PR #21366, opened 2021-04-21
fix CoordSystem._inverse_transformation method

### LINKED ISSUE #21356: `CoordSystem._inverse_transformation` raises a ValueError on any input
Here is an example of an inverse transformation between two coordinate systems, where the forward transformation is a quarter-turn around the origin.

```
import sympy.diffgeom


FIRST = 'first'
SECOND = 'second'

m = sympy.diffgeom.Manifold("M", 2)
p = sympy.diffgeom.Patch("P", m)
RELATIONS = {}

x, y = sympy.symbols("x y")
z, w = sympy.symbols("z w")

matrix = sympy.Matrix([y, - x])
RELATIONS[(FIRST, SECOND)] = sympy.Lambda(
    (x, y),
    sympy.Matrix([y, - x])
)

cartesian2D = sympy.diffgeom.CoordSystem(FIRST, p, [x, y], RELATIONS)
cartesian2D_quarter = sympy.diffgeom.CoordSystem(SECOND, p, [z, w], RELATIONS)


if __name__ == "__main__":
    print(sympy.diffgeom.CoordSystem._inverse_transformation(cartesian2D_quarter, cartesian2D))
```

This example gives me this error: 
```
  File "scratch_3.py", line 26, in <module>
    print(sympy.diffgeom.CoordSystem._inverse_transformation(cartesian2D, cartesian2D_quarter))
  File "sympy\sympy\diffgeom\diffgeom.py", line 361, in _inverse_transformation
    forward_syms, forward_results = forward_transform.args
ValueError: too many values to unpack (expected 2)
```
In this example `forward_transform `is `Matrix([[y], [-x]])`, and `forward_transform.args` is `(2, 1, (y, -x))`, so the ValueError is thrown because the left hand side of the assignment in line 361 of diffgeom.py expects a 2-tuple but the right hand side is a 3-tuple.  The ValueError is thrown for any input because forward_transform.args is always a 3-tuple: the first two elements are the shape of the matrix of transformation equations and the third element is the tuple of matrix transformation equations.

There is a related bug around inverting coordinate systems in the CoordSystem.connect_to method: https://github.com/sympy/sympy/issues/6656

alexcockburn1:
Fixes #21356 

The `CoordSystem._inverse_transformation` method was previously
throwing a ValueError for any input.  This change fixes the broken logic
and adds logic to raise an error if we cannot find an inverse for the
coordinate transformation.

The previous implementation converted variables to Dummy variables
so that all variables in the `solve` method have the same type, but in
this implementation I have converted the variables to 
CoordinateSymbols instead.

#### Release Notes

* diffgeom
  * Fixed a bug in the `CoordSystem._inverse_transformation` method.


Halfogerty:
Is there a way to get this reviewed by someone?  This is my first contribution to any open source project, so please let me know if there's something more I should be doing to get this merged.

oscarbenjamin:
Thanks for contributing this. It sometimes takes a while for anyone to find time to review a pull request.

Personally I don't know the diffgeom module at all so it would be difficult for me to review this. I'll wait and see if anyone else can review it.

JSS95:
I like this fix, but `CoordSystem` needs to be enhanced first. Current design with `Lambda` can be improved by storing the generator `Symbol`s to `CoordSystem`. Then we may not need transformations such as `to_coord_symbol()`.

Halfogerty:
Thanks for taking a look at this.  Could you clarify what you mean by "storing the generator `Symbol`s to `CoordSystem`"?

The way things work at the moment is that a relation_dict is needed to create a `CoordSystem`, and a `CoordSystem` is needed to create a `CoordinateSymbol`.  Therefore using `CoordinateSymbol` in the relation_dict would introduce a cyclic dependency, so we have to use `Symbol` in the relation_dict, and convert between `Symbol` and `CoordinateSymbol`.

Another option for avoiding `to_coord_symbol` would be to add a method to the `CoordSystem` class which returns the symbols as instances of `Symbol`,  in addition to the method which returns the symbols as instances of `CoordinateSymbol`, and only work with `Symbol`s in the `_inverse_tranformation` method.

JSS95:
> Thanks for taking a look at this. Could you clarify what you mean by "storing the generator `Symbol`s to `CoordSystem`"?

It seems that I was confused. `CoordSystem` already stores the symbols. I opened a PR #21464 which makes the signature of `CoordSystem` simpler.

JSS95:
@Halfogerty Will you pull and merge the master branch into this branch, please? Signature for `CoordSystem` is changed a little bit.

Halfogerty:
@JSS95 Sure, I've updated the pull request now.

sympy/diffgeom/diffgeom.py
```diff
@@ -398,26 +398,28 @@ def transformation(self, sys):
...
         return Lambda(signature, expr)
 
     @staticmethod
     def _inverse_transformation(sys1, sys2):
         # Find the transformation relation from sys2 to sys1
-        forward_transform = sys1.transform(sys2)
-        forward_syms, forward_results = forward_transform.args
+        forward_transform_expressions = sys1.transform(sys2)
+        # Coordinate transformations are defined in terms of Symbols rather than CoordSymbols, so must be converted
+        forward_transform_expressions = to_coord_symbol(sys1.symbols, forward_transform_expressions)
```
JSS95, line 411: `sys1.transform(sys2)` already consists of `CoordinateSymbol` so we don't need `to_coord_symbol()` function. (Removing this line showed no different result)

sympy/diffgeom/diffgeom.py
```diff
@@ -398,26 +398,28 @@ def transformation(self, sys):
...
-        inv_syms = [i.as_dummy() for i in forward_syms]
         inv_results = solve(
-            [t[0] - t[1] for t in zip(inv_syms, forward_results)],
-            list(forward_syms), dict=True)[0]
-        inv_results = [inv_results[s] for s in forward_syms]
+            [t[0] - t[1] for t in zip(sys2.symbols, forward_transform_expressions)],
+            list(sys1.symbols), dict=True)
+        if len(inv_results) == 0:
+            raise NotImplementedError(
+                "Cannot solve inverse transformation of {}".format(forward_transform_expressions))
```
JSS95, line 418: I think the error message should tell which systems resulted the failure, e.g. `"Cannot solve inverse transformation between {} and {}".format(sys1, sys2)`.

JSS95, line 418: An error should be raised for ambiguous result (i.e. `len(inv_results) > 1`) as well.

sympy/diffgeom/tests/test_diffgeom.py
```diff
@@ -36,6 +36,34 @@ def test_R2():
...
+        ('first', 'second'): (- a,)
+    }
+
+    R2_a = CoordSystem('first', R2_origin, [a], relations_uninvertible_transformation)
+    R2_b = CoordSystem('second', R2_origin, [b], relations_uninvertible_transformation)
+
+    # The transform method should throw if it cannot invert the coordinate transformation.
+    # This transformation is uninvertible because there is no positive a, b satisfying a = -b
+    with raises(NotImplementedError):
+        R2_b.transform(R2_a)
```
JSS95, line 64: Please remove the space between the negative sign and the expression (`- a` -> `-a`)

sympy/diffgeom/diffgeom.py
Halfogerty: Ah thanks, I hadn't realised that the `transform` method was doing this for me

sympy/diffgeom/diffgeom.py
Halfogerty: Done

Halfogerty: Done

sympy/diffgeom/tests/test_diffgeom.py
Halfogerty: Done

JSS95:
Looks good!

PR #20849, opened 2021-01-22
Make MatchPy a SymPy dependency (egg, pip and conda)

Upabjojr:
Fixes https://github.com/sympy/sympy/issues/20743

#### Brief description of what is fixed or changed

MatchPy is now a dependency of SymPy in:

- `setup.py`
- `.github/workflows`

It will install automatically and be required as a dependency when building the _egg_.

#### Other comments

Removed all workarounds to check if `matchpy` is installed in the code

#### Release Notes

* utilities
  * MatchPy was made a dependency of SymPy.

PR #21367, opened 2021-04-21
refactor(assump) : redesign sathandlers.py

JSS95:

In master, fact for classes are constructed and registered as `UnevaluatedOnFree` object, which tries to behave as a composite function using `rcall` method. The problem is that it requires `Predicate` class to be a subclass of `Boolean`, only to allow the composition between predicates like `(Q.positive & Q.even).rcall(x)` -> `Q.positive(x) & Q.even(x)`. This is mathematically incorrect since predicate is not a boolean but a function which returns boolean. Programmatically, it should be avoided because every type checking for `Boolean` should explicitly exclude `Predicate`.

This change allows handler functions to be registered as fact for classes. It provides more readability and allows first step to make `Predicate` not `Boolean`.

Previously, the registration has been done as
```python
register_fact(Mul, Implies(AllArgs(Q.imaginary | Q.real), Implies(ExactlyOneArg(Q.imaginary), Q.imaginary)))
```

Now, it can be done by 
```python
@class_fact_registry.register(Mul)
def _(expr):
    allargs_imag_or_real = allal
rg(x, Q.imaginary(x) | Q.real(x), expr)
    onearg_imaginary = exactlyonearg(x, Q.imaginary(x), expr)
    return Implies(allargs_imag_or_real, Implies(onearg_imaginary, Q.imaginary(expr)))
```

#### Other comments

`fact_registry` is renamed to `class_fact_registry` because I am planning to introduce another fact registry, `predicate_fact_registry` in next PR. This new registry will store the complicated facts for predicates such as `x > y` implying `x` and `y` being both extended real.

I already tried that change in my local to let `ask(Q.extended_real(x), x>y)` return `True`. However, to keep the diff small that feature is not included in this PR.

Benchmark showed no slowdown by this change.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/assumptions/sathandlers.py
+++ b/sympy/assumptions/sathandlers.py
@@ -2,7 +2,7 @@ from collections import defaultdict
 from collections.abc import MutableMapping
 
 from sympy.assumptions.ask import Q
-from sympy.core import (Add, Mul, Pow, Integer, Number, NumberSymbol,)
+from sympy.core import (Add, Mul, Pow, Integer, Number, NumberSymbol, Symbol)
 from sympy.core.numbers import ImaginaryUnit
 from sympy.functions.elementary.complexes import Abs
 from sympy.logic.boolalg import (Equivalent, Implies, And, Or)
@@ -17,12 +17,24 @@ class ArgFactHandler:
     """
     Class to help apply boolean function to the arguments of expression.
 
+    Parameters
+    ==========
+
+    args : tuple of Symbols.
+        Placeholder symbols representing the arguments of *f*.
+
+    f : Any Boolean expression.
+
     """
-    def __init__(self, func):
-        if not callable(func):
-            raise ValueError(
-                "Argument of ArgFactHandler must be callable, but %s is not." % func)
-        self.func = func
+    def __init__(self, args, f):
+        self.args = args
+        self.f = f
+
+    def __repr__(self):
+        return "%s(%s, %s)" % (type(self).__name__, str(self.args), str(self.f))
+
+    def apply(self, args):
+        return self.f.subs(zip(self.args, args))
 
 
 class AllArgs(ArgFactHandler):
@@ -35,13 +47,17 @@ class AllArgs(ArgFactHandler):
     >>> from sympy import Q
     >>> from sympy.assumptions.sathandlers import AllArgs
     >>> from sympy.abc import x, y
-    >>> allargs = AllArgs(lambda x: Q.negative(x) | Q.positive(x))
+    >>> allargs = AllArgs(x, Q.negative(x) | Q.positive(x))
     >>> allargs(x*y)
     (Q.negative(x) | Q.positive(x)) & (Q.negative(y) | Q.positive(y))
 
     """
+    def __init__(self, x, f):
+        args = (x,)
+        super().__init__(args, f)
+
     def __call__(self, expr):
-        return And(*[self.func(arg) for arg in expr.args])
+        return And(*[self.apply((arg,)) for arg in expr.args])
 
 
 class AnyArgs(ArgFactHandler):
@@ -54,13 +70,17 @@ class AnyArgs(ArgFactHandler):
     >>> from sympy import Q
     >>> from sympy.assumptions.sathandlers import AnyArgs
     >>> from sympy.abc import x, y
-    >>> anyargs = AnyArgs(lambda x: Q.negative(x) & Q.positive(x))
+    >>> anyargs = AnyArgs(x, Q.negative(x) & Q.positive(x))
     >>> anyargs(x*y)
     (Q.negative(x) & Q.positive(x)) | (Q.negative(y) & Q.positive(y))
 
     """
+    def __init__(self, x, f):
+        args = (x,)
+        super().__init__(args, f)
+
     def __call__(self, expr):
-        return Or(*[self.func(arg) for arg in expr.args])
+        return Or(*[self.apply((arg,)) for arg in expr.args])
 
 
 class ExactlyOneArg(ArgFactHandler):
@@ -73,13 +93,17 @@ class ExactlyOneArg(ArgFactHandler):
     >>> from sympy import Q
     >>> from sympy.assumptions.sathandlers import ExactlyOneArg
     >>> from sympy.abc import x, y
-    >>> onearg = ExactlyOneArg(Q.positive)
+    >>> onearg = ExactlyOneArg(x, Q.positive(x))
     >>> onearg(x*y)
     (Q.positive(x) & ~Q.positive(y)) | (Q.positive(y) & ~Q.positive(x))
 
     """
+    def __init__(self, x, f):
+        args = (x,)
+        super().__init__(args, f)
+
     def __call__(self, expr):
-        pred_args = [self.func(arg) for arg in expr.args]
+        pred_args = [self.apply((arg,)) for arg in expr.args]
         res = Or(*[And(pred_args[i], *[~lit for lit in pred_args[:i] +
             pred_args[i+1:]]) for i in range(len(pred_args))])
         return res
@@ -192,6 +216,8 @@ class_fact_registry = ClassFactRegistry()
 
 ### Class fact registration ###
 
+x = Symbol('x')
+
 ## Abs ##
 
 @class_fact_registry.register(Abs)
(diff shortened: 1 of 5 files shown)


oscarbenjamin:
> Previously, the registration has been done as
> 
> ```python
> register_fact(Mul, Implies(AllArgs(Q.imaginary | Q.real), Implies(ExactlyOneArg(Q.imaginary), Q.imaginary)))
> ```
> 
> Now, it can be done by
> 
> ```python
> @class_fact_registry.register(Mul)
> def _(expr):
>     allargs_imag_or_real = AllArgs(lambda x: Q.imaginary(x) | Q.real(x))(expr)
>     onearg_imaginary = ExactlyOneArg(Q.imaginary)(expr)
>     return Implies(allargs_imag_or_real, Implies(onearg_imaginary, Q.imaginary(expr)))
> ```

This seems quite verbose. Can it not just be done by calling the predicates on a symbol? So then
```
AllArgs(Q.imaginary | Q.real)   -->   AllArgs(x, Q.imaginary(x) | Q.real(x))
```

JSS95:
@oscarbenjamin You mean that applying the handler on `expr` be
```python
AllArgs(x, Q.imaginary(x) | Q.real(x)).subs(x, expr)
```
right?

JSS95:
@oscarbenjamin Done.

sympy/assumptions/sathandlers.py
```diff
@@ -2,405 +2,464 @@
...
+
+@class_fact_registry.register(Pow)
+def _(expr):
+    base, exp = expr.base, expr.exp
+    return Implies(Q.nonpositive(base) & Q.odd(exp) & Q.nonnegative(exp), Q.nonpositive(expr))
+
+@class_fact_registry.register(Pow)
+def _(expr):
+    base, exp = expr.base, expr.exp
+    return Equivalent(Q.zero(expr), Q.zero(base) & Q.positive(exp))
```
oscarbenjamin, line 385: Would it make sense to combine these like:
```python
@class_fact_registry.register_many(Pow)
def _(expr):
    base, exp = expr.base, expr.exp
    return [
        Implies(Q.real(base) & Q.even(exp) & Q.nonnegative(exp), Q.nonnegative(expr)),
        Implies(Q.nonnegative(base) & Q.odd(exp) & Q.nonnegative(exp), Q.nonnegative(expr)),
        Implies(Q.nonpositive(base) & Q.odd(exp) & Q.nonnegative(exp), Q.nonpositive(expr)),
        Equivalent(Q.zero(expr), Q.zero(base) & Q.positive(exp)),
    ]
```
Maybe it would also be clearer to use `>>` for implies e.g.:
```python
(Q.real(base) & Q.even(exp) & Q.nonnegative(exp))  >>  Q.nonnegative(expr)
```

sympy/assumptions/sathandlers.py
```diff
@@ -198,6 +199,17 @@ class ClassFactRegistry(FactRegistry):
     >>> reg(Abs(x))
     {Q.nonnegative(Abs(x)), Equivalent(~Q.zero(x), ~Q.zero(Abs(x)))}
 
+    Multiple facts can be registered at once by letting handler return an
+    iterable.
```
oscarbenjamin, line 218: I think it's better to have a separate function for registering multiple handlers rather than checking based on `iterable`.

sympy/assumptions/sathandlers.py
```diff
@@ -220,62 +241,28 @@ def __call__(self, expr):
...
-@class_fact_registry.register(Add)
-def _(expr):
-    allargs_integer = AllArgs(x, Q.integer(x))(expr)
-    return Implies(allargs_integer, Q.integer(expr))
-
-@class_fact_registry.register(Add)
-def _(expr):
-    onearg_notinteger = ExactlyOneArg(x, ~Q.integer(x))(expr)
-    return Implies(onearg_notinteger, ~Q.integer(expr))
+    return [AllArgs(x, Q.positive(x))(expr) >> Q.positive(expr),
```
oscarbenjamin, line 274: Rather than having `AllArgs` be callable would it not make more sense just to combine this to something like:
```
AllArgs(expr, x, Q.positive(x))
```

sympy/assumptions/sathandlers.py
```diff
@@ -367,99 +330,43 @@ def _(expr):
...
 def _(expr):
     from sympy import isprime
-    return Equivalent(Q.prime(expr), isprime(expr))
-
-@class_fact_registry.register(Integer)
-def _(expr):
-    return Equivalent(Q.composite(expr), expr.is_composite)
-
+    return [Equivalent(Q.prime(expr), isprime(expr)),
+            Equivalent(Q.composite(expr), expr.is_composite),
```
oscarbenjamin, line 362: What exactly is this doing?

As far as I can tell `isprime` just returns True or False. It returns False for an arbitrary `Symbol('x')` which doesn't seem right here.

Is it necessary to make a fact saying that `Q.composite` is equivalent to `expr.is_composite`? I thought that there already was a mechanism for checking the old assumptions somewhere else.

sympy/assumptions/sathandlers.py
JSS95: How would it be possible, when we are registering a single function `f`?

JSS95: Oh, maybe we can have a seperate container which stores these kind of functions.

JSS95: There are 4 situations:
1. Register a fact to a class
2. Register a fact to classes
3. Register facts to a class
4. Register facts to classes

Instead of having 4 methods, would it make more sense to have one method with argument(s) for classes and option for single/multiple facts?

sympy/assumptions/sathandlers.py
JSS95: This is registered for `Integer`, not `Symbol`. Basically, this tells satask that `Q.prime(2)` is `True`.

sympy/assumptions/sathandlers.py
oscarbenjamin: Maybe we just need one that can register a single fact to a single class and another that registers multiple facts to multiple classes.

sympy/assumptions/sathandlers.py
oscarbenjamin: Would `expr.is_prime` not do that? In other words is this not just handled by having `Q.prime` querying into the old assumptions?

I thought that the new assumptions checked the old somewhere so we wouldn't need to explicitly register handlers for anything covered by the old assumptions.

I'm not saying this should be changed. Just trying to understand...

sympy/assumptions/sathandlers.py
JSS95: Done.

sympy/assumptions/sathandlers.py
JSS95: Done. `AllArgs` and other helpers do not need to be class anymore, so I changed them to function.

sympy/assumptions/sathandlers.py
JSS95: It seems this can be replaced by the said mechanism.

oscarbenjamin:
Looks good to me

asmeurer:
What was the purpose of this? This changes things to be much more verbose and unreadable in my view, and the previous method had the advantage of being fully symbolic, so you could just print the facts and see them (and also potentially do other things in the future, like bisecting facts to find an unsatisfiable core). 

I would rather see the facts for sathandlers be rewritten in a concise but still readable way, such as using pattern matching. That would also avoid the unapplied predicate "problem" (I'm still not convinced this is really a big deal) because every fact would be evaluated at a Wild symbol. For example simply `Q.positive(Abs(x))` is much simpler than 

```py
@class_fact_registry.multiregister(Abs)
def _(expr):
    arg = expr.args[0]
    return [Q.nonnegative(expr),
    ...
```

For Mul and Add, getting the syntactic sugar right would be harder, and we might have to use something like this instead.

One should consider that the current number of facts in satask is a couple of orders of magnitude less than what could be there, if we were to move all facts in the assumptions system over to it (which we should, if we can get the performance of it good enough). So the method of writing down facts should be concise and readable even if there are hundreds of them. I think there's real potential for the new assumptions to be better than the old in this regard, and it's important because it's really easy to get the math wrong in assumptions facts, so the easier they are to read, the better.

sympy/assumptions/sathandlers.py
```diff
@@ -1,406 +1,324 @@
...
-    result = set()
-    for arg in args:
-        result.update(_find_freepredicate(arg))
-    return result
-
-
-class UnevaluatedOnFree(BooleanFunction):
+### Helper functions ###
+
+def allarg(symbol, fact, expr):
```
asmeurer, line 15: This should be `allargs` plural.

asmeurer:
I do like the allarg(s), anyarg, and so on names better, but they should be unevaluated, so that we can just create expressions, rather than forcing everything to be in a lambda function.

oscarbenjamin:
> For example simply `Q.positive(Abs(x))` is much simpler than

Yes, that is a nicer way of doing it. I haven't looked into the internals of this but how are the facts stored when used? I thought that the purpose of these decorators was to be used with multiple dispatch or something but if the facts are just being stored as a big database of "expressions" then it would definitely be better to create them like that.

JSS95:
Wouldn't using pattern matching be slow?

JSS95:
@asmeurer
1. Fact searching using function (current approach)
```python
In [1]: def f(expr): 
   ...:     x = expr.args[0] 
   ...:     return Q.even(x) >> Q.even(expr) 
   ...: 

In [2]: %timeit f(Abs(x))
6.17 µs ± 62.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

2. Fact searching using pattern matching
```python
In [1]: a = Wild('a') 
   ...: def g(expr): 
   ...:     return (Q.even(a) >> Q.even(Abs(a))).subs(expr.match(Abs(a))) 
   ...:

In [2]: %timeit g(Abs(x))
71.7 µs ± 54.7 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Two functions are designed that both `f(Abs(x))` and `g(Abs(x))` return `Implies(Q.even(x), Q.even(Abs(x)))`. Using Wild symbol is more that 10 times slower, but if you think the benefit compensates the slowdown then I am willing to implement your suggestion.

oscarbenjamin:
The pattern matching approach would potentially make more sense with something like matchpy where we match many patterns at once: https://github.com/sympy/sympy/pull/20849

Potentially if there are a large number of patterns to be matched against an expression in order to extract facts then an efficient pattern matcher can be constructed.

In any case I'm not sure that @asmeurer literally meant pattern matching as part of the assumptions query. I think the point is more that we need to find the most intelligible way of writing down the assumptions facts to make them as easy to understand as possible. Getting the three-valued logic of `_eval_is_positive` etc correct is very difficult in the old assumptions. The idea with the new assumptions is to make the style much more declarative so that it's easy to look at it and see that it's correct. Of course under the hood all those facts can be converted into some other representation that's more efficient for whatever style of matching we want to use.

JSS95:
I am open to any suggestion for storing the facts, as long as we keep `Predicate` not `Boolean`.

PR #21375, opened 2021-04-24
refact(assump) : parametrize the facts

JSS95:

In master, relations between unary predicates are constructed without parametrization such as `Implies(Q.even, ~Q.odd)`, which requires `Predicate` class to be a subclass of `Boolean`.

This is mathematically incorrect since predicate is not a boolean but a function which returns boolean. Programmatically, it should be avoided because every type checking for `Boolean` should explicitly exclude `Predicate` with this design.

This change makes the facts registered with parametrization such as `Implies(Q.even(x), ~Q.odd(x))`. It provides one more step to make `Predicate` not `Boolean`. (Previous step was #21367)

#### Other comments

Benchmark showed no slowdown by this change.

#### Release Notes

NO ENTRY


oscarbenjamin:
These facts should be rearranged/improved. I'll make some notes...

sympy/assumptions/facts.py
```diff
@@ -34,71 +36,138 @@ def get_composite_predicates():
...
     # DO NOT include the predicates in get_composite_predicates()'s keys here!
-    return And(
-
+    x = Symbol('x')
+    fact = And(
         # primitive predicates exclude each other
-        Implies(Q.negative_infinite, ~Q.positive_infinite),
-        Implies(Q.negative, ~Q.zero & ~Q.positive),
-        Implies(Q.positive, ~Q.zero),
+        Implies(Q.negative_infinite(x), ~Q.positive_infinite(x)),
```
oscarbenjamin, line 55: The reverse implication is not listed. Both can be expressed in one rule as:
```python
Not(Q.negative_infinite(x) & Q.positive_infinite(x))
```

sympy/assumptions/facts.py
```diff
@@ -34,71 +36,138 @@ def get_composite_predicates():
...
+        Implies(Q.negative(x), ~Q.zero(x) & ~Q.positive(x)),
+        Implies(Q.positive(x), ~Q.zero(x)),
```
oscarbenjamin, line 57: All of these can be collected together as e.g.
```python
Q.negative_infinite(x) => ~(Q.negative(x) | Q.zero(x) | Q.positive(x) | Q.positive_infinite(x))
Q.negative(x) => ~(...)
...
```

sympy/assumptions/facts.py
```diff
@@ -34,71 +36,138 @@ def get_composite_predicates():
...
 
         # build real line and complex plane
-        Implies(Q.negative | Q.zero | Q.positive, ~Q.imaginary),
-        Implies(Q.negative | Q.zero | Q.positive | Q.imaginary, Q.algebraic | Q.transcendental),
+        Implies(Q.negative(x) | Q.zero(x) | Q.positive(x), ~Q.imaginary(x)),
+        Implies(Q.negative(x) | Q.zero(x) | Q.positive(x) | Q.imaginary(x),
+                Q.algebraic(x) | Q.transcendental(x)),
```
oscarbenjamin, line 62: This should just be
```python
Q.complex(x) >> (Q.algebraic(x) | Q.transcendental(x))
```

sympy/assumptions/facts.py
```diff
@@ -34,71 +36,138 @@ def get_composite_predicates():
...
 
         # other subsets of complex
-        Implies(Q.transcendental, ~Q.algebraic),
-        Implies(Q.irrational, ~Q.rational),
-        Equivalent(Q.rational | Q.irrational, Q.negative | Q.zero | Q.positive),
-        Implies(Q.rational, Q.algebraic),
+        Implies(Q.transcendental(x), ~Q.algebraic(x)),
```
oscarbenjamin, line 65: This implication is in both directions so it should be
```python
Not(Q.transcendental(x) & Q.algebraic(x))
```

sympy/assumptions/facts.py
```diff
@@ -34,71 +36,138 @@ def get_composite_predicates():
...
+        Implies(Q.irrational(x), ~Q.rational(x)),
+        Equivalent(Q.rational(x) | Q.irrational(x),
+                Q.negative(x) | Q.zero(x) | Q.positive(x)),
```
oscarbenjamin, line 68: I think it would be useful to be able to use the composite predicates in these facts like:
```python
Equivalent(Q.real(x), Q.rational(x) | Q.irrational(x))
```

sympy/assumptions/facts.py
```diff
@@ -34,71 +36,138 @@ def get_composite_predicates():
...
+        Implies(Q.rational(x), Q.algebraic(x)),
 
         # integers
-        Implies(Q.even, ~Q.odd),
-        Implies(Q.even | Q.odd, Q.rational),
-        Implies(Q.zero, Q.even),
-        Implies(Q.composite, ~Q.prime),
-        Implies(Q.composite | Q.prime, (Q.even | Q.odd) & Q.positive),
-        Implies(Q.even & Q.positive & ~Q.prime, Q.composite),
+        Implies(Q.even(x), ~Q.odd(x)),
```
oscarbenjamin, line 72:
```
Not(Q.even(x) & Q.odd(x))
```

sympy/assumptions/facts.py
```diff
@@ -34,71 +36,138 @@ def get_composite_predicates():
...
+        Implies(Q.even(x) | Q.odd(x), Q.rational(x)),
```
oscarbenjamin, line 73: This should be
```python
Equivalent(Q.integer(x), Q.even(x) | Q.odd(x))
```

sympy/assumptions/facts.py
```diff
@@ -34,71 +36,138 @@ def get_composite_predicates():
...
+        Implies(Q.zero(x), Q.even(x)),
+        Implies(Q.composite(x), ~Q.prime(x)),
+        Implies(Q.composite(x) | Q.prime(x),
+            (Q.even(x) | Q.odd(x)) & Q.positive(x)),
```
oscarbenjamin, line 77: Surely this should be `Q.integer(x) & Q.positive(x)`. Or is integer considered a composite predicate?

sympy/assumptions/facts.py
```diff
@@ -34,71 +36,138 @@ def get_composite_predicates():
...
+        Implies(Q.negative(x) | Q.zero(x) | Q.positive(x),
+            Q.hermitian(x)),
+        Implies(Q.imaginary(x), Q.antihermitian(x)),
+        Implies(Q.zero(x), Q.hermitian(x) | Q.antihermitian(x)),
 
         # define finity and infinity, and build extended real line
-        Implies(Q.infinite, ~Q.finite),
-        Implies(Q.algebraic | Q.transcendental, Q.finite),
-        Implies(Q.negative_infinite | Q.positive_infinite, Q.infinite),
+        Implies(Q.infinite(x), ~Q.finite(x)),
```
oscarbenjamin, line 87:
```
Not(Q.infinite(x) & Q.finite(x))
```

sympy/assumptions/facts.py
```diff
@@ -34,71 +36,138 @@ def get_composite_predicates():
...
+        Implies(Q.algebraic(x) | Q.transcendental(x), Q.finite(x)),
```
oscarbenjamin, line 88: This is not needed if we recognised that `Q.complex` is equivalent and implies finite.

sympy/assumptions/facts.py
```diff
@@ -34,71 +36,138 @@ def get_composite_predicates():
...
+        Implies(Q.negative_infinite(x) | Q.positive_infinite(x), Q.infinite(x)),
 
         # commutativity
-        Implies(Q.finite | Q.infinite, Q.commutative),
+        Implies(Q.finite(x) | Q.infinite(x), Q.commutative(x)),
```
oscarbenjamin, line 92: This is extremely weird. I think maybe there used to be something like this in the old assumptions but I got rid of it and just made `finite` and `infinite` mutually exclusive. There is a need in the old assumptions for the `is_commutative` attribute although perhaps it should be completely separate from the assumptions system. I don't think `Q.commutative` is needed in the new assumptions. Is there a clear definition of what it is supposed to even mean?

oscarbenjamin:
I made some comments about the facts themselves since you are working on these but I realise that is somewhat orthogonal to this PR which is about modifying `Predicate`.

The changes here look good to me. @asmeurer do you want to review?

sympy/assumptions/facts.py
oscarbenjamin: In fact this is an equivalence rather than an implication

asmeurer:
Yeah, I would make changes that simplify/modify how the facts are written in a separate PR, for simplicity.

sympy/assumptions/facts.py
asmeurer: `Implies(a, ~b)` and `Not(a & b)` are logically equivalent. I guess the latter might be easier to reason about. I'm +0 to changing these facts like this. to_cnf should create the same thing at the end of the day for both. 

Note that this isn't quite the same as an implication in both directions. `Implies(a, ~b)` and `Implies(a, ~b)` are equivalent by contrapositive, but `a iff ~b` (`Equivalent(a, ~b)`) is a stronger statement. It doesn't hold here (and in similar cases) because x might be neither algebraic nor transcendental if it isn't a complex number.

sympy/assumptions/facts.py
asmeurer: The reverse implication is already known

```py
>>> satask(Q.complex(x) << (Q.algebraic(x) | Q.transcendental(x)))
True
>>> satask(Q.algebraic(x) | Q.transcendental(x), Q.complex(x))
True
```

(although browsing through the facts it isn't clear to me exactly why)

sympy/assumptions/facts.py
asmeurer: Also given that the actual facts are compiled with ask_update, it doesn't really matter what form of an expression we use, or if we are redundant with facts. Redundancy might be potentially confusing, but it wouldn't slow things down because of this.

sympy/assumptions/facts.py
```diff
@@ -34,71 +36,138 @@ def get_composite_predicates():
...
+    x : Placeholder symbol for unary facts.
+
+    fact : Known facts in conjugated normal form.
+
+    """
     # We build the facts starting with primitive predicates.
     # DO NOT include the predicates in get_composite_predicates()'s keys here!
-    return And(
-
+    x = Symbol('x')
```
asmeurer, line 52: Instead of changing the return value, it would be simpler to add a default parameter to the function, like `get_known_facts(x=Symbol('x'))`.

asmeurer: Or it might have to be

```py
def get_known_facts(x=None):
    if x is None:
        x = Symbol('x')
```

to avoid using Symbol at the module level.

asmeurer:
This changes the API of get_known_facts, but I guess if we are planning to make bare predicates invalid (?), that will need to happen anyway. get_known_facts is a function that it isn't completely clear if it should be considered public API or not. 

At any rate, the changes here all look fine to me, except for the comment I made about passing x in as a parameter instead of returning it.

sympy/assumptions/facts.py
oscarbenjamin: Yes, I suppose they are equivalent. It seems that a number of facts come into this category where what you want to say is sort of like
```
complex  ==  transcendental xor algebraic 
```
I'm wondering what is the best way to write that so that it's completely clear and clearly symmetric. Maybe:
```
Equivalent(complex, algebraic | transcendental)
Not(algebraic & transcendental)
```
It feels like there should be a simpler way to say this in one somehow.

sympy/assumptions/facts.py
JSS95: This means that every number (extended real numbers, composite numbers and `zoo`) is finite.

`finite` and `infinite` are not always mutually exclusive, since `S.NaN` is not finite nor infinite.

I'm +1 that we do not need `Q.commutative` in the new assumptions. IMO commutativity must be dealt by kind system and shouldn't be included in old/new assumption system. Removing it from old assumptions is impossible since doing so will raise infinite backwards compatibility issues. But I don't think `Q.commutative` is really used anywhere. Perhaps we can just remove it?

oscarbenjamin: Commutative could be removed from the old assumptions in the sense that it can be an attribute/property but just not part of the assumptions system itself. I'm not sure if there is any significant benefit in changing that though.

> `finite` and `infinite` are not always mutually exclusive, since `S.NaN` is not finite nor infinite.

I think NaN can give None for both cases. In all cases where finite=True we should have infinite=False and vice versa so they are mutually exclusive. This particular rule is I think copied from the old assumptions but I have since fixed this in the old assumptions:
https://github.com/sympy/sympy/blob/49df83fdcc04715525b635bb9157113b94d11122/sympy/core/assumptions.py#L257

sympy/assumptions/facts.py
JSS95: I think `Not(Q.negative_infinite(x) & Q.positive_infinite(x))` will be enough instead of `Q.negative_infinite(x) >> ~(Q.negative(x) | Q.zero(x) | Q.positive(x) | Q.positive_infinite(x))`, since finite and and infinite predicate reject each other. Would you take a look at how I changed this?

sympy/assumptions/facts.py
JSS95: `finite` and `infinite` are now mutually exclusive. What should be done with `Q.commutative`, then?

oscarbenjamin: I'm not sure about `Q.commutative` so I would suggest just leaving it for now while working on the refactoring.

sympy/assumptions/facts.py
JSS95: To be clear; `Q.negative` already excludes `Q.negative_infinite` and `Q.positive_infinite` because
1. `Q.negative` implies `Q.finite`
2. Q.negative_infinite` implies `Q.infinite`
3. `Q.finite` and `Q.infinite` are mutually exclusive

oscarbenjamin: I think what you have is fine and I as @asmeurer says I don't think it is good to make non refactoring changes in a refactoring PR. It makes it messy in future with git bisect etc.

My point though (for another PR) is that there are many situations in the facts where we have a predicate that is equivalent to an or of a bunch mutually exclusive predicates like:
```
Equivalent(A, B | C)
Not(B & C)
```
With more variables that is
```
Equivalent(A, B | C | D | E)
Not(B & C)
Not(B & D)
Not(B & E)
Not(C & D)
Not(C & E)
Not(D & E)
```
It would be clearer and more compact if there was a way of saying that some set of predicates is mutually exclusive like:
```
Exclusive(B, C, D, E)
Equivalent(A, B | C | D | E)
```
Perhaps those could be combined like:
```
EquivalentExactlyOne(A, [B, C, D, E])
```
but maybe that's too much. You could use that like:
```
EquivalentExactlyOne(Q.complex(x), [Q.algebraic(x), Q.transcendental(x)])
EquivalentExactlyOne(Q.real(x), [Q.rational(x), Q.irrational(x)])
EquivalentExactlyOne(Q.integer(x), [Q.even(x), Q.odd(x)])

EquivalentExactlyOne(Q.extended_real(x), [
    Q.negative_infinite(x), Q.negative(x), Q.zero(x), Q.positive(x), Q.positive_infinite(x)])
```

Note that n-ary `Xor` doesn't work like `Exclusive`:
```
In [2]: Xor(True, True, True)
Out[2]: True
```

JSS95: The lines are modified anyway by parametrization, so I'll leave them as they are in 10ee1384279f3b903aab0ac17e4b178d4821b52d.

JSS95:
This is ready to be merged.

sympy/assumptions/facts.py
JSS95: I prefer `Not(A & B)` than `Implies(A, ~B)` for its symmetry, so I modified other facts this way and squashed the commits to make the history clean. I'll mark this conversation as resolved.

oscarbenjamin:
It's better not to make changes to the facts in the same PR as the refactoring. Squashing those changes into a single commit is worse though. If a bug bisects to this commit then it might be difficult to identify the change that is relevant.

Can you separate this into two commits where one is the refactoring and the other is the changes to the facts? Actually it would be better to do those in two separate PRs. Can this PR just be the refactoring? Then the changes to the facts can be reviewed more carefully.

JSS95:
@oscarbenjamin It's done. I'll make fact-refactoring PR after this one is merged (since refactored facts needs to be parametrized).

oscarbenjamin:
When one commit reverts changes from the previous commit it is better to squash them

JSS95:
@oscarbenjamin  Done.

JSS95:
I will wait for a few more days and merge if there is no more review.

moorepants:
FYI, master branch CI is failing and this was the last merged PR.

oscarbenjamin:
The CI failure is on Travis and is the Sphinx build failing with these warnings:
```
/home/travis/build/sympy/sympy/sympy/categories/diagram_drawing.py:docstring of sympy.categories.diagram_drawing.DiagramGrid:126: WARNING: more than one target found for cross-reference 'Diagram': sympy.categories.Diagram, sympy.categories.baseclasses.Diagram
/home/travis/build/sympy/sympy/sympy/categories/diagram_drawing.py:docstring of sympy.categories.diagram_drawing.XypicDiagramDrawer:3: WARNING: more than one target found for cross-reference 'Diagram': sympy.categories.Diagram, sympy.categories.baseclasses.Diagram
/home/travis/build/sympy/sympy/sympy/core/mul.py:docstring of sympy.core.mul.Mul:62: WARNING: more than one target found for cross-reference 'MatMul': sympy.matrices.expressions.MatMul, sympy.matrices.expressions.matmul.MatMul
/home/travis/build/sympy/sympy/sympy/core/add.py:docstring of sympy.core.add.Add:78: WARNING: more than one target found for cross-reference 'MatAdd': sympy.matrices.expressions.MatAdd, sympy.matrices.expressions.matadd.MatAdd
/home/travis/build/sympy/sympy/doc/src/modules/matrices/immutablematrices.rst:11: WARNING: more than one target found for cross-reference 'MatrixExpr': sympy.matrices.expressions.MatrixExpr, sympy.matrices.expressions.matexpr.MatrixExpr
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:573: WARNING: more than one target found for cross-reference 'FiniteField': sympy.polys.domains.FiniteField, sympy.polys.domains.finitefield.FiniteField
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:573: WARNING: more than one target found for cross-reference 'PythonFiniteField': sympy.polys.domains.PythonFiniteField, sympy.polys.domains.pythonfinitefield.PythonFiniteField
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:573: WARNING: more than one target found for cross-reference 'GMPYFiniteField': sympy.polys.domains.GMPYFiniteField, sympy.polys.domains.gmpyfinitefield.GMPYFiniteField
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:670: WARNING: more than one target found for cross-reference 'AlgebraicField': sympy.polys.domains.AlgebraicField, sympy.polys.domains.algebraicfield.AlgebraicField
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:961: WARNING: more than one target found for cross-reference 'PolynomialRing': sympy.polys.domains.PolynomialRing, sympy.polys.domains.polynomialring.PolynomialRing
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:1026: WARNING: more than one target found for cross-reference 'RationalField': sympy.polys.domains.RationalField, sympy.polys.domains.rationalfield.RationalField
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:1082: WARNING: more than one target found for cross-reference 'RationalField': sympy.polys.domains.RationalField, sympy.polys.domains.rationalfield.RationalField
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:1163: WARNING: more than one target found for cross-reference 'PolynomialRing': sympy.polys.domains.PolynomialRing, sympy.polys.domains.polynomialring.PolynomialRing
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:1163: WARNING: more than one target found for cross-reference 'RationalField': sympy.polys.domains.RationalField, sympy.polys.domains.rationalfield.RationalField
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:1187: WARNING: more than one target found for cross-reference 'ExpressionDomain': sympy.polys.domains.ExpressionDomain, sympy.polys.domains.expressiondomain.ExpressionDomain
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:1187: WARNING: more than one target found for cross-reference 'AlgebraicField': sympy.polys.domains.AlgebraicField, sympy.polys.domains.algebraicfield.AlgebraicField
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:1201: WARNING: more than one target found for cross-reference 'PolynomialRing': sympy.polys.domains.PolynomialRing, sympy.polys.domains.polynomialring.PolynomialRing
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:1201: WARNING: more than one target found for cross-reference 'RationalField': sympy.polys.domains.RationalField, sympy.polys.domains.rationalfield.RationalField
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:1392: WARNING: more than one target found for cross-reference 'AlgebraicField': sympy.polys.domains.AlgebraicField, sympy.polys.domains.algebraicfield.AlgebraicField
/home/travis/build/sympy/sympy/doc/src/modules/polys/domainsintro.rst:1538: WARNING: more than one target found for cross-reference 'AlgebraicField': sympy.polys.domains.AlgebraicField, sympy.polys.domains.algebraicfield.AlgebraicField
/home/travis/build/sympy/sympy/sympy/polys/domains/domain.py:docstring of sympy.polys.domains.domain.Domain:84: WARNING: more than one target found for cross-reference 'PolynomialRing': sympy.polys.domains.PolynomialRing, sympy.polys.domains.polynomialring.PolynomialRing
/home/travis/build/sympy/sympy/sympy/polys/domains/algebraicfield.py:docstring of sympy.polys.domains.algebraicfield.AlgebraicField:155: WARNING: more than one target found for cross-reference 'AlgebraicNumber': sympy.polys.numberfields.AlgebraicNumber, sympy.core.numbers.AlgebraicNumber
```
I don't think this PR changed those. It might be that we are using the wrong kind of cross-reference somewhere. I don't understand Sphinx referencing very well though...

PR #21380, opened 2021-04-25
functions: Implemented an _eval_expand_trig method for tanh

### LINKED ISSUE #21365: `expand_trig` does not work with `tanh`
Handler for `expand_trig` could be missing for tanh
https://en.wikipedia.org/wiki/Hyperbolic_functions#Sums_of_arguments
```
>>> expand_trig(tan(x + y))
(tan(x) + tan(y))/(-tan(x)*tan(y) + 1)
>>> expand_trig(tanh(x + y))
tanh(x + y)
```

oostben:
**References to other Issues or PRs**
Fixes #21365

**Brief description of what is fixed or changed**
- Implemented _eval_expand_trig  method for tanh funciton. expand_trig now works to expand tanh in a few different ways.
- tanh(coeff*x) will be expanded to coeff\*tanh(x).
- tanh now detects double and half angle formulas, so if coeff is 2 or .5, it will be expanded according to the following formulas:
    - tanh Half Angle Formula: **tanh(.5\*x) = (cosh(x)-1)/sinh(x)**
    - tanh Double Angle Formula: **tanh(2x) = 2\*tanh(x)/(1+tanh^2(x))**
- _eval_expand_trig also expands tanh(x+y) using the sum of arguments identity.
    - tanh Sum of Arguments Identity: **tanh(x+y) = (tanh(x)+tanh(y))/(1+tanh(x)\*tanh(y))**

**Release Notes**

* functions
  * Add _eval_expand_trig method for tanh. 

Example:
```
>>> from sympy import *
>>> x = Symbol('x')
>>> y = Symbol('y')

>>> expand_trig(tanh(x + y))
(tanh(x) + tanh(y))/(tanh(x)*tanh(y) + 1)

>>> expand_trig(tanh(2*x))
2*tanh(x)/(tanh(x)**2 + 1)

>>> expand_trig(tanh(3*x))
3*tanh(x)

>>> expand_trig(tanh(y*x))
tanh(x*y)

>>> expand_trig(tanh((1/2)*x))
(cosh(x) - 1)/sinh(x)
```

### DIFF
--- a/sympy/functions/elementary/hyperbolic.py
+++ b/sympy/functions/elementary/hyperbolic.py
@@ -7,6 +7,7 @@ from sympy.functions.combinatorial.factorials import factorial, RisingFactorial
 from sympy.functions.elementary.exponential import exp, log, match_real_imag
 from sympy.functions.elementary.miscellaneous import sqrt
 from sympy.functions.elementary.integers import floor
+from sympy.utilities.iterables import numbered_symbols
 
 from sympy.core.logic import fuzzy_or, fuzzy_and
 
@@ -623,6 +624,40 @@ class tanh(HyperbolicFunction):
             re, im = self.args[0].as_real_imag()
         denom = sinh(re)**2 + cos(im)**2
         return (sinh(re)*cosh(re)/denom, sin(im)*cos(im)/denom)
+
+    def _eval_expand_trig(self, **hints):
+        arg = self.args[0]
+        x = None
+        if arg.is_Add or arg.is_Mul:
+            from sympy import symmetric_poly
+            n = len(arg.args)
+            TX = []
+            if arg.is_Add:
+                for x in arg.args:
+                    tx = tanh(x, evaluate=False)._eval_expand_trig()
+                    TX.append(tx)
+            else:
+                coeff, terms = arg.as_coeff_Mul()
+                #double angle formula
+                if coeff.is_Integer and coeff == 2:
+                    tx = tanh(terms, evaluate=False)._eval_expand_trig()
+                    TX.append(tx)
+                    TX.append(tx)
+                #half angle formula
+                elif coeff == (1/2):
+                    cx = cosh(terms, evaluate=False)._eval_expand_trig()
+                    sx = sinh(terms, evaluate=False)._eval_expand_trig()
+                    return (cx-1)/sx
+                else:
+                    return coeff*tanh(terms)
+            Yg = numbered_symbols('Y')
+            Y = [ next(Yg) for i in range(n) ]
+
+            p = [0, 0]
+            for i in range(n + 1):
+                p[1 - i % 2] += symmetric_poly(i, Y)*(1)**((i % 4)//2)
+            return (p[0]/p[1]).subs(list(zip(Y, TX)))
+        return tanh(arg)
 
     def _eval_rewrite_as_tractable(self, arg, limitvar=None, **kwargs):
         neg_exp, pos_exp = exp(-arg), exp(arg)
--- a/sympy/functions/elementary/hyperbolic.py
+++ b/sympy/functions/elementary/hyperbolic.py
@@ -649,7 +649,7 @@ class tanh(HyperbolicFunction):
                     sx = sinh(terms, evaluate=False)._eval_expand_trig()
                     return (cx-1)/sx
                 else:
-                    return coeff*tanh(terms)
+                    return tanh(coeff*terms)
             Yg = numbered_symbols('Y')
             Y = [ next(Yg) for i in range(n) ]
 
--- a/sympy/functions/elementary/hyperbolic.py
+++ b/sympy/functions/elementary/hyperbolic.py
@@ -7,7 +7,6 @@ from sympy.functions.combinatorial.factorials import factorial, RisingFactorial
 from sympy.functions.elementary.exponential import exp, log, match_real_imag
 from sympy.functions.elementary.miscellaneous import sqrt
 from sympy.functions.elementary.integers import floor
-from sympy.utilities.iterables import numbered_symbols
 
 from sympy.core.logic import fuzzy_or, fuzzy_and
 
@@ -637,24 +636,22 @@ class tanh(HyperbolicFunction):
                     tx = tanh(x, evaluate=False)._eval_expand_trig()
                     TX.append(tx)
 
-                Yg = numbered_symbols('Y')
-                Y = [ next(Yg) for i in range(n) ]
-
-                p = [0, 0]
+                p = [0, 0]  # [den, num]
                 for i in range(n + 1):
-                    p[1 - i % 2] += symmetric_poly(i, Y)
-                return (p[0]/p[1]).subs(list(zip(Y, TX)))
+                    p[i % 2] += symmetric_poly(i, TX)
+                return p[1]/p[0]
         else:
             from sympy.functions.combinatorial.numbers import nC
             coeff, terms = arg.as_coeff_Mul()
             if coeff.is_Integer and coeff > 1:
-                numerator = 0
-                denominator = 0
-                for k in range(0,floor((coeff-1)/2)+1):
-                    numerator += nC(range(coeff),2*k+1) * tanh(terms)**(2*k+1)
-                for k in range(0,floor(coeff/2)+1):
-                    denominator += nC(range(coeff),(2*k)) * tanh(terms)**(2*k)
-                return numerator / denominator
+                n = []
+                d = []
+                T = tanh(terms)
+                for k in range(1, coeff + 1, 2):
+                    n.append(nC(range(coeff), k)*T**k)
+                for k in range(0, coeff + 1, 2):
+                    d.append(nC(range(coeff), k)*T**k)
+                return Add(*n)/Add(*d)
         return tanh(arg)
 
     def _eval_rewrite_as_tractable(self, arg, limitvar=None, **kwargs):
--- a/sympy/functions/elementary/hyperbolic.py
+++ b/sympy/functions/elementary/hyperbolic.py
@@ -637,7 +637,7 @@ class tanh(HyperbolicFunction):
                 for x in arg.args:
                     tx = tanh(x, evaluate=False)._eval_expand_trig()
                     TX.append(tx)
-                
+
                 Yg = numbered_symbols('Y')
                 Y = [ next(Yg) for i in range(n) ]
 
@@ -647,7 +647,7 @@ class tanh(HyperbolicFunction):
                 return (p[0]/p[1]).subs(list(zip(Y, TX)))
             else:
                 coeff, terms = arg.as_coeff_Mul()
-                #tanh(n*z) (general case)
+                #tanh(n*x) (general case)
                 if coeff.is_Integer and coeff > 1:
                     numerator = 0
                     denominator = 0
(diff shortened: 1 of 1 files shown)


sympy/functions/elementary/hyperbolic.py
```diff
@@ -624,6 +625,40 @@ def as_real_imag(self, deep=True, **hints):
...
+                    sx = sinh(terms, evaluate=False)._eval_expand_trig()
+                    return (cx-1)/sx
+                else:
+                    return coeff*tanh(terms)
+            Yg = numbered_symbols('Y')
+            Y = [ next(Yg) for i in range(n) ]
+
+            p = [0, 0]
+            for i in range(n + 1):
+                p[1 - i % 2] += symmetric_poly(i, Y)*(1)**((i % 4)//2)
```
jksuom, line 658: What does `(1)**((i % 4)//2)` do?

oostben: Sorry, it was left over from the tan _eval_expand_trig method. I have removed it in the most recent commit.

oostben:
I was able to generalize the tanh(coeff*x) case using this formula:

Screen Shot 2021-04-26 at 8 52 47 AM

I also removed the pointless multiplication pointed out in the review.

sylee957:
Does this also work with n-ary addition?
(e.g. `expand_trig(tan(x+y+z))` -> `(-tan(x)*tan(y)*tan(z) + tan(x) + tan(y) + tan(z))/(-tan(x)*tan(y) - tan(x)*tan(z) - tan(y)*tan(z) + 1)`)

oostben:
Yes it does work with n-ary addition.

Screen Shot 2021-04-27 at 8 09 26 PM

Screen Shot 2021-04-27 at 8 10 38 PM

sylee957:
Half angle formula wasn't used for expanding `tan`.
I'm not sure if this can be added later, but at least I think that it's better to align the behaviors

sylee957:
This may need some tests at `test_trigonometric.py`

sympy/functions/elementary/hyperbolic.py
```diff
@@ -624,6 +625,38 @@ def as_real_imag(self, deep=True, **hints):
...
+            from sympy import symmetric_poly
+            n = len(arg.args)
+            TX = []
+            if arg.is_Add:
+                for x in arg.args:
+                    tx = tanh(x, evaluate=False)._eval_expand_trig()
+                    TX.append(tx)
+
+                Yg = numbered_symbols('Y')
+                Y = [ next(Yg) for i in range(n) ]
```
smichr, line 641:
```suggestion
                Y = symbols('Y:%s' % n)
```

sympy/functions/elementary/hyperbolic.py
```diff
@@ -624,6 +625,38 @@ def as_real_imag(self, deep=True, **hints):
...
+
+                p = [0, 0]
+                for i in range(n + 1):
+                    p[1 - i % 2] += symmetric_poly(i, Y)
+                return (p[0]/p[1]).subs(list(zip(Y, TX)))
```
smichr, line 646:
```suggestion
                return (p[0]/p[1]).xreplace(dict(zip(Y, TX)))
```

sympy/functions/elementary/hyperbolic.py
```diff
@@ -624,6 +625,38 @@ def as_real_imag(self, deep=True, **hints):
...
+        else:
+            from sympy.functions.combinatorial.numbers import nC
+            coeff, terms = arg.as_coeff_Mul()
+            if coeff.is_Integer and coeff > 1:
+                numerator = 0
+                denominator = 0
+                for k in range(0,floor((coeff-1)/2)+1):
+                    numerator += nC(range(coeff),2*k+1) * tanh(terms)**(2*k+1)
+                for k in range(0,floor(coeff/2)+1):
+                    denominator += nC(range(coeff),(2*k)) * tanh(terms)**(2*k)
```
smichr, line 656:
```suggestion
                T = tanh(terms)
                for k in range(0, floor((coeff - 1)/2) + 1):
                    numerator += nC(range(coeff), 2*k + 1)*T**(2*k + 1)
                for k in range(0, floor(coeff/2) + 1):
                    denominator += nC(range(coeff), 2*k)*T**(2*k)
```

sympy/functions/elementary/hyperbolic.py
jksuom: We could probably avoid substitutions by using `TX` instead of `Y` in `symmetric_polynomial`. There may be repetitions in `TX` but that could be handled by adding `repetition=True` when calling subsets: https://github.com/sympy/sympy/blob/8f3790a92fe47d33477ebead73f218fadb839d9d/sympy/polys/specialpolys.py#L107

smichr: This would make symmetric polynomials less minimal, e.g. `symmetric_polynomial(2, x, y)` would become `x**2 + x*y + y**2` instead of `x*y`. `symmetric_polynomial(2, x, x, y)` would be `3*x**2 + 2*x*y + y**2` but could be `x*y`, too.

jksuom: Yes, my mistake. `symmetric_poly` seems to have no problems with repeated arguments:
```
In [19]: symmetric_poly(2, x, x, y)                                             
Out[19]: 
 2        
x  + 2⋅x⋅y
```
That could be used with `TX` instead of `Y`.

smichr, line 646:
```suggestion
                p = [0, 0]  # [den, num]
                for i in range(n + 1):
                    p[i % 2] += symmetric_poly(i, TX)
                return p[1]/p[0]
```

sympy/functions/elementary/hyperbolic.py
```diff
@@ -624,6 +625,38 @@ def as_real_imag(self, deep=True, **hints):
...
+                return numerator / denominator
```
smichr, line 657:
```suggestion
                n = []
                d = []
                T = tanh(terms)
                for k in range(1, coeff + 1, 2):
                    n.append(nC(range(coeff), k)*T**k)
                for k in range(0, coeff + 1, 2):
                    d.append(nC(range(coeff), k)*T**k)
                return Add(*n)/Add(*d)
```

sympy/functions/elementary/hyperbolic.py
```diff
@@ -7,6 +7,7 @@
 from sympy.functions.elementary.exponential import exp, log, match_real_imag
 from sympy.functions.elementary.miscellaneous import sqrt
 from sympy.functions.elementary.integers import floor
+from sympy.utilities.iterables import numbered_symbols
```
smichr, line 10:
```suggestion
```

sympy/functions/elementary/hyperbolic.py
```diff
@@ -624,6 +624,36 @@ def as_real_imag(self, deep=True, **hints):
...
+            if arg.is_Add:
+                for x in arg.args:
+                    tx = tanh(x, evaluate=False)._eval_expand_trig()
+                    TX.append(tx)
+
+                p = [0, 0]  # [den, num]
+                for i in range(n + 1):
+                    p[i % 2] += symmetric_poly(i, TX)
+                return p[1]/p[0]
+        else:
```
smichr, line 643:
```suggestion
        if arg.is_Add:
            from sympy import symmetric_poly
            n = len(arg.args)
            TX = [tanh(x, evaluate=False)._eval_expand_trig()
                for x in arg.args]
            p = [0, 0]  # [den, num]
            for i in range(n + 1):
                p[i % 2] += symmetric_poly(i, TX)
            return p[1]/p[0]
        else:
```

smichr:
OP could be updated to reflect that half angle recognition is not included.

smichr:
I wonder if `expand_trig` (or user?) should just be using the conversions from hyperbolic to regular trig functions ala `fu.hyper_as_trig`.  I also see that the `tan._eval_trig` has real and imaginary handling -- is this not needed for hyperbolic `tanh`?

smichr:
The unconditional `else` will cause integer arguments to expand in terms of `tanh(1)` -- I'm not sure that is helpful. Someone can replace int args with `int*Symbol('1')` and expand to see that result if they want.
```python
tanh(5), tanh(5*Symbol('1')) -> (tanh(1)**5 + 5*tanh(1) + 10*tanh(1)**3)/(1 + 5*tanh(1)**4 + 10*tanh(1)**2)
```

jksuom:
> is this not needed for hyperbolic tanh?

The imaginary unit `I` appears in the formulas for `tan` because of Euler's formula

        cos(x) + I*sin(x) = exp(I*x).

It could be avoided by elementary plane trigonometric arguments but it simplifies the mathematics.

`I` is not needed for computations with `tanh` since it does not appear in 'hyperbolic Euler's formula'

        cosh(x) + sinh(x) = exp(x).

oostben:
Is there anything else that needs to be done for this PR to be merged?

Should I add some tests for this anywhere?

Where would I note that half angle recognition is not included @smichr?

smichr:
> Is there anything else that needs to be done for this PR to be merged?

please address this [comment](https://github.com/sympy/sympy/pull/21380#issuecomment-830712304)

oostben:
Originally we had this with the unconditional else:

``` expand_trig(tanh(5)) -> (tanh(1)**5 + 5*tanh(1) + 10*tanh(1)**3)/(1 + 5*tanh(1)**4 + 10*tanh(1)**2) ```
``` expand_trig(tanh(5*Symbol('1'))) -> (tanh(1)**5 + 10*tanh(1)**3 + 5*tanh(1))/(5*tanh(1)**4 + 10*tanh(1)**2 + 1) ``` 

After making the unconditional else into:
``` elif arg.is_Mul: ```
 
I get:
``` expand_trig(tanh(5)) -> tanh(5)``` 
``` expand_trig(tanh(5*Symbol('1'))) -> (tanh(1)**5 + 10*tanh(1)**3 + 5*tanh(1))/(5*tanh(1)**4 + 10*tanh(1)**2 + 1)``` 

Also I wanted to note that the reason, I made that else unconditional was because that is how it's done in tan's ```_eval_expand_trig ``` method:

``` expand_trig(tan(5)) -> (-10*tan(1)**3 + 5*tan(1) + tan(1)**5)/(-10*tan(1)**2 + 1 + 5*tan(1)**4) ``` 

However, I agree that it doesn't make sense to treat all expansions as multiplications if they are not additions. Might be something to change in tan's ```_eval_expand_trig ``` to keep the behavior standard. 

Is this the type of fix you were looking for?

smichr:
> Is this the type of fix you were looking for?

yes -- I am not an expert in trig expression wrangling, but the expansion doesn't seem profitable (and the user can use the symbolic number to expand a certain factor of the numeric arg if they desire, e.g. `tan(4) = tan(2*Symbol('2'))`)

I think it would be ok to make this change in the regular `tan` expansion, too.

smichr:
> ok to make this change in the regular `tan` expansion, too.

Are you willing to do that? If so, I'll wait until that change is present before starting tests.

oostben:
@smichr I added the change to all the trig functions in #21440 because they all share the same behavior.

smichr:
This is in, thanks. Will watch for #21440 to finish up, too.

Issue #21351, opened 2021-04-19
recognize `eq = sin(real) + 1` as nonnegative


smichr:
This is a special case when a determination  of being nonnegative can be made. I think it would have to be a special check for this pattern `(sin or cos + number)` where `(number - 1)` is nonnegative.  See [here](https://stackoverflow.com/a/67155019/1089161) for an example.

smichr:
Here is a possible solution to have Abs recognize this:
```diff
diff --git a/sympy/functions/elementary/complexes.py b/sympy/functions/elementary/complexes.py
index b34ca3a7de..7deddeb713 100644
--- a/sympy/functions/elementary/complexes.py
+++ b/sympy/functions/elementary/complexes.py
@@ -7,10 +7,11 @@
 from sympy.core.numbers import pi, I, oo
 from sympy.core.relational import Eq
 from sympy.functions.elementary.exponential import exp, exp_polar, log
+from sympy.functions.elementary.hyperbolic import tanh
 from sympy.functions.elementary.integers import ceiling
 from sympy.functions.elementary.miscellaneous import sqrt
 from sympy.functions.elementary.piecewise import Piecewise
-from sympy.functions.elementary.trigonometric import atan, atan2
+from sympy.functions.elementary.trigonometric import atan, atan2, cos, sin
 
 ###############################################################################
 ######################### REAL and IMAGINARY PARTS ############################
@@ -589,9 +590,18 @@ def eval(cls, arg):
             return exp(re(arg.args[0]))
         if isinstance(arg, AppliedUndef):
             return
-        if arg.is_Add and arg.has(S.Infinity, S.NegativeInfinity):
-            if any(a.is_infinite for a in arg.as_real_imag()):
-                return S.Infinity
+        if arg.is_Add:
+            if arg.has(S.Infinity, S.NegativeInfinity):
+                if any(a.is_infinite for a in arg.as_real_imag()):
+                    return S.Infinity
+            else:
+                c, t = arg.as_independent(cos, sin, tanh)
+                if t:
+                    m, t = t.as_independent(cos, sin, tanh)
+                    if isinstance(t, (cos, sin, tanh)) and t.is_real:
+                        if arg.subs(t, Dummy(nonnegative=True) - 1).is_nonnegative:
+                            return arg
+                    return
         if arg.is_zero:
             return S.Zero
         if arg.is_extended_nonnegative:
diff --git a/sympy/parsing/sympy_parser.py b/sympy/parsing/sympy_parser.py
[...]

Issue #20789, opened 2021-01-13
Wrong fact in ask_generated.py


JSS95:
`ask_generated.py` contains wrong fact.

https://github.com/sympy/sympy/blob/825a011f3c09101ce5d0c9f75c66387384810433/sympy/assumptions/ask_generated.py#L189

Here, `Q.infinite` implies `Q.extended_real`. This results following behavior in `master`:
```python
In [1]: ask(Q.extended_real(x), Q.infinite(x))
Out[1]: True
```

This is wrong since `zoo` is infinite but not extended real. It seems that we have to fix `get_known_facts()` in `ask.py`.

PR #21383, opened 2021-04-25
fix attempt for issue #20789 and #21351

ianmasc:
#### Release Notes

* assumptions 
  * added extended_real to get_known_facts_dict()

### DIFF
--- a/sympy/functions/elementary/complexes.py
+++ b/sympy/functions/elementary/complexes.py
@@ -7,10 +7,11 @@ from sympy.core.logic import fuzzy_not, fuzzy_or
 from sympy.core.numbers import pi, I, oo
 from sympy.core.relational import Eq
 from sympy.functions.elementary.exponential import exp, exp_polar, log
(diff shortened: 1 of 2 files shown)


smichr:
I think we need something more robust than this. The `t` part here could contain an arbitrary expression containing trig or hyperbolic tangent as shown in the issue page.

sympy/functions/elementary/complexes.py
```diff
@@ -589,9 +590,18 @@ def eval(cls, arg):
             return exp(re(arg.args[0]))
         if isinstance(arg, AppliedUndef):
             return
-        if arg.is_Add and arg.has(S.Infinity, S.NegativeInfinity):
-            if any(a.is_infinite for a in arg.as_real_imag()):
-                return S.Infinity
+         if arg.is_Add:
```
Shekharrajak, line 593: We can have modular sub methods to make code readable and reusable.

PR #21386, opened 2021-04-26
Rewrite ITE as Piecewise in numexpr printer

craymichael:
Fix ITE printing for NumExprPrinter by rewriting as Piecewise (which is already done in numpy printer). Failing to do so results in printing of Python Booleans. MWE-

```python
In [1]: expr = ITE(x > 0, True, False, evaluate=False)

In [2]: expr
Out[2]: 
⎧True   for x > 0
⎨                
⎩False  otherwise

In [3]: printing.lambdarepr.NumExprPrinter().doprint(expr)  # this is incorrect (uses Python Booleans)
Out[3]: "evaluate('((True) if ((x > 0)) else (False))', truediv=True)"

In [4]: lambdify([x], expr, 'numexpr')(1)  # the current state of NumExprPrinter fails
TypeError: You can't use Python's standard boolean operators in NumExpr expressions. You should use their bitwise counterparts instead: '&' instead of 'and', '|' instead of 'or', and '~' instead of 'not'.

In [5]: printing.lambdarepr.NumExprPrinter().doprint(expr)  # this is with the patch applied
Out[5]: "evaluate('where((x > 0), True, False)', truediv=True)"

In [6]: lambdify([x], expr, modules='numexpr')(1)  # the expected result
Out[6]: array(True)
```

`NumExprPrinter` now rewrites `ITE` expressions as `Piecewise`

#### Release Notes

* printing
  * `NumExprPrinter` now properly supports expressions containing `ITE`

### DIFF
--- a/sympy/printing/lambdarepr.py
+++ b/sympy/printing/lambdarepr.py
@@ -152,6 +152,10 @@ class NumExprPrinter(LambdaPrinter):
             ans.append('log(-1)')
         return ''.join(ans) + ')' * parenthesis_count
 
+    def _print_ITE(self, expr):
+        from sympy.functions.elementary.piecewise import Piecewise
+        return self._print(expr.rewrite(Piecewise))
+
     def blacklisted(self, expr):
         raise TypeError("numexpr cannot be used with %s" %
                         expr.__class__.__name__)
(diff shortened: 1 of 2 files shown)


smichr:
Thanks, this is in.

PR #21194, opened 2021-03-29
Removed unnecessary term swap while computing Mul in parse_expr.

dsaw:
#### References to other Issues or PRs
Fixes Mul issue of #21119

#### Brief description of what is fixed or changed
A swapping happens in `parse_expr`(evaluate=False) when operation is `Mul` that was changing the order, is removed since it is since unnecessary now.

#### Other comments
Note the #10560 for which the old swapping #10774 fix was done might be need to reopened since some of the nested terms `('9.2*(4.1+(-4)*3)')` aren't working, even with the old fix.

#### Release Notes

* parsing
  * Fix to maintain the order of Mul terms in parse_expr when evaluate=False

PR #21390, opened 2021-04-26
printing: term reversal and powers with negative exponents

### LINKED ISSUE #21119: `parse_expr` with `evaluate=False` still changes the order of factors
```python
sympy.parsing.sympy_parser.parse_expr("-2*x*y",evaluate=False)
x*(-2)*y
```
According to my understanding of the latest documentation at https://docs.sympy.org/latest/modules/parsing.html#sympy.parsing.sympy_parser.implicit_multiplication_application, the output should be `-2*x*y` or perhaps `(-2)*x*y`, as it says 

> ...the order of the arguments will not be made canonical. This feature allows one to tell exactly how the    expression was entered.

### LINKED ISSUE #21406: `(-x)*y` -> `-x*y`?
When the precedence is computed for Mul factors, those with a negative rational coefficient are printed with surrounding parentheses. This is not needed but is friendly in that it keeps `*-` from appearing in the expression. But this is not needed for the first factor. 

So `x*(-y)` is fine but `(-x)*y` can be `-x*y`.  This would he handled by not parenthesizing the first item in a product. This would have to be caught in `print_Mul` before assembling the factors for the string since precedence would have no way to know if it was dealing with the 1st factor or not...unless it would get an arg to allow for that.

smichr:

close #21194 as extension with retention of original commits
fixes #21119 
fixes #21406

- term swap for unevaluated expression has been undone
- powers with a leading negative sign are now collected in the denominator when they appear as a factor

```python
>>> x**-y
x**(-y)
>>> 2*_
2/x**y
```

This doesn't always lead to a shorter expression but rather than optimize that, a consistent result was selected. This will help with requests as mentioned [here](https://stackoverflow.com/questions/67258281/how-to-make-a-fraction-instead-of-negative-power-for-unevaluatedexpression-divis) -- but there will always be requests for things to appear in non-standard ways, so there is no illusion that this is the best way of printing expressions.

- leading factor with a negative sign in an unevaluated Mul prints without parenthesis now
```python
>>> Mul(-x,2,evaluate=0)
(-x)*2 <--- now prints as -x*2
```
#### Release Notes

* printing
  * fixed bug that caused factors in unevaluated expressions to be swapped
  * powers with negative signs in the exponent are now printed in the denominator
  * parentheses no longer appear on first factors in unevaluated Mul that have a negative sign

### DIFF
--- a/sympy/ntheory/tests/test_factor_.py
+++ b/sympy/ntheory/tests/test_factor_.py
@@ -522,7 +522,7 @@ def test_factorrat():
     assert str(factorrat(Rational(1, 1), visual=True)) == '1'
     assert str(factorrat(S(25)/14, visual=True)) == '5**2/(2*7)'
     assert str(factorrat(Rational(25, 14), visual=True)) == '5**2/(2*7)'
-    assert str(factorrat(S(-25)/14/9, visual=True)) == '(-1)*5**2/(2*3**2*7)'
+    assert str(factorrat(S(-25)/14/9, visual=True)) == '-1*5**2/(2*3**2*7)'
 
     assert factorrat(S(12)/1, multiple=True) == [2, 2, 3]
     assert factorrat(Rational(1, 1), multiple=True) == []
--- a/sympy/ntheory/tests/test_factor_.py
+++ b/sympy/ntheory/tests/test_factor_.py
@@ -522,7 +522,7 @@ def test_factorrat():
     assert str(factorrat(Rational(1, 1), visual=True)) == '1'
     assert str(factorrat(S(25)/14, visual=True)) == '5**2/(2*7)'
     assert str(factorrat(Rational(25, 14), visual=True)) == '5**2/(2*7)'
-    assert str(factorrat(S(-25)/14/9, visual=True)) == '-5**2/(2*3**2*7)'
+    assert str(factorrat(S(-25)/14/9, visual=True)) == '(-1)*5**2/(2*3**2*7)'
 
     assert factorrat(S(12)/1, multiple=True) == [2, 2, 3]
     assert factorrat(Rational(1, 1), multiple=True) == []
--- a/sympy/printing/str.py
+++ b/sympy/printing/str.py
@@ -254,7 +254,24 @@ class StrPrinter(Printer):
         # etc so we display in a straight-forward form that fully preserves all
         # args and their order.
         args = expr.args
-        if args[0] is S.One or any(isinstance(arg, Number) for arg in args[1:]):
+        if args[0] is S.One or any(isinstance(arg, Number) or isinstance(1/arg, Number) for arg in args[1:]):
+            from sympy.utilities.iterables import sift
+            d, n  = sift(args, lambda x:
+                         isinstance(x,Pow) and x.exp == -1,
+                         binary=True
+                         )
+            d = [1/i for i in d]
+            nfactors = [self.parenthesize(arg, prec, strict=False) for arg in n]
+            dfactors = [self.parenthesize(arg, prec, strict=False) for arg in d]
+            numerator_string = '*'.join(nfactors)
+            denominator_string = '*'.join(dfactors)
+            if len(dfactors) > 1:
+                return '%s/(%s)' % (numerator_string, denominator_string)
+            elif dfactors:
+                return '%s/%s' % (numerator_string, denominator_string)
+            else:
+                return numerator_string
+
             factors = [self.parenthesize(a, prec, strict=False) for a in args]
             return '*'.join(factors)
 
--- a/sympy/printing/str.py
+++ b/sympy/printing/str.py
@@ -11,7 +11,7 @@ from sympy.printing.precedence import precedence, PRECEDENCE
 
 from mpmath.libmp import prec_to_dps, to_str as mlib_to_str
 
-from sympy.utilities import default_sort_key
+from sympy.utilities import default_sort_key, sift
 
 
 class StrPrinter(Printer):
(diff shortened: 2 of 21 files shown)


smichr:
Any objections or comments?

sympy/functions/special/error_functions.py
```diff
@@ -84,7 +84,7 @@ class erf(Function):
 
     >>> from sympy import diff
     >>> diff(erf(z), z)
-    2*exp(-z**2)/sqrt(pi)
+    2/(sqrt(pi)*exp(z**2))
```
oscarbenjamin, line 87: This printing seems odd to me

oscarbenjamin: Is this just for the str printer and not the latex and pretty printers? It might be confusing for them to be different.

smichr: Does it look odd because of the `exp`? How about
```python
>>> 2*3**(-z)/sqrt(pi)
2/3**z
>>> x*y**(-3)
x/y**3
```

smichr: > Is this just for the str printer

yes, until the changes are acceptable

oscarbenjamin: Yes, I would usually write an exponential function like `e^{-x}` to be in the numerator even if the argument is negated.

smichr: > usually write an exponential function like `e^{-x}` to be in the numerator

what about the other two examples involving powers?

oscarbenjamin: I think that the other two look good. Actually the latex and pretty printers already show `y**-3` in the numerator.

sympy/printing/str.py
```diff
@@ -6,12 +6,13 @@
 
 from sympy.core import S, Rational, Pow, Basic, Mul, Number
 from sympy.core.mul import _keep_coeff
+from sympy.functions.elementary.exponential import exp
```
smichr, line 9:
```suggestion
```

sympy/stats/matrix_distributions.py
```diff
@@ -369,11 +369,11 @@ def Wishart(symbol, n, scale_matrix):
...
     >>> X = MatrixSymbol('X', 2, 2)
     >>> density(W)(X).doit()
-    2**(-n)*3**(-n/2)*exp(Trace(Matrix([
+    exp(Trace(Matrix([
     [-1/3,  1/6],
-    [ 1/6, -1/3]])*X))*Determinant(X)**(n/2 - 3/2)/(sqrt(pi)*gamma(n/2)*gamma(n/2 - 1/2))
+    [ 1/6, -1/3]])*X))*Determinant(X)**(n/2 - 3/2)/(2**n*3**(n/2)*sqrt(pi)*gamma(n/2)*gamma(n/2 - 1/2))
     >>> density(W)([[1, 0], [0, 1]]).doit()
-    2**(-n)*3**(-n/2)*exp(-2/3)/(sqrt(pi)*gamma(n/2)*gamma(n/2 - 1/2))
+    1/(2**n*3**(n/2)*sqrt(pi)*exp(2/3)*gamma(n/2)*gamma(n/2 - 1/2))
```
smichr, line 376:
```suggestion
    exp(-2/3)/(2**n*3**(n/2)*sqrt(pi)*gamma(n/2)*gamma(n/2 - 1/2))
```

sympy/stats/matrix_distributions.py
```diff
@@ -285,11 +285,11 @@ def MatrixGamma(symbol, alpha, beta, scale_matrix):
...
     >>> X = MatrixSymbol('X', 2, 2)
     >>> density(M)(X).doit()
-    3**(-a)*b**(-2*a)*exp(Trace(Matrix([
+    exp(Trace(Matrix([
     [-2/3,  1/3],
-    [ 1/3, -2/3]])*X)/b)*Determinant(X)**(a - 3/2)/(sqrt(pi)*gamma(a)*gamma(a - 1/2))
+    [ 1/3, -2/3]])*X)/b)*Determinant(X)**(a - 3/2)/(3**a*sqrt(pi)*b**(2*a)*gamma(a)*gamma(a - 1/2))
     >>> density(M)([[1, 0], [0, 1]]).doit()
-    3**(-a)*b**(-2*a)*exp(-4/(3*b))/(sqrt(pi)*gamma(a)*gamma(a - 1/2))
+    1/(3**a*sqrt(pi)*b**(2*a)*exp(4/(3*b))*gamma(a)*gamma(a - 1/2))
```
smichr, line 292:
```suggestion
    exp(-4/(3*b))/(3**a*sqrt(pi)*b**(2*a)*gamma(a)*gamma(a - 1/2))
```

sympy/ntheory/tests/test_factor_.py
```diff
@@ -522,7 +522,7 @@ def test_factorrat():
     assert str(factorrat(Rational(1, 1), visual=True)) == '1'
     assert str(factorrat(S(25)/14, visual=True)) == '5**2/(2*7)'
     assert str(factorrat(Rational(25, 14), visual=True)) == '5**2/(2*7)'
-    assert str(factorrat(S(-25)/14/9, visual=True)) == '-5**2/(2*3**2*7)'
+    assert str(factorrat(S(-25)/14/9, visual=True)) == '(-1)*5**2/(2*3**2*7)'
```
smichr, line 525: We should get rid of these parentheses around the `-1` and check:
```python
>>> S('-3*4',evaluate=False)
(-3)*4
>>> Mul(-3,4,evaluate=False).args
(-3, 4)
>>> S('-3*4',evaluate=False).args
(4, -3)
>>> S('3*4',evaluate=False).args
(3, 4)
```

smichr: This is controlled be precedence of a negated factor in a product while gets precedence of Add and thus is parenthesized. This is a separate issue.

smichr:
@oscarbenjamin -- having done this, I am +1/2. Should string printing just be as fast as possible and not get into pretty semantics? @asmeurer ? If you are both in favor I won't object and this can be committed or I can clean up last commits into a single one.

oscarbenjamin:
It looks good to me.

PR #21397, opened 2021-04-28
to_axis_angle remove multiplying by -1

SlowerPhoton:
to_axis_angle: there is no good reason to multiply the quaternion by -1 before performing the acos. Rotation can go either way and the angle can be negative.


oscarbenjamin:
This will need a test. You should add a reference to the associated issue #21398

PR #21400, opened 2021-04-29
Rational exponent of I < 2

smichr:

Before:
```python
>>> [sqrt(I)**i for i in range(8)]
[1, sqrt(I), I, I**(3/2), -1, I**(5/2), -I, I**(7/2)]
>>> I**(S(11)/3)
I**(11/3)
```

After
```python
>>> [sqrt(I)**i for i in range(8)]
[1, sqrt(I), I, I**(3/2), -1, -sqrt(I), -I, -I**(3/2)]
>>> I**(S(11)/3)
-I**(5/3)
```
#### Release Notes

* core
  * Rational exponents of I are now less than 2

### DIFF
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -3972,17 +3972,22 @@ class ImaginaryUnit(AtomicExpr, metaclass=Singleton):
         I**3 mod 4 -> -I
         """
 
-        if isinstance(expt, Number):
-            if isinstance(expt, Integer):
-                expt = expt.p % 4
-                if expt == 0:
-                    return S.One
-                if expt == 1:
-                    return S.ImaginaryUnit
-                if expt == 2:
-                    return -S.One
+        if isinstance(expt, Integer):
+            expt = expt % 4
+            if expt == 0:
+                return S.One
+            elif expt == 1:
+                return S.ImaginaryUnit
+            elif expt == 2:
+                return S.NegativeOne
+            elif expt == 3:
                 return -S.ImaginaryUnit
-        return
+        if isinstance(expt, Rational):
+            i, r = divmod(expt, 2)
+            rv = Pow(S.ImaginaryUnit, r, evaluate=False)
+            if i % 2:
+                return Mul(S.NegativeOne, rv, evaluate=False)
+            return rv
 
     def as_base_exp(self):
         return S.NegativeOne, S.Half
--- a/sympy/core/tests/test_power.py
+++ b/sympy/core/tests/test_power.py
@@ -318,8 +318,7 @@ def test_issue_6100_12942_4473():
 
 
 def test_issue_6208():
-    from sympy import root, Rational
-    I = S.ImaginaryUnit
+    from sympy import root
     assert sqrt(33**(I*Rational(9, 10))) == -33**(I*Rational(9, 20))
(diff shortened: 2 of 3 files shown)


sympy/core/numbers.py
```diff
@@ -3972,17 +3972,22 @@ def _eval_power(self, expt):
...
+                return S.ImaginaryUnit
+            elif expt == 2:
+                return S.NegativeOne
+            elif expt == 3:
                 return -S.ImaginaryUnit
-        return
+        if isinstance(expt, Rational):
+            i, r = divmod(expt, 2)
+            rv = Pow(S.ImaginaryUnit, r, evaluate=False)
+            if i:
```
jksuom, line 3988: This should probably be `if i % 2`:
```
>>> I**Rational(5, 2)
-sqrt(I)
>>> I**Rational(9, 2)
-sqrt(I)
```
`sqrt(I)**n` will not be evaluated by this code.

smichr, line 3988:
```suggestion
            if i % 2:
```

smichr: Thanks -- I kept switching between Mod and divmod thinking.

Hmm... so `sqrt(I)**5 -> I**(5/2)` because it is a Pow operation and sqrt(I) is the base so it won't pass through this code. Then it should also be handled in the Pow instantiation by looking for a base that is a rational power of `I` and an exponent that is Rational.

sympy/core/tests/test_power.py
```diff
@@ -587,3 +586,8 @@ def _(a, b):
     assert power(a, b) == NewPow(a, b)
     assert power(b, a) == NewPow(b, a)
     assert power(b, b) == NewPow(b, b)
+
+
+def test_powers_of_I():
+    assert [sqrt(I)**i for i in range(9)] == [
+        1, sqrt(I), I, sqrt(I)**3, -1, -sqrt(I), -I, -sqrt(I)**3, 1]
```
smichr, line 593:
```suggestion
        1, sqrt(I), I, sqrt(I)**3, -1, -sqrt(I), -I, -sqrt(I)**3, 1]
    assert sqrt(I)**5 == -sqrt(I)
    assert sqrt(I)**(9/S(2)) == -I**(S(1)/4)
```

sympy/core/numbers.py
smichr: This change has been made. Any other changes? Else I think this is ready and is a pretty low-impact change.

sympy/core/tests/test_power.py
```diff
@@ -587,3 +586,10 @@ def _(a, b):
...
```
smichr, line 593:
```suggestion
    assert [sqrt(I)**i for i in range(13)] == [
        1, sqrt(I), I, sqrt(I)**3, -1, -sqrt(I), -I, -sqrt(I)**3, 1, sqrt(I), I, sqrt(I)**3, -1]
```

sympy/core/numbers.py
smichr: I will confirm, but I don't think special handling in Pow needs to happen since the `_eval_power` should handle it -- it is going to be a lot slower there, however, I think and maybe is worth the quick handling at the outset.

smichr: Unless you think better to do so, I'll let the current Pow code handle `sqrt(I)**n` where `n` is a Rational.

jksuom:
Looks good.

Issue #15086, opened 2018-08-13
Regression: Degradation of det function performance when applied to large matrices of polynomials.


petersbingham:
I was directed to post my issue from [stackoverflow](https://stackoverflow.com/questions/50398584/unhandled-exception-cant-unify-dmp-when-calculating-det-of-sympy-matrix) (although the exception seems to have changed slightly with my repo here).

This report is for different versions of sympy. I have ran the tests on windows 10. Other versions, which remain constant throughout are:
python: 2.7.14, 64 bit
mpmath: 1.0

I have attached a zip file containing five test files. The .py contains the test code. The .txt contain coefficients for the test polynomial. Run with all five files in the same path. There is code at the top of the .py file for loading and constructing the polynomial. The main test code is at the bottom. I can try and simplify on request but the issues only occur with certain polynomials.

Basically when running with sympy 1.0 I encounter no issues running the test code. However, in the .py test file note **line 92, where I wrap the expression in a sym.poly call**. I do this because it provides significant performance improvement. Without this it becomes unfeasible to run my calculations (I run for increasingly larger polynomials).

When I upgrade to sympy 1.1.1 (and also 1.2) I get the following exception from the attached test code (be sure to scroll right to the end):

```
PS C:\Users\Peter\Desktop\sympy_issue> python .\poly_issue_demo.py
Traceback (most recent call last):
  File ".\poly_issue_demo.py", line 96, in <module>
    symPoly = sym.polys.Poly(det, k)
  File "C:\Python27\lib\site-packages\sympy\polys\polytools.py", line 129, in __new__
    return cls._from_expr(rep, opt)
  File "C:\Python27\lib\site-packages\sympy\polys\polytools.py", line 239, in _from_expr
    rep, opt = _dict_from_expr(rep, opt)
  File "C:\Python27\lib\site-packages\sympy\polys\polyutils.py", line 365, in _dict_from_expr
    rep, gens = _dict_from_expr_if_gens(expr, opt)
  File "C:\Python27\lib\site-packages\sympy\polys\polyutils.py", line 304, in _dict_from_expr_if_gens
    (poly,), gens = _parallel_dict_from_expr_if_gens((expr,), opt)
  File "C:\Python27\lib\site-packages\sympy\polys\polyutils.py", line 214, in _parallel_dict_from_expr_if_gens
    raise PolynomialError("%s contains an element of the generators set" % factor)
[...]

PR #21402, opened 2021-04-29
refactor(polys): move smith_normal_form to DomainMatrix

oscarbenjamin:
This commit moves the Smith Normal Code routine from Matrix to
DomainMatrix and replaces the original function with one that can
convert to DomainMatrix for the calculation.

The SNF code was the last place in the codebase that used RawMatrix so
the RawMatrix class is also removed. The RawMatrix class was a bad idea
because it subclasses Matrix in order to insert non-Expr elements
breaking a key internal invariant of the Matrix class. The removal of
RawMatrix is a step towards being able to enforce that the elements of a
Matrix should always be Expr.

This also adds tests to improve coverage in sympy.polys.matrices.

#### Release Notes

* matrices
    * The `smith_normal_form` function now expects in ordinary `Matrix` as input rather than a `RawMatrix`
* polys
    * The `RawMatrix` class has been removed. Use `Matrix`, `DomainMatrix` or `PolyMatrix` instead depending on the application.

### DIFF
--- a/sympy/matrices/normalforms.py
+++ b/sympy/matrices/normalforms.py
@@ -1,9 +1,24 @@
 '''Functions returning normal forms of matrices'''
 
-from sympy.matrices.dense import diag, zeros
+from sympy.polys.polytools import Poly
+from sympy.polys.matrices import DomainMatrix
+from sympy.polys.matrices.normalforms import (
+        smith_normal_form as _snf,
+        invariant_factors as _invf,
+    )
 
 
-def smith_normal_form(m, domain = None):
+def _to_domain(m, domain=None):
+    """Convert Matrix to DomainMatrix"""
+    m = m.applyfunc(lambda e: e.as_expr() if isinstance(e, Poly) else e)
+    dM = DomainMatrix.from_Matrix(m)
+    domain = domain or getattr(m, 'domain', None)
+    if domain is not None:
+        dM = dM.convert_to(domain)
+    return dM
+
+
+def smith_normal_form(m, domain=None):
     '''
     Return the Smith Normal Form of a matrix `m` over the ring `domain`.
     This will only work if the ring is a principal ideal domain.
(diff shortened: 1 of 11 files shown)


sympy/matrices/normalforms.py
```diff
@@ -1,35 +1,43 @@
...
     '''
-    invs = invariant_factors(m, domain=domain)
-    smf = diag(*invs)
-    n = len(invs)
-    if m.rows > n:
-        smf = smf.row_insert(m.rows, zeros(m.rows-n, m.cols))
-    elif m.cols > n:
-        smf = smf.col_insert(m.cols, zeros(m.rows, m.cols-n))
-    return smf
+    dM = _to_domain(m, domain)
```
sylee957, line 36: This can be made to warn deprecation by looking up for `.ring` attribute first, and raising the deprecation warning if there is the attribute.

oscarbenjamin: I'm not sure how that helps with anything. This PR removes the `RawMatrix` class so anyone following the documented way to use the `shmith_normal_form` function will find that their code breaks:
https://github.com/sympy/sympy/blob/d43e9dc5aed4d2b695a3176cdf8ad4e3c79f0e5b/sympy/matrices/normalforms.py#L14-L20
It isn't possible to make `RawMatrix` work and it does need to be completely removed from the codebase in order to establish the internal invariant of `Matrix` that its elements are `Expr`.

The `RawMatrix` class which subclasses `Matrix` and uses unsympifiable domain elements as the elements of the matrix is completely broken. It is a previous attempt to make something like `DomainMatrix` but is based on a flawed design. It isn't really possible to make something that will emulate `RawMatrix` based on `DomainMatrix` because a `RawMatrix` is constructed directly with domain elements or Poly but no domain is supplied and the domain is instead attached as the `ring` attribute.

Anyone who is currently using the `smith_normal_form` function needs to change their code in order to either use `Matrix` with `Expr` or `DomainMatrix` with domain elements and a specified domain. This PR makes both of those possible but I don't see any good way to make that work in a backwards compatible way.

sylee957: Maybe RawMatrix itself should be kept as a deprecating class for now.

Is there a reason that RawMatrix should be removed immediately for now? Maybe you want to change the Matrix internals.

oscarbenjamin: If we keep `RawMatrix` in any way then it will continue to constrain what can be done to improve `Matrix`. There are other classes that need to be changed or removed as well though like `NewMatrix` and `PolyMatrix` so until that is done we do not need to remove `RawMatrix`.

We need to establish the invariant that the matrix elements are Expr which will improve the matrices code. It also makes it possible to use DomainMatrix internally in Matrix.

oscarbenjamin: I've added back a hacky version of RawMatrix that converts all the elements to Expr but manages to pass the previous tests for the Smith normal form. It's hard to be sure that it is really compatible with current RawMatrix just because RawMatrix is so poorly defined.

jksuom: I don't think RawMatrix is needed anywhere. There was an issue with matrices having Poly elements (https://github.com/sympy/sympy/issues/15086) but that was resolved by means of PolyMatrix. That should also have been used in normalforms instead of RawMatrix but it was only being developed at the time.

oscarbenjamin: > I don't think RawMatrix is needed anywhere.

Agreed. This PR removes the only place in the sympy codebase that still uses RawMatrix. I guess it is reasonable to leave a DeprecationWarning though at least until the other issues are resolved given that it is currently the documented way to compute the Smith normal form.

The other issues are:

1. NewMatrix needs to be removed from holonomic which should just use DomainMatrix instead.
2. PolyMatrix also subclasses Matrix and changes the internal elements. It should be reimplemented on top of DomainMatrix instead.

Once those issues are also solved then removing RawMatrix would lead to the actual beneficial result that Matrix can constrain its entries to Expr and then DomainMatrix can be used internally by Matrix.

oscarbenjamin:
Thanks @sylee957 and @jksuom

PR #21404, opened 2021-04-30
Applied the diff and modified the failing tests

### LINKED ISSUE #21166: Wrong integration result involving square root of absolute value
The function

```
sin(x / sqrt(abs(x))
```

is odd, so its definite integral over the range `(-1, 1)` should be identically zero (see [WolframAlpha result here](https://www.wolframalpha.com/input/?i=integrate+sin%28x%2Fsqrt%28abs%28x%29%29%29+from+-1+to+1)).

image

Sympy yields a different, nonzero value:

```python
>>> Integral(sin(x / sqrt(abs(x))), (x, -1, 1)).doit()  
-4*cos(1) + 4*sin(1)
```

```python
>>> Integral(sin(x / sqrt(abs(x))), (x, -1, 1)).doit().evalf()
1.20467471575903
```

The issue is related to the divergence of the numerator as `x -> 0`, and possibly the fact that the absolute value is not differentiable at that point. Note that sympy has no issues when the `sqrt(abs(x))` is in the numerator.

I'm using `sympy` 1.7.1.

jainachal03:
Fixes #21166

This is the implementation of the comments on the main issue page. It also modifies the failing tests.

#### Release Notes

* integrals
  * Fixed a bug in the heurisch module

### DIFF
--- a/sympy/integrals/heurisch.py
+++ b/sympy/integrals/heurisch.py
@@ -743,7 +743,7 @@ def heurisch(f, x, rewrite=False, hints=None, mappings=None, retries=3,
 
     if solution is not None:
         antideriv = solution.subs(rev_mapping)
-        antideriv = cancel(antideriv).expand(force=True)
+        antideriv = cancel(antideriv).expand()
 
         if antideriv.is_Add:
             antideriv = antideriv.as_independent(x)[1]
--- a/sympy/integrals/tests/test_heurisch.py
+++ b/sympy/integrals/tests/test_heurisch.py
@@ -153,9 +153,8 @@ def test_heurisch_symbolic_coeffs():
 def test_heurisch_symbolic_coeffs_1130():
     y = Symbol('y')
     assert heurisch_wrapper(1/(x**2 + y), x) == Piecewise(
-        (-I*log(x - I*sqrt(y))/(2*sqrt(y))
-         + I*log(x + I*sqrt(y))/(2*sqrt(y)), Ne(y, 0)),
-        (-1/x, True))
+    (log(x - sqrt(-y))/(2*sqrt(-y)) - log(x + sqrt(-y))/(2*sqrt(-y)),
+    Ne(y, 0)), (-1/x, True))
     y = Symbol('y', positive=True)
     assert heurisch_wrapper(1/(x**2 + y), x) == (atan(x/sqrt(y))/sqrt(y))
 
@@ -205,8 +204,8 @@ def test_heurisch_wrapper():
         (-log(x - y)/(2*y) + log(x + y)/(2*y), Ne(y, 0)), (1/x, True))
     # issue 6926
     f = sqrt(x**2/((y - x)*(y + x)))
-    assert heurisch_wrapper(f, x) == x*sqrt(x**2)*sqrt(1/(-x**2 + y**2)) \
-        - y**2*sqrt(x**2)*sqrt(1/(-x**2 + y**2))/x
+    assert heurisch_wrapper(f, x) == x*sqrt(x**2/(-x**2 + y**2)) \
+    - y**2*sqrt(x**2/(-x**2 + y**2))/x
 
 def test_issue_3609():
     assert heurisch(1/(x * (1 + log(x)**2)), x) == atan(log(x))
--- a/sympy/integrals/tests/test_heurisch.py
+++ b/sympy/integrals/tests/test_heurisch.py
@@ -123,7 +123,7 @@ def test_heurisch_hyperbolic():
 
 def test_heurisch_mixed():
     assert heurisch(sin(x)*exp(x), x) == exp(x)*sin(x)/2 - exp(x)*cos(x)/2
-    assert heurisch(sin(x/sqrt(-x))) == 2*x*cos(x/sqrt(-x))/sqrt(-x) - 2*sin(x/sqrt(-x))
+    assert heurisch(sin(x/sqrt(-x)), x) == 2*x*cos(x/sqrt(-x))/sqrt(-x) - 2*sin(x/sqrt(-x))
 
 
 def test_heurisch_radicals():
--- a/sympy/integrals/tests/test_heurisch.py
+++ b/sympy/integrals/tests/test_heurisch.py
@@ -29,6 +29,9 @@ def test_components():
 def test_issue_10680():
     assert isinstance(integrate(x**log(x**log(x**log(x))),x), Integral)
 
+def test_issue_21166():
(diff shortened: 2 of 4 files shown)


jainachal03:
@oscarbenjamin please review this pr and approve if this  looks fine. Thanks

oscarbenjamin:
There should be tests for the example from the issue.

jainachal03:
@oscarbenjamin I could only come up with a single test, by using the integral in the question.
```diff
--git a/sympy/integrals/tests/test_heurisch.py b/sympy/integrals/tests/test_heurisch.py
index f872d30228..2a930e4cb1 100644
--- a/sympy/integrals/tests/test_heurisch.py
+++ b/sympy/integrals/tests/test_heurisch.py
@@ -29,6 +29,9 @@ def test_components():
 def test_issue_10680():
     assert isinstance(integrate(x**log(x**log(x**log(x))),x), Integral)
 
+def test_issue_21166():
+    assert integrate(sin(x/sqrt(abs(x))), (x, -1, 1)) == 0
+
 def test_heurisch_polynomials():
     assert heurisch(1, x) == x
     assert heurisch(x, x) == x**2/2
 
```
I am not very sure of how what tests can be added as I believe I do not completely understand heurisch.

oscarbenjamin:
That test looks good. The antiderivative computed by heurisch should also be tested since that is the source of the problem.

jainachal03:
I think that sympy is not able to calculate an antiderivative for the above integral.
```
f =  sin(x/abs(sqrt(x)))
val = integrate(f, x)
print(val) # outputs  Integral(sin(x/Abs(sqrt(x))), x)
```

oscarbenjamin:
As I mentioned in the issue:
> Internally integrate splits this into two integrals over the positive and negative parts to eliminate the abs so the  wrong result comes from this antiderivative from heurisch:
> ```
> In [17]: integrate(sin(x/sqrt(-x)), x, heurisch=True)
> Out[17]: -2⋅ⅈ⋅√x⋅cosh(√x) + 2⋅ⅈ⋅sinh(√x)
> ```
That should be tested directly with the `heurisch` function rather than `integrate` though.

jainachal03:
@oscarbenjamin  thanks!
I have added test in the **test_heurisch** file.
```diff --git a/sympy/integrals/tests/test_heurisch.py b/sympy/integrals/tests/test_heurisch.py
index f872d30228..3a099c5dbf 100644
--- a/sympy/integrals/tests/test_heurisch.py
+++ b/sympy/integrals/tests/test_heurisch.py
@@ -29,6 +29,9 @@ def test_components():
 def test_issue_10680():
     assert isinstance(integrate(x**log(x**log(x**log(x))),x), Integral)
 
+def test_issue_21166():
+    assert integrate(sin(x/sqrt(abs(x))), (x, -1, 1)) == 0
+
 def test_heurisch_polynomials():
     assert heurisch(1, x) == x
     assert heurisch(x, x) == x**2/2
@@ -120,6 +123,7 @@ def test_heurisch_hyperbolic():
 
 def test_heurisch_mixed():
     assert heurisch(sin(x)*exp(x), x) == exp(x)*sin(x)/2 - exp(x)*cos(x)/2
+    assert heurisch(sin(x/sqrt(-x)) == 2*x*cos(x/sqrt(-x))/sqrt(-x) - 2*sin(x/sqrt(-x))
 
 
 def test_heurisch_radicals():
```

oscarbenjamin:
> I have added test in the **test_heurisch** file.

That looks good. Can you push it to the PR?

oscarbenjamin:
```
__________ sympy/integrals/tests/test_heurisch.py:test_heurisch_mixed __________
  File "/home/runner/work/sympy/sympy/sympy/integrals/tests/test_heurisch.py", line 126, in test_heurisch_mixed
    assert heurisch(sin(x/sqrt(-x))) == 2*x*cos(x/sqrt(-x))/sqrt(-x) - 2*sin(x/sqrt(-x))
TypeError: heurisch() missing 1 required positional argument: 'x'
```
You need to pass the symbol argument to heurisch.

jainachal03:
Currently, the pr is failing at this test. 
``` 
The ODE is:
-Derivative(f(x), x)**3 + Derivative(f(x), (x, 2))
The expected solution was
[Eq(f(x), C1 - sqrt(2)*(I*C2 + I*x)*sqrt(1/(C2 + x))), Eq(f(x), C1 + sqrt(2)*(I*C2 + I*x)*sqrt(1/(C2 + x)))]

What dsolve returned is:
[Eq(f(x), C1 - sqrt(2)*sqrt(-1/(C2 + x))*(C2 + x)), Eq(f(x), C1 + sqrt(2)*sqrt(-1/(C2 + x))*(C2 + x))]
```

I tried to change the output, but I didn't really quite understood the workings of that file.

oscarbenjamin:
You've missed the key part which is:
```
AssertionError: Different solution found from dsolve for example reducible_11.
```
Maybe the name of the example should be made more prominent in the error message @Mohitbalwani26 

The example is here:
https://github.com/sympy/sympy/blob/05f5021fff2e74f674e62c56ed0fc58bca17e2d2/sympy/solvers/ode/tests/test_single.py#L1260-L1265
The solution given should be changed.

Mohitbalwani26:
> Maybe the name of the example should be made more prominent in the error message @Mohitbalwani26
> 
Any suggestions what should these examples renamed to?
Also, I have a doubt why didn't this example fail earlier in tests? 🤔

oscarbenjamin:
> > Maybe the name of the example should be made more prominent in the error message @Mohitbalwani26
> 
> Any suggestions what should these examples renamed to?

I just mean that the error message only shows the name of the example as part of the traceback and not as part of the nicely formatted test failure message. The name of the example `reducible_11` is the key piece of information that you would need to find it in the test code so it could be shown more prominently.

> Also, I have a doubt why didn't this example fail earlier in tests? 🤔

Why would it fail? Is the solution returned incorrect or just in a different form?

Mohitbalwani26:
> I just mean that the error message only shows the name of the example as part of the traceback and not as part of the nicely formatted test failure message. The name of the example `reducible_11` is the key piece of information that you would need to find it in the test code so it could be shown more prominently.
> 
Ok, surely will look into it and make a PR for the same


> Why would it fail? Is the solution returned incorrect or just in a different form?
Here as this example failed so earlier in my PR's it should have failed too, right?

oscarbenjamin:
> Here as this example failed so earlier in my PR's it should have failed too, right?

It failed here because the result from `integrate` is changed in this PR. A different result returned from `integrate` has changed the form of the answer returned by `dsolve`.

smichr:
Thanks, this is in.

Issue #15130, opened 2018-08-23
Provide guidance to 3rd part library devs on SymPyDeprecationWarning


bjodah:
We should try to find (and document) an easy way for 3rd parties to **selectively** turn `SymPyDeprecationWarning`s into errors. This will be helpful for people trying to hunt down where in a large codebase they are using deprecated sympy features.

Python 3.7 should offer [quite fine granularity](https://docs.python.org/3.7/library/warnings.html#describing-warning-filters) in `PYTHONWARNINGS` (or the `-W` flag).

Here is an example (test.py):
```python
import imp  # we don't want to turn this import into an error
from sympy import Matrix
Matrix([[1, 0], [0, 1]]).dot([2, 3])
```
running the above file using python3.7 gives the following output:
```
test.py:1: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
/home/bjorn/vc/sympy/sympy/matrices/matrices.py:2494: SymPyDeprecationWarning: 

Dot product of non row/column vectors has been deprecated since SymPy
1.2. Use * to take matrix products instead. See
https://github.com/sympy/sympy/issues/13815 for more info.

  useinstead="* to take matrix products").warn()
```
if we use:
```bash
$ PYTHONWARNINGS='error::DeprecationWarning' python3.7 test.py
```
we get
```
Traceback (most recent call last):
...
DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
```
which is not helpful.

Here are a few things I've tried (none of which have been successful)

- `PYTHONWARNINGS='error:::sympy[.*]' python3.7 test.py`
- `python3.7 -W"error::sympy.utilities.exceptions.SymPyDeprecationWarning" test.py`

asmeurer:
A good blog post about this from @nedbat. https://nedbatchelder.com/blog/201810/why_warnings_is_mysterious.html. 

It looks like

- The module isn't actually a regular expression
- You can't actually use third-party warnings. 

However, if you use pytest there is a way directly in pytest to configure the warnings which apparently works.

bjodah:
Thanks, coincidentally I have just started using pytest's warning-filtering-feature in another project.
We should probably make this pass:
```
python3 -c "from sympy import SymPyDeprecationWarning"
```
so that people can run:
```
pytest -W error::sympy.SymPyDeprecationWarning
```
with their codes.

[...]

PR #21417, opened 2021-05-02
Validate assumptions passed to Symbol

### LINKED ISSUE #21416: Possible assumptions poorly documented and missing validation of assumptions
The `Symbol` class allows one to specify assumptions. Assumptions are frequently needed in engineering and physics to express that certain variables are real and positive (e.g. mass of a particle). Which assumptions are understood by Sympy is not well documented and mistyping the assumptions does not cause an error, which is bad.

#### Documentation issues

1) Docstring of function `symbols`.

The docstring mentions assumptions and shows `integer` and `real` in examples, but does not list all possible assumptions nor does it point to the place where such a list can be found (the latter would be better, the assumptions should be documented in the Symbol class, I suppose).

2) Docstring of class `Symbol`

The docstring mentions assumptions in the class docstring but does not list all possible assumptions. The docstring of `assumptions0` (why the `0` at the end?) also does not list all possible assumptions, they only appear **by chance** in the example code.

3) Searching for "assumption" in Sympy docs yields no useful results

When you search in the Sympy online docs for the string "assumption" you find these hits (in that order):

- https://docs.sympy.org/latest/modules/assumptions/index.html?highlight=assumption#module-sympy.assumptions
  The first hit is very confusing and not what one was searching for.
- https://docs.sympy.org/latest/modules/core.html?highlight=assumption#module-sympy.core.assumptions
  The second hit is finally listing all possible assumptions. This is what the user wants to see.
- https://docs.sympy.org/latest/modules/assumptions/index.html?highlight=assumption
  This is again a variation of the first hit and therefore not useful.

#### Validation issues

The class `Symbol` does not validate the assumptions. One can pass arbitrary arguments. The user is not informed about typos, this is very bad.

```py
>>> from sympy import Symbol

>>> Symbol("x", real=True).assumptions0 # OK, real is a valid assumption
{'real': True, 'complex': True, 'imaginary': False, 'hermitian': True, 'infinite': False, 'finite': True, 'commutative': True, 'extended_real': True}
>>> Symbol("x", is_real=True).assumptions0 # BAD, is_real is not a valid assumption, this should raise an error
{'is_real': True, 'commutative': True}

### PR DESCRIPTION
**Pull request by HDembinski**

This partially fixes #21416 (only the validation part, not the documentation issue).

`Symbol` now validates the assumptions passed to it and produces a nice error message when invalid assumptions are passed to the constructor. `ValueError` is raised in that case.

```sh
python -c 'from sympy import symbols; symbols("x", is_real=True)'
  File "<string>", line 1, in <module>
  File "/Users/hdembinski/Extern/sympy/sympy/core/symbol.py", line 724, in symbols
    symbol = cls(literal(name), **args)
  File "/Users/hdembinski/Extern/sympy/sympy/core/symbol.py", line 282, in __new__
    cls._validate_assumptions(assumptions)
  File "/Users/hdembinski/Extern/sympy/sympy/core/symbol.py", line 236, in _validate_assumptions
    raise ValueError(
ValueError: invalid assumptions: is_real (allowed: real, extended_nonpositive, finite, imaginary, zero, extended_nonnegative, noninteger, extended_nonzero, nonnegative, rational, polar, infinite, extended_negative, commutative, negative, antihermitian, positive, prime, integer, algebraic, extended_real, transcendental, irrational, extended_positive, composite, hermitian, nonzero, complex, nonpositive, even, odd)
```

#### Release Notes

* core
  * Symbol now validates the assumptions and raises a ValueError on invalid assumptions.

### DIFF
--- a/sympy/core/symbol.py
+++ b/sympy/core/symbol.py
@@ -227,8 +227,19 @@ class Symbol(AtomicExpr, Boolean):
         return True
 
     @staticmethod
+    def _validate_assumptions(assumptions):
+        """Validate assumption strings.
+        """
+        valid_assumptions = set(_assume_defined)
+        diff = set(assumptions) - valid_assumptions
+        if diff:
+            raise ValueError(
+                "invalid assumptions: %s " % ", ".join(diff) +
+                "(allowed: %s)" % ", ".join(valid_assumptions))
+
+    @staticmethod
     def _sanitize(assumptions, obj=None):
-        """Remove None, covert values to bool, check commutativity *in place*.
+        """Validate names, remove None, convert to bool, check commutativity *in place*.
         """
 
         # be strict about commutativity: cannot be None
@@ -268,6 +279,7 @@ class Symbol(AtomicExpr, Boolean):
         False
 
         """
+        cls._validate_assumptions(assumptions)
         cls._sanitize(assumptions, cls)
         return Symbol.__xnew_cached_(cls, name, **assumptions)
 
@@ -404,6 +416,7 @@ class Dummy(Symbol):
             dummy_index = Dummy._base_dummy_index + Dummy._count
             Dummy._count += 1
 
+        cls._validate_assumptions(assumptions)
         cls._sanitize(assumptions, cls)
         obj = Symbol.__xnew__(cls, name, **assumptions)
 
@@ -519,6 +532,7 @@ class Wild(Symbol):
     def __new__(cls, name, exclude=(), properties=(), **assumptions):
         exclude = tuple([sympify(x) for x in exclude])
         properties = tuple(properties)
+        cls._validate_assumptions(assumptions)
         cls._sanitize(assumptions, cls)
         return Wild.__xnew__(cls, name, exclude, properties, **assumptions)
 
--- a/sympy/assumptions/tests/test_query.py
+++ b/sympy/assumptions/tests/test_query.py
@@ -2284,6 +2284,9 @@ def test_check_old_assumption():
     assert ask(Q.irrational(x)) is True
     assert ask(Q.rational(x)) is False
 
+    with raises(ValueError):
+        symbols('x', foo=True)
+
 
 def test_issue_9636():
     assert ask(Q.integer(1.0)) is False


oscarbenjamin:
I think there are possibly some situations in which setting arbitrary "assumptions" on symbols is actually used for some things. It looks like this is breaking some of the tests in the test suite but I'm pretty sure it would also break a lot of downstream code as well.

If it is possible to fix up the places in sympy that use this then it could be possible to give a warning for something like this but it shouldn't be an error at this stage. We also need to identify if there is a better way for people who do want to set arbitrary properties. Perhaps there could be another symbol class that can be used for that or something.

smichr:
> need to identify if there is a better way for people who do want to set arbitrary properties. Perhaps there could be another symbol class that can be used for that or something.

I've always been impressed with how strict Poly is in terms of options. Perhaps something that could be done is to add a single keyword `attr` to Symbol which would hold a dictionary of other attributes passed. Then the valid keyword would be what they currently are and `attr`. e.g. Symbol('x', real=True, attr=dict(color='green')).color -> 'green'. This forces the user to separate the applied attribute from the assumption system which would not properly handle other attributes since there are no rules written for them.

smichr:
> Symbol('x', real=True, attr=dict(color='green')).color -> 'green'

I guess that would have to be `Symbol('x', real=True, attr=dict(color='green')).attr['color'] -> 'green'`...and that seems clunky.

smichr:
I would like to see the method added even if it is not hooked up to instantiation. When things aren't working as expected I can often trace it back to a typo like `postive=True` for an assumption. With this method, a user can make their own safe routine for creation of symbols.

asmeurer:
A good solution at this stage would be to give a warning (not a deprecation warning, just a normal warning). That way, people who are using this feature can easily disable the warning using a warnings filter, and by not being a deprecation warning, we don't commit to necessarily keeping it this way or to eventually changing the warning to an error.

oscargus:
I restart (close and reopen) this as I am curious to see which tests didn't work and they have now expired.

I agree that defining a new warning type, `NonStandardAssumptionWarning` or something, could indeed be a quite nice way to handle this.

asmeurer:
I don't know if we necessarily need a separate warning type vs. just UserWarning. The way the Python warnings system works, you can filter warnings by text (and actually, due to some technical reasons it's impossible filter custom warnings types, see https://github.com/sympy/sympy/issues/15130). The reason we have a separate warning type for deprecation warnings is that we want to change the default behavior of all deprecation warnings to always show (normal DeprecationWarning warnings do not show unless you are at the interactive interpreter).

sympy/core/symbol.py
```diff
@@ -226,9 +226,20 @@ def _diff_wrt(self):
         """
         return True
 
+    @staticmethod
+    def _validate_assumptions(assumptions):
```
asmeurer, line 230: It would make more sense to have this in `_sanitize`, since every call is followed by `_sanitize` anyway.

sympy/core/symbol.py
```diff
@@ -226,9 +226,20 @@ def _diff_wrt(self):
...
+        """Validate assumption strings.
+        """
+        valid_assumptions = set(_assume_defined)
+        diff = set(assumptions) - valid_assumptions
+        if diff:
+            raise ValueError(
```
asmeurer, line 236: As noted in the discussion, it would be better to make this a warning (`warnings.warn(...)`) instead of an error. That way the existing behavior continues to work for those using it on purpose, but people will be alerted to typos. The warning message should also be something like "unknown assumptions ..." instead of "invalid assumptions ...".

PR #21421, opened 2021-05-03
Implemented function diagonal() returning kth diagonal of a DomainMatrix

ks147:
Implemented function diagonal() returning kth diagonal of a DomainMatrix

#### Release Notes

* polys
  * Implemented diagonal() for DomainMatrix class

### DIFF
--- a/sympy/polys/matrices/domainmatrix.py
+++ b/sympy/polys/matrices/domainmatrix.py
@@ -553,6 +553,55 @@ class DomainMatrix:
         rows_sympy = [[self.domain.to_sympy(e) for e in row] for row in elemlist]
         return MutableDenseMatrix(rows_sympy)
 
+    def diagonal(self, k=0):
+        r"""
+        Returns the kth diagonal of self. The main diagonal
+        corresponds to `k=0`; diagonals above and below correspond to
+        `k > 0` and `k < 0`, respectively. The values of `self[i, j]`
+        for which `j - i = k`, are returned in order of increasing
+        `i + j`, starting with `i + j = |k|`.
+
+        Examples
+        ========
+
+        >>> from sympy.polys.matrices import DomainMatrix
+        >>> from sympy.polys.domains import ZZ
+        >>> A = DomainMatrix([[ZZ(1), ZZ(2)], [ZZ(3), ZZ(4)]], (2, 2), ZZ)
+        >>> A.diagonal()
+        DomainMatrix([[1, 4]], (1, 2), ZZ)
+        >>> A.diagonal(1)
+        DomainMatrix([[2]], (1, 1), ZZ)
+        >>> A.diagonal(-1)
+        DomainMatrix([[3]], (1, 1), ZZ)
+
+        See Also
+        ========
+        DomainMatrix.diag()
+
+        """
+        from sympy.utilities.misc import filldedent
+
+        diaglist = []
+        elemlist = self.rep.to_list()
+        r = 0 if k > 0 else -k
+        c = 0 if r else k
+
+        nrows, ncols = self.shape
+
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
Generally the `DomainMatrix` class delegates most of the work for implementing something to the lower-level `DDM` or `SDM` classes so that they can each have efficient implementations for their own internal data structures. There shouldn't really be any nontrivial code in the `DomainMatrix` methods themselves.
For example the implementation here would be very inefficient for a large diagonal SDM matrix because it constructs a huge list of lists. Since efficiently getting the diagonal of a large diagonal SDM matrix is an obvious potential use for something like this then I don't think that's a good approach.

You are calling `from_list_sympy` with a list of domain elements which suggests that you have not really understood the distinction between domain elements and Expr. I suggest reading this:
https://docs.sympy.org/latest/modules/polys/domainsintro.html#domain-elements-vs-sympy-expressions

The implementation here would potentially make sense for `DDM` but it is really just a convoluted way of writing this list comprehension:
```
[M[i][i+k] for i in range(max(0,-k), min(rows, cols-k))]
```
The implementation here seems to just be a copy of the `Matrix.diagonal` method and I want to make it clear that while it does sometimes make sense to copy over code from that class the whole point of `DomainMatrix` is to be *better* so if we just copy the code over then that doesn't really help anything.

Finally I'm not exactly sure how useful getting the diagonal of a matrix is compared to other things. I would like to keep `DomainMatrix` as minimal as possible and I certainly don't want to copy all of the Matrix methods over to it because there are far too many Matrix methods. The operations that could make sense need to be considered carefully to set up the primitives that are needed to build efficient high-level classes on top. I doubt that extracting the diagonal of matrix is really a performance critical operation so I would only add it to `DomainMatrix` if it is needed for some other more important algorithm. From `git grep` I can only find two places in the sympy code base that actually use the `Matrix.diagonal` method and they both just want the main diagonal.

There are much more relevant methods missing like `DomainMatrix` doesn't even have a `transpose` method so I think it would be better focus on what is actually important.

We can add a diagonal method if it is improved but there's really no point in having it unless it is somehow better than the trivial code that anyone can write in a list comprehension to get the diagonal elements. For a large sparse matrix this would actually be less efficient than a user using the list comprehension themselves to extract the diagonal elements from the DomainMatrix.

ks147:
> You are calling `from_list_sympy` with a list of domain elements which suggests that you have not really understood the distinction between domain elements and Expr. I suggest reading this:
> https://docs.sympy.org/latest/modules/polys/domainsintro.html#domain-elements-vs-sympy-expressions

I was actually looking for something like this thanks for the link, I'll look into the difference between domain elements and Expr.


> Finally I'm not exactly sure how useful getting the diagonal of a matrix is compared to other things. I would like to keep `DomainMatrix` as minimal as possible and I certainly don't want to copy all of the Matrix methods over to it because there are far too many Matrix methods. The operations that could make sense need to be considered carefully to set up the primitives that are needed to build efficient high-level classes on top. I doubt that extracting the diagonal of matrix is really a performance critical operation so I would only add it to `DomainMatrix` if it is needed for some other more important algorithm. From `git grep` I can only find two places in the sympy code base that actually use the `Matrix.diagonal` method and they both just want the main diagonal.

Alright, looks  like we don't need `diagonal` method then.

> There are much more relevant methods missing like `DomainMatrix` doesn't even have a `transpose` method so I think it would be better focus on what is actually important.


@oscarbenjamin, I'll start working on `transpose` method instead

sylee957:
I think that this can be closed. 
I also think that `diagonal` should have a controversy picking the right representation, like whether to choose python list or 1D array or `n*1` matrix.

PR #21423, opened 2021-05-03
feat(assump) : modify ClassFactRegistry so that facts and handlers can be viewed

JSS95:

This is inspired by @asmeurer 's [comment](https://github.com/sympy/sympy/pull/21367#issuecomment-826044819).

Now, passing an expression to `ClassFactRegistry` instance returns a dictionary which contains the facts and their handlers.

```python
>>> from sympy.assumptions.sathandlers import class_fact_registry
>>> class_fact_registry(Abs(x))
{Equivalent(~Q.zero(x), ~Q.zero(Abs(x))): <function sympy.assumptions.sathandlers.<lambda>(expr)>,
 Implies(Q.odd(x), Q.odd(Abs(x))): <function sympy.assumptions.sathandlers.<lambda>(expr)>,
 Q.nonnegative(Abs(x)): <function sympy.assumptions.sathandlers.<lambda>(expr)>,
 Implies(Q.even(x), Q.even(Abs(x))): <function sympy.assumptions.sathandlers.<lambda>(expr)>,
 Implies(Q.integer(x), Q.integer(Abs(x))): <function sympy.assumptions.sathandlers.<lambda>(expr)>}
```

With this, you can print and see the facts for specific expression, and pick the handlers which return unsatisfiable facts. Or you can extract the handlers you want and build more compact fact registry.

For example,
```python

    # Build the fact registry

    >>> from sympy import Abs, Q
    >>> from sympy.logic.boolalg import Equivalent
    >>> from sympy.assumptions.sathandlers import ClassFactRegistry
    >>> reg = ClassFactRegistry()
    >>> @reg.register(Abs)
    ... def f1(expr):
    ...     return Q.nonnegative(expr)
    >>> @reg.register(Abs)
    ... def f2(expr):
    ...     arg = expr.args[0]
    ...     return Equivalent(~Q.zero(arg), ~Q.zero(expr))

    # Check the facts

    >>> from sympy.abc import x
    >>> reg(Abs(x)) # doctest: +SKIP
    {Q.nonnegative(Abs(x)): <function __main__.f1(expr)>,
     Equivalent(~Q.zero(x), ~Q.zero(Abs(x))): <function __main__.f2(expr)>}

    # Remove unnecessary fact

    >>> h = reg(Abs(x))[Q.nonnegative(Abs(x))]
    >>> reg.remove_handler(h)
    >>> reg(Abs(x)) # doctest: +SKIP
    {Equivalent(~Q.zero(x), ~Q.zero(Abs(x))): <function __main__.f2(expr)>}
```

I expect this feature to be very useful for implementing heuristic approach to evaluate complicated propositions.

`multiregister` method is removed since this change expects one-to-one relation between the handlers and the facts.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/assumptions/sathandlers.py
+++ b/sympy/assumptions/sathandlers.py
@@ -109,11 +109,13 @@ class ClassFactRegistry:
     ===========
 
     ``register`` method registers the handler function for a class. Here,
-    handler function should return a single fact. ``multiregister`` method
-    registers the handler function for multiple classes. Here, handler function
-    should return a container of multiple facts.
+    handler function should return a single fact.
 
-    ``registry(expr)`` returns a set of facts for *expr*.
+    ``multiregister`` method registers the handler function for multiple
+    classes. Here, handler function should return a container of multiple facts.
+
+    ``registry(expr)`` returns a dictionary of computed facts and their handler
+    functions for *expr*.
 
     Examples
     ========
@@ -132,12 +134,13 @@ class ClassFactRegistry:
     ...     arg = expr.args[0]
     ...     return Equivalent(~Q.zero(arg), ~Q.zero(expr))
 
-    Calling the registry with expression returns the defined facts for the
-    expression.
+    Calling the registry with expression returns a dictionary of computed
+    facts and their handler functions for the expression.
 
     >>> from sympy.abc import x
-    >>> reg(Abs(x))
-    {Q.nonnegative(Abs(x)), Equivalent(~Q.zero(x), ~Q.zero(Abs(x)))}
+    >>> reg(Abs(x)) # doctest: +SKIP
+    {Q.nonnegative(Abs(x)): <function __main__.f1(expr)>,
+     Equivalent(~Q.zero(x), ~Q.zero(Abs(x))): <function __main__.f2(expr)>}
 
     Multiple facts can be registered at once by ``multiregister`` method.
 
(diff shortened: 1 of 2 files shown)


sympy/assumptions/sathandlers.py
```diff
@@ -201,28 +188,28 @@ def __call__(self, expr):
...
-            ]
+for fact in [
+    lambda expr: allargs(x, Q.positive(x), expr) >> Q.positive(expr),
+    lambda expr: allargs(x, Q.negative(x), expr) >> Q.negative(expr),
+    lambda expr: allargs(x, Q.real(x), expr) >> Q.real(expr),
+    lambda expr: allargs(x, Q.rational(x), expr) >> Q.rational(expr),
+    lambda expr: allargs(x, Q.integer(x), expr) >> Q.integer(expr),
+    lambda expr: exactlyonearg(x, ~Q.integer(x), expr) >> ~Q.integer(expr)
+]:
+    class_fact_registry.register(Add)(fact)
```
oscarbenjamin, line 211: It's not really clear to me how this is an improvement.

JSS95: `multiregister` method is removed since this change expects one-to-one relations between the handlers and the facts, in order to pick a handler which returns the desired fact.

JSS95:
@asmeurer, can you review this change? I redesigned `ClassFactRegistry` so that user can identify, bisect, select or remove the registered facts.

PR #21424, opened 2021-05-03
Implemented DomainMatrix transpose

ks147:
Added transpose method to DomainMatrix

#### Release Notes

* polys
  * Added transpose method to DomainMatrix class

### DIFF
--- a/sympy/polys/matrices/ddm.py
+++ b/sympy/polys/matrices/ddm.py
@@ -135,6 +135,10 @@ class DDM(list):
         """
 
         MT = ddm_transpose(self)
+        nrows, ncols = self.shape
+        if nrows == 0:
+            MT = [[]]*ncols
+
         return DDM(MT, self.shape[::-1], self.domain)
 
     def __str__(self):
@@ -351,14 +355,4 @@ from .sdm import SDM
 
 
 def ddm_transpose(M):
-
-    nrows, ncols = M.shape
-    MT = []
-
-    for j in range(ncols):
-        row = []
-        for i in range(nrows):
-            row.append(M[i][j])
-        MT.append(row)
-
-    return MT
+    return list(map(list, zip(*M)))
--- a/sympy/polys/matrices/ddm.py
+++ b/sympy/polys/matrices/ddm.py
@@ -120,6 +120,23 @@ class DDM(list):
         rows = ([K.convert_from(e, Kold) for e in row] for row in self)
         return DDM(rows, self.shape, K)
 
+    def transpose(self):
+        """Returns transpose of self
+
+        Examples
+        ========
+
+        >>> from sympy.polys.matrices.ddm import DDM
+        >>> from sympy import ZZ
+        >>> A = DDM([[ZZ(1), ZZ(2), ZZ(3)], [ZZ(4), ZZ(5), ZZ(6)]], (2, 3), ZZ)
+        >>> A.transpose()
+        [[1, 4], [2, 5], [3, 6]]
+
+        """
+
+        MT = list(map(list, zip(*self)))
+        return DDM(MT, self.shape[::-1], self.domain)
+
     def __str__(self):
         rowsstr = ['[%s]' % ', '.join(map(str, row)) for row in self]
         return '[%s]' % ', '.join(rowsstr)
--- a/sympy/polys/matrices/ddm.py
+++ b/sympy/polys/matrices/ddm.py
@@ -134,7 +134,7 @@ class DDM(list):
 
         """
 
-        MT = list(map(list, zip(*self)))
+        MT = ddm_transpose(self)
         return DDM(MT, self.shape[::-1], self.domain)
 
     def __str__(self):
@@ -348,3 +348,17 @@ class DDM(list):
 
 
 from .sdm import SDM
+
+
+def ddm_transpose(M):
+
+    nrows, ncols = M.shape
+    MT = []
+
+    for j in range(ncols):
+        row = []
+        for i in range(nrows):
+            row.append(M[i][j])
+        MT.append(row)
+
+    return MT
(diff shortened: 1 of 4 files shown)


oscarbenjamin:
This should be done for both `SDM` and `DDM`. I think that there are already `ddm_transpose` and `sdm_transpose` functions that should be used in the `SDM.transpose` and `DDM.transpose` methods.

Also make sure to test base cases like matrices with shapes `(0, 2), (2, 0), (0, 0)`.

ks147:
> This should be done for both `SDM` and `DDM`. I think that there are already `ddm_transpose` and `sdm_transpose` functions that should be used in the `SDM.transpose` and `DDM.transpose` methods.

`ddm_transpose` does not exist, I could change `DDM.transpose()` to mimic `SDM.transpose()`

> Also make sure to test base cases like matrices with shapes `(0, 2), (2, 0), (0, 0)`.

I'll add this test case

ks147:
@oscarbenjamin, do you think conjugate could be useful for DomainMatrix class? It could come in handy while adding some decompositions in the future. What do you think?

oscarbenjamin:
> do you think conjugate could be useful for DomainMatrix class?

Do you mean a method that computes the complex conjugate of the domain elements? Or do you mean the conjugate-transpose (Hermitian conjugate)?

I don't think either of those is particularly urgent. What could be useful right now though is a general method for applying a function to the elements of a `DomainMatrix`. The `Matrix` class has `applyfunc` for this:
```python
In [1]: M = randMatrix(2) + randMatrix(2)*I

In [2]: M
Out[2]: 
⎡23 + 32⋅ⅈ  47 + 14⋅ⅈ⎤
⎢                    ⎥
⎣72 + 67⋅ⅈ  43 + 97⋅ⅈ⎦

In [3]: M.applyfunc(conjugate)
Out[3]: 
⎡23 - 32⋅ⅈ  47 - 14⋅ⅈ⎤
⎢                    ⎥
⎣72 - 67⋅ⅈ  43 - 97⋅ⅈ⎦
```
For `DomainMatrix` since it is usually in sparse format I think it would make sense to have two methods: `applyfunc` and `applyfunc_nonzero`. Then for a sparse matrix `applyfunc_nonzero` can act only on the nonzero elements of the matrix so it can be efficient for sparse matrices. Each of these should probably have an optional domain argument to specify the new domain after applying the function.

Probably other methods like `convert_to` or `neg` or `scalarmul` could all be defined in terms of `applyfunc_nonzero`. I suppose `convert_to` still needs to apply to the zero elements of the matrix though so maybe this needs some thought for the best way to make it work.

Potentially good abstractions like this can make it possible to reduce the complexity of the `DDM/SDM` classes by giving them a common base class that absorbs all the repetitive code. That might be a bad way to go though because the other Matrix class tried to separate things into base classes and just became a mess.

ks147:
> For `DomainMatrix` since it is usually in sparse format I think it would make sense to have two methods: `applyfunc` and `applyfunc_nonzero`. Then for a sparse matrix `applyfunc_nonzero` can act only on the nonzero elements of the matrix so it can be efficient for sparse matrices. Each of these should probably have an optional domain argument to specify the new domain after applying the function.

Okay I'll start working on `applyfunc`.



> Potentially good abstractions like this can make it possible to reduce the complexity of the `DDM/SDM` classes by giving them a common base class that absorbs all the repetitive code. That might be a bad way to go though because the other Matrix class tried to separate things into base classes and just became a mess.

Yeah this sounds like a great idea

ks147:
@oscarbenjamin, could you review this PR

sympy/polys/matrices/ddm.py
```diff
@@ -331,3 +348,17 @@ def charpoly(a):
...
+def ddm_transpose(M):
+
+    nrows, ncols = M.shape
+    MT = []
+
+    for j in range(ncols):
+        row = []
+        for i in range(nrows):
+            row.append(M[i][j])
+        MT.append(row)
```
oscarbenjamin, line 362: A function `ddm_transpose` would be expected to operate just on a list of lists so there would be no `shape` attribute.

Also I'm pretty sure that the `zip` version is a lot faster than this.

ks147: `zip` method doesn't really work with shape `(0, 2)`

oscarbenjamin: That case will need special handling. This function that only works with a list of lists can't handle the `(0, 2)` case at all.

You should do timings though. Matrix transpose is a performance sensitive operation and should be cheap. The `ddm_` function for `mul` uses transpose as its first step for example.

ks147: We need the shape to get transpose of list of lists as `[]` can represent DDM of shape `(0, n)` for any n >= 0, but transpose of `[]` varies depending on shape `(0,n)`. The transpose of `[]` for shape `(0,n)` is `[[]]*n`.

Using `zip` method we get `[]` as the transpose  of `DDM([], (0, n), ZZ)`  instead of `DDM([[]]*n, (n, 0), ZZ)`.

ks147: A very simple fix would be to simply pass the `shape` as argument to ddm_transpose like this


```python
def ddm_transpose(M, shape):
    MT = []
    nrows,ncols = shape
    
    for j in range(ncols):
        row = []
        for i in range(nrows):
            row.append(M[i][j])
        MT.append(row)

    return MT
```

oscarbenjamin: In most cases we don't really care about this distinction between the shape of empty matrices like this e.g.:
https://github.com/sympy/sympy/blob/0f807b64436da37b1ab90a595144dbbdbef27d91/sympy/polys/matrices/dense.py#L75-L79
Maybe it would be cleaner if all of the `ddm_*` functions just operated with `DDM` but I think that for this particular case it's easiest just to handle the `(0, 2)` case in `DDM.transpose` and have `ddm_transpose` return an empty list in that case.

Abdullahjavednesar:
Hi @ks147, Can you please resolve conflicts and address the reviews to complete this?

PR #21425, opened 2021-05-03
ExprWithLimits `(free|bound)_symbols` updated

smichr:

When a user-defined function is acting as a bound variable, it's arguments are implicitly bound and changes have been made to reflect this by modifying `bound_symbols` and `free_symbols` to account for those via change to `ExprWithLimits`.

`Basic.canonical_variables` will now include symbols contained within functions acting as bound symbols.

These changes are reflected in expressions returned for `as_dummy`:

```python
>>> f=Function('f')
>>> Integral(x,(f(x), y)).as_dummy()
Integral(_1, (_0, y))  <----- formerly Integral(x, (_0, y))
```

#### Release Notes

* core
  * `canonical_variables` now includes free symbols from the arguments of bound functions
* concrete
  * symbols in functions appearing as bound functions are now considered bound, too
  * ExprWithLimits (like Integral) now reports arguments of functions as free symbols when appearing as singleton limit
  * Integrals like `Integral(x, f(x))` will not treat `x` and independent of `f(x)` and will not respond to `.doit()`

### DIFF
--- a/sympy/concrete/expr_with_limits.py
+++ b/sympy/concrete/expr_with_limits.py
@@ -303,19 +303,46 @@ class ExprWithLimits(Expr):
                     isyms.update(i.free_symbols)
         return isyms
 
+    @property
+    def canonical_variables(self):
+        """Return a minimal number of dummy symbols for
+        the bound variables.
+
+        Examples
+        ========
+
+        >>> from sympy import Integral
+        >>> from sympy.abc import x, y, z
+        >>> Integral(x, (x, y), (y, z), z).canonical_variables
+        {x: _0, y: _1, z: _0}
+
+        A new dummy was not needed for variable z so the ``_0``
+        was re-used.
+
+        >>> Integral(x, (x, y), (y, z), z).as_dummy()
+        Integral(_0, (_0, _1), (_1, _0), (_0, z))
+        """
+        reps = super().canonical_variables
+        if len(self.limits) <= 2:
+            return reps
+        _0 = reps[self.limits[0][0]]
+        for i, ui in enumerate(uniq(i[0] for i in self.limits)):
+            if i > 1:
+                reps[ui] = _0
+        return reps
+
     def as_dummy(self):
-        """Return the integral with integration variables replaced
-        with a minimal set of unique, canonical symbols having only
-        the default assumption for commutativity being True.
+        """Return an expression with any limit with length of 1 replaced
+        with one of length 2.
 
         Examples
         ========
 
         >>> from sympy import Integral, Function
         >>> from sympy.abc import x, y, z
-        >>> eq = Integral(y, (x, y), (y, z), (z, y))
+        >>> eq = Integral(x, y)
         >>> eq.as_dummy()
-        Integral(_1, (_0, _1), (_1, _0), (_0, y))
+        Integral(x, (_0, y))
         >>> _.doit() == eq.doit()
         True
 
@@ -341,13 +368,10 @@ class ExprWithLimits(Expr):
         self2 = self.func(*([self.function] + [
             i if len(i) > 1 else i*2 for i in self.limits]))
         rv = Basic.as_dummy(self2)
-        dums = [i[0] for i in rv.limits]
-        redun = set().union(*[u.free_symbols for l in rv.limits for u in l[1:]]) & set(dums[2:])
-        reps = {k: rv.limits[0][0] for k in redun}
         # undo explicit form of at_pt
         return rv.func(*([rv.function] + [
             i[0] if len(i) == 2 and i[0] == i[1] else i
-            for i in rv.limits])).xreplace(reps)
+            for i in rv.limits]))
 
     @property
     def is_number(self):
--- a/sympy/concrete/expr_with_limits.py
+++ b/sympy/concrete/expr_with_limits.py
@@ -318,12 +318,9 @@ class ExprWithLimits(Expr):
 
         A new dummy was not needed for variable z so the ``_0``
         was re-used.
-
-        >>> Integral(x, (x, y), (y, z), z).as_dummy()
-        Integral(_0, (_0, _1), (_1, _0), (_0, z))
         """
         reps = super().canonical_variables
-        if len(self.limits) <= 2:
+        if len(reps) <= 2:
             return reps
         _0 = reps[self.limits[0][0]]
         free = self.function.free_symbols
--- a/sympy/concrete/expr_with_limits.py
+++ b/sympy/concrete/expr_with_limits.py
@@ -326,8 +326,9 @@ class ExprWithLimits(Expr):
         if len(self.limits) <= 2:
             return reps
         _0 = reps[self.limits[0][0]]
+        free = self.function.free_symbols
         for i, ui in enumerate(uniq(i[0] for i in self.limits)):
-            if i > 1:
+            if i > 1 and ui not in free:
                 reps[ui] = _0
         return reps
 
--- a/sympy/concrete/expr_with_limits.py
+++ b/sympy/concrete/expr_with_limits.py
@@ -304,13 +304,50 @@ class ExprWithLimits(Expr):
         return isyms
 
     def as_dummy(self):
+        """Return the integral with integration variables replaced
+        with a minimal set of unique, canonical symbols having only
+        the default assumption for commutativity being True.
+
+        Examples
+        ========
+
+        >>> from sympy import Integral, Function
+        >>> from sympy.abc import x, y, z
+        >>> eq = Integral(y, (x, y), (y, z), (z, y))
+        >>> eq.as_dummy()
+        Integral(_1, (_0, _1), (_1, _0), (_0, y))
+        >>> _.doit() == eq.doit()
+        True
+
+        If an integration variable contains free symbols that
+        also appear on their own in the integrand, it will not
+        be true that the evaluated form of the original and
+        dummy form will be the same:
+
+        >>> f = Function('f')
+        >>> eq = Integral(x + f(x), f(x))
+        >>> eq.as_dummy()
+        Integral(_0 + _1, (_0, f(x)))
+        >>> _.doit()
+        _1*f(x) + f(x)**2/2
+
+        The non-dummy form of such an integral does not evaluate:
+
+        >>> eq.doit() == eq
+        True
+        """
         from sympy.core.basic import Basic
+        # need to use explicit form of at_pt limit
         self2 = self.func(*([self.function] + [
             i if len(i) > 1 else i*2 for i in self.limits]))
         rv = Basic.as_dummy(self2)
-        if rv == self2:
-             rv = self
-        return rv
+        dums = [i[0] for i in rv.limits]
+        redun = set().union(*[u.free_symbols for l in rv.limits for u in l[1:]]) & set(dums[2:])
+        reps = {k: rv.limits[0][0] for k in redun}
+        # undo explicit form of at_pt
+        return rv.func(*([rv.function] + [
+            i[0] if len(i) == 2 and i[0] == i[1] else i
+            for i in rv.limits])).xreplace(reps)
 
     @property
     def is_number(self):
@@ -367,12 +404,18 @@ class ExprWithLimits(Expr):
                 old.free_symbols.intersection(self.free_symbols):
             sub_into_func = True
             for i, xab in enumerate(limits):
-                if 1 == len(xab) and old == xab[0]:
+                if len(xab) == 1 and old == xab[0]:
                     if new._diff_wrt:
                         xab = (new,)
                     else:
                         xab = (old, old)
-                limits[i] = Tuple(xab[0], *[l._subs(old, new) for l in xab[1:]])
+                _li = Tuple(xab[0], *[l._subs(old, new) for l in xab[1:]])
+                # simplify redundant limits (x, x)  to (x, )
+                # only if they now became so
+                if len(_li) == 2 and _li != limits[i] and \
+                        _li[0] == _li[1]:
+                    _li = _li[0]
+                limits[i] = _li
                 if len(xab[0].free_symbols.intersection(old.free_symbols)) != 0:
                     sub_into_func = False
                     break
@@ -392,10 +435,6 @@ class ExprWithLimits(Expr):
                     limits[i] = Tuple(xab[0], *[l._subs(old, new) for l in xab[1:]])
                     if old == xab[0]:
                         break
-        # simplify redundant limits (x, x)  to (x, )
-        for i, xab in enumerate(limits):
-            if len(xab) == 2 and (xab[0] - xab[1]).is_zero:
-                limits[i] = Tuple(xab[0], )
 
         # Reorder limits back to representation-form
         limits.reverse()
--- a/sympy/concrete/expr_with_limits.py
+++ b/sympy/concrete/expr_with_limits.py
@@ -340,7 +340,7 @@ class ExprWithLimits(Expr):
         ========
 
         >>> from sympy import Integral, Function
-        >>> from sympy.abc import x, y, z
+        >>> from sympy.abc import x, y
         >>> eq = Integral(x, y)
         >>> eq.as_dummy()
         Integral(x, (_0, y))
--- a/sympy/concrete/expr_with_limits.py
+++ b/sympy/concrete/expr_with_limits.py
@@ -292,15 +292,16 @@ class ExprWithLimits(Expr):
         function, limits = self.function, self.limits
         isyms = function.free_symbols
         for xab in limits:
+            free0 = xab[0].free_symbols
             if len(xab) == 1:
-                isyms.add(xab[0])
-                continue
-            # take out the target symbol
-            if xab[0] in isyms:
-                isyms.remove(xab[0])
-            # add in the new symbols
-            for i in xab[1:]:
-                isyms.update(i.free_symbols)
+                # these are all free
+                isyms.update(free0)
+            else:
+                # take out the target symbol
+                isyms -= free0
+                # add in the new symbols
+                for i in xab[1:]:
+                    isyms.update(i.free_symbols)
         return isyms
 
     @property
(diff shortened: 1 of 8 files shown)


smichr:
ping @oscarbenjamin @jksuom

oscarbenjamin:
This doesn't seem consistent to me:
```python
In [1]: integral = Integral(x,(f(x), y))

In [2]: integral
Out[2]: 
y          
⌠          
⎮ x d(f(x))
⌡          
           

In [3]: integral.doit()
Out[3]: x⋅y

In [4]: integral.as_dummy()
Out[4]: 
y         
⌠         
⎮ _1 d(_0)
⌡         
          

In [5]: integral.free_symbols
Out[5]: {y}

In [6]: integral.doit().free_symbols
Out[6]: {x, y}
```
I think that `free_symbols` should be the same for `integral` and `integral.doit()`.

smichr:
> I think that `free_symbols` should be the same for `integral` and `integral.doit()`.

As you point out on the issue page, this is a nonsensical integral. Like `Poly` raising an error for `Poly(x + sqrt(x), x)` I think we should not evaluate an integral written in terms of `f(x)` when `x` appears independently.

oscarbenjamin:
The situation where this kind of integral comes up is something like separation of variables in `dsolve`. Mathematically we write something like this:
```
dx/dt = f(x)*g(t)
int 1/f(x) dx = int g(t) dt
```
The problem is that in sympy the derivative dx/dt only makes sense if x is a function `x(t)` so then it bcomes:
```
t = Symbol('t')
x = Function('x')
Derivative(x(t), t) = f(x(t))*g(t)
Integral(1/f(x(t)), x(t)) = Integral(g(t), t)
```
Here in these integrals we definitely want to keep `x(t)` and `t` separate but I'm not sure how generally that applies.

I think we also want to preserve the idea that an indefinite integral is an antiderivative so those need to be consistent. Situations like `Derivative(f(x(t), t), x(t))` come up a lot in dynamics problems e.g. in the Lagrange equations. Those should be understood as partial derivatives where `t` is a constant when differentiating wrt `x(t)`. Maybe these problem stem from needing to express a derivative wrt a variable rather than a numbered argument of a function.

smichr:
@oscarbenjamin , please see commit message on last commit about when an integral will not evaluate.

smichr:
Hmm...guess that piece of code was uncovered before. Not sure if it is the explication of the eval-at that caused the code path to be traversed or not. I will try to add a direct test of this path.

smichr:
ping @oscarbenjamin -- this is passing now and the objective for the doit and non-doit forms is attained by not doing an integral that does not contain enough information. From the commit message:

> Integral(x + y, y) is not evaluated if y is f(x) just like it is not evaluated when x = f(y).

An error in handling None was also fixed and I searched for `\(.* and .*\) is None` and did not find any other instances.

sympy/core/basic.py
```diff
@@ -582,9 +582,8 @@ def canonical_variables(self):
         names = {i.name for i in self.free_symbols - set(bound)}
         for b in bound:
             d = next(dums)
-            if b.is_Symbol:
-                while d.name in names:
-                    d = next(dums)
+            while d.name in names:
+                d = next(dums)
```
oscarbenjamin, line 586: What happens if StopIteration is raise here?

oscarbenjamin:
This still seems inconsistent to me:
```python
In [5]: integral.free_symbols
Out[5]: {y}

In [6]: integral.doit().free_symbols
Out[6]: {x, y}
```
Is the intention of this PR that we should also change what `Integral.doit` does so the integral remains unevaluated or raises an error? If so I think I would want to see the effect of that change before changing free symbols to match it. Are there many places in the codebase that depend on this?

smichr:
> This still seems inconsistent to me:

hmmm...must have gotten lost in the shuffle. I pushed a change and will have to see the impact now on the rest of the suite.

sympy/integrals/integrals.py
```diff
@@ -397,6 +401,11 @@ def doit(self, **hints):
         from sympy.concrete.summations import Sum
         if not hints.get('integrals', True):
             return self
+        bv = [f[0] for f in self.limits if not f[0].is_Symbol]
+        if bv:
+            free = self.function.free_symbols - set(bv)
```
smichr, line 406:
```suggestion
            free = self.function.expr_free_symbols - set(bv)
```

sympy/integrals/integrals.py
```diff
@@ -397,6 +401,11 @@ def doit(self, **hints):
...
+            if any(i.free_symbols & free for i in bv):
```
smichr, line 407:
```suggestion
            free = self.function.expr_free_symbols - set(bv)
            if free and any(i.expr_free_symbols & free for i in bv):
```

sympy/core/basic.py
smichr: It's an infinite generator so I can't imagine a scenario for this.

sympy/integrals/integrals.py
```diff
@@ -397,6 +401,11 @@ def doit(self, **hints):
         from sympy.concrete.summations import Sum
         if not hints.get('integrals', True):
             return self
+        bv = [f[0] for f in self.limits if not f[0].is_Symbol]
```
smichr, line 404:
```suggestion
        bv = [f[0] for f in self.limits if f[0].expr_free_symbols != {f[0]}]
```

sympy/integrals/integrals.py
```diff
@@ -397,6 +401,11 @@ def doit(self, **hints):
         from sympy.concrete.summations import Sum
         if not hints.get('integrals', True):
             return self
+        bv = [f[0] for f in self.limits if f[0].expr_free_symbols != {f[0]}]
```
smichr, line 404:
```suggestion
        bv = [f[0] for f in self.limits if (
            f[0].expr_free_symbols != {f[0]} and
            f[0].free_symbols != {f[0]}):
```

sympy/integrals/integrals.py
```diff
@@ -397,6 +401,13 @@ def doit(self, **hints):
         from sympy.concrete.summations import Sum
         if not hints.get('integrals', True):
             return self
+        bv = [f[0] for f in self.limits if (
+            f[0].expr_free_symbols != {f[0]} and
+            f[0].free_symbols != {f[0]}):
+        if bv:
+            free = self.function.expr_free_symbols - set(bv)
```
smichr, line 408:
```suggestion
            f[0].free_symbols != {f[0]})]:
        if bv:
            bv = set(bv)
            free = self.function.expr_free_symbols - bv
```

sympy/integrals/integrals.py
```diff
@@ -397,6 +401,14 @@ def doit(self, **hints):
         from sympy.concrete.summations import Sum
         if not hints.get('integrals', True):
             return self
+        bv = [f[0] for f in self.limits if (
+            f[0].expr_free_symbols != {f[0]} and
+            f[0].free_symbols != {f[0]})]:
```
smichr, line 406:
```suggestion
            f[0].free_symbols != {f[0]})]
```

smichr:
> I think I would want to see the effect of that change before changing free symbols to match it. Are there many places in the codebase that depend on this?

The `free_symbols` for `Integral` has only been changed to consider free symbols of the variables of integration. So instead of `f(x)` alone being added when it is the dummy variable, now `x` is added, too. But the change to not evaluate an integral with `f(x)` as a dummy variable if there is a free `x` in the integrand is now enforced and appears to have no impact on the existing suite.

oscarbenjamin:
I wonder if we're just making everything complicated by allowing functions to be integration variables at all.

I don't know whether we should really reserve that kind of usage for something like a Riemann-Stieltjes integral:
https://en.wikipedia.org/wiki/Riemann%E2%80%93Stieltjes_integral

smichr:
> everything complicated by allowing functions to be integration variables at all

The complication will still be there when we allow Indexed and Normal as integration variables. There may be others, but I at least tripped over those two which behave differently than AppliedUndef.

oscarbenjamin:
> The complication will still be there when we allow Indexed and Normal as integration variables.

Do we need to allow these? I think we do need to be able to define somehow what is allowed and exactly what things should be considered "symbol-like" and make sure that it is possible to write code that can handle symbol-like objects without needing to special case these types.

smichr:
> Do we need to allow these?

Maybe we should just make it a printing issue so Subs(Integral(x, x), (x, f(x))) prints as `Integral(f(x), f(x))` and entry of the later generates the former automatically. So the user would never see this. But maybe this is a bad idea since such things are public. 

Maybe a private attribute could hold what the integration variable is. The private variable could be used for representation but when working internally with the integral it would be easier to not have to worry about running into non-symbols. Things like `bound_symbols` and `free_symbols` would still have to take the value into consideration, however, since something like `integrate(x + _, _); _ = f(x)` should still return an unevaluated `Integral`.

Nagging suspicion that this really doesn't solve anything since you still have to look at what the integration variables is when making decisions about `subs`, for example. 

So what can be an integration variable? We hack for this by defining `_diff_wrt`:
```python
>>> from sympy.stats import *
>>> X = Normal('X', 2, 4)
>>> X._diff_wrt
True
```

Maybe we need to make that a public attribute and decide on a good name -- maybe `.is_symbol` in spirit of `is_number` or `is_integer` being different than `is_Number` and `is_Integer`.

smichr:
> code that can handle symbol-like objects without needing to special case these types.

note that the only special attention given in this PR is to check for when something is not `is_Symbol` -- so `is_Symbol` being true indicates the base case.

smichr:
Getting a better definition of what is and is not allowed as a variable of integration can go in another PR. This one at least resolves issues with getting `as_dummy` to correctly identify the dummy variables and stops itnegrals with symbol-dependent variables from evaluating if that symbol is also free in the integrand.

Will commit in a few hours unless there is objection.

oscarbenjamin:
I'm not sure about the way `at_dummy` loses the information relating the different "symbols":
```python

In [10]: I = Integral(x + f(x), (f(x), 0, f(x)))

In [11]: I
Out[11]: 
f(x)                   
 ⌠                     
 ⎮   (x + f(x)) d(f(x))
 ⌡                     
 0                     

In [12]: I.doit()
Out[12]: 
f(x)                   
 ⌠                     
 ⎮   (x + f(x)) d(f(x))
 ⌡                     
 0                     

In [13]: I.as_dummy().doit()
Out[13]: 
           2   
          f (x)
_1⋅f(x) + ─────
            2 
```
Ideally we would have an invariant like `I.doit() == I.as_dummy().doit()`.

On master we have:
```python
In [1]: I = Integral(x + f(x), (f(x), 0, f(x)))

In [2]: I.as_dummy()
Out[2]: 
_0               
⌠                
⎮  (_0 + x) d(_0)
⌡                
0                

In [3]: I.as_dummy().doit()
Out[3]: 
  2       
_0        
─── + _0⋅x
 2        

In [4]: I.doit()
Out[4]: 
          2   
         f (x)
x⋅f(x) + ─────
           2  
```
That also seems wrong because the dummy should be gone after the `doit`.

smichr:
> That also seems wrong because the dummy should be gone after the `doit`.

You are asking `as_dummy` to do more than it was intended to do. The intention was to give a canonical expression -- not necessarily the same mathematically -- of the original expression for purposes of comparison. It should always be true that two expressions that give the same result except for names of variables will be the same when expressed in `as_dummy` form. It's possible to make `as_dummy` be the same as the original but with different names; this requires recreating the class of object that is being replaced with a Symbol. 

So is this a to be considered a bug or a feature?

oscarbenjamin:
I guess what you're saying is that the invariant should be that if the doits can evaluate then
```
e1.doit() == e2.doit()   <-->   e1.as_dummy() == e2.as_dummy()
```

smichr:
> that if the doits can evaluate then

yes; `as_dummy` just yields a canonical expression that can be tested modulo free symbols. Using `doit` on the `as_dummy` result would be like using `evalf` twice on an expression in that the precision of the second instance would no longer reflect the precision -- actual precision -- of the first result.

oscarbenjamin:
Okay how about this:
```python
In [18]: I1 = Integral(x, (x, y), (y, x), (x, y))

In [19]: I2 = Integral(x, (x, y), (y, z), (z, y))

In [20]: I1.doit() == I2.doit()
Out[20]: True

In [21]: I1.as_dummy() == I2.as_dummy()
Out[21]: False

In [22]: I1.as_dummy()
Out[22]: 
y _0 _1                     
⌠ ⌠  ⌠                      
⎮ ⎮  ⎮  _0 d(_0) d(_1) d(_0)
⌡ ⌡  ⌡                      
                            

In [23]: I2.as_dummy()
Out[23]: 
y _2 _1                     
⌠ ⌠  ⌠                      
⎮ ⎮  ⎮  _0 d(_0) d(_1) d(_2)
⌡ ⌡  ⌡ 
```

smichr:
> Okay how about this:

I agree. They should be the same. I'll take a look.

smichr:
This was stress tested with
```python
v = var('a:d')
lim = list(subsets(v,repetition=True))
for i in subsets(v, repetition=True):
 if not i:continue
 i = Add(*i)
 for j in subsets(lim, repetition=True):
  j = [k for k in j if k]
  if not j:continue
  I = Integral(i,*j)
  try:assert I.doit() == I.as_dummy().doit()
  except:
   a,b,c=(I,I.doit(),I.as_dummy())
   assert None, a
```

smichr:
tricky wrangling needed; closing until done

sympy/core/tests/test_basic.py
```diff
@@ -299,14 +299,63 @@ def test_as_dummy():
...
+     if not i:continue
+     i = Add(*i)
+     for j in subsets(lim, repetition=True):
+      j = [k for k in j if k]
+      if not j:continue
+      I = Integral(i,*j)
+      try:assert I.doit() == I.as_dummy().doit()
+      except:
+       a,b,c=(I,I.doit(),I.as_dummy())
+       assert None, a
```
smichr, line 348:
```suggestion


def test_stress():
    from sympy.utilities.iterables import subsets
    v = symbols('a:d')
    lim = list(subsets(v,repetition=True))
    for i in subsets(v, repetition=True):
     if not i:continue
     i = Add(*i)
     for j in subsets(lim, repetition=True):
      j = [k for k in j if k]
      if not j:continue
      I = Integral(i,*j)
      if I.doit() != I.as_dummy().doit():
       assert None, (I, I.doit())
```

sympy/concrete/expr_with_limits.py
```diff
@@ -282,17 +291,89 @@ def free_symbols(self):
...
+
+    def as_dummy(self):
+        """Return an expression with any limit with length of 1 replaced
+        with one of length 2.
+
+        Examples
+        ========
+
+        >>> from sympy import Integral, Function
+        >>> from sympy.abc import x, y, z
```
smichr, line 343:
```suggestion
        >>> from sympy.abc import x, y
```

sympy/core/tests/test_basic.py
```diff
@@ -299,14 +299,61 @@ def test_as_dummy():
...
+    assert I1.as_dummy() == Integral(
+        _0, (_0, x), (_1, _0), _0, (_0, z, y)), I1.as_dummy()
+    assert Integral(x, y, y, x).as_dummy(
+        ) == Integral(_1, _0, (_0, y), (_1, x))
+    assert Integral(x, y, z, (x, x)).as_dummy(
+        ) == Integral(_2, (_0, y), (_1, z), (_2, x))
+
+
+def test_stress():
+    from sympy.utilities.iterables import subsets
```
smichr, line 335:
```suggestion
    from sympy import Add
    from sympy.utilities.iterables import subsets
```

sympy/concrete/expr_with_limits.py
```diff
@@ -282,17 +291,89 @@ def free_symbols(self):
...
+        """
+        reps = super().canonical_variables
+        if len(self.limits) <= 2:
+            return reps
+        _0 = reps[self.limits[0][0]]
+        free = self.function.free_symbols
+        for i, ui in enumerate(uniq(i[0] for i in self.limits)):
+            if i > 1 and ui not in free:
+                reps[ui] = _0
+        return reps
```
smichr, line 333:
```suggestion
        """
        reps = super().canonical_variables
        if len(reps) <= 2:
            return reps
        _0 = reps[self.limits[0][0]]
        free = self.function.free_symbols
        for i, ui in enumerate(uniq(i[0] for i in self.limits)):
            if i > 1 and ui not in free:
                reps[ui] = _0
        return reps
```

sympy/concrete/expr_with_limits.py
```diff
@@ -282,17 +291,89 @@ def free_symbols(self):
...
+        >>> Integral(x, (x, y), (y, z), z).as_dummy()
+        Integral(_0, (_0, _1), (_1, _0), (_0, z))
+        """
+        reps = super().canonical_variables
+        if len(self.limits) <= 2:
+            return reps
+        _0 = reps[self.limits[0][0]]
+        free = self.function.free_symbols
+        for i, ui in enumerate(uniq(i[0] for i in self.limits)):
+            if i > 1 and ui not in free:
```
smichr, line 328:
```suggestion
            if i > 1 and ui in reps:
```

sympy/core/tests/test_basic.py
```diff
@@ -299,14 +299,62 @@ def test_as_dummy():
...
+    lim = list(subsets(v,repetition=True))
+    for i in subsets(v, repetition=True):
+     if not i:continue
+     i = Add(*i)
+     for j in subsets(lim, repetition=True):
+      j = [k for k in j if k]
+      if not j:continue
+      I = Integral(i,*j)
+      if I.doit() != I.as_dummy().doit():
+       assert None, (I, I.doit())
```
smichr, line 347:
```suggestion
       assert None, (I, I.as_dummy())
```

sympy/concrete/expr_with_limits.py
```diff
@@ -282,17 +291,86 @@ def free_symbols(self):
...
+        {x: _0, y: _1, z: _0}
+
+        A new dummy was not needed for variable z so the ``_0``
+        was re-used.
+        """
+        reps = super().canonical_variables
+        if len(reps) <= 2:
+            return reps
+        _0 = reps[self.limits[0][0]]
+        free = self.function.free_symbols
```
smichr, line 326:
```suggestion
```

PR #21427, opened 2021-05-03
Use DomainMatrix in holonomic

oscarbenjamin:

Follows #21402 (see https://github.com/sympy/sympy/pull/21402#discussion_r623897334)

Removes the `NewMatrix` class from the holonomic module and makes the holonomic code use `DomainMatrix` instead.

#### Other comments

After this PR PolyMatrix is the only obstacle to enforcing that the elements of a `Matrix` should be `Expr`.

#### Release Notes

* holonomic
    * The holonomic module now uses `DomainMatrix` instead of `NewMatrix`. The `NewMatrix` class is removed.


oscarbenjamin:
I think that the changes to `DomainMatrix` here can be improved but I also think that those improvements can wait for another PR. The main goal of this PR is just to eliminate NewMatrix which it does. Once PolyMatrix is also built on top of DomainMatrix and DomainMatrix is used internally as part of Matrix it will be easier to see what actually makes sense for the interface that DomainMatrix should provide.

sympy/polys/matrices/domainmatrix.py
```diff
@@ -556,6 +556,18 @@ def to_Matrix(self):
...
+        """Matrix transpose of ``self``"""
+        return self.from_rep(self.rep.transpose())
+
+    def flat(self):
+        rows, cols = self.shape
+        return [self[i,j].element for i in range(rows) for j in range(cols)]
+
+    @property
+    def is_zero_matrix(self):
+        return all(self[i, j].element == self.domain.zero for i in range(self.shape[0]) for j in range(self.shape[1]))
```
jksuom, line 569: I might be slightly more efficient to take a local copy of `self.domain.zero`.

jksuom:
This looks good to me though I didn't check carefully `_find_nonzero_solution` and `_solve`. I guess they might be preliminary anyway.

Issue #9154, opened 2015-03-16
ln(p) is not known to be greater than 1 for an odd prime p


pelegm:
Check this:
```python
>>> p = Symbol('p', prime=True, odd=True)  # hence p is at least 3
>>> p >= 3
True
>>> ln(p) >= 1
log(p) >= 1
```

This is related to #9116.

PR #19948, opened 2020-08-11
Core: Allow Big numbers to be compared

seadavis:
#### References to other Issues or PRs
#6835 

#### Brief description of what is fixed or changed
Fixing very large numbers not being compared,  by not forcing Pow to be evaluated in Multiplication or additions, via multiplication or addition.
If a power cannot be evaluated during Mul._flatten or Add._flatten, it will simply add it to the arguments.
Also, by blocking evalf when a Power would resulting in a floating point overflow error.

Examples:
    >>> from sympy import Pow, S
    >>> Pow(10, Pow(10, 100)) < Pow(10, Pow(10, 101))
    False

Example if we assume this blocks powers of 200 or above are to remain unevaluated.
    >>>print(Mul(Pow(200, Pow(200)), 2))
    200**200 + 2

Before the changes this would have been something like:
1.6069380442589902755419620923412e+460

#### Release Notes

* core
    * enabled large Pow comparisons to occur.

PR #21431, opened 2021-05-04
[wip] enhanced relational comparisons

### LINKED ISSUE #21429: Determining floor of rational expression
I think it would be great if sympy could determine
```
x = sympy.Symbol("x", integer=True, positive=True)
sympy.floor(-1/x) == -1
```
In other words,
```
0 > -1/x >= -1
```
or
```
0 < 1/x <= 1
```

smichr:

fixes #21429

cf #9154

Simplification of Relationals with denominators of known sign is better handled now so

```python
>>> x = var('x',positive=1)
>>> 1/x < 1
1/x < 1
>>> simplify(_)
x > 1
```

In addition, `floor` and `ceiling` now are aware of arguments of size with magnitude larger or smaller than 1:
```python
floor(-1/Dummy(prime=True)) == -1
```
#### Other comments

This picture is useful for checking assumptions that have been coded.

image

What has been difficult is knowing what cases to code in the `_eval_is_ge` routines since some expressions will evaluate with assumptions if these handlers do nothing. So I have tried to add every case. I have also attempted to not do any recalculation of any expression other than to multiple or add Numbers to existing Numbers.

- [ ] fix this

```python
    >>> Dummy(zero=True)**Dummy(infinite=True) >= 0
    True  <-- should be None
    >>> Dummy(zero=True)**Dummy(infinite=True) > 0
    False  <-- should be None
    >>> 0**-oo
    zoo
```
#### Release Notes

* functions
  * floor and ceiling recognize unit fractions whose magnitude is less than 1
* core
  * better simplification of Rationals with rational expressions


sympy/functions/elementary/integers.py
```diff
@@ -20,11 +20,24 @@ class RoundFunction(Function):
...
+            if cls is floor:
+                if n == -1 and d.is_positive:
+                    return n
+                if n == 1 and d.is_negative:
+                    return -n
+            if cls is ceiling:
+                if n == 1 and d.is_positive:
+                    return n
+                if n == -1 and d.is_negative:
+                    return -n
```
oscarbenjamin, line 39: It seems odd to add this as a special case here. Really what this block is doing is:
```python
if arg.is_nonnegative and is_lt(arg, S.One):
    return S.Zero
elif arg.is_negative and is_ge(arg, S.NegativeOne):
    return S.NegativeOne
```
We could make `is_ge/is_lt` handle this by including handlers for `_eval_is_ge(Pow, ...)` and so on.

To me it would seem more useful for the codebase as a whole to have the code here depend on general purpose handlers and to improve the capabilities of the general purpose handlers for use in other parts of the codebase as well.

oscarbenjamin: Note that adding this capability to `is_ge` would immediately make it available to `Relational` as well.

smichr:
In `relational.py` I define
```
@dispatch(Pow, Expr)
def _eval_is_le(lhs, rhs):
```
but the following does not dispatch there:
```python
>>> from sympy.core.relational import is_le
>>> var('i',integer=1,positive=1)
>>> is_le(1/i,S.One)
```

oscarbenjamin:
Everything is defined in terms of `_eval_is_ge` so there is no `_eval_is_le`. In your case
```
is_le(1/i, 1)   -->  is_ge(1, 1/i)
```
So what is needed is 
```
@dispatch(Expr, Pow)
def _eval_is_ge(lhs, rhs):
```
Or perhaps something more precise like:
```
@dispatch(type(S.One), Pow)
def _eval_is_ge(lhs, rhs):
    base, exp = rhs.as_base_exp()
    if exp.is_nonpositive and base.is_positive and is_le(base, S.One):
        return True 
```

It looks like `_eval_is_ge` is barely developed so far. There was some work on it but it stalled https://github.com/sympy/sympy/pull/19948

sympy/functions/elementary/integers.py
```diff
@@ -20,11 +20,23 @@ class RoundFunction(Function):
 
     @classmethod
     def eval(cls, arg):
-        from sympy import im
+        from sympy import im, fraction
```
smichr, line 23:
```suggestion
        from sympy import im
```

sympy/core/relational.py
```diff
@@ -1075,6 +1087,40 @@ def _n2(a, b):
...
+
+@dispatch(Pow, Number)
+def _eval_is_ge(lhs, rhs):
+    if rhs:
+        base, exp = lhs.as_base_exp()
+        if exp.is_nonpositive and is_le(base, 1/rhs):
+            return True
+
+
+@dispatch(Number, Mul)
```
smichr, line 1106:
```suggestion
@dispatch(Number, Mul) # type: ignore
```

sympy/core/relational.py
```diff
@@ -1075,6 +1087,40 @@ def _n2(a, b):
             return dif
 
 
+@dispatch(Number, Pow)
```
smichr, line 1090:
```suggestion
@dispatch(Number, Pow) # type: ignore
```

smichr:
> Everything is defined in terms of `_eval_is_ge`

I get the idea, but this is not an idiom I grew up with. Having worked on the assumption system, this addition is a welcome one. I think I have this working. The tests I added and the floor/ceiling tests passed. If the rest pass, this is ready for a review, @oscarbenjamin .

sympy/core/relational.py
```diff
@@ -1075,6 +1087,40 @@ def _n2(a, b):
             return dif
 
 
+@dispatch(Number, Pow) # type: ignore
+def _eval_is_ge(lhs, rhs):
```
smichr, line 1091:
```suggestion
@dispatch(Number, Pow)
def _eval_is_ge(lhs, rhs): # noqa:F811
```

sympy/core/relational.py
```diff
@@ -1075,6 +1087,40 @@ def _n2(a, b):
...
+    if lhs:
+        base, exp = rhs.as_base_exp()
+        if exp.is_nonpositive and is_ge(base, 1/lhs):
+            return True
+
+
+@dispatch(Pow, Number) # type: ignore
+def _eval_is_ge(lhs, rhs):
```
smichr, line 1099:
```suggestion
@dispatch(Pow, Number)
def _eval_is_ge(lhs, rhs): # noqa:F811
```

sympy/core/relational.py
```diff
@@ -1075,6 +1087,40 @@ def _n2(a, b):
...
+    if rhs:
+        base, exp = lhs.as_base_exp()
+        if exp.is_nonpositive and is_le(base, 1/rhs):
+            return True
+
+
+@dispatch(Number, Mul) # type: ignore
+def _eval_is_ge(lhs, rhs):
```
smichr, line 1107:
```suggestion
@dispatch(Number, Mul)
def _eval_is_ge(lhs, rhs): # noqa:F811
```

sympy/core/relational.py
```diff
@@ -1075,6 +1087,40 @@ def _n2(a, b):
...
+    from sympy.simplify.radsimp import fraction
+    num, den = fraction(rhs)
+    if num.is_Number and den != 1:
+        if is_le(den, lhs/num):
+            return True
+
+
+@dispatch(Mul, Number)  # type: ignore
+def _eval_is_ge(lhs, rhs):
```
smichr, line 1116:
```suggestion
@dispatch(Mul, Number)
def _eval_is_ge(lhs, rhs): # noqa:F811
```

sympy/core/relational.py
```diff
@@ -1075,6 +1087,40 @@ def _n2(a, b):
...
+def _eval_is_ge(lhs, rhs): # noqa:F811
+    from sympy.simplify.radsimp import fraction
+    num, den = fraction(lhs)
+    if num.is_Number and den != 1:
+        if is_ge(den, rhs/num):
+            return True
+
+
 @dispatch(Expr, Expr)
 def _eval_is_ge(lhs, rhs):
```
smichr, line 1125:
```suggestion
def _eval_is_ge(lhs, rhs): # noqa:F811
```

sympy/core/relational.py
```diff
@@ -1075,8 +1087,42 @@ def _n2(a, b):
...
+            return True
+
+
+@dispatch(Number, Mul)
+def _eval_is_ge(lhs, rhs): # noqa:F811
+    from sympy.simplify.radsimp import fraction
+    num, den = fraction(rhs)
+    if num.is_Number and den != 1:
+        if is_le(den, lhs/num):
+            return True
```
smichr, line 1112: if num and num.is_Number and den != 1:
    a, b= den, lhs/num
    if (num < 0):
        a, b = b, a
    if is_le(a, b):
        return True

sympy/core/relational.py
```diff
@@ -1075,8 +1087,42 @@ def _n2(a, b):
...
+
+
+@dispatch(Mul, Number)
+def _eval_is_ge(lhs, rhs): # noqa:F811
+    from sympy.simplify.radsimp import fraction
+    num, den = fraction(lhs)
+    if num.is_Number and den != 1:
+        if is_ge(den, rhs/num):
+            return True
```
smichr, line 1121:
```suggestion
```suggestion
    if num and num.is_Number and den != 1:
        a, b = den, rhs/num
        if (num < 0):
            a, b = b, a
        if is_ge(a, b):
            return True
``````

File: sympy/core/relational.py [hunk 12]
```diff
@@ -1075,8 +1087,49 @@ def _n2(a, b):
...
+            a, b = b, a
+        if is_le(a, b):
+            return True
+
+
+@dispatch(Mul, Number)
+def _eval_is_ge(lhs, rhs): # noqa:F811
+    from sympy.simplify.radsimp import fraction
+    num, den = fraction(lhs)
+```suggestion
```
smichr, line 1122:
```suggestion
```

sympy/core/relational.py
```diff
@@ -1075,8 +1087,48 @@ def _n2(a, b):
             return dif
 
 
+@dispatch(Number, Pow)
+def _eval_is_ge(lhs, rhs): # noqa:F811
+    if lhs:
+        base, exp = rhs.as_base_exp()
+        if exp.is_nonpositive and is_ge(base, 1/lhs):
+            return True
```
oscarbenjamin, line 1095: We should try to avoid creating new Basic instances like `1/lhs`.

Is the logic here correct?

If we have
```
1/2 >= 5 ** 0
```
Then
```
lhs = 1/2
base = 5
exp = 0
```
So `exp.is_nonpositive = True` and `base >= 1/lhs` so we return True but the inequality is not true.

smichr: I think that is why the tests are failing...I have to look more closely at this. 

As to creating different objects, that is the purpose of working with only Number but perhaps I should have limited that to Rational and then we wouldn't be making a new type.

smichr:
working on less evaluation

smichr:
@oscarbenjamin , if you get a chance would you please take a look again. It's not really clear to me what is supposed to be handled where. So there are many other cases that could *potentially* be handled in the Rational/Pow cases but many of them evaluate before getting to these `_eval_is_ge` routines.

And I added Expr/Pow handlers but was warned about ambiguity so I just added the handling directly in the Expr/Expr stub.

sympy/core/relational.py
```diff
@@ -1075,9 +1087,72 @@ def _n2(a, b):
             return dif
 
 
+@dispatch(Rational, Pow)
+def _eval_is_ge(rat, rhs): # noqa:F811
+    if rat.is_zero:
+        return
+    base, exp = rhs.as_base_exp()
+    if base.is_nonnegative and (exp + 1).is_nonpositive:
+        return is_ge(base, rat)
```
oscarbenjamin, line 1096: Here `exp + 1` creates a new Basic instance. It would be better to use `is_ge(exp, 1)`.

sympy/core/relational.py
```diff
@@ -1075,9 +1087,72 @@ def _n2(a, b):
             return dif
 
 
+@dispatch(Rational, Pow)
+def _eval_is_ge(rat, rhs): # noqa:F811
+    if rat.is_zero:
+        return
+    base, exp = rhs.as_base_exp()
+    if base.is_nonnegative and (exp + 1).is_nonpositive:
```
smichr, line 1095:
```suggestion
    # i >= b**((+/-1)/j)
    if rat.is_zero:
        return
    base, exp = rhs.as_base_exp()
    if base.is_nonnegative and is_le(exp, S.NegativeOne):
```

sympy/core/relational.py
```diff
@@ -1075,9 +1087,79 @@ def _n2(a, b):
...
+
+@dispatch(Pow, Rational)
+def _eval_is_ge(lhs, rat): # noqa:F811
+    # b**((+/-1)/j) >= i
+    if rat.is_zero:
+        return
+    base, exp = lhs.as_base_exp()
+    if base.is_nonnegative:
+        if is_ge(base, S.One) and is_ge(exp, S.One):
+            return is_le(rat, S.One)
```
oscarbenjamin, line 1111: So we want to know if `a**b >= c`. The inference here is
```
a >= 1  &  b >= 1    --->     a**b >= c  iff  c <= 1
```
That's not an if and only if though because e.g. if a = 2 and b = 2 and c = 2 then this would return False even though the statement is true.

sympy/core/relational.py
```diff
@@ -1075,9 +1087,79 @@ def _n2(a, b):
...
 
+@dispatch(Rational, Pow)
+def _eval_is_ge(rat, rhs): # noqa:F811
+    # i >= b**((+/-1)/j)
+    if rat.is_zero:
+        return
+    base, exp = rhs.as_base_exp()
+    if base.is_nonnegative:
+        if is_le(exp, S.NegativeOne):
+            return is_ge(base, rat)
```
smichr, line 1098:
```suggestion
            return is_ge(base, 1/rat)  # rat >= b^(e<=-1) -> b^(e>=1) >= 1/rat
```

sympy/core/relational.py
smichr, line 1111:
```suggestion
            if is_le(rat, S.One):
                return True
```

sympy/core/relational.py
```diff
@@ -6,11 +6,15 @@
 from .basic import Basic
 from .evalf import EvalfMixin
 from .function import AppliedUndef
+from .mul import Mul
```
smichr, line 9:
```suggestion
```

sympy/core/relational.py
```diff
@@ -1075,9 +1087,93 @@ def _n2(a, b):
...
+            elif exp < 0:
+                if is_le(base, S.NegativeOne) and is_le(rat, S.NegativeOne):
+                    return True
+
+
 @dispatch(Expr, Expr)
-def _eval_is_ge(lhs, rhs):
-    return None
+def _eval_is_ge(lhs, rhs): # noqa:F811
+    # simplify expression in case one side is Rational
```
smichr, line 1128:
```suggestion
    from sympy.simplify.radsimp import fraction
    # simplify expression in case one side is Rational
```

sympy/core/relational.py
```diff
@@ -1075,9 +1086,94 @@ def _n2(a, b):
...
+def _eval_is_ge(lhs, rhs): # noqa:F811
+    from sympy.simplify.radsimp import fraction
+    # simplify expression in case one side is Rational
+    # to minimize the number of handlers needed, e.g.
+    # -1/x >= 2 is Mul, Integer but can be Pow, Integer by
+    # rewriting as 1/x <= -2
+    op = is_le, is_ge
+    ix = -1
+    L, R = lhs, rhs
+    if L.is_Rational:
```
smichr, line 1135:
```suggestion
    if L.is_Rational:
        if R.is_Rational:
            return  # normal assumptions will handle this
```

sympy/core/relational.py
```diff
@@ -1075,9 +1086,96 @@ def _n2(a, b):
...
+@dispatch(Pow, Rational)
+def _eval_is_ge(lhs, rat): # noqa:F811
+    # b**((+/-1)/j) >= i
+    if rat.is_zero:
+        return  # can resolve based on assumptions of lhs
+    base, exp = lhs.as_base_exp()
+    if base.is_nonnegative:
+        if is_ge(base, S.One) and is_ge(exp, S.One):
+            if is_le(rat, S.One):
+                return True
```
oscarbenjamin, line 1111: So this is `a ** b >= c` and we check `a >= 0` and then `a >= 1` and `b >=1` then if `c <= 1` we return True. I think we only need `b >= 0` here.

Essentially this is:
```
def is_ge(lhs, rat):
    if is_ge(lhs, S.One) and is_le(rat, One):
        return True
```
Maybe it makes sense to write it like that and let a `Pow`, `One` handler take care of further details. Perhaps one of the most important usages of the `is_ge` framework is comparisons against `S.One`...

sympy/core/relational.py
```diff
@@ -1075,9 +1086,92 @@ def _n2(a, b):
...
+    # x >= x**e
+    if isinstance(rhs, Pow) and not isinstance(lhs, Pow):
+        if rhs.base == lhs and lhs.is_nonnegative and is_le(
+                rhs.exp, S.One):
+            return True
+    # x**e >= x
+    if isinstance(lhs, Pow) and not isinstance(rhs, Pow):
+        if lhs.base == rhs and rhs.is_nonnegative and is_ge(
+                lhs.exp, S.One):
+            return True
```
smichr, line 1174:
```suggestion
def _eval_is_ge(lhs, rhs): # noqa:F811
    from sympy.core.function import expand_mul

    if lhs.is_Atom and rhs.is_Atom:
        return  # resolve with assumptions of lhs, rhs

    # x >= x**e
    if isinstance(rhs, Pow) and not isinstance(lhs, Pow):
        if rhs.base == lhs and lhs.is_nonnegative and is_le(
                R.exp, S.One):
            return True
    # x**e >= x
    if isinstance(lhs, Pow) and not isinstance(rhs, Pow):
        if lhs.base == rhs and R.is_nonnegative and is_ge(
                lhs.exp, S.One):
            return True

    # make expression canonical wrt Rationals
    # to minimize the number of handlers needed, e.g.
    # -1/x >= 2 is Mul, Integer but can be Pow, Integer by
    # rewriting as 1/x <= -2
    op = is_le, is_ge
    ix = -1
    L, R = lhs, rhs
    n, d = [expand_mul(i) for i in (L - R).as_numer_denom()]
    if is_ge(d, S.Zero):
        L = n
    elif is_le(d, S.Zero):
        L = n
        ix = ~ix
    else:
        L = n/d
    a, L = L.as_coeff_Add()
    R = -a
    m, L = L.as_coeff_Mul()
    if m < 0:
        ix = ~ix
    R /= m
    if (L, R) != (lhs, rhs):
        return op[ix](L, R)
```

sympy/core/tests/test_relational.py
```diff
@@ -1211,3 +1226,48 @@ def _eval_is_ge(lhs, rhs):
...
+    assert 1/i <= S(1)/3
+    i = Symbol('i', nonnegative=True)
+    e = Pow(10, 1000, evaluate=False)
+    assert (i + 1) <= Pow(i + 1, e, evaluate=False)
+    assert i >= i**-pi
+    assert i**pi >= i
+    assert i**pi >=0
+    j = Symbol('j', integer=True, nonnegative=True)
+    r = i/(i + j)
+    assert r <= 1
```
smichr, line 1273:
```suggestion
    assert r <= 1
    i, j = symbols('i j', integer=True, positive=True)
    assert 1/i +1 <= j + 1
    assert (1/i + 1)/(j + 1) <= 1

```

sympy/core/relational.py
```diff
@@ -1075,9 +1086,82 @@ def _n2(a, b):
...
+def _eval_is_ge(lhs, rhs): # noqa:F811
+    from sympy.core.function import expand_mul
+
+    if lhs.is_Atom and rhs.is_Atom:
+        return  # resolve with assumptions of lhs, rhs
+
+    # x >= x**e
+    if isinstance(rhs, Pow) and not isinstance(lhs, Pow):
+        if rhs.base == lhs and lhs.is_nonnegative and is_le(
+                R.exp, S.One):
```
smichr, line 1134:
```suggestion
                rhs.exp, S.One):
```

sympy/core/relational.py
```diff
@@ -1075,9 +1086,82 @@ def _n2(a, b):
...
+            return True
+    # x**e >= x
+    if isinstance(lhs, Pow) and not isinstance(rhs, Pow):
+        if lhs.base == rhs and R.is_nonnegative and is_ge(
```
smichr, line 1138:
```suggestion
        if lhs.base == rhs and rhs.is_nonnegative and is_ge(
```

sympy/core/relational.py
```diff
@@ -327,6 +330,14 @@ def _eval_simplify(self, **kwargs):
...
                     x = free.pop()
                     dif = r.lhs - r.rhs
+                    # r.func(dif, 0) ->
+                    # r.func(n/d, 0) ->
+                    # r.func(sign(d)*n, 0)
+                    n, d = dif.as_numer_denom()
+                    if d.is_positive:
+                        dif = n
+                    elif d.is_negative:
+                        dif = -n
```
smichr, line 340:
```suggestion
```

sympy/core/tests/test_relational.py
```diff
@@ -787,6 +787,18 @@ def test_simplify_relational():
...
+    assert (-1/p >= -1).simplify() == S.true
+    n = symbols('n', integer=True, negative=True)
+    assert (1/n >= -1).simplify() == S.true
+    assert (-1/n <= 1).simplify() == S.true
+    p = symbols('p', prime=True)
+    assert p/n <= 0
+    n = symbols('n', negative=True)
+    assert (1/n < 1/(n - 1)).simplify() == S.true
+    assert ((n + 1)/(n - 1) > 1).simplify() == S.false
+
```
smichr, line 801:
```suggestion
```

sympy/core/tests/test_relational.py
```diff
@@ -1211,3 +1226,51 @@ def _eval_is_ge(lhs, rhs):
...
+    assert (i + 1) <= Pow(i + 1, e, evaluate=False)
+    assert i >= i**-pi
+    assert i**pi >= i
+    assert i**pi >=0
+    j = Symbol('j', integer=True, nonnegative=True)
+    r = i/(i + j)
+    assert r <= 1
+    i, j = symbols('i j', integer=True, positive=True)
+    assert 1/i +1 <= j + 1
+    assert (1/i + 1)/(j + 1) <= 1
```
smichr, line 1276:
```suggestion
    assert (1/i + 1)/(j + 1) <= 1
    n = symbols('n', negative=True)
    assert 1/n < 1/(n - 1)
    assert ((n + 1)/(n - 1) > 1) == S.false
```

sympy/core/relational.py
```diff
@@ -1075,9 +1078,184 @@ def _n2(a, b):
...
+    c, L = L.as_coeff_Add()
+    R -= c
+    # multiplicative constant to R
+    if not R.is_Add:
+        c, L = L.as_coeff_Mul()
+        if c < 0:
+            ix = ~ix
+        R /= c
+    if R.is_Rational:
+        was = L,R
```
smichr, line 1239:
```suggestion
```

sympy/core/relational.py
```diff
@@ -1075,9 +1078,183 @@ def _n2(a, b):
...
+    if r is S.NegativeOne and is_ge(b, S.Zero):
+        return False  # XXX do we need to handle False cases?
+
+
 @dispatch(Expr, Expr)
-def _eval_is_ge(lhs, rhs):
-    return None
+def _eval_is_ge(lhs, rhs):  # noqa:F811
+    from sympy import Mod, fraction
+    if lhs.is_Atom and rhs.is_Atom:
```
smichr, line 1143:
```suggestion
    if lhs.is_Atom and rhs.is_Atom or (
            lhs.is_number and rhs.is_number):
```

sympy/core/relational.py
```diff
@@ -1075,9 +1078,184 @@ def _n2(a, b):
...
+                r = R
+            else:
+                den = S.Zero
+                r = S.One
+            if is_ge(d, S.Zero):
+                L, R = n/r, den
+            elif is_le(d, S.Zero):
+                ix = ~ix
+                L, R = n/r, den
+    if (L, R) != (lhs, rhs):
```
smichr, line 1257:
```suggestion
    if {L, R} != {lhs, rhs}:
```

sympy/core/relational.py
```diff
@@ -1075,9 +1078,184 @@ def _n2(a, b):
...
+                r = R
+            else:
+                den = S.Zero
+                r = S.One
+            if is_ge(d, S.Zero):
+                L, R = n/r, den
+            elif is_le(d, S.Zero):
+                ix = ~ix
+                L, R = n/r, den
+    if {L, R} != {lhs, rhs}:
```
smichr, line 1257:
```suggestion
    if {L, R} - {lhs, rhs, -lhs, -rhs}:
```

sympy/core/relational.py
```diff
@@ -1075,9 +1078,182 @@ def _n2(a, b):
...
+        return False  # XXX do we need to handle False cases?
+
+
 @dispatch(Expr, Expr)
-def _eval_is_ge(lhs, rhs):
-    return None
+def _eval_is_ge(lhs, rhs):  # noqa:F811
+    from sympy import Mod, fraction
+    if lhs.is_Atom and rhs.is_Atom or (
+            lhs.is_number and rhs.is_number):
```
smichr, line 1144:
```suggestion
            lhs.is_number and rhs.is_number or
            not (lhs.is_finite and rhs.is_finite)):
```

sympy/solvers/inequalities.py
```diff
@@ -962,7 +962,7 @@ def reduce_inequalities(inequalities, symbols=[]):
     (-3 <= x) & (x < oo)
 
     >>> reduce_inequalities(0 <= x + y*2 - 1, [x])
-    (x < oo) & (x >= 1 - 2*y)
+    x >= 1 - 2*y
```
smichr, line 965:
```suggestion
    (x < oo) & (x >= 1 - 2*y)
```

sympy/core/relational.py
```diff
@@ -1075,9 +1078,220 @@ def _n2(a, b):
...
+        L, R = R, L
+        ix = ~ix
+    m, l, b = _aslinear(L)
+    mr, r, br = _aslinear(R)
+    if l.is_Pow or r.is_Pow:
+        L = l
+        R = mr/m*r + (br - b)/m
+        if m < 0:
+            ix = ~ix
+    if R.is_Rational:
```
smichr, line 1272:
```suggestion
    if R.is_Number:
```

smichr:
@oscarbenjamin I will start to add tests specifically to target all uncovered lines but I think the general structure is ready for review. I will clean up the commit history before committing once this has been checked over and iterated to completion.

sympy/core/relational.py
```diff
@@ -1075,9 +1078,220 @@ def _n2(a, b):
...
+    elif r > 0:
+        return is_ge(S(r.p), r.q*_p)
+    if r not in (0, -1):
+        return
+    b, e = _p.as_base_exp()
+    if Mod(e, 2) == 1 and is_ge(S.Zero, b):
+        if r is S.Zero:
+            return is_ge(S.Zero, b)
+        # r == -1
+        if b is S.NegativeOne:
```
smichr, line 1167:
```suggestion
        if b is S.NegativeOne and is_ge(e, S.Zero):
```

sympy/core/relational.py
```diff
@@ -6,11 +6,14 @@
 from .basic import Basic
 from .evalf import EvalfMixin
 from .function import AppliedUndef
+from .numbers import Rational
```
smichr, line 9:
```suggestion
from .mod import Mod
from .numbers import Rational
```

sympy/core/relational.py
```diff
@@ -1075,9 +1078,220 @@ def _n2(a, b):
...
+    (1, 0, 0)
+    """
+    b, t = e.as_coeff_Add()
+    m, x = t.as_coeff_Mul()
+    return m, x, b
+
+
+@dispatch(Pow, Rational) # type: ignore
+def _eval_is_ge(_p, r):  # :F811
+    from sympy import Mod
```
smichr, line 1115:
```suggestion
```

sympy/core/relational.py
```diff
@@ -1075,9 +1078,220 @@ def _n2(a, b):
...
+        # r == 1
+        if is_ge(e, S.Zero):
+            return is_ge(S.NegativeOne, b)
+        if is_ge(S.Zero, e):
+            return _small_neg(b)
+
+
+@dispatch(Rational, Pow) # type: ignore
+def _eval_is_ge(r, _p):  # noqa:F811
+    from sympy import Mod
```
smichr, line 1153:
```suggestion
```

sympy/core/relational.py
```diff
@@ -1075,9 +1078,220 @@ def _n2(a, b):
...
+            return _small_neg(b)
+    if r is S.NegativeOne and is_ge(b, S.Zero):
+        return False
+
+
 @dispatch(Expr, Expr)
-def _eval_is_ge(lhs, rhs):
-    return None
+def _eval_is_ge(lhs, rhs):  # noqa:F811
+    from sympy import Mod, fraction
```
smichr, line 1179:
```suggestion
    from sympy.simplify.radsimp import fraction
```

sympy/core/relational.py
```diff
@@ -1075,9 +1079,218 @@ def _n2(a, b):
...
+            else:
+                den = S.Zero
+                r = S.One
+            if is_ge(d, S.Zero):
+                L, R = n/r, den
+            elif is_le(d, S.Zero):
+                ix = ~ix
+                L, R = n/r, den
+    # if we did more than change sign or switch sides, re-evaluate
+    if {L, R} - {lhs, rhs, -lhs, -rhs}:
```
smichr, line 1290: it seems to me that the return should be something like `is_ge(L,R)` if either is a power else `L >= R` if L and R are not trivially the same as at the start.

sympy/core/relational.py
```diff
@@ -1075,9 +1079,218 @@ def _n2(a, b):
...
+    """
+    b, t = e.as_coeff_Add()
+    m, x = t.as_coeff_Mul()
+    return m, x, b
+
+
+@dispatch(Pow, Rational) # type: ignore
+def _eval_is_ge(_p, r):  # :F811
+    if _p.is_number:
+        return
```
oscarbenjamin, line 1117: Is this return intended to allow the floating point evaluation to handle this case?

Part of the idea of introducing `is_ge` et al was to replace/eliminate the use of numerical evaluation in these queries so that they can be resolved symbolically and therefore be able to handle arbitrarily large numbers.

sympy/core/relational.py
```diff
@@ -1075,9 +1079,218 @@ def _n2(a, b):
...
+    if r <= -1:
+        return is_ge(S.NegativeOne, _invpow(_p))
```
oscarbenjamin, line 1119: Here we are testing `_p >= r` where `_p = a**b` so `a**b >= r`

Now if `r <= -1` we return `-1 >= a**-b`.

If `a = 2` and `b = 3` and `r = -2` then for `2**3 >= -2` (which is true) we return `-1 >= 2**-3` (which is false).

sympy/core/relational.py
smichr: I think I have tightened up the detection of infinite recursion perhaps I can relax this constraint.

sympy/core/relational.py
```diff
@@ -1075,9 +1079,267 @@ def _n2(a, b):
...
+    # if the power is indeterminant; return l if it is not a power
+    if l.is_Pow:
+        if l.base.is_zero:
+            if l.exp.is_zero:
+                l = S.One
+            elif l.exp.is_zero is False:
+                l = S.Zero
+            else:
+                return
+        elif l.base.is_zero is None:
```
smichr, line 1151:
```suggestion
        elif l.base.is_nonnegative is None:
```

sympy/core/relational.py
```diff
@@ -1075,9 +1079,281 @@ def _n2(a, b):
...
+    is_ge(x, S.NegativeOne), is_ge(S.Zero, x)))
+
+_big_ppow = lambda b, e: fuzzy_or((
+        fuzzy_and((b.is_zero, e.is_zero)),  # 1
+        fuzzy_and((Mod(e, 2) == 0, fuzzy_or((
+        is_ge(S.NegativeOne, b) and is_ge(e, S.Zero),
+        _small_neg(b) and is_ge(S.Zero, e))))),
+    _small_pos(b) and is_ge(S.Zero, e),
+    is_ge(b, S.One) and is_ge(e, S.Zero)))
+
```
oscarbenjamin, line 1098: This kind of thing is definitely better as a proper function with a docstring rather than a lambda function. If the function is internal then it does not need to have a "full" docstring but at least an explanation of what it tests. I don't know what `_big_ppow` means and it would take a long time to parse it from the code here!

sympy/core/relational.py
```diff
@@ -1075,9 +1079,281 @@ def _n2(a, b):
...
+                return True
+            if is_ge(S.Zero, R):
+                return True
+        if is_ge(L, S.NegativeOne):
+            if is_ge(S.NegativeOne, R):
+                return True
+
+    if L.is_Pow or R.is_Pow:
+        b, e = L.as_base_exp()
+        B, E = R.as_base_exp()
```
oscarbenjamin, line 1328: The idea with multipledispatch is supposed to be that we can write a `Pow/Pow` handler rather than having all the logic in an `Expr/Expr` handler that checks the types.

sympy/core/relational.py
smichr: I'm in the mid-game right now. I really hope this can be a lot simpler.

smichr: for the record, the helpers identify whether a power is positive or negative and big (>= 1) or small (between 0 and 1).

sympy/core/relational.py
```diff
@@ -1075,9 +1081,375 @@ def _n2(a, b):
...
+                if is_ge(S.Zero, r):
+                    ix = ~ix
+                elif not is_ge(r, S.Zero):
+                    return  # hopefully impossible after passing entrance test
+                L = Pow(l.base, l.exp - r.exp, evaluate=False)
+                R = mr/m
+                if m < 0:
+                    ix = ~ix
+            elif m == mr and l.exp == r.exp and (
+                    l.exp.is_integer or
```
smichr, line 1314: These conditions are too strict and are keeping the `_base_like_pow` lines unused.

sympy/core/assumptions.py
```diff
@@ -385,6 +385,14 @@ def check_assumptions(expr, against=None, **assume):
     >>> check_assumptions(2*x + 1, x)
     True
 
+    To see if a number matches the assumptions of an expression, pass
+    the number as the first argument, else it's specific assumptions
+    may not have a non-None value is the expression:
```
smichr, line 390:
```suggestion
    may not have a non-None value in the expression:
```

sympy/core/relational.py
```diff
@@ -1075,9 +1081,383 @@ def _n2(a, b):
...
+    is_ge(x, S.NegativeOne), is_ge(S.Zero, x)))
+
+# b**e >= 1
+_big_ppow = lambda b, e: fuzzy_or((
+        fuzzy_and((b.is_zero, e.is_zero)),  # 1
+        fuzzy_and((Mod(e, 2) == 0, fuzzy_or((
+        is_ge(S.NegativeOne, b) and is_ge(e, S.Zero),
+        _small_neg(b) and is_ge(S.Zero, e))))),
+    _small_pos(b) and is_ge(S.Zero, e),
+    is_ge(b, S.One) and is_ge(e, S.Zero)))
```
oscarbenjamin, line 1117: I still think these should be proper functions with docstrings and tests.

oscarbenjamin: They don't need to be "public" but the meaning should be clearer in the code.

sympy/core/relational.py
```diff
@@ -1075,9 +1081,383 @@ def _n2(a, b):
...
+        ix = ~ix
+    if SYMPY_DEBUG:
+        print(lhs, rhs, '-trivial->', l, r)
+        if pause: input('press return to continue')
+
+    # work really hard to identify a power for comparison
+    if l.is_Pow and r == l.base:
+        r = Pow(r, 1, evaluate=False)
+    elif r.is_Pow and l == r.base:
+        l = Pow(l, 1, evaluate=False)
```
oscarbenjamin, line 1298: The idea behind introducing multiple dispatch here was supposed to be that we wouldn't just put everything in an `Expr,Expr` handler.

sympy/core/relational.py
smichr: Yes -- once this is pared down as much as possible I will add final touches.

sympy/core/relational.py
smichr: Are we supposed to create a bunch of permutations of the types:

Add, Add
Add, Pow
Pow, Add
Pow, Pow
Add, Mul
Mul, Add
Pow, Mul
Mul, Pow

oscarbenjamin: That was the intention. Maybe it wasn't the best design if that can't be made to work.

This was intended to be more or less a part of the old assumptions and the expectation was not that it would resolve all cases. Instead only "easy" cases would be resolved. The new assumptions are a more natural place to have a slower but more complete implementation for resolving inequalities like this.

smichr: I'll give it a try to put it into the handlers. The good news is that already tests are running significantly faster with even these changes in place.

PR #21438, opened 2021-05-07
Implemented applyfunc and applyfunc_nonzero

ks147:
Implemented applyfunc and applyfunc_nonzero for DomainMatrix

#### Release Notes

* polys
  * Implemented functions applyfunc and applyfunc_nonzero in DomainMatrix class

### DIFF
--- a/sympy/polys/matrices/ddm.py
+++ b/sympy/polys/matrices/ddm.py
@@ -203,6 +203,30 @@ class DDM(list):
             msg = "Shape mismatch: %s %s %s" % (a.shape, op, b.shape)
             raise DDMShapeError(msg)
 
+    def applyfunc(self, f, domain=None):
+        """
+
+        Examples
+        ========
+
+        >>> from sympy.polys.matrices.ddm import DDM
+        >>> from sympy import ZZ, QQ
+        >>> A = DDM([[ZZ(1), ZZ(2)], [ZZ(3), ZZ(4)]], (2, 2), ZZ)
+        >>> A.applyfunc(lambda x: 2 ** x)
+        [[2, 4], [8, 16]]
+
+        We can also specify the :py:class:`~.Domain` of the output
+
+        >>> A.applyfunc(lambda x: x/5, QQ)
+        [[1/5, 2/5], [3/5, 4/5]]
+
+        """
+        if domain is None:
+            domain = self.domain
+
+        Af = ddm_applyfunc(self, f, domain)
+        return DDM(Af, self.shape, domain)
+
     def add(a, b):
         """a + b"""
         a._check(a, '+', b, a.shape, b.shape)
@@ -331,3 +355,6 @@ class DDM(list):
 
 
 from .sdm import SDM
+
+def ddm_applyfunc(M, f, domain):
+    return [[domain.convert(f(x)) for x in Mi] for Mi in M]
--- a/sympy/polys/matrices/ddm.py
+++ b/sympy/polys/matrices/ddm.py
@@ -75,6 +75,8 @@ from .dense import (
         ddm_ilu_split,
         ddm_ilu_solve,
         ddm_berk,
+        ddm_applyfunc,
+        ddm_applyfunc_nonzero,
         )
 
 
@@ -380,8 +382,3 @@ class DDM(list):
 
 from .sdm import SDM
 
-def ddm_applyfunc(M, f):
-    return [[f(x) for x in Mi] for Mi in M]
-
-def ddm_applyfunc_nonzero(M, f, domain):
-    return [[f(x) if not domain.is_zero(x) else domain.zero for x in Mi] for Mi in M]
--- a/sympy/polys/matrices/ddm.py
+++ b/sympy/polys/matrices/ddm.py
@@ -381,4 +381,3 @@ class DDM(list):
 
 
 from .sdm import SDM
-
--- a/sympy/polys/matrices/ddm.py
+++ b/sympy/polys/matrices/ddm.py
@@ -204,7 +204,8 @@ class DDM(list):
             raise DDMShapeError(msg)
 
     def applyfunc(self, f, domain=None):
-        """
+        """Applies a function f to every element of a DDM
+        and converts :py:class:`~.Domain` to domain
 
         Examples
         ========
@@ -224,8 +225,31 @@ class DDM(list):
         if domain is None:
             domain = self.domain
 
-        Af = ddm_applyfunc(self, f, domain)
-        return DDM(Af, self.shape, domain)
+        Mf = ddm_applyfunc(self, f, domain)
+        return DDM(Mf, self.shape, domain)
+
+    def applyfunc_nonzero(self, f, domain=None):
+        """Applies a function f to every non-zero element of a DomainMatrix
+        and converts :py:class:`~.Domain` to domain
+
+        Examples
+        ========
+
+        >>> from sympy.polys.matrices.ddm import DDM
+        >>> from sympy import ZZ, QQ
+        >>> A = DDM([[ZZ(1), ZZ(0)], [ZZ(0), ZZ(3)]], (2, 2), ZZ)
+        >>> A.applyfunc_nonzero(lambda x: 2 ** x)
+        [[2, 0], [0, 8]]
+        >>> A.applyfunc_nonzero(lambda x: 1/x, QQ)
+        [[1, 0], [0, 1/3]]
+
+        """
+
+        if domain is None:
+            domain = self.domain
+
+        Mf = ddm_applyfunc_nonzero(self, f, domain)
+        return DDM(Mf, self.shape, domain)
 
     def add(a, b):
         """a + b"""
@@ -358,3 +382,6 @@ from .sdm import SDM
 
 def ddm_applyfunc(M, f, domain):
     return [[domain.convert(f(x)) for x in Mi] for Mi in M]
+
+def ddm_applyfunc_nonzero(M, f, domain):
+    return [[domain.convert(f(x)) if not domain.is_zero(x) else domain.zero for x in Mi] for Mi in M]
--- a/sympy/polys/matrices/ddm.py
+++ b/sympy/polys/matrices/ddm.py
@@ -218,14 +218,14 @@ class DDM(list):
 
         We can also specify the :py:class:`~.Domain` of the output
 
-        >>> A.applyfunc(lambda x: x/5, QQ)
+        >>> A.applyfunc(lambda x: x/QQ(5), QQ)
         [[1/5, 2/5], [3/5, 4/5]]
 
         """
         if domain is None:
             domain = self.domain
 
-        Mf = ddm_applyfunc(self, f, domain)
+        Mf = ddm_applyfunc(self, f)
         return DDM(Mf, self.shape, domain)
 
     def applyfunc_nonzero(self, f, domain=None):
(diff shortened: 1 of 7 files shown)


sympy/polys/matrices/domainmatrix.py
```diff
@@ -645,6 +645,77 @@ def _check(a, op, b, ashape, bshape):
...
+        ========
+
+        DomainMatrix.applyfunc
+        """
+
+        if self.rep.fmt != 'sparse':
+            raise TypeError("DomainMatrix should be sparse")
+
+        if not callable(f):
+            raise TypeError("`f` must be callable")
```
oscarbenjamin, line 712: Neither of these checks should be here. It should work for dense matrices and also we don't need to check whether it's callable: if it isn't then TypeError will be raised when trying to call it later.

sympy/polys/matrices/sdm.py
```diff
@@ -970,3 +1012,35 @@ def sdm_particular_from_rref(A, ncols, pivots):
...
+    Mf = {}
+    for i in range(nrows):
+        Mf[i] = {}
+        for j in range(ncols):
+            try:
+                domain_elem = domain.convert(f(M[i][j]))
+                if not domain.is_zero(domain_elem):
+                    Mf[i][j] = domain_elem
+            except KeyError:
+                domain_elem = domain.convert(f(domain.zero))
```
oscarbenjamin, line 1028: Here `f(domain.zero)` should only be called once. If it gives zero then `sdm_applyfunc_nonzero` can be used. Otherwise `sdm_applyfunc` should effeciently initialise a matrix full of the result and then only loop over the nonzero elements after.

sympy/polys/matrices/sdm.py
```diff
@@ -970,3 +1012,35 @@ def sdm_particular_from_rref(A, ncols, pivots):
...
+
+
+def sdm_applyfunc_nonzero(M, f, domain):
+    Mf = {}
+    for i, Mi in M.items():
+        for j, Mij in Mi.items():
+            try:
+                Mf[j][i] = domain.convert(f(Mij))
+            except KeyError:
+                Mf[j] = {i: domain.convert(f(Mij))}
```
oscarbenjamin, line 1044: Why would KeyError ever be raised?

This is just:
```python
Mf = {i: {j: f(Mij) for j, Mij in Mi.items()} for i, Mi in M.items()}
```
Possibly it can be faster (if `f` is a fast function) to use `zip` and `map` like:
```python
Mf = {i: dict(zip(Mi, map(f, Mi.values))) for i, Mi in M.items()}
```

sympy/polys/matrices/ddm.py
```diff
@@ -331,3 +379,9 @@ def charpoly(a):
 
 
 from .sdm import SDM
+
+def ddm_applyfunc(M, f, domain):
+    return [[domain.convert(f(x)) for x in Mi] for Mi in M]
```
oscarbenjamin, line 384: It is generally more efficient to use `convert_from` rather than `convert`. Also mostly the old and new domains will be the same in which case we can avoid converting altogether.

Actually though this is misunderstanding the purpose of the domain argument which is that it should specify the domain in the case that `f` will return elements of a different domain. The point is not that applyfunc should call convert but that in some cases `f` will already have converted the elements to another domain in which case the caller of `pplyfunc` needs to tell `DomainMatrix` what that other domain is.

ks147: So you want me to use `convert` or `convert_from` only when the output `domain` is specified?

oscarbenjamin: There is no need to use convert at all. The domain argument is needed if the function `f` already returns elements of a different domain.

ks147: @oscarbenjamin what do you think about this?

```python
def sdm_applyfunc_nonzero(M, f, domain=None):
    Mf = {}
    for i, Mi in M.items():
        for j, Mij in Mi.items():
            fMij = f(M[i][j]) if domain is None else domain.convert(f(M[i][j]))
        
            if not domain.is_zero(fMij):
                try:
                    Mf[i][j] = fMij
                except KeyError:
                    Mf[i] = {j: fMij}

    return Mf
```

Do you think I should also pass current domain to functions `sdm_applyfunc` and `sdm_applyfunc_nonzero` and use `convert_from` instead of `convert` ?

oscarbenjamin: As I already said: there is no need to use convert at all.

sympy/polys/matrices/sdm.py
```diff
@@ -970,3 +1012,34 @@ def sdm_particular_from_rref(A, ncols, pivots):
...
+        for j, Mij in Mi.items():
+                fMij = f(M[i][j])
+                if not domain.is_zero(fMij):
+                    Mf[i][j] = fMij
+                else:
+                    del Mf[i][j]
+    return Mf
+
+
+def sdm_applyfunc_nonzero(M, f, domain):
```
oscarbenjamin, line 1035: There is already a function `unop_dict` that does this and can be used instead.

oscarbenjamin: Perhaps it could be renamed.

sympy/polys/matrices/ddm.py
```diff
@@ -331,3 +379,9 @@ def charpoly(a):
 
 
 from .sdm import SDM
+
+def ddm_applyfunc(M, f):
+    return [[f(x) for x in Mi] for Mi in M]
+
+def ddm_applyfunc_nonzero(M, f, domain):
+    return [[f(x) if not domain.is_zero(x) else domain.zero for x in Mi] for Mi in M]
```
oscarbenjamin, line 387: These should go in `dense.py`.

ks147: ok

sympy/polys/matrices/ddm.py
```diff
@@ -203,6 +203,54 @@ def _check(cls, a, op, b, ashape, bshape):
...
+        >>> from sympy.polys.matrices.ddm import DDM
+        >>> from sympy import ZZ, QQ
+        >>> A = DDM([[ZZ(1), ZZ(2)], [ZZ(3), ZZ(4)]], (2, 2), ZZ)
+        >>> A.applyfunc(lambda x: 2 ** x)
+        [[2, 4], [8, 16]]
+
+        We can also specify the :py:class:`~.Domain` of the output
+
+        >>> A.applyfunc(lambda x: x/QQ(5), QQ)
+        [[1/5, 2/5], [3/5, 4/5]]
```
oscarbenjamin, line 222: Rather than saying that "and converts to a domain" or "we can also specify" this should say: `If the function returns elements from a different domain then the new domain should be given as the *domain* argument.`

sympy/polys/matrices/sdm.py
```diff
@@ -970,3 +1012,34 @@ def sdm_particular_from_rref(A, ncols, pivots):
...
+    fzero = (f(domain.zero))
+    Mf = {i: {j: fzero for j in range(ncols)} for i in range(nrows)}
+
+    for i, Mi in M.items():
+        for j, Mij in Mi.items():
+                fMij = f(M[i][j])
+                if not domain.is_zero(fMij):
+                    Mf[i][j] = fMij
+                else:
+                    del Mf[i][j]
```
oscarbenjamin, line 1031: You already have `Mij` so you don't need to look it up as `M[i][j]`. We can just use `if fMij` rather than `domain.is_zero`. That's what all of the other sparse routines do. It's probably worth doing `Mfi = Mf[i]` at the start of the outer loop. Also the indentation is wrong:
```
    for i, Mi in M.items():
        Mfi = Mf[i]
        for j, Mij in Mi.items():
            fMij = f(Mij)
            if fMij:
                Mfi[j] = fMij
            else:
                del Mfi[j]
```

sympy/polys/matrices/tests/test_sdm.py
```diff
@@ -274,3 +274,12 @@ def test_SDM_particular():
...
     assert A.particular() == Apart
+
+def test_SDM_applyfunc():
+    A = SDM.eye(4, ZZ)
+    assert A.applyfunc(lambda x: ZZ(2) * x) == ZZ(2) * SDM.eye(4, ZZ)
+    assert A.applyfunc(lambda x: ZZ(0)) == SDM.zeros((4, 4), ZZ)
+
+def test_SDM_applyfunc_nonzero():
+    A = SDM.eye(3, ZZ)
+    assert A.applyfunc_nonzero(lambda x: x + QQ(1, 2), QQ) == QQ(3, 2) * SDM.eye(3, QQ)
```
oscarbenjamin, line 285: These tests should also include examples where nonzero elements become zero and vice versa

sympy/polys/matrices/sdm.py
```diff
@@ -680,6 +680,49 @@ def charpoly(A):
...
+
+        """
+        if domain is None:
+            domain = self.domain
+
+        Mf = sdm_applyfunc(self, self.shape, f, domain)
+        return SDM(Mf, self.shape, domain)
+
+    def applyfunc_nonzero(self, f, domain=None):
+        """Returns SDM after applying function f to each non-zero element
```
oscarbenjamin, line 709: This docstring does not explain the purpose of this which is that it is efficient for sparse matrices.

sympy/polys/matrices/sdm.py
ks147: Yeah maybe, I didn't know it was supposed to be doing the exact same thing as `sdm_applyfunc_nonzero`

ks147:
@oscarbenjamin, please review

sympy/polys/matrices/sdm.py
```diff
@@ -970,3 +1017,22 @@ def sdm_particular_from_rref(A, ncols, pivots):
...
             P[j] = Ain / A[i][j]
     return P
+
+
+def sdm_applyfunc(M, shape, f, domain):
+    if not f(domain.zero):
+        return unop_dict(M, f)
+
+    nrows, ncols = shape
+    fzero = f(domain.zero)
```
oscarbenjamin, line 1027: `f(domain.zero)` is computed above. The result should be reused rather than calling `f` again.

sympy/polys/matrices/sdm.py
```diff
@@ -970,3 +1017,22 @@ def sdm_particular_from_rref(A, ncols, pivots):
...
+    Mf = {i: {j: fzero for j in range(ncols)} for i in range(nrows)}
+
+    for i, Mi in M.items():
+        for j, Mij in Mi.items():
+            fMij = f(Mij)
+            if fMij:
+                Mf[i][j] = fMij
+            else:
+                del Mf[i][j]
```
oscarbenjamin, line 1036: This needs to check if row `i` becomes all zero in which case it should be removed. A test should be added for this case.

sympy/polys/matrices/sdm.py
```diff
@@ -970,3 +1017,27 @@ def sdm_particular_from_rref(A, ncols, pivots):
...
+        num_zero_elems = 0
+        for j, Mij in Mi.items():
+            fMij = f(Mij)
+            if fMij:
+                Mf[i][j] = fMij
+            else:
+                del Mf[i][j]
+                num_zero_elems += 1
+        if num_zero_elems == ncols:
+            del Mf[i]
```
oscarbenjamin, line 1042: It would be better to do `Mfi = Mf[i]` once and then use `Mfi` for everything. We don't need `num_zero_elems` because we can just do `if Mfi`.

ks147: ok

sympy/polys/matrices/sdm.py
```diff
@@ -970,3 +1017,29 @@ def sdm_particular_from_rref(A, ncols, pivots):
...
+        Mfi = Mf[i]
+        for j, Mij in Mi.items():
+            fMij = f(Mij)
+            if fMij:
+                Mfi[j] = fMij
+            else:
+                del Mfi[j]
+
+        if Mfi:
+           Mf[i] = Mfi
```
oscarbenjamin, line 1042: There is no need to do `Mf[i] = Mfi`

PR #21439, opened 2021-05-07
Added details to picture allowed-sign-conventions.png

nopria:

This pull request does not address any opened issue.

Added some useful details to picture shown in documentation page https://docs.sympy.org/latest/modules/physics/continuum_mechanics/beam.html

#### Release Notes

NO ENTRY


smichr:
Maybe you could past a picture in the OP so reviewers can see it without having to download your branch.

nopria:
Sorry, I'm a beginner with GitHub, here is the new picture, it is the same picture as before (which also was posted by me) with just a few more details. The reason I added such details is that, looking at documentation after a few months, the current picture seemed not clear enough to me.

allowed-sign-conventions

Below is the current picture.

https://github.com/sympy/sympy/blob/master/doc/src/modules/physics/continuum_mechanics/allowed-sign-conventions.png

smichr:
Ok, I read the page and I don't find either the current or proposed image helpful. I don't know if I am looking at the end of a beam or a "square beam" from the side. If it is a side view, having a rectangle would be useful.

I am not sure what the orange letters are supposed to show -- position of axes after indicated rotation? But the bottom one seems to move opposite of the arrow shown. There must be a better way to show a compound rotation.

nopria:
As stated at the beginning of the documentation "this module can be used to solve 2D beam", so there no doubt that you are looking at the beam from the side. The "square" on the left represents, as usual in civil engineering, an elementary beam segment, and all the arrows represent (positive) directions of axes (forces) and rotation (moment). The letters are possible choices for axes names.
I agree that detached from its context the picture can be interpreted in many ways, but since it relates to the note, and the note talks about sign conventions for shear and moment, in such context should be clear enough that the curved arrows means just the sign convention for rotation and not that the axis is rotating (why should it?).
I think that the proposed image is better than the current one, but of course there is always room for improvements. In the meantime, however, a little improvement is better than nothing.
@smichr

smichr:
> a little improvement is better than nothing.

I agree. What do the curved arrows on the axes represent?

nopria:
> I agree. What do the curved arrows on the axes represent?

Positive rotations and moments. Since we are in 2D the the only rotation is around the axis normal to the 2D plane (represented by the oblique axis).

smichr:
What do the orange background letters represent?

In the top axes does "x" go into the plane and in the bottom axes does it come out of the plane?

nopria:
> What do the orange background letters represent?

The orange background letters represent a possible naming of the axes alternative to the white background letters.

> In the top axes does "x" go into the plane and in the bottom axes does it come out of the plane?

Exactly, the oblique axes are normal to the 2D plane, going into (top) or coming out (bottom) of the plane.

I think it's better to add a caption to the image (is it possible?), which could be:

> At the top and bottom are shown two allowed coordinate systems for the 2D beam. For each coordinate system, the square on the left represents the positives axial force, shear force and bending moment on an elementary beam segment. On the right is represented the corresponding coordinate system with two possible choices for names of the axes (letters with either white or orange background), with the curved arrows showing the positive rotation/moment.

Also, I agree that the curved arrows on the right can be somewhat misleading and are not necessary, they can be erased.

smichr:
What about something like this:

image

The diagram below shows the sense of the sign convention that must be selected when indicating stress, shear and rotation on a location in the beam. If the positive sign convention is selected then positive stress, shear and rotation act in the directions shown in black and vice versa if the negative sign convention is selected. The letters on white or orange background are possible selections for the axes names.

nopria:
This is not the kind of diagram a civil engineer is used to. Here are some examples of elementary beam segment:

https://mechanicalc.com/static/img/Beam/Theory/Original/sign-convention-01.png

https://th.bing.com/th/id/Rb68ce640c1fa2b8a772118fe772ccb2f?rik=wLrr3rSm2avklA&riu=http%3a%2f%2fmechref.engr.illinois.edu%2fsol%2ffaceSignConvention.PNG&ehk=wJ9fXc3Bq%2bJecFg%2bfHlbhnWBQIQ5xA0NfnIJQvPBLeY%3d&risl=&pid=ImgRaw

https://upload.wikimedia.org/wikipedia/commons/thumb/f/f4/Timoshenko_beam_sign_convention.svg/1094px-Timoshenko_beam_sign_convention.svg.png

https://www.researchgate.net/profile/Ehsan_Zafari/publication/262020259/figure/fig2/AS:668919715094530@1536494296167/Positive-sign-convention-of-shear-and-bending-forces-in-beam-element.png

In particular, the schematic representation of an elementary beam segment should be separated from the representation of the coordinate system, because in the elementary beam segment the direction of positive forces changes depending on the side of the segment you are considering, so it must be clear which side of the elementary beam segment a vector is referred to.

However from your image I like the curved arrows representing the rotations, those are much clearer than in mine image.

nopria:
I updated the image with smaller curved arrows

https://github.com/nopria/sympy/blob/master/doc/src/modules/physics/continuum_mechanics/allowed-sign-conventions.png

in my forked sympy and this automatically created a new pull request, would you please tell me if it's ok or I have to do some more steps? Thanks.

smichr:
Is it possible to make the rotation arrow symmetric about the line instead of simply being symmetric on a vertical midline?

nopria:
I don't understand what you mean.

smichr:
> What you mean exactly?

The head of the rotation arrow looks like segments from two circles that are stacked vertically or horizontally. This leads to one of the curved head lines to be much further from the arrow shaft than the other. In a rough way, can you move one of the curved lines to look more like the green one below?

image

nopria:
Yes, I saw that, but I don't know a simple way to fix it. I used LaTeX with tikz package, the command is `\draw [line width=1,->] (0,0) arc (<angle>:<angle>:<radius>);`, if you know how to replace this let mw know.

oscargus:
This may provide some help to improve the arrow head direction: https://tex.stackexchange.com/questions/109689/arrowhead-rotation

nopria:
I finally fixed the head of arrows as required.

PR #19567, opened 2020-06-16
Make PolyMatrix arithmetics rigorous

sylee957:
#### Brief description of what is fixed or changed

1. PolyMatrix sanitizes string types of `ring` specification like `ring='ZZ'` to 

- MatrixShaping methods:
  - [x] `row_del`, `col_del` works and correctly inherits `ring`.
  - [x] `row_insert`, `col_insert` works and correctly unifies `ring`.
  - [x] `row_join`, `col_join` works and correctly unifies `ring`.
  - [x] `vec` works and correctly inherits `ring`.
  - [x] `row`, `col` works and correctly inherits `ring`.
  - [x] `extract` works and correctly inherits `ring`.
  - [x] `get_diag_blocks` works and correctly inherits `ring`.
  - [x] `hstack, vstack` works and correctly unifies `ring`.
  - [x] `reshape` works and correctly inherits `ring`.
  - [x] `diagonal` works and correctly inherits `ring`.
  - [ ] `vech`, `diag` works, but does not inherit the ring from given parameters.
  - [x] `tolist`, `todok` works without modification
- MatrixSpecial methods:
  - [x] `jordan_block` works and correctly inherits `ring` from Poly element given.
  - [x] `companion` works and correctly inherits `ring` from Poly element given.
  - [ ] `eye`, `ones`, `zeros` works, but ones and zeros are `S.One` and `S.Zero` than zeros corresponding to any `Poly`.
- MatrixProperties Methods
  - [x] `atoms`, `free_symbols`, `has`, `is_symbolic` works without modification.
  - [x] `is_square` works without modification.
  - [x] `is_symmetric`, `is_anti_symmetric` works.
  - [ ] `is_hermitian` is not implemented until #19531 is resolved.
  - [x] `is_diagonal` and `is_Identity` works.
  - [ ] `is_strongly_diagonally_dominant` and `is_weakly_diagonally_dominant` are not implemented because `Abs` does not work for `Poly`.
  - [x] `is_upper`, `is_lower`, `is_upper_hessenberg`, `is_lower_hessenberg` works.
  - [x] `is_zero_matrix` works.
- MatrixOperation Methods
  - [ ] `adjoint`, `conjugate` is not implemented until #19531 is resolved.
  - [ ] `as_real_imag`, `evalf`, `expand` will not be implemented because `Poly` does not have them.
  - [x] `transpose` works and correctly inherits `ring`.
  - [x] `rot90` works and correctly inherits `ring`.
[...]

PR #21441, opened 2021-05-08
fix(polys): make PolyMatrix use DomainMatrix

oscarbenjamin:

Built on top of #21427

    This commit makes PolyMatrix a wrapper around DomainMatrix rather than a
    subclass of Matrix. This is done so that Matrix can enforce the expected
    invariant that its elements should of type Expr. Ideally the Risch code
    would use DomainMatrix directly but it is not easy to figure out what
    the domain should be in all cases because the Risch code does not keep
    track of that.

#### Release Notes

* polys
    * The (internal) PolyMatrix class has been changed and now requires generators to be provided on construction.

### DIFF
--- a/sympy/polys/polymatrix.py
+++ b/sympy/polys/polymatrix.py
@@ -1,10 +1,16 @@
-from sympy.matrices.dense import MutableDenseMatrix
+from sympy.core.expr import Expr
+from sympy.core.symbol import Dummy
+from sympy.core.sympify import _sympify
 
-from sympy.polys.polytools import Poly
-from sympy.polys.domains import EX
+from sympy.polys.polyerrors import CoercionFailed
+from sympy.polys.polytools import Poly, parallel_poly_from_expr
+from sympy.polys.domains import QQ
+
+from sympy.polys.matrices import DomainMatrix
+from sympy.polys.matrices.domainscalar import DomainScalar
 
 
-class MutablePolyDenseMatrix(MutableDenseMatrix):
+class MutablePolyDenseMatrix:
     """
     A mutable matrix of objects from poly module or to operate with them.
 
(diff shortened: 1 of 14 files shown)


sympy/polys/polymatrix.py
```diff
@@ -1,87 +1,229 @@
...
+                rows, cols = len(elements), 1
+                items = elements
+        elif [type(a) for a in args[:3]] == [int, int, list]:
+            rows, cols, items, args = args[0], args[1], args[2], args[3:]
+        elif [type(a) for a in args[:3]] == [int, int, type(lambda: 0)]:
+            rows, cols, func, args = args[0], args[1], args[2], args[3:]
+            items = [func(i, j) for i in range(rows) for j in range(cols)]
+        else:
+            1/0
+        return cls.from_list(rows, cols, items, *args)
```
jksuom, line 62: What if there are several args? Should this be just `args`?

sympy/polys/polymatrix.py
```diff
@@ -1,87 +1,229 @@
...
+            p = polys[0]
+            for p2 in polys[1:]:
+                p, _ = p.unify(p2)
+        else:
+            p = Poly(0, gens)
+        if gens is not None:
+            if not isinstance(gens, tuple):
+                gens = (gens,)
+            assert gens == p.gens
+        domain = p.domain.get_field()
```
jksuom, line 83: This is good for `gruntz` where `gens` are given, but if the items are Poly elements and there are no gens, then this is not what the user expects.

sympy/polys/polymatrix.py
oscarbenjamin: I guess if someone does
```
PolyMatrix(items, x, y)
```
then it would be `*args`. For Risch it seems that the Polys are always univariate.

I'm not really sure what sort of constructor options and syntax should be allowed. The main change here is the need to provide a generator if the inputs are not given as Poly.

sympy/polys/polymatrix.py
oscarbenjamin: I'm not sure I understand what you mean. A `Poly` always has gens.

Perhaps the domain should not always be converted to a field though.

For Risch it seems that all polys are over fields so you have something like `QQ(x)[t]`. PolyMatrix (on master) doesn't really distinguish between rational function fields of the ground domain and polynomials so operations like rref use division and implicitly convert from Poly to Expr. Here by ensuring that the domain is always a polynomial ring over a field we know that rref is always possible if all elements are ground which does seem to be the case whenever rref is used in Risch.

Maybe that only needs to be checked in the rref and nullspace functions though and we can leave the ground domain as it is everywhere else. Also the PolyMatrix tests use a `ring=` argument that is so far removed here and doesn't seem to be used anywhere outside of the PolyMatrix tests. Perhaps that should be added back.

oscarbenjamin:
To be clear this is not really a good/efficient implementation of PolyMatrix. It is just an attempt to make something that is usable in Risch without too many modifications and that is built over DomainMatrix rather than being a subclass of Matrix.

I'm not sure what a good approach for PolyMatrix is in general but this PR is really about disentangling PolyMatrix and Matrix and making sure that all matrix calculations in Risch take place over a well defined domain. The main benefit is that I think this is the final step before being able to enforce that the elements in a Matrix are instances of Expr.

Further improvements to PolyMatrix can be made once it is separated from Matrix but that separation is the only part that I want to address in this PR. @sylee957 had some perhaps better ideas for improving PolyMatrix in #19567 but the implementation there makes the classes really entangled whereas this PR disentangles them.

sympy/polys/polymatrix.py
jksuom: I was thinking of the default case `gens=None`. That would probably occur when PolyMatrix is used as a replacement for Matrix: No `gens` are given on the argument list.

oscarbenjamin: If this is called with Poly elements and no generators are given then the Poly are unified and those gens are taken as the gens of the PolyMatrix. There are several different cases:
```python
In [1]: from sympy.polys.polymatrix import PolyMatrix

In [2]: PolyMatrix([[Poly(x, x), Poly(y, y)]])  # gens from the Polys
Out[2]: PolyMatrix([[x, y]], (x, y))

In [3]: PolyMatrix([[x, y]], (x, y))           # gens given explicitly
Out[3]: PolyMatrix([[x, y]], (x, y))

In [4]: PolyMatrix([[x, y]])             # gens inferred from symbols
Out[4]: PolyMatrix([[x, y]], (x, y))

In [5]: PolyMatrix([[1, 2]])             # error just like Poly(1)
GeneratorsNeeded 
```
Which of those are you referring to?

jksuom: The first case. It looks like the `gens` parameter would be `None`. But maybe that does not matter provided there is at least one Poly element with gens.

jksuom:
Would it be possible to make PolyMatrix behave more like a Matrix in a case like this:
```
In [18]: M = Matrix([Poly(x - 1)])                                              

In [19]: type(M[0, 0])                                                          
Out[19]: sympy.polys.polytools.Poly

In [20]: N = PolyMatrix([Poly(x - 1)])                                          

In [21]: type(N[0, 0])                                                          
Out[21]: sympy.core.add.Add
```
The old PolyMatrix would preserve Poly elements.

oscarbenjamin:
> The old PolyMatrix would preserve Poly elements.

That was an unintentional change. I've fixed it now:
```python
In [1]: from sympy.polys.polymatrix import PolyMatrix

In [2]: M = PolyMatrix([[Poly(1, x), Poly(x, x)]])

In [3]: M
Out[3]: PolyMatrix([[1, x]], (x,))

In [4]: M[0,0]
Out[4]: Poly(1, x, domain='QQ')

In [5]: M[0]
Out[5]: Poly(1, x, domain='QQ')

In [6]: M[:2]
Out[6]: [Poly(1, x, domain='QQ'), Poly(x, x, domain='QQ')]
```

jksuom:
It might not be too hard to make `risch` use `DomainMatrix`. The list of fields `K` or domains `K[t]` could be added to DE in the same way as `T`, `D`, `cases`... Then `PolyMatrix` could be left as is for now.

oscarbenjamin:
> It might not be too hard to make `risch` use `DomainMatrix`. The list of fields `K` or domains `K[t]` could be added to DE in the same way as `T`, `D`, `cases`...

I think that might be the best approach for `risch` in the long term. It is harder to work with domain elements than with Poly though so it would be difficult to integrate that right now. Already in many places `Poly` is converted to `Expr` and we would also need to have many other places that convert domain elements to `Expr` or `Poly` for different operations.

So far this PR helps with that because the Risch code is now forced to be explicit about the generators for PolyMatrix and some places have a better separation between Expr and Poly. It would be better though if the risch code was explicit about the domain used at each level as you say. Then the domain could be used explicitly rather letting Poly compute the domain from Expr.

> Then `PolyMatrix` could be left as is for now.

We can't leave `PolyMatrix` as it is because even if it wan't used it would still be a messy `Matrix` subclass that would need to be removed/changed so that `Matrix` can have only `Expr` as its elements. Although it isn't enforced a basic internal invariant of Matrix is that the elements are `Expr`. What `PolyMatrix` has in common with `RawMatrix` and `NewMatrix` is that it subclasses `Matrix` in order to break that internal invariant. That's not good subclassing and it constrains improvements to the more important `Matrix` class.

So far this PR does make significant improvements to `PolyMatrix` because:

1. The elements are always Poly rather than a mix of Poly/Expr and all have the same domain and generators.
2. The internal representation is `DomainMatrix` which is more efficient.
3. Algorithms like rref can use well-defined domains.

The disadvantage of this PR is that many of the other `Matrix` methods are removed from `PolyMatrix`. From here though it would be easy to add those based on the methods that DomainMatrix has. Also although `DomainMatrix` is more efficient than `Matrix/PolyMatrix` there are a lot of inefficient conversions between domain elements, Poly and Expr that could be improved.

jksuom:
> The elements are always Poly

This is good. Yesterday, I got
```
In [20]: N = PolyMatrix([Poly(x - 1)])                                          

In [21]: type(N[0, 0])                                                          
Out[21]: sympy.core.add.Add
```
but it seems to have changed.

jksuom:
This looks pretty good to me now. It is possible that someone is using a method that has been removed but it can then be added if desired.

oscarbenjamin:
I'm going to clean it up a bit and add tests.

sylee957:
I’m not sure if this is finished yet, but at least the commit history can be cleaned up

oscarbenjamin:
I've cleaned up the commit history. Is this good to go?

PR #21442, opened 2021-05-08
Adds _minpoly_tan for minimal_polynomial

### LINKED ISSUE #21430: minpoly raises 'NotAlgebraic' for tan(13*pi/45)
This issue was encountered when we tried to construct  a `DomainMatrix` from a matrix setting `extension=True`:
Try creating DomainMatrix from  `M = Matrix([[1, -1, 8], [-sqrt(3), tan(13*pi/45), 0]])` as `DOM = DomainMatrix.from_Matrix(M, extension=True)`

Since `tan(13*pi/45)` is an algebraic element(`13/45` is rational), `minpoly` shouldn't be raising an error, rather returning a Poly/ Expr.
The error in terminal:
```py
In [1]: minpoly(tan(13*pi/45))
NotAlgebraic                              Traceback (most recent call last)
<ipython-input-1-c699111c7c38> in <module>
----> 1 minpoly(tan(13*pi/45))

~\sympy\sympy\polys\numberfields.py in minimal_polynomial(ex, x, compose, polys, domain)
    670
    671     if compose:
--> 672         result = _minpoly_compose(ex, x, domain)
    673         result = result.primitive()[1]
    674         c = result.coeff(x**degree(result, x))

~\sympy\sympy\polys\numberfields.py in _minpoly_compose(ex, x, dom)
    584         res = _minpoly_rootof(ex, x)
    585     else:
--> 586         raise NotAlgebraic("%s doesn't seem to be an algebraic element" % ex)
    587     return res
    588

NotAlgebraic: tan(13*pi/45) doesn't seem to be an algebraic element
```
On looking closely, it is found that this is caused as there is no computational method for `tan` in `_minpoly_compose` as there are for `sin` and `cos`. Refer [here](https://github.com/sympy/sympy/blob/master/sympy/polys/numberfields.py#L590-L593)

So, on adding condition for `tan`, 
```py
diff --git a/sympy/polys/numberfields.py b/sympy/polys/numberfields.py
index 79f60d3e89..52e28b191c 100644
--- a/sympy/polys/numberfields.py
+++ b/sympy/polys/numberfields.py
@@ -11,7 +11,7 @@
 from sympy.core.exprtools import Factors
 from sympy.core.function import _mexpand
 from sympy.functions.elementary.exponential import exp
-from sympy.functions.elementary.trigonometric import cos, sin
+from sympy.functions.elementary.trigonometric import cos, sin, tan
 from sympy.ntheory import sieve
 from sympy.ntheory.factor_ import divisors
 from sympy.polys.densetools import dup_eval
@@ -578,6 +578,8 @@ def _minpoly_compose(ex, x, dom):
         res = _minpoly_sin(ex, x)
     elif ex.__class__ is cos:
         res = _minpoly_cos(ex, x)
+    elif ex.__class__ is tan:
+        res = _minpoly_sin(sin(ex.args[0]), x) * minimal_polynomial(1/cos(ex.args[0]))
     elif ex.__class__ is exp:
         res = _minpoly_exp(ex, x)
     elif ex.__class__ is CRootOf:
```
This is gets evaluted though as:
```py
>>> minpoly(tan(13*pi/45))
(_x**12 + 24*_x**11 - 144*_x**10 - 248*_x**9 + 1680*_x**8 + 864*_x**7 - 7168*_x**6 - 1152*_x**5 + 13824*_x**4 + 512*_x**3 - 12288*_x**2 + 4096)*(16777216*_x**24 - 100663296*_x**22 + 264241152*_x**20 - 398196736*_x**18 + 379846656*_x**16 - 238436352*_x**14 + 99147776*_x**12 - 26797056*_x**10 + 4485888*_x**8 - 423872*_x**6 + 18912*_x**4 - 288*_x**2 + 1)
```
Even after the diffs, our main issue of constructing the DomainMatrix is unsolved.

hyadav2k:
This PR adds the `_minpoly_tan` method which calculates minimal polynomial of tan(n*pi), if n is rational.

Fixes #21430 

The `_minpoly_tan` method evaluates minimal polynomial by first calculating the binomial coefficient for `(1 + I*x)**n`, then we take the real or imaginary part according to the numerator of the argument of `tan` `* n`, as real part if odd multiple of `pi/2` else imaginary.

#### Other comments
For reference - https://github.com/sympy/sympy/issues/21430#issuecomment-833428491

* polys
    * `_minpoly_tan` method is added for calculating minimal polynomial for tan(n*pi), if n is rational.

### DIFF
--- a/sympy/polys/numberfields.py
+++ b/sympy/polys/numberfields.py
@@ -462,22 +462,13 @@ def _minpoly_tan(ex):
         if c.is_rational:
             c = c * 2
             n = int(c.q)
-            d = [1] * (n + 1)
-            a = 1
-            for k in range(1, n + 1):
-                a = (a * (n - k + 1))//k
-                if k % 4 == 2 or k % 4 == 3:
-                    d[k] = -a
-                else:
-                    d[k] = a
+            a = n if c.p % 2 == 0 else 1
+            terms = []
+            for k in range((c.p+1)%2, n+1, 2):
+                terms.append(a*x**k)
+                a = -(a*(n-k-1)*(n-k)) // ((k+1)*(k+2))
 
-            if c.p % 2 == 0:
-                coeff = {j : d[j] for j in range(1, n+1, 2)}
-            else:
-                coeff = {j : d[j] for j in range(0, n+1, 2)}
-
-            a = [x**i*coeff[i] for i in coeff.keys()]
-            r = Add(*a)
+            r = Add(*terms)
             _, factors = factor_list(r)
             res = _choose_factor(factors, x, ex)
             return res
--- a/sympy/polys/numberfields.py
+++ b/sympy/polys/numberfields.py
@@ -10,7 +10,6 @@ from sympy.functions import sqrt, cbrt
 
 from sympy.core.exprtools import Factors
 from sympy.core.function import _mexpand
-from sympy.core.symbol import Symbol
 from sympy.functions.elementary.exponential import exp
 from sympy.functions.elementary.trigonometric import cos, sin, tan
 from sympy.ntheory import sieve
@@ -451,12 +450,11 @@ def _minpoly_cos(ex, x):
     raise NotAlgebraic("%s doesn't seem to be an algebraic element" % ex)
 
 
-def _minpoly_tan(ex):
+def _minpoly_tan(ex, x):
     """
     Returns the minimal polynomial of ``tan(ex)``
     see https://github.com/sympy/sympy/issues/21430
     """
-    x = Symbol('x', real=True)
     c, a = ex.args[0].as_coeff_Mul()
     if a is pi:
         if c.is_rational:
@@ -605,7 +603,7 @@ def _minpoly_compose(ex, x, dom):
     elif ex.__class__ is cos:
         res = _minpoly_cos(ex, x)
     elif ex.__class__ is tan:
-        res = _minpoly_tan(ex)
+        res = _minpoly_tan(ex, x)
     elif ex.__class__ is exp:
         res = _minpoly_exp(ex, x)
     elif ex.__class__ is CRootOf:
--- a/sympy/polys/numberfields.py
+++ b/sympy/polys/numberfields.py
@@ -10,8 +10,9 @@ from sympy.functions import sqrt, cbrt
 
 from sympy.core.exprtools import Factors
 from sympy.core.function import _mexpand
+from sympy.core.symbol import Symbol
 from sympy.functions.elementary.exponential import exp
-from sympy.functions.elementary.trigonometric import cos, sin
+from sympy.functions.elementary.trigonometric import cos, sin, tan
 from sympy.ntheory import sieve
 from sympy.ntheory.factor_ import divisors
 from sympy.polys.densetools import dup_eval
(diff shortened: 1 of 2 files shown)


sympy/polys/numberfields.py
```diff
@@ -450,6 +451,40 @@ def _minpoly_cos(ex, x):
...
+                    d[k] = -a
+                else:
+                    d[k] = a
+
+            if c.p % 2 == 0:
+                coeff = {j : d[j] for j in range(1, n+1, 2)}
+            else:
+                coeff = {j : d[j] for j in range(0, n+1, 2)}
+
+            a = [x**i*coeff[i] for i in coeff.keys()]
```
jksuom, line 479: A single list of terms should suffice. Something like this in the odd case:
```
        terms = []
        a = 1
        for k in range(0, n+1, 2):
            terms.append(a*x**k)
            a = -(a*(n-k-1)*(n-k)) // ((k+1)*(k+2))
```

hyadav2k: Sure, there was unnecessary memory used.
Kindly review the latest commit.
Thanks.

sympy/polys/numberfields.py
```diff
@@ -450,6 +451,31 @@ def _minpoly_cos(ex, x):
     raise NotAlgebraic("%s doesn't seem to be an algebraic element" % ex)
 
 
+def _minpoly_tan(ex):
+    """
+    Returns the minimal polynomial of ``tan(ex)``
+    see https://github.com/sympy/sympy/issues/21430
+    """
+    x = Symbol('x', real=True)
```
jksuom, line 459: `x` should be given as a call parameter. Creating a new symbol may be confusing if it is also in use elsewhere.

hyadav2k: Sure, it is given as a parameter rather than declared in the function.
Thanks,

sympy/polys/tests/test_numberfields.py
```diff
@@ -263,6 +263,20 @@ def test_minpoly_compose():
     mp = minimal_polynomial(ex, x)
     assert degree(mp) == 48 and mp.subs({x:0}) == -16630256576
 
+    ex = tan(13*pi/45)
+    mp = minimal_polynomial(ex, x)
+    a = Poly(mp).nroots()
```
jksuom, line 270: There are two issues with this type of test: computing all roots of `mp` is a waste of time, and then is is no guarantee that the floating point approximation of `ex` will be *exactly* the same as the element in `nroots`. Comparison of Floats is always unreliable.

The validity of the minimal polynomials can be tested with values of `tan` that can be evaluated exactly by SymPy (see e.g. `test_tan` in `test_trigonometric.py`). It is necessary to set `evaluate=False` to suppress premature evaluation, for example:
```
In [16]: ex = tan(pi/5, evaluate=False); ex, tan(pi/5)                          
Out[16]: 
⎛   ⎛π⎞    __________⎞
⎜tan⎜─⎟, ╲╱ 5 - 2⋅√5 ⎟
⎝   ⎝5⎠              ⎠

In [17]: mp = minimal_polynomial(ex, x)                                         

In [18]: mp.subs(x, tan(pi/5)).is_zero                                          
Out[18]: True
```

jksuom:
```
In [2]: tan(pi/5, evaluate=False) == tan(pi/5)
Out[2]: False
```
This is why we need `evaluate=False`. Otherwise, `tan(pi/5)` will evaluate to an explicit algebraic expression which cannot be used to test `minpoly_tan`.  The unevaluated `tan(x)` will be used to compute the minimal polynomial. The evaluated expression will then be substituted to the polynomial to verify that it is a root.

It is not necessary to compute polynomials of high degree (like 24) for the tests. The logic on `minpoly_tan` can be tested with fairly simple expressions.

jksuom: >  The unevaluated tan(x) will be used to compute the minimal polynomial. The evaluated expression will then be substituted to the polynomial to verify that it is a root.

The argument `x` must be chosen such that the unevaluated and evaluated `tan(x)` are different.

sympy/polys/tests/test_numberfields.py
```diff
@@ -263,6 +263,18 @@ def test_minpoly_compose():
...
+    ex = tan(pi/5, evaluate=False)
+    mp = minimal_polynomial(ex, x)
+    assert mp == x**4 - 10*x**2 + 5
+    assert mp.subs(x, tan(pi/5)).is_zero
+
+    ex = tan(pi/6, evaluate=False)
+    mp = minimal_polynomial(ex, x)
+    assert mp == 3*x**2 - 1
+    assert mp.subs(x, tan(pi/6)).is_zero
+
```
jksuom, line 275: Can you add another example of higher degree for the even case, `pi/10`, for example.

hyadav2k: Sure, will do so.

hyadav2k:
> Can you add release notes?

Sure, I have added, kindly have a look.
Thanks.

jksuom:
Thanks, this is ready to go.

PR #21448, opened 2021-05-09
spacing fixed in beam docs

### LINKED ISSUE #21447: Beam docs have weird spacing of powers
The beam docs have a lot of spacing issues when displaying singularity functions with powers.
For example:
image
image
image

Psycho-Pirate:
Fixes #21447
#### Release Notes

NO ENTRY

### DIFF
--- a/doc/src/modules/physics/continuum_mechanics/beam_problems.rst
+++ b/doc/src/modules/physics/continuum_mechanics/beam_problems.rst
@@ -111,7 +111,7 @@ singularity function form:
    :include-source: True
 
    >>> b.load
-              -1
+             -1
    12⋅<x - 9>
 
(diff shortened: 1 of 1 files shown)


smichr:
If this output was copied from the actual output of these functions then the functions themselves (or the printers they use) should be fixed to put the exponents in the right place. Otherwise this is only looking good in the docs and not on anyone's interactive session or generated output.

Psycho-Pirate:
> 
> 
> If this output was copied from the actual output of these functions then the functions themselves (or the printers they use) should be fixed to put the exponents in the right place. Otherwise this is only looking good in the docs and not on anyone's interactive session or generated output.

I did investigate it further. The actual output was correct both on Sympy Live Shell and my machine, hence it is simply a documentation issue.

PR #21455, opened 2021-05-10
perf(assump) : enhance get_known_facts_keys

JSS95:

When predicate decomposition was introduced, `get_known_facts_keys()` was modified to exclude the composite predicates. But this turned out to be not only unnecessary but slightly slows down the evaluation. This change reverts that.

#### Other comments

`get_known_facts_keys()` is involved in simple, limited case where a single unary predicate is passed as assumption, e.g. `ask(Q.real(x), Q.positive(x))`. That's why the regression was unnoticed.

In master:
```python
In [1]: %timeit ask(Q.real(x), Q.positive(x))
2.01 ms ± 5.94 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

In this branch:
```python
In [1]: %timeit ask(Q.real(x), Q.positive(x))
601 µs ± 2.16 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

#### Release Notes

NO ENTRY


JSS95:
I'm merging this since it's a simple change which reverts the part of previous commit.

PR #21458, opened 2021-05-11
feat(diffgeom) : introduce rewriting of CoordinateSymbol wrt CoordSystem

JSS95:

`CoordinateSymbol` can now be rewritten with respect to another coordnate system. For example, for Cartesian `x, y` and Polar `r, theta`, `x.rewrite(Polar)` returns `r*cos(theta)`.

#### Other comments

For this, `Basic.rewrite()` is modified to be able to take instance as rewrite rule. This does not change public API.

#### Release Notes

* diffgeom
    * `CoordinateSymbol` now can be rewritten to the expression containing the coordinate symbols of other coordinate system.

### DIFF
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -1676,27 +1676,6 @@ class Basic(Printable, metaclass=ManagedProperties):
         from sympy.assumptions import refine
         return refine(self, assumption)
 
-    def _eval_rewrite(self, pattern, rule, **hints):
-        if self.is_Atom:
-            if hasattr(self, rule):
-                return getattr(self, rule)()
-            return self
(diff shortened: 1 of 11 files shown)


oscarbenjamin:
> For this, `Basic.rewrite()` is modified to be able to take instance as rewrite rule. This does not change public API.

I'm not sure about this change. Also it seems that `_eval_rewrite_as_x` needs to take `**kwargs` which wouldn't be backwards compatible for downstream `_eval_rewrite` methods.

Changing coordinate system doesn't feel like a "rewrite" to me. I'm reluctant to make any changes to Basic for this.

Issue #21017, opened 2021-03-01
AppliedPredicate should not be atom


JSS95:
`AppliedPredicate` has `is_Atom = True` by class attribute.

```python
>>> from sympy import Q
>>> from sympy.abc import x
>>> Q.positive(x).is_Atom
True
```

The comment says not to change this and making `is_Atom = False` indeed raised error in `logic/inference` module. However this should be fixed because applied predicate is a function applied to arguments and thus never should be atom. This attribute makes many tree-recursive operations not applied to the arguments in `AppliedPredicate`. There must be some way to allow inference on applied predicate without making it atom.

asmeurer:
I agree with this. I also want to say that the idea that Atom should be the indicator of leaf notes in an expression tree is very misguided. The indicator of such should be an empty .args. Leaf nodes should not need to have a special attribute or be a subclass of a certain class. And conversely I don't think you should be able to fool algorithms into thinking you are a leaf node by pretending to be an Atom with is_Atom.

PR #21464, opened 2021-05-12
refact(diffgeom) : redesign CoordSystem

JSS95:

`Manifold`, `Patch` and `CoordSystem` are now just `Basic`, and not `Atom` anymore. An object being a leaf node should be indicated by empty `.args`, not being a subclass of certain class. (See #21017)

`CoordSystem` now uses simpler construction signature. Previously, transformation relation was built using `Lambda`. Now, only a tuple is suffice.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/diffgeom/diffgeom.py
+++ b/sympy/diffgeom/diffgeom.py
@@ -532,7 +532,7 @@ class CoordSystem(Basic):
         Returns
         =======
 
-        sympy.ImmutableDenseMatrix
+        sympy.ImmutableDenseMatrix containing CoordinateSymbol
 
         Examples
         ========
--- a/sympy/diffgeom/diffgeom.py
+++ b/sympy/diffgeom/diffgeom.py
@@ -6,7 +6,7 @@ from itertools import permutations
 from sympy.combinatorics import Permutation
 from sympy.core import (
     Basic, Expr, Function, diff,
-    Pow, Mul, Add, Atom, Lambda, S, Tuple, Dict
+    Pow, Mul, Add, Lambda, S, Tuple, Dict
 )
 from sympy.core.cache import cacheit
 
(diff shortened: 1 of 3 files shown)


sympy/diffgeom/diffgeom.py
```diff
@@ -499,17 +572,71 @@ def coord_tuple_transform_to(self, to_sys, coords):
...
+        [1, 0],
+        [0, 1]])
+
         """
         result = self.transform(sys).jacobian(self.symbols)
         if coordinates is not None:
             result = result.subs(list(zip(self.symbols, coordinates)))
         return result
-    jacobian_matrix = jacobian
+    jacobian_mat = jacobian_matrix = jacobian
```
sylee957, line 608: Is it necessary to have these shortcuts?

sympy/diffgeom/diffgeom.py
```diff
@@ -320,11 +349,8 @@ def manifold(self):
 
     @property
     def symbols(self):
-        return [
-            CoordinateSymbol(
-                self, i, **s._assumptions.generator
-            ) for i,s in enumerate(self.args[2])
-        ]
+        return Tuple(*[CoordinateSymbol(self, i, **s._assumptions.generator)
+            for i,s in enumerate(self.args[2])])
```
sylee957, line 353: I would consider using built-in `tuple` than `Tuple` if this is not necessary.

sympy/diffgeom/diffgeom.py
JSS95: probably not

sympy/diffgeom/diffgeom.py
```diff
@@ -499,17 +572,70 @@ def coord_tuple_transform_to(self, to_sys, coords):
...
+        Matrix([
+        [1, 0],
+        [0, 1]])
+
         """
         result = self.transform(sys).jacobian(self.symbols)
         if coordinates is not None:
             result = result.subs(list(zip(self.symbols, coordinates)))
         return result
-    jacobian_matrix = jacobian
```
sylee957, line 508: No, I don't think that you should remove this line completely because it was there in the previous version.
But I would rather remove `jacobian_det` below if that is also accidental

JSS95: Oops. This was a mistake.

BTW I think the real names for two methods should be `jacobian_mat()` and `jacobian_det()` for clarity, since the term "jacobian" is ambiguous. (https://mathworld.wolfram.com/Jacobian.html)

JSS95: I don't want this topic to be a blocker for this PR so I'll open a separate issue for the naming of methods.

PR #21465, opened 2021-05-13
Make Predicate not Boolean

JSS95:

`Predicate` is no more `Boolean`.

#### Other comments

Internal structure of `ask` module is enhanced for cleaner code and documentation.

#### Release Notes

* assumptions
  * `Predicate` class is no more a subclass of `Boolean`.

### DIFF
--- a/sympy/assumptions/cnf.py
+++ b/sympy/assumptions/cnf.py
@@ -140,10 +140,17 @@ class AND:
     __repr__ = __str__
 
 
-def to_NNF(expr, composite_map={}):
+def to_NNF(expr, decomposer=None):
     """
     Generates the Negation Normal Form of any boolean expression in terms
     of AND, OR, and Literal objects.
+
+    Parameters
+    ==========
+
+    expr : Boolean
+
+    decomposer : function which decomposes the predicate, optional.
 
     Examples
     ========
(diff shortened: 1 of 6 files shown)


sympy/assumptions/facts.py
```diff
@@ -15,24 +15,54 @@
...
+    >>> decompose_predicate(Q.real(x))
+    Q.negative(x) | Q.positive(x) | Q.zero(x)
+
+    """
+    p, args = pred.function, pred.arguments
+
+    if p == Q.real:
+        return Q.negative(*args) | Q.zero(*args) | Q.positive(*args)
+    if p == Q.integer:
+        return Q.even(*args) | Q.odd(*args)
```
oscarbenjamin, line 44: Maybe there should still be a precomputed dict but like:
```
map = {
    Q.real: [Q.negative, Q.zero, Q.positive],
}
```
Then this function can do:
```
preds = map[p]
return Or(*(pred(*args) for pred in preds))
```
Note that calling `Or` once is more efficient then chaining with `|` for more than two items.

JSS95: Done.

sympy/assumptions/cnf.py
```diff
@@ -140,11 +140,18 @@ def __str__(self):
     __repr__ = __str__
 
 
-def to_NNF(expr, composite_map={}):
+def to_NNF(expr, decomposer=None):
```
smichr, line 143: this is backwards incompatible -- does this matter?

smichr: Can we just leave the args as they are but if modify documentation to indicate that if `composite_map` is `{}` then a full decomposition of the predicate will be supplied by `decompose_predicate`?

sympy/assumptions/facts.py
```diff
@@ -11,29 +11,55 @@
 from sympy.core.cache import cacheit
 from sympy.core.symbol import Symbol
 from sympy.logic.boolalg import (to_cnf, And, Not, Implies, Equivalent,
-    Exclusive,)
+    Exclusive, Or,)
 from sympy.logic.inference import satisfiable
 
 
-@cacheit
-def get_composite_predicates():
```
smichr, line 19: No harm in leaving this here, is there? This can be called from `decompose_predicate` rather than embedding the non-cached version there. That way anyone that wants this mapping can get it -- kind of like being able to retrieve a list of all the old-style assumptions.

sympy/assumptions/cnf.py
JSS95: This argument was introduced in 1.9 dev so it won't be backwards incompatible. But I'm open to any suggestion.

sympy/assumptions/facts.py
smichr: Any thoughts on this?

asmeurer:
My biggest concern here is that this is backwards incompatible. The release notes should indicate this. The release notes are also unclear that only the predicate objects are not Boolean. Q.real is no longer Boolean but Q.real(x) is.

Since Predicate already has a metaclass, we could add `__and__` and so on to give more informative error messages, as well in the logic constructors, so that people are at least aware that this was removed intentionally. We could also do a deprecation cycle, although that may be too much of a burden for continuing with other refactoring work.

asmeurer:
It would also be helpful to give some motivation for this change in the PR description. Generally we should break backwards compatibility without a good reason. Are future refactorings impossible without this?

JSS95:
In fact, by kind system we can make `Predicate` not `Boolean` but still allow them in boolean functions (which will be boolean composite function). I will write a detailed idea in wiki page. Until then, this PR should not be merged.

PR #21467, opened 2021-05-14
feat(assumptions): Introduce predicate fact registration

JSS95:

Facts for the predicates can now be registered. For example, `Q.gt(x, y) >> Q.extended_real(x) & Q.extended_real(y)` can be registered to let `ask(Q.extended_real(x), x > y)` return `True`.

#### Other comments

This system can be used to register the facts such as `Q.eq(x, y) >> Q.zero(x - y)`. With this, relational predicates can be assumed so that `ask(Q.positive(x+1), x > 0)` returns `True`, which will be a great improvement for `Piecewise`. It will be done in succeeding PRs.

Note : equality/inequality evaluation should use advanced algorithms to be efficiently done. Converting equality to being zero should be done only as a fallback.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/assumptions/satask.py
+++ b/sympy/assumptions/satask.py
@@ -104,10 +104,10 @@ def check_satisfiability(prop, _prop, factbase):
         raise ValueError("Inconsistent assumptions")
 
 
-def extract_predargs(proposition, assumptions=None, context=None):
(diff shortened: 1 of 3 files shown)


oscarbenjamin:
I'm sorry this seems to have been unreviewed. I took a quick look but I think I'd need more time to get my head round what these changes do.

PR #21469, opened 2021-05-14
Better handling of Laplace transforms of functions containing DiracDelta

### LINKED ISSUE #10840: laplace_transform of DiracDelta yields wrong result
`laplace_transform(DiracDelta(t),t,s)`
should yield 1, but instead yields 1/2 . 

version 0.7.6.1 of sympy.

Example at
https://github.com/josephcslater/iPythonExamples/blob/master/Laplace_Transforms_Solution.ipynb

hanspi42:

Fixes #10840

`laplace_transform` now checks whether its argument has a `DiracDelta(t)` in it, and 
if yes, treats the cases where `t=0` correctly if the argument of `DiracDelta` is a
first-order polynomial.

#### Release Notes

* integrals
  * `laplace_transform(DiracDelta(t),t,s)` now gives 1 as expected.

### DIFF
--- a/sympy/integrals/tests/test_transforms.py
+++ b/sympy/integrals/tests/test_transforms.py
@@ -451,7 +451,7 @@ def test_inverse_mellin_transform():
 
 @slow
 def test_laplace_transform():
-    from sympy import fresnels, fresnelc
+    from sympy import fresnels, fresnelc, DiracDelta
     LT = laplace_transform
     a, b, c, = symbols('a b c', positive=True)
     t = symbols('t')
@@ -504,6 +504,28 @@ def test_laplace_transform():
         ((s - 1)/(s**2 - 2*s + 2), -oo),
         ((s - 1)/((s - 1)**2 + 1), -oo),
     ]
+
+    # DiracDelta function: standard cases
+    assert LT(DiracDelta(t), t, s) == (1, -oo, True)
+    assert LT(DiracDelta(a*t), t, s) == (1/a, -oo, True)
+    assert LT(DiracDelta(t/42), t, s) == (42, -oo, True)
+    assert LT(DiracDelta(t+42), t, s) == (0, -oo, True)
+    assert LT(DiracDelta(t)+DiracDelta(t-42), t, s) == \
+        (1 + exp(-42*s), -oo, True)
+    assert LT(DiracDelta(t)-a*exp(-a*t), t, s) == (-a/(a + s) + 1, 0, True)
+    assert LT(exp(-t)*(DiracDelta(t)+DiracDelta(t-42)), t, s) == \
+        (exp(-42*s - 42) + 1, -oo, True)
+    # Collection of cases that cannot be fully evaluated and/or would catch
+    # some common implementation errors
+    assert LT(DiracDelta(t**2), t, s) == LaplaceTransform(DiracDelta(t**2), t, s)
+    assert LT(DiracDelta(t**2 - 1), t, s) == (exp(-s)/2, -oo, True)
+    assert LT(DiracDelta(t*(1 - t)), t, s) == \
+        LaplaceTransform(DiracDelta(-t**2 + t), t, s)
+    assert LT((DiracDelta(t) + 1)*(DiracDelta(t - 1) + 1), t, s) == \
+        (LaplaceTransform(DiracDelta(t)*DiracDelta(t - 1), t, s) + \
+         1 + exp(-s) + 1/s, 0, True)
+    assert LT(DiracDelta(2*t - 2*exp(a)), t, s) == \
+        (exp(-s*exp(a))/2, -oo, True)
 
     # Fresnel functions
     assert laplace_transform(fresnels(t), t, s) == \
--- a/sympy/integrals/transforms.py
+++ b/sympy/integrals/transforms.py
@@ -7,6 +7,7 @@ from sympy.core.function import Function
 from sympy.core.relational import _canonical, Ge, Gt
 from sympy.core.numbers import oo
 from sympy.core.symbol import Dummy
+from sympy.functions import DiracDelta
 from sympy.functions.elementary.miscellaneous import Max
 from sympy.integrals import integrate, Integral
 from sympy.integrals.meijerint import _dummy
@@ -15,6 +16,7 @@ from sympy.simplify import simplify
 from sympy.utilities import default_sort_key
 from sympy.utilities.exceptions import SymPyDeprecationWarning
 from sympy.matrices.matrices import MatrixBase
+from sympy.polys.matrices.linsolve import _lin_eq2dict, PolyNonlinearError
 
 
 ##########################################################################
@@ -1005,14 +1007,39 @@ def _simplifyconds(expr, s, a):
     expr = repl(expr, Unequality, replue)
     return S(expr)
 
+def expand_dirac_delta(expr):
+    """
+    Expand an expression involving DiractDelta to get it as a linear
+    combination of DiracDelta functions.
+    """
+    return _lin_eq2dict(expr, expr.atoms(DiracDelta))
 
 @_noconds
 def _laplace_transform(f, t, s_, simplify=True):
     """ The backend function for Laplace transforms. """
     from sympy import (re, Max, exp, pi, Min, periodic_argument as arg_,
-                       arg, cos, Wild, symbols, polar_lift)
+                       arg, cos, Wild, symbols, polar_lift, Add)
     s = Dummy('s')
-    F = integrate(exp(-s*t) * f, (t, 0, oo))
+    a = Wild('a', exclude=[t])
+    deltazero = []
+    deltanonzero = []
+    try:
+        integratable, deltadict = expand_dirac_delta(f)
+    except PolyNonlinearError:
+        raise IntegralTransformError(
+        'Laplace', f, 'could not expand DiracDelta expressions')
+    for dirac_func, dirac_coeff in deltadict.items():
+        p = dirac_func.match(DiracDelta(a*t))
+        if p:
+            deltazero.append(dirac_coeff.subs(t,0)/p[a])
+        else:
+            if dirac_func.args[0].subs(t,0).is_zero:
+                raise IntegralTransformError('Laplace', f,\
+                                             'not implemented yet.')
+            else:
+                deltanonzero.append(dirac_func*dirac_coeff)
+    F = Add(integrate(exp(-s*t) * Add(integratable, *deltanonzero), (t, 0, oo)),
+            Add(*deltazero))
 
     if not F.has(Integral):
         return _simplify(F.subs(s, s_), simplify), -oo, S.true
@@ -1150,17 +1177,24 @@ def laplace_transform(f, t, s, legacy_matrix=True, **hints):
     r"""
     Compute the Laplace Transform `F(s)` of `f(t)`,
 
-    .. math :: F(s) = \int_0^\infty e^{-st} f(t) \mathrm{d}t.
+    .. math :: F(s) = \int_{0^{-}}^\infty e^{-st} f(t) \mathrm{d}t.
 
     Explanation
     ===========
 
-    For all "sensible" functions, this converges absolutely in a
+    For all sensible functions, this converges absolutely in a
     half plane  `a < \operatorname{Re}(s)`.
 
-    This function returns ``(F, a, cond)``
-    where ``F`` is the Laplace transform of ``f``, `\operatorname{Re}(s) > a` is the half-plane
+    This function returns ``(F, a, cond)`` where ``F`` is the Laplace
+    transform of ``f``, `\operatorname{Re}(s) > a` is the half-plane
     of convergence, and ``cond`` are auxiliary convergence conditions.
+
+    The lower bound is `0^{-}`, meaning that this bound should be approached
+    from the lower side. This is only necessary if distributions are involved.
+    At present, it is only done if `f(t)` contains ``DiracDelta``, in which
+    case the Laplace transform is computed as
+
+    .. math :: F(s) = \lim_{\tau\to 0^{-}} \int_{\tau}^\infty e^{-st} f(t) \mathrm{d}t.
 
     If the integral cannot be computed in closed form, this function returns
     an unevaluated :class:`LaplaceTransform` object.
@@ -1182,8 +1216,11 @@ def laplace_transform(f, t, s, legacy_matrix=True, **hints):
 
     >>> from sympy.integrals import laplace_transform
     >>> from sympy.abc import t, s, a
+    >>> from sympy.functions import DiracDelta, exp
     >>> laplace_transform(t**a, t, s)
     (gamma(a + 1)/(s*s**a), 0, re(a) > -1)
+    >>> laplace_transform(DiracDelta(t)-a*exp(-a*t),t,s)
+    (-a/(a + s) + 1, 0, Abs(arg(a)) <= pi/2)
 
     See Also
     ========


sympy/integrals/transforms.py
```diff
@@ -1008,9 +1008,13 @@ def repl(ex, *args):
...
     from sympy import (re, Max, exp, pi, Min, periodic_argument as arg_,
-                       arg, cos, Wild, symbols, polar_lift)
+                       arg, cos, Wild, symbols, polar_lift, limit, DiracDelta)
     s = Dummy('s')
-    F = integrate(exp(-s*t) * f, (t, 0, oo))
+    zero_minus = symbols('zero_minus',real=True)
+    if f.has(DiracDelta):
+        F = limit(integrate(exp(-s*t) * f, (t, zero_minus, oo)), zero_minus, 0, dir='-')
+    else:
+        F = integrate(exp(-s*t) * f, (t, 0, oo))
```
oscarbenjamin, line 1017: Why does this do if f.has(DiracDelta)?

If we are saying that this is basically the definition of the Laplace transform then surely it would also apply when there is no Dirac delta function.

hanspi42: > Why does this do if f.has(DiracDelta)?
> 
> If we are saying that this is basically the definition of the Laplace transform then surely it would also apply when there is no Dirac delta function.

The one-sided `laplace_transform` must work for all functions that behave well for `t>=0`, no matter how crazy they behave for negative `t`, as long as they are OK at infinitesimally small negative values.

I cannot be sure that `integrate(exp(-s*t) * f, (t, zero_minus, oo))` will always converge when `integrate(exp(-s*t) * f, (t, 0, oo))` converges, so I'd rather add an exception to be on the safe side. It also needs less computation time.

Unless there is a `DiracDelta` at `t==0`, it makes no difference anyway. If we ever encounter another delta-like distribution that Laplace needs to know about, we can add it too.

oscarbenjamin: If we're just making a special case for a Dirac delta at t=0 then it would be better just to compute the special case result rather than using `limit` since as you say there is no guarantee that the limit exists anyway.

hanspi42: Great idea! That is very simple, and I tried a few things, it works well.

hanspi42:
The test was cancelled due to an internal github error. Can I somehow re-start it without making another chech-in?

hanspi42:
OK, still does not work. I'll just wait ...

hanspi42:
Could someone please remove the "draft" status?

I though switching there and back might trigger the tests again, but now I cannot switch back from draft anymore.

sympy/integrals/transforms.py
```diff
@@ -1008,9 +1008,33 @@ def repl(ex, *args):
...
+                       arg, cos, Wild, symbols, polar_lift, limit,
+                       Add, DiracDelta)
     s = Dummy('s')
-    F = integrate(exp(-s*t) * f, (t, 0, oo))
+    zero_minus = symbols('zero_minus', real=True)
+    # Expressions with DiracDelta need to be computed with a directional
+    # limit, but this will fail on some sums, so we first use the
+    # linearity of the Laplace transform to split the integral into a sum
+    # of integrals.
+    if f.has(DiracDelta(t)):
```
oscarbenjamin, line 1019: Will this work for something like `DiracDelta(2*t)`?

hanspi42: > Will this work for something like `DiracDelta(2*t)`?

No :(

But fortunately the new part I wrote now makes it unnecessary to restrict the algorithm only to Dirac impulses at 0, so the solution is simply to use if `f.has(DiracDelta):`

hanspi42: I have pushed a solution; github's check fail due to an internal misconfiguration, though.  Will you look at it again before the checks run, or shall I try and force the checks?` before the review?

jksuom:
I think that `expand` is not a good choice for separating distributions with finite support. I would consider implementing a new utility function, say, `expand_delta`, that would expand an expression and return two sums: one linear combination of delta functions (and their derivatives) and the rest.

hanspi42:
> I think that `expand` is not a good choice for separating distributions with finite support. I would consider implementing a new utility function, say, `expand_delta`, that would expand an expression and return two sums: one linear combination of delta functions (and their derivatives) and the rest.

Would you have an example in which `expand()` shows its bad qualities?

jksuom:
`expand` is a general purpose method that does many things most of which are irrelevant to the task of extracting delta functions. It does not necessarily create bad expressions but will be a waste of time. A specialized expand function could raise an error when encountering products and powers of delta functions, delta functions in arguments of functions (what is `exp(DiracDelta(t))`?), etc...

hanspi42:
> ... but will be a waste of time. A specialized expand function could raise an error when encountering products and powers of delta functions, delta functions in arguments of functions (what is `exp(DiracDelta(t))`?), etc...

I like this idea, but it is not something that should be done within `laplace_transform`. I think `expand` should be extended.

However, I have now told `expand` what exactly it should or should not do, so it will now not try much beyond expanding multinomials. This already makes `laplace_transform` with `DiracDelta` faster.

Do you think this is OK for now?

jksuom:
> not something that should be done within laplace_transform

I agree. It could be done for all integral transforms, or even better, for all integrals. It would be harder to implement as an extension of the general `expand` as it would return two expressions instead of one.

> beyond expanding multinomials.

This should only be done if the multinomial does not involve delta functions.

hanspi42:
> > beyond expanding multinomials.
> 
> This should only be done if the multinomial does not involve delta functions.

That one I do not understand. Under what condition is it wrong to expand a multinomial containing DiracDelta?

jksuom:
If a sum contains a delta function, then its expanded powers will contain powers of delta functions. Those should raise an error.

hanspi42:
> 
> 
> If a sum contains a delta function, then its expanded powers will contain powers of delta functions. Those should raise an error.

Weeeelll ... I prefer garbage-in garbage-out ;)  Also, if the two `DiracDelta` do not have the discontinuity at the same location, multiplying them is OK and should give 0. 

With this pull request, the code lines

```
print(laplace_transform((1+DiracDelta(t))*(1+DiracDelta(t-1)),t,s)[0].simplify())
print(laplace_transform((1+DiracDelta(t))*(1+DiracDelta(t)),t,s))
print(laplace_transform((1+DiracDelta(t))*(1+DiracDelta(2*t)),t,s))
print(laplace_transform((1+DiracDelta(t))*(1+DiracDelta(2*t)),t,s))
print(laplace_transform((1+DiracDelta(t))*(1+DiracDelta(2*t)),t,s)[0]-
      laplace_transform((DiracDelta(t))*(DiracDelta(2*t)),t,s)[0])
print(laplace_transform((1+DiracDelta(t)+DiracDelta(2*t)),t,s)[0])
```

will give the result

```
1 + exp(-s) + 1/s
LaplaceTransform((DiracDelta(t) + 1)**2, t, s)
(DiracDelta(0) + 3/2 + 1/s, 0, True)
3/2 + 1/s
3/2 + 1/s
```

The first is correct, the second is a refusal to do it, and the third contains the undefined `DiracDelta(0)` as a warning that something is not good here. The fourth and the fifth show that, even in this situation, the linearity of the Laplace transform is maintained. 

Since this pull request already makes the situation much better, I would use it like this and think about which part of SymPy should complain when it encounters powers of `DiracDelta`. But that must be somewhere more central than `laplace_transform`, I think.

oscarbenjamin:
When would these products of functions involving DiracDelta come up?

I would have thought that it would only be necessary to handle expressions that have linear DiracDelta terms.

hanspi42:
> When would these products of functions involving DiracDelta come up?
> 
> I would have thought that it would only be necessary to handle expressions that have linear DiracDelta terms.

I agree; the others are expressions that should not appear, but of course somebody may try, and somebody may find a way to break `laplace_transform`, but I am not really concerned about that.

hanspi42:
@oscarbenjamin: I have made one last test case that covers the `expand`. I think it is now ready for the final review ;)

sympy/integrals/transforms.py
```diff
@@ -1008,9 +1008,36 @@ def repl(ex, *args):
...
+                       Add, DiracDelta)
     s = Dummy('s')
-    F = integrate(exp(-s*t) * f, (t, 0, oo))
+    zero_minus = symbols('zero_minus', real=True)
+    # Expressions with DiracDelta need to be computed with a directional
+    # limit, but this will fail on some sums, so we first use the
+    # linearity of the Laplace transform to split the integral into a sum
+    # of integrals.
+    if f.has(DiracDelta):
+        if f.func != Add:
```
jksuom, line 1020: Why this test?

hanspi42: It saves the effort of expanding when it is already expanded. Do you think it could be omitted, and expand executed without checking first?

jksuom: If `e` is any expression, then `e + 1` would not be expanded here.

hanspi42: > If `e` is any expression, then `e + 1` would not be expanded here.

You are right, I will change it.

hanspi42: I have run my local tests; since I have restricted what `expand` shall do, I do not see a problem or speed penalty from omitting 'if f.func != Add:'

sympy/integrals/transforms.py
```diff
@@ -1008,9 +1008,34 @@ def repl(ex, *args):
...
+    zero_minus = symbols('zero_minus', real=True)
+    # Expressions with DiracDelta need to be computed with a directional
+    # limit, but this will fail on some sums, so we first use the
+    # linearity of the Laplace transform to split the integral into a sum
+    # of integrals.
+    if f.has(DiracDelta):
+        f = f.expand(deep=True, modulus=None, power_base=False,
+                     power_exp=False, mul=True, log=False, multinomial=True,
+                     basic=True, complex=False, func=False, trig=False)
+        if f.func == Add:
```
jksuom, line 1023: There is an attribute for efficiently testing this condition: `f.is_Add`.

hanspi42: The things I learn here ... thank you! I will update right away, not least because the test pipeline has again failed due to a server problem. Is that normal on github that the tests run so unreliably?

sympy/integrals/transforms.py
```diff
@@ -1008,9 +1008,34 @@ def repl(ex, *args):
 def _laplace_transform(f, t, s_, simplify=True):
     """ The backend function for Laplace transforms. """
     from sympy import (re, Max, exp, pi, Min, periodic_argument as arg_,
-                       arg, cos, Wild, symbols, polar_lift)
+                       arg, cos, Wild, symbols, polar_lift, limit,
+                       Add, DiracDelta)
```
jksuom, line 1012: `Add` should now be removed.

hanspi42: Oh sorry, I ran all tests and failed to notice that flake8 complained.

hanspi42:
I must be doing something wrong with the tests, again the tests have failed (cancelled). The system does not give me a reason.

What am I doing wrong? Could anyone point me towards some instructions that I could read?

oscarbenjamin:
It's just some problem at github:
https://www.githubstatus.com/

hanspi42:
> It's just some problem at github:
> https://www.githubstatus.com/

Phew. And thanks, now I know where to check.

Is there a way for me to re-start the tests without making an additional push?

sympy/integrals/transforms.py
```diff
@@ -1008,9 +1008,34 @@ def repl(ex, *args):
...
+    # linearity of the Laplace transform to split the integral into a sum
+    # of integrals.
+    if f.has(DiracDelta):
+        f = f.expand(deep=True, modulus=None, power_base=False,
+                     power_exp=False, mul=True, log=False, multinomial=True,
+                     basic=True, complex=False, func=False, trig=False)
+        if f.is_Add:
+            _all_parts = []
+            for _part in f.args:
+                if _part.has(DiracDelta(t)):
```
jksuom, line 1026: How do you plan to handle expressions that contain something like `DiracDelta(2*t)`?

hanspi42: Nice catch! I corrected that incorrectly in commit https://github.com/sympy/sympy/pull/21469/commits/7ba0a99e76c8f4659a3dc8bb457be0f78e743ccd , where I removed `(t)` only in one place instead of two places. 

The really curious thing is that it already worked like this. I will still change it with the next commit ;)

oscarbenjamin:
If we only want integrands that are linear in DiracDelta then we can use a function like this:
```python
from sympy.polys.matrices.linsolve import _lin_eq2dict

def expand_dirac_delta(expr):
    """
    Expand an expression involving DiractDelta to get it as a linear
    combination of DiracDelta functions.
    """
    return _lin_eq2dict(expr, expr.atoms(DiracDelta))
```
With that we have:
```python
In [7]: e = t*(DiracDelta(t) + (t + 1)**2)

In [8]: e
Out[8]: 
  ⎛       2       ⎞
t⋅⎝(t + 1)  + δ(t)⎠

In [9]: expand_dirac_delta(e)
Out[9]: 
⎛         2           ⎞
⎝t⋅(t + 1) , {δ(t): t}⎠

In [11]: e = t*(DiracDelta(t) + (t + 1)**2 - DiracDelta(2*t-1)/t)

In [12]: e
Out[12]: 
  ⎛       2          δ(2⋅t - 1)⎞
t⋅⎜(t + 1)  + δ(t) - ──────────⎟
  ⎝                      t     ⎠

In [13]: expand_dirac_delta(e)
Out[13]: 
⎛         2                           ⎞
⎝t⋅(t + 1) , {δ(t): t, δ(2⋅t - 1): -1}⎠
```
This is a special kind of expand that does not expand everything: only those terms involving the DiracDelta are expanded and we get a dict of linear coefficients.

Maybe `_lin_eq2dict` should be made into a public function (but that's a separate PR).

hanspi42:
Wow! I have tried it out, and it gives much simpler code.  However, things like `laplace_transform((1+DiracDelta(t))*(1+DiracDelta(t-1)),t,s)` would throw a `PolyNonlinearError` unless caught, so I catch it by raising an IntegralTransformError which will return the expression unevaluated.  Once I have pushed, try the following. I think the result is quite OK.

```
from sympy import symbols
from sympy.integrals import laplace_transform

from sympy import DiracDelta,exp
t = symbols('t',real=True)
a,omega_0,omega_p,q_p = symbols('a,omega_0,omega_p,q_p',positive=True)
n = symbols('n',integer=True)
s = symbols('s')

expressions = [
    exp(-t),
    exp(-t)*(DiracDelta(t)+DiracDelta(t-42)),
    DiracDelta(t)-a*exp(-a*t),
    (1+DiracDelta(t))*(1+DiracDelta(t-1)),
    (1+DiracDelta(t))*(1+DiracDelta(t)),
    (1+DiracDelta(t))*(1+DiracDelta(2*t)),
    (1+DiracDelta(t)+DiracDelta(2*t)),
    (DiracDelta(2*t))
    ]

print('------------')
for e in expressions:
    f = laplace_transform(e,t,s,noconds=True)
    print(e)
    print('-->',f)
    print('------------')
```

sympy/integrals/transforms.py
```diff
@@ -1008,9 +1008,34 @@ def repl(ex, *args):
...
+                    _all_parts.append(limit(integrate(exp(-s*t) * _part,
```
jksuom: It is inefficient to invoke `limit -> integrate -> Integral -> doit -> ...` separately for each term, in particular, if `expand` creates a great number of terms. It would be better to collect all terms is two lists, one with `DiracDelta`s and the other with the rest, and then integrate the sums (`Add`) of the lists. Those lists can be created with `sift`.

hanspi42: I am aware of the inefficiency, but did it on purpose. I tried your idea first, exactly for efficiency reasons, but found too many cases where `limit`'s `mrv` would not work, but fail with an error. So I chose inefficiency over crashing.

I think this is an OK trade-off, because one of those maybe many terms contains only one Delta function and is not costly to evaluate.

I see in the issues that `mrv` causes other problems too, so maybe we should revisit this discussion when `mrv` is fixed.

Would that be OK for you?

oscarbenjamin: Do we need limit at all?

My suggestion for lineq2dict above decomposes the integrand into a linear combination of DiracDelta with coefficients that are not DiracDelta so we have
```python
a(t) + f_1(t)*δ(g_1(t)) + f_2(t)*δ(g_2(t)) + ...
```
The reason for doing this was just to isolate δ terms that have `g_i(0) = 0` because those are the cases not currently handled correctly by integrate (without using limit). Then we can compute the integral as before but we can add the extra results from some of the δ terms.

All those DiracDelta terms with `g_i(0) = 0` give additional terms that we don't get automatically from integrate. We need to add those terms on and that means adding e.g. `f_i(0) / |g_i'(0)|` provided `g_i'(0) != 0`.

jksuom: > too many cases where limit's mrv would not work,

Can you give examples? Those should be investigated.

hanspi42: > > too many cases where limit's mrv would not work,
> 
> Can you give examples? Those should be investigated.

Sorry, I did not do that professionally enough. I cannot reconstruct those cases immediately. However, if I ever find anything again, I will submit an issue.

hanspi42: > Do we need limit at all?

I think not.

> My suggestion for lineq2dict above decomposes the integrand into a linear combination of DiracDelta with coefficients [...]
> All those DiracDelta terms with `g_i(0) = 0` give additional terms that we don't get automatically from integrate. We need to add those terms on and that means adding e.g. `f_i(0) / |g_i'(0)|` provided `g_i'(0) != 0`.

Brilliant idea! It was not very difficult to integrate, we just have to catch derivatives of `DiracDelta` if we do this. I'll push once my own tests have run through.

hanspi42:
This is a test you can run for your review:

```
from sympy import symbols
from sympy.integrals import laplace_transform

from sympy import DiracDelta,exp
t = symbols('t',real=True)
a,omega_0,omega_p,q_p = symbols('a,omega_0,omega_p,q_p',positive=True)
n = symbols('n',integer=True)
s = symbols('s')

expressions = [
    exp(-t),
    exp(-t)*(DiracDelta(t)+DiracDelta(t-42)),
    DiracDelta(t)-a*exp(-a*t),
    (1+DiracDelta(t))*(1+DiracDelta(t-1)),
    (1+DiracDelta(t))*(1+DiracDelta(t)),
    (1+DiracDelta(t))*(1+DiracDelta(2*t)),
    (1+DiracDelta(t)+DiracDelta(2*t)),
    (DiracDelta(2*t)),
    exp(-a*t)+DiracDelta(a*t)+DiracDelta(t-a,1),
    exp(-a*t)+DiracDelta(a*t)+DiracDelta(t-a),
    DiracDelta(t**2),
    DiracDelta(t**2-1)
    ]

print('------------')
for e in expressions:
    f = laplace_transform(e,t,s,noconds=True)
    print(e)
    print(f)
    print('------------')

```

hanspi42:
@jksuom: I guess you might ask "Why are you integrating each term independently?" The reason is that if I don't, then `integrate()` will not necessarily be able to deal with the sum. I also think it does not cost a lot of computation power to solve each integral independently.

hanspi42:
Comments have stopped appearing here, so does this mean the fix is OK and can be merged?

sympy/integrals/transforms.py
```diff
@@ -1005,14 +1007,37 @@ def repl(ex, *args):
...
+            if len(dd.args)>1 and dd.args[1]>0:
+                raise IntegralTransformError(
+                    'Laplace', f, 'derivatives of DiracDelta not implemented')
+            if dd.args[0].subs(t,0) == 0:
+                _parts.append(deltadict[dd].subs(t,0)/diff(dd.args[0],t).subs(t,0))
+            else:
+                _parts.append(integrate(exp(-s*t)*dd*deltadict[dd],(t, 0, oo)))
+        F = sum(_parts)
+    else:
+        F = integrate(exp(-s*t) * f, (t, 0, oo))
```
oscarbenjamin, line 1040: I think this can be simplified.

We don't need to check `has(DiracDelta)` since `expand_dirac_delta` can handle the case where there is no `DiracDelta`.

We should only call `integrate` once. I think actually the whole expression should be integrated. We only need to extract those DiracDelta terms that would contribute to the result but that would be excluded by `integrate`. Those terms can be added to the result of integrating the whole expression. This is also avoid problems where the argument to delta has multiple zeros e.g. `DiracDelta(x*(1 - x))`.

It's better to use `Add(*terms)` rather than `sum(terms)` because `sum` is `O(n^2)` when used with sympy expressions.

I think that we also need to check for the case that the derivative of the DiracDelta argument is zero.

Checking with `==0` is unreliable. Instead `is_zero` should be used and the three separate cases True, False and None should be handled. In the None case either `IntegralTransformError` could be raised or a Piecewise result could be constructed.

hanspi42: > I think this can be simplified.

Thank you! All of it makes a lot of sense; I will do it next week. One thing I am not sure about:

> I think that we also need to check for the case that the derivative of the DiracDelta argument is zero.

You mean like `DiracDelta(a*x**2)`? So what is the result in that case? Would it always go towards $\infty$, or would it depend on higher derivatives or on the sign of the argument?

hanspi42: P.S. I already found time today, it works well with all examples.

I have done everything except the `slope=0` thing. In case you already want to look at what I have, I push it now.

hanspi42:
Oh bother. I just realized that we have to go back to using limits for the `DiracDelta` terms whose arguments are zero at 't=0'. 

The reason is that if the argument `g(t)` of `DiracDelta(g(t))` has multiple zeros, then all zeros `t_i` of `g(t)` that are in the range `0...oo` will contribute to the Laplace transform, not just the one at `t_i==0`.

So what I have just pushed is mathemathically incorrect.

oscarbenjamin:
> The reason is that if the argument `g(t)` of `DiracDelta(g(t))` has multiple zeros, then all zeros `t_i` of `g(t)` that are in the range `0...oo` will contribute to the Laplace transform, not just the one at `t_i==0`.

What happens if you just leave those in the "integratable" part of the integrand?

I presume that integrate already handles the other zeros (at least it should be expected to)...

hanspi42:
> > The reason is that if the argument `g(t)` of `DiracDelta(g(t))` has multiple zeros, then all zeros `t_i` of `g(t)` that are in the range `0...oo` will contribute to the Laplace transform, not just the one at `t_i==0`.
> 
> What happens if you just leave those in the "integratable" part of the integrand?
> 
> I presume that integrate already handles the other zeros (at least it should be expected to)...

Then the zero at t==0 will again only give half of the contribution to the integral.

Is there a sympy function to check for a 'sufficiently simple' argument of a function? My next attempt would be to only permit linear functions of t in the argument and otherwise not evaluate.

jksuom:
> Is there a sympy function to check for a 'sufficiently simple' argument of a function?

There is a method for that purpose: https://github.com/sympy/sympy/blob/a4327423750065efe9a087a60c20e33b1dba653e/sympy/functions/special/delta_functions.py#L307-L310

> Do you think that would be a viable solution?

Yes, I think so, for the present at least.

hanspi42:
> > Do you think that would be a viable solution?
> 
> Yes, I think so, for the present at least.

I'll implement it, that one is easy.

The problem with the future is that nothing short of finding all zeros of the argument can work.

I could now also reconstruct mrv crashing; I will submit it as a new issue.

hanspi42:
Implemented, tested and pushed.

oscarbenjamin:
If we're only allowing linear linear functions of t in the DiracDelta and it has to be zero at t=0 then we're just looking for something like `DiracDelta(a*t)` where `a != 0`.

I think that the code could be simplified in that case:
```python
In [69]: a = Wild('a')

In [70]: pattern = DiracDelta(a*t)

In [71]: DiracDelta(2*t).match(pattern)
Out[71]: {a: 2}

In [72]: DiracDelta(2*t).match(pattern)[a]
Out[72]: 2
```

hanspi42:
> If we're only allowing linear linear functions of t in the DiracDelta and it has to be zero at t=0 then we're just looking for something like `DiracDelta(a*t)` where `a != 0`.
> 
> I think that the code could be simplified in that case:
> 
> ```python
> In [69]: a = Wild('a')
> 
> In [70]: pattern = DiracDelta(a*t)
> 
> In [71]: DiracDelta(2*t).match(pattern)
> Out[71]: {a: 2}
> 
> In [72]: DiracDelta(2*t).match(pattern)[a]
> Out[72]: 2
> ```

Wow, the things I learn here, thank you!  I was not yet aware of mathemathical pattern matching!  However, it will not simplify it a lot.  Try

`DiracDelta(2*t**2).match(pattern)`

It gives `{a_: 2*t}`, so checking for a first-order polynomial is still necessary.

Therefore, it would only simplify one line, at the cost of introducing two lines more to define the pattern.

So in this case I'd leave the code as it is now.

oscarbenjamin:
> Try
> 
> `DiracDelta(2*t**2).match(pattern)`
> 
> It gives `{a_: 2*t}`, so checking for a first-order polynomial is still necessary.

You need to use `exclude`:
```python
a = Wild('a', exclude=[t])
```

hanspi42:
> You need to use `exclude`:
> 
> ```python
> a = Wild('a', exclude=[t])
> ```

RTFM ;))))

Thanks, works like a charm. I'll push in a minute.

sympy/integrals/transforms.py
```diff
@@ -1005,14 +1007,45 @@ def repl(ex, *args):
...
+    except PolyNonlinearError:
+        raise IntegralTransformError(
+        'Laplace', f, 'could not expand DiracDelta expressions')
+    for dd in deltadict.keys():
+        if len(dd.args)>1 and dd.args[1]>0:
+            raise IntegralTransformError(
+                'Laplace', f, 'derivatives of DiracDelta not implemented')
+        argzero = dd.args[0].subs(t,0).is_zero
+        if argzero is True:
+            p = dd.args[0].match(a*t)
```
oscarbenjamin, line 1036: Since we only match `DiracDelta(a*t)` I don't think that it is necessary to check `argzero`. Instead what we need to know is that `a` is not zero.

sympy/integrals/transforms.py
```diff
@@ -1005,14 +1007,45 @@ def repl(ex, *args):
...
+            if p:
+                deltazero = Add(deltazero,(deltadict[dd].subs(t,0)/p[a]))
+            else:
+                raise IntegralTransformError('Laplace', f, \
+                                             'argument of DiracDelta is not \
+                                              a 1st-order polynomial in t')
+        elif argzero is False:
+            integratable = Add(integratable,dd*deltadict[dd])
```
oscarbenjamin, line 1044: Calling `Add` repeatedly in a loop is `O(N**2)`. You should collect the terms in a list and the call `Add(*terms)` once. 

Also standard formatting puts a space after a comma.

sympy/integrals/transforms.py
oscarbenjamin: If this matched the whole expression `DiracDelta(a*t)` then there also wouldn't be any need to check `len(dd.args)` above.

sympy/integrals/transforms.py
```diff
@@ -1005,14 +1007,45 @@ def repl(ex, *args):
...
     s = Dummy('s')
-    F = integrate(exp(-s*t) * f, (t, 0, oo))
+    a = Wild('a', exclude=[t])
+    deltazero = 0
+    try:
+        integratable, deltadict = expand_dirac_delta(f)
+    except PolyNonlinearError:
+        raise IntegralTransformError(
+        'Laplace', f, 'could not expand DiracDelta expressions')
+    for dd in deltadict.keys():
```
oscarbenjamin, line 1030: It is never necessary in Python to iterate over `d.keys` since iterating directly over `d` is the same i.e.:
```python
for dd in deltadict:
```
Actually though if you also want the corresponding value (which you do below) then it is better to iterate over items:
```python
for dirac_func, dirac_coeff in deltadict.items():
```

sympy/integrals/transforms.py
hanspi42: OK, thanks!

sympy/integrals/transforms.py
hanspi42: And I have again learned something ;)))))

sympy/integrals/transforms.py
hanspi42: OK!

sympy/integrals/tests/test_transforms.py
```diff
@@ -505,6 +505,17 @@ def test_laplace_transform():
         ((s - 1)/((s - 1)**2 + 1), -oo),
     ]
 
+    # Delta function
+    assert LT(DiracDelta(t),t,s) == (1, -oo, True)
+    assert LT(DiracDelta(a*t),t,s) == (1/a, -oo, True)
```
oscarbenjamin, line 510: In this test `a` is declared with `positive=True`. What happens if `a` is just declared as e.g. `Symbol('a', real=True)`?

sympy/integrals/transforms.py
```diff
@@ -1005,14 +1007,35 @@ def repl(ex, *args):
...
+        raise IntegralTransformError(
+        'Laplace', f, 'could not expand DiracDelta expressions')
+    for dirac_func, dirac_coeff in deltadict.items():
+        p = dirac_func.match(DiracDelta(a*t))
+        if p:
+            deltazero.append(dirac_coeff.subs(t,0)/p[a])
+        else:
+            deltanonzero.append(dirac_func*dirac_coeff)
+    F = Add(integrate(exp(-s*t) * Add(integratable, *deltanonzero), (t, 0, oo)),
+            Add(*deltazero))
```
oscarbenjamin, line 1038: What is the output here if I have something like
```python
laplace_transform(DiracDelta(t*(1 - t)), t, s)
```

sympy/integrals/tests/test_transforms.py
hanspi42: This also gives `(1/a, -oo, True)`.

sympy/integrals/transforms.py
hanspi42: `(1/2 + exp(-s), -oo, True)` :(

So again a DiracDelta(t) is integrated incorrectly.  I think checking whether the argument is zero must be done after all.

hanspi42:
After my new commit, the following inputs to laplace_transform give the ouputs on the lines below:

```python
------------
(DiracDelta(t) + DiracDelta(t - 42))*exp(-t)
exp(-42*s - 42) + 1
------------
-a*exp(-a*t) + DiracDelta(t)
-a/(a + s) + 1
------------
(DiracDelta(t) + 1)*(DiracDelta(t - 1) + 1)
LaplaceTransform(DiracDelta(t)*DiracDelta(t - 1), t, s) + 1 + exp(-s) + 1/s
------------
(DiracDelta(t) + 1)**2
LaplaceTransform((DiracDelta(t) + 1)**2, t, s)
------------
(DiracDelta(t) + 1)*(DiracDelta(2*t) + 1)
LaplaceTransform(DiracDelta(t)*DiracDelta(2*t), t, s) + 3/2 + 1/s
------------
DiracDelta(t) + DiracDelta(2*t) + 1
3/2 + 1/s
------------
DiracDelta(2*t - 2*exp(a))
exp(-s*exp(a))/2
------------
DiracDelta(t*exp(a))
exp(-a)
------------
DiracDelta(2*t)
1/2
------------
DiracDelta(a*t) + DiracDelta(-a + t, 1) + exp(-a*t)
s*(1 - Heaviside(a*exp_polar(I*pi), 1/2))*exp(-a*s) + 1/(a + s) + 1/a
------------
DiracDelta(a*t) + DiracDelta(-a + t) + exp(-a*t)
(1 - Heaviside(a*exp_polar(I*pi), 1/2))*exp(-a*s) + 1/(a + s) + 1/a
------------
DiracDelta(t**2)
LaplaceTransform(DiracDelta(t**2), t, s)
------------
DiracDelta(t**2 - 1)
exp(-s)/2
------------
DiracDelta(t*(1 - t))
LaplaceTransform(DiracDelta(-t**2 + t), t, s)
------------
```

sympy/integrals/transforms.py
oscarbenjamin: Maybe we should just match something like `DiracDelta(a*t + b)`. If there is a DiracDelta that doesn't match that then we bail with IntegralTransformError. Otherwise we use something like `p` here:
```python
In [1]: a, b = symbols('a, b')

In [2]: p = Piecewise((0, Ne(b, 0)), (1/a, Ne(a, 0)))

In [3]: p
Out[3]: 
⎧0  for b ≠ 0
⎪            
⎨1           
⎪─  for a ≠ 0
⎩a           

In [4]: p.subs(b, 0)
Out[4]: 
⎧1           
⎨─  for a ≠ 0
⎩a           

In [5]: p.subs(b, 0).subs(a, 0)
Out[5]: nan
```
Should the the condition `Ne(a, 0)` be included in the returned conditions?

oscarbenjamin:
> After my new commit, the following inputs to laplace_transform give the ouputs on the lines below:

Best just to include these in the tests if they aren't already covered. The tests show what is changed.

sympy/integrals/transforms.py
hanspi42: > In [5]: p.subs(b, 0).subs(a, 0)
> Out[5]: nan
> ```
> 
> Should the the condition `Ne(a, 0)` be included in the returned conditions?

No, because `limit(p.subs(b, 0),a, 0)` is valid and correct.

hanspi42:
> Best just to include these in the tests if they aren't already covered. The tests show what is changed.

Very good idea, thanks! I have updated the tests.

sympy/integrals/transforms.py
oscarbenjamin: The limit as a goes to zero is not the same as substituting `a=0`. If a and b are both zero then the integrand has a term like `DiracDelta(0)` and the transform can not have any sensible value.

hanspi42: OK, you are right. Do you have an example for me how to inluce `Ne(a, 0)` into the returned conditions? I have no prior experience with propagating conditions.

oscarbenjamin: I guess where `aux` is returned at the bottom it could be `And(aux, Ne(a, 0))` or something like that. Maybe it's not so important. The result would be `zoo` if you substitute `a=0` anyway...

hanspi42: I think it is not so important. If you calculate the LT of `DiracDelta(a*t - b)`, you get `exp(-b*s/a)/a`.

This cannot come into my code unless b isa already 0 and you actually have `DiracDelta(a*t)`, and for that case the limit would be OK. So I think we can lerave this an conclude. OK?

oscarbenjamin: Yeah, I guess this is fine.

oscarbenjamin:
Okay, I think this looks good. Can you squash the commits?

sympy/integrals/transforms.py
hanspi42: Yippie! .... ehm .... I mean then I can take a breather and then go on to the next topic, inverse LT of rational functions ;)))

hanspi42:
> Okay, I think this looks good. Can you squash the commits?

Probably, I have never done it. Do i need to do it through git commands, or is there a button on github.com to do it?

hanspi42:
Squash: is "The hard(er) and less flexible way" on https://gist.github.com/patik/b8a9dc5cd356f9f6f980 the way to do it?

oscarbenjamin:
> Squash: is "The hard(er) and less flexible way" on https://gist.github.com/patik/b8a9dc5cd356f9f6f980 the way to do it?

I don't think those methods will work since you have merged master into the branch several times.

**IMPORTANT**: Make sure to back up your branch first:
```
git checkout diracdelta_fixes
git checkout -b diracdelta_fixes_backup
```
Then you can squash like this:
```
git checkout diracdelta_fixes
git merge upstream/master
git diff upstream/master > my.diff
git reset --hard HEAD~1000   #  <--- wipe the last 1000 commits
git merge upstream/master     # set the branch to current master
git apply my.diff
git add/commit etc.
git push --force
```

hanspi42:
> Then you can squash like this:
> 

Well, I messed something up (if I only knew what), so the bot reacted, but now it is OK, I think.

oscarbenjamin:
The final diff is good but there is one commit that changes lots of things and then a merge commit that undoes those changes. These two commits still need to be squashed into one so that those changes never "happened" in the history.

The process is the same as I showed above (not sure why it didn't fully work before).

hanspi42:
In order for this to work, might I also need to first merge sympy/master into my fork hanspi42/sympy/master before I even start with your list?

oscarbenjamin:
What I'm referring to as `upstream/master` is probably what you're calling `sympy/master`. My remotes are configured like:
```console
$ git remote -v
origin	git@github.com:oscarbenjamin/sympy.git (fetch)
origin	git@github.com:oscarbenjamin/sympy.git (push)
upstream	https://github.com/sympy/sympy.git (fetch)
upstream	no_push (push)

```

hanspi42:
Same here, but whatever I try, after `git push --force` my branch is 71 commits behind sympy/master and I have to merge again.

oscarbenjamin:
> Same here, but whatever I try, after `git push --force` my branch is 71 commits behind sympy/master and I have to merge again.

Maybe you need a `git fetch upstream`

hanspi42:
> Maybe you need a `git fetch upstream`

That was not the problem, but I found it! It was the diff that was wrong. The following is necessary:

`git diff upstream/master HEAD > my.diff`

And by now I think I might even have understood what I have just done ;)

Thanks for all the support. I hope I will be come lower-maintenance in the future.

oscarbenjamin:
Okay looks good. This is in, thanks!

PR #21472, opened 2021-05-15
Redesign rewrite()

JSS95:

See the discussion https://groups.google.com/g/sympy/c/oemnoHBmfHQ

Previously, parameters of `_eval_rewrite()` were *pattern* and *rule* (as a method name string). Pattern checking and recursive application on the arguments were done by `_eval_rewrite()`

Now, pattern checking and recursive application are done by `rewrite()`. `_eval_rewrite()` takes *rule* (not string anymore, but same with what's passed to `rewrite()`) and *args* (rewritten arguements).

#### Other comments

This change expects `_eval_rewrite()` method to be overridden to define custom rewrite behavior. Previous design did not expect this to be customized and the API method was `_eval_rewrite_as_[...]()`. If user code overrides `_eval_rewrite()` method, this change will break it.

#### Release Notes

* core
    * `rewrite()` method is redesigned. It is now encouraged to define `_eval_rewrite()` method instead of `_eval_rewrite_as_[...]()` method. If user code had overridden `_eval_rewrite()` method, this change may raise backwards incompatibility issue.
    * `_rewrite()` method is defined for `Basic`. If user code had defined `_rewrite()` method, this change may raise backwards incompatibility issue.

* stats
    * `Moment` no more has `None` in `args`.
    * `CentralMoment` no more has `None` in `args`.


JSS95:
@asmeurer 
I redesigned `rewrite` so that rule does not need to be transformed to string.

oscarbenjamin:
It would be more faster to preprocess the patterns and rules once and then recurse with a more efficient representation of the rules.

This should be profiled in particular to see how fast it is when performing a rewrite that has no effect on large expressions e.g.:
```python
x = Symbol('x')
expr = 1 + x
for i in range(10):
    expr = expr + expr**2
```
It should be possible to make this much faster:
```python
In [7]: %time ok = expr.rewrite(Piecewise)
CPU times: user 23.5 ms, sys: 1.44 ms, total: 24.9 ms
Wall time: 23.7 ms

In [10]: %time ok = expr.rewrite(Abs, Piecewise)
CPU times: user 110 ms, sys: 993 µs, total: 111 ms
Wall time: 111 ms
```

JSS95:
@oscarbenjamin 

I enhanced the speed.

In master:
```python
In [1]: x = Symbol('x') 
   ...: expr = 1 + sin(x) 
   ...: for i in range(10): 
   ...:     expr = expr + expr**2 
   ...: 

In [2]: %timeit ok = expr.rewrite(exp)
21.7 ms ± 117 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [3]: x = Symbol('x') 
   ...: expr = 1 + x 
   ...: for i in range(10): 
   ...:     expr = expr + expr**2 
   ...: 

In [4]: %timeit ok = expr.rewrite(Piecewise)
4.72 ms ± 66.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %timeit ok = expr.rewrite(Abs, Piecewise)
1.43 µs ± 24.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

In fccbaa3:
```python
In [1]: x = Symbol('x') 
   ...: expr = 1 + sin(x) 
   ...: for i in range(10): 
   ...:     expr = expr + expr**2 
   ...: 

In [2]: %timeit ok = expr.rewrite(exp)
22.7 ms ± 309 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [3]: x = Symbol('x') 
   ...: expr = 1 + x 
   ...: for i in range(10): 
   ...:     expr = expr + expr**2 
   ...:

In [4]: %timeit ok = expr.rewrite(Piecewise)                                                                                           
6.06 ms ± 44.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %timeit ok = expr.rewrite(Abs, Piecewise)
1.81 µs ± 4.85 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

I think the slight slowdown is due to checking both the old design (`_eval_rewrite_as_Piecewise`) and new design (`_eval_rewrite`).

smichr:
This doesn't appear to regress and we can deprecate the string in the future if desired. This is in.

sympy/core/basic.py
```diff
@@ -1716,73 +1695,119 @@ def _eval_derivative_n_times(self, s, n):
...
-                    pattern = pattern[0]
+                args = expr.args
+            if pattern is None or isinstance(expr, pattern):
+                if hasattr(expr, method):
+                    rewritten = getattr(expr, method)(*args, **hints)
+                else:
+                    rewritten = expr._eval_rewrite(rule, args, **hints)
+                if rewritten is not None:
+                    return rewritten
+            if not args:
```
smichr, line 1802: like the enclosing function, can we move this to the top of this function as a quick exit? And then we could also add a check
```python
if all(i == j for i, j in zip(expr.args, args)):
    return expr
return expr.func(*args)
```
Maybe that would speed things up if there is no change.

oscarbenjamin:
Timings should be compared using `%time` in a fresh process rather than `%timeit` because of the cache.

smichr:
This diff does well -- didn't run full tests on it, however:

```
$ git diff > dif  # changes to this PR
```
```diff
diff --git a/sympy/core/basic.py b/sympy/core/basic.py
index df854874b6..851cd2aae7 100644
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -1786,6 +1786,8 @@ def rewrite(self, *args, deep=True, **hints):
             pattern = None
 
         def _rewrite(expr, pattern, rule, method, **hints):
+            if not expr.args:
+                return expr
             deep = hints.pop('deep', True)
             if deep:
                 args = [_rewrite(a, pattern, rule, method, **hints)
@@ -1799,9 +1801,9 @@ def _rewrite(expr, pattern, rule, method, **hints):
                     rewritten = expr._eval_rewrite(rule, args, **hints)
                 if rewritten is not None:
                     return rewritten
-            if not args:
+            if deep and args == expr.args:
                 return expr
-            return expr.func(*args)
+            return expr.func(*args, evaluate=False)
 
         return _rewrite(self, pattern, rule, method, **hints)
``` 

```python
...
>>> from sympy import *
>>> from sympy.abc import x
>>> e=1+x
>>> for i in range(12):
...  e += e**2
...
>>> from time import time
>>> t=time();q=e.rewrite(Piecewise);(time()-t)
0.13600778579711914
>>> t=time();q=e.rewrite(Piecewise);(time()-t)
0.15700912475585938
>>> ^Z
```
```
$ git stash
```
```python
...
>>> t=time();q=e.rewrite(Piecewise);(time()-t)
0.24701404571533203
>>> t=time();q=e.rewrite(Piecewise);(time()-t)
0.23201322555541992
>>> ^Z
```
```
$ git co master
```
```python
...
>>> t=time();q=e.rewrite(Piecewise);(time()-t)
0.24201393127441406
>>> t=time();q=e.rewrite(Piecewise);(time()-t)
0.21201205253601074
```

smichr:
When I run `bin/test -k rewrite` the times vary a lot, but in the best of 3 runs for each, the timings were off by only a few percent. I think this can be committed. Will wait to see if there is further feedback, however.

asmeurer:
I didn't realize `_eval_rewrite` was already a thing in master. I don't know how much user code was defining that method, but I guess this should be marked as a backwards compatibility break.

sympy/core/basic.py
```diff
@@ -1716,73 +1695,119 @@ def _eval_derivative_n_times(self, s, n):
...
-                # __name__ or __class__.__name__
-                clsname = getattr(args[-1], "__name__", None)
-                if clsname is None:
-                    clsname = args[-1].__class__.__name__
-                rule = '_eval_rewrite_as_' + clsname
+            clsname = rule.__class__.__name__
+            method = "_eval_rewrite_as_%s" % clsname
 
+        if pattern:
+            if iterable(pattern[0]):
```
asmeurer, line 1780: What is the purpose of this code? I see it is in the master version as well. If the first part of the pattern is an iterable, the rest is ignored?

sympy/core/basic.py
```diff
@@ -1716,73 +1695,119 @@ def _eval_derivative_n_times(self, s, n):
...
+
+        >>> expr.rewrite(sin, exp)
+        exp(I*x)/2 + cos(x) - exp(-I*x)/2
+        >>> expr.rewrite([cos,], exp)
+        exp(I*x)/2 + I*sin(x) + exp(-I*x)/2
+        >>> expr.rewrite([cos, sin], exp)
+        exp(I*x)
+
+        Rewriting behavior can be implemented by defining ``_eval_rewrite()``
+        method.
```
asmeurer, line 1744: Should we also document `_eval_rewrite_as`?

sympy/core/basic.py
asmeurer: I guess it was to let you write either `rewrite([a, b, c], rule)` or `rewrite(a, b, c, rule)`. It really should only be one, probably the former. I don't know if it's worth the API break. Personally I'm not really sure if "pattern, rule" is the best API in the first place for expressing more general rewriting patterns.

smichr:
I marked this as "author's turn"...please change to "sympy's turn" if/when @asmeurer's comments have been addressed.

JSS95:
I will attend to this PR in a few days.

sympy/core/basic.py
JSS95: Tests fail with this change.

sympy/core/basic.py
JSS95: I really do dislike this `pattern, rule` signature, but I'd rather bite the bullet than breaking the API here.

sympy/core/basic.py
JSS95: That method is based on the approach where the class name is transformed to method name. We want to get rid of that design, so I deliberately not documented it.

JSS95:
> I didn't realize `_eval_rewrite` was already a thing in master. I don't know how much user code was defining that method, but I guess this should be marked as a backwards compatibility break.

I marked it as you suggested. The right way to define rewrite logic has been to use `_eval_rewrite_as_[...]` method, anyway. User shouldn't have defined `_eval_rewrite()`, and the fact that physics module have done it is a mistake.

sympy/core/basic.py
```diff
@@ -1720,73 +1699,119 @@ def _eval_derivative_n_times(self, s, n):
...
-                rule = '_eval_rewrite_as_' + args[-1]
-            else:
-                # rewrite arg is usually a class but can also be a
-                # singleton (e.g. GoldenRatio) so we check
-                # __name__ or __class__.__name__
-                clsname = getattr(args[-1], "__name__", None)
-                if clsname is None:
-                    clsname = args[-1].__class__.__name__
-                rule = '_eval_rewrite_as_' + clsname
+            clsname = rule.__class__.__name__
```
oscarbenjamin, line 1780: It would be good to keep a comment here that explains what this branch is for

sympy/core/basic.py
oscarbenjamin: I think it's better to document it so that you can say explicitly that it is discouraged.

sympy/core/basic.py
```diff
@@ -1720,73 +1699,119 @@ def _eval_derivative_n_times(self, s, n):
...
+            method = "_eval_rewrite_as_%s" % clsname
 
+        if pattern:
+            if iterable(pattern[0]):
+                pattern = pattern[0]
+            pattern = tuple(p for p in pattern if self.has(p))
```
oscarbenjamin, line 1786: This will recurse the whole of self's tree once for each pattern. Does this actually speed anything up? I would have thought that `_rewrite` below could be made faster if so.

sympy/core/basic.py
```diff
@@ -1720,73 +1699,119 @@ def _eval_derivative_n_times(self, s, n):
...
-                return self._eval_rewrite(None, rule, **hints)
+                return self
+        else:
+            pattern = None
+
+        def _rewrite(expr, pattern, rule, method, **hints):
+            deep = hints.pop('deep', True)
+            if deep:
+                args = [_rewrite(a, pattern, rule, method, **hints)
+                        if isinstance(a, Basic) else a for a in expr.args]
```
oscarbenjamin, line 1796: Is the Basic check here needed? It shouldn't be necessary: all args should be Basic. There was a problem with this in stats before but I think it got fixed.

It's better to not to have an if like this because it's just designed to allow bugs that should really be fixed.

sympy/core/basic.py
```diff
@@ -1720,73 +1699,119 @@ def _eval_derivative_n_times(self, s, n):
...
             else:
-                if iterable(pattern[0]):
-                    pattern = pattern[0]
+                args = expr.args
+            if pattern is None or isinstance(expr, pattern):
```
oscarbenjamin, line 1799: If you make pattern an empty tuple instead of None then you don't need to check for None here.

sympy/core/basic.py
```diff
@@ -1720,73 +1699,119 @@ def _eval_derivative_n_times(self, s, n):
...
+                if hasattr(expr, method):
+                    rewritten = getattr(expr, method)(*args, **hints)
```
oscarbenjamin, line 1801: `hasattr` internally calls `getattr` so this involves two lookups. It can instead be
```
meth = getattr(expr, method, None)
if meth is not None:
     rewritten = meth(*args, **hints)
```

sympy/physics/quantum/tensorproduct.py
```diff
@@ -141,10 +141,8 @@ def flatten(cls, args):
     def _eval_adjoint(self):
         return TensorProduct(*[Dagger(i) for i in self.args])
 
-    def _eval_rewrite(self, pattern, rule, **hints):
-        sargs = self.args
-        terms = [t._eval_rewrite(pattern, rule, **hints) for t in sargs]
-        return TensorProduct(*terms).expand(tensorproduct=True)
+    def _eval_rewrite(self, rule, args, **hints):
+        return TensorProduct(*args).expand(tensorproduct=True)
```
oscarbenjamin, line 145: Can we just get rid of this? This should really just be expand or something rather than rewrite.

sympy/core/basic.py
JSS95: Classes in stats module still has `None` in `args` so I cannot change this in this PR.

sympy/physics/quantum/tensorproduct.py
JSS95: That should be done in another PR, I suppose.

sympy/core/basic.py
oscarbenjamin: Which classes? I really think that we should fix things like that rather than have workarounds in core Basic methods. The workarounds just facilitate broken subclasses. It's much better if those subclasses blow up as soon as they're written.

JSS95: Test failures suggest that `Moment` and `CentralMoment` in `sympy.stats.symbolic_probability` module have `None` in args.

This workaround is in master branch as well. It will be nice if we fix the stats module and remove this in the future, but I think it should not be a blocker for this PR.

JSS95: Test failures suggest that `Moment` and `CentralMoment` in `sympy.stats.symbolic_probability` module have `None` in args.

JSS95: This workaround is in master branch as well because of this reason. Perhaps I can re-define rewrite method for these classes and remove the workaround for `Basic`?

JSS95: @oscarbenjamin 

I removed the Basic check from `Basic.rewrite()`.

oscarbenjamin: It's not actually that difficult to fix them e.g.:
```diff
diff --git a/sympy/stats/symbolic_probability.py b/sympy/stats/symbolic_probability.py
index c74207f2a9..6057e5d5d5 100644
--- a/sympy/stats/symbolic_probability.py
+++ b/sympy/stats/symbolic_probability.py
@@ -603,8 +603,9 @@ def __new__(cls, X, n, c=0, condition=None, **kwargs):
         n = _sympify(n)
         c = _sympify(c)
         if condition is not None:
-            condition = _sympify(condition)
-        return Expr.__new__(cls, X, n, c, condition)
+            return Expr.__new__(cls, X, n, c, _sympify(condition))
+        else:
+            return Expr.__new__(cls, X, n, c)
 
     def doit(self, **hints):
         if not is_random(self.args[0]):
```

JSS95: Done.

oscarbenjamin:
Looks good. Thanks!

PR #21473, opened 2021-05-15
fix: `expr_free_symbols` returns symbols

### LINKED ISSUE #21444: `expr_free_symbols` returns non-symbols
```python
>>> (2*exp(x)-3).expr_free_symbols
{x, 2, -3}
>>> [i.is_Symbol for i in _]
[True, False, False]
```

Non-symbols are being returned. Shouldn't the result just be `{x}`?

hyadav2k:
`expr_free_symbols`  used to return non-symbols i.e. elements of `AtomicExpr`, which leads to having non-symbols within the result.

Fixes #21444 

We have removed the `expr_free_symbols` property from `AtomicExpr` and added it to the `Symbol` class.

#### Other comments
Ouput after the PR:
```py
In [1]: (2*exp(x)-3).expr_free_symbols
Out[1]: {x}

In [2]: (4*y**z+3*x).expr_free_symbols
Out[2]: {x, y, z}
```

#### Release Notes

* core
  * Fixed a bug in `expr_free_symbols`.

### DIFF
--- a/sympy/core/tests/test_expr.py
+++ b/sympy/core/tests/test_expr.py
@@ -2124,5 +2124,12 @@ def test_non_string_equality():
     assert (x != BadRepr()) is True
 
 def test_expr_free_symbols():
-    assert (2*exp(x)-3).expr_free_symbols == {x}
-    assert (4*y+3*x).expr_free_symbols == {x, y}
+    from sympy import IndexedBase
+    assert pi.expr_free_symbols == set()
+    assert (x + 2*y).expr_free_symbols == {x, y}
+    assert Derivative(2*x, y).expr_free_symbols == {x}
+    i = Integral(2*x, (y, z))
+    assert i.expr_free_symbols == i.function.free_symbols
+    assert Function('f')(x + 4).expr_free_symbols == {x}
+    b = 4 * IndexedBase(x)[y]
+    assert b.expr_free_symbols == {b.args[1]}
--- a/sympy/core/tests/test_expr.py
+++ b/sympy/core/tests/test_expr.py
@@ -2122,3 +2122,7 @@ def test_non_string_equality():
 
     assert (x == BadRepr()) is False
     assert (x != BadRepr()) is True
+
+def test_expr_free_symbols():
+    assert (2*exp(x)-3).expr_free_symbols == {x}
+    assert (4*y+3*x).expr_free_symbols == {x, y}
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -3919,10 +3919,6 @@ class AtomicExpr(Atom, Expr):
     def _eval_nseries(self, x, n, logx, cdir=0):
         return self
 
-    @property
-    def expr_free_symbols(self):
-        return {self}
-
 
 def _mag(x):
     """Return integer ``i`` such that .1 <= x/10**i < 1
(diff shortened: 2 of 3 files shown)


sympy/core/tests/test_expr.py
```diff
@@ -2122,3 +2122,7 @@ def __repr__(self):
 
     assert (x == BadRepr()) is False
     assert (x != BadRepr()) is True
+
+def test_expr_free_symbols():
+    assert (2*exp(x)-3).expr_free_symbols == {x}
+    assert (4*y+3*x).expr_free_symbols == {x, y}
```
smichr, line 2128:
```suggestion
    from sympy import IndexedBase
    assert pi.expr_free_symbols == {}
    assert x.expr_free_symbols == {x}
    assert (x + 2*y).expr_free_symbols == {x, y}
    i = Integral(x, (y, z))
    assert i.expr_free_symbols == i.function.free_symbols
    assert Function('f')(x).expr_free_symbols == {x}
    b = IndexedBase(x)[y]
    assert b.expr_free_symbols == {b}
```

hyadav2k: @smichr, the below examples are already `True` for master:
```py
    assert x.expr_free_symbols == {x}
    assert i.expr_free_symbols == i.function.free_symbols
    assert Function('f')(x).expr_free_symbols == {x}
    assert b.expr_free_symbols == {b}
```
Perhaps we can add like
```py
    assert Derivative(2*x, y).expr_free_symbols == {x}
    i = Integral(2*x, (y, z))
    assert i.expr_free_symbols == i.function.free_symbols
    assert Function('f')(x + 4).expr_free_symbols == {x}
    b = 4 * IndexedBase(x)[y]
    assert b.expr_free_symbols == {b.args[1]}
```
Kindly share your opinion.
Thanks.

sympy/core/tests/test_expr.py
```diff
@@ -2122,3 +2122,14 @@ def __repr__(self):
 
     assert (x == BadRepr()) is False
     assert (x != BadRepr()) is True
+
+def test_expr_free_symbols():
+    from sympy import IndexedBase
+    assert pi.expr_free_symbols == set()
+    assert (x + 2*y).expr_free_symbols == {x, y}
+    assert Derivative(2*x, y).expr_free_symbols == {x}
```
smichr, line 2130:
```suggestion
    d = Derivative(2*x, y)
    assert d.expr_free_symbols == d.expr.free_symbols == {x}
```

sympy/core/tests/test_expr.py
```diff
@@ -2122,3 +2122,14 @@ def __repr__(self):
...
+    i = Integral(2*x, (y, z))
+    assert i.expr_free_symbols == i.function.free_symbols
+    assert Function('f')(x + 4).expr_free_symbols == {x}
+    b = 4 * IndexedBase(x)[y]
+    assert b.expr_free_symbols == {b.args[1]}
```
smichr, line 2135:
```suggestion
    b = IndexedBase(x)[y]
    assert b.expr_free_symbols == {b}
```

smichr:
Perhaps some documentation should be added to `expr_free_symbols` to indicate that some objects already have a way to focus on the expression part of the object and that `expr_free_symbols` is a shortcut to get the free symbols from that portion or, for objects that don't support such an expression-focusing property, a way to get free symbols from anywhere in the object that an expression appears as long as it is not in a non-expression node.

I'll have to revisit in which cases `expr_free_symbols` solves a problem. It seems like there is a better way to do this since I can't think of how having a set of symbols that is only in expression nodes is helpful.

PR #21479, opened 2021-05-17
adds degree limit to `primitive_element` and `minpoly`

hyadav2k:
We added a limit to `primitive_element` and `minpoly` functions based on the maximum degree of the polynomial.

We have added the limit argument in `minpoly` and added a mechanism to decide for the `EX` domain depending on the degree of the polynomial, which will be provided by the user, and if the maximum degree is greater than the limit, the domain will fallback to `EX `for `minpoly`.
While for `primitive_element`, we have set the pre-defined limit of `100` as of now.

#### Other comments
The working is as follows:
We try to calculate the maximum degree of the polynomial that the extension elements will generate, using the `minpoly_compose` return types. In general, we will try to calculate the exact degree, but mainly in functions like `sin`, `cos` etc., the degree reduces in general due to factoring of the polynomial results. Since we don't want to be calculating minimal polynomial and using slow method `factor` while calculating degree, we are using the maximum degree possible for that case, like for `sin`, it cannot be more than `q` for odd `q`'s, and `q / 2` for even `q`'s, where `q = sympify(c.q)`, `c` is in `sin(c*pi)`.

#### Release Notes

* polys
  * Added degree limit to `primitive_element` and `minpoly`.

### DIFF
--- a/sympy/polys/constructor.py
+++ b/sympy/polys/constructor.py
@@ -103,6 +103,9 @@ def _construct_algebraic(coeffs, opt):
     exts = list(ordered(exts))
 
     g, span, H = primitive_element(exts, ex=True, polys=True)
+    if (g, span, H) == (False, False, False):
+        return EX, None
+
     root = sum([ s*ext for s, ext in zip(span, exts) ])
 
     domain, g = QQ.algebraic_field((g, root)), g.rep.rep
@@ -365,6 +368,8 @@ def construct_domain(obj, **args):
 
     coeffs = list(map(sympify, coeffs))
     result = _construct_simple(coeffs, opt)
+    if result == (EX, None):
+        return _construct_expression(coeffs, opt)
 
     if result is not None:
         if result is not False:
--- a/sympy/polys/numberfields.py
+++ b/sympy/polys/numberfields.py
@@ -23,6 +23,7 @@ from sympy.polys.polyerrors import (
     IsomorphismFailed,
     NotAlgebraic,
     GeneratorsError,
+    DegreeLimitExceeded
 )
 from sympy.polys.polytools import (
     Poly, PurePoly, invert, factor_list, groebner, resultant,
@@ -627,7 +628,7 @@ def _minpoly_compose(ex, x, dom):
 
 
 @public
-def minimal_polynomial(ex, x=None, compose=True, polys=False, domain=None):
+def minimal_polynomial(ex, x=None, compose=True, polys=False, domain=None, limit=None):
     """
     Computes the minimal polynomial of an algebraic element.
 
@@ -708,6 +709,10 @@ def minimal_polynomial(ex, x=None, compose=True, polys=False, domain=None):
                               "domain %s" % (x, domain))
 
     if compose:
+        if limit is not None:
+            deg = get_max_degree([ex])
+            if deg > limit:
+                raise DegreeLimitExceeded("maximum degree of minpoly(%s) is greater than the limit %s" % (ex, limit))
         result = _minpoly_compose(ex, x, domain)
         result = result.primitive()[1]
         c = result.coeff(x**degree(result, x))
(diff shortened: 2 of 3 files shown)


hyadav2k:
> 1. It would be better if the caller of `minpoly` could set what the limit would be e.g. `limit=100`.
>
Sure, as of now:
```py
In [1]: minpoly(cos(13*pi/45)*sin(17*pi/43), limit=10)
DegreeLimitExceeded                       Traceback (most recent call last)
<ipython-input-1-a79c24423235> in <module>
----> 1 minpoly(cos(13*pi/45)*sin(17*pi/43), limit=10)

~\sympy\sympy\polys\numberfields.py in minimal_polynomial(ex, x, compose, polys, domain, limit)
    713             deg = get_max_degree([ex])
    714             if deg > limit:
--> 715                 raise DegreeLimitExceeded("maximum degree of minpoly(%s) is greater than the limit %s" % (ex, limit))
    716         result = _minpoly_compose(ex, x, domain)
    717         result = result.primitive()[1]

DegreeLimitExceeded: maximum degree of minpoly(sin(17*pi/43)*cos(13*pi/45)) is greater than the limit 10
```
>
> 2. We should avoid duplicating the code from other parts of `minpoly` for this. Maybe the functions like `minpoly_cos` should check `if degree > limit` or perhaps it could be done in the `choose_factor` function or somewhere else.
>
Sure, I was thinking what if we add an argument as `limit=None` to `_minpoly_compose`, which if `True`, then we can return the degree in the function only.
For e.g. in `minpoly_cos`:
```py
In [1]: a = minpoly(cos(13*pi/45))

In [2]: print(a)
4096*_x**12 - 12288*_x**10 + 512*_x**9 + 13824*_x**8 - 1152*_x**7 - 7168*_x**6 + 864*_x**5 + 1680*_x**4 - 248*_x**3 - 144*_x**2 + 24*_x + 1

In [3]: minpoly(cos(13*pi/45), limit = 100)
Out[3]: 1
```
>

@oscarbenjamin, kindly share your opinion.
Thanks.

hyadav2k:
> 4. Possibly the right value for the limit is different for `minpoly` rather than `primitive_element`.
>
Yes, maybe as you instructed, we should let the user provide `limit` for `minpoly`, while for `primtive_element`, I was wondering why not use `_minpoly_compose` as for single radical it will not be stuck, perhaps we can be passing `limit` while calling it, and can handle the exception as well, as we can go for the predefined value as specified in `get_rad_deg` if error raised. Moreover, The passing `limit` could be taken from `get_rad_deg`,  which means, there will be mostly no exception raised.

I am proposing like for a complex radical `cos(13*pi/45)*sin(17*pi/43)`, we can split it as `minpoly(cos(13*pi/45), limit=a)` and `minpoly(sin(17*pi/43), limit=b)`, and the values can be multiplied. Apparently, the product obtained will not be the actual degree of the minimal polynomial of the radical `cos(13*pi/45)*sin(17*pi/43)`, but it will be far more accurate than just multiplying the maximal possible degrees of `cos(13*pi/45)` and `sin(17*pi/43)` and using it as max_deg for the radical.
P.S. `a = get_rad_deg(cos(13*pi/45))`, and `b = get_rad_deg(sin(17*pi/43))`

For e,g,:
```py
In [1]: from sympy.polys.numberfields import get_rad_deg, get_max_degree

In [2]: get_max_degree([cos(13*pi/45)*sin(17*pi/43)])
Out[2]: 924

In [3]: # k = minpoly(cos(13*pi/45)*sin(17*pi/43))

In [4]: d1 = minpoly(cos(13*pi/45), polys=True).degree()

In [5]: d1
Out[5]: 12

In [6]: d2 = minpoly(sin(17*pi/43), polys=True).degree()

In [7]: d2
Out[7]: 42

In [8]: d1*d2
Out[8]: 504
```
So `d1*d2` is far less than the degree obtained from `get_max_degree`.
>

> 5. We should think about what should be used for `Poly(..., extension=True)` to get the equivalent of `limit=100` or whatever the default value for limit would be.
>
We can get all the radicals of the Poly and then simply go ahead by calling `primitve_element`.
Will go through it more, and will get back to you.

oscarbenjamin:
This shouldn't duplicate all of the code for checking for different kinds of algebraic expression. Instead the DegreeLimit exception should be raised at the point in the code that `factor_list` is about to be called.

There could maybe be a helper function that checks the limit and raises DegreeLimit or otherwise calls `factor_list` and `_choose_factor` (since these two functions seemed to be called together a lot).

PR #13452, opened 2017-10-15
Fixed bugs in Subs with Indexed

Upabjojr:
`Subs` substitution now supports derivative by indexed objects in various cases.

PR #21485, opened 2021-05-18
Revert `expr_free_symbols`

### LINKED ISSUE #21444: `expr_free_symbols` returns non-symbols

smichr:

closes #21473
closes #21444

This removes the property `expr_free_symbols` and solves the Subs differentiation problem using existing tools.

#### Other comments

More specific comments for removal at #21494
#### Release Notes

* core
  * removed expr_free_symbols property

### DIFF
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -2394,17 +2394,29 @@ class Subs(Expr):
 
     def _eval_derivative(self, s):
         # Apply the chain rule of the derivative on the substitution variables:
-        val = Add.fromiter(p.diff(s) * Subs(self.expr.diff(v), self.variables, self.point).doit() for v, p in zip(self.variables, self.point))
+        f = self.expr
+        vp = V, P = self.variables, self.point
+        val = Add.fromiter(p.diff(s)*Subs(f.diff(v), *vp).doit()
+            for v, p in zip(V, P))
 
-        # Check if there are free symbols in `self.expr`:
-        # First get the `expr_free_symbols`, which returns the free symbols
-        # that are directly contained in an expression node (i.e. stop
-        # searching if the node isn't an expression). At this point turn the
-        # expressions into `free_symbols` and check if there are common free
-        # symbols in `self.expr` and the deriving factor.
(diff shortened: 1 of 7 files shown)


sympy/core/function.py
```diff
@@ -2394,17 +2389,30 @@ def _eval_subs(self, old, new):
...
+
+        # these are all the free symbols in the expr
+        efree = f.free_symbols
+        # some symbols like IndexedBase include themselves and args
+        # as free symbols
+        complex = {i for i in efree if len(i.free_symbols) > 1}
+        # hide them and see what independent free symbols remain
+        dums = {Dummy() for i in complex}
+        reps = {i: Dummy() for i in complex}
+        masked = f.subs(dict(zip(complex, dums)))
```
smichr, line 2405:
```suggestion
        masked = f.xreplace(dict(zip(complex, dums)))
```

smichr:
ping @asmeurer @jksuom @oscarbenjamin

oscarbenjamin:
I have never heard of this property before. I agree that it should be removed.

I don't think that a deprecation is needed if removing this fixes something. There are no new tests here though so does this fix something or is it just a clean up?

smichr:
> does this fix something or is it just a clean up

The only fix is to not use a new property to solve the issue involving differentiation of Subs (#13452). And there were lots of tests added at that time which helped to check the solution that is here now.

smichr:
I translated the previous behavior into the current behavior of the the routine to evaluate `Subs.diff` and added explanation at each step...I'm not convinced it is right, however. It passes tests but I'm not sure there might be another test that would could it to fail.

oscarbenjamin:
I think we should add a DeprecationWarning. If the method is unused then leaving it there with a warning isn't going to cause any harm. I'm sure someone somewhere is using it for something...

smichr:
ping @oscarbenjamin -- does this look right for the deprecation?

PR #21487, opened 2021-05-18
Ensure finite order before expanding bessel functions

### LINKED ISSUE #21486: expand_func(besselj(oo, x)) -> RecursionError
```
from sympy import *
from sympy.abc import x
expand_func(besselj(oo, x))
...
 line 86, in _eval_expand_func
    if (nu - 1).is_extended_positive:
...
RecursionError: maximum recursion depth exceeded in comparison
```

ehren:

Fixes #21486 

Use is_positive/negative (which now implies finite) to guard against infinite recursion in bessel `expand_func` with e.g. `besselj(oo, x)`

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/functions/special/bessel.py
+++ b/sympy/functions/special/bessel.py
@@ -82,11 +82,11 @@ class BesselBase(Function):
 
     def _eval_expand_func(self, **hints):
         nu, z, f = self.order, self.argument, self.__class__
-        if nu.is_extended_real:
-            if (nu - 1).is_extended_positive:
+        if nu.is_real:
+            if (nu - 1).is_positive:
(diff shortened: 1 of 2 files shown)


ehren:
Might as well change the nearby `is_extended_real` guard to `is_real` as well (done in forcepush).

oscarbenjamin:
I guess this should have some sort of limit for finite cases as well e.g. `expand_func(besselj(10000, x))`

smichr:
> I guess this should have some sort of limit for finite cases as well

This is probably another issue; there aren't such limits for expansion of trig function, however. What has been presented here is an obvious case needing to not expand (forever) so let's commit this as is.

PR #21124, opened 2021-03-19
Refactor Exact ODE solver into a class

habitzreuter:
#### References to other Issues or PRs
Continuing PR #18887 to refactor the exact ODE solver into a class, as suggested in #18348. 

#### Brief description of what is fixed or changed
The work done by @tnzl was was rebased to master. As suggested in the discussion of that PR, the calls to simplify were replaced by more specific functions. The try/except was also removed: the test case fixed by it passed even without this code block and none of the exact and inexact equations tested raised an error. The class docstring was replaced by the original docstring from the ode_1st_exact function.

#### Other comments
I'm in doubt about the `m != 0` conditional. If `m = 0` we can still have an exact equation (although it will be a trivial one): shouldn't `Eq(f(x).diff(x), 0)` (`m = 0, n = 1`) also be classified as `1st_exact`? If this is the case, we could simplify the verify function by removing this conditional and using just the numerator:
```
m, n = self.wilds_match()
m = m.subs(fx,y)
n = n.subs(fx,y)
numerator = cancel(m.diff(y) - n.diff(x))
if numerator.is_zero:
    # It's exact
    return True
else:
    # Try to convert to exact
```

#### Release Notes

* solvers
  * Refactor exact EDO solving into the FirstExact class.

PR #21492, opened 2021-05-19
Add separate flags for xfail raise and xfail wrong

Mohitbalwani26:

SEE [suugestion](https://github.com/sympy/sympy/pull/21124#issuecomment-842561844)

This PR adds two flags for failed examples.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/solvers/ode/tests/test_single.py
+++ b/sympy/solvers/ode/tests/test_single.py
@@ -7,8 +7,8 @@ This File contains test functions for the individual hints used for solving ODEs
 
 Examples of each solver will be returned by _get_examples_ode_sol_name_of_solver.
 
-Examples should have a key 'XFAIL' which stores the list of hints if they are
-expected to fail for that hint.
+Examples should have a key 'XFAIL_raise' which stores the list of hints if they are
+expected to raise an error for that hint.
 
 Functions that are for internal use:
 
@@ -149,7 +149,8 @@ def _add_example_keys(func):
             temp={
                 'eq': solver['examples'][example]['eq'],
                 'sol': solver['examples'][example]['sol'],
-                'XFAIL': solver['examples'][example].get('XFAIL', []),
+                'XFAIL_raise': solver['examples'][example].get('XFAIL_raise', []),
+                'XFAIL_wrong': solver['examples'][example].get('XFAIL_wrong', []),
                 'func': solver['examples'][example].get('func',solver['func']),
                 'example_name': example,
                 'slow': solver['examples'][example].get('slow', False),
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
I think @habitzreuter wanted to add the example from https://github.com/sympy/sympy/pull/21124#issuecomment-841699215 as an XFAIL test.

Can that be added here now?

habitzreuter:
> I think @habitzreuter wanted to add the example from [#21124 (comment)](https://github.com/sympy/sympy/pull/21124#issuecomment-841699215) as an XFAIL test.
> 
> Can that be added here now?

This would work if the example I suggested returned a single wrong result. From [here](https://github.com/sympy/sympy/pull/21124#discussion_r602896976) we see that the first element in the solution array is garbage (`False`), while `checkodesol` expects an equality. If we add the `'XFAIL_wrong': ['1st_exact']` to the test I suggested, we get:
```python
  File "/home/marco/src/sympy/sympy/solvers/ode/subscheck.py", line 143, in checkodesol
    solved = sol.lhs == func and not sol.rhs.has(func)
AttributeError: 'BooleanFalse' object has no attribute 'lhs'
```

Maybe `checkodesol` should raise an exception if `sol` is neither Equality nor Expr. Then we use a `try:` with `checkodesol` at `_test_particular_example`.

But then it seems we would need both `XFAIL_raise` and `XFAIL_wrong` for this test, since the second solution in the array is an Equality with the wrong result.

Mohitbalwani26:
I think that case is too specific. can we mark it as `XFAIL_WRONG` and `checkodesol_XFAIL` and add comment over it why they are marked like this.
In my opinion if we introduce one more flag like `checkodesol_RAISE` it wouldn't be good to maintain so many flags.
what's you opinion on this @oscarbenjamin ?

oscarbenjamin:
We need to have XFAIL tests for cases where `dsolve` returns a wrong result. Generally the best way to do this is like:
```python
@XFAIL
def test_dsolve_wrong_result():
    wrong_sol = ...
    dsolve_sol = dsolve(...)
    assert dsolve_dol != wrong_sol
```
That way when the output of `dsolve` changed for this case the test will go from XFAIL to XPASS and then someone can look at it and see whether the new result returned by `dsolve` is correct. If so then the test can be updated.

Maybe the framework for the dsolve tests can be changed to handle wrong results from dsolve like that.

> In my opinion if we introduce one more flag like `checkodesol_RAISE` it wouldn't be good to maintain so many flags.

It shouldn't be hard to maintain a few flags. Probably it would be easier if the code that handled the flags wasn't so messy (i.e. it can be tidied up).

Mohitbalwani26:
> We need to have XFAIL tests for cases where `dsolve` returns a wrong result. Generally the best way to do this is like:
> ```python
> @XFAIL
> def test_dsolve_wrong_result():
>     wrong_sol = ...
>     dsolve_sol = dsolve(...)
>     assert dsolve_dol != wrong_sol
> ```

Wouldn't this be same as it was earlier? 
If we want to have functions like this for xfail then we would have to write examples inside it. As these can't be reused by passing arguments to it. 

To make the code cleaner we can also make small helper functions like for testing xfail result or checkodesol conditions.

oscarbenjamin:
> Wouldn't this be same as it was earlier?

The example can still be included in the test examples "database" format but there needs to be a separate XFAIL function for testing each example so that we can see the XFAIL/XPASS in the output from bin/test or pytest.

PR #21496, opened 2021-05-20
fix(matrices): Deprecate non-Expr in Matrix

oscarbenjamin:

The deprecation issue for this is #21497
The deprecation issue for the change to `laplace_transform` is https://github.com/sympy/sympy/issues/21504

Follows on from #21441, #21427, #21402

This PR deprecates using non-Expr objects as the elements of a Matrix. The codebase previously contained a number of Matrix subclasses (`NewMatrix`, `RawMatrix` and `PolyMatrix`) that would use non-Expr elements for the entries. In most cases this was done so that the elements could be domain elements or Poly rather than Expr but a better solution to this is to use the new `DomainMatrix` class. After various other PRs it is now possible to deprecate having non-Expr elements in a Matrix.

One notable obstacle to refusing non-Expr elements is the behaviour of the `laplace_transform` function:
```python
In [2]: laplace_transform(eye(2), t, z)
Out[2]: 
⎡⎛1         ⎞                ⎤
⎢⎜─, 0, True⎟   (0, -∞, True)⎥
⎢⎝z         ⎠                ⎥
⎢                            ⎥
⎢               ⎛1         ⎞ ⎥
⎢(0, -∞, True)  ⎜─, 0, True⎟ ⎥
⎣               ⎝z         ⎠ ⎦
```
This has returned a Matrix where each element is a tuple consisting of an Expr and the convergence conditions.In this PR the above Matrix of tuples is deprecated and it is possible to obtain a new behaviour:
```python
In [2]: laplace_transform(eye(2), t, z, legacy_matrix=False)
Out[2]: 
⎛⎡1   ⎤         ⎞
⎜⎢─  0⎥         ⎟
⎜⎢z   ⎥         ⎟
⎜⎢    ⎥, 0, True⎟
⎜⎢   1⎥         ⎟
⎜⎢0  ─⎥         ⎟
⎝⎣   z⎦         ⎠
```
This returns a tuple of the transformed matrix and the convergence conditions for the matrix as a whole and is more consistent with the general behaviour of the `laplace_transform` function for other kinds of inputs. A future PR will make this behaviour the default.

#### Other comments

I plan to use `DomainMatrix` as the internal implementation of `Matrix` but that means that `Matrix` needs to be stricter about the types of its elements. Needing to accommodate for non-Expr elements in the internal routines of `Matrix` has constrained improvements to this important class for some time.

#### Release Notes

* matrices
    * DEPRECATION: Using non-Expr elements in a Matrix is now deprecated.
* integrals
    * DEPRECATION: Using `laplace_transform` with a `Matrix` and without passing `noconds=True` is now deprecated. The behaviour of this function will change in future to return the conditions separately from the Matrix. The new behaviour can be obtained by passing the keyword argument `legacy_matrix=False`.

### DIFF
--- a/sympy/integrals/transforms.py
+++ b/sympy/integrals/transforms.py
@@ -13,6 +13,7 @@ from sympy.integrals.meijerint import _dummy
 from sympy.logic.boolalg import to_cnf, conjuncts, disjuncts, Or, And
 from sympy.simplify import simplify
 from sympy.utilities import default_sort_key
+from sympy.utilities.exceptions import SymPyDeprecationWarning
 from sympy.matrices.matrices import MatrixBase
 
 
@@ -1145,7 +1146,7 @@ class LaplaceTransform(IntegralTransform):
         return plane, cond
 
 
-def laplace_transform(f, t, s, **hints):
+def laplace_transform(f, t, s, legacy_matrix=True, **hints):
     r"""
     Compute the Laplace Transform `F(s)` of `f(t)`,
 
@@ -1168,6 +1169,14 @@ def laplace_transform(f, t, s, **hints):
     :func:`sympy.integrals.transforms.IntegralTransform.doit`. If ``noconds=True``,
     only `F` will be returned (i.e. not ``cond``, and also not the plane ``a``).
 
+    .. deprecated:: 1.9
+        Legacy behavior for matrices where ``laplace_transform`` with
+        ``noconds=False`` (the default) returns a Matrix whose elements are
+        tuples. The behavior of ``laplace_transform`` for matrices will change
+        in a future release of SymPy to return a tuple of the transformed
+        Matrix and the convergence conditions for the matrix as a whole. Use
+        ``legacy_matrix=False`` to enable the new behavior.
+
     Examples
     ========
 
@@ -1181,13 +1190,29 @@ def laplace_transform(f, t, s, **hints):
 
     inverse_laplace_transform, mellin_transform, fourier_transform
     hankel_transform, inverse_hankel_transform
+
     """
 
     if isinstance(f, MatrixBase) and hasattr(f, 'applyfunc'):
-        elements_trans = [laplace_transform(fij, t, s, **hints) for fij in f]
-        elements, avals, conditions = zip(*elements_trans)
-        f_laplace = type(f)(*f.shape, elements)
-        return f_laplace, Max(*avals), And(*conditions)
+
+        conds = not hints.get('noconds', False)
+
+        if conds and legacy_matrix:
+            SymPyDeprecationWarning(
+                feature="laplace_transform of a Matrix with noconds=False (default)",
+                useinstead="the option legacy_matrix=False to get the new behaviour",
+                issue=21504,
+                deprecated_since_version="1.9"
+            ).warn()
+            return f.applyfunc(lambda fij: laplace_transform(fij, t, s, **hints))
+        else:
+            elements_trans = [laplace_transform(fij, t, s, **hints) for fij in f]
+            if conds:
+                elements, avals, conditions = zip(*elements_trans)
+                f_laplace = type(f)(*f.shape, elements)
+                return f_laplace, Max(*avals), And(*conditions)
+            else:
+                return type(f)(*f.shape, elements_trans)
 
     return LaplaceTransform(f, t, s).doit(**hints)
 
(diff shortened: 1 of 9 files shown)


sympy/matrices/tests/test_matrices.py
```diff
@@ -656,9 +656,11 @@ def test_creation():
         Matrix((1, 2))[3] = 5
 
     assert Matrix() == Matrix([]) == Matrix([[]]) == Matrix(0, 0, [])
-    # anything can go into a matrix (laplace_transform uses tuples)
```
smichr, line 659: Does `laplace_transform` still do this?

sympy/matrices/dense.py
```diff
@@ -313,10 +315,21 @@ def _new(cls, *args, copy=True, **kwargs):
         else:
             rows, cols, flat_list = cls._handle_creation_inputs(*args, **kwargs)
             flat_list = list(flat_list) # create a shallow copy
+
+            types = set(map(type, flat_list))
+            if not all(issubclass(typ, Expr) for typ in types):
+                SymPyDeprecationWarning(
+                    feature="non-Expr objects in a Matrix",
+                    useinstead="list of lists or some other data structure",
```
smichr, line 323: perhaps we could advertise `TableForm` here; and maybe slicing of TableForm could be added.

oscarbenjamin: That's a good point. I've mentioned TableForm in the message

sympy/integrals/transforms.py
```diff
@@ -1181,8 +1182,13 @@ def laplace_transform(f, t, s, **hints):
     inverse_laplace_transform, mellin_transform, fourier_transform
     hankel_transform, inverse_hankel_transform
     """
+
     if isinstance(f, MatrixBase) and hasattr(f, 'applyfunc'):
-        return f.applyfunc(lambda fij: laplace_transform(fij, t, s, **hints))
+        elements_trans = [laplace_transform(fij, t, s, **hints) for fij in f]
+        elements, avals, conditions = zip(*elements_trans)
+        f_laplace = type(f)(*f.shape, elements)
+        return f_laplace, Max(*avals), And(*conditions)
```
smichr, line 1190: This can't be done until the deprecation is complete, can it?

oscarbenjamin: This is just fixing something that is completely broken. On master we have:
```python
In [1]: M = Matrix([[t, 0], [0, t]])

In [2]: t, s = symbols('t, s')

In [3]: M = Matrix([[t, 0], [0, t]])

In [4]: laplace_transform(M, t, s)
Out[4]: 
⎡⎛1          ⎞               ⎤
⎢⎜──, 0, True⎟  (0, -∞, True)⎥
⎢⎜ 2         ⎟               ⎥
⎢⎝s          ⎠               ⎥
⎢                            ⎥
⎢               ⎛1          ⎞⎥
⎢(0, -∞, True)  ⎜──, 0, True⎟⎥
⎢               ⎜ 2         ⎟⎥
⎣               ⎝s          ⎠⎦
```
That's a Matrix but with each element being a tuple.

The PR gives:
```python
In [3]: laplace_transform(M, t, s)
Out[3]: 
⎛⎡1     ⎤         ⎞
⎜⎢──  0 ⎥         ⎟
⎜⎢ 2    ⎥         ⎟
⎜⎢s     ⎥         ⎟
⎜⎢      ⎥, 0, True⎟
⎜⎢    1 ⎥         ⎟
⎜⎢0   ──⎥         ⎟
⎜⎢     2⎥         ⎟
⎝⎣    s ⎦         ⎠
```
IOW the PR returns the matrix plus the conditions. On master you get a matrix of tuples with the conditions as part of the tuples in the elements of the matrix. That's so broken that I don't think any deprecation is needed.

smichr: It's not broken for the person that knows this and looks at a given matrix element's tuple to get the condition(s). Not that I don't want to see the change -- I think it makes much more sense like this -- but this is backwards incompatible for anyone that knows and handles this already. So should the cycle be 1) we need a keyword that allows the new output but default to the old. Then when the deprecation occurs, 2) the keyword has to default to new style and be put into deprecation, and finally 3) the keyword gets dropped.

oscarbenjamin: I'm not sure this needs a deprecation cycle but I've added it in.

I realised when looking at it that actually on master reasonable results can be obtained when using `noconds=False` but this PR broke that behaviour. I've fixed that and added tests for it.

oscarbenjamin:
Thanks!

I've update the OP here and also for the related deprecation issues.

PR #20875, opened 2021-01-29
printing: fix latex, pretty to properly render unevaluated Mul in Add

AaronStiff:
#### References to other Issues or PRs

Fixes #19889

#### Brief description of what is fixed or changed
Updated `latex` and `pretty` so that they do not simplify unevaluated Muls with only Integers when a part of an Add. See linked issue for example.

#### Other comments
I've slightly modified the only test which failed as a result of this change. That test was originally intended to test whether or not `pretty` was automatically evaluating Muls (see pull request #8305). Ironically enough, this change causes it to more accurately preserve unevaluated Muls. However, I would still appreciate feedback on whether this is an acceptable modification.

#### Release Notes

* printing
  * Fix both latex and pretty to properly render unevaluated Muls when inside an Add

PR #21499, opened 2021-05-21
fix issue in the parse_latex by not evaluating multiplication with -1

gxyd:
Before:
```
>>> latex_string = "5^0 - 4^0"
>>> parse_latex(latex_string)
−1+5^0
```

Now
```
>>> latex_string = "5^0 - 4^0"
>>> r = parse_latex(latex_string)
>>> print(r)
−4^0+5^0
>>> srepr(r)
_Add(Mul(Integer(-1), Pow(Integer(4), Integer(0),
			  evaluate=0),
	 evaluate=False),
     Pow(Integer(5), Integer(0),
	 evaluate=False),
     evaluate=False)
```

Link to a chat message on gitter: https://gitter.im/sympy/sympy?at=60a3e9b12ee2b13d95b7c3a1

Solution as suggested by @jksuom in the above gitter chat.

#### Other comments

* parsing
  * Fixed a bug in `parse_latex` function for rendering of `-1` Mul to not evaluate in Add operation

### DIFF
--- a/sympy/parsing/latex/_parse_latex_antlr.py
+++ b/sympy/parsing/latex/_parse_latex_antlr.py
@@ -119,7 +119,8 @@ def convert_add(add):
     elif add.SUB():
         lh = convert_add(add.additive(0))
         rh = convert_add(add.additive(1))
-        return sympy.Add(lh, -1 * rh, evaluate=False)
+        return sympy.Add(lh, sympy.Mul(-1, rh, evaluate=False),
+                         evaluate=False)
     else:
         return convert_mp(add.mp())
 
--- a/sympy/parsing/tests/test_latex.py
+++ b/sympy/parsing/tests/test_latex.py
@@ -242,6 +242,7 @@ GOOD_PAIRS = [
     (r"\int x \, dx", Integral(x, x)),
     (r"\log_2 x", _log(x, 2)),
     (r"\log_a x", _log(x, a)),
+    (r"5^0 - 4^0", _Add(_Pow(5, 0), _Mul(-1, _Pow(4, 0))))
 ]
 
 def test_parseable():


sympy/parsing/latex/_parse_latex_antlr.py
```diff
@@ -119,7 +119,8 @@ def convert_add(add):
     elif add.SUB():
         lh = convert_add(add.additive(0))
         rh = convert_add(add.additive(1))
-        return sympy.Add(lh, -1 * rh, evaluate=False)
```
smichr, line 122: I suggest that you use `sympy.core.mul._keep_coeff` so you don't get an unecessary unevaluated Mul for when `rh` is a Rational or a product with a leading Rational like `2*x`. Do either of these cases apply here?

jksuom: Maybe with `sign=True`.

smichr: I was wondering...but I think you would want `1 - 2x` to appear like that instead of as `1 + (-1)*2*x` ... not sure what will happen.

jksuom:
I could not find documentation on how `parse_latex` is expected to behave but this PR seems to better preserve the original latex text.

Current master:
```
>>> latex(parse_latex('x - (y + z)'))
'x + \\left(- y - z\\right)'
```
This PR:
```
>>> latex(parse_latex('x - (y + z)'))
'x - \\left(y + z\\right)'
```

sympy/parsing/tests/test_latex.py
```diff
@@ -242,6 +242,7 @@ def test_import():
     (r"\int x \, dx", Integral(x, x)),
     (r"\log_2 x", _log(x, 2)),
     (r"\log_a x", _log(x, a)),
+    (r"5^0 - 4^0", _Add(_Pow(5, 0), _Mul(-1, _Pow(4, 0))))
```
gxyd, line 245: @asmeurer 's comment [from gitter](https://gitter.im/sympy/sympy?at=60a80aad863ba917ce49dc66)
```
parse_latex isn't designed to do evaluate=False in its present form. 
```

But isn't that how we test things for `parse_latex`? All of the functions with `_` are basically unevaluated form of those.

gxyd:
> I could not find documentation on how parse_latex is expected to behave

Yeah, I think we should probably have that as well.

smichr:
@oscarbenjamin , what was the other recent issue that identified a bug in printing where the unevalulated Mul is not being rendered as such -- didn't it involve hiding the -1 after a 1 as in (1, -1)?

oscarbenjamin:
> @asmeurer, @oscarbenjamin , what was the other recent issue that identified a bug in printing where the unevalulated Mul is not being rendered as such -- didn't it involve hiding the -1 after a 1 as in (1, -1)?

I don't remember. I think there was a PR that tried to make unevaluated Muls print like evaluated Muls where I objected about something like that. But I'm not sure there was a bug in master.

smichr:
> I don't remember

#20875

sympy/parsing/latex/_parse_latex_antlr.py
smichr: The printer works around it but it is parsed as `1 + (-1)*2*x`; a `convert_mul` analogous to `convert_add` has to be added.

sympy/parsing/tests/test_latex.py
smichr: This is just testing that the output of parsing is the same as the sympy expression -- and in this case what is equivalent to the unevaluated expression is obtained.

PR #21500, opened 2021-05-22
perf(matrices): make randMatrix faster for sparse matrices

oscarbenjamin:

Makes randMatrix faster for sparse matrices e.g.:
```python
In [1]: %time ok = randMatrix(1000, percent=1)
CPU times: user 139 ms, sys: 5.56 ms, total: 145 ms
Wall time: 144 ms
```
On master it is much slower:
```python
In [1]: %time ok = randMatrix(1000, percent=1)
CPU times: user 6.69 s, sys: 48 ms, total: 6.74 s
Wall time: 6.79 s
```
Also  a slow test was marked as `@slow` because it takes ~3 minutes

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/matrices/dense.py
+++ b/sympy/matrices/dense.py
@@ -1183,32 +1183,31 @@ def randMatrix(r, c=None, min=0, max=99, seed=None, symmetric=False,
     [70,  0,  0],
     [ 0,  0, 88]
     """
-    if c is None:
-        c = r
     # Note that ``Random()`` is equivalent to ``Random(None)``
     prng = prng or random.Random(seed)
 
-    if not symmetric:
-        m = Matrix._new(r, c, lambda i, j: prng.randint(min, max))
-        if percent == 100:
-            return m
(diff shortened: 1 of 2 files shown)


smichr:
I notice that this (and the old version) are not assured to fill to the desired percent:
```python
>>> len([i for i in randMatrix(10,symmetric=1,percent=50,min=1) if i])
56
>>> len([i for i in randMatrix(10,symmetric=1,percent=50,min=1) if i])
44
```
Note that the min was 1 so non-zero values should have been mapped into 50 slots.

We need to map half of the desired percentage into the lower half of the symmetric matrix. This isn't a mod c calculation, however.

smichr:
> should have been mapped into 50 slots.

ahh - but some hit the diagonal

smichr:
> This isn't a mod c calculation, however.

about half of the selected indices will have r <= c and then these are duplicated to the upper half of the matrix. I either case (symmetric or not) this is only an approximate filling of the matrix but for a random matrix this should be good enough.

PR #21510, opened 2021-05-23
Add Minisat22 solver to Sympy's satisfiable algorithms

migueltorrescosta:
# Add Minisat22 solver to Sympy's satisfiable algorithms

Sympy's satisfiable has 3 supported algorithms (dpll, dpll2 and pycosat), none of which support SAT solvers specialized in finding minimal solutions, in the sense of having the least amount of true statements.
This PR aims to fill this gap.

#### Other comments

- **Existing unmaintained repo:** The current implementation has support for pycosat via a dependence on the [PycoSat repo](https://github.com/ContinuumIO/pycosat) . This has had no activity from the maintainers in 4 years, looking like a stale implementation.  [PySAT's repo](https://github.com/pysathq/pysat) provides more functionality and the maintainers are more active. I added the option to use PySAT's Minisat22 algorithm, but it might be worth removing the dependency on pycosat completely.
- **Docs:** The current docs don't mention the specific algorithms available via satisfiable. As such no changes were needed to the docs
- **New files:** Sympy's bot warned me about new files. These were added intentionally to cover the pySat Wrapper.

#### Release Notes

* solvers
  * Added a new solver, Minisat22, for the satisfiability problem


smichr:
Some tests should be added to know that this works, otherwise I'm not sure a passing run will indicate much.

migueltorrescosta:
Tests added @smichr
If there's anything needed do ask, thank you for the prompt reply!

migueltorrescosta:
@smichr I fixed the whitespaces issue and ran all the tests. It looks fine except for the message `7 expected to fail but passed`. Since these 7 tests are not at all related with what I touched I believe they were passing before, even though they were expected to fail.

```bash
______________________________ xpassed tests _______________________________________
sympy/concrete/tests/test_sums_products.py: test_issue_14313
sympy/core/tests/test_args.py: test_sympy__diffgeom__diffgeom__BaseCovarDerivativeOp
sympy/polys/tests/test_partfrac.py: test_apart_extension_xfail
sympy/sets/tests/test_powerset.py: test_failing_powerset__contains__
sympy/stats/tests/test_rv.py: test_samplingE
sympy/tensor/tests/test_indexed.py: test_Idx_inequalities_current_fails
sympy/utilities/tests/test_pickling.py: test_printing3
```

I believe all is ready now

migueltorrescosta:
Similar to how pycosat is imported, I added the python-sat dependency to:

- bin/test_external_imports.py
- setup.cfg
- sympy/logic/inference.py (via import_module("pysat")

However when running the tests locally they all passed AFTER I did `pip install pycosat python-sat`. If you are familiar with working with optional dependencies I would appreciate any guidelines. Having the package installed as `python-sat` but called as `pysat` doesn't help. 

I haven't yet, but in the meantime I will try to figure it out :)

migueltorrescosta:
I am happy to see all checks pass. Are there any extra steps I can take to have this PR merged?

AUTHORS
```diff
@@ -1050,3 +1050,4 @@ Rodrigo Luger <rodluger@gmail.com>
 Marco Antônio Habitzreuter <mahabitzreuter@gmail.com>
 Ayush Bisht <bisht.ayush2001@gmail.com>
 Akshansh Bhatt <akshansh@tuta.io>
+migueltorrescosta <miguelptcosta1995@gmail.com>
```
asmeurer, line 1053: This file is automatically generated from `./bin/authors_update.py`. If you want to add yourself, the best thing you can do would be to add an entry in .mailmap mapping your full name, and your name will be added here the next time we run this script (you can also run the script yourself but it hasn't been run in a while so a bunch of other people's entries will need to be adjusted if you do).

asmeurer:
What is the benefit of having a minimal solution?

smichr:
cf #8029

migueltorrescosta:
In cases where booleans represent changes to a system and the clauses represent our desired goals, minimising the number of changes reduces the risk of unforeseen consequences. This approach is common in Control Theory.

I started this PR as I will be using this minimal solver at work. We considered using PycoSat directly, but due to my overall satisfaction with Sympy's logic methods I was hoping to instead contribute this option to Sympy.

AUTHORS
migueltorrescosta: On it. I ran `./bin/authors_update.py` first, but as it added a bunch of people it felt like it wasn't doing what was expected

migueltorrescosta:
@asmeurer @smichr  I am not familiar with your Merge Approval process. If there is anything I can do to get this update into Master / Release I would be happy to do so.

oscarbenjamin:
Can you explain what it means for `satisfiable` that the output is minimal? What are examples of minimal and non-minimal output from `satisfiable? Is this related to #8029?

migueltorrescosta:
> Can you explain what it means for `satisfiable` that the output is minimal? What are examples of minimal and non-minimal output from `satisfiable? Is this related to #8029?

The two issues can be made similar, but it isn't a straightforward connection to #8029 .

**Minimality**
The output of satisfiable, assuming symbols `a`, `b`, `c` and `d`, would be something like `{a: True, b: False, c: False, d:True }` . This solution can also be written as the set of `True` statements, `{a, d}`. A solution is minimal when, written in this form, no strict subset of it is also a solution.

**Example**
Let's say we have the expression `Or(x, And(y, z))`. This system has 5 solutions:
```python
{x: True, y:True, z: True }
{x: True, y:True, z: False }
{x: True, y:False, z: True }
{x: True, y:False, z: False }
{x: False, y: True, z: True}
```
which can be written as the sets `{x,y,z}, {x,y}, {x,z}, {x}, {y,z}` . From this the only minimal sets are `{x}, {y,z}`, as all other solutions contain these. Mapping to the original way of presenting the solution, the minimal solutions are
```python
{x: True, y:False, z: False }
{x: False, y: True, z: True}
```

**Why is this useful**
- In real life applications symbols can represent changes to a system ( increase room temperature, add oxygen to a system, add air filtration systems, ... ), with the clauses representing limitations ( we can't add butane and a source of light due to risk of explosion ). Not only is it useful to introduce the system, we might also be interested in minimizing the number of changes due to unforeseen consequences.
- In proof systems it could potentially help find proofs with the least amount of assumptions for a Theorem ( not really sure here though )

Is there any part you would like me to elaborate on @oscarbenjamin ?

oscarbenjamin:
The examples in the tests don't seem to be all "minimal" in the sense described above e.g.:
```
    assert minisat22_satisfiable(        A | B ) in ({A: True}, {B: False}, {A: False, B: True}, {A: True, B: True}, {A: True, B: False})
```
Shouldn't it be just `A:true, B:true`?

Why does the test use `in` rather than `==`?

I think I'm missing something...

migueltorrescosta:
It uses in because a minimum satisfiability solver does not necessarily need to always give the same answer, as some solvers incorporate randomness into their heuristics. As such I have to give room for variability in the answers, as is done in the picosat tests (my tests are mostly copy paste from picosat's solver ).

As for minimality, we can't ensure minimality until we find all solutions to compare. As such the solver is optimized for finding minimal solutions, but it isn't certain that all solutions found are minimal. The way this optimization works is as follows:

- Ensure that negative phases are favoured over positive ones. This is done in the line `r.set_phases([-(i+1) for i in range(r.nof_vars())]) `
- Ensure that we don't return solutions bigger than previously found ones, in the set containment sense. This is done by replacing the addition of clauses done in `results.add_clause([-i for i in sol])` with `results.add_clause([-i for i in sol if i>0])`. The last condition removes all solutions which are bigger than the one we just found.

oscarbenjamin:
Oh, I see. And when `minimal` is not set which is the default then this won't be any slower (for the purposes of minimality)?

I think that these changes look good.

I also think that we should drop support for pycosat if pysat provides the same functionality and pycosat is not currently maintained.

Does anyone have any views on dropping support for pycosat?

migueltorrescosta:
> Oh, I see. And when `minimal` is not set which is the default then this won't be any slower (for the purposes of minimality)?

Exactly, if `minimal=False` (which is the default value), the run time is the same as before, and it returns the usual SAT solver. When `minimal=True`, it applies the mentioned changes.
 
> I also think that we should drop support for pycosat if pysat provides the same functionality and pycosat is not currently maintained.

I agree, but I have only compared them as the end user, and found no differences yet. I can't argue about potentially different inner heuristics though. Maybe we should deprecate pycosat as a different issue from adding support for pySat?

oscarbenjamin:
> Maybe we should deprecate pycosat as a different issue from adding support for pySat?

Yes, it can be a separate issue/PR but I wanted to ask if anyone has any view on dropping support. I'm not familiar with either library so I'd like to hear if anyone else has anything to say before merging this.

asmeurer:
I'm not sure that pycosat is unmaintained. Anyway, I think we should have strong evidence that some other solver is always better before we deprecate any of them.

migueltorrescosta:
> I'm not sure that pycosat is unmaintained.

- The main argument in favour of [pycoSat](https://github.com/ContinuumIO/pycosat) still being maintained is that the maintainers are [Anaconda Inc](https://github.com/ContinuumIO), an active group.
- The main argument against it is that issues aren't being replied to, and even though there are open PRs, [the last code change happened in 2017](https://github.com/ContinuumIO/pycosat/commits/master) .

> Anyway, I think we should have strong evidence that some other solver is always better before we deprecate any of them.

Totally agree

asmeurer:
I know the author of pycosat. I can reach out to him and ask him about the maintenance status. I would hope that at the very least if he can't maintain it any more that he can move the maintenance to someone else.

asmeurer:
So the original author of pycosat no longer maintains it. However, Anaconda does still use pycosat for conda, so they should be maintaining it. I would suggest opening an issue in the repo asking for clarification if the project is maintained, and if it isn't maintained, requesting that they move the maintenance to someone who is willing to do so.

oscarbenjamin:
This PR does not remove support for pycosat but rather adds support for minisat22 which seems reasonable to me.

It is important to review our support for other libraries like pycosat though because I personally find that I spend what seems like wasted time trying to fix things up for sympy to work with some library that I don't know or care about without having any idea whether anyone else actually cares about sympy's support for that library either. From that perspective the primary question is not whether pycosat is maintained but whether anyone using sympy actually cares about pycosat support. We should not fall into the trap of thinking that supporting a library is "free" just because the code for it is already there: sympy has far too many optional dependencies and I think that few of them have any significant use and each one adds cost to general maintenance like CI.

asmeurer:
I generally agree with the sentiment, although that may be more of an argument for reconsidering this PR than reconsidering pycosat support. I personally think having both is fine, as it's actually pretty straightforward to support an external SAT solver. And if the solver is fast, it can make a significant performance improvement, including in the new assumptions which make use of the SAT solver. 

I personally would like to see pycosat remain maintained, regardless if SymPy uses it, which is also why I am concerned if it currently isn't. I should also note that pycosat is currently included with Anaconda, so anyone who obtains SymPy with that will automatically get a faster SAT solver.

At the end of the day, I think SymPy has not choice but to have optional dependencies for fast, C-based libraries, for performance. Things like gmpy2, SAT solvers, and FLINT, and even SymEngine are faster than we can ever achieve with pure Python, and also more work than we really want to do ourselves. Fortunately, of these, SAT solvers are by far the simplest possible library to add as a drop-in replacement in the backend. A typical SAT solver has an extremely simple API, which is generally only one function (with possibly some options) that takes in a very simple input (a list of lists of integers), and returns a very simple output (either a vector of booleans or UNSAT). Furthermore, it doesn't even matter if the exact output is the same, as long as it is correct. SAT solvers make no guarantee about which solution they give to satisfiable formula (except for things like "minimality" for libraries like the one here), and generally no code that uses `satisfiable` will ever make such an assumption about the solution. If you want the solution to satisfy some property, you typically encode that into the formula that you pass to the solver. Most SAT solvers are also standalone libraries, which make they relatively easy to install.

Contrast that with gmpy, which has dozens of API functions, or SymEngine, for which just representing expressions slightly differently can cause issue.

oscarbenjamin:
>  that may be more of an argument for reconsidering this PR than reconsidering pycosat support.

I don't think so. If the feature is that sympy provides users with an easy way to access a fast external library then we can do that without needing to promise continued support for libraries that are not maintained. Users who want to use newer versions of sympy can also install different dependencies that sympy can use to speed up the same computation.

The cost of continuing to "support" a library that is not maintained and also not used (via sympy or otherwise) is much higher than the cost of support something that is maintained and that users actually want to use. If users and maintainers value the sympy support then then they will help maintain it. If no one else cares then the burden falls purely on sympy maintainers to maintain something that is possibly not used by anyone.

> Contrast that with gmpy, which has dozens of API functions

The situation with gmpy is very different because it speeds up calculations with rational numbers which is absolutely fundamental in sympy. (AFAICT no particularly significant benefit comes from any other part of gmpy)

asmeurer:
gmpy has several integer functions that we are (or should be) using, like isqrt and gcd. 

> The cost of continuing to "support" a library that is not maintained and also not used (via sympy or otherwise) is much higher than the cost of support something that is maintained and that users actually want to use. If users and maintainers value the sympy support then then they will help maintain it. If no one else cares then the burden falls purely on sympy maintainers to maintain something that is possibly not used by anyone.

The thing is, pycosat *is* used, by conda. So it really ought to be maintained. So I'd like to first ascertain what the situation is with it. It would be a much happier outcome for everyone if we can get pycosat to be actively maintained again, either by getting Anaconda to actually maintain it or by somehow having someone else maintain it. The fact that there have been several PRs from various people, which have been ignored, means that there are many people who potentially might want to maintain it. It is used by at lot more than just SymPy.

I'm not really sure if you are are just saying these things in the abstract or are specifically talking about pycosat. I agree with your arguments in the abstract, but I really don't think they apply to pycosat. And even if they did, it is a bit premature to state that it is a dead project. Frankly, the sorts of arguments you are making apply more to something like SymEngine, because it has a somewhat small maintenance recently (although it isn't dead), its highly nontrivial to integrate with SymPy and continue to support (contrast with pycosat, which we can support completely with only a few lines of code), and it has a much smaller user base outside of SymPy, so unlike with pycosat, with SymEngine we probably would have to be the ones to step in and maintain it if no one else did. It also applies to mpmath, although mpmath is something that we basically can't live without, so we don't have a choice there.

oscarbenjamin:
> gmpy has several integer functions that we are (or should be) using, like isqrt and gcd.

I think these are both used but they are also both provided by the stdlib math module now so they are less important.

> I'm not really sure if you are are just saying these things in the abstract or are specifically talking about pycosat.

I have no particular objection to supporting pycosat. I am not saying these things in the abstract though. I think it is always important to consider whether we should continue to support something and I am asking specifically whether there is any value in sympy continuing to support the use of pycosat via sympy. That is a separate question from the value of pycosat itself as a library and also from the need for it to be maintained.

In the abstract if the policy is that once support is added for some library then it should be maintained per status quo forever until someone can prove that it is unused by any downstream users then that is clearly unsustainable. Actually I think that is already what happens now though. The optional dependency tests involve a bunch of libraries that I never hear any mention of from actual users. My position is that we need to be able to drop support for libraries that are no longer useful (in sympy) or now have better alternatives (for sympy's needs) and if we can't do that then I object to adding any new support for other libraries because any "support" is unsustainable under that policy.

There is of course a middle ground between "supporting" and "not supporting" a library. Instead what sympy can do is provide functions that make it easy to use external libraries by producing the inputs that they would expect. The documentation can include examples of how to use those libraries from sympy. It is not necessary for sympy to actually try to import those libraries and call their APIs directly though. Supporting other libraries in this way means being able to support current and future libraries without any actual changes in the sympy codebase. The failure to recognise this possibility is where sympy usually goes wrong: it tries to be the frontend for all the optional dependencies that it "supports".

migueltorrescosta:
> So the original author of pycosat no longer maintains it. However, Anaconda does still use pycosat for conda, so they should be maintaining it. I would suggest opening an issue in the repo asking for clarification if the project is maintained, and if it isn't maintained, requesting that they move the maintenance to someone who is willing to do so.

I wrote a PR for pycosat 2 weeks ago ( https://github.com/ContinuumIO/pycosat/pull/39 ) , before considering adding pySAT to sympy , but there was no reply. I tweeted to Anaconda asking about whether they support the repo https://twitter.com/migueltcosta/status/1396119906802978817 They said they would look into it, but I get a feeling they have more urgent concerns than maintaining this repo.

I am not as invested in pycosat maintenance since many SAT solvers have similar performance, but as these interactions showed me they are unlikely to actively maintain it, I started searching for more actively maintained efficient SAT solvers, and pySAT was the best I found.

> I find that I spend what seems like wasted time trying to fix things up for sympy to work with some library that I don't know or care about without having any idea whether anyone else actually cares about sympy's support for that library either.

I added a single solver (minisat 22), but pySat provides many other algorithms, allowing Sympy to provide multiple SAT algorithms while having a single optional library. For me it would be really useful, even though I can't vouch for the average sympy user.

> The optional dependency tests involve a bunch of libraries that I never hear any mention of from actual users.

Do you get any metrics on sympy usage? I can imagine it is impossible to get function call counts, but you might know something I don't 😅

> In the abstract if the policy is that once support is added for some library then it should be maintained per status quo forever until someone can prove that it is unused by any downstream users then that is clearly unsustainable.

I second this sentiment strongly, as I went through the same on a previous project and it was frustrating to support tools without knowing if anyone used it or even cared about it

migueltorrescosta:
Given the conversation above I would suggest merging the `minimal` option that uses pySAT, and I can open a ticket to discuss removing the dependency on picosat, where either we choose to keep it or make a PR removing it. It would be a specific intro for the discussion on how to handle optional dependencies, namely:
- What are the guidelines / criteria for adding an optional dependency?
- What are the guidelines / criteria for removing a dependency?
- How to protect against possibly stale dependencies

What do you think @oscarbenjamin @asmeurer ?

sympy/logic/algorithms/minisat22_wrapper.py
```diff
@@ -0,0 +1,46 @@
...
+
+    # Return UNSAT when False (encoded as 0) is present in the CNF
+    if {0} in expr.data:
+        if all_models:
+            return (f for f in [False])
+        return False
+
+    r = Minisat22(expr.data)
+
+    if minimal:
```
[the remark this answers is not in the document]
migueltorrescosta: You are absolutely right! I added them in my latest commit, and pushed the changes. I took this opportunity to merge master into the branch, to avoid conflicts. Somehow I'm not being able to squash that commit with my changes: my attempts to do so make it so that the master changes are reversed.

oscarbenjamin:
This looks good to me

asmeurer:
> It would be a specific intro for the discussion on how to handle optional dependencies, namely:

This is the sort of thing I'd like to use the SymPEP process to do https://github.com/sympy/sympeps

asmeurer:
pySAT seems like a nice solution to this issue. 

In general I don't think adding or dropping support for SAT solvers is a huge deal, because they don't really affect functionality. They only affect performance. So the worst that can happen is someone will install a library thinking it will help SymPy but it actually won't because support was dropped. That's a much different situation from other kinds of optional dependencies that SymPy has or can have.

> I wrote a PR for pycosat 2 weeks ago , before considering adding pySAT to sympy , but there was no reply. I tweeted Anaconda asking about whether they support the repo, and they said they would look into it, but I get a feeling they have more urgent concerns than pycosat.

As I said, if you want to see pycosat supported better, I would suggest asking Anaconda if they would be interested in either transfering maintainership or adding some external maintainers (especially if you yourself are interested in doing it, or if you can find someone else who is). I don't know if they would do it, but it doesn't hurt to ask.

oscarbenjamin:
Thanks

migueltorrescosta:
Looking at the [merge request](https://github.com/sympy/sympy/commit/110997fe18b9f7d5ba7d22f624d156a29bf40759) one of the [checks failed](https://travis-ci.org/github/sympy/sympy/jobs/774051188). This seems to be an issue with the optional dependency testing and not the code itself. The package is installed via `pip install python-sat` but used as `import pysat`, and my best guess is that it I added the wrong option to the `.travis.yml` file. Is there a way I can check this locally? It doesn't help that `pip install pysat` works but it installs a [Python Satellite Data Analysis Package](https://pypi.org/project/pysat/) , which is completely unrelated to the SAT solver package.

Removing the optional dependency would also fix the test, but it would be a no go solution since it would leave the package untested.

oscarbenjamin:
The master branch is also tested with Travis although it doesn't run before the PR is merged.

The dependency will need to be added in `.travis.yml`.

migueltorrescosta:
**Status Update**
- without installing `python-sat` all works by defaulting to the core packages / code.
- Running `pip install python-sat` all works well using the new code
- The travis testing environment fails when it runs `conda install python-sat` since conda does not recognise packages from https://pypi.org/

**Context**
I was exploring into the differences between `pip install <package_name>` and `import <imported_package>`. Turns out the issue with python-sat is that `conda` and `pip` use different repos for searching packages. `pip` uses the python core https://pypi.org/ , whereas I am still trying to understand how conda handles more recent state of the art packages. Sympy's `.travis.yml` uses conda to create the testing environment, but conda hasn't added support for `python-sat` yet. I tested this when trying to recreate Sympy's testing environment by pulling a clean miniconda docker image

```bash
docker pull conda/miniconda3
docker run -it conda/miniconda3 bash
```

- with CONDA I get

```bash
root@c53daac82749:/# conda install python-sat
WARNING: The conda.compat module is deprecated and will be removed in a future release.
Collecting package metadata: done
Solving environment: /
PackagesNotFoundError: The following packages are not available from current channels:

  - python-sat

Current channels:

  - https://repo.anaconda.com/pkgs/main/linux-64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/free/linux-64
  - https://repo.anaconda.com/pkgs/free/noarch
  - https://repo.anaconda.com/pkgs/r/linux-64
  - https://repo.anaconda.com/pkgs/r/noarch

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org

and use the search bar at the top of the page.
```

- with PIP I get

```bash
root@c53daac82749:/# pip install python-sat
Collecting python-sat
  Downloading https://files.pythonhosted.org/packages/0b/02/197fcb4c0bb7d3af22ec6017cf8ef4bdb05bbc565ba331730d549df4aaf1/python_sat-0.1.7.dev4-cp37-cp37m-manylinux1_x86_64.whl (1.6MB)
    100% |████████████████████████████████| 1.6MB 2.1MB/s
Requirement already satisfied: six in /usr/local/lib/python3.7/site-packages (from python-sat) (1.12.0)
Installing collected packages: python-sat
Successfully installed python-sat-0.1.7.dev4
```

oscarbenjamin:
We could just use `pip` in Travis. Otherwise maybe these tests should be skipped in Travis.

oscarbenjamin:
I opened #21607 which just removes python-sat from the Travis tests. It isn't necessary to test this in the old CI if we are happy that it is well tested in the new CI.

migueltorrescosta:
Is there a timeline for the next sympy version release? I am eager to use this functionality 😅

oscarbenjamin:
You can just use sympy from git for now.

Otherwise now that GSOC has started I think maybe the best time for the release is after GSOC. I could do a quick release now except I don't have time just yet.

migueltorrescosta:
I see, I will do that. Thank you!

Issue #21407, opened 2021-04-30
SetExpr/AccumBounds compute powers of the same


smichr:
There is no handling of `other` being an AccumBounds so the default of (-oo,oo) is returned for 
```python
>>> AccumBounds(1,2)**AccumBounds(1,3)
AccumBounds(-oo, oo)  <-- should be (1, 8)?
```
When trying this with SetExpr I am not sure how to get the max or min and `inf` and `sup` don't work:
```python
>>> SetExpr(Interval(1,2))**SetExpr(Interval(2,3))
SetExpr(ImageSet(Lambda((x, y), x**y), Interval(1, 2), Interval(2, 3)))
```

jksuom:
> should be (1, 8)?

I think so. This case seems to be missing from `AccumBounds.__pow__`.

smichr:
@jksuom , would the SetExpr be the appropriate place to allow `1**oo` to be 1?
```python
>>> from sympy.sets.setexpr import SetExpr as f
>>> f(Interval(0,1))**oo
SetExpr(Interval(0, oo))
```
In this context we know that no value is over 1 so can't we return `SetExpr(Interval(0, 1))`?

smichr:
@oscarbenjamin , since you worked with interval that behaves the same way, is there any context in which a closed interval, [0,1], raised to an arbitrarily large number, can give [0,1] again? And how can we say "arbitrarily large" other than `oo`?

jksuom:
> would the SetExpr be the appropriate place to allow 1**oo to be 1?

I'm not sure what `SetExpr(...)**oo` should represent. If it should be the union of all `x**oo` for all `x` in the SetExpr, then it looks like `1**oo` should be defined as in `numbers`. However, if it should be considered as the limit of `SetExpr(...)**n`, then the result would be different. For example, `[0, 1]**n ` would be `[0. 1]` for all positive integers `n`and so `[0, 1]**oo` could be `[0, 1]` as well. In the case of `AccumBound`, we definitely should consider the base as a limit.

smichr:
> In the case of `AccumBound`, we definitely should consider the base as a limit.

[...]

PR #21516, opened 2021-05-24
enhanced handling of AccumBounds in Pow

smichr:

addresses AccumBounds aspect of #21407

Previously,
```python
>>> 3**AccumBound(2, 3)
3**AccumBounds(2, 3)
>>> AccumBounds(2, 3)**AccumBounds(2, 3)
AccumBounds(-oo, oo)
```
Now
```python
assert 3**AccumBounds(2, 3) == AccumBounds(9, 27)'
assert AccumBounds(2, 3)**AccumBounds(2, 3) == AccumBounds(4, 27)
```
#### Other comments

The handling relies on min/max to work out the details after consider all possible candidates for bases and exponents.

- [x] will an unevaluated form have to be returned instead of raising an error with unsigned bounds?

#### Release Notes

* calculus
  * AccumBounds will be handled if possible as exponents


jksuom:
An instance of `AccumBounds` is expected to contain all accumulation values of a real function `f` at an extended real point `a` when a well defined limit does not exist. (Those are all existing limits of `f(x_n)` for sequences `x_n` tending to `a`. For example, the accumulation values of `sin` at `oo` are the limits `sin(c + 2*n*pi) = sin(c)` which fill the interval `[-1, 1]`.)

A function of the form `f(x) = b**e(x)` where `b` is negative, is real only if `e(x)` is an integer (and therefore constant if it is continuous). So if `e(x)` is not constant, `b` should be non-negative. In that case, it may also be an instance of `AccumBounds` with non-negative minimum. It looks like the easiest way to implement `b**e` in such a case would be `exp(log(b)*e)` as the product `log(b)*e` has already been defined for `AccumBounds`.

smichr:
@jksuom , is there something here that needs to be changed? This is only recognizing `AccumBounds**AccumBounds`.

smichr:
Do you mean something like this:
```diff
diff --git a/sympy/calculus/util.py b/sympy/calculus/util.py
index d438298531..1d2bb22b11 100644
--- a/sympy/calculus/util.py
+++ b/sympy/calculus/util.py
@@ -1,4 +1,4 @@
-from sympy import Order, S, log, limit, lcm_list, im, re, Dummy, Piecewise
+from sympy import Order, S, limit, lcm_list, im, re, Dummy, Piecewise
 from sympy.core import Add, Mul, Pow
 from sympy.core.basic import Basic
 from sympy.core.compatibility import iterable
@@ -7,6 +7,7 @@
 from sympy.core.numbers import _sympifyit, oo
 from sympy.core.relational import is_le, is_lt, is_ge, is_gt
 from sympy.core.sympify import _sympify
+from sympy.functions.elementary.exponential import exp, log
 from sympy.functions.elementary.miscellaneous import Min, Max
 from sympy.logic.boolalg import And
 from sympy.sets.sets import (Interval, Intersection, FiniteSet, Union,
@@ -1461,37 +1462,38 @@ def __pow__(self, other):
 
     @_sympifyit('other', NotImplemented)
     def __rpow__(self, other):
-        if other.is_Number:
+        p = set()
+        if other.is_real:
             if other is S.One:
                 return S.One
-            p = set()
-            if other.is_real:
-                if other < 0:
-                    i = int(self.min)
-                    if i not in self:
-                        i += 1
+            if other.is_negative:
+                i = int(self.min)
+                if i not in self:
+                    i += 1
+                if i in self:
+                    p.add(other**i)
+                    i += 1
                     if i in self:
                         p.add(other**i)
-                        i += 1
-                        if i in self:
-                            p.add(other**i)
-                    i = int(self.max)
+                i = int(self.max)
+                if i in self:
+                    p.add(other**i)
+                    i -= 1
                     if i in self:
                         p.add(other**i)
-                        i -= 1
-                        if i in self:
-                            p.add(other**i)
-                elif other > 0:
-                    p.add(other**self.min)
-                    p.add(other**self.max)
-                else:
-                    try:
-                        if 0 in self:
-                            p.add(1)
-                    except TypeError:
-                        raise ValueError('sign of bounds unknown')
-                    if self.max > 0:
-                        p.add(0)
+            elif other.is_positive:
+                p.add(other**self.min)
+                p.add(other**self.max)
+            elif other.is_zero:
+                try:
+                    if 0 in self:
+                        p.add(1)
+                except TypeError:
+                    raise ValueError('sign of bounds unknown')
+                if self.max > 0:
+                    p.add(0)
+            else:
+                raise NotImplementedError('sign of base is unknown')
             if p:
                 return self.func(min(p), max(p))
             elif other == 0:
@@ -1499,7 +1501,7 @@ def __rpow__(self, other):
             else:
                 raise ValueError('Only real AccumulationBounds are supported')
 
-        return NotImplemented
+        return exp(self*log(other))
 
     def __abs__(self):
         if self.max.is_extended_negative:
```

Allow more than Number to be evaluated
```python
>>> cos(2)**AccumBounds(-3,4)
AccumBounds(cos(2)**(-3), cos(2)**(-2))
```
Allow generic expression to be returned
```python
>>> exp(x)**AccumBounds(-3,4)
exp(AccumBounds(-3, 4)*log(exp(x)))
>>> _.subs(x,I*pi)
exp(AccumBounds(-3, 4)*I*pi)

>>> exp(I*pi)
-1
>>> _**AccumBounds(-3,4)
AccumBounds(-1, 1)
```
But notice how the generic result does not allow one to do much.

sympy/calculus/tests/test_util.py
```diff
@@ -492,6 +494,177 @@ def test_AccumBounds_pow():
         Pow(AccumBounds(-oo, oo), AccumBounds(0, 1))
 
 
+def test_AccumBounds_as_exponent():
+    B = AccumBounds
+    # base = -1
+    assert (-1)**B(-2, -1) == B(-1, 1)  # even and odd
```
jksuom, line 500: `(-1)**x` is not real for `x` in the interval `[-2. -1]` (or any other interval of length > 0). Therefore the result is not of type `AccumBounds` which is expected to point to a subset of extended reals. It would be possible to define "complex accumulation bounds" but those have not been implemented. So I think that all `b**e` where `e` is an instance of `AccumBounds` and `b` or `b.min` is negative should be left unevaluated for now.

sympy/calculus/util.py
```diff
@@ -1358,10 +1358,149 @@ def __pow__(self, other):
                 if den!=1:
                     num_pow = self**num
                     return num_pow**(1 / den)
+
+            if isinstance(other, AccumBounds):
```
jksuom, line 1362: In this case, `self.min` should be non-negative.

sympy/calculus/tests/test_util.py
smichr: I am abusing the bounds notation, right? What I really mean is `{-1, 1}` are the only real answers. Perhaps all this logic is better applied to SetExpr. (Or, as you indicate, just let negative bases be unevaluated or raise an error that 'negative bases have finite solutions only for integer exponents'.)

smichr:
@jksuom, I see you were involved with the original discussion of this. The references refer to Interval Arithmetic and this behaves like interval arithmetic but it also includes (as a subset) the notion of bounds at infinity. So in a sense, `cos(oo)` is using an interval to represent something about itself. But that doesn't mean that it can't still function as a non-Set interval.

Whereas SetExpr can handle more than Interval, AccumBound is like a version of SetExpr that contains only Interval. As such -- and since it only contains real values and since it doesn't make claims about what values are between the bounds -- I wonder if it might be ok to accept the changes that I have made. Consider

1. there is precedent in that sqrt(B(-1,4)) -> B(0, 2): imaginary possibilities are ignored
2. it is documented that the supremum, infimum are not necessarily represented in the interval; this claim (lifted from wikipedia) is a bit hard for me to understand unless it is applied to a case where a function has a discontinuity as `sec(B(0,2)) -> B(-oo,oo)` though it has a minimum positive value of 1 and maximum negative value ~= -2.4. So although the bounds are infinite it does not mean that all values of the function are between those values

In light of those two observations, it might not be wrong to say that `(-1)**B(-2,-1) == B(-1, 1)` even though there are *only* 2 real points possible for the values in the range of the exponent. This is in the same spirit as the first point and it is an extreme case of the second -- here all but two points are excluded from the given interval whereas a range between -2.4 and 1 are actually excluded from the infinite range of the second example.

smichr:
> 2\. a bit hard for me to understand unless it is applied to a case where a function has a discontinuity

OK, I see. Here is an example of a continuous polynomial:

image

```python
>>> (x**3/3 - 3*x**2/2 + x).subs(x,AccumBounds(0,4))
AccumBounds(-24, 76/3)
>>> factor(x**3/3 - 3*x**2/2 + x).subs(x,AccumBounds(0,4))
AccumBounds(-20, 76/3)
>>> from sympy.calculus.singularities import *
>>> function_range(x**3/3 - 3*x**2/2 + x,x,Interval(0,4)).n(3)
Interval(-1.68, 1.33)
```

jksuom:
> there is precedent in that sqrt(B(-1,4)) -> B(0, 2)

Where is this? I think that it is a typo: there should be `sqrt(B(0, 4)`.

smichr:
> Where is this?

That was from here:

https://github.com/sympy/sympy/blob/9dd303ff4fbcf3fc8b73f5013879e253436c9f59/sympy/calculus/util.py#L946-L947

jksuom:
That looks like a bug in the code. Base should be non-negative if the exponent is not an integer.

jksuom:
https://github.com/sympy/sympy/blob/c0993ea7b111efb35d48697148230b7a84cd5fc9/sympy/calculus/util.py#L1354-L1355
This is not correct. It is probably best to return unevaluted `Pow`. There may be other similar cases in the code.

sympy/calculus/util.py
```diff
@@ -1348,20 +1349,66 @@ def __pow__(self, other):
...
-                    num_pow = self**num
-                    return num_pow**(1 / den)
+                        try:
+                            if S.Zero in self:
+                                return AccumBounds(
+                                    real_root(self.min, den),
+                                    real_root(self.max, den))
+                        except TypeError:
+                            AccumBounds(-oo, oo)
+                elif den is not S.One  # e.g. if other is not Float
```
smichr, line 1361:
```suggestion
                elif den is not S.One:  # e.g. if other is not Float
```

sympy/calculus/util.py
```diff
@@ -1348,20 +1349,66 @@ def __pow__(self, other):
                         return AccumBounds(self.min**other, self.max**other)
 
                 num, den = other.as_numer_denom()
```
jksuom, line 1351: At this point, `other` is a number but not an integer and `self.min` is not non-negative. I think that an unevaluated `Pow` should be returned.

sympy/calculus/util.py
```diff
@@ -1348,20 +1349,66 @@ def __pow__(self, other):
...
+            if other.is_positive:
+                p.add(other**self.min)
+                p.add(other**self.max)
+            elif other.is_zero:
+                if (self.min*self.max).is_nonpositive:
+                    p.add(1)
+                    if self.max.is_positive:
+                        p.add(0)
+                elif self.max.is_negative:
+                    raise ZeroDivisionError
```
smichr, line 1406:
```suggestion
            if other.is_extended_positive:
                p.add(other**self.min)
                p.add(other**self.max)
            elif other.is_zero:
                if self.min.is_zero:
                    p.add(0)
                    if self.max.is_extended_positive:
                        p.add(1)
                elif self.min.is_extended_positive:
                    p.add(1)
```

sympy/calculus/util.py
smichr: I might be missing something but at the outset we test `other.is_Integer or self.min.is_extended_nonnegative` so the min *is* non-negative.

sympy/calculus/util.py
```diff
@@ -1348,20 +1349,66 @@ def __pow__(self, other):
...
+            if other.is_extended_positive:
+                p.add(other**self.min)
+                p.add(other**self.max)
+            elif other.is_zero:
+                if self.min.is_zero:
+                    p.add(0)
+                    if self.max.is_extended_positive:
+                        p.add(1)
+                elif self.min.is_extended_positive:
+                    p.add(1)
```
smichr, line 1406: Should I raise `ZeroDivisionError` if `self.min.is_extended_nonpositive`?

jksuom: `0**negative` returns `zoo`. Maybe we could avoid raising error and return `zoo` if also `self.max.is_extended_nonpositive`. Otherwise I think that `nan` could be returned

smichr, line 1404:
```suggestion
                    p.add(1)
                elif self.max.is_extended_nonpositive:
                    return S.ComplexInfinity
```

smichr: I will try just an unevaluated expression since the user could then substitute in a value later.

sympy/calculus/tests/test_util.py
```diff
@@ -492,6 +494,78 @@ def test_AccumBounds_pow():
         Pow(AccumBounds(-oo, oo), AccumBounds(0, 1))
 
 
+def test_AccumBounds_as_exponent():
+    B = AccumBounds
+    # base is 0
+    raises(ZeroDivisionError, lambda: 0**B(-1, -1))  # all neg
```
smichr, line 500:
```suggestion
    assert 0**B(-1, 0) is S.ComplexInfinity
    z = 0**B(a, a + S.Half)
    assert z.subs(a, 0) == B(0, 1)
    assert z.subs(a, 1) == 0
```

sympy/calculus/tests/test_util.py
```diff
@@ -492,6 +494,78 @@ def test_AccumBounds_pow():
...
+    assert 0**B(0, S.Half) == B(0, 1)  # 0 and pos
+    raises(ValueError, lambda: 0**AccumBounds(a, a + 1))
```
smichr, line 502:
```suggestion
```

sympy/calculus/util.py
```diff
@@ -894,22 +894,26 @@ class AccumulationBounds(AtomicExpr):
...
+    When an AccumBounds is raised to a negative power, if 0 is contained
+    between the bounds then an infinite range is returned, otherwise if an
+    endpoint is 0 then a semi-infinite range with consistent sign will be returned.
 
-    `X / Y = \{ z \mid \exists x \in X, y \in Y \mid y \neq 0, z = x/y\}`
-
-    Note: According to this definition the quotient of two AccumulationBounds
-    may not be a AccumulationBounds object but rather a union of
-    AccumulationBounds.
+    >>> 1/AccumBound(-1, 1)
```
smichr, line 901:
```suggestion
    >>> 1/AccumBounds(-1, 1)
```

sympy/calculus/util.py
```diff
@@ -936,19 +940,13 @@ class AccumulationBounds(AtomicExpr):
 
     `X^n = \{ x^n \mid x \in X\}`
 
+    >>> AccumBounds(1, 2)**2
+    AccumBounds(1, 4)
```
smichr, line 944:
```suggestion
    >>> AccumBounds(1, 4)**(S(1)/2)
    AccumBounds(1, 2)
```

sympy/calculus/util.py
```diff
@@ -936,21 +935,25 @@ class AccumulationBounds(AtomicExpr):
...
-    AccumulationBounds object is neglected.
-
-    >>> AccumBounds(-1, 4)**(S(1)/2)
-    AccumBounds(0, 2)
+    >>> 1/AccumBounds(-1, 1)
+    AccumBounds(-oo, oo)
+    >>> 1/AccumBounds(0, 2)
+    AccumBounds(1/2, oo)
+    >>> 1/AccumBounds(-3, 0)
+    AccumBounds(-oo, -1/3)
```
smichr, line 948:
```suggestion
    >>> 1/AccumBounds(-oo, 0)
    AccumBounds(-oo, 0)
```

sympy/calculus/util.py
```diff
@@ -1348,20 +1351,60 @@ def __pow__(self, other):
...
+                        other.min.is_negative is not None):
+                    # negative numbers require only integer exponents to produce
+                    # real bounds and AccumBounds includes non-integers so we
+                    # must deal only with nonnegative bases in self.
+                    p = set()
+                    if self.min.is_zero:
+                        if other.min.is_extended_negative:
+                            raise ZeroDivisionError
+                        if other.max.is_extended_positive:
+                            p.add(0)
```
smichr, line 1372:
```suggestion
                    if self.min.is_zero:
                        if other.min.is_extended_negative:
                            return S.ComplexInfinity
                        if other.min.is_extended_positive:
                            p.add(0)
```

sympy/calculus/util.py
```diff
@@ -1348,20 +1351,60 @@ def __pow__(self, other):
...
+                elif den is not S.One:  # e.g. if other is not Float
+                    return (self**num)**(1/den)  # ok for non-negative base
+
+            if isinstance(other, AccumBounds):
+                if (
+                        self.min.is_zero is not None and
+                        other.min.is_negative is not None):
+                    # negative numbers require only integer exponents to produce
+                    # real bounds and AccumBounds includes non-integers so we
+                    # must deal only with nonnegative bases in self.
```
smichr, line 1366:
```suggestion
                        other.min.is_extended_negative is not None):
                    # negative numbers require only integer exponents to produce
                    # real bounds and AccumBounds includes non-integers so we
                    # must deal only with nonnegative bases in self.
                    if other.min.is_extended_negative:
                        if self.max.is_extended_negative:
                            return S.NaN  # finite complex values
                        elif 0 in self:
                            return S.ComplexInfinity
```

sympy/calculus/util.py
smichr, line 1372:
```suggestion
                    if self.min.is_zero and other.max.is_extended_positive:
                        p.add(0)
```

sympy/calculus/util.py
```diff
@@ -936,21 +935,25 @@ class AccumulationBounds(AtomicExpr):
...
+    AccumBounds(-oo, oo)
+    >>> 1/AccumBounds(0, 2)
+    AccumBounds(1/2, oo)
+    >>> 1/AccumBounds(-oo, 0)
+    AccumBounds(-oo, 0)
 
-    >>> AccumBounds(1, 2)**2
-    AccumBounds(1, 4)
+    If the exponent is itself an AccumBounds instance, the base must be
+    non-negative or else a complex infinity will be returned:
```
smichr, line 951:
```suggestion
    If the exponent is itself an AccumBounds instance and it has
    a negative lower bound then the base must be positive or else
    special values will be returned:
```

sympy/calculus/util.py
```diff
@@ -936,21 +935,25 @@ class AccumulationBounds(AtomicExpr):
...
 
-    >>> AccumBounds(-1, oo)**(-1)
-    AccumBounds(-oo, oo)
+    >>> AccumBounds(0, 2)**AccumBounds(2, 3)
+    Accumbounds(1, 8)
+    >>> AccumBounds(-1, 2)**AccumBounds(2, 3)
+    zoo
```
smichr, line 956:
```suggestion
    >>> AccumBounds(2, 3)**AccumBounds(-1, 2)
    AccumBounds(1/3, 9)
    >>> AccumBounds(0, 2)**AccumBounds(-1, 2)
    zoo
    >>> AccumBounds(-1, 2)**AccumBounds(-1, 2)
    nan
```

sympy/calculus/util.py
```diff
@@ -936,21 +935,27 @@ class AccumulationBounds(AtomicExpr):
...
+    non-negative or else a complex infinity will be returned:
 
-    >>> AccumBounds(-1, oo)**(-1)
-    AccumBounds(-oo, oo)
+    >>> AccumBounds(2, 3)**AccumBounds(-1, 2)
+    AccumBounds(1/3, 9)
+    >>> AccumBounds(0, 2)**AccumBounds(-1, 2)
+    zoo
+    >>> AccumBounds(-1, 2)**AccumBounds(-1, 2)
+    nan
```
smichr, line 958:
```suggestion
    >>> AccumBounds(-2, 0)**AccumBounds(-1, 2)
    zoo
    >>> AccumBounds(-2, -1)**AccumBounds(-1, 2)
    nan
```

smichr:
I don't understand why the first error is occuring:
```python
  File "/home/runner/work/sympy/sympy/sympy/calculus/tests/test_util.py", line 507, in test_AccumBounds_as_exponent
    z = 0**B(a, a + S.Half)
TypeError: unsupported operand type(s) for ** or pow(): 'int' and 'AccumulationBounds'
```
Doesn't the `__rpow__` automatically sympify the 0?

The others, I will check on:
```________________________________________________________________________________
____________ sympy/calculus/tests/test_util.py:test_AccumBounds_pow ____________
  File "/home/runner/work/sympy/sympy/sympy/calculus/tests/test_util.py", line 452, in test_AccumBounds_pow
    assert AccumBounds(-1, 2)**Rational(1, 3) == AccumBounds(-1, 2**Rational(1, 3))
AssertionError
_______________ sympy/series/tests/test_limits.py:test_heuristic _______________
  File "/home/runner/work/sympy/sympy/sympy/series/tests/test_limits.py", line 200, in test_heuristic
    assert limit(log(2 + sqrt(atan(x))*sqrt(sin(1/x))), x, 0) == log(2)
AssertionError
```

sympy/calculus/util.py
```diff
@@ -1348,20 +1353,64 @@ def __pow__(self, other):
...
-                    return AccumBounds(real_root(self.min, den),
-                                       real_root(self.max, den))
-                if den!=1:
-                    num_pow = self**num
-                    return num_pow**(1 / den)
+                if num is S.One:
+                    return AccumBounds(*[i**(1/den) for i in self.args])
+
+                elif den is not S.One:  # e.g. if other is not Float
+                    return (self**num)**(1/den)  # ok for non-negative base
```
smichr, line 1360:
```suggestion
                    return (self**num)**(1/den)  # ok for non-negative base
            elif other.is_number:
                if other.min.is_extended_negative:
                    if (self.min*self.max).is_extended_nonpositive:
                        return S.ComplexInfinity
                    elif self.max.is_extended_negative:
                        return S.NaN
                raise NotImplementedError  # leave unevaluated
```

sympy/calculus/util.py
```diff
@@ -1348,20 +1353,64 @@ def __pow__(self, other):
...
+                        p.add(1)  # x**0 = 1
+                    if 1 in self:
+                        p.add(1)  # 1**x = 1
+                    b = set(self.args) - {0}  # handled above
+                    e = set(other.args)
+                    # compute candidate powers
+                    p |= {i**j for i in b for j in e}
+                    return self.func(min(p), max(p))
+
             return AccumBounds(-oo, oo)
```
smichr, line 1387:
```suggestion
            raise NotImplemented  # to leave unevaluated instead of return AccumBounds(-oo, oo)
```

sympy/calculus/util.py
```diff
@@ -1348,17 +1353,61 @@ def __pow__(self, other):
...
+                        p.add(1)  # x**0 = 1
+                    if 1 in self:
+                        p.add(1)  # 1**x = 1
+                    b = set(self.args) - {0}  # handled above
+                    e = set(other.args)
+                    # compute candidate powers
+                    p |= {i**j for i in b for j in e}
+                    return self.func(min(p), max(p))
+
+            raise NotImplemented  # to leave unevaluated instead of return AccumBounds(-oo, oo)
```
smichr, line 1387:
```suggestion
            raise NotImplementedError  # to leave unevaluated instead of return AccumBounds(-oo, oo)
```

sympy/calculus/tests/test_util.py
```diff
@@ -446,9 +448,8 @@ def test_AccumBounds_pow():
 
     assert AccumBounds(-4, 2)**Rational(2, 3) == AccumBounds(0, 2*2**Rational(1, 3))
 
-    assert AccumBounds(-1, 5)**S.Half == AccumBounds(0, sqrt(5))
-    assert AccumBounds(-oo, 2)**S.Half == AccumBounds(0, sqrt(2))
-    assert AccumBounds(-2, 3)**Rational(-1, 4) == AccumBounds(0, oo)
+    assert AccumBounds(-1, 0)**S.Half is S.NaN
```
jksuom, line 451: This could also be left unevaluated. That would preserve some size information for possible future use.

sympy/calculus/tests/test_util.py
```diff
@@ -446,9 +448,8 @@ def test_AccumBounds_pow():
...
+    assert AccumBounds(-2, 3)**Rational(-1, 4) is S.ComplexInfinity
```
jksuom, line 452: This could be `S.NaN`. The values may be finite but unbounded.

sympy/calculus/util.py
```diff
@@ -1348,17 +1353,61 @@ def __pow__(self, other):
...
+                    if self.max.is_extended_positive:
+                        p.add(1)
+                elif self.min.is_extended_positive:
+                    p.add(1)
+                elif self.max.is_extended_nonpositive:
+                    return S.ComplexInfinity
+        if p:
+            return self.func(min(p), max(p))
 
         return NotImplemented
```
jksuom, line 1412: The TypeError for `z = 0**B(a, a + S.Half)` comes from this line: nothing was added to `p`.

sympy/calculus/tests/test_util.py
smichr: I was thinking that they may be trumped by the `1/0`.

sympy/calculus/tests/test_util.py
```diff
@@ -441,14 +443,15 @@ def test_AccumBounds_pow():
...
-    assert AccumBounds(-4, 2)**Rational(2, 3) == AccumBounds(0, 2*2**Rational(1, 3))
-
-    assert AccumBounds(-1, 5)**S.Half == AccumBounds(0, sqrt(5))
-    assert AccumBounds(-oo, 2)**S.Half == AccumBounds(0, sqrt(2))
-    assert AccumBounds(-2, 3)**Rational(-1, 4) == AccumBounds(0, oo)
+    ni = Symbol('ni', integer=False)
+    neg = Symbol('neg', negative=True)
+    p = AccumBounds(neg, 1)**ni
+    for aneg in (-oo, -1):
+        for ani in (-S.Half, S.Half, S(1)/3, S(2, 3)):
```
smichr, line 452:
```suggestion
        for ani in (-S.Half, S.Half, S(1)/3, S(2)/3):
```

sympy/calculus/util.py
```diff
@@ -1347,21 +1353,58 @@ def __pow__(self, other):
...
+                p.add(other**self.max)
+            elif other.is_zero:
+                if self.min.is_zero:
+                    p.add(1)  # 0**0
+                elif self.min.is_extended_positive:
+                    p.add(0)
+        if p:
+            return self.func(min(p), max(p))
+
+        return Pow(other, self, evaluate=False)
```
smichr, line 1406:
```suggestion
        if other.is_real and other.is_extended_nonnegative and (
                self.max - self.min).is_extended_positive:
            if other is S.One:
                return S.One
            if other.is_extended_positive:
                a, b = [other**i for i in self.args]
                if min(a, b) != a:
                    a, b = b, a
                return self.func(a, b)
            if other.is_zero:
                if self.min.is_zero:
                    return self.func(0, 1)
                if self.min.is_extended_positive:
                    return S.Zero

        return Pow(other, self, evaluate=False)
```

sympy/calculus/util.py
```diff
@@ -1347,21 +1353,58 @@ def __pow__(self, other):
...
+                    if self.min.is_zero and other.max.is_extended_positive:
+                        p.add(0)
+                    if 0 in other:
+                        p.add(1)  # x**0 = 1
+                    if 1 in self:
+                        p.add(1)  # 1**x = 1
+                    b = set(self.args) - {0}  # handled above
+                    e = set(other.args)
+                    # compute candidate powers
+                    p |= {i**j for i in b for j in e}
```
smichr, line 1382:
```suggestion
                    p |= set(
                        (self**other.min).args) | set(
                        (self**other.max).args)
```

jksuom:
Noninteger powers of AccumBounds with negative minimum should be left unevaluated as those powers of negative numbers are not real. Most of these tests should be changed:
https://github.com/sympy/sympy/blob/5d4de364e968fbadc614f3dd374821d797a5fd5a/sympy/calculus/tests/test_util.py#L450-L457
and there are many more.

smichr:
> Noninteger powers of AccumBounds with negative minimum should be left unevaluated

I think it is sifted out now. Also, I left a note in the code but also ask here: should `1**oo` be considered 1 or oo? Perhaps failing tests will answer the question.

sympy/calculus/util.py
```diff
@@ -1290,14 +1297,18 @@ def __rtruediv__(self, other):
...
     def __pow__(self, other):
-        from sympy.functions.elementary.miscellaneous import real_root
         if isinstance(other, Expr):
             if other is S.Infinity:
                 if self.min.is_extended_nonnegative:
                     if self.max < 1:
                         return S.Zero
                     if self.min > 1:
                         return S.Infinity
+                    # XXX is 1**oo to be considered oo or 1?
```
jksuom, line 1307: The result should be `AccumBounds(0, 00)` also in these cases as there are (real) functions `f` and `g` that have limits `1`resp. `oo` but `f(t)**g(t)` may have any nonnegative real as limit or, more generally, as accumulation value, if a limit does not exist . If `h` is any function whose accumulation values (at the same point) are described by `AccumBounds(0, 2)` (like `sint(t) + 1` at `oo`) then `f(t) + h(t)` has limit `AccumBounds(1, 3)` and `(f(t) + h(t))**g(t)` may have any nonnegative extended real as accumulation value (`f(t) + h(t)` may be < 1 repeatedly).

jksuom:
`1**oo` is `nan` ("any extended complex value") in general as explained in the docstring of `Pow`. When considering limits of real functions (as is implicit in case of `Accumbounds(a, b)**oo`), the result could be `AccumBounds(0, oo)` ("any nonnegative extended real").

jksuom:
If a function has limit (or "limit") `AccumBounds(a, b)`, it does not mean that its values are between `a` and `b`. The values may repeatedly go outside of these bounds though eventually less and less. (One can think of the bounds as being `a-` and `b+`.) That small "overdraw" may become significant in an operation where the other operand is infinite.

smichr:
> If a function has limit (or "limit") `AccumBounds(a, b)`, it does not mean that its values are between `a` and `b`. The values may repeatedly go outside of these bounds though eventually less and less. (One can think of the bounds as being `a-` and `b+`.) That small "overdraw" may become significant in an operation where the other operand is infinite.

This might explain the result of 0 that wolfram gives for `((sin(x) + 1)/2)**x` as `x -> oo`; intersetingly, however, they give `frac(x)**x -> (0, 1)`, not `(0, oo)`.

Regarding `1**oo` -- we have internal inconsistency regarding this:
```
>>> 1**x
1
>>> 1**Dummy(infinite=1)
1
>>> 1**-oo
nan
>>> 1**oo
nan
```
but maybe that only exists if one (incorrectly) thinks of AccumBounds as an Interval.

jksuom:
In `frac(x)**x -> (0, 1)`, we know that `frac(x)` is strictly less than 1; there is no "overdraw".

smichr:
> there is no "overdraw".

Though my recollection is that `0.[9] == 1` ;-) But correcting that test in series is beyond the scope of this PR and was already marked as a 'todo'. It seems that we would have to special case `limit(frac(..)**f(x), x, oo)` to give (0, 1) if `limit(f(x), x, oo) -> oo`

smichr:
This is ready for review.

sympy/calculus/tests/test_util.py
```diff
@@ -310,259 +311,325 @@ def test_issue_19869():
...
+
+    assert B(Rational(1, 3), S.Half)**oo is S.Zero
+    assert B(0, S.Half)**oo is S.Zero
+    assert B(S.Half, 1)**oo == B(0, oo)
+    assert B(0, 1)**oo == B(0, oo)
+    assert B(2, 3)**oo is oo
+    assert B(1, 2)**oo is oo
+    assert B(S.Half, 3)**oo == B(0, oo)
+    assert B(Rational(-1, 3), Rational(-1, 4))**oo is S.Zero
+    assert B(-1, Rational(-1, 2))**oo == B(-oo, oo)
```
jksuom, line 497: We cannot assume that the variable `x` tending to `oo` is always an integer (though it may be as in case of `limitseq`). Hence the values of `(-1)**x` as `x -> oo` cover the whole unit circle `|z| = 1`, and so all of its points are accumulation values. As `-1` may be an accumulation value of a function taking values less than `-1`, any extended complex number of arbitrary size could be an accumulation value. This means that the result should be `nan` (not `B(-oo, oo)` that can be regarded as "real nan").

sympy/calculus/tests/test_util.py
```diff
@@ -310,259 +311,325 @@ def test_issue_19869():
...
+    assert B(-3, -2)**oo == FiniteSet(-oo, oo)
```
jksuom, line 498: This should cover also complex infinities, so `zoo`.

sympy/calculus/tests/test_util.py
```diff
@@ -310,259 +311,325 @@ def test_issue_19869():
...
+    assert B(-2, -1)**oo == B(-oo, oo)
```
jksuom, line 499: `nan` again.

sympy/calculus/util.py
```diff
@@ -1290,21 +1308,20 @@ def __rtruediv__(self, other):
...
     @_sympifyit('other', NotImplemented)
     def __pow__(self, other):
-        from sympy.functions.elementary.miscellaneous import real_root
         if isinstance(other, Expr):
             if other is S.Infinity:
                 if self.min.is_extended_nonnegative:
                     if self.max < 1:
                         return S.Zero
-                    if self.min > 1:
+                    if self.min >= 1:
```
jksuom, line 1316: `min` should be strictly greater than 1. If 1 is an accumulation value of a function `f(t)` at `a`, then there may be a sequence `t_n` tending to `a` such that `f(t_n) = 1` for all `n`. Then also `f(t_n)**n = 1`, so 1 is also an accumulation value for exponent `oo`.

smichr: Should B(1,2)**oo be B(0, oo) or B(1, oo)?

jksuom: `B(0, oo` as the function `f` having accumulation points in `[1, 2]` may take values less than 1. In the example above I had chosen `f(t_n) = 1` for simplicity, but we could also have `f(t_n) = 1 - 1/n`. If `g(t_n)` tends to `oo` so that `g(t_n)/n` also tends to `oo` (e.g,, `g(t_n) >= n**2`), then `f(t_n)**g(t_n) = ((1 - 1/n)**n)**(g(t_n)/n) -> (1/E)**oo = 0`.

smichr: Maybe starting to get the idea. Is `sin(oo)` = [-1,1] because the curvature at pi/2 is concave up or down so whether you approach from the right or left, the limit is 1?

sympy/calculus/tests/test_util.py
```diff
@@ -310,259 +311,325 @@ def test_issue_19869():
...
+    assert B(0, oo)**S.Half == B(0, oo)
+    assert B(-oo, 0)**(-2) == B(0, oo)
+    assert B(-2, 0)**(-2) == B(Rational(1, 4), oo)
+
+    assert B(Rational(1, 3), S.Half)**oo is S.Zero
+    assert B(0, S.Half)**oo is S.Zero
+    assert B(S.Half, 1)**oo == B(0, oo)
+    assert B(0, 1)**oo == B(0, oo)
+    assert B(2, 3)**oo is oo
+    assert B(1, 2)**oo is B(0, oo)
```
smichr, line 494:
```suggestion
    assert B(1, 2)**oo == B(0, oo)
```

sympy/calculus/tests/test_util.py
```diff
@@ -310,259 +311,325 @@ def test_issue_19869():
...
+    assert B(2, 3)**B(-3, 2) == B(S(1)/27, 9)
+
+    # bounds generating imaginary parts unevaluated
+    unchanged(Pow, B(-1, 1), B(1, 2))
+    assert B(0, S(1)/2)**B(1, oo) == B(0, S(1)/2)
+    assert B(0, 1)**B(1, oo) == B(0, oo)
+    assert B(0, 2)**B(1, oo) == B(0, oo)
+    assert B(0, oo)**B(1, oo) == B(0, oo)
+    assert B(S(1)/2, 1)**B(1, oo) == B(0, oo)
+    assert B(S(1)/2, 1)**B(-oo, -1) == B(1, oo)
```
smichr, line 549:
```suggestion
    assert B(S(1)/2, 1)**B(-oo, -1) == B(0, oo)
```

sympy/calculus/tests/test_util.py
```diff
@@ -310,259 +311,325 @@ def test_issue_19869():
...
+    assert B(S(1)/2, 1)**B(-oo, oo) == B(0, oo)
+    assert B(S(1)/2, 2)**B(1, oo) == B(0, oo)
+    assert B(S(1)/2, 2)**B(-oo, -1) == B(0, oo)
+    assert B(S(1)/2, 2)**B(-oo, oo) == B(0, oo)
+    assert B(S(1)/2, oo)**B(1, oo) == B(0, oo)
+    assert B(S(1)/2, oo)**B(-oo, -1) == B(0, oo)
+    assert B(S(1)/2, oo)**B(-oo, oo) == B(0, oo)
+    assert B(1, 2)**B(1, oo) == B(1, oo)
```
smichr, line 557:
```suggestion
    assert B(1, 2)**B(1, oo) == B(0, oo)
```

sympy/calculus/tests/test_util.py
```diff
@@ -310,259 +311,325 @@ def test_issue_19869():
...
+    assert B(1, 2)**B(-oo, -1) == B(0, oo)
+    assert B(1, 2)**B(-oo, oo) == B(0, oo)
+    assert B(1, oo)**B(1, oo) == B(1, oo)
```
smichr, line 560:
```suggestion
    assert B(1, oo)**B(1, oo) == B(0, oo)
```

sympy/calculus/util.py
jksuom: `sin(x)` might be a slightly misleading example as its values (on the real axis) are in the interval `[-1, 1]`. A better example would be `f(x) = (1 + 1/x)*sin(x)` that takes repeatedly values not in the interval `[-1, 1]` though its accumulation values at `oo` are there. 

For each `b` in `[-1, 1]`, there is a real `c` such that  `sin(c) = b`. Then `x_n = c + 2*pi*n` is a sequence tending to `oo` such that `f(x_n) -> b`, so `b` is an accumulation value of `f` at `oo`.

smichr:
@jksuom , please merge if you think this is ready. I have nothing more to add.

smichr:
I will clean up commits (deleting the doit commit which had to do with integrals) and then repush. Is there anything else to be done, @jksuom ?

jksuom:
Sorry. I was away from keyboard and didn't see your last comments. In fact, I think this is ready go.

smichr:
@jksuom I will commit after cleaning up and letting tests pass.

sympy/integrals/integrals.py
```diff
@@ -484,10 +484,22 @@ def doit(self, **hints):
...
+                # the default for doit is deep=False
+                # but integrals may have evaluated and
+                # contain objects that will simplify
+                # if their doit method is triggered
+                # so we mask off Integrals and apply doit
+                did = did.replace(
+                    lambda x: isinstance(x, Integral),
+                    lambda x: x.xreplace({x: igl.setdefault(x, Dummy())})
+                    ).doit()
+                igl = {v: k for k, v in igl.items()}
```
smichr, line 498:
```suggestion
```

sympy/integrals/integrals.py
```diff
@@ -484,10 +484,22 @@ def doit(self, **hints):
...
             if type(did) is tuple:  # when separate=True
-                did = tuple([i.xreplace(undo) for i in did])
+                did = tuple([i.xreplace(igl).xreplace(undo) for i in did])
```
smichr, line 500:
```suggestion
                did = tuple([i.xreplace(undo) for i in did])
```

sympy/integrals/integrals.py
```diff
@@ -484,10 +484,22 @@ def doit(self, **hints):
...
             else:
-                did = did.xreplace(undo)
+                did = did.xreplace(igl).xreplace(undo)
```
smichr, line 502:
```suggestion
                did = did.xreplace(undo)
```

Issue #21517, opened 2021-05-24
`parse_latex` introduces unnecessary factor of -1


smichr:
```python
>>> from symyp.parsing.latex import parse_latex
>>> srepr(parse_latex('1 - 2*(x + 1))'))
"Add(Integer(1), Mul(Integer(-1), Integer(2), Add(Symbol('x'), Integer(1))))"
>>> srepr(parse_latex('-3.14'))
"Mul(Integer(-1), Float('3.1400000000000001', precision=53))"
```

The solution to the first expression would involve using `_keep_coeff(S.NegativeOne, mul)` in a `convert_mul` routine that is similar to the `convert_add` routine.

The solution to the second issue involves 
```diff
diff --git a/sympy/parsing/latex/_parse_latex_antlr.py b/sympy/parsing/latex/_parse_latex_antlr.py
index ca50b4958a..2e4a45ad20 100644
--- a/sympy/parsing/latex/_parse_latex_antlr.py
+++ b/sympy/parsing/latex/_parse_latex_antlr.py
@@ -164,11 +164,8 @@ def convert_unary(unary):
         return convert_unary(nested_unary)
     elif unary.SUB():
         numabs = convert_unary(nested_unary)
-        if numabs == 1:
-            # Use Integer(-1) instead of Mul(-1, 1)
-            return -numabs
-        else:
-            return sympy.Mul(-1, convert_unary(nested_unary), evaluate=False)
+        # Use Integer(-n) instead of Mul(-1, n)
+        return -numabs
     elif postfix:
         return convert_postfix_list(postfix)
```

wermos:
I thought I'd mention two things:
1. Your very first example actually has an extra closing parenthesis. ANTLR doesn't seem to mind too much but the Lark-based parser complains (as it should, imo):
```python
>>> srepr(parse_latex_lark('1 - 2*(x + 1))'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/tirthankar/projects/sympy/sympy/parsing/latex/lark/latex_parser.py", line 76, in parse_latex_lark
    return _lark_latex_parser.doparse(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tirthankar/projects/sympy/sympy/parsing/latex/lark/latex_parser.py", line 40, in doparse
    parse_tree = self.parser.parse(s)
                 ^^^^^^^^^^^^^^^^^^^^
[...]

PR #21518, opened 2021-05-24
`_keep coeff` is more careful about not evaluating expressions

smichr:

solves part of #21517

In some cases, `_keep_coeff` evaluates portions of the expression which were not already evaluated. More care is now taken to avoid that (as shown in the added tests). But initial failings show that care must be taken to allow autoevaluation in some cases (like multiplying Order by a constant that gets absorbed).

#### Other comments

`cse` now processes unevaluated Mul without error, too. Formerly,
```python
>>> cse(Mul(1, 2, evaluate=False))
...
ValueError: max() arg is an empty sequence
```
#### Release Notes

* parsing
  *  `parse_latex` does not introduce a factor of -1 on numbers
* core
  * `_keep_coeff` will distribute coeff on non-rational if doing so does not produce a rational
* simplify
  * `cse` will no longer generate an error when passed an unevaluated 2-arg Mul with a Rational arg

### DIFF
--- a/sympy/core/mul.py
+++ b/sympy/core/mul.py
@@ -2050,24 +2050,24 @@ def _keep_coeff(coeff, factors, clear=True, sign=False):
     >>> _keep_coeff(S(-1), x + y, sign=True)
     -(x + y)
     """
-
     if not coeff.is_Number:
         if factors.is_Number:
             factors, coeff = coeff, factors
         else:
             return coeff*factors
+    if factors is S.One:
+        return coeff
     if coeff is S.One:
         return factors
     elif coeff is S.NegativeOne and not sign:
         return -factors
     elif factors.is_Add:
         if not clear and coeff.is_Rational and coeff.q != 1:
-            q = S(coeff.q)
-            for i in factors.args:
-                c, t = i.as_coeff_Mul()
-                r = c/q
-                if r == int(r):
-                    return coeff*factors
+            args = [i.as_coeff_Mul() for i in factors.args]
+            args = [(c*coeff, m) for c, m in args]
+            if any(c == int(c) for c, _ in args):
+                return Add._from_args([Mul._from_args(
+                    i[1:] if i[0] == 1 else i) for i in args])
         return Mul(coeff, factors, evaluate=False)
     elif factors.is_Mul:
         margs = list(factors.args)
@@ -2079,7 +2079,7 @@ def _keep_coeff(coeff, factors, clear=True, sign=False):
             margs.insert(0, coeff)
         return Mul._from_args(margs)
     else:
-        return coeff*factors
+        return Mul._from_args((coeff, factors))
 
 
 def expand_2arg(e):
--- a/sympy/core/mul.py
+++ b/sympy/core/mul.py
@@ -2064,7 +2064,7 @@ def _keep_coeff(coeff, factors, clear=True, sign=False):
     elif factors.is_Add:
         if not clear and coeff.is_Rational and coeff.q != 1:
             args = [i.as_coeff_Mul() for i in factors.args]
-            args = [(c*coeff, m) for c, m in args]
+            args = [(_keep_coeff(c, coeff), m) for c, m in args]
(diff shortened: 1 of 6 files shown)


sympy/core/mul.py
```diff
@@ -2079,8 +2079,10 @@ def _keep_coeff(coeff, factors, clear=True, sign=False):
             margs.insert(0, coeff)
         return Mul._from_args(margs)
     else:
-        return coeff*factors
-
+        m = coeff*factors
```
smichr, line 2082:
```suggestion
        m = -factors if (coeff is S.NegativeOne and not sign) else coeff*factors
```

sympy/core/mul.py
```diff
@@ -2050,24 +2050,22 @@ def _keep_coeff(coeff, factors, clear=True, sign=False):
...
             factors, coeff = coeff, factors
         else:
             return coeff*factors
+    if factors is S.One:
+        return coeff
     if coeff is S.One:
         return factors
-    elif coeff is S.NegativeOne and not sign:
-        return -factors
     elif factors.is_Add:
```
smichr, line 2062:
```suggestion
    elif coeff is S.NegativeOne and not sign:
        if factors.is_Number of factors.is_Add:
            return -factors
        return Mul(coeff, factors, evaluate=False)
    elif factors.is_Add:
```

sympy/core/mul.py
```diff
@@ -2050,24 +2050,26 @@ def _keep_coeff(coeff, factors, clear=True, sign=False):
...
             factors, coeff = coeff, factors
         else:
             return coeff*factors
+    if factors is S.One:
+        return coeff
     if coeff is S.One:
         return factors
     elif coeff is S.NegativeOne and not sign:
-        return -factors
+        if factors.is_Number or factors.is_Add:
```
smichr, line 2063:
```suggestion
        if factors.is_Number or factors.is_Add or factors.as_coeff_Mul()[0].is_Number:
```

sympy/core/mul.py
```diff
@@ -2050,24 +2050,26 @@ def _keep_coeff(coeff, factors, clear=True, sign=False):
...
             factors, coeff = coeff, factors
         else:
             return coeff*factors
+    if factors is S.One:
+        return coeff
     if coeff is S.One:
         return factors
     elif coeff is S.NegativeOne and not sign:
-        return -factors
+        if factors.is_Number or factors.is_Add or factors.as_coeff_Mul()[0].is_Number:
```
smichr, line 2063:
```suggestion
        if factors.is_Number or factors.is_Add or factors.as_coeff_Mul()[0] is not S.One:
```

sympy/core/mul.py
```diff
@@ -2050,24 +2050,26 @@ def _keep_coeff(coeff, factors, clear=True, sign=False):
...
         else:
             return coeff*factors
+    if factors is S.One:
+        return coeff
     if coeff is S.One:
         return factors
     elif coeff is S.NegativeOne and not sign:
-        return -factors
+        if factors.is_Number or factors.is_Add or factors.as_coeff_Mul()[0] is not S.One:
+            return -factors
```
smichr, line 2064:
```suggestion
        m = -factors
        if m.as_coeff_Mul()[0] in (S.NegativeOne, S.One):
            return m
```

sympy/core/mul.py
```diff
@@ -2050,24 +2050,27 @@ def _keep_coeff(coeff, factors, clear=True, sign=False):
...
+    if factors is S.One:
+        return coeff
     if coeff is S.One:
         return factors
     elif coeff is S.NegativeOne and not sign:
-        return -factors
+        m = -factors
+        if m.as_coeff_Mul()[0] in (S.NegativeOne, S.One):
+            return m
+        return Mul._from_args((coeff, factors))
```
smichr, line 2066:
```suggestion
        return -factors
```

smichr:
I will clean up commit history and commit tonight if there are no objections.

PR #21531, opened 2021-05-27
Fix _FuncMinusOne in codegen/rewriting.py

lcontento:
Code optimizations like `exp(x) - 1 --> expm1(x)` may give unusable results in some situations.
For example:
```python
import sympy as sp
from sympy.codegen.rewriting import optimize, expm1_opt
expr = sp.sympify('(2*exp(x) + 1)/(exp(x) + 1) + 1')
optimize(expr, [expm1_opt])
```
returns `(_Dummy_25 + 2*exp(x) + 2)/(_Dummy_25 + 1)`.

This appears to be a bug in the `_try_func_m_1` method of `_FuncMinusOne`.
If multiple instances of `self.func` with the same argument are present, they are assigned different `Dummy` symbols.
However, only one of these is saved in the `old_new` dictionary, making back substitution impossible.

#### Release Notes

* codegen
  * Fixed a bug in the code optimization for `expm1`

### DIFF
--- a/sympy/codegen/rewriting.py
+++ b/sympy/codegen/rewriting.py
@@ -181,9 +181,10 @@ class _FuncMinusOne:
         self.func_m_1 = func_m_1
 
     def _try_func_m_1(self, expr):
-        protected, old_new =  expr.replace(self.func, lambda arg: Dummy(), map=True)
+        old_new = {}
+        protected = expr.replace(self.func, lambda arg: old_new.setdefault(arg, Dummy()))
         factored = protected.factor()
-        new_old = {v: k for k, v in old_new.items()}
+        new_old = {v: self.func(k) for k, v in old_new.items()}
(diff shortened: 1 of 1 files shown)


sympy/codegen/rewriting.py
```diff
@@ -181,9 +181,10 @@ def __init__(self, func, func_m_1):
         self.func_m_1 = func_m_1
 
     def _try_func_m_1(self, expr):
-        protected, old_new =  expr.replace(self.func, lambda arg: Dummy(), map=True)
+        old_new = {}
+        protected = expr.replace(self.func, lambda arg: old_new.setdefault(arg, Dummy()))
```
smichr, line 185: Beautiful application of this dict method!

lcontento: Thanks! :smiley:

Issue #21533, opened 2021-05-28
Multiple integrate problem


UlBoy:
![ssmrp](https://user-images.githubusercontent.com/84957780/119941579-c340c400-bfcb-11eb-9c35-bdff5a0d9b76.jpg)

Hello ! I'm new to using sympy and python. 
I'm trying to solve a multiple integrate problem include 2 variables.
But there are some TypeError and I don't know how to solve this.
```
#Codes
from sympy import *
init_printing()


z1 = symbols('z1')
z2 = symbols('z2')
I = integrate(exp(-0.5*(((-1/z1)+1.39)**2+((-1/z2)+1.23)**2))*(1/(z1**2))(1/(z2**2)),(z1,0,1),(z2,0,1))
print (I)

#results
PS C:\Users\6241\Documents\python\ssmrp> python inte.py
Traceback (most recent call last):
  File "C:\Users\6241\Documents\python\ssmrp\inte.py", line 9, in <module>
    I = integrate(exp(-0.5*(((-1/z1)+1.39)**2+((-1/z2)+1.23)**2))*(1/(z1**2))(1/(z2**2)),(z1,0,1),(z2,0,1))
TypeError: 'Pow' object is not callable
```

Thank you for your helps!!

jksuom:
> `(1/(z1**2))(1/(z2**2))`

There should be a `*` between the factors.

UlBoy:
![image](https://user-images.githubusercontent.com/84957780/119970020-39542380-bfea-11eb-84e6-1dc41ccda33a.png)
Yes I already put '*'. but That's what it looks like on this site.

UlBoy:
Thanks a lot!
I try to do again after adding '*' on my codes.
But It is not calculated and comes out as an expression.
Is the expression itself wrong?

#results
PS C:\Users\6241\Documents\python\ssmrp> python inte.py
0.1786190416574*Integral(exp(-0.5/z1**2)*exp(1.39/z1)/z1**2, (z1, 0, 1))*Integral(exp(-0.5/z2**2)*exp(1.23/z2)/z2**2, (z2, 0, 1))

oscarbenjamin:
It looks like an analytic result can not be computed for this integral currently. It should be possible to evaluate it numerically with `evalf` except that double integrals are not supported (even though mpmath has a routine for double integrals that could be used).

This particular integral can be factored into a product of two single integrals that can be evaluated numerically:
```python
[...]

PR #21543, opened 2021-05-28
perf(polys): make EX faster for trivial operations

oscarbenjamin:
This commit avoids calls to cancel for some trivial operations in the EX
domain. This can lead to a big speed up for some integrals. For example
```
z1, z2 = symbols('z1, z2')
I = Integral(exp(-0.5*(((-1/z1)+1.39)**2+((-1/z2)+1.23)**2))*(1/(z1**2))*(1/(z2**2)),(z1,0,1),(z2,0,1))
I.doit()
```
In this case the integral is not computed with or without this change
but heurisch completes much faster with the change reducing the overall
time by 50%. Previously most of the time was spent in trivial calls to
cancel like cancel(0), cancel(3), etc.

The integral example here is from #21533

The EX arithmetic is changed to reduce unnecessary calls to cancel for trivial cases like `EX(0)*EX(1)`.

#### Other comments

Further improvements could be made but it looks like a bit of a restructure of `ExtensionElement.__new__` would be needed. Ideally all EX elements would be canonicalised at creation time and then further calls to cancel could be eliminated from internal routines.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/polys/domains/expressiondomain.py
+++ b/sympy/polys/domains/expressiondomain.py
@@ -64,10 +64,14 @@ class ExpressionDomain(Field, CharacteristicZero, SimpleDomain):
         def __add__(f, g):
             g = f._to_ex(g)
 
-            if g is not None:
-                return f.simplify(f.ex + g.ex)
-            else:
+            if g is None:
                 return NotImplemented
+            elif g == EX.zero:
+                return f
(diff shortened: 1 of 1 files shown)


jksuom:
Looks good.

PR #21545, opened 2021-05-29
restrict Laplace space var to be strictly complex

namannimmo10:
This restricts the Laplace space input variable in `TransferFunction` to be a strictly complex `Symbol`.

#### Release Notes

* physics.control
  *  Laplace space input variable will be strictly complex

### DIFF
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -148,8 +148,8 @@ class TransferFunction(Basic, EvalfMixin):
     def __new__(cls, num, den, var):
         num, den = _sympify(num), _sympify(den)
 
-        if not isinstance(var, Symbol):
-            raise TypeError("Variable input must be a Symbol.")
+        if not isinstance(var, Symbol) or var.is_real:
+            raise TypeError("Variable input must be a strictly complex symbol.")
         if den == 0:
             raise ValueError("TransferFunction can't have a zero denominator.")
 
--- a/sympy/physics/control/tests/test_lti.py
+++ b/sympy/physics/control/tests/test_lti.py
@@ -1,4 +1,4 @@
-from sympy import symbols, Matrix, factor, Function, simplify, exp, pi, oo, I, \
+from sympy import symbols, Symbol, Matrix, factor, Function, simplify, exp, pi, oo, I, \
     Rational, sqrt, CRootOf
 from sympy.physics.control.lti import TransferFunction, Series, Parallel, Feedback
 from sympy.testing.pytest import raises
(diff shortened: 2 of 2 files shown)


sympy/physics/control/lti.py
```diff
@@ -148,8 +148,8 @@ class TransferFunction(Basic, EvalfMixin):
     def __new__(cls, num, den, var):
         num, den = _sympify(num), _sympify(den)
 
-        if not isinstance(var, Symbol):
-            raise TypeError("Variable input must be a Symbol.")
+        if not isinstance(var, Symbol) or var.is_real:
+            raise TypeError("Variable input must be a strictly complex symbol.")
```
oscarbenjamin, line 152: Should this be `is_complex` rather than `is_real`?

oscarbenjamin:
The general approach in sympy is that if a symbol does not need to have assumptions explicitly set. If the explicit assumptions contradict the context in which the symbols are used then an error can be raised but we don't normally raise an error for a `Symbol('x')` when used in a context where the symbol would be expected to be integer, real etc.

sympy/physics/control/lti.py
namannimmo10: It's true that the reals are a subset of complex numbers. 

> The general approach in sympy is that if a symbol does not need to have assumptions explicitly set.

I agree -- we don't check whether the symbols are real or complex in the polynomial module, for instance.

We want to support symbols with the `complex=True` assumption, and disallow those with the `real=True` assumption. So I figured out two ways to approach this: either check only if `var` is a `Symbol` (no assumption on it whatsoever) or use `var.is_real` (`not var.is_complex` won't work).

namannimmo10: ping @akshanshbhatt & @moorepants here

akshanshbhatt: @namannimmo10, I agree with you. Checking with the `is_complex` assumption won't assure that the symbol is not real, and we don't want that anyway. By 'strictly complex', I meant that the user should be able to substitute any complex number into the `TransferFunction` and get the result. Sympy symbols don't have any sort of "domain" property, so with any number of assumptions while defining the symbol, `subs` and `roots` would return the output considering `CC` as the domain. The only issue will be the fact that `solve()` would return the solutions tailored to these assumptions. Here is what I am trying to say-

```python
>>> from sympy import *
>>> x = Symbol("x", real=True)
>>> n = 5*x**3 + 6*x**2 + 9*x + 8
>>> solve(k)  # It will only return real solution(s)
[-1]
>>> roots(n)
{-1: 1, -1/10 - sqrt(159)*I/10: 1, -1/10 + sqrt(159)*I/10: 1}
>>> n.subs({x: 5*I+4})  # We can substitute even if it is not in accordance with the assumption
44 + 45*I + 6*(4 + 5*I)**2 + 5*(4 + 5*I)**3
```
Apart from `solve()`, these assumptions won't cause any hindrance.

oscarbenjamin:
I expect that there is a better way to do this but I don't really understand the purpose of this PR.

moorepants:
> The general approach in sympy is that if a symbol does not need to have assumptions explicitly set.

Best to stick to this convention.

Issue #18916, opened 2020-03-20
No option to use Common Sub-expression Detection in 'lambdify'


fh9g12:
This is less an issue and more a question / potential lack of understanding on my part. And I thought this was probably the best Forum for it....

A bit of context - I am using Sympy to build up the equations of motion of a flexible multi-link manipulator. As part of this I have matrices with a lot of sin / cos evaluations. 

When numerically integrating my system that largest bottleneck was in evaluating the generated 'Lambda expressions' of these matricies. Below is a snippet of the %prun output showing the lambda bottleneck

![image](https://user-images.githubusercontent.com/15157018/77167117-4a7fd700-6aad-11ea-99e5-5f7e447db02b.png)

To resolve this I have 'monkey patched' the 'doprint' method in the _EvaluatorPrinter class (found in the lambdify.py file) with the following method. The additional lines (in comment blocks) use the 'cse' method to split out the common sub expressions, I then add this to the generated function definition

`
def doprint(self, funcname, args, expr):
        """Returns the function definition code as a string."""
        from sympy import Dummy,cse,Symbol

        funcbody = []

        if not iterable(args):
            args = [args]

        argstrs, expr = self._preprocess(args, expr)

        ## --------------- Addition -----------------
        replacments, exprs = cse(expr,symbols=(Symbol(f'rep_{i}')for i in range(100)))
        expr = exprs[0]
        ## --------------- Addition -----------------

        # Generate argument unpacking and final argument list
        funcargs = []
        unpackings = []

        for argstr in argstrs:
            if iterable(argstr):
                funcargs.append(self._argrepr(Dummy()))
                unpackings.extend(self._print_unpacking(argstr, funcargs[-1]))
            else:
                funcargs.append(argstr)
        arg_vars = ', '.join(funcargs)
        funcsig = f'def {funcname}({arg_vars}):'

        # Wrap input arguments before unpacking
        funcbody.extend(self._print_funcargwrapping(funcargs))

        funcbody.extend(unpackings)

        ## --------------- Addition -----------------
        for variable, expression in replacments:
            funcbody.append(f'{variable} = {expression}')
        ## --------------- Addition -----------------

        funcbody.append('return ({})'.format(self._exprrepr(expr)))

        funclines = [funcsig]
        funclines.extend('    ' + line for line in funcbody)

        return '\n'.join(funclines) + '\n'
` 

This makes some of my lambda functions up to 7 times faster, as shown (rather qualativly) in the image below
 
[...]

PR #21546, opened 2021-05-30
lambdify: Optimization with a substitution of redundant patterns

robinechuca:
In the event that we want to vectorize big sympathetic expressions. Often this one comes from a sympy algorithm which does not give very compact equations. So there is very often a lot of redundancy.
The modification that I propose is to make "change of variables" in order to calculate only once the redundant patterns.

* utilities
  * lambdify has keyword `cse` which allows for more efficient calculation of expressions with highly-repeated subexpressions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,3 @@ sample.tex
 
 # pytest related data file for slow tests
 .ci/durations.log
-
-test_lambdify.py
-
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,6 @@ sample.tex
 
 # pytest related data file for slow tests
 .ci/durations.log
+
+test_lambdify.py
+
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -121,6 +121,9 @@ def cse_minimize_memory(r, e):
     >>> print(rvs)
     (_0, _1, _2, _3, _4)
     """
+    if not r:
+        return r, e
+
     from sympy import symbols
 
     s, p = zip(*r)
@@ -692,7 +695,7 @@ def tree_cse(exprs, symbols, opt_subs=None, order='canonical', ignore=()):
 
 
 def cse(exprs, symbols=None, optimizations=None, postprocess=None,
-        order='canonical', ignore=(), asList=True):
+        order='canonical', ignore=(), as_list=True):
     """ Perform common subexpression elimination on an expression.
 
     Parameters
@@ -723,7 +726,7 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
         concern, use the setting order='none'.
     ignore : iterable of Symbols
         Substitutions containing any Symbol from ``ignore`` will be ignored.
-    asList : bool, (default True)
+    as_list : bool, (default True)
         Returns expression in list or else with same type as input (when False).
 
     Returns
@@ -771,7 +774,7 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     from sympy.matrices import (MatrixBase, Matrix, ImmutableMatrix,
                                 SparseMatrix, ImmutableSparseMatrix)
 
-    if not asList:
+    if not as_list:
         return _cse_homogeneous(exprs,
             symbols=symbols, optimizations=optimizations,
             postprocess=postprocess, order=order, ignore=ignore)
@@ -869,12 +872,10 @@ def _cse_homogeneous(exprs, **kwargs):
 
     Examples
     ========
-    >>> from sympy.simplify.cse_main import _cse_homogeneous
-    >>> from sympy import cos, Tuple
+    >>> from sympy.simplify.cse_main import cse
+    >>> from sympy import cos, Tuple, Matrix
     >>> from sympy.abc import x
-    >>> _cse_homogeneous('[cos(x), cos(x) + 1]')
-    ([(x0, cos(x))], '[x0, x0 + 1]')
-    >>> output = lambda x: type(_cse_homogeneous(x)[1])
+    >>> output = lambda x: type(cse(x, as_list=False)[1])
     >>> output(1)
     <class 'sympy.core.numbers.One'>
     >>> output('cos(x)')
@@ -883,6 +884,8 @@ def _cse_homogeneous(exprs, **kwargs):
     cos
     >>> output(Tuple(1, x))
     <class 'sympy.core.containers.Tuple'>
+    >>> output(Matrix([[1,0], [0,1]]))
+    <class 'sympy.matrices.dense.MutableDenseMatrix'>
     >>> output([1, x])
     <class 'list'>
     >>> output((1, x))
@@ -890,9 +893,8 @@ def _cse_homogeneous(exprs, **kwargs):
     >>> output({1, x})
     <class 'set'>
     """
-    from sympy import sympify
-
     if isinstance(exprs, str):
+        from sympy import sympify
         replacements, reduced_exprs = _cse_homogeneous(
             sympify(exprs), **kwargs)
         return replacements, repr(reduced_exprs)
@@ -902,7 +904,7 @@ def _cse_homogeneous(exprs, **kwargs):
     if isinstance(exprs, dict):
         keys = list(exprs.keys()) # In order to guarantee the order of the elements.
         replacements, values = cse([exprs[k] for k in keys], **kwargs)
-        reduced_exprs = {k: v for k, v in zip(keys, values)}
+        reduced_exprs = dict(zip(keys, values))
         return replacements, reduced_exprs
 
     try:
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -89,7 +89,7 @@ def cse_separate(r, e):
 
 def cse_minimize_memory(r, e):
     """
-    Return tuples giving (a, b) where ``a`` is a symbol and b is
+    Return tuples giving (a, b) where ``a`` is a symbol and ``b`` is
     either an expression or None. The value of None is used when a
     symbol is no longer needed for subsequent expressions.
 
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -806,7 +806,7 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     del temp
 
     if optimizations is None:
-        optimizations = list()
+        optimizations = []
     elif optimizations == 'basic':
         optimizations = basic_optimizations
 
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -744,6 +744,7 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
 
     >>> from sympy import cse, SparseMatrix
     >>> from sympy.abc import x, y, z, w
+    >>> from sympy.simplify.cse_main import cse
     >>> cse(((w + x + y + z)*(w + y + z))/(w + x)**3)
     ([(x0, y + z), (x1, w + x)], [(w + x0)*(x0 + x1)/x1**3])
 
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -742,9 +742,8 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     Examples
     ========
 
-    >>> from sympy import cse, SparseMatrix
+    >>> from sympy import cse, SparseMatrix, cos
     >>> from sympy.abc import x, y, z, w
-    >>> from sympy.simplify.cse_main import cse
     >>> cse(((w + x + y + z)*(w + y + z))/(w + x)**3)
     ([(x0, y + z), (x1, w + x)], [(w + x0)*(x0 + x1)/x1**3])
 
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -89,7 +89,7 @@ def cse_separate(r, e):
 
 def cse_minimize_memory(r, e):
     """
-    Return tuples giving (a, b) where ``a`` is a symbol and ``b`` is
+    Return tuples giving ``(a, b)`` where ``a`` is a symbol and ``b`` is
     either an expression or None. The value of None is used when a
     symbol is no longer needed for subsequent expressions.
 
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -770,6 +770,12 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     >>> cse([y**2*(x + 1), 3*y**2*(x + 1)], ignore=(y,))
     ([(x0, x + 1)], [x0*y**2, 3*x0*y**2])
 
+    >>> output = lambda x, as_list: type(cse(x, as_list=as_list)[1])
+    >>> output(cos(x), as_list=False)
+    cos
+    >>> output(cos(x), as_list=True)
+    <class 'list'>
+    >>>
     """
     from sympy.matrices import (MatrixBase, Matrix, ImmutableMatrix,
                                 SparseMatrix, ImmutableSparseMatrix)
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -88,14 +88,13 @@ def cse_separate(r, e):
 
 
 def cse_minimize_memory(r, e):
-    """return tuples giving (a, b) where ``a`` is either
-        * a cse symbol,
-        * a symbol assigned to each of the original expressions, or
-        * None
-    The list of symbols assigned to original expressions is returned as the last tuple.
+    """
+    Return tuples giving (a, b) where ``a`` is a symbol and b is
+    either an expression or None. The value of None is used when a
+    symbol is no longer needed for subsequent expressions.
 
-    When the first element of a tuple is None, it indicates that the variables
-    given as a list as the second element of the tuple are not needed by any sub-expression.
+    Use of such output can reduce the memory footprint of lambdified
+    expressions that contain large, repeated subexpressions.
 
     Examples
     ========
@@ -128,23 +127,22 @@ def cse_minimize_memory(r, e):
     esyms = symbols('_:%d' % len(e))
     syms = list(esyms)
     s = list(s)
-    ss = set(s)
+    in_use = set(s)
     p = list(p)
     # sort e so those with most sub-expressions appear first
     e = [(e[i], syms[i]) for i in range(len(e))]
     e, syms = zip(*sorted(e,
-        key=lambda x: -sum([p[s.index(i)].count_ops() for i in x[0].free_symbols & ss])))
+        key=lambda x: -sum([p[s.index(i)].count_ops()
+        for i in x[0].free_symbols & in_use])))
     syms = list(syms)
     p += e
-    in_use = ss
-    # del ss # makes failing the quality code test.
     rv = []
     i = len(p) - 1
     while i >= 0:
         _p = p.pop()
         c = in_use & _p.free_symbols
-        if c:
-            rv.extend([(s, None) for s in sorted(c, key=str)]) # sorting for canonical results
+        if c: # sorting for canonical results
+            rv.extend([(s, None) for s in sorted(c, key=str)])
         if i >= len(r):
             rv.append((syms.pop(), _p))
         else:
@@ -183,8 +181,8 @@ def preprocess_for_cse(expr, optimizations):
 
 
 def postprocess_for_cse(expr, optimizations):
-    """ Postprocess an expression after common subexpression elimination to
-    return the expression to canonical sympy form.
+    """Postprocess an expression after common subexpression elimination to
+    return the expression to canonical SymPy form.
 
     Parameters
     ==========
@@ -450,7 +448,6 @@ def match_common_args(func_class, funcs, opt_subs):
         arg_tracker.stop_arg_tracking(i)
 
 
-
 def opt_cse(exprs, order='canonical'):
     """Find optimization opportunities in Adds, Muls, Pows and negative
     coefficient Muls.
@@ -695,7 +692,7 @@ def tree_cse(exprs, symbols, opt_subs=None, order='canonical', ignore=()):
 
 
 def cse(exprs, symbols=None, optimizations=None, postprocess=None,
-        order='canonical', ignore=(), memory=False):
+        order='canonical', ignore=(), asList=True):
     """ Perform common subexpression elimination on an expression.
 
     Parameters
@@ -726,9 +723,8 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
         concern, use the setting order='none'.
     ignore : iterable of Symbols
         Substitutions containing any Symbol from ``ignore`` will be ignored.
-    memory : boolean
-        Allows to activate the search for removable symbols.
-        Seealso ``cse_minimize_memory``.
+    asList : bool, (default True)
+        Returns expression in list or else with same type as input (when False).
 
     Returns
     =======
@@ -775,10 +771,10 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     from sympy.matrices import (MatrixBase, Matrix, ImmutableMatrix,
                                 SparseMatrix, ImmutableSparseMatrix)
 
-    if memory:
-        return cse_minimize_memory(*cse(exprs,
+    if not asList:
+        return _cse_homogeneous(exprs,
             symbols=symbols, optimizations=optimizations,
-            postprocess=postprocess, order=order, ignore=ignore))
+            postprocess=postprocess, order=order, ignore=ignore)
 
     if isinstance(exprs, (int, float)):
         exprs = sympify(exprs)
@@ -851,7 +847,8 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
 
 def _cse_homogeneous(exprs, **kwargs):
     """
-    Same as ``cse`` but the reduced_exprs are expres in the same type as 'exprs'
+    Same as ``cse`` but the ``reduced_exprs`` are returned
+    with the same type as ``exprs`` or a sympified version of the same.
 
     Parameters
     ==========
@@ -872,29 +869,32 @@ def _cse_homogeneous(exprs, **kwargs):
 
     Examples
     ========
-
     >>> from sympy.simplify.cse_main import _cse_homogeneous
-    >>> from sympy import Symbol, cos, Tuple
-    >>> x = Symbol("x")
-    >>> _cse_homogeneous("cos(x + 1)") # str
-    ([], 'cos(x + 1)')
-    >>> _cse_homogeneous(cos(x + 1)) # Basic
-    ([], cos(x + 1))
-    >>> _cse_homogeneous(1) # number
-    ([], 1)
-    >>> _cse_homogeneous(Tuple(1, x))
-    ([], (1, x))
-    >>> _cse_homogeneous([1, x])
-    ([], [1, x])
-    >>> _cse_homogeneous((1, x))
-    ([], (1, x))
-    >>> _cse_homogeneous({1, x})
-    ([], {1, x})
+    >>> from sympy import cos, Tuple
+    >>> from sympy.abc import x
+    >>> _cse_homogeneous('[cos(x), cos(x) + 1]')
+    ([(x0, cos(x))], '[x0, x0 + 1]')
+    >>> output = lambda x: type(_cse_homogeneous(x)[1])
+    >>> output(1)
+    <class 'sympy.core.numbers.One'>
+    >>> output('cos(x)')
+    <class 'str'>
+    >>> output(cos(x))
+    cos
+    >>> output(Tuple(1, x))
+    <class 'sympy.core.containers.Tuple'>
+    >>> output([1, x])
+    <class 'list'>
+    >>> output((1, x))
+    <class 'tuple'>
+    >>> output({1, x})
+    <class 'set'>
     """
     from sympy import sympify
 
     if isinstance(exprs, str):
-        replacements, reduced_exprs = _cse_homogeneous(sympify(exprs), **kwargs)
+        replacements, reduced_exprs = _cse_homogeneous(
+            sympify(exprs), **kwargs)
         return replacements, repr(reduced_exprs)
     if isinstance(exprs, (list, tuple, set)):
         replacements, reduced_exprs = cse(exprs, **kwargs)
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -853,8 +853,8 @@ def _cse_homogeneous(exprs, **kwargs):
     Parameters
     ==========
 
-    exprs : list of sympy expressions, or a single SymPy expression
-        The expressions to reduce.
+    exprs : an Expr, iterable of Expr or dictionary with Expr values
+        the expressions in which repeated subexpressions will be identified
     kwargs : additional arguments for the ``cse`` function
 
     Returns
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -770,12 +770,13 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     >>> cse([y**2*(x + 1), 3*y**2*(x + 1)], ignore=(y,))
     ([(x0, x + 1)], [x0*y**2, 3*x0*y**2])
 
-    >>> output = lambda x, as_list: type(cse(x, as_list=as_list)[1])
-    >>> output(cos(x), as_list=False)
-    cos
-    >>> output(cos(x), as_list=True)
-    <class 'list'>
-    >>>
+    The default return value for the reduced expression(s) is a list, even if there is only
+    one expression. The `as_list` flag preserves the type of the input in the output:
+
+    >>> cse(x)
+    ([], [x])
+    >>> cse(x, as_list=False)
+    ([], x)
     """
     from sympy.matrices import (MatrixBase, Matrix, ImmutableMatrix,
                                 SparseMatrix, ImmutableSparseMatrix)
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -742,7 +742,7 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     Examples
     ========
 
-    >>> from sympy import cse, SparseMatrix, cos
+    >>> from sympy import cse, SparseMatrix
     >>> from sympy.abc import x, y, z, w
     >>> cse(((w + x + y + z)*(w + y + z))/(w + x)**3)
     ([(x0, y + z), (x1, w + x)], [(w + x0)*(x0 + x1)/x1**3])
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -137,10 +137,9 @@ def cse_minimize_memory(r, e):
     syms = list(syms)
     p += e
     in_use = ss
-    del ss
+    # del ss # makes failing the quality code test.
     rv = []
     i = len(p) - 1
-    useless = {}
     while i >= 0:
         _p = p.pop()
         c = in_use & _p.free_symbols
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -695,7 +695,7 @@ def tree_cse(exprs, symbols, opt_subs=None, order='canonical', ignore=()):
 
 
 def cse(exprs, symbols=None, optimizations=None, postprocess=None,
-        order='canonical', ignore=(), as_list=True):
+        order='canonical', ignore=(), list=True):
     """ Perform common subexpression elimination on an expression.
 
     Parameters
@@ -726,7 +726,7 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
         concern, use the setting order='none'.
     ignore : iterable of Symbols
         Substitutions containing any Symbol from ``ignore`` will be ignored.
-    as_list : bool, (default True)
+    list : bool, (default True)
         Returns expression in list or else with same type as input (when False).
 
     Returns
@@ -771,17 +771,17 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     ([(x0, x + 1)], [x0*y**2, 3*x0*y**2])
 
     The default return value for the reduced expression(s) is a list, even if there is only
-    one expression. The `as_list` flag preserves the type of the input in the output:
+    one expression. The `list` flag preserves the type of the input in the output:
 
     >>> cse(x)
     ([], [x])
-    >>> cse(x, as_list=False)
+    >>> cse(x, list=False)
     ([], x)
     """
     from sympy.matrices import (MatrixBase, Matrix, ImmutableMatrix,
                                 SparseMatrix, ImmutableSparseMatrix)
 
-    if not as_list:
+    if not list:
         return _cse_homogeneous(exprs,
             symbols=symbols, optimizations=optimizations,
             postprocess=postprocess, order=order, ignore=ignore)
@@ -879,10 +879,11 @@ def _cse_homogeneous(exprs, **kwargs):
 
     Examples
     ========
+
     >>> from sympy.simplify.cse_main import cse
     >>> from sympy import cos, Tuple, Matrix
     >>> from sympy.abc import x
-    >>> output = lambda x: type(cse(x, as_list=False)[1])
+    >>> output = lambda x: type(cse(x, list=False)[1])
     >>> output(1)
     <class 'sympy.core.numbers.One'>
     >>> output('cos(x)')
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -86,6 +86,76 @@ def cse_separate(r, e):
     e = d[False]
     return [reps_toposort(r), e]
 
+
+def cse_minimize_memory(r, e):
+    """return tuples giving (a, b) where ``a`` is either
+        * a cse symbol,
+        * a symbol assigned to each of the original expressions, or
+        * None
+    The list of symbols assigned to original expressions is returned as the last tuple.
+
+    When the first element of a tuple is None, it indicates that the variables
+    given as a list as the second element of the tuple are not needed by any sub-expression.
+
+    Examples
+    ========
+
+    >>> from sympy import cse
+    >>> from sympy.simplify.cse_main import cse_minimize_memory
+    >>> from sympy.abc import x, y
+    >>> eqs = [(x + y - 1)**2, x, x + y, (x + y)/(2*x + 1) + (x + y - 1)**2, (2*x + 1)**(x + y)]
+    >>> defs, rvs = cse_minimize_memory(*cse(eqs))
+    >>> for i in defs:
+    ...   print(i)
+    ...
+    (x0, x + y)
+    (x1, (x0 - 1)**2)
+    (x2, 2*x + 1)
+    (_3, x0/x2 + x1)
+    (_4, x2**x0)
+    (x2, None)
+    (_0, x1)
+    (x1, None)
+    (_2, x0)
+    (x0, None)
+    (_1, x)
+    >>> print(rvs)
+    (_0, _1, _2, _3, _4)
+    """
+    from sympy import symbols
+
+    s, p = zip(*r)
+    esyms = symbols('_:%d' % len(e))
+    syms = list(esyms)
+    s = list(s)
+    ss = set(s)
+    p = list(p)
+    # sort e so those with most sub-expressions appear first
+    e = [(e[i], syms[i]) for i in range(len(e))]
+    e, syms = zip(*sorted(e,
+        key=lambda x: -sum([p[s.index(i)].count_ops() for i in x[0].free_symbols & ss])))
+    syms = list(syms)
+    p += e
+    in_use = ss
+    del ss
+    rv = []
+    i = len(p) - 1
+    useless = {}
+    while i >= 0:
+        _p = p.pop()
+        c = in_use & _p.free_symbols
+        if c:
+            rv.extend([(s, None) for s in sorted(c, key=str)]) # sorting for canonical results
+        if i >= len(r):
+            rv.append((syms.pop(), _p))
+        else:
+            rv.append((s[i], _p))
+        in_use -= c
+        i -= 1
+    rv.reverse()
+    return rv, esyms
+
+
 # ====end of cse postprocess idioms===========================
 
 
@@ -626,7 +696,7 @@ def tree_cse(exprs, symbols, opt_subs=None, order='canonical', ignore=()):
 
 
 def cse(exprs, symbols=None, optimizations=None, postprocess=None,
-        order='canonical', ignore=()):
+        order='canonical', ignore=(), memory=False):
     """ Perform common subexpression elimination on an expression.
 
     Parameters
@@ -657,6 +727,9 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
         concern, use the setting order='none'.
     ignore : iterable of Symbols
         Substitutions containing any Symbol from ``ignore`` will be ignored.
+    memory : boolean
+        Allows to activate the search for removable symbols.
+        Seealso ``cse_minimize_memory``.
 
     Returns
     =======
@@ -678,7 +751,7 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
 
     Note that currently, y + z will not get substituted if -y - z is used.
 
-     >>> cse(((w + x + y + z)*(w - y - z))/(w + x)**3)
+    >>> cse(((w + x + y + z)*(w - y - z))/(w + x)**3)
      ([(x0, w + x)], [(w - y - z)*(x0 + y + z)/x0**3])
 
     List of expressions with recursive substitutions:
@@ -702,6 +775,11 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     """
     from sympy.matrices import (MatrixBase, Matrix, ImmutableMatrix,
                                 SparseMatrix, ImmutableSparseMatrix)
+
+    if memory:
+        return cse_minimize_memory(*cse(exprs,
+            symbols=symbols, optimizations=optimizations,
+            postprocess=postprocess, order=order, ignore=ignore))
 
     if isinstance(exprs, (int, float)):
         exprs = sympify(exprs)
@@ -770,3 +848,67 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
         return replacements, reduced_exprs
 
     return postprocess(replacements, reduced_exprs)
+
+
+def _cse_homogeneous(exprs, **kwargs):
+    """
+    Same as ``cse`` but the reduced_exprs are expres in the same type as 'exprs'
+
+    Parameters
+    ==========
+
+    exprs : list of sympy expressions, or a single SymPy expression
+        The expressions to reduce.
+    kwargs : additional arguments for the ``cse`` function
+
+    Returns
+    =======
+
+    replacements : list of (Symbol, expression) pairs
+        All of the common subexpressions that were replaced. Subexpressions
+        earlier in this list might show up in subexpressions later in this
+        list.
+    reduced_exprs : list of sympy expressions
+        The reduced expressions with all of the replacements above.
+
+    Examples
+    ========
+
+    >>> from sympy.simplify.cse_main import _cse_homogeneous
+    >>> from sympy import Symbol, cos, Tuple
+    >>> x = Symbol("x")
+    >>> _cse_homogeneous("cos(x + 1)") # str
+    ([], 'cos(x + 1)')
+    >>> _cse_homogeneous(cos(x + 1)) # Basic
+    ([], cos(x + 1))
+    >>> _cse_homogeneous(1) # number
+    ([], 1)
+    >>> _cse_homogeneous(Tuple(1, x))
+    ([], (1, x))
+    >>> _cse_homogeneous([1, x])
+    ([], [1, x])
+    >>> _cse_homogeneous((1, x))
+    ([], (1, x))
+    >>> _cse_homogeneous({1, x})
+    ([], {1, x})
+    """
+    from sympy import sympify
+
+    if isinstance(exprs, str):
+        replacements, reduced_exprs = _cse_homogeneous(sympify(exprs), **kwargs)
+        return replacements, repr(reduced_exprs)
+    if isinstance(exprs, (list, tuple, set)):
+        replacements, reduced_exprs = cse(exprs, **kwargs)
+        return replacements, type(exprs)(reduced_exprs)
+    if isinstance(exprs, dict):
+        keys = list(exprs.keys()) # In order to guarantee the order of the elements.
+        replacements, values = cse([exprs[k] for k in keys], **kwargs)
+        reduced_exprs = {k: v for k, v in zip(keys, values)}
+        return replacements, reduced_exprs
+
+    try:
+        replacements, (reduced_exprs,) = cse(exprs, **kwargs)
+    except TypeError: # For example 'mpf' objects
+        return [], exprs
+    else:
+        return replacements, reduced_exprs
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -87,7 +87,7 @@ def cse_separate(r, e):
     return [reps_toposort(r), e]
 
 
-def cse_minimize_memory(r, e):
+def cse_release_variables(r, e):
     """
     Return tuples giving ``(a, b)`` where ``a`` is a symbol and ``b`` is
     either an expression or None. The value of None is used when a
@@ -100,10 +100,10 @@ def cse_minimize_memory(r, e):
     ========
 
     >>> from sympy import cse
-    >>> from sympy.simplify.cse_main import cse_minimize_memory
+    >>> from sympy.simplify.cse_main import cse_release_variables
     >>> from sympy.abc import x, y
     >>> eqs = [(x + y - 1)**2, x, x + y, (x + y)/(2*x + 1) + (x + y - 1)**2, (2*x + 1)**(x + y)]
-    >>> defs, rvs = cse_minimize_memory(*cse(eqs))
+    >>> defs, rvs = cse_release_variables(*cse(eqs))
     >>> for i in defs:
     ...   print(i)
     ...
--- a/sympy/simplify/tests/test_cse.py
+++ b/sympy/simplify/tests/test_cse.py
@@ -561,8 +561,8 @@ def test_cse_minimize_memory():
     r.reverse()
     assert eqs == [i.subs(r) for i in e]
 
-def test_cse_as_list():
-    _cse = lambda x: cse(x, as_list=False)
+def test_cse_list():
+    _cse = lambda x: cse(x, list=False)
     assert _cse(x) == ([], x)
     assert _cse('x') == ([], 'x')
     it = [x]
--- a/sympy/simplify/tests/test_cse.py
+++ b/sympy/simplify/tests/test_cse.py
@@ -546,13 +546,13 @@ def test_unevaluated_mul():
     eq = Mul(x + y, x + y, evaluate=False)
     assert cse(eq) == ([(x0, x + y)], [x0**2])
 
-def test_cse_minimize_memory():
-    from sympy.simplify.cse_main import cse_minimize_memory
+def test_cse_release_variables():
+    from sympy.simplify.cse_main import cse_release_variables
     _0, _1, _2, _3, _4 = symbols('_:5')
     eqs = [(x + y - 1)**2, x,
         x + y, (x + y)/(2*x + 1) + (x + y - 1)**2,
         (2*x + 1)**(x + y)]
-    r, e = cse(eqs, postprocess=cse_minimize_memory)
+    r, e = cse(eqs, postprocess=cse_release_variables)
     # this can change in keeping with the intention of the function
     assert r, e == ([
     (x0, x + y), (x1, (x0 - 1)**2), (x2, 2*x + 1),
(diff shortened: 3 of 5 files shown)


oscarbenjamin:
The idea seems reasonable.

Since there is a file added in one commit and then removed in another I think that the commits should be squashed.

The code currently has many comments in French. In the SymPy codebase all comments should be in English.

robinechuca:
Done.

.gitignore
```diff
@@ -83,3 +83,6 @@ sample.tex
 
 # pytest related data file for slow tests
 .ci/durations.log
+
+test_lambdify.py
```
oscarbenjamin, line 87: Why has this been added in the gitignore?

sympy/utilities/lambdify.py
```diff
@@ -903,6 +905,97 @@ def _module_present(modname, modlist):
             return True
     return False
 
+class _RedundancySliter:
```
oscarbenjamin, line 908: Why is this a class rather than just some functions?

.gitignore
robinechuca: It is a mistake. I just wanted to do it locally.

sympy/utilities/lambdify.py
robinechuca: I am not sure there are good reasons for this. I needed to do recursive and have 'global' variables. Making a class makes it possible to implement this in a clear way. I am not able to implement this "clean" with a single function. But it is undoubtedly possible. How is it embarrassing to have a class?

oscarbenjamin: I'm not sure what you mean by "global" variables.

I don't find the class to be very clear compared to just some functions e.g.:
```python
def count(expr):
    return 1 + sum(count(child) for child in expr.args)
```
There is already a function for this by the way:
```python
In [1]: e = 1 + x

In [2]: e
Out[2]: x + 1

In [3]: e.count
Out[3]: <bound method Basic.count of x + 1>

In [4]: e.count(Expr)
Out[4]: 3
```

robinechuca: Ha okay, I understood to put everything in a single function! If you prefer to do multiple functions rather than one object, I have nothing against it. Will you be responsible for making the modification or will I take care of it?

oscarbenjamin: It's for you to make the changes. You can just commit and push to the same branch to update the PR.

There are also test failures that need to be addressed.

smichr:
How is this different than the `cse` function?

robinechuca:
> How is this different than the `cse` function?
Indeed, I did not know the existence of this function. you are right, since it exists it is better to use it and not to recode it. However, there is still a small difference, which is that "cse" does not return the now unnecessary variables that can be deleted.
In the case where we work with large arrays, it changes a lot of things.

smichr:
> does not return the now unnecessary variables that can be deleted

please show what you mean (with a small example)

robinechuca:
`
>>> from sympy import *
>>> from sympy.abc import *
>>> x0 = x + cos(x)
>>> x1 = x0 * sin(x0)
>>> x2 = x1 + sqrt(x1)
>>> func = lambdify([x], x2)
>>> print(func.__doc__)`
def _lambdifygenerated(x):
    subvar_1 = x + cos(x)
    subvar_2 = subvar_1*sin(subvar_1)
    del subvar_1
    return sqrt(subvar_2) + subvar_2

smichr (commented):
Does having variables be deleted make the lambdified function perform better?

Does it matter when the deletions happen? Like should they happen as soon as possible or can they occur at the end?

robinechuca:
> Does having variables be deleted make the lambdified function perform better?
If ever the RAM were to saturate and therefore to swap, then we gain a lot to do a little housework in the memory. In an example where I project points from one geometric space to another, with equations derived from the geometric elements of sympy, I gained about 8 GB of RAM which accelerated the execution time of over 1000 by more than 1000. the function ... So yes, in certain situations there is a really noticeable gain. On the other hand, if you don't swap, this modification doesn't improve anything at all but it doesn't hurt either.

robinechuca:
> Does it matter when the deletions happen? Like should they happen as soon as possible or can they occur at the end?
There for once, there is not really the choice. If nothing is deleted at all, the garbage collector will take care of it by itself at the end. It is therefore pointless to tell him which local variable to delete: he already knows.
On the other hand, if we remove the variable as soon as possible (as it is implemented here), it allows to empty the RAM as we go, without accumulating too much stuff.

sympy/utilities/lambdify.py
```diff
@@ -903,6 +905,59 @@ def _module_present(modname, modlist):
...
+    ([(x0, x + cos(x)), (x1, x0*sin(x0))], [set(), {x0}], sqrt(x1) + x1)
+    """
+    def get_free_symbols(expr):
+        if hasattr(expr, "__iter__"):
+            return set.union(*map(get_free_symbols, expr))
+        return expr.free_symbols
+
+    from sympy import cse
+    replacements, reduced_exprs = cse([expr])
+    reduced_exprs = reduced_exprs.pop() # No raison to cast in a list!
```
smichr, line 943: you recognize the need for allowing `expr` to be an iterable in `get_free_symbols`...you should not pop only the first -- and not necessarily the only -- expression.

robinechuca: as 'cse' seemed to put the expression in a list, even though the starting expression was not an iterable, I thought that putting the item in a list from the start would just ensure consistency . But indeed, as you just pointed out to me, this cast is not systematic, so if we put the element in a list, it does not return a list of list but a simple list. It's curious but it's like that! So I corrected the problem.

sympy/utilities/lambdify.py
```diff
@@ -903,6 +905,59 @@ def _module_present(modname, modlist):
...
+    for free_symbol in free_symbols:
+        symbols_cum.insert(0, free_symbol | symbols_cum[0])
+
+    # Detection of useless symbols.
+    useless_symbols = []
+    current_symbols = set() # Buffers memory.
+    for (new_symb, _), following_symb in zip(replacements, symbols_cum[1:]):
+        current_symbols.add(new_symb)
+        useless_symbols.append(current_symbols - following_symb) # The set of useless symbols.
+        current_symbols -= useless_symbols[-1]
```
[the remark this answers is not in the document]
robinechuca: Sorry but I don't understand your proposition...

sympy/utilities/lambdify.py
```diff
@@ -934,16 +934,18 @@ def _cse_del(expr):
     ([(x0, x + cos(x)), (x1, x0*sin(x0))], [set(), {x0}], sqrt(x1) + x1)
     """
     def get_free_symbols(expr):
-        if hasattr(expr, "__iter__") and not isinstance(expr, str):
-            return set.union(*map(get_free_symbols, expr))
         try:
             return expr.free_symbols
         except AttributeError:
             return set()
+        if hasattr(expr, "__iter__") and not isinstance(expr, str):
```
smichr, line 941: if you pass an iterable it will return set() because of the try/except block above

smichr: Also, we typically use `iterable` to do this test:
```python
>>> from sympy.core.compatibility import iterable
>>> iterable('asd')
False
>>> iterable([])
True
```

robinechuca: > if you pass an iterable it will return set() because of the try/except block above

My God! The big beginner's mistake! We will say that I was really not in good shape that day ... Thank you for showing it to me;)

robinechuca: > Also, we typically use `iterable` to do this test:
> 
> ```python
> >>> from sympy.core.compatibility import iterable
> >>> iterable('asd')
> False
> >>> iterable([])
> True
> ```

just like 'cse', i didn't know about the existence of this function. I just integrated it, it's true that it's a little more elegant.

asmeurer:
If I understand the code here, this is not just about adding redundancy removing, but actually about adding automatic calling of `cse` in lambdify. I think this should be done with more thought and care. There have been some discussions about this in the past. See https://github.com/sympy/sympy/issues/18916. The option should probably be added as optional and off by default. 

I can see the use of the redundancies thing, although I would prefer to see just plain CSE fleshed out nicely before we worry about that. I can't help but wonder if the same thing could be achieved at a higher level than labmdify. It's really just a general problem of automatically deleting local variables from a function once they can no longer be used. I wonder if there are more general ways of doing this. For instance, I wouldn't be surprised if there were a library that provides a decorator for any function that rewrites it like this. 

Also CC @moorepants who is interested in this feature.

smichr:
What about something like this (which would be added to the cse post processor function "library" at the top of cse_main.py?
```python
def mark_last_usage(r, e):
    s, p = zip(*r)
    s = set(s)
    p = list(p)
    # sort e so those with most sub-expressions appear first
    e = [(e[i],i) for i in range(len(e))]
    e, perm = zip(*sorted(e,
        key=lambda x: -len(x[0].free_symbols & s)))
    p += e
    in_use = s
    useless = defaultdict(list)
    i = len(p) - 1
    while i > 0 and in_use:
        s = in_use & p.pop().free_symbols
        useless[i] = sorted(s, key=str)
        in_use -= s
        i -= 1
    return [r, list(e), perm, useless]
```
with simulated cse input:
```python
x0,x1,x2,x3 = symbols('x:4')
from sympy.abc import x, y
r = [(x0, x + y), (x1, 2*x + 1), (x2, x0 - 1), (x3, x2**2)]
e = [x3, x, x0, x0/x1 + x3, x1**x0]
r, e, perm, u = mark_last_usage(r,e)
j = 0
for i, eq in enumerate(r + e):
    if type(eq) is tuple:
        print('%s = %s' % eq)
    else:
        print('_%s = %s'%(j, eq))
        j += 1
    if u[i]:
        print('del ' + ', '.join([str(i) for i in u[i]]))
print('return ' + ', '.join('_%s'%perm.index(i) for i in range(j)))
```
output
```
x0 = x + y
x1 = 2*x + 1
x2 = x0 - 1
x3 = x2**2
del x2
_0 = x0/x1 + x3
_1 = x1**x0
del x1
_2 = x3
del x3
_3 = x0
del x0
_4 = x
return _2, _4, _3, _0, _1
```

asmeurer:
Another option, which would be a lot more work, but cleaner in the end, would be to finish the work that I started for making sympy.codegen.CodeBlock.cse() work, and adding an option to it to add deletion nodes. We could then make lambdify print CodeBlocks. The advantage of this is that this could generalise to other code printers for other languages.

robinechuca:
I'm not at all familiar with 'sympy'.
This is the very first time that I have contributed to a project.
Everything you offer sounds very interesting but I don't think I can implement it. While waiting better, I propose a simple modification, with a parameter in 'lambdify' which allows to activate it or not.

smichr:
> with a parameter in 'lambdify' which allows to activate it or not.

the problem with this is that, like `simplify` it is not a descriptive command. The user does not know how the optimization will be done. And it doesn't lend itself well to future improvements.

I still like the idea of a post-processor, something like
```python
def code_form_minimize_memory(r, e):
    """return tuples giving (a, b) where ``a`` is either
        * a cse symbol, 
        * a symbol assigned to each of the original expressions, or 
        * None
    The list of symbols assigned to original expressions is returned as the last tuple.

    When the first element of a tuple is None, it indicates that the variables
    given as a list as the second element of the tuple are not needed by any sub-expression.

    Examples
    ========

    >>> from sympy import cse
    >>> from sympy.simplify.cse_main import code_form_minimize_memory
    >>> eqs = [(x + y - 1)**2, x, x + y, (x + y)/(2*x + 1) + (x + y - 1)**2, (2*x + 1)**(x + y)]
    >>> defs, rvs = code_form_minimize_memory(*cse(eqs))
    >>> for i in defs:
    ...   print(i)
    ...
    (x0, x + y)
    (x1, (x0 - 1)**2)
    (x2, 2*x + 1)
    (_3, x0/x2 + x1)
    (_4, x2**x0)
    (None, [x2])
    (_0, x1)
    (None, [x1])
    (_2, x0)
    (None, [x0])
    (_1, x)
    >>> print(rvs)
    (_0, _1, _2, _3, _4)
    """
    s, p = zip(*r)
    esyms = symbols('_:%s'%len(e))
    syms = list(esyms)
    s = list(s)
    ss = set(s)
    p = list(p)
    # sort e so those with most sub-expressions appear first
    e = [(e[i], syms[i]) for i in range(len(e))]
    e, syms = zip(*sorted(e,
        key=lambda x: -sum([p[s.index(i)].count_ops() for i in x[0].free_symbols & ss])))
    syms = list(syms)
    p += e
    in_use = ss
    del ss
    rv = []
    i = len(p) - 1
    useless = {}
    while i >= 0:
        _p = p.pop()
        c = in_use & _p.free_symbols
        if c:
            rv.append((None, sorted(c, key=str)))  # sorting for canonical results
        if i >= len(r):
            rv.append((syms.pop(), _p))
        else:
            rv.append((s[i], _p))
        in_use -= c
        i -= 1
    rv.reverse()
    return rv, esyms
```

sympy/utilities/lambdify.py
```diff
@@ -905,6 +919,62 @@ def _module_present(modname, modlist):
...
+    >>> _cse_homogeneous((1, x))
+    ([], (1, x))
+    >>> _cse_homogeneous({1, x})
+    ([], {1, x})
+    """
+    from sympy import cse, sympify, Matrix
+    from sympy.core.containers import Tuple
+
+    if isinstance(exprs, str):
+        replacements, (reduced_exprs,) = cse(sympify(exprs), **kwargs)
```
asmeurer: What is the purpose of this helper function? The string parsing here can lead to issues. I would prefer to avoid it entirely. The remaining cases should be handled by cse() itself, if they aren't already.

asmeurer:
> Everything you offer sounds very interesting but I don't think I can implement it. While waiting better, I propose a simple modification, with a parameter in 'lambdify' which allows to activate it or not.

Yes, that is really all I want to see. Personally I'm a little worried about the added complexity deletion stuff, but if others don't see a problem with it we can leave it for now.

We should also decide if `optimize` is the best name for the flag. I would have called it just `cse`, since `optimize` implies that it might do other stuff. OTOH maybe it would make sense for lambdify to do other kinds of optimizations in the future. @moorepants and others, any thoughts?

sympy/utilities/lambdify.py
```diff
@@ -903,6 +919,123 @@ def _module_present(modname, modlist):
...
+        return replacements, reduced_exprs
+
+    try:
+        replacements, (reduced_exprs,) = cse(exprs, **kwargs)
+    except TypeError: # For example 'mpf' objects
+        return [], exprs
+    else:
+        return replacements, reduced_exprs
+
+def _cse_del(expr):
```
asmeurer, line 985: I would

- Move this helper to cse_main.py
- Make the input `replacements, reduced_expr` (i.e., the output of cse()). That way the caller can call cse() with whatever arguments they want, then process the result with this function. 
- Unless `replacements` or `reduced_exprs` are changed (they don't seem to be), this should just return `useless_symbols`.
- "useless" is a bit misleading. I would rather call this "deletable symbols".

sympy/utilities/lambdify.py
```diff
@@ -1059,8 +1192,10 @@ def __init__(self, printer=None, dummify=False):
         # Used to print the generated function arguments in a standard way
         self._argrepr = LambdaPrinter().doprint
 
-    def doprint(self, funcname, args, expr):
-        """Returns the function definition code as a string."""
+    def doprint(self, funcname, args, expr, *, useless_symbols=[], replacements=[]):
```
asmeurer, line 1195: Use `()` instead of `[]` for default arguments. 

```suggestion
    def doprint(self, funcname, args, expr, *, useless_symbols=(), replacements=()):
```

asmeurer: Let's also make these start with an underscore. This API is fine for now, but eventually we should replace it with codegen AST as I described in my other comment.

sympy/utilities/lambdify.py
```diff
@@ -903,6 +919,123 @@ def _module_present(modname, modlist):
...
+        replacements, reduced_exprs = cse(exprs, **kwargs)
+        return replacements, type(exprs)(reduced_exprs)
+    if isinstance(exprs, dict):
+        keys = list(exprs.keys()) # In order to guarantee the order of the elements.
+        replacements, values = cse([exprs[k] for k in keys], **kwargs)
+        reduced_exprs = {k: v for k, v in zip(keys, values)}
+        return replacements, reduced_exprs
+
+    try:
+        replacements, (reduced_exprs,) = cse(exprs, **kwargs)
```
asmeurer, line 979: Are there tests for lambdify taking non-SymPy arguments like str or mpf? I think it should use `_sympify` on `expr` so that things like mpf are converted to SymPy objects first (and it shouldn't accept str at all).

moorepants:
I haven't had time to look at this, but my only comment would be that if this is added it is implemented in a way where we won't have to deprecate the api and can change the internals to a different implementation. A `cse=True/False` flag for lambdify seems like a good api change that will work for future updates too.

sympy/utilities/lambdify.py
robinechuca: The next test failed:
```python
>>> lambdify("x, y, z", "z, x, y")(1, 2, 3) == (3, 1, 2)
```
It is maybe not the best mean to solve it, Do you prefer I make is fixing into the 'cse' function?

sympy/utilities/lambdify.py
smichr: > * Move this helper to cse_main.py

Yes, as I said [here](https://github.com/sympy/sympy/pull/21546#issuecomment-856190288)

> * Make the input `replacements, reduced_expr` (i.e., the output of cse()). That way the caller can call cse() with whatever arguments they want, then process the result with this function.

Yes (as also shown there)

> * Unless `replacements` or `reduced_exprs` are changed (they don't seem to be), this should just return `useless_symbols`.

I disagree -- having a list of symbols that can be deleted at some point in the step-wise computation is not useful. One of the proposals here returns a set of what can be deleted at each step of the way; the version I suggested just sorts in the variables to be deleted along with the replacements and expressions. In order to do something with the symbols you need to know when they can be (manually) garbage collected.

> * "useless" is a bit misleading. I would rather call this "deletable symbols".

agree

> We should also decide if optimize is the best name for the flag. I would have called it just cse

I agree that it should not be `optimize`; `cse` is fairly neutral and as long as it works, I don't think anyone will care if you help it do garbage collection in the process by deleting variables once they are done being used.

* `cse` can handle more than one expression; should code-printing be able to return more than one expression, i.e. should lambdify be able to handle a vector of expression? If so, the code I am advocating for handles that case and allows re-ordering of expressions during calculation to try optimize when they are calculated so memory can be released. I am not guaranteeing that there is not a better way to do it, just that it can handle re-ordering.

bjodah:
Having `cse` available in `lambdify` would be great, I had a need for this functionality today, and I then found this PR. 👍

I opened [a draft PR](https://github.com/sympy/sympy/pull/21601) to at least allow using our exisiting `cse` with `simplify` (I have yet to test that matrices etc. are supported in that PR).

@robinechuca are you sure you need `del`? Or can we just reuse the variable name for the subsequent expressions? What do you think about my suggesting allowing `cse` to be a function? (if one reuses the variable names I think it should work, no?)

robinechuca:
> Having `cse` available in `lambdify` would be great, I had a need for this functionality today, and I then found this PR. +1
> 
> I opened [a draft PR](https://github.com/sympy/sympy/pull/21601) to at least allow using our exisiting `cse` with `simplify` (I have yet to test that matrices etc. are supported in that PR).
> 
> @robinechuca are you sure you need `del`? Or can we just reuse the variable name for the subsequent expressions? What do you think about my suggestion to allow `cse` to be a function? (if one reuses the variable names I think it should work, no?)

The advantage of creating a new symbol for each pattern and deleting the symbols already used is that it is very understandable and there are no surprises.
What you propose seems to me to be more optimized. Let's imagine for example that we run lambify with numpy arrays. If we change the value of a variable, we avoid a call of the garbage collector, and we avoid a c-alloc.
On the other hand, I am afraid that there are bad surprises with pointers. Intuitively there is a risk that changing the state of a variable will also change the state of another variable against our will. I'm not sure, it's just a hunch, but I'm afraid it could lead to some really dark bugs.

robinechuca:
thanks @bjodah . I'm not sure I did the right things but in the idea it must not be far ...

bjodah:
@robinechuca great. Unless I've missed something, I guess the consensus now is that you "just" need to:

1. add some more tests (as suggested by @moorepants [here](https://github.com/sympy/sympy/pull/21601#issuecomment-860009637))
2. move your "deleting cse" to cse_main.py and name it something appropriate.
3. make sure that it can be used with `lambdify` generating code along the lines of what @smichr suggested here:
https://github.com/sympy/sympy/pull/21601#issuecomment-859788950

Point 3 might require an extra function.

If overwriting expressions with `1` does not cut it, and we really need `del` (I'm not convinced), then I would consider looking into employing the `sympy.codegen` submodule (but that would probably be a rabbit hole so I would at least postpone such an attempt to a later subsequent PR).

sympy/utilities/lambdify.py
```diff
@@ -336,6 +336,23 @@ def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
         (if ``args`` is not a string) - for example, to ensure that the
         arguments do not redefine any built-in names.
 
+    cse : bool or callable, optional
+        When the expression that we want to vectorialize is a large expression
```
smichr, line 340:
```suggestion
        When vectorizing a large expression
```

sympy/utilities/lambdify.py
```diff
@@ -336,6 +336,23 @@ def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
...
+        reduce the size in RAM by setting this parameter to True.
+
+        However, the step of finding redundant paterns can be slow.
+        So, if the creation of the 'lambdify' function should be fast,
+        set this parameter to False. If, on the other hand, you want
+        the evaluation of the function returned by 'lambdify' to be
+        as efficient as possible, set this parameter to True.
+
+        Whether or not common subexpression elimination should be performed on
+        the expressions. This can potentially improve performance of the generated
```
smichr, line 351:
```suggestion
        the expressions, this can potentially improve performance of the generated
```

sympy/utilities/lambdify.py
```diff
@@ -905,6 +929,123 @@ def _module_present(modname, modlist):
             return True
     return False
 
+def _cse_homogeneous(exprs, **kwargs):
+    """
+    Same as ``cse`` but the reduced_exprs are expres in the same type as 'exprs'
```
smichr, line 934:
```suggestion
    Same as ``cse`` but the ``reduced_exprs`` will have the same type as ``exprs``
```

sympy/utilities/lambdify.py
```diff
@@ -905,6 +929,123 @@ def _module_present(modname, modlist):
...
+
+    Parameters
+    ==========
+
+    exprs : list of sympy expressions, or a single sympy expression
```
smichr, line 939:
```suggestion
    exprs : list of sympy expressions, or a single SymPy expression
```

sympy/utilities/lambdify.py
```diff
@@ -905,6 +929,123 @@ def _module_present(modname, modlist):
...
+        The expressions to reduce.
+    kwargs : The over argument of the ``cse`` function
```
smichr, line 941:
```suggestion
    kwargs : additional arguments for the ``cse`` function
```

sympy/utilities/lambdify.py
```diff
@@ -905,6 +929,123 @@ def _module_present(modname, modlist):
...
+    ========
+
+    >>> from sympy.utilities.lambdify import _cse_del
+    >>> from sympy import sqrt, cos, sin, Symbol
+    >>> x = Symbol("x")
+    >>> expr = sqrt((x + cos(x))*sin(x + cos(x))) + (x + cos(x))*sin(x + cos(x))
+    >>> _cse_del(expr)
+    ([(x0, x + cos(x)), (x1, x0*sin(x0))], [set(), {x0}], sqrt(x1) + x1)
+    """
+    def get_free_symbols(expr):
```
smichr, line 1023: What will this do when expr is a dictionary?

smichr: And in what situation will you want and use the cse of a dictionary?

smichr:
What do you think about handling the memory minimization call like this (along with other changes):
```diff
diff --git a/sympy/simplify/cse_main.py b/sympy/simplify/cse_main.py
index 5cea125e93..cd3534d923 100644
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -88,14 +88,12 @@ def cse_separate(r, e):
 
 
 def cse_minimize_memory(r, e):
-    """return tuples giving (a, b) where ``a`` is either
-        * a cse symbol,
-        * a symbol assigned to each of the original expressions, or
-        * None
-    The list of symbols assigned to original expressions is returned as the last tuple.
+    """return tuples giving (a, b) where ``a`` is a symbol and b is
+    either an expression or None. The value of None is used when a
+    symbol is no longer needed for subsequent expressions.
 
-    When the first element of a tuple is None, it indicates that the variables
-    given as a list as the second element of the tuple are not needed by any sub-expression.
+    Use of such output can reduce the memory footprint of lambdified
+    expressions that contain large, repeated subexpressions.
 
     Examples
     ========
@@ -128,23 +126,25 @@ def cse_minimize_memory(r, e):
     esyms = symbols('_:%d' % len(e))
     syms = list(esyms)
     s = list(s)
-    ss = set(s)
+    in_use = set(s)
     p = list(p)
     # sort e so those with most sub-expressions appear first
+    # in attempt to minimize memory usage -- this can be
+    # modified if there is a better algorithm for this
     e = [(e[i], syms[i]) for i in range(len(e))]
     e, syms = zip(*sorted(e,
-        key=lambda x: -sum([p[s.index(i)].count_ops() for i in x[0].free_symbols & ss])))
+        key = lambda x: -sum([p[s.index(i)].count_ops()
+        for i in x[0].free_symbols & in_use])))
     syms = list(syms)
     p += e
-    in_use = ss
-    # del ss # makes failing the quality code test.
     rv = []
     i = len(p) - 1
     while i >= 0:
         _p = p.pop()
         c = in_use & _p.free_symbols
         if c:
-            rv.extend([(s, None) for s in sorted(c, key=str)]) # sorting for canonical results
+            # sorting for canonical results
+            rv.extend([(s, None) for s in sorted(c, key=str)])
         if i >= len(r):
             rv.append((syms.pop(), _p))
         else:
@@ -184,7 +184,7 @@ def preprocess_for_cse(expr, optimizations):
 
 def postprocess_for_cse(expr, optimizations):
     """ Postprocess an expression after common subexpression elimination to
-    return the expression to canonical sympy form.
+    return the expression in canonical SymPy form.
 
     Parameters
     ==========
@@ -695,7 +695,7 @@ def _rebuild(expr):
 
 
 def cse(exprs, symbols=None, optimizations=None, postprocess=None,
-        order='canonical', ignore=(), memory=False):
+        order='canonical', ignore=(), asList=True):
     """ Perform common subexpression elimination on an expression.
 
     Parameters
@@ -726,9 +726,8 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
         concern, use the setting order='none'.
     ignore : iterable of Symbols
         Substitutions containing any Symbol from ``ignore`` will be ignored.
-    memory : boolean
-        Allows to activate the search for removable symbols.
-        Seealso ``cse_minimize_memory``.
+    asList : bool (default True)
+        return expression in list or else with same type as input (when False)
 
     Returns
     =======
@@ -775,8 +774,8 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     from sympy.matrices import (MatrixBase, Matrix, ImmutableMatrix,
                                 SparseMatrix, ImmutableSparseMatrix)
 
-    if memory:
-        return cse_minimize_memory(*cse(exprs,
+    if not asList:
+        return _cse_homogeneous(exprs, **dict(
             symbols=symbols, optimizations=optimizations,
             postprocess=postprocess, order=order, ignore=ignore))
 
@@ -851,7 +850,8 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
 
 def _cse_homogeneous(exprs, **kwargs):
     """
-    Same as ``cse`` but the reduced_exprs are expres in the same type as 'exprs'
+    Same as ``cse`` but the ``reduced_exprs`` are returned
+    with the same type as ``exprs`` or a sympified version of the same.
 
     Parameters
     ==========
@@ -874,27 +874,32 @@ def _cse_homogeneous(exprs, **kwargs):
     ========
 
     >>> from sympy.simplify.cse_main import _cse_homogeneous
-    >>> from sympy import Symbol, cos, Tuple
-    >>> x = Symbol("x")
-    >>> _cse_homogeneous("cos(x + 1)") # str
-    ([], 'cos(x + 1)')
-    >>> _cse_homogeneous(cos(x + 1)) # Basic
-    ([], cos(x + 1))
-    >>> _cse_homogeneous(1) # number
-    ([], 1)
-    >>> _cse_homogeneous(Tuple(1, x))
-    ([], (1, x))
-    >>> _cse_homogeneous([1, x])
-    ([], [1, x])
-    >>> _cse_homogeneous((1, x))
-    ([], (1, x))
-    >>> _cse_homogeneous({1, x})
-    ([], {1, x})
+    >>> from sympy import cos, Tuple
+    >>> from sympy.abc import x
+    >>> _cse_homogeneous('[cos(x), cos(x) + 1]')
+    ([(x0, cos(x))], '[x0, x0 + 1]')
+
+    >>> output = lambda x: type(_cse_homogeneous(x)[1])
+    >>> output(1)
+    <class 'sympy.core.numbers.One'>
+    >>> output('cos(x)')
+    <class 'str'>
+    >>> output(cos(x))
+    cos
+    >>> output(Tuple(1, x))
+    <class 'sympy.core.containers.Tuple'>
+    >>> output([1, x])
+    <class 'list'>
+    >>> output((1, x))
+    <class 'tuple'>
+    >>> output({1, x})
+    <class 'set'>
     """
     from sympy import sympify
 
     if isinstance(exprs, str):
-        replacements, reduced_exprs = _cse_homogeneous(sympify(exprs), **kwargs)
+        replacements, reduced_exprs = _cse_homogeneous(
+        sympify(exprs), **kwargs)
         return replacements, repr(reduced_exprs)
     if isinstance(exprs, (list, tuple, set)):
         replacements, reduced_exprs = cse(exprs, **kwargs)
diff --git a/sympy/simplify/tests/test_cse.py b/sympy/simplify/tests/test_cse.py
index 60affd0915..3f822d2c07 100644
--- a/sympy/simplify/tests/test_cse.py
+++ b/sympy/simplify/tests/test_cse.py
@@ -8,6 +8,7 @@
     IndexedBase, Idx, Piecewise, O, signsimp
 )
 from sympy.core.function import count_ops
+from sympy.simplify.cse_main import cse_minimize_memory
 from sympy.simplify.cse_opts import sub_pre, sub_post
 from sympy.functions.special.hyper import meijerg
 from sympy.simplify import cse_main, cse_opts
@@ -20,6 +21,7 @@
 
 w, x, y, z = symbols('w,x,y,z')
 x0, x1, x2, x3, x4, x5, x6, x7, x8, x9, x10, x11, x12 = symbols('x:13')
+_0, _1, _2, _3, _4 = symbols('_:5')
 
 
 def test_numbered_symbols():
@@ -555,3 +557,28 @@ def test_issue_18991():
 def test_unevaluated_Mul():
     m = [Mul(1, 2, evaluate=False)]
     assert cse(m) == ([], m)
+
+
+def test_cse_minimize_memory():
+    eqs = [(x + y - 1)**2, x,
+    x + y, (x + y)/(2*x + 1) + (x + y - 1)**2,
+    (2*x + 1)**(x + y)]
+    r, e = cse(eqs, postprocess=cse_minimize_memory) 
+    # this can change in keeping with the intention of the function
+    assert r, e == ([
+    (x0, x + y), (x1, (x0 - 1)**2), (x2, 2*x + 1),
+    (_3, x0/x2 + x1), (_4, x2**x0), (x2, None), (_0, x1),
+    (x1, None), (_2, x0), (x0, None), (_1, x)], (_0, _1, _2, _3, _4))
+    r.reverse()
+    assert eqs == [i.subs(r) for i in e]
+
+
+def test_cse_asList():
+    _cse = lambda x: cse(x, asList=False)
+    assert _cse(x) == ([], x)
+    assert _cse('x') == ([], 'x')
+    it = [x]
+    for c in (list, tuple, set, Tuple):
+        assert _cse(c(it)) == ([], c(it))
+    d = {x: 1}
+    assert _cse(d) == ([], d)
diff --git a/sympy/utilities/lambdify.py b/sympy/utilities/lambdify.py
index c78d5d3c04..c98264227a 100644
--- a/sympy/utilities/lambdify.py
+++ b/sympy/utilities/lambdify.py
@@ -336,21 +336,17 @@ def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
         (if ``args`` is not a string) - for example, to ensure that the
         arguments do not redefine any built-in names.
 
-    cse : bool or callable, optional
-        When vectorializing a large expression as efficient as possible,
-        set this parameter to True.
-        Whether or not common subexpression elimination should be performed on
-        the expressions, This can potentially improve performance of the generated
-
-        However, the step of finding redundant paterns can be slow.
-        So, if the creation of the 'lambdify' function should be fast,
-        set this parameter to False.
-
-        Whether or not common subexpression elimination should be performed on
-        the expressions. This can potentially improve performance of the generated
-        function. When ``True``, then ``sympy.simplify.cse`` is used, the user
-        may pass their own function matching the signature of said function.
+    cse : bool, None or callable, optional
+        Large expressions can be computed more efficiently when
+        common subexpressions are identified and precomputed before
+        being used multiple time. Finding the subexpressions will make
+        creation of the 'lambdify' function slower, however.
 
+        When ``True``, ``sympy.simplify.cse`` with default settings is
+        used, but any function matching the output of ``cse`` can be
+        used. When ``None``, intermediate variables which are no longer
+        needed will be assigned a value which can help to minimize
+        memory usage.
 
 
     Examples
@@ -864,9 +860,10 @@ def _lambdifygenerated(x):
     else:
         funcprinter = _EvaluatorPrinter(printer, dummify)
 
-    if cse:
-        from sympy.simplify.cse_main import _cse_homogeneous
-        cses, expr = _cse_homogeneous(expr, memory=True)
+    if cse != False:
+        from sympy.simplify.cse_main import _cse_homogeneous, cse_minimize_memory
+        cses, expr = _cse_homogeneous(expr,
+            postprocess=None if cse else cse_minimize_memory)
     else:
         cses = ()
     funcstr = funcprinter.doprint(funcname, args, expr, cses=cses)
diff --git a/sympy/utilities/tests/test_lambdify.py b/sympy/utilities/tests/test_lambdify.py
index 1960cbf17c..d073c2d6e1 100644
--- a/sympy/utilities/tests/test_lambdify.py
+++ b/sympy/utilities/tests/test_lambdify.py
@@ -1408,7 +1408,7 @@ def test_cupy_dotproduct():
         cupy.array([14])
 
 
-def test_cse():
+def test_lambdify_cse():
     def dummy_cse(exprs):
         return (), exprs
 
@@ -1438,49 +1438,57 @@ def assertAllClose(self, result, *, abstol=1e-15, reltol=1e-15):
                 else:
                     assert abs_err/abs(r) < reltol
 
-    case1 = Case(
-        args=(x, y, z),
-        exprs=[
+    cases = [
+        Case(
+            args=(x, y, z),
+            exprs=[
             x + y + z,
             x + y - z,
             2*x + 2*y - z,
             (x+y)**2 + (y+z)**2,
-        ],
-        num_args=(2., 3., 4.)
-    )
-    case2 = Case(
-        args=(x, y, z),
-        exprs=[
+            ],
+            num_args=(2., 3., 4.)
+        ),
+        Case(
+            args=(x, y, z),
+            exprs=[
             x + sympy.Heaviside(x),
             y + sympy.Heaviside(x),
             z + sympy.Heaviside(x, 1),
             z/sympy.Heaviside(x, 1)
-        ],
-        num_args=(0., 3., 4.)
-    )
-    case3 = Case(
-        args=(x, y, z),
-        exprs=[
+            ],
+            num_args=(0., 3., 4.)
+        ),
+        Case(
+            args=(x, y, z),
+            exprs=[
             x + sinc(y),
             y + sinc(y),
             z - sinc(y)
-        ],
-        num_args=(0.1, 0.2, 0.3)
-    )
-    case4 = Case(
-        args=(x, y, z),
-        exprs=[
-            Matrix([[x, x*y], [sin(z) + 4, x**z]]),
-            x*y+sin(z)-x**z,
-            Matrix([x*x, sin(z), x**z])
-        ],
-        num_args=(1.,2.,3.),
-        requires_numpy=True
-    )
-    for case in [case1, case2, case3, case4]:
+            ],
+            num_args=(0.1, 0.2, 0.3)
+        ),
+        Case(
+            args=(x, y, z),
+            exprs=[
+                Matrix([[x, x*y], [sin(z) + 4, x**z]]),
+                x*y+sin(z)-x**z,
+                Matrix([x*x, sin(z), x**z])
+            ],
+            num_args=(1.,2.,3.),
+            requires_numpy=True
+        ),
+        Case(
+            args=(x, y),
+            exprs=[(x + y - 1)**2, x, x + y,
+            (x + y)/(2*x + 1) + (x + y - 1)**2, (2*x + 1)**(x + y)],
+            num_args=(1,2)
+        )
+    ]
+    for case in cases:
         if not numpy and case.requires_numpy:
             continue
-        for cse in [False, True, dummy_cse]:
+        for cse in [False, True, None, dummy_cse]:
             f = case.lambdify(cse=cse)
             result = f(*case.num_args)
             case.assertAllClose(result)
```

asmeurer:
Reusing variable names is a neat idea that would allow using the existing cse API with only a minor modification. On the other hand, I do worry that it could cause issues down the line with other code generators where a variable gets reused on an expression with a different type, which languages like C do not allow.

sympy/utilities/lambdify.py
```diff
@@ -336,6 +336,22 @@ def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
...
+        the expressions, This can potentially improve performance of the generated
+
+        However, the step of finding redundant paterns can be slow.
+        So, if the creation of the 'lambdify' function should be fast,
+        set this parameter to False.
+
+        Whether or not common subexpression elimination should be performed on
+        the expressions. This can potentially improve performance of the generated
+        function. When ``True``, then ``sympy.simplify.cse`` is used, the user
+        may pass their own function matching the signature of said function.
```
moorepants, line 352: Instead of using "None" to signify this low memory option, I think it would be better to show an example of using the callable option, in which you use the low memory cse function.

smichr: Do you think it's ok to do both or do you want the `cse` option to be strictly binary? In either case something like the following could be added:
```python
from sympy.simplify.cse_main import cse_minimize_memory
f = lambdify(exprs, cse=lambda x: cse(x, postprocess=cse_minimize_memory)
```
```

**Comment by smichr** [comment]
> I do worry that it could cause issues down the line with other code generators where a variable gets reused on an expression with a different type, which languages like C do not allow.

This is what has me worried about assigning None to the unused variables -- is this going to be ok or will some language complain?

**Comment by bjodah** [comment]
If we want to make this version of cse useful for C/C++/Fortran we need to pass type information: data type, dimension(s). But reusing variables is exactly what we want to do there, as opposed to "`del`", since `malloc`/`free` is the last thing you want to do in e.g. a hot loop.

File: sympy/utilities/lambdify.py (hunk 16 above)
Comment (moorepants) [reply sympy/utilities/lambdify.py:352]: It just seems odd to link "None" to this specific cse implementation. Having cse is a super common case, but customizing cse optimizations isn't so common and there are many options. Your example seems like a universal way to handle it.

**Comment by asmeurer** [comment]
> If we want to make this version of cse useful for C/C++/Fortran we need to pass type information: data type, dimension(s). But reusing variables is exactly what we want to do there, as opposed to "del", since malloc/free is the last thing you want to do in e.g. a hot loop.

> EDIT: Thinking about this a little more: I don't think we need to worry about mixing of variables of different types (int, double, etc.) and dimensions(s) (scalar, vector, matrix). Any user who does code generation right now already handles those separately. So having None appear should be fine (that will be a no-op in the rendering of the code). And a user who wants to keep their stack variables to a minimum can write another pass which re-use local variable names.

So maybe it would in fact be best to implement this via a flag to cse, which I think should be off by default, that allows it to reuse variables once they are no longer needed. Something like cse(reuse_variables=True). If that is problematic for some implementation, they can either turn it off, or manually process the result to change reused variables into some other way of freeing memory.

File: sympy/simplify/cse_main.py [hunk 17]
```diff
@@ -770,3 +843,71 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
...
+def _cse_homogeneous(exprs, **kwargs):
+    """
+    Same as ``cse`` but the ``reduced_exprs`` are returned
+    with the same type as ``exprs`` or a sympified version of the same.
+
+    Parameters
+    ==========
+
+    exprs : list of sympy expressions, or a single SymPy expression
+        The expressions to reduce.
```
smichr, line 857:
```suggestion
    exprs : an Expr, iterable of Expr or dictionary with Expr values
        the expressions in which repeated subexpressions will be identified
```

sympy/utilities/lambdify.py
```diff
@@ -336,6 +336,17 @@ def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
         (if ``args`` is not a string) - for example, to ensure that the
         arguments do not redefine any built-in names.
 
+    cse : bool, None or callable, optional
```
smichr, line 339:
```suggestion
    cse : bool, or callable, optional
```

sympy/utilities/lambdify.py
```diff
@@ -336,6 +336,17 @@ def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
...
+        Large expressions can be computed more efficiently when
+        common subexpressions are identified and precomputed before
+        being used multiple time. Finding the subexpressions will make
+        creation of the 'lambdify' function slower, however.
+
+        Whether or not common subexpression elimination should be performed on
+        the expressions. This can potentially improve performance of the generated
+        function. When ``True``, then ``sympy.simplify.cse`` is used, the user
+        may pass their own function matching the signature of said function.
```
smichr, line 348:
```suggestion
        When ``True``, the default ``sympy.simplify.cse`` is used, otherwise the user
        may pass a function matching the signature ``cse``.
```

sympy/utilities/lambdify.py
```diff
@@ -847,7 +858,14 @@ def _lambdifygenerated(x):
         funcprinter = _TensorflowEvaluatorPrinter(printer, dummify) # type: _EvaluatorPrinter
     else:
         funcprinter = _EvaluatorPrinter(printer, dummify)
-    funcstr = funcprinter.doprint(funcname, args, expr)
+
+    if cse != False:
+        from sympy.simplify.cse_main import _cse_homogeneous, cse_minimize_memory
+        cses, expr = _cse_homogeneous(expr,
+            postprocess=None if cse else cse_minimize_memory)
```
smichr, line 865:
```suggestion
        from sympy.simplify.cse_main import _cse_homogeneous
        cses, expr = _cse_homogeneous(expr)
```

sympy/utilities/tests/test_lambdify.py
```diff
@@ -1406,3 +1406,91 @@ def test_cupy_dotproduct():
         f3(1, 2, 3) == \
         f4(1, 2, 3) == \
         cupy.array([14])
+
+
```
smichr, line 1410:
```suggestion
```

sympy/utilities/tests/test_lambdify.py
```diff
@@ -1406,3 +1406,91 @@ def test_cupy_dotproduct():
...
+
+
+def test_lambdify_cse():
+    def dummy_cse(exprs):
+        return (), exprs
```
smichr, line 1415:
```suggestion
        return (), exprs

    def minmem(exprs):
        from sympy.simplify.cse_main import cse_minimize_memory
        return cse(expr, postprocess=cse_minimize_memory)
```

sympy/utilities/tests/test_lambdify.py
```diff
@@ -1406,3 +1406,91 @@ def test_cupy_dotproduct():
...
+            args=(x, y),
+            exprs=[(x + y - 1)**2, x, x + y,
+            (x + y)/(2*x + 1) + (x + y - 1)**2, (2*x + 1)**(x + y)],
+            num_args=(1,2)
+        )
+    ]
+    for case in cases:
+        if not numpy and case.requires_numpy:
+            continue
+        for cse in [False, True, None, dummy_cse]:
```
smichr, line 1493:
```suggestion
        for cse in [False, True, minmem, dummy_cse]:
```

sympy/simplify/cse_main.py
```diff
@@ -86,6 +86,73 @@ def cse_separate(r, e):
     e = d[False]
     return [reps_toposort(r), e]
 
+
+def cse_minimize_memory(r, e):
+    """
+    Return tuples giving (a, b) where ``a`` is a symbol and b is
```
smichr, line 92:
```suggestion
    Return tuples giving (a, b) where ``a`` is a symbol and ``b`` is
```

sympy/utilities/tests/test_lambdify.py
```diff
@@ -1406,3 +1406,93 @@ def test_cupy_dotproduct():
...
         cupy.array([14])
+
+
+def test_lambdify_cse():
+    def dummy_cse(exprs):
+        return (), exprs
+
+    def minmem(exprs):
+        from sympy.simplify.cse_main import cse_minimize_memory
+        return cse(expr, postprocess=cse_minimize_memory)
```
smichr, line 1417:
```suggestion
        return cse(exprs, postprocess=cse_minimize_memory)
```

sympy/simplify/cse_main.py
```diff
@@ -86,6 +86,73 @@ def cse_separate(r, e):
     e = d[False]
     return [reps_toposort(r), e]
 
+
+def cse_minimize_memory(r, e):
+    """
+    Return tuples giving (a, b) where ``a`` is a symbol and ``b`` is
```
smichr, line 92:
```suggestion
    Return tuples giving ``(a, b)`` where ``a`` is a symbol and ``b`` is
```

robinechuca:
Are there other things to do or is this version satisfactory?

sympy/simplify/cse_main.py
```diff
@@ -770,3 +843,71 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
         return replacements, reduced_exprs
 
     return postprocess(replacements, reduced_exprs)
+
+
+def _cse_homogeneous(exprs, **kwargs):
```
asmeurer, line 848: I still think we don't need this function. The string support in lambdify does not need to be there, and I'm fine if it doesn't work when cse=True. The remaining cases should be handled in cse itself.

sympy/utilities/tests/test_lambdify.py
```diff
@@ -1406,3 +1406,93 @@ def test_cupy_dotproduct():
...
+
+def test_lambdify_cse():
+    def dummy_cse(exprs):
+        return (), exprs
+
+    def minmem(exprs):
+        from sympy.simplify.cse_main import cse_minimize_memory
+        return cse(exprs, postprocess=cse_minimize_memory)
+
+    class Case:
```
asmeurer, line 1419: I would rather write this test using simple data structures and a loop in the test itself, rather than a class. Like

```py
for args, exprs, num_args in [...]:
    assert ...
```

sympy/simplify/cse_main.py
```diff
@@ -86,6 +86,73 @@ def cse_separate(r, e):
     e = d[False]
     return [reps_toposort(r), e]
 
+
+def cse_minimize_memory(r, e):
```
asmeurer, line 90: What was the decision about rather adding a flag to cse to reuse variable names? 

Either way, I'm not a fan of this function name. Something like `delete_variables` would be more descriptive.

smichr: But it's not deleting any variables; it is just identifying some as not being needed past a certain element in the returned list of replacements.

Reuse (or minimal) is not as trivial to implement but that could be done, too. I am not sure that reusing will actually release all variables that are not needed, however. If I am right then it, alone, would not be a solution.

Maybe this function should be called `cse_minimize_variables` and do both compaction and identification of redundancy.

sympy/simplify/cse_main.py
smichr: > should be handled in cse itself.

not without breaking compatibility or adding a new keyword. `cse` always emits a list of expressions whether there is 1 expressions or more.

> lambdify does not need to be there

It is already there
```python
>>> lambdify(x,'x+1')(1)
2
>>> lambdify('x','x+1')(1)
2
>>> lambdify('x',x+1)(1)
2
```

sympy/simplify/cse_main.py
```diff
@@ -657,6 +723,8 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
         concern, use the setting order='none'.
     ignore : iterable of Symbols
         Substitutions containing any Symbol from ``ignore`` will be ignored.
+    asList : bool, (default True)
```
moorepants, line 726: as_list

smichr:
Or `list`

sympy/simplify/cse_main.py
asmeurer: Most functions that accept string inputs do so by accident (although for whatever reason it's actually tested for lambdify). I'd like to deprecate and eventually remove implicit string parsing from SymPy functions (except for the handful where it actually makes sense, and lambdify isn't one of those).

smichr:
I've been trying to work out use of `connected_components` to order equations with replacements so as few working variables can be used at once.

asmeurer:
That seems to be very similar to this problem https://en.wikipedia.org/wiki/Register_allocation

robinechuca:
I see that the debate is slowly dying down. Do I still have to do things? Or is this version satisfactory?

robinechuca:
oups, I made a wrong manipulation!

sympy/utilities/lambdify.py
```diff
@@ -847,7 +856,13 @@ def _lambdifygenerated(x):
         funcprinter = _TensorflowEvaluatorPrinter(printer, dummify) # type: _EvaluatorPrinter
     else:
         funcprinter = _EvaluatorPrinter(printer, dummify)
-    funcstr = funcprinter.doprint(funcname, args, expr)
+
+    if cse != False:
+        from sympy.simplify.cse_main import cse, cse_minimize_memory
+        cses, _expr = cse(expr, as_list=False, postprocess=cse_minimize_memory)
```
moorepants, line 862: I thought the purpose of allowing a callable for the lambdify cse kwarg was to allow for custom cse functions. I'd prefer that the default cse=True only calls the standard cse with no customization.

sympy/simplify/cse_main.py
```diff
@@ -657,6 +726,8 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
         concern, use the setting order='none'.
     ignore : iterable of Symbols
         Substitutions containing any Symbol from ``ignore`` will be ignored.
+    as_list : bool, (default True)
+        Returns expression in list or else with same type as input (when False).
```
moorepants, line 730: From this explanation it isn't clear what this does. An example would help.

robinechuca: By default, if we do not call 'as_list = False', then no test bench works and lambdify will systematically return a list. Which would profoundly change the nature of 'lambdify'. However, by calling 'as_list = False', the behavior of 'lambdify' remains the same regardless of the value of the 'cse' flag.

sympy/utilities/tests/test_lambdify.py
smichr: I view this as a tidy pure-python way of running the test. Do you feel strongly about this @asmeurer? What problem do you see with organizing it this way.

sympy/simplify/cse_main.py
```diff
@@ -699,10 +770,21 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     >>> cse([y**2*(x + 1), 3*y**2*(x + 1)], ignore=(y,))
     ([(x0, x + 1)], [x0*y**2, 3*x0*y**2])
 
+    >>> output = lambda x, as_list: type(cse(x, as_list=as_list)[1])
+    >>> output(cos(x), as_list=False)
+    cos
+    >>> output(cos(x), as_list=True)
+    <class 'list'>
+    >>>
```
smichr, line 778:
```suggestion
    The default return value for the reduced expression(s) is a list, even if there is only
    one expression. The `as_list` flag preserves the type of the input in the output:

    >>> cse(x)
    ([], [x])
    >>> cse(x, as_list=False)
    ([], x)
```

sympy/simplify/cse_main.py
smichr: There is an example in the docstring.

sympy/simplify/cse_main.py
smichr: This seems like a problem that could be solved with the `connected_components` functions...

sympy/simplify/cse_main.py
```diff
@@ -671,14 +742,14 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     Examples
     ========
 
-    >>> from sympy import cse, SparseMatrix
+    >>> from sympy import cse, SparseMatrix, cos
```
smichr, line 745:
```suggestion
    >>> from sympy import cse, SparseMatrix
```

sympy/utilities/lambdify.py
bjodah: I concur, the doc-string even says the user is allowed to pass such a function.

sympy/utilities/tests/test_lambdify.py
bjodah: I suggest we defer style related refactorings in our non-public test suite to a later PR for now (in the interest of actually getting the functionality into SymPy), does that sound OK to you @asmeurer?

sympy/utilities/lambdify.py
smichr, line 862:
```suggestion
    if cse == True:
        from sympy.simplify.cse_main import cse
        cses, _expr = cse(expr, as_list=False)
    elif cse != False:
        cses, _expr = cse(expr)
```

sympy/utilities/lambdify.py
```diff
@@ -847,7 +856,15 @@ def _lambdifygenerated(x):
         funcprinter = _TensorflowEvaluatorPrinter(printer, dummify) # type: _EvaluatorPrinter
     else:
         funcprinter = _EvaluatorPrinter(printer, dummify)
-    funcstr = funcprinter.doprint(funcname, args, expr)
+
+    if cse == True:
+        from sympy.simplify.cse_main import cse
+        cses, _expr = cse(expr, as_list=False)
+    elif cse != False:
```
smichr, line 863:
```suggestion
    elif cse:  # assume callable
```

sympy/utilities/lambdify.py
```diff
@@ -847,7 +856,15 @@ def _lambdifygenerated(x):
         funcprinter = _TensorflowEvaluatorPrinter(printer, dummify) # type: _EvaluatorPrinter
     else:
         funcprinter = _EvaluatorPrinter(printer, dummify)
-    funcstr = funcprinter.doprint(funcname, args, expr)
+
+    if cse == True:
+        from sympy.simplify.cse_main import cse
+        cses, _expr = cse(expr, as_list=False)
+    elif cse:  # assume callable
```
smichr, line 863:
```suggestion
    elif callable(cse):
```

sympy/utilities/tests/test_lambdify.py
```diff
@@ -1406,3 +1406,93 @@ def test_cupy_dotproduct():
...
         f4(1, 2, 3) == \
         cupy.array([14])
+
+
+def test_lambdify_cse():
+    def dummy_cse(exprs):
+        return (), exprs
+
+    def minmem(exprs):
+        from sympy.simplify.cse_main import cse_minimize_memory
```
smichr, line 1416:
```suggestion
        from sympy.simplify.cse_main import cse_minimize_memory, cse
```

sympy/simplify/cse_main.py
```diff
@@ -626,7 +695,7 @@ def _rebuild(expr):
 
 
 def cse(exprs, symbols=None, optimizations=None, postprocess=None,
-        order='canonical', ignore=()):
+        order='canonical', ignore=(), as_list=True):
```
smichr, line 698:
```suggestion
        order='canonical', ignore=(), list=True):
```

sympy/simplify/cse_main.py
```diff
@@ -657,6 +726,8 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
         concern, use the setting order='none'.
     ignore : iterable of Symbols
         Substitutions containing any Symbol from ``ignore`` will be ignored.
+    as_list : bool, (default True)
```
smichr, line 729:
```suggestion
    list : bool, (default True)
```

sympy/simplify/cse_main.py
```diff
@@ -699,10 +770,22 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     >>> cse([y**2*(x + 1), 3*y**2*(x + 1)], ignore=(y,))
     ([(x0, x + 1)], [x0*y**2, 3*x0*y**2])
 
+    The default return value for the reduced expression(s) is a list, even if there is only
+    one expression. The `as_list` flag preserves the type of the input in the output:
```
smichr, line 774:
```suggestion
    one expression. The `list` flag preserves the type of the input in the output:
```

sympy/simplify/cse_main.py
```diff
@@ -699,10 +770,22 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
...
+
+    >>> cse(x)
+    ([], [x])
+    >>> cse(x, as_list=False)
```
smichr, line 778:
```suggestion
    >>> cse(x, list=False)
```

sympy/simplify/cse_main.py
```diff
@@ -699,10 +770,22 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
...
+    ([], x)
     """
     from sympy.matrices import (MatrixBase, Matrix, ImmutableMatrix,
                                 SparseMatrix, ImmutableSparseMatrix)
 
+    if not as_list:
```
smichr, line 784:
```suggestion
    if not as_list:
```
```suggestion
    if not list:
```

sympy/simplify/cse_main.py
```diff
@@ -770,3 +853,70 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
...
+    replacements : list of (Symbol, expression) pairs
+        All of the common subexpressions that were replaced. Subexpressions
+        earlier in this list might show up in subexpressions later in this
+        list.
+    reduced_exprs : list of sympy expressions
+        The reduced expressions with all of the replacements above.
+
+    Examples
+    ========
+    >>> from sympy.simplify.cse_main import cse
```
smichr, line 882:
```suggestion

    >>> from sympy.simplify.cse_main import cse
```

sympy/simplify/cse_main.py
```diff
@@ -770,3 +853,70 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
...
+    >>> from sympy import cos, Tuple, Matrix
+    >>> from sympy.abc import x
+    >>> output = lambda x: type(cse(x, as_list=False)[1])
```
smichr, line 885:
```suggestion
    >>> output = lambda x: type(cse(x, list=False)[1])
```

sympy/simplify/tests/test_cse.py
```diff
@@ -546,6 +546,30 @@ def test_unevaluated_mul():
...
+    # this can change in keeping with the intention of the function
+    assert r, e == ([
+    (x0, x + y), (x1, (x0 - 1)**2), (x2, 2*x + 1),
+    (_3, x0/x2 + x1), (_4, x2**x0), (x2, None), (_0, x1),
+    (x1, None), (_2, x0), (x0, None), (_1, x)], (_0, _1, _2, _3, _4))
+    r.reverse()
+    assert eqs == [i.subs(r) for i in e]
+
+def test_cse_as_list():
+    _cse = lambda x: cse(x, as_list=False)
```
smichr, line 565:
```suggestion
def test_cse_list():
    _cse = lambda x: cse(x, list=False)
```

sympy/utilities/lambdify.py
```diff
@@ -847,7 +856,15 @@ def _lambdifygenerated(x):
         funcprinter = _TensorflowEvaluatorPrinter(printer, dummify) # type: _EvaluatorPrinter
     else:
         funcprinter = _EvaluatorPrinter(printer, dummify)
-    funcstr = funcprinter.doprint(funcname, args, expr)
+
+    if cse == True:
+        from sympy.simplify.cse_main import cse
+        cses, _expr = cse(expr, as_list=False)
```
smichr, line 862:
```suggestion
        cses, _expr = cse(expr, list=False)
```

sympy/utilities/lambdify.py
```diff
@@ -336,6 +336,15 @@ def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
...
         arguments do not redefine any built-in names.
 
+    cse : bool, or callable, optional
+        Large expressions can be computed more efficiently when
+        common subexpressions are identified and precomputed before
+        being used multiple time. Finding the subexpressions will make
+        creation of the 'lambdify' function slower, however.
+
+        When ``True``, ``sympy.simplify.cse`` is used, otherwise (the default)
+        the user may pass a function matching the signature ``cse``.
```
smichr, line 346:
```suggestion
        the user may pass a function matching the ``cse`` signature.
```

robinechuca:
These changes look good to me.

moorepants:
This looks good to me now too! Great addition.

smichr:
The only thing we haven't done is think of a better name for `cse_minimize_memory`; perhaps `cse_release_variables` or `cse_var_none_when_done`?

moorepants:
`cse_release_variables` seems fine to me.

smichr:
`cse_release_variables` can be changed internally as needed since it no longer advertises to minimize memory (a more delicate task). Since it already filled the need identified by @robinechuca and the use of `cse` makes lambdify more efficient, let's commit this.

PR #21547, opened 2021-05-30
Fix: Bug in MinMax which returned an incorrect simplication which is not true for all ranges of the variables

### LINKED ISSUE #21399: nested Max and Min give incorrect result
While working with different types of nested Max and Min expressions, I have encountered a problem in the following expression:
```
from sympy import sympify 
sympify( "Max( Min(a, b), Min(a, b, c))" )
```

output: `Min(a, b, c)`

This simplification is not correct for all possible values of a, b, and c.
I have not found any problems with other similar expressions, and I cannot find the specific reason why this one evaluates incorrectly.

shashankks0987:

Fixes #21399 

earlier, Max( Min(a, b), Min(a, b, c)) returned Min(a,b,c)
now, Max( Min(a, b), Min(a, b, c)) returns Min(a,b)

- Add test for the above fix
- Add the above example in doc

#### Release Notes

* functions
   * Fixed a bug in MaxMin of elementary functions

### DIFF
--- a/sympy/functions/elementary/miscellaneous.py
+++ b/sympy/functions/elementary/miscellaneous.py
@@ -433,8 +433,6 @@ class MinMaxBase(Expr, LatticeOp):
 
         >>> Min(a, Max(b, Min(c, d, Max(a, e))))
         Min(a, Max(b, Min(a, c, d)))
-        >>> Max( Min(a, b), Min(a, b, c))
-        Min(a,b)
         """
         from sympy.utilities.iterables import ordered
         from sympy.simplify.simplify import walk
--- a/sympy/functions/elementary/miscellaneous.py
+++ b/sympy/functions/elementary/miscellaneous.py
@@ -1,4 +1,5 @@
 from sympy.core import Function, S, sympify
+from sympy.utilities.iterables import sift
 from sympy.core.add import Add
 from sympy.core.containers import Tuple
 from sympy.core.compatibility import ordered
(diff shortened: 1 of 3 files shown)


sympy/functions/elementary/miscellaneous.py
```diff
@@ -432,7 +433,8 @@ def _collapse_arguments(cls, args, **assumptions):
 
         >>> Min(a, Max(b, Min(c, d, Max(a, e))))
         Min(a, Max(b, Min(a, c, d)))
-
+        >>> Max( Min(a, b), Min(a, b, c))
+        Min(a,b)
```
oscarbenjamin, line 437: I don't think that an example needs to be added here.

sympy/functions/elementary/tests/test_miscellaneous.py
```diff
@@ -410,6 +410,10 @@ def test_issue_12638():
     assert Min(a, b, Max(a, b, c)) == Min(a, b)
     assert Min(a, b, Max(a, c)) == Min(a, b)
 
+def test_issue_21399():
+    from sympy.abc import a, b, c
+    assert Max( Min(a, b), Min(a, b, c)) == Min(a,b)
```
oscarbenjamin, line 415: The spacing here is off. There should be no space after `(` but there should be a space after `,` so:
```
assert Max(Min(a, b), Min(a, b, c)) == Min(a, b)
```

oscarbenjamin: Also the line spacing between test functions here should be two blank lines between each function. Can you fix that for this function and also those below?

smichr, line 415:
```suggestion
    assert Max(Min(a, b), Min(a, b, c)) == Min(a, b)
```

sympy/functions/elementary/miscellaneous.py
smichr, line 437:
```suggestion
```

smichr:
Thanks, looks good.

shashankks0987:
https://github.com/sympy/sympy/pull/21547/checks?check_run_id=2710942124 was this test always failing?

PR #21551, opened 2021-05-31
Body.apply_torque takes frame as optional argument

sidhu1012:
An optional parameter, `frame` is added to apply_torque() of Body because it's not necessary that user may want to define Torque only with Body's frame.

#### Other comments

Ping @moorepants @Sc0rpi0n101 
#### Release Notes

* physics.mechanics
    * Torque can be applied to `Body` wrt any frame.


sidhu1012:
@moorepants Please review.

PR #21554, opened 2021-06-01
evaluate limits after clipping in Piecewise._eval_interval

smichr:

After clipping symbolic limits there may be unevaluated Min/Max functions that will evaluate if they are rebuilt:
```python
>>> from sympy import *
>>> x = var('x')
>>> y = symbols('y', real=True)
>>> g = Piecewise(
...     (1 - x, Interval(0, 1).contains(x)),
...     (1 + x, Interval(-1, 0).contains(x)),
...     (0, True)
...     )
>>>
>>> g.integrate((x,1,y))
...
-Min(1, Max(0, y))**2 + Min(1, Max(y, Min(-1, 0)))**2/2 + Min(1, Max(y, Min(-1, 0))) - 1/2
...
Piecewise((Min(1, Max(-1, y))**2/2 + Min(1, Max(-1, y)) - Min(1, Max(0, y))**2 - 1/2, y < 1), (0, True))
>>>
```
#### Release Notes

* functions
    * doit/simplify should no longer be needed to evaluate limits after Piecewise integration in some cases

### DIFF
--- a/sympy/functions/elementary/piecewise.py
+++ b/sympy/functions/elementary/piecewise.py
@@ -590,7 +590,12 @@ class Piecewise(Function):
                 elif hi.is_Symbol:
                     pos = pos.xreplace({hi: lo + p}).xreplace({p: hi - lo})
                     neg = neg.xreplace({hi: lo - p}).xreplace({p: lo - hi})
-
+                # evaluate limits that may have unevaluate Min/Max
+                touch = lambda _: _.replace(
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
This could potentially cause slowdowns but I'm not sure how to measure it. Maybe the timings for the slowest tests should be compared with master.

smichr:
> could potentially cause slowdowns

Rough estimate of about 15%. I was able to trace the issue back to the source, however.

Issue #21539, opened 2021-05-28
[GSoC] Improving `sympy.physics.control` : Phase 1


akshanshbhatt:
This phase will primarily involve completing the pending PR for the `TransferFunctionMatrix` class for MIMO systems. I will follow the [Documentation-Driven Development](https://gist.github.com/zsup/9434452) approach to complete this PR. This was suggested by @namannimmo10  in the meeting which we had yesterday. The steps involved will be-

1. Finalizing the characteristics of `TransferFunctionMatrix` and writing good examples for the users (covering all the cases). Being built upon a symbolic library, We can highlight the symbolic features in the examples.
2. Getting the examples verified from my mentors and also general feedback on the overall implementation.
3. Adding docstrings and writing unit tests before starting the development.
4. Coding according to the finalized docs. I will work in the continuation of PR #19761. It will preserve the commits made by Naman.

moorepants:
That all sounds good to me. I back @namannimmo10 fully on the doc driven development. It is essential to develop some comprehensive examples first and then build the new class(es) to support the examples.

namannimmo10:
Sounds good to me too! @akshanshbhatt, can you show an example through which we can get an idea about your proposed API for the transfer function matrix?

akshanshbhatt:
@namannimmo10 I am working on it. I'll share it in a few hours.

akshanshbhatt:
Hey @namannimmo10 and @moorepants, please look at this [gist](https://gist.github.com/akshanshbhatt/8d8175b6d899114da960c8dae934ed5a) and suggest improvements. It has the implementation details of the `TransferFunctionMatrix`'s API.

akshanshbhatt:
@namannimmo10 @moorepants 
I am at present working on the textbook examples/problems. I will give an update on it very soon. Should I open a new pull request for the `TransferFunctionMatrix` now? 
As the WIP PR (#19761) has some minor merge conflicts, I plan to merge it to a new branch created from the current master and then open a PR. Working this way would preserve Naman's commits and resolve the issue with some of the tests failing on the CI (in #19761) due to obsolete dependencies that were removed in this period of 1 year.
If there is a better way, then please let me know.

namannimmo10:
> Should I open a new pull request for the `TransferFunctionMatrix` now?

Yes! And it's better to take lots of feedback on the API design early on, so that the implementation can be done smoothly.

akshanshbhatt:
@namannimmo10 @moorepants 
[...]

PR #19761, opened 2020-07-14
[GSoC] Add TransferFunctionMatrix class in physics.control

namannimmo10:
#### References to other Issues or PRs

https://github.com/sympy/sympy/issues/19352

#### Brief description of what is fixed or changed
This PR introduces MIMO transfer functions in `sympy.physics.control`.

#### Release Notes

*  physics.control
    * Added `TransferFunctionMatrix` class in physics.control

PR #21556, opened 2021-06-01
[WIP][GSoC] Add `TransferFunctionMatrix` class in physics.control

akshanshbhatt:

#21539 

Support for Multiple-Input Multiple-Output (MIMO) Control Systems will be added along with documentation and examples of use cases.

#### Release Notes

*  physics.control
    * Added `TransferFunctionMatrix` class, enabling support for MIMO systems


akshanshbhatt:
@namannimmo10 @moorepants
From where Naman had left (in #19761), I ensured that all the tests on CI were passing to track down the bug if anything goes wrong. While making commits, I will ensure that all the tests pass before moving ahead to avoid a messy situation later. As Non-Expr objects in a `Matrix`-type objects have been deprecated since SymPy 1.9 (#21497), I have commented out the tests of `TransferFunctionMatrix`, which reused `Matrix` class' printing function for `TransferFunctionMatrix`.
I will start working on this PR from tomorrow, after completing the textbook examples. I wanted to ask that for now, should I consider only the examples of `TransferFunctionMatrix` particularly or control systems in general?

namannimmo10:
Sounds good! I think you can work on adding some examples specifically for TFM. And once this PR gets in, do the same for the SISO Transfer function. Let's try to complete the tasks in small chunks in separate PRs — easy to review and less cluttered.

akshanshbhatt:
ping @namannimmo10 @moorepants 
As discussed, my next commit will include all the docstrings and unit tests. I will try to complete it before the beginning of Week 2.

sympy/physics/control/lti.py
```diff
@@ -224,6 +291,22 @@ def var(self):
         """
         return self._var
 
+    # TODO: Remove num_inputs, num_outputs and shape from TF
```
moorepants, line 294: These would likely have to follow a deprecation: https://github.com/sympy/sympy/wiki/Deprecating-policy

sympy/physics/control/lti.py
```diff
@@ -241,25 +324,27 @@ def _eval_simplify(self, **kwargs):
         num_, den_ = tf[0], tf[1]
         return TransferFunction(num_, den_, self.var)
 
-    def expand(self):
-        """
-        Returns the transfer function with numerator and denominator
-        in expanded form.
+    # TODO: Implement expand(tf) by default and remove it from here.
```
moorepants, line 327: Also would need to be deprecated.

sympy/physics/control/lti.py
```diff
@@ -20,25 +20,47 @@ def _roots(poly, var):
...
+
+            $H(s) = \frac{Y(s)}{X(s)} = \frac{ \mathcal{L}\left\{y(t)\right\} }{ \mathcal{L}\left\{x(t)\right\} }$
+
+    Transfer function, $H$, is generally given as a rational function in $s$ as-
+
+            $H(s) =\ \frac{a_{n}s^{n}+a_{n-1}s^{n-1}+\dots+a_{1}s+a_{0}}{b_{m}s^{m}+b_{m-1}s^{m-1}+\dots+b_{1}s+b_{0}}$
+
+    For more info, refer -
+
+            https://en.wikipedia.org/wiki/Transfer_function
```
namannimmo10, line 45: Links go under the References section. See https://docs.sympy.org/latest/documentation-style-guide.html#docstring-sections

sympy/physics/control/lti.py
```diff
@@ -192,7 +245,7 @@ def num(self):
         p*s + s**2 + 3
         >>> G2 = TransferFunction((p + 5)*(p - 3), (p - 3)*(p + 1), p)
         >>> G2.num
-        (p - 3)*(p + 5)
+        p**2 + 2*p - 15
```
namannimmo10, line 248: Well, that was also my first idea back then, but what if the user wants the unexpanded form? Expansion of `num` and `den` is just one step away. Consider this simple example:
```python
>>> from sympy import *
>>> from sympy.abc import *
>>> a = x**2 + x - 2
>>> b = y - 4
>>> a*b
(y - 4)*(x**2 + x - 2)  # no automatic expansion
```

sympy/physics/control/lti.py
```diff
@@ -20,25 +20,47 @@ def _roots(poly, var):
...
+    The numerator of the transfer function is the Laplace transform of the input signal
+    (The signals are represented as functions of time) and similarly the denominator
+    of the transfer function is the Laplace transform of the output signal. It is also a convention
+    to denote the input and output signal's Laplace transform with capital alphabets like shown below.
+
+            $H(s) = \frac{Y(s)}{X(s)} = \frac{ \mathcal{L}\left\{y(t)\right\} }{ \mathcal{L}\left\{x(t)\right\} }$
+
+    Transfer function, $H$, is generally given as a rational function in $s$ as-
+
+            $H(s) =\ \frac{a_{n}s^{n}+a_{n-1}s^{n-1}+\dots+a_{1}s+a_{0}}{b_{m}s^{m}+b_{m-1}s^{m-1}+\dots+b_{1}s+b_{0}}$
```
namannimmo10, line 41: This looks ok. Control professors usually begin with a differential equation and then obtain a transfer function, `G(s)`, from that. Students are also accustomed to that, I guess. Maybe you can do something like

tf

sympy/physics/control/lti.py
akshanshbhatt: Since it has not been released yet, should I write a deprecation warning for it?
Even on the wiki page, it's mentioned:
> Note: Code that has not yet made it into a release does not need to be deprecated.

akshanshbhatt:
```python
@classmethod
    def from_rational_expressions(cls, expr, var=None):
        if var is None:
            _free_symbols = expr.free_symbols
            _len_free_symbols = len(_free_symbols)
            if _len_free_symbols == 1:
                var = list(_free_symbols)[0]
            elif _len_free_symbols == 0:
                raise TypeError("Positional argument `var` not found in the TransferFunction defined. Specify it manually.")
            else:
                raise TypeError("Conflicting values found for positional argument `var` ({}). Specify it manually.".format(_free_symbols))
            
        _num, _den = expr.as_numer_denom()
        return cls(_num, _den, var)
```
@namannimmo10 @moorepants How about this for expression instantiation for `TransferFunction`?

sympy/physics/control/lti.py
akshanshbhatt: I'll do it accordingly.

sympy/physics/control/lti.py
akshanshbhatt: Sure, I'll change it according to this doc.

sympy/physics/control/lti.py
akshanshbhatt: In my opinion, having a definite form would enable us to compare two `TransferFunction` instances in later phases. Moreover, Transfer functions' numerators and denominators are always polynomials, so it should be better, in my opinion, to show them in their typical polynomial form (`a_n*x**n + a_(n-1)*x**(n-1)+...+a_0`). 
Unfortunately, I didn't find a single example in the docs of MATLAB and Mathematica where they used an expression whose `num` and `den` were not already in the polynomial form.

You can have a look-
https://www.mathworks.com/help/control/ref/tf.html
https://reference.wolfram.com/language/ref/TransferFunctionZeros.html
https://harold.readthedocs.io/en/latest/generated/harold.Transfer.html#harold.Transfer

In MATLAB and Harold, they have strictly implemented the polynomial form. Users have no freedom to instantiate a transfer function by inputting an expression. They can only input the coefficients of the polynomial. There is a form in MATLAB to create a transfer function instance with poles and zeros, but that also maintains its polynomial characteristic.

I'll do what you suggest as a lot of discussions have already been done.

namannimmo10: Okay, fair enough, let's go with the polynomial form!

sympy/physics/control/lti.py
akshanshbhatt: @namannimmo10 I have changed `TransferFunction`'s docstrings according to this. Please review it.

akshanshbhatt:
@moorepants @namannimmo10 
How can we add pretty-printed output in the docstrings?
It shows 
```
AssertionError: File contains a unicode character
```
in the quality test.
Is there any way to fix this?

namannimmo10:
I would not use `pprint`. You can start `init_printing` session like
```python
>>> from sympy import init_printing
>>> init_printing(use_unicode=True, wrap_line=False)
```

But it will pretty-print every possible output.

akshanshbhatt:
> But it will pretty-print every possible output.

That is the downside of `init_printing`. I don't want every output to be pretty-printed.

akshanshbhatt:
I have passed the `use_unicode` flag `False` this time, and all the quality tests pass. Also, it is somehow more visually pleasing than the default.

doc/src/modules/physics/control/control_problems.rst
```diff
@@ -0,0 +1,147 @@
...
+        2 & 2
+        \end{bmatrix}$
+
+* Also, find the **equivalent closed-loop system** *(or the ratio v/u from the block diagram given below)* for the system (negative-feedback loop) having **C(s)** as the **controller** and **P(s)** as **plant** *(Refer to the block diagram given below)*.
+
+        .. image:: https://user-images.githubusercontent.com/53227127/120820301-0b368c80-c572-11eb-84c4-e372190cf0bd.png
+
+Solution-
+
+    >>> from sympy import *
```
moorepants, line 42: Do explicit imports.

oscarbenjamin:
Do we really need a `TransferFunctionMatrix` class?

Can it be implemented more simply as a wrapper around existing matrix classes or does it really need to be a subclass of ImmutableMatrix?

I have spent quite a lot of time trying to simplify the matrices code and to *eliminate* various Matrix subclasses such as NewMatrix, RawMatrix, PolyMatrix, etc. The primary reason I did this was to (re-)establish the internal invariant that the elements of a Matrix should be Expr. In master you will see a DeprecationWarning now if you try to put non-Expr into a Matrix but I see now that the class here subclasses ImmutableMatrix and I haven't included the DeprecationWarning there.

Subclasses that break the internal invariants (e.g. elements are Expr) of a superclass are bad subclasses and make it hard to improve the superclass (i.e. Matrix) for more general use. I have eliminated these subclasses and deprecated non-Expr in a Matrix so that I can replace the internals of the Matrix/ImmutableMatrix classes with a more efficient implementation (#21626)

This PR also seems to duplicate code that exists elsewhere e.g. by reimplementing Matrix multiplication (and not in a way that improves over the existing implementations).

sympy/physics/control/lti.py
moorepants: If it hasn't been released, then no. I thoutht this was TransferFunction code.

moorepants:
Oscar, we were just in a video call discussing the design of this class. We have TransferFunction objects that can be grouped to describe multivariate linear systems in the Laplace domain. This TransferFunctionMatrix would let us do that. We're open for other suggestions. We need to be able to realize block diagram algebra for a MIMO system. Mathematica supports this with a TransferFunctionModel class.

moorepants:
We implemented TransferFunction, Series, Parallel, and Feedback classes that correlate to an Expr, Mul, Add, and solve. We need the same thing for these TransferFunction matrices.

oscarbenjamin:
> We need the same thing for these TransferFunction matrices.

Yes, but why does it need to be a *subclass* of ImmutableMatrix?

Can it just be a class that has an ordinary Matrix as an internal attribute? Or perhaps one Matrix for numerators and another for denominators?

What actual Matrix methods does this need? I presume that e.g. slicing doesn't make any sense here.

If this is just needed for e.g. matrix multiplication then we already have functions that can do that without needing to subclass ImmutableMatrix.

At some point there will be a cleaner way to implement this sort of thing but the current public (Immutable)Matrix classes are not suitable because there is no controllable way for them to understand what the "type" of their elements is. DomainMatrix can do this better but probably needs to be wrapped in a nicer interface.

moorepants:
> Yes, but why does it need to be a subclass of ImmutableMatrix?

I doubt it does. We didn't originally design it like that. I think this is @akshanshbhatt's new idea. It seems reasonable on the face of things, but I don't know the fine details of how matrices work to know if it isn't.

@akshanshbhatt, sounds like we should just make our on class and use some matrix functions for the few operations we need.

oscarbenjamin:
> @akshanshbhatt, sounds like we should just make our on class and use some matrix functions for the few operations we need.

As far as I can tell from the tests the only matrix-like operations that are needed here are `+`, `-`, `*`, `__new__` and printing. Almost all of these are actually overridden in this PR so the inherited implementation is not being used with the exception of `__neg__` (which is a trivial routine anyway).

If you look here you can find examples of implementations of matrix operations that can work with any list of lists:
https://github.com/sympy/sympy/blob/8b06e0aa91e5a161afc22bdb7b663c4badaf51fc/sympy/polys/matrices/dense.py#L79-L85

Example:
```python
In [21]: from sympy.polys.matrices.dense import ddm_imatmul                                          

In [22]: C = [[0, 0], [0, 0]]                                                                        

In [23]: A = [[1, 2], [3, 4]]                                                                        

In [24]: ddm_imatmul(C, A, A)                                                                        

In [25]: C                                                                                           
Out[25]: [[7, 10], [15, 22]]

In [26]: Matrix(A)**2                                                                                
Out[26]: 
⎡7   10⎤
⎢      ⎥
⎣15  22⎦
```
I deliberately designed those functions as separate functions not part of any class so that they could be used without subclassing anything.

Alternatively a Domain object could be created for TransferFunction and then DomainMatrix could be used for all matrix operations:
```python
from sympy import Symbol, sympify
from sympy.polys.matrices import DomainMatrix
from sympy.physics.control import TransferFunction

class TFD:
    def __init__(self, var):
        self.var = var
        self.zero = TransferFunction(0, 1, var)
        self.one = TransferFunction(1, 1, var)

    def to_sympy(self, val): # only needed for to_Matrix
        return val

    def __call__(self, num, den=None):
        if den is None:
            num, den = num.as_numer_denom()
        return TransferFunction(num, den, self.var)

    def __repr__(self):
        return f'TFD({self.var})'

s = Symbol('s')
K = TFD(s)
dM = DomainMatrix([[K(s), K(s**2)], [K(1/s), K(s)]], (2, 2), K)
```
That gives (using `to_Matrix` only for printing purposes):
```python
In [80]: dM.to_Matrix()                                                                              
Out[80]: 
⎡    2⎤
⎢s  s ⎥
⎢─  ──⎥
⎢1  1 ⎥
⎢     ⎥
⎢1  s ⎥
⎢─  ─ ⎥
⎣s  1 ⎦

In [81]: (dM**2).to_Matrix()                                                                         
Out[81]: 
⎡              ⎛ 2⎞              ⎛ 2⎞   ⎛ 2⎞    ⎤
⎢0   ⎛s⎞ ⎛s⎞   ⎜s ⎟ ⎛1⎞  0   ⎛s⎞ ⎜s ⎟   ⎜s ⎟ ⎛s⎞⎥
⎢─ + ⎜─⎟⋅⎜─⎟ + ⎜──⎟⋅⎜─⎟  ─ + ⎜─⎟⋅⎜──⎟ + ⎜──⎟⋅⎜─⎟⎥
⎢1   ⎝1⎠ ⎝1⎠   ⎝1 ⎠ ⎝s⎠  1   ⎝1⎠ ⎝1 ⎠   ⎝1 ⎠ ⎝1⎠⎥
⎢                                               ⎥
⎢                                ⎛ 2⎞           ⎥
⎢0   ⎛1⎞ ⎛s⎞   ⎛s⎞ ⎛1⎞   0   ⎛1⎞ ⎜s ⎟   ⎛s⎞ ⎛s⎞ ⎥
⎢─ + ⎜─⎟⋅⎜─⎟ + ⎜─⎟⋅⎜─⎟   ─ + ⎜─⎟⋅⎜──⎟ + ⎜─⎟⋅⎜─⎟ ⎥
⎣1   ⎝s⎠ ⎝1⎠   ⎝1⎠ ⎝s⎠   1   ⎝s⎠ ⎝1 ⎠   ⎝1⎠ ⎝1⎠ ⎦

In [82]: (dM**2).applyfunc(lambda e: e.doit()).to_Matrix()                                           
Out[82]: 
⎡ 3    2      3  ⎤
⎢s  + s    2⋅s   ⎥
⎢───────   ────  ⎥
⎢   s       1    ⎥
⎢                ⎥
⎢    2     3    2⎥
⎢ 2⋅s     s  + s ⎥
⎢ ────    ───────⎥
⎢   2        s   ⎥
⎣  s             ⎦
```
I've cut them out but if you run the above then you will see deprecation warnings coming from `to_Matrix` which is warning about the use of non-Expr in a Matrix.

akshanshbhatt:
> @akshanshbhatt, sounds like we should just make our own class and use some matrix functions for the few operations we need.

Maybe, we can revert to Naman's implementation (internal structure) of the `TransferFunctionMatrix` and instead of using `ImmutableMatrix` as one of the superclasses, we can add a new attribute to the `TransferFunctionMatrix` object to store the equivalent Expr-Matrix instance.

Something like - 

```python
# Inside __new__() of TransferFunctionMatrix
...
obj = super(TransferFunctionMatrix, cls).__new__(cls, arg)
obj._expr_mat = ImmutableMatrix(expr_list)
# expr_list will be the nested list of Expr, converted from TransferFunction.
# e.g.
# tf_list = [[TransferFunction(1, s, s), TransferFunction(1, 5, s)], [TransferFunction(s, s**2 + 1, s), TransferFunction(1, s - 5, s)]]
# Equivalent expr_list would be:
# expr_list = [[1/s, 1/5], [s/(s**2 + 1), 1/(s - 5)]]
...
```

Using this -
 * Will resolve the issue of being a subclass of `ImmutableMatrix`.
 * We would be able to utilize the efficient addition and multiplication of Matrices without redefining on our own (which was one of the downsides in Naman's implementation).
 * We would have control over the attributes that we really want, instead of inheriting all the methods and attributes of `ImmutableMatrix`.
 * obj.func(*obj.args) would work as expected.

If you have any better idea please suggest.

sympy/physics/control/lti.py
```diff
@@ -135,34 +184,118 @@ class TransferFunction(Basic, EvalfMixin):
...
+            obj._den = den
+            obj._var = var
+            obj._num_inputs, obj._num_outputs = 1, 1 # TODO: Remove _num_inputs and _num_outputs for TF and everywhere else except TFM
+            return obj
+
         else:
             raise TypeError("Unsupported type for numerator or denominator of TransferFunction.")
 
+    @classmethod
+    def from_rational_expression(cls, expr, var=None):
```
namannimmo10, line 226: @oscarbenjamin @asmeurer, does sympy allow such classmethods in the codebase?

namannimmo10: I see several similar methods in this file:
https://github.com/sympy/sympy/blob/21aa3ab7111292aad003827164639716c84c5294/sympy/polys/matrices/domainmatrix.py#L215

sympy/physics/control/lti.py
```diff
@@ -19,15 +20,66 @@ def _roots(poly, var):
...
     are ``num``, ``den``, and ``var``, where ``num`` and ``den`` are numerator and
     denominator polynomials of the ``TransferFunction`` respectively, and the third argument is
     a complex variable of the Laplace transform used by these polynomials of the transfer function.
     ``num`` and ``den`` can be either polynomials or numbers, whereas ``var``
     has to be a Symbol.
 
+    Explanation
+    ===========
+
+    Generally, a dynamical system respresenting a physical system can be described in terms of Linear
```
ilayn, line 35: dynamical systems are more general than physical systems. I think you meant to say, often, physical systems can be modeled by Lin. Ord. Diff. Eq. ....

sympy/physics/control/lti.py
```diff
@@ -19,15 +20,66 @@ def _roots(poly, var):
...
+    Ordinary Differential Equations like -
+
+            $\small{b_{m}y^{\left(m\right)}+b_{m-1}y^{\left(m-1\right)}+\dots+b_{1}y^{\left(1\right)}+b_{0}y=
+            a_{n}x^{\left(n\right)}+a_{n-1}x^{\left(n-1\right)}+\dots+a_{1}x^{\left(1\right)}+a_{0}x}$
+
+    Here, $x$ is the input signal and $y$ is the output signal and superscript on both is the order of derivative
```
ilayn, line 41: x is typically reserved for State variables and inputs are often denoted by u.

ilayn: You can be a bit more explicit at this level I think

```
...superscripts denote the order of differentiation and not the exponentiation, that is to say y^2 is the second derivative and not the square of y, ....
```

sympy/physics/control/lti.py
```diff
@@ -19,15 +20,66 @@ def _roots(poly, var):
...
+    (not exponent). Derivative is taken with respect to the independent variable, $t$. Also, generally $m$ is greater
+    than $n$.
```
ilayn, line 43: Since most of the systems that are modeled are causal, right?

sympy/physics/control/lti.py
```diff
@@ -19,15 +20,66 @@ def _roots(poly, var):
...
+
+    It is not feasible to analyse the properties of such systems in their native form therefore, we use
```
ilayn, line 45: The main advantage of this is because dealing with bunch of polynomials are much easier than solving arbitrary order differential equations. And as is the case with any integral transforms, Laplace transform helps analyzing the diff.eq. set in a straightforward way.

sympy/physics/control/lti.py
```diff
@@ -19,15 +20,66 @@ def _roots(poly, var):
...
+            $\small{b_{m}\mathcal{L}[y^{\left(m\right)}]+\dots+b_{1}\mathcal{L}[y^{\left(1\right)}]+b_{0}\mathcal{L}[y]=
+            a_{n}\mathcal{L}[x^{\left(n\right)}]+\dots+a_{1}\mathcal{L}[x^{\left(1\right)}]+a_{0}\mathcal{L}[x]}$
+
+    Now, applying Derivative property of Laplace transform,
+
+            $\small{b_{m}s^{m}\mathcal{L}[y]+b_{m-1}s^{m-1}\mathcal{L}[y]+\dots+b_{1}s\mathcal{L}[y]+b_{0}\mathcal{L}[y]=
+            a_{n}s^{n}\mathcal{L}[x]+a_{n-1}s^{n-1}\mathcal{L}[x]+\dots+a_{1}s\mathcal{L}[x]+a_{0}\mathcal{L}[x]}$
+
+    Here, the superscript on $s$ is exponent. Note that the zero initial conditions assumption, mentioned above, is very important
+    and cannot be ignored otherwise the dynamical system cannot be considered time-independent and the simplified equation above
```
ilayn, line 65: No the system will still be time-invariant but your equations are going to be messy since there will be X(0)-type variables everywhere. But still, it is time-invariant since these are constants. You wouldn't end up with a simple Y/U ratio.

sympy/physics/control/lti.py
```diff
@@ -19,15 +20,66 @@ def _roots(poly, var):
...
+    cannot be reached.
+
+    Collecting $\mathcal{L}[y]$ and $\mathcal{L}[x]$ terms from both the sides and taking the ratio
+    $\frac{ \mathcal{L}\left\{y\right\} }{ \mathcal{L}\left\{x\right\} }$, we get the typical rational form of transfer
+    function.
+
+    The numerator of the transfer function is therefore, the Laplace transform of the output signal
+    (The signals are represented as functions of time) and similarly the denominator
+    of the transfer function is the Laplace transform of the input signal. It is also a convention
+    to denote the input and output signal's Laplace transform with capital alphabets like shown below.
```
ilayn, line 75: capital letters probably.

ilayn (commented):
Thanks for the extensive PR. It looks really good. I wasted my time on the text instead of doing proper code review. When I have time I'll try to come back to it.

Issue #11402, opened 2016-07-16
pointwise differentiation for Piecewise is wrong


asmeurer:
```
In [7]: Piecewise((sin(x)/x, Ne(x, 0)), (1, True))
Out[7]:
⎧sin(x)
⎪──────  for x ≠ 0
⎨  x
⎪
⎩  1     otherwise

In [8]: Piecewise((sin(x)/x, Ne(x, 0)), (1, True)).diff(x)
Out[8]:
⎧cos(x)   sin(x)
⎪────── - ──────  for x ≠ 0
⎪  x         2
⎨           x
⎪
⎪       0         otherwise
⎩

In [9]: Piecewise((sin(x)/x, Ne(x, 0)), (1, True)).diff(x, x)
Out[9]:
⎧          2⋅cos(x)   2⋅sin(x)
⎪-sin(x) - ──────── + ────────
⎪             x           2
⎪                        x
⎨─────────────────────────────  for x ≠ 0
⎪              x
⎪
⎪              0                otherwise
⎩
```

The otherwise case, to be correct, should use a limit for the derivative. The function is `sinc`. The second derivative at 0 should be -1/3:

```
In [10]: limit(sinc(x).diff(x), x, 0)
Out[10]: 0

In [11]: limit(sinc(x).diff(x, x), x, 0)
Out[11]: -1/3
```

kritkaran94:
I want to take up this issue. @asmeurer How should I start on this ?

asmeurer:
I'm curious how other systems handle this. Does Mathematica do the right thing here?

kritkaran94:
@asmeurer I don't have mathematica installed. So, I will try to check on this.

kritkaran94:
In Mathematica, the output for this particular example is pretty much similar to what we have in Sympy.

````
f[x_] = D[Sinc[x], x];
f[x]
Cos[x]/x - Sin[x]/x^2
````
In its current form, the value at x = 0 is indeterminate. It is only when taking the limit as x->0 that a value emerges. Hence:

````
Limit[f[x], x -> 0]
0
````
@asmeurer  please visit [**Reference**](http://mathematica.stackexchange.com/questions/42159/make-sinc0-return-0-instead-of-indeterminate) for similar problem.

asmeurer:
sinc is a separate issue. The problem there is that there's no simple way to represent the derivative. See https://math.stackexchange.com/questions/1838572/more-convenient-form-of-derivative-of-mathrmsincx.

[...]

PR #21568, opened 2021-06-02
Make sinc(x).diff(x) not return a Piecewise

asmeurer:
Currently, Piecewise.diff() is wrong for pointwise removable singularities.
This lead to sinc(x).diff(x, x) being wrong at x = 0. See
https://github.com/sympy/sympy/issues/11402.

The issue with Piecewise is difficult to solve, but we can work around sinc()
specifically by just returning cos(x)/x - sin(x)/x**2 for sinc(x).diff(x)
rather than a Piecewise. The value of the derivatives of sinc at 0 will have
to be evaluated with limit(), but this is preferable to having an incorrect
answer.

#### Release Notes

- functions
  - The derivatives of sinc(x) no longer return Piecewise results, which were previously incorrect at x=0. The value of the derivatives at x=0 can be computed with `limit`.


smichr:
looks good as a stop-gap

asmeurer:
Worth pointing out that changing differentiation can also affect integration:

```
# master
>>> integrate(sinc(x)*x, x)
-cos(x)
# This branch
>>> integrate(sinc(x)*x, x)
x**2*sinc(x)/2 - x*sin(x)/2 - cos(x)
```

I'm actually surprised integration works in master given that the derivative is a Piecewise.

PR #21574, opened 2021-06-04
Change check of matrix in stats.Normal to allow symengine matrix input

rikardn:

Use `is_Matrix` to check if input is a matrix in `stats.Normal`. This allows for symengine matrices to be used as input.

#### Release Notes

NO ENTRY


smichr:
This looks ok. Is there a test that can be added?

oscarbenjamin:
It looks like we don't really have any significant symengine tests. It would be good to add those somewhere but in the mean time we should just merge this I think.

rikardn:
Thanks @smichr and @oscarbenjamin for reviewing.

Issue #18348, opened 2020-01-15
Refactor the ODE module using classes for solvers


oscarbenjamin:
This is a suitable GSOC idea.

Currently the ODE module (sympy/solvers/ode.py) which implements the `dsolve` function is messy. The code there would be easier to understand and maintain if each of the separate solvers was implemented as a class.

The idea here is to have classes like
```python
class Solver:
    def __init__(self, eq, func):
        # Store the inputs passed to dsolve
        self.eq = eq
        self.func = func

class FirstLinear(Solver):
    def match(self):
        # return True if this method can solve the equation

    def get_general_solution(self):
        # return the general solution
```

The `dsolve` can be rewritten so it looks more like this:
```python
all_solvers = [
    FirstLinear,
    ...
]

def dsolve(eq, func):
    solvers = []
    for solvercls in all_solvers:
        solver = solvercls(eq, func)
        if solver.matches():
            solvers.append(solver)

    # Get the first solver that matched and use it to return the solution
    return solvers[0].get_general_solution()
```
Rewriting the code like this will make it much easier to understand and maintain. This can be done one solver at a time so if interested in this as a GSOC idea you could start beforehand by doing one or two solvers.

The real benefit of doing this will come at the end when `dsolve` can be written like
```python
def dsolve(eq, func):
    solvers = []
    for solvercls in all_solvers:
        solver = solvercls(eq, func)
        if solver.matches():
            # Shortcut this loop and return the solution immediately
            return solver.get_general_solution()
```
This will make dsolve *much* faster for most ODEs.

Mohitbalwani26:
so different methods should have different classes?

risubaba:
@oscarbenjamin I would like to try doing this. However i am not really familiar with `ode.py`. Could you please guide me how i should start with this.

risubaba:
In `_desolve()`
```diff
-    type = kwargs.get('type', None)
+    de_type = kwargs['type']
     xi = kwargs.get('xi')
     eta = kwargs.get('eta')
     x0 = kwargs.get('x0', 0)
     terms = kwargs.get('n')
 
-    if type == 'ode':
+    if de_type == 'ode':
         from sympy.solvers.ode import classify_ode, allhints
         classifier = classify_ode
         string = 'ODE '
         dummy = ''
 
-    elif type == 'pde':
+    else:
         from sympy.solvers.pde import classify_pde, allhints
         classifier = classify_pde
         string = 'PDE '

```
[...]

PR #21578, opened 2021-06-06
Refactor separable reduced

Mohitbalwani26:

SEE #18348

`separable_reduced` is refactored, as it used the same function as ode_separable after extracting coefficients, `SeparableReduced`  is inherited from `Separable` class.

#### Release Notes

NO ENTRY

### DIFF
--- a/doc/src/modules/solvers/ode.rst
+++ b/doc/src/modules/solvers/ode.rst
@@ -157,7 +157,6 @@ linear_coefficients
 separable_reduced
 ^^^^^^^^^^^^^^^^^
 .. autofunction:: sympy.solvers.ode.single::SeparableReduced
-   :members:
 
 lie_group
 ^^^^^^^^^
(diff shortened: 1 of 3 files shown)


oscarbenjamin:
This looks good.

This solver seems like a good example where it's really just match and substitute. It would be better to express everything in terms of some generic matching code and an ODE substitution rather than having a load of special (probably buggy) code to try and process a match for this particular case.

doc/src/modules/solvers/ode.rst
```diff
@@ -156,7 +156,7 @@ linear_coefficients
 
 separable_reduced
 ^^^^^^^^^^^^^^^^^
-.. autofunction:: sympy.solvers.ode.ode::ode_separable_reduced
+.. autofunction:: sympy.solvers.ode.single::SeparableReduced
```
Mohitbalwani26, line 160: I had a doubt as it was inherited from `Separable` class, In `ode.rst` when I wrote :members: it gave an error can yu explain the reason?

PR #21580, opened 2021-06-06
Fix: Bug in is_subset function of set handlers.

### LINKED ISSUE #19513: bug in is_subset(Reals)
```python
In [1]: print( (1/n).is_real )
None

In [2]: imageset(Lambda(n, 1/n), Integers).is_subset(Reals)
Out[2]: True
```
`Out[2]` should also be `None`.

shashankks0987:

Fixes #19513

earlier, imageset(Lambda(n, 1/n), Integers).is_subset(Reals) incorrectly returned True
now, imageset(Lambda(n, 1/n), Integers).is_subset(Reals) returns None

- Add test for the above scenario

#### Release Notes

* sets
  * Fixed a bug in is_subset()

### DIFF
--- a/sympy/sets/handlers/intersection.py
+++ b/sympy/sets/handlers/intersection.py
@@ -1,5 +1,5 @@
 from sympy import (S, Dummy, Lambda, symbols, Interval, Intersection, Set,
-                   EmptySet, FiniteSet, Union, ComplexRegion)
+                   EmptySet, FiniteSet, Union, ComplexRegion, Mul)
 from sympy.multipledispatch import dispatch
 from sympy.sets.conditionset import ConditionSet
 from sympy.sets.fancysets import (Integers, Naturals, Reals, Range,
@@ -304,17 +304,19 @@ def intersection_sets(self, other): # noqa:F811
             return None
         else:
             # univarite imaginary part in same variable
-            x, xis = solve_linear(n, im)
-            if x == n:
+            x, xis = zip(*[solve_linear(i, 0) for i in Mul.make_args(im) if n in i.free_symbols])
+            if x and all(i == n for i in x):
                 base_set -= FiniteSet(xis)
             else:
                 base_set -= ConditionSet(n, Eq(im, 0), S.Integers)
         # exclude values that make denominators 0
         for i in denoms(f):
             if i.has(n):
-                x, xis = solve_linear(i, n)
-                if x == n:
-                    base_set -= FiniteSet(xis)
+                sol = list(zip(*[solve_linear(i, 0) for i in Mul.make_args(im) if n in i.free_symbols]))
+                if sol != []:
+                    x, xis = sol
+                    if x and all(i == n for i in x):
+                        base_set -= FiniteSet(xis)
                 else:
                     base_set -= ConditionSet(n, Eq(i, 0), S.Integers)
         return imageset(lam, base_set)
--- a/sympy/sets/handlers/intersection.py
+++ b/sympy/sets/handlers/intersection.py
@@ -277,8 +277,9 @@ def intersection_sets(self, other): # noqa:F811
                 return FiniteSet(*(fn.subs(n, s[0]) for s in solns))
 
     if other == S.Reals:
-        from sympy.solvers.solveset import solveset_real
         from sympy.core.function import expand_complex
+        from sympy.core.relational import Eq
+        from sympy.solvers.solvers import denoms, solve_linear
 
         f = self.lamda.expr
         n = self.lamda.variables[0]
@@ -293,7 +294,7 @@ def intersection_sets(self, other): # noqa:F811
         im = im.subs(n_, n)
         ifree = im.free_symbols
         lam = Lambda(n, re)
-        if not im:
+        if im.is_zero:
             # allow re-evaluation
             # of self in this case to make
             # the result canonical
@@ -304,7 +305,21 @@ def intersection_sets(self, other): # noqa:F811
             return None
         else:
             # univarite imaginary part in same variable
-            base_set = base_set.intersect(solveset_real(im, n))
+            from sympy.solvers.solvers import denoms, solve_linear
+            from sympy.core.relational import Eq
+            x, xis = solve_linear(im, n)
+            if x == n:
+                base_set -= FiniteSet(xis)
+            else:
+                base_set -= ConditionSet(n, Eq(im, 0), S.Integers)
+        # exclude values that make denominators 0
+        for i in denoms(f):
+            if i.has(n):
+                x, xis = solve_linear(i, n)
+                if x == n:
+                    base_set -= FiniteSet(xis)
+                else:
+                    base_set -= ConditionSet(n, Eq(i, 0), S.Integers)
         return imageset(lam, base_set)
 
     elif isinstance(other, Interval):
--- a/sympy/sets/handlers/intersection.py
+++ b/sympy/sets/handlers/intersection.py
@@ -304,7 +304,7 @@ def intersection_sets(self, other): # noqa:F811
             return None
         else:
             # univarite imaginary part in same variable
-            x, xis = solve_linear(im, n)
+            x, xis = solve_linear(n, im)
             if x == n:
                 base_set -= FiniteSet(xis)
             else:
--- a/sympy/sets/handlers/intersection.py
+++ b/sympy/sets/handlers/intersection.py
@@ -278,9 +278,8 @@ def intersection_sets(self, other): # noqa:F811
 
     if other == S.Reals:
         from sympy.core.function import expand_complex
-        from sympy.core.relational import Eq
         from sympy.solvers.solvers import denoms, solve_linear
-
+        from sympy.core.relational import Eq
         f = self.lamda.expr
         n = self.lamda.variables[0]
 
@@ -305,8 +304,6 @@ def intersection_sets(self, other): # noqa:F811
             return None
         else:
             # univarite imaginary part in same variable
-            from sympy.solvers.solvers import denoms, solve_linear
-            from sympy.core.relational import Eq
             x, xis = solve_linear(im, n)
             if x == n:
                 base_set -= FiniteSet(xis)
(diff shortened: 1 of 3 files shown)


sympy/sets/handlers/intersection.py
```diff
@@ -277,9 +277,9 @@ def intersection_sets(self, other): # noqa:F811
                 return FiniteSet(*(fn.subs(n, s[0]) for s in solns))
 
     if other == S.Reals:
-        from sympy.solvers.solveset import solveset_real
```
smichr, line 280: Although this PR can be patched to use `solve_linear` isn't it better to use the full `solveset_real`? Or maybe you are thinking to keep that for the `simplify` method? But as of now, `simplify` doesn't do anything to the ConditionSet.

In addition, you are incorrectly passing args to `solve_linear`; check parameters.

shashankks0987: Based on the limited knowledge I have, I thought using solveset_real would be computationally expensive. Although using solveset_real would make this implementation a lot simpler

shashankks0987: @smichr Should I use solveset_real or is this cool? I've fixed the other stuff

smichr: There is no choice but to use `solveset_real` or else even a simple system's intersection won't be computed. Is there another method that will resolve this simple failing case? `solve_linear` is ok at instantiation where we don't want to work too hard, but for the `intersection` method, we have to do our best to give a solution.

And you are still not using `solve_linear` correctly. Try solving a simple equation like `x + 1 = 0` after reading the docstring and confirm that you get `(x, -1)` as the solution.

oscarbenjamin: A function like `solveset_real` can be very slow in some cases. I don't think that intersection or instantiation should call a potentially slow function like this. It is actually especially bad in something like an intersection handler because these handlers are called for all *pairs* of sets in an intersection. That makes computing an intersection of a large number of sets very slow e.g.:
```python
In [9]: s = [ImageSet(Lambda(n, x**n), Integers) for x in range(2, 100)]

In [10]: Intersection(*s)
^C---------------------------------------------------------------------------
KeyboardInterrupt  
```
If there is a need to call `solveset` then the whole operation should be handled as part of simplification rather than in any part of evaluation. The intersect handlers need to be extremely fast because they are themselves called in an algorithmically bad `O(n**2)` way.

smichr: OK. Then we need to add this to `simplify` at some point. As for the issue here, it is easy to make `solve_linear` still work for the given example by arg-wise calling `solve_linear`:
```diff
diff --git a/sympy/sets/handlers/intersection.py b/sympy/sets/handlers/intersection.py
index 2a34869063..3691b864eb 100644
--- a/sympy/sets/handlers/intersection.py
+++ b/sympy/sets/handlers/intersection.py
@@ -1,5 +1,5 @@
 from sympy import (S, Dummy, Lambda, symbols, Interval, Intersection, Set,
-                   EmptySet, FiniteSet, Union, ComplexRegion)
+                   EmptySet, FiniteSet, Union, ComplexRegion, Mul)
 from sympy.multipledispatch import dispatch
 from sympy.sets.conditionset import ConditionSet
 from sympy.sets.fancysets import (Integers, Naturals, Reals, Range,
@@ -304,16 +304,16 @@ def intersection_sets(self, other): # noqa:F811
             return None
         else:
             # univarite imaginary part in same variable
-            x, xis = solve_linear(im, n)
-            if x == n:
+            x, xis = zip(*[solve_linear(i, 0) for i in Mul.make_args(im) if n in i.free_symbols])
+            if x and all(i == n for i in x):
                 base_set -= FiniteSet(xis)
             else:
                 base_set -= ConditionSet(n, Eq(im, 0), S.Integers)
         # exclude values that make denominators 0
         for i in denoms(f):
             if i.has(n):
-                x, xis = solve_linear(i, n)
-                if x == n:
+                x, xis = zip(*[solve_linear(i, 0) for i in Mul.make_args(im) if n in i.free_symbols])
+                if x and all(i == n for i in x):
                     base_set -= FiniteSet(xis)
                 else:
                     base_set -= ConditionSet(n, Eq(i, 0), S.Integers)
```

sympy/sets/handlers/intersection.py
```diff
@@ -304,7 +304,7 @@ def intersection_sets(self, other): # noqa:F811
             return None
         else:
             # univarite imaginary part in same variable
-            x, xis = solve_linear(im, n)
+            x, xis = solve_linear(n, im)
```
smichr, line 307: this solves the expression `n = im`...and you need to solve `im == 0`.

sympy/sets/handlers/intersection.py
```diff
@@ -304,17 +304,19 @@ def intersection_sets(self, other): # noqa:F811
...
             else:
                 base_set -= ConditionSet(n, Eq(im, 0), S.Integers)
         # exclude values that make denominators 0
         for i in denoms(f):
             if i.has(n):
-                x, xis = solve_linear(i, n)
-                if x == n:
-                    base_set -= FiniteSet(xis)
+                sol = list(zip(*[solve_linear(i, 0) for i in Mul.make_args(im) if n in i.free_symbols]))
+                if sol != []:
```
shashankks0987, line 316: This is because solve_linear sometimes returns an empty list in some cases

sympy/sets/handlers/intersection.py
shashankks0987: ok understood

sympy/sets/tests/test_fancysets.py
```diff
@@ -656,9 +657,7 @@ def test_infinitely_indexed_set_2():
 def test_imageset_intersect_real():
     from sympy import I
     from sympy.abc import n
-    assert imageset(Lambda(n, n + (n - 1)*(n + 1)*I), S.Integers).intersect(S.Reals) == \
-            FiniteSet(-1, 1)
-
+    assert imageset(Lambda(n, n + (n - 1)*(n + 1)*I), S.Integers).intersect(S.Reals) == Complement(S.Integers, FiniteSet((-1, 1)))
```
oscarbenjamin, line 660: This doesn't seem correct

oscarbenjamin: @smichr @shashankks0987 this output is incorrect:
```python
In [8]: S1 = imageset(Lambda(n, n + (n - 1)*(n + 1)*I), S.Integers)

In [9]: S1
Out[9]: {n + ⅈ⋅(n - 1)⋅(n + 1) │ n ∊ ℤ}

In [10]: 2 in S1
Out[10]: False

In [11]: 2 in S1.intersect(Reals)
Out[11]: True
```
What is actually being returned is:
```python
In [12]: S1.intersect(S2)
Out[12]: ℤ \ {(-1, 1)}
```
That's the set of integers minus the set containing a single tuple `(-1, 1)`. Something has clearly gone wrong there. The previous answer was correct. I think it's fine if this intersection doesn't evaluate but it shouldn't return a wrong result like this.

shashankks0987: Ok I'll try fixing this

shashankks0987:
@oscarbenjamin Does this work? (I'm assuming other cases are tested)
```diff
--- a/sympy/sets/handlers/intersection.py
+++ b/sympy/sets/handlers/intersection.py
@@ -306,7 +306,7 @@ def intersection_sets(self, other): # noqa:F811
             # univarite imaginary part in same variable
             x, xis = zip(*[solve_linear(i, 0) for i in Mul.make_args(im) if n in i.free_symbols])
             if x and all(i == n for i in x):
-                base_set -= FiniteSet(xis)
+                base_set = FiniteSet(*xis)
             else:
                 base_set -= ConditionSet(n, Eq(im, 0), S.Integers)
         # exclude values that make denominators 0
diff --git a/sympy/sets/tests/test_fancysets.py b/sympy/sets/tests/test_fancysets.py
index 098020ca4e..dd67662f01 100644
--- a/sympy/sets/tests/test_fancysets.py
+++ b/sympy/sets/tests/test_fancysets.py
@@ -657,7 +657,7 @@ def test_infinitely_indexed_set_2():
 def test_imageset_intersect_real():
     from sympy import I
     from sympy.abc import n
-    assert imageset(Lambda(n, n + (n - 1)*(n + 1)*I), S.Integers).intersect(S.Reals) == Complement(S.Integers, FiniteSet((-1, 1)))
+    assert imageset(Lambda(n, n + (n - 1)*(n + 1)*I), S.Integers).intersect(S.Reals) == FiniteSet(-1, 1)
```

sympy/sets/tests/test_fancysets.py
shashankks0987:
```
In [4]: S1
Out[4]: {n + ⅈ⋅(n - 1)⋅(n + 1) │ n ∊ ℤ}

In [5]: 2 in S1
Out[5]: False

In [6]: 2 in S1.intersect(Reals)
Out[6]: False

In [7]: S2 = Reals

In [8]: S1.intersect(S2)
Out[8]: {-1, 1}
```
This seems to be working fine though

shashankks0987: With the new diff @oscarbenjamin

oscarbenjamin:
> I'm really not sure this is right though, I feel it's a hack ><

Why does it seem like a hack?

The code as merged definitely has a bug and what you've shown at least fixes it for the given test case.

shashankks0987:
> > I'm really not sure this is right though, I feel it's a hack ><
> 
> Why does it seem like a hack?
> 
> The code as merged definitely has a bug and what you've shown at least fixes it for the given test case.

I feel so because we're assigning the unpacked values to the solution set rather than subtracting/adding from it,
Anyways I'll raise a PR for now as it is still better than the merged code

oscarbenjamin:
Yes, that doesn't seem write. I think it should be `FiniteSet(*xis)` in any case but maybe something else is wrong.

PR #21581, opened 2021-06-06
feat(polys): add EXRAW domain

oscarbenjamin:

Adds a new domain like called EXRAW. This domain is like EX except that it uses Expr directly as the elements and has no automatic cancellation. The intention is that this would be a suitable domain for DomainMatrix to use when providing the internal representation of Matrix.

#### Release Notes

NO ENTRY


sylee957:
Is this ready to merge?

oscarbenjamin:
> Is this ready to merge?

I think so but then there isn't much benefit to this yet. It is needed so that DomainMatrix can handle Expr in order to replace the internals of Matrix. We could just leave this until a PR can be built on top that actually uses this to replace the internals of Matrix.

PR #21588, opened 2021-06-08
refactor dep-div-indep hint

Mohitbalwani26:

SEE #18348

#### Release Notes

NO ENTRY

### DIFF
--- a/doc/src/modules/solvers/ode.rst
+++ b/doc/src/modules/solvers/ode.rst
@@ -90,7 +90,7 @@ factorable
 
 1st_homogeneous_coeff_subs_indep_div_dep
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-.. autofunction:: sympy.solvers.ode.single::HomogeneousCoeffSubsIndepDivDep
+.. autoclass:: sympy.solvers.ode.single::HomogeneousCoeffSubsIndepDivDep
    :members:
 
 1st_linear
--- a/doc/src/modules/solvers/ode.rst
+++ b/doc/src/modules/solvers/ode.rst
@@ -85,7 +85,8 @@ factorable
 
 1st_homogeneous_coeff_subs_dep_div_indep
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-.. autofunction:: sympy.solvers.ode.ode::ode_1st_homogeneous_coeff_subs_dep_div_indep
+.. autoclass:: sympy.solvers.ode.single::HomogeneousCoeffSubsDepDivIndep
+   :members:
 
 1st_homogeneous_coeff_subs_indep_div_dep
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/doc/src/modules/solvers/ode.rst
+++ b/doc/src/modules/solvers/ode.rst
@@ -81,7 +81,7 @@ factorable
 
 1st_homogeneous_coeff_best
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
-.. autofunction:: sympy.solvers.ode.single::HomogeneousCoeffBest
+.. autoclass:: sympy.solvers.ode.single::HomogeneousCoeffBest
 
 1st_homogeneous_coeff_subs_dep_div_indep
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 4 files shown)


oscarbenjamin:
> `ode_1st_homogeneous_coeff_subs_dep_div_indep` is still there in dsolve as it is used by hint `ode_1st_homogeneous_coeff_best`

It's probably better just to refactor both at the same time.

Mohitbalwani26:
> It's probably better just to refactor both at the same time.

yes, I am on it. But for better code reviews I am making different PRs. 
Also I was thinking `ode_1st_homogeneous_coeff_subs_indep_div_dep` to be refactored in same way like this and  `1st_homogeneous_coeff_best` will be a separate class which will inherit these two classes as it just return best solution from these two.
currently `ode_linear_coefficients` also calls `ode_1st_homogeneous_coeff_best` so should it inherit `1st_homogeneous_coeff_best`?

oscarbenjamin:
> yes, I am on it. But for better code reviews I am making different PRs.

It's actually easier to review if you do all the related solvers at once so I can see what the code looks like altogether. The more important part of this is not just moving the solvers into classes but thinking about how to reorganise them afterwards.

> Also I was thinking `ode_1st_homogeneous_coeff_subs_indep_div_dep` to be refactored in same way like this and `1st_homogeneous_coeff_best` will be a separate class which will inherit these two classes as it just return best solution from these two.

Can these not just all be combined? I don't think we need all these similar but slightly different methods.

Mohitbalwani26:
@oscarbenjamin should we keep `_linear_coeff_match` as public in ode.py or in class as private function?

oscarbenjamin:
> @oscarbenjamin should we keep `_linear_coeff_match` as public in ode.py or in class as private function?

Leading underscore names are not public. This should be moved to the class along with the rest of the code for this solver.

The ODE docs list which are public "user" functions:
https://docs.sympy.org/latest/modules/solvers/ode.html

Mohitbalwani26:
Screenshot from 2021-06-09 15-48-56
A: 1st_homogeneous_coeff_subs_dep_div_indep
B: 1st_homogeneous_coeff_subs_indep_div_dep
C: 1st_homogeneous_coeff_best
D: linear_coefficients

This is just a gist for this PR and indicates the flow of dsolve. commenting here for later reference.

Mohitbalwani26:
@oscarbenjamin can you review this?

doc/src/modules/solvers/ode.rst
```diff
@@ -152,7 +154,7 @@ almost_linear
 
 linear_coefficients
 ^^^^^^^^^^^^^^^^^^^
-.. autofunction:: sympy.solvers.ode.ode::ode_linear_coefficients
+.. autofunction:: sympy.solvers.ode.single::LinearCoefficients
```
oscarbenjamin, line 157: Should this be `autoclass`?

Mohitbalwani26: yes, this should be autoclass. will rectify it

doc/src/modules/solvers/ode.rst
```diff
@@ -81,15 +81,17 @@ factorable
...
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-.. autofunction:: sympy.solvers.ode.ode::ode_1st_homogeneous_coeff_subs_dep_div_indep
+.. autoclass:: sympy.solvers.ode.single::HomogeneousCoeffSubsDepDivIndep
+   :members:
 
 1st_homogeneous_coeff_subs_indep_div_dep
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-.. autofunction:: sympy.solvers.ode.ode::ode_1st_homogeneous_coeff_subs_indep_div_dep
+.. autoclass:: sympy.solvers.ode.single::HomogeneousCoeffSubsIndepDivDep
+   :members:
```
oscarbenjamin, line 94: Why do some autoclass blocks have members but others not?

Is it useful to show the members?

Mohitbalwani26: official documentation says if we keep the :members: block then it will include docstrings of member functions also. but in previous class like Factorable, initially it was not there but someone made commit stating 'expanding documentation', so I thought it is the practice to be followed.

sympy/solvers/ode/single.py
```diff
@@ -1310,6 +1309,501 @@ def _get_match_object(self):
         m2 = {self.y: ycoeff, x: 1, 'coeff': 1}
         return m1, m2, x, x**self.r2['power']*fx
 
+class HomogeneousCoeffSubsDepDivIndep(SinglePatternODESolver):
```
oscarbenjamin, line 1312: There should be two blank lines between top-level classes/functions

oscarbenjamin:
I think that this can definitely be cleaned up later. Really we can simplify all of this logic to just something like:

1. Check the basic pattern matches `dy/dt = f(y, t)`
2. Try the substitution `x = t*y` and/or `x = t/y` and see if the result is separable.
3. Solve with the separable solver.

I think that the degenerate case where `d + e*u = 0` actually just corresponds to the case where the substitution reduces the ODE to `dx/dt = 0` which is easily solvable. If the substitution was computed explicitly then it could be handled as well.

The code for matching `linear_coefficients` can be a lot simpler as well.

oscarbenjamin:
For now this can be merged though so the refactoring can continue.

PR #21590, opened 2021-06-08
Changed the sympy.stats.sample() API and updated examples

### LINKED ISSUE #21563: sympy.stats.sample() API change in 1.7 breaks code and is confusing
Spinning off new issue from reemerging discussion on closed issue #19061

`sympy.stats.sample()` has changed behaviour since v1.7, returning an iterator instead of actual sample(s), which is an issue:

- It breaks existing code by breaking the API (now returns an iterator rather than samples)
- The change was AFAIK made without warning about future API change
- The iterator serves no real purpose, as you can only use `next()` on it once
- If you want an iterator (one which works), you can already do that with `sample_iterator()`
- The new behaviour is confusing, non-intuitive and not "pythonic"

Below is a code example that shows the warning message that is triggered, how `sample()` returns an iterator, and how `next()` can only be used once.

```
>>> import sympy.stats as st
>>> X = st.Normal('X', 100, 20)
>>> s = st.sample(X)
C:\Users\analyzethem\Anaconda3\lib\site-packages\sympy\stats\rv.py:1126: UserWarning:
The return type of sample has been changed to return an iterator
object since version 1.7. For more information see
https://github.com/sympy/sympy/issues/19061
  warnings.warn(filldedent(message))
>>> s
<generator object sample_iter.<locals>.return_generator_finite at 0x00000295F020E660>
>>> next(s)
102.97917700376605
>>> next(s)
StopIteration
```

Personally I discovered this issue as a user, having some code which suddenly broke, and following the link to the issue in the warning. It was a bit difficult figuring out what was going on, help text did not really enlighten me too much, and learning that the iterator can only iterate once was doubly confusing.

If you want a random sample from a stochastic variable `X`, then having to do something like `next(sample(X))` to obtain a single sample is really nonintuitive. Also, if you do e.g. `sample(X, size=100)` and the return value is an iterator, you would really think you can iterate over it to get the 100 values - not that you do `next()` once and that then returns a list of 100 values. This is a bit confusing, and so not the pythonic way.

See #19061 for some comments on the topic. My $0.10 how to deal with it:

- There is a problem to keep the new behaviour (breaks old code for no good reason), or just changing it back (breaks new code)
- Since either option is bad, pick the one that is better long-term (return to the behaviour which makes the most sense, which is the previous one) - but do it from the next release, with a warning
- Put in some interim solution, e.g. adding `sympy.stats.future_sampling()` to change global behaviour for `sample()` to no longer return an iterator (this enables fixing old code by adding one line after the `sympy.stats` import`
- If looking into this piece of the API anyways, it might make sense to also introduce a function `samples()` for requesting more than one sample, so instead of `sample(size=100)` you could do `samples(100)` (cleaner, and without the need for keywords)
- And ... if going "full python", why not add behaviour to stochastic variable objects so that for a stochastic variable `X` you could do `iter(X)` and get a random sample iterator

If there is an issue with a new iterator being created any time `sample()` is called (which seems to be the problem one tried to deal with for #19061), one way would be to deal with that e.g. by having a function `sampler()` which returns an object which is iterable (so from that perspective same behaviour as return value of `sample_iterator()`, but which also has methods for pulling sample(s). So you could do e.g.

```
import sympy.stats as sp
X = st.Normal('X', 100, 20)
s = st.sampler(X)
print('Single sample:', s.sample())
print('Ten samples:', s.samples(10))
print('Ten samples (by iterating):', [next(s) for i in range(10)])
```

By reusing the sampler object for generating samples, you avoid the overhead of creating new sample iterators. You could even deprecate `sample()` for a future major release, in order to use the new `sampler()` feature instead - but `sample()` could continue to exist as a shortcut for `sampler().sample()`

mostlyaman:
Fixes #21563 

`sympy.stats.sample()` has changed behaviour since v1.7, returning an iterator instead of actual sample(s), which is an issue. It is now reverted to returning samples and documentation has been updated.

#### Release Notes

* stats
  * BACKWARDS COMPATIBILITY: `sample` now returns samples as float and arrays instead of an iterator. This restores and extends the behavior from SymPy 1.6 which was changed to an iterator in 1.7. Use `sample_iter` to get an iterator consistently in all versions of SymPy.
  * DEPRECATION: `numsamples` parameter in `sample` function is now deprecated. The parameter was added in SymPy 1.7 and continues to work but will emit a DeprecationWarning. The parameter will be removed in a future release.
  * PyMC3 library implementation in `sample` function has been fixed.


mostlyaman:
I believe tests for `sympy.stats.rv.py` also need to be updated for the new API. Please check the new examples and documentation. Also, I have commented out the warning for now, but I believe it also should be removed (line 1123 to 1127).

sympy/stats/rv.py
```diff
@@ -1083,50 +1083,79 @@ def sample(expr, condition=None, size=(), library='scipy',
     >>> X, Y, Z = Die('X', 6), Die('Y', 6), Die('Z', 6) # Finite Random Variable
 
     >>> die_roll = sample(X + Y + Z) # doctest: +SKIP
-    >>> next(die_roll) # doctest: +SKIP
+    >>> die_roll # doctest: +SKIP
```
oscarbenjamin, line 1086: Can the skip be removed now?

mostlyaman:
```python
File "/home/aman/sympy/sympy/stats/rv.py", line 1086, in sympy.stats.rv.sample
Failed example:
    die_roll
Expected:
    6
Got:
    7
```
I dont think so that they can be removed since the output is random.

oscarbenjamin: Maybe set the seed at the start of the docstring?

mostlyaman: Please guide me with that. Do you mean specifying a seed in every example like `die_roll = sample(X+Y+Z, seed = 12)`. That would indeed output a consistent result if we specify seed in every example.

oscarbenjamin: Is there not a way to set the seed just once and have it used in subsequent calls to sample?

oscarbenjamin: I guess this is why it would be good to have a `sampler` class: it could store the seed and then you could do:
```python
state = sampler('numpy', seed)
state.sample(X + Y + Z)
state.sample(X)
```
For now I guess let's just leave the `+SKIP` there.

mostlyaman: Apart from this docstring, where else a change in documentation is required?

mostlyaman:
Do we need tests for checking that `sample` returns the appropriate data type from now on?  Like a float or a list instead of a iterator and reference this issue.
Because such test are absent in `test_rv.py`.

oscarbenjamin:
> Do we need tests for checking that `sample` returns the appropriate data type from now on? Like a float or a list instead of a iterator and reference this issue.

That seems like a good idea.

mostlyaman:
```python
____________________________ sympy.stats.rv.sample _____________________________
File "/home/runner/work/sympy/sympy/sympy/stats/rv.py", line 1084, in sympy.stats.rv.sample
Failed example:
    die_roll = sample(X + Y + Z)
Exception raised:
      File "/opt/hostedtoolcache/Python/3.9.5/x64/lib/python3.9/doctest.py", line 1336, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest sympy.stats.rv.sample[2]>", line 1, in <module>
        die_roll = sample(X + Y + Z)
      File "/home/runner/work/sympy/sympy/sympy/stats/rv.py", line 1146, in sample
        iterator = sample_iter(expr, condition, size=size, library=library,
      File "/home/runner/work/sympy/sympy/sympy/stats/rv.py", line 1264, in sample_iter
        raise ValueError("Failed to import %s" % library)
    ValueError: Failed to import scipy
```
The tests are unable to import scipy, it seems like. Is there a solution to this?
As a temporary fix, I could skip all the doctest in sample.

oscarbenjamin:
The `doctest_depends_on` decorator can be used for this e.g.:
https://github.com/sympy/sympy/blob/110997fe18b9f7d5ba7d22f624d156a29bf40759/sympy/matrices/dense.py#L803-L804

mostlyaman:
The new `sample` API has been implemented and all necessary changes have been made. 
1. Is there any more documentation that need to be updated other than the docstring in `rv.py`
2. Should we consider adding `class sampler`. A discussion might be needed if it is considered.

sympy/stats/rv.py
```diff
@@ -1081,52 +1083,80 @@ def sample(expr, condition=None, size=(), library='scipy',
...
-    sample: iterator object
-        iterator object containing the sample/samples of given expr
+    sample: float/list/numpy.ndarray
+        one sample or a collection of samples of the random expression
+
+        sample(X) --> returns float object
+        sample(X, size=int) or sample(X, numsamples=int) --> returns list object
+        sample(X, size=tuple) --> returns numpy.ndarray object
+        sample(X, size=tuple, numsamples=int) --> returns list containing numpy.ndarray objects
+        sample(X, size=int, numsamples=int) --> returns list containing numpy.ndarray objects
```
oscarbenjamin, line 1143: I don't think that we need all of these combinations. I would just go with the first 3 and not have the `numsamples` option. Perhaps the `numsamples` option could be provided but only so that it is compatible with sympy 1.8. The docstring could list it as deprecated.

Also is a numpy array always returned? What happens if pymc3 is used?

oscarbenjamin:
The docstring should use sphinx' `.. versionchanged` to explain what was changed in different sympy versions e.g. what was changed in 1.7 and then what was changed in 1.9 (i.e. this PR).

sympy/stats/rv.py
mostlyaman: I am trying to fix pymc3 implementation in sample. It seems like `pymc3.sample` takes the size parameter (which is called draws in pymc3) only in `int`. And the sympy implementation transfer the size and numsamples parameters as a tuple to draws which causes a TypeError. Here I have printed them to confirm.
```python
next(x)
size =  (1, 2)
/home/aman/sympy/sympy/stats/rv.py:1574: FutureWarning: In v4.0, pm.sample will return an `arviz.InferenceData` object instead of a `MultiTrace` by default. You can pass return_inferencedata=True or return_inferencedata=False to be safe and silence this warning.
  samps = pymc3.sample(draws=size, chains=1, progressbar=False, random_seed=seed)[:]['X']
```
My solution would be to set a for loop to set up a output in case of a tuple input to size. Also since numsamples does not exist in pymc3, for loop is need to set that up also.

oscarbenjamin: I think that with pymc3 the size parameter should always be passed as an int. Only for numpy/scipy should it be possible to pass a tuple.

mostlyaman: I thought about that too, but also some users might find the size as tuple functionality useful. Possibly a very small amount of users. For code simplicity, it is best to pass a TypeError in case of a tuple.

mostlyaman:
I believe I have fixed pymc3 implementation in sample. I have added a TypeError in case of size = tuple.
```python
>>> from sympy.stats import *
>>> x = Normal('', 10, 2)
>>> sample(x, library='pymc3')
Only 1 samples in chain.
Auto-assigning NUTS sampler...
Initializing NUTS using jitter+adapt_diag...
Sequential sampling (1 chains in 1 job)
NUTS: [X]
Sampling 1 chain for 1_000 tune and 1 draw iterations (1_000 + 1 draws total) took 2 seconds.
12.72763422413447

>>> sample(x, library='pymc3', size=2)
Only 2 samples in chain.
Auto-assigning NUTS sampler...
Initializing NUTS using jitter+adapt_diag...
Sequential sampling (1 chains in 1 job)
NUTS: [X]
Sampling 1 chain for 1_000 tune and 2 draw iterations (1_000 + 2 draws total) took 2 seconds.
array([10.43483746, 10.37401032])

>>> sample(x, library='pymc3', size=(2,3))
  File "/home/aman/sympy/sympy/stats/rv.py", line 1155, in sample
    return next(iterator)
  File "/home/aman/sympy/sympy/stats/rv.py", line 1329, in return_generator_finite
    d = ps.sample(size=(numsamples,) + ((size,) if isinstance(size, int) else size),
  File "/home/aman/sympy/sympy/stats/crv.py", line 478, in sample
    return {self.value: self.distribution.sample(size, library=library, seed=seed)}
  File "/home/aman/sympy/sympy/stats/rv.py", line 1610, in sample
    raise TypeError("Invalid value for `size` in pymc3. Must be int")
TypeError: Invalid value for `size` in pymc3. Must be int
```
Unfortunately, numsamples does not work yet with pymc3. 
I wanted to ask that since numsamples is going to be deprecated after v1.9 and pymc3 does not work before v1.9, is it beneficial to add code to make pymc3 work with numsamples?

oscarbenjamin:
> I wanted to ask that since numsamples is suggested to be deprecated after v1.9 and pymc3 does not work before v1.9, is it beneficial to add code to make pymc3 work with numsamples?

If it doesn't already work and has never worked then we don't need to make it work now.

mostlyaman:
Couple of changes:
1. numsamples has been deprecated. I believe I have updated every test with the change.
2. Pymc3 implementation works now. It took most of the time and ran into bugs with different random variables but now I have tested it to work.
3.sphinx version changes are added. I did not test it, I am new to sphinx and I just looked at some examples and added it in the docstring. Please check it.

sympy/stats/rv.py
```diff
@@ -1076,57 +1079,96 @@ def sample(expr, condition=None, size=(), library='scipy',
...
+    >>> samp > 0
     True
-    >>> samp_list = next(sample(N, size=4)) # doctest: +SKIP
-    >>> [sam in N.pspace.domain.set for sam in samp_list] # doctest: +SKIP
+    >>> samp_list = sample(N, size=4)
+    >>> [sam in N.pspace.domain.set for sam in samp_list]
     [True, True, True, True]
+
+    >>> sample(N, numsamples=3) # doctest: +SKIP
+    [6.07745952932677, 8.024735984727329, 3.7581166319235315]
```
oscarbenjamin, line 1109: If `numsamples` is deprecated then examples like this should go in a section that clearly explains that it is deprecated.

sympy/stats/rv.py
```diff
@@ -1076,57 +1079,96 @@ def sample(expr, condition=None, size=(), library='scipy',
...
+                 deprecated_since_version="1.9").warn()
+
+        if library == 'pymc3':
+            raise NotImplementedError("numsamples is not currently implemented for pymc3")
+        return [next(iterator) for i in range(numsamples)]
+
+    if library == 'pymc3' and size == ():
+        return next(iterator)[0]
+
+    return next(iterator)
```
oscarbenjamin, line 1169: Why does pymc3 need to be special cased here?

Can that not just be handled in the pymc3 code?

Does `sample_iter` with `size=()` return different results for pymc3 than the other implementations?

sympy/stats/rv.py
```diff
@@ -1565,12 +1608,24 @@ def sample(self, size=(), library='scipy', seed=None):
...
+            #size accepts only int in case of pymc3
+            #see https://github.com/sympy/sympy/pull/21590#discussion_r649783860
+
+            size = list(size)[1:]
+            if size == []:
+                size=1
+            elif len(size) == 1:
+                size = size[0]
+            else:
+                raise TypeError("Invalid value for `size` in pymc3. Must be int")
```
oscarbenjamin, line 1624: This block of code appears in at least 3 places. Is there not a better place to add this?

sympy/stats/sampling/tests/test_sample_continuous_rv.py
```diff
@@ -23,11 +23,11 @@ def test_sample_numpy():
     else:
         with ignore_warnings(UserWarning): ### TODO: Restore tests once warnings are removed
```
oscarbenjamin, line 24: Is the `UserWarning` still there?

sympy/stats/tests/test_rv.py
```diff
@@ -210,6 +209,12 @@ def test_Sample():
...
     assert all(i in range(4, 7) for i in density(X, X>3, numsamples=10))
 
+    #Test Issue #21563: Output of sample must be a float or list
+    assert str(type(sample(X))) == "<class 'numpy.int64'>"
+    assert str(type(sample(Y))) == "<class 'numpy.float64'>"
+    assert str(type(sample(X, size = 2))) == "<class 'numpy.ndarray'>"
+
+
 
 @XFAIL
```
oscarbenjamin, line 219: There should only be two blank lines here

sympy/stats/rv.py
mostlyaman: yes, it does.
```python
>>> next(sample_iter(Normal('',10,2), numsamples=1))
8.6219572235293
>>> next(sample_iter(Normal('',10,2), numsamples=1, library='pymc3'))
array([6.29817056])
```
>Can that not just be handled in the pymc3 code?

Regarding that, I notice the the code for producing samples for pymc3 is different from other libraries. While numpy and scipy use the `do_sample_numpy` and `do_sample_scipy` from sampling directory, pymc3 uses its sample function. 

Why was `do_sample_pymc3` not used although it was made? I will be trying to implement it for now.

sympy/stats/rv.py
mostlyaman: There may be a way to add it only in `sample_iter`. But that causes problems in matrix_distributions and joint_rv in this part. Currently, I think that this is the best possible option for now to prevent break working code.
https://github.com/sympy/sympy/blob/a8299e5d3a4309946439dd440421c2a8d7b91ae9/sympy/stats/joint_rv.py#L239-L240

mostlyaman:
I have applied all the suggestions except for 
>This block of code appears in at least 3 places. Is there not a better place to add this?

The new `sample` works as follows:
```scipy size () numsamples 1 result 13.155547152178015
scipy size () numsamples 2 result [12.334068273823242, 6.855375164803271]
scipy size 2 numsamples 1 result [11.3188489   6.46867553]
scipy size 2 numsamples 2 result [array([11.71726809, 10.32243694]), array([10.38108636, 10.38670564])]
scipy size (2, 3) numsamples 1 result [[13.19261421 11.85576395  9.2925102 ]
 [11.35210882  9.54048243  9.10832117]]
scipy size (2, 3) numsamples 2 result [array([[ 7.41443115,  8.40133242, 12.00056096],
       [10.51686578, 12.80475091,  8.13475507]]), array([[ 7.76038915, 11.52534808, 10.16221015],
       [ 9.17799292,  8.45351092,  8.08677963]])]

numpy size () numsamples 1 result 13.468266439580425
numpy size () numsamples 2 result [11.998689835328879, 8.48320379054535]
numpy size 2 numsamples 1 result [12.55275265 10.36745033]
numpy size 2 numsamples 2 result [array([9.23051241, 5.85401366]), array([13.1803515 , 12.58579517])]
numpy size (2, 3) numsamples 1 result [[ 8.39004423  9.8466101  11.06378944]
 [ 9.08373437 10.7414229   8.71392445]]
numpy size (2, 3) numsamples 2 result [array([[10.51265292,  3.09621378,  5.26667608],
       [10.53263655, 10.37663265, 15.35528054]]), array([[14.73952878, 12.99776003, 10.09665147],
       [ 7.40962766,  9.63214763, 15.15499201]])]

pymc3 size () numsamples 1 result 8.173775645457082
pymc3 size () numsamples 2 result NotImplementedError
pymc3 size 2 numsamples 1 result [11.56452123  9.38950679]
pymc3 size 2 numsamples 2 result NotImplementedError
pymc3 size (2, 3) numsamples 1 result TypeError
pymc3 size (2, 3) numsamples 2 result NotImplementedError
```

oscarbenjamin:
@czgdp1807 @Upabjojr what do you think of this?

sympy/stats/tests/test_rv.py
```diff
@@ -210,6 +209,12 @@ def test_Sample():
     assert all(i in range(1, 7) for i in density(X, numsamples=10))
     assert all(i in range(4, 7) for i in density(X, X>3, numsamples=10))
 
+    #Test Issue #21563: Output of sample must be a float or list
+    assert str(type(sample(X))) == "<class 'numpy.int64'>"
+    assert str(type(sample(Y))) == "<class 'numpy.float64'>"
+    assert str(type(sample(X, size = 2))) == "<class 'numpy.ndarray'>"
+
```
czgdp1807, line 216: Use, `isinstance`.

czgdp1807:
So, the change here is that instead of iterator, now `sample` returns arrays/list/float, etc? If that's the case then I am fine with it. Though, what if in future we include tensorflow (or for that matter, any other library) as well then will we convert tensor objects obtained from tensorflow to numpy array and then return it?

oscarbenjamin:
> Though, what if in future we include tensorflow (or for that matter, any other library) as well then will we convert tensor objects obtained from tensorflow to numpy array and then return it?

If in future tensorflow is added then it can return tensorflow arrays when the tensorflow module is used. I think for each library `sample` can return something in the format that the given library would most efficiently return.

mostlyaman:
As mentioned earlier, 
>    There is one problem though. Since PyMC3 prints a bunch of lines while producing the samples, it clutters the output when running tests. Is there a way to disable printing for the library?

Does this also need to be fixed?

oscarbenjamin:
If possible then it should be fixed I think.

I just tested this PR and I got:
```python
In [7]: sample(X, library='pymc3', seed=2)
/home/oscar/current/sympy/sympy.git/sympy/stats/rv.py:1610: FutureWarning: In v4.0, pm.sample will return an `arviz.InferenceData` object instead of a `MultiTrace` by default. You can pass return_inferencedata=True or return_inferencedata=False to be safe and silence this warning.
  progressbar=False, random_seed=seed)[:]['X']]
Only 1 samples in chain.
Auto-assigning NUTS sampler...
Initializing NUTS using jitter+adapt_diag...
Sequential sampling (1 chains in 1 job)
NUTS: [X]
Sampling 1 chain for 1_000 tune and 1 draw iterations (1_000 + 1 draws total) took 1 seconds.
/home/oscar/current/sympy/37venv/lib/python3.7/site-packages/pymc3/sampling.py:643: UserWarning: The number of samples is too small to check convergence reliably.
  warnings.warn("The number of samples is too small to check convergence reliably.")
Out[7]: 1.6129789130217302
```
It looks like at least the first part of the warning can be disabled with `return_inferencedata`.

Also it seems that `sample_iter` doesn't work:
```python
In [4]: next(sample_iter(X, library='pymc3'))
WARNING (theano.tensor.blas): Using NumPy C-API based implementation for BLAS functions.
  File "<ipython-input-4-eb222e895ddf>", line 1, in <module>
    next(sample_iter(X, library='pymc3'))
  File "/home/oscar/current/sympy/sympy.git/sympy/stats/rv.py", line 1303, in return_generator_infinite
    d = ps.sample(size=size, library=library, seed=rand_state)  # a dictionary that maps RVs to values
  File "/home/oscar/current/sympy/sympy.git/sympy/stats/crv.py", line 478, in sample
    return {self.value: self.distribution.sample(size, library=library, seed=seed)}
  File "/home/oscar/current/sympy/sympy.git/sympy/stats/rv.py", line 1610, in sample
    progressbar=False, random_seed=seed)[:]['X']]
  File "/home/oscar/current/sympy/37venv/lib/python3.7/site-packages/pymc3/sampling.py", line 453, in sample
    raise TypeError("Invalid value for `random_seed`. Must be tuple, list or int")
TypeError: Invalid value for `random_seed`. Must be tuple, list or int
```

czgdp1807:
> If in future tensorflow is added then it can return tensorflow arrays when the tensorflow module is used. I think for each library `sample` can return something in the format that the given library would most efficiently return.

Fine then, I am okay with this PR. @Upabjojr Any thoughts on this work? Thanks.

sympy/stats/rv.py
```diff
@@ -1300,11 +1324,11 @@ def return_generator_finite():
         while faulty:
             d = ps.sample(size=(numsamples,) + ((size,) if isinstance(size, int) else size),
                           library=library, seed=seed) # a dictionary that maps RVs to values
+
```
czgdp1807, line 1327: Remove unnecessary new lines. Please make a separate PR for these kind of changes.

mostlyaman:
New changes:
1. `sample` and `sample_iter` now produce appropriate output.

`sample` produces the output as follows with any library:
size=() numsamples=1 --> int/float
size=() numsamples=int --> list containing int/float
size=int/tuple numsamples=1 --> numpy array
size=int/tuple numsamples=int --> list containing numpy array
This is with numsamples being deprecated.
I may add that I have tested only using a Normal RV. I will test with others soon.

2. Excess Printing is minimised while using PyMC3. 

Now, there is only a one time warning when pymc3 is imported. Other log printing is overridden using logging library.

3. numsamples has been implemented for PyMC3.
Although deprecated, I thought it was easy enough to implement and provided consistency with all the libraries.

sympy/stats/joint_rv.py
```diff
@@ -233,9 +234,22 @@ def _sample_pymc3(cls, dist, size, seed):
...
+        if draws == []:
+            draws=1
+        elif len(draws) == 1:
+            draws = draws[0]
+        else:
+            raise TypeError("Invalid value for `size` in pymc3. Must be int")
+        import logging
+        logging.getLogger("pymc3").setLevel(logging.ERROR)
+        with ignore_warnings(UserWarning):
+            with pymc3.Model():
```
oscarbenjamin, line 250: Is there a way to tell pymc3 not to emit the warning rather than using `ignore_warnings` here?

mostlyaman: I guess there's a way. If we set compute_convergence_checks=False, then it will not emit the UserWarning regarding insufficient no of samples for checks. `ignore_warnings` is used to achieve the same effect.

oscarbenjamin: It would be better to use that rather than `ignore_warnings` because it is still useful to be able to see warnings even if there are some that we know we want to disable.

mostlyaman: Yes, I understand.

mostlyaman:
@oscarbenjamin I have tested the changes quite a lot and I think its almost ready to merge. If there's any more documentation or tests to be added, please suggest.
There's one thing i found. `RandomMatrixSymbol`s produces a weird numpy array where each sample is in a `(1,2)` shape instead of a normal `(2,)` shape. It produces the same result in current master. I do not know if its intended to be in a (1,2) instead of (2,).
```python
>>> M = MatrixNormal('M', [[1, 2]], [1], [[1, 0], [0, 1]])
>>> sample(M)
array([[1.83372238, 2.73327529]])
>>> sample(M).shape
(1, 2)
>>> sample(M, size=2)
array([[[1.68397685, 1.4381359 ]],
       [[1.37492516, 0.93570061]]])
```

oscarbenjamin:
> It produces the same result in current master. I do not know if its intended to be in a (1,2) instead of (2,).

I'm not sure. @czgdp1807 do you know what this should be?

Current master:
```python
In [28]: M = MatrixNormal('M', [[1, 2]], [1], [[1, 0], [0, 1]])

In [29]: next(sample(M)).shape
/Users/enojb/current/sympy/sympy/sympy/stats/rv.py:1126: UserWarning: 
The return type of sample has been changed to return an iterator
object since version 1.7. For more information see
https://github.com/sympy/sympy/issues/19061
  warnings.warn(filldedent(message))
Out[29]: (1, 2)

In [30]: next(sample(M, size=2)).shape
/Users/enojb/current/sympy/sympy/sympy/stats/rv.py:1126: UserWarning: 
The return type of sample has been changed to return an iterator
object since version 1.7. For more information see
https://github.com/sympy/sympy/issues/19061
  warnings.warn(filldedent(message))
Out[30]: (2, 1, 2)

In [31]: next(sample_iter(M))
IndexError
~/current/sympy/sympy/sympy/stats/matrix_distributions.py in _sample_scipy(cls, dist, size, seed)
     95         else:
     96             rand_state = seed
---> 97         for _ in range(size[0]):
     98             samp = scipy_rv_map[dist.__class__.__name__](dist, size[1] if len(size) > 1 else 1, rand_state)
     99             samples.append(samp)

IndexError: tuple index out of range
```

czgdp1807:
Probably the default value of `size` is an empty tuple. I think the root cause can be found out by a `pdb` session. 

The other thing about (1, 2) or (2,). `M = MatrixNormal('M', [[1, 2]], [1], [[1, 0], [0, 1]])` specifies that the `x`(sampled values) will be of `[[1, 2]]` dimension (see the second argument). Try giving, `[2, 1]` (or `[[2, 1]]`) and if it still produces `[[1, 2]]` then it would be a bug needing a fix.

mostlyaman:
Yes, I know what is causing this! Its about the way sample_iter interfaces with `pspace(X).distribution.sample` function. 
Currently, sample_iter can use `return_generator_finite` and `return_generator_infinte` and both generators interface with the same `pspace(x).distribution.sample` function but in a different way.

In `return_generator_infinte`, it passes the size param as entered by the user. 
```python
d = ps.sample(size=_size, library=library, seed=rand_state)
```
whereas in `return_generator_finite`, it passes the size param as a tuple combined with numsamples.
```python
d = ps.sample(size=(numsamples,) + ((size,) if isinstance(size, int) else size), library=library, seed=seed)
```
Just like in the example above, size is passed as 2, which causes an error as the function is expecting tuple.
i suggest combining both generators into a single one, and passing numsamples as a separate parameter to distribution.sample, although that mean be a lot of code rewritten.

czgdp1807:
May be doing it in a follow up PR would be better.

mostlyaman:
There are other things which make the generators interface differently with distribution.sample
`return_generator_infinte` sends a numpy generator derived from the seed the user entered whereas `return_generator_finite` sends the seed entered by the user as it is.

Considering both generators interface with the same function, so many differences should not be present as it causes bugs. Like for eg: the seed difference mentioned above caused an error where `return_generator_infinite` would fail when using pymc3 because pymc3 would not use a numpy generator as seed. I had to use an if-condition to solve that.
```python
        if np:
            rand_state = np.random.default_rng(seed=seed) if library != "pymc3" else seed
        else:
            rand_state = None
```
If we just had a single generator, such bugs would not arise or at the very least make them have identical behaviour. I would make a follow-up PR summarising the problem.
For now, I will try to fix it.

oscarbenjamin:
> I would make a follow-up PR summarising the problem.
> For now, I will try to fix it.

So for now does that mean that this PR is finished?

mostlyaman:
Well, I was working on the fix and finally checking all the possibile combinations of parameters for a last time.

mostlyaman:
It works for now, although a better solution will be implemented in the next PR. Its better to push something that isnt completely broken for now.
```python
>>> M = MatrixNormal('M', [[1, 2]], [1], [[1, 0], [0, 1]])
>>> next(sample_iter(M))
array([[1.448871  , 2.09262817]])
>>> next(sample_iter(M, library='pymc3'))
array([[[-0.72141845,  1.24522181]]])
```
I think its ready to be merged.

oscarbenjamin:
Are there tests for all the cases that are fixed? I didn't see any tests in the last commit

sympy/stats/joint_rv.py
```diff
@@ -233,12 +233,24 @@ def _sample_pymc3(cls, dist, size, seed):
         if dist.__class__.__name__ not in dist_list:
             return None
 
+        #size accepts only int in case of pymc3
+        #see https://github.com/sympy/sympy/pull/21590#discussion_r649783860
+
+        draws = list(size)[1:]
+        if draws == []:
+            draws=1
```
oscarbenjamin, line 241: Space before and after `=`

sympy/stats/joint_rv.py
```diff
@@ -233,12 +233,24 @@ def _sample_pymc3(cls, dist, size, seed):
...
         with pymc3.Model():
             pymc3_rv_map[dist.__class__.__name__](dist)
-            samples = pymc3.sample(size, chains=1, progressbar=False, random_seed=seed)[:]['X']
-            if samples.shape[0:len(size)] != size:
-                samples = samples.reshape((size[0],) + samples.shape)
-            return samples
+            samples = pymc3.sample(draws=draws, chains=1, progressbar=False, random_seed=seed, return_inferencedata=False, compute_convergence_checks=False)[:]['X']
+        if samples.shape[0:len(size)] != size:
+            samples = samples.reshape((size[0],) + samples.shape)
+        return samples
```
oscarbenjamin, line 253: If we're going to call `reshape` here then we can presumably allow size to be a tuple. Can it just be something like:
```
draws = prod(size)
samples = pymc3.sample(draws=prod(size), ...)
samples = samples.reshape(size)
```

sympy/stats/rv.py
```diff
@@ -1062,7 +1065,9 @@ def sample(expr, condition=None, size=(), library='scipy',
         Choose any of the available options to sample from as string,
         by default is 'scipy'
     numsamples : int
-        Number of samples, each with size as ``size``
+        Number of samples, each with size as ``size``.
+        numsamples has been deprecated since v1.9.
+        see https://github.com/sympy/sympy/pull/21590#discussion_r649227690.
```
oscarbenjamin, line 1070: You don't need to reference the github comment here. I would say something like
```
numsamples: int
    Number of samples, each with size as ``size``. The ``numsamples`` parameter is
    deprecated and is only provided for compatibility with v1.8. Use a list comprehension
    or an additional dimension in ``size`` instead.
```

sympy/stats/rv.py
```diff
@@ -1040,6 +1042,7 @@ def where(condition, given_condition=None, **kwargs):
     return pspace(condition).where(condition, **kwargs)
 
 
+@doctest_depends_on(modules=('scipy',))
 def sample(expr, condition=None, size=(), library='scipy',
            numsamples=1, seed=None, **kwargs):
```
oscarbenjamin, line 1047: Maybe everything after `size` should be keyword-only. We don't want people use positional parameters here especially if the ``numsamples`` parameter will be removed.

sympy/stats/rv.py
```diff
@@ -1076,57 +1081,73 @@ def sample(expr, condition=None, size=(), library='scipy',
...
         are done by this argument.
 
+    Returns
+    =======
+
+    sample: float/list/numpy.ndarray
+        one sample or a collection of samples of the random expression.
+
+        - sample(X) returns float/numpy.float64/numpy.int64 object.
+        - sample(X, size=int/tuple) returns numpy.ndarray object.
```
oscarbenjamin, line 1091: Does `size=int` always return ndarray?

sympy/stats/rv.py
```diff
@@ -1076,57 +1081,73 @@ def sample(expr, condition=None, size=(), library='scipy',
...
-    >>> [tuple(sam) in MN.pspace.domain.set for sam in samp_list] # doctest: +SKIP
+    [array([2.85768055, 3.38954165]),
+     array([4.11163337, 4.3176591 ]),
+     array([0.79115232, 1.63232916]),
+     array([4.01747268, 3.96716083])]
+    >>> [tuple(sam) in MN.pspace.domain.set for sam in samp_list]
     [True, True, True, True]
 
+    .. versionchanged:: 1.7.0
+        sample returns an iterator containing the samples instead of value.
```
oscarbenjamin, line 1131: This should be in the past tense to indicate that it is no longer the case.

sympy/stats/rv.py
```diff
@@ -1076,57 +1081,73 @@ def sample(expr, condition=None, size=(), library='scipy',
...
-    return sample_iter(expr, condition, size=size, library=library,
+
+    iterator = sample_iter(expr, condition, size=size, library=library,
                                                         numsamples=numsamples, seed=seed)
 
+    if numsamples != 1:
+        SymPyDeprecationWarning(
+                 feature="numsamples parameter",
+                 issue=21563,
+                 deprecated_since_version="1.9").warn()
```
oscarbenjamin, line 1145: There should be a `use_instead` here.

sympy/stats/rv.py
```diff
@@ -1265,19 +1286,21 @@ def given_fn_subs(*args):
...
+
     if condition is not None:
         given_fn = lambdify(rvs, condition, **kwargs)
 
     def return_generator_infinite():
         count = 0
         np = import_module('numpy')
         if np:
-            rand_state = np.random.default_rng(seed=seed)
+            rand_state = np.random.default_rng(seed=seed) if library != "pymc3" else seed
```
oscarbenjamin, line 1297: Ideally we wouldn't special case a particular library here. Can the different classes not just handle this differently?

Is this what you meant about changes in a followup PR?

sympy/stats/rv.py
```diff
@@ -1265,19 +1286,21 @@ def given_fn_subs(*args):
...
         else:
             rand_state = None
+        _size = (1,)+((size,) if isinstance(size, int) else size)
```
oscarbenjamin, line 1300: Why is the `1` being added here?

sympy/stats/sampling/tests/test_sample_continuous_rv.py
```diff
@@ -174,9 +162,9 @@ def test_sample_continuous():
             imported_lib = import_module(lib)
             if imported_lib:
                 s0, s1, s2 = [], [], []
-                s0 = list(sample(Z, numsamples=10, library=lib, seed=0))
-                s1 = list(sample(Z, numsamples=10, library=lib, seed=0))
-                s2 = list(sample(Z, numsamples=10, library=lib, seed=1))
+                s0 = sample(Z, library=lib, seed=0)
+                s1 = sample(Z, library=lib, seed=0)
+                s2 = sample(Z, library=lib, seed=1)
```
oscarbenjamin, line 167: Maybe this should have `size=10`.

czgdp1807: +1. I think the changes are in the representation of outputs(earlier iterator, now a container like arrays). So, input to sampling APIs shouldn't be changed at all. They should be the same and tests should work.

czgdp1807:
I would suggest to only focus on the API right now. Avoid fixes like weird output shapes, etc. We can look into that in follow up PRs. For testing obtain the output from sampling APIs and process them instead of changing the inputs to those APIs to make tests work.

sympy/stats/joint_rv.py
mostlyaman: Fixed that.

sympy/stats/joint_rv.py
mostlyaman: Thats a great idea! It could even make something like this possible for pymc3,
```python
>>> sample(N, library='pymc3', size=(2,2,3))
array([[[11.72402844, 11.72402844, 11.72402844],
        [ 8.20910153,  8.72181852,  8.72181852]],

       [[ 8.72181852,  9.78352304,  9.78352304],
        [ 9.78352304,  9.78352304,  9.78352304]]])
```

sympy/stats/rv.py
mostlyaman: Changed.

sympy/stats/rv.py
mostlyaman: Changed it to `sample used to return an iterator`

sympy/stats/rv.py
mostlyaman: Added.

sympy/stats/sampling/tests/test_sample_continuous_rv.py
mostlyaman: I removed numsamples because it is being deprecated and set off warnings during testing. I will add size parameter in the tests. Should numsamples be also added in the test?

czgdp1807: I think `numsamples` mean number of samples and `size` means size of each sample. Now, if `numsamples` is being removed, then `size` should also be removed and `shape` should be added. This single parameter should be capable of generating `(numsamples,) + size` . So, if users now do, `sample(N, numsamples=10, size=(3, 3))`, then from now onwards they will do, `sample(N, shape=(10, 3, 3))`. Anyways, I would suggest to focus only on changing the output of sampling API. Inputs can be changed in future PRs. Too many things in PR makes it difficult to keep track of things.

sympy/stats/rv.py
```diff
@@ -1076,57 +1081,73 @@ def sample(expr, condition=None, size=(), library='scipy',
...
+    if numsamples != 1:
+        SymPyDeprecationWarning(
+                 feature="numsamples parameter",
+                 issue=21563,
+                 deprecated_since_version="1.9").warn()
+
+        return [next(iterator) for i in range(numsamples)]
+
+    return next(iterator)
+
```
czgdp1807, line 1150: What's the motivation behind this, though?

mostlyaman: If numsamples is 1, then we dont need to iterate over and a single float or array is returned. Otherwise, we need to iterate over to get all the samples in a list.
If we iterate over even when numsamples is 1, then it would return a list of length 1, which is unnecessary.
Also, since numsamples is deprecated, any value other than the default value of numsamples i.e 1 should cause a warning.

czgdp1807: Iteration is bad I believe. `shape`(`size` is confusing) is a better option. The first element can be the number of samples and the rest size of each sample. It would help in using the vectorization benefits provided by the underlying sampling library. But anyways keep it for a separate PR. I am fine with any choice.

mostlyaman: Yes, I agree shape is better. I will probably add that as a parameter in the next PR. I am trying to get the output right in this PR.

sympy/stats/sampling/tests/test_sample_continuous_rv.py
oscarbenjamin: The point is that these tests previously tested 10 outputs from the seeded rng. With the change here only one output is tested.

sympy/stats/rv.py
oscarbenjamin: Is the size parameter new in this PR?

If size was used previously then we should keep it for now for backwards compatibility.

mostlyaman: size parameter was there before the PR. We will keep it for backwards compatibility but we should move onto shape. It seems the best for sample.
Size and numsamples make sense for sample_iterator but not sample.

sympy/stats/rv.py
mostlyaman: Some sampling methods like in matrix distributions and joint rv require both numsamples and size. Since numsamples for return_generator_infinte is effectively 1 as it returns one sample only at one time.So, I added 1.

sympy/stats/rv.py
mostlyaman: This will be removed and handled in the different classes in the next commit.

sympy/stats/rv.py
mostlyaman: Yes, all three libraries return numpy arrays.

sympy/stats/rv.py
mostlyaman: I think we can handle that in the next PR. We should make size and numsamples keyword only for compatibility since they will get replaced by shape which can be set as positional.

mostlyaman:
Thanks to @oscarbenjamin 's suggestion to use prod and reshape to form samples, consistent results are fetched now as sample doesnt depend on if-cases. Regarding the tests, i found tests already present for all libraries. Not sure why they passed despite being pymc3 broken for sample before this PR. Can the tests import pymc3 when they are being tested? 
If they could't, then the tests wouldve skipped wthout any alert.
https://github.com/sympy/sympy/blob/85e9c6c5882e8b628f5717770ce1f1439a204647/sympy/stats/sampling/tests/test_sample_continuous_rv.py#L171-L181

oscarbenjamin:
> Can the tests import pymc3 when they are being tested?

Maybe pymc3 needs to be added here:
https://github.com/sympy/sympy/blob/85e9c6c5882e8b628f5717770ce1f1439a204647/.github/workflows/runtests.yml#L89-L93

We need a separate PR adding that to see that it fails the tests if the tests on master should fail.

mostlyaman:
@oscarbenjamin optional dependencies have passed. Now, should pymc3 be added to all tests?

oscarbenjamin:
> should pymc3 be added to all tests?

No, it's an optional dependency so it should only be installed in the optional dependency tests.

mostlyaman:
So, this PR is complete then? Just wanted to point it out that stochastic_process do not use sample_iter function. It uses its own sample_stochastic_process function which still works like old sample i.e returns an iterator.

oscarbenjamin:
I've edited the release notes in the OP. I think that the changes here look good.

The only thing missing is a deprecation removal issue for the numsamples parameter. See an example here https://github.com/sympy/sympy/issues/21497 although the explanation for `numsamples` can be simpler. The new deprecation removal issue is the one whose issue number should be use in the deprecation warning so that users who see the warning are pointed to that github issue.

czgdp1807:
>  It uses its own sample_stochastic_process function which still works like old sample i.e returns an iterator.

Actually we want to fix that. May be, let's leave that for another PR. See, under Coding period, https://github.com/sympy/sympy/wiki/GSoC-Ideas#probability

mostlyaman:
>The only thing missing is a deprecation removal issue for the numsamples parameter. 

I see. An issue needs to be created detailing the changes and reasons for the change. With numsamples deprecated, users can just add another dimension in size to get the same result. I dont think that shape parameter discussed earlier is really needed anymore. size performs the same thing. Introducing shape would just be renaming size to shape.

mostlyaman:
@oscarbenjamin @czgdp1807  I've created the deprecation removal issue and updated it in rv.py
Please review it. 
Also, 
>I've edited the release notes in the OP. 

After this PR, it's the first time pymc3 sampling works. That should also be mentioned in the release notes, right?

oscarbenjamin:
> After this PR, it's the first time pymc3 sampling works. That should also be mentioned in the release notes, right?

Yeah, I guess so. Did it not work in any previous versions of sympy?

mostlyaman:
No, it did not. Samling from different libraries was introduced in v1.7 but it is bugged just like the current master. I think the devs did not test pymc3 sampling exclusively and the tests did not alert them.

oscarbenjamin:
The a release note should be added about pymc3.

mostlyaman:
Is this ready to be merged?

oscarbenjamin:
Yes, looks good. Thanks!

PR #21600, opened 2021-06-11
Adds `angle_between` to physics.Vector

sidhu1012:
Adds new functions to Vector class in physics module namely - angle_between() and `unit_normal_from()`.

#### Release Notes

* physics.vector
    * Added a new function, `angle_between` to Vector class to calculate angle between two vectors.

### DIFF
--- a/sympy/physics/vector/vector.py
+++ b/sympy/physics/vector/vector.py
@@ -723,6 +723,34 @@ class Vector(Printable, EvalfMixin):
         angle = acos(dot(vec1, vec2))
         return angle
 
+    def unit_normal_from(self, vec):
+        """
+        Returns a unit vector of the cross product of self and vec(vec x self).
+
+        Parameter
+        =========
+
+        vec : Vector
+            The Vector from which cross product is calculated.
+
+        Examples
+        ========
+
+        >>> from sympy.physics.vector import ReferenceFrame
+        >>> A = ReferenceFrame("A")
+        >>> v1 = A.x
+        >>> v2 = A.y
+        >>> v1.unit_normal_from(v2)
+        - A.z
+
+        """
+
+        from sympy.physics.vector.functions import cross
+        vec1 = self.normalize()
+        vec2 = vec.normalize()
+        unit_norm = cross(vec2, vec1)
+        return unit_norm
+
     def free_symbols(self, reference_frame):
         """
         Returns the free symbols in the measure numbers of the vector
--- a/sympy/physics/vector/vector.py
+++ b/sympy/physics/vector/vector.py
@@ -697,7 +697,7 @@ class Vector(Printable, EvalfMixin):
 
     def angle_between(self, vec):
         """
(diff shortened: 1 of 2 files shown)


sympy/physics/vector/vector.py
```diff
@@ -695,6 +695,62 @@ def applyfunc(self, f):
             d[v[1]] = v[0].applyfunc(f)
         return Vector(d)
 
+    def angle_between(self, vec):
+        """
+        Returns the angle between Vector 'vec' and self.
```
moorepants, line 700: the smallest angle between

sympy/physics/vector/vector.py
```diff
@@ -695,6 +695,62 @@ def applyfunc(self, f):
...
+
+        Examples
+        ========
+
+        >>> from sympy.physics.vector import ReferenceFrame
+        >>> A = ReferenceFrame("A")
+        >>> v1 = A.x
+        >>> v2 = A.y
+        >>> v1.angle_between(v2)
+        pi/2
```
moorepants, line 716: Add `Ax.angle_between(A.x + A.y + A.z)` to show 3D case works correctly.

sympy/physics/vector/vector.py
```diff
@@ -695,6 +695,62 @@ def applyfunc(self, f):
...
+
+        """
+
+        from sympy.physics.vector.functions import dot
```
moorepants, line 720: This should be at the top. You don't have to import if you use `Vector.dot()`.

sympy/physics/vector/vector.py
```diff
@@ -695,6 +695,62 @@ def applyfunc(self, f):
...
+        >>> from sympy.physics.vector import ReferenceFrame
+        >>> A = ReferenceFrame("A")
+        >>> v1 = A.x
+        >>> v2 = A.y
+        >>> v1.unit_normal_from(v2)
+        - A.z
+
+        """
+
+        from sympy.physics.vector.functions import cross
```
moorepants, line 748: All imports at top unless there are circular dependency issues.

sympy/physics/vector/vector.py
```diff
@@ -695,6 +695,62 @@ def applyfunc(self, f):
...
+
+        """
+
+        from sympy.physics.vector.functions import dot
+        vec1 = self.normalize()
+        vec2 = vec.normalize()
+        angle = acos(dot(vec1, vec2))
+        return angle
+
+    def unit_normal_from(self, vec):
```
moorepants, line 726: A unit vector normal to both

Maybe this is a more informative name (i.e. just a modified cross product).

vector_a.unit_cross(vector_b)

but this is so trivial. Current code is:

vector_a.cross(vector_b).normalize()

I don't think we really need a method for this, since the existing method is so simple. The angle_between is nice because you don't have to remember the formula, but this one seems unnecessary.

moorepants:
LGTM

PR #21601, opened 2021-06-11
Add kwarg to lambdify: cse

### LINKED ISSUE #18916: No option to use Common Sub-expression Detection in 'lambdify'

bjodah:
This is essentially a minimal subset of gh-21546. No attempt is made here to optimize cse with respect to number of intermediate variables.

This API should, however, allow a user to pass their own `cse` function to reuse intermediate variables, or such an alternative could be included alongside `cse` in `sympy.simplify.cse_main`. I personally don't see where `del` instructions would be useful (as was suggested in mentioned PR), but I'm happy to be proven wrong.

Fixes #18916

`lambdify` gained a keyword argument `cse`

#### Other comments
This PR has not yet been tested with heterogeneous expressions (e.g. mixing scalars and matrices).

#### Release Notes

* utilities
  * `lambdify` gained the ability to exploit common subexpressions (`cse`).

### DIFF
--- a/sympy/utilities/lambdify.py
+++ b/sympy/utilities/lambdify.py
@@ -1103,7 +1103,7 @@ class _EvaluatorPrinter:
         funcbody.extend(self._print_funcargwrapping(funcargs))
 
         funcbody.extend(unpackings)
-        funcbody.extend(['{} = {}'.format(s, e) for s, e in cses])
+        funcbody.extend(['{} = {}'.format(s, self._exprrepr(e)) for s, e in cses])
         funcbody.append('return ({})'.format(self._exprrepr(expr)))
 
         funclines = [funcsig]
--- a/sympy/utilities/lambdify.py
+++ b/sympy/utilities/lambdify.py
@@ -175,7 +175,7 @@ _lambdify_generated_counter = 1
 
 @doctest_depends_on(modules=('numpy', 'tensorflow', ), python_version=(3,))
 def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
-             dummify=False):
+             dummify=False, cse=False):
     """Convert a SymPy expression into a function that allows for fast
     numeric evaluation.
 
@@ -335,6 +335,12 @@ def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
         Set ``dummify=True`` to replace all arguments with dummy symbols
         (if ``args`` is not a string) - for example, to ensure that the
         arguments do not redefine any built-in names.
+
+    cse : bool or callable, optional
+        Whether or not common subexpression elimination should be performed on
+        the expressions. This can potentially improve performance of the generated
+        function. When ``True``, then ``sympy.simplify.cse`` is used, the user
+        may pass their own function matching the signature of said function.
 
 
     Examples
@@ -847,7 +853,15 @@ def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
         funcprinter = _TensorflowEvaluatorPrinter(printer, dummify) # type: _EvaluatorPrinter
     else:
         funcprinter = _EvaluatorPrinter(printer, dummify)
-    funcstr = funcprinter.doprint(funcname, args, expr)
+
+    if cse is True:
+        from sympy.simplify.cse_main import cse
+
+    if cse:
+        cses, expr = cse(expr)
+    else:
+        cses = ()
+    funcstr = funcprinter.doprint(funcname, args, expr, cses=cses)
 
     # Collect the module imports from the code printers.
     imp_mod_lines = []
@@ -1061,7 +1075,7 @@ class _EvaluatorPrinter:
         # Used to print the generated function arguments in a standard way
         self._argrepr = LambdaPrinter().doprint
 
-    def doprint(self, funcname, args, expr):
+    def doprint(self, funcname, args, expr, *, cses=()):
         """Returns the function definition code as a string."""
         from sympy import Dummy
 
@@ -1089,7 +1103,7 @@ class _EvaluatorPrinter:
         funcbody.extend(self._print_funcargwrapping(funcargs))
 
         funcbody.extend(unpackings)
-
+        funcbody.extend(['{} = {}'.format(s, e) for s, e in cses])
         funcbody.append('return ({})'.format(self._exprrepr(expr)))
 
         funclines = [funcsig]
--- a/sympy/utilities/tests/test_lambdify.py
+++ b/sympy/utilities/tests/test_lambdify.py
@@ -1406,3 +1406,30 @@ def test_cupy_dotproduct():
         f3(1, 2, 3) == \
         f4(1, 2, 3) == \
         cupy.array([14])
+
+
+def test_cse():
+    def dummy_cse(exprs):
+        return (), exprs
+
+    args = x, y, z
+    e1 = x + y + z
+    e2 = x + y - z
+    e3 = 2*x + 2*y - z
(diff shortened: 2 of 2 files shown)


smichr:
> but I'm happy to be proven wrong.

see [comments here](https://github.com/sympy/sympy/pull/21546#issuecomment-852776178)

If you would add a formatter option to the lambdify, I think that would be general enough. If someone wants cse they can get the default behavior. If they want to have the cse passed to their own printer (in which deletions and re-ordering would be done) then they can specify that and that formatter routine would return the lines of text to be emitted.

```python
def lambdify(..., cse, printer)
...

def printer(r, e):
   # return lines of text to calculate and return e
```

bjodah:
> > but I'm happy to be proven wrong.
> 
> see [comments here](https://github.com/sympy/sympy/pull/21546#issuecomment-852776178)

Thanks, I did read that part. I still don't see python's garbage collector benefiting by `del` over simply overwriting the variable? For example:
```python
def f(u, v):
    x1 = u + v
    x2 = u*v + x1
    x3 = u/v + x1
    del x1
    x4 = x2*x3
    return x2+1, x2 + 2, x3 + 1, x3 + 2, x4, x4+1
```
vs:
```python
def f(u, v):
    x1 = u + v
    x2 = u*v + x1
    x3 = u/v + x1
    x1 = x2*x3
    return x2+1, x2 + 2, x3 + 1, x3 + 2, x1, x1+1
```

> If you would add a formatter option to the lambdify, I think that would be general enough. If someone wants cse they can get the default behavior. If they want to have the cse passed to their own printer (in which deletions and re-ordering would be done) then they can specify that and that formatter routine would return the lines of text to be emitted.
> 

I think I see what you are suggesting, something along the lines of this then?
```python
def lambdify(..., cse, cse_formatter=default_cse_formatter)
...

def default_cse_formatter(r, e):
    return ["{} = {}".format(rr, ee) for rr, ee in zip(r, e)]
```

smichr:
You got the gist, but on second thought I guess this is not needed. The user would simply define a cse routine to return the desired set of equalities and the vector that is to be returned. I am not sure that your overwriting idea is flexible enough, however, since it could be that you release more variables than are needed for the remaining expressions, couldn't it? Could we allow None to represent items to be deleted (as I showed in the example I gave at the other PR) or perhaps introduce directives that would appear in slot 0 of a tuple and would indicate what to do with the item(s) in slot 1. e.g. `('del', (x0, x1))` would indicate printing directive to write "del x0, x1" or would it be better for the user to return tuples that get formatted or else strings that get emitted directly (so pass back the "del x0, x1" instead of a tuple at some point in the list of "replacements".

smichr:
Perhaps there is no need to delete variables: just assign any into to them. They won't take up much space/time  and since they aren't going to be used after that point, the reverse application of the replacements will still restore the original expressions. So
```python
x0 = x + y
x1 = 2*x + 1
x2 = x0 - 1
x3 = x2**2
del x2
_0 = x0/x1 + x3
_1 = x1**x0
del x1
_2 = x3
del x3
_3 = x0
del x0
_4 = x
return _2, _4, _3, _0, _1
```
would become
```python
x0 = x + y
x1 = 2*x + 1
x2 = x0 - 1
x3 = x2**2
x2 = 1
_0 = x0/x1 + x3
_1 = x1**x0
x1 = 1
_2 = x3
x3 = 1
_3 = x0
x0 = 1
_4 = x
return _2, _4, _3, _0, _1
```

smichr:
So I am +1 to what you have here.

bjodah:
Thanks @smichr. Great, I added a few tests and fixed an oversight in rendering of the intermediate expressions, looks like things are working so far. I think I'm happy with this PR at this point.

Thank you @robinechuca for getting the ball rolling here. If you want, you can "backport" commits from this PR into your. Or, if this looks good to you, we could merge this PR, and then you can just merge master into your PR and you will "only" have to deal with introducing the memory optimzations to cse by providing an alternative to `sympy.simplify.cse`. What do you say?

smichr:
Nice tests! Worth a review just to see some nice wrangling.

moorepants:
LGTM, but would it be worth running any existing lambdify numerical tests with cse to ensure same results? It would put it through its paces a bit more.

robinechuca:
> Thanks @smichr. Great, I added a few tests and fixed an oversight in rendering of the intermediate expressions, looks like things are working so far. I think I'm happy with this PR at this point.
> 
> Thank you @robinechuca for getting the ball rolling here. If you want, you can "backport" commits from this PR into your. Or, if this looks good to you, we could merge this PR, and then you can just merge master into your PR and you will "only" have to deal with introducing the memory optimzations to cse by providing an alternative to `sympy.simplify.cse`. What do you say?

In principle I have merged your branch, but if I have done anything, do not hesitate to do it. I don't care if my branch name appears at the end. If people can benefit from this improvement I would be happy, even if I wouldn't appear in it.

bjodah:
@moorepants, yes, you are right, were you thinking of parameterizing some existing tests with respect to `cse`?

@robinechuca awesome 👍, I'm closing this PR for now then, to avoid spreading the reviewers thin.

moorepants:
> @moorepants, yes, you are right, were you thinking of parameterizing some existing tests with respect to cse?

Sure, it might be overkill but would also be comforting that all lambdify tests passed with or without cse.

PR #21608, opened 2021-06-13
codegen: more capable rewrite to expm1 & cosm1 etc.

bjodah:
This PR also adds a regression test for gh-21531.

I reworked the logic behind rewriting of e.g. `exp(x)-1` and `cos(x)-1` to `expm1(x)` and `cosm1(x)` respectively. I was able to remove the `@XFAIL` decorators for some tests. Hopefully this is an improvement for all cases.

I also made the underlying logic public in the form of a class `FuncMinusOneOptim`.

#### Other comments
cc @lcontento, just FYI. I'm also interested to hear of your experiences of using these optimizations, if you want/can share that is.

Perhaps FuncMinusOneOptim is not flexible enough and we should have a class which generalizes to functions subtracting a number other than one. I could imagine e.g. `arccos(x)-pi/2` being a canddiate. However, I have not been able to find any references to such a function when I did a quick internet search. (I wonder what it would be named, `acosmhalfpi`?)
#### Release Notes

* codegen
  * More capable rewriting of terms such as `2*exp(x) + 3*exp(y) - 5` into `2*expm1(x) + 3*expm1(y)`.

### DIFF
--- a/sympy/codegen/rewriting.py
+++ b/sympy/codegen/rewriting.py
@@ -59,6 +59,9 @@ class Optimization:
         self.cost_function = cost_function
         self.priority=priority
 
+    def cheapest(self, *args):
+        return sorted(args, key=self.cost_function)[0]
+
 
 class ReplaceOptim(Optimization):
     """ Rewriting optimization calling replace on expressions.
@@ -127,9 +130,7 @@ def optimize(expr, optimizations):
         if optim.cost_function is None:
             expr = new_expr
         else:
-            before, after = map(lambda x: optim.cost_function(x), (expr, new_expr))
-            if before > after:
-                expr = new_expr
+            expr = optim.cheapest(expr, new_expr)
     return expr
 
 
@@ -214,7 +215,9 @@ class FuncMinusOneOptim(ReplaceOptim):
     """
 
     def __init__(self, func, func_m_1, opportunistic=True):
-        super().__init__(lambda e: e.is_Add, self.replace_in_Add)
+        weight = 10  # <-- this is an arbitrary number (heuristic)
+        super().__init__(lambda e: e.is_Add, self.replace_in_Add,
+                         cost_function=lambda expr: expr.count_ops() - weight*expr.count(func_m_1))
         self.func = func
         self.func_m_1 = func_m_1
         self.opportunistic = opportunistic
@@ -230,7 +233,6 @@ class FuncMinusOneOptim(ReplaceOptim):
         numsum, terms_with_func, other_non_num_terms = self._group_Add_terms(e)
         if numsum == 0:
             return e
-
         substituted, untouched = [], []
         for with_func in terms_with_func:
             if with_func.is_Mul:
@@ -257,6 +259,11 @@ class FuncMinusOneOptim(ReplaceOptim):
             untouched.append(with_func)
 
         return e.func(numsum, *substituted, *untouched, *other_non_num_terms)
+
+    def __call__(self, expr):
+        alt1 = super().__call__(expr)
+        alt2 = super().__call__(expr.factor())
+        return self.cheapest(alt1, alt2)
 
 
 expm1_opt = FuncMinusOneOptim(exp, expm1)
--- a/sympy/codegen/rewriting.py
+++ b/sympy/codegen/rewriting.py
@@ -12,8 +12,8 @@ expressions for this purpose::
     >>> x = Symbol('x')
     >>> optimize(3*exp(2*x) - 3, optims_c99)
     3*expm1(2*x)
-    >>> optimize(exp(2*x) - 3, optims_c99)
-    exp(2*x) - 3
+    >>> optimize(exp(2*x) - 1 - exp(-33), optims_c99)
+    expm1(2*x) - exp(-33)
     >>> optimize(log(3*x + 3), optims_c99)
     log1p(x) + log(3)
     >>> optimize(log(2*x + 3), optims_c99)
@@ -30,8 +30,7 @@ The ``optims_c99`` imported above is tuple containing the following instances
 
 
 """
-from itertools import chain
-from sympy import cos, exp, log, Max, Min, Wild, expand_log, Dummy, sin, sinc
+from sympy import cos, exp, log, Max, Min, Wild, expand_log, sign, sin, sinc, S
 from sympy.assumptions import Q, ask
 from sympy.codegen.cfunctions import log1p, log2, exp2, expm1
 from sympy.codegen.matrix_nodes import MatrixSolve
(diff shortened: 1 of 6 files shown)


bjodah:
OK, now this PR does more than just add a regression test. I updated the description accordingly. @lcontento would you be interested in reviewing my suggested changes?

lcontento:
> @lcontento would you be interested in reviewing my suggested changes?

I have not that much experience in optimizing expressions for floating point computation and as far as I know it's far from being a solved problem.
There are some automated approaches to do it but they usually result in much less readable expressions.
So I think having some heuristics that deal with the common cases is probably more than enough and the new heuristic you propose seems to be sound and an improvement over the old one. Code looks also good as far as I can see.

However, there is at least one case where the old code does a better job.
```python
from sympy.codegen.rewriting import optimize, expm1_opt
from sympy import sympify
expr = sympify("(1 + exp(x))^2 - 4")
optimize(expr, [expm1_opt])
```
I believe this expression should display the same numerical problems near `x = 0`.
Your branch leaves it unchanged, while the original rewriting logic simplifies it to `(exp(x) + 3)*expm1(x)`, which should be more accurate.
This is due to the fact that the original code always tries factoring before substitution.
Obviously no heuristic can deal with all cases but it may be worth and not too difficult to add back the factorization step.

bjodah:
Yes, heuristics have their limits. I was using this myself for a while (and it did help in some cases), but since then, and in that project, we have moved to polynomials (splines) so right now I'm not using this.

Excellent, exactly the kind of feedback I was hoping for. I'll add your example to the test suite, and will try to rework my patch here (as soon as I get some spare time).

sympy/codegen/rewriting.py
```diff
@@ -175,42 +175,99 @@ def optimize(expr, optimizations):
...
-        factored = protected.factor()
-        new_old = {v: self.func(k) for k, v in old_new.items()}
-        return factored.replace(_d - 1, lambda d: self.func_m_1(new_old[d].args[0])).xreplace(new_old)
-
-    def __call__(self, e):
-        numbers, non_num = sift(e.args, lambda arg: arg.is_number, binary=True)
-        non_num_func, non_num_other = sift(non_num, lambda arg: arg.has(self.func),
-            binary=True)
+    def _group_Add_terms(self, add):
+        numbers, non_num = sift(add.args, lambda arg: arg.is_number, binary=True)
```
smichr, line 226: I wonder if you should further sift numbers into Rational and other:
```python
>>> optimize(2*exp(x)-2,[expm1_opt]).subs(x,1e-11)
2.00000000001000e-11
>>> (2*exp(x)-2).subs(x,1e-11)
2.00000016548074e-11

>>> optimize(2*exp(x)-pi,[expm1_opt])
2*expm1(x) - pi + 2
>>> _.subs(x,1e-11)
2.00000000002 - pi
>>> (2*exp(x)-pi).subs(x,1e-11)
2.00000000002 - pi
```

bjodah: Ah, yes, not very beautiful, but I am not sure how `is_rational` is central to avoiding cancellation? For example, if I carefully construct a fraction of pi close to two:
```python
In [31]: about_two = 2**(58/S(117))*3**(97/S(117))*5**(4/S(39))*7**(92/S(117))/S(30)*pi

In [32]: xval = S(10)**-11

In [33]: ref = (2*exp(x) - about_two).subs(x, xval).n(17); print(ref)
2.0418173913673214e-11

In [34]: no_opt = (2*exp(x) - about_two).subs(x, xval.evalf()).evalf()

In [35]: with_opt = optimize(2*exp(x) - about_two, [expm1_opt]).n(25).subs(x, 1e-11).evalf()

In [36]: with_opt - ref, no_opt - ref
Out[36]: (1.1536301877952077e-26, 1.6547074214222335e-18)
```
somewhat orthogonal to what we are discussing: note how I needed to insert `.n(25)` there, `evalf` didn't cut it. Hopefully our (code) printers compute such sums to the full precision of the targeted data type... (but this is another issue of course).

smichr: What the example I gave might be saying is that there is no benefit to rewriting unless you can absorb an integer; if the number is not rational then it looks like there is no benefit. 

As far as the examples go...I'm nervous about the multiple evalfs that you use to demonstrate the difference. I need a more direct demonstration that the optimization is more precise.

bjodah: Right, I wanted to make sure myself, so I did spend some time adding a test which does a full roundtrip to (generated) C code and back. I just added those commits here.

sympy/codegen/tests/test_rewriting.py
```diff
@@ -297,3 +349,63 @@ def check(d):
...
+    return func_rewritten(x)
+'''
+    with tempfile.TemporaryDirectory() as folder:
+        mod, info = compile_link_import_strings(
+            [('func.c', func_c), ('_func.pyx', func_pyx)],
+            build_dir=folder, compile_kwargs=dict(std='c99')
+        )
+        err_rewritten = abs(mod.py_rewritten(1e-11) - ref)
+        err_unchanged = abs(mod.py_unchanged(1e-11) - ref)
+        assert 1e-27 < err_rewritten < 1e-25  # highly accurate.
```
bjodah, line 404: And here is where I show that the accuracy is improved by several orders of magnitude, even though we are dealing with an irrational number.

sympy/printing/tests/test_str.py
```diff
@@ -252,6 +252,8 @@ def test_Mul():
     # For issue 14160
     assert str(Mul(-2, x, Pow(Mul(y,y,evaluate=False), -1, evaluate=False),
                                                 evaluate=False)) == '-2*x/(y*y)'
+    # issue 21537
+    assert str(Mul(x, Pow(1/y, -1, evaluate=False), evaluate=False)) == 'x/(1/y)'
```
bjodah, line 257: Umm, I don't think I've actively added this line. Weird, I have merged with master...

sympy/codegen/tests/test_rewriting.py
```diff
@@ -297,3 +349,63 @@ def check(d):
...
+
+double func_unchanged(double x) {
+    return %(unchanged)s;
+}
+double func_rewritten(double x) {
+    return %(rewritten)s;
+}
+''' % dict(unchanged=ccode(unchanged.n(NUMBER_OF_DIGITS)),
+           rewritten=ccode(rewritten.n(NUMBER_OF_DIGITS)))
+
```
bjodah, line 387: And this is what `func_c` looks like once rendered:
```C
#include <math.h>

double func_unchanged(double x) {
    return 2.0*exp(x) - 1.9999999999995818;
}
double func_rewritten(double x) {
    return 2.0*expm1(x) + 4.1817391357322551e-13;
}
```

smichr:
OK, thanks @bjodah

bjodah:
Thanks for the review @smichr

bjodah:
And thank you @lcontento !

PR #21609, opened 2021-06-13
Refactor(dsolve): nth order reducible

Mohitbalwani26:

SEE #18348

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/solvers/ode/tests/test_single.py
+++ b/sympy/solvers/ode/tests/test_single.py
@@ -501,7 +501,7 @@ def test_nth_linear_constant_coeff_undetermined_coefficients():
 
 
 def test_nth_order_reducible():
-    F = lambda obj: NthOrderReducible(SingleODEProblem(eq, f(x), x))._matches()
+    F = lambda eq: NthOrderReducible(SingleODEProblem(eq, f(x), x))._matches()
     D = Derivative
     assert F(D(y*f(x), x, y) + D(f(x), x)) == False
     assert F(D(y*f(y), y, y) + D(f(y), y)) == False
--- a/sympy/solvers/ode/tests/test_single.py
+++ b/sympy/solvers/ode/tests/test_single.py
@@ -42,7 +42,7 @@ from sympy.integrals.risch import NonElementaryIntegral
 from sympy.solvers.ode import classify_ode, dsolve
 from sympy.solvers.ode.ode import allhints, _remove_redundant_solutions
 from sympy.solvers.ode.single import (FirstLinear, ODEMatchError,
-    SingleODEProblem, SingleODESolver)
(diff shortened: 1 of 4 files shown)


sympy/solvers/ode/tests/test_single.py
```diff
@@ -501,16 +501,14 @@ def test_nth_linear_constant_coeff_undetermined_coefficients():
...
-    assert F(f(x)*D(f(x), x) + D(f(x), x, 2)) is None
-    assert F(D(x*f(y), y, 2) + D(u*y*f(x), x, 3)) is None  # no simplification by design
-    assert F(D(f(y), y, 2) + D(f(y), y, 3) + D(f(x), x, 4)) is None
-    assert F(D(f(x), x, 2) + D(f(x), x, 3)) == dict(n=2)
+    assert F(SingleODEProblem(D(y*f(x), x, y) + D(f(x), x), f(x), x)) == False
+    assert F(SingleODEProblem(D(y*f(y), y, y) + D(f(y), y), f(x), x)) == False
+    assert F(SingleODEProblem(f(x)*D(f(x), x) + D(f(x), x, 2), f(x), x))== False
+    assert F(SingleODEProblem(D(x*f(y), y, 2) + D(u*y*f(x), x, 3), f(x), x)) == False  # no simplification by design
+    assert F(SingleODEProblem(D(f(y), y, 2) + D(f(y), y, 3) + D(f(x), x, 4), f(x), x)) == False
+    assert F(SingleODEProblem(D(f(x), x, 2) + D(f(x), x, 3), f(x), x)) == True
```
oscarbenjamin, line 511: Maybe just change the function `F` rather than adding the same to all examples

oscarbenjamin:
This is another example of a solver that is really just a substitution.

sympy/solvers/ode/tests/test_single.py
Mohitbalwani26: Sorry, I couldn't understand the part change the function F?

oscarbenjamin: The function `F` is supposed to handle the repetitive parts in these test examples. You've just added repetitive changes to each example when you could have just changed the function `F`.

Mohitbalwani26:
````F = lambda eq: NthOrderReducible(SingleODEProblem(eq, f(x) ,  x) )._matches()
D = Derivative
assert F(D(y*f(x), x, y) + D(f(x), x)) == False
````
Something like this?

Mohitbalwani26:
> This is another example of a solver that is really just a substitution.

Yes, I agree. After moving everything we should come up with either some generalised substitution function which try different substitutions or we can try to handle it inside pattern matching

oscarbenjamin:
Looks good. Thanks

Issue #21560, opened 2021-06-01
`1**x` == 1 but `1**oo -> nan`


smichr:
Both of the autosimplifications cannot be true. In the context of AccumBounds I could believe that `1**oo -> 1` might not be what you want but in other contexts it seems that an exact `1` raised to any power, even `oo`, should be 1 as the symbolic case asserts.

oscarbenjamin:
See also https://github.com/sympy/sympy/issues/17224

smichr:
IMO we need to bear whatever pains come from making the default symbol be finite (and non-zero so `0/x -> 0` is valid).

smichr:
There is a nice discussion of the indetermint nature of 1**oo in the context of a *limit* [here](http://www.math-principles.com/2012/12/indeterminate-form-one-raised-infinity.html). But I wonder why SymPy has chosen to use the limit concept (apparently) to decide that `1**oo` should be nan instead of 1. It seems that there has been a confluence of the idea that a limit leading to `1**oo` is indeterminant with the notion that `1**x` is 1 for any positive number, no matter how large.

Interestingly (to me, not being heavily versed in limits), the `limit(x**oo, x, 1, '-')` is not 0 nor is `limit(x**oo, x, 1, '+')` infinite. Both cases remain unevaluated.

I can understand that `limit` should watch out for using substitution in the case of powers to avoid letting `f(x)**g(x)` with `limit(f(x), x, v) -> 1` and `limit(g(x), x, v) -> oo` but to co-opt the normal meaning of `1**x` as nan seems to be unnecessary.

Can anyone demonstrate otherwise?

smichr:
Since we have the nuanced `1/0->zoo` and `1**zoo` is nan, perhaps we could relax the constraint on `1**oo`?

smichr:
It seems that the consensus is that letting `1**oo` evaluate to 1 is an unwise direction. So fixing assumptions on default symbols is the direction to take before SymPy 2.0.

PR #21619, opened 2021-06-14
let `1**oo == 1`

### LINKED ISSUE #18509: pi raised to infinity gives incorrect result
`pi**oo` should be `oo`.

### LINKED ISSUE #21740: Simplifying zero-like expression discards Integer/Float type
Simplifying a zero-like expression loses information as to whether the original expression was integer-like or float-like.

This behaviour changed somewhere between sympy 1.4 and 1.8; more precisely, it is caused by commit 7649a7c3f1c6c4514b6b1ceaaffede5634673867.

Compare:

**Before:**
```Python
>>> sympy.simplify(sympy.Float("0."))
0.0
>>> type(sympy.simplify(sympy.Float("0.")))
sympy.core.numbers.Float
```

**After:**
```Python
>>> sympy.simplify(sympy.Float("0."))
0
>>> type(sympy.simplify(sympy.Float("0.")))
sympy.core.numbers.Zero
```

Note that this issue does not occur for One, because there is no special case in simplify() to return One for a one-like expression (as was introduced in 7649a7c3f1c6c4514b6b1ceaaffede5634673867 for Zero):

```Python
>>> sympy.simplify(sympy.Float("1."))
1.00000000000000
>>> type(sympy.simplify(sympy.Float("1.")))
sympy.core.numbers.Float
```

This is in any case inconsistent; a solution should address both the Zero and the One case in a consistent manner.

Potential solutions:
- Should zero-like expressions not be simplified to Zero, but remain a Float or Integer? (Basically undo the special case introduced by 7649a7c3f1c6c4514b6b1ceaaffede5634673867.)
- Should a new flag be added to the Zero class (which currently inherits from IntegerConstant) to indicate integer-ness or float-ness?

Potentially related discussion: #6545

smichr:

fixes #18509
fixes #21740 
addresses #21560 

retains Float if one of the two identical items in a difference is a Float (e.g. 1.0 - 1 == 0.0, not 0 and 0.0 - 0 == 0.0)
#### Other comments

This only returns 1 when the base is literally 1. `subs` should not be able to create this result unless the base is a literal `1` to start with, e.g.
```python
>>> eq = (1 + 1/x)**x
>>> eq.subs(x, oo)
nan
>>> limit(eq, x, oo)
E
```

Since we distinguish between 1 and 1.0 (e.g. `1**x == 1` but `1.0**x` is unevaluated) it seems like we could also make `AccumBounds(1,2)**oo` give `AccumBounds(1, oo)` instead of `AccumBounds(0, oo)` but `AccumBounds(1.0, 2)**oo -> AccumBounds(0, oo)`. For reasons articulated by @jksuom, however, AccumBounds retains its inherent skepticism and treats 1 as though it may have been approached from the left or the right.

- [x] update Pow docstring
- [x] should the substitution above be allowed to give 1? to not allow it seems nanny-ish
- [ ] make accurate release notes

- #### Release Notes

* core
  * `1**oo` and `(1**f(x)).subs(x,v)` with `f(v) == oo` now return 1


sympy/core/tests/test_power.py
```diff
@@ -593,3 +594,10 @@ def test_powers_of_I():
         1, sqrt(I), I, sqrt(I)**3, -1, -sqrt(I), -I, -sqrt(I)**3,
         1, sqrt(I), I, sqrt(I)**3, -1]
     assert sqrt(I)**(S(9)/2) == -I**(S(1)/4)
+
+
+def test_indeterminant_forms():
+    x = symbols('x')
+    assert ((1 + 1/x)**x).subs(x, oo) is S.NaN
+    assert (x**oo).subs(x, 1) is S.NaN
+    assert (1**x).subs(x, oo) is S.One
```
smichr, line 603:
```suggestion
    assert Pow(1, x, evaluate=False).subs(x, oo) is S.One
```

oscarbenjamin:
This makes `1**oo = 1` and `1**-oo = 1` which I think seems reasonable although I'm interested to hear what others think.

This leads `1**zoo = nan`. Should that be defined as well?

smichr:
> This leads `1**zoo = nan`. Should that be defined as well?

I would advocate, yes, since it is an exact 1. This is a fundamental understanding in SymPy: Rationals are exact, Floats are not.

Perhaps `1/oo` and `1/zoo` should give `0.0` (Float) instead of `0` and then substitutions would not need any code monitoring: ` ((1 + 1/x)**x).subs(x, oo) -> (1 + 1/oo)**oo -> (1 + 0.0)**oo -> 1.0**oo -> nan` automatically.

jksuom:
I have been testing with substitutions:
```
In [3]: e = Pow(S.One, oo, evaluate=False)                                      

In [4]: e.subs(x, 0)
...
~/test/sympy/sympy/core/power.py in _eval_subs(self, old, new)
    806         if e is S.Infinity:
    807             if self.base is S.One:
--> 808                 return b
    809             b = self.base.subs(old, new)
    810             if b == 1:

UnboundLocalError: local variable 'b' referenced before assignment
```
Did you intend to return `S.One`?

sympy/core/power.py
```diff
@@ -798,6 +801,16 @@ def _eval_subs(self, old, new):
                 return e.__rpow__(b)
             return self.func(b, e)
 
+        # guard against 1**oo
+        e = self.exp.subs(old, new)
+        if e is S.Infinity:
+            if self.base is S.One:
+                return b
```
smichr, line 808:
```suggestion
                return self.base
```

smichr:
> Did you intend to return `S.One`?

yes, thanks

asmeurer:
I guess the point of making indeterminate forms give nan is that you can avoid expressions evaluating to the wrong answer when substituting a finite value. But I see you've guarded against that in the subs code, which is a nice idea. We might want to do the same thing for `0**0`, which is also indeterminate. 

```py
>>> limit((exp(-x))**(1/x), x, oo)
exp(-1)
>>> ((exp(-x))**(1/x)).subs(x, oo)
1
```

I'm not sure if other issues can arise from allowing indeterminate forms to evaluate like this, but it seems fine. This also matches how floats evaluate (`1.**float('inf')`).

asmeurer:
I suppose similar things can happen from cancellation even without using `subs`, like `(exp(-1/(x - x))**(x - x))`. That's a contrived example, but maybe it can happen in a more nontrivial way. 

The `oo` one is harder because most SymPy operations give `zoo` instead of `oo`, even `Symbol('x', positive=True)/0` gives `zoo`.

smichr:
It's interesting to see how python handles some of these issues (per Aaron's comment):
```python
>>> int(1)**float('inf')
1.0
>>> float(1)**float('inf')
1.0
>>> int(1)**int(30)
1
>>> int(1)**1j
(1+0j)
>>> float('inf')**int(0)
1.0
>>> float('inf')**float(0)
1.0

```

I'll see what happens when letting `1/oo = 1/zoo -> 0.0`

sympy/core/numbers.py
```diff
@@ -3386,6 +3386,13 @@ def approximation(self, number_cls):
     def _eval_evalf(self, prec):
         return Float._new(self._as_mpf_val(prec), prec)
 
+    def _eval_power(self, e):
```
smichr, line 3389: check lines 13xx -- are they covered?
```python
        if self == 0:
            if expt.is_positive:
                return S.Zero
            if expt.is_negative:
                return S.Infinity
```

asmeurer:
BTW "Python" here really just means the IEEE standard (except when using complex numbers). We do mention in the Pow docstring that we break from the standard intentionally in some places (e.g., we have zoo in addition to signed infinities, and our nan isn't really like a floating point nan).

smichr:
It would be great for anyone else to test this with problem expressions to see if the behavior is as expected.

sympy/calculus/util.py
```diff
@@ -947,13 +947,23 @@ class AccumulationBounds(AtomicExpr):
...
-    >>> AccumBounds(0, 1)**oo
+    >>> AccumBounds(0, 1.0)**oo
     AccumBounds(0, oo)
 
+    A boundary of 1 will be assumed as exact:
+
+    >>> AccumBounds(1, 2)**oo
+    AccumBounds(1, oo)
+    >>> AccumBounds(0, 1)**oo
+    AccumBounds(0, 1)
```
jksuom, line 965: AccumulationBounds denotes an interval containing the limits determined by some expression. This example would imply that if `f` has limit in the interval [0, 1] and `g` has limit `oo`, then also `f**g` would have limit in the same interval. But that is not true, e.g. for `f = 1 + 1/x, g = x, x -> oo`.

smichr: What does sympy give for the limit of `(1 +1/x)**x`? I think we have to avoid putting possible history on states or else we can't resolve cases that are resolvable. The limit of `sin(x)**x` must lie between -1 and 1 since it strictly never exceeds an absolute value of 1. 0.0 is a better answer since you end up with delta functions at the max/min locations and 0 everywhere else.

If a bound approaches 1 from the positive side then a bound should use 1.0 and the behavior should be right. At least that is my goal in adding this nuance.

oscarbenjamin: I guess `AccumBounds` represents a limit so it has to be handled differently from ordinary interval arithmetic if the result depends on which side a limit would approach the boundary.

smichr: The question for me is, "Does allowing AccumBounds to differentiate between exact and inexact bounds break anything?" If you say that this nuance is just a bad idea for people using AccumBounds as a limit-related quantity used for reasoning then I will trust you. But if the master behavior of `(sin(x)**x).subs(x,oo)` giving `(-oo,oo)` as the result indicates to me that it is not a very powerful reasoning tool at present.

I guess what I am looking for are demonstrations of how this refinement will hurt reasoning with AccumBounds and SymPy.

Here are some current behaviors using @jksuom example:
```python
>>> (1+1/x)**x
(1 + 1/x)**x
>>> eq=_;limit(eq,x,oo)
E
>>> eq.subs(x,oo)
nan
>>> (1+1/x).subs(x,oo)
1.00000000000000
>>> _**oo
nan
```
Hmm...I tried one more and note something interesting:
```python
>>> (1-1/x).subs(x,oo)
1
>>> ((1-1/x)**x).subs(x,oo)
1
```
This is the result of the difference
```
>>> 1 + 0.
1.0
>>> 1 - 0.
1
```

jksuom: > ((1-1/x)**x).subs(x,oo)

`subs` can sometimes be used as a poor man's limit but one must be cautious. The correct limit value is `1/E`.

jksuom: > The limit of sin(x)**x must lie between -1 and 1

This is not a good example as `sin(x)**x` is not real when `sin(x)` is negative, hence not int the interval between -1 and 1. In this case, there is no object in SymPy that would describe the limit better than `nan`. It is not optimal as there is no size information but we don't have complex analogs of `AccumBounds`.

jksuom: > Does allowing AccumBounds to differentiate between exact and inexact bounds break anything?

I would generally try to avoid floating point numbers in SymPy but AccumBounds should behave in the same way even if the bounds are Floats.

smichr: I'll see if I can patch around the changes outside of AccumBounds that affect AccumBounds (my `1oob` branch).

smichr: > AccumBounds should behave in the same way even if the bounds are Floats.

This is where we then end up with inconsistent behavior like `1**x == 1` but `1.0**x -> unevaluated`. We might be leaving the latter unevaluated because we don't know if it goes to 0 or infinity if x is infinite (though it might just be a quirk since `1.0**I` also remains unevaluated).

sympy/calculus/util.py
```diff
@@ -1339,38 +1356,41 @@ def __pow__(self, other):
...
                     if other.is_extended_negative:
                         if self.min.is_zero:
-                            return AccumBounds(self.max**other, oo)
+                            return AccumBounds(_pow(other, self.max), oo)
                         if self.max.is_zero:
-                            return AccumBounds(-oo, self.min**other)
+                            return AccumBounds(-oo, _pow(other, self.min))
                         return AccumBounds(-oo, oo)
-                    return AccumBounds(self.min**other, self.max**other)
+                    return AccumBounds(*_pow(other, self.min, self.max))
```
smichr, line 1393:
```suggestion
                    return AccumBounds(*_pow(other, *self.args))
```

sympy/calculus/util.py
```diff
@@ -1339,38 +1356,41 @@ def __pow__(self, other):
...
                         if self.min.is_zero:
-                            return AccumBounds(self.max**other, oo)
+                            return AccumBounds(_pow(other, self.max), oo)
                         if self.max.is_zero:
-                            return AccumBounds(self.min**other, oo)
+                            return AccumBounds(_pow(other, self.min), oo)
                         return AccumBounds(0, oo)
                     return AccumBounds(
-                        S.Zero, Max(self.min**other, self.max**other))
+                        S.Zero, Max(*_pow(other, self.min, self.max)))
```
smichr, line 1385:
```suggestion
                        S.Zero, Max(*_pow(other, *self.args)))
```

sympy/calculus/util.py
```diff
@@ -1339,38 +1356,41 @@ def __pow__(self, other):
...
-                        return self.func(self.min**other, self.max**other)
+                        return self.func(
+                            _pow(other, self.min),
+                            _pow(other, self.max))
 
             if other.is_zero:
                 return S.One  # x**0 = 1
 
             if other.is_Integer or other.is_integer:
+                args = _pow(other, self.min, self.max)
```
smichr, line 1367:
```suggestion
                args = _pow(other, *self.args)
```

sympy/calculus/util.py
```diff
@@ -1339,38 +1356,41 @@ def __pow__(self, other):
                     # no 0 to worry about
                     if other.is_nonnegative:
                         # no infinity to worry about
-                        return self.func(self.min**other, self.max**other)
+                        return self.func(
+                            _pow(other, self.min),
+                            _pow(other, self.max))
```
smichr, line 1361:
```suggestion
                        return self.func(*_pow(other, *self.args))
```

smichr:
I will recheck behaviors that were commented on above and update the Pow docstrings. Otherwise, I think this is ready to go. AccumBounds behavior has not been changed and `subs` does not require special handling since powers involving `oo` produce Floats which generate nans (and discourage erroneous conclusions about limits in such cases).

sympy/simplify/tests/test_simplify.py
```diff
@@ -868,8 +868,12 @@ def test_issue_7971():
     z = Integral(x, (x, 1, 1))
     assert z != 0
     assert simplify(z) is S.Zero
+    # 21740
+    assert simplify(S.Zero) is S.Zero
+    z = simplify(Float(0))
+    assert z == Float(0) and type(z) is Float
 
-
+
```
smichr, line 876:
```suggestion

```

smichr:
Please see if the behavior as summarized in the table in power.py presents a coherent number model.

sympy/core/power.py
```diff
@@ -193,39 +193,43 @@ class Pow(Expr):
...
     +--------------+---------+-----------------------------------------------+
-    | 1**-1        | 1       |                                               |
-    +--------------+---------+-----------------------------------------------+
-    | oo**-1       | 0       |                                               |
-    +--------------+---------+-----------------------------------------------+
     | 0**oo        | 0       | Because for all complex numbers z near        |
-    |              |         | 0, z**oo -> 0.                                |
+    | 0.0**oo      |         | 0, z**oo -> 0                                 |
+    +--------------+---------+-----------------------------------------------+
+    | 0.0**e with  | 0.0     | 0, z**oo -> 0                                 |
```
smichr, line 213:
```suggestion
    | 0.0**e with  | 0.0     |  a finite exponent does not remove            |
    |              |         | any difference from 0 represented by 0.0      |
```

sympy/core/power.py
```diff
@@ -193,39 +193,43 @@ class Pow(Expr):
     +--------------+---------+-----------------------------------------------+
     | z**1         | z       |                                               |
     +--------------+---------+-----------------------------------------------+
-    | (-oo)**(-1)  | 0       |                                               |
+    | 1**z         | 1       | because 1 is considered exact                 |
+    +--------------+---------+-----------------------------------------------+
+    | oo**-1       | 0.0     |                                               |
+    | (-oo)**-1    |         |                                               |
```
smichr, line 199:
```suggestion
    | oo**-1       | 0.0     | guards against using oo to make               |
    | (-oo)**-1    |         | conclusions best made with limits             |
```

sympy/core/power.py
```diff
@@ -193,39 +193,43 @@ class Pow(Expr):
...
     +--------------+---------+-----------------------------------------------+
     | (-1)**-1     | -1      |                                               |
```
smichr, line 201:
```suggestion
```

jksuom:
I feel uneasy about mixing `Float(0.0)` with exact numbers. There is a lot of code that has problems with Floats: in polys, heurisch, meijerint... Also many functions have issues with inexact numbers.
```
>>> sin((2 + 1/(1 + 1/x))*pi).as_leading_term(x)
0  # should be pi*x
>>> gamma(-2 + 1/(1 + 1/x)).as_leading_term(x)
zoo  # should be 1/(2*x)
```
Wrong results like these are caused by the value of the argument at `x = 0` being a Float that is not recognized as an integer:
```
>>> (-2 + 1/(1 + 1/x)).subs(x, 0)
-2.00000000000000
>>> _.is_integer
False
```

smichr:
Thanks, @jksuom . I will restore special handling in subs and non-float value for 1/oo.

asmeurer:
> Wrong results like these are caused by the value of the argument at x = 0 being a Float that is not recognized as an integer:

That sounds like an argument for making floating point numbers compare exactly equal to rationals (the opposite of https://github.com/sympy/sympy/pull/20033), which also implies making the assumptions on floats treat them the same as rationals. I agree that the change here of producing a float out of nowhere is a bad idea, but this exact same problem can still occur if a user puts a float in an expression somewhere (which real world SymPy users do all the time, https://github.com/sympy/sympy/pull/20033#issuecomment-710417294). (not to hijack this PR too much, since this is unrelated)

asmeurer:
> Please see if the behavior as summarized in the table in power.py presents a coherent number model.

Just to clarify, the only difference between here and master is the `1**oo` mentioned in the PR title? Do the other changes in the table only clarify behavior that is there in master too, or are there other evaluation changes here?

sympy/core/power.py
```diff
@@ -193,39 +193,41 @@ class Pow(Expr):
...
-    | 1**-1        | 1       |                                               |
-    +--------------+---------+-----------------------------------------------+
-    | oo**-1       | 0       |                                               |
-    +--------------+---------+-----------------------------------------------+
     | 0**oo        | 0       | Because for all complex numbers z near        |
-    |              |         | 0, z**oo -> 0.                                |
+    | 0.0**oo      |         | 0, z**oo -> 0                                 |
+    +--------------+---------+-----------------------------------------------+
+    | 0.0**e with  | 0.0     | a finite exponent does not remove             |
+    | 0 < e < oo   |         | any difference from 0 represented by 0.0      |
```
asmeurer, line 212: I don't like listing floats separately from exact values here. I agree that expressions with floating-point inputs should produce floating-point outputs, but that's a general principle in SymPy. Saying `0**x = 0` but `0.0**x = 0.0` here is redundant and distracting IMO. I also don't like calling 0.0 "a difference from 0". If that's true (I don't think it should be), then `0.0**e` ought not to evaluate at all, since "it could be negative" (but again, I don't actually think this).

Is it generally true herethat the value of the result doesn't change from replacing a rational with an equivalent float, only the type? If not, I think it should be.

asmeurer:
I generally like the table in Pow, and I'm a fan of expanding it and expanding the explanations. I actually think we should have a proper document in the documentation that goes into much more mathematical detail for each case, because it's useful to clarify why the various conventions are chosen (not stating that needs to be done here, though).

asmeurer:
I can't reproduce the test failure on CI. Maybe it only occurs when running the full test suite, which I haven't tried. It's also a little odd (why is the same exception chained to itself several times?)

oscarbenjamin:
> why is the same exception chained to itself several times?

A possible cause is if the exception is raised during the printing code. If formatting the exception calls the printers then it can lead to recursive raising.

asmeurer:
The exception is just `TypeError: cannot determine truth value of Relational`. But I just noticed the exact traceback isn't repeated each time, but is slightly different. I think this has to do with the [`except TypeError` line](https://github.com/sympy/sympy/blob/78380bf18ae869f8c14870d2201a186486266f0f/sympy/core/cache.py#L73) in caching, which would also explain why the failure didn't reproduce when I ran just that one test. It seems that line has been there for a while, but it's clearly problematic as it's not acting as intended when the TypeError is an unrelated bug like this one. As far as I can tell, the intention is to guard against functions returning non-hashable results. IMO the caching decorator shouldn't even have that guard. If a cached function returns a non-hashable result, that's a bug that should be fixed, or else that function just isn't a candidate for caching and a lower level function that does always return a hashable result should be cached instead.

asmeurer:
I have a preliminary change along these lines at https://github.com/sympy/sympy/pull/21799. I expect it won't be trivial to fix this, since the cache is so fragile. At any rate, I would guess the failure in CI is legitimate, so you may want to try to figure out what is going on. Based on the traceback, there is an `if x < y` type line in the code that is getting symbolic inputs.

oscarbenjamin:
The `except TypeError` is problematic. I think it also means that a possibly cached call that raises TypeError always gets called twice so if the exception is caught higher up then the failed operation always takes twice as long.

sympy/calculus/util.py
```diff
@@ -1287,27 +1287,44 @@ def __rtruediv__(self, other):
...
                         if other.is_extended_negative:
-                            return AccumBounds(Mul(other, 1 / self.min), oo)
+                            return AccumBounds(Mul(other, invM), oo)
                     return AccumBounds(-oo, oo)
                 else:
-                    return AccumBounds(Min(other / self.min, other / self.max),
-                                       Max(other / self.min, other / self.max))
+                    args = [other/i if not i.is_infinite
+                        else 0 for i in self.args]
+                    if args[0] > args[1]:
```
asmeurer, line 1308: I didn't realize this line from the test failure was actually added in this PR. We can't assume args has non-symbolic values here.

asmeurer:
Yep, that is exactly what is happening. And based on the test failure here, it is even worse, with the same exception being raised multiple times (I counted 64 times in the build log). I guess since so many very basic things like operations are cached, there ends up being a TypeError recursively multiple times. Probably 6 levels deep here based on the number 64.

Clearly if a function took a while to actually raise an exception this could have a serious impact on performance, but even if it doesn't, this exponential growth of exception chaining can do it anyway.

sympy/calculus/util.py
smichr, line 1308:
```suggestion
                    if (args[0] - args[1]).is_extended_positive:
```

jksuom:
> That sounds like an argument for making floating point numbers compare exactly equal to rationals 

The arithmetic operations of floating point numbers are different from those of rational numbers (and floats of different precision). All sufficiently big floats (of certain precision) are represented by an integer. If we make them equal. then both `bigfloat` and `bigfloat + 1/2` would look like integers (`is_integer is True`). That would almost certainly lead to inconsistent results.

sympy/core/power.py
smichr: I reviewed the table value with sympy.live output and note:
```
The following were formerly 0 but now

0.0**3 is 0.0 (and 0.0 permits interpretation as 0 +/- eps so could be + or -)
zoo**-1 is 0.0
oo**-oo is 0.0
(+/-oo)**(-1+I) is 0.0

Also, `(-oo)**(1 + I)` is now zoo instead of `(-1)**(1+I)*zoo`
```
I will have to double check that those table values actually compute as represented.

smichr:
> I will restore special handling in subs and non-float value for `1/oo`.

I will double check tonight, but I don't think that I have restored that in this branch. So an important question to resolve is whether that is really necessary -- tests aren't failing because of this. You [did note](https://github.com/sympy/sympy/pull/21619#issuecomment-881346836), however, that there were expressions that gave trouble in master because of float(0) not being handled properly. I think your concern is that the likelihood of meeting on of those results increases with this PR.

sympy/core/power.py
smichr: **code to run on this PR to affirm table entries**
```python
var('z b')
fzero = lambda x: c == 0 and x is not S.Zero
assert z**1 == z
assert 1**z is S.One
assert fzero(oo**-1)  # @jkalevi wants a change
assert fzero((-oo)**-1)
assert all((-1)**i is S.NaN for i in (-oo, oo))
assert S.Zero**-1 is zoo
assert all(i**oo is S.Zero for i in (0, 0.0))
assert fzero(0.0**S(3))
assert 0**-oo is zoo
assert all(i**j is S.NaN for i in (-1., 1.) for j in (-oo, oo))
assert fzero(zoo**-1)
assert z**zoo is S.NaN
assert oo**oo is oo
assert fzero(oo**-oo)
assert all((-oo)**i is S.NaN for i in (-oo, oo))
assert all(i**(1 + I) is zoo for i in (-oo,oo))
assert all(fzero(i**(-1 + I)) for i in (-oo,oo))
```

sympy/core/power.py
```diff
@@ -194,39 +194,41 @@ class Pow(Expr):
...
+    | b**e with    | nan     | Because there are various cases where         |
+    | b = +/-1.0   |         | lim(x(t),t)=1, lim(y(t),t)=oo (or -oo),       |
+    | e = +/-oo    |         | but lim(x(t)**y(t), t) != 1.  See [3].        |
     +--------------+---------+-----------------------------------------------+
-    | b**zoo       | nan     | Because b**z has no limit as z -> zoo         |
+    | zoo**-1      | 0.0     |                                               |
     +--------------+---------+-----------------------------------------------+
-    | (-1)**oo     | nan     | Because of oscillations in the limit.         |
-    | (-1)**(-oo)  |         |                                               |
+    | b**zoo       | nan     | Because b**z has no limit as z -> zoo         |
```
smichr, line 227:
```suggestion
    | b**zoo       | nan     | Because b**e has no limit as e -> zoo         |
```

sympy/core/power.py
```diff
@@ -236,15 +238,15 @@ class Pow(Expr):
     |              |         | oo. If e is I, then the limit does not exist  |
     |              |         | and nan is used to indicate that.             |
     +--------------+---------+-----------------------------------------------+
-    | oo**(1+I)    | zoo     | If the real part of e is positive, then the   |
-    | (-oo)**(1+I) |         | limit of abs(x**e) is oo. So the limit value  |
-    |              |         | is zoo.                                       |
+    | b**(1+I) and | zoo     | If the real part of e is positive, then the   |
+    | b = +/-oo    |         | limit of abs(x**e) is oo. So the limit value  |
+    |              |         | is zoo. (-oo)**(1+I) simplifies to zoo.       |
```
smichr, line 243:
```suggestion
    | b**e     and | zoo     | If the real part of e is positive, then the   |
    | b = +/-oo    |         | limit of abs(x**e) is oo. So the limit value  |
    | and re(e)>0  |         | is zoo. (-oo)**(1+I) simplifies to zoo.       |
```

sympy/core/power.py
```diff
@@ -236,15 +238,15 @@ class Pow(Expr):
...
     +--------------+---------+-----------------------------------------------+
-    | oo**(-1+I)   | 0       | If the real part of e is negative, then the   |
-    | -oo**(-1+I)  |         | limit is 0.                                   |
+    | b**(-1+I) and| 0.0     | If the real part of e is negative, then the   |
+    | b = +/-oo    |         | limit is 0.                                   |
```
smichr, line 246:
```suggestion
    | b**e     with| 0.0     | If the real part of e is negative, then the   |
    | b = +/-oo and|         | limit is 0.                                   |
    | re(e) < 0    |         |                                               |
```

smichr:
> `1**zoo = nan`. Should that be defined as well?

I think so. But anything else (as the table says) does not have a well defined limit since zoo can have arbitrarily small re or im components which would leave a non vanishing result when raising a finite number to that particular value of zoo.

Considering the discussion in the email regarding `1**oo` it makes me more nervous to move forward with this. Does it open up too many corner cases where a unintentional result will leak out?

jksuom:
> t makes me more nervous to move forward with this.

That is my feeling, too. For a general (i.e. non-integer) exponent andf a finite, nonzero base `b`, the power `b**e`  is defined as `exp(log(b)*e)`. This should be valid also for limits, or accumulation values if limits do not exist. In particular, we would expect that `1**oo` be given by `exp(log(1)*oo)`, so the value should be determined by that of `log(1)*oo`.

smichr:
Let's close this with a commitment to let `1**oo` remain as nan since it is too easy to generate this in normal work.

asmeurer:
> Considering the discussion in the email regarding 1**oo it makes me more nervous to move forward with this. Does it open up too many corner cases where a unintentional result will leak out?

Is this referring to a discussion on the mailing list? 

I also wonder about `0**0`. It is conventionally defined as `1` from combinatorial considerations. In an analytic sense it is just as undefined as `1**oo`.

Issue #20608, opened 2020-12-15
Multivariate polynomial quotient rings


oscarbenjamin:
I want to construct a domain in two symbols c and s that behave as `sin(t)` and `cos(t)` with the requirement that `s**2 + c**2 = 1` so that powers of either `c` or `s` can be transformed into the other. Using `old_poly_ring` and a quotient ring seems to work:
```python
In [1]: s, c = symbols('s, c')

In [2]: K = QQ.old_poly_ring(s, c) / [s**2 + c**2 - 1]

In [3]: p = K.to_sympy

In [4]: c = K.convert(c)

In [5]: s = K.convert(s)

In [6]: p(c**2)
Out[6]: 
 2
c 

In [7]: p(s**2)
Out[7]: 
     2
1 - c 

In [8]: p(s**2 + c**2)
Out[8]: 1

In [9]: p(s**3)
Out[9]: 
   2      
- c ⋅s + s

In [10]: p(s**3 + s*c**2)
Out[10]: s
```
So far so good but now I get:
```python
In [11]: p(s**4)
Out[11]: 
 4    2    2
c  - c  + s 

In [12]: p(s**4 * c**2*s**2)
Out[12]: 
   8      6    4  2    4    2  2
- c  + 2⋅c  - c ⋅s  - c  + c ⋅s 
```
Here it seems that powers of s are not being fully reduced. The last example could be `s**2` or `1 - c**2`.

Is that just a bug in QuotientRing?

Is there a way to construct a domain like this in sympy?

jksuom:
> Is that just a bug in QuotientRing?

It looks like that is how weak normal forms work in sympy.polys.distributeddomains. I suspect that the source SCA is 'A Singular Introduction to Commutative Algebra' but I have currently no easy access to that book so I couldn't check the algorithms.

> Is there a way to construct a domain like this in sympy?

I would consider using `FiniteExtension`:
```
In [1]: from sympy.polys.agca.extensions import FiniteExtension                 

In [2]: s, c = symbols('s, c')                                                  

In [3]: K = FiniteExtension(Poly(s**2 + c**2 - 1, s))                           

In [4]: s = K.generator                                                         

In [5]: s**4                                                                    
Out[5]: c**4 - 2*c**2 + 1
````

oscarbenjamin:
Yes, that seems to be exactly what I wanted.

The docstring says that it is not a domain and so can't be used with Poly. Is there a reason for that or is it just not something that is implemented?

[...]

PR #21626, opened 2021-06-16
Make Matrix use DomainMatrix internally

oscarbenjamin:

This is a work in progress PR to change the internal representation of `Matrix` from the current `._mat` list of `Expr` to a `._rep` DomainMatrix attribute. The `DomainMatrix` still uses `Expr` internally via the `EXRAW` domain added in #21581.

#### Other comments

I am using this PR as a staging area and spinning out most changes to sub-PRs like #21632, #21633, #21637 to keep the diff small for the main change.

The internal `DomainMatrix` can use either the `ZZ`, `QQ` or `EXRAW` domains for elementary arithmetic like matrix multiplication. The `DomainMatrix` sparse implementation is used and can make other operations like `zeros`, `eye` etc more efficient.

Using domains like `ZZ[x,y]` for elementary operations is not backwards compatible because e.g. these domains will automatically expand any powers. However if the internal representation is a sparse DomainMatrix with domain EXRAW then we can efficiently convert to something like `ZZ[x,y]` where it is beneficial as part of an algorithm such as `charpoly`.

I have unified most of the implementation of DenseMatrix and SparseMatrix into a class RepMatrix with an inheritance scheme like:

```python
class RepMatrix(MatrixBase):
    # methods using DomainMatrix
    # Most implementation of matrix classes is here

class MutableRepMatrix(RepMatrix):
    # mutation methods

class ImmutableRepMatrix(RepMatrix, MatrixExpr):
    # a few specialisations for immutable

class DenseMatrix(RepMatrix):
   # Methods specific to dense

class SparseMatrix(RepMatrix):
    # methods specific to sparse
    
class MutableDenseMatrix(DenseMatrix, MutableRepMatrix):
    # 1 method (simplify) specific to this class

class MutableSparseMatrix(SparseMatrix, MutableRepMatrix):
    # 1 method (_new) specific to this case

class ImmutableDenseMatrix(DenseMatrix, ImmutableRepMatrix):
    # A few methods for immutable dense

class ImmutableSparseMatrix(SparseMatrix, ImmutableRepMatrix):
    # A few methods for immutable sparse
```
Internally all of these classes use DomainMatrix with its internal sparse representation (SDM) so there isn't much benefit in having SparseMatrix and DenseMatrix as separate classes. I've mostly unified the implementation of matrices into RepMatrix. Sparse and dense mutable matrices are reasonably unified in MutableRepMatrix. There are only small differences between ImmutableDenseMatrix and ImmutableSparseMatrix but since those differences concern the constructor arguments they are not immediately easy to resolve. Further unification of these classes is deferred for now.

TODO:

- [x] Detect ZZ, QQ domains instead of always using EXRAW.
- [x] Make use of DomainMatrix methods for speed.
- [x] Convert SparseMatrix to use DomainMatrix as well
- [x] Move most methods from DenseMatrix/SparseMatrix up to MatrixBase
- [ ] ~~Make the args of ImmutableDenseMatrix sparse~~
- [ ] ~~Convert ImmutableSparseMatrix args to the same form as for dense~~
- [x] Unify Mutable-dense/sparse.
- [ ] ~~Unify Immutable-dense/sparse~~
- [x] Identify methods that are inconsistent between dense/sparse
- [x] Run benchmarks
- [x] Re-add `_mat` and `_smat` with deprecation warnings

#### Release Notes

* matrices
    * The internal implementation of Matrix and other matrix classes (SparseMatrix etc) is now DomainMatrix. The ZZ and QQ domains are used for matrices with only integer or rational elements. Otherwise the new EXRAW domain is used. This should be backwards compatible although many internal methods and attributes are changed. At the time of this change the DomainMatrix routines are only used for addition and multiplication of matrices and some other simple low-level operations. Further changes will use DomainMatrix routines for operations like `rref`, `det`, `lu` etc and are expected to lead to big speedups for these computations. At this stage those big speedups are not realised but some basic operations such as indexing a matrix like `M[0, 0]` could potentially be slower. The new implementation can be much faster for most operations and is expected to lead to significant speed ups over the next few SymPy releases.


sympy/matrices/sparse.py
```diff
@@ -491,6 +491,10 @@ def _mat(self):
         in DenseMatrix use `_mat` directly to speed up operations."""
         return list(self)
 
+    @property
+    def _flat(self):
+        return list(self)
```
smichr, line 496: Is this better than making it an attribute that is set to `tuple(self)`? If I understand correctly, every call to `_flat` will return a unique, mutable copy of self's elements.

oscarbenjamin: Yes, returning a tuple might be better.

What isn't clear to me is how much we need to respect backwards compatibility wrt `_mat` which is technically private but is also used widely in the SymPy codebase and potentially other downstream code as well. It isn't possible to maintain full backwards compatibility because `_mat` can be modified directly as a way of mutating the Matrix. If we take out the possibility that someone is using `_mat` to mutate the Matrix then returning a list is best for backwards compatibility.

But what if someone is mutating the list? Should we raise a warning or make sure that the operation completely fails (which would be the case if a tuple was returned)?

oscarbenjamin: Actually for mutable matrices we need to construct the list/tuple each time. In the case of immutable matrices a tuple could be stored. Perhaps it could be a tuple in all cases even though it would have to be created fresh each time for mutable matrices.

The performance here should not matter much in the end since the intention is that DomainMatrix would end up handling each of the operations that currently accesses `._mat`.

oscarbenjamin:
A benchmark comparison against current master is below. The results are mixed but that is expected. The change here means that the Matrix class is internally more efficient but routines that operate on the matrix manually from outside are slower. For example getitem is slower and many routines that use getitem are slower as a result but the eventual plan would be to replace those routines with lower-level routines that operate directly on the more efficient internal representation.

The timings below would suggest that det is slower with this PR but actually this can make det much faster. The reason det appears slower is because it is still being computed using the inefficient high-level routine. You can see the difference in this PR between the high and low-level routines like this:
```python
In [4]: for n in range(20):
   ...:     M = randMatrix(n)
   ...:     %time M.det(); n    # <---- old high-level routine
   ...: 
CPU times: user 40 µs, sys: 1 µs, total: 41 µs
Wall time: 45.8 µs
CPU times: user 63 µs, sys: 1e+03 ns, total: 64 µs
Wall time: 69.1 µs
CPU times: user 241 µs, sys: 138 µs, total: 379 µs
Wall time: 274 µs
CPU times: user 413 µs, sys: 1e+03 ns, total: 414 µs
Wall time: 419 µs
CPU times: user 5.17 ms, sys: 176 µs, total: 5.34 ms
Wall time: 5.97 ms
CPU times: user 5.04 ms, sys: 69 µs, total: 5.11 ms
Wall time: 5.13 ms
CPU times: user 9.37 ms, sys: 171 µs, total: 9.54 ms
Wall time: 9.63 ms
CPU times: user 11 ms, sys: 151 µs, total: 11.1 ms
Wall time: 11.2 ms
CPU times: user 21.1 ms, sys: 344 µs, total: 21.4 ms
Wall time: 21.6 ms
CPU times: user 25.9 ms, sys: 278 µs, total: 26.1 ms
Wall time: 26.2 ms
CPU times: user 36.9 ms, sys: 672 µs, total: 37.6 ms
Wall time: 38.2 ms
CPU times: user 48.8 ms, sys: 679 µs, total: 49.5 ms
Wall time: 50 ms
CPU times: user 61.6 ms, sys: 658 µs, total: 62.3 ms
Wall time: 63 ms
CPU times: user 78.5 ms, sys: 600 µs, total: 79.1 ms
Wall time: 80 ms
CPU times: user 100 ms, sys: 840 µs, total: 101 ms
Wall time: 102 ms
CPU times: user 122 ms, sys: 979 µs, total: 123 ms
Wall time: 124 ms
CPU times: user 146 ms, sys: 970 µs, total: 147 ms
Wall time: 148 ms
CPU times: user 176 ms, sys: 1.31 ms, total: 177 ms
Wall time: 179 ms
CPU times: user 212 ms, sys: 1.14 ms, total: 213 ms
Wall time: 216 ms
CPU times: user 245 ms, sys: 1.39 ms, total: 246 ms
Wall time: 249 ms

In [5]: for n in range(20):
   ...:     M = randMatrix(n)
   ...:     %time M._rep.det(); n   # <--- internal (DomainMatrix) routine
   ...: 
CPU times: user 72 µs, sys: 1e+03 ns, total: 73 µs
Wall time: 78 µs
CPU times: user 78 µs, sys: 1e+03 ns, total: 79 µs
Wall time: 83.9 µs
CPU times: user 114 µs, sys: 250 µs, total: 364 µs
Wall time: 108 µs
CPU times: user 81 µs, sys: 1 µs, total: 82 µs
Wall time: 86.8 µs
CPU times: user 122 µs, sys: 0 ns, total: 122 µs
Wall time: 128 µs
CPU times: user 131 µs, sys: 1 µs, total: 132 µs
Wall time: 136 µs
CPU times: user 216 µs, sys: 5 µs, total: 221 µs
Wall time: 228 µs
CPU times: user 161 µs, sys: 1 µs, total: 162 µs
Wall time: 164 µs
CPU times: user 360 µs, sys: 5 µs, total: 365 µs
Wall time: 372 µs
CPU times: user 472 µs, sys: 3 µs, total: 475 µs
Wall time: 481 µs
CPU times: user 618 µs, sys: 1 µs, total: 619 µs
Wall time: 625 µs
CPU times: user 837 µs, sys: 5 µs, total: 842 µs
Wall time: 849 µs
CPU times: user 801 µs, sys: 9 µs, total: 810 µs
Wall time: 819 µs
CPU times: user 1.31 ms, sys: 6 µs, total: 1.32 ms
Wall time: 1.32 ms
CPU times: user 1.7 ms, sys: 20 µs, total: 1.72 ms
Wall time: 1.74 ms
CPU times: user 2.01 ms, sys: 6 µs, total: 2.01 ms
Wall time: 2.02 ms
CPU times: user 2.03 ms, sys: 7 µs, total: 2.04 ms
Wall time: 2.05 ms
CPU times: user 3.01 ms, sys: 13 µs, total: 3.02 ms
Wall time: 3.04 ms
CPU times: user 3.61 ms, sys: 8 µs, total: 3.62 ms
Wall time: 3.63 ms
CPU times: user 3.1 ms, sys: 34 µs, total: 3.13 ms
Wall time: 3.18 ms
```
What this shows is that the new routines can be hundreds of times faster for matrices of any significant size (here the maximum size is 19x19). However the new routines can be a little slower for some very small e.g. 1x1 matrices. The current benchmark suite mainly tests trivially tiny examples that are unlikely to be relevant to the performance of any real work usage so it might seem to show slowdowns for this PR.

This PR does not yet employ the low-level routines for any significant operations like `det`, `inv`,  etc but it represents the first step towards a more efficient implementation. It is expected that there will be small slowdowns here and that speedups from these changes will not yet show in the benchmarks. I want to merge this though and then work on operations like `det` one by one.

Benchmark comparison of the PR against master. Green means that the PR is slower than master (bad) but red means that the PR is faster than master (good). There are some slowdowns but note that they are all in operations that are relatively trivial and that are still quite fast either way.
```diff
$ asv compare --machine=it035770.local upstream/master 04e512627474345c74ac09de02db1089d551aaa4

All benchmarks:

       before           after         ratio
     [fc9acc8f]       [04e51262]
     <master>         <pr_matrix_domainmatrix>
          519±4ns         532±70ns     1.02  baseline.TimeBaseline.time_loop10
         324±30ms         339±60ms     1.05  baseline.TimeBaseline.time_loop10000000
          122±2ns          126±8ns     1.04  baseline.TimeBaseline.time_noop
      2.03±0.03μs      1.99±0.02μs     0.98  baseline.TimeBasicGCD.time_gcd
       4.29±0.3μs       3.82±0.5μs    ~0.89  core.arit.Time_Arit.time_Add_1x
       25.6±0.9μs         29.2±1μs    ~1.14  core.arit.Time_Arit.time_Add_x05
       3.65±0.1μs       3.74±0.3μs     1.02  core.arit.Time_Arit.time_Add_x1
      2.20±0.03μs      2.17±0.08μs     0.99  core.arit.Time_Arit.time_Add_xy
      1.40±0.02μs      1.44±0.02μs     1.02  core.arit.Time_Arit.time_Add_xyz
      4.93±0.04μs       5.05±0.1μs     1.02  core.arit.Time_Arit.time_Div_2y
       3.80±0.3μs       3.46±0.1μs     0.91  core.arit.Time_Arit.time_Div_xy
      2.17±0.03μs       2.23±0.2μs     1.03  core.arit.Time_Arit.time_Mul_xy
         2.22±1μs       1.42±0.1μs    ~0.64  core.arit.Time_Arit.time_Mul_xyz
         692±20μs        815±100μs    ~1.18  core.arit.Time_Arit.time_mod_pow
      2.21±0.03μs       2.27±0.1μs     1.03  core.arit.Time_Arit.time_neg
         999±20ns       1.15±0.3μs    ~1.15  core.arit.Time_Arit.time_pow100
      1.09±0.05μs         992±20ns    ~0.91  core.arit.Time_Arit.time_pow2
       31.2±0.8μs        31.6±10μs     1.01  core.arit.Time_Arit.time_pow_im
       10.6±0.4μs         11.3±1μs     1.07  core.arit.Time_assumption.time_Add_Mul_is_finite
         20.7±2μs         15.4±4μs    ~0.74  core.arit.Time_assumption.time_Add_Mul_is_integer
        56.3±20μs         43.5±1μs    ~0.77  core.arit.Time_assumption.time_ncmul
        69.6±30μs        47.6±10μs    ~0.68  core.arit.Time_assumption.time_ncpow
       16.3±0.7μs         16.4±2μs     1.01  core.expand.TimeExpand.time_expand_arit
         13.3±7μs         12.5±2μs     0.95  core.expand.TimeExpand.time_expand_complex_number_1
       18.6±0.4μs        20.3±10μs     1.09  core.expand.TimeExpand.time_expand_complex_number_2
        24.8±10μs       19.3±0.3μs    ~0.78  core.expand.TimeExpand.time_expand_frac
         49.1±3μs       48.6±0.7μs     0.99  core.expand.TimeExpand.time_expand_log
         8.53±1μs       7.15±0.2μs    ~0.84  core.expand.TimeExpand.time_expand_modulus
         14.9±2μs       13.7±0.2μs     0.92  core.expand.TimeExpand.time_expand_negative_integers_powers
       13.6±0.1μs       13.3±0.2μs     0.98  core.expand.TimeExpand.time_expand_no_log
         23.4±7μs       17.6±0.2μs    ~0.75  core.expand.TimeExpand.time_expand_no_multinomial
        23.3±20μs       19.5±0.2μs    ~0.84  core.expand.TimeExpand.time_expand_non_commutative
          802±6ns        912±200ns    ~1.14  core.expand.TimeExpand.time_expand_nothing_todo
         23.7±3μs       21.0±0.7μs    ~0.89  core.expand.TimeExpand.time_expand_power_base
         27.6±2μs         38.7±3μs    ~1.40  core.expand.TimeExpand.time_expnad_radicals
       1.98±0.4ms       1.71±0.2ms    ~0.87  cse.TimeCSE.time_cse((2, 10))
       1.93±0.4ms       2.29±0.9ms    ~1.18  cse.TimeCSE.time_cse((2, 11))
       1.88±0.7ms       1.61±0.8ms    ~0.85  cse.TimeCSE.time_cse((2, 8))
       1.60±0.3ms       1.60±0.2ms     1.00  cse.TimeCSE.time_cse((2, 9))
      1.33±0.02ms       1.42±0.1ms     1.07  cse.TimeCSE.time_cse((3, 8))
      1.37±0.03ms       1.66±0.3ms    ~1.22  cse.TimeCSE.time_cse((4, 8))
        2.55±0.4s        2.59±0.1s     1.01  dsolve.TimeDsolve01.time_dsolve
         190±30ms        166±0.8ms    ~0.88  integrate.TimeIntegration01.time_doit
        77.5±10ms       71.2±0.6ms     0.92  integrate.TimeIntegration01.time_doit_meijerg
         39.7±4ms       37.2±0.7ms     0.94  integrate.TimeIntegrationRisch01.time_doit(1)
         52.3±5ms       48.1±0.9ms     0.92  integrate.TimeIntegrationRisch01.time_doit(10)
         503±80ms          424±6ms    ~0.84  integrate.TimeIntegrationRisch01.time_doit(100)
         36.5±4ms       36.0±0.4ms     0.99  integrate.TimeIntegrationRisch01.time_doit_risch(1)
-        70.4±9ms       48.3±0.6ms     0.69  integrate.TimeIntegrationRisch01.time_doit_risch(10)
         525±70ms          427±7ms    ~0.81  integrate.TimeIntegrationRisch01.time_doit_risch(100)
      4.54±0.01μs      4.46±0.05μs     0.98  integrate.TimeIntegrationRisch02.time_doit(1)
          130±7ms          120±1ms     0.93  integrate.TimeIntegrationRisch02.time_doit(10)
        6.31±0.1s       6.08±0.03s     0.96  integrate.TimeIntegrationRisch02.time_doit(100)
       6.09±0.1μs       6.23±0.1μs     1.02  integrate.TimeIntegrationRisch02.time_doit_risch(1)
        115±0.9ms        116±0.7ms     1.00  integrate.TimeIntegrationRisch02.time_doit_risch(10)
       6.03±0.01s        6.54±0.2s     1.09  integrate.TimeIntegrationRisch02.time_doit_risch(100)
       47.9±0.8μs         51.2±1μs     1.07  integrate.TimeIntegrationRisch03.time_doit(1)
          241±1ms         290±20ms    ~1.20  integrate.TimeIntegrationRisch03.time_doit(10)
       2.99±0.05s        2.99±0.3s     1.00  integrate.TimeIntegrationRisch03.time_doit(100)
         696±20μs        799±100μs    ~1.15  integrate.TimeIntegrationRisch03.time_doit_risch(1)
          245±1ms         266±30ms     1.09  integrate.TimeIntegrationRisch03.time_doit_risch(10)
       2.96±0.04s        3.03±0.2s     1.03  integrate.TimeIntegrationRisch03.time_doit_risch(100)
         41.4±3ms         39.8±4ms     0.96  lambdify.TimeLambdifyCreation.time_lambdify_create(['math'])
         41.2±3ms         43.8±3ms     1.06  lambdify.TimeLambdifyCreation.time_lambdify_create(['numpy'])
       12.5±0.4μs         12.9±1μs     1.04  lambdify.TimeLambdifyEvaluate.time_lambdify_evaluate(['math'])
         74.7±2μs         77.2±7μs     1.03  lambdify.TimeLambdifyEvaluate.time_lambdify_evaluate(['numpy'])
            51.2M              51M     1.00  large_exprs.TimeLargeExpressionOperations.peakmem_jacobian_wrt_functions
            50.2M            50.4M     1.00  large_exprs.TimeLargeExpressionOperations.peakmem_jacobian_wrt_symbols
              50M            50.1M     1.00  large_exprs.TimeLargeExpressionOperations.peakmem_subs
       59.4±0.4ms         62.8±3ms     1.06  large_exprs.TimeLargeExpressionOperations.time_count_ops
       24.5±0.5ms         25.8±2ms     1.05  large_exprs.TimeLargeExpressionOperations.time_cse
       8.26±0.1ms       8.64±0.4ms     1.05  large_exprs.TimeLargeExpressionOperations.time_free_symbols
       1.35±0.01s       1.38±0.09s     1.02  large_exprs.TimeLargeExpressionOperations.time_jacobian_wrt_functions
       36.8±0.3ms       38.2±0.7ms     1.04  large_exprs.TimeLargeExpressionOperations.time_jacobian_wrt_symbols
       1.35±0.02s       1.36±0.01s     1.01  large_exprs.TimeLargeExpressionOperations.time_manual_jacobian_wrt_functions
          414±8ms          420±7ms     1.02  large_exprs.TimeLargeExpressionOperations.time_subs
       3.80±0.04s       3.80±0.07s     1.00  logic.LogicSuite.time_dpll
          229±5ms          235±6ms     1.03  logic.LogicSuite.time_dpll2
       14.5±0.4ms       14.3±0.2ms     0.99  logic.LogicSuite.time_load_file
+      3.52±0.1ms       4.35±0.1ms     1.24  matrices.TimeDiagonalEigenvals.time_eigenvals
       7.59±0.1μs       7.53±0.2μs     0.99  matrices.TimeMatrixExpression.time_Add
       42.8±0.5μs       43.7±0.5μs     1.02  matrices.TimeMatrixExpression.time_MatAdd
       26.3±0.4ms       26.0±0.4ms     0.99  matrices.TimeMatrixExpression.time_MatAdd_doit
          403±6μs          418±8μs     1.04  matrices.TimeMatrixExpression.time_MatMul
       25.5±0.2ms       25.7±0.2ms     1.01  matrices.TimeMatrixExpression.time_MatMul_doit
       7.41±0.1μs      7.52±0.09μs     1.01  matrices.TimeMatrixExpression.time_Mul
+         130±3μs          217±4μs     1.67  matrices.TimeMatrixGetItem.time_ImmutableDenseMatrix_getitem
+         129±2μs          212±4μs     1.64  matrices.TimeMatrixGetItem.time_ImmutableSparseMatrix_getitem
+         133±6μs          216±3μs     1.62  matrices.TimeMatrixGetItem.time_MutableDenseMatrix_getitem
+         129±1μs          211±2μs     1.63  matrices.TimeMatrixGetItem.time_MutableSparseMatrix_getitem
+      9.91±0.1ms       12.5±0.4ms     1.26  matrices.TimeMatrixPower.time_Case1
          116±3ms          123±2ms     1.06  matrices.TimeMatrixPower.time_Case2
          114±2ms          120±2ms     1.05  matrices.TimeMatrixPower.time_Case3
+     7.32±0.07ms       9.85±0.2ms     1.35  physics.mechanics.kane.KanesMethodMassSpringDamper.time_kanesmethod_mass_spring_damper
       22.1±0.3ms       23.2±0.6ms     1.05  physics.mechanics.lagrange.LagrangesMethodMassSpringDamper.time_lagrangesmethod_mass_spring_damper
       6.88±0.02s       6.92±0.03s     1.01  polygon.PolygonArbitraryPoint.time_bench01
          303±4μs          302±4μs     1.00  polygon.PolygonAttributes.time_area
      1.59±0.02ms      1.61±0.05ms     1.01  polygon.PolygonAttributes.time_centroid
+     5.40±0.04ms       6.71±0.1ms     1.24  polygon.PolygonAttributes.time_create
      1.21±0.02ms      1.20±0.01ms     0.99  polygon.PolygonAttributes.time_perimeter
      1.15±0.03ms      1.15±0.02ms     1.00  polygon.PolygonAttributes.time_sides
+      1.09±0.02s       1.36±0.01s     1.25  polygon.PolygonCutSection.time_bench01
         90.6±9ms         117±10ms    ~1.29  polygon.PolygonDistance.time_bench01
+         137±6ms          188±2ms     1.37  polygon.PolygonEnclosesPoint.time_bench01
+         843±8ms       1.06±0.01s     1.26  polygon.PolygonFirstMoment.time_bench01
+      46.4±0.4ms         55.0±1ms     1.19  polygon.PolygonIsConvex.time_bench01
      5.84±0.08ms       5.85±0.1ms     1.00  polygon.PolygonPolarSecondMoment.time_bench01
       5.80±0.1ms      5.75±0.06ms     0.99  polygon.PolygonSecondMoment.time_bench01
       8.09±0.1ms       8.12±0.1ms     1.00  polygon.PolygonSectionModulus.time_bench01
         92.9±4μs         90.5±2μs     0.97  polys.TimePolyManyGens.time_create_poly(1)
          137±2μs          139±4μs     1.01  polys.TimePolyManyGens.time_create_poly(10)
      1.27±0.01ms      1.27±0.04ms     1.00  polys.TimePolyManyGens.time_create_poly(100)
       22.3±0.4ms       22.3±0.4ms     1.00  polys.TimePolyManyGens.time_create_poly(500)
      2.90±0.05μs      2.95±0.04μs     1.02  polys.TimePolyManyGens.time_is_linear(1)
       33.5±0.4μs         33.7±1μs     1.00  polys.TimePolyManyGens.time_is_linear(10)
         973±10μs      1.00±0.04ms     1.03  polys.TimePolyManyGens.time_is_linear(100)
       21.1±0.2ms       21.2±0.3ms     1.00  polys.TimePolyManyGens.time_is_linear(500)
-         127±2μs         69.2±4μs     0.54  solve.TimeMatrixArithmetic.time_dense_add(10, 0)
+         133±1μs          191±4μs     1.43  solve.TimeMatrixArithmetic.time_dense_add(10, 5)
+      17.3±0.2μs         25.0±1μs     1.44  solve.TimeMatrixArithmetic.time_dense_add(3, 0)
+      21.1±0.5μs         42.7±2μs     2.02  solve.TimeMatrixArithmetic.time_dense_add(3, 5)
       25.6±0.2μs       27.2±0.3μs     1.06  solve.TimeMatrixArithmetic.time_dense_add(4, 0)
+      30.6±0.3μs         53.3±1μs     1.74  solve.TimeMatrixArithmetic.time_dense_add(4, 5)
-        52.5±2μs       39.0±0.4μs     0.74  solve.TimeMatrixArithmetic.time_dense_add(6, 0)
+      55.8±0.9μs         88.7±2μs     1.59  solve.TimeMatrixArithmetic.time_dense_add(6, 5)
-     1.78±0.02ms          302±7μs     0.17  solve.TimeMatrixArithmetic.time_dense_multiply(10, 0)
      2.08±0.03ms      2.24±0.03ms     1.07  solve.TimeMatrixArithmetic.time_dense_multiply(10, 5)
-      69.9±0.8μs       30.8±0.6μs     0.44  solve.TimeMatrixArithmetic.time_dense_multiply(3, 0)
+         110±3μs          186±3μs     1.69  solve.TimeMatrixArithmetic.time_dense_multiply(3, 5)
-         145±1μs       39.4±0.5μs     0.27  solve.TimeMatrixArithmetic.time_dense_multiply(4, 0)
+         226±6μs          309±6μs     1.37  solve.TimeMatrixArithmetic.time_dense_multiply(4, 5)
-         431±6μs         86.8±2μs     0.20  solve.TimeMatrixArithmetic.time_dense_multiply(6, 0)
+         587±6μs          700±7μs     1.19  solve.TimeMatrixArithmetic.time_dense_multiply(6, 5)
+         118±2μs          159±9μs     1.35  solve.TimeMatrixOperations.time_det(3, 0)
          229±4μs         250±10μs     1.09  solve.TimeMatrixOperations.time_det(3, 2)
         220±20μs          237±6μs     1.07  solve.TimeMatrixOperations.time_det(3, 5)
+        861±20μs      1.42±0.03ms     1.65  solve.TimeMatrixOperations.time_det(4, 0)
       5.54±0.2ms       6.21±0.2ms    ~1.12  solve.TimeMatrixOperations.time_det(4, 2)
      1.64±0.06ms      2.06±0.04ms    ~1.26  solve.TimeMatrixOperations.time_det(4, 5)
          121±7μs        155±0.8μs    ~1.28  solve.TimeMatrixOperations.time_det_bareiss(3, 0)
+         221±3μs          253±5μs     1.14  solve.TimeMatrixOperations.time_det_bareiss(3, 2)
+         213±4μs          246±9μs     1.16  solve.TimeMatrixOperations.time_det_bareiss(3, 5)
+        860±10μs      1.38±0.03ms     1.60  solve.TimeMatrixOperations.time_det_bareiss(4, 0)
       5.43±0.1ms       6.01±0.2ms    ~1.11  solve.TimeMatrixOperations.time_det_bareiss(4, 2)
       1.67±0.1ms      2.00±0.05ms    ~1.20  solve.TimeMatrixOperations.time_det_bareiss(4, 5)
+         122±2μs          152±3μs     1.25  solve.TimeMatrixOperations.time_det_berkowitz(3, 0)
          231±4μs          252±7μs     1.09  solve.TimeMatrixOperations.time_det_berkowitz(3, 2)
          219±3μs          235±7μs     1.07  solve.TimeMatrixOperations.time_det_berkowitz(3, 5)
+     1.13±0.05ms      1.76±0.02ms     1.55  solve.TimeMatrixOperations.time_det_berkowitz(4, 0)
+     1.34±0.02ms      2.44±0.03ms     1.82  solve.TimeMatrixOperations.time_det_berkowitz(4, 2)
+     1.40±0.03ms      2.56±0.05ms     1.83  solve.TimeMatrixOperations.time_det_berkowitz(4, 5)
+         384±6μs         693±20μs     1.81  solve.TimeMatrixOperations.time_rank(3, 0)
       67.1±0.4ms       67.7±0.8ms     1.01  solve.TimeMatrixOperations.time_rank(3, 2)
       34.2±0.9ms         35.8±2ms     1.05  solve.TimeMatrixOperations.time_rank(3, 5)
+        618±10μs      1.07±0.03ms     1.72  solve.TimeMatrixOperations.time_rank(4, 0)
       73.3±0.6ms       74.7±0.6ms     1.02  solve.TimeMatrixOperations.time_rank(4, 2)
       75.0±0.2ms         76.8±1ms     1.02  solve.TimeMatrixOperations.time_rank(4, 5)
+         179±3μs          275±5μs     1.54  solve.TimeMatrixOperations.time_rref(3, 0)
       35.8±0.6ms       36.4±0.3ms     1.02  solve.TimeMatrixOperations.time_rref(3, 2)
       69.4±0.8ms       69.8±0.7ms     1.01  solve.TimeMatrixOperations.time_rref(3, 5)
+         372±7μs         519±10μs     1.39  solve.TimeMatrixOperations.time_rref(4, 0)
       90.6±0.7ms         90.6±1ms     1.00  solve.TimeMatrixOperations.time_rref(4, 2)
        102±0.4ms        104±0.4ms     1.02  solve.TimeMatrixOperations.time_rref(4, 5)
       53.0±0.6ms         54.6±1ms     1.03  solve.TimeMatrixPositiveDefinite.time_A_is_positive_definite
         54.6±2ms       52.8±0.5ms     0.97  solve.TimeMatrixPositiveDefinite.time_A_sym_is_positive_definite
          338±3ms          338±5ms     1.00  solve.TimeMatrixSolve.time_solve('ADJ')
          826±7ms          834±8ms     1.01  solve.TimeMatrixSolve.time_solve('GE')
+     1.08±0.02ms      1.67±0.04ms     1.55  solve.TimeMatrixSolve.time_solve('LU')
          248±1ms          254±4ms     1.02  solve.TimeMatrixSolve.time_solve_sym('ADJ')
          687±5ms         683±10ms     0.99  solve.TimeMatrixSolve.time_solve_sym('GE')
+     1.06±0.02ms       1.63±0.1ms     1.55  solve.TimeMatrixSolve.time_solve_sym('LU')
+     1.85±0.02ms      2.24±0.06ms     1.21  solve.TimeMatrixSolve2.time_LDLsolve
+     1.77±0.03ms      1.96±0.04ms     1.11  solve.TimeMatrixSolve2.time_cholesky_solve
+     1.08±0.03ms      1.64±0.09ms     1.52  solve.TimeMatrixSolve2.time_lusolve
+        691±50μs         977±30μs     1.41  solve.TimeMatrixSolvePyDyFast.time_LUsolve(1)
+     1.06±0.08ms      1.61±0.04ms     1.52  solve.TimeMatrixSolvePyDyFast.time_LUsolve(2)
+     1.78±0.04ms      2.66±0.05ms     1.50  solve.TimeMatrixSolvePyDyFast.time_LUsolve(3)
+     1.01±0.01ms      1.41±0.03ms     1.39  solve.TimeMatrixSolvePyDySlow.time_LDLsolve(1)
+        911±20μs      1.20±0.03ms     1.32  solve.TimeMatrixSolvePyDySlow.time_cholesky_solve(1)
         62.1±1ms         66.0±5ms     1.06  solve.TimeMatrixSolvePyDySlow.time_linsolve(1)
         62.7±1ms         65.4±1ms     1.04  solve.TimeMatrixSolvePyDySlow.time_solve(1)
          168±2μs          170±4μs     1.01  solve.TimeRationalSystem.time_linsolve(1)
      3.87±0.06ms      3.93±0.05ms     1.01  solve.TimeRationalSystem.time_linsolve(10)
         446±10μs          441±4μs     0.99  solve.TimeRationalSystem.time_linsolve(3)
         990±60μs          930±7μs     0.94  solve.TimeRationalSystem.time_linsolve(5)
      1.30±0.03ms      1.40±0.03ms     1.07  solve.TimeRationalSystem.time_solve(1)
       44.5±0.4ms       46.2±0.3ms     1.04  solve.TimeRationalSystem.time_solve(10)
      5.62±0.06ms       5.85±0.1ms     1.04  solve.TimeRationalSystem.time_solve(3)
       13.1±0.3ms       13.3±0.2ms     1.01  solve.TimeRationalSystem.time_solve(5)
         604±20μs         590±10μs     0.98  solve.TimeRationalSystemSymbol.time_linsolve(1)
      2.22±0.06ms       2.42±0.2ms     1.09  solve.TimeRationalSystemSymbol.time_linsolve(3)
      5.55±0.07ms       5.63±0.1ms     1.01  solve.TimeRationalSystemSymbol.time_linsolve(5)
      1.37±0.03ms      1.55±0.05ms    ~1.13  solve.TimeRationalSystemSymbol.time_solve(1)
       6.00±0.1ms         7.96±2ms    ~1.33  solve.TimeRationalSystemSymbol.time_solve(3)
       13.7±0.3ms       14.5±0.2ms     1.05  solve.TimeRationalSystemSymbol.time_solve(5)
       39.4±0.7ms         42.4±3ms     1.08  solve.TimeSolve01.time_solve
       39.6±0.3ms         41.1±2ms     1.04  solve.TimeSolve01.time_solve_nocheck
              n/a              n/a      n/a  solve.TimeSolveLinSys.time_solve_lin_sys
              n/a              n/a      n/a  solve.TimeSolveLinSys.time_to_expr_eqs
              n/a              n/a      n/a  solve.TimeSolveLinSys.time_verify_sol
              n/a              n/a      n/a  solve.TimeSolveLinSys10x8.time_solve_lin_sys
              n/a              n/a      n/a  solve.TimeSolveLinSys10x8.time_to_expr_eqs
              n/a              n/a      n/a  solve.TimeSolveLinSys10x8.time_verify_sol
              n/a              n/a      n/a  solve.TimeSolveLinSys165x165.time_solve_lin_sys
              n/a              n/a      n/a  solve.TimeSolveLinSys165x165.time_to_expr_eqs
              n/a              n/a      n/a  solve.TimeSolveLinSys165x165.time_verify_sol
         55.1±4ms       54.7±0.8ms     0.99  solve.TimeSolveLinSys189x49.time_solve_lin_sys
         62.3±3ms       60.5±0.6ms     0.97  solve.TimeSolveLinSys189x49.time_to_expr_eqs
          157±1ms          155±1ms     0.99  solve.TimeSolveLinSys189x49.time_verify_sol
       16.2±0.1ms       16.5±0.1ms     1.02  solve.TimeSolveSparsePolySystem.time_solve(1)
       31.9±0.4ms         34.9±3ms     1.09  solve.TimeSolveSparsePolySystem.time_solve(2)
       48.6±0.5ms       48.8±0.6ms     1.00  solve.TimeSolveSparsePolySystem.time_solve(3)
+     1.76±0.03ms      2.17±0.03ms     1.23  solve.TimeSparseSystem.time_linear_eq_to_matrix(10)
+     5.65±0.06ms      6.69±0.09ms     1.18  solve.TimeSparseSystem.time_linear_eq_to_matrix(20)
+      11.7±0.2ms       14.0±0.2ms     1.20  solve.TimeSparseSystem.time_linear_eq_to_matrix(30)
+      20.1±0.1ms       25.1±0.2ms     1.25  solve.TimeSparseSystem.time_linsolve_Aaug(10)
+      51.4±0.3ms       68.1±0.4ms     1.33  solve.TimeSparseSystem.time_linsolve_Aaug(20)
+      95.4±0.5ms          135±1ms     1.41  solve.TimeSparseSystem.time_linsolve_Aaug(30)
       20.0±0.5ms       24.9±0.6ms    ~1.25  solve.TimeSparseSystem.time_linsolve_Ab(10)
+      51.7±0.4ms         68.8±2ms     1.33  solve.TimeSparseSystem.time_linsolve_Ab(20)
+      94.1±0.5ms          139±5ms     1.47  solve.TimeSparseSystem.time_linsolve_Ab(30)
      1.14±0.01ms      1.12±0.02ms     0.99  solve.TimeSparseSystem.time_linsolve_eqs(10)
      2.06±0.04ms      2.04±0.03ms     0.99  solve.TimeSparseSystem.time_linsolve_eqs(20)
      3.04±0.04ms      3.08±0.08ms     1.01  solve.TimeSparseSystem.time_linsolve_eqs(30)
       26.8±0.5ms       28.7±0.3ms     1.07  solve.TimeSparseSystem.time_solve(10)
       53.1±0.4ms       56.8±0.5ms     1.07  solve.TimeSparseSystem.time_solve(20)
       80.1±0.2ms       85.1±0.7ms     1.06  solve.TimeSparseSystem.time_solve(30)
          179±1ms          178±2ms     1.00  sum.TimeSum.time_doit
```

oscarbenjamin:
I was intending to squash the commits but I now think that they should not be squashed or rearranged. This PR makes major changes to matrices but the commits are organised as incremental changes with each step representing a mostly working state of the code. Preserving these commits would be useful for bisecting bugs or regressions.

This PR represents the first step in a major improvement in the performance of the main matrix subclasses by basing them on the more efficient DomainMatrix class. In this PR the change is that the internal representation of Matrix etc is made to be DomainMatrix. That change will speed up some elementary operations but will also slow down some others. Future changes building on this can speed up many common routines but those are not implemented in this PR.

The most noticeable performance improvement with this PR is for sparse matrices. This PR makes Matrix use DomainMatrix internally and specifically uses the SDM sparse internal representation of DomainMatrix. That means that basic matrix arithmetic is made more efficient for sparse matrices.

Example with the PR:
```python
In [1]: M = randMatrix(100, percent=1)

In [2]: %time ok = M*M
CPU times: user 458 µs, sys: 29 µs, total: 487 µs
Wall time: 496 µs
```
Timings with master:
```python
In [2]: %time ok = M*M
CPU times: user 1.75 s, sys: 16 ms, total: 1.77 s
Wall time: 1.78 s
```
That's a speed difference of more than 1000x.

The other performance improvement is for matrices of integer/rational.

Example with the PR:
```python
In [1]: M = randMatrix(100)

In [2]: %time ok = M*M
CPU times: user 289 ms, sys: 2.25 ms, total: 291 ms
Wall time: 290 ms
```
Timing for master:
```python
In [5]: M = randMatrix(100)

In [6]: %time ok = M*M
CPU times: user 12.9 s, sys: 284 ms, total: 13.1 s
Wall time: 13.5 s
```
That's a 50x speed difference for this case.

oscarbenjamin:
This PR is ready for review.

I would like to get this change in before the next release (and ideally give it good time to sit on master before release).

oscarbenjamin:
> I would like to get this change in before the next release (and ideally give it good time to sit on master before release).

If anyone intends to review this then please say so.

It is a big PR and it will take time to review (which is fine) but without indication that anyone will review it I will at some point just merge it.

oscarbenjamin:

moorepants:
Are the green rows slowdowns?

moorepants:
I see it is in your comment. For the matrix solve commands, this slowdown is very bad as the size of the matrices get larger. For example with LUSolve you have an O(n^3) relationship based on size of the matrix. I'm -1 on adding that kind of major slowdown.

moorepants:
And, in general, there seems to be more green rows that red rows! So how is this an overall performance improvement?

oscarbenjamin:
> And, in general, there seems to be more green rows that red rows! So how is this an overall performance improvement?

This is just the start of a process. The new implementation is much faster but we have to switch various functions to use it internally. Right now this PR changes it so that the Matrix class uses DomainMatrix as its internal representation but does not use the DomainMatrix routines for anything except basic things like addition and multiplication. This is already quite a big diff though so I don't want to change all of the other things at the same time. Also some of the changes would potentially lead to different output being returned and in this PR I am trying to to change any behaviour of Matrix.

The goal here is to switch the internal implementation while making as few other changes as possible and hopefully without causing any major slowdown although there can be small slowdowns for some relatively trivial operation.

> For example with LUSolve you have an O(n^3) relationship based on size of the matrix. I'm -1 on adding that kind of major slowdown.

The benchmarks are unfortunately not always representative. I am interested in some kind of genuine slowdown and the benchmarks can help identify those but those particular changes are not indicative of what this change really means for LUsolve performance in sympy.

For example the LUsolve benchmark is basically this (running with master):
```python
In [1]: M = Matrix(symbols('x:9')).reshape(3, 3)

In [2]: b = Matrix(symbols('b:3')).reshape(3, 1)

In [3]: M
Out[3]: 
⎡x₀  x₁  x₂⎤
⎢          ⎥
⎢x₃  x₄  x₅⎥
⎢          ⎥
⎣x₆  x₇  x₈⎦

In [4]: b
Out[4]: 
⎡b₀⎤
⎢  ⎥
⎢b₁⎥
⎢  ⎥
⎣b₂⎦

In [5]: %time ok = M.LUsolve(b)
CPU times: user 151 ms, sys: 1.81 ms, total: 153 ms
Wall time: 151 ms

In [6]: %time ok = M.LUsolve(b)
CPU times: user 2.91 ms, sys: 0 ns, total: 2.91 ms
Wall time: 2.92 ms
```
The second run is much faster because of the cache and that's the time that the benchmarks report. If I run the same with the PR I get:
```python
In [5]: %time ok = M.LUsolve(b)
CPU times: user 157 ms, sys: 2.95 ms, total: 160 ms
Wall time: 158 ms

In [6]: %time ok = M.LUsolve(b)
CPU times: user 3.91 ms, sys: 0 ns, total: 3.91 ms
Wall time: 3.92 ms
```
So you see that the actual time to compute the LU decomposition (the first call to LUsolve) takes basically the same amount of time. When the result is cached then some small difference in the time to index a matrix or something comes into it. I can probably optimise that just to make the benchmark look better but this is not how we will actually speed up the LU decomposition.

The internal DomainMatrix class has its own `lu_solve` function but we are not using that here yet:
```python
In [7]: %time ok = M._rep.lu_solve(b._rep)
CPU times: user 796 µs, sys: 0 ns, total: 796 µs
Wall time: 804 µs
```
That's how long it takes after caching so you can see that it is faster than what is currently timed by the benchmarks. That is not what is really supposed to speed things up with DomainMatrix though which is when we use a specific domain. That's how we can get it to be faster for the first call to LUsolve which is the one that we care about:
```python
In [1]: M = Matrix(symbols('x:9')).reshape(3, 3)

In [2]: b = Matrix(symbols('b:3')).reshape(3, 1)

In [3]: from sympy.polys.matrices import DomainMatrix

In [4]: dM = DomainMatrix.from_Matrix(M, field=True)

In [5]: db = DomainMatrix.from_Matrix(b, field=True)

In [6]: dM, db = dM.unify(db)

In [7]: %time ok = dM.lu_solve(db)
CPU times: user 80.3 ms, sys: 3.91 ms, total: 84.2 ms
Wall time: 82.4 ms
```
That's a 2x speed up for the result that actually matters. It also produces output in a canonicalised simplified form:
```python
In [9]: dM.lu_solve(db).to_Matrix()
Out[9]: 
⎡b₀⋅x₄⋅x₈ - b₀⋅x₅⋅x₇ - b₁⋅x₁⋅x₈ + b₁⋅x₂⋅x₇ + b₂⋅x₁⋅x₅ - b₂⋅x₂⋅x₄ ⎤
⎢─────────────────────────────────────────────────────────────── ⎥
⎢x₀⋅x₄⋅x₈ - x₀⋅x₅⋅x₇ - x₁⋅x₃⋅x₈ + x₁⋅x₅⋅x₆ + x₂⋅x₃⋅x₇ - x₂⋅x₄⋅x₆ ⎥
⎢                                                                ⎥
⎢-b₀⋅x₃⋅x₈ + b₀⋅x₅⋅x₆ + b₁⋅x₀⋅x₈ - b₁⋅x₂⋅x₆ - b₂⋅x₀⋅x₅ + b₂⋅x₂⋅x₃⎥
⎢────────────────────────────────────────────────────────────────⎥
⎢x₀⋅x₄⋅x₈ - x₀⋅x₅⋅x₇ - x₁⋅x₃⋅x₈ + x₁⋅x₅⋅x₆ + x₂⋅x₃⋅x₇ - x₂⋅x₄⋅x₆ ⎥
⎢                                                                ⎥
⎢b₀⋅x₃⋅x₇ - b₀⋅x₄⋅x₆ - b₁⋅x₀⋅x₇ + b₁⋅x₁⋅x₆ + b₂⋅x₀⋅x₄ - b₂⋅x₁⋅x₃ ⎥
⎢─────────────────────────────────────────────────────────────── ⎥
⎣x₀⋅x₄⋅x₈ - x₀⋅x₅⋅x₇ - x₁⋅x₃⋅x₈ + x₁⋅x₅⋅x₆ + x₂⋅x₃⋅x₇ - x₂⋅x₄⋅x₆ ⎦
```
(Compare that with the output of `M.LUsolve(b)`)

We can also improve this much more by using fraction-free LU solve which is not yet implemented in DomainMatrix although I have written the code for it but not yet shared it anywhere. It needs work but we can bring significant speed ups to the symbolic matrix case with that.

For the time being you can see how DomainMatrix scales up for something like LUsolve with integer/rational if you want to see `O(n^3)` type behaviour:
```python
n [20]: for n in [1, 2, 5, 10, 20, 50, 100]:
    ...:     M = randMatrix(n, n)
    ...:     b = randMatrix(n, 1)
    ...:     print('n =', n)
    ...:     %time ok = M._rep.to_field().lu_solve(b._rep.to_field()) # DomainMatrix routine
    ...: 
n = 1
CPU times: user 363 µs, sys: 0 ns, total: 363 µs
Wall time: 309 µs
n = 2
CPU times: user 419 µs, sys: 0 ns, total: 419 µs
Wall time: 307 µs
n = 5
CPU times: user 622 µs, sys: 21 µs, total: 643 µs
Wall time: 593 µs
n = 10
CPU times: user 1.38 ms, sys: 6 µs, total: 1.39 ms
Wall time: 1.39 ms
n = 20
CPU times: user 6.49 ms, sys: 0 ns, total: 6.49 ms
Wall time: 6.5 ms
n = 50
CPU times: user 80.6 ms, sys: 0 ns, total: 80.6 ms
Wall time: 80.6 ms
n = 100
CPU times: user 991 ms, sys: 0 ns, total: 991 ms
Wall time: 992 ms

In [21]: for n in [1, 2, 5, 10, 20, 50, 100]:
    ...:     M = randMatrix(n, n)
    ...:     b = randMatrix(n, 1)
    ...:     print('n =', n)
    ...:     %time ok = M.LUsolve(b) # Current routine
    ...: 
n = 1
CPU times: user 2.36 ms, sys: 11 µs, total: 2.37 ms
Wall time: 2.09 ms
n = 2
CPU times: user 3.97 ms, sys: 30 µs, total: 4 ms
Wall time: 3.53 ms
n = 5
CPU times: user 16.6 ms, sys: 0 ns, total: 16.6 ms
Wall time: 16.6 ms
n = 10
CPU times: user 73.8 ms, sys: 0 ns, total: 73.8 ms
Wall time: 73.8 ms
n = 20
CPU times: user 334 ms, sys: 0 ns, total: 334 ms
Wall time: 334 ms
n = 50
CPU times: user 4.72 s, sys: 0 ns, total: 4.72 s
Wall time: 4.72 s
n = 100
CPU times: user 38.3 s, sys: 0 ns, total: 38.3 s
Wall time: 38.3 s
```
Notice that the timings for DomainMatrix and Matrix do not necessarily scale proportionality as you might expect for large n e.g. for n=2 we have a speed difference of 10x but for n=10 it's 80x so you really can't just expect results from very small sizes to scale up. Also many Matrix routines do not have the ekind of `n^3` behaviour that you would expect e.g. the `inv` method is more like `2^n` exponential:
```python
In [42]: for n in range(20):
    ...:     M = randMatrix(n, n)
    ...:     print('n =', n)
    ...:     %time M.inv()
    ...: 
n = 0
CPU times: user 641 µs, sys: 0 ns, total: 641 µs
Wall time: 648 µs
n = 1
CPU times: user 1.55 ms, sys: 0 ns, total: 1.55 ms
Wall time: 1.33 ms
n = 2
CPU times: user 3.13 ms, sys: 0 ns, total: 3.13 ms
Wall time: 2.8 ms
n = 3
CPU times: user 5.72 ms, sys: 0 ns, total: 5.72 ms
Wall time: 5.64 ms
n = 4
CPU times: user 8.82 ms, sys: 998 µs, total: 9.82 ms
Wall time: 9.54 ms
n = 5
CPU times: user 12.9 ms, sys: 0 ns, total: 12.9 ms
Wall time: 12.9 ms
n = 6
CPU times: user 18.5 ms, sys: 0 ns, total: 18.5 ms
Wall time: 18.5 ms
n = 7
CPU times: user 24.1 ms, sys: 936 µs, total: 25 ms
Wall time: 25 ms
n = 8
CPU times: user 29.2 ms, sys: 0 ns, total: 29.2 ms
Wall time: 29.2 ms
n = 9
CPU times: user 35.7 ms, sys: 0 ns, total: 35.7 ms
Wall time: 35.7 ms
n = 10
CPU times: user 92.7 ms, sys: 2 µs, total: 92.7 ms
Wall time: 92.7 ms
n = 11
CPU times: user 56.6 ms, sys: 0 ns, total: 56.6 ms
Wall time: 56.7 ms
n = 12
CPU times: user 90.8 ms, sys: 0 ns, total: 90.8 ms
Wall time: 90.8 ms
n = 13
CPU times: user 156 ms, sys: 0 ns, total: 156 ms
Wall time: 156 ms
n = 14
CPU times: user 397 ms, sys: 0 ns, total: 397 ms
Wall time: 397 ms
n = 15
CPU times: user 654 ms, sys: 998 µs, total: 655 ms
Wall time: 655 ms
n = 16
CPU times: user 3.49 s, sys: 2 ms, total: 3.49 s
Wall time: 3.49 s
n = 17
CPU times: user 13.5 s, sys: 7.99 ms, total: 13.5 s
Wall time: 13.5 s
n = 18
CPU times: user 36 s, sys: 16 ms, total: 36 s
Wall time: 36 s
n = 19
CPU times: user 2min 38s, sys: 30.9 ms, total: 2min 38s
Wall time: 2min 38s
```
The DomainMatrix inv can be improved by using fraction-free LU as I mentioned but ever for now it's much faster and scales much more like `n^3`:
```python
In [43]: for n in range(20):
    ...:     M = randMatrix(n, n)
    ...:     print('n =', n)
    ...:     %time M._rep.to_field().inv()
    ...: 
n = 0
CPU times: user 115 µs, sys: 1e+03 ns, total: 116 µs
Wall time: 123 µs
n = 1
CPU times: user 150 µs, sys: 0 ns, total: 150 µs
Wall time: 157 µs
n = 2
CPU times: user 197 µs, sys: 0 ns, total: 197 µs
Wall time: 181 µs
n = 3
CPU times: user 267 µs, sys: 0 ns, total: 267 µs
Wall time: 234 µs
n = 4
CPU times: user 310 µs, sys: 0 ns, total: 310 µs
Wall time: 317 µs
n = 5
CPU times: user 446 µs, sys: 0 ns, total: 446 µs
Wall time: 432 µs
n = 6
CPU times: user 643 µs, sys: 0 ns, total: 643 µs
Wall time: 626 µs
n = 7
CPU times: user 832 µs, sys: 0 ns, total: 832 µs
Wall time: 838 µs
n = 8
CPU times: user 1.06 ms, sys: 0 ns, total: 1.06 ms
Wall time: 1.06 ms
n = 9
CPU times: user 1.35 ms, sys: 0 ns, total: 1.35 ms
Wall time: 1.35 ms
n = 10
CPU times: user 1.77 ms, sys: 0 ns, total: 1.77 ms
Wall time: 1.78 ms
n = 11
CPU times: user 2.2 ms, sys: 0 ns, total: 2.2 ms
Wall time: 2.21 ms
n = 12
CPU times: user 2.82 ms, sys: 0 ns, total: 2.82 ms
Wall time: 2.83 ms
n = 13
CPU times: user 3.39 ms, sys: 1e+03 ns, total: 3.39 ms
Wall time: 3.39 ms
n = 14
CPU times: user 3.93 ms, sys: 0 ns, total: 3.93 ms
Wall time: 3.94 ms
n = 15
CPU times: user 4.84 ms, sys: 0 ns, total: 4.84 ms
Wall time: 4.84 ms
n = 16
CPU times: user 5.54 ms, sys: 0 ns, total: 5.54 ms
Wall time: 5.55 ms
n = 17
CPU times: user 6.4 ms, sys: 0 ns, total: 6.4 ms
Wall time: 6.41 ms
n = 18
CPU times: user 7.29 ms, sys: 0 ns, total: 7.29 ms
Wall time: 7.3 ms
n = 19
CPU times: user 8.18 ms, sys: 0 ns, total: 8.18 ms
Wall time: 8.19 ms
```
Notice that last time difference is 8ms vs 2.5 minutes that's a 20000x speedup and it only gets bigger as n increases.

Please bear in mind that in the end this is going to lead to big speed ups for many operations. We should try to avoid introducing small slowdowns at the same time but we also shouldn't worry too much about them if we can see how big the potential speed up can be for things that are actually slow.

I'll see if there's something that can be done to speed up indexing which will affect those benchmarks.

moorepants:
Thanks for the detailed response. It is all convincing and good to know that even if this PR slows things down you would speed things back up with subsequent PRs.

The main case I'm worried about are operations on dense matrices with symbolic expressions as the elements. The `randMatrix()` function you show creates numerical matrices, so I don't know if those examples represent symbolic matrices also.

Here is an example (uses pydy to get the matrices):

```python
from pydy.models import n_link_pendulum_on_cart
sys = n_link_pendulum_on_cart(n=30)
M = sys.eom_method.mass_matrix
b = sys.eom_method.forcing
%time res = M.LUsolve(b)
```

That solve takes 30 seconds for a 31 x 31 matrix on my laptop.

moorepants:
I tried this PR  with the above example and got about the same time:

````
In [5]: %time res = M.LUsolve(b)
CPU times: user 31.3 s, sys: 32.1 ms, total: 31.4 s
Wall time: 31.4 s
```

So, that's good.

moorepants:
Here is a explicit comparison for reference:

```
moorepants@nandi:sympy(master)$ ipython
Python 3.7.10 | packaged by conda-forge | (default, Feb 19 2021, 16:07:37) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.25.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: %paste
from pydy.models import n_link_pendulum_on_cart
%time sys = n_link_pendulum_on_cart(n=30)
M = sys.eom_method.mass_matrix
b = sys.eom_method.forcing
%time res = M.LUsolve(b)

## -- End pasted text --
/home/moorepants/src/sympy/sympy/printing/ccode.py:12: SymPyDeprecationWarning: 

importing from sympy.printing.ccode has been deprecated since SymPy
1.7. Use Import from sympy.printing.c instead. See
https://github.com/sympy/sympy/issues/20256 for more info.

  deprecated_since_version="1.7").warn()
CPU times: user 1min 53s, sys: 118 ms, total: 1min 53s
Wall time: 1min 53s
CPU times: user 29.5 s, sys: 27.9 ms, total: 29.5 s
Wall time: 29.5 s

In [2]:                                                                                                  
Do you really want to exit ([y]/n)? 
moorepants@nandi:sympy(master)$ git checkout pr-21626
Switched to branch 'pr-21626'
moorepants@nandi:sympy(pr-21626)$ ipython
Python 3.7.10 | packaged by conda-forge | (default, Feb 19 2021, 16:07:37) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.25.0 -- An enhanced Interactive Python. Type '?' for help.

In [2]: %paste
from pydy.models import n_link_pendulum_on_cart
%time sys = n_link_pendulum_on_cart(n=30)
M = sys.eom_method.mass_matrix
b = sys.eom_method.forcing
%time res = M.LUsolve(b)

## -- End pasted text --
/home/moorepants/src/sympy/sympy/printing/ccode.py:12: SymPyDeprecationWarning: 

importing from sympy.printing.ccode has been deprecated since SymPy
1.7. Use Import from sympy.printing.c instead. See
https://github.com/sympy/sympy/issues/20256 for more info.

  deprecated_since_version="1.7").warn()
CPU times: user 1min 53s, sys: 89.7 ms, total: 1min 53s
Wall time: 1min 53s
CPU times: user 30 s, sys: 28 ms, total: 30 s
Wall time: 30 s
```

oscarbenjamin:
> The main case I'm worried about are operations on dense matrices with symbolic expressions as the elements. The `randMatrix()` function you show creates numerical matrices, so I don't know if those examples represent symbolic matrices also.

The slowdown that you are seeing in the benchmarks is essentially that there is a little extra overhead in setting items in the matrix. I think particularly the LUsolve function uses a lot of slice assignment like `M[i,j:N] = some_list` or something like that which is not a case that I tried to optimise here. That could be made faster but if that is a dominant factor in the routine then it suggests that the routine can be made faster still if it just uses lists directly rather than matrix assignment. It shows up as being significant only after caching because otherwise if your matrices have significant symbolic expressions then the cost of slice assignment will be dwarfed by the other calculations.

Where something like the cost of that slice assignment is likely to be noticeable is if your matrix has simple elements like integers which is why I demonstrate that DomainMatrix is already much faster in that case.

Your PyDy examples are matrices with many symbols and also sin/cos terms. The sin/cos are not something that DomainMatrix can handle properly with a domain yet (https://github.com/sympy/sympy/issues/20608) so until that is resolved the EX or EXRAW domains would need to be used which negates the main benefit of DomainMatrix. The old routines can continue to be used though so I don't expect that you will see any significant difference in those cases (unless a significant regression is overlooked somewhere).

The reason this PR does not immediately make use of all the DomainMatrix routines is that each one needs to be considered carefully for what kinds of expressions it should be used with and what domains should be used and how the user can select between the different routines and so on.

moorepants:
> The old routines can continue to be used though so I don't expect that you will see any significant difference in those cases (unless a significant regression is overlooked somewhere).

Sounds good.

oscarbenjamin:
The benchmark results comment has updated: https://github.com/sympy/sympy/pull/21626#issuecomment-883404852

They are looking happier now. I'll see if I can fix up some of the others.

oscarbenjamin:
I've tried looking into the other slowdowns indicated by the benchmarks but:

1. The particular benchmark results are not really representative of actual performance due to cacheing.
2. The slowdowns are seen in benchmarks with very small matrices and are not replicated for larger matrices. From my testing so far any potential change that improves one of the benchmarks slows down handling larger matrices.
3. Cases like det/rank can be much faster if the DomainMatrix routines are used but I don't want to enable that in this PR.

I think at this stage we should just accept these "slowdowns" (as suggested by the benchmarks) and merge this so that further work optimising matrix performance can continue.

Maybe also the cache should be disabled for the benchmark timings because they are very much not representative of the actual time to perform a particular calculation *once* which is actually what matters.

moorepants:
> Maybe also the cache should be disabled for the benchmark timings because they are very much not representative of the actual time to perform a particular calculation once which is actually what matters.

Sounds like a good idea.

moorepants:
The note about the cache reminded me that I used to always run the benchmarks with the cache size bumped up quite a bit. I can't remember why, but I'm guessing I got inconsistent results or something if I didn't do that.

oscarbenjamin:
I'm going to merge this soon unless someone says that they want to review it.

moorepants:
Probably best to specifically ask someone to review a PR like this. It's daunting for most contributors due to the depth.

oscarbenjamin:
I think just a nod to the general idea is still good even if it can't be reviewed in detail.

There aren't many people who could review this in detail. They know who they are and I'm sure they've seen this PR. If they don't want to review it that's fine but I don't think that should hold up merging this. There are lots of other improvements that can be made building on this but it would be best to get this in as the first step.

jksuom:
I haven't been following the development of matrix code closely of late but I like the idea of having a unified low level representation for all matrices. I think that this could be merged.

moorepants:
I only suggest asking someone to review because I've always understood that SymPy's PRs should be reviewed by at least one other person other than the author. Self-merging PRs with no review has traditionally been frowned upon. But if no one can or will review, then some kind of call needs to be made. By who? Not sure.

oscarbenjamin:
I think @sylee957 will understand the detail of this best although others could also review this.

I'm happy to wait if that gives time for a review to happen. If the review won't happen either way then I will just merge this. Reviews are best practice but if we can't merge something like this because we don't have the resource to get it reviewed then it will become impossible to make major improvements to sympy. Since sympy really needs major improvements that is not good for its long term prognosis.

oscarbenjamin:
I'll merge this tomorrow if I don't hear back that anyone intends to review it.

I have a bunch of improvements planned for matrices. Those are all on hold right now because the best way to implement them needs this PR as a starting point.

oscarbenjamin:
Okay, I'm merging this. Thanks all for the discussion.

With this in further improvements will be more straight-forward. The basic change here needs good testing though so it needs to sit in master for some time before the next release.

Issue #14374, opened 2018-03-03
Should 0 times a Quantity keep the units that the Quantity had? 


normalhuman:
The current behavior is that multiplying anything (except an infinite quantity) by 0 returns 0 itself. ([Relevant code](https://github.com/sympy/sympy/blob/master/sympy/core/mul.py#L596)).
```
>>> from sympy.physics.units import meter
>>> 0*meter
0
```
Should this be changed so that the result is 0 meters, instead of unitless 0? This was asked on Stack Overflow ([Why doesn't sympy support units with 0 magnitude?
](https://stackoverflow.com/q/49084526)) and seems to make sense to me.

Upabjojr:
Well, you can still have `Mul(0, meter, evaluate=False)`.

Compare this behaviour with complex numbers. We don't distinguish between real zero and complex zero. The imaginary unit is a variable you put in the expression to have complex numbers.

Quantities are currently treated as variables, so a concept analogous to complex numbers apply.

Also consider that the `Quantity` object supports a scale factor definition, which can be set to zero.

asmeurer:
It makes sense mathematically, but I don't know how hard it is to do in implementation. 

> Compare this behaviour with complex numbers. We don't distinguish between real zero and complex zero. The imaginary unit is a variable you put in the expression to have complex numbers.

The analogy isn't apt here because you can add real and imaginary numbers, but you can't add incompatible quantities, even if they have zero magnitude.

bjodah:
FWIW, I would prefer if e.g.
```
>>> 0*meter + 0*seconds
```
raised an Exception.

schymans:
+1 for `0*meter + 0*secons` raising an exception. See also Issue #13267.

czgdp1807:
This issue appears quite interesting. Can I work on this? I am starting my work on this. :-)

Upabjojr:
@czgdp1807 feel free to work on it if you like.

Upabjojr:
Just beware that `meter + second` may be meaningful in relativity.

czgdp1807:
@Upabjojr Sure. That's an important point.

schymans:
@Upabjojr, could you point us to an example? @czgdp1807 see also: https://github.com/sympy/sympy/issues/14318

Upabjojr:
[...]

PR #21628, opened 2021-06-17
units: Define scale_offset and temperature scales

### LINKED ISSUE #13379: sympy.physics.units documentation doesn't list supported units
I'm wondering whether sympy.physics.units supports degree Celsius as temperature units, but there is no documentation of units that sympy supports in http://docs.sympy.org/latest/modules/physics/units/index.html , so I have no idea whether sympy supports degree Celsius. Compare it to [Pint](https://pint.readthedocs.io/en/latest/) where they have a complete documentation of supported units: https://github.com/hgrecco/pint/blob/master/pint/default_en.txt 

The problem is that Pint units is not supported in Sympy calculations, and Pint doesn't have nice Latex equation output

meganly:

Fixes #13379

Introduces a scale offset to define the temperature scales of degrees Celsius and degrees Fahrenheit.

#### Release Notes

* physics.units
  * Added support for degrees Celsius and degrees Fahrenheit.


meganly:
Unfortunately not.
```
>>> convert_temperature(var('z')*kelvin, degrees_celsius).subs(z, 0).n()
-272.15*degree_celsius
```
The issue is that `(var('z')*kelvin).as_coeff_Mul()` is `(1, kelvin*z)` instead of `(z, kelvin)`.  As a result the magnitude sent to `to_kelvin` is 1 and not z or 0 as desired.

Would it help to be more clever than using `.as_coeff_Mul` to get the magnitude from a quantity?

It would be nice if `0*kelvin`, `0*degrees_celsius`, and `0*degrees_fahrenheit` didn't drop the units.

smichr:
> It would be nice if `0*kelvin`, `0*degrees_celsius`, and `0*degrees_fahrenheit` didn't drop the units.

What about make `convert_temperature(mag, old_unit, new_unit)` so `convert_temperature(0, kelvin, degree_celcius)`

meganly:
That should work!  It also makes it clear that ambiguous temperature conversions like 2 oC + 3 oF are not handled.  Even `2*oC` is slightly unclear whether you mean convert 2 oC to K or convert oC to K and then multiply by 2... though most people would probably take 2*oC to mean the former.

For what it's worth I do think that 0 times a quantity should keep the quantity https://github.com/sympy/sympy/issues/14374.

meganly:
I pushed an update to do `convert_temperature(magnitude, source_units, target_units)` and return the converted temperature as `(magnitude, target_units)`.  This way if the answer is 0 the target units aren't lost.

I'm not sure this solves all the issues with supporting temperature scales.  Ideally `convert_to(0*oC, oF)` would just work.  However, I feel like this is a decent start.

smichr:
> `2*oC` is slightly unclear whether you mean convert 2 oC to K or convert oC to K and then multiply by 2

This raises a good point in terms of how temperatures (other than absolute) can be used in expressions. Perhaps we should have a `T` ~function~class so you would write `n*R*T(0, oC)` and `T(0, oC)` would give the Kelvin temp. `T` might be the same as `convert_termperature(mag, old, new=K)` and would return a `T` object if new is not `K`, e.g. `T(0, oC, oF) -> T(32, oF)`. I'm not sure of all the implications, but maybe we need an `OffsetQuantity(name, base_unit, allowed_units)` that would generalize this for any other offset quantities.

oscarbenjamin:
Maybe absolute temperatures should be a different kind of quantity/unit from temperature differences. This is I guess analogous to the distinction between datetime and timedelta in the datetime module:
```
datetime - datetime -> timedelta
datetime + timedelta -> datetime
number * timedelta -> timedelta
number * datetime -> error
```
Analogously we can have `absolute` and `relative` temperature and
```
absolute - absolute -> relative
absolute + relative -> absolute
expression * relative -> relative
expression * absolute -> undefined
```
Then there could be constants representing particular temperatures e.g.:
```
zero_degC = melting_point_of_water
zero_degF = minus32_degC
zero_K = absolute_zero
```
Converting an absolute temperature from e.g. degC to K would involve the constant zero_degC but converting a relative temperature from degC to K would not.

schymans:
I am not sure what you mean by absolute and relative, @oscarbenjamin. The Kelvin scale is an absolute scale, but you can also express temperature differences in Kelvin. Could you give examples?
I like the idea of `convert_temperature(mag, old_unit, new_unit)`, and I would not try to make `convert_to(0*oC, oF)`  result in `-32*oF`, as in my opinion it does not make any sense to multiply anything by oC or oF.

oscarbenjamin:
> Could you give examples?

A weather report said that the temperature has dropped by "5 degrees" since yesterday. It means a temperature *difference* of 5 degC but I suppose that in the US it would be reported as a difference of 9 degF. An absolute temperature of 5 degC corresponds to 41 degF though.

https://en.wikipedia.org/wiki/Fahrenheit#Conversion_(specific_temperature_point)
https://en.wikipedia.org/wiki/Fahrenheit#Conversion_(temperature_difference_or_interval)

> it does not make any sense to multiply anything by oC or oF.

It makes sense in the context of temperature differences but not for absolute temperatures. This is analogous to saying that you can multiply a distance but not a position.

schymans:
> This is analogous to saying that you can multiply a distance but not a position.

Good point. With oC and K we have the helpful convention that differences are always expressed in K, which is a 1:1 conversion, but you are right, in the Fahrenheit world, such a convention would not work. 
Now that you mention it, all temperatures are relative to a reference point, so they are equivalent to a difference. The same actually applies to the chemical potential, expressed in energy per mass or per amount of substance. It can be positive or negative, or 0. Even pressure can be negative in a liquid.

meganly:
> Perhaps we should have a `T` ~function~class so you would write `n*R*T(0, oC)` and `T(0, oC)` would give the Kelvin temp.

I like this idea.  I don't know of other offset quantities besides temperature so an `OffsetQuantity` class might not be needed?  I will continue thinking it over.  This reminds me of Pint's Quantity class where they store the magnitude and units.  Should `2*oC` throw an error or return `T(2, oC)`?

> Maybe absolute temperatures should be a different kind of quantity/unit from temperature differences.

Yes.  That's a good idea, but probably beyond the scope of this PR.  Pint handles temperature differences with a [delta unit](https://pint.readthedocs.io/en/latest/nonmult.html#id1).

> I would not try to make `convert_to(0*oC, oF)` result in `-32*oF`

Making `convert_to` handle temperature conversions was probably a naive/simplistic hope given the intricacies of temperatures.

meganly:
I created a `MagnitudeQuantity` class to store the magnitude and units of temperatures that `convert_temperature` takes as an argument, e.g. 
```
>>> MagnitudeQuantity(0, oC)
0 degree_celsius
>>> convert_temperature(MagnitudeQuantity(0, oC), oF)
32 degree_fahrenheit
```
Let me know what you think.

smichr:
Looks interesting. I wonder about a different name, however, since this doesn't highlight the reason that this is needed for temp and not any other quantity. Maybe "OffsetQuantity" or just "Offset"? We also need the str printer to give something that can be copied and pasted to recreate the object. So "Offset(32, oC)" would be an option.

meganly:
> I wonder about a different name, however, since this doesn't highlight the reason that this is needed for temp and not any other quantity.

In theory it could be used with other quantities, which would solve #14374.  Some work would be needed to define arithmetic and perhaps make mag * unit an alias for `MagnitudeQuantity(mag, unit)`.  It's food for thought, but I'm also happy to name the class "Offset" and have the str printer return "Offset(32, oC)".

> To do the conversion, what about something like `Offset(0, oC)(K) -> (273.15, K)` or `Offset(0, oC)(oF) -> Offset(32, oF)` -- so a call with a valid unit will call `convert_termperature` transparently.

Sounds good.

>And how to get it to convert to absolute units, `Offset(32, oF).ref -> 273.15*K`?

There should definitely be a method to convert to absolute units.

smichr:
> In theory it could be used with other quantities, which would solve #14374

I'm not sure how far we could go with this, but we can automatically convert `0*unit` into `MagUnit(0, unit)` and MagUnit(3, unit) could automatically denest as `3*unit`, a Mul. The following just shows that the `__mul__` and `__rmul__` can catch this python expression and that it doesn't automatically go to 0:
```python
>>> class unit():
...     def __rmul__(self, other):
...         if other:return 1
...         return -1
...     def __mul__(self, other):
...         if other:return 2
...         return -2
>>> unit()*1
2
>>> unit()*0
-2
>>> 0*unit()
-1
>>> 2*unit()
1
```

meganly:
I would also convert `3*unit` into `MagUnit(3, unit)` instead of denesting to avoid issues with `MagUnit(3, oC)`.  This is what Pint does.  Multiplication could be defined by doing something like `MagUnit(3, N) * MagUnit(2, m) = MagUnit(6, N*m)`.

smichr:
> Multiplication could be defined by doing something like `MagUnit(3, N) * MagUnit(2, m) = MagUnit(6, N*m)`

Since a quantity has a magnitude, unit and uncertainty, I wonder if we could start representing Quantities like any of the above simply as `Quantity(6, N*m)` or something like `Quantity(8*Offset(30, oC), J/K)`.

meganly:
I like that idea, but am not sure how defining units would work.  If quantities are represented with a magnitude, then is a special `Offset` class needed?  I could see an `Offset` class being advantageous to define a `__call__` method to handle temperature conversions as `Offset(0, oC)(oF) -> Offset(32, oF)` as well as defining a `ref` property to convert to absolute units as `Offset(32, oF).ref -> 273.15*K`

smichr:
> then is a special `Offset` class needed

I think so. An offset is needed when you want to give a magnitude relative to a an alternate reference point for the underlying standard quantity. `Offset(0, oC) -> MQ(273.15, K)`.
- MQ might be the default way to represent quantities and guards against loss of units when magnitude is 0
- Offset to allow input of an offset quantity; maybe it converts to absolute units with a doit method?

meganly:
I pushed a change to rename `MagnitudeQuantity` to `Offset` and added a `__call__` and `doit` method for temperature conversions.  I'll leave the idea of representing Quantities as a class with a magnitude and unit as an exercise for another time.

sympy/physics/units/quantities.py
```diff
@@ -209,3 +227,59 @@ def convert_to(self, other, unit_system="SI"):
...
+
+    def _latex(self, printer):
+        return "%s %s" %(printer._print(self.magnitude), printer._print(self.units))
+
+    def __call__(self, other):
+        """Converts temperatures to different temperature scales"""
+        from .util import convert_temperature
+        return convert_temperature(self, other, unit_system="SI")
+
+    def doit(self):
```
smichr, line 281: To keep this Henry useful I think we need to query for the quantity's offset and do the conversion to that...similar to get scale offset function below.

asmeurer:
@Upabjojr if you have a chance, it would be good if you could review this.

asmeurer:
I've seen temperature deltas referred to as "Celsius degrees" C˚ as opposed to the absolute "degrees Celsius" ˚C. I don't know how common that practice is, but I remember learning it in school. 

I think we should definitely distinguish the two. Delta temperature units are more standard as units, whereas absolute temperatures, as with absolute times, are not. Absolute units cannot be multiplied by another unit or even by a scalar (it's meaningless to say today is 2x warmer than yesterday based on the numeric F or C temperatures). In general, `Absolute + absolute` does not make sense, but `absolute +- delta` does. And `absolute - absolute = delta` (since `scalar*absolute` also does not make sense, the ability to do subtraction does not imply addition). 

Honestly, I'm not really sure if the units module should try to handle "absolute" units like Fahrenheit. Another example to think about would be an absolute timestamp like in the datetime module. These aren't really units. There's more to it than just an offset. I would love to hear @Upabjojr's thoughts on this, because I know he has thought a bit about how to define units in a mathematically consistent way.

sympy/physics/units/quantities.py
meganly: In this case the offset of oC and oF would be Kelvin, right?  I want to make sure I understand what to query for correctly.

schymans:
> I think we should definitely distinguish the two. Delta temperature units are more standard as units, whereas absolute temperatures, as with absolute times, are not. Absolute units cannot be multiplied by another unit or even by a scalar (it's meaningless to say today is 2x warmer than yesterday based on the numeric F or C temperatures). In general, `Absolute + absolute` does not make sense, but `absolute +- delta` does. And `absolute - absolute = delta` (since `scalar*absolute` also does not make sense, the ability to do subtraction does not imply addition).

Aren't all "absolute" scales actually Deltas in relation to some commonly accepted reference? For date, each calendar has a 0 related to some event in the past, for time, it is midnight of the previous day, for oC it is the freezing point of water etc. As mentioned earlier, even the zero-point of pressure is somehow arbitrary, i.e. the pressure of vacuum, and in liquids it can be negative. There are some conventions for the zero-point that prohibit negative values, such as that of the Kelvin scale, but in effect, 273 K just means 273 K above absolute 0. If we accept that nothing is absolute and everything is a delta, perhaps we can be more explicit about the convention of the offset or zero-point that changes as we change units for the same dimension. From this perspective,`3 oC + 32 oF = 3 oC = 37.4 oF`.

smichr:
regarding degrees Celcius and Celcius degrees, see [here](https://www.answers.com/Q/Difference_between_degrees_Celsius_and_Celsius_degrees). 

Here's the thorn on the rose of these changes:

If a formula calls for a temperature `Offset(t, oC)` is the way to enter a reading of `t` on a thermometer marked in Celcius. But if a formula required a delta T (as when computing the heat acquired when a gram of water undergoes a change of 3 Celcius degrees) then entering `Offset(3, oC)` would be wrong and `3*oC` would be right. But if there were a 0 Celcius degree change then we are stuck with no way to say this because the units would be inconsistent. Perhaps we need `Delta`, too?
```
Delta(3, oC) = 3*oC
Delta(0, oC) (and does not simplify to 0)

Offset(0, oC) == 273.15*K
```

asmeurer:
This is starting to get deep into the philosophy of science, and I'm willing to bet that someone has thought about this way more than we have. So we might want to look into the literature.

For our purposes, SymPy's type system doesn't always completely faithfully represent mathematical types in their most correct abstract sense. SymPy always takes a more pragmatic approach. Although the units module go a little beyond most parts of SymPy in trying to be "type safe" (I'm talking about mathematical/physical types, not Python types).

For temperatures, we really need to figure something out, because there is a big difference between a C -> F conversion that adds 32 (absolute temperature) and one that doesn't (delta temperature). If the user requests one but gets the other, the final result will be wrong.

smichr:
Could we have `C_deg` unit (of energy/temperature change) and `oC` offset?  The `oC` could only be used in the Offset context, so you couldn't say `3*oC` since `oC` is not a unit: it implies a reference point. You also couldn't say `Offset(3, C_deg)`.

meganly:
I think a `C_deg` unit and a `oC` unit along with the `Offset` class could work.  The intent of this PR was to support absolute temperatures and leave relative/delta temperatures for another PR.

I'll look more into the literature though.  If anyone has any suggestions for references, let me know.

smichr:
> I think a `C_deg` unit and a `oC` unit along with the `Offset` class could work. The intent of this PR was to support absolute temperatures and leave relative/delta temperatures for another PR.
> 
So offset could allow writing absolute values in terms of a different reference and C_deg would actually be the way to differentiate between relative (temp difference) and absolute temperatures.

Upabjojr:
Hi!

I had a quick view on this PR and my impression is that `convert_temperature` seems a bit of unnecessary. It would be preferrable to have its code merged into `convert_to`.

Concerning the offset, I'm fine with it. Though, maybe we can be more generic by adding a customized transformation. For example, instead of:
```python
degree_celsius.set_global_relative_scale_offset(Rational(27315, 100), kelvin)
```
maybe something like:
```python
degree_celsius.set_global_custom_transformation_to(lambda x: x + Rational(27315, 100), kelvin)
```
In this way, we would be able to not only handle offsets, but any generic transformations. 

For example, units like _decibels_ and magnitude of stars will be easily linked to SI units.

meganly:
Thanks everyone for the feedback.  I've looked more into the literature and I think it makes sense to handle temperatures similar to what Pint does.  That is,

Sympy would have the temperature units of kelvin, Rankine, Celsius, Fahrenheit abbreviated as `K`, `deg_R`, `deg_C`, `deg_F`.  Absolute temperatures with the non-multiplicative temperature units of `deg_C` and `deg_F` would need to be represented in the `Offset` class as `Offset(mag, unit)`.

For every non-multiplicative temperature unit, there is a delta counterpart to specify relative temperature differences, namely `C_deg` and `F_deg`.  Subtraction of two temperatures in an offset would yield a delta quantity
```
>>> Offset(20, deg_C) - Offset(5, deg_C)
15*C_deg
```

You would be able to add or subtract an offset and a delta quantity
```
Offset(20, deg_C) + 10*C_deg
Offset(30, deg_C)
```
However adding an absolute temperature to an offset would throw an error about an ambiguous operation.  To avoid the ambiguity you would need to convert the offset to absolute units (kelvin or Rankine), or use delta units.  Similarly multiplication, division, and exponentiation with offsets would also throw an error.

``Convert_to`` can be made to handle temperature conversions such as
```
>>> convert_to(Offset(0, deg_C), deg_F)
Offset(32, deg_F)
```

It's an interesting idea to generalize the `Offset` class to something that would handle generic transforms.  However it would be nice for the `Offset` class to have a `doit` method to easily convert to absolute temperature units, and I don't see how to generalize the `doit` method for a more generic class.

Upabjojr:
Ideally we would have this working:
```python
>>> convert_to(21*celsius, fahrenheit)
69.8*fahrenheit

>>> convert_to(273.15*kelvin, fahrenheit)
32*fahrenheit
```

I'm not entirely sure that `Offset` should be exposed to end-users. SymPy should be as simple as possible to use and `Offset` makes the API a little bit more complicated. Maybe it should just be used internally by the algorithm called by `convert_to`?

meganly:
> I'm not entirely sure that `Offset` should be exposed to end-users. SymPy should be as simple as possible to use and `Offset` makes the API a little bit more complicated. Maybe it should just be used internally by the algorithm called by `convert_to`?

The issue is that `21*deg_C` is ambiguous.  It's unclear whether converting `21*deg_C` means convert 21 deg_C to deg_F or convert deg_C to deg_F and then multiply by 21.  Celsius and Fahrenheit are non-multiplicative units so `convert_to(Offset(21, deg_C), deg_F)` is not the same as`21*convert_to(Offset(1, deg_C), deg_F)`.  However, most people would probably take `21*deg_C` to mean `Offset(21, deg_C)`.  We could make that the default, but exposing `Offset` eliminates any ambiguity.

oscarbenjamin:
I don't think that multiplication should be used for absolute temperature units. It should be something like `celsius(20)` rather than `20 * celsius`.

Upabjojr:
> I don't think that multiplication should be used for absolute temperature units. It should be something like `celsius(20)` rather than `20 * celsius`.

I'm OK with both of them. `celsius(20)` would require the development of new API.

Upabjojr:
@oscarbenjamin can the new `.kind` infrastructure help in this case? Would it make sense to have a new kind for units? By default, expressions containing units could be equipped with an _offset_ method.

oscarbenjamin:
> @oscarbenjamin can the new `.kind` infrastructure help in this case? Would it make sense to have a new kind for units? By default, expressions containing units could be equipped with an _offset_ method.

Yes, I think that units or perhaps quantities should have a kind. Certainly they should not be NumberKind. Ideally something like `tan(2*metres)` would be an error.

I don't know enough about the units module to say what a good design would be but there are various special cases in the codebase for units that should ideally be handled by an extensible mechanism instead. That is the situation that kind was introduced for.

meganly:
> Would it make sense to have a new kind for units?  By default, expressions containing units could be equipped with an _offset_ method.

Is this saying `Mul(20, deg_C)` would have a `QuantityKind`?  How would an _offset_ method for `QuantityKind` work?  A new kind for units could be useful, but I don't see how that helps to represent quantities with degrees Fahrenheit or degrees Celsius.

> I don't think that multiplication should be used for absolute temperature units. It should be something like `celsius(20)` rather than `20 * celsius`.

I agree that multiplication should not be used for absolute temperature units.  I'd rather have one `Offset` class than separate `celsius` and `fahrenheit` classes.  How about making the API simpler by renaming `Offset` as the temperature class `T`?

schymans:
I just searched for "units with offset" and found quite a few interesting discussions, most related to the Fahrenheit and Celsius issue:
https://github.com/labrad/pylabrad/issues/141
https://github.com/hgrecco/pint/issues/147
https://github.com/r-quantities/units/issues/253
https://github.com/df7cb/postgresql-unit/issues/5

Judging by the difficulties pointed out in some of these discussions and the ambiguiities even when adding or subtracting oC or oF quantities from each other (what should 20oC + 32oF give?), I would strongly advocate not to try any arithmetic operations with oC or oF, but just return an error message asking to convert to absolute temperatures or deltas in Kelvin or Rankine. It would be easy to provide a conversion routine, but very hard to avoid ambiguity with oC and oF if we want to allow arithmetic operations with them.

meganly:
I tried pushing a commit to throw an error for any arithmetic operations with oC or oF, but seems to have a negative impact on performance.  Any suggestions for a better way to go about this?  I've thought about creating `__add__` and `__mul__` methods for the `Quantity` class similar to the `Dimension` class or perhaps a `QuantityKind` could be used somehow.

Upabjojr:
> I've thought about creating `__add__` and `__mul__` methods for the `Quantity` class similar to the `Dimension` class or perhaps a `QuantityKind` could be used somehow.

Warning: that would make this PR a lot harder. Overloading `__add__` and `__mul__` in a consistent manner has been an open issue in SymPy for a long time.

meganly:
> Warning: that would make this PR a lot harder. Overloading `__add__` and `__mul__` in a consistent manner has been an open issue in SymPy for a long time.

Yes overloading operators can be tricky, but perhaps the following could work:

```python
 class Quantity(AtomicExpr):
@@ -19,6 +20,8 @@ class Quantity(AtomicExpr):
     is_nonzero = True
     _diff_wrt = True
 
+    _op_priority = 13.0
+
     def __new__(cls, name, abbrev=None, dimension=None, scale_factor=None,
                 latex_repr=None, pretty_unicode_repr=None,
                 pretty_ascii_repr=None, mathml_presentation_repr=None,
@@ -164,6 +167,15 @@ def is_multiplicative(self):
             return False
         return True
 
+    @call_highest_priority('__rmul__')
+    def __mul__(self, other):
+        from sympy.core import Mul
+        args = [self, other]
+        if any(isinstance(arg, Quantity) and arg.is_multiplicative is False for arg in args):
+            raise ValueError("Ambiguous operation with offset unit")
+        return Mul(self, other)
+    __rmul__ = __mul__
```
 I don't know if that will be faster than what I did in 6fc5463 though.

defencedog:
Something ike this: https://github.com/Schlumberger/UOM
```
from uom import conversion_factors

scale, offset = conversion_factors(source='m', target='ft')
```

PR #21630, opened 2021-06-17
Refactor hypergeometric solver

Mohitbalwani26:

SEE #18348

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/solvers/ode/hypergeometric.py
+++ b/sympy/solvers/ode/hypergeometric.py
@@ -1,0 +1,174 @@
+from sympy.core import S, Pow
+from sympy.core.function import expand
+from sympy.core.relational import Eq
+from sympy.core.symbol import Symbol, Wild
+from sympy.functions import sqrt
+from sympy.polys import roots, gcd
+from sympy.polys.polytools import cancel, factor
+from sympy.simplify import collect, simplify
+from sympy.simplify.powsimp import powdenest
+
+
+def equivalence_hypergeometric(A, B, func):
+    # This method for finding the equivalence is only for 2F1 type.
+    # We can extend it for 1F1 and 0F1 type also.
+    x = func.args[0]
+
+    # making given equation in normal form
+    I1 = factor(cancel(A.diff(x)/2 + A**2/4 - B))
+
+    # computing shifted invariant(J1) of the equation
+    J1 = factor(cancel(x**2*I1 + S(1)/4))
+    num, dem = J1.as_numer_denom()
+    num = powdenest(expand(num))
+    dem = powdenest(expand(dem))
+    pow_num = set()
+    pow_dem = set()
+    # this function will compute the different powers of variable(x) in J1.
+    # then it will help in finding value of k. k is power of x such that we can express
+    # J1 = x**k * J0(x**k) then all the powers in J0 become integers.
+    def _power_counting(num):
+        _pow = {0}
+        for val in num:
+            if val.has(x):
+                if isinstance(val, Pow) and val.as_base_exp()[0] == x:
+                    _pow.add(val.as_base_exp()[1])
+                elif val == x:
+                    _pow.add(val.as_base_exp()[1])
+                else:
+                    _pow.update(_power_counting(val.args))
+        return _pow
+
+    pow_num = _power_counting((num, ))
+    pow_dem = _power_counting((dem, ))
+    pow_dem.update(pow_num)
+
+    _pow = pow_dem
+    k = gcd(_pow)
+
+    # computing I0 of the given equation
+    I0 = powdenest(simplify(factor(((J1/k**2) - S(1)/4)/((x**k)**2))), force=True)
+    I0 = factor(cancel(powdenest(I0.subs(x, x**(S(1)/k)), force=True)))
+    num, dem = I0.as_numer_denom()
+
+    max_num_pow = max(_power_counting((num, )))
+    dem_args = dem.args
+    sing_point = []
+    dem_pow = []
+    # calculating singular point of I0.
+    for arg in dem_args:
+        if arg.has(x):
+            if isinstance(arg, Pow):
+                # (x-a)**n
+                dem_pow.append(arg.as_base_exp()[1])
+                sing_point.append(list(roots(arg.as_base_exp()[0], x).keys())[0])
+            else:
+                # (x-a) type
+                dem_pow.append(arg.as_base_exp()[1])
+                sing_point.append(list(roots(arg, x).keys())[0])
+
+    dem_pow.sort()
+    # checking if equivalence is exists or not.
+
+    if equivalence(max_num_pow, dem_pow) == "2F1":
+        return {'I0':I0, 'k':k, 'sing_point':sing_point, 'type':"2F1"}
+    else:
+        return None
+
+
+def match_2nd_2F1_hypergeometric(I, k, sing_point, func):
+    x = func.args[0]
+    a = Wild("a")
+    b = Wild("b")
+    c = Wild("c")
+    t = Wild("t")
+    s = Wild("s")
+    r = Wild("r")
+    alpha = Wild("alpha")
+    beta = Wild("beta")
+    gamma = Wild("gamma")
+    delta = Wild("delta")
+
+    rn = {'type':None}
+    # I0 of the standerd 2F1 equation.
+    I0 = ((a-b+1)*(a-b-1)*x**2 + 2*((1-a-b)*c + 2*a*b)*x + c*(c-2))/(4*x**2*(x-1)**2)
+    if sing_point != [0, 1]:
+        # If singular point is [0, 1] then we have standerd equation.
+        eqs = []
+        sing_eqs = [-beta/alpha, -delta/gamma, (delta-beta)/(alpha-gamma)]
+        # making equations for the finding the mobius transformation
+        for i in range(3):
+            if i<len(sing_point):
+                eqs.append(Eq(sing_eqs[i], sing_point[i]))
+            else:
+                eqs.append(Eq(1/sing_eqs[i], 0))
+        # solving above equations for the mobius transformation
+        _beta = -alpha*sing_point[0]
+        _delta = -gamma*sing_point[1]
+        _gamma = alpha
+        if len(sing_point) == 3:
+            _gamma = (_beta + sing_point[2]*alpha)/(sing_point[2] - sing_point[1])
+        mob = (alpha*x + beta)/(gamma*x + delta)
+        mob = mob.subs(beta, _beta)
+        mob = mob.subs(delta, _delta)
+        mob = mob.subs(gamma, _gamma)
+        mob = cancel(mob)
+        t = (beta - delta*x)/(gamma*x - alpha)
+        t = cancel(((t.subs(beta, _beta)).subs(delta, _delta)).subs(gamma, _gamma))
+    else:
+        mob = x
+        t = x
+
+    # applying mobius transformation in I to make it into I0.
+    I = I.subs(x, t)
+    I = I*(t.diff(x))**2
+    I = factor(I)
+    dict_I = {x**2:0, x:0, 1:0}
+    I0_num, I0_dem = I0.as_numer_denom()
+    # collecting coeff of (x**2, x), of the standerd equation.
+    # substituting (a-b) = s, (a+b) = r
+    dict_I0 = {x**2:s**2 - 1, x:(2*(1-r)*c + (r+s)*(r-s)), 1:c*(c-2)}
+    # collecting coeff of (x**2, x) from I0 of the given equation.
+    dict_I.update(collect(expand(cancel(I*I0_dem)), [x**2, x], evaluate=False))
+    eqs = []
+    # We are comparing the coeff of powers of different x, for finding the values of
+    # parameters of standerd equation.
+    for key in [x**2, x, 1]:
+        eqs.append(Eq(dict_I[key], dict_I0[key]))
+
+    # We can have many possible roots for the equation.
+    # I am selecting the root on the basis that when we have
+    # standard equation eq = x*(x-1)*f(x).diff(x, 2) + ((a+b+1)*x-c)*f(x).diff(x) + a*b*f(x)
+    # then root should be a, b, c.
+
+    _c = 1 - factor(sqrt(1+eqs[2].lhs))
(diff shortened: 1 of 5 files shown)


sympy/solvers/ode/tests/test_ode.py
```diff
@@ -889,9 +889,11 @@ def test_2nd_power_series_ordinary():
     assert checkodesol(eq, sol) == (True, 0)
 
     eq = (1 + x**2)*(f(x).diff(x, 2)) + 2*x*(f(x).diff(x)) -2*f(x)
-    assert classify_ode(eq) == ('2nd_power_series_ordinary',)
+    assert classify_ode(eq) == ('2nd_hypergeometric', '2nd_hypergeometric_Integral',
+    '2nd_power_series_ordinary')
+
     sol = Eq(f(x), C2*(-x**4/3 + x**2 + 1) + C1*x + O(x**6))
```
Mohitbalwani26, line 895: @oscarbenjamin I am not sure about this example whether it should match this equation with `2nd_hypergeometric` hint. Although it returns a solution when called with dsolve.
```python
>>> eq = (1 + x**2)*(f(x).diff(x, 2)) + 2*x*(f(x).diff(x)) -2*f(x)
>>> dsolve(eq)
Eq(f(x), sqrt(x**2 + 1)*(C1*sqrt(x**2/(x**2 + 1))*hyper((0, 0), (3/2,), x**2/(x**2 + 1)) + C2*hyper((-1/2, -1/2), (1/2,), x**2/(x**2 + 1)))*(x**2)**(1/4)/sqrt(x))
```
But when I call checkodesol(eq, sol), It hangs.

According to earlier matching code, it has conditions such that whener a example has `2nd_power_series_ordinary` as a hint, then it will not check for `2nd_hypergeometric`

oscarbenjamin: We basically never want the series solution if one of the other solvers can give a non-series solution so the earlier matching code was incorrect.

oscarbenjamin:
I think that maybe it would be better to put the hypergeometric solver as a bunch of standalone functions in a separate file. This implements a published algorithm and the functions should be available for the different steps of the algorithm. There could be a small class in `single.py` that calls those functions so that the algorithm can be used as part of `dsolve`.

Basically I don't think that these classes in `single.py` should have much non-trivial code in them. They are just wrappers that package up the different solution methods with a consistent interface for `classify_ode/dsolve` to use.

Mohitbalwani26:
> I think that maybe it would be better to put the hypergeometric solver as a bunch of standalone functions in a separate file. This implements a published algorithm and the functions should be available for the different steps of the algorithm. There could be a small class in `single.py` that calls those functions so that the algorithm can be used as part of `dsolve`.
> 
> Basically I don't think that these classes in `single.py` should have much non-trivial code in them. They are just wrappers that package up the different solution methods with a consistent interface for `classify_ode/dsolve` to use.

so what should the new file be called? 
That file should contain methods `equivalence` , `match_2nd_2F1_hypergeometric` and `equivalence_hypergeometric` right?

oscarbenjamin:
> so what should the new file be called?

It can be called hypergeometric.py

> That file should contain methods `equivalence` , `match_2nd_2F1_hypergeometric` and `equivalence_hypergeometric` right?

Yes, the original functions. The class in `single.py` should be kept as minimal as possible.

sympy/solvers/ode/single.py
```diff
@@ -1889,5 +1889,92 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
         return fsol
 
 
+class Hypergeometric2nd(SingleODESolver):
```
oscarbenjamin, line 1892: Most of the code in this class should be in functions in hypergeometric.py. The class should be completely minimal.

sympy/solvers/ode/single.py
```diff
@@ -1889,5 +1889,92 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
...
+
+    def _matches(self):
+        eq = self.ode_problem.eq_preprocessed
+        func = self.ode_problem.func
+        x = self.ode_problem.sym
+        df = func.diff(x)
+        a3 = Wild('a3', exclude=[func, func.diff(x), func.diff(x, 2)])
+        b3 = Wild('b3', exclude=[func, func.diff(x), func.diff(x, 2)])
+        c3 = Wild('c3', exclude=[func, func.diff(x), func.diff(x, 2)])
+        deq = a3*(func.diff(x, 2)) + b3*df + c3*func
```
oscarbenjamin, line 1904: The code for getting the coefficients of a linear ODE should not just be in this particular solver. Many solvers need it.

Mohitbalwani26: In `SinglePatternODESolver`, there is `wilds()` method, so should I use pattern matching for this?

Mohitbalwani26:
@oscarbenjamin Now, I guess it would be easier to implment other types of 2ndHypergeometric as one has to implement `match_0F1hyper() `and `get_sol0F1()` in hypergeometric.py and add conditions in ` 2ndHypergeometric` class.

sympy/solvers/ode/single.py
oscarbenjamin: I think that the ODE problem class should have a method for getting the coefficients of a linear ODE:
https://github.com/sympy/sympy/blob/ce8b52f462ac1017f37bfef944c095a0b5756be0/sympy/solvers/ode/single.py#L144-L148

oscarbenjamin:
The hypergeometric.py module needs to have some kind of docstring or comments at the top that explains what it is including a reference for the algorithm and an explanation of which parts are implemented. Also `Hypergeometric2nd` needs a docstring.

Is this method included in the docs at all?

sympy/solvers/ode/single.py
Mohitbalwani26: > The hypergeometric.py module needs to have some kind of docstring or comments at the top that explains what it is including a reference for the algorithm and an explanation of which parts are implemented. Also `Hypergeometric2nd` needs a docstring.
> 
> Is this method included in the docs at all?

No, there isn't any docstring related to hypergeometric, there were only few comments in the code to make it readable.

Mohitbalwani26:
> The hypergeometric.py module needs to have some kind of docstring or comments at the top that explains what it is including a reference for the algorithm and an explanation of which parts are implemented. Also `Hypergeometric2nd` needs a docstring.
> 
> Is this method included in the docs at all?

No, there isn't any docstring related to hypergeometric, there were only few comments in the code to make it readable.

oscarbenjamin:
> No, there isn't any docstring related to hypergeometric, there were only few comments in the code to make it readable.

Can you add something? The comment basically explains it. The algorithm is from a paper and only the 2F1 case is implemented but the paper describes other cases that are not yet implemented. The docstring could show an example solving the general form of a 2F1 ODE but should explain that the method finds all ODEs that can be transformed to that case via the transformations listed in the paper.

Mohitbalwani26:
> > No, there isn't any docstring related to hypergeometric, there were only few comments in the code to make it readable.
> 
> Can you add something? The comment basically explains it. The algorithm is from a paper and only the 2F1 case is implemented but the paper describes other cases that are not yet implemented. The docstring could show an example solving the general form of a 2F1 ODE but should explain that the method finds all ODEs that can be transformed to that case via the transformations listed in the paper.

I tried to go through this [paper](https://www.researchgate.net/publication/2091304_Non-Liouvillian_solutions_for_second_order_linear_ODEs) but couldn't understand much, will it be possible for you to highlight some points or equations that i should include?

oscarbenjamin:
The key equations are 1, 3 and 9. The first equation in 3 would make a good example for a doctest. The main points are just that

1. Currently only 2F1 is supported but the approach could be extended for 1F1 or 0F1.
2. The algorithm should find all solutions of the form 9 if they exist.

asmeurer:
Just to be clear, you shouldn't try to extend the algorithm beyond what is currently implemented, just document what is there. 

As an aside, I wonder if we couldn't implement a solver for the generalized hypergeometric function https://en.wikipedia.org/wiki/Generalized_hypergeometric_function#Differentiation, or even the Meijer G-function https://en.wikipedia.org/wiki/Meijer_G-function#Differential_equation. The solver could then use `hyperexpand` to rewrite the solution in terms of ordinary functions.

oscarbenjamin:
> As an aside, I wonder if we couldn't implement a solver for the generalized hypergeometric function

It's easy to define a solver that matches the exact form. The tricky part is defining an efficient algorithm that can transform related equations to that form.

oscarbenjamin:
The hypergeometric.py module also needs a docstring or a long comment at the top explaining what it is with reference to the paper. The most important information is the reference to the paper and a statement about the status of the code i.e. that the the algorithm from the paper is not fully implemented because only the 2F1 case is handled but the code could be extended to handle 0F1 and 1F1 as described in the paper.

The docstring you have added suggests that the solver works for 0F1 and 1F1 ODEs when it does not. It should state clearly that although the paper describes an algorithm that should find all solutions of the form of equation 9 for 2F1, 0F1 and 1F1 only the 2F1 case is implemented in sympy right now.

sympy/solvers/ode/single.py
```diff
@@ -1889,5 +1889,87 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
...
+
+    .. math:: y'' + A(x) y' + B(x) y = 0\text{,}
+
+    where `A'/2 + A^2/4 - B`  is a rational function of x.
+
+    These kinds of differential equations have solution of non-Liouvillian form.
+
+    Given linear ODE can be obtained from 2F1 given by
+
+    .. math:: (x^2 - x) y'' + ((a + b + 1) x - c) y' + b a y = 0\text{,}
```
oscarbenjamin, line 1907: I think this would make a better example for the doctest.

sympy/solvers/ode/single.py
```diff
@@ -1889,5 +1889,87 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
...
+class Hypergeometric2nd(SingleODESolver):
+    r"""
+    Solves 2nd order linear differential equations.
+
+    It computes special function solutions which can be expressed using the
+    2F1, 1F1 or 0F1 hypergeometric functions.
+
+    .. math:: y'' + A(x) y' + B(x) y = 0\text{,}
+
+    where `A'/2 + A^2/4 - B`  is a rational function of x.
```
oscarbenjamin, line 1901: This should say that `A` and `B` are rational functions of `x`.

sympy/solvers/ode/single.py
```diff
@@ -1889,5 +1889,87 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
...
+
+    where {a, b, c} are arbitrary constants.
+
+    A note: Although the paper in reference describes an algorithm that should find all solutions
+    of the form of below equation for 2F1, 0F1 and 1F1 only the 2F1 case is implemented in
+    Sympy right now.
+
+    .. math:: y = P(x) _pF_q(..; ..;\frac{\alpha x^k + \beta}{\gamma x^k + \delta})\text{,}
+
+    where pFq is any of 2F1, 1F1 or 0F1.
```
oscarbenjamin, line 1917: This should go in a
```
Notes
=====
```
section and should be worded better. The equation should be part of a sentence.

sympy/solvers/ode/hypergeometric.py
```diff
@@ -0,0 +1,249 @@
...
+            r = collect(eq, [func.diff(x, 2), func.diff(x), func]).match(deq)
+
+    if r and r[a3]!=0:
+        A = cancel(r[b3]/r[a3])
+        B = cancel(r[c3]/r[a3])
+        return [A, B]
+    else:
+        return []
+
+def equivalence_hypergeometric(A, B, func):
```
oscarbenjamin, line 44: There should be two spaces here and also in other places. Currently the spacing is inconsistent in this file.

sympy/solvers/ode/single.py
Mohitbalwani26: Yes, I tried that but only classify_ode works fine for this. dsolve returns "This equation cannot be solved by hypergeometric"

sympy/solvers/ode/hypergeometric.py
```diff
@@ -0,0 +1,249 @@
+r'''
+This File contains all the matching and solution method for hypergeometric solver. Main reference
+for this algorithm is taken from  "Non-Liouvillian solutions for second order linear ODEs" by
+L. Chan, E.S. Cheb-Terrab. Although the paper describes an algorithm that should find all solutions
+with 2F1, 0F1 and 1F1,but Sympy only has 2F1 right now.
+The code could be extended to handle 0F1 and 1F1 as described in the paper.
+'''
```
oscarbenjamin, line 7: This is not well written. Capitalisation is inconsistent. There should be a space after a comma.
```
This module contains the implementation of the 2nd_hypergeometric hint for
dsolve. This is an incomplete implementation of the algorithm described in [1].
The algorithm solves 2nd order linear ODEs of the form

.. math:: y'' + A(x) y' + B(x) y = 0\text{,}

where `A` and `B` are rational functions. The algorithm should find any
solution of the form

.. math:: y = P(x) _pF_q(..; ..;\frac{\alpha x^k + \beta}{\gamma x^k + \delta})\text{,}

where pFq is any of 2F1, 1F1 or 0F1 and `P` is an "arbitrary function".
Currently only the 2F1 case is implemented in SymPy but the other cases are
described in the paper and could be implemented in future (contributions
welcome!).

References
==========

.. [1] L. Chan, E.S. Cheb-Terrab, Non-Liouvillian solutions for second order
       linear ODEs, (2004).
       https://arxiv.org/abs/math-ph/0402063
```

sympy/solvers/ode/single.py
oscarbenjamin: Oh well that's a bug. Can you open an issue? Maybe CC @RituRajSingh878 in the issue

Mohitbalwani26:
@oscarbenjamin I have opened an issue as discussed.

sympy/solvers/ode/hypergeometric.py
Mohitbalwani26: I have fixed this in the latest commit of this PR, But I don't know why it is not showing this review 'outdated'.🤔

oscarbenjamin:
Looks good, thanks!

PR #21633, opened 2021-06-17
fix(matrices): don't use mutable _add attribute of Matrix

oscarbenjamin:

Extracted from #21626

Don't use the mutable `_mat` attribute of `Matrix`. A new `._flat` attribute that returns a copy of the internal list is added and all places that used `_mat` are changed to use that. This is in preparation for replacing the internal `_mat` attribute with a DomainMatrix that would not be mutable in the same way.

#### Other comments

Although this means that accessing `self._flat` now invokes a call to list calling list on a list is very fast so this is unlikely to be slow in any significant usage. Ultimately the plan is that all of these places that currently access `._mat` will be superseded by calls to `DomainMatrix` methods.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -714,7 +714,7 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     temp = []
     for e in exprs:
         if isinstance(e, (Matrix, ImmutableMatrix)):
-            temp.append(Tuple(*e._flat))
+            temp.append(Tuple(*e[:]))
         elif isinstance(e, (SparseMatrix, ImmutableSparseMatrix)):
             temp.append(Tuple(*e._smat.items()))
         else:
--- a/sympy/simplify/cse_main.py
+++ b/sympy/simplify/cse_main.py
@@ -714,7 +714,7 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     temp = []
     for e in exprs:
         if isinstance(e, (Matrix, ImmutableMatrix)):
-            temp.append(Tuple(*e._mat))
+            temp.append(Tuple(*e._flat))
         elif isinstance(e, (SparseMatrix, ImmutableSparseMatrix)):
             temp.append(Tuple(*e._smat.items()))
         else:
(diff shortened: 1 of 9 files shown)


asmeurer:
Does this apply to both mutable and immutable matrices?

sympy/simplify/cse_main.py
```diff
@@ -714,7 +714,7 @@ def cse(exprs, symbols=None, optimizations=None, postprocess=None,
     temp = []
     for e in exprs:
         if isinstance(e, (Matrix, ImmutableMatrix)):
-            temp.append(Tuple(*e._mat))
+            temp.append(Tuple(*e._flat))
```
asmeurer, line 717: Perhaps this attribute shouldn't be used at all outside of the matrices submodule. `e[:]` appears to give the same thing.

oscarbenjamin: Agreed, I've changed it to that.

It should be possible to use just `Tuple(*e)` here but that's very slow for large matrices. Apparently `iter(Matrix)` uses the `__len__/__getitem__` version of the iterator protocol which is slow because `__getitem__` is slow. It would be better to add something like:
```python
def __iter__(self):
    return iter(self[:])  # return a list iterator
```
I'm not even sure what class things like that belong in though because there are so many different matrix classes.

oscarbenjamin:
Any further reviews on this?

I want to push ahead with #21626 so I want this one out of the way if possible.

sylee957:
I think that this can be merged.
I'm not sure if the new `_flat` attribute should stay forever because it looks redundant, but if this is needed for subsequent PRs, it can be a reason

oscarbenjamin:
I'm not sure about this any more. The `_flat` attribute was added here just to make clear that it is returning a separate list rather than one that can be used to mutate the matrix.

Maybe actually `._mat` should be kept but as a property that returns a new list perhaps giving some kind of warning that it is private and can no longer be used to mutate the matrix. Most of these places that use `._mat` should probably be changed to use DomainMatrix methods anyway. In some cases it might be better if there was a public way to access a flattened list of the elements.

oscarbenjamin:
This is not needed since #21626 was merged

PR #21634, opened 2021-06-18
Improve `TransferFunction` docs and add `to_expr`

akshanshbhatt:
* Updated the documentation of `TransferFunction`. Made it more beginner-friendly.
* Added a method `to_expr()` for conversion from `TransferFunction` type to `Expr` along with unit tests.
* Added a classmethod for instantiating `TransferFunction` objects from rational expressions (`from_rational_expression`).
#### Release Notes

* physics.control
  * `to_expr` method added in `TransferFunction` for converting `TransferFunction` to sympy expr.
  * `from_rational_expression` classmethod added for creating `TransferFunction` objects directly from rational expressions.

### DIFF
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -75,10 +75,10 @@ class TransferFunction(Basic, EvalfMixin):
 
             $H(s) = \frac{Y(s)}{X(s)} = \frac{ \mathcal{L}\left\{y(t)\right\} }{ \mathcal{L}\left\{x(t)\right\} }$
 
-    $s$, also known as complex frequency, is a complex variable in the the Laplace domain. It corresponds to the
-    equivalent variable $t$, in the time domain. Independent variable $t$ gets transformed to $s$ after Laplace
-    transform. To get back to the time domain from Laplace domain, Inverse-Laplace transform is used.
-    Transfer function, $H$, is generally given as a rational function in $s$ as-
+    $s$, also known as complex frequency, is a complex variable in the Laplace domain. It corresponds to the
+    equivalent variable $t$, in the time domain. Transfer functions are sometimes also referred as Laplace
+    transform of the system's impulse response. Transfer function, $H$, is generally given as a rational
+    function in $s$ as-
 
             $H(s) =\ \frac{a_{n}s^{n}+a_{n-1}s^{n-1}+\dots+a_{1}s+a_{0}}{b_{m}s^{m}+b_{m-1}s^{m-1}+\dots+b_{1}s+b_{0}}$
 
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -1,7 +1,7 @@
 from sympy import Basic, Mul, Pow, degree, Symbol, expand, cancel, Expr, exp, roots
 from sympy.core.evalf import EvalfMixin
 from sympy.core.logic import fuzzy_and
-from sympy.core.numbers import Integer
+from sympy.core.numbers import Integer, ComplexInfinity
 from sympy.core.sympify import sympify, _sympify
 from sympy.polys import Poly, rootof
 from sympy.series import limit
@@ -223,6 +223,80 @@ class TransferFunction(Basic, EvalfMixin):
 
         else:
             raise TypeError("Unsupported type for numerator or denominator of TransferFunction.")
+
+    @classmethod
+    def from_rational_expression(cls, expr, var=None):
+        r"""
+        Allows an alternative way for the users to instantiate a ``TransferFunction`` object.
+        ``TransferFunction`` objects can be created simply by passing the rational expression.
+        SymPy will smartly create a ``TransferFunction`` object with the properties of the
+        expression.
+
+        Parameters
+        ==========
+        expr : Expr, Number
+            The rational expression representing the ``TransferFunction``.
+        var : Symbol, optional
+            Complex variable of the Laplace transform used by the
+            polynomials of the transfer function.
+
+        Raises
+        ======
+
+        ValueError
+            When ``expr`` is of type ``Number`` and optional parameter ``var``
+            is not passed.
+
+            Wen ``expr`` has more than one variables and optional parameter
+            ``var`` is not passed.
+        ZeroDivisionError
+            When denominator of ``expr`` is zero or it has ``ComplexInfinity``
+            in its numerator.
+
+        Examples
+        ========
+
+        >>> from sympy.abc import s, p, a
+        >>> from sympy.physics.control.lti import TransferFunction
+        >>> expr1 = (s + 5)/(3*s**2 + 2*s + 1)
+        >>> tf1 = TransferFunction.from_rational_expression(expr1)
+        >>> tf1
+        TransferFunction(s + 5, 3*s**2 + 2*s + 1, s)
+        >>> expr2 = (a*p**3 - a*p**2 + s*p)/(p + a**2)  # Expr with more than one variables
+        >>> tf2 = TransferFunction.from_rational_expression(expr2, p)
+        >>> tf2
+        TransferFunction(a*p**3 - a*p**2 + p*s, a**2 + p, p)
+
+        In case of conflict between two or more variables in a expression, SymPy will
+        raise a ``ValueError``, if ``var`` is not passed by the user.
+
+        >>> tf = TransferFunction.from_rational_expression((a + a*s)/(s**2 + s + 1))
+        Traceback (most recent call last):
+        ...
+        ValueError: Conflicting values found for positional argument `var` ({a, s}). Specify it manually.
+
+        This can be corrected by specifying the ``var`` parameter manually.
+
+        >>> tf = TransferFunction.from_rational_expression((a + a*s)/(s**2 + s + 1), s)
+        >>> tf
+        TransferFunction(a*s + a, s**2 + s + 1, s)
+
+        """
+        expr = _sympify(expr)
+        if var is None:
+            _free_symbols = expr.free_symbols
+            _len_free_symbols = len(_free_symbols)
+            if _len_free_symbols == 1:
+                var = list(_free_symbols)[0]
+            elif _len_free_symbols == 0:
+                raise ValueError("Positional argument `var` not found in the TransferFunction defined. Specify it manually.")
+            else:
+                raise ValueError("Conflicting values found for positional argument `var` ({}). Specify it manually.".format(_free_symbols))
+
+        _num, _den = expr.as_numer_denom()
+        if _den == 0 or _num.has(ComplexInfinity):
+            raise ZeroDivisionError("TransferFunction can't have a zero denominator.")
+        return cls(_num, _den, var)
 
     @property
     def num(self):
@@ -587,8 +661,30 @@ class TransferFunction(Basic, EvalfMixin):
         """
         return degree(self.num, self.var) == degree(self.den, self.var)
 
-    def _to_expr(self):
-        """To convert TransferFunction type to SymPy Expr"""
+    def to_expr(self):
+        """
+        To convert ``TransferFunction`` object to SymPy Expr.
+
+        Examples
+        ========
+
+        >>> from sympy.abc import s, p, a, b
+        >>> from sympy.physics.control.lti import TransferFunction
+        >>> from sympy.core.expr import Expr
+        >>> tf1 = TransferFunction(s, a*s**2 + 1, s)
+        >>> tf1.to_expr()
+        s/(a*s**2 + 1)
+        >>> isinstance(_, Expr)
+        True
+        >>> tf2 = TransferFunction(1, (p + 3*b)*(b - p), p)
+        >>> tf2.to_expr()
+        1/((b - p)*(3*b + p))
+        >>> tf3 = TransferFunction((s - 2)*(s - 3), (s - 1)*(s - 2)*(s - 3), s)
+        >>> tf3.to_expr()
+        ((s - 3)*(s - 2))/(((s - 3)*(s - 2)*(s - 1)))
+
+        """
+
         return Mul(self.num, Pow(self.den, -1, evaluate=False), evaluate=False)
 
 
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -99,7 +99,7 @@ class TransferFunction(Basic, EvalfMixin):
     TypeError
         When ``var`` is not a Symbol or when ``num`` or ``den`` is not a
         number or a polynomial.
-    ZeroDivisionError
+    ValueError
         When ``den`` is zero.
 
     Examples
@@ -209,7 +209,7 @@ class TransferFunction(Basic, EvalfMixin):
             raise TypeError("Variable input must be a Symbol.")
 
         if den == 0:
-            raise ZeroDivisionError("TransferFunction can't have a zero denominator.")
+            raise ValueError("TransferFunction can't have a zero denominator.")
 
         if (((isinstance(num, Expr) and num.has(Symbol) and not num.has(exp)) or num.is_number) and
             ((isinstance(den, Expr) and den.has(Symbol) and not den.has(exp)) or den.is_number)):
@@ -414,7 +414,7 @@ class TransferFunction(Basic, EvalfMixin):
             arg_list = list(other.args)
             return Parallel(self, *arg_list)
         else:
-            raise TypeError("TransferFunction cannot be added with {}.".
+            raise ValueError("TransferFunction cannot be added with {}.".
                 format(type(other)))
 
     def __radd__(self, other):
@@ -433,7 +433,7 @@ class TransferFunction(Basic, EvalfMixin):
             arg_list = [-i for i in list(other.args)]
             return Parallel(self, *arg_list)
         else:
-            raise TypeError("{} cannot be subtracted from a TransferFunction."
+            raise ValueError("{} cannot be subtracted from a TransferFunction."
                 .format(type(other)))
 
     def __rsub__(self, other):
@@ -452,7 +452,7 @@ class TransferFunction(Basic, EvalfMixin):
             arg_list = list(other.args)
             return Series(self, *arg_list)
         else:
-            raise TypeError("TransferFunction cannot be multiplied with {}."
+            raise ValueError("TransferFunction cannot be multiplied with {}."
                 .format(type(other)))
 
     __rmul__ = __mul__
@@ -480,7 +480,7 @@ class TransferFunction(Basic, EvalfMixin):
             else:
                 return Feedback(self, Series(*other_arg_list))
         else:
-            raise TypeError("TransferFunction cannot be divided by {}.".
+            raise ValueError("TransferFunction cannot be divided by {}.".
                 format(type(other)))
 
     __rtruediv__ = __truediv__
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -19,14 +19,65 @@ def _roots(poly, var):
 
 
 class TransferFunction(Basic, EvalfMixin):
-    """
+    r"""
     A class for representing LTI (Linear, time-invariant) systems that can be strictly described
-    by ratio of polynomials in the Laplace Transform complex variable. The arguments
+    by ratio of polynomials in the Laplace transform complex variable. The arguments
     are ``num``, ``den``, and ``var``, where ``num`` and ``den`` are numerator and
     denominator polynomials of the ``TransferFunction`` respectively, and the third argument is
     a complex variable of the Laplace transform used by these polynomials of the transfer function.
     ``num`` and ``den`` can be either polynomials or numbers, whereas ``var``
     has to be a Symbol.
+
+    Explanation
+    ===========
+
+    Generally, a dynamical system respresenting a physical model can be described in terms of Linear
+    Ordinary Differential Equations like -
+
+            $\small{b_{m}y^{\left(m\right)}+b_{m-1}y^{\left(m-1\right)}+\dots+b_{1}y^{\left(1\right)}+b_{0}y=
+            a_{n}x^{\left(n\right)}+a_{n-1}x^{\left(n-1\right)}+\dots+a_{1}x^{\left(1\right)}+a_{0}x}$
+
+    Here, $x$ is the input signal and $y$ is the output signal and superscript on both is the order of derivative
+    (not exponent). Derivative is taken with respect to the independent variable, $t$. Also, generally $m$ is greater
+    than $n$.
+
+    It is not feasible to analyse the properties of such systems in their native form therefore, we use
+    mathematical tools like Laplace transform to get a better perspective. Taking the Laplace transform
+    of both the sides in the equation (at zero initial conditions), we get -
+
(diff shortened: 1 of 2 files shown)


moorepants:
Almost everything here is a backwards incompatible change. I think TransferFunction is already in a number of sympy releases. I recommend not changing any of the errors. As for expand, I'd leave the method but make it return the already expanded form.

You don't have an explanation as to why expand should be changed in this way. What is the use case? Why should the expressions be expanded automatically? It isn't really a sympy idiom to do so.

Lastly, I didn't see any tests for the new _to_expr method.

sympy/physics/control/tests/test_lti.py
```diff
@@ -263,6 +249,15 @@ def test_TransferFunction_functions():
...
 
+    # Conversion to Expr with _to_expr()
+    tf8 = TransferFunction(a0*s**5 + 5*s**2 + 3, s**6 - 3, s)
+    tf9 = TransferFunction((5 + s), (5 + s)*(6 + s), s)
+    tf10 = TransferFunction(0, 1, s)
+    tf11 = TransferFunction(1, 1, s)
+    assert tf8._to_expr() == Mul((a0*s**5 + 5*s**2 + 3), Pow((s**6 - 3), -1, evaluate=False), evaluate=False)
+    assert tf9._to_expr() == Mul((s + 5), Pow((s**2 + 11*s + 30), -1, evaluate=False), evaluate=False)
+    assert tf10._to_expr() == Mul(sympify(0), Pow(simplify(1), -1, evaluate=False), evaluate=False)
+    assert tf11._to_expr() == Mul(sympify(1), Pow(simplify(1), -1, evaluate=False), evaluate=False)
```
moorepants, line 260: I missed these on first pass. Scratch my other comment.

moorepants: What is the purpose of simplify and sympify here? You can use `S(0)` to get a SymPy zero if needed.

akshanshbhatt:
> You don't have an explanation as to why expand should be changed in this way. What is the use case? Why should the expressions be expanded automatically? It isn't really a sympy idiom to do so.

@namannimmo10 and I discussed implementing `expand()` by default (https://github.com/sympy/sympy/pull/21556#discussion_r650993372). I provided some use cases and practices used in other popular control packages to make the form of `TransferFunction` more coherent. `expand()` by default was also approved by you in the call we had. I have implemented it after getting approval from both Naman and you.

sympy/physics/control/lti.py
```diff
@@ -394,7 +430,7 @@ def __sub__(self, other):
             arg_list = [-i for i in list(other.args)]
             return Parallel(self, *arg_list)
         else:
-            raise ValueError("{} cannot be subtracted from a TransferFunction."
+            raise TypeError("{} cannot be subtracted from a TransferFunction."
```
namannimmo10, line 433:
```suggestion
            raise ValueError("{} cannot be subtracted from a TransferFunction."
```
This should remain a `ValueError`, no? `TypeError` is raised when an operation or a function is applied to an object of inappropriate type.

sympy/physics/control/lti.py
```diff
@@ -19,15 +19,66 @@ def _roots(poly, var):
...
     are ``num``, ``den``, and ``var``, where ``num`` and ``den`` are numerator and
     denominator polynomials of the ``TransferFunction`` respectively, and the third argument is
     a complex variable of the Laplace transform used by these polynomials of the transfer function.
     ``num`` and ``den`` can be either polynomials or numbers, whereas ``var``
     has to be a Symbol.
 
+    Explanation
+    ===========
+
+    Generally, a dynamical system respresenting a physical model can be described in terms of Linear
```
namannimmo10, line 34:
```suggestion
    Generally, a dynamical system representing a physical model can be described in terms of Linear
```

sympy/physics/control/lti.py
```diff
@@ -19,15 +19,66 @@ def _roots(poly, var):
...
+    function.
+
+    The numerator of the transfer function is therefore, the Laplace transform of the output signal
+    (The signals are represented as functions of time) and similarly the denominator
+    of the transfer function is the Laplace transform of the input signal. It is also a convention
+    to denote the input and output signal's Laplace transform with capital alphabets like shown below.
+
+            $H(s) = \frac{Y(s)}{X(s)} = \frac{ \mathcal{L}\left\{y(t)\right\} }{ \mathcal{L}\left\{x(t)\right\} }$
+
+    Transfer function, $H$, is generally given as a rational function in $s$ as-
```
namannimmo10, line 78: Can you add one line above describing what is $s$?

sympy/physics/control/lti.py
```diff
@@ -526,6 +562,12 @@ def is_biproper(self):
         """
         return degree(self.num, self.var) == degree(self.den, self.var)
 
+    def _to_expr(self):
+        """
+        To convert TransferFunction type to SymPy Expr
+        """
```
namannimmo10, line 568: Make this one line?

namannimmo10: Why private though? This could be a way for users to convert a transfer function to an elementary function.

akshanshbhatt: > Why private though? This could be a way for users to convert a transfer function to an elementary function.

Since the only possible use case, in my view at the time of its implementation, was the internal conversion of `TransferFunction` objects to `Expr` to make them compatible with the native matrix operations (since non-expr operations were deprecated).
If there are other possible use cases that a user might find helpful, I will make it a public method. I will do whatever you suggest.

sympy/physics/control/lti.py
akshanshbhatt: I think it should be `TypeError`. Here the exception is raised when the type of object we are subtracting from the `TransferFunction` is incompatible. The placeholder, too (in the print expression), logs the type of the object passed. Say I try to subtract a `Symbol` object from a `TransferFunction`, it will prompt an error -

```python
>>> from sympy.physics.control.lti import *
>>> from sympy.abc import *
>>> TransferFunction(1, s, s)
TransferFunction(1, s, s)
>>> _ - a
  File "d:\sympy\sympy\physics\control\lti.py", line 436, in __sub__
    raise TypeError("{} cannot be subtracted from a TransferFunction."
TypeError: <class 'sympy.core.symbol.Symbol'> cannot be subtracted from a TransferFunction.
```
It makes sense as the type of object passed is incompatible for tf subtraction. Had this exception been raised when both the params were of the same type, there would have been a possibility for `ValueError`.
I used this as the reference - https://docs.python.org/3/library/exceptions.html#TypeError.

I was anyways, planning to revert to the old error types as suggested by @moorepants since this module has already gone through two release cycles. What is your opinion?

moorepants:
> @namannimmo10 and I discussed implementing expand() by default (#21556 (comment)). I provided some use cases and practices used in other popular control packages to make the form of TransferFunction more coherent. expand() by default was also approved by you in the call we had. I have implemented it after getting approval from both Naman and you.

Yes, it sounded ok in the call, but I still don't know why it is being done. The explanation should be at least in the commit message here. Or show some examples that give an idea why this change is needed. There needs to be a justification for breaking backwards compatibility and not following typical sympy design patterns.

sympy/physics/control/lti.py
moorepants: The errors shouldn't be changed due to backwards incompatibility. It's more trouble for end users to break their code than worring about if the correct error types were originally selected.

akshanshbhatt: > The errors shouldn't be changed due to backwards incompatibility. It's more trouble for end users to break their code than worring about if the correct error types were originally selected.

I will revert the error-type changes.

sympy/physics/control/lti.py
```diff
@@ -19,15 +19,69 @@ def _roots(poly, var):
...
+    The numerator of the transfer function is therefore, the Laplace transform of the output signal
+    (The signals are represented as functions of time) and similarly the denominator
+    of the transfer function is the Laplace transform of the input signal. It is also a convention
+    to denote the input and output signal's Laplace transform with capital alphabets like shown below.
+
+            $H(s) = \frac{Y(s)}{X(s)} = \frac{ \mathcal{L}\left\{y(t)\right\} }{ \mathcal{L}\left\{x(t)\right\} }$
+
+    $s$, also known as complex frequency, is a complex variable in the the Laplace domain. It corresponds to the
+    equivalent variable $t$, in the time domain. Independent variable $t$ gets transformed to $s$ after Laplace
+    transform. To get back to the time domain from Laplace domain, Inverse-Laplace transform is used.
```
namannimmo10, line 80:
```suggestion
    $s$, also known as complex frequency, is a complex variable in the Laplace domain. It corresponds to the
    equivalent variable $t$, in the time domain.
```
This looks sufficient. Since we are explaining this so well, it's also worth mentioning that a transfer function is the Laplace transform of the system's impulse response.

akshanshbhatt: > Since we are explaining this so well, it's also worth mentioning that a transfer function is the Laplace transform of the system's impulse response.

Where exactly should I mention this?

oscarbenjamin:
Are the numerator and denominator always polynomials? If so then they could use Poly which expands automatically. Poly is more efficient than Expr for calculations with expanded polynomials. The internal representation of Poly is just a list of coefficients:
```python
In [3]: Poly(3*s**3 + s + 1).rep.rep
Out[3]: [mpz(3), mpz(0), mpz(1), mpz(1)]
```

I can imagine that it might be more useful in some cases to represent the system in terms of its poles and zeros rather than the expanded polynomial though.

In general I think it would be good to have a rational function class analogous to Poly although maybe it would have automatic cancellation and so not quite fit with the usage here.

namannimmo10:
> Are the numerator and denominator always polynomials?

Actually, no. Last year we decided to support any arbitrary expressions in the numerator and denominator because we can have a [variety of transfer functions](https://tutorial.math.lamar.edu/classes/de/laplace_table.aspx) that aren't the ratio of pure polynomials. So it was not the right choice to use Poly internally.

> I can imagine that it might be more useful in some cases to represent the system in terms of its poles and zeros rather than the expanded polynomial though.

Yes, that was another main reason why I didn't use `expand` by default.
```python
In [3]: num, den = b, (s - a)**2 - b**2

In [4]: pprint(TransferFunction(num, den, s))
       b        
────────────────
   2           2
- b  + (-a + s) 

In [5]: pprint(TransferFunction(num, den, s).expand())
         b          
────────────────────
 2            2    2
a  - 2⋅a⋅s - b  + s 
```
Not sure if using `expand` does anything useful here.

```python
In [9]: _num, _den = factorial(n), (s - a)**(n + 1)

In [10]: pprint(TransferFunction(_num, _den, s))
      n!     
─────────────
        n + 1
(-a + s)     

In [11]: pprint(TransferFunction(_num, _den, s).expand())
             n!            
───────────────────────────
            n             n
- a⋅(-a + s)  + s⋅(-a + s) 

```

But now, even if we decide to expand the numerator and denominator by default, the user can call `factor` to get back to the unexpanded form. Maybe two examples in the docs could be added. 
```python
In [31]: G = TransferFunction(2*a*s**2, (s**2 + a**2)**2, s); pprint(G)
       2  
  2⋅a⋅s   
──────────
         2
⎛ 2    2⎞ 
⎝a  + s ⎠ 

In [32]: _G = TransferFunction(2*a*s**2, (s**2 + a**2)**2, s).expand(); pprint(_G)
           2     
      2⋅a⋅s      
─────────────────
 4      2  2    4
a  + 2⋅a ⋅s  + s 

In [33]: pprint(factor(_G))
       2  
  2⋅a⋅s   
──────────
         2
⎛ 2    2⎞ 
⎝a  + s ⎠ 
```
I would personally stick to the default unexpanded form.

moorepants:
> I would personally stick to the default unexpanded form.

@namannimmo10 why were you suggesting in the meeting to do otherwise?

namannimmo10:
https://github.com/sympy/sympy/pull/21556#discussion_r650887855 convinced me back then. :) Sorry, Akshansh.

akshanshbhatt:
> [#21556 (comment)](https://github.com/sympy/sympy/pull/21556#discussion_r650887855) convinced me back then. :) Sorry, Akshansh.

Well, nothing to be sorry about here. Things that look good in the planning/prototyping period might not work as planned after implementation. Many latent requirements need to be satisfied, which usually come into perspective after the execution. I will change it back to the unexpanded form.

sympy/physics/control/lti.py
namannimmo10:
```python
>>> tf = TransferFunction(s + 1, s**2 + 2, s)
>>> delay = exp(-s/tau)
>>> expr = delay*tf.to_expr()
>>> H1 = TransferFunction.from_rational_expression(expr, s)
>>> 
>>> H2 = TransferFunction(exp(-s/tau)*(s + 1), s**2 + 2, s)
```

Both `H1` and `H2` should work, I guess.

moorepants:
> #21556 (comment) convinced me back then. :) Sorry, Akshansh.

The comment has some good points. In the numeric libraries you pass in coefficients to the polynomial. I'm pretty sure we discussed this when Naman was designing things too and we felt that since we have a symbolic library that method would be too limiting (can't remember the details though). If expand() is expensive to call, then maybe it is worth doing it once on object instantiation. If it isn't expensive then you can get the coefficients whenever you need to with the appropriate function calls. We need a demonstration of why expand() should be called on instantiation, otherwise we shouldn't change it. The only reason presented in the comment is that other libraries in other languages do it differently. But why do they do it differently? What is the benefit?

@akshanshbhatt, you should focus on building on what we have, versus changing what we have already made. If, when building the new things you really need to change TransferFunction, then present the case, explain why and maybe it's worth deprecating things.

moorepants:
You wrote this:

> In my opinion, having a definite form would enable us to compare two TransferFunction instances in later phases.

What are examples of comparing two transfer functions where this is needed?

akshanshbhatt:
> The comment has some good points. In the numeric libraries you pass in coefficients to the polynomial. I'm pretty sure we discussed this when Naman was designing things too and we felt that since we have a symbolic library that method would be too limiting (can't remember the details though). If expand() is expensive to call, then maybe it is worth doing it once on object instantiation. If it isn't expensive then you can get the coefficients whenever you need to with the appropriate function calls. We need a demonstration of why expand() should be called on instantiation, otherwise we shouldn't change it. The only reason presented in the comment is that other libraries in other languages do it differently. But why do they do it differently? What is the benefit?

The purpose was a getting a uniform structure for `TransferFunction` objects, a sort of one-to-one mapping of TransferFunctions no matter the form. Consider these three transfer function objects, 

```
tf1 = TransferFunction(1, s**2 + 2*s + 1, s)
tf2 = TransferFunction(1, (s + 1)**2, s)
tf3 = TransferFunction(1, -s + (s + 1)*(s + 2) - 1, s)
```
Ideally, all these should be equal since all of their properties and attributes are equal (poles, zeros, stability) from the design principle. Yet, they are considered different under our model. 

```python
>>> Parallel(tf1, tf2) - tf3
Parallel(TransferFunction(1, s**2 + 2*s + 1, s), TransferFunction(1, (s + 1)**2, s), TransferFunction(-2, -s + (s + 1)*(s + 2) - 1, s))
>>> _.doit()
TransferFunction(-2*(s + 1)**2*(s**2 + 2*s + 1) + (-s + (s + 1)*(s + 2) - 1)*(s**2 + 2*s + (s + 1)**2 + 1), (s + 1)**2*(-s + (s + 1)*(s + 2) - 1)*(s**2 + 2*s + 1), s)
```

Even after using expand at this stage, the result is not accurate.

```python
>>> _.expand()
TransferFunction(0, s**6 + 6*s**5 + 15*s**4 + 20*s**3 + 15*s**2 + 6*s + 1, s)

# expected
TransferFunction(0, s**2 + 2*s +1, s)
```
Yes, we can use simplify here, but that too would be a resource-consuming choice (in fact, more than `expand()` and also not guaranteed to work).
This was specifically a bad sign since, 

* In `TrasferFunctionMatrix` there are many operations where the same transfer function instances are added and subtracted. Still, since their form is often different (like shown in the example above), it would output something totally gibberish when it should have outputted a simple answer.
* In solving textbook problems, most of the time, it is the case that there is some canceling out of common terms, which ultimately results in simple final answers. Here it won't be happening.

sympy/physics/control/lti.py
akshanshbhatt: > Both `H1` and `H2` should work, I guess.

This example looks fine but in the call which we had, @moorepants suggested not to implement `from_rational_expression()` classmethod.

moorepants:
This is primarily the difference in symoblics and numerics. SymPy, by design, stores symbolics in verbose complex expressions. Symbolic equality is ill-defined mathematically and simplification oriented routines are not typically used in library algorithms, but more as a user tool. The polynomials package has objects for polynomials, but we decided not to use it inside TransferFunctions (for reasons I can't remember). The examples you show are really just simplification to show equality. There isn't anything wrong with it. That's just how sympy works.

moorepants:
> In solving textbook problems, most of the time, it is the case that there is some canceling out of common terms, which ultimately results in simple final answers. Here it won't be happening.

In you examples, use simplification at the end to get a simple form if needed. This is what users will do.

akshanshbhatt:
> This is primarily the difference in symoblics and numerics. SymPy, by design, stores symbolics in verbose complex expressions. Symbolic equality is ill-defined mathematically and simplification oriented routines are not typically used in library algorithms, but more as a user tool. The polynomials package has objects for polynomials, but we decided not to use it inside TransferFunctions (for reasons I can't remember). The examples you show are really just simplification to show equality. There isn't anything wrong with it. That's just how sympy works.

Yes, I understand it now, that's why I have removed it. What I have shared with you was my initial thought, the reason why I brought up expanding the terms by default in the first place.

sympy/physics/control/lti.py
namannimmo10: > ```python
> >>> tf = TransferFunction(s + 1, s**2 + 2, s)
> >>> delay = exp(-s/tau)
> >>> expr = delay*tf.to_expr()
> >>> H1 = TransferFunction.from_rational_expression(expr, s)
> >>> 
> >>> H2 = TransferFunction(exp(-s/tau)*(s + 1), s**2 + 2, s)
> ```

I think it's alright to go with that. I see similar classmethods being used in the codebase. 

There's a `to_expr` method in _holonomic.py_ that converts a holonomic function back to elementary function:
https://github.com/sympy/sympy/blob/8b783d15caae900e05c9236bf001fde908befe7a/sympy/holonomic/holonomic.py#L2066

It would be nice if ours did too. This PR is almost ready after that.

akshanshbhatt: I will complete the docstrings and tests for `from_rational_expression` by today. Hopefully, we can merge this PR after that

sympy/physics/control/lti.py
```diff
@@ -19,15 +19,69 @@ def _roots(poly, var):
...
+    (The signals are represented as functions of time) and similarly the denominator
+    of the transfer function is the Laplace transform of the input signal. It is also a convention
+    to denote the input and output signal's Laplace transform with capital alphabets like shown below.
+
+            $H(s) = \frac{Y(s)}{X(s)} = \frac{ \mathcal{L}\left\{y(t)\right\} }{ \mathcal{L}\left\{x(t)\right\} }$
+
+    $s$, also known as complex frequency, is a complex variable in the Laplace domain. It corresponds to the
+    equivalent variable $t$, in the time domain. Transfer functions are sometimes also referred as Laplace
+    transform of the system's impulse response. Transfer function, $H$, is generally given as a rational
+    function in $s$ as-
```
namannimmo10, line 81:
```suggestion
    The numerator of the transfer function is, therefore, the Laplace transform of the output signal
    (The signals are represented as functions of time) and similarly, the denominator
    of the transfer function is the Laplace transform of the input signal. It is also a convention
    to denote the input and output signal's Laplace transform with capital alphabets like shown below.

            $H(s) = \frac{Y(s)}{X(s)} = \frac{ \mathcal{L}\left\{y(t)\right\} }{ \mathcal{L}\left\{x(t)\right\} }$

    $s$, also known as complex frequency, is a complex variable in the Laplace domain. It corresponds to the
    equivalent variable $t$, in the time domain. Transfer functions are sometimes also referred to as the Laplace
    transform of the system's impulse response. Transfer function, $H$, is represented as a rational
    function in $s$ like,
```

sympy/physics/control/lti.py
```diff
@@ -144,25 +197,107 @@ class TransferFunction(Basic, EvalfMixin):
...
         else:
             raise TypeError("Unsupported type for numerator or denominator of TransferFunction.")
 
+    @classmethod
+    def from_rational_expression(cls, expr, var=None):
+        r"""
+        Allows an alternative way for the users to instantiate a ``TransferFunction`` object.
+        ``TransferFunction`` objects can be created simply by passing the rational expression.
+        SymPy will smartly create a ``TransferFunction`` object with the properties of the
+        expression.
```
namannimmo10, line 233:
```suggestion
        Creates a new ``TransferFunction`` efficiently from a rational expression.
```

sympy/physics/control/lti.py
```diff
@@ -144,25 +197,107 @@ class TransferFunction(Basic, EvalfMixin):
...
+            polynomials of the transfer function.
+
+        Raises
+        ======
+
+        ValueError
+            When ``expr`` is of type ``Number`` and optional parameter ``var``
+            is not passed.
+
+            Wen ``expr`` has more than one variables and optional parameter
```
namannimmo10, line 250:
```suggestion
            When ``expr`` has more than one variables and an optional parameter
```

namannimmo10 (commented):
Minor changes required...but overall looks good.

sympy/physics/control/tests/test_lti.py
```diff
@@ -105,6 +105,26 @@ def test_TransferFunction_construction():
 
 
 def test_TransferFunction_functions():
+    # classmethod from_rational_expression
```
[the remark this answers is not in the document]
akshanshbhatt: Delay is not supported as of now. 

It prompts -

```
TypeError: Unsupported type for numerator or denominator of TransferFunction.
```

It is due to `not num.has(exp)` condition. Should I remove it?

https://github.com/sympy/sympy/blob/8b783d15caae900e05c9236bf001fde908befe7a/sympy/physics/control/lti.py#L155-L158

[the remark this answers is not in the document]
namannimmo10: Yes. Time delay is needed in several LTI models: https://www.mathworks.com/help/control/ug/time-delays-in-linear-systems.html

sympy/physics/control/lti.py
```diff
@@ -144,25 +197,107 @@ class TransferFunction(Basic, EvalfMixin):
...
+    def from_rational_expression(cls, expr, var=None):
+        r"""
+        Allows an alternative way for the users to instantiate a ``TransferFunction`` object.
+        ``TransferFunction`` objects can be created simply by passing the rational expression.
+        SymPy will smartly create a ``TransferFunction`` object with the properties of the
+        expression.
+
+        Parameters
+        ==========
+        expr : Expr, Number
```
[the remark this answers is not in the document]
namannimmo10: Add one example in the docstring for `expr: Number`?

namannimmo10 (approved):
Looks ready! @moorepants, you have any other comments?

namannimmo10:
I will merge this later tonight. Thanks for your contributions, @akshanshbhatt!

PR #21637, opened 2021-06-19
feat(matrices): add DomainMatrix.extract method

oscarbenjamin:

This is a spin out from #21626

Add `DomainMatrix.extract` method analogous to `Matrix.extract`. This is needed because otherwise changing `Matrix.extract` to use `._flat` instead of `._mat` leads to slowdowns as seen in the `sympy/matrices/tests/test_eigen.py::test_issue_8240` test which uses `extract` many times to select small 1x1 submatrices from a large matrix.

#### Release Notes

NO ENTRY


oscarbenjamin:
It seems this doesn't handle negative indices...

oscarbenjamin:
> It seems this doesn't handle negative indices...

All fixed

PR #21639, opened 2021-06-19
caveat about autowrapping blindly

### LINKED ISSUE #6548: Autowrap ((x - y)**25).expand() is a bad example
This was originally https://github.com/sympy/sympy/issues/1593 .

Following the example given in the documentation:

In [1]: from sympy.utilities.autowrap import autowrap
In [2]: from sympy.abc import x,y
In [3]: expr = ((x - y)**(25)).expand()
In [4]: fn = autowrap(expr, language='C', backend='Cython')
In [5]: fn(3, 2)
Out[5]: 0.0
Which is incorrect (subs can be used to trivially show that the correct result is 1). This is simply a limitation of double precision.

I therefore suggest that something more reasonable is selected given that the current example may give the (incorrect) impression that the autowrapping system is broken. Or, better still, use it as an example of a caveat of using autowrap.

Original issue for #6548: http://code.google.com/p/sympy/issues/detail?id=3449

smichr:

closes #6548 by adding warning that autowrapping cannot be done without some insight into numerics

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/utilities/autowrap.py
+++ b/sympy/utilities/autowrap.py
@@ -8,13 +8,33 @@ python.
 This module provides a common interface for different external backends, such
 as f2py, fwrap, Cython, SWIG(?) etc. (Currently only f2py and Cython are
 implemented) The goal is to provide access to compiled binaries of acceptable
-performance with a one-button user interface, i.e.
+performance with a one-button user interface, e.g.,
 
     >>> from sympy.abc import x,y
-    >>> expr = ((x - y)**(25)).expand()
-    >>> binary_callable = autowrap(expr)
-    >>> binary_callable(1, 2)
+    >>> expr = (x - y)**25
+    >>> flat = expr.expand()
+    >>> binary_callable = autowrap(flat)
+    >>> binary_callable(2, 3)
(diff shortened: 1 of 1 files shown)


sympy/utilities/autowrap.py
```diff
@@ -8,14 +8,34 @@
...
+    >>> expr = (x - y)**25
+    >>> flat = expr.expand()
+    >>> binary_callable = autowrap(flat)
+    >>> binary_callable(2, 3)
     -1.0
 
+Although a SymPy user might primarily be interested in working with
+mathematical expressions and not about the details of wrapping tools
+in order to evaluate expressions numerically which are expensive to
+calculate, the user cannot do so without some understanding of the
```
smichr, line 23:
```suggestion
mathematical expressions and not in the details of wrapping tools
needed to evaluate such expressions efficiently in numerical form,
the user cannot do so without some understanding of the
```

smichr: I'll wait to commit this change to see if there are other suggestions so they can be committed together.

Issue #19202, opened 2020-04-26
[Discussion]: Plotting functions for Beam3D


Psycho-Pirate:
We currently don't have any functions for plotting shear force, bending moment, slope, and deflection for 3D beams.
I would like to add these functions as I have described in my GSoC Proposal [here](https://docs.google.com/document/d/1sYoj5F6A9tZEROYa7el6D6gqgqwRsX2qf6zX0Ejlu7Q/edit#heading=h.elaq5ykfk86c)
@moorepants @jashan498 @ishanaj

Psycho-Pirate:
I have added a function for plotting shear force.
If this approach is approved, I will add other functions as well.

Psycho-Pirate:
Please have a look @ishanaj @Sc0rpi0n101

PR #21648, opened 2021-06-22
[GSoC] Function for plotting loading results added

Psycho-Pirate:

See #19202

A function for plotting all loading results is added in Beam3D class.

#### Release Notes

* physics.continuum_mechanics
  * Added ability to plot all loading results in Beam3D class

### DIFF
--- a/sympy/physics/continuum_mechanics/beam.py
+++ b/sympy/physics/continuum_mechanics/beam.py
@@ -2667,3 +2667,104 @@ class Beam3D(Beam):
             Py = self._plot_deflection('y')
             Pz = self._plot_deflection('z')
             return PlotGrid(3, 1, Px, Py, Pz)
+
+    def plot_loading_results(self, dir=None, subs=None):
+
+        """
+
+        Returns a subplot of Shear Force, Bending Moment,
+        Slope and Deflection of the Beam object along the direction specified.
+
+        Parameters
+        ==========
+
+        dir : string
+               Direction along which plots are required.
+               It is necessary to specify the direction for this function.
+        subs : dictionary
+               Python dictionary containing Symbols as key and their
+               corresponding values.
+
+        Examples
+        ========
+        There is a beam of length 20 meters. It it supported by rollers
+        at of its end. A linear load having slope equal to 12 is applied
+        along y-axis. A constant distributed load of magnitude 15 N is
+        applied from start till its end along z-axis.
+
+        .. plot::
+            :context: close-figs
+            :format: doctest
+            :include-source: True
+
+            >>> from sympy.physics.continuum_mechanics.beam import Beam3D
+            >>> from sympy import symbols
+            >>> l, E, G, I, A, x = symbols('l, E, G, I, A, x')
+            >>> b = Beam3D(20, 40, 21, 100, 25, x)
+            >>> b.apply_load(15, start=0, order=0, dir="z")
+            >>> b.apply_load(12*x, start=0, order=0, dir="y")
+            >>> b.bc_deflection = [(0, [0, 0, 0]), (20, [0, 0, 0])]
+            >>> R1, R2, R3, R4 = symbols('R1, R2, R3, R4')
+            >>> b.apply_load(R1, start=0, order=-1, dir="z")
+            >>> b.apply_load(R2, start=20, order=-1, dir="z")
+            >>> b.apply_load(R3, start=0, order=-1, dir="y")
+            >>> b.apply_load(R4, start=20, order=-1, dir="y")
+            >>> b.solve_for_reaction_loads(R1, R2, R3, R4)
+            >>> b.solve_slope_deflection()
+            >>> b.plot_loading_results('y')
+            PlotGrid object containing:
+            Plot[0]:Plot object containing:
+            [0]: cartesian line: -6*x**2 for x over (0.0, 20.0)
+            Plot[1]:Plot object containing:
+            [0]: cartesian line: -15*x**2/2 for x over (0.0, 20.0)
+            Plot[2]:Plot object containing:
+            [0]: cartesian line: -x**3/1600 + 3*x**2/160 - x/8 for x over (0.0, 20.0)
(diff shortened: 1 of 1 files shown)


Psycho-Pirate:
Here is an example:
```
>>> from sympy.physics.continuum_mechanics.beam import Beam3D
            >>> from sympy import symbols
            >>> l, E, G, I, A, x = symbols('l, E, G, I, A, x')
            >>> b = Beam3D(20, 40, 21, 100, 25, x)
            >>> b.apply_load(15, start=0, order=0, dir="z")
            >>> b.apply_load(12*x, start=0, order=0, dir="y")
            >>> b.bc_deflection = [(0, [0, 0, 0]), (20, [0, 0, 0])]
            >>> R1, R2, R3, R4 = symbols('R1, R2, R3, R4')
            >>> b.apply_load(R1, start=0, order=-1, dir="z")
            >>> b.apply_load(R2, start=20, order=-1, dir="z")
            >>> b.apply_load(R3, start=0, order=-1, dir="y")
            >>> b.apply_load(R4, start=20, order=-1, dir="y")
            >>> b.solve_for_reaction_loads(R1, R2, R3, R4)
            >>> b.solve_slope_deflection()
            >>> b.plot_loading_results('y')
```
all

Psycho-Pirate:
@ishanaj @Sc0rpi0n101

Psycho-Pirate:
I tried to build the documentation but I am getting a weird order of functions in the final docs.
docss
Here ```plot_loading_results()``` is shown before ```plot_shear_force()``` even though I wrote them in the correct order in beam.py

ishanaj:
> Here plot_loading_results() is shown before plot_shear_force() even though I wrote them in the correct order in beam.py

I think it comes up in alphabetical order. No issues!

ishanaj:
Could you try a 4x3 grid for the `dir="all"` case and share the results here?

sympy/physics/continuum_mechanics/beam.py
```diff
@@ -2667,3 +2667,104 @@ def plot_deflection(self, dir="all", subs=None):
...
+            Plot[1]:Plot object containing:
+            [0]: cartesian line: -15*x**2/2 for x over (0.0, 20.0)
+            Plot[2]:Plot object containing:
+            [0]: cartesian line: -x**3/1600 + 3*x**2/160 - x/8 for x over (0.0, 20.0)
+            Plot[3]:Plot object containing:
+            [0]: cartesian line: x**5/40000 - 4013*x**3/90300 + 26*x**2/43 + 1520*x/903 for x over (0.0, 20.0)
+
+        """
+
+        if dir == None:
```
ishanaj, line 2727: Should rather be `if not dir:`

sympy/physics/continuum_mechanics/beam.py
```diff
@@ -2667,3 +2667,104 @@ def plot_deflection(self, dir="all", subs=None):
...
+        elif dir == 'y':
+            dir_num = 1
+
+        elif dir == 'z':
+            dir_num = 2
+
+        if subs is None:
+            subs = {}
+
+        for sym in deflection[dir_num].atoms(Symbol):
```
ishanaj, line 2745: I think this should be done for slope, shear force, and bending moment also. Better to directly use the helper functions of the individual plotting functions (`_plot_slope() ..etc`)

Psycho-Pirate: I did it for deflection only because deflection covers all the variables needed for calculating slope, bending moment, and shear force. If deflection is determinable, we can definitely determine other equations as the code works in order.

I didn't use the helper functions because I wanted to impart different colors to different functions. Using helper functions will plot all the x graphs as red, y as blue, etc unless we modify the helper functions to take color as a parameter.

Psycho-Pirate:
> Could you try a 4x3 grid for the `dir="all"` case and share the results here?

If I do not specify a custom size, it looks like this
loadingres2

But if I define a size of 20x10 for each subplot, it looks like this
loadingres
I am not sure how it will look in the documentation with the custom size, I think it will be pretty small for the user to get any valuable information.

sympy/physics/continuum_mechanics/beam.py
ishanaj: I think the idea is to have different line colors within a single plot. So you can change the color of the helper functions in a way that when you use them here all the line colors in this plot are different. (I won't recommend using color as a parameter)

ishanaj: > ....because deflection covers all the variables needed for calculating slope, bending moment, and shear force

IMO there can be a use case where this won't hold true. Or is this something proven?

ishanaj:
Also, the default `dir` of `plot_loading_results()` can be made `x`, possibly because the same function in 2d doesn't take any arguments, so I think this should also work without any arguments(not considering `subs` here).

sympy/physics/continuum_mechanics/beam.py
Psycho-Pirate: > IMO there can be a use case where this won't hold true. Or is this something proven?

The functions are calculated by integrating the lower functions starting from the load. Shear Force is calculated by integrating load, Bending Moment is calculated by integrating shear force, and so on. Deflection is the function at the top so it covers everything below it and adds on some additional variables. Hence it contains all the variables needed for all the other functions. If Deflection can be calculated, it can be proven that other functions have already been calculated and the variables they contain are already present. 

Even in beam2d class, only Deflection is used for ```plot_loading_results()``` because of the same reason.

Psycho-Pirate:
In the latest commit  I have made the following changes:
* changed colors for every plot
* set default direction to x in ```plot_loading_results()```
* used helper functions in ```plot_loading_results()```
* removed the ```if sym == self.variable: continue``` condition
* used subs in the example

@ishanaj @Sc0rpi0n101

ishanaj:
LGTM

ishanaj (approved):

PR #21650, opened 2021-06-22
Fix MultivariateNormal density calculation when symbolic parameters are used

### LINKED ISSUE #21635: MultivariateNormal density calculation fails when matrix params are used
The following code fails:

```python
import sympy as sy
import sympy.stats

S = sy.MatrixSymbol('S', 3, 3)
m = sy.MatrixSymbol('m', 3, 1)

dist = sy.stats.MultivariateNormal('X', m, S)
x = sy.MatrixSymbol('x', 3, 1)

sy.stats.density(dist)(x)
```
with error:
```
~/miniconda3/lib/python3.7/site-packages/sympy/stats/joint_rv_types.py in pdf(self, *args)
    151         return  S.One/sqrt((2*pi)**(k)*det(sigma))*exp(
    152             Rational(-1, 2)*x.transpose()*(sigma.inv()*\
--> 153                 x))[0]
    154 
    155     def _marginal_distribution(self, indices, sym):

TypeError: 'exp' object is not subscriptable
```

Is this intended behaviour?

InfProbSciX:
Fixes #21635

The `pdf` method for the `MultivariateNormalDistribution` has been updated:
 * to use `MatrixElement` to return the `[0, 0]` value of the density calculation, that returns a `[1, 1]` matrix. This has been changed from accessing the element directly using `[0]`, as this throws an error when certain arguments are symbolic.
 * to check if the input is a single matrix, if not, _*then*_ coerce the inputs to an `ImmutableMatrix`. I've added this because passing a matrix with a symbolic number of elements causes the argument to be coerced to a `1x1` matrix, with the original matrix at the `[0, 0]` index.

#### Other comments

While running tests, I noticed that the `test_sample_pymc3` test breaks (I have pymc3 installed). This is also true for the code on master and is unrelated to this PR.

#### Release Notes

* stats
  * Fixed a bug regarding symbolic arguments to the multivariate normal distribution.

### DIFF
--- a/sympy/stats/joint_rv_types.py
+++ b/sympy/stats/joint_rv_types.py
@@ -204,7 +204,7 @@ def MultivariateNormal(name, mu, sigma):
     >>> marginal_distribution(X, X[0])(y)
     exp(-(y - 3)**2/4)/(2*sqrt(pi))
 
-    The example below shows it is also possible to use
+    The example below shows that it is also possible to use
     symbolic parameters to define the MultivariateNormal class.
 
     >>> n = symbols('n', natural=True)
@@ -213,7 +213,7 @@ def MultivariateNormal(name, mu, sigma):
     >>> obs = MatrixSymbol('obs', n, 1)
     >>> X = MultivariateNormal('X', mu, Sg)
 
-    The density of a multivariate normal can also be
+    The density of a multivariate normal can be
     calculated using a matrix argument, as shown below.
 
     >>> density(X)(obs)
--- a/sympy/stats/joint_rv_types.py
+++ b/sympy/stats/joint_rv_types.py
@@ -203,12 +203,19 @@ def MultivariateNormal(name, mu, sigma):
     exp(-(y - 4)**2/4)/(2*sqrt(pi))
     >>> marginal_distribution(X, X[0])(y)
     exp(-(y - 3)**2/4)/(2*sqrt(pi))
-    >>> # symbolic parameters
+
+    The example below shows it is also possible to use
+    symbolic parameters to define the MultivariateNormal class.
+
     >>> n = symbols('n', natural=True)
     >>> Sg = MatrixSymbol('Sg', n, n)
     >>> mu = MatrixSymbol('mu', n, 1)
     >>> obs = MatrixSymbol('obs', n, 1)
     >>> X = MultivariateNormal('X', mu, Sg)
+
+    The density of a multivariate normal can also be
+    calculated using a matrix argument, as shown below.
+
     >>> density(X)(obs)
     (exp(((1/2)*mu.T - (1/2)*obs.T)*Sg**(-1)*(-mu + obs))/sqrt((2*pi)**n*Determinant(Sg)))[0, 0]
 
--- a/sympy/stats/joint_rv_types.py
+++ b/sympy/stats/joint_rv_types.py
@@ -192,7 +192,7 @@ def MultivariateNormal(name, mu, sigma):
     ========
 
     >>> from sympy.stats import MultivariateNormal, density, marginal_distribution
-    >>> from sympy import symbols
+    >>> from sympy import symbols, MatrixSymbol
     >>> X = MultivariateNormal('X', [3, 4], [[2, 1], [1, 2]])
     >>> y, z = symbols('y z')
     >>> density(X)(y, z)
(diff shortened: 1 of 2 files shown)


sympy/stats/tests/test_joint_rv.py
```diff
@@ -46,6 +46,31 @@ def test_Normal():
...
+    assert eval_b == sqrt(2)/(4*pi**Rational(3/2))
+
+    n = symbols('n', natural=True)
+
+    Sg = MatrixSymbol('Sg', n, n)
+    mu = MatrixSymbol('mu', n, 1)
+    obs = MatrixSymbol('obs', n, 1)
+
+    X = MultivariateNormal('X', mu, Sg)
+    density_X_at_obs = density(X)(obs)
```
oscarbenjamin, line 71: It would be better to test what this is equal to like:
```
assert density_X_at_obs == ...
```

InfProbSciX: I'm not sure how to test this. I tried adding:
```python
    expected_density = MatrixElement(
            exp(Rational(-1/2) * (obs - mu).T * Sg**-1 * (obs - mu)) *\
            det(Sg)**(-1/2) * (2*pi)**(-n/2),
        0, 0)
```
and even though the formulas are the same, `expected_density==density_X_at_obs` comes out as `False`. Trying `.subs`, `.doit`, `.simplify` all fail for `density_X_at_obs` giving one error or another... Any ideas?

oscarbenjamin: I've seen this come up a few times. I think it's something to do with the `MatrixSymbol` name being a `Symbol` rather than a `str` or perhaps the same for the `RandomVariable`.

oscarbenjamin: This is what is needed:
```
MatrixElement((exp(((S(1)/2)*mu.T - (S(1)/2)*obs.T)*Sg**(-1)*(-mu + obs))/sqrt((2*pi)**n*Determinant(Sg))), 0, 0)
```

InfProbSciX: Amazing, thanks. Resolved.

oscarbenjamin:
@czgdp1807 @Upabjojr can you review this?

czgdp1807:
If possible (like if it isn't too tricky to do) add an example to the docs as well for this case. Thanks.

InfProbSciX:
@czgdp1807 All done.

InfProbSciX:
@oscarbenjamin @czgdp1807 @Abdullahjavednesar Hello everyone - just following up to see if anything more needs to be done.

sympy/stats/tests/test_joint_rv.py
```diff
@@ -46,6 +48,38 @@ def test_Normal():
...
+    Sg = MatrixSymbol('Sg', n, n)
+    mu = MatrixSymbol('mu', n, 1)
+    obs = MatrixSymbol('obs', n, 1)
+
+    X = MultivariateNormal('X', mu, Sg)
+    density_X = density(X)
+
+    eval_a = density_X(obs).subs({Sg: eye(3),
+        mu: Matrix([0, 0, 0]), obs: Matrix([0, 0, 0])}).doit()
+    eval_b = density_X(0, 0, 0).subs({Sg: eye(3), mu: Matrix([0, 0, 0])}).doit()
```
oscarbenjamin, line 63: I would have expected that the `density_X(0, 0, 0)` would need to be `density_X(Matrix([0, 0, 0]))` or something. Does this convert the three separate positional arguments into a matrix somehow?

oscarbenjamin: I could imagine `density_X([0, 0, 0])` being accepted but it seems odd to allow multiple positional arguments in place of a single positional argument.

InfProbSciX: This is the way the base code has been written - I haven't changed it. I also agree, having one positional argument is probably better, but that might break downstream code...

InfProbSciX: So, this [line in master](https://github.com/sympy/sympy/blob/master/sympy/stats/joint_rv_types.py#L146) takes in `*args` as the input, and forces a conversion of the list `args` to an `ImmutableMatrix`.

oscarbenjamin: Oh, I see. The changes here make it possible to pass a single matrix in instead of multiple args.

InfProbSciX: Yep

oscarbenjamin: Are there examples using multiple args like this in the docs? If so maybe they could be updated to pass a single matrix as you've done here.

sympy/stats/joint_rv_types.py
```diff
@@ -199,6 +203,14 @@ def MultivariateNormal(name, mu, sigma):
     exp(-(y - 4)**2/4)/(2*sqrt(pi))
     >>> marginal_distribution(X, X[0])(y)
     exp(-(y - 3)**2/4)/(2*sqrt(pi))
+    >>> # symbolic parameters
```
oscarbenjamin, line 206: Rather than having this as a comment, break the docstring section and just have some plain text that explains the example below.

sympy/stats/tests/test_joint_rv.py
InfProbSciX: I have added an example in the docs about passing a matrix as an input.

sympy/stats/joint_rv_types.py
InfProbSciX: Sorry - how do you do that?

InfProbSciX: Is it something like:
```
Examples
============

>>> ...

plain text

>>> ...
```

oscarbenjamin: > Is it something like:

Yes. When rendered in the html docs it show up as separate doctest blocks with some text in between.

InfProbSciX: All done.

oscarbenjamin:
Looks good, thanks!

PR #19390, opened 2020-05-21
[GSoC] Add TransferFunction, Series, Parallel and Feedback classes for control package

namannimmo10:
#### References to other Issues or PRs

Built upon #18436.

#### Brief description of what is fixed or changed
This PR adds a class for representing LTI systems in transfer function form.

#### Other comments
**TODO**:

- [x] adding `TransferFunction` class along with its functionality.

- [x] Rewrite unit-tests and make sure they pass.

- [x] Adding documentation.

#### Release Notes

*  physics.control
    *  Add `TransferFunction`, `Series`, `Parallel`, and `Feedback` class for physics.control submodule

PR #21653, opened 2021-06-23
[GSoC] Add `TransferFunctionMatrix` class

akshanshbhatt:

#21556 

Support for Multiple-Input Multiple-Output (MIMO) Control Systems will be added along with documentation and examples of use cases.

#### Other comments

Chunking down PR #21556 into multiple smaller PRs for reviewing purposes.

#### Release Notes

* physics.control
  * Added `TransferFunctionMatrix` class, enabling support for MIMO systems.


akshanshbhatt:
@namannimmo10 please review.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1424,326 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+    It is a matrix of transfer functions (``TransferFunction`` objects).
+    There is only one argument, ``arg`` which is also the compulsory argument.
+    ``arg`` is expected to be strictly of the type list of lists
+    which holds the transfer functions.
+
+    Parameters
+    ==========
+
+    arg: Nested ``List`` (strictly).
+        Users are expected to input a nested list of ``TransferFunction`` objects.
```
oscarbenjamin, line 1453: I would have thought that a Matrix was a more natural input

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1424,326 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+
+        if isinstance(arg, (list, Tuple)):
+            # Making nested Tuple (sympy.core.containers.Tuple) from nested list or nested python tuple
+            arg = Tuple(*(Tuple(*r, sympify=False) for r in arg), sympify=False)
+
+        obj = super(TransferFunctionMatrix, cls).__new__(cls, arg)
+        obj._expr_mat = ImmutableMatrix(expr_mat_arg)
+        obj._var = arg[0][0].var
+        obj._num_outputs, obj._num_inputs = obj._expr_mat.shape
+        return obj
```
oscarbenjamin, line 1623: You should avoid storing additional objects on the instance if possible. Ideally all of these would be derived through property methods from the args. Does this class satisfy `obj == obj.func(*obj.args)`?

akshanshbhatt: Yes. It does satisfy the `obj.func(*obj.args)` condition. In the old PR by @namannimmo10 it was failing and had a `@XFAIL` decorator, which I've removed after it was passing. You can check the changes in test_args.py file.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1424,326 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+    """
+    def __new__(cls, arg):
+
+        expr_mat_arg = []
+        var = arg[0][0].var
+        for row in range(len(arg)):
+            temp = []
+            for col in range(len(arg[row])):
+                if not isinstance(arg[row][col], (TransferFunction, Series, Parallel)):
+                    raise TypeError("Incompatible type found as the element of TransferFunctionMatrix")
```
oscarbenjamin, line 1602: It would seem more extensible to have a common superclass of `TransferFunction, Series, Parallel` that could be tested here.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1424,326 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+            return _to_TFM(trunc, self.var)
+        else:
+            return TransferFunction.from_rational_expression(trunc, self.var)
+
+    def transpose(self):
+        transposed_mat = self._expr_mat.transpose()
+        return _to_TFM(transposed_mat, self.var)
+
+    def subs(self, dict):
+        return self._expr_mat.subs(dict)
```
oscarbenjamin, line 1749: Basic subclasses should not override subs and should not define it to mean that it returns a different kind of object.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1424,326 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+
+    >>> tfm_4 = TransferFunctionMatrix([[tf_2], [-tf_1], [tf_3]])
+    >>> -tfm_4
+    TransferFunctionMatrix(((TransferFunction(-p**4 + 3*p - 2, p + s, s),), (TransferFunction(a + s, s**2 + s + 1, s),), (TransferFunction(-3, s + 2, s),)))
+    >>> tfm_5 = TransferFunctionMatrix([[tf_1, tf_2], [tf_3, -tf_1]])
+    >>> -tfm_5
+    TransferFunctionMatrix(((TransferFunction(-a - s, s**2 + s + 1, s), TransferFunction(-p**4 + 3*p - 2, p + s, s)), (TransferFunction(-3, s + 2, s), TransferFunction(a + s, s**2 + s + 1, s))))
+
+    ``subs()`` returns an ImmutableDenseMatrix object with the value substituted in the expression. This will not
+    mutate your original ``TransferFunctionMatrix``.
```
namannimmo10, line 1580: Based on the constructor down below, I suppose `TransferFunctionMatrix.subs(some_variable, value)` should just work. Var substitution in a transfer function returns a transfer function; TFM must follow the same principle, no?

We don't yet allow the substitution of `var` in `TransferFunction`;
```python
In [19]: G = TransferFunction(x - s, x**2 + 1, s)                      

In [20]: G.subs(x, 2)                                                  
Out[20]: TransferFunction(2 - s, 5, s)

In [21]: G.subs({s: 2, x: 2})                                          
Out[21]: TransferFunction(2 - s, 5, s)
```
But your example here (for TFM) shows otherwise.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1424,326 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+            arg = Tuple(*(Tuple(*r, sympify=False) for r in arg), sympify=False)
+
+        obj = super(TransferFunctionMatrix, cls).__new__(cls, arg)
+        obj._expr_mat = ImmutableMatrix(expr_mat_arg)
+        obj._var = arg[0][0].var
+        obj._num_outputs, obj._num_inputs = obj._expr_mat.shape
+        return obj
+
+    @property
+    def var(self):
```
namannimmo10, line 1626: Update the docstring here.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1424,326 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+        >>> G4 = TransferFunction(s + 1, s**2 + s + 1, s)
+        >>> S1 = Series(G1, G2)
+        >>> S2 = Series(-G3, Parallel(G2, -G1))
+        >>> tfm1 = TransferFunctionMatrix([[G1], [G2], [G3]])
+        >>> tfm1.var
+        p
+        >>> tfm2 = TransferFunctionMatrix([[-S1, -S2], [S1, S2]])
+        >>> tfm2.var
+        p
+        >>> tfm3 = TransferFunctionMatrix([[G4]])
```
namannimmo10, line 1652: Extra square brackets look redundant. Do we really have to restrict `arg` to be nested list? Can this work like Matrix?
```
In [28]: Matrix([TransferFunction(s**2, s - 1, s).to_expr()])          
Out[28]: Matrix([[s**2/(s - 1)]])
```

sympy/printing/pretty/tests/test_pretty.py
```diff
@@ -2551,6 +2552,72 @@ def test_pretty_Feedback():
...
+⎢                                   ⎥\n\
+⎢   2                               ⎥\n\
+⎢- y  + 2⋅y - 1   x - 1      -2     ⎥\n\
+⎢──────────────   ─────      ───    ⎥\n\
+⎣    y + 5        x - y       2     ⎦\
+"""
+    assert upretty(TransferFunctionMatrix([[tf1], [tf2]])) == expected1
+    assert upretty(TransferFunctionMatrix([[tf1], [tf2], [-tf3]])) == expected2
+    assert upretty(TransferFunctionMatrix([[tf1, tf2], [tf3, tf4], [tf5, tf6]])) == expected3
+    assert upretty(TransferFunctionMatrix([[tf2, tf1, tf4], [-tf3, -tf5, -tf6]])) == expected4
```
namannimmo10, line 2618: More tests can be added here.

namannimmo10 (commented):
Looks like a good start! Do you plan to modify the other classes in a separate PR?

sympy/physics/control/lti.py
akshanshbhatt: I don't get it. Creating an instance of one matrix-type class by inputting another matrix-type object seems a bit odd to me. And since non-expr objects are getting deprecated from matrix class, how can I possibly have Matrix of `TransferFunction`s as a valid input?

sympy/physics/control/lti.py
akshanshbhatt: I will remove the existing `subs` method then and update the example.

sympy/physics/control/lti.py
akshanshbhatt: I first planned to do it like this but that would have required another conditional expression in the constructor. Since I wanted to object creation to be as fast and efficient as possible, I thought that would be better to put it as a compulsory condition for the user.

This was there in my implementation in PR #21556. 
https://github.com/sympy/sympy/blob/f6bd2f18eff977ea094a6b9606f352ac1224337a/sympy/physics/control/lti.py#L1981-L1983

I also thought that it would be better for a beginner to have a simple API (PEP 20). If we allow normal lists then we would have to further explain that it is supposed to be a column matrix and not row. For new users who are not accustomed to sympy matrices might find it difficult to adapt. 

I would do whatever your suggest but allowing simple lists as `arg` might be a performance tradeoff.

sympy/printing/pretty/tests/test_pretty.py
akshanshbhatt: I will add more tests after reviewing process of the code is complete. That's why I have also not added any tests of other functions of `TransferFunctionMatrix` (like `subs`, `transpose` etc.). I will add them once the API changes are finalized.

akshanshbhatt:
> Looks like a good start! Do you plan to modify the other classes in a separate PR?

Yes. In my opinion, that would be better. In smaller PRs, it would be easier to keep track of vulnerabilities and would also easier to review.

sympy/physics/control/lti.py
oscarbenjamin: If I can create a `TransferFunction` as `TransferFunction(expr, s)` then why not `TransferFunctionMatrix` as `TransferFunctionMatrix(M, s)` where `M` is a matrix of expressions involving `s`?

That seems to be how it works in Mathematica:
https://reference.wolfram.com/language/ref/TransferFunctionModel.html

sympy/printing/pretty/tests/test_pretty.py
oscarbenjamin: More tests are not needed. There are already too many tests. The function being tested here is a trivial forwarding method so a single test is fine.

sympy/printing/tests/test_latex.py
```diff
@@ -2287,6 +2288,16 @@ def test_Parallel_printing():
...
 
 
+def test_TransferFunctionMatrix_printing():
+    tf1 = TransferFunction(p, p + x, p)
+    tf2 = TransferFunction(-s + p, p + s, p)
+    tf3 = TransferFunction(p, y**2 + 2*y + 3, p)
+    assert latex(TransferFunctionMatrix([[tf1], [tf2]])) == \
+        '\\left[\\begin{matrix}\\frac{p}{p + x}\\\\\\frac{p - s}{p + s}\\end{matrix}\\right]'
+    assert latex(TransferFunctionMatrix([[tf1, tf2], [tf3, -tf1]])) == \
+        '\\left[\\begin{matrix}\\frac{p}{p + x} & \\frac{p - s}{p + s}\\\\\\frac{p}{y^{2} + 2 y + 3} & \\frac{\\left(-1\\right) p}{p + x}\\end{matrix}\\right]'
```
oscarbenjamin, line 2298: Use raw strings

sympy/printing/pretty/tests/test_pretty.py
namannimmo10: Maybe update `expected4` with the test in which `Series` or `Parallel` instances are passed into the nested list.

sympy/physics/control/lti.py
```diff
@@ -5,8 +5,10 @@
 from sympy.core.sympify import sympify, _sympify
 from sympy.polys import Poly, rootof
 from sympy.series import limit
+from sympy.core.containers import Tuple
```
namannimmo10, line 8: Alphabetical order in absolute imports.

sympy/physics/control/lti.py
akshanshbhatt: > If I can create a `TransferFunction` as `TransferFunction(expr, s)` then why not `TransferFunctionMatrix` as `TransferFunctionMatrix(M, s)` where `M` is a matrix of expressions involving `s`?
> 
> That seems to be how it works in Mathematica:
> https://reference.wolfram.com/language/ref/TransferFunctionModel.html

Oh, you meant a classmethod. I get your point now. It seems to be a good idea. @namannimmo10, what do you say? Should I make a classmethod for creating `TransferFunctionMatrix` objects this way?

namannimmo10: Now that we have an option to create a new transfer function from a rational expression, a similar thing can be done for TFM.

sympy/physics/control/lti.py
namannimmo10: > We don't yet allow the substitution of `var` in `TransferFunction`;

Although, I think we might need this while evaluating a transfer function over a range of frequencies.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1424,326 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+        p
+        >>> tfm2 = TransferFunctionMatrix([[-S1, -S2], [S1, S2]])
+        >>> tfm2.var
+        p
+        >>> tfm3 = TransferFunctionMatrix([[G4]])
+        >>> tfm3.var
+        s
+
+        """
+        return self._var
```
oscarbenjamin, line 1657: Basic instances should extract properties from their args and should not use other internal variables e.g.:
```
class My(Basic):
    def __new__(cls, var):
        obj = super().__new__(var)
    @property
    def var(self):
        return self.args[0]
```
There can be exceptions to this but the classes in this module don't have any reason to be exceptional.

sympy/physics/control/lti.py
oscarbenjamin: > We don't yet allow the substitution of `var` in `TransferFunction`;

Why doesn't that work?

namannimmo10: See https://github.com/sympy/sympy/pull/19390#issuecomment-640556483.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1424,359 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+                    raise TypeError("Incompatible type found as the element of TransferFunctionMatrix")
+
+                if var != arg[row][col].var:
+                    raise ValueError("Conflicting value(s) found for `var`. All TransferFunction instances in "
+                            "TransferFunctionMatrix should use the same complex variable in Laplace domain.")
+
+                if isinstance(arg[row][col], TransferFunction):
+                    temp.append(arg[row][col].to_expr())
+                else:
+                    temp.append(arg[row][col].doit().to_expr())
```
namannimmo10, line 1657: Why are we calling `doit()` here? In my previous PR, I defined a `doit()` method in TFM class which:
```
        Returns the resultant transfer function matrix obtained after evaluating
        the transfer functions in series or parallel configurations (if any present)
        for all entries in a transfer function matrix.
```
Now it does simplify things, but is this the right way of "doing" it?

sympy/physics/control/lti.py
namannimmo10: OK, let's go with this.

sympy/physics/control/lti.py
akshanshbhatt: > Why are we calling `doit()` here? In my previous PR, I defined a `doit()` method in TFM class which:
> 
> ```
>         Returns the resultant transfer function matrix obtained after evaluating
>         the transfer functions in series or parallel configurations (if any present)
>         for all entries in a transfer function matrix.
> ```
> 
> Now it does simplify things, but is this the right way of "doing" it?

I am calling `doit` here for the formation of the `obj._expr_mat`. The actual state of the `Series` or `Parallel` element passed is not changed. 

https://github.com/sympy/sympy/blob/7a6c86302e7fc02cdcfabfee733dc17c5d34cdbc/sympy/physics/control/lti.py#L1661-L1663

akshanshbhatt: I got your point. I have removed `doit()`.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1445,426 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+        if isinstance(arg, (list, Tuple)):
+            # Making nested Tuple (sympy.core.containers.Tuple) from nested list or nested python tuple
+            arg = Tuple(*(Tuple(*r, sympify=False) for r in arg), sympify=False)
+
+        obj = super(TransferFunctionMatrix, cls).__new__(cls, arg)
+        obj._expr_mat = ImmutableMatrix(expr_mat_arg)
+        return obj
+
+    @classmethod
+    def from_matrix(cls, matrix, var):
```
namannimmo10, line 1691:
```suggestion
    def from_Matrix(cls, matrix, var):
```

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1445,426 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+        ========
+
+        >>> from sympy.abc import s, p, a
+        >>> from sympy.physics.control.lti import TransferFunctionMatrix
+        >>> from sympy.matrices import ImmutableMatrix
+        >>> from sympy.printing import pprint
+        >>> expr1 = (s + 5)/(3*s**2 + 2*s + 1)
+        >>> expr2 = 1/s
+        >>> expr3 = 1/s**2
+        >>> expr4 = 1/(s + 1)
```
namannimmo10, line 1714:
```suggestion
        >>> expr_1 = (s + 5)/(3*s**2 + 2*s + 1)
        >>> expr_2 = 1/s
        >>> expr_3 = 1/s**2
        >>> expr_4 = 1/(s + 1)
```

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1445,426 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+
+class TransferFunctionMatrix(Basic):
+    r"""
+    A class for representing the MIMO (multiple-input and multiple-output)
+    generalization of the SISO (single-input and single-output) transfer function.
+
+    It is a matrix of transfer functions (``TransferFunction`` objects).
+    There is only one argument, ``arg`` which is also the compulsory argument.
+    ``arg`` is expected to be strictly of the type list of lists
+    which holds the transfer functions.
```
namannimmo10, line 1468: and Series/Parallel configurations too.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1445,426 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+        Creates a new ``TransferFunctionMatrix`` efficiently from a SymPy Matrix of ``Expr`` objects.
+
+        Parameters
+        ==========
+
+        matrix : ``ImmutableMatrix`` having ``Expr``/``Number`` elements.
+        var : Symbol
+            Complex variable of the Laplace transform which will be used by the
+            all the ``TransferFunction`` objects in the ``TransferFunctionMatrix``
+            object.
```
namannimmo10, line 1702:
```suggestion
            A complex variable of the Laplace transform which will be used by
            all the ``TransferFunction`` objects in ``TransferFunctionMatrix``.
```
Avoid writing _object(s)_ everywhere.

sympy/physics/control/tests/test_lti.py
```diff
@@ -1,9 +1,16 @@
-from sympy import symbols, Matrix, factor, Function, simplify, exp, oo, I, \
-    Rational, sqrt, CRootOf, S, Mul, Pow, Add
-from sympy.physics.control.lti import TransferFunction, Series, Parallel, Feedback
+from sympy import symbols, factor, Function, simplify, exp, oo, I, \
+    S, Mul, Pow, Add, Rational, sqrt, CRootOf
+from sympy.matrices import ImmutableMatrix, Matrix
+from sympy.physics.control import TransferFunction, Series, Parallel, \
+    Feedback, TransferFunctionMatrix
+from sympy.core.containers import Tuple
```
namannimmo10, line 6:
```suggestion
from sympy import (symbols, factor, Function, simplify, exp, oo, I, 
    S, Mul, Pow, Add, Rational, sqrt, CRootOf)
from sympy.core.containers import Tuple
from sympy.matrices import ImmutableMatrix, Matrix
from sympy.physics.control import (TransferFunction, Series, Parallel,
    Feedback, TransferFunctionMatrix)
```

namannimmo10 (commented):
You can also add property related to IO-coupling here.

namannimmo10:
@akshanshbhatt @moorepants, we discussed in the last meeting, the possibility of also having common denominators in a transfer function matrix. I think that's the reason why Wolfram language provides the option for users to pass in a collection of numerator and denominator polynomials separately; see [`TransferFunctionModel`](https://reference.wolfram.com/language/ref/TransferFunctionModel.html).

image
image

I can also see that Harold (by @ilayn) does the same:
https://github.com/ilayn/harold/blob/2bfa00fca4549313d47386991a27c84c8fc91637/harold/_classes.py#L43

Can we provide this option?

akshanshbhatt:
> @akshanshbhatt @moorepants, we discussed in the last meeting, the possibility of also having common denominators in a transfer function matrix. I think that's the reason why Wolfram language provides the option for users to pass in a collection of numerator and denominator polynomials separately; see [`TransferFunctionModel`](https://reference.wolfram.com/language/ref/TransferFunctionModel.html).
> 
> image
> image
> 
> I can also see that Harold (by @ilayn) does the same:
> https://github.com/ilayn/harold/blob/2bfa00fca4549313d47386991a27c84c8fc91637/harold/_classes.py#L43
> 
> Can we provide this option?

We indeed can allow this, but we need to discuss the API once. Should it be a separate classmethod (common denominator passed during object creation itself), or should it be a method in the class to convert all the TransferFunctions into common denominator form?

ilayn:
`is_IO_coupled` is, in my opinion, an overkill basically asking whether it is diagonal or not. I am not sure if it warrants a method in the namespace.

akshanshbhatt:
> `is_IO_coupled` is, in my opinion, an overkill basically asking whether it is diagonal or not. I am not sure if it warrants a method in the namespace.

Initially, I also thought the same, but then I thought that configurations like these-
```
[[ 1/(s+1),  0,   0],
 [ 0,        0,   1],
 [ 0,        1,   0]]
```
are also not IO_coupled.

It is a valid MIMO transfer function, and each input is interacting with one output only. I might be wrong, and in that case, I will change it to check for the diagonal condition.

ilayn:
Certainly true but I have never come across a case where I checked a system whether it is a permuted diagonal. I am not sure if even matlab has something for this. Not that they are very good at relevant commands but at least it is a data point.

sylee957:
I think that `Matrix.connected_components` can reveal all permuted diagonal or triangular structure of a matrix.

akshanshbhatt:
> Certainly true but I have never come across a case where I checked a system whether it is a permuted diagonal. I am not sure if even matlab has something for this. Not that they are very good at relevant commands but at least it is a data point.

So, should I change it to check for diagonal matrix only?

akshanshbhatt:
> I think that `Matrix.connected_components` can already reveal all permuted diagonal or triangular structure of a matrix.

I just saw its [doc ](https://docs.sympy.org/dev/modules/matrices/matrices.html?highlight=connected#sympy.matrices.matrices.MatrixBase.connected_components)and it states that it is only applicable for square matrices. `TransferFunctionMatrix` may or may not be a square matrix.

sylee957:
> > I think that `Matrix.connected_components` can already reveal all permuted diagonal or triangular structure of a matrix.
> 
> I just saw its [doc ](https://docs.sympy.org/dev/modules/matrices/matrices.html?highlight=connected#sympy.matrices.matrices.MatrixBase.connected_components)and it states that it is only applicable for square matrices. `TransferFunctionMatrix` may or may not be a square matrix.

I think that any matrix can be augmented with zeros to make that work though

oscarbenjamin:
What does IO-coupled mean? The docstring says:
```
A MIMO system is IO-coupled when each input affects many outputs.
```
That's not very clear to me but from that I would guess that having two or more nonzero entries in each row (or column?) would make the system IO-coupled. If that's what it means then the code for determining this seems somewhat convoluted.

akshanshbhatt:
I looked for examples of decoupled MIMO systems in books and on the internet. Unfortunately, I could only find diagonal matrices, so I think I should follow the advice from @ilayn and check for the diagonal condition instead of convoluting it without any evidence.

ilayn:
I think this is partially due my fault since this was coming from the gist I saw before and there it was using the term "interactive". Then I mentioned that it is a very nonstandard terminology and I heard IO-coupled or cross-talk for this. 

The original source is probably an ETH slide but again I can safely say it is not used and it seems to me that the author made it up themselves.

akshanshbhatt:
> I think this is partially due my fault since this was coming from the gist I saw before and there it was using the term "interactive". Then I mentioned that it is a very nonstandard terminology and I heard IO-coupled or cross-talk for this.
> 
> The original source is probably an ETH slide but again I can safely say it is not used and it seems to me that the author made it up themselves.

So, should I discard the property?

ilayn:
From controls perspective I never saw this, but if sympy has this for other objects such as the `connected` mentioned above then I can't judge so maybe you would like to keep it as a lazy property.

akshanshbhatt:
> From controls perspective I never saw this, but if sympy has this for other objects such as the `connected` mentioned above then I can't judge so maybe you would like to keep it as a lazy property.

@namannimmo10 what do you say?

namannimmo10:
> The original source is probably an ETH slide but again I can safely say that this metric is not used anywhere else and it seems to me that the author made it up themselves. Because none bothers with this in MIMO systems and switch directly to state-space anyways

In that case, I would personally discard this property. But it's up to you.

namannimmo10:
> > @akshanshbhatt @moorepants, we discussed in the last meeting, the possibility of also having common denominators in a transfer function matrix. I think that's the reason why Wolfram language provides the option for users to pass in a collection of numerator and denominator polynomials separately; see [`TransferFunctionModel`](https://reference.wolfram.com/language/ref/TransferFunctionModel.html).
> > image
> > image
> > I can also see that Harold (by @ilayn) does the same:
> > https://github.com/ilayn/harold/blob/2bfa00fca4549313d47386991a27c84c8fc91637/harold/_classes.py#L43
> > Can we provide this option?
> 
> We indeed can allow this, but we need to discuss the API once. Should it be a separate class method (common denominator passed during object creation itself), or should it be a method in the class to convert all the TransferFunctions into common denominator form?

I think we could have a separate classmethod -- but the standard way of constructing a `TransferFunctionMatrix` would still be from `TransferFunction`s:
```
G1 = TF()
G2 = TF()
TFM = TransferFunctionMatrix([[G1, G2]])
``` 
> should it be a method in the class to convert all the TransferFunctions into common denominator form?

Can you show us one example?

akshanshbhatt:
> Can you show us one example?

In the Mathematica example, you shared with us, say that the user inputs this-

```python
>>> mat_1 = ImmutableMatrix([
...    [3/(2 + s), (6 + s)/((2 + s)*(1 + s))],
...    [(3 + s)/((2 + s)*(1 + s)), 1/(1 + s)]])
>>> tfm_1 = TransferFunctionMatrix.from_Matrix(mat_1, s)
>>> pprint(tfm_1)
[       3              s + 6     ]
[     -----       ---------------]
[     s + 2       (s + 1)*(s + 2)]
[                                ]
[     s + 3              1       ]
[---------------       -----     ]
[(s + 1)*(s + 2)       s + 1     ]
```

After this, if he wants to have a common denominator form, he can-

```python
>>> tfm_denom = tfm_1.common_denom()
[    3*s + 3           s + 6     ]
[---------------  ---------------]
[(s + 1)*(s + 2)  (s + 1)*(s + 2)]
[                                ]
[     s + 3            s + 2     ]
[---------------  ---------------]
[(s + 1)*(s + 2)  (s + 1)*(s + 2)]
```

namannimmo10:
> After this, if he wants to have a common denominator form, he can-
> 
> ```python
> >>> tfm_denom = tfm_1.common_denom()
> >>> pprint(tfm_denom)
> [    3*s + 3           s + 6     ]
> [---------------  ---------------]
> [(s + 1)*(s + 2)  (s + 1)*(s + 2)]
> [                                ]
> [     s + 3            s + 2     ]
> [---------------  ---------------]
> [(s + 1)*(s + 2)  (s + 1)*(s + 2)]
> ```
@akshanshbhatt: this looks better to me.

`poles`, `zeros`, and `doit` methods are also needed. Will you include them in this PR?

akshanshbhatt:
> `poles`, `zeros`, and `doit` methods are also needed. Will you include them in this PR?

I am working on element-wise poles and zeros (The one shown in my example and present in [Mathematica's API](https://wolfram.com/xid/0bh07vf2uxor9g-f7nt6i)). As far as the actual poles and zeros of TFM go, I am planning to implement it later due to its complicated nature and multiple possibilities depending on the matrix.

akshanshbhatt:
> @akshanshbhatt: this looks better to me.

@namannimmo10 , so should I finally implement this or the classmethod we talked earlier?

namannimmo10:
> @namannimmo10 , so should I finally implement this or the classmethod we talked earlier?

IMO, the reason why any other toolbox provides this option of instantiating a TFM with a common denominator is to avoid rewriting it for every Transfer function. So I would first go with this. Common denominators are observed in many scenarios.
But in any case later on, if they'd require a denominator polynomial that is the least common multiple, `common_denom()` can do the job. Maybe add this later on if needed.
@ilayn @moorepants might have different opinions.

akshanshbhatt:
> IMO, the reason why any other toolbox provides this option of instantiating a TFM with a common denominator is to avoid rewriting it for every Transfer function. So I would first go with this.
> @ilayn @moorepants might have different opinions.

In that case, what should I name the classmethod?

ilayn:
Just to mention what I had in mind a few ages ago and why I coded what I coded is that 

> If there is a denominator with same shape of numerator, nice, go ahead, But if there is one and only one entry in the denominator, then, oh they are being lazy, so assume it to be common to all of the numerator entries,

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1445,545 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+            doit_arg.append(temp)
+            temp = []
+        return TransferFunctionMatrix(doit_arg)
+
+    def elem_poles(self):
+        """
+        Returns the poles of each element of the ``TransferFunctionMatrix``.
+
+        .. note::
+            Actual poles of a MIMO system are NOT the poles of individual elements.
```
namannimmo10, line 1929: I don't understand this. If you say the following is true:
>  Actual poles of a MIMO system are NOT the poles of individual elements.

Then why are you doing it this way? I don't think we need two methods to find poles.

namannimmo10:
```diff
$ git diff
diff --git a/sympy/physics/control/lti.py b/sympy/physics/control/lti.py
index a89c72858b..4ff5fb6be5 100644
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -1449,11 +1449,15 @@ def __neg__(self):
 
 def _to_TFM(mat, var):
     """Private method to convert ImmutableMatrix to TransferFunctionMatrix efficiently"""
-    def group(flat, size): return [flat[i:i+size] for i in range(0, len(flat), size)]
+
+    def group(flat, size):
+        return [flat[i:i+size] for i in range(0, len(flat), size)]
+
     tfm_cols = mat.args[1]
     flat_list = list(mat.args[2])
     flat_list = [TransferFunction.from_rational_expression(elem, var) for elem in flat_list]
     arg = group(flat_list, tfm_cols)
+
     return TransferFunctionMatrix(arg)
 
 
@@ -1862,8 +1866,8 @@ def __getitem__(self, key):
         trunc = self._expr_mat.__getitem__(key)
         if isinstance(trunc, ImmutableMatrix):
             return _to_TFM(trunc, self.var)
-        else:
-            return TransferFunction.from_rational_expression(trunc, self.var)
+
+        return TransferFunction.from_rational_expression(trunc, self.var)
 
     def transpose(self):
         transposed_mat = self._expr_mat.transpose()
```

sympy/physics/control/lti.py
```diff
@@ -891,6 +896,15 @@ def __truediv__(self, other):
     def __neg__(self):
         return Series(TransferFunction(-1, 1, self.var), self)
 
+    def _to_expr(self):
+        temp = []
+        for arg in self.args:
+            if isinstance(arg, TransferFunction):
+                temp.append(arg.to_expr())
+            else:
+                temp.append(arg._to_expr())
```
oscarbenjamin, line 905: This block seems to be repeated in a few places. Why don't all classes just have a method of the same name?

akshanshbhatt: I thought that the `to_expr` method for `Series` and `Parallel` objects might not have any application for the end-user. However, in the `TransferFunction` class, @namannimmo10 demonstrated a few use cases (https://github.com/sympy/sympy/pull/21634#discussion_r654944089) where it was genuinely helpful for the user.

sympy/physics/control/lti.py
akshanshbhatt: Firstly, what I meant was - _Actual poles of a MIMO system are NOT **always** the poles of individual elements_. For an uncoupled TFM, the poles of the system are **poles of the individual components**. It might be beneficial for the user to visualize all the poles and zeros of the elements at once. Later, when adding the Poles and Zeros of the system, a beginner user might find it convenient to compare the poles and zeros to element-wise poles and zeros and get an intuitive understanding of the MIMO system stability. We can add an example in the textbook problems section where we could ask the user to find the poles and zeros of the system and compare them with the poles and zeros of the individual elements. In my opinion, these methods have practical use cases, and we should keep them.

namannimmo10:
@akshanshbhatt This PR is super close to being mergeable. How much time do you need to complete the remaining tasks?  
What about a goal to merge this on Wednesday? We'd like to get you into the next steps and some new features. @moorepants, do you have any comments on the present API?

On a side note... You could have a TODO list in the following PRs, it helps.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1445,563 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+        """
+        doit_arg = []
+        for i in self.args[0]:
+            temp = []
+            for j in i:
+                if not isinstance(j, TransferFunction):
+                    j = j.doit()
+                temp.append(j)
+            doit_arg.append(temp)
+        return TransferFunctionMatrix(doit_arg)
```
oscarbenjamin, line 1940: This code is somewhat convoluted:
```
arg = [[element.doit() for element in row] for row in self.args[0]]
return TransferFunctionMatrix(arg)
```
Then again is that even needed?

Surely the `Basic.doit` method already handles this.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1445,563 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+        >>> tfm_1.elem_poles()
+        [[[-1], [-2, -1]], [[-2, -1], [-5/2 + sqrt(65)/2, -sqrt(65)/2 - 5/2]]]
+
+        """
+        poles_list = []
+        for i in self.args[0]:
+            temp = []
+            for j in i:
+                if not isinstance(j, TransferFunction):
+                    j = j.doit()
```
oscarbenjamin, line 1970: Do you need to test `isinstance` here? Why not just call `doit` on the argument regardless?

sympy/physics/control/lti.py
oscarbenjamin: The methods should just have the same name for all classes so you don't have to check `isinstance` like this

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1445,563 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+
+    """
+    def __new__(cls, arg):
+
+        expr_mat_arg = []
+        var = arg[0][0].var
+        for row_index, row in enumerate(arg):
+            temp = []
+            for col_index, element in enumerate(row):
+                if not isinstance(element, (TransferFunction, Series, Parallel)):
```
oscarbenjamin, line 1687: These should have a common superclass that you can check for so you don't need an explicit list of types in a situation like this.

sympy/physics/control/tests/test_lti.py
```diff
@@ -727,3 +734,108 @@ def test_Feedback_functions():
...
+    assert -tfm2 == TransferFunctionMatrix([[-TF1, TF2], [-TF3, -tf5]])
+
+    tfm2_ = TransferFunctionMatrix([[TF1, -TF2], [TF3, tf5]])
+    assert -tfm2_ == TransferFunctionMatrix([[-TF1, TF2], [-TF3, -tf5]])
+
+    tfm3 = TransferFunctionMatrix([[TF1, TF2, TF3], [tf5, -TF1, -TF3]])
+    assert -tfm3 == TransferFunctionMatrix([[-TF1, -TF2, -TF3], [-tf5, TF1, TF3]])
+
+    tfm3_ = TransferFunctionMatrix([[TF1, TF2, TF3], [tf5, -TF1, -TF3]])
+    assert -tfm3_ == TransferFunctionMatrix([[-TF1, -TF2, -TF3], [-tf5, TF1, TF3]])
```
oscarbenjamin, line 841: These tests are very repetitive and seem excessive for testing something as basic as this. If there are no corner cases to handle then one test is fine.

Note that having too many tests is actually a problem - it's a waste of code as well as a waste of time every time the test suite runs. The excessive tests make it harder to see what is meaningfully being tested when looking at the test suite.

sympy/printing/latex.py
```diff
@@ -2482,6 +2482,9 @@ def _print_Feedback(self, expr):
         denom = self._print(den)
         return r"\frac{%s}{%s}" % (numer, denom)
 
+    def _print_TransferFunctionMatrix(self, expr):
+        return self._print(expr._expr_mat)
```
oscarbenjamin, line 2486: Should this not look different from an ordinary matrix in some way? It can be very confusing when different objects print the same but behave differently.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1445,563 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+                if not isinstance(j, TransferFunction):
+                    j = j.doit()
+                temp.append(j.poles())
+            poles_list.append(temp)
+            temp = []
+        return poles_list
+
+    def elem_zeros(self):
+        """
+        Returns the zeros of each element of the ``TransferFunctionMatrix``.
```
oscarbenjamin, line 1978: This could have a `see also` for poles.

sympy/physics/control/tests/test_lti.py
akshanshbhatt: I have removed redundant tests.

sympy/printing/latex.py
akshanshbhatt: It is represented like a typical matrix in every academic material I have read related to it, even in control packages in other modules. If you have any suggestions then maybe we can discuss them.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1445,551 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+
+
+def _to_TFM(mat, var):
+    """Private method to convert ImmutableMatrix to TransferFunctionMatrix efficiently"""
+    def group(flat, size):
+        return [flat[i:i+size] for i in range(0, len(flat), size)]
+    tfm_cols = mat.args[1]
+    flat_list = list(mat.args[2])
+    flat_list = [TransferFunction.from_rational_expression(elem, var) for elem in flat_list]
+    arg = group(flat_list, tfm_cols)
```
oscarbenjamin, line 1457: Don't access `args` directly like this. You should not presume the layout of the args of a class from outside. Instead the class should provide accessors like `mat.cols`.

In any case this code can just be:
```python
to_tf = lambda expr: TransferFunction.from_rational_expression(expr, var)
arg = [[to_tf(expr) for expr in row] for row in mat.tolist()]
```

sympy/physics/control/lti.py
akshanshbhatt: Can you give an example?
I am very clear about what you are trying to say, but what should I name the class?

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1445,551 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+        >>> pprint(tfm_2, use_unicode=False)
+        [ 10        1    ]
+        [ --    ---------]
+        [ 1     s*(s - 1)]
+        [                ]
+        [  p        a    ]
+        [-----   ------- ]
+        [s - 1   a*s + p ]
+
+        """
```
oscarbenjamin, line 1801: Is it really necessary to have so many examples and for them to be so long?

Note that these examples are for the documentation and are to be seen by users. If they do not convey information to users then they shouldn't be there.

I would have thought that this could just be one example:
```python
>>> M = Matrix([[s, 1/s], [1/(s+1), s]])
>>> Mtf = TransferFunctionMatrix(M, s)
>>> pprint(Mtf)
```
As far as I can tell the other examples don't really convey any meaningful information. Since the same as normal matrix printing the matrix is not really informative. Perhaps it would be better to briefly show some property that `Mtf` has that `M` does not like the poles or something.

Also why does this need to be `ImmutableMatrix`. Normal `Matrix` sympifies to `ImmutableMatrix` so if you use `sympify` then a normal `Matrix` could be used here.

Also Mathematica's version of this allows a matrix of numerators and a matrix of denominators. Is that supported here? Perhaps there could be a separate method for that.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1445,551 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+        >>> tf1 = TransferFunction(a*p**2 + b*s, s - p, s)
+        >>> tf2 = TransferFunction(s**3 - 2, s**4 + 5*s + 6, s)
+        >>> tf3 = TransferFunction(a*s - 4, s**4 + 1, s)
+        >>> S1, S2 = Series(tf1, -tf2), Series(tf2, tf3, -tf1)
+        >>> TFM1 = TransferFunctionMatrix([[tf1], [tf2], [tf3]])
+        >>> TFM1.num_outputs
+        3
+        >>> TFM2 = TransferFunctionMatrix([[S1, S2], [-S2, -S1]])
+        >>> TFM2.num_outputs
+        2
```
oscarbenjamin, line 1876: I would aim for the simplest demonstration possible here:
```
>>> TFM = TransferFunction.from_Matrix([[s], [1/s]], s)
>>> print(TFM)
>>> TFM.num_outputs
2
```
Multiple examples are hardly needed for something this simple. What would be more useful is an explicit statement that this is just the number of rows in the matrix.

This could have a see also for `num_inputs`

sympy/physics/control/tests/test_lti.py
```diff
@@ -727,3 +734,96 @@ def test_Feedback_functions():
...
+    assert tfm5.var == s
+
+    tfm6 = TransferFunctionMatrix([[TF1, TF2, TF3], [tf5, tf6, -tf6]])
+    assert tfm6.shape == (tfm6.num_outputs, tfm6.num_inputs) == (2, 3)
+    assert tfm6.args == Tuple(Tuple(Tuple(TF1, TF2, TF3), Tuple(tf5, tf6, -tf6)))
+    assert tfm6.var == s
+
+    tfm7 = TransferFunctionMatrix([[TF1, TF2], [TF3, -tf5], [-tf5, TF2]])
+    assert tfm7.shape == (tfm7.num_outputs, tfm7.num_inputs) == (3, 2)
+    assert tfm7.args == Tuple(Tuple(Tuple(TF1, TF2), Tuple(TF3, -tf5), Tuple(-tf5, TF2)))
```
oscarbenjamin, line 787: Do we really need more than one of these?

oscarbenjamin: I can imagine that maybe two or three might be needed for full line coverage but we certainly don't need to repeat this 7 times

sympy/physics/control/lti.py
oscarbenjamin: The lazy is to call it something like `TransferFunctionBase`. That's clear to programmers but makes no mathematical sense to users. For users it's better if you have a meaningful name for the kinds of objects that can go into a `TransferFunctionMatrix`.

It would be the superclass of the others and would define the common methods like `to_expr`. The name should match up with whatever these conceptually represent if possible.

sympy/printing/latex.py
oscarbenjamin: If it is a Matrix then we don't need a separate class. Apparently it needs to be different from a matrix so it should probably look a little different. For example in Mathematica it seems to display with a funny `tau` next to it.

sympy/physics/control/tests/test_lti.py
oscarbenjamin: There can be a reason for having multiple cases in order to test the corner cases like `TransferFunctionMatrix([])`.

sympy/printing/latex.py
akshanshbhatt: It makes sense, but what can be the best way to add this `tau` here? 
The `tau` in Mathematica is reminiscent of `TransferFunction`, and similarly, they have curly `S` for State Space representation. A comparable argument should be valid for `TransferFunction` and `Expr`. Should we also consider adding `tau` to `TransferFunction`?

sympy/physics/control/lti.py
akshanshbhatt: >Also Mathematica's version of this allows a matrix of numerators and a matrix of denominators. Is that supported here? Perhaps there could be a separate method for that.

The API has not been finalized yet, so I plan to add it in a separate PR later.

sympy/physics/control/lti.py
akshanshbhatt: @namannimmo10, what do you suggest?

namannimmo10: We did consider introducing a class called `TransferFunctionExpr` last year so that we could subclass `TransferFunction`, `Series`, and `Parallel` from it and make `__add__`, `__mul__` overloaded for `TransferFunctionExpr` with `Series` and `Parallel`. It could somewhat mimic the `MatrixExpr` superclass in the matrices module:
https://github.com/sympy/sympy/blob/200788d6d52657ae36d21737ed780471ec38be27/sympy/matrices/expressions/matexpr.py#L35
Finally, though, didn't find any harm in overloading operators in Series and Parallel classes. 

I'm not sure if we really need to introduce such a class even now, but since @oscarbenjamin is well versed with the matrices module, maybe it _is_ needed to structure the code. Can this be done later on?

oscarbenjamin: I wouldn't try to emulate the design of the matrices module. The reason for wanting a superclass is not so the subclasses can share code but rather:

1. Be able to define in a meaningful way what kinds of objects can be used in what situations
2. Define the interface that is expected of those objects
3. Document that clearly for users and contributors

sympy/printing/latex.py
akshanshbhatt: @namannimmo10 @oscarbenjamin, should I go ahead with adding `tau` as a subscript?

image

After it is implemented, `TransferFunctionMatrix` would render like this.

akshanshbhatt:
@namannimmo10 @oscarbenjamin, can this be merged now?
I have started working on `Series`, `Parallel`, and `Feedback` classes for MIMO systems, so it would be better to merge this PR as soon as possible. Let me know if there are further changes to be made.

sympy/physics/control/tests/test_lti.py
```diff
@@ -727,3 +734,64 @@ def test_Feedback_functions():
...
+    # length of all the lists/tuples in the first arg of TFM should be equal.
+    raises(ValueError, lambda: TransferFunctionMatrix([[TF1], [TF3, tf5]]))
+    raises(ValueError, lambda: TransferFunctionMatrix([[TF1, TF3], [tf5]]))
+
+    # lists/tuples only support transfer functions in them.
+    raises(TypeError, lambda: TransferFunctionMatrix([[TF1, TF2], [TF3, Matrix([1, 2])]]))
+    raises(TypeError, lambda: TransferFunctionMatrix([[TF1, Matrix([1, 2])], [TF3, TF2]]))
+
+
+def test_TransferFunctionMatrix_functions():
```
namannimmo10, line 771: Not yet tested: `doit`, `elem_poles`, `elem_zeros`, `subs`, et al.

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1441,520 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+    [ 2     3  ]
+    [s     s   ]
+
+    See Also
+    ========
+
+    TransferFunction, Series, Parallel, Feedback
+
+    """
+    def __new__(cls, arg):
```
namannimmo10, line 1700:
```python
>>> G = TransferFunction(s, s + 1, s)
>>> TransferFunctionMatrix([G])
  File "/home/namannimmo/oss/sympy/sympy/physics/control/lti.py", line 1703, in __new__
    var = arg[0][0].var
TypeError: 'TransferFunction' object is not subscriptable
```
Raise an error mentioning `arg` can only be a nested list. Add test for that as well.

sympy/physics/control/tests/test_lti.py
```diff
@@ -727,3 +734,64 @@ def test_Feedback_functions():
...
+    raises(ValueError, lambda: TransferFunctionMatrix([[TF1], [TF2], [tf4]]))
+    raises(ValueError, lambda: TransferFunctionMatrix([[TF1, tf4], [TF3, tf5]]))
+
+    # length of all the lists/tuples in the first arg of TFM should be equal.
+    raises(ValueError, lambda: TransferFunctionMatrix([[TF1], [TF3, tf5]]))
+    raises(ValueError, lambda: TransferFunctionMatrix([[TF1, TF3], [tf5]]))
+
+    # lists/tuples only support transfer functions in them.
+    raises(TypeError, lambda: TransferFunctionMatrix([[TF1, TF2], [TF3, Matrix([1, 2])]]))
+    raises(TypeError, lambda: TransferFunctionMatrix([[TF1, Matrix([1, 2])], [TF3, TF2]]))
```
namannimmo10, line 768:
```suggestion
    # length of all the lists in TFM should be equal.
    raises(ValueError, lambda: TransferFunctionMatrix([[TF1], [TF3, tf5]]))
    raises(ValueError, lambda: TransferFunctionMatrix([[TF1, TF3], [tf5]]))

    # lists only support transfer functions in them.
    raises(TypeError, lambda: TransferFunctionMatrix([[TF1, TF2], [TF3, Matrix([1, 2])]]))
    raises(TypeError, lambda: TransferFunctionMatrix([[TF1, Matrix([1, 2])], [TF3, TF2]]))
```

namannimmo10:
```diff
$ git diff
diff --git a/sympy/physics/control/lti.py b/sympy/physics/control/lti.py
index 223ba89da0..c44694f7b5 100644
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -1909,16 +1909,7 @@ def elem_poles(self):
         elem_zeros
 
         """
-        poles_list = []
-        for i in self.args[0]:
-            temp = []
-            for j in i:
-                if not isinstance(j, TransferFunction):
-                    j = j.doit()
-                temp.append(j.poles())
-            poles_list.append(temp)
-            temp = []
-        return poles_list
+        return [[element.poles() for element in row] for row in self.doit().args[0]]
 
     def elem_zeros(self):
         """
@@ -1948,13 +1939,4 @@ def elem_zeros(self):
         elem_poles
 
         """
-        zeros_list = []
-        for i in self.args[0]:
-            temp = []
-            for j in i:
-                if not isinstance(j, TransferFunction):
-                    j = j.doit()
-                temp.append(j.zeros())
-            zeros_list.append(temp)
-            temp = []
-        return zeros_list
+        return [[element.zeros() for element in row] for row in self.doit().args[0]]
```

sympy/printing/latex.py
namannimmo10: We can add `tau` as a superscript for it to look different from an ordinary matrix.

akshanshbhatt: > We can add `tau` as a superscript for it to look different from an ordinary matrix.

Superscript might confuse users with exponent. Subscript is more apt, in my opinion.

akshanshbhatt: Also, should I change only the latex method or both latex and pprint?

namannimmo10: Both of them.

sympy/physics/control/lti.py
```diff
@@ -891,6 +896,13 @@ def __truediv__(self, other):
     def __neg__(self):
         return Series(TransferFunction(-1, 1, self.var), self)
 
+    def to_expr(self):
+        """Returns the equivalent ``Expr`` object."""
+        temp = []
+        for arg in self.args:
+            temp.append(arg.to_expr())
+        return Mul(*temp, evaluate=False)
```
oscarbenjamin, line 904: Using `temp` as a variable name is almost always a bad choice. The `temp` list isn't needed anyway:
```
return Mul(*(arg.to_expr() for arg in self.args))
```

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1441,520 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+        zeros_list = []
+        for i in self.args[0]:
+            temp = []
+            for j in i:
+                if not isinstance(j, TransferFunction):
+                    j = j.doit()
+                temp.append(j.zeros())
+            zeros_list.append(temp)
+            temp = []
+        return zeros_list
```
oscarbenjamin, line 1960: This is
```
return [element.doit().zeros() for row in self.args[0] for element in row]
```
It would probably be useful to add a `._flat()` method that returns all elements as a flattened list and then this would just be:
```
return [element.doit().zeros() for element in self._flat()]
```

sympy/printing/latex.py
akshanshbhatt: @namannimmo10, I have updated the latex support for TFM, and it renders as expected. Here is a sample -

image

However, I am not well versed with the `pprint` parser. I tried to implement subscript in `pprint`, and the results were unsatisfactory. If you could help.

sympy/physics/control/lti.py
akshanshbhatt: > This is
> 
> ```
> return [element.doit().zeros() for row in self.args[0] for element in row]
> ```

@oscarbenjamin, actually, the `zeros_list` variable mentioned above is not a flat list of zeros. I have anyways added the `_flat()` method as it would surely be useful in later phases.

sympy/printing/latex.py
namannimmo10: I think in this case you need to define `_print_TransferFunctionMatrix` in _mathml.py_ file. Once you're done with that, one test can be added in _test_mathml.py_.

akshanshbhatt: > I think in this case you need to define `_print_TransferFunctionMatrix` in _mathml.py_ file. Once you're done with that, one test can be added in _test_mathml.py_.

I'll do that, but how will the MathML parser help us in getting the pretty-printed output?

sympy/physics/control/lti.py
```diff
@@ -1422,3 +1435,510 @@ def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
...
+            Complex variable of the Laplace transform which will be used by the
+            all the ``TransferFunction`` objects in the ``TransferFunctionMatrix``.
+
+        Examples
+        ========
+
+        >>> from sympy.abc import s
+        >>> from sympy.physics.control.lti import TransferFunctionMatrix
+        >>> from sympy.matrices import Matrix
+        >>> from sympy.printing import pprint
```
oscarbenjamin, line 1744: Both `Matrix` and `pprint` can be imported as 
```
from sympy import Matrix, pprint
```
In the docs we should recommend that top-level names be imported in this way because we don't want users to import directly from submodules like `sympy.matrices`.

oscarbenjamin: This comment applies to all of the docstrings

namannimmo10:
@akshanshbhatt, scratch my comment earlier. Best I can do right now is ... 
```diff
$ git diff
diff --git a/sympy/printing/pretty/pretty.py b/sympy/printing/pretty/pretty.py
index f48b47d70e..23369f1817 100644
--- a/sympy/printing/pretty/pretty.py
+++ b/sympy/printing/pretty/pretty.py
@@ -1014,7 +1014,10 @@ def _print_Feedback(self, expr):
         return self._print(num)/self._print(den)
 
     def _print_TransferFunctionMatrix(self, expr):
-        return self._print(expr._expr_mat)
+        D = self._print(expr._expr_mat)
+        D.baseline = D.height()//2
+        D = prettyForm(*D.right('\n'*(D.baseline) + 'τ'))
+        return D
 
     def _print_BasisDependent(self, expr):
         from sympy.vector import Vector
```
```python
In [6]: G = TransferFunction(s, 1 - s, s)                               

In [7]: G_2 = TransferFunction(1 - s, s + 1, s)                         

In [8]: pprint(TransferFunctionMatrix([[G, G_2], [-G_2, -G]]))          
⎡  s    1 - s⎤ 
⎢─────  ─────⎥ 
⎢1 - s  s + 1⎥ 
⎢            ⎥ 
⎢s - 1   -s  ⎥ 
⎢─────  ─────⎥ 
⎣s + 1  1 - s⎦τ
```
You might have to modify it further to make sure it prints fine for every case.

akshanshbhatt:
> You might have to modify it further to make sure it prints fine for every case.

That's great!
We just need to find a replacement for the `tau` symbol (τ) because it will not render when `use_unicode` `kwarg` is set to `False`. For regular `pprint`, it will print fine.

namannimmo10:
```diff
$ git diff
diff --git a/sympy/printing/pretty/pretty.py b/sympy/printing/pretty/pretty.py
index f48b47d70e..4b60f21a26 100644
--- a/sympy/printing/pretty/pretty.py
+++ b/sympy/printing/pretty/pretty.py
@@ -1014,7 +1014,12 @@ def _print_Feedback(self, expr):
         return self._print(num)/self._print(den)
 
     def _print_TransferFunctionMatrix(self, expr):
-        return self._print(expr._expr_mat)
+        mat = self._print(expr._expr_mat)
+        mat.baseline = mat.height()//2
+        pos = mat.baseline if mat.baseline % 2 else mat.baseline - 1
+        c = greek_unicode['tau'] if self._use_unicode else 't'
+        mat = prettyForm(*mat.right('\n'*pos + c))
+        return mat
 
     def _print_BasisDependent(self, expr):
         from sympy.vector import Vector
```

akshanshbhatt:
> ```diff
> $ git diff
> diff --git a/sympy/printing/pretty/pretty.py b/sympy/printing/pretty/pretty.py
> index f48b47d70e..4b60f21a26 100644
> --- a/sympy/printing/pretty/pretty.py
> +++ b/sympy/printing/pretty/pretty.py
> @@ -1014,7 +1014,12 @@ def _print_Feedback(self, expr):
>          return self._print(num)/self._print(den)
>  
>      def _print_TransferFunctionMatrix(self, expr):
> -        return self._print(expr._expr_mat)
> +        mat = self._print(expr._expr_mat)
> +        mat.baseline = mat.height()//2
> +        pos = mat.baseline if mat.baseline % 2 else mat.baseline - 1
> +        c = greek_unicode['tau'] if self._use_unicode else 't'
> +        mat = prettyForm(*mat.right('\n'*pos + c))
> +        return mat
>  
>      def _print_BasisDependent(self, expr):
>          from sympy.vector import Vector
> ```

@namannimmo10, there were some inconsistencies in the output of this code (when the height of the Matrix is even). I fixed it with slight modifications (↓) and it works perfectly fine now. 
```diff
$ git diff
diff --git a/sympy/printing/pretty/pretty.py b/sympy/printing/pretty/pretty.py
index f48b47d70e..fe9ad9a5bf 100644
--- a/sympy/printing/pretty/pretty.py
+++ b/sympy/printing/pretty/pretty.py
@@ -1014,7 +1014,17 @@ def _print_Feedback(self, expr):
         return self._print(num)/self._print(den)

     def _print_TransferFunctionMatrix(self, expr):
-        return self._print(expr._expr_mat)
+       mat = self._print(expr._expr_mat)
+       mat.baseline = mat.height() - 1
+       subscript = greek_unicode['tau'] if self._use_unicode else r'{t}'
+       mat = prettyForm(*mat.right(f"{subscript}"))
+       return mat

     def _print_BasisDependent(self, expr):
         from sympy.vector import Vector
```
I just wanted to ask one thing before finally making the changes in all the tests. Should I use `{t}` as the subscript when `use_unicode` is `False` OR just the simple `t`?
You see, `τ` renders fine when `use_unicode` is `True`. It also fits perfectly well. So, there is no need to modify it. However, regular `t` (when `use_unicode` is `False`) as a subscript looks like a glitch or an overflow from the Matrix. That's why I thought it would be better to use `{t}` as the subscript rather than `t`.

```python
>>> M_1 = Matrix([[5, 6*s, 7*s**2], [5, 6*s**3, 7*s]])
>>> T_1 = TransferFunctionMatrix.from_Matrix(M_1, s)
>>> pprint(T_1, use_unicode=True)
⎡            2⎤ 
⎢5  6⋅s   7⋅s ⎥
⎢─  ───   ────⎥
⎢1   1     1  ⎥
⎢             ⎥
⎢      3      ⎥
⎢5  6⋅s   7⋅s ⎥
⎢─  ────  ─── ⎥
⎣1   1     1  ⎦τ
```
```python
>>> pprint(T_1, use_unicode=False) # 1
[            2]   
[5  6*s   7*s ]
[-  ---   ----]
[1   1     1  ]
[             ]
[      3      ]
[5  6*s   7*s ]
[-  ----  --- ]
[1   1     1  ]{t}
```
```python
>>> pprint(T_1, use_unicode=False) # 2
[            2]   
[5  6*s   7*s ]
[-  ---   ----]
[1   1     1  ]
[             ]
[      3      ]
[5  6*s   7*s ]
[-  ----  --- ]
[1   1     1  ]t
```

Should I go ahead with 1st one or the 2nd one? Let me know your thoughts.

namannimmo10:
OK, you can go with the first one

namannimmo10 (approved):
Thanks and congrats! This is in.

namannimmo10:
@akshanshbhatt, perhaps you could also discuss how you would like to add a common superclass in the following PRs.

akshanshbhatt:
> @akshanshbhatt, perhaps you could also discuss how you would like to add a common superclass in the following PRs.

Actually, I have found a better alternative. After `Series` and `Parallel` support is implemented for MIMO systems, that condition wouldn't be true anyways (because `isinstance(elem, Series/Parallel) == True` wouldn't imply that elem is SISO or MIMO Series/Parallel configuration). Therefore like your previous implementation, I plan to bring back `is_SISO` property in all the classes; after which we just need to check for the **one** `is_SISO` condition.

How does this sound to you? 
Let me know if you think there is some flaw in this or I am missing something.

PR #21654, opened 2021-06-23
refactor const homogenous hints

Mohitbalwani26:

SEE #18348

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/solvers/ode/ode.py
+++ b/sympy/solvers/ode/ode.py
@@ -1059,6 +1059,9 @@ def classify_ode(eq, func=None, dict=False, ics=None, *, prep=True, xi=None, eta
         LinearCoefficients: ('linear_coefficients',),
         NthOrderReducible: ('nth_order_reducible',),
         Hypergeometric2nd: ('2nd_hypergeometric',),
+        NthConstCoeffHomogen: ('nth_linear_constant_coeff_homogeneous',),
+        NthConstCoeffVar: ('nth_linear_constant_coeff_variation_of_parameters',),
+        NthConstCoeffUndet: ('nth_linear_constant_coeff_undetermined_coefficients',),
     }
     for solvercls in solvers:
         solver = solvercls(ode)
@@ -1179,25 +1182,6 @@ def classify_ode(eq, func=None, dict=False, ics=None, *, prep=True, xi=None, eta
         # a_n(x)y^(n) + ... + a_1(x)y' + a_0(x)y = F(x) = b
 
         r = _nth_linear_match(reduced_eq, func, order)
-
-        # Constant coefficient case (a_i is constant for all i)
-        if r and not any(r[i].has(x) for i in r if i >= 0):
-            # Inhomogeneous case: F(x) is not identically 0
-            if r[-1]:
-                eq_homogeneous = Add(eq,-r[-1])
-                undetcoeff = _undetermined_coefficients_match(r[-1], x, func, eq_homogeneous)
-                s = "nth_linear_constant_coeff_variation_of_parameters"
-                matching_hints[s] = r
-                matching_hints[s + "_Integral"] = r
-                if undetcoeff['test']:
-                    r['trialset'] = undetcoeff['trialset']
-                    matching_hints[
-                        "nth_linear_constant_coeff_undetermined_coefficients"
-                            ] = r
-
-            # Homogeneous case: F(x) is identically 0
-            else:
-                matching_hints["nth_linear_constant_coeff_homogeneous"] = r
 
         # nth order Euler equation a_n*x**n*y^(n) + ... + a_1*x*y' + a_0*y = F(x)
         #In case of Homogeneous euler equation F(x) = 0
@@ -1845,42 +1829,6 @@ def odesimp(ode, eq, func, hint):
     if eq.lhs == func and not eq.rhs.has(func):
         # The solution is already solved
         eq = [eq]
-
-        # special simplification of the rhs
-        if hint.startswith("nth_linear_constant_coeff"):
-            # Collect terms to make the solution look nice.
-            # This is also necessary for constantsimp to remove unnecessary
-            # terms from the particular solution from variation of parameters
-            #
-            # Collect is not behaving reliably here.  The results for
-            # some linear constant-coefficient equations with repeated
-            # roots do not properly simplify all constants sometimes.
-            # 'collectterms' gives different orders sometimes, and results
-            # differ in collect based on that order.  The
-            # sort-reverse trick fixes things, but may fail in the
-            # future. In addition, collect is splitting exponentials with
-            # rational powers for no reason.  We have to do a match
-            # to fix this using Wilds.
-            #
-            # XXX: This global collectterms hack should be removed.
-            global collectterms
-            collectterms.sort(key=default_sort_key)
-            collectterms.reverse()
-            assert len(eq) == 1 and eq[0].lhs == f(x)
-            sol = eq[0].rhs
-            sol = expand_mul(sol)
-            for i, reroot, imroot in collectterms:
-                sol = collect(sol, x**i*exp(reroot*x)*sin(abs(imroot)*x))
-                sol = collect(sol, x**i*exp(reroot*x)*cos(imroot*x))
-            for i, reroot, imroot in collectterms:
-                sol = collect(sol, x**i*exp(reroot*x))
-            del collectterms
-
-            # Collect is splitting exponentials with rational powers for
-            # no reason.  We call powsimp to fix.
-            sol = powsimp(sol)
-
-            eq[0] = Eq(f(x), sol)
 
     else:
         # The solution is not solved, so try to solve it
@@ -3874,82 +3822,6 @@ def _undetermined_coefficients_match(expr, x, func=None, eq_homogeneous=S.Zero):
     return retdict
 
 
-def ode_nth_linear_constant_coeff_variation_of_parameters(eq, func, order, match):
-    r"""
-    Solves an `n`\th order linear differential equation with constant
-    coefficients using the method of variation of parameters.
-
-    This method works on any differential equations of the form
-
-    .. math:: f^{(n)}(x) + a_{n-1} f^{(n-1)}(x) + \cdots + a_1 f'(x) + a_0
-                f(x) = P(x)\text{.}
-
-    This method works by assuming that the particular solution takes the form
-
-    .. math:: \sum_{x=1}^{n} c_i(x) y_i(x)\text{,}
-
-    where `y_i` is the `i`\th solution to the homogeneous equation.  The
-    solution is then solved using Wronskian's and Cramer's Rule.  The
-    particular solution is given by
-
-    .. math:: \sum_{x=1}^n \left( \int \frac{W_i(x)}{W(x)} \,dx
-                \right) y_i(x) \text{,}
-
-    where `W(x)` is the Wronskian of the fundamental system (the system of `n`
-    linearly independent solutions to the homogeneous equation), and `W_i(x)`
-    is the Wronskian of the fundamental system with the `i`\th column replaced
-    with `[0, 0, \cdots, 0, P(x)]`.
-
-    This method is general enough to solve any `n`\th order inhomogeneous
-    linear differential equation with constant coefficients, but sometimes
-    SymPy cannot simplify the Wronskian well enough to integrate it.  If this
-    method hangs, try using the
-    ``nth_linear_constant_coeff_variation_of_parameters_Integral`` hint and
-    simplifying the integrals manually.  Also, prefer using
-    ``nth_linear_constant_coeff_undetermined_coefficients`` when it
-    applies, because it doesn't use integration, making it faster and more
-    reliable.
-
-    Warning, using simplify=False with
-    'nth_linear_constant_coeff_variation_of_parameters' in
-    :py:meth:`~sympy.solvers.ode.dsolve` may cause it to hang, because it will
-    not attempt to simplify the Wronskian before integrating.  It is
-    recommended that you only use simplify=False with
-    'nth_linear_constant_coeff_variation_of_parameters_Integral' for this
-    method, especially if the solution to the homogeneous equation has
-    trigonometric functions in it.
-
-    Examples
-    ========
-
-    >>> from sympy import Function, dsolve, pprint, exp, log
-    >>> from sympy.abc import x
-    >>> f = Function('f')
-    >>> pprint(dsolve(f(x).diff(x, 3) - 3*f(x).diff(x, 2) +
-    ... 3*f(x).diff(x) - f(x) - exp(x)*log(x), f(x),
-    ... hint='nth_linear_constant_coeff_variation_of_parameters'))
-           /       /       /     x*log(x)   11*x\\\  x
-    f(x) = |C1 + x*|C2 + x*|C3 + -------- - ----|||*e
-           \       \       \        6        36 ///
-
-    References
-    ==========
-
-    - https://en.wikipedia.org/wiki/Variation_of_parameters
-    - http://planetmath.org/VariationOfParameters
-    - M. Tenenbaum & H. Pollard, "Ordinary Differential Equations",
-      Dover 1963, pp. 233
-
-    # indirect doctest
-
-    """
-
-    gensol = ode_nth_linear_constant_coeff_homogeneous(eq, func, order, match,
-        returns='both')
-    match.update(gensol)
-    return _solve_variation_of_parameters(eq, func, order, match)
-
-
 def _solve_variation_of_parameters(eq, func, order, match):
     r"""
     Helper function for the method of variation of parameters and nonhomogeneous euler eq.
@@ -5901,4 +5773,5 @@ from .single import (NthAlgebraic, Factorable, FirstLinear, AlmostLinear,
         Bernoulli, SingleODEProblem, SingleODESolver, RiccatiSpecial,
         SecondNonlinearAutonomousConserved, FirstExact, Liouville, Separable,
         SeparableReduced, HomogeneousCoeffSubsDepDivIndep, HomogeneousCoeffSubsIndepDivDep,
-        HomogeneousCoeffBest, LinearCoefficients, NthOrderReducible, Hypergeometric2nd)
+        HomogeneousCoeffBest, LinearCoefficients, NthOrderReducible, Hypergeometric2nd,
+        NthConstCoeffHomogen, NthConstCoeffVar, NthConstCoeffUndet)
--- a/sympy/solvers/ode/ode.py
+++ b/sympy/solvers/ode/ode.py
@@ -5774,4 +5774,5 @@ from .single import (NthAlgebraic, Factorable, FirstLinear, AlmostLinear,
         SecondNonlinearAutonomousConserved, FirstExact, Liouville, Separable,
         SeparableReduced, HomogeneousCoeffSubsDepDivIndep, HomogeneousCoeffSubsIndepDivDep,
         HomogeneousCoeffBest, LinearCoefficients, NthOrderReducible, Hypergeometric2nd,
-        NthLinearConstantCoeffHomogeneous, NthLinearConstantCoeffVariationOfParameters, NthLinearConstantCoeffUndeterminedCoefficients)
+        NthLinearConstantCoeffHomogeneous, NthLinearConstantCoeffVariationOfParameters,
+        NthLinearConstantCoeffUndeterminedCoefficients)
--- a/sympy/solvers/ode/ode.py
+++ b/sympy/solvers/ode/ode.py
@@ -1059,9 +1059,9 @@ def classify_ode(eq, func=None, dict=False, ics=None, *, prep=True, xi=None, eta
         LinearCoefficients: ('linear_coefficients',),
         NthOrderReducible: ('nth_order_reducible',),
         Hypergeometric2nd: ('2nd_hypergeometric',),
-        NthConstCoeffHomogen: ('nth_linear_constant_coeff_homogeneous',),
-        NthConstCoeffVar: ('nth_linear_constant_coeff_variation_of_parameters',),
-        NthConstCoeffUndet: ('nth_linear_constant_coeff_undetermined_coefficients',),
+        NthLinearConstantCoeffHomogeneous: ('nth_linear_constant_coeff_homogeneous',),
+        NthLinearConstantCoeffVariationOfParameters: ('nth_linear_constant_coeff_variation_of_parameters',),
+        NthLinearConstantCoeffUndeterminedCoefficients: ('nth_linear_constant_coeff_undetermined_coefficients',),
     }
     for solvercls in solvers:
         solver = solvercls(ode)
@@ -5774,4 +5774,4 @@ from .single import (NthAlgebraic, Factorable, FirstLinear, AlmostLinear,
         SecondNonlinearAutonomousConserved, FirstExact, Liouville, Separable,
         SeparableReduced, HomogeneousCoeffSubsDepDivIndep, HomogeneousCoeffSubsIndepDivDep,
         HomogeneousCoeffBest, LinearCoefficients, NthOrderReducible, Hypergeometric2nd,
-        NthConstCoeffHomogen, NthConstCoeffVar, NthConstCoeffUndet)
+        NthLinearConstantCoeffHomogeneous, NthLinearConstantCoeffVariationOfParameters, NthLinearConstantCoeffUndeterminedCoefficients)
--- a/sympy/solvers/ode/single.py
+++ b/sympy/solvers/ode/single.py
@@ -120,7 +120,7 @@ class SingleODEProblem:
                     den = self.func**r[c1]
                     reduced_eq = Add(*[arg/den for arg in self.eq.args])
         if not reduced_eq:
-            reduced_eq = self.eq
+            reduced_eq = expand(self.eq)
         return reduced_eq
 
     @cached_property
@@ -2321,7 +2321,7 @@ class NthConstCoeffVar(SingleODESolver):
         The general solution.
 
         """
-        eq = self.ode_problem.eq_high_order_free
+        eq = self.ode_problem.eq
         f = self.ode_problem.func.func
         order = self.ode_problem.order
         x = self.ode_problem.sym
@@ -2629,7 +2629,7 @@ class NthConstCoeffUndet(SingleODESolver):
         ``_undetermined_coefficients_match()['trialset']``.
 
         """
-        eq = self.ode_problem.eq_high_order_free
+        eq = self.ode_problem.eq
         x = self.ode_problem.sym
         f = self.ode_problem.func.func
         order = self.ode_problem.order
--- a/sympy/solvers/ode/single.py
+++ b/sympy/solvers/ode/single.py
@@ -29,6 +29,8 @@ from sympy.solvers.solvers import solve
 from sympy.matrices import wronskian
 from sympy.solvers.deutils import ode_order, _preprocess
 from .subscheck import sub_func_doit
+from sympy.polys.matrices.linsolve import _lin_eq2dict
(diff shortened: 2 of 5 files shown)


sympy/solvers/ode/single.py
```diff
@@ -101,6 +106,24 @@ def order(self) -> int:
...
+            deriv_coef = self.eq.coeff(self.func.diff(self.sym, self.order))
+            if deriv_coef not in (1, 0):
+                r = deriv_coef.match(a*self.func**c1)
+                if r and r[c1]:
+                    den = self.func**r[c1]
+                    reduced_eq = Add(*[arg/den for arg in self.eq.args])
+        if not reduced_eq:
+            reduced_eq = self.eq
+        return reduced_eq
+
```
Mohitbalwani26, line 126: @oscarbenjamin This snippet is kept as separate function for now because when I keep it inside `preprocess_eq`, other solvers( pattern matching ones) fail to classify many examples as everytime the highest order coeff is removed. example : `f(x).diff(x)*f(x) + f(x)*f(x) - x*f(x)`. If you could please help in this, how this should be handled as this manipulation is required for const coeff hints.

sympy/solvers/ode/single.py
```diff
@@ -1975,5 +2045,654 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
...
+        x = self.ode_problem.sym
+        homogen_instance = NthConstCoeffHomogen(SingleODEProblem(eq, f(x), x))
+        gensols, collectterms = homogen_instance._get_sols(self.r)
+        # A generator of constants
+        constants = self.ode_problem.get_numbered_constants(num=len(gensols))
+        gsol = Add(*[i*j for (i, j) in zip(constants, gensols)])
+        gsol = Eq(f(x), gsol)
+        self.r.update({'list': gensols, 'sol': gsol, 'simpliy_flag': simplify_flag})
+        gsol = self._solve_undetermined_coefficients(self.r)
+        if simplify_flag:
```
Mohitbalwani26, line 2692: Also, this raises an exception  
```
File "/home/runner/work/sympy/sympy/sympy/solvers/ode/single.py", line 2691, in _get_general_solution
    gsol = self._solve_undetermined_coefficients(self.r)
 File "/home/runner/work/sympy/sympy/sympy/solvers/ode/single.py", line 2668, in _solve_undetermined_coefficients
    coeffvals = solve(list(coeffsdict.values()), coefflist)
    raise GeneratorsNeeded(
sympy.polys.polyerrors.GeneratorsNeeded: can't initialize from 'dict' without generators
```
whereas when I used inheritance, it was working fine.

sympy/solvers/ode/single.py
oscarbenjamin: I think this is fine.

sympy/solvers/ode/single.py
```diff
@@ -141,6 +163,53 @@ def is_autonomous(self):
...
+        to be good.
+
+        Examples
+        ========
+
+        >>> from sympy import Function, cos, sin
+        >>> from sympy.abc import x
+        >>> from sympy.solvers.ode.ode import _nth_linear_match
+        >>> f = Function('f')
+        >>> _nth_linear_match(f(x).diff(x, 3) + 2*f(x).diff(x) +
```
oscarbenjamin, line 185: These examples should be updated to show how to use the method here

sympy/solvers/ode/single.py
```diff
@@ -141,6 +163,53 @@ def is_autonomous(self):
         syms = self.eq.subs(self.func, u).free_symbols
         return x not in syms
 
+    def get_linear_coefficients(self, eq, func, order):
+        r"""
+        Matches a differential equation to the linear form:
+
+        .. math:: a_n(x) y^{(n)} + \cdots + a_1(x)y' + a_0(x) y + B(x) = 0
+
+        Returns a dict of order:coeff terms, where order is the order of the
```
oscarbenjamin, line 172: Can this function just use the `_lineq2dict` function defined here:
https://github.com/sympy/sympy/blob/6a57794ca820ac055106e7f421f12e99e5aac832/sympy/polys/matrices/linsolve.py#L171-L172

sympy/solvers/ode/ode.py
```diff
@@ -5901,4 +5773,5 @@ def _nonlinear_3eq_order1_type5(x, y, z, t, eq):
         Bernoulli, SingleODEProblem, SingleODESolver, RiccatiSpecial,
         SecondNonlinearAutonomousConserved, FirstExact, Liouville, Separable,
         SeparableReduced, HomogeneousCoeffSubsDepDivIndep, HomogeneousCoeffSubsIndepDivDep,
-        HomogeneousCoeffBest, LinearCoefficients, NthOrderReducible, Hypergeometric2nd)
+        HomogeneousCoeffBest, LinearCoefficients, NthOrderReducible, Hypergeometric2nd,
+        NthConstCoeffHomogen, NthConstCoeffVar, NthConstCoeffUndet)
```
oscarbenjamin, line 5777: We should keep the naming consistent. None of the other solvers has abbreviations like `Undet`. Also the other class names correspond directly to the hint names except in camel case rather than snake case.

sympy/solvers/ode/tests/test_single.py
```diff
@@ -2563,9 +2563,9 @@ def _get_examples_ode_sol_nth_linear_constant_coeff_homogeneous():
     'lin_const_coeff_hom_37': {
         'eq': f(x).diff(x, 10) - 2*f(x).diff(x, 6) + 2*f(x).diff(x, 5)
         + f(x).diff(x, 2) - 2*f(x).diff(x, 1) + f(x),
-        'sol': [Eq(f(x), (C1 + C2*x)*exp(x*r21) + (C10*sin(x*im(r24)) + C7*x*sin(x*im(r24)) + (
-        C8 + C9*x)*cos(x*im(r24)))*exp(x*re(r24)) + (C3*x*sin(x*im(r22)) + C6*sin(x*im(r22)
-        ) + (C4 + C5*x)*cos(x*im(r22)))*exp(x*re(r22)))],
+        'sol': [Eq(f(x), (C1 + C2*x)*exp(x*r21) + (-((C3 + C4*x)*sin(x*im(r22)))
+        + (C5 + C6*x)*cos(x*im(r22)))*exp(x*re(r22)) + (-((C7 + C8*x)*sin(x*im(r24)))
+        + (C10*x + C9)*cos(x*im(r24)))*exp(x*re(r24)))],
```
oscarbenjamin, line 2568: Why has this changed?

sympy/solvers/ode/ode.py
Mohitbalwani26: `NthLinearConstantCoeffUndeterminedCoefficients`  but wouldn't it be too long as a class name?

sympy/solvers/ode/tests/test_single.py
Mohitbalwani26: I think because of ordering of constants and simplification function is called inside solver earlier this was handled in `odesimp`.

sympy/solvers/ode/ode.py
oscarbenjamin: Hardly anyone ever needs to type the class name. The hint is what users have to type and is `nth_linear_constant_coeff_undetermined_coefficients`. It would have been good to make that shorter but it's too late for that now. For the class names consistency is more important than length.

sympy/solvers/ode/tests/test_single.py
oscarbenjamin: Can we not just let `odesimp` handle this?

Mohitbalwani26: In odesimp there was explicit condition for the hints having `nth_linear_const_coeff`, with the help of global variable it was handling the simplification which we discussed in the meeting. But now that simplification is moved in the solver.

sympy/solvers/ode/single.py
Mohitbalwani26:
```python
>>> _lin_eq2dict(eq, {f(x),f(x).diff(x),f(x).diff(x,2),f(x).diff(x,3)})
(x - sin(x), {f(x): -1, Derivative(f(x), x): cos(x) + 2, Derivative(f(x), (x, 2)): x, Derivative(f(x), (x, 3)): 1})
```
This also looks fine, only output format is changed. so should I change to use this function?

oscarbenjamin: Yes. that should simplify it a bit.

The docstring should be rewritten just to say that it gets the coefficients of the ODE. Is it more convenient for the calling code to have a dict with e.g. `f(x)` as the key or just a list of coefficients in order of derivatives?

sympy/solvers/ode/tests/test_single.py
oscarbenjamin: I still don't understand why we can't just let `odesimp` do all the simplification. If it can't figure out which expressions to collect on then maybe odesimp needs to be improved.

Mohitbalwani26: so If we keep that simplification part in `odesimp` then we also need to pass the list of collectterms to it so that it can handle.
If you can suggest some other way, it would be helpful.

sympy/solvers/ode/single.py
Mohitbalwani26: I think dictionary is good as it provides more readability. If you want me to change to list, I will do that

sympy/solvers/ode/tests/test_single.py
oscarbenjamin: Why can't `odesimp` just figure out which terms to collect on itself?

Mohitbalwani26: In this case the terms are roots of that equation with real and imaginary parts also. I will try to think if it can be managed

sympy/solvers/ode/single.py
Mohitbalwani26: @oscarbenjamin `_lin_eq2dict()` does not check whether the given ODE is linear. so should I keep `nth_linear_match` code only?
```python
>>> eq=f(x).diff(x, 3) + 2*f(x).diff(x) +x*f(x).diff(x, 2) + cos(x)*f(x).diff(x) + x - f(x) -sin(f(x))
>>> _nth_linear_match(eq, f(x), 3)
None
>>> _lin_eq2dict(eq,{f(x),f(x).diff(x),f(x).diff(x,2),f(x).diff(x,3)})
(x - sin(f(x)), {f(x): -1, Derivative(f(x), x): cos(x) + 2, Derivative(f(x), (x, 2)): x, Derivative(f(x), (x, 3)): 1})
```

sympy/solvers/ode/tests/test_single.py
Mohitbalwani26: I had a look at `odesimp` it solves for `func` basically and handles constants. I couldn't come up with a way to handle our case. if you can have a look and suggest please.

sympy/solvers/ode/single.py
oscarbenjamin: Oh yes, I forgot about that.

You could check if the rhs or any of the values of the dict has `f(x)` like:
```python
In [1]: sin(f(x)).has(f(x))
Out[1]: True
```

sympy/solvers/ode/tests/test_single.py
oscarbenjamin: Wherever the old code is calling collect with `collectterms` it should be possible to add some code that calculates which terms to collect with by checking for e.g. `exp`, `sin` or `cos` terms that involve `t`.

Mohitbalwani26: Old code is calling collect on `x**i*exp(reroot*x)*sin(abs(imroot)*x)` where i is the multiplicity. so if we want odesimp to handle this then it will require `(i,imroot,reroot)` also.

oscarbenjamin: Why does it need them? Why can't it just do `expr.atoms(exp)`?

The `simpsol` function in `systems.py` doesn't need to be told exactly which terms to collect on so why does `odesimp`?

Mohitbalwani26:
```python
sol = expand_mul(sol)
arr=eq[0].rhs.atoms(exp)
arr2=eq[0].rhs.atoms(sin)
arr3=eq[0].rhs.atoms(cos)
for i in arr:
    sol = collect(sol,i)
for i in arr2:
    sol = collect(sol,i)
for i in arr3:
    sol = collect(sol,i)
sol = powsimp(sol)
return Eq(f(x), sol)
```
I tried this, it changes a lot of current expected solutions for these hints.

oscarbenjamin: If you can make it pass the same arguments to `collect` then it should give the same results.

Mohitbalwani26: how do i pass that because old code does `sol = collect(sol, x**i*exp(reroot*x)*sin(abs(imroot)*x))`

oscarbenjamin: Can you give an example of input and expected output as well as the arguments that are passed to `collect`?

Mohitbalwani26:
```python
>>> eq=-x**4*sin(x - 1) + f(x) + Derivative(f(x), (x, 2))
[(0, 0, -1), (0, 0, 1)]  #(multiplicity, reroot, imroot)
sol = collect(sol, x**i*exp(reroot*x)*sin(abs(imroot)*x))
sol = collect(sol, x**i*exp(reroot*x)*cos(imroot*x))
sol = collect(sol, x**i*exp(reroot*x))
sol = powsimp(sol)
expected_sol = Eq(f(x), C1*sin(x) + C2*cos(x) - x**5*cos(x - 1)/10 + x**4*sin(x - 1)/4 + x**3*cos(x - 1)/2 - 3*x**2*sin(x - 1)/4 - 3*x*cos(x - 1)/4)
```
The code snippet which I shared [here](https://github.com/sympy/sympy/pull/21654#discussion_r658712291) gives `Eq(f(x), C1*sin(x) + C2*cos(x) + x**2*(x**2 - 3)*sin(x - 1)/4 + x*(-2*x**4 + 10*x**2 - 15)*cos(x - 1)/20)`

oscarbenjamin: This discussion would be a lot quicker if you would communicate more clearly.

What is `sol` in this example?

Mohitbalwani26: sorry for the miscommunication.
Let's take the example `eq=-8*f(x) + 12*Derivative(f(x), x) - 6*Derivative(f(x), (x, 2)) + Derivative(f(x), (x, 3))` the root is 2. so the collectterms (which is an array of tupe having `(multiplicity, realroot, imaginaryPart)`) will be  `[(2, 2, 0), (1, 2, 0), (0, 2, 0)]`.
general sol returned will be `Eq(f(x), C1*exp(2*x) + C2*x*exp(2*x) + C3*x**2*exp(2*x))`.
After doing collect on roots the expected simplified solution is `Eq(f(x), (C1 + x*(C2 + C3*x))*exp(2*x))`.
If it is still not clear, can we have discuss this on meet?

oscarbenjamin: We can do that like:
```python
In [24]: C1, C2, C3 = symbols('C1:4')

In [25]: eq = Eq(f(x), C1*exp(2*x) + C2*x*exp(2*x) + C3*x**2*exp(2*x)).rhs

In [26]: eq
Out[26]: 
    2⋅x         2⋅x       2  2⋅x
C₁⋅ℯ    + C₂⋅x⋅ℯ    + C₃⋅x ⋅ℯ   

In [27]: eq.collect(eq.atoms(exp))
Out[27]: 
⎛                2⎞  2⋅x
⎝C₁ + C₂⋅x + C₃⋅x ⎠⋅ℯ 


In [28]: eq.collect(eq.atoms(exp), horner)
Out[28]: 
                      2⋅x
(C₁ + x⋅(C₂ + C₃⋅x))⋅ℯ 
```
Actually I prefer `Out[27]` though. I don't see why we would want to change it to `Out[28]`.

Mohitbalwani26: Yes `out[27]` is achievable, but this was just one of the examples. I think I should commit and change the examples and then we can discuss further if the examples look simplified or not.

Mohitbalwani26:
@oscarbenjamin can you review this now?

sympy/solvers/ode/ode.py
```diff
@@ -5901,4 +5741,5 @@ def _nonlinear_3eq_order1_type5(x, y, z, t, eq):
         Bernoulli, SingleODEProblem, SingleODESolver, RiccatiSpecial,
         SecondNonlinearAutonomousConserved, FirstExact, Liouville, Separable,
         SeparableReduced, HomogeneousCoeffSubsDepDivIndep, HomogeneousCoeffSubsIndepDivDep,
-        HomogeneousCoeffBest, LinearCoefficients, NthOrderReducible, Hypergeometric2nd)
+        HomogeneousCoeffBest, LinearCoefficients, NthOrderReducible, Hypergeometric2nd,
+        NthLinearConstantCoeffHomogeneous, NthLinearConstantCoeffVariationOfParameters, NthLinearConstantCoeffUndeterminedCoefficients)
```
oscarbenjamin, line 5745: This line is too long

sympy/solvers/ode/ode.py
```diff
@@ -1845,39 +1829,27 @@ def odesimp(ode, eq, func, hint):
     if eq.lhs == func and not eq.rhs.has(func):
         # The solution is already solved
         eq = [eq]
-
-        # special simplification of the rhs
         if hint.startswith("nth_linear_constant_coeff"):
```
oscarbenjamin, line 1832: Do we still need to check the hint here? It would be better not to have simplification that is specific to one hint if possible.

sympy/solvers/ode/ode.py
```diff
@@ -1845,39 +1829,27 @@ def odesimp(ode, eq, func, hint):
...
-                sol = collect(sol, x**i*exp(reroot*x)*cos(imroot*x))
-            for i, reroot, imroot in collectterms:
-                sol = collect(sol, x**i*exp(reroot*x))
-            del collectterms
-
-            # Collect is splitting exponentials with rational powers for
-            # no reason.  We call powsimp to fix.
+            arr=eq[0].rhs.atoms(exp)
+            arr2=eq[0].rhs.atoms(sin)
+            arr3=eq[0].rhs.atoms(cos)
```
oscarbenjamin, line 1837: Can we just do `rhs.atoms(exp, sin, cos)` and then collect on that?

sympy/solvers/ode/single.py
```diff
@@ -12,19 +12,25 @@
...
+from sympy.simplify import collect, simplify, separatevars, logcombine, powsimp, trigsimp
 from sympy.simplify.radsimp import fraction
 from sympy.utilities import numbered_symbols
 from sympy.solvers.solvers import solve
+from sympy.matrices import wronskian
 from sympy.solvers.deutils import ode_order, _preprocess
+from .subscheck import sub_func_doit
+from sympy.polys.matrices.linsolve import _lin_eq2dict
+from sympy.polys.solvers import PolyNonlinearError
 from .hypergeometric import equivalence_hypergeometric, match_2nd_2F1_hypergeometric, get_sol_2F1_hypergeometric, match_2nd_hypergeometric
```
oscarbenjamin, line 34: This line is too long

sympy/solvers/ode/single.py
```diff
@@ -141,6 +165,53 @@ def is_autonomous(self):
...
+        >>> obj = SingleODEProblem(eq, f(x), x)
+        >>> obj.get_linear_coefficients(eq, f(x), 3) == None
+        True
+
+        """
+        f = func.func
+        x = func.args[0]
+        symset = {Derivative(f(x), x, i) for i in range(order+1)}
+        try:
+            rhs,lhs_terms = _lin_eq2dict(eq, symset)
```
oscarbenjamin, line 205: Space after comma

sympy/solvers/ode/tests/test_ode.py
```diff
@@ -196,7 +196,7 @@ def test_classify_ode():
          '1st_homogeneous_coeff_subs_dep_div_indep',
          '1st_power_series',
          'lie_group',
-         'nth_linear_constant_coeff_homogeneous',
+        #  'nth_linear_constant_coeff_homogeneous',
```
oscarbenjamin, line 199: If this is not supposed to match any more then this list should be deleted

sympy/solvers/ode/ode.py
Mohitbalwani26: The reason I did separately was because we also want to do `collect(expr, exp(x)*sin(x))` basically on all combinations of `exp*sin` and `exp*cos`

sympy/solvers/ode/ode.py
Mohitbalwani26: yes, because the way collect is done is specifically for the solutions returned by `nth_linear_constant_coeff` due to roots of characterstic equation. 
If we remove this if statement, It will change the form of solutions of other hints too as they already return solution solved for func.
Example:
```python
>>> eq = (x*log(f(x)/x) - 2*x)*Derivative(f(x), x) + f(x)
>>> dsolve(eq)  # keeping this If condition
Eq(f(x), -exp(C1)*LambertW(-x*exp(1 - C1)))

>>> dsolve(eq) #removing If condition
Eq(f(x), -exp(C1)*LambertW(-E*x*exp(-C1)))
```

sympy/solvers/ode/ode.py
oscarbenjamin: Does it make a different to the end result?

sympy/solvers/ode/ode.py
oscarbenjamin: Ideally we would apply the same simplification routines to the solutions from all solvers. For now we can leave this here but ultimately I think it should be removed.

Perhaps the best result would be to use the `simpsol` function from `systems.py`.

sympy/solvers/ode/ode.py
Mohitbalwani26: The ans is correct without it also, but it was combining some examples in a nicer way. If you want me to remove, I will do that

oscarbenjamin: When you say "in a nicer way" it's hard for me to know what you mean without any examples.

Mohitbalwani26: My bad. It was related to ordering only. 
I tried with `rhs.atoms(exp, sin, cos)` and ran collect over it. It change solution like:
```
The ODE is:
f(x) - Derivative(f(x), x) + Derivative(f(x), (x, 2))

The expected solution was
Eq(f(x), (C1*sin(sqrt(3)*x/2) + C2*cos(sqrt(3)*x/2))*exp(x/2))

What dsolve returned is:
Eq(f(x), (C1*sin(sqrt(3)*x/2) + C2*cos(sqrt(3)*x/2))*sqrt(exp(x)))

```

oscarbenjamin: Why does that come out different? There must be something else turning `exp(x/2)` into `sqrt(exp(x))` because `collect` shouldn't do that.

Mohitbalwani26: The reason was as `.atoms()` returns a set so it was returning `{cos(sqrt(3)*x/2), sin(sqrt(3)*x/2), exp(x/2)}` for the above mentioned example whereas we wanted to collect first on exp terms. That's why I have used a list now.

oscarbenjamin: What happens if you use the `simpsol` function from `systems.py` here? Does it give different results?

Mohitbalwani26: As far as I understood, simpsol takes list of symbols to be collected on but we want this collect to be on given functions i.e exp,sin and cos.
Although I tried by passing constants like `simpsol(eq, [C1], constants)`, here is the example of simple equations: 
```python
>>> eq=f(x).diff(x, 3) - f(x).diff(x) - sinh(x) - exp(x)
>>> dsolve(eq)
[Eq(f(x), C1*exp(-x) + C2 + C3*exp(x) + x*(exp(x) + sinh(x))/2)]   #solution returned by simpsol
Eq(f(x), C1 + C3*exp(-x) + x*sinh(x)/2 + (C2 + x/2)*exp(x))   #expected solution

>>> eq=f(x).diff(x, 2) - f(x) - exp(x - 1)
>>> dsolve(eq)
[Eq(f(x), C1*exp(-x) + C2*exp(x) + x*exp(x - 1)/2)]  #solution returned by simpsol
Eq(f(x), C2*exp(-x) + (C1 + x*exp(-1)/2)*exp(x))  #expected solution
```

oscarbenjamin: The idea with `simpsol` is that you tell it to collect like `simpsol(eq, [C1, C2, ...], [t])` and then it understands that `t` could mean `exp(t)`, `sin(t)*exp(t)` etc. It finds everything that involves `t`. You can also tell it to collect like `simpsol(eq, [t], [C1, C2, ...])` to reverse the order between collecting on `t` vs collecting on the constants.

Mohitbalwani26: I tried `simpsol(eq, [x], constants)`  as well as `simpsol(eq, [x], constants)` . But for some of the examples it doesn't collect properly. Two of them are:
```
The ODE is:
-exp(x) - sinh(x) - Derivative(f(x), x) + Derivative(f(x), (x, 3))

The expected solution was
[Eq(f(x), C1 + C3*exp(-x) + x*sinh(x)/2 + (C2 + x/2)*exp(x))]

What dsolve returned is:
Eq(f(x), C1 + C2*exp(-x) + C3*exp(x) + x*exp(x)/2 + x*sinh(x)/2)

```

```
The ODE is:
-x - x*exp(-x) - f(x) + Derivative(f(x), x) + 5*Derivative(f(x), (x, 2)) + 3*Derivative(f(x), (x, 3))

The expected solution was
[Eq(f(x), C3*exp(x/3) - x + (C1 + C2*x - x**3/24 - 3*x**2/32)*exp(-x) - 1)]

What dsolve returned is:
Eq(f(x), C1*x*exp(-x) + C2*exp(-x) + C3*exp(x/3) - x**3*exp(-x)/24 - 3*x**2*exp(-x)/32 - x - 1)

```

oscarbenjamin: Maybe `simpsol` can be improved. I'm still not sure which of these forms I prefer. I think that this is something good to discuss in the meeting. Can you collect examples of how solutions look when collected in different ways so that we can compare them in the meeting?

Mohitbalwani26: okay, sure I will write them.

Mohitbalwani26: Here is the list of examples. [Link](https://docs.google.com/document/d/1i4Zr22VXeDRA0eialHRlLmp-2zbp_ALUdGr21ceOnd4/edit?usp=sharing)

oscarbenjamin: Which way round are you passing the arguments to `simpsol`? Does that make a difference? How is `simpsol` called in `systems.py`?

Mohitbalwani26: These examples in the doc were generated using `simpsol(eq, constants, [x])` . In most of the examples it wasn't making any difference when simpsol was called like this `simpsol(eq, [x],  constants)`.  In this [comment](https://github.com/sympy/sympy/pull/21654#discussion_r659735157) only these two examples doesn't change when the order of arguments was swapped.

Mohitbalwani26: In `systems.py` simpsol is called like `simpsol(sol, [t], constants, doit=doit)` where t is Independent variable in the system of ODEs

Mohitbalwani26:
@oscarbenjamin PTAL. Is there anything which needs to be changed?

oscarbenjamin:
Looks good.

Issue #21519, opened 2021-05-25
[GSoC] [Track]Implement Joint classes


sidhu1012:
The following tasks need to be implemented(#9809)-

- [x] `Joints` abstract class
- [x] `PinJoint` class
- [x] `SlidingJoint` class
- [x] `CylindricalJoint` class
- [x] `PlanarJoint` class

ping @moorepants @Sc0rpi0n101

moorepants:
As before, please try to build off of the old prs so those commits are included in sympy.

sidhu1012:
I'll start a fresh PR but would `co-author` to include others, fresh PR would help in focusing on less code and less mess.

sidhu1012:
@moorepants [Three Link Conical Pendulum](https://pydy.readthedocs.io/en/latest/examples/three-link-conical-pendulum.html) is an example of which type of joint?

moorepants:
I don't know what "co-author" means. Are you building on the prior work or not? I prefer incorporating prior author commits so that they get proper credit.

sidhu1012:
By "co-author" I mean, I'll make a multiple author commit , so that everyone can get their credits .

namannimmo10:
> I'll make a multiple-author commit

I think you should pull down the original PR and edit on a branch that starts from that PR, then it will include commits from the prior author and you. (cherry-pick all the commits)

sidhu1012:
@moorepants  Can you please share a couple of examples for all the four Joint type, something like:
```
PinJoint - examples
SlidingJoint -  examples
CylindricalJoint - examples
PlanarJoint - Examples
```

So that I can differentiate between join types.

Thanks

moorepants:
I shared them all in the meeting. The pydy `examples` and `docs` folders has them.

sidhu1012:
Diagram to be done :- https://github.com/sympy/sympy/pull/21564#issuecomment-865745168

TJStienstra:
@sidhu1012 and @moorepants the joints have been implemented.

moorepants:
Congrats!

PR #21674, opened 2021-06-28
Implement SlidingJoint

sidhu1012:

#21519

Added SlidingJoint Class

#### Other comments
ping @moorepants @Sc0rpi0n101 

#### Release Notes

* physics.mechanics
    * Implemented a PrismaticJoint class that generate the kinematics of a prismatic joint.


sympy/physics/mechanics/joint.py
```diff
@@ -0,0 +1,485 @@
...
+    connected via SlidingJoint through their masscenters.
+
+    >>> from sympy.physics.mechanics import Body, SlidingJoint
+    >>> parent = Body('parent')
+    >>> child = Body('child')
+    >>> P = SlidingJoint('P', parent, child)
+    >>> P.coordinates()
+    [P_x(t)]
+    >>> P.speeds()
+    [P_v(t)]
```
moorepants, line 406: x_P(t)

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -0,0 +1,326 @@
...
+    assert A.ang_vel_in(N) == omega*N.y
+    assert A.ang_vel_in(N).express(A) == omega* A.x
+    assert A.ang_vel_in(N).magnitude() == omega
+    assert expand_mul(A.ang_vel_in(N).angle_between(A.x)) == 0
+    assert C.masscenter.vel(N).simplify() == - omega*N.z
+    assert C.masscenter.pos_from(P.masscenter) == N.x
+
+    #Both joint pos id defined but different axes
+    N, A, P, C = _generate_body()
+    PinJoint('J', P, C, parent_joint_pos=N.x, child_joint_pos=A.x, child_axis=A.x+A.y)
```
moorepants, line 178: Looking only at the orientation for this example with the child rotation axis as `rC = A.x + A.y` I have this figure:
image

`uC = cross(As.x, N.x).normalize()`

My code to generation the orientation and angular velocity expressions is:

```
In [1]: import sympy as sm

In [2]: import sympy.physics.mechanics as me

In [3]: N = me.ReferenceFrame('N')

In [4]: A = me.ReferenceFrame('A')

In [5]: As = me.ReferenceFrame('As')

In [6]: rC = A.x + A.y

In [7]: As.orient_axis(N, -N.z, sm.pi/4)  # note the negative N.z here

In [8]: theta = me.dynamicsymbols('theta')

In [9]: A.orient_axis(As, N.x, theta)

In [11]: sm.simplify(A.dcm(N))
Out[11]: 
Matrix([
[sqrt(2)/2, -sqrt(2)*cos(theta(t))/2, -sqrt(2)*sin(theta(t))/2],
[sqrt(2)/2,  sqrt(2)*cos(theta(t))/2,  sqrt(2)*sin(theta(t))/2],
[        0,           -sin(theta(t)),            cos(theta(t))]])

In [13]: A.ang_vel_in(N).express(N)
Out[13]: Derivative(theta(t), t)*N.x

In [14]: A.ang_vel_in(N).express(A).simplify()
Out[14]: sqrt(2)*Derivative(theta(t), t)/2*A.x + sqrt(2)*Derivative(theta(t), t)/2*A.y

In [16]: A.x.express(N).simplify()
Out[16]: sqrt(2)/2*N.x - sqrt(2)*cos(theta(t))/2*N.y - sqrt(2)*sin(theta(t))/2*N.z

In [16]: A.x.express(N).simplify()
Out[16]: sqrt(2)/2*N.x - sqrt(2)*cos(theta(t))/2*N.y - sqrt(2)*sin(theta(t))/2*N.z

In [17]: A.z.express(N).simplify()
Out[17]: - sin(theta(t))*N.y + cos(theta(t))*N.z

In [18]: A.y.express(N).simplify()
Out[18]: sqrt(2)/2*N.x + sqrt(2)*cos(theta(t))/2*N.y + sqrt(2)*sin(theta(t))/2*N.
```

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -0,0 +1,326 @@
...
+    assert expand_mul(N.x.angle_between(A.x + A.y)) == 0 #Axis are aligned
+    assert (A.x + A.y).express(N).simplify() == sqrt(2)*N.x
+    assert A.dcm(N).simplify() == Matrix([[sqrt(2)/2, -sqrt(2)*cos(theta)/2, -sqrt(2)*sin(theta)/2], \
+        [sqrt(2)/2, sqrt(2)*cos(theta)/2, sqrt(2)*sin(theta)/2], [0, -sin(theta), cos(theta)]])
+    assert A.ang_vel_in(N) == omega*N.x
+    assert A.ang_vel_in(N).express(A).simplify() == (omega*A.x + omega*A.y)/sqrt(2)
```
moorepants, line 184: I get `(omega*A.x + omega*A.y)*sqrt(2)/2` in my example above.

moorepants:
```
In [14]: A.ang_vel_in(N).express(A).simplify()
Out[14]: sqrt(2)*Derivative(theta(t), t)/2*A.x + sqrt(2)*Derivative(theta(t), t)/2*A.y
```

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -0,0 +1,326 @@
...
+    assert C.masscenter.vel(N).simplify() == - omega*N.z
+    assert C.masscenter.pos_from(P.masscenter) == N.x
+
+    #Both joint pos id defined but different axes
+    N, A, P, C = _generate_body()
+    PinJoint('J', P, C, parent_joint_pos=N.x, child_joint_pos=A.x, child_axis=A.x+A.y)
+    assert expand_mul(N.x.angle_between(A.x + A.y)) == 0 #Axis are aligned
+    assert (A.x + A.y).express(N).simplify() == sqrt(2)*N.x
+    assert A.dcm(N).simplify() == Matrix([[sqrt(2)/2, -sqrt(2)*cos(theta)/2, -sqrt(2)*sin(theta)/2], \
+        [sqrt(2)/2, sqrt(2)*cos(theta)/2, sqrt(2)*sin(theta)/2], [0, -sin(theta), cos(theta)]])
```
moorepants, line 182: This direction cosine matrix matches mine above.

sympy/physics/mechanics/tests/test_joint.py
moorepants: Sorry, put this in the wrong PR.

sympy/physics/mechanics/tests/test_joint.py
sidhu1012: Yes it's same as mine, I just simplified the output, they are same. This way it was easier to write

sympy/physics/mechanics/tests/test_joint.py
sidhu1012: Yes all other cases would match , I guess

sidhu1012: @moorepants Btw it's sliding joint PR

sidhu1012:
@moorepants SlidingJoint is also ready, it may have some merge conflicts when PinJoint would be merged though

sympy/physics/mechanics/joint.py
```diff
@@ -0,0 +1,485 @@
...
+        self._child.frame.set_ang_vel(self._parent.frame, self._speeds[0] * self._parent_axis.normalize())
+
+    def _set_linear_velocity(self):
+        self._parent_joint.set_vel(self._parent.frame, 0)
+        self._child_joint.set_vel(self._parent.frame, 0)
+        self._child_joint.set_pos(self._parent_joint, 0)
+        self._child.masscenter.v2pt_theory(self._parent.masscenter, self._parent.frame, self._child.frame)
+
+
+class SlidingJoint(Joint):
```
moorepants, line 354: Maybe we should simplify call this `PrismaticJoint`, as the word "sliding" is pretty ambiguous. The planar joint is also "sliding".

moorepants: The PinJoint is also probably better named "RevoluteJoint".

https://en.wikipedia.org/wiki/Prismatic_joint

https://en.wikipedia.org/wiki/Revolute_joint

sidhu1012: Okay, I'll change the name.

sidhu1012: @moorepants `PinJoint` seems nice, shall I change it too?

moorepants: Either way is fine for pin joint. Pin is shorter.

Prismatic could be "SliderJoint" to maybe match the accepted terminology better. I'm also fine with either too. It's not too big a deal.

sympy/physics/mechanics/joint.py
```diff
@@ -543,3 +543,137 @@ def _set_linear_velocity(self):
...
+    >>> J = SlidingJoint('J', C, P, coordinates=x, speeds=v)
+
+    We can check the kinematics now.
+
+    >>> P.masscenter.pos_from(C.masscenter)
+    x(t)*N.x
+    >>> P.masscenter.vel(N)
+    v(t)*N.x
+    >>> J.kdes()
+    [v(t) - Derivative(x(t), t)]
```
moorepants, line 639: Similar to pin joint you could also add a basic chained slider example. There is this one in PyDy: https://github.com/pydy/pydy/blob/master/pydy/models.py#L16, just multiple masses attached together by springs.

sympy/physics/mechanics/joint.py
```diff
@@ -543,3 +543,137 @@ def _set_linear_velocity(self):
...
+    speed : dynamicsymbol, optional
+        Speed of the joint.
+    parent_joint : Point
+        The joint's point where parent body is connected to the joint.
+    child_joint : Point
+        The joint's point where child body is connected to the joint.
+    parent_axis : Vector
+        The axis of parent frame.
+    child_axis : Vector
+        The axis of child frame.
```
moorepants, line 604: I updated the wording for all these parameters and attributes in the pin joint pr. Make them consistent here.

sympy/physics/mechanics/joint.py
```diff
@@ -543,3 +543,137 @@ def _set_linear_velocity(self):
...
+            y = dynamicsymbols('v' + '_' + self._name)
+            speed = y
+        speeds.append(speed)
+        return speeds
+
+    def _orient_frames(self):
+        self._child.frame.orient_axis(self._parent.frame, self._parent_axis, 0)
+        angle, axis = self._align_axis(self._parent_axis, self._child_axis)
+        if axis != Vector(0):
+            self._child.frame.orient_axis(self._parent.frame, axis, angle)
```
moorepants, line 669: I think this will need to be the same as PinJoint's implementation, no? Maybe this method belongs in `Joint`.

sympy/physics/mechanics/joint.py
```diff
@@ -543,3 +543,137 @@ def _set_linear_velocity(self):
...
+
+    def _set_angular_velocity(self):
+        return
```
moorepants, line 672: This should set the angular velocity between the frames to 0.

sympy/physics/mechanics/joint.py
sidhu1012: I already have `linear mass spring damper` example in docstrings, isn't one enough in docstring? Others can go into examples module?

moorepants: I don't see it here. Where is it?

In PinJoint we have two examples in the docstring: 1) shows the simplest pin joint and it's class features, 2) super simple multibody system kinematics (simple double pendulum). If we follow that pattern it makes sense to have a 3 body slider here.

sympy/physics/mechanics/joint.py
sidhu1012: `PinJoint` seems fine to me, I have changed `SlidingJoint` to `PrismaticJoint`

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -316,3 +317,111 @@ def test_pinjoint_pi():
...
                 child_axis=-C.frame.x-C.frame.y-C.frame.z)
     assert J._generate_vector() == P.frame.x
+
+
+def test_slidingjoint():
+    P = Body('P')
+    C = Body('C')
+    x, v = dynamicsymbols('x_S, v_S')
+    S = PrismaticJoint('S', P, C)
+    assert S._name == 'S'
```
moorepants, line 327:
```suggestion
    assert S.name == 'S'
```

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -316,3 +317,111 @@ def test_pinjoint_pi():
...
+    assert S.coordinates == [x]
+    assert S.speeds == [v]
+    assert S.kdes == [v - x.diff(t)]
+    assert S.parent_axis == P.frame.x
+    assert S.child_axis == C.frame.x
+    assert S.child_point.pos_from(C.masscenter) == Vector(0)
+    assert S.parent_point.pos_from(P.masscenter) == Vector(0)
+    assert S.parent_point.pos_from(S.child_point) == - x * P.frame.x
+    assert P.masscenter.pos_from(C.masscenter) == - x * P.frame.x
+    assert C.masscenter.vel(P.frame) == v * P.frame.x
```
moorepants, line 339: Check if expressed in C.frame is correct.

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -316,3 +317,111 @@ def test_pinjoint_pi():
...
+
+    l, m = symbols('l m')
+    S = PrismaticJoint('S', P, C, parent_joint_pos= l * P.frame.x,
+                    child_joint_pos= m * C.frame.y, parent_axis = P.frame.z)
+
+    assert S.parent_axis == P.frame.z
+    assert S.child_point.pos_from(C.masscenter) == m * C.frame.y
+    assert S.parent_point.pos_from(P.masscenter) == l * P.frame.x
+    assert S.parent_point.pos_from(S.child_point) == - x * P.frame.z
+    assert P.masscenter.pos_from(C.masscenter) == - x * P.frame.x
```
moorepants, line 350: Should this have `l` and `m` in it?

moorepants: This should be `m*C.y - x*P.z + l*P.x`. Right?

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -316,3 +317,111 @@ def test_pinjoint_pi():
...
+    assert C.masscenter.vel(P.frame) == v * P.frame.z
```
moorepants, line 351: For all angular velocity and velocity assertions check in both frames to see if they are correct.

sympy/physics/mechanics/tests/test_joint.py
moorepants: For future joint implementations you need to post a figure for these examples. Right now, I'm having to draw a figure and determine the solutions myself to check these results. If you post a figure I can look at it and the results and check what you've done without having to work out the whole problem myself.

sidhu1012: All the cases here are same as of `PinJoint`.

moorepants: Why isn't m and l in the position vector from mass center to mass center?

sidhu1012: > Why isn't m and l in the position vector from mass center to mass center?

Because `parent_joint_pos` is defined in `P.frame.x` and `child_joint_pos` is defined in `C.frame.y` but the `parent_axis` is `P.frame.z'. The algined axes are `P.z` and `C.x` , that's why `m` and 'l' are not in `pos_vector`.

If I do something like-
```
>>> S = PrismaticJoint('S', P, C, parent_joint_pos= l * P.frame.z, child_joint_pos= m * C.frame.x, parent_axis = P.frame.z)
>>> P.masscenter.pos_from(C.masscenter)
(-l - x_S(t))*P_frame.z + m*C_frame.x
```

This way we have `l` and `m` in `pos_vector` because in `PrismaticJoint` the `velocity` and `pos_vector` are always about `parent_axis`.

sidhu1012: I have added it to tests.

sympy/physics/mechanics/joint.py
sidhu1012: Please check the examples now.

sidhu1012:
@moorepants Please review now.

sympy/physics/mechanics/tests/test_joint.py
moorepants: I took these lines:

```python
l, m = symbols('l m')
S = PrismaticJoint('S', P, C, parent_joint_pos= l * P.frame.x, child_joint_pos= m * C.frame.y, parent_axis = P.frame.z)
```

and I drew this figure:

image

The position vector (red r) from the mass center of the child to the parent is:

`m*C.y -x*C.x - l*P.x`

from what I can tell.

Please show my your figure and your vector calculation.

moorepants: C.y doesn't have to be -P.x (because this choice is arbitrary), but regardless of the direction m, l, and x should all be present.

sidhu1012: Yes, you are right, I got the problem , it is not in implementation but in writing tests , I'll fix it first thing tomorrow morning.

sidhu1012: Fixed!

doc/src/modules/physics/mechanics/api/joint.rst
```diff
@@ -15,3 +15,9 @@ PinJoint
 
 .. autoclass:: PinJoint
    :members:
+
+PrismaticJoint
+============
```
moorepants, line 20:
```suggestion
==============
```

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -315,4 +316,142 @@ def test_pinjoint_pi():
...
+        (x - sqrt(3)/3 + 1)*N.x + sqrt(3)/3*N.y - sqrt(3)/3*N.z
+    assert C.masscenter.vel(N) == v*N.x
+    assert C.masscenter.vel(N).express(A) == sqrt(3)*v/3*A.x + sqrt(3)*v/3*A.y - sqrt(3)*v/3*A.z
+    assert A.ang_vel_in(N) == 0
+    assert N.ang_vel_in(A) == 0
+
+    N, A, P, C = _generate_body()
+    m, n = symbols('m n')
+    PrismaticJoint('S', P, C, parent_joint_pos=m*N.x, child_joint_pos=n*A.x,
+                    child_axis=A.x+A.y-A.z, parent_axis=N.x-N.y+N.z)
```
sidhu1012, line 445: @moorepants My figure for this case

vector

The angle between the axes is theta(constant angle), so the child frame is rotated at an angle theta, wrt parent_frame where rotation_axis=cross(parent_axis, parent_frame.x)

sympy/physics/mechanics/joint.py
```diff
@@ -272,9 +336,9 @@ class PinJoint(Joint):
         The parent body of joint.
     child : Body
         The child body of joint.
-    coordinates: List of dynamicsymbols, optional
+    coordinates: dynamicsymbol, optional
         Generalized coordinates of the joint.
-    speeds : List of dynamicsymbols, optional
+    speeds : dynamicsymbol, optional
```
moorepants, line 341: Future joints will require a list. I think we should return lists of the symbols for consistency with the future joints.

sympy/physics/mechanics/joint.py
```diff
@@ -543,3 +533,197 @@ def _set_linear_velocity(self):
...
+
+
+class PrismaticJoint(Joint):
+    """Prismatic (Sliding) Joint.
+
+    Explanation
+    ===========
+
+    It is defined such that the child body translates with respect to the parent
+    body along the body fixed parent axis. The point of joint (sliding joint) is defined
```
moorepants, line 545:
```suggestion
    body along the body fixed parent axis. The location of the joint is defined
```

sympy/physics/mechanics/joint.py
```diff
@@ -543,3 +533,197 @@ def _set_linear_velocity(self):
...
+    by two points in each body which coincides initially. The direction cosine matrix between
```
moorepants, line 546:
```suggestion
    by two points in each body which coincides when the generalized coordinate is zero. The direction cosine matrix between
```

sympy/physics/mechanics/joint.py
```diff
@@ -543,3 +533,197 @@ def _set_linear_velocity(self):
...
+    Parameters
+    ==========
+
+    name : string
+        A unique name for the joint.
+    parent : Body
+        The parent body of joint.
+    child : Body
+        The child body of joint.
+    coordinates: dynamicsymbol, optional
```
moorepants, line 560:
```suggestion
    coordinates: list of dynamicsymbol, optional
```

sympy/physics/mechanics/joint.py
```diff
@@ -543,3 +533,197 @@ def _set_linear_velocity(self):
...
+        Generalized coordinates of the joint.
+    speeds : dynamicsymbol, optional
```
moorepants, line 562:
```suggestion
    speeds : list of dynamicsymbol, optional
```

sympy/physics/mechanics/joint.py
```diff
@@ -543,3 +533,197 @@ def _set_linear_velocity(self):
...
+    Matrix([
+    [1, 0, 0],
+    [0, 1, 0],
+    [0, 0, 1]])
+    >>> joint.child_point.pos_from(joint.parent_point)
+    x_PC(t)*P_frame.x
+
+    To further demonstrate the use of the prismatic joint, the kinematics of linear mass
+    spring damper that translates about the X axis of each connected body can be
+    created as follows.
```
moorepants, line 647: There is no spring and damper in this example because there are no forces. This is the kinematics of two masses that can slide on the x axis.

sympy/physics/mechanics/joint.py
```diff
@@ -543,3 +533,197 @@ def _set_linear_velocity(self):
...
+
+    >>> from sympy.physics.mechanics import PrismaticJoint, Body
+
+    First create bodies to represent the fixed ceiling and one to represent
+    a particle.
+
+    >>> ceiling = Body('C')
+    >>> Part = Body('P')
```
moorepants, line 655: Create two bodies that are chained, just like the pin joint double pendulum example. One body is no different than the prior example.

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -315,4 +316,142 @@ def test_pinjoint_pi():
...
+    assert P.ang_vel_in(C) == 0
+
+
+def test_slidingjoint_arbitrary_axis():
+    x, v = dynamicsymbols('x_S, v_S')
+
+    N, A, P, C = _generate_body()
+    PrismaticJoint('S', P, C, child_axis=-A.x)
+
+    assert -A.x.angle_between(N.x) == -pi
```
moorepants, line 376: Why would this return -pi? The function should return an angle that satisfies: 0 <= angle <= pi

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -315,4 +316,142 @@ def test_pinjoint_pi():
...
+    assert C.masscenter.pos_from(P.masscenter).express(N) == (x - sqrt(2)/2 + 1)*N.x + sqrt(2)/2*N.y
+    assert C.masscenter.vel(N).express(A) == v * (A.x + A.y)/sqrt(2)
+    assert C.masscenter.vel(N) == v*N.x
+    assert A.ang_vel_in(N) == 0
+    assert N.ang_vel_in(A) == 0
+
+    N, A, P, C = _generate_body()
+    PrismaticJoint('S', P, C, parent_joint_pos=N.x, child_joint_pos=A.x, child_axis=A.x+A.y-A.z)
+    assert N.x.angle_between(A.x + A.y - A.z) == 0 #Axis are aligned
+    assert (A.x + A.y - A.z).express(N) == sqrt(3)*N.x
```
moorepants, line 430: The parent axis is `N.x` (default) and the child axis is `A.x + A.y - A.z`. Shouldn't `A.x + A.y - A.z == N.x`?

moorepants: Actually, this is correct.

sympy/physics/mechanics/tests/test_joint.py
moorepants: The figure I was asking for is one that shows the position vectors (as I did in the figure I shared above).

sympy/physics/mechanics/tests/test_joint.py
sidhu1012: Because it was considering the above case to - `-(A.x.angle_between(N.x))`

sidhu1012: Fixed!

sympy/physics/mechanics/joint.py
sidhu1012: `PinJoint` and `PrismaticJoint` takes `speeds` and `coordinates` as dynamicsymbols because only one argument is required, but returns them as list.

sympy/physics/mechanics/joint.py
sidhu1012: I have updated the docstring example.

sympy/physics/mechanics/joint.py
sidhu1012: Done!

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -373,7 +373,7 @@ def test_slidingjoint_arbitrary_axis():
     N, A, P, C = _generate_body()
     PrismaticJoint('S', P, C, child_axis=-A.x)
 
-    assert -A.x.angle_between(N.x) == -pi
+    assert (-A.x).angle_between(N.x) == 0
```
moorepants, line 376: This is a fix for this test, but isn't a fix for the method. `angle_between()` doesn't behave as expected. In general with sympy objects you'd expect `-x.something()` to operate on `-x` not `x`. Is this an issue because `.x` is an attribute of `A`?

An issue should be opened about `angle_between()` not respecting the leading negative sign. And, for now, `angle_between()` needs an example like this in its doc string, otherwise users will be caught by this unexpected issue.

sympy/physics/mechanics/tests/test_joint.py
moorepants: Please see me comment in the new code.

sympy/physics/mechanics/joint.py
moorepants: Understood. This is the input parameter list. I thought it was the attribute list.

sympy/physics/mechanics/joint.py
```diff
@@ -543,3 +533,217 @@ def _set_linear_velocity(self):
...
+    >>> joint.child.frame.dcm(joint.parent.frame)
+    Matrix([
+    [1, 0, 0],
+    [0, 1, 0],
+    [0, 0, 1]])
+    >>> joint.child_point.pos_from(joint.parent_point)
+    x_PC(t)*P_frame.x
+
+    To further demonstrate the use of the prismatic joint, the kinematics of n-link
+    pendulum(n=2) about the X axis of each connected body can be created as follows.
```
moorepants, line 646: You can't create a pendulum with a sliding joint. You can create this https://github.com/pydy/pydy/blob/master/pydy/models.py#L16 (but with out springs and dampers in between the masses). This example should be two masses sliding, one moving relative to a fixed body and the other relative to the moving body.

sympy/physics/mechanics/joint.py
```diff
@@ -543,3 +533,217 @@ def _set_linear_velocity(self):
...
+    0
+
+    And finally, the linear velocities of the two particles can be computed
+    with respect to the ceiling.
+
+    >>> Part1.masscenter_vel(ceiling)
+    v_J1(t)*C_frame.x
+
+    >>> Part2.masscenter_vel(ceiling)
+    v_J1(t)*C_frame.x + Derivative(x_J1(t), t)*P1_frame.x
```
moorepants, line 707: This is supposed to be:

```
v_J1(t)*C_frame.x + v_J2(t)*P1_frame.x
```

sympy/physics/mechanics/joint.py
sidhu1012: I'll change that.

sympy/physics/mechanics/joint.py
sidhu1012: Bur derivative (x) = v

moorepants: Yes, but the joint's should ensure that the generalized speeds are used. We shouldn't see `Derivatie(coordinate, t)` in any of the kinematics.

moorepants: Also, it says `Derivative(x_J1(t), t)` J1 is not correct either.

sympy/physics/mechanics/tests/test_joint.py
sidhu1012: I think this problem is not limited to `angle_between()`, all member functions of class `Vector` are ignoring the leading negative signs.

sympy/physics/mechanics/joint.py
sidhu1012: > Also, it says `Derivative(x_J1(t), t)` J1 is not correct either.

Yes , I named joint 2 'J1' by mistake.

sidhu1012: > Yes, but the joint's should ensure that the generalized speeds are used. We shouldn't see `Derivatie(coordinate, t)` in any of the kinematics.

I agree, but how can we do that, part2 velocity is defined in part1's frame, we want it's velocity in C's frame.
The current velocity function returns the `velovity of part1 in C + derivative(pos_vector between part1 and part2)`. So how can we change this behaviour?

sidhu1012: PR #21730 might solve this issue.

sympy/physics/mechanics/tests/test_joint.py
moorepants: Maybe this only incidentally works:

```
In [1]: import sympy.physics.mechanics as me

In [3]: A = me.ReferenceFrame('A')

In [4]: -A.x.cross(A.y)
Out[4]: - A.z

In [5]: A.x.cross(A.y)
Out[5]: A.z
```

moorepants: If there is no way to make this work correctly without the parentheses, then we need to make a clear "gotcha" statement in the documentation of `angle_between()`, because it will give the incorrect values. We also need to make sure the use of `angle_between()` in `Joint` can't be caught by this issue in an edge case.

sympy/physics/mechanics/joint.py
moorepants: All joints should be setting the correctly velocities, no computing them.

moorepants: For example, if you are doing a `pos_vec.diff(t)` in your joint code, that's not the way to do it. The joint is defined and we know what the velocity is, so just use `.set_vel()` to set the correct velocities.

sidhu1012: We are set velocity for parent frame, but we can't for parent body' parent body.  So we have to compute it for that.

sympy/physics/mechanics/tests/test_joint.py
sidhu1012: Yes I made sure, `angle_between` in `Joint` would always be correct.

sympy/physics/mechanics/joint.py
sidhu1012: Fixed!

sidhu1012:
@moorepants This one is ready!

moorepants:
> @moorepants This one is ready!

What about the figure explaining the last, most general, test?

sidhu1012:
> What about the figure explaining the last, most general, test?

Will try to show it tomorrow.

sympy/physics/mechanics/tests/test_joint.py
sidhu1012: slider

@moorepants I hope I got it right this time.

sidhu1012:
> What about the figure explaining the last, most general, test?

https://github.com/sympy/sympy/pull/21674/files#r671642338

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -315,4 +316,142 @@ def test_pinjoint_pi():
...
+    assert (N.x-N.y+N.z).angle_between(A.x+A.y-A.z) == 0 #Axis are aligned
+    assert (A.x-A.y+A.z).express(N) == - 5/3*N.x - 1/3*N.y + 1/3*N.z
+    assert A.dcm(N) == Matrix([[-S(1)/3, -S(2)/3, S(2)/3],
+                                [S(2)/3, S(1)/3, S(2)/3],
+                                [-S(2)/3, S(2)/3, S(1)/3]])
+    assert C.masscenter.pos_from(P.masscenter) == (m + x)*N.x - x*N.y + x*N.z - n*A.x
+    assert C.masscenter.pos_from(P.masscenter).express(N) == \
+        (m + n/3 + x)*N.x + (2*n/3 - x)*N.y + (-2*n/3 + x)*N.z
+    assert C.masscenter.vel(N) == v*N.x - v*N.y + v*N.z
+    assert C.masscenter.vel(N).express(A) == v*A.x + v*A.y - v*A.z
```
moorepants, line 455: The magnitudes of these two vectors should be `v`.

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -315,4 +316,142 @@ def test_pinjoint_pi():
...
+    assert C.masscenter.vel(N).express(A) == sqrt(3)*v/3*A.x + sqrt(3)*v/3*A.y - sqrt(3)*v/3*A.z
+    assert A.ang_vel_in(N) == 0
+    assert N.ang_vel_in(A) == 0
+
+    N, A, P, C = _generate_body()
+    m, n = symbols('m n')
+    PrismaticJoint('S', P, C, parent_joint_pos=m*N.x, child_joint_pos=n*A.x,
+                    child_axis=A.x+A.y-A.z, parent_axis=N.x-N.y+N.z)
+    assert (N.x-N.y+N.z).angle_between(A.x+A.y-A.z) == 0 #Axis are aligned
+    assert (A.x-A.y+A.z).express(N) == - 5/3*N.x - 1/3*N.y + 1/3*N.z
```
moorepants, line 447:
```suggestion
    assert (A.x+A.y-A.z).express(N) == - 5/3*N.x - 1/3*N.y + 1/3*N.z
```

sympy/physics/mechanics/tests/test_joint.py
moorepants: Does this example imply that:

```
N.x = A.y
N.y = A.z
N.z = A.x
```

moorepants:
This looks good but we do not have tests for the most general cases for child_axis, parent_axis, etc.

PR #21689, opened 2021-06-30
Add new features to class `Body`

sidhu1012:
#### Release Notes

* physics.mechanics
    * Add `masscenter_vel()` to class `Body` to calculate velocity of one body's masscenter wrt other body's frame. 
    * Add `ang_vel_in()` to class `Body` to calculate angular velocity of one body's frame wrt other body's frame. 
    * Add `dcm()` to class Body to calculate dcm of one body's frame wrt other body's frame.

### DIFF
--- a/sympy/physics/mechanics/body.py
+++ b/sympy/physics/mechanics/body.py
@@ -275,3 +275,32 @@ class Body(RigidBody, Particle):  # type: ignore
         """
 
         return self.frame.ang_vel_in(body.frame)
+
+    def dcm(self, body):
+        """
+        Returns the direction cosine matrix relative to the provided body's
+        reference frame.
+
+        Parameters
+        ==========
+
+        body: Body
+            The body's reference frame which the direction cosine matrix of this body's frame
+            is formed relative to.
+
+        Example
+        =======
+
+        >>> from sympy.physics.mechanics import Body
+        >>> A = Body('A')
+        >>> B = Body('B')
+        >>> A.frame.orient_axis(B.frame, B.frame.x, 5)
+        >>> A.dcm(B)
+        Matrix([
+        [1,       0,      0],
+        [0,  cos(5), sin(5)],
+        [0, -sin(5), cos(5)]])
+
+        """
+
+        return self.frame.dcm(body.frame)
--- a/sympy/physics/mechanics/body.py
+++ b/sympy/physics/mechanics/body.py
@@ -246,7 +246,7 @@ class Body(RigidBody, Particle):  # type: ignore
 
         if body is None:
             body = self
-        frame = body.frame      
+        frame = body.frame
         return self.masscenter.vel(frame)
 
     def ang_vel_in(self, body):
--- a/sympy/physics/mechanics/body.py
+++ b/sympy/physics/mechanics/body.py
@@ -217,15 +217,14 @@ class Body(RigidBody, Particle):  # type: ignore
 
     def masscenter_vel(self, body):
         """
-        Returns the velocity of self's masscenter wrt body's referenceframe or
-        self's referenceframe.
+        Returns the velocity of the mass center with respect to the provided
+        rigid body or reference frame.
 
         Parameters
         ==========
 
-        body: Body/ReferenceFrame
-            The body about which the velocity is needed to be calculated.
-            If body is self by default.
+        body: Body or ReferenceFrame
+            The rigid body or reference frame to calculate the velocity in.
 
         Example
         =======
@@ -249,15 +248,14 @@ class Body(RigidBody, Particle):  # type: ignore
 
     def ang_vel_in(self, body):
         """
-        Returns the angular velocity of self's frame wrt body's referenceframe or
-        self's referenceframe.
+        Returns this body's angular velocity with respect to the provided
+        rigid body or reference frame.
 
         Parameters
         ==========
 
-        body: Body
-            The body about which the angluar velocity is needed to be calculated.
-            If body is self by default.
+        body: Body or ReferenceFrame
+            The rigid body or reference frame to calculate the angular velocity in.
 
         Example
         =======
@@ -282,15 +280,14 @@ class Body(RigidBody, Particle):  # type: ignore
 
     def dcm(self, body):
         """
-        Returns the direction cosine matrix relative to the provided body's
-        reference frame.
+        Returns the direction cosine matrix of this body relative to the
+        provided rigid body or reference frame.
 
         Parameters
         ==========
 
-        body: Body
-            The body's reference frame which the direction cosine matrix of this body's frame
-            is formed relative to.
+        body: Body or ReferenceFrame
+            The rigid body or reference frame to calculate the dcm.
 
         Example
         =======
--- a/sympy/physics/mechanics/body.py
+++ b/sympy/physics/mechanics/body.py
@@ -215,7 +215,7 @@ class Body(RigidBody, Particle):  # type: ignore
             raise TypeError("A Vector must be supplied to add torque.")
         self.loads.append((self.frame, vec))
 
-    def vel(self, body=None):
+    def masscenter_vel(self, body):
         """
         Returns the velocity of self's masscenter wrt body's referenceframe or
         self's referenceframe.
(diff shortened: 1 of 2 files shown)


sympy/physics/mechanics/body.py
```diff
@@ -214,3 +214,64 @@ def apply_torque(self, vec):
         if not isinstance(vec, Vector):
             raise TypeError("A Vector must be supplied to add torque.")
         self.loads.append((self.frame, vec))
+
+    def vel(self, body=None):
```
moorepants, line 218: The mass center does not move wrt to its body's frame, so you always need another body or frame here. I think you should accept a Body or a ReferenceFrame. Also a body can have an infinite number of points, so there are an infinite number of velocities. This should probably be named `masscenter_vel()` to be explicit about that.

sympy/physics/mechanics/body.py
```diff
@@ -214,3 +214,64 @@ def apply_torque(self, vec):
...
+        5*N.x
+
+        """
+
+        if body is None:
+            body = self
+        frame = body.frame
+        return self.masscenter.vel(frame)
+
+    def ang_vel_in(self, body):
```
moorepants, line 252: This should also take a body or frame.

sympy/physics/mechanics/body.py
```diff
@@ -214,3 +214,93 @@ def apply_torque(self, vec):
...
+        >>> B = Body('B', frame=N)
+        >>> A.frame.set_ang_vel(N, 5*N.x)
+        >>> A.ang_vel_in(B)
+        5*N.x
+
+        """
+
+        return self.frame.ang_vel_in(body.frame)
+
+    def dcm(self, body):
```
moorepants, line 279: Also a body or a frame.

sympy/physics/mechanics/body.py
moorepants: I'm trying to think whether we will every store other points on in a Body object other than the mass center. If we wouldn't then sticking with `vel()` as the name is ok. I'll think a bit about it.

sympy/physics/mechanics/body.py
sidhu1012: I think it should be body only, because in `Joints` we would always be having 2 bodies.

moorepants: I think users (including me) will be more fluid about the use of frames and bodies. Both will be present in different scripts and there is not reason not to accept frames in all these methods. If it accepts both, then it simplifies setting up problems. A frame and a body are quite synonymous. In fact, in my course I teach that. Please make these methods accept frames or bodies.

sidhu1012: Done

sympy/physics/mechanics/body.py
sidhu1012: > I'm trying to think whether we will every store other points on in a Body object other than the mass center. If we wouldn't then sticking with `vel()` as the name is ok. I'll think a bit about it.

Current implementation stores masscenter only.

moorepants: I know that. But I'm trying to imagine the future. A body has an infinite number of points and some of those are important for any giving multi-body model. The first that comes to mind is locations that forces are applied on the body. But that only makes sense if we store forces in bodies.

sidhu1012: `Body.loads` does store point, force tuple.

moorepants: Ah, so there is already a `.loads` attribute. That's interesting. More thought needs to go into this, especially since forces/torques have to ensure equal and opposite behavior.

sympy/physics/mechanics/body.py
```diff
@@ -214,3 +214,106 @@ def apply_torque(self, vec):
...
+    def masscenter_vel(self, body):
+        """
+        Returns the velocity of self's masscenter wrt body's referenceframe or
+        self's referenceframe.
+
+        Parameters
+        ==========
+
+        body: Body/ReferenceFrame
+            The body about which the velocity is needed to be calculated.
```
moorepants, line 227: The rigid body or reference frame to calculate the velocity in.

sympy/physics/mechanics/body.py
```diff
@@ -214,3 +214,106 @@ def apply_torque(self, vec):
...
+
+    def masscenter_vel(self, body):
+        """
+        Returns the velocity of self's masscenter wrt body's referenceframe or
+        self's referenceframe.
+
+        Parameters
+        ==========
+
+        body: Body/ReferenceFrame
```
moorepants, line 226:
```suggestion
        body: Body or ReferenceFrame
```

moorepants: See https://numpydoc.readthedocs.io/en/latest/format.html#parameters

sympy/physics/mechanics/body.py
```diff
@@ -214,3 +214,106 @@ def apply_torque(self, vec):
...
+            The body about which the velocity is needed to be calculated.
+            If body is self by default.
```
moorepants, line 228:
```suggestion
```

sympy/physics/mechanics/body.py
```diff
@@ -214,3 +214,106 @@ def apply_torque(self, vec):
...
+        if isinstance(body,  ReferenceFrame):
+            frame=body
+        elif isinstance(body, Body):
+            frame = body.frame
+        return self.masscenter.vel(frame)
+
+    def ang_vel_in(self, body):
+        """
+        Returns the angular velocity of self's frame wrt body's referenceframe or
+        self's referenceframe.
```
moorepants, line 253: Returns this body's angular velocity with respect to the provided rigid body or reference frame.

sympy/physics/mechanics/body.py
```diff
@@ -214,3 +214,106 @@ def apply_torque(self, vec):
...
+
+        Parameters
+        ==========
+
+        body: Body
+            The body about which the angluar velocity is needed to be calculated.
+            If body is self by default.
```
moorepants, line 260:
```suggestion
```

sympy/physics/mechanics/body.py
```diff
@@ -214,3 +214,106 @@ def apply_torque(self, vec):
...
+    def ang_vel_in(self, body):
+        """
+        Returns the angular velocity of self's frame wrt body's referenceframe or
+        self's referenceframe.
+
+        Parameters
+        ==========
+
+        body: Body
+            The body about which the angluar velocity is needed to be calculated.
```
moorepants, line 259:
```suggestion
            The rigid body or reference frame to calculate the angular velocity in.
```

sympy/physics/mechanics/body.py
```diff
@@ -214,3 +214,106 @@ def apply_torque(self, vec):
...
+
+    def ang_vel_in(self, body):
+        """
+        Returns the angular velocity of self's frame wrt body's referenceframe or
+        self's referenceframe.
+
+        Parameters
+        ==========
+
+        body: Body
```
moorepants, line 258:
```suggestion
        body: Body or ReferenceFrame
```

sympy/physics/mechanics/body.py
```diff
@@ -214,3 +214,106 @@ def apply_torque(self, vec):
...
+        if isinstance(body,  ReferenceFrame):
+            frame=body
+        elif isinstance(body, Body):
+            frame = body.frame
+        return self.frame.ang_vel_in(frame)
+
+    def dcm(self, body):
+        """
+        Returns the direction cosine matrix relative to the provided body's
+        reference frame.
```
moorepants, line 286: Returns the direction cosine matrix of this body relative to the provided rigid body or reference frame.

sympy/physics/mechanics/body.py
```diff
@@ -214,3 +214,106 @@ def apply_torque(self, vec):
...
+
+        Parameters
+        ==========
+
+        body: Body
+            The body's reference frame which the direction cosine matrix of this body's frame
+            is formed relative to.
```
moorepants, line 293: Adjust same as above.

moorepants:
The `Body` object should actually have all the methods and attributes of a `ReferenceFrame` and have all the methods and attributes of  a `Point` (but with `masscenter_` prepended to the methods/attributes names. We can start with these though.

sympy/physics/mechanics/body.py
sidhu1012: @moorepants Yes, I think it would make applying forces and torques a bit easier.

```
class Forces:
    pass

F = Force(Body1, Body2)
```

Forces acting on Body1 can be +ve, and equal and opposite on Body 2, and can append the `.loads` and `.torque` of respective bodies. There is just one things that is tough is , to apply force on point other than masscenter, as we may need to take multiple parameters in `Force` for each body something like-
`F = Forces(Body1, Body1_point=None, Body2, Body2_point=None)` 
where `Body_point` is the point on each body on which force is applied.

Now problem 2 is gravity, as it should be wrt Newtonian frame, so shal we do something like-
`F = Forces(body1, body1_point=None, body2, body2_point=None, gravity=Frame/Body/None)`
If gravity is None then gravity is not applied, else a frame/Body should be supplied in which gravity is applied.

moorepants: I don't think we need to special case a constant gravitational field. You'd simple do `Force(Body1, Body1, g*Body1.x)` where Body1 is the newtonian frame. You have to do it for all bodies, but we can think of ways to ease things if you have lots of bodies to apply it to.

We should open a new issue or discussion for this topic, to try to work out the design.

moorepants: Once again, have some example problems that show how the api will be used on a real problem is essential.

moorepants:
LGTM! Thanks.

PR #21692, opened 2021-06-30
refactor euler solvers

Mohitbalwani26:

SEE #18348 

Refactor Euler solvers and cleanup the constant coff code from ode.py

#### Release Notes

NO ENTRY


sympy/solvers/ode/tests/test_ode.py
```diff
@@ -604,114 +605,115 @@ def test_collect_respecting_exponentials():
 
 
 def test_undetermined_coefficients_match():
-    assert _undetermined_coefficients_match(g(x), x) == {'test': False}
-    assert _undetermined_coefficients_match(sin(2*x + sqrt(5)), x) == \
+    F = lambda eq: NthLinearConstantCoeffUndeterminedCoefficients(SingleODEProblem(eq, f(x), x))._undetermined_coefficients_match(eq,x)
```
oscarbenjamin, line 608: It's helpful during refactoring if the tests are changed as little as possible so we can be clear that the mostly unchanged tests still pass. This could be:
```
def _undetermined_coefficients_match(eq, x):
    problem = SingleODEProblem(eq, f(x), x)
    solver = NthLinearConstantCoeffUndeterminedCoefficients(problem)
    return solver. _undetermined_coefficients_match(eq, x)
```
Then the rest of the testing code would not need to be changed.

sympy/solvers/ode/single.py
```diff
@@ -2482,12 +2482,17 @@ def _undetermined_coefficients_match(self,expr, x, func=None, eq_homogeneous=S.Z
...
-        >>> from sympy import log, exp
-        >>> from sympy.solvers.ode.ode import _undetermined_coefficients_match
+        >>> from sympy import log, exp, Function
+        >>> from sympy.solvers.ode.single import NthLinearConstantCoeffUndeterminedCoefficients,SingleODEProblem
         >>> from sympy.abc import x
-        >>> _undetermined_coefficients_match(9*x*exp(x) + exp(-x), x)
+        >>> f = Function('f')
+        >>> eq = 9*x*exp(x) + exp(-x)
+        >>> obj = NthLinearConstantCoeffUndeterminedCoefficients(SingleODEProblem(eq,f(x),x))
+        >>> obj._undetermined_coefficients_match(eq,x)
```
oscarbenjamin, line 2491: There should be a space after a comma (here and everywhere else)

sympy/solvers/ode/single.py
```diff
@@ -2482,12 +2482,17 @@ def _undetermined_coefficients_match(self,expr, x, func=None, eq_homogeneous=S.Z
         Examples
         ========
 
-        >>> from sympy import log, exp
-        >>> from sympy.solvers.ode.ode import _undetermined_coefficients_match
+        >>> from sympy import log, exp, Function
+        >>> from sympy.solvers.ode.single import NthLinearConstantCoeffUndeterminedCoefficients,SingleODEProblem
```
oscarbenjamin, line 2486: This line is too long

sympy/solvers/ode/single.py
```diff
@@ -2482,12 +2482,17 @@ def _undetermined_coefficients_match(self,expr, x, func=None, eq_homogeneous=S.Z
...
+        >>> f = Function('f')
+        >>> eq = 9*x*exp(x) + exp(-x)
+        >>> obj = NthLinearConstantCoeffUndeterminedCoefficients(SingleODEProblem(eq,f(x),x))
+        >>> obj._undetermined_coefficients_match(eq,x)
         {'test': True, 'trialset': {x*exp(x), exp(-x), exp(x)}}
-        >>> _undetermined_coefficients_match(log(x), x)
+        >>> eq = log(x)
+        >>> obj = NthLinearConstantCoeffUndeterminedCoefficients(SingleODEProblem(eq,f(x),x))
+        >>> obj._undetermined_coefficients_match(eq,x)
         {'test': False}
```
oscarbenjamin, line 2496: It's probably better to give an example using `dsolve` here.

sympy/solvers/ode/single.py
```diff
@@ -2207,7 +2207,7 @@ def _get_simplified_sol(self, sol, collectterms):
         for i, reroot, imroot in collectterms:
             sol = collect(sol, x**i*exp(reroot*x))
         sol = powsimp(sol)
-        return [Eq(f(x), sol)]
+        return Eq(f(x), sol)
```
oscarbenjamin, line 2210: Returning anything other than a list should fail here. Is there some code somewhere else that checks for this? Why didn't this fail before?

sympy/solvers/ode/single.py
```diff
@@ -2695,7 +2700,381 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
...
+                sin_form = ln(x)**i*x**reroot*sin(abs(imroot)*ln(x))
+                if sin_form in gensols:
+                    cos_form = ln(x)**i*x**reroot*cos(imroot*ln(x))
+                    gensols.append(cos_form)
+                else:
+                    gensols.append(sin_form)
+        return gensols
+
+
+    def _get_general_solution(self, *, simplify_flag: bool = True):
```
oscarbenjamin, line 2884: Only one blank line here

sympy/solvers/ode/single.py
```diff
@@ -2695,7 +2700,381 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
...
+        self._get_sols(self.r)
+        return [self.gsol]
```
oscarbenjamin, line 2886: Shouldn't this just be `return self._get_sols()`?

It would be better if the methods in these classes don't use internal variables like this. We should minimise instance attributes.

Why is `_get_sols` a different method from `_get_general_solution`?

sympy/solvers/ode/single.py
```diff
@@ -2695,7 +2700,381 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
...
+
+    def _get_sols(self,r):
+        x = self.ode_problem.sym
+        f = self.ode_problem.func.func
+
+        # First, set up characteristic equation.
+        chareq, symbol = S.Zero, Dummy('x')
+
+        for i in r.keys():
+            if not isinstance(i, str) and i >= 0:
```
oscarbenjamin, line 2832: If `r` is a dict then it isn't necessary to call `keys` here (the dict is already iterable).

Why would `i` be a `str`?

oscarbenjamin:
Side note: these Euler solvers take up 400 lines that would be reduced to almost nothing if the ODE was simply solved by substitution.

sympy/solvers/ode/single.py
Mohitbalwani26: `r` can have key named 'trialset' which will be used for undetermined coefficient.

sympy/solvers/ode/single.py
Mohitbalwani26: `_get_sols() ` is the method which calcutates the roots of characteristic equation. And to form that we need match_object.
`self.r`  was kept as an argument because this method is also used by euler variation of parameters and euler undetermined coeffs so we need to pass their match object to this.

sympy/solvers/ode/single.py
oscarbenjamin: It would be better to store trialset as a separate attribute rather than mix it up with the roots:
```
self.roots = ...
self.trialset = ...
```

sympy/solvers/ode/single.py
oscarbenjamin: Rather than passing the match object it would be better if this had names arguments like `roots, trialset` or something.

sympy/solvers/ode/single.py
Mohitbalwani26: This docstring is just for getting trial set method. For examples of undet_coeff solver the docstring at class level has examples with dsolve.

sympy/solvers/ode/single.py
Mohitbalwani26: Currently `test_single.py` handles both type of solutions with list as well as without as some of the solvers are yet to be refactored. After refactoring I will remove the conditional part from test suite so that it should expect list of solution from every  solver.

sympy/solvers/ode/single.py
Mohitbalwani26: > Why can't the vop and undet classes just use the euler class `get_general_solution` method?


because `get_general_solution` will return a equation where as in vop and undet we just want the list of roots of characterstic equation.

sympy/solvers/ode/single.py
```diff
@@ -570,7 +570,7 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
         m, n = self.wilds_match()
         fx = self.ode_problem.func
         x = self.ode_problem.sym
-        (C1,) = self.ode_problem.get_numbered_constants(num=1)
+        (C1, ) = self.ode_problem.get_numbered_constants(num=1)
```
oscarbenjamin, line 573: This space should not be here

sympy/solvers/ode/single.py
```diff
@@ -590,7 +590,7 @@ class FirstLinear(SinglePatternODESolver):
     .. math:: dy/dx + P(x) y = Q(x)\text{.}
 
     These kinds of differential equations can be solved in a general way.  The
-    integrating factor `e^{\int P(x) \,dx}` will turn the equation into a
+    integrating factor `e^{\int P(x) \, dx}` will turn the equation into a
```
oscarbenjamin, line 593: The `\,` in latex is different from other commas since it represents a small space. I don't think that we need to also add a space after this.

sympy/solvers/ode/single.py
```diff
@@ -1702,7 +1702,7 @@ class LinearCoefficients(HomogeneousCoeffBest):
     The general form of a differential equation with linear coefficients is
 
     .. math:: y' + F\left(\!\frac{a_1 x + b_1 y + c_1}{a_2 x + b_2 y +
-                c_2}\!\right) = 0\text{,}
+                c_2}\!\right) = 0\text{, }
```
oscarbenjamin, line 1705: This space should not be here

oscarbenjamin:
It looks like you've done a find and replace to add spaces after commas but some of these places shouldn't be changed so can you check through those?

Mohitbalwani26:
> It looks like you've done a find and replace to add spaces after commas but some of these places shouldn't be changed so can you check through those?

Yes I did that. I will check once. Also I am trying to move all the reusable functions to another file so that classes look minimal and kind of wrapper for them.

Mohitbalwani26:
@oscarbenjamin does this way seem good? 
Also if yes then what should be the name of `euler_nonhomogen_helpers.py`?

oscarbenjamin:
> @oscarbenjamin does this way seem good?
> Also if yes then what should be the name of `euler_nonhomogen_helpers.py`?

Yes, I think so. I think it should just be called `nonhomogeneous.py`. Other functions like variation of parameters etc can go there later.

Mohitbalwani26:
> Yes, I think so. I think it should just be called `nonhomogeneous.py`. Other functions like variation of parameters etc can go there later.

So should I again move functions like `solve_variation_of_param` , `solve_unde_coeff` back to classes? 

Currently all the functions which are there in new file they all are used by more than 1 solver. If we again keep any of them into solver, we will have to create instance of that class into another. 

So what all functions should I keep in classes back?

sympy/solvers/ode/nonhomogeneous.py
```diff
@@ -0,0 +1,462 @@
...
+        else:
+            sin_form = ln(x)**i*x**reroot*sin(abs(imroot)*ln(x))
+            if sin_form in gensols:
+                cos_form = ln(x)**i*x**reroot*cos(imroot*ln(x))
+                gensols.append(cos_form)
+            else:
+                gensols.append(sin_form)
+    return gsol, gensols
+
+def _solve_variation_of_parameters(eq, func, roots, homogen_sol, order, match_obj, simplify_flag=True):
```
oscarbenjamin, line 107: Two blank lines

sympy/solvers/ode/nonhomogeneous.py
```diff
@@ -0,0 +1,462 @@
...
+    for i in roots:
+        psol += negoneterm*Integral(wronskian([sol for sol in roots if sol != i], x)*r[-1]/wr, x)*i/r[order]
+        negoneterm *= -1
+
+    if simplify_flag:
+        psol = simplify(psol)
+        psol = trigsimp(psol, deep=True)
+    return Eq(f(x), homogen_sol.rhs + psol)
+
+def _get_const_characterstic_eq_sols(r, func, order):
```
oscarbenjamin, line 157: characteristic

oscarbenjamin:
> So should I again move functions like `solve_variation_of_param` , `solve_unde_coeff` back to classes?

No I think this is fine. Eventually these functions should be made into something that it is more usefully reusable though.

sympy/solvers/ode/nonhomogeneous.py
```diff
@@ -0,0 +1,463 @@
+from collections import defaultdict
```
oscarbenjamin, line 1: There should be some kind of comment or module level docstring at the top of the module explaining what it is.

sympy/solvers/ode/nonhomogeneous.py
```diff
@@ -0,0 +1,463 @@
...
+            return False
+        return x**order in coeff.args
+    elif coeff.is_Pow:
+        return coeff.as_base_exp() == (x, order)
+    elif order == 1:
+        return x == coeff
+    return False
+
+
+def _get_euler_characterstic_eq_sols(eq, func, match_obj):
```
oscarbenjamin, line 47: The functions in this file should have at least a one line docstring explaining what they are as well as comments explaining why they are here and what they are used for.

sympy/solvers/ode/nonhomogeneous.py
```diff
@@ -0,0 +1,512 @@
...
+        return coeff.as_base_exp() == (x, order)
+    elif order == 1:
+        return x == coeff
+    return False
+
+
+def _get_euler_characteristic_eq_sols(eq, func, match_obj):
+    r"""
+    It returns the solution of homogeneous part of the linear euler ODE and
+    the list of roots of characteristic equation.
```
oscarbenjamin, line 84: Don't use "It" here. Just "Returns ..." is fine.

This doesn't really explain the function very well. What is the `match_obj` parameter supposed to be for example?

sympy/solvers/ode/nonhomogeneous.py
```diff
@@ -0,0 +1,512 @@
...
+nth_linear_constant_coeff_variation_of_parameters,
+and nth_linear_euler_eq_nonhomogeneous_variation_of_parameters.
+
+All the functions in this file are used by more than one solvers so, instead of creating
+instances in other classes for using them it is better to keep it here as separate helpers.
+
+Functions that are for internal use:
+
+1) _test_term(coeff, func, order) - It takes coefficient of partiular term, function and
+    order to which it is associated. It returns whether given term matches with linear euler ODEs.
```
oscarbenjamin, line 13: I don't think that you need to detail each function at the top here. These kinds of comments/description are better placed in or immediately before the function.

This module level docstring just needs to explain the general purpose of the module.

sympy/solvers/ode/single.py
```diff
@@ -1030,7 +1029,7 @@ class SecondNonlinearAutonomousConserved(SinglePatternODESolver):
 
     def _wilds(self, f, x, order):
         fy = Wild('fy', exclude=[0, f(x).diff(x), f(x).diff(x, 2)])
-        return (fy,)
+        return (fy, )
```
oscarbenjamin, line 1032: There shouldn't be a space here

oscarbenjamin:
Okay, looks good.

sympy/solvers/ode/single.py
Mohitbalwani26: Okay, will remove it in next PR.

PR #21695, opened 2021-07-02
New Documentation layout

nanjekyejoannah:
This commit reorganizes the documentation following the diátaxis framework.

Main goal:

The main focus is to ensure files are moved to the right folders.

Not a goal:

Theming and navigation bar.

I will have small follow-up PRs for each section and theme/navigation.

NO ENTRY


namannimmo10:
Please write "NO ENTRY" in the release notes:
```
<!-- BEGIN RELEASE NOTES -->
NO ENTRY
<!-- END RELEASE NOTES -->
```

asmeurer:
I'm worried that if we move the pages around that all the old URLs pointing to the docs around the internet will be broken. We should see if we can generate redirects so that we can make them continue to work. I found two packages that can do this

https://documatt.gitlab.io/sphinx-reredirects/usage.html
https://github.com/sphinx-contrib/redirects

We should see if either of these will work.

asmeurer:
There are a few references to rst files in sympy/testing/runtests.py which need to be updated.

asmeurer:
@namannimmo10 substantial changes should include a release notes entry. Please do not tell people to use `NO ENTRY` unless the change is internal only or trivial. See https://github.com/sympy/sympy/wiki/Writing-Release-Notes. If you want, you can also edit the PR description to add an entry for someone. 

For this PR, the plan is to close it and redo it in smaller chunks, so you do not need to worry about it for this one.

nanjekyejoannah:
Closing for a smaller change as discussed with @asmeurer

namannimmo10:
Thank you! I apologize for the noise.

PR #21698, opened 2021-07-02
Fix the docs build when gmpy2 is installed

### LINKED ISSUE #21697: Error when building docs: py:class reference target not found: mpz
I get this error when building the docs

```
<unknown>:1: WARNING: py:class reference target not found: mpz
<unknown>:1: WARNING: py:class reference target not found: mpz
<unknown>:1: WARNING: py:class reference target not found: mpq
<unknown>:1: WARNING: py:class reference target not found: mpq
```

This is with the latest version of Sphinx. I have gmpy2 installed.

asmeurer:
Previously the Sphinx would try to automatically generate a cross-reference to
gmpy2.mpz and gmpy2.mpq because of the dtype and tp attributes of the
GMPYIntegerRing and GMPYRationalField classes, which fails because we don't
have references to those classes in our docs. We don't really need to document
these, so we exclude them.

Ideally, we would be installing gmpy2 on CI when building the docs, which
would have caught this issue sooner, but this does not address that.

Fixes #21697.

#### Release Notes

NO ENTRY

### DIFF
--- a/doc/src/modules/polys/domainsref.rst
+++ b/doc/src/modules/polys/domainsref.rst
@@ -131,6 +131,7 @@ when available.
 .. autoclass:: PythonIntegerRing
 .. autoclass:: GMPYIntegerRing
    :members:
+   :exclude-members: dtype, tp
 
 .. _QQ:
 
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
> Ideally, we would be installing gmpy2 on CI when building the docs

For a longer term fix we can just delete the GMPYIntegerRing and GMPYRationalField classes now. They aren't used any more.

Issue #12134, opened 2017-02-07
Sympy numbers don't adhere to Python's numbers ABC


inducer:
I'd expect this to be `True`:

```
>>> import sympy
>>> x = sympy.numbers.Float(5)
>>> import numbers
>>> isinstance(x, numbers.Number)
False
```

https://docs.python.org/3/library/numbers.html

(Sorry if this has been considered before--a quick search didn't reveal anything.)

asmeurer:
Should be easy to fix. I think the last time this was brought up we still supported Python versions that didn't have ABCs.

Abdullahjavednesar:
Why does it return ```False```?

madan96:
@Abdullahjavednesar I guess because `x` is not an instance of `Number` ABC in Python numbers.

madan96:
@asmeurer Will inheriting the `Number` class (from numbers module) into `Number` class in sympy.core.numbers work?

Abdullahjavednesar:
@madan96 Hopefully inheriting ```Number```class should work.

asmeurer:
We might need to define some attributes required by the ABCs, but as far as I know it should work.

meet7202:
```
>>> import sympy
>>> x = sympy.numbers.Float(5)
>>> isinstance(x, numbers.Number)
True
```
importing numbers changes datatype of x to string from Number that's why Number ABC returns False.
@asmeurer , This issue can be solved if I will change 
`num = str(num)`  to `num = mpf(num)` in "sympy/core/numbers.py" (Line Number 780-781) , But I am not able to figure our that it will misbehave in other cases or not please clear my doubt so I can send you PR.

mohit3011:
@meet7202  are you working on this issue ?

meet7202:
@mohit3011 , Yes I am. Just Needed @asmeurer's confirmation...!!

madan96:
@meet7202 You can run `tests` locally to check if there is any anomaly for other cases.

asmeurer:
The fix here should to add `numbers.Number` as a superclass to `Float`.

meet7202:
@asmeurer sir, Please review PR #12152

meet7202:
```
>>> import sympy
>>> x = sympy.numbers.Number(5)
>>> import numbers
>>> isinstance(x, numbers.Number)
False
```
`sympy.numbers.Number()` also gives WA and If I am not wrong Float is inherited from `sympy.numbers.Number()`.

[...]

PR #21699, opened 2021-07-03
feat(core): make sympy.Integer compatible with Integral ABC

### LINKED ISSUE #19311: "sympy.core.numbers.Integer" should be registered with "numbers.Integral"
While recently perusing the `sympy.core.numbers` submodule, I espied with my rheumy little eye a trivial typo with respect to type registration. SymPy's `Integer` type is [currently registered as satisfying](https://github.com/sympy/sympy/blob/5d1b2c6f04f583ec0beeea4b8238d4151b497332/sympy/core/numbers.py#L3941) the `numbers.Rational` API ~~(which it obviously doesn't)~~ (which is technically true but less useful), but should instead be registered as satisfying the `numbers.Integral` API ~~(which it obviously does)~~ (which is also true and much more useful): e.g.,

```
    # This less useful line...
    numbers.Rational.register(Integer)

    # ...should instead resemble this more useful line.
    numbers.Integral.register(Integer)
```

Thanks for all the symbolic maths, beloved SymPy volunteers! :yellow_heart:

posita:
Make `sympy.Integer` compatible with the `numbers.Integral` ABC as defined
in PEPs 3141 and 238. (See also PEP 239.) The necessary bitwise operations
(e.g. `1 << 2`) are added to `sympy.Integer` and `sympy.Integer` is
registered with the `Integral` ABC such that:

    isinstance(sympify(1), Integral) -> True

Prior to this commit, `sympy.Integer.__truediv__` and
`sympy.Integer.__rtruediv__` already returned `sympy.Rational` values,
which is not only ergonomic, but also appears compliant.

Fix sympy/sympy#19311.

* core
  * Add missing bitwise operator implementations to sympy.core.numbers.Integer and register with numbers.Integral

### DIFF
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -2422,6 +2422,76 @@ class Integer(Rational):
     def __rfloordiv__(self, other):
         return Integer(Integer(other).p // self.p)
 
+    # These bitwise operations (__lshift__, __rlshift__, ..., __invert__) are defined
+    # for Integer only and not for general sympy expressions. This is to achieve
+    # compatibility with the numbers.Integral ABC which only defines these operations
+    # among instances of numbers.Integral. Therefore, these methods check explicitly for
+    # integer types rather than using sympify because they should not accept arbitrary
+    # symbolic expressions and there is no symbolic analogue of numbers.Integral's
+    # bitwise operations.
+    def __lshift__(self, other):
+        if isinstance(other, (int, Integer, numbers.Integral)):
+            return Integer(self.p << int(other))
+        else:
+            return NotImplemented
+
+    def __rlshift__(self, other):
+        if isinstance(other, (int, numbers.Integral)):
+            return Integer(int(other) << self.p)
+        else:
+            return NotImplemented
+
+    def __rshift__(self, other):
+        if isinstance(other, (int, Integer, numbers.Integral)):
+            return Integer(self.p >> int(other))
+        else:
+            return NotImplemented
+
+    def __rrshift__(self, other):
+        if isinstance(other, (int, numbers.Integral)):
+            return Integer(int(other) >> self.p)
+        else:
+            return NotImplemented
+
+    def __and__(self, other):
+        if isinstance(other, (int, Integer, numbers.Integral)):
+            return Integer(self.p & int(other))
+        else:
+            return NotImplemented
+
+    def __rand__(self, other):
+        if isinstance(other, (int, numbers.Integral)):
+            return Integer(int(other) & self.p)
+        else:
+            return NotImplemented
+
+    def __xor__(self, other):
+        if isinstance(other, (int, Integer, numbers.Integral)):
+            return Integer(self.p ^ int(other))
+        else:
+            return NotImplemented
+
+    def __rxor__(self, other):
+        if isinstance(other, (int, numbers.Integral)):
+            return Integer(int(other) ^ self.p)
+        else:
+            return NotImplemented
+
+    def __or__(self, other):
+        if isinstance(other, (int, Integer, numbers.Integral)):
+            return Integer(self.p | int(other))
+        else:
+            return NotImplemented
+
+    def __ror__(self, other):
+        if isinstance(other, (int, numbers.Integral)):
+            return Integer(int(other) | self.p)
+        else:
+            return NotImplemented
+
+    def __invert__(self):
+        return Integer(~self.p)
+
 # Add sympify converters
 converter[int] = Integer
 
@@ -4054,6 +4124,6 @@ def _register_classes():
     numbers.Number.register(Number)
     numbers.Real.register(Float)
     numbers.Rational.register(Rational)
-    numbers.Rational.register(Integer)
+    numbers.Integral.register(Integer)
 
 _register_classes()
--- a/sympy/core/tests/test_numbers.py
+++ b/sympy/core/tests/test_numbers.py
@@ -1226,6 +1226,127 @@ def test_powers_Float():
     assert str((S('-1/10')**S('3/10')).n()) == str(Float(-.1)**(.3))
 
 
+def test_lshift_Integer():
+    assert Integer(0) << Integer(2) == Integer(0)
+    assert Integer(0) << 2 == Integer(0)
+    assert 0 << Integer(2) == Integer(0)
+
+    assert Integer(0b11) << Integer(0) == Integer(0b11)
+    assert Integer(0b11) << 0 == Integer(0b11)
+    assert 0b11 << Integer(0) == Integer(0b11)
+
+    assert Integer(0b11) << Integer(2) == Integer(0b11 << 2)
+    assert Integer(0b11) << 2 == Integer(0b11 << 2)
+    assert 0b11 << Integer(2) == Integer(0b11 << 2)
+
+    assert Integer(-0b11) << Integer(2) == Integer(-0b11 << 2)
+    assert Integer(-0b11) << 2 == Integer(-0b11 << 2)
+    assert -0b11 << Integer(2) == Integer(-0b11 << 2)
+
+    raises(TypeError, lambda: Integer(2) << 0.0)
+    raises(TypeError, lambda: 0.0 << Integer(2))
+    raises(ValueError, lambda: Integer(1) << Integer(-1))
+
+
+def test_rshift_Integer():
+    assert Integer(0) >> Integer(2) == Integer(0)
+    assert Integer(0) >> 2 == Integer(0)
+    assert 0 >> Integer(2) == Integer(0)
+
+    assert Integer(0b11) >> Integer(0) == Integer(0b11)
+    assert Integer(0b11) >> 0 == Integer(0b11)
+    assert 0b11 >> Integer(0) == Integer(0b11)
+
+    assert Integer(0b11) >> Integer(2) == Integer(0)
+    assert Integer(0b11) >> 2 == Integer(0)
+    assert 0b11 >> Integer(2) == Integer(0)
+
+    assert Integer(-0b11) >> Integer(2) == Integer(-1)
+    assert Integer(-0b11) >> 2 == Integer(-1)
+    assert -0b11 >> Integer(2) == Integer(-1)
+
+    assert Integer(0b1100) >> Integer(2) == Integer(0b1100 >> 2)
+    assert Integer(0b1100) >> 2 == Integer(0b1100 >> 2)
+    assert 0b1100 >> Integer(2) == Integer(0b1100 >> 2)
+
+    assert Integer(-0b1100) >> Integer(2) == Integer(-0b1100 >> 2)
+    assert Integer(-0b1100) >> 2 == Integer(-0b1100 >> 2)
+    assert -0b1100 >> Integer(2) == Integer(-0b1100 >> 2)
+
+    raises(TypeError, lambda: Integer(0b10) >> 0.0)
+    raises(TypeError, lambda: 0.0 >> Integer(2))
+    raises(ValueError, lambda: Integer(1) >> Integer(-1))
+
+
+def test_and_Integer():
+    assert Integer(0b01010101) & Integer(0b10101010) == Integer(0)
+    assert Integer(0b01010101) & 0b10101010 == Integer(0)
+    assert 0b01010101 & Integer(0b10101010) == Integer(0)
+
+    assert Integer(0b01010101) & Integer(0b11011011) == Integer(0b01010001)
+    assert Integer(0b01010101) & 0b11011011 == Integer(0b01010001)
+    assert 0b01010101 & Integer(0b11011011) == Integer(0b01010001)
+
+    assert -Integer(0b01010101) & Integer(0b11011011) == Integer(-0b01010101 & 0b11011011)
+    assert Integer(-0b01010101) & 0b11011011 == Integer(-0b01010101 & 0b11011011)
+    assert -0b01010101 & Integer(0b11011011) == Integer(-0b01010101 & 0b11011011)
+
+    assert Integer(0b01010101) & -Integer(0b11011011) == Integer(0b01010101 & -0b11011011)
+    assert Integer(0b01010101) & -0b11011011 == Integer(0b01010101 & -0b11011011)
+    assert 0b01010101 & Integer(-0b11011011) == Integer(0b01010101 & -0b11011011)
+
+    raises(TypeError, lambda: Integer(2) & 0.0)
+    raises(TypeError, lambda: 0.0 & Integer(2))
+
+
+def test_xor_Integer():
+    assert Integer(0b01010101) ^ Integer(0b11111111) == Integer(0b10101010)
+    assert Integer(0b01010101) ^ 0b11111111 == Integer(0b10101010)
+    assert 0b01010101 ^ Integer(0b11111111) == Integer(0b10101010)
+
+    assert Integer(0b01010101) ^ Integer(0b11011011) == Integer(0b10001110)
+    assert Integer(0b01010101) ^ 0b11011011 == Integer(0b10001110)
+    assert 0b01010101 ^ Integer(0b11011011) == Integer(0b10001110)
+
+    assert -Integer(0b01010101) ^ Integer(0b11011011) == Integer(-0b01010101 ^ 0b11011011)
+    assert Integer(-0b01010101) ^ 0b11011011 == Integer(-0b01010101 ^ 0b11011011)
+    assert -0b01010101 ^ Integer(0b11011011) == Integer(-0b01010101 ^ 0b11011011)
+
+    assert Integer(0b01010101) ^ -Integer(0b11011011) == Integer(0b01010101 ^ -0b11011011)
+    assert Integer(0b01010101) ^ -0b11011011 == Integer(0b01010101 ^ -0b11011011)
+    assert 0b01010101 ^ Integer(-0b11011011) == Integer(0b01010101 ^ -0b11011011)
+
+    raises(TypeError, lambda: Integer(2) ^ 0.0)
+    raises(TypeError, lambda: 0.0 ^ Integer(2))
+
+
+def test_or_Integer():
+    assert Integer(0b01010101) | Integer(0b10101010) == Integer(0b11111111)
+    assert Integer(0b01010101) | 0b10101010 == Integer(0b11111111)
+    assert 0b01010101 | Integer(0b10101010) == Integer(0b11111111)
+
+    assert Integer(0b01010101) | Integer(0b11011011) == Integer(0b11011111)
+    assert Integer(0b01010101) | 0b11011011 == Integer(0b11011111)
+    assert 0b01010101 | Integer(0b11011011) == Integer(0b11011111)
+
+    assert -Integer(0b01010101) | Integer(0b11011011) == Integer(-0b01010101 | 0b11011011)
+    assert Integer(-0b01010101) | 0b11011011 == Integer(-0b01010101 | 0b11011011)
+    assert -0b01010101 | Integer(0b11011011) == Integer(-0b01010101 | 0b11011011)
+
+    assert Integer(0b01010101) | -Integer(0b11011011) == Integer(0b01010101 | -0b11011011)
+    assert Integer(0b01010101) | -0b11011011 == Integer(0b01010101 | -0b11011011)
+    assert 0b01010101 | Integer(-0b11011011) == Integer(0b01010101 | -0b11011011)
+
+    raises(TypeError, lambda: Integer(2) | 0.0)
+    raises(TypeError, lambda: 0.0 | Integer(2))
+
+
+def test_invert_Integer():
+    assert ~Integer(0b01010101) == Integer(-0b01010110)
+    assert ~Integer(0b01010101) == Integer(~0b01010101)
+    assert ~(~Integer(0b01010101)) == Integer(0b01010101)
+
+
 def test_abs1():
     assert Rational(1, 6) != Rational(-1, 6)
     assert abs(Rational(1, 6)) == abs(Rational(-1, 6))
@@ -2037,6 +2158,10 @@ def test_abc():
     assert(isinstance(y, nums.Rational))
     z = numbers.Integer(3)
     assert(isinstance(z, nums.Number))
+    assert(isinstance(z, numbers.Number))
+    assert(isinstance(z, nums.Rational))
+    assert(isinstance(z, numbers.Rational))
+    assert(isinstance(z, nums.Integral))
 
 def test_floordiv():
     assert S(2)//S.Half == 4


oscarbenjamin:
As discussed in #19311 we shouldn't register with the ABC unless the methods of the ABC are actually supported:
```
In [75]: S.One << 3
<ipython-input-75-d6d0393686a2> in <module>
----> 1 S.One << 3

TypeError: unsupported operand type(s) for <<: 'One' and 'int'
```

posita:
> As discussed in #19311 we shouldn't register with the ABC unless the methods of the ABC are actually supported:
> 
> ```
> In [75]: S.One << 3
> ---------------------------------------------------------------------------
> TypeError                                 Traceback (most recent call last)
> <ipython-input-75-d6d0393686a2> in <module>
> ----> 1 S.One << 3
> 
> TypeError: unsupported operand type(s) for <<: 'One' and 'int'
> ```

I guess I don't understand the rational of making that a requirement in light of the fact that existing registrations with the ABC were already kind of lies. Wouldn't this at least get closer to the intended truth? Or is it that the current registration is now technically correct (in light of https://github.com/sympy/sympy/issues/19311#issuecomment-873376426), and all that remains for registering `sympy.core.numbers.Integer` with `numbers.Integral` are the bitwise functions?

Is supporting bitwise functions even feasible in light of https://github.com/sympy/sympy/issues/19311#issuecomment-628419934 ? Forgive me if I don't appreciate all the implications how existing bitwise operators (e.g., `Implies`) might overlap/fight.

oscarbenjamin:
> I guess I don't understand the rational of making that a requirement in light of the fact that existing registrations with the ABC were already kind of lies.

Don't use one bug to justify introducing another!

I don't understand why anyone cares about using these ABCs if they don't also care that registration with an ABC actually means something.

The rest of the discussion should be at #19311 rather than here.

oscarbenjamin:
> The rest of the discussion should be at #19311 rather than here.

I've commented on the issue

posita:
> > I guess I don't understand the rational of making that a requirement in light of the fact that existing registrations with the ABC were already kind of lies.
> 
> Don't use one bug to justify introducing another!

Perhaps that was the source of my misunderstanding. I didn't think I was introducing another bug. I thought I was merely not completely fixing an existing one.

> I don't understand why anyone cares about using these ABCs if they don't also care that registration with an ABC actually means something.
> 
> The rest of the discussion should be at #19311 rather than here.

Fair. I have attempted to update the PR accordingly and in light of https://github.com/sympy/sympy/issues/19311#issuecomment-873399610.

sympy/core/numbers.py
```diff
@@ -2422,6 +2422,89 @@ def __floordiv__(self, other):
     def __rfloordiv__(self, other):
         return Integer(Integer(other).p // self.p)
 
+    def __lshift__(self, other):
+        if global_parameters.evaluate:
+            if isinstance(other, int):
+                return Integer(self.p << other)
+            elif isinstance(other, Integer):
+                return Integer(self.p << other.p)
+        return NotImplemented
```
oscarbenjamin, line 2431: I don't think we need to check evaluate here since there is no unevaluated form. This would raise TypeError when evaluate=False which isn't particularly useful and probably contradicts the Integral ABC anyway.

I'd write this as:
```python
if not isinstance(other, (int, Integer)):
    return NotImplemented
return Integer(self.p << int(other))
```
I'm not sure what the numbers ABC say about interaction between different implementations of each ABC. Should this method work for any `numbers.Integral`? If so then maybe it should be:
```python
if not isinstance(other, (int, Integer, Integral)):
    return NotImplemented
return Integer(self.p << int(other))
```

sympy/core/numbers.py
```diff
@@ -2422,6 +2422,89 @@ def __floordiv__(self, other):
...
+
+    def __rlshift__(self, other):
+        if global_parameters.evaluate:
+            if isinstance(other, int):
+                return Integer(other << self.p)
+            elif isinstance(other, Integer):
+                return Integer(other.p << self.p)
```
oscarbenjamin, line 2438: Have you checked coverage of these methods? I think this is unreachable.

sympy/core/numbers.py
posita: Oh! I like that much better. I took a stab with a slight twist on style. (I've developed a habitual nervousness around `if not …` because it tends to mess with my brain, even though it's perfectly readable in this case.) If you prefer the flow as you've written it, let me know and I'll change it. I also like including `Integral`, so I've done that as well.

Technically, one could probably get away with `isinstance(other, (Integer, numbers.Integral))`, since `isinstance(int(0), numbers.Integral)` should evaluate to `True`.

sympy/core/numbers.py
posita: Hmmm…now that you mention it, I _**think**_ all we need is something like this:

```python
    def __lshift__(self, other):
        if isinstance(other, numbers.Integral):
            return Integer(self.p << int(other))
        else:
            return NotImplemented

    def __rlshift__(self, other):
        if isinstance(other, numbers.Integral):
            return Integer(int(other) << self.p)
        else:
            return NotImplemented
```

Both `int` and `Integer` should be registered with `numbers.Integral`, so the above should be sufficient. Its symmetry is a _**tiny**_ bit misleading, since, as you point out, the `__r…__` methods will never be called where `other` is an `Integer`. If that hinting was important, this is redundant, but may be helpful in cultivating that understanding:

```python
    def __lshift__(self, other):
        if isinstance(other, (numbers.Integral, Integer)):
            return Integer(self.p << int(other))
        else:
            return NotImplemented

    def __rlshift__(self, other):
        if isinstance(other, numbers.Integral):
            return Integer(int(other) << self.p)
        else:
            return NotImplemented
```

sympy/core/numbers.py
oscarbenjamin: > Technically, one could probably get away with `isinstance(other, (Integer, numbers.Integral))`, since `isinstance(int(0), numbers.Integral)` should evaluate to `True`.

Wouldn't the changes here also mean that `isinstance(Integer(0), numbers.Integral` would be True as well?

I doubt performance is an issue here (anyone looking for better performance should just use `int` rather than `Integer`) but testing `isinstance` with an ABC is slow so it's faster if you can shortcut the common cases:
```python
In [119]: from numbers import Integral

In [120]: x = 1

In [121]: %timeit isinstance(x, int)
218 ns ± 0.917 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [122]: %timeit isinstance(x, Integral)
1.64 µs ± 74 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [123]: %timeit isinstance(x, (int, Integral))
269 ns ± 36.8 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

posita: Interesting. I could make one slight tweak to remove `Integer` from the `isinstance` checks in the `__r…__` methods (to signal that `other` won't be an `Integer` on that path), but I think its presence is logically harmless.

This is an example of what I mean:

```diff
diff --git a/sympy/core/numbers.py b/sympy/core/numbers.py
index af20ef64e3..73f828de4e 100644
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -2429,7 +2429,7 @@ def __lshift__(self, other):
             return NotImplemented

     def __rlshift__(self, other):
-        if isinstance(other, (int, Integer, numbers.Integral)):
+        if isinstance(other, (int, numbers.Integral)):
             return Integer(int(other) << self.p)
         else:
             return NotImplemented
```

Just let me know what you'd prefer.

oscarbenjamin: My preference would be to test for Integer explicitly so that it is very clear that this works with Integer but honestly I don't think it matters much so I don't mind.

A more important confusion that someone would likely have when looking at this code is to wonder why `sympify` isn't being used. There should be a long comment above these methods saying something like:

Bitwise operations are defined for Integer only and not for general sympy expressions. This is for compatibility with the numbers.Integer ABC which only defines these operations between instances of numbers.Integral. These methods should check explicitly for integer types rather than using sympify because they should not accept arbitrary symbolic expressions and there is no symbolic analogue of the bitwise operations.

sympy/core/numbers.py
```diff
@@ -2422,6 +2422,69 @@ def __floordiv__(self, other):
...
         return Integer(Integer(other).p // self.p)
 
+    def __lshift__(self, other):
+        if isinstance(other, (int, Integer, numbers.Integral)):
+            return Integer(self.p << int(other))
+        else:
+            return NotImplemented
+
+    def __rlshift__(self, other):
+        if isinstance(other, (int, numbers.Integral)):
```
posita, line 2432: Let me know if you think it's necessary/helpful to add `Integer` to this list. In the current implementation, I don't think it's necessary, and its omission signals the asymmetry of the `__r…__` methods. That being said, maybe there's some weird future where there's a `class Integer2(Integer): …` that overrides half of these methods such that these `__r…__` methods become "slow" again? Dunno.

oscarbenjamin:
Are the other methods of `Integer` compatible with `Integral`?

An obvious difference would be:
```python
In [128]: Integer(1) / Integer(2)
Out[128]: 1/2

In [129]: int(1) / int(2)
Out[129]: 0.5
```
All methods should be checked carefully

posita:
> Are the other methods of `Integer` compatible with `Integral`?
> 
> An obvious difference would be:
> 
> ```python
> In [128]: Integer(1) / Integer(2)
> Out[128]: 1/2
> 
> In [129]: int(1) / int(2)
> Out[129]: 0.5
> ```
> 
> All methods should be checked carefully

Ah, yes. `__truediv__` is its own animal. There is precious little guidance on definite semantics, there. There are [the release notes from 2.2](https://docs.python.org/3/whatsnew/2.2.html#pep-238-changing-the-division-operator), where PEP 238 was introduced. There is [PEP 238 itself](https://www.python.org/dev/peps/pep-0238/), and there are the python/typeshed definitions for [`int`](https://github.com/python/typeshed/blob/1a131a489efc86e5de9e01fcb2083d93226b4f57/stdlib/builtins.pyi#L198) and [`Number`](https://github.com/python/typeshed/blob/17dcea4a683d0892c26ba79234341e9b575bdac3/stdlib/numbers.pyi#L36) (from which `Integral` inherits).

We do have this hint from [PEP 238's "Semantics of True Division"](https://www.python.org/dev/peps/pep-0238/#semantics-of-true-division) section:

> If and when a rational type is added to Python (see [PEP 239](https://www.python.org/dev/peps/pep-0239) …), true division for ints and longs should probably return a rational. This avoids the problem with true division of ints and longs losing information. But until then, for consistency, float is the only choice for true division.

That being said, PEP 239 was [rejected](https://www.python.org/dev/peps/pep-0239/#bdfl-pronouncement). Standard library implementations are slightly inconsistent. Consider:

```python
>>> from fractions import Fraction
>>> Fraction(1) / 4
Fraction(1, 4)
>>> Fraction(1) // 4
0
>>> Fraction(1) // Fraction(4)
0
>>> from decimal import Decimal
>>> Decimal("1.0") / 4
Decimal('0.25')
>>> Decimal("1.0") // 4
Decimal('0')
>>> Decimal("1.0") // Decimal("4.0")
Decimal('0')
```

Given the ambiguities, I think returning a `Rational` (vs. a `Float`) is not only perfectly defensible, but likely preferred by SymPy users, and seems internally consistent with what's already present:

```python
>>> simplify("1 / 4")
1/4
>>> simplify("1.0 / 4")
0.250000000000000
>>> simplify("1 // 4")
0
>>> simplify("1.0 // 4")
0
```

With the exception of `__matmul__` (which is not required), spot checking [all other operations](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types) (including the oft-forgot `__index__`) seem to yield sane results.

sympy/core/numbers.py
posita: Added. I tweaked the language _slightly_, but reproduced it largely verbatim.

oscarbenjamin:
> That being said, PEP 239 was [rejected](https://www.python.org/dev/peps/pep-0239/#bdfl-pronouncement).

I think the appropriate PEP is PEP 3141 -- A Type Hierarchy for Numbers:
https://www.python.org/dev/peps/pep-3141/

>  Standard library implementations are slightly inconsistent with `__floordiv__`, but `Rational`s return other `Rational`s for `__truediv__`. Consider:

Note that Decimal is not included in the Real ABC:
https://bugs.python.org/issue43602

I guess that for true division we have `int / int -> float` which in terms of the numeric tower is `Integral / Integral -> Real`. SymPy has `Integer / Integer -> Rational` which might seem inconsistent except that any Rational is also an instance of Real (`numbers.Rational` is a subclass of `numbers.Real`). That's probably fine then...

The other methods listed for `Integral` should  be checked though.

oscarbenjamin:
I wonder whether operations like `Integer(1) << Integer(2)` should return `Integer` rather than just a plain `int`. There's really no benefit in using `Integer` here...

posita:
Thanks for the continued attention. Hopefully this is close to acceptance.

> I think the appropriate PEP is PEP 3141 -- A Type Hierarchy for Numbers:
> https://www.python.org/dev/peps/pep-3141/

PEP 3141 provides no guidance on division. IMO, `Integral / Integral` -> `Rational` is proper for anyone who cares about math. I agree it's the right way to go.

> The other methods listed for `Integral` should be checked though.

What does this mean? Do I need to write more tests? Am I waiting on you to check those methods? Are you waiting on me to provide something?

> I wonder whether operations like `Integer(1) << Integer(2)` should return `Integer` rather than just a plain `int`. There's really no benefit in using `Integer` here...

My preference is to leave such decisions to the caller rather than force them to deal with an implicit conversion. My recommendation is to live with it like this for awhile and change it in a subsequent PR if it seems important.

oscarbenjamin:
> > The other methods listed for `Integral` should be checked though.
> 
> What does this mean? Do I need to write more tests? Am I waiting on you to check those methods? Are you waiting on me to provide something?

In so far as there is a spec for what methods should exist and what they should return I would like you to check it. Where the spec is lacking at least the stdlib implementations of these classes can be compared with what sympy's Integer gives in this PR.

I ask for this because I didn't personally see any value in supporting these ABCs at all (https://github.com/sympy/sympy/issues/12134#issuecomment-472221844) but I ended up spending time fixing the bugs from the implementation to support the Rational ABC. I'm happy to merge this but please check it properly so that merging this can put the issue to rest without me needing to fill in the gaps in future.

If this is something that *you* care about then please make sure it is done properly. This has no value to sympy internally but I understand that there are downstream use cases where this is somehow beneficial although apparently no one will answer what they are: https://github.com/sympy/sympy/issues/19311#issuecomment-628517308.

posita:
I'm fine doing what is required to get this accepted, but first I need to understand what that is. I'm hoping this satisfies the request:

```python
In [26]: for binop in operator.add, operator.and_, operator.eq, operator.floordiv, operator.ge, operator.gt, operator.le, operator.lshift, operator.lt, operator.mod, operator.mul, operator.ne, operator.or_, operator.pow, operator.rshift, operator.sub, operator.truediv, operator.xor:
    ...:     a, b = S("-12"), S("5")
    ...:     r = binop(a, b)
    ...:     print(f"{binop.__name__!r}({a!r}, {b!r}) -> {r!r} ({type(r)!r})")
'add'(-12, 5) -> -7 (<class 'sympy.core.numbers.Integer'>)
'and_'(-12, 5) -> 4 (<class 'sympy.core.numbers.Integer'>)
'eq'(-12, 5) -> False (<class 'bool'>)
'floordiv'(-12, 5) -> -3 (<class 'sympy.core.numbers.Integer'>)
'ge'(-12, 5) -> False (<class 'sympy.logic.boolalg.BooleanFalse'>)
'gt'(-12, 5) -> False (<class 'sympy.logic.boolalg.BooleanFalse'>)
'le'(-12, 5) -> True (<class 'sympy.logic.boolalg.BooleanTrue'>)
'lshift'(-12, 5) -> -384 (<class 'sympy.core.numbers.Integer'>)
'lt'(-12, 5) -> True (<class 'sympy.logic.boolalg.BooleanTrue'>)
'mod'(-12, 5) -> 3 (<class 'sympy.core.numbers.Integer'>)
'mul'(-12, 5) -> -60 (<class 'sympy.core.numbers.Integer'>)
'ne'(-12, 5) -> True (<class 'bool'>)
'or_'(-12, 5) -> -11 (<class 'sympy.core.numbers.Integer'>)
'pow'(-12, 5) -> -248832 (<class 'sympy.core.numbers.Integer'>)
'rshift'(-12, 5) -> -1 (<class 'sympy.core.numbers.NegativeOne'>)
'sub'(-12, 5) -> -17 (<class 'sympy.core.numbers.Integer'>)
'truediv'(-12, 5) -> -12/5 (<class 'sympy.core.numbers.Rational'>)
'xor'(-12, 5) -> -15 (<class 'sympy.core.numbers.Integer'>)

In [27]: for unop in complex, int, float, round, math.ceil, math.floor, math.trunc, operator.abs, operator.index, operator.invert, operator.neg, operator.pos:
    ...:     a = S("-12")
    ...:     r = unop(a)
    ...:     print(f"{unop.__name__!r}({a!r}) -> {r!r} ({type(r)!r})")
'complex'(-12) -> (-12+0j) (<class 'complex'>)
'int'(-12) -> -12 (<class 'int'>)
'float'(-12) -> -12.0 (<class 'float'>)
'round'(-12) -> -12 (<class 'sympy.core.numbers.Integer'>)
'ceil'(-12) -> -12 (<class 'sympy.core.numbers.Integer'>)
'floor'(-12) -> -12 (<class 'sympy.core.numbers.Integer'>)
'trunc'(-12) -> -12 (<class 'sympy.core.numbers.Integer'>)
'abs'(-12) -> 12 (<class 'sympy.core.numbers.Integer'>)
'index'(-12) -> -12 (<class 'int'>)
'invert'(-12) -> 11 (<class 'sympy.core.numbers.Integer'>)
'neg'(-12) -> 12 (<class 'sympy.core.numbers.Integer'>)
'pos'(-12) -> -12 (<class 'sympy.core.numbers.Integer'>)

In [28]: for binop in "+", "&", "==", "//", ">=", ">", "<=", "<<", "<", "%", "*", "!=", "|", "**", ">>", "-", "/", "^":
    ...:     s = "-12 {} 5".format(binop)
    ...:     r = S(s)
    ...:     print(f"{s} -> {r!r} ({type(r)!r})")
-12 + 5 -> -7 (<class 'sympy.core.numbers.Integer'>)
-12 & 5 -> 4 (<class 'sympy.core.numbers.Integer'>)
-12 == 5 -> False (<class 'bool'>)
-12 // 5 -> -3 (<class 'sympy.core.numbers.Integer'>)
-12 >= 5 -> False (<class 'sympy.logic.boolalg.BooleanFalse'>)
-12 > 5 -> False (<class 'sympy.logic.boolalg.BooleanFalse'>)
-12 <= 5 -> True (<class 'sympy.logic.boolalg.BooleanTrue'>)
-12 << 5 -> -384 (<class 'sympy.core.numbers.Integer'>)
-12 < 5 -> True (<class 'sympy.logic.boolalg.BooleanTrue'>)
-12 % 5 -> 3 (<class 'sympy.core.numbers.Integer'>)
-12 * 5 -> -60 (<class 'sympy.core.numbers.Integer'>)
-12 != 5 -> True (<class 'bool'>)
-12 | 5 -> -11 (<class 'sympy.core.numbers.Integer'>)
-12 ** 5 -> -248832 (<class 'sympy.core.numbers.Integer'>)
-12 >> 5 -> -1 (<class 'sympy.core.numbers.NegativeOne'>)
-12 - 5 -> -17 (<class 'sympy.core.numbers.Integer'>)
-12 / 5 -> -12/5 (<class 'sympy.core.numbers.Rational'>)
-12 ^ 5 -> -248832 (<class 'sympy.core.numbers.Integer'>)
```

If the above does not meet the request, I likely (still) misunderstood it. If so, can you please let me know what's still missing?

oscarbenjamin:
> If the above does not meet the request, I likely (still) misunderstood it.

I just wanted you to say "yes, I've checked all the methods" :)

I think that this looks good. I'll give some time to see if there are other reviews but otherwise I think this is good to merge.

sympy/core/numbers.py
smichr: > I don't think we need to check evaluate here since there is no unevaluated form.
```python
>>> Rational(4, 8, gcd=1)
4/8
```

smichr:
I commented above to show that unevaluated Rationals are possible:
```python
>>> Rational(4,8,gcd=1)
4/8
```

sympy/core/numbers.py
posita: @smichr, does this mean I should re-introduce the `if global_parameters.evaluate: …` check from 82f7de3?

oscarbenjamin:
> I commented above to show that unevaluated Rationals are possible:

It's not unevaluated Rationals but rather unevaluated bitwise operations. SymPy has no `LeftShift` or `BitwiseAnd` etc symbolic classes so there is no way for e.g. `1 << 2` to be unevaluated. For satisfying the numbers ABCs the operation `a << b` would have to return an Integral result to returning e.g. `LeftShift` because global evaluate is False would probably just break things.

posita:
What is the general preference for squashing/rebasing of an in-flight PR? I have recently rebased to `master` in my own repo, but am reluctant to overwrite history here, because I understand it can lead to inconveniences for reviewers. I looked at the various contributing docs, but couldn't find an answer to my question. (Apologies if I missed it.) Any guidance is appreciated.

oscarbenjamin:
I don't think rebasing is problematic for reviewers. The way I check out a branch for review is just (need to adjust git config to make this work)
```
git fetch upstream
git checkout upstream/pr/21699
```
and this command will happily update to the latest in the PR if it gets force pushed.

I think that this is small enough and simple enough for one commit although the 4 commits you have here are fine except for the commit messages:

1. The first commit refers to `numpy.Integral` rather than `numbers.Integral`.
2. The other commits which contain all of the substantive changes are all titled "Respond to PR feedback".

As a general rule I think that "Respond to PR feedback" is not a good commit message because it's useless when looking at either the blame or the log:
https://github.com/sympy/sympy/blame/b315e10118cbeb5cab716abd14e81cfaedf6c9c8/sympy/core/numbers.py#L2425

I would say squash this to one commit with a message like:
```
feat(core): make sympy.Integer compatible with Integral ABC

This commit makes sympy.Integer compatible with the numbers.Integral ABC
as defined in PEP XXX. The necessary bitwise operations (e.g. 1 << 2) are added
to sympy.Integer and sympy.Integer is registered with the Integral ABC so that

    isinstance(sympify(1), Integral) -> True
```

posita:
Commit history cleaned up. Please let me know if there's anything out of place, and thanks for the guidance!

asmeurer:
We register the classes because it is faster but you should be able to temporarily switch it to subclassing and that will tell you if any methods are missing.

oscarbenjamin:
> We register the classes because it is faster but you should be able to temporarily switch it to subclassing and that will tell you if any methods are missing.

Doesn't work because of metaclass issues:
```
  File "/space/enojb/sympy/36venv/bin/isympy", line 11, in <module>
    load_entry_point('sympy', 'console_scripts', 'isympy')()
  File "/space/enojb/sympy/sympy.git/isympy.py", line 338, in main
    from sympy.interactive import init_session
  File "/space/enojb/sympy/sympy.git/sympy/__init__.py", line 51, in <module>
    from .core import (sympify, SympifyError, cacheit, Basic, Atom,
  File "/space/enojb/sympy/sympy.git/sympy/core/__init__.py", line 9, in <module>
    from .expr import Expr, AtomicExpr, UnevaluatedExpr
  File "/space/enojb/sympy/sympy.git/sympy/core/expr.py", line 4087, in <module>
    from .mul import Mul
  File "/space/enojb/sympy/sympy.git/sympy/core/mul.py", line 2114, in <module>
    from .numbers import Rational
  File "/space/enojb/sympy/sympy.git/sympy/core/numbers.py", line 2039, in <module>
    class Integer(Rational, Integral_):
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

oscarbenjamin:
Looks good, merging. Thanks!

PR #21700, opened 2021-07-03
Add basis vector to body

sidhu1012:
#### Release Notes

* physics.mechanics
    * Add basis vectors to `Body`.

### DIFF
--- a/sympy/physics/mechanics/body.py
+++ b/sympy/physics/mechanics/body.py
@@ -136,24 +136,20 @@ class Body(RigidBody, Particle):  # type: ignore
         else:
             RigidBody.__init__(self, name, masscenter, frame, _mass, _inertia)
 
-        self._x = self.frame.x
-        self._y = self.frame.y
-        self._z = self.frame.z
-
     @property
     def x(self):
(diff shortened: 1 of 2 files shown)


sympy/physics/mechanics/body.py
```diff
@@ -136,6 +136,25 @@ def __init__(self, name, masscenter=None, mass=None, frame=None,
...
             RigidBody.__init__(self, name, masscenter, frame, _mass, _inertia)
 
+        self._x = self.frame.x
+        self._y = self.frame.y
+        self._z = self.frame.z
+
+    @property
+    def x(self):
+        """The basis Vector for the Body, in the x direction. """
+        return self._x
```
moorepants, line 146:
```suggestion
        return self,frame,x
```

moorepants: No reason not to do this, and it would ensure it always returns the correct value.

sidhu1012: I have declared `self._x` to `self.frame.x` in line 139

sidhu1012:
@moorepants Please review.

PR #21702, opened 2021-07-04
Added case for Str name for MatrixSymbol and a test for the same

### LINKED ISSUE #21623: Bug in assumption computation of expression with matrix element
The following code throws an exception when trying to compute whether `e` is finite with Sympy 1.8
```
from sympy import *
a = MatrixSymbol("a",5,1)
e = cosh(a[4, 0]**2)
e.is_finite
```

mostlyaman:

Fixes #21623 

MatrixSymbol name attribute is a Str, which is a Basic subclass and has no args. It does not support func-args invariant causing errors.

#### Other comments
Solution proposed:
https://github.com/sympy/sympy/issues/21623#issuecomment-861785256

#### Release Notes

NO ENTRY


mostlyaman:
@oscarbenjamin please review this PR

PR #21703, opened 2021-07-04
Implement `MIMOSeries` and `MIMOParallel`.

akshanshbhatt:

`MIMOSeries` and `MIMOParallel` classes will support [valid] MIMO transfer functions as arguments. `*` and `+` operators between two `TransferFunctionMatrix` object will return equivalent `MIMOSeries` and `MIMOParallel` objects.

#### Other comments

Checklist:

- [x] ~~Add MIMO support in the constructor of `Series` and `Parallel` classes respectively.~~ Add `MIMOSeries` and `MIMOParallel` classes.
- [x] Implement `doit()` in both `MIMOSeries` and `MIMOParallel`.
- [x] Add `num_input`, `num_output`, and `shape` properties in `MIMOSeries` and `MIMOParallel`.
- [x] ~~Modify `to_expr`, `is_proper`, `is_strictly_proper` and `is_biproper` in `Series` and `Parallel`.~~
- [x] Implement `_eval_rewrite_as_TransferFunctionMatrix()` method in both the classes.
- [x] Modify `__add__` and `__mul__` of the `TransferFunctionMatrix` class to return `MIMOParallel` and `MIMOSeries` objects respectively.
- [x] Modify `__add__`, `__mul__` and `__neg__` in `MIMOSeries` and `MIMOParallel`.
- [x] Add proper docstrings in both the classes with examples.
- [x] Add a common superclass for all the LTI systems.
- [x] Add unit tests for MIMO `Series` and `Parallel` classes covering all the possible outcomes.
  - [x] Tests for `MIMOSeries`
  - [x] Tests for `MIMOParallel`
- [x] Modify `latex.py` and add tests.
  - [x] Update `_print_MIMOSeries()`.
  - [x] Update `_print_MIMOParallel()`.
  - [x] Add tests.
- [x] Modify `pretty.py` and add tests.
  - [x] Update `_print_MIMOSeries()`.
  - [x] Update `_print_MIMOParallel()`.
  - [x] Add tests.
- [x] Modify `_SISO_doit()` to return `Add(*args)` and `Mul(*args)` instead of present implementation. Some tests need to be modified in that case. (**Disussion required before implementing**)
- [x] Add `evalf`, `expand` and `simplify` methods in `TransferFunctionMatrix` class.

#### Release Notes

* physics.control
  * Added Series and Parallel configuration's support for MIMO systems (`MIMOSeries` and `MIMOParallel` classes).

### DIFF
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -20,42 +20,22 @@ def _roots(poly, var):
         r = [rootof(poly, var, k) for k in range(n)]
     return r
 
-class LTICommon:
+
+class LinearTimeInvariant(Basic, EvalfMixin):
     """A common class for all the Linear Time-Invariant Dynamical Systems."""
     # Users should not directly interact with this class.
-    def __new__(cls, system, **kwargs):
-        if cls is LTICommon:
+    def __new__(cls, *system, **kwargs):
+        if cls is LinearTimeInvariant:
             raise NotImplementedError('The LTICommon class is not meant to be used directly.')
-        obj = super(LTICommon, cls).__new__(cls)
-        obj.__var = system.var
-        obj.__num_inputs = getattr(system, 'num_inputs', None)
-        obj.__num_outputs = getattr(system, 'num_outputs', None)
-        return obj
+        return super(LinearTimeInvariant, cls).__new__(cls, *system, **kwargs)
 
     @property
     def is_SISO(self):
         """Returns `True` if the passed LTI system is SISO else returns False."""
         return self._is_SISO
 
-    def _to_expr(self):
-        """Returns the equivalent ``Expr`` object."""
-        if not self.is_SISO:
-            raise TypeError("`to_expr` method is only valid for SISO Systems.")
 
-        if isinstance(self, Series):
-            return Mul(*(arg.to_expr() for arg in self.args), evaluate=False)
-        elif isinstance(self, Parallel):
-            return Add(*(arg.to_expr() for arg in self.args), evaluate=False)
-        elif isinstance(self, TransferFunction):
-            if self.num != 1:
-                return Mul(self.num, Pow(self.den, -1, evaluate=False), evaluate=False)
-            else:
-                return Pow(self.den, -1, evaluate=False)
-        else:
-            return ValueError("Cannot convert this  LTI system to expr.")
-
-
-class TransferFunction(Basic, EvalfMixin, LTICommon):
+class TransferFunction(LinearTimeInvariant):
     r"""
     A class for representing LTI (Linear, time-invariant) systems that can be strictly described
     by ratio of polynomials in the Laplace transform complex variable. The arguments
@@ -726,7 +706,11 @@ class TransferFunction(Basic, EvalfMixin, LTICommon):
         ((s - 3)*(s - 2))/(((s - 3)*(s - 2)*(s - 1)))
 
         """
-        return self._to_expr()
+
+        if self.num != 1:
+            return Mul(self.num, Pow(self.den, -1, evaluate=False), evaluate=False)
+        else:
+            return Pow(self.den, -1, evaluate=False)
 
 
 def _mat_mul_compatible(*args):
@@ -738,7 +722,7 @@ def _mat_mul_compatible(*args):
     )
 
 
-class Series(Basic, LTICommon):
+class Series(LinearTimeInvariant):
     r"""
     A class for representing series configuration of SISO and MIMO transfer functions.
 
@@ -854,7 +838,7 @@ class Series(Basic, LTICommon):
 
     """
     def __new__(cls, *args, evaluate=False):
-        if args is None:
+        if len(args) == 0:
             raise ValueError("Atleast 1 argument must be passed.")
 
         try:
@@ -935,16 +919,9 @@ class Series(Basic, LTICommon):
 
         """
         def _SISO_doit():
-            res = None
-            for arg in self.args:
-                arg = arg.doit()
-                if res is None:
-                    res = arg
-                else:
-                    num_ = arg.num * res.num
-                    den_ = arg.den * res.den
-                    res = TransferFunction(num_, den_, self.var)
-            return res
+            _arg = (arg.doit().to_expr() for arg in self.args)
+            res = Mul(*_arg, evaluate=True)
+            return TransferFunction.from_rational_expression(res, self.var)
 
         def _MIMO_doit():
             _arg = (arg.doit()._expr_mat for arg in reversed(self.args))
@@ -1033,7 +1010,9 @@ class Series(Basic, LTICommon):
 
     def to_expr(self):
         """Returns the equivalent ``Expr`` object."""
-        return self._to_expr()
+        if not self.is_SISO:
+            raise ValueError("`to_expr` is only valid for SISO systems.")
+        return Mul(*(arg.to_expr() for arg in self.args), evaluate=False)
 
     @property
     def is_proper(self):
@@ -1123,7 +1102,7 @@ class Series(Basic, LTICommon):
         return self.doit().is_biproper if self.is_SISO else None
 
 
-class Parallel(Basic, LTICommon):
+class Parallel(LinearTimeInvariant):
     r"""
     A class for representing parallel configuration of SISO and MIMO transfer functions.
 
@@ -1182,7 +1161,7 @@ class Parallel(Basic, LTICommon):
     You can get the resultant transfer function by using ``.doit()`` method:
 
     >>> Parallel(tf1, tf2, -tf3).doit()
-    TransferFunction(-p**2*(-p + s)*(s**4 + 5*s + 6) + (p + s)*((-p + s)*(s**3 - 2) + (a*p**2 + b*s)*(s**4 + 5*s + 6)), (-p + s)*(p + s)*(s**4 + 5*s + 6), s)
+    TransferFunction(-p**2*(-p + s)*(s**4 + 5*s + 6) + (-p + s)*(p + s)*(s**3 - 2) + (p + s)*(a*p**2 + b*s)*(s**4 + 5*s + 6), (-p + s)*(p + s)*(s**4 + 5*s + 6), s)
     >>> Parallel(tf2, Series(tf1, -tf3)).doit()
     TransferFunction(-p**2*(a*p**2 + b*s)*(s**4 + 5*s + 6) + (-p + s)*(p + s)*(s**3 - 2), (-p + s)*(p + s)*(s**4 + 5*s + 6), s)
 
@@ -1235,7 +1214,7 @@ class Parallel(Basic, LTICommon):
 
     """
     def __new__(cls, *args, evaluate=False):
-        if args is None:
+        if len(args) == 0:
             raise ValueError("Atleast 1 argument must be passed.")
 
         try:
@@ -1316,16 +1295,9 @@ class Parallel(Basic, LTICommon):
         """
 
         def _SISO_doit():
-            res = None
-            for arg in self.args:
-                arg = arg.doit()
-                if res is None:
-                    res = arg
-                else:
-                    num_ = res.num * arg.den + res.den * arg.num
-                    den_ = res.den * arg.den
-                    res = TransferFunction(num_, den_, self.var)
-            return res
+            _arg = (arg.doit().to_expr() for arg in self.args)
+            res = Add(*_arg, evaluate=True)
+            return TransferFunction.from_rational_expression(res, self.var)
 
         def _MIMO_doit():
             _arg = (arg.doit()._expr_mat for arg in self.args)
@@ -1393,7 +1365,9 @@ class Parallel(Basic, LTICommon):
 
     def to_expr(self):
         """Returns the equivalent ``Expr`` object."""
-        return self._to_expr()
+        if not self.is_SISO:
+            raise ValueError("`to_expr` is only valid for SISO systems.")
+        return Add(*(arg.to_expr() for arg in self.args), evaluate=False)
 
     @property
     def is_proper(self):
@@ -1692,7 +1666,7 @@ def _to_TFM(mat, var):
     return TransferFunctionMatrix(arg)
 
 
-class TransferFunctionMatrix(Basic, LTICommon):
+class TransferFunctionMatrix(LinearTimeInvariant):
     r"""
     A class for representing the MIMO (multiple-input and multiple-output)
     generalization of the SISO (single-input and single-output) transfer function.
@@ -2057,7 +2031,7 @@ class TransferFunctionMatrix(Basic, LTICommon):
         for row_index, row in enumerate(arg):
             temp = []
             for col_index, element in enumerate(row):
-                if not isinstance(element, LTICommon):
+                if not isinstance(element, LinearTimeInvariant):
                     raise TypeError("Incompatible type found as the element of "
                         "TransferFunctionMatrix.")
 
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -795,7 +795,7 @@ class Series(Basic):
     [ 1 ]{t}*[   6*s ]{t}*[s  1]{t}
     >>> Series(tfm_c, tfm_b, tfm_a).doit()
     TransferFunctionMatrix(((TransferFunction(25*(6*s**3 + 1), 6*s**2, s), TransferFunction(5*(30*s**3 + 1), 6*s, s)), (TransferFunction(25*(6*s**3 + 1), 6*s**3, s), TransferFunction(5*(30*s**3 + 1), 6*s**2, s))))
-    >>> pprint(_, use_unicode=False)  # (2 Inputs -A-> 2 Outputs) -> (2 Inputs -B-> 1 Output) -> (1 Input -C-> 2 Outputs) is equivalent to (2 Inputs -Series Equivalent-> 2 Outputs). 
+    >>> pprint(_, use_unicode=False)  # (2 Inputs -A-> 2 Outputs) -> (2 Inputs -B-> 1 Output) -> (1 Input -C-> 2 Outputs) is equivalent to (2 Inputs -Series Equivalent-> 2 Outputs).
     [   /   3    \    /    3    \]
     [25*\6*s  + 1/  5*\30*s  + 1/]
     [-------------  -------------]
@@ -1346,7 +1346,7 @@ class Parallel(Basic):
 
         self_arg_list = list(self.args)
         return Parallel(*self_arg_list, other)
-    
+
     __radd__ = __add__
 
     def __sub__(self, other):
@@ -1921,11 +1921,14 @@ class TransferFunctionMatrix(Basic):
 
     Addition, subtraction, and multiplication of transfer function matrices can form
     unevaluated ``Series`` or ``Parallel`` objects.
-    For addition and subtraction:
-        All the transfer function matrices must have the same shape.
-    For multiplication (C = A * B):
-        The number of inputs of the first transfer function matrix (A) must be equal to the
-        number of outputs of the second transfer function matrix (B).
+
+    - For addition and subtraction:
+      All the transfer function matrices must have the same shape.
+
+    - For multiplication (C = A * B):
+      The number of inputs of the first transfer function matrix (A) must be equal to the
+      number of outputs of the second transfer function matrix (B).
+
     Also, use pretty-printing (``pprint``) to analyse better.
 
     >>> tfm_8 = TransferFunctionMatrix([[tf_3], [tf_2], [-tf_1]])
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -1296,8 +1296,8 @@ class Parallel(LinearTimeInvariant):
 
         def _SISO_doit():
             _arg = (arg.doit().to_expr() for arg in self.args)
-            res = Add(*_arg, evaluate=True)
-            return TransferFunction.from_rational_expression(res, self.var)
+            res = Add(*_arg).as_numer_denom()
+            return TransferFunction(*res, self.var)
 
         def _MIMO_doit():
             _arg = (arg.doit()._expr_mat for arg in self.args)
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -7,6 +7,7 @@ from sympy.core.sympify import sympify, _sympify
 from sympy.polys import Poly, rootof
 from sympy.series import limit
 from sympy.matrices import ImmutableMatrix
+from sympy.matrices.expressions import MatMul, MatAdd
 
 __all__ = ['TransferFunction', 'Series', 'Parallel', 'Feedback', 'TransferFunctionMatrix']
 
@@ -221,6 +222,7 @@ class TransferFunction(Basic, EvalfMixin):
             obj._num = num
             obj._den = den
             obj._var = var
+            obj.is_SISO = True
             return obj
 
         else:
@@ -697,13 +699,53 @@ class TransferFunction(Basic, EvalfMixin):
             return Pow(self.den, -1, evaluate=False)
 
 
+def _mat_mul_compatible(*args):
+    """To check whether shapes are compatible for matrix mul."""
+    compatible = True
+    for i, arg in enumerate(args, start=1):
+        try:
+            if arg.num_outputs != args[i].num_inputs:
+                compatible = False
+        except IndexError:
+            return compatible
+    return compatible
+
+
 class Series(Basic):
-    """
-    A class for representing product of transfer functions or transfer functions in a
-    series configuration.
+    r"""
+    A class for representing series configuration of SISO and MIMO transfer functions.
+
+    Parameters
+    ==========
+
+    args : TransferFunction, TransferFunctionMatrix, Series, Parallel
+        Systems in series configuration.
+    evaluate : Boolean, Keyword
+        When passed ``True``, returns the equivalent
+        Series(*args).doit(). Set to ``False`` by default.
+
+    Raises
+    ======
+
+    ValueError
+        When no argument is passed.
+
+        ``var`` attribute is not same for every system.
+
+        ``num_outputs`` of the MIMO system is not equal to the
+        ``num_inputs`` of its adjacent MIMO system. (Matrix
+        multiplication constraint, basically)
+    TypeError
+        Any of the passed ``*args`` has unsupported type
+
+        A combination of SISO and MIMO systems is
+        passed. There should be homogeneity in the
+        type of systems passed.
 
     Examples
     ========
+
+    SISO-System Examples -
 
     >>> from sympy.abc import s, p, a, b
     >>> from sympy.physics.control.lti import TransferFunction, Series, Parallel
@@ -738,8 +780,11 @@ class Series(Basic):
     Notes
     =====
 
-    All the transfer functions should use the same complex variable
-    ``var`` of the Laplace transform.
+    All the transfer functions should use the same complex variable ``var`` of the Laplace transform.
+
+    ``Series(A, B)`` is not equivalent to ``A*B``. It is always in the reverse order, that is ``B*A``.
+    In, SISO systems, it hardly matters, but in MIMO systems, there can be issues due to non-commutative
+    nature of Matrix Multiplication.
 
     See Also
     ========
@@ -748,20 +793,29 @@ class Series(Basic):
 
     """
     def __new__(cls, *args, evaluate=False):
-        if not all(isinstance(arg, (TransferFunction, Parallel, Series)) for arg in args):
-            raise TypeError("Unsupported type of argument(s) for Series.")
+        try:
+            type_set = {arg.is_SISO for arg in args}
+            if len(type_set) != 1:
+                raise TypeError("Invalid system. `args` passed contain SISO as well "
+                    "as MIMO elements.")
+            var_set = {arg.var for arg in args}
+            if len(var_set) != 1:
+                raise ValueError("All transfer functions should use the same complex variable"
+                    f" of the Laplace transform. {len(var_set)} different values found.")
+        except AttributeError:
+            raise TypeError("Unsupported type of argument passed.")
 
-        obj = super().__new__(cls, *args)
-        obj._var = None
-        for arg in args:
-            if obj._var is None:
-                obj._var = arg.var
-            elif obj._var != arg.var:
-                raise ValueError("All transfer functions should use the same complex"
-                    " variable of the Laplace transform.")
-        if evaluate:
-            return obj.doit()
-        return obj
+        if list(type_set)[0]:
+            obj = super().__new__(cls, *args)
+            obj.is_SISO = True
+        elif _mat_mul_compatible(*args):
+            obj = super().__new__(cls, *args)
+            obj.is_SISO = False
+        else:
+            raise ValueError("Number of input signals do not match the number"
+                " of output signals for some args.")
+
+        return obj.doit() if evaluate else obj
 
     @property
     def var(self):
@@ -782,7 +836,30 @@ class Series(Basic):
         p
 
         """
-        return self._var
+        return self.args[0].var
+
+    @property
+    def num_inputs(self):
+        # If the Series is for TFMs, then return num_inputs of the first TFM arg as num_input for the series system
+        try:
+            return self.args[0].num_inputs
+        # If no such attribute is found, then its TF instead of TFM and num inputs for tf should be None
+        except AttributeError:
+            return None
+
+    @property
+    def num_outputs(self):
+        try:
+            return self.args[-1].num_outputs
+        except AttributeError:
+            return None
+
+    @property
+    def shape(self):
+        try:
+            return self.args[-1].num_outputs, self.args[0].num_inputs
+        except AttributeError:
+            return None
 
     def doit(self, **kwargs):
         """
@@ -802,16 +879,27 @@ class Series(Basic):
         TransferFunction((2 - s**3)*(-a*p**2 - b*s), (-p + s)*(s**4 + 5*s + 6), s)
 
         """
-        res = None
-        for arg in self.args:
-            arg = arg.doit()
-            if res is None:
-                res = arg
-            else:
-                num_ = arg.num * res.num
-                den_ = arg.den * res.den
-                res = TransferFunction(num_, den_, self.var)
-        return res
+        def _SISO_doit():
+            res = None
+            for arg in self.args:
+                arg = arg.doit()
+                if res is None:
+                    res = arg
+                else:
+                    num_ = arg.num * res.num
+                    den_ = arg.den * res.den
+                    res = TransferFunction(num_, den_, self.var)
+            return res
+
+        def _MIMO_doit():
+            _arg = (arg.doit()._expr_mat for arg in reversed(self.args))
+            res = MatMul(*_arg, evaluate=True)
+            return TransferFunctionMatrix.from_Matrix(res, self.var)
+
+        if self.is_SISO:
+            return _SISO_doit()
+        else:
+            return _MIMO_doit()
 
     def _eval_rewrite_as_TransferFunction(self, *args, **kwargs):
         return self.doit()
@@ -977,15 +1065,42 @@ class Series(Basic):
 
 
 class Parallel(Basic):
-    """
-    A class for representing addition of transfer functions or transfer functions
-    in a parallel configuration.
+    r"""
+    A class for representing parallel configuration of SISO and MIMO transfer functions.
+
+    Parameters
+    ==========
+
+    args : TransferFunction, TransferFunctionMatrix, Series, Parallel
+        Systems in parallel arrangement
+    evaluate : Boolean, Keyword
+        When passed ``True``, returns the equivalent
+        Parallel(*args).doit(). Set to ``False`` by default.
+
+    Raises
+    ======
+
+    ValueError
+        When no argument is passed.
+
+        ``var`` attribute is not same for every system.
+
+        All MIMO systems passed don't have same shape.
+    TypeError
+        Any of the passed ``*args`` has unsupported type
+
+        A combination of SISO and MIMO systems is
+        passed. There should be homogeneity in the
+        type of systems passed.
 
     Examples
     ========
 
+    SISO-System examples -
+
     >>> from sympy.abc import s, p, a, b
-    >>> from sympy.physics.control.lti import TransferFunction, Parallel, Series
+    >>> from sympy.physics.control.lti import TransferFunction, TransferFunctionMatrix, Parallel, Series
+    >>> from sympy import Matrix
     >>> tf1 = TransferFunction(a*p**2 + b*s, s - p, s)
     >>> tf2 = TransferFunction(s**3 - 2, s**4 + 5*s + 6, s)
     >>> tf3 = TransferFunction(p**2, p + s, s)
@@ -1012,6 +1127,18 @@ class Parallel(Basic):
     >>> Parallel(tf2, Series(tf1, -tf3)).doit()
     TransferFunction(-p**2*(a*p**2 + b*s)*(s**4 + 5*s + 6) + (-p + s)*(p + s)*(s**3 - 2), (-p + s)*(p + s)*(s**4 + 5*s + 6), s)
 
+    MIMO-System example -
+
+    >>> expr_1 = 1/s
+    >>> expr_2 = s/(s**2-1)
+    >>> expr_3 = (2 + s)/(s**2 - 1)
+    >>> expr_4 = 5
+    >>> tfm_a = TransferFunctionMatrix.from_Matrix(Matrix([[expr_1, expr_2], [expr_3, expr_4]]), s)
+    >>> tfm_b = TransferFunctionMatrix.from_Matrix(Matrix([[expr_2, expr_1], [expr_4, expr_3]]), s)
+    >>> tfm_c = TransferFunctionMatrix.from_Matrix(Matrix([[expr_3, expr_4], [expr_1, expr_2]]), s)
+    >>> Parallel(tfm_a, tfm_b, tfm_c)
+    Parallel(TransferFunctionMatrix(((TransferFunction(1, s, s), TransferFunction(s, s**2 - 1, s)), (TransferFunction(s + 2, s**2 - 1, s), TransferFunction(5, 1, s)))), TransferFunctionMatrix(((TransferFunction(s, s**2 - 1, s), TransferFunction(1, s, s)), (TransferFunction(5, 1, s), TransferFunction(s + 2, s**2 - 1, s)))), TransferFunctionMatrix(((TransferFunction(s + 2, s**2 - 1, s), TransferFunction(5, 1, s)), (TransferFunction(1, s, s), TransferFunction(s, s**2 - 1, s)))))
+
     Notes
     =====
 
@@ -1025,20 +1152,28 @@ class Parallel(Basic):
 
     """
     def __new__(cls, *args, evaluate=False):
-        if not all(isinstance(arg, (TransferFunction, Series, Parallel)) for arg in args):
-            raise TypeError("Unsupported type of argument(s) for Parallel.")
+        try:
+            type_set = {arg.is_SISO for arg in args}
+            if len(type_set) != 1:
+                raise TypeError("Invalid system. `args` passed contain SISO as well "
+                    "as MIMO elements.")
+            var_set = {arg.var for arg in args}
+            if len(var_set) != 1:
+                raise ValueError("All transfer functions should use the same complex variable"
+                    f" of the Laplace transform. {len(var_set)} different values found.")
+        except AttributeError:
+            raise TypeError("Unsupported type of argument passed.")
 
-        obj = super().__new__(cls, *args)
-        obj._var = None
-        for arg in args:
-            if obj._var is None:
-                obj._var = arg.var
-            elif obj._var != arg.var:
-                raise ValueError("All transfer functions should use the same complex"
-                    " variable of the Laplace transform.")
-        if evaluate:
-            return obj.doit()
-        return obj
+        if list(type_set)[0]:
+            obj = super().__new__(cls, *args)
+            obj.is_SISO = True
+        elif all(arg.shape == args[0].shape for arg in args):
+            obj = super().__new__(cls, *args)
+            obj.is_SISO = False
+        else:
+            raise TypeError("Shape of all the args is not equal.")
+
+        return obj.doit() if evaluate else obj
 
     @property
     def var(self):
@@ -1059,7 +1194,28 @@ class Parallel(Basic):
         p
 
         """
-        return self._var
+        return self.args[0].var
+
+    @property
+    def num_inputs(self):
+        try:
+            return self.args[0].num_inputs
+        except AttributeError:
+            return None
+
+    @property
+    def num_outputs(self):
+        try:
+            return self.args[0].num_outputs
+        except AttributeError:
+            return None
+
+    @property
+    def shape(self):
+        try:
+            return self.args[0].shape
+        except AttributeError:
+            return None
 
     def doit(self, **kwargs):
         """
@@ -1079,16 +1235,28 @@ class Parallel(Basic):
         TransferFunction((2 - s**3)*(-p + s) + (-a*p**2 - b*s)*(s**4 + 5*s + 6), (-p + s)*(s**4 + 5*s + 6), s)
 
         """
-        res = None
-        for arg in self.args:
-            arg = arg.doit()
-            if res is None:
-                res = arg
-            else:
-                num_ = res.num * arg.den + res.den * arg.num
-                den_ = res.den * arg.den
-                res = TransferFunction(num_, den_, self.var)
-        return res
+
+        def _SISO_doit():
+            res = None
+            for arg in self.args:
+                arg = arg.doit()
+                if res is None:
+                    res = arg
+                else:
+                    num_ = res.num * arg.den + res.den * arg.num
+                    den_ = res.den * arg.den
+                    res = TransferFunction(num_, den_, self.var)
+            return res
+
+        def _MIMO_doit():
+            _arg = (arg.doit()._expr_mat for arg in self.args)
+            res = MatAdd(*_arg, evaluate=True)
+            return TransferFunctionMatrix.from_Matrix(res, self.var)
+
+        if self.is_SISO:
+            return _SISO_doit()
+        else:
+            return _MIMO_doit()
 
     def _eval_rewrite_as_TransferFunction(self, *args, **kwargs):
         return self.doit()
@@ -1702,9 +1870,13 @@ class TransferFunctionMatrix(Basic):
         for row_index, row in enumerate(arg):
             temp = []
             for col_index, element in enumerate(row):
-                if not isinstance(element, (TransferFunction, Series, Parallel)):
+                try:
+                    if not element.is_SISO:
+                        raise TypeError("MIMO Series/Parallel object found as the element of "
+                        f"TransferFunctionMatrix at index {(row_index, col_index)}.")
+                except AttributeError:
                     raise TypeError("Incompatible type found as the element of "
-                    f"TransferFunctionMatrix at index {(row_index, col_index)}.")
+                        f"TransferFunctionMatrix at index {(row_index, col_index)}.")
 
                 if var != element.var:
                     raise ValueError("Conflicting value(s) found for `var`. All TransferFunction instances in "
@@ -1720,6 +1892,7 @@ class TransferFunctionMatrix(Basic):
 
         obj = super(TransferFunctionMatrix, cls).__new__(cls, arg)
         obj._expr_mat = ImmutableMatrix(expr_mat_arg)
+        obj.is_SISO = False
         return obj
 
     @classmethod
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -745,7 +745,8 @@ class Series(Basic):
     SISO-System Examples -
 
     >>> from sympy.abc import s, p, a, b
-    >>> from sympy.physics.control.lti import TransferFunction, Series, Parallel
+    >>> from sympy.physics.control.lti import TransferFunction, Series, Parallel, TransferFunctionMatrix
+    >>> from sympy import Matrix, pprint
     >>> tf1 = TransferFunction(a*p**2 + b*s, s - p, s)
     >>> tf2 = TransferFunction(s**3 - 2, s**4 + 5*s + 6, s)
     >>> tf3 = TransferFunction(p**2, p + s, s)
@@ -773,6 +774,39 @@ class Series(Basic):
     >>> S4 = Series(tf2, Parallel(tf1, -tf3))
     >>> S4.doit()
     TransferFunction((s**3 - 2)*(-p**2*(-p + s) + (p + s)*(a*p**2 + b*s)), (-p + s)*(p + s)*(s**4 + 5*s + 6), s)
+
+    SISO-System Examples -
+
+    >>> mat_a = Matrix([[5*s], [5]])  # 2 Outputs 1 Input
+    >>> mat_b = Matrix([[5, 1/(6*s**2)]])  # 1 Output 2 Inputs
+    >>> mat_c = Matrix([[1, s], [5/s, 1]])  # 2 Outputs 2 Inputs
+    >>> tfm_a = TransferFunctionMatrix.from_Matrix(mat_a, s)
+    >>> tfm_b = TransferFunctionMatrix.from_Matrix(mat_b, s)
+    >>> tfm_c = TransferFunctionMatrix.from_Matrix(mat_c, s)
+    >>> Series(tfm_c, tfm_b, tfm_a)
+    Series(TransferFunctionMatrix(((TransferFunction(1, 1, s), TransferFunction(s, 1, s)), (TransferFunction(5, s, s), TransferFunction(1, 1, s)))), TransferFunctionMatrix(((TransferFunction(5, 1, s), TransferFunction(1, 6*s**2, s)),)), TransferFunctionMatrix(((TransferFunction(5*s, 1, s),), (TransferFunction(5, 1, s),))))
+    >>> pprint(_, use_unicode=False)  #  For Better Visualization
+    [5*s]                 [1  s]
+    [---]                 [-  -]
+    [ 1 ]                 [1  1]
+    [   ]    [5   1  ]    [    ]
+    [ 5 ]    [-  ----]    [5  1]
+    [ - ]    [1     2]    [-  -]
+    [ 1 ]{t}*[   6*s ]{t}*[s  1]{t}
+    >>> Series(tfm_c, tfm_b, tfm_a).doit()
+    TransferFunctionMatrix(((TransferFunction(25*(6*s**3 + 1), 6*s**2, s), TransferFunction(5*(30*s**3 + 1), 6*s, s)), (TransferFunction(25*(6*s**3 + 1), 6*s**3, s), TransferFunction(5*(30*s**3 + 1), 6*s**2, s))))
+    >>> pprint(_, use_unicode=False)  # (2 Inputs -A-> 2 Outputs) -> (2 Inputs -B-> 1 Output) -> (1 Input -C-> 2 Outputs) is equivalent to (2 Inputs -Series Equivalent-> 2 Outputs). 
+    [   /   3    \    /    3    \]
+    [25*\6*s  + 1/  5*\30*s  + 1/]
+    [-------------  -------------]
+    [        2           6*s     ]
+    [     6*s                    ]
+    [                            ]
+    [   /   3    \    /    3    \]
+    [25*\6*s  + 1/  5*\30*s  + 1/]
+    [-------------  -------------]
+    [        3              2    ]
+    [     6*s            6*s     ]{t}
 
     Notes
     =====
@@ -1104,7 +1138,7 @@ class Parallel(Basic):
 
     >>> from sympy.abc import s, p, a, b
     >>> from sympy.physics.control.lti import TransferFunction, TransferFunctionMatrix, Parallel, Series
-    >>> from sympy import Matrix
+    >>> from sympy import Matrix, pprint
     >>> tf1 = TransferFunction(a*p**2 + b*s, s - p, s)
     >>> tf2 = TransferFunction(s**3 - 2, s**4 + 5*s + 6, s)
     >>> tf3 = TransferFunction(p**2, p + s, s)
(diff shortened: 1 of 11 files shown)


namannimmo10:
Thanks for adding a checklist this time.  Looks like a good start. Please get feedback on the class level docstrings and unit tests first, before further moving on to the implementation.

sympy/printing/latex.py
```diff
@@ -2449,7 +2449,7 @@ def _print_Parallel(self, expr):
         args = list(expr.args)
         parens = lambda x: self.parenthesize(x, precedence_traditional(expr),
                                              False)
-        return ' '.join(map(parens, args))
+        return '+'.join(map(parens, args))
```
akshanshbhatt, line 2452: @namannimmo10, In my opinion, we should change this function. First of all, it is not printing equivalent to `expr`'s pretty form, and secondly, it adds unnecessary parentheses. It is just the Series' latex parsing function. If you can, suggest any good fix.

image

compared to
```
                                       [   3      ]   
                                       [5*s    s  ]
[5]                                    [----   -  ]
[-]                                    [ 1     1  ]
[1]                                    [          ]
[ ]    /[      3]      [      3]   \   [         2]
[5]    |[5  6*s ]      [5  6*s ]   |   [ 5    5*s ]
[-]    |[-  ----]      [-  ----]   |   [ -    ----]
[1]{t}*\[1   1  ]{t} + [1   1  ]{t}/ + [ 1     1  ]{t}
```
For Series too we should consider adding dot instead of space.

sympy/physics/control/lti.py
```diff
@@ -735,11 +775,47 @@ class Series(Basic):
...
+
+    >>> mat_a = Matrix([[5*s], [5]])  # 2 Outputs 1 Input
+    >>> mat_b = Matrix([[5, 1/(6*s**2)]])  # 1 Output 2 Inputs
+    >>> mat_c = Matrix([[1, s], [5/s, 1]])  # 2 Outputs 2 Inputs
+    >>> tfm_a = TransferFunctionMatrix.from_Matrix(mat_a, s)
+    >>> tfm_b = TransferFunctionMatrix.from_Matrix(mat_b, s)
+    >>> tfm_c = TransferFunctionMatrix.from_Matrix(mat_c, s)
+    >>> Series(tfm_c, tfm_b, tfm_a)
+    Series(TransferFunctionMatrix(((TransferFunction(1, 1, s), TransferFunction(s, 1, s)), (TransferFunction(5, s, s), TransferFunction(1, 1, s)))), TransferFunctionMatrix(((TransferFunction(5, 1, s), TransferFunction(1, 6*s**2, s)),)), TransferFunctionMatrix(((TransferFunction(5*s, 1, s),), (TransferFunction(5, 1, s),))))
+    >>> pprint(_, use_unicode=False)  #  For Better Visualization
```
namannimmo10, line 788: Let's pass `use_unicode=True` in the docstrings? It's much readable for the users.
```python
In [76]: G1 = TransferFunctionMatrix.from_Matrix(Matrix([[5*s], [5]]), s)                 

In [77]: G2 = TransferFunctionMatrix.from_Matrix(Matrix([[5, 1/(6*s**2)]]), s)            

In [78]: G3 = TransferFunctionMatrix.from_Matrix(Matrix([[1, s], [5/s, 1]]), s)           

In [79]: pprint(Series(G3, G2, G1))                                                       
⎡5⋅s⎤             ⎡1  s⎤ 
⎢───⎥             ⎢─  ─⎥ 
⎢ 1 ⎥             ⎢1  1⎥ 
⎢   ⎥  ⎡5   1  ⎤  ⎢    ⎥ 
⎢ 5 ⎥  ⎢─  ────⎥  ⎢5  1⎥ 
⎢ ─ ⎥  ⎢1     2⎥  ⎢─  ─⎥ 
⎣ 1 ⎦τ⋅⎣   6⋅s ⎦τ⋅⎣s  1⎦τ
```
Dot should be placed slightly above; notice how it is done here:
```python
In [27]: mat_a = Matrix([[5*s], [5]])                                                 

In [28]: mat_b = Matrix([[5, 1/(6*s**2)]])                                                                              
                                                  
In [32]: pprint(MatMul(mat_a, mat_b))                                                                    
      ⎡    1  ⎤
⎡5⋅s⎤ ⎢5  ────⎥
⎢   ⎥⋅⎢      2⎥
⎣ 5 ⎦ ⎣   6⋅s ⎦
```

namannimmo10:
When we try to connect several MIMO systems, the TFM breaks;
```python
In [101]: tfm_a, tfm_b, tfm_c, tfm_d                                                       
Out[101]: 
⎛⎡  1      1  ⎤                                                                ⎞
⎜⎢  ─    ─────⎥                                                                ⎟
⎜⎢  s    s + 1⎥                           ⎡     2      ⎤                       ⎟
⎜⎢            ⎥                           ⎢────────────⎥                       ⎟
⎜⎢  s      1  ⎥                           ⎢ 2          ⎥   ⎡   1         0   ⎤ ⎟
⎜⎢─────    ─  ⎥                           ⎢s  - 4⋅s + 1⎥   ⎢────────     ─   ⎥ ⎟
⎜⎢s + 3    s  ⎥                           ⎢            ⎥   ⎢10⋅s + 1     1   ⎥ ⎟
⎜⎢            ⎥                           ⎢    2⋅s     ⎥   ⎢                 ⎥ ⎟
⎜⎢  1      1  ⎥   ⎡  1      3    s + 1⎤   ⎢────────────⎥   ⎢   0         1   ⎥ ⎟
⎜⎢─────  ─────⎥   ⎢─────  ─────  ─────⎥   ⎢ 2          ⎥   ⎢   ─      ───────⎥ ⎟
⎝⎣s + 1  s + 3⎦τ, ⎣s + 1  s + 1  s + 3⎦τ, ⎣s  - 4⋅s + 1⎦τ, ⎣   1      5⋅s + 1⎦τ⎠

In [102]: Series(tfm_a, tfm_b, tfm_c, tfm_d).doit()                                                                                                                               
Out[102]: 
⎡  ⎛                                               2                                
⎢  ⎜   2                           2 ⎛ 2          ⎞             2                   
⎢2⋅⎝3⋅s ⋅(s + 1)⋅(s + 3)⋅(10⋅s + 1) ⋅⎝s  - 4⋅s + 1⎠  + s⋅(s + 1) ⋅(s + 3)⋅(10⋅s + 1)
⎢───────────────────────────────────────────────────────────────────────────────────
⎢                                                                                   
⎢                                                           2        2           3 ⎛
⎢                                                  s⋅(s + 1) ⋅(s + 3) ⋅(10⋅s + 1) ⋅⎝
⎢                                                                                   
⎢   ⎛                                              2                                
⎢   ⎜   2                          2 ⎛ 2          ⎞             2                  2
⎢ 2⋅⎝3⋅s ⋅(s + 1)⋅(s + 3)⋅(5⋅s + 1) ⋅⎝s  - 4⋅s + 1⎠  + s⋅(s + 1) ⋅(s + 3)⋅(5⋅s + 1) 
⎢ ──────────────────────────────────────────────────────────────────────────────────
⎢                                                                                   
⎢                                                           2        2          3 ⎛ 
⎣                                                    (s + 1) ⋅(s + 3) ⋅(5⋅s + 1) ⋅⎝s

                2                                              2⎞    ⎛              
2 ⎛ 2          ⎞                   2           2 ⎛ 2          ⎞ ⎟    ⎜         4    
 ⋅⎝s  - 4⋅s + 1⎠  + (s + 1)⋅(s + 3) ⋅(10⋅s + 1) ⋅⎝s  - 4⋅s + 1⎠ ⎠  2⋅⎝s⋅(s + 1) ⋅(10
─────────────────────────────────────────────────────────────────  ─────────────────
             3                                                                      
 2          ⎞                                                                       
s  - 4⋅s + 1⎠                                                                       
                                                                                    
               2                                             2⎞       ⎛             
 ⎛ 2          ⎞                   2          2 ⎛ 2          ⎞ ⎟       ⎜         4   
⋅⎝s  - 4⋅s + 1⎠  + (s + 1)⋅(s + 3) ⋅(5⋅s + 1) ⋅⎝s  - 4⋅s + 1⎠ ⎠     2⋅⎝s⋅(s + 1) ⋅(5
───────────────────────────────────────────────────────────────     ────────────────
            3                                                                       
2          ⎞                                                                        
  - 4⋅s + 1⎠                                                                        

                       2                                                2           
       2 ⎛ 2          ⎞                     2           2 ⎛ 2          ⎞            
⋅s + 1) ⋅⎝s  - 4⋅s + 1⎠  + s⋅(s + 1)⋅(s + 3) ⋅(10⋅s + 1) ⋅⎝s  - 4⋅s + 1⎠  + 3⋅(s + 1
────────────────────────────────────────────────────────────────────────────────────
                                                                            3       
                                       3        2           3 ⎛ 2          ⎞        
                              s⋅(s + 1) ⋅(s + 3) ⋅(10⋅s + 1) ⋅⎝s  - 4⋅s + 1⎠        
                                                                                    
                       2                                               2            
       2 ⎛ 2          ⎞                     2          2 ⎛ 2          ⎞             
⋅s + 1) ⋅⎝s  - 4⋅s + 1⎠  + s⋅(s + 1)⋅(s + 3) ⋅(5⋅s + 1) ⋅⎝s  - 4⋅s + 1⎠  + 3⋅(s + 1)
────────────────────────────────────────────────────────────────────────────────────
                                                                          3         
                                      3        2          3 ⎛ 2          ⎞          
                               (s + 1) ⋅(s + 3) ⋅(5⋅s + 1) ⋅⎝s  - 4⋅s + 1⎠          

                                      2⎞⎤ 
 2        2           2 ⎛ 2          ⎞ ⎟⎥ 
) ⋅(s + 3) ⋅(10⋅s + 1) ⋅⎝s  - 4⋅s + 1⎠ ⎠⎥ 
────────────────────────────────────────⎥ 
                                        ⎥ 
                                        ⎥ 
                                        ⎥ 
                                        ⎥ 
                                    2⎞  ⎥ 
2        2          2 ⎛ 2          ⎞ ⎟  ⎥ 
 ⋅(s + 3) ⋅(5⋅s + 1) ⋅⎝s  - 4⋅s + 1⎠ ⎠  ⎥ 
──────────────────────────────────────  ⎥ 
                                        ⎥ 
                                        ⎥ 
                                        ⎦τ


```

sympy/physics/control/lti.py
```diff
@@ -1702,9 +2043,13 @@ def __new__(cls, arg):
...
             for col_index, element in enumerate(row):
-                if not isinstance(element, (TransferFunction, Series, Parallel)):
+                try:
+                    if not element.is_SISO:
+                        raise TypeError("MIMO Series/Parallel object found as the element of "
+                        f"TransferFunctionMatrix at index {(row_index, col_index)}.")
+                except AttributeError:
                     raise TypeError("Incompatible type found as the element of "
-                    f"TransferFunctionMatrix at index {(row_index, col_index)}.")
+                        f"TransferFunctionMatrix at index {(row_index, col_index)}.")
```
namannimmo10, line 2052: I see. This is what you meant. 
More like, 
```python
                try:
                    if not element.is_SISO:
                        raise TypeError("MIMO Series/Parallel object found as the element of "
                        f"TransferFunctionMatrix at index {(row_index, col_index)}.")
                except AttributeError as err:
                    raise TypeError("Incompatible type found as the element of "
                        f"TransferFunctionMatrix at index {(row_index, col_index)}.") from err
```

oscarbenjamin: We shouldn't catch AttributeError. AttributeError is usually raised because of bugs in the code. Instead just use `getattr`:
```python
if getattr(element, 'is_SISO', False):
    ...
```
Why are you checking if this attribute exists anyway?

It would be better to check the type of the arguments and then ensure that all acceptable types have the `is_SISO` attribute:
```python
rows_list = arg
var = rows_list[0][0]  # Can there be zero rows or columns?

values = set.union(*rows_list)
if not all(isinstance(v, SomeCls) for v in values):
    raise TypeError("Inputs should be SomeCls")
if not all(v.is_SISO for v in values):
    raise TypeError("Inputs should be SISO")

expr_mat = ImmutableMatrix([[v.to_expr() for v in row] for row in rows_list])
```
As I mentioned somewhere else there should be a class `SomeCls` that you can test for here.

sympy/physics/control/lti.py
akshanshbhatt: > Let's pass `use_unicode=True` in the docstrings? It's much readable for the users.

It cannot be done, in my opinion. I've already tried asking for a solution in the past -

https://github.com/sympy/sympy/pull/21556#issuecomment-862621542

I also tried to lookup for Unicode-containing `pprint` examples in docstrings, but I couldn't find one.

akshanshbhatt: > Dot should be placed slightly above; notice how it is done here:

I had already noticed that and tried to make it equivalent to the `MatMul`'s output but even after multiple attempts, couldn't get it done.

sympy/physics/control/lti.py
akshanshbhatt: @namannimmo10, should I reintroduce the `TransferFunctionExpr` class, or you prefer some other name?

sylee957: I think that TransferFunctionExpr sounds like a restrictive superclass that only applies to transfer function like instances.
We could have other objects like StateSpace (to appear), so a common name like ControlSystems could be appropriate for separating the objects from this library from the rest of the sympy.

But I don’t know if others prefer duck typing (like SISO statespace can also be recognized as something like SISO transferfunction) or they should be in a completely different and incompatible objects.

But I believe that this is fine since I think that we can have many references how these are taxonomically organized.

namannimmo10:
> I think that TransferFunctionExpr sounds like a restrictive superclass that only applies to transfer functions like instances.

I agree. I think we need a parent class for all the linear time-invariant system classes (including `TransferFunction`, `TransferFunctionMatrix`, `StateSpace`). This class won't be used directly but contains several common attributes and methods from all subclasses. Methods for interconversion can also be added to this. Plus, we could also make support for a timebase (dt) that could be used for discrete-time systems later on.

Python Control has [LTI superclass](https://github.com/python-control/python-control/blob/da27d3572c016750ce4776d0379aec72f433d54c/control/lti.py#L24) for this purpose. Perhaps, if the name looks terse, we could instead name it `LinearTimeInvariant`.

SciPy also has `LinearTimeInvariant`, `lti`, and `dlti` classes: https://github.com/scipy/scipy/blob/d1a683f5716e50ef84a61473518c15deb4a4d9f7/scipy/signal/ltisys.py#L47
Although, I believe we don't need three separate classes for this purpose.

sympy/physics/control/lti.py
namannimmo10: > I also tried to lookup for Unicode-containing pprint examples in docstrings, but I couldn't find one.

```
$ git grep ">>> pprint(" | grep "use_unicode=True)"
doc/src/modules/rewriting.rst:    >>> pprint(cse(sqrt(sin(x))), use_unicode=True)
doc/src/modules/rewriting.rst:    >>> pprint(cse(sqrt(sin(x)+5)*sqrt(sin(x)+4)), use_unicode=True)
doc/src/modules/rewriting.rst:    >>> pprint(cse((x-y)*(z-y) + sqrt((x-y)*(z-y))), use_unicode=True)
doc/src/modules/solvers/solveset.rst:   >>> pprint(ImageSet(Lambda(n, 2*pi*n), S.Integers), use_unicode=True)
doc/src/modules/solvers/solveset.rst:   >>> pprint(ComplexRegion(FiniteSet(1)*Interval(0, 2*pi), polar=True), use_unicode=True)
doc/src/modules/solvers/solveset.rst:   >>> pprint(ComplexRegion(Interval(-oo, oo)*Interval(0, oo)), use_unicode=True)
doc/src/modules/solvers/solveset.rst:    >>> pprint(solveset(abs(x) - n, x, domain=S.Reals), use_unicode=True)
doc/src/tutorial/printing.rst:    >>> pprint(Integral(sqrt(1/x), x), use_unicode=True)
sympy/physics/optics/polarization.py:>>> pprint(x0, use_unicode=True)
sympy/physics/optics/polarization.py:>>> pprint(s0, use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(jones_vector(psi, chi), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(jones_vector(0, 0), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(jones_vector(pi/2, 0), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(jones_vector(pi/4, 0), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(jones_vector(-pi/4, 0), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(jones_vector(0, pi/4), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(jones_vector(0, -pi/4), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(stokes_vector(psi, chi, p, I), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(stokes_vector(0, 0), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(stokes_vector(pi/2, 0), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(stokes_vector(pi/4, 0), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(stokes_vector(-pi/4, 0), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(stokes_vector(0, pi/4), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(stokes_vector(0, -pi/4), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(stokes_vector(0, 0, 0), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(J, use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(R, use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(HWP, use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(QWP, use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(NDF, use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(reflective_filter(R), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(mueller_matrix(linear_polarizer(theta)), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(mueller_matrix(half_wave_retarder(theta)), use_unicode=True)
sympy/physics/optics/polarization.py:    >>> pprint(mueller_matrix(quarter_wave_retarder(theta)), use_unicode=True)
sympy/physics/wigner.py:    >>> pprint(wigner_d_small(half, beta), use_unicode=True)
sympy/physics/wigner.py:    >>> pprint(wigner_d_small(2*half, beta), use_unicode=True)
sympy/physics/wigner.py:    >>> pprint(wigner_d_small(half, beta).subs({beta:pi/2}), use_unicode=True)
sympy/physics/wigner.py:    >>> pprint(wigner_d(half, alpha, beta, gamma), use_unicode=True)
```

oscarbenjamin: There needs to be a decision about using unicode in docstrings. It is deliberately not used in most of the codebase even if there are some exceptions.

eric-wieser: The unicode test seems to be entirely based around conventions from python 2:

https://github.com/sympy/sympy/blob/cc67e93c1e05a46fb3d5c4f445086aa89dc09353/sympy/testing/quality_unicode.py#L5-L28

sympy/physics/control/lti.py
```diff
@@ -748,20 +824,29 @@ class Series(Basic):
...
-                raise ValueError("All transfer functions should use the same complex"
-                    " variable of the Laplace transform.")
-        if evaluate:
-            return obj.doit()
-        return obj
+        try:
+            type_set = {arg.is_SISO for arg in args}
+            if len(type_set) != 1:
+                raise TypeError("Invalid system. `args` passed contain SISO as well "
+                    "as MIMO elements.")
```
eric-wieser, line 831: This will falsely claim that the args contain a mixture when `args=[]`

sympy/physics/control/lti.py
```diff
@@ -894,11 +1001,18 @@ def __truediv__(self, other):
...
+        arg_list[0] = -arg_list[0]
+        return Series(*arg_list)
 
     def to_expr(self):
         """Returns the equivalent ``Expr`` object."""
-        return Mul(*(arg.to_expr() for arg in self.args), evaluate=False)
+        if self.is_SISO:
+            return Mul(*(arg.to_expr() for arg in self.args), evaluate=False)
+        else:
+            TypeError("`to_expr` method is only valid for SISO Series.")
```
eric-wieser, line 1015:
```suggestion
            raise TypeError("`to_expr` method is only valid for SISO Series.")
```

sympy/physics/control/lti.py
sylee957: I encourage not to use unicode in source files and docstrings because they need special monospace fonts to look not misaligned.

sympy/printing/latex.py
akshanshbhatt: **Update:**

Latex

image

Pretty-Print

```
                                       [   3      ]   
                                       [5*s    s  ]   
[5]                                    [----   -  ]   
[-]    /[      3]      [      3]   \   [ 1     1  ]   
[1]    |[5  6*s ]      [5  6*s ]   |   [          ]   
[ ]   *|[-  ----]    + [-  ----]   | + [         2]   
[5]    \[1   1  ]{t}   [1   1  ]{t}/   [ 5    5*s ]   
[-]                                    [ -    ----]   
[1]{t}                                 [ 1     1  ]{t}
```

namannimmo10:
@akshanshbhatt, please add unit tests covering all the functionality and corner cases. The examples in the docstrings look fine overall. Completing the tests will verify that the implementation is correct.

Perhaps @oscarbenjamin could comment on this added superclass.

sympy/physics/control/lti.py
```diff
@@ -1079,85 +1314,86 @@ def doit(self, **kwargs):
...
+
+        def _MIMO_doit():
+            _arg = (arg.doit()._expr_mat for arg in self.args)
+            res = MatAdd(*_arg, evaluate=True)
+            return TransferFunctionMatrix.from_Matrix(res, self.var)
+
+        if self.is_SISO:
+            return _SISO_doit()
+        else:
+            return _MIMO_doit()
```
namannimmo10, line 1338: More like,
```python
        def _SISO_doit():
            _arg = (arg.doit().to_expr() for arg in self.args)
            res = Add(*_arg, evaluate=False)
            return TransferFunction.from_rational_expression(res, self.var)

        def _MIMO_doit():
            _arg = (arg.doit()._expr_mat for arg in self.args)
            res = MatAdd(*_arg, evaluate=False)
            return TransferFunctionMatrix.from_Matrix(res, self.var)
        
        return _SISO_doit() if self.is_SISO else _MIMO_doit()
```
Similarly, pass `evaluate=False` in `Series` class.

sympy/physics/control/lti.py
```diff
@@ -20,8 +20,42 @@ def _roots(poly, var):
...
+            if self.num != 1:
+                return Mul(self.num, Pow(self.den, -1, evaluate=False), evaluate=False)
+            else:
+                return Pow(self.den, -1, evaluate=False)
+        else:
+            return ValueError("Cannot convert this  LTI system to expr.")
 
-class TransferFunction(Basic, EvalfMixin):
+
+class TransferFunction(Basic, EvalfMixin, LTICommon):
```
oscarbenjamin, line 58: Why not just make LTICommon a subclass of Basic and EvalfMixin?

sympy/physics/control/lti.py
```diff
@@ -20,8 +20,42 @@ def _roots(poly, var):
         r = [rootof(poly, var, k) for k in range(n)]
     return r
 
+class LTICommon:
+    """A common class for all the Linear Time-Invariant Dynamical Systems."""
```
oscarbenjamin, line 24: I wonder if a better name is possible than `LTICommon`. Ideally the name would be usable in the docs to describe what is allowed as e.g. input to a particular function.

sympy/physics/control/lti.py
```diff
@@ -20,8 +20,42 @@ def _roots(poly, var):
...
+    # Users should not directly interact with this class.
+    def __new__(cls, system, **kwargs):
+        if cls is LTICommon:
+            raise NotImplementedError('The LTICommon class is not meant to be used directly.')
+        obj = super(LTICommon, cls).__new__(cls)
+        obj.__var = system.var
+        obj.__num_inputs = getattr(system, 'num_inputs', None)
+        obj.__num_outputs = getattr(system, 'num_outputs', None)
+        return obj
```
oscarbenjamin, line 33: Avoid using double underscore private variables in Python as a general rule.

Avoid having any attributes in a Basic subclass. Presumably these private attributes are not used anywhere if they aren't used in this class so maybe just get rid of them.

The base classes should be passing their args to this class but it calls `super.__new__` without any args.

sympy/physics/control/lti.py
oscarbenjamin: This docstring could describe what properties or methods can be expected to be provided by any instance of LTICommon

sympy/physics/control/lti.py
```diff
@@ -20,8 +20,42 @@ def _roots(poly, var):
...
+    def _to_expr(self):
+        """Returns the equivalent ``Expr`` object."""
+        if not self.is_SISO:
+            raise TypeError("`to_expr` method is only valid for SISO Systems.")
+
+        if isinstance(self, Series):
+            return Mul(*(arg.to_expr() for arg in self.args), evaluate=False)
+        elif isinstance(self, Parallel):
+            return Add(*(arg.to_expr() for arg in self.args), evaluate=False)
+        elif isinstance(self, TransferFunction):
```
oscarbenjamin, line 49: Special-casing on subclasses is not really how these things usually work: just implement this method on each subclass.

sympy/physics/control/lti.py
akshanshbhatt: @namannimmo10 
I don't get why `evaluate` should be `False`. If we pass `evaluate=False` there is no purpose of having `doit` as it will return the same output as input.

```
>>> pprint(Series(tfm_c, tfm_b, tfm_a), use_unicode=False)
[5*s]                 [1  s]
[---]    [5   1  ]    [-  -]
[ 1 ]    [-  ----]    [1  1]
[   ]   *[1     2]   *[    ]
[ 5 ]    [   6*s ]{t} [5  1]
[ - ]                 [-  -]
[ 1 ]{t}              [s  1]{t}

File "d:\sympy\sympy\physics\control\lti.py", line 812, in sympy.physics.control.lti.Series
Failed example:
    pprint(Series(tfm_c, tfm_b, tfm_a).doit(), use_unicode=False)  # (2 Inputs -A-> 2 Outputs) -> (2 Inputs -B-> 1 Output) -> (1 Input -C-> 2 Outputs) is equivalent to (2 Inputs -Series Equivalent-> 2 Outputs).
Expected:
    [   /   3    \    /    3    \]
    [25*\6*s  + 1/  5*\30*s  + 1/]
    [-------------  -------------]
    [        2           6*s     ]
    [     6*s                    ]
    [                            ]
    [   /   3    \    /    3    \]
    [25*\6*s  + 1/  5*\30*s  + 1/]
    [-------------  -------------]
    [        3              2    ]
    [     6*s            6*s     ]{t}
Got:
    [5*s]                 [1  s]
    [---]    [5   1  ]    [-  -]
    [ 1 ]    [-  ----]    [1  1]
    [   ]   *[1     2]   *[    ]
    [ 5 ]    [   6*s ]{t} [5  1]
    [ - ]                 [-  -]
    [ 1 ]{t}              [s  1]{t}
```

namannimmo10: You're right!

namannimmo10:
`.evalf()`, `simplify()`, `expand()`, etc., should work on a Transfer function matrix.

sympy/physics/control/tests/test_lti.py
```diff
@@ -524,6 +567,55 @@ def test_Series_functions():
...
+    # Transfer function matrix in the arguments.
+    assert Series(tfm2, tfm1, evaluate=True) == Series(tfm2, tfm1).doit() == \
+        TransferFunctionMatrix(((TransferFunction(-k**2*(a2*s + p)**2*(s**2 + 2*s*wn*zeta + wn**2)**2 + (-a2*p + s)*(a2*p - s)*(s**2 + 2*s*wn*zeta + wn**2)**2 - \
+            (a2*s + p)**2, (a2*s + p)**2*(s**2 + 2*s*wn*zeta + wn**2)**2, s),), (TransferFunction(k**2, 1, s),)))
+
+    # calling doit() once will not expand the internal Series and Parallel objects.
+    assert Series(-tfm3, -tfm2, tfm1, evaluate=True) == Series(-tfm3, -tfm2, tfm1).doit() == \
+        TransferFunctionMatrix(((TransferFunction(k**2*(a2*s + p)**2*(s**2 + 2*s*wn*zeta + wn**2)**2 + (a2*p - s)**2*(s**2 + 2*s*wn*zeta + wn**2)**2 + (a2*s + p)**2, \
+            (a2*s + p)**2*(s**2 + 2*s*wn*zeta + wn**2)**3, s),), (TransferFunction(-k**2*(a2*s + p)*(s**2 + 2*s*wn*zeta + wn**2), (a2*s + p)*(s**2 + 2*s*wn*zeta + wn**2)**2, s),)))
+
```
namannimmo10, line 617: Add one test with `Parallel` instance as the argument and call `.rewrite(TransferFunctionMatrix)`?

sympy/physics/control/tests/test_lti.py
```diff
@@ -462,6 +462,49 @@ def test_Series_construction():
...
+    # all the TFMs must use the same complex variable.
+    raises(ValueError, lambda: Series(tfm_3, tfm_5))
+
+    # Number or expression not allowed in the arguments.
+    raises(TypeError, lambda: Series(2, tfm_2, tfm_3))
+    raises(TypeError, lambda: Series(s**2 + p*s, -tfm_2, tfm_3))
+
+
 def test_Series_functions():
     zeta, wn = symbols('zeta, wn')
```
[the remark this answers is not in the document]
namannimmo10: This applies to all such declarations.

sympy/physics/control/lti.py
```diff
@@ -802,80 +918,72 @@ def doit(self, **kwargs):
...
-        for arg in self.args:
-            arg = arg.doit()
-            if res is None:
-                res = arg
-            else:
-                num_ = arg.num * res.num
-                den_ = arg.den * res.den
-                res = TransferFunction(num_, den_, self.var)
-        return res
+        def _SISO_doit():
```
namannimmo10, line 921: Whichever method you modify in Series and Parallel, update the docstring and add one example related to MIMO TF.

akshanshbhatt: I had modified the tests for this particular change.

sympy/physics/control/lti.py
```diff
@@ -697,16 +713,54 @@ def to_expr(self):
...
-    """
-    A class for representing product of transfer functions or transfer functions in a
-    series configuration.
+def _mat_mul_compatible(*args):
+    """To check whether shapes are compatible for matrix mul."""
+    return all(
+        arg.num_outputs == args[i].num_inputs
+        for i, arg in enumerate(args, start=1)
+        if i < len(args)
+    )
```
oscarbenjamin, line 722: This use of enumerate is probably more confusing then just using an integer index like:
```
return all(args[i].num_outputs == args[i+1].num_inputs for i in range(len(args)-1))
```
Otherwise `zip` is the tool to use rather than enumerate:
```
pairs = zip(args[:-1], args[1:])
return all(a.num_outputs == b.num_inputs for a, b in pairs)
```

sympy/physics/control/lti.py
```diff
@@ -748,20 +838,32 @@ class Series(Basic):
...
+            type_set = {arg.is_SISO for arg in args}
+            if len(type_set) != 1:
+                raise TypeError("Invalid system. `args` passed contain SISO as well "
+                    "as MIMO elements.")
+            var_set = {arg.var for arg in args}
+            if len(var_set) != 1:
+                raise ValueError("All transfer functions should use the same complex variable"
+                    f" of the Laplace transform. {len(var_set)} different values found.")
+        except AttributeError:
+            raise TypeError("Unsupported type of argument passed.")
```
oscarbenjamin, line 854: AttributeError should not be caught. Why is it necessary to catch this?

Whatever raises the error can be handled in a better way.

sympy/physics/control/lti.py
```diff
@@ -782,7 +884,21 @@ def var(self):
         p
 
         """
-        return self._var
+        return self.args[0].var
+
+    @property
+    def num_inputs(self):
+        return getattr(self.args[0], "num_inputs", None)
```
oscarbenjamin, line 891: Why would `self.args[0]` sometimes be an object that has this attribute and sometimes not?

That suggests a messy design somewhere.

sympy/physics/control/lti.py
```diff
@@ -802,80 +918,72 @@ def doit(self, **kwargs):
...
     def __add__(self, other):
-        if isinstance(other, (TransferFunction, Series)):
-            if not self.var == other.var:
-                raise ValueError("All the transfer functions should use the same complex variable "
-                    "of the Laplace transform.")
+        try:
+            if self.is_SISO != other.is_SISO:
+                raise TypeError("Cannot add SISO and MIMO systems together.")
+        except AttributeError:
+            raise ValueError(f"Cannot add Series type with {type(other)}.")
```
oscarbenjamin, line 953: Don't catch `AttributeError`. An `__add__` method should not raise if the argument is not of the expected type. Rather it should return `NotImplemented` which will lead to `TypeError` if the other object also returns `NotImplemented`.

oscarbenjamin:
It seems like a lot of methods do something like:
```
if self.is_SISO:
    do_SISO()
else:
    do_MIMO()
```
Or even:
```
return something() if self.is_SISO else None
```
There are also various mentions in docstrings that some methods are only applicable for one of SISO or MIMO.

That makes me wonder if it would be a better design for SISO and MIMO to be handled as separate classes. I guess this is basically like mixing scalars and matrices and sympy ended up having separate classes for those e.g. Add vs MatAdd.

akshanshbhatt:
> That makes me wonder if it would be a better design for SISO and MIMO to be handled as separate classes. I guess this is basically like mixing scalars and matrices and sympy ended up having separate classes for those e.g. Add vs MatAdd.

If we handle them as a separate class, then that would require two different classes for each `Series` and `Parallel` (`MIMOSeries` and `MIMOParallel` with their SISO counterparts), and maybe `Feedback` too. These are merely configurations (and not operations like `Add` and `Mul`). It might cause inconvenience for the user to check whether the systems are in a `MIMOParallel` configuration or `SISOParallel` (an example).
@namannimmo10 if you have something to say?

namannimmo10:
IMO, we do not need separate classes. While doing the SISO or MIMO Transfer function algebra, the user need not worry about the classes that are being used for series or parallel configurations of Transfer functions (or matrices). They will do this like it is normally done on paper; 
```
G_1 = TransferFunction(num_1, den_1, s)
G_2 = TransferFunction(num_2, den_2, s)
G_3 = TransferFunction(num_3, den_3, s)
G_4 = TransferFunction(num_4, den_4, s)
pprint(G_1 * G_2) # SISO in series configuration

tfm_1 = TransferFunctionMatrix([[G_1, G_2], [G_3, G_4]]) # (2, 2)
tfm_2 = TransferFunctionMatrix([[-G_3, -G_2], [-G1, -G4]]) # (2, 2)
pprint(tfm_1 * tfm_2)
pprint(tfm_2 * tfm_1) # MIMO in series configuration
pprint(tfm_1 + tfm_2) # MIMO in parallel configuration

tfm_feedback = Feedback(tfm_1, tfm_2, type='neg')
pprint(tfm_feedback.doit())
```
As you can see, the user doesn't actually deal with `Series` or `Parallel` directly.
Instead, we should give time to other features/classes that would be really useful while solving problems.

eric-wieser:
> As you can see, the user doesn't actually deal with `Series` or `Parallel` directly.

To me this sounds like an argument _for_ separate classes not against them. The implementation avoids a lot of `if` statements with separate classes, and your example shows that the simpler implementation doesn't impact the user negatively, as it's a hidden implementation detail.

oscarbenjamin:
Alternatively would it make sense to say that a transfer function is just a 1x1 transfer function matrix? Then we don't need separate classes and a transfer function would have `num_inputs = 1` and is_SISO would just be:
```python
@property
def is_SISO(self):
    return self.shape == (1, 1)
```

In the context of control theory what exactly does the distinction between a 1x1 matrix and a scalar represent? Algebraically the distinction between scalars and matrices is about how multiplication is handled with scalar-matrix multiplication being different from matrix-matrix multiplication.

My understanding of all of these operations like series, parallel etc is that what the user is doing is assembling a dynamical system through a block diagram representation as in e.g. Simulink. In that context every block with a transfer function will have an input and an output: it would be SISO if it had one input and one output but that could naturally be represented in the framework as a 1x1 matrix.

namannimmo10:
Yes, our SISO system could function as a transfer function matrix object with 1x1 dimensions.

Albeit, a solid SISO transfer function object should be used internally in the MIMO transfer function matrix without duplicating functionality and thus providing both a _simple_ SISO user interface (that would map nicely to introductory control textbooks) and the more complex MIMO interface.

sympy/physics/control/lti.py
namannimmo10: > I had modified the tests for this particular change.

I don't get it.  `doit()` method in `Series` and `Parallel` should include at least one example for MIMO system.

namannimmo10:
@akshanshbhatt, how much time do you need to address the comments? We also need to deal with the `Feedback` class after this.

akshanshbhatt:
> @akshanshbhatt, how much time do you need to address the comments?

I am not able to think of an efficient way to add separate SISO and MIMO classes. If either you or @oscarbenjamin could suggest the API of these classes.
This PR is hopefully complete after that.

sympy/physics/control/lti.py
akshanshbhatt: Done.

namannimmo10:
> Alternatively would it make sense to say that a transfer function is just a 1x1 transfer function matrix? Then we don't need separate classes and a transfer function would have `num_inputs = 1` and is_SISO would just be:
> 
> ```python
> @property
> def is_SISO(self):
>     return self.shape == (1, 1)
> ```
@oscarbenjamin, can you give some more details on what ideas you have in your mind about the implementation?
If the transfer function is represented as a 1x1 transfer function matrix, then there's also a possibility that we could connect SISO with MIMO systems in series and parallel configurations.

oscarbenjamin:
> @oscarbenjamin, can you give some more details on what ideas you have in your mind about the implementation?

I don't have a detailed idea in mind. It seems clear to me from the code that something isn't quite right in the current design though. I've suggested two approaches that seem cleaner to me:

1. Have separate MIMO and SISO classes for Series and Parallel etc.
2. Unify MIMO and SISO so that an SISO is just an MIMO with one input and one output.

It's not clear to me why there needs to be a distinction between a TransferFunction and a 1x1 TransferFunctionMatrix. If that distinction is important then maybe the same class Series should not be used in both cases.

namannimmo10:
I would advocate for the separation of SISO and MIMO because of the following reasons.

*  As I already mentioned, a solid SISO transfer function (`TransferFunction`) should be used inside the MIMO transfer function (`TransferFunctionMatrix`) so that we provide both a simple SISO user interface and the more complex MIMO interface.
*  If, for instance, you take a look at [`TransferFunctionModel`](https://reference.wolfram.com/language/ref/TransferFunctionModel.html) in Mathematica (go under the Applications section) you would see that most applications involve the usage of a SISO transfer function.
* Separation of SISO and MIMO are done in most, if not all, other control software (_at least from a user's perspective_), so it will be easy for people to pick up ours.
* Most use cases for SymPy control will likely be elementary linear controls as symbolic become less useful for more complicated systems. A SISO transfer function object will map nicely to introductory control textbooks.
* `TransferFunction` had been included in the previous SymPy release. . .so I would not completely change/discard that.

We can have `SISOSeries`, `SISOParallel`, `MIMOSeries`, and `MIMOParallel` classes for a better design and to make the arithmetic code less messy.

akshanshbhatt:
> We can have `SISOSeries`, `SISOParallel`, `MIMOSeries`, and `MIMOParallel` classes for a better design and to make the arithmetic code less messy.

Of the two options given, this seemed better to me too. I'll try to implement it asap and get this merged.

akshanshbhatt:
@namannimmo10 just wanted to confirm that the usual `Series` and `Parallel` classes should be deprecated, right?

sympy/physics/control/lti.py
```diff
@@ -1099,7 +1004,222 @@ def is_biproper(self):
...
+
+        >>> from sympy.abc import p
+        >>> from sympy.physics.control.lti import TransferFunction, Series, Parallel
+        >>> G1 = TransferFunction(p**2 + 2*p + 4, p - 6, p)
+        >>> G2 = TransferFunction(p, 4 - p, p)
+        >>> G3 = TransferFunction(0, p**4 - 1, p)
+        >>> Series(G1, G2).var
+        p
+        >>> Series(-G3, Parallel(G1, G2)).var
+        p
```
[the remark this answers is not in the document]
namannimmo10: Similarly, update the docstring in `MIMOParallel`.

sympy/physics/control/lti.py
```diff
@@ -1099,7 +1004,222 @@ def is_biproper(self):
...
+    def num_outputs(self):
+        return self.args[-1].num_outputs
+
+    @property
+    def shape(self):
+        return self.args[-1].num_outputs, self.args[0].num_inputs
+
+    def doit(self, **kwargs):
+        """
+        Returns the resultant transfer function obtained after evaluating
```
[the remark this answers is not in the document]
namannimmo10: Similarly, update the docstring in `MIMOParallel`

sympy/physics/control/lti.py
```diff
@@ -1458,7 +1491,216 @@ def is_biproper(self):
...
+
+        """
+        _arg = (arg.doit()._expr_mat for arg in self.args)
+        res = MatAdd(*_arg, evaluate=True)
+        return TransferFunctionMatrix.from_Matrix(res, self.var)
+
+    def _eval_rewrite_as_TransferFunctionMatrix(self, *args, **kwargs):
+        if self.is_SISO:
+            raise ValueError("Only transfer function matrices or a collection of transfer function"
+                " matrices is allowed in the arguments.")
```
namannimmo10, line 1662:
```suggestion
```

sympy/physics/control/lti.py
```diff
@@ -1458,7 +1491,216 @@ def is_biproper(self):
...
+            raise TypeError("All arguments must be of type LinearTimeInvariant.")
+
+        type_set = {arg.is_SISO for arg in args}
+        if len(type_set) != 1:
+            raise TypeError("Invalid system. `args` passed contain SISO as well "
+                "as MIMO elements.")
+        var_set = {arg.var for arg in args}
+        if len(var_set) != 1:
+            raise ValueError("All MIMO systems should use the same complex variable"
+                f" of the Laplace transform. {len(var_set)} different values found.")
```
namannimmo10, line 1592: I see this code block getting repeated in these four classes -- let's make a helper function which will check the `args` passed.

namannimmo10 (commented):
Looks good. I've added some minor comments.

namannimmo10:
@akshanshbhatt, can you address https://github.com/sympy/sympy/pull/21703#issuecomment-879328717 in this PR?

```diff
$ git diff
diff --git a/sympy/physics/control/lti.py b/sympy/physics/control/lti.py
index 21e7d1446f..86991b253b 100644
--- a/sympy/physics/control/lti.py
+++ b/sympy/physics/control/lti.py
@@ -2272,18 +2272,18 @@ def __new__(cls, arg):
         expr_mat_arg = []
         try:
             var = arg[0][0].var
-        except TypeError:
+        except TypeError as err:
             raise ValueError("`arg` param in TransferFunctionMatrix should "
-            "strictly be a nested list containing TransferFunction objects.")
-        for row_index, row in enumerate(arg):
+            "strictly be a nested list containing TransferFunction objects.") from err
+        for _, row in enumerate(arg):
             temp = []
-            for col_index, element in enumerate(row):
+            for _, element in enumerate(row):
                 if not isinstance(element, LinearTimeInvariant):
                     raise TypeError("Incompatible type found as the element of "
                         "TransferFunctionMatrix.")
 
                 if not element.is_SISO:
-                        raise TypeError("MIMO Series/Parallel object found as the element of "
+                    raise TypeError("MIMO Series/Parallel object found as the element of "
                         "TransferFunctionMatrix. Each element is expected to be a SISO LTI system.")
 
                 if var != element.var:
@@ -2548,3 +2548,8 @@ def elem_zeros(self):
     def _flat(self):
         """Returns flattened list of args in TransferFunctionMatrix"""
         return [elem for tup in self.args[0] for elem in tup]
+
+    def _eval_evalf(self, prec):
+        """Calls evalf() on each transfer function in the transfer function matrix"""
+        mat = self._expr_mat.evalf(prec)
+        return _to_TFM(mat, self.var)
```
```python
>>> G_1 = TransferFunction(2*s + 1, s**2 - 1, s)
>>> G_2 = TransferFunction(4, s**2 + 1, s)
>>> pprint(TransferFunctionMatrix([[G_1, G_2]]))
⎡2⋅s + 1    4   ⎤ 
⎢───────  ──────⎥ 
⎢  2       2    ⎥ 
⎣ s  - 1  s  + 1⎦τ
>>> pprint(TransferFunctionMatrix([[G_1, G_2]]).evalf())
⎡2.0⋅s + 1.0    4.0   ⎤ 
⎢───────────  ────────⎥ 
⎢   2          2      ⎥ 
⎣  s  - 1.0   s  + 1.0⎦τ
```

sympy/physics/control/lti.py
```diff
@@ -744,24 +783,32 @@ class Series(Basic):
...
+        var_set = {arg.var for arg in args}
+        if len(var_set) != 1:
+            raise ValueError("All transfer functions should use the same complex variable"
+                f" of the Laplace transform. {len(var_set)} different values found.")
+
+        if list(type_set)[0]:
+            obj = super().__new__(cls, *args)
+            obj._is_SISO = True
+        else:
+            raise TypeError("MIMO systems are not allowed in `Series`. Use `MIMOSeries` instead.")
```
eric-wieser, line 809: I don't really see the point in checking for a mix of MIMO and SISO, then checking for any MIMO. Why not just check SISO from the beginning?
```suggestion
        if not all(arg.is_SISO for arg in args):
            raise TypeError("MIMO systems are not allowed in `Series`. Use `MIMOSeries` instead.")

        var_set = {arg.var for arg in args}
        if len(var_set) != 1:
            raise ValueError("All transfer functions should use the same complex variable"
                f" of the Laplace transform. {len(var_set)} different values found.")

        obj = super().__new__(cls, *args)
        obj._is_SISO = True
```

sympy/physics/control/lti.py
```diff
@@ -976,10 +1007,247 @@ def is_biproper(self):
...
+
+    def __neg__(self):
+        arg_list = list(self.args)
+        arg_list[0] = -arg_list[0]
+        return MIMOSeries(*arg_list)
+
+
+class Parallel(LinearTimeInvariant):
+    r"""
+    A class for representing parallel configuration of SISO and MIMO transfer functions.
```
eric-wieser, line 1227:
```suggestion
    A class for representing parallel configuration of SISO transfer functions.
```
this one doesn't support MIMO, right?

sympy/physics/control/lti.py
```diff
@@ -221,6 +237,7 @@ def __new__(cls, num, den, var):
             obj._num = num
             obj._den = den
             obj._var = var
+            obj._is_SISO = True
```
oscarbenjamin, line 240: I guess this could be a class attribute

sympy/physics/control/lti.py
oscarbenjamin: There are lots of places that do this:
```python
if not all(isinstance(arg, LinearTimeInvariant) for arg in args):
    raise ...

if not all(arg.is_SISO for arg in args):
    raise ...

var_set = {arg.var for arg in args}
if len(var_set) != 1:
    raise ...
```
Now that SISO and MIMO are always distinct classes it would be possible to make a common superclass for SISO like `SISOLinearTimeInvariant`. Then the first two tests could be combined like:
```python
if not all(isinstance(arg, SISOLinearTimeInvariant) for arg in args):
    raise...
```
Also since this checking is quite common it should be factored out into a helper in the super class so it looks like this:
```python
class LinearTimeInvariant:

    @classmethod
    def _check_args(cls, args):
        if not all(isinstance(arg, cls._clstype) for arg in args):
            raise ...        
        var_set = {arg.var for arg in args}
        if len(var_set) != 1:
            raise ...
        (var,) = var_set
        return var  # does the caller need this?

    @property
    def is_SISO(self):
        return self._is_SISO


class SISOLinearTimeInvariant(LinearTimeInvariant):
    _is_SISO = True


class MIMOLinearTimeInvariant(LinearTimeInvariant
    _is_SISO = False


SISOLinearTimeInvariant._clstype = SISOLinearTimeInvariant
MIMOLinearTimeInvariant._clstype = MIMOLinearTimeInvariant


class Series(SISOLinearTimeInvariant):

    def __new__(cls, *args):
        if not args:
            raise ... # maybe this could be handled in _check_args as well?
        var = cls._check_args(args)
        return super().__new__(cls, *args)
```

sympy/physics/control/lti.py
```diff
@@ -802,77 +849,49 @@ def doit(self, **kwargs):
...
-                res = arg
-            else:
-                num_ = arg.num * res.num
-                den_ = arg.den * res.den
-                res = TransferFunction(num_, den_, self.var)
-        return res
+
+        _arg = (arg.doit().to_expr() for arg in self.args)
+        res = Mul(*_arg, evaluate=True)
+        return TransferFunction.from_rational_expression(res, self.var)
```
oscarbenjamin, line 855: Using `Mul` here could lead to pole-zero cancellation. Is that intended?

sympy/physics/control/lti.py
```diff
@@ -802,77 +849,49 @@ def doit(self, **kwargs):
...
+        if not other.is_SISO:
+            raise TypeError("Cannot add SISO and MIMO systems together.")
 
+        if isinstance(other, Parallel):
+            arg_list = list(other.args)
             return Parallel(self, *arg_list)
-        else:
-            raise ValueError("This transfer function expression is invalid.")
+
+        return Parallel(self, other)
```
oscarbenjamin, line 871: Maybe this flattening should be handled in `Parallel.__new__`. Are these supposed to be equivalent:
```python
tf1 + tf2 + tf3
Parallel(tf1, Parallel(tf2, tf3))
```

sympy/physics/control/lti.py
```diff
@@ -802,77 +849,49 @@ def doit(self, **kwargs):
...
-        elif isinstance(other, Parallel):
-            if not self.var == other.var:
-                raise ValueError("All the transfer functions should use the same complex variable "
-                    "of the Laplace transform.")
-            arg_list = [-i for i in list(other.args)]
-
-            return Parallel(self, *arg_list)
-        else:
-            raise ValueError("This transfer function expression is invalid.")
+        return self + (-other)
```
oscarbenjamin, line 876: This should return `NotImplemented` if `other` is not a recognised type.

sympy/physics/control/lti.py
```diff
@@ -802,77 +849,49 @@ def doit(self, **kwargs):
...
 
     def __rsub__(self, other):
         return -self + other
```
oscarbenjamin, line 879: This should check the type of other and return `NotImplemented` if other is not a recognised type.

sympy/physics/control/lti.py
```diff
@@ -907,6 +926,10 @@ def is_proper(self):
         function is less than or equal to degree of the denominator polynomial of
         the same, else False.
 
+        .. note::
+            ``is_proper`` is defined only for SISO ``Series`` configuration.
+            ``Series(*args).is_proper`` for MIMO ``*args`` will return ``None``.
```
oscarbenjamin, line 931: Is this still correct?

sympy/physics/control/lti.py
```diff
@@ -1099,7 +1004,222 @@ def is_biproper(self):
...
+
+class MIMOSeries(LinearTimeInvariant):
+    r"""
+    A class for representing series configuration of MIMO systems.
+
+    Parameters
+    ==========
+
+    args : TransferFunctionMatrix, MIMOSeries, MIMOParallel
+        MIMO Systems in series configuration.
```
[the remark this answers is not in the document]
oscarbenjamin: If there was a common superclass for MIMO then it could mentioned as the type here rather than listing all the particular classes.

sympy/physics/control/tests/test_lti.py
```diff
@@ -641,8 +789,58 @@ def test_Parallel_functions():
...
+    # (tfm3 - tfm2) has (3, 1) shape while tfm4*tfm5 has (2, 2) shape.
+    raises(ValueError, lambda: (tfm3 - tfm2)*tfm4*tfm5)
+    # (tfm1 - tfm2) has (3, 1) shape while tfm5 has (2, 2) shape.
+    raises(ValueError, lambda: (tfm1 - tfm2)*tfm5)
+
+    # TFM in the arguments.
+    assert MIMOParallel(tfm1, tfm2, evaluate=True) == MIMOParallel(tfm1, tfm2).doit() == MIMOParallel(tfm1, tfm2).rewrite(TransferFunctionMatrix) == \
+        TransferFunctionMatrix(((TransferFunction(-k*(s**2 + 2*s*wn*zeta + wn**2) + 1, s**2 + 2*s*wn*zeta + wn**2, s),), \
+            (TransferFunction(-a0 + a1*s**2 + a2*s + k*(a0 + s), a0 + s, s),), (TransferFunction(-a2*s - p + (a2*p - s)* \
+                (s**2 + 2*s*wn*zeta + wn**2), (a2*s + p)*(s**2 + 2*s*wn*zeta + wn**2), s),)))
```
oscarbenjamin, line 840: Long lines should be broken. You can use `(` to extend them e.g.:
```python
assert (asdsdfsdfsd
        == asdasdasdasd
        == asdasdasdasdasd)
```
Is there not a less verbose way of creating a `TransferFunctionMatrix`?

sympy/printing/latex.py
```diff
@@ -2434,22 +2434,31 @@ def _print_Morphism(self, morphism):
...
-        parens = lambda x: self.parenthesize(x, precedence_traditional(expr),
-                                             False)
-        return ' '.join(map(parens, args))
+        func = lambda x: self._print(x)
+        return '+'.join(map(func, args))
+
+    def _print_MIMOParallel(self, expr):
+        args = list(expr.args)
+        func = lambda x: self._print(x)
+        return '+'.join(map(func, args))
```
oscarbenjamin, line 2461: When you find yourself using `map` with `lambda` it is usually clearer just to use a comprehension:
```python
return '+'.join(self._print(arg) for arg in expr.args)
```
Sometimes `lambda` is useful for breaking long lines (I prefer not to have a multiline comprehension) but that doesn't apply here.

Should this have spaces around the `+` like `1 + 2`?

sympy/physics/control/lti.py
akshanshbhatt: It ultimately calls `__add__(self, -other)`, and I have handled it there. So, should I do it again?

akshanshbhatt:
> @akshanshbhatt, can you address #21703 (comment) in this PR?

@namannimmo10, I will open a separate PR for this.

sympy/physics/control/lti.py
oscarbenjamin: Does this method actually return NotImplemented of other is not a recognised type?

That is what is needed for cooperative dispatch.

namannimmo10 (approved):
@akshanshbhatt: I think you have addressed almost every major comment that needed to be.
This is excellent; thanks for the hard work and adjusting to all the feedback.  :tada: 

I'll merge this later tonight if anyone else has no further comments.

namannimmo10:
Removing "`LinearTimeInvariant` is now the common super class for all the linear time-invariant systems." from the release notes because this class is not available to the users.

sympy/physics/control/lti.py
```diff
@@ -976,10 +1008,236 @@ def is_biproper(self):
...
+        >>> tf2 = TransferFunction(s**3 - 2, s**4 + 5*s + 6, s)
+        >>> tfm1 = TransferFunctionMatrix([[tf1, tf2], [tf2, tf2]])
+        >>> tfm2 = TransferFunctionMatrix([[tf2, tf1], [tf1, tf1]])
+        >>> MIMOSeries(tfm2, tfm1).doit()
+        TransferFunctionMatrix(((TransferFunction(2*(s**3 - 2)*(a*p**2 + b*s), (-p + s)*(s**4 + 5*s + 6), s), TransferFunction((-p + s)**2*(s**3 - 2)*(a*p**2 + b*s) + (-p + s)*(a*p**2 + b*s)**2*(s**4 + 5*s + 6), (-p + s)**3*(s**4 + 5*s + 6), s)), (TransferFunction((-p + s)*(s**3 - 2)**2*(s**4 + 5*s + 6) + (s**3 - 2)*(a*p**2 + b*s)*(s**4 + 5*s + 6)**2, (-p + s)*(s**4 + 5*s + 6)**3, s), TransferFunction(2*(s**3 - 2)*(a*p**2 + b*s), (-p + s)*(s**4 + 5*s + 6), s))))
+
+        """
+        _arg = (arg.doit()._expr_mat for arg in reversed(self.args))
+        res = MatMul(*_arg, evaluate=True)
+        return TransferFunctionMatrix.from_Matrix(res, self.var)
```
oscarbenjamin, line 1166: Is it possible that this can lead to pole-zero cancellation?

sympy/physics/control/lti.py
```diff
@@ -1079,78 +1328,51 @@ def doit(self, **kwargs):
...
-                res = arg
-            else:
-                num_ = res.num * arg.den + res.den * arg.num
-                den_ = res.den * arg.den
-                res = TransferFunction(num_, den_, self.var)
-        return res
+
+        _arg = (arg.doit().to_expr() for arg in self.args)
+        res = Add(*_arg).as_numer_denom()
+        return TransferFunction(*res, self.var)
```
oscarbenjamin, line 1334: Can this lead to pole-zero cancellation?

sympy/physics/control/lti.py
```diff
@@ -1941,3 +2494,18 @@ def elem_zeros(self):
     def _flat(self):
         """Returns flattened list of args in TransferFunctionMatrix"""
         return [elem for tup in self.args[0] for elem in tup]
+
+    def _eval_evalf(self, prec):
+        """Calls evalf() on each transfer function in the transfer function matrix"""
+        mat = self._expr_mat.applyfunc(lambda a: a.evalf(prec))
+        return _to_TFM(mat, self.var)
```
oscarbenjamin, line 2501: The prec passed to `_eval_evalf` is binary but you are passing it as the decimal prec to `evalf`.

sympy/physics/control/lti.py
```diff
@@ -1941,3 +2494,18 @@ def elem_zeros(self):
     def _flat(self):
         """Returns flattened list of args in TransferFunctionMatrix"""
         return [elem for tup in self.args[0] for elem in tup]
```
oscarbenjamin, line 2496: After #21626 Matrix has a `.flat` method which could be used here.

sympy/physics/control/lti.py
```diff
@@ -1867,6 +2391,35 @@ def __neg__(self):
...
+        if not isinstance(other, LinearTimeInvariant):
+             return NotImplemented
+
+        if other.is_SISO:
+            raise TypeError("Cannot add SISO and MIMO systems together.")
+
+        if not isinstance(other, MIMOParallel):
+            return MIMOParallel(self, other)
+        other_arg_list = list(other.args)
+        return MIMOParallel(self, *other_arg_list)
```
oscarbenjamin, line 2404: If flattening is happening here in `__add__` but not in `__new__` then it won't work if `Parallel` is used directly like `Parallel(tf1, Parallel(tf2, tf3))`.

Is that intentional?

sympy/physics/control/lti.py
```diff
@@ -1235,6 +1457,202 @@ def is_biproper(self):
...
+            return MIMOParallel(*self_arg_list, *other_arg_list)
+
+        self_arg_list = list(self.args)
+        return MIMOParallel(*self_arg_list, other)
+
+    __radd__ = __add__
+
+    def __sub__(self, other):
+        if not isinstance(other, MIMOLinearTimeInvariant):
+             return NotImplemented
```
oscarbenjamin, line 1632: Maybe there should be a decorator to do this check and return `NotImplemented` since it is used so many times. Already int he codebase there is `sympifyit` and `polyfyit` that do something similar

sympy/physics/control/lti.py
```diff
@@ -1235,6 +1457,202 @@ def is_biproper(self):
...
+
+    @property
+    def num_outputs(self):
+        """Returns the number of output signals of the parallel system."""
+        return self.args[0].num_outputs
+
+    @property
+    def shape(self):
+        """Returns the shape of the equivalent MIMO system."""
+        return self.args[0].num_outputs, self.args[0].num_inputs
```
oscarbenjamin, line 1586: I would just make this: `return (self.num_outputs, self.num_inputs)`.

sympy/physics/control/lti.py
```diff
@@ -1079,78 +1328,51 @@ def doit(self, **kwargs):
...
-            return Series(self, other)
-        elif isinstance(other, Series):
-            if not self.var == other.var:
-                raise ValueError("All the transfer functions should use the same complex variable "
-                    "of the Laplace transform.")
+        if not isinstance(other, LinearTimeInvariant):
+             return NotImplemented
+
+        if not other.is_SISO:
+            raise TypeError("Cannot multiply SISO and MIMO systems together.")
```
oscarbenjamin, line 1369: I guess this is analogous to multiplying a scalar and a matrix. Should that be allowed?

Also is it important to show this particular error message because this second check wouldn't be needed if the first check was for SISOLinearTimeInvariant

namannimmo10: > I guess this is analogous to multiplying a scalar and a matrix. Should that be allowed?
> 
We can make that work later on if needed.

akshanshbhatt: > I guess this is analogous to multiplying a scalar and a matrix. Should that be allowed?

We can allow that, but I don't think that it would have any physical significance.

sympy/physics/control/lti.py
akshanshbhatt: I tried a few cases where there was a possibility of pole-zero cancelation, but that didn't happen.

sympy/physics/control/lti.py
namannimmo10: Taken from [`SystemsModelParallelConnect`](https://reference.wolfram.com/language/ref/SystemsModelParallelConnect.html):

Add a disturbance model to the output of a system:
image
image
image

sympy/physics/control/lti.py
akshanshbhatt: In some cases, yes. I've fixed it and added a test case.

akshanshbhatt:
@namannimmo10 now that this PR is merged, I guess I can close #21556.

PR #21706, opened 2021-07-05
Refactor bessel solver

Mohitbalwani26:
SEE #18348 

#### Release Notes

NO ENTRY


sympy/solvers/ode/single.py
```diff
@@ -2615,5 +2615,196 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
...
+        if order==2 and r and r[a3] != 0:
+            p = cancel(r[b3]/r[a3])  # Used below
+            q = cancel(r[c3]/r[a3])
+            if p.is_zero:
+                self.rn = q.match(a4+b4*x)
+                if self.rn and self.rn[b4] != 0:
+                    self.rn = {'b':self.rn[a4],'m':self.rn[b4]}
+                    does_match = True
+        return does_match
+
```
Mohitbalwani26, line 2790: @oscarbenjamin According to my understanding about `2nd_linear_airy`, I attempted to simplify the match function something like :
```python
    def _matches(self):
        eq = self.ode_problem.eq_high_order_free
        f = self.ode_problem.func
        order = self.ode_problem.order
        x = self.ode_problem.sym
        df = f.diff(x)
        a4 = Wild('a4', exclude=[x,f,df])
        b4 = Wild('b4', exclude=[x,f,df])
        match = self.ode_problem.get_linear_coefficients(eq, f, order)
        does_match = False
        if order == 2 and match and match[2] != 0:
            if match[1].is_zero:
                self.rn = match[0].match(a4+b4*x)
                if self.rn and self.rn[b4] != 0:
                    self.rn = {'b':self.rn[a4],'m':self.rn[b4]}
                    does_match = True
        return does_match
```
It passes all the tests but wanted to ask you does this diff look ok?

oscarbenjamin: Does `get_linear_coefficients` divide by the leading coefficient?

What happens if you have something like `x y'' + x y = 0`?

Mohitbalwani26: > x y'' + x y = 0

It fails for this. I thought `self.ode_problem.eq_high_order_free` passing in the `get_linear_coefficients` would work as eq_high_order_free removes leading coefficient but it only remove `f(x)` not x.

Mohitbalwani26: if we replace `self.rn = match[0].match(a4+b4*x)` with `self.rn = cancel(match[0]/match[2]).match(a4+b4*x)` , It will solve the problem :thinking:

sympy/solvers/ode/single.py
```diff
@@ -2615,5 +2615,196 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
...
+
+    def _get_general_solution(self, *, simplify_flag: bool = True):
+        f = self.ode_problem.func.func
+        x = self.ode_problem.sym
+        (C0, C1) = self.ode_problem.get_numbered_constants(num=2)
```
asmeurer, line 2795: Shouldn't this be C1 and C2?

Mohitbalwani26:
@oscarbenjamin I have updated the match function. PTAL

oscarbenjamin:
I just noticed this bug on master:
```python
In [43]: dsolve(x*f(x).diff(x, 2) - x*f(x))
Out[43]: f(x) = √x⋅(C₁⋅besselj(1/2, ⅈ⋅x) + C₂⋅bessely(1/2, ⅈ⋅x))
```
Is that bug also in this PR?

Mohitbalwani26:
> ```python
> dsolve(x*f(x).diff(x, 2) - x*f(x))
> ```

Isn't it the correct solution as `checkodesol(eq,sol)` also returns `[(True, 0)]`?
According to the reference mentioned in the docstring [Link](https://www.math24.net/bessel-differential-equation) it is reducible to bessel equation (4th point)
x<sup>2</sup>y'' + axy' + (x<sup>2</sup> - n<sup>2</sup>) can be reduced to bessel form by substituting y = x<sup>(1-a)/2</sup> z

oscarbenjamin:
> Isn't it the correct solution as `checkodesol(eq,sol)` also returns `[(True, 0)]`?

It might be correct for some part of the domain but it is definitely not as good as this:
```python
In [68]: eq = f(x).diff(x, 2) - f(x)

In [69]: dsolve(eq)
Out[69]: 
           -x       x
f(x) = C₁⋅ℯ   + C₂⋅ℯ 

In [70]: dsolve(expand(x*eq))
Out[70]: f(x) = √x⋅(C₁⋅besselj(1/2, ⅈ⋅x) + C₂⋅bessely(1/2, ⅈ⋅x))
```
The Bessel solution is not continuous at `x=0` and is significantly more complicated then the basic exponential solution here.

Mohitbalwani26:
> The Bessel solution is not continuous at `x=0` and is significantly more complicated then the basic exponential solution here.

So should we first check if the equation is analytic at 0 or some point? 🤔

oscarbenjamin:
> So should we first check if the equation is analytic at 0 or some point? 🤔

This should definitely match the constant coefficients solver. I'm not sure what should be done with the Bessel solver.

Mohitbalwani26:
> This should definitely match the constant coefficients solver. I'm not sure what should be done with the Bessel solver.

It doesn't not match directly with constant coefficients solver because in `eq_higher_order_free`, we are removing the leading coefficient having f(x).  Although this equation has `x` as common so it should match with Factorable solver and then recursively call the constant coefficient solver.

Mohitbalwani26:
> It doesn't not match directly with constant coefficients solver because in `eq_higher_order_free`, we are removing the leading coefficient having f(x). Although this equation has `x` as common so it should match with Factorable solver and then recursively call the constant coefficient solver.

@oscarbenjamin Thanks for pointing out this example! 
This was a bug and got introduced while refactoring the `Factorable` hint.  As earlier in sympy version 1.5 it was classified by factorable hint and solution given was correct. 
I have fixed it now and added the example.

sympy/solvers/ode/tests/test_single.py
```diff
@@ -521,6 +521,10 @@ def test_factorable():
     _ode_solver_test(_get_examples_ode_sol_factorable)
 
 
+def test_slow_examples_factorable():
+    _ode_solver_test(_get_examples_ode_sol_factorable, run_slow_test=True)
```
oscarbenjamin, line 525: Shouldn't this be marked as slow?

Mohitbalwani26: yes, Done.

sympy/solvers/ode/single.py
```diff
@@ -2615,5 +2617,182 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
...
+    r"""
+    Gives solution of the Bessel differential equation
+
+    .. math :: x^2 \frac{d^2y}{dx^2} + x \frac{dy}{dx} y(x) + (x^2-n^2) y(x)
+
+    if n is integer then the solution is of the form Eq(f(x), C0 besselj(n,x)
+    + C1 bessely(n,x)) as both the solutions are linearly independent else if
+    n is a fraction then the solution is of the form Eq(f(x), C0 besselj(n,x)
+    + C1 besselj(-n,x)) which can also transform into Eq(f(x), C0 besselj(n,x)
+    + C1 bessely(n,x)).
```
oscarbenjamin, line 2630: The equations here should be in LaTeX. If they are supposed to be code rather than equations then they should be in double backticks.

sympy/solvers/ode/ode.py
```diff
@@ -1059,6 +1058,8 @@ class in it.  Note that a hint may do this anyway if
         NthLinearEulerEqHomogeneous: ('nth_linear_euler_eq_homogeneous',),
         NthLinearEulerEqNonhomogeneousVariationOfParameters: ('nth_linear_euler_eq_nonhomogeneous_variation_of_parameters',),
         NthLinearEulerEqNonhomogeneousUndeterminedCoefficients: ('nth_linear_euler_eq_nonhomogeneous_undetermined_coefficients',),
+        LinearBessel2nd: ('2nd_linear_bessel',),
+        LinearAiry2nd: ('2nd_linear_airy',),
```
oscarbenjamin, line 1062: It would be better to call these e.g. `SecondLinearBessel`

oscarbenjamin: Also Hypergeometric2nd should be renamed.

oscarbenjamin:
Looks good. Thanks

Mohitbalwani26:
@oscarbenjamin should I raise PR for lie_group first and then speedup the process of calling classify_ode?

oscarbenjamin:
> @oscarbenjamin should I raise PR for lie_group first and then speedup the process of calling classify_ode?

Up to you. You can also just work on both in parallel.

PR #21712, opened 2021-07-06
Implement Forces and Torques

sidhu1012:
Automate applying forces and torques

#### Other comments

Ping @moorepants @Sc0rpi0n101 
#### Release Notes

* physics.mechanics
    * Update `apply_force` function of class `Body` to apply force on one body, or equal and opposite forces on two bodies.
    * Update `apply_torque` function of class `Body` to apply torque on one body, or equal and opposite torques on two bodies.
    * `Body.loads` is now an immutable attribute.
    * Add `clear_loads` function to class `Body` to clear Body's load list.
    * Add `remove_load` function to `Body` to remove a load from Body's load list.


sidhu1012:
@moorepants I have implemented `Force` and `Torque` function as we discussed earlier today, its still wip but reviews would be appreciated.

sympy/physics/mechanics/functions.py
```diff
@@ -723,3 +723,118 @@ def flist_iter():
         unzip = lambda l: list(zip(*l)) if l[0] else [(), ()]
         vel_list, f_list = unzip(list(flist_iter()))
     return vel_list, f_list
+
+def Force(force, body1, body2=None, point1=None, point2=None):
```
moorepants, line 727: Functions names should follow PEP8 recommendations for capitalization and ideally be a verb not a noun.

sidhu1012: Corrected!

sidhu1012:
@moorepants If this implementation seems correct , then I can start adding the tests.

moorepants:
> @moorepants If this implementation seems correct , then I can start adding the tests.

You've done it backwards. Write the tests and example first. The tests and examples will tell you what the implementation needs to be. The implementation is only correct if it correctly computes the tests and solves example problems correctly. But you have neither, so there is no way to know!

sidhu1012:
> You've done it backwards. Write the tests and examples first. The tests and examples will tell you what the implementation needs to be. The implementation is only correct if it correctly computes the tests and solves example problems correctly. But you have neither, so there is no way to know!

Any specific problem that I must include in tests?

moorepants:
I gave you two today in the meeting and have given you 10s of others in prior discussions.

sidhu1012:
@moorepants 

Example for this implementation:- 

```
# For 2 bodies attach with spring

>>> B1 = Body('B1')
>>> B2 = Body('B2')
>>> k, g = symbols('k g')
>>. x, v = dynamicsymbols('x v')

>>> spring_force = k*x

#Applying equal and opposite spring force to bodies masscenter
>>> apply_force(spring_force, B1, B2)

#Checking body load
>>> B1.loads
[(B1_masscenter, k*x(t))]

>>> B2.loads
[(B2_masscenter, -k*x(t))]

#Now if we apply gravity to masscenters it will be added to spring force
>>> g1 = B1.mass *g
>>> g2 = B2.mass*g
>>> apply_force(g1, B1)
>>> B2.apply_force(g2) #Also a way for single force

#Checking loads now
>>> B1.loads
[(B1.masscenter, B1_mass*g + k*x(t))]

>>> B2.loads
[(B2_masscenter, B2_mass*g - k*x(t))]

#Applying force on a point other than masscenter[hand of god 😄 ]

>>> P = Point('P')
>>> force = x
>>> B1.apply_force(force, P) # or apply_force(force, B1, point1=P)

>>> B1.loads
[(B1.masscenter, B1_mass*g + k*x(t)), (P, force)]

# Applying equal and opposite torque now
>>> torque = v
>>> apply_torque(torque, B1, B2)

>>> B1.loads
[(B1.masscenter, B1_mass*g + k*x(t)), (P, force), (B1_frame, torque)]

>>> B2.loads
[(B2_masscenter, B2_mass*g - k*x(t)), (B2_frame, -torque)]

# Since .loads attribute is immutable we can't change it directly
# We can use Body's .remove_load() to remove torque from body's frame or net force on any point on Body

>>> B1.remove_load(P) 
>>> B1.loads
[(B1.masscenter, B1_mass*g + k*x(t)), (B1_frame, torque)]

>>> B1.remove_load(B1.frame)
>>> B1.loads
[(B1.masscenter, B1_mass*g + k*x(t))]

# We can also clear a body's loads list by using body's clear_load()

>>> B2.clear_load()
>>> B2.loads
[]

```

moorepants:
Please commit the examples in the documentation so that we can comment on them using github's commenting feature.

sidhu1012:
> Please commit the examples in the documentation so that we can comment on them using github's commenting feature.

I'll add them to tests.

sidhu1012:
@moorepants I have added one example to docstring and other to tests. Please review now.

sympy/physics/mechanics/body.py
```diff
@@ -136,6 +136,10 @@ def __init__(self, name, masscenter=None, mass=None, frame=None,
         else:
             RigidBody.__init__(self, name, masscenter, frame, _mass, _inertia)
 
+    @property
+    def loads(self):
+        return self._loads
```
moorepants, line 141: I think that having `loads` as a mutable list is a good thing in this case. Users may want to remove loads from the list and manipulating the list is the easiest way to do so.

moorepants: In other words, a user should be able to add and remove loads manually and not need the new functions you've made. Those functions are really just convenience functions.

sympy/physics/mechanics/body.py
```diff
@@ -191,6 +195,13 @@ def apply_force(self, vec, point=None):
             ...                                   body.frame.y)
             >>> body.apply_force(body.mass * g * body.frame.x, point)
 
+        Notes
+        =====
+
+        If a new force is applied to a body on a point which already has some
+        force applied on it, then the new force is added to the already applied
+        force on that point.
```
moorepants, line 203: This isn't how we handle forces/torques in KanesMethod or LagrangesMethod. I think we shouldn't combine forces automatically. The user can do that if needed themselves.

sympy/physics/mechanics/body.py
```diff
@@ -224,11 +241,83 @@ def apply_torque(self, vec):
...
+        >>> B.apply_force(force)
+        >>> B.loads
+        [(B_masscenter, B_frame.x + B_frame.y)]
+        >>> B.clear_load()
+        >>> B.loads
+        []
+
+        """
+
+        self._loads = []
```
moorepants, line 281: If `.load` is mutable, this method is not necessary. A rule of thumb to remember, if the implementation of the method is only 1 line, you likely don't need the method. (that often goes for 2 or even 3 lines too)

sympy/physics/mechanics/body.py
```diff
@@ -224,11 +241,83 @@ def apply_torque(self, vec):
...
+
+    def remove_load(self, about):
```
moorepants, line 283: This seems potentially useful. If you have many loads acting on the same point you can just say "remove all forces from this point". Maybe this shoudl be `remove_loads_from(point|frame)`.

sympy/physics/mechanics/tests/test_functions.py
```diff
@@ -250,3 +252,76 @@ def test_center_of_mass():
...
+    assert B2.loads == [(P2, -force), (B2.masscenter, g2)]
+
+    force2 = x*N.x
+
+    apply_force(force2, B1, B2) #Applying time varying force on masscenter
+
+    assert B1.loads == [(P1, force), (B1.masscenter, force2+g1)]
+    assert B2.loads == [(P2, -force), (B2.masscenter, -force2+g2)]
+
+def test_apply_loads_on_multi_degree_freedom_holonomic_system():
```
moorepants, line 305:
```suggestion
def test_apply_loads_on_multi_degree_freedom_holonomic_system():
    """Example based on: https://pydy.readthedocs.io/en/latest/examples/multidof-holonomic.html"""
```

sympy/physics/mechanics/tests/test_functions.py
```diff
@@ -250,3 +252,76 @@ def test_center_of_mass():
...
+    W = Body('W') #Wall
+    B = Body('B') #Block
+    P = Body('P') #Pendulum
+    b = Body('b') #bob
+    q1, q2 = dynamicsymbols('q1 q2') #generalized coordinates
+    k, c, g, kT = symbols('k c g kT') #constants
+    F, T = dynamicsymbols('F T') #Specified forces
+
+    #Applying forces
+    B.apply_force(F*W.x)
```
moorepants, line 315: What is the purpose of having and `apply_force()` method on `Body` and a function called `apply_force`? Why can't, for example, the method do everything the function can?

sympy/physics/mechanics/tests/test_functions.py
```diff
@@ -250,3 +252,76 @@ def test_center_of_mass():
...
+    P.apply_force(P.mass*g*W.y)
+    b.apply_force(b.mass*g*W.y)
+
+    #Applying torques
+    apply_torque(kT*q3*W.z, P, b)
+    P.apply_torque(T*W.z)
+
+    assert B.loads == [(B.masscenter, (F - k*q1 - c*q1.diff())*W.x)]
+    assert P.loads == [(P.masscenter, P.mass*g*W.y), (P.frame, (T + kT*q3)*W.z)]
+    assert b.loads == [(b.masscenter, b.mass*g*W.y), (b.frame, -kT*q3*W.z)]
```
moorepants, line 327: Missing loads on the wall.

sympy/physics/mechanics/tests/test_functions.py
```diff
@@ -250,3 +252,76 @@ def test_center_of_mass():
...
+    #Applying forces
+    B.apply_force(F*W.x)
+    apply_force(k*q1*W.x, W, B) #Spring force
+    apply_force(c*q1.diff()*W.x, W, B) #dampner
+    P.apply_force(P.mass*g*W.y)
+    b.apply_force(b.mass*g*W.y)
+
+    #Applying torques
+    apply_torque(kT*q3*W.z, P, b)
+    P.apply_torque(T*W.z)
```
moorepants, line 323: This should be equal and opposite between B and P.

sympy/physics/mechanics/body.py
sidhu1012: I think we should keep it immutable else user may be able to append that list. But if it is immutable only way to append is by using functions, which would ensure that force is vector, point is a Point and all.

sympy/physics/mechanics/body.py
sidhu1012: I think we should add forces/torques because, like in `test_apply_loads_on_multi_degree_freedom_holonomic_system()` example, the force on 1 Body, is sum of dampner, spring and gravity on all same point. If we don't add them then `loads` would store tuples of same point with different forces.

sympy/physics/mechanics/body.py
sidhu1012: I think it should be immutable.

sympy/physics/mechanics/body.py
sidhu1012: This function would only be useful if `.loads` is immutable.

sympy/physics/mechanics/tests/test_functions.py
sidhu1012: I was thinking `apply_force()` could be used to apply equal and opposite forces with added functionality of single forces, whereas `Body.apply_force` could be used to add single forces. Combining them might make it a bit complicated as it was in orient().

sympy/physics/mechanics/tests/test_functions.py
sidhu1012: I think `kT*q3*W.z` is equal and opposite. `T*W.z` is only on P.

moorepants: It is between B and P. With no assumption of a newtonian frame, all forces/torques should be account for. Also this is a unit test, it is helpful to show that the general cases work, not just special cases.

sympy/physics/mechanics/tests/test_functions.py
moorepants: The function `apply_force()` already has a certain level of complexity and you have an identical method on Body that doesn't do the same thing. I'm not seeing why we need the function if we have the method? Here are scenarios:

1. User has to manually handle equal and opposite forces and `.apply_force()` only accepts a single force.

```python
A = Body('A')
B = Body('B')
A.apply_force(F*A.x)
B.apply_force(-F*A.x)
```

2. Equal and opposite can be handled automatically by `.apply_force()` by `.apply_force()` accepting a body.

```python
A = Body('A')
B = Body('B')
A.apply_force(F*A.x, reaction_body=B)
```

3. Equal and opposite can be handled automatically by `apply_force()` (function).

```python
A = Body('A')
B = Body('B')
apply_force(F*A.x, reaction_body1=A, reaction_body2=B)
```

2. seems the most straight forward with no redundancy to me and less typing, at least from this minimal example. 3. has redundant methods and functions. It isn't clear why having two ways to add the equal & opposite force gives benefit.

If there are example problems that show why the function is better, show a demonstration.

sympy/physics/mechanics/body.py
moorepants: Why is it bad for someone to append to that list? It isn't like something has to be calculated in Body when that list is modified. It is just a simple storage container.

moorepants: In joint, you'd have to recompute the joint's position and velocities if coordinate or speed was changed, thus we make it immutable to force the user to choose the coordinate and speed only on initialization. But the user may not know what all the loads are on initialization, then add them slowly in the problem construction. So `.load` has to be mutable. There is no reason to make the mutable `.load` a property unless you do something in it's setter method. That is, mutable attributes only need to be properties if there is a setter method. Body does nothing with `.load` so it should be a simply mutable attribute.

sympy/physics/mechanics/body.py
moorepants: If you sum the forces, then it makes it harder to separate the forces if you need to work with the force list later. There is no reason to sum the forces here. It adds no advantage, only a disadvantage.

> If we don't add them then loads would store tuples of same point with different forces.

Yes, that's what we've done for 10+ years and exactly what KanesMethod and LagrangesMethod expect.

sympy/physics/mechanics/body.py
moorepants: I explained above.

sympy/physics/mechanics/body.py
moorepants: That's not true. This function can only work if `.load` is assume mutable. You can remove something from something that is immutable.

sympy/physics/mechanics/tests/test_functions.py
moorepants: And, to be honest, point 1. is pretty clean and simple. It just requires the user to do the correct thing (garbage in, gargage out philosophy).

sidhu1012: Yes, I told you about way 2 in meeting, but you asked me to proceed with 3.

I will change it to 2 , I also prefer 2.

But there is one other way also, i.e , we can use Body's attribute for single forces and `apply_force` function for equal and opposite forces.

So way 2 or this way that I explained?

sympy/physics/mechanics/body.py
sidhu1012: If we define `self._loads = [ ]` , and make `loads` a property as I have done in this PR, user can't alter loads of bodies directly but we can.

sympy/physics/mechanics/body.py
sidhu1012: But if we have `JointsMethod`, and then load changes on Body then it may affect result because loads are Body's attributes.

sympy/physics/mechanics/body.py
sidhu1012: In holonomic problem I did, multiple forces are applied and summed together before passing, that's why I added and by this way we can have all forces applied on one point together and by remove_load we can remove the entire load from one point.

sympy/physics/mechanics/tests/test_functions.py
moorepants: I recall you suggesting a function not a method in the meeting, but maybe I misunderstood. I seemed reasonable then, but we hadn't seen the examples of its use. Regardless, I don't know what the best option is until we've explored the possibilities. This is the purpose of trying to come up with examples for use and writing them down. We can then examine the proposed API design in the context of the examples.

The process is to:

1. Think of all possible examples (including edge cases) that we want the API to be used for. These examples should be derived from existing problems we've created over the years as well as looking at other similar tools and how they approach the problems.
2. Propose an API design.
3. Show the API design used in the examples in a PR.
4. Discuss and debate whether the API design is good or needs improvement.
5. Modify the API to make it better.
6. Repeat starting at 1. if you think of new examples that you didn't originally think of.

Steps 1 to 6 are repeated until everyone involved in the design is happy with the result.

You job is to do this process and I will help with point 1. by pointing you to examples and providing my past experience and in 4. by discussing your proposal.

For this particular API idea, I'm not yet sure point 1. is fully fleshed out and if it is, then the API is too complex, thus my suggestions.

sidhu1012: We will discuss it further in upcoming meet.

sidhu1012:
@moorepants It is ready to be merged.

sidhu1012:
@moorepants I you could review this PR, then I could spend my weekend trying to figure out proper implementation of JointsMethod.

sympy/physics/mechanics/body.py
```diff
@@ -151,84 +155,258 @@ def z(self):
         """The basis Vector for the Body, in the z direction. """
         return self.frame.z
 
-    def apply_force(self, vec, point=None):
-        """
-        Adds a force to a point (center of mass by default) on the body.
+    def apply_force(self, force, point1=None, other=None, point2=None):
```
moorepants, line 158:
```suggestion
    def apply_force(self, force, point=None, reaction_body=None, reaction_point=None):
```

sympy/physics/mechanics/body.py
```diff
@@ -151,84 +155,258 @@ def z(self):
...
+        Now let's apply equal and opposite spring force to the masscenter of the bodies.
+
+        >>> B1.apply_force(spring_force, other=B2)
+
+        We can check the loads(forces) applied to bodies now.
+
+        >>> B1.loads
+        [(B1_masscenter, x(t)*N_frame.x)]
+        >>> B2.loads
+        [(B2_masscenter, - x(t)*N_frame.x)]
```
moorepants, line 229: Add spring force acting on two different points other than mass center.

sympy/physics/mechanics/body.py
```diff
@@ -151,84 +155,258 @@ def z(self):
...
+        >>> from sympy.physics.mechanics import Body, dynamicsymbols
+        >>> N = Body('N') #Newtonion frame
+        >>> B1 = Body('B1')
+        >>> B2 = Body('B2')
+        >>> v = dynamicsymbols('v')
+        >>> T = v*N.y #Torque
+
+        Now let's apply equal and opposite torque to the bodies.
+
+        >>> B1.apply_torque(T, B2)
```
moorepants, line 317:
```suggestion
        >>> B1.apply_torque(T, reaction_body=B2)
```

sympy/physics/mechanics/body.py
```diff
@@ -151,84 +155,258 @@ def z(self):
...
+
+        We can check the loads(torquess) applied to bodies now.
```
moorepants, line 319:
```suggestion
        We can check the loads (torques) applied to bodies now.
```

sympy/physics/mechanics/body.py
```diff
@@ -151,84 +155,258 @@ def z(self):
...
+        self._loads = []
+
+    def remove_load(self, about=None):
+        """
+        Remove load about a point or frame.
+
+        Parameters
+        ==========
+
+        about : Point, optional
```
moorepants, line 375:
```suggestion
        about : Point or ReferenceFrame, optional
```

moorepants:
LGTM!

Issue #19069, opened 2020-04-04
[Discussion]: Influence Line Diagrams for Beams


Psycho-Pirate:
I would like to implement influence line diagrams for reactions, shear and moment as described in the phase 1 of my [GSoC proposal here.](https://docs.google.com/document/d/1sYoj5F6A9tZEROYa7el6D6gqgqwRsX2qf6zX0Ejlu7Q/edit#heading=h.p6wewc7lx5n6)
It will basically involve making three functions that will return the equations along with plots of reactions, shear and moments.
The details regarding implementation can be found in my GSoC Proposal.

Ping @moorepants @jashan498 @ishanaj

moorepants:
I've never heard of these diagrams. I looked at your proposal but didn't really understand their utility from my brief review of what you wrote. How are these useful? What do they tell us that isn't already provided by the existing plots?

Psycho-Pirate:
We currently have functions for shear force diagrams(SFD) and bending moment diagrams(BMD).
SFD and BMD represent the effect of **_fixed loads_** at _**all points**_ along the axis of the member. Whereas Influence line diagrams represent the effect of a **_moving load_** only at a **_specified point_** on a member. 
They are really useful in describing the shear and moment variation when the structure is subjected to a _live_ or _moving_ load.
They are really useful in designing beams used in bridges, crane rails, conveyor belt, floor girders, and other structures where loads will move along their span.
They can be used to determine the maximum shear stress or bending moment caused by a moving load and the structure can be designed accordingly. The SFD and BMD functions only work for dead or fixed loads. @moorepants

Psycho-Pirate:
The influence line diagram(ILD) for reaction forces basically depicts the variation of reaction force due to the effect of a moving load on the beam.
The ILD for shear depicts the variation of shear force at a particular point due to the effect of a moving load on the beam.
Similarly,
The ILD for moment depicts the variation of bending moment at a particular point due to the effect of a moving load on the beam.

Influence lines are both scalar and additive.

This means that they can be used even when the load that will be applied is not a unit load or if there are multiple loads applied. To find the effect of any non-unit load on a structure, the ordinate results obtained by the influence line are multiplied by the magnitude of the actual load to be applied. The entire influence line can be scaled, or just the maximum and minimum effects experienced along the line. The scaled maximum and minimum are the critical magnitudes that must be considered while designing a beam.

Psycho-Pirate:
I want to implement these as they have huge real-life applications.
[...]

PR #21714, opened 2021-07-06
[GSoC] Function for I.L.D. for reaction loads added

Psycho-Pirate:

See #19069

Function `solve_for_ild_reactions()` is added which returns the reaction equations under the effect of a moving load.
Function `plot_ild_reactions()` is added which plots these equations.

#### Release Notes

* physics.continuum_mechanics 
   * Added ability to calculate and plot I.L.D. Reaction Equations


Psycho-Pirate:
Here's an example:
ildrxm
```
>>> from sympy.physics.continuum_mechanics.beam import Beam
>>> from sympy import symbols
>>> E, I = symbols('E, I')
>>> R1, R2 = symbols('R1, R2')
>>> b = Beam(10, E, I)
>>> b.apply_load(R1, 0, -1)
>>> b.apply_load(R2, 10, -1)
>>> print(b.solve_for_ild_reactions(1,R1,R2))
{R1: 1 - X/10, R2: X/10}
>>> b.plot_ild_reactions()
```
ildr1
ildr2

Psycho-Pirate:
@ishanaj @Sc0rpi0n101

sympy/physics/continuum_mechanics/beam.py
```diff
@@ -1561,6 +1562,62 @@ def plot_loading_results(self, subs=None):
...
+            Magnitude of moving load
+        reactions :
+            The reaction forces applied on the beam.
+
+        """
+        x = self.variable
+        l = self.length
+        C3 = Symbol('C3')
+        C4 = Symbol('C4')
+        X = symbols('X')
```
ishanaj, line 1584: What is the purpose of this `X` when we have `self.variable`?

Psycho-Pirate: In order to solve the slope and moment equations, a limit (lim x->L) is applied. Here 'x' represents the length of the beam. The uppercase 'X' represents the distance of the moving load from origin. I have also explained this in my gsoc proposal.
Screenshot (6)

ishanaj: Firstly the big `X` is something the user never defined and thus would never know what it is. So you might have to provide user information about this `X` or create a similar flow of how the small `x` is taken as a parameter when a Beam object is defined.

Secondly, according to what I understand, `x` and `X` both vary over the beam length, and represent any random cross-section over the length, so we can use the same variable for them, unless there is some issue in solving the equations.

Psycho-Pirate: > Secondly, according to what I understand, `x` and `X` both vary over the beam length, and represent any random cross-section over the length, so we can use the same variable for them, unless there is some issue in solving the equations.

The only reason I added 'X' was to avoid confusion as X is not any distance on the beam, it is the distance at which the moving load is acting. But if it causes more confusion then it can be definitely removed.

Psycho-Pirate:
In the recent commit I made the following changes:
* Added two different examples in both functions along with an image.
* The `plot_ild_reactions()` now raises an error if `solve_for_ild_reactions()` is not called
* All the reaction graphs are plotted on a single plot
* removed 'X' variable
* fixed sign convention
@Sc0rpi0n101 @ishanaj

sympy/physics/continuum_mechanics/beam.py
```diff
@@ -1561,6 +1562,129 @@ def plot_loading_results(self, subs=None):
...
+            [0]: cartesian line: x/7 - 22/7 for x over (0.0, 10.0)
+            Plot[1]:Plot object containing:
+            [0]: cartesian line: -x/7 - 20/7 for x over (0.0, 10.0)
+
+        """
+        if not self._ild_reactions:
+            raise KeyError("solve_for_ild_reactions() not called")
+
+        x = self.variable
+        p = []
```
ishanaj, line 1682: Variable name should be something more relevant for code readability

sympy/physics/continuum_mechanics/beam.py
```diff
@@ -1561,6 +1562,129 @@ def plot_loading_results(self, subs=None):
...
+        for i in self._ild_reactions:
```
ishanaj, line 1683: Here too, it can be `for reaction, equation in self._ild_reactions` or something better than just `i`

sympy/physics/continuum_mechanics/beam.py
```diff
@@ -131,6 +131,7 @@ def __init__(self, length, elastic_modulus, second_moment, area=Symbol('A'), var
         self._support_as_loads = []
         self._applied_loads = []
         self._reaction_loads = {}
+        self._ild_reactions = {}
```
ishanaj, line 134: A property can also be defined for the same for the user to access

ishanaj:
I am getting this plot for the example present in `plot_ild_reactions()`. The y-axis has -ve values even above the origin. Why is it so?

image

ishanaj:
Two cases that need to be considered: 
Case 1: If the user has already called `solve_for_reaction_loads()`, the current code seems to fail.

Case 2: If length is a symbol, or if there are any other symbols present, we might need to keep a `subs={}` parameter in the plotting part. I haven't tried Case 2, but I believe this condition will occur. You can test an example and share the results if it works.

sympy/physics/continuum_mechanics/beam.py
```diff
@@ -1561,6 +1562,129 @@ def plot_loading_results(self, subs=None):
...
+            >>> b.plot_ild_reactions()
+            PlotGrid object containing:
+            Plot[0]:Plot object containing:
+            [0]: cartesian line: x/7 - 22/7 for x over (0.0, 10.0)
+            Plot[1]:Plot object containing:
+            [0]: cartesian line: -x/7 - 20/7 for x over (0.0, 10.0)
+
+        """
+        if not self._ild_reactions:
+            raise KeyError("solve_for_ild_reactions() not called")
```
ishanaj, line 1679: Why a keyerror? I think it should be a ValueError instead. Also, the message should be elaborated like: "I.L.D. reaction equations not found. Please use `solve_for_ild_reactions()` to generate the I.L.D. reaction equations"

Psycho-Pirate:
I have implemented most of the changes you suggested.


> I am getting this plot for the example present in `plot_ild_reactions()`. The y-axis has -ve values even above the origin. Why is it so?
> 
I think this depends on the plot function as I have tried plotting the same equation just by using the plot function and it still shows negative values above origin.


> Two cases that need to be considered:
> Case 1: If the user has already called `solve_for_reaction_loads()`, the current code seems to fail.
> 
> Case 2: If length is a symbol, or if there are any other symbols present, we might need to keep a `subs={}` parameter in the plotting part. I haven't tried Case 2, but I believe this condition will occur. You can test an example and share the results if it works.

There are some things that need to be discussed here, after that I will push all the changes together.

Psycho-Pirate:
In the recent commit I have made the following changes:
* Added a property which returns ild_reactions
* Added an attribute `_original_load` which retains an unsubstituted copy of load function
* Added a helper function `_solve_for_ild_reactions()` which calculates and returns shear force and bending moment using original load
* Added subs functionality in the `plot_ild_reactions()` function
* Changed variable names to something relevant to the beam module

The `solve_for_ild_reactions` function is now independent of `solve_for_reaction_loads`

Here is a working example:
```
>>> from sympy import symbols
>>> from sympy.physics.continuum_mechanics.beam import Beam
>>> E, I = symbols('E, I')
>>> R_0, R_7, l = symbols('R_0, R_7, l')
>>> subs = {l:10, E:100, I:25}
>>> b = Beam(l, E, I)
>>> b.apply_support(0, 'roller')
>>> b.apply_support(7, 'roller')
>>> b.apply_load(5, 4, -1)

>>> b.solve_for_reaction_loads(R_0,R_7)

>>> b.solve_for_ild_reactions(1,R_0,R_7)
>>> b.plot_ild_reactions(subs)
```
ildild

Psycho-Pirate:
> I am getting this plot for the example present in `plot_ild_reactions()`. The y-axis has -ve values even above the origin. Why is it so?

I have opened an issue regarding this [here](https://github.com/sympy/sympy/issues/21777)

sympy/physics/continuum_mechanics/beam.py
```diff
@@ -1561,6 +1577,159 @@ def plot_loading_results(self, subs=None):
...
+            eqs = deflection_curve.subs(x, position) - value
+            deflection_eqs.append(eqs)
+
+        solution = list((linsolve([shear_curve, moment_curve] + slope_eqs
+                            + deflection_eqs, (C3, C4) + reactions).args)[0])
+        solution = solution[2:]
+
+        # Determining the equations and solving them.
+        self._ild_reactions = dict(zip(reactions, solution))
+        return self._ild_reactions
```
ishanaj, line 1663: `solve_for_ild_reactions` probably shouldn't return anything, it should just update the value of `self._ild_reactions`.
It goes with the behavior of the other solve functions in the module.

ishanaj:
I have left a review. Apart from that you should also include some release notes with this PR.
Rest looks good to me!

Psycho-Pirate:
> I have left a review. Apart from that you should also include some release notes with this PR.
> Rest looks good to me!

Done!

sympy/physics/continuum_mechanics/beam.py
```diff
@@ -1561,6 +1577,161 @@ def plot_loading_results(self, subs=None):
...
+            eqs = deflection_curve.subs(x, position) - value
+            deflection_eqs.append(eqs)
+
+        solution = list((linsolve([shear_curve, moment_curve] + slope_eqs
+                            + deflection_eqs, (C3, C4) + reactions).args)[0])
+        solution = solution[2:]
+
+        # Determining the equations and solving them.
+        self._ild_reactions = dict(zip(reactions, solution))
+        self._ild_reactions
```
ishanaj, line 1664: This line is not required anymore then.

Psycho-Pirate:
I have removed the line

ishanaj (approved):

ishanaj:
LGTM! Merging..!

PR #21720, opened 2021-07-08
Added pymc3 to optional-dependencies

mostlyaman:
#21590 

stats module uses pymc3 in sampling random variables. Tests for pymc3 were being skipped because the library was not installed during the tests.

#### Release Notes

NO ENTRY


mostlyaman:
@oscarbenjamin I tested the current master locally and it fails. Indeed, it was pymc3 not installed during the tests.

oscarbenjamin:
Okay great. The optional dependencies test has failed in this PR. Can you add this in #21590 instead then so that we can see that the tests are passing with the changes there?

PR #21730, opened 2021-07-10
Check if velocity is defined in frame before taking derivative of pos

sidhu1012:
Earlier when velocity was not defined in a frame but pos vector was given, the velocity was calculated(#20049) . But now if we have a point defined with velocity in that frame, then we first check if current point's velocity is available in given frame or not.

#### Release Notes

* physics.vector
    * `Point.vel()` checks velocity in intermediate frames before taking derivative of position vector.

### DIFF
--- a/sympy/physics/vector/tests/test_point.py
+++ b/sympy/physics/vector/tests/test_point.py
@@ -298,3 +298,14 @@ def test_auto_vel_multiple_path_warning_msg():
         assert issubclass(w[-1].category, UserWarning)
         assert 'Velocity automatically calculated based on point' in str(w[-1].message)
         assert 'Velocities from these points are not necessarily the same. This may cause errors in your calculations.' in str(w[-1].message)
+
+def test_vel_frame():
+    N = ReferenceFrame('N')
+    A = ReferenceFrame('A')
+    A.orient_axis(N, N.x, 0)
+    P = Point('P')
+    P.set_vel(N, 5*N.x)
+    P1 = Point('P1')
+    P1.set_pos(P, A.x+N.x)
+    P1.set_vel(A, 10*A.x)
+    assert P1.vel(N) == 5*N.x + 10*A.x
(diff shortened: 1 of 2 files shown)


moorepants:
So, the behavior that was introduced in https://github.com/sympy/sympy/pull/20049 overwrites all velocities in the branch of the tree even if they've already been defined?

sidhu1012:
Earlier behaviour was to find the closest point in tree having velocity defined in the required frame and then add that velocity to the derivative of `pos_vector` between these two points.

Now the behaviour would be, before computing this, we try to find a frame in which velocity is defined for current point .

If velocity is defined then we don't compute derivate and use that velocity.

moorepants:
> Earlier behaviour was to find the closest point in tree having velocity defined in the required frame and then add that velocity to the derivative of pos_vector between these two points.

If that was the behavior, then why would it overwrite a velocity that is already set? 

Velocities should never be overwritten, in any situation, unless explicitly done by the user calling `.set_vel()` or `v1pt_theory() or `v2p_theory`. If the algorithm introduced in #20049 overwrites velocities, then that should be corrected, as it will cause all kinds of trouble for users.

sidhu1012:
No it never overwrites

sympy/physics/vector/tests/test_point.py
```diff
@@ -298,3 +298,14 @@ def test_auto_vel_multiple_path_warning_msg():
...
+def test_vel_frame():
+    N = ReferenceFrame('N')
+    A = ReferenceFrame('A')
+    A.orient_axis(N, N.x, 0)
+    P = Point('P')
+    P.set_vel(N, 5*N.x)
+    P1 = Point('P1')
+    P1.set_pos(P, A.x+N.x)
+    P1.set_vel(A, 10*A.x)
+    assert P1.vel(N) == 5*N.x + 10*A.x
```
moorepants, line 311: As @TJStienstra points out, this test is incorrect. `A_v_P1` is set to `10*A.x` but A = N, so `N_v_P1` also should be `10*A.x`. For some reason the code adds the velocity from `N_v_P`, which makes no sense.

Issue #21721, opened 2021-07-08
Bug in integration solver


pctablet505:
For calculating the integration of pdf of cauchy distribution I tried sympy, but it produced wrong result.

`integrate(1/(pi*(1+(x-a)**2)),(x,-oo,oo))` 
the **expected** output was **1** but it **produced output 0 which is wrong**.

![image](https://user-images.githubusercontent.com/55033230/124940822-ef6a4e80-e027-11eb-8772-214cbb31fd8a.png)

![image](https://user-images.githubusercontent.com/55033230/124941424-67387900-e028-11eb-833e-e1b44d074e06.png)
![image](https://user-images.githubusercontent.com/55033230/124941649-95b65400-e028-11eb-99cf-26861bdab0f3.png)

oscarbenjamin:
If `a` is declared as real then the correct answer is given:
```python
In [8]: a = Symbol('a', real=True)

In [9]: integrate(1/(pi*(1+(x-a)**2)),(x,-oo,oo))
Out[9]: 1
```

pctablet505:
Thank you so much :)

oscarbenjamin:
Reopening because the answer should not be given as zero when `a` has not been declared as real

oscarbenjamin:
Seems to be a bug in the limit code:
```python
In [1]: a = Symbol('a')

In [2]: I = integrate(1/(pi*(1+(x-a)**2)),x)

In [3]: I
Out[3]: 
  ⅈ⋅log(-a + x - ⅈ)   ⅈ⋅log(-a + x + ⅈ)
- ───────────────── + ─────────────────
          2                   2        
───────────────────────────────────────
                   π                   

In [4]: I.limit(x, oo)
Out[4]: 0

In [5]: I.limit(x, -oo)   # <-- should be -1
Out[5]: 0
```
Seems to be also a bug for this limit (even though the integral works in this case):
```python
In [9]: a = Symbol('a', real=True)

In [10]: I = integrate(1/(pi*(1+(x-a)**2)),x)

In [11]: I
Out[11]: 
-atan(a - x) 
─────────────
      π      

In [12]: I.limit(x, oo)  # <-- should be +1/2
Out[12]: -1/2

In [13]: I.limit(x, -oo)
Out[13]: -1/2
```

oscarbenjamin:
> ```python
> In [12]: I.limit(x, oo)  # <-- should be +1/2
> Out[12]: -1/2
> ```

[...]

PR #21589, opened 2021-06-08
[GSoC] A few leading term methods changed

0sidharth:
#### References to other Issues or PRs

Fixes #20365
Fixes #21031
Fixes #21606

#### Brief description of what is fixed or changed
`Pow`, `exp`, `factorial` and `gamma` `_eval_as_leading_term` now raise `PoleError` when the argument is unbounded, instead of returning the original expression itself.
Add `LambertW._eval_as_leading_term` method
Fixed `besselj` and `bessely` leading term methods
Added `erfi` leading term method
Added `SingularityFunction` leading term method and power series expansion
Added `fresnels` and `fresnelc` leading term methods

#### Other comments
The previous implementation of `bound` parameter for `aseries` was buggy and misleading, and has broken after these changes. After thinking about what the `bound` parameter should actually represent, it should be coded properly. Consequently, the corresponding tests have been commented out for now.

#7529 and #7552 may be closed now as all the code in those PRs will be present in master
(For #7552, the code is there in #17303, so only one of the two need to remain open)

#### Release Notes

* functions
  * `exp`, `factorial` and `gamma` `._eval_as_leading_term` now raises `PoleError` when the argument is unbounded
  * Add `LambertW`, `erfi`, `fresnels`, `fresnelc` and `SingularityFunction` leading term methods
  * Add `SingularityFunction` power series expansion
  * Fixed `besselj` and `bessely` leading term methods
* series
  * Limit code has better handling of `sign`
* core
  * `Pow._eval_as_leading_term` now raises `PoleError` when the argument is unbounded
  * `leadterm` now takes `logx` argument

PR #21731, opened 2021-07-11
[GSoC] Improvements to `Limit.doit`

### LINKED ISSUE #21530: Incorrect limit
`limit(sinh(n+1)/sinh(n), n,oo)` should give `e` instead of `1`.

### LINKED ISSUE #21701: Incorrect limit with bessel function besselj(m,z)
Hello everyone,
I think I found a bug which was not reported yet. There are a few issues which might be related, but as I don't know the details of the background work, I'm not sure about it. The issues I'm talking about are #11776 and #7224. But they are from the user persepective different from the issue I would like to report about.

When taking the following limit
```py
limit(
    (besselj(m,z)/z**m).subs(m,7),
    z, 0 )
```
the sympy result is `1` for all values of `m`. Here, I have already substituted an arbitrary integer value. In the generic case, without `.subs(m,7)`  an exception is thrown, reporting
```
NotImplementedError: Result depends on the sign of sign(m - 2)
```

However, that's not what I'm talking about. The above result is actually wrong. The limit is not `1` but `1/(2**m * m!)`, see for instance [wolfram alpha](https://www.wolframalpha.com/input/?i=lim++BesselJ%5Bm%2C+x%5D%2Fx%5Em+as+x-%3E0). This can also be derived by hand, using the [series expansion](https://en.wikipedia.org/wiki/Bessel_function#Bessel_functions_of_the_first_kind:_J%CE%B1) from the Frobenius method. When multiplying with `x^{-m}` all terms except the first one `m=0` vanish for `x --> 0`. Any idea, what the origin for the wrong result could be? Is it related to the aforementioned issues?
Thanks in advance!

System:
- Sympy 1.8
- Python 3.8.10
- Windows 10, 20H2

0sidharth:

Fixes #21530 
Fixes #21701 
#21721 

Some heuristics present in the `Limit.doit` code are no longer needed after #21589, so they are removed.
A factor of `1/gamma(nu + 1)` was missing from the `besselj._eval_as_leading_term`, which is now added.

#### Release Notes

* functions
  * Fixed bugs related to direction in `floor` and `ceiling` leading terms
  * Fixed bugs in `sinh` and `cosh` leading terms
  * Added `Ei`, `Ci`, `Shi`, `Chi` leading term methods 
  * `sin` and `cos` now return `AccumBounds(-1, 1)` as the leading term at real infinities

### DIFF
--- a/sympy/functions/elementary/hyperbolic.py
+++ b/sympy/functions/elementary/hyperbolic.py
@@ -234,7 +234,7 @@ class sinh(HyperbolicFunction):
         arg0 = arg.subs(x, 0)
 
         if arg0 is S.NaN:
-            arg0 = arg.limit(x, 0, dir='-' if cdir == -1 else '+')
+            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
         if arg0.is_zero:
             return arg
         elif arg0.is_finite:
@@ -424,7 +424,7 @@ class cosh(HyperbolicFunction):
         arg0 = arg.subs(x, 0)
 
         if arg0 is S.NaN:
-            arg0 = arg.limit(x, 0, dir='-' if cdir == -1 else '+')
+            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
         if arg0.is_zero:
             return S.One
         elif arg0.is_finite:
--- a/sympy/functions/elementary/hyperbolic.py
+++ b/sympy/functions/elementary/hyperbolic.py
@@ -230,13 +230,17 @@ class sinh(HyperbolicFunction):
         return 2*coth_half/(coth_half**2 - 1)
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy import Order
-        arg = self.args[0].as_leading_term(x)
+        arg = self.args[0].as_leading_term(x, logx=logx, cdir=cdir)
+        arg0 = arg.subs(x, 0)
 
-        if x in arg.free_symbols and Order(1, x).contains(arg):
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir == -1 else '+')
+        if arg0.is_zero:
             return arg
+        elif arg0.is_finite:
+            return self.func(arg0)
         else:
-            return self.func(arg)
+            return self
 
     def _eval_is_real(self):
         arg = self.args[0]
@@ -416,13 +420,17 @@ class cosh(HyperbolicFunction):
         return (coth_half + 1)/(coth_half - 1)
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy import Order
-        arg = self.args[0].as_leading_term(x)
+        arg = self.args[0].as_leading_term(x, logx=logx, cdir=cdir)
+        arg0 = arg.subs(x, 0)
 
-        if x in arg.free_symbols and Order(1, x).contains(arg):
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir == -1 else '+')
+        if arg0.is_zero:
             return S.One
+        elif arg0.is_finite:
+            return self.func(arg0)
         else:
-            return self.func(arg)
+            return self
 
     def _eval_is_real(self):
         arg = self.args[0]
--- a/sympy/functions/elementary/integers.py
+++ b/sympy/functions/elementary/integers.py
@@ -137,7 +137,7 @@ class floor(RoundFunction):
         if arg0 is S.NaN:
             arg0 = arg.limit(x, 0, dir='-' if cdir == -1 else '+')
         r = self._eval_nseries(x, n=1, logx=None, cdir=cdir)
-        if r.is_finite:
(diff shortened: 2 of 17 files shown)


sympy/functions/elementary/integers.py
```diff
@@ -132,11 +132,21 @@ def _eval_number(cls, arg):
...
         arg0 = arg.subs(x, 0)
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir == -1 else '+')
+        r = self._eval_nseries(x, n=1, logx=None, cdir=cdir)
         if arg0.is_finite:
-            return self._eval_nseries(x, n=1, logx=None, cdir=cdir)
-        return arg.as_leading_term(x)
+            if r == arg0:
+                direction = (arg - arg0).leadterm(x, cdir=cdir)[0]
+                if direction.is_positive and cdir != -1:
```
jksuom, line 143: `cdir` cannot be used in this condition. It is the direction of `x`, not of `arg` (which may be `-x`).

I think that `dir()` could be applied here.

sympy/functions/elementary/hyperbolic.py
```diff
@@ -230,13 +230,17 @@ def _eval_rewrite_as_coth(self, arg, **kwargs):
...
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy import Order
-        arg = self.args[0].as_leading_term(x)
+        arg = self.args[0].as_leading_term(x, logx=logx, cdir=cdir)
+        arg0 = arg.subs(x, 0)
 
-        if x in arg.free_symbols and Order(1, x).contains(arg):
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir == -1 else '+')
```
jksuom, line 237: `cdir == -1` is too restrictive. `cdir`, when recomputed by `dir`, may have any value. `t*cdir` denotes the same direction for all positive values of `t`. In principle, `cdir` could be normalized to have absolute value `1` but then the expression would often have `sqrt` expressions in the denominator. In this case, a proper condition would be `cdir.is_negative`.

sympy/functions/elementary/trigonometric.py
```diff
@@ -478,12 +478,15 @@ def _eval_expand_trig(self, **hints):
...
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
+        from sympy.calculus.util import AccumBounds
         arg = self.args[0]
         x0 = arg.subs(x, 0).cancel()
         n = x0/S.Pi
         if n.is_integer:
             lt = (arg - n*S.Pi).as_leading_term(x)
             return ((-1)**n)*lt
+        if x0.is_infinite:
```
jksuom, line 488: This would only apply to the real infinities `+/-oo`.

sympy/functions/elementary/integers.py
```diff
@@ -132,11 +132,21 @@ def _eval_number(cls, arg):
             return arg.approximation_interval(Integer)[0]
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        arg = self.args[0]
+        arg = self.args[0].as_leading_term(x, logx=logx, cdir=cdir)
         arg0 = arg.subs(x, 0)
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
+        r = self._eval_nseries(x, n=1, logx=None, cdir=cdir)
```
jksuom, line 139: I would have expected something like this (without the comments):
```
    r = self.subs(x, 0)  # Can this be nan? Probably not in the current examples. 
    if arg0 == r:  # arg0 is an integer
        ndir = arg.dir(x, cdir)  # new direction depending on arg
        return r - 1 if ndir.is_negative else r
```
`dir` is cheaper than `leadterm`. Besides, `leadterm` is usually independent of `cdir` so the coefficient it returns may have wrong sign. (Consider `arg = arg0 + x`: `(arg - arg0).leadterm(x, cdir=cdir)` is `(1, 1)`.)

0sidharth: The problem is that `x.dir(x, cdir)` and `(-x).dir(x, cdir)` are both `0`, when `cdir=0`

jksuom: We should try to make `cdir` nonzero in a case like this where it would matter. I think that is how it is done in `Limit.doit`. Anyway, we could probably set `cdir = 1` if it is zero.

jksuom: We could also raise if `arg0 == r` and `cdir == 0` as then there is not twosided limit.

0sidharth: In that case, should `ceiling(x).series(x)` raise? And what error should it raise, `ValueError`?

jksuom: It seems that a slightly more careful investigation is needed if `cdir == 0`, We could have something like `limit(floor(-x**2, x, 0, dir='+-')`. I think that both `arg.dir(x, 1)` and `arg.dir(x, -1)` should be computed and an error raised only if they disagree. Otherwise the common value can be used in the sequel.

jksuom:
Looks good, thanks!

PR #21737, opened 2021-07-12
[GSoC] Kamke ODE Test Suite

naveensaigit:
This PR aims to make a test suite of the Kamke ODEs in SymPy.

Only basic work is done now. I have added the first 300 ODEs.

#### Release Notes

NO ENTRY


sympy/solvers/ode/tests/ode_test_suit/test_kamke.py
```diff
@@ -0,0 +1,112 @@
...
+        timer.cancel()
+
+
+start = time.time()
+# Read all parsed ODEs from text file
+f = open("kamke_parsed.txt", "r")
+eqs = f.readlines()
+f.close()
+# Open a file to log failing ODEs
+f = open("test_report.txt", "w")
```
oscarbenjamin, line 42: Reading from a file should use `with`:
```
with open(filename) as input_file:
    lines = input_file.readlines()
```
This kind of thing should not be done at top-level in the module (it should be in a function).

naveensaigit: Yes, this was just a very basic draft to get it running.

sympy/solvers/ode/tests/ode_test_suite/test_kamke.py
```diff
@@ -0,0 +1,413 @@
...
+    timer.start()
+    try:
+        yield
+    except KeyboardInterrupt:
+        raise TimeOutError("Timed out for operation {}".format(msg))
+    finally:
+        # Cancel timer if process finishes in time
+        timer.cancel()
+
+A, B, a, a0, a1, a2, a3, a4, alpha, b, b0, b1, b2, b3, b4, bbeta, c, d, ggamma, k, m, n, nu, x = symbols('A, B, a, a0, a1, a2, a3, a4, alpha, b, b0, b1, b2, b3, b4, bbeta, c, d, ggamma, k, m, n, nu, x')
```
asmeurer, line 36: Why is it bbeta and ggamma instead of beta and gamma?

sympy/solvers/ode/tests/ode_test_suite/test_kamke.py
```diff
@@ -0,0 +1,413 @@
...
+        if report_path is None:
+            report_path = "test_report.txt"
+
+        start = time.time()
+        # Counts for result
+        solved = 0
+        check_slow = 0
+        fail = 0
+
+        with open(report_path, "w") as output_file:
```
asmeurer, line 395: I think it would be better to just print the output. If you want it in a file, you can redirect the terminal output.

sympy/solvers/ode/tests/ode_test_suite/test_kamke.py
```diff
@@ -0,0 +1,413 @@
...
+                sol, output = self.test_example(example, hint, verify, dsolve_timelim, checkodesol_timelim, log=True)
+                output_file.write(output)
+                output_file.flush()
+
+            # Summary
+            output_file.write("Total time taken" + str(time.time() - start))
+            output_file.write("No. of ODEs solved:" + str(solved))
+            output_file.write("No. of ODEs failed:" + str(fail))
+            output_file.write("No. of ODEs solved, but not checked:" + str(check_slow))
+            output_file.close()
```
asmeurer, line 413: At the end of the file you can put 

```py
if __name__ == '__main__':
    test_all_examples()
```

and you will be able to run it as a script (`python -m sympy.solvers.ode.tests.test_kamke`). You could also put a little argparse parsing in a function to add the options are command line arguments if you wanted.

sympy/solvers/ode/tests/ode_test_suite/test_kamke.py
```diff
@@ -0,0 +1,413 @@
...
+
+        try:
+            # Try to find the solution to the equation
+            with time_limit(dsolve_timelim):
+                sol = dsolve(eq, y(x), hint)
+                if isinstance(sol, Eq):
+                    sol = [sol]
+        except (TimeOutError, ValueError, NotImplementedError, TypeError):
+            # Solution not found / timeout
+            output += "dsolve failed"
```
asmeurer, line 361: You should actually record the full exception.

sympy/solvers/ode/tests/ode_test_suite/test_kamke.py
naveensaigit: I used `bbeta` and `ggamma` since `beta` and `gamma` might be used in the ODEs. That is not the case though, so I will rename them.

naveensaigit:
I've now added all examples from Kamke. They are segregated into chapters based on the classification given on [this](https://12000.org) website. A particular example/chapter or even all examples can be tested. The code generates a HTML page for each ODE. It also logs the status onto the terminal. The following is an example of the output in the terminal -
```
py sympy/solvers/ode/tests/test_kamke.py -ch 5

kamke_5.1 dsolve failed in 1.082000732421875 seconds
solve: Cannot solve a**4*y(x) - 2*a**2*Derivative(y(x), (x, 2)) - lambda_*(a*x - b)*(-a**2*y(x) + Derivative(y(x), (x, 2))) + Derivative(y(x), 
(x, 4))

kamke_5.2 solved in 1.0556578636169434 seconds
Equation: -a*x - b*sin(x) - c*cos(x) + Derivative(y(x), x) + 2*Derivative(y(x), (x, 3)) + Derivative(y(x), (x, 5))
Solution: [Eq(y(x), C1 + a*x**2/2 + (C2 + x*(C3 - c*x/8))*sin(x) + (C4 + x*(C5 + b*x/8))*cos(x))]

kamke_5.3 dsolve failed in 10.02199912071228 seconds
Timed out for operation: dsolve

kamke_5.4 dsolve failed in 0.3489997386932373 seconds
solve: Cannot solve -a*x*y(x) - b + Derivative(y(x), (x, 5))

kamke_5.5 dsolve failed in 0.8682498931884766 seconds
solve: Cannot solve a*nu*x**(nu - 1)*y(x) + a*x**nu*Derivative(y(x), x) + Derivative(y(x), (x, 5))

kamke_5.6 solved in 0.3670005798339844 seconds
Equation: a*Derivative(y(x), (x, 4)) - f + Derivative(y(x), (x, 5))
Solution: [Eq(y(x), C1 + C2*x + C3*x**2 + C4*x**3 + C5*exp(-a*x) + f*x**4/(24*a))]

kamke_5.7 dsolve failed in 0.40700387954711914 seconds
solve: Cannot solve a*x*y(x) - 5*m*Derivative(y(x), (x, 4)) + x*Derivative(y(x), (x, 5))

kamke_5.8 dsolve failed in 10.023767948150635 seconds
Timed out for operation: dsolve

kamke_5.10 dsolve failed in 0.42600011825561523 seconds
solve: Cannot solve -a*y(x) + x**5*Derivative(y(x), (x, 10))

kamke_5.11 dsolve failed in 0.3120002746582031 seconds
solve: Cannot solve -a*y(x) + x**10*Derivative(y(x), (x, 5))

kamke_5.12 dsolve failed in 0.5030002593994141 seconds
solve: Cannot solve -a*y(x) + x**5.5*Derivative(y(x), (x, 11))

kamke_5.13 dsolve failed in 5.191370964050293 seconds
solve: Cannot solve -c*y(x) + (-a + x)**5*(-b + x)**5*Derivative(y(x), (x, 5))

Total time taken: 41.558265686035156
No. of ODEs solved: 2
No. of ODEs failed: 10
No. of ODEs solved, but not checked: 0
```

naveensaigit:
The quality check fails because I am using a multi-line f-string. Is there any fix for this?

oscarbenjamin:
The error message says `SyntaxError: invalid escape sequence \(`. Does your string contain those characters?

naveensaigit:
Yes, I am using MathJax in the html and it requires using `\(`. It runs fine locally. I don't know why it says it is a syntax error.

oscarbenjamin:
> Yes, I am using MathJax in the html and it requires using `\(`

Apart from in rawstrings in Python `\x` means some kind of escape character. If you want a literal `\` then you need to use two slashes: `\\(`. Older versions of Python ignored invalid escape sequences but the test that fails runs in a mode that warns about invalid escapes.

oscarbenjamin:
It would be good if there was a main html page that just gave a short listing in a long table of each of the different ODEs and linked to them. Something like:
```
name | order | linearity | solved | hint | checked | time
kamke_1.123 | 1st | nonlinear | solved | separable | unchecked 1 sec
kamke_1.124 | 1st | nonlinear | not solved | separable | checked | N/A
...
```
Then the names could like `kamke_1.123` could be links that lead to the page for the ODE with more detailed information about that particular example.

Also I would put all of the code at the top of the file (with a main function at the very top) so that at the bottom it's just something like:
```
if __name__ == "__main__":
    import sys
    main(*sys.argv[1:])
```

The docstring at the top should give more of an explanation.

There should be a test-case for the change to checkodesol.

naveensaigit:
I've now added code to automatically generate a main page which links to different chapter pages. It looks something like this-

image

The chapter page links to a page for each example. It looks like this -

image

Finally, the example page looks like this -

image


One thing that is unclear to me is what the example page should look like for a system of ODEs (present in chapter 8 and 9). There are multiple ODEs in each system, so will they all be of the same order and linearity? Also, is there a function to classify system of ODEs like `classify_ode`? I tried using `classify_ode`, but it only works for individual ODEs.

oscarbenjamin:
That looks nice. I guess it would be good to know what the other hints give for any particular ODE but let's just see if we can get the whole thing to run first.

> There are multiple ODEs in each system, so will they all be of the same order and linearity?

I don't know about the examples in Kamke/Murphy but in general a system of ODEs can have mixed order e.g.:
```python
In [4]: eqs = [f(x).diff(x) - g(x), g(x).diff(x, 2) - x]

In [5]: eqs
Out[5]: 
⎡                         2      ⎤
⎢        d               d       ⎥
⎢-g(x) + ──(f(x)), -x + ───(g(x))⎥
⎢        dx               2      ⎥
⎣                       dx       ⎦

In [6]: dsolve(eqs)
Out[6]: 
⎡                       2    4                      3⎤
⎢                   C₃⋅x    x                      x ⎥
⎢f(x) = C₁ + C₂⋅x + ───── + ──, g(x) = C₂ + C₃⋅x + ──⎥
⎣                     2     24                     6 ⎦
```
I guess a system like that could have the order listed as `1st,2nd` or something.

> Also, is there a function to classify system of ODEs like classify_ode? I tried using classify_ode

There is `classify_sysode` which is used by `dsolve` for systems:
```python
In [7]: from sympy.solvers.ode.ode import classify_sysode

In [8]: classify_sysode(eqs)
Out[8]: 
{'no_of_equation': 2,
 'eq': [-g(x) + Derivative(f(x), x), -x + Derivative(g(x), (x, 2))],
 'func': [f(x), g(x)],
 'order': {g(x): 2, f(x): 1},
 'func_coeff': {(0, f(x), 0): 0,
  (0, f(x), 1): 1,
  (0, g(x), 0): -1,
  (0, g(x), 1): 0,
  (0, g(x), 2): 0,
  (1, f(x), 0): 0,
  (1, f(x), 1): 0,
  (1, g(x), 0): 0,
  (1, g(x), 1): 0,
  (1, g(x), 2): 1},
 'is_linear': True,
 'type_of_equation': None}
```
The newer systems code added last year uses a different system though for classifying linear ODEs:
```python
In [5]: from sympy.solvers.ode.systems import _classify_linear_system

In [6]: eqs = [f(x).diff(x) - g(x), g(x).diff(x, 2) - x]

In [7]: _classify_linear_system(eqs, [f(x), g(x)], x)
Out[7]: 
{'no_of_equation': 2,
 'eq': [-g(x) + Derivative(f(x), x), -x + Derivative(g(x), (x, 2))],
 'func': [f(x), g(x)],
 'order': {f(x): 1, g(x): 2},
 'is_linear': True,
 'is_homogeneous': False,
 'is_general': True,
 'rhs': Matrix([
 [0],
 [x]]),
 'type_of_equation': 'type0',
 'is_higher_order': True}
```
Here I think `type0` is the key that determines how it is solved:
https://github.com/sympy/sympy/blob/4f50dd78e42fb9b132ecd0073bb72a70992c0b68/sympy/solvers/ode/systems.py#L1801-L1802

oscarbenjamin:
For basic website navigation it would be good if each page has a "home" link that takes you back to the main page. The main page should have some text that explains what the website is showing.

naveensaigit:
> That looks nice. I guess it would be good to know what the other hints give for any particular ODE but let's just see if we can get the whole thing to run first.

All the matching hints for a particular ODE are listed in the example page.

> For basic website navigation it would be good if each page has a "home" link that takes you back to the main page. The main page should have some text that explains what the website is showing.

Yes, this is a nice idea. I will work on it.

oscarbenjamin:
> All the matching hints for a particular ODE are listed in the example page.

What I mean is that the different hints do not necessarily give the same form for the solution. It would be nice to see the different solutions that they return. That would make generating the pages a lot slower though because every ODE would need to be solved with every different hint.

oscarbenjamin:
It would be nice if each html page had a code repr for the example being tested so you could copy/paste to try it out.

oscarbenjamin:
> I'm sorry, I don't know what a repr is. Is it like a code snippet?

I mean the output of the Python function `repr`:
```python
In [109]: eq = f(x).diff(x) - 2

In [110]: repr(eq)
Out[110]: 'Derivative(f(x), x) - 2'
```
Ideally there would be a code snippet like:
```python
from sympy import *
y = Function('y')
x = Symbol('x')
a, b, c = symbols('a, b, c')
eq = Derivative(y(x), x) + a*y(x) + b*x + c
sol = dsolve(eq)
checkodesol(eq, sol)
```
Then when looking at a particular example you can copy paste the code to try it out yourself.

oscarbenjamin:
> What should I do to make the `summary.html` the home page for the github pages deployment?

It seems to be working if I go to the URL:
https://naveensaigit.github.io/kamke_test_suite/

naveensaigit:
> It seems to be working if I go to the URL:
> https://naveensaigit.github.io/kamke_test_suite/

Yes, I think the last change I pushed in took sometime to reflect. Now the only thing left is to batch the ODEs and run different jobs for each of them.

naveensaigit:
If I split it up into different jobs, the files generated in only the last chapter are being committed. @oscarbenjamin Could you please take a look in the `gh-pages` branch?

oscarbenjamin:
> Could you please take a look in the `gh-pages` branch?

I guess you mean this:
https://github.com/naveensaigit/kamke_test_suite/blob/b096812a9db1b6b0def6a28f91776d8d5d6e6d0d/.github/workflows/build.yml#L1

The problem is that a force push from one job will overwrite the commits from the other. You should avoid force pushing in automatic scripts like this because it is quite dangerous if something goes wrong (e.g. if the repo wipes all commits or something).

I think that you should have each job save its output as an artifact and then have a final job that downloads all artifacts, updates and generates commits once only.

This will be easier if each job saves its results in a json file. Then the final job collects all the json files and updates the html files, commits the json and html (maybe as two separate commits) and finally pushes (not using a force push).

Using `git add .` is also risky. You should be as specific as possible e.g. `git add somefile.json` and then use the output of `git status` to confirm that there are no other unexpected files.

naveensaigit:
> The problem is that a force push from one job will overwrite the commits from the other. You should avoid force pushing in automatic scripts like this because it is quite dangerous if something goes wrong (e.g. if the repo wipes all commits or something).

Yes, so should I pull before pushing normally?

> I think that you should have each job save its output as an artifact and then have a final job that downloads all artifacts, updates and generates commits once only.

Do files from different jobs stay in memory after the job is completed? I wasn't sure about this. Also, where would the artifacts be stored?

oscarbenjamin:
> Yes, so should I pull before pushing normally?

No just collect the artifacts together and have a single job make the commit so each update is made as a single commit.

> Do files from different jobs stay in memory after the job is completed? I wasn't sure about this. Also, where would the artifacts be stored?

See here:
https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts

naveensaigit:
I've now made it such that each chapter runs as a separate job and saves a json file for each example as an artifact. Finally, there is one job which downloads all the artifacts, generates html and pushes it. You can now check the page at  https://naveensaigit.github.io/kamke_test_suite/

PR #21743, opened 2021-07-14
Improve dsolve

Mohitbalwani26:

#18348 

Earlier the flow of code was dsolve -> _desolve -> classify_ode (which checked for every hint whether it matches and return all the matching hints) -> particular solver's general solution was called.

This PR now changes the flow in classify_ode as whenever the user calls with explicit hint, It won't be matching all the possible solvers. It will just create the instance of that particular class and if it matches it will return.

#### Other comments
This also significantly improves the speed of test suite.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/solvers/ode/ode.py
+++ b/sympy/solvers/ode/ode.py
@@ -608,7 +608,7 @@ def dsolve(eq, func=None, hint="default", simplify=True,
         if all_:
             retdict = {}
             failed_hints = {}
-            gethints = classify_ode(eq, dict=True,hint='all')
+            gethints = classify_ode(eq, dict=True, hint='all')
             orderedhints = gethints['ordered_hints']
             for hint in hints:
                 try:
@@ -1034,7 +1034,8 @@ def classify_ode(eq, func=None, dict=False, ics=None, *, prep=True, xi=None, eta
             if user_map[hint].has_integral:
                 matching_hints[hint + "_Integral"] = solver
             if dict and early_exit:
-                break
+                matching_hints["default"] = hint
+                return matching_hints
 
     eq = expand(eq)
     # Precondition to try remove f(x) from highest order derivative
--- a/sympy/solvers/ode/ode.py
+++ b/sympy/solvers/ode/ode.py
@@ -3565,16 +3565,3 @@ def _nonlinear_3eq_order1_type5(x, y, z, t, eq):
     sol2 = dsolve(diff(fv(t), t) - (v*(a*F3 - c*F1)).subs(u, x_y).subs(w, z_y).subs(v, fv(t))).rhs
     sol3 = dsolve(diff(fw(t), t) - (w*(b*F1 - a*F2)).subs(u, x_z).subs(v, y_z).subs(w, fw(t))).rhs
     return [sol1, sol2, sol3]
-
-
-#This import is written at the bottom to avoid circular imports.
-from .single import (NthAlgebraic, Factorable, FirstLinear, AlmostLinear,
-        Bernoulli, SingleODEProblem, SingleODESolver, RiccatiSpecial,
-        SecondNonlinearAutonomousConserved, FirstExact, Liouville, Separable,
-        SeparableReduced, HomogeneousCoeffSubsDepDivIndep, HomogeneousCoeffSubsIndepDivDep,
-        HomogeneousCoeffBest, LinearCoefficients, NthOrderReducible, SecondHypergeometric,
-        NthLinearConstantCoeffHomogeneous, NthLinearConstantCoeffVariationOfParameters,
-        NthLinearConstantCoeffUndeterminedCoefficients, NthLinearEulerEqHomogeneous,
-        NthLinearEulerEqNonhomogeneousVariationOfParameters, LieGroup,
-        NthLinearEulerEqNonhomogeneousUndeterminedCoefficients, SecondLinearBessel, SecondLinearAiry,
-        solver_map)
--- a/sympy/solvers/ode/ode.py
+++ b/sympy/solvers/ode/ode.py
@@ -1017,42 +1017,54 @@ def classify_ode(eq, func=None, dict=False, ics=None, *, prep=True, xi=None, eta
     # Any ODE that can be solved with a combination of algebra and
     # integrals e.g.:
     # d^3/dx^3(x y) = F(x)
+    solver_map = {
+        'factorable': Factorable,
+        'nth_algebraic': NthAlgebraic,
+        'separable': Separable,
+        '1st_exact': FirstExact,
+        '1st_linear': FirstLinear,
+        'Bernoulli': Bernoulli,
+        'Riccati_special_minus2': RiccatiSpecial,
+        '1st_homogeneous_coeff_best': HomogeneousCoeffBest,
+        '1st_homogeneous_coeff_subs_indep_div_dep': HomogeneousCoeffSubsIndepDivDep,
+        '1st_homogeneous_coeff_subs_dep_div_indep': HomogeneousCoeffSubsDepDivIndep,
+        'almost_linear': AlmostLinear,
+        'linear_coefficients': LinearCoefficients,
+        'separable_reduced': SeparableReduced,
+        'lie_group': LieGroup,
+        'nth_linear_constant_coeff_homogeneous': NthLinearConstantCoeffHomogeneous,
+        'nth_linear_euler_eq_homogeneous': NthLinearEulerEqHomogeneous,
+        'nth_linear_constant_coeff_undetermined_coefficients': NthLinearConstantCoeffUndeterminedCoefficients,
+        'nth_linear_euler_eq_nonhomogeneous_undetermined_coefficients': NthLinearEulerEqNonhomogeneousUndeterminedCoefficients,
+        'nth_linear_constant_coeff_variation_of_parameters': NthLinearConstantCoeffVariationOfParameters,
+        'nth_linear_euler_eq_nonhomogeneous_variation_of_parameters': NthLinearEulerEqNonhomogeneousVariationOfParameters,
+        'Liouville': Liouville,
+        '2nd_linear_airy': SecondLinearAiry,
+        '2nd_linear_bessel': SecondLinearBessel,
+        '2nd_hypergeometric': SecondHypergeometric,
+        'nth_order_reducible': NthOrderReducible,
+        '2nd_nonlinear_autonomous_conserved': SecondNonlinearAutonomousConserved,
+        }
+
     ode = SingleODEProblem(eq_orig, func, x, prep=prep, xi=xi, eta=eta)
-    solvers = {
-        NthAlgebraic: ('nth_algebraic',),
-        FirstExact:('1st_exact',),
-        FirstLinear: ('1st_linear',),
-        AlmostLinear: ('almost_linear',),
-        Bernoulli: ('Bernoulli',),
-        Factorable: ('factorable',),
-        RiccatiSpecial: ('Riccati_special_minus2',),
-        SecondNonlinearAutonomousConserved: ('2nd_nonlinear_autonomous_conserved',),
-        Liouville: ('Liouville',),
-        Separable: ('separable',),
-        SeparableReduced: ('separable_reduced',),
-        HomogeneousCoeffSubsDepDivIndep: ('1st_homogeneous_coeff_subs_dep_div_indep',),
-        HomogeneousCoeffSubsIndepDivDep: ('1st_homogeneous_coeff_subs_indep_div_dep',),
-        HomogeneousCoeffBest: ('1st_homogeneous_coeff_best',),
-        LinearCoefficients: ('linear_coefficients',),
-        NthOrderReducible: ('nth_order_reducible',),
-        SecondHypergeometric: ('2nd_hypergeometric',),
-        NthLinearConstantCoeffHomogeneous: ('nth_linear_constant_coeff_homogeneous',),
-        NthLinearConstantCoeffVariationOfParameters: ('nth_linear_constant_coeff_variation_of_parameters',),
-        NthLinearConstantCoeffUndeterminedCoefficients: ('nth_linear_constant_coeff_undetermined_coefficients',),
-        NthLinearEulerEqHomogeneous: ('nth_linear_euler_eq_homogeneous',),
-        NthLinearEulerEqNonhomogeneousVariationOfParameters: ('nth_linear_euler_eq_nonhomogeneous_variation_of_parameters',),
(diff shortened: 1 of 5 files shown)


sympy/solvers/ode/ode.py
```diff
@@ -1017,42 +1017,54 @@ class in it.  Note that a hint may do this anyway if
     # Any ODE that can be solved with a combination of algebra and
     # integrals e.g.:
     # d^3/dx^3(x y) = F(x)
+    solver_map = {
```
oscarbenjamin, line 1020: This should probably be a module level constant. Perhaps it should be defined at the bottom of `single.py`.

sympy/solvers/ode/ode.py
```diff
@@ -1017,42 +1017,54 @@ class in it.  Note that a hint may do this anyway if
...
+            matching_hints[user_hint] = solver
+            if solver_map[user_hint].has_integral:
+                matching_hints[user_hint + "_Integral"] = solver
+    else:
+        for hint in solver_map:
+            solver = solver_map[hint](ode)
+            if solver.matches():
+                matching_hints[hint] = solver
+                if solver_map[hint].has_integral:
+                    matching_hints[hint + "_Integral"] = solver
```
oscarbenjamin, line 1066: I would probably restructure this slightly:
```python
# Used when dsolve is called without an explicit hint.
# We exit early to return the first valid match
early_exit = (user_hint == 'default')

# An explicit hint has been given to dsolve
# Skip matching code for other hints
if user_hint not in ('default', 'all', 'all_Integral', 'best'):
    solver_map = ... # only the relevant items

for hint, cls in solver_map.items()
   ...
   if solver.matches():
       ...
       if early_exit:
           # maybe there needs to be a check here because of the unrefactored
           # solvers that are matched below. I guess this should only return if the
           # hint has a higher priority than all of those...
           return matching_hints
```
I suppose ultimately if all solvers were refactored then we wouldn't need to call `classify_ode` at all when a hint is given but for now the explicit hint case can be handled here.

oscarbenjamin:
Can you write an explanation of what this does in the OP?

sympy/solvers/ode/ode.py
Mohitbalwani26: I am still trying to figure out what to do with the `default` hint as Factorable and NthorderReducible solvers call dsolve recursively.

Mohitbalwani26:
> Can you write an explanation of what this does in the OP?

sure, I will update it

oscarbenjamin:
The OP doesn't explain the early exit

oscarbenjamin:
> This also significantly improves the speed of test suite.

Can you give numbers for this?

sympy/solvers/ode/ode.py
oscarbenjamin: You have changed this code now but you seemed to have missed part of the point which is that by changing `solver_map` we can avoid duplicating the code that calls match.

Mohitbalwani26:
> Can you give numbers for this?

1) `bin/doctest sympy/solvers/ode` took **233.06** seconds on master and **195.56** seconds on this branch.

2) ```python 
    In [2]: import sympy
    In [3]: %time sympy.test('sympy/solvers/ode/tests/test_ode.py')   #on master
    CPU times: user 644 µs, sys: 7.56 ms, total: 8.2 ms
    Wall time: 1min 51s
    ***************************************************************************************************
    In [2]: %time sympy.test('sympy/solvers/ode/tests/test_ode.py')   #on this Branch
    CPU times: user 1.43 ms, sys: 8.07 ms, total: 9.5 ms
    Wall time: 1min 25s
    ```
3) ```python 
    In [2]: import sympy
    In [3]: %time sympy.test('sympy/solvers/ode/tests/test_single.py')   #on master
    CPU times: user 0 ns, sys: 9.96 ms, total: 9.96 ms
    Wall time: 25min 39s
    ***************************************************************************************************
    In [2]: %time sympy.test('sympy/solvers/ode/tests/test_single.py')   #on this Branch
    CPU times: user 4.75 ms, sys: 4.29 ms, total: 9.04 ms
    Wall time: 24min 43s
    ```

These doesn't include examples marked as XFAIL or Slow.
Also can you suggest some way to see the difference in speed if this is not the correct way.

oscarbenjamin:
Timing the test suite is good. I think that the best thing to do is to focus on a particular example and you should start with the simplest examples first. So let's take an ODE that should be trivial and time solving it on master:
```python
In [1]: eq = f(x).diff(x, 4)

In [2]: eq
Out[2]: 
  4      
 d       
───(f(x))
  4      
dx       

In [3]: %time sol = dsolve(eq, f(x))
CPU times: user 1.37 s, sys: 15.2 ms, total: 1.38 s
Wall time: 1.38 s

In [4]: sol
Out[4]: 
                       2       3
f(x) = C₁ + C₂⋅x + C₃⋅x  + C₄⋅x 
```
Clearly it should be possible to compute this in much less than 1 second. There are several hints that match this and we can try others:
```python
In [5]: classify_ode(eq)
Out[5]: 
('nth_algebraic',
 'nth_linear_constant_coeff_homogeneous',
 'nth_linear_euler_eq_homogeneous',
 'nth_algebraic_Integral')

In [6]: %time sol = dsolve(eq, f(x), hint='nth_linear_constant_coeff_homogeneous')
CPU times: user 1.79 s, sys: 22.3 ms, total: 1.81 s
Wall time: 1.85 s
```
With the PR we get better time if we specify the right hint:
```python
In [1]: eq = f(x).diff(x, 4)

In [2]: %time sol = dsolve(eq, f(x))
CPU times: user 1.28 s, sys: 13.4 ms, total: 1.3 s
Wall time: 1.3 s

In [3]: %time sol = dsolve(eq, f(x), hint='nth_linear_constant_coeff_homogeneous')
CPU times: user 96.3 ms, sys: 2.45 ms, total: 98.8 ms
Wall time: 97.7 ms
```
We should be able to make this a lot faster still but you can see that it can be 10x faster if we specify the right hint. However by default it is still much slower than that which is because the `nth_algebraic` solver is called first and that is slow.

We can use a profiler to see what is slow:
```python
In [4]: %prun -s cumulative dsolve(eq)

         2324044 function calls (2175047 primitive calls) in 3.251 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    197/1    0.036    0.000    3.252    3.252 {built-in method builtins.exec}
        1    0.000    0.000    3.252    3.252 ode.py:354(dsolve)
      2/1    0.000    0.000    2.945    2.945 deutils.py:134(_desolve)
        1    0.000    0.000    2.938    2.938 ode.py:811(classify_ode)
        3    0.000    0.000    2.924    0.975 single.py:283(matches)
        1    0.000    0.000    2.905    2.905 single.py:380(_matches)
        2    0.000    0.000    2.677    1.339 solvers.py:379(solve)
      3/2    0.000    0.000    2.672    1.336 solvers.py:1281(_solve)
     79/3    0.008    0.000    2.557    0.852 simplify.py:411(simplify)
        2    0.000    0.000    2.503    1.252 solvers.py:1731(<listcomp>)
        2    0.000    0.000    2.503    1.251 solvers.py:187(checksol)
  790/388    0.010    0.000    2.390    0.006 basic.py:1241(replace)
 8284/388    0.017    0.000    2.361    0.006 basic.py:1466(walk)
8284/3376    0.010    0.000    2.341    0.001 basic.py:1488(rec_replace)
```
Line 380 of single.py is the NthAlgebraic match method and it is slow because it calls `solve` which then also calls `simplify`.

I don't know whether the `nth_algebraic` solver is still needed any more but it is very slow so we should either make it faster or get rid of it. In any case a method like constant coefficients can get the solution without integrals and so should always be preferred over a method that computes the result using integrals.

We can also check what makes it slow when using the hint:
```python
In [4]: %prun -s cumulative sol = dsolve(eq, f(x), hint='nth_linear_constant_coeff_homogeneous')

         141421 function calls (133382 primitive calls) in 0.211 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.211    0.211 {built-in method builtins.exec}
        1    0.000    0.000    0.211    0.211 <string>:1(<module>)
        1    0.000    0.000    0.210    0.210 ode.py:354(dsolve)
        1    0.000    0.000    0.178    0.178 ode.py:640(_helper_simplify)
        1    0.000    0.000    0.103    0.103 single.py:292(get_general_solution)
        1    0.000    0.000    0.103    0.103 single.py:2133(_get_general_solution)
        1    0.000    0.000    0.093    0.093 nonhomogeneous.py:251(_get_simplified_sol)
   162/12    0.007    0.000    0.085    0.007 radsimp.py:21(collect)
        1    0.000    0.000    0.075    0.075 ode.py:671(<listcomp>)
      3/1    0.000    0.000    0.075    0.075 multidimensional.py:105(wrapper)
        1    0.000    0.000    0.075    0.075 ode.py:1561(odesimp)
        2    0.000    0.000    0.069    0.034 ode.py:1916(constantsimp)
2638/2253    0.006    0.000    0.052    0.000 cache.py:69(wrapper)
      582    0.007    0.000    0.039    0.000 basic.py:1560(match)
  250/144    0.000    0.000    0.038    0.000 {built-in method builtins.sum}
       32    0.000    0.000    0.037    0.001 basic.py:1519(count)
      448    0.001    0.000    0.037    0.000 basic.py:1522(<genexpr>)
    91/32    0.005    0.000    0.035    0.001 function.py:1268(__new__)
       30    0.001    0.000    0.035    0.001 basic.py:765(subs)
        2    0.000    0.000    0.033    0.017 ode.py:1863(__remove_linear_redundancies)
```
Here it is slow because of simplification. If we disable that then it is faster:
```python
In [5]: %time sol = dsolve(eq, f(x), hint='nth_linear_constant_coeff_homogeneous', simplify=False)
CPU times: user 22.9 ms, sys: 1.76 ms, total: 24.6 ms
Wall time: 24.9 ms
```
Now 25ms is a lot better than 1.3 seconds but I'm sure there are many things that can be optimised here because this really is a simple ODE.

The simplification code here can be improved I think because that it currently quite slow.

Mohitbalwani26:
@oscarbenjamin I think first we should take a look on the ordering of hints. should I take some example which can be solved by every hint and then according to the time I should place them?

oscarbenjamin:
I do think that we should look at changing the order of the hints and optimising things but it should be separate from this PR

Mohitbalwani26:
> I do think that we should look at changing the order of the hints and optimising things but it should be separate from this PR

so any changes you want me to make in this PR before getting it merged?

sympy/solvers/ode/ode.py
```diff
@@ -608,7 +608,7 @@ def recur_len(l):
         if all_:
             retdict = {}
             failed_hints = {}
-            gethints = classify_ode(eq, dict=True)
+            gethints = classify_ode(eq, dict=True,hint='all')
```
oscarbenjamin, line 611: Space after comma

sympy/solvers/ode/ode.py
```diff
@@ -1014,45 +1014,28 @@ class in it.  Note that a hint may do this anyway if
...
         if solver.matches():
-            for hints in solvers[solvercls]:
-                matching_hints[hints] = solver
-                if solvercls.has_integral:
-                    matching_hints[hints + "_Integral"] = solver
+            matching_hints[hint] = solver
+            if user_map[hint].has_integral:
+                matching_hints[hint + "_Integral"] = solver
+            if dict and early_exit:
+                break
```
oscarbenjamin, line 1037: Why is this break rather than return?

Presumably it's because some of the other hints are not yet refactored and might come earlier than the match that we just found. Which hints are not refactored?

Looking at `allhints` it seems that everything is refactored apart from the series solvers. If that's the case then I think it might be better to just return here. I don't see why we would want to return a series solution when it is still possible to return a non-series solution. Note that the matching code for the series solvers can be slow.

sympy/solvers/ode/ode.py
```diff
@@ -1014,45 +1014,29 @@ class in it.  Note that a hint may do this anyway if
...
-        NthLinearEulerEqHomogeneous: ('nth_linear_euler_eq_homogeneous',),
-        NthLinearEulerEqNonhomogeneousVariationOfParameters: ('nth_linear_euler_eq_nonhomogeneous_variation_of_parameters',),
-        NthLinearEulerEqNonhomogeneousUndeterminedCoefficients: ('nth_linear_euler_eq_nonhomogeneous_undetermined_coefficients',),
-        SecondLinearBessel: ('2nd_linear_bessel',),
-        SecondLinearAiry: ('2nd_linear_airy',),
-        LieGroup: ('lie_group',),
-    }
-    for solvercls in solvers:
-        solver = solvercls(ode)
+    user_hint = kwargs.get('hint','default')
```
oscarbenjamin, line 1018: There should be a space after a comma.

oscarbenjamin:
There should be tests for the new behaviour of `classify_ode`.

Mohitbalwani26:
> There should be tests for the new behaviour of `classify_ode`.

@oscarbenjamin can you please explain a bit what the test would be as behavior remains same.

oscarbenjamin:
The behaviour is not the same if there is an early exit

sympy/solvers/ode/tests/test_ode.py
```diff
@@ -298,6 +298,11 @@ def test_classify_ode():
         ('1st_power_series', 'lie_group')
     assert isinstance(classify_ode(Eq(f(x), 5), f(x), dict=True), dict)
 
+    #This is for new behavior of classify_ode when called internally with default, It should
+    # return the first hint which matches therefore, 'ordered_hints' key will not be there.
+    assert classify_ode(Eq(f(x).diff(x), 0), f(x),dict=True).get('ordered_hints') == None
+    assert classify_ode(Eq(f(x).diff(x), 0), f(x),dict=True).get('default') == 'nth_algebraic'
```
oscarbenjamin, line 304: There should be a space after a comma.

Why doesn't this assert what the actual output of `classify_ode` is?

Also is the behaviour changed in cases where a hint is specified?

Mohitbalwani26: > Why doesn't this assert what the actual output of `classify_ode` is?

ok, will change it to that.

> Also is the behaviour changed in cases where a hint is specified?

Earlier classify_ode didn't take hint as argument but now it has so if we pass specific hint it will return solver for that and also checks for the series solution

Mohitbalwani26: Also the output of this test is something like:
```
{'order': 1,
 'nth_algebraic': <sympy.solvers.ode.single.NthAlgebraic at 0x7f94c63cc160>,
 'nth_algebraic_Integral': <sympy.solvers.ode.single.NthAlgebraic at 0x7f94c63cc160>,
 'default': 'nth_algebraic'}
```
so how should we check with assert?

oscarbenjamin:
Looks good. Thanks!

PR #21747, opened 2021-07-15
Docs: Getting started category

nanjekyejoannah:

This PR categorizes the docs according to the diataxis format.

NO ENTRY

### DIFF
--- a/doc/src/conf.py
+++ b/doc/src/conf.py
@@ -27,8 +27,12 @@ sys.path = ['ext'] + sys.path
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.addons.*') or your custom ones.
 extensions = ['sphinx.ext.autodoc', 'sphinx.ext.linkcode', 'sphinx_math_dollar',
-              'sphinx.ext.mathjax', 'numpydoc', 'sympylive',
+              'sphinx.ext.mathjax', 'numpydoc', 'sympylive', 'sphinx_reredirects',
               'sphinx.ext.graphviz', 'matplotlib.sphinxext.plot_directive']
+
+redirects = {
+    "install.rst": "getting_started/install.rst",
+}
 
 # Use this to use pngmath instead
 #extensions = ['sphinx.ext.autodoc', 'sphinx.ext.viewcode', 'sphinx.ext.pngmath', ]
@@ -123,7 +127,8 @@ html_static_path = ['_static']
 # using the given strftime format.
 html_last_updated_fmt = '%b %d, %Y'
 
-html_theme = 'classic'
+# was classic
+html_theme = "classic"
 
 html_logo = '_static/sympylogo.png'
 html_favicon = '../_build/logo/sympy-notailtext-favicon.ico'
--- a/doc/src/getting_started/install.rst
+++ b/doc/src/getting_started/install.rst
@@ -1,7 +1,7 @@
 .. _installation:
 
 Installation
-------------
+=============
 
 The SymPy CAS can be installed on virtually any computer with Python.
 SymPy does require `mpmath`_ Python library to be installed first.  The
@@ -12,7 +12,7 @@ distributions have SymPy packages available.
 SymPy officially supports Python 3.5, 3.6, 3.7, and PyPy.
 
 Anaconda
-========
+---------
 
 `Anaconda <https://www.anaconda.com/download/>`_ is a free Python distribution from
 Continuum Analytics that includes SymPy, Matplotlib, IPython, NumPy, and many
@@ -28,7 +28,7 @@ use::
     conda update sympy
 
 Git
-===
+----
 
 If you wish to contribute to SymPy or like to get the latest updates as they
 come, install SymPy from git. To download the repository, execute the
@@ -49,7 +49,7 @@ This will cause the installed version to always point to the version in the git
 directory.
 
 Other Methods
-=============
+--------------
 
 You may also install SymPy using pip or from source. In addition, most Linux
 and Python distributions have some SymPy version available to install using
@@ -61,7 +61,7 @@ their package manager. Here is a list of several such Python distributions:
 * `Spack <https://spack.io/>`_
 
 Run SymPy
-=========
+---------
 
 After installation, it is best to verify that your freshly-installed SymPy
 works. To do this, start up Python and import the SymPy libraries::
@@ -80,7 +80,7 @@ From here, execute some simple SymPy statements like the ones below::
 For a starter guide on using SymPy effectively, refer to the :ref:`tutorial`.
 
 Mpmath
-======
+-------
 
 Versions of SymPy prior to 1.0 included `mpmath`_, but it now depends on it as
 an external dependency.  If you installed SymPy with Anaconda, it will already
@@ -104,7 +104,7 @@ before the code that imports ``sympy.mpmath``. It is recommended to change
 code that uses ``sympy.mpmath`` to use ``mpmath`` directly wherever possible.
 
 Questions
-=========
+----------
 
 If you have a question about installation or SymPy in general, feel free to
 visit our chat on `Gitter`_. In addition, our `mailing list`_ is an excellent
--- a/doc/src/index.rst
+++ b/doc/src/index.rst
@@ -5,6 +5,7 @@
    contain the root `toctree` directive.
 
 Welcome to SymPy's documentation!
+
 =================================
 
 A PDF version of these docs can be found `here <https://github.com/sympy/sympy/releases>`_.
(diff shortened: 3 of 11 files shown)


moorepants:
We've typically had developer docs in the sympy wiki. Is it the plan that they stay there?

doc/src/conf.py
```diff
@@ -30,6 +30,14 @@
               'sphinx.ext.mathjax', 'numpydoc', 'sympylive',
               'sphinx.ext.graphviz', 'matplotlib.sphinxext.plot_directive']
 
+redirects = {
+    "install.rst": "getting_started/install.rst",
+    "aboutus.rst": "getting_started/aboutus.rst",
+    "wiki.rst": "getting_started/wiki.rst",
```
moorepants, line 36: The wiki page is quite brief https://docs.sympy.org/dev/wiki.html and if the wiki houses the development docs, then maybe we only need a link pointing developers to the wiki. This page could be deleted.

doc/src/conf.py
```diff
@@ -30,6 +30,14 @@
               'sphinx.ext.mathjax', 'numpydoc', 'sympylive',
               'sphinx.ext.graphviz', 'matplotlib.sphinxext.plot_directive']
 
+redirects = {
+    "install.rst": "getting_started/install.rst",
+    "aboutus.rst": "getting_started/aboutus.rst",
```
moorepants, line 35: This has a list of contributors (not sure if this matches the AUTHORS file), the license, and a list of financial support. I think the list of contributors and financial support better on the sympy website, instead of the docs.

doc/src/conf.py
```diff
@@ -30,6 +30,14 @@
...
+    "wiki.rst": "getting_started/wiki.rst",
+    "citing.rst": "getting_started/citing.rst",
+    "outreach.rst": "getting_started/outreach.rst",
```
moorepants, line 38: This page is also not so useful. I'd delete and move any useful content to the sympy website.

doc/src/getting_started/aboutus.rst
```diff
@@ -4,7 +4,7 @@ About
 SymPy Development Team
 ----------------------
 
-.. literalinclude:: ../../AUTHORS
+.. literalinclude:: ../../../AUTHORS
```
moorepants, line 7: I see, this is the actual AUTHORS file. That's good that it is. But I don't see why it belongs in the documentation.

doc/src/index.rst
```diff
@@ -14,20 +15,73 @@ If you are new to SymPy, start with the :ref:`Tutorial <tutorial>`.
...
+
+..  rst-class:: clearfix row custom-headings
+
+=================================
+
+
+:ref:`Contributing <contributing>`
+-------------------------------------------------------
+
+Covers instructions for SymPy contributors.
```
moorepants, line 67: If we have this section, will you be moving content from the wiki here? As that is where it is housed now.

doc/src/getting_started/aboutus.rst
asmeurer: It used to be a curated list of people and what they contributed, but we moved to just copying AUTHORS because curating a list of contributors doesn't scale well. I think it's fine to have a page for it in the docs. That way people who have contributed can get recognition and have somewhere to point to to show that.

moorepants: As per my other comment, I think this would be more appropriate and visible on the website.

asmeurer: I can see what you mean, but it would be lot harder technically to include it on the website. The AUTHORS file lives in the repo so it's easy to include it in the docs build, which gets updated with every release. For the website, we would probably need to include it as part of the release script. I'm not opposed to that, but I think it might be out of scope for @nanjekyejoannah's project.

moorepants: That's fine. There are just things in the sphinx docs that don't really fit in the new four categories and there are things in wiki that probably belong in these new categories. I assumed getting this sorted out was part of the project. Maybe it isn't.

doc/src/index.rst
nanjekyejoannah: As discussed with @asmeurer and @oscarbenjamin , it's my understanding that we are not moving things from the wiki. In my initial audit, I suggested doing so but I was told it's out of scope for the project.

nanjekyejoannah:
Can I get some consensus on which `rst` files should remain under this `getting started category`. Otherwise, what I gather is that the pages `wiki` and `outreachy` and `aboutus` should be removed from the docs. I plan to update the PR latest tomorrow after getting a response.

asmeurer:
The wiki and outreach pages probably can be removed. If anything the wiki page can be put in the developer section, since most of the wiki is developer documentation. The outreach page is perhaps useful, but I don't know if it belongs in "getting started". I think we should keep aboutus, although I don't know what category it should go in. It also doesn't seem to be related to getting started. Maybe there should be another section for this sort of miscellaneous information about the project.

doc/src/getting_started/aboutus.rst
asmeurer: We decided earlier to leave mostly out of scope moving stuff from the wiki into the docs, because that is a lot more work and just cleaning up what is in the docs already is a bit of work. So we should make the categories based on the idea that things will be moved eventually, but unless there is time at the end, we likely won't be able to do it as part of this project. Moving the authors to the website is probably something we should just open an issue about.

doc/src/index.rst
asmeurer: Yes, we should make room for stuff eventually being moved from the wiki, so that when it happens we have a place for it to go. But moving it is if not out of scope at least the lowest priority thing in this project, compared to cleaning up what is already in the docs.

doc/src/getting_started/aboutus.rst
moorepants: Sounds good!

doc/src/index.rst
moorepants: Maybe at least issues could be opened up to encourage people to build out these new categories, particularly with wiki content. It would make for some nice first contributor issues.

nanjekyejoannah:
I updated this first PR to contain the new proposed documentation layout and getting-started section.

For context, I added a SymPy overview section to enhance the general flow.

asmeurer:
Just to check, is this in a state where we should be able to merge it now (with additional changes coming in future PRs)? I built the docs on this branch and it seems like it to me, in the sense that every page in the docs can still be accessed from the main docs page, or from direct links, but I want to confirm this before I click merge here.

doc/src/index.rst
```diff
@@ -14,20 +15,67 @@ If you are new to SymPy, start with the :ref:`Tutorial <tutorial>`.
...
 This is the central page for all of SymPy's documentation.
 
+==================================================================
+
+..  rst-class:: column
+
+:ref:`Getting started <installation>`
+-------------------------------------------------------
+
+An introduction to SymPy for developers and detailed tutorials.
```
asmeurer, line 25: This is not really accurate. The section as it is now is just installation instructions.

doc/src/getting_started/install.rst
```diff
@@ -1,5 +1,17 @@
 .. _installation:
 
+SymPy Overview
+----------------
+
+SymPy is a computer algebra system (CAS) written in the Python programming language. SymPy is easy to use and install (see the download instructions and tutorial for more information), and works everywhere where Python 2.7 or newer is installed (Linux, Windows, Mac OS X, ...). SymPy's features include:
```
oscarbenjamin, line 6: This test is outdated. It would be better to take the first two paragraphs (about and sympy is...) from here:
https://www.sympy.org/en/index.html

Also it would be better at the top of the main index page rather than the install page.

doc/src/conf.py
```diff
@@ -30,6 +30,10 @@
               'sphinx.ext.mathjax', 'numpydoc', 'sympylive',
               'sphinx.ext.graphviz', 'matplotlib.sphinxext.plot_directive']
 
+redirects = {
```
asmeurer, line 33: I think you need to add the extension to make this actually work.

doc/src/getting_started/install.rst
nanjekyejoannah: Moved all docs to index, trying to avoid a ToC for a section that is still this small, or you wanted the installation ToC @oscarbenjamin ?

oscar-uob: I don't understand what you are asking

oscarbenjamin: I don't understand what you are asking.

nanjekyejoannah: Creating a separate index page then adding the installation page will lead to a small doc for the category plus a Table of contents for the installation section when you navigate to the getting-started section. Was this intended?

nanjekyejoannah: re: Also it would be better at the top of the main index page rather than the install page.

oscarbenjamin: I meant to move this to the actual main index page. The very first page of the whole docs website (`doc/src/index.rst`).

oscarbenjamin:
Looks good, thanks.

PR #21755, opened 2021-07-17
Add N-ary DomainMatrix hstack and vstack

sylee957:

I made `hstack` and `vstack` work with n-ary arguments, and in the requirements, I also had to change `unify` to work with n-ary arguments.
The documentation should be edited, but before that, I want to make sure that this design is agreed on.
I think that those APIs could better be designed as `classmethod`, but it's fine as long as `DomainMatrix.hstack` doesn't have problem even if it's not invoked from the instance

#### Release Notes

- polys
  - Made `DomainMatrix.hstack`, `DomainMatrix.vstack` and `DomainMatrix.unify` work with N-ary arguments.

### DIFF
--- a/sympy/polys/matrices/domainmatrix.py
+++ b/sympy/polys/matrices/domainmatrix.py
@@ -9,6 +9,7 @@ convenience routines for converting between Expr and the poly domains as well
 as unifying matrices with different domains.
 
 """
+from functools import reduce
 from sympy.core.sympify import _sympify
 
 from ..constructor import construct_domain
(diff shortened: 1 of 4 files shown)


sympy/polys/matrices/domainmatrix.py
```diff
@@ -462,36 +463,34 @@ def to_dense(self):
...
-        """Convert self and other to the same format.
-
-        If both are sparse or both are dense then return both unmodified.
+    @classmethod
+    def _unify_domain(cls, *matrices):
+        """Convert matrices to a common domain"""
+        domains = [matrix.domain for matrix in matrices]
+        if len(set(domains)) == 1:
+            return matrices
+        domain = reduce(lambda x, y: x.unify(y), domains)
```
oscarbenjamin, line 472: It would be better to use `set(domains)` here.

oscarbenjamin: I wonder if there are any reasonable uses where an empty list of matrices would be given here and a default domain such as ZZ could be specified.

oscarbenjamin:
If `_unify_fmt` is a classmethod then I think I would have it take a single iterable arg rather than `*args`.

I wonder if hstack and vstack should also be classmethods.

sylee957:
What is the advantage of using single iterable arg than *args?
I think that the advantage of using *args in general was that it was easy to extend the stuff that were applied for single argument without backward incompatibility, if you use single iterable arg, you need to deal with compatibility for iterables with length 1 and

sympy/polys/matrices/domainmatrix.py
sylee957: As long as `Domain.unify` can't handle nullary union, I think that it can be left as exception rather than defining `ZZ`.
I haven't decided that `ZZ` could be the natural answer for that, I rather thought about `EmptyDomain`, but I'm not sure if that's mathematical.
I think that we may also extend `Domain.unify` itself to work with n-ary arguments, but I'm afraid that this can be complicated

sylee957: > It would be better to use `set(domains)` here.

I'm concerned about whether the `Domain.union` is guaranteed to give unique result regardless of the permutation of the arguments. If it is not the case, I'm afraid that using `set` directly can give indeterministic result.

oscarbenjamin: I meant that the default could be give like:
```
K = cls._unify_domain(matrices, ZZ)
```
but maybe it's best just to leave that until there's a use for it.

If `unify` works then I think it should be expected to give the same result regardless of the order of the domains.

oscarbenjamin:
> if you use single iterable arg, you need to deal with compatibility for iterables with length 1

I don't think we need to worry so much about compatibility for DomainMatrix right now. These methods are new (possibly not released yet) and I doubt anyone is using them.

> What is the advantage of using single iterable arg than *args?

I think it's generally better to avoid `*args` altogether unless there is some good reason for it to be used. Using an iterable makes it clearer that any number of inputs can be provided and also avoids awkwards `*` unpacking like:
```
A.rep.hstack(*(Bk.rep for Bk in B))   # *args
A.rep.hstack(Bk.rep for Bk in B)   # iterable
```
Notice that every call to `_unify_fmt` or `_unify_domain` in the diff here uses `*` unpacking: the objects are already in an iterable but have to be converted into a tuple of `*args`.

Also probably not relevant here but `*` unpacking forces the exhaustion of otherwise lazy constructs like generators.

sylee957:
I think that your ideas have some technical difficulties because it needs significant hierarchical changes to `DomainMatrix.unify` to be made `classmethod`, to accept single iterable argument, and so on.
Also, I don't think that it's easy to extend some functions like `DomainMatrix.unify` to work with lazy generators because it needs to inspect all the formats of the matrices at once.

oscarbenjamin:
> I think that your ideas have some technical difficulties

Fair enough. They were just suggestions but we can leave them.

oscarbenjamin:
Looks good to me

Issue #21585, opened 2021-06-07
Matrix exponential using RootSum


oscarbenjamin:
The code below computes a matrix exponential using the inverse Laplace transform and partial fraction expansion. This makes it possible to expression the result using `RootSum` rather than in terms of radical or RootOf expressions for the eigenvalues. The method is based on the fact that the matrix exponential is the inverse Laplace transform of the resolvent:
```python
exp(M*t)  ==  L^-1 (s*I - M)^-1
```
The advantage of using `RootSum` is that we can avoid any calculations with explicit expressions for the roots. That means that it is possible to compute the matrix exponential for matrices of any size even if there are symbolic coefficients (although the result will not be valid for degenerate cases).

In essence the method is just this:
```python
In [112]: M = Matrix([[0, 1], [-1, 0]])

In [113]: M
Out[113]: 
⎡0   1⎤
⎢     ⎥
⎣-1  0⎦

In [114]: s = Symbol('s')

In [115]: inverse_laplace_transform((s*eye(2) - M).inv(), s, t)
Out[115]: 
⎡cos(t)⋅θ(t)   sin(t)⋅θ(t)⎤
⎢                         ⎥
⎣-sin(t)⋅θ(t)  cos(t)⋅θ(t)⎦

In [116]: M.exp()
Out[116]: 
⎡cos(1)   sin(1)⎤
⎢               ⎥
⎣-sin(1)  cos(1)⎦

In [117]: (M*t).exp()
Out[117]: 
⎡   ⅈ⋅t    -ⅈ⋅t         ⅈ⋅t      -ⅈ⋅t⎤
⎢  ℯ      ℯ          ⅈ⋅ℯ      ⅈ⋅ℯ    ⎥
⎢  ──── + ─────    - ────── + ───────⎥
⎢   2       2          2         2   ⎥
⎢                                    ⎥
⎢   ⅈ⋅t      -ⅈ⋅t      ⅈ⋅t    -ⅈ⋅t   ⎥
⎢ⅈ⋅ℯ      ⅈ⋅ℯ         ℯ      ℯ       ⎥
⎢────── - ───────     ──── + ─────   ⎥
⎣  2         2         2       2     ⎦
```
However it is implemented differently because:

1. `inverse_laplace_transform` is slow and does not automatically use `RootSum`.
2. The matrix `inv` method is slow.

It should be possible to make other speed improvements:

[...]

PR #21763, opened 2021-07-20
Add graphical analyses in `physics.control`

akshanshbhatt:

Implement common graphical analyses related to control theory.

#### Other comments

__Checklist:__

- [x] Pole-Zero Plot
- [x] Step Response
- [x] Impulse Response
- [x] Ramp Response
- [x] Bode Plot
- [x] Methods for getting numerical data
- [x] Docs
- [x] Tests

__Future Work:__

- [ ] ~~Root Locus~~ (Effective sampling needed. See https://github.com/sympy/sympy/pull/21763#discussion_r675772024 and https://github.com/sympy/sympy/pull/21763#discussion_r675882746)
- [ ] ~~Nyquist Plot **~~ (Out of scope of my proposed project. Will add sometime in the future along with Nichols plot)

#### Release Notes

* physics.control
  * Added common plots related to control theory.


akshanshbhatt:
### Step Response:

#### For `TransferFunction(4, s**2 + 2*s + 10, s)`

##### MATLAB

image

##### THIS IMPLEMENTATION

output

#### For `TransferFunction(8*s**2 + 18*s + 32, 1*s**3 + 6*s**2 + 14*s + 24, s)`

##### MATLAB

image

##### THIS IMPLEMENTATION

output2

akshanshbhatt:
### Pole-Zero Plot

#### `TransferFunction(2*s**2 + 5*s + 1, s**2 + 3*s + 5, s)`

##### MATLAB
image

##### THIS IMPLEMENTATION
pole_zero_1

#### `TransferFunction(s - 1, s**4 + 6*s**3 + 12*s**2 + 11*s + 6, s)`

##### MATLAB
image

##### THIS IMPLEMENTATION
pole_zero_2

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,48 @@
...
+    zero_points = [zero.as_real_imag() for zero in zeros]
+
+    x_poles = list(map(lambda x: x[0], pole_points))
+    y_poles = list(map(lambda x: x[1], pole_points))
+
+    x_zeros = list(map(lambda x: x[0], zero_points))
+    y_zeros = list(map(lambda x: x[1], zero_points))
+
+    plt.plot(x_poles, y_poles, 'x', mfc='none', markersize=15)
+    plt.plot(x_zeros, y_zeros, 'o', markersize=10)
```
namannimmo10, line 30:
```suggestion
    plt.plot(x_poles, y_poles, 'x', mfc='none', markersize=10)
    plt.plot(x_zeros, y_zeros, 'o', markersize=7)
```
A small markersize will help us point out the zeros more precisely, no?

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,48 @@
...
+
+def step_response(system, **kwargs):
+    r"""
+    Return the unit step response of a continuous-time system.
+    """
+    x = Symbol("x")
+    expr = system.to_expr()/(system.var)
+    y = inverse_laplace_transform(expr, system.var, x)
+    return plot(y, (x, 0, 6), show=True, title="Unit Step Response", \
+        xlabel="Time (Seconds)", ylabel="Amplitude")
```
namannimmo10, line 48:
```suggestion
        xlabel="Time (s)", ylabel="Amplitude")
```

namannimmo10 (commented):
These plots look really nice!

akshanshbhatt:
This PR
image
MATLAB
image

(For `TransferFunction(1*s**2 + 0.1*s + 7.5, 1*s**4 + 0.12*s**3 + 9*s**2, s)`, First example on - https://in.mathworks.com/help/ident/ref/lti.bode.html)

@namannimmo10, the text-overflow/overlap shouldn't be an issue, in my opinion, right?

akshanshbhatt:
Is there some way to fix the axis on the extreme left instead of floating in the middle of the screen (like in MATLAB)?
And also,
How can we turn the grid on while using sympy's native `plot()` (sympy.plotting.plot) method?

akshanshbhatt:
image

After changing the marker size.

namannimmo10:
> Is there some way to fix the axis on the extreme left instead of floating in the middle of the screen (like in MATLAB)?

I'm not aware. If there's a way to obtain a better-looking (with no overlaps or overflows) bode plot using matplotlib or other backends, you can go with that.

akshanshbhatt:
### `TransferFunction(2*s**2 + 5*s + 1, s**2 + 2*s + 3, s)`
image
(MATLAB)
rootloci
(This PR)

### `TransferFunction(s + 7, s*(s + 5)*(s + 15)*(s + 20), s)`
image
(MATLAB)
rootloci_2
(This PR)

akshanshbhatt:
With this, all the plots are complete.
Some minor API changes are required, along with docs and tests, which I will try to finish before next week.

Let me know if you find any issues.

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,160 @@
...
+    y = inverse_laplace_transform(expr, system.var, x)
+    return plot(y, (x, 0, upper_limit), show=show, title="Ramp Response", \
+        xlabel="Time (s)", ylabel="Amplitude")
+
+
+def bode_plot(system, initial_exp=-5, final_exp=5, show=True, **kwargs):
+    r"""
+    Plot a Bode plot of a continuous-time system.
+    """
+    expr = system.to_expr()
```
namannimmo10, line 96: Check the arguments of the plotting functions?

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,160 @@
...
+
+    mag = 20*log(sqrt(real**2 + imag**2), 10)
+
+    mag_plot = plot(mag, (w, 10**initial_exp, 10**final_exp), xscale='log', title="Bode Plot (Magnitude)", \
+        xlabel="Frequency (Hz) [Log Scale]", ylabel="Magnitude (dB)", show=False)
+
+    phase = arg(w_expr)
+
+    phase_plot = plot(phase, (w, 10**initial_exp, 10**final_exp), xscale='log', title="Bode Plot (Phase)", \
+        xlabel="Frequency (Hz) [Log Scale]", ylabel="Phase (rad)", show=False)
```
namannimmo10, line 109: Unfortunately, I don't think this is so easy. Can you give me some links which you followed so that I could check the correctness?

namannimmo10:
@ilayn, @murrayrm: please take a look.

ilayn:
Looks very nice ! Much appreciated

For the step responses, I'd suggest my annoying pair of systems `Transfer(1,[1,1001,1000])` and `Transfer([1, 2, 0],[1,1001,1000])` for testing. 

I'm a bit worried about the inverse laplace way since fast poles might trip up the exponential calculations say pole at -120 is no problem for numeric step responses but exp(-120*t) might underflow too quickly. Emphasis on might :D

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,160 @@
...
+        return
+    return plt
+
+
+def step_response(system, show_input=False, colour='r', show=True, grid=True, upper_limit=6, **kwargs):
+    r"""
+    Return the unit step response of a continuous-time system.
+    """
+    x = Symbol("x")
+    expr = system.to_expr()/(system.var)
```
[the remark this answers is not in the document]
akshanshbhatt: As I said earlier --

> Some minor API changes are required, along with docs and tests, which I will try to finish before next week.

This was one of them. All of these plots are meant for SISO lti systems with no time delay term.

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,160 @@
...
+    np = numpy  # Matplotlib already has numpy as a compulsory dependency. No need to install it separately.
+
+
+def pole_zero(system, pole_colour='r', zero_colour='b', grid=True, show=True, **kwargs):
+    r"""
+    Returns the Pole-Zero plot (also known as PZ Plot or PZ Map) of a system.
+    """
+    system = system.doit()  # Get the equivalent TransferFunction object.
+
+    num_poly = Poly(system.num, system.var).all_coeffs()
```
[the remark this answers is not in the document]
akshanshbhatt: The reason behind using `Poly()` was speed. I was using NumPy for root-locus and getting a huge performance advantage because- (a) SymPy is a symbolic computation library and NumPy is a numeric computation library. (b) Matplotlib requires NumPy for internal numeric computation and hence NumPy is also a listed dependency for matplotlib. It is installed **with** matplotlib.

image

Also just to be clear, these plots will not support transfer functions with time delay. Maybe Ill allow Pole-Zero because time delay doesn't affect the poles and zeroes of the system.

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,160 @@
...
+    num_poly = np.array(num_poly, dtype=np.float64)
+    den_poly = np.array(den_poly, dtype=np.float64)
+
+    zeros, poles = _find_roots(num_poly, den_poly)
+
+    zero_real = np.real(zeros)
+    zero_imag = np.imag(zeros)
+
+    pole_real = np.real(poles)
+    pole_imag = np.imag(poles)
```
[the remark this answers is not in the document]
akshanshbhatt: SymPy can do it but by taking 1000x more time. SymPy is not optimized for large numerical computations, NumPy however, is a numerical computation library. Let me explain.

This root computation operation in SymPy took about 50ms per operation.
```python 
from sympy import Poly, roots, nroots
from sympy.abc import x

root_list = []

for k in range(0, 100):
    s = Poly(k*x**8 + 9*x**7 + 7*x**6 + 32*x**5 + 89*x**2 + 78*x + 10, x)
    %time root_list.append(nroots(s, n=8))

print(root_list)
```

OUTPUT:
```
Wall time: 59 ms
Wall time: 58.3 ms
Wall time: 86.6 ms
Wall time: 62 ms
Wall time: 82 ms
Wall time: 53 ms
Wall time: 67.8 ms
Wall time: 51 ms
Wall time: 54 ms
Wall time: 53.1 ms
Wall time: 47 ms
Wall time: 54 ms
Wall time: 59 ms
Wall time: 57 ms
Wall time: 71 ms
Wall time: 50 ms
Wall time: 50 ms
Wall time: 51 ms
Wall time: 65.2 ms
Wall time: 45.8 ms
Wall time: 46.3 ms
Wall time: 67.7 ms
Wall time: 44.9 ms
Wall time: 47.9 ms
Wall time: 61.8 ms
Wall time: 55.9 ms
Wall time: 44.4 ms
Wall time: 48.9 ms
Wall time: 42.7 ms
Wall time: 42.1 ms
Wall time: 48 ms
Wall time: 55.9 ms
Wall time: 70.4 ms
Wall time: 111 ms
Wall time: 68.2 ms
Wall time: 41.7 ms
Wall time: 44 ms
Wall time: 47 ms
Wall time: 39.3 ms
Wall time: 45 ms
Wall time: 23 ms
Wall time: 21 ms
Wall time: 23 ms
Wall time: 22.6 ms
Wall time: 24 ms
Wall time: 29 ms
Wall time: 22 ms
Wall time: 20 ms
Wall time: 19.5 ms
Wall time: 18.7 ms
Wall time: 24.4 ms
Wall time: 30.4 ms
Wall time: 35 ms
Wall time: 24 ms
Wall time: 20 ms
Wall time: 25 ms
Wall time: 27 ms
Wall time: 21 ms
Wall time: 23 ms
Wall time: 17 ms
Wall time: 18 ms
Wall time: 20 ms
Wall time: 22.1 ms
Wall time: 21 ms
Wall time: 22.8 ms
Wall time: 15.6 ms
Wall time: 16.6 ms
Wall time: 15.6 ms
Wall time: 18.7 ms
Wall time: 20 ms
Wall time: 28 ms
Wall time: 16 ms
Wall time: 16 ms
Wall time: 16 ms
Wall time: 19 ms
Wall time: 17.9 ms
Wall time: 21 ms
Wall time: 24 ms
Wall time: 16.6 ms
Wall time: 16.7 ms
Wall time: 18.7 ms
Wall time: 17.8 ms
Wall time: 19.5 ms
Wall time: 22.1 ms
Wall time: 24.9 ms
Wall time: 35 ms
Wall time: 29 ms
Wall time: 35 ms
Wall time: 19 ms
Wall time: 23 ms
Wall time: 16 ms
Wall time: 4.44 ms
Wall time: 15.6 ms
Wall time: 36.5 ms
Wall time: 11 ms
Wall time: 28.3 ms
Wall time: 33.1 ms
Wall time: 31 ms
Wall time: 31 ms
Wall time: 24 ms
```

The same operation on NumPy took:
```python 
import numpy as np

root_list = []

for k in range(0, 100):
    s = np.array([k, 9, 7, 32, 0, 0, 89, 78, 10])
    %time root_list.append(np.roots(s))

print(root_list)
```
OUTPUT:

```
Wall time: 29.7 ms
Wall time: 0 ns
Wall time: 1e+03 µs
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 1 ms
Wall time: 0 ns
Wall time: 0 ns
Wall time: 996 µs
Wall time: 0 ns
Wall time: 0 ns
Wall time: 1 ms
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 1.01 ms
Wall time: 0 ns
Wall time: 998 µs
Wall time: 0 ns
Wall time: 1.02 ms
Wall time: 0 ns
Wall time: 1.01 ms
Wall time: 0 ns
Wall time: 0 ns
Wall time: 1.01 ms
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 1e+03 µs
Wall time: 0 ns
Wall time: 0 ns
Wall time: 999 µs
Wall time: 0 ns
Wall time: 0 ns
Wall time: 1 ms
Wall time: 0 ns
Wall time: 998 µs
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 997 µs
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
Wall time: 0 ns
```

And yes, the results are not cached. you can try this on your local.

This speed in computation is required especially during plotting. This is not something out of the blue. NumPy is used internally in the SymPy plotting module too. Using NumPy is not a choice rather a necessity.

https://github.com/sympy/sympy/blob/645bbb06cd289165a9144d4c2c64bf17b664f8ac/sympy/plotting/plot.py#L1322-L1335

sympy/physics/control/control_plots.py
akshanshbhatt: As told earlier, the user API is crude now. I have not dealt with input checking yet. Even the keyword arguments passed are unused in most of the function signatures.

sympy/physics/control/control_plots.py
akshanshbhatt: > Can you give me some links which you followed so that I could check the correctness?

https://en.wikipedia.org/wiki/Bode_plot
https://www.youtube.com/watch?v=_eh1conN6YM
https://www.youtube.com/watch?v=CSAp9ooQRT0

I didn't look at someone's implementation. I just read the Wikipedia page saw a couple of related videos/univ lectures. I tested it with 10s of different examples which I could find online and all the plots matched exactly. There might be few vulnerabilities, if you find any particular case just let me know.

> Unfortunately, I don't think this is so easy.

The point here is not about being easy. It is the **definition** of the Bode plot. This is how it is defined. I really don't get that how can it be wrong if the calculation of points is done by the sole definition of the plot.

sympy/physics/control/control_plots.py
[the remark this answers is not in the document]
akshanshbhatt: I forgot to add that the time results that I shared were only for 100 different `k` values.
As you can see here, for root locus, we are computing the roots 10000 times (`num=10000`) so there is absolutely no point in using sympy.

https://github.com/sympy/sympy/blob/ec68ebf5b089f9166b759c4fe7fd5ef42fcf2e7b/sympy/physics/control/control_plots.py#L133-L140

https://github.com/sympy/sympy/blob/ec68ebf5b089f9166b759c4fe7fd5ef42fcf2e7b/sympy/physics/control/control_plots.py#L114-L121

sympy/physics/control/control_plots.py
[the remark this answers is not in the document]
oscarbenjamin: It is possible to use sympy with other plotting backends that do not necessarily use numpy:
https://pypi.org/project/sympy-plot-backends/

Generally when sympy has numeric routines they are more accurate than numpy's:
```python
In [8]: p = expand(prod(x - i for i in range(1, 21)))

In [9]: nroots(p)
Out[9]: 
[1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0, 12.0, 13.0, 14.0, 15
.0, 16.0, 17.0, 18.0, 19.0, 20.0]

In [10]: numpy.roots(Poly(p).all_coeffs())
Out[10]: 
array([19.99980929, 19.00190982, 17.99092135, 17.02542715, 15.94628672,
       15.0754938 , 13.91475559, 13.07431403, 11.95328325, 11.02502293,
        9.99041304,  9.00291529,  7.99935583,  7.000102  ,  5.99998925,
        5.00000067,  3.99999998,  3.        ,  2.        ,  1.        ])
```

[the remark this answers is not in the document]
akshanshbhatt: Plotting backend is not an issue here. Whichever backend we may use, it requires efficient sampling to generate plots. Some form of numerical computation will always occur under the hood. SymPy undoubtedly is more accurate in numerical operations. Still, during sampling with a large number of points (10000), accuracy hardly matters, especially when a much faster approach (more than 1000 times) is accurate up to 3-4 decimal points.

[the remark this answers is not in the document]
oscarbenjamin: > is accurate up to 3-4 decimal places.

The accuracy can degrade much more than that.

The idea with using a symbolic library is that you should have greater confidence in things like the accuracy of any computed result even if that means waiting longer for things to be computed. It also means that different approaches are often better to make something efficient.

As an example your `root_locus` function computes the roots of 10000 different polynomials. Why 10000? No one needs that many points in the plot. Many of the lines are just joining points on the real axis. To plot a straight line you only need to know its start and end and those can be computed exactly. Then you don't need to evaluate the polynomial at 10000 points.

I'm not sure I've understood the calculation that you want precisely but I think it's something like this:
```python
In [35]: eq
Out[35]: k⋅(s + 7) + s⋅(s + 5)⋅(s + 15)⋅(s + 20)

In [36]: discriminant(eq)
Out[36]: 
      4          3             2                               
- 27⋅k  + 47088⋅k  - 37907500⋅k  + 3881250000⋅k + 1265625000000

In [37]: nroots(discriminant(eq))
Out[37]: [-130.564918275438, 307.765516989741, 783.399700642849 - 743.513266939402⋅ⅈ, 783.399700642849 + 743.513266939402⋅ⅈ]
```
The positive real roots for `k` are the points on the real axis where lines start and end. This now shows you what values of `k` are interesting for the plot.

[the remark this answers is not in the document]
akshanshbhatt: For `TransferFunction(2*s**2 + 5*s + 1, s**2 + 3*s + 5, s)`

* If `num=10000` (sampled 10000 times, default)

  scatter_rootLocus_10k

  The scatter plot looks like this.
  And the actual root-locus plot (corresponding to the regression of these sampled points) looks like this-

  rl

Similarly,

* If `num=1000`
  
  1k

  1k_plot

* If `num=300`

  scatter_300

  plot_300


> To plot a straight line you only need to know its start and end and those can be computed exactly. Then you don't need to evaluate the polynomial at 10000 points.

Straight Lines in root locus plots are not very common. In root-locus plots, we plot the movement of the roots of the polynomial `den+ k*num` when k varies from 0 to inf. As you can tell when k is 0 we get the roots of den (poles) and as k approaches inf, the roots of `den + k*num` are roots of num only (zeros). Therefore, when num and den have the same degree, the root locus plot is a path from poles to zeros of the transfer function. This is not always a straight line. 

equal

Only one is a straight line path here and knowing this fact can't help us in any way. I don't understand how knowing the discriminant of the polynomial will help us in knowing the points on the rl plot. Maybe I'm missing something.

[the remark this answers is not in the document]
oscarbenjamin: > Straight Lines in root locus plots are not very common.

Your example with `num=1000` illustrates the point I wanted to make perfectly. To generate the plot you compute the roots of the polynomial for 1000 different values of k ranging from 0 to kmax where kmax is arbitrarily chosen to be 400. Out of those 1000 values of k only 4 have non-real roots so the other 996 polynomials whose roots are computed are only used to plot two straight lines. The straight line could be computed more efficiently and more accurately by simply computing its start and end points which you can do like this:
```python
In [93]: k, s = symbols('k, s')

In [94]: n, d = 2*s**2 + 5*s + 1, s**2 + 3*s + 5

In [95]: eq = d + k*n

In [96]: eq
Out[96]: 
  ⎛   2          ⎞    2          
k⋅⎝2⋅s  + 5⋅s + 1⎠ + s  + 3⋅s + 5

In [97]: nroots(n) # k = oo
Out[97]: [-2.28077640640441, -0.219223593595585]

In [98]: discriminant(eq, s)
Out[98]: 
    2            
17⋅k  - 14⋅k - 11

In [99]: nroots(discriminant(eq, s))
Out[99]: [-0.491899499749248, 1.31542891151395]

In [100]: roots(discriminant(eq, s), multiple=True)
Out[100]: 
⎡7    2⋅√59  7    2⋅√59⎤
⎢── - ─────, ── + ─────⎥
⎣17     17   17     17 ⎦

In [101]: eq.subs(k, roots(discriminant(eq, s), multiple=True)[1])
Out[101]: 
 2         ⎛7    2⋅√59⎞ ⎛   2          ⎞    
s  + 3⋅s + ⎜── + ─────⎟⋅⎝2⋅s  + 5⋅s + 1⎠ + 5
           ⎝17     17 ⎠                     

In [102]: roots(eq.subs(k, roots(discriminant(eq, s), multiple=True)[1]), s)
Out[102]: {-9 + √59: 2}

In [103]: list(_)[0].n()
Out[103]: -1.31885425213139
```
Your suggestion to use num=10000 and compute the roots of 10000 polynomials is a workaround for having the wrong value of kmax. In fact by computing the roots of the discriminant it is possible to compute what the relevant values of k are.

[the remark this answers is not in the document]
akshanshbhatt: Fine. I get your point now. But there is still an issue. You see, with your calculations, we get these points:

image

Which is basically poles and zeroes of the Transfer Function + The point where the discriminant is zero (equal roots). I agree that for the two straight lines on the axis, this will work fine but what about the two symmetric arms of the curve?
We cannot plot them by simply knowing the start and endpoints. We still need to compute the roots of the polynomials from 0 to `k_c` (value of k for which the 0 discriminant point is reached) in uniform intervals.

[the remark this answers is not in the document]
oscarbenjamin: > We still need to compute the roots of the polynomials from 0 to `k_c`

Yes, but we don't need 10000 points to plot the curve. The `num=10000` plot above shows that an acceptable plot is obtained from only 40 points (since the other 9960 points are on the real axis).

I expect that uniform spacing is not optimal. Naturally I would guess that something like logarithmic or harmonic spacing would be better. You can see in the `num=10000` plot that using uniform intervals for `k` does not lead to a uniform distance between the points in the plot.

Ideally the curve would be generated using pseudo-arc-length continuation so you can generate them at a fixed 2D distance within the plot but that's probably more complicated than it is worth.

namannimmo10:
You can set the [`linestyle`](https://matplotlib.org/stable/api/_as_gen/matplotlib.lines.Line2D.html#matplotlib.lines.Line2D.set_linestyle) as `'dotted'` (or something else) so that these RGB colors are differentiable on the axis;
tf

sympy/physics/control/control_plots.py
namannimmo10: > I really don't get that how can it be wrong if the calculation of points is done by the sole definition of the plot.

I did not say that you are wrong. I just had some doubts and wanted to verify the correctness myself so I asked you to provide me some links.

> I tested it with 10s of different examples which I could find online and all the plots matched exactly. 

Thanks!

sympy/physics/control/control_plots.py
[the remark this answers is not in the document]
namannimmo10: You could have just said, "I'm using NumPy for a better computation speed." :) 
Thanks, I'm satisfied now.

plaes:
Although this is a cool and much wanted feature, I'm not that sure whether it makes sense to put these plotting functions in as public API. 

Although it would be nice to have out-of-the box plotting tools, we end up with a lot of customization and localization issues which eventually forces users to create a copy of existing functionality anyway if they want to do even the simplest things like changing the plot titles and labels.

oscarbenjamin:
> if they want to do even the simplest things like changing the plot titles and labels.

I think that the solution to that is that the plotting APIs should expose the lower level a bit so it's easy to e.g. get the matplotlib handle for the figure and add a title. At the same time any plotting APIs should make it easier to just get the underlying data so that they can be used with different plotting libraries.

CC @Davide-sd

moorepants:
At our meeting with @akshanshbhatt yesterday we suggested a few things:

1. The root locus plots really need adaptive sampling to look good. There are tools like https://github.com/python-adaptive/adaptive that do a nice job at this and sympy.plot() also does this in some ways.
2. Move the root locus plot to another PR, so this can be merged as is. The other plots look good.
3. Open an issue about the plot labels over lapping for sympy.plot() in the one example. It doesn't need to be solved here, as that is the current default (and seemingly not able to be adjusted).
4. Add tests, documentation, and examples in the docstrings, etc.

I'm +1 to merge this if root locus is moved to another PR and the documetation/tests are added.

moorepants:
@plaes and @oscarbenjamin I think the issues you raise are good points, but should be opened in a general issue about sympy.plot() and addressed in that module. For better or worse, we have sympy.plot() and it is designed in the way it is. We definitely want plotting commands that offer quick ways to examine these transfer functions. Calling something like `TransferFunction.plot_response('step')` and getting an immediate result is super useful. If you need custom plots for publications or something then users will always be better served rolling their own sympy -> lambdify -> matplotlib (or other plotting library) script. Providing simple, good in most cases, plotting functionality is very useful in a symbolic library and many other popular libraries provide the same. It saves the user enormous time in interactive computing.

oscarbenjamin:
> @plaes and @oscarbenjamin I think the issues you raise are good points, but should be opened in a general issue about sympy.plot() and addressed in that module.

The part that is relevant to this PR is that when we have a function like `impulse_reponse` that computes and plots the impulse response then there should be an easy way for the user to get the expression that is being plotted rather than showing the plot. E.g. something like:
```python
def impulse_reponse(tf):
    return inverse_laplace_transform(...)

def impulse_reponse_plot(tf, ...):
    response = impulse_response(tf)
    return plot(response, ...)
```
This way if the user wants to use another plotting library or control other details of the plot then they can but they can still use the `impulse_response` function to do the computation. Of course if the API was designed that way then you could question the benefit of having the `impulse_response_plot` function at all e.g.:
```python
# Which of these is better:
impulse_response_plot(tf)
plot(impulse_response(tf))
```
I guess it depends how nontrivial the code in the `_plot` function is and whether the specific choices it makes are generally more useful than the default/customisable behaviour of the ordinary `plot` function. Probably just a clear example in the docs showing how to use the ordinary `plot` function with options that are likely to be useful when plotting an impulse response is more useful to users than having an actual function that both computes and plots the impulse response.

Personally as a user I would rather have a smaller set of reusable API to learn. If `impulse_response` is doing something clever to choose the plot limits or something then that's useful but otherwise having to read the docs to learn that I need to provide `upper_limit=100` is just awkward if I already know how to set the horizontal limits using the *ordinary* plot function.

A function that computes something and then has no option to access the computed result except as an output side effect (e.g. a plot) is usually poor API. Similarly, I regularly teach novice programmers not to make functions that compute something and print out the result. If the result is *returned* rather than printed then the caller of the function can choose whether to print it or not and then the function is reusable, composable etc. In other words as much code/API as possible should be in *pure* functions that can be composed with input/output functions.

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def impulse_response(system, show_input=False, colour='r', show=True, grid=True, upper_limit=6, **kwargs):
+    r"""
+    Return the unit impulse response (Input is the Dirac-Delta Function) of a
+    continuous-time system.
+    """
+    x = Symbol("x")
+    expr = system.to_expr()
+    y = inverse_laplace_transform(expr, system.var, x)
```
oscarbenjamin, line 75: What if `expr` already contains a symbol called `"x"`?

akshanshbhatt: I'll replace it with a dummy variable.

moorepants:
I agree that having `plot_impulse_response()` and `calculate_impulse_response()` is good to have. Users may want each (at least when it is possible to provide an analytic result).

 For SISO systems, as soon as the order of the transfer function is > 4 it is impossible to provide an analytic solution. So, there is limited use for computing the solution to these linear differential equations.

Lastly, if this PR introduces only the plot() functions, there is nothing preventing the functions that output the expressions that are plotted from being added later. You don't have to solve this expression output problem now. I think we can add these plotting functions as is without hindrance to future features.

sympy/physics/control/control_plots.py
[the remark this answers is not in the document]
moorepants: Do these even work with high order systems?

moorepants:
Test driven and/or documentation driven development is not being used here, to your detriment. You are designing these functions without first creating a fully encompassing set of possible inputs. This absolutely needs to be done because I believe your functions will only work with some very simple transfer functions as designed.

For example, what happens if I want to plot the impulse response function of: `tf = S(1)/(s**5 + s**4 + s**3 + s**2 + s + 1)` or `tf = (a + b*s)/(s**2 + d*s + 3)`?

moorepants:
Here's another example, with a different result: `tf = sm.S(1)/(s**8 + s**4 + s**3 + s**2 + s + 1)`.

akshanshbhatt:
> For example, what happens if I want to plot the impulse response function of: `tf = S(1)/(s**5 + s**4 + s**3 + s**2 + s + 1)` 

image

As compared to MATLAB,

image

> or `tf = (a + b*s)/(s**2 + d*s + 3)`?

This will prompt the user with an error as without knowing the numerical values of these constants, there is no way we can plot the impulse response.

akshanshbhatt:
It is possible to find the numerical values (approx.) of unevaluated `InverseLaplaceTransform` objects.

moorepants:
And how long did it take to make the first plot?

sympy/physics/control/control_plots.py
moorepants: > There might be few vulnerabilities, if you find any particular case just let me know.

@akshanshbhatt it is your job to figure these out and show us that you've fully though this through. Right now, it doesn't look like you have.

oscarbenjamin:
> Test driven and/or documentation driven development is not being used here

Another advantage of pure functions is that they are easier to test.

oscarbenjamin:
> For SISO systems, as soon as the order of the transfer function is > 4 it is impossible to provide an analytic solution.

This is not entirely correct. If the transfer function has symbolic coefficients then it will usually not be possible for `order > 4` however with rational coefficients we can go to any order using `RootOf` or `RootSum`. See #21585 for related examples.

moorepants:
For transfer functions with numeric coefficients, making a plot of something like an impulse response is going to be easier and much faster if done with numeric floating point calculations. If all I want is the plot, computing the `inverse_laplace_transform()` with sympy is not the most direct method and may fail on some transfer functions. I imagine `plot_impulse_response()` not even calling `calculate_symblic_impluse_response()` internally.

akshanshbhatt:
> And how long did it take to make the first plot?

```
In [9]: %time impulse_response(TransferFunction(1, s**5 + s**4 + s**3 + s**2 + s + 1, s), upper_limit=60)
Out[9]: 
Wall time: 12.5 s
<sympy.plotting.plot.Plot at 0x220ff5b3e50>
```

> For transfer functions with numeric coefficients, making a plot of something like an impulse response is going to be easier and much faster if done with numeric floating point calculations. If all I want is the plot, computing the `inverse_laplace_transform()` with sympy is not the most direct method and may fail on some transfer functions. I imagine `plot_impulse_response()` not even calling `calculate_symblic_impluse_response()` internally.

That was exactly how I planned it initially -- Using `mpmath` to calculate the numerical Inverse Laplace Transform. https://mpmath.org/doc/current/calculus/inverselaplace.html. I went ahead with the symbolic approach because I wanted to utilize sympy's native plotting module wherever possible. Although, now it makes sense to use the numerical method instead.

oscarbenjamin:
> I imagine `plot_impulse_response()` not even calling `calculate_symblic_impluse_response()` internally.

Maybe not but then the same principle applies: there should be a function like `calculate_numeric_impulse_reponse()` so that the user can get the numbers that are being plotted rather than just getting the plot.

> If all I want is the plot, computing the `inverse_laplace_transform()` with sympy is not the most direct method and may fail on some transfer functions.

This might not be what you want to use here but perhaps for some other calculations in the control module: in #21585 I have shown a method that can compute the inverse laplace transform for polynomials of any order including with symbolic coefficients. It is not restricted to degree less than 5 because it does not use radical expressions although it can be converted to radical expressions when possible. With symbolic coefficients the results are valid for the generic case but break down for degenerate values of the symbols (e.g. giving multiple roots). The method is also faster than the ordinary `inverse_laplace_transform` function:
```python
In [2]: s, t, beta = symbols('s, t, beta')

In [3]: tf = 1/(s**5 - beta*s + 1)

In [4]: tf
Out[4]: 
      1      
─────────────
        5    
-β⋅s + s  + 1

In [5]: %time tf_apart = apart(tf, s, full=True)  # partial fraction expansion using rootsum
CPU times: user 72.7 ms, sys: 4.34 ms, total: 77.1 ms
Wall time: 75.1 ms

In [6]: %time ilt(tf_apart, s, t)   # inverse laplace transform
CPU times: user 14.6 ms, sys: 342 µs, total: 15 ms
Wall time: 15 ms
Out[6]: 
       ⎛ 5                ⎛     4  3        3  2        2                  4⎞  a⋅t⎞
RootSum⎝w  - w⋅β + 1, a ↦ ⎝320⋅a ⋅β  + 400⋅a ⋅β  + 500⋅a ⋅β + 625⋅a - 256⋅β ⎠⋅ℯ   ⎠
───────────────────────────────────────────────────────────────────────────────────
                                        5                                          
                                   256⋅β  - 3125
```
This expression can be manipulated symbolically and is valid for any value of `beta` except those that give multiple roots making the denominator zero:
```python
In [7]: discriminant(1/tf, s)
Out[7]: 
            5
3125 - 256⋅β 
```

namannimmo10:
@akshanshbhatt: any updates?

akshanshbhatt:
> @akshanshbhatt: any updates?

@namannimmo10 this PR is nearly complete. I have to push the updated unit tests to the remote (which are complete btw). The only thing remaining is examples in `MIMOFeedback` docstring. You can review the rest.

akshanshbhatt:
This PR is complete from my end.

I have tried to address every comment in this thread:

>Naming should be consistent for all the plots. Although, I wonder if we really want this naming style. `pole_zero`, by itself, doesn't make much sense.

- Every plotting function in the module has consistent naming now. `(plot_name)_plot(sys)` for plots and `(plot_name)_numerical_data()` for getting the plotted data.

---
> For the step responses, I'd suggest my annoying pair of systems `Transfer(1,[1,1001,1000])` and `Transfer([1, 2, 0],[1,1001,1000])` for testing.
> 
> I'm a bit worried about the inverse laplace way since fast poles might trip up the exponential calculations say pole at -120 is no problem for numeric step responses but exp(-120*t) might underflow too quickly. 

- Both these tf systems can be plotted without any issue with the module. Here is the plot comparison with MATLAB generated ones for both (switch to light mode for a better view) -

image

image

---
>  we end up with a lot of customization and localization issues which eventually forces users to create a copy of existing functionality anyway if they want to do even the simplest things like changing the plot titles and labels.

> I think that the solution to that is that the plotting APIs should expose the lower level a bit so it's easy to e.g. get the matplotlib handle for the figure and add a title.

- Users can get the matplotlib handle for the figure by inputting `False` to the `show` parameter. Users can change the titles, add legends, combine multiple plots, etc.

---
> The part that is relevant to this PR is that when we have a function like `impulse_reponse` that computes and plots the impulse response then there should be an easy way for the user to get the expression that is being plotted rather than showing the plot. E.g. something like:
> 
> ```python
> def impulse_reponse(tf):
>     return inverse_laplace_transform(...)
> 
> def impulse_reponse_plot(tf, ...):
>     response = impulse_response(tf)
>     return plot(response, ...)
> ```
> 
> This way if the user wants to use another plotting library or control other details of the plot then they can but they can still use the `impulse_response` function to do the computation.

> I agree that having plot_impulse_response() and calculate_impulse_response() is good to have. Users may want each (at least when it is possible to provide an analytic result).

> At the same time any plotting APIs should make it easier to just get the underlying data so that they can be used with different plotting libraries.

- I have added functions for getting the numerical data of each of the plots. If the user wants to use any other module for plotting or analyzing the data, he can simply use the `(plot_name)_numerical_data(sys)` method. This seemed better than the alternative (`plot(impulse_response)` one) as this gives more customizability. All the data is computed natively in SymPy itself. Users can choose if they want to use adaptive sampling or not. 

---
> For example, what happens if I want to plot the impulse response function of: tf = S(1)/(s**5 + s**4 + s**3 + s**2 + s + 1) or tf = (a + b*s)/(s**2 + d*s + 3)

> Here's another example, with a different result: tf = sm.S(1)/(s**8 + s**4 + s**3 + s**2 + s + 1).

- I have handled every case now. Users will be prompted with error messages if coefficients are not numbers. Time delay terms, MIMO systems, etc. - I have handled every case and made sure only suitable systems are allowed.

- Secondly, higher-order systems can be plotted highly efficiently now. Thanks to the inverse Laplace transform function suggested by @oscarbenjamin. It is even faster as well as more accurate than most numerical methods. Last time the function suggested by you took around 12 seconds to compute the plot. Using Oscar's method, I was able to get the plot of the same 8 order system in about a second and it was highly accurate too. I would suggest you try it yourself.

output

This system has order 3 polynomial in the num and 6 order polynomial in den. Yet, this plot only took 996 ms. It was even faster than MATLAB's `impulse()` command. MATLAB's output for the same system -

image
---
> Another advantage of pure functions is that they are easier to test.

- All the unit tests in the module have utilized this.

> For transfer functions with numeric coefficients, making a plot of something like an impulse response is going to be easier and much faster if done with numeric floating point calculations. If all I want is the plot, computing the inverse_laplace_transform() with sympy is not the most direct method and may fail on some transfer functions. I imagine plot_impulse_response() not even calling calculate_symblic_impluse_response() internally.

- @oscarbenjamin's method for calculating the inv Laplace transform was far more efficient than any other numerical method out there. I tried mpmath's method for numerical calculation and it was 6x slower than Oscar's method, even using their fastest method (Talbot).

---
Also,
- We don't require other external dependencies for adaptive sampling as sympy handles it itself now.
- All plots now have a coherent structure. Earlier, Bode and response plots were plotted using SymPy's plotting module, and PZ and RL plots were plotted using matplotlib, but now every plot is plotted using Matplotlib.

It would really be helpful if this PR is reviewed and merged before GSoC ends.

sympy/physics/control/tests/test_control_plots.py
```diff
@@ -0,0 +1,258 @@
...
+
+def test_impulse_response():
+    if not matplotlib:
+        skip("Matplotlib not the default backend")
+
+    assert _to_tuple(*impulse_response_numerical_data(tf1, adaptive=False, nb_of_points=10)) == \
+        ((0.0, 1.1111111111111112, 2.2222222222222223, 3.3333333333333335, 4.444444444444445,
+        5.555555555555555, 6.666666666666667, 7.777777777777779, 8.88888888888889, 10.0),
+        (0.0, 0.544019738507865, 0.01993849743234938, -0.31140243360893216, -0.022852779906491996, 0.1778306498155759,
+        0.01962941084328499, -0.1013115194573652, -0.014975541213105696, 0.0575789724730714))
```
oscarbenjamin, line 151: This test seems fragile by testing for exact equality with floats.

akshanshbhatt: Is there any better alternative?

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,964 @@
...
+
+    if show:
+        plt.show()
+        return
+
+    return plt
+
+
+def _fast_inverse_laplace(e, s, t):
+    """Fast inverse Laplace transform of rational function including RootSum"""
```
oscarbenjamin, line 922: I think this should be added in the integrals module somewhere so that someone who looks at the Laplace transform code will see it there.

sympy/physics/control/tests/test_control_plots.py
oscarbenjamin: It's better to test that the numbers are close to some precision e.g. `abs(result - expected) < 1e-15` or something like that.

sympy/physics/control/control_plots.py
oscarbenjamin: It should probably be in the same place as the `inverse_laplace_transform` function

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,964 @@
...
+    sys = system.to_expr()
+    len_free_symbols = len(sys.free_symbols)
+    if len_free_symbols > 1:
+        raise ValueError("Extra degree of freedom found. Make sure"
+            " that there are no free symbols in the dynamical system other"
+            " than the variable of Laplace transform.")
+    if sys.has(exp):
+        raise NotImplementedError("Time delay terms are not supported.")
+
+    return
```
namannimmo10, line 45:
```suggestion
```

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,964 @@
...
+    Parameters
+    ==========
+
+    system : SISOLinearTimeInvariant
+        The system for which the pole-zero data is to be computed.
+
+    Returns
+    =======
+
+    tuple : (zeros, poles)
```
namannimmo10, line 65:
```suggestion
    tuple : (poles, zeros)
```

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,964 @@
...
+
+
+def pole_zero_plot(system, pole_color='blue', pole_markersize=10,
+    zero_color='orange', zero_markersize=7, grid=True, show_axes=True,
+    show=True, **kwargs):
+    r"""
+    Returns the Pole-Zero plot (also known as PZ Plot or PZ Map) of a system.
+    A Pole-Zero plot is a graphical representation of a system's poles and
+    zeros. It is plotted on a complex plane, with circular markers representing
+    the system's zeros and 'x' shaped markers representing the system's poles.
```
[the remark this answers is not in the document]
namannimmo10: See [PEP 257](https://www.python.org/dev/peps/pep-0257/);

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,964 @@
...
+    _x = Dummy("x")
+    expr = system.to_expr()/(system.var)
+    expr = apart(expr, system.var, full=True)
+    _y = _fast_inverse_laplace(expr, system.var, _x).evalf(prec)
+    return LineOver1DRangeSeries(_y, (_x, lower_limit, upper_limit),
+        **kwargs).get_points()
+
+
+def step_response_plot(system, color='b', show=True, lower_limit=0,
+    upper_limit=10, prec=8, show_axes=False, grid=True, **kwargs):
```
namannimmo10, line 282:
```suggestion
def step_response_plot(system, color='b', prec=8, lower_limit=0,
    upper_limit=10, show_axes=False, grid=True, show=True, **kwargs):
```
I'd try to keep the order consistent.

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,964 @@
...
+        **kwargs).get_points()
+
+
+def ramp_response_plot(system, slope=1, color='b', show=True, lower_limit=0,
+    upper_limit=10, prec=8, show_axes=False, grid=True, **kwargs):
+    r"""
+    Returns the ramp response of a continuous-time system. Ramp
+    function is defined as the the straight line passing through
+    origin ($f(x) = mx$). The slope of the ramp function can be
+    varied by the user and the default value is 1.
```
namannimmo10, line 596:
```suggestion
    Returns the ramp response of a continuous-time system.
    
    Ramp function is defined as the straight line passing through
    origin ($f(x) = mx$). The slope of the ramp function can be
    varied by the user and the default value is 1.
```

namannimmo10 (commented):
Great! Can you finish this off by showing in the PR how all the plots look like?

sympy/physics/control/control_plots.py
akshanshbhatt: I guess the order is fine. Zeros should be first because arg of tf is `(num, den)` - so it makes sense to return `(root(num), roots(den))`. This might confuse the user.

sympy/physics/control/tests/test_control_plots.py
akshanshbhatt: Done

akshanshbhatt:
> Can you finish this off by showing in the PR how all the plots look like?

Sure!

These are all the plots from the documentation page rendered on my local -

image
image
image
image
image

 All these plots will be visible on the doc page after merging this PR.

sympy/integrals/transforms.py
```diff
@@ -1412,14 +1414,60 @@ def inverse_laplace_transform(F, s, t, plane=None, **hints):
...
+        if expr.is_Mul:
+            raise NotImplementedError
+        return coeff * _ilt(expr)
+
+    def _ilt_pow(e):
+        match = e.match((a*s + b)**n)
+        if match is not None:
+            nm, am, bm = match[n], match[a], match[b]
+            if nm.is_Integer and nm < 0:
+                if nm == 1:
```
namannimmo10, line 1457: `nm` can never be equal to `1`.

sympy/integrals/transforms.py
```diff
@@ -1412,14 +1414,60 @@ def inverse_laplace_transform(F, s, t, plane=None, **hints):
...
+                    return exp(-(bm/am)*t) / am
+                else:
+                    return t**(-nm-1)*exp(-(bm/am)*t)/(am**-nm*gamma(-nm))
```
namannimmo10, line 1460:
```suggestion

                return t**(-nm-1)*exp(-(bm/am)*t)/(am**-nm*gamma(-nm))
```

sympy/integrals/transforms.py
```diff
@@ -1412,14 +1414,60 @@ def inverse_laplace_transform(F, s, t, plane=None, **hints):
...
-    laplace_transform
+    laplace_transform, fast_inverse_laplace
     hankel_transform, inverse_hankel_transform
     """
     if isinstance(F, MatrixBase) and hasattr(F, 'applyfunc'):
         return F.applyfunc(lambda Fij: inverse_laplace_transform(Fij, s, t, plane, **hints))
     return InverseLaplaceTransform(F, s, t, plane).doit(**hints)
 
 
+def fast_inverse_laplace(e, s, t):
```
namannimmo10, line 1425: Can you test this? Also, include in the release notes?

oscarbenjamin: It should be tested.

It should either be private (with a leading underscore name) or it should have a proper docstring etc. It should only be mentioned in the release notes if it is public.

akshanshbhatt: > It should either be private (with a leading underscore name) or it should have a proper docstring etc.

I have made it private since I don't know much about the inner implementation. You can make it a public function and add docs/tests later according to the need.

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,919 @@
...
+    pole_markersize : Number, optional
+        The size of the markers used to mark the poles in the plot.
+        Default pole markersize is 10.
+    zero_color : str, tuple, optional
+        The color of the zero points on the plot. Default color
+        is orange. The color can be provided as a matplotlib color string,
+        or a 3-tuple of floats each in the 0-1 range.
+    zero_markersize : Number, optional
+        The size of the markers used to mark the zeros in the plot.
+        Default pole markersize is 7.
```
namannimmo10, line 139:
```suggestion
        Default zero markersize is 7.
```

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,919 @@
...
+    return LineOver1DRangeSeries(_y, (_x, lower_limit, upper_limit),
+        **kwargs).get_points()
+
+
+def ramp_response_plot(system, slope=1, color='b', prec=8, lower_limit=0,
+    upper_limit=10, show_axes=False, grid=True, show=True, **kwargs):
+    r"""
+    Returns the ramp response of a continuous-time system.
+
+    Ramp function is defined as the the straight line
```
namannimmo10, line 594:
```suggestion
    Ramp function is defined as the straight line
```

sympy/physics/control/control_plots.py
```diff
@@ -0,0 +1,919 @@
...
+    See Also
+    ========
+
+    bode_magnitude_plot, bode_phase_plot
+
+    References
+    ==========
+
+    .. [1] https://en.wikipedia.org/wiki/Ramp_function
+
```
namannimmo10, line 905:
```suggestion
```

namannimmo10:
Can you check locally whether it is all rendering fine?  Some cleanups might also be needed.

We can improve these plots in the future. I can merge this tonight.

akshanshbhatt:
> Can you check locally whether it is all rendering fine? Some cleanups might also be needed.

Do you mean the docs?

namannimmo10:
Yes!

akshanshbhatt:
> Yes!

[Have a look](https://github.com/akshanshbhatt/temp/blob/master/control_plots.pdf)

namannimmo10 (approved):
:+1: Thank you!

oscarbenjamin:
Having just released sympy 1.9 I've found that the tests added here fail:
```console
$ bin/test sympy/physics/control/tests/test_control_plots.py 
======================================== test process starts ========================================
executable:         /home/oscar/current/sympy/37venv/bin/python  (3.7.10-final-0) [CPython]
architecture:       64-bit
cache:              yes
ground types:       gmpy 2.0.8
numpy:              1.21.0
random seed:        40766605
hash randomization: on (PYTHONHASHSEED=3506573775)

sympy/physics/control/tests/test_control_plots.py[6] .F.FFF                                    [FAIL]

_____________________________________________________________________________________________________
_________________ sympy/physics/control/tests/test_control_plots.py:test_pole_zero __________________
  File "/home/oscar/current/sympy/sympy.git/sympy/physics/control/tests/test_control_plots.py", line 100, in test_pole_zero
    ((0.0,), ((-0.5000000000000004+0.8660254037844395j),
AssertionError
_____________________________________________________________________________________________________
______________ sympy/physics/control/tests/test_control_plots.py:test_impulse_response ______________
  File "/home/oscar/current/sympy/sympy.git/sympy/physics/control/tests/test_control_plots.py", line 192, in test_impulse_response
    assert impulse_res_tester(tf3, exp3)
AssertionError
_____________________________________________________________________________________________________
_______________ sympy/physics/control/tests/test_control_plots.py:test_step_response ________________
  File "/home/oscar/current/sympy/sympy.git/sympy/physics/control/tests/test_control_plots.py", line 241, in test_step_response
    assert step_res_tester(tf3, exp3)
AssertionError
_____________________________________________________________________________________________________
_______________ sympy/physics/control/tests/test_control_plots.py:test_ramp_response ________________
  File "/home/oscar/current/sympy/sympy.git/sympy/physics/control/tests/test_control_plots.py", line 283, in test_ramp_response
    assert ramp_res_tester(tf3, 10, exp3, 1.5)
AssertionError

======================== tests finished: 2 passed, 4 failed, in 4.73 seconds ========================
DO *NOT* COMMIT!

```
I guess these tests are not properly running in CI and also they seem to be incorrect.

PR #21768, opened 2021-07-21
Update the link to python's documentation

jond01:
Just that

NO ENTRY


oscarbenjamin:
Thanks for this. I fixed the release note section in the OP. You can just put NO ENTRY in the release note (unless a release note is actually needed) but the release note section needs to be there.

PR #21775, opened 2021-07-24
[GSoC] Limit bug fixes

### LINKED ISSUE #9497: polylog(S(3)/2, -z).series(z, 0, 5) fails
As of f74ad72890e91c09bb945449438da2ec2986f30a:

```
In [1]: polylog(S(3)/2, -z).series(z, 0, 5)
<ipython-input-1-cb514b271258> in <module>()
----> 1 polylog(S(3)/2, -z).series(z, 0, 5)

/home/ondrej/repos/sympy/sympy/core/expr.py in series(self, x, x0, n, dir, logx)
   2474             # replace x with an x that has a positive assumption
   2475             xpos = Dummy('x', positive=True, finite=True)
-> 2476             rv = self.subs(x, xpos).series(xpos, x0, n, dir, logx=logx)
   2477             if n is None:
   2478                 return (s.subs(xpos, x) for s in rv)

/home/ondrej/repos/sympy/sympy/core/expr.py in series(self, x, x0, n, dir, logx)
   2481 
   2482         if n is not None:  # nseries handling
-> 2483             s1 = self._eval_nseries(x, n=n, logx=logx)
   2484             o = s1.getO() or S.Zero
   2485             if o:

/home/ondrej/repos/sympy/sympy/core/function.py in _eval_nseries(self, x, n, logx)
    599                 #let's try the general algorithm
    600                 term = e.subs(x, S.Zero)
--> 601                 if term.is_finite is False or term is S.NaN:
    602                     raise PoleError("Cannot expand %s around 0" % (self))
    603                 series = term

AttributeError: 'int' object has no attribute 'is_finite'
```

### LINKED ISSUE #18363: Assume real root in limit
`Limit(((x - 6)**(Integer(1)/3) + 2)/(x + 2),x,-2,'-').doit()` returns `-oo*sign(2+2*(-1)^(1/3))`. How to make it real-valued?

### LINKED ISSUE #21598: Wrong limit of sum of logs
It seems like sometimes some cancellations involving logarithms are not taken into account by `.limit`. Here is an example:

``` python
>>> a, b, c = sym.symbols('a, b, c', positive=True)
>>> A = sym.log(a/b) - (sym.log(a) - sym.log(b))
>>> A.limit(a, sym.oo)
0          # correct
>>> (A * c).limit(a, sym.oo)
-c*log(b)  # wrong!!
```
The expression `A` is equivalent to zero, so its limit is (correctly) picked up as zero. However, when I multiply it by a symbol, the cancellation is not picked up. This doesn't happen when `c` is a concrete number, or when `a, b` are not declared `positive`.

This may seem like a trivial problem and one that I shouldn't need SymPy for, but I actually came across it when doing something much more complicated, which boils down to lome logarithms. Here it is, for completeness:
``` python
>>> tau, x, c = sym.symbols('tau, x, c', positive=True)
>>> res = manualintegrate(1 / x / (1+x)**2 * tau**2 / (tau**2 + x**2) * x**2, x)
>>> res.limit()
0  # wrong, with consequences for my research
>>> res.simplify().limit()
pi*tau**3/(tau**2 + 1)**2  # correct
>>> res.series(x, sym.oo, 1).removeO()
pi*tau**3/(tau**4 + 2*tau**2 + 1)  # also correct
```
In this example, if I omit the `positive=True`, even the second result (taking the limit after `.simplify()`) is wrong: it also gives zero! The series expansion remains correct.

(I'm running version 1.8, apologies, if this has been fixed since the latest release.)

### LINKED ISSUE #21756: Incorrect limit with ratio of complex exponentials
From SO:
https://stackoverflow.com/questions/68406242/managing-cases-of-an-expression-containing-positive-integers?noredirect=1#comment120914902_68406242

These limits all give 0 but the correct value is 5:
```python
In [1]: term = (1 - exp(-2*I*pi*z))/(1 - exp(-2*I*pi*z/5))

In [2]: term
Out[2]: 
     -2⋅ⅈ⋅π⋅z 
1 - ℯ         
──────────────
     -2⋅ⅈ⋅π⋅z 
     ─────────
         5    
1 - ℯ         

In [3]: term.limit(z, 0)
Out[3]: 0

In [4]: term.subs(z, 1e-6).n()
Out[4]: 4.99999999997631 - 1.25663706143261e-5⋅ⅈ

In [5]: re(term).limit(z, 0)
Out[5]: 0

In [6]: abs(term).limit(z, 0)
Out[6]: 0
```

0sidharth:

Fixes #9497
Fixes #18363
Fixes #21598
Fixes #21756 

#### Release Notes

* functions
  * Added `polylog` nseries expansion
  * Fixes bug in `erf` nseries expansion leading to wrong number of terms
  * `log` leading term method handles `zoo` better, thus raises more often
  * `binomial` leading term method added
* series
  * `Limit` now throws `NotImplementedError` if the point contains the variable wrt which limit is being found

### DIFF
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -979,9 +979,12 @@ class Add(Expr, AssocOp):
         return (self.func(*re_part), self.func(*im_part))
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy import expand_mul, Order
+        from sympy import expand_mul, Order, Piecewise, piecewise_fold
 
         old = self
+
+        if old.has(Piecewise):
+            old = piecewise_fold(old)
 
         expr = expand_mul(old)
         if not expr.is_Add:
@@ -1009,7 +1012,7 @@ class Add(Expr, AssocOp):
         if is_zero is None:
             new_expr = new_expr.trigsimp().cancel()
             is_zero = new_expr.is_zero
-        if is_zero is True:
+        if is_zero is True and min.expr != 0:
             # simple leading term analysis gave us cancelled terms but we have to send
             # back a term, so compute the leading term (via series)
             n0 = min.getn()
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -979,14 +979,22 @@ class Add(Expr, AssocOp):
         return (self.func(*re_part), self.func(*im_part))
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy import expand_mul, Order, Piecewise, piecewise_fold
+        from sympy import expand_mul, Order, Piecewise, piecewise_fold, log
 
         old = self
 
         if old.has(Piecewise):
             old = piecewise_fold(old)
 
+        # This expansion is the last part of expand_log. expand_log also calls
+        # expand_mul with factor=True, which would be more expensive
+        if any(isinstance(a, log) for a in self.args):
+            logflags = dict(deep=True, log=True, mul=False, power_exp=False,
+                power_base=False, multinomial=False, basic=False, force=False,
+                factor=False)
+            old = old.expand(logflags)
         expr = expand_mul(old)
+
         if not expr.is_Add:
             return expr.as_leading_term(x, logx=logx, cdir=cdir)
 
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -1020,7 +1020,7 @@ class Add(Expr, AssocOp):
         if is_zero is None:
             new_expr = new_expr.trigsimp().cancel()
             is_zero = new_expr.is_zero
-        if is_zero is True and min.expr != 0:
+        if is_zero is True:
             # simple leading term analysis gave us cancelled terms but we have to send
             # back a term, so compute the leading term (via series)
             n0 = min.getn()
--- a/sympy/series/limits.py
+++ b/sympy/series/limits.py
@@ -293,10 +293,8 @@ class Limit(Expr):
             newe = e.subs(z, z + z0)
         try:
             coeff, ex = newe.leadterm(z, cdir=cdir)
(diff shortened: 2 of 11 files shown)


sympy/series/limits.py
```diff
@@ -261,48 +283,48 @@ def set_signs(expr):
...
-                                return S.ComplexInfinity
+        if abs(z0) is S.Infinity:
+            # if e.is_Mul:
+            e = factor_terms(e)
+            newe = e.subs(z, 1/z)
+            cdir = -cdir
+        else:
+            newe = e.subs(z, z + z0)
+        try:
+            # cdir changes sign as oo- should become 0+
```
jksuom, line 294: This comment relates to line 290 above.

sympy/series/limits.py
```diff
@@ -261,48 +283,48 @@ def set_signs(expr):
...
+            coeff, ex = newe.leadterm(z, cdir=cdir)
+        except (ValueError, NotImplementedError, PoleError):
+            # The NotImplementedError catching is for custom functions
+            # AttributeError may be removed one TupleArg leading term
+            # is handled
+            if e.is_Pow:
+                r = self.pow_heuristics()
+                if r is not None:
+                    return r
+            pass
```
jksuom, line 304: `pass` could be omitted.

sympy/core/add.py
```diff
@@ -1009,7 +1020,7 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
         if is_zero is None:
             new_expr = new_expr.trigsimp().cancel()
             is_zero = new_expr.is_zero
-        if is_zero is True:
+        if is_zero is True and min.expr != 0:
```
0sidharth, line 1023: This change is not needed, I will remove it in my next commit

jksuom:
Looks good. Thanks!

PR #21474, opened 2021-05-15
Use strongly_connected_components for Matrix

sylee957:
#### References to other Issues or PRs

Fixes #21470 

#### Brief description of what is fixed or changed

In this pass, I'm attempting to use `strongly_connected_components` for places like `eigenvals`, `det` to replace the triangular heuristics and to replace the already present improvement done by `connected_components`.

~For computing `charpoly`, I took the reverse approach by computing the multiplication of polynomial
However, I don't think that any method using determinant or eigenvalues relies directly on `charpoly` after this, so it would just be an individual improvement. (Since there is no class in sympy that can correspond to `PolyMul`, I don't think that it can use the benefit of already computed factorization)
I may rather disentangle the `charpoly` to avoid doubly computing `strongly_connected_components`~
I won't attempt to improve `charpoly` here since it looks like messing up the organization of the module

#### Other comments

I would also want to correct the documentation `A = P B P^{-1}` to `A = P^{-1} B P` since I discovered that was mathematically incorrect by testing with random matrix.

I would also investigate if giving upper triangular decomposition is easy to build on top of this framework (like with transposition) or has some mathematical difficulty such that we have reason to provide upper or lower option.
The only reason that this gives lower triangular result is because it corresponds to the output of `iterable.strongly_connected_components`, but since lots of matrix methods like `rref` are based on upper triangulation, we need a way to convert it to upper triangular decomposition.

#### Release Notes

- matrices
  - Added `Matrix.strongly_connected_components` and `Matrix.strongly_connected_components_decomposition` for computing block triangulation of the matrix.

PR #21776, opened 2021-07-24
perf(matrices): speed up eigenvals for triangular matrices

oscarbenjamin:

Corrects a small performance regression from #21474

Speed up eigenvals for triangular matrices e.g.:
```python
def make(n):
    def entry(i, j):
        if i == j:
            return i
        elif i > j:
            return j
        else:
            return 0
    return Matrix(n, n, entry)

In [3]: M = make(5)

In [4]: M
Out[4]: 
⎡0  0  0  0  0⎤
⎢             ⎥
⎢0  1  0  0  0⎥
⎢             ⎥
⎢0  1  2  0  0⎥
⎢             ⎥
⎢0  1  2  3  0⎥
⎢             ⎥
⎣0  1  2  3  4⎦

In [5]: %time M.eigenvals()
CPU times: user 810 µs, sys: 313 µs, total: 1.12 ms
Wall time: 1.42 ms
Out[5]: {0: 1, 1: 1, 2: 1, 3: 1, 4: 1}
```
This had slowed down on master since 1.8 as a result of https://github.com/sympy/sympy/pull/21474 which is a performance improvement in general but slows down the particular case of a perfectly triangular matrix that factors completely into 1x1 blocks. The slowdown is due to calling calling charpoly and roots for trivial 1x1 matrices. This  PR just uses the value from the matrix directly as the eigenvalue if it is 1x1 which mostly restores the previous performance.

#### Other comments

This routine could become a lot faster for this kind of matrix after #21626.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/matrices/eigen.py
+++ b/sympy/matrices/eigen.py
@@ -10,7 +10,7 @@ from sympy.core.logic import fuzzy_and, fuzzy_or
 from sympy.core.numbers import Float
 from sympy.core.sympify import _sympify
 from sympy.functions.elementary.miscellaneous import sqrt
-from sympy.polys import roots, CRootOf, EX
+from sympy.polys import roots, CRootOf, ZZ, QQ, EX
 from sympy.polys.matrices import DomainMatrix
 from sympy.polys.matrices.eigen import dom_eigenvects, dom_eigenvects_to_sympy
 from sympy.simplify import nsimplify, simplify as _simplify
@@ -223,7 +223,16 @@ def _eigenvals_dict(
     M, error_when_incomplete=True, simplify=False, **flags):
     iblocks = M.strongly_connected_components()
     all_eigs = {}
+    is_dom = M._rep.domain in (ZZ, QQ)
     for b in iblocks:
+
+        # Fast path for a 1x1 block:
+        if is_dom and len(b) == 1:
+            index = b[0]
+            val = M[index, index]
+            all_eigs[val] = all_eigs.get(val, 0) + 1
+            continue
+
         block = M[b, b]
 
         if isinstance(simplify, FunctionType):
--- a/sympy/matrices/eigen.py
+++ b/sympy/matrices/eigen.py
@@ -157,8 +157,10 @@ def _eigenvals(
     if not M.is_square:
         raise NonSquareMatrixError("{} must be a square matrix.".format(M))
 
-    if all(x.is_number for x in M) and M.has(Float):
-        return _eigenvals_mpmath(M, multiple=multiple)
+    if M._rep.domain not in (ZZ, QQ):
+        # Skip this check for ZZ/QQ because it can be slow
+        if all(x.is_number for x in M) and M.has(Float):
+            return _eigenvals_mpmath(M, multiple=multiple)
 
     if rational:
         M = M.applyfunc(
--- a/sympy/matrices/eigen.py
+++ b/sympy/matrices/eigen.py
@@ -190,7 +190,16 @@ def _eigenvals_list(
     M, error_when_incomplete=True, simplify=False, **flags):
     iblocks = M.strongly_connected_components()
     all_eigs = []
+    is_dom = M._rep.domain in (ZZ, QQ)
     for b in iblocks:
+
+        # Fast path for a 1x1 block:
+        if is_dom and len(b) == 1:
+            index = b[0]
+            val = M[index, index]
+            all_eigs.append(val)
+            continue
+
         block = M[b, b]
 
         if isinstance(simplify, FunctionType):
--- a/sympy/matrices/graph.py
+++ b/sympy/matrices/graph.py
@@ -64,6 +64,7 @@ def _strongly_connected_components(M):
     if not M.is_square:
         raise NonSquareMatrixError
 
+    # RepMatrix uses the more efficient DomainMatrix.scc() method
     rep = getattr(M, '_rep', None)
     if rep is not None:
         return rep.scc()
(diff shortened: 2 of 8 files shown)


sylee957:
You also need to change `_eigenvals_list`

oscarbenjamin:
A problem with the approach here:
```python
In [1]: M = Matrix([[(x + y)/z, 0], [0, x/z + y/z]])

In [2]: M
Out[2]: 
⎡x + y       ⎤
⎢─────    0  ⎥
⎢  z         ⎥
⎢            ⎥
⎢       x   y⎥
⎢  0    ─ + ─⎥
⎣       z   z⎦

In [3]: M.eigenvals()
Out[3]: 
⎧x + y     x   y   ⎫
⎨─────: 1, ─ + ─: 1⎬
⎩  z       z   z   ⎭
```
On master that gives:
```python
In [2]: M.eigenvals()
Out[2]: 
⎧x + y   ⎫
⎨─────: 2⎬
⎩  z     ⎭
```
It's possible to fix that here by adding a call to `cancel` but then the speed benefit is lost.

Probably a better way is that after #21626 we could do this only when a proper domain is being used rather than for EXRAW.

smichr:
> possible to fix that here by adding a call

are there cases that need `cancel` or would `factor_terms` (which was made for the obvious cases) be sufficient:
```python
>>> factor_terms(x/z+y/z)
(x + y)/z
```

oscarbenjamin:
In SymPy 1.8 we have this:
```python
In [21]: M = Matrix([[(x**2 - y**2)/(x - y), 0], [0, x + y]])

In [22]: M
Out[22]: 
⎡ 2    2       ⎤
⎢x  - y        ⎥
⎢───────    0  ⎥
⎢ x - y        ⎥
⎢              ⎥
⎣   0     x + y⎦

In [23]: M.eigenvals()
Out[23]: 
⎧ 2    2             ⎫
⎪x  - y              ⎪
⎨───────: 1, x + y: 1⎬
⎪ x - y              ⎪
⎩                    ⎭

In [24]: roots(M.charpoly())
Out[24]: {x + y: 2}
```
However on master that is
```python
In [2]: M.eigenvals()
Out[2]: {x + y: 2}

In [3]: roots(M.charpoly())
Out[3]: {x + y: 2}
```
Getting the multiplicity right is important because e.g. with sympy 1.8 we have
```python
In [1]: M = Matrix([[(x**2 - y**2)/(x - y), 0], [0, x + y]])

In [2]: M.exp()
NonInvertibleMatrixError
```
whereas master gives
```python
In [1]: M = Matrix([[(x**2 - y**2)/(x - y), 0], [0, x + y]])

In [2]: M.exp()
Out[2]: 
⎡ x + y        ⎤
⎢ℯ         0   ⎥
⎢              ⎥
⎢         x + y⎥
⎣  0     ℯ     ⎦
```
Currently PR would restore the buggy sympy 1.8 behaviour.

smichr:
Does the following work without slowing things down too much?
```python
def merge_multiple(ev):
    from sympy.solvers.solvers import denoms
    from sympy.utilities.iterables import ordered
    from sympy.core.function import _mexpand
    from collections import defaultdict
    rv = {}
    d = defaultdict(list)
    for i in ev:
        d[tuple(ordered(i.free_symbols))].append(i)
    for i in d:
        if len(d[i]) == 1:
            rv[d[i]] = ev[d[i]]
            continue
        den = Mul(*denoms(Tuple(*d[i])), evaluate=False)
        uniq = defaultdict(list)
        for i in d[i]:
            uniq[_mexpand(den*i)].append(i)
        for i in uniq:
            k = next(ordered(uniq[i]))
            rv[k] = sum(ev[i] for i in uniq[i])
    return rv

    
>>> merge_multiple({(1/x+1/y):1,(x + y)/x/y:1})
{1/y + 1/x: 2}
>>> merge_multiple(Matrix([[(x**2 - y**2)/(x - y), 0], [0, x + y]]).eigenvals())
{x + y: 2}
```

oscarbenjamin:
> Does the following work without slowing things down too much?

I'm not sure but I decided just to only apply this optimisation when the domain of the matrix is ZZ or QQ since then all expressions are in a canonical form.

asmeurer:
Do the domains not include rational functions or other complex domains that also have canonical forms? Maybe we should add a `has_canonical` attribute to the domains. Or add `is_canonical` to the elements. That also opens the interesting possibility of not necessarily keeping things like rational functions in canonical form, which could be useful since cancelling after every operation might be slower than cancelling once at the end in some cases.

oscarbenjamin:
> Do the domains not include rational functions or other complex domains that also have canonical forms?

They do but the implementation I just merged in #21626 only uses the domains automatically for ZZ and QQ. Otherwise the EXRAW domain is used precisely because the canonical forms would not be backwards compatible e.g. right now you can do:
```python
In [15]: M = Matrix([[(x + 1)**10000000000000]])

In [16]: M
Out[16]: 
⎡       10000000000000⎤
⎣(x + 1)              ⎦

In [17]: M**2
Out[17]: 
⎡       20000000000000⎤
⎣(x + 1)              ⎦
```
If that automatically used `ZZ[x]` then those binomials would need to be expanded.

Of course for charpoly/eigenvals we probably do want to use the domains but we need some code to identify the appropriate domain at the start of the routine because Matrix will not do it at construction.

oscarbenjamin:
This is ready for review

sylee957:
> You also need to change `_eigenvals_list`

I don’t think that this comment is resolved yet. I expect simultaneous change for eigenvals_list

oscarbenjamin:
> I don’t think that this comment is resolved yet

Sorry, I forgot that one after the other changes...

oscarbenjamin:
> I expect simultaneous change for eigenvals_list

I've made those changes now. Those functions should be merged somehow (one should call the other).

oscarbenjamin:
Okay, now this is ready for review.

PR #21778, opened 2021-07-25
Add new abstract base class for methods

### LINKED ISSUE #21772: Create an abstract base class for the mechanics `*Method` classes and create a commit set of attributes & methods
We currently have `KanesMethod`, `LagrangesMethod`, and very soon `JointMethod`. `JointMethod` is unique in that it will use `KanesMethod` or `LagrangesMethod` internally, but, even so, it would be helpful if all three of these methods had some common attributes and methods so that duck typing can work well and there is consistency among the methods.

Some reasons making these more consistent will be helpful:

1. `JointMethod(..., method=KanesMethod|LagrangesMethod|OtherNewMethods)`: `JointMethod` should be able to use a number of other `Method` classes to actually formulate the equations of motion. `JointMethod` simple collects the joints and bodies then passes the correct information to the internal methods. Being able to pass in these and future methods to `JointMethod` would give a clean, simple API.
2. `pydy.system.System` currently consumes `KanesMethod` in its constructor. It should be able to consume any `*Method` object and also the `SymbolicSystem` object (https://github.com/sympy/sympy/blob/master/sympy/physics/mechanics/system.py).
3. The idea for `SymbolicSystem` was/is that each method object could produce the `SymbolicSystem`, e.g. `KanesMethod.export_symbolic_system()`. The symbolic system is supposed to be the most general container that can describe any symbolic multibody system. If that works, then `pydy.system.System()` could consume it instead of (or in addition) to the `*Methods` classes.
4. There are many more `*Methods` classes that could be created, e.g. `HamiltonsMethod`, `TMTMethod`, `NewtonEulerMethod`, etc. So having a common class to build them from is helpful.

Key things that `Method()` should likley have:

- coordinates
- speeds (if not the time derivative of coordinates)
- mass matrix, M of `M*u' = F`
- forcing equations, F of `M*u' = F`
- algebraic constraint equations
- equations for constraint forces (explicit or Lagrange multiplier eqs).
- kinematic different equations (if not q' = u)
- maybe more

sidhu1012:

Partially deals with #21772 

Added a new class `_Methods` which would serve as an abstract base class for all method(KanesMethod, JointsMethod etc)

#### Release Notes

* physics.mechanics
    * `KanesMethod`  can now take body list and force list in object instantiation.
    * `KanesMethod.kanes_equations` both parameters are optional and can be taken automaticaly if they are passed at object instantiation. 
    * `.bodies` attribute is added to `KanesMethod` which is advised to use instead of `.bodylist` attribute.

### DIFF
--- a/sympy/physics/mechanics/kane.py
+++ b/sympy/physics/mechanics/kane.py
@@ -533,8 +533,8 @@ class KanesMethod(_Methods):
         """
         if bodies is None:
             bodies = self.bodies
-        if  loads is None and self.loads is not None:
-            loads = self.loads
+        if  loads is None and self._forcelist is not None:
+            loads = self._forcelist
         if not self._k_kqdot:
             raise AttributeError('Create an instance of KanesMethod with '
                     'kinematic differential equations to use this method.')
--- a/sympy/physics/mechanics/kane.py
+++ b/sympy/physics/mechanics/kane.py
@@ -533,7 +533,7 @@ class KanesMethod(_Methods):
         """
         if bodies is None:
             bodies = self.bodies
-        if  loads is None and self._forcelist is not None:
+        if  loads is None and self._forcelist is not None and self._forcelist != []:
             loads = self._forcelist
         if not self._k_kqdot:
             raise AttributeError('Create an instance of KanesMethod with '
--- a/sympy/physics/mechanics/kane.py
+++ b/sympy/physics/mechanics/kane.py
@@ -533,8 +533,11 @@ class KanesMethod(_Methods):
         """
         if bodies is None:
             bodies = self.bodies
-        if  loads is None and self._forcelist is not None and self._forcelist != []:
-            loads = self._forcelist
+        if loads == []:
+            loads = None
+        if  loads is None and self._forcelist is not None:
+            if self._forcelist != []:
+                loads = self._forcelist
         if not self._k_kqdot:
             raise AttributeError('Create an instance of KanesMethod with '
                     'kinematic differential equations to use this method.')
--- a/sympy/physics/mechanics/kane.py
+++ b/sympy/physics/mechanics/kane.py
@@ -3,6 +3,7 @@ from sympy import solve_linear_system_LU
 from sympy.utilities import default_sort_key
 from sympy.physics.vector import (ReferenceFrame, dynamicsymbols,
                                   partial_velocity)
+from sympy.physics.mechanics.method import _Methods
 from sympy.physics.mechanics.particle import Particle
 from sympy.physics.mechanics.rigidbody import RigidBody
 from sympy.physics.mechanics.functions import (msubs, find_dynamicsymbols,
@@ -13,7 +14,7 @@ from sympy.utilities.iterables import iterable
 __all__ = ['KanesMethod']
 
 
-class KanesMethod:
+class KanesMethod(_Methods):
     """Kane's method object.
 
     Explanation
--- a/sympy/physics/mechanics/kane.py
+++ b/sympy/physics/mechanics/kane.py
@@ -31,7 +31,7 @@ class KanesMethod(_Methods):
 
     q, u : Matrix
         Matrices of the generalized coordinates and speeds
-    bodylist : iterable
+    bodies : iterable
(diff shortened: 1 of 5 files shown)


sidhu1012:
@moorepants This one's ready !

sidhu1012:
@moorepants This PR is ready to be merged.

sympy/physics/mechanics/kane.py
```diff
@@ -117,7 +118,7 @@ class KanesMethod:
     def __init__(self, frame, q_ind, u_ind, kd_eqs=None, q_dependent=None,
             configuration_constraints=None, u_dependent=None,
             velocity_constraints=None, acceleration_constraints=None,
-            u_auxiliary=None):
+            u_auxiliary=None, bodies=None, forcelist=None):
```
moorepants, line 121: Docstring about these new kwargs is needed.

sympy/physics/mechanics/kane.py
```diff
@@ -530,6 +531,10 @@ def kanes_equations(self, bodies, loads=None):
             Must be either a non-empty iterable of tuples or None which corresponds
             to a system with no constraints.
         """
+        if bodies is None:
+            bodies = self.bodies
+        if  loads is None and self._forcelist is not None and self._forcelist != []:
```
moorepants, line 536: This needs a unit test for these cases.

sympy/physics/mechanics/kane.py
```diff
@@ -530,6 +531,10 @@ def kanes_equations(self, bodies, loads=None):
             Must be either a non-empty iterable of tuples or None which corresponds
             to a system with no constraints.
         """
+        if bodies is None:
```
moorepants, line 534: Unit test for this is needed.

sympy/physics/mechanics/tests/test_kane.py
```diff
@@ -21,8 +21,10 @@ def test_one_dof():
     pa = Particle('pa', P, m)
     BL = [pa]
 
-    KM = KanesMethod(N, [q], [u], kd)
-    KM.kanes_equations(BL, FL)
+    KM = KanesMethod(N, [q], [u], kd, bodies=BL, forcelist=FL)
+    KM.kanes_equations()
```
moorepants, line 25: Both cases should still work, so don't remove the test, only add a new one.

sympy/physics/mechanics/tests/test_method.py
```diff
@@ -0,0 +1,5 @@
+from sympy.physics.mechanics.method import _Methods
+from sympy.testing.pytest import raises
+
+def test_method():
+    raises(TypeError, lambda: _Methods())
```
moorepants, line 5: Not sure what the purpose of this is.

sidhu1012: To make sure `_Method` class is working as desired.

sympy/physics/mechanics/tests/test_kane.py
sidhu1012: Okay

sympy/physics/mechanics/kane.py
sidhu1012: @moorepants  I don't think all other `__init__`  kwargs are documented.

sidhu1012:
@moorepants Review please.

sympy/physics/mechanics/kane.py
moorepants: That's unfortunate!

moorepants:
You need to add release notes for this, as it is a new addition.

sympy/physics/mechanics/tests/test_method.py
moorepants: So "as desired" means that if someone tries to use the class by itself a type error should be raised?

sidhu1012:
> You need to add release notes for this, as it is a new addition.

okay

sympy/physics/mechanics/tests/test_method.py
sidhu1012: Yes

sidhu1012:
@moorepants Release notes added.

PR #21799, opened 2021-07-30
Don't try to guard against non-hashable results in the cache

asmeurer:
See https://github.com/sympy/sympy/pull/21619#issuecomment-889540868. This can
cause issues when the function itself raises TypeError. Really, if a function
has the cacheit decorator and returns a non-hashable result, that is a bug.
Either the function should be fixed to return a hashable result, or some lower
level function that does always return hashable results should be cached
instead.

This commit is to see what sorts of tests will fail when this change is made.
We may also want to run the test suite in random order to try to find
failures, since that can affect caching.

#### Release Notes

- core
  - Fix the internal caching code to not run a cached function twice if it raises a TypeError.

### DIFF
--- a/sympy/core/cache.py
+++ b/sympy/core/cache.py
@@ -61,7 +61,28 @@ def __cacheit(maxsize):
         to force cacheit to check returned results mutability and consistency,
         set environment variable SYMPY_USE_CACHE to 'debug'
     """
-    return lru_cache(maxsize, typed=True)
+    def func_wrapper(func):
+        from .decorators import wraps
+
+        cfunc = lru_cache(maxsize, typed=True)(func)
+
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            try:
+                retval = cfunc(*args, **kwargs)
+            except TypeError as e:
+                if not e.args[0].startswith('unhashable type:'):
+                    raise
+                retval = func(*args, **kwargs)
+            return retval
+
+        wrapper.cache_info = cfunc.cache_info
+        wrapper.cache_clear = cfunc.cache_clear
+
+        CACHE.append(wrapper)
+        return wrapper
+
+    return func_wrapper
 ########################################
 
 
--- a/sympy/core/cache.py
+++ b/sympy/core/cache.py
@@ -61,26 +61,7 @@ def __cacheit(maxsize):
         to force cacheit to check returned results mutability and consistency,
         set environment variable SYMPY_USE_CACHE to 'debug'
     """
-    def func_wrapper(func):
-        from .decorators import wraps
-
-        cfunc = lru_cache(maxsize, typed=True)(func)
-
-        @wraps(func)
-        def wrapper(*args, **kwargs):
-            try:
-                retval = cfunc(*args, **kwargs)
-            except TypeError:
-                retval = func(*args, **kwargs)
-            return retval
-
-        wrapper.cache_info = cfunc.cache_info
-        wrapper.cache_clear = cfunc.cache_clear
-
-        CACHE.append(wrapper)
-        return wrapper
-
-    return func_wrapper
+    return lru_cache(maxsize, typed=True)
 ########################################
 
 
--- a/sympy/core/cache.py
+++ b/sympy/core/cache.py
@@ -71,7 +71,7 @@ def __cacheit(maxsize):
             try:
                 retval = cfunc(*args, **kwargs)
             except TypeError as e:
-                if not e.args[0].startswith('unhashable type:'):
+                if not e.args or not e.args[0].startswith('unhashable type:'):
                     raise
                 retval = func(*args, **kwargs)
             return retval
--- a/sympy/codegen/tests/test_algorithms.py
+++ b/sympy/codegen/tests/test_algorithms.py
@@ -6,7 +6,7 @@ from sympy.codegen.fnodes import bind_C
 from sympy.codegen.futils import render_as_module as f_module
 from sympy.codegen.pyutils import render_as_module as py_module
 from sympy.external import import_module
-from sympy.printing import ccode
+from sympy.printing.codeprinter import ccode
 from sympy.utilities._compilation import compile_link_import_strings, has_c, has_fortran
 from sympy.utilities._compilation.util import may_xfail
 from sympy.testing.pytest import skip, raises
--- a/sympy/codegen/tests/test_applications.py
+++ b/sympy/codegen/tests/test_applications.py
@@ -3,7 +3,7 @@
 import tempfile
 
 from sympy.external import import_module
-from sympy.printing import ccode
+from sympy.printing.codeprinter import ccode
 from sympy.utilities._compilation import compile_link_import_strings, has_c
 from sympy.utilities._compilation.util import may_xfail
 from sympy.testing.pytest import skip
--- a/sympy/codegen/tests/test_cnodes.py
+++ b/sympy/codegen/tests/test_cnodes.py
@@ -1,5 +1,5 @@
 from sympy.core.symbol import symbols
-from sympy.printing import ccode
+from sympy.printing.codeprinter import ccode
 from sympy.codegen.ast import Declaration, Variable, float64, int64, String, CodeBlock
 from sympy.codegen.cnodes import (
     alignof, CommaOperator, goto, Label, PreDecrement, PostDecrement, PreIncrement, PostIncrement,
--- a/sympy/codegen/tests/test_cxxnodes.py
+++ b/sympy/codegen/tests/test_cxxnodes.py
@@ -1,7 +1,7 @@
 from sympy.core.symbol import Symbol
 from sympy.codegen.ast import Type
 from sympy.codegen.cxxnodes import using
-from sympy.printing import cxxcode
+from sympy.printing.codeprinter import cxxcode
 
 x = Symbol('x')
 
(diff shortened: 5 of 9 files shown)


asmeurer:
Oh I think I got mixed up here. It's the input to the function that needs to be hashable, not the output. So the guard is useful. But it still should be done in a way that doesn't repeat the function call if the exception is coming from the function.

asmeurer:
The caching code itself seems to be confused about this as well. It states that the output should be hashable, but only the inputs need be. See https://docs.python.org/3/library/functools.html#functools.lru_cache.

Incidentally, found a bug here anyway

```diff
diff --git a/sympy/core/add.py b/sympy/core/add.py
index 2e33672ff3..3a33d39714 100644
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -992,7 +992,7 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
             logflags = dict(deep=True, log=True, mul=False, power_exp=False,
                 power_base=False, multinomial=False, basic=False, force=False,
                 factor=False)
-            old = old.expand(logflags)
+            old = old.expand(**logflags)
         expr = expand_mul(old)

         if not expr.is_Add:
```

asmeurer:
> Incidentally, found a bug here anyway

This same bug would be found if we used keyword-only arguments in `expand`. I think it's high time we did some keyword-only cleanups in the codebase.

asmeurer:
Several test failures from running the tests in random order, although I expect several of these are also present in master.

```
____________________________________________________________ sympy/codegen/tests/test_cnodes.py:test_sizeof _____________________________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_cnodes.py", line 78, in test_sizeof
    assert ccode(sz) == 'sizeof(%s)' % typename
TypeError: 'module' object is not callable
_________________________________________________________________________________________________________________________________________________________________________
_________________________________________________________ sympy/codegen/tests/test_cnodes.py:test_CommaOperator _________________________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_cnodes.py", line 20, in test_CommaOperator
    assert ccode(expr) == '(++(x), 2*x)'
TypeError: 'module' object is not callable
_________________________________________________________________________________________________________________________________________________________________________
__________________________________________________________ sympy/codegen/tests/test_cnodes.py:test_goto_Label ___________________________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_cnodes.py", line 29, in test_goto_Label
    assert ccode(g) == 'goto early_exit'
TypeError: 'module' object is not callable
_________________________________________________________________________________________________________________________________________________________________________
____________________________________________________________ sympy/codegen/tests/test_cnodes.py:test_struct _____________________________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_cnodes.py", line 93, in test_struct
    assert ccode(s) == (
TypeError: 'module' object is not callable
_________________________________________________________________________________________________________________________________________________________________________
_____________________________________________________________ sympy/codegen/tests/test_cnodes.py:test_union _____________________________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_cnodes.py", line 108, in test_union
    assert ccode(u) == (
TypeError: 'module' object is not callable
_________________________________________________________________________________________________________________________________________________________________________
_________________________________________________________ sympy/codegen/tests/test_cnodes.py:test_PreIncrement __________________________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_cnodes.py", line 66, in test_PreIncrement
    assert ccode(p) == '++(x)'
TypeError: 'module' object is not callable
_________________________________________________________________________________________________________________________________________________________________________
_________________________________________________________ sympy/codegen/tests/test_cnodes.py:test_PreDecrement __________________________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_cnodes.py", line 54, in test_PreDecrement
    assert ccode(p) == '--(x)'
TypeError: 'module' object is not callable
_________________________________________________________________________________________________________________________________________________________________________
____________________________________________________________ sympy/codegen/tests/test_cnodes.py:test_alignof ____________________________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_cnodes.py", line 14, in test_alignof
    assert ccode(ax) == 'alignof(x)'
TypeError: 'module' object is not callable
_________________________________________________________________________________________________________________________________________________________________________
_________________________________________________________ sympy/codegen/tests/test_cnodes.py:test_PostIncrement _________________________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_cnodes.py", line 72, in test_PostIncrement
    assert ccode(p) == '(x)++'
TypeError: 'module' object is not callable
_________________________________________________________________________________________________________________________________________________________________________
_________________________________________________________ sympy/codegen/tests/test_cnodes.py:test_PostDecrement _________________________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_cnodes.py", line 60, in test_PostDecrement
    assert ccode(p) == '(x)--'
TypeError: 'module' object is not callable
_________________________________________________________________________________________________________________________________________________________________________
_______________________________________________ sympy/codegen/tests/test_rewriting.py:test_create_expand_pow_optimization _______________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_rewriting.py", line 262, in test_create_expand_pow_optimization
    assert cc(x**4) == 'x*x*x*x'
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_rewriting.py", line 259, in <lambda>
    cc = lambda x: ccode(
TypeError: 'module' object is not callable
_________________________________________________________________________________________________________________________________________________________________________
____________________________________________________________ sympy/codegen/tests/test_cxxnodes.py:test_using ____________________________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/codegen/tests/test_cxxnodes.py", line 11, in test_using
    assert cxxcode(u1) == 'using std::vector'
TypeError: 'module' object is not callable

_________________________________________________________________________________________________________________________________________________________________________
___________________________________________ sympy/multipledispatch/tests/test_core.py:test_inheritance_and_multiple_dispatch ____________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/multipledispatch/tests/test_core.py", line 78, in test_inheritance_and_multiple_dispatch
    assert f(A(), C()) == (A, C)
AssertionError
_________________________________________________________________________________________________________________________________________________________________________
_____________________________________________________ sympy/physics/units/tests/test_quantities.py:test_convert_to ______________________________________________________
  File "/Users/aaronmeurer/Documents/Python/sympy/sympy/sympy/physics/units/tests/test_quantities.py", line 37, in test_convert_to
    assert q.convert_to(m) == 5000*m
AssertionError
```

asmeurer:
In fact it doesn't seem like any of those have to do with caching, although I'll investigate further, and also do some more randomized runs.

asmeurer:
I've pushed some minor fixes which aren't strictly related to the caching fix. I can separate them out if that is desired.

asmeurer:
Not seeing any more failures here. This shouldn't really bring up new failures, only prevent the double calling issue when a function does raise an exception.

asmeurer:
I pushed up a test for this. This is ready for review.

sympy/core/cache.py
```diff
@@ -70,7 +70,9 @@ def func_wrapper(func):
         def wrapper(*args, **kwargs):
             try:
                 retval = cfunc(*args, **kwargs)
-            except TypeError:
+            except TypeError as e:
+                if not e.args or not e.args[0].startswith('unhashable type:'):
+                    raise
```
oscarbenjamin, line 75: Parsing error strings seems flakey to me.

What if the cached function fails because of an unhashable type?

oscarbenjamin:
Can we just remove this error handling altogether?

Presumably there are some places in the codebase that depend on this but then maybe we should just remove the cacheit decorator in those places.

Maybe it would be difficult to do this but I guess the problem comes from using cacheit on a function that accepts unsympified types. After sympification any Basic instances should be hashable.

sympy/core/cache.py
asmeurer: The more correct way would be to check `hash(args)`, but this is more expensive, so probably not a good idea for this function.

asmeurer:
You can see the test failures from my first commit here for an idea of what that would be. Things like functions that can accept mutable matrices would no longer work. My feeling is that if we did this, a bunch of functions would just end up special-casing mutable input types, so we might as well just do that automatically in the decorator.

oscarbenjamin:
From the test failures in the first commit here it seems like there are two kinds of failures if this exception handling is removed:

1. The cache is not being cleared when global parameters change (I guess that's fixable).
2. Passing a mutable matrix as an argument to a Function fails because `Function.__new__` and `Application.__new__` use `cacheit`.

The matrix would be hashable after sympification since mutable matrices sympify to immutable matrices. If the caching was only applied after sympification then it would work. Maybe there should be an alternate form of `cacheit` that sympifies the arguments before calling the cached function.

Alternately I dislike things like `exp(M)` where `M` is a matrix. That should be a separate `MatrixExp` class or something like that but cases like `re(M)` seem more reasonable I guess (although it isn't returning a real number...).

oscarbenjamin:
The changes here look good to me.

PR #21801, opened 2021-07-30
diophantine(solvers) -  fixes issue#18081 and diophantine now handles simple cases with no solutions

### LINKED ISSUE #18081: Incorrect result for Intersection
`imageset(Lambda(n, n*log(2)), S.Integers).intersection(S.Integers)` should give {0}, but it doesn't.

praneethratna:
Fixes #18081  by modifying the code in sympy/solvers/diophantine/diophantine.py and also test cases are modified so that now diophantine handles simple cases with no solutions and also,
imageset(Lambda(n, n*log(2)), S.Integers).intersection(S.Integers) now returns FiniteSet(0)

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/solvers/diophantine/diophantine.py
+++ b/sympy/solvers/diophantine/diophantine.py
@@ -1387,7 +1387,7 @@ def diophantine(eq, param=symbols("t", integer=True), syms=None,
                 if len(i) == len(nzvar) and rv:
                     continue
                 rv.append(tuple(
-                        [0 if i in zvar else i for i in var]))
+                        [S.Zero if i in zvar else i for i in var]))
         return set(rv)
     try:
         p = Poly(eq)
--- a/sympy/solvers/diophantine/diophantine.py
+++ b/sympy/solvers/diophantine/diophantine.py
@@ -10,7 +10,7 @@ from sympy.core.numbers import igcdex, ilcm, igcd
 from sympy.core.power import integer_nthroot, isqrt
 from sympy.core.relational import Eq
 from sympy.core.singleton import S
-from sympy.core.symbol import Symbol, symbols
+from sympy.core.symbol import Symbol, symbols, Dummy
 from sympy.core.sympify import _sympify
 from sympy.functions.elementary.complexes import sign
 from sympy.functions.elementary.integers import floor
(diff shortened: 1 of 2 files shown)


sympy/solvers/diophantine/diophantine.py
```diff
@@ -1349,30 +1349,47 @@ def diophantine(eq, param=symbols("t", integer=True), syms=None,
     if isinstance(eq, Eq):
         eq = eq.lhs - eq.rhs
 
+    var = list(eq.expand(force=True).free_symbols)
```
smichr, line 1352: What is the purpose of expansion here?

sympy/solvers/diophantine/diophantine.py
```diff
@@ -1349,30 +1349,47 @@ def diophantine(eq, param=symbols("t", integer=True), syms=None,
...
+    if eq.xreplace(nzvar).is_integer is False:
+        # perhaps zero if 1 or more of the symbols is 0
+        rv = []
+        for i in subsets(nzvar):
+            zvar = {s: 0 for s in i}
+            if eq.xreplace(zvar).is_integer:
+                if len(i) == len(nzvar) and rv:
+                    continue
+                rv.append(tuple(
+                        [0 if i in zvar else i for i in var]))
```
smichr, line 1390:
```suggestion
                        [S.Zero if i in zvar else i for i in var]))
```

praneethratna (commented):
also should a test be included for imageset(Lambda(n, n*log(2)), S.Integers).intersection(S.Integers) that asserts it equal to FiniteSet(0) ?

smichr:
> should a test be included for imageset(Lambda(n, n*log(2)), S.Integers).intersection(S.Integers)

In my opinion, no: we test `sin(x)` and `x*pi` with `diophantine` and they confirm that the root issue has been addressed.

smichr:
Let's see if there are other comments, otherwise I think this is ready.

praneethratna:
> Let's see if there are other comments, otherwise I think this is ready.

@smichr can this be merged?

smichr:
While factoring and finding integer solutions is ok, reporting the lucky 0-is-a-solution case without assuring that this is the only integer that is a solution is problematic. Until we have a method of knowing what integer solutions a transcendental equation might have we should not report a guessed solution without a warning or indication that the solution is incomplete.

So if a factor in the equation is univariate we should use `solveset` over integers to see if there is a solution.

oscargus:
Probably makes sense to add the original issue from #18081 as a test so that it can be closed as well.

sympy/solvers/diophantine/diophantine.py
```diff
@@ -1349,30 +1349,47 @@ def diophantine(eq, param=symbols("t", integer=True), syms=None,
...
+        good = diophantine(n) - dsol
+        return {s for s in good if _mexpand(d.subs(zip(var, s)))}
+    else:
+        eq = n
+    eq = factor_terms(eq)
+    assert not eq.is_number
+    eq = eq.as_independent(*var, as_Add=False)[1]
+
+    # check for plausible integer non-solutions
+    ivar = {s: Dummy(integer=True) for s in var if not s.is_integer}
```
Shekharrajak, line 1377: I think we should define private methods for these regular checks, to make code readable and reusable.

Shekharrajak (changes requested):

Issue #17944, opened 2019-11-21
Using keyword only arguments


asmeurer:
Once we have dropped Python 2, we can use keyword only arguments for function definitions. This means a function like

```py
def simplify(expr, ratio=1.7, measure=count_ops, rational=False, inverse=False, doit=True, **kwargs):
```

can become 

```py
def simplify(expr, *, ratio=1.7, measure=count_ops, rational=False, inverse=False, doit=True, **kwargs):
```

(alternately you can add the keyword arguments after `**kwargs`). 

The extra `*` in the parameters list makes it so that the keywords have to be named explicitly when calling the function. So `simplify(expr, 1.6)` would no longer work. It would have to be `simplify(expr, ratio=1.7). The advantage of this is that it makes it easy to add new keyword arguments in the future without breaking compatibility. It also makes the function calls be more self-documenting.

An issue is that this is technically a backwards compatibility break for any existing function that we change. However, I think that calling keyword arguments by position is similar to using a private API, because there is never a guarantee that the arguments would stay in that order. 

Not all keyword arguments should be keyword-only. Sometimes a keyword argument is actually a positional argument with a default value. The ones that should be keyword-only are the ones that are actual flags or options to the function. For example, `dsolve` has the definition

```
def dsolve(eq, func=None, hint="default", simplify=True,
    ics= None, xi=None, eta=None, x0=0, n=6, **kwargs):
```

`func` is a keyword argument because it can be omitted, but it should not be keyword-only. I think `hint` also shouldn't be keyword-only, because including it as the third argument is used somewhat often, and is even documented as such. But the rest should be. 

This is somewhat related to https://github.com/sympy/sympy/issues/14030. 

Note that https://github.com/sympy/sympy/issues/17943 needs to be fixed (in particular, removing Python 2 from Travis) before this can be done. If you want to start work on it, you can, but the PR won't be mergeable until that happens.

PR #21803, opened 2021-07-30
Convert various functions to use keyword-only arguments

asmeurer:

https://github.com/sympy/sympy/issues/17944 (shouldn't mark as closed because this PR isn't going to cover every function in SymPy)

I am fixing various function signatures to use keyword only arguments. So for example,

```py
def expand(self, deep=True, modulus=None, power_base=True, power_exp=True,
           mul=True, log=True, multinomial=True, basic=True, **hints):
```

becomes

```py
def expand(self, *, deep=True, modulus=None, power_base=True, power_exp=True,
           mul=True, log=True, multinomial=True, basic=True, **hints):
```

This means that things like `expr.expand(True)` are no longer valid. You will have to use `expr.expand(deep=True)` instead.

#### Other comments

This is a huge undertaking and I probably won't get to every function in SymPy. Please feel free to merge this at any time if the tests are passing, and to pick up this work. 

#### Release Notes

- core
  - Change several function signatures to use keyword-only arguments.
- functions
  - Change several function signatures to use keyword-only arguments.

### DIFF
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -509,8 +509,8 @@ class Add(Expr, AssocOp):
             return terms[0].matches(expr - coeff, repl_dict)
         return
 
-    def matches(self, expr, repl_dict={}, old=False):
-        return self._matches_commutative(expr, repl_dict, old)
+    def matches(self, expr, repl_dict={}, *, old=False):
+        return self._matches_commutative(expr, repl_dict, old=old)
 
     @staticmethod
     def _combine_inverse(lhs, rhs):
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -992,7 +992,7 @@ class Add(Expr, AssocOp):
             logflags = dict(deep=True, log=True, mul=False, power_exp=False,
                 power_base=False, multinomial=False, basic=False, force=False,
                 factor=False)
-            old = old.expand(logflags)
+            old = old.expand(**logflags)
         expr = expand_mul(old)
 
         if not expr.is_Add:
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -1639,10 +1639,10 @@ class Basic(Printable, metaclass=ManagedProperties):
         # done
         return m
 
-    def count_ops(self, visual=None):
+    def count_ops(self, *, visual=None):
         """wrapper for count_ops that returns the operation count."""
         from sympy import count_ops
-        return count_ops(self, visual)
+        return count_ops(self, visual=visual)
 
     def doit(self, **hints):
         """Evaluate objects that are not evaluated by default like limits,
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -1521,7 +1521,7 @@ class Basic(Printable, metaclass=ManagedProperties):
         query = _make_find_query(query)
         return sum(bool(query(sub)) for sub in preorder_traversal(self))
 
-    def matches(self, expr, repl_dict={}, old=False):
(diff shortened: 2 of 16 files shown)


asmeurer:
I should note that this isn't always a trivial fix. In some cases, we really do want to allow keyword arguments to be passed as positional arguments. I'm trying to be conservative here. For instance, I have changed ` def matches(self, expr, repl_dict={}, old=False)` to `def matches(self, expr, repl_dict={}, *, old=False)` because it seems reasonable to me to pass the repl_dict as a positional argument.

oscarbenjamin:
> Significantly changed benchmark results (PR vs master)
> 
> ```diff
>        before           after         ratio
>      [78380bf1]       [14dd8937]
> !        544±10ns           failed      n/a  core.expand.TimeExpand.time_expand_nothing_todo
> ```

Looks like this causes a benchmark to fail:
https://github.com/sympy/sympy_benchmarks/blob/c9e500f8c7a2c9ed6496103dc7c1278e8a900635/benchmarks/core/expand.py#L10-L20
(I guess `bench_expand_32` isn't even run!)

Those benchmarks look broken to me but it makes me wonder if a deprecation warning should be given for this change.

asmeurer:
But the code in the benchmarks is nonsense. It's calling `p.expand(self)`, which amounts to `p.expand(deep=self)` where `self` is the benchmarking class. This is exactly the sort of bug that keyword-only arguments are designed to fix. If there is some case where an argument is used positionally "legitimately", we might consider that a sign that it shouldn't be keyword-only.

oscarbenjamin:
> But the code in the benchmarks is nonsense.

I agree but the thing is that currently that code "works" and the change here breaks it. Also in principle someone could be using this legitimately like `p.expand(False)`. I'm just wondering whether or not there should be a warning rather than causing downstream code to blow up e.g.:
```python
def expand(*args, deep=True, ...):
    if args:
        warn("positional arguments to expand are ignored and will raise an exception soon")
    ...
```
If we do intend to switch to introduce keyword-only arguments wholesale then there is a lot of potential for breakage, especially with commonly used core routines like `expand`. A deprecation warning could reference a deprecation removal issue that explains why these are deprecated.

asmeurer:
If we do do that we should do it with a decorator. It's a little more nontrivial than that. We'd want the existing syntax to continue to work (with a warning), and for `help(f)` to still show the correct signature (I think it can be done by setting `__signature__`). I still personally think it's not necessary but I don't feel strongly about it.

oscarbenjamin:
I'm not sure that it's necessary either but a decorator would be the best approach if it does not lead to significant slowdowns.

@bjodah does your thumbs up above mean that you think we should preserve compatibility with a warning?

Does anyone else have any opinions about this?

bjodah:
Yes, I think that using deprecation notifications for 1 or 2 releases for all potentially breaking changes is good practice. Using a decorator to achieve this, in this case, sounds like a rather clean way. If there's some (small) performance degradation I think it's acceptable for the transition period as well (since those would soon be removed either way).

oscarbenjamin:
I'm not sure if 1 or 2 releases is the right way to define it given that we don't have a regular release cycle. Maybe 2 years is a better way to think about it?

It's not uncommon to see bug reports of regressions since versions that are more than 2 years old which shows how often some users update their SymPy installation.

bjodah:
Sure, time matters more.

asmeurer:
Our deprecation policy needs to be refined so that we actually specify how long deprecations last. I don't know what the right answer is here, but if we do do a deprecation, the actual length of time isn't so important to decide now. We can remove it when we remove it.

PR #21817, opened 2021-08-03
stats - fixes issue#21057 by change value check to positive semi-definite

### LINKED ISSUE #21057: Cannot create joint normal distribution with positive semidefinite covariance matrix
I want to create a joint normal distributed random variable with a covariance matrix that is positive semidefinite but not positive definite. This seems to be impossible:

```python
import sympy.stats

sympy.stats.Normal("x", [0, 0], [[0,0], [0,0]])
```
gives

```
  File "<ipython-input-28-3bd23a2d609a>", line 1, in <module>
    sympy.stats.Normal("x", [0, 0], [[0,0], [0,0]])

  File "/home/.local/lib/python3.8/site-packages/sympy/stats/crv_types.py", line 2966, in Normal
    return MultivariateNormal(name, mean, std)

  File "/home/.local/lib/python3.8/site-packages/sympy/stats/joint_rv_types.py", line 209, in MultivariateNormal
    return multivariate_rv(MultivariateNormalDistribution, name, mu, sigma)

  File "/home/.local/lib/python3.8/site-packages/sympy/stats/joint_rv_types.py", line 28, in multivariate_rv
    dist.check(*args)

  File "/home/.local/lib/python3.8/site-packages/sympy/stats/joint_rv_types.py", line 143, in check
    _value_check(sigma.is_positive_definite,

  File "/home/.local/lib/python3.8/site-packages/sympy/stats/rv.py", line 1487, in _value_check
    raise ValueError(message)

ValueError: The covariance matrix must be positive definite. 
```

I don't know if there is some implementation reason for this, but mathematically any positive semidefinite matrix should be accepted. I checked numpy and there it works to do:

```python
import numpy as np

np.random.multivariate_normal([0, 0], [[0, 0], [0, 0]])
```

with the expected output:

```
Out[30]: array([0., 0.])
```

praneethratna:
Fixes #21057 and now joint normal distribution with a positive semi-definite covariance matrix

#### Release Notes

* stats
    * Added support for positive semi-definite covariance matrix in joint Multivariate normal distribution.

### DIFF
--- a/sympy/stats/joint_rv_types.py
+++ b/sympy/stats/joint_rv_types.py
@@ -176,6 +176,8 @@ def MultivariateNormal(name, mu, sigma):
 
     The density of the multivariate normal distribution can be found at [1].
 
+    Sigma can now be positive semidefinite but only supports sampling.
+
     Parameters
     ==========
 
--- a/sympy/stats/joint_rv_types.py
+++ b/sympy/stats/joint_rv_types.py
@@ -176,14 +176,13 @@ def MultivariateNormal(name, mu, sigma):
 
     The density of the multivariate normal distribution can be found at [1].
 
-    Sigma can now be positive semidefinite but only supports sampling.
-
     Parameters
     ==========
 
     mu : List representing the mean or the mean vector
     sigma : Positive definite square matrix
         Represents covariance Matrix
+        If `sigma` is noninvertible then only sampling is supported currently
 
     Returns
     =======
--- a/sympy/stats/joint_rv_types.py
+++ b/sympy/stats/joint_rv_types.py
@@ -180,7 +180,7 @@ def MultivariateNormal(name, mu, sigma):
     ==========
 
     mu : List representing the mean or the mean vector
-    sigma : Positive definite square matrix
+    sigma : Positive semidefinite square matrix
         Represents covariance Matrix
         If `sigma` is noninvertible then only sampling is supported currently
 
--- a/sympy/stats/joint_rv_types.py
+++ b/sympy/stats/joint_rv_types.py
@@ -139,10 +139,10 @@ class MultivariateNormalDistribution(JointDistribution):
     def check(mu, sigma):
         _value_check(mu.shape[0] == sigma.shape[0],
             "Size of the mean vector and covariance matrix are incorrect.")
-        #check if covariance matrix is positive definite or not.
+        #check if covariance matrix is positive semi definite or not.
         if not isinstance(sigma, MatrixSymbol):
-            _value_check(sigma.is_positive_definite,
-            "The covariance matrix must be positive definite. ")
+            _value_check(sigma.is_positive_semidefinite,
+            "The covariance matrix must be positive semi definite. ")
 
     def pdf(self, *args):
         mu, sigma = self.mu, self.sigma
--- a/sympy/stats/tests/test_joint_rv.py
+++ b/sympy/stats/tests/test_joint_rv.py
@@ -357,6 +357,7 @@ def test_sample_seed():
 def test_issue_21057():
     m = Normal("x", [0, 0], [[0, 0], [0, 0]])
     n = MultivariateNormal("x", [0, 0], [[0, 0], [0, 0]])
+    p = Normal("x", [0, 0], [[0, 0], [0, 1]])
     assert m == n
     libraries = ['scipy', 'numpy', 'pymc3']
     for library in libraries:
@@ -365,6 +366,9 @@ def test_issue_21057():
             if imported_lib:
                 s1 = sample(m, size=8)
                 s2 = sample(n, size=8)
+                s3 = sample(p, size=8)
                 assert tuple(s1.flatten()) == tuple(s2.flatten())
+                for s in s3:
+                    assert tuple(s.flatten()) in p.pspace.distribution.set
         except NotImplementedError:
             continue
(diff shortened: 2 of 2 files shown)


Maelstrom6:
Sorry about the late response, studying is getting hectic.

Some other tests that we can add to make sure that it's usable:
- Check that E(X) is a vector of zeros
- Check that Var(X) is a matrix of zeros
- It might be a good idea to test that the sample produces a vector of zeros too
- This might be a big problem but preferably `E(X*Y.T)` should be able to work too. But I think that might be beyond the scope of the stats module at the moment

Are there any other things that people might use a random variable with a semidefinite matrix for?

praneethratna:
@Maelstrom6 i'm kinda struck up here , E(X) returns an error stating sympy.matrices.common.NonInvertibleMatrixError: Matrix det == 0; not invertible. and similar error for variance(X) , could you help me figure this out?

Maelstrom6:
So I found out today that SymPy computes expectations by first principles every time for every kind of distribution. Nothing is precomputed. This module is long overdue for an overhaul and, when it was first designed, it didn't have multivariate distributions in mind. We might just have to leave out expectation calculations for now.

The sampling seems to work fine so it might be helpful to add a test for that. You can look at similar tests done by other distributions in the testing module. 

I'm worried that with general use-cases other than sampling, the user will get lots of `NonInvertibleMatrixError`s.

praneethratna:
> So I found out today that SymPy computes expectations by first principles every time for every kind of distribution. Nothing is precomputed. This module is long overdue for an overhaul and, when it was first designed, it didn't have multivariate distributions in mind. We might just have to leave out expectation calculations for now.
> 
> The sampling seems to work fine so it might be helpful to add a test for that. You can look at similar tests done by other distributions in the testing module.
> 
> I'm worried that with general use-cases other than sampling, the user will get lots of `NonInvertibleMatrixError`s.

But does sympy support  a pdf of dirac delta function ? Then should'nt it be feasible to create a Normal distribution with 0 Standard deviation and expectation?

Maelstrom6:
SymPy does have a Dirac delta function and we could return it as a PDF but then again, Wikipedia says that if the matrix is noninvertible, then the PDF does not exist. I'm not sure how correct they are there.

I'd guess a last thing that you can change for now is to change the docstring of `MultivariateNormal`. Sigma can now be positive semidefinite but only supports sampling.

praneethratna:
> SymPy does have a Dirac delta function and we could return it as a PDF but then again, Wikipedia says that if the matrix is noninvertible, then the PDF does not exist. I'm not sure how correct they are there.
> 
> Edit:
> With `Normal('X', [0, 0], [0, 0, 0, 1])`, I don't think it is possible to express the PDF even with Dirac Deltas. It might be useful to add a test for some normal distribution of that nature for sampling as well just to make sure it works.
> 
> I'd guess a last thing that you can change for now is to change the docstring of `MultivariateNormal`. Sigma can now be positive semidefinite but only supports sampling.

I've modified the docstring , is it fine or should i add that statement in parameters part?

Maelstrom6:
Maybe combine both statements so that we get something like:

```
sigma : Positive semidefinite square matrix
        Represents covariance Matrix
        If `sigma` is noninvertible then only sampling is supported currently
```

Once that's done, we could get @oscarbenjamin to see if they have any other suggestions.

oscarbenjamin:
Maybe if the covariance matrix is a zero matrix then a constant expression can be returned instead of a random variable.

praneethratna:
@Maelstrom6 I've modified the docstrings, What else changed should be made?

sympy/stats/joint_rv_types.py
```diff
@@ -182,6 +182,7 @@ def MultivariateNormal(name, mu, sigma):
     mu : List representing the mean or the mean vector
     sigma : Positive definite square matrix
```
Maelstrom6, line 183: Should it say `sigma : Positive semidefinite square matrix` instead?

Maelstrom6:
It will also be a good idea to check the sampling for `Normal([0, 0], [0, 0, 0, 1])`. Once that's done it should be good to me.

praneethratna:
> It will also be a good idea to check the sampling for `Normal([0, 0], [0, 0, 0, 1])`. Once that's done it should be good to me.

@Maelstrom6 I've made the changes , and also added sampling for `Normal([0, 0], [[0, 0], [0, 1]])` could you please review the same?

Maelstrom6:
> Maybe if the covariance matrix is a zero matrix then a constant expression can be returned instead of a random variable.

Maybe the PR that implements sampling of constants should do that instead.

This PR looks good to me.

sympy/stats/tests/test_joint_rv.py
```diff
@@ -354,3 +352,23 @@ def test_sample_seed():
...
+    p = Normal("x", [0, 0], [[0, 0], [0, 1]])
+    assert m == n
+    libraries = ['scipy', 'numpy', 'pymc3']
+    for library in libraries:
+        try:
+            imported_lib = import_module(library)
+            if imported_lib:
+                s1 = sample(m, size=8)
+                s2 = sample(n, size=8)
+                s3 = sample(p, size=8)
```
oscarbenjamin, line 369: This doesn't pass the library argument to sample so only scipy is tested. The test actually fails for pymc3. I've fixed the test in #23512 but I had to XFAIL the pymc3 part because it doesn't work.

PR #21833, opened 2021-08-06
Implement `MIMOFeedback`

akshanshbhatt:

- Support for closed-loop feedback with MIMO systems.
- Support for positive feedback loop added.
- Made the doc for the `Feedback` class more beginner-friendly.
- Fixed bugs in the `Feedback` class from previous releases.

#### Release Notes

* physics.control
  * Added `MIMOFeedback` class to support closed-loop MIMO systems.


sympy/physics/control/lti.py
```diff
@@ -1719,37 +1745,44 @@ class Feedback(Basic):
...
-            raise ValueError("The numerator cannot be equal to the denominator.")
-        if not num.var == den.var:
-            raise ValueError("Both numerator and denominator should be using the"
+    def __new__(cls, plant, feedback_controller=None, ftype=-1):
+        if not feedback_controller:
+            feedback_controller = TransferFunction(1, 1, plant.var)
+
+        if not (isinstance(plant, (TransferFunction, Series))
+            and isinstance(feedback_controller, (TransferFunction, Series))):
+            raise TypeError("Unsupported type for `plant` or `feedback_contoller` of Feedback.")
```
plaes, line 1764: typo: contoller

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1889,188 @@ def doit(self, **kwargs):
...
+    See Also
+    ========
+
+    Feedback, MIMOSeries, MIMOParallel
+
+    """
+    def __new__(cls, plant, feedback_controller, ftype=-1):
+        if not (isinstance(plant, (TransferFunctionMatrix, MIMOSeries))
+            and isinstance(feedback_controller, (TransferFunctionMatrix, MIMOSeries))):
+            raise TypeError("Unsupported type for `plant` or `feedback_contoller` of MIMO Feedback.")
```
plaes, line 1982: typoe: contoller

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,57 @@ def __neg__(self):
...
-        The feedback plant (often a feedback controller).
+    plant : Series, TransferFunction
+        The primary plant. Also commonly known as the open
+        loop gain / open-loop tranfer function.
+    feedback_controller : Series, TransferFunction, optional
+        The feedback plant (often a feedback controller). Also
+        commonly known as the feedback factor.
+        If not specified explicitly, the feedback_controller is
+        assumed to be unit (1) transfer function.
+    ftype : int, optional
```
plaes, line 1697: Don't you want to make the feedback to enum-like variable? `Feedback.Positive`, `Feedback.Negative`?
And then use the `cls._ftype` to silently handle the conversion to `{+,-}1`?

plaes (commented):
Bunch of general comments regarding code style and typos.

akshanshbhatt:
@namannimmo10 This PR is done from my side. Please review.

doc/src/modules/physics/control/control.rst
```diff
@@ -10,8 +10,9 @@ property is used for transfer functions in the ``Parallel`` class, and the
 multiplicative property is used for transfer functions in the ``Series`` class.
 Also, there is a ``Feedback`` class which is used to represent negative feedback
 interconnection between two input/output systems. MIMO systems are also supported
-with ``TransferFunctionMatrix`` as the base class for representing one. ``MIMOSeries``
-and ``MIMOParallel`` are MIMO equivalent of ``Series`` and ``Parallel`` classes.
+with ``TransferFunctionMatrix`` as the base class for representing one. ``MIMOSeries``,
+``MIMOParallel``  and ``MIMOFeedback`` are MIMO equivalent of ``Series``, ``Parallel``
+and ``Feedback`` classes.
```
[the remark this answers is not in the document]
akshanshbhatt: That can perhaps be added in the subsequent _control_problems_ PR.

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1905,398 @@ def doit(self, **kwargs):
...
+        [                 s                                  1 - s              ]
+        [              -------                            -----------           ]
+        [              9*s + 1                            s*(9*s + 1)           ]
+        [                                                                       ]
+        [100*s*(s - 1) + 10*(1 - s)*(9*s + 1)  s*(s - 1)*(9*s + 1) + 100*s - 100]
+        [------------------------------------  ---------------------------------]
+        [         (1 - s)*(9*s + 1)                       s*(9*s + 1)           ]{t}
+
+        If the user wants the the resultant ``TransferFunctionMatrix`` object without
+        cancelling the common factors then the ``cancel`` arg should be passed ``False``.
```
namannimmo10, line 2252:
```suggestion
        If the user wants the resultant ``TransferFunctionMatrix`` object without
        canceling the common factors then the ``cancel`` arg should be passed ``False``.
```
Can we provide an example where it actually avoids canceling the common factors? It is not apparent in this one.

namannimmo10 (commented):
The unit tests look solid! Give me some time, and I'll review the implementation and docstrings.

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1905,398 @@ def doit(self, **kwargs):
...
+
+        return TransferFunction(_num, _den, self.var)
 
-    def _eval_rewrite_as_TransferFunction(self, num, den, **kwargs):
+    def _eval_rewrite_as_TransferFunction(self, num, den, ftype, **kwargs):
         return self.doit()
 
     def __neg__(self):
-        return Feedback(-self.num, self.den)
+        return Feedback(-self.plant, -self.feedback_controller, self.ftype)
```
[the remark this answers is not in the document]
akshanshbhatt: I was a bit confused on this one.
This is what I thought -
```
P => Plant
C => Feedback Controller
F => Feedback
```
Then,
```
F(P, C) = P/(1 + PC)
-F(P, C) = -P/(1 + PC) = -P/(1 + (-P)*(-C)) = F(-P, -C)
```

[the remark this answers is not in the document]
namannimmo10:
```
>>> P = TransferFunction(1, s, s)
>>> C = TransferFunction(s, s - 1, s)
>>> P*C == (-P)*(-C)
False
```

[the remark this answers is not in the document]
akshanshbhatt: Indeed both these expressions would not be "symbolically" equal per se, but they are both identical in every other way. It reminds me of the comment by @moorepants (https://github.com/sympy/sympy/pull/21634#issuecomment-864504405). In symbolic, as there are a dozen different ways to write a single expression, there is no definitive approach to check for equality. Let me know what you think.
Is there a better way to implement this? I am open to suggestions.

sympy/physics/control/tests/test_lti.py
```diff
@@ -866,44 +866,44 @@ def test_Feedback_construction():
...
 
     f5 = Feedback(tf5, TransferFunction(1, 1, s))
-    assert f5.args == (tf5, TransferFunction(1, 1, s))
+    assert f5.args == (tf5, TransferFunction(1, 1, s), -1)
     assert f5.var == s
 
     f6 = Feedback(TransferFunction(1, 1, p), tf4)
-    assert f6.args == (TransferFunction(1, 1, p), tf4)
+    assert f6.args == (TransferFunction(1, 1, p), tf4, -1)
     assert f6.var == p
```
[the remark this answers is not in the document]
akshanshbhatt: Done

sympy/physics/control/lti.py
akshanshbhatt: Done

sympy/physics/control/tests/test_lti.py
```diff
@@ -866,44 +866,44 @@ def test_Feedback_construction():
...
+    assert f7.args == (Series(TransferFunction(-1, 1, p), Series(tf4, tf6)), -TransferFunction(1, 1, p), -1)
+    assert f7.plant == Series(TransferFunction(-1, 1, p), Series(tf4, tf6))
 
     # denominator can't be a Parallel instance
     raises(TypeError, lambda: Feedback(tf1, tf2 + tf3))
     raises(TypeError, lambda: Feedback(tf1, Matrix([1, 2, 3])))
     raises(TypeError, lambda: Feedback(TransferFunction(1, 1, s), s - 1))
     raises(TypeError, lambda: Feedback(1, 1))
-    raises(ValueError, lambda: Feedback(TransferFunction(1, 1, s), TransferFunction(1, 1, s)))
+    # raises(ValueError, lambda: Feedback(TransferFunction(1, 1, s), TransferFunction(1, 1, s)))
```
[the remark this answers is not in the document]
namannimmo10: (Marked unresolved!)

This need to be fixed:
```python
>>> Feedback(TransferFunction(1, 1, s), TransferFunction(1, 1, s))
  File "/home/namannimmo/oss/sympy/sympy/physics/control/lti.py", line 1770, in __new__
    raise ValueError("The equivalent system will have zero denominator.")
ValueError: The equivalent system will have zero denominator.
```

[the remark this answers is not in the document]
akshanshbhatt: > (Marked unresolved!)

Thank for pointing this out. I almost forgot this one. Fixed now - was a minor sign error in the conditional.

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1905,399 @@ def doit(self, **kwargs):
...
+        _expand_num_den = lambda _num, _den: Mul(_num.expand(),
+            Pow(_den.expand(), -1, evaluate=False))
+
+        if expand:
+            _mat = _mat.applyfunc(lambda a: _expand_num_den(*a.as_numer_denom()))
+
+        return _to_TFM(_mat, self.var)
+
+    def _eval_rewrite_as_TransferFunctionMatrix(self, plant, feedback_controller, ftype, **kwargs):
+        return self.doit(cancel=False, expand=False)
```
namannimmo10, line 2297:
```suggestion
        return self.doit()
```

sympy/physics/control/tests/test_lti.py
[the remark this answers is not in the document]
namannimmo10: > I guess this should work!
> 
> ```python
> >>> F = Feedback(TransferFunction(1, 1, s), TransferFunction(1, 1, s)) 
> >>> F.doit()
> TransferFunction(1, 2, s)
> >>>
> ```

Can you test this?  Unit tests should cover all corner cases.

sympy/printing/latex.py
```diff
@@ -2461,35 +2461,46 @@ def _print_MIMOParallel(self, expr):
...
+        num, tf = expr.plant, TransferFunction(1, 1, expr.var)
         num_arg_list = list(num.args) if isinstance(num, Series) else [num]
-        den_arg_list = list(expr.den.args) if isinstance(expr.den, Series) else [expr.den]
-
-        if isinstance(num, Series) and isinstance(expr.den, Series):
-            den = Parallel(tf, Series(*num_arg_list, *den_arg_list))
-        elif isinstance(num, Series) and isinstance(expr.den, TransferFunction):
-            if expr.den == tf:
-                den = Parallel(tf, Series(*num_arg_list))
+        den_arg_list = list(expr.feedback_controller.args) if isinstance(expr.feedback_controller, Series) else [expr.feedback_controller]
```
namannimmo10, line 2468: This line is too long.

sympy/printing/pretty/pretty.py
```diff
@@ -1018,33 +1018,58 @@ def _print_MIMOParallel(self, expr):
...
+        from sympy.physics.control import MIMOSeries, TransferFunctionMatrix
+
+        inv_mat = self._print(MIMOSeries(expr.feedback_controller, expr.plant))
+        plant = self._print(expr.plant)
+        _feedback = prettyForm(*stringPict.next(inv_mat))
+        _feedback = prettyForm(*stringPict.right("I + ", _feedback)) if expr.ftype == -1 \
+            else prettyForm(*stringPict.right("I - ", _feedback))
+        _feedback = prettyForm(*stringPict.parens(_feedback))
+        _feedback.baseline = 0
+        # _feedback = _feedback**(prettyForm("-1"))
```
namannimmo10, line 1065:
```suggestion
```

sympy/printing/pretty/pretty.py
```diff
@@ -1018,33 +1018,58 @@ def _print_MIMOParallel(self, expr):
...
 
     def _print_Feedback(self, expr):
-        from sympy.physics.control import TransferFunction, Parallel, Series
+        from sympy.physics.control import TransferFunction, Series
 
-        num, tf = expr.num, TransferFunction(1, 1, expr.num.var)
+        num, tf = expr.plant, TransferFunction(1, 1, expr.var)
         num_arg_list = list(num.args) if isinstance(num, Series) else [num]
-        den_arg_list = list(expr.den.args) if isinstance(expr.den, Series) else [expr.den]
+        den_arg_list = list(expr.feedback_controller.args) if isinstance(expr.feedback_controller, Series) else [expr.feedback_controller]
```
namannimmo10, line 1025: Line too long.

sympy/physics/control/tests/test_lti.py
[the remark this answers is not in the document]
akshanshbhatt: > Can you test this? Unit tests should cover all corner cases.

Added this test in test_lti.py

sympy/physics/control/lti.py
```diff
@@ -1719,37 +1745,44 @@ class Feedback(Basic):
...
+            feedback_controller = TransferFunction(1, 1, plant.var)
+
+        if not (isinstance(plant, (TransferFunction, Series))
+            and isinstance(feedback_controller, (TransferFunction, Series))):
+            raise TypeError("Unsupported type for `plant` or `feedback_controller` of Feedback.")
+
+        if ftype not in [-1, 1]:
+            raise ValueError("Unsupported type for feedback. ")
+
+        if Mul(plant.to_expr(), feedback_controller.to_expr(), evaluate=True) == ftype:
```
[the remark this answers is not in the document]
akshanshbhatt: > This needs to be tested for both positive or negative feedback.

```py
if Mul(plant.to_expr(), feedback_controller.to_expr(), evaluate=True) == ftype:
```
I assume you are talking about this line?
It already covers both cases.

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1905,399 @@ def doit(self, **kwargs):
...
+    type : int, optional
+        The type of closed-loop MIMO feedback. Can either be ``1``
+        (for positive feedback) or ``-1`` (for negative feedback).
+        Default value is `-1`.
+
+    Raises
+    ======
+
+    ValueError
+        When ``plant`` is equal to ``feedback_controller`` or when they are not using the
```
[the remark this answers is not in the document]
akshanshbhatt: > Just like in the Feedback class, these can be equal, no?

Yes.

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1910,400 @@ def doit(self, **kwargs):
...
+
+        _num, _den = _tf.num, _tf.den
+
+        if cancel:
+            _num, _den = Mul(_num, 1/_den, evaluate=False).cancel(expand=False).as_numer_denom()
+
+        if expand:
+            _num, _den = (_num.expand(), _den.expand())
+
+        return TransferFunction(_num, _den, self.var)
```
namannimmo10, line 1940:
```suggestion

        resultant_tf = TransferFunction(F_n.num * F_d.den, F_n.den * F_d.num, F_n.var)

        if cancel:
            resultant_tf = resultant_tf.simplify()
        if expand:
            resultant_tf = resultant_tf.expand()

        return resultant_tf
```

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1910,400 @@ def doit(self, **kwargs):
...
+    @property
+    def sign(self):
+        r"""
+        Returns the type of closed-loop MIMO Feedback. ``1``
+        for Positive and ``-1`` for Negative.
+        """
+        return self._sign
+
+    @property
+    def sensitivity(self):
```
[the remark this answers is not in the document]
akshanshbhatt: Sensitivity can be thought of as a property of closed-loop feedback systems. It should be helpful later.
https://www.mathworks.com/help/robust/ref/lti.loopsens.html

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1910,399 @@ def doit(self, **kwargs):
...
+        """
+        _plant_mat = self.plant.doit()._expr_mat
+        _controller_mat = self.feedback_controller.doit()._expr_mat
+
+        return (eye(self.plant.num_inputs) - \
+            self.sign*_plant_mat*_controller_mat).inv()
+
+    def doit(self, cancel=True, expand=False, **kwargs):
+        r"""
+        Returns the resultant closed loop transfer function matrix obtained by the
```
namannimmo10, line 2222:
```suggestion
        Returns the resultant transfer function matrix obtained by the
```

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1910,399 @@ def doit(self, **kwargs):
...
+            _mat = _mat.applyfunc(lambda a: _cancel_num_den(*a.as_numer_denom()))
+
+        _expand_num_den = lambda _num, _den: Mul(_num.expand(),
+            Pow(_den.expand(), -1, evaluate=False))
+
+        if expand:
+            _mat = _mat.applyfunc(lambda a: _expand_num_den(*a.as_numer_denom()))
 
-        return TransferFunction(F_n.num*F_d.den, F_n.den*F_d.num, F_n.var)
+        return _to_TFM(_mat, self.var)
```
namannimmo10, line 2299: More like,
```python
        _mat = self.sensitivity * self.plant.doit()._expr_mat
        _resultant_tfm = _to_TFM(_mat, self.var)

        if cancel:
            _resultant_tfm = _resultant_tfm.simplify()
        if expand:
            _resultant_tfm = _resultant_tfm.expand()

        return _resultant_tfm
```

sympy/physics/control/tests/test_lti.py
```diff
@@ -950,6 +952,113 @@ def test_Feedback_functions():
...
+    assert f1.sign == -1
+    assert -(-f1) == f1
+
+    f2 = MIMOFeedback(tfm_2, tfm_1, 1)
+    assert f2.args == (tfm_2, tfm_1, 1)
+    assert f2.plant == tfm_2
+    assert f2.feedback_controller == tfm_1
+    assert f2.var == s
+    assert f2.sign == 1
+
```
namannimmo10, line 979: Add this one test:
```
>>> mat = Matrix([[1, 1/s], [0, 1]])
>>> plant = controller = TransferFunctionMatrix.from_Matrix(mat, s)
>>> MIMOFeedback(plant, controller)
```

sympy/physics/control/lti.py
```diff
@@ -1823,10 +1860,40 @@ def var(self):
...
+        >>> from sympy.physics.control.lti import TransferFunction, Feedback
+        >>> C = TransferFunction(5*p + 10, p + 10, p)
+        >>> P = TransferFunction(1 - p, p + 2, p)
+        >>> F_1 = Feedback(P, C)
+        >>> F_1.sensitivity
+        1/((1 - p)*(5*p + 10)/((p + 2)*(p + 10)) + 1)
+
+        """
+
+        return 1/(1 - self.sign*self.plant.to_expr()*self.feedback_controller.to_expr())
```
namannimmo10, line 1891: Add test for this method?

namannimmo10:
@ilayn: PTAL.

sympy/physics/control/lti.py
```diff
@@ -1719,37 +1749,45 @@ class Feedback(Basic):
...
-    def __new__(cls, num, den):
-        if not (isinstance(num, (TransferFunction, Series))
-            and isinstance(den, (TransferFunction, Series))):
-            raise TypeError("Unsupported type for numerator or denominator of Feedback.")
-
-        if num == den:
-            raise ValueError("The numerator cannot be equal to the denominator.")
-        if not num.var == den.var:
-            raise ValueError("Both numerator and denominator should be using the"
+    def __new__(cls, plant, feedback_controller=None, sign=-1):
```
namannimmo10, line 1762:
```suggestion
    def __new__(cls, plant, controller=None, sign=-1):
```

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1910,395 @@ def doit(self, **kwargs):
...
+    Feedback, MIMOSeries, MIMOParallel
+
+    """
+    def __new__(cls, plant, feedback_controller, sign=-1):
+        if not (isinstance(plant, (TransferFunctionMatrix, MIMOSeries))
+            and isinstance(feedback_controller, (TransferFunctionMatrix, MIMOSeries))):
+            raise TypeError("Unsupported type for `plant` or `feedback_controller` of MIMO Feedback.")
+
+        if not (plant.num_inputs == plant.num_inputs == feedback_controller.num_inputs == \
+                feedback_controller.num_inputs):
```
namannimmo10, line 2048:
```suggestion
        if plant.shape != controller.shape or plant.num_outputs != plant.num_inputs:
```

namannimmo10:
This looks good overall!  I will take a look at the docstrings when I have time ... Meanwhile, let's wait for someone else's review.
Since there aren't any major changes required, I believe this can go in asap.

sympy/physics/control/lti.py
```diff
@@ -1719,37 +1749,45 @@ class Feedback(Basic):
...
+
+        if sign not in [-1, 1]:
+            raise ValueError("Unsupported type for feedback. `sign` arg should "
+            "either be 1 (positive feedback loop) or -1 (negative feedback loop).")
+
+        if Mul(plant.to_expr(), controller.to_expr()) == sign:
+            raise ValueError("The equivalent system will have zero denominator.")
+        if plant.var != controller.var:
+            raise ValueError("Both `plant` and `controller` should be using the"
                 " same complex variable.")
```
namannimmo10, line 1778: I removed these lines of code and all tests passed. Can you test this properly?

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1910,394 @@ def doit(self, **kwargs):
...
+            and isinstance(controller, (TransferFunctionMatrix, MIMOSeries))):
+            raise TypeError("Unsupported type for `plant` or `controller` of MIMO Feedback.")
+
+        if plant.shape != controller.shape or plant.num_outputs != plant.num_inputs:
+            raise ValueError("`plant` and `controller` must be square matrices of"
+                "the same shape.")
+
+        if sign not in [-1, 1]:
+            raise ValueError("Unsupported type for feedback. `sign` arg should "
+            "either be 1 (positive feedback loop) or -1 (negative feedback loop).")
```
namannimmo10, line 2053:
```suggestion
        if sign not in [-1, 1]:
            raise ValueError("Unsupported type for feedback. `sign` arg should "
                "either be 1 (positive feedback loop) or -1 (negative feedback loop).")
```

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1910,394 @@ def doit(self, **kwargs):
...
+
+    """
+    def __new__(cls, plant, controller, sign=-1):
+        if not (isinstance(plant, (TransferFunctionMatrix, MIMOSeries))
+            and isinstance(controller, (TransferFunctionMatrix, MIMOSeries))):
+            raise TypeError("Unsupported type for `plant` or `controller` of MIMO Feedback.")
+
+        if plant.shape != controller.shape or plant.num_outputs != plant.num_inputs:
+            raise ValueError("`plant` and `controller` must be square matrices of"
+                "the same shape.")
```
namannimmo10, line 2049: Same comment as above; not tested properly!

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1910,394 @@ def doit(self, **kwargs):
...
+        for Positive and ``-1`` for Negative.
+        """
+        return self._sign
+
+    @property
+    def sensitivity(self):
+        r"""
+        Returns the sensitivity function matrix of the feedback loop.
+        Sensitivity of a closed-loop system is the ratio of change
+        in the open loop gain to the change in the closed loop gain.
```
namannimmo10, line 2172: It seems like you are extending the concepts of sensitivity in SISO systems to MIMO systems -- and this can be problematic. There are many issues in designing this using SISO techniques. I would not have a separate property in `MIMOFeedback`.

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1910,394 @@ def doit(self, **kwargs):
...
+        Default value is `-1`.
+
+    Raises
+    ======
+
+    ValueError
+        When ``plant`` and ``controller`` are not using the
+        same complex variable of the Laplace transform.
+
+        When shapes of ``plant`` and ``controller`` are not the same.
```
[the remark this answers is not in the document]
namannimmo10: I don't know why `plant` and `controller` have to be square matrices. 

Say the plant has _m_ inputs and _n_ outputs; hence a matrix with (n, m) shape. Now the controller can be an (m, n) matrix.

sympy/physics/control/lti.py
```diff
@@ -1823,10 +1860,40 @@ def var(self):
...
+        return self._sign
+
+    @property
+    def sensitivity(self):
         """
-        Returns the resultant closed loop transfer function obtained by the
-        negative feedback interconnection.
+        Returns the sensitivity function of the feedback loop. Sensitivity
+        of a closed-loop system is the ratio of change in the open
+        loop gain to the change in the closed loop gain.
```
namannimmo10, line 1876:
```suggestion
        Returns the sensitivity function of the feedback loop.
        
        Sensitivity of a closed-loop system is the ratio of change in the open
        loop gain to the change in the closed loop gain.
```

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,61 @@ def __neg__(self):
...
+class Feedback(SISOLinearTimeInvariant):
+    r"""
+    A class for representing closed-loop feedback interconnection between two
+    SISO input/output systems. The first argument, ``plant``, is the
+    primary plant of the closed-loop system or in simple words, the transfer
+    function representing the system/process to be controlled. Most of the time, a controller system is also
+    placed in series with the plant. This controller is known as the **plant controller** as it controls
+    the plant by varying the input that goes into plant. The second argument, ``controller``,
+    as the name suggests, controls the fed back signal to the ``plant``. It is the equivalent transfer
+    function of the feedback controling system which is generally a sensor system that constantly
```
namannimmo10, line 1668:
```suggestion
    function of the feedback controlling system which is generally a sensor system that constantly
```

sympy/physics/control/lti.py
namannimmo10: @ilayn,  what's your take on it?

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,61 @@ def __neg__(self):
...
-    """
-    A class for representing negative feedback interconnection between two
-    input/output systems. The first argument, ``num``, is called as the
-    primary plant or the numerator, and the second argument, ``den``, is
-    called as the feedback plant (which is often a feedback controller) or
-    the denominator. Both ``num`` and ``den`` can either be ``Series`` or
+class Feedback(SISOLinearTimeInvariant):
+    r"""
+    A class for representing closed-loop feedback interconnection between two
+    SISO input/output systems. The first argument, ``plant``, is the
```
ilayn, line 1662: This is typically true however, in the block diagram algebra the first item doesn't have to be the plant. Nor the second should be controller. You could very well composing different parts of a dynamical system. Hence this might confuse users. sys1, sys2 terminology is fine I think. 

Besides in the SISO case, the order won't matter since closed loop is GK/(1+GK) and G and K commutes.

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,61 @@ def __neg__(self):
...
+    primary plant of the closed-loop system or in simple words, the transfer
```
ilayn, line 1663: Transfer function is not mandatory can be any dynamical model.

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,61 @@ def __neg__(self):
...
+    function representing the system/process to be controlled. Most of the time, a controller system is also
+    placed in series with the plant. This controller is known as the **plant controller** as it controls
```
ilayn, line 1665: Do you need all this explanation here? Feels a bit more like a textbook rather than a docstring.

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,61 @@ def __neg__(self):
...
+    the plant by varying the input that goes into plant. The second argument, ``controller``,
+    as the name suggests, controls the fed back signal to the ``plant``. It is the equivalent transfer
+    function of the feedback controling system which is generally a sensor system that constantly
+    takes in the output signal produced and feeds it back to the primary plant (or a plant-controller
+    if one is connected in series with the plant). Note that the feedback controller is not same as the
+    plant controller.
+
+    Generally, in control theory, we want the output signal to follow the input signal or in
```
ilayn, line 1673: Just mention that you can change the feedback sign leading to either direct sum or a negated sum. This paragraph is a bit wordy.

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,61 @@ def __neg__(self):
...
 
     Parameters
     ==========
 
-    num : Series, TransferFunction
-        The primary plant.
-    den : Series, TransferFunction
-        The feedback plant (often a feedback controller).
+    plant : Series, TransferFunction
+        The primary plant. Also commonly known as the open
```
ilayn, line 1690: Primary is misplaced here. plant has no significance. Also open-loop gain is `plant*controller` shown with `L` in many textbooks. Hence this needs correction.

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,61 @@ def __neg__(self):
...
+        loop gain / open-loop tranfer function.
+    controller : Series, TransferFunction, optional
+        The feedback plant (often a feedback controller). Also
```
ilayn, line 1693: The model sitting on the feedback path.

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,61 @@ def __neg__(self):
...
+        commonly known as the feedback factor.
```
ilayn, line 1694: Never heard of a `feedback factor`

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,61 @@ def __neg__(self):
...
+        If not specified explicitly, the controller is
+        assumed to be unit (1) transfer function.
```
ilayn, line 1696: assumed to be 1.0 since this is SISO

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,61 @@ def __neg__(self):
...
+    sign : int, optional
```
ilayn, line 1697: you want int but you have only two options maybe convert this to a bool with keyword name `feedback_sign` or something like that?

sympy/physics/control/lti.py
```diff
@@ -1692,17 +1722,17 @@ class Feedback(Basic):
...
-    Feedback(TransferFunction(3*s**2 + 7*s - 3, s**2 - 4*s + 2, s), TransferFunction(5*s - 10, s + 7, s))
+    Feedback(TransferFunction(3*s**2 + 7*s - 3, s**2 - 4*s + 2, s), TransferFunction(5*s - 10, s + 7, s), -1)
     >>> F1.var
     s
     >>> F1.args
-    (TransferFunction(3*s**2 + 7*s - 3, s**2 - 4*s + 2, s), TransferFunction(5*s - 10, s + 7, s))
+    (TransferFunction(3*s**2 + 7*s - 3, s**2 - 4*s + 2, s), TransferFunction(5*s - 10, s + 7, s), -1)
 
-    You can get the primary and the feedback plant using ``.num`` and ``.den`` respectively.
+    You can get the primary and the feedback controller using ``.plant`` and ``.controller`` respectively.
```
ilayn, line 1731: Again I don't know what primary controller means but now it is pointing to plant

ilayn (commented):
I went through the SISO case but in my humble opinion, this is a bit too specific to a particular control problem where sys1 is sitting on the forward path and sys2 on the feedback path and sys1 is the plant to be controlled and sys2 is the controller as the compansator as it is directly operating on the output. 

However the use case of this function is much broader and here it is only addressing one. I would avoid the parts that can be read from the textbook and leave a few generic comments about which argument does what.

sympy/physics/control/lti.py
akshanshbhatt: I understand what you are trying to say. I initially chose the sys1-sys2 naming convention since it was used in both MATLAB and python-control. I went ahead with the `plant` and `feedback_controller` convention as the goal was to make it beginner-friendly. 
`plant` and `feedback_controller` made more sense than `plant` and `controller` because the controller doesn't need to be always in the feedback path. Some controller systems can also be placed in the feedforward path. `feedback_controller` specified this fact.

sympy/physics/control/lti.py
akshanshbhatt: I can add this explanation in an explanation section as I did in `TransferFunction`'s doc.
@namannimmo10?

sympy/physics/control/lti.py
akshanshbhatt: https://www.electronics-tutorials.ws/systems/feedback-systems.html

image

I referred to this block diagram.

sympy/physics/control/lti.py
akshanshbhatt: > you want int but you have only two options maybe convert this to a bool with keyword name `feedback_sign` or something like that?

I tried that but that didn't work, some tests failed. `Feedback` has `Basic` as its superclass, therefore it must satisfy the `obj.func(*obj.args)` criteria. Only Basic instances can be passed as args.

https://github.com/sympy/sympy/blob/19c5014a9f324d59e2fddb44f239ccf6c8f7b1ba/sympy/core/tests/test_args.py#L90

The same reason why I have to sympify the `sign` arg.

https://github.com/sympy/sympy/blob/4196e76bffb061c60a498ad6a3f300cf8626c95b/sympy/physics/control/lti.py#L1779

sympy/physics/control/lti.py
ilayn: Factor is basically a scalar, here they probably don't consider an LTI system apparently since both the reference and the output is of the same kind.

sympy/physics/control/lti.py
akshanshbhatt: If `plant`-`controller` is not a general terminology, then maybe we can go with the `open_loop_gain`-`feedback_gain` or `sys1`-`sys2`. Whichever suits better.

namannimmo10: I think we can go with sys1-sys2 terminology.

ilayn: G-H is also fine if we do a very simple ASCII-art of the loop

akshanshbhatt: > G-H is also fine if we do a very simple ASCII-art of the loop

Slightly off-topic but always wanted to know whether there is some tool for creating block diagram using ASCII art.

sympy/physics/control/lti.py
```diff
@@ -1719,37 +1749,47 @@ class Feedback(Basic):
...
-        obj = super().__new__(cls, num, den)
-        obj._num = num
-        obj._den = den
-        obj._var = num.var
+
+        obj = super().__new__(cls, plant, controller, _sympify(sign))
+        obj._plant = plant
+        obj._controller = controller
+        obj._var = plant.var
+        obj._sign = sign
```
namannimmo10, line 1785:
```suggestion
```

sympy/physics/control/lti.py
```diff
@@ -1759,23 +1799,22 @@ def num(self):
...
         >>> C = TransferFunction(5*p + 10, p + 10, p)
         >>> P = TransferFunction(1 - s, p + 2, p)
         >>> F2 = Feedback(TransferFunction(1, 1, p), G*C*P)
-        >>> F2.num
+        >>> F2.plant
         TransferFunction(1, 1, p)
 
         """
-        return self._num
+        return self._plant
```
namannimmo10, line 1812:
```suggestion
        return self.args[0]
```

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1914,399 @@ def doit(self, **kwargs):
...
+
+        if plant.var != controller.var:
+            raise ValueError("Both `plant` and `controller` should be using the"
+                " same complex variable.")
+
+        obj = super().__new__(cls, plant, controller, _sympify(sign))
+        obj._plant = plant
+        obj._controller = controller
+        obj._var = plant.var
+        obj._sign = sign
```
namannimmo10, line 2072:
```suggestion
```

namannimmo10:
@akshanshbhatt, any updates on this one?  I think this is ready once you address these comments.

akshanshbhatt:
> I think this is ready once you address these comments.

Addressed.

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,60 @@ def __neg__(self):
...
+        Also commonly known as the open
+        loop gain.
+    sys2 : Series, TransferFunction, optional
+        The feedback path system (often a feedback controller).
+        It is the model sitting on the feedback path.
+
+        If not specified explicitly, the sys2 is
+        assumed to be unit (1.0) transfer function.
+    sign : int, optional
+        The type of closed-loop feedback. Can either be ``1``
```
namannimmo10, line 1697:
```suggestion
        The sign of feedback. Can either be ``1``
```

sympy/physics/control/lti.py
```diff
@@ -1656,31 +1656,60 @@ def __neg__(self):
...
     ==========
 
-    num : Series, TransferFunction
-        The primary plant.
-    den : Series, TransferFunction
-        The feedback plant (often a feedback controller).
+    sys1 : Series, TransferFunction
+        The feedforward path system.
+        Also commonly known as the open
+        loop gain.
```
namannimmo10, line 1689:
```suggestion
```
See https://github.com/sympy/sympy/pull/21833#discussion_r688267765.

sympy/physics/control/lti.py
```diff
@@ -1719,37 +1748,41 @@ class Feedback(Basic):
...
 
-        return obj
+        return super().__new__(cls, sys1, sys2, _sympify(sign))
 
     @property
-    def num(self):
+    def sys1(self):
         """
-        Returns the primary plant of the negative feedback closed loop.
+        Returns the feedforward system of the closed-loop feedback.
```
namannimmo10, line 1785:
```suggestion
        Returns the feedforward system of the feedback interconnection.
```
I'd refrain from using "closed-loop" in all the docstrings.

sympy/physics/control/lti.py
```diff
@@ -1759,23 +1792,22 @@ def num(self):
...
-        return self._num
+        return self.args[0]
 
     @property
-    def den(self):
+    def sys2(self):
         """
-        Returns the feedback plant (often a feedback controller) of the
-        negative feedback closed loop.
+        Returns the feedback controller of the closed-loop feedback.
```
namannimmo10, line 1810: Same comment as in `sys1` property.

sympy/physics/control/lti.py
```diff
@@ -1843,19 +1907,393 @@ def doit(self, **kwargs):
...
+        r"""
+        Returns the type of feedback interconnection of two models. ``1``
+        for Positive and ``-1`` for Negative.
+        """
+        return self.args[2]
+
+    @property
+    def sensitivity(self):
+        r"""
+        Returns the sensitivity function matrix of the feedback loop.
```
namannimmo10, line 2165: There are two types of sensitivities, right?
```
S = (I + P*C)^-1
T = (I + P*C)^-1 * (P*C)
S + T = I
```
Check out [this video](https://www.youtube.com/watch?v=hTu36q5yx20).
You should also mention here:
This would not return the complementary sensitivity.

akshanshbhatt: Added a note.

PR #21848, opened 2021-08-10
integrals - fixes issue#21711 and now manual integrate is tried if meijerg returns unevaluated integral

### LINKED ISSUE #21711: odd result for  integrate(sqrt(1 - (x-1)*(x-1)), (x, 0, 1))
```
In [12]: integrate(sqrt(1 - (x-1)*(x-1)), (x, 0, 1))
Out[12]: Integral(Piecewise((I*(1 - x)*(x - 1)**3/(2*((x - 1)**2 - 1)**(3/2)) - I*(1 - x)*(x - 1)/(2*((x - 1)**2 - 1)**(3/2)) + 3*I*(x - 1)**2/(2*sqrt((x - 1)**2 - 1)) - I/sqrt((x - 1)**2 - 1), (x - 1)**2 > 1), ((1 - x)*(x - 1)/(2*sqrt(1 - (x - 1)**2)) + sqrt(1 - (x - 1)**2)/2 + 1/(2*sqrt(1 - (x - 1)**2)), True)), (x, 0, 1))

In [13]: integrate(sqrt(1 - (x)*(x)), (x, 0, 1))
Out[13]: pi/4
```

maxima:
```
(%i4) integrate(sqrt(1 - x^2), x, 0, 1);
                                     %pi
(%o4)                                 ---
                                       4


(%i5) integrate(sqrt(1 - (x-1)^2), x, 0, 1)
;
                                      %pi
(%o5)                                 ---
                                       4
```

praneethratna:
Fixes #21711 by trying for manual integration when meijerg returns an unevaluated integral

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/integrals/integrals.py
+++ b/sympy/integrals/integrals.py
@@ -679,28 +679,28 @@ class Integral(AddWithLimits):
                                     args.append(g)
                             return Mul(*args)
 
-                        integrals, others, piecewises = [], [], []
-                        for f in Add.make_args(antideriv):
-                            if any(is_indef_int(g, x)
-                                   for g in Mul.make_args(f)):
-                                integrals.append(f)
-                            elif any(isinstance(g, Piecewise)
-                                     for g in Mul.make_args(f)):
-                                piecewises.append(piecewise_fold(f))
-                            else:
-                                others.append(f)
-                        uneval = Add(*[eval_factored(f, x, a, b)
-                                       for f in integrals])
+                        def sep_values(antideriv, x):
+                            integrals, others, piecewises = [], [], []
+                            for f in Add.make_args(antideriv):
+                                if any(is_indef_int(g, x)
+                                      for g in Mul.make_args(f)):
+                                    integrals.append(f)
+                                elif any(isinstance(g, Piecewise)
+                                        for g in Mul.make_args(f)):
+                                    piecewises.append(piecewise_fold(f))
+                                else:
+                                    others.append(f)
+                            uneval = Add(*[eval_factored(f, x, a, b)
+                                          for f in integrals])
+                            return uneval, piecewises, others
                         try:
+                            uneval, piecewises, others = sep_values(antideriv, x)
                             evalued = Add(*others)._eval_interval(x, a, b)
                             evalued_pw = piecewise_fold(Add(*piecewises))._eval_interval(x, a, b)
                             if isinstance(evalued_pw, Integral):
-                                antideriv = self._eval_integral(function, xab[0], manual=True)
-                                piecewises = []
-                                for f in Add.make_args(antideriv):
-                                    if any(isinstance(g, Piecewise)
-                                           for g in Mul.make_args(f)):
-                                               piecewises.append(piecewise_fold(f))
+                                antideriv = manualintegrate(function, xab[0])
+                                uneval, piecewises, others = sep_values(antideriv, x)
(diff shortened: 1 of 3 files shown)


praneethratna:
@oscarbenjamin I'm not sure about this particular fix but it is kinda passing all the tests and is working perfectly fine as far as i have tested. Is there any other easier fix to #21711  that can be more reasonable?

oscarbenjamin:
I don't know the integration code so well myself. The control flow seems to be very complicated soI don't know what's the best way to fall back from meijerg to manualintegrate. Normally that happens automatically if meijerg fails.

sympy/integrals/integrals.py
```diff
@@ -694,6 +694,14 @@ def eval_factored(f, x, a, b):
                         try:
                             evalued = Add(*others)._eval_interval(x, a, b)
                             evalued_pw = piecewise_fold(Add(*piecewises))._eval_interval(x, a, b)
+                            if isinstance(evalued_pw, Integral):
+                                antideriv = self._eval_integral(function, xab[0], manual=True)
```
oscargus, line 698: Here you try out the manual integrator.

sympy/integrals/integrals.py
```diff
@@ -694,6 +694,14 @@ def eval_factored(f, x, a, b):
...
+                                piecewises = []
+                                for f in Add.make_args(antideriv):
+                                    if any(isinstance(g, Piecewise)
+                                           for g in Mul.make_args(f)):
+                                               piecewises.append(piecewise_fold(f))
```
oscargus, line 703: Here you get the pieces from the Piecewise?

sympy/integrals/integrals.py
```diff
@@ -694,6 +694,14 @@ def eval_factored(f, x, a, b):
...
+                                evalued_pw = piecewise_fold(Add(*piecewises))._eval_interval(x, a, b)
```
oscargus, line 704: Here you recreate it for those that were Piecewise. What I do not understand is how do you know that `antideriv` from the manual integrator will end up with the same `Piecewise` sections as the original integral. That is, how can you replace evaluated_pw with this line?

It does indeed seem to work for the example, but the logic is not obvious to me.

sympy/integrals/integrals.py
praneethratna: yes but i don't know whether this particular method is reasonable or not since I've reapeated some of the code again 
https://github.com/sympy/sympy/blob/c39a6848728a3c84a92f348a4fc5f1a4f4ad2b84/sympy/integrals/integrals.py#L682-L705

sympy/integrals/integrals.py
praneethratna: I mean I missed some part of the code here since i was not sure of this solution there will be needing of a isinstance check for type indef integrals and rest which are both not piecewise and indef integrals but that code is already above this so basically all i did was repeating some part of code but can we do something like making the below part of code a function and call it twice?https://github.com/sympy/sympy/blob/c39a6848728a3c84a92f348a4fc5f1a4f4ad2b84/sympy/integrals/integrals.py#L682-L693

sympy/integrals/integrals.py
praneethratna: Yes But i'm not sure of this approach

praneethratna:
@oscargus please review , I've tried an alternative fix

praneethratna:
@oscargus please review , I've tried an alternative fix

Issue #14700, opened 2018-05-05
boolalg.SOPform() does not give the simplest form.


aiifabbf:
Hi, developers of Sympy.

@mpraiser and I were using `boolalg.SOPform()` to get the smallest sum of products. But it does not seem to give the *smallest* sum of products.

```python
In [136]: v
Out[136]: [A, B, C, D, E, F, G, H]

In [137]: q
Out[137]: 
(B ∧ D ∧ H ∧ ¬F) ∨ (B ∧ H ∧ ¬C ∧ ¬D) ∨ (B ∧ H ∧ ¬C ∧ ¬F) ∨ (B ∧ H ∧ ¬D ∧ ¬G) ∨
 (B ∧ H ∧ ¬F ∧ ¬G) ∨ (C ∧ G ∧ ¬B ∧ ¬D) ∨ (C ∧ G ∧ ¬D ∧ ¬H) ∨ (C ∧ G ∧ ¬F ∧ ¬H)
 ∨ (D ∧ F ∧ H ∧ ¬B) ∨ (D ∧ F ∧ ¬G ∧ ¬H) ∨ (B ∧ D ∧ F ∧ ¬C ∧ ¬H) ∨ (D ∧ E ∧ F ∧
 ¬B ∧ ¬C) ∨ (D ∧ F ∧ ¬A ∧ ¬B ∧ ¬C) ∨ (D ∧ F ∧ ¬A ∧ ¬C ∧ ¬H) ∨ (A ∧ B ∧ D ∧ F ∧
 ¬E ∧ ¬H)

In [138]: qq
Out[138]: 
(B ∧ D ∧ H ∧ ¬F) ∨ (B ∧ H ∧ ¬C ∧ ¬D) ∨ (B ∧ H ∧ ¬D ∧ ¬G) ∨ (C ∧ G ∧ ¬B ∧ ¬D) ∨
 (C ∧ G ∧ ¬D ∧ ¬H) ∨ (C ∧ G ∧ ¬F ∧ ¬H) ∨ (D ∧ F ∧ H ∧ ¬B) ∨ (D ∧ F ∧ ¬G ∧ ¬H) 
∨ (B ∧ D ∧ F ∧ ¬C ∧ ¬H) ∨ (D ∧ E ∧ F ∧ ¬C ∧ ¬H) ∨ (D ∧ F ∧ ¬A ∧ ¬C ∧ ¬H) ∨ (A 
[...]

PR #21854, opened 2021-08-11
Fixes for logic simplification

### LINKED ISSUE #7950: Simplifying logic and equality constraints
``` python
>>> x = sympy.Symbol('x')
>>> sympy.simplify_logic(sympy.And(sympy.Eq(x, 1), sympy.Eq(x, 2)))
And(x == 1, x == 2)
```

Wouldn't one expect it to be just `False` since `x` cannot be simultaneously equal to two different values? If not, is there any way to simplify it to `False`, at least in the case when `x` is an integer?

Could be somewhat related: https://code.google.com/p/sympy/issues/detail?id=4009

### LINKED ISSUE #8373: Simplification of Relational and BooleanFunction
Consider the following:

``` python
In [6]: x = Symbol('x', real=True)

In [7]: simplify(Or(x<1, x>= 1))
Out[7]: Or(x < 1, x >= 1)
```

I'd expect the result to be `True` or somehow else to designate "all `x`".

Even if I force the symbol to be a positive integer, this does not simplify:

``` python
In [3]: r = Symbol('r', integer=True, positive=True)

In [4]: Or(r<1, r>= 1)
Out[4]: Or(r < 1, r >= 1)

In [5]: simplify(Or(r<1, r>= 1))
Out[5]: Or(r < 1, r >= 1)
```

which is a bit odd. All of this causes over-complicated outputs sometimes. For example, trying to sum the PMF of a negative binomial distribution with `p=1/2`, I get

``` python
In [3]: k = Symbol('k', integer=True, nonnegative=True)

In [4]: r = Symbol('r', integer=True, positive=True)

In [5]: Sum(binomial(k+r-1,k)*(S.Half)**r*S.Half**k, (k, 0, oo)).doit()
Out[5]: Piecewise((1, Or(r < 1, r >= 1)), (Sum(2**(-k)*2**(-r)*binomial(k + r - 1, k), (k, 0, oo)), True))
```

### LINKED ISSUE #18645: simplify And(Le|Eq(x, 3), Ge(x, 3)) to Eq(x, 3)
```
x = Symbol("x")
```
With
```
expr = And(Ge(x, 3), Le(x, 3))
```
Or:
```
expr = And(Eq(x, 3), Le(x, 3))
```
I would expect that `expr.simplify()` yields `Eq(x, 3)` in both cases.
Maybe this is related to #18642 or #18389 but I'm not sure.

### LINKED ISSUE #19114: `sympy.to_dnf(expr, simplify=True)` does not return the simplest form
Hi ! 

In python 3.6, sympy 1.4 does not return the minimal DNF of `(A & ~C) | (~A & ~B) | (B & C)` (which is already a min DNF in the sense it has been explicated in #13649). 

The following code returns `(A & B) | (B & C) | (A & ~C) | (C & ~A) | (~A & ~B) | (~B & ~C)`

```python
import sympy
expr = sympy.sympify("(A & ~C) | (~A & ~B) | (B & C)")
expr_minDNF = sympy.to_dnf(expr, simplify=True)
print(expr_minDNF)
```

Do you have any idea how this could be fixed? 

Also, are you aware of a check we could run so we don't have to run a simplification when the input expression is already minimal?

Thanks in advance. :)

### LINKED ISSUE #19822: And is not simplified
`And(Gt(n-2,1), Gt(n,1))` is not simplified to `And(Gt(n-2,1))`. Is there some function (or at least direction) to simplify such things?

### LINKED ISSUE #20870: Logic simplification not simplifying good enough
Just stumbled across this case which seems to trigger an issue in the logic simplification (not sure if it is solved by #20245, do not have time to test now, just wanted to post the case).

```
from sympy import *

a, b, c, d = symbols('a b c d')
bad = SOPform([a, b, c, d], [1, 2, 3, 4, 5, 6, 8, 9, 11, 12, 14, 15])
````

This gives `(b & ~d) | (d & ~b) | (a & b & c) | (a & c & d) | (a & ~b & ~c) | (a & ~c & ~d) | (b & ~a & ~c) | (c & ~a & ~b) | (c & ~a & ~d) | (d & ~a & ~c)` but e.g. the `(a & b & c)` and `(a & c & d)` terms are redundant, only one is needed. It should be enough with `(b & ~d) | (d & ~b)` and four of the remaining eight terms (pair-wise redundant).

### PR DESCRIPTION
**Pull request by oscargus**

Fixes #8373
Fixes #18645
Fixes #19114
Fixes #19822
Fixes #20870
Closes #7950 (not fixed here, but test added and it seems fixed elsewhere)
Added (XFAIL) tests for #18389 #18642

Fixed logic simplification so that the pattern based simplification is used even if there are no `Eq`-constrains involved (removed early exit)

Improved set cover algorithm which only used to work in case one term fully covered another term. Now, a greedy heuristic is used in that case.

#### Other comments
As enabling the pattern based simplification killed some other tests (as `zoo` is not comparable), I removed that the relational simplification replaces with expression including symbolic Max/Min. This is something that breaks later simplifications anyway and something I regret introducing.

So `And(x >= 1, x >= 2)` will be simplified to `x >= 2`, but `And(x >= a, x >= b)` will no longer be simplified to `And(x >= Max(a, b))`. As this part was quite often disabled earlier, I do not expect it to become a major issue. 

In the long term, one should have some function that goes from boolean relations to Min/Max and vice versa. Probably affects #17443 and #17330 in a positive way (this is where I started regretting adding the initial simplifications if I recall correctly).

#### Release Notes

* logic
  * Pattern based logic simplification is actually used when doing `simplify` with relations.
  * Improved logic simplification, dealing better with some symmetric cases and always producing consistent results.
  * The relational simplification no longer results in symbolic `Min`/`Max`.

### DIFF
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -629,10 +629,9 @@ class BooleanFunction(Application, Boolean):
                                 # will be replacementvalue
                                 return replacementvalue
                             # add replacement
-                            if (not isinstance(np, ITE) and np.count(Max) == 0
-                                and np.count(Min) == 0):
-                                # We only want to use ITE replacements if
-                                # they simplify to a relational
+                            if not isinstance(np, ITE) and not np.has(Min, Max):
+                                # We only want to use ITE and Min/Max
+                                # replacements if they simplify away
                                 costsaving = measure(oldexpr) - measure(np)
                                 if costsaving > 0:
                                     results.append((costsaving, (i, j, np)))
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -2310,16 +2310,14 @@ def _rem_redundancy(l1, terms):
 
     # Create dominating matrix
     dommatrix = [[0]*nl1 for n in range(nterms)]
+    colcount = [0]*nl1
+    rowcount = [0]*nterms
     for primei, prime in enumerate(l1):
         for termi, term in enumerate(terms):
             if _compare_term(term, prime):
                 dommatrix[termi][primei] = 1
-
-    # Non-dominated prime implicants, dominated to be removed
-    ndprimeimplicants = set(range(nl1))
-
-    # Non-dominated terms, dominated to be removed
-    ndterms = set(range(nterms))
+                colcount[primei] += 1
+                rowcount[termi] += 1
 
     # Keep track if anything changed
     anythingchanged = True
@@ -2327,63 +2325,75 @@ def _rem_redundancy(l1, terms):
     while anythingchanged:
         anythingchanged = False
 
-        # Make copy for iteration
-        oldndterms = ndterms.copy()
-        # Filter matrix to only get non-dominated items
-        filteredrows = [[dommatrix[rowi][i] for i in list(ndprimeimplicants)]
-                        for rowi in oldndterms]
-        for n, rowi in enumerate(oldndterms):
+        for rowi in range(nterms):
             # Still non-dominated?
-            if rowi in ndterms:
-                row = filteredrows[n]
-                for n2, row2i in enumerate(oldndterms):
+            if rowcount[rowi]:
+                row = dommatrix[rowi]
+                for row2i in range(nterms):
                     # Still non-dominated?
-                    if n != n2 and row2i in ndterms:
-                        if all(a >= b for (a, b) in zip(filteredrows[n2], row)):
+                    if rowi != row2i and rowcount[rowi] and (rowcount[rowi] <= rowcount[row2i]):
+                        row2 = dommatrix[row2i]
+                        if all(a >= b for (a, b) in zip(row2, row)):
                             # row2 dominating row, remove row2
-                            ndterms.remove(row2i)
+                            rowcount[row2i] = 0
                             anythingchanged = True
+                            for primei, prime in enumerate(row2):
+                                if prime:
+                                    # Make corresponding entry 0
+                                    dommatrix[row2i][primei] = 0
+                                    colcount[primei] -= 1
 
-        # Make copy for iteration
-        oldndprimeimplicants = ndprimeimplicants.copy()
-        # Filter matrix to only get non-dominated items
-        filteredcols = [[dommatrix[i][coli] for i in list(ndterms)]
-                        for coli in oldndprimeimplicants]
-        for n, coli in enumerate(oldndprimeimplicants):
+        colcache = dict()
+
+        for coli in range(nl1):
             # Still non-dominated?
-            if coli in ndprimeimplicants:
-                col = filteredcols[n]
-                for n2, col2i in enumerate(oldndprimeimplicants):
+            if colcount[coli]:
+                if coli in colcache:
+                    col = colcache[coli]
+                else:
+                    col = [dommatrix[i][coli] for i in range(nterms)]
+                    colcache[coli] = col
+                for col2i in range(nl1):
                     # Still non-dominated?
-                    if coli != col2i and col2i in ndprimeimplicants:
-                        if all(a >= b for (a, b) in zip(col, filteredcols[n2])):
+                    if coli != col2i and colcount[col2i] and (colcount[coli] >= colcount[col2i]):
+                        if col2i in colcache:
+                            col2 = colcache[col2i]
+                        else:
+                            col2 = [dommatrix[i][col2i] for i in range(nterms)]
+                            colcache[col2i] = col2
+                        if all(a >= b for (a, b) in zip(col, col2)):
                             # col dominating col2, remove col2
-                            ndprimeimplicants.remove(col2i)
+                            colcount[col2i] = 0
                             anythingchanged = True
+                            for termi, term in enumerate(col2):
+                                if term and dommatrix[termi][col2i]:
+                                    # Make corresponding entry 0
+                                    dommatrix[termi][col2i] = 0
+                                    rowcount[termi] -= 1
 
         if not anythingchanged:
             # Heuristically select the prime implicant covering most terms
             maxterms = 0
             bestcolidx = -1
-            bestcol = []
-            for n, coli in enumerate(oldndprimeimplicants):
-                s = sum(filteredcols[n])
+            for coli in range(nl1):
+                s = colcount[coli]
                 if s > maxterms:
                     bestcolidx = coli
                     maxterms = s
-                    bestcol = filteredcols[n]
 
             # In case we found a prime implicant covering at least two terms
             if bestcolidx != -1 and maxterms > 1:
                 for primei, prime in enumerate(l1):
                     if primei != bestcolidx:
-                        for termi, term in enumerate(bestcol):
+                        for termi, term in enumerate(colcache[bestcolidx]):
                             if term and dommatrix[termi][primei]:
                                 # Make corresponding entry 0
                                 dommatrix[termi][primei] = 0
                                 anythingchanged = True
+                                rowcount[termi] -= 1
+                                colcount[primei] -= 1
 
-    return [l1[i] for i in ndprimeimplicants]
+    return [l1[i] for i in range(nl1) if colcount[i]]
 
 
 def _input_to_binlist(inputlist, variables):
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -2358,6 +2358,28 @@ def _rem_redundancy(l1, terms):
                             ndprimeimplicants.remove(col2i)
                             anythingchanged = True
 
+        if not anythingchanged:
+            # Heuristically select the prime implicant covering most terms
+            maxterms = 0
+            bestcolidx = -1
+            bestcol = []
+            for n, coli in enumerate(oldndprimeimplicants):
+                s = sum(filteredcols[n])
+                if s > maxterms:
+                    bestcolidx = coli
+                    maxterms = s
+                    bestcol = filteredcols[n]
+
+            # In case we found a prime implicant covering at least two terms
+            if bestcolidx != -1 and maxterms > 1:
+                for primei, prime in enumerate(l1):
+                    if primei != bestcolidx:
+                        for termi, term in enumerate(bestcol):
+                            if term and dommatrix[termi][primei]:
+                                # Make corresponding entry 0
+                                dommatrix[termi][primei] = 0
+                                anythingchanged = True
+
     return [l1[i] for i in ndprimeimplicants]
 
 
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -743,10 +743,11 @@ class And(LatticeOp, BooleanFunction):
             return rv
         eqs, other = sift(Rel, lambda i: isinstance(i, Equality), binary=True)
 
-        measure, ratio = kwargs['measure'], kwargs['ratio']
-        reps = {}
-        sifted = {}
+        measure = kwargs['measure']
         if eqs:
+            ratio = kwargs['ratio']
+            reps = {}
+            sifted = {}
             # group by length of free symbols
             sifted = sift(ordered([
                 (i.free_symbols, i) for i in eqs]),
@@ -783,11 +784,11 @@ class And(LatticeOp, BooleanFunction):
                         f = e.free_symbols
                         resifted[len(f)].append((f, e))
                 sifted = resifted
-        for k in sifted:
-            eqs.extend([e for f, e in sifted[k]])
-        nonlineqs = [ei.subs(reps) for ei in nonlineqs]
-        other = [ei.subs(reps) for ei in other]
-        rv = rv.func(*([i.canonical for i in (eqs + nonlineqs + other)] + nonRel))
+            for k in sifted:
+                eqs.extend([e for f, e in sifted[k]])
+            nonlineqs = [ei.subs(reps) for ei in nonlineqs]
+            other = [ei.subs(reps) for ei in other]
+            rv = rv.func(*([i.canonical for i in (eqs + nonlineqs + other)] + nonRel))
         patterns = simplify_patterns_and()
         return self._apply_patternbased_simplification(rv, patterns,
                                                        measure, False)
@@ -2187,8 +2188,8 @@ def _convert_to_varsSOP(minterm, variables):
     Converts a term in the expansion of a function from binary to its
     variable form (for SOP).
     """
-    temp = [variables[n] if val == 1 else Not(variables[n])
-            for n, val in enumerate(minterm) if val != 3]
+    temp = (variables[n] if val == 1 else Not(variables[n])
+            for n, val in enumerate(minterm) if val != 3)
     return And(*temp)
 
 
@@ -2197,8 +2198,8 @@ def _convert_to_varsPOS(maxterm, variables):
     Converts a term in the expansion of a function from binary to its
     variable form (for POS).
     """
-    temp = [variables[n] if val == 0 else Not(variables[n])
-            for n, val in enumerate(maxterm) if val != 3]
+    temp = (variables[n] if val == 0 else Not(variables[n])
+            for n, val in enumerate(maxterm) if val != 3)
     return Or(*temp)
 
 
@@ -2251,11 +2252,9 @@ def _simplified_pairs(terms):
 
     # Count number of ones as _check_pair can only potentially match if there
     # is at most a difference of a single one
-    def ones_count(term):
-        return sum([1 for t in term if t == 1])
     termdict = defaultdict(list)
     for n, term in enumerate(terms):
-        ones = ones_count(term)
+        ones = sum([1 for t in term if t == 1])
         termdict[ones].append(n)
 
     variables = len(terms[0])
@@ -2279,8 +2278,7 @@ def _simplified_pairs(terms):
         simplified_terms = _simplified_pairs(simplified_terms)
 
     # Add remaining, non-simplified, terms
-    simplified_terms.extend(
-        [terms[i] for i in [_ for _ in todo if _ is not None]])
+    simplified_terms.extend([terms[i] for i in todo if i is not None])
     return simplified_terms
 
 
@@ -2333,7 +2331,7 @@ def _rem_redundancy(l1, terms):
                     # Still non-dominated?
                     if rowi != row2i and rowcount[rowi] and (rowcount[rowi] <= rowcount[row2i]):
                         row2 = dommatrix[row2i]
-                        if all(a >= b for (a, b) in zip(row2, row)):
+                        if all(row2[n] >= row[n] for n in range(nl1)):
                             # row2 dominating row, remove row2
                             rowcount[row2i] = 0
                             anythingchanged = True
@@ -2361,7 +2359,7 @@ def _rem_redundancy(l1, terms):
                         else:
                             col2 = [dommatrix[i][col2i] for i in range(nterms)]
                             colcache[col2i] = col2
-                        if all(a >= b for (a, b) in zip(col, col2)):
+                        if all(col[n] >= col2[n] for n in range(nterms)):
                             # col dominating col2, remove col2
                             colcount[col2i] = 0
                             anythingchanged = True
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by oscargus** [comment]
@oscarbenjamin There is a test that fails using optional dependencies (currently XFAILed). I have no idea why and which dependency might affect it. Any ideas? Anything being used to simplify dnf-forms now?

**Comment by oscargus** [comment]
Ahh, OK! I guess I need to sort at some stages then to get consistent results.

**Comment by oscargus** [comment]
The results should now be consistent. Also improved the run time a bit for `simplify_logic` by not testing for cnf or dnf forms unless required.

**Comment by oscargus** [comment]
For now, I just check that the resulting expression is equal and minimal. Not sure how to deal with it to get results that are consistent.

**Comment by oscarbenjamin** [comment]
> Is there are way to access the test server to see what is happening? If I recall correctly simply printing intermediate results doesn't work.

It's not hard to reproduce the setup outside of CI:
https://github.com/sympy/sympy/blob/12827227e1d71aef6967724df3f8dd019fc18a4e/.github/workflows/runtests.yml#L80-L99

**Comment by oscargus** [comment]
You need to try the penultimate commit for that... But you are probably correct in that it cannot produce consistent results (I think because all boolean functions use a set for the args, so you cannot even guarantee that the order you type it in is retained.)

**Comment by oscargus** [comment]
Is there anything here to be sorted out? I am quite convinced that the inconsistency is not caused by the current fixes.

**Comment by smichr** [comment]
> I still see nondeterministic results:

With this diff I get `B & C` always appearing first:

```diff
diff --git a/sympy/logic/boolalg.py b/sympy/logic/boolalg.py
index 5021354dd5..d59141528d 100644
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -2865,7 +2865,7 @@ def simplify_logic(expr, form=None, deep=True, force=False):
     if not force and len(variables) > 8:
         return expr
     # group into constants and variable values
-    c, v = sift(variables, lambda x: x in (True, False), binary=True)
+    c, v = sift(ordered(variables), lambda x: x in (True, False), binary=True)
     variables = c + v
     truthtable = []
     # standardize constants to be 1 or 0 in keeping with truthtable
```

oscargus:
Thanks @smichr ! I've added that one now.

Interesting that this has never manifested itself earlier. I didn't have to change any of the other tests (in `test_boolalg.py` and `test_simplify.py` at least, let's see how the rest goes...).

oscarbenjamin:
> With this diff I get `B & C` always appearing first:

I also get deterministic results now with the PR.

oscargus:
Nice! And the tests pass! Good to go?

oscarbenjamin:
It looks good to me. My only question would be about performance since I know that some of these routines can be very slow in large cases.

Do we have any benchmarks for this sort of stuff?

Could you maybe suggest an example for how to time this in a potentially slow case?

smichr:
> My only question would be about performance

Does this concern block this merge? With the number of issues being closed, it seems like this would be good to merge. And if it can be slow then we would still want the issues to be fixed, but in a better way.

oscarbenjamin:
> Does this concern block this merge? With the number of issues being closed, it seems like this would be good to merge. And if it can be slow then we would still want the issues to be fixed, but in a better way.

I think this is a good fix but I'd like to at least consider the performance implications. If there are slowdowns it doesn't mean that it should not be merged but at least we can consider whether there is a better place to implement this.

I ask about this because the tests are all for relatively simple examples but I can see several quadruply nested loops which is the kind of thing that can be slow so it would be good to consider under what conditions those loops could have large numbers of iterations e.g.:
```python
    while anythingchanged:
        ...
        for n, rowi in enumerate(oldndterms):
                ...
                for n2, row2i in enumerate(oldndterms):
                        ...
                        if all(a >= b for (a, b) in zip(filteredrows[n2], row)):
```
That's a loop in a loop in a loop in a loop. Maybe all loops only ever have a couple of iterations but I don't know. Maybe this is the best algorithm even if it's `O(n^4)` since many logic algorithms are `O(2^n)` or worse.

Testing with master it's not hard to see that some of these things are slow:
```python
In [10]: xs = symbols('x:10')

In [11]: fact = And(*(x1 < x2 for x1 , x2 in zip(xs[:-1], xs[1:])))

In [12]: %time simplify(fact) == fact
CPU times: user 8.91 s, sys: 49.8 ms, total: 8.96 s
Wall time: 9.01 s
Out[12]: True
```
That's 9 seconds to simplify an `And` with 10 terms and the simplified result as still the same as the input.

Increasing the size of the input leads to a nonlinear increase in time:
```python
In [18]: xs = symbols('x:30')

In [19]: fact = And(*(x1 < x2 for x1 , x2 in zip(xs[:-1], xs[1:])))

In [20]: %time ok = simplify(fact)
CPU times: user 1min 10s, sys: 870 ms, total: 1min 10s
Wall time: 1min 15s
```
(For some reason with 30 terms the result does not come back identical...)

Profiling suggests that the code here isn't involved:
```python
In [21]: xs = symbols('x:10')

In [22]: fact = And(*(x1 < x2 for x1 , x2 in zip(xs[:-1], xs[1:])))

In [24]: %prun -s cumulative simplify(fact)

         11091774 function calls (10736229 primitive calls) in 14.515 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      2/1    0.000    0.000   14.515   14.515 {built-in method builtins.exec}
        1    0.000    0.000   14.515   14.515 <string>:1(<module>)
     91/1    0.004    0.000   14.515   14.515 simplify.py:411(simplify)
        1    0.000    0.000   14.515   14.515 boolalg.py:730(_eval_simplify)
        1    0.024    0.024   11.914   11.914 boolalg.py:543(_apply_patternbased_simplification)
     3050    0.063    0.000   10.945    0.004 operations.py:481(__new__)
     3050    0.095    0.000    9.841    0.003 boolalg.py:692(_new_args_filter)
    12319    0.197    0.000    7.970    0.001 relational.py:202(canonical)
24796/24688    0.119    0.000    5.309    0.000 expr.py:2452(could_extract_minus_sign)
74558/49708    0.863    0.000    4.741    0.000 expr.py:2182(extract_multiplicatively)
235829/233547    0.562    0.000    4.262    0.000 cache.py:69(wrapper)
172874/62092    0.947    0.000    4.256    0.000 compatibility.py:501(ordered)
185857/160614    0.930    0.000    3.371    0.000 expr.py:143(__eq__)
        1    0.000    0.000    2.504    2.504 boolalg.py:444(_eval_simplify)
       18    0.002    0.000    2.481    0.138 relational.py:303(_eval_simplify)
    54/18    0.002    0.000    2.421    0.135 expr.py:710(equals)
51223/51200    0.088    0.000    2.340    0.000 decorators.py:254(_func)
51189/51166    0.078    0.000    2.215    0.000 decorators.py:129(binary_op_wrapper)
       18    0.001    0.000    2.049    0.114 expr.py:538(is_constant)
50142/50139    0.149    0.000    2.005    0.000 expr.py:259(__truediv__)
    36966    0.548    0.000    1.947    0.000 compatibility.py:315(default_sort_key)
       18    0.001    0.000    1.895    0.105 expr.py:457(_random)
       18    0.000    0.000    1.879    0.104 expr.py:498(<listcomp>)
       36    0.001    0.000    1.879    0.052 randtest.py:12(random_complex_number)
   144/72    0.004    0.000    1.878    0.026 simplify.py:1439(nsimplify)
       72    0.003    0.000    1.864    0.026 simplify.py:1578(_real_to_rational)
       36    0.001    0.000    1.722    0.048 simplify.py:1533(nsimplify_real)
    59/36    0.015    0.000    1.689    0.047 identification.py:533(identify)
   257942    0.499    0.000    1.664    0.000 symbol.py:306(_hashable_content)
      898    1.242    0.001    1.620    0.002 identification.py:17(pslq)
    45/27    0.000    0.000    1.325    0.049 basic.py:1673(simplify)
        1    0.000    0.000    1.325    1.325 boolalg.py:445(<listcomp>)
374293/373977    0.566    0.000    1.201    0.000 sympify.py:92(sympify)
        1    0.000    0.000    1.168    1.168 boolalg.py:2796(simplify_logic)
```
I don't know if that is the right sort of example to test this code but it does show the impact that some of these algorithms have as soon as you scale up beyond tiny inputs. The input I used here is not what I would call "large":
```python
In [25]: fact
Out[25]: x₀ < x₁ ∧ x₁ < x₂ ∧ x₂ < x₃ ∧ x₃ < x₄ ∧ x₄ < x₅ ∧ x₅ < x₆ ∧ x₆ < x₇ ∧ x₇ < x₈ ∧ x₈ < x₉
```
@oscargus can you suggest what would be a good way to time this code in a case where it succeeds in simplifying and also a case where it does not?

oscargus:
Well, it is a set cover problem that is solved, one of Karp's original 21 NP-complete problems, https://en.wikipedia.org/wiki/Set_cover_problem. So even now it is a heuristic, but slightly better than the previous (and I have updated it at least twice earlier, first to get better results/smaller expressions, second to get better performance). The new code is not really affecting the performance as it is "only" n^3 as far as I can tell. Another issue is that n is here (up to) 2^terms, hence, the `force` keyword argument which, unless set to `True`, disables this part if there are more than eight variables.

It would indeed be interesting to get some benchmark numbers in though. I think I've looked into that earlier, but didn't really come up with anything (and at that time, the current benchmark action didn't exist).

In your example, you have nine variables (each inequality becomes a variable in this respect), so not really sure if this part of the code is called. On the other hand, there is only one expression for the boolean simplification, so it wouldn't really become that time consuming. Basically, for this part of the code, your problem is just simplifying A & B & C & D ..., so that is very easy (already simplified as there is a single AND-expression, cannot really be any smaller).

oscargus:
I think that there are probably slightly more efficient ways to obtain a similar heuristic as the current one. Not sure if it is really the purpose of this PR though. No worse asymptotic performance and better results in some cases (not worse compares to the previous).

oscarbenjamin:
Okay that makes sense. Then this looks good to merge.

I do think that the performance of logic simplification needs to be improved in general though.

oscargus:
Sounds good!

I'm rewriting it a bit now, but not sure if there is any major changes in performance. Some things will be better, some may be worse, so probably depends on case if it improves. A bit better structure though compared to the current one...

oscargus:
The thing is that for small problems the current algorithm is enough and I think it may give close to optimal results. For larger problems, it is not. One could possibly use things like https://en.wikipedia.org/wiki/Espresso_heuristic_logic_minimizer but then we are talking external dependencies etc. Fundamentally hard problem when the problem size grows.

oscargus:
Nah, I didn't manage to get the improved version to work and will need to start all over at some later stage. So if this is merged, I can rebase other related PRs.

oscargus:
Gah! Found the problem, so will push a new version in a few minutes...

oscarbenjamin:
> Fundamentally hard problem when the problem size grows.

Yes, but then it shouldn't be attempted even by `simplify` without some additional `force` flag or something. The flipside of any algorithm like this in a system like SymPy is that if the algorithm is being invoked indirectly then users don't actually control when that happens. Here this is called by `simplify` so in many cases the cost of the algorithm will be paid for no benefit e.g. the example above where it takes 10 seconds for simplify to achieve nothing.

We have to remember that performance is not orthogonal to capability: often in SymPy something is so slow that effectively the result can not be computed (e.g. "simplify hangs for this expression"). When that happens the impact can be much wider than the type of problem that the slow algorithm was designed for.

oscargus:
I just bench marked a bit for the problem in #14700 which has eight variables.

Of the 2.5 seconds it takes, 5 ms is in the currently discussed code (n^4). In your code example above, simplify_logic takes 1.168 seconds out of the 14. I would expect that the time for the n^4-code is about the same there. Most of the time in simplify_logic is spent doing `xreplace` for small number of variables.

Yes, hanging is one thing, getting "why isn't this Boolean expression fully simplified?" is another. Yes, the time may increase further for larger problems, but currently it is repeated calls to small pieces of code that takes quite a bit of time. For example, in my test, 47193 calls to `sympify`, requiring a total of 127 ms local time, for simplifying an expression with plain symbols and operators.

One should also distinguish between `simplify` and `simplify_logic`. Two quite different functions, where the second is only a small part of the first.

sympy/logic/boolalg.py
```diff
@@ -1363,7 +1364,7 @@ def __new__(cls, *args, **options):
             return And(*argset)
         if False in argset:
             argset.discard(False)
-            return And(*[~arg for arg in argset])
+            return And(*(~arg for arg in argset))
```
oscarbenjamin, line 1367: Does this kind of thing actually lead to any speed up?
```python
In [1]: %timeit [].extend([a for a in range(1000)])
44.1 µs ± 107 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [2]: %timeit [].extend(a for a in range(1000))
69 µs ± 1.21 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
I think that the chief difference between a list comprehension and a generator expression in performance terms is that in a generator expression the active frame is switched twice for each item yielded effectively adding the cost of a function call for each item.

Internally a list comprehension
```python
a.extend([e for e in stuff])
```
translates to:
```python
def _tmp():
    res = []
    for e in stuff:
        res.append(e)
    return res

_res = _tmp()
a.extend(_res)
```
whereas a generator expression
```
a.extend(e for e in stuff)
```
translates to
```python
def _tmp():
    for e in stuff:
        yield e

_gen = _tmp()
a.extend(_gen)
```
Now as the `a.extend` method iterates over `_gen` the `_tmp` function is repeatedly suspended and restarted to invoke the code between each pair of yields.

What this means is that actually iterating over a generator expression is slower than iterating over a list comprehension. The advantage of the generator expression is that it is lazy so if there is any possibility to shortcut the iteration then we don't have to compute all of the items and then it can be a big performance boost. Also if the full list of items would e.g. use too much memory or something then the generator makes it possible to process one at a time with `O(1)` memory. 

If all items from the iterable are consumed and ultimately stored in memory simultaneously then the advantages of a generator are gone and it's faster just to use a list comprehension. Here the use of `*` unpacking forces the generator into a tuple to be passed to `And.__new__` I think it's faster to create that tuple from a list than from a generator expression. In any case if we don't get to shortcut then any potential optimisations here are unlikely to be noticeable. I expect that `And.__new__` plus sympifying every element will be much slower.

If we did want to optimise this kind of thing then potentially a good approach is adding an optimised `Add._fromiter` function could work. The idea would be that all inputs are already sympified and the argument is potentially a generator that can be shortcutted. Then `_fromiter` could iterate over the generator looking for `S.false` to exit early. (Note that there is a `Basic.fromiter` method but it has no performance benefits in any situation.)

In any case with optimisations like this we need some reasoning or benchmark or timings to justify them or otherwise it's best just to leave things alone.

oscargus: For the `_convert_to_varsSOP/POS` functions I tested a bit and there it is better to not create a list. Not sure about the other cases, so I'll revert those.

oscargus:
Now I think I am happy with this (if the tests pass). Reverted most generators to lists (even a few more). Also modified some other parts for slight improvements, primarily using `map`.

sympy/logic/boolalg.py
```diff
@@ -2289,10 +2287,7 @@ def _compare_term(minterm, term):
     Return True if a binary term is satisfied by the given term. Used
     for recognizing prime implicants.
     """
-    for m, t in zip(minterm, term):
-        if t != 3 and m != t:
-            return False
-    return True
+    return all(t == 3 or minterm[n] == t for n, t in enumerate(term))
```
oscarbenjamin, line 2290: Is it not better to use zip?

sympy/logic/boolalg.py
```diff
@@ -2373,7 +2401,7 @@ def _input_to_binlist(inputlist, variables):
             nonspecvars = list(variables)
             for key in val.keys():
                 nonspecvars.remove(key)
-            for t in product([0, 1], repeat=len(nonspecvars)):
+            for t in product((0, 1), repeat=len(nonspecvars)):
```
oscarbenjamin, line 2404: This is fine but in general it would be better to make these kinds of changes in a separate PR from one that actually changes behaviour or fixes bugs etc.

sympy/logic/boolalg.py
```diff
@@ -2822,20 +2852,22 @@ def simplify_logic(expr, form=None, deep=True, force=False):
...
+            form_ok = is_dnf(expr)
+
+        if form_ok and all(is_literal(a)
+                for a in expr.args):
+            return expr
     if deep:
         variables = _find_predicates(expr)
         from sympy.simplify.simplify import simplify
-        s = [simplify(v) for v in variables]
+        s = tuple(map(simplify, variables))
```
oscarbenjamin, line 2870: Generally I prefer list comprehensions for things like this just because it looks clearer to me. Is there actually a speed advantage here?

Does the change from list to tuple have any observable effect on the output from this function?

In my experience `map` is faster if the iterable and the function are all builtins but the speed difference evaporates once the function is a python function or if the iterable is something like a generator.

oscargus: I've bench-marked some things, not all, and tried to generalize it and it seems like tuples are faster, even with your own functions (as long as it is not lambdas, `map` should not be slower from what I've read). In this file it was a mix of `map` and list comprehensions for exactly this case, so I thought I'd unify it and went for this.

sympy/logic/boolalg.py
oscargus: You are correct. Tends to just see lots of things and, a bit manically, change them...

sympy/logic/boolalg.py
oscargus: No, as this enables early exit which zip doesn't. Clear speed-advantage here, statistically at least.

There is a handful of `any` and `all`-statements in the code that uses `zip` that one may want to consider at a later stage. One should probably check how large the iterators are though, for just a few items it may be slower I think, but for "enough" there will be a significant improvement.

oscarbenjamin: Why doesn't `zip` enable an early exit?

I was thinking of something like:
```
return all(t == 3 or t == mt for t, mt in zip(term, minterm)):
return all(t in (3, mt) for t, mt in zip(term, minterm)):
```

oscargus: You are right in that it is not the reason. However, my experiments show that it is faster. I think that the reason is that you can avoid fetching the minterm argument every now and then since it is enough to test if the term argument is 3 and only if not fetch the minterm argument. But I can try to recreate the timings if you think it is required.

Compare with the `_check_pairs` function which is faster without a zip, although that one also includes an enumerate. See https://github.com/sympy/sympy/commit/d3e5f14c5241b9da96560c4deffef9eaf4991228#diff-0fa274ccbfd01ea485764211bac9ecce2ac3819d1c110968d0b13c7da3bab6f0 for the earlier version.

(Now that it is a one-liner one can also replace it in the code, but that hold for all one-liners...)

oscargus: Nah, I may be confusing things. I realize that I change from a  loop with enumerate to the current zip-loop earlier... I'll try the three single-line cases and come back...

oscargus:
Where on earth are the benchmarks defined? I've found multiple benchmarks, but not the ones from the GitHub Actions...

oscarbenjamin:
> Where on earth are the benchmarks defined? I've found multiple benchmarks, but not the ones from the GitHub Actions...

https://github.com/sympy/sympy_benchmarks

oscarbenjamin:
> > Where on earth are the benchmarks defined? I've found multiple benchmarks, but not the ones from the GitHub Actions...
> 
> https://github.com/sympy/sympy_benchmarks

The benchmarks that are still in the main sympy repo are unused and potentially broken. All benchmarks should be moved over to the benchmarks repo and then removed from the main sympy repo.

oscarbenjamin:
Does this have full diff coverage from the tests?

If so this looks good to me so go ahead and merge if you're done.

sympy/logic/boolalg.py
oscargus: Your first oneliner seems marginally faster then mine (times vary, maybe hard to say any is better). The second clearly slower than both.

oscargus:
I do not recall how to check coverage, but it should be. The case I have added is for sure executed as it is the one that solves #19114 and #20870. Then there are some generic rewrites, but not really any specific cases one can trigger or not.

PR #21855, opened 2021-08-11
Correct calculation of variance and central moment for non-random expressions

rikardn:
Calculate `sympy.stats.variance` and `sympy.stats.cmoment` correctly for non-random expressions.

This is a small part of the solution to #21057

Before:

```
>>> sympy.stats.variance(4 + x)
x + 4
```

After:

```
>>> sympy.stats.variance(4 + x)
0
```

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/stats/rv_interface.py
+++ b/sympy/stats/rv_interface.py
@@ -422,6 +422,9 @@ def median(X, evaluate=True, **kwargs):
     .. [1] https://en.wikipedia.org/wiki/Median#Probability_distributions
 
     """
+    if not is_random(X):
+        return X
+
     from sympy.stats.crv import ContinuousPSpace
     from sympy.stats.drv import DiscretePSpace
     from sympy.stats.frv import FinitePSpace
--- a/sympy/stats/symbolic_probability.py
+++ b/sympy/stats/symbolic_probability.py
@@ -609,8 +609,6 @@ class Moment(Expr):
             return super().__new__(cls, X, n, c)
 
     def doit(self, **hints):
-        if not is_random(self.args[0]):
-            return self.args[0]
         return self.rewrite(Expectation).doit(**hints)
 
     def _eval_rewrite_as_Expectation(self, X, n, c=0, condition=None, **kwargs):
@@ -668,8 +666,6 @@ class CentralMoment(Expr):
             return super().__new__(cls, X, n)
 
     def doit(self, **hints):
-        if not is_random(self.args[0]):
-            return self.args[0]
         return self.rewrite(Expectation).doit(**hints)
 
     def _eval_rewrite_as_Expectation(self, X, n, condition=None, **kwargs):
--- a/sympy/stats/tests/test_rv.py
+++ b/sympy/stats/tests/test_rv.py
@@ -6,7 +6,7 @@ from sympy.stats import (Die, Normal, Exponential, FiniteRV, P, E, H, variance,
         random_symbols, sample, Geometric, factorial_moment, Binomial, Hypergeometric,
         DiscreteUniform, Poisson, characteristic_function, moment_generating_function,
         BernoulliProcess, Variance, Expectation, Probability, Covariance, covariance, cmoment,
-        moment)
+        moment, median)
 from sympy.stats.rv import (IndependentProductPSpace, rs_swap, Density, NamedArgsMixin,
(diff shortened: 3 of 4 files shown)


sympy/stats/symbolic_probability.py
```diff
@@ -668,8 +668,6 @@ def __new__(cls, X, n, condition=None, **kwargs):
             return super().__new__(cls, X, n)
 
     def doit(self, **hints):
-        if not is_random(self.args[0]):
-            return self.args[0]
```
oscarbenjamin, line 672: I guess this should have been something like `self.args[0] ** self.args[1]`.

Should the Moment.doit method be fixed as well?
https://github.com/sympy/sympy/blob/0f6d73ba2a844ce24bb8cd6b8a57177e1bccca23/sympy/stats/symbolic_probability.py#L611-L614

rikardn: Thanks @oscarbenjamin for reviewing. I did the same fix for `moment` and added tests for the n=0 case.

rikardn:
I had to change the following test:

```
Moment(5, 5) = 5
```

to 

```
Moment(5, 5) == 5**5
```

I believe it was incorrect before since the moment is `E[X**n] = E[5**5] = 5**5` if I am not mistaken.

Maelstrom6:
Would sampling also be another use-case for constants?

oscarbenjamin:
Ah yes that's another problem:
```python
In [48]: sample(1)
...
ValueError: Expression containing Random Variable expected, not 1
```

sympy/stats/rv_interface.py
```diff
@@ -422,6 +422,9 @@ def median(X, evaluate=True, **kwargs):
     .. [1] https://en.wikipedia.org/wiki/Median#Probability_distributions
 
     """
+    if not is_random(X):
+        return X
```
oscarbenjamin, line 426: Is it necessary to special case this? What fails otherwise (and can the other thing that fails be fixed instead)?

rikardn: The problem is that `pspace` cannot be called on an expression with no random variables. It would raise `ValueError: Expression containing Random Variable expected`. One alternative to handling it as a special case would be to let `pspace` return something useful for constant expressions. Perhaps something like a "Dirac distribution pspace", "constant distribution pspace" or "degenerate distribution pspace". Perhaps this would remove the need for special cases also in other situations.

rikardn:
Yes, you are right @Maelstrom6. Some other cases are also mentioned in the #21057 that triggered this PR. I didn't fix these at the same time to keep the PR small to simplify reviewing.

sympy/stats/rv_interface.py
oscarbenjamin: Okay, so that goes back to what you originally suggested about having a Dirac distribution. Right now a constant expression needs special case handling to work in some situations. However making it work in general requires implementing a pspace. Maybe we should avoid adding special case code if there is a more general way to fix it.

Maelstrom6: I'd assume adjusting `pspace` like that would simply move the `if` statement around and wouldn't actually remove it. Removing similar statements from higher-level functions might create bugs with unconsidered edge-cases.

Maelstrom6:
> Yes, you are right @Maelstrom6. Some other cases are also mentioned in the #21057 that triggered this PR. I didn't fix these at the same time to keep the PR small to simplify reviewing.

Understood. Hopefully we'll get to it eventually.

sympy/stats/rv_interface.py
rikardn: I think adding a new pspace is the right thing to do. That being said I don't have the full overview of the stats module so I cannot tell if it would directly affect more code than this. One way of finding out is to try it and ask the unit tests.

oscarbenjamin: > I'd assume adjusting `pspace` like that would simply move the `if` statement around and wouldn't actually remove it.

The question is whether we add a special case to every function that does anything with random variables or whether we implement the trivial pspace so all of those functions can work without needing to care that the expression is constant or not.

I don't think that this needs to hold up the PR but adding more checks like this to every function that doesn't work properly with constant RVs would not be the right approach.

oscarbenjamin:
This looks good to me. Further improvements like `sample` etc should be made by implementing a pspace I think.

rikardn:
Thanks for reviewing and merging

PR #21856, opened 2021-08-11
Updated slowness of tests

oscargus:
Some tests have not been marked as slow and some was marked, but not really slow. The threshold was set a bit arbitrary (so there will still show up entries in the corresponding lists), but at least it is slightly better than before.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/concrete/tests/test_sums_products.py
+++ b/sympy/concrete/tests/test_sums_products.py
@@ -988,6 +988,7 @@ def test_indexed_idx_sum():
     raises(ValueError, lambda: Product(A[k], (k, 2, oo)))
 
 
+@slow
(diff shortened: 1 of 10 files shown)


oscarbenjamin:
I've been thinking about changing it so that a test that is not marked as slow but then takes a long time (or vice versa) leads to a CI failure. That way contributors would be forced to keep the slow markers up to date with each PR. Any significant slowdown in a test would be visible in the diff for PR reviewers and also newly introduced but needlessly slow tests can be queried or simplified. Ideally contributors would be encouraged to write fast tests in the first place but the time taken to run newly introduced tests is not immediately obvious to a reviewer.

PR #21871, opened 2021-08-14
[GSoC] Function for I.L.D. for Moment added

Psycho-Pirate:

See #19069

Function `solve_for_ild_moment()` is added which calculates the moment equation under the effect of a moving load.
Function `plot_ild_moment()` is added which plots these equations.

#### Release Notes

* physics.continuum_mechanics 
   * Added ability to calculate and plot I.L.D. Moment Equations

### DIFF
--- a/sympy/physics/continuum_mechanics/beam.py
+++ b/sympy/physics/continuum_mechanics/beam.py
@@ -133,7 +133,7 @@ class Beam:
         self._reaction_loads = {}
         self._ild_reactions = {}
         self._ild_shear = 0
-
+        self._ild_moment = 0
         # _original_load is a copy of _load equations with unsubstituted reaction
         # forces. It is used for calculating reaction forces in case of I.L.D.
         self._original_load = 0
@@ -159,6 +159,11 @@ class Beam:
     def ild_reactions(self):
         """ Returns the I.L.D. reaction forces in a dictionary."""
         return self._ild_reactions
+
+    @property
+    def ild_moment(self):
+        """ Returns the I.L.D. moment equation."""
+        return self._ild_moment
 
     @property
     def length(self):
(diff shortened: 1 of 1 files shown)


Psycho-Pirate:
@ishanaj

sympy/physics/continuum_mechanics/beam.py
```diff
@@ -1884,6 +1889,146 @@ def plot_ild_shear(self,subs=None):
         return plot(self._ild_shear.subs(subs), (x, 0, l),  title='I.L.D. for Shear',
                xlabel=r'$\mathrm{X}$', ylabel=r'$\mathrm{V}$', line_color='blue',show=True)
 
+    def _solve_for_ild_moment(self):
```
ishanaj, line 1892: Instead of having 3 different `_solve_for_ild_....()` functions, we can have a single `_solve_for_ild_equations`, since they more or less perform the same operations.

Psycho-Pirate: I'm not sure that will work because the function will return both shear eq and moment eq. In the shear function, moment eq is not required, similarly for moment function, shear eq is not required. These will be unused variables and the Travis tests for sympy shows error when you leave something unused.

ishanaj: Try using an underscore instead. Like for the case where shear is not required use:

 `_ , moment = _solve_for_ild_equations()`

I think this should work!

Psycho-Pirate: I have removed the additional functions

ishanaj:
LGTM!

ishanaj (approved):

ishanaj:
Merging!

PR #21873, opened 2021-08-14
Fixed bug when trying to simplify non-linear relationals 

### LINKED ISSUE #21869: Incorrect relational simplification with squares
Current master:
```
from sympy import *
x = Symbol('x', real=True)
y = Symbol('y', real=True)
expr = And(Eq(x**2, 4), Ge(y, x))
expr.simplify()
```
returns
`(y >= 4) & Eq(x**2, 4)`
so the square term in the equality is not accounted for (the code seems to just handle linear relations, but doesn't detect that this is not linear). The code is here:
https://github.com/sympy/sympy/blob/ec0fe8c5f3e59840e8aa5d3d6a7c976e40f76b64/sympy/logic/boolalg.py#L760-L761

I think it is related to the recently(?) introduced exception `NonlinearError` which is thrown, but not caught here. A bit strange that it is not passed on though.

oscargus:

Closes #21869

The previous code indirectly assumed that certain forms of non-linear expressions didn't show up. Here, it is fixed so that this doesn't happen. Also, this will simplify slightly better in terms of several Eq with the same variable, earlier, if the first one didn't allow extraction of the value (i.e. being linear) it was not simplified later either.

#### Release Notes

* logic
  * Fixed simplification bug when equalities contained certain non-linear expressions.

### DIFF
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -752,10 +752,11 @@ class And(LatticeOp, BooleanFunction):
                 (i.free_symbols, i) for i in eqs]),
                 lambda x: len(x[0]))
             eqs = []
+            nonlineqs = []
             while 1 in sifted:
                 for free, e in sifted.pop(1):
                     x = free.pop()
-                    if e.lhs != x or x in e.rhs.free_symbols:
+                    if (e.lhs != x or x in e.rhs.free_symbols) and x not in reps:
                         try:
                             m, b = linear_coeffs(
                                 e.rewrite(Add, evaluate=False), x)
@@ -768,12 +769,13 @@ class And(LatticeOp, BooleanFunction):
                         except ValueError:
                             pass
                     if x in reps:
-                        eqs.append(e.func(e.rhs, reps[x]))
-                    elif e.lhs == x:
+                        eqs.append(e.subs(x, reps[x]))
+                    elif e.lhs == x and x not in e.rhs.free_symbols:
                         reps[x] = e.rhs
                         eqs.append(e)
                     else:
-                        eqs.append(e)
+                        # x is not yet identified, but may be later
+                        nonlineqs.append(e)
                 resifted = defaultdict(list)
                 for k in sifted:
                     for f, e in sifted[k]:
@@ -783,8 +785,9 @@ class And(LatticeOp, BooleanFunction):
                 sifted = resifted
         for k in sifted:
             eqs.extend([e for f, e in sifted[k]])
+        nonlineqs = [ei.subs(reps) for ei in nonlineqs]
         other = [ei.subs(reps) for ei in other]
-        rv = rv.func(*([i.canonical for i in (eqs + other)] + nonRel))
+        rv = rv.func(*([i.canonical for i in (eqs + nonlineqs + other)] + nonRel))
         patterns = simplify_patterns_and()
         return self._apply_patternbased_simplification(rv, patterns,
                                                        measure, False)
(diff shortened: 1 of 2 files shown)


sympy/logic/boolalg.py
```diff
@@ -752,10 +752,11 @@ def _eval_simplify(self, **kwargs):
                 (i.free_symbols, i) for i in eqs]),
                 lambda x: len(x[0]))
             eqs = []
+            nonlineqs = []
             while 1 in sifted:
                 for free, e in sifted.pop(1):
                     x = free.pop()
```
smichr, line 758: `x = next(ordered(free))` would be canonical, but I'm not sure it matters since all single-symbol expressions will eventually be solved. But thinking about this and a recent SO question where recursive resolution of linear expressions is a desirable action, maybe this should be a solve-routine, maybe `extract_known(eqs) -> {xi: vi}, nonlinear_eqs`?

sympy/logic/boolalg.py
```diff
@@ -768,10 +769,13 @@ def _eval_simplify(self, **kwargs):
...
                     if x in reps:
-                        eqs.append(e.func(e.rhs, reps[x]))
-                    else:
+                        eqs.append(e.subs(x, reps[x]))
+                    elif e.lhs == x and x not in e.rhs.free_symbols:
                         reps[x] = e.rhs
                         eqs.append(e)
+                    else:
+                        # x is not yet identified, but may be later
+                        nonlineqs.append(e)
```
smichr, line 778: I would suggest a slight reordering as:
```python
                        except ValueError:
                            # x is not yet identified, but may be later
                            nonlineqs.append(e)
                    elif x in reps:
                        eqs.append(e.subs(x, reps[x]))
                    else:
                        reps[x] = e.rhs
                        eqs.append(e)
```

sympy/logic/boolalg.py
oscargus: OK! Not sure if I get it fully, but say that we have `And(Eq(x**2, x), Eq(x, 1))` or `And(Eq(x**2, 1), Eq(x, 1))`, do we always know that `Eq(x,1)` is processed first? Or `And(Eq(y**2, 1), Eq(x, y), Eq(x, 1))`, isn't there a risk that `Eq(y**2, 1)` is passed on to `eqs` and that `y=1` is never substituted there?

sympy/logic/boolalg.py
oscargus: Makes sense!

oscargus: Ohh, but that would require 
```
reps[x] = e.rhs
eqs.append(e)
```
before `except ValueError` as well, right? Can be done and maybe clearer, but do you think it should be like that?

smichr: No. I think what you have is right. The blocks cover, in order "either false", "one true", "other true", "both true".

sympy/logic/boolalg.py
smichr: Looking more closely. We are dealing with 1 free symbol so there is nothing non-canonical about popping that one symbol from `free` ;-). If we order `(i.free_symbols, i) for i in eqs])` as `(i.free_symbols, i) for i in ordered(eqs)])` then equality `Eq(x, 1)` will come before `Eq(x**2, x)`
```python
>>> list(ordered((Eq(x**2,x),Eq(x,1))))
[Eq(x, 1), Eq(x**2, x)]
```
I am not sure if the equations get made canonical by this step, but if not then
```python
(i.free_symbols, i) for i in [i.canonical for i in ordered(eqs)]])
```
would be best. That will put `Eq(1,x)` as `Eq(x, 1)`.

smichr: no -- the solving for linear term handles rearrangement. I think this is fine as it is.

PR #21878, opened 2021-08-15
Minor speed-ups by not constructing lists for all and any and using tuples for constant iterators

oscargus:
Not creating an explicit list is faster for `any` and `all`, which can terminate early if an iterator is provided instead.

When creating an iterator with constants, tuples tend to be faster than lists.

#### Other comments
Some minor modifications to e.g. avoid computing lengths several times.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/categories/diagram_drawing.py
+++ b/sympy/categories/diagram_drawing.py
@@ -2125,15 +2125,15 @@ class XypicDiagramDrawer:
         if dom_i == 0:
             free_up = True
         else:
-            free_up = all([grid[dom_i - 1, j] for j in
-                           range(start, end + 1)])
+            free_up = all(grid[dom_i - 1, j] for j in
+                          range(start, end + 1))
 
         # Check for free space below.
         if dom_i == grid.height - 1:
             free_down = True
         else:
-            free_down = all([not grid[dom_i + 1, j] for j in
-                             range(start, end + 1)])
+            free_down = not any(grid[dom_i + 1, j] for j in
+                                range(start, end + 1))
 
         return (free_up, free_down, backwards)
 
@@ -2155,14 +2155,14 @@ class XypicDiagramDrawer:
         if dom_j == 0:
             free_left = True
         else:
-            free_left = all([not grid[i, dom_j - 1] for i in
-                             range(start, end + 1)])
+            free_left = not any(grid[i, dom_j - 1] for i in
+                                range(start, end + 1))
 
         if dom_j == grid.width - 1:
             free_right = True
         else:
-            free_right = all([not grid[i, dom_j + 1] for i in
-                              range(start, end + 1)])
+            free_right = not any(grid[i, dom_j + 1] for i in
+                                 range(start, end + 1))
 
         return (free_left, free_right, backwards)
 
--- a/sympy/combinatorics/fp_groups.py
+++ b/sympy/combinatorics/fp_groups.py
@@ -138,9 +138,9 @@ class FpGroup(DefaultPrinting):
 
         '''
 
-        if not all([isinstance(g, FreeGroupElement) for g in gens]):
+        if not all(isinstance(g, FreeGroupElement) for g in gens):
             raise ValueError("Generators must be `FreeGroupElement`s")
-        if not all([g.group == self.free_group for g in gens]):
+        if not all(g.group == self.free_group for g in gens):
                 raise ValueError("Given generators are not members of the group")
         if homomorphism:
             g, rels, _gens = reidemeister_presentation(self, gens, C=C, homomorphism=True)
@@ -259,7 +259,7 @@ class FpGroup(DefaultPrinting):
         used_gens = set()
         for r in self.relators:
             used_gens.update(r.contains_generators())
-        if any([g not in used_gens for g in self.generators]):
+        if not set(self.generators) <= used_gens:
             return True
         # Abelianisation test: check is the abelianisation is infinite
         abelian_rels = []
@@ -1030,7 +1030,7 @@ def elimination_technique_1(gens, rels, identity):
         # don't look for a redundant generator in a relator which
         # depends on previously found ones
         contained_gens = rel.contains_generators()
-        if any([g in contained_gens for g in redundant_gens]):
+        if any(g in contained_gens for g in redundant_gens):
             continue
         contained_gens = list(contained_gens)
         contained_gens.sort(reverse = True)
--- a/sympy/combinatorics/homomorphisms.py
+++ b/sympy/combinatorics/homomorphisms.py
@@ -290,9 +290,9 @@ def homomorphism(domain, codomain, gens, images=[], check=True):
         raise TypeError("The codomain must be a group")
 
     generators = domain.generators
-    if any([g not in generators for g in gens]):
+    if not all(g in generators for g in gens):
         raise ValueError("The supplied generators must be a subset of the domain's generators")
-    if any([g not in codomain for g in images]):
+    if not all(g in codomain for g in images):
         raise ValueError("The images must be elements of the codomain")
 
     if images and len(images) != len(gens):
--- a/sympy/combinatorics/perm_groups.py
+++ b/sympy/combinatorics/perm_groups.py
@@ -773,7 +773,7 @@ class PermutationGroup(Basic):
 
         """
         der = self.derived_series()
-        if not (all(g.is_identity for g in der[-1].generators)):
+        if not all(g.is_identity for g in der[-1].generators):
             raise NotImplementedError('Group should be solvable')
         series = []
 
@@ -845,7 +845,7 @@ class PermutationGroup(Basic):
                 b = base_ordering[base[l]^u]
                 for t in T:
                     p = t*u
-                    if all([base_ordering[h^p] >= b for h in orbits[l]]):
+                    if all(base_ordering[h^p] >= b for h in orbits[l]):
                         T_next.append(p)
                     if t_len + len(T_next) == indices[l]:
                         break
@@ -3933,7 +3933,7 @@ class PermutationGroup(Basic):
            elements of the group
         """
 
-        if not all([g in self for g in gens]):
+        if not all(g in self for g in gens):
             raise ValueError("The group doesn't contain the supplied generators")
 
         G = PermutationGroup(gens)
--- a/sympy/combinatorics/permutations.py
+++ b/sympy/combinatorics/permutations.py
@@ -983,7 +983,7 @@ class Permutation(Atom):
         temp = set(temp)
 
         if not is_cycle:
-            if any(i not in temp for i in range(len(temp))):
+            if temp != set(range(len(temp))):
                 raise ValueError('Integers 0 through %s must be present.' %
                 max(temp))
             if size is not None and temp and max(temp) + 1 > size:
@@ -1152,8 +1152,7 @@ class Permutation(Atom):
         [[0], [1, 2]]
         """
         need = set(range(self.size)) - set(flatten(self.cyclic_form))
-        rv = self.cyclic_form
-        rv.extend([[i] for i in need])
+        rv = self.cyclic_form + [[i] for i in need]
         rv.sort()
         return rv
 
--- a/sympy/combinatorics/tensor_can.py
+++ b/sympy/combinatorics/tensor_can.py
@@ -403,7 +403,7 @@ def double_coset_can_rep(dummies, sym, b_S, sgens, S_transversals, g):
     g = g.array_form
     num_dummies = size - 2
     indices = list(range(num_dummies))
-    all_metrics_with_sym = all([_ is not None for _ in sym])
+    all_metrics_with_sym = not any(_ is None for _ in sym)
     num_types = len(sym)
     dumx = dummies[:]
     dumx_flat = []
@@ -763,12 +763,12 @@ def canonicalize(g, dummies, msym, *v):
     """
     from sympy.combinatorics.testutil import canonicalize_naive
     if not isinstance(msym, list):
-        if not msym in [0, 1, None]:
+        if not msym in (0, 1, None):
             raise ValueError('msym must be 0, 1 or None')
         num_types = 1
     else:
         num_types = len(msym)
-        if not all(msymx in [0, 1, None] for msymx in msym):
+        if not all(msymx in (0, 1, None) for msymx in msym):
             raise ValueError('msym entries must be 0, 1 or None')
         if len(dummies) != num_types:
             raise ValueError(
--- a/sympy/geometry/entity.py
+++ b/sympy/geometry/entity.py
@@ -171,7 +171,7 @@ class GeometryEntity(Basic):
             # will fall back to the next representation
             return None
 
-        if any([not x.is_number or not x.is_finite for x in bounds]):
+        if not all(x.is_number and x.is_finite for x in bounds):
             return None
 
         svg_top = '''<svg xmlns="http://www.w3.org/2000/svg"
--- a/sympy/physics/continuum_mechanics/beam.py
+++ b/sympy/physics/continuum_mechanics/beam.py
@@ -900,10 +900,9 @@ class Beam:
                 points = solve(shear_slope, x)
                 val = []
                 for point in points:
-                    val.append(shear_curve.subs(x, point))
+                    val.append(abs(shear_curve.subs(x, point)))
                 points.extend([singularity[i-1], s])
-                val.extend([limit(shear_curve, x, singularity[i-1], '+'), limit(shear_curve, x, s, '-')])
-                val = list(map(abs, val))
+                val += [abs(limit(shear_curve, x, singularity[i-1], '+')), abs(limit(shear_curve, x, s, '-'))]
                 max_shear = max(val)
                 shear_values.append(max_shear)
                 intervals.append(points[val.index(max_shear)])
@@ -986,10 +985,9 @@ class Beam:
                 points = solve(moment_slope, x)
                 val = []
                 for point in points:
-                    val.append(bending_curve.subs(x, point))
+                    val.append(abs(bending_curve.subs(x, point)))
                 points.extend([singularity[i-1], s])
-                val.extend([limit(bending_curve, x, singularity[i-1], '+'), limit(bending_curve, x, s, '-')])
-                val = list(map(abs, val))
+                val += [abs(limit(bending_curve, x, singularity[i-1], '+')), abs(limit(bending_curve, x, s, '-'))]
                 max_moment = max(val)
                 moment_values.append(max_moment)
                 intervals.append(points[val.index(max_moment)])
--- a/sympy/sets/handlers/intersection.py
+++ b/sympy/sets/handlers/intersection.py
@@ -264,8 +264,7 @@ def intersection_sets(self, other): # noqa:F811
             # not helpful here: multiple parametric solutions.
             if len(solns) == 0:
                 return EmptySet
-            elif any(not isinstance(s, int) and s.free_symbols
-                     for tupl in solns for s in tupl):
+            elif any(s.free_symbols for tupl in solns for s in tupl):
                 if len(solns) == 1:
                     soln, solm = solns[0]
                     (t,) = soln.free_symbols
--- a/sympy/sets/sets.py
+++ b/sympy/sets/sets.py
@@ -213,8 +213,6 @@ class Set(Basic, EvalfMixin):
             return S.EmptySet
 
         elif isinstance(other, FiniteSet):
-            from sympy.utilities.iterables import sift
-
             sifted = sift(other, lambda x: fuzzy_bool(self.contains(x)))
             # ignore those that are contained in self
             return Union(FiniteSet(*(sifted[False])),
@@ -867,7 +865,7 @@ class ProductSet(Set):
         return reduce(lambda a, b: a*b, (len(s) for s in self.args))
 
     def __bool__(self):
-        return all([bool(s) for s in self.sets])
+        return all(self.sets)
 
 
 class Interval(Set):
@@ -1265,12 +1263,12 @@ class Union(Set, LatticeOp):
             # Clear out duplicates
             sos_list = []
             sets_list = []
-            for set in sets:
-                if set[0] in sos_list:
+            for _set in sets:
+                if _set[0] in sos_list:
                     continue
                 else:
-                    sos_list.append(set[0])
-                    sets_list.append(set)
+                    sos_list.append(_set[0])
+                    sets_list.append(_set)
             sets = sets_list
 
             # Flip Parity - next time subtract/add if we added/subtracted here
@@ -1906,7 +1904,6 @@ class FiniteSet(Set):
 
     def as_relational(self, symbol):
         """Rewrite a FiniteSet in terms of equalities and logic operators. """
-        from sympy.core.relational import Eq
         return Or(*[Eq(symbol, elem) for elem in self])
 
     def compare(self, other):
--- a/sympy/solvers/ode/hypergeometric.py
+++ b/sympy/solvers/ode/hypergeometric.py
@@ -46,7 +46,7 @@ def match_2nd_hypergeometric(eq, func):
     r = collect(eq,
         [func.diff(x, 2), func.diff(x), func]).match(deq)
     if r:
-        if not all([r[key].is_polynomial() for key in r]):
+        if not all(val.is_polynomial() for val in r.values()):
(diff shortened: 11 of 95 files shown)


sympy/categories/diagram_drawing.py
```diff
@@ -2155,14 +2155,14 @@ def _check_free_space_vertical(dom_i, cod_i, dom_j, grid):
...
+            free_left = all(not grid[i, dom_j - 1] for i in
+                            range(start, end + 1))
 
         if dom_j == grid.width - 1:
             free_right = True
         else:
-            free_right = all([not grid[i, dom_j + 1] for i in
-                              range(start, end + 1)])
+            free_right = all(not grid[i, dom_j + 1] for i in
+                             range(start, end + 1))
```
oscarbenjamin, line 2165: I suppose it is probably faster to do `not any` rather than `all not` because only one negation is needed.

sympy/combinatorics/homomorphisms.py
```diff
@@ -301,8 +301,8 @@ def homomorphism(domain, codomain, gens, images=[], check=True):
     gens = list(gens)
     images = list(images)
 
-    images.extend([codomain.identity]*(len(generators)-len(images)))
-    gens.extend([g for g in generators if g not in gens])
+    images.extend((codomain.identity,) * (len(generators)-len(images)))
+    gens.extend(g for g in generators if g not in gens)
```
oscarbenjamin, line 305: I'm not sure the second change has a speed improvement

oscarbenjamin: Also if `gens` can be a large list then `g not in gens` is an `O(N)` containment test whereas a set could give `O(1)` containment testing. Potentially that makes the difference between the operation as a whole being `O(N)` or `O(N^2)`.

sympy/combinatorics/perm_groups.py
```diff
@@ -3679,7 +3679,7 @@ def schreier_sims_incremental(self, base=None, gens=None, slp_dict=False):
                 strong_gens_slp[g] = [g]
             return (_base, strong_gens, strong_gens_slp)
 
-        strong_gens.extend([k for k, _ in strong_gens_slp])
+        strong_gens.extend(k for k, _ in strong_gens_slp)
```
oscarbenjamin, line 3682: Not sure this is an improvement

sympy/categories/diagram_drawing.py
oscargus: You are probably correct. I have not considered that type of changes, but can have a quick look.

sympy/combinatorics/permutations.py
```diff
@@ -1153,7 +1153,7 @@ def full_cyclic_form(self):
         """
         need = set(range(self.size)) - set(flatten(self.cyclic_form))
         rv = self.cyclic_form
-        rv.extend([[i] for i in need])
+        rv.extend([i] for i in need)
```
oscarbenjamin, line 1156: Maybe
```
rv = self.cyclic_form + [[i] for i in need]
```

sympy/combinatorics/tensor_can.py
```diff
@@ -1141,11 +1141,11 @@ def _get_bsgs(G, base, gens, free_indices):
         a.extend(ind2)
         a.extend(ind1)
         base_comm.append(ind1[0])
-        a.extend(list(range(ind2[-1] + 1, nr)))
+        a.extend(range(ind2[-1] + 1, nr))
         if sym == 0:
-            a.extend([nr, nr + 1])
+            a.extend((nr, nr + 1))
```
oscarbenjamin, line 1146: Is this sort of thing actually faster?

I get that if the tuple elements are compile-time constants then the tuple can be pre-created and marshalled into the pyc file but if the tuple elements depend on local variables as is the case here then we're just talking about the difference between the time to construct a tuple vs a list:
```python
def f1(l, a):
    for _ in range(100):
        l.extend([a, a+1])

def f2(l, a):
    for _ in range(100):
        l.extend((a, a+1))

def f3(l, a):
    for _ in range(100):
        l += [a, a+1]

def f4(l, a):
    for _ in range(100):
        l.append(a)
        l.append(a+1)
```
Timings:
```python
In [28]: l = []

In [29]: %timeit f1(l, 5)
30 µs ± 105 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [30]: l = []

In [31]: %timeit f2(l, 5)
26.3 µs ± 107 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [32]: l = []

In [33]: %timeit f3(l, 5)
23.1 µs ± 2.55 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [34]: l = []

In [35]: %timeit f4(l, 5)
31.5 µs ± 244 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
Looks like `+=` with a list is fastest.

sympy/geometry/entity.py
```diff
@@ -171,7 +171,7 @@ def _repr_svg_(self):
             # will fall back to the next representation
             return None
 
-        if any([not x.is_number or not x.is_finite for x in bounds]):
+        if any(not x.is_number or not x.is_finite for x in bounds):
```
oscarbenjamin, line 174: Might be clearer and faster as
```python
if not all(x.is_number and x.is_finite for x in bounds):
```
More importantly it might be more correct. What should happen here if `x.is_finite` gives `None`?

sympy/physics/continuum_mechanics/beam.py
```diff
@@ -901,8 +901,8 @@ def max_shear_force(self):
                 val = []
                 for point in points:
                     val.append(shear_curve.subs(x, point))
-                points.extend([singularity[i-1], s])
-                val.extend([limit(shear_curve, x, singularity[i-1], '+'), limit(shear_curve, x, s, '-')])
+                points.extend((singularity[i-1], s))
+                val.extend((limit(shear_curve, x, singularity[i-1], '+'), limit(shear_curve, x, s, '-')))
                 val = list(map(abs, val))
                 max_shear = max(val)
```
oscarbenjamin, line 907: Maybe
```python
val += [
    abs(limit(shear_curve, x, singularity[i-1], '+')),
    abs(limit(shear_curve, x, s, '-'))),
]
```
I guess this isn't tested with symbolic inputs because `max(val)` would probably fail in that case.

sympy/solvers/ode/hypergeometric.py
```diff
@@ -46,7 +46,7 @@ def match_2nd_hypergeometric(eq, func):
     r = collect(eq,
         [func.diff(x, 2), func.diff(x), func]).match(deq)
     if r:
-        if not all([r[key].is_polynomial() for key in r]):
+        if not all(r[key].is_polynomial() for key in r):
```
oscarbenjamin, line 49: Maybe
```python
if not all(val.is_polynomial() for val in r.values()):
```

sympy/stats/frv_types.py
```diff
@@ -785,7 +785,7 @@ def set(self):
 
     @property
     def is_symbolic(self):
-        return not all([self.k.is_number, self.c.is_number, self.delta.is_number])
+        return not all((self.k.is_number, self.c.is_number, self.delta.is_number))
```
oscarbenjamin, line 788:
```python
return not (self.k.is_number and self.c.is_number and self.delta.is_number)
```
or perhaps
```python
return not all(s.is_number for s in (self.k, self.c, self.delta))
```
Not sure if classes in stats have something like `self.parameters` for this.

sympy/stats/tests/test_rv.py
```diff
@@ -214,12 +214,11 @@ def test_Sample():
     if not numpy:
         skip('Numpy is not installed. Abort tests')
     #Test Issue #21563: Output of sample must be a float or array
-    assert isinstance(sample(X), numpy.int64)
+    assert isinstance(sample(X), (numpy.int32, numpy.int64))
```
oscarbenjamin, line 217: Why is this changed?

sympy/tensor/index_methods.py
```diff
@@ -188,7 +188,7 @@ def _get_indices_Add(expr):
     if not non_scalars:
         return set(), {}
 
-    if not all([x == non_scalars[0] for x in non_scalars[1:]]):
+    if not all(x == non_scalars[0] for x in non_scalars[1:]):
```
oscarbenjamin, line 191: Maybe
```
if len(set(non_scalars)) > 1:
```

oscarbenjamin:
The changes to make shortcutting possible are definitely good e.g.
```python
all([x for x in e]) -> all(x for x in e)
```

I'm not sure about changes like
```python
a.extend([y, z]) -> a.extend((y, z))
```
partly just because calling `extend` with a tuple looks weird to me. The fastest version of this according to my timings above is
```python
a += [y, z]
```
which I also think looks nicer and clearer.

sympy/combinatorics/homomorphisms.py
oscargus: You are probably correct. I'll revert those (and see if I find other where a list may help).

sympy/combinatorics/tensor_can.py
oscargus: I didn't consider adding to the list as an option, but it indeed looks faster. I read https://stackoverflow.com/questions/14446128/python-append-vs-extend-efficiency where no one seem to have thought about it either.

sympy/geometry/entity.py
oscargus: I do not analyze the code to that extent... :-) But I do indeed see your point!

sympy/physics/continuum_mechanics/beam.py
oscargus: Yes, plus `val.append(abs(shear_curve.subs(x, point)))` on the lines above.

sympy/stats/tests/test_rv.py
oscargus: For some reason it returned `numpy.int32` on my computer. Not sure why though. I do not think it relates to the changes I've made, so I guess that it may return that on 32-bit machines/numpy/python (which I do not think I have...).

sympy/combinatorics/permutations.py
```diff
@@ -983,7 +983,7 @@ def __new__(cls, *args, size=None, **kwargs):
         temp = set(temp)
 
         if not is_cycle:
-            if any(i not in temp for i in range(len(temp))):
```
oscargus, line 986: Not sure if this actually helps performance, but tends to be more readable in my opinion. It may avoid multiple not, but not sure if `not in` is one or two statements.

sympy/combinatorics/tensor_can.py
oscargus: I guess there is a threshold for when `+=` is faster? Either way, for most cases I kept the tuples and extended.

sympy/categories/diagram_drawing.py
```diff
@@ -2155,14 +2155,14 @@ def _check_free_space_vertical(dom_i, cod_i, dom_j, grid):
         if dom_j == 0:
             free_left = True
         else:
-            free_left = all([not grid[i, dom_j - 1] for i in
-                             range(start, end + 1)])
+            free_left = not all(grid[i, dom_j - 1] for i in
+                                range(start, end + 1))
```
oscarbenjamin, line 2159: Should this not be changed to `any`?

sympy/combinatorics/fp_groups.py
```diff
@@ -259,7 +259,7 @@ def _is_infinite(self):
         used_gens = set()
         for r in self.relators:
             used_gens.update(r.contains_generators())
-        if any([g not in used_gens for g in self.generators]):
+        if not all(g in used_gens for g in self.generators):
```
oscarbenjamin, line 262: I'm not sure what the type of `self.generators` is but `used_gens` is a set and this looks a subset test to me. The most efficient way would be:
```python
if not set(self.generators) <= used_gens:
```
It's not possible to negate the `<=` because this is a partial order.

sympy/solvers/ode/hypergeometric.py
oscargus: One should try to figure out a way to detect this in the code...

sympy/stats/tests/test_rv.py
oscargus: And any integer type should be OK considering that there are 6 options...

sympy/categories/diagram_drawing.py
oscargus: It should!

I'll also try to revert the changes fully in files where there are now only white space changes...

sympy/sets/handlers/intersection.py
```diff
@@ -264,7 +264,7 @@ def intersection_sets(self, other): # noqa:F811
             # not helpful here: multiple parametric solutions.
             if len(solns) == 0:
                 return EmptySet
-            elif any(not isinstance(s, int) and s.free_symbols
+            elif any(s.free_symbols and not isinstance(s, int)
```
oscarbenjamin, line 267: It is possible for `s` to be an `int` here? If it is then `s.free_symbols` would give an attribute error.

My guess is that this is a workaround for a bug that resulted in unsympified types being returned from `diophantine`. I think the `not isinstance(s, int)` should be removed if the tests pass with this change. If there is a failure then I guess that's a bug in diophantine. It should be fixed rather than worked around.

oscargus: You are probably correct. I realized that these changes were not always good when I found the `x.is_Integer and ...`, but forgot about this one. Will not fix diophantine though. :-)

oscargus: Seems like it passes. Should we assume that the bug is fixed or that a proper test case is missing?

sympy/tensor/array/expressions/array_expressions.py
```diff
@@ -44,7 +44,7 @@ def __getitem__(self, item):
         return ArrayElement(self, item)
 
     def as_explicit(self):
-        if any(not isinstance(i, (int, Integer)) for i in self.shape):
+        if not all(isinstance(i, (int, Integer)) for i in self.shape):
```
oscarbenjamin, line 47: This looks like something to be fixed elsewhere. The shape should really be required to be sympified or required to be an `int`. Allowing both is flakey.

sympy/sets/handlers/intersection.py
oscargus: Or rather the diophantine tests passed. Better wait for the other ones as well...

sympy/combinatorics/permutations.py
oscarbenjamin: Since `temp` is a set here this can be something like:
```python
if temp != set(range(len(temp))):
```

sympy/sets/handlers/intersection.py
oscarbenjamin: Something like this is designed to hide bugs. I'd rather remove it so that the bugs can be seen and fixed even if that has to happen in future rather than now.

sympy/categories/diagram_drawing.py
```diff
@@ -670,7 +670,7 @@ def _weld_triangle(tri, welding_edge, fringe, grid, skeleton):
         a = (a[0] + offset[0], a[1] + offset[1])
         b = (b[0] + offset[0], b[1] + offset[1])
 
-        fringe.extend([(a, target_cell), (b, target_cell)])
+        fringe.extend(((a, target_cell), (b, target_cell)))
```
oscarbenjamin, line 673: There are still quite a few changes like this. I think this would be better left unchanged or should be changed to
```python
fringe += [(a, target_cell), (b, target_cell)]
```
since that's the fastest version.

oscargus: Now, there are a few aspects here, just to be clear:
* If the content is compile-time constant?
* More than two terms?
* Some compile-time constants?

sympy/combinatorics/homomorphisms.py
```diff
@@ -301,7 +301,7 @@ def homomorphism(domain, codomain, gens, images=[], check=True):
     gens = list(gens)
     images = list(images)
 
-    images.extend([codomain.identity]*(len(generators)-len(images)))
+    images.extend((codomain.identity,) * (len(generators)-len(images)))
```
oscarbenjamin, line 304: What advantage does a change like this give?

It could alternatively be:
```python
images += [codomain.identity] * (len(generators)-len(images))
```
I think that's potentially fastest and also nicer to read IMO.

oscarbenjamin: There are many other changes like this one

oscargus: Speed, I'd say. But your argument is primarily against variables not constants?

Not sure one can say that += will always be faster?

sympy/combinatorics/tensor_can.py
oscarbenjamin: I think that `+=` with a list is faster when the list of new items is small (e.g. length 2 here). The reason is just that in CPython an operator like `+` is more efficient than a function/method call for e.g. `extend`.

sympy/categories/diagram_drawing.py
oscarbenjamin: If the tuple is entirely a compile time constant then a tuple expression like `(1, 2)` is efficient because it is *fully* precomputed. Otherwise I don't think tuples have any advantage in this situation. Creating a small tuple from a literal tuple expression should be just as fast as creating a small list and both should be just as fast when iterating.

Using `+=` will usually be faster than `extend` because operators are faster than methods in CPython. This will only be significant when extending by a small list.

Using a generator with `extend` is slower on a per-item basis because both `list.extend` and `list.__iadd__` have fast paths for concrete containers like list and tuple that can not be used with a generator.

sympy/combinatorics/homomorphisms.py
oscarbenjamin: Can you show a timing that demonstrates the speed improvement?

oscarbenjamin: Given these examples:
```python
def f1(a):
    l = []
    for _ in range(100):
        l.extend([a] * 100)

def f2(a):
    l = []
    for _ in range(100):
        l.extend((a,) * 100)

def f3(a):
    l = []
    for _ in range(100):
        l += [a] * 100
```
I get the timings
```python
In [22]: %timeit f1(5)
210 µs ± 8.98 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [23]: %timeit f2(5)
248 µs ± 2.79 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [24]: %timeit f3(5)
213 µs ± 11.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
According to that `f2` (using a tuple) is slowest. I'm not sure how well that generalises but it doesn't support making the change here.

Ultimately I think that if we're not sure about a change it's best just to stick with what we have.

sympy/combinatorics/homomorphisms.py
smichr: What about `gens = list(uniq(gens + generators))`?

sympy/stats/tests/test_rv.py
oscargus: I can just confirm that I get this on my current master as well. Must have installed a 32-bit version by mistake?

Anyway, I think it makes sense that this passes even if you have a 32-bit version locally. (Noted that I was running pywin32 although I did install the 64-bit miniconda as far as I (and my web history) can tell...)

sympy/categories/diagram_drawing.py
oscargus: One could argue that creation of tuples is slightly faster, at least in terms of "python operations", and I've seen minor benefits even then. I guess there is a reason why the points are tuples and not lists. (Even though one can also argue that they should be tuples as they do not need to be manipulated.)

I'll get rid of the generator extends I've added, but here I do actually only see benefits (unless we should change all `extend` to `+=` which I will not do currently).

sympy/combinatorics/homomorphisms.py
oscargus: OK! I might have generalized things a bit to far. With constants there is a huge speed-up using tuples, but not using variables.

oscargus:
```
%timeit (x,)*100
621 ns ± 3.82 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit [x,]*100
578 ns ± 7.2 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit [1,]*100
562 ns ± 29.7 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit (1,)*100
14.6 ns ± 0.25 ns per loop (mean ± std. dev. of 7 runs, 100000000 loops each)
```

sympy/stats/tests/test_rv.py
oscarbenjamin: Yes, I think this is fine.

sympy/combinatorics/homomorphisms.py
oscarbenjamin: I think that the huge speedup is only when the whole tuple can be precomputed at compile time.

Otherwise there is no reason in principle why creating a tuple should be faster than creating a list either for small or large sizes.

sympy/tensor/index_methods.py
oscargus:
```
    if len(set(non_scalars)) > 1:
TypeError: unhashable type: 'set'
```

oscargus:
Now I think I've reverted all the extends.

I only changed to tuples where there were constants only.

If it is OK now, I will squash the commits before merging.

sympy/geometry/entity.py
```diff
@@ -171,7 +171,7 @@ def _repr_svg_(self):
             # will fall back to the next representation
             return None
 
-        if any([not x.is_number or not x.is_finite for x in bounds]):
+        if not all(x.is_number and x.is_finite for x in bounds):
             return None
```
oscarbenjamin, line 175: Does this change have an observable effect in the case that `is_finite` is None?

Maybe there could be a test for that...

oscarbenjamin, line 175: Does this change have an observable effect in the case that `is_finite` is None?

Maybe there could be a test for that...

sympy/sets/sets.py
```diff
@@ -867,7 +867,7 @@ def __len__(self):
         return reduce(lambda a, b: a*b, (len(s) for s in self.args))
 
     def __bool__(self):
-        return all([bool(s) for s in self.sets])
+        return all(bool(s) for s in self.sets)
```
oscarbenjamin, line 870: Is `bool` needed inside `all`?

oscarbenjamin: I would have thought this could just be `all(self.sets)`.

oscarbenjamin:
I left some comments but they are not significant.

I think this looks good. You said you wanted to squash so go ahead and squash and merge.

Is there a good reason for squashing? Is it because there are changes that were reverted?

sympy/geometry/entity.py
oscargus: It may indeed be, but since this is printing code, I would assume that we will get some feedback pretty quickly... Right now, I do not really understand the code well enough to try to test it.

sympy/tensor/array/expressions/array_expressions.py
oscargus: Indeed! I will not change it here though.

sympy/sets/sets.py
oscargus: Thanks! Changed it and it seems to work.

oscargus:
> Is it because there are changes that were reverted?

Yes. Thinking it is a hassle if this leads to merge conflicts somewhere else because of reverted changes.

oscargus:
Fixed the `bool` thing and some minor things pointed out by pylint in the sets file (duplicate imports and using `set` as a variable name) as well and rebased. Hopefully this is good to go now.

sympy/combinatorics/tensor_can.py
```diff
@@ -763,12 +763,12 @@ def canonicalize(g, dummies, msym, *v):
     """
     from sympy.combinatorics.testutil import canonicalize_naive
     if not isinstance(msym, list):
-        if not msym in [0, 1, None]:
+        if not msym in (0, 1, None):
```
oscarbenjamin, line 766: Not really important but I think `x not in y` is generally better for clarity then `not x in y` and also it's very slightly faster.

oscarbenjamin:
Thanks

PR #21879, opened 2021-08-15
quaternion - to_axis_angle supports negative rotations

### LINKED ISSUE #21398: Quaternion.to_axis_angle doesn't respect negative rotations
Is there any reason to multiply the quaternion by -1 if the first term is negative? It should support negative rotations (and angles) instead and remove the multiplication.

praneethratna:
Fixes #21398 and now to_axis_angle function can give Rotations in either way and the angle can be negative.
`Quaternion(-1,1,1,1)` returns an angle of `4 * pie/3` instead of `2*pie/3`

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/algebras/quaternion.py
+++ b/sympy/algebras/quaternion.py
@@ -651,8 +651,6 @@ class Quaternion(Expr):
 
         """
         q = self
-        if q.a.is_negative:
-            q = q * -1
 
         q = q.normalize()
         angle = trigsimp(2 * acos(q.a))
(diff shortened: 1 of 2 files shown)


oscargus:
Since the code is identical to #21397 it would make sense to base the PR on that.

Also, it would make sense to add a test that uses a negative angle.

praneethratna:
> Since the code is identical to #21397 it would make sense to base the PR on that.
> 
> Also, it would make sense to add a test that uses a negative angle.

I changed the to_axis_angle test in test_issue_16318(), will that be a sufficient test for this issue?

oscargus:
To me it is not clear that the test shows that "to_axis_angle supports negative rotations" as both the axis and the angle are positive in that test. I would expect a test where it is more clearly shown that there are negative rotations supported.

While I can imagine that the two different rotations (before and after the fix) are the same, it would be nice to have some more explicit tests. Could you use the axis-angle-pair and perform a rotation of another Quaternion and illustrate that this is the same as multiplying(?) them? Or recreate the same Quaternion from them?

PR #21880, opened 2021-08-15
maint(ci): add Python 3.10 jobs to CI tests

oscarbenjamin:
Adds Python 3.10 to the list of Python versions tested in CI.

Python 3.10 is not released yet but will be soon.

NO ENTRY

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.10'
+          python-version: '3.10.0-rc.1'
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
This will need to be updated when the final release of Python 3.10 is out

asmeurer:
It seems this would also need to be updated when the next release candidate is released. Is there no way to do 3.10-dev like you could with Travis?

oscarbenjamin:
According to this:
https://github.com/actions/setup-python#usage

You can do something like:
```
- uses: actions/setup-python@v2
  with:
    python-version: '3.9.0-alpha - 3.9.0' # SemVer's version range syntax
```
I guess you could make that `3.9.0-alpha - 3.9.x`.

Issue #17186, opened 2019-07-13
Solving already solved equations takes a long time


drhagen:
When solving a system of nonlinear equations, if the equations are already solved for all unknowns, then it should be very fast to return the solution. However, this is not the case. For example, this script:

```python
import time
from sympy import *

var('kn v ah ldkd_1 ldkd_2 eh_1 lc_1 lc_2 rec_1 rec_2')
var('lf_1 lf_2 rh_1 rh_2 sc_1 sc_2 sh_1 sh_2 lrkd_1 lrkd_2')
var('dv_1 dv_2 d m')

dependent_vars = (var('ka kl_D1 kn2_T1 kf_D1T1 kf_L1R1 kl_R1 kl_L1R1 kl_L1 kl_S1 t_R1')
+ var('L1_0 t_S1 t_R1 L1_R1_0 R1_0 L1_S1_0 S1_0 kS_1 ky_L1 ky_R1 a_1_0')
+ var('kn2_T2 kf_D1T2 kf_L2R2 kl_R2 kl_L2R2 kl_R2 kl_L2R2 kl_L2 kl_S2 t_R2')
+ var('L2_0 t_S2 L2_R2_0 R2_0 L2_S2_0 S2_0 kS_2 ky_L2 ky_R2 a_2_0'))


eqs = [
    Eq(ka, log(2) / (ah * d)),
    Eq(kl_D1, log(2) / (eh_1 * d)),
    Eq(kn2_T1, floor(dv_1 / 2) * kn),
    Eq(kf_D1T1, ldkd_1 * kn),
    Eq(kf_L1R1, lrkd_1 * kn),
    Eq(kl_R1, log(2) / (rh_1 * m)),
    Eq(kl_L1R1, kl_R1),
    Eq(kl_L1, log(2) / (lf_1 * m)),
    Eq(kl_S1, log(2) / (sh_1 * m)),
    Eq(t_R1, rec_1 * v),
    Eq(L1_0, lc_1 * v),
    Eq(t_S1, sc_1 * v),
    Eq(L1_R1_0, (kn * L1_0 * t_R1) / (v * (kf_L1R1 + kl_L1R1 + kn * L1_0 / v))),
    Eq(R1_0, t_R1 - L1_R1_0),
    Eq(L1_S1_0, (kn * L1_0 * t_S1) / (v * (kf_L1R1 + kl_L1 + kn * L1_0 / v))),
    Eq(S1_0, t_S1 - L1_S1_0),
[...]

PR #21883, opened 2021-08-16
fix(solvers): split separable systems of equations properly

### LINKED ISSUE #21882: Incorrect solutions given by `solve`
The following code tries to find the solutions of a system of (multivariate quadratic) equations. However, the single solution returned by `solve` is not correct.
It seems like `f = 0` should be part of the solution but is not.

```python
from sympy import solve, sympify

equations = [
    "-k*a + b + 5*f/6 + 2*c/9 + 5*d/6 + 4*a/3",
    "-k*f + 4*f/3 + d/2",
    "-k*d + f/6 + d",
    "13*b/18 + 13*c/18 + 13*a/18",
    "-k*c + b/2 + 20*c/9 + a",
    "-k*b + b + c/18 + a/6",
    "5*b/3 + c/3 + a",
    "2*b/3 + 2*c + 4*a/3",
    "-g"
]
equations = [sympify(e) for e in equations]
unknowns = {s for e in equations for s in e.free_symbols}

solutions = solve(equations, unknowns, dict=True)
# Should all be zero but are not:
for solution in solutions:
    for e in equations:
        print(e.subs(solution).simplify())

```

Output:
```
0
f*(-12*k**2 + 28*k - 15)/(12*(k - 1))
0
65*f*(5 - 6*k)/(108*(3*k**2 - 5*k + 2))
5*f*(6*k - 5)*(9*k - 17)/(36*(k - 1)*(3*k - 2))
0
0
35*f*(5 - 6*k)/(18*(3*k**2 - 5*k + 2))
0
```

Expected Output:
```
0
0
0
0
0
0
0
0
0
```

The bug exists in sympy versions `1.8` and `1.7` but NOT in version `1.6`.

In the meantime, is there a way to circumvent this problem until the issue is fixed? We thought about calling `solve` again on the equations of the output but are not sure if that is the best way.

oscarbenjamin:

Fixes #21882

#### Release Notes

* solvers
    * A bug when splitting separable systems of equations in solve was fixed.

### DIFF
--- a/doc/src/tutorial/solvers.rst
+++ b/doc/src/tutorial/solvers.rst
@@ -142,9 +142,9 @@ In the ``solveset`` module, the non linear system of equations is solved using
    ``solve`` can be used for such cases:
 
    >>> solve([x**2 - y**2/exp(x)], [x, y], dict=True)
-   ⎡⎧      ⎛-y ⎞⎫  ⎧      ⎛y⎞⎫⎤
-   ⎢⎨x: 2⋅W⎜───⎟⎬, ⎨x: 2⋅W⎜─⎟⎬⎥
-   ⎣⎩      ⎝ 2 ⎠⎭  ⎩      ⎝2⎠⎭⎦
+   ⎡⎧         ____⎫  ⎧        ____⎫⎤
+   ⎢⎨        ╱  x ⎬  ⎨       ╱  x ⎬⎥
+   ⎣⎩y: -x⋅╲╱  ℯ  ⎭, ⎩y: x⋅╲╱  ℯ  ⎭⎦
 
    3. Currently ``nonlinsolve`` is not properly capable of solving the system of equations
    having trigonometric functions.
--- a/doc/src/tutorial/solvers.rst
+++ b/doc/src/tutorial/solvers.rst
@@ -145,6 +145,10 @@ In the ``solveset`` module, the non linear system of equations is solved using
    ⎡⎧         ____⎫  ⎧        ____⎫⎤
    ⎢⎨        ╱  x ⎬  ⎨       ╱  x ⎬⎥
    ⎣⎩y: -x⋅╲╱  ℯ  ⎭, ⎩y: x⋅╲╱  ℯ  ⎭⎦
+   >>> solve(x**2 - y**2/exp(x), x, dict=True)
+   ⎡⎧      ⎛-y ⎞⎫  ⎧      ⎛y⎞⎫⎤
+   ⎢⎨x: 2⋅W⎜───⎟⎬, ⎨x: 2⋅W⎜─⎟⎬⎥
+   ⎣⎩      ⎝ 2 ⎠⎭  ⎩      ⎝2⎠⎭⎦
 
    3. Currently ``nonlinsolve`` is not properly capable of solving the system of equations
    having trigonometric functions.
--- a/sympy/solvers/solvers.py
+++ b/sympy/solvers/solvers.py
@@ -1758,17 +1758,6 @@ def _solve_system(exprs, symbols, **flags):
                 for e in subexpr:
                     subsyms |= exprsyms[e]
                 subsyms = list(sorted(subsyms, key = lambda x: sym_indices[x]))
-                # use canonical subset to solve these equations
-                # since there may be redundant equations in the set:
-                # take the first equation of several that may have the
-                # same sub-maximal free symbols of interest; the
-                # other equations that weren't used should be checked
-                # to see that they did not fail -- does the solver
-                # take care of that?
-                choices = sift(subexpr, lambda x: tuple(ordered(exprsyms[x])))
-                subexpr = choices.pop(tuple(ordered(subsyms)), [])
-                for k in choices:
-                    subexpr.append(next(ordered(choices[k])))
                 flags['_split'] = False  # skip split step
                 subsol = _solve_system(subexpr, subsyms, **flags)
                 if not isinstance(subsol, list):
--- a/sympy/solvers/solvers.py
+++ b/sympy/solvers/solvers.py
@@ -1951,7 +1951,16 @@ def _solve_system(exprs, symbols, **flags):
                             rnew[k] = v.subs(s, sol)
                         # and add this new solution
                         rnew[s] = sol
-                        newresult.append(rnew)
+                        # check that it is independent of previous solutions
+                        iset = set(rnew.items())
+                        for i in newresult:
+                            if len(i) < len(iset) and not set(i.items()) - iset:
+                                # this is a superset of a known solution that
+                                # is smaller
+                                break
+                        else:
+                            # keep it
+                            newresult.append(rnew)
                     hit = True
                     got_s.add(s)
                 if not hit:
--- a/sympy/solvers/solvers.py
+++ b/sympy/solvers/solvers.py
@@ -1884,9 +1884,22 @@ def _solve_system(exprs, symbols, **flags):
         # be solved.
         def _ok_syms(e, sort=False):
             rv = (e.free_symbols - solved_syms) & legal
+
+            # Solve first for symbols that have lower degree in the equation.
+            # Ideally we want to solve firstly for symbols that appear linearly
+            # with rational coefficients e.g. if e = x*y + z then we should
+            # solve for z first.
+            def key(sym):
+                ep = e.as_poly(sym)
+                if ep is None:
+                    complexity = (S.Infinity, S.Infinity, S.Infinity)
+                else:
+                    coeff_syms = ep.LC().free_symbols
+                    complexity = (ep.degree(), len(coeff_syms & rv), len(coeff_syms))
+                return complexity + (default_sort_key(sym),)
+
             if sort:
-                rv = list(rv)
-                rv.sort(key=default_sort_key)
+                rv = sorted(rv, key=key)
             return rv
 
         solved_syms = set(solved_syms)  # set of symbols we have solved for
--- a/sympy/solvers/solvers.py
+++ b/sympy/solvers/solvers.py
@@ -51,7 +51,7 @@ from sympy.functions.elementary.piecewise import piecewise_fold, Piecewise
 from sympy.utilities.lambdify import lambdify
 from sympy.utilities.misc import filldedent
 from sympy.utilities.iterables import (cartes, connected_components,
-    generate_bell, uniq, sift)
+    generate_bell, uniq)
 from sympy.utilities.decorator import conserve_mpmath_dps
 
 from mpmath import findroot
--- a/sympy/solvers/tests/test_solvers.py
+++ b/sympy/solvers/tests/test_solvers.py
@@ -1402,7 +1402,6 @@ def test_issue_21882():
         {a: 0, f: 0, b: 0, d: 0, c: 0, g: 0},
         {a: 0, f: -d, b: 0, k: S(5)/6, c: 0, g: 0},
         {a: -2*c, f: 0, b: c, d: 0, k: S(13)/18, g: 0},
-        {a: 0, f: 0, b: 0, k: S(3)/2, c: 0, d: 0, g: 0}
     ]
 
     assert solve(equations, unknowns, dict=True) == answer
--- a/sympy/solvers/tests/test_solvers.py
+++ b/sympy/solvers/tests/test_solvers.py
@@ -1316,6 +1316,11 @@ def test_issue_5114_solvers():
 
 
 def test_issue_5849():
+    #
+    # XXX: This system does not have a solution for most values of the
+    # parameters. Generally solve returns the empty set for systems that are
+    # generically inconsistent.
+    #
     I1, I2, I3, I4, I5, I6 = symbols('I1:7')
     dI1, dI4, dQ2, dQ4, Q2, Q4 = symbols('dI1,dI4,dQ2,dQ4,Q2,Q4')
 
@@ -1332,21 +1337,24 @@ def test_issue_5849():
     )
 
     ans = [{
-    I1: I2 + I6,
-    dI1: -4*I2 - 4*I3 - 4*I5 - 10*I6 + 24,
-    I4: -I5 + I6,
-    dQ4: -I5 + I6,
-    Q4: 3*I5/2 - I6/2 - dI4/2,
+    I1: I2 + I3,
+    dI1: -4*I2 - 8*I3 - 4*I5 - 6*I6 + 24,
+    I4: I3 - I5,
+    dQ4: I3 - I5,
+    Q4: -I3/2 + 3*I5/2 - dI4/2,
     dQ2: I2,
     Q2: 2*I3 + 2*I5 + 3*I6}]
 
     v = I1, I4, Q2, Q4, dI1, dI4, dQ2, dQ4
     assert solve(e, *v, manual=True, check=False, dict=True) == ans
-    assert solve(e, *v, manual=True) == ans[0]
+    assert solve(e, *v, manual=True, check=False) == ans[0]
+    assert solve(e, *v, manual=True) == []
+    assert solve(e, *v) == []
+
     # the matrix solver (tested below) doesn't like this because it produces
     # a zero row in the matrix. Is this related to issue 4551?
     assert [ei.subs(
-        ans[0]) for ei in e] == [-I3 + I6, I3 - I6, 0, 0, 0, 0, 0, 0, 0]
+        ans[0]) for ei in e] == [0, 0, I3 - I6, -I3 + I6, 0, 0, 0, 0, 0]
 
 
 def test_issue_5849_matrix():
@@ -1373,14 +1381,34 @@ def test_issue_5849_matrix():
         2*I3 + 2*I5 + 3*I6 - Q2,
         I4 - 2*I5 + 2*Q4 + dI4
     )
-    assert solve(e, I1, I4, Q2, Q4, dI1, dI4, dQ2, dQ4) == {
-    I1: I2 + I6,
-    dI1: -4*I2 - 4*I3 - 4*I5 - 10*I6 + 24,
-    I4: -I5 + I6,
-    dQ4: -I5 + I6,
-    Q4: 3*I5/2 - I6/2 - dI4/2,
-    dQ2: I2,
-    Q2: 2*I3 + 2*I5 + 3*I6}
+    assert solve(e, I1, I4, Q2, Q4, dI1, dI4, dQ2, dQ4) == []
+
+
+def test_issue_21882():
+
+    a, b, c, d, f, g, k = unknowns = symbols('a, b, c, d, f, g, k')
+
+    equations = [
+        -k*a + b + 5*f/6 + 2*c/9 + 5*d/6 + 4*a/3,
+        -k*f + 4*f/3 + d/2,
+        -k*d + f/6 + d,
+        13*b/18 + 13*c/18 + 13*a/18,
+        -k*c + b/2 + 20*c/9 + a,
+        -k*b + b + c/18 + a/6,
+        5*b/3 + c/3 + a,
+        2*b/3 + 2*c + 4*a/3,
+        -g,
+    ]
+
+    answer = [
+        {a: 0, d: 0, b: 0, f: 0, c: 0, g: 0},
+        {a: 0, d: -f, b: 0, k: S(5)/6, c: 0, g: 0},
+        {a: -2*c, d: 0, b: c, f: 0, k: S(13)/18, g: 0},
+        {a: 0, d: 0, b: 0, k: S(3)/2, c: 0, f: 0, g: 0}
+    ]
+
+    assert solve(equations, unknowns, dict=True) == answer
+
 
 
 def test_issue_5901():
--- a/sympy/solvers/tests/test_solvers.py
+++ b/sympy/solvers/tests/test_solvers.py
@@ -1410,7 +1410,6 @@ def test_issue_21882():
     assert solve(equations, unknowns, dict=True) == answer
 
 
-
 def test_issue_5901():
     f, g, h = map(Function, 'fgh')
     a = Symbol('a')
--- a/sympy/solvers/tests/test_solvers.py
+++ b/sympy/solvers/tests/test_solvers.py
@@ -870,12 +870,12 @@ def test_issue_5132():
         {(log(-sin(2)), -S(2)), (log(sin(2)), S(2))}
     eqs = [exp(x)**2 - sin(y) + z**2, 1/exp(y) - 3]
     assert solve(eqs, set=True) == \
-        ([x, y], {
-        (log(-sqrt(-z**2 - sin(log(3)))), -log(3)),
-        (log(-z**2 - sin(log(3)))/2, -log(3))})
+        ([y, z], {
+        (-log(3), sqrt(-exp(2*x) - sin(log(3)))),
+        (-log(3), -sqrt(-exp(2*x) - sin(log(3))))})
     assert solve(eqs, x, z, set=True) == (
         [x, z],
-        {(log(-z**2 + sin(y))/2, z), (log(-sqrt(-z**2 + sin(y))), z)})
+        {(x, sqrt(-exp(2*x) + sin(y))), (x, -sqrt(-exp(2*x) + sin(y)))})
     assert set(solve(eqs, x, y)) == \
         {
             (log(-sqrt(-z**2 - sin(log(3)))), -log(3)),
@@ -885,12 +885,10 @@ def test_issue_5132():
             (-log(3), -sqrt(-exp(2*x) - sin(log(3)))),
         (-log(3), sqrt(-exp(2*x) - sin(log(3))))}
     eqs = [exp(x)**2 - sin(y) + z, 1/exp(y) - 3]
-    assert solve(eqs, set=True) == ([x, y], {
-        (log(-sqrt(-z - sin(log(3)))), -log(3)),
-            (log(-z - sin(log(3)))/2, -log(3))})
+    assert solve(eqs, set=True) == ([y, z], {
+        (-log(3), -exp(2*x) - sin(log(3)))})
     assert solve(eqs, x, z, set=True) == (
-        [x, z],
-        {(log(-sqrt(-z + sin(y))), z), (log(-z + sin(y))/2, z)})
+        [x, z], {(x, -exp(2*x) + sin(y))})
     assert set(solve(eqs, x, y)) == {
             (log(-sqrt(-z - sin(log(3)))), -log(3)),
             (log(-z - sin(log(3)))/2, -log(3))}
@@ -1401,10 +1399,10 @@ def test_issue_21882():
     ]
 
     answer = [
-        {a: 0, d: 0, b: 0, f: 0, c: 0, g: 0},
-        {a: 0, d: -f, b: 0, k: S(5)/6, c: 0, g: 0},
-        {a: -2*c, d: 0, b: c, f: 0, k: S(13)/18, g: 0},
-        {a: 0, d: 0, b: 0, k: S(3)/2, c: 0, f: 0, g: 0}
+        {a: 0, f: 0, b: 0, d: 0, c: 0, g: 0},
+        {a: 0, f: -d, b: 0, k: S(5)/6, c: 0, g: 0},
+        {a: -2*c, f: 0, b: c, d: 0, k: S(13)/18, g: 0},
+        {a: 0, f: 0, b: 0, k: S(3)/2, c: 0, d: 0, g: 0}
     ]
 
     assert solve(equations, unknowns, dict=True) == answer


sympy/solvers/solvers.py
```diff
@@ -1758,17 +1758,6 @@ def _solve_system(exprs, symbols, **flags):
                 for e in subexpr:
                     subsyms |= exprsyms[e]
                 subsyms = list(sorted(subsyms, key = lambda x: sym_indices[x]))
-                # use canonical subset to solve these equations
```
smichr, line 1761: How is this situation being handled now?

oscarbenjamin: I don't understand what situation the deleted code handles.

The code in this block is intended to split a system that can be solved as smaller independent systems e.g. if we have:
```
x + y
x - y - 1
w + t
w - t - 1
```
then it's clear that we can solve the first two equations for x and y separately from the other two equations which can be solved for w and t.

That's the purpose of this block. What I don't understand is the purpose of the code that is deleted in this PR...

smichr: What happens if you have
```
solve((
x + y,
1/x + 1/y,
x - y - 2))
```

oscarbenjamin: For that particular example the result is the same in both PR and master:
```
In [1]: solve((
   ...: x + y,
   ...: 1/x + 1/y,
   ...: x - y - 2))
Out[1]: {x: 1, y: -1}
```

smichr: if this gives `[]` then I it seems that all is ok:
```
solve((
    x + y,
    x*y,
    x - y - 2))
```

oscarbenjamin: Yes, that is the same on master and in the PR:
```python
In [1]: solve((
   ...:     x + y,
   ...:     x*y,
   ...:     x - y - 2))
Out[1]: []
```

oscarbenjamin: @smichr are you happy with this change then?

The PR is not ready to be merged because I'll add some more tests both from #21882 and the examples you've just listed here.

doc/src/tutorial/solvers.rst
```diff
@@ -142,9 +142,9 @@ In the ``solveset`` module, the non linear system of equations is solved using
    ``solve`` can be used for such cases:
 
    >>> solve([x**2 - y**2/exp(x)], [x, y], dict=True)
-   ⎡⎧      ⎛-y ⎞⎫  ⎧      ⎛y⎞⎫⎤
-   ⎢⎨x: 2⋅W⎜───⎟⎬, ⎨x: 2⋅W⎜─⎟⎬⎥
-   ⎣⎩      ⎝ 2 ⎠⎭  ⎩      ⎝2⎠⎭⎦
+   ⎡⎧         ____⎫  ⎧        ____⎫⎤
+   ⎢⎨        ╱  x ⎬  ⎨       ╱  x ⎬⎥
+   ⎣⎩y: -x⋅╲╱  ℯ  ⎭, ⎩y: x⋅╲╱  ℯ  ⎭⎦
```
oscarbenjamin, line 147: While I do think this is an improvement in the output of `solve` this change does not make sense in this docstring so a different example should be used here.

smichr: With this PR, I don't think you will ever get a solution in terms of `LambertW` anymore since `y` (in this sort of equation) will have lower complexity than `x`. So the example can be removed.

oscarbenjamin: The example could be changed to solve for `x` only.

sympy/solvers/solvers.py
smichr: > are you happy with this change then

I don't see any issues. I like the idea of avoiding solutions that require division by expressions with free symbols. I would like to see the direction (eventually) that solves #17186. Splitting the linear from the non-linear before diving into the non-linear part is important.

oscarbenjamin: Yes, I want to have a think about that before merging this change. I think that something like strongly connected components could be used somehow but I haven't quite figured out the right way to pose the system as a directed graph.

doc/src/tutorial/solvers.rst
smichr: doh...good point

smichr:
It's too late to think much on this now, but this is what I was trying in terms of removing linear parts. This just uses recursion and removal of linearly solvable symbols (taking the simplest solution at each step for any variable that can be solve).

```python
def repsort(*replace):
    """Return sorted replacement tuples `(o, n)` such that `(o_i, n_i)` will appear before
    `(o_j, n_j)` if `o_j` appears in `n_i`. An error will be raised if o_j appears in n_i and o_i
    appears in n_k if k >= i.

    Examples
    ========
    >>> repsort((x, y + 1), (z, x + 2))
    [(z, x + 2), (x, y + 1)]
    >>> repsort((x, y + 1), (z, x**2))
    [(z, x**2), (x, y + 1)]

    Any two of the following 3 tuples will not raise an error,
    but together they contain a cycle that raises an error:

    >>> repsort((x, y), (y, z), (z, x))
    ...
    raise ValueError("cycle detected")
    """
    from itertools import permutations
    from sympy import default_sort_key, topological_sort
    edges = [(i, j) for i, j in permutations(replace, 2) if 
        i[1].has(j[0]) and (not j[0].is_Symbol or j[0] in i[1].free_symbols)]
    return topological_sort([replace, edges], default_sort_key)

def linreduce(eqs):
    from collections import defaultdict
    psol = {}
    while 1:
        d = dict(sift(eqs, lambda x: solve_linear(x, 0)))
        bysym = defaultdict(list)
        for di in d:
            if di[0].is_Symbol and not di[1].has(di[0]):
                bysym[di[0]].append(di[1])
        if not bysym:
            try:
                r = repsort(*[i for i in psol.items()])
                sol = {}
                for i in range(len(r)-1,-1,-1):
                    k, v = r[i]
                    sol[k] = v.xreplace(sol)
                return sol, eqs
            except:
                return psol, eqs
        for k in bysym:
            bysym[k] = next(ordered(bysym[k]))
            psol[k] = bysym[k]
        eqs = [i.subs(bysym) for i in eqs]

>>> eqs
[-a*k + 4*a/3 + b + 2*c/9 + 5*d/6 + 5*f/6, d/2 - f*k + 4*f/3, -d*k + d
+ f/6, 13*a/18 + 13*b/18 + 13*c/18, a + b/2 - c*k + 20*c/9, a/6 - b*k
+ b + c/18, a + 5*b/3 + c/3, 4*a/3 + 2*b/3 + 2*c, -g]


>>> linreduce(eqs)
({g: 0, f: 0, d: 0, c: 0, b: 0, a: 0}, [0, 0, 0, 0, 0, 0, 0, 0, 0])

==> solution is all 0s and k arbitrary

===============================================================
>>> eqs = (
...         I1 - I2 - I3,
...         I3 - I4 - I5,
...         I4 + I5 - I6,
...         -I1 + I2 + I6,
...         -2*I1 - 2*I3 - 2*I5 - 3*I6 - dI1/2 + 12,
...         -I4 + dQ4,
...         -I2 + dQ2,
...         2*I3 + 2*I5 + 3*I6 - Q2,
...         I4 - 2*I5 + 2*Q4 + dI4
...     )
>>> linreduce(ieqs)
({Q4: -dI1/18 - dI4/2 - 2*dQ2/9 - 23*dQ4/18 + 4/3, Q2: -7*dI1/18 -
14*dQ2/9 - 4*dQ4/9 + 28/3, I6: -dI1/18 - 2*dQ2/9 + 2*dQ4/9 + 4/3, I5:
-dI1/18 - 2*dQ2/9 - 7*dQ4/9 + 4/3, I4: dQ4, I3: -dI1/18 - 2*dQ2/9 +
2*dQ4/9 + 4/3, I2: dQ2, I1: -dI1/18 + 7*dQ2/9 + 2*dQ4/9 + 4/3}, [0, 0, 0, 0, 0, 0, 0, 0, 0])

==> solution is these and dQ2, dI4, dQ4, dI1 are arbitrary

================================================================
>>> linreduce((x+y,x**2+y**2-4))
({x: -y}, [0, 2*y**2 - 4])

==> so x = -y and 2*y**2 - 4 must be solved before x,y pairs in solution are found.
```

sympy/solvers/solvers.py
smichr: > something like strongly connected components

There is potentially a related idea [here](https://github.com/sympy/sympy/issues/6662#issuecomment-895382853) in what Tim Peters was directing me towards as a means of doing a derangement efficiently.

oscarbenjamin:
> This just uses recursion and removal of linearly solvable symbols

That's quite close to what I was thinking. The primary difference is that we should find the strongly connected components of that graph rather than using a topological sort. Cycles are okay as they just imply a subsystem that needs to be solved together. Also we should try to avoid the quadratic loop over `permutations(equations, 2)` and we should compute this once for the whole system rather than again and again for each equation if possible.

We define a graphs as follows:

1. Each vertex represents an equation.
2. If `eq_i` can be trivially solved for an unknown `sym_k` which also appears in `eq_j` then there is an edge from `eq_i` to `eq_j`.

So given the system:
```
(0) -a*k + 4*a/3 + b + 2*c/9 + 5*d/6 + 5*f/6
(1) d/2 - f*k + 4*f/3
(2) -d*k + d + f/6
(3) 13*a/18 + 13*b/18 + 13*c/18
(4) a + b/2 - c*k + 20*c/9
(5) a/6 - b*k + b + c/18
(6) a + 5*b/3 + c/3
(7) 4*a/3 + 2*b/3 + 2*c
(8) -g
```
We have a graph like:
```python
graph = {
    0: {1, 2, 3, 4, 5, 6, 7} , # b, c, d, f
    1: {0, 2}, # d
    2: {0, 1}, # f
    3: {0, 4, 5, 6, 7}, # a, b, c
    4: {0, 3, 5, 6, 7}, # a, b
    5: {0, 3, 4, 6, 7}, # a, c
    6: {0, 3, 4, 5, 7}, # a, b, c
    7: {0, 3, 4, 5, 6}, # a, b, c
    8: {}, # g
}
```
Okay, hm that's not quite right as the SCC would be too big. I think that the approach I described above would work for #17186 but something more is needed to further break down systems like this one. The definition of when there is an edge needs to be tightened somehow.

smichr:
> definition of when there is an edge needs to be tightened somehow.

could it be "if the coefficient of sym in equations ei and ej is constant (or are independent) then the equations are connected"?

so `x*y + 1` and `(x + 1)*(y - 2) + z` are not connected but `x*z + 1` and `(x + 1)*(y - 2) + z` are.

oscarbenjamin:
> > definition of when there is an edge needs to be tightened somehow.
> 
> could it be "if the coefficient of sym in equations ei and ej is constant (or are independent) then the equations are connected"?

I want an asymmetric definition so that it's a directed graph that can be sorted topologically like in your code above. The idea is to identify where one equation can be solved in its own before substituting into another. The example from #17186 is (with the unknowns renamed as `xi`:
```python
Eq(x0, log(2)/(ah*d))
Eq(x1, log(2)/(d*eh_1))
Eq(x2, kn*floor(dv_1/2))
Eq(x3, kn*ldkd_1)
Eq(x4, kn*lrkd_1)
Eq(x5, log(2)/(m*rh_1))
Eq(x6, x5)
Eq(x7, log(2)/(lf_1*m))
Eq(x8, log(2)/(m*sh_1))
Eq(x9, rec_1*v)
Eq(x10, lc_1*v)
Eq(x11, sc_1*v)
Eq(x12, kn*x10*x9/(v*(kn*x10/v + x4 + x6)))
Eq(x13, -x12 + x9)
Eq(x14, kn*x10*x11/(v*(kn*x10/v + x4 + x7)))
Eq(x15, x11 - x14)
Eq(x16, (kn*x10*x15/v - x14*x4 + x15*x8)/x13)
Eq(x17, kn*x10*x13/v + kn*x10*x15/v + x10*x7 - x12*x4 - x14*x4)
Eq(x18, kn*x10*x13/v - x12*x4 + x13*x16 + x13*x5)
Eq(x19, x12)
Eq(x20, kn*floor(dv_2/2))
Eq(x21, kn*ldkd_2)
Eq(x22, kn*lrkd_2)
Eq(x23, log(2)/(m*rh_2))
Eq(x24, x23)
Eq(x25, log(2)/(lf_2*m))
Eq(x26, log(2)/(m*sh_2))
Eq(x27, rec_2*v)
Eq(x28, lc_2*v)
Eq(x29, sc_2*v)
Eq(x30, kn*x27*x28/(v*(kn*x28/v + x22 + x24)))
Eq(x31, x27 - x30)
Eq(x32, kn*x28*x29/(v*(kn*x28/v + x22 + x25)))
Eq(x33, x29 - x32)
Eq(x34, (kn*x28*x33/v - x22*x32 + x26*x33)/x31)
Eq(x35, kn*x28*x31/v + kn*x28*x33/v - x22*x30 - x22*x32 + x25*x28)
Eq(x36, kn*x28*x31/v - x22*x30 + x23*x31 + x31*x34)
Eq(x37, x30)
```
This is already in topological order but in general we want to be able to identify a topological order so we see that e.g. the equation for `x5` can be solved first before substituting into the equation for `x6`.

Perhaps for that case the situation is quite simple... as a first pass any equation trivially solvable for only a single unknown with no other unknowns involved can be solved and those results can be substituted into everything else.

But then same thing can happen with more than one equation e.g.:
```
(1) x + y
(2) x - y
(3) z + x + y + t
(4) t + x + z
```
Here (1) and (2) can be solved for x and y first before substituting into (3) and (4). I'm trying to think how to define a graph such that (1) and (2) would be a connected component and (3) and (4) would be another.

oscarbenjamin:
The main limitation of the changes in this PR is really just that it doesn't sort the equations again after some symbols have been solved. Perhaps we don't need to overthink this so much and should just solve that problem. I think your example code does that.

smichr:
I wonder if `pairwise_most_common` (from #11232) applied to sets of free symbols from each expression that appear with constant coefficients (i.e. do not contain any symbols of interest) in an expression.

```python
>>> eqs
[-a*k + 4*a/3 + b + 2*c/9 + 5*d/6 + 5*f/6, 
d/2 - f*k + 4*f/3, 
-d*k + d + f/6, 
13*a/18 + 13*b/18 + 13*c/18, 
a + b/2 - c*k + 20*c/9, 
a/6 - b*k + b + c/18, 
a + 5*b/3 + c/3, 4*a/3 + 2*b/3 + 2*c, 
-g]
>>> [set([i for i in e.free_symbols if e.diff(i).is_number])  for e in eqs]
[{f, d, b, c}, {d}, {f}, {a, b, c}, {b, a}, {a, c}, {a, b, c}, {b, a, c}, {g}]
>>> pairwise_most_common(_)
[({b, a, c}, [(3, 6), (3, 7), (6, 7)])]
```
So we can solve single equations for `d, f, g`:
>>> solve((eqs[1],eqs[2],eqs[-1]),d,f,g)
{d: 0, f: 0, g: 0}
>>> sol = _
and 3,6,7 for `a,b,c`
>>> solve((eqs[3],eqs[7],eqs[6]),a,b,c)
{a: -2*c, b: c}
>>> sol.update(_)
>>> [i.subs(sol).simplify() for i in eqs]
[c*(18*k - 13)/9, 0, 0, 0, c*(13 - 18*k)/18, c*(13 - 18*k)/18, 0, 0, 0]
>>> solve(_)
[{c: 0}, {k: 13/18}]
```
So the solution is `{d: 0, f: 0, g: 0, a: -2*c, b: c}` and `c=0 | k = 13/18`.

**Comment by smichr** [comment]
Here is as far as I've gotten on this. See what you think:

<details><summary>POC</summary>
<p>

```python
def repsort(*replace):
    """Return sorted replacement tuples ``(o, n)`` such that
    ``(o_i, n_i)`` will appear before ``(o_j, n_j)`` if ``o_j``
    appears in ``n_i``. An error will be raised if ``o_j``
    appears in ``n_i`` and ``o_i`` appears in ``n_k`` if ``k >= i``.

    Examples
    ========

    >>> repsort((x, y + 1), (z, x + 2))
    [(z, x + 2), (x, y + 1)]
    >>> repsort((x, y + 1), (z, x**2))
    [(z, x**2), (x, y + 1)]

    Any two of the following 3 tuples will not raise an error,
    but together they contain a cycle that raises an error:

    >>> repsort((x, y), (y, z), (z, x))
    ...
    raise ValueError("cycle detected")
    """
    from itertools import permutations
    from sympy import default_sort_key, topological_sort
    edges = [(i, j) for i, j in permutations(replace, 2) if 
        i[1].has(j[0]) and (not j[0].is_Symbol or j[0] in i[1].free_symbols)]
    return topological_sort([replace, edges], default_sort_key)

def resolve(sol):
    if not sol:
        return {}
    elif isinstance(sol, (Dict,dict)):
        sol = list(sol.items())
    else:
        assert all(isinstance(i, tuple) for i in sol)
    try:
        r = repsort(*sol)
    except ValueError:  # cycles
        return {}
    sol = {}
    for i in range(len(r)-1,-1,-1):
        k, v = r[i]
        sol[k] = v.xreplace(sol)
    return sol

def presolve(eqs, symset=None):
    Eqs, other = sift(eqs,
        lambda x: isinstance(x, Eq) and (
            not symset or x.lhs in symset), binary=True)
    # can we resolve the Eqs?
    sol = resolve([i.args for i in Eqs])
    if not sol:
        other.extend(i.rewrite(Add) for i in Eqs)
    if not other:
        return sol, other
    # solve for linear symbols in the others
    eqs = other
    while 1:
        # find free symbols of interest that appear linearly in each eq
        lfs = []
        for i in range(len(eqs)):
            f = set()
            free = eqs[i].free_symbols
            if symset:
                free = free & symset
            for j in free:
                if eqs[i].diff(j).is_number:
                    f.add(j)
            lfs.append(f)
        for i in range(len(lfs)-1, -1, -1):
            free = lfs[i]
            if len(free) == 1:
                # this is an easy solution
                x = free.pop()
                sol[x] = solve(eqs[i], x)[0]
                # remove from further consideration
                lfs.pop(i)
                eqs.pop(i)
                # update others with this information
                for i in range(i):
                    lfs[i] -= {x}
                    eqs[i] = eqs[i].subs(x, sol[x])
        # find the largest subsets of symbols that share symbols in common
        # and take the one that had most in common and appeared in the most
        # equations, e.g. x,y,z might appear in two equations or more -- use
        # as many as possible
        pmc = sorted(pairwise_most_common(lfs),
            key=lambda x: (len(i) for i in x))
        if not pmc:
            # we've removed the linear variables; return resolved
            # solution and the set of remaining equations to be solved
            return resolve(sol), set(eqs)
        s,l =  pmc.pop()
        l = set(flatten(l))
        # we are solving without passing s since we don't know whether
        # those symbols appear in a linearly independent way
        nsol = solve([eqs[i] for i in l])
        assert nsol and isinstance(nsol, dict)
        sol.update(nsol)
        # update remaining equations
        eqs = list(set([e.subs(sol) for i, e in enumerate(eqs) if i not in l]))

>>> var('a:z')
>>> eqs2 = [-a*k + 4*a/3 + b + 2*c/9 + 5*d/6 + 5*f/6, d/2 - f*k + 4*f/3, -d*k + d + f/6, 13*a/18 + 13*b/18 + 13*c/18, a + b/2 - c*k + 20*c/9, a/6 - b*k + b + c/18, a + 5*b/3 + c/3, 4*a/3 + 2*b/3 + 2*c, -g]
>>> presolve(eqs2)
({g: 0, d: 0, f: 0, b: c, a: -2*c}, {2*c*k - 13*c/9, -c*k + 13*c/18})
>>> eqs2[0] = eqs2[0]/-2
>>> presolve(eqs2)
({g: 0, d: 0, f: 0, b: c, a: -2*c}, {-c*k + 13*c/18})
>>> dQ4, dI4, I5, I2, I4, I6, Q4, dI1, Q2, dQ2, I1, I3 =var('dQ4, dI4, I5, I2, I4, I6, Q4, dI1, Q2, dQ2, I1, I3')
>>> eqs0 = (I1 - I2 - I3, I3 - I4 - I5, I4 + I5 - I6, -I1 + I2 + I6, -2*I1 - 2*I3 - 2*I5 - 3*I6 - dI1/2 + 12, -I4 + dQ4, -I2 + dQ2, 2*I3 + 2*I5 + 3*I6 - Q2, I4 - 2*I5 + 2*Q4 + dI4)
>>> presolve(eqs0)
({Q4: -dI1/18 - dI4/2 - 2*dQ2/9 - 23*dQ4/18 + 4/3, Q2: -7*dI1/18 - 14*dQ2/9 - 4*
dQ4/9 + 28/3, I6: -dI1/18 - 2*dQ2/9 + 2*dQ4/9 + 4/3, I5: -dI1/18 - 2*dQ2/9 - 7*d
Q4/9 + 4/3, I4: dQ4, I3: -dI1/18 - 2*dQ2/9 + 2*dQ4/9 + 4/3, I2: dQ2, I1: -dI1/18
 + 7*dQ2/9 + 2*dQ4/9 + 4/3}, set())
>>> d, kl_L2R2, dv_1, kl_R2, S1_0, kf_D1T2, ky_R2, kf_L2R2, ky_L2, kS_1, kS_2, kn, kl_R1, lf_1, kl_S1, rec_1, rh_1, R2_0, ky_R1, sh_2, a_1_0, kf_D1T1, R1_0, L2_S2_0, lrkd_1, lc_2, sh_1, rec_2, ky_L1, sc_2, a_2_0, eh_1, ldkd_2, ka, L2_R2_0, L1_S1_0, kn2_T2, kn2_T1, ah, kl_L2, t_S2, kl_D1, kf_L1R1, kl_L1R1, dv_2, lc_1, lrkd_2, sc_1, rh_2, S2_0, kl_L1, t_R1, L2_0, ldkd_1, t_S1, v, L1_0, kl_S2, lf_2, m, t_R2, L1_R1_0 = symbols('d, kl_L2R2, dv_1, kl_R2, S1_0, kf_D1T2, ky_R2, kf_L2R2, ky_L2, kS_1, kS_2, kn, kl_R1, lf_1, kl_S1, rec_1, rh_1, R2_0, ky_R1, sh_2, a_1_0, kf_D1T1, R1_0, L2_S2_0, lrkd_1, lc_2, sh_1, rec_2, ky_L1, sc_2, a_2_0, eh_1, ldkd_2, ka, L2_R2_0, L1_S1_0, kn2_T2, kn2_T1, ah, kl_L2, t_S2, kl_D1, kf_L1R1, kl_L1R1, dv_2, lc_1, lrkd_2, sc_1, rh_2, S2_0, kl_L1, t_R1, L2_0, ldkd_1, t_S1, v, L1_0, kl_S2, lf_2, m, t_R2, L1_R1_0')
>>> v = (ka, kl_D1, kn2_T1, kf_D1T1, kf_L1R1, kl_R1, kl_L1R1, kl_L1, kl_S1, t_R1, L1_0, t_S1, L1_R1_0, R1_0, L1_S1_0, S1_0, kS_1, ky_L1, ky_R1, a_1_0, kn2_T2, kf_D1T2, kf_L2R2, kl_R2, kl_L2R2, kl_L2, kl_S2, t_R2, L2_0, t_S2, L2_R2_0, R2_0, L2_S2_0, S2_0, kS_2, ky_L2, ky_R2, a_2_0)
>>> eqs = [Eq(ka, log(2)/(ah*d)), Eq(kl_D1, log(2)/(d*eh_1)), Eq(kn2_T1, kn*floor(dv_1/2)), Eq(kf_D1T1, kn*ldkd_1), Eq(kf_L1R1, kn*lrkd_1), Eq(kl_R1, log(2)/(m*rh_1)), Eq(kl_L1R1, kl_R1), Eq(kl_L1, log(2)/(lf_1*m)), Eq(kl_S1, log(2)/(m*sh_1)), Eq(t_R1, rec_1*v), Eq(L1_0, lc_1*v), Eq(t_S1, sc_1*v), Eq(L1_R1_0, L1_0*kn*t_R1/(v*(L1_0*kn/v + kf_L1R1 + kl_L1R1))), Eq(R1_0, -L1_R1_0 + t_R1), Eq(L1_S1_0, L1_0*kn*t_S1/(v*(L1_0*kn/v + kf_L1R1 + kl_L1))), Eq(S1_0, -L1_S1_0 + t_S1), Eq(kS_1, (L1_0*S1_0*kn/v - L1_S1_0*kf_L1R1 + S1_0*kl_S1)/R1_0), Eq(ky_L1, L1_0*R1_0*kn/v + L1_0*S1_0*kn/v + L1_0*kl_L1 - L1_R1_0*kf_L1R1 - L1_S1_0*kf_L1R1), Eq(ky_R1, L1_0*R1_0*kn/v - L1_R1_0*kf_L1R1 + R1_0*kS_1 + R1_0*kl_R1), Eq(a_1_0, L1_R1_0), Eq(kn2_T2, kn*floor(dv_2/2)), Eq(kf_D1T2, kn*ldkd_2), Eq(kf_L2R2, kn*lrkd_2), Eq(kl_R2, log(2)/(m*rh_2)), Eq(kl_L2R2, kl_R2), Eq(kl_L2, log(2)/(lf_2*m)), Eq(kl_S2, log(2)/(m*sh_2)), Eq(t_R2, rec_2*v), Eq(L2_0, lc_2*v), Eq(t_S2, sc_2*v), Eq(L2_R2_0, L2_0*kn*t_R2/(v*(L2_0*kn/v + kf_L2R2 + kl_L2R2))), Eq(R2_0, -L2_R2_0 + t_R2), Eq(L2_S2_0, L2_0*kn*t_S2/(v*(L2_0*kn/v + kf_L2R2 + kl_L2))), Eq(S2_0, -L2_S2_0 + t_S2), Eq(kS_2, (L2_0*S2_0*kn/v - L2_S2_0*kf_L2R2 + S2_0*kl_S2)/R2_0), Eq(ky_L2, L2_0*R2_0*kn/v + L2_0*S2_0*kn/v + L2_0*kl_L2 - L2_R2_0*kf_L2R2 - L2_S2_0*kf_L2R2), Eq(ky_R2, L2_0*R2_0*kn/v - L2_R2_0*kf_L2R2 + R2_0*kS_2 + R2_0*kl_R2), Eq(a_2_0, L2_R2_0)]
>>> presolve(eqs, set(v))
({t_S2: sc_2*v, t_S1: sc_1*v, t_R2: rec_2*v, t_R1: rec_1*v, kl_S2: log(2)/(m*sh_2), kl_R2: log(2)/(m*rh_2), kl_L2R2: log(2)/(m*rh_2), kl_L2: log(2)/(lf_2*m), kf_L2R2: kn*lrkd_2, L2_0: lc_2*v, L2_S2_0: kn*lc_2*sc_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(lf_2*m)), S2_0: -kn*lc_2*sc_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(lf_2*m)) + sc_2*v, L2_R2_0: kn*lc_2*rec_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(m*rh_2)), R2_0: -kn*lc_2*rec_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(m*rh_2)) + rec_2*v, kS_2: (-kn**2*lc_2*lrkd_2*sc_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(lf_2*m)) + kn*lc_2*(-kn*lc_2*sc_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(lf_2*m)) + sc_2*v) + (-kn*lc_2*sc_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(lf_2*m)) + sc_2*v)*log(2)/(m*sh_2))/(-kn*lc_2*rec_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(m*rh_2)) + rec_2*v), ky_R2: -kn**2*lc_2*lrkd_2*rec_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(m*rh_2)) - kn**2*lc_2*lrkd_2*sc_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(lf_2*m)) + kn*lc_2*(-kn*lc_2*rec_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(m*rh_2)) + rec_2*v) + kn*lc_2*(-kn*lc_2*sc_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(lf_2*m)) + sc_2*v) + (-kn*lc_2*sc_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(lf_2*m)) + sc_2*v)*log(2)/(m*sh_2) + (-kn*lc_2*rec_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(m*rh_2)) + rec_2*v)*log(2)/(m*rh_2), kl_S1: log(2)/(m*sh_1), kl_R1: log(2)/(m*rh_1), kl_L1R1: log(2)/(m*rh_1), kl_L1: log(2)/(lf_1*m), kf_L1R1: kn*lrkd_1, L1_0: lc_1*v, L1_S1_0: kn*lc_1*sc_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(lf_1*m)), S1_0: -kn*lc_1*sc_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(lf_1*m)) + sc_1*v, L1_R1_0: kn*lc_1*rec_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(m*rh_1)), R1_0: -kn*lc_1*rec_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(m*rh_1)) + rec_1*v, kS_1: (-kn**2*lc_1*lrkd_1*sc_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(lf_1*m)) + kn*lc_1*(-kn*lc_1*sc_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(lf_1*m)) + sc_1*v) + (-kn*lc_1*sc_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(lf_1*m)) + sc_1*v)*log(2)/(m*sh_1))/(-kn*lc_1*rec_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(m*rh_1)) + rec_1*v), ky_R1: -kn**2*lc_1*lrkd_1*rec_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(m*rh_1)) - kn**2*lc_1*lrkd_1*sc_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(lf_1*m)) + kn*lc_1*(-kn*lc_1*rec_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(m*rh_1)) + rec_1*v) + kn*lc_1*(-kn*lc_1*sc_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(lf_1*m)) + sc_1*v) + (-kn*lc_1*sc_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(lf_1*m)) + sc_1*v)*log(2)/(m*sh_1) + (-kn*lc_1*rec_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(m*rh_1)) + rec_1*v)*log(2)/(m*rh_1), ky_L2: -kn**2*lc_2*lrkd_2*rec_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(m*rh_2)) - kn**2*lc_2*lrkd_2*sc_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(lf_2*m)) + kn*lc_2*(-kn*lc_2*rec_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(m*rh_2)) + rec_2*v) + kn*lc_2*(-kn*lc_2*sc_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(lf_2*m)) + sc_2*v) + lc_2*v*log(2)/(lf_2*m), ky_L1: -kn**2*lc_1*lrkd_1*rec_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(m*rh_1)) - kn**2*lc_1*lrkd_1*sc_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(lf_1*m)) + kn*lc_1*(-kn*lc_1*rec_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(m*rh_1)) + rec_1*v) + kn*lc_1*(-kn*lc_1*sc_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(lf_1*m)) + sc_1*v) + lc_1*v*log(2)/(lf_1*m), kn2_T2: kn*floor(dv_2/2), kn2_T1: kn*floor(dv_1/2), kl_D1: log(2)/(d*eh_1), kf_D1T2: kn*ldkd_2, kf_D1T1: kn*ldkd_1, ka: log(2)/(ah*d), a_2_0: kn*lc_2*rec_2*v/(kn*lc_2 + kn*lrkd_2 + log(2)/(m*rh_2)), a_1_0: kn*lc_1*rec_1*v/(kn*lc_1 + kn*lrkd_1 + log(2)/(m*rh_1))}, [])
```
</p>
</details>

I envision `solve` having a `first` keyword so this would only be done on the first pass. Any equations in the set returned by `presolve` would then needed to be solved and a solution for each new solution of the remaining eq(s) along with the previouis linear solutions would be returned. Part of the speed of this comes from the fact that there is no need for simplification or checking of linear solutions when the coefficient is constant and not zero.  (But I modified this to consider as constant coefficients that were free of symbols of interest so maybe that is not totally true.)

oscarbenjamin:
> Here is as far as I've gotten on this. See what you think:

It looks good although I'm finding it hard to fully follow. I think it's similar to what I was thinking.

Why is it advantageous to find the most common pairwise subset rather than say a symbol that can be solved without reference to any other symbols?

smichr:
> Why is it advantageous to find the most common pairwise subset rather than say a symbol that can be solved without reference to any other symbols?

the latter is done by solving for the singleton sets before the pairwise most common step

I suspect the pairwise most common step might be important because it identifies a subset of equations that share the common condition on the symbols to be solved for. If one blindly just solves for a single symbol and substitutes it may create a dead-end. Not sure. But trying the greedy algorithm with the condition that you only solve for symbols with constant coefficients seems to work:

```python
def do(eqs):
 sol={}
 eqs2 = eqs[:]
 free = [i.free_symbols for i in eqs2]
 while 1:
     changed = False
     for f, e in zip(free, eqs2):
      for fi in f:
       m = e.diff(fi)
       if m and m.is_number:
        changed = True
        v = -e.subs(fi,0)/m
        for k in sol:
          sol[k] = sol[k].xreplace({fi:v})
        sol[fi] = v
        for i in range(len(eqs2)):
         if fi in free[i]:
           eqs2[i]=eqs2[i].xreplace({fi:v})
           free[i] = eqs2[i].free_symbols
        break
     if not changed:
         break
 if not any(eqs2):
   return [sol]
 rv = []
 for i in solve(eqs2):
   d = {k:v.xreplace(i) for k,v in sol.items()}
   d.update(i)
   rv.append(d)
 return rv
```
With this, 4 solutions for `eqs = [-a*k + 4*a/3 + b + 2*c/9 + 5*d/6 + 5*f/6, d/2 - f*k + 4*f/3, -d*k + d + f/6, 13*a/18 + 13*b/18 + 13*c/18, a + b/2 - c*k + 20*c/9, a/6 - b*k + b + c/18, a + 5*b/3 + c/3, 4*a/3 + 2*b/3 + 2*c, -g]` are identified:
```python
>>> do(eqs)
[
{d: 0, b: 0, c: 0, g: 0, f: 0, a: 0}, 
{d: -f, b: 0, c: 0, g: 0, k: 5/6, a: 0}, 
{d: 0, b: -a/2, c: -a/2, g: 0, f: 0, k: 13/18}, 
{d: 0, b: 0, c: 0, g: 0, k: 3/2, a: 0, f: 0}]
```

oscarbenjamin:
> With this, 4 solutions for

I wonder if any of these rearrangements can eliminate the k=3/2 case.

smichr:
> I wonder if any of these rearrangements can eliminate the k=3/2 case.

`k` is arbitrary if `{d: 0, b: 0, c: 0, g: 0, f: 0, a: 0}`; I haven't investigated how to eliminate specifying a value that is redundant. I did update the `do` code, however.

oscarbenjamin:
I would like to get some part of this in for the 1.9 release. What I have so far in the PR is a very minimal improvement. I don't want to make a major change right before the release so I just wonder if a simple improvement can be made (beyond what is here already). The main limitation is that the PR doesn't sort the equations after solving for some variables but it still seems to be sufficient for the examples in the motivating issue #21882.

@smichr what do you think of the changes in this PR as they stand? Is there a simple way to improve them?

(There are a couple of todos like adding tests still)

smichr:
I will take a closer look tomorrow.

smichr:
If you modify `_solve_system` down through the last line indicated it will find the 3 solutions to the motivating system of equations very quickly.

```python
def _solve_system(exprs, symbols, **flags):
    if not exprs:
        return []

    if flags.pop('_split_linear', True):
        flags['_split_linear'] = False
        symset = set(symbols)
        eqs = list(exprs)
        sol = {}
        while True:
            for i, e in enumerate(eqs):
                lsym = _syms_with_constant_coeff(e) & symset
                if not lsym:
                    continue
                x = next(ordered(lsym))
                y = -e.xreplace({x: 0})/e.coeff(x)
                s = {x: y}
                for k in sol:
                    sol[k] = sol[k].xreplace(s)
                sol.update(s)
                eqs.pop(i)
                for i in range(len(eqs)):
                    eqs[i] = eqs[i].xreplace(s)
                symset -= {x}
                break  # start again
            else:
                # no more linear symbols
                if not any(eqs):
                    # nothing left to do
                    return [sol]
                if not symset:
                    # there equations that were not 0
                    # but no symbols left to solve for
                    return []
                # get solutions for remaining equations
                rv = []
                nsol = _solve_system(eqs, symset, **flags)
                if isinstance(nsol, dict):  # if len(symset) == 1
                    nsol = [nsol]
                for d in nsol:
                    # update known solution values with d
                    r = {k: v.xreplace(d) for k, v in sol.items()}
                    # add d to the solution
                    r.update(d)
                    # this is one of the solutions to the system
                    rv.append(r)
                return rv

    if flags.pop('_split', True):
        flags['_split'] = False  # skip split step
        # Split the system into connected components
        V = exprs
        symset = set(symbols)
        exprsyms = {e: e.free_symbols & symset for e in exprs}
        E = []
        sym_indices = {sym: i for i, sym in enumerate(symbols)}
        for n, e1 in enumerate(exprs):
            for e2 in exprs[:n]:
                # Equations are connected if they share a symbol
                if exprsyms[e1] & exprsyms[e2]:
                    E.append((e1, e2))
        G = V, E
        subexprs = connected_components(G)
        if len(subexprs) > 1:
            subsols = []
            for subexpr in subexprs:
                subsyms = set()
                for e in subexpr:
                    subsyms |= exprsyms[e]
                subsyms = list(sorted(subsyms, key = lambda x: sym_indices[x]))
                subsol = _solve_system(subexpr, subsyms, **flags)
                if not isinstance(subsol, list):
                    subsol = [subsol]
                subsols.append(subsol)
            # Full solution is cartesion product of subsystems
            sols = []
            for soldicts in cartes(*subsols):
                sols.append(dict(item for sd in soldicts
                    for item in sd.items()))
            # Return a list with one dict as just the dict
            if len(sols) == 1:
                return sols[0]
            return sols

    polys = []
    dens = set()
    failed = []
    result = False
    linear = False
    manual = flags.get('manual', False)
    checkdens = check = flags.get('check', True)

    for j, g in enumerate(exprs):
        dens.update(_simple_dens(g, symbols))
        i, d = _invert(g, *symbols)
        g = d - i
        g = g.as_numer_denom()[0]
        if manual:
            failed.append(g)
            continue

        poly = g.as_poly(*symbols, extension=True)

        if poly is not None:
            polys.append(poly)
        else:
            failed.append(g)

    if not polys:
        solved_syms = []
    else:
        if all(p.is_linear for p in polys):
            n, m = len(polys), len(symbols)
            matrix = zeros(n, m + 1)

            for i, poly in enumerate(polys):
                for monom, coeff in poly.terms():
                    try:
                        j = monom.index(1)
                        matrix[i, j] = coeff
                    except ValueError:
                        matrix[i, m] = -coeff

            # returns a dictionary ({symbols: values}) or None
            if flags.pop('particular', False):
                result = minsolve_linear_system(matrix, *symbols, **flags)
            else:
                result = solve_linear_system(matrix, *symbols, **flags)
            if failed:
                if result:
                    solved_syms = list(result.keys())
                else:
                    solved_syms = []
            else:
                linear = True

        else:
            if len(symbols) > len(polys):
                from sympy.utilities.iterables import subsets

                free = set().union(*[p.free_symbols for p in polys])
                free = list(ordered(free.intersection(symbols)))
                got_s = set()
                result = []
                for syms in subsets(free, len(polys)):
                    try:
                        # returns [] or list of tuples of solutions for syms
                        res = solve_poly_system(polys, *syms)
                        if res:
                            for r in res:
                                skip = False
                                for r1 in r:
                                    if got_s and any([ss in r1.free_symbols
                                           for ss in got_s]):
                                        # sol depends on previously
                                        # solved symbols: discard it
                                        skip = True
                                if not skip:
                                    got_s.update(syms)
                                    result.extend([dict(list(zip(syms, r)))])
                    except NotImplementedError:
                        pass
                if got_s:
                    solved_syms = list(got_s)
                else:
                    raise NotImplementedError('no valid subset found')
            else:
                try:
                    result = solve_poly_system(polys, *symbols)
                    if result:
                        solved_syms = symbols
                        # we don't know here if the symbols provided
                        # were given or not, so let solve resolve that.
                        # A list of dictionaries is going to always be
                        # returned from here.
                        result = [dict(list(zip(solved_syms, r))) for r in result]
                except NotImplementedError:
                    failed.extend([g.as_expr() for g in polys])
                    solved_syms = []
                    result = None

    if result:
        if isinstance(result, dict):
            result = [result]
    else:
        result = [{}]

    if failed:
        # For each failed equation, see if we can solve for one of the
        # remaining symbols from that equation. If so, we update the
        # solution set and continue with the next failed equation,
        # repeating until we are done or we get an equation that can't
        # be solved.
        def _ok_syms(e, sort=False):
            rv = (e.free_symbols - solved_syms) & legal

            # Solve first for symbols that have lower degree in the equation.
            # Ideally we want to solve firstly for symbols that appear linearly
            # with rational coefficients e.g. if e = x*y + z then we should
            # solve for z first.
            def key(sym):
                ep = e.as_poly(sym)
                if ep is None:
                    complexity = (S.Infinity, S.Infinity, S.Infinity)
                else:
                    coeff_syms = ep.LC().free_symbols
                    complexity = (ep.degree(), len(coeff_syms & rv), len(coeff_syms))
                return complexity + (default_sort_key(sym),)

            if sort:
                rv = sorted(rv, key=key)
            return rv

        solved_syms = set(solved_syms)  # set of symbols we have solved for
        legal = set(symbols)  # what we are interested in
        # sort so equation with the fewest potential symbols is first
        u = Dummy()  # used in solution checking
        for eq in ordered(failed, lambda _: len(_ok_syms(_))):
            newresult = []
            bad_results = []
            got_s = set()
            hit = False
            for r in result:
                # update eq with everything that is known so far
                eq2 = eq.subs(r)
                # if check is True then we see if it satisfies this
                # equation, otherwise we just accept it
                if check and r:
                    b = checksol(u, u, eq2, minimal=True)
                    if b is not None:
                        # this solution is sufficient to know whether
                        # it is valid or not so we either accept or
                        # reject it, then continue
                        if b:
                            newresult.append(r)
                        else:
                            bad_results.append(r)
                        continue
                # search for a symbol amongst those available that
                # can be solved for
                ok_syms = _ok_syms(eq2, sort=True)
                if not ok_syms:
                    if r:
                        newresult.append(r)
                    break  # skip as it's independent of desired symbols
                for s in ok_syms:
                    try:
                        soln = _solve(eq2, s, **flags)
                    except NotImplementedError:
                        continue
                    # put each solution in r and append the now-expanded
                    # result in the new result list; use copy since the
                    # solution for s in being added in-place
                    for sol in soln:
                        if got_s and any([ss in sol.free_symbols for ss in got_s]):
                            # sol depends on previously solved symbols: discard it
                            continue
                        rnew = r.copy()
                        for k, v in r.items():
                            rnew[k] = v.subs(s, sol)
                        # and add this new solution
                        rnew[s] = sol
                        # check that it is independent of previous solutions
                        iset = set(rnew.items())
                        for i in newresult:
                            if len(i) < len(iset) and not set(i.items()) - iset:
                                # this is a superset of a known solution that
                                # is smaller
                                break
                        else:
                            # keep it
                            newresult.append(rnew)
                    hit = True
                    got_s.add(s)
                if not hit:
                    raise NotImplementedError('could not solve %s' % eq2)
            else:
                result = newresult
                for b in bad_results:
                    if b in result:
                        result.remove(b)

    default_simplify = bool(failed)  # rely on system-solvers to simplify
```

sympy/solvers/tests/test_solvers.py
```diff
@@ -1373,14 +1379,33 @@ def test_issue_5849_matrix():
...
+        5*b/3 + c/3 + a,
+        2*b/3 + 2*c + 4*a/3,
+        -g,
+    ]
+
+    answer = [
+        {a: 0, f: 0, b: 0, d: 0, c: 0, g: 0},
+        {a: 0, f: -d, b: 0, k: S(5)/6, c: 0, g: 0},
+        {a: -2*c, f: 0, b: c, d: 0, k: S(13)/18, g: 0},
+        {a: 0, f: 0, b: 0, k: S(3)/2, c: 0, d: 0, g: 0}
```
smichr, line 1405:
```suggestion
```

smichr:
If tests pass, I think this is a good stopping point for this PR.

oscarbenjamin:
Yes, I think it looks good. Thanks @smichr!

PR #21887, opened 2021-08-17
simplify - fixes issue #21355 and now some equations give correct result

### LINKED ISSUE #21355: `radsimp` gives wrong result for `1 / (x + sqrt(x**2))`
I got into an accidental discovery of this issue while developing #21343
Generally, even if `1 / (x + sqrt(x**2))` can have a nonzero denominator for `x > 0`, the expression is not rationalizable because multiplying `x - sqrt(x**2)` makes the denominator zero. (`x - sqrt(x**2)` should also not be rationalizable.)

I saw that `radsimp(1 / (x + sqrt(x**2)))` gives `zoo*(x + sqrt(x**2))` which should be a wrong result.
I think that it needs a test for intermediate zero or `NaN, zoo`.

praneethratna:
Fixes #21355 and now `radsimp(1/(x + sqrt(x**2)))` returns `1/(x + sqrt(x**2))` instead of `zoo*(x + sqrt(x**2))` and also added regression test in test_radsimp.py

#### Release Notes

* simplify
   * fixed bug in radsimp.

### DIFF
--- a/sympy/simplify/radsimp.py
+++ b/sympy/simplify/radsimp.py
@@ -967,8 +967,7 @@ def radsimp(expr, symbolic=True, max_terms=4):
             d *= num
             d = powdenest(_mexpand(d), force=symbolic)
             if d.has(S.Zero, nan, zoo):
-                n, d = fraction(expr)
-                return _unevaluated_Mul(n, 1/d)
+                return expr
             if d.is_Atom:
                 break
 
--- a/sympy/simplify/radsimp.py
+++ b/sympy/simplify/radsimp.py
@@ -961,6 +961,7 @@ def radsimp(expr, symbolic=True, max_terms=4):
                     keep = False
                 break
             from sympy.simplify.powsimp import powsimp, powdenest
+
             num = powsimp(_num(rterms))
             n *= num
             d *= num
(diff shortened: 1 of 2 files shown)


praneethratna:
@sylee957 please review

sympy/simplify/radsimp.py
```diff
@@ -966,6 +966,9 @@ def handle(expr):
             n *= num
             d *= num
             d = powdenest(_mexpand(d), force=symbolic)
+            if d.has(S.Zero, nan, zoo):
+                n, d = fraction(expr)
+                return _unevaluated_Mul(n, 1/d)
```
sylee957, line 971: what about `return expr`

praneethratna: @sylee957 I've made the changes , any other changes to make?

sympy/simplify/radsimp.py
```diff
@@ -966,6 +966,8 @@ def handle(expr):
             n *= num
             d *= num
             d = powdenest(_mexpand(d), force=symbolic)
+            if d.has(S.Zero, nan, zoo):
```
smichr, line 969: Does the problem already exist with `num` a few lines earlier?

praneethratna: @smichr No , the problem exists only after line 968.

PR #21894, opened 2021-08-18
Docs: Add Contributor Guide

nanjekyejoannah:

This PR adds docs for the ``contributing`` category.

NO ENTRY


doc/src/conf.py
```diff
@@ -32,6 +32,7 @@
 
 redirects = {
     "install.rst": "getting_started/install.rst",
+    "documentation-style-guide.rst": "contributing/documentation-style-guide.rst",
```
asmeurer, line 35: These should be redirecting to `html`, not `rst`.

doc/src/contributing/dev-setup.rst
```diff
@@ -0,0 +1,246 @@
...
+================================
+Development Environment Setup
+================================
+
+The first step to contributing to the code base is creating your development environment.
+
+Git Setup
+-----------
+
+SymPy is available on `GitHub <https://github.com/micropython/micropython>`_ and uses
```
asmeurer, line 12: This is the wrong URL.

doc/src/contributing/dev-setup.rst
```diff
@@ -0,0 +1,246 @@
...
+To start sympy in debug mode set the SYMPY_DEBUG variable. For instance in a unix-like system you would do
+
+    $ SYMPY_DEBUG=True bin/isympy
+
+or in Windows
+
+    > set SYMPY_DEBUG=True
+    > python bin/isympy
+
+Now just use for example the `limit()` function. You will get a nice printed tree, which is very useful for debugging.
```
asmeurer, line 246: `limit()` needs two backticks

doc/src/contributing/dev-setup.rst
```diff
@@ -0,0 +1,246 @@
...
+This will run tests for the ``core`` and ``utilities`` modules.
+
+Similary, run quality tests with:
+
+.. code-block:: bash
+
+    $ bin/test code_quality
+
+Build the Documentation
+------------------------
```
oscarbenjamin, line 127: This section (from here down to debugging) would be better on a separate page or perhaps as part of one of the pages about writing the docs.

There's a lot of instructions here about installing dependencies that are needed to build the docs but are not otherwise needed for sympy development.

doc/src/contributing/documentation-style-guide.rst
```diff
@@ -0,0 +1,409 @@
+========================
+Writing Documentation
```
asmeurer, line 2: We should keep "style guide" in the title of this page.

doc/src/contributing/cross-referencing.rst
```diff
@@ -0,0 +1,110 @@
+.. _style_guide_cross-referencing:
+
+Cross-Referencing
```
asmeurer, line 3: This doesn't really make sense as a top-level page. It makes more sense to put it as a section in the docstrings page.

oscarbenjamin:
It would be better if a PR like this had a reference to some other issue/PR in the OP. It looks like you deleted the PR template but it normally has a section for "references to other issues/PRs". For example if this PR is built on top of another you could reference it there. Ideally all of these PRs would link back to an issue where there would be some explanation or discussion about the purpose. I think anyone looking at this PR would struggle to understand why any changes are being made.

nanjekyejoannah:
This PR is now building on master, hence no need for reference but will ensure for the other PRs.

moorepants:
+1 from moving the contributor guide from the wiki to the docs.

Two general comments:

1. I think this page should be the starting point for new contributors: https://github.com/sympy/sympy/wiki/introduction-to-contributing, as it doesn't' start with dense technical git/github kung-fu

2. The contributor guide is super thorough, which is good in one sense, but also bad in another because it is so dense and long. Most other projects don't have such an overwhelming guide for newcomers. If it can be toned and reshaped some to give the essentials, I think it would be better for newcomers.

oscarbenjamin:
This is good to merge but let's wait until after I've created the 1.8 release branch

asmeurer:
We should move the wiki stuff here, but let's do it in a separate pull request.

PR #21896, opened 2021-08-19
implemented domain matrix transpose function

shivank070701:

a very simple fix would be simply pass the shape as argument to ddm transpose matrix function


oscarbenjamin:
What's the purpose of this?

There already is a DomainMAtrix transpose method:
```python
In [8]: from sympy.polys.matrices import DomainMatrix

In [9]: dM = DomainMatrix([[1, 2], [3, 4]], (2, 2), ZZ)

In [10]: dM
Out[10]: DomainMatrix([[1, 2], [3, 4]], (2, 2), ZZ)

In [11]: dM.transpose()
Out[11]: DomainMatrix([[1, 3], [2, 4]], (2, 2), ZZ)
```

Is there an open issue about adding this?

I'm going to close this PR because it does not seem to be needed.

Issue #21860, opened 2021-08-12
xreplace hangs


proy87:
`(-3*t + y**3 + sin(2*t + y)).subs({t: Rational(7,125), y: -1.23516983720720999822e+472802}, simultaneous=True)` hangs or takes too long.

oscargus:
The problem comes from Python trying to evaluate `2**(200000000000-1)` which it does in an exact way.

An observation:
https://github.com/sympy/sympy/blob/d7d65f6c2baaddde2c659bace54fb13de8e6eac3/sympy/core/numbers.py#L1821-L1825

Line 1821 results in calling 
https://github.com/sympy/sympy/blob/d7d65f6c2baaddde2c659bace54fb13de8e6eac3/sympy/core/numbers.py#L2339
which calls lines 1823-1825, at least in this case. It seems a bit inefficient to create an Integer objects to again, create a Rational object. It may be more efficient to do the complete rewrite on line 1821. 

Not that it solves this issue...

For the current issue, the lines 
https://github.com/sympy/sympy/blob/d7d65f6c2baaddde2c659bace54fb13de8e6eac3/sympy/core/numbers.py#L2368-L2391
tries to do some clever stuff. However, on lines 2373 and 2391, numerical powers are computed. Here is where the current code "hangs" (I think 2391, was a few days ago). Probably one should add some upper bound on the expected size of the result (`exp*log2(base)` should give an estimate of the number of bits required to represent the result), and not perform these parts if it is a to large value.

oscarbenjamin:
There are a number of open issues about this and related problems. See #6835 and other issues referenced from there.

oscargus:
I was wrong above, it was the line 
https://github.com/sympy/sympy/blob/d7d65f6c2baaddde2c659bace54fb13de8e6eac3/sympy/core/numbers.py#L2341
that is called in the second step (as the base is positive).

However, evaluating `Rational(a, b)**Rational(c, d)` for positive `Rational(a, b)` does the following

```
Rational(a, b)**Rational(c, d) =            # 1821
a**Rational(c, d)*b**(-Rational(c, d))   # 2341
a**Rational(c, d)*b**Rational(c*(d-1), d)/(b**c) # 1823-1825
```
It seems like writing this directly would be faster if nothing else, even though that requires a bit of additional code to check for the sign of `self` and use a slightly different expression otherwise. Also, I'm wondering if not just
`a**Rational(c, d)/b**Rational(c, d)`
[...]

Issue #21762, opened 2021-07-19
xreplace hangs


proy87:
`((x**2 + 6)**(Integer(33333333333333333)/50000000000000000)).xreplace({x: S.Half})`

oscarbenjamin:
A more direct way to reproduce this is just with something like:
```python
4 ** Rational(111111111111111111111111111111111, 2222222222222)
```
Basically sympy should not attempt to evaluate such large integer powers as this.

smichr:
duplicates #6835

PR #21900, opened 2021-08-19
Faster Rational power of Rational

### LINKED ISSUE #21647: xreplace hangs
`log((Integer(567)/500)**(811*(Integer(567)/500)**x/100)).xreplace({x: 6})`

oscargus:

Fixes #21647
Part of #21762 (original issue not example in https://github.com/sympy/sympy/issues/21762#issuecomment-882655739)
Part of #21860

Rewrote parts of `Rational**Rational` code to direct it to a wanted form. 

#### Other comments
Maybe one would like to do something similar for `Integer**Rational`, but I have not figured out the logic there yet, basically, factor out the integer multiple of the rational number (it might already be done though, not clear).

Not really sure that this is the "minimal" solution, i.e., that all the branches are needed etc, but it seems to work...

#### Release Notes

* core
   * `Rational**Rational` works better for large numbers.

### DIFF
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -1799,9 +1799,9 @@ class Rational(Number):
                 # (3/4)**-2 -> (4/3)**2
                 ne = -expt
                 if (ne is S.One):
-                    return Rational(self.q, self.p)
+                    return Rational(self.q, self.p, 1)
                 if self.is_negative:
-                    return S.NegativeOne**expt*Rational(self.q, -self.p)**ne
+                    return S.NegativeOne**expt*Rational(self.q, -self.p, 1)**ne
                 else:
                     return Rational(self.q, self.p)**ne
             if expt is S.Infinity:  # -oo already caught by test for negative
@@ -1822,14 +1822,14 @@ class Rational(Number):
                     remfracpart = intpart*expt.q - expt.p
                     ratfracpart = Rational(remfracpart, expt.q)
                     if self.p != 1:
-                        return Integer(self.p)**expt*Integer(self.q)**ratfracpart*Rational(1, self.q**intpart)
-                    return Integer(self.q)**ratfracpart*Rational(1, self.q**intpart)
+                        return Integer(self.p)**expt*Integer(self.q)**ratfracpart*Rational(1, self.q**intpart, 1)
+                    return Integer(self.q)**ratfracpart*Rational(1, self.q**intpart, 1)
                 else:
                     remfracpart = expt.q - expt.p
                     ratfracpart = Rational(remfracpart, expt.q)
                     if self.p != 1:
-                        return Integer(self.p)**expt*Integer(self.q)**ratfracpart*Rational(1, self.q)
-                    return Integer(self.q)**ratfracpart*Rational(1, self.q)
+                        return Integer(self.p)**expt*Integer(self.q)**ratfracpart*Rational(1, self.q, 1)
+                    return Integer(self.q)**ratfracpart*Rational(1, self.q, 1)
 
         if self.is_extended_negative and expt.is_even:
             return (-self)**expt
@@ -1843,7 +1843,7 @@ class Rational(Number):
         return mpmath.make_mpf(mlib.from_rational(self.p, self.q, prec, rnd))
 
     def __abs__(self):
-        return Rational(abs(self.p), self.q)
+        return Rational(abs(self.p), self.q, 1)
 
     def __int__(self):
         p, q = self.p, self.q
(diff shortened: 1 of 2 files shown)


oscargus:
Guess I will have to install 3.6... Or is support for that soon to be dropped? 🙏🏻

sympy/core/numbers.py
```diff
@@ -1816,13 +1816,18 @@ def _eval_power(self, expt):
...
-                    return Integer(self.p)**expt*Integer(self.q)**(-expt)
-                # as the above caught negative self.p, now self is positive
-                return Integer(self.q)**Rational(
-                expt.p*(expt.q - 1), expt.q) / \
-                    Integer(self.q)**Integer(expt.p)
+                intpart = expt.p // expt.q
+                if intpart:
+                    remfracpart = expt.q - expt.p
+                    if self.p != 1:
+                        return Integer(self.p)**expt*Integer(self.q)**Rational(remfracpart, expt.q)*Rational(1, self.q)
```
smichr, line 1823: Isn't this just `return Integer(self.p)**expt*Integer(self.q)**(-expt)`? or `Integer(self.p)**expt*Integer(self.q)**Rational(remfracpart - 1, expt.q)`?

oscargus: Note that the logic of `if intpart` was wrong in the code you commented on, so there is clearly a limited amount of test with a rational exponent > 1... (only the gamma-tests it seem).

smichr:
Note: one thing that can speed things up, if the `Rational` you are producing has no gcd for numerator and denominator, is to pass `gcd=1` to `Rational` to avoid recomputing this.

oscargus:
I though about that, but I am not sure that is the case somewhere? Unless it will be beneficial even for Rational(1, XX)? In that case there are several places (30-40 in the whole code base after a quick search) where it can be added.

oscargus:
I can say that this is a "two-stage fix". First, I figured out how to avoid the very long run time. Second, I realized that the end result typically ends up to be in the stated form, so that was a good guess to avoid recursing a bit. Amended the initial commit though, so not really visible here.

oscargus:
This is a bit surprising:
```
%timeit Rational(1, 3)
571 ns ± 44.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit Rational(1, 3, 1)
602 ns ± 34.1 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Despite the first statement creating another Rational object and calls the gcd function. Maybe related to optional keyword arguments?

So the last commit should maybe be changed?

smichr:
> So the last commit should maybe be changed?

Try use two non-trivial large primes as the numerator and denominator.

sympy/core/numbers.py
```diff
@@ -1799,9 +1799,9 @@ def _eval_power(self, expt):
                 # (3/4)**-2 -> (4/3)**2
                 ne = -expt
                 if (ne is S.One):
-                    return Rational(self.q, self.p)
+                    return Rational(self.q, self.p, 1)
                 if self.is_negative:
-                    return S.NegativeOne**expt*Rational(self.q, -self.p)**ne
+                    return S.NegativeOne**expt*Rational(self.q, -self.p, 1)**ne
```
smichr, line 1804: But here, you can't be sure that `self` was computed with `gcd=1` or not and since an evaluation is now happening it should be reduced. And is there any reason we are doing that instead of `Rational(self.q**ne, -self.p**ne, 1)`?

oscargus: Not sure about it, but my idea was that self was already reduced (since it is a `Rational`), but not fully sure.

The second rewrite is only possible if ne is integer (and some other rare cases) as the arguments to `Rational` must be integers (or `Rational`). An option is of course that `Rational(1, sqrt(2))` returns `1/sqrt(2)`, but currently this is not how it works.

oscargus: Still,  there seems to be limited benefits of adding the gcd-1 based on the timings. Although that is quite surprising considering the execution pattern of the code...

smichr: > my idea was that self was already reduced (since it is a `Rational`),

There can be unevaluated Rationals:
```python
>>> a,b=Rational(2,4,1),Rational(3,6,1);print(a,b)
(2/4, 3/6)
>>> a**b
sqrt(2)/2
```

smichr: > limited benefits of adding the gcd-1 based on the timings

The test may not be helpful in showing the difference, perhaps because of caching. But 
```python
got=[]
for gcd in range(2):
  for do in range(5):
    t=time.time();ok=[Rational(1,i,1) for i in range(10000)];got.append(time.time()-t)
print([sum(i) for i in got)
```
shows that there is about a 25% savings when using `gcd=1`, regardless of whether the out loop runs over `range(2)` or `reversed(range(2))`.

oscargus: Ahh, clever. Then we cannot know for sure. I'll edit.

Good that there is a gain. I was really confused as not providing the gcd seemed to execute more code, but still no gain was visible.

PR #21901, opened 2021-08-19
matrices - is_positive_semidefinite() now gives correct result

### LINKED ISSUE #20752: is_positive_semidefinite() gives wrong result on sympy==1.7.1
The code
```python
from sympy import Matrix, Symbol
b = Symbol('b', nonzero=True)
m = Matrix([[0, 0, 0], [0, b, 0], [0, 0, b]])
m.is_positive_semidefinite
```
gives wrong result `True`. Since `b` could be either positive or negative, the positive semidefinite-ness is inconclusive and should return `None`. The reason comes from the following line.
https://github.com/sympy/sympy/blob/347542065a39bef13d6fb9500dce631f2cf72f55/sympy/matrices/eigen.py#L825-L826
The test rejects only provably negative pivots and fails to consider _potentially_ negative pivots. It seems that adding strict nonnegativity tests like
```python
elif not (M[k, k].is_nonnegative and pivot_val.is_nonnegative):
    return None
```
will resolve this issue, but as I'm not 100% familiar with the implementation the details could be wrong.

praneethratna:
Fixes #20752 and now is_positive_semidefinite returns `None` for `Matrix([[0, 0, 0], [0, b, 0], [0, 0, b]])` where b is nonzero instead of `True`.

#### Release Notes

* matrices
   * fixed bug in is_positive_semidefinite() in eigen.py.

### DIFF
--- a/sympy/matrices/tests/test_eigen.py
+++ b/sympy/matrices/tests/test_eigen.py
@@ -681,3 +681,9 @@ def test_issue_20275():
         [Matrix([[S(1)/2 + sqrt(5)/2], [0], [1], [1], [0]]),
          Matrix([[S(1)/2 + sqrt(5)/2], [1], [0], [0], [1]])]
     )
+
+
+def test_issue_20752():
+    b = symbols('b',nonzero=True)
+    m = Matrix([[0, 0, 0], [0, b, 0], [0, 0, b]])
+    assert m.is_positive_semidefinite == None
--- a/sympy/matrices/tests/test_eigen.py
+++ b/sympy/matrices/tests/test_eigen.py
@@ -684,6 +684,6 @@ def test_issue_20275():
 
 
 def test_issue_20752():
-    b = symbols('b',nonzero=True)
+    b = symbols('b', nonzero=True)
     m = Matrix([[0, 0, 0], [0, b, 0], [0, 0, b]])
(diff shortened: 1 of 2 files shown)


sympy/matrices/tests/test_eigen.py
```diff
@@ -681,3 +681,9 @@ def test_issue_20275():
         [Matrix([[S(1)/2 + sqrt(5)/2], [0], [1], [1], [0]]),
          Matrix([[S(1)/2 + sqrt(5)/2], [1], [0], [0], [1]])]
     )
+
+
+def test_issue_20752():
+    b = symbols('b',nonzero=True)
+    m = Matrix([[0, 0, 0], [0, b, 0], [0, 0, b]])
+    assert m.is_positive_semidefinite == None
```
oscarbenjamin, line 689: There should be a space after a comma and also `== None` is better as `is None`.

praneethratna: I've made changes but can this test go with similar others : ?
https://github.com/sympy/sympy/blob/cf846222b245629f331c9ad38b80843f90cc0d56/sympy/matrices/tests/test_eigen.py#L505-L599

oscarbenjamin:
Looks good, thanks!

PR #18776, opened 2020-03-04
WIP: Add UnitQuaternion class

mgeier:
This is a slight variation of #17694.

This never tries to actually calculate the norm of quaternions. The unit-ness of `UnitQuaternion` is just assumed. This way, the components can be symbolic expressions.
If this assumption is violated, the results will be nonsensical (as expected?):

```pycon
>>> from sympy.algebras import UnitQuaternion
>>> q = UnitQuaternion(6, 7, 8, 9)
>>> q
6 + 7*i + 8*j + 9*k
>>> q.norm()
1
```

#### References to other Issues or PRs

This is (hopefully) an improvement over #17694, therefore it is also an alternative to  #18065 and #18086.

#### Brief description of what is fixed or changed

Same motivation as #17694.

#### Other comments

I've tried to make the result type a `UnitQuaternion` wherever this is guaranteed.
I might have missed some places, though.

Currently, a `UnitQuaternion` is never equal to a `Quaternion`, even if they have the same components:

```pycon
>>> from sympy.algebras import Quaternion, UnitQuaternion
>>> Quaternion(1, 0, 0, 0) == UnitQuaternion(1, 0, 0, 0)
False
```

I'm not sure whether this makes sense or not.

This is a "draft" PR because some docs and deprecation warnings are missing for now.
Probably some methods are missing as well?

#### Release Notes

* algebras
  * Added a UnitQuaternion class

PR #18086, opened 2019-12-20
Quaternion: Add normalize flag

mgeier:
... as suggested by @oscarbenjamin in https://github.com/sympy/sympy/pull/18065#issuecomment-566808193.

#### References to other Issues or PRs

This is an alternative to  #17694 and #18065. Probably the third time it's the charm?

#### Brief description of what is fixed or changed

I added a `normalize` flag to some of the `Quaternion` methods.

The default behavior is unchanged, but if `normalize=False` is given, the quaternion is assumed to already have unit length and no normalization will happen.

It works like this:

```pycon
>>> import sympy as sp
>>> sp.init_printing()
>>> a, b, c, d = sp.symbols('a:d')
>>> q = sp.Quaternion(a, b, c, d)
>>> q.to_rotation_matrix(normalize=False)
⎡     2      2                                          ⎤
⎢- 2⋅c  - 2⋅d  + 1   -2⋅a⋅d + 2⋅b⋅c      2⋅a⋅c + 2⋅b⋅d  ⎥
⎢                                                       ⎥
⎢                        2      2                       ⎥
⎢  2⋅a⋅d + 2⋅b⋅c    - 2⋅b  - 2⋅d  + 1   -2⋅a⋅b + 2⋅c⋅d  ⎥
⎢                                                       ⎥
⎢                                           2      2    ⎥
⎣ -2⋅a⋅c + 2⋅b⋅d      2⋅a⋅b + 2⋅c⋅d    - 2⋅b  - 2⋅c  + 1⎦
```

#### Other comments

This is a smaller and less intrusive change than my previous attempts (and I hope it has less problems, too), but it is also less powerful: If `inverse()` is called (either explicitly or implicitly by some other operation), a division by the squared norm still happens (which could be avoided in my previous attempts).

#### Release Notes

* algebras
  * Added `normalize` flag to `Quaternion`

PR #18065, opened 2019-12-17
Add assume_unit to Quaternion class

mgeier:
... and disable `rotate_point()`, `to_axis_angle()` and `to_rotation_matrix()`
for non-unit quaternions.

#### References to other Issues or PRs

This is an alternative to #17694.

#### Brief description of what is fixed or changed

I've added an `assume_unit` argument to the `Quaternion` constructor as well as an `.assume_unit()` method.
Probably one of them would suffice?

The `_assume_unit` attribute is propagated through mathematical operations, as far as I am aware that's possible. Most likely I missed something.

Here's an example that illustrates the change:

**Before this PR:**

```pycon
>>> import sympy as sp
>>> sp.init_printing()
>>> a, b, c, d = sp.symbols('a:d')
>>> q = sp.Quaternion(a, b, c, d)
>>> q.to_rotation_matrix()
⎡       ⎛ 2    2⎞                                                         ⎤
⎢     2⋅⎝c  + d ⎠              2⋅(-a⋅d + b⋅c)           2⋅(a⋅c + b⋅d)     ⎥
⎢- ───────────────── + 1     ─────────────────        ─────────────────   ⎥
⎢   2    2    2    2          2    2    2    2         2    2    2    2   ⎥
⎢  a  + b  + c  + d          a  + b  + c  + d         a  + b  + c  + d    ⎥
⎢                                                                         ⎥
⎢                                ⎛ 2    2⎞                                ⎥
⎢     2⋅(a⋅d + b⋅c)            2⋅⎝b  + d ⎠              2⋅(-a⋅b + c⋅d)    ⎥
⎢   ─────────────────     - ───────────────── + 1     ─────────────────   ⎥
⎢    2    2    2    2        2    2    2    2          2    2    2    2   ⎥
⎢   a  + b  + c  + d        a  + b  + c  + d          a  + b  + c  + d    ⎥
⎢                                                                         ⎥
⎢                                                         ⎛ 2    2⎞       ⎥
⎢     2⋅(-a⋅c + b⋅d)           2⋅(a⋅b + c⋅d)            2⋅⎝b  + c ⎠       ⎥
⎢   ─────────────────        ─────────────────     - ───────────────── + 1⎥
⎢    2    2    2    2         2    2    2    2        2    2    2    2    ⎥
⎣   a  + b  + c  + d         a  + b  + c  + d        a  + b  + c  + d     ⎦
```

This normalizes the quaternion without asking.

[...]

PR #17694, opened 2019-10-03
Proof of Concept: Add UnitQuaternion class

mgeier:
#### Brief description of what is fixed or changed

Previously, the `Quaternion` class had methods like `from_axis_angle()` and `from_rotation_matrix()`, which are supposed to create a unit quaternion.
That's OK, because a unit quaternion *is a* quaternion, but then there were also the methods `to_axis_angle()`, `to_rotation_matrix()` and `rotate_point()`, which really only make sense for unit quaternions.

In this PR I'm adding a separate `UnitQuaternion` class for those cases.

#### Other comments

I'm not sure if that makes sense from a mathematical standpoint, please tell me if it doesn't.
I've gotten the idea for a separate class from https://www.nalgebra.org/rustdoc/nalgebra/geometry/type.UnitQuaternion.html, but I'm not sure if it makes sense within SymPy.

For now, I've just moved some methods from `Quaternion` to `UnitQuaternion`, which is a breaking change.

I would like to get some feedback first, but if this turns out to be considered for merging, we'd have to add some compatibility stubs and probably some deprecation warnings.

#### Release Notes

* algebras
  * Added a UnitQuaternion class

PR #21902, opened 2021-08-19
Removed unnecessary normalization in quaternion.py

dhyams:
removed unnecessary normalization of the created quaternion in the from_axis_angle Quaternion method.  The quaternion is normalized by construction.

Also added an is_normalized flag to Quaternion.rotate_point, which allows the user to skip the forced normalization of the input quaternion, if the user knows that it is already normalized.

Forgoing these normalizations is sometimes a key operation to be able to successfully simplify complex algebraic output from these routines.

None

removed unnecessary normalization of the created quaternion in the from_axis_angle Quaternion method.  The quaternion is normalized by construction.

Also added an is_normalized flag to Quaternion.rotate_point, which allows the user to skip the forced normalization of the input quaternion, if the user knows that it is already normalized.

#### Other comments

Forgoing these normalizations is sometimes a key operation to be able to successfully simplify complex algebraic output from these routines.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/algebras/quaternion.py
+++ b/sympy/algebras/quaternion.py
@@ -116,7 +116,7 @@ class Quaternion(Expr):
         c = y * s
         d = z * s
 
-        return cls(a, b, c, d).normalize()
+        return cls(a, b, c, d)
 
     @classmethod
     def from_rotation_matrix(cls, M):
@@ -585,7 +585,7 @@ class Quaternion(Expr):
                           integrate(self.c, *args), integrate(self.d, *args))
 
     @staticmethod
-    def rotate_point(pin, r):
+    def rotate_point(pin, r, is_normalized=False):
         """Returns the coordinates of the point pin(a 3 tuple) after rotation.
 
         Parameters
@@ -599,6 +599,9 @@ class Quaternion(Expr):
 
             It's important to note that when r is a tuple, it must be of the form
             (axis, angle)
+
+        is_normalized: boolean
+            Whether or not the quaternion r is already normalized.
 
         Returns
         =======
@@ -625,7 +628,7 @@ class Quaternion(Expr):
             q = Quaternion.from_axis_angle(r[0], r[1])
         else:
             # if r is a quaternion
-            q = r.normalize()
+            q = r if is_normalized else r.normalize()
         pout = q * Quaternion(0, pin[0], pin[1], pin[2]) * conjugate(q)
         return (pout.b, pout.c, pout.d)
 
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
See also #18065 and #18086.

I'm not sure what this comment means:
```
removed unnecessary normalization of the created quaternion in the
from_axis_angle Quaternion method. The quaternion is normalized by construction.
```

This would need tests and any change of behaviour would need to be explained in a release note.

dhyams:
Hello Oscar; after seeing the other pull requests, my contribution should be resubmitted, as there are other people trying to address situations where people are working with unit quaternions (which is the most common use case).

**My one contribution that I think is important** is directly related to your question above, and there are already unit tests that cover it.

Specifically to answer your question, it can be proven that the quaternion that is being constructed is already normalized, and thus does not need to have normalize() called on it again.  Small proof, using the same names as inside of the from_axis_angle routine:

c^2 + (s*x)*2 + (s*y)^2 + (s*z)^2 = c^2 + s^2*(x^2 + y^2 + z^2) = c^2 + s^2 * 1 = c^2 + s^2 = 1

dhyams:
@oscarbenjamin, I am in the process of resubmitting the pull request.

oscarbenjamin:
> there are other people trying to address situations where people are working with unit quaternions (which is the most common use case).

I don't think that @mgeier is still working on that so it is still an open issue.

dhyams:
Oscar; all right.  If I have time I will address.  I do think it is a good idea to separate out the "by construction" pull request, because that is a sure thing, whereas any fixes that involve optional normalization will be subject to opinion.  I should have separated it out to begin with.

oscarbenjamin:
When you open a new PR please add a reference to the old one so that there are links in both directions for anyone reading this in future trying to understand what happened.

mgeier:
> See also #18065 and #18086.

Other failed (or at least unfinished) attempts are at #17694 and  #18776, with the latter probably being the most promising.

A large part of all those is to remove unnecessary normalization.

> I don't think that @mgeier is still working on that so it is still an open issue.

I indeed stopped working on this, but not due to lack of interest, but just because I don't know how to continue.

I've described the main problem in https://github.com/sympy/sympy/pull/18776#issuecomment-787466433, @oscarbenjamin has given a hint in https://github.com/sympy/sympy/pull/18776#issuecomment-787476443, but I don't know how to continue from there.

oscarbenjamin:
@mgeier it seems that all of the discussion around this has taken place across multiple different PRs. Is there an issue somewhere for this? Could you either add to that issue or open an issue?

It would be good to summarise the different attempts and what works or doesn't work. Most importantly I think it would be good to clearly explain what the actual problem is before attempting to make a fix for it. I have some ideas that might work but I need to be clear what the problem is first.

PR #21903, opened 2021-08-19
nD (counting) and multiset_derangements (generating)

smichr:

A function for counting derangements of a sequence and another for generating the derangement of elements of the sequence is added.

#### Other comments

There have been many comments about combinatorial functions in general and possibilities for enhancement of the generation of the function added here in #6662. This PR is intended as a place to now elicit comments for improvement from the wider community.

#### Release Notes

* functions
  * combinatorial numbers enumeration function for deranngements, nD, added that is similar to nP, nC and nT.
* utilities
  * iterables function for generating derangements of a sequence was added

### DIFF
--- a/sympy/functions/combinatorial/numbers.py
+++ b/sympy/functions/combinatorial/numbers.py
@@ -15,11 +15,12 @@ from sympy.core.evalf import pure_complex
 from sympy.core.expr import Expr
 from sympy.core.function import Function, expand_mul
 from sympy.core.logic import fuzzy_not
-from sympy.core.mul import prod
-from sympy.core.numbers import E, pi, Rational, Integer
-from sympy.core.relational import LessThan, StrictGreaterThan
+from sympy.core.mul import Mul, prod
+from sympy.core.numbers import E, pi, oo, Rational, Integer
+from sympy.core.relational import is_le, is_gt
 from sympy.external.gmpy import SYMPY_INTS
-from sympy.functions.combinatorial.factorials import binomial, factorial
+from sympy.functions.combinatorial.factorials import (binomial,
+    factorial, subfactorial)
 from sympy.functions.elementary.exponential import log
 from sympy.functions.elementary.integers import floor
 from sympy.functions.elementary.miscellaneous import sqrt, cbrt
@@ -830,16 +831,15 @@ class harmonic(Function):
         if m.is_zero:
             return n
 
-        if n is S.Infinity and m.is_Number:
-            # TODO: Fix for symbolic values of m
+        if n is S.Infinity:
             if m.is_negative:
                 return S.NaN
-            elif LessThan(m, S.One):
+            elif is_le(m, S.One):
                 return S.Infinity
-            elif StrictGreaterThan(m, S.One):
+            elif is_gt(m, S.One):
                 return zeta(m)
             else:
-                return cls
+                return
 
         if n == 0:
             return S.Zero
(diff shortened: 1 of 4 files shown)


oscargus:
LGTM!

smichr:
There is some ongoing discussion on the issue page about this, so I am changing this to "draft".

sympy/utilities/iterables.py
```diff
@@ -2285,14 +2285,244 @@ def generate_involutions(n):
...
+    rv = [None]*n
+    # special cases
+
+    # 1) singletons
+    if len(ms) == n:
+        for p in generate_derangements(s):
+            yield p
+        return
+
+    # 2) aaabbb-like
```
rathmann, line 2320: I don't think case #2 is needed any more -- the finish_derangements code handles these cases efficiently.
For example:
```
assert(['b', 'a'] * 25 ==
           list(list(d) for d in
                multiset_derangements_mod2(['a', 'b'] * 25)) [0]
```
returns instantly.

rathmann:
Is this PR at a state where we are mostly discussing the algorithm, or are we ready to consider the API issues as well?

If this were merged now, wouldn't we have two competing multiset derangements routines?   (the old `generate_derangements()` and the new `multiset_derangements()`)

Near as I can tell, the new code is a great improvement over the old, **except** that there are two changes in semantics:

1. The old code emits derangements relative to a cannonical order, the new code is relative to the argument order.
2. The old gave the derangements in lexicographical order, the new produces a predictable-but-idiosyncratic order.

I suspect that both of these are fine, but should be mentioned in the documentation and release notes.

smichr:
> I suspect that both of these are fine, but should be mentioned in the documentation and release notes.

They both work the same way and can be used interchangeably. Since `generate_derangements` was there first, `multiset_derangements` can be thought of as a helper for it.

smichr:
I wonder if `nD` should support a flag like `counts` or `multiplicity' which would allow `s` to be interpreted as multiplicities rather than elements so [1,1,1,2,2] or {1:3, 2:2} would correspond to 'abcddee'.

smichr:
I wonder if the special case of only 1 repeated element should be special cased (e.g. abcdefff).

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,3 +2177,124 @@ def eval(cls, n):
...
+    Since the number of derangments depends on the multiplicity of the
+    elements and not the elements themselves, it may be more convenient
+    to give a list or multiset of multiplicities using keyword ``m``:
+
+    >>> assert nD('abc') == nD(m=(1,1,1)) == nD(m={1:3}) == 2
+
+    """
+    from sympy.core.compatibility import SYMPY_INTS
+    from sympy.utilities.iterables import multiset, multiset_derangements, iterable
+    from sympy.integrals.integrals import integrate, Integral
```
smichr, line 2228:
```suggestion
    from sympy.integrals.integrals import integrate
```

sympy/functions/combinatorial/tests/test_comb_numbers.py
```diff
@@ -14,9 +14,7 @@
 from sympy.core.numbers import GoldenRatio, Integer
 
 from sympy.testing.pytest import XFAIL, raises, nocache_fail
-
-
-x = Symbol('x')
+from sympy.abc import x, y
```
smichr, line 17:
```suggestion
x = Symbol('x')
```

rathmann:
> I wonder if the special case of only 1 repeated element should be special cased (e.g. abcdefff).

Not sure how often this case arises, but if it is justified, I would think such a case might be wired in as 

```python
if mx ==1:   # Then ALL remaining elements to place are singletons
   # special case code - basically a set-based permutation algorithm
   . . .
```

That way any multiset which has a few singletons will get the special case treatment.  A generic permutation algorithm for the remaining elements, modified to respect the derangement constraints, should be reasonably easy to code in a way that is faster than having a stack of calls to subsets().

sympy/utilities/iterables.py
```diff
@@ -2285,13 +2285,238 @@ def generate_involutions(n):
...
+            rv[i] = None
+        for i in forced_b:
+            rv[i] = None
+
+    def iopen(v):
+        return [i for i in range(n) if rv[i] is None and s[i] != v]
+
+    def do(j):
+        if j == -1:
+            yield rv
```
smichr, line 2408:
```suggestion
        if j == 1:
            yield from finish_derangements()
```

sympy/utilities/iterables.py
```diff
@@ -2168,13 +2168,238 @@ def generate_involutions(n):
...
+                return rv
+    return rv
 
-    TODO: This will be rewritten to use the
-    ECO operator approach once the permutations
-    branch is in master.
+
+def generate_derangements(perm):
+    """
+    Routine to generate unique derangements or sets or multisets.
```
smichr, line 2402:
```suggestion
    Routine to generate unique derangements of sets or multisets.
```

sympy/utilities/iterables.py
```diff
@@ -2195,9 +2420,21 @@ def generate_derangements(perm):
...
-        if not any(i == j for i, j in zip(perm, p)):
-            yield p
+    if not has_dups(perm):
+        s = perm
+        if len(perm) == 2:
+            yield [s[1],s[0]]
+            return
+        if len(perm) == 3:
+            yield [s[1],s[2],s[0]]
+            yield [s[2],s[0],s[1]]
```
smichr, line 2430:
```suggestion
            yield [s[1], s[0]]
            return
        if len(perm) == 3:
            yield [s[1], s[2], s[0]]
            yield [s[2], s[0], s[1]]
```

smichr:
@rathman, I don't have additions to this. Let's commit this since it is a significant improvement over what we already have.

PR #21904, opened 2021-08-19
Quaternion - exp() and _ln() of quaternion with 0 imaginary part now returns 0 instead of NaN

### LINKED ISSUE #21576: Wrong exponent of (a, 0, 0, 0) quaternion
In case of a quaternion with zero imaginary part division by `vector_norm` causes NaNs instead of 0 in the folowing lines of code. 
https://github.com/sympy/sympy/blob/d17907e248abe9f1b3d8aa98701ba6d48bd948d9/sympy/algebras/quaternion.py#L459-L463
Example:
wrong_quaternion_exponent

praneethratna:
Fixes #21576 and  now `Quaternion(0, b, 0, 0).exp().subs(b, 0)` returns `1 + 0*i + 0*j + 0*k`  instead of  
`1 + NaN*i + NaN*j + NaN*k` and `Quaternion(w, x, y, z)._ln().subs(w**2 + x**2 + y**2 + z**2, 1).subs(x, 0).subs(y, 0).subs(z, 0)` returns `0 + 0*i + 0*j + 0*k` instead of `0 + 0*i + 0*j + NaN*k` and also added regression tests in test_quaternion.py

#### Release Notes

* algebras
   * fixed bug in Quaternion.exp and Quaternion._ln


praneethratna:
@oscarbenjamin please review

sympy/algebras/quaternion.py
```diff
@@ -459,9 +459,9 @@ def exp(self):
         q = self
         vector_norm = sqrt(q.b**2 + q.c**2 + q.d**2)
         a = exp(q.a) * cos(vector_norm)
-        b = exp(q.a) * sin(vector_norm) * q.b / vector_norm
-        c = exp(q.a) * sin(vector_norm) * q.c / vector_norm
-        d = exp(q.a) * sin(vector_norm) * q.d / vector_norm
+        b = exp(q.a) * sin(vector_norm) * q.b / (Piecewise((vector_norm, vector_norm > 0), (1, True)))
+        c = exp(q.a) * sin(vector_norm) * q.c / (Piecewise((vector_norm, vector_norm > 0), (1, True)))
+        d = exp(q.a) * sin(vector_norm) * q.d / (Piecewise((vector_norm, vector_norm > 0), (1, True)))
```
oscarbenjamin, line 464: There should be tests checking that these Piecewise are returned.

Rather than `vector_norm > 0` I think it would be better to use `Ne(vector_norm, 0)`.

The common factor could be precomputed like:
```python
scale = sin(vector_norm) / Piecewise((vector_norm, Ne(vector_norm, 0)), (1, True))
```
Actually this Piecewise is just `sinc` so I think it would be better to use `sinc` here.

praneethratna: I've made the changes , any more changes to make?

sympy/algebras/quaternion.py
```diff
@@ -459,9 +459,9 @@ def exp(self):
         q = self
         vector_norm = sqrt(q.b**2 + q.c**2 + q.d**2)
         a = exp(q.a) * cos(vector_norm)
-        b = exp(q.a) * sin(vector_norm) * q.b / vector_norm
-        c = exp(q.a) * sin(vector_norm) * q.c / vector_norm
-        d = exp(q.a) * sin(vector_norm) * q.d / vector_norm
+        b = exp(q.a) * sinc(vector_norm).rewrite(sin) * q.b
+        c = exp(q.a) * sinc(vector_norm).rewrite(sin) * q.c
+        d = exp(q.a) * sinc(vector_norm).rewrite(sin) * q.d
```
oscarbenjamin, line 464: I wonder if it would be better to keep this as `sinc` rather than turn it into a Piecewise...

praneethratna: Okay , Then there are more things to change as if we use sinc instead of sin
```
>>> Quaternion(1, 2, 3, 4).exp()
>>> E*cos(sqrt(29)) + 2*E*sinc(sqrt(29))*i + 3*E*sinc(sqrt(29))*j + 4*E*sinc(sqrt(29))*k
```
instead of `E*cos(sqrt(29)) + 2*sqrt(29)*E*sin(sqrt(29))/29*i + 3*sqrt(29)*E*sin(sqrt(29))/29*j + 4*sqrt(29)*E*sin(sqrt(29))/29*k`

oscarbenjamin: I think it would be best to discuss the different possibilities on the issue rather than here.

praneethratna: I feel we can go ahead with sinc and also change the Docstrings to match the output? @oscarbenjamin

Issue #20598, opened 2020-12-13
Use stdlib and gmpy implementations of rationals


oscarbenjamin:
There are two implementations of rational numbers in sympy: Rational (which is Basic) and PythonRational (QQ in the poly domains). When gmpy2 is installed the poly domains will use mpq instead which is significantly faster. 

The Python stdlib has the fractions module which provides the Fraction type. This is well maintained and optimised as a pure Python implementation. The gmpy implementation is in C and is much faster and should be used when possible. There is no need for sympy to reimplement rationals like this when they are both well implemented in pure Python in the stdlib and better implemented in C in an optional dependency.

We should have a compatibility module somewhere that checks for the existence of gmpy2 and uses either mpq or Fraction as an implementation of rationals. That should then be used both by the poly domains and also by Rational. Each Rational instance would internally hold a Fraction/mpz instance and use that for all calculations. The poly domains should just use mpz/Fraction directly as the instances for QQ.

Here are some timings for Fraction, PythonRational, gmpy2 and Rational:
```python
# timefrac.py

from fractions import Fraction
from sympy.polys.domains.pythonrational import PythonRational
from sympy import Rational
from gmpy2 import mpq
from random import randrange as R

nums = [(R(100), R(1, 10)) for _ in range(500)]
nums_frac = [Fraction(n, d) for n, d in nums]
nums_pr = [PythonRational(n, d) for n, d in nums]
nums_rational = [Rational(n, d) for n, d in nums]
nums_gmpy = [mpq(n, d) for n, d in nums]

def time_frac(n):
    return sum(nums_frac[:n])

def time_pr(n):
    return sum(nums_pr[:n])

def time_rational(n):
    return sum(nums_rational[:n])

def time_gmpy(n):
    return sum(nums_gmpy[:n])
```
Running those with timeit:
```console
$ python -m timeit -s 'from timefrac import time_frac' 'time_frac(500)'
200 loops, best of 5: 1.58 msec per loop
$ python -m timeit -s 'from timefrac import time_pr' 'time_pr(500)'
100 loops, best of 5: 2.19 msec per loop
$ python -m timeit -s 'from timefrac import time_rational' 'time_rational(500)'
50 loops, best of 5: 4.17 msec per loop
$ python -m timeit -s 'from timefrac import time_gmpy' 'time_gmpy(500)'
5000 loops, best of 5: 89.3 usec per loop
```
[...]

PR #21910, opened 2021-08-20
Added gcd to Rational

oscargus:
Based on discussions in #21900, it seems to be slightly faster to explicitly mention the gcd of rational numbers on creation.

#### Other comments
Moved some computation to speed things up slightly.

#### Release Notes

NO ENTRY


smichr:
LGTM

oscarbenjamin:
I think `Rational(1, 2, 1)` seems quite cryptic. Doesn't look like the gcd parameter is documented anywhere.

It would be better to have a different constructor for this I think.

If we really want to make Rational faster then we should have it use `MPF` internally (https://github.com/sympy/sympy/issues/20598)

There is a `MPQ` class now that can be used:
https://github.com/sympy/sympy/blob/81f1401300956b27e5dff870c37022421c7e37bb/sympy/external/gmpy.py#L86

oscarbenjamin:
Maybe there could be a function like `rat_no_gcd` or something that makes it clearer what this is doing. Perhaps some of these files files could do:
```
from sympy.core.numbers import Rational

R = Rational.no_gcd

half = R(1, 2)
```

robinechuca:
Maybe I'm wrong but it seems to me that for better performances, we shouldn't rebuild a rational each time from a class but just use an instance.
```
sqrt(4 - 2*sqrt(2)): S.Pi*Rational(3, 8, 1),
2/sqrt(2 + sqrt(2)): S.Pi*Rational(3, 8, 1),
```
Could be
```
pi_rat_3_8 =  S.Pi*Rational(3, 8, 1)
sqrt(4 - 2*sqrt(2)): pi_rat_3_8,
2/sqrt(2 + sqrt(2)): pi_rat_3_8,
```

oscargus:
I think that you are correct @robinechuca 

There are many instances in the code where the same thing is computed again.

Another, semi-related, question: Rational uses standard integers (non-sympified). Wouldn't these be sympified every time they are used?

smichr:
> Doesn't look like the gcd parameter is documented anywhere.

I add documentation in #21944. The additions here, at least, mark where future code should focus on using a better Rational construction. This, at least, is doing no harm and is easy to find and undo. Except for large primes, however, I doubt that any of this has an impact on the overall runtime of code in which it is used.

oscargus:
Not sure how to properly benchmark it. Just a simple `%timeit` seems to implicate that not using the last 1 is actually slightly faster, which seems unintuitive.

PR #14497, opened 2018-03-13
logarithm : handling of different bases symbolically

abaokar-07:
As @asmeurer suggested in issue #6888 that `log(x, b)` should return `log(x, b)` instead of `log(x)/log(b)`, this returns `log(x, b)`

PR #20819, opened 2021-01-18
New exponential function represented as power

Upabjojr:
Splitting the work in https://github.com/sympy/sympy/pull/13211 - that PR is huge.

This PR introduces a new E constant (`S.Exp1New` or `_E_new`) whose exponential evaluates to `Pow` (and not `exp`, unlike `S.Exp1`).

This is a first step into removing the class `exp` in order to represent the exponential as a `Pow` object.

#### Release Notes

* core
  * Introduce a second E constant whose exponent is evaluated into Pow.
* functions
  * Introduce an second hidden exponential functions evaluating to Pow.

PR #21911, opened 2021-08-20
[WIP] Introduced LogWithBase class

oscargus:

Fixes #6888
Closes #14497

Introduced the `LogWithBase` class.

#### Other comments
Apart from the failing test, one thing that is bothering me is
`log(x, y, evaluate=False)` which will not return a `LogWithBase`. Seems like one will have to create a `__new__` method for `log` that can return a `LogWithBase` (still unevaluated), but as I understand it, it is tricky to combine with keyword arguments.

#### Release Notes

* functions
     * Better handling of logs with different base than e.


oscargus:
@asmeurer was it something like this you were looking for regarding LogWithBase? If so, I'll try to figure out what happens with the last test.

Not sure about the current release plans (@oscarbenjamin  ?) but it would sort of feel safer to introduce such a change in master just after a release so that any complications are kept from the standard users. I guess that same holds for #20819 which is sort of the power counter part to this.

asmeurer:
Sorry, I think I missed this when it was originally opened. 

It looks like it would be better to have a common base class here to avoid writing `log` and `LogWithBase` in so many places (unfortunately, `LogBase` is a rather confusing name). Or perhaps `log` should itself be a subclass of `LogWithBase`, although I'm not sure how well that plays with the Liskov substitution principle, although it does seem like there is a lot of code that could be shared.

Also LogWithBase itself is not the greatest name so if you have a better suggestion we can use it instead. Maybe `log_b`?

Issue #15497, opened 2018-11-15
Parametrised tests like in pytest


oscarbenjamin:
It doesn't look as if sympy's test module has a way to do parametrised tests like pytest does. These could be very useful in e.g. the `test_ode.py` which has many repetitive but not always consistent tests.

With `pytest` it works like this:
```python
# test_solve.py

from sympy import solve, Symbol
from pytest import mark
parametrize = mark.parametrize

x = Symbol('x')

eqns = {
    'linear': (x-1, [1]),
    'quadratic': (x**2 - x, [0, 1]),
    'cubic': (x**3, [1])
}

@parametrize('eqname', eqns.keys())
def test_solve(eqname):
    eq, sol = eqns[eqname]
    assert solve(eq) == sol
```

Then each case becomes a separate test:
```console
$ pytest solve_test.py
========================================== test session starts ==========================================
platform linux -- Python 3.6.2, pytest-4.0.0, py-1.7.0, pluggy-0.8.0
rootdir: /space/enojb/current/sympy/sympy, inifile:
collected 3 items                                                                                       

solve_test.py ..F                                                                                 [100%]

=============================================== FAILURES ================================================
___________________________________________ test_solve[cubic] ___________________________________________

eqname = 'cubic'

    @parametrize('eqname', eqns.keys())
    def test_solve(eqname):
        eq, sol = eqns[eqname]
>       assert solve(eq) == sol
E       assert [0] == [1]
E         At index 0 diff: 0 != 1
E         Use -v to get the full diff

solve_test.py:16: AssertionError
================================== 1 failed, 2 passed in 0.48 seconds ===================================
```

You can the easily re-run only the failed test with
```console
$ pytest solve_test.py -k 'solve[cubic]'
========================================== test session starts ==========================================
platform linux -- Python 3.6.2, pytest-4.0.0, py-1.7.0, pluggy-0.8.0
rootdir: /space/enojb/current/sympy/sympy, inifile:
collected 3 items / 2 deselected                                                                        

solve_test.py F                                                                                   [100%]

=============================================== FAILURES ================================================
___________________________________________ test_solve[cubic] ___________________________________________

eqname = 'cubic'

    @parametrize('eqname', eqns.keys())
    def test_solve(eqname):
        eq, sol = eqns[eqname]
>       assert solve(eq) == sol
E       assert [0] == [1]
E         At index 0 diff: 0 != 1
E         Use -v to get the full diff

solve_test.py:16: AssertionError
[...]

PR #21916, opened 2021-08-20
Removed unnecessary normalization of the created quaternion in from_axis_angle

dhyams:
Removed unnecessary normalization of the created quaternion in the from_axis_angle Quaternion method. The quaternion is normalized by construction.

Foregoing these normalizations is sometimes a key operation to be able to successfully simplify complex algebraic output from these routines.

References to other Issues or PRs
None

Brief description of what is fixed or changed
Removed unnecessary normalization of the created quaternion in the from_axis_angle Quaternion method. The quaternion is normalized by construction.

Other comments
Foregoing these normalizations is sometimes a key operation to be able to successfully simplify complex algebraic output from these routines.

Release Notes

NO ENTRY

### DIFF
--- a/sympy/algebras/tests/test_quaternion.py
+++ b/sympy/algebras/tests/test_quaternion.py
@@ -1,5 +1,5 @@
 from sympy import symbols, re, im, sign, I, Abs, Symbol, \
-     cos, sin, sqrt, conjugate, log, acos, E, pi, \
+     cos, sin, sqrt, conjugate, log, acos, asin, E, pi, \
      Matrix, diff, integrate, trigsimp, S, Rational
 from sympy.algebras.quaternion import Quaternion
 from sympy.testing.pytest import raises
@@ -24,7 +24,6 @@ def test_quaternion_construction():
     raises(ValueError, lambda: Quaternion(w, x, nc, z))
 
 
-
 def test_quaternion_axis_angle():
 
     test_data = [ # axis, angle, expected_quaternion
@@ -41,6 +40,14 @@ def test_quaternion_axis_angle():
 
     for axis, angle, expected in test_data:
         assert Quaternion.from_axis_angle(axis, angle) == Quaternion(*expected)
+
+
+def test_quaternion_axis_angle_simplification():
+    result = Quaternion.from_axis_angle((1, 2, 3), asin(4))
+    assert result.a == cos(asin(4)/2)
+    assert result.b == sqrt(14)*sin(asin(4)/2)/14
+    assert result.c == sqrt(14)*sin(asin(4)/2)/7
+    assert result.d == 3*sqrt(14)*sin(asin(4)/2)/14
 
 def test_quaternion_complex_real_addition():
     a = symbols("a", complex=True)
--- a/sympy/algebras/tests/test_quaternion.py
+++ b/sympy/algebras/tests/test_quaternion.py
@@ -24,6 +24,24 @@ def test_quaternion_construction():
     raises(ValueError, lambda: Quaternion(w, x, nc, z))
 
 
+
+def test_quaternion_axis_angle():
+
+    test_data = [ # axis, angle, expected_quaternion
+        ((1, 0, 0), 0, (1, 0, 0, 0)),
+        ((1, 0, 0), pi/2, (sqrt(2)/2, sqrt(2)/2, 0, 0)),
+        ((0, 1, 0), pi/2, (sqrt(2)/2, 0, sqrt(2)/2, 0)),
+        ((0, 0, 1), pi/2, (sqrt(2)/2, 0, 0, sqrt(2)/2)),
+        ((1, 0, 0), pi, (0, 1, 0, 0)),
+        ((0, 1, 0), pi, (0, 0, 1, 0)),
+        ((0, 0, 1), pi, (0, 0, 0, 1)),
+        ((1, 1, 1), pi, (0, 1/sqrt(3),1/sqrt(3),1/sqrt(3))),
+        ((sqrt(3)/3, sqrt(3)/3, sqrt(3)/3), pi*2/3, (S.Half, S.Half, S.Half, S.Half))
+    ]
+
+    for axis, angle, expected in test_data:
+        assert Quaternion.from_axis_angle(axis, angle) == Quaternion(*expected)
+
 def test_quaternion_complex_real_addition():
     a = symbols("a", complex=True)
     b = symbols("b", real=True)
--- a/.mailmap
+++ b/.mailmap
@@ -583,3 +583,4 @@ Rahil Parikh <r.parikh@somaiya.edu> rprkh <r.parikh@somaiya.edu>
 Yaser AlOsh <yaseralosh@outlook.com> Yaser <yaseralosh@outlook.com>
 Joannah Nanjekye <joannah.nanjekye@ibm.com> nanjekyejoannah <joannah.nanjekye@ibm.com>
 Kshitij <kshitijparwani.mat18@itbhu.ac.in> Kshitij Parwani <44468674+P-Kshitij@users.noreply.github.com>
+Daniel Hyams <dhyams@gmail.com> dhyams <dhyams@gmail.com>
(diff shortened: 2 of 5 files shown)


oscarbenjamin:
Can you add some test code in `sympy/algebras/tests/test_quaternion.py`?

dhyams:
This pull request derived from #21902

dhyams:
added unit tests as requested.

sympy/algebras/tests/test_quaternion.py
```diff
@@ -23,6 +24,22 @@ def test_quaternion_construction():
     nc = Symbol('nc', commutative=False)
     raises(ValueError, lambda: Quaternion(w, x, nc, z))
 
+@pytest.mark.parametrize(
```
oscarbenjamin, line 27: SymPy has its own test runner which is sort of a copy of pytest. While it is possible to use pytest with the sympy codebase (I always do) we do need to make sure that the tests run without pytest i.e. using the `bin/test` script when `pytest` is not installed.

dhyams:
Oh, so is there not a way to do a parametrize with this mechanism as in pytest?

oscarbenjamin:
> Oh, so is there not a way to do a parametrize with this mechanism as in pytest?

No, there isn't: #15497 (CC @asmeurer)

dhyams:
unit tests updated to not use pytest.parametrize.

sympy/algebras/tests/test_quaternion.py
```diff
@@ -24,6 +24,24 @@ def test_quaternion_construction():
...
+        ((0, 0, 1), pi/2, (sqrt(2)/2, 0, 0, sqrt(2)/2)),
+        ((1, 0, 0), pi, (0, 1, 0, 0)),
+        ((0, 1, 0), pi, (0, 0, 1, 0)),
+        ((0, 0, 1), pi, (0, 0, 0, 1)),
+        ((1, 1, 1), pi, (0, 1/sqrt(3),1/sqrt(3),1/sqrt(3))),
+        ((sqrt(3)/3, sqrt(3)/3, sqrt(3)/3), pi*2/3, (S.Half, S.Half, S.Half, S.Half))
+    ]
+
+    for axis, angle, expected in test_data:
+        assert Quaternion.from_axis_angle(axis, angle) == Quaternion(*expected)
```
oscarbenjamin, line 43: It seems like these tests all pass on the current sympy master. Can you add a test for the *changed* behaviour i.e. a test that fails on master but passes with the changes in this PR?

dhyams: That won't be possible because this change has zero impact to the result.  I can think on it a little to see if I can contrive a case where the simplification doesn't work well due to the normalization.  But I wouldn't hold my breath ;)

oscarbenjamin: When you say "this change has zero impact" I'm left wondering why the change should be made... What exactly does this do?

oscarbenjamin:
This statement implies that these changes have some impact:

> Foregoing these normalizations is sometimes a key operation to be able to successfully simplify complex algebraic output from these routines.

What impact is that? (Is something better or faster?)

dhyams:
Yes it is:
  1) better.  Does not inject unneeded algebraic expressions into the result that are hard to simplify, or sympy fails to simplify, later.  This is how I ran across the problem in the first place.
  2) faster; there is no call to normalize() at all, because it's not needed.

This change is something like finding the following line of code:

a  = b / 2 * 2

and simplifying it to 

a = b

I will put in a good faith effort to create a case that it helps for, but the problem is that my own example is so complex it would not make for a very good unit test.

oscarbenjamin:
> 1. better.  Does not inject unneeded algebraic expressions into the result that are hard to simplify, or sympy fails to simplify, later.  This is how I ran across the problem in the first place.

That suggests that a test could be written that would demonstrate the betterness :)

mgeier (approved):

oscarbenjamin:
> I will put in a good faith effort to create a case that it helps for, but the problem is that my own example is so complex it would not make for a very good unit test.

I think I see the problem now. The `normalize` method calls `norm` which calls `trigsimp`. So in simple examples the normalization is simplified away to dividing by 1. The issue arises when trigsimp is unable to simplify the norm to one. By definition then any test for this would actually be testing the failure of trigsimp.

Perhaps it would be good to open an issue about trigsimp for the complicated example then.

An example where `norm` does not simplify is:
```python
In [23]: Quaternion.from_axis_angle((1, 2, 3), asin(4))
Out[23]: cos(asin(4)/2) + sqrt(14)*sin(asin(4)/2)/14*i + sqrt(14)*sin(asin(4)/2)/7*j + 3*sqrt(14)*sin(asin(4)/2)/14*k

In [24]: Quaternion.from_axis_angle((1, 2, 3), asin(4)).norm()
Out[24]: 
    _______________________
   ╱ 1 - √15⋅ⅈ   1 + √15⋅ⅈ 
  ╱  ───────── + ───────── 
╲╱       2           2  
```
Note that `radsimp` can simplify this:
```python
In [27]: radsimp(Quaternion.from_axis_angle((1, 2, 3), asin(4)).norm())
Out[27]: 1
```
So perhaps it would actually be good to use `radsimp` as well. Ideally simplification would not be done automatically in `norm` though or at least there should be some way to disable it.

dhyams:
I can't tell in the above comment that you agree that this PR should be merged.  I think it should, as there is no need to call normalize() to begin with.

oscarbenjamin:
> I can't tell in the above comment that you agree that this PR should be merged

We were discussing coming up with a test for the changed behaviour. You said you weren't sure that you could so I found this case:
```python
In [3]: Quaternion.from_axis_angle((1, 2, 3), asin(4))
Out[3]: 
cos(asin(4)/2)/sqrt((1 - sqrt(15)*I)/2 + (1 + sqrt(15)*I)/2) + sqrt(14)*sin(asin(4)/2)/(14*sqrt((1 - sqrt(15)*I)/2 + (1 + sqrt(15)*I)/2))*i + sqrt(14)*sin(a
sin(4)/2)/(7*sqrt((1 - sqrt(15)*I)/2 + (1 + sqrt(15)*I)/2))*j + 3*sqrt(14)*sin(asin(4)/2)/(14*sqrt((1 - sqrt(15)*I)/2 + (1 + sqrt(15)*I)/2))*k
```
With the PR that gives:
```python
In [1]: Quaternion.from_axis_angle((1, 2, 3), asin(4))
Out[1]: cos(asin(4)/2) + sqrt(14)*sin(asin(4)/2)/14*i + sqrt(14)*sin(asin(4)/2)/7*j + 3*sqrt(14)*sin(asin(4)/2)/14*k
```

dhyams:
Oh, I see, forgive me for being dense.  Thanks!

oscarbenjamin:
I think there also needs to be a line like:
```
Daniel Hyams <dhyams@gmail.com> Daniel Hyams <daniel.hyams@peopletec.com>
```
This is because all commits need to map to a unique name/email address.

oscarbenjamin:
Rebasing this I get the merge conflict:
```
<<<<<<< HEAD
Kshitij <kshitijparwani.mat18@itbhu.ac.in> Kshitij Parwani <44468674+P-Kshitij@users.noreply.github.com>
=======
Daniel Hyams <dhyams@gmail.com> dhyams <dhyams@gmail.com>
>>>>>>> updated mailmap
```
This is what was discussed on the mailing list:
https://groups.google.com/g/sympy/c/3DXOKgoeX24/m/1eDTik55BAAJ

The problem is that new names have to go at the end of the AUTHORS file so as soon as one new contributor PR is merged any that were opened before have to be rebased over.

oscarbenjamin:
Oh, I didn't realise the other problem which is that the number of authors is listed explicitly:
```diff
diff --git a/AUTHORS b/AUTHORS
index c0ff31c722..0eed4cb71a 100644
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,7 +4,7 @@ those who explicitly didn't want to be mentioned. People with a * next
 to their names are not found in the metadata of the git history. This
 file is generated automatically by running `./bin/authors_update.py`.
 
-There are a total of 1092 authors.
+There are a total of 1093 authors.
 
 Ondřej Čertík <ondrej@certik.cz>
 Fabian Pedregosa <fabian@fseoane.net>
```
I've added a line to show the diff when the script fails. Just checking that works and then I'll fix the AUTHORS file.

oscarbenjamin:
Thanks!

PR #21918, opened 2021-08-21
Fix correctness issues of unify

### LINKED ISSUE #21917: Returning idempotent most general unifier with `unify`
SymPy's unify returns just an mgu (most general unifier), but not an idempotent mgu (mgu which has no variables on right hand sides that appears on left hand sides)
I think that this comes from the lack of the elimination steps both in the implemented algorithm and the reference (_Artificial intelligence : a modern approach_), but I'm not sure if this can be trivially implemented by applying the elimination step posteriori or it needs the modification of the algorithm internally.

```python
from sympy.unify.core import Compound, Variable, unify

x = Variable('x')
y = Variable('y')
a = Compound('a', ())

gen = unify(Compound('f', (x, y)), Compound('f', (Compound('h', (a,)), x)))
mgu = next(gen)
for k, v in mgu.items():
    print(f'{k} : {v}')
```
```
Variable(x) : h[a[]]
Variable(y) : Variable(x)
```

sylee957:

Fixes #21917

Unify should satisfy the identity `all(substitute(A, s) == substitute(B, s) for s in unify(A, B))` but the previous results does not guarantee this for some expressions.
This is explainable by the lack of variable elimination step where similar references all has.
I also find that the second statement of `unify_var` is missing as in the original source, but I see the original textbook implementation was also questioned by https://github.com/aimacode/aima-python/issues/1053 so I ported the changes from the update of the reference source.

This also fixes the issue with function with commutativity, but not with associativity $C$
$C(1, 2, 3) = C(x, y, z)$ should give the 6 permutation of the result.

#### Release Notes

- unify
  - Fixed wrong results of `unify` not giving the idempotent unifier, and not giving correct result for function symbols only with `is_commutative` defined

### DIFF
--- a/sympy/unify/core.py
+++ b/sympy/unify/core.py
@@ -73,18 +73,26 @@ class CondVariable:
         return "CondVariable(%s)" % str(self.arg)
 
 def unify(x, y, s=None, **fns):
-    """ Unify two expressions.
+    """Returns idempotent most general unifier of the pairs of expressions
 
     Parameters
     ==========
 
-        x, y - expression trees containing leaves, Compounds and Variables.
-        s    - a mapping of variables to subtrees.
+    x
+        Expression trees containing leaves, Compounds and Variables.
+
+    y
+        Expression trees containing leaves, Compounds and Variables.
+
+    s : dict, optional
+        A mapping of variables to subtrees.
 
     Returns
     =======
 
-        lazy sequence of mappings {Variable: subtree}
+    generator
+        Lazy sequence of dictionaries in the form of
+        ``Variable: Union[Variable, Compound]``
 
     Examples
     ========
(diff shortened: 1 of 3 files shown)


sympy/unify/core.py
```diff
@@ -1,179 +1,358 @@
...
+http://aima.cs.berkeley.edu/python/logic.html
 
-Artificial Intelligence: A Modern Approach by Stuart Russel and Peter Norvig
-Second edition, section 9.2, page 276
+We extended the original algorithm with the capability to
 
-It is modified in the following ways:
+- Handle associative-commutative unification capability.
+- Designed type system to make this work across almosst every objects of
+  python by composition.
```
smichr, line 10:
```suggestion
- Designed type system to make this work with almost all Python objects by composition.
```

smichr: (not exactly sure what you mean so this might still not quite capture the intent)

sympy/unify/core.py
```diff
@@ -1,179 +1,358 @@
...
+
+    >>> Compound('f', (x,))
+    f[x]
+    >>> Compound('f', (x, y))
+    f[x, y]
+
+    How to represent constants?
+
+    The standard way of encoding 'constant' as term is defining
+    them as nullary compound term.
```
smichr, line 74:
```suggestion
    A 'constant' term is defined as a nullary compound term.
```

sympy/unify/core.py
```diff
@@ -1,179 +1,358 @@
...
+    2[]
+
+    For legacy reason, the following behavior is also supported
+
+    >>> Compound('Add', (1, 2))
+    Add[1, 2]
+
+    However, the identity between the nullary compound term its head is
+    not preserved, so you need to be consistent in the way how to
+    encode constants into the term.
```
smichr, line 88:
```suggestion
    However, the identity between the nullary compound term and its head is
    not preserved, so you need to be consistent in how constants are encoded
    into the term.
```

sympy/unify/core.py
```diff
@@ -1,179 +1,358 @@
...
+
+class Variable(VariableBase):
+    """A variable term
 
-def unify(x, y, s=None, **fns):
-    """ Unify two expressions.
+    Explanation
+    ===========
+
+    It is similar as :class:`Symbol` in sympy, however, unlike sympy's
```
smichr, line 161:
```suggestion
    It is similar as :class:`Symbol` in SymPy, however, unlike SymPy's
```

sympy/unify/core.py
```diff
@@ -1,179 +1,358 @@
...
     >>> pattern = Compound("Add", ("x", Variable("a")))
     >>> next(unify(expr, pattern, {}))
-    {Variable(a): 'y'}
+    {a: 'y'}
+
+    Notes
+    =====
+
+    Also, this is not a standard approach of associative-commutative
+    unification, but rather be list unification, which has similar
```
smichr, line 232:
```suggestion
    unification, but rather of list unification, which has similar
```

Issue #16643, opened 2019-04-13
solveset weird behaviour


proy87:
Why `solveset(x**2*sin(x), x)` outputs ConditionSet instead of `pi*n`?
 `solveset(x**2*sin(x), x, S.Reals)` gives the desired result.

oscargus:
solveset does not know how to solve the equation for complex x (and as far as solveset knows, x is complex).

oscargus:
Btw, this also works:

```
x=Symbol('x', real=True)
solveset(x**2*sin(x), x)
```

proy87:
Very weird.
Anyway, 
x=Symbol('x', real=True)
solveset((x**2+1)*sin(x)*sqrt(x))
Gives imaginary solutions and incorrect imageset (since x should be nonnegative).

oscargus:
Well, the assumptions are indeed not propagating perfectly. However, 
`solveset((x**2+1)*sin(x)*sqrt(x), x, S.Reals)`
at least does not give the imaginary solutions, although still the negative ones are provided.

gschintgen:
With master:
```
In [1]: solveset(x**2*sin(x), x)
Out[1]: {2⋅n⋅π | n ∊ ℤ} ∪ {2⋅n⋅π + π | n ∊ ℤ}

In [2]: x=Symbol('x', real=True)

In [3]: solveset((x**2+1)*sin(x)*sqrt(x))
Out[3]: {-ⅈ, ⅈ} ∪ {2⋅n⋅π | n ∊ ℤ} ∪ {2⋅n⋅π + π | n ∊ ℤ}
```
This is my take on the different issues:

- `Out[1]` not simplified, but that's a known issue (#12340)

- `Out[3]` includes imaginary solutions, but that's by design:
"What is this domain argument about?
Solveset is designed to be independent of the assumptions on the variable being solved for and instead, uses the domain argument to decide the solver to dispatch the equation to, namely solveset_real or solveset_complex. It’s unlike the old solve which considers the assumption on the variable."
https://docs.sympy.org/latest/modules/solvers/solveset.html#what-is-this-domain-argument-about

- `Out[3]` includes negative solutions. That's incorrect and the reason I'm leaving this issue open. (There are other issues filed mentioning incomplete domain handling in `solveset`, especially with `ImageSets`) I'll adapt the title accordingly.

oscargus:
Point 3 is related to #16648

PR #21925, opened 2021-08-21
Fixed solveset and added tests for already solved issues

### LINKED ISSUE #10085: invert_real returns non-real value for exp(x)
``` python
>>> from sympy import *; from sympy.abc import x
>>> from sympy.solvers.solveset import invert_real
>>> invert_real(exp(x),0,x)
(x, {zoo})
```

Probably empty set should be returned?

### LINKED ISSUE #10086: solve generates a NaN comparison error
```python
>>> solve(Abs(x-4)-(x-1)/(x+2)-(x-3)/(x+1))
  File "sympy\solvers\solvers.py", line 1194, in solve
    solution.sort(key=default_sort_key)
  File "sympy\core\numbers.py", line 1807, in __lt__
    return Rational.__lt__(self, other)
  File "sympy\core\numbers.py", line 1485, in __lt__
    return Expr.__lt__(expr, other)
  File "sympy\core\expr.py", line 301, in __lt__
    raise TypeError("Invalid NaN comparison")
TypeError: Invalid NaN comparison
```

It looks like the problem is that it is trying to sort the solution which contains a nan. One can see the solutions it is working with by requesting a set:

```
>>> print filldedent(solve(Abs(x-4)-(x-1)/(x+2)-(x-3)/(x+1), set=True))

([x], set([(1 + sqrt(6),), (Piecewise((2**(2/3)*(-4 + (1 +
sqrt(3)*I)*(2 + (-1 - sqrt(3)*I)*(6 + 2*sqrt(7)*I)**(1/3))*(6 +
2*sqrt(7)*I)**(1/3)/4)/((1 + sqrt(3)*I)*(3 + sqrt(7)*I)**(1/3)),
2**(2/3)*(4 + (1 + sqrt(3)*I)*(6 + 2*sqrt(7)*I)**(1/3)*(6 + (1 +
sqrt(3)*I)*(6 + 2*sqrt(7)*I)**(1/3))/4)/((1 + sqrt(3)*I)*(3 +
sqrt(7)*I)**(1/3)) <= 0), (nan, True)),), (-sqrt(6) + 1,), (nan,),
(-3,), (Piecewise((1 + 2*2**(2/3)/(3 + sqrt(7)*I)**(1/3) + 2**(1/3)*(3
+ sqrt(7)*I)**(1/3), -3 + 2*2**(2/3)/(3 + sqrt(7)*I)**(1/3) +
2**(1/3)*(3 + sqrt(7)*I)**(1/3) >= 0), (nan, True)),)]))
```

It's interesting to see that a Piecewise is present because it couldn't decide the sign of an expression.

BTW, the actual solutions (confirmed visually) are:

```
>>> nsolve(eq,-3)
mpf('-3.0')
>>> nsolve(eq,-1.5)
mpf('-1.4494897427831781')
>>> nsolve(eq,3.2)
mpf('3.4494897427831781')
>>> nsolve(eq,5)
mpf('4.8844837019393323')
```

### LINKED ISSUE #10876: solveset errors if numerator doesn't have symbol
```
>>> solveset(1/sqrt(x),x)
  File "sympy\solvers\solveset.py", line 792, in solveset
    return _solveset(f, symbol, domain, _check=True)
  File "sympy\solvers\solveset.py", line 613, in _solveset
    solver)
  File "sympy\solvers\solveset.py", line 492, in _solve_radical
    eq, cov = unrad(f)
TypeError: 'NoneType' object is not iterable
```

...but not always:

```
>>> solveset(1/x**2,x)
EmptySet()
```

### LINKED ISSUE #12032: Error in solveset for polynomial equation
The "solve" gives fastly the result of:
`solve(x**4 +x -1, x)`
whereas "solveset" for the same equation fails:
`solveset(x**4 +x -1, x)`
It seems to be traped in infinite loop.

### LINKED ISSUE #13396: solveset() unable to solve -2*y*exp(-x**2 - y**2)*Abs(x)
```python
>>> expr = -2*y*exp(-x**2 - y**2)*Abs(x)
>>> solveset(expr, y, domain=S.Reals)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/solvers/solveset.py", line 925, in solveset
    return _solveset(f, symbol, domain, _check=True)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/solvers/solveset.py", line 699, in _solveset
    result = Union(*[solver(m, symbol) for m in f.args])
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/solvers/solveset.py", line 699, in <listcomp>
    result = Union(*[solver(m, symbol) for m in f.args])
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/solvers/solveset.py", line 678, in <lambda>
    solver = lambda f, x, domain=domain: _solveset(f, x, domain)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/solvers/solveset.py", line 715, in _solveset
    lhs, rhs_s = inverter(f, 0, symbol)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/solvers/solveset.py", line 683, in <lambda>
    inverter = lambda f, rhs, symbol: inverter_func(f, rhs, symbol, domain)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/solvers/solveset.py", line 118, in invert_real
    return _invert(f_x, y, x, domain)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/solvers/solveset.py", line 107, in _invert
    return x1, s.intersection(domain)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/sets/sets.py", line 114, in intersection
    return self.intersect(other)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/sets/sets.py", line 108, in intersect
    return Intersection(self, other)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/sets/sets.py", line 1497, in __new__
    args = list(ordered(args, Set._infimum_key))
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/core/compatibility.py", line 608, in ordered
    d[f(a)].append(a)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/sets/sets.py", line 58, in _infimum_key
    infimum = expr.inf
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/sets/sets.py", line 263, in inf
    return self._inf
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/sets/sets.py", line 2022, in _inf
    return Min(*self)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/functions/elementary/miscellaneous.py", line 343, in __new__
    args = cls._find_localzeros(set(_args), **assumptions)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/functions/elementary/miscellaneous.py", line 425, in _find_localzeros
    con = cls._is_connected(v, z)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/functions/elementary/miscellaneous.py", line 447, in _is_connected
    r = hit(x >= y, Max, Min)
  File "/Users/arihantparsoya/anaconda/lib/python3.6/site-packages/sympy-1.1.2.dev0-py3.6.egg/sympy/core/expr.py", line 255, in __ge__
    raise TypeError("Invalid comparison of complex %s" % me)
TypeError: Invalid comparison of complex sqrt(-x**2 + zoo)
```

### LINKED ISSUE #14541: solveset fails if floating-point numbers are used
For example `solveset(sqrt(-x**2 - 2), x)` returns` {-sqrt(2)*I, sqrt(2)*I}`, as expected. However, when instead of integer a floating point is used (`solveset(sqrt(-x**2 - 2.0), x)`), then `EmptySet()` is returned.

### LINKED ISSUE #14642: ConditionSet solveset cannot calculate a complex I polynomials, return conditionSet
case 1：
```
from sympy import *
x = symbols('x',dtype = complex)
n1 = 0.5*x**3+x**2+0.5
solveset(n1,x)
```
_**result**_ : `{-2.20556943040059, 0.102784715200295 - 0.665456951152813*I, 0.102784715200295 + 0.665456951152813*I}`

case 2：
```
from sympy import *
x = symbols('x',dtype = complex)
n1 = 0.5*x**3+x**2+0.5+I  #**add I in the Polynomials**
solveset(n1,x)
```
**_result_**: `ConditionSet(0.5*x**3 + x**2 + 0.5 + I, x, Complexes(S.Reals x S.Reals, False))`

1. what is the means "Complexes(S.Reals x S.Reals, False)"
2. why case 2 cannot calculate a result ????
thanks you very much !

### LINKED ISSUE #17580: Weird result in solveset
`solveset(1/(1-x**3)**2, x, S.Reals)` returns `Intersection`

### LINKED ISSUE #17906: Runtime error in transolve
`solveset(7**(x^2)-49**x, x)` correctly gives `{0, 2}`, while `solveset(7**(x^2-80)-49**x, x)` gives runtime error. That's because inside transolve `factor` rewrites 7**80 and `powdenest` inside transolve rewrites it back.

### LINKED ISSUE #17940: Solveset returns ConditionSet for a simple exponential equation
`exp(exp(x))-5`

### LINKED ISSUE #19506: solveset mishandles equation with arg, raises TypeError
```
eq = arg(x+I)
solveset(eq)
```
results in:
```
<ipython-input-55-69ecc2ebbd2c> in <module>
      1 get_ipython().run_line_magic('xmode', 'Verbose')
      2 eq = arg(x+I)
----> 3 solveset(eq)
        global solveset = <function solveset at 0x7fa7937ac378>
        global eq = arg(x + I)

/opt/tljh/user/lib/python3.6/site-packages/sympy/solvers/solveset.py in solveset(f=arg(x + I), symbol=x, domain=S.Complexes)
   2009     f = piecewise_fold(f)
   2010 
-> 2011     return _solveset(f, symbol, domain, _check=True)
        global _solveset = <function _solveset at 0x7fa7937abd90>
        f = arg(x + I)
        symbol = x
        domain = S.Complexes
        global _check = undefined
   2012 
   2013 

/opt/tljh/user/lib/python3.6/site-packages/sympy/solvers/solveset.py in _solveset(f=arg(x + I), symbol=x, domain=S.Complexes, _check=True)
    914     elif isinstance(f, arg):
    915         a = f.args[0]
--> 916         result = solveset_real(a > 0, symbol)
        result = EmptySet
        global solveset_real = <function solveset_real at 0x7fa7937ac400>
        a = x + I
        symbol = x
    917     elif f.is_Piecewise:
    918         result = EmptySet

/opt/tljh/user/lib/python3.6/site-packages/sympy/solvers/solveset.py in solveset_real(f=x + I > 0, symbol=x)
   2013 
   2014 def solveset_real(f, symbol):
-> 2015     return solveset(f, symbol, S.Reals)
        global solveset = <function solveset at 0x7fa7937ac378>
        f = x + I > 0
        symbol = x
        global S.Reals = Reals
   2016 
   2017 

/opt/tljh/user/lib/python3.6/site-packages/sympy/solvers/solveset.py in solveset(f=x + I > 0, symbol=x, domain=Reals)
   1991             try:
   1992                 r = Dummy('r', **assumptions)
-> 1993                 return solveset(f.xreplace({symbol: r}), r, domain
        global solveset = <function solveset at 0x7fa7937ac378>
        f.xreplace = <bound method Basic.xreplace of x + I > 0>
        symbol = x
        r = _r
        domain.xreplace = <bound method Basic.xreplace of Reals>
   1994                     ).xreplace({r: symbol})
   1995             except InconsistentAssumptions:

/opt/tljh/user/lib/python3.6/site-packages/sympy/core/basic.py in xreplace(self=x + I > 0, rule={x: _r})
   1190 
   1191         """
-> 1192         value, _ = self._xreplace(rule)
        value = undefined
        _ = undefined
        self._xreplace = <bound method Basic._xreplace of x + I > 0>
        rule = {x: _r}
   1193         return value
   1194 

/opt/tljh/user/lib/python3.6/site-packages/sympy/core/basic.py in _xreplace(self=x + I > 0, rule={x: _r})
   1212             args = tuple(args)
   1213             if changed:
-> 1214                 return self.func(*args), True
        self.func = <class 'sympy.core.relational.StrictGreaterThan'>
        args = (_r + I, 0)
   1215         return self, False
   1216 

/opt/tljh/user/lib/python3.6/site-packages/sympy/core/relational.py in __new__(cls=<class 'sympy.core.relational.StrictGreaterThan'>, lhs=_r + I, rhs=0, **options={})
    745             # exception).  In that case, it must call us with
    746             # `evaluate=False` to prevent infinite recursion.
--> 747             r = cls._eval_relation(lhs, rhs)
        r = undefined
        cls._eval_relation = <bound method StrictGreaterThan._eval_relation of <class 'sympy.core.relational.StrictGreaterThan'>>
        lhs = _r + I
        rhs = 0
    748             if r is not None:
    749                 return r

/opt/tljh/user/lib/python3.6/site-packages/sympy/core/relational.py in _eval_relation(cls=<class 'sympy.core.relational.StrictGreaterThan'>, lhs=_r + I, rhs=0)
   1053     def _eval_relation(cls, lhs, rhs):
   1054         # We don't use the op symbol here: workaround issue #7951
-> 1055         return _sympify(lhs.__gt__(rhs))
        global _sympify = <function _sympify at 0x7fa7a8646620>
        lhs.__gt__ = <bound method Expr.__gt__ of _r + I>
        rhs = 0
   1056 
   1057 

/opt/tljh/user/lib/python3.6/site-packages/sympy/core/expr.py in __gt__(self=_r + I, other=0)
    393     def __gt__(self, other):
    394         from sympy import StrictGreaterThan
--> 395         return self._cmp(other, ">", StrictGreaterThan)
        self._cmp = <bound method Expr._cmp of _r + I>
        other = 0
        StrictGreaterThan = <class 'sympy.core.relational.StrictGreaterThan'>
    396 
    397     def __lt__(self, other):

/opt/tljh/user/lib/python3.6/site-packages/sympy/core/expr.py in _cmp(self=_r + I, other=0, op='>', cls=<class 'sympy.core.relational.StrictGreaterThan'>)
    338         for me in (self, other):
    339             if me.is_extended_real is False:
--> 340                 raise TypeError("Invalid comparison of non-real %s" % me)
        global TypeError = undefined
        me = _r + I
    341             if me is S.NaN:
    342                 raise TypeError("Invalid NaN comparison")

TypeError: Invalid comparison of non-real _r + I
```

oscargus:

Closes #10085 (added test, solved elsewhere)
Closes #10876 (added test, solved elsewhere)
Closes #12032 (added test, solved elsewhere)
Fixes #13396 (solved underlying problem)
Added test for #13961 (gives two identical solutions)
Closes #14541 (added test, solved elsewhere)
Closes #14642 (added test, solved elsewhere)
Added test for #16643 (original issue solved, but not follow-up equation)
Fixes #17580 (solved underlying problem)
Closes #17906 (added test, solved elsewhere(?))
Fixes #17940 (solved underlying problem)
Fixes #19506 (solved underlying problem)

Added tests and closed where appropriate.

#### Release Notes

* solvers
    * `solveset` correctly solves `arg` when the argument is complex.

### DIFF
--- a/sympy/solvers/solveset.py
+++ b/sympy/solvers/solveset.py
@@ -197,6 +197,11 @@ def invert_real(f_x, y, x, domain=S.Reals):
 def _invert_real(f, g_ys, symbol):
     """Helper function for _invert."""
 
+    # Some inverse functions evaluate to zoo, remove if present
+    fs = FiniteSet(S.ComplexInfinity)
+    if fs.is_subset(g_ys):
+        g_ys -= fs
+
     if f == symbol:
         return (f, g_ys)
 
@@ -295,16 +300,16 @@ def _invert_real(f, g_ys, symbol):
     if isinstance(f, TrigonometricFunction):
         if isinstance(g_ys, FiniteSet):
             def inv(trig):
-                if isinstance(f, (sin, csc)):
-                    F = asin if isinstance(f, sin) else acsc
+                if isinstance(trig, (sin, csc)):
+                    F = asin if isinstance(trig, sin) else acsc
                     return (lambda a: n*pi + (-1)**n*F(a),)
-                if isinstance(f, (cos, sec)):
-                    F = acos if isinstance(f, cos) else asec
+                if isinstance(trig, (cos, sec)):
+                    F = acos if isinstance(trig, cos) else asec
                     return (
                         lambda a: 2*n*pi + F(a),
                         lambda a: 2*n*pi - F(a),)
-                if isinstance(f, (tan, cot)):
-                    return (lambda a: n*pi + f.inverse()(a),)
+                if isinstance(trig, (tan, cot)):
+                    return (lambda a: n*pi + trig.inverse()(a),)
 
             n = Dummy('n', integer=True)
             invs = S.EmptySet
@@ -708,7 +713,7 @@ def _solve_trig1(f, symbol, domain):
 def _solve_trig2(f, symbol, domain):
     """Secondary helper to solve trigonometric equations,
     called when first helper fails """
-    from sympy import ilcm, expand_trig, degree
+    from sympy import ilcm, expand_trig
     f = trigsimp(f)
     f_original = f
     trig_functions = f.atoms(sin, cos, tan, sec, cot, csc)
@@ -862,7 +867,7 @@ def _solve_radical(f, unradf, symbol, solveset_solver):
         result = Union(*[imageset(Lambda(y, g_y), f_y_sols)
                          for g_y in g_y_s])
 
-    if isinstance(result, Complement) or isinstance(result,ConditionSet):
+    if isinstance(result, (Complement, ConditionSet)):
         solution_set = result
     else:
         f_set = []  # solutions for FiniteSet
@@ -1025,8 +1030,10 @@ def _solveset(f, symbol, domain, _check=False):
             _is_function_class_equation(HyperbolicFunction, f, symbol):
         result = _solve_trig(f, symbol, domain)
     elif isinstance(f, arg):
+        from sympy.functions.elementary.complexes import re, im
         a = f.args[0]
-        result = solveset_real(a > 0, symbol)
+        result = Intersection(_solveset(re(a) > 0, symbol, domain),
+                              _solveset(im(a), symbol, domain))
     elif f.is_Piecewise:
         expr_set_pairs = f.as_expr_set_pairs(domain)
         for (expr, in_set) in expr_set_pairs:
@@ -1116,7 +1123,16 @@ def _solveset(f, symbol, domain, _check=False):
         if isinstance(result, ConditionSet):
             # it wasn't solved or has enumerated all conditions
             # -- leave it alone
-            return result
+            if domain is S.Complexes:
+                return result
+            # Add domain
+            new_res = Intersection(result, domain)
+            # Check if the domain is really contributing
+            if new_res.is_subset(result):
+                return new_res
+            else:
+                return result
+
 
         # whittle away all but the symbol-containing core
         # to use this for testing
@@ -2935,9 +2951,6 @@ def substitution(system, symbols, result=[{}], known_symbols=[],
       ImageSet(Lambda(_n, sqrt(-exp(2*x) + sin(2*_n*I*pi - log(3)))), Integers))}
 
     """
-
-    from sympy import Complement
-    from sympy.core.compatibility import is_sequence
 
     if not system:
         return S.EmptySet
--- a/sympy/solvers/solveset.py
+++ b/sympy/solvers/solveset.py
@@ -197,11 +197,6 @@ def invert_real(f_x, y, x, domain=S.Reals):
 def _invert_real(f, g_ys, symbol):
     """Helper function for _invert."""
 
-    # Some inverse functions evaluate to zoo, remove if present
-    fs = FiniteSet(S.ComplexInfinity)
-    if fs.is_subset(g_ys):
-        g_ys -= fs
-
     if f == symbol:
         return (f, g_ys)
 
--- a/sympy/solvers/tests/test_solveset.py
+++ b/sympy/solvers/tests/test_solveset.py
@@ -1,10 +1,12 @@
+from sympy.core.add import Add
 from sympy.core.containers import Tuple
 from sympy.core.compatibility import ordered
 from sympy.core.function import (Function, Lambda, nfloat, diff)
 from sympy.core.mod import Mod
-from sympy.core.numbers import (E, I, Rational, oo, pi, Integer)
-from sympy.core.relational import (Eq, Gt,
-    Ne, Ge)
+from sympy.core.mul import Mul
+from sympy.core.numbers import (E, I, Rational, oo, pi, Integer, zoo)
+from sympy.core.relational import (Eq, Gt, Ne, Ge)
+from sympy.core.power import Pow
 from sympy.core.singleton import S
 from sympy.core.symbol import (Dummy, Symbol, symbols)
 from sympy.functions.elementary.complexes import (Abs, arg, im, re, sign)
@@ -1392,6 +1394,10 @@ def test_issue_16577():
         Matrix([[2*a, 3*a + 4]]), Matrix([[5]]))
 
 
+def test_issue_10085():
+    assert invert_real(exp(x),0,x) == (x, S.EmptySet)
+
+
 def test_linsolve():
     x1, x2, x3, x4 = symbols('x1, x2, x3, x4')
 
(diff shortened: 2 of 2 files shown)


sympy/solvers/tests/test_solveset.py
```diff
@@ -1724,6 +1731,130 @@ def test_solve_nonlinear_trans():
     assert nonlinsolve([x**2 - y**2/exp(x)], [x, y]) == soln4
 
 
+def test_issue_14642():
+    x = symbols('x',dtype = complex)
```
oscarbenjamin, line 1735: What does `dtype=complex` do?

oscargus: No idea. The original code was like this...

oscarbenjamin: I guess it is supposed to be `complex=True` or something. It should be removed though because it doesn't do anything (ideally this would raise an error).

oscargus: OK! Will do!

oscargus:
I've updated the tests and original comment so that the status is correct.

sympy/solvers/tests/test_solveset.py
```diff
@@ -1724,6 +1731,150 @@ def test_solve_nonlinear_trans():
...
+@XFAIL
+def test_issue_13396_failing():
+    expr = -2*y*exp(-x**2 - y**2)*Abs(x)
+    # The second part may be empty
+    sol = Union(FiniteSet(Integer(0)), Intersection(S.Reals, FiniteSet(Pow(Add(Mul(Integer(-1),
+                Pow(x, Integer(2))), zoo), Rational(1, 2)), Mul(Integer(-1),
+                 Pow(Add(Mul(Integer(-1), Pow(x, Integer(2))), zoo), Rational(1, 2))))))
+
+     # Should work, but will not. See #21943
+    assert solveset(expr, y, domain=S.Reals) == sol
```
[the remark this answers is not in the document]
oscarbenjamin: Problem is that the args are flipped in `zoo - x**2`.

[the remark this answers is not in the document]
oscarbenjamin: I guess that the *cause* of the problem is evaluate=False being used somewhere...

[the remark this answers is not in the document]
oscargus: I tried to rewrite the expressions some time ago, but couldn't figure out where to put the `evaluate=False`. May have another go, but indeed it would be better to only get one solution...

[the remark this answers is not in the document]
oscarbenjamin: To recreate the output of `solveset` we need `evaluate=False` for the `Add` in `zoo - x**2`:
```python
In [15]: sol = Union(FiniteSet(Integer(0)), Intersection(S.Reals, FiniteSet(Pow(Add(Mul(Integer(-1),
    ...:             Pow(x, Integer(2))), zoo, evaluate=False), Rational(1, 2)), Mul(Integer(-1),
    ...:             Pow(Add(Mul(Integer(-1), Pow(x, Integer(2))), zoo, evaluate=False), Rational(1, 2))))))
    ...: 
    ...: expr = -2*y*exp(-x**2 - y**2)*Abs(x)
    ...: 
    ...: print(solveset(expr, y, Reals) == sol)
True
```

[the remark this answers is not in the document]
oscarbenjamin: It's the 2-arg Mul:
```python
In [6]: e = zoo - x**2

In [7]: e.args
Out[7]: 
⎛       2⎞
⎝zoo, -x ⎠

In [8]: e2 = -1 * e

In [9]: e2
Out[9]: 
 2      
x  + zoo

In [10]: e2.args
Out[10]: 
⎛ 2     ⎞
⎝x , zoo⎠

In [11]: e2 == e2.func(*e2.args)
Out[11]: False
```
It can be fixed with
```diff
diff --git a/sympy/core/mul.py b/sympy/core/mul.py
index 9933eb601a..960b5baf25 100644
--- a/sympy/core/mul.py
+++ b/sympy/core/mul.py
@@ -281,7 +281,7 @@ def flatten(cls, seq):
                         ar = a*r
                         if ar:
                             bargs.insert(0, ar)
-                        bargs = [Add._from_args(bargs)]
+                        bargs = [Add(*bargs)]
                         rv = bargs, [], None
             if rv:
                 return rv
```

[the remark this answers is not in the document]
smichr: or maybe `_unevaluated_Add(*args)` -- I read the issue but am not seeing enough context to understand the problem. `Add._from_args` is much faster than `Add` and since you are making this change in core it makes me nervous.

oscargus:
I added the `evalute=False` in the test with a comment about that they should ideally not be required.

Didn't want to change the core mul-code...

oscargus:
I have fixed the `zoo` problems in a slightly more general way than discussed in #22019.

Also fixed solving `arg` in a way that at least is correct, but may not always return "minimal" solutions. For example, `solveset(arg(x) , x)` returns

`Intersection(ConditionSet(x, Eq(im(x), 0), Complexes), ConditionSet(x, re(x) > 0, Complexes))`

where the first ConditionSet is identical to x being real.

On the other hand, `solveset(arg(x-I) , x)` now do not throw and error, but returns

`Intersection(ConditionSet(x, Eq(im(x) - 1, 0), Complexes), ConditionSet(x, re(x) > 0, Complexes))`

sympy/solvers/solveset.py
```diff
@@ -197,6 +197,11 @@ def invert_real(f_x, y, x, domain=S.Reals):
 def _invert_real(f, g_ys, symbol):
     """Helper function for _invert."""
 
+    # Some inverse functions evaluate to zoo, remove if present
+    fs = FiniteSet(S.ComplexInfinity)
+    if fs.is_subset(g_ys):
+        g_ys -= fs
```
oscarbenjamin, line 203: I'm not sure this does fix things in a more general way. It removes zoo but loses the reason for removing zoo.

oscargus: Well, the option is to check each and every inverse function that the input will not lead to zoo as you suggested. I cannot really see that keeps the reason better.

oscargus: My point is that once the function class is identified it will in almost all cases call `_invert_real` again with an evaluated version of the inverse function. So rather than checking for every evaluation if the input may yield `zoo`, which is a bit of a hassle and cases will have to be added once identified, it seems more general to check if the input to (the next call of) `_invert_real` is `zoo`. The only possible way that this may not work is if there are some functions that map `zoo` back to reals in a "controlled" way (not just inverting again). It doesn't seem to be any of those cases in the tests so far.

I added a test when I wrote this code to illustrate the point:
`assert solveset(atan(x**2 - y**2)-pi/2, y, S.Reals) == EmptySet()` which currently returns `Intersection({-sqrt(x**2 + zoo), sqrt(x**2 + zoo)}, Reals)`.

sympy/solvers/tests/test_solveset.py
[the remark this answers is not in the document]
smichr: > It can be fixed with

I think this is best. Otherwise it's dicey with special handling of objects in Add which get promoted to the coefficient position (like TensExpr, AccumBounds, ComplexInfinity).

And there is a difference between the handling of `-x` vs `-1*x`. The following diff will make the behavior more uniform:
```diff
diff --git a/sympy/core/mul.py b/sympy/core/mul.py
index 9933eb601a..1de76341f1 100644
--- a/sympy/core/mul.py
+++ b/sympy/core/mul.py
@@ -267,7 +267,7 @@ def flatten(cls, seq):
                 r, b = b.as_coeff_Mul()
                 if b.is_Add:
                     if r is not S.One:  # 2-arg hack
-                        # leave the Mul as a Mul?
+                        # 2*Mul(3, x + 1, evaluate=False) -> 6*(x + 1)
                         ar = a*r
                         if ar is S.One:
                             arb = b
@@ -275,14 +275,12 @@ def flatten(cls, seq):
                             arb = cls(a*r, b, evaluate=False)
                         rv = [arb], [], None
                     elif global_parameters.distribute and b.is_commutative:
-                        r, b = b.as_coeff_Add()
-                        bargs = [_keep_coeff(a, bi) for bi in Add.make_args(b)]
-                        _addsort(bargs)
+                        r, nonr = b.as_coeff_Add()
+                        bargs = [_keep_coeff(a, bi) for bi in Add.make_args(nonr)]
                         ar = a*r
                         if ar:
                             bargs.insert(0, ar)
-                        bargs = [Add._from_args(bargs)]
-                        rv = bargs, [], None
+                        rv = [Add(*bargs)], [], None
             if rv:
                 return rv
 
@@ -2144,4 +2142,4 @@ def do(e):
 
 from .numbers import Rational
 from .power import Pow
-from .add import Add, _addsort, _unevaluated_Add
+from .add import Add, _unevaluated_Add
diff --git a/sympy/core/tests/test_arit.py b/sympy/core/tests/test_arit.py
index 16580c14fa..8a35d99a49 100644
--- a/sympy/core/tests/test_arit.py
+++ b/sympy/core/tests/test_arit.py
@@ -2354,3 +2354,24 @@ def test_issue_17130():
 def test_issue_21034():
     e = -I*log((re(asin(5)) + I*im(asin(5)))/sqrt(re(asin(5))**2 + im(asin(5))**2))/pi
     assert e.round(2)
+
+
+def test_issue_22021():
+    from sympy.calculus.util import AccumBounds
+    from sympy.tensor.tensor import TensExpr
+    from sympy.utilities.iterables import permutations
+    # these objects are special cases in Mul
+    from sympy.tensor.tensor import TensorIndexType, tensor_indices, TensorSymmetry,    get_symmetric_group_sgs, TensorIndex, tensor_mul, TensAdd,tensor_heads
+    L = TensorIndexType("L")
+    i = tensor_indices("i", L)
+    A, B = tensor_heads("A B", [L])
+    e = A(i) + B(i)
+    assert -e == -1*e
+    e = zoo + x
+    assert -e == -1*e
+    a = AccumBounds(1, 2)
+    e = a + x
+    assert -e == -1*e
+    for args in permutations((zoo, a, x)):
+        e = Add(*args, evaluate=False)
+        assert -e == -1*e
```

sympy/solvers/solveset.py
oscarbenjamin: I think it would be better if `invert_real` only returned sets that are strictly contained in the domain of a function. There are other ways this can go wrong like if if the `subset` test is indeterminate then a zoo case is not removed.

smichr: > I think it would be better if `invert_real` only returned sets that are strictly contained in the domain of a function

Since it has the `domain` argument, I agree. I think I fixed a problem where the domain was integers but it wasn't return only integers. Those are troublesome bugs because you assume that the routine can be counted on to respect the domain and that you won't have worry about this in the the calling routine.

smichr:
Hmmm. The signature of `invert_real` should be like `solveset_real` and not accept the domain argument per its own docstring

```python
def invert_real(f_x, y, x):
    """
    Inverts a real-valued function. Same as _invert, but sets
    the domain to ``S.Reals`` before inverting.
    """
    return _invert(f_x, y, x, S.Reals)
```
And then the calling program intersects the real result with the actual domain.

Or else it should be something like this:

```python
def invert_real(f_x, y, x, domain=S.Reals):
    """
    Inverts a real-valued function. Same as _invert, but sets
    the domain to ``S.Reals`` before inverting.
    """
    return _invert(f_x, y, x, domain.intersection(S.Reals))
```

smichr:
```diff
diff --git a/sympy/solvers/solveset.py b/sympy/solvers/solveset.py
index acb0a81052..434520fd79 100644
--- a/sympy/solvers/solveset.py
+++ b/sympy/solvers/solveset.py
@@ -188,10 +188,10 @@ def _invert(f_x, y, x, domain=S.Complexes):
 
 def invert_real(f_x, y, x, domain=S.Reals):
     """
-    Inverts a real-valued function. Same as _invert, but sets
-    the domain to ``S.Reals`` before inverting.
+    Inverts a real-valued function. Same as _invert, but
+    limits the domain to ``S.Reals`` before inverting.
     """
-    return _invert(f_x, y, x, domain)
+    return _invert(f_x, y, x, domain.intersection(S.Reals))
 
 
 def _invert_real(f, g_ys, symbol):
diff --git a/sympy/solvers/tests/test_solveset.py b/sympy/solvers/tests/test_solveset.py
index fbbc6cddae..3ff2094bf0 100644
--- a/sympy/solvers/tests/test_solveset.py
+++ b/sympy/solvers/tests/test_solveset.py
@@ -65,6 +65,7 @@ def ireal(x, s=S.Reals):
     assert invert_real(x, 0, x, Interval(1, 2)) == (x, S.EmptySet)
 
     assert invert_real(exp(x), z, x) == (x, ireal(FiniteSet(log(z))))
+    assert invert_real(exp(x), z, x, S.Complexes) == (x, ireal(FiniteSet(log(z))))
 
     y = Symbol('y', positive=True)
     n = Symbol('n', real=True)
```
passes all solveset tests

sympy/solvers/solveset.py
oscargus: I agree in general. The only other way I might think of is 
```
g_ys = Intersection(g_ys, S.Reals)
```
which will also cover other cases (however, I believe that all the inverse functions return real-valued results for real-valued input, except for those that returns zoo, but I may be wrong). I'm a bit worried that this could explode in certain cases. (`Intersection(ImageSet(..., Intersection(ImageSet(..., Intersection(...)), S.Reals), S.Reals)`) Also, I am not sure how computationally efficient this is, but can give it a try.

oscargus: Will also be interesting see if that deals with: `solveset(I*pi-log(x), x, S.Reals)`, which, if I guess correctly, will have an imaginary intermediate value.

smichr: I tried to fix this at the intersection-level in #22045.

sympy/solvers/solveset.py
```diff
@@ -197,6 +197,11 @@ def invert_real(f_x, y, x, domain=S.Reals):
...
+
```
smichr, line 204:
```suggestion
```

sympy/solvers/solveset.py
```diff
@@ -1116,7 +1118,16 @@ def _solveset(f, symbol, domain, _check=False):
         if isinstance(result, ConditionSet):
             # it wasn't solved or has enumerated all conditions
             # -- leave it alone
-            return result
+            if domain is S.Complexes:
+                return result
+            # Add domain
+            new_res = Intersection(result, domain)
+            # Check if the domain is really contributing
+            if new_res.is_subset(result):
```
oscarbenjamin, line 1126: Isn't this always True? (Or rather is it not safe to assume that it is always True?)

oscargus: It may be that you are correct.

The thing was that just adding the domain sometimes led to some bad looking results. Typically, the domain was not really required in some cases and still remained. I do not remember an exact example, but, say, `Intersection(ConditionSet(... S.Reals), S.Reals)` or something similar. This was one of many attempts to get rid of those that at least didn't add anything "stupid".

Maybe it would be better to just add the domain unless it is `S.Complexes` and use that as input for improving `Intersection`?

oscargus: Ahh, forgot what I had already written. I'll try to find a better way of doing this.

oscargus: This was the problem:
`solveset(Abs(x) - n, x, S.Reals)` currently returns `ConditionSet(x, Contains(n, Interval(0, oo)), {-n, n})`, by Intersecting with the Reals domain we get `ConditionSet(x, Contains(_n, Interval(0, oo)), Intersection({-_n, _n}, Reals))`.

oscargus: I realize that I was trying to solve issues that I didn't fully understood. So I'll just revert this part.

PR #21928, opened 2021-08-22
Fixed issue with determining is_integer for round functions (#21651)

### LINKED ISSUE #21651: doit() method *sometimes* ignores floor and ceiling within Sum
**Summary of problem**: the `doit()` method will sometimes (depending on the expression) ignore any `floor` or `ceiling` that is located inside a `Sum`.

**Example detailing the bug**

The following code sets up the problem:

```
from sympy import floor, Sum, Symbol
i = Symbol('i')
a = Sum(floor(2*2**(-i)), (i, 1, 2))
b = floor(2*2**(-1)) + floor(2*2**(-2))
```

Note that these two expressions `a` and `b` are the same, just `a` is written with the summation notation and `b` is written out explicitly. They both equal `1`. 

Printing these variables give us the expected results:

image

However, if we run the `doit()` method on them we see a discrepancy:

image

where we see that the `floor` function inside the summation has been ignored. The same issue also arises when using the `ceiling` method (i.e. the `doit()` method ignores it when it is in a summation).

However it seems like this problem is not consistent, e.g. if you remove the `2*` from the expression it works fine. More precisely if you run

```
from sympy import floor, Sum, Symbol
i = Symbol('i')
a = Sum(floor(2**(-i)), (i, 1, 2))
b = floor(2**(-1)) + floor(2**(-2))
print(a.doit())
print(b.doit())
```
you get the exptected results (this works for `ceiling` too)

(bug discovered on sympy version 1.8 on python 3.7.3)

oscargus:

Closes #21651

For some odd cases, `Mul.is_integer` returned `True` where it should have been `None`.

#### Other comments
There are three tests added:
1. To check that the final result in #21651 is correct.
2. To check that the floor is not removed from the expression.
3. To check that `is_integer` returns correctly.

#### Release Notes

* core
   * Fixed issue where `is_integer` returned incorrectly for multiplications.

### DIFF
--- a/sympy/core/mul.py
+++ b/sympy/core/mul.py
@@ -1351,7 +1351,7 @@ class Mul(Expr, AssocOp):
                 b, e = a.as_base_exp()
                 if not b.is_integer or not e.is_integer: return
                 if e.is_negative:
-                    denominators.append(Pow(b, S.NegativeOne*e))
+                    denominators.append(2 if a is S.Half else Pow(a, S.NegativeOne, evaluate=False))
                 else:
                     # for integer b and positive integer e: a = b**e would be integer
                     assert not e.is_positive
--- a/sympy/core/mul.py
+++ b/sympy/core/mul.py
@@ -1351,7 +1351,7 @@ class Mul(Expr, AssocOp):
                 b, e = a.as_base_exp()
                 if not b.is_integer or not e.is_integer: return
                 if e.is_negative:
-                    denominators.append(b)
+                    denominators.append(Pow(b, S.NegativeOne*e))
                 else:
                     # for integer b and positive integer e: a = b**e would be integer
                     assert not e.is_positive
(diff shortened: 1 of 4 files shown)


sympy/core/mul.py
```diff
@@ -1351,7 +1351,7 @@ def _eval_is_integer(self):
                 b, e = a.as_base_exp()
                 if not b.is_integer or not e.is_integer: return
                 if e.is_negative:
-                    denominators.append(b)
+                    denominators.append(Pow(b, S.NegativeOne*e))
```
smichr, line 1354: You can probably safely make this `2 if a is S.Half else Pow(a, -1, evaluate=False)`.

oscargus: Fixed that and it seems to work for the test cases at least.

oscargus: Nah, the tests didn't pass... `is_odd` and `is_even` returns `None` for unevaluated results.

`Pow(a, S.NegativeOne)` is about seven times faster compared to my previous approach though.

smichr:
Thanks, this is in.

PR #21929, opened 2021-08-22
solvers - fixes issue #19814 and nonlinsolve() for some equations returns correct output

### LINKED ISSUE #19814: nonlinsolve throwing AttributeError
```python3
>>> from sympy import symbols, nonlinsolve
>>> n,m = symbols('n m')
>>> nonlinsolve([ 2**m - 2**(2*n),  4*2**m - 2**(4*n) ], m,  n)
```
This code is throwing the following error: `AttributeError: 'ConditionSet' object has no attribute 'as_coeff_Mul'`
Shouldn't the expected output be m=2 and n=1?

praneethratna:
Fixes #19814 and now `nonlinsolve([ 2**m - 2**(2*n),  4*2**m - 2**(4*n) ], m,  n)` returns `FiniteSet((log(2**(2*n))/log(2), S.Complexes))` instead of `AttributeError: 'ConditionSet' object has no attribute 'as_coeff_Mul'` and also added a regression test in test_solveset.py

#### Release Notes

* solvers
   * fixed a bug in nonlinsolve() method in solveset.

### DIFF
--- a/sympy/solvers/solveset.py
+++ b/sympy/solvers/solveset.py
@@ -3303,7 +3303,7 @@ def substitution(system, symbols, result=[{}], known_symbols=[],
                         # in the new result list (solution for symbol `s`)
                         # along with old results.
                         for k, v in res.items():
-                            if isinstance(v, Expr) and not sol is S.Complexes:
+                            if isinstance(v, Expr) and isinstance(sol, Expr):
                                 # if any unsolved symbol is present
                                 # Then subs known value
                                 rnew[k] = v.subs(sym, sol)
--- a/sympy/solvers/solveset.py
+++ b/sympy/solvers/solveset.py
@@ -3303,7 +3303,7 @@ def substitution(system, symbols, result=[{}], known_symbols=[],
                         # in the new result list (solution for symbol `s`)
                         # along with old results.
                         for k, v in res.items():
-                            if isinstance(v, Expr):
+                            if isinstance(v, Expr) and not sol is S.Complexes:
                                 # if any unsolved symbol is present
                                 # Then subs known value
                                 rnew[k] = v.subs(sym, sol)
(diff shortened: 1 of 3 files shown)


sympy/solvers/solveset.py
```diff
@@ -3303,7 +3303,7 @@ def _solve_using_known_values(result, solver):
                         # in the new result list (solution for symbol `s`)
                         # along with old results.
                         for k, v in res.items():
-                            if isinstance(v, Expr):
+                            if isinstance(v, Expr) and not sol is S.Complexes:
```
oscarbenjamin, line 3306: Could this be
```
if isinstance(v, Expr) and isinstance(sol, Expr)
```
?

oscarbenjamin:
I don't know `nonlinsolve` so well so I'm hoping someone else will review this.

oscargus:
I think that it is probably a step in the right direction, but I do not think that the solution is correct.
`log(2**(2*n))/log(2)` seems to come from solving the first equation for `m`. Then, I guess, the idea is to replace `m` with `log(2**(2*n))/log(2)` and solve the second equation. That equation is now `-2**(4*n) + 4*2**(2*n)` and solving that with `nonlinsolve` gives `FiniteSet((Union(ConditionSet(n, Eq(2**n - 2, 0), Complexes), ConditionSet(n, Eq(2**n + 2, 0), Complexes)),))` so it cannot be fully solved (`solve` works though).

So while it is nice that there isn't any exception, the solution seems to wrong. Not sure if this is caused by the change here, but it may be worth checking if the current change somehow affects the recombination of solutions as well.

oscargus:
The solution to the second equation is `[1, 1 + I*pi/log(2)]`, so ideally the solution should be something like FiniteSet({2, 1}, {2, 1 + I*pi/log(2)}) if I'm doing the maths correctly. (It may be that there are more solutions as well.)

smichr:
> the solution seems to wrong

It is ok, just deficient in not knowing how to solve `2**n +/-2`. It is producing the correct conditions for the solution that exists. It is just not solving those conditions.

PR #21933, opened 2021-08-23
piecewise rewrite sign and factor

robinechuca:
an instance of Piecewise can sometimes be rewritten with the sign function.

When trying to simplify a piecewise expression, it now tries to factor and express itself as a sign function when possible.

* functions
  * When applying factor to piecewise, it attempts to output a common factor to all terms.
  * If the Piecewise expression is mathematically equivalent to "x1*sign(x2)", then it is now possible to perform the transformation.


oscargus:
Please do a rebase and squash the commits (`git rebase -i master` and squash or fixup the deletion commits) and then do `git push --force`. If not, all these extra files will still end up in the repository once merged.

For the PR itself, I like the idea!

sympy/functions/elementary/piecewise.py
```diff
@@ -806,6 +807,25 @@ def _eval_nseries(self, x, n, logx, cdir=0):
...
+        from sympy.polys.polyerrors import PolynomialError
+
+        args = [(e if hasattr(e, "_eval_factor") else factor(e), c) for e, c in self.args]
+        commonfact = None
+        for expr, cond in args:
+            if commonfact is not None:
+                try:
+                    commonfact = gcd(commonfact, expr)
+                except PolynomialError:
+                    commonfact = 1
```
oscargus, line 821: Better to use `S.One` here.

sympy/functions/elementary/piecewise.py
```diff
@@ -806,6 +807,25 @@ def _eval_nseries(self, x, n, logx, cdir=0):
...
+            else:
+                commonfact = expr
+        if commonfact == 1 and args == self.args:
```
oscargus, line 824: Why do you have `args == self.args` here?

What if some of the arguments are modified by `factor` and `commonfact` is still 1? In that case you will divide by one and factor out a one which doesn't make sense (or am I missing something?).

sympy/functions/elementary/piecewise.py
```diff
@@ -806,6 +807,25 @@ def _eval_nseries(self, x, n, logx, cdir=0):
...
+            return self
+        args = [(cancel(e/commonfact), c) for e, c in args]
+        return commonfact*Piecewise(*args)
```
oscargus, line 827: I think that you can use `evaluate=False` for `Piecewise` here to speed things up as there should be no need to reevaluate the arguments. (Unless there is an instance of the same expression which cannot be identified upon the first creation...)

sympy/functions/elementary/piecewise.py
```diff
@@ -1045,6 +1065,79 @@ def rewrite(cond):
...
+                if len(piecewises) != 1:
+                    return
+                from sympy.core.mul import Mul
+                args = piecewises.pop().args
+                fact = Mul(*[f for f in fact_args if not f.is_Piecewise])
+        args = list(args)
+
+        # the comparison elements are reported to 0
+        for i, (expr, cond) in list(enumerate(args)):
+            if isinstance(cond, (GreaterThan, LessThan,
```
oscargus, line 1095: `Unequality`?

sympy/functions/elementary/piecewise.py
robinechuca: Yes, I have nothing against it, but how is it better or worse?

sympy/functions/elementary/piecewise.py
robinechuca: You're right, dividing and factoring by 1 is not very efficient. There is only one case where you want factor to factor Piecewise arguments, and that is when one of those arguments is a gamma expression, because "gammasimp" counts on that.

sympy/functions/elementary/piecewise.py
oscargus: 1 is a plain Python integer. S.One is a SymPy integer. So 1 will have to be converted to a SymPy integer when used. Here on line 827, so two times per argument. One will probably not notice the speed difference (but there is one), more like good practice.

sympy/functions/elementary/piecewise.py
oscargus: Just to clarify (maybe I just do not understand the code...), but Unequality is ~=, which wondered if it should be there.

If you want to test for all (==, <, <=, >, >=, ~=) you can use `Relational`.

sympy/functions/elementary/piecewise.py
robinechuca: Okay, I replace it. I imagine is the same think for the sets in _eval_rewrite_as_sign ?

sympy/functions/elementary/piecewise.py
robinechuca: The aim here is just to reduce it to a comparison with 0 in order to facilitate the change of variable. Maybe there are even other classes besides Relational? I'm not very familiar with sympy. I'm interested in all these little tips! thank you ;)

sympy/functions/elementary/tests/test_piecewise.py
```diff
@@ -1188,6 +1198,30 @@ def _ITE(*args):
...
+    approx_args = [p_arg for arg in approx_args for p_arg in permutations(arg)]
+    strict_args += [arg[:-1]+((arg[-1][0], True),) for arg in strict_args]
+    approx_args += [arg[:-1]+((arg[-1][0], True),) for arg in approx_args]
+
+    for args in strict_args:
+        assert Piecewise(*args).rewrite(sign) == sign(x)
+        assert Piecewise(*mult(-1, args)).rewrite(sign) == -sign(x)
+        assert Piecewise(*mult(2, args)).rewrite(sign) == 2*sign(x)
+        assert Piecewise(*mult(x, args)).rewrite(sign) == x*sign(x)
+        assert Piecewise(*mult(-x, args)).rewrite(sign) == -x*sign(x)
```
oscarbenjamin, line 1216: I find this way of writing tests quite hard to read because the test input is far away from the expected output. I think it reads better if you do something more like:
```python
def test_perm(expected, *args):
    for arg in permutations(args):
        assert Piecewise(*arg) == expected

test_perm(sign(x), (1, x > 0), (0, Eq(x, 0)), (-1, x < 0))
test_perm(-sign(x), (1, x >= 0), (-1, x < 0)), ((1, x > 0), (-1, x <= 0))
...
```

robinechuca: Okay! It is do.

sympy/functions/elementary/tests/test_piecewise.py
```diff
@@ -1188,6 +1198,36 @@ def _ITE(*args):
...
+    strict_args = [((1, x > 0), (0, Eq(x, 0)), (-1, x < 0))]
+    approx_args = [((1, x >= 0), (-1, x < 0)),
+                   ((1, x > 0), (-1, x <= 0))]
+
+    for args in strict_args:
+        test_perm(args          ,  sign(x))
+        test_perm(mult(-1, args), -sign(x))
+        test_perm(mult( 2, args),  2*sign(x))
+        test_perm(mult( x, args),  x*sign(x))
+        test_perm(mult(-x, args), -x*sign(x))
```
oscarbenjamin, line 1222: This is still quite hard to decipher. For one thing `strict_args` is a list of length one so we don't need a loop.

Also I think that:
```
test_perm([(1, x > 0), (0, Eq(x, 0)), (-1, x < 0)], sign(x))
```
is the most readable way to write this.

robinechuca: I put it in a single item list because I imagined he would have more later.

robinechuca:
I don't understand, at the moment the test benches are running locally on my computer and it is also running on my github repository. I can't reproduce the error. I really don't see where it comes from...

robinechuca:
I'm not super strong in math. There are 2 points that question me:
- If we work with complexes, does the simplification of the Piecewise by the "sign" function make sense? In my opinion no, but the problem comes from the fact that we can write "x > 0, x€C" . So the question is: before trying to rewrite with the "sign" function, should we check that we are working with Real numbers?
- The other point concerns the value in 0, "sign(0) = 0". In this pull request, I make sure that this point is respected. But for that, we make several calculations of limits and change of variables, which seems to me perhaps excessive and heavy. I did this to ensure that the results are mathematically correct. But maybe there is a more elegant way to handle this point?

robinechuca:
From your example, I have the impression that sign(a.exp(i.t)) = exp(i.t). And in the complex, we have a.exp(i.t) / abs(a.exp(i.t)) = exp(i.t).
I have the impression that we can maybe generalize this pool request with complex numbers. I don't see how for the moment.

robinechuca:
I don't think that the rewriting of exp(x) should be done directly in Piecewice. Maybe it should be added to the Exponential class.
The current version seems complete to me.

Issue #13430, opened 2017-10-10
move Sage conversion into Sage


rwst:
After the Sage ticket
https://trac.sagemath.org/ticket/24006

 - the `_sage_` member functions
 - the file `sympy/external/tests/test_sage.py`
 - Travis tests and scripts with `TEST_SAGE`

can be removed.

Motivation:
The real problems of #13395 raise the question if the conversion methods named `_sage_()` shouldn't better be part of Sage? The methods can be added dynamically there. 

Not only is Travis testing a PITA, I have just now several interdependent Sage tickets that add `_sage_` functions, and I need not only patches against SymPy master but also 1.0. And a SymPy patch and a Sage ticket for each.

rwst:
It looks like @certik started the Sage interface in https://github.com/sympy/sympy/commit/98e996ca4b89c6cebcec60f5574621cd8342006e. We certainly would like to hear from him on this.

OTOH the move could be done completely independently in Sage, and I will try to experiment this for principle.

certik:
@rwst I am fine either way. The only disadvantage of adding it in Sage is that if something has to be modified in those `_sage_` methods that is tied to a sympy version, then you have to maintain specific patches to specific SymPy versions. But I guess Sage will use just one version of sympy at a time, so I guess that's fine. How do other Python libraries handle this?

rwst:
Yes, Sage always has one specific SymPy version. I'll now change the issue description.

certik:
I agree that if we don't have to test Sage, it will become simpler. I am just a bit worried that it might happen to be a lot of work to keep things working if things are not tested in SymPy itself. But perhaps it will not be too much work to update the _sage_ methods for each new SymPy release to be put into Sage.

rwst:
I'm afraid the Sage ticket will have to use `_sage_()` instead of `sage()` because all other interfaces in Sage use that too. So it will overwrite what is in SymPy.

asmeurer:
Can you guys test SymPy master in your testing infrastructure so that you aren't surprised by a bunch of accidental breaks every time we do a SymPy release? That's the main disadvantage of moving everything to Sage, is that we can break stuff in SymPy and not really know about it.

rwst:
[...]

PR #21941, opened 2021-08-25
Add portability/integration testing with SageMath on GH Actions

mkoeppe:

#13430

Add a GH Actions workflow that builds an sdist and then launches SageMath's portability/integration testing.

#### Release Notes

NO ENTRY

### DIFF
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -1,0 +1,173 @@
+name: Run Sage CI for Linux/Cygwin/macOS
+
+## This GitHub Actions workflow provides:
+##
+##  - portability testing, by building and testing this project on many platforms
+##
+##  - continuous integration, by building and testing other software
+##    that depends on this project.
+##
+## The testing can be monitored in the "Actions" tab of the GitHub repository.
+##
+## After all jobs have finished (or are canceled) and a short delay,
+## tar files of all logs are made available as "build artifacts".
+##
+## This GitHub Actions workflow uses the portability testing framework
+## of SageMath (https://www.sagemath.org/).  For more information, see
+## https://doc.sagemath.org/html/en/developer/portability_testing.html
+
+## The workflow consists of two jobs:
+##
+##  - First, it builds a source distribution of the project
+##    and generates a script "update-pkgs.sh".  It uploads them
+##    as a build artifact named upstream.
+##
+##  - Second, it checks out a copy of the SageMath source tree.
+##    It downloads the upstream artifact and replaces the project's
+##    package in the SageMath distribution by the newly packaged one
+##    from the upstream artifact, by running the script "update-pkgs.sh".
+##    Then it builds a small portion of the Sage distribution.
+##
+## Many copies of the second step are run in parallel for each of the tested
+## systems/configurations.
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+  push:
+    tags:
+      - '*'
+  workflow_dispatch:
+    # Allow to run manually
+
+env:
+  # Ubuntu packages to install so that the project's "make dist" can succeed
+  DIST_PREREQ:      python3-setuptools
+  # Name of this project in the Sage distribution
+  SPKG:             sympy
+  # Sage distribution packages to build
+  TARGETS_PRE:      build/make/Makefile
+  TARGETS:          SAGE_CHECK=no SAGE_CHECK_PACKAGES=sympy build ptest
+  TARGETS_OPTIONAL: build/make/Makefile
+  # Standard setting: Test the current beta release of Sage:
+  SAGE_REPO:        sagemath/sage
+  SAGE_REF:         develop
+  # Uncomment to temporarily test with the branch from a sage ticket
+  # (this is a no-op after that ticket is merged)
+  ## SAGE_TRAC_GIT:    https://github.com/sagemath/sagetrac-mirror.git
+  ## SAGE_TICKET:
+  ## REMOVE_PATCHES:   "*"
+
+jobs:
+
+  dist:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out ${{ env.SPKG }}
+        uses: actions/checkout@v2
+        with:
+          path: build/pkgs/${{ env.SPKG }}/src
+      - name: Install prerequisites
+        run: |
+          sudo DEBIAN_FRONTEND=noninteractive apt-get update
+          sudo DEBIAN_FRONTEND=noninteractive apt-get install $DIST_PREREQ
+      - name: Run make dist, prepare upstream artifact
+        run: |
+          (cd build/pkgs/${{ env.SPKG }}/src && python3 setup.py sdist ) \
+          && mkdir -p upstream && cp build/pkgs/${{ env.SPKG }}/src/dist/*.tar.gz upstream/${{ env.SPKG }}-git.tar.gz \
+          && echo "sage-package create ${{ env.SPKG }} --version git --tarball ${{ env.SPKG }}-git.tar.gz --type=standard" > upstream/update-pkgs.sh \
+          && if [ -n "${{ env.REMOVE_PATCHES }}" ]; then echo "(cd ../build/pkgs/${{ env.SPKG }}/patches && rm -f ${{ env.REMOVE_PATCHES }}; :)" >> upstream/update-pkgs.sh; fi \
+          && ls -l upstream/
+      - uses: actions/upload-artifact@v2
+        with:
+          path: upstream
+          name: upstream
+
+  docker:
+    runs-on: ubuntu-latest
+    needs: [dist]
+    strategy:
+      fail-fast: false
+      max-parallel: 32
+      matrix:
+        # debian-buster uses system python 3.7
+        # ubuntu-groovy uses system python 3.8
+        # fedora-34 uses system python 3.9
+        # archlinux-latest is at the cutting edge
+        tox_system_factor: [debian-buster, ubuntu-groovy, fedora-34, archlinux-latest]
+        # "standard" installs lots of system packages, reducing the full build and test of the
+        # Sage distribution to 3-4 hours
+        tox_packages_factor: [standard]
+    env:
+      TOX_ENV: docker-${{ matrix.tox_system_factor }}-${{ matrix.tox_packages_factor }}
+      LOGS_ARTIFACT_NAME: logs-commit-${{ github.sha }}-tox-docker-${{ matrix.tox_system_factor }}-${{ matrix.tox_packages_factor }}
+      DOCKER_TARGETS: configured with-targets with-targets-optional
+    steps:
+      - name: Check out SageMath
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ env.SAGE_REPO }}
+          ref: ${{ env.SAGE_REF }}
+          fetch-depth: 2000
+        if: env.SAGE_REPO != ''
+      - name: Check out git-trac-command
+        uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-sage.yml
```diff
@@ -0,0 +1,173 @@
...
+    needs: [dist]
+    strategy:
+      fail-fast: false
+      max-parallel: 32
+      matrix:
+        # debian-buster uses system python 3.7
+        # ubuntu-groovy uses system python 3.8
+        # fedora-34 uses system python 3.9
+        # archlinux-latest is at the cutting edge
+        tox_system_factor: [debian-buster, ubuntu-groovy, fedora-34, archlinux-latest]
```
asmeurer, line 98: This seems like a fragile way to test various versions of Python. Whenever a new version of Python is released we have to find a new linux distro that includes it. Is it not possible to just build a docker container specifically for Sage with all the Python versions required? Or alternately use conda environments, since all the Sage dependencies are on conda-forge (assuming they work with the development version).

.github/workflows/ci-sage.yml
```diff
@@ -0,0 +1,173 @@
+name: Run Sage CI for Linux/Cygwin/macOS
```
asmeurer, line 1: I think this is just Linux.

.github/workflows/ci-sage.yml
asmeurer: It's also opaque in the test matrix. We should at least name the matrix entries appropriately, assuming that's possible.

asmeurer: Actually, is it really necessary for us to test every version of Python? Do we anticipate problems that only occur in one version?

mkoeppe: The comments there just explain my selection among the dozens of platforms that we test in Sage.

I am not suggesting that you should replace your testing of supported Python versions by my CI.

.github/workflows/ci-sage.yml
mkoeppe: True, I cut out the other platforms to make the file simpler. Given that sympy is pure Python, there's probably relatively little point in the full portability testing that we do.

.github/workflows/ci-sage.yml
mkoeppe: Well, this proposed workflow does integration testing with Sage (= the Sage distribution), which is what is most relevant for Sage development.

Using conda-forge would certainly be another way for sympy to test integration with some version of sage; but you won't get testing with the Sage development releases that way.

mkoeppe: > Actually, is it really necessary for us to test every version of Python? Do we anticipate problems that only occur in one version?

There are many differences between these 4 platforms, not just the Python version. Sage makes use of system packages that its `configure` scripts determines to be suitable. For example, it picks up openblas from the system; so if sympy contains numerical tests using numpy, the workflow will test it with that on these platforms. Whether this is useful for you, you will have to decide. It's fine with me if you want to test fewer platforms than the ones I put there.

mkoeppe:
The test on `archlinux-latest` shows some failures apparently caused by some differences of SymPy master and release 1.8. 
The test on `ubuntu-groovy` showed the same failures but there were additional numerics-related failures that are likely unrelated to SymPy. I have removed this platform from the script.

oscarbenjamin:
Thanks for this.

oscarbenjamin:
The failures are:
```
File "src/sage/matrix/matrix1.pyx", line 542, in sage.matrix.matrix1.Matrix._sympy_
Failed example:
    sA = A._sympy_(); sA
Expected:
    Matrix([
    [1, 2, 3],
    [4, 5, 6],
    [7, 8, 9]])
Got:
    Matrix([
    [1],
    [2],
    [3],
    [4],
    [5],
    [6],
    [7],
    [8],
    [9]])

**********************************************************************
File "src/sage/matrix/matrix1.pyx", line 572, in sage.matrix.matrix1.Matrix._sympy_
Failed example:
    sA._sage_() == A
Expected:
    True
Got:
    False

**********************************************************************
File "src/sage/matrix/matrix1.pyx", line 580, in sage.matrix.matrix1.Matrix._sympy_
Failed example:
    sM = M._sympy_(); sM
Expected:
    Matrix([
    [ sin(x), cos(x)],
    [-cos(x), sin(x)]])
Got:
    Matrix([
    [ sin(x)],
    [ cos(x)],
    [-cos(x)],
    [ sin(x)]])

**********************************************************************
File "src/sage/matrix/matrix1.pyx", line 584, in sage.matrix.matrix1.Matrix._sympy_
Failed example:
    sM.subs(x, pi/4)
Expected:
    Matrix([
    [ sqrt(2)/2, sqrt(2)/2],
    [-sqrt(2)/2, sqrt(2)/2]])
Got:
    Matrix([
    [ sqrt(2)/2],
    [ sqrt(2)/2],
    [-sqrt(2)/2],
    [ sqrt(2)/2]])
```
Not sure if that's a bug on Sage's side but this might be something important to fix for the upcoming sympy release... Do you know how the sympy Matrix is being constructed here?

These are changes in printing:
```
**********************************************************************
File "src/sage/tests/books/computational-mathematics-with-sagemath/recequadiff_doctest.py", line 377, in sage.tests.books.computational-mathematics-with-sagemath.recequadiff_doctest
Failed example:
    rsolve(f, u(n), {u(0):-1,u(1):1})
Expected:
    3 - 4*2**(-n)
Got:
    3 - 4/2**n

**********************************************************************
File "src/sage/calculus/test_sympy.py", line 196, in sage.calculus.test_sympy
Failed example:
    rsolve(f,u(n))
Expected:
    2**(-n)*(C0 + C1*n)
Got:
    (C0 + C1*n)/2**n
```
Maybe those doctests just need to be updated in Sage?

The Heaviside function has changed how it handles `H(0)` (#21469):
```
**********************************************************************
File "src/sage/calculus/calculus.py", line 1628, in sage.calculus.calculus.laplace
Failed example:
    laplace(dirac_delta(t), t, s, algorithm='sympy')
Expected:
    (-heaviside(0) + 1, -oo, True)
Got:
    (1, -oo, True)
```
The new output is better. Maybe doctest needs updating in Sage.

There is a change in how `Heaviside(x)` prints:
```
**********************************************************************
File "src/sage/functions/generalized.py", line 206, in sage.functions.generalized.FunctionHeaviside
Failed example:
    heaviside(x)._sympy_()
Expected:
    Heaviside(x)
Got:
    Heaviside(x, 1/2)
```
That could be changed to suppress the 1/2 since it is default and isn't shown in sympy 1.8:
```
In [1]: Heaviside(x)
Out[1]: θ(x)

In [2]: str(Heaviside(x))
Out[2]: 'Heaviside(x, 1/2)'
```

mkoeppe:
> The failures are:
> 
> ```
> File "src/sage/matrix/matrix1.pyx", line 542, in sage.matrix.matrix1.Matrix._sympy_
> Failed example:
>     sA = A._sympy_(); sA
> Expected:
>     Matrix([
>     [1, 2, 3],
>     [4, 5, 6],
>     [7, 8, 9]])
> Got:
>     Matrix([
>     [1],
>     [2],
>     [3],
>     [4],
>     [5],
>     [6],
>     [7],
>     [8],
>     [9]])
> 
> Not sure if that's a bug on Sage's side but this might be something important to fix for the upcoming sympy release... Do you know how the sympy Matrix is being constructed here?

These doctests test the new matrix conversion code that was added in Sage 9.4 in https://trac.sagemath.org/ticket/31942; see https://git.sagemath.org/sage.git/diff/src/sage/matrix/matrix1.pyx?id=3c9a3f547c17acd9aeac8044f577dbb287f449e5

Has the `ImmutableMatrix` constructor changed since 1.8?


> 
> These are changes in printing:
> 
> ```
> **********************************************************************
> File "src/sage/tests/books/computational-mathematics-with-sagemath/recequadiff_doctest.py", line 377, in sage.tests.books.computational-mathematics-with-sagemath.recequadiff_doctest
> Failed example:
>     rsolve(f, u(n), {u(0):-1,u(1):1})
> Expected:
>     3 - 4*2**(-n)
> Got:
>     3 - 4/2**n
> 
> Maybe those doctests just need to be updated in Sage?

Yes, this is something to update on the Sage side. Currently, according to `build/pkgs/sympy/install-requires.txt`, we claim to support `sympy >=1.6, <2.0`, so in Sage we'll have to see whether we can update these doctests so that they work with older and newer versions.

> 
> The Heaviside function has changed how it handles `H(0)` (#21469):
> 
> ```
> **********************************************************************
> File "src/sage/calculus/calculus.py", line 1628, in sage.calculus.calculus.laplace
> Failed example:
>     laplace(dirac_delta(t), t, s, algorithm='sympy')
> Expected:
>     (-heaviside(0) + 1, -oo, True)
> Got:
>     (1, -oo, True)
> ```
> 
> The new output is better. Maybe doctest needs updating in Sage.
> 
> There is a change in how `Heaviside(x)` prints:
> 
> ```
> **********************************************************************
> File "src/sage/functions/generalized.py", line 206, in sage.functions.generalized.FunctionHeaviside
> Failed example:
>     heaviside(x)._sympy_()
> Expected:
>     Heaviside(x)
> Got:
>     Heaviside(x, 1/2)
> ```
> 
> That could be changed to suppress the 1/2 since it is default and isn't shown in sympy 1.8:
> 
> ```
> In [1]: Heaviside(x)
> Out[1]: θ(x)
> 
> In [2]: str(Heaviside(x))
> Out[2]: 'Heaviside(x, 1/2)'
> ```

Thanks. I have opened https://trac.sagemath.org/ticket/32420 for the necessary work on the Sage side.

oscarbenjamin:
> Now that the Sage tests have completed I can see that the change I tried fixed the matrix problems there at least. More investigation is needed to figure out exactly what has changed and how to fix it without breaking the SymPy tests that are currently failing in this PR.

According to git bisect the problem emerged from 8b63ecd86b87998faac06a663435c97ae162b3da but I don't see anything there that looks suspicious for this...

I think it was the change to this line:
https://github.com/sympy/sympy/blob/6c605115a7ed99601e222ce99e485e0889dda83f/sympy/matrices/matrices.py#L1010

Previously that was just `return x`. I'm not sure why it was changed in that commit. The change there means that we have:
```python
(Pdb) p x
(1, 2, 3)
(Pdb) p cls._sympify(x)
Matrix([
[1],
[2],
[3]])
```
So the 3-tuple (`Vector_integer_dense`) becomes a 3x1 column matrix when sympified. I think maybe if that was a row matrix then all would work nicely. I'm not sure where the code controlling that sympification is but I guess it's in sage.

The problem really though is that `x` is being treated as an element of the matrix rather than a row of elements and that is due to the `list/int` check.

oscarbenjamin:
> The docstring appears to promise that it works with arbitrary iterables, not just sequences

Which docstring are you referring to?

mkoeppe:
> > The docstring appears to promise that it works with arbitrary iterables, not just sequences
> 
> Which docstring are you referring to?

`MatrixBase._handle_creation_inputs`
https://github.com/sympy/sympy/blob/fe3325288047312fbba6c8229e37438a0e2477c6/sympy/matrices/matrices.py#L917

mkoeppe:
> due to the `list/tuple` check.

Related ancient issue: #5113

oscarbenjamin:
The matrix construction issue is fixed. The only failures here now are the doctests.

I opened  #21945 for the change to Heaviside printing which needs a decision before the upcoming SymPy release.

I think the other doctests just need to be updated or modified in sage.

I'm going to push a commit so that the sage tests only run after a merge to master and then merge this in.

mkoeppe:
Wonderful, thanks very much!
I'll work on the doctest changes needed on the Sage side.

PR #21944, opened 2021-08-25
enhanced int/odd/even detection

smichr:

fixes #8648
closes #12399 as alternate

Consideration of the nominal multiplicity of 2 allows us to return True/False for queries on integer/odd/even as tests indicate.

#### Release Notes

* core
    * better detection of odd/even/integer for expressions with symbolic powers of even bases

### DIFF
--- a/sympy/core/mul.py
+++ b/sympy/core/mul.py
@@ -1334,6 +1334,7 @@ class Mul(Expr, AssocOp):
     #_eval_is_integer = lambda self: _fuzzy_group(
     #    (a.is_integer for a in self.args), quick_exit=True)
     def _eval_is_integer(self):
+        from sympy import trailing
         is_rational = self._eval_is_rational()
         if is_rational is False:
             return False
@@ -1345,8 +1346,10 @@ class Mul(Expr, AssocOp):
                 numerators.append(a)
             elif a.is_Rational:
                 n, d = a.as_numer_denom()
-                numerators.append(n)
-                denominators.append(d)
+                if abs(n) is not S.One:
+                    numerators.append(n)
+                if d is not S.One:
+                    denominators.append(d)
             elif a.is_Pow:
                 b, e = a.as_base_exp()
                 if not b.is_integer or not e.is_integer: return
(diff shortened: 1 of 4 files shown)


sympy/core/mul.py
```diff
@@ -1545,37 +1564,48 @@ def _eval_is_extended_negative(self):
         return self._eval_pos_neg(-1)
 
     def _eval_is_odd(self):
-        is_integer = self.is_integer
```
oscargus, line 1548: Wouldn't it make sense to keep this to avoid potentially have to evaluate it twice?

sympy/core/mul.py
```diff
@@ -1545,37 +1564,48 @@ def _eval_is_extended_negative(self):
...
+                elif acc != 1 and (acc + t).is_odd:
                     r = False
-                elif t.is_integer:
-                    if r is False:
-                        pass
-                    elif acc != 1 and (acc + t).is_odd:
-                        r = False
-                    elif t.is_odd is None:
-                        r = None
+                elif t.is_odd is None:
```
oscargus, line 1586: Is there a way that `t.is_odd` is different from `fuzzy_not(t.is_even)`?

oscargus:
Looks good! Just two minor comments regarding avoiding redundant evaluations.

sympy/core/mul.py
smichr: If `e` is even then `(2/e).is_even` is False but `(2/e).is_odd` is None.

smichr: But in the context of this code, `t.is_integer` is True and `denom(t) == 1` so I don't think there would be a difference. Why?

oscargus: Since you tested for `t.is_even`, I was just thinking about if one could draw some conclusions based on that result and not have to evaluate `t.is_odd`, which may or may not be time consuming.

sympy/core/mul.py
oscargus: OK!

PR #21958, opened 2021-08-27
Remove obsolete _sage_ methods and old scripts for Sage integration testing

mkoeppe:

#13430

Remove almost all `_sage_` methods. In all supported versions of Sage, these methods are provided (via monkey patching) by Sage.

Also remove the old scripts for Sage integration testing; they are obsolete after #21941.

#### Release Notes

* core
   * The `_sage_` SymPy/Sage interface methods are no longer included with SymPy, but are now initialized on the first use of a `_sage_` method from within Sage.


mkoeppe:
One new failure:
```
File "src/sage/rings/infinity.py", line 570, in sage.rings.infinity.UnsignedInfinityRing_class.__init__
Failed example:
    SR(sympy.zoo)
Exception raised:
      File "/sage/local/lib/python3.9/site-packages/sage/doctest/forker.py", line 718, in _run
        self.compile_and_execute(example, compiler, test.globs)
      File "/sage/local/lib/python3.9/site-packages/sage/doctest/forker.py", line 1137, in compile_and_execute
        exec(compiled, globs)
      File "<doctest sage.rings.infinity.UnsignedInfinityRing_class.__init__[2]>", line 1, in <module>
        SR(sympy.zoo)
      File "sage/structure/parent.pyx", line 898, in sage.structure.parent.Parent.__call__ (build/cythonized/sage/structure/parent.c:9338)
        return mor._call_(x)
      File "sage/symbolic/ring.pyx", line 1417, in sage.symbolic.ring.UnderscoreSageMorphism._call_ (build/cythonized/sage/symbolic/ring.cpp:14822)
        return self.codomain()(a._sage_())
    AttributeError: 'ComplexInfinity' object has no attribute '_sage_'
```

mkoeppe:
This doctest in Sage tries to access the `_sage_` method right after importing `sympy`. This can't work because `sage.interfaces.sympy.sympy_init` must be called first.

sympy/matrices/matrices.py
```diff
@@ -1001,8 +1001,8 @@ def _handle_creation_inputs(cls, *args, **kwargs):
                 if evaluate:
                     def do(x):
                         # make Block and Symbol explicit
-                        if isinstance(x, (list, tuple)):
-                            return type(x)([do(i) for i in x])
+                        if is_sequence(x):
+                            return list([do(i) for i in x])
```
oscargus, line 1005: `return [do(i) for i in x]` should be enough.

(If this should be changed or not, I cannot really tell though...)

oscargus:
The test error is in (and should be fixed in) Sage right? I really have limited experience with this, but just happened to see a small fix.

Makes sense to remove it, if it isn't used anyway.

sympy/matrices/matrices.py
mkoeppe: This comments should go to #21941

mkoeppe:
> The test error is in (and should be fixed in) Sage right? 
Yes, I have a fix for it in https://trac.sagemath.org/ticket/32420, which makes sure to call `sympy_init` before attempting to call any `_sage_` method. 
But I cannot retroactively fix Sage versions <= 9.4 this way.

mkoeppe:
Here is a complete solution that will also work with already released Sage versions. Also managed to get rid of `Add._sage_` on the way.

mkoeppe:
Sage doctests all pass (modulo the ones caused by changes in the sympy 1.9 series, discussed in #21941).
Ready for review

sympy/core/basic.py
```diff
@@ -1855,6 +1855,22 @@ def _exec_constructor_postprocessors(cls, obj):
 
         return obj
 
+    def _sage_(self):
```
asmeurer, line 1858: Just a heads up, there are a few SymPy classes that do not subclass from Basic. I don't know if you interface with them in Sage, but if you do, you shouldn't assume that they will have this logic.

asmeurer:
Thanks. I'm glad to see this. I made the release notes entry a little clearer.

I wonder if we still need to keep the tests in sympy/external/tests/test_sage.py. I'm not opposed to keeping them, but with this change, they test code that is part of Sage rather than part of SymPy. Either way, Sage itself should probably make sure it has those tests as part of the migration of the code that used to be here.

sympy/core/basic.py
mkoeppe: Thanks! Do you have an example/pointer so I can check this?

mkoeppe:
> I made the release notes entry a little clearer.

Thanks, looking great.

> I wonder if we still need to keep the tests in sympy/external/tests/test_sage.py. I'm not opposed to keeping them, but with this change, they test code that is part of Sage rather than part of SymPy. Either way, Sage itself should probably make sure it has those tests as part of the migration of the code that used to be here.

I haven't checked all details of the tests in the file, but we have very good coverage in the doctests of the conversion methods on the Sage side. @rwst in https://github.com/sympy/sympy/issues/13430#issue-264261809 indicated that `sympy/external/tests/test_sage.py` can be removed. So I'll remove it now.

sympy/core/basic.py
asmeurer: The main example is Matrix, which defaults to a mutable version (so can't subclass from Basic). There are also some things in physics, which it is unlikely you are using. Also a lot of internal stuff like the classes in the polys.

mkoeppe: Thanks. I've taken care of `Matrix` in 554ca46cf4de32b9ffa4af81a667d38d6f220bde

mkoeppe:
Rebased on master.

oscarbenjamin:
Thanks @mkoeppe. This looks good so merging.

mkoeppe:
Great! Thanks very much.

certik:
@mkoeppe thank you for this work!

PR #20853, opened 2021-01-23
 Hamilton's "operators"

mayankray2020:
#### References to other Issues or PRs

Fixes #20769 

#### Brief description of what is fixed or changed
Added the features mentioned in #20769

* algebras
  * quaternion

PR #21966, opened 2021-08-28
quaternion - added Hamilton's "operators"

### LINKED ISSUE #20769: Add the Hamilton's "operators" to the Quaternion class (classical Quaternion approach)
Hamilton in his work describe quite a few "unary operators" (or "characteristic extractors"), some of them required to work with quaternions in the classical approach. Given `q = w + xi + yj + zk`, we have these unary operators:

```
S(q) = w                           # Scalar part
V(q) = xi + yj + zk                # Vector part
U(q) = q / T(q)                    # Unary (versor)
T(q) = sqrt(w² + x² + y² + z²)     # Modulus/magnitude of q ("tensor")
K(q) = S(q) - V(q)                 # Conjugate
N(q) = [T(q)]²                     # Squared magnitude ("norm")
Ax(q) = U[V(q)]                    # Axis
∠(q) = atan2[|V(q)|, S(q)]         # Angle
I(q) = |q| Ax(q)                   # Index vector
R(q) = 1/q = K(q)/[T(q)]²          # Reciprocal (inverse)
M(q) = ln[T(q)]                    # Mensor
```

The names "tensor" and "norm" as used by Hamilton are deprecated, but I think these misleading names should at least appear in the documentation, including an equation with the Hamilton operator (perhaps with the operator symbol in bold face, like what Wikipedia does). At a minimum, the "norm" method documentation should be explicit that it's not the "squared magnitude" definition from Hamilton. Another misleading name is the `am(q)`, which he calls "amplitude", but it's just an even multivalued version of the angle, namely `am(q) = 2*n*π ± ∠(q)`, where "n" is any integer number (the angle ranges from `0` to `π`, and the conjugation doesn't change it).

There are a few other operators, but Hamilton himself tells that only the first five above are required to properly deal with quaternions. On the other hand, "Ax" and "angle" are two ones that he also uses a lot, and these are quite helpful when we think on quaternions, since `q = T(q) * exp(Ax(q) * ∠(q))`, and `pv ln(q) = ln(T(q)) + Ax(q) * ∠(q)`. I can see how to get these with:

```python
from sympy import atan2, ln, Quaternion, symbols
w, x, y, z = symbols("w x y z", real=True)
q = Quaternion(w, x, y, z)
vq = Quaternion(0, q.b, q.c, q.d)  # I'm really missing this one

q.a                        # S(q)
vq                         # V(q)
q.normalize()              # U(q)
q.norm()                   # T(q)
q.conjugate()              # K(q)
q.norm() ** 2              # N(q)
vq.normalize()             # Ax(q)
atan2(vq.norm(), q.a)      # angle(q)
q.norm() * vq.normalize()  # I(q)
q.inverse()                # R(q)
ln(q.norm())               # M(q)
```

I also think this is somehow strange:

```python
>>> # A pure/right quaternion (i.e., purely imaginary)
>>> Quaternion(0, q.b, q.c, q.d).is_imaginary
False
```

It would be nice if methods like `q.scalar()` and `q.vector()` would be available, so that at least the five main Hamilton operators can be used as methods. The `normalize` would perhaps be better named as `q.versor()`, but it's not a big issue. As a workaround, one can use this:

```python
S = lambda q: q.a
V = lambda q: Quaternion(0, q.b, q.c, q.d)
U = Quaternion.normalize
T = Quaternion.norm
K = Quaternion.conjugate
N = lambda q: q.norm() ** 2
Ax = lambda q: Quaternion(0, q.b, q.c, q.d).normalize()
ang = lambda q: atan2(Quaternion(0, q.b, q.c, q.d).norm(), q.a)
I = lambda q: q.norm() * Quaternion(0, q.b, q.c, q.d).normalize()
R = Quaternion.inverse
M = lambda q: ln(q.norm())
```

There are some binary operations that Hamilton uses on quaternions and 3D vectors seen as pure quaternions, apart from the common ones. Two pure quaternions are parallel when their product is commutative (`self * other == other * self`), and they're orthogonal when their product is anti-commutative (`self * other == -other * self`). The plane of a quaternion is the one normal to its axis, and two quaternions are coplanar when they share the same plane. Three pure quaternions are coplanar when one of them is orthogonal to the axis of the division of the other two, that is, when `q ⟂ Ax(r/s)`.

What's this issue is all about? It's mostly a feature request to create some façade methods and to add the classical quaternion terminology to the documentation, but as a summary, it's about:

- [ ] Add `Quaternion.vector()` method
- [ ] Add `Quaternion.scalar()` method
- [ ] Add Hamilton operators to the Quaternion documentation.
- [ ] Add the deprecated terminology "norm" and "tensor" to the documentation, telling these are deprecated, but being really explicit about the fact that `Quaternion.norm()` gives what's currently called "norm", not what Hamilton used to call by that name
- [ ] Fix `Quaternion.is_imaginary` to just check if `q.a == 0`, or create `Quaternion.is_pure`
- [ ] Add `Quaternion.axis()` method
- [ ] Add `Quaternion.angle()` method

Some stuff not really required for most cases, but that might help:

- [ ] Add `Quaternion.versor()` as a new name to `Quaternion.normalize()`
- [ ] Add `Quaternion.coplanar(other)` method to check if `Ax(q) == Ax(r) or Ax(q) == -Ax(r)`
- [ ] Add `Quaternion.ternary_coplanar(q1, q2)` method to check if the pure quaternions `self`, `q1` and `q2` are coplanar in 3D (should enforce that `self.a == q1.a == q2.a == 0`)
- [ ] Add `Quaternion.parallel(other)` method to check if two pure quaternions are parallel (should enforce that `self.a == other.a == 0`)
- [ ] Add `Quaternion.orthogonal(other)` method to check if two pure quaternions are orthogonal (should enforce that `self.a == other.a == 0`)
- [ ] Add `Quaternion.index_vector()`
- [ ] Add `Quaternion.mensor()`

Actually, there are some stuff that might get implemented in another way, since we have some properties that are seen as valid for all quaternions, like `T(U(q)) = 1` for all `q`, although `U(0)` is undefined (that is, even undefined, it still has a magnitude of 1), so `Quaternion(0, 0, 0, 0).normalize().norm()` should be `1`, not `nan`, but I don't know how that could be done. The `Ax(q)` also has unit norm, but it's undefined when `V(q) = 0`. The `∠(0)` is undefined, but it's limited since its range is from `0` to `π`.

The source of all of these is at least one of these (all from Hamilton himself):

- On Quaternions; or on a new System of Imaginaries in Algebra (1844-1850)
- Lectures on Quaternions (1853)
- Elements of Quaternions (1866)

praneethratna:
Fixes #20769 also most part of this PR is taken from unmerged PR #20853 and made changes in code and tests.

#### Release Notes

* algebras
   * Added hamilton's operators in Quaternion class in quaternion.py.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -554,6 +554,7 @@ Bhaskar Joshi <bhaskar.joshi@research.iiit.ac.in> BhaskarJoshi-01 <bhaskar.joshi
 Miguel Torres Costa <miguelptcosta1995@gmail.com>
 Elisha Hollander <just4now666666@gmail.com> donno2048 <just4now666666@gmail.com>
 Mamidi Ratna Praneeth <praneethratna@gmail.com> praneeth <praneethratna@gmail.com>
+Mamidi Ratna Praneeth <praneethratna@gmail.com> Praneeth Ratna <63547155+praneethratna@users.noreply.github.com>
 Zach Carmichael <20629897+craymichael@users.noreply.github.com> Zach <20629897+craymichael@users.noreply.github.com>
 Anup Parikh <parikhanupk@gmail.com> Anup Parikh <84572003+parikhanupk@users.noreply.github.com>
 Harshit Yadav <harshityadav2k@gmail.com> Harshit Yadav <45384915+hyadav2k@users.noreply.github.com>
--- a/sympy/algebras/quaternion.py
+++ b/sympy/algebras/quaternion.py
@@ -3,13 +3,14 @@
 # https://en.wikipedia.org/wiki/Quaternion
 from sympy import S, Rational
 from sympy import re, im, conjugate, sign
-from sympy import sqrt, sin, cos, acos, exp, ln
+from sympy import sqrt, sin, cos, acos, exp, ln, atan2
 from sympy import trigsimp
 from sympy import integrate
 from sympy import Matrix
 from sympy import sympify
 from sympy.core.evalf import prec_to_dps
 from sympy.core.expr import Expr
+from sympy.core.logic import fuzzy_not, fuzzy_or
 
 
 class Quaternion(Expr):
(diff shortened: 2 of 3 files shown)


praneethratna:
@oscarbenjamin please review

praneethratna:
> It seems like PR #20853 is already finished and waiting for review.
> 
> And for the second thing if you take over someone's work , you should start from where they left that is should start over their work.

It has been a long time since that PR has been stalled , so I've taken over it and made some minor changes over that PR's work.

sidhu1012:
I think it would be better to comment and suggest changes on that PR first rather than taking over. If it's really stalled than you can take over.

PR #21967, opened 2021-08-28
functions now accept body arguments

### LINKED ISSUE #21964: Make `Body` and `ReferenceFrame` analogous
`Body` and `ReferenceFrame` are analogous in dynamics, and all the joint classes , jointsmethod and class body follow it, but rest of physics module doesn't.

Example - 

```
>>> A = Body('A')
>>> N = ReferenceFrame('N')
>>> A.dcm(N) # Works
>>> A.dcm(A) # Works

>>> N.dcm('A') # Error but should work

```

So where ever ReferenceFrame works on being passed as arg, passing body to same arg should also work.

It can be done simply by - 
```
if isinstance(frame, Body):
    frame = Body.frame
```

cc - @moorepants

mohajain:

functions now accepts body arguments along with reference frame
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/physics/mechanics/functions.py
+++ b/sympy/physics/mechanics/functions.py
@@ -9,7 +9,6 @@ from sympy.physics.mechanics.rigidbody import RigidBody
 from sympy import simplify
 from sympy.core.backend import (Matrix, sympify, Mul, Derivative, sin, cos,
                                 tan, AppliedUndef, S)
-from sympy.physics.mechanics.body import Body
 
 __all__ = ['inertia',
            'inertia_of_point_mass',
@@ -82,6 +81,7 @@ def inertia(frame, ixx, iyy, izz, ixy=0, iyz=0, izx=0):
     (N.x|N.x) + 2*(N.y|N.y) + 3*(N.z|N.z)
 
     """
+    from sympy.physics.mechanics import Body
     if isinstance(frame, Body):
         frame = Body.frame
     if not isinstance(frame, ReferenceFrame):
@@ -169,6 +169,7 @@ def linear_momentum(frame, *body):
     10*N.x + 500*N.y
 
     """
+    from sympy.physics.mechanics import Body
     if isinstance(frame, Body):
         frame = Body.frame
     if not isinstance(frame, ReferenceFrame):
@@ -229,6 +230,7 @@ def angular_momentum(point, frame, *body):
     10*N.z
 
     """
+    from sympy.physics.mechanics import Body
     if isinstance(frame, Body):
         frame = Body.frame
     if not isinstance(frame, ReferenceFrame):
@@ -292,6 +294,7 @@ def kinetic_energy(frame, *body):
     350
 
     """
+    from sympy.physics.mechanics import Body
     if isinstance(frame, Body):
         frame = Body.frame
     if not isinstance(frame, ReferenceFrame):
@@ -499,6 +502,7 @@ def Lagrangian(frame, *body):
     -M*g*h - g*h*m + 350
 
     """
+    from sympy.physics.mechanics import Body
     if isinstance(frame, Body):
         frame = Body.frame
     if not isinstance(frame, ReferenceFrame):
--- a/sympy/physics/mechanics/functions.py
+++ b/sympy/physics/mechanics/functions.py
@@ -81,7 +81,8 @@ def inertia(frame, ixx, iyy, izz, ixy=0, iyz=0, izx=0):
     (N.x|N.x) + 2*(N.y|N.y) + 3*(N.z|N.z)
 
     """
-
+    if isinstance(frame, Body):
+        frame = Body.frame
     if not isinstance(frame, ReferenceFrame):
         raise TypeError('Need to define the inertia in a frame')
     ol = sympify(ixx) * (frame.x | frame.x)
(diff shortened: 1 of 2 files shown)


mohajain:
Please have a look @sidhu1012

sidhu1012:
I'll look it tomorrow.

moorepants:
You'll need to provide a set of unit tests for these changes.

sympy/physics/mechanics/functions.py
```diff
@@ -81,7 +81,8 @@ def inertia(frame, ixx, iyy, izz, ixy=0, iyz=0, izx=0):
     (N.x|N.x) + 2*(N.y|N.y) + 3*(N.z|N.z)
 
     """
-
+    if isinstance(frame, Body):
```
mostlyaman, line 84: Body class is not defined in this file. Please add the following line at the start of this file with the other import statements.
```python
from sympy.physics.mechanics.body import Body
```

sympy/physics/vector/frame.py
```diff
@@ -53,6 +53,8 @@ def __new__(cls, name, frame, index):
         assumptions = {}
         super()._sanitize(assumptions, cls)
         obj = super().__xnew__(cls, name, **assumptions)
+        if isinstance(frame, Body):
```
mostlyaman, line 56: Body is not defined here too. Add the same line at the start of this file too with the other import statements.

mostlyaman (changes requested):
Undefined Body caused the code quality to fail for these 2 files. Please try pushing the changes requested.

sympy/physics/vector/frame.py
mohajain: I did add this line but it is giving another error
```
ImportError: cannot import name 'CoordinateSym' from partially initialized module 'sympy.physics.vector' (most likely due to a circular import) (/home/psychopirate/Documents/sym/mj/sympy/sympy/physics/vector/__init__.py)
```

mohajain: I added the line and it is giving me this error
```
ImportError: cannot import name 'CoordinateSym' from partially initialized module 'sympy.physics.vector' (most likely due to a circular import) (/home/mj/Documents/sym/mj/sympy/sympy/physics/vector/__init__.py)
```

mostlyaman: I see import Body may have caused a circular import. This can be solved by removing import body from the top of the document.  Instead of at the top, we need to import body whenever it is required i.e. just before every `ifinstance(****, Body)` statement. Something like this but before every `ifinstance` case.
```python
   from sympy.physics.mechanics import Body  
    if isinstance(otherframe, Body):
        otherframe = Body.frame
```
Please note to use `from sympy.physics.mechanics` instead of `from sympy.physics.mechanics.body`. Also do the same for the other file too. I have tested these changes and the tests run fine.

sidhu1012:
> You'll need to provide a set of unit tests for these changes.

And also update the docstrings for changes.

sympy/physics/vector/frame.py
mohajain: Thank you so much! @mostlyaman

sympy/physics/mechanics/functions.py
```diff
@@ -81,7 +81,9 @@ def inertia(frame, ixx, iyy, izz, ixy=0, iyz=0, izx=0):
     (N.x|N.x) + 2*(N.y|N.y) + 3*(N.z|N.z)
 
     """
-
+    from sympy.physics.mechanics import Body
```
sidhu1012, line 84: Imports should go at the top of file unless it gives circular import issue.

sympy/physics/mechanics/functions.py
```diff
@@ -81,7 +81,9 @@ def inertia(frame, ixx, iyy, izz, ixy=0, iyz=0, izx=0):
...
+    if isinstance(frame, Body):
+        frame = Body.frame
```
sidhu1012, line 86: It should be `frame = frame.frame` , correct it here and everywhere else.

sympy/physics/vector/frame.py
```diff
@@ -53,6 +53,9 @@ def __new__(cls, name, frame, index):
         assumptions = {}
         super()._sanitize(assumptions, cls)
         obj = super().__xnew__(cls, name, **assumptions)
+        from sympy.physics.mechanics import Body
+        if isinstance(frame, Body):
+            frame = Body.frame
```
sidhu1012, line 58: Remove it.

sympy/physics/vector/frame.py
```diff
@@ -361,7 +365,9 @@ def variable_map(self, otherframe):
         {A_x: B_x*cos(q(t)) - B_y*sin(q(t)), A_y: B_x*sin(q(t)) + B_y*cos(q(t)), A_z: B_z}
 
         """
-
```
sidhu1012, line 364: Don't remove these blank lines.

sympy/physics/vector/frame.py
```diff
@@ -361,7 +365,9 @@ def variable_map(self, otherframe):
...
+        from sympy.physics.mechanics import Body
+        if isinstance(otherframe, Body):
+            otherframe = Body.frame
```
sidhu1012, line 370: This is not required, private functions most probably don't need this change.

vikas-chaudhary-2802:
Is this Issue Open ??

Issue #20262, opened 2020-10-14
Converting Boolean formulas to Boolean polynomials


jarble:
Does [sympy.logic](https://docs.sympy.org/latest/modules/logic.html) have a method to convert Boolean expressions to Boolean polynomials?

If sympy.logic doesn't have this feature yet, then it could easily be implemented using [these substitution rules](https://math.stackexchange.com/a/3293627/32826).

sylee957:
There is algebraic normal form #13686

Daisyprakash17:
Hey there, i am a beginner i want to work on this issues ,so can you guide me how to start with this issue

oscargus:
This should be relatively simple to implement. However, one should consider what happens when the "variables" are relations. Should it only be possible to do this for symbols? Probably. Since `1-(Eq(a, 0))` gives an error.

Verissimo03:
Has anyone implemented this?

asmeurer:
Not as far as I know, though might grep around the codebase just to check.

Verissimo03:
I'm starting to work on this feature

Verissimo03:
Just to make sure, how should i treat the input?
Can you give me an example so I know how I should implement?

oscarbenjamin:
It is not clear to me what the intended purpose is here. Is it just to display the polynomial? Or is there some expectation of doing algebra with the polynomials?

If the expectation is to do algebra then these polynomials have different properties from other polynomials. Firstly they have coefficients in `GF(2)` which is fine since SymPy has that but not in a way that can be used directly with ordinary expressions. Secondly and this is where it is different from a polynomial over `GF(2)` is that all positive exponents can be reduced to 1 e.g. `x**2 -> x`. Really then if you want to do actual algebra with the polynomials then it needs some special representation that does not exist in SymPy.

On the other hand if the goal is just to output the polynomial as an expression so that someone can look at it then that should be simple to do. Presumably you would also want to be able to do that transformation in reverse i.e. to turn a polynomial expression into a boolean expression.

Verissimo03:
What i did for now, and how I thought it was for example:
Or(a, b) this would print a + b - a * b;
Implies(a, b) this would print 1 - a + a * b;
Equivalent(a, b) this would print 1 - a - b + 2 * a * b;
And(Not(a), Not(b)) this would print (1 - a)*(1 - b)
But I'm not sure that this is what is expected, what do you think?

oscarbenjamin:
[...]

PR #21973, opened 2021-08-29
Added gateinputcount method for Boolean functions

### LINKED ISSUE #18982: [Feature Request]:Calculate GIC in logic expression.
GIC means gate-input-cost, it equals to how many wires entering gates in the logic circuit.

oscargus:

Fixes #18982

A function `gateinputcount()` is added to `boolalg.py` and works for the boolean functions corresponding to gates. This is quite handy as the complexity of a digital circuit is often measures/estimated based on this.

#### Other comments
Ideally one should not rewrite `Nand` and `Not(And())` and correspondingly for `Nor` for this to really make sense. However, it is explicitly mentioned in the documentation.

#### Release Notes

* logic
   * Function `gateinputcount` added to returns the total number of inputs for all classes corresponding to gates in the expression.

### DIFF
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -1338,6 +1338,9 @@ class ITE(BooleanFunction):
     ``ITE(A, B, C)`` evaluates and returns the result of B if A is true
     else it returns the result of C. All args must be Booleans.
 
+    From a logic gate perspective, ITE corresponds to a 2-to-1 multiplexer,
+    where A is the select signal.
+
     Examples
     ========
 
@@ -1754,6 +1757,7 @@ def is_anf(expr):
 
     Examples
     ========
+
     >>> from sympy.logic.boolalg import And, Not, Xor, true, is_anf
     >>> from sympy.abc import A, B, C
     >>> is_anf(true)
@@ -3422,3 +3426,63 @@ def simplify_univariate(expr):
                 c = And(a < x, x < b)
         args.append(c)
     return Or(*args)
+
+
+# Classes corresponding to logic gates
+# Used in gateinputcount method
+BooleanGates = (And, Or, Xor, Nand, Nor, Not, Xnor, ITE)
+
+def gateinputcount(expr):
+    """
+    Return the total number of inputs for the logic gates realizing the
+    Boolean expression.
+
+    Returns
+    =======
+
+    int
+        Number of gate inputs
+
+    Note
+    ====
+
+    Not all Boolean functions count as gate here, only those that are
+    considered to be standard gates. These are: ``And``, ``Or``, ``Xor``,
+    ``Not``, and ``ITE`` (multiplexer). ``Nand``, ``Nor``, and ``Xnor`` will
+    be evalutated to ``Not(And())`` etc.
+
+    Examples
+    ========
+
+    >>> from sympy.logic import And, Or, Nand, Not, gateinputcount
+    >>> from sympy.abc import x, y, z
+    >>> expr = And(x, y)
+    >>> gateinputcount(expr)
+    2
+    >>> gateinputcount(Or(expr, z))
+    4
+
+    Note that ``Nand`` is automatically evaluated to ``Not(And())`` so
+    >>> gateinputcount(Nand(x, y, z))
+    4
+    >>> gateinputcount(Not(And(x, y, z)))
+    4
+
+    Although this can be avoided by using ``evaluate=False``
+    >>> gateinputcount(Nand(x, y, z, evaluate=False))
+    3
+
+    Also note that a comparison will count as a Boolean variable:
+    >>> gateinputcount(And(x > z, y >= 2))
+    2
+
+    As will a symbol:
+    >>> gateinputcount(x)
+    0
+
+    """
+    if not isinstance(expr, Boolean):
+        raise TypeError("Expression must be Boolean")
+    if isinstance(expr, BooleanGates):
+        return len(expr.args) + sum(gateinputcount(x) for x in expr.args)
+    return 0
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -3449,7 +3449,7 @@ def gateinputcount(expr):
     Not all Boolean functions count as gate here, only those that are
     considered to be standard gates. These are: ``And``, ``Or``, ``Xor``,
     ``Not``, and ``ITE`` (multiplexer). ``Nand``, ``Nor``, and ``Xnor`` will
-    be evalutated to ``Not(And())`` etc.
+    be evaluated to ``Not(And())`` etc.
 
     Examples
     ========
--- a/doc/src/modules/logic.rst
+++ b/doc/src/modules/logic.rst
@@ -112,6 +112,8 @@ Disjunctive, and Negated Normal forms:
 
 .. autofunction:: sympy.logic.boolalg::is_nnf
 
+.. autofunction:: sympy.logic.boolalg::gateinputcount
+
 Simplification and equivalence-testing
 --------------------------------------
 
--- a/sympy/logic/__init__.py
+++ b/sympy/logic/__init__.py
@@ -1,11 +1,12 @@
 from .boolalg import (to_cnf, to_dnf, to_nnf, And, Or, Not, Xor, Nand, Nor, Implies,
-    Equivalent, ITE, POSform, SOPform, simplify_logic, bool_map, true, false)
+    Equivalent, ITE, POSform, SOPform, simplify_logic, bool_map, true, false,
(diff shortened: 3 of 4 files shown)


sympy/logic/boolalg.py
```diff
@@ -540,6 +540,40 @@ def _eval_derivative(self, x):
...
             return S.Zero
 
+    def gateinputcount(self):
+        """
+        Return the total number of inputs for the logic gates realizing the
+        Boolean expression.
+
+        Examples
+        ========
+        >>> from sympy.logic import And, Or, Nand, Not
```
smichr, line 550: Is a blank line needed after the `========`?

oscargus: Not needed, but for consistency it should definitely be there! Fixed that.

sympy/logic/boolalg.py
```diff
@@ -540,6 +540,42 @@ def _eval_derivative(self, x):
...
+
+        Returns
+        =======
+
+        int
+            Number of gate inputs
+
+        """
+        return len(self.args) + sum(x.gateinputcount() if isinstance(x, BooleanFunction)
+                                    else 0 for x in self.args)
```
oscarbenjamin, line 577: I think it's better not to split ternary if/else or generator/list comprehensions over multiple lines:
```python
        subgates = [x for x in self.args if isinstance(x, BooleanFunction)]
        return len(self.args) + sum(g.gateinputcount() for g in subgates)
```

oscargus:
Good find! Probably one should create a tuple of `LogicGates` containing `(And, Or, Xor, Not)` for readability.

oscarbenjamin:
I think this might work better as a function rather than a method.

oscarbenjamin:
@oscargus what do you want to do with this? Should it be changed?

oscargus:
@oscarbenjamin I should correct it. Should have some time over the weekend.

oscargus:
I finally read the part about a function, and, yes, it makes much more sense. It will maybe be a bit confusing that `gateinputcount(Integral(sin(x), (x, 0, 3)))` has a gate input count of zero, but it also makes sense as we only are interested in the boolean operations.

I am also reminded that it from this perspective would make sense to not rewrite `Nand` and `Nor` directly (as these gate are actually quite often preferred over `And` and `Or` due to that they are easier to realize in silicon), but that is a different problem.

oscargus:
One may consider if `ITE` should also be supported. It is basically a 2-1 multiplexer, which is a very common gate to use.

oscargus:
I added `ITE` as well and added a minor comment in the documentation.

oscarbenjamin:
> It will maybe be a bit confusing that `gateinputcount(Integral(sin(x), (x, 0, 3)))` has a gate input count of zero

Maybe
```
if not isinstance(expr, Boolean):
    raise TypeError
```

oscargus:
I was considering raising an error, but then I thought about your example, `gateinputcount(And(x, True))` and thought that if one tries to automatically compute this, it is a bit of a hassle if an exception is raised for expressions that are simplified to a single symbol. Where, of course, the symbol may be a relational. And since we do not distinguish between boolean symbols and others, that made sense to me. It now also enable easier recursive implementation (which is not the main benefit, but still shows why it is probably a good idea).

oscarbenjamin:
The benchmarks failure will be fixed by #22015

oscarbenjamin:
Symbol is a subclass of Boolean:
```python
In [2]: Symbol.mro()
Out[2]: 
[sympy.core.symbol.Symbol,
 sympy.core.expr.AtomicExpr,
 sympy.core.basic.Atom,
 sympy.core.expr.Expr,
 sympy.logic.boolalg.Boolean,
 sympy.core.basic.Basic,
 sympy.printing.defaults.Printable,
 sympy.core.evalf.EvalfMixin,
 object]
```
Checking for Boolean will still work for Symbol.

In the future though we should change it so that Symbol is not used for Booleans. The plan would be something like:

1. Introduce a new BooleanSymbol.
2. Ensure that nowhere in the sympy codebase uses plain Symbol in a boolean context.
3. Update all docs to use BooleanSymbol and explain somewhere that Symbol should be changed to BooleanSymbol.
4. After some time/releases have plain Symbol give a deprecation warning when used in a boolean context.
5. After more time/releases make it so that Symbol does not subclass Boolean any more.

None of that would have any effect on this PR checking for Boolean if that is the right thing to check for.

oscarbenjamin:
One more step:

6. Remove the awkward workarounds resulting from Symbol being a Boolean subclass.

sympy/logic/boolalg.py
```diff
@@ -3118,3 +3122,51 @@ def simplify_patterns_xor():
...
+    >>> gateinputcount(Or(expr, z))
+    4
+
+    Note that Nand is automatically rewritten to Not(And()) so
+    >>> gateinputcount(Nand(x, y, z))
+    4
+    >>> gateinputcount(Not(And(x, y, z)))
+    4
+
+    Also note that a comparison will count as a Boolean vairable:
```
oscarbenjamin, line 3153: variable

oscargus: Gah! Will fix!

sympy/logic/boolalg.py
```diff
@@ -3118,3 +3122,51 @@ def simplify_patterns_xor():
                       And(Lt(a, Max(b, c)), Ge(a, Min(b, c)))),
                      )
     return _matchers_xor
+
+
+# Classes corresponding to logic gates
+# Used in gateinputcount method
+BooleanGates = (And, Or, Xor, Nand, Nor, Not, Xnor, ITE)
```
oscarbenjamin, line 3129: Maybe the supported gate types should be listed in the docstring. Could the existing BooleanFunction class be used here instead?

sympy/logic/boolalg.py
```diff
@@ -3118,3 +3122,51 @@ def simplify_patterns_xor():
...
+
+def gateinputcount(expr):
```
oscarbenjamin, line 3131: Will this function automatically show up in the docs somewhere or does a reference to it need be added in an rst file?

sympy/logic/boolalg.py
oscargus: Ideally it would be like that, but I do not think that `Implies`, `Equivalent`, and `Exclusive` should be in there, as there are typically no "gates" that directly implements them (one can of course build gates, but they are not really considered standard and for some of them, the internal complexity is pretty high). This sort of holds for `ITE` as well, a 2-1 multiplexer is more complicated than, say, a three-input `And`-gate, but that is still a very commonly used gate. (`Xor` is also more complicated.)

But it is a quite crude measure...

And, yes, it should be mentioned in the docstring which functions counts as gates.

sympy/logic/boolalg.py
oscargus: Good question! Will try to figure out.

oscargus:
>     1. Introduce a new BooleanSymbol.

And/or introduce a binary/boolean assumption? Will help the transition between the domains if nothing else. See, e.g.,  #20262

(Or then `ElementOf`...)

oscarbenjamin:
> And/or introduce a binary/boolean assumption? Will help the transition between the domains if nothing else

I think boolean assumption would muddy the waters more. We should reserve at least the old assumptions for Expr/NumberKind. A Boolean is just a completely different kind of thing. There is no overlap between booleans and numbers at all.

oscargus:
> There is no overlap between booleans and numbers at all.

Except for, from the top of my head, Zhegalkin polynomials, integer binary vs pseudo-boolean optimization, and Python's `1 == True`. But I'm not going to argue strongly for introducing a `binary` as `ElementOf` is a much more attractive option anyway for the general case of restricted numerical symbols and if there is a `BinarySymbol`, it would be easy to map between these two.

oscargus:
This is updated, including adding it to the documentation.

sympy/logic/boolalg.py
```diff
@@ -3422,3 +3426,63 @@ def simplify_univariate(expr):
...
+    int
+        Number of gate inputs
+
+    Note
+    ====
+
+    Not all Boolean functions count as gate here, only those that are
+    considered to be standard gates. These are: ``And``, ``Or``, ``Xor``,
+    ``Not``, and ``ITE`` (multiplexer). ``Nand``, ``Nor``, and ``Xnor`` will
+    be evalutated to ``Not(And())`` etc.
```
oscarbenjamin, line 3452: evaluated

oscarbenjamin, line 3452:
```suggestion
    be evaluated to ``Not(And())`` etc.
```

oscarbenjamin:
Looks good!

PR #21975, opened 2021-08-29
Minor clean-ups in logic

oscargus:
Some minor code improvements.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -226,7 +226,6 @@ class BooleanAtom(Boolean):
 
     # /// drop when Py2 is no longer supported
     def __lt__(self, other):
-        from sympy.utilities.misc import filldedent
         raise TypeError(filldedent('''
             A Boolean argument can only be used in
             Eq and Ne; all other relationals expect
@@ -450,7 +449,6 @@ class BooleanFunction(Application, Boolean):
         return simplify(self, **kwargs)
 
     def __lt__(self, other):
-        from sympy.utilities.misc import filldedent
         raise TypeError(filldedent('''
             A Boolean argument can only be used in
             Eq and Ne; all other relationals expect
@@ -464,12 +462,12 @@ class BooleanFunction(Application, Boolean):
     def binary_check_and_simplify(self, *args):
         from sympy.core.relational import Relational, Eq, Ne
         args = [as_Boolean(i) for i in args]
-        bin = set().union(*[i.binary_symbols for i in args])
+        bin_syms = set().union(*[i.binary_symbols for i in args])
         rel = set().union(*[i.atoms(Relational) for i in args])
         reps = {}
-        for x in bin:
+        for x in bin_syms:
             for r in rel:
-                if x in bin and x in r.free_symbols:
+                if x in bin_syms and x in r.free_symbols:
                     if isinstance(r, (Eq, Ne)):
                         if not (
                                 S.true in r.args or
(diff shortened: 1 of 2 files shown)


sympy/logic/boolalg.py
```diff
@@ -464,12 +462,12 @@ def __lt__(self, other):
     def binary_check_and_simplify(self, *args):
         from sympy.core.relational import Relational, Eq, Ne
         args = [as_Boolean(i) for i in args]
-        bin = set().union(*[i.binary_symbols for i in args])
```
oscargus, line 467: bin is a keyword

sympy/logic/boolalg.py
```diff
@@ -226,7 +226,6 @@ def _noop(self, other=None):
 
     # /// drop when Py2 is no longer supported
     def __lt__(self, other):
-        from sympy.utilities.misc import filldedent
```
oscargus, line 229: Already imported at top.

sympy/logic/boolalg.py
```diff
@@ -527,9 +525,9 @@ def diff(self, *symbols, **assumptions):
         return Derivative(self, *symbols, **assumptions)
 
     def _eval_derivative(self, x):
-        from sympy.core.relational import Eq
-        from sympy.functions.elementary.piecewise import Piecewise
         if x in self.binary_symbols:
+            from sympy.core.relational import Eq
```
oscargus, line 529: No need to import unless this branch is used.

sympy/logic/boolalg.py
```diff
@@ -595,7 +593,7 @@ def _apply_patternbased_simplification(self, rv, patterns, measure,
             results = []
             # Try all combinations
             for ((i, pi), (j, pj)) in combinations(enumerate(Rel), 2):
-                for k, (pattern, simp) in enumerate(patterns):
+                for pattern, simp in patterns:
```
oscargus, line 596: k is not used.

sympy/logic/boolalg.py
```diff
@@ -640,7 +638,7 @@ def _apply_patternbased_simplification(self, rv, patterns, measure,
                 results = list(reversed(sorted(results,
                                                key=lambda pair: pair[0])))
                 # Replace the one providing most simplification
-                cost, replacement = results[0]
+                replacement = results[0][1]
```
oscargus, line 641: cost not used.

sympy/logic/boolalg.py
```diff
@@ -1285,7 +1283,7 @@ def eval(cls, *args):
             newargs = []
             for x in args:
                 if isinstance(x, Number) or x in (0, 1):
-                    newargs.append(True if x else False)
+                    newargs.append(bool(x))
```
oscargus, line 1286: Should be the same.

sympy/logic/inference.py
```diff
@@ -132,7 +132,7 @@ def valid(expr):
     return not satisfiable(Not(expr))
 
 
-def pl_true(expr, model={}, deep=False):
+def pl_true(expr, model=None, deep=False):
```
oscargus, line 135: Safer to not have empty dicts or lists as default arguments, although it may not always be a problem.

smichr:
looks ok to me

sympy/logic/boolalg.py
oscarbenjamin: It would be good to resolve the circular dependency between `sympy.logic.boolalg` and `sympy.core.relational`. Maybe `boolalg.py` really belongs in the core or maybe `Relational` belongs in the logic module.

sympy/logic/boolalg.py
oscarbenjamin: I often use this pattern to make the code more readable and also to force an error to be raises if `results[0]` has the wrong length:
```python
In [1]: cost, replacement = [1]
ValueError                                Traceback (most recent call last)
<ipython-input-1-98a393f56ecc> in <module>
----> 1 cost, replacement = [1]

ValueError: not enough values to unpack (expected 2, got 1)
```

sympy/logic/boolalg.py
```diff
@@ -1344,7 +1342,7 @@ def __new__(cls, *args, **options):
         for x in args:
             if isinstance(x, Number) or x in [True, False]:  # Includes 0, 1
                 argset.discard(x)
-                argset.add(True if x else False)
+                argset.add(bool(x))
```
oscarbenjamin, line 1345: The condition should probably be `if x in (S.Zero, S.One)`.

I think it's a mistake to accept numbers here altogether.

PR #21976, opened 2021-08-29
Fixed partial subs in Boolean expressions

### LINKED ISSUE #21971: subs is not working in logic expression
subs is not working in logic expression, for a long expression it only replacement the first match subexpression

oscargus:

Fixes #21971

It is now possible to use `subs` where the expression to be replaced is part of a Boolean function.

#### Release Notes

* logic
    * `subs` works better when replacing parts of a Boolean expression.

### DIFF
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -728,8 +728,10 @@ class And(LatticeOp, BooleanFunction):
         # If old is And, replace the parts of the arguments with new if all
         # are there
         if isinstance(old, And):
-            if all(a in args for a in old.args):
-                args = [new] + [a for a in args if a not in old.args]
+            old_args = set(old.args)
+            sd = old_args.symmetric_difference(set(args))
+            if not sd & old_args:
+                args = [new] + list(sd)
 
         return self.func(*args)
 
@@ -881,8 +883,10 @@ class Or(LatticeOp, BooleanFunction):
         # If old is Or, replace the parts of the arguments with new if all
         # are there
         if isinstance(old, Or):
-            if all(a in args for a in old.args):
-                args = [new] + [a for a in args if a not in old.args]
+            old_args = set(old.args)
+            sd = old_args.symmetric_difference(set(args))
+            if not sd & old_args:
+                args = [new] + list(sd)
 
         return self.func(*args)
 
@@ -1152,9 +1156,11 @@ class Xor(BooleanFunction):
         # If old is Xor, replace the parts of the arguments with new if all
         # are there
         if isinstance(old, Xor):
-            if all(a in self.args for a in old.args):
-                args = [new] + [a for a in self.args  if a not in old.args]
-                return self.func(*args)
+            old_args = set(old.args)
+            sd = old_args.symmetric_difference(set(self.args))
+            if not sd & old_args:
+                sd.add(new)
+                return self.func(*sd)
 
 
 class Nand(BooleanFunction):
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -728,10 +728,10 @@ class And(LatticeOp, BooleanFunction):
         # If old is And, replace the parts of the arguments with new if all
         # are there
         if isinstance(old, And):
-            old_args = set(old.args)
-            sd = old_args.symmetric_difference(set(args))
-            if not sd & old_args:
-                args = [new] + list(sd)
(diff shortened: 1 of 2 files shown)


sympy/logic/boolalg.py
```diff
@@ -1136,6 +1148,14 @@ def _eval_simplify(self, **kwargs):
...
             kwargs['measure'], None)
 
+    def _eval_subs(self, old, new):
+        # If old is Xor, replace the parts of the arguments with new if all
+        # are there
+        if isinstance(old, Xor):
+            if all(a in self.args for a in old.args):
+                args = [new] + [a for a in self.args  if a not in old.args]
+                return self.func(*args)
+
```
smichr, line 1158: Might the following be better? Not sure:
```python
>>> o={1,2}
>>> s={1,2,3}
>>> sd = o.symmetric_difference(s)
>>> if not sd & old:  # all args in s are in o
...  print(sd)  # these are the args in s that are not in o
...
{3}
```

oscargus: Probably! Felt a bit wrong to write it like that, but too tired to really think of anything else than brute force...

oscarbenjamin: I find the logic with symmetric difference confusing. Basically this is:
```python
if old_args.issubset(self_args):
    self_args -= old_args
    self_args.add(new)
```

oscarbenjamin: Note that `issubset` can be called with a list/tuple as argument:
```python
In [38]: {1}.issubset((1, 2, 3))
Out[38]: True

In [39]: {1}.issubset((2, 2, 3))
Out[39]: False
```
That means that the common case (condition is false) can be a fast-path that doesn't create any new sets apart from `set(old.args)`.

oscargus: I find both set notations more confusing than the original list comprehension, but I can of course change to any of them.

smichr: I have been confused by `symmetric_difference` but now think of it as the non-trivial difference between two sets: it gives the difference no matter what order you put the args, kind of like taking `abs(x-y)` for two scalars. 

I like @oscarbenjamin suggestion:
```
old_set = set(old)
if old_set.is_subset(args):
    args = set(args) - old_set
    args.update(new)
    return self.func(*args)
```

(with the last line unindented in Xor)

sympy/logic/boolalg.py
```diff
@@ -1136,6 +1152,16 @@ def _eval_simplify(self, **kwargs):
         return self._apply_patternbased_simplification(rv, patterns,
             kwargs['measure'], None)
 
+    def _eval_subs(self, old, new):
+        # If old is Xor, replace the parts of the arguments with new if all
+        # are there
+        if isinstance(old, Xor):
+            old_set = set(old.args)
+            if old_set.issubset(args):
+                args = set(args) - old_set
```
smichr, line 1161:
```suggestion
            if old_set.issubset(self.args):
                args = set(self.args) - old_set
```

sympy/logic/boolalg.py
```diff
@@ -1136,6 +1152,16 @@ def _eval_simplify(self, **kwargs):
...
+                args.update(new)
```
smichr, line 1162:
```suggestion
                args.add(new)
```

sympy/logic/boolalg.py
```diff
@@ -872,6 +880,14 @@ def _eval_subs(self, old, new):
         if bad is not None:
             # let it raise
             bad.subs(old, new)
+        # If old is Or, replace the parts of the arguments with new if all
+        # are there
+        if isinstance(old, Or):
+            old_set = set(old.args)
+            if old_set.issubset(args):
+                args = set(args) - old_set
+                args.update(new)
```
smichr, line 889:
```suggestion
                args.add(new)
```

sympy/logic/boolalg.py
```diff
@@ -725,6 +725,14 @@ def _eval_subs(self, old, new):
         if bad is not None:
             # let it raise
             bad.subs(old, new)
+        # If old is And, replace the parts of the arguments with new if all
+        # are there
+        if isinstance(old, And):
+            old_set = set(old.args)
+            if old_set.issubset(args):
+                args = set(args) - old_set
+                args.update(new)
```
smichr, line 734:
```suggestion
                args.add(new)
```

PR #21979, opened 2021-08-29
Better `is_zero` for LambertW, hyperbolic, and trigonometric functions

oscargus:
Improved zero detection. Sometimes it reported incorrectly earlier, e.g. `cosh(5*I*S.Pi/2, evaluate=False).is_zero` returned `False` while the correct is `True`.

#### Other comments
Primarily pushed to see if it breaks anything else. Not sure if I tested everything...

#### Release Notes

* functions
    * Better zero detection for `LambertW` and some hyperbolic and trigonometric functions.

### DIFF
--- a/sympy/core/tests/test_arit.py
+++ b/sympy/core/tests/test_arit.py
@@ -443,12 +443,12 @@ def test_Add_Mul_is_finite():
     x = Symbol('x', extended_real=True, finite=False)
 
     assert sin(x).is_finite is True
-    assert (x*sin(x)).is_finite is None
+    assert (x*sin(x)).is_finite is False
     assert (x*atan(x)).is_finite is False
     assert (1024*sin(x)).is_finite is True
     assert (sin(x)*exp(x)).is_finite is None
     assert (sin(x)*cos(x)).is_finite is True
-    assert (x*sin(x)*exp(x)).is_finite is None
+    assert (x*sin(x)*exp(x)).is_finite is False
 
     assert (sin(x) - 67).is_finite is True
     assert (sin(x) + exp(x)).is_finite is not True
--- a/sympy/core/tests/test_arit.py
+++ b/sympy/core/tests/test_arit.py
@@ -443,12 +443,12 @@ def test_Add_Mul_is_finite():
     x = Symbol('x', extended_real=True, finite=False)
 
     assert sin(x).is_finite is True
-    assert (x*sin(x)).is_finite is False
+    assert (x*sin(x)).is_finite is None
     assert (x*atan(x)).is_finite is False
     assert (1024*sin(x)).is_finite is True
     assert (sin(x)*exp(x)).is_finite is None
     assert (sin(x)*cos(x)).is_finite is True
-    assert (x*sin(x)*exp(x)).is_finite is False
+    assert (x*sin(x)*exp(x)).is_finite is None
 
     assert (sin(x) - 67).is_finite is True
     assert (sin(x) + exp(x)).is_finite is not True
--- a/sympy/functions/elementary/hyperbolic.py
+++ b/sympy/functions/elementary/hyperbolic.py
@@ -272,8 +272,8 @@ class sinh(HyperbolicFunction):
 
     def _eval_is_zero(self):
         rest, ipi_mult = _peeloff_ipi(self.args[0])
-        return fuzzy_and([ipi_mult.is_integer,
-                          rest.is_zero])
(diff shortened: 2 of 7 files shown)


sympy/core/tests/test_arit.py
```diff
@@ -443,12 +443,12 @@ def test_Add_Mul_is_finite():
     x = Symbol('x', extended_real=True, finite=False)
 
     assert sin(x).is_finite is True
-    assert (x*sin(x)).is_finite is None
+    assert (x*sin(x)).is_finite is False
```
oscargus, line 446: Here it is not obvious what should be returned.

`(x/S.Pi).is_integer` returns `False`, but can `"sin(oo)"` still possibly be zero? And is `oo*0 (= NaN)` in that case finite?

oscarbenjamin: It's correct that `(x/S.Pi).is_integer` should give `False` because an integer is always finite. Perhaps something somewhere should be checking for `finite` or perhaps more precisely `real`.

I think that `(x*sin(x)).is_finite` should be `None`.

sympy/functions/elementary/hyperbolic.py
```diff
@@ -269,9 +269,7 @@ def _eval_is_finite(self):
         return arg.is_finite
 
     def _eval_is_zero(self):
-        arg = self.args[0]
-        if arg.is_zero:
-            return True
+        return (self.args[0]/(S.Pi*I)).is_integer
```
oscarbenjamin, line 272: This kind of thing is very tricky and should be avoided if possible. What this does is create a new Basic instance as part of an assumptions query. There are several reasons this can be problematic:

1. The assumptions system should lie on a level below evaluation so that it can be called as part of evaluation. Invoking evaluation as part of the assumptions system itself is circular.
2. Apart from making a messy design the circularity creates a risk of infinite recursion where an object directly or indirectly recreates itself as part of the assumptions query.
3. The `is_zero` predicate is called in many places e.g. `Mul.flatten` and as a result any change can have wide performance impact.
4. If the expression is large then putting it into a Mul can be expensive as `Mul.flatten` needs to rerun.

Also it is incorrect to conclude that `(self.args[0]/(S.Pi*I)).is_integer = False` implies `self.is_zero = False` because of e.g. `zoo`, `nan` etc. There needs to be a guard here like checking `self.args[0].is_complex`.

sympy/core/tests/test_arit.py
oscargus: How can it possibly be finite if x is infinite?

sympy/functions/elementary/hyperbolic.py
oscargus: Regarding incorrectness, see my other comment. Can really `sin(zoo)` be considered to possibly be zero?

As far as I can tell, there is no need to check for complex here as a real value will return an imaginary value that is never integer.

For the rest of the comments, I see the potential risk of circular execution. I just wonder how to get around that `cosh(5*I*S.Pi/2, evaluate=False).is_zero` currently returns `False`.

Is there a "safe" way to extract `Pi` from the expression?

sympy/core/tests/test_arit.py
oscarbenjamin: You can have something well defined like:
```python
In [1]: limit(x*sin(1/x**2), x, oo)
Out[1]: 0
```
Here the limit is:
```python
In [1]: limit(x*sin(x), x, oo)
Out[1]: ∞⋅sign(<-1, 1>)
```
which is a round about way of saying that in the limit this ends up stretching over all possible real numbers including zero.

It's important that the assumptions system gives `None` in any case that is ambiguous. More complicated techniques like limits and series and so on should be used to resolve the ambiguity.

sympy/functions/elementary/hyperbolic.py
oscarbenjamin: There's a function somewhere called something like `peel_off_pi`.

jksuom: It seems that the tests run by `_peeloff_pi` are almost the same: https://github.com/sympy/sympy/blob/281b1904acc9e15ed04402cff33ebd6ef7812026/sympy/functions/elementary/trigonometric.py#L127-L129

oscarbenjamin: That can definitely be improved. In the meantime it would be better if all places doing similar things use the same function so that it can be improved in one place.

Issue #18497, opened 2020-01-28
Add ElementOf and ArbitraryConstant symbols


oscarbenjamin:
I have two different ideas here that are related in that they could share a lot their implementation. Perhaps having just one is useful.

The first is that I think it would be good to create an `ElementOf` symbol type that represents an element of a particular `Set` e.g.:
```python
>>> x = ElementOf('x', Reals)
>>> x.is_real
True
>>> y = ElementOf('x', Interval(1, 2))
>>> y.is_positive
True
```
The idea is that this forms a connection between sets and expressions that you can use as a more powerful way of expressing some "assumptions" about an expression. The new assumptions could be taught to recognise this for more complicated queries and perhaps some other operations could work like
```python
>>> y in Reals
True
>>> Eq(y, 3)
False
```
The `ElementOf` instances represent some symbol with the knowledge that it is an element of some set so like `Symbol` they have
```python
>>> ElementOf('x', Reals) == ElementOf('x', Reals)
True
```

That last result would not hold in the case of `ArbitraryConstant`. The idea of this class is that it is used in the return from solvers like `dsolve`. We currently have e.g.:
```julia
In [5]: dsolve(f(x).diff(x))
Out[5]: f(x) = C₁

In [6]: diophantine(x**2 + y**2 - z**2)
Out[6]: 
⎧⎛        2    2   2    2⎞⎫
⎨⎝2⋅p⋅q, p  - q , p  + q ⎠⎬
⎩                         ⎭

In [15]: diophantine(x+2*y)
Out[15]: {(2⋅t₀, -t₀)}

In [14]: solve([x*y**2, z+y], [x,y,z])
Out[14]: [(0, -z, z), (x, 0, 0)]

In [9]: nonlinsolve([x+y], [x,y])
Out[9]: {(-y, y)}
```
I think that in all these cases it would be better to return a symbol that is registered as being a member of a particular set and that is explicitly identified as an arbitrary constant. These would be symbols similar to `ElementOf` except they function as `Dummy` symbols under comparison e.g.:
```julia
>>> ArbitraryConstant('C1', Complexes) == ArbitraryConstant('C1', Complexes)
False
```
These need to compare unequal like `Dummy` because e.g. two distinct integration constants are not equal just because they have the same name (there are bugs in dsolve as a result of not using `Dummy` for this #16902).

[...]

PR #21985, opened 2021-08-30
Replace symbol with Dummy that has correct assumptions in function_range (mininum/maximum)

oscargus:

Pretty much sorts out #21956, but the rewrite for `Abs` is not implemented so not sure if we should close it.

As `maximum`, `minimum`, `function_range` only accepts a real-valued domain it makes sense to replace the symbol with a Dummy that has correct assumptions (i.e. real and possibly nonnegative/nonpositive). 

#### Other comments
It can probably be further improved by considering open intervals so `Interval.Lopen(0, 1)` defines a positive Dummy rather than a nonnegative one.

Also, `function_range` now explicitly checks that the domain is indeed real.

#### Release Notes

* calculus
    * Better support for `minimum` and `maximum` by automatically adding assumptions on variables.


sympy/calculus/util.py
```diff
@@ -190,7 +193,33 @@ def function_range(f, symbol, domain):
                 else:
                     vals += FiniteSet(f.subs(symbol, limit_point))
 
-            solution = solveset(f.diff(symbol), symbol, interval)
+            # Replace symbol with dummy with proper assumptions if the symbol
+            # not already has it
+            l = (interval.inf, interval.sup)
+            if all(i.is_nonnegative for i in l) and not symbol.is_nonnegative:
+                d = Dummy(positive=True)
+            elif all(i.is_nonpositive for i in l) and not symbol.is_nonpositive:
```
oscarbenjamin, line 201: I think that this should be abstracted out to a function (even if incomplete and private) because it is something that can be useful in many situations.

oscarbenjamin:
See also #18497 which suggests a more general approach along these lines.

sympy/calculus/util.py
oscargus: Yeah, I thought that there was some, but stole it directly from the integral code.

However, I seem to recall that there may be something in concrete that possibly can be modified (as it probably has integer constraints).

oscargus:
`ElementOf` would be very useful in many situations for sure. I've earlier been thinking about being able to provide `min` and `max` values to a Symbol, which is sort of similar. Although sets can of course be made discontinuous (and complex).

sympy/calculus/util.py
smichr: cf #19523

oscargus:
The "hanging"(?) computation is `variance(X + a*Y + b)`.

oscargus:
The failing test will most likely pass after #22052 is merged. That may also affect #22075 so will wait and see.

sympy/concrete/summations.py
```diff
@@ -14,7 +14,8 @@
 from sympy.core.relational import Eq
 from sympy.core.singleton import S
 from sympy.core.sorting import ordered
-from sympy.core.symbol import Dummy, Wild, Symbol, symbols
+from sympy.core.symbol import Dummy, Wild, Symbol, symbols, _dummy_with_inherited_properties_concrete
+from sympy.functions.special.zeta_functions import zeta
```
oscargus, line 18:
```suggestion
```

sympy/calculus/util.py
smichr:
```python
>>> from sympy.core.assumptions import *
>>> l=Tuple(1, oo)
>>> symbol = Dummy(real=1)
>>> common = common_assumptions(l)
>>> if failing_assumptions(symbol, **common):
...    reps = {symbol: Dummy(**common)}
...
>>> failing_assumptions(reps[symbol], **common)
{}
>>> len(failing_assumptions(symbol, **common))
13
```

smichr:
Comment above with reference to using `failing_assumptions`.

Issue #21968, opened 2021-08-28
failure in master?


smichr:
I saw the following in the most recent run of #21903

```console
________________________________________________________________________________
 sympy/matrices/expressions/tests/test_blockmatrix.py:test_BlockMatrix_2x2_inverse_numeric 
Traceback (most recent call last):
  File "/home/runner/work/sympy/sympy/sympy/matrices/expressions/blockmatrix.py", line 313, in schur
    inv = (d[mat].T*d[mat]).inv()*d[mat].T if generalized else d[mat].inv()
  File "/home/runner/work/sympy/sympy/sympy/matrices/matrices.py", line 2204, in inv
    return _inv(self, method=method, iszerofunc=iszerofunc,
  File "/home/runner/work/sympy/sympy/sympy/matrices/inverse.py", line 459, in _inv
    rv = M.inverse_GE(iszerofunc=iszerofunc)
  File "/home/runner/work/sympy/sympy/sympy/matrices/matrices.py", line 2189, in inverse_GE
    return _inv_GE(self, iszerofunc=iszerofunc)
  File "/home/runner/work/sympy/sympy/sympy/matrices/inverse.py", line 245, in _inv_GE
    raise NonInvertibleMatrixError("Matrix det == 0; not invertible.")
sympy.matrices.common.NonInvertibleMatrixError: Matrix det == 0; not invertible.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/runner/work/sympy/sympy/sympy/matrices/expressions/tests/test_blockmatrix.py", line 234, in test_BlockMatrix_2x2_inverse_numeric
    assert block_collapse(K.inv()).as_explicit() == K.as_explicit().inv()
  File "/home/runner/work/sympy/sympy/sympy/matrices/expressions/blockmatrix.py", line 722, in block_collapse
    result = rule(expr)
  File "/home/runner/work/sympy/sympy/sympy/strategies/core.py", line 8, in exhaustive_rl
    new, old = rule(expr), expr
  File "/home/runner/work/sympy/sympy/sympy/strategies/core.py", line 41, in chain_rl
    expr = rule(expr)
  File "/home/runner/work/sympy/sympy/sympy/strategies/core.py", line 8, in exhaustive_rl
    new, old = rule(expr), expr
  File "/home/runner/work/sympy/sympy/sympy/strategies/core.py", line 30, in conditioned_rl
    return rule(expr)
[...]

PR #11196, opened 2016-06-06
NEW: longest_repetition, seq_replace, split, find

smichr:
I propose utility functions for `iterables` that will be useful for #10271 but do not otherwise depend on that PR.

The main function is something that can be used to identify repeated subsequences of contiguous elements (more precisely a repeated "substring" although we are not working with strings) in one or more sequences. (The proper scope of the subsequence problem is to identify an ordered sequence of elements -- not necessarily contiguous -- that appears is two sequences, e.g. 'abe' for 'abcde' and 'alabaster'. A description and code [are here](http://www.geeksforgeeks.org/printing-longest-common-subsequence/).)

``` python
>>> s = [[1, 2, 3], [1, 1, 2], [1, 1, 2, 3]]
>>> extract_repetitions(numbered_symbols(), *s)
([(x1, [1, 2]), (x0, [1, x1])], [[x1, 3], [x0], [x0, 3]])
```

The supporting functions allow for the following:
- Find a subsequence

``` python
>>> s=ibin(23)
>>> s
[1, 0, 1, 1, 1]
>>> find(s, [0,1])
1
>>> find(s, [1], all=True)
[0, 2, 3, 4]
```
- Split a sequence on given elements

``` python
>>> s
[1, 0, 1, 1, 1]
>>> split(s, 0)
[[1], [1, 1, 1]]
```
- replace a subsequence in-place

``` python
>>> s
[1, 0, 1, 1, 1]
>>> seq_replace(s,[0,1], 2);s
[1, 2, 1, 1]
```
- find repetitions in a sequence

``` python
>>> s=ibin(23545)
>>> s
[1, 0, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 0, 0, 1]
>>> longest_repetition(s)  # first lexical
[0, 1, 1]
>>> longest_repetition(s, all=True)
[[0, 1, 1], [1, 0, 1], [1, 1, 0], [1, 1, 1]]
>>> longest_repetition(s, all=True, max=2)
[[0, 1], [1, 0], [1, 1]]
```

To demonstrate how it can be used with several open issues related to cse, consider the following:

``` python
def _cse(*e):
    muls = all(i.is_Mul or i.is_Pow for i in e)
    if not muls:
        assert all(i.is_Add for i in e)
        F = Add
    else:
        F = Mul
    def remap(*e):
        args = list(ordered(set(flatten(F.make_args(i) for i in e))))
        mapping = dict(zip(args, range(len(args))))
[...]

PR #21986, opened 2021-08-30
Modified coeff() to give correct value for a sum

### LINKED ISSUE #16752: coeff() fails with non-commutative argument
Diffgeom results in a wrong result for the double Lie Bracket of three vectors 
Commutator(Commutator(v1,v2),v3)
where
v1 = exp(r)*e_r + exp(r**2)*e_theta
v2 = sin(theta)*e_r
v3 = e_theta+e_r 
xs = r,    theta   = spherical.coord_functions()
es = e_r,  e_theta = spherical.base_vectors() 

Full example in [sympy googlegroup thread](https://groups.google.com/forum/#!topic/sympy/zmGIsxmzV0c).

This leads to violation of Jacobi identities for Commutator.

Ideas:
> It looks like there is a bug in Expr.coeff. It cannot find the coefficient of e_r in Commutator(v1, v2), apparently because of non-commutativity.  K Souminen

### LINKED ISSUE #17917: Wrong result in LieDerivative
I'll reproduce the steps that go to a wrong result:

```python
from sympy.diffgeom import Manifold, Patch, CoordSystem, LieDerivative
M = Manifold('M', 2)
P = Patch('P', M)
coord = CoordSystem("coord", P, ["x", "y"])
dx, dy = coord.base_oneforms()
ex, ey = coord.base_vectors()
x, y = coord.coord_functions()
X = x*ex-y*ey
Y = (x**2+y**2)*ex-x*y*ey
LieDerivative(X, Y)
>> -x*y*ey
```
But this is absolutely wrong, the correct Lie Derivative of Y w.r.t X is (x^2 - 3*y^2)&PartialD;x-xy&PartialD;y as you can easily check. Sympy is omitting the first term

What is happening here?

t-ens:
Fixes #16752 
Fixes #17917

Given expressions e and n which are both noncommutative when computing e.coeff(n) instead of immediately returning 0 if n is found in more than one summand of e it now only returns 0 when finding n in a new summand if, after removing n from both summands, the remaining noncommutative part of the new summand  and previous summand are not equal.  Added tests for this behaviour as well.

#### Other comments
Before if everything was commutative then for example `(x*a+y*a+z*b).coeff(a)` returned `x+y` and even if `a,b` were not commutative `(x*a+y*a).coeff(a)` returned `x+y` however for `a,b` not commutative `(x*a+y*a+z*b).coeff(a)` returned `0`.  With this change now `(x*a+y*a+z*b).coeff(a)` returns `x+y` in all cases. This fixes the incorrect answers given by `LieDerivative` and `Commutator` from the diffgeom submodule referenced above however it is slower so there is a performance hit.

* core
  * Extended coeff() to return nonzero value for some sums with noncommutative variables

### DIFF
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -1629,7 +1629,7 @@ class Expr(Basic, EvalfMixin):
                         if right:
                             np = n[:ii]
                         else:
-                            np = n[ii+len(nx):]
+                            np = n[ii + len(nx):]
                     else:
                         if right:
                             if n[:ii] != np:
@@ -1638,11 +1638,11 @@ class Expr(Basic, EvalfMixin):
                             if n[ii+len(nx):] != np:
                                 break
                     if right:
-                        hit.append(({},n[ii+len(nx):]))
+                        hit.append(({}, n[ii + len(nx):]))
                     else:
                         hit.append((r, n[:ii]))
             else:
-                if hit is not None:
+                if hit:
                     return Add(*[Mul(*(list(r) + n)) for r, n in hit])
 
             return S.Zero
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -1619,22 +1619,31 @@ class Expr(Basic, EvalfMixin):
                         return Add(*[Mul(*(list(r) + n[:-len(end) + ii])) for r, n in co])
                     else:
                         return Mul(*end[ii + len(nx):])
-            # look for single match
-            hit = None
+            # look for a single hit but allow a sum of terms if each have the
+            # same remaining expression after removing coefficient
+            hit = []
             for i, (r, n) in enumerate(co):
                 ii = find(n, nx, right)
                 if ii is not None:
                     if not hit:
-                        hit = ii, r, n
+                        if not right:
+                            np = n[ii+len(nx):]
+                        else:
+                            np = n[:ii]
                     else:
-                        break
-            else:
-                if hit:
-                    ii, r, n = hit
+                        if not right:
+                            if n[ii+len(nx):] != np:
+                              break
+                        else:
+                            if n[:ii] != np:
+                                break
                     if not right:
-                        return Mul(*(list(r) + n[:ii]))
+                        hit.append((r, n[:ii]))
                     else:
-                        return Mul(*n[ii + len(nx):])
+                        hit.append(({},n[ii+len(nx):]))
+            else:
+                if hit is not None:
+                    return Add(*[Mul(*(list(r) + n)) for r, n in hit])
 
             return S.Zero
 
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -1558,6 +1558,39 @@ class Expr(Basic, EvalfMixin):
                 i = len(l) - (i + n)
             return i
 
+        def special_cases(co, nx, right):
+            if all(n == co[0][1] for r, n in co):
+                ii = find(co[0][1], nx, right)
+                if ii is not None:
+                    if right:
+                        return Mul(*co[0][1][ii + len(nx):])
+                    else:
+                        return Mul(Add(*[Mul(*r) for r, c in co]), Mul(*co[0][1][:ii]))
+            beg = reduce(incommon, (n[1] for n in co))
+            if beg:
+                ii = find(beg, nx, right)
+                if ii is not None:
+                    if right:
+                        m = ii + len(nx)
+                        return Add(*[Mul(*(list(r) + n[m:])) for r, n in co])
+                    else:
+                        gcdc = co[0][0]
+                        for i in range(1, len(co)):
+                            gcdc = gcdc.intersection(co[i][0])
+                            if not gcdc:
+                                break
+                        return Mul(*(list(gcdc) + beg[:ii]))
+            end = list(reversed(
+                reduce(incommon, (list(reversed(n[1])) for n in co))))
+            if end:
+                ii = find(end, nx, right)
+                if ii is not None:
+                    if right:
+                        return Mul(*end[ii + len(nx):])
+                    else:
+                        return Add(*[Mul(*(list(r) + n[:-len(end) + ii])) for r, n in co])
+            return None
+
         co = []
         args = Add.make_args(self)
         self_c = self.is_commutative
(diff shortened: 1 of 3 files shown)


oscarbenjamin:
The tests failed because of the changes to the authors file. I'm about to disable them in #21987. Once that is merged you can close and reopen the PR to rerun the tests.

smichr:
cf #11196

sympy/core/expr.py
```diff
@@ -1619,22 +1619,31 @@ def find(l, sub, first=True):
...
+                        if not right:
+                            np = n[ii+len(nx):]
+                        else:
+                            np = n[:ii]
                     else:
-                        break
-            else:
-                if hit:
-                    ii, r, n = hit
+                        if not right:
```
oscargus, line 1634: Minor thing, but I'd swap the order of these if-else clauses so that the negation is avoided.

sympy/core/tests/test_expr.py
```diff
@@ -1414,6 +1414,8 @@ def test_coeff():
     assert (n*m + o*m*n).coeff(m*n) == o
     assert (n*m + o*m*n).coeff(m*n, right=1) == 1
     assert (n*m + n*m*n).coeff(n*m, right=1) == 1 + n  # = n*m*(n + 1)
+    assert (x*n + y*n + z*m).coeff(n) == x+y
+    assert (n*m +n*o + o*l).coeff(n,right=1) == m+o
```
oscargus, line 1418: Spaces around +

sympy/core/expr.py
t-ens: I did it this way because whoever wrote the function originally wrote if-else clauses with negation(for example line 1594 and a few other places.) I fixed my own contribution to avoid the negation. Should I also change the other occurrences of this in the coeff() function which I did not write?

sympy/core/tests/test_expr.py
t-ens: Oops, just noticed there should also be a space before right in the second one.  I'll commit this together with the other if-else clause changes if I need to do those.

sympy/core/expr.py
oscargus: I think it is enough to change it here (although not required). Unless there is a "pedagogical" benefit of starting with the not-clause it is marginally better not to. But there is indeed a lot of code where it is not like that. I actually tried the other day to fix it systematically, but in some parts of the code base, it really needs to be written with not first...

oscargus:
Oh, I didn't mention that, but it really makes sense to add tests for the issues you close. For #16752 it is probably enough to add a comment where you test x*a + y*a + z*b like `# issue #16752`, but for #17917 please add a test `test_issue_17917` in https://github.com/sympy/sympy/blob/master/sympy/diffgeom/tests/test_diffgeom.py

t-ens:
After testing the function a bit more I realized that even though it passed all the tests my change is not correct. For example if `e=xuvu+yuvw+zw` my code gives `e.coeff(v, right=True)=u+w` because I only checked whether the noncommutative parts to the left of `v` are equal not the commutative parts and then added the coefficient from each summand. This is incorrect. If I simply  ensure that both the noncommutative part of the expression to the left of `v` and the commutative parts are equal, then I would get `e.coeff(v, right=True)=0` which is also incorrect since `xuvu+yuvw=uv(xu+yw)` so the correct answer should be `xu+yw`.

On the other hand using the original code before I modified anything we correctly get `(xuvu+yuvw).coeff(v, right=True)=xu+yw` as we should, however now `e.coeff(v,right=True) = 0`. The problem is because of the extra term `zw` which does not contain `v` at all which causes the tests for special cases where there is a common factor in the noncommutative part of each term to fail. What I really need to do is remove all terms which do not have `v` at all, and then use the algorithms already implemented on the remaining terms which do more complicated factoring than I was doing.

So the most obvious solution would be to simply revert my changes back to the way they were and add the line
`co = [t for t in co if find(t[1], nx, right) is not None]`
in line 1589. I've tested this and it solves the problems with the issues in the PR, passes all the tests and also gives the right answer for the above example.

Is this an acceptable solution? I ask because it seems like it is quite inefficient to apply the `find()` function to every term right away like this  since the special cases sometimes only require you to apply this function to a single term so for large expressions containing many terms adding my above line could cause a significant slowdown. One way I see to only have the slowdown when it is actually required would be to add an argument to the coeff() function, say `reduced` and have it be false by default. When `e.coeff(n)` is called the function would run to line 1622. If the all the special cases fail and `reduced` is False then use my line above to remove any term not containing `n` and then call coeff() again this time with `reduced=True` and the new expression so that now maybe the special cases apply now that any terms not containing `n` have been removed. If all the special cases fail, now reduced is True so now search for a single match.

Also since I am essentially changing my pull request completely, is it best to close this pull request and start a new one or just add another commit to this one? Sorry, I'm new to this and not really sure what the best way to proceed is.

t-ens:
I have added the requested test and addressed all the other issues that were still relevant with my new update. As per my previous post I changed approach in my latest commit since my previous algorithm did not work in all cases. Rather than running the line
` co = [t for t in co if find(t[1], nx, right) is not None]`
for every expression I thought it would be more efficient to first try the special cases, then only run this line if necessary since the special cases only require performing the find() one time rather than once for every term in the expression. I also added a test for the example in the previous post which was incorrect in both my updated version and the original version of the function.

I'm not sure if I should close this pull request and make a new one or not since I changed my approach completely and reverted most of my original code changes. Let me know if I should or if there is a problem with the new approach.

oscargus:
I think that you should continue in the same PR.

A bit strange that it fails in 3.10 and passes in the other versions...

oscarbenjamin:
> A bit strange that it fails in 3.10 and passes in the other versions...

There is an intermittent failure in 3.10 in CI (#21968).

t-ens:
I believe the issue with the test was fixed in #22029 so closed and reopened to rerun the tests and also removed the draft status since I think it works correctly now.

sympy/core/tests/test_expr.py
```diff
@@ -1414,6 +1414,9 @@ def test_coeff():
     assert (n*m + o*m*n).coeff(m*n) == o
     assert (n*m + o*m*n).coeff(m*n, right=1) == 1
     assert (n*m + n*m*n).coeff(n*m, right=1) == 1 + n  # = n*m*(n + 1)
+    assert (x*n + y*n + z*m).coeff(n) == x + y #i issue #16752
+    assert (n*m + n*o + o*l).coeff(n, right=1) == m + o
+    assert (x*n*m*n + y*n*m*o + z*l).coeff(m, right=1) == x*n + y*o
```
smichr, line 1419: I think we need an assurance the coefficient of the left and right will recreate the terms that were matched:

```
>>> (x*n*m*n + y*n*m*o + z*l).coeff(m, right=0)
n
>>> (x*n*m*n + y*n*m*o + z*l).coeff(m, right=1)
x*n + y*o
```
That's right because `factor_nc(x*n*m*n + y*n*m*o) == n*m*(x*n + y*o)`. But consider the following:
```python
>>> (x*n*m*n + x*n*m*o + z*l).coeff(m, right=0)
x*n
>>> (x*n*m*n + x*n*m*o + z*l).coeff(m, right=1)
x*n + x*o
```
The `x` cannot be part of the left and part of the right.

smichr:
Doesn't this single change accomplish the same thing as this proposal?
From this
```python
    if self_c and not x_c:
        return S.Zero
```
to this:
```python
    if self_c and not x_c:
        return S.Zero
    if self.is_Add and not self_c and not x_c:
        d = self.as_independent(x)[1]
        return coeff(d, x, right=right)
```
The issue of returning the commutative part as both left and right coeff still remains, however, and is present in master.

smichr:
Am trying alternative in #22137

smichr:
@t-ens , what do you think about #22137?

t-ens:
I agree that if `e` is some expression containing `x` and `a`, `b` are the left and right coefficients of `x` in `e` then it is reasonable to expect that `e = axb`. #22140 looks like a better implementation of what I was trying to do here plus it addresses this reconstruction issue as well so this pull request can be closed in favour of that one.

PR #21988, opened 2021-08-31
Added orient functions to Body class

### LINKED ISSUE #21965: Add `orient` functions to `Body`
Since `Body` and `ReferenceFrame` are analogous , all the orient functions should be added to class body so that we can simply do:-

```
>>> A = Body('A')
>>> B = Body('B')
>>> A.orent_axis(B, axis , angle)
```

cc - @moorepants

mostlyaman:

Fixes #21965

orient_axis, orient_explicit, orient_body_fixed, orient_space_fixed, orient_quaternion methods are added to Body class.

#### Other comments
I have also found some errors in the documentation of physics.vector.frame and they have been fixed here as well.

#### Release Notes

* physics.mechanics
  * `orient_axis`, `orient_explicit`, `orient_body_fixed`, `orient_space_fixed`, `orient_quaternion` methods are added to Body class.


sidhu1012:
You need to add tests cases to ensure it runs fine amd gives correct answer.

mostlyaman:
https://github.com/sympy/sympy/blob/ed4604016775a113f0c4aa754ece296699bb24ce/sympy/physics/mechanics/body.py#L523-L527
I have used this bit of code in this PR. If by user error, if `body` is neither a `Body` object or `ReferenceFrame` object, then this will cause a UnboundLocalError as frame variable will never be defined.
```python
UnboundLocalError: local variable 'frame' referenced before assignment
```
This error is not telling the user what the problem is. I suggest adding another else condition which can raise an helpful error message.
```python
if isinstance(body,  ReferenceFrame): 
     frame=body 
elif isinstance(body, Body): 
     frame = body.frame 
else:
    raise(TypeError, "body should be Body or ReferenceFrame object.")
```
Should this be added in this PR?

sidhu1012:
@mostlyaman In tests check the dcm matrix also, add one more assertion in each test to check ghe value of dcm matrix.

mostlyaman:
@sidhu1012 I have updated the test with your suggestion. Also, is my suggestion above worth adding in this PR?
> This error is not telling the user what the problem is. I suggest adding another else condition which can raise an helpful error message.

sympy/physics/mechanics/body.py
```diff
@@ -562,3 +562,368 @@ def dcm(self, body):
...
+        >>> B.orient_axis(A, -A.x, q1)
+        >>> B.orient_axis(A, A.x, -q1)
+
+        """
+
+        if isinstance(parent, Body):
+            frame = parent.frame
+        elif isinstance(parent, ReferenceFrame):
+            frame = parent
+        self.frame.orient_axis(frame, axis, angle)
```
sidhu1012, line 631: Regarding https://github.com/sympy/sympy/pull/21988#issuecomment-909147200

This code can be
```suggestion
        if isinstance(parent, Body):
            parent = parent.frame
        self.frame.orient_axis(parent, axis, angle)
```

This way the `Body` is turned into frame in parent variable and `orient` function would handle proper error mg.

sidhu1012: Correct it everywhere.

mostlyaman: Done.

sympy/physics/mechanics/body.py
```diff
@@ -557,8 +553,361 @@ def dcm(self, body):
...
+        [0, sin(q1),  cos(q1)]])
+
+        The following two lines show that the sense of the rotation can be
+        defined by negating the vector direction or the angle. Both lines
+        produce the same result.
+
+        >>> B.orient_axis(A, -A.x, q1)
+        >>> B.orient_axis(A, A.x, -q1)
+
+        """
```
sidhu1012, line 619: I don't think we need such detailed examples here as they are already present in original function so it would be wise to add a simple example here and everywhere else and add a see also section which points to original orient function.

sidhu1012: Same goes to docstrings , don't elaborate as in original just the necessary part.

mostlyaman:
I have updated the docstrings according to your suggestion. I have kept the information that may be required for any new user to see the API syntax and one example. I have also added See Also linking the orient functions to their ReferenceFrame counterparts.

sidhu1012:
Will review it in detail tomorrow.

sidhu1012:
@moorepants Do you want to review it before I merge?

sympy/physics/mechanics/tests/test_body.py
```diff
@@ -273,3 +274,53 @@ def test_apply_loads_on_multi_degree_freedom_holonomic_system():
...
     assert W.loads == [(W.masscenter, (c*q1.diff() + k*q1)*W.x)]
+
+def test_orient_axis():
+    A = Body('A')
+    B = Body('B')
+    N = ReferenceFrame('N')
+    A.orient_axis(N, N.x + N.y, 10)
+    B.orient_axis(A, A.x + A.y, 10)
+    dcm = Matrix([[cos(10)/2 + S.Half, S.Half - cos(10)/2, -sqrt(2)*sin(10)/2], [S.Half - cos(10)/2,
+    cos(10)/2 + S.Half, sqrt(2)*sin(10)/2], [sqrt(2)*sin(10)/2, -sqrt(2)*sin(10)/2, cos(10)]])
```
sidhu1012, line 285: Use tabs for proper indentation , else everything seems fine.

sidhu1012: Check other matrices indentation.

sidhu1012:
Symengine test is failing.

sympy/physics/mechanics/tests/test_body.py
mostlyaman: https://github.com/sympy/sympy/blob/2c83657ff1c62fc2761b639469fdac7f7561a72a/sympy/physics/vector/tests/test_frame.py#L470
I saw matrices being written in a single line here so I used this format. Let me know if this is ok.

sympy/physics/vector/frame.py
```diff
@@ -884,10 +884,10 @@ def orient_body_fixed(self, parent, angles, rotation_order):
...
                                            'body', rot_order)
             templist = [expand(i) for i in templist]
-            td = solve(templist, [u1, u2, u3])
-            u1 = expand(td[u1])
-            u2 = expand(td[u2])
-            u3 = expand(td[u3])
+            td = solve(templist, [u1, u2, u3], dict = True)
+            u1 = expand(td[0][u1])
+            u2 = expand(td[0][u2])
+            u3 = expand(td[0][u3])
```
mostlyaman, line 890: I have fixed the error which symengine was giving earlier. Although, it looks like it is unable to find a solution to the equations it is trying to solve which returns empty list causing index error.

mostlyaman: @sidhu1012 I have been unable to find out why the `td[0]` is giving a keyerror. Maybe someone familiar with the code can help me there. But I was successful in recreating the dictionary functionality using sets. I did the following changes and ran symengine test and it has passed.
```python
            td = solve(templist, [u1, u2, u3], set = True)
            u1 = expand(list(td[1])[0][td[0].index(u1)])
            u2 = expand(list(td[1])[0][td[0].index(u2)])
            u3 = expand(list(td[1])[0][td[0].index(u3)])
```

oscarbenjamin: > I have found that changing u1, u2, u3 to dynamicsymbols does yield a result but idk why then it cannot find u1 as a dictionary index.

Can you show the code needed to setup the inputs to solve that leads to this problem?

There is probably a bug but it needs a reproducer (that doesn't involve calling the functions in the mechanics module).

mostlyaman: It is indeed a bug and a very odd one at that. Somehow changing some other function changed whether it passed the test or not. I can open a new issue describing all the oddities that I experienced with it.

sympy/physics/vector/frame.py
```diff
@@ -1,6 +1,7 @@
-from sympy.core.backend import (diff, expand, sin, cos, sympify, eye, symbols,
+from sympy.core.backend import (diff, expand, sin, cos, sympify, eye,
                                 ImmutableMatrix as Matrix, MatrixBase)
-from sympy import (trigsimp, solve, Symbol, Dummy)
+from sympy.core.sympify import sympify as sympify2
```
mostlyaman, line 3: I have found that `core.backend.sympify` and `core.sympify.sympify` have different behaviours to symengine. While the former changes to `symengine.lib.symengine_wrapper.sympify`, the latter remains a sympy function. This has the effect that the former starts converting its inputs to symengine objects while the latter works as usual converting its inputs to sympy objects. So, I had to import the latter as a separate function.

sympy/physics/mechanics/body.py
```diff
@@ -488,11 +488,9 @@ def masscenter_vel(self, body):
...
         """
 
-        if isinstance(body,  ReferenceFrame):
-            frame=body
-        elif isinstance(body, Body):
-            frame = body.frame
-        return self.masscenter.vel(frame)
+        if isinstance(body, Body):
+            body = body.frame
+        return self.masscenter.vel(body)
```
sidhu1012, line 493: Class `Body` doesn't need these changes.

mostlyaman: Noted. I thought this suffered the same problem as the other one where I made this change, so I added it here too.

sidhu1012: No remove changes from Body.

sidhu1012:
```
-from sympy.core.backend import (diff, expand, sin, cos, sympify, eye,
+from sympy.core.backend import (diff, expand, sin, cos, sympify, eye, symbols,
                                 ImmutableMatrix as Matrix, MatrixBase)
-from sympy.core.sympify import sympify as sympify2
-from sympy import (trigsimp, solve, Symbol)
+from sympy import (trigsimp, solve, Symbol, Dummy)
 from sympy.physics.vector.vector import Vector, _check_vector
 from sympy.utilities.misc import translate
 
@@ -879,17 +878,17 @@ def orient_body_fixed(self, parent, angles, rotation_order):
 
         try:
             from sympy.polys.polyerrors import CoercionFailed
-            from sympy.physics.vector.functions import kinematic_equations, dynamicsymbols
+            from sympy.physics.vector.functions import kinematic_equations
             q1, q2, q3 = amounts
-            u1, u2, u3 = dynamicsymbols('u1, u2, u3')
+            u1, u2, u3 = symbols('u1, u2, u3', cls=Dummy)
             templist = kinematic_equations([u1, u2, u3], [q1, q2, q3],
                                            'body', rot_order)
             templist = [expand(i) for i in templist]
-            td = solve(templist, [u1, u2, u3], dict = True)
-            u1, u2, u3 = map(sympify2, [u1, u2, u3])
-            u1 = expand(td[0][u1])
-            u2 = expand(td[0][u2])
-            u3 = expand(td[0][u3])
+            td = solve(templist, [u1, u2, u3])
+            u1, u2, u3 = map(sympify, [u1, u2, u3])
+            u1 = expand(td[u1])
+            u2 = expand(td[u2])
+            u3 = expand(td[u3])
             wvec = u1 * self.x + u2 * self.y + u3 * self.z
         except (CoercionFailed, AssertionError):
             wvec = self._w_diff_dcm(parent)
@@ -1002,17 +1001,17 @@ def orient_space_fixed(self, parent, angles, rotation_order):
 
         try:
             from sympy.polys.polyerrors import CoercionFailed
-            from sympy.physics.vector.functions import kinematic_equations, dynamicsymbols
+            from sympy.physics.vector.functions import kinematic_equations
             q1, q2, q3 = amounts
-            u1, u2, u3 = dynamicsymbols('u1, u2, u3')
+            u1, u2, u3 = symbols('u1, u2, u3', cls=Dummy)
             templist = kinematic_equations([u1, u2, u3], [q1, q2, q3],
                                            'space', rot_order)
             templist = [expand(i) for i in templist]
-            td = solve(templist, [u1, u2, u3], dict = True)
-            u1, u2, u3 = map(sympify2, [u1, u2, u3])
-            u1 = expand(td[0][u1])
-            u2 = expand(td[0][u2])
-            u3 = expand(td[0][u3])
+            td = solve(templist, [u1, u2, u3])
+            u1, u2, u3 = map(sympify, [u1, u2, u3])
+            u1 = expand(td[u1])
+            u2 = expand(td[u2])
+            u3 = expand(td[u3])
             wvec = u1 * self.x + u2 * self.y + u3 * self.z
         except (CoercionFailed, AssertionError):
             wvec = self._w_diff_dcm(parent)
```

This will fix the failing symengine test without breaking other tests.

sympy/physics/mechanics/tests/test_body.py
sidhu1012: https://github.com/sympy/sympy/blob/25a3b8ec67a2c0ae99d0be2e9f01d294e207e1ad/sympy/physics/mechanics/tests/test_joint.py#L82-L84

Make it like this.

sidhu1012:
The problem is when we `Symengine` the solve output is a list and without it is a dict.

sympy/physics/mechanics/tests/test_body.py
sidhu1012: On the second though it seems fine.

sidhu1012:
with `Symengine` flag set to true , the answer is wrong. The test case is failing as solve is not being able to find a answer and is returning an empty list instead of a solution.

Without symengine the test case runs fine.

cc @oscarbenjamin

mostlyaman:
> The test case is failing as solve is not being able to find a answer and is returning an empty list instead of a solution.

Maybe the solution is not computed because symengine is unable to convert u1, u2, u3 to symengine objects. In my previous changes, when u1, u2, u3 were dynamicsymbols, solve was able to find an answer. But now when they are Dummy symbols, it is not able to. And here we can also see that dynamicsymbols are converted to symengine objects while symbols are not.
```python
>>> type(dynamicsymbols('a'))
<class 'symengine.lib.symengine_wrapper.FunctionSymbol'>
>>> type(symbols('a', cls=Dummy))
<class 'sympy.core.symbol.Dummy'>
>>> type(symbols('a'))
<class 'sympy.core.symbol.Symbol'>
>>> type(Symbol('a'))
<class 'sympy.core.symbol.Symbol'>
```
Here we can see that only dynamicsymbols are automatically converted to symengine objects while symbols are not converted in any case. To furthur support my theory, the test run fine when we specify u1, u2, u3 as symengine symbols from the start indicating that u1, u2, u3 are being passed as sympy objects and symengine is unable to compute any answers with them. 
```python
>>> from symengine.lib.symengine_wrapper import Symbol
>>> u1, u2, u3 = map(Symbol, 'u1 u2 u3'.split())
>>> type(u1)
<class 'symengine.lib.symengine_wrapper.Symbol'>
>>> from sympy import symbols, expand, solve
>>> from sympy.physics.vector import kinematic_equations
>>> q1, q2, q3 = symbols('q1 q2 q3')
>>> eq = kinematic_equations([u1, u2, u3], [q1, q2, q3], 'body', 'XZX')
>>> eq = [expand(i) for i in eq]
>>> td = solve(eq, [u1, u2, u3])
>>> td
{u1: 0, u2: 0, u3: 0}
```
This is just from my tinkering with symengine for a few days without any previous knowledge of how symengine tests work, so I may be very wrong. Still passing symengine symbols explicitly when USE_SYMENGINE is 1 may be a way to make the tests work.

mostlyaman:
> The test case is failing as solve is not being able to find a answer and is returning an empty list instead of a solution.

Maybe the solution is not computed because symengine is unable to convert u1, u2, u3 to symengine objects. In my previous changes, when u1, u2, u3 were dynamicsymbols, solve was able to find an answer. But now when they are Dummy symbols, it is not able to. And here we can also see that dynamicsymbols are converted to symengine objects while symbols are not.
```python
>>> type(dynamicsymbols('a'))
<class 'symengine.lib.symengine_wrapper.FunctionSymbol'>
>>> type(symbols('a', cls=Dummy))
<class 'sympy.core.symbol.Dummy'>
>>> type(symbols('a'))
<class 'sympy.core.symbol.Symbol'>
>>> type(Symbol('a'))
<class 'sympy.core.symbol.Symbol'>
```
Here we can see that only dynamicsymbols are automatically converted to symengine objects while symbols are not converted in any case. To furthur support my theory, the test run fine when we specify u1, u2, u3 as symengine symbols from the start indicating that u1, u2, u3 are being passed as sympy objects and symengine is unable to compute any answers with them.
```python
>>> from symengine.lib.symengine_wrapper import Symbol
>>> u1, u2, u3 = map(Symbol, 'u1 u2 u3'.split())
>>> type(u1)
<class 'symengine.lib.symengine_wrapper.Symbol'>
>>> from sympy import symbols, expand, solve
>>> from sympy.physics.vector import kinematic_equations
>>> q1, q2, q3 = symbols('q1 q2 q3')
>>> eq = kinematic_equations([u1, u2, u3], [q1, q2, q3], 'body', 'XZX')
>>> eq = [expand(i) for i in eq]
>>> td = solve(eq, [u1, u2, u3])
>>> td
{u1: 0, u2: 0, u3: 0}
```
This is just from my tinkering with symengine for a few days without any previous knowledge of how symengine tests work, so I may be very wrong. Still passing symengine symbols explicitly when USE_SYMENGINE is 1 may be a way to make the tests work.
I have no idea why symbols are not being converted.

oscarbenjamin:
Rather than importing directly from symengine you should use `sympy.core.backend`. If that is used consistently (in the mechanics code and the tests) then it should work with and without symengine. Otherwise there might be a bug somewhere else.

mostlyaman:
I never bothered to look at core.backend. I didn't knew it linked to symengine. My bad!

mostlyaman:
I apologize if I was not clear in my explanation above. I meant to say that passing cls = sympy.Dummy prevents them from being converted to symengine objects. 

If we remove the argument, the tests run fine. Is this change acceptable?

I searched a lot to find a Dummy variable for symengine but I could not find docs for the symengine wrapper. The Dummy available in symengine wrapper gives error with solve.
```python
>>> solve(a-1, a)
TypeError: symengine.lib.symengine_wrapper.Symbol.__new__(symengine.lib.symengine_wrapper.Dummy) is not safe, use symengine.lib.symengine_wrapper.Dummy.__new__()
```

sidhu1012:
@mostlyaman I'll try to look into this.

sidhu1012:
@mostlyaman Use symbols in your tests, I have fixed that `symengine` part that was failing.

sidhu1012:
```
______ sympy/physics/mechanics/tests/test_body.py:test_orient_body_fixed _______
  File "/home/runner/work/sympy/sympy/sympy/physics/mechanics/tests/test_body.py", line 304, in test_orient_body_fixed
    A.orient_body_fixed(N, (1,1,0), 'XYX')
  File "/home/runner/work/sympy/sympy/sympy/physics/mechanics/body.py", line 721, in orient_body_fixed
    self.frame.orient_body_fixed(parent, angles, rotation_order)
  File "/home/runner/work/sympy/sympy/sympy/physics/vector/frame.py", line 885, in orient_body_fixed
    templist = kinematic_equations([u1, u2, u3], [q1, q2, q3],
  File "/home/runner/work/sympy/sympy/sympy/physics/vector/functions.py", line 306, in kinematic_equations
    return [q1d - (w2 * s3 + w3 * c3) / s2, q2d - w2 * c3 + w3 *
  File "symengine_wrapper.pyx", line 873, in symengine.lib.symengine_wrapper.Basic.__mul__
  File "symengine_wrapper.pyx", line 74, in symengine.lib.symengine_wrapper.c2py
TypeError: symengine.lib.symengine_wrapper.Symbol.__new__(symengine.lib.symengine_wrapper.Dummy) is not safe, use symengine.lib.symengine_wrapper.Dummy.__new__()
______ sympy/physics/mechanics/tests/test_body.py:test_orient_space_fixed ______
  File "/home/runner/work/sympy/sympy/sympy/physics/mechanics/tests/test_body.py", line 315, in test_orient_space_fixed
    A.orient_space_fixed(N, (1,2,3), 'ZXY')
  File "/home/runner/work/sympy/sympy/sympy/physics/mechanics/body.py", line 776, in orient_space_fixed
    self.frame.orient_space_fixed(parent, angles, rotation_order)
  File "/home/runner/work/sympy/sympy/sympy/physics/vector/frame.py", line 1010, in orient_space_fixed
    td = solve(templist, [u1, u2, u3])
  File "/home/runner/work/sympy/sympy/sympy/solvers/solvers.py", line 849, in solve
    f, symbols = (_sympified_list(w) for w in [f, symbols])
  File "/home/runner/work/sympy/sympy/sympy/solvers/solvers.py", line 849, in <genexpr>
    f, symbols = (_sympified_list(w) for w in [f, symbols])
  File "/home/runner/work/sympy/sympy/sympy/solvers/solvers.py", line 840, in _sympified_list
    return list(map(sympify, w if iterable(w) else [w]))
  File "/home/runner/work/sympy/sympy/sympy/core/sympify.py", line 389, in sympify
    return a._sympy_()
  File "symengine_wrapper.pyx", line 2150, in symengine.lib.symengine_wrapper.Add._sympy_
  File "/home/runner/work/sympy/sympy/sympy/core/cache.py", line 72, in wrapper
  File "/home/runner/work/sympy/sympy/sympy/core/operations.py", line 53, in __new__
    args = list(map(_sympify_, args))
  File "/home/runner/work/sympy/sympy/sympy/core/sympify.py", line 505, in _sympify
  File "/home/runner/work/sympy/sympy/sympy/core/sympify.py", line 389, in sympify
    return a._sympy_()
  File "symengine_wrapper.pyx", line 2201, in symengine.lib.symengine_wrapper.Mul._sympy_
  File "symengine_wrapper.pyx", line 975, in symengine.lib.symengine_wrapper.Basic.args.__get__
  File "symengine_wrapper.pyx", line 809, in symengine.lib.symengine_wrapper.vec_basic_to_tuple
  File "symengine_wrapper.pyx", line 815, in symengine.lib.symengine_wrapper.vec_basic_to_list
  File "symengine_wrapper.pyx", line 74, in symengine.lib.symengine_wrapper.c2py
TypeError: symengine.lib.symengine_wrapper.Symbol.__new__(symengine.lib.symengine_wrapper.Dummy) is not safe, use symengine.lib.symengine_wrapper.Dummy.__new__()
```

This has been fixed in upstream and should start passing once the changes have been included in the new release.

mostlyaman:
Maybe I should add a TODO comment to replace the symbols used with a Dummy after the new symengine release.

sidhu1012:
I don't think we need a todo for this.

mostlyaman:
There's quite a bit of different things that are changed when symengine is being used. ReferenceFrame.variable_map is one of such functions.
```python
>> A = ReferenceFrame('A')
>>> A.variable_map(A)
{A_x: A_x, A_y: A_y, A_z: A_z}
>>> type(list(A.variable_map(A).values())[0])                 # with USE_SYMENGINE = 0
<class 'sympy.physics.vector.frame.CoordinateSym'>
>>> type(list(A.variable_map(A).values())[0])                 # with USE_SYMENGINE = 1
<class 'symengine.lib.symengine_wrapper.Symbol'>
```
I am not sure if this is the expected behaviour but it breaks this test only while using symengine.
https://github.com/sympy/sympy/blob/6c10460a45535b860dbbc9b2e8c1071ec8ffec13/sympy/physics/vector/tests/test_frame.py#L52

sidhu1012:
The only cause of failure in this case was that the `Dummy` instance was not being converted into symengine type, since we now import Dummy from symengine all variables would be of symengine type and everything should work as desired.

sympy/physics/vector/frame.py
```diff
@@ -1002,14 +1004,14 @@ def orient_space_fixed(self, parent, angles, rotation_order):
...
             templist = kinematic_equations([u1, u2, u3], [q1, q2, q3],
                                            'space', rot_order)
             templist = [expand(i) for i in templist]
             td = solve(templist, [u1, u2, u3])
-            u1 = expand(td[u1])
-            u2 = expand(td[u2])
-            u3 = expand(td[u3])
+            u1 = expand(td[sympy.sympify(u1)])
+            u2 = expand(td[sympy.sympify(u2)])
+            u3 = expand(td[sympy.sympify(u3)])
```
sidhu1012, line 1014: @moorepants `symengine` tests fail in case of `orient_body` and `orient_space` , this fix by @mostlyaman is some what a temporary fix, i.e, replace dummy by symbols.

The correct fix according to me is https://github.com/sympy/sympy/pull/21988/commits/25bac97129d4b59a4a4b367e2929aa0119faba1b, which is to import `Dummy` from `symengine` itself, it was failing earlier due to an error in `symengine` that was fixed in #22082. But by doing this the symengine tests would fail until the newer symengine version is released.

So I suggest we put that test in `xfail` until the newer symengine is released.

What do you suggest @moorepants ?

moorepants: If you want to put a symengine only test into xfail that is fine. Just open an issue about fixing the xfail, so it is tracked.

sidhu1012: @mostlyaman https://github.com/sympy/sympy/commit/25bac97129d4b59a4a4b367e2929aa0119faba1b is the right fix according to me, you shouldn't use symbols like that, so please revert  these changes and add the failing symengine tests to `xfail` which can be tracked and resolved once new symengine release is published.

mostlyaman:
I think there is another bug. Its with `sympy.sympify` adding an extra underscore when converting an symengine Dummy object to sympy Dummy object. I am using symengine development version from github so it contains all the latest changes.
```python
>>> import sympy, symengine
>>> a = symengine.symbols('a', cls = symengine.Dummy)
>>> a
_a
>>> sympy.sympify(a)
__a
```
`sympify` should return a dummy without the additional underscore. Since, solve converts symengine Dummy to sympy Dummy, the tests are still failing.

sidhu1012:
@mostlyaman Can you also share the exact error message solve gives.

sidhu1012:
I found the bug and it's solution, will fix it tomorrow.

mostlyaman:
I didn't understand you. What "test with symbols" is required?

mostlyaman:
> Just add a test with symbols in every new test function to make syre everything is working.

@sidhu1012 Could you please explain what sort of test is required in this PR, if you have some time. Other than that, this PR was ready to be merged. Maybe some other person can review this too.

sujantkumarkv:
Hey! I request updates on this from the team if [this issue](https://github.com/sympy/sympy/issues/21965) it's solved or needs work?

Issue #20374, opened 2020-11-02
Travis CI .com migration and free plans limits


asmeurer:
Travis CI is migrating all .org builds to .com by the end of the year (see https://docs.travis-ci.com/user/migrate/open-source-repository-migration#frequently-asked-questions). We are currently using .org. I don't know if we need to do anything for this, or if it will happen automatically.

A more important issue is that Travis is limiting the number of builds per month on their free plan to 10000 minutes per month. See https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing. I don't how many minutes we use per month currently. If we are well under 10000, then we don't need to worry about this. But if we go over, we may want to consider moving part or all of our builds to another CI service. I don't think we will be able to tell how many minutes we use until the .com migration happens, though we can probably estimate it from the current build logs.

asmeurer:
Actually it's 10000 credits per month, but there are 10 credits per minute for Linux builds. 

Based on https://travis-ci.org/github/sympy/sympy/pull_requests, each successful build takes about 9 total hours, so this means we would only have enough free credits for 2 builds. They say we can also request more free credits, but this doesn't seem like it is even close to what we are currently using, unless I am doing the math on it wrong.

oscarbenjamin:
We should look seriously at github actions.

asmeurer:
The existing alternatives are GitHub Actions, Circle CI, and Azure Pipelines. I agree GitHub actions is probably the best alternative, though we might still want to investigate the others. My biggest concern with GitHub actions is that they don't retain old build logs past 90 days https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration#artifact-and-log-retention-policy. It's pretty common to reference a build log in an issue or something, and we would have to avoid doing that. It also means if we go back to an old PR that had a failing build, we can't see why. 

I also dislike the UI and configuration for GitHub actions, but I actually have that issue with all the services (Travis does much better than them in this area IMO). Azure Pipelines is actually the worst here. We tried it in the past, but unless they have improved things since then, I would recommend avoiding it, because figuring out how their configuration works and the permissions in their UI is a major pain. Although we do have an existing Azure Pipelines setup in the git history, so it would be the quickest to move to it.

oscarbenjamin:
Travis is throttling our builds and in general has a backlog of open source jobs. You can see the current status here:
https://www.traviscistatus.com/

[...]

PR #21992, opened 2021-08-31
maint(ci): run symengine tests in CI (USE_SYMENGINE=1)

oscarbenjamin:

#20374

Although symengine was being installed in the optional dependency job `USE_SYMENGINE=1` was not set so the tests did not actually use symengine. This adds a new `bin/test_symengine.py` script that can be used with or without `USE_SYMENGINE=1` to run the tests that can involve symengine wither with or without symengine.

This also ensures that symengine is actually being tested in CI

#### Other comments

There are already failures with symengine in mechanics. It seems to be to do with a change in symengine rather than in sympy.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/physics/mechanics/tests/test_joint.py
+++ b/sympy/physics/mechanics/tests/test_joint.py
@@ -1,4 +1,5 @@
 from sympy import sin, cos, Matrix, sqrt, pi, expand_mul, S
+from sympy.core.backend import _simplify_matrix
 from sympy.core.symbol import symbols
 from sympy.physics.mechanics import dynamicsymbols, Body, PinJoint, PrismaticJoint
 from sympy.physics.mechanics.joint import Joint
@@ -78,9 +79,9 @@ def test_pin_joint_double_pendulum():
                                [sin(q1), cos(q1), 0], [0, 0, 1]])
     assert A.dcm(B) == Matrix([[cos(q2), -sin(q2), 0],
                                [sin(q2), cos(q2), 0], [0, 0, 1]])
-    assert N.dcm(B).simplify() == Matrix([[cos(q1 + q2), -sin(q1 + q2), 0],
-                                          [sin(q1 + q2), cos(q1 + q2), 0],
-                                          [0, 0, 1]])
+    assert _simplify_matrix(N.dcm(B)) == Matrix([[cos(q1 + q2), -sin(q1 + q2), 0],
+                                                 [sin(q1 + q2), cos(q1 + q2), 0],
+                                                 [0, 0, 1]])
 
     # Check Angular Velocity
     assert A.ang_vel_in(N) == u1 * N.z
@@ -206,7 +207,7 @@ def test_pinjoint_arbitrary_axis():
              child_axis=A.x+A.y)
     assert expand_mul(N.x.angle_between(A.x + A.y)) == 0  # Axis are aligned
     assert (A.x + A.y).express(N).simplify() == sqrt(2)*N.x
-    assert A.dcm(N).simplify() == Matrix([
+    assert _simplify_matrix(A.dcm(N)) == Matrix([
         [sqrt(2)/2, -sqrt(2)*cos(theta)/2, -sqrt(2)*sin(theta)/2],
         [sqrt(2)/2, sqrt(2)*cos(theta)/2, sqrt(2)*sin(theta)/2],
         [0, -sin(theta), cos(theta)]])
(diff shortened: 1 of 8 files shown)


oscarbenjamin:
CC @moorepants there are CI failures with the mechanics module when using symengine.

The same tests fail with older sympy versions so I presume that this is due to a change in symengine rather than sympy. It might be easy to fix in sympy even if it is a bug in symengine.

oscarbenjamin:
The mechanics module is inconsistent in how it handles mutable and immutable matrices e.g.:
https://github.com/sympy/sympy/blob/281b1904acc9e15ed04402cff33ebd6ef7812026/sympy/physics/vector/frame.py#L524-L526
Here `eye(3)` creates a mutable matrix. Depending on the other matrices it is multiplied with in the loop the return value may be mutable or may be immutable.

There is also an inconsistency between symengine's matrix and sympy's in the simplify method:
```python
>>> print(Matrix([[1]]).simplify())
None
>>> from sympy.core.backend import ImmutableMatrix as Matrix
>>> print(Matrix([[1]]).simplify())
None
>>> from sympy import Matrix as Matrix
>>> print(Matrix([[1]]).simplify())
None
>>> from sympy import ImmutableMatrix as Matrix
>>> print(Matrix([[1]]).simplify())
Matrix([[1]])
```
The symengine ImmutableMatrix.simplify method does not return the simplified result.

oscarbenjamin:
Thanks for reviewing. I wanted this fixed before the release.

I've reported the SymEngine bug in https://github.com/symengine/symengine.py/issues/363

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -176,7 +177,7 @@ def test_pinjoint_arbitrary_axis():
     assert A.ang_vel_in(N).express(A) == omega * A.y
     assert A.ang_vel_in(N).magnitude() == sqrt(omega**2)
     angle = A.ang_vel_in(N).angle_between(A.y)
-    assert expand_mul(angle).xreplace({omega: 1}) == 0
+    assert angle.xreplace({omega: 1}) == 0
```
oscarbenjamin, line 180: I'm not sure why this needed to be changed. The `expand_mul` is not needed and doesn't seem to do anything to the expression with or without symengine. However with symengine after `expand_mul` the call to `xreplace` fails:
```python
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/enojb/current/sympy/sympy/sympy/physics/mechanics/tests/test_joint.py(181)test_pinjoint_arbitrary_axis()
-> assert angle.xreplace({omega: 1}) == 0
(Pdb) p angle
acos(omega_J(t)/sqrt(omega_J(t)**2))
(Pdb) p expand_mul(angle)
acos(omega_J(t)/sqrt(omega_J(t)**2))
(Pdb) p angle.xreplace({omega:1})
0
(Pdb) p expand_mul(angle).xreplace({omega:1})
acos(omega_J(t)/sqrt(omega_J(t)**2))
```

oscarbenjamin: Oh, it's because `expand_mul` sympifies the expression from a SymEngine expression to a SymPy expression:
```python
(Pdb) p type(angle)
<class 'symengine.lib.symengine_wrapper.acos'>
(Pdb) p type(expand_mul(angle))
acos
```
If we also apply `expand_mul` to `omega` then xreplace works:
```python
(Pdb) p expand_mul(angle).xreplace({expand_mul(omega):1})
0
(Pdb) p expand_mul(angle).xreplace({omega:1})
acos(omega_J(t)/sqrt(omega_J(t)**2))
```
Looks like `xreplace` does not sympify its args:
https://github.com/sympy/sympy/blob/281b1904acc9e15ed04402cff33ebd6ef7812026/sympy/core/basic.py#L1081-L1166

oscarbenjamin: A SymEngine expression compares equal to a SymPy expression via `sympify`. However the hashes don't match so they are not identified as equivalent by dict/set:
```python
(Pdb) p angle == expand_mul(angle)
True
(Pdb) p angle in {expand_mul(angle)}
False
(Pdb) hash(angle)
879987935
(Pdb) hash(expand_mul(angle))
-1535576704778178373
```
This breaks the hash invariant:
```
x = y    -->    hash(x) = hash(y)
```

Issue #21991, opened 2021-08-31
Should UnevaluatedExpr.is_real query its argument?


bjodah:
Consider this:
```python
>>> from sympy import symbols as s, UnevaluatedExpr as UE
>>> r, s = s('r s', real=True)
>>> (r + s).is_real
True
>>> UE(r + s).is_real
>>>  # that is, it was "None"
```
Currently I need to treat UnevaluatedExpr as a special case in some logic. I don't see any reason why `.is_real`, `.is_complex` etc. can't be made to return `self.args[0].is_real`, ...?

But perhaps I'm missing something?

oscarbenjamin:
The question is whether you want to allow evaluation of expressions containing UnevaluatedExpr:
```python
In [1]: cos(UnevaluatedExpr(0))
Out[1]: cos(0)

In [2]: x = Symbol('x', zero=True)

In [3]: cos(0)
Out[3]: 1
```

bjodah:
I don't think so. I would argue that it's unrelated for my use case, and that I think it is (very) good that it's not evaluated.

Here's an example:
```python
In [4]: x, y, z = symbols('x y z', real=True)
In [5]: ccode(Abs(exp(UnevaluatedExpr(UnevaluatedExpr(y+z) + x))))
Out[5]: '// Not supported in C:\n// re\nexp(re(x + y + z))'
```

oscarbenjamin:
> I would argue that it's unrelated for my use case

The assumptions can be implemented like this:
```diff
diff --git a/sympy/core/expr.py b/sympy/core/expr.py
index 6ff45211dd..265994bc81 100644
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -4006,6 +4006,9 @@ def doit(self, **kwargs):
         else:
             return self.args[0]
 
+    def _eval_is_zero(self):
+        return self.args[0].is_zero
+
 
 
 def unchanged(func, *args):
```
However that change will cause expressions involving UnevaluatedExpr to evaluate in some situations:
```python
In [1]: cos(UnevaluatedExpr(0))
Out[1]: 1
```
The mechanism that prevents expressions involving UnevaluatedExpr from evaluating is basically just the fact that it has no properties that any evaluation code checks for. If you add assumptions to it then that will change:
https://github.com/sympy/sympy/blob/ed4604016775a113f0c4aa754ece296699bb24ce/sympy/functions/elementary/trigonometric.py#L576-L577

[...]

PR #21993, opened 2021-08-31
Treat UnevaluatedExpr specially in CodePrinter.doprint

bjodah:
https://github.com/sympy/sympy/issues/21991

When printing e.g. C or Fortran code for the an expression such as `abs(exp(x + (y + z)))`, (where `UnevaluatedExpr` is used to emit parenthesis around (y+z) for cases where the non-associativity floating point addition is an issue), the printer would emit `re(...)` nodes which aren't supported by the printer.

This PR replaces `UnevaluatedExpr` by instances which passes `is_real` queries through.

#### Release Notes

NO ENTRY


sympy/codegen/rewriting.py
```diff
@@ -294,6 +295,11 @@ def create_expand_pow_optimization(limit, *, base_req=lambda b: b.is_symbol):
     base_req : function returning bool
          Requirement on base for expansion to happen, default is to return
          the ``is_symbol`` attribute of the base.
+    UnevaluatedClass : subclass of UnevaluatedExpr
+         By default UnevaluatedExpr is used to prevent automatic simplification
+         of the expanded powers. For certain purposes, you may want to use a
+         subclass thereof instead (which e.g. propagates .is_real by overloading
```
oscarbenjamin, line 301: This seems like an unnecessarily general solution to a very specific problem that should perhaps be solved in a different way.

Why is it necessary to control the assumptions like this?

oscarbenjamin: Maybe the same effect could be better achieved by a specific rewrite.

bjodah: What prompted me to do this was to avoid `re(...)` being emitted by the printer. This was the smallest patch I could come up with, but of course it would be nice with something less intrusive.

A rewrite pass which removes `re(...)` if it's argument can be shown to be real is one way, the rewriting pass then needs to manually "jump" into the arg of UnevaluatedExpr instanceses, does it not?

oscarbenjamin: Maybe or what causes the `re` in the first place? Maybe the printers should be fixed or something.

bjodah: The printer cannot determine that the argument is real, perhaps best illustrated with this example:
```python
>>> r, s, t = symbols('r s t', real=True)
>>> ccode(abs(exp(r+s+t)))
'exp(r + s + t)'
>>> ccode(abs(exp(r+UnevaluatedExpr(s+t))))
'// Not supported in C:\n// re\nexp(r + re(s + t))'
```

oscarbenjamin: So maybe the printer should look inside an UnevaluatedExpr

bjodah: Hmmm, yes, that's probably the better fix. Will see what I can come up with.

bjodah:
OK, so I added a pre-processing step in the CodePrinter instead, what do you think of this approach?

oscarbenjamin:
I don't know this part of the codebase so well. This seems better then the previous version though.

I keep thinking there must be a better way than needing to invoke a new class but I can't think what it is.

sympy/printing/codeprinter.py
```diff
@@ -67,6 +69,15 @@ def __init__(self, settings=None):
...
             self.reserved_words = set()
 
+    def _handle_UnevaluatedExpr(self, expr):
+        if expr is None:
+            return expr
+        elif isinstance(expr, Basic):
+            return expr.replace(re, lambda arg: arg if isinstance(
+                arg, UnevaluatedExpr) and arg.args[0].is_real else re(arg))
+        else:
+            return type(expr)(self._handle_UnevaluatedExpr(item) for item in expr)
```
oscarbenjamin, line 79: Why does this need to handle non-Basic?

bjodah:
Tests for lambdify passes those through lambdarepr printer (which is
questionable in my opinion, but that's where we are at).

sympy/printing/codeprinter.py
bjodah: I just pushed a commit which moves this non-Basic handling into lambdify's internal class.

oscarbenjamin:
Is the List change related to the other changes?

Also are you finished with this or are you still working on it?

bjodah:
@oscarbenjamin, this is good to go from my point of view (barring my latest merge of master causes tests to fail).

Yes, the List change is related: in order for e.g. `_handle_UnevaluatedExpr` to be able to call `.replace`, we cannot have e.g. python-list instances at that point. But for example the tests of the OctavePrinter (in test_octave.py) explicitly exercises printing of python lists. Therefore, I introduced `List` and also added what I hope are some sensible tests of such a class. I noted that `sympify` by default changes python tuple into Tuple instances, but leaves list instances be. By having a "code-printer specific List-class" (`sympy.codegen.pynodes.List`), we can continue to support this behavior without touching non-code-printing related code.

So the design at this point is quite influenced by the constraints set by our existing test suite. Perhaps there's a more elegant solution, but right now, this is the best I've got...

oscarbenjamin:
Okay, great. I've set to auto-merge. Thanks

PR #21994, opened 2021-08-31
Print with parenthesis around UnevaluatedExpr in CodePrinter

### LINKED ISSUE #21955: Printing of UnevaluatedExpr needs parenthesis around Add terms
Floating point "add" is non-associative, i.e. `(x+y)+z` often differs from `x+(y+z)`. In SymPy `Add` (re)orders its arguments (for good reasons), but for representing e.g. an algorithm in finite precision this creates a problem (see e.g. [Kahan summation on wikipedia](https://en.wikipedia.org/wiki/Kahan_summation_algorithm)).

Typically we use `UnevaluatedExpr` for situations like this:
```python
>>> from sympy import UnevaluatedExpr as ue
>>> from sympy.abc import k,l,m
>>> ue(k+m)+l
l + k + m
>>> ue(k+m)+l + k
k + l + k + m
```
but, as you see the printer is not emitting the required parenthesis to make this useful for representing non-associative add.
 
Note that SymEngine gets this right:
```python
>>> from symengine import Symbol as S, UnevaluatedExpr as UE
>>> x,y,z = map(S, 'xyz')
>>> UE(y + x) + z
z + (x + y)
```

I propose that we fix this in the printers (I suspect updating precedence for `UnevaluatedExpr` might be enough, I haven't tried yet though). At the very least, the code printers should output parenthesis, but I think it's reasonable for the `StrPrinter` to do so as well?

bjodah:
Fixes #21955

Floating point add is not associative, it would be nice if we can use UnevaluatedExpr to represent this in the subclasses of CodePrinter.

#### Release Notes

NO ENTRY

### DIFF
--- a/doc/src/tutorial/manipulation.rst
+++ b/doc/src/tutorial/manipulation.rst
@@ -546,7 +546,7 @@ expression. Combine both of the methods to prevent both inside and outside
 evaluations:
 
     >>> UnevaluatedExpr(sympify("x + x", evaluate=False)) + y
-    y + x + x
+    y + (x + x)
 
 ``UnevalutedExpr`` is supported by SymPy printers and can be used to print the
 result in different output forms. For example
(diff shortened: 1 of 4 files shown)


Upabjojr:
It's a bit off topic, but I'm starting to think that the precedence module should have subclasses fitted for the various printers.

PR #21995, opened 2021-09-01
Fix check whether _sage_ method has been replaced by Sage

mkoeppe:

For SymPy classes for which Sage does not provide a `_sage_` method, code added in #21958 leads to:
```
sage: import sympy.physics
....: from sympy.physics.units import meter
sage: meter._sage_()
~/s/sage/sympy/sympy/core/basic.py in _sage_(self)
   1869         else:
   1870             # call the freshly monkey-patched method
-> 1871             return self._sage_()
   1872 
   1873 

RecursionError: maximum recursion depth exceeded
```

Fixed in this PR so it correctly says:
```
NotImplementedError: conversion to SageMath is not implemented
```

Sorry @oscarbenjamin for not noticing this earlier.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -1864,7 +1864,7 @@ class Basic(Printable, metaclass=ManagedProperties):
         old_method = self._sage_
         from sage.interfaces.sympy import sympy_init
         sympy_init()  # may monkey-patch _sage_ method into self's class or superclasses
-        if old_method is self._sage_:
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
> ```
> NotImplementedError: conversion to SageMath is not implemented
> ```

No problem. Thanks for the fix.

I'm glad we can get all this sorted before the sympy 1.9 release.

PR #22001, opened 2021-09-01
Docs: Add Explanation Category

nanjekyejoannah:
This PR adds docs for the ``explanation`` category.

It builds on top of   #21894 

The docs in directory `special_topics` are moved to `explanation/special_topics`, `gotchs.rst` is also moved to `explanation/gotchs.rst`.

#### Other comments

Opened to replace #21893

#### Release Notes

NO ENTRY


doc/src/explanation/gotchas.rst
```diff
@@ -7,7 +7,7 @@ Gotchas and Pitfalls
 .. role:: input(strong)
 
 Introduction
-============
+--------------
```
asmeurer, line 10: This doesn't actually do anything, because of the way rst works. We should revert the header changes in this file.

PR #20885, opened 2021-01-31
Fixed evalf(chop=True) for log

Maelstrom6:
#### References to other Issues or PRs

Fixes #18516
Fixes #19623

#### Brief description of what is fixed or changed
* `log(x)` can now be evaluated to a float when `x` is chopped to 0. Please advise on how I can improve this fix. What should I return when I want to return an mpf tuple representing `-oo`?
* Added a docstring to the "private" function `evalf`.

#### Other comments
`evalf_log` assumed the output when evaluating `x` is always an mpf tuple. This is not always the case: `x` can be an array of `None`. 

I know that `evalf` can return Nones but please can someone explain why that was the design choice? I've been trying to figure out why for a few days. Why not always have it return an mpf tuple (and return the 0 mpf tuple when it would have returned Nones)? This would allow for far fewer if statements.

#### Release Notes

NO ENTRY

PR #22022, opened 2021-09-05
Fixed #22020 by checking for oo in numerical equals

### LINKED ISSUE #22020: Comparing two operations that contain `log` sometimes leads to TypeError exception
This originally comes from PrairieLearn/PrairieLearn#4793, but it seems the problem is in sympy. We weren't able to narrow it down much further, other than to identify that this is specific to `log` (other functions like square root don't seem to have this problem).

In essence, two expressions can lead to an exception when being compared if the `log` expression is multiplied by a variable instead of a fixed coefficient.

```
Python 3.8.5 (default, May 27 2021, 13:30:53)
[GCC 9.3.0] on linux
>>> from sympy.parsing.sympy_parser import parse_expr
>>> x = parse_expr("log((2*V/3-V)/C)/-(R+r)*C")
>>> y = parse_expr("log((2*V/3-V)/C)/-(R+r)*2")
>>> x.equals(y)
  File "/home/jonatan/.local/lib/python3.8/site-packages/sympy/core/expr.py", line 751, in equals
    constant = diff.is_constant(simplify=False, failing_number=True)
  File "/home/jonatan/.local/lib/python3.8/site-packages/sympy/core/expr.py", line 674, in is_constant
    if b is not None and b is not S.NaN and b.equals(a) is False:
  File "/home/jonatan/.local/lib/python3.8/site-packages/sympy/core/expr.py", line 741, in equals
    diff = factor_terms(simplify(self - other), radical=True)
  File "/home/jonatan/.local/lib/python3.8/site-packages/sympy/simplify/simplify.py", line 635, in simplify
    _e = cancel(expr)
  File "/home/jonatan/.local/lib/python3.8/site-packages/sympy/polys/polytools.py", line 6688, in cancel
    f = factor_terms(f, radical=True)
  File "/home/jonatan/.local/lib/python3.8/site-packages/sympy/core/exprtools.py", line 1274, in factor_terms
    return do(expr)
  File "/home/jonatan/.local/lib/python3.8/site-packages/sympy/core/exprtools.py", line 1271, in do
    rv = _keep_coeff(cont, p, clear=clear, sign=sign)
  File "/home/jonatan/.local/lib/python3.8/site-packages/sympy/core/mul.py", line 2069, in _keep_coeff
    if r == int(r):
  File "/home/jonatan/.local/lib/python3.8/site-packages/sympy/core/expr.py", line 332, in __int__
    raise TypeError("can't convert %s to int" % r)
TypeError: can't convert oo to int
>>>
```

oscargus:

Closes #22020

It seems like `evalf` now returns `oo`  instead of throwing `ZeroDivisionError` for some cases. This breaks equals resulting in an error that `oo` cannot be turned into an `int`.

#### Release Notes

* core
    * Fixed issue when performing `is_constant()` on certain expressions which may result in division by zero.

### DIFF
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -666,7 +666,8 @@ class Expr(Basic, EvalfMixin):
                     a = expr._random(None, 0, 0, 0, 0)
             except ZeroDivisionError:
                 a = None
-            if a is not None and a not in (S.NaN, S.Infinity, S.NegativeInfinity):
+            from sympy.polys.polyutils import illegal
+            if a is not None and a not in illegal:
                 try:
                     b = expr.subs(list(zip(free, [1]*len(free))),
                         simultaneous=True)
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -666,7 +666,7 @@ class Expr(Basic, EvalfMixin):
                     a = expr._random(None, 0, 0, 0, 0)
             except ZeroDivisionError:
                 a = None
-            if a is not None and a is not S.NaN:
+            if a is not None and a not in (S.NaN, S.Infinity, S.NegativeInfinity):
                 try:
                     b = expr.subs(list(zip(free, [1]*len(free))),
                         simultaneous=True)
--- a/sympy/core/tests/test_expr.py
+++ b/sympy/core/tests/test_expr.py
@@ -2135,6 +2135,13 @@ def test_non_string_equality():
     assert (x != BadRepr()) is True
 
 
+def test_issue_22020():
+    from sympy.parsing.sympy_parser import parse_expr
+    x = parse_expr("log((2*V/3-V)/C)/-(R+r)*C")
+    y = parse_expr("log((2*V/3-V)/C)/-(R+r)*2")
+    assert x.equals(y) is None
+
+
 def test_21494():
     from sympy.testing.pytest import warns_deprecated_sympy
     with warns_deprecated_sympy():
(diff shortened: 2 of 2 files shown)


oscarbenjamin:
> It seems like `evalf` now returns `oo` instead of throwing `ZeroDivisionError` for some cases.

May it is `evalf` that should be fixed... The problem originates from #20885

oscargus:
> May it is `evalf` that should be fixed...

I do not disagree with it. However, there *may* be cases where we actually want to return `oo` from `evalf`, while this should take care of those cases as well (as I assume that just because two expressions both evaluate to `oo` does not make them equal and that `oo` may show up from `evalf` in cases that does not involve a division by zero).

So either wait for a "proper" fix or go with this that will not break anything (as far as I can tell, maybe one should check for `S.ComplexInfinity` as well though).

sympy/core/expr.py
```diff
@@ -666,7 +666,7 @@ def check_denominator_zeros(expression):
                     a = expr._random(None, 0, 0, 0, 0)
             except ZeroDivisionError:
                 a = None
-            if a is not None and a is not S.NaN:
+            if a is not None and a not in (S.NaN, S.Infinity, S.NegativeInfinity):
```
smichr, line 669: from sympy.polys.polyutils import illegal
...
if a is not None and a not in illegals:

oscargus: Weirdly enough adding `S.ComplexInfinity`, i.e., using `illegal` (should be in a better location, right? In core somewhere?) leads to that the following test fails:

```
    x = Symbol('x', positive=True)
    e5 = (1 + x)/x
    assert e5.is_constant() is False
```
I realize by checking the blame that I added it a while back, but do not really remember the circumstances. It is stated in the documentation that it is tested for denominators being zero. However, in this case returning `None` should be more correct than `False` since `x` cannot be zero...

oscargus:
Found a better way to fix it, so I close this (as it is done in a completely different way).

PR #22024, opened 2021-09-05
Fixed issue with simplifying oo + imaginary

### LINKED ISSUE #22020: Comparing two operations that contain `log` sometimes leads to TypeError exception

oscargus:

Closes #22020

Simplifying expressions of the type `oo + I/2` used to throw an error. Not anymore.

#### Other comments
To be honest, I do not fully understand the logic behind the code I added, but it seems to work and avoids the `int(c)` that was the cause of all evil here.

#### Release Notes

* simplify
     * Fixed issue with simplifying infinity plus rational imaginary.

### DIFF
--- a/sympy/core/mul.py
+++ b/sympy/core/mul.py
@@ -2112,7 +2112,7 @@ def _keep_coeff(coeff, factors, clear=True, sign=False):
         if not clear and coeff.is_Rational and coeff.q != 1:
             args = [i.as_coeff_Mul() for i in factors.args]
             args = [(_keep_coeff(c, coeff), m) for c, m in args]
-            if any(not c.is_infinite and c == int(c) for c, _ in args):
+            if any(c.is_Integer for c, _ in args):
                 return Add._from_args([Mul._from_args(
                     i[1:] if i[0] == 1 else i) for i in args])
         return Mul(coeff, factors, evaluate=False)
--- a/sympy/core/mul.py
+++ b/sympy/core/mul.py
@@ -2112,7 +2112,7 @@ def _keep_coeff(coeff, factors, clear=True, sign=False):
         if not clear and coeff.is_Rational and coeff.q != 1:
             args = [i.as_coeff_Mul() for i in factors.args]
             args = [(_keep_coeff(c, coeff), m) for c, m in args]
-            if any(c == int(c) for c, _ in args):
+            if any(not c.is_infinite and c == int(c) for c, _ in args):
                 return Add._from_args([Mul._from_args(
                     i[1:] if i[0] == 1 else i) for i in args])
         return Mul(coeff, factors, evaluate=False)
--- a/sympy/core/tests/test_expr.py
+++ b/sympy/core/tests/test_expr.py
@@ -2113,6 +2113,13 @@ def test_ExprBuilder():
     assert eb.build() == x**2
 
 
+def test_issue_22020():
+    from sympy.parsing.sympy_parser import parse_expr
+    x = parse_expr("log((2*V/3-V)/C)/-(R+r)*C")
+    y = parse_expr("log((2*V/3-V)/C)/-(R+r)*2")
+    assert x.equals(y) is False
+
+
 def test_non_string_equality():
     # Expressions should not compare equal to strings
     x = symbols('x')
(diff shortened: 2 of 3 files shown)


sympy/core/mul.py
```diff
@@ -2112,7 +2112,7 @@ def _keep_coeff(coeff, factors, clear=True, sign=False):
         if not clear and coeff.is_Rational and coeff.q != 1:
             args = [i.as_coeff_Mul() for i in factors.args]
             args = [(_keep_coeff(c, coeff), m) for c, m in args]
-            if any(c == int(c) for c, _ in args):
+            if any(not c.is_infinite and c == int(c) for c, _ in args):
```
jksuom, line 2115: The condition should probably be something like `c.is_Number` as `int(c)` is not implemented for finite complex numbers:
```
In [23]: I.is_infinite                                                          
Out[23]: False

In [24]: int(I)                                                                 

TypeError: can't convert complex to int

In [25]: I.is_Number                                                            
Out[25]: False
```

oscargus: Ahh, I only noted the `oo` error, but that clearly makes sense. (For some reason the code doesn't come here for e.g. `oo + I`, only for `oo+I/2` etc, and apparently not for many other cases with complex numbers...)

oscargus: However, as `oo` is a Number, I am not sure exactly what it should be. Right now it works and it seems like `c` rarely (if ever) an imaginary/complex number.

Also, `c` should by sympified as far as I can tell, so not obvious what the purpose of `c == int(c)` is? Check that `c` is an integer or something else? Maybe `is_Integer` is enough?

oscargus: I tried `is_Integer` only and so far so good. Pushing to run the complete test set.

smichr: > Also, `c` should by sympified as far as I can tell, so not obvious what the purpose of `c == int(c)` is

Yes, `c` was generated from `as_coeff_Mul` so it will be sympified. Even I am not sure why the `int` was used.

Issue #20068, opened 2020-09-10
Use flake8-bugbear on Travis and fix warnings


oscarbenjamin:
See e.g. https://github.com/sympy/sympy/pull/20061#discussion_r485996905

Currently flake8 is used on most files under the sympy folder. Mostly it warns about unused variables and imports in code and also non-existent variables which can help to pick up on typos and helps to keep the imports up to date. This was recently extended to docstrings (#19374, #19406).

It would be good to pick up on some other things though such as the use of mutable default arguments. There is flake-mutable but it seems that it basically doesn't work e.g. https://github.com/ebeweber/flake8-mutable/issues/21. The recommendation there is to use flake8-bugbear instead.

I've listed all the warnings that show up when using flake8-bugbear below. If we agree that they should all be fixed then I think that maybe this can be marked as easy to fix although it's not as simple as #19374. Once the warnings are fixed we could enable the use of flake8-bugbear on Travis with:
```diff
diff --git a/.travis.yml b/.travis.yml
index b21eef3c3a..a578f95460 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -214,7 +214,7 @@ before_install:
       coverage debug sys;
     fi
   - if [[ "${TEST_FLAKE8}" == "true" ]]; then
-      pip install flake8;
+      pip install flake8 flake8-bugbear;
     fi
   - if [ "$TRAVIS_PYTHON_VERSION" == "3.8" ]; then
         pip uninstall -y Cython;
```

We can test this with
```console
$ pip install flake8-bugbear
```
The warnings from flake8-bugbear need to be enabled in setup.cfg with this diff:
```diff
diff --git a/setup.cfg b/setup.cfg
index 112c744bb4..f86e990e45 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,7 +5,7 @@ long_description_content_type = text/markdown
 [flake8]
 doctests = True
 ignore = F403
-select = F,E722
+select = F,E722,B
 exclude = sympy/parsing/latex/_antlr/*,
           sympy/parsing/autolev/_antlr/*,
           sympy/parsing/autolev/test-examples/*,
```
Then you can run
```console
$ flake8 sympy
```
and you will see a list of warnings. We can count the warning types with:
```
$ flake8 sympy > bugbear.txt
```
And then this Python script:
```python
from itertools import chain
import re
from collections import Counter
from pprint import pprint

with open('bugbear.txt') as fin:
[...]

Issue #19372, opened 2020-05-19
don't use mutable defaults for functions


smichr:
A vexing problem was encountered while working on #19261 which was the result of using a mutable default for `repl_dict` in the `matches` routines. Such mutable defaults are used when setting arguments to `=[]` and `={}` -- there are about 70 cases of each of these.

These can be fixed by using tuple or None (and then internally providing [] or {} if the value is None).

Perhaps a code-quality test should added that checks for the use of such mutable objects?

asmeurer:
We can add https://pypi.org/project/flake8-mutable/ to our flake8 tests. Or else we can write our own check in the code quality check. I agree they should be avoided.

- `=[]` can often be replaced with `=()` without any loss of functionality.
- The only "valid" use-case I've seen of this is using mutable defaults for a function cache or memo. But it's not strictly necessary (the cache could just as well be defined above the function definition), and in most cases in SymPy we should be using the @cacheit or @memoize decorators.
- We should always make sure that the true default is documented in the parameters section of the docstring.

oscarbenjamin:
> Or else we can write our own check

Better to make use of existing tools.

PR #22025, opened 2021-09-05
Minor cleanups in core

oscargus:
Fixed some minor code issues (duplicate imports, unused variables, etc)

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -101,7 +101,7 @@ class Expr(Basic, EvalfMixin):
             else:
                 expr, exp = expr.args
         else:
-            expr, exp = expr, S.One
+            exp = S.One
 
         if expr.is_Dummy:
             args = (expr.sort_key(),)
@@ -867,7 +867,6 @@ class Expr(Basic, EvalfMixin):
             return False
 
     def _eval_is_extended_positive_negative(self, positive):
-        from sympy.core.numbers import pure_complex
         from sympy.polys.numberfields import minimal_polynomial
         from sympy.polys.polyerrors import NotAlgebraic
         if self.is_number:
@@ -1193,8 +1192,6 @@ class Expr(Basic, EvalfMixin):
 
     def as_terms(self):
         """Transform an expression to a list of terms. """
-        from .add import Add
-        from .mul import Mul
         from .exprtools import decompose_power
 
         gens, terms = set(), []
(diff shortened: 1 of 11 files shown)


sympy/core/expr.py
```diff
@@ -3619,7 +3616,7 @@ def _expand_hint_key(hint):
             use_hint = hints[hint]
             if use_hint:
                 hint = '_eval_expand_' + hint
-                expr, hit = Expr._expand_hint(expr, hint, deep=deep, **hints)
+                expr = Expr._expand_hint(expr, hint, deep=deep, **hints)[0]
```
oscarbenjamin, line 3619: I consider this kind of thing to be defensive programming. If the function is always expected to return a sequence of length 2 then this asserts that it has exactly two items:
```python
>>> a, b = [1, 2, 3]
ValueError: too many values to unpack (expected 2)
```
The variable name can also be used to help understand what the function returns even if the `hit` variable is unused.

sympy/core/mul.py
```diff
@@ -1113,7 +1113,7 @@ def _matches_noncomm(nodes, targets, repl_dict={}):
     def _matches_add_wildcard(dictionary, state):
         node_ind, target_ind = state
         if node_ind in dictionary:
-            begin, end = dictionary[node_ind]
+            begin = dictionary[node_ind][0]
```
oscarbenjamin, line 1116: I wouldn't change this.

sympy/core/expr.py
oscargus: I was also considering this and, e.g., for the mp-function returning unused accuracy variables I did not change it. I see the point to some extent here as well. (And next comment.)

sympy/core/mul.py
smichr: But here and above you could use `_` for the unused arg.

smichr, line 1116:
```suggestion
            begin, _ = dictionary[node_ind]
```

sympy/core/expr.py
smichr, line 3619:
```suggestion
                expr, _ = Expr._expand_hint(expr, hint, deep=deep, **hints)
```

sympy/core/mul.py
oscarbenjamin: I prefer `end`: it shows that the dictionary value is a 2-tuple marking the beginning and end which is helpful information when reading the code.

oscargus:
Thanks @smichr . I changed them in my local copy (back to using a name, since it may make sense here) and also removed some mutable default arguments as I found #19372 and #20068. Hence, I disabled auto-merge.

oscargus:
There is a bit of `repl_dict.copy()` here which I do not fullt know are required or if they were just a way to avoid mutating the default argument. I've kept them though.

oscargus:
@smichr You are right! I had an idea at some stage, but then I automatically replaced all `if ...` with `if ... else .copy` and then I forgot this. Thanks!

PR #22028, opened 2021-09-05
Further dropping Python 2

oscargus:
Removed imports from compatibility, either because they are not needed or that they should be imported directly from the source now that there is no Python 2 support.

However, I did not remove them from compatibility as there may be other software relying on them.

#### Release Notes

NO ENTRY


smichr:
> either because they are not needed or that they should be imported directly from the source

My only concern is that if they are located in `compatiblity` then it makes more sense to me to control the import location there so it's centralized, otherwise you have to edit all the direct imports in other files.

oscargus:
To me it seems like the long term plan should be to get id of compatibility. Now, that may be more or less simple for two reasons:

1) Some of the things there are functions that are actually non-trivial, not just some simple redefinition.
2) People may use import from there in their own projects built upon an earlier version of SymPy.

In compatibility there are the following lines:
https://github.com/sympy/sympy/blob/c4528bfd76bc762cf0203dbe4b82635629b1ead4/sympy/core/compatibility.py#L83-L84

Out of those six imports, only some odd Callable imports actually still imports from compatibility. Some things, like unicode, cStringIO, u_decode etc are not used anywhere in the code base as far as I can tell.

Ideally, we should deprecate all imports from compatibility, at least those that are not proper functions, so that people have time to change their projects, and then remove these. There is no point jumping to compatibility if you can import directly from the source.

sympy/functions/combinatorial/factorials.py
```diff
@@ -165,7 +165,7 @@ def eval(cls, n):
 
                     # GMPY factorial is faster, use it when available
                     elif HAS_GMPY:
-                        from sympy.core.compatibility import gmpy
+                        from sympy.external.gmpy import gmpy
                         result = gmpy.fac(n)
```
oscarbenjamin, line 169: Ideally anything `HAS_GMPY` logic should be handled in the `gmpy` compatibility module. For example there could just be
```
from sympy.external.gmpy import factorial
```
at the top of this file.

oscarbenjamin:
The compatibility module is used for different purposes:

1. Incompatible changes in Python e.g. py2 vs py3.
2. Vendoring backported Python features not available in oldest supported Python versions.
3. Handling gmpy2 being installed or not.

For 1 and 2 the feature in the compatibility module should only exist as long as needed. Once all supported Python versions are consistent then the feature should be removed from the compatibility module. Maybe a deprecation is needed but certainly sympy itself should just use the Python stdlib directly. These compatibility functions are only temporarily needed and should not be expected to remain there indefinitely.

For 3 I think that it is better to handle everything to do with gmpy2 in one place. Ideally the only module that checks `HAS_GMPY` would be the `sympy.external.gmpy` module itself. If there is a need for an alternative implementation of something that gmpy2 provides then that should be implemented in `sympy.external` just like `PythonMPQ`. The rest of the codebase should just do e.g. `from sympy.external.gmpy import factorial` and not need to worry about `gmpy` being installed or not.

If the compatibility module is not clearly documented as being internal to sympy then it should be. Downstream users/libraries should not use the functions from there although I know that there are some that do.

sympy/functions/combinatorial/factorials.py
oscargus: I see the point and that could work here and for bincoeff, but e.g. primetest would require quite a bit of rewrite in a not obvious way (or one could refactor, call the `gmpy` compatibility module, which call primetest again.

I'd prefer to open an issue for these things as there are probably more than those I have touched upon here. I assume that `HAS_GMPY` really should never be used in the code base?

oscarbenjamin: I would prefer for all of the `HAS_GMPY` logic to be as contained as possible but yes that can be a separate issue/PR.

> I assume that `HAS_GMPY` really should never be used in the code base?

Ideally but there doesn't need to be a hard rule. It's just better if you can see in one place what gmpy is doing and reason about what happens when installing it.

I would like to add support for using python_flint instead of gmpy2 because it's faster for many things and also does much more that is useful for sympy:
https://fredrikj.net/python-flint/

Example:
```
In [8]: import flint

In [9]: p = Poly(random_poly(x, 200, 0, 2))

In [10]: coeffs = [int(c) for c in p.all_coeffs()[::-1]]

In [11]: pf = flint.fmpq_poly(coeffs)

In [12]: %time ok = (pf ** 2).factor()
CPU times: user 45.3 ms, sys: 899 µs, total: 46.2 ms
Wall time: 48.5 ms

In [13]: %time ok = (p ** 2).factor_list()
CPU times: user 8.33 s, sys: 93 ms, total: 8.43 s
Wall time: 8.55 s
```
It will be a lot easier to control making use of external dependencies if all the code for doing it is as concentrated as possible.

Also we can avoid duplication and make better use of external libraries better by e.g. using `MPQ` internally as part of Rational.

smichr:
I am ok with whatever you guys decide.

oscargus:
OK! That is sort of how I thought of it, maybe except for the downstream aspects. 

If I understand correctly, then there are a number of exports from compatibility that can be removed as well. I can add them to this PR later on (tonight).

asmeurer:
`compatibility` really should be treated as internal and not used by external code. I don't know if we've been clear about that, though.

sympy/core/compatibility.py
```diff
@@ -6,131 +6,16 @@
...
-
-Metaclasses:
-    * Use `with_metaclass()`, examples below
-        * Define class `Foo` with metaclass `Meta`, and no parent:
-            class Foo(with_metaclass(Meta)):
-                pass
-        * Define class `Foo` with metaclass `Meta` and parent class `Bar`:
-            class Foo(with_metaclass(Meta, Bar)):
-                pass
+Some utility functios origniating from Python 2 and Python 3 compatible imports.
```
oscargus, line 11: Disastrous spelling....

sympy/core/compatibility.py
```diff
@@ -6,131 +6,16 @@
...
 __all__ = [
-    'PY3', 'int_info', 'SYMPY_INTS', 'clock',
-    'unicode', 'u_decode', 'get_function_code', 'gmpy',
-    'get_function_globals', 'get_function_name', 'builtins', 'reduce',
-    'StringIO', 'cStringIO', 'exec_', 'Mapping', 'Callable',
-    'MutableMapping', 'MutableSet', 'Iterable', 'Hashable', 'unwrap',
-    'accumulate', 'with_metaclass', 'NotIterable', 'iterable', 'is_sequence',
-    'as_int', 'default_sort_key', 'ordered', 'GROUND_TYPES', 'HAS_GMPY',
+    'NotIterable', 'iterable', 'is_sequence',
+    'as_int', 'default_sort_key', 'ordered',
```
oscargus, line 16: https://github.com/sympy/sympy/blob/414fc3df3090fe67afd6c8312e4ebf202124d308/sympy/utilities/iterables.py#L11-L13

Adding `NotIterable` to that list and moving them there?

oscargus:
Maybe we should just state in the 1.9 Release Notes that `compatibility` will be removed in the next release? In that way we can at least give the users a bit of a heads-up. The problem will be the six(?) remaining functions here that, as far as I can tell will still be around in `utilities.iterables`. 

Should I create a separate PR where these are moved to `utilities.iterables` (including `NotIterable`) and the import from `compatibility` is deprecated? Then the note that `compatibility` will be removed can go along with that deprecation notice (and that it was never meant for downstream use).

oscargus:
Or rather, maybe, update this to deprecate the remaining functions, keep the import-only stuff and add a comment in the release notes. No point in not merging the correct import location part of this PR.

oscarbenjamin:
> Should I create a separate PR where these are moved to `utilities.iterables` (including `NotIterable`) and the import from `compatibility` is deprecated?

I have often wondered utilities.iterables is itself in the right place for the things that it contains. It creates a lot of circular import problems because the core uses things from iterables but utilities also includes lambdify so importing anything from there brings the whole kitchen sink. Ideally there would be a clear separation between those modules on which the core depends and those that depend on the core. I think potentially having something like a `sympy.compat` module for things that can be freely used anywhere else without even needing to import the core would perhaps be better than either `sympy.core.compatibility` or `sympy.utilities.iterables`.

oscargus:
Or should one have a `sympy.core.iterables` for selected functions (and classes)? To me it seems like the only remaining functions here are not so much about compatibility.

For `sympy.utilities.iterables`there is the `import sympy` line that is only used to get `sympy.Id` in a single function. Most things are imported explicitly in the functions when needed. So clearly one can improve `sympy.utilities.iterables` a bit when it comes to imports without actually moving stuff. Maybe that is a first step. (It also imports from `sympy.utilities.enumerative` but that doesn't have any top-level imports at all.)

oscarbenjamin:
This has merge conflicts now

PR #20507, opened 2020-11-29
add checking if modules is an empty collection in lambdify

NChechulin:
#### References to other Issues or PRs
Fixes #17737 

#### Brief description of what is fixed or changed
Added checking is `modules` is a collection with length 0.
Now `lambdify([x, y], expr, modules=[])` behaves same way as `lambdify([x, y], expr)`

#### Release Notes

* utilities
  * fix issue with incorrect behaviour of `Mod` when `modules = []`

PR #22032, opened 2021-09-06
Add checking if modules is an empty collection in lambdify

oscargus:

Fixes #17737
Closes #20507 (couldn't continue there since it was built on master)

#### Release Notes

* utilities
  * Fixed issue with incorrect lamdifying behaviour of `Mod` when `modules = []`.

### DIFF
--- a/sympy/utilities/lambdify.py
+++ b/sympy/utilities/lambdify.py
@@ -754,7 +754,7 @@ def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
     from sympy.core.symbol import Symbol
 
     # If the user hasn't specified any modules, use what is available.
-    if modules is None:
+    if not modules:
         try:
             _import("scipy")
         except ImportError:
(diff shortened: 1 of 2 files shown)


oscargus:
Hmm, that test passes locally...

bjodah:
Superficially this looks like the wrong fix. `modules=[]` should mean exactly that: no modules, `modules=None` should mean "default", and due to historic(?) reasons that means "import numpy if you can", which turns mod into a function call (so that operator precedence does not need to enter the picture). I think the precedence weights are either wrong for mod, or we are missing a `self.parenthesize(...)` call in a `_print_XXX` method...

oscargus:
OK! Just thought that it seemed to fix it, so worth rebasing. But clearly something went wrong...

I'll close this in favor of your fix! (One should maybe fix the actual Python printing as well, maybe by simply adding brackets around a modulo-operation.)

bjodah:
It's a bit unclear to me, exactly what are you referring to with "actual Python printing"? Are there other printers than subclasses of `CodePrinter` that also fail to parentesize unary minus and modulo operator correctly? (I wouldn't be surprised if that's the case, I just haven't found any)

oscargus:
Basically, test `pycode(-Mod(x, y))` explicitly (the `python` printer uses `Mod` so that is not an issue, the `ccode` printer sprinkles brackets, most other printers do not seem to support it).

bjodah:
Ah, I see. I added some tests in d11dd0c . Was it something like that, that you had in mind?

oscargus:
Yes, exactly! Thanks!

PR #22037, opened 2021-09-06
Replaced isinstance(..., (int, Integer)) with is_Integer

oscargus:

Based on a discussion with @oscarbenjamin earlier, we saw a case where an already sympified object was tested for integrality with `isinstance(x, (int, Integer))`. This can hide bugs, such as the objects not being sympified, so itis better to test using `x.is_Integer`. Found a few of those (and some that cannot be changed, e.g. in `qubit.py`).

#### Release Notes

NO ENTRY


oscarbenjamin:
Looks good to me

PR #22039, opened 2021-09-06
maint: add version_tuple instead of LooseVersion

### LINKED ISSUE #22026: DeprecationWarning under Python 3.10 when importing sympy
Running under Python 3.10 importing SymPy gives a deprecation warning:
```
>>> import sympy
/Users/enojb/current/sympy/sympy/sympy/external/importtools.py:4: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  from distutils.version import LooseVersion
```
If this is not fixed then when Python 3.10 is released the latest versions of sympy and python will show this warning every time anyone imports sympy.

This is because sympy uses disutils which is now deprecated:
https://www.python.org/dev/peps/pep-0632/

According to the PEP the replacement for distutils.version is packaging. Since this is used at runtime using packaging would mean having a runtime dependency.

There are several uses of LooseVersion in the codebase:
```console
$ git grep LooseVersion
bin/authors_update.py:from distutils.version import LooseVersion
bin/authors_update.py:if LooseVersion(git_ver) < LooseVersion(minimal):
bin/mailmap_update.py:from distutils.version import LooseVersion
bin/mailmap_update.py:if LooseVersion(git_ver) < LooseVersion(minimal):
setup.py:    from distutils.version import LooseVersion
setup.py:        if mpmath.__version__ < LooseVersion(min_mpmath_version):
sympy/conftest.py:from distutils.version import LooseVersion as V
sympy/external/importtools.py:from distutils.version import LooseVersion
sympy/external/importtools.py:        if LooseVersion(modversion) < LooseVersion(min_module_version):
sympy/external/tests/test_numpy.py:from distutils.version import LooseVersion
sympy/external/tests/test_numpy.py:    if LooseVersion(numpy.__version__) >= LooseVersion('1.17'):
sympy/interactive/printing.py:from distutils.version import LooseVersion as V
sympy/interactive/session.py:from distutils.version import LooseVersion as V
sympy/printing/tensorflow.py:from distutils.version import LooseVersion as V
```
The most important case is `sympy/external/importtools.py` which uses `LooseVersion` as part of `import_module` which is itself very widely used around the codebase. The reason for using `LooseVersion` is to compare the versions of dependencies like numpy to check that the minimum supported version is installed.

As I see it there are two options:

1. Add a runtime dependency on packaging and use the version handling tools it provides:

    https://packaging.pypa.io/en/latest/version.html

2. Don't bother checking versions at runtime in `import_module`. It is the job of installers like pip and conda to make sure that the user has a valid combination of versions. SymPy should just make sure to provide accurate version requirements in setup.py.

I would lean towards option 2. I also doubt that many of the version requirements are actually needed anyway.

oscarbenjamin:
Previously sympy used the LooseVersion class from distutils.versions. As
of Python 3.10 distutils is deprecated and shows a warning on import.
This commit extracts the basic algorithm of LooseVersion into a simple
function version_tuple to be used for comparing versions.

Fixes #22026 on the 1.9 branch

#### Release Notes

- external
    - No longer use deprecated distutils. Previously a deprecation warning was seen when importing sympy on Python 3.10.

### DIFF
--- a/sympy/external/importtools.py
+++ b/sympy/external/importtools.py
@@ -1,7 +1,7 @@
 """Tools to assist importing optional external modules."""
 
 import sys
-from distutils.version import LooseVersion
+import re
 
 # Override these in the module to change the default warning behavior.
 # For example, you might set both to False before running the tests so that
@@ -26,6 +26,23 @@ def __sympy_debug():
 if __sympy_debug():
     WARN_OLD_VERSION = True
     WARN_NOT_INSTALLED = True
+
+
+_component_re = re.compile(r'(\d+ | [a-z]+ | \.)', re.VERBOSE)
+
+def version_tuple(vstring):
+    # Parse a version string to a tuple e.g. '1.2' -> (1, 2)
+    # Simplified from distutils.version.LooseVersion which was deprecated in
+    # Python 3.10.
+    components = []
+    for x in _component_re.split(vstring):
+        if x and x != '.':
+            try:
+                x = int(x)
+            except ValueError:
+                pass
+            components.append(x)
+    return components
 
 
 def import_module(module, min_module_version=None, min_python_version=None,
@@ -155,7 +172,7 @@ def import_module(module, min_module_version=None, min_python_version=None,
         modversion = getattr(mod, module_version_attr)
         if module_version_attr_call_args is not None:
             modversion = modversion(*module_version_attr_call_args)
-        if LooseVersion(modversion) < LooseVersion(min_module_version):
+        if version_tuple(modversion) < version_tuple(min_module_version):
             if warn_old_version:
                 # Attempt to create a pretty string version of the version
                 if isinstance(min_module_version, str):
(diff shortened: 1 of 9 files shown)


moorepants:
This looks acceptable to me.

sympy/external/importtools.py
```diff
@@ -28,6 +28,23 @@ def __sympy_debug():
...
+    # Python 3.10.
+    components = []
+    for x in _component_re.split(vstring):
+        if x and x != '.':
+            try:
+                x = int(x)
+            except ValueError:
+                pass
+            components.append(x)
+    return components
```
smichr, line 45: `return tuple(components)`?

oscarbenjamin: Yeah, good point. It doesn't matter if it's only used in version comparison but I suppose in principle it could be used in something like a set. Mainly it just means the name is misleading.

PR #22045, opened 2021-09-07
remove domain arg from invert_real

smichr:
This private function does not need the domain
argument. All the domain handling happens at a
higher level. This function just assures the caller
that the result will be real.

Also, expressions like `x + zoo` were not filtered from FiniteSet when intersected with Reals
but since the only possible values are nan and zoo (when x is infinity or finite, respectively) and
neither of those are currently  in Reals then the sum can be excluded, too.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/sets/sets.py
+++ b/sympy/sets/sets.py
@@ -1085,7 +1085,9 @@ class Interval(Set):
 
     def _contains(self, other):
         if (not isinstance(other, Expr) or other is S.NaN
-            or other.is_real is False):
+            or other.is_real is False or other.has(S.ComplexInfinity)):
+                # if an expression has zoo it will be zoo or nan
+                # and neither of those is real
                 return false
 
         if self.start is S.NegativeInfinity and self.end is S.Infinity:
--- a/sympy/sets/tests/test_fancysets.py
+++ b/sympy/sets/tests/test_fancysets.py
@@ -584,6 +584,9 @@ def test_Reals():
     assert S.Reals != Interval(0, oo)
     assert S.Reals.is_subset(Interval(-oo, oo))
     assert S.Reals.intersect(Range(-oo, oo)) == Range(-oo, oo)
+    assert S.ComplexInfinity not in S.Reals
+    assert S.NaN not in S.Reals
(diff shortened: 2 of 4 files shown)


oscargus:
I see the point and know that restricted intervals are not really used anyway. However, assuming that they can be used properly, would one like to propagate the interval to the `invert`-functions? Or put another way, would one ever like to send the domain to `_invert_real`? I can somehow see the point of that, but it is not that hard to add the `domain` argument again then. If not, LGTM!

smichr:
> assuming that they can be used properly, would one like to propagate the interval to the `invert`-functions

From what I saw, it looks like the real domain restriction is being used to decide how to invert things like powers. The clipping of the real result is made in the calling program. Since it's a private function, the arg (as you say) can be added in if it later looks like that is the right way to do things. But my hunch is that it won't -- real/complex is the only distinction that matters at the invert level.

smichr:
@oscargus, you might try your solveset changes after rebasing on this if the `x + zoo` handling looks ok to you.

smichr:
All the tests you added in #21925 still pass without the zoo-handling addition:
```diff
diff --git a/sympy/solvers/solveset.py b/sympy/solvers/solveset.py
index 21b460d7d5..856f17ec5e 100644
--- a/sympy/solvers/solveset.py
+++ b/sympy/solvers/solveset.py
@@ -197,11 +197,6 @@ def invert_real(f_x, y, x):
 def _invert_real(f, g_ys, symbol):
     """Helper function for _invert."""

-    # Some inverse functions evaluate to zoo, remove if present
-    fs = FiniteSet(S.ComplexInfinity)
-    if fs.is_subset(g_ys):
-        g_ys -= fs
-
     if f == symbol:
         return (f, g_ys)

```
I will leave this up to see if anyone else has comments, otherwise will commit in 24 h.

oscargus:
Great! Commit at will and I will update #21925 by removing those lines.

smichr:
> Commit at will and I will update #21925 by removing those lines.

@oscargus , this is committed now so your PR can be updated and merged.

PR #17443, opened 2019-08-18
[WIP] Refactoring and improvement of Piecewise simplify

oscargus:
#### References to other Issues or PRs

Closes #15929 
Closes #17407
Fixes #17283 

#### Brief description of what is fixed or changed
Refactored the simplification of `Piecewise` to have separate methods for separate steps. Will have to think a bit to figure out how to combine them...

#### Other comments
Still a bit of work needed, especially to make some of the code more pythonic, and add more comments. Right now, just trying to see which other tests may fail...

#### Release Notes

* core 
   * `.is_constant` is more robust.
   * Matching of relational expressions with division is more robust.
* logic
   * Better simplification of relationals with `Abs`.
* functions
   * Better simplification of `Piecewise`.

Issue #17354, opened 2019-08-06
Error when simplifying Max-relation


oscargus:
When investigating #13552 I found the following (maybe not related to the original issue, but this is a minimal example of failing simplification of the generated `Piecewise`).
```
In [27]: Max(y/(x - 1), y/(x + 1)) >= 0

In [28]: _.simplify()
Traceback (most recent call last):

  File "<ipython-input-28-20cad44b6496>", line 1, in <module>
    _.simplify()

  File "/nobackup/data/sympy/sympy/core/expr.py", line 3546, in simplify
    return simplify(self, **kwargs)

  File "/nobackup/data/sympy/sympy/simplify/simplify.py", line 552, in simplify
    return _eval_simplify(**kwargs)

  File "/nobackup/data/sympy/sympy/core/relational.py", line 292, in _eval_simplify
    elif dif.equals(0):  # XXX this is expensive

  File "/nobackup/data/sympy/sympy/core/expr.py", line 759, in equals
    constant = diff.is_constant(simplify=False, failing_number=True)

  File "/nobackup/data/sympy/sympy/core/expr.py", line 683, in is_constant
    simultaneous=True)

  File "/nobackup/data/sympy/sympy/core/basic.py", line 998, in subs
    return rv.xreplace(reps)

  File "/nobackup/data/sympy/sympy/core/basic.py", line 1192, in xreplace
    value, _ = self._xreplace(rule)

  File "/nobackup/data/sympy/sympy/core/basic.py", line 1214, in _xreplace
    return self.func(*args), True

  File "/nobackup/data/sympy/sympy/functions/elementary/miscellaneous.py", line 349, in __new__
    args = frozenset(cls._new_args_filter(args))

  File "/nobackup/data/sympy/sympy/functions/elementary/miscellaneous.py", line 522, in _new_args_filter
    raise ValueError("The argument '%s' is not comparable." % arg)

ValueError: The argument 'zoo' is not comparable.
```

oscargus:
I think it comes from a division by zero returning `zoo`. If instead `oo` is returned the error goes away. Wouldn't it make sense to return `oo` if whatever is divided by 0 is real? Ping @smichr who added the `zoo` in #16770.

smichr:
> Wouldn't it make sense to return `oo` if whatever is divided by 0 is real? [as in `1/S(0)`]

ping @jksuom

oscargus:
[...]

PR #22047, opened 2021-09-07
Fixed issue with matching

oscargus:

Extracted a part of #17443 that doesn't have to wait for the rest of the code.
Solves parts of #17354 (the matching issue, not the original one).

`match` with some expressions now work without throwing an error.

#### Other comments
Yes, catching `ValueError` is bad, but that is what we have to live with for now.

#### Release Notes

* core
    * `match` no longer throws are exception for certain relationals.

### DIFF
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -1556,7 +1556,13 @@ class Basic(Printable, metaclass=ManagedProperties):
         for arg, other_arg in zip(self.args, expr.args):
             if arg == other_arg:
                 continue
-            d = arg.xreplace(d).matches(other_arg, d, old=old)
+            if arg.is_Relational:
+                try:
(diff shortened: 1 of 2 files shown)


sympy/core/basic.py
```diff
@@ -1556,7 +1556,13 @@ def matches(self, expr, repl_dict=None, old=False):
         for arg, other_arg in zip(self.args, expr.args):
             if arg == other_arg:
                 continue
-            d = arg.xreplace(d).matches(other_arg, d, old=old)
+            if arg.is_Relational:
+                try:
+                    d = arg.xreplace(d).matches(other_arg, d, old=old)
+                except TypeError: # Should be InvalidCompatisonError when introduced
```
smichr, line 1562:
```suggestion
                except TypeError: # Should be InvalidComparisonError when introduced
```

PR #22052, opened 2021-09-08
fix(core): improve checks in pow substitutions

### LINKED ISSUE #18527: integrate((cos(x)/(4+(sin(x))**2)), manual=True) gives the wrong answer
Integrating with manual=True incorrectly adds an absolute value sign to the answer.
```
>>> x = symbols('x', real=True)
>>> expr = (cos(x)/(4+(sin(x))**2))
>>> integrate(expr, x, manual=True)
atan(Abs(sin(x))/2)/2
>>> integrate(expr, x)
atan(sin(x)/2)/2
```

### LINKED ISSUE #21671: Incorrect result from `manualintegrate` for definite integral
sympy vs sagecell error with triple integral?

I'm teaching a course in Multivariate Calculus and decided to convert my notes from Sage to Jupyter using SymPy. I have rewritten nearly all my notes as Jupyter Notebooks and am very impressed how I can use multiple cells like Mathematica and I can use MarkDown cells with LaTEX as well as all the great features of matplotlib, NumPy and SymPy.

I'm nearly done converting my segelets to python scripts on Colab and found a discrepency,

This Sage code:

`show(integral(integral(integral(1,z,x^2+y^2,2-x^2-y^2),y,-sqrt(1-x^2),sqrt(1-x^2)),x,-1,1))`

resolves as pi.

This SymPy code:

`display(Integral(1,(z,x**2+y**2,2-x**2-y**2),(y,-sqrt(1-x**2),sqrt(1-x**2)),(x,-1,1)).doit())`

resolves as -pi/2.

See #3 in this Jupyter Notebook:

[https://colab.research.google.com/drive/1OlT9nfPG8TzoR_WpDavx-SAa07HLg3hV?usp=sharing](https://colab.research.google.com/drive/1OlT9nfPG8TzoR_WpDavx-SAa07HLg3hV?usp=sharing)

Shouldn't these be equal? What am I missing? Any help would be greatly appreciated as I've done A LOT of work on this course using sympy and woud like to use it in class this summer session!

Please help,

A. Jorge Garcia

Applied Math & CS

Nassau Community College

http://shadowfaxrant.blogspot.com

PS: Here's the SAGECELL version,

[https://sagecell.sagemath.org/?z=eJzFk8FuwjAMhu9IvIMFB9qRTk1A07TrtE07cNuuSBkEGlGS4qRA-_RLQ1uQuDCGtlMsx_ns_5fTn-AbsJg-3scPoxg-J89sDB8os1TAu7JiiTw13U6fhvCy5WnOrQCu5v4OMxT2CWRdFpwHlJQkJiwkhTtGIdm7Yxx2OybRu6B3U2jPYbccg0FBykHT4seUdrYK52SzX0xIo-KArZvQo_DbYnsXyz1_e5A5unqe_ZQNiykjLHJR5KIKHJkN2oBWqZCcxFXviJ4a8deNL7XqOvrBzHEIr9JpsYmANTcG9MLHuZIWZvmXcNJ8YiHRWNAzy5WFXSIUYKJBGshQZxqt1IqnYLUvNpuco2hYc2ncq5ljoF77jEa5lKo19j-HaL_i6pKPuLoLareHZWWs39NmPf2yOmO_Adu6eJ4=&lang=sage](https://sagecell.sagemath.org/?z=eJzFk8FuwjAMhu9IvIMFB9qRTk1A07TrtE07cNuuSBkEGlGS4qRA-_RLQ1uQuDCGtlMsx_ns_5fTn-AbsJg-3scPoxg-J89sDB8os1TAu7JiiTw13U6fhvCy5WnOrQCu5v4OMxT2CWRdFpwHlJQkJiwkhTtGIdm7Yxx2OybRu6B3U2jPYbccg0FBykHT4seUdrYK52SzX0xIo-KArZvQo_DbYnsXyz1_e5A5unqe_ZQNiykjLHJR5KIKHJkN2oBWqZCcxFXviJ4a8deNL7XqOvrBzHEIr9JpsYmANTcG9MLHuZIWZvmXcNJ8YiHRWNAzy5WFXSIUYKJBGshQZxqt1IqnYLUvNpuco2hYc2ncq5ljoF77jEa5lKo19j-HaL_i6pKPuLoLareHZWWs39NmPf2yOmO_Adu6eJ4=&lang=sage)

### LINKED ISSUE #22033: A bug of integrate function
When we normalize the 2nd order Chebyshev polynomial x^2 - 1/4 of the 2nd kind by the L^2 norm on [-1, 1] with density function sqrt(1 - x^2)dt, we encounter a wrong result.
```
>>> from sympy import *
>>> from sympy import __version__
>>> __version__
'1.8'
>>> x = Symbol('x')
>>> f = (x**2 - Rational(1, 4))**2 * sqrt(1 - x**2); f
sqrt(1 - x**2)*(x**2 - 1/4)**2
>>> integrate(f, (x, -1, 1))
0
```
The norm is 0?.
The integrand of the above definite integral is strictly positive, so it is a bug.

If we divide the interval [-0, 1] into [-1, 0] and [0, 1], we have correct results as follows:
```
>>> integrate(f, (x, -1, 0))
pi/64
>>> integrate(f, (x, 0, 1))
pi/64
```
And, expanding the integrand, we also have the correct.
```
>>> g = f.expand(); g
x**4*sqrt(1 - x**2) - x**2*sqrt(1 - x**2)/2 + sqrt(1 - x**2)/16
>>> integrate(g, (x, -1, 1))
pi/32
```
This bug exists since version 1.5 of SymPy.
```
>>> from sympy import *
>>> from sympy import __version__
>>> __version__
'1.4'
>>> x = Symbol('x')
>>> f = (x**2 - Rational(1, 4))**2 * sqrt(1 - x**2); f
sqrt(1 - x**2)*(x**2 - 1/4)**2
>>> integrate(f, (x, -1, 1))
pi/32
```
This is correct.

oscarbenjamin:
In a substitution like (1/x).subs(x**2, y) there need to be checks to
see whether it is valid to replace x with e.g. sqrt(y). Previously those
checks were heuristic and depended on the evaluation behaviour of
_eval_power. This commit changes those to depend directly on the
assumptions of the base and exponent in the base pow.

Fixes #21671
Fixes #22033
Fixes #18527

#### Release Notes

- core
   - Handling of power substitutions like `(1/x).subs(x**2, y)` was made more robust. This indirectly fixed problems leading to incorrect evaluation of integrals.


sympy/core/power.py
```diff
@@ -832,9 +832,10 @@ def _check(ct1, ct2, old):
                         as_int(pow, strict=False)
                         combines = True
                     except ValueError:
-                        combines = isinstance(Pow._eval_power(
-                            Pow(*old.as_base_exp(), evaluate=False),
-                            pow), (Pow, exp, Symbol))
+                        b, e = old.as_base_exp()
+                        # These conditions ensure that (b**e)**f == b**(e*f) for any f
+                        combines = b.is_positive and e.is_real or b.is_nonnegative and e.is_nonnegative
```
smichr, line 837: `(b**e)**f` = b**(e*f)` also if `f.is_integer and (b.is_positive or e.is_integer)`. I think it was me that put this `combines` there and I wanted to make sure that we didn't hard-code in a test like this which would miss valid combinations or misapply invalid combinations.  (It took a long time to track down and remove those hard-coded applications down.) I don't know if @asmeurer note [here](https://github.com/sympy/sympy/pull/83#issuecomment-670411) has ever had a good reference point to consult.

smichr: I positively view the addition of a function that validates a combing of exponents in the power.py file -- one that doesn't require evaluation of a power (which, as you note, has brought troubles of its own).

oscarbenjamin: In general I think that substitutions like `x.subs(x**2, y)` are not a good idea for subs. It seems like subs is used for several different things that I would have made separate functions for e.g.:
```
(1 + x).subs(x, 1) # replace a symbol with a "value"
sin(exp(x)).subs(exp(x), y) # replace a subexpression with a different subexpression
(x**4).subs(x**2, y) # attempt a basic elimination of x from the equation
```
I think it would have been better if `subs` only did the first of these things: replace a symbol with a value. For the second I think something like replace is better. For the third case there should just be a different function called e.g. `eliminate`. The third case is very different from the others because it ends up replacing things that are not literally in the expression tree e.g. there is no `Pow(x, 2)` in the expression tree for `Pow(x, 4)`.

Given that I (personally) think that using subs for elimination is a mistake I'm not sure how comprehensive the checking here should be.

oscarbenjamin: My intention for this PR was for the change to be as minimal as possible though. I would like to apply this to the 1.9 release branch but I don't want to do that if it looks like it would potentially break things.

oscarbenjamin: Although there is a failure of the symengine job (#22054) it looks like all other tests have passed. If you can think of any examples that would fail under this PR then it would be good to add them to the tests and fix them here.

smichr: > not a good idea for subs

This has been my sentiment from old. We called these "algebraic substitutions". 

I can perhaps look more closely tonight or tomorrow.

oscargus:
Have you had the chance to try #18527 as well?

oscarbenjamin:
> Have you had the chance to try #18527 as well?

Yes, I've added a test for that now.

smichr:
> Previously those
> checks were heuristic and depended on the evaluation behaviour of
> _eval_power. This commit changes those to depend directly on the
> assumptions of the base and exponent in the base pow.

It is not accurate to say that the checks were heuristic. The check was directly dependent on the logic of `Pow._eval_power`. This modification captures only a subset of the rules encoded there. I have added comments to the discussion in #22033.

smichr:
It could be that the improvement is needed in the integration routines where substitution is being used as a means of performing the integral.

smichr:
Is is sufficient to limit power substitutions only when `pow >= 1`:
```python
combines = (pow - 1).is_nonnegative and isinstance(Pow._eval_power(
                            Pow(*old.as_base_exp(), evaluate=False),
                            pow), (Pow, exp, Symbol))
```
Then `(x**(S(3)/2)).subs(x**2, y)` will fail but `(x**(S(5)/2)).subs(x**2, y)` will succeed.

smichr:
> Is is sufficient to limit power substitutions only when `pow >= 1`:

or just when `pow.is_Integer`?

oscarbenjamin:
> > Previously those
> > checks were heuristic and depended on the evaluation behaviour of
> > _eval_power. This commit changes those to depend directly on the
> > assumptions of the base and exponent in the base pow.
> 
> It is not accurate to say that the checks were heuristic. The check was directly dependent on the logic of `Pow._eval_power`. This modification captures only a subset of the rules encoded there.

I think that it is potentially an improvement to handle only a subset of those rules.

Perhaps "heuristic" is the wrong word. I think checking the assumptions on the base and the exponent is a principled way of deciding whether the substitution can go ahead even if the method for evaluating those assumptions is heuristic itself.

The part that is flakey about checking `_eval_power` is checking the type of the expression node. In this case `Abs(x)` is not a `Pow` but `1/Abs(x)` is and in fact `1/anything` is. Potentially checking `_eval_power is not None` would make sense if the purpose was to know if evaluation would happen but I don't see how `isinstance(expr, Pow)` really tells us something meaningful.

smichr:
> really tells us something meaningful.

I think my intent was to check if the result evaluated to something like a power meaning that the assumptions allowed the exponents to combine. If they didn't you would get None. If they did in a less trivial way you could end up with a Mul. So ending with a single Pow/exp/Symbol was confirmation that the exponents were allowed to combine.

smichr:
I am not opposed to changing `subs`. I would like to see it, when using algebraic subs, give a result of "least surprise." And you identified the code path leading to the surprise. Consider the following in master:
```python
>>> x = Symbol('x', real=True)
>>> [(x**(i*S.Half)).subs(x**2,y) for i in range(2, 9)]
[x, y**(3/4), y, y**(5/4), x**3, y**(7/4), y**2]

vs

>>> x = Symbol('x', positive=1)
>>> [(x**(i*S.Half)).subs(x**2,y) for i in range(2, 9)]
[sqrt(y), y**(3/4), y, y**(5/4), y**(3/2), y**(7/4), y**2]
```

In the first set, given that `x**3` remains unchanged, I think it surprising that `x**(3/2)` is changed.

oscarbenjamin:
> I am not opposed to changing `subs`

Do you think it would be make this change on the 1.9 branch (or better to wait fo rnext release so it has more testing in master)?

I'm not sure about this odd evaluation rule:
```python
In [15]: Abs(x)**3
Out[15]: 
 2    
x ⋅│x│
```
The original `Abs(x)**3` seems better to me. For one thing making the symbol `x` appear twice in the expression significantly increases the complexity of some operations/algorithms (e.g. solving for `x`).

smichr:
> increases the complexity of some operations

I agree. 

As far as the change goes, I feel nervous about putting this in for 1.9 for these reasons:
1) this masks but does not fix the lack of splitting that is needed when doing a change of variables in `integrate`
2) this selects a subset of combining rules that will change the unrelated algebraic-subs behavior and I would rather see a more coherent change to favor the "least suprise" result. I don't think we should take a less careful change here to fix the irritating and harder to fix change in integration.

But having some basic integrals give correct answers is a pretty strong motivation to come up with a reasonable solution for `subs`. Does the requirement that the new exponent, `pow`, be an integer solve the issue without raising other significant issues?

oscarbenjamin:
Okay I've changed the milestone from 1.9 to 1.10.

smichr:
> In the first set, given that `x**3` remains unchanged, I think it surprising that `x**(3/2)` is changed.

Also, `x` remains unchanged because that doesn't pass through the `Pow._eval_power` routine. `Symbol._eval_power` would need to pass it off as `Pow(self, 1, evaluate=False).subs(old, new)` when old was a Pow.

oscargus:
Can you test the following code:
```
from sympy import *
from sympy.integrals.manualintegrate import manualintegrate
x = Symbol('x')
print(manualintegrate(log(log(x)) / x**2, x))
x = Symbol('x', extended_positive=True)
print(manualintegrate(log(log(x)) / x**2, x))
```

It seems like in current master the first one allows a substitution (or maybe matching) that the second one does not. Related to #21985 where the corresponding test breaks if `extended_positive` is automatically added. A few hours of debugging leads to that in the first case `EiRule` is matched if there are no assumptions, but not otherwise.

oscarbenjamin:
> Can you test the following code:

With master that gives:
```
Ei(-log(x)) - log(log(x))/x
Ei(log(x**(-2))/2) - log(-log(x**(-2)))/x + log(2)/x
```
This PR gives:
```
Ei(-log(x)) - log(log(x))/x
Ei(-log(x)) - log(log(x))/x
```

oscargus:
Thanks! I was hoping that it would be solved by this. (And it is.)

oscargus:
Should we merge this now or is there any remaining doubt? Seems like some other PRs would benefit from it.

oscarbenjamin:
I think this is good to go but @smichr seemed to have some doubts above.

oscargus:
I had the impression that it was primarily to "force" it into 1.9. Just thinking that the earlier it gets in, the longer the time to figure out surprising consequences of this (and other PRs that relies on this). But let us wait and see what @smichr thinks.

oscarbenjamin:
@smichr I think that this is a good change and I think that @oscargus is waiting on it. What do you think?

smichr:
Sorry for the delay -- this fell off my radar. Comparing `x.subs(x**2,y)` to `(1/x).subs(x**2,y)` is the clincher for me. If the former fails, the latter should not succeed. I would really like to see [the surprises here](https://github.com/sympy/sympy/pull/22052/#issuecomment-916225591) addressed at some time.

sympy/core/tests/test_subs.py
```diff
@@ -879,9 +882,17 @@ def test_issue_19558():
...
+
+
 def test_guard_against_indeterminate_evaluation():
     eq = x**y
     assert eq.subs([(x, 1), (y, oo)]) == 1  # because 1**y == 1
     assert eq.subs([(y, oo), (x, 1)]) is S.NaN
     assert eq.subs({x: 1, y: oo}) is S.NaN
-    assert eq.subs([(x, 1), (y, oo)],simultaneous=True) is S.NaN
+    assert eq.subs([(x, 1), (y, oo)], simultaneous=True) is S.NaN
+
```
smichr, line 898:
```suggestion
```

sympy/core/power.py
smichr: I think I see the problem: the conclusion was being based on the structure of the result (imagining that the result collapsed to a symbol or single exponential function or a power with combined exponent). But consider
```
>>> eq=x**Rational(3,2)
>>> eq**Rational(3,4)
(x**(3/2))**(3/4)

```
The last power does not represent successful combination of the exponents. So perhaps `_combines` should have been:
```python
combines = Pow._eval_power(
                            Pow(*old.as_base_exp(), evaluate=False),
                            pow).as_base_exp()[0] == old.as_base_exp()[0]
```

smichr:
please see [here](https://github.com/sympy/sympy/pull/22052#discussion_r753596335)

sympy/core/power.py
```diff
@@ -861,9 +861,10 @@ def _check(ct1, ct2, old):
...
```
smichr, line 866:
```suggestion
                        obase, oexp =old.as_base_exp()
                        p = Pow._eval_power(
                            Pow(obase, oexp, evaluate=False), pow)
                        if p is None:
                            combines = False
                        else:
                            pbase = p.as_base_exp()[0]
                            combines = pbase == obase

```

smichr: Note that this will now keep `sqrt(x).subs(x**2,y)` from changing but will allow `sqrt(p).subs(p**2,y)` to become `sqrt(sqrt(p))` if `p` is positive. So to disable algebraic subs, Integral need only use vanilla symbols.

smichr:
I will make the suggested mod after this is committed. So let's go with what you have now since this fixes something and the next mod won't break it (and will require another run of tests anyway.)

oscarbenjamin:
Thanks @smichr

PR #22064, opened 2021-09-10
Cleanup in utilities.iterables including no more usage of cartes

oscargus:

Related to #22028

Some general cleanup of `utilities.iterables` to improve the imports (not import sympy at top level).

As `cartes` is now supported as `itertools.product` it is not use in the code base anymore. However, I kept it here, but probably it should be deprecated before removal?

#### Other comments
Some of the imports can probably be made more "directed".

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/symbol.py
+++ b/sympy/core/symbol.py
@@ -9,12 +9,13 @@ from .function import FunctionClass
 from .kind import NumberKind, UndefinedKind
 from sympy.core.logic import fuzzy_bool
 from sympy.logic.boolalg import Boolean
-from sympy.utilities.iterables import cartes, sift
+from sympy.utilities.iterables import sift
 from sympy.core.containers import Tuple
 
 import string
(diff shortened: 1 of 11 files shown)


smichr:
+1/2. SymPy already uses `product` like `summation`, however:
```python
>>> help(product)
Help on function product in module sympy.concrete.products:

product(*args, **kwargs)
    Compute the product.
    
    The notation for symbols is similar to the notation used in Sum or
    Integral. product(f, (i, a, b)) computes the product of f with
    respect to i from a to b, i.e.,
```
In context, perhaps it's clear.

oscargus:
> SymPy already uses `product` like `summation`

Good point!

smichr:
Since `cartes` is still available there is no harm in using the library version in the codebase. If a user doesn't like the name clash in their own work they can use `cartes`.

PR #22069, opened 2021-09-10
Some minor fixes for (now non-slow) tests

oscargus:
Unmarked a few tests as not being slow.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/functions/special/tests/test_bessel.py
+++ b/sympy/functions/special/tests/test_bessel.py
@@ -13,7 +13,7 @@ from sympy.testing.randtest import (random_complex_number as randcplx,
                                       test_derivative_numerically as td,
                                       _randint)
 from sympy.simplify import besselsimp
-from sympy.testing.pytest import raises, slow
+from sympy.testing.pytest import raises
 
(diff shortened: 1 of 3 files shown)


oscarbenjamin:
I think we should have something that makes tests fail if they are too slow or if they are not slow when marked as slow so that:

1. The markers have to be kept up to date continuously.
2. We can pick up on speedups and slowdowns.

Sometimes a change is made that massively slows down a test but it isn't obvious to reviewers. If the test had to be edited to show something like `@slow('20 seconds')` then it would be possible for the reviewer to question that change.

oscargus:
> I think we should have something that makes tests fail if they are too slow

That would indeed be nice. Or just something that keeps track of the test times and posts significant changes.

Right now, some of the slow tests takes 0 seconds. However, as this is because they are disabled, I do not really see the point of changing them to non-slow...

It seems (after a quick googling) that either reporting all "slowest" tests and then figuring out a way to compare them or configuring pytest-benchmark in some way to report all tests (which I have not found a way to do) are the options.

PR #22072, opened 2021-09-11
Fixed syntax error

oscargus:
Found a typo...

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/functions/special/error_functions.py
+++ b/sympy/functions/special/error_functions.py
@@ -2146,7 +2146,7 @@ class Shi(TrigonometricIntegral):
         elif not arg0.is_infinite:
             return self.func(arg0)
         elif arg0.is_infinite:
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
Is this already fixed on master?

oscargus:
I think I changed it in another PR that should go to master, but I cannot seem to recall which one right now... (Can be one branch where there is no PR yet). Thought it was important to get it into 1.9 though so made a separate one.

oscarbenjamin:
It gets very confusing if a fix is made on the release branch but not on master. Effectively it is an immediate regression. I think it's better that something gets fixed on master and then the fix is "backported" to the release branch.

Can you open a PR for master as well so both can be merged at once?

oscargus:
Sure! Wasn't sure exactly how the different branches were handled. If one could "forward-port" or not. #22095

PR #22077, opened 2021-09-11
Partial fix for `matches` being unable to match

### LINKED ISSUE #21942: Algebraic re-arrangement of exponential equations
Sympy version: 1.7.1
Environment: Google Colab
Python version: 3.7

I am trying to use Sympy 1.7.1 in a google Colab environment running python 3.7 to automate some tedious algebra in macroeconomic models, and I'm running into an issue when trying to re-arrange equations of the form:
```
mc = (a * r ** (1 - p) + (1 - a) * w ** (1 - p)) ** (1 / (1 - p))
```
With parameters 0 < a < 1, and p > 0, p != 1 given, and w, r > 0. I would like to be able to re-arrange this expression to obtain functions r(mc, w) or w(mc, r). In the case of r, my expected output is:
 ```
r = ((1 / a) * (mc ** (1 - p) - (1 - a) * w ** (1 - p))) ** (1 / (1 - p))
```

To solve this equation I am trying the following code:

```
import sympy as sp

a, p, mc, r, w = sp.symbols(['a', 'p', 'mc', 'r', 'w'], real=True, nonnegative=True)
eq = mc - (a * r ** (1 - p) + (1 - a) * w ** (1 - p)) ** (1 / (1 - p))
sp.solve(eq, r)
```

Which results in the following error and trace-back: 

```
<ipython-input-15-a95f68afce93> in <module>()
----> 1 sp.solve(eq, r)

/usr/local/lib/python3.7/dist-packages/sympy/solvers/solvers.py in _solve(f, *symbols, **flags)
   1458                 d = dgen.pop()
   1459                 w = Wild('g')
-> 1460                 gen = f_num.match(D.xreplace({d: w}))[w]
   1461                 spart = gen.as_independent(symbol)[1].as_base_exp()[0]
   1462                 if spart == symbol:

TypeError: 'NoneType' object is not subscriptable
```

When i define a new parameter q = (1 - p), however, so that:
```
mc = (a * r ** q + (1 - a) * w ** q) ** (1 / q)
```

Sympy.solve finds expressions for r or w without any problem. Indeed, it seems only the outside exponent, (1 / (1 - p)) is important to change: the system is able to solve `mc = (a * r ** (1 - p) + (1 - a) * w ** (1 - p)) ** (1 / p)` without any problem. Evidently, that (1 - p) in the denominator of the exponent is causing the problem, but I don't know enough about how the solve function works to understand why that would be the case.

Are there best practices for writing functions of exponents, such as the one I have presented here, in order to facilitate the solve function? 

Thanks for any help in advance! This is my first time posting an issue on github, so I apologize if I did it wrong.

### PR DESCRIPTION
**Pull request by Maelstrom6**
Fixes #21942

A central idea about symbolic mathematics is pattern matching. 

A common question is something like: "what does `W` have to be so that `2*x*W` is the same as `2*x*exp(y)`"? In this case, `W` maps to `exp(y)`. `W` is a Wild while everything else is a symbol. I'm not sure what they are supposed to be called, but I will call `2*x*W` the "pattern" and I will call `2*x*exp(y)` the "expression". 

SymPy's pattern matching is very complex (and possibly overly convoluted compared to JuliaLang's SymbolicUtils.jl), but it does seem to work for the most part. However, there is a problem that #21942 picked up: in the process of matching, the expression mutates and information is lost. 

In the very first example above, the way SymPy goes about finding what `W` is, is by going step by step: it sees that the 2 is common and drops that. That leaves a new pattern of `x*W` and a new expression of `x*exp(y)`. It then applies the same process recursively. It drops the `x` and it is left with `W` and `exp(y)`. It is then obvious that it should take `W` to be `exp(y)`. 

Extending this further, let us consider the pattern to be `-W**(1/(1-p))` and the expression to be `-x**(1/(1-p))`. SymPy begins by dropping the -1 in the front of both expressions. It does this through a function called `Mul._combine_inverse` on the expression to get rid of the -1. However, in the last step, it performs a signsimp on the resulting expression. So after it has dropped the -1, it is left with a pattern of `W**(1/(1-p))` but a mutated expression of `x**(-1/(p-1))`. It cannot match this and it says that a match is not possible. 

What would be first prize is to not simplify the expression but instead just divide it straight to get `x**(1/(1-p))` instead. However, there was a unit test that explicitly checked that it would simplify. I fear that it may have longer reaching effects further down in the code base. This is probably the simplest and most correct way to go though.

However, for some reason, I chose to do it another way: to simplify the pattern as well. That way we get the pattern of `W**(-1/(p-1))` with a mutated expression of `x**(-1/(p-1))` which SymPy can then match. In the end SymPy says that `W` maps to `x`. This is fine. But if we started with a different expression that would simplify into the same one we had in the last step, then we would have a problem. 

Consider if we had a pattern of `-W**(1/(1-p))` but an expression of `-x**(-1/(p-1))`. Since these expressions are not exactly the same, it should not find a match. But this is not the case. The signsimp on the expression will do nothing but the signsimp on the pattern will result in `W**(-1/(p-1))` which means SymPy is then able to identify a match when none exists.

I'm open to suggestions. (I'm not sure why I didn't remove the simplification in the first place).

Edit:
It may have been because there is a `_combine_inverse` for every function such as `Add`, `Mul` and `Pow` and all of them perform a signsimp before returning. If we change all of them, there will be dozens of unit test failures. But the problem may extend to all of these classes and not just the `Mul` class. This is possibly why I chose a more universal but slightly incorrect fix.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/tests/test_match.py
+++ b/sympy/core/tests/test_match.py
@@ -710,15 +710,12 @@ def test_match_issue_17397():
 
 
 def test_match_issue_21942_1():
-    a, r, w = symbols('a, r, w', nonnegative=True)
+    a, x, r, w = symbols('a, x, r, w', nonnegative=True)
     p = symbols('p', positive=True)
     g_ = Wild('g')
-    pattern = g_ ** (1 / (1 - p))
-    eq = (a * r ** (1 - p) + w ** (1 - p) * (1 - a)) ** (1 / (1 - p))
-    assert pattern.matches(eq) == {g_: a * r ** (1 - p) + w ** (1 - p) * (1 - a)}
-
-    eq = signsimp(eq)
-    assert pattern.matches(eq) is None
+    pattern = x - g_ ** (1 / (1 - p))
+    eq = x - (a * r ** (1 - p) + w ** (1 - p) * (1 - a)) ** (1 / (1 - p))
+    assert pattern.matches(eq) == {g_: a * r ** (1 - p) - w ** (1 - p) * (a - 1)}
 
 
 @XFAIL
--- a/sympy/core/tests/test_match.py
+++ b/sympy/core/tests/test_match.py
@@ -1,5 +1,5 @@
 from sympy import (abc, Add, cos, collect, Derivative, diff, exp, Float, Function,
-    I, Integer, log, Mul, oo, Poly, Rational, S, sin, sqrt, Symbol, symbols,
+    I, Integer, log, Mul, oo, Poly, Rational, S, signsimp, sin, sqrt, Symbol, symbols,
     Wild, pi, meijerg, Sum
 )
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by smichr** [comment]
Wouldn't it be sufficient to do `signsimp` on the expression and pattern once at the start of the match routine?

**Comment by Maelstrom6** [comment]
I think we'd be left with a compromise in every solution. For example, if we signsimp everything at the start and we inputted the pattern `W` and the expression `1/(1-p)`, then `W` would map to `-1/(p-1)` instead.

If we signsimp the pattern only when `_combine_inverse` signsimps the expression, then in the above example, the matcher would return a value before a signsimp takes place. This would result in a more correct mapping to `1/(1-p)`. But then again, as notes before, for more complicated expressions, `W` might map to something that doesn't syntactically exist in the expression.

**Comment by smichr** [comment]
Perhaps signsimp should not be used at all but +/-W could be used?

**Comment by Maelstrom6** [comment]
I'm not sure I understand. Could you please explain further?

File: sympy/core/tests/test_match.py [hunk 1]
```diff
@@ -709,6 +709,37 @@ def test_match_issue_17397():
     assert r == {a3: x - 4 + 4/x, b3: 1 - 2/x, c3: x - 4}
 
 
+def test_match_issue_21942_1():
```
smichr, line 712: This test already passes in master. How has your change changed the behavior of match?

Maelstrom6: Sorry, I got a bit carried away by trying to make the XFAIL look like this one. I'll update it with a test that actually makes a difference.

Maelstrom6:
Maybe we should continue on your PR instead? It's a lot simpler, a lot more readable and I have no more spare time again.

PR #22078, opened 2021-09-11
Fixed incorrect answer when adding an identity matrix to a block matrix

### LINKED ISSUE #22034: Bug in sum between BlockMatrix and Matrix
The following simple operations provide unexpected results. Let us define
```python
import sympy as sym
n = 10
I = sym.Identity(n)
O = sym.ZeroMatrix(n, n)
I2_block = sym.BlockMatrix([[I, O], [O, I]])
I2 = sym.Identity(2*n)
```

Then, the line
```python
print(sym.block_collapse(I2_block - I2))
```
produces the following (clearly incorrect) result
```
Matrix([
[2*I,   0],
[  0, 2*I]]) - I
```

On the other hand, the line
```python
print(sym.block_collapse(I2_block + I2))
```
never terminates. In particular, in enters in an infinite loop at
https://github.com/sympy/sympy/blob/6ef14a954288079cde3d1bf304189f6b8f9296dc/sympy/strategies/core.py#L9

t-ens:
Fixes #22034 

Previously `bc_block_plus_ident(expr)` added any block identity matrix in `expr` to both the list of block matrices and the list of identity matrices so that when everything is summed together in the end any block identity matrix was added twice. I fixed this and added a test.

#### Other comments
As suggested in issue #22034 I am treating every block identity matrix  the same regardless of internal block structure. If there is at least one non-identity block matrices and all non-identity block matrices have the same internal block structure then every identity matrix is converted to a matrix with this same block structure as well. However if the only block matrices in `expr` are also identity matrices then I chose to forget the internal block structure and just treat it as a usual identity matrix rather than a block matrix so in this case the result of `bc_block_plus_ident(expr)` no longer contains any block matrices in it at all.  In the original code the function returned `expr` unchanged if any two block matrices in `expr` had a different internal block structure, even if only an identity matrix was different whereas now the block structure of the identity matrices is ignored. This change does not seem to cause any problems and the tests pass however if this is a problem or just not the behaviour `bc_block_plus_ident` was intended to have I will modify it.

#### Release Notes

* matrices
  * Fixed a bug when adding identity matrix to block matrix.

### DIFF
--- a/sympy/matrices/expressions/blockmatrix.py
+++ b/sympy/matrices/expressions/blockmatrix.py
@@ -747,17 +747,31 @@ def bc_matadd(expr):
         return block
 
 def bc_block_plus_ident(expr):
-    idents = [arg for arg in expr.args if arg.is_Identity]
+    idents = 0
+    block_ident_size = 0
+    blocks, rest = [], []
+    for arg in expr.args:
+        if isinstance(arg, BlockMatrix):
+            if arg.is_Identity:
+                idents += 1
+                block_ident_size = arg.shape[0]
+            else:
+                blocks.append(arg)
+        elif arg.is_Identity:
+            idents += 1
+        else:
+            rest.append(arg)
+
     if not idents:
         return expr
-
-    blocks = [arg for arg in expr.args if isinstance(arg, BlockMatrix)]
-    if (blocks and all(b.structurally_equal(blocks[0]) for b in blocks)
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
This PR needs a review

oscargus:
@naveensaigit do you have time to take a look at this? You seem to be the one that most recently made major contribution to the block matrix (although this particular part has been unchanged for quite some time).

PR #22079, opened 2021-09-12
21942: remove match from solve (1.9 branch)

### LINKED ISSUE #21942: Algebraic re-arrangement of exponential equations

smichr:

fixes #21942
#### Release Notes

* solvers
  * use of `match` was removed to make solving more reliable

### DIFF
--- a/sympy/core/tests/test_match.py
+++ b/sympy/core/tests/test_match.py
@@ -1,5 +1,5 @@
 from sympy import (abc, Add, cos, collect, Derivative, diff, exp, Float, Function,
-    I, Integer, log, Mul, oo, Poly, Rational, S, sin, sqrt, Symbol, symbols,
+    I, Integer, log, Mul, oo, Poly, Rational, S, signsimp, sin, sqrt, Symbol, symbols,
     Wild, pi, meijerg, Sum
 )
 
@@ -509,7 +509,7 @@ def test_issue_3883():
     a, b, c = symbols('a b c', cls=Wild, exclude=(gamma,))
 
     assert f.match(a * log(gamma) + b * gamma + c) == \
-        {a: Rational(-1, 2), b: -(mu - x)**2/2, c: log(2*pi)/2}
+        {a: Rational(-1, 2), b: -(-mu + x)**2/2, c: log(2*pi)/2}
     assert f.expand().collect(gamma).match(a * log(gamma) + b * gamma + c) == \
         {a: Rational(-1, 2), b: (-(x - mu)**2/2).expand(), c: (log(2*pi)/2).expand()}
     g1 = Wild('g1', exclude=[gamma])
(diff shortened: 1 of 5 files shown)


oscarbenjamin:
Is there also a PR for the master branch?

sympy/core/tests/test_match.py
```diff
@@ -647,7 +647,7 @@ def test_issue_4319():
     ans = {S.Zero, y - S.One/8}
 
     def ok(pat):
-        assert set(p.match(pat).values()) == ans
+        assert set(p.match(pat).values()) == ans, set(p.match(pat).values())
```
smichr, line 650:
```suggestion
        assert set(p.match(pat).values()) == ans
```
```suggestion
        assert set(p.match(pat).values()) == ans, set(p.match(pat).values())
```

PR #20650, opened 2020-12-22
Improve the set handler

sayandip18:
#### References to other Issues or PRs

Fixes #20535 

#### Brief description of what is fixed or changed
The set handler can now deal with symbolic intervals

#### Release Notes

NO ENTRY

PR #21196, opened 2021-03-29
Fix(solvers): ImageSet Union simplify (#18489 improved a bit)

MaqOwais:
**References to other Issues or PRs**
Closes #17838
Closes #11188
re-uses tests from #17838 / #17079 / #18489 (with attribution)

**Brief description of what is fixed or changed**
This PR adds a union handler for Integer-based ImageSets such as those often returned by solveset for trigonometric equations.

Before (i.e. with master):

```
In [1]: solveset(cos(x) + cos(3*x) + cos(5*x), x, S.Reals)
Out[1]:
⎧        π        ⎫   ⎧        3⋅π        ⎫   ⎧        4⋅π        ⎫   ⎧        2⋅π        ⎫   ⎧        5⋅π        ⎫   ⎧        π        ⎫
⎨2⋅n⋅π + ─ | n ∊ ℤ⎬ ∪ ⎨2⋅n⋅π + ─── | n ∊ ℤ⎬ ∪ ⎨2⋅n⋅π + ─── | n ∊ ℤ⎬ ∪ ⎨2⋅n⋅π + ─── | n ∊ ℤ⎬ ∪ ⎨2⋅n⋅π + ─── | n ∊ ℤ⎬ ∪ ⎨2⋅n⋅π + ─ | n ∊ ℤ⎬
⎩        2        ⎭   ⎩         2         ⎭   ⎩         3         ⎭   ⎩         3         ⎭   ⎩         3         ⎭   ⎩        3        ⎭

   ⎧        7⋅π        ⎫   ⎧        5⋅π        ⎫   ⎧        11⋅π        ⎫   ⎧        π        ⎫
 ∪ ⎨2⋅n⋅π + ─── | n ∊ ℤ⎬ ∪ ⎨2⋅n⋅π + ─── | n ∊ ℤ⎬ ∪ ⎨2⋅n⋅π + ──── | n ∊ ℤ⎬ ∪ ⎨2⋅n⋅π + ─ | n ∊ ℤ⎬
   ⎩         6         ⎭   ⎩         6         ⎭   ⎩         6          ⎭   ⎩        6        ⎭
```
After:

```
In [1]: solveset(cos(x) + cos(3*x) + cos(5*x), x, S.Reals)
Out[1]:
⎧      π        ⎫   ⎧n⋅π   π        ⎫   ⎧n⋅π   π        ⎫
[...]

PR #17330, opened 2019-08-02
Simplification of relations with min/max

oscargus:
#### References to other Issues or PRs

Fixes #13605

#### Brief description of what is fixed or changed
This PR introduces simplification of relations with `Min` or `Max` in them. For example, it can do:
```
(Min(x, y, z) >= z).simplify() => (z <= Min(x, y))
(Max(x, y, z) >= z).simplify() => S.true
(Eq(Min(x, y, z), z)).simplify() => (z <= Min(x, y))
 (Ne(Max(x, y, z), z)).simplify() => (z < Max(x, y))
```

I realized the lack of simplification when seeing:
```
a = Symbol('a', positive=True)
integrate(sqrt(1/(1 - x**2)), (x, 0, a))
```
which returns
```
⎧asin(Min(1, a))  for a ≤ Min(1, a)
⎨                                  
⎩      nan            otherwise    
```

Now, instead we get (after simplify):
```
⎧asin(Min(1, a))  for a ≤ 1
⎨                          
⎩      nan        otherwise
```

(Of course, the expression can also be simplified, but that is more the role of Piecewise simplification than Relational simplification.)

#### Other comments

The commented out code is related to #17331. If/when that gets sorted out, it should be possible to (more or less) just uncomment those rows.

One can think about querying for extended_real here. At least when returning `S.true` or `S.false`. Not sure if it is worth it or not. Using `Min` and `Max` (and `Relational`) more or less only makes sense for real values.

#### Release Notes

* functions
   * `expand_minmax` introduced to remove `Min/Max` from relations.
* logic
   * Faster `simplify_logic` when used with relations.
* simplify
    * Improved simplification of relations.

PR #22083, opened 2021-09-12
Better intersection of symbolic intervals

### LINKED ISSUE #20535: Error thrown by `integrate`
```python
from sympy import *
_z = Dummy('z')
a = symbols('a', positive=True)
x = Integral(exp(-_z), (_z, Intersection(Interval(0, oo), Interval(a, oo))))
print(x.doit())
```

_Originally posted by @Maelstrom6 in https://github.com/sympy/sympy/issues/20529#issuecomment-737711255_

oscargus:

Fixes #20535
Relates to #20650 but as this solves it in a slightly different way, only comparisons that evaluate are used here, so no `Min` or `Max`.

Evaluating Intersection of Intervals does some simple modifications of bounds when possible. The requirements are that no `Min` or `Max` are introduced. This helps some cases with symbolic intervals and will also tighten the intervals in some other cases.

#### Release Notes

* sets
   * Slightly better handling of `Intersection` of `Interval`s.

### DIFF
--- a/sympy/sets/handlers/intersection.py
+++ b/sympy/sets/handlers/intersection.py
@@ -404,7 +404,7 @@ def intersection_sets(a, b): # noqa:F811
             return b
 
 
-    # We can't intersect [0,3] with [x,6] -- we don't know if x>0 or x<0
+    # Check if any start and end point combinations can be compared
     comps = [is_le(a.start, b.start), is_le(a.start, b.end), is_le(a.end, b.start), is_le(a.end, b.end)]
     if all(c is None for c in comps):
         return None
@@ -445,12 +445,12 @@ def intersection_sets(a, b): # noqa:F811
             return Interval(start, end, left_open, right_open)
 
     if start is not None and not start_equal:
-        return Intersection(Interval(start, a.end, left_open, a.right_open),
-                            Interval(start, b.end, left_open, b.right_open))
+        return (Interval(start, a.end, left_open, a.right_open),
+                Interval(start, b.end, left_open, b.right_open))
 
     if end is not None and not end_equal:
-        return Intersection(Interval(a.start, end, a.left_open, right_open),
-                            Interval(b.start, end, b.left_open, right_open))
+        return (Interval(a.start, end, a.left_open, right_open),
+                Interval(b.start, end, b.left_open, right_open))
 
     return None
 
--- a/sympy/sets/handlers/intersection.py
+++ b/sympy/sets/handlers/intersection.py
@@ -1,5 +1,6 @@
 from sympy import (S, Dummy, Lambda, symbols, Interval, Intersection, Set,
                    EmptySet, FiniteSet, Union, ComplexRegion, Mul)
+from sympy.core.relational import is_le, is_lt, is_eq, is_gt
 from sympy.multipledispatch import dispatch
 from sympy.sets.conditionset import ConditionSet
 from sympy.sets.fancysets import (Integers, Naturals, Reals, Range,
@@ -402,43 +403,57 @@ def intersection_sets(a, b): # noqa:F811
         if l.is_real or l in infty or r.is_real or r in infty:
             return b
 
+
     # We can't intersect [0,3] with [x,6] -- we don't know if x>0 or x<0
-    if not a._is_comparable(b):
+    comps = [is_le(a.start, b.start), is_le(a.start, b.end), is_le(a.end, b.start), is_le(a.end, b.end)]
+    if all(c is None for c in comps):
         return None
 
-    empty = False
-
-    if a.start <= b.end and b.start <= a.end:
-        # Get topology right.
-        if a.start < b.start:
-            start = b.start
-            left_open = b.left_open
-        elif a.start > b.start:
-            start = a.start
-            left_open = a.left_open
-        else:
-            start = a.start
-            left_open = a.left_open or b.left_open
-
-        if a.end < b.end:
-            end = a.end
-            right_open = a.right_open
-        elif a.end > b.end:
-            end = b.end
-            right_open = b.right_open
-        else:
-            end = a.end
-            right_open = a.right_open or b.right_open
-
-        if end - start == 0 and (left_open or right_open):
-            empty = True
-    else:
-        empty = True
-
-    if empty:
+    if comps[1] is False and comps[2]:
         return S.EmptySet
 
-    return Interval(start, end, left_open, right_open)
+    start = None
+    start_equal = False
+    if is_lt(a.start, b.start):
+        start = b.start
+        left_open = b.left_open
+    elif is_gt(a.start, b.start):
+        start = a.start
+        left_open = a.left_open
+    elif is_eq(a.start, b.start):
+        start = a.start
+        left_open = a.left_open or b.left_open
+        start_equal = (a.left_open == b.left_open)
+
+    end = None
+    end_equal = False
+    if is_lt(a.end, b.end):
+        end = a.end
+        right_open = a.right_open
+    elif is_gt(a.end, b.end):
+        end = b.end
+        right_open = b.right_open
+    elif is_eq(a.end, b.end):
+        end = a.end
+        right_open = a.right_open or b.right_open
+        end_equal = (a.right_open == b.right_open)
+
+    if start is not None and end is not None:
+        if is_eq(start, end) and (left_open or right_open):
+            return S.EmptySet
+        else:
+            return Interval(start, end, left_open, right_open)
+
+    if start is not None and not start_equal:
+        return Intersection(Interval(start, a.end, left_open, a.right_open),
+                            Interval(start, b.end, left_open, b.right_open))
+
+    if end is not None and not end_equal:
+        return Intersection(Interval(a.start, end, a.left_open, right_open),
+                            Interval(b.start, end, b.left_open, right_open))
+
+    return None
+
 
 @dispatch(type(EmptySet), Set)  # type: ignore # noqa:F811
 def intersection_sets(a, b): # noqa:F811
--- a/sympy/sets/sets.py
+++ b/sympy/sets/sets.py
@@ -2457,7 +2457,10 @@ def simplify_intersection(args):
                 # with t. Returns the newly intersected set otherwise
 
                 if new_set is not None:
-                    new_args = (args - {s, t}).union({new_set})
+                    if isinstance(new_set, tuple):
+                        new_args = (args - {s, t}).union(new_set)
+                    else:
+                        new_args = (args - {s, t}).union({new_set})
                     break
             if new_args:
                 args = new_args
--- a/sympy/integrals/tests/test_integrals.py
+++ b/sympy/integrals/tests/test_integrals.py
@@ -6,7 +6,7 @@ from sympy import (
     Ne, O, oo, pi, Piecewise, polar_lift, Poly, polygamma, Rational, re, S, Si, sign,
     simplify, sin, sinc, SingularityFunction, sqrt, sstr, Sum, Symbol, summation,
     symbols, sympify, tan, trigsimp, Tuple, lerchphi, exp_polar, li, hyper,
-    Float
+    Float, Intersection
 )
 from sympy.core.expr import unchanged
 from sympy.functions.elementary.complexes import periodic_argument
@@ -1253,6 +1253,13 @@ def test_issue_4153():
         -12*log(3) - 3*log(6)/2 + 3*log(8)/2 + 5*log(2) + 7*log(4),
         6*log(2) + 8*log(4) - 27*log(3)/2, 22*log(2) - 27*log(3)/2,
         -12*log(3) - 3*log(6)/2 + 47*log(2)/2]
+
+
+def test_issue_20535():
+    _z = Symbol('z')
+    a = symbols('a', positive=True)
+    x = Integral(exp(-_z), (_z, Intersection(Interval(0, oo), Interval(a, oo))))
+    assert x.doit() == exp(-a)
 
 
 def test_issue_4326():
--- a/sympy/sets/tests/test_powerset.py
+++ b/sympy/sets/tests/test_powerset.py
@@ -5,7 +5,7 @@ from sympy.sets.contains import Contains
 from sympy.sets.fancysets import Interval
 from sympy.sets.powerset import PowerSet
 from sympy.sets.sets import FiniteSet
-from sympy.testing.pytest import raises, XFAIL
+from sympy.testing.pytest import raises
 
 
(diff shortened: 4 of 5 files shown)


sympy/sets/sets.py
```diff
@@ -2457,7 +2457,10 @@ def simplify_intersection(args):
                 # with t. Returns the newly intersected set otherwise
 
                 if new_set is not None:
-                    new_args = (args - {s, t}).union({new_set})
+                    if isinstance(new_set, tuple):
+                        new_args = (args - {s, t}).union(new_set)
```
smichr, line 2461: which `intersection_sets` is returning a tuple? That should maybe be fixed?

oscargus: The new one... If you have a better idea, I'd be happy to implement it. The thing is that with the new code one would like to return an `Intersection`/two `Interval`s for the case when one have rewritten one of the bounds (but did not manage to merge it into one).

oscarbenjamin:
The comment here might apply to these changes as well: https://github.com/sympy/sympy/pull/21196#issuecomment-814376676

Basically any code in the intersection handlers needs to very simple because the handlers are called in a pairwise `O(N^2)` routine.

It would probably be better to make a `simplify_intersection_of_intervals` that can do this more efficiently than pairwise handlers. Also it would be better if this happened as part of simplify rather than evaluation.

sympy/sets/handlers/intersection.py
```diff
@@ -402,43 +403,57 @@ def intersection_sets(a, b): # noqa:F811
...
-        empty = True
+    if start is not None and not start_equal:
+        return (Interval(start, a.end, left_open, a.right_open),
+                Interval(start, b.end, left_open, b.right_open))
 
-    if empty:
-        return S.EmptySet
+    if end is not None and not end_equal:
+        return (Interval(a.start, end, a.left_open, right_open),
+                Interval(b.start, end, b.left_open, right_open))
```
asmeurer, line 453: Couldn't these cases be represented with Min() and Max()?

asmeurer:
It seems like the algorithm here could be implemented more efficiently. Maybe the fuzzy logic makes this untrue, but it seems like a lot of the is_* function calls are redundant from the previous calls. Can we assume that is_ge(s.start, s.end) is not False, because otherwise the interval would have evaluated to a singleton finite set or the empty set? 

I think instead of checking for <, >, and = in each case, you can check just <= and >, being careful to order things so that if one interval is open and the other closed that it gives the right answer.

sympy/sets/handlers/intersection.py
asmeurer: Oh I guess the problem is that one is open and one is closed, the open/closedness of the final interval depends on which ends up being bigger. Perhaps this can be represented by a union of a ConditionSet that includes the left (right) endpoint iff it is bigger (smaller) and closed. 

In any case, couldn't this function just return a Set object (like Union) instead of a tuple here?

asmeurer:
You should check my math here, but it seems to me that given `[a, b]` and `[x, y]`, their intersection is always `[max(a, x), min(b, y)]`. This works even if `b < x`, `y < a`, or even `b < a` or `y < x`, where the intersection is empty (ditto for open (a, b) and (x, y) using <= instead of <). 

Here is some code that verifies the formula for all combinations of 4 different orderings of values (using the existing intersection code for explicit numeric values, which I'm assuming is correct).

```py
for a in range(4):
    for b in range(4):
        for x in range(4):
            for y in range(4):
                assert Interval(a, b).intersect(Interval(x, y)) == Interval(max(a, x), min(b, y))
```

If there are open ends to the intervals we may need to union some conditional endpoint, e.g., for `[a, b]` and `[x, y)` it would be `[max(a, x), min(b, y)) U {i ∈ {b} | i < y }`. I'm not sure if there's a cleaner way to represent this.

Therefore, we can represent the intersection completely symbolically without needing to check cases. The symbolic assumptions checking that is already the `Min`, `Max`, and `Interval` constructors should be sufficient.

oscargus:
Some quick notes.

There was an earlier version that wasn't part of the intersection-handler that was linear in time, but only handled closed intervals. I seem to have force-pushed that away though... It may be possible to make rhat handle open bounds as well. (Maybe I stashed it...)

The idea is to not end up with Min/Max expressions, but extract the known improvements to bounds, if any, and apply them. But indeed open/close would make it hard to get Min/Max correct anyway.

It may be that one can avoid checking all three combinations. At least one can reuse the earlier checks, leading to six in total. 

Once should also account for the probability to reducing N when looking at N**2 complexity. And what N typically is.

I doubt that #20535 can be solved with relying on simplify. As I understand it, this is something that comes from SymPy, and, hence, it would make sense to be able to handle it without manual manipulation. (There are quite few issues that can be closed if we just replied "Use simplify".)

asmeurer:
The point is that Min and Max themselves will automatically simplify to improved bounds if possible. And when they don't it's still a nicer symbolic result than an unevaluated Intersection. The mixed open/closed cases are still annoying, but I think it's better to fix them up by unioning or complementing some other set, both from an internal logic point of view and from the point of view of returning the simplest expression (its simpler in the sense that it's still a closed form of the intersection).

oscarbenjamin:
> There are quite few issues that can be closed if we just replied "Use simplify"

The point is not that we should use the `simplify` function but rather there should be specific functions analogous to `signsimp`, `powsimp`, `logcombine`, `expand`, `to_cnf` etc but for simplifying/reexpressing sets. Then those specific functions should be used judiciously in e.g. `solveset` and could perhaps be used as part of `simplify` so that end users can use e.g. `simplify(solveset(...))` even if that wouldn't be used in the codebase.

Some functions like `powsimp` and `expand` do opposite things which is nice because it allows using different representations of a symbolic expression. Automatic evaluation makes that impossible though because it forces expressions back into a particular form even after you try to convert them to another form. For example if all powers would automatically expand then the `(x*y)**n` form that `powsimp` creates would become impossible to represent.

Adding simplification routines as part of evaluation is a significant performance problem because it means that any code that uses those classes has no control over how costly any routine is. Even basic things like subs and xreplace need to rerun the evaluation code e.g.:
```python
In [5]: xs = symbols('x:100')

In [6]: ys = symbols('y:100')

In [8]: %time i = Intersection(*(Interval(x, y) for x, y in zip(xs, ys)))
CPU times: user 1.03 s, sys: 4 ms, total: 1.04 s
Wall time: 1.06 s

In [9]: %time ok = i.subs(xs[0], z)  
CPU times: user 656 ms, sys: 0 ns, total: 656 ms
Wall time: 654 ms
```
Currently sympy's set manipulation falls over as soon as you create a complicated set like an intersection of many sets.

The basic design where solveset depends on the automatic evaluation of sets is a mistake and each new thing added to evaluation of sets makes that worse. The problem is not necessarily that more evaluation leads to a bigger slowdown in itself but that it further increases the dependency of other code on automatic evaluation when we should be trying to reduce that dependency so that we can at least eliminate the slow parts.

oscargus:
> The point is that Min and Max themselves will automatically simplify to improved bounds if possible.

Indeed. That is why I used it for my first attempt. However, with some experience of rewriting relationals to Min/Max and regretting it as it is the long term resulted in issues elsewhere, I wasn't so keen to keep the Min/Max here (things like `x < Min(y, z) and x >= z` was hard to simplify to `x < y`). But it is indeed a very simple thing to do from a programming perspective (and in #17330 there is an `expand_minmax` that gets rewrites relationals with Min/Max for this reason) . Just loop over all the Intervals and add Min/Max of it (at least for the closed ones). Would also be easy to incorporate with Range etc.

oscargus:
> Adding simplification routines as part of evaluation

Where is the cut between evaluation and simplification? Many functions have quite complicated evaluation routines. Comparing to #22098, you were not that keen on the `doit`. To me it seems to make sense to actually use `doit` more so that `eval` only perform some basic evaluation and `doit` performs the more time-consuming ones. So `doit` on an `Intersection` would execute the stuff that is now done every time it is evaluated (the pair-wise quadratic stuff etc). Basically `evaluate=False` all over, except for trivial cases (for some definition of trivial). I think that is actually what most people would expect. If you write `1 +1` you would probably expect `2`, but if you write `sin(pi/12)` you may not always expect `sqrt(6)/4 - sqrt(2)/4` directly. On the other hand, you may expect `Integral(0, x)` to evaluate...

So I think we should carefully consider this. To what degree should automatic evaluation happen? Wouldn't it make sense to actually use the `doit` functionality to a sort of "function-based simplification", evaluating more advanced expressions (as happens with `Integral` et al).

Either way, if I get you correctly, you think that most (all?) of the intersection handlers should be removed from evaluation and moved to `setsimplify` (or just `doit` then, which would be much easier for people to remember and use and avoid the risk of infinite recursion in `simplify`). Or specifically, the quadratic part.

asmeurer:
> Indeed. That is why I used it for my first attempt. However, with some experience of rewriting relationals to Min/Max and regretting it as it is the long term resulted in issues elsewhere, I wasn't so keen to keep the Min/Max here (things like x < Min(y, z) and x >= z was hard to simplify to x < y). 

I think you meant a different example. If x < min(y, z) then x < z. 

If you instead had x <= min(y, z), you can use the fact that if a = min(b, c) then by definition a <= b and a <= c, so we can rewrite the above as `x <= min(y, z) <= y and x <= min(y, z) <= z and x >= z`, which gives `x = z <= y`.

In any case, I think it's a good idea to improve simplification of inequalities with min and max, because you can represent a lot of things symbolically using min and max. I would rather have robust inequality and min/max simplification that we can use implicitly than trying to effectively redo the same logic over and over again like is being done here.

> But it is indeed a very simple thing to do from a programming perspective (and in #17330 there is an expand_minmax that gets rewrites relationals with Min/Max for this reason) . Just loop over all the Intervals and add Min/Max of it (at least for the closed ones). Would also be easy to incorporate with Range etc.

I think it's not just simple from a programming perspective, but it also produces a very simple result for the end user in the general case (at least for the both open or both closed cases, I'm still not completely sure what the best result should be in the mixed cases).

> Where is the cut between evaluation and simplification? Many functions have quite complicated evaluation routines. Comparing to #22098, you were not that keen on the doit. To me it seems to make sense to actually use doit more so that eval only perform some basic evaluation and doit performs the more time-consuming ones. So doit on an Intersection would execute the stuff that is now done every time it is evaluated (the pair-wise quadratic stuff etc). Basically evaluate=False all over, except for trivial cases (for some definition of trivial). I think that is actually what most people would expect. If you write 1 +1 you would probably expect 2, but if you write sin(pi/12) you may not always expect sqrt(6)/4 - sqrt(2)/4 directly. On the other hand, you may expect Integral(0, x) to evaluate...

I think for sets we shouldn't do any automatic evaluation. That's how the matrix expressions work, and it works fine. For normal expressions it's hard because users expect certain evaluations like x + x, but I think for sets we can get away with no automatic evaluation. 

See also https://github.com/sympy/sympy/wiki/Automatic-Simplification. The problem with automatic evaluation is that it creeps in and causes problems, but ends up being hard to remove.

oscargus:
> I think you meant a different example. If x < min(y, z) then x < z.

Correct, sort of, I meant `Min(y, z) or x >= y` it turns out... With #17330, the relationals are replaced with symbols, so `x < y & x < z | x >= z` turns into `a & b | ~a` which is then simple to simplify to `b`.

> I think for sets we shouldn't do any automatic evaluation. 
So evaluate on `doit()`? Which seems to be the original idea.

It actually makes quite a bit of sense since there is `.intersect`, so `a. intersect(b)` would be `Intersection(a, b).doit()` as opposed to `Intersection(a, b)` which is left "unevaluated". (Or possibly `intersect(a, b)` evaluates, if there is a huge benefit of `a. intersect(b)` not doing so.) This makes it behave like `integrate`/`Integral`.

However, my main point is probably that `doit` should do "advanced" evaluation, on an function base, and `simplify_xx`, simplify things involving more than one function (although I realize that this is not a clear-cut line either). 

For Intersections (and other set methods), it would more or less mean moving the `eval` functionality to `doit`.

oscargus:
If we ignore the recent history, it seems like most functions should require `doit`. Compare it to other CAS where you often have to press shift+Enter to evaluate. Of course, basic arithmetic functions such as +-*/ could evaluate, but most others should probably not.

oscarbenjamin:
> That's how the matrix expressions work, and it works fine. For normal expressions it's hard because users expect certain evaluations like x + x

Actually with the matrix expressions x + x does evaluate and is equivalent to `MatAdd(x, x).doit()`:
```python
In [1]: x = MatrixSymbol('x', 2, 2)

In [2]: y = MatrixSymbol('y', 2, 2)

In [3]: x + x
Out[3]: 2⋅x

In [4]: MatAdd(x, x)
Out[4]: x + x

In [5]: MatAdd(x, x).doit()
Out[5]: 2⋅x
```
I'm not sure if I completely agree with that design but it is much better than having automatic evaluation in the constructor because it means that the func-args invariant is satisfied so the behaviour of e.g. substitutions is predictable:
```python
In [6]: MatAdd(x, x).subs(x, y)
Out[6]: y + y

In [7]: x, y = symbols('x, y')

In [8]: e = Add(x, x, evaluate=False)

In [9]: e
Out[9]: x + x

In [10]: e.subs(x, y)
Out[10]: 2⋅y
```
The last result here is a consequence of the func-args invariant not being satisfied when `evaluate=False` is used.

>  it seems like most functions should require `doit`

Having evaluation in doit rather than `__new__` is an improvement but actually I also think that we should limit what happens in `doit` as well. The fundamental problem with any kind of evaluation is the premise that there is a unique best way to represent something that is always desirable over any other equivalent form.

For example with `And/Or` (analogous to `Intersection/Union`) we might say that `doit` should always convert to CNF or something. Those experienced with logic though know that there are many different normal forms and that different ones should be used in different situations.

Really what is needed is not evaluation but controlled *manipulation*. Putting everything into `__new__/doit` is nice if it means that the calling code does not have to think about what manipulation it really wants. Actually though there is no substitute for the caller choosing what manipulation to use and what form they want the expression to be in and using e.g. `to_cnf/to_nnf` or `factor/expand` and so on.

A function like `simplify` can use heuristics to choose the best representation so that a user can call the function knowing that it is slow but expecting it to figure out the details for them. Otherwise we need well behaved manipulation but it isn't possible to have that when there is automatic evaluation. Putting the evaluation into `doit` is an improvement but it should still be limited carefully to well-defined behaviour there as well.

sympy/sets/sets.py
smichr: Ahh, it's `new_set` not `new_args` that is a tuple. I would be more worried about the former.

oscarbenjamin:
This PR needs a review

PR #22086, opened 2021-09-12
No errors on intersection of symbolic ranges, range + interval

oscargus:
Earlier it was not possible to perform an intersection with a symbolic range without throwing an exception.

#### Other comments
Nothing clever happens yet, but better to not throw an exception.

Also added some xfail-tests to illustrate what can happen if it is clever.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/sets/handlers/intersection.py
+++ b/sympy/sets/handlers/intersection.py
@@ -80,14 +80,15 @@ def intersection_sets(a, b): # noqa:F811
 
 @dispatch(Range, Interval)  # type: ignore # noqa:F811
 def intersection_sets(a, b): # noqa:F811
-    from sympy.functions.elementary.integers import floor, ceiling
-    if not all(i.is_number for i in b.args[:2]):
+    # Check that there are no symbolic arguments
+    if not all(i.is_number for i in a.args + b.args[:2]):
         return
 
     # In case of null Range, return an EmptySet.
     if a.size == 0:
         return S.EmptySet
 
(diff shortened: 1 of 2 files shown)


sympy/sets/handlers/intersection.py
```diff
@@ -80,14 +80,15 @@ def intersection_sets(a, b): # noqa:F811
 
 @dispatch(Range, Interval)  # type: ignore # noqa:F811
 def intersection_sets(a, b): # noqa:F811
-    from sympy.functions.elementary.integers import floor, ceiling
-    if not all(i.is_number for i in b.args[:2]):
+    # Check that there are no symbolic arguments
+    if not all(i.is_number for i in a.args + b.args[:2]):
```
smichr, line 84: Should this be `if not any(i.is_number for i in a.args + b.args[2:])`?

smichr: Maybe then the XFAIL would pass.

oscargus: The logic is that all the `Range` arguments and the two first `Interval` arguments should be numbers (the two last ones are booleans as I understand it). So rather than just testing that the `Interval` arguments are numbers, we also test that the `Range` arguments are numbers, as that caused the errors earlier. The tests fail as there is no code corresponding to #22083 that tests the start and end arguments separately and tries to make the most out of it. Yet.

At least based on how I interpreted the code.

smichr:
OK, better no error than error.

PR #22087, opened 2021-09-12
Added is_empty method to Range

oscargus:
Range did not have a working `is_empty` method for symbolic ranges.

#### Release Notes

* sets
   * `Range` now has a working `is_empty` method for symbolic ranges.

### DIFF
--- a/sympy/sets/fancysets.py
+++ b/sympy/sets/fancysets.py
@@ -770,6 +770,13 @@ class Range(Set):
             return True
         return self.size.is_finite
 
+    @property
+    def is_empty(self):
+        try:
+            return self.size.is_zero
+        except ValueError:
+            return None
+
     def __bool__(self):
         # this only distinguishes between definite null range
         # and non-null/unknown null; getting True doesn't mean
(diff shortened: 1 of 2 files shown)


smichr:
LGTM

sympy/sets/fancysets.py
```diff
@@ -770,6 +770,13 @@ def is_finite_set(self):
             return True
         return self.size.is_finite
 
+    @property
+    def is_empty(self):
+        try:
+            return self.size.is_zero
+        except ValueError:
+            return None
```
oscarbenjamin, line 778: I don't understand why `self.size` raises `ValueError`. It should be changed to not raise.

oscargus: Well, earlier it raised for any type of symbolic size, now it just raises if the symbols are not integers. (I added a PR some time ago.)

One need to decide what e.g. `Range(a, b, c)` should return. `(a-b)/c`? Which may be negative. So `Max((a-b)/c, 0)`. Which may be fractional. So `Max(floor((a-b)/c), 0)`?

PR #22088, opened 2021-09-12
Improved LaTeX printing of symbolic Range

oscargus:
Better printing of symbolic ranges.
a) Use `\operatorname` for `Range`
image

b) Only print required arguments
image

c) Print as set when possible.
image

#### Release Notes

* printing
    * Better LaTeX printing of symbolic `Range`.

### DIFF
--- a/sympy/printing/latex.py
+++ b/sympy/printing/latex.py
@@ -2057,13 +2057,8 @@ class LatexPrinter(Printer):
             it = iter(s)
             printset = next(it), next(it), dots
         else:
-            try:
-                if (s.size < 4) == True:
-                    printset = tuple(s)
-                else:
-                    it = iter(s)
-                    printset = next(it), next(it), dots, s[-1]
-            except (TypeError, ValueError):
+            def _print_symbolic_range():
+                # Symbolic Range that cannot be resolved
                 if s.args[0] == 0:
                     if s.args[2] == 1:
                         cont = self._print(s.args[1])
@@ -2075,11 +2070,23 @@ class LatexPrinter(Printer):
                     else:
                         cont = ", ".join(self._print(arg) for arg in s.args)
 
-                return(f"\\operatorname{{Range}}\\left({cont}\\right)")
+                return(f"\\text{{Range}}\\left({cont}\\right)")
 
+            if s.is_empty is not None:
+                if (s.size < 4) == True:
+                    printset = tuple(s)
+                else:
+                    try:
+                        it = iter(s)
+                        printset = next(it), next(it), dots, s[-1]
+                    except TypeError:
+                        return _print_symbolic_range()
+            else:
+                return _print_symbolic_range()
         return (r"\left\{" +
                 r", ".join(self._print(el) if el is not dots else r'\ldots' for el in printset) +
                 r"\right\}")
+
 
     def __print_number_polynomial(self, expr, letter, exp=None):
         if len(expr.args) == 2:
--- a/sympy/printing/latex.py
+++ b/sympy/printing/latex.py
@@ -2046,9 +2046,6 @@ class LatexPrinter(Printer):
     def _print_Range(self, s):
         dots = object()
 
-        if s.has(Symbol):
-            return self._print_Basic(s)
-
         if s.start.is_infinite and s.stop.is_infinite:
             if s.step.is_positive:
                 printset = dots, -1, 0, 1, dots
@@ -2059,11 +2056,26 @@ class LatexPrinter(Printer):
         elif s.stop.is_infinite:
             it = iter(s)
             printset = next(it), next(it), dots
-        elif len(s) > 4:
-            it = iter(s)
-            printset = next(it), next(it), dots, s[-1]
         else:
-            printset = tuple(s)
+            try:
+                if (s.size < 4) == True:
+                    printset = tuple(s)
+                else:
+                    it = iter(s)
+                    printset = next(it), next(it), dots, s[-1]
+            except (TypeError, ValueError):
+                if s.args[0] == 0:
+                    if s.args[2] == 1:
+                        cont = self._print(s.args[1])
+                    else:
+                        cont = ", ".join(self._print(arg) for arg in s.args)
+                else:
+                    if s.args[2] == 1:
+                        cont = ", ".join(self._print(arg) for arg in s.args[:2])
+                    else:
+                        cont = ", ".join(self._print(arg) for arg in s.args)
+
+                return(f"\\operatorname{{Range}}\\left({cont}\\right)")
 
         return (r"\left\{" +
                 r", ".join(self._print(el) if el is not dots else r'\ldots' for el in printset) +
--- a/sympy/printing/latex.py
+++ b/sympy/printing/latex.py
@@ -2044,6 +2044,21 @@ class LatexPrinter(Printer):
     _print_frozenset = _print_set
 
     def _print_Range(self, s):
+        def _print_symbolic_range():
+            # Symbolic Range that cannot be resolved
+            if s.args[0] == 0:
+                if s.args[2] == 1:
+                    cont = self._print(s.args[1])
+                else:
+                    cont = ", ".join(self._print(arg) for arg in s.args)
+            else:
+                if s.args[2] == 1:
+                    cont = ", ".join(self._print(arg) for arg in s.args[:2])
+                else:
+                    cont = ", ".join(self._print(arg) for arg in s.args)
+
+            return(f"\\text{{Range}}\\left({cont}\\right)")
+
         dots = object()
 
         if s.start.is_infinite and s.stop.is_infinite:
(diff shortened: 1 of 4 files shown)


smichr:
+1

sympy/printing/latex.py
```diff
@@ -2054,11 +2051,26 @@ def _print_Range(self, s):
...
-            printset = next(it), next(it), dots, s[-1]
         else:
-            printset = tuple(s)
+            try:
+                if (s.size < 4) == True:
+                    printset = tuple(s)
+                else:
+                    it = iter(s)
+                    printset = next(it), next(it), dots, s[-1]
+            except (TypeError, ValueError):
```
oscarbenjamin, line 2061: Rather than catch ValueError or TypeError can this have an explicit check. Also what raises these exceptions? Can we eliminate them at the source?

I don't like raising or catching either ValueError or TypeError since those exception types are regularly raised when there is a bug in the code.

oscargus: This is also related to that the symbolic symbols in `Range` must be integer for some things to work. Both `size` and getting `next` (or maybe at `iter`) will throw an exception if it cannot get anything, so it was way to deal with both of them.

The thing is that `Range(i, j)` will not work, but `Range(i, i+3)` will, so the easy way is to try and catch the exception (the alternative is to have the same tests as those throwing the exception, which then gives two places to change if this is improved). I know that it can hide bugs, but I think that in the printing code it is less of a problem.

jksuom:
I think that `\operatorname` is intended for functions that act like operators, i.e., single argument and no parentheses. Latex will automatically separate such a function from its argument by adding some space.

oscargus:
I was considering if I should have used that or `\text`. Normally, I am more of a `\text` person, so I'd be happy to change it to that.

sympy/printing/latex.py
smichr: Can you decide based on ` is_empty `being None

oscarbenjamin: Can there not just be an `is_iterable` attribute for Range? Basically if `is_iterable` is True then it should be safe to iterate or something.

jksuom:
Actually, names of mathematical functions are often formatted with `\mathrm` though that may not be much different from `\text`.

oscargus:
I beg to disagree there. The problem with `\mathrm` is if you have fonts that are not Roman, which is the case in e.g. the standard Beamer template. Unfortunately, there seems to be no way to get the math font in upright style. `\mathup` should really be added and almost always be what we would like to use here, as `\text` uses the text font (which often, but not always is from a similar family).

If I understand it correctly, `\text` was originally thought of for things like `x = y \text{ and } z = w`, so actual text, rather than `\text{Range}(n)` or `V_\text{DD}`.

jksuom:
> The problem with \mathrm is if you have fonts that are not Roman, 

I agree that this may be a problem. The output of `latex` is probably seldom printed on paper where more fonts are available.

Anyway, I am fine with `\text`.

sympy/printing/latex.py
oscargus: > Can you decide based on `is_empty`being None

It is probably possible to get rid of one of the exceptions then (the one coming from `size`). I do not think it existed when I wrote this code though. :-) (At least not in master.)

oscargus: One exception removed now. Would have to do something like `is_iterable` to get rid of the other one. However, `Range` defines it to be `True` already. Should I replace it with a function checking if it actually can be iterated?

asmeurer:
Can you clarify what \mathrm looks like in Beamer? I think we need to figure out which is generally better, \mathrm or \text, and use it everywhere. I always thought they were the same.

asmeurer:
Based on https://tex.stackexchange.com/questions/19502/is-there-a-preference-of-when-to-use-text-and-mathrm and https://tex.stackexchange.com/questions/98008/is-mathrm-really-preferable-to-text, it looks like `\mathrm` always gives a Roman (upright) font, whereas `\text` always matches the surrounding text. This matters in environments like Theorem where the surrounding text is italics. `\text` also requires recursive $$ to do any math formatting.

To me it seems like \text should only be used if you are putting actual text in your math (e.g., a label in a formula with \underbrace), not just a mathematical operator. 

For Range, another alternative is to use neither, and have it just print in italics like any other mathematical function. I'm somewhat partial to just doing that, as personally I like to limit the use of Roman font in LaTeX math.

Yet another option would be to print it not using "Range" at all, but rather using direct set notation.

oscargus:
If you have a san-serif font, like Helvetica or Arial, for your equations, \mathrm will replace it with a roman/serif font, like Times. Using \text will replace it with the current text font. So if similar fonts are used for both maths and text, using \text is always the better option (I changed all printing to \text a few years ago). Ideally, one should use the font used for \operatorname,, which is typically an upright version of the math font. However, there is strangely no support for that in any standard package.

(An option that may be used if one have fundamentally different fonts in ones template is to redefine \mathrm to use an upright style, however, that is not a solution suitable for SymPy... At least not if we would like the latex output to be useable...)

oscargus:
And, yes, you are correct that one should ideally only use \text for text. However, \mathrm doesn't mean upright, it means Roman, so you get a Roman font which looks quite bad in an otherwise san-serif document. So one have to pick between two evil. It is really strange that there isn't a \mathup command in e.g. amsmath.

To clarify, in Beamer you get this, but reversed, if you use mathrm:
image

(Firefox cannot copy and paste properly, but I think you get the idea, the subscript is using a completely different font compared to the name.)

asmeurer:
It seems we already had this discussion https://github.com/sympy/sympy/issues/15934. That issue is closed, but it isn't clear to me what the consensus was. The PR that closed it switched to using \text everywhere https://github.com/sympy/sympy/pull/16056. 

The serif thing makes sense. But I'm still wondering about the Theorem environment comment which was mentioned in the StackExchange posts. The answer in this question is helpful https://tex.stackexchange.com/questions/244657/how-to-display-upright-letters-in-math-mode-without-manually-creating-space. It seems to me that the context of the math matters. The "for" and "otherwise" in a Piecewise is different from an operator like Range. \text looks bad in the sans-serif example there, but I wonder if that's because of poor font choices. What does that look like in Beamer?

asmeurer:
> \text looks bad in the sans-serif example there, but I wonder if that's because of poor font choices. What does that look like in Beamer?

It is. Here is what it actually looks like in Beamer.

Screen Shot 2021-09-14 at 2 29 11 PM

It seems to me that `\textup` is correct for inline text like in Piecewise. I'm still not sure what should be used for operator like things like Range.

oscargus:
I'd say that the "correct" method depends on how you most commonly use it. To me, I do not often write theorems with italic text, but I do often reuse equations for papers and presentations. As long as the font style is the same, it is not a problem using \text. If one have san-serif text font and Roman math font, \text is as bad as using \mathrm in an otherwise san-serif document.

An option is to use \operatorname with the drawback that it has some spacing before and after. The one after can be removed, but I have not seen a solution to remove the one in front of it (which is maybe not a problem in many cases).

I think the reason it looks bad there is because they have a Roman math font. That is sort of the thing: if you have different font styles for text and math (serif/san-serif) it will look bad, independent of which way around it is. I think this table sort of explains it

| Text | Math | \text | \mathrm | \mathsf |
|-----| ---- | ---- | ---- | ---- |
| Roman | Roman | Good | Good | Bad |
| Roman | San-serif | Bad | Bad | Good |
| San-serif | Roman | Bad | Good | Bad |
| San-serif | San-serif | Good | Bad | Good |

So all ties. However, the most common case is the top one. I would claim that the second most common is the last one (Beamer default and it makes sense to use similar fonts for both text and math), but I sometimes publish in a conference that use the penultimate case and then \text is bad. But that template is bad in general:
image

oscargus:
> It seems to me that `\textup` is the most correct thing
Or `\mathup`, if that ever existed. Yes, there is probably a benefit of `\textup` over `\text` when using it in Theorems. Haven't though of that, but then I rarely use them.

asmeurer:
I agree that mixing serif and sans-serif is already bad. It's rather unfortunate that that sort of thing is becoming common on the web due to the use of MathJax, which uses its own separate font from the text. 

There may not be a uniform correct thing if you consider such cases, but we need to do something by default. latex() can't make any assumptions about where the resulting text will be used. Most often it will be used for display in the notebook (although the notebook display can pass custom arguments to latex() if necessary), but it is also used for copy-pasting into latex documents with all sorts of preset document classes. 

Perhaps the best thing would be to split out the different types of uses of `\text`/`\mathrm` (operators vs. inline text, I don't know if it needs to be finer than that) into helper functions in the printer so that they can easily be changed with a flag. It might also be worth documenting this stuff since I think most people aren't aware of the subtle differences.

oscargus:
There may be a third case as well: multi-character variable names and sub-scripts that are not indexes. Not sure if the use of \mathrm actually plays a role here anyway. If so, one should be able to configure between that and \mathsf. \operatorname will adapt to the font. (There is a variable, `\operator@text` if I recall correctly, which one theoretically could use, but I have not figured out how. I think you need to use it in the preamble to define a new math command.)

asmeurer:
Preamble stuff isn't really an option. We also have to make sure whatever we use works with MathJax and KaTeX.

oscargus:
> Preamble stuff isn't really an option.

Totally agree. Convincing AMS to add `\mathup` seems more reasonable...

Not sure how those support the AMS extensions. Noted that most of the documentation use `\mathrm`.

oscargus:
No more exceptions here.

oscarbenjamin:
Looks good to me

asmeurer:
Let's not worry about the \text thing here. Let's reopen https://github.com/sympy/sympy/issues/15934 and summarize the discussion that we had here about it. I feel like we should examine the options a little more closely. I didn't realize the actual differences between \text and \mathrm before, nor did I realize that there are potentially other options as well. It seems like we might at least want to add some options to the printer for it.

PR #22089, opened 2021-09-12
Added Mod printing for js and corrected Mod printing for C

oscargus:
`%` is remainder, not modulo, in C (and js).

#### Release Notes

* printing
    * Corrected C-printing of `Mod` (for negative divisors)
    * Added js-printing of `Mod`

### DIFF
--- a/sympy/printing/c.py
+++ b/sympy/printing/c.py
@@ -295,8 +295,11 @@ class C89CodePrinter(CodePrinter):
         if num.is_integer and den.is_integer:
             PREC = precedence(expr)
             snum, sden = [self.parenthesize(arg, PREC) for arg in expr.args]
-            # % is remainder, not modulo in C, % only works for non-negative numbers
-            if num.is_nonnegative and den.is_nonnegative:
+            # % is remainder (same sign as numerator), not modulo (same sign as
+            # denominator), in js. Hence, % only works as modulo if both numbers
+            # have the same sign
+            if (num.is_nonnegative and den.is_nonnegative or
+                num.is_nonpositive and den.is_nonpositive):
                 return f"({snum} % {sden})"
             return f"((({snum} % {sden}) + {sden}) % {sden})"
         # Not guaranteed integer
--- a/sympy/printing/c.py
+++ b/sympy/printing/c.py
@@ -293,9 +293,14 @@ class C89CodePrinter(CodePrinter):
     def _print_Mod(self, expr):
         num, den = expr.args
         if num.is_integer and den.is_integer:
-            return "(({}) % ({}))".format(self._print(num), self._print(den))
-        else:
-            return self._print_math_func(expr, known='fmod')
+            PREC = precedence(expr)
+            snum, sden = [self.parenthesize(arg, PREC) for arg in expr.args]
+            # % is remainder, not modulo in C, % only works for non-negative numbers
+            if num.is_nonnegative and den.is_nonnegative:
+                return f"({snum} % {sden})"
+            return f"((({snum} % {sden}) + {sden}) % {sden})"
+        # Not guaranteed integer
+        return self._print_math_func(expr, known='fmod')
 
     def _print_Rational(self, expr):
         p, q = int(expr.p), int(expr.q)
--- a/sympy/printing/c.py
+++ b/sympy/printing/c.py
@@ -296,12 +296,12 @@ class C89CodePrinter(CodePrinter):
             PREC = precedence(expr)
             snum, sden = [self.parenthesize(arg, PREC) for arg in expr.args]
             # % is remainder (same sign as numerator), not modulo (same sign as
-            # denominator), in js. Hence, % only works as modulo if both numbers
+            # denominator), in C. Hence, % only works as modulo if both numbers
             # have the same sign
             if (num.is_nonnegative and den.is_nonnegative or
                 num.is_nonpositive and den.is_nonpositive):
-                return f"({snum} % {sden})"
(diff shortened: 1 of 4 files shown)


oscargus:
Will need to read up a bit further. This may not work when both numerator and denominator is negative, although that expression is a standard one.

sympy/printing/c.py
```diff
@@ -293,9 +293,17 @@ def _print_Pow(self, expr):
     def _print_Mod(self, expr):
         num, den = expr.args
         if num.is_integer and den.is_integer:
-            return "(({}) % ({}))".format(self._print(num), self._print(den))
-        else:
-            return self._print_math_func(expr, known='fmod')
+            PREC = precedence(expr)
+            snum, sden = [self.parenthesize(arg, PREC) for arg in expr.args]
+            # % is remainder (same sign as numerator), not modulo (same sign as
+            # denominator), in js. Hence, % only works as modulo if both numbers
```
oscargus, line 299: in C...

(Will change, but waiting for other comments as well.)

sympy/printing/tests/test_c.py
```diff
@@ -145,13 +145,20 @@ def test_ccode_exceptions():
...
     r, s = symbols('r,s', real=True)
-    assert ccode(Mod(ceiling(r), ceiling(s))) == "((ceil(r)) % (ceil(s)))"
+    assert ccode(Mod(ceiling(r), ceiling(s))) == '((((ceil(r)) % (ceil(s))) + '\
+                                                 '(ceil(s))) % (ceil(s)))'
     assert ccode(Mod(r, s)) == "fmod(r, s)"
-
+    p1, p2 = symbols('p1 p2', integer=True, positive=True)
+    assert ccode(Mod(p1, p2)) == '(p1 % p2)'
+    assert ccode(Mod(p1, p2 + 3)) == '(p1 % (p2 + 3))'
+    assert ccode(Mod(-3, -7, evaluate=False)) == '((-3) % (-7))'
```
oscargus, line 161: Additional empty line after.

sympy/printing/jscode.py
```diff
@@ -111,6 +111,18 @@ def _print_Rational(self, expr):
...
+        num, den = expr.args
+        PREC = precedence(expr)
+        snum, sden = [self.parenthesize(arg, PREC) for arg in expr.args]
+        # % is remainder (same sign as numerator), not modulo (same sign as
+        # denominator), in js. Hence, % only works as modulo if both numbers
+        # have the same sign
+        if (num.is_nonnegative and den.is_nonnegative or
+            num.is_nonpositive and den.is_nonpositive):
+         return f"({snum} % {sden})"
+        return f"((({snum} % {sden}) + {sden}) % {sden})"
```
bjodah, line 124: do we need the explicit extra parenthesis? I'm leaning towards that we should just trust that `_print_WhateverOperator` actually calls `parenthesize` just like we are doing here.

sympy/printing/jscode.py
```diff
@@ -111,6 +111,18 @@ def _print_Rational(self, expr):
...
+    def _print_Mod(self, expr):
+        num, den = expr.args
+        PREC = precedence(expr)
+        snum, sden = [self.parenthesize(arg, PREC) for arg in expr.args]
+        # % is remainder (same sign as numerator), not modulo (same sign as
+        # denominator), in js. Hence, % only works as modulo if both numbers
+        # have the same sign
+        if (num.is_nonnegative and den.is_nonnegative or
+            num.is_nonpositive and den.is_nonpositive):
+         return f"({snum} % {sden})"
```
bjodah, line 123: is there just a single space indentation for this return statement?

oscargus: Seems like it indeed!

sympy/printing/jscode.py
oscargus: Maybe not. Strictly, the `({snum} % {sden})` should never be needed in any of the expressions. For the latter it was more that the expression I saw was printed like this, so it looked "better" somehow. The outer in the second one should not be needed either. It was more because of the issues with mod elsewhere that I thought it was safer to have them there since there is limited harm in generated code, especially compiled.

Now that #22080 is merged, I can revisit it and see if double parenthesis show up anywhere.

sympy/printing/jscode.py
```diff
@@ -111,6 +111,18 @@ def _print_Rational(self, expr):
...
+        num, den = expr.args
+        PREC = precedence(expr)
+        snum, sden = [self.parenthesize(arg, PREC) for arg in expr.args]
+        # % is remainder (same sign as numerator), not modulo (same sign as
+        # denominator), in js. Hence, % only works as modulo if both numbers
+        # have the same sign
+        if (num.is_nonnegative and den.is_nonnegative or
+            num.is_nonpositive and den.is_nonpositive):
+            return f"{snum} % {sden}"
+        return f"(({snum} % {sden}) + {sden}) % {sden}"
```
oscargus, line 124: The inner parenthesis are not strictly needed, but I think it looks clearer that way. Also, the middle `sden` can be parenthesized with a different precedence (add), but would require another call without any actual benefit.

sympy/printing/jscode.py
oscargus: Turns out that I(?) had managed to get some tabs in.

sympy/printing/jscode.py
oscargus: Seems like #22080 changed the behaviour a bit (to the better). So with added tests, I am quite confident in removing the outer parenthesis.

bjodah:
Alright, this looks good to me!

PR #22091, opened 2021-09-13
Improved and corrected LaTeX printing of Mod

oscargus:
Earlier `Mod(x, y + 1)` was printed as
image

Now it is printed as
image

Also, the LaTeX used `\bmod{7}` which is incorrect. It should just be `\bmod 7` as `\bmod` is not a command. (Probably a confusion with `\pmod` which is used like that.)

#### Other comments
Still an issue with unary -, where both `-Mod(x, y)` and `Mod(-x, y)` ends up with the same output.

#### Release Notes

* printing
     * Improved and corrected LaTeX-printing of `Mod`.

### DIFF
--- a/sympy/printing/latex.py
+++ b/sympy/printing/latex.py
@@ -1712,13 +1712,18 @@ class LatexPrinter(Printer):
 
     def _print_Mod(self, expr, exp=None):
         if exp is not None:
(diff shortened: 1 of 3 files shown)


smichr:
+1

oscarbenjamin:
Thanks

Issue #21681, opened 2021-06-28
feat(Rational) allow limit to numerator


smichr:
When seeking a simple fraction to approximate a rational, it might be useful to be able to request that the numerator of the result have a value less than a limit, too:
```python
>>> r = Rational(1819831678027117, 9007199254740992)
>>> dmax = 10**5
>>> r.limit_denominator(dmax)
11003/54459
>>> nmax = 1000
>>> duse = min(ceiling(nmax/r), dmax)
>>> Rational(a,b).limit_denominator(duse)
376/1861
```
Something like this:
```python
def approx(self, a, b=None):
    # limit(den) or limit(num, den)
    a = as_int(a)
    assert a > 0
    if self == 0:
        return self
    elif self < 0:
        return -(-self).approx(a, b)
    if b is not None:
        b = as_int(b)
        assert b > 0
        if self*b < 1:
            return self.limit_denominator(b)  # 0 or 1/b
        if b == 1:
            return Integer(min(a, max(1, int(self))))
        if a == 1:
            if self >= 1:
                return self.func(1, 1)
            return self.func(1, int(1/self))
        if self >= 1:
            b = min(a, b)
        else:
            a = min(a, b)
        duse = min(int(a/self), b)
    else:
        duse = a
    return self.limit_denominator(duse)

>>> Rational.approx = approx
>>> r.approx(10**5)
11003/54459
>>> r.approx(10**3, 10**5)
376/1861
```
-- from this [SO answer](https://stackoverflow.com/a/68169534/1089161)

smichr:
OK, let's pass on this until there is stronger motivation. The existing functionality can easily handle this infrequent case.

PR #22094, opened 2021-09-13
added function for limiting numerator

praneethratna:
Fixes issue #21681 and now we can request that the numerator of the result have a value less than a limit too using limit_num_denom() function.

#### Release Notes

* core
   * added method limit_num_denom() for limiting numerator in Rational class.

### DIFF
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -1714,25 +1714,23 @@ class Rational(Number):
         elif self < 0:
             return -(-self).limit_num_denom(a, b)
 
-        if b is not None:
-            if b < 0:
-                raise AssertionError("Value should be greater than 0")
-            if self*b < 1:
+        if b is None:
+            return self.limit_denominator(a)
+        if b < 0:
+            raise AssertionError("Value should be greater than 0")
+        if self.p*b < self.q:
                 return self.limit_denominator(b)
-            if b == 1:
-                return Integer(min(a, max(1, int(self))))
-            if a == 1:
-                if self >=1:
-                    return self.limit_num_denom(1, 1)
-                return self.limit_num_denom(1, int(1/self))
-            if self >=1:
-                b = min(a, b)
-            else:
-                a = min(a, b)
-            d_use = min(int(a/self), b)
+        if b == 1:
+            return Integer(min(a, max(1, int(self))))
+        if a == 1:
+            if self.p >= self.q:
+                return self.limit_num_denom(1, 1)
+            return self.limit_num_denom(1, self.q//self.p)
+        if self.p >= self.q:
+            b = min(a, b)
         else:
-            d_use = a
-        return self.limit_denominator(d_use)
+            a = min(a, b)
+        return self.limit_denominator(min(int(a*self.q/self.p), b))
 
     def __getnewargs__(self):
         return (self.p, self.q)
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -1691,6 +1691,49 @@ class Rational(Number):
         f = fractions.Fraction(self.p, self.q)
         return Rational(f.limit_denominator(fractions.Fraction(int(max_denominator))))
 
+    def limit_num_denom(self, a, b = None):
+        """Closest Rational to self with numerator at most a
+           and denominator at most b if not None else Closest
+           Rational to self with denominator at most a.
+
+        Examples
+        ========
+
+        >>> from sympy import Rational
+        >>> Rational(1819831678027117, 9007199254740992).limit_num_denom(10**5)
+        11003/54459
+        >>> Rational(1819831678027117, 9007199254740992).limit_num_denom(10**3, 10**5)
+        376/1861
+
+        """
+        if a < 0:
+            raise AssertionError("Value should be greater than 0")
+
+        if self == 0:
+            return self
+        elif self < 0:
+            return -(-self).limit_num_denom(a, b)
+
+        if b is not None:
+            if b < 0:
+                raise AssertionError("Value should be greater than 0")
+            if self*b < 1:
+                return self.limit_denominator(b)
+            if b == 1:
+                return Integer(min(a, max(1, int(self))))
+            if a == 1:
+                if self >=1:
+                    return self.limit_num_denom(1, 1)
+                return self.limit_num_denom(1, int(1/self))
+            if self >=1:
+                b = min(a, b)
+            else:
+                a = min(a, b)
+            d_use = min(int(a/self), b)
+        else:
+            d_use = a
+        return self.limit_denominator(d_use)
+
     def __getnewargs__(self):
         return (self.p, self.q)
 
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -1691,7 +1691,7 @@ class Rational(Number):
         f = fractions.Fraction(self.p, self.q)
         return Rational(f.limit_denominator(fractions.Fraction(int(max_denominator))))
 
-    def limit_num_denom(self, a, b = None):
+    def limit_div(self, a, b = None):
         """Closest Rational to self with numerator at most a
            and denominator at most b if not None else Closest
            Rational to self with denominator at most a.
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -1700,9 +1700,9 @@ class Rational(Number):
         ========
 
         >>> from sympy import Rational
-        >>> Rational(1819831678027117, 9007199254740992).limit_num_denom(10**5)
+        >>> Rational(1819831678027117, 9007199254740992).limit_div(10**5)
         11003/54459
-        >>> Rational(1819831678027117, 9007199254740992).limit_num_denom(10**3, 10**5)
+        >>> Rational(1819831678027117, 9007199254740992).limit_div(10**3, 10**5)
         376/1861
 
         """
@@ -1712,7 +1712,7 @@ class Rational(Number):
         if self == 0:
             return self
         elif self < 0:
-            return -(-self).limit_num_denom(a, b)
+            return -(-self).limit_div(a, b)
 
         if b is None:
             return self.limit_denominator(a)
@@ -1724,8 +1724,8 @@ class Rational(Number):
             return Integer(min(a, max(1, int(self))))
         if a == 1:
             if self.p >= self.q:
-                return self.limit_num_denom(1, 1)
-            return self.limit_num_denom(1, self.q//self.p)
+                return self.limit_div(1, 1)
+            return self.limit_div(1, self.q//self.p)
         if self.p >= self.q:
             b = min(a, b)
         else:
--- a/sympy/core/tests/test_numbers.py
+++ b/sympy/core/tests/test_numbers.py
@@ -321,16 +321,15 @@ def test_Rational_new():
     assert Rational(19, 25).limit_denominator(4) == n3_4
     assert Rational('19/25').limit_denominator(4) == n3_4
     n2_3 = Rational(2, 3)
-    assert Rational('2/3') == n2_3
-    assert -Rational('-2/3') == n2_3
     assert Rational('.68').limit_num_denom(3, 5) == n2_3
-    assert Rational(-17, 25).limit_num_denom(3, 5) == -n2_3
     assert Rational('17/25').limit_num_denom(5) == Rational('17/25').limit_denominator(5)
+    assert S.Zero.limit_num_denom(3, 5) == S.Zero
+    assert Rational(-17, 25).limit_num_denom(3, 5) == -n2_3
     assert Rational(7, 25).limit_num_denom(2, 3) == Rational(7, 25).limit_denominator(3)
     assert Rational(17, 25).limit_num_denom(3, 1) == 1
     assert Rational(17, 25).limit_num_denom(1, 5) == 1
     assert Rational(25, 17).limit_num_denom(1, 5) == 1
-    assert Rational(25, 17).limit_num_denom(3, 5) == Rational(1, n2_3)
+    assert Rational(25, 17).limit_num_denom(3, 5) == Rational(3, 2)
     raises(AssertionError, lambda: Rational(17, 25).limit_num_denom(-5))
     raises(AssertionError, lambda: Rational(17, 25).limit_num_denom(3, -5))
     assert Rational(1.0, 3) == Rational(1, 3)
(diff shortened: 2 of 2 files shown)


oscargus:
Please add some tests for it as well, including a negative one.

praneethratna:
@oscargus Added tests , please review.

sympy/core/numbers.py
```diff
@@ -1691,6 +1691,49 @@ def limit_denominator(self, max_denominator=1000000):
...
+        >>> Rational(1819831678027117, 9007199254740992).limit_num_denom(10**3, 10**5)
+        376/1861
+
+        """
+        if a < 0:
+            raise AssertionError("Value should be greater than 0")
+
+        if self == 0:
+            return self
+        elif self < 0:
```
oscargus, line 1714: I was especially thinking about testing this line, which I do not think happens at the moment.

sympy/core/numbers.py
```diff
@@ -1691,6 +1691,49 @@ def limit_denominator(self, max_denominator=1000000):
...
+            return -(-self).limit_num_denom(a, b)
+
+        if b is not None:
+            if b < 0:
+                raise AssertionError("Value should be greater than 0")
+            if self*b < 1:
+                return self.limit_denominator(b)
+            if b == 1:
```
oscargus, line 1722: This doesn't seem like it is tested either.

sympy/core/numbers.py
```diff
@@ -1691,6 +1691,49 @@ def limit_denominator(self, max_denominator=1000000):
...
+                return Integer(min(a, max(1, int(self))))
+            if a == 1:
```
oscargus, line 1724: Nor this.

sympy/core/numbers.py
```diff
@@ -1691,6 +1691,49 @@ def limit_denominator(self, max_denominator=1000000):
...
+                if self >=1:
+                    return self.limit_num_denom(1, 1)
+                return self.limit_num_denom(1, int(1/self))
+            if self >=1:
```
oscargus, line 1728: Nor this.

sympy/core/tests/test_numbers.py
```diff
@@ -320,6 +320,16 @@ def test_Rational_new():
     assert Rational('.76').limit_denominator(4) == n3_4
     assert Rational(19, 25).limit_denominator(4) == n3_4
     assert Rational('19/25').limit_denominator(4) == n3_4
+    n2_3 = Rational(2, 3)
+    assert Rational('2/3') == n2_3
```
oscargus, line 324: Doesn't this (and the next few test) just test the parsing? Not needed to test again.

sympy/core/tests/test_numbers.py
```diff
@@ -320,6 +320,16 @@ def test_Rational_new():
...
+    assert -Rational('-2/3') == n2_3
+    assert Rational('.68').limit_num_denom(5) == n2_3
+    assert Rational('17/25').limit_num_denom(5) == Rational('17/25').limit_denominator(5)
+    assert Rational(17, 25).limit_num_denom(3, 5) == n2_3
+    raises(AssertionError, lambda: Rational('17/25').limit_num_denom(-5))
```
oscargus, line 329: Better not use the parsing version here. Already tested above.

oscargus:
Try to add tests that actually executes all the different cases that you cover in the code.

Out of the tests that you added, one does actually test the workings of the new code (and four tests that there will be exceptions, which is also important, although the last one may seem redundant, and maybe the second to last as well).

sympy/core/tests/test_numbers.py
```diff
@@ -320,6 +320,19 @@ def test_Rational_new():
...
+    assert Rational('2/3') == n2_3
+    assert -Rational('-2/3') == n2_3
+    assert Rational('.68').limit_num_denom(3, 5) == n2_3
+    assert Rational(-17, 25).limit_num_denom(3, 5) == -n2_3
+    assert Rational('17/25').limit_num_denom(5) == Rational('17/25').limit_denominator(5)
+    assert Rational(7, 25).limit_num_denom(2, 3) == Rational(7, 25).limit_denominator(3)
+    assert Rational(17, 25).limit_num_denom(3, 1) == 1
+    assert Rational(17, 25).limit_num_denom(1, 5) == 1
+    assert Rational(25, 17).limit_num_denom(1, 5) == 1
+    assert Rational(25, 17).limit_num_denom(3, 5) == Rational(1, n2_3)
```
oscargus, line 333: I think this relies on additional evaluations and in this case there is no speed improvement compared to `Rational(3, 2)`.

sympy/core/tests/test_numbers.py
```diff
@@ -320,6 +320,19 @@ def test_Rational_new():
     assert Rational('.76').limit_denominator(4) == n3_4
     assert Rational(19, 25).limit_denominator(4) == n3_4
     assert Rational('19/25').limit_denominator(4) == n3_4
+    n2_3 = Rational(2, 3)
+    assert Rational('2/3') == n2_3
```
oscargus, line 324: Maybe I wasn't clear, but as these do not add anything in terms of testing, I would remove them.

sympy/core/numbers.py
oscargus, line 1714: I cannot see the this is tested.

sympy/core/numbers.py
```diff
@@ -1691,6 +1691,49 @@ def limit_denominator(self, max_denominator=1000000):
...
+        """
+        if a < 0:
+            raise AssertionError("Value should be greater than 0")
+
+        if self == 0:
+            return self
+        elif self < 0:
+            return -(-self).limit_num_denom(a, b)
+
+        if b is not None:
```
oscargus, line 1717: Maybe it is better to restructure the code to deal with "early exits" quite early in the code as well (not from an execution time perspective, but from a logical perspective).
```
if b is None:
    return self.limit_denominator(a)
if b < 0:
    raise AssertionError("Value should be greater than 0")
The rest of the cases.
```

sympy/core/numbers.py
```diff
@@ -1691,6 +1691,49 @@ def limit_denominator(self, max_denominator=1000000):
...
+                return Integer(min(a, max(1, int(self))))
+            if a == 1:
+                if self >=1:
+                    return self.limit_num_denom(1, 1)
+                return self.limit_num_denom(1, int(1/self))
+            if self >=1:
+                b = min(a, b)
+            else:
+                a = min(a, b)
+            d_use = min(int(a/self), b)
```
oscargus, line 1732: It may be more efficient to write this as `min((a*self.q)//self.p), b)` to just use the standard integer operations. (I think that `int` does the same as `//`)

sympy/core/numbers.py
```diff
@@ -1691,6 +1691,49 @@ def limit_denominator(self, max_denominator=1000000):
...
+            if b < 0:
+                raise AssertionError("Value should be greater than 0")
+            if self*b < 1:
+                return self.limit_denominator(b)
+            if b == 1:
+                return Integer(min(a, max(1, int(self))))
+            if a == 1:
+                if self >=1:
+                    return self.limit_num_denom(1, 1)
+                return self.limit_num_denom(1, int(1/self))
```
oscargus, line 1727: `int(1/self)` => `self.q//self.p`

sympy/core/numbers.py
oscargus: Nah, `int` and `//` behaves differently for negative numbers, but it would still be better to use `self.q` and `self.p` rather than relying on division by a `Rational`. So `min(int(a*self.q/self.p), b)` (and similar above in the other comment).

sympy/core/numbers.py
oscargus, line 1728: Maybe `self.p >= self.q`.

sympy/core/numbers.py
```diff
@@ -1691,6 +1691,49 @@ def limit_denominator(self, max_denominator=1000000):
...
+
+        if self == 0:
+            return self
+        elif self < 0:
+            return -(-self).limit_num_denom(a, b)
+
+        if b is not None:
+            if b < 0:
+                raise AssertionError("Value should be greater than 0")
+            if self*b < 1:
```
oscargus, line 1720: Maybe `self.p*b < self.q`

sympy/core/numbers.py
praneethratna: The case with negative numbers has been handled previously so i think can we go ahead with `//` ?

sympy/core/numbers.py
praneethratna: This test `assert Rational(-17, 25).limit_num_denom(3, 5) == -n2_3` is corresponding to `self < 0` . or do you mean a test case for `self == 0` ?

sympy/core/numbers.py
oscargus: Yes, forgot to mention that it may still be possible. (This is where tests for all cases are handy, easy to see if one breaks anything by changing it.)

sympy/core/tests/test_numbers.py
praneethratna: Sorry , got confused will remove them now.

sympy/core/numbers.py
oscargus: Sorry, missed that. But the `self == 0` would indeed be good to test as well, for completeness.

sympy/core/numbers.py
```diff
@@ -1691,6 +1691,47 @@ def limit_denominator(self, max_denominator=1000000):
         f = fractions.Fraction(self.p, self.q)
         return Rational(f.limit_denominator(fractions.Fraction(int(max_denominator))))
 
+    def limit_num_denom(self, a, b = None):
```
smichr, line 1694: I'm not sure that it should be this way. If the user wants to limit denominator then there is a function for that. I'm not sure it's a good idea to mix that into this function.

praneethratna: So , we can change this by removing default for `b` so that this particular method can be used when the user wants to limit both numerator and denominator.

oscargus: Or maybe just add a wrapper `limit_numerator` that calls this?

oscargus: I take that back. I thought that a single argument input was limiting the numerator.

praneethratna: Or, the limit_denominator method can be removed and a single method for limiting both numerator and denominator/ for only denominator can be added ?

oscargus: I think it is probably easier to have three functions: numerator, denominator and both. At least at the top level. Then one function can use any other function if required.

If there is only one function then one will have to call it as `limit_num_den(None, 5)` and so on. Which of course can be easily solved by `limit_denominator(5)` returning `limit_num_den(None, 5)`.

So as long as there are three functions to call, the actual implementation can be in multiple ways. I think that the "problem" right now is not that your new function calls `limit_denominator`, it is more that one cannot get the `limit_numerator` behavior (if I understand it correctly).

smichr:
My proposal was named `approx` and was not suggestive of what the order of args would be whereas yours is `limit_num_den` which suggests that you should be limiting both values. What about naming yours `limit_div` (as in dividend and/or divisor). What I like about `limit_div(10)` giving `limit_denominator` is that it is like saying "I want the divisor to be less than 10" and then `limit_div(10, 100)` is like saying "make the numerator less than 10 and denom less than 100" -- args are in the same order as they would be for writing the limits in a fraction (as `Rational(10, 100)`). This is also how range works: `range(5)` means "go up to 5 from normal start" and `range(3, 5)` means "start from 3 and go up to 5".

praneethratna:
> My proposal was named `approx` and was not suggestive of what the order of args would be whereas yours is `limit_num_den` which suggests that you should be limiting both values. What about naming yours `limit_div` (as in dividend and/or divisor). What I like about `limit_div(10)` giving `limit_denominator` is that it is like saying "I want the divisor to be less than 10" and then `limit_div(10, 100)` is like saying "make the numerator less than 10 and denom less than 100" -- args are in the same order as they would be for writing the limits in a fraction (as `Rational(10, 100)`). This is also how range works: `range(5)` means "go up to 5 from normal start" and `range(3, 5)` means "start from 3 and go up to 5".

@smichr Yes, I've changed the name of the function to `limit_div` .

oscargus:
But wouldn't it make sense to be able to limit the numerator only? That seems to be the original issue.

praneethratna:
> But wouldn't it make sense to be able to limit the numerator only? That seems to be the original issue.

The original issue had an approach for limiting either denominator or both numerator and denominator.

oscargus:
The code, yes, but the title of that (and this) is limiting numerator.

smichr:
> denominator or both numerator and denominator

Maybe best to keep the order the same so `limit_div(den)`, `limit_div(den, num)`, `limit_div(None, num)`?

praneethratna:
> > denominator or both numerator and denominator
> 
> Maybe best to keep the order the same so `limit_div(den)`, `limit_div(den, num)`, `limit_div(None, num)`?

The order was `limit_div(den)` , `limit_div(num, den)` in the original approach mentioned by you.

smichr:
> The order was `limit_div(den)` , `limit_div(num, den)` in the original approach mentioned by you.

You know, maybe this is all just overkill. What about just making it `limit_numerator` and if someone wants both they can call `x.limit_numerator(num).limit_denominator(den)`.

sympy/core/numbers.py
```diff
@@ -1691,6 +1691,47 @@ def limit_denominator(self, max_denominator=1000000):
         f = fractions.Fraction(self.p, self.q)
         return Rational(f.limit_denominator(fractions.Fraction(int(max_denominator))))
 
+    def limit_div(self, a, b = None):
```
smichr, line 1694: I think this can just be like the following
```python
def limit_numerator(self, n):
    if not n or not self:
        return S.Zero
    if self < 0:
        return -(-self).limit_numerator(n)
    return 1/(1/self).limit_denominator(n)
```
Note that for the doctest example this gives a better answer:
```
>>> r
1819831678027117/9007199254740992
>>> a
11003/54459
>>> b = 1/(1/r).limit_denominator(10**5); b
76269/377491
>>> abs(b-r) < abs(a-r)
True
```
So I'm not sure where I went wrong, but the current routine is not right.

smichr: Also, I wonder if an improper fraction of a number greater than 1 should have the numerator of the fractional part reduced. Consider:
```python
>>> p=Rational(pi.n())
>>> p
884279719003555/281474976710656
>>> r=p-int(p)
>>> r
39854788871587/281474976710656

For this case they are the same

>>> 1/(1/p).limit_denominator(1000)
355/113
>>> 3 + 1/(1/r).limit_denominator(1000)
355/113


But not for this case

>>> a=1/(1/p).limit_denominator(10**7)
>>> b=3+1/(1/r).limit_denominator(10**7)
>>> abs(a-p)=<abs(b-p)
False
>>> a,b
(5419351/1725033, 165707065/52746197)
```

smichr:
closing per comment in main issue

Issue #12628, opened 2017-05-10
as_numer_denom non-deterministically influences the state of Python’s random number generator


Wrzlprmft:
When I apply `as_numer_denom` to a power expression, this changes the state of Python’s random number generator. This happens in a non-deterministic fashion, i.e., it isn’t just that random numbers are generated for whatever reason, but the state seems to be reset using some external information (like the system time).

Here is a minimal non-working example:

``` Python

import random
from sympy.abc import x

results = set()
for _ in range(10):
	random.seed(42)
	(x**2).as_numer_denom()
	results.add(random.random())
assert len(results)==1
```

asmeurer:
We should be using a random.Random instance so that it doesn't affect the random seed for other applications. 

As a workaround, I suggest doing the same in your own code, so that nothing else can mess with random like this.

Wrzlprmft:
> As a workaround, I suggest doing the same in your own code, so that nothing else can mess with random like this.

That’s what I eventually did. I stumbled upon this when checking something quickly and dirtily and since I didn’t expect SymPy to use random numbers for this (let alone in a way that is not even deterministic), I was rather baffled and it took me a while to find out what happened.

caph1993:
Same here. Spent about two hours until finding the issue.

I've written another short example because I found something that might be useful: the issue happens only at the first call of `my_test`. Subsequent calls do not affect python's random state.

```python
import sympy as sp
import random

def my_test():
    random.seed('HolaAmigo')
    # Should print -1735515812038708070 (may vary on your laptop)
    print(hash(random.getstate()))
    for i in range(100):
        x = random.randrange(2) * sp.Symbol('x')
    # Should print 3791330111857148225 (may vary on your laptop)
    print(hash(random.getstate()))
    return

my_test() # fails
my_test() # ok
my_test() # ok
my_test() # ok
```
You may comment out the sp.Symbol to check what the value would be if no intervention from sympy occurred.

laolux:
I ran into a similar problem, using `sympy.acos`.
The following program shows different results for variable `after` for consecutive runs.

    import sympy
    import random
    random.seed(28)
    
    before = random.uniform(0,1)
    print('before: ',before)
    m0, m1= sympy.symbols('m_0 m_1', real=True)
    result = sympy.acos(-m0/m1)
    after = random.uniform(0,1)
    print('after: ',after)

What about adding a warning in the documentation about this behavior? I think the section `Gotchas and Pitfalls` would be good for this.

[...]

PR #22099, opened 2021-09-14
adding module sympy.matrices.random

sonntagsgesicht:

- generates random matrices of various types like
  - orthogonal
  - unitary
  - invertible
  - singular
  - symmetric
  - nilpotent
  - idempotent
  - normal
  - symmetric
  - hermite 
  - elementary
  - rotation
  - reflection

#### Release Notes

* matrices
  * Added random matrix generation, i.e. for orthogonal, unitary, invertible and more matrix types.

### DIFF
--- a/doc/src/modules/matrices/random.rst
+++ b/doc/src/modules/matrices/random.rst
@@ -13,7 +13,8 @@ In order to setup distinct mathematical problems, like
  - non-standard geometries by non-standard scalar products
  - orthogonal basis of vector space,
 
-require specific types of matrices.
+specific types of matrices are required.
+
 A purely random generation of matrix entries does not grant any
 specific shape or property of the matrix.
 So, those matrices have to be created carefully using
@@ -25,20 +26,22 @@ Generators on different types of matrices are presented below.
 
    Even for carefully sampled matrices which are elements in topological matrix groups,
    the sample distribution does not claim to meet any uniform distribution
-   (with respect to the groups haar measure). The sample procedures focuses
-   mainly on educational and academic problems than a proper pre-defined
-   distribution.
+   (with respect to the groups Haar measure).
+   So the sampling procedures may be used to produce educational or academic problems
+   but do not generate proper pre-defined distribution.
 
 There are three different types of matrix generators:
 
-   - the *base matrices* which from the atomic building block of the later
+   - the *base matrices* which form the atomic building block of the later
    - the *compound matrices* which are simply products of base matrices of given types
-   - the *conjugate matrices* which are usually given as base or compound matrix $A$
-     conjugate by another compound matrix $S$, i.e.
+   - the *conjugate matrices* which are usually given as base or compound matrix $\mathbf{A}$
+     conjugate by another compound matrix $\mathbf{S}$,  i.e.
 
       .. math::
 
-         B = S \cdot A \cdot S^{-1} \text{ or } C = S \cdot A \cdot S^{t}
+         \mathbf{B} = \mathbf{S} \cdot \mathbf{A} \cdot \mathbf{S}^{-1}
+         \text{ or }
+         \mathbf{C} = \mathbf{S} \cdot \mathbf{A} \cdot \mathbf{S}^{T}
 
 
  1. *base matrices* are simple matrix types with only a few non trivial entries.
@@ -79,9 +82,10 @@ There are three different types of matrix generators:
     - :py:func:`triangular <sympy.matrices.random.triangular>` as product of *triangular*
       :py:func:`elementary <sympy.matrices.random.elementary>` matrices
 
- 3. *conjugate matrices* that arise as a product $SAS^{-1}$ of specific matrices $A$
+ 3. *conjugate matrices* that arise as a product $\mathbf{SAS}^{-1}$
+    of specific matrices $\mathbf{A}$
     - usually given as a *normal form* - conjugate by an
-    :py:func:`invertible <sympy.matrices.random.invertible>` $S$
+    :py:func:`invertible <sympy.matrices.random.invertible>` $\mathbf{S}$
     such as
 
     - :py:func:`diagonalizable <sympy.matrices.random.diagonalizable>` with
@@ -96,8 +100,9 @@ There are three different types of matrix generators:
     - :py:func:`nilpotent <sympy.matrices.random.nilpotent>` with
       :py:func:`jordan_normal <sympy.matrices.random.jordan_normal>` (with zero eigenvalues)
 
-    *isometry conjugate matrices* that arise as a product $OAO^{-1} = OAO^{t}$ of specific matrices $A$
-    where $A$ sets the (complex) eigenvalue spectrum.
+    *isometry conjugate matrices* that arise as a product $\mathbf{OAO}^{-1} = \mathbf{OAO}^{T}$
+    of specific matrices $A$
+    where $\mathbf{A}$ sets the (complex) eigenvalue spectrum.
 
     - :py:func:`orthogonal <sympy.matrices.random.orthogonal>`
       can be a conjugate of
@@ -119,13 +124,14 @@ There are three different types of matrix generators:
       or
       :py:func:`unitary <sympy.matrices.random.unitary>`
 
-    finally *symmetric matrices* given as a as product $SS^t$ resp. $S\bar{S}^t$ of $S$, such as
+    finally *symmetric matrices* given as a as product $\mathbf{MM}^T$
+    resp. $\mathbf{MM}^H$ of $\mathbf{M}$, such as
 
-    - :py:func:`symmetric <sympy.matrices.random.symmetric>` $SS^t$ with
-      :py:func:`invertible <sympy.matrices.random.invertible>` $S$
+    - :py:func:`symmetric <sympy.matrices.random.symmetric>` $\mathbf{MM}^T$ with
+      :py:func:`invertible <sympy.matrices.random.invertible>` $\mathbf{M}$
 
-    - :py:func:`hermite <sympy.matrices.random.hermite>` $S\bar{S}^t$ with
-      :py:func:`invertible <sympy.matrices.random.orthogonal>` $S$
+    - :py:func:`hermite <sympy.matrices.random.hermite>` $\mathbf{MM}^H$ with
+      :py:func:`invertible <sympy.matrices.random.orthogonal>` $\mathbf{M}$
 
 
 In addition to the type of matrix, also the type of entries (as a commutative ring with one)
@@ -163,7 +169,7 @@ The product of *elementary* matrices
     [-1, 1, 0],
     [ 0, 0, 1]])
 
-gives an *invertible* matrix ``M``
+gives an *invertible* matrix $\mathbf{M}$
 
     >>> M = A*B*C
     >>> M
@@ -177,6 +183,35 @@ gives an *invertible* matrix ``M``
     [1, 0, 2],
     [1, 1, 1],
     [0, 0, 1]])
+
+
+Random Generator
+----------------
+
+The module uses its own random generator ``rand`` which is a instance of ``random.Random()``.
+Hence, the state can be set by invoking ``rand.seed()``.
+
+
+    >>> from sympy.matrices.random import rand, orthogonal
+    >>> rand.seed(1)
+    >>> orthogonal(3)
+    Matrix([
+    [-1,          0,          0],
+    [ 0,  sqrt(2)/2, -sqrt(2)/2],
+    [ 0, -sqrt(2)/2, -sqrt(2)/2]])
+
+    >>> orthogonal(3)
+    Matrix([
+    [sqrt(2)/2,          0, -sqrt(2)/2],
+    [      1/2, -sqrt(2)/2,        1/2],
+    [     -1/2, -sqrt(2)/2,       -1/2]])
+
+    >>> rand.seed(1)
+    >>> orthogonal(3)
+    Matrix([
+    [-1,          0,          0],
+    [ 0,  sqrt(2)/2, -sqrt(2)/2],
+    [ 0, -sqrt(2)/2, -sqrt(2)/2]])
 
 
 Matrix Functions Reference
--- a/doc/src/modules/matrices/random.rst
+++ b/doc/src/modules/matrices/random.rst
@@ -1,0 +1,187 @@
+.. _matrices-random:
+
+Generate Specific Random Matrices
+=================================
+
+In order to setup distinct mathematical problems, like
+
+ - systems of linear equations,
+ - invertible matrices,
+ - triangular matrices
+ - differential equations with given eigenvalues
+ - linear isometries
+ - non-standard geometries by non-standard scalar products
+ - orthogonal basis of vector space,
+
+require specific types of matrices.
+A purely random generation of matrix entries does not grant any
+specific shape or property of the matrix.
+So, those matrices have to be created carefully using
+classical classification theorems of linear algebra.
+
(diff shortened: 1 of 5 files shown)


doc/src/modules/matrices/random.rst
```diff
@@ -0,0 +1,187 @@
...
+So, those matrices have to be created carefully using
+classical classification theorems of linear algebra.
+
+Generators on different types of matrices are presented below.
+
+.. note::
+
+   Even for carefully sampled matrices which are elements in topological matrix groups,
+   the sample distribution does not claim to meet any uniform distribution
+   (with respect to the groups haar measure). The sample procedures focuses
```
oscargus, line 28: Haar, focus (or sample)

sympy/matrices/random.py
```diff
@@ -0,0 +1,2248 @@
...
+_unitary_scalar_set = tuple(c * 1 + s * _i for c, s in _rotation_scalar_set)
+
+
+# === fundamental constructor ===
+
+def super_elementary_matrix(dim,
+                            index=None,
+                            value=None,
+                            *scalar_set):
+    r"""super elementary matrix n x n, i.e. identity_matrix with on 2 x 2 block
```
oscargus, line 63: one(?)

sympy/matrices/random.py
```diff
@@ -0,0 +1,2248 @@
...
+
+    Explanation
+    ===========
+    The super elementary matrix $A$ is a gerealization of elementary matrices
```
oscargus, line 67: generalization

sympy/matrices/random.py
```diff
@@ -0,0 +1,2248 @@
...
+
+        A[i,i] = a, \quad A[i,j] = b, \quad A[j,i] = -c, \quad A[j,j] = d
+
+    and $ad - bc \neq 0$.
+    So multiplication with $A$ gives for
+
+    * $a=0=d$ and $b=1=-c$ a ``transposition``, i.e. swapping rows $i$ and $j$
+    * $a=1$, $d=\lambda$ and $b=0=-c$ this matrix describes scaling the
+        row $j$ by $\lambda$
+    * $a=1=d$, $b=\mu, $-c=0$ adding the $\mu multiple
```
oscargus, line 629: $\mu$

oscargus:
Very nice and valuable contribution!

Just pointed out some minor things.

I was also a bit confused with notation, but I am not sure if there is a uniform SymPy matrix notation, so I will just note it open for consideration. In general, I think we should use `\mathbf{}` for matrices. I was also surprised to see a small t for transpose (rather than capital T or \intercal) and I managed to figure out that `\bar` meant conjugate (right?). I'm used to `^H` for conjugate transpose, but then I now that it is not always used. I have not checked any other matrix documentation, so it may be that you have used exactly the same formatting. (I also thing that it is a "cultural" thing how the notations are used.)

Anyway, welcome to SymPy. Impressive first contribution!

doc/src/modules/matrices/random.rst
```diff
@@ -0,0 +1,187 @@
...
+   mainly on educational and academic problems than a proper pre-defined
+   distribution.
+
+There are three different types of matrix generators:
+
+   - the *base matrices* which from the atomic building block of the later
```
oscarbenjamin, line 34: form

doc/src/tutorial/matrices.rst
```diff
@@ -231,6 +231,163 @@ filled with `0`\ s.
...
+Random Matrices
+---------------
+
+To create Matrices with specific properties use one of
+the random matrix generation functions, e.g.
+``invertible``, ``triangular``, ``orthogonal``.
+Find a :ref:`full list of random matrices <matrices-random>`.
+
+    >>> import random
+    >>> random.seed(1)
```
oscarbenjamin, line 244: A doctest should not modify the global random seed. Also there should be a better way to seed these functions.

oscarbenjamin: See e.g. https://github.com/sympy/sympy/issues/12628

sympy/matrices/tests/test_random.py
```diff
@@ -0,0 +1,463 @@
...
+TEST_PRECISION = 7
+
+random.seed(1)
+
+
+def _check_triangular(m):
+    s, t = m.shape
+    for i in range(s):
+        for j in range(i):
+            assert m[i, j] == 0, m
```
oscarbenjamin, line 25: The tests below would be clearer if this was just a boolean function e.g.:
```python
def is_triangular(m):
    s, t = m.shape
    return all(m[i, j] == 0 for i in range(s) for j in range(i))
```
Then it can be used like:
```python
assert is_triangular(M)
assert not is_triangular(N)
```

sympy/matrices/tests/test_random.py
```diff
@@ -0,0 +1,463 @@
...
+    # with raises(ValueError):
+    #    rotation(3, scalar=4)
+    with raises(ValueError):
+        jordan_normal(3, (None, 1, 2))
+
+
+if __name__ == '__main__':
+    import sympy
+
+    sympy.test(__file__, verbose=True, subprocess=False)
```
oscarbenjamin, line 463: We shouldn't have this in a module

sympy/matrices/random.py
```diff
@@ -0,0 +1,2248 @@
...
+
+_eps = 1e-13
+
+
+def _get_value(o, i, j):
+    return o[i, j]
+
+
+def _set_value(o, i, j, v):
+    o[i, j] = v
```
oscarbenjamin, line 29: What is the purpose of this function? I find the array subscript easier to understand than the function call.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2248 @@
...
+
+def _cs(scalar):
+    sgn = random.choice((-1, 1))
+    if isinstance(scalar, (tuple, list)) and len(scalar) == 2:
+        return scalar
+    if isinstance(scalar, complex) or getattr(scalar, 'is_real', True) is False:
+        return _re(scalar), _im(scalar)
+    c = scalar
+    s = sgn * _sqrt(1 - c * c)
+    return c, s
```
oscarbenjamin, line 44: What exactly does this function do? Is it returning a random cos/sin pair?

sympy/matrices/random.py
```diff
@@ -0,0 +1,2248 @@
...
+
+
+def _inv(o):
+    return o.inv()
+
+
+def _cs(scalar):
+    sgn = random.choice((-1, 1))
+    if isinstance(scalar, (tuple, list)) and len(scalar) == 2:
+        return scalar
```
oscarbenjamin, line 39: random.choice should not be called until it is needed

sympy/matrices/random.py
```diff
@@ -0,0 +1,2248 @@
...
+    is super elementary, i.e. $ad+bc \neq 0$. In higher dimensions,
+    any matrix $A$ looking like identity_matrix matrix but with entries
+
+    .. math::
+
+        A[i,i] = a, \quad A[i,j] = b, \quad A[j,i] = -c, \quad A[j,j] = d
+
+    and $ad - bc \neq 0$.
+
+    This inculdes elementary matrices of matrix operation for Gauss elimination.
```
oscarbenjamin, line 85: includes

sympy/matrices/random.py
```diff
@@ -0,0 +1,2248 @@
...
+    sympy.matrices.dense.eye
+
+    """
+
+    return super_elementary_matrix(dim)
+
+
+def projection(dim,
+               index=None):
+    r"""a projection matrix n x n
```
oscarbenjamin, line 367: A randomly generated n x n projection matrix

sympy/matrices/random.py
```diff
@@ -0,0 +1,2248 @@
...
+
+    Explanation
+    ===========
+    A projection is a identity_matrix like matrix
+    but with zero diagonal entiries off **index**.
```
oscarbenjamin, line 372: Single `*` like `*index*` for a parameter.

Should be a blank line under `=========`

oscarbenjamin: entries

sympy/matrices/random.py
```diff
@@ -0,0 +1,2248 @@
...
+    d = _multiply(p.inv(), d, p)
+    md = _multiply(mat, d)
+    return md + _multiply(md, mat, i - d)
+
+
+# === base matrices ===
+
+
+def identity_matrix(dim):
+    r"""identity_matrix matrix n x n
```
oscarbenjamin, line 331: What is the difference between this and the `eye` function.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2248 @@
...
+    Explanation
+    ===========
+    Let $I$ be the $n \times n$ identity_matrix matrix
+    and $D$ an diagonal matrix with only zero and one entries
+    of rank $r$ (i.e. all entries but $n-r$ diagonal entries are zero).
+
+    If a full rank matrix $A$ is given, then
+
+    .. math::
+        B = D * (I + A * (I-D)) = D + D * A * (I-D)
```
oscarbenjamin, line 281: Should this be LaTeX?

sonntagsgesicht:
Thx for the feedback and review. I'll take care of your good suggestions within the next days. Happy to adopt SymPy matrix notation.

doc/src/tutorial/matrices.rst
sonntagsgesicht: Thx for the hint.

sympy/matrices/tests/test_random.py
sonntagsgesicht: Good point.

sympy/matrices/tests/test_random.py
sonntagsgesicht: Removed it.

sympy/matrices/random.py
sonntagsgesicht: Used some of these wrappers to wrap the matrix and complex numbers framework (e.g. like `_inv` or `_re`). This keeps the code agnostic so it can be used for other frameworks (like `numpy`, `mpmath` or other.). 
I would appreciate it if we keep it that way.

sympy/matrices/random.py
sonntagsgesicht: This functions provides proper eigenvalues of isometries. In latest version it can have real output (cos/sin) or complex (cos + sin *I). For convenience, different input types are handled (cos/sin, complex units or just cos (with random sin)). It ensures same behaviour for all isometry related matrices (like `rotation`, `reflection`, `isometry_normal` (form), `orthogonal`, `unitary` and `normal`).

sympy/matrices/random.py
sonntagsgesicht: fixed

sympy/matrices/random.py
```diff
@@ -0,0 +1,2248 @@
...
+       A = \left[\begin{array}{cc}a & b \\ -c & d\end{array}\right]
+
+    is super elementary, i.e. $ad+bc \neq 0$. In higher dimensions,
+    any matrix $A$ looking like identity_matrix matrix but with entries
+
+    .. math::
+
+        A[i,i] = a, \quad A[i,j] = b, \quad A[j,i] = -c, \quad A[j,j] = d
+
+    and $ad - bc \neq 0$.
```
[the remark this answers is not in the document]
sonntagsgesicht: yes, fixed.

sympy/matrices/random.py
sonntagsgesicht: kind of redundancy -> removed it.

sympy/matrices/random.py
sonntagsgesicht: As suggested, use now `\mathbf` for matrices in LaTeX - and `\cdot` instead of *

sonntagsgesicht:
Had to re-open this as PR #22103 by my academic account. 
Code and commits are still the same. 
Sorry for the inconvenience.

sympy/matrices/random.py
oscarbenjamin: Can you name a framework that doesn't use `o[i, j] =v` for setting a matrix element?

sympy/matrices/tests/test_random.py
```diff
@@ -0,0 +1,535 @@
...
+        assert _is_eye(m.H * m), repr(simplify(m.H * m))
+
+
+def test_normal():
+    for d in list(TEST_DIMS)[1::3]:
+        spec = (2, 3, 4)
+        m = normal(d, spec, length=d)
+        assert all(x.is_real for x in simplify(m)), repr(simplify(m))
+        assert _is_zeros(m.T * m - m * m.T), \
+            repr(simplify(m.T * m - m * m.T))
```
oscarbenjamin, line 493: With random tests like this we need to make sure that when the test fails the error message gives a way to reproduce the failure (e.g. it prints the seed or something). Otherwise if you run the tests once and it fails then you have no way to reproduce that failure afterwards.

sympy/matrices/random.py
jan-philipp-hoffmann: I think `tensorflow` does not support such value assignment. Well yes, `tf` doesn't support any assignment at all. 

But some naive frameworks which simply use lists of lists `[[a,  b], [c, d]]` to represent matrices would take `o[i][j]=v`. Maybe we can use `o[i, j] = v` as you prefer.

sympy/matrices/tests/test_random.py
```diff
@@ -0,0 +1,535 @@
...
+from sympy.matrices import random as _random
+from sympy.testing.pytest import raises
+
+TEST_DIMS = dict((d, tuple(cartes(range(d), range(d)))) for d in range(2, 6))
+TEST_PRECISION = 7
+TEST_EPSILON = 1e-7
+
+phi, psi = symbols('phi psi')
+
+rand.seed(1)
```
jan-philipp-hoffmann, line 22: seed is set here

sympy/matrices/tests/test_random.py
```diff
@@ -0,0 +1,535 @@
...
+        assert _is_eye(m.T * m)
+
+
+def test_unitary():
+    for d in list(TEST_DIMS)[1::3]:
+        i = unitary(d, (1,), length=0)
+        assert _is_eye(i), repr(i)
+
+        m = unitary(d)
+        assert _is_eye(m.H * m, TEST_EPSILON), repr(simplify(m.H * m))
```
jan-philipp-hoffmann, line 470: random test with prints

sympy/matrices/tests/test_random.py
jan-philipp-hoffmann: seed is set once for all tests at the top of the test file
https://github.com/sympy/sympy/pull/22099/files/3375a9ded74f4a3350e74ea1a042c199fc18b138#r757045756
so any failure could be reproduced. 

In case of a failure, one can add some prints as done here already
https://github.com/sympy/sympy/pull/22099/files/3375a9ded74f4a3350e74ea1a042c199fc18b138#r757047379

Since there are plenty of tests (more than 600 lines of code) I don't think it would be efficient to add prints to all of them now. 

But if there are special reason to do this only for test of `normal` I think it would be feasible.

sympy/matrices/random.py
oscarbenjamin: It's possible to just convert between different matrix formats so it's not clear to me what the value is in needing to support so many rather than just saying that the matrix can be constructed in one format and then converted to another.

sonntagsgesicht: Since `sympy` is quite slow, creating in `sympy ` and converting it would not be efficient. Nevertheless, I already removed the `_set_value` ;)

PR #22002, opened 2021-09-01
Docs:  Add Tutorials Category

nanjekyejoannah:
This PR adds docs for the ``tutorials`` category.

#### References to other Issues or PRs

It builds on top of   #22001 

#### Brief description of what is fixed or changed

The docs in the directory `tutorials ` have been reorganized.

#### Release Notes

NO ENTRY

Issue #22007, opened 2021-09-02
Reorganizing the SymPy Documentation


nanjekyejoannah:
## Abstract

SymPy documentation is currently difficult to navigate for users since the key topics are hidden in a big table of contents. After an audit of the documentation, this documentation discusses a way forward for the reorganization of the SymPy documentation. 

## Scope

The main aim of the reorganization is to achieve the following goals:

- Reorganize the documentation following the [Diátaxis documentation structure](https://diataxis.fr/).
- Migrate a few agreed-upon contents from the wiki pages.
- Ensure the documentation pages are responsive.

## Proposed Documentation Structure

As pointed out earlier, we propose using the Diátaxis documentation structure. The structure dictates that the documentation is divided into four main parts; *Getting started, How-to guides, Reference, and Explanation*. According to SymPy's needs, Diátaxis's four-category documentation structure works better than other tools we know of like DITA and Redhat's modular documentation system. The structure gives a clear scope and sets a clear tone on where each piece of information should go,  for both users and documentation authors.

### Content for each of the categories

This is a discussion on what type of SymPy documentation should live in each of the Diátaxis documentation categories.

**Getting Started**

The sections on installation and tutorial from the documentation can be migrated to this category. In addition, we propose collecting any other information and quick links under a common "About SymPy" section, also in this category. See the detailed table of contents for the full layout of sections.

**How-to Guides**

The documentation style guide, including any information on how to write tutorials and documentation, should go in this category.

**Reference**

This category can have the SymPy Modules reference. As advised, we will then have a clear distinction between public and internal APIs.

**Explanation**

The SymPy documentation under Gotchas and Pitfalls should perfectly fit in this category but we can also have the documentation on SymPy special topics here as well.

**Contributor Guide**

This consists of details on environment setup, developer workflow, and style guides to aid comprehension of the SymPy development process.

**Tutorials**

This is a collection of SymPy tutorials.

## A New Sphinx Theme

We also discussed the potential of changing the documentation theme from ``classic``. I proposed the ``Furo`` theme used by CPython, NumPy, and Pandas, etc., but this is still up for decision and suggestions. In my view, this would give us a better layout for seamless navigation of the documentation.

## Progress

- [x] Getting Started (#21747)
- [x] Contributor Guide (#21894 )
- [x] Tutorials (#22002)
- [x] Explanation (#22001)
- [x] How-to Guides (#22102)
- [ ] Public reference (#22209 #22006)
[...]

PR #22102, opened 2021-09-16
Docs: Add How-to Guide Category

nanjekyejoannah:
This PR adds docs for the ``How-to Guides`` category.

It builds on top of   #22002  

The docs in the directories ``guides``, ``contributing`` and ``getting-started`` are moved to the ``how-to guide`` category.

#### Release Notes

NO ENTRY


asmeurer:
It looks like there are merge conflicts here. You might want to redo this so it is just against master instead of based on the other PRs.

doc/src/how_to_guide/index.rst
```diff
@@ -4,7 +4,14 @@
...
+Content
+=======
+
 .. toctree::
    :maxdepth: 2
 
-   ../guides/index.rst
+   getting_started/index.rst
+   contributing/index.rst
+   guides/index.rst
```
asmeurer, line 17: This creates a subsection "SymPy Guides", but it seems to me that they should just be at the top level on this page.

asmeurer: How about instead of changing guides to how_to_guides/guides, we just keep guides. Then we don't need to rename what is already there.

nanjekyejoannah:
See the respective item on [Elaborate on the meaning of the categories](https://github.com/sympy/sympy/issues/22007)

PR #22103, opened 2021-09-16
Added module `sympy.matrices.random`

jan-philipp-hoffmann:

(Re-)opened PR by my academic account. PR was originally opened as PR #22099 by _sonntagsgesicht_

- generates random matrices of various types like
  - orthogonal
  - unitary
  - invertible
  - singular
  - symmetric
  - nilpotent
  - idempotent
  - normal
  - symmetric
  - hermite 
  - elementary
  - rotation
  - reflection

#### Release Notes

* matrices
  * Added random matrix generation, i.e. for orthogonal, unitary, invertible and triangular


doc/src/modules/matrices/random.rst
```diff
@@ -0,0 +1,222 @@
...
+
+Generators on different types of matrices are presented below.
+
+.. note::
+
+   Even for carefully sampled matrices which are elements in topological matrix groups,
+   the sample distribution does not claim to meet any uniform distribution
+   (with respect to the groups Haar measure).
+   So the sampling procedures may be used to produce educational or academic problems
+   but do not generate proper pre-defined distribution.
```
oscarbenjamin, line 31: Also worth mentioning that the matrices are useful for testing algorithms. I expect we can build some good randomised testing from these that would show up bugs in many of the matrix routines.

sympy/matrices/tests/test_random.py
```diff
@@ -0,0 +1,535 @@
...
+    return _is_zeros(m - eye(*m.shape), precision)
+
+
+def _is_triangular(m, precision=None):
+    s, t = m.shape
+    if precision is None:
+        return all(m[i, j] == 0 for i in range(s) for j in range(i))
+    else:
+        n = m.evalf()
+        return all(abs(n[i, j]) < precision for i in range(s) for j in range(i))
```
oscarbenjamin, line 42: There is an `allclose` function in sympy/testing/matrices.py .

sympy/matrices/random.py
```diff
@@ -0,0 +1,2292 @@
...
+          'permutation', 'elementary', 'rotation', 'reflection', \
+          'diagonal_normal', 'jordan_normal', 'isometry_normal', \
+          'triangular', 'invertible', 'singular', \
+          'idempotent', 'nilpotent', 'diagonalizable', 'trigonalizable', \
+          'orthogonal', 'unitary', 'normal', \
+          'symmetric', 'hermite', \
+          'regular_to_singular', 'complex_to_real', 'rand'
+
+""" random generator """
+rand = Random()
```
oscarbenjamin, line 20: Is this thread-safe?

doc/src/modules/matrices/random.rst
```diff
@@ -0,0 +1,222 @@
...
+Random Generator
+----------------
+
+The module uses its own random generator ``rand`` which is a instance of ``random.Random()``.
+Hence, the state can be set by invoking ``rand.seed()``.
+
+
+    >>> from sympy.matrices.random import rand, orthogonal
+    >>> rand.seed(1)
+    >>> orthogonal(3)
```
oscarbenjamin, line 197: I think it would be better if there was a way of seeding that didn't require global state. For example maybe a function like `orthogonal` should be able to take a seed or a RNG as a parameter like:
```python
>>> from random import Random
>>> rng = Random()
>>> orthogonal(3, seed=rng)
>>> orthogonal(3, seed=1)
```
We should probably try to make this consistent with other random functions in sympy if possible. There has been a lot of discussion about how to do seeding in the stats module.

When the user does not supply a seed then I think that sympy should have a centralised rng to use for all things. There are long open issues about this - I'm not proposing that you need to go around and fix those issues across the codebase but rather than adding a module-specific seed I think it would be better to add a centralised seed for all of sympy and use that here.

doc/src/modules/matrices/random.rst
```diff
@@ -0,0 +1,222 @@
...
+ - linear isometries
+ - non-standard geometries by non-standard scalar products
+ - orthogonal basis of vector space,
+
+specific types of matrices are required.
+
+A purely random generation of matrix entries does not grant any
+specific shape or property of the matrix.
+So, those matrices have to be created carefully using
+classical classification theorems of linear algebra.
```
oscarbenjamin, line 21: This could reference the existing `randMatrix` function:
```python
In [4]: randMatrix(5)
Out[4]: 
⎡86  75  27  10  38⎤
⎢                  ⎥
⎢45  93  13  23  13⎥
⎢                  ⎥
⎢32  13  35  16  15⎥
⎢                  ⎥
⎢97  24  89  10  7 ⎥
⎢                  ⎥
⎣2   93  34  31  46⎦
```

doc/src/tutorial/matrices.rst
```diff
@@ -231,6 +231,149 @@ filled with `0`\ s.
...
+    ⎢          ⎥
+    ⎣0   0   1 ⎦
+    >>> M.inv()
+    ⎡1   -1  0 ⎤
+    ⎢          ⎥
+    ⎢-2  1   -1⎥
+    ⎢          ⎥
+    ⎣0   0   1 ⎦
+
+    >>> M = invertible(3, scalar_set=(sqrt(2),))  # entries as sums and products of sqrt(2) and 1
```
oscarbenjamin, line 263: I'm not sure about `scalar_set`. I think perhaps a domain argument could be good. Ideally there should be some way for the user to pass in some distribution of expressions that they have designed themselves.

oscarbenjamin: I previously wrote some code for random matrix generation that I didn't finish although my emphasis was more on generating different kinds of matrix expressions and in particular working with domain elements. I wonder if an approach like that can be used here:
```python
from pytest import raises

from random import randint, randrange, random, sample, Random
from itertools import product

from sympy import QQ, ZZ, GF, ZZ_I, QQ_I, sqrt
from sympy.abc import x, y
from sympy.polys.matrices import DomainMatrix



class RandomDomain:
    """Superclass for random element generators"""
    rng = Random()

    def __call__(self, rng=None):
        return self.generate(rng or self.rng)


class RandomGF(RandomDomain):
    """Random finite field element GF(p)"""
    def __init__(self, K):
        self.domain = K

    def generate(self, rng):
        return self.domain(rng.randint(0, self.domain.mod))


class _RandomZZ(RandomDomain):
    """Base class for RandomZZ and RandomZZpos"""

    domain = ZZ

    def __init__(self, bits):
        self.bits = bits
        self.intmax = 1 << bits - 1


class RandomZZ(_RandomZZ):
    """Uniform random integer with maximum bitsize"""
    def generate(self, rng):
        return ZZ(rng.randint(-self.intmax, self.intmax))


class RandomZZpos(_RandomZZ):
    """Uniform random positive integer ZZ with maximum bitsize"""
    def generate(self, rng):
        return ZZ(rng.randint(1, self.intmax))


class RandomZZlog(_RandomZZ):
    """Random positive integer ZZ the logarithmic distribution"""
    def generate(self, rng):
        logbits = 0
        bits = self.bits
        while bits:
            bits //= 2
            logbits += 1
        bits = rng.randint(1, 1 << logbits - 1)
        return ZZ(rng.randint(0, 1 << bits - 1))


class RandomQQ(RandomDomain):
    """Random rational QQ with given num/den distribution"""

    domain = QQ

    def __init__(self, numrand, denrand):
        self.numrand = numrand
        self.denrand = denrand

    @classmethod
    def bits(cls, bits):
        """RandomQQ with uniform num/den distribution and max bits"""
        return RandomQQ(RandomZZ(bits), RandomZZpos(bits))

    def generate(self, rng):
        return QQ(self.numrand(rng), self.denrand(rng))


class _RandomGaussian(RandomDomain):
    """Base class for RandomZZ_I and RandomQQ_I"""
    def __init__(self, rand_x, rand_y=None):
        rand_y = rand_y or rand_x
        self.rand_x = rand_x
        self.rand_y = rand_y

    def generate(self, rng):
        return self.domain(self.rand_x(rng), self.rand_y(rng))


class RandomZZ_I(_RandomGaussian):
    """Random Gaussian integer ZZ_I with given x,y distribution"""
    domain = ZZ_I


class RandomQQ_I(_RandomGaussian):
    """Random Gaussian rational QQ_I with given x,y distribution"""
    domain = QQ_I


class RandomALG(RandomDomain):
    """Random element of an algebraic number field QQ(a)"""
    def __init__(self, K, rand_coeff):
        self.domain = K
        self.degree = K.mod.degree()
        self.rand = rand_coeff

    def generate(self, rng):
        coeffs = [self.rand(rng) for _ in range(self.degree + 1)]
        return self.domain.new(coeffs)


class RandomPolyRing(RandomDomain):
    """Random polynomial in K[x,y]"""
    def __init__(self, R, rand_coeff, degree):
        self.domain = R
        self.rand = rand_coeff
        self.gens = R.gens
        self.degree = degree
        self.monomials = list(product(range(degree+1), repeat=len(R.gens)))

    def generate(self, rng):
        gens = self.gens
        poly_d = {m: self.rand(rng) for m in self.monomials}
        return self.domain.new(poly_d)


class RandomFracField(RandomDomain):
    """Random element of a fraction field K(x,y)"""
    def __init__(self, K, rand_poly_num, rand_poly_den=None):
        self.domain = K
        self.rand_num = rand_poly_num
        self.rand_den = rand_poly_den or rand_poly_num

    def generate(self, rng):
        num = self.rand_num(rng)
        den = self.rand_den(rng)
        return self.domain(num) / self.domain(den)


class RandomDDM(RandomDomain):
    """Random dense DomainMatrix"""
    def __init__(self, rand_element, shape):
        self.rand = rand_element
        self.shape = shape
        self.domain = rand_element.domain

    def generate(self, rng):
        rand = self.rand
        rows, cols = self.shape
        elements = [[rand(rng) for _ in range(cols)] for _ in range(rows)]
        dM = DomainMatrix(elements, self.shape, self.domain)
        return dM


class RandomSDM(RandomDomain):
    """Random sparse DomainMatrix"""
    def __init__(self, rand_element, shape, density):
        self.rand = rand_element
        self.shape = shape
        self.density = density
        self.domain = rand_element.domain

    def generate(self, rng):
        rand = self.rand
        rows, cols = self.shape
        num_nonzero = round((rows * cols) * self.density)
        indices = rng.sample(range(rows * cols), num_nonzero)
        sdm = {}
        for ij in indices:
            i, j = divmod(ij, cols)
            val = rand(rng)
            try:
                sdm[i][j] = val
            except KeyError:
                sdm[i] = {j: val}
        dM = DomainMatrix(sdm, self.shape, self.domain)
        return dM


def random_domain_generator(domain, **kwargs):
    """Make a random generator for the given domain"""
    bits = kwargs.get('bits', 6)
    degree = kwargs.get('degree', 2)
    logarithmic = kwargs.get('logarithmic', False)

    if domain.is_FiniteField:
        return RandomGF(domain)

    elif domain == ZZ:
        if not logarithmic:
            return RandomZZ(bits)
        else:
            return RandomZZlog(bits)

    elif domain == QQ:
        return RandomQQ.bits(bits)

    elif domain == ZZ_I:
        return RandomZZ_I(RandomZZ(bits))

    elif domain == QQ_I:
        return RandomQQ_I(RandomQQ.bits(bits))

    elif domain.is_AlgebraicField:
        return RandomALG(domain, RandomQQ.bits(bits))

    elif domain.is_PolynomialRing:
        rand_coeff = random_domain_generator(domain.domain, **kwargs)
        return RandomPolyRing(domain, rand_coeff, degree)

    elif domain.is_FractionField:
        rand_poly = random_domain_generator(domain.get_ring(), **kwargs)
        return RandomFracField(domain, rand_poly)

    else:
        raise ValueError


def random_element(domain, rng=None, **kwargs):
    """Return a random element from the given domain"""
    rand = random_domain_generator(domain, **kwargs)
    return rand(rng)


def random_matrix(domain_or_rand, shape, *, rng=None, density=1, **kwargs):
    """Random matrix with elements of given domain"""
    rows, cols = shape

    if isinstance(domain_or_rand, RandomDomain):
        rand = domain_or_rand
    else:
        domain = domain_or_rand
        rand = random_domain_generator(domain, **kwargs)

    if density == 1:
        rand_DM = RandomDDM(rand, shape)
    else:
        rand_DM = RandomSDM(rand, shape, density)

    dM = rand_DM(rng)

    return dM


def test_rand_element():
    seed = Random().randint(1, 10**6)
    rng1 = Random(seed)
    rng2 = Random(seed)

    domains = [
        ZZ, QQ, GF(7), ZZ_I, QQ_I,
        QQ.algebraic_field(sqrt(2)),
        ZZ[x], QQ[x],
        ZZ[x,y], QQ[x,y],
        ZZ.frac_field(x), QQ.frac_field(x),
        ZZ.frac_field(x,y), QQ.frac_field(x,y),
    ]
    for K in domains:
        e = random_element(K)
        assert K.of_type(e)

        assert random_element(K, rng=rng1) == random_element(K, rng=rng2)

        for density in [0.5, 1]:
            for r in [K, random_domain_generator(K)]:
                dM = random_matrix(r, (3, 3), density=density)
                assert isinstance(dM, DomainMatrix)
                assert dM.domain == K

                dM1 = random_matrix(r, (3, 3), density=density, rng=rng1)
                dM2 = random_matrix(r, (3, 3), density=density, rng=rng2)
                assert dM1 == dM2

    from sympy.polys.domains import Domain
    class Undefined(Domain):
        def __init__(self):
            pass
    K = Undefined()
    raises(ValueError, lambda: random_domain_generator(K))
```

oscarbenjamin: The main point is that with someone can design their own random expression generator and pass it in to `random_matrix` to get a matrix generated from that distribution of expressions.

jan-philipp-hoffmann: I think, I get the idea. Note, there are mainly two different ways a rng is used. 

1. for sampling _dimensions_/_shapes_, _indices_ or _sign_ in `sgn * sqrt(1 - c**2)` which are probably uniform distributed 
2. for sampling matrix entries. But here, one has to distinguish between 
   - _isometries_ which take elements of the circle of complex units `scalar_set` as an argument
   - _elementary_  which takes elements `scalar_set`  **and** units `unit_set` of a ring as arguments
   - _normal forms_ which take eigenvalues `spec` as arguments

My suggestion would be, let `scalar_set`,  `unit_set` and `spec` possibly be a rng, too. If so, they should implement a `sample(k)` method. Otherwise (e.g. if just a `list` or `tuple`) they will be sampled using the _default rng_ `rand.sample()` which is given by an extra _seed_ argument.

Here what it could look like:
```python
def diagonal_normal(dim, spec=None, seed=None):
    """ diagonal normal form"""
    
    rng = seed if has_attr(seed, 'sample') else random.Random(seed)

    if has_attr(spec, 'sample'):
        spec = tuple(spec.sample(1) for _ in range(dim))
    elif not len(spec)==dim:
        spec = tuple(rng.sample(spec, 1) for _ in range(dim))
```

Then the matrix entries can come from some `RandomDomain` or just a finite `list`. The `list` is very handy to draw easy problems for undergraduate _linear algebra_ courses. Note, even a few elements could give a quite huge matrices. Arbitrary rationals would give very easy unsolvable problems. 

```
>>> orthogonal(3, scalar_set=(0, 1/2, sqrt(2)/2, sqrt(3)/2))
⎡   ⎛  1   √2⎞      ⎛√2   1⎞             ⎛√2   1⎞      ⎛  1   √2⎞⎤
⎢√2⋅⎜- ─ + ──⎟   √2⋅⎜── + ─⎟          √2⋅⎜── + ─⎟   √2⋅⎜- ─ + ──⎟⎥
⎢   ⎝  2   4 ⎠      ⎝4    2⎠             ⎝4    2⎠      ⎝  2   4 ⎠⎥
⎢───────────── + ───────────  1/2   - ─────────── + ─────────────⎥
⎢      2              2                    2              2      ⎥
⎢                                                                ⎥
⎢   ⎛  1   √2⎞      ⎛√2   1⎞             ⎛  1   √2⎞      ⎛√2   1⎞⎥
⎢√2⋅⎜- ─ + ──⎟   √2⋅⎜── + ─⎟          √2⋅⎜- ─ + ──⎟   √2⋅⎜── + ─⎟⎥
⎢   ⎝  2   4 ⎠      ⎝4    2⎠             ⎝  2   4 ⎠      ⎝4    2⎠⎥
⎢───────────── + ───────────  1/2   - ───────────── + ───────────⎥
⎢      2              2                     2              2     ⎥
⎢                                                                ⎥
⎢            √2               -√2                                ⎥
⎢            ──               ────                0              ⎥
⎣            2                 2                                 ⎦

>>> orthogonal(3, length=10)
⎡                 √2   1               1   √2      ⎤
⎢  -1/2           ── + ─               ─ - ──      ⎥
⎢                 4    2               2   4       ⎥
⎢                                                  ⎥
⎢                    ⎛  1   √2⎞     ⎛  1   √2⎞     ⎥
⎢                 √2⋅⎜- ─ - ──⎟  √2⋅⎜- ─ - ──⎟     ⎥
⎢ 1   √2     √2      ⎝  2   4 ⎠     ⎝  2   4 ⎠   √2⎥
⎢ ─ - ──   - ── - ─────────────  ───────────── - ──⎥
⎢ 2   4      4          2              2         4 ⎥
⎢                                                  ⎥
⎢                     ⎛1   √2⎞             ⎛1   √2⎞⎥
⎢                  √2⋅⎜─ - ──⎟          √2⋅⎜─ - ──⎟⎥
⎢  1   √2     √2      ⎝2   4 ⎠     √2      ⎝2   4 ⎠⎥
⎢- ─ - ──   - ── - ───────────   - ── + ───────────⎥
⎣  2   4      4         2          4         2     ⎦
```

jan-philipp-hoffmann: btw, no problem to rename `scalar_set` and `unit_set` arguments by `domain` and `units`. 

As mentioned above, it might be more consistent to use `units` instead of `scalar_set` for isometries like `orthogonal` or `unitary` (as `cos/sin` are just representing a point on the _circle of complex units_).

oscarbenjamin: I'm not sure what is a good interface for seeding here. What you suggest sounds reasonable.

jan-philipp-hoffmann: Me either. I've been thinking more than a year.. 

Nevertheless latest version offers at least most flexibility - by the cost of heavily overloading the `seed` argument. Hope it is intuitive enough. 

To explain it, I added a extra section in [`doc/src/modules/matrices/random.rst`](https://github.com/jan-philipp-hoffmann/sympy/blob/pre_master/doc/src/modules/matrices/random.rst).

oscarbenjamin:
Sorry I lost track of this. I see now that many changes were made and will try to review them at some point.

oscarbenjamin:
I think it might be confusing that this uses `domain` in a way that is inconsistent with the existing usage of "domain" in sympy to return to implementations of rings and fields. The domain argument here is a list that generates the elements in some way so perhaps something like "basis" is a better name, or perhaps just `scalar_set` as you had it before. I'm not sure but I think "domain" would be confusing in the context of sympy.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2466 @@
...
+              "may be supplied, but not %i"
+        raise ValueError(msg % len(domain))
+    return obj
+
+
+# === fundamental matrix functions ===
+
+
+def complex_to_real(mat=None):
+    r""" returns real version of a complex matrix
```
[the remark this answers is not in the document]
jan-philipp-hoffmann: Correct, it does not split real and imaginary parts. It gives the real matrix of a complex homomorphism where the complex basis `b_1,...,b_n` is turned into a real basis `b_1,b_1*I,...,b_n,b_n*I` of dimension `2*n`. It is used to turn multiplication with a complex root of unity into a real rotation.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2466 @@
...
+            raise ValueError(
+                "if row==col no further args than value may be supplied.")
+    elif not domain:
+        # elementary
+        if value is None:
+            # transposition
+            _set_value(obj, row, row, 0)
+            _set_value(obj, row, col, 1)
+            _set_value(obj, col, row, 1)
+            _set_value(obj, col, col, 0)
```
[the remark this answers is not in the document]
jan-philipp-hoffmann: As I mentioned in https://github.com/sympy/sympy/pull/22099#discussion_r709255948

jan-philipp-hoffmann:
Sorry, didn't forget ist bus was busy with my lectures. Try to comeback to the project asap.

oscarbenjamin:
No need to apologise but thanks for chiming in. Some PRs get abandoned so it's good to know that it is still expected to be worked on.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2466 @@
...
+            0 & 1 & 0 & 0 \\
+            0 & 0 & 2 & 0 \\
+            0 & 0 & 0 & 2
+        \end{array}\right]
+
+    Alternativly, a sequence of pairs *(eigenvalue, block size)*,
+    i.e. tuple of length 2, can be provided.
+
+    So the above example is equivalent to
+    **spec** = [(1,2),(2,1),(2,1)].
```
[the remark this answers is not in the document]
jan-philipp-hoffmann: I agree that the second one is clearer. That's the reason why I implemented it. But the first one is more inline the previous concept of setting eigenvalues. And it is very handy if you want to quickly setup a Jordan matrix. So, why not keeping both?

sympy/matrices/random.py
```diff
@@ -0,0 +1,2466 @@
...
+    >>> m.det()
+    1
+    >>> m.inv()
+    Matrix([
+    [ 1, 0, phi],
+    [-1, 1,   1],
+    [ 0, 0,   1]])
+
+    """
+    return square(dim, None, domain, units, length, seed=seed)
```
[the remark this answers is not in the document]
jan-philipp-hoffmann: Tried to make this more clear by explicitly referencing to `square`. Same for `invertible`.

sympy/matrices/random.py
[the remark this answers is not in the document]
jan-philipp-hoffmann: I'll swap it in the doc. So the clearer interface will come first.

sympy/matrices/random.py
jan-philipp-hoffmann: I did search for a common interface but found more than on. Finally, I switched to a flexible but hopefully intuitiv version https://github.com/sympy/sympy/pull/22103#discussion_r711015017

jan-philipp-hoffmann:
Finally, I incorporated most of your recommendations. 

I tried to make the docs clearer specially when it comes to the sublime differences of `square` and `invertible` as well as `square` and `singular`.
Also changed `domain` into `scalars`. 

Only kept the `_get_value`/`_set_value` functions as well as both ways to set Jordan blocks via `spec`  in `jordan_normal`.  Nevertheless I tried to make the reasons for both clear in the documentation.

Hope, we come close to somewhat that could become a finale version.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2467 @@
...
+        v = orthogonal(dim,
+                       scalars=scalars, length=length - half, seed=rng)
+        return _multiply(u, d, v)
+
+    normal_form = isometry_normal(dim, spec, real=False, seed=rng)
+    s = unitary(dim, scalars=scalars, length=length, seed=rng)
+    return _multiply(_c(s), normal_form, s)
+
+
+def normal(dim,
```
oscarbenjamin, line 2220: When I test this function it always seems to return a symmetric matrix. I would have thought that symmetric matrices could be made more easily than this though.

jan-philipp-hoffmann: Yes, but only if all _spec_ entries are real, this gives a symmetric matrix `M`. But this one is different from simply some `A*A.T` since `M` has a given set of eigenvalues (as set by _spec_). Note, those ev must all be unit roots.

If any _spec_ entry is complex, this yields a complex normal matrix `M`, i.e. `M.H*M=M*M.H` (which is true for symmetric matrices). Again, with a given set of eigenvalues.

oscarbenjamin: Maybe the docstring could be clearer about what this expected to return

jan-philipp-hoffmann: I tried to improve the docs on _unitary_ and _normal_ matrices.

doc/src/modules/matrices/random.rst
```diff
@@ -0,0 +1,371 @@
...
+
+    >>> M.inv()
+    Matrix([
+    [1, 0, 2],
+    [1, 1, 1],
+    [0, 0, 1]])
+
+
+Seeding Random Number Generators
+--------------------------------
```
oscarbenjamin, line 194: In the time since this PR was created a new `sympy/core/random` module has been added so that there is centralised control of seeding for random function in SymPy:
https://github.com/sympy/sympy/blob/4489b4c7c51a65966b1a95f2af6f7e0c8c461794/sympy/core/random.py#L1-L10

sympy/matrices/random.py
```diff
@@ -0,0 +1,2525 @@
...
+    else:
+        smpl = rand.sample(scalars, _k)
+    return smpl if k else smpl[0]
+
+
+# === helper ===
+
+
+def _set_value(o, i, j, v):
+    o[i, j] = v
```
oscarbenjamin, line 79: This function doesn't seem to be used any more

sympy/matrices/random.py
```diff
@@ -0,0 +1,2525 @@
...
+            msg += " or - if real - norm less than 1"
+            msg += " not abs%s=%s" % (str(scalar), str(abs_cs))
+            raise ValueError(msg)
+    return (c, s) if real else (c + s * _i,)
+
+
+def _is_complex(z):
+    return isinstance(z, complex) or \
+           (getattr(z, 'is_complex', False) is True and
+            getattr(z, 'is_real', True) is False)
```
oscarbenjamin, line 103: The `is_complex` and `is_real` attributes are part of the assumptions system. I'm not sure exactly what the intended purpose of this check is but `is_real` might not be what is wanted here.

Usually in SymPy it isn't necessary to try to handle built in types at the same time as SymPy types because wen just use sympify to turn everything into SymPy types e.g.:
```python
In [1]: from sympy.core.sympify import _sympify

In [2]: _sympify(1j)
Out[2]: 1.0⋅ⅈ

In [3]: _sympify(1j).is_real
Out[3]: False
```

sympy/matrices/random.py
```diff
@@ -0,0 +1,2525 @@
...
+    transposition
+
+    """
+    rng = _rand(seed)
+    perm = perm or _sample(range(dim), dim, rng=rng)
+    obj = _eye(dim)
+    for i, j in enumerate(perm):
+        obj[i, i] = 0  # aka zeros(dim)
+        obj[i, j] = 1
+    return obj  # aka _eye(dim).permute(perm)
```
oscarbenjamin, line 657: Could this just use `eye(dim).permute(perm)`?

sympy/matrices/random.py
jan-philipp-hoffmann: yes, just forgot to remove it.

sympy/matrices/random.py
jan-philipp-hoffmann: Could, but would work for `sympy` only. neither `numpy` nor `mpmath` support `eye(dim).permute(perm)`.

oscarbenjamin: I have to say I don't really understand the desire to support code here (in the SymPy codebase) that should work with the matrix classes from the other codebases. The obvious way to get a numpy array here would be to compute the sympy Matrix and then convert it to a numpy array at the end. Is there a reason that support for numpy matrices can not just work like that?

The main advantage of using numpy in the internal calculations is lost when most of the code that does the setting of individual elements is done by non-vectorised pure Python code. On the other hand if we were to make performance improvements to sympy's  `permute` functions then this code would not benefit from those because it reimplements them rather than reusing them.

Also how would I even use this function with those other libraries?

jan-philipp-hoffmann: 1. Frist of all, this module wouldn't exists as I started with `numpy`. 
2. Second, using different framework is a good way of testing. If same code works for multiple frameworks, the implementation is kind of generic. 
3. Third, in my opinion the current implementation fits into coding style of the remaining module. The `eye(dim).permute(perm)` approach would be kind of singular compared to the other functions.
4. Forth, a dependency via `eye(dim).permute(perm)` would not be in favour of the module. Assume there will be changes on `eye(dim).permute(perm)`, this doesn't mean `permutation(dim, perm)` should change the same way, too.

To run this code with another framework, just replace all lines before line 120 with your framework wrapper and use `eval`.

sympy/matrices/random.py
jan-philipp-hoffmann: Good hint, thanks.

doc/src/modules/matrices/random.rst
jan-philipp-hoffmann: This dosen't work properly. See #22592.

jan-philipp-hoffmann:
Here is an update on `sympy.matrices.random` (PR #22103) to adapt the new `sympy.core.random` framework.

In order to fix issue #22592 the PR incorporates some changes based on the ideas of @oscarbenjamin and @mohajain in #22649 plus some tests which prove the issue is really fixed.

Moreover, all signatures, docs and tests are reworked. Finally all the stuff seems to work together well.

.mailmap
```diff
@@ -1497,6 +1498,7 @@ shruti <shrutishrm512@gmail.com>
 shubhayu09 <guptashubhayu601@gmail.com> shubhayu09 <87441886+shubhayu09@users.noreply.github.com>
 sirnicolaf <43586954+sirnicolaf@users.noreply.github.com>
 slacker404 <pchantza@gmail.com>
+sonntagsgesicht <sonntagsgesicht@icloud.com>
```
oscarbenjamin, line 1501: Is this a different person from Jan-Philipp Hoffman?

Adding a separate entry like this implies that it is and that two separate names should be added to the AUTHORS file.

jan-philipp-hoffmann: No,  _sonntagsgesicht_ is not a different person from _Jan-Philipp Hoffmann_. But since there are commits form both accounts, the tests required both mentioned in the AUTHORS file. 
If there is any way to drop _sonntagsgesicht_, it would be fine. But I don't know how.

oscarbenjamin: One way would be to squash the commits and add a single commit under one author name.

On the other hand if sonntagsgesicht is a different person and did author part of this work then I would have thought that they should be credited although they might prefer for their real name to be listed in the AUTHORS file. If that is the case then instead of the above a line can be added like:
```
Real Name <preferred email address> sonntagsgesicht <sonntagsgesicht@icloud.com>
```
That tells `git` and the `mailmap_check` script that commits with the latter metadata should be attributed to Real Name and ensures that Real Name will be listed in the AUTHORS file.

oscarbenjamin: Sorry, I missed the word "not" above.

jan-philipp-hoffmann: As I'm not familiar to squash commits and really afraid to ruin the whole pull request, so I preferred to fix the `.mailmap` file.  Therefore I changed my entries to

```
Jan-Philipp Hoffmann <sonntagsgesicht@icloud.com> Jan-Philipp Hoffmann <90828785+jan-philipp-hoffmann@users.noreply.github.com>
Jan-Philipp Hoffmann <sonntagsgesicht@icloud.com> sonntagsgesicht <sonntagsgesicht@icloud.com>
```

Sorry, if this doesn't work properly.  Please help me, how to fix this. I'm a bit lost.

oscarbenjamin: I think that the .mailmap file is correct now.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2421 @@
...
+          'permutation', 'elementary', 'rotation', 'reflection', \
+          'diagonal_normal', 'jordan_normal', 'isometry_normal', \
+          'triangular', 'invertible', 'singular', \
+          'idempotent', 'nilpotent', 'diagonalizable', 'trigonalizable', \
+          'orthogonal', 'unitary', 'normal', \
+          'symmetric', 'hermite', 'square', \
+          'regular_to_singular', 'complex_to_real'
+
+
+_ALT = False
```
oscarbenjamin, line 21: There should be some comment explaining what `_ALT` is. I've looked through the code and I don't understand its purpose.

Probably this variable should just be removed. If there is a reason to choose between the different codepaths then it should not be done using a global variable.

jan-philipp-hoffmann: Good point. Removed `_ALT` and moved alternative implementations to `test_random.py` as it is used in testing different codepaths.

oscarbenjamin:
This looks generally good and also better than it was when I previously reviewed. It will take some time for me to go through it in detail though.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2378 @@
...
+# === helper ===
+
+
+def _cs(scalar, real=True):
+    if isinstance(scalar, (tuple, list)) and len(scalar) == 2:
+        c, s = scalar
+    elif _is_complex(scalar):
+        c, s = _re(scalar), _im(scalar)
+    else:
+        c, s = scalar, _sample((-1, 1)) * _sqrt(1 - scalar ** 2)
```
oscarbenjamin, line 45: This function name is quite cryptic. It could do with at least comments or a docstring but probably a more descriptive name e.g. one that describes what it returns or does.

The signature is quite confusing since it accepts either a tuple or a "scalar" and then returns either a tuple or a complex number. The function is called in only two places only one of which uses the `real` argument which makes me wonder if it really makes sense for this combination of functionality to be grouped together in a single function.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2378 @@
...
+    abs_cs = abs(c ** 2 + s ** 2 - 1)
+    if isinstance(scalar, (int, float, complex)):
+        if abs(scalar) > 1 or abs_cs > _EPS:
```
oscarbenjamin, line 48: I think that this can just be `if (c**2 + s**2 - 1).is_zero is False:` or using the internal `is_ne` function `if is_ne(c**2 + s**2, 1):`

sympy/matrices/random.py
```diff
@@ -0,0 +1,2378 @@
...
+
+    """
+    obj = _eye(dim)
+    if index is None:
+        return obj
+
+    if not isinstance(index, (list, tuple, set)) or not len(index) == 2:
+        raise ValueError(
+            "index argument must be tuple of two matrix index integer.")
+    row, col = index
```
oscarbenjamin, line 179: I don't think that a set should be allowed here if the intention is to iterate over index.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2378 @@
...
+    obj = super_elementary_matrix(2 * dim)
+    for i in range(dim):
+        for j in range(dim):
+            z_value = mat[i, j]
+            a, b = _re(z_value), _im(z_value)
+            obj[2 * i, 2 * j] = a
+            obj[2 * i, 2 * j + 1] = b
+            obj[2 * i + 1, 2 * j] = -b
+            obj[2 * i + 1, 2 * j + 1] = a
+    return obj
```
oscarbenjamin, line 282: It should be possible to do this with something like:
```python
M = Matrix([[a + I*b, 0], [1, I]])

Mr, Mi = M.as_real_imag()

Mplus = Mr + Mi
Mminus = Mr - Mi
Mnew = zeros(2*Mr.rows, 2*Mr.cols)
Mnew[::2,::2] = Mminus
Mnew[1::2,1::2] = Mminus
Mnew[1::2,::2] = -Mplus
Mnew[::2,1::2] = Mplus
```
It is preferable to use higher level operations like addition, slicing etc because they can be optimised for sparse matrices.

The code shown above doesn't work because of a bug though.

oscarbenjamin: > The code shown above doesn't work because of a bug though.

I mean that the code that I wrote does not work because of a bug:
```python
In [51]: M = zeros(4, 4)

In [52]: M[::2,::2] = ones(2, 2)
ShapeError 
```

sympy/matrices/random.py
```diff
@@ -0,0 +1,2378 @@
...
+        return mat
+    dim = _shape(mat)[0]
+    if rank == dim:
+        return mat
+    i = _eye(dim)
+    p = permutation(dim)
+    d = projection(dim, (0, rank))
+    d = _multiply(p.inv(), d, p)
+    md = _multiply(mat, d)
+    return md + _multiply(md, mat, i - d)
```
oscarbenjamin, line 346: Unless I misunderstand the `_multiply` function here is `Mul` (it's confusing to give it a different name) which is really just a roundabout way of using `*` i.e. the above can just be:
```python
d = p.inv() * d * p
md = mat * d
return md + md * mat * (i - d)
```

sympy/matrices/random.py
```diff
@@ -0,0 +1,2378 @@
...
+    diagonal_normal
+
+    """
+
+    index = index or _sample(range(dim + 1), 2)
+    obj = _eye(dim)
+    start, end = sorted(index)
+    for i in range(dim):
+        v = 1 if start <= i < end else 0
+        obj[i, i] = v
```
oscarbenjamin, line 403: Why not just loop like `for i in range(start, end)`?

On the other hand it would seem more natural to me for this function to take a list of indices rather than a start stop tuple.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2378 @@
...
+    return obj
+
+
+def jordan(dim,
+           index=None,
+           scalar=None):
+    r"""n x n matrix with a single Jordan block of given eigenvalue
```
oscarbenjamin, line 410: I find this function slightly odd. It returns an identity matrix with a given Jordan block. It would make more sense to me to have a function that takes a list of `(eigenvalue, blocksize)` pairs e.g.:
```
>>> jordan([(2, 2), (3, 1)])
⎡2  1  0⎤
⎢       ⎥
⎢0  2  0⎥
⎢       ⎥
⎣0  0  3⎦
```
The only random element here seems to be choosing the eigenvalue for the nontrivial Jordan block. Maybe that could be done by only specifying the sizes of the blocks e.g. `jordan([2, 1])`.

sympy/matrices/random.py
jan-philipp-hoffmann: You're right the functions signature is a mess. I reworked depending functions like `isometry_normal` and `unitary` to rely only on `rotation`. So I was able to moved the remaining parts of `_cs` into `rotation`. Finally, I was able to remove the `real` argument.

Hope this makes the concept of the whole module more clear. 
As it constructs _high level matrices_ (like `orthogonal`, `unitary` or `normal`) by _conjugating_ a _normal form_ (here `isometry_normal`) with a product of _basis matrices_ (of type `rotation`).

And now the default behaviour of `scalars` arguments are set in `rotation`.

sympy/matrices/random.py
jan-philipp-hoffmann: simplified this part by removing the `abs_cs > _EPS` condition. But something similar happens in `_is_abs_one`. Currently I use buildins `isclose` as a strikt compare fails for
```
>>> z = complex(1, 2)
>>> abs(z/abs(z)) == 1
False
```

jan-philipp-hoffmann: If there is any way to improve this, I would be happy to adopt it.

sympy/matrices/random.py
jan-philipp-hoffmann: removed `set` here.

sympy/matrices/random.py
jan-philipp-hoffmann: Slices seem not to work but I came up with a different approach permuting block matrices
```
    a, b = mat.as_real_imag()
    M = zeros(2 * dim)
    M[0, 0] = a
    M[0, dim] = b
    M[dim, 0] = -b
    M[dim, dim] = a

    p = list()
    for i in range(dim):
        p.extend([i, dim + i])
    M = M.permute(p).T.permute(p).T
```

sympy/matrices/random.py
jan-philipp-hoffmann: Fine, removed all renaming of sympy function and names, e.g. use now `I` instead of `_i` and `Mul` or `*` instead of `_multiply`. Hope this makes the code less misunderstanding.

sympy/matrices/random.py
jan-philipp-hoffmann: As the module has the concept of _basis types matrices_ which usually operate on a single subspace of a vector space (hence have a single non trivial block). Here `jordan` is such basic type and acts on a single [space of generalized eigenvectors](https://en.wikipedia.org/wiki/Generalized_eigenvector).

In order to keep the signature consistent for all basic matrices, `index` sets start and end indices of the non trivial block.

Multiple Jordan blocks come into play with the `jordan_normal` matrix which are build by blocks of _basis type matrices_. Same as `isometry_normal` is build by multiple `rotation` blocks. The `spec` argument of _normal form matrices_ follows exactly the pattern `(eigenvalue, blocksize)` you suggest.

sympy/matrices/random.py
jan-philipp-hoffmann: same for `_is_complex`

oscarbenjamin: We have:
```
def _is_abs_one(x):
    return isinstance(abs(x), (int, float, Number)) and isclose(abs(x), 1)
```
This is basically just `x in [-1, 1]` if I understand correctly. Why do we need the `isclose`? There shouldn't be any rounding error.

oscarbenjamin: Oh, I see the idea is that if `x` is complex then `abs(x)` could be float.

In general it is better to work with sympified objects rather than having `int`, `float` or `complex`. The idea is that when receiving input from the user you call `_sympify` on it and then you know everywhere internally that you have symbolic expressions.

sympy/matrices/random.py
oscarbenjamin: The slicing assignment should be fixed. I'll open a PR for that.

sympy/matrices/random.py
oscarbenjamin: With exact symbolic expressions there shouldn't be any rounding error so no need for `isclose` etc.

oscarbenjamin: I'm not sure what the purpose of `_is_complex` is supposed to be.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+    ========
+    transposition
+
+    """
+
+    perm = perm or _sample(range(dim), dim)
+    obj = zeros(dim)
+    for i, j in enumerate(perm):
+        obj[i, j] = 1
+    return obj  # aka eye(dim).permute(perm)
```
oscarbenjamin, line 568: Is there a reason this shouldn't just use `eye(dim).permute(perm)`?

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+    * $a=1$, $d=\lambda$ and $b=0=-c$ this matrix describes scaling the
+      row $j$ by $\lambda$
+    * $a=1=d$, $b=\mu$, $-c=0$ adding the $\mu$ multiple
+      of the row $i$ to row $j$.
+
+    Examples
+    ========
+    >>> from sympy.core.random import seed
+    >>> from sympy.matrices.random import elementary
+    >>> seed(1)
```
oscarbenjamin, line 614: I'm not sure what's the best way to handle this. Ideally the doctest wouldn't set seed like this because it would influence the behaviour of other doctests. Maybe it's better just to skip the tests that are random with `DOCTEST_SKIP`.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+    works with symbols too
+
+    >>> cos_phi = cos(symbols('phi'))
+    >>> r = rotation(3, scalar=cos_phi)
+    >>> r
+    Matrix([
+    [              cos(phi), 0, sqrt(1 - cos(phi)**2)],
+    [                     0, 1,                     0],
+    [-sqrt(1 - cos(phi)**2), 0,              cos(phi)]])
+
```
oscarbenjamin, line 743: Maybe as a special case it would be good to check if the input expression is a `cos` and then use `sin` in that case.

I'm wondering how this works for all `phi`. Obviously it isn't always true that `sin(phi) == sqrt(1 - cos(phi)**2)`.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+            continue
+        val, cnt = s, 1
+    if cnt:
+        block_list.append((val, cnt))
+    spec = block_list
+    del block_list
+
+    # choose block randomly if dim and len(spec) does not meet
+    if not dim == sum(cnt for val, cnt in spec):
+        spec = tuple(_sample(spec) for _ in range(dim))
```
oscarbenjamin, line 1127: This signature is a bit strange e.g.:
```python
In [7]: jordan_normal(2, (1,2,3)) # random
Out[7]: 
⎡3  0⎤
⎢    ⎥
⎣0  3⎦

In [8]: jordan_normal(3, (1,2,3)) # not random
Out[8]: 
⎡1  0  0⎤
⎢       ⎥
⎢0  2  0⎥
⎢       ⎥
⎣0  0  3⎦

In [9]: jordan_normal(4, (1,2,3)) # random
Out[9]: 
⎡1  0  0  0⎤
⎢          ⎥
⎢0  1  0  0⎥
⎢          ⎥
⎢0  0  3  0⎥
⎢          ⎥
⎣0  0  0  1⎦
```
Also this seems odd:
```python
In [24]: jordan_normal(4, ((7,2),(9,3)))
Out[24]: 
⎡7  1  0  0⎤
⎢          ⎥
⎢0  7  0  0⎥
⎢          ⎥
⎢0  0  1  0⎥
⎢          ⎥
⎣0  0  0  1⎦
```

oscarbenjamin: It feels like maybe there's too much going on in one function here and with a single very overloaded argument.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+    which is called *orthogonal*
+    and $\mathbf{A}^H\mathbf{A} = \mathbf{I}$ for complex $\mathbf{A}$
+    (called *unitary*).
+
+    In normal from complex isometries have only diagonal entries of
+    norm 1, i.e. for a diagonal element $z$ the following holds.
+
+    .. math::
+
+        |z| = z * \bar{z} = 1
```
oscarbenjamin, line 1172: I guess it should be `|z|^2`

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+
+    with $c^2 + s^2 = 1$.
+
+    .. math::
+
+    So, **spec** is assumed to be a list of entries which are either
+    scalars $z$ with $|z| = z * \bar{z} = 1$
+    or
+    pairs $(c,s)$ with $c^2 + s^2 = 1$
+    to form the above rotaion square.
```
oscarbenjamin, line 1189: rotation

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+    singular
+
+    """
+
+    if length == 0:
+        return regular_to_singular(eye(dim), rank)
+
+    length = length or 2 * dim
+    lwr = triangular(dim, None, scalars, units,
+                     int(length / 2))
```
oscarbenjamin, line 1418: `int(length / 2)` is `length // 2` (which avoids the use of floats as an intermediary).

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+    ===========
+    A full rank square matrix has **rank=dim**.
+    Such a matrix is constructed as
+    product of :func:`elementary` matrices.
+
+    If **rank** is less than **dim**,
+    it is constructed from a full rank matrix $\mathbf{S}$
+    by choosing randomly $r$ basis columns
+    and build $n-r$ columes as a linear combinations
+    of the basis colums.
```
oscarbenjamin, line 1370: columns (twice)

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+        number of invertible matrices to build the resulting matrix.
+
+    See Also
+    ========
+    invertible
+    triangular
+
+    """
+    rank = dim - 1 if rank is None else rank
+    return square(dim, rank, scalars, units, length)
```
oscarbenjamin, line 1532: Maybe an error should be raised here if `rank == dim`

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+              rank=None,
+              scalars=_elementary_scalars,
+              units=_elementary_units,
+              length=None):
+    r"""a nilpotent matrix of dinemsion n x n.
+
+    Explanation
+    ===========
+    A *nilpotent* matrix is a matrix $\mathbf{A}$ such that there is
+    an integer $n$ with $\mathbf{A}^n = 0$.
```
oscarbenjamin, line 1616: It is perhaps confusing to use `n` in two different ways here.

oscarbenjamin: It seems odd that this function allows specifying rank when I would have expected it to allow specifying the `k` such that `A^k = 0` and `A^{k-1} != 0` instead.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+
+
+# === compound matrices, i.e. product of base matrices ===
+
+
+def triangular(dim,
+               rank=None,
+               scalars=_elementary_scalars,
+               units=_elementary_units,
+               length=None):
```
oscarbenjamin, line 1292: I think it would be good to decide that some of these arguments should be keyword-only e.g.:
```python
 def triangular(dim,
￼               rank=None,
                *,
￼               scalars=_elementary_scalars,
￼               units=_elementary_units,
￼               length=None):
```
It's hard to keep track of more than a couple of optional positional arguments.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+    units : tuple or list of symbols (optional)
+        see :func:`invertible`
+    length : integer (optional with default 2 * **dim**)
+        see :func:`invertible`
+
+    """
+
+    normal_form = diagonal_normal(dim, spec)
+    s = invertible(dim, scalars, units, length)
+    return s.inv() * normal_form * s
```
oscarbenjamin, line 1764: Does the method used by `invertible` make it possible to compute the inverse of `s` more efficiently than just computing `s` and calling `s.inv()`?

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+                   scalars=_elementary_scalars,
+                   units=_elementary_units,
+                   length=None):
+    r"""Creates a square matrix n x n of a given rank
+
+    Explanation
+    ===========
+    Creates a square matrix n x n of a given rank
+    which is trigonalizable (has echelon form)
+    and has eigenvalues from a given set.
```
oscarbenjamin, line 1778: What exactly does trigonalizable mean? Every matrix has an echelon form.

How is this different from `square`?

sympy/matrices/random.py
```diff
@@ -0,0 +1,2343 @@
...
+
+def hermite(dim,
+            scalars=_elementary_scalars,
+            units=_elementary_units,
+            length=None):
+    r"""Creates a Hermintian square matrix n x n.
+
+    Explanation
+    ===========
+    A *Hermintian* matrix is a matrix such that $\mathbf{A}^H = \mathbf{A}$,
```
oscarbenjamin, line 2282: Hermitian (twice)

sympy/matrices/random.py
jan-philipp-hoffmann: Replaced `_abs_one(x)` by `abs(x) == 1` to avoid `isclose`. Note that as `complex` has numerical issues, will not work some cases. Even not for
```
>>> z = sympify(complex(1, 2))
>>> abs(z/abs(z)) == 1
False
```

> I'm not sure what the purpose of `_is_complex` is supposed to be.

replaced `_is_complex` by `_is_real` in order to check if the values have no imaginary part.

sympy/matrices/random.py
jan-philipp-hoffmann: Ok, fine. Changed.

sympy/matrices/random.py
jan-philipp-hoffmann: as long as `phi` is a real number this will hold as a rotation matrix. Even 
```
>>> from sympy.abc import phi
>>> from sympy.matrices.random import *
>>> rotation(3, scalar=phi)
Matrix([
[              phi, 0, sqrt(1 - phi**2)],
[                0, 1,                0],
[-sqrt(1 - phi**2), 0,              phi]])
```
looks fine for me. And for `1 - phi**2<0` this is _complex_ rotation, i.e. here in $SU_3$.

So I am not sure what I missed?

sympy/matrices/random.py
jan-philipp-hoffmann: Well, but very useful im practice. Moreover, the `spec` argument is used by `trigonalizable`. More precise it is directly forwarded. Here, I think,
```
trigonalizable(5, spec=[-1, 1, 2])
```
is by far more convenient (and natural) than 
```
trigonalizable(5, spec=[(-1,1), (1,1), (2,1)])
```
which is equivalent.

Nevertheless, at first glance it might be confusing as
```
>>> jordan_normal(3, (1, 1, 1))  # not random
Matrix([
[1, 1, 0],
[0, 1, 1],
[0, 0, 1]])
>>> jordan_normal(3, (1, 1, 1, 1))  # random
Out[5]: 
Matrix([
[1, 0, 0],
[0, 1, 0],
[0, 0, 1]])
```
nevertheless I would recommend to keep it.

sympy/matrices/random.py
jan-philipp-hoffmann: Yes. Thx.

sympy/matrices/random.py
jan-philipp-hoffmann: Nice.

sympy/matrices/random.py
jan-philipp-hoffmann: Hope it's better now.

sympy/matrices/random.py
jan-philipp-hoffmann: Good point.

sympy/matrices/random.py
jan-philipp-hoffmann: currently the have a common order for _all_ random matrix function: 
```
func(dim, index, rank, spec, scalars, units, length)
```
But which are primarily used depends on the matrix type. But even for `invertible(3, [1,2,3])` a positional argument `scalars` is quite handy.

sympy/matrices/random.py
jan-philipp-hoffmann: Yes, `trigonalizable` means it has an echelon form but with a fixed **spectrum** i.e. diagonal entries in echelon form. As given by `spec`. `square` gives a general square matrix which does only have general diagonal entries in echelon form. 

This function is very useful for generating matrix exercises in undergrad math courses.

sympy/matrices/random.py
jan-philipp-hoffmann: Fixed

sympy/matrices/random.py
oscarbenjamin: Making the code invisible does not really help. the problem is that calling the seed like this alters global state. When the doctests are run this doctest will be executed and then afterwards other doctests will be executed. Ideally the test runner would be in control of the seeds for doctests so that when some doctest fails the failure is something that can be reproduced by the test runner. Also ideally we wouldn't use the same seed always when running the doctests but any doctest that runs after this one would have its seed fixed by this one.

sympy/matrices/random.py
oscarbenjamin: I'm just wondering whether it would be better to use `sin(phi)` rather than `sqrt(1 - cos(phi)**2)`. The two are not always equal because sometimes `sin(phi) = -sqrt(1 - cos(phi)**2)`.

sympy/matrices/random.py
oscarbenjamin: What I mean is that when you see a function call like
```python
f(a,b,c,d,e)
```
It's hard to mentally keep track of what all of the different arguments are supposed to mean (especially if they mean very different things) and what their defaults would be if the arguments are not provided. Calling less common arguments by keyword argument makes it clearer when reading the code:
```python
f(a, b, scalars=c, units=d, length=e)
```
Much of SymPy's codebase dates back to before keyword-only arguments existed:
https://peps.python.org/pep-3102/
Otherwise these would have been used more widely in the codebase.

sympy/matrices/random.py
oscarbenjamin: What I'm wondering is if it could be more efficient to generate both a matrix and its inverse simultaneously like:
```python
sinv, s = invertible_with_inverse(...)
return sinv * normal_form * s
```
The code that builds `s` has access to all the matrices that make up `s` which are probably more easily invertible.

sympy/matrices/random.py
oscarbenjamin: Do you have any source for this meaning of trigonalizable? When I search this word I get e.g.:
https://encyclopediaofmath.org/wiki/Trigonalizable_element
Translating that into our language it says that a square matrix over a field K is trigonalizable if all its eigenvalues lie in K rather than some algebraic extension of K. If we allow K to be algebraically closed (e.g. the complex numbers) then every square matrix is trigonalizable so there needs to be some notion of the field that we are working over for trigonalizable to be a meaningful property of a square matrix.

Here there is no mention of a field but I guess that these matrices are implicitly defined to be in a ring something like `Z[s1,s2,...][e1,e2,...]` where the `si` are the scalars and the `ei` are the spec (eigenvalues).

I think perhaps this docstring could be a bit clearer in defining what a trigonalizable matrix is. I'm not sure what the best language is but I found this confusing to read also because the meaning of trigonalizable is not obvious from a quick web search.

jan-philipp-hoffmann: _trigonalizable_ means for square matrix $A$ that it is similar to an upper triangular matrix $T$ see [Wikipedia](https://en.wikipedia.org/wiki/Triangular_matrix#Triangularisability), i.e. there is an $S \in GL(n)$ s.th. $S^{-1} T S = A$.

Same as a diagonizable matrix is similar to a diagonal matrix.

Here the field $k$ is actually given by the _spec_, _scalars_ and _units_ arguments. If all of them are elements of $k$.

sympy/matrices/random.py
jan-philipp-hoffmann: This might work but the matrices to build `s` are quite low level (in `super_elementary`). So feeding them up the chain ist quite a bit of work. I'll take it on my backlog for a future version.

sympy/matrices/random.py
jan-philipp-hoffmann: I prefer keyword arguments, too. Turned all non-optional arguments into keyword only arguments.

sympy/matrices/random.py
jan-philipp-hoffmann: Both, `sin(phi)` and `-sin(phi)`, define proper rotation matrices `Matrix([[c, s], [-s, c]])` and `Matrix([[c, -s], [s, c]])`. Same for any variable `x` as long as `sqrt(x**2-1)` is real. Here `cos(phi)` is just a example for some `x`. Nothing specific about `cos` or `sin`.

As mentioned, helpful to generate sample rotation matrices with two angels `phi` and `psi` for exercises.
```
>>> simplify(orthogonal(4, spec=(cos(phi), cos(phi))))
Matrix([
[                          cos(phi), -(sqrt(2) + 2)*sqrt(sin(phi)**2)/4,                sqrt(sin(phi)**2)/2, (2 - sqrt(2))*sqrt(sin(phi)**2)/4],
[ (sqrt(2) + 2)*sqrt(sin(phi)**2)/4,                           cos(phi), (-2 + sqrt(2))*sqrt(sin(phi)**2)/4,               sqrt(sin(phi)**2)/2],
[              -sqrt(sin(phi)**2)/2,  (2 - sqrt(2))*sqrt(sin(phi)**2)/4,                           cos(phi), (sqrt(2) + 2)*sqrt(sin(phi)**2)/4],
[(-2 + sqrt(2))*sqrt(sin(phi)**2)/4,               -sqrt(sin(phi)**2)/2, -(sqrt(2) + 2)*sqrt(sin(phi)**2)/4,                          cos(phi)]])

```

sympy/matrices/random.py
jan-philipp-hoffmann: I though of something like storing the state of random before calling `seed()`and re-establishing the state at the end of the docstring. Since there seem no seed setting possible in `sympy.testing.runtests.doctest`.

If we could add [`sphinx.ext.doctest`](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#module-sphinx.ext.doctest) as sphinx framework (in `conf.py`) one could make use of [`rng.getstate()`](https://docs.sympy.org/latest/modules/core.html?highlight=random#module-sympy.core.random)  

```
    Examples
    ========

    .. testsetup::

       >>> from sympy.core.random import rng
       >>> _rng_state = rng.getstate()

    >>> from sympy import sqrt, eye, expand
    >>> from sympy.core.random import seed
    >>> from sympy.matrices.random import orthogonal
    >>> seed(1)

    >>> orthogonal(3, scalars=(-1,1))
    Matrix([
    [-1,  0, 0],
    [ 0, -1, 0],
    [ 0,  0, 1]])

    .. testcleanup::

       >>> assert not rng.getstate() == _rng_state
       >>> rng.setstate(_rng_state)
       >>> assert rng.getstate() == _rng_state

```

This seems to work fine. Wonder if this would be ok?

Otherwise I don't know better than `DOCTEST_SKIP`.

sympy/matrices/random.py
oscarbenjamin: I have heard of the word `triangularizable` (the word used in wikipedia) before but `trigonalizable` is new to me. Is there a reason to prefer `trigonalizable`?

sonntagsgesicht: Sorry, didn't realize the difference in spelling. _triangularizable_ is fine for me also I think _trigonalizable_ wouldn't be wrong. Nevertheless, I'll change it tomorrow.

sympy/matrices/random.py
jan-philipp-hoffmann: I tried [sphinx.ext.doctest](https://github.com/sympy/sympy/pull/22103#:~:text=we%20could%20add-,sphinx.ext.doctest,-as%20sphinx%20framework) on a different [branch](https://github.com/jan-philipp-hoffmann/sympy/tree/pre_master_ext_doctest) and it seems to work well; both docs as well as doctests. The examples look more natural as invoking `seed()` could be hidden in `.. testsetup::`.

If you agree, I would merge this with this PR.

oscarbenjamin: The sphinx doctest plugin looks good but I'm not sure it's compatible with SymPy's existing doctest runner or the pytest-doctestplus plugin.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2388 @@
...
+_unitary_scalars = tuple(c * 1 + s * I for c, s in _rotation_scalars)
+
+
+# === fundamental constructor ===
+
+
+def super_elementary(dim,
+                     index=None,
+                     value=None,
+                     *scalars):
```
oscarbenjamin, line 63: Every other function with a scalars argument takes a keyword argument rather than `*args`. It would seem more consistent for this one to do the same.

sympy/matrices/random.py
```diff
@@ -0,0 +1,2388 @@
...
+
+    .. math::
+
+       \left[\begin{array}{cc}a & b \\ -c & d\end{array}\right]
+       =
+       \left[\begin{array}{cc}
+        \cos \phi & \sin \phi \\ -\sin \phi & \cos \phi \end{array}\right]
+
+    In this module the super elementary matrix serves as the base class to
+    create futher matrices of given type.
```
oscarbenjamin, line 106: Perhaps "base case" because this isn't really a class.

sympy/matrices/random.py
oscarbenjamin: It seems that most cases where `invertible` is called by the other functions here also want the inverse.

sympy/matrices/random.py
jan-philipp-hoffmann: Best would be to remove this function. It is kind of legacy as it is called only from `elementary`, `transposition` and `rotation` and each has it own codepath within `super_elementary`. It is more clear to create those matrices explicitly within each function.

sympy/matrices/random.py
jan-philipp-hoffmann: no more relevant once `super_elementary` is removed

sympy/matrices/random.py
jan-philipp-hoffmann: found even better solution. We just put the hidden code into a comment like 
```
.. ..testsetup::

   >>> from sympy.core.random import rng, seed
   >>> _rng_state = rng.getstate()
   >>> seed(0)
```
so we need no additional sphinx plugin. The doc test still execute the code but it is invisible in the docs.

sympy/matrices/random.py
oscarbenjamin: Calculating the inverse can be a lot more expensive for matrices with different kinds of entries rather than just integers.

oscarbenjamin:
Can anyone else review this?

This is a new module for generating random matrices and is 4000 lines of code altogether.

jan-philipp-hoffmann:
Maybe the full module is too much to be reviewed at once? It's new sliced it into smaller chunks and starts with `elementary` + `triangular` + `square`.

This fixes at least the seed issue #22592 and states the concept of seeding random doctest. Once this is merged we might move on step by step adding more cases.

Otherwise, the PR might become stale again and all efforts so far would be wasted.

oscarbenjamin:
> Maybe the full module is too much to be reviewed at once? It's new sliced it into smaller chunks and starts with `elementary` + `triangular` + `square`.

If you want to break this into smaller chunks then I suggest keeping this PR as the whole module and just making a separate PR with say the first chunk. If we want to make it simpler for someone to review then it would be better to have a separate PR that does not have all of the comments above. It's useful to keep the full module around somewhere though for reference. Also this has 89 commits. I would suggest squashing those but if you're going to copy out the changes into separate PRs then it isn't necessary to squash them.

jan-philipp-hoffmann:
> > Maybe the full module is too much to be reviewed at once? It's new sliced it into smaller chunks and starts with `elementary` + `triangular` + `square`.
> 
> If you want to break this into smaller chunks then I suggest keeping this PR as the whole module and just making a separate PR with say the first chunk. If we want to make it simpler for someone to review then it would be better to have a separate PR that does not have all of the comments above. It's useful to keep the full module around somewhere though for reference. Also this has 89 commits. I would suggest squashing those but if you're going to copy out the changes into separate PRs then it isn't necessary to squash them.

opened new PR https://github.com/sympy/sympy/pull/24519

PR #22111, opened 2021-09-18
Fixed syntax error

oscargus:
Three instances of `S.zero` changed to `S.Zero`.

#### Other comments
I cannot figure out tests for the relational. What I would need is an expression which returns `True` for `.is_zero`, but does not evaluate or simplify to 0. (Using `evaluate=False` is not really an option, unless running the whole test part with it.)

Will backport to 1.9 if this seems OK.
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/relational.py
+++ b/sympy/core/relational.py
@@ -336,7 +336,7 @@ class Relational(Boolean, EvalfMixin):
                         else:
                             r = r.func(x, -b / m)
                     else:
-                        r = r.func(b, S.zero)
+                        r = r.func(b, S.Zero)
                 except ValueError:
                     # maybe not a linear function, try polynomial
                     from sympy.polys import Poly, poly, PolynomialError, gcd
@@ -375,7 +375,7 @@ class Relational(Boolean, EvalfMixin):
                             r = r.func(lhsterm, -newexpr)
 
                     else:
-                        r = r.func(constant, S.zero)
+                        r = r.func(constant, S.Zero)
                 except ValueError:
                     pass
         # Did we get a simplified result?
--- a/sympy/core/tests/test_relational.py
+++ b/sympy/core/tests/test_relational.py
@@ -5,7 +5,7 @@ from sympy.testing.pytest import XFAIL, raises, warns_deprecated_sympy
 from sympy import (S, Symbol, symbols, nan, oo, I, pi, Float, And, Or,
                    Not, Implies, Xor, zoo, sqrt, Rational, simplify, Function,
                    log, cos, sin, Add, Mul, Pow, floor, ceiling, trigsimp, Reals,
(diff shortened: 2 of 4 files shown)


sympy/core/relational.py
```diff
@@ -336,7 +336,7 @@ def _eval_simplify(self, **kwargs):
                         else:
                             r = r.func(x, -b / m)
                     else:
-                        r = r.func(b, S.zero)
+                        r = r.func(b, S.Zero)
```
oscarbenjamin, line 339: Are there tests covering this line?

oscargus: No, as I wrote in the original comment, I cannot figure out how to get there.

oscargus: Basically, this is for a case where `dif` is `m*x + b` where `m.is_zero` is None or True, but `m` should have been evaluated to zero in the subtraction (or previous simplification of arguments or re-evaluation of arguments).

oscarbenjamin: Okay, so we need `m` to have no symbols and to be possibly zero but not something that `simplify` can simplify to zero. This can do it:
```python
from sympy import *

x = Symbol('x')
m = exp(1) - exp_polar(1)

simplify(m*x < 1)
```
That gives:
```
  File "t.py", line 6, in <module>
    simplify(m*x < 1)
  File "/home/oscar/current/sympy/sympy.git/sympy/simplify/simplify.py", line 592, in simplify
    return _eval_simplify(**kwargs)
  File "/home/oscar/current/sympy/sympy.git/sympy/core/relational.py", line 339, in _eval_simplify
    r = r.func(b, S.zero)
  File "/home/oscar/current/sympy/sympy.git/sympy/core/singleton.py", line 122, in __getattr__
    name, self))
AttributeError: Attribute 'zero' was not installed on SymPy registry S
```

oscarbenjamin: For a test I guess it's better if it's something that `simplify` would never be able to handle e.g.:
```python
from sympy import *

class A(Expr):
    pass

x = Symbol('x')
m = A()

simplify(m*x < 1)
```

oscargus: Great! Thanks! I tried quite a few, but couldn't find anything working.

oscarbenjamin:
Future changes in `simplify` might mean that `simplify(exp(1)-exp_polar(1))` simplified to zero which would defeat the purpose of the test. I think it would be good to also test something that can never be simplified like a completely arbitrary Expr but that's not hugely important.

This looks good to me and can be backported to the 1.9 branch.

oscargus:
Yes, it is indeed an issue that the currently added tests can be invalidated by improvements elsewhere...

I'll create a separate 1.9 PR then.

PR #22117, opened 2021-09-18
fix(printing): suppress printing default Heaviside arg

oscarbenjamin:
Heaviside(x) was changed to have a default value of 1/2 for the second
argument rather than leaving it undefined. This changed the args from
(x,) to (x, 1/2) which then changed the output of several printers. This
commit restores the previous printing behaviour so that if the second
arg for Heaviside(x, y) is equal to the default 1/2 then it is
suppressed from printing in the str/repr/latex printers.

Fixes the printing issue from #21945 on the master branch.

NO ENTRY

### DIFF
--- a/sympy/printing/printer.py
+++ b/sympy/printing/printer.py
@@ -325,9 +325,10 @@ class Printer:
                     c.__name__ == classes[0].__name__ or \
                     c.__name__.endswith("Base")) + classes[i:]
             for cls in classes:
-                printmethod = '_print_' + cls.__name__
-                if hasattr(self, printmethod):
-                    return getattr(self, printmethod)(expr, **kwargs)
+                printmethodname = '_print_' + cls.__name__
+                printmethod = getattr(self, printmethodname, None)
+                if printmethod is not None:
+                    return printmethod(expr, **kwargs)
             # Unknown object, fall back to the emptyPrinter.
             return self.emptyPrinter(expr)
         finally:
--- a/sympy/printing/tests/test_repr.py
+++ b/sympy/printing/tests/test_repr.py
@@ -1,7 +1,7 @@
 from typing import Any, Dict
 
 from sympy.testing.pytest import raises
-from sympy import (symbols, sympify, Function, Integer, Matrix, Abs,
+from sympy import (symbols, sympify, Function, Integer, Matrix, Abs, Heaviside,
     Rational, Float, S, WildFunction, ImmutableDenseMatrix, sin, true, false, ones,
     sqrt, root, AlgebraicNumber, Symbol, Dummy, Wild, MatrixSymbol, Q)
 from sympy.combinatorics import Cycle, Permutation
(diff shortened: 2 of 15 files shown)


sympy/printing/tests/test_repr.py
```diff
@@ -67,6 +67,11 @@ def test_Function():
     sT(sin, "sin")
 
 
+def test_Heaviside():
+    sT(Heaviside(x), "Heaviside(Symbol('x'))")
+    sT(Heaviside(x, 1), "Heaviside(Symbol('x'), Integer(1))")
```
oscarbenjamin, line 72: Maybe the `srepr` printer should not be changed...

hanspi42: No, it should not, or else you also have to change lambdify again ;)

hanspi42: In case you are looking for the documentation you have to fix, search for regular expression ´Heaviside\([^\(]+1/2\)´

sympy/printing/printer.py
```diff
@@ -325,9 +325,10 @@ def _print(self, expr, **kwargs):
                     c.__name__ == classes[0].__name__ or \
                     c.__name__.endswith("Base")) + classes[i:]
             for cls in classes:
-                printmethod = '_print_' + cls.__name__
-                if hasattr(self, printmethod):
-                    return getattr(self, printmethod)(expr, **kwargs)
+                printmethodname = '_print_' + cls.__name__
+                printmethod = getattr(self, printmethodname, None)
+                if printmethod is not None:
+                    return printmethod(expr, **kwargs)
```
oscarbenjamin, line 331: This change makes it possible to do:
```
_print_Heaviside = None
```
in the CodePrinter class so that all subclasses will ignore the parent `StrPrinter._print_Heaviside` method.

sympy/printing/tests/test_repr.py
oscarbenjamin: I think it's the StrPrinter method that affects the code printers but I fond a fix for that (see below).

sympy/printing/printer.py
hanspi42: Wow. That was the knowledge I was missing when I first worked on this.

oscarbenjamin:
Okay, the main tests have passed. @oscargus @jksuom @hanspi42 can anyone review this?

hanspi42:
> 
> 
> Okay, the main tests have passed. @oscargus @jksuom @hanspi42 can anyone review this?

I have reviewed it, internally it looks OK. I think using ´pargs´ is robust.

The one thing I cannot review properly (because I do not understand printers enough) is whether the ´_print_Heaviside = None´ mechanism you introduced can have effects somewhere else because it affects a whole class. I think not, but cannot be sure, so would be happy if another reviewer confirms or you, Oscar, say it cannot be a problem because "xxx".

oscarbenjamin:
> ´_print_Heaviside = None´ mechanism you introduced can have effects somewhere else

I'm not 100% sure about that myself as I have not done much work on the printers before. When I realised that adding `_print_Heaviside` to the str-printer affected the Octave printer I thought that it seems an odd design. Probably it's a mistake for CodePrinter to subclass StrPrinter and it should be the other way around i.e. the StrPrinter is a particular kind of CodePrinter so it should subclass CodePrinter.

oscarbenjamin:
I opened #22118 which is the same as this but for the 1.9 branch

jksuom:
This looks good to me.

oscarbenjamin:
Thanks for the reviews. I'll get this in.

PR #22114, opened 2021-09-18
Added .det()-method to MatExpr

oscargus:
#### Brief description of what is fixed or changed
Convenience method to be able to call the determinant using a method rather than a function.

There are basically three ways to determine the determinant:
`Determinant().doit()`, `det()`, `.det()`.
The last one didn't work on matrix expressions (although for example `.inv()` did).

#### Release Notes

* matrices
    * It is now possible to use `.det()` on matrix symbols and matrix expressions.

PR #22119, opened 2021-09-19
Import cleanup - part 1

oscargus:
In many files, the same thing is imported more than once, see: https://github.com/sympy/sympy/pull/22114#discussion_r711608622

This is the first PR to mitigate this.

#### Other comments
For `meijerint.py` and `transforms.py`, I went a bit longer and moved most imports to the top level. Some cannot be moved due to circular imports, some was deemed to be rare enough to stay.

I'm thinking about opening an issue about this. Actually a quite good way to get to know the code-base.

What are the reasons to not move imports to the top? Circular imports is one. Performance in case of rarely used functions another.

I've seen some files having imports at the bottom. What is the benefit?

Is there any benefit of having fully specified paths to the imports, .e.g., `from sympy import sin` vs `from sympy.functions.elementary.trignometric import sin`? Relative paths vs absolute?

(I've read up a bit, but there seems to be a bit of preferences involved and the code base is not consistent enough to be able to tell...)

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/assumptions/refine.py
+++ b/sympy/assumptions/refine.py
@@ -232,7 +232,6 @@ def refine_atan2(expr, assumptions):
     nan
     """
     from sympy.functions.elementary.trigonometric import atan
-    from sympy.core import S
     y, x = expr.args
     if ask(Q.real(y) & Q.positive(x), assumptions):
         return atan(y / x)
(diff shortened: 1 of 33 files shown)


oscarbenjamin:
Looks good to me

PR #22110, opened 2021-09-17
geometry - fixes issue #22109 and now Point class will subclass EvalfMixin

praneethratna:
#### Brief description of what is fixed or changed

Fixes #22109 , https://github.com/sympy/sympy/issues/22109#issuecomment-921785187 and now Point2D will subclass `EvalfMixin` .

#### Release Notes

NO ENTRY

PR #20493, opened 2020-11-26
[WIP]Add a new helper method for evalf

sidhu1012:
#### References to other Issues or PRs

Fixes #20479 

#### Brief description of what is fixed or changed
Added a new helper method for `evalf` to make `evalf` consistent without compromising `precision` and `options` parameter.

#### Other comments
Read discussion #20479 

#### Release Notes

* core
    * Added a new helper method for evalf()

PR #22131, opened 2021-09-21
all geometry entities have evalf capability

### LINKED ISSUE #22109: evalf(), .n or N of Point2D produces error.
The following code produces the error `TypeError: evalf() got multiple values for argument 'n'`.  I would expect it to produce `𝑃𝑜𝑖𝑛𝑡2𝐷(6,6)`

```python
from sympy import N, symbols
from sympy.geometry import Point2D, Segment2D

a,b,c = symbols('a,b,c')

eq1 = a + b + c
subs = {a:1,b:2,c:3}
eq1.n(n=5, subs=subs)

eq2 = a * b * c
eq2.n(n=5, subs=subs)

p = Point2D(eq1 ,eq2)
p.n(n=5, subs=subs)
```

smichr:

closes #22110 as a continuation
fixes #22109

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/geometry/ellipse.py
+++ b/sympy/geometry/ellipse.py
@@ -8,6 +8,7 @@ Contains
 
 from sympy import Expr, Eq
 from sympy.core import S, pi, sympify
+from sympy.core.evalf import prec_to_dps
 from sympy.core.parameters import global_parameters
 from sympy.core.logic import fuzzy_bool
 from sympy.core.numbers import Rational, oo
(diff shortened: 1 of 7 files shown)


smichr:
@oscarbenjamin , does this look complete, now?

sympy/geometry/ellipse.py
```diff
@@ -1605,6 +1606,13 @@ def __new__(cls, *args, **kwargs):
 
             raise GeometryError("Circle.__new__ received unknown arguments")
 
+    def _eval_evalf(self, prec=15, **options):
+        pt, r = self.args
+        dps = prec_to_dps(prec)
+        pt = pt.evalf(n=dps, **options)
+        r = r.evalf(n=dps, **options)
+        return self.func(pt, r, evaluate=False)
```
praneethratna, line 1614: @smichr Can this be written something like `self.func(*[x.evalf(n=prec_to_dps(prec), **options) for x in self.args], evaluate = False)` similar to one in Point class ?

oscarbenjamin: This method computes `dps` only once which is good. Ideally `_eval_evalf` would be able to call a method that takes binary precision. There is a PR for that but it is unfinished  #20493

oscarbenjamin:
Looks good to me

PR #22136, opened 2021-09-22
Cleanup and some minor refactoring of perm_groups

oscargus:
pylint reported many issues related to attributes defined outside `__init__`. Also found and fixed some other stuff.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/combinatorics/perm_groups.py
+++ b/sympy/combinatorics/perm_groups.py
@@ -143,39 +143,40 @@ class PermutationGroup(Basic):
             args = list(uniq([_af_new(list(a)) for a in args]))
         if len(args) > 1:
             args = [g for g in args if not g.is_identity]
-        obj = Basic.__new__(cls, *args, **kwargs)
-        obj._generators = args
-        obj._order = None
-        obj._center = []
-        obj._is_abelian = None
-        obj._is_transitive = None
-        obj._is_sym = None
-        obj._is_alt = None
-        obj._is_primitive = None
-        obj._is_nilpotent = None
-        obj._is_solvable = None
-        obj._is_trivial = None
-        obj._transitivity_degree = None
-        obj._max_div = None
-        obj._is_perfect = None
-        obj._is_cyclic = None
-        obj._r = len(obj._generators)
-        obj._degree = obj._generators[0].size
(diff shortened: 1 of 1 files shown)


sympy/combinatorics/perm_groups.py
```diff
@@ -3231,9 +3229,9 @@ def is_cyclic(self):
         order = self.order()
 
         if order < 6:
-            self._is_abelian == True
+            self._is_abelian = True
             if order != 4:
-                self._is_cyclic == True
+                self._is_cyclic = True
```
oscargus, line 3234: I doubt these lines are tested... (Unfortunately, I know nothing about the subject.)

sympy/combinatorics/perm_groups.py
```diff
@@ -5217,12 +5216,8 @@ class SymmetricPermutationGroup(Basic):
...
     """
 
-    def __new__(cls, deg):
-        deg = _sympify(deg)
-        obj = Basic.__new__(cls, deg)
-        obj._deg = deg
-        obj._order = None
-        return obj
+    def __init__(self, *args, **kwargs):
+        self._order = factorial(self.args[0])
```
smichr, line 5220: Why this instead of the lazy computation upon a `__contains__` request?

oscargus: I can revert it for sure. Just seemed to make more sense to compute it once and for all rather than checking every time. But no strong feelings there...

oscargus: (I know it is computed once in the other setup as well.)

sympy/combinatorics/perm_groups.py
smichr: docstring affirms
```
        - If the order of the group is less than $6$ and is not $4$, the
          group is cyclic.
```
I don't see assurance that it is Abelian, however. It could be None or True at this point in the code,

oscargus: @sylee957 do you remember how this should work?

sylee957: I think that the new behavior should work if it passes the tests

PR #22146, opened 2021-09-23
Matrix power now fully supports .as_explicit()

Upabjojr:

Method `.as_explicit()` applied on `MatPow` objects was returning the full explicit matrix only when the exponent was a positive integer. In all other cases, a fake explicit matrix was returned (i.e. the matrix expression in the base of the `MatPow` object was not made explicit).

This PR fixes this issue and extends the explicit matrix to all exponents. Only matrices with symbolic shapes won't get an explicit form.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/matrices/expressions/matpow.py
+++ b/sympy/matrices/expressions/matpow.py
@@ -4,8 +4,7 @@ from .special import Identity
 from sympy.core import S
 from sympy.core.sympify import _sympify
 from sympy.matrices import MatrixBase
-from ... import cacheit, Integer
-from ...external.gmpy import SYMPY_INTS
+from ... import cacheit
 
 
 class MatPow(MatrixExpr):
@@ -48,7 +47,7 @@ class MatPow(MatrixExpr):
             # We still have a MatPow, make an explicit MatMul out of it.
             if A.exp.is_Integer and A.exp.is_positive:
                 A = MatMul(*[A.base for k in range(A.exp)])
-            elif isinstance(self.rows, (SYMPY_INTS, Integer)) and isinstance(self.cols, (SYMPY_INTS, Integer)):
+            elif not self._is_shape_symbolic():
                 return A._get_explicit_matrix()[i, j]
             else:
                 # Leave the expression unevaluated:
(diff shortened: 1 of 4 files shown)


sympy/matrices/expressions/matpow.py
```diff
@@ -47,7 +48,7 @@ def _entry(self, i, j, **kwargs):
             # We still have a MatPow, make an explicit MatMul out of it.
             if A.exp.is_Integer and A.exp.is_positive:
                 A = MatMul(*[A.base for k in range(A.exp)])
-            elif A.exp.is_number:
+            elif isinstance(self.rows, (SYMPY_INTS, Integer)) and isinstance(self.cols, (SYMPY_INTS, Integer)):
```
oscarbenjamin, line 51: It's not good if self.rows/cols is not known to be sympified. Perhaps there should be a different attribute for getting symbolic shape vs explicit int shape.

Upabjojr: It checks both `int` and `Integer`, so non-sympified expressions should be OK.

oscarbenjamin: I mean that it is problematic if an attribute is not known to be sympified or not e.g.:
```python
In [6]: M1 = Matrix([[1]])

In [7]: M2 = MatrixSymbol('x', n, n)

In [8]: type(M1.shape[0])
Out[8]: int

In [9]: type(M2.shape[0])
Out[9]: sympy.core.symbol.Symbol
```
Maybe if the shape is sometimes a symbolic expression it should always be symbolic e.g. `Integer(1)` rather than `int(1)`.

Upabjojr: Well, I'm just keeping the code check the same way as they are on master (and according to `git blame` the checks for both `int` and `Integer` have been there since 2015).

Technically `.shape` is not part of `.args` of the explicit valued matrix, so it is not required to be sympified. OTOH, the `.shape` of `MatrixSymbol` is sympified.

Upabjojr:
I am going to merge this PR.

oscarbenjamin:
Looks good to me

PR #22147, opened 2021-09-23
Make ArrayAdd.as_explicit() work correctly

Upabjojr:
NO ENTRY

### DIFF
--- a/sympy/tensor/array/expressions/array_expressions.py
+++ b/sympy/tensor/array/expressions/array_expressions.py
@@ -7,7 +7,7 @@ from typing import Optional, List, Dict
 
 import typing
 
-from sympy import Expr, ImmutableDenseNDimArray, S, Symbol, ZeroMatrix, Basic, tensorproduct, Add, permutedims, \
+from sympy import Expr, ImmutableDenseNDimArray, S, Symbol, ZeroMatrix, Basic, tensorproduct, permutedims, \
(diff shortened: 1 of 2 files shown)


sympy/tensor/array/expressions/array_expressions.py
```diff
@@ -288,7 +288,7 @@ def _flatten_args(cls, args):
         return new_args
 
     def as_explicit(self):
-        return Add.fromiter([arg.as_explicit() for arg in self.args])
```
smichr, line 291: why didn't this work?

Upabjojr: It's creating an `Add` object instead of an array object in case `arg.as_explicit()` returns an array. It works in case it returns a matrix. I didn't investigate that much.

Do you think this should be fixed at the `Add` level?

Upabjojr: OK, I had a look at the code. Matrices are handled in `Add`'s postprocessors. There is no postprocessing for N-dim arrays, that's the reason it won't work. I would like to keep this PR this way as I would prefer not to venture into some edits of `Add`'s postprocessors.

oscarbenjamin: Is this not handled for matrix using kinds now? Couldn't the same be done for Array?

Upabjojr: > Is this not handled for matrix using kinds now? Couldn't the same be done for Array?

Probably, but I don't really want to go into the kinds for arrays until the array expressions module is finished and ready to use.

PR #22150, opened 2021-09-24
Improved type checking - part 2

oscargus:
Improved type checking

#### Release Notes

NO ENTRY


oscargus:
Will have to dig into the errors later. The ODEs work on my local machine and the LaTeX parsing works in Spyder, but not from a shell... Will try to extract tests and move them closer to the source...

oscargus:
This seems to pass now.

oscarbenjamin:
Looks good!

Issue #20914, opened 2021-02-05
Hypothesis testing


asmeurer:
I'd like to use this issue to discuss the idea of using [hypothesis](https://hypothesis.readthedocs.io/en/latest/index.html) in SymPy. 

For those who don't know, hypothesis is a library that lets you do property based testing. You tell it what the input to your function should look like and assert what properties should always hold, and it tries to find inputs that falsify that. Here's an example to show what a hypothesis test looks like

```py
from hypothesis import given
from hypothesis.strategies import integers

 @given(integers())
 def test_factorint(x):
     f = factorint(x)
     assert Mul(*[b**e for b, e in f.items()]) == x
     for b in f:
         assert abs(b) in [0, 1] or isprime(b)
```

(note that this test as written will fail because it takes too long for some inputs, but it's just to give the idea of what a property based test looks like)

Hypothesis is an extremely powerful tool. It is very good at finding examples that fail your tests, things which you would never think to test yourself. However, it's also very picky. As soon as it finds a failing test, it remembers it and always reports it. So it's only useful to add it to SymPy in a place where either the code currently works, or we are willing to fix any bugs that it finds.

There has been a lot of discussion on this in the past. See https://github.com/sympy/sympy/pull/17190 and https://github.com/sympy/sympy/pull/20906. 

My idea with hypothesis is to start small. Much smaller than what was proposed in https://github.com/sympy/sympy/pull/17190 (although that approach can still have some merits as something to run independently to see if anything interesting pops up). For example, the example test I wrote in https://github.com/sympy/sympy/pull/20906 passes on that branch. I just ran over 20000 examples. Even it, though, is still too much to start with, IMO, because the "correct" strategy to get it to find interesting examples is nontrivial. Random polynomials do not factor or have interesting roots. So you need to generate things in a way that matches what you are looking for. 

I'll have to think a bit on where a good place to start would be. Ideally it would be something that is easy to generate with the builtin hypothesis strategies, something that doesn't blow up in terms of performance on certain inputs, and something where interesting inputs aren't difficult to find from the naive way of generating them. 

[...]

PR #20635, opened 2020-12-20
Made ordered() calls faster for large expressions

kc611:
#### References to other Issues or PRs

Related to #20359 

#### Brief description of what is fixed or changed

Added changes suggested by @oscarbenjamin in the issue page. 

#### Release Notes

- core
  - ordered() made faster for large expressions

PR #22152, opened 2021-09-25
print -oo relational first

### LINKED ISSUE #14933: RecursionError with Piecewise.xreplace containing Indexed objects
Minimal example to reproduce the issue:

```python
from sympy import Piecewise, Symbol, ITE, MatrixSymbol

x = Symbol('x')
y = Symbol('y')

inp = MatrixSymbol('inp', 1, 1)
rep_dict = {y: inp[0, 0], x: inp[0, 0]}

p = Piecewise((1, ITE(y > 0, x < 0, True)))
z = p.xreplace(rep_dict)
```

```
RecursionError                            Traceback (most recent call last)
<ipython-input-44-46c498a48320> in <module>()
----> 1 z = p.xreplace(rep_dict)
      2 print(z)

/home/rotis/anaconda/envs/calphadpy3/lib/python3.5/site-packages/sympy/core/basic.py in xreplace(self, rule)
   1113 
   1114         """
-> 1115         value, _ = self._xreplace(rule)
   1116         return value
   1117 

/home/rotis/anaconda/envs/calphadpy3/lib/python3.5/site-packages/sympy/core/basic.py in _xreplace(self, rule)
   1134             args = tuple(args)
   1135             if changed:
-> 1136                 return self.func(*args), True
   1137         return self, False
   1138 

/home/rotis/anaconda/envs/calphadpy3/lib/python3.5/site-packages/sympy/functions/elementary/piecewise.py in __new__(cls, *args, **options)
    134 
    135         if options.pop('evaluate', True):
--> 136             r = cls.eval(*newargs)
    137         else:
    138             r = None

/home/rotis/anaconda/envs/calphadpy3/lib/python3.5/site-packages/sympy/functions/elementary/piecewise.py in eval(cls, *_args)
    183                     x = free.pop()
    184                     try:
--> 185                         c = c.as_set().as_relational(x)
    186                     except NotImplementedError:
    187                         pass

/home/rotis/anaconda/envs/calphadpy3/lib/python3.5/site-packages/sympy/logic/boolalg.py in as_set(self)
    144             reps = {}
    145             for r in self.atoms(Relational):
--> 146                 if periodicity(r, x) not in (0, None):
    147                     s = r._eval_as_set()
    148                     if s in (S.EmptySet, S.UniversalSet, S.Reals):

/home/rotis/anaconda/envs/calphadpy3/lib/python3.5/site-packages/sympy/calculus/util.py in periodicity(f, symbol, check)
    471     elif period is None:
    472         from sympy.solvers.decompogen import compogen
--> 473         g_s = decompogen(f, symbol)
    474         num_of_gs = len(g_s)
    475         if num_of_gs > 1:

/home/rotis/anaconda/envs/calphadpy3/lib/python3.5/site-packages/sympy/solvers/decompogen.py in decompogen(f, symbol)
     55         f1 = f.subs(gens[0], symbol)
     56         f2 = gens[0]
---> 57         result += [f1] + decompogen(f2, symbol)
     58         return result
     59 

/home/rotis/anaconda/envs/calphadpy3/lib/python3.5/site-packages/sympy/solvers/decompogen.py in decompogen(f, symbol)
     55         f1 = f.subs(gens[0], symbol)
     56         f2 = gens[0]
---> 57         result += [f1] + decompogen(f2, symbol)
     58         return result
     59 

/home/rotis/anaconda/envs/calphadpy3/lib/python3.5/site-packages/sympy/solvers/decompogen.py in decompogen(f, symbol)
     55         f1 = f.subs(gens[0], symbol)
     56         f2 = gens[0]
---> 57         result += [f1] + decompogen(f2, symbol)
     58         return result
     59 

... last 3 frames repeated, from the frame below ...

/home/rotis/anaconda/envs/calphadpy3/lib/python3.5/site-packages/sympy/solvers/decompogen.py in decompogen(f, symbol)
     55         f1 = f.subs(gens[0], symbol)
     56         f2 = gens[0]
---> 57         result += [f1] + decompogen(f2, symbol)
     58         return result
     59 

RecursionError: maximum recursion depth exceeded
```

Note that this will occur even if you pass `evaluate=False` to `Piecewise`.
This may be related to #14899

smichr:

When a `Relational` in an `And` involves `-oo`, it is put as the first arg.

A note was added inline to `ordered` and `_node_count` to indicate why they might fail and what should be done.
For example, a TODO in OracleGate indicates that the second arg is not Basic; hence OracleGate cannot be used with `ordered` or `count_ops`.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/compatibility.py
+++ b/sympy/core/compatibility.py
@@ -473,8 +473,15 @@ def default_sort_key(item, order=None):
     return (cls_index, 0, item.__class__.__name__
             ), args, S.One.sort_key(), S.One
 
+
 def _node_count(e):
+    # this not only counts nodes, it affirms that the
+    # args are Basic (i.e. have an args property). If
+    # some object has a non-Basic arg, it needs to be
+    # fixed since it is intended that all Basic args
+    # are of Basic type (though this is not easy to enforce).
     return 1 + sum(map(_node_count, e.args))
+
 
 def _nodes(e):
     """
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -3147,6 +3147,10 @@ def count_ops(expr, visual=False):
         while args:
             a = args.pop()
 
+            # if the following fails because the object is
+            # not Basic type, then the object should be fixed
+            # since it is the intention that all args of Basic
+            # should themselves be Basic
             if a.is_Rational:
                 #-1/3 = NEG + DIV
                 if a is not S.One:
--- a/sympy/printing/str.py
+++ b/sympy/printing/str.py
@@ -7,6 +7,7 @@ from typing import Any, Dict
 from sympy.core import S, Rational, Pow, Basic, Mul, Number
 from sympy.core.mul import _keep_coeff
 from sympy.core.function import _coeff_isneg
+from sympy.core.relational import Relational
 from sympy.sets.sets import FiniteSet
 from .printer import Printer, print_function
 from sympy.printing.precedence import precedence, PRECEDENCE
@@ -78,7 +79,12 @@ class StrPrinter(Printer):
         return '~%s' %(self.parenthesize(expr.args[0],PRECEDENCE["Not"]))
 
     def _print_And(self, expr):
-        return self.stringify(expr.args, " & ", PRECEDENCE["BitwiseAnd"])
+        args = list(expr.args)
+        for j, i in enumerate(args):
+            if isinstance(i, Relational) and (
+                    i.canonical.rhs is S.NegativeInfinity):
+                args.insert(0, args.pop(j))
+        return self.stringify(args, " & ", PRECEDENCE["BitwiseAnd"])
 
     def _print_Or(self, expr):
         return self.stringify(expr.args, " | ", PRECEDENCE["BitwiseOr"])
--- a/sympy/solvers/inequalities.py
+++ b/sympy/solvers/inequalities.py
@@ -436,7 +436,7 @@ def solve_univariate_inequality(expr, gen, relational=True, domain=S.Reals, cont
     >>> x = Symbol('x')
 
     >>> solve_univariate_inequality(x**2 >= 4, x)
-    ((2 <= x) & (x < oo)) | ((x <= -2) & (-oo < x))
+    ((2 <= x) & (x < oo)) | ((-oo < x) & (x <= -2))
 
     >>> solve_univariate_inequality(x**2 >= 4, x, relational=False)
     Union(Interval(-oo, -2), Interval(2, oo))


sympy/functions/elementary/tests/test_piecewise.py
```diff
@@ -622,6 +622,7 @@ def test_piecewise_interval():
         (S.Half, True))
 
 
+@slow
 def test_piecewise_collapse():
```
oscarbenjamin, line 626: Has this become slower or are you just adding the flag because it was already slow?

smichr: On my slow system it was very slow. I'll try remove it and see if it gets flagged by the test suite.

sympy/functions/elementary/tests/test_piecewise.py
```diff
@@ -622,6 +622,7 @@ def test_piecewise_interval():
         (S.Half, True))
 
 
+@slow
```
smichr, line 625:
```suggestion
```

sympy/functions/elementary/tests/test_piecewise.py
```diff
@@ -871,6 +872,7 @@ def test_holes():
         (Piecewise((B*x, x < 1), (nan, True)), True))
 
 
+@slow
```
smichr, line 875:
```suggestion
```

sympy/functions/elementary/tests/test_piecewise.py
```diff
@@ -1136,6 +1138,7 @@ def test_issue_8919():
                  assert ans.subs(reps) == I.subs(reps).doit()
 
 
+@slow
```
smichr, line 1141:
```suggestion
```

sympy/functions/elementary/tests/test_piecewise.py
oscarbenjamin: For me this test takes 0.75 seconds on both master and the PR (with or without gmpy2 being installed):
```console
$ pytest sympy/functions/elementary/tests/test_piecewise.py::test_piecewise_collapse
======================================== test session starts ========================================
platform linux -- Python 3.7.10, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
architecture: 64-bit
cache:        yes
ground types: python 

rootdir: /home/oscar/current/sympy/sympy.git, configfile: pytest.ini
plugins: xdist-2.3.0, forked-1.3.0
collected 1 item                                                                                    

sympy/functions/elementary/tests/test_piecewise.py .                                          [100%]

========================================= 1 passed in 0.75s =========================================
```

sympy/functions/elementary/piecewise.py
```diff
@@ -168,41 +166,7 @@ def eval(cls, *_args):
...
-                                    # new Relationals with +/-oo
-                                    reps[i] = S.true
-                                elif ('=' not in ic.rel_op and
-                                        c.xreplace({x: i.rhs}) !=
-                                        _c.xreplace({x: i.rhs})):
-                                    reps[i] = Relational(
-                                        i.lhs, i.rhs, i.rel_op + '=')
-                        c = c.xreplace(reps)
-            args.append((e, _canonical(c)))
+        args = [(e, _canonical(c)) for e, c in _args]
```
smichr, line 169:
```suggestion
        args = [(e, _canonical(c, evaluate=True)) for e, c in _args]
```

sympy/functions/elementary/tests/test_piecewise.py
smichr: all slows reverted

smichr:
Am I correct in removing `as_set` calculation from `ITE`? We don't have a Piecewise set. I don't know how we can represent, with a single set, the output of `ITE`.

oscargus:
> Am I correct in removing `as_set` calculation from `ITE`?

I am not sure. As one can rewrite an ITE to a Boolean expressions (which is what has been done earlier), I believe that it makes sense to be able to still rewrite it. To me it seems like one sometimes prefer entering it as an ITE, but it should be OK to rewrite it when beneficial. I think the same holds for e.g. Implies.

sympy/logic/boolalg.py
```diff
@@ -1533,6 +1533,10 @@ def _eval_rewrite_as_Piecewise(self, *args, **kwargs):
         from sympy.functions import Piecewise
         return Piecewise((args[1], args[0]), (args[2], True))
 
+    def as_set(self):
```
oscargus, line 1536: This should probably be `_eval_as_set`, which then already exists a few lines up. Otherwise, I guess, expressions like `And(ITE,...))` will still be valid to perform `as_set` on but not `ITE.as_set()`. But then I also think that it makes sense to keep the functionality to rewrite `ITE`.

oscargus:
Assuming that we shouldn't remove the canonicalization of Piecewise conditions, I think merging that code into the `_canonical` function is the correct thing to do. I added #22156 which instead raise an exception if the free symbol in `as_set` is not a `NumberKind` and also if the symbol in `periodicity` is not `NumberKind`.

oscarbenjamin:
Maybe `ITE(A, B, C).as_set()` could be defined as
```python
(B.as_set() & A.as_set()) | (C.as_set() - A.as_set())
```
Example:
```python
In [18]: A = x > 0

In [19]: B = x > 1

In [20]: C = x < -1

In [21]: ITE(A, B, C)
Out[21]: 
⎧x > 1   for x > 0
⎨                 
⎩x < -1  otherwise

In [22]: (B.as_set() & A.as_set()) | (C.as_set() - A.as_set())
Out[22]: (-∞, -1) ∪ (1, ∞)
```
Not sure if some additional checks would be needed to ensure that gives a sensible result.

oscargus:
That is probably correct. Now it is rewritten (`to_nnf`) as `A & B | ~A & C` which should lead to that.

sympy/core/tests/test_count_ops.py
```diff
@@ -27,6 +28,9 @@ def count(val):
     assert count(Equivalent(x,y)) == 1
     assert count(ITE(x,y,z)) == 1
     assert count(ITE(True,x,y)) == 0
+    # not all args or Basic are Basic; the following
```
smichr, line 31:
```suggestion
    # not all args of Basic are Basic; the following
```

smichr:
ping @oscargus

sympy/core/tests/test_count_ops.py
```diff
@@ -27,6 +28,9 @@ def count(val):
     assert count(Equivalent(x,y)) == 1
     assert count(ITE(x,y,z)) == 1
     assert count(ITE(True,x,y)) == 0
+    # not all args of Basic are Basic; the following
+    # might be 1 but should not fail to compute
+    assert count(OracleGate(1, lambda x: x == 2)) == 0
```
oscarbenjamin, line 33: I didn't realise what this PR was doing. I think that the OracleGate class should be fixed to have all Basic args.

We shouldn't try to accommodate for broken subclasses. They should be fixed. It is a good thing that various methods fail if a subclass breaks the invariant expected by the superclass because it makes it clear that there is a problem with that subclass.

oscarbenjamin:
This PR undoes the work in https://github.com/sympy/sympy/pull/20635 which took some time to get merged because of the various broken classes that needed to be fixed. Allowing non-Basic in the args of Basic breaks a fundamental invariant and will just be an eternal source of bugs. It is better that these things raise errors so that the author of a class like `OracleGate` can be shown straight away that they have designed the class incorrectly.

smichr:
> This PR undoes the work

Not according to this test that was given there:
```python
>>> def go():
...  from time import time
...  t=time()
...  xs = symbols('x:400')
...  b = True
...  for x in xs:
...     b &= x
...  return time()-t
...
>>> go()
9.347534656524658
```
In master it gives 9.46 (which I would just call an insignificant difference). 

In addition, until the all-Basic-args-are-Basic invariant is enforced -- and it's not, presently -- this will allow those that didn't write things like OracleGate to continue to use core functionality.

How are we going to enforce the invariant? I assume we are not going to backwards-incompatibly deprecate all such violating objects. So while they exist and no plan is in place to change them, I would like to see this in place.

smichr:
> this PR attempts to undo the bug protection that was gained by that.

I agree. Like adding randomization to the tests, we eventual expose problems.

smichr:
> can it not just use a `Lambda` instead

I need to emulate something like this:
```python
>>> f=lambda x: x==1
>>> f(1)
True
>>> Lambda(x, f(x))(1)  # doesn't work
False
```

asmeurer:
I agree with Oscar. Just look at the tons of bugs and special case code we used to have for MatrixSymbol, until we finally replaced the args with a Symbol (it used to store a string in its args). We don't see as many bugs with OracleGate because it's not used outside of its own module, and not many people use the quantum module, but I guarantee that just as many of them exist (I would present an example, but I have no idea how anything in the quantum module is supposed to work).

> I agree. Like adding randomization to the tests, we eventual expose problems.

https://github.com/sympy/sympy/issues/20914

oscarbenjamin:
The PR is now changed into something very different from what was discussed above. The current changes in the PR look good to me although I don't know the printing code so well.

asmeurer:
The PR as written looks fine to me. It's perhaps a little bit of a special case. However, I'm hesitant to add too much sorting to the printers, as we already do too much of that for Add and Mul and it leads to performance issues (i.e., the printers should *not* be calling evalf()). So while a more general sorting of And args might be called for, we should be very careful about how we do it.

PR #22157, opened 2021-09-26
Rewrote == comparisons to in

oscargus:
Replace checks with use of `in` where suitable. Better structured code and sometimes performance benefits.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/assumptions/handlers/calculus.py
+++ b/sympy/assumptions/handlers/calculus.py
@@ -101,7 +101,7 @@ def _(expr, assumptions):
         # is None and Bounded is None or there was already
         # an unknown sign, return None
         if sign != -1 and s != sign or \
-                s is None and (s == _bounded or s == sign):
+                s is None and (s in (_bounded, sign)):
             return None
         else:
             sign = s
--- a/sympy/assumptions/handlers/calculus.py
+++ b/sympy/assumptions/handlers/calculus.py
@@ -101,7 +101,7 @@ def _(expr, assumptions):
         # is None and Bounded is None or there was already
         # an unknown sign, return None
         if sign != -1 and s != sign or \
-                s is None and (_bounded is None or sign is None):
+                s is None and None in (_bounded, sign):
             return None
         else:
             sign = s
--- a/sympy/assumptions/handlers/calculus.py
+++ b/sympy/assumptions/handlers/calculus.py
@@ -101,7 +101,7 @@ def _(expr, assumptions):
         # is None and Bounded is None or there was already
         # an unknown sign, return None
         if sign != -1 and s != sign or \
-                s is None and (s in (_bounded, sign)):
+                s is None and (_bounded is None or sign is None):
             return None
         else:
             sign = s
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -703,7 +703,7 @@ class Add(Expr, AssocOp):
                 return
         if z == len(self.args):
             return True
-        if len(nz) in (0, len(self.args)):
+        if not nz or len(nz) == len(self.args):
             return None
         b = self.func(*nz)
         if b.is_zero:
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -703,7 +703,7 @@ class Add(Expr, AssocOp):
                 return
         if z == len(self.args):
             return True
-        if len(nz) == 0 or len(nz) == len(self.args):
+        if len(nz) in (0, len(self.args)):
             return None
         b = self.func(*nz)
         if b.is_zero:
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -703,7 +703,7 @@ class Add(Expr, AssocOp):
                 return
         if z == len(self.args):
             return True
-        if not nz or len(nz) == len(self.args):
+        if len(nz) in [0, len(self.args)]:
             return None
         b = self.func(*nz)
         if b.is_zero:
--- a/sympy/integrals/singularityfunctions.py
+++ b/sympy/integrals/singularityfunctions.py
@@ -55,7 +55,7 @@ def singularityintegrate(f, x):
         n = sympify(f.args[2])
         if n.is_positive or n.is_zero:
             return SingularityFunction(x, a, n + 1)/(n + 1)
-        elif n == -1 or n == -2:
+        elif n in (-1, -2):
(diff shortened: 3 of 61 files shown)


sympy/integrals/singularityfunctions.py
```diff
@@ -55,7 +55,7 @@ def singularityintegrate(f, x):
         n = sympify(f.args[2])
         if n.is_positive or n.is_zero:
             return SingularityFunction(x, a, n + 1)/(n + 1)
-        elif n in (-1, -2):
+        elif n in (S.NegativeOne, S(-2)):
```
oscarbenjamin, line 58: In general it is better to use `Integer(2)` rather than `S(-2)` to avoid going through sympify. Otherwise for clarity `-1, -2` is fine but it does imply sympification.

oscarbenjamin:
Looks good to me

sympy/core/add.py
```diff
@@ -703,7 +703,7 @@ def _eval_is_zero(self):
                 return
         if z == len(self.args):
             return True
-        if len(nz) == 0 or len(nz) == len(self.args):
+        if not nz or len(nz) == len(self.args):
```
asmeurer, line 706: Why not `len(nz) in [0, len(self.args)]` for this one?

sympy/integrals/singularityfunctions.py
asmeurer: I think `n in (-1, -2)`. is better. If that `x == 1` slow because `x` has to sympify the rhs we could improve the performance of that by improving `__eq__`.

sympy/printing/maple.py
```diff
@@ -120,9 +120,9 @@ def _print_Pow(self, expr, **kwargs):
         PREC = precedence(expr)
         if expr.exp == -1:
             return '1/%s' % (self.parenthesize(expr.base, PREC))
-        elif expr.exp == 0.5 or expr.exp == S(1) / 2:
+        elif expr.exp in (0.5, S.Half):
```
asmeurer, line 123: These lines are pointless. Floats compare equal to exactly equal rationals:

```py
>>> 0.5 == S(1)/2
True
```

This could just be `expr.exp == S.Half` and so on.

sympy/simplify/epathtools.py
```diff
@@ -52,7 +52,7 @@ def __new__(cls, path):
             index = 0
 
             for c in selector:
-                if c.isalnum() or c == '_' or c == '|' or c == '?':
+                if c.isalnum() or c in ('_', '|', '?'):
```
asmeurer, line 55: Checks like this could also be `c in '_|?'`. No preference though, as one could consider it to be less readable.

asmeurer:
A few minor comments that aren't critical. I would also generally prefer using `0` instead of `S.Zero` for readability, except in cases where the latter is required, but that also isn't critical.

Did you use a tool to produce these changes or did you find them by hand?

sympy/integrals/singularityfunctions.py
asmeurer: For instance, `Basic.__eq__` could short-circuit False on `int`, which would be overridden by `Integer.__eq__`. I expect optimizing equality checks in this way could lead to significant performance gains throughout the library, as == checks are done everywhere (any time an object is in a container like a dict or a list it is going to have a lot of == checks done on it). Right now the very first thing `Basic.__eq__` (and `Expr.__eq__`) does is call sympify, but actually constructing a sympified object can be avoided for almost every sympifyable type.

sympy/printing/maple.py
oscargus: OK! Didn't try it, although I reflected upon it... I'll change it.

sympy/simplify/epathtools.py
oscargus: Say some instance of it, but, again, just went for a minimum change.

sympy/core/add.py
oscargus: I think I had that, but then made a guess that not nz should be faster to evaluate (as checking len() == 0 is often reported as a bad idea by linters). But it is indeed a guessing game.

sympy/integrals/singularityfunctions.py
oscargus: Not sure I fully get the `__eq__` aspects. I've just noted that there is a tremendous amount of calls to `sympify` in almost every benchmark I've tested.

`Integer(-2)` is indeed better. (Except for readability...)

oscargus:
> Did you use a tool to produce these changes or did you find them by hand?

I tried a few different regex searches, but changed them by hand. Basically of the type `==.* or .*==` and variants (`is`, `!=`, `is not`, the latter with `and`). I know that pylint reports some of them.

sympy/integrals/singularityfunctions.py
oscargus: Another aspect of performance here is that `(-1, -2)` can be generated at compile time. Even though `(S.NegativeOne, Integer(-2))` are constants, it may be that they cannot.

(I know this is just a very small part of a not often used function, it is more the general concept of trying to write code in an efficient way wherever possible that I am looking for...)

asmeurer: Again, readability should be the forefront. Microbenchmarks are a waste of time, and usually end up making code harder to maintain. They should only be considered for code that gets run millions of times because it is deep in the core.

sympy/assumptions/handlers/calculus.py
```diff
@@ -101,7 +101,7 @@ def _(expr, assumptions):
         # is None and Bounded is None or there was already
         # an unknown sign, return None
         if sign != -1 and s != sign or \
-                s is None and (s == _bounded or s == sign):
+                s is None and (_bounded is None or sign is None):
```
smichr, line 104:
```suggestion
                s is None and None in (_bounded, sign):
```

sympy/simplify/epathtools.py
oscarbenjamin: That's not equivalent:
```python
>>> 'ab' in ('a', 'b')
False
>>> 'ab' in 'ab'
True
```

sympy/printing/maple.py
oscarbenjamin: Ideally they wouldn't compare equal (#20033)

Issue #21494, opened 2021-05-20
deprecate `expr_free_symbols`


smichr:
This is a general property that was added to solve a very
specific problem but it adds a layer of abstraction that is
not necessary in general.

1. objects that have structural "non-expression" nodes already
   allow one to focus on on the expression node if desired, e.g.
```python
   >>> Derivative(f(x), x).expr
   f(x)
```
   introduction of this property encourages imprecise thinking
   when requesting free_symbols since it allows one to get
   symbols from a specific node of an object without specifying
   the node

2. it incorrectly added the property to AtomicExpr so numbers
   are returned as expr_free_symbols
```python
   >>> S(2).expr_free_symbols
   2
```
3. the application of the concept was misapplied to define
   Subs.expr_free_symbols: it added in expr_free_symbols of
   the point but the point is a Tuple so nothing was added

4. it is not used anywhere else in the code base except in the
   context of differentiating a Subs object which suggests that
   it is not something of general use, this is also confirmed by
   the fact that,

5. it was added without specific tests except for test of the
   derivatives of the Subs object for which it was introduced

Upabjojr:
`expr_free_symbols` is meant to represent indexed objects such as `MatrixElement` and `Indexed` as free symbols. Currently `.free_symbols` keeps iterating over their indices.

If I have `A[i, j]` in an expression, I would like to have a way to return `{A[i, j]}` as a free symbol, and not `{A, i, j}`, which by itself doesn't make sense.

Unfortunately much of the code has already been written in order to use `{A, i, j}` as free symbols, so it's easier to introduce a new method rather than fixing the old one.

I this `.expr_free_symbols` can be useful.

oscarbenjamin:
This deprecation has not been released yet. There needs to be a clear decision before the imminent release.

The points raised by @smichr are good ones but an alternative to deprecation would be fixing the bugs and completing the implementation.

Generally though I would prefer not to keep adding attributes and methods to Expr which already has far too many. This would be better as a public function even if it uses an internal method to allow extensibility by different classes. At least the implementation should be complete before adding any new API to Expr/Basic because it creates future problems with the need for incompatible changes.

I suppose the need for something like `expr_free_symbols` is related to #20809 and #20809.

smichr:
> There needs to be a clear decision before the imminent release.

Please get rid of the current implementation for all the reasons I cite. And perhaps create a new method `free_expressions` (after thorough discussion) that does not conflate symbols and non-symbols.

[...]

PR #22156, opened 2021-09-26
Fixed two issues pointed out related to #14933

oscargus:
#### References to other Issues or PRs

Fixes #14933

#### Brief description of what is fixed or changed
Two of the three points in https://github.com/sympy/sympy/issues/14933#issuecomment-925300998 are fixed here. This avoids the error in #14933 if nothing else, but it may still be slower than before. 

#### Other comments
This is independent of #22152.

#### Release Notes

NO ENTRY

PR #22158, opened 2021-09-26
maint(core): undeprecate expr_free_symbols

oscarbenjamin:
The expr_free_symbols property was deprecated in GH-21485 but as
discussed in GH-21494 was being used by pycalphad because of the bug
reported there. It is too late in the release process to fix the
underlying issues so this commit undeprecates expr_free_symbols in the
1.9 branch while a broader fix is worked out for the SymPy 1.10 release.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -509,10 +509,6 @@ class Basic(Printable, metaclass=ManagedProperties):
 
     @property
     def expr_free_symbols(self):
-        from sympy.utilities.exceptions import SymPyDeprecationWarning
-        SymPyDeprecationWarning(feature="expr_free_symbols method",
-                                issue=21494,
-                                deprecated_since_version="1.9").warn()
         return set()
 
     def as_dummy(self):
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -2430,9 +2430,6 @@ class Expr(Basic, EvalfMixin):
         Examples
         ========
 
-        >>> from sympy.utilities.exceptions import SymPyDeprecationWarning
-        >>> import warnings
-        >>> warnings.simplefilter("ignore", SymPyDeprecationWarning)
         >>> from sympy.abc import x, y
         >>> (x + y).expr_free_symbols
         {x, y}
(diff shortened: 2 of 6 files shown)


smichr:
Doesn't the #22156 solve this problem?

oscarbenjamin:
> Doesn't the #22156 solve this problem?

I'm losing track of what problem it is that precisely needs to be fixed any more...

@richardotis @bocklund does #22156 solve the problem for pycalphad?

smichr:
> solve the problem for pycalphad?

I suspect this solves the recursion but not the slow-down.

richardotis:
@smichr Correct. It works but it's still slow.

We really appreciate all of you investigating this issue and going the extra mile for us. We're re-architecting downstream to no longer rely on the behavior that has the slow-down, so issues with pycalphad shouldn't be considered a blocker here anymore. In any case, we can stay pinned to sympy 1.8 until we get our fixes implemented and released.

cc @bocklund

oscarbenjamin:
> shouldn't be considered a blocker here anymore

Great. Do you want any of the various fixes for the next release? Otherwise I'll just release 1.9rc1 as 1.9.

oscarbenjamin:
> > shouldn't be considered a blocker here anymore
> 
> Great. Do you want any of the various fixes for the next release? Otherwise I'll just release 1.9rc1 as 1.9.

I'll put the release out tomorrow

richardotis:
I saw that #22156 was merged, and I don't think anything else is needed.

oscarbenjamin:
> I saw that #22156 was merged, and I don't think anything else is needed.

It was merged to master but not the 1.9 branch which is what will be used for the release. I'll open a PR to port it to 1.9...

oscarbenjamin:
> I'll open a PR to port it to 1.9...

Or someone else can. Bedtime for me...

richardotis:
Even without the fix, we'll be fine downstream. Do what you think is best. From our perspective, it can wait.

oscarbenjamin:
Okay, I'll close this

PR #22035, opened 2021-09-06
[WIP] Removed as_set in Piecewise eval

oscargus:
#### Brief description of what is fixed or changed

Testing out  #21360
Fixes #14933

#### Release Notes

NO ENTRY

PR #22162, opened 2021-09-27
move more expensive simplification steps from Piecewise.eval to simplify function

### LINKED ISSUE #22175: Incorrect results with Piecewise expression MatrixSymbol substitutions.

```python
x = Symbol("x")
print(fn)
print(subfn)
# 1.0
```

Thanks

smichr:

fixes #22175 
closes #22035 while keeping the functionality available through simplify

As discussed in #14933, there was code that causes a significant slow down in Piecewise work. This code has now been moved to a simplify-related function for Piecewise.

Better handling of `And` expressions containing Relationals with infinite bounds are handled better, now. Formerly, bounds that didn't include infinity were returned as including it.

#### Release Notes

* functions
  * Piecewise does less automatic simplication at instantiation; old functionality is available through `piecewise_simplify_arguments` (which is available through `simplify`).


smichr:
ping @richardotis @bocklund, this may solve the speed issues you were having with Piecewise.

richardotis:
@smichr On my local testing this PR (at `cefa96a`) preserves both correctness and performance in pycalphad, compared to sympy 1.8, but without us doing patching downstream. (cc @bocklund this was tested with https://github.com/pycalphad/pycalphad/pull/375 at `542c3b`)

As I mentioned before, pycalphad's latest version is currently pinned to sympy 1.8, so it's not time-sensitive for this to make it into sympy's next release, but we are happy to see progress made here. Thanks for your work.

sympy/functions/elementary/piecewise.py
```diff
@@ -1189,9 +1158,58 @@ def _clip(A, B, k):
...
+                                    c = x >= s.inf
+                                else:
+                                    c = x > s.inf
+                            elif c.xreplace({x: s.inf}) == True:
+                                c = And(x >= s.inf, x < s.sup)
+                            else:
+                                c = And(x > s.inf, x < s.sup)
+                    else:
+                        c = s.as_relational(x)
+                except NotImplementedError:
```
oscarbenjamin, line 1210: The try/except should be around precisely the statement from which the error is to be caught i.e. the `as_set` call.

sympy/integrals/integrals.py
```diff
@@ -468,6 +469,13 @@ def doit(self, **hints):
         reps = {}
         for xab in self.limits:
             if len(xab) != 3:
+                # it makes since to just make
```
oscargus, line 472: sense?

This happens (sort of) a number of lines down and there is a "better" approach in #21985 (in the sense that the Dummy-symbols will inherit properties from the bounds). However, that is only for definite integrals.

oscargus:
I'm wondering how much of this is handled by #17330? Still it is better to just be able to call it separately as here.

(I'm also a bit worried about the merge conflicts for #17443, which would be nice to get in soon.)

oscargus:
Also wondering how much of this will be sorted by merging #22052? Seems like at least some of the issues with setting real/extended_real to Symbols/Dummys are sorted out by this. I'd say we merge that quite soon (it also affects some of my recent PRs).

In general I think that it is good to refactor the code like this. Still. the requirement to run simplify after the first of the two integrals in a double-integral is a bit unintuitive... But hopefully it is possible to continue working on the root cause of that.

(I'm about to close #22035 anyway.)

smichr:
OK, there were two issues. The one impacting integration was the expectation that a condition like `x <= oo` be treated as `True` in the integration context. Without the simplification provided by that portion of `Piecewise.eval` certain itegrals will fail. The other issue was that a relationship that didn't initially include `-oo` was incorrectly written to include it.

Knowing this, it will be easy to work around your PRs in progress. So let's do those first, if you want.

sympy/functions/elementary/piecewise.py
```diff
@@ -273,22 +238,23 @@ def eval(cls, *_args):
...
                     continue
 
-            newargs.append(ExprCondPair(expr, cond))
+            newargs.append((expr, cond))
 
         # some conditions may have been redundant
         missing = len(newargs) != len(_args)
         # some conditions may have changed
-        same = all(a == b for a, b in zip(newargs, _args))
+        same = all(a == b for a, b in zip(newargs, args))
```
oscarbenjamin, line 257: Why is this checking `args` but for `missing` above it checks `_args`?

Can this not just be `if newargs != args`?

oscarbenjamin:
This is a good improvement although I haven't studied in detail.

Should this use `evaluate=False`?
https://github.com/sympy/sympy/blob/9831987abba5ba3eab0786c1d524141e9896a804/sympy/functions/elementary/piecewise.py#L265-L266

oscarbenjamin:
Also it looks like `ExprCondPair` does not respect `evaluate=False` and `Piecewise.__new__` does not check `globalevaluate`.

sympy/functions/elementary/piecewise.py
smichr: input `_args` was changed to `args` and is a tuple while newargs is a list. So instead of converting one type to the other I ust compare them element-wise

smichr:
> Should this use `evaluate=False`?

no, changes are made recursively until the expression is stable. e.g. `Piecewise((x, x < 1), (x, x < 2), (x, True))` needs to recurse.

sympy/integrals/tests/test_integrals.py
```diff
@@ -1344,37 +1343,34 @@ def test_issue_2884():
...
+    x4 = Abs(arg(s))
+    x5 = 2*x4 < pi
+    x6 = sqrt(x2)*cos(x4) - 1 > 0
+    cse_ans = [
+    Piecewise((pi*Piecewise((-s*x0/(1 - x1), x2 < 1), (x0/(s*(1 - 1/x1)),
+    1/x2 < 1), (meijerg(((1/2,), (0, 0)), ((0, 1/2), (0,)),
+    polar_lift(s)**2), True)), (x3 & x5 & x6 & (s > 1)) | (x3 & x5 & x6 &
+    (s < -1)) | (x3 & x5 & x6 & (s > -1) & (s < 1))),
+    (Integral(exp(-s*x)*cosh(x), (x, 0, oo)), True))]
+    assert got == cse_ans[0]
```
oscarbenjamin, line 1360: This part of the diff makes me a little nervous. I can't tell how much the result has actually changed just from the diff here.

smichr:
> like `ExprCondPair` does not respect `evaluate=False`

The only thing it evaluates is Piecewise condition to make sure that it becomes a boolean. But I added an evaluate option for `piecewise_fold`. (Not sure if that is useful or not, though.)

sympy/integrals/tests/test_integrals.py
smichr: I tested the old and new expressions with a few random() numbers.

smichr:
```python
for i in range(100):
  p = Rational(randint(1,30),randint(1,30))
  if not e(p):
   print(p)
   break
```

smichr: But that second expression has a cosh instead of a sinh...checking

smichr: doh - checked the first against the second. When checking each against the correct old Piecewise and the new cse, the loop never emits.

PR #22168, opened 2021-09-28
Add integral basis and prime factorization for number fields.

skieffer:
### Brief description of what is fixed or changed

Hello, and thanks for your work on this amazing project!

I thought it might be good to extend functionality for computational algebraic number theory, so I have implemented calculation of:

* integral bases for number fields, via Zassenhaus's Round 2 algorithm
* prime decomposition in number fields

The new features are meant to be accessed through an `AlgebraicField`:

```python
  >>> from sympy import Poly, QQ
  >>> from sympy.abc import x, theta
  >>> T = Poly(x ** 3 + x ** 2 - 2 * x + 8)
  >>> K = QQ.algebraic_field((T, theta))
  >>> K.integral_basis
  Matrix([
  [1, 0,   0],
  [0, 1, 1/2],
  [0, 0, 1/2]])
  >>> K.discriminant
  -503
  >>> K.primes_above(2)
  [[ (2, x**2 + 1) e=1, f=1 ], [ (2, (x**2 + 3*x + 2)/2) e=1, f=1 ], [ (2, (3*x**2 + 3*x)/2) e=1, f=1 ]]

```

### Other comments

The new code needed a home, and it seemed to me that the most appropriate thing was to turn the existing module `sympy.polys.numberfields` into a whole package.

Also some supporting functionality was needed in other areas of the library.

Note: I have left some TODOs and FIXMEs, doctext should be improved, more unit tests are needed... I thought this is okay since this is just a draft PR.

Note: I have split my changes into three commits to make them easier to inspect:
1. First commit: General purpose functionality.
2. Second commit: Turning `sympy.polys.numberfields` into a package.
3. Third commit: The new functions for algebraic number theory.

### Questions

I guess my main question at this point is whether classes I have defined in `sympy.polys.numberfields.modules` ought to be subclasses of similarly-named classes defined in `sympy.polys.agca`. (Modules, Ideals, Homomorphisms, etc.)

TBH, I have not yet studied `agca` too, too closely. I was hoping to get some feedback first.

Maybe these classes could converge in the future, as the need/utility became more clear?

#### Release Notes

* core
  * Provide a `binpow` function: Raise to power efficiently, using binary expansion.
* matrices
  * Make the `_to_domain()` function defined in `.normalforms` available as a method of the `MatrixBase` class.
  * Provide an `hermite_normal_form()` function.
  * Provide a `supplementary_space()` function.
* polys
  * Supply missing `setitem` method in `DDM` class.
  * Supply missing `columnspace` and `rowspace` methods in `DomainMatrix`.
  * Split `numberfields` into three modules: `minpoly`, `isomorphism`, `numbers`.
  * Add calculation of integral bases for number fields in `basis.py`.
  * Add prime factorization in number fields in `primes.py`.
  * Other supporting functionality and classes.


sympy/polys/numberfields/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+from .basis import nilradical_mod_p, round_two
+
+from .primes import prime_valuation, prime_decomp
+
+# Before `sympy.polys.numberfields` was a package, it was a module, containing
+# what is now defined in the `minpoly`, `isomorphism` and `numbers` modules.
+# To maintain backwards compatibility, we have to import everything that was
+# available in the old `sympy.polys.numberfields` module.
+# The lists below may contain some (harmless) redundancies between them, since
+# each was automatically generated by scanning the module.
```
oscargus, line 24: I think that one probably should remove, at least, the obvious ones not originating from those modules (like S, Rational, Add, Dummy, etc). 

Also, you will need to mark these to avoid the test errors, I think with `# noqa: F401`.

oscargus:
Nice work! Always nice to see a major contribution. (For the technical aspects, I cannot really say a lot.)

sympy/core/power.py
```diff
@@ -181,6 +181,35 @@ def integer_log(y, x):
...
+    """
+    Raise base b to integer exponent e > 1 efficiently, using binary
+    representation of e.
+
+    (The reason for not supporting e == 1 is that in that case b itself would
+    be returned, while in all other cases a new instance is computed. Callers
+    should handle the case e == 1 separately, so they can return a clone of b,
+    rather than b itself.)
+    """
+    e = _sympify(e)
```
oscargus, line 194: I think that you should either sympify b as well to return a sympified result or not sympify e and get the speed benefit of integer arithmetic (and check e.is_integer in another way, like `int(e) == e`).

Considering that this is in the core it should probably return a sympified result, but that can of course also happen at the return.

sympy/core/tests/test_power.py
```diff
@@ -630,3 +630,11 @@ def test_powers_of_I():
         1, sqrt(I), I, sqrt(I)**3, -1, -sqrt(I), -I, -sqrt(I)**3,
         1, sqrt(I), I, sqrt(I)**3, -1]
     assert sqrt(I)**(S(9)/2) == -I**(S(1)/4)
+
+
+def test_bin_pow():
+    raises(ValueError, lambda: binpow(3, 1))
+    raises(ValueError, lambda: binpow(3, 1/2))
```
oscargus, line 637: Note the difference between `1/2 = 0.5` and `S(1)/2 = Rational(1, 2) = S.Half`. The test will raise anyway, more if the purpose was to test for a Rational number or a float. What about `binpow(3, 2.0)`? Should that work and in that case how?

sympy/core/power.py
```diff
@@ -181,6 +181,35 @@ def integer_log(y, x):
     return e, r == 0 and y == 1
 
 
+def binpow(b, e):
```
jksuom, line 184: It seems that Python implements this algorithm [in C](https://github.com/python/cpython/blob/953e733e6d0bc02fde0e4f808259218a95a3b902/Objects/longobject.c#L4328-L4360). That would be more efficient.

jksuom:
> whether classes I have defined in sympy.polys.numberfields.modules ought to be subclasses of similarly-named classes defined in sympy.polys.agca. (Modules, Ideals, Homomorphisms, etc.)

I don't think that it would be necessary. The structures are different.

sympy/matrices/matrices.py
```diff
@@ -2235,6 +2236,9 @@ def strongly_connected_components(self):
     def strongly_connected_components_decomposition(self, lower=True):
         return _strongly_connected_components_decomposition(self, lower=lower)
 
+    def to_domain(self, domain=None):
+        return _to_domain(self, domain=domain)
```
oscarbenjamin, line 2240: The `_to_domain` function is a hack that should not be made public. It is just there as an attempt to maintain backward compatibility for the deprecated way of using the `smith_normal_form` function:
https://github.com/sympy/sympy/blob/ed6a33049b42a16b87a6d8685dd012594aa62dfa/sympy/matrices/tests/test_normalforms.py#L31-L39

There could be a `to_domain` method for `Matrix` but it should not be based on the one `normalforms.py`.

Rather these algorithms should just be implemented for `DomainMatrix` in `sympy/polys/matrices` (like `smith_normal_forms` now is). Then there can be a convenience method for Matrix that calls into the lower-level routine.

See fa6b1f3494d909c43bb589d937b01e72e1f476b2 and the associated PR #21402 for background to why the code is the way that it is.

oscarbenjamin: A `to_domain` method would be good but it should just be the function before the changes in fa6b1f3

sympy/polys/numberfields/forms.py
```diff
@@ -0,0 +1,449 @@
...
+
+    def __mul__(self, r):
+        if isinstance(r, StandardRep):
+            f, g = self.poly(), r.poly()
+            e, h = (f * g % self.T).clear_denoms()
+            return StandardRep.from_poly(self.T, h) // e
+        if is_rat(r):
+            r = sympify(r)
+            a, b = r.as_numer_denom()
+            s = StandardRep(self.T, [a*c for c in self.coeffs], b*self.denom)
```
oscarbenjamin, line 158: If this is supposed to use the ZZ domain then `sympify` should not be used as it gives `Integer` rather than `MPZ`. SymPy `Expr` types are infectious: if a single one creeps in it can ruin everything.

I suggest adding a check in `__init__`:
```python
assert all(ZZ.of_type(c) for c in coeffs)
```
If the check slows things down then it can be disabled later but at least the internals here need to be sorted out first.

sympy/polys/numberfields/forms.py
```diff
@@ -0,0 +1,449 @@
...
+        if not W.is_square or not W.is_upper:
+            raise ValueError('HNF matrix must be square and upper triangular')
+        super().__init__(W, d)
+        self.T = T
+        self._mult_tab = None
+
+    @classmethod
+    def for_power_basis(cls, T):
+        from sympy.matrices import Matrix
+        return cls(T, Matrix.eye(T.degree()), 1)
```
oscarbenjamin, line 260: The polys module should not use the matrices module. This should be DomainMatrix.

sympy/polys/numberfields/numbers.py
```diff
@@ -0,0 +1,96 @@
...
+
+        if coeffs is not None:
+            return AlgebraicNumber(theta, coeffs)
+        else:
+            raise IsomorphismFailed(
+                "%s is not in a subfield of %s" % (root, theta.root))
+
+
+class IntervalPrinter(MpmathPrinter, LambdaPrinter):
+    """Use ``lambda`` printer but print numbers as ``mpi`` intervals. """
```
oscarbenjamin, line 48: I realise that this PR is just moving this into a package but I really don't understand why this class is defined here.

oscarbenjamin:
This looks like a lot of excellent work. Thanks for this!

I'm finding it hard to review though because it's so much... You have separated this nicely into commits but I think I would rather review each of those as a separate PR. For example just the first commit has a much more manageable diff but already in there some changes are needed. I think it would be helpful to go through a much smaller diff like that in detail. Likewise a refactoring change should be reviewed and merged with the changes being as close as possible to a pure refactor without any other observable change.

Could you perhaps open a PR for just the first commit and then another for just the second commit?

I imagine it's not easy to disentangle the third from the second so it would be better just to try and get the second (refactoring) merged and then rebase the third commit with all the new stuff on top.

In general there is a mixing of sympified and unsympified types in a few places and the code sometimes uses Matrix and sometimes DomainMatrix. I think that there should be a clear separation between the lower-level implementation algorithms in the polys module built in terms of domains and DomainMatrix and the higher-level classes like Expr and Matrix that are built on top. What that means is that anything in polys that needs to work with matrices should be using DomainMatrix: in simple terms there should be no imports from the matrices module in the polys module.

There are a few places in polys that use Matrix but it's because they predate DomainMatrix (they should be updated). Likewise the `matrices.normalforms` module should really be in `polys` but it predates DomainMatrix. Now that we have DomainMatrix and it is used by all other matrix classes we can make these things somewhat less circular.

sympy/polys/numberfields/__init__.py
skieffer: Okay, I agree that omitting `Rational`, `Add`, etc. is cleaner and is probably the right thing. My intention was to cater to those users who may have -- against all good coding principles! -- imported such things _from this module_. But, I think we could hardly be faulted for failing to support that kind of behavior!

May need some clarification on the `# noqa: F401` thing, but will cross that bridge when I come to it.....

sympy/core/power.py
skieffer: Aha! Thank you for the link. You know, I was wondering if this might have been the case, but didn't know how to investigate it.

Well, my question now is whether there's a good way to bring Python's internal powering algorithm to apply to powers of user-defined classes. In other words, suppose I'm defining `class Foo` and I've defined `Foo.__mul__`. Is there a way to then tell Python that, when it sees

```python
f = Foo()
g = f ** 3
```

it is to use its internal binary powering algorithm, while calling on `Foo.__mul__` each time it needs to actually multiply two `Foo`s together in the process? If so, that would be great!

sympy/core/power.py
```diff
@@ -181,6 +181,35 @@ def integer_log(y, x):
...
+
+    (The reason for not supporting e == 1 is that in that case b itself would
+    be returned, while in all other cases a new instance is computed. Callers
+    should handle the case e == 1 separately, so they can return a clone of b,
+    rather than b itself.)
+    """
+    e = _sympify(e)
+    if not e.is_integer or e < 2:
+        raise ValueError('Exponent must be integer greater than 1.')
+    bits = [int(d) for d in reversed(bin(e)[2:])]
```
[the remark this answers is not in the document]
skieffer: Not sure....will think about it......

sympy/core/power.py
skieffer: I see your point regarding speed vs sympifying. Should be one or the other.

Well, as I think more about it, I am beginning to think `binpow` really doesn't belong in `core`. Certainly, I want to be able to use it to compute powers of bases `b` that are "not sympifyable", such as my custom classes.

sympy/matrices/matrices.py
skieffer: Okay, so I will move my `hermite_normal_form` functions into `polys.matrices.normalforms`.

skieffer:
Thank you for your kind words, and all of this useful feedback!

Okay, I am starting to understand the architecture better, and in particular that there's meant to be a division between the way `polys` does things, and the way `matrices` and `core` do things. We want to stick to just domain elements here, and don't want to use things that are subclasses of `core.basic.Basic`. Got it.

Also, the discussion on `binpow` is making me think this function should come out of `core`, and instead go into a utilities module at some level. (Unless there's a way to leverage Python's internal version, and then this just isn't needed.)

---------

The plan of processing the three "layers" of this PR one PR at a time sounds good.

So, let's talk about the first layer.

As I go back and put everything belonging to "layer 3" in terms of DomainMatrices, I'm probably going to want to implement a couple of new methods in the `DomainMatrix` class, to serve as counterparts to some of the convenience methods already offered by `Matrix`. For example, I use a lot of `row_join()`.

In other words, the revisions are going to make "layer 1" grow a bit. So, would you be happy for me to go back, do all the revisions first, and only then begin opening the new PRs? (Just to get it all done at once?)

sympy/polys/matrices/domainmatrix.py
```diff
@@ -1170,15 +1170,71 @@ def rref(self):
...
+        ...    [QQ(2), QQ(-2)]], (2, 2), QQ)
+        >>> A.rowspace()
+        DomainMatrix([[1, -1]], (1, 2), QQ)
+
+        """
+        if not self.domain.is_Field:
+            raise ValueError('Not a field')
+        rref, pivots = self.rref()
+        rows, cols = self.shape
+        return self.extract(pivots, range(cols))
```
skieffer, line 1227: Oops, I think this should actually be

```python
return self.extract(range(len(pivots)), range(cols))
```

I'll fix this.

sympy/polys/matrices/tests/test_domainmatrix.py
```diff
@@ -438,6 +438,36 @@ def test_DomainMatrix_rref():
...
+
+
+def test_DomainMatrix_rowspace():
+    A = DomainMatrix([[QQ(1), QQ(-1)], [QQ(2), QQ(-2)]], (2, 2), QQ)
+    Arow = DomainMatrix([[QQ(1), QQ(-1)]], (1, 2), QQ)
+    assert A.rowspace() == Arow
+
+    Az = DomainMatrix([[ZZ(1), ZZ(-1)], [ZZ(2), ZZ(-2)]], (2, 2), ZZ)
+    raises(ValueError, lambda: Az.rowspace())
+
```
skieffer, line 463: - [ ] Let's use non-square matrices in these unit tests, which will be better for exposing errors.

sympy/core/power.py
oscarbenjamin: I think that the C version is only for `int`. This probably does not belong in core though. I don't think that core is a good place for this (especially if it isn't used there) but maybe `sympy.external` can do for now.

oscarbenjamin:
> In other words, the revisions are going to make "layer 1" grow a bit. So, would you be happy for me to go back, do all the revisions first, and only then begin opening the new PRs? (Just to get it all done at once?)

I think that's fine. The `DomainMatrix` class is still a work in progress so it's not surprising if it needs new additions. I would like it to have cautiously defined additions rather than just slowly grow all the same methods as Matrix (which IMO has too many methods some of which are not useful). If it is lacking a key primitive though then that should be added.

In particular if there is something that has very different performance characteristics for dense vs sparse then I think it makes sense to define that at the DomainMatrix (rather than DDM or SDM) level. Ideally DomainMatrix and DDM/SDM would provide a small number of efficiently implemented operations. Those operations should be sufficient for building higher-level algorithms without needing to fiddle around with the individual elements of the matrix and with performance that is favourable to both the dense and sparse implementations so e.g. the higher-level algorithm just calls rref and doesn't need to worry about how that is implemented.

sympy/matrices/normalforms.py
```diff
@@ -63,3 +64,180 @@ def invariant_factors(m, domain=None):
...
+                for jj in range(n - k):
+                    for ii in range(m):
+                        W[ii, jj] = A[ii, jj + k]
+                return W
+            else:
+                i -= 1
+                k -= 1
+                j = k
+                state = 2
+
```
jksuom, line 139: I think that this could be coded using for-loops instead of explicitly setting, decrementing and testing the index values. Maybe something like this:

    for i in range(n - 1, ell - 1, -1):
        for j in range(k - 1, -1, -1):
            if A[i, j] == 0:
                continue
            do state 3 stuff
        do state 4 stuff up to if i == ell: line
        decrement k
    compute and return W

skieffer: Ah yes, I meant to come back to this. Was working from a text that uses numbered Steps and `GOTO`s, so a state-based system was just the most immediate translation. But I think you're right, it should be "modernized" and put into ordinary looping constructs. Will do...

PR #22169, opened 2021-09-28
parse_expr converts builtin abs, max, min, pow

### LINKED ISSUE #17811: sympify("abs(x)*5",evaluate=False) crashes
Long story short: The following program crashes:
```
import sympy
sympy.sympify("abs(x)*5",evaluate=False)
```
with the message:
>   File "bugreport2.py", line 3, in <module>
>     sympy.sympify("abs(x)*5",evaluate=False)
>   File "/home/volker/.local/lib/python3.7/site-packages/sympy/core/sympify.py", line 384, in sympify
>     expr = parse_expr(a, local_dict=locals, transformations=transformations, evaluate=evaluate)
>   File "/home/volker/.local/lib/python3.7/site-packages/sympy/parsing/sympy_parser.py", line 964, in parse_expr
>     code = compile(evaluateFalse(code), '<string>', 'eval')
>   File "/home/volker/.local/lib/python3.7/site-packages/sympy/parsing/sympy_parser.py", line 974, in evaluateFalse
>     node = EvaluateFalseTransformer().visit(node)
>   File "/usr/lib/python3.7/ast.py", line 262, in visit
>     return visitor(node)
>   File "/usr/lib/python3.7/ast.py", line 317, in generic_visit
>     value = self.visit(value)
>   File "/usr/lib/python3.7/ast.py", line 262, in visit
>     return visitor(node)
>   File "/usr/lib/python3.7/ast.py", line 326, in generic_visit
>     new_node = self.visit(old_value)
>   File "/usr/lib/python3.7/ast.py", line 262, in visit
>     return visitor(node)
>   File "/home/volker/.local/lib/python3.7/site-packages/sympy/parsing/sympy_parser.py", line 1047, in visit_BinOp
>     new_node.args = self.flatten(new_node.args, sympy_class)
>   File "/home/volker/.local/lib/python3.7/site-packages/sympy/parsing/sympy_parser.py", line 996, in flatten
>     if isinstance(arg, ast.Call) and arg.func.id == func:
> AttributeError: 'Call' object has no attribute 'id'

sympy.\_\_version\_\_ is '1.4' and sys.version is '3.7.3 (default, Oct  7 2019, 12:56:13) \n[GCC 8.3.0]'

eagleoflqj:

Fixes #17811

See https://github.com/sympy/sympy/issues/17811#issuecomment-547191254
#### Release Notes

* parsing
  * parse_expr converts builtin abs, max, min, pow to sympy functions

### DIFF
--- a/sympy/parsing/sympy_parser.py
+++ b/sympy/parsing/sympy_parser.py
@@ -8,6 +8,8 @@ from keyword import iskeyword
 import ast
 import unicodedata
 from io import StringIO
+import builtins
+import types
 
 from sympy.assumptions.ask import AssumptionKeys
 from sympy.core.basic import Basic
@@ -15,7 +17,7 @@ from sympy.core import Symbol
 from sympy.core.function import arity, Function
 from sympy.utilities.iterables import iterable
 from sympy.utilities.misc import filldedent, func_name
-
+from sympy.functions.elementary.miscellaneous import Max, Min
 
 
 def _token_splittable(token):
@@ -1072,6 +1074,14 @@ def parse_expr(s, local_dict=None, transformations=standard_transformations,
                 raise TypeError(filldedent('''
                     a transformation should be function that
                     takes 3 arguments'''))
+
+    builtins_dict = vars(builtins)
+    for name, obj in builtins_dict.items():
+        if isinstance(obj, types.BuiltinFunctionType):
+            global_dict[name] = obj
+    global_dict['max'] = Max
+    global_dict['min'] = Min
+
     code = stringify_expr(s, local_dict, global_dict, transformations)
 
     if not evaluate:
--- a/sympy/parsing/tests/test_sympy_parser.py
+++ b/sympy/parsing/tests/test_sympy_parser.py
@@ -318,3 +318,15 @@ def test_parsing_definitions():
     assert T[:5, 8] == standard_transformations + (t[8],)
     assert parse_expr('0.3x^2', transformations='all') == 3*x**2/10
     assert parse_expr('sin 3x', transformations='implicit') == sin(3*x)
+
+
+def test_builtins():
+    cases = [
+        ('abs(x)', 'Abs(x)'),
+        ('max(x, y)', 'Max(x, y)'),
+        ('min(x, y)', 'Min(x, y)'),
+        ('pow(x, y)', 'Pow(x, y)'),
+    ]
+    for built_in_func_call, sympy_func_call in cases:
+        assert parse_expr(built_in_func_call) == parse_expr(sympy_func_call)
+    assert str(parse_expr('pow(38, -1, 97)')) == '23'


oscargus:
Seems to make sense! I'm not really experienced with advanced usage of the parsing, but I think it would make even more sense if this was added independently of the user providing a `global_dict` (or otherwise this functionality would break unless the user re-adds them explicitly). Not sure how that should be done though. Can one update the dict with these four if they are not already present? Should it be possible to avoid using them?

eagleoflqj:
Any comments? @asmeurer
What about
```py
def parse_expr(s, local_dict=None, transformations=standard_transformations,
global_dict=None, evaluate=True, convert_builtins=(abs, max, min, pow)):
```
that still convert these four by default, but allow user to assign a false value or a subset tuple?

smichr:
Don't these automatically call the SymPy equivalent? That way you don' thave to worry about ever importing these; Python will just do the right thing and call the SymPy version.

eagleoflqj:
> Don't these automatically call the SymPy equivalent? That way you don' thave to worry about ever importing these; Python will just do the right thing and call the SymPy version.

I'm afraid no on current master:
```py
>>> import sympy
>>> sympy.__version__
'1.10.dev'
>>> sympy.parse_expr('abs(1)')
abs(1)
>>> sympy.parse_expr('Abs(1)')
1
```

oscargus:
> What about

Something like that would make sense, yes. I am not fully convinced about using a tuple with functions though, but not really sure exactly how to do it.

Maybe have a default value of `None` and then accept a list with string names? That is probably the easiest way since to can then create the dict and update `global_dict` with those items (or all for `None`)? 

Although it is somehow more correct to have the default value as you do, it is typically not how it is done in Sympy. Also, it may lead to a quite long signature in editors supporting it.

sympy/parsing/sympy_parser.py
```diff
@@ -989,7 +989,15 @@ def parse_expr(s, local_dict=None, transformations=standard_transformations,
         raise TypeError('expecting local_dict to be a dict')
 
     if global_dict is None:
-        global_dict = {}
+        from sympy.core.power import Pow
+        from sympy.functions.elementary.complexes import Abs
+        from sympy.functions.elementary.miscellaneous import Max, Min
```
oscargus, line 994: Please move these imports to the top level (if possible, some may have to stay here to avoid cyclic imports).

sympy/parsing/sympy_parser.py
```diff
@@ -989,7 +989,15 @@ def parse_expr(s, local_dict=None, transformations=standard_transformations,
...
+        global_dict = {  # map built-in functions to sympy functions
+            'abs': Abs,
+            'max': Max,
+            'min': Min,
+            'pow': Pow,
```
asmeurer, line 999: This should be `pow`, which will still work because sympy types define `__pow__`, but will also support the 3-argument version.

asmeurer:
I would expect all builtin functions to just parse as themselves. Then abs and pow would not need to be handled separately. min and max still make sense though.

asmeurer:
I'm not really sure this needs to be a flag. What's the use-case for disabling built-ins parsing?

eagleoflqj:
> I'm not really sure this needs to be a flag. What's the use-case for disabling built-ins parsing?

idk, maybe just provide a way compatible to previous behavior?

smichr:
> maybe just provide a way compatible to previous behavior?

This is stalled because of this. Let's just leave that option off until we have good reason to add it. People should not be using builtins as personal function names in general. And if they do, can't they just pass a `local_dict` indicating that the function should not be parsed as a SymPy function?

eagleoflqj:
ping @smichr

oscargus:
LGTM!

eagleoflqj:
ping @oscargus
Let's merge before next conflict?

eagleoflqj:
ping @oscargus

eagleoflqj:
ping @oscargus @smichr

oscargus:
I do hardly recall the discussion, but the current solution seems to make sense. Thanks!

Volker-Weissmann:
Thanks for (finally) fixing https://github.com/sympy/sympy/issues/17811

oscarbenjamin:
> > I'm not really sure this needs to be a flag. What's the use-case for disabling built-ins parsing?
> 
> idk, maybe just provide a way compatible to previous behavior?

Exactly. The change here broke the `unyt` library (see #23148). Because the parser allows any unrecognised identifiers to become symbols it is necessary for the user to be able to control the `global_dict`.

PR #22170, opened 2021-09-28
series - added fix for sympifying expr in Order.contains

### LINKED ISSUE #22165: `contains` assumes its argument to be sympyfied already
```
In [1]: f = log(x)

In [2]: g = 2

In [3]: O(f).contains(g)
<ipython-input-3-3016094d13c3> in <module>
----> 1 O(f).contains(g)

~/sympy/sympy/core/cache.py in wrapper(*args, **kwargs)
     70         def wrapper(*args, **kwargs):
     71             try:
---> 72                 retval = cfunc(*args, **kwargs)
     73             except TypeError as e:
     74                 if not e.args or not e.args[0].startswith('unhashable type:'):

~/sympy/sympy/series/order.py in contains(self, expr)
    372         """
    373         from sympy import powsimp
--> 374         if expr.is_zero:
    375             return True
    376         if expr is S.NaN:

AttributeError: 'int' object has no attribute 'is_zero'
```
`Order.contains` assumes that its argument `expr` has already been sympified, which leads to an error being thrown in a case like shown above. Just adding a line to `sympify` `expr` should be enough, so this should be easy to fix

praneethratna:
Fixes #22165 by sympifying `expr` in the contains method and now `O(log(x)).contains(2)` returns `True` instead of error. Also added a test case function `test_issue_22165()` .

#### Release Notes

* series
   * fixed bug in Order.contains

### DIFF
--- a/sympy/series/order.py
+++ b/sympy/series/order.py
@@ -371,6 +371,7 @@ class Order(Expr):
         (e.g. when self and expr have different symbols).
         """
         from sympy import powsimp
+        expr = sympify(expr)
         if expr.is_zero:
             return True
         if expr is S.NaN:
--- a/sympy/series/tests/test_order.py
+++ b/sympy/series/tests/test_order.py
@@ -446,3 +446,7 @@ def test_issue_15539():
 
 def test_issue_18606():
     assert unchanged(Order, 0)
+
+
+def test_issue_22165():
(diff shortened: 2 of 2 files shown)


0sidharth:
Bug fixes should have release notes entries

oscargus:
LGTM! Thanks!

PR #22171, opened 2021-09-28
Fixed return values and some Integer conversions

oscargus:
Replaced some `S` usage with `Integer`. In general it should improve the performance. Also fixed some odd case of function not always returning sympified value.

#### Other comments
It is not always obvious when explicitly converting to Integer is better than just leave it and let the functions handle it...

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/codegen/cfunctions.py
+++ b/sympy/codegen/cfunctions.py
@@ -7,7 +7,7 @@ as a SymPy function for symbolic manipulation.
 
 """
 from sympy.core.function import ArgumentIndexError, Function
-from sympy.core.numbers import Rational
+from sympy.core.numbers import Rational, Integer
 from sympy.core.power import Pow
 from sympy.core.singleton import S
 from sympy.functions.elementary.exponential import exp, log
@@ -160,7 +160,7 @@ class log1p(Function):
     def _eval_is_nonnegative(self):
         return self.args[0].is_nonnegative
 
-_Two = S(2)
+_Two = Integer(2)
 
 def _exp2(x):
     return Pow(_Two, x)
@@ -326,7 +326,7 @@ class fma(Function):
         return _fma(arg)
 
 
-_Ten = S(10)
+_Ten = Integer(10)
 
 
 def _log10(x):
--- a/sympy/combinatorics/schur_number.py
+++ b/sympy/combinatorics/schur_number.py
@@ -42,7 +42,7 @@ class SchurNumber(Function):
             if k is S.Infinity:
                 return S.Infinity
             if k.is_zero:
-                return 0
+                return S.Zero
             if not k.is_integer or k.is_negative:
                 raise ValueError("k should be a positive integer")
             first_known_schur_numbers = {1: 1, 2: 4, 3: 13, 4: 44}
@@ -51,7 +51,7 @@ class SchurNumber(Function):
 
     def lower_bound(self):
         f_ = self.args[0]
-        return (3**f_ - 1)/2
+        return (Integer(3)**f_ - S.One)/Integer(2)
 
 
 def _schur_subsets_number(n):
--- a/sympy/functions/combinatorial/numbers.py
+++ b/sympy/functions/combinatorial/numbers.py
@@ -890,10 +890,10 @@ class harmonic(Function):
                 p = p - u * q
                 if u.is_nonnegative and p.is_positive and q.is_positive and p < q:
                     k = Dummy("k")
-                    t1 = q * Sum(1 / (q * k + p), (k, 0, u))
-                    t2 = 2 * Sum(cos((2 * pi * p * k) / S(q)) *
-                                   log(sin((pi * k) / S(q))),
-                                   (k, 1, floor((q - 1) / S(2))))
+                    t1 = q * Sum(S.One / (q * k + p), (k, S.Zero, u))
+                    t2 = 2 * Sum(cos((2 * pi * p * k) / q) *
+                                   log(sin((pi * k) / q)),
+                                   (k, S.One, floor((q - S.One) / Integer(2))))
                     t3 = (pi / 2) * cot((pi * p) / q) + log(2 * q)
                     return t1 + t2 - t3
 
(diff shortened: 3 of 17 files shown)


sympy/combinatorics/schur_number.py
```diff
@@ -51,7 +51,7 @@ def eval(cls, k):
 
     def lower_bound(self):
         f_ = self.args[0]
-        return (3**f_ - 1)/2
+        return (Integer(3)**f_ - S.One)/Integer(2)
```
asmeurer, line 54: This shouldn't be needed. If `f_` is an Expr (and it should be if its in an object's `args`) the numbers combined with it will automatically be converted to SymPy types.

sympy/codegen/cfunctions.py
```diff
@@ -160,7 +160,7 @@ def _eval_is_zero(self):
     def _eval_is_nonnegative(self):
         return self.args[0].is_nonnegative
 
-_Two = S(2)
+_Two = Integer(2)
```
asmeurer, line 163: Why do we even need `_Two` at all in this file. Everywhere where it is used can be replaced with 2 as far as I can tell, because every function will automatically sympify its input. In fact the only benefit of doing this is if we instead wrote things like `2**x` instead of `Pow(_Two, x)`. Doing both is redundant. Alternatively, functions like `_exp2` could sympify their arguments if they could receive non-SymPy inputs.

sympy/functions/combinatorial/numbers.py
```diff
@@ -890,10 +890,10 @@ def _eval_expand_func(self, **hints):
                 p = p - u * q
                 if u.is_nonnegative and p.is_positive and q.is_positive and p < q:
                     k = Dummy("k")
-                    t1 = q * Sum(1 / (q * k + p), (k, 0, u))
-                    t2 = 2 * Sum(cos((2 * pi * p * k) / S(q)) *
-                                   log(sin((pi * k) / S(q))),
-                                   (k, 1, floor((q - 1) / S(2))))
+                    t1 = q * Sum(S.One / (q * k + p), (k, S.Zero, u))
```
asmeurer, line 893: None of these should be necessary either. q, k, and p are already SymPy types. Even if they weren't, a better fix would be to make sure they are sympified wherever they are defined, rather than trying to chase every place where they are used and avoiding integer/integer divisions.

sympy/functions/special/error_functions.py
```diff
@@ -1694,7 +1694,7 @@ class Li(Function):
     def eval(cls, z):
         if z is S.Infinity:
             return S.Infinity
-        elif z == S(2):
+        elif z is Integer(2):
```
asmeurer, line 1697: Don't use `is` to compare things unless they are `S.something` (S(2) is not singletonized in SymPy). Also this can just be `z == 2`. The type of the rhs doesn't matter for `==`. `S(2) == 2`, `S(2) == 2`, and `2 == 2` all give True.

asmeurer:
Most of these are not needed at all. The only place we really have to be careful about making sure we produce a SymPy type is in the lines that directly return something (like `return S(2)` instead of `return 2`). Anywhere where you have a number in an expression it will automatically be converted to a SymPy type by the operator. The exception is direct integer/integer, which is generally easy to spot because it almost always involves integer literals rather than variables. 

I'm wondering if it would be better to avoid this issue more generally in SymPy with some decorators or metaclass trickery. But that's a more difficult question because while that might prevent mistakes, it could have a significant impact on performance if not done correctly.

asmeurer:
In general though I like this change. I'm actually not a huge fan of `S` in general, although I don't mind it in library code. But Integer is indeed faster (even though both take on the order of microseconds so it usually wouldn't matter).

sympy/codegen/cfunctions.py
oscargus: I think the idea is to save additional sympifying/object creation, but not sure. Will add to the import time, but the be "free". In this context (codegen), it may not be worth it though.

sympy/combinatorics/schur_number.py
oscargus: It is not needed, but is it faster than relying on the automatic conversion? S.One is probably faster, for the other numbers not sure. (This is one of the cases I was talking about as not being obvious in terms of performance.)

sympy/functions/combinatorial/numbers.py
oscargus: One of the change was to remove the S from q. I think that adding S.One and S.Zero will speed things up, but at the cost of readability.

oscargus: In general, I think it would be nice to be able to bypass sympifying when one know that the object are already sympified (as in functions like this).

sympy/functions/special/error_functions.py
oscargus: Correct, but `z == 2`  will call sympify one way or the other, so providing that is better.

No really sure why it works, but even `Integer(1) + Integer(1) is S(2)` returns True. And it seems faster than ==.

oscargus:
There is clearly a trade-off between leaving the sympifying to the automatic conversion or doing it yourself in terms of performance and readability. It is not always obvious which one is the best even for performance (or at least not how to test it, since caching may happen for larger expressions).

Integer should always be better than S, there we agree it seems.

I think that in library code it may make sense to change it, although tedious in the long run. For the tests, I have not bothered to change anything, the primary reason would be to bypass the limitation of 1000 search hits in Spyder, although it will of course have a small performance benefit even there, especially if one could reuse the object creation (some tests do S(2) tens of times).

I'll edit this a bit.

smichr:
Unless there is compelling evidence of significant benefit, I would recommend to make things more legible and shorter.

sympy/functions/special/error_functions.py
asmeurer: > Correct, but z == 2 will call sympify one way or the other, so providing that is better.

Not sure what you are saying here, but I would say `z == 2` is better because it is more readable. The `Integer` is unnecessary.

> No really sure why it works, but even Integer(1) + Integer(1) is S(2) returns True. And it seems faster than ==.

It may work due to the cache, but you shouldn't rely on it. is comparisons are only guaranteed to work for singletonized objects. And in those cases it would be good to be explicit by using `S.whatever`. 

On a related note, `2 is 2` will work in CPython but won't work in other implementations like PyPy. Basically, just because `is` happens to work in an example doesn't mean it will always work, and it can lead to weird and hard to reproduce bugs when it doesn't. It's unfortunate. I really wish Python didn't have the `is` operator, because all it does is trip people up who don't realize that it isn't actually equality testing.

Issue #22073, opened 2021-09-11
Interval with oo


JYLFamily:
Hi:
when use Interval with oo, can not get Lopen or Ropen interval.
```
Interval.Lopen(-oo, 5)
Out[1]: Interval(-oo, 5)
Interval.Ropen(5, +oo)
Out[2]: Interval(5, oo)
```
how to solve this？
Thanks

smichr:
It is impossible to include oo in an interval so the Interval is not printed with that redundant information. But can check to see that it is not included:
```
>>> -oo in Interval(-oo, 5)
False
>>> -oo in Interval.Lopen(-oo, 5)
False
>>> -3 in Interval.Lopen(-3, 5)
False
>>> -3 in Interval(-3, 5)
True
```

oscarbenjamin:
If you really wanted to include `oo` you could use a `Union`:
```python
In [2]: Interval(0, oo) | {oo}
Out[2]: [0, ∞) ∪ {∞}
```

jksuom:
I would not be hard to implement intervals optionally including `+/-oo`. That might even simplify some code.

oscarbenjamin:
Okay, maybe Interval should be changed.

asmeurer:
It could also complicate code that assumes Interval(x, y) is a subset of reals. 

It should be noted that it would be a backwards compatibility break to make Interval(-oo, oo) return a closed interval. 

If we don't change this I would suggest making the class constructor methods like Lopen error when oo is included on a closed end. The default constructor would be annoying if Interval(-oo, oo, True, True) was required, so maybe that shouldn't error.

smichr:
> would be annoying if Interval(-oo, oo, True, True) was required, so maybe that shouldn't error.

The user could also write
```python
>>> Interval.open(-oo,oo)
Interval(-oo, oo)
>>> Interval.Lopen(-oo,5)
Interval(-oo, 5)
>>> Interval.Ropen(1,oo)
Interval(1, oo)
```
And you could raise an error whenever an infinite value is used on a closed end. Following your suggestion, @asmeurer the user would now have to write

```python
Interval.open(1, oo) instead of Interval.Lopen(1, oo)
Interval.open(-oo, 5) insteed of Interval.Ropen(-oo, 5)
```
But the ones I listed above would be unchanged from current practice:
```python
Interval(-oo, oo)
Interval(-oo, 5)
Interval(1, oo)
```

asmeurer:
[...]

PR #22172, opened 2021-09-28
corrections to Piecewise.eval when handling relationals with oo

### LINKED ISSUE #22175: Incorrect results with Piecewise expression MatrixSymbol substitutions.

smichr:

fixes #22175 
minimal changes like those in #22162

#### Other comments
- [x] add tests of all oo-involving And from #22162

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/functions/elementary/piecewise.py
+++ b/sympy/functions/elementary/piecewise.py
@@ -174,6 +174,7 @@ class Piecewise(Function):
         for e, c in _args:
             if (not c.is_Atom and not isinstance(c, Relational) and
                     not c.has(im, re)):
+                orig_c = c
                 x = unigen(c)
                 if x is not None:
                     if not isinstance(x, Symbol):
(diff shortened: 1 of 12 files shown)


sympy/integrals/tests/test_meijerint.py
```diff
@@ -624,11 +624,11 @@ def test_messy():
                        fourier_transform)
     assert laplace_transform(Si(x), x, s) == ((-atan(s) + pi/2)/s, 0, True)
 
-    assert laplace_transform(Shi(x), x, s) == (acoth(s)/s, 1, s > 1)
+    assert laplace_transform(Shi(x), x, s) == (acoth(s)/s, 1, (s > -oo) & (s < -1))
```
oscargus, line 627: Are we sure what the result should be here and below? I also made some changes that affected these, but didn't really come to any conclusion what it actually should be.

smichr: `acoth(x)/x` is symmetric above 1 and below -1 so I think this is fine.

sympy/functions/elementary/piecewise.py
```diff
@@ -192,16 +200,26 @@ def eval(cls, *_args):
...
+                                        orig_c.xreplace({x: inf}) !=
+                                        c.xreplace({x: inf})):
+                                    # hack around inability of Interval to
+                                    # contain +/-oo
+                                    # >>> And(x >= -oo, x <= 3).as_set()
+                                    # Interval(-oo, 3)
+                                    # >>> _.as_relational(x)
+                                    # (x <= 3) & (-oo < x)
                                     reps[i] = Relational(
                                         i.lhs, i.rhs, i.rel_op + '=')
```
oscarbenjamin, line 217: Is it possible to take a different approach and just not use `as_set` at all?

Also if we want to be able to have intervals with `oo` then that's fine but wouldn't it be better just to fix `Interval` rather than work around it here?

smichr: > wouldn't it be better just to fix `Interval` rather than work around it here?

see [here](https://github.com/sympy/sympy/issues/22073#issuecomment-925377833)

smichr: But recall that someone suggested just returning a Union when infinities are included. I am trying that.

sympy/integrals/tests/test_meijerint.py
oscargus: So it should really say `abs(s) > 1`? (Not that it should be fixed in the PR.)

PR #22183, opened 2021-09-29
Fix inconsistent number substitution

aeflores:
Currently number substitutions accross number types (Float vs Integer) behave inconsistently.
```
e = sympy.Integer(9)
e.subs(sympy.Float(9.0000),0)
```
will result in `9` (nothing gets substituted) but:
```
e = sympy.Integer(9)
e.subs(-sympy.Float(9.0000),0)
```
will result in `0` (the substitution is applied).

This is because in the first case, the decision to not substitute is made by `_aresame` https://github.com/sympy/sympy/blob/master/sympy/core/basic.py#L1959 which checks for the class of the number, whereas in the second case the decision is made in `Number._eval_subs` which does not.

Changed `Number._eval_subs` to perform a substitution only if the two number are of the same class.

#### Other comments

I guess another option to fix the inconsistency would be to use `_aresame` inside `Number._eval_subs` though `_aresame` is supposed to be private:
```
if _aresame(old, -self):
```
This way the two behaviors won't diverge again.

#### Release Notes

* core
    * Fix bug in substitution of different classes of numbers

### DIFF
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -696,7 +696,7 @@ class Number(AtomicExpr):
         return Order(S.One, *symbols)
 
     def _eval_subs(self, old, new):
-        if old == -self:
+        if old == -self and old.__class__ == self.__class__:
             return -new
         return self  # there is no other possibility
 


aeflores:
I see some tests are failing, so maybe this is not the right solution. Should `_aresame` be behaving differently instead?
I just noticed that `replace` does not behave like that. ie:
```
e = sympy.Integer(9)
e.replace(sympy.Float(9.0000),0)`
```
yields `0`

oscarbenjamin:
See also #20033

asmeurer:
This gets pretty deep into some fundamental questions about how SymPy treats integers and equivalent floating point numbers. See the discussion in the issues linked to in @oscarbenjamin's PR, particularly https://github.com/sympy/sympy/issues/11707. 

There are basically two options for how SymPy could be treating the objects Integer(9) and Float(9.0): either always treat them as always equal, or always treat them as unequal. Right now SymPy is inconsistent in that it treats them as equal in some places and unequal in others, and there are a lot of places in the code that use special-cases to work around one or the other (like _aresame). I'd rather not add more special cases like the one here, unless there is a very good reason for it. 

The change here also would affect other classes than just integer and float. By the looks of it, it would change the behavior of subs on subclasses. IMO, regardless of which direction we go with the integer/float thing, the change here is not good. If two objects say they are equal (with ==), then subs should treat them as equal (whether or not that would apply to `Integer(9)` and `Float(9.0)` is the point that's up for question). 

You could argue that subs is a bit special in that it does try to look at mathematical equivalence in some cases, so it could potentially apply a rule break in the case of Rational/Float, but if it does, it should do so in a way that only applies to those specific classes (e.g., using _aresame). Even then, if we decide to go with the "unequal" route (as per Oscar's PR), my take is that subs ought to respect that for consistency.

aeflores:
@oscarbenjamin  and @asmeurer thanks for your feedback! I had not fully grasped all the implications of the Float vs Rational (and Integer) classes when I created this PR. I agree with @asmeurer that this is not the right change to do, and since the discussion is ongoing in https://github.com/sympy/sympy/pull/20033, I think I can safely close this PR. It might be good to keep my example above as a test case for when changes are made one way or the other.

PR #22187, opened 2021-09-30
unigen; as_set and non-Symbols; as_relational and +/-oo

### LINKED ISSUE #22073: Interval with oo

### LINKED ISSUE #22175: Incorrect results with Piecewise expression MatrixSymbol substitutions.

smichr:

fixes #22175 by introducing `unigen` which looks for the smallest symbol-like univariate generator for an expression.

fixes #22073 by returning a union to show inclusion of infinity, Relationals are unbounded and hold in extended real whereas `Interval(-oo,oo) == Reals`
```python
>>> (x>9).as_set()
Union({oo}, Interval.open(9, oo))
>>> _.as_relational(x)
x > 9
```
To maintain economy when dealing with such sets being conveted to Relationals, special tests are done to return something similar to the starting Relational. So instead of `Eq(x, oo) | (x > 9) & (x < oo)` for the above, `x > 9` is returned.

With this in place, the corner-case introduction of `=` to relationships expressed without equality in Piecewise conditions has been fixed.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/sets/sets.py
+++ b/sympy/sets/sets.py
@@ -1192,6 +1192,9 @@ class Union(Set, LatticeOp):
         # flatten inputs to merge intersections and iterables
         args = _sympify(args)
 
+        # replace Interval(-oo, oo) with S.Reals
+        args = [S.Reals if i == S.Reals else i for i in args]
+
         # Reduce sets using known rules
         if evaluate:
             args = list(cls._new_args_filter(args))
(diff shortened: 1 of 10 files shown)


sympy/sets/sets.py
```diff
@@ -1192,6 +1192,9 @@ def __new__(cls, *args, **kwargs):
         # flatten inputs to merge intersections and iterables
         args = _sympify(args)
 
+        # replace Interval(-oo, oo) with S.Reals
```
smichr, line 1195: Perhaps better to just do this in the `Union.as_relational` routine?

PR #22188, opened 2021-09-30
allow as_set to represent relationals more accurately to not exclude +/-oo

smichr:
#### Release Notes

NO ENTRY

### DIFF
--- a/doc/src/tutorial/calculus.rst
+++ b/doc/src/tutorial/calculus.rst
@@ -211,7 +211,7 @@ definite integrals.  Here is a sampling of some of the power of ``integrate``.
     ⌡
     0
     >>> integ.doit()
-    ⎧ Γ(y + 1)    for re(y) > -1
+    ⎧ Γ(y + 1)    for y > -1
     ⎪
     ⎪∞
     ⎪⌠
@@ -222,7 +222,7 @@ definite integrals.  Here is a sampling of some of the power of ``integrate``.
     ⎩
 
 This last example returned a ``Piecewise`` expression because the integral
-does not converge unless `\Re(y) > 1.`
+does not converge unless `y > -1` (and real).
 
 Limits
 ======
--- a/sympy/functions/elementary/piecewise.py
+++ b/sympy/functions/elementary/piecewise.py
@@ -956,8 +956,7 @@ class Piecewise(Function):
                 free = c.free_symbols
                 x = free.pop()
                 try:
-                    byfree[x] = byfree.setdefault(
-                        x, S.EmptySet).union(c.as_set())
+                    cs = c.as_set()
                 except NotImplementedError:
                     if not default:
                         raise NotImplementedError(filldedent('''
@@ -968,6 +967,13 @@ class Piecewise(Function):
                             This error would not occur if a default expression
                             like `(foo, True)` were given.
                             ''' % c))
+                else:
+                    # remove extraneous infinities
+                    cs -= {S.Infinity, S.NegativeInfinity}
+                    # update x
+                    byfree[x] = byfree.setdefault(
+                        x, S.EmptySet).union(cs)
+
                 if byfree[x] in (S.UniversalSet, S.Reals):
                     # collapse the ith condition to True and break
                     args[i] = list(args[i])
--- a/sympy/functions/elementary/piecewise.py
+++ b/sympy/functions/elementary/piecewise.py
@@ -171,7 +171,7 @@ class Piecewise(Function):
         args = []
         for e, c in _args:
             if (not c.is_Atom and not isinstance(c, Relational) and
-                not c.has(im, re)):
+                    not c.has(im, re)):
                 free = c.free_symbols
                 if len(free) == 1:
                     funcs = [i for i in c.atoms(Function)
@@ -187,21 +187,6 @@ class Piecewise(Function):
                         c = c.as_set().as_relational(x)
                     except NotImplementedError:
                         pass
-                    else:
-                        reps = {}
-                        for i in c.atoms(Relational):
-                            ic = i.canonical
-                            if ic.rhs in (S.Infinity, S.NegativeInfinity):
-                                if not _c.has(ic.rhs):
-                                    # don't accept introduction of
-                                    # new Relationals with +/-oo
-                                    reps[i] = S.true
-                                elif ('=' not in ic.rel_op and
-                                        c.xreplace({x: i.rhs}) !=
-                                        _c.xreplace({x: i.rhs})):
-                                    reps[i] = Relational(
-                                        i.lhs, i.rhs, i.rel_op + '=')
-                        c = c.xreplace(reps)
             args.append((e, _canonical(c)))
 
         for expr, cond in args:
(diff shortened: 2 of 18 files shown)


sympy/integrals/integrals.py
```diff
@@ -8,7 +8,7 @@
 from sympy.core.logic import fuzzy_bool
 from sympy.core.mul import Mul
 from sympy.core.numbers import oo, pi
-from sympy.core.relational import Ne
+from sympy.core.relational import Ne, Relational
```
smichr, line 11:
```suggestion
from sympy.core.relational import Ne
```

sympy/logic/boolalg.py
```diff
@@ -144,36 +143,36 @@ def as_set(self):
...
         >>> Or(x < -2, 2 < x).as_set()
-        Union(Interval.open(-oo, -2), Interval.open(2, oo))
+        Union({-oo, oo}, Interval.open(-oo, -2), Interval.open(2, oo))
 
         """
         from sympy.calculus.util import periodicity
+        from sympy.core.numbers import oo
         from sympy.core.relational import Relational
+        from sympy.sets.sets import Union
+        from sympy.core.symbol import Symbol, Dummy
```
smichr, line 157:
```suggestion
```

sympy/logic/boolalg.py
```diff
@@ -144,36 +143,36 @@ def as_set(self):
...
+                    s = r._eval_as_set()
+                    if s in (S.EmptySet, S.UniversalSet, S.Reals
+                            ) or s == Union({-oo, oo}, S.Reals):
+                        reps[r] = s.as_relational(x)
+                        continue
+                    raise NotImplementedError(filldedent('''
+                        as_set is not implemented for relationals
+                        with periodic solutions
+                        '''))
+            return self.subs(reps)._eval_as_set()
```
smichr, line 173:
```suggestion
            if x.kind is NumberKind:
                reps = {}
                for r in self.atoms(Relational):
                    if periodicity(r, x) not in (0, None):
                        s = r._eval_as_set()
                        if s in (S.EmptySet, S.UniversalSet, S.Reals):
                            reps[r] = s.as_relational(x)
                            continue
                        raise NotImplementedError(filldedent('''
                            as_set is not implemented for relationals
                            with periodic solutions
                            '''))
                return self.subs(reps)._eval_as_set()
            return self._eval_as_set()
```

sympy/logic/boolalg.py
```diff
@@ -161,7 +162,8 @@ def as_set(self):
                 for r in self.atoms(Relational):
                     if periodicity(r, x) not in (0, None):
                         s = r._eval_as_set()
-                        if s in (S.EmptySet, S.UniversalSet, S.Reals):
+                        if s in (S.EmptySet, S.UniversalSet, S.Reals
+                                ) or s == Union({-oo, oo}, S.Reals)):
```
smichr, line 166:
```suggestion
                                ) or s == Union({-oo, oo}, S.Reals):
```

sympy/logic/boolalg.py
```diff
@@ -10,7 +10,6 @@
 from sympy.core.compatibility import ordered, as_int
 from sympy.core.decorators import sympify_method_args, sympify_return
 from sympy.core.function import Application, Derivative
-from sympy.core.kind import NumberKind
 from sympy.core.numbers import Number
```
smichr, line 13:
```suggestion
from sympy.core.kind import NumberKind
from sympy.core.numbers import Number
```

smichr:
Although Interval is supposed to be a subset or Reals, would it make more sense to introduce `Extended` or `ExtendedInterval` that could contain the infinite bound? `Extended` could replace `Interval` and then the new `Interval` would subclass from `Extended` and have the constraint that infinities would not be contained. I would want `isinstance(foo, Interval)` to return True for either class.

sympy/functions/elementary/piecewise.py
```diff
@@ -968,11 +951,18 @@ def _eval_rewrite_as_ITE(self, *args, **kwargs):
...
                             ''' % c))
-                if byfree[x] in (S.UniversalSet, S.Reals):
-                    # collapse the ith condition to True and break
-                    args[i] = list(args[i])
-                    c = args[i][1] = True
-                    break
+                else:
+                    # remove extraneous infinities
+                    cs -= {S.Infinity, S.NegativeInfinity}
+                    # update x
```
smichr, line 957: how do we know that len(free) was 1 and x is unique?

doc/src/tutorial/calculus.rst
```diff
@@ -211,7 +211,7 @@ definite integrals.  Here is a sampling of some of the power of ``integrate``.
     ⌡
     0
     >>> integ.doit()
-    ⎧ Γ(y + 1)    for re(y) > -1
+    ⎧ Γ(y + 1)    for y > -1
```
smichr, line 214: the integration variable changed, too, from `x` to `y`!

sympy/integrals/integrals.py
```diff
@@ -1523,7 +1539,7 @@ def integrate(*args, meijerg=None, conds='piecewise', risch=None, heurisch=None,
     in interactive sessions and should be avoided in library code.
 
     >>> integrate(x**a*exp(-x), (x, 0, oo)) # same as conds='piecewise'
-    Piecewise((gamma(a + 1), re(a) > -1),
+    Piecewise((gamma(a + 1), a > -1),
         (Integral(x**a*exp(-x), (x, 0, oo)), True))
```
smichr, line 1543: the second integral changed from x to a

sympy/integrals/integrals.py
```diff
@@ -559,10 +561,24 @@ def try_meijerg(function, xab):
...
+                                # not having them in Piecewise doesn't allow
+                                # x < oo to evaluate, so we handle it here
+                                # before creating the Piecewise
+                                free = cond.free_symbols
+                                d = None
+                                if len(free) == 1:
+                                    x = free.pop()
+                                    if not x.is_real:
+                                        d = Dummy(real=True)
+                                        cond = cond.xreplace({x: d})
```
smichr, line 575:
```suggestion
                                if len(free) == 1:
                                    cx = free.pop()
                                    if not cx.is_real:
                                        d = Dummy(real=True)
                                        cond = cond.xreplace({cx: d})
```

sympy/integrals/integrals.py
```diff
@@ -559,10 +561,24 @@ def try_meijerg(function, xab):
...
                                 ret = Piecewise(
                                     (f, cond),
                                     (self.func(
                                     function, (x, a, b)), True))
+                                if d:
+                                    ret = ret.xreplace({d: x})
```
smichr, line 581:
```suggestion
                                    ret = ret.xreplace({d: cx})
```

oscarbenjamin:
> Although Interval is supposed to be a subset or Reals, would it make more sense to introduce `Extended` or `ExtendedInterval` that could contain the infinite bound? `Extended` could replace `Interval` and then the new `Interval` would subclass from `Extended` and have the constraint that infinities would not be contained. I would want `isinstance(foo, Interval)` to return True for either class.

That sounds reasonable although maybe `Interval` should just be allowed to include infinities. The code seems to be somewhat complicated by the seemingly arbitrary requirement not to have the infinities included. It makes sense to exclude infinities from Reals but I don't see why Interval should not just be able to include infinity.

smichr:
> I don't see why Interval should not just be able to include infinity.

@asmeurer had said that this would be a backwards incompatible change since Interval is supposed to be a subset of Reals.

PR #22192, opened 2021-09-30
enhanced Interval creation that can include oo

smichr:
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/sets/sets.py
+++ b/sympy/sets/sets.py
@@ -1039,27 +1039,12 @@ class Interval(Set):
 
     @classmethod
     def xx(cls, a, b):
-        """Return an closed interval."""
+        """Return a closed interval."""
         return cls(a, b, False, False, extended=True)
 
     @classmethod
     def xo(cls, a, b):
-        """Return an interval open on the rigth."""
-        return cls(a, b, False, True, extended=True)
-
-    @classmethod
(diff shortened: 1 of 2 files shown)


sympy/sets/sets.py
```diff
@@ -1035,6 +1037,56 @@ def Ropen(cls, a, b):
...
+    @classmethod
+    def xo(cls, a, b):
+        """Return an interval open on the rigth."""
+        return cls(a, b, False, True, extended=True)
+
+    @classmethod
+    def ox(cls, a, b):
+        """Return an interval open on the left."""
+        return cls(a, b, True, False, extended=True)
+
```
smichr, line 1069:
```suggestion
    @classmethod
    def xx(cls, a, b):
        """Return a closed interval."""
        return cls(a, b, False, False, extended=True)

    @classmethod
    def xo(cls, a, b):
        """Return an interval open on the right."""
        return cls(a, b, False, True, extended=True)

    @classmethod
    def ox(cls, a, b):
        """Return an interval open on the left."""
        return cls(a, b, True, False, extended=True)

```

smichr:
@asmeurer comment [here](https://groups.google.com/g/sympy/c/9KrMk2z1FAI) suggests that what is presented here is not the way to go.

PR #22193, opened 2021-10-01
Fixed the Sum.doit() cases where eventually after splitting the componenets we would end up in an oo-oo form

### LINKED ISSUE #14410: Sum(3-1/sqrt(x), (x,0,oo)).doit() is still sum
`Sum(3-1/sqrt(x), (x,0,oo)).doit()` gives sum instead of `oo`

### LINKED ISSUE #14579: Can't solve limit in sqrt
```
>>> from sympy import *
>>> x=symbols('x')
>>> Sum(sqrt(x), (x,1,oo)).doit()
Sum(sqrt(x), (x, 1, oo))
```
where the answer should have been ```oo```(related to https://github.com/sympy/sympy/issues/14393)

### LINKED ISSUE #22178: Improvements in Sum.doit() for functions having the +/- operator .
While trying to approach another issue(#14410) , I realized that Sum.doit() doesn't work as expected and certainly needs improvements to solve a couple of issues I browsed through .

1) `Sum.doit()` for functions having the  +/-  operator doesn't always give expected outputs, for eg
```
>>> Sum((3+x) , (x,1,oo)).doit()
oo
>>> Sum((3-x) , (x,1,oo)).doit()
Sum(3 - x, (x, 1, oo))
>>> Sum((3-x) , (x,4,oo)).doit()
-oo
```
At first look this indicates  that maybe `doit`  is checking for the first term whether +ve/-ve and then returns `oo/-oo` accordingly .But the next example shows that might not be the case 

```
>>> Sum( x**2 - x**3, (x,1,oo)).doit()
-oo
>>> Sum( x**2 - x**3, (x,0,oo)).doit()
Sum(-x**3 + x**2, (x, 0, oo))
>>> Sum( x**2 - x**3, (x,-1,oo)).doit()
Sum(-x**3 + x**2, (x, -1, oo))
```

This example shows that if there's a possibility to get both +ve and -ve terms from the Sum ,then `Sum.doit()` doesn't perform as expected and the first intuition is probably not correct .

2) An alternating series passed inside Sum either with or without  the +/- operators  eg `Sum(((-1)**x *(x**2 + 1)/x) , (x,1,oo))`  or  `Sum((3 + (-1)**x *(x**2 + 1)/x) , (x,1,oo))`  which would eventually diverge should be returning `None `cause this would oscillating between +oo and -oo eventually . As of now this is yet to be implemented
```
>>> Sum(((-1)**x *(x**2 + 1)/x) , (x,1,oo)).doit()
Sum((-1)**x*(x**2 + 1)/x, (x, 1, oo))
>>> Sum((3 + (-1)**x *(x**2 + 1)/x) , (x,1,oo)).doit()
Sum((-1)**x*(x**2 + 1)/x + 3, (x, 1, oo))
```

anutosh491:

Fixes #22178
Closes #14579 
Closes #14410

The problem was that Sum.doit() deals with such summations having +/- operator by splitting them into parts and getting result seperately which is not wrong at all but in cases like `Sum(3 - sqrt(x) , (x, 1, oo)).doit()` , the 3 part would return oo and something like `-sqrt(x)` would return `-oo` ,hence oo - oo gives nan rather than considering which side would the summation end up preferring (obviously sqrt(x) would dominate 3 as x goes to infinity) hence it would return `Sum(3 - sqrt(x) , (x, 1, oo)).doit()` instead of `-oo`
Some examples to demonstrate the fix
```
>>> from sympy import Sum, sqrt, Symbol, oo
>>> x = Symbol('x')
>>> Sum(sqrt(x) , (x, 1, oo)).doit()
oo
>>> Sum(3 - x**2 , (x, 1, oo)).doit()
-oo
>>> Sum(3 - x , (x, 1, oo)).doit()
-oo
>>> Sum(3 - sqrt(x) , (x, 1, oo)).doit()
-oo
>>> Sum(sqrt(x) - 3, (x, 1, oo)).doit()
oo
>>> Sum(3 - 1/x, (x, 1, oo)).doit()
oo
>>> Sum(x**2 - x**3, (x, 1, oo)).doit()
-oo
>>> Sum(x**3 - x**2, (x, 1, oo)).doit()
oo
>>> Sum(x**3 + x**2, (x, 1, oo)).doit()
oo
>>> Sum(1/(sqrt(x)) - 3, (x, 1, oo)).doit()
-oo
```
All these cases weren't returning the expected result as of now .

#### Other comments
I will be adding test cases after the initial review .Any suggestions are welcome

#### Release Notes

* series
  * Fixed the Sum.doit() cases where eventually after splitting the components we would end up in an oo-oo form.


sympy/concrete/summations.py
```diff
@@ -1137,9 +1137,16 @@ def eval_sum_symbolic(f, limits):
         if without_i != 0:
             s = eval_sum_symbolic(with_i, (i, a, b))
             if s:
-                r = without_i*(b - a + 1) + s
-                if r is not S.NaN:
-                    return r
+                if((s == S.NegativeInfinity and (b - a + 1) == S.Infinity) or (s == S.Infinity and (b - a + 1) == S.Infinity)):
```
smichr, line 1140: did you mean the last S.Infinity to be S.NegativeInfinity?  If so, you could write `if (b - a - s).is_infinite:`

sympy/concrete/summations.py
```diff
@@ -1137,9 +1137,16 @@ def eval_sum_symbolic(f, limits):
...
+                    Max = max(abs(f.subs(i, a)), abs(f.subs(i, S.Infinity)))
```
oscargus, line 1141: Slightly better to use the Sympy `Max` and `Abs` directly (and therefore rename the variable).

sympy/concrete/summations.py
```diff
@@ -1137,9 +1137,16 @@ def eval_sum_symbolic(f, limits):
...
+                    if(Max > abs(without_i)):
```
oscargus, line 1142: No brackets around the comparison. You may also want to consider that this may not evaluate and then you will get an error. Better to use `is_gt` and then check the three cases: `True`, `False` and `None`. (Or maybe you have some indication from previous code flow that it will evaluate?)

sympy/concrete/summations.py
oscargus: An option (in case the above is not what you want to to) is to do the comparison like `s is S.Infinity` or `s in (S.Infinity, S. NegativeInfinity)` which is slightly more efficient.

sympy/concrete/summations.py
```diff
@@ -1158,7 +1165,7 @@ def eval_sum_symbolic(f, limits):
     if result is not None:
         n = result[n]
 
-        if n.is_Integer:
+        if n.is_Integer or n.is_real:
```
oscargus, line 1168: Which problem will this solve? It seems like the requirements are quite different. `is_Integer` will only be true if `n` is an actual integer, i.e., a number. `is_real` will be True for quite a lot of expressions, including symbols that are real, but not necessarily integer.

I'm not saying that it is a bad idea, just a bit surprising. `is_integer` would be less surprising.

oscargus: If n is a symbol or an expression, the next comparison may throw an error and so on. So it seems like maybe this is not needed (as in not tested for the case where n is an unevaluated expression).

oscargus: Reading up a bit more. It seems like n will match the exponent, so I guess this is for sqrt? Then, one should consider if the expressions below are valid for Rational numbers (bernoulli is at least not) and probably test with `n.is_Rational` instead of `n.is_real`. Even though `harmonic(oo, S.Half)` evaluates to `oo`, I am not sure it will be correct for other cases. So probably better to have a test after this to deal with the cases that can be solved explicitly.

You then should add some tests that uses this for sqrt-summations that doesn't evaluate to `oo`.

sympy/concrete/summations.py
anutosh491: yupp cool will get this done . the Second check  by the way is for these cases where s would be returning `oo` but still won't  be dominating the other term as eventually returns `-oo` .These cases won't be addressed without the second one 
```
>>> Sum(1/x - 3, (x, 1, oo)).doit()
-oo
```

sympy/concrete/summations.py
anutosh491: Yeah  exactly what I wanted to fix , `Sum( sqrt(x), (x, 1, oo)).doit()` or rather any non-integer exponent like `Sum( x**(2.5), (x, 1, oo)).doit()` wasn't returning `oo` and the check is for that . Yess I'll be adding all tests for all these , was waiting for initial reviews by you'll !

sympy/concrete/summations.py
anutosh491: ahh yeah the general case where the limits are of the form `(x , int, oo)` is handled properly by `eval_sum_hyper() `method ,no need to worry about that . 
```
    # Now b == oo, a != -oo
    res = _eval_sum_hyper(f, i, a)
    if res is not None:
        r, c = res
        if c == False:
            if r.is_number:
                f = f.subs(i, Dummy('i', integer=True, positive=True) + a)
                if f.is_positive or f.is_zero:
                    return S.Infinity
                elif f.is_negative:
                    return S.NegativeInfinity
            return None
        return Piecewise(res, (old_sum, True))
```
To be honest this actually boils down to a very very specific case that i've tried to fix . Sympy wasn't able to deal with any sort of numbers  or non integer exponents contained in Sum but as obviously as integers and real are infinite in number even this specific case becomes very important to fix .I've shown above what all would be fixed as an end result ,and all of these weren't responding as expected .

sympy/concrete/summations.py
anutosh491: I used `max` and `abs` to maintain consistency in the code , all throughout summations.py these ones have been used everywhere . Even the is_absolutely_convergent() method which just a a single line uses `abs` .

sympy/concrete/summations.py
anutosh491: > Reading up a bit more. It seems like n will match the exponent, so I guess this is for sqrt? Then, one should consider if the expressions below are valid for Rational numbers (bernoulli is at least not) and probably test with `n.is_Rational` instead of `n.is_real`. Even though `harmonic(oo, S.Half)` evaluates to `oo`, I am not sure it will be correct for other cases. So probably better to have a test after this to deal with the cases that can be solved explicitly.
> 
> You then should add some tests that uses this for sqrt-summations that doesn't evaluate to `oo`.
Well we won't get anything of this sort with .is_Rational  probably .is_real would be better !
```
>>> Sum(x**2.66, (x, 1, oo)).doit()
Sum(x**2.66, (x, 1, oo))
>>> Sum(1/(x**0.99), (x, 1, oo)).doit()
Sum(x**(-0.99), (x, 1, oo)) 
```

sympy/concrete/summations.py
oscargus: OK! It will work, in the sense that Sympy will turn them into `Abs` and `Max` with some additional processing, but probably they should be changed elsewhere as well (although it must be done a bit carefully, sometimes they do compute the `max` and `abs` of standard Python types, like the result of `len`). 

I will probably change it at some stage, so feel free to do that already now...

sympy/concrete/summations.py
oscargus: Floats are tricky in general. I suggest add `Sum(x**(S(8)/3), (x, 1, oo)).doit()` and, say, `Sum(x**S(9)/10, (x, 1, oo)).doit()` instead.

And for sqrt-test that doesn't sum up to `oo`, say, `Sum(1/sqrt(x), (x, 1, 1000)).doit()` which currently is a 1000 term series, but I assume could be something else. (If it is still a 1000 term series, maybe it is a bad idea to add it...)

oscargus: Nah, I may be a bit confused here. Add what you think are good tests here, but it would be interesting to see a case evaluating the bernoulli and the harmonic lines where n is real and the result is not `oo` or `-oo`. Those are the ones I am a bit concerned about.

(Still, floats are a bit non-intuitive and may lead to weird errors, although they may work here.)

sympy/concrete/summations.py
oscargus: At least, please rename the `Max` variable to make it less confusing.

anutosh491: > OK! It will work, in the sense that Sympy will turn them into `Abs` and `Max` with some additional processing, but probably they should be changed elsewhere as well (although it must be done a bit carefully, sometimes they do compute the `max` and `abs` of standard Python types, like the result of `len`).
> 
> I will probably change it at some stage, so feel free to do that already now...

Yeah exactly it's not that I didn't try , my plan earlier was to go with Abs, Max but and I had made relevant changes at a couple of places where I found it necessary for this PR but that was breaking some things from my side hence I decided to switch back .

sympy/concrete/summations.py
anutosh491: yeah even I wasn't completely sure what all tests (specifically those which don't evaluate to `oo`) would be a good fit here ,will give a more thorough thought regarding what to add according to your comments . Yeah truly speaking I didn't consider much when I added `n.is_real` .I'll try some stuff out , if you have something very particular in mind let me know !Thanks

oscargus: Those suggested above would be good. They will trigger the new parts of the code I think (n not being integer).

In addition:
m = Symbol('m', real=True)
Sum(x**m, (x, 1, oo)).doit()
```

Chances are that this will throw an error though, which then indicates that `is_real` is not what you want.
Sum(

Comment (anutosh491) [reply sympy/concrete/summations.py:1168]: hey @oscargus I have been trying a bit now since couple of hours but I'm not sure does the bin test for test_sums_products.py give me an assertion error , locally on my system I get :
```
>>> Sum(x**bernoulli(4), (x, 1, 1000)).doit()
harmonic(1000, 1/30)
>>> Sum(1/sqrt(x), (x, 1, 1000)).doit()
harmonic(1000, 1/2)
>>> Sum(1/sqrt(x) , (x, 1, 10)).doit()
sqrt(10)/10 + sqrt(7)/7 + sqrt(6)/6 + sqrt(5)/5 + sqrt(3)/3 + 3*sqrt(2)/4 + 11/6
```
But no matter how I try to mold these I am getting an assertion error whenever I run the mandatory bin/test concrete as I have been trying to commit . Some examples 
```
    assert Sum(1/sqrt(x), (x, 1, 10)).doit() == sqrt(10)/10 + sqrt(7)/7 + sqrt(6)/6 + sqrt(5)/5 + sqrt(3)/3 + 3*sqrt(2)/4 + 11/6
AssertionError
    assert Sum(1/sqrt(x), (x, 1, 10)).doit() == Sum(1/sqrt(x), (x, 1, 10))
AssertionError
    assert Sum(x**bernoulli(4), (x, 1, 1000)).doit() == harmonic(1000, 1/30)
AssertionError
    assert Sum(x**bernoulli(6), (x, 1, 10)).doit() == 1 + 2**(1/42) + 3**(1/42) + 2**(1/21) + 5**(1/42) + 6**(1/42) + 7**(1/42) + 2**(1/14) + 3**(1/21) + 10**(1/42)
AssertionError
    assert Sum(1/sqrt(h), (h, 1, n)).doit() == harmonic(n, 1/2)
AssertionError
```
No matter how I'm trying to get through i'm not able too . Btw this doesn't trouble me much 
```
assert Sum(x**bernoulli(4), (x, 1, oo)).doit() == oo
assert Sum(x**bernoulli(5), (x, 1, 1000)).doit() == 1000
```
No idea what might be going wrong , will try to figure it out . I thought I was close ...... but may be not .Anyways the cases/above issues which I specifically wanted to fix , deal with `doit()` with upper bound as `oo ` and atleast those work really great as of now !

Comment (oscargus) [reply sympy/concrete/summations.py:1168]: Not really sure what the problem is, but `bernoulli(5) = 0`, so not sure why you test it like that. My comment was about testing a series which returns `bernoulli`, like `Sum(x**(S(9)/10), (x, 1, n)).doit()`

**Comment by anutosh491** [comment]
> It is worth noting that `Sum(1/x**(S(3)/2), (x, 1, oo)).doit()` do work in current master, so it is only rational exponents > -1 that doesn't evaluate. Probably the thing to do is to check if the exponent is > -1 and one of the bounds is oo (in a correct way) and then return `oo`.

```
        if n.is_Integer or n.is_Rational:
            if n >= 0:
                if (b is S.Infinity and not a is S.NegativeInfinity) or \
                   (a is S.NegativeInfinity and not b is S.Infinity):
                    return S.Infinity
                if n.is_Integer:
                    return ((bernoulli(n + 1, b + 1) - bernoulli(n + 1, a))/(n + 1)).expand()
```

Well the first line and the next two lines check exactly what you wanted , they check whether the exponent is `n>-1` and whether `b` would be `oo` hence the latest code I pushed would evaluate all related cases concerning rational exponents firstly because we now use `n.is_Rational` and secondly because these checks would operate exactly as you said . For eg
```
>>> Sum(x**(S(-3/2)) , (x, 1, oo)).doit()
zeta(3/2)
>>> Sum(x**(S(3/2)) , (x, 1, oo)).doit()
oo
>>> Sum(x**(S(-1/2)) , (x, 1, oo)).doit()
oo
```

**Comment by anutosh491** [comment]
> > bigger than -1
> 
> Yes, that is what I meant. -3/2 < -1 and that works now, but will (probably) get a new behavior is this PR. However, in the range 0 to -1 there are no explicit checks as far as I can tell.
> 
> I am also wondering if one should support symbolic exponents here?
> 
> ```
> x = Symbol('x', real=True)
> m = Symbol('m', positive=True)
> n = Symbol('n', integer=True, postive=True)
> Sum(x**m, (x, 1, oo)).doit()
> Sum(x**n, (x, 1, oo)).doit()
> Sum(x**m, (x, 1, n)).doit()
> ```
> 
> Basically, you should be able to replace, e.g., `is_Integer` with `is_integer` and `n >= 0` with `is_ge(n, 0)`. However, `is_ge` can return `True`, `False` and `None`, so the first branch should happen if `True` and the second if `False` (no branch if `None`).

Yupp I have checked the symbolic ones and they work perfectly fine , will go through the suggestion nevertheless . I just replied above to one of your comments , you could review that as of now ! . I'll get back after checking this

**Comment by anutosh491** [comment]
> I still think that you should add a couple of tests (with and without oo limits) for a negative exponent < -1 as this is a new case that your code handles.

I've added some tests for the negative exponent thinking about what all possible code these cases might trigger .Now that we have unlocked `harmonic(a, b)` where b is not integer , it returns a consistent result to what used to get from when b was integer on master. I'll demonstrate what I'm trying to convey 
```

>>> Sum(x**(-3) , (x, 3, oo)).doit() # This doesn't just shift the zeta function giving `zeta(3, 3)  `
-9/8 + zeta(3)                        # but returns a more appropriate result      
 >>> Sum(x**(-S(3)/2) , (x, 3, oo)).doit()
-harmonic(2, 3/2) + zeta(3/2)         # Previously returned this : zeta(3/2, 3) , as it was unable to deal with 3/2 in harmonic , it 
                                      # ended up shifting the zeta function  
```                            
 And I think this is good enough because` harmonic(2, 3/2)` would involve  adding to `sqrt(2)` somewhere and hence in most cases where we have a non integer second term in `harmonic` we wouldn't be giving a presentable result like `9/8` above .I would like to know if you would want be to add any other test cases

**Comment by anutosh491** [comment]
Hello @oscargus thanks for the extensive review on this and I would be glad if we could work towards getting this in . I guess you haven't seen this after you suggested some changes and I made my latest commit . I don't remember exactly what I has changed will go through again I'll point you out to my comments which I made after the latest commits for you're convenience and for easy review 

1) I had made a compilation of all results we might be expecting now that I've have introduced some decent changes in the codebase , it's listed in https://github.com/sympy/sympy/pull/22193#discussion_r730879134

2) I would like you to go through a review thread from a particular reply of mine which shows how the Pr deals better with cases than on master , it's this one https://github.com/sympy/sympy/pull/22193#discussion_r730584092

3) And finally replying to your review .....
"**I still think that you should add a couple of tests (with and without oo limits) for a negative exponent < -1 as this is a new case that your code handles.**"

I had made a comment which is this one https://github.com/sympy/sympy/pull/22193#issuecomment-946324465

I am satisfied from my side that all cases have been addressed thoroughly and maybe this could go in after you review . Let me know what you think . Thanks for the help !

**Comment by anutosh491** [comment]
When I made this Pr , it was just my second one in sympy , hence I wasn't sure about the best possible logic here hence I ended up using a trivial logic to get the preferred term but now that I know `Order` could help me out with that , I realized I should make this change . The previous logic would fail for few cases where we would have more than 2 arguments in the expressions like the following and would return self back as the trivial code written for calculating the dominant term wasn't perfect. Using Order helps us get rid of that !
```
>>> Sum(3 - sqrt(x) + x, (x, 1, oo)).doit()
oo
>>> Sum(1/x -3 + sqrt(x) -x**2, (x, 1, oo)).doit()
-oo
```

anutosh491:
ping @oscargus for review

anutosh491:
ping @oscargus for review !

Issue #12345, opened 2017-03-18
nonlinsolve (solve_biquadratic) gives no solution with radical


smichr:
solve_biquadratic is called in the process of using nonlinsolve but it does not give a solution (or error) when an equation contains a radical:

```
>>> nonlinsolve((x**2-y-sqrt(2), x**2+x*y-y**2), x,y)
EmptySet()
```

Replacing the radical with a rational approximation gives

```
>>> for i in nonlinsolve((x**2-y-1414/S(1000), x**2+x*y-y**2), x,y):
...     Point(i).n(2)
...
Point2D(-1.5, 0.95)
Point2D(2.2, 3.6)
Point2D(0.92, -0.57)
Point2D(-0.63, -1.0)
```

(and those roots were checked visually by graphing)

maurogaravello:
I think the problem is in finding roots of polynomials. Indeed at line 89 of polysys.py the instruction
q_roots = list(roots(q).keys())
gives an empty list.
q in this case is the polynomial of degree 4: 
Poly(y**4 - 3*y**3 + (-3*sqrt(2) + 1)*y**2 + 2*sqrt(2)*y + 2, y, domain='QQ<sqrt(2)>')

smichr:
Yes, and I tried raising an error there so the calling routine would fallback to something else and it didn't work either. I didn't look the other direction to see what has to be fixed in roots.

szymag:
I run debugger during solving your equation and I found interesting things.
In your case, `roots` has problem with finding roots for this Poly: `Poly(y**4 - 3*y**3 + (-3*sqrt(2) + 1)*y**2 + 2*sqrt(2)*y + 2, y, domain='QQ<sqrt(2)>')`

I did some calculation for another polynomials and I found that, if poly has variables to the power greater than 3, the results may be wrong. For example: 
`Poly(y ** 12 - 3 * y ** 3 + (-3 * sqrt(2) + 1) * y ** 2 + 2 * sqrt(2) * y + 2, y, domain='QQ<sqrt(2)>') `
has roots: 
`{{y -> -1.1612 - 0.866924 I}, {y -> -1.1612 + 0.866924 I}, {y -> -0.528763}, {y -> 0.944865}, {y -> 1.9063}}` (calculated in Mathematica). Using `roots` I got `{}`.
What's more for poly without radical I also get wrong result. For example: 
[...]

PR #22194, opened 2021-10-01
Improved sign handling in signsimp (fixes #22189)

### LINKED ISSUE #22189: simplify with Abs
Hi!

```python
>>> a - b
```

How i can get `0`?

Thanks!

oscargus:

Fixes #22189

Based on @asmeurer's insight, `could_extract_minus_sign` is used in `Abs` to make the argument canonical with respect to sign. 

#### Other comments
One may also consider making it canonical with respect to `I` (or `exp(I*x)`), but that is left.

#### Release Notes

* functions
    * `Abs` use canonical argument with respect to sign of argument.

### DIFF
--- a/sympy/functions/elementary/complexes.py
+++ b/sympy/functions/elementary/complexes.py
@@ -528,9 +528,6 @@ class Abs(Function):
             raise TypeError("Bad argument type for Abs(): %s" % type(arg))
 
         # handle what we can
-        if arg.could_extract_minus_sign():
-            return cls(-arg)
-
         arg = signsimp(arg, evaluate=False)
         n, d = arg.as_numer_denom()
         if d.free_symbols and not n.free_symbols:
--- a/sympy/functions/elementary/complexes.py
+++ b/sympy/functions/elementary/complexes.py
@@ -528,6 +528,9 @@ class Abs(Function):
             raise TypeError("Bad argument type for Abs(): %s" % type(arg))
 
         # handle what we can
+        if arg.could_extract_minus_sign():
+            return cls(-arg)
+
         arg = signsimp(arg, evaluate=False)
         n, d = arg.as_numer_denom()
         if d.free_symbols and not n.free_symbols:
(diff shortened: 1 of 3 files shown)


smichr:
I wonder if `signsimp` is used in that function...just a thought. I don't have time to look more closely now.
```python
>>> eq=(2+x)/(y*(-1-x))
>>> eq.could_extract_minus_sign()
False
>>> signsimp(eq)
-(x + 2)/(y*(x + 1))
>>> abs(eq)
Abs((x + 2)/(y*(x + 1)))
```

oscargus:
It is used, but doesn't seem to arrange it canonically.

smichr:
I think the better solution is to fix `signsimp`:
```python
    if e.is_Add:
        return e.func(*[signsimp(a, evaluate) for a in e.args])
```
should maybe be
```python
    if e.is_Add:
        if e.could_extract_minus_sign():
            neg = True
            e = -e
        else:
            neg = False
        rv = e.func(*[signsimp(a, evaluate) for a in e.args])
        if evaluate:
            return -rv if neg else rv
        return Mul(-1, rv, evaluate=False) if neg else rv
```

smichr:
rabbit trail... 
```python
>>> (1-x).could_extract_minus_sign()
True
>>> (x-1).could_extract_minus_sign()
False
>>> ((1-x)*y).could_extract_minus_sign()  # one of these must be True so extract_multiplicatively needs help (I think)
False
>>> ((x-1)*y).could_extract_minus_sign()
False
```

oscargus:
The signsimp change seems to work (at least for complexes and simplify tests).

smichr:
I think the root of the problem is tie breaking for Add and Mul in `could_extract_minus_sign` in expr.py.

smichr:
Is this change necessary if #22201 were fixed?

oscargus:
> Is this change necessary if #22201 were fixed?

Not sure, but I think so since neither `sign_simp` or `Abs.eval` do use `could_extract_minus_sign` otherwise.

oscarbenjamin:
I think it would be better if we didn't have PRs/issues with titles like "Fixed #22189". That could make sense at the end like "Handle mul of adds in could_extract_minus_sign (fixes #12345)". The PR title should convey what is changed directly without needing reference to other issues.

smichr:
#22232 will close this in 24 hours if there is no objection.

PR #22196, opened 2021-10-01
meijerg indefinite_1 bug eliminated

smichr:

The loop over expr, cond overwrote the previously defined cond. A test was added to show that this will fail if it is not changed: an unevaluated integral results. The result that is there is confirmed for this case:

```python
>>> integrate((b + t)**(-a), (t),meijerg=True).subs(b,1).subs(a,2)
-t/(t + 1)**2 - 1/(t + 1)**2
>>> simplify(_)
1/(-t - 1)
>>> signsimp(_)
-1/(t + 1)
>>> integrate(((b + t)**(-a)).subs(b,1).subs(a,2), t)
-1/(t + 1)
```
#### Other comments

An evaluate option was given to `piecewise_fold`.

#### Release Notes

* functions
    * `piecewise_fold` has the option to avoid evaluation
* integrals
    * a bug in indefinite meijerg integration was fixed

### DIFF
--- a/sympy/functions/elementary/piecewise.py
+++ b/sympy/functions/elementary/piecewise.py
@@ -1045,11 +1045,16 @@ class Piecewise(Function):
             return result
 
 
-def piecewise_fold(expr):
+def piecewise_fold(expr, evaluate=True):
     """
     Takes an expression containing a piecewise function and returns the
     expression in piecewise form. In addition, any ITE conditions are
     rewritten in negation normal form and simplified.
+
+    The final Piecewise is evaluated (default) but if the raw form
+    is desired, send ``evaluate=False``; if trivial evaluation is
+    desired, send ``evaluate=None`` and duplicate conditions and
+    processing of True and False will be handled.
 
     Examples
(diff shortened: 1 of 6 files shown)


smichr:
@oscarbenjamin , what do you think about the evaluate option on the folding?

oscarbenjamin:
> @oscarbenjamin , what do you think about the evaluate option on the folding?

I think it's good. I'm not sure I understand what the unpolarify does.

smichr:
> what the unpolarify does.

I didn't write that part; I'm just the editor. :-)

jksuom:
Typical G=functions are branched at the origin. `meijerint` tries to handle different branches internally by using "polar numbers", i.e., complex numbers with an attached argument that belongs to a not necessarily principal branch of `arg`. It is expected that the argument `z` of final result has `arg(z) <= pi` and can therefore be represented by an ordinary complex number. `unpolarify` implements that by dropping the extra argument value. So it is essentially just a cleaning operation.

PR #17215, opened 2019-07-17
KroneckerDelta simplification and rewrite(Piecewise)

ethankward:
#### References to other Issues or PRs
Closes #17214.

#### Brief description of what is fixed or changed
Add `_eval_rewrite_as_Piecewise` to `KroneckerDelta`. Add `simplify` recognition of expressions containing `KroneckerDelta` and attempt to simplify via `Piecewise`.

#### Release Notes

* functions
  * KroneckerDelta can be rewritten as Piecewise.
  * Piecewise can be rewritten as KroneckerDelta
* simplify
  * simplify attempts to simplify expressions containing KroneckerDelta.

PR #22199, opened 2021-10-01
Fixed return values

oscargus:

Closes #22171

Fixed return values that were not sympified.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/simplify/simplify.py
+++ b/sympy/simplify/simplify.py
@@ -1241,7 +1241,7 @@ def kroneckersimp(expr):
             args1 = delta1.args
             args2 = delta2.args
             if args_cancel(args1, args2):
-                return 0*m
+                return S.Zero
         return m
 
     if not expr.has(KroneckerDelta):
--- a/sympy/simplify/simplify.py
+++ b/sympy/simplify/simplify.py
@@ -1241,7 +1241,7 @@ def kroneckersimp(expr):
             args1 = delta1.args
             args2 = delta2.args
             if args_cancel(args1, args2):
-                return S.Zero
+                return S.Zero * m # In case of oo etc
         return m
 
     if not expr.has(KroneckerDelta):
--- a/sympy/simplify/tests/test_simplify.py
+++ b/sympy/simplify/tests/test_simplify.py
@@ -7,7 +7,7 @@ from sympy import (
     Matrix, MatrixSymbol, Mul, nsimplify, oo, pi, Piecewise, Poly, posify, rad,
     Rational, S, separatevars, signsimp, simplify, sign, sin,
     sinc, sinh, solve, sqrt, Sum, Symbol, symbols, sympify, tan,
-    zoo, And, Gt, Ge, Le, Or)
+    zoo, And, Gt, Ge, Le, Or, eye)
 from sympy.core.mul import _keep_coeff
 from sympy.core.expr import unchanged
 from sympy.simplify.simplify import nthroot, inversecombine
@@ -938,6 +938,11 @@ def test_simplify_kroneckerdelta():
                                      [0, K(0, n), 0, K(1, n)],
                                      [0, 0, K(0, n), 0],
                                      [0, 0, 0, K(0, n)]])
+    assert simplify(eye(1) * KroneckerDelta(0, n) *
+                    KroneckerDelta(1, n)) == Matrix([[0]])
+
+    assert simplify(S.Infinity * KroneckerDelta(0, n) *
+                    KroneckerDelta(1, n)) is S.NaN
 
 
 def test_issue_17292():
--- a/sympy/combinatorics/fp_groups.py
+++ b/sympy/combinatorics/fp_groups.py
@@ -687,7 +687,7 @@ class FpSubgroup(DefaultPrinting):
 
     def order(self):
         if not self.generators:
-            return 1
+            return S.One
         if isinstance(self.parent, FreeGroup):
             return S.Infinity
         if self.C is None:
--- a/sympy/combinatorics/free_groups.py
+++ b/sympy/combinatorics/free_groups.py
@@ -250,7 +250,7 @@ class FreeGroup(DefaultPrinting):
 
         """
         if self.rank == 0:
-            return 1
+            return S.One
         else:
             return S.Infinity
 
@@ -582,7 +582,7 @@ class FreeGroupElement(CantSympify, DefaultPrinting, tuple):
 
         """
         if self.is_identity:
-            return 1
+            return S.One
         else:
             return S.Infinity
 
--- a/sympy/combinatorics/schur_number.py
+++ b/sympy/combinatorics/schur_number.py
@@ -42,7 +42,7 @@ class SchurNumber(Function):
             if k is S.Infinity:
                 return S.Infinity
             if k.is_zero:
-                return 0
+                return S.Zero
             if not k.is_integer or k.is_negative:
                 raise ValueError("k should be a positive integer")
             first_known_schur_numbers = {1: 1, 2: 4, 3: 13, 4: 44}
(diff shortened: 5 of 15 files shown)


sympy/simplify/simplify.py
```diff
@@ -1241,7 +1241,7 @@ def cancel_kronecker_mul(m):
             args1 = delta1.args
             args2 = delta2.args
             if args_cancel(args1, args2):
-                return 0*m
```
smichr, line 1244: this one makes me a little nervous since `0*obj` might not return 0: it might return a type of zero as represented by that object.
```python
>>> Matrix([[1,2]])*0
Matrix([[0, 0]])
```

asmeurer: Surely this function only operates on Expr objects

asmeurer:
Everything here looks good to me.

sympy/simplify/simplify.py
smichr:
```python
>>> var('i', integer=1)
i
>>> simplify(eye(1)*KroneckerDelta(0, i) * KroneckerDelta(1, i))
Matrix([[0]])
```

smichr:
see previous comments and example - KD factors in a Mul may go to 0 but other non-KD factors may be present, hence we cann't return 0; a test should be added since your changing this did not trigger a failure.

sympy/simplify/simplify.py
oscargus: Yeah, I removed it to see if any tests failed. Which didn't happen. But we should probably add a test like above. I'll do that.

oscargus:
However, even with returning `S.Zero`:
```

In [3]: print(simplify(eye(1) * KroneckerDelta(0, n) *
   ...:                     KroneckerDelta(1, n)))
Matrix([[0]])
```

So another test is required. I have no idea which though....

oscargus:
It seems like the matrices are not fed to this part of the code, which is probably why it works. So as far as I can tell (not reading the code, just printing the arguments etc), this only handles the scalar part, and, hence, it should work with just returning zero.

`simplify(oo * KroneckerDelta(0, n) * KroneckerDelta(1, n))` differs though.

smichr:
> So another test is required.

My guess is that simplify separates the product into those with and those without the KD. The direct test would be `kroneckersimp(eye(1) * KroneckerDelta(0, n) * KroneckerDelta(1, n))`. Which routine should do the sorting? I favor the routine for which it is important (so the KD simplification); I would expect `if expr.has(KD): expr = KDsimp(expr)` in simplify -- or better, just `expr = KDsimp(expr)` and let `KDsimplify` decide if it should make a change or not. Then `simplify` is cleaner and one can see an overview of what sorts of steps are being taken.

oscargus:
In this case it seems like the following happens:
```
In [4]: srepr(eye(1) * KroneckerDelta(0, n) *
   ...:                     KroneckerDelta(1, n))
Out[4]: "MutableDenseMatrix([[Mul(KroneckerDelta(Integer(0), Symbol('n', integer=True)), KroneckerDelta(Integer(1), Symbol('n', integer=True)))]])"
```
so the KD ends up inside the matrix.

But the `oo` test that I added do require `0*m` to remain.

sympy/simplify/tests/test_simplify.py
```diff
@@ -938,6 +938,11 @@ def test_simplify_kroneckerdelta():
                                      [0, K(0, n), 0, K(1, n)],
                                      [0, 0, K(0, n), 0],
                                      [0, 0, 0, K(0, n)]])
+    assert simplify(eye(1) * KroneckerDelta(0, n) *
```
smichr, line 941: You already showed that these do not fail, so the necessity of the `0*m` has not been secured with a test. I'm guessing you have a push in progress, but just in case...

asmeurer:
The original `m*0` is indeed weird and we should try to figure out why it was put there. But it seems like we shouldn't have to put `0*x` everywhere instead of `0` or `S.Zero` to make things type safe. If we do there's a serious design issue somewhere.

sympy/simplify/tests/test_simplify.py
oscargus: Check two lines down. The oo test fails with returning 0.

smichr:
> a serious design issue somewhere

As I mentioned above, the sifting out of args to be simplified or not should probably happen in the `KDsimp` routine. And then a test affirming that a product with a Matrix and KDs sent to it returns the right answer even when returning 0.

asmeurer:
Traced it back to https://github.com/sympy/sympy/pull/17215, which did indeed seem to involve matrices. The original version was written by @oscarbenjamin in https://github.com/sympy/sympy/pull/17215#issuecomment-512563087. @oscarbenjamin was the `Mul(*((0,) + args))` (which later got turned into `0*m`) instead of just `0` there to handle matrices or for some other reason?

smichr:
That's another issue, however. For the sake of this PR, it is sufficient to leave that line alone and just add a test that calling KDsimp with such a product. That test should never fail regardless of how the internals are edited; edits should change the need to return `0*m`, however.

oscargus:
> If we do there's a serious design issue somewhere.

`0 * oo => NaN` which seems like the only case I can figure out. Matrices seems to move multiplication by scalars into the matrix, so there it seems to not happen (at least not after a quick check with some options...).

sympy/simplify/tests/test_simplify.py
```diff
@@ -938,6 +938,11 @@ def test_simplify_kroneckerdelta():
...
+                    KroneckerDelta(1, n)) == Matrix([[0]])
+
+    assert simplify(S.Infinity * KroneckerDelta(0, n) *
```
oscargus, line 944: @smichr Here is the actual test that will break if 0 * m is removed. I added the earlier just because I didn't want to remove it...

asmeurer:
> 0 * oo => NaN which seems like the only case I can figure out. Matrices seems to move multiplication by scalars into the matrix, so there it seems to not happen (at least not after a quick check with some options...).

It can happen with matrix expressions (MatrixSymbol), like `x*MatrixSymbol('A', n, n)`.

> As I mentioned above, the sifting out of args to be simplified or not should probably happen in the KDsimp routine. And then a test affirming that a product with a Matrix and KDs sent to it returns the right answer even when returning 0.

My point is that you could call any function on a MatMul, and expect it to just work (assuming it knows how to handle non-commutatives). And for that to happen it shouldn't need to know about matrices and filtering them, or little gotchas like using 0*x instead of a plain 0. Maybe a big function like simplify should manually filter out the scalar part (or commutative part if it thinks about it in that way). Otherwise we're forcing every simplification function to know about matrices, even if what it is doing has nothing to do with matrices or even noncommutatives (as would be the case with kroneckersimp).

oscargus:
> It can happen with matrix expressions (MatrixSymbol)

Actually not here. Only `Mul` is dealt with (not `MatMul`), so it works even when returning `S.Zero`.
```
In [3]: print(simplify(MatrixSymbol('a',1 ,1 ) * KroneckerDelta(0, n) *
   ...:                     KroneckerDelta(1, n)))
0*a

In [4]: srepr(MatrixSymbol('a',1 ,1 ) * KroneckerDelta(0, n) *
   ...:                     KroneckerDelta(1, n))
Out[4]: "MatMul(Mul(KroneckerDelta(Integer(0), Symbol('n', integer=True)), KroneckerDelta(Integer(1), Symbol('n', integer=True))), MatrixSymbol(Str('a'), Integer(1), Integer(1)))"
```
so only `Mul(KroneckerDelta(Integer(0), Symbol('n', integer=True)), KroneckerDelta(Integer(1), Symbol('n', integer=True)))` is simplified here.

smichr:
LGTM

oscarbenjamin:
> Otherwise we're forcing every simplification function to know about matrices, even if what it is doing has nothing to do with matrices or even noncommutatives (as would be the case with kroneckersimp).

We're not forcing anything. The `0*m` is really just replacing the Kronecker deltas with zero. It's up to `Mul.flatten` to determine the consequences of that.

PR #22202, opened 2021-10-02
Better return values and used S.Half

oscargus:
Follow-up on useful parts from #22171 not present in #22199.

* Used `S.Half` in library code instead of `S(1)/2` (still many tests using the latter).
* Removed some not required `S` in return values
* Used `Integer` instead of `S` for constant integer returns. (Or `S.One` etc...)

#### Other comments
The removed `S` in ambient dimension is for most similar functions returns a plain `int`.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/geometry/parabola.py
+++ b/sympy/geometry/parabola.py
@@ -96,7 +96,7 @@ class Parabola(GeometrySet):
         2
 
         """
-        return S(2)
+        return 2
 
     @property
     def axis_of_symmetry(self):
--- a/sympy/core/exprtools.py
+++ b/sympy/core/exprtools.py
@@ -74,27 +74,27 @@ def _monotonic_sign(self):
         s = self
         if s.is_prime:
             if s.is_odd:
-                return S(3)
+                return Integer(3)
             else:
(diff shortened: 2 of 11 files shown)


sympy/geometry/parabola.py
```diff
@@ -96,7 +96,7 @@ def ambient_dimension(self):
         2
 
         """
-        return S(2)
+        return 2
```
oscarbenjamin, line 99: I'm not sure about this. Should this be a symbolic return value?

oscargus: At the time I just went for a feeling that it is similar to `len`. However, checking now, all other `ambient_dimension` methods return `int` typically `len` of something.

For example:
https://github.com/sympy/sympy/blob/d168c74babaf92ed523cf503e7398a30dc9a235d/sympy/geometry/line.py#L138
https://github.com/sympy/sympy/blob/9dd704b071e5134a0e8beeeb1a2d2599536f5144/sympy/geometry/ellipse.py#L185

So at least it is consistent now. Correct? I do not know.

PR #22205, opened 2021-10-02
Restructured the imports in simplify with related changes

oscargus:
"Better" import order in simplify and a bit less top level dependencies. As much as possible is moved to the top level imports and the remaining are distributed to where they are needed (as there are returns earlier and so on).

The changes in `hyperexpand.py` are primarily to be able to top-level import that in `simplify.py`, which I think make sense as `simplify` should be the top level.

Change in `gruntz.py` is a "left-over" from a non-successful attempt to make it not import from simplify, but no harm in keeping it. 

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/diffgeom/diffgeom.py
+++ b/sympy/diffgeom/diffgeom.py
@@ -968,7 +968,6 @@ class BaseScalarField(Expr):
         coords = point.coords(self._coord_sys)
         # XXX Calling doit  is necessary with all the Subs expressions
         # XXX Calling simplify is necessary with all the trig expressions
-        from sympy.simplify.simplify import simplify
         return simplify(coords[self._index]).doit()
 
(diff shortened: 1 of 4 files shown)


sympy/diffgeom/diffgeom.py
```diff
@@ -1856,6 +1855,8 @@ def intcurve_diffequ(vector_field, param, start_point, coord_sys=None):
     intcurve_series
 
     """
+    from sympy.simplify import simplify
```
oscarbenjamin, line 1858: Can the import here just go at the end of the file?

Importing at the end of the file is better than importing in a function because it means that the import only happens once.

oscargus: Yes, it was possible here. Tried the other files as well, but not possible there.

PR #22206, opened 2021-10-02
Moved bottom_up, use, and walk to core

### LINKED ISSUE #22204: Move bottom_up and walk to simplify.traversaltools and move that file elsewhere
I've started to nest a bit in the import dependencies and one thing that will at least not make it harder is to move `bottom_up` and `walk` (and maybe more?) to `simplify.traversaltools`. On the other hand, it seems like these functions should maybe not be in `simplify` at all, but in `core` or `utilities` as they do not really rely on anything in `simplify` (they do hardly rely on anything from Sympy).

Then, the questions is where they should end up?

I assume that this involves deprecating the current locations as part of it.

oscargus:

Closes #22204 

See #22204 for discussion. 

#### Other comments
No deprecations yet.

Will add release notes when things converge regarding location etc.

#### Release Notes

NO ENTRY


sympy/simplify/simplify.py
```diff
@@ -1156,57 +1157,6 @@ def f(rv):
     return bottom_up(expr, f)
 
 
-def walk(e, *target):
```
oscarbenjamin, line 1159: Maybe add this back as a function with a deprecation warning that calls the other walk function from core.

oscargus: Will do that for all removed functions.

oscarbenjamin:
> This is ready for review. (Not sure about the performance degradation. Doesn't seem to be related.)

The performance change has shown up on a few PRs I think. I think that something has slowed down but it's close to the threshold for reporting so it doesn't always show up. It should be investigated with asv.

PR #21183, opened 2021-03-28
Fixes, #21147 . evalf()  doesn't works properly with evaluate = False

ayushbisht2001:
#### References to other Issues or PRs

fixed: #21147

#### Release Notes

NO ENTRY

PR #22211, opened 2021-10-03
Fix handling of zoo in evalf

### LINKED ISSUE #8242: (1/x).evalf(subs={x: 0}) produces 0
Here's how it looks like:

``` python
In [1]: sp.__version__
Out[1]: '0.7.5-git'

In [2]: x = sp.abc.x

In [3]: (1/x).evalf(subs={x: 0})
Out[3]: 0

In [4]: (1/x).subs({x: 0})
Out[4]: zoo

In [5]: (1/x).xreplace({x: 0})
Out[5]: zoo
```

I did investigate it some time ago and, AFAIR, it happens in `Pow` evaluation that short-circuits to 0 whenever the base is 0 while it should return infs if the exponent is negative.

### LINKED ISSUE #21147: Evalf and evaluate=False issues
`Mul(2, Pow(0,-1, evaluate=False), evaluate=False).evalf()` returns 0 instead of complex infinity.

eagleoflqj:

Fixes #8242
Fixes #21147
Closes #21183

Fix by allowing `evalf`s to return `zoo` in addition to mpf value tuple, as there's no equivalent mpc representation in mpmath. Check this return value in each `evalf`s and handle it.
#### Other comments

Some additional logics are added to `equals` and `is_constant` in order to fix newly broken tests. One case is leaving unfixed, which imho is acceptable because it’s a bit complicated and previously based on wrong evalf intermediate result.

#### Release Notes

* core
  * Evaluation of expressions involving zoo is now handled properly in evalf without raising obscure exceptions.

### DIFF
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -12,7 +12,7 @@ from mpmath import (
     make_mpc, make_mpf, mp, mpc, mpf, nsum, quadts, quadosc, workprec)
 from mpmath import inf as mpmath_inf
 from mpmath.libmp import (from_int, from_man_exp, from_rational, fhalf,
-        fnan, fnone, fone, fzero, mpf_abs, mpf_add,
+        fnan, finf, fninf, fnone, fone, fzero, mpf_abs, mpf_add,
         mpf_atan, mpf_atan2, mpf_cmp, mpf_cos, mpf_e, mpf_exp, mpf_log, mpf_lt,
         mpf_mul, mpf_neg, mpf_pi, mpf_pow, mpf_pow_int, mpf_shift, mpf_sin,
         mpf_sqrt, normalize, round_nearest, to_int, to_str)
@@ -206,6 +206,8 @@ def complex_accuracy(result):
     In the worst case (re and im equal), this is wrong by a factor
     sqrt(2), or by log2(sqrt(2)) = 0.5 bit.
     """
+    if result is S.ComplexInfinity:
+        return INF
     re, im, re_acc, im_acc = result
     if not im:
         if not re:
@@ -221,8 +223,10 @@ def complex_accuracy(result):
 
 
 def get_abs(expr, prec, options):
-    re, im, re_acc, im_acc = evalf(expr, prec + 2, options)
-
+    result = evalf(expr, prec + 2, options)
+    if result is S.ComplexInfinity:
+        return finf, None, prec, None
+    re, im, re_acc, im_acc = result
     if not re:
         re, re_acc, im, im_acc = im, im_acc, re, re_acc
     if im:
@@ -246,6 +250,8 @@ def get_complex_part(expr, no, prec, options):
     i = 0
     while 1:
         res = evalf(expr, workprec, options)
+        if res is S.ComplexInfinity:
+            return fnan, None, prec, None
         value, accuracy = res[no::2]
         # XXX is the last one correct? Consider re((1+I)**2).n()
         if (not value) or accuracy >= prec or -value[2] > prec:
@@ -289,6 +295,8 @@ def chop_parts(value, prec):
     """
     Chop off tiny real or complex parts.
     """
+    if value is S.ComplexInfinity:
+        return value
     re, im, re_acc, im_acc = value
     # Method 1: chop based on absolute value
     if re and re not in _infs_nan and (fastlog(re) < -prec + 4):
@@ -323,7 +331,10 @@ def get_integer_part(expr, no, options, return_ints=False):
     from sympy.functions.elementary.complexes import re, im
     # The expression is likely less than 2^30 or so
     assumed_size = 30
-    ire, iim, ire_acc, iim_acc = evalf(expr, assumed_size, options)
+    result = evalf(expr, assumed_size, options)
+    if result is S.ComplexInfinity:
+        raise ValueError("Cannot get integer part of Complex Infinity")
+    ire, iim, ire_acc, iim_acc = result
 
     # We now know the size, so we can calculate how much extra precision
     # (if any) is needed to get within the nearest integer
@@ -532,10 +543,17 @@ def evalf_add(v, prec, options):
         options['maxprec'] = min(oldmaxprec, 2*prec)
 
         terms = [evalf(arg, prec + 10, options) for arg in v.args]
+        n = terms.count(S.ComplexInfinity)
+        if n >= 2:
+            return fnan, None, prec, None
         re, re_acc = add_terms(
-            [a[0::2] for a in terms if a[0]], prec, target_prec)
+            [a[0::2] for a in terms if isinstance(a, tuple) and a[0]], prec, target_prec)
         im, im_acc = add_terms(
-            [a[1::2] for a in terms if a[1]], prec, target_prec)
+            [a[1::2] for a in terms if isinstance(a, tuple) and a[1]], prec, target_prec)
+        if n == 1:
+            if re in (finf, fninf, fnan) or im in (finf, fninf, fnan):
+                return fnan, None, prec, None
+            return S.ComplexInfinity
         acc = complex_accuracy((re, im, re_acc, im_acc))
         if acc >= target_prec:
             if options.get('verbose'):
@@ -568,19 +586,30 @@ def evalf_mul(v, prec, options):
     args = list(v.args)
 
     # see if any argument is NaN or oo and thus warrants a special return
+    has_zero = False
     special = []
     from .numbers import Float
     for arg in args:
-        arg = evalf(arg, prec, options)
-        if arg[0] is None:
+        result = evalf(arg, prec, options)
+        if result is S.ComplexInfinity:
+            special.append(result)
             continue
-        arg = Float._new(arg[0], 1)
-        if arg is S.NaN or arg.is_infinite:
-            special.append(arg)
+        if result[0] is None:
+            if result[1] is None:
+                has_zero = True
+            continue
+        num = Float._new(result[0], 1)
+        if num is S.NaN:
+            return fnan, None, prec, None
+        if num.is_infinite:
+            special.append(num)
     if special:
+        if has_zero:
+            return fnan, None, prec, None
         from .mul import Mul
-        special = Mul(*special)
-        return evalf(special, prec + 4, {})
+        return evalf(Mul(*special), prec + 4, {})
+    if has_zero:
+        return None, None, None, None
 
     # With guard digits, multiplication in the real case does not destroy
     # accuracy. This is also true in the complex case when considering the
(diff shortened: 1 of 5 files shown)


eagleoflqj:
> I don't want to add any type hints until mypy is running in CI

Then I’d wait #22212 to be merged first

sympy/core/expr.py
```diff
@@ -47,6 +47,8 @@ class Expr(Basic, EvalfMixin):
 
     is_scalar = True  # self derivative is 1
 
+    args: tuple
```
oscarbenjamin, line 50: It should be a tuple of Basic. See the same hint in `basic.py`. Is it necessary to add this hint for Expr or can it just inherit the hint from Basic?

sympy/core/evalf.py
```diff
@@ -82,7 +93,13 @@ class PrecisionExhausted(ArithmeticError):
 """
 
 
-def fastlog(x):
+MPF_VALUE_TUPLE = tTuple[int, int, int, int]
+SCALED_ZERO_TUPLE = tTuple[List[int], int, int, int]
+TEMP_RESULT = Any
```
oscarbenjamin, line 98: These names are quite long and I'm not sure what they really mean. If we're going to come up with names for things like this then they need to be informative names and there should be some comments here explaining what they mean. The name `TEMP_RESULT` does not convey much information to me and the fact that it is an alias for `Any` means that it gives the appearance of a meaningful type hint when in fact the hint says nothing.

sympy/core/evalf.py
```diff
@@ -82,7 +93,13 @@ class PrecisionExhausted(ArithmeticError):
...
+# tUnion[tTuple[Optional[MPF_VALUE_TUPLE], Optional[MPF_VALUE_TUPLE], Optional[int], Optional[int]], 'ComplexInfinity']
+
+
+def fastlog(x: Optional[MPF_VALUE_TUPLE]) -> tUnion[int, Any]:
```
oscarbenjamin, line 102: Why is this `Optional`?

sympy/core/evalf.py
```diff
@@ -186,13 +210,13 @@ def scaled_zero(mag, sign=1):
         raise ValueError('scaled zero expects int or scaled_zero tuple.')
 
 
-def iszero(mpf, scaled=False):
+def iszero(mpf: tUnion[MPF_VALUE_TUPLE, SCALED_ZERO_TUPLE, None], scaled=False) -> Optional[bool]:
     if not scaled:
         return not mpf or not mpf[1] and not mpf[-1]
     return mpf and isinstance(mpf[0], list) and mpf[1] == mpf[-1] == 1
```
oscarbenjamin, line 216: Is the return type here not just `bool`?

sympy/core/evalf.py
```diff
@@ -205,6 +229,8 @@ def complex_accuracy(result):
     In the worst case (re and im equal), this is wrong by a factor
     sqrt(2), or by log2(sqrt(2)) = 0.5 bit.
     """
+    if not isinstance(result, tuple):
+        return INF
```
oscarbenjamin, line 233: When would this be called with something that isn't a tuple? Maybe that should be fixed at the calling site rather than here.

sympy/core/evalf.py
```diff
@@ -284,10 +315,12 @@ def finalize_complex(re, im, prec):
     return re, im, re_acc, im_acc
 
 
-def chop_parts(value, prec):
+def chop_parts(value: TEMP_RESULT, prec: int) -> TEMP_RESULT:
     """
     Chop off tiny real or complex parts.
     """
+    if not isinstance(value, tuple):
+        return value
```
oscarbenjamin, line 323: When is this called with something that is not a tuple?

sympy/core/expr.py
eagleoflqj: Indeed unnecessary. I should instead add type hints to sub classes of `Expr`.

sympy/core/evalf.py
eagleoflqj: For `MPF_VALUE_TUPLE` and `TEMP_RESULT`, I didn't come up with them randomly, they are just below the comment https://github.com/sympy/sympy/blob/9019765edb0d0b4bbd29674a58b93ef4e75b92f8/sympy/core/evalf.py#L60-L83 which already "define" these names.

eagleoflqj: For `SCALED_ZERO_TUPLE`, it's used in the functions related to `scaled_zero` and can be found at https://github.com/sympy/sympy/blob/9019765edb0d0b4bbd29674a58b93ef4e75b92f8/sympy/core/evalf.py#L152-L156

eagleoflqj: `TEMP_RESULT` is `Any` because I failed to find a way that makes mypy happy. I originally use the type `tUnion[tTuple[Optional[MPF_VALUE_TUPLE], Optional[MPF_VALUE_TUPLE], Optional[int], Optional[int]], 'ComplexInfinity']` (commented at the following line) and some variants of it, but they couldn't reflect the fact that if the first argument of the tuple is `None` then the third must also be `None`, and sometimes a temp_result is impossible to be `ComplexInfinity` but mypy still requires to check this case. Dealing with these cases neither helps to fix the bug that this PR aims to fix, nor helps understand and refactor the code in evalf.py (at least now).
Moreover, `TEMP_RESULT` provides a way that unifys the return type of the `evalf`s. So imho it's better than nothing or tons of `Any`s here.

sympy/core/evalf.py
eagleoflqj: Because the body of `fastlog` expect this case.
https://github.com/sympy/sympy/blob/9019765edb0d0b4bbd29674a58b93ef4e75b92f8/sympy/core/evalf.py#L117

sympy/core/evalf.py
eagleoflqj: `iszero(None, True)` returns `None`. This may happen in https://github.com/sympy/sympy/blob/9019765edb0d0b4bbd29674a58b93ef4e75b92f8/sympy/core/evalf.py#L554

sympy/core/evalf.py
eagleoflqj: `zoo`, which isn't represented in mpmath. Originally `TEMP_RESULT` is not `Any` and `if not isinstance(result, tuple)` may make my type checker happy. I will change all of this to `if result is S.ComplexInfinity`.

sympy/core/evalf.py
eagleoflqj: same above

sympy/core/evalf.py
oscarbenjamin: If `TEMP_RESULT` is just an alias for `Any` then how is it better than `Any`?

Note that my main point in the comment above is that there should be comments in the code explaining what these types mean and why they are defined in this way.

sympy/core/evalf.py
oscarbenjamin: I guess what I mean is when is this called with `None`?

sympy/core/expr.py
```diff
@@ -752,6 +749,12 @@ def equals(self, other, failing_expression=False):
             # then this can't be a zero
             return False
 
+        factors = diff.as_coeff_mul()[1]
+        if len(factors) > 1:  # avoid infinity recursion
+            fac_zero = [fac.equals(0) for fac in factors]
+            if None not in fac_zero:  # every part can be decided
+                return any(fac_zero)
```
oscarbenjamin, line 756: Apart from the type hints is this the main change in this PR?

oscarbenjamin: Note that mixing changes together like this makes the diff much more difficult to review because the important behavioural changes are buried in a sea of unrelated changes.

eagleoflqj: No the main change is in `evalf_pow`.
Sorry about mixing changes together. Originally I only want to fix the `0**-1` evalf bug, then I feel the `evalf.py` is hard to modify without type hints, then the fix makes some tests fail and I have to patch other modules, then `mypy` reports tons of errors...
Since CI passes now, I will re-arrange the commits by logic.

sympy/core/evalf.py
eagleoflqj: > If `TEMP_RESULT` is just an alias for `Any` then how is it better than `Any`?

Because you can not tell 2 `Any`s are actually the same type or not, but alias convey the information that they are the same type.

eagleoflqj: > Note that my main point in the comment above is that there should be comments in the code explaining what these types mean and why they are defined in this way.

`An mpf value tuple is a tuple of integers (sign, man, exp, bc) ...` clearly explains why `MPF_VALUE_TUPLE = tTuple[int, int, int, int]` and what these 4 `int`s are. In fact I don't quite understand what scaled_zero tuple is, this name already appears in the comment, confusing me, so I simply add type hints for it.

oscarbenjamin: It's okay to write a comment in the code saying that you don't understand something. That kind of comment is actually very helpful to other readers,

sympy/core/evalf.py
eagleoflqj: https://github.com/sympy/sympy/blob/d04fa46664c3aeda08fb41a2b2d62055056e8926/sympy/core/evalf.py#L880-L881
or
https://github.com/sympy/sympy/blob/d04fa46664c3aeda08fb41a2b2d62055056e8926/sympy/core/evalf.py#L1000-L1006
where the argument of `fastlog` is directly from a call to `evalf`s without checking `None`.

oscarbenjamin:
Can you split this PR into two PRs where one fixes the cited issues and the other adds type hints?

eagleoflqj:
> Can you split this PR into two PRs where one fixes the cited issues and the other adds type hints?

Much clearer now

sympy/core/tests/test_evalf.py
```diff
@@ -628,3 +633,28 @@ def test_issue_20291():
...
+
+
+def test_evalf_with_zoo():
+    assert (1/x).evalf(subs={x: 0}) == zoo  # issue 8242
+    assert (-1/x).evalf(subs={x: 0}) == zoo  # PR 16150
+    assert (0 ** x).evalf(subs={x: -1}) == zoo  # PR 16150
+    assert (0 ** x).evalf(subs={x: -1 + I}) == nan
+    assert Mul(2, Pow(0, -1, evaluate=False), evaluate=False).evalf() == zoo  # issue 21147
+    assert Mul(x, 1/x, evaluate=False).evalf(subs={x: 0}) == Mul(x, 1/x, evaluate=False).subs(x, 0) == nan
+    assert Mul(1/x, 1/x, evaluate=False).evalf(subs={x: 0})
```
oscarbenjamin, line 645: There seems to be a missing `== zoo` here

sympy/functions/elementary/exponential.py
```diff
@@ -1055,6 +1055,9 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
             return self.func(x0) - 2*I*S.Pi
         return self.func(arg)
 
+    def is_constant(self, *wrt, **flags):
+        return self.args[0].is_constant(*wrt, **flags)
```
oscarbenjamin, line 1059: Why is this needed for `log`? No other function class has this.

oscarbenjamin:
This is much clearer, thanks. I think this looks good although I left a couple of comments.

It seems a bit awkward that everywhere in `evalf` has to check for `zoo` but I can't really think of a better way of handling it.

sympy/functions/elementary/exponential.py
eagleoflqj: It was imported when fixing test_issue_22020 in test_expr.py
Seems without it that case can still pass now. I'll remove it to see what CI says.

oscarbenjamin:
I changed the title of the PR and also edited the release note.

This looks good. Thanks for this!

oscarbenjamin:
I bet this fixes many more issues that aren't explicitly listed in the OP as well.

PR #20173, opened 2020-09-30
maint(ci): add mypy type checking to Travis

oscarbenjamin:
#### References to other Issues or PRs

Previous work is in #20140 and #18244

#### Brief description of what is fixed or changed

Adds a mypy job on Travis so that mypy is tested on all PRs.

#### Other comments

I deliberately didn't do this at the time of #18244 because I found mypy somewhat flakey. I think it can help improve the codebase but it is not necessarily that intuitive. If we enable this then all contributors will need to deal with mypy (it is not as easy as flake8 so help will be needed).

#### Release Notes

NO ENTRY

PR #20140, opened 2020-09-24
maint: add type hints to fix mypy warnings

oscarbenjamin:
#### Brief description of what is fixed or changed

Adds type hints to fix new mypy warnings that have emerged since #18244

#### Release Notes

NO ENTRY

PR #18244, opened 2020-01-06
Fix some mypy errors to get mypy working

oscarbenjamin:
#### Brief description of what is fixed or changed

This PR makes a lot of small changes and adds a lot of type-hints to make it possible to use myp on the sympy codebase.

Particular changes:
* Contains a fix for using py3 syntax even though the PY2 import test is still there (see .travis.yml).
* Remove all uses of `with_metaclass`. These were needed for PY2 compatibility but are not needed now.
* Replace all uses of list for `__slots__` with tuple. I thought this was necessary for mypy. I no longer think that it is but I still think it's a good change because `__slots__` doesn't need to be mutable.
* Add type hints in lots of places that create empty dicts/lists - this is comething that mypy complains about.
* Ensure that all definitions of a function are equivalent from a type perspective when types are defined/imported conditionally.
* Add type:ignore in lots of places particularly in multipledispatch.
* Tell mypy to ignore some internal code (rubi, antr, benchmarks and pygletplot) because they are autogenerated or unmaintained.

#### Other comments

This PR is the minimum needed to get mypy working so that we can *begin* adding type-hints and finding/preventing bugs.

At this stage I don't recommend making mypy a compulsory part of the Travis test suite because it is confusing. It could be added as an allowed failure but not in this PR which is intentionally minimal.

#### Release Notes

NO ENTRY

PR #22212, opened 2021-10-03
quality(mypy): Add type hints and fix all mypy warnings/errors

oscarbenjamin:

The general issue for adding type hints is #17945 so probably high-level discussion belongs there.

See also #20173 which did not get merged. At the time it fixed all mypy errors and there were only a few. The fixes in this PR are what is needed after 1 year of not running mypy in CI.

Also previous PRs that originally fixed all the mypy complaints: #20140 and #18244

Adds type hints and `type: ignore` to silence all mypy errors. There are a few places where the mypy errors indicate genuine bugs such as a method being defined in a superclass where it can only really work for a particular subclass. Otherwise mostly this is just about adding hints and adding ignores.

#### Other comments

Most of the ignores are needed due to https://github.com/python/mypy/issues/2904.

Some of them actually indicate genuine bugs that I can't easily fix for example the tensor module has a lot of code that does inconsistent things with different types of object that probably would fail under some conditions. Ironically it is also the module with the most type hints and many of those hints are incorrect and inconsistent.

The bulk of the work in making this PR was not about adding hints but about fixing incorrect hints. I think that since contributors are beginning to add type hints anyway sympy needs to start running mypy in CI to make sure that the hints are not garbage. I'll do that in a separate PR though after this one is merged.

#### Release Notes

* other
    * Type hints were added to silence all mypy errors in the sympy codebase.

### DIFF
--- a/sympy/stats/stochastic_process_types.py
+++ b/sympy/stats/stochastic_process_types.py
@@ -11,8 +11,8 @@ from sympy import (Matrix, MatrixSymbol, S, Indexed, Basic, Tuple, Range,
                    linsolve, eye, Or, Not, Intersection, factorial, Contains,
                    Union, Expr, Function, exp, cacheit, sqrt, pi, gamma,
                    Ge, Piecewise, Symbol, NonSquareMatrixError, EmptySet,
-                   ceiling, MatrixBase, ConditionSet, ones, zeros, Identity,
-                   Rational, Lt, Gt, Le, Ne, BlockMatrix, Sum, MatrixExpr)
+                   ceiling, ConditionSet, ones, zeros, Identity,
+                   Rational, Lt, Gt, Le, Ne, BlockMatrix, Sum)
 from sympy.core.relational import Relational
 from sympy.logic.boolalg import Boolean
 from sympy.utilities.exceptions import SymPyDeprecationWarning
--- a/sympy/stats/stochastic_process_types.py
+++ b/sympy/stats/stochastic_process_types.py
@@ -347,6 +347,13 @@ class MarkovProcess(StochasticProcess):
     """
 
     @property
+    def number_of_states(self) -> tUnion[Integer, Symbol]:
+        """
+        The number of states in the Markov Chain.
+        """
+        return _sympify(self.args[2].shape[0]) # type: ignore
+
+    @property
     def _state_index(self):
         """
         Returns state index as Range.
@@ -918,13 +925,6 @@ class DiscreteMarkovChain(DiscreteTimeStochasticProcess, MarkovProcess):
                 indices[state] = index
         obj.index_of = indices
         return obj
-
-    @property
-    def number_of_states(self) -> tUnion[Integer, Symbol]:
-        """
-        The number of states in the Markov Chain.
-        """
-        return _sympify(self.args[2].shape[0]) # type: ignore
 
     @property
     def transition_probabilities(self):
--- a/sympy/stats/stochastic_process_types.py
+++ b/sympy/stats/stochastic_process_types.py
@@ -2,7 +2,8 @@ from __future__ import print_function, division
 import random
 
 import itertools
-from typing import Sequence as tSequence, Union as tUnion, List as tList, Tuple as tTuple
+from typing import (Sequence as tSequence, Union as tUnion, List as tList,
+        Tuple as tTuple, Set as tSet)
 
 from sympy import (Matrix, MatrixSymbol, S, Indexed, Basic, Tuple, Range,
                    Set, And, Eq, FiniteSet, ImmutableMatrix, Integer, igcd,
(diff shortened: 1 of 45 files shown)


oscarbenjamin:
To test the effect of this PR try running `mypy sympy` on master. With this PR `mypy sympy` does not report any errors. On master you get 168 errors but actually once I started fixing some of them more and more appeared. It seems like mypy can't handle the sympy codebase on current master and just bails after a while.

sympy/stats/stochastic_process_types.py
```diff
@@ -2,15 +2,16 @@
 import random
 
 import itertools
-from typing import Sequence as tSequence, Union as tUnion, List as tList, Tuple as tTuple
+from typing import (Sequence as tSequence, Union as tUnion, List as tList,
+        Tuple as tTuple, Set as tSet)
```
asmeurer, line 6: This naming convention isn't the greatest (not that I can think of a better one). Maybe we should require all typing imports to come from `sympy.typing` which could provide common alias names. Even the fact that some files import Tuple or Dict as is because they don't use the SymPy Tuple or Dict is confusing.

sympy/tensor/array/expressions/conv_matrix_to_array.py
```diff
@@ -1,11 +1,11 @@
-from sympy import Mul, Basic, MatMul, MatAdd, Transpose, Trace, Pow, \
+from sympy import Integer, Mul, Basic, MatMul, MatAdd, Transpose, Trace, Pow, \
     MatPow, symbols, Dummy, Lambda, HadamardProduct, HadamardPower, S
 from sympy.matrices.expressions.matexpr import MatrixExpr
 from sympy.tensor.array.expressions.array_expressions import ArrayDiagonal, ArrayTensorProduct, \
     PermuteDims, ArrayAdd, ArrayContraction, ArrayElementwiseApplyFunc
 
 
-def convert_matrix_to_array(expr: MatrixExpr) -> Basic:
+def convert_matrix_to_array(expr: Basic) -> Basic:
```
asmeurer, line 8: Why is Expr not good enough here?

oscarbenjamin: The function calls itself recursively with the args of its argument. In general the args of `Expr` are not known to be `Expr` e.g.:
```python
In [11]: M = MatrixSymbol('X', 2, 2)

In [12]: isinstance(M, Expr)
Out[12]: True

In [13]: isinstance(M.args[0], Expr)
Out[13]: False
```
It should be possible to add type hints to each of the different classes here like `MatMul` could have
```python
    args: Tuple[Expr, ...]
```
I expect that adding those type hints will throw up more errors though when I'm actually trying to silence the errors first so that we can run mypy and then add type-hints properly later.

sympy/stats/stochastic_process_types.py
oscarbenjamin: Given the ubiquity of typing now in common editors and so on I wonder if sympy needs to bite the bullet and start calling them `symTuple` or `BasicTuple` or something so that `Tuple` can just be for typing. Or actually maybe there was a PEP that was going to change it so you could just do `tuple[int]` etc...

eagleoflqj: I'm also adding type hint, and I expect an official naming convention to be documented somewhere.

oscarbenjamin:
I realise that there can be improvements here but really I would just like to get this in and enable mypy in CI and then further improvements can be made while CI is running mypy. I have fixed all mypy errors several times now and I just want mypy to be running in CI so that the next time I come to look at this there won't be a load of new errors to fix.

As I said in the OP:

> See also #20173 which did not get merged. At the time it fixed all mypy errors and there were only a few. The fixes in this PR are what is needed after 1 year of not running mypy in CI.

Does anyone object to this being merged as it is?

redeboer (changes requested):
Great work, good to see that mypy can soon be enforced through CI!

Could you perhaps add (and install) a file `py.typed` next to `sympy/__init__.py`? Then mypy sees SymPy as typed when it checks downstream packages.
https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

oscarbenjamin:
> Could you perhaps add (and install) a file `py.typed` next to `sympy/__init__.py`?

I can but can that also be done in a subsequent PR? The intention in this PR is to silence warnings from `mypy sympy`. My next planned PR is to enable mypy checking in GitHub actions. After that many changes can be made by many contributors to improve the typing situation but we can then do those with mypy enabled in CI. This PR already has a sizeable diff and I don't want to add to that any more than is needed.

redeboer (approved):
Okay LGTM

oscargus:
Let's merge this and then the repackaging and CI integration can come in subsequent PRs.

Issue #22105, opened 2021-09-16
Use autoapi for the internal reference documentation


nanjekyejoannah:
@nanjekyejoannah didn't mention it, but we will have a section "internal reference" section which will document all functions in SymPy, both public and private, and will be organized completely based on the code structure. We can use a tool like autoapi and autosummary to generate this automatically (see https://stackoverflow.com/questions/2701998/sphinx-autodoc-is-not-automatic-enough/).

asmeurer:
I think we might run into issues with docstrings that are not yet in Sphinx and will fail the build. For testing, we should try to do this for just one file or submodule just to see what it looks like so we avoid issues with trying to pull in the entire codebase.

nanjekyejoannah:
I opened a PR that exercises one submodule see #22214

asmeurer:
Another option is to use sphinx.ext.autosummary with the recursive option. That is what is described in one of the stackoverflow answers. 

I created a file with

```
.. autosummary::
   :recursive:

   sympy.calculus
```

and it generated a file that looks fine to me. It also didn't have any build errors. Furthermore, based on the SO answer, it looks like autosummary is easier to configure, since you can modify the template. And it doesn't involve running any commands to generate rst files. It is itself just an rst directive. So if we need to do something special for some specific module for some reason, it will be easy to do.

asmeurer:
Maybe we can copy the template files from https://github.com/JamesALeedham/Sphinx-Autosummary-Recursion

moorepants:
I'm curious what will be labeled as "private". SymPy doesn't really have much private code, i.e. most all of SymPy has always been importable with no notion of privateness. If things are now going to be called private, I think it is most appropriate to have a fairly long deprecation cycles to transform the names of private classes, methods, variables, functions by preprending underscores. Leading underscores are the only softly accepted method of signalling to users not to rely on that functionally. Suddenly calling some things private in the docs and thus allowing people to change those things without deprecation is only going to cause downstream frustration.

asmeurer:
"Private" may not be the best name for this anymore. We originally wanted a way to include internal functions in Sphinx, but ended up with a design that includes everything, without discrimination. There is also "public" which we want to make only include things that are to be considered public API going forward. 

Having better documentation on what is and isn't private is definitely something we want to improve. I don't think we can rely solely on underscores. 

[...]

Issue #22006, opened 2021-09-02
Categorize SymPy Modules


nanjekyejoannah:
To avoid having a big TOC, I propose categorizing the modules in the SymPy Modules Reference section of the documentation.

cc @asmeurer @oscarbenjamin

asmeurer:
We should try to come up with ~8 high level sections that we can organize the public API reference documentation into.

asmeurer:
Here's the current reference documentation. It's organized by SymPy submodules, but that's not necessarily the best organization for end-users. https://docs.sympy.org/latest/modules/index.html

Here's my initial thoughts on organization, based on the headers that are there. 
This is just my initial thoughts, so please critique. 

**Main category**. I'm not sure what this should actually be called, but it's
what 90% of users of SymPy will use. It consists of anything using basic
expressions, and functions that interact with them. This might need to be
split into sub-categories. Some of the categories below may make more sense as
subcategories of this.
```
abc
Calculus
Concrete
Core
Numerical Evaluation
Functions
Inequality Solvers
Integrals
Numeric Computation
ODE
PDE
Series
Simplify
Solvers
Solveset
```

**Code Generation**
```
Code Generation
```

**Matrices** Maybe this should just be a subsection of the main category. Or alternately the subsections of the main category could be top-level sections (but there would be more than ~8 in that case).
```
Matrices
Tensor
Vector
```

**Logic**
```
Logic
Assumptions
Sets
```

**Number Theory**
```
Combinatorics
Diophantine
Number Theory
```

**Geometry**
```
Geometry
```

**Polynomials** This might need a better name too. It's anything that involves
more direct calculations on things like polynomials or rings.
```
Algebras
Discrete
Polynomial Manipulation
```

**Plotting**
```
Plotting
```

**Physics** This should have subcategories for the different sub-submodules of sympy.physics.
```
Physics
```

**Utilities** Anything that isn't directly symbolic computation.
```
Interactive
Parsing
Printing
Testing
Utilities
```

**Other** Things that are mostly orthogonal to each other and don't fit
anywhere else. We could also make other categories for some of these items. "Stats" and "differential geometry" are a little more important so maybe they shouldn't be relegated to a miscellaneous category. 
```
Category Theory
Cryptography
Differential Geometry
Holonomic
Lie Algebra
Stats
```

I didn't include the Meijer G-function sections, which are part of integrals
(and shouldn't be in the reference section), or term rewriting, which isn't a
reference section and should probably go in the user guides. 

[...]

PR #22214, opened 2021-10-04
Internal API reference: sympy.abc

nanjekyejoannah:
This PR adds docs for the ``Private API reference `` section and specifically the ``sympy.abc`` module.

https://github.com/sympy/sympy/issues/22105

The docs in the directories ``modules`` have been re-categorized, see [Categorize SymPy Modules](https://github.com/sympy/sympy/issues/22006)

#### Release Notes

NO ENTRY

### DIFF
--- a/doc/src/_templates/custom-class-template.rst
+++ b/doc/src/_templates/custom-class-template.rst
@@ -1,0 +1,35 @@
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
+
+.. autoclass:: {{ objname }}
+   :members:
+   :show-inheritance:
+   :inherited-members:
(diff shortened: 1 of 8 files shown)


nanjekyejoannah:
This is the first iteration of the internal reference as agreed on in #22105, we can use this to discuss this.

asmeurer:
Can you update the installation instructions? I tried installing `autoapi` but it still told me it couldn't find the extension.

asmeurer:
This is generating rst files with the docstrings. We should use sphinx-apidoc instead, which will generate files with autodoc references. We will need to add :noindex: to the templates. We will also need to fix any RST issues with a bunch of docstrings.

PR #22222, opened 2021-10-05
Documentation cleanup 2

oscargus:
More documentation improvements in terms of formatting.

Added lots of functions from `logic` that were currently not in the documentation.

#### Release Notes

* logic
    * Many more functions added to documentation.


doc/src/modules/logic.rst
```diff
@@ -85,27 +90,76 @@ Boolean functions
...
+.. autofunction:: sympy.logic.boolalg::is_nnf
+
 Simplification and equivalence-testing
 --------------------------------------
 
 .. autofunction:: sympy.logic.boolalg::simplify_logic
 
-SymPy's simplify() function can also be used to simplify logic expressions to their
+SymPy's ``simplify()`` function can also be used to simplify logic expressions to their
 simplest forms.
```
oscarbenjamin, line 121: Can this be a clickable link like
```
    :py:func:`~.simplify`
```

sympy/logic/boolalg.py
```diff
@@ -360,13 +360,13 @@ def as_set(self):
...
-    This is the SymPy version of False, for use in the logic module. The
-    primary advantage of using false instead of False is that shorthand boolean
-    operations like ~ and >> will work as expected on this class, whereas with
-    False they act bitwise on 0. Functions in the logic module will return this
-    class when they evaluate to false.
+    This is the SymPy version of ``False``, for use in the logic module. The
+    primary advantage of using ``false`` instead of ``False`` is that shorthand
+    Boolean operations like ``~`` and ``>>`` will work as expected on this class,
+    whereas with ``False`` they act bitwise on 0. Functions in the logic module
+    will return this class when they evaluate to false.
```
oscarbenjamin, line 369: It also supports all the other Basic methods like `subs` etc. That's the real reason it is needed: Basic args need to be Basic.

sympy/logic/boolalg.py
```diff
@@ -1584,7 +1593,7 @@ def conjuncts(expr):
 
 
 def disjuncts(expr):
-    """Return a list of the disjuncts in the sentence s.
+    """Return a list of the disjuncts in ``expr``.
```
oscarbenjamin, line 1596: I think the style guide recommends using `*expr*` for function parameters.

sympy/logic/boolalg.py
```diff
@@ -2782,14 +2797,14 @@ def bool_monomial(k, variables):
 
     Each boolean function can be uniquely represented by a
     Zhegalkin Polynomial (Algebraic Normal Form). The Zhegalkin
-    Polynomial of the boolean function with n variables can contain
-    up to 2^n monomials. We can enumarate all the monomials.
+    Polynomial of the boolean function with `n` variables can contain
+    up to `2^n` monomials. We can enumarate all the monomials.
```
oscarbenjamin, line 2801: I also wonder if things like this should use `$2^n$` since the mathdollar sphinx extension is used.

doc/src/modules/logic.rst
oscargus: Would make sense!

sympy/logic/boolalg.py
oscargus: I was not really considering exactly what it said, focused on the formatting... ;-)

sympy/logic/boolalg.py
oscargus: OK!

sympy/logic/boolalg.py
oscargus: I know about that, but I use Spyder and there $ does not work. Backtick doesn't work properly either, but at least turns it into italic and removed the backspaces. Hence, I'm not really keen on the $ as it should do the same thing as backtick (but worse).

sympy/logic/boolalg.py
oscargus: I do not agree with this. 1) it looks bad, using double backticks makes it look like code and it is much more clear what one refer to, 2) it quite often says "if simplify=True" in the text and formatting that like `if *simplify*=...` doesn't make sense.

oscarbenjamin: I'm not sure where the `*param*` recommendation came from

PR #22227, opened 2021-10-05
could_extact_minus_sign, etc.. (distinguish between 1-x and x-1)

### LINKED ISSUE #22189: simplify with Abs

### LINKED ISSUE #22201: invariant is violated with current `Expr.could_extract_minus_sign`
It should be true that `Eq(e, -e) | Ne(e.could_extract_minus_sign(), (-e).could_extract_minus_sign())` but it is not:
```python
>>> (1-x).could_extract_minus_sign()
True
>>> (x-1).could_extract_minus_sign()
False
>>> ((x-1)*y).could_extract_minus_sign()
False
>>> ((1-x)*y).could_extract_minus_sign()  # should be true
False
```

cf, #4077, #22189, #22194

smichr:

fixes #22201
fixes #22189 
closes #22194 as alternate

Once the invariant that `e` and `-e` must give different results from `could_extract_minus_sign` (or else be equal) was enforced (as far as I can now tell) it was possible to fix `signsimp` so it would distinguish between `1 - x` and `x - 1` and return one of them as a `Mul` when `evaluate=False`:

```python
>>> signsimp(1-x, evaluate=False)
-(x - 1)
>>> signsimp(x-1, evaluate=False)
x - 1
>>> abs(1 - sqrt(x)) == abs(sqrt(x) - 1)
True
```

#### Other comments

The purpose of `could_extract_minus_sign` is to quickly distinguish between two forms of an expression; it is not necessary to have an expression be in canonical order wrt signs -- that is the purpose of `signsimp`. So the tests should be fast and in the case of a tie a fast tie-breaker should be used. It's nice if that tie-breaker will select something that mathematically looks like it should be negated to look "normal". In keeping with the current convention, the expression `x - 1` is considered "normal" wrt to signs while `1 - x` is the one that is thought of as `-(x - 1)`. In order to make `could_extract_minus_sign` faster, it no longer uses the `extract_multiplicatively` routine: it looks for literal negative signs.

`extract_multiplicatively` was modified to better handle the case of extracting a `-1` from an expression. This is more expensive than `could_extract_minus_sign` because it uses recursion, e.g.

```python
>>> 1_x
1 - x
>>> _.extract_multiplicatively(-1)
x - 1
>>> _.extract_multiplicatively(-1)  is None  # expr is canonical
True

>>> y*1_x
y*(1 - x)
>>> _.extract_multiplicatively(-1)
y*(x - 1)
>>> _.extract_multiplicatively(-1) is None # because no more can be extracted
True

>>> -y*1_x
-y*(1 - x)
>>> _.extract_multiplicatively(-1) # doesn't give None because we stop after finding an extraction
y*(1 - x)
>>> _.extract_multiplicatively(-1)  # there is still another
y*(x - 1)
>>> _.extract_multiplicatively(-1) is None # because there are no more
True
```
I think `could_extract_minus_sign` is working too hard on the `Mul` and could/should be more literal as `extract_multiplicatively` is:

```python
>>> 1_x.could_extract_minus_sign()
True
>>> (-1_x).could_extract_minus_sign()
False
>>> (y*1_x).could_extract_minus_sign()  ### perhaps this one should be false instead of recursing through args
True
>>> (-y*1_x).could_extract_minus_sign()  ### perhaps this should be true by merit of the leading sign
False
```

`signsimp` works the hardest and it should always be true that `signsimp(a) == -signsimp(-a)`

```python
>>> signsimp(y*1_x)
-y*(x - 1)
>>> signsimp(-y*1_x)
y*(x - 1)
```

**EXTENSIONS**

* The function `hide_sign` was created to push a sign into an expression if possible

```
>>> from sympy.core.mul import hide_sign
>>> Mul(-1, 1 - x,evaluate=False)
-(1 - x)
>>> hide_sign(_)
x - 1

>>> hide_sign(-y*1_x)  # it doesn't push if expr is not an Add
-y*(1 - x)
>>> hide_sign(-y*1_x, force=True)  # but you can force it to do so
y*(x - 1)
```

* _excess_neg_args is used to indicate whether there are an excess of negative signs in an Add; None indicates a tie. Maybe it would be better to return 0 for a tie so `>` and `<` could be used without an explicit test for None.
* _more_minus will handle any basic expression and combines `_excess_neg_args` and `_coeff_isneg`
* `_negate` pushes a sign onto an expression, into an Add factor of a Mul if necessary so `_negate(y*(1 - x))` is `y*(x - 1)` rather than `-y*(1 - x)`. This might be an unneeded feature outside of `hide_sign`.

**KEEP IT SIMPLE**

One of the themes that keeps coming up in discussions is the need to keep simplification separated from eval of instantiation. The trig functions are a place where some modifications were made. The desire in eval is to have `cos(-x) == cos(x)`. At the simplest level, this just means that there is a negative sign that should be removed. I think what happens in a lot of `eval` codes is that the idea of `canonical` is confused with evaluation rules. e.g. If we remove a negative sign from a Mul or Symbol, should we put an Add in canonical form, too, so `cos(1 - x) == cos(x - 1)`? This is the motivation for a function like `could_extract_minus_sign`. But what about `cos(y*(1 - x))`? `could_extract_minus_sign` says that a sign could be extracted -- should it? -- but negating it would not lead to a simpler expression, it would give `cos(-y*(1 - x))`. So I changed the query for the trig functions to be more literal in extraction of signs (though we might change the `could_extract_minus_sign` to be more literal in the case of Mul) with hopes that they would be faster. This should help with expressions involving many trig functions with non-trivial args.

I am leaving this as a draft, hoping to get some helpful feedback and suggested directions. I think that making `could_extract_minus_sign` more literal for a Mul would be a good idea. I'll have to check the code-base to see how it is being used.

#### Release Notes

NO ENTRY


sympy/core/expr.py
```diff
@@ -2284,16 +2284,23 @@ def extract_multiplicatively(self, c):
...
-                    else:
-                        xc = cs.extract_multiplicatively(c)
-                    if xc is not None:
-                        return xc*ps  # rely on 2-arg Mul to restore Add
-                return  # |c| != 1 can only be extracted from cs
+            if c.is_Number:
+                if c is S.NegativeOne:
+                    neg_args = sum(1 for _ in self.args if _coeff_isneg(_))
+                    xs = 2*neg_args - len(self.args)  # excess negatives
+                    neg = (xs > 0) if xs else bool(
```
smichr, line 2291:
```suggestion
                    xs = self._excess_neg_args
                    neg = xs if xs is not None else bool(
```

sympy/core/expr.py
```diff
@@ -4108,7 +4112,7 @@ def search_element(self, elem):
 from .mul import Mul
 from .add import Add
 from .power import Pow
-from .function import Function, _derivative_dispatch
+from .function import Function, _derivative_dispatch, _coeff_isneg
```
smichr, line 4115:
```suggestion
from .function import Function, _derivative_dispatch
```

smichr:
cf #22232

PR #22232, opened 2021-10-07
`signsimp` handles Add better and `could_extract_minus_sign` more literal

### LINKED ISSUE #22189: simplify with Abs

smichr:

fixes #22189
closes #22194 as an extension

`signsimp` was not extracting the sign from an expression like `e = sqrt(2*x) - 1` when evaluate was false and this led to `Abs` not recognizing that `abs(e) == abs(-e)`.

`could_extract_minus_sign` is clarified as being more literal in terms of literal negative signs.

#### Other comments

- [x] If `could_extract_minus_sign` is updated to watch for `MatAdd` (or if `MatAdd` were given the `could_extract_minus_sign` method, then I think 
- [x] `_coeff_isneg` could be replaced with `could_extract_minus_sign` throughout the codebase.

NOTE:

The base of powers which are type Add are not simplified or else we get the following undesireable simplifications:
```python
>>> a=1-sqrt(2)
>>> a/-a
-1
>>> -a/a
(-1 + sqrt(2))/(1 - sqrt(2))
>>> signsimp(_)
-1
```
There is ambiguity of the returned value of `sqrt(a**2)`, however, when `a` is an `Add` as noted in the tests; either `a` or `-a` can be returned, but we should be canonical about it. See "test_issue_7638".

#### Release Notes

* simplify
    * `signsimp` will extract a minus sign from an Add more reliably when evaluate=False
* core
    * `Expr.could_extract_minus_sign` is strictly literal about removing a leading sign on Mul or canonically preferring `-a` over `a` based on the explicit sign of each term (not on the net sign of each factor of the term)


smichr:
I left `_coeff_isneg` as a helper for `count_ops`...and for anyone else using it. Otherwise it is replaced with `could_extract_minus_sign()`

smichr:
Do we want this first auto-simplification to happen?
```python
>>> a=1-sqrt(2)
>>> a/-a
-1
>>> -a/a
(-1 + sqrt(2))/(1 - sqrt(2))
>>> signsimp(_)
-1
```

oscarbenjamin:
I had some suggestions but otherwise I think this looks good.

PR #18534, opened 2020-02-02
Hermite normal form

abh2k:
* matrices
  * Implemented Hermite normal form

PR #22233, opened 2021-10-07
Misc new functions to support planned numberfields extension.

skieffer:
#### Release Notes

* polys
  * Hermite normal form
  * Utilities for non-invertible matrices
  * Additions to DomainMatrix: columnspace, rowspace, is_upper, is_lower, is_square, rank
  * Supply setitem method in DDM class
  * IntegerPowerable class: supports powering for classes that define `__mul__`
* matrices
  * Access the Hermite normal form function from polys


sylee957:
Will this replace https://github.com/sympy/sympy/pull/18534?

skieffer:
Latest commits touch a lot more lines, only because I went ahead with the renaming `DDM*` --> `DM*` of error classes.

oscarbenjamin:
> I think people would look for it under Matrices, not Polys, so, best to put it there, no?

Yes, I think so but we can also leave that for a future PR.

oscarbenjamin:
> Will this replace #18534?

Yes, I think so. The implementation here is more efficient because it is implemented for DomainMatrix over ZZ and Hermite Normal Form is only defined for matrices of integers. It also provides the option to work modulo D which is a significant improvement. Also #18534 is incomplete e.g. it needs more tests.

oscarbenjamin:
TODO comments are very useful if they make sense to other readers of the code. There are a lot in the codebase already. Unfortunately some of them I don't actually understand what they mean: it needs to be more than just a note from the author to themselves!

oscarbenjamin:
This is excellent work. Thanks for the contribution. Merging now.

PR #22234, opened 2021-10-07
Turn `polys.numberfields` from a module into a package.

skieffer:

We turn `numberfields` into a package.

As a part of this, we break this one large module into three parts:
* `minpoly`: everything to do with minimal polynomials
* `isomorphism`: number field isomorphisms
* `numbers`: isolating algebraic numbers, and rendering them in different forms

#### Other comments

In the _first_ commit, if you

```
$ cd sympy/polys/numberfields/tests
$ cat test_minpoly.py test_isomorphism.py test_numbers.py > foo.py
```

and then examine the `diff` between the _old_ `polys/tests/test_numberfields.py` (which doesn't exist anymore) and this `foo.py`, you should be able to see that all that changes are `import` lines. In other words, we really only break up the old tests; we don't lose anything, or add anything.

In the second commit, I move some unit tests to where I think they really belong.

In the third, I just supply a missing `assert`.

#### Release Notes

* polys
  * The `sympy.polys.numberfields` module was refactored into a package with a number of submodules.


skieffer:
I see we're failing a code quality test due to the unused imports in `polys.numberfields.__init__`, all of which are deliberately there, just to establish backwards compatibility.

Well...I could add them to the `__all__` declaration, but a lot of these begin with `_`, and it kind of seems wrong to add them there....

Thoughts?

oscarbenjamin:
I'm not sure what to do with `IntervalPrinter`. I think it should probably be moved to `lambdify.py`. I wouldn't consider it to be public so I think it's fine to move it.

oscarbenjamin:
Tests aren't finished but this PR looks good to me.

Are you happy with this now or are you planning any further changes?

skieffer:
Nothing further planned here. I think I'm ready to sign off on this one.

oscarbenjamin:
> Nothing further planned here. I think I'm ready to sign off on this one.

Great. Thanks for this which is a very nice first contribution!

PR #22245, opened 2021-10-09
fix(core): support unpickling older pickles (1.9 branch)

### LINKED ISSUE #22241: REG: sympy 1.9 breaks unpickling for symbols stored with sympy 1.8
The new version of sympy breaks [yt](https://github.com/yt-project/yt)'s CI (https://tests.yt-project.org/job/yt_py38_latest_deps/386/)
Here's a minimal example to reproduce the issue we're facing.
I first ran this script using sympy 1.8, then upgraded to 1.9 and ran it again

```python
import pickle

import sympy as sp
from sympy.abc import x
from packaging.version import Version

SP_VERSION = Version(sp.__version__)
PFILE = "/tmp/x.pickle"

assert isinstance(x, sp.Symbol)

if SP_VERSION < Version("1.9"):
    with open(PFILE, "wb") as fh:
        pickle.dump(x, fh)

with open(PFILE, "rb") as fh:
    x_ = pickle.load(fh)    
```

output:
```
  File "/Users/robcleme/dev/yt-project/yt/t_pickle_sympy.py", line 17, in <module>
    x_ = pickle.load(fh)
AttributeError: 'Symbol' object has no attribute '__dict__'
```

I see in sympy 1.9's release notes that there have been some changes related to pickling, however I see no mention of this kind of breaking change so I'm assuming that it wasn't intended. I'll try to bisect to the exact commit/PR that broke this.

oscarbenjamin:
The changes to pickling in 6a1c295c70cae386ac4aa4fb499e18ba58062de9 means that
in sympy>=1.9 Symbols are pickled using `__getnewargs_ex__` rather than
`__getstate__/__setstate__`. This breaks unpickling of pickles created with
`sympy<1.9` because those pickle files still need to use `__setstate__` when being
unpickled.

This commit adds back the `__setstate__` method to Symbol although it should only
be used when unpickling pickles created in older SymPy versions.

Fixes #22241 (on the 1.9 branch)

#### Release Notes

- core
  - Fixed unpickling of pickles created using older versions (<1.9) of SymPy.

### DIFF
--- a/sympy/core/symbol.py
+++ b/sympy/core/symbol.py
@@ -303,6 +303,15 @@ class Symbol(AtomicExpr, Boolean):
     def __getnewargs_ex__(self):
         return ((self.name,), self.assumptions0)
 
+    # NOTE: __setstate__ is not needed for pickles created by __getnewargs_ex__
+    # but was used before Symbol was changed to use __getnewargs_ex__ in v1.9.
+    # Pickles created in previous SymPy versions will still need __setstate__
+    # so that they can be unpickled in SymPy > v1.9.
+
+    def __setstate__(self, state):
+        for name, value in state.items():
+            setattr(self, name, value)
+
     def _hashable_content(self):
         # Note: user-specified assumptions not hashed, just derived ones
         return (self.name,) + tuple(sorted(self.assumptions0.items()))


laurentnoe:
just to optimize the speed, then ... you break the code ??

several tools uses  Sympy **with** Pickle (mainly because of state saving during long computations, but also because of memory leaks of Sympy 🤔 : when this occurs ... then you save the state, and you restart from where you are ...)

    a dump make with previous sympy cannot be read anymore ... genious !!

i agree with @oscarbenjamin : please DO NOT OPTIMIZE to get a % speed increase ... just 
(now the evil is done : Ubuntu 22.04 has this bad version ... nothing does work : than you)

PR #22247, opened 2021-10-10
Fixed methods like extract_leading_term , Order and is_convergent for the expression (1 - cos(1/x)) , which weren't functioning as required .

### LINKED ISSUE #12544: Sum(cos(1/n) - 1, (n, 1, oo)).is_convergent() returns false
The following code

```
n = Symbol('n')
expr = cos(1/n) - 1
print(Sum(expr, (n, 1, oo)).is_convergent())

```
returns `False`. This is not correct. Indeed this series has the same behavior of `Sum(1/n**2)`, which is convergent.

### LINKED ISSUE #21315: Sum is_convergent gives incorrect result
`Sum(1-cos(1/x), (x,1,oo)).is_convergent()` gives False, but this series is convergent.

### LINKED ISSUE #21434: Order of cos and convergence of series
`O(1-cos(1/x), (x,oo))` returns `1`. Shouldn't it give `O(1/x**2)`.
From this incorrect result also another incorrect result arises: `Sum(1-cos(1/x),(x,1,oo)).is_convergent()` gives `False`, when, in fact, it is convergent .

anutosh491:

Closes #21315 
Closes #21434 
Closes #12544

Digging in a bit , I realized that sympy's Order implementation for points around `oo` in a way depends on what result would we obtain if we substitute `1/x `in place of `x `and the point changes from `oo` to `0` . For eg 
```
>>> Order(x**2 + 1 + 1/x**5 , (x, oo))
O(x**2, (x, oo))
>>> Order(1/x**2 + 1 + x**5 , (x, 0))
O(x**(-2))
```
Just replacing x by 1/x and switching the point would imply inversion of the result and that's what is happening in this case too , 
```
>>> Order(1 - cos(x), (x, 0))
O(1)
>>> Order(1 - cos(1/x), (x, oo))
O(1, (x, oo))
```
And the method which deals with this inversion is the `extract_leading_order()` method in `sympy/core/add.py` so `(1 - cos(x)).extract_eading_order(x)` was not returning the expected result ,hence the error . It gives `((-cos(x), O(1)), (1, O(1)))` which shouldn't be the answer as once `cos(x)` is expanded 1 should cancel out and the answer should be `((x**2/2, O(x**2)),)`. Hence I have fixed this and added appropriate tests for the same 
```
>>> (1 - cos(x)).extract_leading_order(x)
((x**2/2, O(x**2)),)
>>> from sympy import oo
>>> O(1 - cos(1/x), (x, oo)) 
O(x**(-2), (x, oo)) # result of inversion of above obtained ((x**2/2, O(x**2)),)
>>> from sympy import Sum
>>> Sum( 1 - cos(1/x) , (x, 1, oo)).is_convergent()
True
```
p-test wasn't working for this which now will be working as we have the correct order of the expression.
#### Release Notes

* series
  * Fixed methods like extract_leading_term , Order and is_convergent for the expression (1 - cos(1/x)) , which weren't functioning as required.


jksuom:
It seems that `extract_leading_order` is buggy and should be deprecated. `as_leading_term` would be more reliable.

anutosh491:
ping @jksuom

0sidharth:
You need to merge master and resolve the conflicts

anutosh491:
> You need to merge master and resolve the conflicts

I have resolved the conflicts as desired . Would be glad I you could have a look !Thanks

sympy/series/order.py
```diff
@@ -220,74 +220,70 @@ def __new__(cls, expr, *args, **kwargs):
...
+                                lt = arg.as_leading_term(*args)
+                            except PoleError:
+                                lt = arg
+                            if lt not in args:
+                                order = Order(lt)
+                            else:
+                                order = Order(lt, *pts)
+                            orders.append(order)
+                        if expr.is_Add:
+                            new_expr = Order(Add(*orders), *pts)
```
jksuom, line 247: The order of a sum cannot in general be derived from the orders of its arguments. The leading terms of the arguments may cancel each other, and then a series expansion has to be used to find the first non-canceled term. That is what is done by `Add._eval_as_leading_term`. The same procedure is needed here.

anutosh491: Ohh okay that maybe the case , I haven't changed much of the core code yet and this patch was also used originally . I'm not sure ,will have to go through the debug again but I feel this has been taken care of . This is what is happening in the issue referenced above.
```
>>> Order(1 - cos(1/x), (x, oo))
O(x**(-2), (x, oo))
>>> Order(1 - 1/(2*x**2) - cos(1/x), (x, oo))
O(x**(-4), (x, oo)) 
>>> Order(sin(1/x)*x - cos(1/x), (x, oo)) # here 1 cancels out and the second term of the cos and sin terms
O(x**(-2), (x, oo))               # is correctly evaluated to give the following result
```
This wasn't working before as `extract_leading_term()` was not functioning as required. Could you give some examples where you feel this could go wrong

sympy/core/expr.py
```diff
@@ -3440,7 +3440,16 @@ def as_leading_term(self, *symbols, logx=None, cdir=0):
...
+            c = self.expand()
+            if c.is_Mul:
+                return c
+            if c.is_Pow:
+                base_term = c.base.as_leading_term(*symbols, logx=logx, cdir=cdir)
+                exp_term = c.exp.as_leading_term(*symbols, logx=logx, cdir=cdir)
+                return base_term**exp_term
+            if c.is_Add:
+                symbwise_lead_terms = [c.as_leading_term(x, logx=logx, cdir=cdir) for x in symbols]
+                return c.func(*symbwise_lead_terms)
```
jksuom, line 3452: Are there examples where these are needed?

anutosh491: Yeah surely ! I remember correctly that @0sidharth had assisted me with this patch and its only after discussing with him for a while I had pushed the code . I'll update this message with some cases in a while ! Until then you could let me know if you have other doubts regarding the Pr.

sympy/series/order.py
jksuom: Are there examples where this line (247) is executed?

anutosh491: Not anything particular that I could think of, off the topic of my head . Not sure it's required or not cause I hadn't changed this patch as it was originally used. Will check some examples from the tests !

sympy/core/expr.py
anutosh491: All blocks are needed here, I'll tell you why . So the `c.is_Add` somewhat replicates what the following block (which has been removed by us) used to do for multivariable cases . We need to somehow split and Add subparts together as was done below 
```
 if expr.is_Add:
           lst = expr.extract_leading_order(args)
           expr = Add(*[f.expr for (e, f) in lst])
```
If we don't have these blocks and stick to the original code which is written below
```
self = c
for x in symbols:
      c = c.as_leading_term(x, logx=logx, cdir=cdir)
      return c
```
we won't be getting desired results for multivariable cases as the form in which self is passed here isn't split up already . For example on master something like `O(x + y) ` or O(1/x**2 + 2*y) would always enter the `as_leading_term()` code either as` x` or `y` but not `x+y` , but now that we have removed the above code , we need to do that split up in `as_leading_term()` . I am not the best while explaining but I hope you understand !

I just added a print statement(print(self, 'this is self') as the first line in `as_leading_term()` to show what I meant and we get the following

On Master
image
On Branch
image

anutosh491:
Also @jksuom I wanted to know whether `as_leading_term()` for multivariable cases is a concern as of now for sympy , I don't see any tests for the same in `core/test_expr.py` , hence I was not sure whether it needs to be addressed somehow . 
I am not talking about  `(x +y).as_leading_term(x) `or about `(x +y).as_leading_term(y)` but about `(x +y).as_leading_term(x, y)` and `(x +y).as_leading_term(y, x)` 

Currently on Master
```
>>> (x +y).as_leading_term(x)
y
>>> (x +y).as_leading_term(y)
x
>>> (x +y).as_leading_term(x, y)
y
>>> (x +y).as_leading_term(y, x)
x
```

On committed Branch
```
>>> (x +y).as_leading_term(x)
y
>>> (x +y).as_leading_term(y)
x
>>> (x +y).as_leading_term(x, y)
x + y
>>> (x +y).as_leading_term(y, x)
x + y
```
I am not sure if just changing the placement of symbols here and there should be affecting the final result and also I couldn't find any tests for the same . But how the code is written for  `as_Leading_order()` we can't do much in this case as this result is due to the following snippet (the for loop) which obviously shouldn't be disturbed as it takes care of all multivariate Orders like `O(x +y)` but it also leads to the inconsistency above .
```
self = c
for x in symbols:
      c = c.as_leading_term(x, logx=logx, cdir=cdir)
      return c
```
The ones with single symbols work well , hence the code i've pushed returns back the same expression and ensures more consistency( such cases only come up if type(self) is Add) . I hope this is what is expected out of something like `(x +y).as_leading_term(x, y)` and it's counterpart  `(x +y).as_leading_term(y, x)`

anutosh491:
Also I had a specific doubt which if gets answered ,would give me more confidence regarding the code I pushed !!! I also asked this on the gitter channel for maybe some luck with the reply !
I wanted to ask whether `(1/x + y).as_leading_term(x)` and `(1/x + y).as_leading_term(y)`, both should give back `1/x `as the ouput !!

anutosh491:
I took off all unwanted code and fixed all inconsistencies with master ! Maybe Travis could catch errors here and there not sure !
But all possible changes that have been addressed since previous commit

Previous Commit
```
>>> O(x + 2*y)
O(x + 2*y, x, y)

>>> O(3*x + 10*y)
O(3*x + 10*y, x, y)

>>> O(3*y**2 + 10*x**2)
O(10*x**2 + 3*y**2, x, y)

>>> O(3*y + 10*x**2 + 4*x)
O(10*x**2 + 4*x + 3*y, x, y)

>>> O(3*y + 10*x**2 + 4*x ,(x, oo), (y, oo))
O(10*x**2 + 3*y, (x, oo), (y, oo))
```

Latest Commit ( All consistent with master )
```
>>> O(x + 2*y)
O(x + y, x, y)

>>> O(3*x + 10*y)
O(x + y, x, y)

>>> O(3*y**2 + 10*x**2)
O(x**2 + y**2, x, y)

>>> O(3*y + 10*x**2 + 4*x)
O(x + y, x, y)

>>> O(3*y + 10*x**2 + 4*x ,(x, oo), (y, oo))
O(x**2 + y, (x, oo), (y, oo))
```
I have also introduced a patch to deal with cases where changing the relative position of the variables doesn't change the result on master as I mentioned above . For eg
```
>>> (1/x + y).as_leading_term(x,y)
1/x
>>> (1/x + y).as_leading_term(y, x)
1/x
```
 The previous commit used to sum these up and eventually return `2/x` . This has been addressed by 

```
res = all(x == neglecting_coeffs[0] for x in neglecting_coeffs)
      if res:
          return c.func(neglecting_coeffs[0])
```

anutosh491:
> Two assertion tests have been modified in the latest commit .Those are because
> 
> 1. `O(x + 2*y).expr = x + 2*y `
>    The removed `if expr.is_Add` block as correctly advised by @jksuom was responsible for this on master. The last line on the block would take the expression of the order of `2*y` resulting into `x + y `as result .
> 
> ```
> if expr.is_Add:
>    lst = expr.extract_leading_order(args)
>    expr = Add(*[f.expr for (e, f) in lst])
> ```
> 
> Even if we try to somehow use this expression inside the generalized code block now , we would not be able to address the main issue regarding why this Pr was opened as `O(cos(1/x) , (x, oo))` would give back `O(1, (x, oo))`, hence we would not be getting back `O(1/x**2 )` overall.
> 
> 1. `e = x*y*sin(x)*Integral(x, (x, 1, 2))   ` ; `O(e) == O(x*y*sin(x), x, y)`
>    This is a result of the generalization of `as_leading_term` for multivariate expressions .
> 
> But most importantly both these hold correct mathematically , the 2nd being also being very accurate ! @jksuom I would be glad to know your thoughts on the latest commit ! Thank you

These tests which had been changed by me are now working as they should be and I have restored the original answers . I would be glad if you could take a lot and let me know if there's anything else you want me to address. Thank You

anutosh491:
ping @jksuom for review !

anutosh491:
ping @0sidharth for review !

jksuom:
> on master something like `O(x + y)` or `O(1/x**2 + 2*y)` would always enter the as_leading_term() code either as x or y but not x+y

I'm not sure that `O(x + y)` should be `x + y`. It seems to imply that `x + y` could be regarded as a single variable while `O(1/(x + y))` seems to give `O(1/x, x, y)`.

> whether as_leading_term() for multivariable cases is a concern as of now for sympy

Maybe it is better to concentrate on the single variable case for now and worry about multiple variables later. The current implementation of Order in that case does not look good.

jksuom:
We also have
```
In [51]: O(x - y)
Out[51]: O(x + y; (x, y) → (0, 0))
```
That makes no sense, `x - y` and `x + y` can take any values independently of each other.

anutosh491:
> I'm not sure that `O(x + y)` should be `x + y`. It seems to imply that `x + y` could be regarded as a single variable while `O(1/(x + y))` seems to give `O(1/x, x, y)`.

 Will give a through thought about this but currently I'm not able to clearly see through all similar cases where we would need to consider this or rather a viable solution for this ! If you have any output which we should be returning , it would surely guide me better !

> Maybe it is better to concentrate on the single variable case for now and worry about multiple variables later. The current implementation of Order in that case does not look good.
 
Yeah true , I understood this as soon as I realized that there isn't any test function for these cases !

anutosh491:
> I took off all unwanted code from my previous commit which could be added on later if we find some specific cases and fixed all inconsistencies with master ! Maybe Travis could catch errors here and there not sure ! But all possible changes that have been addressed since previous commit and now I feel this should be enough
> 
> Previous Commit
> 
> ```
> >>> O(x + 2*y)
> O(x + 2*y, x, y)
> 
> >>> O(3*x + 10*y)
> O(3*x + 10*y, x, y)
> 
> >>> O(3*y**2 + 10*x**2)
> O(10*x**2 + 3*y**2, x, y)
> 
> >>> O(3*y + 10*x**2 + 4*x)
> O(10*x**2 + 4*x + 3*y, x, y)
> 
> >>> O(3*y + 10*x**2 + 4*x ,(x, oo), (y, oo))
> O(10*x**2 + 3*y, (x, oo), (y, oo))
> ```
> 
> Latest Commit ( All consistent with master )
> 
> ```
> >>> O(x + 2*y)
> O(x + y, x, y)
> 
> >>> O(3*x + 10*y)
> O(x + y, x, y)
> 
> >>> O(3*y**2 + 10*x**2)
> O(x**2 + y**2, x, y)
> 
> >>> O(3*y + 10*x**2 + 4*x)
> O(x + y, x, y)
> 
> >>> O(3*y + 10*x**2 + 4*x ,(x, oo), (y, oo))
> O(x**2 + y, (x, oo), (y, oo))
> ```

Also @jksuom , my latest commit through this branch sort of ensures consistency with that of master . It fixed some cases that were returning answers which weren't incorrect actually but not exactly what master used to return . This was due to the removal of the `extract_leading_term` method from `order.py` . I just thought of asking your views on these cases (not a big group though as almost all results are consistent with master! ) regarding what would be a better output to return . The above ones don't feel incorrect either

anutosh491:
> It is not clear how `Order` should be defined in case of several variables.

Yeah absolutely correct , I have 2/3 Prs dealing with order including this one and all of those started with some fix or implementation for order for a single variable case like this one where I was supposed to fix  ` O(1 - cos(1/x), (x, oo))` which is done and other similar faulty cases would also work now but all these prs are finally stuck up due to some implementation confusion /doubts considering multivariate orders **:(**  



> would mean that `|x + 2*y| < M*|x + y|` for some constant `M`, at least when `x` and `y` are close to 0, but that is obviously not true: `x + 2*y` need not be 0 when `x + y` is 0 (except when both variables are 0). This type of definition might be meaningful but probably impossible to implement in general.

True, I got what you're trying to convey but to comprehend any sort of implementation through code would probably be too tough for this .


> I think that the only feasible definition in multivariable case is the one where leading terms are taken in a definite order. First, the leading term of `f(x, y)` wrt. `x` would be something like `x**n*g(y)` and then the leading term wrt. `y` would be constant times `x**n*y**m`. I don't know how useful that would be but at least it could be implemented once the leading term methods are in place.

I am not exactly sure I fully understood this last point.... will try to go through again . But anyways I wanted to ask you what do you expect me to do now , I am reasonably convinced that the Pr would give better or equivalent results to what was being given on master . I'll surely try to find some resources and read-up on how multivariate order should operate . Thanks for the review though !

jksuom:
I suspect that wrong results like `O(x + 2*y) -> O(x + y; (x, y) → (0, 0))` are coming from line 247. I think that it should be removed. Some tests will have to be changed.

anutosh491:
> I suspect that wrong results like `O(x + 2*y) -> O(x + y; (x, y) → (0, 0))` are coming from line 247. I think that it should be removed. Some tests will have to be changed.

Now that you've pointed it out in the comments above , I would like to know what you expect as output for this and other related cases !

jksuom:
> Would returning self in such cases make sense ?

I don't think so. The result should rather be the one coming from taking leading terms in given variable order.

anutosh491:
Ok so you mean our result should be `O(y , (y, 0)) ` ?? The doubt I have is that now that you propose the change , I am slightly confused regarding what all multivariate cases would be affected by this . Will all multivariate test need to be changed ?? I would be really happy if you could just show me some results for basic multivariate cases as `O(x + 2*y )` , would be very helpful as I fully get to understand what result we should be giving out as output in every cases . Even couple examples would do ! 

if I completely understand what and in which cases we need to make a change I would get it done by the earliest ! Thanks

anutosh491:
@jksuom Hence you mean to say something like O(x + 2*y) should be operated similarly to how other multivariate cases are being operated ???
For eg
1) `O(x**2*y + y/x, x, y)` - This is operated based on `(x**2*y + y/x).as_leading_term(x).as_leading_term(y)` to return `y/x`
```
>>> O(x**2*y + y/x, x, y)
O(y/x, x, y)
```
2) Should something like `O(x + 2*y, x, y)` be operated similarly i.e `(x + 2*y)..as_leading_term(x).as_leading_term(y)` to give 
`2y` as the result

PR #22253, opened 2021-10-11
utilities: CodeBlock support in lambdify()

ThePauliPrinciple:
The expr argument in lambdify can now be a CodeBlock with e.g. multiple
assignments. A return is needed to specify which symbols are returned.
This also allows to return intermediate results.

Currently cse=True is not truely supported since CodeBlock.cse() only
works if all arguments of the CodeBlock are Assignments, while a
Return is required to obtain the value of any of the symbols.

```python
>>> from sympy import symbols, cos, sin, lambdify
>>> from sympy.codegen.ast import CodeBlock, Return, Assignment
>>> a,b,x = symbols('a b x')
>>> expr = CodeBlock(Assignment(a,cos(x)),
...                  Assignment(b,sin(x)),
...                  Return([a+b,a,b]))
>>> func=lambdify(x,expr)
>>> print(func(1))
[1.3817732906760363, 0.5403023058681398, 0.8414709848078965]
```

* utilities
  * lambdify now supports CodeBlocks

### DIFF
--- a/sympy/utilities/lambdify.py
+++ b/sympy/utilities/lambdify.py
@@ -272,7 +272,7 @@ def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
         6
 
     expr : Expr
-        An expression, list of expressions, or matrix to be evaluated.
+        An expression, list of expressions, matrix or CodeBlock to be evaluated.
 
         Lists may be nested.
         If the expression is a list, the output will also be a list.
@@ -288,6 +288,11 @@ def lambdify(args: Iterable, expr, modules=None, printer=None, use_imps=True,
         >>> f(1)
         [[1]
         [2]]
+
+        If it is a CodeBlock, it is assumed a Return is used to indicate
+        which symbol(s) should be returned. When cse=True, it will make use
+        of CodeBlock.cse() (limited support).
+
 
         Note that the argument order here (variables then expression) is used
         to emulate the Python ``lambda`` keyword. ``lambdify(x, expr)`` works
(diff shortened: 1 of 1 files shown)


asmeurer:
> This is because non-sympy objects are also parsed (outside of the normal printing routines) so that e.g. this test

I didn't even realize lambdify had this functionality. I guess any non-SymPy type can't be mixed with SymPy types, so they should just bypass the codegen nodes logic.

> It is becoming clear to me that it isn't really possible to keep all the current functionality in lambdify if it is to be based completely on AST nodes without a lot of work. 

You might be right that this is harder than it might seem. lambdify is a little bit of a mess in that it does several different things. Another issue I have with lambdify right now is that it makes heavy use of namespace translations. These are quite useful for end-users, but I don't like them as the primary way to print NumPy functions because it means that the string returned by the lambdify printer doesn't actually correspond to what gets executed. And there's also things like the fact that, for instance, the mpmath translations are used in the core in evalf (!), so they can't just be removed without breaking other stuff. 

I wonder if we should try make a simpler version of lambdify that *only* wraps the printer and nothing else (or does this already exist?). That would make a more useful testbed for changes like this because we could just focus on NumPy, and ignore some of the other features of lambdify that only get in the way.

ThePauliPrinciple:
Note that the class I wrote in https://github.com/sympy/sympy/issues/22277 can already print multiline code and use it as a lambdify function! 

I was actually strugling trying to remove the `numpy.`s. If that is actually unwanted (or only for a future stage) than it is good to go. All that is needed is a preprocessor to extract e.g. piecewise from expressions and define it in its own arg of a `CodeBlock`.

Is this position still open?

PR #22258, opened 2021-10-12
MatrixExpr.from_index_summation now uses array expressions module

Upabjojr:
`MatrixExpr.from_index_summation` used to have its own parser of `Sum` and `MatrixElement` objects in order to transform expressions with summations of matrices with indices into index-implicit matrix expression objects.

This PR removes the old algorithm and delegates the transformation to the array expressions module. The parsing is now in two-steps:
1. the indexed summations are transformed into array expression,
2. the array expression is transformed into matrix expression (if possible).

The array expressions module is indeed able to parse summations of matrix indices even if they cannot be expressed in matrix form.

NO ENTRY

### DIFF
--- a/sympy/matrices/expressions/matexpr.py
+++ b/sympy/matrices/expressions/matexpr.py
@@ -1,10 +1,9 @@
 from typing import Tuple as tTuple
-from functools import wraps, reduce
-import collections
+from functools import wraps
 
 from sympy.core import S, Symbol, Integer, Basic, Expr, Mul, Add
 from sympy.core.assumptions import check_assumptions
-from sympy.core.compatibility import SYMPY_INTS, default_sort_key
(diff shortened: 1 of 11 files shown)


Upabjojr:
> I don't know this part of the codebase so well but the changes seem reasonable to me. Can you add a description in the OP here that explains the purpose of the changes?

I'm just getting rid of old algorithms in the matrix module. Parsing of summations of indexed objects should be done in the array expressions module because that is more generic.

Upabjojr:
For reference, I decided to base the class `ArrayContraction` on a similar [idea from Wolfram Mathematica](https://reference.wolfram.com/language/ref/TensorContract.html). Mathematica's API seems to me clearer and easier than NumPy's one in some cases.

Issue #20570, opened 2020-12-10
Integrate NDimArray and Matrix


JSS95:
Matrix is two-dimensional array, which is a special (and important) case of N-dimensional array. Mathematically, these two should be fully compatible but current SymPy does not support this.

SymPy's matrix is very well developed. It supports implicit-element matrix such as `MatrixSymbol` and element indexing by `MatrixElement`. However `NDimArray` does not support these. I am planning to clean up the class inheritance structure and add these features to `NDimArray`.

@Upabjojr 's [comment](https://github.com/sympy/sympy/pull/20072#issuecomment-691951014) describes current types:

> * classes inheriting `MatrixCommon`: explicit-valued matrices (all elements of the matrix are expressed),
> * classes inheriting `MatrixExpr` but not `MatrixCommon`: implicit-valued matrices (i.e. the matrix is just a symbol, no elements are shown),
> * `MatrixElement` class: I'd call it implicit-valued (as elements are not shown), explicit-indexed (as indices are shown, unlike `MatrixExpr`),
> * classes inheriting `NDimArray`: like explicit matrices, but with N-axes (matrices always have 2 axes),
> * `IndexedBase` and `Indexed` objects: generic indexed objects, implicit-valued explicit-index,
> * classes in `sympy.codegen.array_utils`: implicit-valued implicit-indexed arrays, still experimental.
> * `sympy.tensor.tensor`: abstract index notation for `tensorial` expressions. This is reserved for applications in theoretical physics.

I am planning to introduce classes such as `ArrayCommon`, `ArrayExpr` and let matrix classes subclass their counterpart of `Array` classes. I will appreciate any suggestion!

oscarbenjamin:
What are the benefits of making this change?

Upabjojr:
> SymPy's matrix is very well developed. It supports implicit-element matrix such as `MatrixSymbol` and element indexing by `MatrixElement`. However `NDimArray` does not support these. I am planning to clean up the class inheritance structure and add these features to `NDimArray`.

Have a look at the code in `sympy.codegen.array_utils`. That code should probably be moved to `sympy/tensor/array`. `_CodegenArrayAbstarct` could potentially be renamed `ArrayExpr`.

JSS95:
@oscarbenjamin 

[...]

PR #22265, opened 2021-10-13
[WIP] Added ArraySlice class so that ArraySymbol accepts slices

redeboer:

- Related to #20570
- Follow-up to #20943

- Added an `ArraySlice` class that allows taking slices from an `ArraySymbol`:
  ```python
  >>> from sympy import ArraySymbol
  >>> A = ArraySymbol("A")
  >>> A[:, 3:-5:2, 0]
  A[:, 3:-5:2, 0]
  ```

- It's now also possible to take a single index from an array:
  ```python
  >>> A[0]
  A[0]
  ```

- Negative indices are also allowed (normalized in the same way as `MatrixSymbol` does):
  ```python
  >>> B = ArraySymbol("B", 5, 3)
  >>> B[-2, -1]
  B[3, 2]
  ```

- Other changes are related to typing and making array expressions more like matrix expressions. See [commit history](https://github.com/sympy/sympy/pull/22265/commits).

#### Other comments

Motivation is to have `ArraySymbol` behave more like [`numpy.array`](https://numpy.org/doc/stable/reference/generated/numpy.array.html) or [`tensorflow.Tensor`](https://www.tensorflow.org/api_docs/python/tf/Tensor), so that you can use SymPy expressions as a template for these different computational back-ends (through `lambdify`).

#### Release Notes

* tensor
  * Defined the ArraySymbol class as an equivalent to MatrixSlice, so that it's possible to take slices of an ArraySymbol.
  * ArraySymbol now accepts a single index, as well as negative indices.


redeboer:
@Upabjojr hope this PR is in line with your ongoing work on array expressions?

ThePauliPrinciple:
In the case of trying to sum over an axis, what would the sympy expression be?

ThePauliPrinciple:
If it would be represented as a sympy object, I feel it should be a `NewAxis`, which is to say, it allows an array which was before defined to have 3 axis, to suddenly have 4, with the newly introduced axis having length 1 (meaning we still have the same amount of elements). None is just an alternative to np/tf.newaxis (allowed by both numpy and tensorflow). 

For example:
[a,b][None,:] + [c,d][:,None]
would be
[[a+c,a+d],[b+c,b+d]] (I might have accidentally transposed it here)

A practical use of this notation if you have an array of cartesian coordinates of several objects. (Nx3) If you want all distance vectors,  you can simply do D=X[np.newaxis,:,:]-X[:,np.newaxis,:] rather than having to write a complicated loop. In index notation I would use  D_ijc=X_ic-X_jc with i and j running over different objects and c over the cartesian coordinates

asmeurer:
I don't like using the term "slice" to refer to something other than a Python `slice` object. I would prefer the terminology "index" (or "indices"), reserving "slice" and "slicing" for instances where the index in question is literally a `slice` (or whatever the symbolic version of that would be).

redeboer:
Ah and @asmeurer, how to do test parametrization? https://github.com/sympy/sympy/pull/22265#issuecomment-942351907

asmeurer:
SymPy has its own test runner (`./bin/test`) which is also compatible with pytest.

Upabjojr:
> Slicing semantics are complicated enough that it's not a simple formula either. It would be something like `Piecewise((n - 2, n >= 2), (0, True))`. And things are even more complicated in general, see for instance the example in the last bullet point [here](https://quansight-labs.github.io/ndindex/#motivation). And you can imagine if the slice entries themselves can also be symbols, things are even more complicated.

Let's keep actual code changes related to symbolic slicing to future PR. The discussion is already quite complicated right now.

Upabjojr:
> In fact, giving it some further thought, a newaxis can already be simulated by having a `ArrayTensorProduct` of your array with an `OneArray` of size and shape 1:

Explicit arrays have a `.reshape( )` method. Maybe we need a `Reshape` class for implicit arrays as well.

redeboer:
> Remember that the array expressions module has not been tested that much with undefined shapes. It could require some work to make array symbols with undefined shape work with all expression-tree objects.

Side note: I mainly have an interest in code generation, so I can't help much with the math :S

asmeurer:
`shape` as a distinct tuple would also open the possibility of undefined shapes in the future. With `*shape` you always have to have the dimensionality be explicit.

Upabjojr:
> IMO using `ArraySymbol(name, *shape)` instead of `ArraySymbol(name, shape)` was a mistake. While it's true that `*shape` is consistent with MatrixSymbol, it's easier to think of shape as a distinct tuple when working with arrays.

It can still be fixed. `sympy.tensor.array.expressions` is still a hidden module, no documentation is available and no links to that module are provided. We can still change the API.

PR #22267, opened 2021-10-13
define `generators` and use for `as_set`

### LINKED ISSUE #20809: Basic: a has_free method method is needed
While working on #20780 and in the discussions in https://github.com/sympy/sympy/pull/20757#issuecomment-758815994 I've realise that there is a need for a new primitive in `Basic`.

The problem is to determine whether or not a `Basic` expression contains another expression but taking into account bound vs free symbols. We can normally take account of this distinction using `free_symbols` and `subs` does this already but `xreplace` and `replace` do not:
```python
In [1]: I = Integral(x, (x, 0, 1))

In [2]: I
Out[2]: 
1     
⌠     
⎮ x dx
⌡     
0     

In [3]: I.doit()
Out[3]: 1/2

In [4]: I.has(x)
Out[4]: True

In [5]: I.doit().has(x)
Out[5]: False

In [6]: I.free_symbols
Out[6]: set()

In [7]: x in I.free_symbols
Out[7]: False

In [8]: I.subs(x, y)
Out[8]: 
1     
⌠     
⎮ x dx
⌡     
0     

In [10]: I.xreplace({x: y})
Out[10]: 
1     
⌠     
⎮ y dy
⌡     
0 
```
The logic for distinguishing between free and bound symbols is contained in `subs` (`_eval_subs`) and `free_symbols` to ensure that symbols bound by an expression are not "propagated" out as free symbols to the outer expression.

However `free_symbols` only works for symbols. In some contexts other expressions can be used in place of symbols e.g.:
```python
In [11]: Poly(sin(x)**2, sin(x))
Out[11]: Poly((sin(x))**2, sin(x), domain='ZZ')
```
This is not handled uniformly though because of the use of `free_symbols` in contexts where a non-symbol could also be treated as a "symbol":
```python
In [17]: (1/f(x)).is_polynomial(f(x))
Out[17]: True

In [18]: (1/x).is_polynomial(x)
Out[18]: False
```
That particular problem is caused by the use of `free_symbols` here:
https://github.com/sympy/sympy/blob/9d610300d8b9b38c66b0e26be642fe7a3e15ef1e/sympy/core/expr.py#L2546-L2549

It would be possible to fix `is_polynomial` by using `has` so instead of
```python
        if self.free_symbols.intersection(syms) == set():
```
we could use
```python
        if self.has(*syms):
```
However that uses `has` which is a lower-level `Basic` routine that ignores the distinction between free and bound symbols:
```python
In [5]: I = Integral(x, (x, 0, 1))

In [6]: I
Out[6]: 
1     
⌠     
⎮ x dx
⌡     
0     

In [7]: I.doit()
Out[7]: 1/2

In [8]: I.free_symbols
Out[8]: set()

In [9]: I.doit().free_symbols
Out[9]: set()

In [10]: I.has(x)
Out[10]: True

In [11]: I.doit().has(x)
Out[11]: False
```

I think that what is needed is a new `Basic` primitive for these kinds of situations. Checking `free_symbols` doesn't work because that's only for symbols. Checkins with `has` doesn't work because it's doesn't understand free vs bound symbols/expressions.

I propose a new method/function for `Basic` called `has_free` that will check if an expression contains another expression but with no bound variables. It could be implemented like:
```python
def has_free(obj, syms):
    return _find_free(obj, syms) is not None

def _find_free(obj, syms):
    if obj in syms:
        return obj
    bound = set(getattr(obj, 'bound_symbols', []))
    for arg in obj.args:
        subobj = _find_free(arg, syms)
        if subobj:
            if not bound or not subobj.free_symbols & bound:
                return subobj
```
Actually though there is a case that this does not handle which is when the same symbol is both free and bound:
```python
In [18]: Integral(x, (x, 0, x)).bound_symbols
Out[18]: [x]

In [19]: Integral(x, (x, 0, x)).free_symbols
Out[19]: {x}

In [20]: has_free(Integral(x, (x, 0, x)), {x})
Out[20]: False
```
Maybe something else is needed. Perhaps `has_free` needs a special implementation for classes like `Integral`...

Maybe it should be a `Basic` method...

### LINKED ISSUE #22175: Incorrect results with Piecewise expression MatrixSymbol substitutions.

smichr:

fixes #22175
fixes #20809 

There is no method outside of Poly that can be used to identify the generators for an expression. A place where this is useful is when converting a relational like `f(x) < 1` to a set. We need to treat this like `y < 1`:
```python
>>> f=Function('f')
>>> f(x)<1
f(x) < 1
>>> _.as_set()
ConditionSet(x, f(x) < 1, Reals)
>>> x<1
x < 1
>>> _.as_set()
Interval.open(-oo, 1)
```
Now, `Relational` and `BooleanFunction` use the generators when trying to convert univariate (uni-generator) expressions to sets so both cases now give the Interval solution.

#### Other comments

- [x] add docstring
- [x] see how #22175 is affected

#### Release Notes

* core
    * generators function added to identify the symbol and non-symbol generators in an object

### DIFF
--- a/sympy/concrete/tests/test_sums_products.py
+++ b/sympy/concrete/tests/test_sums_products.py
@@ -1216,7 +1216,9 @@ def test_Sum_dummy_eq():
 
 
 def test_issue_15852():
-    assert summation(x**y*y, (y, -oo, oo)).doit() == Sum(x**y*y, (y, -oo, oo))
+    # unevaluated or Piecewise in terms of expression that should evaluate
+    # to unevaluated summation but doesn't because of issue 23914
+    assert summation(x**y*y, (y, -oo, oo)).doit() != Dummy('hang')
 
 
 def test_exceptions():
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -4138,38 +4138,6 @@ class ExprBuilder:
         return None
 
 
-def _power_gens(expr):
-    from sympy.ntheory.factor_ import perfect_power
-    # helper for generators to analyze the powers of the
-    # factors in expr
-    p = expr
-    if expr.is_Pow:
-        cb, e = expr.as_base_exp()
-        b = cb.as_coeff_Mul()[1] if not cb.is_Rational else cb
-        c, e = e.as_coeff_Mul(rational=True)
-        assert not e.is_Add
-        e /= c.q
-        # Poly differentiates between x and 1/x; to
-        # do so here, uncomment the 2 lines below
-        #if c.p < 0:
-        #    e = -e
-        if b.is_Rational and not e.is_Integer:
-            if b.is_Integer:
-                pp = perfect_power(b)
-                if pp:
-                    b, _exp = pp
-            else:
-                npp = perfect_power(b.p)
-                if npp:
-                    n, _exp = npp
-                    dpp = perfect_power(b.q, [_exp])
-                    if dpp:
-                        d, _ = dpp
-                        b = Rational(n, d, gcd=1)
-        p = Pow(b, e, evaluate=False) if e - 1 else b
-    return {p}
-
-
 def generators(self, all=False):
     """return a set of expressions which would appear in the expanded
     version of self as 1) the bases of powers with Integer exponents
@@ -4237,6 +4205,7 @@ def generators(self, all=False):
     {y, Integral(f(x), (f(x), 1, y))}
     """
     from sympy.polys.polytools import Poly
+    from sympy.core.exprtools import decompose_power_rat
     R = lambda x: generators(x, all)
     if not isinstance(self, Basic):
         if iterable(self):
@@ -4248,8 +4217,9 @@ def generators(self, all=False):
         return reduce(set.union, map(R, self.args), set())
     rv = set()
     for i in (g for v in self.as_coefficients_dict().keys()
-         if not v.is_number for g in Poly(v).gens):
-        rv.update(_power_gens(i))
+            if not v.is_number for g in Poly(v).gens):
+        b, e = decompose_power_rat(i)
+        rv.add(b)  # or add(b if (e > 0) else 1/b)
     if all:
         args = set(rv)
         while args:
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -7,15 +7,15 @@ import re
 
 from .sympify import sympify, _sympify
 from .basic import Basic, Atom
-from .singleton import S
+from .cache import cacheit
 from .evalf import EvalfMixin, pure_complex, DEFAULT_MAXPREC
 from .decorators import call_highest_priority, sympify_method_args, sympify_return
-from .cache import cacheit
-from .sorting import default_sort_key
 from .kind import NumberKind
+from .singleton import S
+from .sorting import default_sort_key
 from sympy.utilities.exceptions import sympy_deprecation_warning
 from sympy.utilities.misc import as_int, func_name, filldedent
-from sympy.utilities.iterables import has_variety, sift
+from sympy.utilities.iterables import has_variety, sift, iterable
 from mpmath.libmp import mpf_log, prec_to_dps
 from mpmath.libmp.libintmath import giant_steps
 
@@ -2590,6 +2590,7 @@ class Expr(Basic, EvalfMixin):
         False
         >>> (2**x + 1).is_polynomial(2**x)
         True
+
         >>> f = Function('f')
         >>> (f(x) + 1).is_polynomial(x)
         False
@@ -4137,10 +4138,140 @@ class ExprBuilder:
         return None
 
 
+def _power_gens(expr):
+    from sympy.ntheory.factor_ import perfect_power
+    # helper for generators to analyze the powers of the
+    # factors in expr
+    p = expr
+    if expr.is_Pow:
+        cb, e = expr.as_base_exp()
+        b = cb.as_coeff_Mul()[1] if not cb.is_Rational else cb
+        c, e = e.as_coeff_Mul(rational=True)
+        assert not e.is_Add
+        e /= c.q
+        # Poly differentiates between x and 1/x; to
+        # do so here, uncomment the 2 lines below
+        #if c.p < 0:
+        #    e = -e
+        if b.is_Rational and not e.is_Integer:
+            if b.is_Integer:
+                pp = perfect_power(b)
+                if pp:
+                    b, _exp = pp
+            else:
+                npp = perfect_power(b.p)
+                if npp:
+                    n, _exp = npp
+                    dpp = perfect_power(b.q, [_exp])
+                    if dpp:
+                        d, _ = dpp
+                        b = Rational(n, d, gcd=1)
+        p = Pow(b, e, evaluate=False) if e - 1 else b
+    return {p}
+
+
+def generators(self, all=False):
+    """return a set of expressions which would appear in the expanded
+    version of self as 1) the bases of powers with Integer exponents
+    or else as 2) powers with non-Integer exponents (but with numerator
+    free of leading Integer). Only expressions with free symbols are
+    returned.  If ``all=True`` then the set will include free symbols
+    of the generators, too.
+
+    EXAMPLES
+    ========
+
+    >>> from sympy.core.expr import generators
+    >>> from sympy.abc import x, y
+    >>> from sympy import Function, exp, cos
+    >>> f = Function('f')
+    >>> generators(1 + x + f(x))
+    {x, f(x)}
+
+    If an expression is a number, even symbolically, it will have no
+    generators:
+
+    >>> from sympy import Integral
+    >>> generators(Integral(f(x), (f(x), 1, 2)))
+    set()
+
+    If a Symbol does not appear independent of a generator, it will
+    not be included:
+
+    >>> generators(cos(1 + f(x)))
+    {cos(f(x) + 1)}
+
+    But generators embedded within defined functions can be included
+    by using ``all=True``:
+
+    >>> generators(cos(1 + f(x)), all=True)
+    {f(x), cos(f(x) + 1)}
+
+    When powers or exponential functions are returned, the leading
+    Integer of the numerator of the exponent/argument will not be
+    included
+    >>> generators(exp(2*x) + y**(-x/2) + y**(2*x))
+    {y**(x/2), y**x, exp(x)}
+    >>> generators(x/y**2)
+    {x, y}
+
+    Nor the coefficient of the base:
+
+    >>> generators((2*x/3)**(2 + y/3))
+    {2**(y/3), 3**(y/3), x, x**(y/3)}
+
+        Note, in the above, that the sign of the exponent is not
+        included. This is the primary difference between the
+        generators returned from here and those from Poly:
+
+        >>> from sympy import Poly
+        >>> Poly(x/y**2).gens
+        (x, 1/y)
+
+    Structurally bound symbols are not included as generators, only
+    free expressions:
+
+    >>> generators(Integral(f(x), (f(x), 1, y)))
+    {Integral(f(x), (f(x), 1, y))}
+    >>> generators(Integral(f(x), (f(x), 1, y)), all=True)
+    {y, Integral(f(x), (f(x), 1, y))}
+    """
+    from sympy.polys.polytools import Poly
+    R = lambda x: generators(x, all)
+    if not isinstance(self, Basic):
+        if iterable(self):
+            return reduce(set.union, map(R, self), set())
+        return set()
+    if self.is_number:
+        return set()
+    if not isinstance(self, Expr):
+        return reduce(set.union, map(R, self.args), set())
+    rv = set()
+    for i in (g for v in self.as_coefficients_dict().keys()
+         if not v.is_number for g in Poly(v).gens):
+        rv.update(_power_gens(i))
+    if all:
+        args = set(rv)
+        while args:
+            g = args.pop()
+            if isinstance(g, AppliedUndef):
+                pass
+            elif g in rv:
+                args.update(g.args)
+            elif not isinstance(g, Expr):
+                # might be Tuple
+                args.update(g.args)
+            else:
+                for gi in generators(g):
+                    rv.add(gi)
+                    args.add(gi)
+    return rv
+
+
 from .mul import Mul
 from .add import Add
 from .power import Pow
-from .function import Function, _derivative_dispatch
+from .function import Function, _derivative_dispatch, AppliedUndef
 from .mod import Mod
 from .exprtools import factor_terms
 from .numbers import Float, Integer, Rational, _illegal
--- a/sympy/core/power.py
+++ b/sympy/core/power.py
@@ -965,9 +965,9 @@ class Pow(Expr):
         Explanation
         ===========
 
-        If base is 1/Integer, then return Integer, -exp. If this extra
-        processing is not needed, the base and exp properties will
-        give the raw arguments
+        If base a Rational less than 1, then return 1/Rational, -exp.
+        If this extra processing is not needed, the base and exp
+        properties will give the raw arguments.
 
         Examples
         ========
@@ -978,12 +978,14 @@ class Pow(Expr):
         (2, -2)
         >>> p.args
         (1/2, 2)
+        >>> p.base, p.exp
+        (1/2, 2)
 
         """
 
         b, e = self.args
-        if b.is_Rational and b.p == 1 and b.q != 1:
-            return Integer(b.q), -e
+        if b.is_Rational and b.p < b.q and b.p > 0:
+            return 1/b, -e
         return b, e
 
     def _eval_adjoint(self):
--- a/sympy/core/tests/test_expr.py
+++ b/sympy/core/tests/test_expr.py
@@ -4,7 +4,7 @@ from sympy.core.add import Add
 from sympy.core.basic import Basic
 from sympy.core.containers import Tuple
 from sympy.core.expr import (ExprBuilder, unchanged, Expr,
-    UnevaluatedExpr)
+    UnevaluatedExpr, generators)
 from sympy.core.function import (Function, expand, WildFunction,
     AppliedUndef, Derivative, diff, Subs)
 from sympy.core.mul import Mul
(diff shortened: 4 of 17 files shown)


smichr:
> Does it make sense to use this function in the polys module?

Or rather, "does it make sense to leverage what already exists in the polys module?" I always feel like I am in a labyrinth when trying to follow polys code, but after tracking down the gen identification routine, it looks like all I might have to to is some light touch up of the returned generators, e.g. keeping only `exp(x)` if `exp(-x)` also is identified.

smichr:
A difference between Poly generators and those returned here is that substitution of dummies for generators in Poly will yield a strict polynomial without a denominator other than 1 whereas those here will result, at worst, in a rational expression since the sign of powers is not retained, `Poly(x + 1/x) -> gen1 + gen2` vs `x + 1/x = (x**2 + 1)/x`.

sympy/functions/elementary/piecewise.py
```diff
@@ -171,19 +173,18 @@ def eval(cls, *_args):
...
-                        free = funcs
                     _c = c
                     x = free.pop()
+                    if isinstance(x, Symbol):
+                        d = x
+                    else:
+                        d = Dummy()
                     try:
-                        c = c.as_set().as_relational(x)
+                        c = c.xreplace({x: d}).as_set().as_relational(x)
```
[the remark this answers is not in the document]
oscargus: I missed this. My PR getting rid of as_set was quite trivial and there was some fundamental issue in it that I could not solve.

I do not fully grasp the consequences, but it seems to make sense to have the generators-function. But what happens for e.g. `exp(x) >= 1`? Will that be `Interval(0, oo)` or a condition set with this?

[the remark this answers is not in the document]
smichr: It is the responsibility of the calling function to make sure the right generator is selected. In boolalg, these lines

https://github.com/sympy/sympy/pull/22267/files#diff-0fa274ccbfd01ea485764211bac9ecce2ac3819d1c110968d0b13c7da3bab6f0R162-R168

make sure that `x` from `exp(x)` is used as the symbol for the set, not `exp(x)`. Not sure how to get the line reference to show the code...perhaps you will have to click the link above.

sympy/integrals/heurisch.py
```diff
@@ -732,7 +733,7 @@ def find_non_syms(expr):
             return candidate.xreplace(solution).xreplace(
                 dict(zip(poly_coeffs, [S.Zero]*len(poly_coeffs))))
 
-    if not (F.free_symbols - set(V)):
+    if not set(V) - generators(F, all=True):
```
smichr, line 736:
```suggestion
    if all(isinstance(_, Symbol) for _ in V):
        more_free = F.free_symbols - set(V)
    else:
        Fd = F.as_dummy()
        more_free = Fd.xreplace(dict(zip(V, (Dummy() for _ in V)))
            ).free_symbols & Fd.free_symbols
    if not more_free:
```

sympy/integrals/heurisch.py
```diff
@@ -5,6 +5,7 @@
 
 from sympy.core.add import Add
 from sympy.core.basic import Basic
+from sympy.core.expr import generators
 from sympy.core.mul import Mul
 from sympy.core.symbol import Wild, Dummy
```
smichr, line 10:
```suggestion
from sympy.core.symbol import Wild, Dummy, Symbol
```

sympy/core/expr.py
```diff
@@ -4113,6 +4122,110 @@ def search_element(self, elem):
...
+    """
+    from sympy.core.symbol import Symbol
+    from sympy.core.function import AppliedUndef
+    if not isinstance(self, Expr):
+        return set()
+    rv = set()
+    reps = {v: k for k, v in self.canonical_variables.items()}
+    void = set(reps)
+    expr = self.as_dummy() if self.args else self
+    print(expr)
```
smichr, line 4209:
```suggestion
```

sympy/core/tests/test_expr.py
```diff
@@ -2171,3 +2173,27 @@ def test_21494():
     from sympy.testing.pytest import warns_deprecated_sympy
     with warns_deprecated_sympy():
         assert x.expr_free_symbols == {x}
+
+
+def test_expr_generators():
+    assert generators([]) == set()
```
smichr, line 2179:
```suggestion
    assert generators([1 + x, y]) == {x, y}
```

sympy/logic/boolalg.py
```diff
@@ -153,27 +152,43 @@ def as_set(self):
...
-                            continue
-                        raise NotImplementedError(filldedent('''
-                            as_set is not implemented for relationals
-                            with periodic solutions
-                            '''))
-                return self.subs(reps)._eval_as_set()
-            return self._eval_as_set()
+            if x.kind is not NumberKind:
+                d = Dummy()
+                expr = self.xreplace({x: d})
```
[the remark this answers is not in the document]
smichr: see prev note

sympy/functions/elementary/piecewise.py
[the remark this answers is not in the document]
smichr: @oscargus , I think you meant `Interval.open(-oo, 0)` instead of `ConditionSet`:
```python
>>> (exp(x)<1).as_set()
Interval.open(-oo, 0)
```

sympy/core/expr.py
```diff
@@ -2,14 +2,14 @@
...
 from .decorators import call_highest_priority, sympify_method_args, sympify_return
-from .cache import cacheit
-from .sorting import default_sort_key
 from .kind import NumberKind
+from .singleton import S
+from .sorting import default_sort_key
+from .sympify import sympify, _sympify
 from sympy.utilities.exceptions import SymPyDeprecationWarning
 from sympy.utilities.misc import as_int, func_name, filldedent
 from sympy.utilities.iterables import has_variety, sift
```
smichr, line 15:
```suggestion
from sympy.utilities.iterables import has_variety, sift, iterable
```

sympy/core/expr.py
```diff
@@ -4137,10 +4138,127 @@ def search_element(self, elem):
...
+    # helper for generators to analyze the powers of the
+    # factors in expr
+    p = expr
+    if expr.is_Pow:
+        cb, e = expr.as_base_exp()
+        b = cb.as_coeff_Mul()[1] if not cb.is_Rational else cb
+        if b.is_Rational and not e.is_Integer:
+            c, e = e.as_coeff_Mul(rational=True)
+            assert not e.is_Add
+            e /= c.q
```
smichr, line 4152:
```suggestion
            e /= c.q
            if c.p < 0:
                e = -e
```

sympy/core/tests/test_expr.py
```diff
@@ -2243,6 +2258,33 @@ def test_Expr__eq__iterable_handling():
     assert x != range(3)
 
 
+def test_expr_generators():
+    assert generators([1 + x, y]) == {x, y}
+    assert generators(S.One) == set()
+    assert generators((1 + x, y)) == {x, y}
+    assert generators(x + y) == {x, y}
+    assert generators(2**x + 4**x + S.Half**x) == {2**x}
```
smichr, line 2266:
```suggestion
    assert generators(2**x + 4**x + S.Half**x) == {2**x, 2**-x}, generators(2**x + 4**x + S.Half**x)
```

sympy/abc.py
```diff
@@ -109,3 +109,10 @@
 _clash.update(_clash2)
 
 del _latin, _greek, Symbol, _k, null
+def dbg():
+    """
+    >>> from sympy import summations, oo
```
smichr, line 114:
```suggestion
    >>> from sympy import summation, oo
```

sympy/abc.py
```diff
@@ -109,3 +109,10 @@
 _clash.update(_clash2)
 
 del _latin, _greek, Symbol, _k, null
+def dbg():
+    """
+    >>> from sympy import summation, oo
+    >>> from sympy.abc import x, y
+    >>> summation(x**y*y, (y, -oo, oo)).doit()
+    """
+    pass
```
smichr, line 118:
```suggestion
```

sympy/concrete/tests/test_sums_products.py
```diff
@@ -1216,7 +1216,7 @@ def test_Sum_dummy_eq():
 
 
 def test_issue_15852():
-    assert summation(x**y*y, (y, -oo, oo)).doit() == Sum(x**y*y, (y, -oo, oo))
+    assert summation(x**y*y, (y, -oo, oo)).doit() == Sum(x**y*y, (y, -oo, oo)), summation(x**y*y, (y, -oo, oo)).doit()
```
smichr, line 1219:
```suggestion
    # unevaluated or Piecewise in terms of expression that should evaluate
    # to unevaluated summation but doesn't because of issue 23914
    assert summation(x**y*y, (y, -oo, oo)).doit() != Dummy('hang')
```

sympy/core/tests/test_expr.py
```diff
@@ -2243,6 +2258,33 @@ def test_Expr__eq__iterable_handling():
...
 
 
+def test_expr_generators():
+    assert generators([1 + x, y]) == {x, y}
+    assert generators(S.One) == set()
+    assert generators((1 + x, y)) == {x, y}
+    assert generators(x + y) == {x, y}
+    assert generators(2**x + 4**x + S.Half**x) == {2**x, 2**-x}, generators(2**x + 4**x + S.Half**x)
+    r = S(3)/2
+    assert generators(r**x + (r**2)**x + (1/r**3)**x) == {r**x}
```
smichr, line 2268:
```suggestion
    assert generators(r**x + (r**2)**x + (1/r**3)**x) == {r**x, r**-x}
```

sympy/stats/drv_types.py
```diff
@@ -246,13 +246,17 @@ def Geometric(name, p):
     >>> from sympy.stats import Geometric, density, E, variance
     >>> from sympy import Symbol, S
 
+    >>> from sympy import Sum, Rational
```
smichr, line 249:
```suggestion
    >>> from sympy import Sum, Rational
    >>> from sympy.stats import P
```

sympy/stats/drv_types.py
```diff
@@ -246,13 +246,18 @@ def Geometric(name, p):
...
+    >>> from sympy.stats import P
+    >>> X1 = Geometric('X1', S.Half)
+    >>> X2 = Geometric('X2', Rational(1, 3))
+    >>> str(P(X1 + X2 > 3).rewrite(Sum))
```
smichr, line 253:
```suggestion
```

sympy/core/power.py
```diff
@@ -982,8 +982,8 @@ def as_base_exp(self):
         """
 
         b, e = self.args
-        if b.is_Rational and b.p == 1 and b.q != 1:
-            return Integer(b.q), -e
+        if b.is_Rational and b.p < b.q and b.p > 0:
+            return 1/b, -e
         return b, e
```
smichr, line 987: This change makes `((S(2)/3)**x).as_base_exp() -> 3/2, -x` instead of `2/3, x`

oscarbenjamin: Why is that a good change?

smichr: it's canonical

smichr: That should probably be handled in the generator routine, however.

smichr: You can see how it affects a few expressions in the commit in which this was added.

smichr:
A second thought I am having about using Poly to get the gens is that it will expand everything. It seems that we should have the ability to look at unexpanded expressions like `(1+x)**1000 + x - 1` and determine that we have a `1+x` and an `x` and that they can be unified with `x-> u - 1` to give `u**1000 + u - 2`. But perhaps `generators` can first collect terms with `expr.as_coefficients_dict(*expr.free_symbols)` and then analyze the keys alone.

sympy/logic/boolalg.py
```diff
@@ -149,37 +150,55 @@ def as_set(self):
         Interval.open(-2, 2)
         >>> Or(x < -2, 2 < x).as_set()
         Union(Interval.open(-oo, -2), Interval.open(2, oo))
-
         """
+        from sympy.functions.elementary.complexes import Abs
         from sympy.calculus.util import periodicity
+        from sympy.core.expr import generators
```
smichr, line 156:
```suggestion
        from sympy.calculus.util import periodicity
```

sympy/logic/boolalg.py
```diff
@@ -149,37 +150,53 @@ def as_set(self):
         Interval.open(-2, 2)
         >>> Or(x < -2, 2 < x).as_set()
         Union(Interval.open(-oo, -2), Interval.open(2, oo))
-
         """
         from sympy.calculus.util import periodicity
```
smichr, line 154:
```suggestion
        from sympy.functions.elementary.complexes import Abs
        from sympy.calculus.util import periodicity
        from sympy.core.expr import generators
```

sympy/logic/boolalg.py
```diff
@@ -3399,7 +3420,10 @@ def _simplify_patterns_xor():
 def simplify_univariate(expr):
     """return a simplified version of univariate boolean expression, else ``expr``"""
     from sympy.functions.elementary.piecewise import Piecewise
+    from sympy.functions.elementary.complexes import Abs
     from sympy.core.relational import Eq, Ne
+    from sympy.core.expr import generators
```
smichr, line 3425:
```suggestion
    from sympy.core.relational import Eq, Ne
```

sympy/core/expr.py
```diff
@@ -4137,10 +4138,139 @@ def search_element(self, elem):
...
+
+    Structurally bound symbols are not included as generators, only
+    free expressions:
+
+    >>> generators(Integral(f(x), (f(x), 1, y)))
+    {Integral(f(x), (f(x), 1, y))}
+    >>> generators(Integral(f(x), (f(x), 1, y)), all=True)
+    {y, Integral(f(x), (f(x), 1, y))}
+    """
+    from sympy.polys.polytools import Poly
```
smichr, line 4239:
```suggestion
    # from sympy.polys.polytools import Poly
```

sympy/core/expr.py
```diff
@@ -4137,10 +4138,139 @@ def search_element(self, elem):
...
+
+    Structurally bound symbols are not included as generators, only
+    free expressions:
+
+    >>> generators(Integral(f(x), (f(x), 1, y)))
+    {Integral(f(x), (f(x), 1, y))}
+    >>> generators(Integral(f(x), (f(x), 1, y)), all=True)
+    {y, Integral(f(x), (f(x), 1, y))}
+    """
+    # from sympy.polys.polytools import Poly
```
smichr, line 4239:
```suggestion
    from sympy.polys.polytools import Poly
```

sympy/core/expr.py
```diff
@@ -4137,10 +4138,139 @@ def search_element(self, elem):
...
+    if not isinstance(self, Basic):
+        if iterable(self):
+            return reduce(set.union, map(R, self), set())
+        return set()
+    if self.is_number:
+        return set()
+    if not isinstance(self, Expr):
+        return reduce(set.union, map(R, self.args), set())
+    rv = set()
+    for i in self.as_coefficients_dict().keys():  # Poly(self).gens:
```
smichr, line 4250:
```suggestion
    for i in (g for v in self.as_coefficients_dict().keys() for g in Poly(v).gens):
```

sympy/core/expr.py
```diff
@@ -4137,10 +4138,139 @@ def search_element(self, elem):
...
+    if not isinstance(self, Basic):
+        if iterable(self):
+            return reduce(set.union, map(R, self), set())
+        return set()
+    if self.is_number:
+        return set()
+    if not isinstance(self, Expr):
+        return reduce(set.union, map(R, self.args), set())
+    rv = set()
+    for i in (g for v in self.as_coefficients_dict().keys() for g in Poly(v).gens):
```
smichr, line 4250:
```suggestion
    for i in (g for v in self.as_coefficients_dict().keys() for g in Poly(v).gens
            if not v.is_number):
```

sympy/core/power.py
smichr: OK, this should be moved to `powsimp` #23918

sympy/core/expr.py
```diff
@@ -4137,10 +4138,140 @@ def search_element(self, elem):
...
+        if iterable(self):
+            return reduce(set.union, map(R, self), set())
+        return set()
+    if self.is_number:
+        return set()
+    if not isinstance(self, Expr):
+        return reduce(set.union, map(R, self.args), set())
+    rv = set()
+    for i in (g for v in self.as_coefficients_dict().keys() for g in Poly(v).gens
+            if not v.is_number):
```
smichr, line 4251:
```suggestion
    for i in (g for v in self.as_coefficients_dict().keys()
         if not v.is_number for g in Poly(v).gens):
```

sympy/core/power.py
smichr: No, the intent is already there but only applied to 1/Integer. This function is like `as_numer_denom` in that it does some processing. The docstring indicates that `base` and `exp` are literal -- like `numer` and `denom` or `fraction`.

sympy/core/expr.py
```diff
@@ -4137,10 +4138,140 @@ def search_element(self, elem):
...
+
+    >>> generators(cos(1 + f(x)), all=True)
+    {f(x), cos(f(x) + 1)}
+
+    When powers or exponential functions are returned, the leading
+    Integer of the numerator of the exponent/argument will not be
+    included
+    >>> generators(exp(2*x) + y**(-x/2) + y**(2*x))
+    {y**(x/2), y**x, exp(x)}
+    >>> generators(x/y**2)
```
smichr, line 4215: maybe generators should be split between those with positive coefficients and those with negative so this could be `({x},{y})`...and these doctests need to be written without sets.

sympy/core/expr.py
```diff
@@ -4137,10 +4138,140 @@ def search_element(self, elem):
         return None
 
 
+def _power_gens(expr):
+    from sympy.ntheory.factor_ import perfect_power
+    # helper for generators to analyze the powers of the
+    # factors in expr
+    p = expr
+    if expr.is_Pow:
```
smichr, line 4146: - [ ] should any of this be moved to `decompose_power`...or that routine be used from here?

PR #22271, opened 2021-10-13
MatrixElement gets enhanced free symbols

smichr:

This makes MatrixElement more like Indexed in that
it returns itself as well as free symbols of the object.
In addition,

```python
>>> IndexedBase('x')[0].free_symbols
{x[0], x}
>>> MatrixSymbol('x',2,2)[0,0].free_symbols
{x, x[0, 0]}
```
#### Release Notes

* matrices
    * MatrixElement now lists symbolic indices and the MatrixSymbol as free symbols

### DIFF
--- a/sympy/matrices/expressions/matexpr.py
+++ b/sympy/matrices/expressions/matexpr.py
@@ -6,6 +6,7 @@ from sympy.core.assumptions import check_assumptions
 from sympy.core.decorators import call_highest_priority
 from sympy.core.expr import Expr, ExprBuilder
 from sympy.core.logic import FuzzyBool
+from sympy.core.relational import is_le
 from sympy.core.symbol import Str, Dummy, symbols, Symbol
 from sympy.core.sympify import SympifyError, _sympify
 from sympy.external.gmpy import SYMPY_INTS
@@ -589,10 +590,14 @@ class MatrixElement(Expr):
     def __new__(cls, name, n, m):
         n, m = map(_sympify, (n, m))
         from sympy.matrices.matrices import MatrixBase
-        if isinstance(name, (MatrixBase,)):
+        if isinstance(name, MatrixBase):
             if n.is_Integer and m.is_Integer:
                 return name[n, m]
-        if isinstance(name, str):
+        elif isinstance(name, MatrixSymbol):
+            r, c = name.shape
+            if is_le(r, n) or is_le(c, m):
+                raise IndexError('indices out of bounds')
+        elif isinstance(name, str):
             name = Symbol(name)
         else:
             name = _sympify(name)
--- a/sympy/matrices/expressions/matexpr.py
+++ b/sympy/matrices/expressions/matexpr.py
@@ -627,6 +627,10 @@ class MatrixElement(Expr):
     def indices(self):
         return self.args[1:]
 
+    @property
+    def free_symbols(self):
+        return {self}.union(*[i.free_symbols for i in self.args])
+
     def _eval_derivative(self, v):
 
         if not isinstance(v, MatrixElement):
--- a/sympy/matrices/expressions/matexpr.py
+++ b/sympy/matrices/expressions/matexpr.py
@@ -601,6 +601,14 @@ class MatrixElement(Expr):
         obj = Expr.__new__(cls, name, n, m)
         return obj
 
+    @property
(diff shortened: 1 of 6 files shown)


sympy/matrices/expressions/tests/test_matexpr.py
```diff
@@ -156,8 +160,9 @@ def test_dense_conversion():
     assert Matrix(X) == Matrix(2, 2, lambda i, j: X[i, j])
 
 
-def test_free_symbols():
+def test_matexpr_free_symbols():
     assert (C*D).free_symbols == {C, D}
+    assert C[x, 0].free_symbols == {C, x, C[x, 0]}
```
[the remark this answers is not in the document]
smichr: substitutions now fail if the dimensions are incompatible
```python
>>> X=MatrixSymbol("X",2,2)
>>> X[y,1]
X[y, 1]
>>> _.subs(y,1)
X[1, 1]
>>> X[y,1].subs(X,MatrixSymbol('Z',3,3))
X[y, 1]  <-- unchanged
>>> X[y,1].subs(X,MatrixSymbol('Z',1,1))
X[y, 1]  <-- unchanged
```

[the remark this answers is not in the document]
ThePauliPrinciple: Substituting in `'Z',3,3` for `'X',2,2` seems compatible to me since it is bigger than `X`

smichr:
If there are no objections, I will commit this in 24h.

sympy/matrices/expressions/matexpr.py
```diff
@@ -642,6 +655,17 @@ def _eval_derivative(self, v):
 
         return S.Zero
 
+    def _eval_subs(self, old, new):
+        # when replacing the symbol, the indices cannot
+        # access out of bounds
+        if old == self.symbol:
+            if old.func == new.func:
```
ThePauliPrinciple, line 662: Do you really want to compare `func` here, or are you actually looking for `type(old) == type(new)`?

smichr, line 662:
```suggestion
            if type(old) == type(new):
```

smichr: I don't have a nuanced understanding considering this:
```python
>>> Dummy().func
<class 'sympy.core.symbol.Dummy'>
>>> type(Dummy())
<class 'sympy.core.symbol.Dummy'>
```
So for me these seem like the same thing.

ThePauliPrinciple: Currently, `func` always returns the type, however, that means that `atom.func(*atom.args) == atom` generally can't work for `Atom`s (unless the atom stores no information) if `.func` always returns its type since it never stores it in `args`. So it should be made possible that `func` can also return something else.

I think the use of `func` as an alias for `type()` like this is more a remainder of python 2 style classes rather than something that should generally be prefered. If the question is whether it is the same class, then `type()` should be used since it is clearer to the general python reader and it would allow for a more flexible use of `func`.

Maybe @oscarbenjamin has some further insight.

oscarbenjamin:
I'm not sure I understand the purpose of this change. Now that we have `has_free` do we need e.g. `M[1, 1]` to be in its own free symbols?

smichr:
One can differentiate wrt a free symbol so it makes sense to show in a query for free symbols those items that have that quality:
```python
>>> a=MatrixSymbol('A',2,2)
>>> a.diff(a)
PermuteDims(ArrayTensorProduct(I, I), (3)(1 2))
>>> eq=a[0,0]*2
>>> eq.diff(a[0,0])
2
>>> eq = 2*x + 1
>>> eq.diff(x)
2
```
`has`/`has_free` simply tell whether a given node or subset of args for Add/Mul have a given expression.
```python
>>> eq=2*x + 1
>>> eq.has(2*x)
True
>>> eq.has(x)
True
>>> eq.has(2*x + 1)
True
>>> eq.has(2*x + 2)
False
```
But "having" something doesn't indicate whether that thing is symbol like:
```
>>> eq.diff(2*x)
...
ValueError:
Can't calculate derivative wrt 2*x.
```

smichr:
This is the wrong direction to be going. Indexed should not list itself as a free symbol; we should not make MatrixElement behave the same way.

Issue #22260, opened 2021-10-12
Simplify should simplify argument of Sum and Product


oscargus:
```
In [20]: from sympy import *

In [21]: n = Symbol('n', integer=True)

In [22]: Sum(sin(n)**2 + cos(n)**2 - 1, (n, 1, oo)).simplify()
Out[22]: 
    ∞         ∞                      
   ___       ___                     
   ╲         ╲                       
    ╲         ╲   ⎛   2         2   ⎞
-   ╱   1 +   ╱   ⎝sin (n) + cos (n)⎠
   ╱         ╱                       
   ‾‾‾       ‾‾‾                     
  n = 1     n = 1                    
```

Hence, it would make sense to apply `simplify` on the argument first like `Integral` seems to do:
```
In [26]: Integral(sin(x)**2 + cos(x)**2 - 1, (x, 1, oo)).simplify()
Out[26]: 0
```

Issue #17224, opened 2019-07-18
Add finite assumptions to Symbol by default


oscarbenjamin:
Currently a symbol defined with no assumptions e.g. `Symbol('x')` is not assumed to be finite:
```julia
In [34]: x = Symbol('x')

In [35]: print(x.is_finite)
None
```
Pow.is_zero correctly handles this:
```julia
In [42]: m = Mul(0, x, evaluate=False)

In [43]: m
Out[43]: 0⋅x

In [44]: print(m.is_zero)
None
```
However:
```julia
In [45]: m.doit()
Out[45]: 0

In [46]: 0*x
Out[46]: 0
```
Note that if x is *explicitly* declared as infinite then we get:
```julia
In [41]: 0*Symbol('x', infinite=True)
Out[41]: nan
```
There is a contradiction here. Treating the assumptions strictly as defined `0*x` should not evaluate with default assumptions on x. That seems too painful to contemplate changing though.

It seems to me that the assumption that an arbitrary symbol is finite is baked in here and probably some other places. Rather than change that I propose to add `finite=True` by default for a Symbol. Then `0*x -> 0` can be formally correct except when someone goes out of their way to define a possibly infinite symbol.

asmeurer:
I'd be curious if you add it as a default what breaks. I suspect this could be a very difficult thing to change (but I could be wrong).

asmeurer:
I opened a testing PR to see what Travis tests fail https://github.com/sympy/sympy/pull/17227

oscarbenjamin:
I tried this:
```diff
diff --git a/sympy/core/symbol.py b/sympy/core/symbol.py
index 07ff5d4..0e5da3b 100644
--- a/sympy/core/symbol.py
+++ b/sympy/core/symbol.py
@@ -231,6 +231,8 @@ def __new__(cls, name, **assumptions):
         False
 
         """
+        if not assumptions:
+            assumptions = {'finite': True}
         cls._sanitize(assumptions, cls)
         return Symbol.__xnew_cached_(cls, name, **assumptions)
```
Running the tests in core I get a few failures but they are just assumptions tests and the odd test for corner cases to do with symbols being possibly infinite e.g.:
```python
   assert Eq(1/(1/x + 1), 1).func is Eq
```
That test would be invalidated by the change (a PR could set the assumptions on x to fix it). There's a failure that I think is because this needs to be done for Dummy as well:
```python
    def test_disambiguate():
[...]

PR #22278, opened 2021-10-14
Simplify function of Sum and Product on simplify

### LINKED ISSUE #14219: Sum((diag(0, 2, -3))**n, (n, 0, 3)).doit() causes an error
```
>>> k  = symbols('k', positive = True, integer = True)
>>> Sum(x**n/(n + 1), (n, 0, k)).doit()
⎧   k + 1                         log(-x + 1)                     
⎪- x     ⋅lerchphi(x, 1, k + 2) - ───────────  for │x│ ≤ 1 ∧ x ≠ 1
⎪                                      x                          
⎪                                                                 
⎪                  k                                              
⎪                 ____                                            
⎪                 ╲                                               
⎨                  ╲      n                                       
⎪                   ╲    x                                        
⎪                   ╱  ─────                        otherwise     
⎪                  ╱   n + 1                                      
⎪                 ╱                                               
⎪                 ‾‾‾‾                                            
⎪                n = 0                                            
⎩                                                                 
```
An evaluation of a finite sum containing a convergence condition

```
>>> from sympy.concrete.gosper import gosper_sum
>>> gosper_sum(x**n, (n, 0, k))
   k    
x⋅x  - 1
────────
 x - 1  
>>> gosper_sum(n*x**n, (n, 0, k))
  ⎛     k      k    k    ⎞
x⋅⎝k⋅x⋅x  - k⋅x  - x  + 1⎠
──────────────────────────
                2         
         (x - 1)          
```
The above have `x*x**k` rather than `x**(k+1)`

```
>>> A = diag(0, 2, -3)
>>> A**0 + A**1 + A**2 + A**3
⎡1  0    0 ⎤
⎢          ⎥
⎢0  15   0 ⎥
⎢          ⎥
⎣0  0   -20⎦
>>> Sum(A**n, (n, 0, 3)).doit() 

  File "<ipython-input-83-bb0cff02949f>", line 1, in <module>
    Sum(A**n, (n, 0, 3))

  File ".\sympy\core\decorators.py", line 132, in binary_op_wrapper
    return func(self, other)

  File ".\sympy\matrices\common.py", line 2052, in __pow__
    return a._matrix_pow_by_jordan_blocks(num)

  File ".\sympy\matrices\matrices.py", line 1888, in _matrix_pow_by_jordan_blocks
    jordan_cell_power(j, num)

  File ".\sympy\matrices\matrices.py", line 1873, in jordan_cell_power
    raise ValueError("Matrix det == 0; not invertible")

ValueError: Matrix det == 0; not invertible
```

```
>>> Sum((n**(-2/n))**n, (n, 1, oo)).doit()
  ∞          
_____        
╲            
 ╲          n
  ╲   ⎛ -2 ⎞ 
   ╲  ⎜ ───⎟ 
   ╱  ⎜  n ⎟ 
  ╱   ⎝n   ⎠ 
 ╱           
╱            
‾‾‾‾‾        
n = 1        
>>> Sum(n**(-2), (n, 1, oo)).doit()
 2
π 
──
6 
```

oscargus:

Relates to #22260 (also see  #22264 which applies simplification as part of `doit` which also makes sense)
Closes #14219 (just adding test, fixed elsewhere)

The function in the `Sum` and `Product` is simplified when calling `simplify`.

Also, `product_simplify` is improved a bit, honoring simplify keyword arguments and using `sift`.

#### Release Notes

* concrete
   * The function of `Sum` and `Product` is now simplified when calling `simplify`.

### DIFF
--- a/sympy/simplify/simplify.py
+++ b/sympy/simplify/simplify.py
@@ -714,7 +714,7 @@ def simplify(expr, ratio=1.7, measure=count_ops, rational=False, inverse=False, 
             i: factor_terms(i) for i in expr.atoms(Integral)})
 
     if expr.has(Product):
-        expr = product_simplify(expr)
+        expr = product_simplify(expr, **kwargs)
 
     from sympy.physics.units import Quantity
 
@@ -883,15 +883,20 @@ def sum_add(self, other, method=0):
     return Add(self, other)
 
 
-def product_simplify(s):
+def product_simplify(s, **kwargs):
     """Main function for Product simplification"""
     terms = Mul.make_args(s)
     p_t = [] # Product Terms
     o_t = [] # Other Terms
 
+    deep = kwargs.get('deep', True)
     for term in terms:
         if isinstance(term, Product):
-            p_t.append(term)
+            if deep:
+                p_t.append(Product(term.function.simplify(**kwargs),
+                                   *term.limits))
+            else:
+                p_t.append(term)
         else:
             o_t.append(term)
 
@@ -902,8 +907,9 @@ def product_simplify(s):
             if not used[i]:
                 for j, p_term2 in enumerate(p_t):
                     if not used[j] and i != j:
-                        if isinstance(product_mul(p_term1, p_term2, method), Product):
-                            p_t[i] = product_mul(p_term1, p_term2, method)
+                        tmp_prod = product_mul(p_term1, p_term2, method)
+                        if isinstance(tmp_prod, Product):
+                            p_t[i] = tmp_prod
                             used[j] = True
 
     result = Mul(*o_t)
--- a/sympy/concrete/products.py
+++ b/sympy/concrete/products.py
@@ -392,7 +392,7 @@ class Product(ExprWithIntLimits):
 
     def _eval_simplify(self, **kwargs):
         from sympy.simplify.simplify import product_simplify
-        rv = product_simplify(self)
+        rv = product_simplify(self, **kwargs)
         return rv.doit() if kwargs['doit'] else rv
 
     def _eval_transpose(self):
--- a/sympy/concrete/summations.py
+++ b/sympy/concrete/summations.py
@@ -332,8 +332,13 @@ class Sum(AddWithLimits, ExprWithIntLimits):
 
     def _eval_simplify(self, **kwargs):
 
+        function = self.function
+
+        if kwargs.get('deep', True):
+            function = function.simplify(**kwargs)
+
         # split the function into adds
-        terms = Add.make_args(expand(self.function))
+        terms = Add.make_args(expand(function))
         s_t = [] # Sum Terms
         o_t = [] # Other Terms
 
@@ -348,7 +353,7 @@ class Sum(AddWithLimits, ExprWithIntLimits):
                     # go through each term
                     if isinstance(subterm, Sum):
                         # if it's a sum, simplify it
-                        out_terms.append(subterm._eval_simplify())
+                        out_terms.append(subterm._eval_simplify(**kwargs))
                     else:
                         # otherwise, add it as is
                         out_terms.append(subterm)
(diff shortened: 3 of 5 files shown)


oscargus:
Seems like the problem causing the failed test was fixed elsewhere.

sympy/simplify/simplify.py
```diff
@@ -896,34 +896,37 @@ def __refactor(val):
...
+                             lambda i: isinstance(i, Product), binary=True)
 
-    for term in terms:
-        if isinstance(term, Product):
-            p_t.append(term)
+    tmp_prod_terms = []
+    for term in prod_terms:
+        f = term.function.simplify(**kwargs)
+        if f.is_zero:
+            return S.Zero
```
oscarbenjamin, line 910: Thee probably needs to be some checking here:
```python
In [4]: p = Product(Product(sin(t)**2+cos(t)**2-1, (k, 0, 1))*Product(1/k, (k, 0, 1)), (n, 0, 1))

In [5]: p
Out[5]: 
       1                                               
─┬───────────┬─                                        
 │           │     1         1                         
 │           │  ─┬────┬─   ─┬──┬─                      
 │           │   │    │  1  │  │     2         2       
 │           │   │    │  ─⋅ │  │  sin (t) + cos (t) - 1
 │           │   │    │  k  │  │                       
 │           │   │    │    k = 0                       
 │           │   k = 0                                 
 │           │                                         
     n = 0                                             

In [6]: simplify(p)
Out[6]: 0

In [7]: p.doit()
Out[7]: 
                           4
    ⎛   2         2       ⎞ 
zoo⋅⎝sin (t) + cos (t) - 1⎠ 

In [8]: p.doit().simplify()
Out[8]: nan
```

oscargus: In which way should they evaluate? But of course it is possible to skip the early exit.

oscarbenjamin: The most important thing is that we don't get a definite-looking result for a case like this that is really undefined.

oscargus: I came back to this now and it turns out that the problem is actually not in this implementation (although return 0 is not a good idea).

In current master:
```
In [3]: p = Product(sin(t)**2+cos(t)**2-1, (k, 0, 1))*Product(1/k, (k, 0, 1))

In [4]: p.simplify()
Out[4]: 0

In [5]: p.doit()
Out[5]: zoo*(sin(t)**2 + cos(t)**2 - 1)**2
```

oscargus: Which seems to come from:
```
In [3]: p = 0*Product(1/k, (k, 0, 1))

In [4]: p
Out[4]: 0
```

oscargus: I think this comes from:
https://github.com/sympy/sympy/blob/a3826118a64c252bf6170a10b87d15a65728a8f4/sympy/core/mul.py#L684-L691

where especially `is_finite == False` actually should be `is_finite != True`, but that will also give some other interesting effects, like `0*x` not evaluating to `0` with default assumptions.

oscargus: Nah, that should rather be something like:
``` python
            # we know for sure the result will be 0 except the multiplicand
            # is infinity or a matrix
            if any(isinstance(c, MatrixExpr) for c in nc_part):
                return [coeff], nc_part, order_symbols
            if any(c.is_finite == False for c in c_part):
                return [S.NaN], [], order_symbols
            possibly_not_finite = [c for c in c_part if c.is_finite == None]
            return [coeff], possibly_not_finite, order_symbols
```

oscargus: Changing this leads to that the code breaks hard. Not only test failures, but test errors. Will update once everything is ran.

oscargus: Previous was wrong. Correct is probably:
``` python
            # we know for sure the result will be 0 except the multiplicand
            # is infinity or a matrix
            if any(isinstance(c, MatrixExpr) for c in nc_part):
                return [coeff], nc_part, order_symbols
            if any(c.is_finite == False for c in c_part):
                return [S.NaN], [], order_symbols
            c_part = [c for c in c_part if c.is_finite == None]
            if not c_part:
                return [coeff], [], order_symbols
```

But quite a lot of failures and `simplify` still evaluates incorrectly to 0.

oscarbenjamin: See also #17224

oscarbenjamin:
Looks good.

PR #22286, opened 2021-10-14
modify _condsimp to handle missed cases; remove expensive steps from Piecewise.eval

smichr:

Piecewise has an expensive cond->set->relational simplification step that is very expensive. This was removed. But `_condsimp` in meijerint.py depended on those to happen in order to give simpler results. So specific rules for simplification were expanded to deal with this change.

This should address concerns raised in #14933

### Notes for each commit (and I suggest reviewing by commit)

* `_condsimp`/`Piecewise.eval` mods
  * the most important thing to look at are the rules that were added; if these are wrong, results are wrong

* trivial test changes
  * nothing to see here, just things like `2* x + 2  > 0 -> x > -1`

* less trivial changes 
  * changes here are a reflection of Piecewise no longer handling the simplification since Piecewise is built with False in integrals and because of changes to the rules

- [x] use evaluate in Piecewise from integrals.py 
- [x] can probably be put back to False if there are no failures.

#### Release Notes

* functions
    * Piecewise.eval is faster: removed steps that should be in simplification

### DIFF
--- a/sympy/concrete/tests/test_sums_products.py
+++ b/sympy/concrete/tests/test_sums_products.py
@@ -938,11 +938,13 @@ def test_issue_2787():
     res = s.doit().simplify()
     assert res == Piecewise(
         (n*p, p/Abs(p - 1) <= 1),
-        ((-p + 1)**n*Sum(k*p**k*(-p + 1)**(-k)*binomial(n, k), (k, 0, n)),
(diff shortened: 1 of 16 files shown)


smichr:
blocked by #22319

bocklund:
@smichr thanks for the ping and the PR. I haven't done any in-depth profiling, but installing this branch gets our test suite back to what we were experiencing in our monkey patched sympy=1.8, if not better.

smichr:
I will squash the last 11 commits and then this is ready to go.

PR #22292, opened 2021-10-15
Removed pre-1.0 deprecations warnings

### LINKED ISSUE #7013: Prefer init_printing() to %load_ext sympy.interactive.ipythonprinting
The IPython printing extension exists as an extension for historical purposes, but it would be better to use init_printing(), which works exactly the same way, but also works outside of IPython.  This way we can unify our documentation, and remove ambiguity. Plus, init_printing lets you customize the behavior, so you can, for example, choose to use png printing with real LaTeX in the notebook instead of MathJax.

Also, we will eventually make this automatic in IPython anyway ( issue 7003 ), so its primary use will be if you do want to customize it, or use it outside IPython. 

For the even older extension that also imports stuff, use init_session, which imports everything, defines common symbols, and initializes printing.

Original issue for #7013: http://code.google.com/p/sympy/issues/detail?id=3914
Referenced issues: #7003

### LINKED ISSUE #7659: Move gaussopt to optics.gaussopt (DeprecationRemoval)
This is just a `DeprecationRemoval` issue for the same. 

I'm not sure about what release version I should put there. See https://github.com/sympy/sympy/pull/7607

oscargus:

Closes #7013
Closes #7659

Removed the remaining pre-1.0 deprecations.

#### Other comments
Had to change a bit in an example notebook (that still used the Python 2 print-syntax without (), so clearly those are not tested in any way).

Will add a comment in the release notes once merged.

#### Release Notes

NO ENTRY


oscarbenjamin:
Looks good

PR #22293, opened 2021-10-15
Fixes bug when simplifying block matrices

t-ens:
Fixes #22238

When simplifying a block matrix using `bc_dist(expr)` first `expr` is split into a coefficient `factor` and the remaining matrix `mat`. If `mat` was not of an appropriate form rather than returning the original `expr` unchanged it instead returned only `mat` so that the coefficient `factor` was lost. I fixed this and added a test for the issue this PR fixes. 

#### Release Notes

* matrices
  * Fixed bug when simplifying block matrices by distributing a coefficient.


oscargus:
Restarting tests. Seems to make sense.

oscargus:
@t-ens sorry for the delay. Since you contributed this, we have added a test to make sure that all authors are handled correctly. Please see the error message of the failing test.

t-ens:
The authors test now passes.

PR #22296, opened 2021-10-16
Documentation cleanup 6

oscargus:
Some formatting changes.

#### Other comments
Note that argument names with underscores will be type-set strangely by using the recommended back-tick as this gives a single subscript letter. Also, using back-tick for math is a SymPy-convention, so it doesn't work in e.g. Spyder ($ will work after https://github.com/spyder-ide/spyder/pull/16586 is merged).

#### Release Notes

NO ENTRY


smichr:
Looks good. Stop auto-merge if there are objections to edits.

PR #22303, opened 2021-10-17
Add new err classes for DomainMatrix.

skieffer:

We introduce `DM....` counterparts for error classes that were formerly imported from `sympy.matrices`, so that the latter need not be imported here.

#### Release Notes

NO ENTRY


oscarbenjamin:
Looks good, thanks. I've set to automerge after CI

PR #22311, opened 2021-10-18
Code printing improvements for Boolean functions

oscargus:
Added rewriting of BooleanFunction to NNF to enable printing of all Boolean Functions, even though only And, Or and Not are supported.

Added rewriting support for `sinc`, so corresponding code can be removed.

#### Other comments

Changed some explicit rewrites to use `.rewrite`.

Added support for printing Xor and Not in aesara, but do not really understand how the tests work. (But better to have it and not to...)

#### Release Notes

* printing
   * Better support for Boolean functions (`Xor`, `Implies`, etc) in code generation through rewriting.

### DIFF
--- a/sympy/printing/aesaracode.py
+++ b/sympy/printing/aesaracode.py
@@ -51,8 +51,10 @@ if aesara:
             sympy.StrictLessThan: aet.lt,
             sympy.LessThan: aet.le,
             sympy.GreaterThan: aet.ge,
-            sympy.And: aet.and_,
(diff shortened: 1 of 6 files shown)


oscargus:
> I think this looks good in general.

My idea is that it is better that the code generation succeeds, although with possibly sub-optimal code than that it returns non-operational code.

There are some other functions that can be rewritten, to e.g. erf, but that also introduces imaginary units. Am I right that most code generators do not handle that very well?

PR #22314, opened 2021-10-18
Do not evaluate relations in the latex parser

fergusq:

Fixes #22305

Change the LaTeX parser so that it doesn't evaluate relations (it already doesn't evaluate other kinds of expressions).

#### Release Notes

* parsing
  * LaTeX parser doesn't evaluate relations anymore


fergusq:
Hmm, I noticed `parse_expr` evaluates relations too even if `evaluate=False`. It should be fixed too.

asmeurer:
I agree an option would be better. I expect most users would like True as a default, but you definitely should be able to disable it. Ideally relations wouldn't evaluate at all, especially when created directly from the classes.

PR #22316, opened 2021-10-18
Add integral basis and prime factorization for number fields.

skieffer:

Replaces #22168.

Add functions to compute integral bases and prime factorizations in number fields.

The functions can be accessed directly, or through new methods in the `AlgebraicField` class.

#### Release Notes

* polys
  * Add calculation of integral bases for number fields in `polys.numberfields.basis`.
  * Add prime factorization in number fields in `polys.numberfields.primes`.
  * Add other supporting functionality and classes.
  * New functionality can be accessed through new methods in the existing `polys.domains.algebraicfield.AlgebraicField` class.


skieffer:
By the way, I am writing some doctext for `modules.py` which I had meant to include earlier, to help explain its design plan. I also found one or two small things to fix. (In case you want to hold off in reviewing until I can add these things.)

jksuom:
I have finished reviewing `primes`. It looks good but one problem remains: I could not see where the steps 3 - 5 of Algorithm 6.2.9 are implemented.

skieffer:
Quite right -- you do not find them because they are not there! I took the alternative approach described in the remark following the algorithm because (a) I wanted to provide Alg 4.8.17 for valuations at prime ideals anyway, and (b) Cohen says that in practice it is faster.

(I actually think the idea embodied in steps 3-5 is really cool, but still this seemed the better approach.)

I have had other stuff going on, but am hoping to get all the changes discussed so far on this PR made soon. Thanks very much, to both, for your reviewing.

jksuom:
> I took the alternative approach described in the remark following the algorithm

I should have noticed that. It seems to work as `Ip` is a product of maximal ideals and therefore `O/Ip` is separable. Only the valuations have to be computed.

I concur with Oscar that this is extraordinary work. I think that it is ready to be merged after some changes that have already been discussed.

sympy/polys/numberfields/modules.py
```diff
@@ -0,0 +1,1242 @@
...
+
+    def pb_elt_from_col(self, col, denom=1):
+        return self.container(col, denom=denom)
+
+
+class Order(HNF):
+    """
+    An order in a number field.
+
+    This is an HNF that includes unity.
```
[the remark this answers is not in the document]
oscarbenjamin: > then it's their fault if they get bad results?

If it's easy to provide the code that can check this then then the docstring here can point to the `is_order()` method and just explain that this will not be checked automatically.

sympy/polys/numberfields/modules.py
```diff
@@ -0,0 +1,1242 @@
...
+
+    def _first_power(self):
+        return self.copy()
+
+    def __floordiv__(self, a):
+        if is_rat(a):
+            a = QQ(a)
+            if a == 0:
+                raise ZeroDivisionError
+            return self * (1/a)
```
oscarbenjamin, line 1038: We don't need to explicitly raise ZeroDivisionError here because `1/QQ(0)` already does that.

sympy/polys/numberfields/modules.py
```diff
@@ -0,0 +1,1242 @@
...
+        return self.inverse() ** abs(e)
+
+
+class ModuleHomomorphism:
+    """A homomorphism from one module to another."""
+
+    def __init__(self, domain, codomain, mapping):
+        self.domain = domain
+        self.codomain = codomain
+        self.mapping = mapping
```
oscarbenjamin, line 1107: The type and meaning of these parameters needs to be explained. I presume domain and codomain are both of type Module. What should `mapping` be (e.g. a dict)? A doctest example helps a lot in situations like this.

oscarbenjamin: I mapping an arbitrary callable?

sympy/polys/numberfields/primes.py
```diff
@@ -0,0 +1,490 @@
...
+        return f'[ ({p}, {alpha_rep}) e={e}, f={f} ]'
+
+    def __repr__(self):
+        return self.pretty()
+
+    def as_ideal(self):
+        return self.p * self.ZK + self.alpha * self.ZK
+
+    def copy(self):
+        return PrimeIdeal(self.p, self.alpha.copy(), self.e, self.f, self.ZK.copy())
```
oscarbenjamin, line 55: There are a few `copy` methods on different classes in this PR. As far as I can tell all of these classes are immutable so it isn't really meaningful to make copies.

skieffer:
Okay, I will be taking all this into account, and hope to get it pulled together over this week and possibly the next.

(Btw, thanks for the kind words @jksuom ! : )

sympy/polys/domains/algebraicfield.py
```diff
@@ -362,6 +366,58 @@ def from_GaussianRationalField(K1, a, K0):
...
+        ========
+
+        maximal_order()
+
+        """
+        ZK = self.maximal_order()
+        return ZK.QQ_matrix
+
+    def discriminant(self):
+        """Get the discriminant of the field."""
```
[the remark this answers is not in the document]
skieffer: I'm having some trouble with the Sphinx build. Was wondering if you could offer some advice.

I have added some text to the docstring for the `AlgebraicField` class (defined in `sympy.polys.domains.algebraicfield`), which says:

```
    The maximal order, or ring of integers, of the field can also be
    obtained from the :py:meth:`~.maximal_order` method, as a
    :py:class:`~sympy.polys.numberfields.modules.Submodule`.
```
but Sphinx can't seem to resolve the
```
:py:class:`~sympy.polys.numberfields.modules.Submodule`
```
reference at the end. It says,
```
docstring of sympy.polys.domains.algebraicfield.AlgebraicField:171: WARNING: py:class reference target not found: sympy.polys.numberfields.modules.Submodule
```

It's already an absolute path, i.e. starts from `sympy....` so I don't know how I can help it to find this class. Does the class have to be imported into the module where this docstring lives? Because I can't do that, since it causes a circular import.

Have you run into this before? Thanks.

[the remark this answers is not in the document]
oscarbenjamin: I'm not sure. Maybe just `~.Submodule`?

Easiest way to get help with debugging is to push what you have to the PR.

sympy/polys/numberfields/modules.py
```diff
@@ -0,0 +1,1242 @@
...
+        ``g[i] * g[j] == sum(c[k]*g[k], k in range(self.n))``,
+        where ``g`` is the list of generators of this module.
+        If ``j < i`` then ``M[i][j]`` is undefined.
+
+        """
+        raise NotImplementedError
+
+    @property
+    def container(self):
+        """A larger module, in which this one is regarded as a submodule."""
```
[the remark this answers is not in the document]
skieffer: Agreed, and the maximal order itself is a good example, often being larger than the power basis, in the sense of Z-span. Calling it `parent` also makes the notion of "ancestor" more obvious.

sympy/polys/numberfields/modules.py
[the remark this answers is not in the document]
skieffer: Okay, I looked at this one in many different ways. Ultimately, what the functions in `primes` call for is a `Submodule` representing not just an order, but the maximal order, and I don't know of any straightforward way to test for that, aside from essentially running the Round Two algorithm all over again, which would be crazy.

So I've implemented a middle-of-the-road approach:

1. these functions begin by checking some "formal preconditions" for being the right type of `Submodule`, which will at least raise excepetions on obviously ill-formed cases.

2. the boolean properties being checked are pre-set by the `round_two` function, so that, under standard usage, the checks are instantaneous or "free".

Meanwhile, the `Order`, `Ideal`, and `HNF` classes are gone. They weren't needed, so this is a good simplification.

sympy/polys/numberfields/modules.py
skieffer: Yup, it is. Docstring added.

sympy/polys/numberfields/primes.py
skieffer: Agreed. Copying is gone now.

skieffer:
Okay, I think I have addressed all the points discussed above, including adding `polys.numberfields` to the docs.

Sorry there are so many commits. Most are just working on the docs. Only a few commits change code, and the messages should make this clear. 

I've tried to summarize the changes, outside of improving the docs:

* Eliminate the HNF, Order, and Ideal classes
    - Introduce new function to help test formal properties of Submodules
* Introduce `polys.numberfields.exceptions` module
* Revise `ModuleElement.__eq__()`, add `ModuleElement.equiv()`
* Improve division for `ModuleElement`
* Rename `Submodule.container` --> `Submodule.parent`
* Redesign `AlgebraicField.integral_basis()`: now it returns a list of `ANP` instances
* Rename `PrimeIdeal.beta` --> `PrimeIdeal.test_factor`
* Rename `PrimeIdeal.as_ideal` --> `PrimeIdeal.as_submodule`
* Eliminate superfluous copying -- classes are immutable, so no need
* Move `int` conversion from Hermite normal form function in `polys` into corresp. func in `sympy.matrices`
* Remove unnecessary import of `AlgebraicNumber` into `polys.numberfields`
* Build special case into `DomainMatrix.convert_to` (passing `None` just gets a copy)
* Simplify some code
* Use fully qualified imports

skieffer:
I spotted a couple of things to fix up in the docs, which I've just pushed.

Okay, I think I'm done for now.

doc/src/modules/polys/numberfields.rst
```diff
@@ -0,0 +1,289 @@
...
+    such that $\mathbb{Q}(\alpha_1, \ldots, \alpha_m) = \mathbb{Q}(\theta)$.
+13. The Field Isomorphism Problem: Decide whether two number fields
+    $\mathbb{Q}(\alpha)$, $\mathbb{Q}(\beta)$ are isomorphic.
+14. The Field Membership Problem: Given two algebraic numbers $\alpha$,
+    $\beta$, decide whether $\alpha \in \mathbb{Q}(\beta)$, and if so write
+    $\alpha = f(\beta)$ for some $f(x) \in \mathbb{Q}[x]$.
+
+At present only a subset of the tasks enumerated in the lists above is yet
+supported in SymPy, and if you are interested in expanding support, you are
+encouraged to contribute!
```
oscarbenjamin, line 65: This note here is an important one. I think it would be good to make clear that everything needed is in Cohen (if that is the case).

Am I right in thinking that you yourself intend to build on this by implementing some of tasks 4-9?

doc/src/modules/polys/numberfields.rst
```diff
@@ -0,0 +1,289 @@
...
+
+Internals
+=========
+
+Algebraic number fields
+-----------------------
+
+Algebraic number fields are represented in SymPy by the
+:py:class:`~.AlgebraicField` class, which is a part of the
+polynomial domains system.
```
oscarbenjamin, line 92: I think it would be good to add a
```
:ref:`polys-domainsref`
```
here which links through to here:
https://docs.sympy.org/latest/modules/polys/domainsref.html#reference-docs-for-the-poly-domains

doc/src/modules/polys/numberfields.rst
```diff
@@ -0,0 +1,289 @@
...
+forms may be preferable for different computational tasks.
+See [Cohen93]_ Section 4.2.
+
+
+As number field elements
+````````````````````````
+
+In SymPy, there is a distinction between number and expression classes defined
+in the :py:mod:`sympy.core.numbers` module on the one hand, and domains and
+domain elements defined in the :py:mod:`~sympy.polys` module on the other.
```
oscarbenjamin, line 109: This is explained in more detail in
```
:ref:`polys-domainsintro`
```
which links to here:
https://docs.sympy.org/latest/modules/polys/domainsintro.html

doc/src/modules/polys/numberfields.rst
```diff
@@ -0,0 +1,289 @@
...
+
+In order to select a particular conjugate, we can identify
+it with a point in the complex plane. This means identifying two real
+intervals, one on the real axis and one on the imaginary axis, such that
+exactly one root of the given irreducible polynomial over $\mathbb{Q}$ lies
+within the complex region so defined.
+
+Support for this in SymPy is currently limited. The :py:func:`~.isolate`
+function is currently applicable only to algebraic numbers lying on the
+real line.
```
oscarbenjamin, line 133: How exactly is this different from RootOf?
```python
In [8]: rs = [RootOf(x**5 - x + 1, n) for n in range(5)]

In [9]: [r.n(4) for r in rs]
Out[9]: [-1.167, -0.1812 - 1.084⋅ⅈ, -0.1812 + 1.084⋅ⅈ, 0.7649 - 0.3525⋅ⅈ, 0.7649 + 0.3525⋅ⅈ]

In [10]: rs[0]
Out[10]: 
       ⎛ 5           ⎞
CRootOf⎝x  - x + 1, 0⎠

In [11]: rs[0].is_real
Out[11]: True

In [12]: rs[1].is_real
Out[12]: False

In [13]: rs[1].conjugate() == rs[2]
Out[13]: True
```

sympy/polys/numberfields/modules.py
```diff
@@ -0,0 +1,1794 @@
...
+        Parameters
+        ==========
+
+        other : int, :ref:`ZZ`, :ref:`QQ`, :py:class:`~.ModuleElement`, :py:class:`~.Submodule`
+        hnf : boolean, optional (default=True)
+            If ``True``, reduce the matrix of the product module to its
+            Hermite Normal Form.
+        hnf_modulus : :ref:`ZZ`, None, optional
+            If a positive integer is provided, use this as modulus in the
+            HNF reduction.
```
oscarbenjamin, line 1013: This is explained in the `hermite_normal_form` function docstring so a clickable link from here to there like "see the explanation of this parameter at :py:func:..." would be good or at least a "See also" here.

sympy/polys/numberfields/modules.py
```diff
@@ -0,0 +1,1794 @@
...
+
+        Module
+
+        """
+        return None
+
+    def represent(self, elt):
+        """
+        Represent an element of an ancestor module as an integer-linear
+        combination over the generators of this module.
```
oscarbenjamin, line 293: Maybe this docstring should be the one that actually explains what `represent` means rather than the two subclasses. I think it would be good to have (somewhere) a demonstration of how represent works showing how a particular number is represented in a particular module and how you could reconstruct the `Expr` form of a number from the generators and the vector returned here by represent.

sympy/polys/numberfields/modules.py
```diff
@@ -0,0 +1,1794 @@
...
+        number or another :py:class:`~.ModuleElement`, if they represent the
+        same algebraic number.
+
+        Explanation
+        ===========
+
+        When *other* is another :py:class:`~.ModuleElement`, the test for
+        equivalence requires not just that the two elements represent the same
+        algebraic number, but that they have the same :py:class:`~.PowerBasis`
+        ancestor.
```
oscarbenjamin, line 1282: It seems odd to me that it's okay for other to be something completely different that isn't even a ModuleElement but if it is a ModuleElement then it needs to have exactly the same ancester. Why is it useful to have a method that works like this?

sympy/polys/numberfields/modules.py
```diff
@@ -0,0 +1,1794 @@
...
+        Explanation
+        ===========
+
+        For example, if this :py:class:`~.ModuleElement` represents the number
+
+        $\frac{15 a_1 + a_0}{2} = \frac{a_1 + a_0}{2} + 7 a_1$
+
+        then reducing mod 7 should mean throwing away that part that is a poly
+        in the basis elements $a_i$, with content divisible by 7. But this is
+        achieved by reducing our coeffs mod $2 \times 7$.
```
oscarbenjamin, line 1433: I'm finding it hard to get my head round what this means.

sympy/polys/numberfields/modules.py
```diff
@@ -0,0 +1,1794 @@
...
+        """
+        Test whether other is another :py:class:`~.ModuleElement` with same
+        module.
+        """
+        return isinstance(other, ModuleElement) and other.module == self.module
+
+    def make_compat(self, other):
+        """
+        Try to make a compatible pair of :py:class:`~.ModuleElement`, one
+        equivalent to this one, and one equivalent to the other.
```
oscarbenjamin, line 1241: In other contexts in the polys domains this is referred to as unification and the method would be called `unify`. Maybe it would make sense to use that name here or maybe it would be confusing...

sympy/polys/numberfields/modules.py
```diff
@@ -0,0 +1,1794 @@
...
+        Let $M$ be a module, and $E$ its ring of endomorphisms. Let $N$ be
+        another module, and consider a homomorphism $\varphi: N \rightarrow E$.
+        In the event that $\varphi$ is to be represented by a matrix $A$, each
+        column of $A$ must represent an element of $E$. This is possible when
+        the elements of $E$ are themselves representable as matrices, by
+        stacking the columns of such a matrix into a single column.
+
+        This method supports calculating such matrices $A$, by representing
+        an element of this endomorphism ring first as a matrix, and then
+        stacking that matrix's columns into a single column.
```
oscarbenjamin, line 1673: Is it possible to have a simple example here?

doc/src/modules/polys/numberfields.rst
```diff
@@ -0,0 +1,289 @@
...
+   :members:
+
+   .. automethod:: AlgIntPowers.__init__
+
+.. autofunction:: coeff_search
+.. autofunction:: supplement_a_subspace
+
+
+Solving the Main Problems
+=========================
```
oscarbenjamin, line 232: Reading this in the rendered docs this title seems odd coming after the reference docs for the classes and functions above. I wonder if this belongs further up.

doc/src/modules/polys/numberfields.rst
```diff
@@ -0,0 +1,289 @@
...
+---------------------------
+
+.. autofunction:: minimal_polynomial
+.. autofunction:: minpoly
+
+
+The Primitive Element Problem
+-----------------------------
+
+.. autofunction:: primitive_element
```
oscarbenjamin, line 269: This function is badly under-documented. When I wrote the other docs for domains this one slipped under the radar.

doc/src/modules/polys/numberfields.rst
```diff
@@ -0,0 +1,289 @@
...
+The Field Isomorphism Problem
+-----------------------------
+
+.. autofunction:: field_isomorphism
+
+
+The Field Membership Problem
+----------------------------
+
+.. autofunction:: to_number_field
```
oscarbenjamin, line 281: The docstrings here are lacking.

doc/src/modules/polys/numberfields.rst
skieffer: Yes, I would like to implement tasks 4-9, and/or work with others who might be keen on it. And yes, I believe it's all in Cohen. Will add note.

doc/src/modules/polys/numberfields.rst
skieffer: Ah, good. I didn't know how to make links of this kind (i.e. using `:ref:`).

doc/src/modules/polys/numberfields.rst
skieffer: Interesting. TBH, on the code that I didn't write (anything in `numberfields.(minpoly|numbers|isomorphism)` I was doing my best to comment, without knowing the full story. I'll look into this some more.

sympy/polys/numberfields/modules.py
skieffer: Well, it was restricted to these cases only because they are the easy ones (and the only ones I needed!).

Without a common `PowerBasis` ancestor, it essentially means asking whether an element of `QQ(alpha)` and an element of `QQ(beta)` are equal, with `alpha` and `beta` being algebraic numbers not necessarily bearing any obvious relation to one another.

I guess it wouldn't hurt to handle this case too, using the `to_number_field` function. It just means bringing in some heavy machinery that's not required for the cases currently handled.

doc/src/modules/polys/numberfields.rst
skieffer: I'll give it a shot. (Again, this is not my code, but I should probably be getting familiar enough with these to write something reasonable!)

doc/src/modules/polys/numberfields.rst
oscarbenjamin: The code underlying RootOf is in `sympy/polys/rootisolation.py` e.g. `dup_isolate_all_roots`:
```python
In [1]: from sympy.polys.rootisolation import dup_isolate_all_roots

In [2]: dup_isolate_all_roots([ZZ(1), ZZ(0), ZZ(2)], ZZ)
Out[2]: 
([],
 [(((mpq(0,1), mpq(-4,1)), (mpq(4,1), mpq(0,1))), 1),
  (((mpq(0,1), mpq(0,1)), (mpq(4,1), mpq(4,1))), 1)])
```
This shows that `x**2 + 2` has no real roots and two complex roots of multiplicity 1 which are bounded by `(0, 4) + (-4, 0)*I` and `(0, 4) + (0, 4)*I`

doc/src/modules/polys/numberfields.rst
oscarbenjamin: Sorry I thought that you had added these functions.

skieffer:
I've now reviewed all the public functions that existed under `numberfields` before I started doing any work
on SymPy, and this has resulted in a few things:

* I have written expanded docstrings for all these functions, and improved the test coverage.
* I have realized that the way I split it up before (into three modules, `minpoly.py`, `isomorphism.py`, and `numbers.py`) is really not the best way.

I've opened a PR, #22464, that

* reorganizes
* adds docs
* improves test coverage

and I'm wondering what you think. In particular, what would you think about merging that one, and then I could rebase this one on it.

doc/src/modules/polys/numberfields.rst
skieffer: I have looked more closely at this funny `isolate()` function under `numberfields`, and here's what I find:

It is not used at all in SymPy (outside of the one unit test devoted to it), but it is
included in `from sympy import *`.

My best guess is that it exists to support something involving NumPy, SciPy, or other.
This is suggested by the use of `lambdify()` in its implementation. (Docstring for `lambdify` talks
about making SymPy objects usable by NumPy etc.)


Q: How does it compare to `RootOf` / `dup_isolate_all_roots`?

A: It _uses_ such things internally, and its purpose is to accept an algebraic number (not a poly),
_find_ a poly for that number, and then do some isolation.


Q: Why the restriction to _real_ algebraic numbers, since SymPy clearly does support isolating boxes in `CC`?

A: Not clear. Probably pertains to the intended external application, involving NumPy or whatever it was.


Having now reviewed this, I'd like to (a) improve the docstring, and (b) move the function. Please see my
latest comment on this PR.

doc/src/modules/polys/numberfields.rst
skieffer: No worries, I've been meaning to familiarize myself with these anyway.

oscarbenjamin:
> and I'm wondering what you think. In particular, what would you think about merging that one, and then I could rebase this one on it.

That sounds reasonable. I'll focus on that one first.

sympy/polys/numberfields/modules.py
skieffer: On second thought, I think the remaining case (`ModuleElement`s lacking a common ancestor) has to either wait, or just be dropped. It's really not needed at all, and it was just going to be a "bonus" for "completeness".

I think it's perfectly fine to raise `UnificationFailed` here. Also I now make it clear in the docstring that this is a deliberate limitation.

The issue is, I am actually having trouble getting `minpoly.py` and `subfield.py` to answer this question, in the form in which it's given. I'm not yet able to say whether they actually _can't_ do it, or I simply haven't figured out how. In any case, it doesn't seem to be easy.

If it turns out to be easy, I will come back and supply this case. If it turns out to be hard, then we might be talking about some sizable new developments in `minpoly.py` and/or `subfield.py`, which would have to be done in a  separate PR.

skieffer:
I think I've now addressed all the points raised above.

Besides that, I gave the `PrimeIdeal` class a `reduce_poly()` method.
In the commit before that, I also made some minor improvements in `PrimeIdeal`.
The commit messages explain.

oscarbenjamin:
@jksuom I haven't yet gone through this again but does it look good to you?

jksuom:
This looks fine to me, ready to be merged.

oscarbenjamin:
Looks good, thanks!

Thanks for the review @jksuom.

@skieffer this is fantastic. I'm looking forward to what comes later!

skieffer:
Thank you @oscarbenjamin and @jksuom for reviewing! It's a ton of work to review this much code, and I truly appreciate it. Your comments resulted in many substantial improvements. I am definitely proud of this contribution to SymPy! And yes, looking forward to more!

PR #22318, opened 2021-10-18
Fix: abs for numexpr, not Abs

jonas-eschle:

NumexprPrinter has an invalid entry, it's "abs", not "Abs". The name of the function "abs" is "abs" not "Abs".

#### Release Notes

Fix bug in numexpr lambdify that used the invalid identifier Abs instead of abs

 * printing
 	* Fix invalid NumExprPrinter identifier "Abs" to "abs"

### DIFF
--- a/sympy/printing/lambdarepr.py
+++ b/sympy/printing/lambdarepr.py
@@ -95,7 +95,7 @@ class NumExprPrinter(LambdaPrinter):
         'log': 'log',
         'exp': 'exp',
         'sqrt' : 'sqrt',
-        'Abs' : 'abs',
+        'abs' : 'abs',
         'conjugate' : 'conj',
         'im' : 'imag',
         're' : 'real',


smichr:
> Not exactly,

So it would be more accurate to say that it returns Abs or a SymPy object when the argument is of SymPy type.

oscargus:
`key, value pairs correspond to sympy name and numexpr name`

Well, the SymPy name is Abs, so I guess that it is correct. Is this causing any issues?

jonas-eschle:
> `key, value pairs correspond to sympy name and numexpr name`
> 
> Well, the SymPy name is Abs, so I guess that it is correct. Is this causing any issues?

Yes, indeed. Sorry, I thought this is a no-brainer, but it's truly more complex than thought.

Here it is:
```
import sympy
sympy_expr = sympy.sympify('abs(-42)')
sympy.lambdify([], sympy_expr, modules='numexpr')
```

But actually, as I digged in now, the error seems to be in the parsing, not in the lambdification.

This parses correctly
```
sympy.sympify('Abs(-42)', evaluate=False)
```

I'll close this RP and open an issue then

oscargus:
Would it be possible to just add both lines?
```
'Abs' : 'abs'
'abs' : 'abs'
```

(I have no idea how it works, so this is just a random stab...)

(I also noted that it looked like the problem was actually elsewhere, but in case it reappears here...)

PR #22323, opened 2021-10-19
Moved postorder_traversal and interactive_traversal

oscargus:
Moved `postorder_traversal` and `preorder_traversal` in accordance with the previous moves, see #22288 

#### Other comments
Should `interactive_traversal` also move? This would get rid of the `Basic` import (but there will still be `deprecated` imports and imports of the deprecated function).

#### Release Notes

* utilities
   * `postorder_traversal` is moved from `utilities.iterables` to `core.traversal`.
   * `interactive_traversal` is moved from `utilities.iterables` to `interactive.traversal`.

### DIFF
--- a/doc/src/modules/core.rst
+++ b/doc/src/modules/core.rst
@@ -577,6 +577,10 @@ bottom_up
 ^^^^^^^^^
 .. autofunction:: bottom_up
 
+postorder_traversal
+^^^^^^^^^^^^^^^^^^^
+.. autofunction:: postorder_traversal
+
 preorder_traversal
 ^^^^^^^^^^^^^^^^^^
 .. autofunction:: preorder_traversal
--- a/sympy/__init__.py
+++ b/sympy/__init__.py
@@ -200,11 +200,10 @@ from .geometry import (Point, Point2D, Point3D, Line, Ray, Segment, Line2D,
         GeometryError, Curve, Parabola)
 
 from .utilities import (flatten, group, take, subsets, variations,
-        numbered_symbols, cartes, capture, dict_merge,
-        interactive_traversal, prefixes, postfixes, sift, topological_sort,
-        unflatten, has_dups, has_variety, reshape, default_sort_key, ordered,
-        rotations, filldedent, lambdify, source, threaded, xthreaded, public,
-        memoize_property, timed)
+        numbered_symbols, cartes, capture, dict_merge, prefixes, postfixes,
+        sift, topological_sort, unflatten, has_dups, has_variety, reshape,
+        default_sort_key, ordered, rotations, filldedent, lambdify, source,
+        threaded, xthreaded, public, memoize_property, timed)
 
 from .integrals import (integrate, Integral, line_integrate, mellin_transform,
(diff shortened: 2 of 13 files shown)


oscargus:
> Yes, I think it should move.

To `core.traversal` or `interactive`? (I have never used it, but it seems to, maybe, belong more in `interactive` or elsewhere than in `core`.)

oscarbenjamin:
> @oscarbenjamin `mypy` just bit me... No idea what to do since it is some completely other code compared to what I changed...

We're going to see things like this for some time I think. I'm not sure how to describe this but it's almost like mypy is not deterministic some how. Obviously it is deterministic but what exactly it picks up on can be quite sensitive to changes in the codebase. I don't fully understand how it works but I imagine that in mypy's mind the codebase is just a massive mess and sometimes it can pick out something it doesn't like but potentially a small change somewhere else might mean that it doesn't notice any more.

In the short term there is always a quick fix which is just to add `type:ignore` at the exact line that mypy refers to.

The particular error you are seeing is
```console
$ mypy sympy
sympy/core/function.py:2298: error: Incompatible types in assignment (expression has type "Callable[[Any, KwArg(Any)], Any]", base class "EvalfMixin" defined the type as "Callable[[Any, Any, Any, Any, Any, Any, Any], Any]")
Found 1 error in 1 file (checked 1440 source files)
```
The line that it complains about is the assignment `n = evalf` below:
https://github.com/sympy/sympy/blob/e451705e52ed22ffae88da0cbf685b5facb3b7a2/sympy/core/function.py#L2295-L2298
Really the problem though is the fact that the `evalf` method here does not have a consistent signature with the evalf method in the `EvalfMixin` superclass:
https://github.com/sympy/sympy/blob/e451705e52ed22ffae88da0cbf685b5facb3b7a2/sympy/core/evalf.py#L1427

There are several possible fixes:

1. Make both evalf methods have the same signature (the correct fix)
2. Define n as a function rather than using assignment
3. Add a typehint for n
4. Use `type:ignore`

The reason 2. probably works is that it is (apparently) okay for a subclass to redefine a superclass method and change the signature. However assigning an object of one type to replace/shadow one of another type seems to be an error.

In general muliple inheritance is something that leads to complaints from mypy in the absence of type-hints. If you do:
```python
class A:
    def f(self):
        pass

class B:
    def f(self, x):
        return 2

class C(A, B):
    pass
```
Then you get:
```console
$ mypy t.py
t.py:9: error: Definition of "f" in base class "A" is incompatible with definition in base class "B"
Found 1 error in 1 file (checked 1 source file)
```
However this is okay apparently:
```python
class A:
    def f(self):
        pass

class B:
    def f(self):
        pass

class C(A, B):
    def f(self, x):
        return 2
```
I'm not 100% how this all works but what you need to understand is that we're looking at the errors where there are no type hints and mypy gets confused just trying to analyse the code because there are so many examples of bad practice in the codebase. It will work more reliably when there are lots of type hints and when basic problems like this have been ironed out.

oscargus:
OK! I added an ignore comment for now as this is probably something to be dealt with elsewhere.

PR #22324, opened 2021-10-19
ArraySymbol shape parameter is now a tuple

Upabjojr:
Making the parameter _shape_ of _ArraySymbol_ a tuple, so that it is possible to avoid defining the shape.

- tensor
  - Change the signature of `ArraySymbol` from `ArraySymbol(name, *shape)` to `ArraySymbol(name, shape)`.

### DIFF
--- a/sympy/core/tests/test_args.py
+++ b/sympy/core/tests/test_args.py
@@ -4440,7 +4440,7 @@ def test_sympy__tensor__array__array_derivatives__ArrayDerivative():
 def test_sympy__tensor__array__expressions__array_expressions__ArraySymbol():
     from sympy.tensor.array.expressions.array_expressions import ArraySymbol
     m, n, k = symbols("m n k")
-    array = ArraySymbol("A", m, n, k, 2)
+    array = ArraySymbol("A", (m, n, k, 2))
     assert _test_args(array)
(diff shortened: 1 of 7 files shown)


redeboer (approved):
Thanks!
Suggested some minor type hint improvements, but looks ok.

smichr:
> take @Upabjojr's word as to whether this module is public

Sure, no problem. I'm just wondering how this fixes the objective stated in the OP: "so that it is possible to avoid defining the shape." How is requiring star args to be entered as an explicit tuple -- which, as you noted -- is unlike other shape-like things solving the problem. Just a short explanation or an example would be appreciated.

Upabjojr:
Anyway, NumPy always acts on arrays with a defined shape, because arrays are explicit there. Undefined shapes is something of interest to SymPy only.

smichr:
> arbitrary amount of indices

`A(shape=(1, 2, 3))` mean there are 3 indices and each index can take on any value? Or does the shape parameter tell you something about how large each index can be? e.g. `A(2, 2, 3)`. Could `Range` be of any use to indicate a symbolic number of indices with unspecified range for each, otherwise how would you specify that there are `n` indices?

PR #22327, opened 2021-10-19
Ne written in terms of Complement

smichr:

Whereas inequalities require arguments to be real (and are rewritten as sets under that assumption), `Ne` is more general -- even `Ne(nan, 1)` is True. So now rewriting of `Ne` is done in terms of UniversalSet in a Complement (and that set round-trips back to an instance of Ne).

Complement also sympies args, now, so sets will be converted to FiniteSet.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/relational.py
+++ b/sympy/core/relational.py
@@ -709,6 +709,10 @@ class Unequality(Relational):
             return self.func(*eq.args)
         return eq.negated  # result of Ne is the negated Eq
 
+    def _eval_as_set(self):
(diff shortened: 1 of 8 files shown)


smichr:
Ahh, thanks. It must be simplification involving the Not that does this. And that came back when  the Complement with more than one element in excluded FiniteSet was converted back to a Relational...so the problem is there in that a Universal set was reduced to Reals (less the exclusions).

smichr:
This has been instructive, thanks.

PR #22335, opened 2021-10-20
Moved iterable, is_sequence, and NotIterable to utilities.iterables

oscargus:

The three should obviously be in utilities (and the doctests already state that as they were imported through that file).

Had to change about half of the is_sequence imports and 30% of the iterable imports, so many were already pointing to the right location.

#### Other comments

Importing inside `sympify` is not ideal, but it should be possible to move it back to the top-level once the import cleanup is (enough) finished.

There will be a single text in release note outlining this and other moves.

#### Release Notes

NO ENTRY

### DIFF
--- a/doc/src/modules/core.rst
+++ b/doc/src/modules/core.rst
@@ -521,14 +521,6 @@ compatibility
 -------------
 .. module:: sympy.core.compatibility
 
-iterable
-^^^^^^^^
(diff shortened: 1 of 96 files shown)


smichr:
My 22335-eimport is merged with this. Tests are running.

PR #22320, opened 2021-10-18
Run mypy in GitHub Actions on all pull requests

oscarbenjamin:
#### References to other Issues or PRs

Base issue is #17945 but see also #22212, #20173, #20140 and #18244

#### Brief description of what is fixed or changed

Adds a `mypy sympy` step to CI in GH Actions.

#### Other comments

This PR is intentionally minimal. I had to fix one mypy error but otherwise all this does is enable mypy. I would like to get that in so that all further improvements around typing can take place with type hints being checked in CI.

#### Release Notes

NO ENTRY

PR #22337, opened 2021-10-21
Make SymPy PEP 561 compatible

redeboer:

Follow-up to #22212 and and #22320

Now that SymPy is type-checked with mypy, it should signal to downstream packages that it is typed (even if partially). This is achieved by installing a `py.typed` file:
https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

#### Release Notes

* other
  * SymPy is now [PEP 561](https://www.python.org/dev/peps/pep-0561) compatible

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -915,7 +915,9 @@ Raymond Wong <rayman_407@yahoo.com> <rayman@CoolGuy.(none)>
 Raymond Wong <rayman_407@yahoo.com> <rayman@CoolGuyJr.(none)>
 Raymond Wong <rayman_407@yahoo.com> <rayman@CoolGuyJr>
 Rehas Sachdeva <aquannie@gmail.com>
+Remco de Boer <29308176+redeboer@users.noreply.github.com>
 Remco de Boer <redeboer@gmx.com>
+Remco de Boer <redeboer@outlook.com>
 Renato Coutinho <renato.coutinho@gmail.com>
 Renato Orsino <renato.orsino@gmail.com>
 Rhea Parekh <rheaparekh12@gmail.com>
--- a/setup.py
+++ b/setup.py
@@ -446,6 +446,7 @@ if __name__ == '__main__':
               'sympy.parsing.latex': ['*.txt', '*.g4'],
               'sympy.integrals.rubi.parsetools': ['header.py.txt'],
               'sympy.plotting.tests': ['test_region_*.png'],
+              'sympy': ['py.typed']
               },
           data_files=[('share/man/man1', ['doc/man/isympy.1'])],
           cmdclass={'test': test_sympy,


oscarbenjamin:
Would this potentially report typing problems from within the SymPy codebase when a user type-checks their code with mypy or pyright?

I'm worried that SymPy only barely passes the basic checks with mypy. Small changes in the codebase can cause new errors to be thrown up and it will take some time to get it to stabilise.

redeboer:
If I understand correctly, enforces that mypy passes #22320. So externally, won't see any issues. For the rest, types are just `Any` if unspecified.

oscarbenjamin:
> If I understand correctly, #22320 enforces that mypy passes

It passes but it is flakey. My experience of using mypy with SymPy means that I will not presume to understand when or when not mypy will complain about something. The SymPy codebase is messy and I have found numerous things that I presume are bugs in mypy but that I can't reproduce in smaller examples. Seemingly unrelated changes can cause mypy to blow up:
https://github.com/sympy/sympy/pull/22323#issuecomment-946687559

oscarbenjamin:
I'm about to make a new release so a decision on this is needed.

oscarbenjamin:
The test failure is:
```
Remco de Boer <29308176+redeboer@users.noreply.github.com>
Remco de Boer <redeboer@outlook.com>
```
Something needs to be added to .mailmap to specifywhat the author information should be,

redeboer:
> I'm about to make a new release so a decision on this is needed.

Nice, good to hear.

So, to recap, this `py.typed` file is needed when type checking code that uses SymPy as a library
https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages
If not shipped with the library, you'll get (`mypy==0.931`):
```
my_script.py:10: error: Skipping analyzing "sympy": module is installed, but missing library stubs or py.typed marker
```

The main question (https://github.com/sympy/sympy/pull/22337#issuecomment-950160277) is whether SymPy wants to signal that there are type hints (they may not yet be perfect). I'm in favour of shipping the file. If you do run into trouble downstream, you can give feedback or simply ignore SymPy or one of its submodules in your [mypy config](https://mypy.readthedocs.io/en/stable/config_file.html):
```ini
[mypy-sympy.*]
ignore_errors = True
```

oscarbenjamin:
In the absence of any other feedback I say let's go with this.

Thanks @redeboer!

asmeurer:
Does py.typed need to be an empty file? Is it possible to add a comment to it explaining what it does?

.mailmap
```diff
@@ -915,7 +915,9 @@ Raymond Wong <rayman_407@yahoo.com> <rayman@CoolGuy.(none)>
 Raymond Wong <rayman_407@yahoo.com> <rayman@CoolGuyJr.(none)>
 Raymond Wong <rayman_407@yahoo.com> <rayman@CoolGuyJr>
 Rehas Sachdeva <aquannie@gmail.com>
+Remco de Boer <29308176+redeboer@users.noreply.github.com>
 Remco de Boer <redeboer@gmx.com>
+Remco de Boer <redeboer@outlook.com>
```
oscarbenjamin, line 920: This records each name as a separate entry in the AUTHORs file. Which of these do you actually want to record as your name and email address for the AUTHORS file?

merwok:
I think a comment can be included: the PEP mentions special meaning for a line equal to `partial\n`: https://www.python.org/dev/peps/pep-0561/#partial-stub-packages
I didn’t find a note that other lines would cause errors, but I did not re-read the full PEP.  It would probably be worth checking if the most popular type checkers accept comments in the marker file.

.mailmap
redeboer: 29308176+redeboer@users.noreply.github.com

redeboer: Should I update that in a separate PR?

redeboer:
Ah interesting. If I understand correctly, `partial\n` doesn't apply here, because that's only for stubs (`.pyi`) packages. Otherwise, it would be good to add that line.

As for an explanatory comment, I would leave this file empty. It's becoming more common to see it in python libraries and if you want to know why the file was added, commit history can lead you to this PR discussion.

.mailmap
oscarbenjamin: I can fix it during the release. I just need to know which name/email it is that you want to use.

redeboer: Ok thanks. So just `Remco de Boer <29308176+redeboer@users.noreply.github.com>`

oscarbenjamin:
Now that I've been playing around with Python typing for a bit I think it was definitely premature to merge this. SymPy's typing is nowhere near the level needed before exposing that to downstream users.

oscarbenjamin:
I wonder about retracting this until the codebase is ready.

oscarbenjamin:
> Have there been issues filed that suggest removing `py.typed`?

Not directly but there were for example complaints about pylance being extremely slow on the sympy codebase.

What I have found in some code (outside of sympy) where I am using type hints is that just having `import sympy` takes mypy from a 2 second run to a 2 minute run and then doesn't actually produce any usable typing information. Everything still seems to be `Any` or if I try to type things as Expr or Basic it quickly turns out that I can't write any meaningful hinted code that mypy will accept.

That's basically because sympy doesn't have anywhere near enough hints and the code is miles away from being something that can be statically analysed. At the same time it is hard to see how meaningful hints can be added in most places because sympy makes such heavy use of dynamic typing and has heavily overloaded signatures for so many core functions.

> So I think removing `py.typed` doesn't change or improve things for users, while it removes potential for feedback from developers

I'm not sure that we are getting any meaningful feedback. What is really needed is for someone who actually wants to make use of exported typing information to work on it. I expect that potential contributors for this are discourage though because  at least some sympy contributors seem to be opposed to adding hints to the codebase at least partly because they end up being so messy (gh-23249).

oscarbenjamin:
> I've had similar experience. For mypy, SymPy can be ignored with [`follow_imports`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-follow_imports) though, e.g.
> 
> ```ini
> [mypy-sympy.*]
> follow_imports = skip
> ```

Thanks, that's the option I was looking for. Are you already using that option?

> Afaik, there is no way to get mypy to consider SymPy's existing type hints if you do not ship it with `py.typed`.

Yes, but as far as I can tell the existing type hints are almost useless for downstream code. If downstream code just decides to set `follow_imports = skip` then shouldn't sympy just not have the py.typed file in the first place?

I tried using vscode with some fully hinted code that I wrote and all the autocomplete etc is great. Having fully typed code also means that mypy really can pick up on bugs and you can consider to some extent depending on the validation of the type checker rather than only tests. It also forces thinking more carefully about the design of things. None of this is the case with SymPy though because so many core things are not hinted properly. It would be a bunch of work to figure out how to make even vaguely correct hints for most of the commonly used functions.

> > Not directly but there were for example complaints about pylance being extremely slow on the sympy codebase.
> 
> Is that the case since this PR?

I don't remember if it was before or after but to be clear the complaint was about code that uses sympy rather than when editing the sympy code itself. I presume pylance checks py.typed for that.

> Could it be that stubs are more efficient for pylance and perhaps mypy too?

I don't see why they would be more efficient but at least it makes it possible not to complicate the .py files with absurdly complicated type hints for the cases that are hard to define. I am finding that even when you start out wanting to write fully typed code the syntax ends up being very verbose compared to other languages that were designed around static typing in the first place.

In SymPy's case the full correct type hints with all overloads would be very complicated even for the most commonly used functions. Often the way that someone would bring sympy objects into their code is by using `symbols` for example:
```python
from sympy import symbols

x = symbols('x')
poly = x**2 + 2*x + 1
```
The symbols function is impossible to type so that a type-checker can understand that it returns symbols because it also returns tuples of symbols depending on the value of the string:
```
x = symbols('x')
a, b = symbols('a, b')
poly = x**2 + a*x + b
```
It is statically knowable whether the return type would be `Symbol` or `tuple[Symbol]` by looking at the spaces and commas in the string. There is no way any type-checker is going to be able to resolve that though. So then we have to define it like:
```python
def symbols(names: str) -> Symbol | tuple[Symbol, ...]:
    ...
```
Now any type checker would choke on the assignment `a, b = ` because `Symbol` is not iterable or on the power `x**2` because a tuple cannot be squared.

Currently `symbols` is specified to return `Any` because it isn't clear that we can do much better. If we can't sort out things like `symbols` or `sympify` then it is not clear that there is much benefit in exporting type hints. Without the hints there is nothing to stop downstream code from adding their own hints like `x: sympy.Expr` but the value of the SymPy giving explicit hints should be to do more than just expose `Any` (if the goal is to stop the checker then removing `py.typed` is a more effective way).

I tried to write hints for `symbols`:
```python
from typing import TypeVar, Callable, cast

T = TypeVar('T')

def symbols(names: str,
            *,
            cls: Callable[[str], T] = Symbol,
            **args: Any,
            ) -> T | tuple[T, ...]:
```
I'm not sure how to get mypy to understand what the `cls` argument is supposed to be or that it should also be able to receive `**args`. In any case it fails with:
```
sympy/core/symbol.py:592: error: Incompatible default for argument "cls" (default has type "Callable[[str], Symbol]", argument has type "Callable[[str], T]")  [assignment]
```
Can it not understand that `Callable[[str], Symbol]` is a case of `Callable[[str], T]`?

Probably this needs a bunch of overloads etc. Instead if I just change the return type to `Symbol | tuple[Symbol, ...]` then I get 200 errors from the sympy codebase (mostly tests) like:
```
sympy/core/tests/test_expr.py:50: error: "Symbol" object is not iterable
```
That's presumably indicative of what would happen to any type-hinted downstream codebases if sympy actually added type hints for all of these things. It would "break" their typing so now having the py.typed file means that adding type hints is almost a backwards compatibility consideration.

oscarbenjamin:
> > I've had similar experience. For mypy, SymPy can be ignored with [`follow_imports`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-follow_imports) though, e.g.
> > ```ini
> > [mypy-sympy.*]
> > follow_imports = skip
> > ```
> 
> Thanks, that's the option I was looking for

Bizarrely when using this I got a failure:
```
error: Cannot find implementation or library stub for module named
"sympy"  [import]
        import sympy
    ^
```
So I had to add 
```
ignore_missing_imports = true
```
as well.

oscarbenjamin:
See also gh-26150 and https://github.com/microsoft/pyright/issues/7159.

I think that the py.typed file should be removed. It was a mistake to add it before adding the corresponding type hints.

Issue #17587, opened 2019-09-08
CSE with Order around nonzero point loses terms


oscarbenjamin:
Using `cse` with an Order works fine if the Order is around 0:
```julia
In [16]: from sympy.simplify import cse

In [17]: e = exp(x).series(n=5)

In [18]: e
Out[18]: 
         2    3    4        
        x    x    x     ⎛ 5⎞
1 + x + ── + ── + ── + O⎝x ⎠
        2    6    24        

In [19]: cse(e)
Out[19]: 
⎛    ⎡         2    3    4        ⎤⎞
⎜    ⎢        x    x    x     ⎛ 5⎞⎥⎟
⎜[], ⎢1 + x + ── + ── + ── + O⎝x ⎠⎥⎟
⎝    ⎣        2    6    24        ⎦⎠
```
If the Order is around any other point though then `cse` absorbs all the terms into the Order:
```julia
In [20]: e = exp(x).series(x, 1, n=5)

In [21]: e
Out[21]: 
                         2            3            4                     
                ℯ⋅(x - 1)    ℯ⋅(x - 1)    ℯ⋅(x - 1)     ⎛       5       ⎞
ℯ + ℯ⋅(x - 1) + ────────── + ────────── + ────────── + O⎝(x - 1) ; x → 1⎠
                    2            6            24                         

In [22]: cse(e)
Out[22]: ([(x₀, x - 1)], [O(1; x → 1)])
```

RituRajSingh878:
I think Order has a problem.

eg.
```
In [3]: o=Order(x**6)

In [4]: o=o.subs(x,x-1)

In [5]: o
Out[5]: 
 ⎛       6       ⎞
O⎝(x - 1) ; x → 1⎠

In [6]: o=o.subs(x-1,x)

In [7]: o
Out[7]: O(1; x → 1)

```

oscarbenjamin:
Is that what happens in cse?

oscarbenjamin:
I guess that could be fixed in here:
https://github.com/sympy/sympy/blob/fcf84d99dd243b9921eaebe19281b692feb7869a/sympy/series/order.py#L419

RituRajSingh878:
> Is that what happens in cse?

yes

anutosh491:
> I think Order has a problem.
> In [3]: o=Order(x**6)
> 
> In [4]: o=o.subs(x,x-1)
> 
> In [5]: o
> Out[5]: 
>  ⎛       6       ⎞
[...]

PR #22339, opened 2021-10-21
Fixes Subs where the original and the substituted expression both have an Order term

### LINKED ISSUE #19120: sin works well but cos return wrong
```
from sympy import *
cos(x).series(n=8).subs(x,cos(x).series(n=8))
```

this return type error

image

but if we use sin

```
from sympy import *
sin(x).series(n=8).subs(x,sin(x).series(n=8)).expand()
```

image

anutosh491:

Fixes #19120 

Subs wasn't working for cases where both the original and the expression to be substituted have an Order term . All return the same error For eg
```
>>> O(x).subs(x,  O(x))
>>> (x + O(x**2)).subs(x, x + O(x**2))
>>> O(x**2).subs(x, x + O(x**2))
>>> O(x**2 + 1, (x, oo)).subs(x, x + O(x**2, (x, oo)))
>>> (x - x**3/6 + x**5/120 - x**7/5040 + O(x**8)).subs(x,sin(x).series(n=8)).expand()
>>> (cos(x).series(n=8)).subs(x,sin(x).series(n=8)).expand()

  File "c:\users\anuto\sympy\sympy\sympy\core\basic.py", line 953, in subs
    rv = rv._subs(old, new, **kwargs)
  File "c:\users\anuto\sympy\sympy\sympy\core\cache.py", line 72, in wrapper
  File "c:\users\anuto\sympy\sympy\sympy\core\basic.py", line 1067, in _subs
    rv = fallback(self, old, new)
  File "c:\users\anuto\sympy\sympy\sympy\core\basic.py", line 1039, in fallback
    arg = arg._subs(old, new, **hints)
  File "c:\users\anuto\sympy\sympy\sympy\core\cache.py", line 72, in wrapper
  File "c:\users\anuto\sympy\sympy\sympy\core\basic.py", line 1065, in _subs
    rv = self._eval_subs(old, new)
  File "c:\users\anuto\sympy\sympy\sympy\series\order.py", line 479, in _eval_subs
    res = [dict(zip((d, ), sol))]
TypeError: 'ConditionSet' object is not iterable
```

Now many of these can be solved and though all expressions return the same error statement there are 2 reasons why this was happening . Firstly we should realize that `_eval_subs()` method of `order.py` would be used when the original expression has an order term
Reason 1) The logic implemented won't work correctly when the expression to be substituted has an Order term because of the `point = new.subs(var, self.point[i])` (line 472). Here we need to get the point at which the order has to be implemented . If `new` has an order term we would get something like `O(0)` or `O(oo)` because ` O(x).subs(x,0)` won't be giving us 0 rather would give us `O(0)` .Hence before comparisons we need to strip down the `O` from the expression and still there is a chance for longer expressions where we could land on a point having Order hence we need to strip that again ! Results below 
I have tried to show mathematical equivalence for one of them 
```
>>> O(x**2).subs(x, x + O(x))
O(x**2)
>>> (x + O(x**2)).subs(x, x + O(x**2))
x + O(x**2)
>>> O(x**2).subs(x, x + O(x**2))
O(x**2)
>>> (x**2+ O(x**2)).subs(x, x + O(x**2))
O(x**2)
>>> (x**2+ O(x**2) + 1/x**2).subs(x, x + O(x**2))
(x + O(x**2))**(-2) + O(x**2)
>>> (x**2+ O(x**2) + 1).subs(x, x + O(x**2)) # nothing but (x + O(x**2))**2 + O((x + O(x**2))**2) + 1 
1 + O(x**2)                                 # which returns same as shown below
>>> (x + O(x**2))**2 + O((x + O(x**2))**2) + 1
1 + O(x**2)
>>> (1 + x + O(x**2, (x, oo))).subs(x, x + O(x , (x, oo))) # nothing but 1 + (x + O(x , (x, oo))) + O((x + O(x , (x, oo)))**2)
O(x**2, (x, oo))
>>> 1 + (x + O(x , (x, oo))) + O((x + O(x , (x, oo)))**2)
O(x**2, (x, oo))
```
Point needs to be simplified too for solving longer expressions. I have checked the mathematical correctness here . There is no change in point in these cases though.
```
>>> (x - x**3/6 + x**5/120 - x**7/5040 + O(x**8)).subs(x,sin(x).series(n=8)).expand() 
x - x**3/3 + x**5/10 - 8*x**7/315 + O(x**8)

>>> (cos(x).series(n=8)).subs(x,sin(x).series(n=8)).expand()
1 - x**2/2 + 5*x**4/24 - 37*x**6/720 + O(x**8)

>>> tan(x).series(n=8).subs(x,tan(x).series(n=8)).expand()
x + 2*x**3/3 + 3*x**5/5 + 181*x**7/315 + O(x**8)
```

Reason2) There are cases which can't be approached as of now because they haven't been implemented in `solveset` and also importantly the point changes from the original point. But as the workflow was similar, it was showing the same error which shouldn't be the case as it is misguiding hence I have raised a `NotImplementedError` there!! This comes into picture because of (line 479) `sol = solveset(old - new.subs(var, d), d)` . Here sol is expected to be used as a point but in many cases it returns a `ConditionSet` which is not wrong but also not what we expect here .Even `solvify` returns `NotImplementedError` for these. I guess `sol` would be using the real solution of the following

```
>>> solveset(x - (1 - x**2/2 + x**4/24 - x**6/720 + O(x**8)), x)
ConditionSet(x, Eq(-1 + x + x**2/2 - x**4/24 + x**6/720 + O(x**8), 0), Complexes)

>>> solveset(x - cos(x), x,)  # should return in form of LambertW 
ConditionSet(x, Eq(x - cos(x), 0), Complexes)

>>> solveset(x - sin(x), x) # should return in form of LambertW  
ConditionSet(x, Eq(x - sin(x), 0), Complexes)

>>> solveset(x - exp(x), x) # should return in form of LambertW 
ConditionSet(x, Eq(x - exp(x), 0), Complexes)

>>> solveset(x - log(x), x) # should return in form of LambertW  
ConditionSet(x, Eq(x - log(x), 0), Complexes)
```
See #7872 which fails for the same reason as SymPy currently can't solve `x**3 = exp(-x**2) ` which is probably not a basic case as above to return in terms of `LambertW` and maybe need a numerical solution as given by other CAS' like Wolfram etc

For eg implementation in Pr
```
>>> (cos(x).series(n=8)).subs(x,cos(x).series(n=8)).expand()
NotImplementedError: sympy is unable to solve the following equation: -1 + x + x**2/2 - x**4/24 + x**6/720 + O(x**8) = 0
```
Now even if we would try some cheap tricks this would fail because we actually don't have anything to solve this . If we strip of the `O(x**8)`, this leads to `NotImplementedError` for all such expressions from other place (line 166 in `order.py`) but the raised statement doesn't portray why we land up on this error as it is originally Implemented for something else. if we rewrite it as `cos(x)` even that hasn't been implemented . 
Only option we had was to check if original expr is Add and strip the Order term in the original expression 

```
>>>(cos(x).series(n=8).removeO()).subs(x,cos(x).series(n=8)).expand()
389/720 + 101*x**2/240 - 37*x**4/360 - 53*x**6/10800 + O(x**8)
```
But this is not correct and would lead to mathematical incorrectness at large number of places including most of the  results shown above. 

#### Release Notes

* series
  * Fixes Subs where the original and the substituted expression both have an Order term.


anutosh491:
Also while trying to figure what was going on here . I realized that any subs except for the major symbol (x) won't be working as required and might luckily give the correct result .See the following 
```
>>> (O(x**(n + 2))).subs(n, 2)
O(x**2)
>> O(x + n*x**2, (x, oo)).subs(n,0)
O(x**2, (x, oo)). 
```
Obviously Sympy considers the above expression as Multivariate O `O(f(x, y), x, y)`  and transforms the expressions into `O(f(x, y).as_leading_term(x,y).as_leading_term(y), x, y)` automatically hence almost any subs which is not intended to be multivariate would fail. Not exactly sure how we should be dealing with these

anutosh491:
I have also tried to address a subtle issue which might be leading to errors here and there as pointed out in #17587 !
An example for the same would be :
```
In [3]: o=Order(x**6)

In [4]: o=o.subs(x,x-1)

In [5]: o
Out[5]: 
 ⎛       6       ⎞
O⎝(x - 1) ; x → 1⎠

In [6]: o=o.subs(x-1,x)

In [7]: o
Out[7]: O(1; x → 1)

```
Here we could see that while substituting back `x` for `x-1`, we might not be getting the expected results ( `O(x**6)` ), as it might be substituting back the expression but doesn't transform the point as should be done like the first initial case.

anutosh491:
ping @jksuom for review !

sympy/series/tests/test_order.py
```diff
@@ -409,6 +409,16 @@ def test_order_subs_limits():
     assert Order(x**2).subs(x, y - 1) == Order((y - 1)**2, (y, 1))
     assert Order(10*x**2, (x, 2)).subs(x, y - 1) == Order(1, (y, 3))
 
+    # issue 19120
+    assert O(x).subs(x, O(x)) == O(x)
+    assert O(x**2).subs(x, x + O(x)) == O(x**2)
+    assert (x + O(x**2)).subs(x, x + O(x**2)) == x + O(x**2)
```
0sidharth, line 415: Is this correct? Substituting would lead to O((x + O(x**2))**2) which would have an O(x**4) term.

anutosh491: Yeah , I've made sure all results are consistent with master to check for any errors . See below for manual substitution of the same ! The second term `O(x**2)` captures the `O(x**4)` term as you pointed out .
```
>>> (x + O(x**2) + O((x + O(x**2))**2))
x + O(x**2)
```

sympy/series/order.py
```diff
@@ -488,6 +497,12 @@ def _eval_subs(self, old, new):
                 else:
                     return
             return Order(newexpr, *zip(newvars, newpt))
+        else:
+            point = list(self.point)
+            old_update = old + point[0]
+            new_update = new + point[0]
+            if old_update !=old and new_update !=new:
```
0sidharth, line 504: Space around binary operators

sympy/series/order.py
```diff
@@ -467,17 +467,26 @@ def _eval_subs(self, old, new):
...
+                        sol_str = str(old - new)
                         sol = solveset(old - new.subs(var, d), d)
                         if isinstance(sol, Complement):
                             e1 = sol.args[0]
                             e2 = sol.args[1]
                             sol = set(e1) - set(e2)
-                        res = [dict(zip((d, ), sol))]
-                        point = d.subs(res[0]).limit(old, self.point[i])
+                        try:
+                            res = [dict(zip((d, ), sol))]
```
0sidharth, line 485:
```suggestion
                            res = [dict(zip((d,), sol))]
```

0sidharth:
Currently this PR gives `NotImplementedError` for - 
```py
In [16]: O(x).subs(x, O(1/y))
NotImplementedError                       Traceback (most recent call last)
...
~/sympy/sympy/series/order.py in __new__(cls, expr, *args, **kwargs)
    163                 if v in new_vp.keys():
    164                     if p != new_vp[v]:
--> 165                         raise NotImplementedError(
    166                             "Mixing Order at different points is not supported.")
    167                 else:

NotImplementedError: Mixing Order at different points is not supported.
```

Not sure about these but - 

Since `O(1/y)` is also around `y = 0`, the error message does not seem very accurate. Also, should the answer just be `O(1/y)`, instead of throwing an error?

anutosh491:
> Currently this PR gives `NotImplementedError` for -
> 
> ```python
> In [16]: O(x).subs(x, O(1/y))
> ---------------------------------------------------------------------------
> NotImplementedError                       Traceback (most recent call last)
> ...
> ~/sympy/sympy/series/order.py in __new__(cls, expr, *args, **kwargs)
>     163                 if v in new_vp.keys():
>     164                     if p != new_vp[v]:
> --> 165                         raise NotImplementedError(
>     166                             "Mixing Order at different points is not supported.")
>     167                 else:
> 
> NotImplementedError: Mixing Order at different points is not supported.
> ```
> 
> Not sure about these but -
> 
> Since `O(1/y)` is also around `y = 0`, the error message does not seem very accurate. Also, should the answer just be `O(1/y)`, instead of throwing an error?

Thanks for pointing this out , I didn't really think of the replacing a positive power with a negative power while making this Pr. Not sure why something like `O(x**a).subs(x, O(x**b))` where `a` is +ve and `b` is -ve , didn't come to my mind ! 
I'll look into this and get back soon

anutosh491:
This had failed due to the benchmarks issue which happened some days back , else this should work !

anutosh491:
ping @jksuom @0sidharth @smichr for review

anutosh491:
ping @jksuom @0sidharth @smichr for review . I am satisfied from the changes made as the issue didn't demand much, I would be glad if I could have your views for the same !

anutosh491:
Ping @czgdp1807 for review !

anutosh491:
ping @jksuom for review ! Haven't heard from you on this Pr yet , would be glad if you could review this . Looks fairly good to me !

anutosh491:
CC @0sidharth for reviews !

anutosh491:
ping @jksuom for review ! Haven't heard from you on this Pr yet , would be glad if you could review this . Looks fairly good to me !

anutosh491:
ping @jksuom for review. Would be glad to hear from you on this !

PR #22344, opened 2021-10-21
added Rem(ainder) function

### LINKED ISSUE #22090: Add Rem(ainder) function?
Related to #22089

So far, `Mod(x, y)` has been printed in C as `x % y`. This is only correct if the sign of x and y are the same (or rather both being either non-negative or non-positive). Typically, if a language has support for both modulo and remainder, `Mod` has the same sign as the divisor/denominator and the remainder the same sign as the dividend/numerator, see this historic version of the Wikipedia page: https://en.wikipedia.org/w/index.php?title=Modulo_operation&oldid=764138354 The new version of the page use the terms floored (modulo) and truncated (remainder): https://en.wikipedia.org/wiki/Modulo_operation There is also a Euclidean modulo operation, where the result is always non-negative. The Python operation `%` corresponds to the modulo/floored version, as do SymPy's `Mod`.

It may make sense to add a `Rem` function to SymPy, if nothing else to simplify printing of these things (by rewriting to the version that is supported in the particular language) and to clarify the difference between these.

| x | y | mod(x, y) | rem(x, y) |
|--|--|--------|-----|
| 5 | 3 | 2 | 2|
| -5 | 3 | 1 | -2 |
| 5 | -3 | -1 | 2 |
| -5 | -3 | -2 | -2 |

To rewrite `Mod(x, y)` to `Rem`, one can use `Rem(Rem(x, y) + y,  y)` (unless one know that x and y are both non-negative or non-positive, then it is just `Rem(x, y)`).

gurpartap0306:

Fixes #22090 

There was no function to find the remainder if one number was negative

#### Release Notes

* core
  * Added a function `Rem` to find remainder of both positive and negative numbers


oscarbenjamin:
I think that making a special module in core for Mod is a mistake. We shouldn't repeat that by making a special module in core for Rem. This belongs in `sympy/functions`.

PR #22345, opened 2021-10-21
codegen/ast: Argument checks for Return

ThePauliPrinciple:
In order to easily identify a class from codegen.ast, a (empty)
class was created from which all codegen.ast classes inherit.

`Return` did not have appropriate checks for its input arguments
It now no longer allows non `Basic` input arguments. It is now also
a subclass of `Token` and the documentation is updated to reflect
this.

`Statement` was indicated in the documentation of codegen.ast.py,
however, it is never is defined. It is now indicated in the documentation
that `Statement` is not yet implemented

Previous example:
```python
isinstance(ast_object,(AssignmentBase,CodeBlock,Token,Statement,Return))
```
Example:
```python
isinstance(ast_object,CodegenAST)
```
```python
    >>> from sympy.codegen.ast import Return
    >>> from sympy.printing.pycode import pycode
    >>> from sympy import Symbol
    >>> x = Symbol('x')
    >>> print(pycode(Return(x)))
    return x
    >>> print(pycode(Return(x,x)))
    return (x,x)
```
#### Release Notes

NO ENTRY


ThePauliPrinciple:
I have reverted the "not implemented" comment in the docstring (and started a seperate PR on this topic). I also changed the `Return` code to the simplified version suggested by @bjodah

bjodah:
Thanks!

PR #22346, opened 2021-10-21
Update Matrix shape documentation

angelagomba:
Documentation to get the shape of a Matrix is outdated. Current documentation causes `ImportError: cannot import name 'shape' from 'sympy'` error. Updated the documentation to use M.shape as opposed to shape(M) function.

Updated the documentation for getting the shape of a matrix 

#### Release Notes

NO ENTRY


oscarbenjamin:
SymPy Live is using an out of date SymPy version

PR #22351, opened 2021-10-22
[WIP] Move codegen files from utilities

oscargus:
An additional step in the cleanup of utilities. Files related to code generation is moved to codegen. The most doubtful is probably lambdify.

[ ] Decide on which functions should have re-added deprecated versions in utilities.
[ ] Add those functions.
[ ] Update documentation.
[ ] Update labels.

#### Other comments

I have not fully understood the relation between utilities.codegen and codegen. There seems to be some connections though, at least in terms of the final goal of the code...

Will add a post in the release notes about the restructuring.

#### Release Notes

NO ENTRY


oscargus:
There is an alternative solution to the more direct import order problem in #22386 by basically removing all the top-level core imports in `lambdify`. It will not break anything, enable some sort of hierarchy between the packages, but will not move `lambdify` and the `codegen` packages (so not solve that "problem").

I still think that the files that are moved here are misplaced, but it is not obvious where they should end up. As there are some efforts to generalize lambdify and autowrap etc in #22365, hopefully that will be the better method to deal with this in the long term.

Also, this gave at least me a good insight into the importance of documenting things in a clever way to enable restructuring of code. If we in the documentation specify the full path, the users will more likely use the full path.

I'm close to closing this in preference of #22386 as that seems to be a less controversial way to obtain what I aim for.

oscargus:
> We could also make a `softdeprecated`

Since I've moved a few functions recently, I've been thinking about the need for a different wrapper for functions that are just relocated. These are not deprecated as such, so the message is a bit misleading. Basically, it would be better to have something say directly: "just change you import path to this to get rid of the warning". Any opinions?

oscarbenjamin:
> > We could also make a `softdeprecated`
> 
> Since I've moved a few functions recently, I've been thinking about the need for a different wrapper for functions that are just relocated. These are not deprecated as such, so the message is a bit misleading. Basically, it would be better to have something say directly: "just change you import path to this to get rid of the warning". Any opinions?

I agree but I also think that we need a softdeprecated which is something that shows as deprecated in the docs and help etc but does not actually emit a warning. If in the current version of sympy the only way to import something is by using `from sympy.x.y import z` then if that starts to give a warning in the next sympy version it makes it difficult for a downstream library to support multiple sympy versions. We should ideally deprecate things in the documentation and release notes some time before actually emitting a warning.

Exactly when to do this does depend to some extent on whether or not it is documented to import things in a certain way so if someone is importing something that should really be considered private or they are importing it in a way that is not documented then that is potentially different. If the docs currently say to do `from sympy.x.y import z` or if that's the only way to use a public function `z` then it's premature to jump to a warning in the next release.

oscargus:
I close this since other PRs achieves the goal from an import order aspect.

PR #22365, opened 2021-10-25
[WIP] CodeParsers for preprocessing expressions given to CodePrinters

ThePauliPrinciple:
SymPy features a very extensive printing system. In this sytem, the `doprint`
method of an appropriate `Printer` is used to obtain a string. Internally,
doprint will use the `_print` method which looks for an appropriate
`_print_<class>` based on the class (or its parent classes) of the object
being printed. Additionally, an object can define its own printing method
by giving it the appropriate name (Printer dependent) such that users
can define the printing behaviour for their custom objects. These
`_print_<class>` methods will often call `_print` again. For example: `Mul`,
the object in SymPy representing multiplication, might use `_print` on all
of its arguments and then join the obtained strings with a "*" character.

Here, the `Parser` class is introduced which generalises the above concepts.
The `Parser` class allows the use of not only `doprint`, `_print` and
`_print_<class>`, but now allows print to be replaced with any name (without
underscores "_") using `do<parsername>`, `_<parsername>` and
`_<parsername>_<class>`. Multiple such parsers can be stored in the same
object since they should not conflict. Additionally, `do<parsername>` and
`_<parsername>`	are automatically constructed if they are not provided.

Currently, several parts of SymPy rely heavily on the code printers. For
example, `lambdify` and `autowrap` use the code printers to convert SymPy
expressions into code that can be run in several different languages.
However, both use their own code generation to obtain the code needed to
embed the printed expression in a form that is expected of a programming
language (e.g. function definitions, variable declarations etc.). Code
parsers could be used to perform these tasks, such that new functionality
is immediately available in both `autowrap` and `lambdify` as well as to
the user simply wanting to obtain the code and reuse it in their own
projects.

Recently, common subexpression extraction (CSE) was implemented in
`lambdify`, which was also used to delete temporary variables to reduce
the memory usage. Automatic (de)allocation would also be useful for
`autowrap` but would now need to be implemented again. On top of that,
putting new functionality in code parsers would stop bloating the already
lengthy lambdify and autowrap functions.

#### Other comments

Examples:

This google colab contains a simple example of how to use code parsers (more to follow)
https://colab.research.google.com/drive/1ugRSYuA6fdvNpZlHss7aLx3r3YSLJ7tB?usp=sharing
- Creating a function definition around a SymPy `Expr` with `Parser`
- Creating a (simplified version of) `lambdify` with `Parser`

Known issues:

- [ ] Currently pickling does not work for the new Parser class
- [x] Some doc tests are still failing.  

To do:

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/physics/vector/printing.py
+++ b/sympy/physics/vector/printing.py
@@ -62,7 +62,7 @@ class VectorLatexPrinter(LatexPrinter):
                 base = self.parenthesize_super(base)
                 return r"%s^{%s}" % (base, exp)
             else:
-                return super()._print(expr)
+                return self._print(expr)
         else:
             return super()._print_Function(expr, exp)
 
--- a/sympy/printing/mathml.py
+++ b/sympy/printing/mathml.py
@@ -67,7 +67,7 @@ class MathMLPrinterBase(Printer):
         """
         Prints the expression as MathML.
         """
-        mathML = Printer._print(self, expr)
+        mathML = self._print(expr)
         unistr = mathML.toxml()
         xmlbstr = unistr.encode('ascii', 'xmlcharrefreplace')
         res = xmlbstr.decode()
--- a/sympy/printing/printer.py
+++ b/sympy/printing/printer.py
@@ -218,9 +218,7 @@ from functools import cmp_to_key, update_wrapper
 from sympy.core.add import Add
 from sympy.core.basic import Basic
 
-from sympy.core.core import BasicMeta
-from sympy.core.function import AppliedUndef, UndefinedFunction, Function
-
+from sympy.utilities.parser import Parser
 
 
 @contextmanager
@@ -233,7 +231,7 @@ def printer_context(printer, **kwargs):
         printer._context = original
 
 
-class Printer:
+class Printer(Parser):
     """ Generic printer
 
     Its job is to provide infrastructure for implementing new printers easily.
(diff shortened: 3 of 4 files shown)


oscarbenjamin:
There are two fundamental problems with the current printing dispatch system:

1. It is based on the class name: what if two classes have the same name?

2. Everything needs to be add to the main printer class so it is not easily extended.

Your suggestion to add a method makes it extensible from outside (solving problem 2) but single dispatch solves problems 1 and 2. With single dispatch downstream classes can simply register with the appropriate method and there doesn't need to be any upstream code changes.

ThePauliPrinciple:
I see now why using the classes themselves would be helpful. In fact, a lot of the code inside `_print` is actually dealing with that, when someone subclassed the `Function` class it needs to not use that name but instead print through `_print_Function` because the class might now be called something that already is in SymPy. 

What still needs some thought:
- singledispatchmethod being introduced only in 3.8 (so could not be used in SymPy until support for 3.7 is dropped)
- Currently, a class can define its own print method, which would prioretise over anything provided in the printer (I can say as a user I would be very disappointed if this were dropped. It is not so easy as to just dispatch that method since it is part of the object being printed, not of the printer (those methods have their first and second argument reversed compared to the printer methods)
- Possible circular inputs. If I understand correctly using the single dispatch system would require all classes to already be loaded when the module gets loaded (when the class objects get created), this might be tricky since the printers are used in `__str__`/`__repr__` 

I tried to use the forward declaration for codehints, but it seems this can not straightforwardly be used with single dispatch:
```python
from functools import singledispatchmethod
class A():
    @singledispatchmethod
    def _print(self,obj):
        print('fall back')
    @_print.register
    def _(self,obj:"undefined"):
        print('got undefined')
class undefined(): pass
A()._print(undefined)
```

ThePauliPrinciple:
As for the second point, it seems that it should be possible to do something along the lines of subclassing `singledispatchmethod`'s `__get__` method:
```python
    def __get__(self, obj, cls=None):
        def _method(*args, **kwargs):
           #check if we can use the printmethod defined in the class being printed
           #self.func.__name__ should also be available, so it should be possible to do this also for
           #dispatchers other than `_print`
            if hasattr(args[0],obj.printmethod):
                return getattr(args[0],obj.printmethod)(obj,*args[1:],**kwargs)
            method = self.dispatcher.dispatch(args[0].__class__)
            return method.__get__(obj, cls)(*args, **kwargs)

        _method.__isabstractmethod__ = self.__isabstractmethod__
        _method.register = self.register
        update_wrapper(_method, self.func)
        return _method
```

ThePauliPrinciple:
I started a new issue on this topic: https://github.com/sympy/sympy/issues/22387. Considering using single dispatch comes with its own (unresolved) issues, I would like to continue discussion on this PR though

ThePauliPrinciple:
Would `Codifier` and wrapper function `codify` make sense as the name here? Along the same lines of `lambdify`, `sympify` and `simplify`, this changes a sympy object into some symbolic code object after all. 

This would also better define the scope of this project. Not reusing the printing system but going for a dispatch system makes sense then as well (all problems I see related to dispatchers are for using it for the printing system, which doesn't need to be solved here). 

I would prefer something along the lines of a meta class automatically initializing dispatchers and decorating methods based on some naming scheme. (Only for which methods belong to the same dispatcher, the dispatching itself would be based on type annotation).

ThePauliPrinciple:
I don't think this is the right idea.

PR #22366, opened 2021-10-25
fix _eval_transpose for matrices

danielalcalde:
Fix for _eval_transpose for Matrices, so that the transposition is applied recursievely.

If you want to transpose a Matrix with quantum operators inside you would expect the following behavior:

but currently this is the behaviour:

#### Release Notes

* matrices
  * Transposition is now applied recursievely.


oscarbenjamin:
It would be better to open an issue first to discuss what the problem is and how it could be fixed.

You should provide the code needed to reproduce the problem.

PR #22373, opened 2021-10-26
series - added test for issue #9917

### LINKED ISSUE #9917: O(n*sin(n) + 1, (n, oo)) returns O(n*sin(n), (n, oo))
```
>>> O(n*sin(n) + 1, (n, oo))
O(n*sin(n), (n, oo))
```

rather the expression should be returned un-evaluated as `O(n*sin(n) + 1, (n, oo))`

praneethratna:
Fixes #9917 by adding a test function test_issue_9917 in test_order.py

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/series/tests/test_order.py
+++ b/sympy/series/tests/test_order.py
@@ -450,3 +450,7 @@ def test_issue_18606():
 
 def test_issue_22165():
     assert O(log(x)).contains(2)
+
+
+def test_issue_9917():
+    assert O(x*sin(x) + 1, (x, oo)) == O(x, (x, oo))


anutosh491:
Shouldn't we still prefer `self` here , I not convinced to the fullest here . Considering Order by fundamentally how its defined, which is growth of functions if we just plot `x*sin(x) + 100` and compare it with `x` as x tends to `oo`, we can see
image

Which shows that `x*sin(x) + 100` will be dominating `x` by quite some margin periodically , so I would rather say we should have returned `self` here. But now on master we would have the following
```
>>> O(x*sin(x) + 1000, (x, oo))
O(x, (x, oo))
```
 Maybe `O(x*sin(x) , (x, oo))` could have been `O(x, (x, oo))` as it's bounded in all cases (still returning `self` here isn't a bad option either)

image

@smichr what are your thoughts here ?

Issue #21202, opened 2021-03-30
laplace_transform(cosh(2*x), x, s) raises RecursionError


MaqOwais:
```
>>> from sympy import *
>>> from sympy.abc import x,y,z,s
>>> laplace_transform(cosh(x), x, s)
(s/(s**2 - 1), 1, True)

>>> laplace_transform(cosh(2*x), x, s)
RecursionError: maximum recursion depth exceeded while calling a Python object
```

jksuom:
This piece of code looks incorrect: https://github.com/sympy/sympy/blob/f9badb21b01f4f52ce4d545d071086ee650cd282/sympy/functions/elementary/piecewise.py#L179-L183
It should be removed.

hanspi42:
Laplace transform of `cosh` is now implemented properly in #22376. Comments welcome.

oscarbenjamin:
Since #22376 was merged this issue is no longer seen when calling `laplace_transform`. The underlying bug with `simplify` is still seen though:
```python
In [5]: expr = parse_expr(
   ...:     "Piecewise((s*exp_polar(I*pi)/(4*(1 - s**2/4)), Abs(s**2)/4 < 1), (1/(s*(1 - 4/s**2)), 4/
   ...: Abs(s**2) < 1), (pi*meijerg(((1/2,), (0, 0)), ((0, 1/2), (0,)), s**2/4)/2, True))"
   ...: )

In [6]: expr
Out[6]: 
⎧              ⅈ⋅π                  │ 2│    
⎪           s⋅ℯ                     │s │    
⎪           ──────              for ──── < 1
⎪                2                   4      
⎪           4 - s                           
⎪                                           
⎪             1                      4      
⎪         ──────────            for ──── < 1
⎪           ⎛    4 ⎞                │ 2│    
⎨         s⋅⎜1 - ──⎟                │s │    
⎪           ⎜     2⎟                        
⎪           ⎝    s ⎠                        
⎪                                           
⎪          ⎛             │  2⎞              
⎪  ╭─╮2, 1 ⎜ 1/2    0, 0 │ s ⎟              
⎪π⋅│╶┐     ⎜             │ ──⎟              
⎪  ╰─╯3, 3 ⎝0, 1/2   0   │ 4 ⎠              
⎪─────────────────────────────   otherwise  
⎩              2                            

In [7]: simplify(expr)
...
RecursionError: maximum recursion depth exceeded while calling a Python object
```
The infinite recursion seems to be here:
[...]

Issue #7219, opened 2013-12-04
Make Laplace transform work on undefined functions


sympy-issue-migrator:
Example:
```
In [2]: x = Function('x')

In [3]: from sympy.abc import s

In [4]: laplace_transform(x(t).diff(), t, s)
Out[4]: LaplaceTransform(Derivative(x(t), t), t, s)
```
I expected `s * x(s)`.

The inverse doesn't work either:
```
In [5]: inverse_laplace_transform(s * x(s), s, t)
Out[5]: InverseLaplaceTransform(s*x(s), s, t, _None)
```
Aaron Meurer suggested either to implement a table lookup for some simple rules and properties or to just make the Integral work.

Original issue for #7219: http://code.google.com/p/sympy/issues/detail?id=4120
Original author: https://code.google.com/u/104623634006529252977/

eric-wieser:
Here's a crappy implementation that catches the simple cases correctly (ignoring initial conditions):

```python
def laplace(e, t, s):
    res = laplace_transform(e, t, s, noconds=True)
    wf = Wild('f')
    lw = LaplaceTransform(wf, t, s)

    
    for exp in res.find(lw):
        e = exp.match(lw)[wf]
        
        if not isinstance(e, Derivative):
            continue
            
        if all(arg == t for arg in e.args[1:]):
            newexp = s**len(e.args[1:]) * LaplaceTransform(e.args[0], t, s)
            res = res.replace(exp, newexp)
            
    return res
```

awa5114:
Has this been fixed yet?

physici:
Any news here? Have the same problem that I need to calculate the Laplace transformation of an unknown function.

adeak:
I updated Eric's ["crappy implementation"](https://github.com/sympy/sympy/issues/7219#issuecomment-154768904) for the corresponding question on Stack Overflow. It seems to me that the representation of `Derivative`s has changed, and that we can shoehorn in a hack for `Integral`s similarly. Here's my updated workaround for completeness' sake:
```
from sympy import *

def laplace(e, t, s):
    res = laplace_transform(e, t, s, noconds=True)
    wf = Wild('f')
    lw = LaplaceTransform(wf, t, s)

    for exp in res.find(lw):
        e = exp.match(lw)[wf]
        args = e.args

        if isinstance(e, Derivative):
            # for derivative check that there's only d/dt^n with n>0
            if len(args) == 2 and args[1][0] == t:
                n = args[1][1]
[...]

Issue #7176, opened 2013-10-29
laplace_transform(diff(y(t), t, 2), t, s) not calculated


ppuedom:
```
t = symbols('t', real=True, positive=True)
w = symbols('w', real=True)
s = symbols('s')
y = Function('y')
laplace_transform(diff(y(t), t, 2), t, s).doit()

Does not calculate transform and return:
LaplaceTransform(Derivative(y(t), t, t), t, s)
```

Original issue for #7176: http://code.google.com/p/sympy/issues/detail?id=4077
Original author: https://code.google.com/u/118279007905392951376/

hanspi42:
Laplace transform of diff is now implemented in #22376. Comments welcome.

hanspi42:
@oscarbenjamin I think this can be closed.

oscarbenjamin:
Current output is
```python
In [5]: t = symbols('t', real=True, positive=True)
   ...: w = symbols('w', real=True)
   ...: s = symbols('s')
   ...: y = Function('y')
   ...: laplace_transform(diff(y(t), t, 2), t, s).doit()
Out[5]: 
 2                                         ⎛d       ⎞│   
s ⋅LaplaceTransform(y(t), t, s) - s⋅y(0) - ⎜──(y(t))⎟│   
                                           ⎝dt      ⎠│t=0
```

PR #22376, opened 2021-10-28
Making the Laplace Transform Rule-Based

### LINKED ISSUE #5813: Use general rules in integral transforms code
```
for example, laplace_transform(f(t).diff(t), t, x) should give LaplaceTransform(f(t), t, x) - f(0)

This should actually be fairly easy: there is a "TODO" comment in integrals/transforms/transforms.py function IntegralTransform.doit where calls to methods that are overridden in the derived classes can be inserted.
```

Original issue for #5813: http://code.google.com/p/sympy/issues/detail?id=2714
Original author: https://code.google.com/u/104531927090589914088/
Original owner: https://code.google.com/u/104531927090589914088/

hanspi42:

Fixes #5813, #7176, #7219, #21202

Major change in the LaplaceTransform: The Laplace transfom now applies a large number of known rules before it tries to integrate. It can now transform a far wider range of expressions than before, and is now also able to transform differential equations.

#### Release Notes

* integrals
  * The Laplace transform is now rule-based, can transform a far wider range of expressions, and works better with undefined functions, including the ability to transform differential equations.

### DIFF
--- a/sympy/integrals/tests/test_transforms.py
+++ b/sympy/integrals/tests/test_transforms.py
@@ -1,6 +1,6 @@
 from sympy.integrals.transforms import (mellin_transform,
-    inverse_mellin_transform, laplace_transform, inverse_laplace_transform,
-    fourier_transform, inverse_fourier_transform,
+    inverse_mellin_transform, laplace_transform,
+    inverse_laplace_transform, fourier_transform, inverse_fourier_transform,
     sine_transform, inverse_sine_transform,
     cosine_transform, inverse_cosine_transform,
     hankel_transform, inverse_hankel_transform,
@@ -8,20 +8,20 @@ from sympy.integrals.transforms import (mellin_transform,
     InverseLaplaceTransform, InverseFourierTransform,
     InverseSineTransform, InverseCosineTransform, IntegralTransformError)
 from sympy.core.function import (Function, expand_mul)
-from sympy.core import EulerGamma
+from sympy.core import EulerGamma, Subs, Derivative, diff
 from sympy.core.numbers import (I, Rational, oo, pi)
-from sympy.core.relational import Eq, Ne
+from sympy.core.relational import Eq
 from sympy.core.singleton import S
 from sympy.core.symbol import (Symbol, symbols)
 from sympy.functions.combinatorial.factorials import factorial
-from sympy.functions.elementary.complexes import (Abs, arg, re, unpolarify)
+from sympy.functions.elementary.complexes import (Abs, re, unpolarify)
 from sympy.functions.elementary.exponential import (exp, exp_polar, log)
-from sympy.functions.elementary.hyperbolic import (cosh, sinh)
+from sympy.functions.elementary.hyperbolic import (cosh, sinh, coth, asinh)
 from sympy.functions.elementary.miscellaneous import sqrt
 from sympy.functions.elementary.trigonometric import (atan, atan2, cos, sin, tan)
 from sympy.functions.special.bessel import (besseli, besselj, besselk, bessely)
 from sympy.functions.special.delta_functions import Heaviside
-from sympy.functions.special.error_functions import (erf, erfc, expint)
+from sympy.functions.special.error_functions import (erf, erfc, expint, Ei)
 from sympy.functions.special.gamma_functions import gamma
 from sympy.functions.special.hyper import meijerg
 from sympy.simplify.gammasimp import gammasimp
@@ -38,10 +38,7 @@ def test_undefined_function():
     f = Function('f')
     assert mellin_transform(f(x), x, s) == MellinTransform(f(x), x, s)
     assert mellin_transform(f(x) + exp(-x), x, s) == \
-        (MellinTransform(f(x), x, s) + gamma(s), (0, oo), True)
-
-    assert laplace_transform(2*f(x), x, s) == 2*LaplaceTransform(f(x), x, s)
-    # TODO test derivative and other rules when implemented
+        (MellinTransform(f(x), x, s) + gamma(s + 1)/s, (0, oo), True)
 
 
 def test_free_symbols():
@@ -468,71 +465,263 @@ def test_inverse_mellin_transform():
 
 @slow
 def test_laplace_transform():
+    from sympy import lowergamma
     from sympy.functions.special.delta_functions import DiracDelta
     from sympy.functions.special.error_functions import (fresnelc, fresnels)
     LT = laplace_transform
-    a, b, c, = symbols('a b c', positive=True)
-    t = symbols('t')
-    w = Symbol("w")
+    a, b, c, = symbols('a, b, c', positive=True)
+    t, w, x = symbols('t, w, x')
     f = Function("f")
+    g = Function("g")
 
-    # Test unevaluated form
-    assert laplace_transform(f(t), t, w) == LaplaceTransform(f(t), t, w)
+    # Test rule-base evaluation according to
+    # http://eqworld.ipmnet.ru/en/auxiliary/inttrans/
+    # Power-law functions (laplace2.pdf)
+    assert LT(a*t+t**2+t**(S(5)/2), t, s) ==\
+        (a/s**2 + 2/s**3 + 15*sqrt(pi)/(8*s**(S(7)/2)), 0, True)
+    assert LT(b/(t+a), t, s) == (-b*exp(-a*s)*Ei(-a*s), 0, True)
+    assert LT(1/sqrt(t+a), t, s) ==\
+        (sqrt(pi)*sqrt(1/s)*exp(a*s)*erfc(sqrt(a)*sqrt(s)), 0, True)
+    assert LT(sqrt(t)/(t+a), t, s) ==\
+        (-pi*sqrt(a)*exp(a*s)*erfc(sqrt(a)*sqrt(s)) + sqrt(pi)*sqrt(1/s),
+         0, True)
+    assert LT((t+a)**(-S(3)/2), t, s) ==\
+        (-2*sqrt(pi)*sqrt(s)*exp(a*s)*erfc(sqrt(a)*sqrt(s)) + 2/sqrt(a),
+         0, True)
+    assert LT(t**(S(1)/2)*(t+a)**(-1), t, s) ==\
+        (-pi*sqrt(a)*exp(a*s)*erfc(sqrt(a)*sqrt(s)) + sqrt(pi)*sqrt(1/s),
+         0, True)
+    assert LT(1/(a*sqrt(t) + t**(3/2)), t, s) ==\
+        (pi*sqrt(a)*exp(a*s)*erfc(sqrt(a)*sqrt(s)), 0, True)
+    assert LT((t+a)**b, t, s) ==\
+        (s**(-b - 1)*exp(-a*s)*lowergamma(b + 1, a*s), 0, True)
+    assert LT(t**5/(t+a), t, s) == (120*a**5*lowergamma(-5, a*s), 0, True)
+    # Exponential functions (laplace3.pdf)
+    assert LT(exp(t), t, s) == (1/(s - 1), 1, True)
+    assert LT(exp(2*t), t, s) == (1/(s - 2), 2, True)
+    assert LT(exp(a*t), t, s) == (1/(s - a), a, True)
+    assert LT(exp(a*(t-b)), t, s) == (exp(-a*b)/(-a + s), a, True)
+    assert LT(t*exp(-a*(t)), t, s) == ((a + s)**(-2), -a, True)
+    assert LT(t*exp(-a*(t-b)), t, s) == (exp(a*b)/(a + s)**2, -a, True)
+    assert LT(b*t*exp(-a*t), t, s) == (b/(a + s)**2, -a, True)
+    assert LT(t**(S(7)/4)*exp(-8*t)/gamma(S(11)/4), t, s) ==\
+        ((s + 8)**(-S(11)/4), -8, True)
+    assert LT(t**(S(3)/2)*exp(-8*t), t, s) ==\
+        (3*sqrt(pi)/(4*(s + 8)**(S(5)/2)), -8, True)
+    assert LT(t**a*exp(-a*t), t, s) ==  ((a+s)**(-a-1)*gamma(a+1), -a, True)
+    assert LT(b*exp(-a*t**2), t, s) ==\
+        (sqrt(pi)*b*exp(s**2/(4*a))*erfc(s/(2*sqrt(a)))/(2*sqrt(a)), 0, True)
+    assert LT(exp(-2*t**2), t, s) ==\
+        (sqrt(2)*sqrt(pi)*exp(s**2/8)*erfc(sqrt(2)*s/4)/4, 0, True)
+    assert LT(b*exp(2*t**2), t, s) == b*LaplaceTransform(exp(2*t**2), t, s)
+    assert LT(t*exp(-a*t**2), t, s) ==\
+        (1/(2*a) - s*erfc(s/(2*sqrt(a)))/(4*sqrt(pi)*a**(S(3)/2)), 0, True)
+    assert LT(exp(-a/t), t, s) ==\
+        (2*sqrt(a)*sqrt(1/s)*besselk(1, 2*sqrt(a)*sqrt(s)), 0, True)
+    assert LT(sqrt(t)*exp(-a/t), t, s) ==\
+        (sqrt(pi)*(2*sqrt(a)*sqrt(s) + 1)*sqrt(s**(-3))*exp(-2*sqrt(a)*\
+                                                    sqrt(s))/2, 0, True)
+    assert LT(exp(-a/t)/sqrt(t), t, s) ==\
+        (sqrt(pi)*sqrt(1/s)*exp(-2*sqrt(a)*sqrt(s)), 0, True)
+    assert LT( exp(-a/t)/(t*sqrt(t)), t, s) ==\
+        (sqrt(pi)*sqrt(1/a)*exp(-2*sqrt(a)*sqrt(s)), 0, True)
+    assert LT(exp(-2*sqrt(a*t)), t, s) ==\
+        ( 1/s -sqrt(pi)*sqrt(a) * exp(a/s)*erfc(sqrt(a)*sqrt(1/s))/\
+         s**(S(3)/2), 0, True)
+    assert LT(exp(-2*sqrt(a*t))/sqrt(t), t, s) == (exp(a/s)*erfc(sqrt(a)*\
+        sqrt(1/s))*(sqrt(pi)*sqrt(1/s)), 0, True)
+    assert LT(t**4*exp(-2/t), t, s) ==\
+        (8*sqrt(2)*(1/s)**(S(5)/2)*besselk(5, 2*sqrt(2)*sqrt(s)), 0, True)
+    # Hyperbolic functions (laplace4.pdf)
+    assert LT(sinh(a*t), t, s) == (a/(-a**2 + s**2), a, True)
+    assert LT(b*sinh(a*t)**2, t, s) == (2*a**2*b/(-4*a**2*s**2 + s**3),
+                                        2*a, True)
+    # The following line confirms that issue #21202 is solved
+    assert LT(cosh(2*t), t, s) == (s/(-4 + s**2), 2, True)
+    assert LT(cosh(a*t), t, s) == (s/(-a**2 + s**2), a, True)
+    assert LT(cosh(a*t)**2, t, s) == ((-2*a**2 + s**2)/(-4*a**2*s**2 + s**3),
+                                      2*a, True)
+    assert LT(sinh(x + 3), x, s) == (
+        (-s + (s + 1)*exp(6) + 1)*exp(-3)/(s - 1)/(s + 1)/2, 0, Abs(s) > 1)
+    # The following line replaces the old test test_issue_7173()
+    assert LT(sinh(a*t)*cosh(a*t), t, s) == (a/(-4*a**2 + s**2), 2*a, True)
+    assert LT(sinh(a*t)/t, t, s) == (log((a + s)/(-a + s))/2, a, True)
+    assert LT(t**(-S(3)/2)*sinh(a*t), t, s) ==\
+        (-sqrt(pi)*(sqrt(-a + s) - sqrt(a + s)), a, True)
+    assert LT(sinh(2*sqrt(a*t)), t, s) ==\
+        (sqrt(pi)*sqrt(a)*exp(a/s)/s**(S(3)/2), 0, True)
+    assert LT(sqrt(t)*sinh(2*sqrt(a*t)), t, s) ==\
+        (-sqrt(a)/s**2 + sqrt(pi)*(a + s/2)*exp(a/s)*erf(sqrt(a)*\
+                                            sqrt(1/s))/s**(S(5)/2), 0, True)
+    assert LT(sinh(2*sqrt(a*t))/sqrt(t), t, s) ==\
+        (sqrt(pi)*exp(a/s)*erf(sqrt(a)*sqrt(1/s))/sqrt(s), 0, True)
+    assert LT(sinh(sqrt(a*t))**2/sqrt(t), t, s) ==\
+        (sqrt(pi)*(exp(a/s) - 1)/(2*sqrt(s)), 0, True)
+    assert LT(t**(S(3)/7)*cosh(a*t), t, s) ==\
+        (((a + s)**(-S(10)/7) + (-a+s)**(-S(10)/7))*gamma(S(10)/7)/2, a, True)
+    assert LT(cosh(2*sqrt(a*t)), t, s) ==\
+        (sqrt(pi)*sqrt(a)*exp(a/s)*erf(sqrt(a)*sqrt(1/s))/s**(S(3)/2) + 1/s,
+         0, True)
+    assert LT(sqrt(t)*cosh(2*sqrt(a*t)), t, s) ==\
+        (sqrt(pi)*(a + s/2)*exp(a/s)/s**(S(5)/2), 0, True)
+    assert LT(cosh(2*sqrt(a*t))/sqrt(t), t, s) ==\
+        (sqrt(pi)*exp(a/s)/sqrt(s), 0, True)
+    assert LT(cosh(sqrt(a*t))**2/sqrt(t), t, s) ==\
+        (sqrt(pi)*(exp(a/s) + 1)/(2*sqrt(s)), 0, True)
+    # logarithmic functions (laplace5.pdf)
+    assert LT(log(t), t, s) == (-log(s+S.EulerGamma)/s, 0, True)
+    assert LT(log(t/a), t, s) == (-log(a*s + S.EulerGamma)/s, 0, True)
+    assert LT(log(1+a*t), t, s) == (-exp(s/a)*Ei(-s/a)/s, 0, True)
+    assert LT(log(t+a), t, s) == ((log(a) - exp(s/a)*Ei(-s/a)/s)/s, 0, True)
+    assert LT(log(t)/sqrt(t), t, s) ==\
+        (sqrt(pi)*(-log(s) - 2*log(2) - S.EulerGamma)/sqrt(s), 0, True)
+    assert LT(t**(S(5)/2)*log(t), t, s) ==\
+        (15*sqrt(pi)*(-log(s)-2*log(2)-S.EulerGamma+S(46)/15)/(8*s**(S(7)/2)),
+         0, True)
+    assert (LT(t**3*log(t), t, s, noconds=True)-6*(-log(s) - S.EulerGamma\
+                                    + S(11)/6)/s**4).simplify() == S.Zero
+    assert LT(log(t)**2, t, s) ==\
+        (((log(s) + EulerGamma)**2 + pi**2/6)/s, 0, True)
+    assert LT(exp(-a*t)*log(t), t, s) ==\
+        ((-log(a + s) - S.EulerGamma)/(a + s), -a, True)
+    # Trigonometric functions (laplace6.pdf)
+    assert LT(sin(a*t), t, s) == (a/(a**2 + s**2), 0, True)
+    assert LT(Abs(sin(a*t)), t, s) ==\
+        (a*coth(pi*s/(2*a))/(a**2 + s**2), 0, True)
+    assert LT(sin(a*t)/t, t, s) == (atan(a/s), 0, True)
+    assert LT(sin(a*t)**2/t, t, s) == (log(4*a**2/s**2 + 1)/4, 0, True)
+    assert LT(sin(a*t)**2/t**2, t, s) ==\
+        (a*atan(2*a/s) - s*log(4*a**2/s**2 + 1)/4, 0, True)
+    assert LT(sin(2*sqrt(a*t)), t, s) ==\
+        (sqrt(pi)*sqrt(a)*exp(-a/s)/s**(S(3)/2), 0, True)
+    assert LT(sin(2*sqrt(a*t))/t, t, s) == (pi*erf(sqrt(a)*sqrt(1/s)), 0, True)
+    assert LT(cos(a*t), t, s) == (s/(a**2 + s**2), 0, True)
+    assert LT(cos(a*t)**2, t, s) ==\
+        ((2*a**2 + s**2)/(s*(4*a**2 + s**2)), 0, True)
+    assert LT(sqrt(t)*cos(2*sqrt(a*t)), t, s) ==\
+        (sqrt(pi)*(-2*a + s)*exp(-a/s)/(2*s**(S(5)/2)), 0, True)
+    assert LT(cos(2*sqrt(a*t))/sqrt(t), t, s) ==\
+        (sqrt(pi)*sqrt(1/s)*exp(-a/s), 0, True)
+    assert LT(sin(a*t)*sin(b*t), t, s) ==\
+        (2*a*b*s/((s**2 + (a - b)**2)*(s**2 + (a + b)**2)), 0, True)
+    assert LT(cos(a*t)*sin(b*t), t, s) ==\
+        (b*(-a**2 + b**2 + s**2)/((s**2 + (a - b)**2)*(s**2 + (a + b)**2)),
+         0, True)
+    assert LT(cos(a*t)*cos(b*t), t, s) ==\
+        (s*(a**2 + b**2 + s**2)/((s**2 + (a - b)**2)*(s**2 + (a + b)**2)),
+         0, True)
+    assert LT(c*exp(-b*t)*sin(a*t), t, s) == (a*c/(a**2 + (b + s)**2),
+                                              -b, True)
+    assert LT(c*exp(-b*t)*cos(a*t), t, s) == ((b + s)*c/(a**2 + (b + s)**2),
+                                              -b, True)
+    assert LT(cos(x + 3), x, s) == ((s*cos(3) - sin(3))/(s**2 + 1), 0, True)
+    # Error functions (laplace7.pdf)
+    assert LT(erf(a*t), t, s) == (exp(s**2/(4*a**2))*erfc(s/(2*a))/s, 0, True)
+    assert LT(erf(sqrt(a*t)), t, s) == (sqrt(a)/(s*sqrt(a + s)), 0, True)
+    assert LT(exp(a*t)*erf(sqrt(a*t)), t, s) ==\
+        (sqrt(a)/(sqrt(s)*(-a + s)), a, True)
+    assert LT(erf(sqrt(a/t)/2), t, s) == ((1-exp(-sqrt(a)*sqrt(s)))/s, 0, True)
+    assert LT(erfc(sqrt(a*t)), t, s) ==\
+        ((-sqrt(a) + sqrt(a + s))/(s*sqrt(a + s)), 0, True)
+    assert LT(exp(a*t)*erfc(sqrt(a*t)), t, s) ==\
+        (1/(sqrt(a)*sqrt(s) + s), 0, True)
+    assert LT(erfc(sqrt(a/t)/2), t, s) == (exp(-sqrt(a)*sqrt(s))/s, 0, True)
+    # Bessel functions (laplace8.pdf)
+    assert LT(besselj(0, a*t), t, s) == (1/sqrt(a**2 + s**2), 0, True)
+    assert LT(besselj(1, a*t), t, s) ==\
+        (a/(sqrt(a**2 + s**2)*(s + sqrt(a**2 + s**2))), 0, True)
+    assert LT(besselj(2, a*t), t, s) ==\
+        (a**2/(sqrt(a**2 + s**2)*(s + sqrt(a**2 + s**2))**2), 0, True)
+    assert LT(t*besselj(0, a*t), t, s) ==\
+        (s/(a**2 + s**2)**(S(3)/2), 0, True)
+    assert LT(t*besselj(1, a*t), t, s) ==\
+        (a/(a**2 + s**2)**(S(3)/2), 0, True)
+    assert LT(t**2*besselj(2, a*t), t, s) ==\
+        (3*a**2/(a**2 + s**2)**(S(5)/2), 0, True)
+    assert LT(besselj(0, 2*sqrt(a*t)), t, s) == (exp(-a/s)/s, 0, True)
+    assert LT(t**(S(3)/2)*besselj(3, 2*sqrt(a*t)), t, s) ==\
+        (a**(S(3)/2)*exp(-a/s)/s**4, 0, True)
+    assert LT(besselj(0, a*sqrt(t**2+b*t)), t, s) ==\
+        (exp(b*s - b*sqrt(a**2 + s**2))/sqrt(a**2 + s**2), 0, True)
+    assert LT(besseli(0, a*t), t, s) == (1/sqrt(-a**2 + s**2), a, True)
+    assert LT(besseli(1, a*t), t, s) ==\
+        (a/(sqrt(-a**2 + s**2)*(s + sqrt(-a**2 + s**2))), a, True)
+    assert LT(besseli(2, a*t), t, s) ==\
+        (a**2/(sqrt(-a**2 + s**2)*(s + sqrt(-a**2 + s**2))**2), a, True)
+    assert LT(t*besseli(0, a*t), t, s) == (s/(-a**2 + s**2)**(S(3)/2), a, True)
+    assert LT(t*besseli(1, a*t), t, s) == (a/(-a**2 + s**2)**(S(3)/2), a, True)
+    assert LT(t**2*besseli(2, a*t), t, s) ==\
+        (3*a**2/(-a**2 + s**2)**(S(5)/2), a, True)
+    assert LT(t**(S(3)/2)*besseli(3, 2*sqrt(a*t)), t, s) ==\
+        (a**(S(3)/2)*exp(a/s)/s**4, 0, True)
+    assert LT(bessely(0, a*t), t, s) ==\
+        (-2*asinh(s/a)/(pi*sqrt(a**2 + s**2)), 0, True)
+    assert LT(besselk(0, a*t), t, s) ==\
+        (log(s + sqrt(-a**2 + s**2))/sqrt(-a**2 + s**2), a, True)
+    assert LT(sin(a*t)**8, t, s) ==\
+        (40320*a**8/(s*(147456*a**8 + 52480*a**6*s**2 + 4368*a**4*s**4 +\
+                        120*a**2*s**6 + s**8)), 0, True)
+
+    # Test general rules and unevaluated forms
+    # These all also test whether issue #7219 is solved.
+    assert LT(Heaviside(t-1)*cos(t-1), t, s) == (s*exp(-s)/(s**2 + 1), 0, True)
+    assert LT(a*f(t), t, w) == a*LaplaceTransform(f(t), t, w)
+    assert LT(a*Heaviside(t+1)*f(t+1), t, s) ==\
+        a*LaplaceTransform(f(t + 1)*Heaviside(t + 1), t, s)
+    assert LT(a*Heaviside(t-1)*f(t-1), t, s) ==\
+        a*LaplaceTransform(f(t), t, s)*exp(-s)
+    assert LT(b*f(t/a), t, s) == a*b*LaplaceTransform(f(t), t, a*s)
+    assert LT(exp(-f(x)*t), t, s) == (1/(s + f(x)), -f(x), True)
+    assert LT(exp(-a*t)*f(t), t, s) == LaplaceTransform(f(t), t, a + s)
+    assert LT(exp(-a*t)*erfc(sqrt(b/t)/2), t, s) ==\
+        (exp(-sqrt(b)*sqrt(a + s))/(a + s), -a, True)
+    assert LT(sinh(a*t)*f(t), t, s) ==\
+        LaplaceTransform(f(t), t, -a+s)/2 - LaplaceTransform(f(t), t, a+s)/2
+    assert LT(sinh(a*t)*t, t, s) ==\
+        (-1/(2*(a + s)**2) + 1/(2*(-a + s)**2), a, True)
+    assert LT(cosh(a*t)*f(t), t, s) ==\
+        LaplaceTransform(f(t), t, -a+s)/2 + LaplaceTransform(f(t), t, a+s)/2
+    assert LT(cosh(a*t)*t, t, s) ==\
+        (1/(2*(a + s)**2) + 1/(2*(-a + s)**2), a, True)
+    assert LT(sin(a*t)*f(t), t, s) ==\
+        I*(-LaplaceTransform(f(t), t, -I*a + s) +\
+           LaplaceTransform(f(t), t, I*a + s))/2
+    assert LT(sin(a*t)*t, t, s) ==\
+        (2*a*s/(a**4 + 2*a**2*s**2 + s**4), 0, True)
+    assert LT(cos(a*t)*f(t), t, s) ==\
+        LaplaceTransform(f(t), t, -I*a + s)/2 +\
+        LaplaceTransform(f(t), t, I*a + s)/2
+    assert LT(cos(a*t)*t, t, s) ==\
+        ((-a**2 + s**2)/(a**4 + 2*a**2*s**2 + s**4), 0, True)
+    # The following two lines test whether issues #5813 and #7176 are solved.
+    assert LT(diff(f(t), (t, 1)), t, s) == s*LaplaceTransform(f(t), t, s)\
+        - f(0)
+    assert LT(diff(f(t), (t, 3)), t, s) == s**3*LaplaceTransform(f(t), t, s)\
+        - s**2*f(0) - s*Subs(Derivative(f(t), t), t, 0)\
+            - Subs(Derivative(f(t), (t, 2)), t, 0)
+    assert LT(a*f(b*t)+g(c*t), t, s) == a*LaplaceTransform(f(t), t, s/b)/b +\
+        LaplaceTransform(g(t), t, s/c)/c
     assert inverse_laplace_transform(
         f(w), w, t, plane=0) == InverseLaplaceTransform(f(w), w, t, 0)
+    assert LT(f(t)*g(t), t, s) == LaplaceTransform(f(t)*g(t), t, s)
 
-    # test a bug
-    spos = symbols('s', positive=True)
-    assert LT(exp(t), t, spos) == (1/(spos - 1), 0, spos > 1)
-
-    # basic tests from wikipedia
+    # additional basic tests from wikipedia
     assert LT((t - a)**b*exp(-c*(t - a))*Heaviside(t - a), t, s) == \
         ((s + c)**(-b - 1)*exp(-a*s)*gamma(b + 1), -c, True)
-    assert LT(t**a, t, s) == (s**(-a - 1)*gamma(a + 1), 0, True)
-    assert LT(Heaviside(t), t, s) == (1/s, 0, True)
-    assert LT(Heaviside(t - a), t, s) == (exp(-a*s)/s, 0, True)
-    assert LT(1 - exp(-a*t), t, s) == (a/(s*(a + s)), 0, True)
-
     assert LT((exp(2*t) - 1)*exp(-b - t)*Heaviside(t)/2, t, s, noconds=True) \
         == exp(-b)/(s**2 - 1)
 
-    assert LT(exp(t), t, s) == (1/(s - 1), 0, abs(s) > 1)
-    assert LT(exp(2*t), t, s) == (1/(s - 2), 0, abs(s) > 2)
-    assert LT(exp(a*t), t, s) == (1/(s - a), a, Ne(s/a, 1))
-
-    assert LT(log(t/a), t, s) == (-(log(a*s) + EulerGamma)/s, 0, True)
-
-    assert LT(erf(t), t, s) == (erfc(s/2)*exp(s**2/4)/s, 0, True)
-
-    assert LT(sin(a*t), t, s) == (a/(a**2 + s**2), 0, True)
-    assert LT(cos(a*t), t, s) == (s/(a**2 + s**2), 0, True)
-    # TODO would be nice to have these come out better
-    assert LT(exp(-a*t)*sin(b*t), t, s) == (b/(b**2 + (a + s)**2), -a, True)
-    assert LT(exp(-a*t)*cos(b*t), t, s) == \
-        ((a + s)/(b**2 + (a + s)**2), -a, True)
-
-    assert LT(besselj(0, t), t, s) == (1/sqrt(1 + s**2), 0, True)
-    assert LT(besselj(1, t), t, s) == (1 - 1/sqrt(1 + 1/s**2), 0, True)
-    # TODO general order works, but is a *mess*
-    # TODO besseli also works, but is an even greater mess
-
-    # test a bug in conditions processing
-    # TODO the auxiliary condition should be recognised/simplified
-    assert LT(exp(t)*cos(t), t, s)[:-1] in [
-        ((s - 1)/(s**2 - 2*s + 2), -oo),
-        ((s - 1)/((s - 1)**2 + 1), -oo),
-    ]
-
     # DiracDelta function: standard cases
-    assert LT(DiracDelta(t), t, s) == (1, -oo, True)
-    assert LT(DiracDelta(a*t), t, s) == (1/a, -oo, True)
-    assert LT(DiracDelta(t/42), t, s) == (42, -oo, True)
-    assert LT(DiracDelta(t+42), t, s) == (0, -oo, True)
+    assert LT(DiracDelta(t), t, s) == (1, 0, True)
+    assert LT(DiracDelta(a*t), t, s) == (1/a, 0, True)
+    assert LT(DiracDelta(t/42), t, s) == (42, 0, True)
+    assert LT(DiracDelta(t+42), t, s) == (0, 0, True)
     assert LT(DiracDelta(t)+DiracDelta(t-42), t, s) == \
-        (1 + exp(-42*s), -oo, True)
-    assert LT(DiracDelta(t)-a*exp(-a*t), t, s) == (-a/(a + s) + 1, 0, True)
+        (1 + exp(-42*s), 0, True)
+    assert LT(DiracDelta(t)-a*exp(-a*t), t, s) == (s/(a + s), 0, True)
     assert LT(exp(-t)*(DiracDelta(t)+DiracDelta(t-42)), t, s) == \
         (exp(-42*s - 42) + 1, -oo, True)
+
     # Collection of cases that cannot be fully evaluated and/or would catch
     # some common implementation errors
     assert LT(DiracDelta(t**2), t, s) == LaplaceTransform(DiracDelta(t**2), t, s)
@@ -542,8 +731,17 @@ def test_laplace_transform():
     assert LT((DiracDelta(t) + 1)*(DiracDelta(t - 1) + 1), t, s) == \
         (LaplaceTransform(DiracDelta(t)*DiracDelta(t - 1), t, s) + \
          1 + exp(-s) + 1/s, 0, True)
-    assert LT(DiracDelta(2*t - 2*exp(a)), t, s) == \
-        (exp(-s*exp(a))/2, -oo, True)
+    assert LT(DiracDelta(2*t-2*exp(a)), t, s) == (exp(-s*exp(a))/2, 0, True)
+    assert LT(DiracDelta(-2*t+2*exp(a)), t, s) == (exp(-s*exp(a))/2, 0, True)
+
+    # Heaviside tests
+    assert LT(Heaviside(t), t, s) == (1/s, 0, True)
+    assert LT(Heaviside(t - a), t, s) == (exp(-a*s)/s, 0, True)
+    assert LT(Heaviside(t-1), t, s) == (exp(-s)/s, 0, True)
+    assert LT(Heaviside(2*t-4), t, s) == (exp(-2*s)/s, 0, True)
+    assert LT(Heaviside(-2*t+4), t, s) == ((1 - exp(-2*s))/s, 0, True)
+    assert LT(Heaviside(2*t+4), t, s) == (1/s, 0, True)
+    assert LT(Heaviside(-2*t+4), t, s) == ((1 - exp(-2*s))/s, 0, True)
 
     # Fresnel functions
     assert laplace_transform(fresnels(t), t, s) == \
@@ -553,6 +751,7 @@ def test_laplace_transform():
         ((2*sin(s**2/(2*pi))*fresnelc(s/pi) - 2*cos(s**2/(2*pi))*fresnels(s/pi)
         + sqrt(2)*cos(s**2/(2*pi) + pi/4))/(2*s), 0, True))
 
+    # Matrix tests
     Mt = Matrix([[exp(t), t*exp(-t)], [t*exp(-t), exp(t)]])
     Ms = Matrix([[    1/(s - 1), (s + 1)**(-2)],
                  [(s + 1)**(-2),     1/(s - 1)]])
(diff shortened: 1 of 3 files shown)


hanspi42:
> The easiest is probably to define a doit method in LaplaceTransform, run the code you want and then, if needed, call super().doit() (or whatever the correct syntax is).

> An option may be to check isinstance(self, LaplaceTransform), but I do not know if it works. (May be even easier if it does.)

Afer many experiments and thoughts, I have now switched to writing an independent `doit()` for the Laplace transform. The order "try directly, then split into sum" is not optimum for Laplace, I have changed it to "split into sum, then work on sum terms", and it already gives better results for many Laplace transforms.

My next steps are:
- time shift rule
- differentiation rule

For differentiation of an undefined function `f(t)`, I will need to return an unevaluated expression that means "the n-th derivative evaluated at t=0". I don't know how to best do this, does anyone have am idea?  Would returning `Derivative(f(t), (t, n)).subs(t,0)` be OK?

oscarbenjamin:
> Would returning `Derivative(f(t), (t, n)).subs(t,0)` be OK?

Yes, that's basically the only way to represent a derivative evaluated at a point.

oscarbenjamin:
I think this was previously marked as a draft. Is this ready for review now?

hanspi42:
> I think this was previously marked as a draft. Is this ready for review now?

Thanks for asking, it is not ready yet. I will tell you here when I think it is at a stage where a review makes sense.

hanspi42:
@oscarbenjamin: it is now ready for a first review. Do you think this way of writing a `doit()` is OK? If you say yes, I will next implement the derivatives of undefined functions.

oscarbenjamin:
> Do you think this way of writing a `doit()` is OK?

I just looked quickly but I think it looks okay. I would prefer to implement the bulk (or all) of the logic in a separate function that is called by doit rather than in the doit function itself but otherwise this seems fine.

hanspi42:
> 
> 
> > Do you think this way of writing a `doit()` is OK?
> 
> I just looked quickly but I think it looks okay. I would prefer to implement the bulk (or all) of the logic in a separate function that is called by doit rather than in the doit function itself but otherwise this seems fine.

OK, I have changed it. I will put all rules for `AppliedUndef` into the new `_apply_rules_undef` and rules for concrete functions into `_compute_transform`.

hanspi42:
I have now also added a first concrete rule. From here on it should be smooth flying ;)

hanspi42:
I am now dealing with `exp` expressions and have seen that many convergence regions were wrong up to now. I am correcting this, but do we have somebody who'll be able to review that part from a mathematical point of view?

sympy/integrals/transforms.py
```diff
@@ -1150,7 +1155,75 @@ class LaplaceTransform(IntegralTransform):
...
+        omega = Wild('omega', exclude=[t])
+        # Expressions with exponential functions:
+        # http://eqworld.ipmnet.ru/en/auxiliary/inttrans/laplace3.pdf
+        # 3.1
+        ma1 = f.match(b*exp(x))
+        if ma1:
+            ma2 = ma1[x].collect(t).match(a*t-tau)
+            if ma2:
+                return self._cr(ma1[b]*exp(-ma2[tau])/(s-ma2[a]),
+                                ma2[a], S.true, **hints)
```
oscarbenjamin, line 1180: This block of code seems to be repeated many times. Maybe we can just collect all the rules together like:
```python
rules = [
    (b*exp(a*t - tau), b*exp(-tau)/(s - a)),
    ... 
]
```
Then the main code can just do something like:
```python
for pattern, replacement in rules:
     if f.match(pattern)
         ...
```
Also maybe leading constants can be factored out of the pattern at the beginning so it's something like:
```python
for term in Add.make_args(f):
    coeff, func = f.as_independent(t, as_Add=False)
    # Now laplace transform func
```

hanspi42: I tried that first, but matching to `b*exp(a*t - tau)` will not necessarily work, which is why I do a `collect(t)` on `ma1[x]`. Sometimes I also have to check parameter values, I will mark an example.

sympy/integrals/transforms.py
```diff
@@ -1150,7 +1155,83 @@ class LaplaceTransform(IntegralTransform):
...
+        ma1 = f.match(b*t*exp(x))
+        if ma1:
+            ma2 = ma1[x].collect(t).match(a*t-tau)
+            if ma2:
+                return self._cr(ma1[b]*exp(-ma2[tau])/(s-ma2[a])**2,
+                                ma2[a], S.true, **hints)
+        # 3.3
+        ma1 = f.match(b*t**c*exp(a*t))
+        if ma1:
+            if ma1[c]>-1:
```
hanspi42, line 1192: Here an additional check is needed whether the rule can be applied.

hanspi42:
I found a way to introduce checks etc. into the rules. Now it is much shorter.

sympy/integrals/transforms.py
```diff
@@ -1150,7 +1155,37 @@ class LaplaceTransform(IntegralTransform):
...
+            (b*exp(a*t-tau), b*exp(-tau)/(s-a), S.true, a), # 3.1
+            (b*t*exp(a*t-tau), b*exp(-tau)/(s-a)**2, S.true, a), # 3.2
+            (b*t**c*exp(a*t), b*gamma(c+1)/(s-a)**(c+1), c>-1, a), # 3.3
+            (a*sin(omega*t), a*omega/(s**2+omega**2), S.true, S.Zero), #6.1
+            (a*sin(omega*t)/t, a*atan(omega/s), S.true, S.Zero), # 6.6
+            (a*cos(omega*t), a*s/(s**2+omega**2), S.true, S.Zero), # 6.11
+            (c*exp(b*t)*sin(a*t), c*a/((s-b)**2+a**2), S.true, b), # 6.21
+            (c*exp(b*t)*cos(a*t), c*(s-b)/((s-b)**2+a**2), S.true, b), # 6.22
+            ]
+        for rule in simple_rules:
```
oscargus, line 1182: Maybe one would like to unpack the rule here, like:
```
for time_domain, laplace_domain, condition, convergence_plane in simple_rules:
```
(or something more compact).

Will maybe make the code a bit easier to read.

hanspi42:
Very annoying: when I do `LT(cos(t*(w - 1)), t, s)`, the expression that is handed down to `doit()` is unfortunately `cos(t*w - t)`, and this does not match. `a*t`.

Does anybody know how I could automatically apply `.collect(t)` to the arguments of all functions in an expression?

hanspi42:
@oscargus: I have implemented your suggestion, thanks again.

@oscarbenjamin: I found a solution, I first tidy up the expression with a new `_deep_collect(f, t)` function by traversing the tree leaves first and applying `.collect()` each time the function in the tree node is `Add()`.

I implemented it such that every table entry now has a fifth element, a one-argument lambda function that is applied before pattern matching. Like this other prep functions could be used as well.

oscarbenjamin:
This is looking nice and also much faster than master:
```python
In [3]: %time laplace_transform(exp(-3*t), t, z)
CPU times: user 12 ms, sys: 0 ns, total: 12 ms
Wall time: 12.5 ms
Out[3]: 
⎛  1            ⎞
⎜─────, -3, True⎟
⎝z + 3          ⎠
```
We should try to structure the code in such a way that there is a function that can be called that will attempt rule-based evaluation of the Laplace transform and does not fall back on the integration code. Then the `laplace_transform` function or `LaplaceTransform.doit` can call that function and handle falling back to integration. That way it becomes possible for the integration code itself to make use of the rule-based Laplace transform function when a suitable integral arrives.

hanspi42:
> We should try to structure the code in such a way that there is a function that can be called that will attempt rule-based evaluation of the Laplace transform and does not fall back on the integration code. Then the `laplace_transform` function or `LaplaceTransform.doit` can call that function and handle falling back to integration. That way it becomes possible for the integration code itself to make use of the rule-based Laplace transform function when a suitable integral arrives.

OK, I will prepare for that by having all rules in the same place. Let me first complete the rules; then we can do this step. I guess I'll need some guidance for doing this ;)

sympy/integrals/transforms.py
```diff
@@ -1167,38 +1167,56 @@ def _compute_transform(self, f, t, s, **hints):
...
+                (exp(a*t-tau), exp(-tau)/(s-a),
+                     S.true, a, dco), # 3.1
+                (t*exp(a*t-tau), exp(-tau)/(s-a)**2,
+                     S.true, a, dco), # 3.2
+                (t**c*exp(a*t), gamma(c+1)/(s-a)**(c+1),
+                     c>-1, a, dco), # 3.3
+                (exp(-a*t**2), sqrt(pi/4/a)*exp(s**2/4/a)*erfc(s/sqrt(4*a)),
+                     a>0, S.Zero, dco), # 3.6
+                (t*exp(-a*t**2),
+                     1/(2*a)-2/sqrt(pi)/(4*a)**(3/2)*s*erfc(s/sqrt(4*a)),
```
oscargus, line 1187: Change `3/2` to `Rational(3, 2)`, `S(3)/2`, or `Integer(3)/2` to avoid floats (created by Python before SymPy gets hold of the input). (Not clear where the 0.25 in the test comes from though. Does it print like that or is it a consequence of `Eq(0.25, Rational(1, 4)) = True`?)

hanspi42: > ... to avoid floats (created by Python before SymPy gets hold of the input) ...

Thanks! This single sentence solved SOOO many of my open questions ;)  Changed, and added four more rules with test cases.

sympy/integrals/transforms.py
```diff
@@ -1150,9 +1156,89 @@ class LaplaceTransform(IntegralTransform):
...
+
     def _compute_transform(self, f, t, s, **hints):
+        a = Wild('a', exclude=[t])
+        b = Wild('b', exclude=[t])
+        c = Wild('c', exclude=[t])
+        g = WildFunction('g', nargs=1)
+        tau = Wild('tau', exclude=[t])
+        omega = Wild('omega', exclude=[t])
+        dco = lambda f: self._deep_collect(f,t)
+        if not any(func.has(t) for func in f.atoms(AppliedUndef)):
```
oscarbenjamin, line 1174: Why is this `if` needed?

hanspi42: Everything after the `else:`, if it triggers any rule, returns a `LaplaceTransform` and will not be evaulated further. The `if` is there because I fear infinite loops.

oscarbenjamin: I don't think I understand. Why is it checking `func.has(t)`?

hanspi42: Because if it does not have `t`, it should use the normal rules.

With the present code, 

```LT(exp(-f(x)*t), t, s)```

gives 

```(1/(s + f(x)), -f(x), True)```,

processed by rule 3.1, as it should.

**Comment by hanspi42** [comment]
Now `diff` is implemented, it looks beuatiful ;))))

diff

**Comment by oscarbenjamin** [comment]
Yes, that is nice. I wonder if the pretty printer can be improved to look more like the latex printer:
```python
In [2]: s = Symbol('s')

In [3]: laplace_transform(diff(f(t), (t, 4)), t, s)
Out[3]: 
                                                                  ⎛  2      ⎞│      ⎛  3      ⎞│   
 4                                 3         2 ⎛d       ⎞│        ⎜ d       ⎟│      ⎜ d       ⎟│   
s ⋅LaplaceTransform(f(t), t, s) - s ⋅f(0) - s ⋅⎜──(f(t))⎟│    - s⋅⎜───(f(t))⎟│    - ⎜───(f(t))⎟│   
                                               ⎝dt      ⎠│t=0     ⎜  2      ⎟│      ⎜  3      ⎟│   
                                                                  ⎝dt       ⎠│t=0   ⎝dt       ⎠│t=0
```

hanspi42:
Would be nice to have a better pretty printer. Meanwhile I have written a small helper function to solve differential equations.

Check my test fiile.

[0003_all_asserts.zip](https://github.com/sympy/sympy/files/7554363/0003_all_asserts.zip)


The code lines

'''
s1 = LT(diff(f(t), (t, 3)), t, s)
print('\n-->',s1)
ic = [c0, c1, c2, c3, c4]
s2 = laplace_transform_replace_ic(s1, f, F, t, s, ic)
print('-->',s2)
print('-->',solve(s2, F(s)))
'''

give 

'''
--> s**3*LaplaceTransform(f(t), t, s) - s**2*f(0) - s*Subs(Derivative(f(t), t), t, 0) - Subs(Derivative(f(t), (t, 2)), t, 0)
--> -c0*s**2 - c1*s - c2 + s**3*F(s)
--> [(c0*s**2 + c1*s + c2)/s**3]
'''

Before I code all checks, tests, etc: is this a proper way to go?

oscargus:
I think the "problem" with the pretty printer is related to `⋅` and `⎜` and the used font which aren't the same width as the rest of the characters. Which is a bad thing when the printer relies on a fixed-width font and counting characters...

This is with DejaVu Sans Mono (from my editor, hence the vertical line and so on):
image

Btw, really nice PR @hanspi42 !

oscarbenjamin:
> I think the "problem" with the pretty printer is related to `⋅` and `⎜` and the used font which aren't the same width as the rest of the characters.

You seem to be paying much more attention to some things than me but much less to others! To me the problem is that whereas in the latex version the Laplace transform of an undefined function comes out like `L_t[f(t)](s)`  the pretty print version is `LaplaceTransform(f(t), t, s)`. I think that some people get confused about whether SymPy has "computed" the Laplace transform because of this difference (https://github.com/sympy/sympy/issues/7219#issuecomment-802092986).

Also yes, much kudos to @hanspi42 for making headway on this longstanding issue!

hanspi42:
This is going to give a lot to review, so I will guide you reviewers through the review process. I have re-checked all rules from the top to 3.14 and think they are implemented correctly now. The pdf below shows the first four fields of each rule converted automatically to LaTeX for better readability:

[0006_table.pdf](https://github.com/sympy/sympy/files/7574933/0006_table.pdf)

Could somebody check this against the respective tables on http://eqworld.ipmnet.ru/en/auxiliary/inttrans/ ?

sympy/integrals/transforms.py
```diff
@@ -1150,9 +1175,193 @@ class LaplaceTransform(IntegralTransform):
...
+                # 2.2 makes little sense here; it is covered by Heaviside and
+                #     piecewise linear.
+                (t, 1/s**2,
+                     S.true, S.Zero, dco), # 2.3
+                (1/(t+a), -exp(-a*s)*Ei(-a*s),
+                     S.true, S.Zero, dco), # 2.4
+                (t**n, factorial(n)/(s**(n+1)),
+                     S.true, S.Zero, dco), # 2.5
+                # 2.6 is not worth the effort, if users do not precisely know
+                #     how to code the request, it will never match.
```
oscarbenjamin, line 1221: Isn't 2.6 just something like `sqrt(t)`? Is it not needed because it is handled by 2.5?

sympy/integrals/transforms.py
```diff
@@ -1014,9 +1018,30 @@ def expand_dirac_delta(expr):
...
+        k+=1
+    return res
+
 @_noconds
 def _laplace_transform(f, t, s_, simplify=True):
-    """ The backend function for Laplace transforms. """
+    """ The backend function for Laplace transforms.
+
+    This backend assumes that the frontend has already split sums
+    such that `f` is to an addition anymore.
```
oscarbenjamin, line 1043: What function should be called if we don't want to make this assumption?

This should ideally be structured so that there is a single function that can compute the Laplace transform using rules or otherwise return `None`. Or perhaps the function should return say `(A, B)` where `A` is a successfully computed part and `B` is the part that could not be handled and interpretation is that `L(f) = A + L(B)`.

sympy/integrals/transforms.py
```diff
@@ -1150,9 +1175,193 @@ class LaplaceTransform(IntegralTransform):
...
+        b = Wild('b', exclude=[t])
+        n = Wild('n', exclude=[t])
+        g = WildFunction('g', nargs=1)
+        tau = Wild('tau', exclude=[t])
+        omega = Wild('omega', exclude=[t])
+        dco = lambda f: self._deep_collect(f,t)
+        if not any(func.has(t) for func in f.atoms(AppliedUndef)):
+            simple_rules = [
+                # ( time domain, laplace domain,
+                #       condition, convergence plane, preparation function )
```
oscarbenjamin, line 1196: It would be better to define all these rules as a global variable in the module I think.

oscarbenjamin: Or maybe not because then it would happen at import time.

Either way there should be a way to get these rules directly i.e. a function that can be called like `_get_laplace_transform_rules()`.

sympy/integrals/transforms.py
```diff
@@ -1150,9 +1175,193 @@ class LaplaceTransform(IntegralTransform):
...
+                (t**n*sinh(a*t), gamma(n+1)/2*((s-a)**(-n-1)-(s+a)**(-n-1)),
+                     n>-2, S.Zero, dco), # 4.4
+                (sinh(2*sqrt(a*t)), sqrt(pi*a)/s/sqrt(s)*exp(a/s),
+                     S.true, S.Zero, dco), # 4.5
+                (sqrt(t)*sinh(2*sqrt(a*t)), pi**(S(1)/2)*s**(-S(5)/2)*\
+                     (s/2+a)*exp(a/s)*erf(sqrt(a/s))-a**(S(1)/2)*s**(-2),
+                     S.true, S.Zero, dco), # 4.6
+                (sinh(2*sqrt(a*t))/sqrt(t), pi**(S(1)/2)*s**(-S(1)/2)*\
+                     exp(a/s)*erf(sqrt(a/s)),
+                     S.true, S.Zero, dco), # 4.7
```
oscarbenjamin, line 1286: In terms of formatting, it is nicer to split at comma rather than an arithmetic operator like `*`.

oscarbenjamin: Maybe something like:
```python
def _get_laplace_transform_rules():
    rules = {
        ...
        '4.7': sinh(2*sqrt(a*t))/sqrt(t),
               pi**(S(1)/2)*s**(-S(1)/2)*exp(a/s)*erf(sqrt(a/s)),
               S.true, S.Zero, dco),
        '4.8': ...
    }
    return rules
```

sympy/integrals/transforms.py
hanspi42: > What function should be called if we don't want to make this assumption?
> 
> This should ideally be structured so that there is a single function that can compute the Laplace transform using rules or otherwise return `None`. Or perhaps the function should return say `(A, B)` where `A` is a successfully computed part and `B` is the part that could not be handled and interpretation is that `L(f) = A + L(B)`.

I am assuming that normally you would not call `_laplace_transform` yourself, this is only called by `self.doit()`. The partitioning into sum terms is done there, so the comment means "don't remove that functionality from `doit()`"

sympy/integrals/transforms.py
hanspi42: agreed, will do

sympy/integrals/transforms.py
hanspi42: I don't see why the names of the rules should be in the data space. Is a dict always odered? 

But I will structure it more nicely.

sympy/integrals/transforms.py
hanspi42: Actually it should not be covered by 2.5. There I have to check for `n=1, 2, 3, 4 ...`. How can I do this? If I try to use `.isinteger` in the rule, it fails because the first thing to appear there will be a `Wild` object that does not have an `.isinteger` method.

sympy/integrals/transforms.py
oscarbenjamin: I think that we would also want to call this from the main `integrate` function so that if someone asks for an integral that is equivalent to a Laplace transform then we can use this routine to calculate it. Since the main `laplace_transform` falls back on `integrate` we need a function that just applies these rules but does not fall back on `integrate`.

sympy/integrals/transforms.py
hanspi42: Just a question: that function will need all the wildcards. Should these be passed as arguments?

sympy/integrals/transforms.py
hanspi42: Forget this ... 2.5 and 2.6 are both covered wonderfully by 2.11 because `gamma` does the right thing ;)

sympy/integrals/transforms.py
oscarbenjamin: I think the function should return the wildcards.

hanspi42: Do users of the function need them at all, or can they just be part of the returned expressions?

sympy/integrals/transforms.py
hanspi42: Done.

sympy/integrals/transforms.py
hanspi42: I have written something that should be useful.

sympy/integrals/transforms.py
hanspi42: I have written something that should be useful :)

oscarbenjamin:
Is the special handling of `DiracDelta` in `_laplace_transform` still needed after the changes in this PR?

hanspi42:
> Is the special handling of `DiracDelta` in `_laplace_transform` still needed after the changes in this PR?

No, it is not, but I will build it back only after I have implemented all rules. I tried to remove it once and too many things broke at once.

Other question: I would like to add time scaling to all rules instead of implementing it as an additional step. This has the disadvantage that the same kind of code is in multiple places, but the distinct advantage that it is visible in the rule list. One example is here:

https://github.com/sympy/sympy/blob/49fbd49b631a25a6ce785afdb775321c0a3decac/sympy/integrals/transforms.py#L1363-L1366

In many rules this is alrerady built in naturally, such as

https://github.com/sympy/sympy/blob/49fbd49b631a25a6ce785afdb775321c0a3decac/sympy/integrals/transforms.py#L1335-L1338

Is it OK for you if I just include it everyhere where it makes sense? The function `dco` given in each rule makes this very easy to match.

oscarbenjamin:
> Is it OK for you if I just include it everyhere where it makes sense?

I think so although I'm not sure exactly what the two alternatives you are suggesting are. Do you just mean that you will generalise the rules from [1] rather than apply them literally? In other words this rule would be change to match `1/sqrt(a*t + b)`:
https://github.com/sympy/sympy/blob/49fbd49b631a25a6ce785afdb775321c0a3decac/sympy/integrals/transforms.py#L1222-L1224

[1] http://eqworld.ipmnet.ru/en/auxiliary/aux-inttrans.htm

hanspi42:
> In other words this rule would be change to match `1/sqrt(a*t + b)`:
> 
> https://github.com/sympy/sympy/blob/49fbd49b631a25a6ce785afdb775321c0a3decac/sympy/integrals/transforms.py#L1222-L1224

Yes, it would then be

```
    (1/sqrt(b*t+a),
     sqrt(pi/s)*exp(a/b*s)*erfc(sqrt(a/b*s))/b,
     b>0, S.Zero, dco),
```

oscarbenjamin:
That seems reasonable. I'm not sure if there are other situations where a general delay handler would still be needed perhaps for delays involving undefined functions or something.

hanspi42:
> > Is the special handling of `DiracDelta` in `_laplace_transform` still needed after the changes in this PR?
> 
> No, it is not, but I will build it back only after I have implemented all rules. I tried to remove it once and too many things broke at once.

Correction: yes, it is. The present code includes the masking property and can transform the following:

```
LT(exp(-t)*(DiracDelta(t)+DiracDelta(t-42)), t, s)
```

I do not see how I could implement the masking property with rules.

hanspi42:
I found something very annoying: Test case `test_issue_14692` uses `I` as a factor to `t`.  Now I am at the limit of my mathemathical knowledge. I don't know all the conditions for all the rules in [1] except that they surely work if coefficients are real. I think the only way to play it safe is to check whether they are real and only apply the rules if this is so.

The best way I found to check whether something is real that will only be evaluated after `.xreplace` is this:

https://github.com/sympy/sympy/blob/c8c1981f4849021706d9880ed0f1b43dc3cfd05f/sympy/integrals/transforms.py#L1183

applied like that:

https://github.com/sympy/sympy/blob/c8c1981f4849021706d9880ed0f1b43dc3cfd05f/sympy/integrals/transforms.py#L1218-L1221

It works with `test_issue_14692`, but I will also apply it to `b`, of course.

I think I will add this to all rules. Then expressions with non-real coefficients wil be solved as an integral, and it is later possible to relax conditions once we know they can be relaxed for some rules.

What do you think about it?

hanspi42:
Aaargh, the above does not work.  The code on line 1436 of

https://github.com/sympy/sympy/blob/c8c1981f4849021706d9880ed0f1b43dc3cfd05f/sympy/integrals/transforms.py#L1433-L1438

triggers in this case when the replace operation is successful.

I have tried many things but am stuck: does anyone have an idea how I can provide a sympy expression in the list that is to be taken by my code and then 

1. replace wildcards by values
2. evaluate and check whether true

Whatever I tried up to now, my `isreal()` function is evaluated before the wildcards are replaced.

Does anyone have an idea how to do this?

hanspi42:
I think I have a solution to make the new code work recursively such that general rules can also apply to concrete functions.

https://github.com/sympy/sympy/blob/58efc63ad93041ee593a06a142c5c6466b0cd3d9/sympy/integrals/transforms.py#L1547-L1559

@oscarbenjamin: this solves the "why do we need `if not any(func.has(t) for func in f.atoms(AppliedUndef))`" that you asked about two weeks ago.

oscargus:
> Does anyone have an idea how to do this?

Not sure this is applicable here, but would it be possible to make a `is_extended_real` check and if `None` return a Piecewise like `Piecewise((result, Eq(re(a), a)), (LaplaceTransform(input), True))`? Or `Eq(im(a), 0)`.

I checked the Laplace transform for 1/(b*x+a) with Mathematica and that returns: 

(E^((a s)/b) (Gamma[0,(a s)/b]+Log[a/b]+Log[b/a]))/b

not sure if that is just a rewrite of Ei or not. For b = I it returns a value separated into a CI and SI, which, as far as I can tell, is just a rewrite. However, it is also commonly stated that Ei is just for real-values and E1 should be used for complex arguments. Or that may just be the integral definition that only works for real-valued arguments and Ei is defined for complex arguments in SymPy. 

Maybe @jksuom can shed some light on this?

oscarbenjamin:
We can also just have the pattern rules written with Piecewise where needed. The Piecewise will evaluate if the value is real:
```python
In [1]: p = Piecewise((1, Contains(x, Reals)), (2, True))

In [2]: p
Out[2]: 
⎧1  for x ∈ ℝ
⎨            
⎩2  otherwise

In [3]: p.subs(x, 4)
Out[3]: 1

```

hanspi42:
Very good ideas. Meanwhile, I could do miracles if I was able to use a new hint called `applyrules` in addition to 'needeval' etc.

However, I could not find out in where and in which file the selection of allowed hints can be found. I tried code like

```
                hints['applyrules'] = False
                return k*LaplaceTransform(ma1[g].func(t), t, s, **hints)
```

but LaplaceTransform bails out with 

```
  File "sympy\core\function.py", line 284, in __new__
    raise ValueError("Unknown options: %s" % options)

ValueError: Unknown options: {'applyrules': False}
```

so I guess I have to add this hint name to some list somewhere.

hanspi42:
> I checked the Laplace transform for 1/(b*x+a) with Mathematica and that returns:
> 
> (E^((a s)/b) (Gamma[0,(a s)/b]+Log[a/b]+Log[b/a]))/b
> 
> not sure if that is just a rewrite of Ei or not.

Well, for non-zero arguments, Log[a/b]+Log[b/a] is zero, and Gamma(0,x) is Ei(x), so it is the same as we have ;)

hanspi42:
> For b = I it returns a value separated into a CI and SI, which, as far as I can tell, is just a rewrite.

I am currently planning to let all rules just work on parameters that I am really sure are correct; all other cases will be treated as before.  It will always be possible to extend the rules later, but I first need to have something not too big but consistent, or it may blow my mind ;))

hanspi42:
Just in case you are wondering what I did with the hints: I added a hint that prevents endless loops and makes it possible to apply rules without evaluating integrals

https://github.com/sympy/sympy/blob/904e1926f52232d881741636b2cba21210f62262/sympy/integrals/transforms.py#L1502-L1504

hanspi42:
Dear reviewers,

I think I have finished the task and the code is ready for review.

After the review, while doing necessary corrections, I will also
- [ ] update the documentation a bit,
- [ ] add an example for solving differential equations to the documentation,
- [ ] write proper notes,
- [ ] squash the commits.

I have omitted all general rules containing integrals (1.15 ... 1.25) as I do not really see how this would help anybody much; if you think people need them and can actually use them efficiently, tell me. 

Parameter ranges for some rules may be more restrictive than necessary, but I would very much prefer to just take it as we have it (and later increase parameter ranges, when necessary, with comments from the field) such that I can turn to working on the rules for the inverse Laplace transform next.

hanspi42:
By the way, if you execute the tests like so,

```
import sympy
from sympy.integrals.tests.test_transforms import test_laplace_transform
sympy.SYMPY_DEBUG=True

import sys
file_path = '0007_debug_tests_out.txt'
sys.stderr = open(file_path, "w")
test_laplace_transform()
```

then you will see in the output file which rules were triggered for which test cases.  Here the output for your convenience:

[0007_debug_tests_out.zip](https://github.com/sympy/sympy/files/7625616/0007_debug_tests_out.zip)

sympy/integrals/transforms.py
```diff
@@ -1169,6 +1734,59 @@ def _collapse_extra(self, extra):
...
+            for x in res:
+                if not isinstance(x, tuple):
+                    x = [x]
+                ress.append(x[0])
+                if len(x) == 2:
+                    # only a condition
+                    extra.append(x[1])
+                elif len(x) > 2:
+                    # some region parameters and a condition (Mellin, Laplace)
+                    extra += [x[1:]]
```
oscargus, line 1770: extra.append? (Compare three rows up. Or change there?)

hanspi42: Hmm, I have no clue ;)  `LaplaceTransform` did not have its own `doit` before, so this part of the code is a verbatim copy of the `doit` of `IntegralTransform`

https://github.com/sympy/sympy/blob/b5191b805eba30488a54cc0e46dbab66052a0429/sympy/integrals/transforms.py#L175-L180

If we change this in one place, we should also cgange it in the other place.

hanspi42:
> I have checked it and have nothing to object against (on the opposite: a nice, huge, step!).

Thank you ;)

> Some minor things for the sake of it:
> 
> Are all the "removed" tests just moved elsewhere? I cannot seem to find the cosh*sinh-test, but I may just not see it.

There are no removed tests, they are just re-sorted.  `cosh*sinh`, e.g., is here:

https://github.com/sympy/sympy/blob/b5191b805eba30488a54cc0e46dbab66052a0429/sympy/integrals/tests/test_transforms.py#L549

> It would be nice to add a comment for the tests that corresponds to close issues as it right now is not that clear without checking the issues etc.

Good point.  Should these be comment lines next to the tests, or can it be a description under "Other comments" together with the release notes?

oscargus:
> > Some minor things for the sake of it:
> > Are all the "removed" tests just moved elsewhere? I cannot seem to find the cosh*sinh-test, but I may just not see it.
> 
> There are no removed tests, they are just re-sorted. `cosh*sinh`, e.g., is here:

Great!

> > It would be nice to add a comment for the tests that corresponds to close issues as it right now is not that clear without checking the issues etc.
> 
> Good point. Should these be comment lines next to the tests, or can it be a description under "Other comments" together with the release notes?

I'd say lines in the tests (as often there are separate, explicit, tests for the issues). If nothing else for later developers to know that the test is "special" and shouldn't be removed in the first place.

sympy/integrals/transforms.py
```diff
@@ -1137,6 +1160,541 @@ def sbs(expr):
...
+            L = _laplace_apply_rules(ma1[g].func(t), t, s, **hints)
+            try:
+                r, p, c = L
+                return (k*exp(-ma2[a]*s)*r, p, c)
+            except TypeError:
+                return k*exp(-ma2[a]*s)*L
+    ma1 = func.match(exp(y)*z)
+    if ma1:
+        ma2 = ma1[y].collect(t).match(a*t)
+        if ma2:
```
oscarbenjamin, line 1618: The code in this section seems to be a bit repetitive. Can these rules be abstracted out somehow or at least made into named functions or something?

sympy/integrals/transforms.py
```diff
@@ -1137,6 +1160,541 @@ def sbs(expr):
...
+            debug('_laplace_apply_rules match:')
+            debug('      f:    %s ( %s, %s )'%(f, ma1, ma2))
+            debug('      rule: multiply with sinh (1.5)')
+            L = _laplace_apply_rules(ma1[z], t, s, **hints)
+            try:
+                r, p, c = L
+                return ((r.subs(s, s-ma2[a])-r.subs(s, s+ma2[a])).simplify()/2,
+                        p+Abs(ma2[a]), c)
+            except TypeError:
```
oscarbenjamin, line 1640: Why is this catching TypeError? What raises the error?

In general it is better to avoid catching TypeError because it risks masking bugs. There are many situations where a bug/typo etc can lead to a TypeError but this would suppress the traceback.

sympy/integrals/transforms.py
```diff
@@ -1137,6 +1160,541 @@ def sbs(expr):
...
+        for k in range(ma1[n]):
+            if k==0:
+                y = ma1[g].func(t).subs(t, 0)
+            else:
+                y = Derivative(ma1[g].func(t), (t, k)).subs(t, 0)
+            d.append(s**(ma1[n]-k-1)*y)
+        r = s**ma1[n]*_laplace_apply_rules(ma1[g].func(t), t, s, **hints)
+        return r - Add(*d)
+    raise IntegralTransformError(
+                'Laplace', None, 'No rules could be applied.')
```
oscarbenjamin, line 1697: Maybe this could just return None?

sympy/integrals/transforms.py
```diff
@@ -1151,7 +1709,14 @@ class LaplaceTransform(IntegralTransform):
     _name = 'Laplace'
 
     def _compute_transform(self, f, t, s, **hints):
-        return _laplace_transform(f, t, s, **hints)
+        try:
+            hints['doit'] = True
+            return _laplace_apply_rules(f, t, s, **hints)
```
oscarbenjamin, line 1714: I don't understand what the `doit` flag does here. Is that so that the `_laplace_apply_rules` function could be used in a different context?

Is `doit` already a flag used by `laplace_transform`?

Maybe `laplace_apply_rules` could use a different parameter name for this to avoid conflicting with whatever `doit` does.

sympy/integrals/transforms.py
```diff
@@ -1137,6 +1160,541 @@ def sbs(expr):
...
+            debug('      f:    %s'%(func,))
+            debug('      rule: %s o---o %s'%(t_dom, s_dom))
+            try:
+                debug('      try   %s'%(check,))
+                c = check.xreplace(ma)
+                debug('      check %s -> %s'%(check, c))
+                if c==True:
+                    debug(' ---> go')
+                    return _laplace_cr(k*s_dom.xreplace(ma),
+                                plane.xreplace(ma), S.true, **hints)
```
oscarbenjamin, line 1573: The `doit=True` was popped from hints above and so will not be passed on here. Is that intentional?

sympy/integrals/transforms.py
oscarbenjamin: It's better to avoid duplicating the same code in multiple places. Can one method not just call another or can there be a third method that they both call or something?

sympy/integrals/transforms.py
```diff
@@ -1169,6 +1734,59 @@ def _collapse_extra(self, extra):
...
+                pass
+        else:
+            hints['needeval'] = needeval
+            res = [self.__class__(*([x] + list(self.args[1:]))).doit(**hints)
+                   for x in fn.args]
+            extra = []
+            ress = []
+            for x in res:
+                if not isinstance(x, tuple):
+                    x = [x]
```
oscarbenjamin, line 1763: If `x` is not a tuple then turn it into a list?

sympy/integrals/transforms.py
```diff
@@ -1169,6 +1734,59 @@ def _collapse_extra(self, extra):
                 'Laplace', None, 'No combined convergence.')
         return plane, cond
 
+    def doit(self, **hints):
+        """
+        Try to evaluate the Laplace transform in closed form.
+        """
+
+        needeval = hints.pop('needeval', False)
```
oscarbenjamin, line 1742: Rather than using `pop` it would be better to use explicit keyword only arguments:
```python
def doit(self, *, needeval=False, **hints):
```

sympy/integrals/transforms.py
hanspi42: Yes, that flag makes just that possible. This is how I understood your original suggestion to write the rules engine such that it can be used, e.g., by integrate.

The doit flag does not exist yet in the Laplace transform.

sympy/integrals/transforms.py
hanspi42: Maybe. I knew that you would ask this, and thought about it, but did not find a good way. Since there are only five similar rules (exp, sin, cos, sinh, cosh) and more are not expected to come, and there are small differences in what they return, I decided against abstracting it.

sympy/integrals/transforms.py
hanspi42: The line `r, p, c = L` raises a TypeError if the unpacking fails. `L = _laplace_apply_rules(ma1[z], t, s, **hints)` above can return a tuple or a `LaplaceTransform` object.  This was my attempt to write "if it is a tuple, do it on the first entry, otherwise do it on everything"

sympy/integrals/transforms.py
hanspi42: An then write the part 

https://github.com/sympy/sympy/blob/b5191b805eba30488a54cc0e46dbab66052a0429/sympy/integrals/transforms.py#L1711-L1719

as 

```
L = _laplace_apply_rules(f, t, s, **hints)
if L==None:
    return _laplace_transform(f, t, s, **hints)
else:
    return L
```

?

sympy/integrals/transforms.py
hanspi42: Here it is even vital that it is not passed on, it would give an error above if this hint is passed onwarts to a `LaplaceTransform` object.

However, when `_laplace_apply_rules` calls itself, the hint needs to be passed on. This is an error I need to correct.

sympy/integrals/transforms.py
hanspi42: I don't know. This is a copy-paste from the present `doit` definition of the class `IntegralTransform`, I did not change that part.

sympy/integrals/transforms.py
hanspi42: I believe you, but this is also from the original:

https://github.com/sympy/sympy/blob/b5191b805eba30488a54cc0e46dbab66052a0429/sympy/integrals/transforms.py#L151-L153

There would then be three keywords that would need to be made explicit. Should we change this in both locations?

hanspi42:
> > > It would be nice to add a comment for the tests that corresponds to close issues as it right now is not that clear without checking the issues etc.
 
> > 
> > Good point. Should these be comment lines next to the tests, or can it be a description under "Other comments" together with the release notes?

> I'd say lines in the tests (as often there are separate, explicit, tests for the issues). If nothing else for later developers to know that the test is "special" and shouldn't be removed in the first place.

I will do this.

sympy/integrals/transforms.py
oscarbenjamin: Yes

sympy/integrals/transforms.py
oscarbenjamin: Okay so if this flag only exists for the purpose of using `_laplace_apply_rules` then it can have a more descriptive name. To me `doit` in this context suggests "evaluate any integrals". How can we describe in a name what this flag is for?

Also we don't need `hints['doit'] = True`. It's just:
```python
return _laplace_apply_rules(f, t, s, doit=True, **hints)
```

sympy/integrals/transforms.py
oscarbenjamin: They could at least be in named functions so that the high-level logic in this function is easier to parse. Something like:
```python
def _exp_rule(...):
    ...

def _laplace_transform(func):
    rules = [_exp_rule, _sin_rule, ...]
    for rule in rules:
        result = rule(func)
        if result is not None:
            return result
```

sympy/integrals/transforms.py
oscarbenjamin: Is there a need for `_laplace_apply_rules` to return ` LaplaceTransform` object? Does that just represent a failure to compute the Laplace transform? If so then it could just yse `return None` instead and then:
```python
L = f()
if L is None:
   ...
```

sympy/integrals/transforms.py
oscarbenjamin: The original code probably predates explicit keyword-only argments:
https://www.python.org/dev/peps/pep-3102/

> There would then be three keywords that would need to be made explicit.

Do they all need to be made explicit? The point is just that mutating data structures like this is fragile whereas using an explicit parameter for `needeval` is less bug-prone.

hanspi42: I see the point, but I suggest the following: I will re-write `_laplace_apply_rules` such that it uses a keryword-only argument, but will not touch the needeval/doit part. For that I suggest to open a new "code cleanup" issue and solve it for all integral transforms independent of the functionality change I am doing here.

sympy/integrals/transforms.py
hanspi42: Thanks, I will do precisely this. I used the other strategy by learning from the present (and apparently very old) code.

sympy/integrals/transforms.py
oscarbenjamin: I think really the problem is that the `doit` method from the superclass should not be duplicated here.

sympy/integrals/transforms.py
hanspi42: This whole block is duplicated:

https://github.com/sympy/sympy/blob/b5191b805eba30488a54cc0e46dbab66052a0429/sympy/integrals/transforms.py#L165-L201

So I could convert it to a new function inside `IntegralTransform`. Is that a proper way to solve it?

oscarbenjamin: There are many possible ways to do this without duplication. The question is which part of the code needs to be *different* in the subclass from that of the superclass? That part of the code can be moved into a separate method that is called by the superclass `doit` method. Then the subclass can override that method and have it do something different while both classes can share the code that needs to be the *same* for both classes.

sympy/integrals/transforms.py
hanspi42: > Is there a need for `_laplace_apply_rules` to return `LaplaceTransform` object?

Oh yes, `_laplace_apply_rules` must return `LaplaceTransform` objects when it simplifies expressions with undefined functions, see e.g. the `diff` stuff in https://github.com/sympy/sympy/pull/22376#issuecomment-971412783 .

sympy/integrals/transforms.py
hanspi42: OK!

hanspi42:
My to-do-list for the final submission:

- [ ] Re-write `_laplace_apply_rules` such that it uses a keyword-only argument instead of a hint;
- [ ] make sure that the `doit` argument is passed on properly in recursive calls;
- [ ] re-write `_laplace_apply_rules` such that it returns `None` instead of raising an error and adapt `_compute_transform` to cover this;
- [ ] convert the exp, sin, cos, sinh, cosh rules to high-level rules;
- [ ] un-duplicate code in `doit`;
- [ ] add comments to those test cases that test issues fixed with this pull request;
- [ ] update the documentation;
- [ ] add an example for solving differential equations to the documentation;
- [ ] write proper release notes;
- [ ] squash the commits.

sympy/integrals/transforms.py
oscarbenjamin: Should it not still return a tuple in that case (with the convergence conditions)?

hanspi42: There is no convergence condition for undefined functions, so I think it should not return any in that case. 

What could, in principle, be done is to implement a "convergence plane modifier", so, e.g., `sinh(a*t)*f(t)` would mean that the correct convergencer plane is the maximum of `Abs(a)` and the convergence plane of transforming `f(t)` on its own, but this would be a major hassle to implement.

I also think it is not necessary. Convergence in the Laplace Transform just means "right of the right-most singularity", so I think it is OK if determining the convergence condition after substituting a concrete function for an undefined function in the result is left as an exercise for the user. If they want to have an automatic convergence plane, they shall use defined functions ;)

sympy/integrals/transforms.py
hanspi42: The lines

https://github.com/sympy/sympy/blob/b5191b805eba30488a54cc0e46dbab66052a0429/sympy/integrals/transforms.py#L152-L153

in the superclass hurt me badly.  I also need `AppliedUndef` to be deferred to my `_compute_transform`; this is why duplicated `doit`. Is there a possibility to change the `doit` in the superclass such that it behaves differently when used in the `LaplaceTranform`?

sympy/integrals/transforms.py
oscarbenjamin: I just mean that it's a bit weird to have a function that sometimes returns a tuple and sometimes returns a LaplaceTransform. Looking at it now I realise that's what `laplace_transform` does by default:
```python
In [7]: laplace_transform(f(t), t, s)
Out[7]: LaplaceTransform(f(t), t, s)

In [8]: laplace_transform(exp(-t), t, s)
Out[8]: 
⎛  1           ⎞
⎜─────, 0, True⎟
⎝s + 1         ⎠
```
That's a bit unfortunate...

hanspi42:
For the reviewers' convenience: the built-in rules automatically converted to LaTeX:

[0006_table.pdf](https://github.com/sympy/sympy/files/7648681/0006_table.pdf)

sympy/integrals/transforms.py
hanspi42: We can discuss it here: https://github.com/sympy/sympy/pull/22376#discussion_r761809377

If that one is solved, this one is automatically solved as well.

sympy/integrals/transforms.py
hanspi42: If https://github.com/sympy/sympy/pull/22376#discussion_r761809377 is solved, this one is resolved too, so I resolve it right away :)

sympy/integrals/transforms.py
hanspi42: It is unfortunate, yes!  I think we should leave it as is because changing this will affect a lot of users.

sympy/integrals/transforms.py
oscarbenjamin: Yes, just move that code into a separate method and override the method in the subclass.

hanspi42: Thank you for teaching me proper object-oriented programming here!

(I guess you must already found out that OOP is not something I have actually learnt properly and that I am making things up as I go along ;)

Understood, I will do this!

oscarbenjamin: No problem. The existing code here is quite messy. We need to do our best not to make that worse... (when possible)

hanspi42:
Dear @oscarbenjamin and @oscargus, I think I have finished all tasks. Once you give me the green lights, I will squash the commits and then signal to @oscarbenjamin that a merge of this pull request is possible.

sympy/integrals/tests/test_transforms.py
```diff
@@ -468,71 +465,274 @@ def mysimp(expr):
...
+    assert LT(exp(a*(t-b)), t, s) == (exp(-a*b)/(-a + s), a, True)
+    assert LT(t*exp(-a*(t)), t, s) == ((a + s)**(-2), -a, True)
+    assert LT(t*exp(-a*(t-b)), t, s) == (exp(a*b)/(a + s)**2, -a, True)
+    assert LT(b*t*exp(-a*t), t, s) == (b/(a + s)**2, -a, True)
+    assert LT(t**(S(7)/4)*exp(-8*t)/gamma(S(11)/4), t, s) ==\
+        ((s + 8)**(-S(11)/4), -8, True)
+    assert LT(t**(S(3)/2)*exp(-8*t), t, s) ==\
+        (3*sqrt(pi)/(4*(s + 8)**(S(5)/2)), -8, True)
+    assert LT(t**42*exp(-42*t), t, s) == \
+        (1405006117752879898543142606244511569936384000000000/(s + 42)**43,
```
oscarbenjamin, line 514: It's clearer to write `factorial(42)` here.

oscarbenjamin: Maybe there could be a symbolic test like:
```python
In [8]: k = Symbol('k', positive=True)

In [9]: laplace_transform(t**k*exp(-k*t), t, s)
Out[9]: 
⎛       -k - 1                   ⎞
⎝(k + s)      ⋅Γ(k + 1), -k, True⎠
```
Is this result valid for noninteger k?

sympy/integrals/tests/test_transforms.py
```diff
@@ -868,6 +1061,7 @@ def test_issue_8882():
         **{'as_meijerg': True, 'needeval': True}))
 
 
+@XFAIL
 def test_issue_7173():
```
oscarbenjamin, line 1065: Why is this XFAIL now?

sympy/integrals/transforms.py
```diff
@@ -1137,6 +1190,525 @@ def sbs(expr):
...
+     a>0, S.Zero, dco),
+    # 8.18
+    (besselk(0, a*t),
+     (log(s+sqrt(s**2-a**2)))/(sqrt(s**2-a**2)),
+     a>0, Abs(a), dco)
+    ]
+    return laplace_transform_rules
+
+def _laplace_cr(f, a, c, **hints):
+    conds = not hints.get('noconds', False)
```
oscarbenjamin, line 1575: A one-line docstring at least would help here

sympy/integrals/transforms.py
```diff
@@ -1137,6 +1190,525 @@ def sbs(expr):
...
+            try:
+                r, p, c = L
+                return (r, p+ma2[a], c)
+            except TypeError:
+                return L
+    trigrules = [(sinh(y), '1.5',  1, -1, 1, 1), (cosh(y), '1.6', 1, 1, 1, 1),
+                 (sin(y),  '1.7', -I, -1, I, 0), (cos(y),  '1.8', 1, 1, I, 0)]
+    for trigrule in trigrules:
+        fm, nu, s1, s2, ff, kk = trigrule
+        ma1 = func.match(fm*z)
```
oscarbenjamin, line 1676: This is quite cryptic. I think if this was a function with its own docstring then that would help to provide the space to explain what this is doing.

sympy/integrals/transforms.py
```diff
@@ -1169,6 +1748,20 @@ def _collapse_extra(self, extra):
                 'Laplace', None, 'No combined convergence.')
         return plane, cond
 
+    def _try_directly(self, **hints):
```
oscarbenjamin, line 1751: This looks better

oscarbenjamin:
I think this is almost there. The main thing is it would be better to split out the repetitive sections from `_laplace_apply_rules` into separate named functions so that they could have their own docstrings and be tested separately. Right now looking at that function it's hard to see what it's doing from a high-level.

sympy/integrals/tests/test_transforms.py
hanspi42: Ah, I forgot to remove it. This actually tested a failure in `meijerint`, corrected by https://github.com/sympy/sympy/pull/8905, but `sinh(a*x)*cosh(a*x)` does not go through `meijerint` anymore.

I will remove the test and write a comment in the main test list where issue #7173 is tested.

sympy/integrals/tests/test_transforms.py
hanspi42: Yes, it is. I will use your test case.

sympy/integrals/transforms.py
hanspi42: OK!

sympy/integrals/transforms.py
hanspi42: :)

sympy/integrals/transforms.py
hanspi42: I will do this.

hanspi42:
> I think this is almost there. The main thing is it would be better to split out the repetitive sections from `_laplace_apply_rules` into separate named functions so that they could have their own docstrings and be tested separately. Right now looking at that function it's hard to see what it's doing from a high-level.

I totally agree and have found a way to do it.

hanspi42:
Final task, squash commits, still open ;)

sympy/integrals/transforms.py
```diff
@@ -147,18 +166,13 @@ def doit(self, **hints):
...
+        simplify = hints.pop('simplify', True)
+        hints['simplify'] = simplify
 
-        fn = self.function
-        if not fn.is_Add:
-            fn = expand_mul(fn)
+        fn, T = self._try_directly(**hints)
+
+        if not T==None:
+            return T
```
oscarbenjamin, line 175: Not so important but `if T is not None` is usually the preferred way to spell this.

sympy/integrals/transforms.py
```diff
@@ -1137,6 +1190,607 @@ def sbs(expr):
         aux = _simplifyconds(aux, s, a)
     return _simplify(F.subs(s, s_), simplify), sbs(a), _canonical(sbs(aux))
 
+def _laplace_deep_collect(f, t):
+    func = f.func
+    args = [*f.args]
```
oscarbenjamin, line 1195: I guess this is equivalent to `args = list(args)` but I haven't seen it written this way before.

sympy/integrals/transforms.py
```diff
@@ -1137,6 +1190,607 @@ def sbs(expr):
...
+        return f
+    else:
+        for k in range(len(args)):
+            args[k] = _laplace_deep_collect(args[k], t)
+        if func.is_Add:
+            return func(*args).collect(t)
+        else:
+            return func(*args)
+
+def _laplace_build_rules(t, s):
```
oscarbenjamin, line 1206: This function should have some sort of comment or docstring that explains what it returns and what the purpose is.

sympy/integrals/transforms.py
```diff
@@ -1137,6 +1190,607 @@ def sbs(expr):
...
+    conds = not hints.get('noconds', False)
+    if conds:
+        return f, a, c
+    else:
+        return f
+
+def _laplace_rule_timescale(f, t, s, doit=True, **hints):
+    """
+    This function tries to apply the time-scaling rule of the Laplace
+    transform and returns `None` if it cannot do it.
```
oscarbenjamin, line 1588: This should explain what the "time-scaling" rule of the Laplace transform is and it would be good to have a doctest example.

sympy/integrals/transforms.py
```diff
@@ -1014,9 +1034,42 @@ def expand_dirac_delta(expr):
     """
     return _lin_eq2dict(expr, expr.atoms(DiracDelta))
 
+def laplace_transform_replace_ic(expr, f, F, t, s, ic):
```
oscarbenjamin, line 1037: Should this function be considered public? In other words do we expect users to use this function? If so it should have better docs following the style guide:
https://docs.sympy.org/latest/documentation-style-guide.html

sympy/integrals/transforms.py
oscarbenjamin: Also if this is an internal function rather than a function for end-users then I think it's good to state that explicitly in the docstring somehow such as "Internal helper function to do X"

sympy/integrals/transforms.py
```diff
@@ -1137,6 +1190,607 @@ def sbs(expr):
...
+    will attempt to integrate.
+
+    If it is called with `doit=False`, then it will instead return
+    `LaplaceTransform` objects.
+    """
+
+    k, func = f.as_independent(t, as_Add=False)
+
+    simple_rules = _laplace_build_rules(t, s)
+    for t_dom, s_dom, check, plane, prep in simple_rules:
```
oscarbenjamin, line 1769: Maybe this part could also be factored out into a function and then listed with the other functions in `prog_rules`

sympy/integrals/transforms.py
```diff
@@ -1137,6 +1190,607 @@ def sbs(expr):
...
+    """
+    This function tries to transform a product containing the `Heaviside`
+    function and returns `None` if it cannot do it.
+    """
+    hints.pop('simplify', True)
+    a = Wild('a', exclude=[t])
+    b = Wild('b', exclude=[t])
+    y = Wild('y')
+    g = WildFunction('g', nargs=1)
+    k, func = f.as_independent(t, as_Add=False)
```
oscarbenjamin, line 1629: It looks like most (all?) rules call `as_independent` like this. Maybe that should be done before the rule function is called.

asmeurer:
I'm happy to see this finally being done. I have a few comments:

- We should obviously be doing this for every transform, not just Laplace. Can we set up the basic infrastructure for all the transforms, even if we don't have any rules for them yet? I think once we have the design down, adding rules will be the easy part, and it will also be easy for people to contribute additional rules.
- It isn't clear to me if the rules are applied to the exact expression provided by the user or if the input is pre-processed. I think it would be a good idea to at least split the input expression linearly, so `laplace_transform(a*f + b*g)` only pattern matches against `f` and `g`. Otherwise even a simple constant factor will break the pattern match (if this is actually already being done, then pay this comment no mind).
- Let's design this in a way that makes it possible for users to inject their own rules if they want to. Should be as simple as doing something like

```py
laplace_rules = []
def _laplace_build_rules():
    rules.append((...))
    rules.append((...))
```

- Just a note that ideally we would use matchpy for this, but that requires a lot of other refactorings first. 
- You could also consider modeling this after the `manualintegrate` module. The way that module is implemented allows things like extracting the steps used (this is how the integral steps part of sympygamma.com works). OTOH manualintegrate is a bit more complex than this because here we generally only expect one rule to match, and any matching rule to always "work", which isn't the case with manualintegrate.

sympy/utilities/tests/test_wester.py
```diff
@@ -2945,14 +2944,13 @@ def test_Y5_Y6():
...
     F, _, _ = laplace_transform(diff(y(t), t, 2)
                                 + y(t)
                                 - 4*(Heaviside(t - 1)
                                 - Heaviside(t - 2)), t, s)
-    # Laplace transform for diff() not calculated
-    # https://github.com/sympy/sympy/issues/7176
-    assert (F == s**2*LaplaceTransform(y(t), t, s) - s
-            + LaplaceTransform(y(t), t, s) - 4*exp(-s)/s + 4*exp(-2*s)/s)
+    G = laplace_transform_replace_ic(F, y, Y, t, s, [1, 0])
+    assert (G == s**2*Y(s) - s + Y(s) - 4*exp(-s)/s + 4*exp(-2*s)/s)
```
oscarbenjamin, line 2953: I wonder if it would be better to extend the signature of `laplace_transform` rather than have a separate function for this. I can see the argument both ways but I expect that a lot of users would want `laplace_transform` to directly convert `L{y(t)}` to `Y(t)` rather than call a separate function afterwards. I haven't really considered what a good way of allowing users to specify that would be...

oscarbenjamin: Or maybe there could be a separate function called e.g. `laplace_transform_ode` that is specifically for applying the Laplace transform to ODEs.

oscarbenjamin: Perhaps this particular part of the PR could be deferred to another PR so we can get the rest in and consider these details later.

hanspi42:
> I'm happy to see this finally being done. I have a few comments:
> 
> We should obviously be doing this for every transform, not just Laplace. Can we set up the basic infrastructure for all the transforms, even if we don't have any rules for them yet? I think once we have the design down, adding rules will be the easy part, and it will also be easy for people to contribute additional rules.

I will be happy to do this, but I really would like you to let me to do it in this order:

1. Finish this pull request without the possibility to let users inject their own rules.
2. In a second pull request, re-write the inverse Laplace transform such that it uses rules to deal with rational fractions. Then it is possible to solve linear differential equations by Laplace transform, and it may be ready for deployment in sympy 1.10
3. Then I will have enough knowledge to be able to participate in a discussion about how a basic infrastructure for all integral transfroms should be designed. So we specify an architecture together.
4. We (or I) implement that architecture, and in that step, we make it possible for people to append their own rules, and we also tidy up the messy code in `IntegralTransform` that prevents you from seeing the answer to the next question ;)

>  It isn't clear to me if the rules are applied to the exact expression provided by the user or if the input is pre-processed. I think it would be a good idea to at least split the input expression linearly, so `laplace_transform(a*f + b*g)` only pattern matches against `f` and `g`. Otherwise even a simple constant factor will break the pattern match (if this is actually already being done, then pay this comment no mind).

It is like this already, but not easy to read. So I would like to defer the following points also to step 4 of my list above.

>  Let's design this in a way that makes it possible for users to inject their own rules if they want to. 

>  Just a note that ideally we would use matchpy for this, but that requires a lot of other refactorings first.

>  You could also consider modeling this after the `manualintegrate` module. ...

Is it OK for you if I proceed like this?

sympy/utilities/tests/test_wester.py
hanspi42: If all of you agree with the project plan I outlined in the reply to @asmeurer below, we could do this in step 4 later.

asmeurer:
I'm fine with doing things in further PRs. It's not like anything here is public API, so it's OK if we end up completely refactoring it to something different later.

hanspi42:
Very good, thanks, so I will reply to the rest of @oscarbenjamin's comments with that plan in mind.

sympy/integrals/transforms.py
hanspi42: Presently it is for all rules, but this cannot be guaranteed to be true for rules that will come later, or that users can supply themselves according to what @asmeurer suggests below, so I will just leave it as is.

sympy/integrals/transforms.py
hanspi42: I thought about it, but I would like to keep it as is, because if we  unify it for all integral transforms later and decide on a way how to allow users to supply their own rules to the function, it will lead to a re-design anyway.

sympy/integrals/transforms.py
hanspi42: Yes, it should, so I will write an appropriate doc with example.

sympy/integrals/transforms.py
hanspi42: I thought the trailing underbar meant precisely that, but I will change it in all relevant docstrings.

sympy/integrals/transforms.py
hanspi42: Will do.

sympy/integrals/transforms.py
hanspi42: I will change it, then, I love the '*' unoack notation, but may have gone too far here ;)

sympy/integrals/transforms.py
hanspi42: I will change it everywhere, it looks nicer.

sympy/integrals/transforms.py
oscarbenjamin: Yes, the leading underscore does indicate that it is internal.

hanspi42:
Once more I believe I have everything now ;))

If you give me green lights, I will squash the commits.

oscarbenjamin:
> It's not like anything here is public API

The PR in its current state adds a new public function `laplace_transform_replace_ic`. I am not sure about the interface of this function and whether it is the best way to implement the feature expected by users so I would rather postpone that change to a future PR so it can be discussed in detail on its own.

hanspi42:
> > It's not like anything here is public API
> 
> The PR in its current state adds a new public function `laplace_transform_replace_ic`. I am not sure about the interface of this function and whether it is the best way to implement the feature expected by users so I would rather postpone that change to a future PR so it can be discussed in detail on its own.

So would I delete it again, or make it an interrnal helper function for the time being and document it as such?

oscarbenjamin:
> So would I delete it again, or make it an interrnal helper function for the time being and document it as such?

Is it actually used by any of the other code here? AFAICT it is only used in the tests so it's not really an internal helper. I think I would rather remove it from this PR entirely.

oscarbenjamin:
> I think I would rather remove it from this PR entirely.

But we can open a new issue or a separate PR to add this right now and discuss the details there.

hanspi42:
I have now created a new branch for the helper function: https://github.com/hanspi42/sympy/tree/laplace_diff_eq

Is it a good idea to already create the new pull request, or should I wait until this one is merged?

sympy/integrals/tests/test_transforms.py
```diff
@@ -553,42 +751,26 @@ def test_laplace_transform():
...
+        Ms_conds = Matrix([[(1/(s - 1), 1, True), ((s + 1)**(-2),
+            -1, True)], [((s + 1)**(-2), -1, True), (1/(s - 1), 1, True)]])
     with warns_deprecated_sympy():
         assert LT(Mt, t, s) == Ms_conds
-
     # The new behavior is to return a tuple of a Matrix and the convergence
     # conditions for the matrix as a whole:
-    assert LT(Mt, t, s, legacy_matrix=False) == (Ms, 0, Abs(s) > 1)
-
+    assert LT(Mt, t, s, legacy_matrix=False) == (Ms, 1, True)
```
oscarbenjamin, line 768: I just want to note here that the `legacy_matrix=False` argument is a terrible API. I added it myself because there were backwards compatibility concerns with #21496 (https://github.com/sympy/sympy/pull/21496#discussion_r637461504). We should fix this so that a matrix is returned by default. Also the fact that `laplace_transform` sometimes returns an `Expr` and sometimes a `tuple` should be fixed somehow.

sympy/integrals/transforms.py
```diff
@@ -1137,6 +1162,624 @@ def sbs(expr):
...
+            args[k] = _laplace_deep_collect(args[k], t)
+        if func.is_Add:
+            return func(*args).collect(t)
+        else:
+            return func(*args)
+
+def _laplace_build_rules(t, s):
+    """
+    This is an internal helper function that returns the table of Laplace
+    transfrom rules in terms of the time variable `t` and the frequency
```
oscarbenjamin, line 1187: transfrom -> transform

oscarbenjamin:
Okay looks good. Let's get this in so further improvements can be made later.

Thanks @hanspi42!

oscarbenjamin:
@Upabjojr the changes here would make a good candidate for demonstrating the benefits of matchpy. With SymPy's current match routine a number of manual steps are needed around the calls to match but also it's a classic many-to-one match routine which is precisely what matchpy is supposed to help with.

sympy/integrals/tests/test_transforms.py
hanspi42: The other integral transforms also do this, don't they?

hanspi42:
> > I think I would rather remove it from this PR entirely.
> 
> But we can open a new issue or a separate PR to add this right now and discuss the details there.

Here you go: https://github.com/sympy/sympy/pull/22630

sympy/integrals/tests/test_transforms.py
oscarbenjamin: I'm not sure but it's a bad API. We shouldn't have tuples in a matrix and we also shouldn't have functions that sometimes return Expr and sometimes return tuples.

Upabjojr:
> @Upabjojr the changes here would make a good candidate for demonstrating the benefits of matchpy. With SymPy's current match routine a number of manual steps are needed around the calls to match but also it's a classic many-to-one match routine which is precisely what matchpy is supposed to help with.

SymPy can already use MatchPy if it is installed. The only obstacle remains making MatchPy a SymPy dependency.

oscarbenjamin:
> making MatchPy a SymPy dependency

The point is that we should do that when there is some actual benefit for users e.g. if we have a PR that demonstrates the improvement brought by matchpy in computing Laplace transforms or something. Right now making matchpy a dependency is just asking users to install an extra package for no actual benefit.

hanspi42:
> > making MatchPy a SymPy dependency
> 
> The point is that we should do that when there is some actual benefit for users e.g. if we have a PR that demonstrates the improvement brought by matchpy in computing Laplace transforms or something. Right now making matchpy a dependency is just asking users to install an extra package for no actual benefit.

I'll read https://dl.acm.org/doi/10.1145/3149869.3149871 and form an opinion to contribute ;)

hanspi42:
> > > making MatchPy a SymPy dependency
> > 
> > 
> > The point is that we should do that when there is some actual benefit for users e.g. if we have a PR that demonstrates the improvement brought by matchpy in computing Laplace transforms or something. Right now making matchpy a dependency is just asking users to install an extra package for no actual benefit.
> 
> I'll read https://dl.acm.org/doi/10.1145/3149869.3149871 and form an opinion to contribute ;)

OK, I am stuck.  Even after reading everything I found, I don't see how to use it.

@Upabjojr: it would be nice if you could tell us how to implement, e.g., this rule in matchpy:

https://github.com/sympy/sympy/blob/6632ed8b1d0ad351a1c8dbf65c8bb57bdc78fd4b/sympy/integrals/transforms.py#L1261-L1264

It matches Line 1262 and replaces by 1263 if ``And(n>-1, a>0)`` is true.

Upabjojr:
> > > > making MatchPy a SymPy dependency
> > > 
> > > 
> > > The point is that we should do that when there is some actual benefit for users e.g. if we have a PR that demonstrates the improvement brought by matchpy in computing Laplace transforms or something. Right now making matchpy a dependency is just asking users to install an extra package for no actual benefit.
> > 
> > 
> > I'll read https://dl.acm.org/doi/10.1145/3149869.3149871 and form an opinion to contribute ;)
> 
> OK, I am stuck. Even after reading everything I found, I don't see how to use it.
> 
> @Upabjojr: it would be nice if you could tell us how to implement, e.g., this rule in matchpy:
> 
> https://github.com/sympy/sympy/blob/6632ed8b1d0ad351a1c8dbf65c8bb57bdc78fd4b/sympy/integrals/transforms.py#L1261-L1264
> 
> It matches Line 1262 and replaces by 1263 if `And(n>-1, a>0)` is true.

Using the stuff in https://github.com/sympy/sympy/blob/master/sympy/utilities/matchpy_connector.py , maybe:

```python
from sympy.utilities.matchpy_connector import *

a = WildDot("a")
n = WildDot("n")
constraint = CustomConstraint(lambda n, a: And(n > -1, a > 0) == True)
replacer = Replacer()
replacer.add((a*t+b)**n, lowergamma(n+1, b/a*s)*exp(-b/a*s)/s**(n+1)/a, conditions_true=constraint)
```

(I didn't test this code, but it should be something similar to this)

PR #16967, opened 2019-06-04
Removed undesired random state change

caph1993:
Sympy was altering the Python's internal random state. With this fix it doesn't.

#### References to other Issues or PRs
Fixes #12628 
Fixes #14589 

#### Brief description of what is fixed or changed

Importing and calling `shuffle` changes the Python's internal random state. I replaced that with `Random().shuffle`, which is still a randomized shuffle, but does not change the random state.

#### Other comments

You can run the code below (with the current version) to verify that just by calling the attribute `is_finite` on a symbol, the internal random state changes. This happens also when multiplying two symbols. With this fix, it doesn't.

```python
import sympy as sp
import random

h = hash(random.getstate())
def random_state_changed():
    return h != hash(random.getstate())

print(random_state_changed())
z = sp.Symbol('z')
print(random_state_changed())
z.is_finite
print(random_state_changed())
```

#### Release Notes

* core
  * Using `random.seed` leads now to the same results on consecutive runs.

PR #22385, opened 2021-10-30
[OK except CI] add numerically_zero to randtest

smichr:

as discussed in #8516

#### Release Notes

* other
    * added `numerically_zero` testing function

### DIFF
--- a/sympy/testing/randtest.py
+++ b/sympy/testing/randtest.py
@@ -66,7 +66,7 @@ def numerically_zero(f, number=10, prec=10, a=2, b=-1, c=3, d=1, maxfail=100, ma
     tol = S(10)**-prec if prec else None
     f = S(f)
     if not f.args:
-        if f.is_number and abs(f) <= (tol or 0):
+        if f.is_number and f.is_comparable and abs(f) <= (tol or 0):
             return True
         return False
     s = f.free_symbols
--- a/sympy/testing/randtest.py
+++ b/sympy/testing/randtest.py
@@ -3,7 +3,9 @@
 from random import uniform, Random, randrange, randint
 
 from sympy.core.containers import Tuple
+from sympy.core.function import _coeff_isneg
 from sympy.core.numbers import comp, I
+from sympy.core.singleton import S
 from sympy.core.symbol import Symbol
 from sympy.simplify.simplify import nsimplify
 from sympy.utilities.iterables import is_sequence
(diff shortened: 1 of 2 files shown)


sympy/testing/randtest.py
```diff
@@ -27,6 +28,80 @@ def random_complex_number(a=2, b=-1, c=3, d=1, rational=False, tolerance=None):
...
+    If you are not sure what precision to use you can set ``prec=None`` and
+    a numerical estimate of the error will be made from two numerical
+    estimates.
+
+    Examples
+    ========
+
+    >>> from sympy import sin, cos, pi, S, Add
+    >>> from sympy.abc import x
+    >>> from sympy.utilities.randtest import numerically_zero as is_zero
```
smichr, line 49:
```suggestion
    >>> from sympy.testing.randtest import numerically_zero as is_zero
```

sympy/testing/randtest.py
```diff
@@ -27,6 +28,80 @@ def random_complex_number(a=2, b=-1, c=3, d=1, rational=False, tolerance=None):
...
+    False
+
+    See Also
+    ========
+    random_complex_number
+    """
+    tol = S(10)**-prec if prec else None
+    f = S(f)
+    if not f.args:
+        if f.is_number and abs(f) <= (tol or 0):
```
smichr, line 69:
```suggestion
        if f.is_number and f.is_comparable and abs(f) <= (tol or 0):
```

sympy/testing/tests/test_randtest.py
```diff
@@ -0,0 +1,46 @@
+from sympy.core.add import Add
+from sympy.core.numbers import pi
+from sympy.core.singleton import S
+from sympy.functions.elementary.trigonometric import cos, sin
+from sympy.functions.elementary.miscellaneous import sqrt
+from sympy.solvers.solvers import solve
+from sympy.testing.randtest import numerically_zero
+from sympy.utilities.pytest import slow
```
smichr, line 8:
```suggestion
from sympy.testing.pytest import slow
```

oscarbenjamin:
Where is this explanation for this PR? The OP seems to have the wrong issue link.

sympy/testing/randtest.py
```diff
@@ -4,6 +4,7 @@
 
 from sympy.core.containers import Tuple
 from sympy.core.numbers import comp, I
+from sympy.core.singleton import S
```
oscarbenjamin, line 7: The testing module should ideally not import anything from the rest of sympy except potentially from utilities. Since I moved the testing module here I have been meaning to move this submodule out. It depends on core and so should not be part of testing.

smichr:
> Where is this explanation for this PR? The OP seems to have the wrong issue link.

The issue is correct. See especially dicussion after [here](https://github.com/sympy/sympy/issues/8516#issuecomment-64695226).  The function adds the ability to do less than `expr.equals(0)` and more than checking `expr._random()` to see if it is zero.

sympy/testing/randtest.py
smichr: So should randtest.py go back to utilities?

oscarbenjamin: I don't think it should be in utilities either since I don't think that utilities should import anything from core. Probably only `_randint` and `_randrange` belong in utilities although I'm not convinced that either is needed at all.

Maybe the rest should go into the `sympy.core.random` or something. The only place any of this is used in non-testing code is in `sympy.core.expr`.

See also #16967

smichr: > although I'm not convinced that either is needed at all.

The routines were provided as a convenience to allow a routine to control the numbers being provided by the same function that would otherwise be selecting them on its own based on the value of seed.

```python
>>> r = _randrange(range(10))
>>> [r(10) for i in range(10)]
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]

>>> r=_randrange([2,3,2,1,4,6,5])
>>> while 1:
...   r(3,5)
...
3
4
...
ValueError: _randrange sequence was too short
```

PR #22386, opened 2021-10-30
Moved imports to top level

oscargus:

Sort of an alternative solution to #22351 (both can still be merged, but lambdify is no longer directly importing from core).

Moved "all" core and utilities to the top level.

#### Other comments

It would make sense to move the `deprecated` decorator to `utilities` as this is the one causing cyclic imports at the moment. This will be in a later PR though as the current one really doesn't affect any functionality (just better importing).

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/assumptions/refine.py
+++ b/sympy/assumptions/refine.py
@@ -1,6 +1,7 @@
 from typing import Dict, Callable
 
-from sympy.core import S, Add, Expr, Basic, Mul
+from sympy.core import S, Add, Expr, Basic, Mul, Pow, Rational
+from sympy.core.logic import fuzzy_not
 from sympy.logic.boolalg import Boolean
 
 from sympy.assumptions import ask, Q  # type: ignore
@@ -83,7 +84,6 @@ def refine_abs(expr, assumptions):
     -x
 
     """
-    from sympy.core.logic import fuzzy_not
     from sympy.functions.elementary.complexes import Abs
     arg = expr.args[0]
     if ask(Q.real(arg), assumptions) and \
@@ -133,7 +133,6 @@ def refine_Pow(expr, assumptions):
     (-1)**(x + 1)
 
     """
-    from sympy.core import Pow, Rational
     from sympy.functions.elementary.complexes import Abs
     from sympy.functions import sign
     if isinstance(expr.base, Abs):
--- a/sympy/codegen/ast.py
+++ b/sympy/codegen/ast.py
@@ -134,7 +134,8 @@ from sympy.core.basic import Basic
 from sympy.core.expr import Expr
 from sympy.core.numbers import Float, Integer, oo
 from sympy.core.sympify import _sympify, sympify, SympifyError
-from sympy.utilities.iterables import iterable
+from sympy.utilities.iterables import (iterable, topological_sort,
+                                       numbered_symbols, filter_symbols)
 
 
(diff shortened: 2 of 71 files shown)


sympy/assumptions/refine.py
```diff
@@ -1,6 +1,7 @@
 from typing import Dict, Callable
 
-from sympy.core import S, Add, Expr, Basic, Mul
+from sympy.core import S, Add, Expr, Basic, Mul, Pow, Rational
```
ThePauliPrinciple, line 3: Is it safe to import from the core package? Or should there be imports from their respective modules e.g.:
```python
from sympy.core.add import Add
from sympy.core.power import Pow
```

ThePauliPrinciple (commented):
1)
In several places, imports are performed from the `core` package rather than from the respective module. Is there a reasoning behind sometimes importing from the module and sometimes from the package?

2)
Are these import moves possible due to recent changes? Or is this more a clean up of code that used local imports without a good reason in the first place?

sympy/assumptions/refine.py
oscargus: Ideally from the modules, but I only wanted to spend so much time (sometimes I changed it to modules, but not everywhere).

oscargus: But it is in general safe. Makes it clearer though to import from the modules.

oscargus:
1. Ideally they should be imported from the module for clarity (I do not know about efficiency), but it is a question of time to spend. I basically copied it over and sometimes changed it to fully specified module. Working by the step-by-step principle...

2. I made an effort to make `utilities` and `external` independent of `core` (and all other sympy-modules) and only import from those two in `core`, to try to reduce the cyclic import issues. So as a consequence, it should be possible to import  `core`, `external`, and `utilities` from top-level in all other modules. At some stage (once it is fully implemented) this should be documented and ideally some tests be added to at least avoid cyclic imports in that part of the code.

(Currently `logic` is also imported in `core`, but that can/will be changed by moving parts of `boolalg` to `core`. I just tied it out locally and seems to be pretty straightforward. Still a few PRs I would like to have merged before making that change though...)

oscarbenjamin:
> Ideally they should be imported from the module for clarity (I do not know about efficiency), but it is a question of time to spend.

I agree that it is difficult to change everything all at once so it isn't necessary to do that in this PR but I just want to clarify what I think is the real reason that internal imports should be direct to the module when there are cyclic import problems.

Consider something like this:
```python
# sympy/__init__.py

from sympy.a import AA, BB

# sympy/a/__init__.py

from sympy.a.aa import AA

from sympy.a.bb import BB

# sympy/a/aa.py

from sympy.a import BB  # <--- ImportError here

class AA:
    pass

# sympy/a/bb.py

class BB:
    pass
```
This will give an ImportError because BB does not yet exist in the sympy.a module at the time that`aa.py` is imported. If the import was direct to the module `sympy.a.bb` using either a relative or an absolute import path then it could succeed.

This should be fixed for everything that is imported before the core has finished being imported. It doesn't need to apply to things that are not used at all by core like `sympy.physics`.

oscargus:
The other side, I guess, is that end users can import import `BB` from `sympy` directly without having to consider where it is located, making restructuring the code easier. (Although that is of course only an argument if the code is badly structured...)

PR #22388, opened 2021-10-30
Move deprecated

oscargus:

Built on #22386, so the diff will be much smaller once that is merged.

Move the `deprecated` decorator from `core` to `utilities` to allow better import structure.

#### Other comments

I do not think there is anything special needed here as this is an internal function.

#### Release Notes

NO ENTRY


oscarbenjamin:
Looks good to merge once flake8 issues are resolved.

PR #22391, opened 2021-10-30
Str is now non-scalar Symbol

smichr:

starts to address #22272 

Str now differs from Symbol only in that it is not a scalar. Otherwise, it can accept assumptions and now Indexed and MatrixSymbol both use Str as the object to represent the symbol for the object. A decision was made in #17652 to allow assumptions to remain on the label of Indexed so that is why Str was made to subclass from Symbol (an object which has all the assumption machinery built into it).

#### Release Notes

* core
    * Str is a subclass of Symbol


oscarbenjamin:
> Str now differs from Symbol only in that it is not a scalar.

I don't understand what "not a scalar" means or what the rationale for this change is. The point of Str is that it is just supposed to be a Basic wrapper for str for situations where we need to store a string as part of a symbolic expression of some kind. I think that Str should not subclass Expr or Boolean.

Also the issue number referenced in the OP is invalid. Should it be #17652?

smichr:
> I don't understand what "not a scalar" means

For `x` to be scalar means `diff(x) == 1`. Str was introduced because it was thought to be bad to have Symbol carrying label information along with the `scalar=True`. 

As I mentioned in the OP, #17652 allows Indexed to keep attributes on the Symbol. By doing so it requires an object that can hold assumptions. That object has been, until now, Symbol. But this is allowing a scalar into the name slot for IndexedBase when it should really be a Str -- and that's why MatrixSymbol and IndexedBase are different: the former uses Str and the latter, Symbol. They should both be using the same thing for the semantically identical "label".

This PR is attempting to fix that.

oscarbenjamin:
Okay that makes sense.

I don't think that the Str should hold any assumptions. The Indexed object itself can hold assumptions that could perhaps be derived from the corresponding IndexedBase but the name itself is just a string so properties like `is_real` don't make sense there.

smichr:
> Str should not subclass Expr or Boolean

I agree with this since I don't think we want `Str('a') + Str('b')` to do anything mathematical. I am not sure how involved it will be to get Indexed to work with a non-Symbol object. But if you are in favor the direction of the change I can give it a try.

smichr:
> I don't think that the Str should hold any assumptions. The Indexed object itself can hold assumptions

We let a 3rd party drive a decision on this in #17652. Please check that issue and see if you have a better idea.

oscarbenjamin:
> > I don't think that the Str should hold any assumptions. The Indexed object itself can hold assumptions
> 
> We let a 3rd party drive a decision on this in #17652. Please check that issue and see if you have a better idea.

Looking at #17652 it doesn't seem to have anything to do with assumptions. I think they just wanted `IndexedBase` to store the label argument without modification.

CC @theHamsta

smichr:
> anything to do with assumptions

They wanted it stored unmodified so they could get at the assummptions on it.  So what we set up was a requirement that an object deriving from Symbol not be modified...so I don't know how to do anthing but make Str derive from Symbol if we are going to keep that requirement.

sympy/stats/tests/test_mix.py
```diff
@@ -43,23 +44,24 @@ def test_MarginalDistribution():
...
+    Indexed(IndexedBase('C'), 1)), 2)),
+    ExprCondPair(0, True)), Pow(gamma(Symbol('a1', positive=True)),
+    -1), gamma(Add(Symbol('a1', positive=True),
+    Indexed(IndexedBase('C'), 0))),
+    Pow(gamma(Indexed(IndexedBase('C'), 0)), -1),
+    Pow(Indexed(IndexedBase('B'), 0),
+    Add(Symbol('a1', positive=True), -1)),
+    Pow(Indexed(IndexedBase('B'), 1),
+    Add(Indexed(IndexedBase('C'), 0), -1)))
+    assert MGR(C) == mgrc, MGR(C)
```
smichr, line 63:
```suggestion
    from sympy import srepr
    assert MGR(C) == mgrc, ('\n',srepr(mgrc),'\n',srepr(MGR(C)))
```

smichr:
> I don't know how to do anthing but make Str derive from Symbol if we are going to keep that requirement.

I think the better thing to do is ask @theHamsta if they can just get the assumptions from the Indexed instead of the label of the Indexed. Then Str can keep deriving from Atom and then it is just a matter of trying to get Indexed to work with Str instead of Symbol.

smichr:
In #16085 the assumptions were allowed to be specified on the Symbol that was being used to define the IndexedBase. In light of "one way" principle, it seems like it would be better to require the name to be given as a string and the assumptions to be given explicitly. Is there a reason one would want to keep Symbols around with with assumptions separate from the object? And if someone says that it is just a convenient way to store the assumptions, I would suggest:
```
Instead of:

x = Symbol('x', positive=True, integer=True)
IndexedBase(x)

use:

a = dict(positive=True, integer=True)
IndexedBase('x', **a)
```
Then

oscarbenjamin:
The first argument (label) to IndexedBase should be treated purely as an identifier just like the argument to Symbol. Any semantic properties like assumptions should be part of the IndexedBase rather than its label.

smichr:
> should be part of the IndexedBase rather than its label.

This will need a deprecation.

sympy/core/symbol.py
```diff
@@ -1,5 +1,5 @@
 from .assumptions import StdFactKB, _assume_defined
-from .basic import Basic, Atom
+from .basic import Basic
```
smichr, line 2:
```suggestion
from .basic import Atom, Basic
```

PR #13007, opened 2017-07-19
Sorted imports in hanging grid style

jirikuncar:
cherry-picked from #12970 
closes #12929
closes #12935

PR #22394, opened 2021-10-31
OneArray and ZeroArray shape parameter is now a tuple

redeboer:
The `shape` parameter of `OneArray` and `ZeroArray` is now a single positional argument, as to match the new interface of `ArraySymbol` (#22324). Note that this matches the signature of e.g. [`numpy.ones()`](https://numpy.org/doc/stable/reference/generated/numpy.ones.html).

See also #22321.

#### Release Notes

* tensor
  * Changed the signature of `OneArray` and `ZeroArray` from `OneArray(*shape)` to `OneArray(shape)`.

### DIFF
--- a/sympy/tensor/array/expressions/array_expressions.py
+++ b/sympy/tensor/array/expressions/array_expressions.py
@@ -3,7 +3,7 @@ from collections import defaultdict, Counter
 from functools import reduce
 import itertools
 from itertools import accumulate
-from typing import Optional, List, Dict as tDict, Tuple as tTuple
+from typing import Iterable, Optional, List, Dict as tDict, Tuple as tTuple
 
 import typing
 
@@ -43,13 +43,11 @@ class ArraySymbol(_ArrayExpr):
     name = property(lambda self: self._args[0])
     shape = property(lambda self: self._args[1])
 
-    def __new__(cls, symbol, shape: typing.Iterable) -> "ArraySymbol":
+    def __new__(cls, symbol, shape: Iterable) -> "ArraySymbol":
         if isinstance(symbol, str):
             symbol = Symbol(symbol)
-        # symbol = _sympify(symbol)
-        shape = Tuple(*map(_sympify, shape))
-        obj = Expr.__new__(cls, symbol, shape)
-        return obj
+        sympified_shape = Tuple(*map(_sympify, shape))
+        return Expr.__new__(cls, symbol, sympified_shape)
 
     def __getitem__(self, item):
         return ArrayElement(self, item)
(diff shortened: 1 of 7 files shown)


sympy/tensor/array/expressions/array_expressions.py
```diff
@@ -1,34 +1,46 @@
...
+)
 from sympy.tensor.array.ndim_array import NDimArray
-from sympy.tensor.indexed import (Indexed, IndexedBase)
-from sympy.matrices.expressions.matexpr import MatrixElement
-from sympy.tensor.array.expressions.utils import _apply_recursively_over_nested_lists, _sort_contraction_indices, \
-    _get_mapping_from_subranks, _build_push_indices_up_func_transformation, _get_contraction_links, \
-    _build_push_indices_down_func_transformation
-from sympy.combinatorics import Permutation
-from sympy.combinatorics.permutations import _af_invert
-from sympy.core.sympify import _sympify
```
Upabjojr, line 31: is there a reason for changing the import lines so much?

redeboer: [isort](https://pycqa.github.io/isort). But ok, I'll kick that commit out

redeboer: Done :heavy_check_mark: 
But tbh, those unsorted, unorganised import statements in sympy give me goosebumps... :sweat_smile:

Upabjojr: Well, I am actually OK with reorganizing the imports. I usually don't edit those lines manually, as I use PyCharm's auto-import capabilities.

Maybe this discussion could apply to all of SymPy (and not just to this module)?

Upabjojr: (maybe reorganizing the imports can be done in a separate PR...)

redeboer: I looked this up once in the issues and vaguely remembered someone already tried applying isort (can't find that issue/PR anymore). Problem is that the order of imports matters in sympy: ordering the imports can cause circular imports (huge code smell imho :grimacing:).

At any rate, I don't think it's good to start applying isort on arbitrary modules without an official policy on this. That's why I kicked out that commit.

redeboer: There it is: https://github.com/sympy/sympy/pull/13007#issuecomment-316875190

ThePauliPrinciple: I think sorting imports could be a great idea, but should be done as a seperate issue/PR, especially considering work has been done on the imports recently (https://github.com/sympy/sympy/pull/22342).

redeboer: Ah yeah good point.

> I use PyCharm's auto-import capabilities.

As a side note, not a PyCharm user, but I'm quite sure it can call isort automatically upon saving as well.

redeboer:
https://github.com/sympy/sympy/pull/22396#issuecomment-960614666 same here

PR #22396, opened 2021-10-31
Allow taking a single index from ArraySymbol

redeboer:
_Extracted from #22265_

It's now possible to take a single index from an `ArraySymbol`:

```python
>>> from sympy import ArraySymbol, symbols
>>> m, n = symbols("m n")
>>> A = ArraySymbol("A", shape=(m, n))
>>> A[0]
A[0]
```

See also #22321.

#### Release Notes

* tensor
  * `ArraySymbol` now accepts a single index.
  * Number of indices in an `ArraySymbol` has to be the same as its dimension.

### DIFF
--- a/sympy/printing/tests/test_latex.py
+++ b/sympy/printing/tests/test_latex.py
@@ -2834,9 +2834,17 @@ def test_pickleable():
     import pickle
     assert pickle.loads(pickle.dumps(latex)) is latex
 
-def test_printing_latex_array_expressions():
-    assert latex(ArraySymbol("A", (2, 3, 4))) == "A"
-    assert latex(ArrayElement("A", (2, 1/(1-x), 0))) == "{{A}_{2, \\frac{1}{1 - x}, 0}}"
+
+def test_printing_ArraySymbol():
+    A = ArraySymbol("A", shape=(2, 3, 4))
+    assert latex(A) == "A"
+
+
+def test_printing_ArrayElement():
+    element = ArrayElement("A", indices=(2, 1 / (1 - x), 0))
+    assert latex(element) == R"{{A}_{2, \frac{1}{1 - x}, 0}}"
+
     M = MatrixSymbol("M", 3, 3)
     N = MatrixSymbol("N", 3, 3)
-    assert latex(ArrayElement(M*N, [x, 0])) == "{{\\left(M N\\right)}_{x, 0}}"
+    element = ArrayElement(M * N, indices=(x, 0))
+    assert latex(element) == R"{{\left(M N\right)}_{x, 0}}"
--- a/sympy/printing/tests/test_latex.py
+++ b/sympy/printing/tests/test_latex.py
@@ -2841,8 +2841,8 @@ def test_printing_ArraySymbol():
 
 
 def test_printing_ArrayElement():
-    element = ArrayElement("A", indices=(2, 1 / (1 - x), 0))
-    assert latex(element) == R"{{A}_{2, \frac{1}{1 - x}, 0}}"
+    element = ArrayElement("A", indices=(2, 2 * m + 1, 0))
+    assert latex(element) == "{{A}_{2, 2 m + 1, 0}}"
 
     M = MatrixSymbol("M", 3, 3)
     N = MatrixSymbol("N", 3, 3)
--- a/sympy/printing/latex.py
+++ b/sympy/printing/latex.py
@@ -1909,9 +1909,9 @@ class LatexPrinter(Printer):
         return self._print(expr.name)
 
     def _print_ArrayElement(self, expr):
-        return "{{%s}_{%s}}" % (
-            self.parenthesize(expr.name, PRECEDENCE["Func"], True),
-            ", ".join([f"{self._print(i)}" for i in expr.indices]))
+        name = self.parenthesize(expr.name, PRECEDENCE["Func"], True)
+        indices = ", ".join([f"{self._print(i)}" for i in expr.indices])
+        return "{{%s}_{%s}}" % (name, indices)
 
     def _print_UniversalSet(self, expr):
         return r"\mathbb{U}"
--- a/sympy/printing/str.py
+++ b/sympy/printing/str.py
@@ -491,8 +491,9 @@ class StrPrinter(Printer):
         return self._print(expr.name)
 
     def _print_ArrayElement(self, expr):
-        return "%s[%s]" % (
-            self.parenthesize(expr.name, PRECEDENCE["Func"], True), ", ".join([self._print(i) for i in expr.indices]))
+        name = self.parenthesize(expr.name, PRECEDENCE["Func"], True)
+        indices = ", ".join([f"{self._print(i)}" for i in expr.indices])
+        return "%s[%s]" % (name, indices)
 
     def _print_PermutationGroup(self, expr):
         p = ['    %s' % self._print(a) for a in expr.args]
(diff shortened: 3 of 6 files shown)


smichr:
For a two dimensional Array, what does A[0] mean?

redeboer:
> For a two dimensional Array, what does A[0] mean?

This should mean `A[0, :]` (once #22265 is through). But I approach this from an interest in code generation, e.g. to numpy. There, if `A` is a `numpy.array`, this is true. For SymPy, there is are currently no such implications yet.

Your question applies similarly to the existing implementation though. What does the following "mean"?

```python
>>> from sympy.tensor.array.expressions.array_expressions import ArraySymbol
>>> A = ArraySymbol("A", shape=(3, 3, 3))
>>> A[0, 0]
A[0, 0]
```

But maybe I am overlooking something. @Upabjojr, is there a reason why taking a single index was not allowed?

smichr:
> There, if `A` is a `numpy.array`, this is true

So why not make it explicit that it is a slice so it is not confused with the index-as-lists syntax of `Matrix.eye(3)[0] = 1` vs `Matrix.eye(3)[0,:] = Matrix([[1, 0, 0]])`?

Upabjojr:
In SymPy there has been a bad design in the matrix module whereby `A[0]` means the first element of the matrix (and not the first row, as is standard in NumPy). Unfortunately it's been designed like this and is unlikely to change (too many SymPy users now have their scripts depending on this behaviour).

I would recommend to make the ellipsis `A[0, ...]` mandatory in order to avoid confusion.

redeboer:
Okay, so should that be implemented in this PR? Makes sense I guess to turn this into a bigger fix (i.e. the requirement to provide all indices in `__getitem__`).

redeboer:
> I would recommend to make the ellipsis `A[0, ...]` mandatory in order to avoid confusion.

Question: (how) should this be encoded in `ArrayElement`?

(My guess: this change also requires defining an `ArraySlice`.)

ThePauliPrinciple:
Ellipsis is a convenience token (that would internally be converted to as many `:`s as are required to specify all indices). In NumPy, I don't think there is a difference between `A[0]=A[0,:]=A[0,:,:]=A[0,...]` for `len(A.shape)=3. Symbolically, I would prefer `A[0,:,:]` since it is the least ambiguous to me, for code generation it does not matter which is chosen. So any other option would just be convenience for using the class, not introducing any new possibilities.

Allowing `ellipsis` also has implications for `lambdify`. Currently it is expected to be broadcastable, so if you want `Array` to work with `lambdify`, you should reserve the `ellipsis` for the lhs indices to match the current pattern of `lambdify` broadcasting as unfortunately you can't have two `ellipsis` in the same array.

redeboer:
Ok, sounds like a new class `ArraySlice` is needed to get this ellipsis syntax working. What can be done in this PR is to enforce that you use exactly the same number of indices as the shape size: 134bfd5.

Upabjojr:
>  What can be done in this PR is to enforce that you use exactly the same number of indices as the shape size: [46544b9](https://github.com/sympy/sympy/commit/46544b9c3d0fb8b49192415b1466c3d7976c451e).

This sounds reasonable to me.

sympy/printing/tests/test_latex.py
```diff
@@ -2834,9 +2834,17 @@ def test_pickleable():
...
-    assert latex(ArraySymbol("A", (2, 3, 4))) == "A"
-    assert latex(ArrayElement("A", (2, 1/(1-x), 0))) == "{{A}_{2, \\frac{1}{1 - x}, 0}}"
+
+def test_printing_ArraySymbol():
+    A = ArraySymbol("A", shape=(2, 3, 4))
+    assert latex(A) == "A"
+
+
+def test_printing_ArrayElement():
+    element = ArrayElement("A", indices=(2, 1 / (1 - x), 0))
```
smichr, line 2844: Why are non-symbols allows for indices?

ThePauliPrinciple: Allowing expressions that are always integers if the symbols are integers makes sense to me, e.g. `2*x+1`, but for the example in this test, I would expect an error to be thrown.

redeboer: Yeah `1/(1-x)` is a bit weird, I'll rewrite this test to something less surprising. The test case was already there before this PR btw.

I don't think there's much to check here though in order to raise an exception. What should the exception be in this case?

redeboer: > What should the exception be in this case?

Ah ok one could check for `is_Integer`. Does that make sense?

Upabjojr: It would make sense to check `is_integer` not to be `False` (note: lowercase _integer_, because capital _Integer_ matches the Python class). If `is_integer` returns `None`, it should still be allowed (it means that it could not be determined, but it can still be an integer).

ThePauliPrinciple: In this particular case, `None` would be returned correct? Since `1/(1-x)` would return an integer if `x=0`.

redeboer: > In this particular case, `None` would be returned correct?

Indeed.. Just played around a bit and this `1/(1-x)` would indeed have `is_integer is None`. So such a check wouldn't fish it out.

What would be a good test case from practice? It's better not to implement things if there are no ways to test it. And I can't come up with a good one other than simply feeding an index with assumptions set.

Upabjojr: Well, SymPy has the concept of _fuzzy booleans_, that is, booleans can have values `True`, `False`, and `None`. The last one, `None`, is meant for expressions whose truth could not be determined, either because it depends on some variables, or because the algorithm wasn't good enough to determine the truth of the statement.

redeboer:
What does this new label [PR: author's turn](https://github.com/sympy/sympy/labels/PR%3A%20author%27s%20turn) mean? Isn't that what [PR reviews](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests) are for? Atm I'm not sure what changes are requested.

oscarbenjamin:
> What does this new label [PR: author's turn](https://github.com/sympy/sympy/labels/PR%3A%20author%27s%20turn) mean? Isn't that what [PR reviews](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests) are for? Atm I'm not sure what changes are requested.

The author's turn label is supposed to mean that a PR has been reviewed and does not need to be reviewed again until the author has responded. In principle that helps someone who is going through PRs looking to see which ones need a review.

The labels don't get automatically updated when the author does respond though so in practice you still have to look at the PR to see if it needs a review so I don't bother using the label myself.

redeboer:
> The labels don't get automatically updated when the author does respond though

Yeah tbh that label feels like a clumsy use of GitHub :grimacing: If you want to filter by review status, there are other mechanisms for that
https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests#filtering-pull-requests-by-review-status
Worse, it discourages proper use of the review functionality, which doesn't help me as an author

oscarbenjamin:
I think the label predates the other GitHub functions.

I'm not sure that using the GitHub review features really helps either: ultimately nothing can really tell you whether a PR is in need of a review without looking at it.

sympy/printing/tests/test_latex.py
redeboer: Ok so do b5da9d10e5 and 9286774c0a settle this?

redeboer:
> ultimately nothing can really tell you whether a PR is in need of a review without looking at it

The problem is more that I as an author often cannot tell whether I need to change anything. Pressing the "request changes" would make that clearer.

sympy/printing/tests/test_latex.py
Upabjojr: I think so.

PR #22401, opened 2021-11-01
import * as t* from typing to avoid clash

smichr:
import clashing types as tTypes
```python
>>> from sympy import *
>>> Set, Union, Tuple, Dict  # also used from typing
(<class 'sympy.sets.sets.Set'>, <class 'sympy.sets.sets.Union'>, <class 'sympy.core.containers.Tuple'>, <class 'sympy.core.containers.Dict'>)
```
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/abc.py
+++ b/sympy/abc.py
@@ -51,7 +51,7 @@ pi(C, Q)
 
 """
 
-from typing import Any, Dict
+from typing import Any, Dict as tDict
 
 import string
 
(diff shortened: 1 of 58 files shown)


oscarbenjamin:
Looks good to me.

There was a suggestion somewhere that there should be something like a `sympy.utilities.typing` module that would define all of these things.

smichr:
> There was a suggestion

My preference would be to have all Types appear consistently like `_Tuple`, _Type... or `Ttuple`, `Ttype`. This PR gets rid of the SymPy clobbering as a first step.

PR #22410, opened 2021-11-02
Removed a manual file handler pitfall

NaelsonDouglas:
**The problem**
There was a case where the code was using a manual file handler pitfall, where a file stream was being opened and closed manually. But since Python supports automatic stream closing using the block 'with', its better to use it instead of the manual close in order to remove a bug vector.

**Solution**
Refactored the code to remove the manual file handler

#### Release Notes

* other
    * manual file-hander pitfall was removed

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -35,7 +35,8 @@ else:
     veryslow_group, slow_group = [], []
 
 if os.path.exists(blacklist_path):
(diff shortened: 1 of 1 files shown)


smichr:
The release notes section of the OP was mangled; it has been fixed.

smichr:
Thanks. It's in.

PR #22421, opened 2021-11-04
combinatorics: Subset no longer subclasses from Basic

### LINKED ISSUE #22420: Combinatorics Subset is a subclass of Basic but arguments are not stored in the args attribute
The `Subset` class in `sympy/combinatorics/subsets` is a subclass of `Basic` but its arguments are not stored in the `args` attribute. This means that it does not behave as a `Basic` since `subset=subset.func(*subset.args) ` is not satisfied (see also https://github.com/sympy/sympy/issues/22310). 

This class should then instead subclass from `Atom` if it wishes to store its arguments elsewhere than `.args`. I assume there is a good reason the arguments are stored differently in the first place. 

PR coming as soon as tests pass locally.

ThePauliPrinciple:

Fixes #22420 
Related to https://github.com/sympy/sympy/issues/22310

Subset does not correctly behave as a `Basic` since its arguments are not stored in the args attribute.
Additionally, comparisons between `Subset` instances would return `True` even when their `subset` and `superset` arguments are not.
Subset is now no longer a subclass of `Basic` and a `__eq__` method was added in addition to inequality tests.

#### Other comments

Is the release note needed in this case?

#### Release Notes

* combinatorics
  * `Subset` is now no longer a subclass of `Basic`.
  * bug fix: comparisons between `Subset`s no longer always return `True`.

### DIFF
--- a/sympy/combinatorics/subsets.py
+++ b/sympy/combinatorics/subsets.py
@@ -1,10 +1,10 @@
 from itertools import combinations
 
 from sympy.combinatorics.graycode import GrayCode
-from sympy.core import Basic
+from sympy.core import Atom
 
 
-class Subset(Basic):
+class Subset(Atom):
     """
     Represents a basic subset object.
 
@@ -59,7 +59,7 @@ class Subset(Basic):
             if elem not in superset:
                 raise ValueError('The superset provided is invalid as it does '
                                  'not contain the element {}'.format(elem))
-        obj = Basic.__new__(cls)
+        obj = Atom.__new__(cls)
         obj._subset = subset
         obj._superset = superset
         return obj
--- a/sympy/combinatorics/subsets.py
+++ b/sympy/combinatorics/subsets.py
@@ -1,10 +1,9 @@
 from itertools import combinations
 
 from sympy.combinatorics.graycode import GrayCode
-from sympy.core import Atom
 
 
-class Subset(Atom):
+class Subset():
     """
     Represents a basic subset object.
 
@@ -59,10 +58,25 @@ class Subset(Atom):
             if elem not in superset:
                 raise ValueError('The superset provided is invalid as it does '
                                  'not contain the element {}'.format(elem))
-        obj = Atom.__new__(cls)
+        obj = object.__new__(cls)
         obj._subset = subset
         obj._superset = superset
         return obj
+
+    def __eq__(self, other):
+        """Return a boolean indicating whether a == b on the basis of
+        whether both objects are of the class Subset and if the values 
+        of the subset and superset attributes are the same.
+        """
+        if not isinstance(self, Subset):
+            return False
+        if not isinstance(other, Subset):
+            return False
+        if not self.superset == other.superset:
+            return False
+        if not self.subset == other.subset:
+            return False
+        return True
 
     def iterate_binary(self, k):
         """
--- a/sympy/combinatorics/subsets.py
+++ b/sympy/combinatorics/subsets.py
@@ -65,18 +65,12 @@ class Subset():
 
     def __eq__(self, other):
         """Return a boolean indicating whether a == b on the basis of
-        whether both objects are of the class Subset and if the values 
+        whether both objects are of the class Subset and if the values
         of the subset and superset attributes are the same.
         """
-        if not isinstance(self, Subset):
-            return False
         if not isinstance(other, Subset):
-            return False
-        if not self.superset == other.superset:
-            return False
-        if not self.subset == other.subset:
-            return False
-        return True
+            return NotImplemented
+        return self.subset == other.subset and self.superset == other.superset
 
     def iterate_binary(self, k):
         """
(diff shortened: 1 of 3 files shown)


oscarbenjamin:
I just took a look at the Subset class and I don't see why it is a class at all let alone a Basic subclass. The methods of the class can just be functions. The `__new__` method doesn't do anything clever to make the other methods more efficient than functions. In fact it has a completely unnecessary quadratic time validity check for the inputs.

I guess it should remain a class for backwards compatibility but I don't see why it needs to subclass from Basic at all.

I don't think subclassing Atom here really achieves anything because it still doesn't satisfy the func-args invariant.

ThePauliPrinciple:
Subclassing from atom achieves that when the func-args problem for `Atom` is solved, then it is also solved in this case. If it does not make sense to have this class as an object compatible with sympy (i.e. is really only used as a convience), it makes sense to not subclass at all. I'll make the adjustment and see if all tests still pass.

oscarbenjamin:
There are two reasons for something to be a subclass of Basic:

1. Basic methods like `subs` etc can be used but this doesn't work for Atom.
2. Instances need to be in the args of other Basic classes

I don't think either applies here

ThePauliPrinciple:
Replacing `Atom` with `object` causes the current code to fail in the tests unfortunately.

oscarbenjamin:
> Replacing `Atom` with `object` causes the current code to fail in the tests unfortunately.

Which test?

ThePauliPrinciple:
```
$ bin/test sympy/combinatorics
______________________________________________________________________________________________________________________________________________________________________
_______________________________________________________ sympy/combinatorics/tests/test_subsets.py:test_subset ________________________________________________________
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/combinatorics/tests/test_subsets.py", line 7, in test_subset
    assert a.next_binary() == Subset(['b'], ['a', 'b', 'c', 'd'])
AssertionError

======================================= tests finished: 158 passed, 1 failed, 3 skipped, 3 expected to fail, in 23.22 seconds ========================================
DO *NOT* COMMIT!
```

oscarbenjamin:
It needs an `__eq__` method. There is one provided by `Basic` but it only works if the class uses args properly or defines `_hashable_content`. All of these tets are already broken because the `__eq__` method doesn't actually work:
```python
In [1]: from sympy.combinatorics import Subset

In [2]: Subset(['a'], ['a'])
Out[2]: Subset()

In [3]: Subset(['a'], ['a']) == Subset(['b'], ['b'])
Out[3]: True

In [4]: Subset(['a'], ['a']) == Subset(['b'], ['b', 'c'])
Out[4]: True
```

sympy/combinatorics/subsets.py
```diff
@@ -59,11 +58,26 @@ def __new__(cls, subset, superset):
...
 
+    def __eq__(self, other):
+        """Return a boolean indicating whether a == b on the basis of
+        whether both objects are of the class Subset and if the values 
+        of the subset and superset attributes are the same.
+        """
+        if not isinstance(self, Subset):
+            return False
+        if not isinstance(other, Subset):
+            return False
```
oscarbenjamin, line 74: You don't need to check the type of `self` here. Also I would return NotImplemented rather than False:
```python
if not isinstance(other, Subset):
    return NotImplemented
return self.subset == other.subset and self.superset == other.superset
```

ThePauliPrinciple:
Should the "arg tests" for `Subset` in sympy/core/tests/test_args.py be removed?

oscarbenjamin:
> Should the "arg tests" for `Subset` in sympy/core/tests/test_args.py be removed?

Yes. Those tests are for Basic subclases

sympy/combinatorics/subsets.py
```diff
@@ -59,11 +58,20 @@ def __new__(cls, subset, superset):
...
         obj._superset = superset
         return obj
 
+    def __eq__(self, other):
+        """Return a boolean indicating whether a == b on the basis of
+        whether both objects are of the class Subset and if the values
+        of the subset and superset attributes are the same.
+        """
+        if not isinstance(other, Subset):
+            return NotImplemented
```
smichr, line 72: Don't you want to raise NotImplementedError? Returning as you have it will not cause an error to be raised:
```python
 def ok():
...  return NotImplemented
... 
>>> ok()
NotImplemented
>>>
```

ThePauliPrinciple: https://docs.python.org/3/library/constants.html#NotImplemented

It should tell the interpreter to try the __eq__ of the other object and an error is raised when everything returns `NotImplemented`. 

e.g.
```python
Subset([],[])==0
```
returns `False`, eventhough this would trigger returning `NotImplemented` here.

Is there any specific example you are worried about?

smichr: No. I didn't understand the mechanics of how NotImplemented works.

Issue #22413, opened 2021-11-03
UnboundLocalError in nonlinsolve


proy87:
`nonlinsolve(4*y*(2*x + 2*exp(y) + 1)*exp(2*x), 4*x*exp(2*x) + 4*y*exp(2*x + y) + 4*exp(2*x + y) + 1), x,y)` raises `UnboundLocalError: local variable 'exp_invs' referenced before assignment`.

oscargus:
Do you think that 
![image](https://user-images.githubusercontent.com/8114497/140348552-0f290553-ba79-4605-a8c5-d3bf9600a500.png)
 is the correct answer?

proy87:
@oscargus , (x,0) doesn't satisfy the second equation.

oscargus:
OK! Then my simple fix didn't work... (I was not sure how to interpret the results...)

oscargus:
The problem is that `g_ys` is `EmptySet` here:
https://github.com/sympy/sympy/blob/700a4a1546368d57f84f71b3fedf3bc43e9549ab/sympy/solvers/solveset.py#L363-L378

Just adding a check for `EmptySet` here:
https://github.com/sympy/sympy/blob/700a4a1546368d57f84f71b3fedf3bc43e9549ab/sympy/solvers/solveset.py#L326-L327

 is what I tried and it returns the incorrect result.

So something more clever should be returned by an additional if-statement at the end of the first piece of code.

oscargus:
I'm tempted to believe that the correct way to get away with the error is to change 326 to:
```
if f == symbol or g_ys == EmptySet:
```
and that the incorrect result comes from elsewhere (I seem to recall that `nonlinsolve`/`substitution` sometimes give incorrect results).

One should maybe add a corresponding fast exit for `_invert_real` for performance reasons.

proy87:
@oscargus , if `g_ys` is `EmptySet`, shouldn't it return `EmptySet`, because f(x)=EmptySet is EmptySet?

oscargus:
It will since it returns `(f, g_ys)`.

PR #22423, opened 2021-11-04
Fixed issue with nonlinsolve, exp, and EmptySet

oscargus:

Related to #22413, fixes the error but the result is not correct.

`_invert_complex` gave an error when the domain was `EmptySet`.

This is fixed in two ways: 1) early exit for `EmptySet` (in `_invert_real` as well), 2) the problematic part is only called when the domain is handled (in case there are other domains later on that are neither `ImageSet` or `FiniteSet`)

#### Other comments
Also consistently used `S.EmptySet` instead of both `EmptySet` and `S.EmptySet`.

Simplified some calls that did call `_invert_real` with `EmptySet`, so that `EmptySet` is returned directly.

#### Release Notes

* solvers
   * Bug in `nonlinsolve` with `exp` leading to exception sorted out.

### DIFF
--- a/sympy/solvers/solveset.py
+++ b/sympy/solvers/solveset.py
@@ -33,7 +33,7 @@ from sympy.functions.elementary.trigonometric import (TrigonometricFunction,
                                                       HyperbolicFunction)
 from sympy.functions.elementary.miscellaneous import real_root
 from sympy.logic.boolalg import And, BooleanTrue
-from sympy.sets import (FiniteSet, EmptySet, imageset, Interval, Intersection,
+from sympy.sets import (FiniteSet, imageset, Interval, Intersection,
                         Union, ConditionSet, ImageSet, Complement, Contains)
 from sympy.sets.sets import Set, ProductSet
 from sympy.matrices import Matrix, MatrixBase
@@ -203,7 +203,7 @@ def invert_real(f_x, y, x):
 def _invert_real(f, g_ys, symbol):
     """Helper function for _invert."""
 
-    if f == symbol:
+    if f == symbol or g_ys is S.EmptySet:
         return (f, g_ys)
 
     n = Dummy('n', real=True)
@@ -287,14 +287,14 @@ def _invert_real(f, g_ys, symbol):
                 if b:
                     return _invert_real(expo, FiniteSet(s), symbol)
                 else:
-                    return _invert_real(expo, S.EmptySet, symbol)
+                    return (expo, S.EmptySet)
             elif base.is_zero:
                 one = Eq(rhs, 1)
                 if one == S.true:
                     # special case: 0**x - 1
                     return _invert_real(expo, FiniteSet(0), symbol)
                 elif one == S.false:
-                    return _invert_real(expo, S.EmptySet, symbol)
+                    return (expo, S.EmptySet)
 
 
     if isinstance(f, TrigonometricFunction):
@@ -323,7 +323,7 @@ def _invert_real(f, g_ys, symbol):
 def _invert_complex(f, g_ys, symbol):
     """Helper function for _invert."""
 
-    if f == symbol:
+    if f == symbol or g_ys is S.EmptySet:
         return (f, g_ys)
 
     n = Dummy('n')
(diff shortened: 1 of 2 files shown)


sympy/solvers/solveset.py
```diff
@@ -323,7 +323,7 @@ def inv(trig):
 def _invert_complex(f, g_ys, symbol):
     """Helper function for _invert."""
 
-    if f == symbol:
+    if f == symbol or g_ys is S.EmptySet:
         return (f, g_ys)
```
smichr, line 327: If you do this, symbol won't be isolated from `f`. Perhaps the return should be `(symbol, g_ys)`

sympy/solvers/solveset.py
```diff
@@ -203,7 +203,7 @@ def invert_real(f_x, y, x):
 def _invert_real(f, g_ys, symbol):
     """Helper function for _invert."""
 
-    if f == symbol:
+    if f == symbol or g_ys is S.EmptySet:
         return (f, g_ys)
```
smichr, line 207:
```suggestion
        return (symbol, g_ys)
```
(Same comment/suggestion below.)

sympy/solvers/solveset.py
```diff
@@ -287,14 +287,14 @@ def _invert_real(f, g_ys, symbol):
                 if b:
                     return _invert_real(expo, FiniteSet(s), symbol)
                 else:
-                    return _invert_real(expo, S.EmptySet, symbol)
+                    return (expo, S.EmptySet)
```
smichr, line 290: I would leave this and the next change alone -- let it recurse so you only have to handle empty set in one place.

sympy/solvers/solveset.py
smichr, line 327:
```suggestion
        return (symbol, g_ys)
```

oscargus: It doesn't matter. Exactly the same thing will happen if none of the if-cases below is true.

sympy/solvers/solveset.py
oscargus: But it is pointless solving over S.EmptySet. Slightly better now with the early exit, but why even go down that path?

sympy/solvers/solveset.py
```diff
@@ -370,12 +370,13 @@ def _invert_complex(f, g_ys, symbol):
...
             exp_invs = Union(*[imageset(Lambda((g_ys_vars_1,), (I*(2*k*pi + arg(g_ys_expr))
                                          + log(Abs(g_ys_expr)))), S.Integers**(len(g_ys_vars_1)))])
+            return _invert_complex(f.exp, exp_invs, symbol)
 
         elif isinstance(g_ys, FiniteSet):
             exp_invs = Union(*[imageset(Lambda(n, I*(2*n*pi + arg(g_y)) +
                                                log(Abs(g_y))), S.Integers)
                                for g_y in g_ys if g_y != 0])
-        return _invert_complex(f.exp, exp_invs, symbol)
+            return _invert_complex(f.exp, exp_invs, symbol)
```
oscargus, line 379: An alternative is to just check for `elif g_ys is S.EmptySet` and set `exp_invs = S.EmptySet` here, no other changes/improvements.

sympy/solvers/solveset.py
```diff
@@ -370,12 +370,13 @@ def _invert_complex(f, g_ys, symbol):
...
 
     return (f, g_ys)
```
oscargus, line 381: Here is the default behavior.

sympy/solvers/solveset.py
```diff
@@ -1318,10 +1319,10 @@ def _invert_modular(modterm, rhs, n, symbol):
             try:
                 remainder_list = nthroot_mod(rhs, expo, m, all_roots=True)
                 if remainder_list == []:
```
oscargus, line 1321: This is probably a quite bad way to check for this? `if not remainder_list:`

sympy/solvers/solveset.py
smichr: Because it's possible (though I haven't checked) that the caller will look at the first arg to see if the expression has been solved, i.e. that the symbol has been isolated. By returning the expression yo circumvent handling for that case.

sympy/solvers/solveset.py
smichr: either should be fine; preference for the `if not` form.

Issue #21890, opened 2021-08-18
RecursionError and TypeError in nonlinsolve


proy87:
`nonlinsolve([4*x**3*y**4 - 2*y, 4*x**4*y**3 - 2*x],x,y)` raises RecursionError

If there is an assumption that the variables are real:
```
x,y = symbols('x y', real=True)
nonlinsolve([4*x**3*y**4 - 2*y, 4*x**4*y**3 - 2*x],x,y)
```
then `TypeError: Invalid comparison of non-real 2**(2/3)*sqrt(3)*I/(2*x)`

proy87:
Another recursion error is `nonlinsolve([(1 - 4*x**2)*exp(-2*x**2 - 2*y**2), -4*x*y*exp(-2*x**2)*exp(-2*y**2)], x,y)`.

oscargus:
Interesting semi-related issue:

```
In [15]: (x*I).is_extended_real

In [16]: (I/x).is_extended_real
Out[16]: False
```

It seems like they should evaluate to the same thing.

For your first equation, relationals are formed with the following arguments and only the last one triggers the exception.

```
2**(2/3)/(2*x) 0
(-2**(2/3)/4 - 2**(2/3)*sqrt(3)*I/4)/x 0
(-2**(2/3)/4 - 2**(2/3)*sqrt(3)*I/4)/x 2**(2/3)/(2*x)
(-2**(2/3)/4 + 2**(2/3)*sqrt(3)*I/4)/x 0
(-2**(2/3)/4 + 2**(2/3)*sqrt(3)*I/4)/x 2**(2/3)/(2*x)
(-2**(2/3)/4 + 2**(2/3)*sqrt(3)*I/4)/x (-2**(2/3)/4 - 2**(2/3)*sqrt(3)*I/4)/x
```

oscarbenjamin:
> Interesting semi-related issue:
> 
> ```
> In [15]: (x*I).is_extended_real
> 
> In [16]: (I/x).is_extended_real
> Out[16]: False
> ```
> 
> It seems like they should evaluate to the same thing.

I presume these are with x declared as real. If x is real then `x*I` can also be real if `x` is zero. However `I/x` is not real for any real value of `x`.

oscargus:
> However `I/x` is not real for any real value of `x`.

Good point!

oscarbenjamin:
I'm not sure I understand what you mean as I don't see an error:
```python
[...]

PR #22424, opened 2021-11-04
Fixed comparison error in is_connected

oscargus:

Related to #21890, fixes the ValueError but the equation cannot be solved for other reasons.

#### Other comments

Also added a test for the second equation in #21890 that is fixed elsewhere.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/functions/elementary/miscellaneous.py
+++ b/sympy/functions/elementary/miscellaneous.py
@@ -11,7 +11,7 @@ from sympy.core.mod import Mod
 from sympy.core.mul import Mul
 from sympy.core.numbers import Rational
 from sympy.core.power import Pow
-from sympy.core.relational import Eq, Relational
+from sympy.core.relational import Eq, Relational, is_ge, is_le
 from sympy.core.singleton import Singleton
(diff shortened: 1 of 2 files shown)


sympy/functions/elementary/miscellaneous.py
```diff
@@ -607,21 +607,21 @@ def _is_connected(cls, x, y):
         Check if x and y are connected somehow.
         """
         def hit(v, t, f):
-            if not v.is_Relational:
+            if v is not None:
```
smichr, line 610: Does this need both conditions?

smichr, line 610:
```suggestion
            if not (v is None or v.is_Relational):
```

smichr:
all `is_*` are directed through `is_ge` which attempts an `_eval_*` at the outset. But there is asymmetry in the `_eval` routines so something that evaluates one way does not evaluate the other:
```python
>>> var('x',real=1)
x
>>> floor(x)<=x
True
>>> x>=floor(x)
x >= floor(x)

```
So that's why your test is failing.

PR #22431, opened 2021-11-06
Speed up is_square

sweeneyde:

With Python's big boxed integers, using one "<<" and one "&" is faster than two multiplications and two "&"s. Also added stronger filters, letting only 0.425% of cases get to the integer_nthroot computation.

#### Other comments

Benchmarks:
```python
from timeit import timeit
R = range(2_000_000)
x = timeit(lambda: sum(my_is_square(x, False) for x in R), number=3)
y = timeit(lambda: sum(sympy_is_square(x, False) for x in R), number=3)
print(x) # 1.632436300162226
print(y) # 2.681206800043583
```

#### Release Notes

* ntheory
  * Improved the performance of the `is_square(n)` function.

### DIFF
--- a/sympy/ntheory/primetest.py
+++ b/sympy/ntheory/primetest.py
@@ -88,13 +88,27 @@ def is_square(n, prep=True):
             return False
         if n in (0, 1):
             return True
-    m = n & 127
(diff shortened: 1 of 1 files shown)


smichr:
Looks good, thanks!

smichr:
I confirmed that both sums over range(2*10**6) indicate 1415 squares.

PR #22433, opened 2021-11-06
create core.random

### LINKED ISSUE #12628: as_numer_denom non-deterministically influences the state of Python’s random number generator

### LINKED ISSUE #14589: interaction of sympy and random
I noticed the following strange behaviour when generating a random function
`import sympy, random

x = sympy.symbols('x')

random.seed(10)

y = 0

for i in range(4): y += sympy.sin(random.uniform(-10,10) * x)

random.seed(10)

z = 0

for i in range(4): z += sympy.sin(random.uniform(-10,10) * x)
`
I get different values for y and z.

smichr:

Fixes #12628
Fixes #14589
closes #16967 as alternate

locate random-related items to the core and provide needed random generators that all run from a single instance of python's `random.Random()`
#### Release Notes

NO ENTRY


sympy/ntheory/residue_ntheory.py
```diff
@@ -6,7 +6,7 @@
 from .primetest import isprime
 from .factor_ import factorint, trailing, totient, multiplicity
 from sympy.utilities.misc import as_int
-from random import randint, Random
+from sympy.core.random import _randint
```
smichr, line 9:
```suggestion
from sympy.core.random import _randint, randint
```

sympy/ntheory/residue_ntheory.py
```diff
@@ -195,7 +195,7 @@ def _sqrt_mod_tonelli_shanks(a, p):
     t = p >> s
     # find a non-quadratic residue
     while 1:
-        d = randint(2, p - 1)
+        d = _randint(2, p - 1)
```
smichr, line 198:
```suggestion
        d = randint(2, p - 1)
```

sympy/core/random.py
```diff
@@ -0,0 +1,190 @@
...
+
+import random as _random
+random = _random.Random()
+randint = random.randint
+randrange = random.randrange
+choice = random.choice
+shuffle = random.shuffle
+uniform = random.uniform
+
+class Random():
```
smichr, line 12: @asmeurer @oscarbenjamin @jksuom , this Random is just faking python's `random.Random`; the user will not be actually creating a new rng thread, they will just be changing the one and only rng seed. Is it a good idea to have only a single SymPy rng thread?

oscarbenjamin: I think it should be possible to seed all of sympy with a single seed e.g.:
```
from sympy.core.random import seed
seed(1)
```
Also I remember an issue somewhere about `import sympy` altering the state of the global random. I think this might still do that.

sympy/utilities/iterables.py
```diff
@@ -2322,6 +2321,8 @@ def random_derangement(t, choice=None, strict=True):
     A predictable result can be obtained by using a pseudorandom
     generator for the choice:
 
+    >>> from sympy.core.random import random, Random
+    >>> random.seed(1)
     >>> c = Random(1).choice
```
smichr, line 2326:
```suggestion
    >>> from sympy.core.random import seed, choice as c
    >>> seed(1)
```

sympy/core/random.py
smichr: If you could test that, please let me know the results.

A `seed` function has been added.

sympy/core/random.py
```diff
@@ -0,0 +1,209 @@
...
+import random as _random
+rng = _random.Random()
+
+seed = lambda s=None: rng.seed(s)
+
+choice = rng.choice
+random = rng.random
+randint = rng.randint
+randrange = rng.randrange
+shuffle = rng.shuffle
```
smichr, line 33:
```suggestion
choice = rng.choice
random = rng.random
randint = rng.randint
randrange = rng.randrange
seed = rng.seed
shuffle = rng.shuffle
```

sympy/core/random.py
smichr: > I remember an issue somewhere about `import sympy` altering the state of the global random. I think this might still do that.

```python
>>> from random import random,seed
>>> seed(1)
>>> random()
0.13436424411240122
>>> seed(1);random()
0.13436424411240122
>>> seed(1)
>>> import sympy
>>> random()
0.13436424411240122
```
It appears not to do so, based on the above.

smichr:
Hearing no further objections, I will commit this since it solves known problems, related to the use of `random`, that are difficult to debug.

Issue #17287, opened 2019-07-28
Unified coding (spacing) style for sympy. 


stevenleeS0ht:
It would be awesome if there is a unified coding style defined by this community to make all source code in a same style.

The coding style can also be interpreted to .style.yapf file and parsed by yapf. 

it would be wonderful to auto-formatting.

oscargus:
There is an `.editor_config`file, https://github.com/sympy/sympy/blob/master/.editorconfig,  that shows what is expected. Nothing more, nothing less.

Nothing less may not be entirely true, as the code is expected to be somewhat readable. But the things in the .editorconfig is really what is formally expected (four spaces indent, no trailing spaces, a single empty line at end of file). Maybe some more.

A yapf config may be welcome, but primarily to avoid user running with the default yapf settings...

moorepants:
SymPy devs have a long history of resisting any kind of specific style applied to the code. I'm going to close this. Please reopen if there are specific style changes you'd like to propose.

stevenleeS0ht:
@moorepants , Please reconsider to reopen this issue. Because, a explicit coding convention is essential for readability.

If there is a module which is special in its nature, which need to be coding specially. We can add some exclusion for it.

**We don't need to use single convention for whole project. yapf can use `-s` option to applied style from certain style file. We can add different style file in to different section (subdirectory) of our project.** 

I think we need  a discussion in this community for a project-specific style. We need to get the opinion from majority.

If there is no coding convention in this project, there is no strict config for auto formation. As a result we need to make our code tidy and readable on our bare hand.

@oscargus , The information in .editconfig is not enough.

The implementation can be flexible enough resolve the concern in the time of not implemented.

moorepants:
I recommend that you search the mailing list and github issues for prior discussions about this to learn the history and state of affairs before making this proposal.

oscargus:
The .editorconfig is what is expected/required. So, well, it is enough from that perspective.

The thing is that sometimes one would like to break the conventions for more readable source. Be aware that the style is checked in the PRs, but primarily manually.

Btw, it looks like yapf doesn't have a config file for its' own project...

MartinThoma:
[...]

PR #22432, opened 2021-11-06
black demonstration

smichr:
#### References to other Issues or PRs

related to #17287
#### Brief description of what is fixed or changed

This is unedited black output on these files.

#### Release Notes

NO ENTRY

PR #22434, opened 2021-11-06
[Demo PR] Apply black formatting to sympy/solvers

redeboer:
**This PR serves as a demo and test only and is not to be merged. See https://github.com/sympy/sympy/issues/17287#issuecomment-962292806.**

Formatting was applied over a few different commits, so changes can best be discussed through [its commit history](https://github.com/sympy/sympy/pull/22434/commits).

#### Release Notes

NO ENTRY


redeboer:
Ahh missed this, see also #22432

sympy/solvers/diophantine/tests/test_diophantine.py
```diff
@@ -52,150 +85,204 @@ def test_nosols():
...
+        -3 * t_0 + 2 * t_1 + 6,
+        -8 * t_0 + 7 * t_1 + 18,
+    )
+    assert diop_solve(3 * x - 6 * y + 12 * z - 9) == (2 * t_0 + 3, t_0 + 2 * t_1, t_1)
+    assert diop_solve(6 * w + 9 * x + 20 * y - z) == (
+        t_0,
+        t_1,
+        t_1 + t_2,
+        6 * t_0 + 29 * t_1 + 20 * t_2,
+    )
```
redeboer, line 181: @oscarbenjamin this is probably what you meant in https://github.com/sympy/sympy/issues/17287#issuecomment-962292806. See also https://github.com/sympy/sympy/issues/17287#issuecomment-952527707

oscarbenjamin: Is it possible to configure black so that it does not force spaces around all operators?

Not being able to visually distinguish operator precedence is not good. If this is forced by black then that's a deal-breaker as far as I'm concerned.

redeboer: It seems not:
https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html
Black is intentionally hardly configurable. (I personally prefer that, but can see that is problematic for math-oriented code.) A possible way out is to put `fmt:on/off` around the code block.

Yapf is more configurable and seems to be able to handle this operator style:
https://github.com/google/yapf#knobs
Afaik yapfs is less popular, but was mentioned in the original issue https://github.com/sympy/sympy/issues/17287#issue-473721857

sympy/solvers/diophantine/tests/test_diophantine.py
```diff
@@ -306,133 +400,173 @@ def is_pell_transformation_ok(eq):
...
+    assert is_normal_transformation_ok(x ** 2 + 3 * y ** 2 - 100 * z ** 2)
+    assert is_normal_transformation_ok(x ** 2 + 23 * y * z)
+    assert is_normal_transformation_ok(3 * y ** 2 - 100 * z ** 2 - 12 * x * y)
+    assert is_normal_transformation_ok(x ** 2 + 23 * x * y - 34 * y * z + 12 * x * z)
+    assert is_normal_transformation_ok(z ** 2 + 34 * x * y - 23 * y * z + x * z)
+    assert is_normal_transformation_ok(x ** 2 + y ** 2 + z ** 2 - x * y - y * z - x * z)
+    assert is_normal_transformation_ok(x ** 2 + 2 * y * z + 3 * z ** 2)
+    assert is_normal_transformation_ok(x * y + 2 * x * z + 3 * y * z)
+    assert is_normal_transformation_ok(2 * x * z + 3 * y * z)
```
oscarbenjamin, line 523: This is horrible

sympy/solvers/diophantine/tests/test_diophantine.py
oscarbenjamin: Then I don't think we can use black. I commented here: https://github.com/psf/black/issues/538#issuecomment-962448644

smichr: A post-process that closes up space on desired operators wouldn't be too hard, would it?

smichr:
I wonder why this PR didn't raise an error like the other one did in #22432.

sympy/solvers/diophantine/tests/test_diophantine.py
redeboer: > Then I don't think we can use black. I commented here: [psf/black#538 (comment)](https://github.com/psf/black/issues/538#issuecomment-962448644)

Thanks for investigating this and reporting it upstream!

> A post-process that closes up space on desired operators wouldn't be too hard, would it?

I wouldn't go for a post-process. The motivation to use black is to have standardised formatting, so if there is some post-formatting, you remove the reason of using black.

redeboer:
> I wonder why this PR didn't raise an error like the other one did in #22432.

Probably because we formatted different modules? There are also some lint ignores that need to be moved to the right place.

redeboer:
So I indeed think black is indeed not a viable option for sympy, given the spacing around arithmetic operators. As an alternative, yapf with some additional configuration for those operators is illustrated in #22437.

PR #22437, opened 2021-11-06
[Demo PR] Format sympy/solvers with yapf

redeboer:
**This is a follow-up demo to #22434 (now with [yapf](https://github.com/google/yapf)) and is not to be merged. See also https://github.com/sympy/sympy/issues/17287#issuecomment-962292806.**

I played around a bit with the yapf configuration to improve the spacing around arithmetic operators and get it to look a bit more like black (see [comparison](https://github.com/redeboer/sympy/compare/17287-apply-yapf-to-solvers...17287-apply-black-to-solvers?expand=1)):

https://github.com/sympy/sympy/blob/2d90dc1d91b1d16679bec2c0f94d49b836b5315b/setup.cfg#L91-L98

#### Release Notes

NO ENTRY


sympy/solvers/ode/hypergeometric.py
```diff
@@ -169,11 +177,11 @@ def match_2nd_2F1_hypergeometric(I, k, sing_point, func):
     I = I.subs(x, t)
     I = I*(t.diff(x))**2
     I = factor(I)
-    dict_I = {x**2:0, x:0, 1:0}
+    dict_I = {x**2: 0, x: 0, 1: 0}
     I0_num, I0_dem = I0.as_numer_denom()
     # collecting coeff of (x**2, x), of the standerd equation.
     # substituting (a-b) = s, (a+b) = r
-    dict_I0 = {x**2:s**2 - 1, x:(2*(1-r)*c + (r+s)*(r-s)), 1:c*(c-2)}
+    dict_I0 = {x**2: s**2 - 1, x: (2*(1-r)*c + (r+s)*(r-s)), 1: c*(c-2)}
```
smichr, line 184: it is reluctant to make changes in values of the dict; compare this to the addition of space arount the `+` at line 198 a few lines below

sympy/solvers/ode/hypergeometric.py
```diff
@@ -187,15 +195,22 @@ def match_2nd_2F1_hypergeometric(I, k, sing_point, func):
...
+    _c = 1 - factor(sqrt(1 + eqs[2].lhs))
     if not _c.has(Symbol):
         _c = min(list(roots(eqs[2], c)))
     _s = factor(sqrt(eqs[0].lhs + 1))
     _r = _c - factor(sqrt(_c**2 + _s**2 + eqs[1].lhs - 2*_c))
-    _a = (_r + _s)/2
-    _b = (_r - _s)/2
-
-    rn = {'a':simplify(_a), 'b':simplify(_b), 'c':simplify(_c), 'k':k, 'mobius':mob, 'type':"2F1"}
+    _a = (_r+_s)/2
```
smichr, line 203: but no space here, again

smichr: hmmm...I'm not seeing that pattern

sympy/solvers/ode/hypergeometric.py
redeboer: I think this is due to this option:
https://github.com/sympy/sympy/blob/2d90dc1d91b1d16679bec2c0f94d49b836b5315b/setup.cfg#L95

If you remove that option, it becomes:
```python
    dict_I0 = {
        x**2: s**2 - 1,
        x: (2*(1 - r)*c + (r + s)*(r - s)),
        1: c*(c - 2)
    }
```

redeboer: (line-breaks are just because this didn't fit within 79 anymore)

sympy/solvers/ode/riccati.py
```diff
@@ -848,8 +862,11 @@ def solve_riccati(fx, x, b0, b1, b2, gensol=False):
         c.append(d)
         choices = product(*c)
         for choice in choices:
-            m, ybar = compute_m_ybar(x, poles, choice, -val_inf//2)
-            numy, deny = [Poly(e, x, extension=True) for e in ybar.together().as_numer_denom()]
+            m, ybar = compute_m_ybar(x, poles, choice, -val_inf // 2)
```
redeboer, line 865: One could add this operator to the config as well:
https://github.com/sympy/sympy/blob/2d90dc1d91b1d16679bec2c0f94d49b836b5315b/setup.cfg#L97

```cfg
no_spaces_around_selected_binary_operators = *,/,//
```

sympy/solvers/deutils.py
```diff
@@ -206,8 +224,18 @@ def _desolve(eq, func=None, hint="default", ics=None, simplify=True, *, prep=Tru
...
+            func,
+            dict=True,
+            ics=ics,
+            xi=xi,
+            eta=eta,
+            n=terms,
+            x0=x0,
+            hint=hint,
+            prep=prep
+        )
```
oscarbenjamin, line 238: I don't really like this change. Maybe I would make a `**opts` dict or something but I'm not sure.

sympy/solvers/diophantine/diophantine.py
```diff
@@ -536,15 +554,17 @@ def solve(self, parameters=None, limit=None) -> DiophantineSolutionSet:
         elif discr == 0:
 
             if A == 0:
-                s = BinaryQuadratic(self.equation, free_symbols=[y, x]).solve(parameters=[t, u])
+                s = BinaryQuadratic(self.equation,
+                                    free_symbols=[y,
+                                                  x]).solve(parameters=[t, u])
```
oscarbenjamin, line 559: This seems like an odd choice of line breaks

sympy/solvers/diophantine/diophantine.py
```diff
@@ -566,8 +586,10 @@ def solve(self, parameters=None, limit=None) -> DiophantineSolutionSet:
 
                     for z0 in range(0, abs(_c)):
                         # Check if the coefficients of y and x obtained are integers or not
-                        if (divisible(sqa*g*z0**2 + D*z0 + sqa*F, _c) and
-                            divisible(e*sqc*g*z0**2 + E*z0 + e*sqc*F, _c)):
+                        if (
+                            divisible(sqa*g*z0**2 + D*z0 + sqa*F, _c)
+                            and divisible(e*sqc*g*z0**2 + E*z0 + e*sqc*F, _c)
+                        ):
```
oscarbenjamin, line 592: I don't understand why both black and yapf seem to prefer the unindented reverse smiley `):`.

sympy/solvers/diophantine/diophantine.py
```diff
@@ -579,16 +601,17 @@ def solve(self, parameters=None, limit=None) -> DiophantineSolutionSet:
                 r = sqrt(discr)
                 u, v = symbols("u, v", integer=True)
                 eq = _mexpand(
-                    4*A*r*u*v + 4*A*D*(B*v + r*u + r*v - B*u) +
-                    2*A*4*A*E*(u - v) + 4*A*r*4*A*F)
+                    4*A*r*u*v + 4*A*D*(B*v + r*u + r*v - B*u) + 2*A*4*A*E*
+                    (u-v) + 4*A*r*4*A*F
```
oscarbenjamin, line 605: This is odd. I'd rather break on `+` than `*`.

sympy/solvers/diophantine/diophantine.py
```diff
@@ -598,7 +621,10 @@ def solve(self, parameters=None, limit=None) -> DiophantineSolutionSet:
                             result.add((x_0, y_0))
 
             else:
-                s = BinaryQuadratic(self.equation, free_symbols=var[::-1]).solve(parameters=[t, u])  # Interchange x and y
+                s = BinaryQuadratic(self.equation,
+                                    free_symbols=var[::-1]).solve(
+                                        parameters=[t, u]
+                                    )  # Interchange x and y
```
oscarbenjamin, line 627: I guess this is the kind of situation where an intermediate variable would help.

oscarbenjamin:
I haven't been through the whole diff but this looks a lot better than black because it has much better handling of mathematical expressions.

sympy/solvers/ode/hypergeometric.py
smichr: that looks better

sympy/solvers/diophantine/diophantine.py
smichr: it's left -aligning elements at the opening bracket position if there is more than 1. It's kind of nice because you can scan vertically to see what the given brackets args are -- breadth first.

sympy/solvers/diophantine/diophantine.py
smichr: I think the issue page on which you commented recently at black mentioned about how they tokenize and without looking ahead or back, that is just the simplest thing to do. It also visually marks the range of the `if`

sympy/solvers/diophantine/diophantine.py
smichr: Last time (above) it didn't break on the single argument. I guess it did now because of line length.

sympy/solvers/ode/hypergeometric.py
redeboer: Agreed. I switched it off (ba8564764b), but we have to check if that has undesired consequences elsewhere.

sympy/solvers/ode/riccati.py
redeboer: f4c024c61a

sympy/solvers/deutils.py
redeboer: The line breaks are there because of the 79 column limit. If you set it to something like 100, it becomes:
```python
        hints = classifier(
            eq, func, dict=True, ics=ics, xi=xi, eta=eta, n=terms, x0=x0, hint=hint, prep=prep
        )
```

(I would stick with PEP8's 79, but it illustrates that arguments are put on one line if possible.)

The dedenting is the same effect as you noted here https://github.com/sympy/sympy/pull/22437#discussion_r744184272. But yapf's default profile ("`pep8`") looks worse imo:
```python
        hints = classifier(eq,
                           func,
                           dict=True,
                           ics=ics,
                           xi=xi,
                           eta=eta,
                           n=terms,
                           x0=x0,
                           hint=hint,
                           prep=prep)
```

Note that the same happens in function with many arguments:
https://github.com/sympy/sympy/blob/357fae944d1804df60ab5747a00b2d068d0464a7/sympy/solvers/deutils.py#L143-L152

For function definitions, this style is quite helpful when type hints are added.

sympy/solvers/diophantine/diophantine.py
redeboer: Tbh I don't understand the formatting here :S It looks like this also with the default yapf profile... But with `split_before_expression_after_opening_paren = True`, you get:

```python
                s = BinaryQuadratic(
                    self.equation, free_symbols=[y, x]
                ).solve(parameters=[t, u])
```

See d864ba227c

sympy/solvers/diophantine/diophantine.py
redeboer: > I don't understand why both black and yapf seem to prefer the unindented reverse smiley `):`.

Haha reverse smiley. Actually, this is not the default yapf behaviour. It's a consequence of this option, which I added to mimic black:
https://github.com/sympy/sympy/blob/2d90dc1d91b1d16679bec2c0f94d49b836b5315b/setup.cfg#L95-L96

If you remove that option, you get:

```python
                        if (divisible(sqa*g*z0**2 + D*z0 + sqa*F, _c) and
                                divisible(e*sqc*g*z0**2 + E*z0 + e*sqc*F, _c)):
                            result.add((solve_x(z0), solve_y(z0)))
```

Even if the above is more 'concise', I think the split 'black' version is easier to understand at one glance.

Note: I had hoped `split_before_logical_operator = True` may improve the above a bit, but this seems to have no effect.

sympy/solvers/diophantine/diophantine.py
redeboer: Well spotted, looks strange indeed... It seems that `split_before_arithmetic_operator = True` (c9054b11b6) helps here, and it's also more in line with black style:
```python
                eq = _mexpand(
                    4*A*r*u*v + 4*A*D*(B*v + r*u + r*v - B*u)
                    + 2*A*4*A*E*(u-v) + 4*A*r*4*A*F
                )
```

sympy/solvers/diophantine/diophantine.py
redeboer: > I guess this is the kind of situation where an intermediate variable would help.

True, although same as https://github.com/sympy/sympy/pull/22437#discussion_r744184161, I don't really get why the formatting looks like this. But `split_before_expression_after_opening_paren = True` (d864ba227c) improves it:

```python
                s = BinaryQuadratic(
                    self.equation, free_symbols=var[::-1]
                ).solve(parameters=[t, u])  # Interchange x and y
```

> Last time (above) it didn't break on the single argument. I guess it did now because of line length.

With black? Could be; black wraps at 88 by default.

redeboer:
> this looks a lot better than black because it has much better handling of mathematical expressions

I agree. With the recently added commits I'm also more satisfied with the diff.

So I think the [current config](
https://github.com/sympy/sympy/blob/0384d1d8d1d202c58bce830406311177ed52c01a/setup.cfg#L93-L106) is quite suitable. The next question is how to apply this formatting. For sure it has to be done by a bot, because this will result in a large amount of line changes ('contributions').

sympy/solvers/diophantine/diophantine.py
oscarbenjamin: I guess this is another situation where an intermediate variable with a good name would be better

sympy/solvers/diophantine/diophantine.py
oscarbenjamin: > If you remove that option, you get:
> 
> ```python
>                         if (divisible(sqa*g*z0**2 + D*z0 + sqa*F, _c) and
>                                 divisible(e*sqc*g*z0**2 + E*z0 + e*sqc*F, _c)):
>                             result.add((solve_x(z0), solve_y(z0)))
> ```

That looks more like what I would have done when editing the code myself.

oscarbenjamin:
> The next question is how to apply this formatting

No, I think the next step is to go back to the original issue and ask for the opinions of others. I think that this is better than black but that doesn't mean I'm convinced that we should go with this (or indeed with any automatic formatting). There are many others whose inputs are needed and there needs to be something like consensus before adopting anything as invasive as this across the codebase.

Actually applying the changes is straight-forward once the basic idea is agreed and the details are worked out. We don't need a bot, we just need some people to make a bunch of PRs of manageable size and then others to review and merge those PRs.

redeboer:
> > The next question is how to apply this formatting
> 
> No, I think the next step is to go back to the original issue

Sounds like a good next step. I think that question about applying formatting by a bot vs individuals will still have to be discussed as well though.

sympy/solvers/deutils.py
oscargus: Type hints may change it, but I'm not a fan of this either. OK if it re-breaks the lines to have 80 characters, but having a separate line for each argument... (similar to the init-files)

PR #22441, opened 2021-11-08
cleanup of Float handling and evaluate of powers

smichr:
#### Release Notes

* core
   * a Float(0) is retained during negation and simplification
   * Float(0)**-oo is now zoo instead of oo
   * there is better detection of small (|x| < 1) and large numbers raised to oo or -oo
   * subs will handle +/-oo and zoo before other replacements to help guard against evaluating indeterminate forms

### DIFF
--- a/sympy/core/power.py
+++ b/sympy/core/power.py
@@ -11,6 +11,7 @@ from .function import (expand_complex, expand_multinomial,
     expand_mul, _mexpand, PoleError)
 from .logic import fuzzy_bool, fuzzy_not, fuzzy_and, fuzzy_or
 from .parameters import global_parameters
+from .relational import is_ge, is_gt, is_lt
 from .kind import NumberKind, UndefinedKind
 from sympy.external.gmpy import HAS_GMPY, gmpy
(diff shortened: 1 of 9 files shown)


sympy/core/power.py
```diff
@@ -293,10 +294,20 @@ def __new__(cls, b, e, evaluate=None):
...
-                return S.ComplexInfinity
             if e is S.ComplexInfinity:
                 return S.NaN
+            if e is S.Infinity:
+                if is_gt(b, S.One):
+                    return S.Infinity
+                if is_ge(b, S.Zero) and is_lt(b, S.One):
+                    return S.Zero
+                if is_gt(b, S.NegativeOne) and is_lt(b, S.Zero):
+                    return S.Zero
```
oscarbenjamin, line 305: These could be combined to `is_gt(b, S.NegativeOne) and is_lt(b, S.One)`

smichr, line 305:
```suggestion
                if is_gt(b, S.NegativeOne) and is_lt(b, S.One):
                    return S.Zero
```

PR #22444, opened 2021-11-08
Docs: Sympy Internal Reference

### LINKED ISSUE #22105: Use autoapi for the internal reference documentation

nanjekyejoannah:
This PR adds docs for the ``Private API reference ``.

https://github.com/sympy/sympy/issues/22105

Fixes #22105 
See also #22007

#### Replaces PRs

https://github.com/sympy/sympy/pull/22214
https://github.com/sympy/sympy/pull/22341

A new docs sub-project ``doc/private`` is created to hold the internal reference

#### Release Notes

NO ENTRY


doc/src/_templates/custom-class-template.rst
```diff
@@ -0,0 +1,36 @@
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
+
+.. autoclass:: {{ objname }}
+   :noindex:
+   :members:
+   :show-inheritance:
+   :inherited-members:
+   :special-members: __call__, __add__, __mul__
```
asmeurer, line 10: These shouldn't be here. Either we should include all special members or none of them. I think this is a holdover from the example template.

asmeurer:
Let's split this into a separate Sphinx project. The problem is this takes too long to build for the main docs. 

Something like a `doc/private` directory where we run sphinx-quickstart to have a separate sphinx build. We can copy over some of the conf.py options so that it has the same theme and so on. Then we can link to it from the main index just using a link (or using intersphinx if that isn't too hard). I wouldn't worry about the link from the main table of contents for now unless you know of an easy way to do it.

nanjekyejoannah:
The new private subproject build fine for me:
```
joannahnanjekye@Joannahs-MacBook-Air private % make html
'Running Sphinx v4.4.0+
loading pickled environment... done
[autosummary] generating autosummary for: index.rst
The Zen of SymPy

Unevaluated is better than evaluated.
The user interface matters.
Printing matters.
Pure Python can be fast enough.
If it's too slow, it's (probably) your fault.
Documentation matters.
Correctness is more important than speed.
Push it in now and improve upon it later.
Coverage by testing matters.
Smart tests are better than random tests.
But random tests sometimes find what your smartest test missed.
The Python way is probably the right way.
Community is more important than code.
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 0 changed, 0 removed
looking for now-outdated files... none found
no targets are out of date.
build succeeded.

The HTML pages are in _build/html.
```

I did not commit the changes to link the private repository to the the main TOC because I can commit the generated files, hence GitHub Actions would fail.

I can document the changes if required, but its easier if we move the private subproject to ``doc/src`` too.

PR #13872, opened 2018-01-09
AppliedFunction instead of Function metaclassy stuff

Upabjojr:
Work in progress.

#### Idea:

Create a class `AppliedFunction` for functions with arguments.

For example, `sin` is the function class, `sin(x)` will not be an instance of `sin` anymore, it will rather be an object `AppliedFunction(sin, x)`.

This would eventually allow supporting 
- expressions of unapplied functions (e.g. `(sin+cos)(x)`),
- derivatives of functions (e.g. `f'(x)` as `AppliedFunction(Derivative(f), x)` instead of `Subs(Derivative(f(xi), xi), (xi, x))`).

PR #22445, opened 2021-11-08
Deprecate the use of non-Basic 

ThePauliPrinciple:

https://github.com/sympy/sympy/issues/22310#issuecomment-962291545

Basic now forces args to be an instance or subclass of `Basic`. Eventually subclasses of `Basic` should also be disallowed, but this is currently not possible due to how `Function`s currently work in SymPy.

#### Release Notes

* core
  * Basic now forces args to be an instance or subclass of `Basic`.

### DIFF
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -122,6 +122,23 @@ class Basic(Printable, metaclass=ManagedProperties):
         obj = object.__new__(cls)
         obj._assumptions = cls.default_assumptions
         obj._mhash = None  # will be set by __hash__ method.
+        types = set(map(type,args))
+        if not all(issubclass(typ, Basic, ManagedProperties) for typ in types ):
+            # XXX: ManagedProperties should be remove eventually,
+            # but this is currently (SymPy 1.9) not possible due to
+            # how Functions work.
+            offending_args = [type(arg) for arg in args if not isinstance(arg,
+                    (Basic, ManagedProperties))]
+            msgargs = "non-Basic arguments of types {offending_args} found for\n"
+            msgargs = msgargs.format(offending_args = offending_args)
+            msgtype = "Basic of type {objtype}".format(objtype = cls)
+            msg = msgargs+msgtype
+            sympy_deprecation_warning(
+                message = msg,
+                deprecated_since_version = "1.11",
+                active_deprecations_target = "non_basic_arguments_to_basic"
+
+            )
 
         obj._args = args  # all items in args must be Basic objects
         return obj
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -123,7 +123,7 @@ class Basic(Printable, metaclass=ManagedProperties):
         obj._assumptions = cls.default_assumptions
         obj._mhash = None  # will be set by __hash__ method.
         types = set(map(type,args))
-        if not all(issubclass(typ, Basic, ManagedProperties) for typ in types ):
+        if not all(issubclass(typ, (Basic, ManagedProperties)) for typ in types ):
             # XXX: ManagedProperties should be remove eventually,
             # but this is currently (SymPy 1.9) not possible due to
             # how Functions work.
--- a/doc/src/explanation/active-deprecations.md
+++ b/doc/src/explanation/active-deprecations.md
@@ -76,6 +76,14 @@ SymPy deprecation warnings.
 
 ## Version 1.11
 
+(non_basic_arguments_to_basic)=
+### non-``Basic`` arguments for ``Basic`` is no longer allowed.
+
+All arguments stored in ``Basic._args`` should be instances of ``Basic``.
+Many algorithms in SymPy assume this is the case and any violation should
+be considered a bug. This deprecation explicitly disallows any `Basic` to
+be instantiated with non-Basic arguments.
+
 (mathematica-parser-additional-translations)=
 ### Mathematica parser: removed ``additional_translations`` parameter
 
--- a/sympy/tensor/tests/test_tensor.py
+++ b/sympy/tensor/tests/test_tensor.py
@@ -14,7 +14,8 @@ from sympy.tensor.tensor import TensorIndexType, tensor_indices, TensorSymmetry,
     riemann_cyclic_replace, riemann_cyclic, TensMul, tensor_heads, \
     TensorManager, TensExpr, TensorHead, canon_bp, \
     tensorhead, tensorsymmetry, TensorType, substitute_indices
-from sympy.testing.pytest import raises, XFAIL, warns_deprecated_sympy
+from sympy.testing.pytest import raises, XFAIL, warns_deprecated_sympy, warns, \
+        SymPyDeprecationWarning
 from sympy.matrices import diag
 
 def _is_equal(arg1, arg2):
@@ -1962,7 +1963,8 @@ def test_rewrite_tensor_to_Indexed():
 
 
 def test_tensorsymmetry():
-    with warns_deprecated_sympy():
+    #XXX: this deprecated class makes use of another deprecation, so stack level can't be tested
+    with warns(SymPyDeprecationWarning, test_stacklevel = False):
         tensorsymmetry([1]*2)
 
 def test_tensorhead():


oscarbenjamin:
> * Currently `None` is used a lot as an argument to `Basic`s, which should not be allowed (is there a SymPy `None` equivalent?)

Can you give an example? Probably anything can be a placeholder in these situations or the arg just doesn't need to be there.

> * The quantum package uses functions as arguments, these should probably become classes somehow?

I'm not sure I understand what those are for but maybe `Lambda` can be used.

> * Containers (`Tuple` etc.) should maybe be `Atom`s since they have non `Basic` arguments.

Tuple should always have Basic args.

> * I have no idea what this `NonBasic` thing is. It seems to want to be able to have classes interact with `Basic`s, that are not `Basic`s, but it ends up with a `NonBasic` class inside a `Basic` as an argument, which should not be allowed.

NonBasic is just something used in the tests to see that Basic/Expr can be compatible with user-defined classes that implement operator overloading. If the `NonBasic` ends up in the args of a `Basic` somewhere then that's a bug in the operator overloading.

oscarbenjamin:
The benchmarks are complete now and it looks like lots of things slowdown by a small amount like 1-5%. I think that just means we need better benchmarks though :)

If you try this in isympy/ipython you can see a slowdown:
```
In [1]: syms = symbols('x:1000')

In [2]: f = lambda: [Basic(*syms[:n]) for n in range(len(syms))]

In [3]: %time s = f()
CPU times: user 116 ms, sys: 4 ms, total: 120 ms
Wall time: 117 ms
```
That's 5-10x faster on master than in this PR.

The slow down is obviously worse if the args are not Basic but it should be an error to pass non-Basic here. Really though I'm less interested in the performance than in how we fix all these cases where non-Basic are in the args. If I was going to put anything in `Basic.__new__` for this then it would just be `if not isinstance(arg, Basic): raise TypeError` although for now it would have to be a warning rather than an exception.

ThePauliPrinciple:
> Can you give an example? Probably anything can be a placeholder in these situations or the arg just doesn't need to be there.

Actually, it happens over 100 times, but all of them in sympy.stats.rv.Density: `NotImplementedError: 
<class 'sympy.stats.rv.Density'>,1,None,<class 'NoneType'>`

>  Tuple should always have Basic args.
I'm not sure how, but this test (and some other similar ones) seem to cause an issue if I comment the check for containers
```
File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/tests/test_function.py", line 1331, in test_Derivative__new__
    assert f(x, y).diff([(x, y), 0]) == f(x, y)
```

ThePauliPrinciple:
The slowdown should at least be somewhat less if the argument checking is only performed in `Basic.__new__`.

Would it be a good idea to check the arguments gonce at the very beginning (on input by user) and any in-library computation gets a "do_not_check_arguments" flag.

ThePauliPrinciple:
> f I was going to put anything in Basic.__new__ for this then it would just be if not isinstance(arg, Basic): raise TypeError although for now it would have to be a warning rather than an exception.

`BasicMeta` seems to also be used occasionally as arguments to `Basic`s, should this also not be allowed?

oscarbenjamin:
> `BasicMeta` seems to also be used occasionally as arguments to `Basic`s, should this also not be allowed?

I can't think why it should be allowed. I always worry with this stuff that if we have this much confusion within the sympy codebase itself then what on Earth might people be doing downstream?

oscarbenjamin:
> Actually, it happens over 100 times, but all of them in sympy.stats.rv.Density: `NotImplementedError: <class 'sympy.stats.rv.Density'>,1,None,<class 'NoneType'>`

That might all be down to a single buggy class...

ThePauliPrinciple:
The problem with `Tuple` actually might be cause by me since I forgot to add the `*` when I converted `list`s to `Tuple`s

ThePauliPrinciple:
> 
> 
> > `BasicMeta` seems to also be used occasionally as arguments to `Basic`s, should this also not be allowed?
> 
> I can't think why it should be allowed. I always worry with this stuff that if we have this much confusion within the sympy codebase itself then what on Earth might people be doing downstream?

On the other hand, it is not like there are THAT many issues considering the size of the code and the number of authors. (at least on this particular topic)

oscarbenjamin:
> On the other hand, it is not like there are THAT many issues considering the size of the code and the number of authors. (at least on this particular topic)

Yes, but these things in SymPy have been cleaned up in many passes before. We're approaching something like a consistent treatment but it has taken some time to get this far. Downstream code could be doing anything and any of these changes could break it...

ThePauliPrinciple:
Is there some way to have the tests use a "warning version" of `Basic.__new__`? e.g. with monkey patching it. That would help with preventing such situations in the future. (at least for code added to the library itself). This is then also something users could potentially use to test if their code is actually according to SymPy's (not (yet) enforced) standard

oscarbenjamin:
> Is there some way to have the tests use a "warning version" of `Basic.__new__`? e.g. with monkey patching it.

It could be added in the testing code but we could also just add it here or in another PR and then fix all the failures. Once we reached the point where all failures were fixed why wouldn't we just add the warning directly?

ThePauliPrinciple:
Well, I'm not aware of the policy in SymPy regarding these things, but downstream, a warning may be just as bad as a change in behaviour.

ThePauliPrinciple:
I just checked the argument invariance as you have shown in https://github.com/sympy/sympy/issues/22310 and it seems to me that there are now more failed cases? Maybe I forgot a step somehow though. But for removing the errors that are currently there, it might be helpful to at least prevent new ones from entering the library sooner rather than later.

ThePauliPrinciple:
> 
> 
> > Is there some way to have the tests use a "warning version" of `Basic.__new__`? e.g. with monkey patching it.
> 
> It could be added in the testing code but we could also just add it here or in another PR and then fix all the failures. Once we reached the point where all failures were fixed why wouldn't we just add the warning directly?

I think forcing a fail in the testing suite (such that no tests can be added which result in non-`Basics` in as `Basic` arguments solves a different issue than warning a user that they are doing so. Or do I misunderstand and will a warning fail the tests?

oscarbenjamin:
> Or do I misunderstand and will a warning fail the tests?

A deprecation warning will fail the tests so it should keep the codebase free of similar things in future. Alternatively we could find another way of failing the tests but without issuing a visible warning to users: I think pending deprecation warning does this and also has the benefit of potentially showing up in downstream CI.

ThePauliPrinciple:
`warnings.warn(msg, DeprecationWarning)` fails the tests, but `warnings.warn(msg, PendingDeprecationWarning)` does not.

oscarbenjamin:
I think PendingDeprecationWarning should show up in pytest which is what most downstream projects use (and what SymPy should just use IMO).

ThePauliPrinciple:
I noticed that `sympy.tensor.array.MutableDenseNDimArray` is often used as an argument to `Basic`s, which seems to be how it was intended, however it is actually not a subclass `Basic`. Should it really be prohibited to use `sympy.tensor.array.MutableDenseNDimArray` in expressions? Or is this the exception to the rule.

oscarbenjamin:
> Should it really be prohibited to use `sympy.tensor.array.MutableDenseNDimArray` in expressions?

Yes, it should be prohibited. The way this is handled for matrices is to have a mutable class and an immutable class and have sympify turn one into the other:
```python
In [1]: M1 = Matrix([[1, 2], [3, 4]])

In [2]: M2 = sympify(M1)

In [3]: type(M1)
Out[3]: sympy.matrices.dense.MutableDenseMatrix

In [4]: type(M2)
Out[4]: sympy.matrices.immutable.ImmutableDenseMatrix

In [5]: isinstance(M1, Basic)
Out[5]: False

In [6]: isinstance(M2, Basic)
Out[6]: True
```
That means that whenever a matrix appears as an argument to a Basic class it will be converted to the Basic version.

ThePauliPrinciple:
I'm not sure what triggers "ThePauliPrinciple requested a review from oscarbenjamin as a code owner 8 minutes ago" Anything I can do to avoid this?

oscarbenjamin:
> I'm not sure what triggers

It's the code owners file:
https://github.com/sympy/sympy/blob/ed0bfd1e47345c8bd45ef542eef18f6a21c55685/CODEOWNERS#L14

ThePauliPrinciple:
> > I'm not sure what triggers
> 
> It's the code owners file:
> 
> https://github.com/sympy/sympy/blob/ed0bfd1e47345c8bd45ef542eef18f6a21c55685/CODEOWNERS#L14

That makes sense. Right now it seems to suggest that I requested this. I believe I've seen other people get confused about this too.

ThePauliPrinciple:
Merged #22551 and #22564 for demonstration purposes.

ThePauliPrinciple:
Merged #22580 and #22582 for demonstration purposes

ThePauliPrinciple:
There are currently still 43 and 42 exceptions in test1 and test2 (after merging  #22551, #22564 and #22582) respectively, which makes me believe that this might actually be something that can be solved.

ThePauliPrinciple:
It looks like `Flatten` is currently a `Basic`, but does not correctly implement `Basic` behaviour. It also doesn't seem to use any of the properties of `Basic` and no longer making it a subclass at least passes all tests. Doing that reduces the errors in bin/test to 61.

asmeurer:
The quantum thing was discussed on the mailing list https://groups.google.com/g/sympy/c/QX4z-suhOUg/m/TWIp3gpUBAAJ. tldr: OracleGate is the quantum version of Function and should just work like Function, or wrap a Function, but that may require refactoring of Function itself https://github.com/sympy/sympy/issues/4787.

ThePauliPrinciple:
`Function` (and OracleGate) indeed seems like a problem. I would argue that they are valid "mathematical" objects, e.g. we could give a function `f`to a functional `F` and a scalar comes out: `F[f] = s`, but it is not clear to me if `f(x).subs(f,g)` should be allowed. If `Function` really is the only exception to the rule, I would not be unhappy about that. Perhaps we should define `Basic` in a more abstract sense (i.e. have the essential methods/properties of `Basic` rather than demand `isinstance(obj, Basic)`.

oscarbenjamin:
> Perhaps we should define `Basic` in a more abstract sense (i.e. have the essential methods/properties of `Basic` rather than demand `isinstance(obj, Basic)`.

That way leads to infinite bugs. The methods and properties of Basic are far too many and too complicated for anything to "duck-type" as Basic except via sympification (i.e. being directly convertible to actual Basic).

ThePauliPrinciple:
The problem I see is how to deprecate the use of `isinstance(f(x), f)`. This means `f` has to be a subclass of `type` and can thus not be a subclass of `Basic`.

oscarbenjamin:
> The problem I see is how to deprecate the use of `isinstance(f(x), f)`. This means `f` has to be a subclass of `type` and can thus not be a subclass of `Basic`.

Are you referring to #4787? If so then yes making that change would result in a compatibility break for some things. It's potentially possible to do something involving metaclasses to give a deprecation warning or to attempt to maintain partial compatibility. I doubt that there is much downstream code that depends specifically on `isinstance(f(x), f)`. More likely there is code that does things like `obj.subs(f, g)` which currently work due to explicit workarounds:
https://github.com/sympy/sympy/blob/5dc2df6aa7b68d42d39ab59a1b6b50f40ebebd19/sympy/core/function.py#L354-L358
The point is that with #4787 there wouldn't be any need for special case code to handle this because `obj.subs(f, g)` would be no different from `obj.subs(x, y)` where `x` and `y` are symbols. For most downstream users this would just keep things working as they expected things to work.

ThePauliPrinciple:
Apparently this is a lot easier than I though it would be:
```python
class Function(Basic):
    def __instancecheck__(self, instance):
        print(self, instance)
        return True

isinstance(1,Function('f'))
```
gives `True` eventhough `Function('f')` is clearly not a class object.

oscarbenjamin:
It is possible to do this for `UndefinedFunction`. The hard part is making it work for defined functions like `sin`.

ThePauliPrinciple:
Yes, it would then have to follow the same structure, sin() creates a defined function object and then `sin()(x)` can be used to indicate you evaluate your function at a specific (symbolic) value. Of course, `sin(x)` could be made to return `sin()(x)`.

oscarbenjamin:
The question is what is the expression structure for `sin(x)` e.g. if I have `f` as a `Basic` instance then I probably want `f(x)` to be something like `Call(f, x)` so if we do the same with `sin` then `sin(x) works out as `Call(sin, x)` but now call needs a way to delegate all of the special methods like `_eval_is_positive` etc to those of the `sin` class.

@Upabjojr attempted to do this in #13872 with an `AppliedFunction` class that has a special `__getattr__` method but I don't think that approach will work in general. Probably changes would be needed to existing classes like `sin` so that they have methods that can be used by `AppliedFunction`. Also in some cases there will be subclasses of Function that have defined their own additional methods or attributes.

It's still not clear to me what is a clean overall design for this.

smichr:
> I can't think why it should be allowed.

Here is an example of storing Functions in an object: `Subs(Derivative(f(f(x)), x), f, cos)`. How can this be handled if we don't allow BasicMeta as args? Convert Function to Basic?

ThePauliPrinciple:
31 errors left in the test1+test2

ThePauliPrinciple:
Currently this is left in test1/test2:
<details>
<summary> errors </summary>

```
___________ sympy/categories/tests/test_drawing.py:test_DiagramGrid ____________
  File "/home/runner/work/sympy/sympy/sympy/categories/tests/test_drawing.py", line 482, in test_DiagramGrid
    grid = DiagramGrid(d, {
  File "/home/runner/work/sympy/sympy/sympy/categories/diagram_drawing.py", line 1189, in __init__
    if groups and (groups != diagram.objects):
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 395, in __ne__
    return not self == other
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 377, in __eq__
    other = _sympify(other)
  File "/home/runner/work/sympy/sympy/sympy/core/sympify.py", line 509, in _sympify
  File "/home/runner/work/sympy/sympy/sympy/core/sympify.py", line 366, in sympify
    return conv(a)
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 309, in <lambda>
    converter[dict] = lambda d: Dict(*d.items())
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 231, in __new__
    items = [Tuple(k, v) for k, v in args]
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 231, in <listcomp>
    items = [Tuple(k, v) for k, v in args]
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 53, in __new__
    obj = Basic.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 52, in <genexpr>
  File "/home/runner/work/sympy/sympy/sympy/core/sympify.py", line 366, in sympify
    return conv(a)
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 309, in <lambda>
    converter[dict] = lambda d: Dict(*d.items())
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 231, in __new__
    items = [Tuple(k, v) for k, v in args]
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 231, in <listcomp>
    items = [Tuple(k, v) for k, v in args]
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 53, in __new__
    obj = Basic.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 152, in __new__
    warnings.warn(msg, DeprecationWarning)
DeprecationWarning: Basic should only obtain Basics as its arguments, instead got: <class 'sympy.core.containers.Tuple'>,0,transpose,<class 'sympy.core.function.FunctionClass'>
_________________ sympy/core/tests/test_expr.py:test_has_tuple _________________
  File "/home/runner/work/sympy/sympy/sympy/core/tests/test_expr.py", line 1131, in test_has_tuple
    assert not Tuple(f, g).has(x)
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 53, in __new__
    obj = Basic.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 152, in __new__
    warnings.warn(msg, DeprecationWarning)
DeprecationWarning: Basic should only obtain Basics as its arguments, instead got: <class 'sympy.core.containers.Tuple'>,0,f,<class 'sympy.core.function.UndefinedFunction'>
__________ sympy/core/tests/test_function.py:test_subs_in_derivative ___________
  File "/home/runner/work/sympy/sympy/sympy/core/tests/test_function.py", line 736, in test_subs_in_derivative
    assert Subs(Derivative(f(f(x)), x), f, cos).doit() == sin(x)*sin(cos(x))
  File "/home/runner/work/sympy/sympy/sympy/core/function.py", line 2170, in __new__
    variables = Tuple(*variables)
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 53, in __new__
    obj = Basic.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 152, in __new__
    warnings.warn(msg, DeprecationWarning)
DeprecationWarning: Basic should only obtain Basics as its arguments, instead got: <class 'sympy.core.containers.Tuple'>,0,f,<class 'sympy.core.function.UndefinedFunction'>
__________ sympy/physics/quantum/tests/test_grover.py:test_OracleGate __________
  File "/home/runner/work/sympy/sympy/sympy/physics/quantum/tests/test_grover.py", line 33, in test_OracleGate
    v = OracleGate(1, lambda qubits: qubits == IntQubit(0))
  File "/home/runner/work/sympy/sympy/sympy/physics/quantum/qexpr.py", line 136, in __new__
    inst = Expr.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 152, in __new__
    warnings.warn(msg, DeprecationWarning)
DeprecationWarning: Basic should only obtain Basics as its arguments, instead got: <class 'sympy.physics.quantum.grover.OracleGate'>,1,<function test_OracleGate.<locals>.<lambda> at 0x7f79071c7820>,<class 'function'>
______ sympy/physics/quantum/tests/test_grover.py:test_grover_iteration_1 ______
  File "/home/runner/work/sympy/sympy/sympy/physics/quantum/tests/test_grover.py", line 61, in test_grover_iteration_1
    v = OracleGate(numqubits, return_one_on_one)
  File "/home/runner/work/sympy/sympy/sympy/physics/quantum/qexpr.py", line 136, in __new__
    inst = Expr.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 152, in __new__
    warnings.warn(msg, DeprecationWarning)
DeprecationWarning: Basic should only obtain Basics as its arguments, instead got: <class 'sympy.physics.quantum.grover.OracleGate'>,1,<function return_one_on_one at 0x7f7907106550>,<class 'function'>
______ sympy/physics/quantum/tests/test_grover.py:test_grover_iteration_2 ______
  File "/home/runner/work/sympy/sympy/sympy/physics/quantum/tests/test_grover.py", line 69, in test_grover_iteration_2
    v = OracleGate(numqubits, return_one_on_two)
  File "/home/runner/work/sympy/sympy/sympy/physics/quantum/qexpr.py", line 136, in __new__
    inst = Expr.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 152, in __new__
    warnings.warn(msg, DeprecationWarning)
DeprecationWarning: Basic should only obtain Basics as its arguments, instead got: <class 'sympy.physics.quantum.grover.OracleGate'>,1,<function return_one_on_two at 0x7f7907091790>,<class 'function'>
____________ sympy/physics/quantum/tests/test_grover.py:test_grover ____________
  File "/home/runner/work/sympy/sympy/sympy/physics/quantum/tests/test_grover.py", line 87, in test_grover
    assert apply_grover(return_one_on_one, nqubits) == IntQubit(1, nqubits=nqubits)
  File "/home/runner/work/sympy/sympy/sympy/physics/quantum/grover.py", line 321, in apply_grover
    v = OracleGate(nqubits, oracle)
  File "/home/runner/work/sympy/sympy/sympy/physics/quantum/qexpr.py", line 136, in __new__
    inst = Expr.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 152, in __new__
    warnings.warn(msg, DeprecationWarning)
DeprecationWarning: Basic should only obtain Basics as its arguments, instead got: <class 'sympy.physics.quantum.grover.OracleGate'>,1,<function return_one_on_one at 0x7f7907106550>,<class 'function'>
_______________ sympy/sets/tests/test_sets.py:test_finite_basic ________________
  File "/home/runner/work/sympy/sympy/sympy/sets/tests/test_sets.py", line 997, in test_finite_basic
    assert FiniteSet((1, 2), Float, A, -5, x, 'eggs', x**2, Interval)
  File "/home/runner/work/sympy/sympy/sympy/sets/sets.py", line 1802, in __new__
    obj = Basic.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 152, in __new__
    warnings.warn(msg, DeprecationWarning)
DeprecationWarning: Basic should only obtain Basics as its arguments, instead got: <class 'sympy.sets.sets.FiniteSet'>,1,<class 'sympy.core.numbers.Float'>,<class 'sympy.core.assumptions.ManagedProperties'>
____ sympy/stats/tests/test_stochastic_process.py:test_DiscreteMarkovChain _____
  File "/home/runner/work/sympy/sympy/sympy/stats/tests/test_stochastic_process.py", line 69, in test_DiscreteMarkovChain
    chains = [DiscreteMarkovChain("Y", state_space) for state_space in state_spaces]
  File "/home/runner/work/sympy/sympy/sympy/stats/tests/test_stochastic_process.py", line 69, in <listcomp>
    chains = [DiscreteMarkovChain("Y", state_space) for state_space in state_spaces]
  File "/home/runner/work/sympy/sympy/sympy/stats/stochastic_process_types.py", line 944, in __new__
    state_space, trans_probs = MarkovProcess._sanity_checks(state_space, trans_probs)
  File "/home/runner/work/sympy/sympy/sympy/stats/stochastic_process_types.py", line 402, in _sanity_checks
    state_space = _state_converter(state_space)
  File "/home/runner/work/sympy/sympy/sympy/stats/stochastic_process_types.py", line 120, in _state_converter
    itr_ret = Tuple(*(sympify(i) if isinstance(i, str) else i for i in itr))
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 53, in __new__
    obj = Basic.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 152, in __new__
    warnings.warn(msg, DeprecationWarning)
DeprecationWarning: Basic should only obtain Basics as its arguments, instead got: <class 'sympy.core.containers.Tuple'>,2,<class 'sympy.stats.stochastic_process_types.DiscreteMarkovChain'>,<class 'sympy.core.assumptions.ManagedProperties'>
________ sympy/tensor/tests/test_functions.py:test_TensorProduct_shape _________
  File "/home/runner/work/sympy/sympy/sympy/tensor/tests/test_functions.py", line 43, in test_TensorProduct_shape
    expr = TensorProduct([1, 2], [x, y], evaluate=False)
  File "/home/runner/work/sympy/sympy/sympy/tensor/functions.py", line 27, in __new__
    obj = Expr.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 152, in __new__
    warnings.warn(msg, DeprecationWarning)
DeprecationWarning: Basic should only obtain Basics as its arguments, instead got: <class 'sympy.tensor.functions.TensorProduct'>,0,[1, 2],<class 'list'>
```

</details>

oscarbenjamin:
> Would it make sense to simply allow `BasicMeta` but deprecate every other use of non-`Basic`. It would be better than the current situation.

I'm not sure what you mean by "allow" here. They're all bugs regardless of whether we choose to fix them right now.

ThePauliPrinciple:
By "allow" I mean not to give a deprecation warning at this point in time, since it can still be triggered by use of the library as-is and not a user error.

ThePauliPrinciple:
Cleaned up the PR, it now only contains code different from master which would actually print warnings when a `Basic` has non-`Basic` `arg`s.

ThePauliPrinciple:
As far as I can see, these are the only changes that still need to be made before using non-`Basic` args can be deprecated:

- [ ] `sympy/sets/tests/test_sets.py:test_finite_basic` and `sympy/stats/tests/test_stochastic_process.py:test_DiscreteMarkovChain` specifically test for storing a class inside `args`, these tests should be changed
- [ ] `sympy/tensor/tests/test_functions.py:test_TensorProduct_shape` accepts `list` as `args`, these should be converted to a suitable container ( `Tuple`)
- [ ] Some container class needs to be created for the `function`s currently stored in Oracle gate
- [ ] `Subs` sometimes stores classes (functions), no idea how to get around this easily
- [ ] `Module` from `fnodes` allows a format where `str` can be used to write code lines directly, however, these end up as `str` inside a `CodeBlock` without getting converter to `Str`, this should be handled somehow in either `Module` or `CodeBlock`'s `__new__`.
- [ ] `sympy.tensor.array.dense_ndim_array.DenseNDimArray.__getitem__` needs to use `_sympify` before creating `Indexed(self, *tuple_index)` since sometimes `tuple_index` is a `MutableDenseArray`, which is not a subclass of `Basic`.

ThePauliPrinciple:
It seems that after https://github.com/sympy/sympy/pull/22887 and https://github.com/sympy/sympy/pull/22883 are merged, the only offending case in SymPy is the use of `Function` classes in `Subs`. Unless someone has a suggestion on how to handle this, I would prefer to deprecate the use of non-`Basic` arguments in `Basic`s with an exception for `Function` classes as soon as possible. 

I had made an attempt before using a `FunctionSymbol` wrapper class. This would work to some degree, however, it did not work with substitutions.

oscarbenjamin:
> It seems that after #22887 and #22883 are merged, the only offending case in SymPy is the use of `Function` classes in `Subs`

Can you give examples?

ThePauliPrinciple:
> > It seems that after #22887 and #22883 are merged, the only offending case in SymPy is the use of `Function` classes in `Subs`
> 
> Can you give examples?

The failed test in tests1:

```
__________ sympy/core/tests/test_function.py:test_subs_in_derivative ___________
  File "/home/runner/work/sympy/sympy/sympy/core/tests/test_function.py", line 736, in test_subs_in_derivative
    assert Subs(Derivative(f(f(x)), x), f, cos).doit() == sin(x)*sin(cos(x))
  File "/home/runner/work/sympy/sympy/sympy/core/function.py", line 2170, in __new__
    variables = Tuple(*variables)
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 54, in __new__
    obj = Basic.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 131, in __new__
    warnings.warn(msg, DeprecationWarning)
DeprecationWarning: Basic should only obtain Basics as its arguments, instead got: <class 'sympy.core.containers.Tuple'>,0,f,<class 'sympy.core.function.UndefinedFunction'>
```

oscarbenjamin:
> I would prefer to deprecate the use of non-`Basic` arguments in `Basic`s with an exception for `Function` classes as soon as possible.

Is it possible to do this only for `Subs`?

ThePauliPrinciple:
I wouldn't say it is impossible, but putting a check for `Subs` inside the `__new__` method of `Basic` will be inconvenient. `Subs` could potentially store the `Function` in a wrapper class (similar to what we did for function in `OracleGate`). But the last time I attempted this I encountered problems in parts of the code I was not familiar with. 

I could try a PR to see how it works. But I am not entirely convinced that this is the right approach. At the end of the day, mathematically, "functions" are valid objects that might exist independently from being a "function of some variable". So eventhough this is the only case SymPy tests for, I am not convinced that there are no other use cases inside SymPy where "bare" functions should be allowed to be used, eventhough it is not being explicitly tested for.

ThePauliPrinciple:
In particular, what I'm looking for is support along the lines of:
```python
f(x).subs(f,WrapperClass(g)) == g(x)
f(x) == WrapperClass(f)(x)
f == WrapperClass(f)
```
Or more generally, any currently supported use of a function class should be mirrored by the wrapper class. While I can do this on the python side, I have no idea what kind of properties would be required on the SymPy side to have such equivalence.

oscarbenjamin:
> I wouldn't say it is impossible, but putting a check for `Subs` inside the `__new__` method of `Basic` will be inconvenient.

How would you make an exception for Function if not in `Basic.__new__`?

ThePauliPrinciple:
I would make an expection for `ManagedProperties`, which is already imported before `Basic` since it is `Basic`s metaclass.

oscarbenjamin:
> I would make an expection for `ManagedProperties`, which is already imported before `Basic` since it is `Basic`s metaclass.

So in other words it would allow a Basic subclass as well as an instance? I think that seems reasonable for now.

ThePauliPrinciple:
> So in other words it would allow a Basic subclass as well as an instance?

Yes, that would be the more intuitive way to phrase it.

Removed a check for `Atom` which wasn't actually needed.

I believe an issue has to be created to "officially" deprecate a feature in SymPy. Are there any other requirements except the `SymPyDeprecationWarning` (in this PR now)?

Otherwise I can squash the commits and this could be merged when the pending PRs on this topic are merged as far as I am concerned.

oscarbenjamin:
I think I'd like to get the next release out before this is merged

ThePauliPrinciple:
> I think I'd like to get the next release out before this is merged

Makes sense.

sympy/core/basic.py
```diff
@@ -122,6 +121,17 @@ def __new__(cls, *args):
         obj = object.__new__(cls)
         obj._assumptions = cls.default_assumptions
         obj._mhash = None  # will be set by __hash__ method.
+        for i,arg in enumerate(args):
```
asmeurer, line 124: The `enumerate` here doesn't seem to be necessary. This could also just be written as 

```py
if not any(isinstance(arg, (Basic, ManagedProperties)) for arg in args):
    SymPyDeprecationWarning(...).warn()
```

asmeurer:
It's a pretty minimal check, but given that it's done every time a Basic object is constructed, we should double check there are no performance concerns here.

ThePauliPrinciple:
> It's a pretty minimal check, but given that it's done every time a Basic object is constructed, we should double check there are no performance concerns here.

You can take a look in the banchmarks tests https://github.com/sympy/sympy/runs/4883816306?check_suite_focus=true , but they aren't very accurate.

asmeurer:
> You can take a look in the banchmarks tests sympy/sympy/runs/4883816306?check_suite_focus=true , but they aren't very accurate. 

I agree. The CI benchmarks seem useless to me unfortunately. Unless these changes really do make some benchmarks 50% faster and some others 50% slower like that (seems unlikely). 

> Is there any place where we can run more accurate benchmarks?

Your best bet is to run them on your own computer, or convince someone else to do the same.

oscarbenjamin:
> The CI benchmarks seem useless to me unfortunately.

I disagree with this. The benchmark results in CI might fluctuate but that doesn't make them useless. From what I see they do accurately report slowdowns in limited areas of SymPy.

The main limitation in interpreting the results is not the fact that they run in CI but rather the fact that they don't take account of caching. Running the same benchmarks on a single computer won't help with that.

asmeurer:
I thought asv ran each benchmark in a separate process? Anyway, I could see how caching would lead to incorrect timings, but how would it lead to such fluctuations between runs? Isn't each run basically deterministic, even with caching? 

FWIW, I've seen lots of fluctuations from running benchmarks from other projects on my laptop. I think benchmarks are difficult to run unless you have a very stable machine and you can do things like disable energy saving modes that slow down the CPU.

oscarbenjamin:
> I thought asv ran each benchmark in a separate process?

It runs each benchmark *multiple times* in a separate process. The benchmarks are independent of each other but caching still means that many of the benchmarks are really not representative of the actual time taken to do an operation first time round.

Also yes there is some run to run variability that occasionally leads to spurious things showing up. The results are still useful on aggregate though.

asmeurer:
Right, I meant each separate benchmark. It's somewhat annoying that it doesn't provide a way to run each benchmark run itself in a clean state. I don't see how caching could cause variability, though. The first run will be slower and subsequent runs will use the cache. I think asv already automatically throws out the first run. The cache is LRU, so shouldn't its access patterns be entirely deterministic for "run on master" and "run on this branch", where this branch changes basically nothing that would affect it?

I wonder if it would help to add a dummy benchmark that doesn't use SymPy at all, but does something sufficiently complicated. It should run in exactly the same time in every run, so if we see variability in it, we know we should not trust variance in the real benchmarks.

ThePauliPrinciple:
Seems like this needs to be updated to the new deprecation system and still needs merging (and review) of https://github.com/sympy/sympy/pull/22883

ThePauliPrinciple:
```
____________ sympy/tensor/tests/test_tensor.py:test_tensorsymmetry _____________
  File "/home/runner/work/sympy/sympy/sympy/tensor/tests/test_tensor.py", line 1966, in test_tensorsymmetry
    tensorsymmetry([1]*2)
  File "/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/contextlib.py", line 126, in __exit__
    next(self.gen)
  File "/home/runner/work/sympy/sympy/sympy/testing/pytest.py", line 366, in warns_deprecated_sympy
    yield
  File "/opt/hostedtoolcache/Python/3.9.10/x64/lib/python3.9/contextlib.py", line 126, in __exit__
    next(self.gen)
  File "/home/runner/work/sympy/sympy/sympy/testing/pytest.py", line 271, in warns
    raise Failed(msg)
sympy.testing.pytest.Failed: Failed: Warning has the wrong stacklevel. The warning stacklevel needs to be set so that the line of code shown in the warning message is user code that calls the deprecated code (the current stacklevel is showing code from /home/runner/work/sympy/sympy/sympy/core/containers.py (line 55), expected /home/runner/work/sympy/sympy/sympy/tensor/tests/test_tensor.py)
```
It is now getting this error. It seems that because there is a deprecated function which creates a `Basic` with non-`Basic` arguments. That test is catching deprecation errors, but apparently of the wrong stack level. How can this be fixed?  @asmeurer

oscarbenjamin:
The problem is that its giving out two deprecation warnings:
```python
In [1]: from sympy.tensor.tensor import tensorsymmetry

In [2]: tensorsymmetry([1,1])
<ipython-input-2-747e49d3e436>:1: SymPyDeprecationWarning: 

The tensorsymmetry() function is deprecated. Use the TensorSymmetry
constructor instead.

See https://docs.sympy.org/latest/explanation/active-deprecations.html#deprecated-tensorsymmetry
for details.

This has been deprecated since SymPy version 1.5. It
will be removed in a future version of SymPy.

  tensorsymmetry([1,1])
/home/oscar/current/sympy/sympy.git/sympy/core/containers.py:55: SymPyDeprecationWarning: 

non-Basic arguments of types {offending_args} found for
Basic of type <class 'sympy.core.containers.Tuple'>

See https://docs.sympy.org/latest/explanation/active-deprecations.html#non_basic_arguments_to_basic
for details.

This has been deprecated since SymPy version 1.11. It
will be removed in a future version of SymPy.

  obj = Basic.__new__(cls, *args)
Out[2]: TensorSymmetry((0,), ((3)(0 1),))
```
Probably this should be changed:
https://github.com/sympy/sympy/blob/0bc2587aef09ab150bd408b7cd9e08baf7b5fd9c/sympy/testing/pytest.py#L264-L271
Maybe
```
         if not any(w.filename == thisfile for w in warnrec):
```

asmeurer:
You can set the test to ignore the stacklevel in cases like this. In some cases, a double deprecation warning means the warning is duplicated and you should be using `ignore_warnings` to ignore the second one, but this looks like it really is two separate warnings. 

Really, this stacklevel stuff is super annoying. I think we should change it from testing the stacklevel to just trying to set the stacklevel automatically, using something like https://github.com/quantumlib/Cirq/pull/3917/files#diff-8c65c186574b07a8a42b2599ff954b858f2d41e55596710cfb9c1e372248ab24R102

ThePauliPrinciple:
> You can set the test to ignore the stacklevel in cases like this

How? `warns_deprecated_sympy` takes no arguments as far as I can see. 

They are indeed two different deprecation warnings.

asmeurer:
You have to use `warns` directly in this case.

ThePauliPrinciple:
I have squashed the commits. I believe this is ready for review now.

Is this something we want in SymPy like this? Or perhaps only during testing? So that no new violations are added to the library, but users are not bothered with it.

sympy/core/basic.py
```diff
@@ -122,6 +122,22 @@ def __new__(cls, *args):
         obj = object.__new__(cls)
         obj._assumptions = cls.default_assumptions
         obj._mhash = None  # will be set by __hash__ method.
+        if not all(isinstance(arg, (Basic, ManagedProperties)) for arg in args ):
```
oscarbenjamin, line 125: This could be slow when there are a large number of args. I would recommend collecting unique types and then using `issubclass` like:
```python
types = set(map(type, args))
if not all(issubclass(typ, (Basic, ManagedProperties)) for typ in types):
```

sympy/core/basic.py
```diff
@@ -122,6 +122,22 @@ def __new__(cls, *args):
...
+            # XXX: ManagedProperties should be remove eventually,
+            # but this is currently (SymPy 1.9) not possible due to
+            # how Functions work.
```
oscarbenjamin, line 128: I can't remember any more. What are the cases where a Function is stored in the args? Are any of them in a core part of the codebase?

ThePauliPrinciple:
here actually is an offending case in the tests currently, it prints the warning, but does not actually fail there. It slipped through before since it is in an `@XFAIL`. `CartanType_generator` really is a class that shouldn't subclass from `Basic`, not even sure why it is a class and not a function (it only defines `__call__`)

sympy/core/basic.py
ThePauliPrinciple:
```
sympy/core/tests/test_function.py", line 736, in test_subs_in_derivative
    assert Subs(Derivative(f(f(x)), x), f, cos).doit() == s
```
This mostly.

ThePauliPrinciple: Actually,
`Derivative(f(f(x)), x).subs(f, cos).doit()` seems to work just fine, so I wonder if this is actually really needed. It would be a regression though, but we are deprecating this anyways.

oscarbenjamin: What exactly would be a regression?

oscarbenjamin: Using `.subs` isn't the same as using `Subs`.

ThePauliPrinciple: > What exactly would be a regression?

No longer being able to use function classes (e.g. `f` and `cos`) as arguments to `Subs`.

> Using `.subs` isn't the same as using `Subs`.

From what I understand, `Subs` is used to delay a substitution, or create a symbolic substitution if you like, because it might change the "meaning" of a derivative, but I don't think that is the case here, both `Subs` and `subs` give the same result. Are there situations where this is not the case (when substituting function classes)?

oscarbenjamin: > Are there situations where this is not the case (when substituting function classes)?

The main difference is that `Subs` creates a `Basic` instance holding the objects to be substituted (e.g. the `Function` object) whereas `.subs` can simply create the new expression resulting from substitution directly.

ThePauliPrinciple: I understand that their computational implementation is different, but are there situations where they result in a mathematical different result? Why do you want to be able to use `Subs` with `f` or `cos`? For variables it potentially matters if we substitute them before or after taking the derivative, but are there situations where that matters for functions?

oscarbenjamin: > are there situations where that matters for functions?

I'm not sure if this is really well defined at all but maybe something weird like this:
```python
In [6]: f(f(x)).diff(f(x)).subs(f, cos)
Out[6]: 
⎛ d          ⎞│         
⎜───(cos(ξ₀))⎟│         
⎝dξ₀         ⎠│ξ₀=cos(x)
```

ThePauliPrinciple: I think
```python
Subs(Derivative(f(f(x)), f(x)), f, cos).doit()
Derivative(f(f(x)), f(x)).subs(f, cos).doit()
f(f(x)).diff(f(x)).subs(f, cos).doit()
```
Are all the same

ThePauliPrinciple: For variables, it matters a lot:
```python
>>> Derivative(sin(x),x).subs(x,0).doit()
1
>>> Derivative(sin(x).subs(x,0),x).doit()
0
```
whether we sub first or take the derivative first.

oscarbenjamin: Maybe something like this:
```python
In [6]: cos(x).diff(x, 2).subs(cos, exp)
Out[6]: 
  x
-ℯ 

In [7]: cos(x).subs(cos, exp).diff(x, 2)
Out[7]: 
 x
ℯ 
```
I'm not really sure what kind of example we're looking for here.

ThePauliPrinciple: Interesting, it seems that we can't diff w.r.t. `cos(x)` but we can do it w.r.t. `f(x)` (due to how `_diff_wrt` is set).

ThePauliPrinciple: Ah I see, yes, if you are substituting two different defined functions then the order is going to matter. That's unfortunate.

asmeurer: Is that something that's tested? There's not a lot of clarity on the difference between `subs` and `replace`, but I would generally expect something like that to only be supported by `replace`. 

It would be possible to that symbolically using something more explicit like `Subs(exp(x), exp(a), cos(a))`, where `a` is a `Wild`. But I rather expect this is something that is only supported accidentally.

ThePauliPrinciple: I feel that I've not really stated my concerns clearly on this topic:

`Subs` is a SymPy object which is created when `expr.subs` is used if expr contains a `Derivative` object. This is because someone might want to represent symbolically the derivative `f'(x)` at `x=0`. Here the substitution then also needs to be represented symbolically because `f '(0)` would be equal to zero since it is constant.

`.subs` has been specifically coded to work with substitutions of functions (which are classes) as in `f(x).subs(f, cos) == cos(x)`.

When subs is used for function substitutions and there is a `Derivative` in that expression, a `Subs` object is created that now has as its arguments `Function`s, which are subclasses of `Basic` and not instances of `Basic`. 

I was hoping the creation of such a `Subs` object could be avoided altogether by simply performing the actual substitution rather than creating the symbolic `Subs` object because I thought there would never be an ambiguity (unlike in the case of substituting a variable with a constant), but it seems that this is not the case.

A unit test exists for a `Subs` object with `Function`s as its arguments: sympy/core/tests/test_function.py, line 736

oscarbenjamin: Ultimately the problem is that functions aren't instances of Basic. In the mean time maybe it's possible for `Subs` to store a function using a dummy so it actually stores the called function like `f(_x)`.

ThePauliPrinciple: I considered something like this too, but it would leave some problems:
- how do we distinguish between a substitution with a function and substitution with a function as a function of a dummy? Do we store the dummy in the `Subs` object too if we create it for this purpose?
- The printing methods would have to be rewritten to take this into account.

oscarbenjamin: Yes, it would leave some problems. If a function was a proper symbolic object then I doin't think that Subs would be needed at all. We use it for something like this:
```python
In [1]: f(x).diff(x).subs(x, 0)
Out[1]: 
⎛d       ⎞│   
⎜──(f(x))⎟│   
⎝dx      ⎠│x=0
```
If Function was a proper symbolic object though then we could represent that as just `Derivative(f)(0)` which would in turn print as `f'(0)`. Then there's no need for Subs and no need to specify any symbols at all.

oscarbenjamin: We can't really "allow" Basic subclasses because they generally won't work:
```python
>>> from sympy import *
>>> f = Function('f')
>>> x = Symbol('x')
>>> e = Subs(f(x), f, cos)
```
If you check `dir(Expr)` it's clear that a large number of methods don't work and it's not reasonable to make them work e.g.:
```python
>>> e.apart()
  File "/home/oscar/current/sympy/sympy.git/sympy/core/expr.py", line 3696, in apart
    return apart(self, x, **args)
  File "/home/oscar/current/sympy/sympy.git/sympy/utilities/decorator.py", line 35, in threaded_func
    return func(expr, *args, **kwargs)
  File "/home/oscar/current/sympy/sympy.git/sympy/polys/partfrac.py", line 81, in apart
    (P, Q), opt = parallel_poly_from_expr((P, Q), x, **options)
  File "/home/oscar/current/sympy/sympy.git/sympy/polys/polytools.py", line 4403, in parallel_poly_from_expr
    return _parallel_poly_from_expr(exprs, opt)
  File "/home/oscar/current/sympy/sympy.git/sympy/polys/polytools.py", line 4440, in _parallel_poly_from_expr
    expr = expr.expand()
  File "/home/oscar/current/sympy/sympy.git/sympy/core/cache.py", line 70, in wrapper
  File "/home/oscar/current/sympy/sympy.git/sympy/core/expr.py", line 3627, in expand
    expr, hit = Expr._expand_hint(expr, hint, deep=deep, **hints)
  File "/home/oscar/current/sympy/sympy.git/sympy/core/expr.py", line 3560, in _expand_hint
    arg, arghit = Expr._expand_hint(arg, hint, **hints)
  File "/home/oscar/current/sympy/sympy.git/sympy/core/expr.py", line 3560, in _expand_hint
    arg, arghit = Expr._expand_hint(arg, hint, **hints)
  File "/home/oscar/current/sympy/sympy.git/sympy/core/expr.py", line 3559, in _expand_hint
    for arg in expr.args:
TypeError: 'property' object is not iterable
```
That's because `f.args` gives a property rather than a tuple which breaks a fundamental invariant that is supposed to be expected from Basic.

PR #22446, opened 2021-11-09
add type hints to evalf.py

eagleoflqj:

As a continuation of #22211
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -3,7 +3,8 @@ Adaptive numerical evaluation of SymPy expressions, using mpmath
 for mathematical functions.
 """
 
-from typing import Tuple as tTuple
+from typing import Tuple as tTuple, Optional, Union as tUnion, Callable, List, Dict as tDict, Type, TYPE_CHECKING, \
+    Any, overload
 
 import math
 
@@ -28,6 +29,22 @@ from sympy.external.gmpy import SYMPY_INTS
 from sympy.utilities.iterables import is_sequence
 from sympy.utilities.lambdify import lambdify
 from sympy.utilities.misc import as_int
+
+if TYPE_CHECKING:
+    from sympy.core.expr import Expr
+    from sympy.core.add import Add
+    from sympy.core.mul import Mul
+    from sympy.core.power import Pow
+    from sympy.core.symbol import Symbol
+    from sympy.integrals.integrals import Integral
+    from sympy.concrete.summations import Sum
+    from sympy.concrete.products import Product
+    from sympy.functions.elementary.exponential import exp, log
+    from sympy.functions.elementary.complexes import Abs, re, im
+    from sympy.functions.elementary.integers import ceiling, floor
+    from sympy.functions.elementary.trigonometric import atan
+    from sympy.functions.combinatorial.numbers import bernoulli
+    from .numbers import Float, Rational, Integer
 
 LG10 = math.log(10, 2)
 rnd = round_nearest
@@ -62,7 +79,10 @@ An mpf value tuple is a tuple of integers (sign, man, exp, bc)
 representing a floating-point number: [1, -1][sign]*man*2**exp where
 sign is 0 or 1 and bc should correspond to the number of bits used to
 represent the mantissa (man) in binary notation, e.g.
+"""
+MPF_TUP = tTuple[int, int, int, int]  # mpf value tuple
 
+"""
 Explanation
 ===========
 
@@ -81,9 +101,19 @@ relative accuracy of the respective complex part, but may be anything
 if the corresponding complex part is None.
 
 """
+TMP_RES = Any  # temporary result, should be some variant of
+# tUnion[tTuple[Optional[MPF_TUP], Optional[MPF_TUP],
+#               Optional[int], Optional[int]],
+#        'ComplexInfinity']
+# but mypy reports error because it doesn't know as we know
+# 1. re and re_acc are either both None or both MPF_TUP
+# 2. sometimes the result can't be zoo
+
+# type of the "options" parameter in internal evalf functions
+OPT_DICT = tDict[str, Any]
 
 
-def fastlog(x):
+def fastlog(x: Optional[MPF_TUP]) -> tUnion[int, Any]:
     """Fast approximation of log2(x) for an mpf value tuple x.
 
     Explanation
@@ -119,7 +149,7 @@ def fastlog(x):
     return x[2] + x[3]
 
 
-def pure_complex(v, or_real=False):
+def pure_complex(v: 'Expr', or_real=False) -> Optional[tTuple['Expr', 'Expr']]:
     """Return a and b if v matches a + I*b where b is not zero and
     a and b are Numbers, else None. If `or_real` is True then 0 will
     be returned for `b` if `v` is a real number.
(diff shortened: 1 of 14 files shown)


sympy/core/evalf.py
```diff
@@ -83,7 +101,13 @@ class PrecisionExhausted(ArithmeticError):
 """
 
 
-def fastlog(x):
+MPF_VALUE_TUPLE = tTuple[int, int, int, int]
+SCALED_ZERO_TUPLE = tTuple[List[int], int, int, int]
+TEMP_RESULT = Any
+# tUnion[tTuple[Optional[MPF_VALUE_TUPLE], Optional[MPF_VALUE_TUPLE], Optional[int], Optional[int]], 'ComplexInfinity']
```
oscarbenjamin, line 107: There should be comments explaining what these types are and what they mean in practise.

oscarbenjamin: They could also have shorter names

sympy/integrals/integrals.py
```diff
@@ -41,6 +43,8 @@ class Integral(AddWithLimits):
 
     __slots__ = ('is_commutative',)
 
+    args: tTuple[Expr, Any]
```
oscarbenjamin, line 46: Why is Any here?

sympy/core/evalf.py
```diff
@@ -944,7 +997,7 @@ def evalf_log(expr, prec, options):
         return re, None, re_acc, None
 
 
-def evalf_atan(v, prec, options):
+def evalf_atan(v: 'atan', prec: int, options: tDict[str, Any]) -> TEMP_RESULT:
```
oscarbenjamin, line 1000: Maybe a type should be defined at the top for options so it doesn't get repeated everywhere. I think we can be more precise than Any here e.g.:
```
tDict[str, Optional[tUnion[bool, int, Dict[Basic, Sympifiable]]]
````
(I realise we don't yet have sympifiable)

Comment (eagleoflqj) [reply sympy/core/evalf.py:1000]: This produces many errors as different keys are associated with different types. TypedDict is needed, which is new in python 3.8

File: sympy/integrals/integrals.py (hunk 2 above)
Comment (eagleoflqj) [reply sympy/integrals/integrals.py:46]: https://github.com/sympy/sympy/blob/6a5cb5dabd6d090cc1d45f7ef591462cb0026fc6/sympy/core/evalf.py#L1018-L1020
This expects the second arg to be a trinary `sympy.core.containers.Tuple`, which is guaranteed **dynamically** at
https://github.com/sympy/sympy/blob/6a5cb5dabd6d090cc1d45f7ef591462cb0026fc6/sympy/core/evalf.py#L1114-L1117
but no type hints can be **statically** applied.

File: sympy/core/evalf.py (hunk 1 above)
Comment (eagleoflqj) [reply sympy/core/evalf.py]: I move the definitions to the nearest comments that explain them. Shouldn't cause any confusion this time.
https://github.com/eagleoflqj/sympy/blob/afb1af8cd5b2fc876d795f6aaabc22d21c8a81b7/sympy/core/evalf.py#L78-L83

File: sympy/integrals/integrals.py (hunk 2 above)
Comment (oscarbenjamin) [reply sympy/integrals/integrals.py:46]: Can it not be `tTuple[Expr, Tuple[Expr, ...]]`?

I guess mypy doesn't understand that Tuple is like tuple so maybe `tTuple[Expr, Tuple]` with the implicit understanding that the args of Tuple are Basic.

File: sympy/core/evalf.py [hunk 4]
```diff
@@ -81,9 +101,16 @@ class PrecisionExhausted(ArithmeticError):
 if the corresponding complex part is None.
 
 """
+TMP_RES = Any  # temporary result, should be some variant of
+# tUnion[tTuple[Optional[MPF_TUP], Optional[MPF_TUP], Optional[int], Optional[int]], 'ComplexInfinity']
```
oscarbenjamin, line 105: This line is too long

sympy/core/evalf.py
```diff
@@ -81,9 +101,16 @@ class PrecisionExhausted(ArithmeticError):
...
+# but mypy reports error because it doesn't know as we know
+# 1. re and re_acc are either both None or both MPF_TUP
+# 2. sometimes the result can't be zoo
+
+OPT_DICT = tDict[str, Any]  # type of the "options" parameter in internal evalf functions
 
 
-def fastlog(x):
+def fastlog(x: Optional[MPF_TUP]) -> tUnion[int, Any]:
```
oscarbenjamin, line 113: Is the return here not `tUnion[float, int]`?

sympy/core/evalf.py
```diff
@@ -140,16 +167,25 @@ def pure_complex(v, or_real=False):
...
+    if t:
+        c, i = t.as_coeff_Mul()
+        if i is S.ImaginaryUnit:
+            return h, c
+    elif or_real:
+        return h, t
+    return None
+
+
+SCALED_ZERO_TUP = tTuple[List[int], int, int, int]  # I don't know what this is, see function scaled_zero below
```
oscarbenjamin, line 179: Another too long line

sympy/core/evalf.py
```diff
@@ -187,13 +223,13 @@ def scaled_zero(mag, sign=1):
         raise ValueError('scaled zero expects int or scaled_zero tuple.')
 
 
-def iszero(mpf, scaled=False):
+def iszero(mpf: tUnion[MPF_TUP, SCALED_ZERO_TUP, None], scaled=False) -> Optional[bool]:
     if not scaled:
         return not mpf or not mpf[1] and not mpf[-1]
     return mpf and isinstance(mpf[0], list) and mpf[1] == mpf[-1] == 1
```
oscarbenjamin, line 229: Why is the return type `Optional` here? It looks like it always returns a bool to me.

sympy/core/evalf.py
```diff
@@ -187,13 +223,13 @@ def scaled_zero(mag, sign=1):
...
 
 
-def complex_accuracy(result):
+def complex_accuracy(result: TMP_RES) -> tUnion[int, Any]:
```
oscarbenjamin, line 232: Is the return type here not `tUnion[int, float]`?

sympy/integrals/integrals.py
eagleoflqj: `tTuple[Expr, Tuple]` is feasible. I didn't notice `Integral(x, x)` automatically converts the second arg to sympy tuple.
`Tuple[Expr, ...]` produces
```
sympy/integrals/integrals.py:46: error: Unexpected "..."
sympy/integrals/integrals.py:46: error: "Tuple" expects no type arguments, but 2 given
```

sympy/core/evalf.py
eagleoflqj: `iszero(None, True)` returns `None`. This may happen in https://github.com/sympy/sympy/blob/dfd966437a2ac0551c2c55174187f92dd6cba278/sympy/core/evalf.py#L572
Here `re` is returned by `add_terms`
https://github.com/sympy/sympy/blob/dfd966437a2ac0551c2c55174187f92dd6cba278/sympy/core/evalf.py#L549-L550
thus could be `None`

sympy/core/evalf.py
eagleoflqj: https://github.com/sympy/sympy/blob/dfd966437a2ac0551c2c55174187f92dd6cba278/sympy/core/evalf.py#L557-L566
float will cause a float value to be assigned to int `prec`

sympy/core/evalf.py
eagleoflqj: Declaring float return type will cause many operations to generate float value

oscarbenjamin: I don't understand what you mean. The type annotations do not affect the behaviour of the code so I guess you mean that in some cases the type is inferred to be float but I'm not sure how that's a problem given that the return type can be a float.

My reading of the function is that it either returns `MINUS_INF` which is a float or it returns `x[2]+x[3]` where `x` has type `MPF_TUP` so that's a sum of ints which gives an int.

sympy/core/evalf.py
oscarbenjamin: Okay, I see. I wonder if that's intentional. I would have written that more explicitly if I wanted to return None.

sympy/core/evalf.py
oscarbenjamin: So what does happen in this code path if `complex_accuracy` returns `INF`?

oscarbenjamin:
Maybe the difficulty with the hints here stems from the fact that so many different types are being mixed up and that should just be fixed in evalf by making it more type consistent.

sympy/core/evalf.py
eagleoflqj: https://github.com/sympy/sympy/blob/dfd966437a2ac0551c2c55174187f92dd6cba278/sympy/core/evalf.py#L275-L291
When line 283 is executed, we already know `re` is neither `fzero` nor `None`, so `size_re` should be int, but mypy doesn't know that. Here is an example that mypy reports error:
```py
def f(x: int) -> Union[int, None]:
    if x == 0:
        return None
    return x

a: int = f(3)
print(a, type(a))
```
Maybe `Literal` can be used to solve this problem, but that requires python 3.8
Or I can simply remove these kinds of return type hint if you prefer so.

oscarbenjamin: I don't actually know what I prefer. I'm basically learning how type hints work in practice as I go here (hence all my questions).

oscarbenjamin:
I think this looks good. This is the first attempt to add comprehensive type hints to a significant part of the codebase. I'll merge this now and then we can see what happens.

Thanks for this @eagleoflqj!

sympy/core/evalf.py
eagleoflqj: `target_prec - acc` is minus inf then, so the `max` returns `10 + 2**i` and `prec` remains int.

PR #22449, opened 2021-11-09
21890b: as_real/imag more sensitive to denom

smichr:
In #21890 the observation was made that the following didn't
raise an error even though the lhs is not real.

```python
>>> var('x', real=1)
>>> eq = (-2**(S(2)/3)/4 - 2**(S(2)/3)*sqrt(3)*I/4)/x >= 2**(S(2)/3)/(2*x)
```
Now `is_extended_real` and `is_imaginary` are aware of how the
denominator will affect an answer and the expression raises an error. 
The helper routine that each
of them uses works for an expression having a denominator of 1.
The individual routines now only call that helper with flat
expression and do the analysis based on the denominator in their
respective routines.

`is_nonzero` was changed to `is_zero == False` to allow for an infinite
base in Pow's routine for `is_finite`.

`(1/Dummy(zero=0, extended_real=1)).is_nonzero` is `None` because it is not known
if the value is infinite; but having an infinite base in a power with
a negative exponent does not change the fact that it is still finite.

#### Other comments
With these definitions,
```python
x = Symbol('x')
O = Symbol('O', zero=True)
i = Symbol('i', imaginary=True)  # is finite
xr = Symbol('xr', extended_real=True)
nz = Symbol('nz', extended_real=True, zero=False)
nr = Symbol('Z', complex=True, extended_real=False)  # finite
ni = Symbol('Z', complex=True, imaginary=False)  # finite
Z = Symbol('Z', complex=True, extended_real=False, imaginary=False)  # finite
```
these give False in master
```python
assert (i/xr).is_extended_real is None
assert (i/xr).is_real is None
assert (i/xr).is_zero is None
```
These all give None in master
```python
assert (O/nz).is_extended_real is True
assert (O/nz).is_imaginary is False
assert (O/nz).is_zero is True
assert (Z/nz).is_extended_real is False
assert (Z/nz).is_imaginary is False
assert (Z/nz).is_real is False
assert (Z/nz).is_zero is False
assert (Z/xr).is_imaginary is False
assert (i/Z).is_extended_real is False
assert (i/Z).is_real is False
assert (ni/nz).is_imaginary is False
assert (nr/nz).is_extended_real is False
assert (nr/nz).is_real is False
assert (nr/nz).is_zero is False
assert (nz/Z).is_imaginary is False
assert (nz/xr).is_imaginary is False
assert (xr/Z).is_imaginary is False
```

#### Release Notes

* core
    * the is_real/is_imaginary tests for expressions with a denominator other than 1 are more aware of the denominator
* simplify
    * `fraction`, `numer` and `denom`: still available from `radsimp` but now in `core.mul` and cannot be imported from `simplify`


sympy/core/assumptions.py
```diff
@@ -74,7 +74,7 @@
         object has the value of 0.
 
     nonzero
-        object is a real number that is not zero.
+        object is a extended real number that is not zero.
```
smichr, line 77:
```suggestion
        object is a real number that is not zero.
```

sympy/core/mul.py
```diff
@@ -1426,9 +1426,27 @@ def _eval_is_polar(self):
             all(arg.is_polar or arg.is_positive for arg in self.args)
 
     def _eval_is_extended_real(self):
+        from sympy.simplify.radsimp import fraction
+        n, d = fraction(self)
```
oscarbenjamin, line 1430: It is problematic to have anything that leads to evaluation during assumptions queries. Maybe this should use `exact=True`.

It's also problematic to have circular imports like this. Maybe the `fraction` function should be defined her in `mul.py` rather than `radsimp.py`.

oscarbenjamin, line 1430: It is problematic to have anything that leads to evaluation during assumptions queries. Maybe this should use `exact=True`.

It's also problematic to have circular imports like this. Maybe the `fraction` function should be defined her in `mul.py` rather than `radsimp.py`.

smichr: fraction, numer and denom have been moved. They can still be imported from radsimp but not from simplify (because they are now available from `*, core, mul`).

smichr: `fraction` now uses `exact`

sympy/core/mul.py
```diff
@@ -1328,6 +1336,17 @@ def _eval_is_algebraic(self):
...
     def _eval_is_zero(self):
+        n, d = fraction(self, exact=True)
+        if not d.is_Integer:
+            if n.is_zero:
+                if d.is_zero is False:
+                    return True
+            if d.is_finite is None:
+                return
+            if n.is_zero is False:
+                return False
```
oscarbenjamin, line 1347: This gives:
```python
In [5]: n = Symbol('n', extended_nonzero=True)

In [6]: d = Symbol('d', finite=False)

In [7]: (n/d).is_zero
Out[7]: False
```
We don't know though if this is `oo/oo` so it should give `None`.

sympy/core/mul.py
```diff
@@ -1328,6 +1336,17 @@ def _eval_is_algebraic(self):
...
+            if n.is_finite and d.is_infinite:
+                return True
```
smichr, line 1349:
```suggestion
            if n.is_finite:
                if d.is_infinite:
                    return True
                if d.is_infinite is not None and n.is_zero is False:
                    return False
```

sympy/core/tests/test_assumptions.py
```diff
@@ -1292,3 +1292,121 @@ def test_common_assumptions():
...
+    assert (nz/x).is_real is None
+    assert (nz/i).is_real is False
+    assert (nz/xr).is_real is None
+    assert (nz/Z).is_real is False
+    assert (Z/x).is_real is None
+    assert (Z/i).is_real is False
+    assert (Z/xr).is_real is None
+    assert (Z/nz).is_real is False
+    assert (nr/nz).is_real is False
+    assert (ni/nz).is_real is None
```
smichr, line 1412:
```suggestion
    assert (ni/nz).is_real is None

    n = Symbol('n', extended_nonzero=True)
    d = Symbol('d', finite=False)
    assert (n/d).is_zero is None  # could be oo/oo->nan->None for is_zero
Out[7]: False
```

sympy/core/tests/test_assumptions.py
```diff
@@ -1292,3 +1292,126 @@ def test_common_assumptions():
...
+
+    n = Symbol('n', extended_nonzero=True)
+    d = Symbol('d', finite=False)
+    assert (n/d).is_zero is None  # could be oo/oo->nan->None for is_zero
+Out[7]: False
```
smichr, line 1417:
```suggestion
```

sympy/core/mul.py
```diff
@@ -1328,6 +1336,18 @@ def _eval_is_algebraic(self):
...
+            if n.is_zero:
+                if d.is_zero is False:
+                    return True
+            if d.is_finite is None:
+                return
+            if n.is_finite:
+                if d.is_infinite:
+                    return True
+                if d.is_infinite is not None and n.is_zero is False:
+                    return False
```
smichr, line 1350:
```suggestion
        if not d.is_Integer:
            if n.is_zero:
                if d.is_zero is False:
                    return True
            elif n.is_finite:
                if d.is_infinite:
                    return True
                if d.is_finite and n.is_zero is False:
                    return False
            return
```

sympy/core/mul.py
```diff
@@ -1429,9 +1449,79 @@ def _eval_is_polar(self):
...
+        # Matrix([      /   \       -1
+        # [ 0,  1,  1,  1,  1, 1],  -1
+        # [-1,  1, -1, -1, -1, 1],   0
+        # [-1, -1,  1, -1, -1, 1],  -1
+        # [-1, -1, -1,  0, -1, 1],\\
+        # [-1, -1, -1, -1,  0, 1],// -1, -1, -1, 0, 1
+        # [ 1,  1, -1, -1, -1, 0]])
+        n, d = fraction(self, exact=True)
+        if not d.is_Integer:
+            nri = lambda x: x.is_complex and x.is_extended_real is False and x.is_imaginary is False
```
smichr, line 1469:
```suggestion
            # check for indeterminants
            if None in (n.is_infinite, d.is_infinite):
                return
            if None in (n.is_zero, d.is_zero):
                return
            nri = lambda x: x.is_complex and x.is_extended_real is False and x.is_imaginary is False
```

sympy/core/mul.py
```diff
@@ -1474,6 +1564,55 @@ def _eval_real_imag(self, real):
...
+        # Matrix([      /   \        |
+        # [ 0, -1, -1, -1, -1, -1],  |
+        # [-1, -1,  1, -1, -1, -1],  |
+        # [-1,  1, -1, -1, -1, -1],  \
+        # [-1, -1, -1,  0, -1, -1],}\_ becomes -1, -1, -1, 0, -1
+        # [-1, -1, -1, -1,  0, -1],}/
+        # [-1, -1, -1, -1, -1,  0]])
+        n, d = fraction(self, exact=True)
+        if not d.is_Integer:
+            nri = lambda x: x.is_complex and x.is_extended_real is False and x.is_imaginary is False
```
smichr, line 1584:
```suggestion
            # check for indeterminants
            if None in (n.is_infinite, d.is_infinite):
                return
            if None in (n.is_zero, d.is_zero):
                return
            nri = lambda x: x.is_complex and x.is_extended_real is False and x.is_imaginary is False
```

sympy/core/mul.py
```diff
@@ -1429,9 +1448,84 @@ def _eval_is_polar(self):
...
+        # ...     ix = ns.index(i), ns.index(j)
+        # ...     m[ix] = {None:0,True:1,False:-1}[(i/j).is_extended_real]
+        # ...
+        # >>> m           +--------> 1
+        # Matrix([      /   \       -1
+        # [ 0,  1,  1,  1,  1, 1],  -1
+        # [-1,  1, -1, -1, -1, 1],   0
+        # [-1, -1,  1, -1, -1, 1],  -1
+        # [-1, -1, -1,  0, -1, 1],\\
+        # [-1, -1, -1, -1,  0, 1],// -1, -1, -1, 0, 1
```
smichr, line 1464:
```suggestion
        # >>> m = eye(len(ns))
        # >>> for i in ns:
        # ...   for j in ns:
        # ...     ix = ns.index(i), ns.index(j)
        # ...     m[ix] = {None:0,True:1,False:-1}[(i/j).is_extended_real]
        # ...
        # >>> m           +--------> 1
        # Matrix([      /   \       -1
        # [ 0,  1,  1,  1,  1, 1],  -1
        # [-1,  1, -1, -1, -1, 1],   0
        # [-1, -1,  1, -1, -1, 1],  -1
        # [-1, -1, -1,  0, -1, 1],\\
        # [-1, -1, -1, -1,  0, 1],// -->  -1, -1, -1, 0, 1
```

PR #22458, opened 2021-11-10
vector: all classes are now argument invariant

### LINKED ISSUE #22457: Argument invariance of vector
Several classes in the vector package do not have the correct argument invariance (BaseVector, BaseDyadic and BaseScalar). It seems to me that this is more due to them being an unfortunate mix and use of parent classes, rather than some inherent problem with their parent classes or the classes themselves.

ThePauliPrinciple:

Fixes #22457
See also #22310

Several classes were not argument invariant. This commit "fixes"
the invariance, but does not solve the issues in the design of
these classes that causes the issue in the first place.

#### Release Notes

NO ENTRY


sympy/vector/dyadic.py
```diff
@@ -225,6 +225,13 @@ def _sympystr(self, printer):
         return "({}|{})".format(
             printer._print(self.args[0]), printer._print(self.args[1]))
 
+    #this should not be needed, but this class is indicated as an atom
+    #eventhough, it has `Basic` args, this makes the argument invariance
+    #behave as if it is an atom. It might be better to redesign this class
+    #since it uses conflicting designs.
+    @property
+    def func(self):
+        return lambda *args, **kwargs: self
```
oscarbenjamin, line 234: Why is this needed for `BaseDyadic`? The class seems to work okay to me:
```python
In [11]: from sympy.vector import CoordSys3D

In [12]: N = CoordSys3D('N')

In [13]: d = N.i | N.j

In [14]: type(d)
Out[14]: sympy.vector.dyadic.BaseDyadic

In [15]: d.func
Out[15]: sympy.vector.dyadic.BaseDyadic

In [16]: d.args
Out[16]: (i_N, j_N)

In [17]: d.func(*d.args)
Out[17]: (i_N|j_N)

In [18]: d.func(*d.args) == d
Out[18]: True
```

sympy/vector/scalar.py
```diff
@@ -69,3 +69,7 @@ def system(self):
 
     def _sympystr(self, printer):
         return self._name
+
+    def func(self, *args):
+        return self.__new__(self.__class__, *args, pretty_str =
+                self._pretty_form[1:], latex_str = self._latex_form[2:])
```
oscarbenjamin, line 75: Something very strange here (on master):
```python
In [33]: from sympy.vector import CoordSys3D

In [34]: N = CoordSys3D('N')

In [35]: s = N.x

In [36]: s
Out[36]: x_N

In [37]: s.func(*s.args)
Out[37]: x0

In [38]: s.func(*s.args) == s
Out[38]: True
```
Is it nos possible to change it so that the optional arguments for `BaseScalar` are passed to `super().__new__` and appear in args?

oscarbenjamin: Same question applies to `BaseVector`.

sympy/vector/dyadic.py
ThePauliPrinciple: `isinstance(d, Atom)` is `True` since it is an `AtomicExpr`. This means it will no longer work when `Atom` is updated to be argument invariant (if done so according to #22310). One can ask the question why this is an `AtomicExpr` in the first place, since it doesn't behave atomic at all)

oscarbenjamin: Maybe just remove the `AtomicExpr` subclass. The other superclass already subclasses `Expr` whch looks like the correct superclass to me.

sympy/vector/scalar.py
ThePauliPrinciple: Currently the names are stored as `str`, which is why they are not in `args`. Storing them in a `String` wrapper is possible, but it would change the api (although only in `_` attributes). Note that this is currently an `AtomicExpr`, so its behaviour might be unexpected when there is an update for `func` of `Atom`. Redesigning these classes to simply not be `Atom` subclasses might be the best approach, but I felt that might be a bit invasive.

oscarbenjamin: Can we not just store `String` in `_args` (i.e. by passing to `super().__new__`) even if it is redundant. It's usually good practice that all public attributes for a non-Atom Basic class should be properties that access `args`. In some cases for performance reasons it can be better to attach some attributes directly on the instance but we still need to make sure that the func-args invariant is satisfied and that all of these attributes are correctly initialised by `obj.func(*obj.args)`.

ThePauliPrinciple: Well putting them in `_args` means, if I understand correctly, that it is possible to use `subs` to substitute these values, the question is whether that was intended. I'll be honest, this confuses me about these objects. Are they intended to behave like pure `Atom`s, that non of its arguments can be substituted, should it be a mix, that the arguments storted in _args can be substituted and those that are not in there can not, or was it intended that all arguments can be substituted, but the original author did not know how.

oscarbenjamin: I don't think we will get an answer to the question of what the original author intended so we should try now to decide what makes sense. @moorepants might be able to help with understanding this design.

ThePauliPrinciple: I think what makes most sense is 
> Is it not possible to change it so that the optional arguments for BaseScalar are passed to super().__new__ and appear in args?
Which is what I tried to accomplish with
```python
    def func(self, *args):
        return self.__new__(self.__class__, *args, pretty_str =
                self._pretty_form[1:], latex_str = self._latex_form[2:])
```
The slices are there because the `str` inputs get modified before they are stored.
This way, the non `args` arguments behave "atomic".

oscarbenjamin: Okay but can we make that work by defining `args` and `__new__` rather than `func`?

ThePauliPrinciple: I can make it work now, but these are subclasses of `Atom`, so they are expected to have no `args` in the first place and would fail if `Atom`s `func` returns `lambda: self`.

oscarbenjamin: But why are they subclasses of Atom? Maybe they shouldn't be. There is clearly something wrong here so we can't start from the "if it ain't broke don't fix it" perspective.

ThePauliPrinciple: The problems does remain though, if it works with `args` and `__new__` it always means that those arguments can be substituted as far as I understand. If you want any argument to behave "atomic", you will have to have some kind of "manipulation" in `func`

oscarbenjamin: It's not clear to me what the original authors intended. Few people understand how Basic works so few people understand how to subclass it correctly. What matters is not necessarily what those original authors intended but rather what is actually reasonable to us now and also whether any change is compatible with previous versions of SymPy.

ThePauliPrinciple: If it is required to be compatible with previous versions, then `pretty_str` and `latex_str` can not be stored in `_args` since that would make them substitutable. Then that only leaves the option of using a modified `func` so that these arguments are given to a newly created instance.

Currently, it seems to me that the "supported" way to store non-basic arguments (that then have the property that they can't be substituted or in general be modified with `obj.func( modification(obj.args) )`, but are still maintained under such operations) is to create a metaclass which stores those non-basic arguments as class attributes (e.g. I think `Function` does this with the name argument). This is needed, because you don't want the function to be replaced when you perform a sub with `obj.subs(Str('gamma'),Str('beta))` but only with `obj.subs(Function('gamma'),Function('beta'))`.

I think the lack of a generic way to introduce non `args` into `Basic`s and the (perceived) need to do so is what causes these kind of inconsistent codes. Something simple of demanding all `args` arguments to be first in `__new__` would work nicely, since `func` could then be written as:
```python
def __new__(self, a, b, c, d):
    obj = Basic.__new__(a, b)
    obj._non_args = (c, d)
def func(self, *args):
    return self.__class__(*args, *self._non_args)
```
This would work for `Atom`, `Basic` and "mixed" (with both arguments that behave "basic" and that behave "atomic"), although for performance reasons one might want to return self for `Atom` and leave out the `_non_args` in the case of `Basic`. (I do not suggest to add this in this PR or at all, it is just to indicate what I expected `Basic` to be able to do, and what I think also others expect it to do, or "feel" the need for `Basic` to do)

The proposed solution does exactly what is described above, except it isn't generic but specific for this class ( `*self._non_args` replaced with the actual arguments). 

I could work with any way to indicate an element of `args` is "atomic" (can't be replaced/modified). I could also make it work to simply treat these classes as `Atom` and don't store any of the arguments in `_args`. Which might actually be a good property for these classes.

oscarbenjamin: > If it is required to be compatible with previous versions, then `pretty_str` and `latex_str` can not be stored in `_args` since that would make them substitutable

It's not clear to me that substitutability matters in this situation or more importantly whether the current behaviour would be preferred over a more consistent alternative.

ThePauliPrinciple: Because when the intention is to use subs on a `Str` elsewhere in an `expr` also containing a `sympy.vector.scalar.BaseScalar` the `pretty_str` value might be replaced unintentionally simply because it coincidentally has the same `str` value, eventhough it represents a completely different kind of object.

oscarbenjamin: I suspect that that is very unlikely in practice so that it isn't really a compatibility problem. But that's why I brought @bjodah in because he might have more of an idea how this is used.

ThePauliPrinciple: I made an attempt at putting the strings in `args`. This leads to a bunch of test fails, which I think are due to the hash changing which changes the order in e.g. `Add._sorted_args`. Is it safe to assume that this change in behaviour will not cause issues? Or does this preclude the option of adding the strings to `args`?

ThePauliPrinciple:
The current fails are due to a change in the hash of `BaseScalar`, which causes `Add` to print its terms in a different order, failing the test comparisons with static strings.

oscarbenjamin:
It fails `flake8` checks.

ThePauliPrinciple:
> 
> 
> It fails `flake8` checks.

Previously, it converted `Symbol` to `str`, but this is no longer needed since `str` is now converted to `Str` and both `Str` and `Symbol` store their `str` in the `name` attribute. The `Symbol` import was thus no longer used.

ThePauliPrinciple:
It actually seems that `Symbol` doesn't compare with `Str`, so everything should be converted to `Str` afterall

ThePauliPrinciple:
In particular, a test fails because
```python
    A = CoordSys3D('A')
    print(A.x.args)
    print(BaseScalar(0, A, 'A_x', r'\mathbf{{x}_{A}}').args)
```
```
(0, A, x_A, \mathbf{{x}_{A}})
(0, A, A_x, \mathbf{{x}_{A}})
```
But I don't understand what I did that changed this behaviour

ThePauliPrinciple:
Alright, so it appears that the only reason this test passed in the first place was that `pretty_str` was never used for comparison since it was not in `args` before, but it was always different for the lhs and rhs in the respective assert.

ThePauliPrinciple:
It is not clear to me whether the current behaviour should be maintained (instances with different `pretty_str` and `latex_str` arguments still leading to an equality).

oscarbenjamin:
> It is not clear to me whether the current behaviour should be maintained (instances with different `pretty_str` and `latex_str` arguments still leading to an equality).

Having instances that behave different but compare equal with `__eq__` is problematic because of how it interacts with sets, dicts, the cache etc so ideally they wouldn't compare equal. I bet this causes bugs but it's also not possible to change in a compatible way.

There are various things wrong with the classes in sympy.vector. My inclination in general is that they need a complete rewrite.

ThePauliPrinciple:
Alright, I will consider the failure of __eq__ here as a bug and update the other classes along the same lines together with the concomitant tests.

I feel that only supporting argument invariance (which is currently bugged and fails silently) is something which benefits the entire library and would like to limit the scope in this PR to that.

oscarbenjamin:
> would like to limit the scope in this PR to that.

That's fine. We should make sure to clearly note in the code where it is not using best practice though so it is clear that the design should not be replicated elsewhere.

Does that mean that you need to change the PR?

ThePauliPrinciple:
Yes, `BaseVector`, `BaseDyadic` and the tests still need to be updated along the same lines as `BaseScalar`

ThePauliPrinciple:
Rewriting the tests turned out to be very involved. Due to the `args` having changed, many tests failed to a change in printing order (I suspect `_sorted_args` uses a hash based on `args`)

oscarbenjamin:
> I suspect `_sorted_args` uses a hash based on `args`

Is this the code that handles this:
https://github.com/sympy/sympy/blob/a8e583c4e5eae5c4263a5f2d7b088bf48f0282fe/sympy/printing/pretty/pretty.py#L1086

ThePauliPrinciple:
I don't think so, I think the problem is that e.g. `Add` and `Mul` sort their arguments when printing them 
afbeelding
So now that the the objects have additional arguments, they seem to sort to a different order, hence changing the printing output

oscarbenjamin:
But where does the sorting take place?

Are the args being sorted in the Basic instances or is it just that the printer sorts the args?

ThePauliPrinciple:
I'm not sure if there is a consistent pattern, but e.g. `Mul`, `Add` and `Cross` change the order of their arguments when the `Mul`, `Add` or `Cross` object is instanced

ThePauliPrinciple:
This ordering is now somehow changed due to the changes and hence the printing changes

ThePauliPrinciple:
Maybe this shows clearly what the problem is:
afbeelding

ThePauliPrinciple:
With some further testing, it seems like this can already happen when only changing the baseclass from `AtomicExpr` to `Expr`

ThePauliPrinciple:
```python
    @classmethod
    def class_key(cls):
        return 2, 0, cls.__name__
    @cacheit
    def sort_key(self, order = None):
        return Atom.sort_key(self, order)
```
It seems that adding this to the classes preserves the ordering (at least in some cases), would that be prefered over changing the printing behaviour?

ThePauliPrinciple:
It seems that the above then fails some other checks, I don't think I can solve this issue

oscarbenjamin:
This looks weird to me:
```python
In [21]: from sympy.vector import CoordSys3D, Cross

In [22]: A = CoordSys3D('A')

In [23]: Cross(A.i, A.x)
Out[23]: -(x_A)×(i_A)
```
Isn't `A.x` a scalar? How can it be an argument to `Cross`? Have I misunderstood something basic here?

ThePauliPrinciple:
I would agree, but the tests are filled with such statements.

oscarbenjamin:
@moorepants are we misunderstanding this? What does `Cross(A.i, A.x)` mean?

PR #22460, opened 2021-11-10
utilities/matchpy_connector: Wildcard.__eq__ now works without matchpy

ThePauliPrinciple:

See also https://github.com/sympy/sympy/issues/22310

The `Wildcard` class has an interface for when matchpy is not available,
but a comparsion with `==` results in an error because `__init__` does
not correctly initialize the class.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/integrals/rubi/symbol.py
+++ b/sympy/integrals/rubi/symbol.py
@@ -2,14 +2,8 @@ from sympy.external import import_module
 matchpy = import_module("matchpy")
 from sympy.utilities.decorator import doctest_depends_on
 
-if matchpy:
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
There should be a test of the change in behaviour.

ThePauliPrinciple:
It currently fails on the argument invariance test, but it is skipped (`if not obj.args`) since there are no arguments (removing the skip is how it popped up for me). Does it still require an explicit test in this case?

oscarbenjamin:
> Does it still require an explicit test in this case?

No, this is good.

Thanks!

PR #22463, opened 2021-11-11
Implemented function return_piecewise for evaluating limit when z0 is passed as Symbol

### LINKED ISSUE #21029: Limits involving division by x
Take the function `(sinh(x)+cosh(x)-1)/4`.  This function when divided by `x`, has a limit of 1/4 at 0:

```
In [77]: cc=(sinh(x)+cosh(x)-1)/4                                                                                                                  

In [78]: limit(cc/x,x,0)                                                                                                                           
Out[78]: 1/4

```

However, when the limit is taken as "`x` approaches a symbol" such as `z`, the result of `limit` doesn't take account the fact that `z` could be 0:

```
In [79]: limit(cc/x,x,z)                                                                                                                           
Out[79]: 
 z    
ℯ  - 1
──────
 4⋅z  

In [80]: limit(cc/x,x,z).subs(z,0)                                                                                                                 
Out[80]: nan
```

In this case the correct solution to the limit would be a piecewise function:

- 1/4 if `z` = 0.
- `(exp(z) - 1)/(4*z)` otherwise.

anutosh491:

Fixes #21029

Currently on Master

```
from sympy import *
x = Symbol('x')
z = Symbol('z')
>>> limit((exp(x)-1)/x, x, z) # Not the correct output 
(exp(z) - 1)/z
>>> limit((exp(x)-1)/x, x, z).subs(z, 0)
nan
>>> limit(ln(1+x)/x, x, z)
log(z + 1)/z
>>> limit(ln(1+x)/x, x, z).subs(z, 0) 
nan
>>> limit(sin(x)/x, x, z)
sin(z)/z
>>> limit(sin(x)/x, x, z).subs(z, 0)
nan
```
Hence whenever the point is passed as Symbol , we get wrong answers for the same . This is because the returned answers need a piecewise output and something like `sin(z)/z` is not the correct output for  `limit(sin(x)/x, x, z)`. This had to be addressed through a piecewise implementation as discussed in the issue .
The issue reported is basically the same thing 
```
In [79]: limit(cc/x,x,z)                                                                                                                           
Out[79]: 
 z    
ℯ  - 1
──────
 4⋅z  

In [80]: limit(cc/x,x,z).subs(z,0)                                                                                                                 
Out[80]: nan
```

So as I couldn't find any explicit check for `z0.is_Symbol` in `limits.py`, I realized something like this hasn't been implemented yet. Hence I've added a function `return_peicewise` and I've added a check at whichever places we could get an expression like above as output ! The implemented function would return a Piecewise Limit Function with a limit check at all concerned points starting from `-oo` to `oo`
Some examples being

```
>>> limit(sin(x)/x, x, z)
Piecewise((1, Eq(z, 0)), (sin(z)/z, True))
>>> limit(sin(x)/x, x, z).subs(z, 0)
1

>>> limit( sin(x)/(x**2 -4) , x, z)
Piecewise((oo, Eq(z, -2) | Eq(z, 2)), (sin(z)/((z - 2)*(z + 2)), True))

>>> limit((1- cos(x))/x**2, x, z)
Piecewise((1/2, Eq(z, 0)), (-(cos(z) - 1)/z**2, True))

>>> limit(1/(5 - x)**3, x, z, dir="+")
Piecewise((-oo, Eq(z, 5)), ((5 - z)**(-3), True))

>>> limit(1/(5 - x)**3, x, z, dir="-")
Piecewise((oo, Eq(z, 5)), ((5 - z)**(-3), True))

>>> limit(log(1 + x)/x, x, z)
Piecewise((0, Eq(z, -oo)), (oo, Eq(z, -1)), (1, Eq(z, 0)), (0, Eq(z, oo)), (log(z + 1)/z, True))

>>> limit((1 + x)/x, x, z)
Piecewise((1, Eq(z, -oo)), (oo, Eq(z, 0)), (1, Eq(z, oo)), ((z + 1)/z, True))

>>> limit((1 + 1/x)**x, x, z)
Piecewise((E, Eq(z, -oo)), (1, Eq(z, 0)), (E, Eq(z, oo)), ((1 + 1/z)**z, True))

>>> limit((1 + 1/(x**2 -4))**x, x, z)
Piecewise((1, Eq(z, -oo)), (0, Eq(z, -2)), (oo, Eq(z, 2)), (1, Eq(z, oo)), ((z**2/(z**2 - 4) - 3/(z**2 - 4))**z, True))

>>> limit((5**(1/x) + 3**(1/x))**x, x, z)
Piecewise((5, Eq(z, 0)), ((exp(log(3)/z) + exp(log(5)/z))**z, True))

>>> limit((sinh(x) + cosh(x) -1)/(4*x), x, z)    # this is the issue being solved above
Piecewise((1/4, Eq(z, 0)), (oo, Eq(z, oo)), ((exp(z) - 1)/(4*z), True))
>>> limit((sinh(x) + cosh(x) -1)/(4*x), x, z).subs(z, 0)
1/4

>>> limit(exp(1/x) , x, z)
Piecewise((oo, Eq(z, 0)), (exp(1/z), True))

>>> limit(1/sqrt(x), x, z, dir='-')
Piecewise((-oo*I, Eq(z, 0)), (1/sqrt(z), True))
```

I realized that explicit checks at `-oo` and `oo` are quite important because
1) They won't be coming under singularities usually 
2) Secondly this would lead to wrong answers . For eg if we don't have a check for them
```
>>> limit((1 + x)/x, x, z)
Piecewise((oo, Eq(z, 0)), ((z + 1)/z, True))
>>> limit((1 + x)/x, x, z).subs(z, oo)
nan
```

#### Other comments

The implementation currently works only for expressions have finite discontinuities but I feel this could be extended for expressions having periodic infinite discontinuities . Hence for something like an `imageset` for eg `limit(tan(x)/x, x, z)` or  `limit(tan(x), x, z)`  where we have a periodic repetition of discontinuous points , we could maybe try to calculate limit at the first point and generalize it overall !!
```
>>>singularities(tan(x)/x , x)
Union(FiniteSet(0), ImageSet(Lambda(_n, 2*_n*pi + pi/2), Integers), ImageSet(Lambda(_n, 2*_n*pi + 3*pi/2), Integers))
```
I wasn't sure how one could maybe implement this due to my inexperience with the ` Solvers` Module. Can we somehow get the generalized expression from the `ImageSet` ??  . Also I'll add tests as soon as the first set of review is done ! 

#### Release Notes

* series
  * Added Piecewise implementation for limit evaluation at symbolic points


sympy/series/limits.py
```diff
@@ -128,6 +128,47 @@ def heuristics(e, z, z0, dir):
...
+    Piecewise((1, Eq(z, 0)), (sin(z)/z, True))
+    >>> limit(log(x + 1)/x, x, z)
+    Piecewise((0, Eq(z, -oo)), (oo, Eq(z, -1)), (1, Eq(z, 0)), (0, Eq(z, oo)), (log(z + 1)/z, True))
+    >>> limit((1 + 1/(x**2 -4))**x, x, z)
+    Piecewise((1, Eq(z, -oo)), (0, Eq(z, -2)), (oo, Eq(z, 2)), (1, Eq(z, oo)), ((z**2/(z**2 - 4) - 3/(z**2 - 4))**z, True))
+    >>> limit(1/(5 - x)**3, x, z, dir="-")
+    Piecewise((oo, Eq(z, 5)), ((5 - z)**(-3), True))
+
+    """
+    from sympy import singularities, Eq, Piecewise, FiniteSet
```
oscargus, line 153: Please use more precis import locations.

sympy/series/limits.py
```diff
@@ -128,6 +128,47 @@ def heuristics(e, z, z0, dir):
     return rv
 
 
+def return_piecewise(expr, z0, cdir):
```
oscargus, line 131: I would call this `_return_piecewise` to indicate that it is for internal use.

sympy/series/limits.py
```diff
@@ -128,6 +128,47 @@ def heuristics(e, z, z0, dir):
...
+    """Helper function to evaluate limits when the point, ``z0``, is passed as a Symbol.
+    Returns the limit in piecewise form over the complete real number line from
+    Negative Infinity to Positive Infinity.
+
+    Note:This method is currently implemented for expressions having finite discontinuities.
+
+    Examples
+    ========
+
+    >>> from sympy import limit, sin, log
```
oscargus, line 141: It probably makes sense to actually import and use the new function here and/or add an example in `limit` showing that it can handle symbolic points.

sympy/series/limits.py
```diff
@@ -128,6 +128,47 @@ def heuristics(e, z, z0, dir):
     return rv
 
 
+def return_piecewise(expr, z0, cdir):
+    """Helper function to evaluate limits when the point, ``z0``, is passed as a Symbol.
+    Returns the limit in piecewise form over the complete real number line from
+    Negative Infinity to Positive Infinity.
+
+    Note:This method is currently implemented for expressions having finite discontinuities.
```
oscargus, line 136: Space after :

sympy/series/limits.py
```diff
@@ -283,6 +324,8 @@ def set_signs(expr):
                 if ex > 0:
                     return S.Zero
                 elif ex == 0:
+                    if z0.is_Symbol and not z0 in e.free_symbols:
+                        coeff = return_piecewise(coeff, z0, cdir)
```
oscargus, line 328: `return return_piecewise(...`

(which probably shows that the name is actually not that good, maybe `_limit_for_symbolic_point` or something similar?)

oscargus:
Nice idea! I left some comment on the code.

anutosh491:
> Nice idea! I left some comment on the code.

Thanks @oscargus for the extensive review. I had just pushed a general code for review , will go through all suggestions and make requested changes

sympy/series/limits.py
anutosh491: Yeah makes much more sense than what something like `return_piecewise` would convey . The name suggested justifies the work correctly !

sympy/series/limits.py
anutosh491: Yeah makes sense that we should be adding couple of exams in limit docstring too !

anutosh491:
ping @jksuom for review

anutosh491:
ping @0sidharth for review !

sympy/series/limits.py
```diff
@@ -128,6 +134,50 @@ def heuristics(e, z, z0, dir):
...
+            piecewise_list += [(limit(expr, z0, S.NegativeInfinity),Eq(z0, S.NegativeInfinity))]
+        for singularity in singularities(expr, z0):
+            piecewise_list.append((limit(expr, z0, singularity, dir),Eq(z0, singularity)))
+        if limit(expr, z0, S.Infinity) != expr.subs(z0, S.Infinity):
+            piecewise_list += [(limit(expr, z0, S.Infinity),Eq(z0, S.Infinity))]
+        piecewise_list += [(expr,True)]
+        expr = Piecewise(*piecewise_list)
+        return expr
+    else:
+        return expr
```
0sidharth, line 179:
```suggestion
    return expr
```

sympy/series/limits.py
```diff
@@ -360,4 +414,6 @@ def set_signs(expr):
             if r is None:
                 return self
 
+        if z0.is_Symbol and not z0 in e.free_symbols:
+            return _limit_for_symbolic_point(r, z0, cdir)
```
0sidharth, line 418: Can this just be called once sometime before gruntz, instead of having 3 different locations to call it?

sympy/series/limits.py
```diff
@@ -315,6 +367,8 @@ def set_signs(expr):
                 if ex.is_positive:
                     return S.Zero
                 elif ex == 0:
```
0sidharth, line 369: `is_meromorphic` and `z0 is S.Infinity` block have this else part quite similar, is it possible to move this code to a new heuristic function (`_limit_from_leading_term`) and use that as a substitute for both? Will decrease code repetition

sympy/series/limits.py
```diff
@@ -128,6 +134,50 @@ def heuristics(e, z, z0, dir):
...
+        piecewise_list = []
+        if cdir == 1:
+            dir = "+"
+        elif cdir == -1:
+            dir = "-"
+        if limit(expr, z0, S.NegativeInfinity) != expr.subs(z0, S.NegativeInfinity):
+            piecewise_list += [(limit(expr, z0, S.NegativeInfinity),Eq(z0, S.NegativeInfinity))]
+        for singularity in singularities(expr, z0):
+            piecewise_list.append((limit(expr, z0, singularity, dir),Eq(z0, singularity)))
+        if limit(expr, z0, S.Infinity) != expr.subs(z0, S.Infinity):
```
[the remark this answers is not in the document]
anutosh491: Yeah can be done,  I just thought of checking through all relevant points systematically ( `-oo` followed by singulairites followed by `oo`) , Would return outputs in a systematic order I guess through the whole number line!

sympy/series/limits.py
anutosh491: Yeah but only if somehow I could implement the suggestion below , otherwise would have to use it seperately for both the `is_meromorphic` and ` z0 is S.Infinity` blocks

sympy/series/limits.py
anutosh491: I'll try this out and get back !

anutosh491:
ping @jksuom for review

0sidharth:
A Release Notes Entry is needed here too. Please go through the guidelines about when they are needed, most of your PRs need one.

sympy/series/limits.py
0sidharth: Yes, I meant the entire `else` block, which includes `if ex.is_positive` and everything below

anutosh491: Oh okay I got what you were trying to convey . Nope no wrong answers really , atleast I couldn't find one while running the test locally and also mathematically shouldn't make much difference . I'll make a commit sometime soon for this . Thanks

anutosh491:
I've also tried to add a condition block for dealing with bi-directional limits. The basic procedure is to compare the left side sided symbolic limit and the right sided symbolic limit. If both are equal , we return the result (the left limit in my case) and if there are differences in between the limits we raise nan error pointing out  equality conditions at which the symbolic limits would not be equal 
Some examples where we raise `ValueError`
```
>>> limit(1/(5 - x)**3, x, z, dir="+-")
ValueError: Limit at following equalities do not exist : [Eq(z, 5)]

>>> limit(1/x, x, z, dir = '+-')
ValueError: Limit at following equalities do not exist : [Eq(z, 0)]

>>> limit((x**3 + 2*x + 1)/(x**2 -1), x, z, dir = '+-')
ValueError: Limit at following equalities do not exist : [Eq(z, -oo) | Eq(z, -1) | Eq(z, 1), Eq(z, oo)]

>>> limit((1 + x)/x, x, z, dir = '+-')
ValueError: Limit at following equalities do not exist : [Eq(z, 0)]
```

And others which work perfectly are 
```
>>> limit((1 - cos(x))/x**2, x, z, dir = '+-')
Piecewise((1/2, Eq(z, 0)), (-(cos(z) - 1)/z**2, True))

>>> limit( sin(x)/x, x, z, dir = '+-')
Piecewise((1, Eq(z, 0)), (sin(z)/z, True))
```

sympy/series/limits.py
anutosh491: I have implemented a basic layout for this ,You could surely review this now !

anutosh491:
The latest commit removes equalities at` oo` and `-oo` at which obviously directional limits are equal . The commit before this was comparing the `args` of the right handed limit and the left handed limit and whichever was different it was giving that as output but we may have expressions as type `or`  and some other types too which would eventually hold on to equality at `oo` and `-oo`
If you don't fully relate to what I say here's an example
```
>>> x = Symbol('x')
>>> z = Symbol('z')
>>> limit((x**3 + 2*x + 1)/(x**2 -1), x, z, dir = '+-')
# This basically evaluates the limit((x**3 + 2*x + 1)/(x**2 -1), x, z, dir = '+') and limit((x**3 + 2*x + 1)/(x**2 -1), x, z, dir = '-')


>>> limit((x**3 + 2*x + 1)/(x**2 -1), x, z, dir = '+')
Piecewise((-oo, Eq(z, -oo)), (oo, Eq(z, -1) | Eq(z, 1) | Eq(z, oo)), ((z**3 + 2*z + 1)/(z**2 - 1), True))

>>> limit((x**3 + 2*x + 1)/(x**2 -1), x, z, dir = '-')
Piecewise((-oo, Eq(z, -oo) | Eq(z, -1) | Eq(z, 1)), (oo, Eq(z, oo)), ((z**3 + 2*z + 1)/(z**2 - 1), True))
```
Now if we compares args obviously `(-oo, Eq(z, -oo))` and `(-oo, Eq(z, -oo) | Eq(z, -1) | Eq(z, 1)` are different and when evaluating such limits at singularities we may encounter expression of such type in this case of `Or` type . Similarly the second expression 
`(oo, Eq(z, -1) | Eq(z, 1) | Eq(z, oo))` and `(oo, Eq(z, oo))` are different and hence the answer for some cases was also including limits at `oo` and `-oo` even though they are same , hence the commit explicitly removes them from the equality conditions!

oscargus:
I cannot really say anything regarding the maths except that it seems to work. What happens if you get the symbolic limit of a `Piecewise`?

anutosh491:
Well something like `singularities(type(Piecewise) , sym, Domain)` hasn't been implemented anyway and doesn't return expected results so yeah that's something for the future and that's the reason this won't work but otherwise i'm certainly sure it would work.
Actually I had thought of implementing singularities for Piecewise some days ago . I'll give it a try in another Pr in a couple of days (don't think I should be disturbing this one and maintain the purpose of the Pr) .
Another work for the future is that I'll try to implement symbolic limits for expressions having periodic discontinuities . Being the first we implement limits at symbolic points , I would like to get the basic cases down in this Pr and build on top for other corner cases in following prs !

oscargus:
OK! I was just curious since this returns a Piecewise, so I wondered if it would be a Piecewise of Piecewises or a folded version. But I realize that there are other things limiting this than the current code.

anutosh491:
ping @0sidharth for review!

anutosh491:
hey @oscargus @smichr @jksuom I am satisfied with the changes from my side and and the Pr would almost certainly deals with any case having a finite singularities ! This is ready to go from my side . I have been finding it tough to get any reviews on my pr's on `series` and `concrete` lately but this one I am convinced that it would do it's job correctly !

anutosh491:
```
========================================================================= test process starts =========================================================================
executable:         C:\Users\anuto\anaconda3\envs\sympy-dev\python.exe  (3.9.6-final-0) [CPython]
architecture:       64-bit
cache:              yes
ground types:       python
numpy:              1.20.3
random seed:        18910121
hash randomization: on (PYTHONHASHSEED=1942652505)

sympy\series\tests\test_limits.py[144] ......................f................f........f...............................................f...............................
................                                                                                                                                                   [OK]

================================================== tests finished: 140 passed, 4 expected to fail, in 114.25 seconds ==================================================
```
Not sure why the test is failing , doesn't fail for me locally and I also reverted back to the previous commit thinking there was a mistake from my side but just one of the tests from the test function I've added fails but shouldn't

anutosh491:
ping @0sidharth for review ! I believe this is done from my side . I would be glad if this could be reviewed .

anutosh491:
ping @jksuom @0sidharth. I believe this is done from my side ! I would be glad if this can be reviewed !

anutosh491:
ping @jksuom @0sidharth. I believe this is done from my side and could be merged . I would be glad if this can be reviewed !

anutosh491:
ping @czgdp1807 for review!

anutosh491:
ping @jksuom , this looks like good work to me . Haven't heard from you on the pr before . Would be great if you could review !

anutosh491:
Cc: @0sidharth , I would like your views on this Pr, I have tried my best to improve this after the last set of reviews and think I might be good to go but a set of reviews would do great !

anutosh491:
ping @jksuom for review !

anutosh491:
@0sidharth @oscargus what are your thoughts on this Pr ? I feel this is good work and could go in after a set of reviews from your side . There have been some good reviews already by you'll on this but haven't heard on since then !

anutosh491:
Hey @oscarbenjamin 

This is one of my PRs from long back. I've seen that you've suggested all general results (maybe in piecewise form) being returned for series/limits. The idea here was somewhere along those lines. I would be glad if you can go through the PR [description ](https://github.com/sympy/sympy/pull/22463#issue-1050525512) and tell me if this might be worth a try.

oscarbenjamin:
I think it is reasonable that Piecewise might be needed if the limit point is a symbol or more generally some expression.

I'm not sure in general though whether a Piecewise is better than an unevaluated limit. There is always a tricky balance with this because using Piecewise seems nice in simple examples and when there are few cases but can explode if there are many difference cases that need to be considered.

It seems like the real work here is being done by the `singularities` function. I'm not sure how exactly that function works and when it can be expected to succeed.

PR #22464, opened 2021-11-11
Reorganize polys.numberfields

skieffer:
The set of problems solved in `polys.numberfields` can be organized in a better way, which the present module structure fails to reflect.

What we currently have in `polys.numberfields` is:
1. A way to compute minimal polynomials for algebraic numbers.
2. A solution to the Subfield Problem and its several allied problems (see [Cohen93], Section 4.5, or the docstring for the new `subfield.py` module).
3. An ad hoc function called `isolate()`, for returning an isolating interval for a real algebraic number.

Accordingly, we should have three modules:
1. minpoly.py
2. subfield.py
3. utilities.py

Therefore this PR makes the following moves:

* Rename `isomorphism.py` --> `subfield.py`
* Move `primitive_element()` function from `minpoly.py` --> `subfield.py`
* Move `to_number_field()` function from `numbers.py` --> `subfield.py`
* Rename `numbers.py` --> `utilities.py`
* Move unit tests in a corresponding way.

ALSO

This PR...

* Expands the docstrings for the public functions under `polys.numberfields`.
* Puts `polys.numberfields` into the docs.
* Improves test coverage in `polys.numberfields`.

NO ENTRY


skieffer:
I want to ask what you think about the definition of the `minpoly` function.
It used to be as follows (in `polys/numberfields/minpoly.py`):


```python
minpoly = minimal_polynomial
```

In other words, it was a pure synonym.

The problem with that is that when you include `minpoly` in the docs, you get an exact replica of the docstring for `minimal_polynomial`, which is really bad for users. ("Is this identical? Why did they repeat it?")


I have now changed the definition to:

```python
@public
def minpoly(ex, x=None, compose=True, polys=False, domain=None):
    """This is a synonym for :py:func:`~.minimal_polynomial`."""
    return minimal_polynomial(ex, x=x, compose=compose, polys=polys, domain=domain)

```

so that the docs look a lot better.

Now, I guess this incurs one extra stack frame if you call `minpoly` instead of `minimal_polynomial`. To me, that seems negligable, and worth it to make better docs. Wondering if you agree.

oscarbenjamin:
This all looks good to me.

We have a CI problem right now with the benchmarks suite that is unrelated to this PR.

oscarbenjamin:
> To me, that seems negligable, and worth it to make better docs. Wondering if you agree.

I do agree.

sympy/polys/numberfields/subfield.py
```diff
@@ -0,0 +1,486 @@
...
+from sympy.polys.numberfields.minpoly import _choose_factor, minimal_polynomial
+from sympy.polys.polyerrors import IsomorphismFailed
+from sympy.polys.polytools import Poly, PurePoly, factor_list
+from sympy.utilities import public
+
+from mpmath import pslq, mp
+
+
+def is_isomorphism_possible(a, b):
+    """Returns `True` if there is a chance for isomorphism. """
```
oscarbenjamin, line 53: Does this mean that the condition is necessary but not sufficient?

sympy/polys/numberfields/subfield.py
```diff
@@ -0,0 +1,486 @@
...
+
+    for i in range(1, 5):
+        A = a.root.evalf(n)
+        B = b.root.evalf(n)
+
+        basis = [1, B] + [ B**i for i in range(2, m) ] + [A]
+
+        dps, mp.dps = mp.dps, n
+        coeffs = pslq(basis, maxcoeff=int(1e10), maxsteps=1000)
+        mp.dps = dps
```
oscarbenjamin, line 101: Does this change the global `dps` in mpmath? If so there should be a better way. At least this could use `try/finally` or a context manager but also is `dps` in mpmath threadsafe?

Is there an mpmath function that can be called that does not require altering the global dps?

oscarbenjamin: Actually I see now that you are just moving this function from one place to another. Still this isn't good.

oscarbenjamin:
I went through this but I realise that it is mostly just moving stuff around so my comments above are not really comments on the changes made here in this PR.

This looks good to me.

Thanks!

sympy/polys/numberfields/subfield.py
skieffer: There are a few implementation details I wanted to revise in these modules as well, so maybe we can return to this and other points in another PR.

PR #22465, opened 2021-11-11
Removed distutil since it is deprecated in Python 3.10

oscargus:
`distutils` is deprecated in Python 3.10, so removed all use.

```
/home/runner/work/sympy/sympy/sympy/utilities/_compilation/compilation.py:8: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  from distutils.errors import CompileError
/home/runner/work/sympy/sympy/sympy/utilities/_compilation/compilation.py:9: DeprecationWarning: The distutils.sysconfig module is deprecated, use sysconfig instead
  from distutils.sysconfig import get_config_var, get_config_vars
```
#### Other comments

There is a template code in autowrap that imports distutils as a backup to setuptools. It may be a bit early to remove that part yet(?).
https://github.com/sympy/sympy/blob/c62bb34279944303f9d34748e9ff51f75e346f04/sympy/utilities/autowrap.py#L236-L255

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/utilities/_compilation/compilation.py
+++ b/sympy/utilities/_compilation/compilation.py
@@ -5,8 +5,7 @@ import subprocess
 import sys
 import tempfile
 import warnings
-from distutils.errors import CompileError
-from distutils.sysconfig import get_config_var, get_config_vars
+from sysconfig import get_config_var, get_config_vars, get_path
 
 from .runners import (
     CCompilerRunner,
@@ -16,10 +15,8 @@ from .runners import (
 from .util import (
     get_abspath, make_dirs, copy, Glob, ArbitraryDepthGlob,
     glob_at_depth, import_module_from_file, pyx_is_cplus,
-    sha256_of_string, sha256_of_file
+    sha256_of_string, sha256_of_file, CompileError
 )
-
-sharedext = get_config_var('EXT_SUFFIX')
 
 if os.name == 'posix':
     objext = '.o'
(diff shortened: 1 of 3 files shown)


asmeurer:
Looks good assuming all the replacements are available in versions of Python we support. Is all the code you changed covered in tests?  I agree that code that does `try: import setuptools; except: import distutils` can be left alone for now (including setup.py).

Why does the benchmarks build fail? That seems unrelated.

oscargus:
I think it is all supported, yes. Although I do not think that it is
actually tested (apart from the import part).

Checked a few of them and those were introduced in 3.2, but I am not sure I
checked everything.

Yes, the benchmark test is unrelated and quite weird as the filename is
changed without reason and it didn't happen before. No change of version
either.

Den tors 11 nov. 2021 22:24Aaron Meurer ***@***.***> skrev:

oscarbenjamin:
Looks good to me

oscargus:
The "new" methods were introduced in 3.2 and 3.3 so should be OK.

However, I could not find a replacement for `CompileError` so I defined my own type. Not sure if that should be noted somewhere?

bjodah:
Just a PSA if someone finds their way here when debugging compilation errors after upgrading SymPy:

Ubuntu 22.04 (or upstream Debian) shipped broken config for python 3.10.3:
```
Python 3.10.3 (main, Mar 16 2022, 17:19:40) [GCC 11.2.0] on linux
>>> import sysconfig
>>> sysconfig.get_path('include')
'/usr/local/include/python3.10'
```

but upgrading python to 3.10.6:
```console
$ apt update && apt install python3-dev
```
solves the issue:
```
Python 3.10.6 (main, Aug 10 2022, 11:40:04) [GCC 11.3.0] on linux
>>> import sysconfig
>>> sysconfig.get_path('include')
'/usr/include/python3.10'
```

PR #22467, opened 2021-11-11
Use MathJax version 3 in the docs

asmeurer:

MathJax 3 is now the default in Sphinx, so simply removing our manual use of MathJax 2 will use it. We also have to update the config as the config format has changed (see https://www.sphinx-doc.org/en/master/usage/extensions/math.html).

MathJax 3 is faster than MathJax 2 and also has several other features https://docs.mathjax.org/en/latest/upgrading/whats-new-3.0.html. For the most part, this shouldn't really affect anything. I noticed some formulas render slightly differently with MathJax 3 due to some minor font differences. I primarily need this for a future PR I plan to make that was not working properly without this change. 

#### Release Notes

* other
  * Use MathJax version 3 in the docs

### DIFF
--- a/doc/src/conf.py
+++ b/doc/src/conf.py
@@ -54,17 +54,12 @@ nitpick_ignore = [
 # To stop docstrings inheritance.
 autodoc_inherit_docstrings = False
 
-# MathJax file, which is free to use.  See https://www.mathjax.org/#gettingstarted
-# As explained in the link using latest.js will get the latest version even
-# though it says 2.7.5.
-mathjax_path = 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/latest.js?config=TeX-AMS_HTML-full'
-
 # See https://www.sympy.org/sphinx-math-dollar/
-mathjax_config = {
-    'tex2jax': {
-        'inlineMath': [ ["\\(","\\)"] ],
-        'displayMath': [["\\[","\\]"] ],
-    },
+mathjax3_config = {
+  "tex": {
+    "inlineMath": [['\\(', '\\)']],
+    "displayMath": [["\\[", "\\]"]],
+  }
 }
 
 # Add any paths that contain templates here, relative to this directory.


ThePauliPrinciple:
Are there any compatibility issues for viewing the documentation with mathjax 3? e.g. in old browser versions?

asmeurer:
According to https://docs.mathjax.org/en/latest/output/browser.html it might not work in IE11.

 Personally I wouldn't really worry about supporting older browsers, unless someone complains about it. And even then, we should really consider if it's worth it. IE11 is going to be end-of-life in under a year https://docs.microsoft.com/en-us/lifecycle/faq/internet-explorer-microsoft-edge. I think the only real reason someone might be using an older browser is if they have a mobile browser on an older phone OS.

asmeurer:
I thought it was working OK, but now I'm not so sure. The longest formulas I know of in the SymPy docs are the ones on this page https://docs.sympy.org/dev/modules/integrals/g-functions.html. In the 1.9 docs, which are still using MathJax v2, the long formulas scroll independently https://docs.sympy.org/latest/modules/integrals/g-functions.html. In dev, they overflow into the sidebar. Honestly, both look kind of bad. In the v2 way, it's hard to tell that the formulas scroll at all, at least on Mac where the scrollbars are hidden by default. 

I thought that we were doing line wrapping, but maybe we don't have that feature turned on? 

By the way, this is how the themes proposed at https://github.com/sympy/sympy/issues/22716 are handling this (scroll to the very bottom on each page and look at the last formula on the page)

https://bertiewooster.github.io/sympy-doc/book/modules/simplify/hyperexpand.html
https://bertiewooster.github.io/sympy-doc/pydata/modules/simplify/hyperexpand.html
https://bertiewooster.github.io/sympy-doc/furo/modules/simplify/hyperexpand.html
https://bertiewooster.github.io/sympy-doc/rtd/modules/simplify/hyperexpand.html

Furo and Book put the formula in a box, and RTD and Pydata overflow them on the page. Book and Furo must be putting MathJax formulas in special divs that prevent them from overflowing on the page.

I quickly scrolled through every page in the docs (boy are there a lot of them) to see if I noticed any other overflowing formulas. The only pages I spotted any on were the above two, https://docs.sympy.org/dev/modules/physics/wigner.html, and https://docs.sympy.org/dev/modules/functions/special.html. None of them really overflow that much, so while wrapping would be nice, I don't think it's essential for our present documentation. Certainly not as essential as when using SymPy output in a notebook, where you can often get really long math formulas.

moorepants:
> I thought that we were doing line wrapping, but maybe we don't have that feature turned on?

v3 can't even do it: https://docs.mathjax.org/en/latest/output/linebreaks.html

moorepants:
Also, I guess this is effecting me more because I'm using jupyter-sphinx to output results from sympy calcs. We don't really do that (at least not for rendered latex) in our docs, so it probably doesn't matter.

asmeurer:
By the way, I just checked the PDF documentation for 1.10rc1, and some of the above examples overflow off the page on the PDF (for example, some of the G-Function convergence conditions, see page 558). We could try using `breqn`, although given how few there are, and that they have problems in HTML too, we might want to just add manual line breaks to these equations. 

We might even be able to configure the LaTeX build to fail if there is an equation that is too long. 

Another idea, which we should open an issue for, is to add something to the `latex()` printer to do manual line breaks. This might be difficult, though, as we would have to estimate how long a given piece of latex math is.

asmeurer:
Actually it looks like the MathJax overflow behavior is coming from a CSS variable that we have set https://github.com/sympy/sympy/blob/2192b60ab127c56391a9c7c2a680078326f50de7/doc/src/_static/default.css_t#L348-L353. So most likely we just need to update that variable for MathJax 3. I'm not going to worry about it since we are going to update our theme soon anyway.

PR #22468, opened 2021-11-12
Use headless Chrome instead of Inkscape to convert SVG files to PDF in the PDF docs build

asmeurer:

Previously this used inkscape, but it was misconverting several of the SVG
images, in some cases in ways that made the image completely broken. There are
several other possibilities (see
https://superuser.com/questions/381125/how-do-i-convert-an-svg-to-a-pdf-on-linux#),
including rsvg-convert, convert, and cairosvg, but all have similar issues (I can demonstrate this if anyone is interested).
Additionally, inkscape is a pain to install (at least on Mac), and newer
versions of inkscape changed the command line flags, breaking the Makefile.

A script convert-svg-to-pdf.sh is added that uses headless Chrome or Chromium
to convert the SVG files to PDF, which is based on this gist
https://gist.github.com/s417-lama/84bf66de1096c4587e8187092fb41684. 

- [x] Note: I am still waiting for the author of this gist to confirm that I have permission to include it here. [Edit: [confirmed](https://gist.github.com/s417-lama/84bf66de1096c4587e8187092fb41684#gistcomment-3959850)]

Note that this is only required for building the PDF docs (which also requires
LaTeX). The HTML documentation uses the SVG files directly and does not
require any conversions. Inkscape is no longer a build dependency for the
documentation.

Here's a before after example:

**Master (Inkscape)**

Screen Shot 2021-11-11 at 6 22 10 PM

**This branch (Headless Chrome)**
Screen Shot 2021-11-11 at 6 22 16 PM

You can see the Headless Chrome version matches the [version on the HTML page](https://docs.sympy.org/dev/modules/physics/vector/kinematics.html#angular-velocity).

**Note: as far as I'm aware, it's impossible to know if Travis will work until after this PR is merged, so if something breaks there we will need to fix it in a separate PR.**

CC @moorepants 

#### Release Notes

- physics.vector
  - Fix the rendering of some diagrams in the PDF documentation.


moorepants:
I'm good with this. If you want to add this, it would be nice to include the dependency instructions for building the docs using the `make latexpdf` command. I noticed that was absent from the documentation style guide, for example.

smichr:
I highlighted the TODO that you mentioned in the OP. Do we have the green light for this?

doc/convert-svg-to-pdf.sh
```diff
@@ -0,0 +1,67 @@
...
+tmpfile=$(mktemp XXXXXX.html)
+trap "rm -f $tmpfile" EXIT
+echo $HTML > "$tmpfile"
+
+if command -v google-chrome &> /dev/null
+then
+    CHROME=google-chrome
+elif command -v "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" &> /dev/null
+then
+    CHROME="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
```
oscarbenjamin, line 58: Does this work on Linux?

doc/convert-svg-to-pdf.sh
```diff
@@ -0,0 +1,67 @@
...
+# https://gist.github.com/s417-lama/84bf66de1096c4587e8187092fb41684. See also
+# https://superuser.com/questions/381125/how-do-i-convert-an-svg-to-a-pdf-on-linux
+# and
+# https://stackoverflow.com/questions/44970113/how-can-i-change-paper-size-in-headless-chrome-print-to-pdf
+#
+# Note: if you end up with a pdf with a blank page, you need to edit the SVG
+# file to trim down the width and/or height.
+
+if [ $# -ne 2 ]; then
+  echo "Usage: ./svg2pdf.bash input.svg output.pdf" 1>&2
```
oscarbenjamin, line 15: This usage does not seem correct

doc/convert-svg-to-pdf.sh
asmeurer: The lines above will cover Linux (it works on CI). Basically it's just using the first valid "chrome" command it can find. I don't know if there's a simpler way to do that in bash.

asmeurer:
@smichr yes, see https://gist.github.com/s417-lama/84bf66de1096c4587e8187092fb41684#gistcomment-3959850

smichr:
Here's to more clarity!

PR #22470, opened 2021-11-12
Add 'make latexpdf' to build the PDF documentation in one step

asmeurer:

Previously to build the pdf docs you had to run

```
make latex
export LATEXMKOPTS='-halt-on-error -xelatex'
cd _build/latex
make
```

now you can just run

```
make latexpdf
```

This was also something the output of `make latex` said you could do but didn't actually work.

#### Release Notes

NO ENTRY

### DIFF
--- a/bin/test_sphinx.sh
+++ b/bin/test_sphinx.sh
@@ -6,12 +6,9 @@ echo "Testing SPHINX"
 cd doc
 make html
 make man
-make latex
-cd _build/latex
-export LATEXMKOPTS="-halt-on-error -xelatex -silent"
-make all || {
+make pdflatex LATEXMKOPTS="-halt-on-error -xelatex -silent" || {
     echo "An error had occured during the LaTeX build";
-    tail -n 1000 *.log;
+    tail -n 1000 ./*.log;
     sleep 1; # A guard against travis running tail concurrently.
-    exit -1;
+    exit 1;
 }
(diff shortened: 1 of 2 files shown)


moorepants:
LGTM

smichr:
SPHINX is failing

bin/test_sphinx.sh
```diff
@@ -6,12 +6,9 @@ echo "Testing SPHINX"
 cd doc
 make html
 make man
-make latex
-cd _build/latex
-export LATEXMKOPTS="-halt-on-error -xelatex -silent"
-make all || {
+make pdflatex LATEXMKOPTS="-halt-on-error -xelatex -silent" || {
```
oscargus, line 9:
```suggestion
make latexpdf LATEXMKOPTS="-halt-on-error -xelatex -silent" || {
```

oscarbenjamin:
Looks good.

There is also this:
https://github.com/sympy/sympy/blob/210636b390261ef2a13dccc541853869f3aec1e0/bin/test_travis.sh#L25-L33
and this
https://github.com/sympy/sympy/blob/210636b390261ef2a13dccc541853869f3aec1e0/release/build_docs.py#L41-L47
I guess something should also be added here:
https://github.com/sympy/sympy/blob/master/doc/src/guides/contributing/build-docs.rst

PR #22472, opened 2021-11-12
Add MyST to the documentation build

asmeurer:
MyST allows writing documentation in Markdown. See
https://myst-parser.readthedocs.io/en/latest/index.html for more information.

In particular, MyST allows mixing Markdown and RST documents. It also includes
extensions to the Markdown syntax to support things that would otherwise only
work in RST, such as cross-references and directives.

I do not propose at this point to convert any existing documentation from RST
to Markdown. However, I do plan to write all new documentation that I will be
writing as part of my CZI grant project in Markdown. I should also note that
MyST does not yet support autodoc, so it will not be possible yet to write
docstrings in Markdown, only separate Sphinx documentation guides.

Note that LaTeX math will not work properly with Markdown documents until https://github.com/sympy/sympy/pull/22467 is merged. 

I have pushed, to a separate branch, a demo of what this looks like. [Here is the Markdown source](https://github.com/asmeurer/sympy/blob/myst-demo/doc/src/markdown_demo.md?plain=1) and [here is what the rendered document looks like](https://www.asmeurer.com/sympy-myst-demo/markdown_demo.html).

#### Release Notes

NO ENTRY


moorepants:
I'm not much a fan of this, mostly because we will now have two ways of doing one thing and there are no reported deficiencies in RestructuredText for our documentation that this solves. One consequence is that the documentation style guide will now have to layout parallel descriptions for two markup languages and we have to learn two languages Myst markdown and RestructuredText to work on the documentation.

Could you provide some rationale for this addition? For example, what benefit does this offer other than personal preference of markup language?

I don't intend to try to block this because you are going to be writing a year's worth of documentation and you should enjoy doing that, but it's worth documenting why this change is made.

moorepants:
Also, I think this needs to be edited in this pr: https://docs.sympy.org/latest/documentation-style-guide.html, there are references to narrative documenation, e.g. "reStructuredText: Narrative documentation files and documentation strings embedded in Python code follow the reStructuredText format. Advanced features not described in this document can be found at http://docutils.sourceforge.net/rst.html.". If you add this, the documentation should be updated to explain to users when and how it is used.

asmeurer:
My rationale is that whenever I write anything significant in RST, I invariably find myself spending an enormous amount of time wrestling with its syntax and idiosyncrasies. That's not something I want to deal with over the next two years. I've used MyST for a few different projects now and I can say that it's much simpler to work with. 

It's not just for me, though. Markdown is a syntax that virtually every potential SymPy contributor will know, due to its ubiquitousness in places like GitHub, StackOverflow, Jupyter notebooks, and others. It has a much simpler and easier to remember syntax, which doesn't have any of the unnecessary limitations of RST like the inability to nest markup. Just look at how many people incorrectly use single backticks for code in docstrings (which due to our configuration renders the result as math). If you don't believe me, do an experiment. Ask anybody what the syntax is for creating an external link in Markdown, and what the syntax is in RST. You won't find anyone who knows the RST syntax but not the Markdown syntax. That's not just because Markdown is used more, but also because the RST syntax is unnecessarily complicated and difficult to remember.

The only argument for RST over Markdown has been that RST supports things that Markdown doesn't like cross references, but with MyST, this is no longer an argument, because it supports all those things. And it even does it in a way that's completely compatible with the existing Sphinx tooling. That and the fact that RST is the existing thing that we already use. But it's clearly not working in many ways, and Markdown (with MyST) fixes a lot of the problems with it. I'd honestly like to hear your own reasons for preferring RST, given that MyST now supports all the same markup in Markdown.

Eventually I'd like to move all SymPy documentation (and ideally the whole ecosystem) from RST to Markdown, but that's a much harder task. Markdown docstrings are not yet possible. And even for the existing RST files, they would require translation. I'm not worried about this though because MyST explicitly supports both formats, so we can do a gradual transition rather than trying to go all or nothing.

You're right that I need to update the style guide. Most of the style guide deals with docstrings, which as I noted still must be in RST, but I do need to document the Markdown.

moorepants:
> I'd honestly like to hear your own reasons for preferring RST, given that MyST now supports all the same markup in Markdown.

I stated my reasons in my first comment.

doc/src/guides/contributing/documentation-style-guide.rst
```diff
@@ -172,6 +171,19 @@ Reference documentation should go in the docstrings and be pulled into the RST
...
 that is not a reference for a single specific function.
 
+Documentation using Markdown
+----------------------------
+
+Narrative documentation can be written using either Restructured Text
+(``.rst``) or Markdown (``.md``). Markdown documentation uses `MyST
+<https://myst-parser.readthedocs.io/en/latest/index.html>`_. See `this guide
+<https://myst-parser.readthedocs.io/en/latest/syntax/syntax.html>`_ for more
+information on how to write documents in MArkdown. Markdown is only supported
```
oscarbenjamin, line 181: MArkdown -> Markdown

oscarbenjamin:
Aaron could you maybe add something somewhere that demonstrates what the difference would be between markdown and RST? For example if we took a nontrivial rst file and rewrote it as markdown then how would that look?

asmeurer:
Sure, I can translate one of the existing pages to Markdown as an demonstration. The installation page should enough markup to make it clear, and it is distinct from the other pages in the docs.

smichr:
> mostly because we will now have two ways of doing one thing

But if the two ways can do the same thing, then the easier way will become the preferred way. And the harder way will be used to do what can't be done the other way (or the other way will learn how to do it).

Let's see what Aaron's example of the differences looks like.

asmeurer:
I've translated the installation page from RST to Markdown using [rst-2-myst](https://rst-to-myst.readthedocs.io/en/latest/index.html). The resulting page is rendered exactly the same.

oscarbenjamin:
I think I would find it easier to write docs in this format.

bertiewooster:
@asmeurer Great idea, so folks can leverage content from the wiki, etc. Could you share how to use MyST to create documentation pages in Markdown for docs.sympy.org? Thanks!

asmeurer:
Once this PR is merged it will be as simple as using `.md` as your file suffix instead of `.rst`.

bertiewooster:
Great! Between Markdown and [making it easier to build documentation](https://github.com/sympy/sympy/issues/22655), more people will be able to contribute to the documentation.

asmeurer:
Merging this because I am going to start using it.

PR #22476, opened 2021-11-12
Matrix expression derivatives are now computed with the array expressions module

Upabjojr:
- matrices
  - Derivatives of matrix expressions are now computed in the array expressions module. The result is therefore able to represent higher dimensional arrays. If trivial dimensions are present, an attempt is made at converting the resulting array expression back to matrix expression.

### DIFF
--- a/sympy/matrices/expressions/matexpr.py
+++ b/sympy/matrices/expressions/matexpr.py
@@ -12,7 +12,7 @@ from sympy.external.gmpy import SYMPY_INTS
 from sympy.functions import conjugate, adjoint
 from sympy.functions.special.tensor_functions import KroneckerDelta
 from sympy.matrices.common import NonSquareMatrixError
-from sympy.matrices.matrices import MatrixKind
+from sympy.matrices.matrices import MatrixKind, MatrixBase
 from sympy.multipledispatch import dispatch
 from sympy.simplify import simplify
 from sympy.utilities.misc import filldedent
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
This is the second time that I see a bug report bisected back to this PR:
```python
from sympy import *

x = MatrixSymbol('x', 2, 1)
fx = x.T * MatMul(Matrix([[1, 2], [4, 7]]), x) + x.T * Matrix([3, 5]) + 6*Identity(1)

fx.diff(x)
```
That code used to work but as of this PR fails with RecursionError.

oscarbenjamin:
https://stackoverflow.com/a/69876854/9450991

Upabjojr:
This PR has expanded considerable the matrix derivation algorithm capabilities. It could be that something has been broken. Consider that the old algorithm is still there (in `_matrix_derivative_old_algorithm` ).

Upabjojr:
First observations:

* `convert_matrix_to_array( )` only converts matrix expressions into array expressions. Maybe it should support component-explicit matrices as well?
* `array_derive` does not overload `MatrixBase`. A case could be added to handle that expression.

Upabjojr:
@oscarbenjamin 

I would try to add something like:
```python
@array_derive.register(MatrixBase)
def _(expr: MatrixBase, x: Expr):
    if len(set.intersection(expr.free_symbols, x.free_symbols)) == 0:
        return ZeroArray(*x.shape, *expr.shape)
    if isinstance(x, MatrixExpr) and all(isinstance(i, (int, Integer)) for i in x.shape):
        x = x.as_explicit()
    if isinstance(x, MatrixBase):
        return derive_by_array(expr, x)
    raise NotImplementedError("could not determine derivative")
```
in file _sympy/tensor/array/expressions/arrayexpr_derivatives.py_ , I think this fixes the issue.

oscarbenjamin:
> This PR has expanded considerable the matrix derivation algorithm capabilities.

I'm sure it has but I very often find when bisecting a bug that I land in a PR that has been self-merged without review from anyone else. There are only a few contributors who do this but unreviewed PRs turn up disproportionately when investigating bugs.

This PR has no reviews. There are no references to other issues or PRs or anything that anyone could use to investigate the problem because clearly this was opened and merged without any intention that anyone would review it.

Upabjojr:
> I'm sure it has but I very often find when bisecting a bug that I land in a PR that has been self-merged without review from anyone else. There are only a few contributors who do this but unreviewed PRs turn up disproportionately when investigating bugs.

I have implemented all of the matrix derivation algorithm on my own. I'm not even sure there's anyone else who's looked into the algorithm.

> This PR has no reviews. There are no references to other issues or PRs or anything that anyone could use to investigate the problem because clearly this was opened and merged without any intention that anyone would review it.

Waiting for a review would have likely lasted eternally.

oscarbenjamin:
> I have implemented all of the matrix derivation algorithm on my own. I'm not even sure there's anyone else who's looked into the algorithm.

One thing I have looked into is how some of these changes have messed up ordinary differentiation which is now more convoluted and slower because of the array differentiation dispatching. If I had noticed those changes being made then I would have objected to them but I guess it happened in a different PR that was merged without significant review or explanation.

> Waiting for a review would have likely lasted eternally.

If it is difficult to get a review then sometimes it can be reasonable to merge something without a significant review but here as in many other cases you clearly had no intention of allowing anyone to review this. The PR was opened without any explanation and then merged on the same day.

Sometimes actually it is better for something not to be merged at all than to merge in things that are unreviewed, not well tested or buggy. Review is supposed to improve code before it gets merged. The code that you write and self-merge is not perfect and often has bugs or other things that could be improved. That is why you should not routinely self-merge PRs especially if it is a large or significant change as you claim this one is.

Upabjojr:
The old matrix derivative algorithm is unmaintained (and a real mess, if you want to look at the code), so this PR had to happen.

Issue #22453, opened 2021-11-09
zoo + inf -> ?


smichr:
```python
>>> zoo+oo
nan
>>> zoo+oo*I
zoo + oo*I
>>> _+ oo
nan
```
`zoo` could be `+/-1 +/- I*oo`, `+/-oo +/- I*oo`, or '+/-oo +/- I`. I'm not sure why adding `oo` should give nan and not `I*oo`.

oscarbenjamin:
See also #17701

asmeurer:
zoo isn't just one of four different directional infinities. It represents infinity in *all* directions. It's only a coherent quantity in contexts where the result would be the same in any direction. This happens to be true in quite a few contexts due to the nature of analytic functions in the complex plane. 

oo is a fundamentally real concept and it's hard relate it to zoo in a reasonable way. zoo + oo being nan seems like a fine relation. That's not dissimilar from oo - oo. zoo + oo == zoo also seems fine. If we understand oo as a limiting value on the positive real axis and I*oo as a limiting value on the positive imaginary axis, then they both ought to behave in the same way.

jksuom:
In principle, the sum `a + b` of two "infinite numbers" should be computed as follows: For two sequences of complex numbers, `a_n` and `b_n`, converging to `a` and `b`, respectively, consider the sequence `a_n + b_n`.  If it has a limit that is independent of the choices of `a_n` and `b_n`, then that should be the value of `a + b`, otherwise the limit should be `nan`.

A question remains to be answered:

What does it mean that `a_n -> a` when `a` is infinite?

For `a = zoo`, the answer is simple: this means that the absolute values `|a_n|` grow without bound (`limit(abs(a_n), n, oo) == oo`).
For `a = oo`, the same condition needed but, in addition, we also have to restrict the arguments of the `a_n`. The most natural condition seems to be `limit(arg(a_n), n, oo) == 0`.
Similarly, for `a = I*oo`, the latter limit would be `pi/2`. In fact, this definition can be extended to all values of type `c*oo`, where `c` is a non-zero complex number: the condition would be `limit(arg(a_n), n, oo) == arg(c)`.

[...]

PR #22480, opened 2021-11-13
zoo + inf -> nan

smichr:

discussed in #22453

#### Other comments

#### Release Notes

* core

### DIFF
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -216,15 +216,20 @@ class Add(Expr, AssocOp):
                         # e.g. 3 + ...
         order_factors = []
 
-        extra = []
-        saw_inf = None
-        _nan = [S.NaN], [], None
+        saw_inf = None  # flag to consult when encountering zoo
+
+        extra, seq = sift(seq, lambda _: isinstance(_, MatExpr), binary=True)
+        def _nan():
+            if extra:
+                extra.insert(0, S.NaN)
+                return [] extra, None
+            return [S.NaN], [], None
 
         for o in seq:
 
             if o.is_infinite:
                 if saw_inf and S.ComplexInfinity in (o, coeff):
-                    return _nan
+                    return _nan()
                 if saw_inf is None:
                     saw_inf = True
                 if o is S.ComplexInfinity:
@@ -247,19 +252,16 @@ class Add(Expr, AssocOp):
             # 3 or NaN
             elif o.is_Number:
                 if o is S.NaN:
-                    return _nan
+                    return _nan()
                 elif coeff.is_Number or isinstance(coeff, AccumBounds):
                     coeff += o
+                    if coeff is S.NaN:
+                        return _nan()
                 # XXX ignore Number if coeff is TensExpr?
                 continue
 
             elif isinstance(o, AccumBounds):
                 coeff = o.__add__(coeff)
-                continue
-
-            elif isinstance(o, MatrixExpr):
-                # can't add 0 to Matrix so make sure coeff is not 0
-                extra.append(o)
                 continue
 
             elif isinstance(o, TensExpr):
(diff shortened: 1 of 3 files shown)


sympy/core/add.py
```diff
@@ -191,7 +191,6 @@ def flatten(cls, seq):
         ========
 
         sympy.core.mul.Mul.flatten
```
smichr, line 193:
```suggestion
        sympy.core.mul.Mul.flatten

```

sympy/core/add.py
```diff
@@ -281,7 +279,7 @@ def flatten(cls, seq):
             elif o.is_Pow:
                 b, e = o.as_base_exp()
                 if b.is_Number and (e.is_Integer or
-                                   (e.is_Rational and e.is_negative)):
+                                    (e.is_Rational and e.is_negative)):
```
smichr, line 282:
```suggestion
                                   (e.is_Rational and e.is_negative)):
```

sympy/core/add.py
```diff
@@ -216,10 +216,25 @@ def flatten(cls, seq):
...
         order_factors = []
 
-        extra = []
+        saw_inf = None  # flag to consult when encountering zoo
+
+        extra, seq = sift(seq, lambda _: isinstance(_, MatExpr), binary=True)
+        def _nan():
+            if extra:
+                extra.insert(0, S.NaN)
+                return [] extra, None
```
smichr, line 225:
```suggestion
                return [], extra, None
```

sympy/core/add.py
```diff
@@ -216,10 +216,25 @@ def flatten(cls, seq):
                         # e.g. 3 + ...
         order_factors = []
 
-        extra = []
+        saw_inf = None  # flag to consult when encountering zoo
+
+        extra, seq = sift(seq, lambda _: isinstance(_, MatExpr), binary=True)
```
smichr, line 221:
```suggestion
        extra, seq = sift(seq, lambda _: isinstance(_, MatrixExpr), binary=True)
```

sympy/core/add.py
```diff
@@ -216,10 +216,25 @@ def flatten(cls, seq):
                         # e.g. 3 + ...
         order_factors = []
 
-        extra = []
+        saw_inf = None  # flag to consult when encountering zoo
+
+        extra, seq = sift(seq, lambda _: isinstance(_, MatrixExpr), binary=True)
```
jksuom, line 221: I don't think that a sum could be a mixture of scalars and matrices (or tensors). It should be an error (maybe `ValueError`) if both `extra` and `seq` were nonempty. If `seq` is empty, then there is no need to deal with infinities or `nan`, and `extra` could probably be returned without much further processing; in particular, it would not be necessary to add `nan` to the argument list.

oscarbenjamin:
@smichr please don't add branches in the sympy repo. Do it from your fork and then open the PR from there.

oscarbenjamin:
This needs close attention. The `Add.flatten` method is already way too complicated and this could make that worse.

sympy/core/add.py
smichr: > don't add branches in the sympy repo

I guess this happens if a file is modified online and a PR is started. I am not sure how to avoid this other than not starting a pr this way. I think it does this even if I have a URL with smichr/sympy in it since when I edit a file it doesn't have smichr/sympy it has sympy/sympy in the path.


I will reopen and delete this version.

smichr: > I don't think that a sum could be a mixture of scalars and matrices

This is a separate issue and an inline in `test_matexpr` says,
```
    # Some of these are nonsensical, but we do not raise errors for Add
    # because that breaks algorithms that want to replace matrices with dummy
    # symbols.
```
In this PR I am not focusing on what to do with MatExpr, just trying to faithfully preserve the intent of the code as it is wrt MatExpr and Add.flatten.

PR #22481, opened 2021-11-13
Fix benchmarks build on GitHub Actions

### LINKED ISSUE #22459: Benchmarks builds are failing on CI
Is there any known issue that can cause this? Considering this is also happening in a PR not from me, I don't think this is related to the PRs directly.

https://github.com/sympy/sympy/pull/22209
https://github.com/sympy/sympy/pull/22458
https://github.com/sympy/sympy/pull/22456
https://github.com/sympy/sympy/pull/22449
https://github.com/sympy/sympy/pull/22461
https://github.com/sympy/sympy/pull/22463
https://github.com/sympy/sympy/pull/22464
https://github.com/sympy/sympy/pull/22390

```
Run asv machine --yes --config asv.conf.actions.json
  asv machine --yes --config asv.conf.actions.json
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.9.8/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.9.8/x64/lib
· Config file /home/runner/work/sympy/sympy/asv.conf.json not found.
```

asmeurer:
The build is broken because of an argparse bug introduced in Python 3.9.8,
which breaks asv machine --config (it ignores the given filename and only
looks at the default file). This won't be fixed until 3.9.10, so for now, we
work around this by renaming asv.conf.actions.json to the expected
asv.conf.json. Once 3.9.10 is released and Actions is updated, we can revert
this commit. Travis also runs the benchmarks, but it is still using 3.9.1, so
this issue is not present there (if Travis started failing that wouldn't be a
huge deal anyway).

Fixes #22459.

Please merge as soon as we can confirm this does indeed fix the problem so that the builds can stop failing on other PRs.

#### Release Notes

NO ENTRY


oscarbenjamin:
Thanks @asmeurer and @smichr

asmeurer:
Looks like 3.9.9 was just released fixing this issue. I don't know how long it will take to make its way to GitHub Actions.

PR #22483, opened 2021-11-13
recursive canonical

### LINKED ISSUE #8243: relationals created with symbols can be made canonical
``` python
>>> a=1<x
>>> b=S(1)<x
>>> c=x>S(1)
>>> a==b
False
>>> b==c
False
```

The `a==b` failure is expected since Python flips the args around. But it and the `b==c` case could all be the same if Relationals created from symbols (<, >, etc...) would sort args canonically. This, like sorting arguments of an evaluated Add (so that `x + 2` == `2 + x`), would make working with inequalities a little friendlier.

smichr:
This is for Eq and Ne which can have relational args.

closes #8243 

#### Release Notes

* core
    * canonical will recurse into Relational args in Eq and Ne so `Eq(y<x,x>y).canonical -> True`

### DIFF
--- a/sympy/core/relational.py
+++ b/sympy/core/relational.py
@@ -111,8 +111,7 @@ class Relational(Boolean, EvalfMixin):
             # other than Eq/Ne;
             # Note: Symbol is a subclass of Boolean but is considered
             # acceptable here.
-            if any(map(_nontrivBool, (lhs, rhs))) or isinstance(lhs, Relational
-                    ) or isinstance(rhs, Relational)):
+            if any(map(_nontrivBool, (lhs, rhs))):
                 raise TypeError(filldedent('''
                     A Boolean argument can only be used in
                     Eq and Ne; all other relationals expect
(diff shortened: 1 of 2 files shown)


sympy/core/relational.py
```diff
@@ -111,7 +111,8 @@ def __new__(cls, lhs, rhs, rop=None, **assumptions):
             # other than Eq/Ne;
             # Note: Symbol is a subclass of Boolean but is considered
             # acceptable here.
-            if any(map(_nontrivBool, (lhs, rhs))):
+            if any(map(_nontrivBool, (lhs, rhs))) or isinstance(lhs, Relational
+                    ) or isinstance(rhs, Relational)):
```
smichr, line 115:
```suggestion
            if any(map(_nontrivBool, (lhs, rhs))):
```

sympy/core/relational.py
```diff
@@ -266,7 +266,7 @@ def canonical(self):
         >>> (-y < -x).canonical
         x < y
         """
-        args = self.args
+        args = [i.canonical if isinstance(i, Relational) else i for i in self.args]
```
smichr, line 269:
```suggestion
        args = tuple([i.canonical if isinstance(i, Relational) else i for i in self.args])
```

oscargus:
LGTM.

PR #22485, opened 2021-11-13
test for bare expressions in test suite

smichr:
Bare expressions can represent a logical error that passes silently. e.g.
```
x + 1 instead of x += 1
x == 1 instead of x = 1
```
This PR adds checking for such expressions using ast.

It increases code-quality checking from 6 seconds to 32 seconds (based on observation of run-times of this and another PR).

#### Other comments

Got a tutorial on [ast here](https://www.mattlayman.com/blog/2018/decipher-python-ast/) and made this attempt at checking for relationals and binary ops that appear alone as an expression.

- [x] add tests -- where? I put code in the docstring which passes locally when run interactively; but this is not tested by the docstring tester, so it is there for show
- [x] get the body of if-blocks to be tested (and while-blocks, I suppose) -- see doctest with ">>"
    this turns out to not be necessary; by simply removing the If visitor handler, all Expr are traversed whereever they may appear. The one exception is code in a `with-block`; no checking is done in that context since the visitor handler for `with` does not process the nodes in that node.
- [x] only run on test files; code itself should be tested already

#### Release Notes

* other
    * bare Relational (x == 1) and binary operations (x + 1) will raise a code-quality error if they are in the test suite and not in a `with-block`

### DIFF
--- a/sympy/core/tests/test_arit.py
+++ b/sympy/core/tests/test_arit.py
@@ -1809,7 +1809,7 @@ def test_Mod():
     assert Mod(1, nan) is nan
     assert Mod(nan, nan) is nan
 
-    Mod(0, x) == 0
+    assert Mod(0, x) == 0
     with raises(ZeroDivisionError):
         Mod(x, 0)
 
@@ -1917,13 +1917,13 @@ def test_Mod():
     assert (factorial(n + 4) % (n + 5)).func is Mod
 
     # Wilson's theorem
-    factorial(18042, evaluate=False) % 18043 == 18042
+    assert factorial(18042, evaluate=False) % 18043 == 18042
     p = Symbol('n', prime=True)
-    factorial(p - 1) % p == p - 1
-    factorial(p - 1) % -p == -1
-    (factorial(3, evaluate=False) % 4).doit() == 2
+    assert factorial(p - 1) % p == p - 1
+    assert factorial(p - 1) % -p == -1
+    assert (factorial(3, evaluate=False) % 4).doit() == 2
     n = Symbol('n', composite=True, odd=True)
-    factorial(n - 1) % n == 0
+    assert factorial(n - 1) % n == 0
 
     # symbolic with known parity
     n = Symbol('n', even=True)
@@ -1955,7 +1955,7 @@ def test_Mod():
     from sympy.functions.elementary.piecewise import Piecewise
 
     x_r, y_r = symbols('x_r y_r', real=True)
-    (Piecewise((x_r, y_r > x_r), (y_r, True)) / z) % 1
+    assert (Piecewise((x_r, y_r > x_r), (y_r, True)) / z) % 1
     expr = exp(sinh(Piecewise((x_r, y_r > x_r), (y_r, True)) / z))
     expr.subs({1: 1.0})
     sinh(Piecewise((x_r, y_r > x_r), (y_r, True)) * z ** -1.0).is_zero
--- a/sympy/geometry/tests/test_polygon.py
+++ b/sympy/geometry/tests/test_polygon.py
@@ -570,7 +570,7 @@ def test_first_moment():
     assert p2.first_moment_of_area((a/8, b/6)) == (-25*a*b**2/648, -5*a**2*b/768)
 
     p2 = Polygon((0, 0), (12, 0), (12, 30))
-    assert p2.first_moment_of_area() == (1600/3, -640/3)
+    assert p2.first_moment_of_area() == (S(1600)/3, -S(640)/3)
 
 
 def test_section_modulus_and_polar_second_moment_of_area():
--- a/sympy/geometry/tests/test_polygon.py
+++ b/sympy/geometry/tests/test_polygon.py
@@ -104,7 +104,7 @@ def test_polygon():
     assert p5.encloses_point(Point(4, 0)) is False
     assert p1.encloses(Circle(Point(2.5, 2.5), 5)) is False
     assert p1.encloses(Ellipse(Point(2.5, 2), 5, 6)) is False
-    p5.plot_interval('x') == [x, 0, 1]
+    assert p5.plot_interval('x') == [x, 0, 1]
     assert p5.distance(
         Polygon(Point(10, 10), Point(14, 14), Point(10, 14))) == 6 * sqrt(2)
     assert p5.distance(
@@ -570,7 +570,7 @@ def test_first_moment():
     assert p2.first_moment_of_area((a/8, b/6)) == (-25*a*b**2/648, -5*a**2*b/768)
 
     p2 = Polygon((0, 0), (12, 0), (12, 30))
-    p2.first_moment_of_area() == (1600/3, -640/3)
+    assert p2.first_moment_of_area() == (1600/3, -640/3)
 
 
 def test_section_modulus_and_polar_second_moment_of_area():
--- a/sympy/matrices/tests/test_matrices.py
+++ b/sympy/matrices/tests/test_matrices.py
@@ -1063,7 +1063,7 @@ def test_simplify():
     M = Matrix([[eq]])
     M.simplify()
     assert M == Matrix([[eq]])
-    M.simplify(ratio=oo) == M
+    assert M.simplify(ratio=oo) == M
     assert M == Matrix([[eq.simplify(ratio=oo)]])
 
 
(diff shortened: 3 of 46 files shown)


smichr:
@oscarbenjamin said,

> This looks like a good check. I wonder if flake8 or similar can pick up on this.

Someone more familiar with that would have to see.

Interesting to see that there were many exceptions raised.

See XXX in the diff for questions. I am not sure what to do about the timing tests. For now I just skip files with "bench" in their name.

sympy/core/tests/test_arit.py
```diff
@@ -1955,7 +1955,7 @@ def test_Mod():
     from sympy.functions.elementary.piecewise import Piecewise
 
     x_r, y_r = symbols('x_r y_r', real=True)
-    (Piecewise((x_r, y_r > x_r), (y_r, True)) / z) % 1
+    assert (Piecewise((x_r, y_r > x_r), (y_r, True)) / z) % 1
```
smichr, line 1958:
```suggestion
    assert unchanged(Mod, Piecewise((x_r, y_r > x_r), (y_r, True)) / z, 1)
```

sympy/simplify/tests/test_gammasimp.py
```diff
@@ -67,8 +67,8 @@ def test_gammasimp():
     assert gammasimp(e**2/gamma(k + 1)) == k/gamma(k)
     a = R(1, 2) + R(1, 3)
     b = a + R(1, 3)
```
smichr, line 69: This was an assertion where the left and rhs of the `==` were disconnected; so searching for more than `==` expressions caught this.

sympy/testing/tests/test_code_quality.py
```diff
@@ -147,6 +196,11 @@ def test(fname):
             _test_this_file_encoding(fname, test_file)
 
     def test_this_file(fname, test_file):
+        idx = line_with_bare_expr(test_file.read()) if 'bench' not in fname else None  # XXX what should be done about bench tests
```
oscarbenjamin, line 199: The benchmarks in the sympy repo are broken and should be removed so I think it's fine to skip them here.

sympy/testing/tests/test_pytest.py
```diff
@@ -43,7 +43,7 @@ def test_expected_exception_is_silent_with():
 def test_lack_of_exception_triggers_AssertionError_with():
     try:
         with raises(Exception):
-            1 + 1
+            _ = 1 + 1
```
oscarbenjamin, line 46: This kind of thing is less good but I don't know if there's a way to detect this.

smichr: I'm not sure yet how to handle visitation of a "with" ( or "if") node. If the node includes the head and body of the block then I need to recurse into the body. Perhaps we can ignore contents of "with" blocks. But we need to look at the body of "if" blocks (per note in doctest).

sympy/matrices/tests/test_matrices.py
```diff
@@ -1063,7 +1063,7 @@ def test_simplify():
     M = Matrix([[eq]])
     M.simplify()
     assert M == Matrix([[eq]])
-    M.simplify(ratio=oo) == M
+    assert M.simplify(ratio=oo) == M
```
smichr, line 1066:
```suggestion
    M.simplify(ratio=oo)
```

sympy/polys/domains/tests/test_domains.py
```diff
@@ -983,10 +983,10 @@ def test_CC_double():
 def test_gaussian_domains():
     I = S.ImaginaryUnit
     a, b, c, d = [ZZ_I.convert(x) for x in (5, 2 + I, 3 - I, 5 - 5)]
```
smichr, line 985:
```suggestion
    a, b, c, d = [ZZ_I.convert(x) for x in (5, 2 + I, 3 - I, 5 - 5*I)]
```

sympy/printing/tests/test_mathml.py
```diff
@@ -1249,7 +1249,7 @@ class TestAssocOp(AssocOp):
         identity = 0
 
     expr = TestAssocOp(1, 2)
-    mpp.doprint(expr) == \
+    assert mpp.doprint(expr) == \
         '<mrow><mi>testassocop</mi><mn>2</mn><mn>1</mn></mrow>'
```
smichr, line 1253:
```suggestion
        '<mrow><mi>testassocop</mi><mn>1</mn><mn>2</mn></mrow>'
```

sympy/solvers/tests/test_solvers.py
```diff
@@ -172,7 +172,7 @@ def test_solve_args():
     assert solve(
         (exp(x) - x, exp(y) - y)) == {x: -LambertW(-1), y: -LambertW(-1)}
     # --  when symbols given
-    solve([y, exp(x) + x], x, y) == [(-LambertW(1), 0)]
+    assert solve([y, exp(x) + x], x, y) == [(-LambertW(1), 0)]
```
smichr, line 175:
```suggestion
    assert solve([y, exp(x) + x], x, y) == {y: 0, x: -LambertW(0)}
```

sympy/vector/tests/test_implicitregion.py
```diff
@@ -85,7 +85,7 @@ def test_rational_parametrization():
     assert sphere.rational_parametrization(parameters=(s, t)) == (2/(s**2 + t**2 + 1), 2*t/(s**2 + t**2 + 1), 2*s/(s**2 + t**2 + 1))
 
     conic = ImplicitRegion((x, y), Eq(x**2 + 4*x*y + 3*y**2 + x - y + 10, 0))
-    conic.rational_parametrization(t) == (17/2 + 4/(3*t**2 + 4*t + 1), 4*t/(3*t**2 + 4*t + 1) - 11/2)
+    assert conic.rational_parametrization(t) == (17/2 + 4/(3*t**2 + 4*t + 1), 4*t/(3*t**2 + 4*t + 1) - 11/2)
```
smichr, line 88:
```suggestion
    assert conic.rational_parametrization(t) == (S(17)/2 + 4/(3*t**2 + 4*t + 1), 4*t/(3*t**2 + 4*t + 1) - S(11)/2)
```

sympy/geometry/tests/test_polygon.py
```diff
@@ -570,7 +570,7 @@ def test_first_moment():
     assert p2.first_moment_of_area((a/8, b/6)) == (-25*a*b**2/648, -5*a**2*b/768)
 
     p2 = Polygon((0, 0), (12, 0), (12, 30))
-    p2.first_moment_of_area() == (1600/3, -640/3)
+    assert p2.first_moment_of_area() == (1600/3, -640/3)
```
smichr, line 573:
```suggestion
    assert p2.first_moment_of_area() == (S(1600)/3, -S(640)/3)
```

sympy/utilities/tests/test_wester.py
```diff
@@ -1056,9 +1056,10 @@ def test_M26():
 def test_M27():
     x = symbols('x', real=True)
     b = symbols('b', real=True)
-    with assuming(sin(cos(1/E**2) + 1) + b > 0):
-        # TODO: Replace solve with solveset
-        solve(log(acos(asin(x**R(2, 3) - b) - 1)) + 2, x) == [-b - sin(1 + cos(1/E**2))**R(3/2), b + sin(1 + cos(1/E**2))**R(3/2)]
+    # TODO: Replace solve with solveset which gives both [+/- current answer]
+    # note that there is a typo in this test in the wester.pdf; there is no
+    # real solution for the equation as it appears in wester.pdf
+    assert solve(log(acos(asin(x**R(2, 3) - b)) - 1) + 2, x) == [b + sin(1 + cos(1/E**2 + 1))**R(3/2)]
```
smichr, line 1062:
```suggestion
    with assuming(b + sin(1 + cos(1/E**2 + 1))**R(3, 2) > 0):
        assert solve(log(acos(asin(x**R(2, 3) - b)) - 1) + 2, x) == [b + sin(1 + cos(1/E**2 + 1))**R(3, 2)]
```

smichr:
If this tests pass, this is ready to go.

sympy/testing/tests/test_pytest.py
smichr: expressions in `with-blocks` are not tested

smichr:
The constraint of filtering only test files can be changed. At least we have a working check against changes that easily make their way into the test suite.

PR #22489, opened 2021-11-13
Fixed some import and code quality issues

oscargus:

Started by moving imports to top level and ended up replacing `(-1)**` with `S.NegativeOne` and `Integer(1)` etc.

```
In [5]: %timeit (-1)**x
3.17 µs ± 260 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [6]: %timeit S.NegativeOne**x
1.94 µs ± 33 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

#### Release Notes

NO ENTRY


oscarbenjamin:
Thanks!

PR #22491, opened 2021-11-14
Fixed limit(1/asin(x) , x, 0) which was returning wrong outputs along both directions .

anutosh491:
I discovered this when I was debugging one of my other Prs
On Master
```
>>> limit(1/asin(x), x, 0, dir = '+')
-oo
>>> limit(1/asin(x), x, 0, dir = '-')
oo
```
But this is clearly wrong !

image

And it won't be that hard to find more like these . When I went through the code , there weren't any obvious errors in `limits.py`. The workflow was straightforward in returning the above results but then I realized that we should be using ` cdir` rather than `str(dir)` in the final checks which are in the lines(320 -333) 
```
                    if ex.is_integer:
                        if str(dir) == "-" or ex.is_even:
                            return S.Infinity*sign(coeff)
                        elif str(dir) == "+":
                            return S.NegativeInfinity*sign(coeff)
                        else:
                            return S.ComplexInfinity
                    else:
                        if str(dir) == "+":
                            return S.Infinity*sign(coeff)
                        elif str(dir) == "-":
                            return S.NegativeInfinity*sign(coeff)*S.NegativeOne**(S.One + ex)
                        else:
                            return S.ComplexInfinity
```
This would help us generally and also for the above reported cases . This is because any expression of the form 
`limit( f(x), x, +-oo)` is converted into something like `limit( f(1/x), x, 0)` followed by a change in `cdir` which is done by line 299
```
            cdir = -cdir
```
But we haven't used the changed `cdir` anywhere after the change( nothing but switching signs) . Now that we have made the above explained transformation , we should be dealing with the changed `cdir` rather than the original direction ! This would also solve the above problem and also deal with all general cases better . 

```
>>> limit(1/asin(x), x, 0, dir = '+')
oo
>>> limit(1/asin(x), x, 0, dir = '-')
-oo
```
As I said it won't be that tough to come across this on master . Some basic examples could be something as follows
```
>>> limit(1/sinh(x), x, 0, dir = '+') # Both should be returning oo
-oo
>>> limit(ln(1/x) + 1/sin(x), x, 0, dir = '+')
-oo
>>> limit(ln(1/x) + 1/x, x, 0, dir = '+')
-oo
```
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/series/tests/test_limits.py
+++ b/sympy/series/tests/test_limits.py
@@ -308,6 +308,15 @@ def test_series_AccumBounds():
     assert limit(e, x, oo) == AccumBounds(0, oo)
 
 
+def test_pull_request_22491():
+    assert limit(1/asin(x), x, 0, dir = '+') == oo
+    assert limit(1/asin(x), x, 0, dir = '-') == -oo
+    assert limit(1/sinh(x), x, 0, dir = '+') == oo
+    assert limit(1/sinh(x), x, 0, dir = '-') == -oo
+    assert limit(log(1/x) + 1/sin(x), x, 0, dir = '+') == oo
+    assert limit(log(1/x) + 1/x, x, 0, dir = '+') == oo
+
+
 @XFAIL
 def test_doit2():
     f = Integral(2 * x, x)
--- a/sympy/series/tests/test_limits.py
+++ b/sympy/series/tests/test_limits.py
@@ -85,7 +85,7 @@ def test_basic1():
     assert limit(1/sqrt(x), x, 0, dir='-') == (-oo)*I
     assert limit(x**2, x, 0, dir='-') == 0
     assert limit(sqrt(x), x, 0, dir='-') == 0
-    assert limit(x**-pi, x, 0, dir='-') == -oo*(-1)**(1 - pi)
+    assert limit(x**-pi, x, 0, dir='-') == oo/(-1)**pi
     assert limit((1 + cos(x))**oo, x, 0) == Limit((cos(x) + 1)**oo, x, 0)
 
 
--- a/sympy/series/tests/test_limits.py
+++ b/sympy/series/tests/test_limits.py
@@ -88,6 +88,14 @@ def test_basic1():
     assert limit(x**-pi, x, 0, dir='-') == oo/(-1)**pi
     assert limit((1 + cos(x))**oo, x, 0) == Limit((cos(x) + 1)**oo, x, 0)
 
+    # test pull request 22491
+    assert limit(1/asin(x), x, 0, dir = '+') == oo
+    assert limit(1/asin(x), x, 0, dir = '-') == -oo
+    assert limit(1/sinh(x), x, 0, dir = '+') == oo
+    assert limit(1/sinh(x), x, 0, dir = '-') == -oo
+    assert limit(log(1/x) + 1/sin(x), x, 0, dir = '+') == oo
+    assert limit(log(1/x) + 1/x, x, 0, dir = '+') == oo
+
 
 def test_basic2():
     assert limit(x**x, x, 0, dir="+") == 1
@@ -306,15 +314,6 @@ def test_series_AccumBounds():
     # https://github.com/sympy/sympy/issues/12312
     e = 2**(-x)*(sin(x) + 1)**x
     assert limit(e, x, oo) == AccumBounds(0, oo)
-
-
-def test_pull_request_22491():
-    assert limit(1/asin(x), x, 0, dir = '+') == oo
-    assert limit(1/asin(x), x, 0, dir = '-') == -oo
-    assert limit(1/sinh(x), x, 0, dir = '+') == oo
-    assert limit(1/sinh(x), x, 0, dir = '-') == -oo
-    assert limit(log(1/x) + 1/sin(x), x, 0, dir = '+') == oo
-    assert limit(log(1/x) + 1/x, x, 0, dir = '+') == oo
 
 
 @XFAIL
--- a/sympy/series/limits.py
+++ b/sympy/series/limits.py
@@ -273,7 +273,9 @@ class Limit(Expr):
 
         if e.is_meromorphic(z, z0):
             if abs(z0) is S.Infinity:
-                newe = e.subs(z, -1/z)
+                newe = e.subs(z, 1/z)
+                # cdir changes sign as oo- should become 0+
+                cdir = -cdir
             else:
                 newe = e.subs(z, z + z0)
             try:
@@ -329,7 +331,7 @@ class Limit(Expr):
                         if cdir == 1:
                             return S.Infinity*sign(coeff)
                         elif cdir == -1:
-                            return S.NegativeInfinity*sign(coeff)*S.NegativeOne**(S.One + ex)
+                            return S.Infinity*sign(coeff)*S.NegativeOne**ex
                         else:
                             return S.ComplexInfinity
 
--- a/sympy/series/limits.py
+++ b/sympy/series/limits.py
@@ -270,6 +270,7 @@ class Limit(Expr):
             e = nsimplify(e)
         e = set_signs(e)
 
+
         if e.is_meromorphic(z, z0):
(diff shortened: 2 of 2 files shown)


oscarbenjamin:
I'm not so familiar with the limits code myself so I won't comment on the change you've made but I will say that tests should be added for the behaviour that is changed.

anutosh491:
> I'm not so familiar with the limits code myself so I won't comment on the change you've made but I will say that tests should be added for the behaviour that is changed.

Cool , will add in a bit ! I have couple of tests in mind . As you might have seen,all  these small changes are in the if-elif conditional statements which are followed by return statements so many of the limit tests in test_limits.py would have to go through the changes, before entering the correct condition block to return the output .
So its not a major change , an enhancement of the previous code but it would also subtly work for cases like the one reported above and the general ones !

oscargus:
I think it makes sense to move the `cdir`-assignments (line ~240) to just before it is used. That may mean that the first set of changes that you did can be reverted to use `dir` (depending on where you put that code, it is strictly not required until line 295.

But good find! It clearly makes sense that what you have done now is the correct way to do it.

anutosh491:
> I think it makes sense to move the `cdir`-assignments (line ~240) to just before it is used. That may mean that the first set of changes that you did can be reverted to use `dir` (depending on where you put that code, it is strictly not required until line 295.
> 
> But good find! It clearly makes sense that what you have done now is the correct way to do it.

Thanks @oscargus , will make required changing. Just had a doubt regarding this block (`if e.is_meromorphic(z, z0): `) In this block the transformation which is done is as follows (line 274)
```
        if e.is_meromorphic(z, z0):
            if abs(z0) is S.Infinity:
                newe = e.subs(z, -1/z)
```
And in the following block , we actually want to achieve a similar transformation but we do it differently like (line 293)
```
        if abs(z0) is S.Infinity:
            newe = e.subs(z, 1/z)
            cdir = -cdir
```
1) So for the first case, we basically preserve `cdir` and change `f(x)` to `f(-1/x)` . Probably used for `limit(x, x, oo)` having default `dir ='-'` is transformed to` limit(-1/x, x, 0)` having `dir = '-'`. Now all such cases obviously work as intended but ...

2) In the former case we reverse `cdir` and change` f(x)` to `f(1/x)` . So now this would be used for something like `limit(log(x) , x, oo)` which would be transformed to `limit(log(1/x) , x, 0, dir='+')` . So this changes direction but doesn't change the sign of the expression as above . 

I tried out that both transformations can be used in each others place without any issues . All tests pass but just wanted to know whether there is any particular reason for the first transformation . I personally feel the second one is more apt .
something like the following would work without any issues 

```
        if e.is_meromorphic(z, z0):
            if abs(z0) is S.Infinity:
            cdir = -cdir
```
Let me know what you think of this , other than I'm satisfied that the Pr would tackle all sorts of cases that were being missed earlier like the ones i reported and could be closed . !

oscargus:
I do not really know the code that well, just browsed it from a clarity perspective. (I guess what you say makes sense though, but wouldn't trust me.)

Let's ping @0sidharth and @sachin-4099 and see what they say.

oscargus:
This line: https://github.com/sympy/sympy/blob/02a21dc58189c0316a8cc05f7b71c6a22e5fc7b3/sympy/series/limits.py#L331

can be changed to 

`return S.Infinity*sign(coeff)*S.NegativeOne**ex`

if you want to do that as well.

anutosh491:
> I do not really know the code that well, just browsed it from a clarity perspective. (I guess what you say makes sense though, but wouldn't trust me.)
> 
> Let's ping @0sidharth and @sachin-4099 and see what they say.

Cool no issues , maybe Kalevi @jksuom can have a look too !! Both changes vice-versa don't hinder any test cases whatsoever

anutosh491:
> This line:
> 
> https://github.com/sympy/sympy/blob/02a21dc58189c0316a8cc05f7b71c6a22e5fc7b3/sympy/series/limits.py#L331
> 
> can be changed to
> 
> `return S.Infinity*sign(coeff)*S.NegativeOne**ex`
> 
> if you want to do that as well.

Yeah I agree even this should be done . Thanks for pointing this out !

anutosh491:
> I do not really know the code that well, just browsed it from a clarity perspective. (I guess what you say makes sense though, but wouldn't trust me.)
> 
> Let's ping @0sidharth and @sachin-4099 and see what they say.

I dug into their work and I could make out that the latter had implemented the `if e.is_meromorphic(z, z0):`  block . The  `if abs(z0) is S.Infinity:`  block was introduced by the former because `is_meromorphic` wasn't able to capture all meromorphic functions and many issues were piling up due to this hence that was necessary.  Both implementations took place in their GSoC tenures 
There was not much difference in their objective , wanted to tackle simlar issues but the point is that when `if e.is_meromorphic(z, z0):` was first implemented the cdir keyword wasn't in use ! Hence it hasn't been used in the block even after being a more appropriate transformation ( `newe = e.subs(z, 1/z) , cdir = -cdir`) .
This gives me a much clear idea what needs to be done

0sidharth:
> I tried out that both transformations can be used in each others place without any issues . All tests pass but just wanted to know whether there is any particular reason for the first transformation .

The reason that the `z0 is Infinity` block has a different transformation than `is_meromorphic` block is due to some other issues present in the codebase. I had tried out using a similar transformation as the `is_meromorphic` block when I wrote the `z0 is Infinity` block, but it was throwing some issues which were difficult to fix at the time, so this was a workaround.

If the tests pass with the transformation of `is_meromorphic`  block in `z0 is Infinity` block, then the code can be changed so.

Glancing over the PR, I do have a concern with the usage of `cdir == 1` - the `cdir` parameter need not be restricted to 1, -1 and 0. It can also take other real and even complex values, although I am not sure if this case will arise (maybe `re(cdir) > 0` is a better check). I will be able to look into the PR in more detail after a couple of days.

anutosh491:
> > I tried out that both transformations can be used in each others place without any issues . All tests pass but just wanted to know whether there is any particular reason for the first transformation .
> 
> The reason that the `z0 is Infinity` block has a different transformation than `is_meromorphic` block is due to some other issues present in the codebase. I had tried out using a similar transformation as the `is_meromorphic` block when I wrote the `z0 is Infinity` block, but it was throwing some issues which were difficult to fix at the time, so this was a workaround.
> 
> If the tests pass with the transformation of `is_meromorphic` block in `z0 is Infinity` block, then the code can be changed so.
> 
> Glancing over the PR, I do have a concern with the usage of `cdir == 1` - the `cdir` parameter need not be restricted to 1, -1 and 0. It can also take other real and even complex values, although I am not sure if this case will arise (maybe `re(cdir) > 0` is a better check). I will be able to look into the PR in more detail after a couple of days.

Thank you for your views ! Yeah currently there's no issue using the transformation you did is the above block and as we are using ` cdir` anyways , it's sensible to make the change. 
Ohh I haven't given a thought for cases lying outside 1, 0, -1 will ponder on that . Do you have any specific case in mind ! I dealt with a decent number of cases in the past few days didn't notice any particularly regarding the same . Still will think about it and yeah if we can find something surely `re(cdir) > 0` this would make a better check than the current one

sympy/series/tests/test_limits.py
```diff
@@ -308,6 +308,15 @@ def test_series_AccumBounds():
     assert limit(e, x, oo) == AccumBounds(0, oo)
 
 
+def test_pull_request_22491():
```
0sidharth, line 311: These tests can be added to one of the `test_basic` and `test_log` functions - no need to make a new function. A reference to this PR can be put in the comments under those functions.

0sidharth:
The current `cdir` check seems fine. Once the tests are moved to proper functions, this can be marged.

anutosh491:
> The current `cdir` check seems fine. Once the tests are moved to proper functions, this can be marged.

Thanks , I've made the commit . I feel this is ready to be merged once you have a look !

anutosh491:
Thanks for the reviews @oscargus @0sidharth @smichr !!!

PR #22492, opened 2021-11-14
Reordered imports and moved to top level

oscargus:

Another stab in bringing some order among the imports. This time for integrals and concrete.

The general rationale is that the more general classes and functions should be imported at the top-level at the detailed classes. So `Integral` should be imported at top-level in e.g. manualintegrate rather than the other way around.

#### Other comments

This also affected the solvers module, where imports were moved into the methods. However, there was a clear benefit of moving in that direction (two imports in solvers, many more imports in the other modules)

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/concrete/delta.py
+++ b/sympy/concrete/delta.py
@@ -7,11 +7,15 @@ References
 .. [1] http://mathworld.wolfram.com/KroneckerDelta.html
 
 """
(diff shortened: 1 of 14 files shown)


oscarbenjamin:
Looks good

PR #22496, opened 2021-11-15
zoo + inf -> nan

smichr:

discussed in #22453

#### Other comments

#### Release Notes

* core


smichr:
comments from previous master-based PR:

image

PR #22497, opened 2021-11-15
zoo + inf -> nan

smichr:

discussed in #22453

#### Other comments

#### Release Notes

* core

### DIFF
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -218,6 +218,8 @@ class Add(Expr, AssocOp):
 
         extra = []
 
+        saw_Ioo = False
+
         for o in seq:
 
             # O(x)
@@ -236,15 +238,8 @@ class Add(Expr, AssocOp):
 
             # 3 or NaN
             elif o.is_Number:
-                if (o is S.NaN or coeff is S.ComplexInfinity and
-                        o.is_finite is False) and not extra:
-                    # we know for sure the result will be nan
-                    return [S.NaN], [], None
                 if coeff.is_Number or isinstance(coeff, AccumBounds):
                     coeff += o
-                    if coeff is S.NaN and not extra:
-                        # we know for sure the result will be nan
-                        return [S.NaN], [], None
                 continue
 
             elif isinstance(o, AccumBounds):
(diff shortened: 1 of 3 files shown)


smichr:
comments from previous master-based PR:

image

jksuom:
I think that it is a good idea to collect `extra` first. It is a pity that there are currently "nonsensical" tests that involve sums with both scalars and matrices. I would expect that symbols that are used in matrix algorithms would be of `MatrixKind`. Then the `kind` attribute could be used for creating `extra`. But before that is implemented maybe this is the best we can do.

Otherwise this seems to work as expected wrt. `zoo + inf`.

smichr:
> It is a pity that there are currently "nonsensical" tests

Let's see what happens when I "do the right thing" and reject such interactions. 

I stopped collecting `extra` because there might be items like `Add(M + N, L)`. Instead I just do a check for presence of the non-scalar kinds of interest.

Using `kind` would be nice but that would require that no symbols be used for matrices -- only MatrixSymbol. I'm not sure if the kind system is mature enough to use as a precondition check: that there be only one kind present in the summands.

asmeurer:
I remember `Matrix + nan` being something that I wanted to disable, but I didn't want to remove the hard-coded nan processing that happens in the Add constructor, so I left it as is. Apparently I just wrote this as a comment in the tests rather than making it xfail.

sympy/matrices/expressions/tests/test_matexpr.py
```diff
@@ -405,38 +405,15 @@ def test_MatMul_postprocessor():
...
-    # This is difficult to get working because of the way that Add processes
-    # its args.
-    z = zeros(2)
-    assert Add(z, S.NaN) == Add(S.NaN, z)
-
-
 def test_MatAdd_postprocessor():
-    # Some of these are nonsensical, but we do not raise errors for Add
-    # because that breaks algorithms that want to replace matrices with dummy
-    # symbols.
```
asmeurer, line 419: I'm not sure how many of these are covered by the tests. I know it came up in situations of trying to call various simplification functions on matrix expressions. It might be worth checking a few to make sure there aren't any regressions here. See https://github.com/sympy/sympy/pull/15948 for the nitty-gritty details.

sympy/core/add.py
```diff
@@ -216,10 +216,28 @@ def flatten(cls, seq):
...
 
+        saw_inf = None  # flag to consult when encountering zoo
+
         extra = []
 
+        _nan = False
+
+        tensAdd = None
+
+        oocoeff = S.Infinity, S.ComplexInfinity, S.NegativeInfinity, S.ImaginaryUnit*S.Infinity, S.NegativeOne*S.ImaginaryUnit*S.Infinity
```
asmeurer, line 227: We should check the performance implications of this change. For instance, does constructing these objects and checking against them in every Add.flatten call slow it down?

PR #22500, opened 2021-11-15
Array expression objects are not evaluated by default anymore

Upabjojr:
This PR makes are classes in `sympy.tensor.array.expressions` unevaluated by default. That is, the class constructor will not perform any transformation or simplification. They will indeed be abstract syntax trees (AST) for array expressions.

Use functions like `tensorproduct( ... )`, `tensordiagonal( ... )`, `tensorcontraction( ... )`, and `permutedims( ... )` to perform automatic simplifications of array expressions.

NO ENTRY

### Previous behaviour:
```python
In [1]: from sympy.tensor.array.expressions.array_expressions import *

In [2]: A = ArraySymbol("A", shape=(2, 2, 2, 2))

In [4]: ArrayContraction(PermuteDims(A, [3, 2, 1, 0]), (1, 2))
Out[4]: PermuteDims(ArrayContraction(A, (1, 2)), (0 1))
```

_PermuteDims_ and _ArrayContraction_ are swapped, as the current normal form for array expressions reorders the operations so that the array-operations that decrease dimensions are applied before array-operations that do not decrease (or decrease less) the number of dimensions.

### New behaviour:
```python
>>> ArrayContraction(PermuteDims(A, [3, 2, 1, 0]), (1, 2))
ArrayContraction(PermuteDims(A, (0 3)(1 2)), (1, 2))
```
That is, no reordering takes place.

To achieve the same behaviour as before, use the functions instead of the classes:
```python
>>> tensorcontraction(permutedims(A, [3, 2, 1, 0]), (1, 2))
PermuteDims(ArrayContraction(A, (1, 2)), (0 1))
```

### DIFF
--- a/sympy/tensor/array/arrayop.py
+++ b/sympy/tensor/array/arrayop.py
@@ -160,12 +160,12 @@ def tensorcontraction(array, *contraction_axes):
     [a*e + b*g, a*f + b*h],
     [c*e + d*g, c*f + d*h]])
     """
-    from sympy.tensor.array.expressions.array_expressions import ArrayContraction
+    from sympy.tensor.array.expressions.array_expressions import _array_contraction
     from sympy.tensor.array.expressions.array_expressions import _CodegenArrayAbstract
     from sympy.tensor.array.expressions.array_expressions import _ArrayExpr
     from sympy.matrices.expressions.matexpr import MatrixSymbol
     if isinstance(array, (_ArrayExpr, _CodegenArrayAbstract, MatrixSymbol)):
-        return ArrayContraction(array, *contraction_axes)
+        return _array_contraction(array, *contraction_axes)
 
     array, remaining_indices, remaining_shape, summed_deltas = _util_contraction_diagonal(array, *contraction_axes)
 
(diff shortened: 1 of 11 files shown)


Upabjojr:
@redeboer @ThePauliPrinciple I ping you to warn you about this change. It significantly affects the way the module `sympy.tensor.array.expressions` works.

redeboer:
Thanks for notifying! I'll have a better look at this PR tomorrow, but I like what I see in the description and from a quick look through the diff

ThePauliPrinciple:
This looks like a good change. Do you feel it is now at a point where it is time to set up some documentation with a "design pattern"

Upabjojr:
> Do you feel it is now at a point where it is time to set up some documentation with a "design pattern"

We are getting closer. Once we are sure the API will not change anymore, documenting the code will be an option.

But we have to be sure the API will be in its final version. I don't want to risk having people on the forum complaining about the changes, or worse, having examples on StackOverflow pointing to the outdated API.

ThePauliPrinciple:
Is there a way to normalize after you created the tree? e.g. `.simplify()`

Upabjojr:
> Is there a way to normalize after you created the tree? e.g. `.simplify()`

There is `._normalize()`, but it doesn't apply recursively. Maybe we should add support to call it from `.doit()`, as it is already the case for matrix expressions:

```python
In [2]: A = MatrixSymbol("A", 3, 3)

In [3]: MatAdd(A, A)
Out[3]: A + A

In [4]: MatAdd(A, A).doit()
Out[4]: 2*A
```

Upabjojr:
In general, `.doit()` is OK for simple rewritings or triggering unevaluated algorithms. OTOH, `.simplify()` is meant to call computationally intense algorithms, and should usually be avoided for performance reasons (unless you really need to simplify the expressions).

ThePauliPrinciple:
Perhaps 
```python
def normalize(self):
    return self.func( *[arg.normalize() for arg in self.args] )._normalize()
```
Could already catch a bunch of recursive cases?

Upabjojr:
> Could already catch a bunch of recursive cases?

I would add it to `.doit()` so we are compatible with the matrix expressions module (and, in a certain way, that would be the desired behaviour for all of SymPy).

PR #22505, opened 2021-11-16
[feat] mexpand added to core.functions

smichr:

addressed #8516 (though a good automatic solution is still very slow for the equation set in that issue's OP)

When roots of higher-order polynomials are substituted into the equation that generated them, the expansion to check whether the root satisfies the equation can take a long time. The use of `cse` and expansion after each step has made this much more efficient as judged by the solution checking of the added/modified tests. With the current approach, those expansions essentially hang.

#### Other comments

If you want to experiment with an alternative to `mexpand_cse`, pass the alternative simplification function to `check`. Calling as `check()` will use the version in this PR.
```python
def check(simplify=None):
    """return Eq(target, expr) if expr failed to simplify to target,
    otherwise the number of seconds it took to succeed"""
    if simplify is None:
        def simplify(e):
            from sympy.simplify.cse_main import cse
            try:from sympy.core.function import _mexpand as mexpand
            except:from sympy.core.function import mexpand
            r, e = cse(e, list=False)
            for r in reversed(r):
                e = mexpand(e, recursive=True).xreplace(dict([r]))
            e, d = e.as_numer_denom()
            e = mexpand(e, recursive=True)
            return e/d
    from sympy.abc import x, y, z, a, b, c
    from sympy import solve, ordered, Eq
    from time import time
    eqs, v = [x + y + z - a, x*y + y*z + x*z - b, x*y*z - c], [x, y, z]
    sol = solve(eqs, v, manual=True, simplify=False, check=False)
    nroots = 0
    timer = time()
    for s in sol:
        nroots += 1
        reps = dict(zip((x, y, z), s))
        zeros = [e.xreplace(reps) for e in eqs]
        simplified = [simplify(i) for i in zeros]
        if simplified.count(0) == 3:
            continue
        if nroots > 2:
            target = 0
        elif simplified.count(S.NaN) != 3:
            target = S.NaN
        else:
            continue
        return Eq(next(ordered([i for i,j in zip(zeros, simplified) if j != target])), target)
    return time() - timer
```
#### Release Notes

* core
    * `_mexpand` is now called `mexpand` 
    * `mexpand_cse` will used the cse of an expression to inject 1 subexpression at a time while applying `mexpand`; use for expressions that contain nesting and repeated sub-expressions


sympy/solvers/solvers.py
```diff
@@ -232,6 +235,8 @@ def checksol(f, symbol, sol=None, **flags):
            make positive all symbols without assumptions regarding sign.
 
     """
+    from sympy.simplify.radsimp import numer
+    from sympy.core.function import mexpand
```
oscargus, line 239: This (mexpand) should be possible to import at the top level.

oscargus: Probably numer as well.

smichr:
@oscargus @oscarbenjamin  what do you think about naming the new function `cse_mexpand` and making `_mexpand` into `mexpand`?

oscarbenjamin:
Is the name `mexpand` rather than `_mexpand` intended to mean that it is a public function for users? If so it needs better docs.

sympy/solvers/tests/test_solvers.py
```diff
@@ -2427,3 +2427,27 @@ def test_issue_21942():
...
+    # demonstrate that the solutions satisfy the original equation
+    # for 6 of the roots. So not this:
+    # reps = {a:6, b: 11, c: 6}
+    # for s in sol:
+    #     print([i.xreplace(reps).n(2) for i in s])
+    # but this:
+    nroots = 0
+    for s in sol:
+        reps = dict(zip((x, y, z), s))
+        for e in eqs:
```
smichr, line 2449:
```suggestion
```

sympy/core/function.py
```diff
@@ -2824,19 +2824,60 @@ def _eval_expand_hint(self, **hints):
...
+            was, expr = expr, expand_mul(expand_multinomial(expr))
+            if not recursive:
+                break
+        return expr
+    n, d = fraction(expr, exact=True)
+    if numer or numer is None:
+        n = do(n)
+    if denom or denom is None:
+        if not d.is_Atom:
+            d = do(n)
```
smichr, line 2862:
```suggestion
            d = do(d)
```

sympy/core/function.py
```diff
@@ -2824,19 +2824,60 @@ def _eval_expand_hint(self, **hints):
...
+    """
+    from sympy.simplify.cse_main import cse
+    r, e = cse(e, list=False)
+    for r in reversed(r):
+        e = mexpand(e, recursive=True).xreplace(dict([r]))
+    e, d = e.as_numer_denom()
+    if _final_denom:
+        e, d = [mexpand(_, recursive=True) for _ in (e, d)]
+    else:
+        e = mexpand(e, recursive=True)
```
smichr, line 2879:
```suggestion
    e = mexpand(e, recursive=True)
    if _final_denom:
        d = mexpand(d, recursive=True)
```

sympy/core/tests/test_function.py
```diff
@@ -1026,9 +1026,14 @@ class A(UndefinedFunction):
...
-    assert _mexpand(1) is S.One
-    assert _mexpand(x*(x + 1)**2) == (x*(x + 1)**2).expand()
+    assert mexpand(None) is None
+    assert mexpand(1) is S.One
+    assert mexpand(x*(x + 1)**2) == (x*(x + 1)**2).expand()
+    n, d = (x + 1)**2, (x - 1)**2
+    xn, xd = x**2 + 2*x + 1, x**2 - 2*x + 1
+    assert mexpand(n/d) == xn/xd
+    assert mexpand(n/d, numer=True) == xn/d
+    assert mexpand(n/d, denom=True) == n/xd
```
smichr, line 1036:
```suggestion
```

sympy/core/function.py
```diff
@@ -2824,12 +2824,21 @@ def _eval_expand_hint(self, **hints):
...
+    give a fully expanded expression.
+
+    EXAMPLES
+    ========
+
+    >>> from sympy.core.function import mexpand
+    >>> from sympy.abc import x
+    >>> eq = ((x + 1)**2 + 1)/(x + 1)**2
+    >>> mexpand(eq)
+    1
```
smichr, line 2840:
```suggestion
    x**2/(x**2 + 2*x + 1) + 2*x/(x**2 + 2*x + 1) + 2/(x**2 + 2*x + 1)
```

sympy/core/tests/test_function.py
```diff
@@ -1026,9 +1026,12 @@ class A(UndefinedFunction):
...
     from sympy.abc import x
-    assert _mexpand(None) is None
-    assert _mexpand(1) is S.One
-    assert _mexpand(x*(x + 1)**2) == (x*(x + 1)**2).expand()
+    assert mexpand(None) is None
+    assert mexpand(1) is S.One
+    assert mexpand(x*(x + 1)**2) == (x*(x + 1)**2).expand()
+    n, d = (x + 1)**2, (x - 1)**2
+    xn, xd = x**2 + 2*x + 1, x**2 - 2*x + 1
+    assert mexpand(n/d) == xn/xd
```
smichr, line 1034:
```suggestion
    assert mexpand(n/d) == (xn/xd).expand()
```

smichr, line 1034:
```suggestion
    n, d = (x + 1)**2, (x - 1)**2
    xn, xd = x**2 + 2*x + 1, x**2 - 2*x + 1
    assert mexpand(n/d) == (xn/xd).expand()
```

sympy/core/function.py
```diff
@@ -2824,19 +2824,95 @@ def _eval_expand_hint(self, **hints):
...
+def mexpand_cse(e, _final_denom=True):
+    """return a recursively expanded expression using
+    multinomial and product expansions, but introduce
+    any repeated subexpressions 1 at a time while doing so.
+    Unlike `mexpand`, the results here will always be
+    expressed as a rational expression (but common factors
+    will not necessarily be automatically be cancelled).
+
+    EXAMPLES
+    ========
```
oscarbenjamin, line 2897: The one line summary should be more like one line at the start of the docstring and EXAMPLEs should be capitalised like Examples.

sympy/core/function.py
```diff
@@ -2824,19 +2824,95 @@ def _eval_expand_hint(self, **hints):
...
+
+    >>> from sympy.core.function import mexpand_cse, mexpand
+    >>> from sympy.abc import x
+    >>> eq = 1 + (x + 1)**2/(1 - (x + 1)**2)
+    >>> mexpand_cse(eq)
+    1/(-x**2 - 2*x)
+    >>> mexpand(eq)
+    x**2/(-x**2 - 2*x) + 2*x/(-x**2 - 2*x) + 1 + 1/(-x**2 - 2*x)
+    """
```
oscarbenjamin, line 2906: This could use a See Also section

sympy/core/power.py
```diff
@@ -1757,7 +1757,7 @@ def mul(d1, d2):
             res += terms[e1]*inco*x**(ex)
 
         if not (e.is_integer and e.is_positive and (e*d - n).is_nonpositive and
-                res == _mexpand(self)):
+                res == mexpand(self)):
```
oscarbenjamin, line 1760: What are the implications of this for performance?

sympy/core/function.py
smichr: That's where it is. Notice that this is the `mexpand_cse` function, not the `mexpand` function.

sympy/core/power.py
smichr: none -- there are two functions. `_mexpand` has been promoted to public as `mexpand` and a new function `mexpand_cse` is available but not imported by default.

Actually there will be some impact: expressions like those added will test very quickly because `check_sol` is using `mexpand_cse` now.

sympy/core/function.py
```diff
@@ -2824,19 +2824,105 @@ def _eval_expand_hint(self, **hints):
...
+    >>> mexpand(eq, numer=True)
+    0
+    >>> mexpand(eq, denom=True)
+    -1 + (x - 1)**2/(1 - x)**2
+    >>> mexpand(eq)
+    0
+
+    See Also
+    ========
+    mexpand_cse
```
oscarbenjamin, line 2864: We also want the see also for `expand` I think which in turn should reference all the `expand_*` functions. It probably makes sense to add a see also to `expand` linking back here as well.

smichr:
@sylee957 , can you help me with the Sphinx error? Where does sphinx get this information?

sylee957:
There are some errors for the missing references.
```
/home/runner/work/sympy/sympy/sympy/core/function.py:docstring of sympy.core.function.expand:319: WARNING: py:obj reference target not found: mexpand

/home/runner/work/sympy/sympy/sympy/core/function.py:docstring of sympy.core.function.expand:319: WARNING: py:obj reference target not found: mexpand_cse
```
You may need to add the added function docs to `core.rst` or the new seealso links would not automatically work.

sympy/solvers/solvers.py
```diff
@@ -39,12 +39,14 @@
...
     powdenest, nsimplify, denom, logcombine, sqrtdenest, fraction,
     separatevars)
+from sympy.simplify.radsimp import numer
 from sympy.simplify.sqrtdenest import sqrt_depth
 from sympy.simplify.fu import TR1, TR2i
 from sympy.matrices.common import NonInvertibleMatrixError
 from sympy.matrices import Matrix, zeros
 from sympy.polys import roots, cancel, factor, Poly
 from sympy.polys.polyerrors import GeneratorsNeeded, PolynomialError
+from sympy.polys.polyutils import illegal as _illegal
```
smichr, line 49:
```suggestion
```

PR #22506, opened 2021-11-16
Bessel and Meijer-G functions added to MpmathPrinter

scimax:
Hello everyone,

I was working with some bessel functions, but wanted to export it to `mpmath` via pretty-printing. However, I realized that some functions were not included in the `MpmathPrinter`. So I've added those which I needed. Maybe it's also useful for others, but I have to say, I didn't go through all different functions. Is there a need for other functions? Then the PR would just work as a starting point.

Further, is it useful to add a test?

I hope this is helpful for others.

Thanks!

- `hyper`, `meijerg`, `besselj`, `bessely`, `besseli`, `besselk` add to `_known_functions_mpmath` in pycode printer.

#### Release Notes

* printing
  * Meijer-G function and Bessel functions in MpmathPrinter

### DIFF
--- a/sympy/printing/pycode.py
+++ b/sympy/printing/pycode.py
@@ -523,6 +523,12 @@ _known_functions_mpmath = dict(_in_mpmath, **{
     'fresnels': 'fresnels',
     'sign': 'sign',
     'loggamma': 'loggamma',
+    'hyper': 'hyper',
+    'meijerg': 'meijerg',
+    'besselj': 'besselj',
+    'bessely': 'bessely',
+    'besseli': 'besseli',
+    'besselk': 'besselk',
 })
 _known_constants_mpmath = {
     'Exp1': 'e',


oscargus:
Looks good! Thanks!

I think that there may be some more functions. The best would probably be to go through https://github.com/sympy/sympy/tree/master/sympy/functions/special and https://mpmath.org/doc/current/functions/index.html to check which ones can be added. The error and airy functions for example seems to not be supported.

Regarding tests, I believe that the basic printing functionality is tested for loggamma so should work for other functions in the dict as well, but in case some functions have multiple arguments it can make sense to add a test for that to make sure the arguments ends up in the expected order. (It is quite likely though that they match.)

oscargus:
I'll merge this for now, but if you continue to add functions, which is appreciated, please add some tests for argument orders. (I guess they are the same as probably whoever wrote the SymPy code was inspired by the mpmath functions.) Thanks!

PR #22508, opened 2021-11-16
stats/rv: Density can no longer have None as its argument

### LINKED ISSUE #22507: Density class in stats/rv can have None as one of its arguments
Currently the `Density` class will have `None` as one of its arguments when it is constructed with the `density` function and no value for `condition` is provided. This should not be allowed since `Density` is a subclass of `Basic`

ThePauliPrinciple:

Fixes #22507
See also: https://github.com/sympy/sympy/pull/22445
Previously, `Density` would take `None` as its argument for `condition`.
This has been changed such that it only takes a single argument if
`condition` is `None`. Having one argument less rather than a `None`
value was already supported elsewhere in the class.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/stats/rv.py
+++ b/sympy/stats/rv.py
@@ -852,11 +852,11 @@ def probability(condition, given_condition=None, numsamples=None,
 class Density(Basic):
     expr = property(lambda self: self.args[0])
 
-    def __new__(self, expr, condition = None):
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
Looks good

PR #22513, opened 2021-11-17
Add support for .doit() to array expressions

Upabjojr:
- Add support for .doit() to array expressions
- Using the correct term `canonicalize` for canonicalization of expressions (previous term _normalize_ was not quite proper)
- ArrayElement is now commutative and differentiatable

NO ENTRY

### DIFF
--- a/sympy/tensor/array/expressions/array_expressions.py
+++ b/sympy/tensor/array/expressions/array_expressions.py
@@ -107,7 +107,7 @@ class ArrayElement(_ArrayExpr):
         if s.name != self.name:
             return S.Zero
 
-        return KroneckerDelta(s.indices[0], self.indices[0])*KroneckerDelta(s.indices[1], self.indices[1])
+        return Mul.fromiter(KroneckerDelta(i, j) for i, j in zip(self.indices, s.indices))
 
 
 class ZeroArray(_ArrayExpr):
--- a/sympy/tensor/array/expressions/array_expressions.py
+++ b/sympy/tensor/array/expressions/array_expressions.py
@@ -7,7 +7,7 @@ from typing import Optional, List, Dict as tDict, Tuple as tTuple
 
 import typing
 
-from sympy import Integer
+from sympy import Integer, KroneckerDelta
 from sympy.core.basic import Basic
 from sympy.core.containers import Tuple
 from sympy.core.expr import Expr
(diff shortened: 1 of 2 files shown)


sympy/tensor/array/expressions/tests/test_array_expressions.py
```diff
@@ -677,3 +710,15 @@ def test_array_expr_construction_with_functions():
 
     expr = permutedims(PermuteDims(tp, [1, 0, 2, 3]), [0, 1, 3, 2])
     assert expr == PermuteDims(tp, [1, 0, 3, 2])
+
+
+def test_array_element_expressions():
```
ThePauliPrinciple, line 715: Should there be tests for symbolic elements too?

sympy/tensor/array/expressions/array_expressions.py
```diff
@@ -257,7 +274,14 @@ def _normalize(self):
             diagonal_indices = [tuple(cumulative_ranks2[i] + k for k in j) for i, arg in diagonals.items() for j in arg.diagonal_indices]
             return _permute_dims(_array_diagonal(tp, *diagonal_indices), _af_invert(inverse_permutation))
 
-        return self.func(*args, normalize=False)
+        return self.func(*args, canonicalize=False)
+
+    def doit(self, **kwargs):
+        deep = kwargs.get("deep", True)
```
ThePauliPrinciple, line 280: Just a quick check, should the default behaviour be that the canonicalization is performed recursively?

Upabjojr: Yes, this is compatible with the default behaviour of `.doit()` in SymPy.

sympy/tensor/array/expressions/tests/test_array_expressions.py
Upabjojr: There is one already in the derivatives. The one returning Kronecker deltas. Do you think we need more?

ThePauliPrinciple: I'm looking for the behaviour of e.g. `M[0,0].diff(Symbols('x'))`. For which I'm not even sure how to handle this as we don't know if the element depends on `x`, so perhaps it should never evaluate?

Upabjojr: `M[0, 0]` is a symbol which is different from `x`, so it should return zero. An array symbol is a symbol and its elements are symbols, not more complex expressions.

It could be more complicated if you have an `ArrayElement` object of a non-`ArraySymbol`. Say, you call `ArrayElement( ... some array or array expr ... , ... indices ... )`. In this case we should probably return some unevaluated expression.

PR #22516, opened 2021-11-17
22453a: (oo + I*oo)**(not 0 or 1) -> nan

smichr:
#### Other comments

This is in response to an inline comment in `test_add_flatten` which noted that `1/(oo + I*oo) -> 0*(oo - I*oo)`.

#### Release Notes

* core
    * nan is returned for any power (with exponent not 0 or 1) with infinite real and imaginary portions, e.g. 1/(oo + I*oo)->nan

### DIFF
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -480,10 +480,14 @@ class Add(Expr, AssocOp):
     def _eval_power(self, e):
         from .evalf import pure_complex
         from .relational import is_eq
-        if e.is_zero is False and is_eq(e, S.One
-                    ) is False and all(_.is_infinite for _
-                    in self.as_real_imag()):
-            return S.NaN
+        if len(self.args) == 2 and all(_.is_infinite for _ in self.args):
+            if e.is_zero is False and is_eq(e, S.One) is False:
+                a, b = self.args
+                if not a.coeff(S.ImaginaryUnit):
+                    a, b = b, a
+                ico = a.coeff(S.ImaginaryUnit)
+                if ico and ico.is_extended_real and b.is_extended_real:
+                    return S.NaN
         if e.is_Rational and self.is_number:
             ri = pure_complex(self)
             if ri:
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -478,8 +478,13 @@ class Add(Expr, AssocOp):
     # issue 5524.
 
     def _eval_power(self, e):
+        from .evalf import pure_complex
+        from .relational import is_eq
+        if e.is_zero is False and is_eq(e, S.One
+                    ) is False and all(_.is_infinite for _
+                    in self.as_real_imag()):
+            return S.NaN
         if e.is_Rational and self.is_number:
-            from .evalf import pure_complex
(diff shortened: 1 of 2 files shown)


sympy/core/add.py
```diff
@@ -478,8 +478,13 @@ def as_coeff_Add(self, rational=False, deps=None):
     # issue 5524.
 
     def _eval_power(self, e):
+        from .evalf import pure_complex
+        from .relational import is_eq
+        if e.is_zero is False and is_eq(e, S.One
+                    ) is False and all(_.is_infinite for _
+                    in self.as_real_imag()):
+            return S.NaN
```
smichr, line 486:
```suggestion
        if len(self.args) == 2 and all(_.is_infinite for _ in self.args):
            if e.is_zero is False and is_eq(e, S.One) is False:
                a, b = self.args
                if not a.coeff(S.ImaginaryUnit):
                    a, b = b, a
                ico = a.coeff(S.ImaginaryUnit)
                if ico and ico.is_extended_real and b.is_extended_real:
                    return S.NaN
```

eagleoflqj:
Shouldn't `1/(oo+I*oo)` be just 0?

smichr:
> Shouldn't `1/(oo+I*oo)` be just 0?

`1/(1 + I)` is `(1 - I)/((1 + I)*(1 - I)) -> (1 - I)/2` so I am returning nan by analogy since the denominator would be undefined.

Checking with semi-finite versions, the issue also arises (but is not addressed in this PR, but could be):
```python
>>> 1/(oo+3*I)
0*(oo - 3*I)
>>> (oo+3*I)*(oo-3*I)
(oo - 3*I)*(oo + 3*I)
>>> expand(_)
nan

>>> 1/(3+I*oo)
0*(3 - oo*I)
>>> (3+oo*I)*(3-oo*I)
(3 - oo*I)*(3 + oo*I)
>>> expand(_)
nan
```

smichr:
Any objections?

jksuom:
`nan` is always safe (like `None` in tests) but I think that more precise results could be given. `oo + I*oo` can be regarded as a refinement of `zoo`: its absolute value is infinite but its argument can be thought of being between 0 and `pi/2`. When it is raised to a power greater than 4 (or less than -4), all information of its argument will be lost as `4*pi/2 = 2*pi`, so I think that it could be treated as `zoo` when raised to a power (even if the exponent is between -4 and 4). So, for positive exponents, we could have `zoo` instead of `nan`:
```
In [23]: zoo*2                                                                  
Out[23]: zoo

In [24]: zoo*(S(1)/2)                                                           
Out[24]: zoo
```
and for negative exponents , `1/zoo` which is 0. Only `(oo + I*oo)**0` cannot be determined and could be `nan`.

smichr:
@eagleoflqj , @jksuom -- does this look right now? And is this right to all exponents of +/-oo to give results similar to +/-2 (finite version)?

jksuom:
This looks good to me.

smichr:
Thanks @jksuom

PR #22521, opened 2021-11-18
Added _eval_nseries , _eval_as_leading_term  methods for asech and acsch functions

### LINKED ISSUE #10868: sympy is  unable to solve limit for function log( x ) + asech( x ) at both point x = 0 and x = oo.
while Calculating the limit of function  log(x)+asech(x) at two different point sympy gives unexpected result 

```
>>> limit(log(x)+asech(x),x,oo)
    ⅈ⋅π
∞ + ───
     2 
>>> limit(log(x)+asech(x),x,0)
 lim (log(x) + asech(x))
x─→0⁺                   
>>> 
```

I think that gruntz algorithm has problem while  calculating limit for  function whose two side limit is not exist .
Reference  -https://www.wolframalpha.com/input/i=limit+of+log%28x%29%2Basech%28x%29+where+x+tends+0

anutosh491:

Fixes #10868 

Currently on master , there was no series expansion(maclaurin or puiseux series including logarithm at `0`)  for `asech` and `acsch ` resulting into wrong output for limits etc which is shown as follows . So the `series` method didn't work because the `_eval_nseries` method didn't work which depends on the `taylor_term` method which has been added for the same . Similarly the `_eval_as_leading_term` method has been implemented for these.
```
>>> asech(x).series(x)
nan
>>> acsch(x).series(x)
nan
>>> limit(log(x)+asech(x),x,0, dir = '+') # This issue which has been addressed
oo
>>> limit(log(x)+asech(x),x,0, dir = '-')
oo + I*pi
>>> limit(log(x)+acsch(x),x,0, dir = '-')
zoo
>>> limit(log(x)+acsch(x),x,0, dir = '+')
zoo
```
Changes made 

```
>>> asech(x).series(x, 0, 8)
log(2) - log(x) - x**2/4 - 3*x**4/32 - 5*x**6/96 + O(x**8)

>>> acsch(x).series(x, 0, 8)
log(2) - log(x) + x**2/4 - 3*x**4/32 + 5*x**6/96 + O(x**8)

>>> limit(log(x)+asech(x),x,0, dir = '+')
log(2)
>>> limit(log(x)+asech(x),x,0, dir = '-')
log(2) + 2*I*pi
>>> limit(log(x)+acsch(x),x,0, dir = '-')
-oo
>>> limit(log(x)+acsch(x),x,0, dir = '+')
log(2)

>>> asech(x).as_leading_term(x)
-log(x)
>>> acsch(x).as_leading_term(x)
-log(x)
```

The issue originally also reported calculating `limit(log(x)+asech(x), x , oo)` which maybe ..... could return `oo` (as done by few other CAS libraries) as we would be computing something like `log(oo) + I*pi/2`  . But in this case sympy gives `oo + I*pi/2` which isn't wrong by any means . I guess in sympy as per convention , I don't think `oo + I*(some constant)` is mapped to `oo` hence this is not really an issue and we could maybe have a discussion over it .

#### Release Notes

* series
  * Added _eval_nseries , _eval_as_leading_term methods for some inverse hyperbolic functions.

### DIFF
--- a/sympy/functions/elementary/hyperbolic.py
+++ b/sympy/functions/elementary/hyperbolic.py
@@ -681,13 +681,17 @@ class tanh(HyperbolicFunction):
         return 1/coth(arg)
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy.series.order import Order
         arg = self.args[0].as_leading_term(x)
+        arg0 = arg.subs(x, 0)
 
-        if x in arg.free_symbols and Order(1, x).contains(arg):
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
+        if arg0.is_zero:
             return arg
+        elif arg0.is_finite:
+            return self.func(arg0)
         else:
-            return self.func(arg)
+            return self
 
     def _eval_is_real(self):
         arg = self.args[0]
@@ -883,13 +887,17 @@ class coth(HyperbolicFunction):
             return self.args[0].is_negative
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy.series.order import Order
         arg = self.args[0].as_leading_term(x)
+        arg0 = arg.subs(x, 0)
 
-        if x in arg.free_symbols and Order(1, x).contains(arg):
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
+        if arg0.is_zero:
             return 1/arg
+        elif arg0.is_finite:
+            return self.func(arg0)
         else:
-            return self.func(arg)
+            return self
 
     def _eval_expand_trig(self, **hints):
         arg = self.args[0]
@@ -1201,13 +1209,17 @@ class asinh(InverseHyperbolicFunction):
                 return S.NegativeOne**k * R / F * x**n / n
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy.series.order import Order
         arg = self.args[0].as_leading_term(x)
+        arg0 = arg.subs(x, 0)
 
-        if x in arg.free_symbols and Order(1, x).contains(arg):
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
+        if arg0.is_zero:
             return arg
+        elif arg0.is_finite:
+            return self.func(arg0)
         else:
-            return self.func(arg)
+            return self
 
     def _eval_rewrite_as_log(self, x, **kwargs):
         return log(x + sqrt(x**2 + 1))
@@ -1348,13 +1360,17 @@ class acosh(InverseHyperbolicFunction):
                 return -R / F * S.ImaginaryUnit * x**n / n
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy.series.order import Order
         arg = self.args[0].as_leading_term(x)
+        arg0 = arg.subs(x, 0)
 
-        if x in arg.free_symbols and Order(1, x).contains(arg):
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
+        if arg0.is_zero:
             return S.ImaginaryUnit*S.Pi/2
+        elif arg0.is_finite:
+            return self.func(arg0)
         else:
-            return self.func(arg)
+            return self
 
     def _eval_rewrite_as_log(self, x, **kwargs):
         return log(x + sqrt(x + 1) * sqrt(x - 1))
@@ -1456,13 +1472,17 @@ class atanh(InverseHyperbolicFunction):
             return x**n / n
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy.series.order import Order
         arg = self.args[0].as_leading_term(x)
+        arg0 = arg.subs(x, 0)
 
-        if x in arg.free_symbols and Order(1, x).contains(arg):
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
+        if arg0.is_zero:
             return arg
+        elif arg0.is_finite:
+            return self.func(arg0)
         else:
-            return self.func(arg)
+            return self
 
     def _eval_rewrite_as_log(self, x, **kwargs):
         return (log(1 + x) - log(1 - x)) / 2
@@ -1554,13 +1574,17 @@ class acoth(InverseHyperbolicFunction):
             return x**n / n
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy.series.order import Order
         arg = self.args[0].as_leading_term(x)
+        arg0 = arg.subs(x, 0)
 
-        if x in arg.free_symbols and Order(1, x).contains(arg):
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
+        if arg0.is_zero:
             return S.ImaginaryUnit*S.Pi/2
+        elif arg0.is_finite:
+            return self.func(arg0)
         else:
-            return self.func(arg)
+            return self
 
     def _eval_rewrite_as_log(self, x, **kwargs):
         return (log(1 + 1/x) - log(1 - 1/x)) / 2
@@ -1694,13 +1718,18 @@ class asech(InverseHyperbolicFunction):
                 return -1 * R / F * x**n / 4
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy.series.order import Order
         arg = self.args[0].as_leading_term(x)
+        arg0 = arg.subs(x, 0)
 
-        if x in arg.free_symbols and Order(1, x).contains(arg):
-            return -1*log(arg)
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
+        if arg0.is_zero:
+            c, e = arg.as_coeff_exponent(x)
+            return log(2) - log(c) - e*log(x)
+        elif arg0.is_finite:
+            return self.func(arg0)
         else:
-            return self.func(arg)
+            return self
 
     def inverse(self, argindex=1):
         """
@@ -1823,13 +1852,18 @@ class acsch(InverseHyperbolicFunction):
                 return S.NegativeOne**(k +1) * R / F * x**n / 4
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy.series.order import Order
         arg = self.args[0].as_leading_term(x)
+        arg0 = arg.subs(x, 0)
 
-        if x in arg.free_symbols and Order(1, x).contains(arg):
-            return -1*log(arg)
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
+        if arg0.is_zero:
+            c, e = arg.as_coeff_exponent(x)
+            return log(2) - log(c) - e*log(x)
+        elif arg0.is_finite:
+            return self.func(arg0)
         else:
-            return self.func(arg)
+            return self
 
     def inverse(self, argindex=1):
         """
--- a/sympy/functions/elementary/hyperbolic.py
+++ b/sympy/functions/elementary/hyperbolic.py
@@ -1677,7 +1677,7 @@ class asech(InverseHyperbolicFunction):
 
     @staticmethod
     @cacheit
-    def expansion_term(n, x, *previous_terms):
+    def taylor_term(n, x, *previous_terms):
         if n == 0:
             return log(2 / x)
         elif n < 0 or n % 2 == 1:
@@ -1692,6 +1692,15 @@ class asech(InverseHyperbolicFunction):
                 R = RisingFactorial(S.Half, k) *  n
                 F = factorial(k) * n // 2 * n // 2
                 return -1 * R / F * x**n / 4
+
+    def _eval_as_leading_term(self, x, logx=None, cdir=0):
+        from sympy.series.order import Order
+        arg = self.args[0].as_leading_term(x)
+
+        if x in arg.free_symbols and Order(1, x).contains(arg):
+            return -1*log(arg)
+        else:
+            return self.func(arg)
 
     def inverse(self, argindex=1):
         """
@@ -1794,6 +1803,33 @@ class acsch(InverseHyperbolicFunction):
 
         if arg.could_extract_minus_sign():
             return -cls(-arg)
+
+    @staticmethod
+    @cacheit
+    def taylor_term(n, x, *previous_terms):
+        if n == 0:
+            return log(2 / x)
+        elif n < 0 or n % 2 == 1:
+            return S.Zero
+        else:
+            x = sympify(x)
+            if len(previous_terms) > 2 and n > 2:
+                p = previous_terms[-2]
+                return p * (n - 1)**2 // (n // 2)**2 * x**2 / 4
+            else:
+                k = n // 2
+                R = RisingFactorial(S.Half, k) *  n
+                F = factorial(k) * n // 2 * n // 2
+                return S.NegativeOne**(k +1) * R / F * x**n / 4
+
+    def _eval_as_leading_term(self, x, logx=None, cdir=0):
+        from sympy.series.order import Order
+        arg = self.args[0].as_leading_term(x)
+
+        if x in arg.free_symbols and Order(1, x).contains(arg):
+            return -1*log(arg)
+        else:
+            return self.func(arg)
 
     def inverse(self, argindex=1):
         """
--- a/sympy/series/tests/test_limits.py
+++ b/sympy/series/tests/test_limits.py
@@ -641,10 +641,9 @@ def test_issue_10801():
 
 
 def test_issue_10868():
-    assert limit(log(x)+asech(x),x,0, dir = '+') == log(2)
-    assert limit(log(x)+asech(x),x,0, dir = '-') == log(2) + 2*I*pi
-    assert limit(log(x)+acsch(x),x,0, dir = '+') == log(2)
-    assert limit(log(x)+acsch(x),x,0, dir = '-') == -oo
+    assert limit(log(x)+asech(x),x,0, dir = '+-') == log(2)
+    assert limit(log(x)+acsch(x),x,0, dir = '+-') == log(2)
+
 
 def test_issue_10976():
     s, x = symbols('s x', real=True)
--- a/sympy/series/tests/test_limits.py
+++ b/sympy/series/tests/test_limits.py
@@ -9,7 +9,7 @@ from sympy.core.symbol import (Symbol, symbols)
 from sympy.functions.combinatorial.factorials import (binomial, factorial, subfactorial)
 from sympy.functions.elementary.complexes import (Abs, re, sign)
 from sympy.functions.elementary.exponential import (LambertW, exp, log)
-from sympy.functions.elementary.hyperbolic import (acoth, atanh, sinh)
+from sympy.functions.elementary.hyperbolic import (acoth, atanh, sinh, asech, acsch)
 from sympy.functions.elementary.integers import (ceiling, floor, frac)
 from sympy.functions.elementary.miscellaneous import (cbrt, real_root, sqrt)
 from sympy.functions.elementary.trigonometric import (acos, acot, acsc, asec, asin, atan, cos, cot, sec, sin, tan)
@@ -639,6 +639,12 @@ def test_issue_10801():
     # make sure limits work with binomial
     assert limit(16**k / (k * binomial(2*k, k)**2), k, oo) == pi
 
+
+def test_issue_10868():
+    assert limit(log(x)+asech(x),x,0, dir = '+') == log(2)
+    assert limit(log(x)+asech(x),x,0, dir = '-') == log(2) + 2*I*pi
+    assert limit(log(x)+acsch(x),x,0, dir = '+') == log(2)
+    assert limit(log(x)+acsch(x),x,0, dir = '-') == -oo
 
 def test_issue_10976():
     s, x = symbols('s x', real=True)
--- a/sympy/functions/elementary/tests/test_hyperbolic.py
+++ b/sympy/functions/elementary/tests/test_hyperbolic.py
@@ -713,9 +713,11 @@ def test_asech():
 
 def test_asech_series():
     x = Symbol('x')
-    t6 = asech(x).expansion_term(6, x)
+    assert asech(x).series(x, 0, 8) == \
+        log(2) - log(x) - x**2/4 - 3*x**4/32 - 5*x**6/96 + O(x**8)
+    t6 = asech(x).taylor_term(6, x)
     assert t6 == -5*x**6/96
-    assert asech(x).expansion_term(8, x, t6, 0) == -35*x**8/1024
+    assert asech(x).taylor_term(8, x, t6, 0) == -35*x**8/1024
 
 
 def test_asech_rewrite():
@@ -791,6 +793,15 @@ def test_acsch_infinities():
     assert acsch(oo) == 0
     assert acsch(-oo) == 0
     assert acsch(zoo) == 0
+
+
+def test_acsch_series():
+    x = Symbol('x')
+    assert acsch(x).series(x, 0, 8) == \
+        log(2) - log(x) + x**2/4 - 3*x**4/32 + 5*x**6/96 + O(x**8)
+    t6 = acsch(x).taylor_term(6, x)
+    assert t6 == 5*x**6/96
+    assert acsch(x).taylor_term(10, x, t6, 0) == 63*x**10/2560
 
 
 def test_acsch_rewrite():
@@ -953,6 +964,8 @@ def test_leading_term():
     assert coth(x).as_leading_term(x) == 1/x
     assert acosh(x).as_leading_term(x) == I*pi/2
     assert acoth(x).as_leading_term(x) == I*pi/2
+    assert asech(x).as_leading_term(x) == -log(x)
+    assert acsch(x).as_leading_term(x) == -log(x)
     for func in [sinh, tanh, asinh, atanh]:
         assert func(x).as_leading_term(x) == x
     for func in [sinh, cosh, tanh, coth, asinh, acosh, atanh, acoth]:
--- a/sympy/functions/elementary/tests/test_hyperbolic.py
+++ b/sympy/functions/elementary/tests/test_hyperbolic.py
@@ -964,8 +964,8 @@ def test_leading_term():
     assert coth(x).as_leading_term(x) == 1/x
     assert acosh(x).as_leading_term(x) == I*pi/2
     assert acoth(x).as_leading_term(x) == I*pi/2
-    assert asech(x).as_leading_term(x) == -log(x)
-    assert acsch(x).as_leading_term(x) == -log(x)
+    assert asech(x).as_leading_term(x) == -log(x) + log(2)
+    assert acsch(x).as_leading_term(x) == -log(x) + log(2)
     for func in [sinh, tanh, asinh, atanh]:
         assert func(x).as_leading_term(x) == x
     for func in [sinh, cosh, tanh, coth, asinh, acosh, atanh, acoth]:


sympy/functions/elementary/hyperbolic.py
```diff
@@ -1693,6 +1693,15 @@ def expansion_term(n, x, *previous_terms):
                 F = factorial(k) * n // 2 * n // 2
                 return -1 * R / F * x**n / 4
 
+    def _eval_as_leading_term(self, x, logx=None, cdir=0):
+        from sympy.series.order import Order
+        arg = self.args[0].as_leading_term(x)
+
+        if x in arg.free_symbols and Order(1, x).contains(arg):
```
0sidharth, line 1700: `contains` is an expensive method. This check can simply be - 
```py
        arg0 = arg.subs(x, 0)
        if arg0 is S.NaN:
            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
        if arg0.is_zero:
            return -1*log(arg)
```

anutosh491: Yeah this had crossed this mind actually but then all `trig`, `inverse trig`, `hyperbolic` , `inverse hyperbolic ` use `.contains` for their respective `_eval_as_leading_term` method, hence I thought of going forward with contains for consistency . ! Though won't be an issue if the change is introduced everywhere !! I'll try that out and get back

sympy/functions/elementary/hyperbolic.py
```diff
@@ -1693,6 +1693,15 @@ def expansion_term(n, x, *previous_terms):
...
+            return -1*log(arg)
```
jksuom, line 1701: It seems that this should be something like `return log(2) - log(c) - e*log(x)` when `arg` is `c*x**e`. If `log(2)` is not included, then the leading term of `log(x) + asech(x)` will not be correct. `log(c*x**e)` should be written in the form `log(c) + e*log(x)` so that the `log(x)` terms coming from different parts of a sum would combine without explicit simplifications.

anutosh491: Thanks for pointing this out , I had completely overlooked this aspect but yeah surely what you've said if correct

anutosh491: I was about to make a commit but it seems that this test would fail as it gives back log(2) rather than the output below when I make the above change , will have to look into this !
```
>>> limit(log(x)+asech(x),x,0, dir = '-')
log(2) + 2*I*pi
```

jksuom: log(x) + asech(x) = log(x) + log((1 + sqrt(1 - x**2))/x)
                      = log(1 + sqrt(1 - x**2))

is meromorphic (even holomorphic) at `x = 0` so its limit from any direction should be equal to its value `log(2)`.

anutosh491: Ahh okay now I realize what led to my confusion ! The above returns True and below one returns false but yeah what you said is correct `log(2)` should be the output for the this
```
>>> (log(1 + sqrt(1 - x**2))).is_meromorphic(x, 0)
True
>>> (log(x) + asech(x)).is_meromorphic(x, 0)
>>>
```

sympy/functions/elementary/hyperbolic.py
anutosh491: I have removed the `contains` method from all `hyperbolic` and `inverse hyperbolic` functions of `hyperbolic.py` as this block shown above is more than sufficient for all functions and should perform better than using than `contains`.

anutosh491:
@jksuom @0sidharth  I've made requested change , you'll could have a look . Seems fine to me . Some basic cases have been verified below !

```
>>> asech(x).as_leading_term(x)
-log(x) + log(2)

>>> asech(2*x).series()
-log(x) - x**2 - 3*x**4/2 + O(x**6)
>>> asech(2*x).as_leading_term(x)
-log(x)

>>> asech(x**2).series()
log(2) - 2*log(x) - x**4/4 + O(x**6)
>>> asech(x**2).as_leading_term(x)
-2*log(x) + log(2)

>>> (log(x) + asech(x**2)).as_leading_term(x)
-log(x) + log(2)
```

sympy/series/tests/test_limits.py
```diff
@@ -640,6 +640,11 @@ def test_issue_10801():
     assert limit(16**k / (k * binomial(2*k, k)**2), k, oo) == pi
 
 
+def test_issue_10868():
+    assert limit(log(x)+asech(x),x,0, dir = '+-') == log(2)
```
0sidharth, line 644: Spaces after commas, no space around = in `dir = `keep the formatting consistent and follow PEP8

0sidharth:
A Release Notes Entry should be added. Apart from the one formatting change, LGTM.

jksuom:
Leading term and series expansion methods of multivalued functions like `asech` are fairly hard to implement because of branch cuts. According to Wikipedia, `asech` has branch cuts from `-oo` to 0 and from 1 to `oo`. However, it should be checked that this is how it is defined in `mpmath`. There should be special handling for points on branch cuts, and even more special treatment for endpoints like 0, -1 and 1. The code at `zoo` will also need special care as it is on the branch cut.

jksuom:
Inverse hyperbolic (and trigonometric) functions can be expressed as logarithms of algebraic expressions. It might be possible to use those expressions to derive their leading terms and series expansions without having to deal with each function separately in detail. For that purpose, those methods of `log` have to be improved at 0 (see [this comment](https://github.com/sympy/sympy/issues/22453#issuecomment-967249025) for handling of the branch cut) and at infinities (possibly by using `log(1/x) = - log(x)`). Currently, `log(2*x).as_leading_term(x)` returns `log(2*x)` but it should rather be `log(2) + log(x)`, and `log(x**2).as_leading_term(x)` should be `2*log(x)`. It seems that logarithms should be allowed as "constants" in leading terms as they are expected by `gruntz`.

anutosh491:
Hmm, I get what you're trying to convey but I'll have to give a thought regarding the implementation behind this . As of now , I can see that the `series` method won't work on the committed branch only for two points which are the end points 1 and -1 where I get a `PoleError` in return but a log rewrite works properly for the same !
 
Maybe for now this could be used rather than making the `series` and `as_leading_term` methods of all `inverse hyberbolic` and other  `trig`  functions dependent on `log` as it seems that some good work has to be done on those methods of `log` which deal with branch cuts before this change could be made.

We would just end up using this rewrite(log) method at 4 places which are asec, acsc, asech, acsch  @jksuom and all would work as expected !
```
>>> asech(x).series(x, 1)
sympy.core.function.PoleError: Cannot expand asech(_x + 1) around 0
>>> asech(x).rewrite(log).series(x, 1, 4)
sqrt(2)*I*sqrt(x - 1) - 5*sqrt(2)*I*(x - 1)**(3/2)/12 + 43*sqrt(2)*I*(x - 1)**(5/2)/160 - 177*sqrt(2)*I*(x - 1)**(7/2)/896 + O((x - 1)**4, (x, 1))
```

sympy/functions/elementary/hyperbolic.py
anutosh491: > asech(x)

@jksuom I have been having doubts in this recently , are we always allowed to use ` log` operations ( I don't think there any restrictions cause i've never heard of any)  
For eg
```
>>> asech(x).rewrite(log)
log(sqrt(-1 + 1/x)*sqrt(1 + 1/x) + 1/x)

>>> limit(asech(x).rewrite(log) + log(x), x, 0,'-')
log(2) + 2*I*pi

>>> limit(log(sqrt(-1 + 1/x)*sqrt(1 + 1/x) + 1/x) + log(x), x, 0,'-')
log(2) + 2*I*pi

>>> limit(log(x*sqrt(-1 + 1/x)*sqrt(1 + 1/x) + 1), x, 0,'-') # After performing log simplifcation 
log(2)
```

anutosh491: I'll paste a set of results which I feel are the correct ones (after confirming with wolfram and other CAS)
Ignore the rewrite required in these, all results are correct but in sympy we could only get the correct results as of now on master using the rewrite ! But other than that wolfram agrees to all these answers !
```
>>> limit(log(x) + asech(x).rewrite(log), x,0, '-')
log(2) + 2*I*pi
>>> limit(log(x) + asech(x).rewrite(log), x,0, '+')
log(2)
>>> limit(-log(x) + asech(x).rewrite(log), x,0, '-')
oo
>>> limit(-log(x) + asech(x).rewrite(log), x,0, '+')
oo
>>>
>>>
>>> limit(log(x) + acsch(x).rewrite(log), x,0, '-')
-oo
>>> limit(log(x) + acsch(x).rewrite(log), x,0, '+')
log(2)
>>> limit(-log(x) + acsch(x).rewrite(log), x,0, '-')
-log(2) - I*pi
>>> limit(-log(x) + acsch(x).rewrite(log), x,0, '+')
oo
```
The best part is my first commit was giving all answers perfectly and after trying to consider the suggestion you pointed above now this is what we get 
```
>>> limit(log(x) + asech(x), x, 0, '-')
log(2)
>>> limit(log(x) + asech(x), x, 0, '+')
log(2)
>>> limit(-log(x) + asech(x), x, 0, '-')
oo
>>> limit(-log(x) + asech(x), x, 0, '+')
oo
>>> limit(log(x) + acsch(x), x, 0, '-')
log(2)
>>> limit(log(x) + acsch(x), x, 0, '+')
log(2)
>>> limit(-log(x) + acsch(x), x, 0, '-')
-log(2) - I*pi
>>> limit(-log(x) + acsch(x), x, 0, '+')
oo
```

anutosh491:
The `acsch(x)` implementation here doesn't trouble us though and works perfectly fine as confirmed with other CAS , cause the branch cuts are along the imaginary axis ! , hence once I'll address the above pointed out thoughts https://github.com/sympy/sympy/pull/22521#discussion_r797295998
i'll maybe create a new PR for the` acsch` one and we could continue our discussion here for how to handle the branch cuts for `asech` along real axis !

sympy/functions/elementary/hyperbolic.py
anutosh491: Pic from wolfram for reference 
image

anutosh491: > It seems that this should be something like `return log(2) - log(c) - e*log(x)` when `arg` is `c*x**e`. If `log(2)` is not included, then the leading term of `log(x) + asech(x)` will not be correct. `log(c*x**e)` should be written in the form `log(c) + e*log(x)` so that the `log(x)` terms coming from different parts of a sum would combine without explicit simplifications.

I think these should be the answers for the following
```
>>> asech(x).as_leading_term(x)
log(2/x)
>>> asech(2*x).as_leading_term(x)
log(1/x)
>>> (2*asech(x)).as_leading_term(x)
2*log(2/x)
>>> ( log(x) + asech(x)).as_leading_term(x)
log(2/x) + log(x)
```
Returning this will give us correct answer for limits !

anutosh491: ping @0sidharth

anutosh491: Ping @jksuom

anutosh491: ping @jksuom @0sidharth

sympy/functions/elementary/hyperbolic.py
```diff
@@ -681,13 +681,17 @@ def _eval_rewrite_as_coth(self, arg, **kwargs):
         return 1/coth(arg)
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
-        from sympy.series.order import Order
         arg = self.args[0].as_leading_term(x)
+        arg0 = arg.subs(x, 0)
 
-        if x in arg.free_symbols and Order(1, x).contains(arg):
+        if arg0 is S.NaN:
+            arg0 = arg.limit(x, 0, dir='-' if cdir.is_negative else '+')
```
jksuom, line 688: Can cdir be - 1?

anutosh491: I took inspiration for this from the `sinh` function but I don't think we have a test yet where we encounter a case where `arg0` is `S.NaN`, so not too sure on this but if you have anything in mind let me know , i'll check it out ! I'll try to think when could we encounter such case till then !

jksuom: > the suggestion you gave me

Do you mean the suggestion to use an improved version of `log`?

anutosh491: I'm actually a bit confused about this comment now , can we do all log operations at all places( not that I know any restrictions) but I came across some results which make me oppose this simplifcation here !
For eg
```
>>> asech(x).rewrite(log)
log(sqrt(-1 + 1/x)*sqrt(1 + 1/x) + 1/x)

>>> expr = log(x*sqrt(-1 + 1/x)*sqrt(1 + 1/x) + 1) # basically simplying log(x) + asech(x)

>>> limit(asech(x).rewrite(log) + log(x), x, 0,'-')
log(2) + 2*I*pi 

>>> limit(expr, x, 0, '-')
log(2) # This answer opposes the one from wolfram and few other CAS(sage for example)
```

anutosh491: I'll paste a set of results which I feel are the correct ones (after confirming with wolfram and other CAS)
Ignore the rewrite required in these, all results are correct but in sympy we could only get the correct results as of now on master using the rewrite ! But other than that wolfram agrees to all these answers !
```
>>> limit(log(x) + asech(x).rewrite(log), x,0, '-')
log(2) + 2*I*pi
>>> limit(log(x) + asech(x).rewrite(log), x,0, '+')
log(2)
>>> limit(-log(x) + asech(x).rewrite(log), x,0, '-')
oo
>>> limit(-log(x) + asech(x).rewrite(log), x,0, '+')
oo
>>>
>>>
>>> limit(log(x) + acsch(x).rewrite(log), x,0, '-')
-oo
>>> limit(log(x) + acsch(x).rewrite(log), x,0, '+')
log(2)
>>> limit(-log(x) + acsch(x).rewrite(log), x,0, '-')
-log(2) - I*pi
>>> limit(-log(x) + acsch(x).rewrite(log), x,0, '+')
oo
```

The best part is my first commit was giving all answers perfectly and after trying to consider the suggestion you pointed above now this is what we get
```
>>> limit(log(x) + asech(x), x, 0, '-')
log(2)
>>> limit(log(x) + asech(x), x, 0, '+')
log(2)
>>> limit(-log(x) + asech(x), x, 0, '-')
oo
>>> limit(-log(x) + asech(x), x, 0, '+')
oo
>>> limit(log(x) + acsch(x), x, 0, '-')
log(2)
>>> limit(log(x) + acsch(x), x, 0, '+')
log(2)
>>> limit(-log(x) + acsch(x), x, 0, '-')
-log(2) - I*pi
>>> limit(-log(x) + acsch(x), x, 0, '+')
oo

```
Some tests I had added in the first commit, which agree with all what wolfram says and the rewrite ones !
```
    assert limit(log(x)+asech(x),x,0, dir = '+') == log(2)
    assert limit(log(x)+asech(x),x,0, dir = '-') == log(2) + 2*I*pi
    assert limit(log(x)+acsch(x),x,0, dir = '+') == log(2)
    assert limit(log(x)+acsch(x),x,0, dir = '-') == -oo
```

anutosh491: >It seems that this should be something like return log(2) - log(c) - e*log(x) when arg is c*x**e. If log(2) is not included, then the leading term of log(x) + asech(x) will not be correct. log(c*x**e) should be written in the form log(c) + e*log(x) so that the log(x) terms coming from different parts of a sum would combine without explicit simplifications.

I feel these should be the correct results 
```

>>> asech(x).as_leading_term(x)
log(2/x)
>>> asech(2*x).as_leading_term(x)
log(1/x)
>>> (2*asech(x)).as_leading_term(x)
2*log(2/x)
>>> ( log(x) + asech(x)).as_leading_term(x)
log(2/x) + log(x)  # If we cancel the x here , the essence of the imaginary part of the limit is being lost for some reason
```
Returning this will give us correct answer for limits !

anutosh491: Once I have some clarity on this , I plan to do the following !
The `acsch(x)` implementation here doesn't trouble us though and works perfectly fine as confirmed with other CAS , cause the branch cuts are along the imaginary axis ! , hence once we've address the above pointed out thoughts 
i'll maybe create a new PR for the `acsch` one and we could continue our discussion here for how to handle the branch cuts for `asech` along real axis !

Reference from wolfram 
image
image

jksuom: It is possible that the two logarithms cannot be combined as in https://github.com/sympy/sympy/pull/22521#discussion_r753669192. That would explain the extra `2*pi`.

anutosh491: Yeah I agree , I haven't given the log based approach your mentioning a fill fledged attempt ,will give it a shot in a day or two and see all areas where we could improve !

jksuom: The pole error, or at least some of them, should be removed and some log term returned instead.

anutosh491: True true will try to approach those too ,I realized if we do some basic log operations as we had done before ( which lead to missing of the 2 pi term) in that case the as leading term works but in the purest form it won't! 
Will give it a shot by a day or two !

anutosh491: Actually I realize that the pr is good to go by current standard(where I see that `_eval_as_leading_term`  which are already implemented for functions eg `asinh, acosh`  don't really work at branch cuts) but I wouldn't want that otherwise the essence of the discussion would be lost !
Check out this issue #22986
Here we have the following 
```
>>> limit(acosh(1 + 1/x)*sqrt(x), x, oo) # this comes from _eval_as_leading_term of acosh function
0                                       # acosh(x + 1).as_leading_term(x) is sqrt(2)*sqrt(x) but returns 0
>>> limit(acosh(1 + 1/x).rewrite(log)*sqrt(x), x, oo)
sqrt(2)
>>> gruntz(acosh(1 + 1/x)*sqrt(x), x, oo)
sympy.core.function.PoleError: Cannot expand acosh(_w + 1) around 0
>>> gruntz(acosh(1 + 1/x).rewrite(log)*sqrt(x), x, oo)
sqrt(2)
```
So basically this tells us that `acosh(x).series(x, 1, 6)` won't work ! Hence what changes are introduced in log need to cater to all functions also those which have been implemented . Obviously a basic fix for this issue is adding this block 

```
    def _eval_nseries(self, x, n, logx):
        x0 = self.args[0].limit(x, 0)
        if x0 == 1:
            return self._eval_rewrite_as_log(self.args[0])._eval_nseries(x, n, logx)
        else:
            return super()._eval_nseries(x, n, logx)
```
I had given a thought on the reported issue few days back and obviously this was solv

anutosh491: **Will this solution work for the reported issue ?**
I ask this because 
**1)** The `_eval_nseries` implementation makes gruntz work ! So now we have
```
>>> gruntz(acosh(1 + 1/x)*sqrt(x), x, oo)
sqrt(2)
```
**2)** The `_eval_as_leading_term` improvement makes the leading term work and then as `leadterm` won't be independent of `x`  , we could return the working solution through gruntz
```
>>> acosh(1 + x).as_leading_term(x)
sqrt(2)*sqrt(x)
>>> 
>>> (acosh(1 + x)*sqrt(1/x)).as_leading_term(x)
sqrt(2)*sqrt(x)*sqrt(1/x)
>>> 
>>> (acosh(1 + x)*sqrt(1/x)).leadterm(x)
cannot compute leadterm(sqrt(1/x)*acosh(x + 1), x). The coefficient
should have been free of x but got sqrt(2)*sqrt(x)*sqrt(1/x)
```
Hence this addresses the issue along with giving us series at branch cuts` (-1, 1)`
```
>>> acosh(1 + x).series(x)
sqrt(2)*sqrt(x) - sqrt(2)*x**(3/2)/12 + 3*sqrt(2)*x**(5/2)/160 - 5*sqrt(2)*x**(7/2)/896 + 35*sqrt(2)*x**(9/2)/18432 - 63*sqrt(2)*x**(11/2)/90112 + O(x**6)
```
I don't see anything wrong or suspicious with the solution reported above ! Hence if you approve of this , I could implement this for one function atleast and `acosh` seems to be a good place to start , cause it has branch cuts along the real axis

jksuom: Does this also work with acosh(-1 + x)?

anutosh491: Yes it does , I've checked related cases but just wanted some confidence behind the solution ! Thought of discussing it here as it's somewhat related !

jksuom: What is the result?

anutosh491:
```
>>> acosh(-1 + x).series(x)
I*pi - sqrt(2)*I*sqrt(x) - sqrt(2)*I*x**(3/2)/12 - 3*sqrt(2)*I*x**(5/2)/160 - 5*sqrt(2)*I*x**(7/2)/896 - 35*sqrt(2)*I*x**(9/2)/18432 - 63*sqrt(2)*I*x**(11/2)/90112 + O(x**6)
>>> 
>>> acosh(1 + x).series(x)
sqrt(2)*sqrt(x) - sqrt(2)*x**(3/2)/12 + 3*sqrt(2)*x**(5/2)/160 - 5*sqrt(2)*x**(7/2)/896 + 35*sqrt(2)*x**(9/2)/18432 - 63*sqrt(2)*x**(11/2)/90112 + O(x**6)
```
Result seem good when compared with wolfram !

jksuom: The expansion of acosh(-1 + x) has a branch cut for x < 0 where the function is single-valued.

anutosh491: I agree that `acosh(x -1 )` has a branch cut from `(-oo, 0)` and `(0, 2)` but it's single valued after 2 right ? 
image
[Check out this wolfram link for reference](https://www.wolframalpha.com/input?i=limit+calculator&assumption=%7B%22F%22%2C+%22Limit%22%2C+%22limit%22%7D+-%3E%220%22&assumption=%7B%22F%22%2C+%22Limit%22%2C+%22limitfunction%22%7D+-%3E%22acosh%28x+-1%29%22) , cause we can see limit calculations supporting the point too .
We also have the correct result while calculating series at 2
```
>>> acosh(x - 1).series(x, 2)
sqrt(2)*sqrt(x - 2) - sqrt(2)*(x - 2)**(3/2)/12 + 3*sqrt(2)*(x - 2)**(5/2)/160 - 5*sqrt(2)*(x - 2)**(7/2)/896 + 35*sqrt(2)*(x - 2)**(9/2)/18432 - 63*sqrt(2)*(x - 2)**(11/2)/90112 + O((x - 2)**6, (x, 2))
```

anutosh491: Also we know that `acosh(z) = ±i acos(z)`, where the sign depends on the location of `z` in the complex plane !
Hence the results seem quite relevant , don't you think so ?
```
>>> acos(x - 1).series(x)
pi - sqrt(2)*sqrt(x) - sqrt(2)*x**(3/2)/12 - 3*sqrt(2)*x**(5/2)/160 - 5*sqrt(2)*x**(7/2)/896 - 35*sqrt(2)*x**(9/2)/18432 - 63*sqrt(2)*x**(11/2)/90112 + O(x**6)
>>> acosh(x - 1).series(x)
I*pi - sqrt(2)*I*sqrt(x) - sqrt(2)*I*x**(3/2)/12 - 3*sqrt(2)*I*x**(5/2)/160 - 5*sqrt(2)*I*x**(7/2)/896 - 35*sqrt(2)*I*x**(9/2)/18432 - 63*sqrt(2)*I*x**(11/2)/90112 + O(x**6)
>>>
>>> 
>>> acos(x +1).series(x)
sqrt(2)*I*sqrt(x) - sqrt(2)*I*x**(3/2)/12 + 3*sqrt(2)*I*x**(5/2)/160 - 5*sqrt(2)*I*x**(7/2)/896 + 35*sqrt(2)*I*x**(9/2)/18432 - 63*sqrt(2)*I*x**(11/2)/90112 + O(x**6)
>>> acosh(x +1).series(x)
sqrt(2)*sqrt(x) - sqrt(2)*x**(3/2)/12 + 3*sqrt(2)*x**(5/2)/160 - 5*sqrt(2)*x**(7/2)/896 + 35*sqrt(2)*x**(9/2)/18432 - 63*sqrt(2)*x**(11/2)/90112 + O(x**6)
```

anutosh491: Does this explanation convince you or am I missing something here ? I guess we could try this out !

anutosh491: cc @jksuom

jksuom:
The implementations of `_eval_as_leading_term` of inverse functions are very old. They should be completely rewritten. For example, starting with `arg = self.args[0].as_leading_term(x)` is not efficient, something like

    arg = self.args[0]
    arg0 = arg.subs(x, 0)

should suffice in most cases to find out if a branch point is involved.

anutosh491:
> The implementations of `_eval_as_leading_term` of inverse functions are very old. They should be completely rewritten. For example, starting with `arg = self.args[0].as_leading_term(x)` is not efficient.

True , I agree will try this out soon ! Dealing with couple other prs related to series as of now which are close to go in !Will look into this soon !

anutosh491:
Hello @jksuom . I had a doubt regarding using log based rewrites . As I plan to improve **leading term** and **series** method for **log** and then use log based rewrites for handling series for **inverse trig/hyperbolic functions** at **branch cuts** as a part of my gsoc proposal , I had a doubt which I am not sure could be answered through the current cdir/direction system .

After I improved the method for log , I was able to answer limits, series for all points other than points like `I*oo` (though the limits can be answered here through heuristics here so even that's not a concern, will be shown in proposal)
  
Consider the following example
```
>>> asinh(x).rewrite(log)
log(x + sqrt(x**2 + 1))
>>> _.series(x, I*oo)
  File "C:\Users\anuto\sympy\sympy\sympy\core\power.py", line 1723, in _eval_nseries
    raise NotImplementedError()
NotImplementedError
```
First of all the series call is transformed to `log(_x + sqrt((_x + oo*I)**2 + 1) + oo*I).series(x)` , which is not correct as this should be dealt as what we do with `oo` , which is an inversion based transformation . So the following block from `expr.py` should be improved first of all 
```
        if x0 in [S.Infinity, S.NegativeInfinity]:    # this needs to be improved obviously 
            from .function import PoleError
            try:
                sgn = 1 if x0 is S.Infinity else -1
                s = self.subs(x, sgn/x).series(x, n=n, dir='+', cdir=cdir)
                if n is None:
                    return (si.subs(x, sgn/x) for si in s)
                return s.subs(x, sgn/x)
            except PoleError:
                s = self.subs(x, sgn*x).aseries(x, n=n)
                return s.subs(x, sgn*x)
```
Consider for the time being , we go with `if abs(x0) in [S.Infinity, S.NegativeInfinity]` then we get the oo sort of transformation here then we would land upon the similar answer to what we would get from `asinh(x).rewrite(log).series(x, oo)` .
This is because 
```
>>> # asinh(x).rewrite(log).series(x, oo)   and   asinh(x).rewrite(log).series(x, I*oo)
>> # would end up in log._eval_nseries() 
>>> # and call log(sqrt(1 + _x**(-2)) + 1/_x).series(x)
>>> # and return same results 
>>> -3/(32*x**4) + 1/(4*x**2) + log(2) - log(1/x) + O(x**(-6), (x, oo))
```

But series at `oo` and `I*oo` would differ  by a constant factor of `I*pi/2` , rest everything is good .
image
image

How do we create this distinction of ` I*pi/2` ?
I currently don't see a clear method to approach this . Other than this I have clear proposed implementations for everything I plan to tackle , hence would be great if you could go guide me here .

jksuom:
I would first implement `log. _eval_as_leading_term` at the branch points, especially 0 and `-oo`.

anutosh491:
> 

I now have a basic implementation for this and the code provided in our discussion on the other issue works flawlessly and fixes all wrong results based on branch cuts from both directions . 
Though I am still doubtful regarding handling of this case  `asinh(x).rewrite(log).series(x, I*oo)` . We anyways use transformation through the series code and transform oo cases to 0+ cases and use _eval_nseries() methods . But can the same be done for this ??  Obviously we need to do a inversion based transformation only (`sub(z, 1/z)`) rather than an additive one (`subs(z , z+ z0)` or `subs(z , z + I*oo)`) .

What I feel is do the same thing as oo but then cdir can't be 1 , it needs to be something else to take decisions on and add the `I*pi/2` term because the series is completely same except the constant term
image

Issue #22529, opened 2021-11-19
aim for "least surprise" substitutions


smichr:
In #22052 was raised the following behavior of algebraic subs:
```python
>>> x = Symbol('x', real=True)
>>> [(x**(i*S.Half)).subs(x**2,y) for i in range(2, 9)]
[x, y**(3/4), y, y**(5/4), x**3, y**(7/4), y**2]
```
vs
```python
>>> x = Symbol('x', positive=1)
>>> [(x**(i*S.Half)).subs(x**2,y) for i in range(2, 9)]
[sqrt(y), y**(3/4), y, y**(5/4), y**(3/2), y**(7/4), y**2]
```
In the first set, given that `x**3` remains unchanged, I think it surprising that x**(3/2) is changed.

Do we really need algebraic subs like this instead of having the user specify `subs(x, sqrt(y))`? The latter is totally unambiguous.
```

**Comment by ThePauliPrinciple** [comment]
If you indicate that x is a complex, the result is
```python
>>>x = Symbol('x', complex= True)
>>>Eq(x**Rational(5,2),(x**2)**Rational(5,4)).subs(x,I)
Eq(I**(5/2), -(-1)**(1/4))
```
undertermined. 

I think such substitutions with non integer powers of negative/imaginary numbers are generally not allowed. I think the underlying problem is the same as the "standard" trick to "prove" that any number equals any other number (which we know is not true).

smichr:
> undertermined.

It should be allowed. If you evaluate the lhs and rhs of the Eq, they are the same.



> I think such substitutions with non integer powers of negative/imaginary numbers are generally not allowed

Did you find a case that is wrong?

smichr:
I think this is working in master, now.

PR #22534, opened 2021-11-20
use Pow.combines to limit subs

smichr:

fizes #22529 which gives examples of subs doing algebraic substitutions that are not always valid

It also does not break anything that was committed in #22052.

The conditions for `(b**e)**x -> b**(e*x)` were, near SymPy's beginning, spread out and applied inconsistently in the code base. Now they are all localized in the logic of the `Pow._eval_power` routine (or should be). 

But in the code base one might want to know if a combination should be made or not. The canonical way to do that now is to see if `self.combines(x) == True` where `self = b**e`.

#### Other comments

- [x] add tests from #22529 to demonstrate that less surprising (and correct) algebraic substitutions are taking place.

#### Release Notes

* functions
    * Pow and ExpBase now have `combines` method that should return True if self's exponent can combine with other

### DIFF
--- a/sympy/core/power.py
+++ b/sympy/core/power.py
@@ -823,32 +823,14 @@ class Pow(Expr):
         return self.base.is_polar
 
     def combines(self, x):
-        """return +/-1 if ``(b**e)**x == +/-b**(e*x)`` else None"""
+        """return True if ``(b**e)**x == b**(e*x)`` else False"""
         try:
             as_int(x, strict=False)
-            return S.One
+            return True
         except ValueError:
             b, e = self.as_base_exp()
-            p = Pow._eval_power(
-                Pow(b, e, evaluate=False), x)
-            if p is None:
-                return
-            c = S.One
-            # we might get Pow or -Pow
-            if p.is_Mul:
-                if p.args[0] == -1 and len(p.args) == 2:
-                    p = p.args[1]
-                    c = -c
-                else:
-                    return
-            # trickier input involving numbers could collapse;
-            # leave that out unless it is needed, e.g.
-            # Pow(1, x)**y -> 1 or nan and the result will
-            # match the result of 1**(x*y) unless x,y=0,oo
-            # >>> (1*oo)**0,1**(0*oo)
-            # (1, nan)
-            if p.as_base_exp()[0] == b:
-                return c
+            # These conditions ensure that (b**e)**f == b**(e*f) for any f
+            return b.is_positive and e.is_real or b.is_nonnegative and e.is_nonnegative
 
     def _eval_subs(self, old, new):
         from sympy.calculus.accumulationbounds import AccumBounds
@@ -885,7 +867,7 @@ class Pow(Expr):
                 if old.is_commutative:
                     # Allow fractional powers for commutative objects
                     pow = coeff1/coeff2
-                    return bool(old.combines(pow)), pow, None
+                    return old.combines(pow), pow, None
                 else:
                     # With noncommutative symbols, substitute only integer powers
                     if not isinstance(terms1, tuple):
--- a/sympy/core/power.py
+++ b/sympy/core/power.py
@@ -822,6 +822,34 @@ class Pow(Expr):
     def _eval_is_polar(self):
         return self.base.is_polar
 
+    def combines(self, x):
+        """return +/-1 if ``(b**e)**x == +/-b**(e*x)`` else None"""
+        try:
+            as_int(x, strict=False)
+            return S.One
+        except ValueError:
+            b, e = self.as_base_exp()
+            p = Pow._eval_power(
+                Pow(b, e, evaluate=False), x)
+            if p is None:
+                return
+            c = S.One
+            # we might get Pow or -Pow
+            if p.is_Mul:
+                if p.args[0] == -1 and len(p.args) == 2:
+                    p = p.args[1]
+                    c = -c
+                else:
+                    return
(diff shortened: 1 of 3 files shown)


sympy/core/power.py
```diff
@@ -822,6 +822,19 @@ def _eval_is_composite(self):
     def _eval_is_polar(self):
         return self.base.is_polar
 
+    def combines(self, exp):
+        """return True if ``(b**e)**x == b**(e*x)`` else False"""
+        try:
+            as_int(pow, strict=False)
+            combines = True
```
smichr, line 829:
```suggestion
            return True
```

sympy/functions/elementary/exponential.py
```diff
@@ -115,11 +116,18 @@ def _eval_is_rational(self):
     def _eval_is_zero(self):
         return self.exp is S.NegativeInfinity
 
+    def combines(self, pow):
+        # since the base is positive E, it combines if
+        # other is int or E's power is real
+        if as_int(pow, strict=False):
+            return True
+        return self.args[0].is_real
```
smichr, line 124:
```suggestion
        try:
            as_int(pow, strict=False):
            return True
        except ValueError:
            return self.args[0].is_real
```

sympy/functions/elementary/exponential.py
```diff
@@ -115,11 +116,20 @@ def _eval_is_rational(self):
     def _eval_is_zero(self):
         return self.exp is S.NegativeInfinity
 
+    def combines(self, pow):
+        # since the base is positive E, it combines if
+        # other is int or E's power is real
+        try:
+            as_int(pow, strict=False):
```
smichr, line 123:
```suggestion
            as_int(pow, strict=False)
```

sympy/functions/elementary/exponential.py
```diff
@@ -115,11 +116,20 @@ def _eval_is_rational(self):
...
         return self.exp is S.NegativeInfinity
 
+    def combines(self, pow):
+        # since the base is positive E, it combines if
+        # other is int or E's power is real
+        try:
+            as_int(pow, strict=False)
+            return True
+        except ValueError:
+            return self.args[0].is_real
```
smichr, line 126:
```suggestion
        try:
            as_int(pow, strict=False)
            return True
        except ValueError:
            # treat like Pow
            b, e = self.as_base_exp()
            return Pow(b, e, evaluate=False).combines(pow)
```

sympy/core/power.py
```diff
@@ -822,6 +822,19 @@ def _eval_is_composite(self):
...
+        try:
+            as_int(pow, strict=False)
+            return True
+        except ValueError:
+            b, e = self.as_base_exp()
+            p = Pow._eval_power(
+                Pow(b, e, evaluate=False), exp)
+            if p is None:
+                return False
+            return p.as_base_exp()[0] == b
```
smichr, line 836:
```suggestion
            # we might get Pow or -Pow
            if p.is_Mul:
                assert len(p.args) == 2 and p.args[0] == -1, 'unanticipated return value'
                p = p.args[1]
            # could it have evaluated to non-Pow by combining exponent?
            return not p.is_Pow or p.as_base_exp()[0] == b
```

sympy/core/power.py
```diff
@@ -822,6 +822,24 @@ def _eval_is_composite(self):
     def _eval_is_polar(self):
         return self.base.is_polar
 
+    def combines(self, exp):
+        """return True if ``(b**e)**x == b**(e*x)`` else False"""
+        try:
+            as_int(pow, strict=False)
```
oscargus, line 828:
```suggestion
            as_int(exp, strict=False)
```

smichr: Thanks...I'll add it in the squash

sympy/core/power.py
```diff
@@ -822,6 +822,34 @@ def _eval_is_composite(self):
...
 
+    def combines(self, x):
+        """return +/-1 if ``(b**e)**x == +/-b**(e*x)`` else None"""
+        try:
+            as_int(x, strict=False)
+            return S.One
+        except ValueError:
+            b, e = self.as_base_exp()
+            p = Pow._eval_power(
+                Pow(b, e, evaluate=False), x)
```
oscarbenjamin, line 833: In #22052 I deliberately changed this so that it did not depend on `_eval_power` but rather a principled check based on the available assumptions. Is it not possible to do the same here?

smichr: `_eval_power` *is* the principled check of conditions. It was my goal (long ago) to remove all those decisions from everywhere in the code base and localize them there.

I experimented with pulling out the code from there and putting it in `combines` and then having `_eval_power` ask `combines` if the exponents should be joined, but the results of `_eval_power` are a bit more general than the notion of `combines` so it didn't look that clean. So I decided to make `combines` query the result of `_eval_power`.

smichr:
@oscarbenjamin , it could just be that I'm nervous without cause. Perhaps the other rules in `Pow._evalf_power` don't apply to the case where subs will yield `b**(e*other)` and the cases that don't are already handled elsewhere.

To use the `combines` idea that way, `exp._eval_evalf` checks for the specific simple case where the result is strictly `exp(e*other)` and also allows for more general combining by deferring to `Pow._eval_power` if the simple conditions do not apply. (And tests fail without that.)

smichr:
I think this is not actually changing any behavior in current master other than putting the logic for combining in a function with that name. So  this looks like it is not needed.

PR #22536, opened 2021-11-20
calculus: AccumulationBounds is now an Expr

ThePauliPrinciple:
@oscarbenjamin 

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/calculus/accumulationbounds.py
+++ b/sympy/calculus/accumulationbounds.py
@@ -1,6 +1,6 @@
 from sympy.core import Add, Mul, Pow, S
 from sympy.core.basic import Basic
-from sympy.core.expr import AtomicExpr, Expr
+from sympy.core.expr import Expr
 from sympy.core.numbers import _sympifyit, oo, zoo
 from sympy.core.relational import is_le, is_lt, is_ge, is_gt
 from sympy.core.sympify import _sympify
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
The test fails ultimately because `is_number` has changed. The PR gives:
```python
In [1]: AccumBounds(-S(3)/2, S(1)/2).is_number
Out[1]: True
```
However master gives:
```python
In [1]: AccumBounds(-S(3)/2, S(1)/2).is_number
Out[1]: False
```
The correct answer is `False` because `is_number=True` means that the expression can be numerically evaluated to a complex number in floating point but an AccumBounds represents an interval rather than a complex number.

That change is because `AtomicExpr` sets `is_number` to `False` as a class attribute:
https://github.com/sympy/sympy/blob/ef854dad43c9cef361857782dd0b8523e43f2430/sympy/core/expr.py#L3920-L3927

So `AccumBounds` just needs to set that as well. Also all subclasses of `AtomicExpr` should be reviewed to see whether this value of `False` is correct or whether this class attribute should actually just be removed from `AtomicExpr` because it doesn't really make sense to set `is_number=False` for all `AtomicExpr` (although it does make sense to set it for `AccumBounds`).

ThePauliPrinciple:
The class also defines: `is_extended_real = True`, does that make sense if `is_number = False`?

smichr:
I feel like `AccumBounds` is like sneaking SetExpr behavior into Expr. But perhaps this is just how it has to be...like special casing Order, too, in Add.flatten.

oscarbenjamin:
> The class also defines: `is_extended_real = True`, does that make sense if `is_number = False`?

All the numbers contained by an `AccumBounds` are extended real but it is not itself an extended real. I'm not entirely sure how that should be handled but it is probably best to leave it unchanged for now.

oscarbenjamin:
Looks good. Thanks!

PR #22539, opened 2021-11-22
has_free defined and used to solve issues with non-Symbol generators

### LINKED ISSUE #20809: Basic: a has_free method method is needed

### LINKED ISSUE #22527: integrate doesn't seem to handle Functions properly
I get different results from integrating ``f(z(t))`` as compared to ``f(t)``.
```python
t, R = symbols(r't R')
z = Function('z')(t)

def f(x):
  return x/sqrt(R**2 - x**2)

Uz = integrate(f(z), z)
Fz = diff(Uz, z)

Ut = integrate(f(t), t)
Ft = diff(Ut, t)
```

```python
[f(z), Uz, Fz]
```
image

```python
[f(t), Ut, Ft]
```
image

The Function `z` is integrated wrongly, but the symbol `t` is integrated correctly.

### LINKED ISSUE #22583: is_polynomial right for wrong reasons (and sometimes wrong)
```python
>>> (f(x) + 1).is_polynomial(f(x))
True
```
That is right for the same reason that `(x + 1).is_polynomial(y) -> True`: because the expression is perceived as being independent of the "symbol" provided. This can be seen by considering a case known to be false:
```python
>>> (1/x+1).is_polynomial(x)
False
>>> (1/f(x)+1).is_polynomial(f(x))
True
```
This is addressed by #22539

### LINKED ISSUE #22624: Wrong indefinite integration when variable is element of MatrixSymbol
The integral 1/x is wrongly evaluated by sympy when x is a `MatrixElement`:
\intx x^{-1})

## Code to reproduce
```python
integrate(1/x, x)  # -> log(x)

a = MatrixSymbol("a", 1, 1)
integrate(1/a[0, 0], a[0, 0]) # -> 1 ≠ log(a[0, 0])
```

However, the definite integral is correctly computed:
```python
integrate(1/x, (x, 1, 2)) # -> log(2)
integrate(1/a[0, 0], (a[0, 0], 1, 2)) # -> log(2)
```

## Workaround
Substitute `a[0, 0]` with some other variable, integrate, and substitute back:
```python
integrate((1/a[0, 0]).subs(a[0, 0], x), x).subs(x, a[0, 0]) # -> log(a[0, 0])
```

## Notes
- 1 is never a good answer to an indefinite integral (as pointed out in the documentation, terms independent of the variable are dropped).
- If the integrand is a polynomial in `a[0, 0]`, the integral is evaluated correctly.

Have a great day!

smichr:

fixes #20809
fixes #22527
fixes #22583
fixes #22624  # no new test needed

* now there is a query to tell if an expression (like a user-defined function) appears as a free object

```python
>>>Integral(f(x), (f(x), 1, 3)).has(f(x))
True
>>>Integral(f(x), (f(x), 1, 3)).has_free(f(x))
False
```
#### Other comments

Internally, modification were made to how `has` works which lead to performance improvements:
```python
>>> from sympy.abc import x, y, z, t
>>> expr = 1+cos(t)
>>> for i in range(10):
...     expr = 1 + expr**2 + expr
...
>>> x10 = symbols('x:10')
>>> x100 = symbols('x:100')
>>> t1=time();ok=expr.has(t);time()-t1
0.0  # same as before
>>> t1=time();ok=expr.has(cos);time()-t1
0.0  # same as before

missing function

>>> t1=time();ok=expr.has(sin);time()-t1
0.017002883911132812 # formerly 0.031

missing symbol(s)

>>> t1=time();ok=expr.has(x);time()-t1
0.009001016616821289  # formerly 0.078
>>> t1=time();ok=expr.has(*x10);time()-t1
0.009001016616821289  # formerly 0.78
>>> t1=time();ok=expr.has(*x100);time()-t1
0.011001110076904297  # formerly 7.75
```
So the biggest change has come for searching for symbols not present in the expression.

The query using `has` can be compared to that using `symbol in expr.free_symbols` which is slower in both the positive and negative case:
```python
>>> t1=time();ok = x in expr.free_symbols;time()-t1
0.032003164291381836
>>> t1=time();ok = t in expr.free_symbols;time()-t1
0.03300309181213379
```
#### Release Notes

* core
    * to query whether a object appears as a non-bound object, use `expr.has_free(obj)`


sympy/core/basic.py
```diff
@@ -124,6 +124,57 @@ def __new__(cls, *args):
...
+
+        This only works for expressions, not types:
+
+        >>> expr.has_free(f)
+        False
+
+        """
+        from sympy.core.symbol import Symbol
+        from sympy.utilities.iterables import iterable, sift
+        masked = self.as_dummy() if not isinstance(self, Symbol) else self
```
oscarbenjamin, line 154: If we need to use `as_dummy` then why isn't this just `return self.as_dummy().has(x)`? Any performance improvements could then be made in `has`.

Or have I misunderstood what this method does?

smichr:
wait for #22541

sympy/core/basic.py
smichr: I made it so for the case of `any` (i.e. `all=False`); but handling for `all=True` is also included.

smichr: Also, the failing docstring is a result of using `has` instead of the previous code. `f` is not an expression -- does not appear as any object's `arg` so the former code that only traversed args worked. The `has` walk is more general. If we want to use `has` then I need a way to identify the generators. With #22267 `has_free` is:
```python
def has_free(self, x):
    return x in generators(self)

def all_free(self, s):
    return not (generators(self) - s)
```
But then we really only need the `generators` since this is a pretty straightforward expression.

sympy/integrals/heurisch.py
```diff
@@ -732,7 +732,7 @@ def find_non_syms(expr):
             return candidate.xreplace(solution).xreplace(
                 dict(zip(poly_coeffs, [S.Zero]*len(poly_coeffs))))
 
-    if not (F.free_symbols - set(V)):
+    if F.has_free(V, all=True):
```
oscarbenjamin, line 735: I don't think this is equivalent. `F.free_symbols - set(V)` is the set of symbols in `F` that are *not* in `V`. Then `not (F.free_symbols - set(V))` means "F has no symbols that are not in V". However `F.has_free(V, all=True)` means that "F has all the symbols that are in V".

This change reverses the sense of the subsets so before it was
```python
if F.free_symbols <= V
```
but the changed code is basically
```python
if V <= F.free_symbols
```

smichr: So this is a case where has can't help. We need to know what are all the free generators are are in F...and we need to know what qualifies as a generator. Another way to look at this would be `F.xreplace(dict(zip (V, (Dummy() for I in V)))).free_symbols & F.free_symbols` is empty.

Having spent many hours on the two approaches, I can say that it is very disconcerting to have to think about the types of generators that one must consider at the algorithm level. I would be MUCH in favor of replacing generators with symbols at the routine outset.

smichr, line 735:
```suggestion
    Fd = F.as_dummy()
    if not Fd.xreplace(dict(zip(V, (Dummy() for _ in V)))).free_symbols & Fd.free_symbols:
```

oscarbenjamin: I wonder what the purpose of the check is in the first place. Maybe @asmeurer would know.

sympy/core/basic.py
```diff
@@ -124,6 +124,56 @@ def __new__(cls, *args):
...
+        sym -= free
+        non = set(non)
+        args = [masked]
+        while args:
+            a = args.pop()
+            if a in non:
+                if all:
+                    non.remove(a)
+                else:
+                    return True
```
smichr, line 173:
```suggestion
        sym = set(sym) - free
        non = set(non)
        args = [masked]
        while args:
            non.discard(args.pop())
```

sympy/core/expr.py
```diff
@@ -2642,13 +2637,17 @@ def is_polynomial(self, *syms):
...
+        if not self.has_free(syms):
             # constant polynomial
             return True
         else:
             return self._eval_is_polynomial(syms)
 
+    def _eval_is_polynomial(self, syms):
+        # classes should return True if self is a
+        # polynomial in all of passed syms
+        return False
```
smichr, line 2649:
```suggestion
        return self._eval_is_polynomial(syms)

    def _eval_is_polynomial(self, syms):
        if not self.has_free(syms):
            # constant polynomial
            return True
        else:
            return False
```

sympy/integrals/tests/test_heurisch.py
```diff
@@ -210,25 +217,28 @@ def test_heurisch_function_derivative():
     assert heurisch(f(x)**2*df, x) == f(x)**3/3
     assert heurisch(df/f(x), x) == log(f(x))
 
+
 def test_heurisch_wrapper():
     f = 1/(y + x)
-    assert heurisch_wrapper(f, x) == log(x + y)
+    assert heurisch_wrapper(f, x) == log(y + x)
```
smichr, line 223:
```suggestion
    assert heurisch_wrapper(f, x) == log(x + y)
```

sympy/integrals/tests/test_heurisch.py
```diff
@@ -210,25 +217,28 @@ def test_heurisch_function_derivative():
...
     f = 1/(y - x)
-    assert heurisch_wrapper(f, x) == -log(x - y)
+    assert heurisch_wrapper(f, x) == -log(y - x)
```
smichr, line 225:
```suggestion
    assert heurisch_wrapper(f, x) == -log(x - y)
```

sympy/integrals/tests/test_heurisch.py
```diff
@@ -210,25 +217,28 @@ def test_heurisch_function_derivative():
...
     f = 1/((y - x)*(y + x))
     assert heurisch_wrapper(f, x) == Piecewise(
-        (-log(x - y)/(2*y) + log(x + y)/(2*y), Ne(y, 0)), (1/x, True))
+        (-log(y - x)/(2*y) + log(x + y)/(2*y), Ne(y, 0)), (1/x, True))
```
smichr, line 228:
```suggestion
        (-log(x - y)/(2*y) + log(x + y)/(2*y), Ne(y, 0)), (1/x, True))
```

sympy/core/basic.py
```diff
@@ -124,6 +124,51 @@ def __new__(cls, *args):
...
+        """
+        from sympy.core.symbol import Symbol
+        from sympy.utilities.iterables import iterable, sift
+        if not iterable(x):
+            x = [x]
+        xset = set(x)
+        # have as_dummy hide all bound objects (which occur in unevaluated
+        # nodes like Integral and are fast to rebuild after putting Dummy
+        # symbols in place) then see if any objects of interest are present
+        masked = self.as_dummy() if not isinstance(self, Symbol) else self
```
smichr, line 162:
```suggestion
        from sympy.utilities.iterables import iterable
        if not iterable(x):
            x = [x]
        xset = set(x)
        # have as_dummy hide all bound objects (which occur in unevaluated
        # nodes like Integral and are fast to rebuild after putting Dummy
        # symbols in place) then see if any objects of interest are present
        masked = self.as_dummy() if self.args else self
```

sympy/core/basic.py
smichr: The problem with generators is that we have to traverse the whole tree to see if they are there, but a `has` query can exit as soon as one is hit. Building `as_dummy` must also traverse the tree, however. Perhaps an improvement would be to have `iter_args_free` methods. The Basic default would be:
```python
def iter_args_free(self):
    for a in self.args:
        yield a
        yield from a.iter_args_free
```
But the routines that have bound symbols would have their own semantics for how to do this. But all the logic is already present in the `as_dummy` construction` so there would be duplication of efforts to do the iterargs.

sympy/integrals/heurisch.py
```diff
@@ -732,7 +732,14 @@ def find_non_syms(expr):
             return candidate.xreplace(solution).xreplace(
                 dict(zip(poly_coeffs, [S.Zero]*len(poly_coeffs))))
 
-    if not (F.free_symbols - set(V)):
+    if all(isinstance(_, Symbol) for _ in V):
+        more_free = F.free_symbols - set(V)
+    else:
+        Fd = F.as_dummy()
+        more_free = Fd.xreplace(dict(zip(V, (Dummy() for _ in ns)))
```
smichr, line 739:
```suggestion
        more_free = Fd.xreplace(dict(zip(V, (Dummy() for _ in V)))
```

smichr:
Where are these benchmark defined? I wonder why the time at 30 is nearly neutral wrt master.
```
0s
        362±0.6μs        358±0.7μs     0.99  solve.TimeRationalSystemSymbol.time_linsolve(1)
      1.25±0.01ms         1.25±0ms     1.00  solve.TimeRationalSystemSymbol.time_linsolve(3)
      2.97±0.01ms      2.95±0.01ms     0.99  solve.TimeRationalSystemSymbol.time_linsolve(5)
      1.02±0.01ms      1.01±0.01ms     0.99  solve.TimeRationalSystemSymbol.time_solve(1)
      3.74±0.03ms      3.74±0.01ms     1.00  solve.TimeRationalSystemSymbol.time_solve(3)
      7.92±0.02ms      7.97±0.05ms     1.01  solve.TimeRationalSystemSymbol.time_solve(5)
      23.4±0.09ms       23.3±0.1ms     1.00  solve.TimeSolve01.time_solve
      23.4±0.09ms       23.3±0.2ms     0.99  solve.TimeSolve01.time_solve_nocheck
              n/a              n/a      n/a  solve.TimeSolveLinSys.time_solve_lin_sys
              n/a              n/a      n/a  solve.TimeSolveLinSys.time_to_expr_eqs
              n/a              n/a      n/a  solve.TimeSolveLinSys.time_verify_sol
              n/a              n/a      n/a  solve.TimeSolveLinSys10x8.time_solve_lin_sys
              n/a              n/a      n/a  solve.TimeSolveLinSys10x8.time_to_expr_eqs
              n/a              n/a      n/a  solve.TimeSolveLinSys10x8.time_verify_sol
              n/a              n/a      n/a  solve.TimeSolveLinSys165x165.time_solve_lin_sys
              n/a              n/a      n/a  solve.TimeSolveLinSys165x165.time_to_expr_eqs
              n/a              n/a      n/a  solve.TimeSolveLinSys165x165.time_verify_sol
      31.0±0.07ms      30.5±0.04ms     0.98  solve.TimeSolveLinSys189x49.time_solve_lin_sys
      34.7±0.06ms      34.2±0.07ms     0.98  solve.TimeSolveLinSys189x49.time_to_expr_eqs
       76.1±0.1ms       76.6±0.3ms     1.01  solve.TimeSolveLinSys189x49.time_verify_sol
      11.0±0.05ms      11.1±0.02ms     1.01  solve.TimeSolveSparsePolySystem.time_solve(1)
       22.0±0.1ms      22.1±0.06ms     1.01  solve.TimeSolveSparsePolySystem.time_solve(2)
      33.1±0.04ms       33.3±0.1ms     1.01  solve.TimeSolveSparsePolySystem.time_solve(3)
         1.02±0ms         1.15±0ms     1.13  solve.TimeSparseSystem.time_linear_eq_to_matrix(10)
+     2.91±0.01ms      14.6±0.03ms     5.01  solve.TimeSparseSystem.time_linear_eq_to_matrix(20)
+     5.82±0.03ms      32.2±0.08ms     5.54  solve.TimeSparseSystem.time_linear_eq_to_matrix(30)
       14.0±0.4ms       14.0±0.3ms     1.00  
```

smichr:
@oscarbenjamin , do you have a sense of which is better: the last commit or the one before it? Other than that, I don't have further changes planned.

sympy/core/basic.py
```diff
@@ -1228,27 +1228,84 @@ def has(self, *patterns):
...
+
+        """
+        from .traversal import iterfreeargs
+        return self._has(iterfreeargs, *patterns)
+
+    def _has(self, iterargs, *patterns):
+        from sympy.utilities.iterables import sift
+        from sympy.core.expr import Expr
+        from sympy.core.function import UndefinedFunction
+        from sympy.core.basic import BasicMeta
```
oscarbenjamin, line 1269: Why do these things need to be special cased?

It should be possible to have a clean design that does not need special cases. It shouldn't be necessary to distinguish Expr or UndefinedFunction, This operates at the Basic level and so there are only three cases to consider:

- Explicit Basic instances
- Basic instances with Wild
- Basic subclasses

smichr: Expressions are node arguments, classes/UndefinedFunction (except in the case of a Tuple holding UndefinedFunctions) are not: traversing args will expose Expr; traversing args and emitting `arg.func` will expose the classes. Expr and UndefinedFunction can be matched with equality whereas others might be subclasses of the classes being sought.

Symbols (not special cased) are located at the tips of branches; no comparison need be done until you get to the no-arg terminus of a branch. Doesn't it make sense to look where you are likely to find something? e.g. there is no point in looking at `arg.func` if you are looking for Expr and vice versa.

oscarbenjamin: An UndefinedFunction is a subclass of Basic so it can be handled just like any other subclass e.g. exp. I don't think that func should be used at all here. I showed efficient and simple ways to handle either expressions (Basic instances) or classes (Basic subclasses) here: https://github.com/sympy/sympy/issues/20809#issuecomment-830783820
The only other thing needed is the fallback slow path for Wild pattern matches.

smichr: The `_has_matcher` does not have to do with Wild patterns -- it is not `match` based; it is separating and expr into commutative and noncommutative parts. That is why I try to avoid that be separating out expressions that *won't* need the matcher. 

I am happy with the performance relative to master.
```
# PR

>>> from sympy.abc import x, y, z, t
>>> expr = 1+cos(t)
>>> for i in range(10):
...     expr = 1 + expr**2 + expr
...
>>> x10 = symbols('x:10')
>>> x100 = symbols('x:100')
>>> t1=time();ok=expr.has(x);time()-t1
0.009001016616821289
>>> t1=time();ok=expr.has(t);time()-t1
0.0
>>> t1=time();ok=expr.has(cos);time()-t1
0.0
>>> t1=time();ok=expr.has(sin);time()-t1
0.017002883911132812
>>> t1=time();ok=expr.has(*x10);time()-t1
0.009001016616821289
>>> t1=time();ok=expr.has(*x100);time()-t1
0.011001110076904297


#master

>>> t1=time();ok=expr.has(x);time()-t1
0.07800793647766113
>>> t1=time();ok=expr.has(t);time()-t1
0.0
>>> t1=time();ok=expr.has(cos);time()-t1
0.0
>>> t1=time();ok=expr.has(sin);time()-t1
0.031003236770629883
>>> t1=time();ok=expr.has(*x10);time()-t1
0.7810781002044678
>>> t1=time();ok=expr.has(*x100);time()-t1
7.705770492553711
```
If you prefer the `has_expression` and `has_type` interface, that has the advantage of making the user decide which path they want and then the same sifting is not needed. But when you are dealing with BasicMeta, the user might not know whether to look for the BasicMeta like `Function('f')` as a literal `Tuple(f, g)` or as an instance `f(x) + g(y)`.

At this point in this effort, it seems to me that the use of `has` for most of the cases where `free_symbols` is used is the better thing to do. There are over 700 uses of `free_symbols` (outside of test assertions) and many of them could probably be recast using `has_free`. Reviewing them would be a good "sprint" project.

I would prefer to commit this as is (per sympy.this). I've battled with the tests to keep benchmarks from increasing. I am happy to yield to any other implementation that improves upon this while doing the same.

smichr:
The reason I am using `i.func` is because a common query is for the simplification check to see if certain classes are present; those can use the faster `i in m` instead of the slower `isinstance(i, j) for j in m`.

sympy/core/basic.py
oscarbenjamin: I think I got confused by the use of the word match in has. I see that it doesn't do Wild patterns.

Let me have a go at writing an alternative version of `has` so we can compare them.

sympy/core/traversal.py
```diff
@@ -5,6 +5,34 @@
 
 
 
+def iterargs(expr):
+    if isinstance(expr, Basic):
```
oscarbenjamin, line 9: Why does this check if the arg is Basic? Surely it should only be used with Basic or otherwise it should use sympify.

Also are these functions public? They don't have leading underscore names but they also don't have docstrings.

sympy/core/basic.py
oscarbenjamin: > I would prefer to commit this as is (per sympy.this "commit now", principle).

To be clear that principle holds no weight and it is in fact an extremely bad idea to have a principle like that in a large software project that is used by many people downstream. We absolutely should think carefully about things before making changes because we always need to be careful about backwards compatibility. There are two sides to backwards compatibility:
1. Trying to avoid making breaking changes now
2. Trying not to create situations where breaking changes would be needed in future

Almost all discussions about "backwards compatibility" in SymPy focus on the first of these rules but the second is actually far more important. When introducing a new public method on Basic or when changing an existing method we need to be very careful to try and ensure that we will never want to change the exact semantics of that method in future. Following this principle all methods should be conservative and the possible set of valid inputs should be clearly defined.

The implementation of `has/has_free` currently in this PR is sufficiently complex (the cryptic variable names don't help) that I am not even sure what its exact semantics are. Below I show a simpler version but that fails some tests: that shows to me that the version in the PR is semantically different so that it is clear to me that it has behaviour that I would not expect even if it isn't clear exactly how they differ.

I would prefer something simpler and clearer like this:
```python
    def _has(self, iterargs, patterns):
        patterns_instance = set()
        patterns_class = []
        patterns_matcher = []
        for p in patterns:
            if isinstance(p, type):
                assert issubclass(p, Basic)
                patterns_class.append(p)
            else:
                p = _sympify(p)
                get_matcher = getattr(p, '_has_matcher', None)
                if get_matcher is not None:
                    patterns_matcher.append(get_matcher())
                else:
                    patterns_instance.add(p)

        if patterns_instance:
            if any(arg in patterns_instance for arg in iterargs(self)):
                return True

        if patterns_class:
            types = tuple(patterns_class)
            types_found = set(map(type, iterargs(self)))
            if any(issubclass(t, types) for t in types_found):
                return True

        if patterns_matcher:
            # This is the slow path (to be avoided where possible):
            for arg in iterargs(self):
                for matcher in patterns_matcher:
                    if matcher(arg):
                        return True

        return False

# Also delete the Basic _has_matcher method
```
I can see that this version fails a couple of tests. This test should be changed:
https://github.com/sympy/sympy/blob/ed0bfd1e47345c8bd45ef542eef18f6a21c55685/sympy/core/tests/test_expr.py#L1129-L1131
Since `f` and `g` are not `Basic` instances `Tuple(f, g)` should really raise an error. In any case it should not be necessary for `has/has_free` to support broken Basic instances like this.

The other test that fails is this one:
https://github.com/sympy/sympy/blob/ed0bfd1e47345c8bd45ef542eef18f6a21c55685/sympy/core/tests/test_power.py#L257-L258
The problem is something to do with `exp(x)` as a function vs `E**x` as a power. That possibly needs some special case handling via `_has_matcher` in `Pow` and/or `exp` but ideally we should avoid `_has_matcher` where possible as it takes us back down the slow path. Probably there is a bug somewhere that causes us to have a `Pow` when we should really have `exp` and that bug should be fixed rather than using `_has_matcher`.

smichr: > Tuple(f, g) should really raise an error

I don't see a good reason to disallow a user to use a SymPy container to hold anything defined by SymPy.

Please have another look at the code. The approach is `has` or `has_free` selects the appropriate traversal iterator and calls `_has` which for Expr will isolate easy expressions to deal with before passing off to the Basic version which does two things: checks for BasicMeta and then uses a matcher which defaults to literal equality testing or (in the case of `operations._has_matcher` uses subset testing and is commutative/noncommutative aware in doing so.

smichr: > We absolutely should think carefully about things before making changes

I made the suggestion because I thought the publicly exposed addition, `has_free`, is needed and it's signature matches that of `has` -- everything else is a detail that we are free to work out. `iterargs` and `iterfreeargs` are light versions of `preorder_traversal` and are straightforward.

smichr: > ideally we should avoid _has_matcher where possible

That's why I sift out the args that don't need a matcher. I don't see a similar guard in the suggestion that you made. And I think that `pattern_instance` will be empty since `Basic._has_matcher` is defined and defaults to an equality tests.

smichr: > problem is something to do with exp(x) as a function vs E**x as a power

```python
>>> from sympy.core.traversal import iterargs
>>> for i in iterargs(Pow(E, x, evaluate=0)):
...  print(i, isinstance(i, exp))
...
E**x True
E False
x False
>>> isinstance(exp(x), exp)
True
```

Your suggestion doesn't identify this because the type of `E**x` is `Pow`, not `exp`. I'm not sure where the magic methods are located that allow `isinstance` to say that the `Pow` with base `E` is to be treated like `exp`.

smichr: > magic methods
It's defined in `__instancecheck`:
```python
class ExpMeta(FunctionClass):
    def __instancecheck__(cls, instance):
        if exp in instance.__class__.__mro__:
            return True
        return isinstance(instance, Pow) and instance.base is S.Exp1
```

smichr:
@oscarbenjamin , I've removed the special handling for Expr and removed the default `_has_matcher` for Basic and just put all the fast equality testing up front. This is much more in line with what you were expecting to find. There's been a lot of thrashing here; please take a fresh look at the current state.

sympy/geometry/line.py
```diff
@@ -1157,7 +1157,7 @@ class Line(LinearEntity):
     """
     def __new__(cls, *args, **kwargs):
         if len(args) == 1 and isinstance(args[0], (Expr, Eq)):
-            missing = uniquely_named_symbol('?', args)
+            missing = uniquely_named_symbol('?', args).name
```
smichr, line 1160:
```suggestion
            missing = uniquely_named_symbol('?', args)
```

sympy/core/expr.py
```diff
@@ -2599,15 +2594,21 @@ def is_polynomial(self, *syms):
...
         >>> (2**x + 1).is_polynomial(x)
         False
-
+        >>> (2**x + 1).is_polynomial(2**x)
+        True
+        >>> f = Function('f')
+        >>> (f(x) + 1).is_polynomial(x)
+        False
+        >>> (f(x) + 1).is_polynomial(f(x))
+        True
```
smichr, line 2611:
```suggestion
        True
        >>> (1/f(x) + 1).is_polynomial(f(x))
        False
```

sympy/core/basic.py
```diff
@@ -1228,31 +1228,90 @@ def has(self, *patterns):
...
+        >>> (x + y + 1).has_free(y + 1)
+        True
+
+        """
+        return self._has(iterfreeargs, *patterns)
+
+    def _has(self, iterargs, *patterns):
+        # separate out types and unhashable objects
+        type_set = set()  # only types
+        p_set = set()  # all patterns so Tuple(f, Add).has(f) will pass
```
oscarbenjamin, line 1266: The comment suggests that `Tuple(f, Add).has(f)` is a supported use case but it shouldn't be.

oscarbenjamin: The `_has` function still seems more complicated than I think it should be. How is this better/different than the version that I proposed?

smichr: > Tuple(f, Add)...shouldn't be [supported]

It would be backward compatible to not support it. If we are going to drop support I think there should be a broader discussion. In python you can store anything in a list; why can't we store any SymPy object in a Tuple? I don't know what a user wants to do with such a Tuple nor what type of object they might want to make that uses it, but I don't see a reason to limit their options by removing this possibility. Maybe they want to make an object that generates random equations so they want to store different operators in the Tuple.

> How is it different [that the one at https://github.com/sympy/sympy/pull/22539#discussion_r758286349]

1. it supports currently supported syntax (`Tuple(f, Add)`)
2. it supports the dual-nature of `E**x` vs `exp(x)` for which `isinstance(E**x, exp)` is True
2. it checks for exact matches before using `has_match` even for things that have a matcher
3. it recognizes objects that are not hashable
5. it ignores non-Basic types (yours raises an error which is probably better)
6. it sympifies only things that are not Basic (so avoids that path with an `isinstance` test)
7. it uses `any(i in set_types for i in iterargs(self))` instead of any(instance(i, types) for i in set(iterargs(self)))`; I'm not sure of the performance difference there, however. I know that `i in tuple` is slower than `i in set`.

oscarbenjamin: > In python you can store anything in a list; why can't we store any SymPy object in a Tuple?

The question is what is a "SymPy object"? The args of Basic should be Basic (instances not subclasses). That is an assumption that is implicit in almost all Basic routines e.g.:
```
In [86]: Tuple(f, Add).subs(x, y)
...
~/current/sympy/sympy.git/sympy/core/basic.py in fallback(self, old, new)
   1053                 if not hasattr(arg, '_eval_subs'):
   1054                     continue
-> 1055                 arg = arg._subs(old, new, **hints)
   1056                 if not _aresame(arg, args[i]):
   1057                     hit = True
...
TypeError: _subs() missing 1 required positional argument: 'new' 
```
There are a few Basic routines that have workarounds to try and support broken subclasses that have non-Basic args but those broken classes will always be unreliable and buggy. Instead of trying to work around these things we should do the opposite: ensure that use of non-Basic args blows up as soon as possible. We should also remove anything from the test suite that gives the impression that such cases should be supported so that contributors are not tempted to add workarounds.

@ThePauliPrinciple has been working (in a range of different PRs) on trying to eliminate all cases of non-Basic args. e.g. #22551

Of course it would be nice if an undefined function was a Basic instance (#4787).

smichr: >  why can't we store any SymPy object in a Tuple

Is this because of the invariant that we are aiming for: `obj.func(*obj.args)`. I know @ThePauliPrinciple has been working on this (as I recall); but special handling is needed for Atom anyway and I'm not sure that special handling shouldn't be done for classes, too, in a detached state.

oscarbenjamin: There are two separate invariants:

1. `obj.func(obj.args)` should recreate `obj` (or evaluate to something equivalent to `obj`)
2. All args are Basic.

It is necessary to assume that all args are Basic because Basic is a very complicated class and has two many methods and attributes for anything to meaningfully duck-type as Basic:
```python
In [2]: len(Basic.__dict__)
Out[2]: 127
```
Wherever we have non-Basic in the args we have bugs. Wherever we have tests that assert that we can have non-Basic in the args it forces contributors into workarounds that degrade the quality, reliability and performance of routines that would be quite simple if the two invariants above could be relied upon.

sympy/core/tests/test_basic.py
```diff
@@ -112,7 +112,11 @@ def test_has():
     assert b21.has(Basic)
     assert not b1.has(b21, b3)
     assert not b21.has()
+    assert not b21.has(type)
     raises(SympifyError, lambda: Symbol("x").has("x"))
+    # if this fails, then it will not be necessary to
+    # test for types by equality in _has
+    assert Tuple(Function('f')).has(Function('f'))
```
smichr, line 119:
```suggestion
    raises(AssertionError, lambda: Tuple(Function('f')))
    from sympy.core.collections import OrderedSet
    raises(AssertionError, lambda: OrderedSet([Function('f')]))
```

sympy/core/basic.py
```diff
@@ -1228,31 +1228,96 @@ def has(self, *patterns):
...
+        for i in other:
+            if not hasattr(i, '_has_matcher'):
+                if any(i == arg for arg in iterargs(self)):
+                    return True
+            else:
+                match = i._has_matcher()
+                if any(match(arg) for arg in iterargs(self)):
+                    return True
+        # now do only non-equality testing for those in p_set
+        for i in p_set
```
smichr, line 1312:
```suggestion
        for i in p_set:
```

sympy/core/tests/test_basic.py
```diff
@@ -112,7 +112,11 @@ def test_has():
     assert b21.has(Basic)
     assert not b1.has(b21, b3)
     assert not b21.has()
+    assert not b21.has(type)
     raises(SympifyError, lambda: Symbol("x").has("x"))
+    raises(AssertionError, lambda: Tuple(Function('f')))
+    from sympy.core.collections import OrderedSet
```
smichr, line 118:
```suggestion
    from sympy.core.containers import OrderedSet
```

sympy/core/containers.py
```diff
@@ -50,6 +50,7 @@ class Tuple(Basic):
     def __new__(cls, *args, **kwargs):
         if kwargs.get('sympify', True):
             args = (sympify(arg) for arg in args)
+            assert all(isinstance(i, Basic) for i in args), 'only Basic args allowed'
```
smichr, line 53:
```suggestion
            # design goal: assert all(isinstance(i, Basic) for i in args), 'only Basic args allowed'
```

sympy/core/containers.py
```diff
@@ -311,6 +312,7 @@ def _sorted_args(self):
 class OrderedSet(MutableSet):
     def __init__(self, iterable=None):
         if iterable:
+            assert all(isinstance(i, Basic) for i in iterable), 'only Basic args allowed'
```
smichr, line 315:
```suggestion
            # design goal: assert all(isinstance(i, Basic) for i in iterable), 'only Basic args allowed'
```

sympy/core/tests/test_basic.py
```diff
@@ -112,7 +112,11 @@ def test_has():
     assert b21.has(Basic)
     assert not b1.has(b21, b3)
     assert not b21.has()
+    assert not b21.has(type)
     raises(SympifyError, lambda: Symbol("x").has("x"))
+    raises(AssertionError, lambda: Tuple(Function('f')))
+    from sympy.core.containers import OrderedSet
+    raises(AssertionError, lambda: OrderedSet([Function('f')]))
```
smichr, line 119:
```suggestion
    assert not b21.has(type)  # XXX should this raise an error?
    raises(SympifyError, lambda: Symbol("x").has("x"))
```

smichr:
Here is an example of storing Functions in an object: `Subs(Derivative(f(f(x)), x), f, cos)`. How can this be handled if we don't allow BasicMeta as args?

oscarbenjamin:
> Here is an example of storing Functions in an object: `Subs(Derivative(f(f(x)), x), f, cos)`. How can this be handled if we don't allow BasicMeta as args?

It can't be handled. Note that such an instance is already broken though:
```python
In [2]: Subs(Derivative(f(f(x)), x), f, cos).subs(y, z)
...
TypeError: _subs() missing 1 required positional argument: 'new'
```
We would need #4787 to make this work.

smichr:
> Note that such an instance is already broken though:

half-full/half-empty:
```python
>>> Subs(Derivative(f(f(x)), x), f, cos).doit() == cos(cos(x)).diff(x)
True
```

oscarbenjamin:
Okay so:
```python
In [2]: e = Subs(Derivative(f(f(x)), x), f, cos)

In [6]: e.simplify()
TypeError

In [9]: e.xreplace({z:y})
TypeError 

In [10]: e.subs({z:y})
TypeError

In [14]: e.nsimplify()
TypeError

In [15]: e.trigsimp()
TypeError

In [16]: e.factor()
TypeError 

In [18]: e.series()
TypeError

In [23]: e.leadterm(z)
TypeError

In [24]: e.integrate(z)
TypeError

In [29]: e.count_ops()
AttributeError

In [31]: e.diff(x)
TypeError 

In [32]: e.expand()
TypeError

In [36]: e.apart()
TypeError

# I could go on...
```
I'd say that the glass is about 99% empty :)

smichr:
> 99% empty

OK, I see how tenuous the system is hanging together given this foundational problem! 

This is what it takes to get `subs` and `xreplace` working (and the methods that fail because of either of them):
```diff
diff --git a/sympy/core/basic.py b/sympy/core/basic.py
index 5075553d18..daffac79f9 100644
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -945,6 +945,25 @@ def subs(self, *args, **kwargs):
                 for i in reversed(redo):
                     sequence.insert(0, sequence.pop(i))
 
+        if kwargs.pop('meta', True):
+            # MetaBasic does not have a tuple of args so special
+            # handling is needed
+            from sympy.core.traversal import bottom_up
+            kwargs['meta'] = False
+            # get meta from self
+            self_meta = set(i for i in iterargs(self) if isinstance(i, BasicMeta))
+            if self_meta:
+                # replace meta with symbols
+                reps = {m: Dummy() for m in self_meta}
+                self = bottom_up(self, lambda x: reps.get(x, x), nonbasic=True)
+                # do the normal subs
+                self = self.subs(sequence, **kwargs)
+                # now replace the dummies with the original meta or a new
+                # meta from the sequence
+                dseq = dict(sequence)
+                u = {v: k for k, v in reps.items()}
+                return bottom_up(self, lambda x: dseq.get(*[u.get(x, x)]*2), atoms=True)
+
         if simultaneous:  # XXX should this be the default for dict subs?
             reps = {}
             rv = self
@@ -1172,7 +1191,13 @@ def _xreplace(self, rule):
             for a in self.args:
                 _xreplace = getattr(a, '_xreplace', None)
                 if _xreplace is not None:
-                    a_xr = _xreplace(rule)
+                    try:
+                        a_xr = _xreplace(rule)
+                    except TypeError:
+                        if a in rule:
+                            changed = True
+                        args.append(rule.get(a, a))
+                        continue
                     args.append(a_xr[0])
                     changed |= a_xr[1]
                 else:
@@ -1540,7 +1565,7 @@ def walk(rv, F):
             """Apply ``F`` to args and then to result.
             """
             args = getattr(rv, 'args', None)
-            if args is not None:
+            if type(args) is tuple:
                 if args:
                     newargs = tuple([walk(a, F) for a in args])
                     if args != newargs:
diff --git a/sympy/core/core.py b/sympy/core/core.py
index 44a1281c8c..a3156fa5a6 100644
--- a/sympy/core/core.py
+++ b/sympy/core/core.py
@@ -102,3 +102,4 @@ def __gt__(cls, other):
         if cls.__cmp__(other) == 1:
             return True
         return False
+
diff --git a/sympy/core/tests/test_basic.py b/sympy/core/tests/test_basic.py
index 0645b091d0..b9357f3d1d 100644
--- a/sympy/core/tests/test_basic.py
+++ b/sympy/core/tests/test_basic.py
@@ -136,6 +136,18 @@ def test_subs():
     assert Symbol("s").subs({"s": 1}) == 1
 
 
+def test_subs_with_MetaBasic():
+    f = Function('f')
+    x, y = symbols('x y')
+    m = Tuple(1, f)
+    assert m.subs(x, y) == m
+    assert m.subs(f, y) == Tuple(1, y)
+    assert m.subs(f, cos) == Tuple(1, cos)
+    assert m.xreplace({x: y}) == m
+    assert m.xreplace({f: y}) == Tuple(1, y)
+    assert m.xreplace({f: cos}) == Tuple(1, cos)
+
+
 def test_subs_with_unicode_symbols():
     expr = Symbol('var1')
     replaced = expr.subs('var1', 'x')
diff --git a/sympy/core/traversal.py b/sympy/core/traversal.py
index 1c5a8ffa47..53cd724c8c 100644
--- a/sympy/core/traversal.py
+++ b/sympy/core/traversal.py
@@ -232,7 +232,7 @@ def bottom_up(rv, F, atoms=False, nonbasic=False):
     if ``nonbasic`` is True, try to apply ``F`` to non-Basic objects.
     """
     args = getattr(rv, 'args', None)
-    if args is not None:
+    if type(args) is tuple:
         if args:
             args = tuple([bottom_up(a, F, atoms, nonbasic) for a in args])
             if args != rv.args:
```
I've removed code to try match non-Basic args (that appear as args) and the tests that would have failed because of that. The changes shown in the diff are not included.

sympy/combinatorics/tests/test_perm_groups.py
```diff
@@ -32,6 +32,13 @@ def test_has():
     assert PermutationGroup(a, b).degree == \
         PermutationGroup(a, b).degree == 6
 
+    # this must be tested by equality since
+    # containment in a set cannot be checked since
+    # g is not hashable
+    g = PermutationGroup(Permutation(0, 2, 1))
+    raises(TypeError, lambda: {g})
+    assert Tuple(1, g).has(g)
```
oscarbenjamin, line 40: I consider this a bug in PermutationGroup arising from abuse of `__eq__`. It is said that in SymPy `a == b` tests structural equality and that works fine with the `__eq__` method that is defined in `Basic`. Practically any time a subclass overrides the `__eq__` method it leads to problems.

oscarbenjamin: This passes all tests in sympy/combinatorics:
```diff
diff --git a/sympy/combinatorics/perm_groups.py b/sympy/combinatorics/perm_groups.py
index c0b91c8..387bc24 100644
--- a/sympy/combinatorics/perm_groups.py
+++ b/sympy/combinatorics/perm_groups.py
@@ -202,7 +202,7 @@ def __contains__(self, i):
     def __len__(self):
         return len(self._generators)
 
-    def __eq__(self, other):
+    def is_equivalent(self, other):
         """Return ``True`` if PermutationGroup generated by elements in the
         group are same i.e they represent the same PermutationGroup.
 
@@ -217,6 +217,8 @@ def __eq__(self, other):
         >>> G.generators == H.generators
         False
         >>> G == H
+        False
+        >>> G.is_equivalent(H)
         True
 
         """
@@ -1786,7 +1788,7 @@ def is_perfect(self):
 
         """
         if self._is_perfect is None:
-            self._is_perfect = self == self.derived_subgroup()
+            self._is_perfect = self.is_equivalent(self.derived_subgroup())
         return self._is_perfect
 
     @property
diff --git a/sympy/combinatorics/tests/test_perm_groups.py b/sympy/combinatorics/tests/test_perm_groups.py
index ff91a33..b496608 100644
--- a/sympy/combinatorics/tests/test_perm_groups.py
+++ b/sympy/combinatorics/tests/test_perm_groups.py
@@ -72,10 +72,12 @@ def test_equality():
     g_1 = PermutationGroup(p_1, p_2)
     g_2 = PermutationGroup(p_3, p_4)
     g_3 = PermutationGroup(p_2, p_1)
+    g_4 = PermutationGroup(p_1, p_2)
 
-    assert g_1 == g_2
+    assert g_1 != g_2
     assert g_1.generators != g_2.generators
-    assert g_1 == g_3
+    assert g_1 != g_3
+    assert g_1 == g_4
 
 
 def test_stabilizer():
```

sympy/core/basic.py
```diff
@@ -1228,31 +1229,88 @@ def has(self, *patterns):
...
+                p_set.add(p)
+            except TypeError:
+                other.append(p)
+
+        for i in iterargs(self):
+            try:
+                if i in p_set:
+                    return True
+            except TypeError:
+                pass  # unhashable i, e.g. PermutationGroup([(0 2 1)])
```
oscarbenjamin, line 1288: Are there other unhashable Basic types besides PermutationGroup?

sympy/integrals/heurisch.py
```diff
@@ -732,7 +732,14 @@ def find_non_syms(expr):
...
 
-    if not (F.free_symbols - set(V)):
+    if all(isinstance(_, Symbol) for _ in V):
+        more_free = F.free_symbols - set(V)
+    else:
+        Fd = F.as_dummy()
+        more_free = Fd.xreplace(dict(zip(V, (Dummy() for _ in V)))
+            ).free_symbols & Fd.free_symbols
+    if not more_free:
+        # all free generators are identified in V
```
oscarbenjamin, line 742: Let me get my head around this. With symbols:
```python
In [22]: F = Integral(x, (x, y, z))

In [23]: V = [x, y]

In [24]: F.free_symbols - set(V)
Out[24]: {z}
```
With functions:
```python
In [25]: F = Integral(f(x), (f(x), g(x), h(x)))

In [26]: V = [f(x), g(x)]

In [27]: Fd = F.as_dummy()

In [28]: Fd.xreplace(dict(zip(V, (Dummy() for _ in V))))
Out[28]: 
  h(x)          
   ⌠            
   ⎮    _0 d(_0)
   ⌡            
Dummy₃₅         

In [29]: Fd.xreplace(dict(zip(V, (Dummy() for _ in V)))).free_symbols
Out[29]: {Dummy₃₇, x}

In [30]: Fd.free_symbols
Out[30]: {x}

In [31]: Fd.xreplace(dict(zip(V, (Dummy() for _ in V)))).free_symbols & Fd.free_symbols
Out[31]: {x}
```
So we want to see if there are any free symbols in Fd after masking off any dummies and anything that is in `V`. I guess that's one way to generalise the original test.

I'm trying to understand what that should mean though. Maybe the premise in using `free_symbols` here is something different.

The difference afterwards is just whether we pass `filter='Q'` to `root_factors`:
https://github.com/sympy/sympy/blob/52f606a503cea5e9588de14150ccb9f7f9ed4752/sympy/integrals/heurisch.py#L643
It looks like `root_factors` passes that to `roots`:
https://github.com/sympy/sympy/blob/52f606a503cea5e9588de14150ccb9f7f9ed4752/sympy/polys/polyroots.py#L1132
From there it looks like `root_factors` applies that to filter out the roots that it is interested in:
https://github.com/sympy/sympy/blob/52f606a503cea5e9588de14150ccb9f7f9ed4752/sympy/polys/polyroots.py#L1063
If we just wanted the rational roots of the polynomial we should be able to compute that more efficiently than having roots compute all the roots and filter them though.

oscarbenjamin: I'm not sure what `root_factors` is supposed to do that is different from `factor`:
```python
In [24]: p = expand((x**2 - 2)*(x-5))

In [25]: p
Out[25]: 
 3      2           
x  - 5⋅x  - 2⋅x + 10

In [26]: factor(p)
Out[26]: 
        ⎛ 2    ⎞
(x - 5)⋅⎝x  - 2⎠

In [27]: factor_list(p)
Out[27]: 
⎛   ⎡            ⎛ 2       ⎞⎤⎞
⎝1, ⎣(x - 5, 1), ⎝x  - 2, 1⎠⎦⎠

In [28]: root_factors(p)
Out[28]: [x - 5, x - √2, x + √2]

In [29]: root_factors(p, filter='Q')
Out[29]: 
⎡        2    ⎤
⎣x - 5, x  - 2⎦
```
With a symbol in the coefficients:
```python
In [30]: p = expand((x**2 - y)*(x-5))

In [31]: p
Out[31]: 
 3      2            
x  - 5⋅x  - x⋅y + 5⋅y

In [32]: factor(p)
Out[32]: 
        ⎛ 2    ⎞
(x - 5)⋅⎝x  - y⎠

In [33]: root_factors(p, x)
Out[33]: [x - 5, x - √y, x + √y]

In [34]: root_factors(p, x, filter='Q')
Out[34]: 
⎡        2    ⎤
⎣x - 5, x  - y⎦
```
Is it possible that this code just predates the factorisation code and should really just use `factor`?

oscarbenjamin: The `Out[33]` above is the only thing that `factor` will not do.

sympy/core/basic.py
smichr: This is the only one that triggered. Why is it not hashable? It derives from Basic.

oscarbenjamin: A subclass that refines `__eq__` must also redefine `__hash__` to be hashable. The `__hash__` method for `PermutationGroup` would probably be quite expensive if it should match the behaviour of `__eq__` (`x == y  -->  hash(x)==hash(y)`). Really it is best not to redefine `__eq__` in `Basic` subclasses and to accept that `a == b` is for structural equality.

sympy/integrals/heurisch.py
oscarbenjamin: I haven't finished with this yet but the following seems to pass all tests:
```diff
diff --git a/sympy/core/numbers.py b/sympy/core/numbers.py
index e87b039..93c07c8 100644
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -2536,7 +2536,7 @@ def __new__(cls, expr, coeffs=None, alias=None, **args):
                 from sympy.polys.polytools import Poly
                 minpoly = Poly(minpoly)
         elif expr.is_AlgebraicNumber:
-            minpoly, root = expr.minpoly, expr.root
+            minpoly, root, coeffs = expr.minpoly, expr.root, expr.coeffs()
         else:
             minpoly, root = minimal_polynomial(
                 expr, args.get('gen'), polys=True), expr
diff --git a/sympy/integrals/heurisch.py b/sympy/integrals/heurisch.py
index 6167936..9d0a85f 100644
--- a/sympy/integrals/heurisch.py
+++ b/sympy/integrals/heurisch.py
@@ -29,7 +29,7 @@
 from sympy.logic.boolalg import And, Or
 from sympy.utilities.iterables import uniq
 
-from sympy.polys import quo, gcd, lcm, factor, cancel, PolynomialError
+from sympy.polys import quo, gcd, lcm, factor_list, cancel, PolynomialError
 from sympy.polys.monomials import itermonomials
 from sympy.polys.polyroots import root_factors
 
@@ -610,37 +610,28 @@ def _exponent(g):
     reducibles = set()
 
     for poly in polys:
-        if poly.has(*V):
-            try:
-                factorization = factor(poly, greedy=True)
-            except PolynomialError:
-                factorization = poly
-
-            if factorization.is_Mul:
-                factors = factorization.args
-            else:
-                factors = (factorization, )
-
-            for fact in factors:
-                if fact.is_Pow:
-                    reducibles.add(fact.base)
-                else:
-                    reducibles.add(fact)
+        coeff, factors = factor_list(poly, *V)
+        reducibles.add(coeff)
+        for fact, mul in factors:
+            reducibles.add(fact)
 
     def _integrate(field=None):
-        irreducibles = set()
         atans = set()
         pairs = set()
 
-        for poly in reducibles:
-            for z in poly.free_symbols:
-                if z in V:
-                    break  # should this be: `irreducibles |= \
-            else:          # set(root_factors(poly, z, filter=field))`
-                continue   # and the line below deleted?
-                           #               |
-                           #               V
-            irreducibles |= set(root_factors(poly, z, filter=field))
+        if field == 'Q':
+            irreducibles = set(reducibles)
+        else:
+            irreducibles = set()
+            for poly in reducibles:
+                for z in poly.free_symbols:
+                    if z in V:
+                        break  # should this be: `irreducibles |= \
+                else:          # set(root_factors(poly, z, filter=field))`
+                    continue   # and the line below deleted?
+                               #               |
+                               #               V
+                irreducibles |= set(root_factors(poly, z, filter=field))
 
         log_part, atan_part = [], []
```

smichr: Feel free to send a PR to this. Note that I modified that loop over `poly.free_symbols`, however.

smichr: So interesting find -- I put a test to see that only 1 `z` was in a `poly` since only one was being processed in master and `root_factors` is supposed to be working with a univariate expression...but that assertion failed.

smichr: Since it was formerly iterating over a set and breaking after the first, this means the routine was not deterministic.

smichr: But the poly's were linear in the cases I checked. I put a new test in place.

smichr: I added the changes.

smichr:
OK, I will be responding less frequently on this for about a week.

smichr:
@oscarbenjamin , I used `equals` instead of `is_equivalent` since that is the prevailing idiom.

sympy/core/basic.py
```diff
@@ -1228,31 +1229,72 @@ def has(self, *patterns):
...
+                except SympifyError:
+                    continue  # Basic won't have this in it
+            p_set.add(p)  # fails if object defines __eq__ but
+                          # doesn't define __hash__
+                                  #
+        for i in iterargs(self):  #
+            if i in p_set:        # <--- here, too
+                return True
+            if type(i) in type_set or any(isinstance(i, t) for t in type_set):
+                return True
```
oscarbenjamin, line 1284: When there are lots of types this can be slow:
```python
In [1]: p = 1 + x**2

In [2]: for _ in range(10):
   ...:     p = p**2 + 1/p
   ...: 

In [3]: types = cos, sin, tan, cosh, sinh, tanh, exp, log, cot, sec, csc, coth, sinh, csch

In [4]: %time p.has(*types)
CPU times: user 76 ms, sys: 0 ns, total: 76 ms
Wall time: 74.8 ms
Out[4]: False
```
It's a bit faster to use a tuple:
```diff
diff --git a/sympy/core/basic.py b/sympy/core/basic.py
index acf28a8..20ac83c 100644
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -1276,11 +1276,12 @@ def _has(self, iterargs, *patterns):
                     continue  # Basic won't have this in it
             p_set.add(p)  # fails if object defines __eq__ but
                           # doesn't define __hash__
+        types = tuple(type_set)
                                   #
         for i in iterargs(self):  #
             if i in p_set:        # <--- here, too
                 return True
-            if type(i) in type_set or any(isinstance(i, t) for t in type_set):
+            if isinstance(i, types):
                 return True
 
         # use matcher if defined, e.g. operations defines
```
That gives:
```python
In [4]: %time p.has(*types)
CPU times: user 52 ms, sys: 0 ns, total: 52 ms
Wall time: 52.8 ms
Out[4]: False
```

sympy/solvers/solveset.py
```diff
@@ -2438,10 +2438,13 @@ def linear_coeffs(eq, *syms, **_kw):
...
     """
+    from sympy.core.traversal import iterfreeargs
     d = defaultdict(list)
     eq = _sympify(eq)
     symset = set(syms)
-    has = eq.free_symbols & symset
+    if len(symset) != len(syms):
+        raise ValueError('duplicate symbols given')
+    has = set(iterfreeargs(eq)) & symset
     if not has:
```
oscarbenjamin, line 2448: Shouldn't this just be `if eq.has_free(*symset)`?

oscarbenjamin:
Okay I think this looks good and it's waited long enough. Let's get it in.

Thanks @smichr!

sympy/solvers/solveset.py
smichr: The set `has` is needed later, so instead of traversing twice, it is collected once.

PR #22540, opened 2021-11-22
Fixed issue with piecewise integration

### LINKED ISSUE #22533: Integral(1/abs(x), x).doit() returns an incomplete answer as output
This was pointed out on the gitter channel a week back or two but I'm not sure if the user remembered to raise the issue on the github page and hence I am raising it before it gets lost in the chat !

On Master
```
>>> x = Symbol('x', real=True)
>>> Integral(1/abs(x), x).doit()
Piecewise((-log(x), x <= 0), (nan, True))
```
The correct answer for this should `sign(x)*log(abs(x))` for all points except 0 where it should be returning `nan` . I'm not sure where this might be going wrong due to my inexperience with the `Integrals` module but the output seems to capture one half of the number line correctly and returns `nan` for all other values of `x`

NikhilSDate:
Fixes #22533

Currently, `integrate(1/Abs(x))` (where x is real) gives `Piecewise((-log(x), x<= 0), (nan, True))`. The same issue could also occur for piecewise functions whose antiderivatives on some interval are infinite at one of the endpoints. I changed the `_eval_integral` method of sympy/functions/elementary/piecewise.py to detect such a situation and try to use the antiderivative expression instead of the running sum. With the changes I made, `integrate(1/Abs(x))` now returns `Piecewise((-log(x), x<= 0), (log(x), True))`. The change also gives the correct result for other piecewise functions for which this bug exists.

#### Other comments
None

#### Release Notes

* functions
  * fixed a bug with integrating certain piecewise functions

### DIFF
--- a/sympy/functions/elementary/piecewise.py
+++ b/sympy/functions/elementary/piecewise.py
@@ -8,6 +8,7 @@ from sympy.core.sorting import ordered
 from sympy.functions.elementary.miscellaneous import Max, Min
 from sympy.logic.boolalg import (And, Boolean, distribute_and_over_or, Not,
     true, false, Or, ITE, simplify_logic, to_cnf, distribute_or_over_and)
+from sympy.polys.polyutils import illegal
 from sympy.utilities.iterables import uniq, sift, common_prefix
 from sympy.utilities.misc import filldedent, func_name
 
@@ -501,9 +502,14 @@ class Piecewise(Function):
                 sum = anti
             else:
                 sum = sum.subs(x, a)
-                if sum == Undefined:
-                    sum = 0
-                sum += anti._eval_interval(x, a, x)
+                if sum.has(*illegal):
+                    sum = anti
+                else:
+                    e = anti._eval_interval(x, a, x)
+                    if e.has(*illegal):
+                        sum = anti
+                    else:
+                        sum += e
             # see if we know whether b is contained in original
             # condition
             if b is S.Infinity:
@@ -512,11 +518,7 @@ class Piecewise(Function):
                 cond = (x < b)
             else:
                 cond = (x <= b)
-            if sum.has(S.Infinity, S.NegativeInfinity, S.ComplexInfinity, Undefined):
-                args.append((anti, cond))
-                sum = S.Zero
-            else:
-                args.append((sum, cond))
+            args.append((sum, cond))
         return Piecewise(*args)
 
     def _eval_interval(self, sym, a, b, _first=True):
--- a/sympy/functions/elementary/piecewise.py
+++ b/sympy/functions/elementary/piecewise.py
@@ -502,10 +502,8 @@ class Piecewise(Function):
                 sum = anti
             else:
                 sum = sum.subs(x, a)
-                if sum.has(*illegal):
-                    sum = S.Zero
                 e = anti._eval_interval(x, a, x)
-                if e.has(*illegal):
+                if sum.has(*illegal) or e.has(*illegal):
                     sum = anti
                 else:
                     sum += e
--- a/sympy/functions/elementary/piecewise.py
+++ b/sympy/functions/elementary/piecewise.py
@@ -512,7 +512,10 @@ class Piecewise(Function):
                 cond = (x < b)
             else:
                 cond = (x <= b)
-            args.append((sum, cond))
+            if sum.has(oo, -oo, S.ComplexInfinity, Undefined):
+                args.append((anti, cond))
+            else:
+                args.append((sum, cond))
         return Piecewise(*args)
 
     def _eval_interval(self, sym, a, b, _first=True):
--- a/sympy/functions/elementary/piecewise.py
+++ b/sympy/functions/elementary/piecewise.py
@@ -512,8 +512,9 @@ class Piecewise(Function):
                 cond = (x < b)
             else:
                 cond = (x <= b)
-            if sum.has(oo, -oo, S.ComplexInfinity, Undefined):
+            if sum.has(S.Infinity, S.NegativeInfinity, S.ComplexInfinity, Undefined):
                 args.append((anti, cond))
+                sum = S.Zero
             else:
                 args.append((sum, cond))
         return Piecewise(*args)
--- a/sympy/functions/elementary/piecewise.py
+++ b/sympy/functions/elementary/piecewise.py
@@ -503,13 +503,12 @@ class Piecewise(Function):
             else:
                 sum = sum.subs(x, a)
                 if sum.has(*illegal):
+                    sum = S.Zero
+                e = anti._eval_interval(x, a, x)
+                if e.has(*illegal):
                     sum = anti
                 else:
-                    e = anti._eval_interval(x, a, x)
-                    if e.has(*illegal):
-                        sum = anti
-                    else:
-                        sum += e
+                    sum += e
             # see if we know whether b is contained in original
             # condition
             if b is S.Infinity:
--- a/sympy/functions/elementary/tests/test_piecewise.py
+++ b/sympy/functions/elementary/tests/test_piecewise.py
@@ -34,43 +34,44 @@ z = symbols('z', nonzero=True)
 
 
 def test_piecewise1():
+
     # Test canonicalization
     assert unchanged(Piecewise, ExprCondPair(x, x < 1), ExprCondPair(0, True))
     assert Piecewise((x, x < 1), (0, True)) == Piecewise(ExprCondPair(x, x < 1),
                                                          ExprCondPair(0, True))
     assert Piecewise((x, x < 1), (0, True), (1, True)) == \
-           Piecewise((x, x < 1), (0, True))
+        Piecewise((x, x < 1), (0, True))
     assert Piecewise((x, x < 1), (0, False), (-1, 1 > 2)) == \
-           Piecewise((x, x < 1))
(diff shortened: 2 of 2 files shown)


NikhilSDate:
This pull request was accidental so I'm closing it

NikhilSDate:
I now realize that I should not have closed the PR, but should have instead edited it. I have edited the PR and am reopening it.

sympy/functions/elementary/piecewise.py
```diff
@@ -512,7 +512,10 @@ def handler(ipw):
                 cond = (x < b)
             else:
                 cond = (x <= b)
-            args.append((sum, cond))
+            if sum.has(oo, -oo, S.ComplexInfinity, Undefined):
```
jksuom, line 515: How about testing infinities on line 504?

ThePauliPrinciple:
Is there a reason for all the formatting? It would help for reviewing if the code only gets changed where it is relevant for the topic of the PR.

NikhilSDate:
> Is there a reason for all the formatting? It would help for reviewing if the code only gets changed where it is relevant for the topic of the PR.

Yes, I had accidently used PyCharm's autoformatter in the first commit. I think that has been fixed in the second commit though. Is there anything else I need to do in this regard?

ThePauliPrinciple:
No, I mistook the restoring of the format as changes in the format.

sympy/functions/elementary/piecewise.py
NikhilSDate: Testing infinities on line 504 does not catch infinities produced on line 506. I could change it so that it tests infinities right line 506 if that would help somehow.

oscargus: I'd suggest using `S.NegativeInfinity` instead of `-oo` to speed up the code (although both are very fast, an expression is created for the current code). An therefore also `S.Infinity` for consitency. What about `S.NaN`?

jksuom: If `sum` is infinite or undefined, then it should probably be reset to 0.

NikhilSDate: > If `sum` is infinite or undefined, then it should probably be reset to 0.

Yes, this seems like a good idea. I'll add it in along with oscargus's suggestion.

NikhilSDate: > I'd suggest using `S.NegativeInfinity` instead of `-oo` to speed up the code (although both are very fast, an expression is created for the current code). An therefore also `S.Infinity` for consistency. What about `S.NaN`?


I'll replace `oo` and `oo` with `S.NegativeInfinity` and `S.Infinity` respectively. In this file, `Undefined` is set to `S.NaN`, so `S.NaN` is already covered.

sympy/functions/elementary/piecewise.py
```diff
@@ -512,7 +512,11 @@ def handler(ipw):
                 cond = (x < b)
             else:
                 cond = (x <= b)
-            args.append((sum, cond))
+            if sum.has(S.Infinity, S.NegativeInfinity, S.ComplexInfinity, Undefined):
```
smichr, line 515: Two notes:
1) 
```python
from sympy.polys.polyutils import illegal  # [nan, oo, -oo, zoo] but import at top
sum.has(*illegal)
```
2) this may only be correct if there is a single tuple in done. If there are more then the proper change needs to happen were sum is being accumulated. And I'm not exactly sure what has to be done but suspect it needs to be done there. Please check.

smichr: ;-) maybe `illegal` should be named `nanoozoo`

NikhilSDate: > Two notes: 1)
> 
> ```python
> from sympy.polys.polyutils import illegal  # [nan, oo, -oo, zoo] but import at top
> sum.has(*illegal)
> ```
Yes, using this sounds like a good idea. I'll add it in.
> 
>     2. this may only be correct if there is a single tuple in done. If there are more then the proper change needs to happen were sum is being accumulated. And I'm not exactly sure what has to be done but suspect it needs to be done there. Please check.
I'm not sure what exactly you mean here. I tried my code with cases where `done` has more than one tuple in it and it seems to be working fine.

jksuom:
> Testing infinities on line 504 does not catch infinities produced on line 506. 

When are infinities produced on line 506?

NikhilSDate:
> When are infinities produced on line 506?

Integrating `Piecewise((x, x <= 0), (1 / x, True))`, for example, will produce an infinity on line 506. Currently, this returns `Piecewise((x**2/2, x <= 0), (log(x) + oo, True))`

jksuom:
> Integrating Piecewise((x, x <= 0), (1 / x, True)), for example, will produce an infinity on line 506. 

That seems to come from `anti._eval_interval(x, a, x)`. How about checking that before adding to `sum`? Something like
```
        e = anti._eval_interval(x, a, x)
        if e.has(...):
            e = anti
        sum += e
```

NikhilSDate:
> That seems to come from anti._eval_interval(x, a, x). How about checking that before adding to sum? Something like

Is there any advantage over using this over the current code? This could cause problems if line 503 produces an infinity but line 506 does not. For example, using your code, integrating something like `Piecewise((-1 / x, x <= 0), (x, True))` results in `Piecewise((-log(x), x <= 0), (x**2/2 + zoo, True))`.

jksuom:
> `Piecewise((-log(x), x <= 0), (x**2/2 + zoo, True))`

There should be no `zoo` if infinities are also checked on line 504.

NikhilSDate:
> There should be no `zoo` if infinities are also checked on line 504.

So do you recommend checking infinities both on line 504 and right after line 506? What advantages would this provide over the current code?

jksuom:
The goal is to make the antiderivative continuous at `a` if both `sum` and `anti` are there finite. It either one is infinite or undefined, then we can define `sum = anti`. In both cases we can use `args.append((sum, cond))`.

NikhilSDate:
I see your point. So are there cases for the current code where the antiderivative would not be continuous even though it is possible for such an antiderivative to exist? I would be happy to incorporate the changes you suggest but it just seems to me that checking infinities in one place instead of two is a bit cleaner. To take your suggestion of setting `sum = anti`, I could change it so that it does something like this
```
if sum.has(...):
    sum = anti
args.append((sum, cond))
```
instead of 
```            
if sum.has(...):
    args.append((anti, cond))
    sum = S.Zero
else:
    args.append((sum, cond))
```
Would that be alright?

jksuom:
> checking infinities in one place instead of two is a bit cleaner. 

There are two potential sources of infinities: `sum.subs(x, a)` and `anti._eval_interval(x, a, x)`, and both can be checked in one place:

        sum = sum.subs(x, a)
        e = anti._eval_interval(x, a, x)
        if  sum.has(...) or e.has(...):
            sum = anti
        else:
            sum += e

It would also be possible to check `sum` first and compute `e` only if `sum` is finite. There is no need to add `e` to `sum` if `e` has infinities.

NikhilSDate:
Yes, this sounds like a good idea. I'll make the change. @smichr any thoughts on this?

smichr:
> Yes, this sounds like a good idea. I'll make the change. @smichr any thoughts on this?

Anything that moves the test more back toward the source of the problem sounds good. And that looks like the direction this is going.

NikhilSDate:
> There are two potential sources of infinities: `sum.subs(x, a)` and `anti._eval_interval(x, a, x)`, and both can be checked in one place:

I tried your approach but it is failing the `test_holes` test in `test_piecewise.py`. The issue is that checking both `sum` and `e` in the same place does not handle the case where `sum` has infinities but `e` does not in the way the test requires it to. The test requires that in this case, sum should be set to `e`, not `anti`. It seems that to move the test for infinities near the source, it will be necessary to have two checks.
Is there anything else you suggest I should do before the pull request is merged?
Also, my most recent commit is failing the benchmark test with the error message `Config file /home/runner/work/sympy/sympy/asv.conf.actions.json not found.` This does not seem like a problem with my code. Could this be related to #22459?

jksuom:
> The test requires that in this case, sum should be set to e, not anti

Have you considered changing the test? Both `e` and `anti` should be valid antiderivatives. I would choose the one that leads to simpler expression.

jksuom:
I don't understand why github actions tries to run sympy/sympy/asv.conf.actions.json. @oscarbenjamin ?

ThePauliPrinciple:
Github actions used to use that file, but a while ago it suddenly changed to look for a different file (which made tests fail at that time). It seems like it has changed back again. You might have to wait a bit until a dev unmerges the previous fix.

oscarbenjamin:
The change was from #22481. Maybe the upstream bug was fixed and that change should be reverted now

NikhilSDate:
> The change was from #22481. Maybe the upstream bug was fixed and that change should be reverted now

I could submit a PR that changes the name of the file back to `asv.conf.actions.json` (from `asv.conf.json`). Should I do that?

jksuom:
> Should I go ahead with that?

I think that it is better to wait. That should not take too much time.

NikhilSDate:
Since the benchmark check isn't required for merging, is there any specific reason we are waiting for that check to be fixed before moving ahead with this pull request?

jksuom:
> the benchmark check isn't required for merging

So it seems. The PR looks good. Thanks!

PR #22544, opened 2021-11-23
Update blockmatrix.py

### LINKED ISSUE #21866: Error in inverse of BlockMatrix
The inverse computation of a square `BlockMatrix` with more than 3x3 blocks sometimes provides a incorrect solution. For example:

```python
import sympy as sym
from IPython.display import display

n  = 10
I  = sym.Identity(n)
O  = sym.ZeroMatrix(n, n)

A  = sym.BlockMatrix([[  I,  O,  O,  O ],
                      [  O,  I,  O,  O ],
                      [  O,  O,  I,  O ],
                      [  I,  O,  O,  I ]])
Ainv = sym.block_collapse(A.inv())
Am1A = sym.block_collapse(Ainv @ A)
display(Am1A)
```
```
>>> Matrix([
[  I, 0, 0, 0],
[  0, I, 0, 0],
[  0, 0, I, 0],
[2*I, 0, 0, I]])
```

```python
Id = sym.BlockMatrix([[  I,  O,  O,  O ],
                      [  O,  I,  O,  O ],
                      [  O,  O,  I,  O ],
                      [  O,  O,  O,  I ]])
print('inv(A)*A == Id: %s' % (Am1A == Id))
```
```
>>> False
```

Notice that the correct solution should be the following `B` matrix:
```python
B  = sym.BlockMatrix([[  I,  O,  O,  O ],
                      [  O,  I,  O,  O ],
                      [  O,  O,  I,  O ],
                      [ -I,  O,  O,  I ]])
BA = sym.block_collapse(B @ A)
print('B*A  == Id: %s' % (BA == Id))
```
```
>>> True
```

*Remark*: the same issue is present with any of the following methods:
```python
Ainv = sym.block_collapse(A**-1)
Ainv = sym.block_collapse(A.inverse())
Ainv = sym.block_collapse(sym.Inverse(A))
```

AkuBrains:
Fixes collapse Block Matrix Inverse 

Fixes #21866

When collapsing a matrix expression, in some cases, it happens that `bc_dist(expr)` forget to return the factor.
We fixed that by returning `expr` and not just `unpacked`.

#### Release Notes

* matrices
   * Fixed bug when collapsing inverse block matrices


oscargus:
Can you please add a test that shows that the problem has been fixed, i.e., a case that did not work before this fix?

Franck2111:
I used the issue's code to test the fix :
```Python
import sympy as sym
from IPython.display import display

n  = 10

I  = sym.Identity(n)
O  = sym.ZeroMatrix(n, n)

A  = sym.BlockMatrix([[  I,  O,  O,  O ],
                      [  O,  I,  O,  O ],
                      [  O,  O,  I,  O ],
                      [  I,  O,  O,  I ]])
Ainv = sym.block_collapse(A.inv())
Am1A = sym.block_collapse(Ainv @ A)
display(Am1A)
```
Before the fix, the test displays the following matrix while we expected the Indentity Matrix : 
```
>>> Matrix([
[  I, 0, 0, 0],
[  0, I, 0, 0],
[  0, 0, I, 0],
[2*I, 0, 0, I]])
```
The Matrix Inversion algorithm implemented in Sympy use several blockmatrix operation functions and in particular `bc_dist()'. Returning the full matrix expression gives the correct matrix as it returns now : 
```
>>> Matrix([
[  I, 0, 0, 0],
[  0, I, 0, 0],
[  0, 0, I, 0],
[ 0, 0, 0, I]])
```

oscargus:
That is probably the easiest way to do it, yes. If you can add a function `test_issue_21866` in 
https://github.com/sympy/sympy/blob/master/sympy/matrices/expressions/tests/test_blockmatrix.py that includes the code above (ideally without the `sym` part as most of the functions are imported anyway, better to import at the top in case you require some additional ones) and finishes with something like `assert Am1A == Id` (using Id from the original issue), that would solve it. In that way it cannot (easily) be broken again without noticing.

(In general, it is better to create a separate branch and not use master as master is your local copy of SymPy and if you later on update that, it may go into this PR and mess things up, primarily for you. Also, if you had used a different branch, I could have added the test code in your branch.)

I checked the surrounding code and your change indeed makes sense. What I found a bit weird is that `unpacked` is in this case a matrix multiplication (so not a plain matrix). Hence, it is not clear what the method *really* should return (but just returning `unpacked` is not correct for sure). Maybe one would like to have an `elif isinstance(unpacked, MatMul):` and apply factor to one of the matrices in the matrix multiplication? But that is a later issue and there should still be a default return of `expr` rather than `unpacked` for any uncovered cases.

Thanks for your contribution and efforts in finding this bug!

oscargus:
And without the display-part.

oscargus:
Thanks!

Some trailing white-space it seems. I've enable auto-merge so once the tests pass it will be merged.

PR #22551, opened 2021-11-26
core/containers: no longer create Tuple's with lists as arguments.

### LINKED ISSUE #22550: Tuple using sympify instead of _sympify incorrectly supports lists
Currently, `Tuple` supports the keyword `sympify` (`True` by default). However, this uses `sympify` rather than `_sympify`. As a result, `lists` appear to be supported as arguments to `Tuple` but this may result in unexpected behaviour since the resulting `Tuple` is now in principle mutable. Instead, I suggest to use `_sympify` over `sympify` and check for any `list` argument and convert those using `Tuple(*list_arg)`.

ThePauliPrinciple:

Fixes #22550

Several classes in SymPy use `sympify` on their arguments. In some cases, this leads to the creation of `Tuple` objects that have `lists` as their arguments. This PR solves those cases as far as they show up in the tests, by explicitly looping over iterables when a mix of `tuple` and `list` can be used for the input arguments.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/codegen/abstract_nodes.py
+++ b/sympy/codegen/abstract_nodes.py
@@ -1,0 +1,15 @@
+"""This module provides containers for python objects that are valid
+printing targets but are not a subclass of SymPy's Printable.
+"""
+
+
+from sympy.core.containers import Tuple
+
+
+class List(Tuple):
+    """Represents a (frozen) (Python) list (for code printing purposes)."""
+    def __eq__(self, other):
+        if isinstance(other, list):
+            return self == List(*other)
+        else:
+            return self.args == other
--- a/sympy/core/containers.py
+++ b/sympy/core/containers.py
@@ -15,6 +15,7 @@ from .sympify import _sympify, sympify, converter, SympifyError
 from sympy.utilities.iterables import iterable
 from sympy.utilities.misc import as_int
 
+import warnings
 
 class Tuple(Basic):
     """
@@ -49,10 +50,11 @@ class Tuple(Basic):
 
     def __new__(cls, *args, **kwargs):
         if kwargs.get('sympify', True):
-            args = (sympify(arg) for arg in args)
+            args = tuple((sympify(arg) for arg in args))
         for arg in args:
             if isinstance(arg, list):
-                raise TypeError("Tuple should not contain list as arguments")
+                msg = "Tuple should not contain list as arguments: {0}".format(arg)
+                warnings.warn( DeprecationWarning(msg))
         obj = Basic.__new__(cls, *args)
         return obj
 
--- a/sympy/core/containers.py
+++ b/sympy/core/containers.py
@@ -15,6 +15,7 @@ from .sympify import _sympify, sympify, converter, SympifyError
 from sympy.utilities.iterables import iterable
 from sympy.utilities.misc import as_int
 
+
 class Tuple(Basic):
     """
     Wrapper around the builtin tuple object.
--- a/sympy/core/containers.py
+++ b/sympy/core/containers.py
@@ -14,6 +14,7 @@ from .sorting import default_sort_key
 from .sympify import _sympify, sympify, converter, SympifyError
 from sympy.utilities.iterables import iterable
 from sympy.utilities.misc import as_int
+import warnings
 
 class Tuple(Basic):
     """
@@ -48,7 +49,13 @@ class Tuple(Basic):
 
     def __new__(cls, *args, **kwargs):
         if kwargs.get('sympify', True):
-            args = tuple((Tuple(*arg) if isinstance(arg, list) else sympify(arg) for arg in args))
+            args = (sympify(arg) for arg in args)
+        #if kwargs.get('denest_lists', False):
+        #    args = (Tuple(*arg) if isinstance(arg, list) else arg for arg in args)
+        args = tuple(args)
+        for arg in args:
+            if isinstance(arg, list):
+                warnings.warn(DeprecationWarning('Tuple',arg))
         obj = Basic.__new__(cls, *args)
         return obj
 
--- a/sympy/core/containers.py
+++ b/sympy/core/containers.py
@@ -14,7 +14,6 @@ from .sorting import default_sort_key
 from .sympify import _sympify, sympify, converter, SympifyError
 from sympy.utilities.iterables import iterable
 from sympy.utilities.misc import as_int
-import warnings
 
 class Tuple(Basic):
     """
--- a/sympy/core/containers.py
+++ b/sympy/core/containers.py
@@ -50,6 +50,9 @@ class Tuple(Basic):
     def __new__(cls, *args, **kwargs):
         if kwargs.get('sympify', True):
             args = (sympify(arg) for arg in args)
+        for arg in args:
+            if isinstance(arg, list):
+                raise TypeError("Tuple should not contain list as arguments")
         obj = Basic.__new__(cls, *args)
         return obj
 
--- a/sympy/core/containers.py
+++ b/sympy/core/containers.py
@@ -50,12 +50,6 @@ class Tuple(Basic):
     def __new__(cls, *args, **kwargs):
         if kwargs.get('sympify', True):
             args = (sympify(arg) for arg in args)
-        #if kwargs.get('denest_lists', False):
-        #    args = (Tuple(*arg) if isinstance(arg, list) else arg for arg in args)
-        args = tuple(args)
-        for arg in args:
-            if isinstance(arg, list):
-                warnings.warn(DeprecationWarning('Tuple',arg))
         obj = Basic.__new__(cls, *args)
         return obj
 
--- a/sympy/core/containers.py
+++ b/sympy/core/containers.py
@@ -15,8 +15,6 @@ from .sympify import _sympify, sympify, converter, SympifyError
 from sympy.utilities.iterables import iterable
 from sympy.utilities.misc import as_int
 
-import warnings
-
 class Tuple(Basic):
     """
     Wrapper around the builtin tuple object.
--- a/sympy/core/containers.py
+++ b/sympy/core/containers.py
@@ -50,11 +50,7 @@ class Tuple(Basic):
 
     def __new__(cls, *args, **kwargs):
         if kwargs.get('sympify', True):
-            args = tuple((sympify(arg) for arg in args))
-        for arg in args:
-            if isinstance(arg, list):
-                msg = "Tuple should not contain list as arguments: {0}".format(arg)
-                warnings.warn( DeprecationWarning(msg))
+            args = tuple((Tuple(*arg) if isinstance(arg, list) else sympify(arg) for arg in args))
         obj = Basic.__new__(cls, *args)
         return obj
 
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -1270,7 +1270,7 @@ class Derivative(Expr):
                         must be supplied to differentiate %s''' % expr))
 
         # Standardize the variables by sympifying them:
-        variables = list(sympify(variables))
+        #variables = sympify(list(variables))
 
         # Split the list of variables into a list of the variables we are diff
         # wrt, where each element of the list has the form (s, count) where
@@ -1278,11 +1278,12 @@ class Derivative(Expr):
         # derivative.
         variable_count = []
         array_likes = (tuple, list, Tuple)
+        integer_likes = (int, Integer)
 
         from sympy.tensor.array import Array, NDimArray
 
         for i, v in enumerate(variables):
-            if isinstance(v, Integer):
+            if isinstance(v, integer_likes):
                 if i == 0:
                     raise ValueError("First variable cannot be a number: %i" % v)
                 count = v
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -1269,9 +1269,6 @@ class Derivative(Expr):
                         expression, the variable(s) of differentiation
                         must be supplied to differentiate %s''' % expr))
 
-        # Standardize the variables by sympifying them:
-        #variables = sympify(list(variables))
-
         # Split the list of variables into a list of the variables we are diff
         # wrt, where each element of the list has the form (s, count) where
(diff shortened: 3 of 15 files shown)


sympy/core/containers.py
```diff
@@ -50,6 +50,12 @@ class Tuple(Basic):
     def __new__(cls, *args, **kwargs):
         if kwargs.get('sympify', True):
             args = (sympify(arg) for arg in args)
+        #if kwargs.get('denest_lists', False):
```
ThePauliPrinciple, line 53: This is to be removed before merging, but shows that currently there are no longer any `Tuple` with `list` args being created during the testing.

sympy/core/function.py
```diff
@@ -1270,19 +1270,20 @@ def __new__(cls, expr, *variables, **kwargs):
                         must be supplied to differentiate %s''' % expr))
 
         # Standardize the variables by sympifying them:
-        variables = list(sympify(variables))
+        #variables = sympify(list(variables))
```
ThePauliPrinciple, line 1273: This is actually not needed here since all variables are eventually stored in `Tuple` (or `Array` which again stores it in `Tuple`) where `sympify` is also used.

sympy/printing/tests/test_julia.py
```diff
@@ -212,7 +212,8 @@ def test_containers():
     assert julia_code(Tuple(*[1, 2, 3])) == "(1, 2, 3)"
     assert julia_code((1, x*y, (3, x**2))) == "(1, x.*y, (3, x.^2))"
     # scalar, matrix, empty matrix and empty list
-    assert julia_code((1, eye(3), Matrix(0, 0, []), [])) == "(1, [1 0 0;\n0 1 0;\n0 0 1], zeros(0, 0), Any[])"
+    from sympy.codegen.pynodes import List
```
ThePauliPrinciple, line 215: Should there be sympy/codegen/{julia,maple,octave}nodes.py files defining `List` or perhaps this means that `List` should be moved to a more general file in sympy/codegen. Obviously the import should be made top-level but I put it here for convenience of review.

sympy/simplify/tests/test_cse.py
```diff
@@ -584,8 +584,10 @@ def test_cse_list():
     assert _cse(x) == ([], x)
     assert _cse('x') == ([], 'x')
     it = [x]
-    for c in (list, tuple, set, Tuple):
+    for c in (list, tuple, set):
```
ThePauliPrinciple, line 587: `Tuple` doesn't do typecasting, the equivalent of `tuple(x)` is `Tuple(*x)`.

sympy/core/containers.py
```diff
@@ -50,6 +50,12 @@ class Tuple(Basic):
     def __new__(cls, *args, **kwargs):
```
oscarbenjamin, line 50: Probably this should use an explicit keyword argument:
```python
def __new__(cls, *args, sympify=False):
```

sympy/core/containers.py
```diff
@@ -50,6 +50,12 @@ class Tuple(Basic):
...
         if kwargs.get('sympify', True):
             args = (sympify(arg) for arg in args)
+        #if kwargs.get('denest_lists', False):
+        #    args = (Tuple(*arg) if isinstance(arg, list) else arg for arg in args)
+        args = tuple(args)
+        for arg in args:
+            if isinstance(arg, list):
+                warnings.warn(DeprecationWarning('Tuple',arg))
```
oscarbenjamin, line 58: Deprecation warnings in SymPy are usually given like this:
https://github.com/sympy/sympy/blob/ed0bfd1e47345c8bd45ef542eef18f6a21c55685/sympy/core/sympify.py#L453-L460
The SymPyDeprecationWarning class is handled specially by the test runner so that it is an error if it is emitted in the test suite.

ThePauliPrinciple: As far as I am concerned, this warning will be removed before the merge (see also the first review comment by me). If such a warning is to be added, it should be added to `Basic` when all such issues are resolved. This is for demonstration purposes. (that there are no longer such violations)

sympy/core/containers.py
ThePauliPrinciple: Would it be confusing if this gets changed in this PR? Since it solves a different issue. Note also that the current default value is `sympify=True`.

sympy/printing/tests/test_julia.py
oscarbenjamin: I'm not sure that List has the same meaning in all languages or if this should be considered an abstract "list" that would have different implementations in different languages.

ThePauliPrinciple: I think each code printer has its own way to print a `list, but since `sympify` is used on the input to `CodePrinter`s, this line causes the creation of a `Tuple`, which then has a `list` in it. Changing this to a `List` seems to avoid this problem, I'm not sure there is a better way. `List` is actually just a wrapper around `Tuple` which gets printed differently (if I understand correctly, it defaults to printing the same as `list`.

oscarbenjamin: I think you know more about this than I do so I can't really advise but I trust your judgement.

ThePauliPrinciple: @bjodah do you have any preference for how this is dealt with?

bjodah: I'm fine with moving it. I put it in pynodes to signal equivalence with Python lists.

sympy/core/function.py
oscarbenjamin: Maybe this should just be deleted.

oscarbenjamin:
Is this finished now?

ThePauliPrinciple:
@bjodah Do you have an opinion about the new location of `List`? I also noticed that `CodePrinter` actually uses `Lists` (regardless of language), but only after creating `Tuple`s first. 

I think it is possible to have `sympify` use extra rules, so it might be better to have `CodePrinter` already convert `list`s to `List`s inside `sympify` so that the current tests do not have to be changed.

Is `abstract_nodes` good name for the new file? Or would `container_nodes` be better?

sympy/printing/tests/test_julia.py
```diff
@@ -212,7 +213,7 @@ def test_containers():
     assert julia_code(Tuple(*[1, 2, 3])) == "(1, 2, 3)"
     assert julia_code((1, x*y, (3, x**2))) == "(1, x.*y, (3, x.^2))"
     # scalar, matrix, empty matrix and empty list
-    assert julia_code((1, eye(3), Matrix(0, 0, []), [])) == "(1, [1 0 0;\n0 1 0;\n0 0 1], zeros(0, 0), Any[])"
```
bjodah, line 215: this looks like it introduces backwards incompatibility? If so, I'm afraid it needs a deprecation cycle.

sympy/printing/tests/test_maple.py
```diff
@@ -233,8 +234,7 @@ def test_containers():
     assert maple_code(Tuple(*[1, 2, 3])) == "[1, 2, 3]"
     assert maple_code((1, x * y, (3, x ** 2))) == "[1, x*y, [3, x^2]]"
     # scalar, matrix, empty matrix and empty list
-
-    assert maple_code((1, eye(3), Matrix(0, 0, []), [])) == \
+    assert maple_code((1, eye(3), Matrix(0, 0, []), List(*[]))) == \
```
bjodah, line 237: same thing here and for the octave test too.

bjodah (changes requested):
I'm fine with putting the class in `.abstract_nodes`.

sympy/printing/codeprinter.py
```diff
@@ -134,8 +134,19 @@ def _handle_assign_to(expr, assign_to):
                         type(self).__name__, type(assign_to)))
             return Assignment(assign_to, expr)
 
+        #for codeprinting, all lists should be converted to abstract_nodes.Lists.
+        from sympy.codegen.abstract_nodes import List
+        store_list_converter=converter.get(list, None)
+        converter[list]=lambda x: List(*x)
```
ThePauliPrinciple, line 140: I'm adding `list->List` conversion to `sympify` for the purposes of `CodePrinting`, this will allow current tuple/list nesting without creating `list`s in `Tuple`s but not change anything for the user. Is there a better way to do this?

sympy/printing/codeprinter.py
```diff
@@ -134,8 +134,19 @@ def _handle_assign_to(expr, assign_to):
...
+
         expr = _handle_assign_to(expr, assign_to)
-        expr = _convert_python_lists(expr)
+
+        #restore previous behaviour
+        if store_list_converter is None:
+            del converter[list]
+        else:
+            converter[list] = store_list_converter
+
```
ThePauliPrinciple, line 149: Restore `sympify` to normal behaviour. Does it make sense to add this as an option to `symplify`?

sympy/printing/codeprinter.py
bjodah: This kind of coding is fragile in my humble opinion: importing a mutable object, modifying it in-place, calling a function using it, and then reverting the change. If the `converter` in `.sympify` is intended to be modifiable, then I'd rather see a context manager class in `.sympify` allowing to override the default converters for a region of code, would you agree?

oscarbenjamin: This is also not thread-safe. The converters are only really supposed to be registered during start up.

Probably somewhere there should explicit code that checks for `list` and converts to `List` rather than depending on `sympify` to do this implicitly.

sympy/printing/codeprinter.py
```diff
@@ -134,8 +135,19 @@ def _handle_assign_to(expr, assign_to):
                         type(self).__name__, type(assign_to)))
             return Assignment(assign_to, expr)
 
+        #for codeprinting, all lists should be converted to abstract_nodes.Lists.
+        from sympy.codegen.abstract_nodes import List
+        store_list_converter=converter.get(list, None)
+        converter[list]=lambda x: List(*x)
+
         expr = _handle_assign_to(expr, assign_to)
-        expr = _convert_python_lists(expr)
```
bjodah, line 138: This is where the explicit conversion was done previously.

ThePauliPrinciple: It needs to happen before `_handle_assign_to` (since it will gladly convert `([],)` to `Tuple([])` in `sympify` which should not be allowed).

sympy/printing/codeprinter.py
ThePauliPrinciple: Point taken, I'll try to see of sympy.strategies can be of any help.

sympy/printing/codeprinter.py
bjodah: can we simply move the call up?

ThePauliPrinciple: Currently `_convert_python_lists` only goes through `list`s (so it would stop as soon as it sees the `tuple`). I could try to change it to loop through `tuple` too.

sympy/printing/codeprinter.py
ThePauliPrinciple: I think this has been solved.

sympy/printing/codeprinter.py
ThePauliPrinciple: I updated `_convert_python_lists` to recurse through `tuple` and moved it forward. I believe this solves all issues.

bjodah (approved):

bjodah:
Thanks! The changes regarding List now looks good to me 🙂👍

sympy/codegen/abstract_nodes.py
```diff
@@ -0,0 +1,15 @@
...
+printing targets but are not a subclass of SymPy's Printable.
+"""
+
+
+from sympy.core.containers import Tuple
+
+
+class List(Tuple):
+    """Represents a (frozen) (Python) list (for code printing purposes)."""
+    def __eq__(self, other):
```
oscarbenjamin, line 11: Is the `__eq__` method needed for this class to function or is it just for the tests?

In general it's better if Basic subclass don't override `__eq__`.

ThePauliPrinciple:
```python
from sympy import *
class List(Tuple):
    pass
print([1] == List(1))

class List(Tuple):
    """Represents a (frozen) (Python) list (for code printing purposes)."""
    def __eq__(self, other):
        if isinstance(other, list):
            return self == List(*other)
        else:
            return self.args == other
print([1] == List(1))
```
First case is `False`, second is `True`, so I do believe that is needed here.

ThePauliPrinciple:
```python
from sympy import *
class List(Tuple):
    pass
print([1] == List(1))

class List(Tuple):
    """Represents a (frozen) (Python) list (for code printing purposes)."""
    def __eq__(self, other):
        if isinstance(other, list):
            return self == List(*other)
        else:
            return self.args == other
print([1] == List(1))
```
Shows that the first case is `False` while the second case is `True`

oscarbenjamin: Yes, but why does that matter? Does the List class actually need to be compared (with `==`) to the list class in order to function properly in actual use?

ThePauliPrinciple: I'm not sure, that would have to be tested. Note that I only moved this class. It already had this functionality, so I assume there was a reason (perhaps a bit naive). It was previously in pynodes.py, which should indicate that this is specific to python printers, which it actually is not since it was already being used for all `CodePrinter`s. The 'bug' related to this class was actually that it had to be used earlier in `CodePrinter.doprint` to avoid converting `Tuple`s with `list`s as arguments.

oscarbenjamin: > Note that I only moved this class

Oh, okay. Let's just leave it for now then.

oscarbenjamin:
Looks good. Thanks!

sympy/codegen/abstract_nodes.py
bjodah: I think we can remove the __eq__ method and the test if we want. I didn't reflect over implications of overriding __eq__ in a Basic subclass.

PR #22555, opened 2021-11-27
Series/limits : Fixes basic bugs encountered while evaluating limits of Piecewise Functions

anutosh491:
I have two Pr's (under review currently) which somewhat deal with limits for Piecewise functions and while trying stuff out I encountered some issues which I've tried to resolve , (not sure if these have been reported somewhere before will do a check on that). It goes as follows ... so lets consider a function as follows

image

On Master
```
z = Piecewise((x - 3, x < 0), ((x + 1)**2, x < 10), (5, True))
>>> limit(z, x, 0, dir= '-')
-3
>>> limit(z, x, 0, dir= '+')
-3
>>> limit(z, x, 10, dir= '-')
5
>>> limit(z, x, 10, dir= '+')
5
>>> limit(z, x, oo)
TypeError: Invalid comparison of non-real zoo

>>> limit(z, x, -oo)
TypeError: Invalid comparison of non-real zoo
```
And we know that, if the function's value at the limit is different from the limit's value, it doesn't mean the limit doesn't exist! So obviously these outputs are not what we expect . This was due to the fact that `_eval_as_leading_term` method of class `Piecewise` wasn't taking directions into account . I'm confident about my approach in this case but the implementation is a bit naive .

 I wish we could have something like `get_cond`  method in class `Piecewise` where we could pass in the point and get the condition which it fits into . This would have helped a lot but as of now, we are dealing with directions and limit values might not be function values so we need some sort of substitutions to confirm our check around 0 . For that I have used `1e-3` .

Also I discovered any expression having a number with True as a `ExprCondPair` wasn't working irrespective of the other error. 
For eg something like `Piecewise((x , x<0) , (5, True))` wasn't working at any given point and I have added a sperate check for that to pass .

Hence results on this branch
```
expr = Piecewise((x - 3, x < 0), ((x + 1)**2, x < 10), (5, True))
>>> limit(expr, x, 0, dir = '+') 
1
>>> limit(expr, x, 0, dir = '-') 
-3
>>> limit(expr, x, 10, dir = '+')
5
>>> limit(expr, x, 10, dir = '-')
121
>>> limit(expr, x, oo)
5
>>> limit(expr, x, -oo)
-oo
```

#### Release Notes

* series
  * Fixes basic bugs encountered while evaluating limits of Piecewise Functions.


sympy/core/function.py
```diff
@@ -621,8 +621,11 @@ def _eval_is_commutative(self):
         return fuzzy_and(a.is_commutative for a in self.args)
 
     def _eval_is_meromorphic(self, x, a):
+        from sympy.functions.elementary.piecewise import Piecewise
         if not self.args:
             return True
+        if isinstance(self, Piecewise) and any((arg.has(x) or arg[0].is_number) for arg in self.args[1:]):
```
jksuom, line 627: Why should arg in self.arg[1:] not have x? A univariate Piecewise function may have x in each argument.

jksuom: This method if for general functions that may have several arguments. To say that such a function is meromorphic means that it is meromorphic in the first argument. This does not work with Piecewise functions. There should be a `Piecewise.is_meromorphic` that would invoke `_eval_is_meromorphic` for the piece specified by `a`.

anutosh491: Yeah True , this was just a low level implementation and I expecting that someone points out that `Piecewise.is_meromorphic` , should be implemented . I'll try this out and get back soon !

anutosh491:
@jksuom , Current implementation works properly for all cases except the ones having `dir` passed as `+-` . This doesn't give error but gives one of the limits as answer . To tackle this i've implemented a dummy helper function which could be used to fix this !
Procedure is as follows (considering some  random example                                                                                                            `f = Piecewise(( asinh(x), x < 1), ( x + 5, x < 5), (log(x), True))`)

Step 1 : Suppose we need to check for two sided limit at `0` , I first check if some margin to 0 from both sides falls in the same interval (for eg `0 - 1e-3` and `0 + 1e+3` are both less than `1` , in that case conisder `e` to be `asinh(x)` and calculate 2 sided limit

Step2 : If the above step is not the case for eg suppose we need the limit at 1 , then we calculate two limits, the right handed side would conisder `x + 5` and the left sided one would consider `asinh(x)`


```
        def _eval_piecewise_limits(e, z, z0, cdir):                                   
                i = 0
                for arg in e.args:
                    if arg[1].subs(z, z0 - 1e-3) and arg[1].subs(z, z0 + 1e-3):
                        e = arg[0]
                        return limit(e, z, z0)
                    elif arg[1].subs(z, z0 - 1e-3) and not arg[1].subs(z, z0 + 1e-3):
                        l = limit(e.args[i][0], z, z0, '-')
                        r = limit(e.args[i + 1][0], z, z0, '+')
                        if l != r:
                            raise ValueError("The limit does not exist since "
                                         "left hand limit = %s and right hand limit = %s"
                                         % (l, r))
                        else:
                            return r
                    i = i+1
```
I tried out the above implementation for few examples and it works fine for eg
```
>>> limit(f, x, 10, dir = '+-')
log(10)
>>> limit(f, x, 5, dir = '+-')
ValueError: The limit does not exist since left hand limit = 10 and right hand limit = log(5)
>>> limit(f, x, 1, dir = '+-')
ValueError: The limit does not exist since left hand limit = -log(-1 + sqrt(2)) and right hand limit = 6
```
I would be glad to know your thoughts regarding the implementation and the approach (what can be improved and if I have missed something) . @0sidharth , I would be happy your have your views too on the same

sympy/functions/elementary/piecewise.py
```diff
@@ -305,12 +306,42 @@ def doit(self, **hints):
...
+            if arg[1].subs(x, a):
+                return arg[0]._eval_is_meromorphic(x, a)
+
     def _eval_simplify(self, **kwargs):
         return piecewise_simplify(self, **kwargs)
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
         for e, c in self.args:
-            if c == True or c.subs(x, 0) == True:
+            if (cdir == -1 and c.subs(x, -1e-3)) or ((cdir == 1 or cdir == 0) and c.subs(x, 1e-3)):
```
jksuom, line 344: This should not depend on exact values like `cdir == 1`. Although such values are currently set in limit code, any positive value could be used instead. All numbers of the form `t*c` for positive `t` define the same complex direction. Also values like `-1e-3` will not work in general.

jksuom: We should probably set `cdir = S.One` in the limit code. Then the sign could be checked by `cdir.is_positive`.

smichr:
> I wish we could have something like get_cond method in class Piecewise where we could pass in the point and get the condition which it fits into . 

How would this help? Does `Piecewise._intervals` help? Could you just do
```python
for i, (e, c) in enumerate(pw.args):
   b = c.subs(x, v)
    if b == True:
        b = i, c
        break
else:
    b = None
if b is not None:
    # condition i which is c is True at x == v
```

sympy/functions/elementary/piecewise.py
anutosh491: > 

Yeah agreed , I was just trying to get things working first of all and convey my approach and then implement all relevant suggestions . The basic approach was that if we consider an example as follows `expr = Piecewise((x - 3, x < 0), ((x + 1)**2, x < 10), (5, True))`

image

We want the `as_leading_term(x)` from the negative side to be `-3` given by the this piece `(x - 3, x < 0)` and `+1` from the right  given by the this piece `((x + 1)**2, x < 10)`  right ?? Now how should I approach this ? I thought of having a margin which gets us a value just lesser than `0` and a value just greater than `0`  to approach this and I ended up using `1e-3` just as a dummy 

Replying to @smichr , I thought of getting the expression from the negative side when `cdir = -1` and the following consecutive expression from the positive side when `cdir = 1` and for the same I thought of proposing a `get_cond ` method . Surely I would try out your suggestion , I was actually trying to using the` ._intervals method` sometime back and I'll try to use your soln for the same

anutosh491:
> > I wish we could have something like get_cond method in class Piecewise where we could pass in the point and get the condition which it fits into .
> 
> How would this help? Does `Piecewise._intervals` help? Could you just do
> 
> ```python
> for i, (e, c) in enumerate(pw.args):
>    b = c.subs(x, v)
>     if b == True:
>         b = i, c
>         break
> else:
>     b = None
> if b is not None:
>     # condition i which is c is True at x == v
> ```


https://github.com/sympy/sympy/pull/22555#discussion_r757912745

sympy/functions/elementary/piecewise.py
anutosh491: > 

Yupp this should surely be done

smichr:
So it seems you need to know if a limit point is at the boundary between two pieces. Seems like '_intervals' would be helpful since it will flatten the condition logic and give you the boundaries for each expression. If the limit point is a boundary value then there are two functions to consider, else only one.

anutosh491:
> So it seems you need to know if a limit point is at the boundary between two pieces. Seems like '_intervals' would be helpful since it will flatten the condition logic and give you the boundaries for each expression. If the limit point is a boundary value then there are two functions to consider, else only one.
```

        upper_bound_list, cond_list =[], [] 
        for a, b, e, i in self._intervals(x)[1]:
            upper_bound_list.append(b)
            cond_list.append(e)
        i = 0
        lst = list(zip(cond_list, upper_bound_list))
        for pair in lst:
            if pair[1].is_positive:                            # Step 1
                return pair[0].as_leading_term(x)              
            elif pair[1].is_zero:                              # Step 2
                if cdir == -1:
                    return pair[0].as_leading_term(x)
                else:
                    return pair[i + 1][0].as_leading_term(x)
            i = i + 1
```
This is a somewhat basic implementation of the idea above . For example 
Step1 
`expr = Piecewise((x - 3, x <= 2), ((x + 1)**2, x < 10), (5, True))` . First using `._intervals` method we create an upper bound , condition pair. Hence suppose we want to calculate `limit(expr, x, 0)` we get a the pair as `[(x - 3, 2), ((x + 1)**2, 10), (5, oo)]` and now that we know that while crossing 0 the term would be `x-3` without discontinuity and hence we use that .

Step2 
`expr = Piecewise((x - 3, x <= 0), ((x + 1)**2, x < 10), (5, True))` . Now if we want to calculate `limit(expr, x, 0)` where there is a change at 0 , we have the pair as  `[(x - 3, 0), ((x + 1)**2, 10), (5, oo)]` and now according to `cdir` value , we use either `x-3` or `(x + 1)**2` . 

And I find everything working fine for most of the test cases I tried . But sometimes while taking limit at `oo` , we have some minor bugs . For example consider `limit(expr, x, oo)` . we have 
`self` =` Piecewise((-3 + 1/x, 1/x <= 0), ((1 + 1/x)**2, 1/x < 10), (5, True))` due to the transformation in `limits.py` line 275 then 
`self._intervals[1] ` = `[(-oo, 0, -3 + 1/x, 0), (1/10, oo, (1 + 1/x)**2, 1), (-oo, 0, (1 + 1/x)**2, 1), (-oo, oo, 5, 2)]` and we end up with the pair `[(-3 + 1/x, 0), ((1 + 1/x)**2, oo), ((1 + 1/x)**2, 0), (5, oo)]` and as we have two values with the same upper bound the method returns the wrong answer . 
I am not sure how should I be tackling this ! Any suggestions @smichr @jksuom

jksuom:
For `is_meromorphic`, it should be first checked whether `a` is a boundary point. If it is, the `None` can be returned. (Actually, `False` could be returned except in case where expressions are the same on both sides but we should probably not need to check that).

For `_eval_as_leading_term`, it should also be checked if `x = 0` is a boundary point. If it is, then `cdir` should be nonzero. If it is zero, then `ValueError` can be raised. (It is possible that the same leading term appears on both sides but that need not be checked here.)

anutosh491:
Yeah ,the above code I've shared works just like you and Christopher pointed out ,  for most cases except some and as soon as I get some help or realize how to tackle take i'll make the commit !
The point is that tackling limit cases  for expressions like this  `expr = Piecewise((x - 3, x <= 0), ((x + 1)**2, x < 10), (5, True))` at `oo` is troubling me .
 This is because code from `limits.py` would transform this into `Piecewise((-3 + 1/x, 1/x <= 0), ((1 + 1/x)**2, 1/x < 10), (5, True))` . Now as @smichr pointed it is really convinient to use `._intervals` method for getting the boundary points and by basic logic the code goes as
```
check if boundary point is 0:
          if cdir.is_negative:
                 take the piece
          else:
                take the next piece
```
This works for mostly all cases like if we want `limit(expr, x, 0)` , we would use use `(x -3)` for negative cdir and `(x + 1)**2` for positive cdir.
 But the problem with taking  `limit(expr, x, oo)` is that after transofrmation ( `subs(x, 1/x) ` happens and we get `Piecewise((-3 + 1/x, 1/x <= 0), ((1 + 1/x)**2, 1/x < 10), (5, True))`) and applying `self._intervals(x)[1]` gives us `[(-oo, 0, -3 + 1/x, 0), (1/10, oo, (1 + 1/x)**2, 1), (-oo, 0, (1 + 1/x)**2, 1), (-oo, oo, 5, 2)]` which has multiple `0` as boundary points ( boundary points being `[0, oo, 0, oo]` ) . Hence now if we take up the next piece we would get this piece `(1 + 1/x)**2` rather than the `5` piece .

If I somehow understand how this could be tackled i'm sure that all cases would work as expected !!!

jksuom:
> the problem with taking limit(expr, x, oo) is that after transofrmation ( subs(x, 1/x) happens and we get Piecewise((-3 + 1/x, 1/x <= 0), ((1 + 1/x)**2, 1/x < 10), (5, True))

The implementation of `subs(x, 1/x)` for a piecewise function is complicated (not a simple `replace` operation) because the mapping `x -> 1/x` inverts the order on both the negative and positive real axis. The substituted expression should be completely rewritten to heed the new order. Fortunately, a full implementation of `subs` is not necessary for the purpose of computing leading terms and series expansions at infinities as only the first (starting from `-oo`) and last (ending at `oo`) pieces are needed. I would implement a special method, say `Piecewise.at_infinity`, that would return a piecewise expression having two pieces with the origin as their common boundary point. The relevant bit of limit code would be something like:

        if abs(z0) is S.Infinity:
            if e.is_Mul:
                e = factor_terms(e)
            if isinstance(e, Piecewise):
                newe = e.at_infinity(z)
            else:
                newe = e.subs(z, 1/z)
            cdir = -cdir
 
The meromorphic case need not be changed as it is expected that piecewise function is never meromorphic at infinity (or even if it happens to be, `is_meromorphic` will return `None`).

anutosh491:
> The implementation of `subs(x, 1/x)` for a piecewise function is complicated (not a simple `replace` operation) because the mapping `x -> 1/x` inverts the order on both the negative and positive real axis. The substituted expression should be completely rewritten to heed the new order. Fortunately, a full implementation of `subs` is not necessary for the purpose of computing leading terms and series expansions at infinities as only the first (starting from `-oo`) and last (ending at `oo`) pieces are needed. I would implement a special method, say `Piecewise.at_infinity`, that would return a piecewise expression having two pieces with the origin as their common boundary point. The relevant bit of limit code would be something like:
> 
> ```
>     if abs(z0) is S.Infinity:
>         if e.is_Mul:
>             e = factor_terms(e)
>         if isinstance(e, Piecewise):
>             newe = e.at_infinity(z)
>         else:
>             newe = e.subs(z, 1/z)
>         # cdir changes sign as oo- should become 0+
>         cdir = -cdir
> ```
> 
> The meromorphic case need not be changed as it is expected that piecewise function is never meromorphic at infinity (or even if it happens to be, `is_meromorphic` will return `None`).

Thanks for this ! Gives me a much better direction to proceed . I'll try the suggestions out and get back in a days' time

anutosh491:
The latest commit I made works completely for any case . I ended up checking lot of cases for this . Actually I went through `test_piecewise.py` and tried to understand what sort of expressions we might encounter in our code . So i came across these set of expressions
```
    p0 = Piecewise((0, Or(x < -2, x > 2)), (1, True))
    p1 = Piecewise((0, x < -2), (0, x > 2), (1, True))
    p2 = Piecewise((0, x > 2), (0, x < -2), (1, True))
    p3 = Piecewise((0, x < -2), (1, x < 2), (0, True))
    p4 = Piecewise((0, x > 2), (1, x > -2), (0, True))
    p5 = Piecewise((1, And(-2< x, x < 2)), (0, True))
```
which is nothing but 1 from -2 to 2 and 0 everywhere else but these are 6 ways to represent different pieces and way to represent the same expression . Considering this set I understood all loop holes in my code and earlier it was only catering to `p3, p4` but this gave me much better direction to proceed. I've checked limit for all relevant points (`-oo, -2, 0, 2, oo`) from both directions(`+ve, -ve`) and everything works as expected . 
Finally I ended up trying random expressions with random boundaries .... some being these and everything's working fine
```
expr = Piecewise((x - 3, x <= -3), ((x + 1)**2, x < -2),(1/x, x<10), (5, True))
expr = Piecewise((x - 3, x <= -3), ((x + 1)**2, x <= -2),(1/x, x<0), (5, True))
expr = Piecewise((x - 3, x <= -5), ((x + 1)**2, x < 0),(1/x, x<10), (5, True))
```

sympy/functions/elementary/piecewise.py
```diff
@@ -305,13 +306,87 @@ def doit(self, **hints):
...
+        # Check whether 0 lies in the Interval of any piece.
+        # If yes, we consider that piece for evaluation.
+        for bound in self.as_expr_set_pairs():
+            if S.Zero in bound[1] and not S.Zero in bound[1].boundary:
+                return bound[0].as_leading_term(x)
+
+    def _at_infinity(self, x):
+        at_infinity = []
+        for i, pair in enumerate(self.as_expr_set_pairs()):
+            if pair[1].inf is S.NegativeInfinity or pair[1].sup is S.Infinity:
```
jksuom, line 387: Are these the first and last items of `self.as_expr_set_pairs()`? It so, then we could use the indices 0 and -1 without `enumerate`.

sympy/functions/elementary/piecewise.py
```diff
@@ -305,13 +306,87 @@ def doit(self, **hints):
...
+                at_infinity.append(self.args[i])
+        return Piecewise(*tuple(at_infinity))
```
jksuom, line 389: There should be 2 pieces. The first condition should be `x < 0` in order to make 0 the single boundary point. The expressions should have `1/x` substituted for `x`. (After all, this is expected to be a simplified `subs(x, 1/x)`.)

anutosh491: Ohh ok , I haven't disturbed the `subs(x, 1/x)` transformation from limits.py yet .
```
            if isinstance(e, Piecewise):
                e = e._at_infinity(z)
            newe = e.subs(z, 1/z)
```
So currently what happens is as follows. We eliminate all pieces in between and only keep those pieces whose lower limit is `-oo` and upper limit is `oo` . So basically `expr` changes to `expr1` everytime while we calculate limit at `oo` and `-oo` . . The subs thingy still takes place in `limits.py` . Doesn't seem wrong though , atleast I couldn't find any exceptions out of the cases i tried
```
>>> expr = Piecewise((x - 3, x <= -5), ((x + 1)**2, x < 0),(1/x, x<10), (5, True))
>>> expr._at_infinity(x)
Piecewise((x - 3, x <= -5), (5, True))
>>> limit(expr, x, oo)
5
>>> limit(expr, x, -oo)
-oo
>>> expr1 = expr._at_infinity(x)
>>> limit(expr1, x, oo)
5
>>> limit(expr1, x, -oo)
-oo
```

sympy/functions/elementary/piecewise.py
anutosh491: Nope not necessarily first and last , it's basically taking out the pieces whose either lower boundary is `-oo` or upper boundary is `oo` . Check this out 
```
>>> p0 = Piecewise((0, Or(x < -2, x > 2)), (1, True))
>>> p0._at_infinity(x)
Piecewise((0, (x > 2) | (x < -2)))
```

sympy/functions/elementary/piecewise.py
jksuom: What will happen to a condition of the type `x < a` if `subs(x. 1/x)` is applied?

anutosh491: The test I've added would be a good example for this 
```
>>> expr2 = Piecewise((x - 3, x < 0), ((x + 1)**2, x < 10), (5, True))
>>> expr3 = expr2._at_infinity(x)
>>> expr3
Piecewise((x - 3, x < 0), (5, True))
>>> expr3.subs(x, 1/x)
Piecewise((-3 + 1/x, 1/x < 0), (5, True))
>>> newe = expr3.subs(x, 1/x)
>>> newe.leadterm(x, cdir = 1)
(5, 0)
>>> newe.leadterm(x, cdir = -1)
(1, -1)
```

jksuom: We want the condition `x < 0` for the first piece. `Piecewise((x - 3, x < 0), ((x + 1)**2, x < 10), (5, True))` is not a good example as there is `x < 0`. We should be prepared for `x < a` for any `a`.

anutosh491: I'm sure it would work for any a though ! This is what we want right ? cause the logic works correctly like this .
```
>>> expr = Piecewise((x - 3, x < -2), ((x + 1)**2, x < 10), (5, True))
>>> expr._at_infinity(x)
Piecewise((x - 3, x < -2), (5, True))
>>> expr = Piecewise((x - 3, x < 2), ((x + 1)**2, x < 10), (5, True))
>>> expr._at_infinity(x)
Piecewise((x - 3, x < 2), (5, True)) 
```
I hope I am understanding what your trying to convey . is this there anything wrong with the logic above ? ( only keeping pieces with either lower bound at `-oo `or those with upper bound at `oo` fllowed by `subs(x, 1/x)` in `limits.py` . Maybe you could give me an example of what you would like , cause I'm a bit blinded now as I don't see any case which the previous commit is not able to answer !

anutosh491:
ping @0sidharth for review !

sympy/functions/elementary/piecewise.py
jksuom: The final result should be a piecewise expression where 0 is a boundary point. That does not happen if `subs(x, 1/x)` is applied to Piecewise:
```
n [1]: e = Piecewise((x - 3, x < 2), (5, True))                                

In [2]: e.subs(x, 1/x)                                                          
Out[2]: 
⎧     1      1    
⎪-3 + ─  for ─ < 2  # this should be x < 0
⎨     x      x    
⎪                 
⎩  5     otherwise
```

anutosh491: Cool I'll maybe do this , though I still have my doubts regarding why do you think the code I pushed latest would fail to give expected results .I'll try again maybe to explain throughly I guess ? 

1) We get rid of all pieces which won't contribute to limit at `oo/-oo` using `_at_infinity` helper method. Hence after this our expression ` e = Piecewise((x - 3, x < -2), ((x + 1)**2, x < 10), (5, True))` would become `e = Piecewise((x - 3, x < 2), (5, True)) ` considering we eliminate the piece `((x + 1)**2, x < 10)` as it won't contribute anything in this case !

2) Now we only have two pieces and subs would never disappoint us in such cases .
`newe = e.subs(x, 1/x) `, where `e` is the updated `e` . Hence we have `newe  = Piecewise((-3 + 1/x, 1/x < 2), (5, True))` and as my code uses `._intervals(x)` , we have 
```
>>> (Piecewise((-3 + 1/x, 1/x < 2), (5, True)))._intervals(x)[1]
[(1/2, oo, -3 + 1/x, 0), (-oo, 0, -3 + 1/x, 0), (-oo, oo, 5, 1)]
```
This is correctly representing the intervals we want which is `-3 + 1/x` for `Union(Interval(1/2, oo), Interval(-oo, 0))`n and 5 for the rest which would eventually be `Interval(0, 1/2)` 
image

Hence from positive side (cdir = 1, which is basically the case for oo as cdir = -1, is reversed in `limits.py` by cdir = -cdir) we have the 5 case . This basically means `limit(5, x, 0, dir = '+') == limit(e, x, oo, dir = '-')`
And from the negative side (cdir = -1) we have the `-3 + 1/x` . This implies `limit(-3 + 1/x ,x, 0 , dir = '-') == limit(e, x, -oo, dir = '+') ` . 
And as we would always have only two pieces whenever we deal with limit at `oo` and `-oo` , we would always end up with a similar codeflow ,with one piece transformed on one side of `0` and another piece transformed ( `1/x` in place of `x`) on another side of `0`. I really don't see any case failing with this , maybe you could point something out if you have your doubts regarding certain expressions !

sympy/functions/elementary/piecewise.py
jksuom: > not necessarily first and last

I would first take all boundary points in ascending order:

    -oo < a_1 < a_2 < ... < a_n < oo

Then the first and last intervals will be `(-oo, a_1)` and `(a_n, oo)`. These are not necessarily the same as the intervals defined by the first and last `ExprCondPair`s of a `Piecewise` expression. Some processing is needed for finding the boundary points.

anutosh491: Yupp correct ! I realized that first we should always be covering the `(-oo, a_1)` interval and only then should we append the piece capturing `(a_n, oo)` . I'll make these changes asap !

anutosh491: Now limit for p works as expected for all points, wasn't working properly at `oo` before 
```
>>> p = Piecewise((2, x > 2), (0, x < -2), (1, True))
>>> limit(p, x, 4)
2
>>> limit(p, x, oo)
2
>>> limit(p, x, -oo)
0
>>> limit(p, x, 2)
2
>>> limit(p, x, -2)
1
>>> limit(p, x, 2, '-')
1
>>> limit(p, x, -2, '-')
0
```

anutosh491:
@jksuom I would be glad to have your views/suggestions on couple related things .

1) I've seen it multiple times on different issues and also on couple of my Pr's . There has been a suggestion that `cdir == 1/-1` should be changed to `cdir.is_positive`  and `cdir.is_negative` . Currently whatever issues i've tried to tackle have been revolving around `cdir` being eithe `0, 1, -1` so I still don't see the full potential of this change !

2) Also my latest commit makes sure all one directional limits would work as expected even if we manipulate the pieces back and forth  but I wanted to know what would be the apt position in `limits.py` and also the method for a bi-directional check . I have 1 in mind , maybe I could implement after having your views

jksuom:
> Currently whatever issues i've tried to tackle have been revolving around cdir being eithe 0, 1, -1

That is true. Currently. And probably it will be true even some time in the future so the code can probably be left as is for now. But  complex direction (`cdir`) is a work in progress and, in principle, it could take any complex value. It *could* be normalized to have absolute value 1 but it seems that it is more convenient to allow, for example, `1 + I` instead of always changing that to `(sqrt(2) + I*sqrt(2))/2` When leading term (and series expansion) methods are calling each other, they are allowed to modify `cdir` though that may not happen currently.

anutosh491:
>  Currently. And probably it will be true even some time in the future so the code can probably be left as is for now.

Makes complete sense , Would be glad if you could give your views on the second point (bi-directional limits for piecewise expressions) and the other changes too !

anutosh491:
The latest commit adds 
1) A check for bi-directional limits .for eg

```
>>> x = Symbol('x')
>>> expr = Piecewise((x - 3, x <= 0), ((x + 1)**2, x < 10), (5, True))
>>> limit(expr, x, 0, '+-')
    raise ValueError("The limit does not exist since "
ValueError: The limit does not exist since left hand limit = -3 and right hand limit = 1
>>> limit(expr, x, 10, '+-')
  File "C:\Users\anuto\sympy\sympy\sympy\series\limits.py", line 360, in doit
    raise ValueError("The limit does not exist since "
ValueError: The limit does not exist since left hand limit = 121 and right hand limit = 5
```
2) an exhaustive test for 6 piecewise functions which are all the same (1 from `x = (-2, 2)` and 0 otherwise) and those are all possible ways to represent the function using the Piecewise class . 
```
    p0 = Piecewise((0, Or(x < -2, x > 2)), (1, True))
    p1 = Piecewise((0, x < -2), (0, x > 2), (1, True))
    p2 = Piecewise((0, x > 2), (0, x < -2), (1, True))
    p3 = Piecewise((0, x < -2), (1, x < 2), (0, True))
    p4 = Piecewise((0, x > 2), (1, x > -2), (0, True))
    p5 = Piecewise((1, And(-2< x, x < 2)), (0, True))
```
I'm convinced the Pr would work perfectly fine for all finite x ! Once I confirm the working of code at `+-oo`, this can certainly be merged ! @jksuom

sympy/functions/elementary/piecewise.py
anutosh491: > There should be 2 pieces. The first condition should be `x < 0` in order to make 0 the single boundary point. The expressions should have `1/x` substituted for `x`. (After all, this is expected to be a simplified `subs(x, 1/x)`.)

Resolving conversation as my latest commit implements this !

anutosh491:
My latest commit improves the level of code involved to compute the function of a piecewise limit ! It gets a bit lenghty with checks sometimes because there are **so many ways** in which a piecewise function can be implemented in sympy ! So I had to write code which caters to each and all of these forms ! I've also added another test which would encounter all parts of the code 
```
def test_piecewise_basic3():
    p0 = Piecewise((0, x < -2), (2, x > 2), (1, True))
    p1 = Piecewise((2, x > 2), (0, x < -2), (1, True))
    p2 = Piecewise((0, x < -2), (1, x < 2), (2, True))
    p3 = Piecewise((2, x > 2), (1, x > -2), (0, True))
    Functions = [p0, p1, p2, p3]
    for function in Functions:
        assert limit(function, x, -oo) == 0
        assert limit(function, x, -2, '-') == 0
        assert limit(function, x, -2, '+') == 1
        raises(ValueError, lambda: limit(function, x, -2, dir='+-'))
        assert limit(function, x, 0, '+-') == 1
        assert limit(function, x, 2, '-') == 1
        assert limit(function, x, 2, '+') == 2
        raises(ValueError, lambda: limit(function, x, 2, dir='+-'))
        assert limit(function, x, oo) == 2
```
I am satisfied with the changes and will also explain the changes /code below ! I have tried out exhaustive cases not to miss any form of a piecewise expression , hence I believe it'll work fine .

anutosh491:
I'll explain one bit of the code and other things can be related.

**Overview :** 

- We first compute if 0 is the upper boundary or lower boundary of any function . This code explains the case when the upper boundary of a piece is 0 . 
- If a piece has 0 as it's upper boundary then we can make out that the leading term from the negative side would be the based on the function itself and the leading term from the positive side would be based on the immediate next peice . See figure in OP for reference(https://github.com/sympy/sympy/pull/22555#issue-1065032028)

**Step 1**: Line **3** helps us check which function's interval has an upper boundary of 0. This might be an interval or a Union of Intervals , hence we need to check for both .
**Step 2**: Line **5** returns the leading term based on the function which has an upper boundary of 0 (**Negative side**).
**Step 3**: Line **7 - 11** helps us compute the immediate piece following the one we found above . Now as I've written above , there being so many ways to write a piecewise function....... the immediate next piece may/may-not be next in line, hence we run a loop unless and untill the lower boundary of a piece is 0. Here we could encounter Unions too hence we have to deal with that too ! Now that we have found the function we return the leading term (**Positive Side**). 

```
        # Check whether 0 is the upper boundary of a piece.
        # If yes, we consider that piece and the next one for evaluation.
(1)        if S.Zero in upper_bound_list:
(2)            for i, pair in enumerate(expr_pairs):
(3)                if pair[1].sup == S.Zero or any([arg.sup.is_zero for arg in pair[1].args if isinstance(pair[1], Union)]):
(4)                   if cdir == -1:
(5)                        return pair[0].as_leading_term(x)
(6)                    else:
(7)                        while(expr_pairs[i + 1][1].inf != S.Zero):
(8)                           if isinstance(expr_pairs[i + 1][1], Union):
(9)                               if any([arg.inf.is_zero for arg in expr_pairs[i + 1][1].args
(10)                                        if isinstance(expr_pairs[i + 1][1], Union)]):
(11)                                    break
                            i = i + 1
                        return expr_pairs[i + 1][0].as_leading_term(x)
```

anutosh491:
ping @0sidharth @jksuom . I would be glad if you'll could have a look at the tests , the overall code and the comment(https://github.com/sympy/sympy/pull/22555#issuecomment-1030850348) above ! I am satisfied with the changes and can go in after set of reviews.

anutosh491:
ping @oscarbenjamin @oscargus @smichr  . Could you'll help me figure out why test/optional-dependencies fail . The Pr is based on a new implementation . It is not fixing any existing issue which would likely involve breaking any tests !
Also to my surprise the exceptions arise in `physics/quantum` module which I am not sure is closely related to the series module and that too to this implementation !
Some tests which fail . I honestly have no clue as of now which this might be happening !

```
test_sympy__physics__quantum__anticommutator__AntiCommutator E
test_sympy__physics__quantum__cartesian__PositionBra3D E
test_sympy__physics__quantum__cartesian__PositionKet3D E
test_sympy__physics__quantum__cartesian__PositionState3D E
test_sympy__physics__quantum__cartesian__PxBra E
test_sympy__physics__quantum__cartesian__PxKet E
test_sympy__physics__quantum__cartesian__PxOp E
test_sympy__physics__quantum__cartesian__XBra E
test_sympy__physics__quantum__cartesian__XKet E
test_sympy__physics__quantum__cartesian__XOp E
test_sympy__physics__quantum__cartesian__YOp E
test_sympy__physics__quantum__cartesian__ZOp E
test_sympy__physics__quantum__cg__CG E
test_sympy__physics__quantum__cg__Wigner3j E
test_sympy__physics__quantum__cg__Wigner6j E
test_sympy__physics__quantum__cg__Wigner9j E
test_sympy__physics__quantum__circuitplot__Mz E
test_sympy__physics__quantum__circuitplot__Mx E
test_sympy__physics__quantum__commutator__Commutator E
test_sympy__physics__quantum__constants__HBar E
test_sympy__physics__quantum__dagger__Dagger E
test_sympy__physics__quantum__gate__CGate E
test_sympy__physics__quantum__gate__CGateS E
test_sympy__physics__quantum__gate__CNotGate E
test_sympy__physics__quantum__gate__Gate E
test_sympy__physics__quantum__gate__HadamardGate E
test_sympy__physics__quantum__gate__IdentityGate E
test_sympy__physics__quantum__gate__OneQubitGate E
test_sympy__physics__quantum__gate__PhaseGate E
test_sympy__physics__quantum__gate__SwapGate E
test_sympy__physics__quantum__gate__TGate E
test_sympy__physics__quantum__gate__TwoQubitGate E
test_sympy__physics__quantum__gate__UGate E
test_sympy__physics__quantum__gate__XGate E
```

anutosh491:
ping @jksuom for reviews !

anutosh491:
cc: @0sidharth for reviews !

anutosh491:
ping @jksuom for reviews . What are you thoughts on this pr ? I feel this is good work and could go in after a set of reviews from your side . I have improved the code and added extensive tests for testing the new functionality/implementation leaving no room for error (obviously you can point out if you see any but currently I don't see this not answering anything incorrectly) . 

There have been some good reviews by you already in the past but haven't heard on it since a while after updating the code and making requested changes !

sympy/functions/elementary/piecewise.py
```diff
@@ -305,13 +306,115 @@ def doit(self, **hints):
...
+        from sympy.sets.sets import Union
+        if cdir == S.Zero:
+            l = self._eval_as_leading_term(x, logx=None, cdir=-1)
+            r = self._eval_as_leading_term(x, logx=None, cdir=1)
+            if l != r:
+                raise ValueError
+            return r
+
+        upper_bound_list, lower_bound_list = [], []
+        for a, b, e, i in self._intervals(x)[1]:
```
jksuom, line 359: `_intervals` may fail. That should be checked.

anutosh491:
> I don't think that bidirectional limits should be checked by low level methods like this. `cdir=0` should usually be treated in the same way as a positive `cdir`.

Yeah correct, I agree.... I knew that this isn't the best approach for this . I'll have to think how I could maybe approach this !

sympy/functions/elementary/piecewise.py
```diff
@@ -305,13 +306,115 @@ def doit(self, **hints):
...
+            upper_bound_list.append(b)
+            lower_bound_list.append(a)
```
jksuom, line 361: Why do you collect all bounds? It should suffice to take the first interval that contains the origin, possibly as an endpoint depending on `cdir`.

anutosh491: Yeah , taking all bounds isn't super necessary here , will try to improve on that !

anutosh491:
> I don't think that bidirectional limits should be checked by low level methods like this. cdir=0 should usually be treated in the same way as a positive cdir.

I tried to think of a method to tackle bi-directional limits but I'm not sure if  I got one .This is because it's only `cdir `as a parameter that gives us some clarity regarding the direction from which the leading term should be claculated
```
>>> expr2 = Piecewise((x - 3, x <= 0), ((x + 1)**2, x < 10), (5, True))
>>> limit(expr2, x, 0)  # enters method with cdir 1
>>> limit(expr2, x, 0, '+')  # enters method with cdir 1
>>> limit(expr2, x, 0, '-') # enters method with cdir -1
>>> limit(expr2, x, 0, '+-') # enters method with cdir 0
```
I went through some other `_eval_as_leading_term` methods to get some idea on this . And I realized that there are quite some functions(mostly those where there are significant differences in limits from both sides periodically) . Some prominent ones are `floor`, `ceil.` Even these take decision for the bi-directional check based on `cdir = 0`. The code snippet for `floor` dealing with bi-directional checks goes as follows
```
                if cdir == 0:
                    ndirl = arg.dir(x, cdir=-1)
                    ndir = arg.dir(x, cdir=1)
                    if ndir != ndirl:
                        raise ValueError("Two sided limit of %s around 0"
                                    "does not exist" % self)
```
We could maybe try an approach like this , not sure if this is what we need.
```
            if str(dir) == '+-':
                if isinstance(e, Piecewise):
                    r = limit(e, z, z0, '+')
                    l = limit(e, z, z0, '-')
                else:
                    r = gruntz(e, z, z0, '+')
                    l = gruntz(e, z, z0, '-')
```

sympy/functions/elementary/piecewise.py
anutosh491: Will try to address this too , I see `_intervals` can return `False` too , will return what we already have on master in that case.

anutosh491:
CC: @0sidharth
I am not sure if I am able to catch the best approach for including a bi-directional check as I've stated about . Could you maybe try suggesting what is a good approach here !

sympy/functions/elementary/piecewise.py
anutosh491: @jksuom is this the how we should be approaching this ..... cause everytime we encounter `False` for `self._intervals(x)[0]` , a message is returned . As the docstring says for _intervals , we have 
```
If there are any relationals not involving sym, or any
relational cannot be solved for sym, the bool will be False
a message be given as the second return value.
```
The message raised goes like this `'Unable to solve relational %s for %s.' % (r, sym)`
So should the following diff be used ! 
```
--- a/sympy/functions/elementary/piecewise.py
+++ b/sympy/functions/elementary/piecewise.py
@@ -357,6 +357,8 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
             return r

         zero_in_upper_bound, zero_in_lower_bound = None, None
+        if not self._intervals(x)[0]:
+            raise NotImplementedError("%s" % (self._intervals(x)[1]))
         for a, b, e, i in self._intervals(x)[1]:
             if b is S.Zero:
                 zero_in_upper_bound = True
```

anutosh491: Also as you had pointed out that the low level fix for the bi-directional case wasn't the best way to approach it . I had also replied to that some days back https://github.com/sympy/sympy/pull/22555#issuecomment-1047394766 and still I am not completely sure how can one go about this . So I wanted your guidance on this too , like how you could you maybe approach this ? If I get some hint of a workaround or so I'll surely try that out !

anutosh491: Another doubt I had was why haven't we considered having a separate number for bi-directional cases, like was it ever thought that we could have a sperate number or even `cdir = 0` reserved only for bi-directional cases . It might help us at multiple places !

Issue #22537, opened 2021-11-21
InconsistentAssumptions sympy error


jdmallo25:
My program crushes with error. (you can find code here: https://github.com/jdmallo25/beam_deflection_computing/blob/main/program.py)
It happens in different moments (100 iterations - 200 iterations), but error is always the same.
This is 80th iteration. I start the program like this (previously ):  python d.py -lbi -274 -rbi 0
I would like to know what I did wrong. Thank you!
`
[13:24:42]       [0]     solve(24.0, -776.0, 10, 0.001)
[13:24:42]       [1]     solve(24.0, -774.6666666666666, 10, 0.001)
[13:24:42]       [2]     solve(24.0, -773.3333333333334, 10, 0.001)
[13:24:42]       [3]     solve(24.0, -772.0, 10, 0.001)
[13:24:44]       [0]     iteration #0
[13:24:44]       [3]     iteration #0
[13:24:44]       [2]     iteration #0
[13:24:45]       [1]     iteration #0
Exception in thread 3:
Traceback (most recent call last):
  File "C:\Users\Andrey\AppData\Local\Programs\Python\Python38\lib\site-packages\sympy\core\assumptions.py", line 462, in getit
    return self._assumptions[fact]
KeyError: 'zero'
[13:24:53]       [2]     iteration #1

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Andrey\AppData\Local\Programs\Python\Python38\lib\site-packages\sympy\core\assumptions.py", line 462, in getit
    return self._assumptions[fact]
KeyError: 'extended_real'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Andrey\AppData\Local\Programs\Python\Python38\lib\site-packages\sympy\core\assumptions.py", line 462, in getit
    return self._assumptions[fact]
KeyError: 'finite'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Andrey\AppData\Local\Programs\Python\Python38\lib\site-packages\sympy\core\assumptions.py", line 462, in getit
    return self._assumptions[fact]
KeyError: 'integer'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Andrey\AppData\Local\Programs\Python\Python38\lib\threading.py", line 932, in _bootstrap_inner
    self.run()
  File "d.py", line 16, in run
[...]

Issue #16265, opened 2019-03-15
performance issue with random.shuffle call in assumptions


asmeurer:
In the core assumptions code, there is a call to `random.shuffle`. This was added in ba6e49fa0b7. https://github.com/sympy/sympy/blob/ce64896148f965a117619143aac2c320f7bd3a68/sympy/core/assumptions.py#L298

In trying to profile https://github.com/sympy/sympy/issues/16249, I found that it causes serious slowdowns. In master, with `x = symbols('x:50')` takes 2.3 seconds, but commenting out the shuffle line, it takes 1.75 seconds. The rest of the time is taken from other things, but this is appears to be a serious slowdown, since it's a core part of SymPy that is called many times. 

CC @smichr

ShubhamKJha:
This can be removed. The performance gain is important and the commit [ba6e49f](https://github.com/sympy/sympy/commit/ba6e49fa0b72a9ee93def4aca821cb3b77aaaede) suggests some fail safe to another issue rather than adding anything to the `_ask` function.

oscarbenjamin:
I don't think that random.shuffle particularly slows things down in normal usage. At least that should be demonstrated.

I have found that shuffling the prerequisites can help for detecting inconsistent rules in the assumptions system. The problem manifests itself as non-deterministic behaviour which is awkward but at least will be picked up by test code.

ShubhamKJha:
Hi @oscarbenjamin, can you explain the second part. I couldn't get it from the commit also.

oscarbenjamin:
When you query `is_x` if `_eval_is_x` returns None then prerequisites can be checked. Suppose that y->x and z-->!x then we can check `_eval_is_y` or `_eval_is_z` but which should we check first? This call to shuffle ensures that we don't always try the same thing first so if there is a bug and `_eval_is_y` and `_eval_is_z` are inconsistent with each other then the difference manifests in `is_x` giving non-deterministically different results. This will then show up in testing if you depend on `is_x` giving a particular result.

asmeurer:
I just confirmed the same 1.3x slowdown on another computer for `symbols('x:50')`. Putting some print statements there, the function seems to be called a lot. For instance, calling `sin(x)` caused the shuffle to be called 2642 times (fortunately only the first time because of the cache). The time to create `sin(x)` with the shuffle is 5.73 ms vs. 1.21 without it. That's 4.7x slower.

asmeurer:
[...]

PR #22557, opened 2021-11-28
Make the core assumptions fully deterministic

oscarbenjamin:

See e.g. here: https://github.com/sympy/sympy/issues/22537#issuecomment-975485692

Rework the `_ask` function in the core assumptions module.

The core assumptions system could be non-deterministic in some cases due
to the storing a key value of None in the _assumptions dict while
handling an assumptions query. For example in a query like e.is_zero the
assumptions system would:

1. Store None in self._assumptions['zero']
2. Look for and evaluate _eval_is_zero
3. If _eval_is_zero then other facts (is_integer, is_real, etc) are
checked.
4. While checking the other facts zero=None is still stored in the
_assumptions dict so any code that checks e.is_zero receives a None
value.
5. Possibly later the value of is_zero becomes known e.g. if is_real
turns out to be False than is_zero must be False.

The problem here is that in step 4 is_zero gives None even though a
fully resolved is_zero query should give e.g. False. Since the different
facts attempted in step 3 are tried in a random order this can lead to
non-deterministic behaviour.

The reason None is stored in self._assumptions['zero'] is that without
then a query like e.is_zero that occurs recursively while trying to
evaluate e.is_zero leads to infinite recursion. The None is stored there
to prevent the infinite recursion but that is really just masking bugs
and replacing the infinite recursion with non-deterministic behaviour.

With the change in this commit it is now necessary that no assumptions
query ever recursively queries an assumption on self. Mostly there is no
reason for assumptions queries to do that and in many cases the change
here just deletes a redundant `_eval_is_*` method that isn't actually
needed because the resolver in the assumptions system can figure out the
result itself.

The other situation where this could be problematic is where an
assumptions query calls higher-level code that should not really be used
within the assumptions system because it makes things circular. The
higher-level code should be free to query assumptions on e and so should
not be called as part of an assumptions query on e.

#### Release Notes

* core
    * A bug resulting in the core assumptions system being sometimes unreliable was fixed meaning that the assumptions system should now be fully reliable and deterministic including under multithreaded usage. The fix has a side effect that any code that is used as part of an `_eval_is_*` handler for an assumptions predicate (e.g. `_eval_is_real`) must not recursively query another assumption on the same object (e.g. by checking something like `self.is_positive`) or it will lead to infinite recursion. Any downstream code that implements such assumption handlers will need to be reviewed for bugs of this kind. Note that it isn't necessary to query something like `x.is_positive`` in a handler for `x.is_real` because the assumptions system already understands the relationship between these so if `x_eval_is_positive()` returns True then the system will conclude that `x.is_real` should be true.

### DIFF
--- a/sympy/core/assumptions.py
+++ b/sympy/core/assumptions.py
@@ -210,7 +210,7 @@ References
 
 """
 
-from .facts import FactRules, FactKB, InconsistentAssumptions
+from .facts import FactRules, FactKB
 from .core import BasicMeta
 from .sympify import sympify
 
@@ -510,57 +510,84 @@ def _ask(fact, obj):
     In all cases, when we settle on some fact value, its implications are
     deduced, and the result is cached in ._assumptions.
     """
+    # FactKB which is dict-like and maps facts to their known values:
     assumptions = obj._assumptions
+
+    # A dict that maps facts to their handlers:
     handler_map = obj._prop_handler
 
-    # Store None into the assumptions so that recursive attempts at
-    # evaluating the same fact don't trigger infinite recursion.
-    #
-    # Potentially in a multithreaded context it is possible that the
-    # assumptions query for fact was already resolved in another thread. If
-    # that happens and the query was resolved as True or False then
-    # assumptions._tell will raise InconsistentAssumptions because of the
-    # attempt to replace True/False with None. In that case it should be safe
-    # to catch the exception and return the result that was computed in the
-    # other thread.
-    #
-    # XXX: Ideally this call to assumptions._tell would be removed. Its purpose
-    # is to guard against infinite recursion if a query for one fact attempts
-    # to evaluate a related fact for the same object. However really this is
-    # just masking bugs because a query for a fact about obj should only query
-    # the properties of obj.args and not obj itself. This is not easy to change
-    # though because it requires fixing all of the buggy _eval_is_* handlers.
-    try:
-        assumptions._tell(fact, None)
-    except InconsistentAssumptions:
+    # This is our queue of facts to check:
+    facts_to_check = [fact]
+    facts_queued = {fact}
+
+    # Loop over the queue as it extends
+    for fact_i in facts_to_check:
+
+        # If fact_i has already been determined then we don't need to rerun the
+        # handler. There is a potential race condition for multithreaded code
+        # though because it's possible that fact_i was checked in another
+        # thread. The main logic of the loop below would potentially skip
+        # checking assumptions[fact] in this case so we check it once after the
+        # loop to be sure.
+        if fact_i in assumptions:
+            continue
+
+        # Now we call the associated handler for fact_i if it exists.
+        fact_i_value = None
+        handler_i = handler_map.get(fact_i)
+        if handler_i is not None:
+            fact_i_value = handler_i(obj)
+
+        # If we get a new value for fact_i then we should update our knowledge
+        # of fact_i as well as any related facts that can be inferred using the
+        # inference rules connecting the fact_i and any other fact values that
+        # are already known.
+        if fact_i_value is not None:
+            assumptions.deduce_all_facts(((fact_i, fact_i_value),))
+
+        # Usually if assumptions[fact] is now not None then that is because of
+        # the call to deduce_all_facts above. The handler for fact_i returned
+        # True or False and knowing fact_i (which is equal to fact in the first
+        # iteration) implies knowing a value for fact. It is also possible
+        # though that independent code e.g. called indirectly by the handler or
+        # called in another thread in a multithreaded context might have
+        # resulted in assumptions[fact] being set. Either way we return it.
+        fact_value = assumptions.get(fact)
+        if fact_value is not None:
+            return fact_value
+
+        # Extend the queue with other facts that might determine fact_i. Here
+        # we randomise the order of the facts that are checked. This should not
+        # lead to any non-determinism if all handlers are logically consistent
+        # with the inference rules for the facts. Non-deterministic assumptions
+        # queries can result from bugs in the handlers that are exposed by this
+        # call to shuffle. These are pushed to the back of the queue meaning
+        # that the inference graph is traversed in breadth-first order.
+        new_facts_to_check = list(_assume_rules.prereq[fact_i] - facts_queued)
+        shuffle(new_facts_to_check)
+        facts_to_check.extend(new_facts_to_check)
+        facts_queued.update(new_facts_to_check)
+
+    # The above loop should be able to handle everything fine in a
+    # single-threaded context but in multithreaded code it is possible that
+    # this thread skipped computing a particular fact that was computed in
+    # another thread (due to the continue). In that case it is possible that
+    # fact was inferred and is now stored in the assumptions dict but it wasn't
+    # checked for in the body of the loop. This is an obscure case but to make
+    # sure we catch it we check once here at the end of the loop.
+    if fact in assumptions:
         return assumptions[fact]
 
-    # First try the assumption evaluation function if it exists
-    try:
-        evaluate = handler_map[fact]
-    except KeyError:
-        pass
-    else:
-        a = evaluate(obj)
-        if a is not None:
-            assumptions.deduce_all_facts(((fact, a),))
-            return a
-
-    # Try assumption's prerequisites
-    prereq = list(_assume_rules.prereq[fact])
-    shuffle(prereq)
-    for pk in prereq:
-        if pk in assumptions:
-            continue
-        if pk in handler_map:
-            _ask(pk, obj)
-
-            # we might have found the value of fact
-            ret_val = assumptions.get(fact)
-            if ret_val is not None:
-                return ret_val
-
-    # Note: the result has already been cached
+    # This query can not be answered. It's possible that e.g. another thread
+    # has already stored None for fact but assumptions._tell does not mind if
+    # we call _tell twice setting the same value. If this raises
+    # InconsistentAssumptions then it probably means that another thread
+    # attempted to compute this and got a value of True or False rather than
+    # None. In that case there must be a bug in at least one of the handlers.
+    # If the handlers are all deterministic and are consistent with the
+    # inference rules then the same value should be computed for fact in all
+    # threads.
+    assumptions._tell(fact, None)
     return None
 
 
--- a/sympy/series/order.py
+++ b/sympy/series/order.py
@@ -259,7 +259,14 @@ class Order(Expr):
                                 e = expr.exp
                                 b = expr.base
                                 expr = exp(e * log(b))
-                    expr = expr.as_independent(*args, as_Add=False)[1]
+
+                    # It would probably be better to handle this somewhere
+                    # else. This is needed for a testcase in which there is a
+                    # symbol with the assumptions zero=True.
+                    if expr.is_zero:
+                        expr = S.Zero
+                    else:
+                        expr = expr.as_independent(*args, as_Add=False)[1]
 
                     expr = expand_power_base(expr)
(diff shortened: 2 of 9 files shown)


oscarbenjamin:
Looks like these changes to assumptions have broken this test:
https://github.com/sympy/sympy/blob/649cd163d9015289743d6f7132d69bc995a842b8/sympy/series/tests/test_limits.py#L950-L953
It leads into an infinite recursion at this line:
https://github.com/sympy/sympy/blob/649cd163d9015289743d6f7132d69bc995a842b8/sympy/core/add.py#L1071

oscarbenjamin:
There's something strange going on in `Add.as_leading_term`. For the failing test it only works on master because an exception is raised at line 1049 here:
https://github.com/sympy/sympy/blob/649cd163d9015289743d6f7132d69bc995a842b8/sympy/core/add.py#L1046-L1056
The exception is essentially this:
```python
In [2]: O(0) in O(x)
<ipython-input-2-81275f1954d0> in <module>
----> 1 O(0) in O(x)

~/current/sympy/sympy.git/sympy/series/order.py in __contains__(self, other)
    446         result = self.contains(other)
    447         if result is None:
--> 448             raise TypeError('contains did not evaluate to a bool')
    449         return result
    450 

TypeError: contains did not evaluate to a bool
```
I don't know this code so well so I find it hard to judge whether this is reasonable or not.

oscarbenjamin:
> Looks like these changes to assumptions have broken this test:

I have temporarily commented out this test because I'm not really sure what to do with it. It would be great if someone could help with that.

oscarbenjamin:
So it looks like everything passed except the test that I commented out for limits.

oscarbenjamin:
The problem for the commented out test is this:
```python
from sympy import *

p = Symbol('p', zero=True)
x = Symbol('x')
print(Order(p, x))
```
On master this gives `O(0)` but with this PR it gives `O(1)`.

oscarbenjamin:
That in turn is caused by this changing:
```python
from sympy import *

p = Symbol('p', zero=True)
x = Symbol('x')
print(p.as_independent(x, as_Add=False))
```
On master that gives `(0, 0)` whereas the PR gives `(p, 1)`.

oscarbenjamin:
This is ready for review. I've edited the OP.

oscarbenjamin:
This is still ready for review. Any input welcome.

ThePauliPrinciple:
I already gave this a couple of attempts, but it is very difficult to review for me.

I feel there are three types of changes in this PR:
- resolving the actual bug
- solving the issue that came from the proposed solution (no recursive assumption queries)
- some clean ups that were encountered but aren't actually involved in both of the above

Due to the complexity of the first point, it is difficult for me to confirm which change belongs to which group. I realise this is not very nice to ask, but would it be possible to split the PR in one with more trivial changes that would benefit SymPy without solving the underlying bug (e.g. in `_eval_is_extended_positive` you put a `fuzzy_not` to replace a handmade version of that, if I understand correctly) and the ones which actually solve the bug.

I don't think this hasn't gotten a review yet because no one is interested, but rather because is a difficult topic.

oscarbenjamin:
> I don't think this hasn't gotten a review yet because no one is interested, but rather because it is a difficult topic.

I realise that. I'm asking for reviews because I'm pretty sure that no one else knows the assumptions system as well as I do and I'm not sure that anyone can review these changes in detail. Rather than leaving this unreviewed I would prefer it if people could just say that so at least we are clear about it.

If necessary it can be merged without detailed review if other contributors are happy to take it partly on trust that I know what I'm doing with these changes. In many different contexts reviewing something does not mean that the reviewer should understand everything in full detail. It's a judgement call: based on your experience and what you do understand of this does it seem reasonable or not or are there any major concerns/uncertainties that you see?

> would it be possible to split the PR in one with more trivial changes that would benefit SymPy without solving the underlying bug

I could do that. Basically the change in `sympy/core/assumptions.py` is the main one. Everything else can be a single separate PR. The other changes could also be broken into smaller PRs to some extent but some of them are interdependent so in some cases it would be a lot of work to break this down smaller. Either way we would still be left with the changes to `assumptions.py` which would still be the difficult part to review.

ThePauliPrinciple:
> Rather than leaving this unreviewed I would prefer it if people could just say that so at least we are clear about it.

My thought exactly.

> Everything else can be a single separate PR. 

Having this confirmed actually really helped. I'm convinced about the non-`assumption.py` changes. The hermitian tests look right. The changes in `assumptions.py` are very well documented.

It would have been convenient if there could be a test to expose the previous issues that could arrise with multi-threading, but I don't think this is possible in the current CI (or is it?).

As for the compatibility issue, I don't think it is a big problem, but my concern is that the user might not get a useful error message when they do make recursive assumption calls. Is there any way to raise an explicit error which states that it is not allowed?

oscarbenjamin:
> It would have been convenient if there could be a test to expose the previous issues that could arrise with multi-threading, but I don't think this is possible in the current CI (or is it?).

It is possible to test multithreaded code in CI. The problem though is that the noticeable multithreaded issue was most likely already fixed but in a suboptimal way: https://github.com/sympy/sympy/issues/22537#issuecomment-975490896

Constructing examples which exhibit the nondeterminism is tricky because they need to be quite complicated. Basically you query predicate `obj.is_x`. Since there is no `obj._eval_is_x` or `obj._eval_is_x` returns None we then query all of the handlers for any predicates that could (perhaps in combination) imply `x`. These are called in a randomised order which is where the nondeterminism comes from so we might begin by querying `obj.is_y` or `obj.is_z`. Depending on the order we might find that an intermediate query resolves to `None` changing the final result. Here's an example:
```python
from sympy import *

class B(Expr):

    def _eval_is_extended_nonnegative(self):
        if self.is_finite is None:
            return
        return True

    def _eval_is_finite(self):
        if self.is_extended_positive is None:
            return
        return True

    def _eval_is_real(self):
        return True

try:
    print(B().is_nonnegative)
except Exception as e:
    print(type(e).__name__)
```
With master this will sometimes print True and sometimes print None. The determining factor is whether `_eval_is_real` is called before the other handlers. If it is then `real=True` implies `finite=True` and at that stage both of the other handlers would give True. If `_eval_is_finite` is called first then `is_finite=None` is stored in the assumptions dict so that `_eval_is_extended_nonnegative` would then return None and then `is_finite` is stored as None until `_eval_is_real` is called. At that point though it's too late to go back and call `_eval_is_extended_nonnegative` again.

The change with this PR is that for an example like that the cases where it would have hit a stored None potentially affecting the output are now just infinite recursion. So the above script will either print True or RecursionError, but it will never print None.

The title of this PR is perhaps misleading as things are still nondeterministic. The point is that if there are bugs of this type in the handlers that currently make them nondeterministic then this will expose those by having them manifest more visibly as RecursionError.

asmeurer:
I'm worried as always about the performance implications here. Just at a high level, it looks like this is replacing a single block of code with a loop. Also the `shuffle` call, which is already known to be a performance bottleneck (https://github.com/sympy/sympy/issues/16265), is now inside that loop. 

> The fix has a side effect that any code that is used as part of an _eval_is_* handler for an assumptions predicate (e.g. _eval_is_real) must not recursively query another assumption on the same object (e.g. by checking something like self.is_positive) or it will lead to infinite recursion. Any downstream code that implements such assumption handlers will need to be reviewed for bugs of this kind.

I think this could be clearer on why exactly that is a bug (and it should be emphasized that it's only recursive checking *on the same object* that is problematic). It took me a moment of thinking to realize it's because the system already knows about the relationship between `real` and `positive`, so a check in the handler is redundant. That's something that won't at all be obvious to anyone who don't understand how the assumptions system works internally (almost everyone).

Also is it possible to detect this somehow and raise an exception with a helpful message?

sympy/core/assumptions.py
```diff
@@ -510,57 +510,84 @@ def _ask(fact, obj):
...
+        # Extend the queue with other facts that might determine fact_i. Here
+        # we randomise the order of the facts that are checked. This should not
+        # lead to any non-determinism if all handlers are logically consistent
+        # with the inference rules for the facts. Non-deterministic assumptions
+        # queries can result from bugs in the handlers that are exposed by this
+        # call to shuffle. These are pushed to the back of the queue meaning
+        # that the inference graph is traversed in breadth-first order.
+        new_facts_to_check = list(_assume_rules.prereq[fact_i] - facts_queued)
+        shuffle(new_facts_to_check)
+        facts_to_check.extend(new_facts_to_check)
```
asmeurer, line 568: While this specific usage is technically fine, modifying a list while looping over it in general doesn't work as expected. It might be cleaner to make facts_to_check a `deque` and use a while loop.

asmeurer:
We should make sure to test this with the cache off. Otherwise many potential bugs would never be hit because the code would only be run once globally per assumptions query (I'm not clear, but the original bug may fall into that category).

oscarbenjamin:
> Just at a high level, it looks like this is replacing a single block of code with a loop.

It's replacing a recursive call with a loop.

oscarbenjamin:
> Also is it possible to detect this somehow and raise an exception with a helpful message?

I don't think so but I think that infinite recursion is a reasonable failure mode here. Note that infinite recursion would occur naturally here except there is a fudge that tries to hide it:
https://github.com/sympy/sympy/blob/7dbc2f49370b31ac6960524ea7e5444e2e5a50d5/sympy/core/assumptions.py#L516-L536

oscarbenjamin:
> That's something that won't at all be obvious to anyone who don't understand how the assumptions system works internally (almost everyone).

I can improve the release note but it isn't clear to me if or why anyone would have implemented their own assumptions handlers in downstream code. Most downstream code extends SymPy by adding new types of expression for which the assumptions predicates don't really make any sense. As you say "almost everyone" does not understand the internals here. It's just hypothetically possible that someone will have made a class that uses some `_eval_is_*` handlers since there's nothing to say whether that mechanism should be considered public or private.

asmeurer:
> It's replacing a recursive call with a loop.

I see now. That alone makes this better actually. Those recursive `_ask` calls are always clogging up tracebacks (esp. from Control-C), and loops are typically faster than recursion in Python anyway.

smichr:
> On master that gives `(0, 0)` whereas the PR gives `(p, 1)`.

The former is an evaluated form of the request, the latter an unevaluated form. They both mean the same thing in terms of result since `0*0` is equivalent to `p*1` when `p` is declared to be `zero=True`

sympy/series/order.py
```diff
@@ -256,7 +256,12 @@ def __new__(cls, expr, *args, **kwargs):
                                 expr = Mul(*[a.expr for a in orders])
                             elif expr.is_Pow:
                                 expr = orders[0].expr**orders[1].expr
-                    expr = expr.as_independent(*args, as_Add=False)[1]
+
+                    # XXX: There is probably a better place to check for this:
+                    if expr.is_zero:
+                        expr = S.Zero
+                    else:
+                        expr = expr.as_independent(*args, as_Add=False)[1]
```
smichr, line 264:
```suggestion
                    c, expr = expr.as_independent(*args, as_Add=False)
                    if c.is_zero:
                        expr = S.Zero
```

oscarbenjamin: What effect would this suggested change have?

oscarbenjamin: Note that these lines were already changed to something else on master. Let's see if there is still any need to change anything here.

oscarbenjamin:
I've fixed merge conflicts and added a bit to the release note.

I think this is good to merge. It's had a couple of reasonable reviews but otherwise just been sitting for a while (4 months) without any significant changes.

Please merge or object. Otherwise I will merge soon.

oscarbenjamin:
Okay, I'll merge this now.

PR #22562, opened 2021-11-29
Bugfix in `minimal_polynomial()` and `field_isomorphism()`.

### LINKED ISSUE #22561: Some functions misinterpret `AlgebraicNumber`s.
In certain cases, the functions `minimal_polynomial()` and `field_isomorphism()`
fail to operate on the numbers that are actually represented by instances of
`AlgebraicNumber`.

## Explanation

An `AlgebraicNumber` instance represents an element `alpha` of a number field `Q(theta)`. The primitive
element `theta` for the number field is represented by its minimal polynomial, and a particular root thereof.
The element `alpha` is represented as a polynomial in `theta`.

The `minimal_polynomial()` and `field_isomorphism()` functions are treating `AlgebraicNumber`
instances as if they represent not `alpha` but `theta`. In other words, these functions are operating
on the primitive element of the number field, not on the represented element itself.

## Example 1, `minimal_polynomial()`

```python
>>> from sympy import AlgebraicNumber, minimal_polynomial, sqrt, S
>>> from sympy.abc import x
>>> theta = sqrt(2) + sqrt(3)
>>> a = AlgebraicNumber(theta, [S(1) / 2, 0, S(-9) / 2, 0])
>>> a.as_expr()
sqrt(2)
```

The `AlgebraicNumber` `a` represents the number `sqrt(2)`, as an element of the number field `Q(theta)`.

(The coefficients `[1/2, 0, -9/2, 0]` in the definition of `a` correspond to the fact that `sqrt(2) = theta**3/2 - 9*theta/2`.)

But when we ask for the minimal polynomial of `a`, expecting `x**2 - 2`, we instead get the minimal polynomial for `theta`:

```python
>>> minimal_polynomial(a, x)
x**4 - 10*x**2 + 1
>>> minimal_polynomial(theta, x)
x**4 - 10*x**2 + 1
```

## Example 2, `field_isomorphism()`

```python
>>> from sympy import to_number_field, field_isomorphism, sqrt
>>> theta = sqrt(2) + sqrt(3)
>>> eta   = sqrt(2) + sqrt(5)
>>> a = to_number_field(sqrt(2), theta)
>>> b = to_number_field(sqrt(2), eta)
>>> a.as_expr()
sqrt(2)
>>> b.as_expr()
sqrt(2)
>>> print(field_isomorphism(a, b))
None
```

Here, `a` and `b` are `AlgebraicNumber`s representing the number `sqrt(2)` in the fields
`Q(theta)` and `Q(eta)`, respectively.

When we ask whether there is a field isomorphism from `Q(a)` to `Q(b)` we are asking
for an embedding from `Q(sqrt(2))` into `Q(sqrt(2))`. The answer should be `[1, 0]`,
meaning we can map `a` to `1*b + 0`. But instead we get `None`, because what `field_isomorphism()`
is actually telling us is that there is no embedding from `Q(theta)` into `Q(eta)` (which is true, but
it's not the question we asked).

skieffer:

Fixes #22561 

The `minimal_polynomial()` and `field_isomorphism()` functions were not always doing what they advertised to do.
This was due to a misapprehension of what an instance of the `AlgebraicNumber` class represents.

An `AlgebraicNumber` instance represents an element `alpha` of a number field `Q(theta)`. The primitive
element `theta` for the number field is represented by its minimal polynomial, and a particular root thereof.
The element `alpha` is represented as a polynomial in `theta`.

The `minimal_polynomial()` and `field_isomorphism()` functions were written as if an `AlgebraicNumber`
instance represented not `alpha` but `theta`. In other words, these functions operated on the primitive
element of the number field, not on the represented element itself.

We repair this as follows:
* Provide new docstrings for the `AlgebraicNumber` class, explaining what it represents and how.
* Provide new methods for the `AlgebraicNumber` class, making it easy to:
  - get the minimal polynomial for alpha
  - test whether alpha == theta
  - convert into an instance were alpha does equal theta
* Update the `minimal_polynomial()` and `field_isomorphism()` functions to operate on alpha, not theta.
* Provide new unit tests that failed before the fix.

#### Release Notes

* polys
  * Fixed related issues in `minimal_polynomial()` and `field_isomorphism()`.

### DIFF
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -2842,7 +2842,9 @@ class AlgebraicNumber(Expr):
         Say whether this algebraic number $\alpha \in \mathbb{Q}(\theta)$ is
         equal to the primitive element $\theta$ for its field.
         """
-        return self.coeffs() == [1, 0]
+        c = self.coeffs()
+        # Second case occurs if self.minpoly is linear:
+        return c == [1, 0] or c == [self.root]
 
     def primitive_elt(self):
         r"""
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -2545,24 +2545,47 @@ class AlgebraicNumber(Expr):
         Construct a new algebraic number $\alpha$ belonging to a number field
         $k = \mathbb{Q}(\theta)$.
 
+        There are four instance attributes to be determined:
+
+        ===========  ============================================================================
+        Attribute    Type/Meaning
+        ===========  ============================================================================
+        ``root``     :py:class:`~.Expr` for $\theta$ as a complex number
+        ``minpoly``  :py:class:`~.Poly`, the minimal polynomial of $\theta$
+        ``rep``      :py:class:`~sympy.polys.polyclasses.DMP` giving $\alpha$ as poly in $\theta$
+        ``alias``    :py:class:`~.Symbol` for $\theta$, or ``None``
+        ===========  ============================================================================
+
+        See Parameters section for how they are determined.
+
         Parameters
         ==========
 
-        expr : :py:class:`~.Expr`, or pair (m, r)
-            This defines the primitive element $\theta$ of the number field
-            $k$. If *expr* is an :py:class:`~.AlgebraicNumber`, then our
-            primitive element $\theta$ will be the same as that of *expr*.
-            If it is any other type of :py:class:`~.Expr`, then it itself will
-            be our primitive element. Therefore it must express an algebraic
-            quantity, and we will compute its minimal polynomial.
-            Otherwise *expr* must be an ordered pair
-            $(m, r)$ giving the minimal polynomial $m$, and a root $r$
-            thereof, which together define $\theta$. In this case $m$ may be
-            either a univariate :py:class:`~.Poly` or any :py:class:`~.Expr`
-            which represents the same, while $r$ must be some
-            :py:class:`~.Expr` representing a complex number that is a root of
-            $m$, including both explicit expressions, and instances of
-            :py:class:`~.ComplexRootOf`.
+        expr : :py:class:`~.Expr`, or pair $(m, r)$
+            There are three distinct modes of construction, depending on what
+            is passed as *expr*.
+
+            **(1)** *expr* is an :py:class:`~.AlgebraicNumber`:
+            In this case we begin by copying
+            all four instance attributes from *expr*. If *coeffs* were also
+            given, we compose the two coeff polynomials (see below).
+
+            **(2)** *expr* is any other type of :py:class:`~.Expr`:
+            Then ``root`` will equal *expr*. Therefore it
+            must express an algebraic quantity, and we will compute its
+            ``minpoly``.
+
+            **(3)** *expr* is an ordered pair $(m, r)$ giving the
+            minimal polynomial $m$, and a root $r$ thereof, which together
+            define $\theta$. In this case $m$ (which defines ``minpoly``) may
+            be either a univariate :py:class:`~.Poly` or any :py:class:`~.Expr`
+            which represents the same, while $r$ (which defines ``root``) must
+            be some :py:class:`~.Expr` representing a complex number that is a
+            root of $m$, including both explicit expressions in radicals, and
+            instances of :py:class:`~.ComplexRootOf`. If $r$ is an
+            :py:class:`~.AlgebraicNumber` it will be converted into another
+            type of :py:class:`~.Expr`, in order to be consistent with modes of
+            construction **(1)** and **(2)**.
 
         coeffs : list, :py:class:`~.ANP`, None, optional (default=None)
             This defines the algebraic number $\alpha$ as an element of $k$,
@@ -2571,7 +2594,10 @@ class AlgebraicNumber(Expr):
             If an :py:class:`~.ANP`, we take its coefficients (using its
             :py:meth:`~.ANP.to_list()` method). If ``None``, then the list of
             coefficients defaults to ``[1, 0]``, meaning that $\alpha = \theta$
-            is the primitive element of the field.
+            is the primitive element of the field. If *expr* was an
+            :py:class:`~.AlgebraicNumber`, let $g(x)$ be its coeff polynomial,
+            and let $f(x)$ be the polynomial given by *coeffs*. Then
+            ``self.rep`` will represent the composition $(f \circ g)(x)$.
 
         alias : str, :py:class:`~.Symbol`, None, optional (default=None)
             This is a way to provide a name for the primitive element. We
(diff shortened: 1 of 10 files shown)


sympy/core/numbers.py
```diff
@@ -2641,6 +2748,119 @@ def _eval_simplify(self, **kwargs):
...
+            return self
+        m = self.minpoly_of_elt()
+        r0 = self.n(prec)
+        closest_root, min_distance = None, oo
+        for r in m.all_roots():
+            d = abs(r0 - r.n(prec))
+            if d < min_distance:
+                min_distance = d
+                closest_root = r
+        return AlgebraicNumber((m, closest_root))
```
skieffer, line 2848: @oscarbenjamin Do you know if there's an existing function for this problem? I couldn't find one, so wrote an ad hoc solution.

You have a `Poly` `m`, and you have a number (in this case `self`), and you know that the number is equal to one of the roots of `m`. You want to determine which of the roots it is, and then you want to obtain an exact expression for that root (such as a `CRootOf`, or an explicit expression using radicals etc.).

oscarbenjamin: I'm not sure there is a function but one could be added to rootisolation.py. The key function is `dup_isolate_all_roots` which gives bounding intervals for each of the roots. Then evalf (i.e. .n) can tell us if our expression is in the bounding interval.

sympy/core/numbers.py
```diff
@@ -2509,9 +2509,24 @@ def __invert__(self):
...
+    Formally, the primitive element $\theta$ is given by two data points: (1)
+    its minimal polynomial (which defines $\mathbb{Q}(\theta)$), and (2) a
+    particular complex number that is a root of this polynomial (which defines
+    the embedding $\mathbb{Q}(\theta) \hookrightarrow \mathbb{C}$). Finally,
+    the algebraic number $\alpha$ which we represent is then given by the
+    coefficients of a polynomial in $\theta$.
+    """
 
-    __slots__ = ('rep', 'root', 'alias', 'minpoly')
+    __slots__ = ('rep', 'root', 'alias', 'minpoly', '_own_minpoly')
```
oscarbenjamin, line 2529: I don't think there is any real benefit in using `__slots__` for this class

sympy/core/numbers.py
oscarbenjamin: This functions is the only thing I'm not sure about

jksuom: Maybe this could be based on the same idea that is used in `_choose_factor`:
https://github.com/sympy/sympy/blob/551c16facb2081c7b7a085aeaf20316244a1da01/sympy/polys/numberfields/minpoly.py#L80-L86
All distances are computed and two least ones are compared. It they have different orders of magnitude (probably even less than `10**6` would suffice), then the minimum is chosen, otherwise the precision in incremented in a loop. There is no need to give a precision parameter as an argument.

skieffer: Wait...... First, thanks @oscarbenjamin and @jksuom for your ideas. I think you're on the right track for the problem
I asked to solve. However, maybe it's not needed here.

What we're talking about could go in a method called `to_CRootOf()`, which would turn an `AlgebraicNumber` into a `CRootOf` equal to the same complex number.

But I think all we need here is any `Expr` that can be `evalf`-ed to arbitrary precision, and `self` already is that.
So the whole method becomes:

```python
def to_primitive_elt(self):
    if self.is_primitive_elt:
        return self
    m = self.minpoly_of_elt()
    return AlgebraicNumber((m, self))
```

Sorry for the confusion! I didn't see this earlier.

skieffer: Cancel that. This turns out to be quite tricky.

## TL;DR

I think we were on the right track in the first place, and I'll push a new commit that does a better job of determining the root.

## Details

It seems that using one `AlgebraicNumber` as the `root` arg to another `AlgebraicNumber` is going to be very error prone and should probably be avoided.

As a demonstration of what can happen, let's start with an `AlgebraicNumber` `a` and its minpoly `m`:

```python
>>> from sympy import sqrt, to_number_field, AlgebraicNumber
>>> a = to_number_field(sqrt(2), sqrt(2) + sqrt(3))
>>> a.as_expr()
sqrt(2)
>>> m = a.minpoly_of_elt()
```

and use these as the `(m, r)` when constructing a new `AlgebraicNumber` `b`:

```python
>>> b = AlgebraicNumber((m, a))
>>> b.as_expr()
sqrt(2)
```

So far, so good.

But now, due to the funny way `AlgebraicNumber.__new__()` treats its first arg, when that arg is an `AlgebraicNumber`,

https://github.com/sympy/sympy/blob/519f1b75eda8830189b1385197f2c4fb49d333aa/sympy/core/numbers.py#L2541-L2542

`b` is not going to be picklable:

```python
>>> import pickle
>>> c = pickle.loads(pickle.dumps(b))
>>> c.as_expr()
sqrt(2) + sqrt(3)
```

Yipes!

But maybe we can repair pickling by changing this line:

https://github.com/sympy/sympy/blob/519f1b75eda8830189b1385197f2c4fb49d333aa/sympy/core/numbers.py#L2564

to

```python
        sargs = (expr, scoeffs)
```

which seems totally reasonable. Let our args be exactly as they were passed to us.

But this turns out to have too many far-flung ramifications.

For just one example, [this unit test](https://github.com/sympy/sympy/blob/519f1b75eda8830189b1385197f2c4fb49d333aa/sympy/polys/numberfields/tests/test_minpoly.py#L416-L422) is broken by it, because this method:

https://github.com/sympy/sympy/blob/519f1b75eda8830189b1385197f2c4fb49d333aa/sympy/polys/domains/gaussiandomains.py#L320-L323

expects the first arg of an `AlgebraicNumber` to always equal its root. In how many other places might similar expectations be embedded? This approach begins to look like much more trouble than it's worth, so I'll revert to the original approach.

sympy/core/numbers.py
skieffer: I'm not super well-versed in the pros and cons of declaring `__slots__`. Would you advise deleting the whole declaration?

oscarbenjamin: This is probably fine. I'm not sure that there is much benefit in using `__slots__` at all any more but maybe it's best just to leave it.

sympy/core/numbers.py
oscarbenjamin: I have to say that I've never quite understood how `AlgebraicNumber` works. Maybe it is not used consistently and should be fixed.

jksuom: It seems that an instance of `AlgebraicNumber` is practically always the same as the primitive element except in some tests. Maybe `coeffs` could be deprecated (and the tests removed).

skieffer: The possibility did cross my mind, but I think there are strong reasons for keeping it.

* A computer algebra system should have a class that represents "an element of a number field".
Such elements are usually thought of as polynomials in the primitive element, and the class should
provide things like printing support for displaying the element in this form. `AlgebraicNumber`
currently provides all this, but it will cease to if we give up `coeffs`.

* The `ANP` class does things the same way -- i.e. it represents the minimal polynomial of the
primitive element, plus the coeffs of the given element -- but this is on the `polys` side of things.
It is not an `Expr`.

* An `AlgebraicNumber` with `coeffs` is needed as the return value of the public `to_number_field()`
function.

* If we give up `coeffs`, then we just have a `minpoly` and a `root`, in which case we're essentially
the same thing as a `CRootOf`. (Which, to be honest, is a fine representation of an algebraic number for
certain applications -- but then it would be redundant.)

I guess those are the main reasons that I have considered.

Also, I'm still hoping to implement more things from Cohen. So, you're probably right that right now the library mostly treats these things as their primitive elements, but I'm hoping to add some cool stuff which will make an "element of a number field" class of this kind more important!

oscarbenjamin: I think that `AlgebraicNumber` should just be fixed to work properly.

oscarbenjamin: > But now, due to the funny way AlgebraicNumber.__new__() treats its first arg, when that arg is an AlgebraicNumber,

There `expr.coeffs` should be checked as well. That should be combined with the newly provided coeffs using e.g. `dup_compose`

skieffer: But isn't it a backwards compatibility issue? Otherwise I'd be all for changing the behavior here. (I think the special case should just be eliminated.)

oscarbenjamin: Can you give an example of what would be incompatible if this was just fixed?

skieffer:
I've implemented a `to_root()` method for `AlgebraicNumber`. This converts it into either a `CRootOf` or an expression in radicals, which is equal to it as a complex number.

We discussed two methods for this: (1) using isolating intervals, and (2) using the idea from `_choose_factor()`.

I like (1) because it dodges the "magic number" of method (2) (currently set as `10**6` in `_choose_factor()`), but I think we also need method (2) if we want to work with roots expressed in radicals.

So I've written a method that takes the best of both worlds. When working with only `CRootOf`, we use isolating intervals; if we have a root expressed in radicals, we use the other approach.

Note: This is independent of the pickling bug we're now discussing. I think we want this `to_root()` method in any case.

sympy/core/numbers.py
skieffer: I guess the real question I was raising was where we draw the line between breaking change and bugfix. But it dawns on me now that this entire special case of `AlgebraicNumber.__new__(expr, coeffs, ...)` where `expr` is an `AlgebraicNumber` could be regarded as a part of the same bug this PR sets out to fix; namely, places in the code where the `coeffs` of an `AlgebraicNumber` were forgotten/ignored.

So, I think the composing of polynomials here is clever, it brings the coeffs back into play, and it means that on unpickling we recover a number which is the same as a complex number, all of which is good. But unfortunately the internal structure still may differ. Example:

Adding this in `.__new__()`:

```python
        if expr.is_AlgebraicNumber:
            from sympy.polys.densetools import dup_compose
            c = dup_compose(rep.rep, expr.rep.rep, dom)
            rep = DMP.from_list(c, 0, dom)
```

we now get:

```python
>>> import pickle
>>> a = to_number_field(sqrt(2), sqrt(2) + sqrt(3))
>>> m = a.minpoly_of_elt()
>>> b = AlgebraicNumber((m, a))
>>> c = pickle.loads(pickle.dumps(b))
>>> for e in [b, c]:
>>>     print('--------')
>>>     print('As a complex number:         ', e.as_expr())
>>>     print('Our coeffs:                  ', e.rep)
>>>     print('Minpoly of primitive element:', e.minpoly)
>>>     print('Primitive element:           ', e.root)
>>> print('--------')
>>> print('Equal after pickling:', b == c)
--------
As a complex number:          sqrt(2)
Our coeffs:                   DMP([mpq(1,1), mpq(0,1)], QQ, None)
Minpoly of primitive element: PurePoly(_x**2 - 2, _x, domain='QQ')
Primitive element:            sqrt(2)
--------
As a complex number:          sqrt(2)
Our coeffs:                   DMP([mpq(1,2), mpq(0,1), mpq(-9,2), mpq(0,1)], QQ, None)
Minpoly of primitive element: PurePoly(_x**4 - 10*_x**2 + 1, _x, domain='QQ')
Primitive element:            sqrt(2) + sqrt(3)
--------
Equal after pickling: False
```

I'm thinking the fix could instead be as simple as just deleting the special case,

https://github.com/sympy/sympy/blob/52f606a503cea5e9588de14150ccb9f7f9ed4752/sympy/core/numbers.py#L2541-L2542

altogether. Then we do test as equal after pickling:

```python
--------
As a complex number:          -9*sqrt(2) + sqrt(3)/2 + sqrt(2) + sqrt(3)**3/2
Our coeffs:                   DMP([mpq(1,1), mpq(0,1)], QQ, None)
Minpoly of primitive element: PurePoly(_x**2 - 2, _x, domain='QQ')
Primitive element:            -9*sqrt(2) + sqrt(3)/2 + sqrt(2) + sqrt(3)**3/2
--------
As a complex number:          -9*sqrt(2) + sqrt(3)/2 + sqrt(2) + sqrt(3)**3/2
Our coeffs:                   DMP([mpq(1,1), mpq(0,1)], QQ, None)
Minpoly of primitive element: PurePoly(_x**2 - 2, _x, domain='QQ')
Primitive element:            -9*sqrt(2) + sqrt(3)/2 + sqrt(2) + sqrt(3)**3/2
--------
Equal after pickling: True

```

Note that the printing under `.as_expr()` is also changed in this case. This is because we are now succeeding in restoring `self.root` as an instance of `AlgebraicNumber`, which stays atomic in simplifications.

I think this is a good thing. If the old behavior is desired, then `b` can just be constructed a little differently:

```python
>>> b = AlgebraicNumber((m, a.to_root()))
>>> print(b.as_expr())
sqrt(2)
```

Here I'm using the `to_root()` method which I've now implemented, and have just pushed. In this case `a` is transformed into a `Pow`, which does participate in expression simplifications.

It's complex. I want to continue thinking about it a bit more...

oscarbenjamin: Okay, I see the problem now. Your examples use lots of complicated methods but it's actually just this:
```python
In [15]: a = AlgebraicNumber(sqrt(2), [1, 2])

In [16]: a
Out[16]: √2 + 2

In [17]: AlgebraicNumber(a)
Out[17]: √2
```
That's a clear bug to be fixed. Does the fix need to be much more complicated than this:
```diff
diff --git a/sympy/core/numbers.py b/sympy/core/numbers.py
index e87b039..93c07c8 100644
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -2536,7 +2536,7 @@ def __new__(cls, expr, coeffs=None, alias=None, **args):
                 from sympy.polys.polytools import Poly
                 minpoly = Poly(minpoly)
         elif expr.is_AlgebraicNumber:
-            minpoly, root = expr.minpoly, expr.root
+            minpoly, root, coeffs = expr.minpoly, expr.root, expr.coeffs()
         else:
             minpoly, root = minimal_polynomial(
                 expr, args.get('gen'), polys=True), expr
```
Could also check if coeffs was provided as well as an AlgebraicNumber and if so use `dup_compose`.

skieffer: Okay, this has the virtue of sticking closer to existing behavior, so I think it's good.

But while we're making `AlgebraicNumber(a)` behave more like a "copy constructor" when `a` is an `AlgebraicNumber`, we should probably also copy the `self.alias` from `a`, allowing this to be overridden by any `alias` that may have been passed.

Also, we need to modify the case where `expr` is an ordered pair `(m, r)`. If `r` is an `AlgebraicNumber` we have to convert it with `r = r.to_root()`. This (a) repairs pickling, and (b) makes this case consistent with the others.

I've pushed some code that implements this, and adapts the docs.

skieffer: Scratch the bit about `r = r.to_root()`. I tried it, but reverted it since it broke some things.

Anyway this was intended to address a separate issue -- pickling of `AlgebraicNumber` and `AlgebraicField` -- and I'd like to handle this in a separate PR.

skieffer:
I think this one might be ready now.

sympy/polys/rootisolation.py
```diff
@@ -1732,13 +1732,36 @@ def center(self):
...
     def as_tuple(self):
         """Return tuple representation of real isolating interval. """
         return (self.a, self.b)
 
     def __repr__(self):
         return "(%s, %s)" % (self.a, self.b)
 
+    def __contains__(self, item):
+        """
+        Say whether a complex number belongs to this interval.
```
oscarbenjamin, line 1749: Maybe this should clarify that it is a real interval.

sympy/polys/rootisolation.py
```diff
@@ -2002,6 +2031,24 @@ def conjugate(self):
         return ComplexInterval(self.a, self.b, self.I, self.Q,
             self.F1, self.F2, self.f1, self.f2, self.dom, conj=True)
 
+    def __contains__(self, item):
+        """
+        Say whether a complex number belongs to this interval.
```
oscarbenjamin, line 2036: Maybe this should clarify that it is a complex rectangular region. Is it common to use the word interval for this?

sympy/core/numbers.py
```diff
@@ -2526,11 +2541,180 @@ class AlgebraicNumber(Expr):
...
+            polynomial in $\theta$.
+
+            If a list, the elements should be integers or rational numbers.
+            If an :py:class:`~.ANP`, we take its coefficients (using its
+            :py:meth:`~.ANP.to_list()` method). If ``None``, then the list of
+            coefficients defaults to ``[1, 0]``, meaning that $\alpha = \theta$
+            is the primitive element of the field.
+
+            If *expr* was an :py:class:`~.AlgebraicNumber`, let $g(x)$ be its
+            ``rep`` polynoial, and let $f(x)$ be the polynomial defined by
```
oscarbenjamin, line 2598: typo: polynoial

sympy/core/numbers.py
```diff
@@ -2526,11 +2541,180 @@ class AlgebraicNumber(Expr):
...
+            ``minpoly``.
+
+            **(3)** *expr* is an ordered pair $(m, r)$ giving the
+            ``minpoly`` $m$, and a ``root`` $r$ thereof, which together
+            define $\theta$. In this case $m$ may be either a univariate
+            :py:class:`~.Poly` or any :py:class:`~.Expr` which represents the
+            same, while $r$ must be some :py:class:`~.Expr` representing a
+            complex number that is a root of $m$, including both explicit
+            expressions in radicals, and instances of
+            :py:class:`~.ComplexRootOf` or :py:class:`~.AlgebraicNumber`.
```
oscarbenjamin, line 2585: In future we should avoid making constructors that are this hard to explain. It's better to have methods like `AlgebraicNumber.from_poly_and_root(m, r)` rather than overload things like this.

sympy/core/numbers.py
```diff
@@ -2526,11 +2541,180 @@ class AlgebraicNumber(Expr):
...
+        Recall that we are constructing an algebraic number as a field element
+        $\alpha \in \mathbb{Q}(\theta)$.
+
+        >>> from sympy import AlgebraicNumber, sqrt, CRootOf, S
+        >>> from sympy.abc import x
+
+        Example (1): $\alpha = \theta = \sqrt{2}$
+
+        >>> a1 = AlgebraicNumber(sqrt(2))
+        >>> a1.minpoly_of_elt().as_expr(x)
```
oscarbenjamin, line 2623: I find the `minpoly_of_elt` method name confusing. I presume `elt` is short for "element" but it doesn't read that way to me. It comes across more like some acronym ELT that I've not heard of. I think it would be better to spell it out more as `minpoly_of_element` or otherwise come up with a different name. In SymPy "gen" is used a lot so it could be `minpoly_of_gen` or `minpoly_of_generator` but that might also be confusing since gen is usually used for polynomial variables.

oscarbenjamin: Actually looking further down at the docstring and code of `minpoly_of_elt` I'm really not sure I understand what the method does.

sympy/polys/rootisolation.py
skieffer: Ah, good point. No, I haven't heard the word used this way outside this context.

sympy/core/numbers.py
skieffer: Agreed.

sympy/core/numbers.py
oscarbenjamin: We should be careful about whether "element" means the primitive element of the algebraic number field or the element of the algebraic numberfield that is a defined in terms of that primitive element.

oscarbenjamin: Actually the new wording is careful about this

oscarbenjamin:
Again this is great work. Thanks!

PR #22535, opened 2021-11-20
core/basic: argument invariance for all Atom classes - lambda

ThePauliPrinciple:
#### References to other Issues or PRs

See https://github.com/sympy/sympy/pull/22522 for an alternative solution.

#### Brief description of what is fixed or changed
`Atom` classes now follow the invariance `atom.func(*atom.args) == atom`.
This is done by having `expr.func` return a `lambda` function.
Considering that SymPy currently uses `expr.func` as an alias for `type(expr)`,
several `expr.func` had to be changes to either `type(expr)` or something
along the lines of `isinstance(expr, <some class or tuple of classes>)`.

#### Release Notes

* core
  * `obj.func` no longer always returns `obj.__class__`, so use `type(obj)` instead.
  * All subclasses of `Atom` now follow the invariance `obj.func(*obj.args) == obj`.

PR #22563, opened 2021-11-29
basic: use type instead of func for dictionary keys

ThePauliPrinciple:

This would make https://github.com/sympy/sympy/pull/22535 safer to use.

Currently SymPy uses `obj.func` to obtain the classes of expressions
to use as keys in dictionaries. This might not be safe to do when
`obj.func` does not return `obj.__class__`. Instead, `type(obj)` can
be used safely. It is also more efficient.

#### Release Notes

NO ENTRY


oscarbenjamin:
Looks good!

PR #22564, opened 2021-11-30
liealgebras: subclass from atom instead of basic where appropriate

ThePauliPrinciple:

Solves part of the issues reported in https://github.com/sympy/sympy/pull/22445

The changed objects make sense as an `Atom` rather than a `Basic`
since their arguments should not be substitutable. Currently they
also store non-`Basic` arguments in `_args`, which should not be
allowed. Finally the changes also make the instances of these
classes pass the arg invariance test which was previously skipped
with `@XFAIL`.

#### Release Notes

NO ENTRY


oscarbenjamin:
Looks good. Thanks!

PR #22572, opened 2021-11-30
Min/Max now supports sets as argument

### LINKED ISSUE #22568: Max and Min of sets
It seems like the support for `Max` and `Min` over sets in very limited.

```
In [51]: Max(Range(5))

  File "C:\Users\Oscar\AppData\Local\Temp/ipykernel_3492/1594959331.py", line 1, in <module>
    Max(Range(5))

  File "C:\Users\Oscar\sympy\sympy\functions\elementary\miscellaneous.py", line 393, in __new__
    args = frozenset(cls._new_args_filter(args))

  File "C:\Users\Oscar\sympy\sympy\functions\elementary\miscellaneous.py", line 566, in _new_args_filter
    raise ValueError("The argument '%s' is not comparable." % arg)

ValueError: The argument 'Range(0, 5, 1)' is not comparable.
```

It may make sense to use `.sup` and `.inf` in case of sets, which will introduce at least basic support for `Range` and maybe some other sets.

oscargus:

Closes #22568

It is now possible to use sets as arguments to `Max` and `Min`. I needed a way to write something like

`Max(imageset(Lambda(x, f(x), Range(J))`

and later on replace J with an integer.

Although this is not yet supported as it doesn't seem possible to convert `Range` to `FiniteSet` at will, see #22571, it is a step in the right direction.

#### Release Notes

* functions
    * `Max` and `Min` now accept `Set` arguments.

### DIFF
--- a/sympy/functions/elementary/tests/test_miscellaneous.py
+++ b/sympy/functions/elementary/tests/test_miscellaneous.py
@@ -14,6 +14,9 @@ from sympy.functions.elementary.miscellaneous import (sqrt, cbrt, root, Min,
 from sympy.functions.elementary.trigonometric import cos, sin
 from sympy.functions.special.delta_functions import Heaviside
 
+from sympy.sets.sets import FiniteSet
+from sympy.sets.fancysets import Range
+
 from sympy.utilities.lambdify import lambdify
 from sympy.testing.pytest import raises, skip, ignore_warnings
 
@@ -449,6 +452,7 @@ def test_rewrite_as_Abs():
     test(Min(x, y, z))
     test(Min(Max(w, x), Max(y, z)))
 
+
 def test_issue_14000():
     assert isinstance(sqrt(4, evaluate=False), Pow) == True
     assert isinstance(cbrt(3.5, evaluate=False), Pow) == True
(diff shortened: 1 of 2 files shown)


sympy/functions/elementary/tests/test_miscellaneous.py
```diff
@@ -461,12 +465,27 @@ def test_issue_14000():
...
+
 def test_issue_6899():
     from sympy.core.function import Lambda
     x = Symbol('x')
     eqn = Lambda(x, x)
     assert eqn.func(*eqn.args) == eqn
 
+
+def test_MinMax_with_Sets():
+    assert Max(Range(3)) == 2
```
smichr, line 477: I think this might be conflating ideas of objects and object elements. In a similar way we needed to distinguish between `FiniteSet(FiniteSet(1, 2)) != FiniteSet(1, 2)`. When the user writes `Max(a, b)` then `a` and `b` must be compared, not the elements within. If there is code that needs to find the largest item in sets then it should be `max(i.sup for i in sets)`.

smichr: I think `max(Range)` is ok but have doubts about `max(Range, FiniteSet)`. Note how the builtin behaves:
```python
>>> max([1,2,3],[2])
[2]
```
So it returned the element that had the largest first difference. And mixed comparisons fail:
```python
>>> max([1,2,3],[2],7)
TypeError: '>' not supported between instances of 'int' and 'list'
```

oscargus:
OK! I can see the point of not allowing `Max(set1, set2)` (as there is no way to compare sets afaIk). However, from a mathematical perspective, the `MaxSet` function is, well, max. So it will print exactly like max (although ideally for an ImageSet with the domain below the max).

So I think that it does make sense to have a "max"-function which allows a single set (and, hence, takes the maximum set member). If that then leads to a different class being executed, so be it. More work for limited benefit.

(This means that while `Max(set1, set2)` is not allowed but `Max(Max(set1), Max(set2))` is and `set1.sup` can return `Max(set1)` if nothing else is available.)

The whole point is to be able to manipulate these types of expressions without having to do additional coding, but to benefit from the built-in support. As an added benefit it will generate mathematically correct/valid output for intermediate steps.

smichr:
What about `max(Union(*sets))` or `max(i.sup for i in sets)`? I pefer the latter. And even though you don't know of a meaningful way to compare sets, Python has chosen a way to compare iterables. Since we are in Python, we might aim for the 'least surprise' result.

smichr:
Note, too, that FiniteSet can hold FiniteSets, so even `max(FiniteSet)` would have to decide what the biggest element is when comparing the two sets, for example, in `FiniteSet({1,2},{3,4})`. It just doesn't seem like `max(*sets)` makes sense; and at least the example I gave recognizes that one can't compare the two sets directly and raises a Value Error when requesting the `.sup` of that set of sets.

oscarbenjamin:
I think it would be a mistake to allow the type of the args for Min/Max to be confused like this. If there should be something to represent the maximum of a set then it should be a separate expression type.

PR #22575, opened 2021-11-30
update find docstring

smichr:
Also, make the `missing` a symbol that won't need to be sympified
since trying to sympify '?' will fail. The code currently does not
encounter this issue, so this is just a safety precaution in case
code is changed that will require it to be sympified.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/geometry/util.py
+++ b/sympy/geometry/util.py
@@ -39,7 +39,10 @@ def find(x, equation):
     x
     >>> find('x', Dummy('x'))
     _x
-    >>> Dummy('x').name == 'x'
+
+    The dummy symbol is returned since it has a matching name:
+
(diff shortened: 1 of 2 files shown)


sympy/geometry/util.py
```diff
@@ -23,9 +23,28 @@
...
+    ========
+
+    >>> from sympy.geometry.util import find
+    >>> from sympy import Dummy
+    >>> from sympy.abc import x
+    >>> find('x', x)
+    x
+    >>> find('x', Dummy('x'))
+    _x
+    >>> Dummy('x').name == 'x'
```
smichr, line 42:
```suggestion

    The dummy symbol is returned since it has a matching name:

    >>> _.name == 'x'
```

PR #22576, opened 2021-11-30
Improved time for generating truth tables in simplify_logic

oscargus:

`simplify_logic` is faster by improving the truth table generation.

The largest test, `test_issue_14700` takes about 2.5 s to execute on my computer. With this it takes about 0.5 s.

#### Other comments

I also added a bit of code from #17330 which replaces e.g. `x < 0` and `x >= 0` with a single variable (one is `not` the other). This primarily helps avoiding an issue where one of them may become rewritten to the other under certain circumstances and then the sequential replacement as is done here may miss one of them. This also gives an additional benefit of automatically simplifying expressions like `Or(x < 0, x >= 0)` without relying on the pattern-based simplification and that the limit of eight variables is enforced a bit more rarely (earlier these were two variables, now one).

#### Release Notes

* logic
   * `simplify_logic` is faster.

### DIFF
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -2913,9 +2913,9 @@ def simplify_logic(expr, form=None, deep=True, force=False):
             var = variables.pop()
             undo[d] = var
             repl[var] = d
-            nvar = ~var
+            nvar = var.negated
             if nvar in variables:
-                repl[nvar] = ~d
+                repl[nvar] = d.negated
                 variables.remove(nvar)
 
     expr = expr.xreplace(repl)
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -657,7 +657,7 @@ class BooleanFunction(Application, Boolean):
                 # Remove the old relationals
                 del Rel[j]
                 del Rel[i]
-                if dominatingvalue is None or newrel != ~dominatingvalue:
+                if dominatingvalue is None or newrel != Not(dominatingvalue):
                     # Insert the new one (no need to insert a value that will
                     # not affect the result)
                     Rel.append(newrel)
@@ -1015,30 +1015,30 @@ class Not(BooleanFunction):
         func, args = expr.func, expr.args
 
         if func == And:
-            return Or._to_nnf(*[~arg for arg in args], simplify=simplify)
+            return Or._to_nnf(*[Not(arg) for arg in args], simplify=simplify)
 
         if func == Or:
-            return And._to_nnf(*[~arg for arg in args], simplify=simplify)
+            return And._to_nnf(*[Not(arg) for arg in args], simplify=simplify)
 
         if func == Implies:
             a, b = args
-            return And._to_nnf(a, ~b, simplify=simplify)
+            return And._to_nnf(a, Not(b), simplify=simplify)
 
         if func == Equivalent:
-            return And._to_nnf(Or(*args), Or(*[~arg for arg in args]),
+            return And._to_nnf(Or(*args), Or(*[Not(arg) for arg in args]),
                                simplify=simplify)
 
         if func == Xor:
             result = []
             for i in range(1, len(args)+1, 2):
                 for neg in combinations(args, i):
-                    clause = [~s if s in neg else s for s in args]
+                    clause = [Not(s) if s in neg else s for s in args]
                     result.append(Or(*clause))
             return And._to_nnf(*result, simplify=simplify)
 
         if func == ITE:
             a, b, c = args
-            return And._to_nnf(Or(a, ~c), Or(~a, ~b), simplify=simplify)
+            return And._to_nnf(Or(a, Not(c)), Or(Not(a), Not(b)), simplify=simplify)
 
         raise ValueError("Illegal operator %s in expression" % func)
 
@@ -1145,7 +1145,7 @@ class Xor(BooleanFunction):
         args = []
         for i in range(0, len(self.args)+1, 2):
             for neg in combinations(self.args, i):
-                clause = [~s if s in neg else s for s in self.args]
+                clause = [Not(s) if s in neg else s for s in self.args]
                 args.append(Or(*clause))
         return And._to_nnf(*args, simplify=simplify)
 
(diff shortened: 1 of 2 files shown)


sympy/logic/boolalg.py
```diff
@@ -2898,21 +2901,52 @@ def simplify_logic(expr, form=None, deep=True, force=False):
...
+        from sympy.core.symbol import Dummy
+        while variables:
+            d = Dummy()
+            var = variables.pop()
+            undo[d] = var
+            repl[var] = d
+            nvar = ~var
+            if nvar in variables:
+                repl[nvar] = ~d
+                variables -= set([nvar])
```
smichr, line 2919:
```suggestion
                variables.remove(nvar)
```

sympy/logic/boolalg.py
```diff
@@ -2898,21 +2901,52 @@ def simplify_logic(expr, form=None, deep=True, force=False):
     # get variables in case not deep or after doing
     # deep simplification since they may have changed
     variables = _find_predicates(expr)
+    # Replace variables with Dummys in case there are Relationals to possibly
+    # reduce the number of variables
+    repl = dict()
+    undo = dict()
+    from sympy.core.relational import Relational
```
smichr, line 2908: Can this import and the one a few lines down get moved out of the function or at least to the beginning? I don't fully understand the motivations to where they are located and appreciate wanting to keep them near the code that needs them...but just checking.

oscargus: Not outside due to circular imports. Of course they can go to the top of the function, but I thougt I'd keep them nearby (Dummy is not needed unless in the if.)

oscargus: If there are inlined imports, I usually try to import them where needed if there are returns before they are needed. If not, at the top of the method/function.

If only needed inside an if-statement I try to import them there.

(Unless the import ends up in a loop.)

sympy/logic/boolalg.py
```diff
@@ -2898,21 +2901,52 @@ def simplify_logic(expr, form=None, deep=True, force=False):
...
+    if expr.has(Relational):
+        from sympy.core.symbol import Dummy
+        while variables:
+            d = Dummy()
+            var = variables.pop()
+            undo[d] = var
+            repl[var] = d
+            nvar = ~var
```
oscargus, line 2916:
```suggestion
            nvar = var.negated
```

sympy/logic/boolalg.py
```diff
@@ -2898,21 +2901,52 @@ def simplify_logic(expr, form=None, deep=True, force=False):
...
+            if nvar in variables:
+                repl[nvar] = ~d
```
oscargus, line 2918:
```suggestion
                repl[nvar] = d.negated
```

sympy/logic/boolalg.py
ThePauliPrinciple: Can this potentially be done with "expr.is_Relational"?

sympy/logic/boolalg.py
```diff
@@ -2898,21 +2901,52 @@ def simplify_logic(expr, form=None, deep=True, force=False):
...
+    if expr.has(Relational):
+        from sympy.core.symbol import Dummy
+        while variables:
+            d = Dummy()
+            var = variables.pop()
+            undo[d] = var
+            repl[var] = d
+            nvar = var.negated
+            if nvar in variables:
+                repl[nvar] = d.negated
```
oscargus, line 2918:
```suggestion
                repl[nvar] = ~d
```

sympy/logic/boolalg.py
oscargus: You right. I'll give it a try!

oscargus, line 2908:
```suggestion
```

sympy/logic/boolalg.py
```diff
@@ -2898,21 +2901,52 @@ def simplify_logic(expr, form=None, deep=True, force=False):
     # get variables in case not deep or after doing
     # deep simplification since they may have changed
     variables = _find_predicates(expr)
+    # Replace variables with Dummys in case there are Relationals to possibly
+    # reduce the number of variables
+    repl = dict()
+    undo = dict()
+    from sympy.core.relational import Relational
+    if expr.has(Relational):
```
oscargus, line 2909:
```suggestion
    if expr.is_Relational:
```

sympy/logic/boolalg.py
oscargus: Ahh, no, it cannot. Since expr is the logic expression: 

```
And(x, y > 0).is_Relational
Out[4]: False
```

ThePauliPrinciple: Would something like this be suitable:
`any([expr.is_Relational for expr in preorder_traversal(And(x, y > 0))])`

ThePauliPrinciple: Or does that also cause circular imports?

In the end I suppose the issue really is that sympy.core heavily makes use of `Boolean`, but `Boolean` also shares the same source file as functions that use sympy.core.

oscargus:
Motivation for the ~-commit:

```
%timeit a = ~x
1.64 µs ± 167 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit a = Not(x)
1.14 µs ± 113 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

sympy/logic/boolalg.py
oscargus: I think it may work (just quickly looking at the import patterns), but not really sure if it is worth it compared to having a more directly readable code.

Yes, parts of boolalg (Boolean and And, Or etc) should probably move to core. I made a quick attempt some time back and if I recall correctly it was still quite feasible. However, I am not sure it actually solves the circular import since relational will have to import from boolalg (or whatever that should be called) and vice versa.

oscargus: Also, there are import from relational in quite a few functions and methods in boolalg, so from that perspective it is maybe not very beneficial either.

oscargus: An option, after browsing the other code, is to use self.atoms(Relational) for the initial set. Will not help with the import issue, but simplify the code otherwise. (As we are only interested in replacing the Relationals.)

ThePauliPrinciple: It would be nice if `And(x, y > 0).find(lambda expr: expr.is_Relational)` had an alias like 
`And(x, y > 0).find(is_Relational = True)`

PR #22578, opened 2021-11-30
powsimp recognizes non-symbolic negated exponents

smichr:
If an exponent in powsimp is a sum from which a -1 can be
factored then if that exponent is an integer or the base on which
it occurs is positive, then `b**e == (1/b)**-e` and this can be used to combine
bases that were previously not combined, so now
```python
>>> i = Symbol('i', integer=True)
>>> powsimp(x**(i + 1)/y**(i + 1))
(x/y)**(i + 1)

>>> x, y = symbols('x y', positive=True)
>>> from sympy.abc import z
>>> powsimp(x**(z + 1)/y**(z + 1))
(x/y)**(z + 1)
```

#### Release Notes

* simplify
    * `powsimp` now recognizes any negated exponent that is an integer or that occurs on positive bases


oscarbenjamin:
Looks good to me

PR #22580, opened 2021-12-01
core/containers: Dict now stores arguments as `Basic`

### LINKED ISSUE #22579: Dict storing key,value pairs as non-Basic
Currently, `Dict` stores its `key,value` pairs as `frozenset`. I believe this might be a typo:
https://github.com/sympy/sympy/blob/ed0bfd1e47345c8bd45ef542eef18f6a21c55685/sympy/core/containers.py#L227-L238
line 235 should probably read:
```python
        obj = Basic.__new__(cls, *elements)
```
Although it might be intended for efficiency reasons.

ThePauliPrinciple:

Fixes #22579 

Currently `Dict` stores its `key,value` pairs inside a `frozenset` which is not a `Basic`, so should not be stored in `_args`. Note that `args` already provides the correct interface, so this behaviour might be acceptable. 

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/containers.py
+++ b/sympy/core/containers.py
@@ -11,7 +11,7 @@ from collections.abc import MutableSet
 from typing import Any, Callable
 
 from .basic import Basic
-from .sorting import default_sort_key
+from .sorting import default_sort_key, ordered
 from .sympify import _sympify, sympify, converter, SympifyError
 from sympy.utilities.iterables import iterable
 from sympy.utilities.misc import as_int
@@ -233,7 +233,7 @@ class Dict(Basic):
         else:
             raise TypeError('Pass Dict args as Dict((k1, v1), ...) or Dict({k1: v1, ...})')
         elements = frozenset(items)
-        obj = Basic.__new__(cls, elements)
+        obj = Basic.__new__(cls, *ordered(items))
         obj.elements = elements
         obj._dict = dict(items)  # In case Tuple decides it wants to sympify
         return obj
@@ -249,17 +249,6 @@ class Dict(Basic):
 
     def __setitem__(self, key, value):
         raise NotImplementedError("SymPy Dicts are Immutable")
-
-    @property
-    def args(self):
-        """Returns a tuple of arguments of 'self'.
-
-        See Also
-        ========
-
-        sympy.core.basic.Basic.args
-        """
-        return tuple(self.elements)
 
     def items(self):
         '''Returns a set-like object providing a view on dict's items.


ThePauliPrinciple:
Seems like the `frozenset` was used to ensure the order doesn't matter during comparison. I'm not sure why it only fails in pypy though.

smichr:
Can you add a test that fails in master but not with this change?

ThePauliPrinciple:
> Can you add a test that fails in master but not with this change?

No. The current code works correctly, however, it uses `frozenset` as an argument to `Basic.__new__`, thus storing a non `Basic` object in `_args`. But it seems that this is not trivial to solve. Perhaps it should be allowed since it has a custom `args` defined?

oscarbenjamin:
Maybe the args should sorted using `ordered` and then `_hashable_content` would not need to be defined.

ThePauliPrinciple:
I decided to try to keep the `frozenset` sorting over `_sorted_args` so that it would remain consistent with previous versions.

oscarbenjamin:
> I decided to try to keep the `frozenset` sorting over `_sorted_args` so that it would remain consistent with previous versions.

Do we not need the sorting to be consistent from one run to another?
```console
$ cat t.py 
print(tuple(frozenset(['asd', 'qwe', 'zxc'])))
$ python t.py 
('asd', 'zxc', 'qwe')
$ python t.py 
('zxc', 'qwe', 'asd')
```

sympy/core/containers.py
```diff
@@ -231,8 +231,9 @@ def __new__(cls, *args):
             items = [Tuple(k, v) for k, v in args]
         else:
             raise TypeError('Pass Dict args as Dict((k1, v1), ...) or Dict({k1: v1, ...})')
+        items = tuple(sorted(items, key=default_sort_key))
```
oscarbenjamin, line 234: I think that `ordered` might be more efficient than directly using `sort` with `default_sort_key`.

ThePauliPrinciple: I am not familiar with what you are refering to with `ordered`

oscarbenjamin: https://github.com/sympy/sympy/blob/5dc2df6aa7b68d42d39ab59a1b6b50f40ebebd19/sympy/core/sorting.py#L201-L210

ThePauliPrinciple:
@oscarbenjamin could you give this an attempt?

sympy/core/containers.py
```diff
@@ -231,8 +231,9 @@ def __new__(cls, *args):
             items = [Tuple(k, v) for k, v in args]
         else:
             raise TypeError('Pass Dict args as Dict((k1, v1), ...) or Dict({k1: v1, ...})')
+        items = ordered(items)
         elements = frozenset(items)
-        obj = Basic.__new__(cls, elements)
+        obj = Basic.__new__(cls, *items)
         obj.elements = elements
         obj._dict = dict(items)  # In case Tuple decides it wants to sympify
```
oscarbenjamin, line 238: Since `ordered` returns a one-shot iterable it will already be consumed before we reach this line giving:
```python
In [1]: Dict({x:1})._dict
Out[1]: {}
```

ThePauliPrinciple:
There is currently still this error:

```
_____________________________________________________________________________________________________________________________________________________
______________________________________________ sympy/categories/tests/test_drawing.py:test_DiagramGrid ______________________________________________
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/categories/tests/test_drawing.py", line 482, in test_DiagramGrid
    grid = DiagramGrid(d, {
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/categories/diagram_drawing.py", line 1189, in __init__
    if groups and (groups != diagram.objects):
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/basic.py", line 375, in __ne__
    return not self == other
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/basic.py", line 348, in __eq__
    other = _sympify(other)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/sympify.py", line 509, in _sympify
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/sympify.py", line 366, in sympify
    return conv(a)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/containers.py", line 298, in <lambda>
    converter[dict] = lambda d: Dict(*d.items())
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/containers.py", line 231, in __new__
    items = [Tuple(k, v) for k, v in args]
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/containers.py", line 231, in <listcomp>
    items = [Tuple(k, v) for k, v in args]
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/containers.py", line 53, in __new__
    obj = Basic.__new__(cls, *args)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/containers.py", line 52, in <genexpr>
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/sympify.py", line 366, in sympify
    return conv(a)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/containers.py", line 298, in <lambda>
    converter[dict] = lambda d: Dict(*d.items())
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/containers.py", line 235, in __new__
    obj = Basic.__new__(cls, *tuple(ordered(items)))
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/sorting.py", line 306, in ordered
    yield from d[k]
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/sorting.py", line 288, in ordered
    d[f(a)].append(a)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/sorting.py", line 192, in _nodes
    return _node_count(e)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/sorting.py", line 175, in _node_count
    return 1 + sum(map(_node_count, e.args))
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/sorting.py", line 175, in _node_count
    return 1 + sum(map(_node_count, e.args))
TypeError: 'property' object is not iterable
```

ThePauliPrinciple:
Looks like this error is similar to #22582: it is a problem which arrises from `sympify` being called in `Tuple` (due to an __eq__ and in this case `Dict` using `Tuple`). In this case, it converts the string `"transpose"` to the object `transpose`, whose `.args` is indeed a `property` object.

oscarbenjamin:
The fact that `sympify` tries to parse strings is a strong reason for why it shouldn't be used in internal sympification.

ThePauliPrinciple:
I don't think we want something like this:
`assert Dict({transpose:'foo'}) == {'transpose': 'foo'}`

oscarbenjamin:
> I don't think we want something like this:

Is that in the tests somewhere?

ThePauliPrinciple:
No, but I just realised that this would be a consequence of the current situation (that assert works for the current master)

ThePauliPrinciple:
Should I just use `sorted` leaving a comment that it should use `ordered` when `Tuple` can be used safely?

oscarbenjamin:
> Should I just use `sorted` leaving a comment that it should use `ordered` when `Tuple` can be used safely?

It's more important to fix the bug so that `ordered` works if possible.

ThePauliPrinciple:
Let's see if with the recent merge related to `Tuple` this is now possible without further changes.

ThePauliPrinciple:
Yeah, was afraid of this, there are still a bunch of places in the library that use non-`Basic` arguments in `Tuple`s. (either intentionally or not)

oscarbenjamin:
> Yeah, was afraid of this, there are still a bunch of places in the library that use non-`Basic` arguments in `Tuple`s. (either intentionally or not)

It's okay, we're getting closer!

ThePauliPrinciple:
Yes, I will continue trying to solve these cases and at some point this should just work.

ThePauliPrinciple:
Unfortunately it seems that this still fails.

ThePauliPrinciple:
It's probably not ideal, but I think it is either needed that `ordered` deals with `BasicMeta`/`Function` or we can't use `ordered` for `Dict` here.

oscarbenjamin:
> It's probably not ideal, but I think it is either needed that `ordered` deals with `BasicMeta`/`Function` or we can't use `ordered` for `Dict` here.

Where is the code that needs this?

ThePauliPrinciple:
<details>
<summary> error </summary>

```
___________ sympy/categories/tests/test_drawing.py:test_DiagramGrid ____________
  File "/home/runner/work/sympy/sympy/sympy/categories/tests/test_drawing.py", line 482, in test_DiagramGrid
    grid = DiagramGrid(d, {
  File "/home/runner/work/sympy/sympy/sympy/categories/diagram_drawing.py", line 1189, in __init__
    if groups and (groups != diagram.objects):
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 375, in __ne__
    return not self == other
  File "/home/runner/work/sympy/sympy/sympy/core/basic.py", line 348, in __eq__
    other = _sympify(other)
  File "/home/runner/work/sympy/sympy/sympy/core/sympify.py", line 509, in _sympify
  File "/home/runner/work/sympy/sympy/sympy/core/sympify.py", line 366, in sympify
    return conv(a)
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 299, in <lambda>
    converter[dict] = lambda d: Dict(*d.items())
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 231, in __new__
    items = [Tuple(k, v) for k, v in args]
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 231, in <listcomp>
    items = [Tuple(k, v) for k, v in args]
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 53, in __new__
    obj = Basic.__new__(cls, *args)
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 52, in <genexpr>
  File "/home/runner/work/sympy/sympy/sympy/core/sympify.py", line 366, in sympify
    return conv(a)
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 299, in <lambda>
    converter[dict] = lambda d: Dict(*d.items())
  File "/home/runner/work/sympy/sympy/sympy/core/containers.py", line 236, in __new__
    obj = Basic.__new__(cls, *ordered(items))
  File "/home/runner/work/sympy/sympy/sympy/core/sorting.py", line 306, in ordered
    yield from d[k]
  File "/home/runner/work/sympy/sympy/sympy/core/sorting.py", line 288, in ordered
    d[f(a)].append(a)
  File "/home/runner/work/sympy/sympy/sympy/core/sorting.py", line 192, in _nodes
    return _node_count(e)
  File "/home/runner/work/sympy/sympy/sympy/core/sorting.py", line 175, in _node_count
    return 1 + sum(map(_node_count, e.args))
  File "/home/runner/work/sympy/sympy/sympy/core/sorting.py", line 175, in _node_count
    return 1 + sum(map(_node_count, e.args))
TypeError: 'property' object is not iterable
```
</details>

ThePauliPrinciple:
Adding
```python
        if groups:
            print('groups:', groups, type(groups))
            print('diagram.objects:', diagram.objects, type(diagram.objects))
```
above `if groups and (groups != diagram.objects):` in "sympy/categories/diagram_drawing.py", line 1189

gives

```
groups: {{Object("A"), Object("B"), Object("C"), Object("D"), Object("E")}: {'layout': 'sequential', 'transpose': True}, {Object("A'"), Object("B'"), Object("C'"), Object("D'"), Object("E'")}: {'layout': 'sequential', 'transpose': True}} <class 'dict'>
diagram.objects: {Object("A"), Object("A'"), Object("B"), Object("B'"), Object("C"), Object("C'"), Object("D"), Object("D'"), Object("E"), Object("E'")} <class 'sympy.sets.sets.FiniteSet'>
```

I think the problem is that @smichr 's solution doesn't work for `dict`s, but I might be mistaken.

ThePauliPrinciple:
looks like `Basic` needs something similar to `Expr` in #22593.

oscarbenjamin:
> looks like `Basic` needs something similar to `Expr` in #22593.

Maybe a better approach in general would be to change it so that `Basic/Expr.__eq__` do not attempt to sympify anything. Then if any particular Basic subclass does want to be able to compare equal to an unsympified type it can do something like:
```python
class Dict(Basic):
    def __eq__(self, other):
        if isinstance(other, dict):
            return self._dict == dict
        else:
            return super().__eq__(self, other)
```

ThePauliPrinciple:
I think it is done to automatically to have something like `1 == Integer(1)` work.

oscarbenjamin:
> I think it is done to automatically to have something like `1 == Integer(1)` work.

Yes, but only a handful of SymPy classes could possible compare equal to `1` (Integer, Float, ...?). Why should every other class have to use sympify to find out that `1` becomes `Integer` which is a different Basic class?

ThePauliPrinciple:
My concern is mostly about how many classes would need a new `__eq__` in that case.

oscarbenjamin:
I don't think it's that many classes. It's basically just where a non-sympy type sympifies to a sympy one:
```console
$ git grep 'converter\['
sympy/core/containers.py:converter[tuple] = lambda tup: Tuple(*tup)
sympy/core/containers.py:converter[dict] = lambda d: Dict(*d.items())
sympy/core/numbers.py:converter[float] = converter[decimal.Decimal] = Float
sympy/core/numbers.py:converter[int] = Integer
sympy/core/numbers.py:converter[fractions.Fraction] = sympify_fractions
sympy/core/numbers.py:    converter[type(gmpy.mpz(1))] = sympify_mpz
sympy/core/numbers.py:    converter[type(gmpy.mpq(1, 2))] = sympify_mpq
sympy/core/numbers.py:converter[type(mpmath.rational.mpq(1, 2))] = sympify_mpmath_mpq
sympy/core/numbers.py:converter[mpnumeric] = sympify_mpmath
sympy/core/numbers.py:converter[complex] = sympify_complex
sympy/core/sympify.py:            return converter[complex](a.item())
sympy/core/sympify.py:    object, e.g. ``converter[MyList] = lambda x: Matrix(x)``.
sympy/core/sympify.py:    >>> converter[MyList2] = lambda x: Matrix(x)
sympy/core/sympify.py:            return converter[superclass](a)
sympy/logic/boolalg.py:converter[bool] = lambda x: S.true if x else S.false
sympy/matrices/immutable.py:sympify_converter[MatrixBase] = sympify_matrix
sympy/matrices/immutable.py:sympify_converter[_matrix] = sympify_mpmath_matrix
sympy/polys/domains/pythonrational.py:converter[PythonRational] = sympify_pythonrational
sympy/sets/fancysets.py:converter[range] = lambda r: Range(r.start, r.stop, r.step)
sympy/sets/sets.py:converter[set] = lambda x: FiniteSet(*x)
sympy/sets/sets.py:converter[frozenset] = lambda x: FiniteSet(*x)
```
From there I make it
- Tuple
- Dict
- Number
- BooleanAtom
- Matrix
- FiniteSet

ThePauliPrinciple:
So then Basic can just be something simple like
```python
class Basic():
   _compare_to = []
   def __eq__(self, other):
       if type(other) in self._compare_to:
           other = _sympify(other)
       if isinstance(other, Basic) and type(self) == type(other) and  self.args == other.args:
           return True
       return NotImplemented       
```
and we define non-`Basic` types in `_compare_to`?

oscarbenjamin:
Maybe for `Basic.__eq__` we can keep it even simpler and not check for any anything (i.e. don't bother with `compare_to`). The subclasses that want to compare to other types can do that in their `__eq__` methods.

ThePauliPrinciple:
And then we have something like this in `BooleanAtom`
```python
    def __eq__(self, other):
        if isinstance(other, bool):
            return self.__bool__ == other
        else:
            return super().__eq__(other)

    __hash__ = Basic.__hash__
```
I suppose there is no need to convert with `_sympify` if you can just access your `bool` value?

ThePauliPrinciple:
Looks like something went wrong with the tests, they didn't stop, so closing this to stop them.

ThePauliPrinciple:
I wonder if `Dict` should maybe inherit from `FiniteSet` instead of `Tuple` and make it structurally like `FiniteSet(Tuple(key, value) for key, value in dict)`. The problem is that a `dict` is unordered, so it's more like a `set` than a `tuple`.

oscarbenjamin:
Dict does not inherit from Tuple. Inheriting from FiniteSet wouldn't be right because then it would semantically be a Set which it isn't. It could have a single arg that is a FiniteSet though e.g.:
```python
In [3]: Dict(FiniteSet(Tuple(1, 2), Tuple(3, 4)))
Out[3]: {(1, 2): (3, 4)}
```

ThePauliPrinciple:
Looks like comparison with `._dict` is possible, at least for the tests inside SymPy. It might still cause issues downstream. It is also worth it to consider just converting to `Dict`, since it only happens when comparing directly with `Dict` now, which should not be that bad for performance and doesn't cause any backwards compatibility issues. It will also be easier maintainance wise if for some reason `Dict` gets changed in the future.

I'm fine with either.

This still needs a commit to remove the option not chosen.

oscarbenjamin:
> Looks like comparison with `._dict` is possible, at least for the tests inside SymPy.

Presumably there is no test for this:
```python
In [15]: Dict({1:1})._dict == Dict({1:1.0})._dict
Out[15]: True

In [16]: Dict({1:1}) == Dict({1:1.0})
Out[16]: False
```
I prefer `False` in this instance and I don't think that performance is a significant issue with `Dict`. Of course if we had #20033 then these would both give the same result.

oscarbenjamin:
Looks good. Can you squash the commits?

ThePauliPrinciple:
Apparently, no I cannot.

ThePauliPrinciple:
No clue if github counts this as squashed, but it seems to locally in git.

oscarbenjamin:
If you look above you can see "Commits (1)" so this is now a single commit. IT also has the correct changes.

PR #22582, opened 2021-12-01
core/evalf: no longer create unneeded Tuples with None arguments

### LINKED ISSUE #22581: evalf unintentionally creates Tuple with None as argument
In `terms.count` the terms get converted to `Tuple` during the comparison with `S.ComplexInfinity` but it will contain `None` as arguments
https://github.com/sympy/sympy/blob/ed0bfd1e47345c8bd45ef542eef18f6a21c55685/sympy/core/evalf.py#L599-L604
I believe this can be resolved with
```python
n = sum([term.count(S.ComplexInfinity) for term in terms])
```
to avoid the creation of `Tuple`s.

ThePauliPrinciple:

Fixes #22581

Helps with #22445

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -600,8 +600,7 @@ def evalf_add(v: 'Add', prec: int, options: OPT_DICT) -> TMP_RES:
         options['maxprec'] = min(oldmaxprec, 2*prec)
 
         terms = [evalf(arg, prec + 10, options) for arg in v.args]
-        # remove any None so they won't be sympified during comparison
-        n = [term for term in terms if not isinstance(term, tuple)].count(S.ComplexInfinity)
+        n = sum(1 for i in terms if i is S.ComplexInfinity)
         if n >= 2:
             return fnan, None, prec, None
         re, re_acc = add_terms(
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -601,7 +601,7 @@ def evalf_add(v: 'Add', prec: int, options: OPT_DICT) -> TMP_RES:
 
         terms = [evalf(arg, prec + 10, options) for arg in v.args]
         # remove any None so they won't be sympified during comparison
-        n = filter(None, terms).count(S.ComplexInfinity)
+        n = [term for term in terms if not isinstance(term, tuple)].count(S.ComplexInfinity)
         if n >= 2:
             return fnan, None, prec, None
         re, re_acc = add_terms(
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -600,7 +600,8 @@ def evalf_add(v: 'Add', prec: int, options: OPT_DICT) -> TMP_RES:
         options['maxprec'] = min(oldmaxprec, 2*prec)
 
         terms = [evalf(arg, prec + 10, options) for arg in v.args]
-        n = sum([term.count(S.ComplexInfinity) for term in terms])
+        # remove any None so they won't be sympified during comparison
+        n = filter(None, terms).count(S.ComplexInfinity)
         if n >= 2:
             return fnan, None, prec, None
         re, re_acc = add_terms(
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -600,7 +600,7 @@ def evalf_add(v: 'Add', prec: int, options: OPT_DICT) -> TMP_RES:
         options['maxprec'] = min(oldmaxprec, 2*prec)
 
         terms = [evalf(arg, prec + 10, options) for arg in v.args]
-        n = terms.count(S.ComplexInfinity)
+        n = sum([term.count(S.ComplexInfinity) for term in terms])
         if n >= 2:
             return fnan, None, prec, None
         re, re_acc = add_terms(


sympy/core/evalf.py
```diff
@@ -600,7 +600,7 @@ def evalf_add(v: 'Add', prec: int, options: OPT_DICT) -> TMP_RES:
         options['maxprec'] = min(oldmaxprec, 2*prec)
 
         terms = [evalf(arg, prec + 10, options) for arg in v.args]
-        n = terms.count(S.ComplexInfinity)
+        n = sum([term.count(S.ComplexInfinity) for term in terms])
```
smichr, line 603:
```suggestion
        n =filter(None, terms).count(S.ComplexInfinity)  # else None will be sympified during comparison
```

ThePauliPrinciple: `terms` is a `list` of `tuple` that potentially has `None`, so this would not filter the problematic `None`s and would restore the original problem (that during the comparison with `S.ComplexInfinity`, the `tuple` that contains `None` gets converted to `Tuple`)

ThePauliPrinciple: Note that I don't know if `terms` is always a `list` of `tuple` it just always was when I was printing it during debugging.

ThePauliPrinciple: The suggested change does not solve the underlying issue.

ThePauliPrinciple: It is not a problem that `terms` contains `None` it is a problem that `terms` contains `tuple` with `None`

ThePauliPrinciple: `n = [ term for term in terms if not isinstance(term, tuple)].count(S.ComlexInfinity)` could be used I think

oscarbenjamin: The problem is that `__eq__` should not create this Tuple with None in the first place. If sympify refused to sympify a tuple with unsympifiable arguments then everything would be fine:
https://github.com/sympy/sympy/blob/4489b4c7c51a65966b1a95f2af6f7e0c8c461794/sympy/core/expr.py#L148-L154

smichr: > list of tuples
Sorry, @ThePauliPrinciple

sympy/core/evalf.py
```diff
@@ -600,7 +600,8 @@ def evalf_add(v: 'Add', prec: int, options: OPT_DICT) -> TMP_RES:
         options['maxprec'] = min(oldmaxprec, 2*prec)
 
         terms = [evalf(arg, prec + 10, options) for arg in v.args]
-        n = terms.count(S.ComplexInfinity)
+        # remove any None so they won't be sympified during comparison
+        n = [term for term in terms if not isinstance(term, tuple)].count(S.ComplexInfinity)
```
smichr, line 604:
```suggestion
        n = sum(1 for i in terms if i is S.ComplexInfinity)
```

smichr: Here's a more direct workaround.

sympy/core/evalf.py
ThePauliPrinciple: I think the confusion arised because my proposed solution doesn't solve the source of the problem as @oscarbenjamin indicates above.

Changing from `sympify` to `_sympify` in `Tuple` solves this issue I believe, but that should probably be deprecated properly. How do we solve this specific problem during the deprecation period?

sympy/core/evalf.py
ThePauliPrinciple: I considered `is S.ComplexInfinity` too, but I wonder if there are any cases that we do want to catch where a term in `terms` evaluates to `S.ComplexInfinity`

ThePauliPrinciple:
I think it would be better to not solve this in a PR like this. If `Tuple` correctly rejected non-`Basic` arguments, there would be no issue. So adding a solution here only causes more code that is eventually not needed (when `Tuple` behaves properly).

sympy/core/evalf.py
oscarbenjamin: You could change `Tuple.__new__` to do
```python
try:
    args = tuple(_sympify(arg) for arg in args)
except SympifyError:
    # give a warning and then fall back to existing behaviour
    SymPyDeprecationWarning(...).warn()
    args = tuple(sympify(arg) for arg in args)

Comment (ThePauliPrinciple) [reply sympy/core/evalf.py]: I was hoping to do it in such a way that such warnings would only ever be shown if a user uses the library in a deprecated way and have the library always (at least for the available tests) avoid such situations.

Comment (oscarbenjamin) [reply sympy/core/evalf.py]: Oh, actually that's a good point. We don't want `a==b` to give a warning. There's probably a way to make this work but I also think that sometimes it's necessary to just make a change that isn't 100% compatible.

Comment (smichr) [reply sympy/core/evalf.py]: A tuple doesn't need sympification to find out that it's not equal to zoo. We can catch this by reporting failure for any iterable that doesn't have a `_sympy` method (or whatever a 3rd party object must implement so `sympify` knows how to sympify it) since no iterable will equal an Expr.

File: sympy/core/evalf.py (hunk 2 above)
Comment (smichr) [reply sympy/core/evalf.py:604]: You mean an element in `term`?

File: sympy/core/evalf.py (hunk 1 above)
Comment (smichr) [reply sympy/core/evalf.py]: `Expr.__eq__`:
```python
    def __eq__(self, other):
        if not hasattr(other, '_sympy_') and iterable(other):
            # no expression is an iterable
            return False
        try:
            other = _sympify(other)
            if not isinstance(other, Expr):
                return False
...
```

ThePauliPrinciple: Well, here is the problem: `assert _sympify(tuple()) == Tuple()`  so any `tuple` gets converted to a `Tuple` (on `other = _sympify(other)` even if that causes a `Tuple` that should not exist e.g. `(None,)` gets converted to `Tuple(None)`.

ThePauliPrinciple: Changing `sympify` in `Tuple` to `_sympify` (because a `None` anywhere would raise a `SympifyError`) solves the issue, but SymPy generously abuses `Tuple` to apply `sympify` to multiple objects (along the lines of `x,y,z = Tuple(x, y, z)`). All those cases need to be fixed and then this can be solved with the aforementioned change in `Tuple`.

sympy/core/evalf.py
smichr: It doesn't look like `zoo` will be in a tuple (based on the use of `ComplexInfinity` in evalf.py). In addition the docstring says:
```
    An optional tuple, ``(re, im, re_acc, im_acc)``
    which are the real, imaginary, real accuracy
    and imaginary accuracy respectively. ``re`` is
    an mpf value tuple and so is ``im``. ``re_acc``
    and ``im_acc`` are ints.
    NB: all these return values can be ``None``.
    If all values are ``None``, then that represents 0.
    Note that 0 is also represented as ``fzero = (0, 0, 0, 0)``.

```
It should also note that `zoo` will possibly be returned. I don't know why it says an 'optional tuple' unless I missed something. It returns `r`, possibly the result after `chop_parts` so it should be a tuple or `zoo`.

ThePauliPrinciple: I see. Then yes, this would solve the problem. However, it essentially introduces an extra check because comparisons between `tuple`s and `Basic`s are bugged. So that should be fixed rather than introducing a check here.

We should definately keep this as a backup solution if it turns out that solving the underlying problem is not (at this point) possible.

sympy/core/evalf.py
smichr: The `iterable(other)` catches the tuple case and returns False. I tried this locally and it works fine. And if you ignore an iterable you can, with one test, ignore everything inside, `x == [i for i in range(1,10**6)]` will not need to sympify a million integers to indicate False ;-)

smichr: I am posting this in several PRs hoping you will consider it and have a plan before going too far down the nonBasic erradication trail:  here is an example of storing Functions in an object: Subs(Derivative(f(f(x)), x), f, cos). How can this be handled if we don't allow BasicMeta as args?

oscarbenjamin: See https://github.com/sympy/sympy/pull/22539#issuecomment-985279336

ThePauliPrinciple: > 

But the "tuple case" `assert _sympify(tuple()) == Tuple()` is expected to return `True` as far as I understand

ThePauliPrinciple: I do agree with you on `Function`, I don't know how to solve that case, but I do believe that all other uses of non-`Basic` can be removed from the library. At which point figuring out how to make `Function` a `Basic` (or at the very least behave exactly like a `Basic`) might start to be worthwhile.

ThePauliPrinciple:
fixed by https://github.com/sympy/sympy/pull/22593

Issue #17087, opened 2019-06-24
Remove args sorting from And and Or?


asmeurer:
As you can see from https://github.com/sympy/sympy/issues/17066#issuecomment-504774120, the sorting in the And and Or constructor (from LatticeOp) can take up a significant amount of time when using large logic expressions. 

We should see if we can remove it. The `_hashable_content`, which is used by `__eq__`, already uses the frozenset `_args`, so equality checking would work even if the `args` order is different. However, it will break any code that expects args to be ordered, or expects the current order, so I do not expect this to be an easy fix. 

An important warning here, currently `LatticeOp.args` is cached. We should remove the cache when testing this. Otherwise the cache will make it so that equal objects always have the same args, and potential test failures will be hidden. The cache wouldn't be needed without sorting anyway. We could just set `.args` in the constructor.

This is related to https://github.com/sympy/sympy/issues/5954, but I think removing the ordering from Add/Mul will be much harder, as they are used everywhere. We should try doing it with the logic classes first, to see what are the issues that come up.

asmeurer:
Oh and I forgot to mention that I expect this will not be an easy thing to change. Usually changes like this tend to bring up issues that are difficult to fix. I could be wrong, but a heads up that this is most likely not an easy to fix issue.

smichr:
> it will break any code that expects args to be ordered

The only major expectation that I am aware of is that if there is an argument that is a Number in an Add, it is in slot 0. (This applies to Mul, too, but you aren't mentioning that here.)

asmeurer:
The expectation would be implicit. The issue is that the same object could appear twice with different arg orderings. 

Here's a test failure from making the change

```
______________________________________________________________________________________________________________
______________________ sympy/logic/tests/test_boolalg.py:test_relational_simplification ______________________
Traceback (most recent call last):
  File "/home/aaronmeurer/Documents/python/sympy/sympy/sympy/logic/tests/test_boolalg.py", line 970, in test_relational_simplification
    assert Or(x >= y, w < z, x <= y).simplify() == S.true
AssertionError
```

(`x`, `y`, `z`, and `w` are `real=True`)

I haven't digged into it, but somehow simplify() isn't handling the different arg ordering correctly. 

This is the test change (the real change should probably set `args` in the constructor as I noted)

```diff
diff --git a/sympy/core/operations.py b/sympy/core/operations.py
index 4c37663f4..2d10678f5 100644
--- a/sympy/core/operations.py
[...]

PR #22585, opened 2021-12-01
Improved pattern-based logic simplification

oscargus:

Pulled the logic simplification part from #17330.

Better pattern-based simplification. Fewer patterns and combinations tested. Three-term patterns (only for `And` so far).

#### Other comments

Some minor code improvements in the first commit.

#### Release Notes

* logic
    * Improved simplification of Boolean expressions with relations.

### DIFF
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -2493,7 +2493,7 @@ def SOPform(variables, minterms, dontcares=None):
     .. [1] https://en.wikipedia.org/wiki/Quine-McCluskey_algorithm
 
     """
-    if minterms == []:
+    if not minterms:
         return false
 
     variables = tuple(map(sympify, variables))
@@ -2569,7 +2569,7 @@ def POSform(variables, minterms, dontcares=None):
     .. [1] https://en.wikipedia.org/wiki/Quine-McCluskey_algorithm
 
     """
-    if minterms == []:
+    if not minterms:
         return false
 
     variables = tuple(map(sympify, variables))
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
I think this looks good. I'm not entirely sure what effect the changes have.

oscargus:
There are four(?) things:

1. Refactoring to a function instead of a method.
2. Better checking of relationals, earlier the code tried all possible combinations of reversing the relationals to make them match, now only the Eq/Ne-relationals are reversed, the others are canonicalized to Le/Lt (unfortunately, I selected Ge/Gt for three-terms, but will fix that later). This also lead to that fewer patterns must be tested (left them commented for now, both in case I was wrong and for other to see that they were considered).
3. Introduced three-term pattern, such that `And(a > b, b > c, c > a)` can be simplified to `False`.
4. Patterns are stored (and matched) as Tuple rather than the Boolean function which makes things a bit better/quicker as there is no evaluation etc.

oscargus:
This is probably an enabler for #17087 as well.

oscarbenjamin:
Okay well it looks good to me although it sounds like you still intend to make more changes.

oscargus:
There are things that can be changed/improved/extended, but hard to say when... I expect to continue working on the simplification, but it is also good to get things in reasonably small batches and based on experience I may not always do linear development, rather jump back and forth to do something that seems intriguing at the moment (based on the amount of time that I believe I have).

oscarbenjamin:
Okay so what does this mean:

> unfortunately, I selected Ge/Gt for three-terms, but will fix that later

Is that a bug or just something that is potentially less efficient or leads to less simple code or something?

oscargus:
The three-term normalizes all inequalities to Gt/Ge. The result is then sent to the two-term where they are normalized to Lt/Le. So inefficient, although the effect is probably minimal (considering that the matching takes much, much longer time), more a bit annoying and "stupid".

oscarbenjamin:
> the matching takes much, much longer time

Is this the kind of thing that matchpy could help with?

https://github.com/HPAC/matchpy#many-to-one-matching

oscarbenjamin:
Okay well looks good for now.

PR #22587, opened 2021-12-02
tensor/array: Flatten is no longer a subclass of Basic

ThePauliPrinciple:

See https://github.com/sympy/sympy/pull/22445

Currently `Flatten` is a subclass of `Basic`. It is meant to avoid
evaluating its argument for as long as possible, thus it can't make
sure that its arguments are a `Basic` (e.g. this is meant to work
with mutable arrays, which can not be a `Basic`. Moreover, currently
this class seems to not use any of the properties of a `Basic`.
I realise this module is currently under development, so I would
like to know if `Basic` can savely be removed considering future
plans for this module.

#### Release Notes

NO ENTRY


ThePauliPrinciple:
Note that the failed doctest is because it now no longer uses `Basic` printing.

oscarbenjamin:
Looks good

PR #18594, opened 2020-02-06
Use autosummary for documentation

sylee957:
#### Brief description of what is fixed or changed

SymPy documentation had grown up too much and it is really annoying to scroll down thousands of lines.

One of the problem is that sympy classes have so many functions, attributes, keywords that have to be documented and I don't think that it's a good idea to paste all the docs in a single page using `autodoc`.

For example, I see numpy and scipy documents a class, a method in a single page and use redirectives to link them, instead of having all the class, functions documented in a single page.

I think that we should start to use `autosummary` and use templates to automatically generate stub pages for each `rst` files

I have an example to showcase for matrix documentation
Such that users can see the main page like below, only containing the classes, loose functions and summaries

Screenshot_2020-02-07 Matrices (linear algebra) — SymPy 1 6 dev documentation

And also each classes should list only the summaries for each methods, because sympy classes often have massive numbers of methods and properties.

image

And finally, it should have one doc under one separate page.

image

#### Release Notes

NO ENTRY

PR #22589, opened 2021-12-03
Use autosummary in the docs

asmeurer:

This is a redux of the work that was started in https://github.com/sympy/sympy/pull/18594. 

Our [documentation poll](
https://forms.gle/EjkFEdnLXaYxjvfn6) is still ongoing, but one thing that is already becoming clear from some of the responses is that our documentation would benefit from splitting the documentation of each function into separate pages. This is achieved with the autosummary Sphinx extension. 

Not only does this make each page of the documentation more manageable, but it also makes it easier for Google search results to return the exact function in question, rather than just a large page that contains the function somewhere on it. 

For now, I have only modified the matrices.rst file. Please review what this looks like, both in the RST and the generated HTML. If we like how this looks, we can add it to other pages.

#### Other comments

- You have to manually add `:toctree: ref` to each autosummary to tell it where
  to put the autogenerated files. You should always use the name "ref" for
  this for consistency, but you can use something like `:toctree: ../ref` to put
  it on level up. The word "ref" is part in the URL for each generated page.

- The classes I have documented here really need to be organized in a better
  way. However, I have not done that in this PR. An advantage of autosummary
  is that we can move where a class is documented in the organization without
  breaking the URL to it, since it lives on a separate page.

- This will break any existing links to specific anchors on this page.

- The class template hardcodes :members:. However, it would be easy to use
  other combinations by adding additional templates, although each autosummary
  list can only use one template at a time.

- The docs for classes still list each method on the same page. This is
  because autosummary is not really capable of doing recursive autosummary on
  class members in a way that only includes the members we want to include.
  See the commit message of a0e955bfc56267877442ab318e9ec92f29a806d6.

- Autosummary is very particular about how you spell the name of the function
  so that it can find it. I recommend avoiding `.. module::` or `..
  currentmodule::`and just spelling everything explicitly, using `~` before the
  name to truncate the module name for public names (eventually, we will have
  *only* public names in this part of documentation, see
  #22105, but this is a separate issue to
  clean this up).

- When building, be aware that files for the autosummary are generated
  automatically, but are only cleaned when you do a `make clean`. Thus if you
  add something to an autosummary and then delete it, you will get warnings in
  the build from the generated file that no longer is referenced anywhere.

#### Release Notes

- other
  - Start using the autosummary extension in the docs, which puts documentation for each function and class on a separate page. 

#### Screenshots

Here is what the matrices page looks like
Screen Shot 2021-12-02 at 7 38 34 PM

and here is what a page for one of the clases looks like

Screen Shot 2021-12-02 at 7 38 59 PM

### DIFF
--- a/doc/src/modules/matrices/matrices.rst
+++ b/doc/src/modules/matrices/matrices.rst
@@ -1,8 +1,6 @@
 Matrices (linear algebra)
 =========================
 
-.. module:: sympy.matrices.matrices
-
 Creating Matrices
 -----------------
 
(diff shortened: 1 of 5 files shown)


asmeurer:
I've noticed a somewhat important limitation of this, as it's currently implemented. This puts the fully qualified name as the title of every autosummary page. This also goes in the URL, so you'd get something like docs.sympy.org/latest/modules/matrices/ref/sympy.matrices.dense.ones.html instead of just docs.sympy.org/latest/modules/matrices/ref/ones.html (and it would also have `sympy.matrices.dense.ones` as the title of the page). This is bad because it would encourage people to import top-level names from submodules. We're already not great at distinguishing these anyway, but a bigger problem is that with this, it affects the URL too. So we ever wanted to move the function around or rename the submodule, it would break the URL. 

I think I might have a workaround for this, by making separate "public" and "private" templates. I'll have to play with it tomorrow.

doc/src/modules/matrices/matrices.rst
```diff
@@ -1,8 +1,6 @@
 Matrices (linear algebra)
 =========================
 
-.. module:: sympy.matrices.matrices
-
 Creating Matrices
 -----------------
```
oscarbenjamin, line 5: All of the content here above the autodoc lines at the bottom of the rst file should be moved into a separate rst file that could be linked to from here (perhaps a guide). The number one benefit of using autosummary is making all the different functions/methods more discoverable so we need it to be clear when you come to this page that it has a list of those functions which is not yet immediately obvious because I still need to scroll through a load of stuff to see that list.

asmeurer: I agree. I wanted to minimize the diff so that we can focus on just the specific details about autosummary to make sure everything looks tenable before going all in on it. So far there are some concerns that need addressing, like the one I mentioned above.

oscarbenjamin:
The basic idea here looks good. In combination with the other changes to the top level organisation of the docs I think we're getting to something much better.

PR #22591, opened 2021-12-03
not foo in -> foo not in; not foo is -> foo is not

smichr:

Just tidy up code to reflect the possible usage of the 'is not' and the 'not in' operators. It is easy to foget about these and just write negations as `not cond`; but when `cond` involves `is` or `in`, the `not` can be put more legibly inside:
```python
if not 1 in args -> if 1 not in args
if not 1 is S.One -> if 1 is not S.One
```
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/combinatorics/subsets.py
+++ b/sympy/combinatorics/subsets.py
@@ -172,7 +172,7 @@ class Subset():
             else:
                 indices.remove(i)
                 i = i - 1
-                while not i in indices and i >= 0:
+                while i >= 0 and i not in indices:
                     i = i - 1
                 if i >= 0:
                     indices.remove(i)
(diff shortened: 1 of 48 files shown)


sympy/combinatorics/subsets.py
```diff
@@ -172,7 +172,7 @@ def next_lexicographic(self):
             else:
                 indices.remove(i)
                 i = i - 1
-                while not i in indices and i >= 0:
+                while i >= 0 and i not in indices:
```
smichr, line 175: Here, I figured it would be faster to do the relational than to check for containment, so put it first.

oscarbenjamin: I think `i` is guaranteed to be `int` so this would be faster. It would also be better if `indices` was a `set`.

oscarbenjamin:
Looks good

PR #22593, opened 2021-12-03
Iterable

smichr:

spinoff from #22582 with some additional changes

1. `Expr.__eq__` automatically rejects as unequal anything that is iterable (and also saves time by not having to sympify anything in the iterable
2. The default arg z=None is handled without putting it into a Tuple
3. it looked like NotIterable was in campatibililty at one time -- the tests for compatibility wanted to import it from there -- so it was added along with the others that are deprecated in favor of importing them from iterables.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -155,7 +155,7 @@ class Expr(Basic, EvalfMixin):
             try:
                 other = _sympify(other)
             except (SympifyError, SyntaxError):
-                return False
+                return NotImplemented
         if not isinstance(other, Expr):
             return False
         # check for pure number expr
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -146,11 +146,11 @@ class Expr(Basic, EvalfMixin):
         return self._args
 
     def __eq__(self, other):
-        assert not iterable(self), 'needs __eq__ definition'
         if not isinstance(other, Basic):
-            if iterable(other):
-                # XXX assume that an iterable with a _sympy_
-                # method will not sympify to a non-iterable
+            if iterable(other) and not hasattr(other, '_sympy_'):
+                # XXX iterable self should have it's own __eq__
+                # method if the path gives a false negative
+                # comparison
                 return False
             try:
                 other = _sympify(other)
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -12,7 +12,7 @@ from .sorting import default_sort_key
 from .kind import NumberKind
 from sympy.utilities.exceptions import SymPyDeprecationWarning
 from sympy.utilities.misc import as_int, func_name, filldedent
-from sympy.utilities.iterables import has_variety, sift, iterable
+from sympy.utilities.iterables import has_variety, sift, iterable, NotIterable
 from mpmath.libmp import mpf_log, prec_to_dps
 from mpmath.libmp.libintmath import giant_steps
 
@@ -147,7 +147,8 @@ class Expr(Basic, EvalfMixin):
 
     def __eq__(self, other):
         if not isinstance(other, Basic):
-            if iterable(other) and not hasattr(other, '_sympy_'):
+            if iterable(other, exclude=(str, NotIterable)
+                    ) and not hasattr(other, '_sympy_'):
                 # XXX iterable self should have it's own __eq__
                 # method if the path gives a false negative
                 # comparison
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -12,7 +12,7 @@ from .sorting import default_sort_key
 from .kind import NumberKind
 from sympy.utilities.exceptions import SymPyDeprecationWarning
 from sympy.utilities.misc import as_int, func_name, filldedent
-from sympy.utilities.iterables import has_variety, sift
+from sympy.utilities.iterables import has_variety, sift, iterable
 from mpmath.libmp import mpf_log, prec_to_dps
 from mpmath.libmp.libintmath import giant_steps
 
@@ -146,11 +146,15 @@ class Expr(Basic, EvalfMixin):
         return self._args
 
     def __eq__(self, other):
-        try:
-            other = _sympify(other)
-            if not isinstance(other, Expr):
+        if not isinstance(other, Basic):
+            if iterable(other):
+                # no expression is iterable
                 return False
-        except (SympifyError, SyntaxError):
+            try:
+                other = _sympify(other)
+            except (SympifyError, SyntaxError):
+                return False
+        if not isinstance(other, Expr):
             return False
         # check for pure number expr
         if  not (self.is_Number and other.is_Number) and (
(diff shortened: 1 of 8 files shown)


sympy/core/random.py
```diff
@@ -72,10 +72,13 @@ def verify_numerically(f, g, z=None, tol=1.0e-6, a=2, b=-1, c=3, d=1):
     True
     """
     from sympy.core.symbol import Symbol
+    from sympy.core.sympify import sympify
     from sympy.core.numbers import comp
-    from sympy.core.containers import Tuple
-    f, g, z = Tuple(f, g, z)
-    z = [z] if isinstance(z, Symbol) else (f.free_symbols | g.free_symbols)
+    f, g = (sympify(i) for i in (f, g))
```
ThePauliPrinciple, line 77: There is a similar (mis)use of `Tuple` here: https://github.com/sympy/sympy/blob/ed0bfd1e47345c8bd45ef542eef18f6a21c55685/sympy/integrals/integrals.py#L808-L813

ThePauliPrinciple: And here:
https://github.com/sympy/sympy/blob/ed0bfd1e47345c8bd45ef542eef18f6a21c55685/sympy/tensor/indexed.py#L326-L334

sympy/core/expr.py
```diff
@@ -146,11 +146,15 @@ def _hashable_content(self):
         return self._args
 
     def __eq__(self, other):
-        try:
-            other = _sympify(other)
-            if not isinstance(other, Expr):
+        if not isinstance(other, Basic):
```
ThePauliPrinciple, line 149: Note that `Tuple` defines its own custom `__eq__`: https://github.com/sympy/sympy/blob/ed0bfd1e47345c8bd45ef542eef18f6a21c55685/sympy/core/containers.py#L96-L99

Although I do believe this is handled correctly.

ThePauliPrinciple:
Looks good.

If it is not possible to solve the two other uses of `Tuple` (provided in the review) I think this can be accepted.

sympy/core/expr.py
```diff
@@ -146,11 +146,15 @@ def _hashable_content(self):
...
+            if iterable(other):
```
ThePauliPrinciple, line 150: looks like `dict` is not considered an `iterable`, but I think it should get a similar treatment

oscarbenjamin:
ImmutableMatrix is one example of an iterable Expr. So is MatrixSymbol but iterable thinks that it isn't:
```python
In [4]: from sympy.utilities.iterables import iterable

In [5]: iterable(ImmutableMatrix([[1, 2]]))
Out[5]: True

In [7]: x = MatrixSymbol('x', 2, 2)

In [8]: iterable(x)
Out[8]: False

In [9]: list(x)
Out[9]: [x₀₀, x₀₁, x₁₀, x₁₁]
```

ThePauliPrinciple:
But `MatrixSymbol` is a `Basic`. So it shouldn't matter here since `iterable` is only used for non-`Basic` equalities.

oscarbenjamin:
Yes, but the question is whether an Expr could ever be considered equal to a non-Basic iterable e.g. something like:
```python
In [6]: import sympy as sym

In [7]: import numpy as np

In [8]: sym.Matrix([[1, 2], [3, 4]]) == np.array([[1, 2], [3, 4]])
Out[8]: False
```
I don't know of any examples but I'm just considering it.

ThePauliPrinciple:
`Tuple(1, 2, 3) == (1, 3, 4)`

Note that `Matrix` is not an `Expr`.

oscarbenjamin:
> Note that `Matrix` is not an `Expr`.

ImmutableMatrix is though.

ThePauliPrinciple:
```python
>>> from sympy import ImmutableMatrix
>>> import numpy as np
>>> print(ImmutableMatrix([[1, 2], [3, 4]]) == np.array([[1, 2], [3, 4]]))
None
```
Looks like a bug (on 1.9)

smichr:
> Looks like a bug (on 1.9)

Since `_sympify` is strict, the np array is not converted.

```python
>>> a=(np.array([[1, 2], [3, 4]]))
>>> s=S(a); type(s)
<class 'sympy.tensor.array.dense_ndim_array.ImmutableDenseNDimArray'>
>>> s==a
False
```

Also, the result is not a Matrix (and not Expr) so the changed code does not apply in this case.

But if something does define `_symy_` is should be handled. I made a change towards that end.

ThePauliPrinciple:
Yes, it isn't caused due to this PR, but surely `==` should not return `None`.

smichr:
I'm happy with how often this shortcut actually gets used in the tests.

smichr:
This now only returns a False if the iterable other doesn't have a `_sympy_` method, so I think it addresses the concern of @oscarbenjamin . I think this is ready if this run passes.

ThePauliPrinciple:
Does it work yet for `dict`? It doesn't pass `iterable({})` but I think it should be treated in the same way

sympy/core/expr.py
```diff
@@ -146,11 +146,17 @@ def _hashable_content(self):
...
     def __eq__(self, other):
-        try:
-            other = _sympify(other)
-            if not isinstance(other, Expr):
+        if not isinstance(other, Basic):
+            if iterable(other) and not hasattr(other, '_sympy_'):
+                # XXX iterable self should have it's own __eq__
+                # method if the path gives a false negative
+                # comparison
+                return False
```
oscarbenjamin, line 154: This should be `return NotImplemented`.

sympy/core/expr.py
smichr: Dict is Basic but not Expr so it will get reject without sympification.

sympy/core/expr.py
smichr: done

sympy/core/expr.py
ThePauliPrinciple: I mean a python `dict`, which is not a `Basic`, nor will it go to the `iterable` branch, so it will try to `sympify` the `dict`, potentially creating `Dict` with `None` just like it did with `tuple->Tuple` before

sympy/core/expr.py
```diff
@@ -146,11 +146,17 @@ def _hashable_content(self):
         return self._args
 
     def __eq__(self, other):
-        try:
-            other = _sympify(other)
-            if not isinstance(other, Expr):
+        if not isinstance(other, Basic):
+            if iterable(other) and not hasattr(other, '_sympy_'):
```
ThePauliPrinciple, line 150:
```python
if (iterable(other) or isinstance(other, dict)) and not hasattr(other, '_sympy_')
```

ThePauliPrinciple: preferable : 
```python
if iterable(other, exclude = (str, NotIterable)) and not hasattr(other, `_sympy_`):
```
Looks like `dict` is excluded by default

ThePauliPrinciple:
Thanks a lot! I believe this will solve a lot of the remaining errors in https://github.com/sympy/sympy/pull/22445 and would allow for https://github.com/sympy/sympy/pull/22580 to work.

smichr:
> Thanks a lot! I believe this will solve a lot of the remaining errors 

Thanks for your (and @oscarbenjamin ) feedback.

Issue #15416, opened 2018-10-22
45X slowdown in (matrix) differentiation


moorepants:
PR ##15049 introduced a significant slowdown in differentiation. See these benchmark results:

- 813% slowdown http://www.moorepants.info/misc/sympy-asv/#large_exprs.TimeLargeExpressionOperations.time_jacobian_wrt_functions
-  3747% slowdown http://www.moorepants.info/misc/sympy-asv/#large_exprs.TimeLargeExpressionOperations.time_jacobian_wrt_symbols
- 816% slowdown http://www.moorepants.info/misc/sympy-asv/#large_exprs.TimeLargeExpressionOperations.time_manual_jacobian_wrt_functions
- 206% slowdown http://www.moorepants.info/misc/sympy-asv/#physics.mechanics.kane.KanesMethodMassSpringDamper.time_kanesmethod_mass_spring_damper
- 1104% slowdown http://www.moorepants.info/misc/sympy-asv/#physics.mechanics.lagrange.LagrangesMethodMassSpringDamper.time_lagrangesmethod_mass_spring_damper
- http://www.moorepants.info/misc/sympy-asv/#dsolve.TimeDsolve01.time_dsolve
![selection_307](https://user-images.githubusercontent.com/276007/47322518-1c266f80-d60d-11e8-8713-e98ce542c99c.png)

They are all likely the same issue related to slower `.jacobian()` calls.

iamprayush:
I'd like to help with this issue. Could you please tell me where and how to start?

moorepants:
You can profile the benchmark code and then try to come up with ways to speed things up without breaking any sympy tests. I recommend pyinstrument to profile.

asmeurer:
Bisected the physics.mechanics.lagrange.LagrangesMethodMassSpringDamper.time_lagrangesmethod_mass_spring_damper
 slowdown to 03ffa11604bcd59a75e3bb6acebe9e34416f8c65 specifically.

asmeurer:
It's not clear to me from an instrument or from that commit's diff what is causing the slowdown. The instrument shows the same code path, just each call takes longer. The whole thing is just a bunch of recursive calls through the diff machinery. Is there a parameter I can tweak in that benchmark to make it run longer?

moorepants:
Unfortunately, that benchmark can't be scaled up. This function:

https://github.com/pydy/pydy/blob/master/pydy/models.py#L16

[...]

Issue #11003, opened 2016-04-13
Disallow string input to API functions


asmeurer:
Most SymPy functions "accidentally" accept strings as input, because they call `sympify` on the input, and sympify converts strings to SymPy expressions. 

This would be somewhat of an API break, but I think we should disallow this. The problems with strings as input are outlined [here](https://github.com/sympy/sympy/wiki/Idioms-and-Antipatterns#strings-as-input). Additionally, it only works by accident, and for some functions, it doesn't work at all. 

I think we should remove it. We can use `_sympify` to convert expressions, which doesn't allow strings. Perhaps there's a better way, which prevents `sympify` from creeping back in. 

This is technically an API break, so we should consider if we want to run a deprecation cycle. It's not documented API, and I at least personally have stated several times that this could change at any time. Regardless, we should make the error message for string input clear that strings can be converted to SymPy expressions via `sympify`. We may also want to copy the contents of [that wiki page](https://github.com/sympy/sympy/wiki/Idioms-and-Antipatterns#strings-as-input) to the official docs.

asmeurer:
This will be easy to fix once we decide how we want to do it.

oscarbenjamin:
I agree. String input should be disallowed. I see a lot of examples using this though so maybe it does need a deprecation cycle. 

Actually I'd rather not allow string input to simpify either. There should be a separate function for parsing strings as expressions since it is an entirely different operation.

We should consider and document the alternatives carefully either way.

oscarbenjamin:
Since we're now talking about milestones for 1.4 I think that some kind of deprecation for this should go in there.

asmeurer:
> I agree. String input should be disallowed. I see a lot of examples using this though so maybe it does need a deprecation cycle.

I've always considered functions supporting strings to be accidental behavior (with a couple of exceptions) that shouldn't be relied upon. I wouldn't be opposed to removing it without a deprecation, but maybe we should do it. That's what I wrote [here](https://github.com/sympy/sympy/wiki/Idioms-and-Antipatterns#strings-as-input) (though that isn't technically part of the documentation). Then again my experience from things like StackOverflow and looking around GitHub for uses of SymPy is that a *lot* of users use string inputs, so maybe a deprecation is better. 

[...]

PR #22607, opened 2021-12-05
[proof of concept] only use _sympify in __eq__ when required 

ThePauliPrinciple:
Currently `__eq__` almost always tries to apply `_symplify` to `other`.
While this does follow the ideas of "pythonic duck-typing", it is also
somewhat inefficient. It also seems to sometimes cause the creation of
invalid SymPy objects (that are only used for comparison) which results
in the need for non-pythonic instance checking anyways. Instead, it
was suggested in sympy#22580 (comment) to have any class that 
wants to compare with non-`Basic` classes to perform `_sympify` only 
for those classes it knows that it can compare against. This PR 
implements this, except for `Number` and `Matrix` subclasses, which 
already have their own `__eq__` logic.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -145,26 +145,27 @@ class Expr(Basic, EvalfMixin):
         been defined by a class. See note about this in Basic.__eq__."""
         return self._args
 
-    def __eq__(self, other):
-        if not isinstance(other, Expr):
-            return super().__eq__(other)
-        # check for pure number expr
-        result = True
-        if  not (self.is_Number and other.is_Number) and (
-                type(self) != type(other)):
-            result = False
-        a, b = self._hashable_content(), other._hashable_content()
-        if a != b:
-            result = False
-        # check number *in* an expression
-        for a, b in zip(a, b):
-            if not isinstance(a, Expr):
-                continue
-            if a.is_Number and type(a) != type(b):
-                result = False
-        if not result is Basic.__eq__(self, other):
-            raise TypeError(result, Basic.__eq__(self, other), type(self), type(other), self.__repr__(), other.__repr__())
-        return result
+    #def __eq__(self, other):
+    #    if not isinstance(other, Expr):
+    #        return super().__eq__(other)
+    #    # check for pure number expr
+    #    result = True
+    #    if  not (self.is_Number and other.is_Number) and (
+    #            type(self) != type(other)):
+    #        result = False
+    #    a, b = self._hashable_content(), other._hashable_content()
+    #    if a != b:
+    #        result = False
+    #    # check number *in* an expression
+    #    for a, b in zip(a, b):
+    #        if not isinstance(a, Expr):
+    #            continue
+    #        if a.is_Number and type(a) != type(b):
+    #            result = False
+    #    if not result is Basic.__eq__(self, other):
+    #        raise TypeError(result, Basic.__eq__(self, other), type(self), type(other), self.__repr__(), other.__repr__(),
+    #                type(self) is type(other))
+    #    return result
 
     # ***************
     # * Arithmetics *
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -2,7 +2,7 @@ from typing import Tuple as tTuple
 from collections.abc import Iterable
 from functools import reduce
 
-from .sympify import sympify, _sympify, SympifyError
+from .sympify import sympify, _sympify
 from .basic import Basic, Atom
 from .singleton import S
 from .evalf import EvalfMixin, pure_complex, DEFAULT_MAXPREC
@@ -12,7 +12,7 @@ from .sorting import default_sort_key
 from .kind import NumberKind
 from sympy.utilities.exceptions import SymPyDeprecationWarning
 from sympy.utilities.misc import as_int, func_name, filldedent
-from sympy.utilities.iterables import has_variety, sift, iterable, NotIterable
+from sympy.utilities.iterables import has_variety, sift
 from mpmath.libmp import mpf_log, prec_to_dps
 from mpmath.libmp.libintmath import giant_steps
 
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -127,7 +127,7 @@ class Expr(Basic, EvalfMixin):
 
         return expr.class_key(), args, exp, coeff
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         # hash cannot be cached using cache_it because infinite recurrence
         # occurs as hash is needed for setting cache dictionary keys
         h = self._mhash
@@ -146,7 +146,22 @@ class Expr(Basic, EvalfMixin):
         return self._args
 
     def __eq__(self, other):
-        return super().__eq__(other)
+        if not isinstance(other, Expr):
+            return super().__eq__(other)
+        # check for pure number expr
+        if  not (self.is_Number and other.is_Number) and (
+                type(self) != type(other)):
+            return False
+        a, b = self._hashable_content(), other._hashable_content()
+        if a != b:
+            return False
+        # check number *in* an expression
+        for a, b in zip(a, b):
+            if not isinstance(a, Expr):
+                continue
+            if a.is_Number and type(a) != type(b):
+                return False
+        return True
 
     # ***************
     # * Arithmetics *
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -162,8 +162,8 @@ class Expr(Basic, EvalfMixin):
                 continue
             if a.is_Number and type(a) != type(b):
                 result = False
-        if not result == super().__eq__(other):
-            print(self, other, type(self), type(other))
+        if not result is super().__eq__(other):
+            print(result, super().__eq__(other))
         return result
 
     # ***************
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -145,25 +145,6 @@ class Expr(Basic, EvalfMixin):
         been defined by a class. See note about this in Basic.__eq__."""
         return self._args
 
-    def __eq__(self, other):
-        # This is not responsible for comparisons with non Expr
-        if not isinstance(other, Expr):
-            return super().__eq__(other)
-        # check for pure number expr
-        if  not (self.is_Number and other.is_Number) and (
-                type(self) != type(other)):
-            return False
-        a, b = self._hashable_content(), other._hashable_content()
-        if a != b:
-            return False
-        # check number *in* an expression
-        for a, b in zip(a, b):
-            if not isinstance(a, Expr):
-                continue
-            if a.is_Number and type(a) != type(b):
-                return False
-        return True
-
     # ***************
     # * Arithmetics *
     # ***************
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -127,6 +127,15 @@ class Expr(Basic, EvalfMixin):
 
         return expr.class_key(), args, exp, coeff
 
+    def __hash__(self):
+        # hash cannot be cached using cache_it because infinite recurrence
+        # occurs as hash is needed for setting cache dictionary keys
+        h = self._mhash
+        if h is None:
+            h = hash((type(self).__name__,) + self._hashable_content())
+            self._mhash = h
+        return h
+
     def _hashable_content(self):
         """Return a tuple of information about self that can be used to
         compute the hash. If a class defines additional attributes,
@@ -135,6 +144,9 @@ class Expr(Basic, EvalfMixin):
         Defining more than _hashable_content is necessary if __eq__ has
         been defined by a class. See note about this in Basic.__eq__."""
         return self._args
+
+    def __eq__(self, other):
+        return super().__eq__(other)
 
     # ***************
     # * Arithmetics *
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -149,19 +149,22 @@ class Expr(Basic, EvalfMixin):
         if not isinstance(other, Expr):
             return super().__eq__(other)
         # check for pure number expr
+        result = True
         if  not (self.is_Number and other.is_Number) and (
                 type(self) != type(other)):
-            return False
+            result = False
         a, b = self._hashable_content(), other._hashable_content()
         if a != b:
-            return False
+            result = False
         # check number *in* an expression
         for a, b in zip(a, b):
             if not isinstance(a, Expr):
                 continue
             if a.is_Number and type(a) != type(b):
-                return False
-        return True
+                result = False
+        if not result == super().__eq__(other):
+            print(self, other, type(self), type(other))
+        return result
 
     # ***************
     # * Arithmetics *
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -146,19 +146,9 @@ class Expr(Basic, EvalfMixin):
         return self._args
 
     def __eq__(self, other):
-        if not isinstance(other, Basic):
-            if iterable(other, exclude=(str, NotIterable)
-                    ) and not hasattr(other, '_sympy_'):
-                # XXX iterable self should have it's own __eq__
-                # method if the path gives a false negative
-                # comparison
-                return False
-            try:
-                other = _sympify(other)
-            except (SympifyError, SyntaxError):
-                return NotImplemented
+        # This is not responsible for comparisons with non Expr
         if not isinstance(other, Expr):
-            return False
+            return super().__eq__(other)
         # check for pure number expr
         if  not (self.is_Number and other.is_Number) and (
                 type(self) != type(other)):
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -162,8 +162,8 @@ class Expr(Basic, EvalfMixin):
                 continue
             if a.is_Number and type(a) != type(b):
                 result = False
-        if not result is super().__eq__(other):
-            print(result, super().__eq__(other))
+        if not result is Basic.__eq__(self, other):
+            raise TypeError(result, Basic.__eq__(self, other), type(self), type(other), self.__repr__(), other.__repr__())
         return result
 
     # ***************
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -127,15 +127,6 @@ class Expr(Basic, EvalfMixin):
 
         return expr.class_key(), args, exp, coeff
 
-    def __hash__(self) -> int:
-        # hash cannot be cached using cache_it because infinite recurrence
-        # occurs as hash is needed for setting cache dictionary keys
-        h = self._mhash
-        if h is None:
-            h = hash((type(self).__name__,) + self._hashable_content())
-            self._mhash = h
-        return h
-
     def _hashable_content(self):
         """Return a tuple of information about self that can be used to
         compute the hash. If a class defines additional attributes,
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -38,7 +38,7 @@ from sympy.utilities.exceptions import SymPyDeprecationWarning
 
 _LOG2 = math.log(2)
 
-number_types=[]
+number_types:list[type] = []
 
 def comp(z1, z2, tol=None):
     """Return a bool indicating whether the error between z1 and z2
@@ -1484,7 +1484,7 @@ class Float(Number):
 # Add sympify converters
 converter[float] = converter[decimal.Decimal] = Float
 number_types.append(float)
-nubmer_types.append(decimal.Decimal)
+number_types.append(decimal.Decimal)
 # this is here to work nicely in Sage
 RealNumber = Float
 
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -1,3 +1,4 @@
+#pinging this file so it can be used in the review
 import numbers
 import decimal
 import fractions
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -766,7 +766,7 @@ class Number(AtomicExpr):
             elif other in (S.Infinity, S.NegativeInfinity):
                 return S.Zero
         return AtomicExpr.__truediv__(self, other)
-
+#ping
     def __eq__(self, other):
         raise NotImplementedError('%s needs .__eq__() method' %
             (self.__class__.__name__))
@@ -1375,7 +1375,7 @@ class Float(Number):
         if self._mpf_ == fzero:
             return 0
         return int(mlib.to_int(self._mpf_))  # uses round_fast = round_down
-
+#ping
     def __eq__(self, other):
         from sympy.logic.boolalg import Boolean
         try:
@@ -1873,7 +1873,7 @@ class Rational(Number):
 
     def __ceil__(self):
         return self.ceiling()
-
+#ping
     def __eq__(self, other):
         try:
             other = _sympify(other)
@@ -2244,7 +2244,7 @@ class Integer(Rational):
                 return Integer(other.p % self.p)
             return Rational.__rmod__(self, other)
         return Rational.__rmod__(self, other)
-
+#ping
     def __eq__(self, other):
         if isinstance(other, int):
             return (self.p == other)
@@ -3036,7 +3036,7 @@ class Infinity(Number, metaclass=Singleton):
 
     def __hash__(self):
         return super().__hash__()
-
+#ping
     def __eq__(self, other):
         return other is S.Infinity or other == float('inf')
 
@@ -3202,7 +3202,7 @@ class NegativeInfinity(Number, metaclass=Singleton):
 
     def __hash__(self):
         return super().__hash__()
-
+#ping
     def __eq__(self, other):
         return other is S.NegativeInfinity or other == float('-inf')
 
@@ -3333,7 +3333,7 @@ class NaN(Number, metaclass=Singleton):
 
     def __hash__(self):
         return super().__hash__()
-
+#ping
     def __eq__(self, other):
         # NaN is structurally equal to another NaN
         return other is S.NaN
@@ -3457,7 +3457,7 @@ class NumberSymbol(AtomicExpr):
 
     def _eval_evalf(self, prec):
         return Float._new(self._as_mpf_val(prec), prec)
-
+#ping
     def __eq__(self, other):
         try:
             other = _sympify(other)
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -38,6 +38,7 @@ from sympy.utilities.exceptions import SymPyDeprecationWarning
 
 _LOG2 = math.log(2)
 
+number_types=[]
 
 def comp(z1, z2, tol=None):
     """Return a bool indicating whether the error between z1 and z2
@@ -1378,9 +1379,9 @@ class Float(Number):
 #ping
     def __eq__(self, other):
         from sympy.logic.boolalg import Boolean
-        try:
+        if type(other) in number_types:
             other = _sympify(other)
-        except SympifyError:
+        if not isinstance(other, Basic):
             return NotImplemented
         if isinstance(other, Boolean):
             return False
@@ -1482,7 +1483,8 @@ class Float(Number):
 
 # Add sympify converters
 converter[float] = converter[decimal.Decimal] = Float
-
+number_types.append(float)
+nubmer_types.append(decimal.Decimal)
 # this is here to work nicely in Sage
 RealNumber = Float
 
@@ -1875,9 +1877,9 @@ class Rational(Number):
         return self.ceiling()
 #ping
     def __eq__(self, other):
-        try:
+        if type(other) in number_types:
             other = _sympify(other)
-        except SympifyError:
+        if not isinstance(other, Basic):
             return NotImplemented
         if not isinstance(other, Number):
             # S(0) == S.false is False
@@ -2507,7 +2509,7 @@ class Integer(Rational):
 
 # Add sympify converters
 converter[int] = Integer
-
+number_types.append(int)
 
 class AlgebraicNumber(Expr):
     """Class for representing algebraic numbers in SymPy. """
@@ -3459,9 +3461,9 @@ class NumberSymbol(AtomicExpr):
         return Float._new(self._as_mpf_val(prec), prec)
 #ping
     def __eq__(self, other):
-        try:
+        if type(other) in number_types:
             other = _sympify(other)
-        except SympifyError:
+        if not isinstance(other, Basic):
             return NotImplemented
         if self is other:
             return True
@@ -4059,6 +4061,7 @@ def sympify_fractions(f):
     return Rational(f.numerator, f.denominator, 1)
 
 converter[fractions.Fraction] = sympify_fractions
+number_types.append(fractions.Fraction)
 
 if HAS_GMPY:
     def sympify_mpz(x):
@@ -4072,26 +4075,28 @@ if HAS_GMPY:
 
     converter[type(gmpy.mpz(1))] = sympify_mpz
     converter[type(gmpy.mpq(1, 2))] = sympify_mpq
-
+    number_types.append(type(gmpy.mpz(1)))
+    number_types.append(type(gmpy.mpq(1, 2)))
 
 def sympify_mpmath_mpq(x):
     p, q = x._mpq_
     return Rational(p, q, 1)
 
 converter[type(mpmath.rational.mpq(1, 2))] = sympify_mpmath_mpq
-
+number_types.append(type(mpmath.rational.mpq(1, 2)))
 
 def sympify_mpmath(x):
     return Expr._from_mpmath(x, x.context.prec)
 
 converter[mpnumeric] = sympify_mpmath
-
+number_types.append(mpnumeric)
 
 def sympify_complex(a):
     real, imag = list(map(sympify, (a.real, a.imag)))
     return real + S.ImaginaryUnit*imag
 
 converter[complex] = sympify_complex
+number_types.append(complex)
 
 from .power import Pow, integer_nthroot
 from .mul import Mul
--- a/sympy/matrices/repmatrix.py
+++ b/sympy/matrices/repmatrix.py
@@ -46,7 +46,7 @@ class RepMatrix(MatrixBase):
     #
 
     _rep: DomainMatrix
-
+#ping
     def __eq__(self, other):
         # Skip sympify for mutable matrices...
         if not isinstance(other, RepMatrix):
--- a/sympy/polys/matrices/domainmatrix.py
+++ b/sympy/polys/matrices/domainmatrix.py
@@ -1641,7 +1641,7 @@ class DomainMatrix:
 
         """
         return cls.from_rep(DDM.ones(shape, domain))
-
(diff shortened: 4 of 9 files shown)


sympy/core/expr.py
```diff
@@ -146,19 +146,9 @@ def _hashable_content(self):
         return self._args
 
     def __eq__(self, other):
-        if not isinstance(other, Basic):
```
oscarbenjamin, line 149: Is this method needed at all? Can the Basic method not also work here?

ThePauliPrinciple: Removing it caused failures in the test. Probably because of the check "number *in* an expression" part.

sympy/core/numbers.py
```diff
@@ -765,7 +766,7 @@ def __truediv__(self, other):
             elif other in (S.Infinity, S.NegativeInfinity):
                 return S.Zero
         return AtomicExpr.__truediv__(self, other)
-
+#ping
     def __eq__(self, other):
```
ThePauliPrinciple, line 770: Numbers forces its subclasses to define a `__eq__`

sympy/core/numbers.py
```diff
@@ -1374,7 +1375,7 @@ def __int__(self):
         if self._mpf_ == fzero:
             return 0
         return int(mlib.to_int(self._mpf_))  # uses round_fast = round_down
-
+#ping
     def __eq__(self, other):
```
ThePauliPrinciple, line 1379: It would be nice if this only uses `_sympify` on appropriate classes, rather than any, but which are those?

It might be better to replace `isinstance(other, Boolean)` with `other.is_Bool` to avoid the circular import.

ThePauliPrinciple: I think it is prefered to use `return NotImplemented` over `return other.__eq__(self)` and let python try the other class.

sympy/core/numbers.py
```diff
@@ -1872,7 +1873,7 @@ def __floor__(self):
 
     def __ceil__(self):
         return self.ceiling()
-
+#ping
     def __eq__(self, other):
         try:
             other = _sympify(other)
```
ThePauliPrinciple, line 1879: Similar: only convert when appropriate, use `NotImplemented` and possibly use `other.is_Number` instead of `isinstance(other, Number)

sympy/core/numbers.py
```diff
@@ -3456,7 +3457,7 @@ def approximation(self, number_cls):
 
     def _eval_evalf(self, prec):
         return Float._new(self._as_mpf_val(prec), prec)
-
+#ping
     def __eq__(self, other):
         try:
             other = _sympify(other)
```
ThePauliPrinciple, line 3463: Again, only convert when appropriate. I also don't think the last if statement `if other.is_Number and self.is_irrational:` does anything, since it returns `False` regardless of whether the if statement is true or false

sympy/matrices/repmatrix.py
```diff
@@ -46,7 +46,7 @@ class RepMatrix(MatrixBase):
     #
 
     _rep: DomainMatrix
-
+#ping
     def __eq__(self, other):
         # Skip sympify for mutable matrices...
         if not isinstance(other, RepMatrix):
```
ThePauliPrinciple, line 52: Only do `_sympify` for `MatrixBase` and mpmath matrices. This could possibly be extended to also compare with lists or numpy arrays.

sympy/polys/matrices/domainmatrix.py
```diff
@@ -1641,7 +1641,7 @@ def ones(cls, shape, domain):
 
         """
         return cls.from_rep(DDM.ones(shape, domain))
-
+#ping
     def __eq__(A, B):
```
ThePauliPrinciple, line 1645: looks good to me.

sympy/polys/polymatrix.py
```diff
@@ -186,7 +186,7 @@ def to_poly(v):
             return to_poly(dm[i, j].element)
         else:
             return self.from_dm(dm[i, j])
-
+#ping
     def __eq__(self, other):
```
ThePauliPrinciple, line 190: looks good to me.

sympy/core/expr.py
oscarbenjamin: Which tests failed?

oscarbenjamin:
With `Expr.__eq__` the failed tests are:
https://github.com/sympy/sympy/blob/82cc5a598f236b32b5f7adf530802503c121b080/sympy/core/tests/test_complex.py#L101-L105
https://github.com/sympy/sympy/blob/82cc5a598f236b32b5f7adf530802503c121b080/sympy/core/tests/test_numbers.py#L1882
https://github.com/sympy/sympy/blob/82cc5a598f236b32b5f7adf530802503c121b080/sympy/core/tests/test_power.py#L315
Those all make sense.

Then there's a bunch of failures all in the ODE tests like this that I don't understand:
```
______ sympy/solvers/ode/tests/test_single.py:test_Riccati_special_minus2 ______
  File "/home/runner/work/sympy/sympy/sympy/solvers/ode/tests/test_single.py", line 553, in test_Riccati_special_minus2
    _ode_solver_test(_get_examples_ode_sol_riccati)
  File "/home/runner/work/sympy/sympy/sympy/solvers/ode/tests/test_single.py", line 185, in _ode_solver_test
    result = _test_particular_example(example['hint'], example, solver_flag=True)
  File "/home/runner/work/sympy/sympy/sympy/solvers/ode/tests/test_single.py", line 235, in _test_particular_example
    dsolve_sol = dsolve(eq, func, simplify=simplify_flag,hint=our_hint)
  File "/home/runner/work/sympy/sympy/sympy/solvers/ode/ode.py", line 605, in dsolve
    hints = _desolve(eq, func=func,
  File "/home/runner/work/sympy/sympy/sympy/solvers/deutils.py", line 234, in _desolve
    raise ValueError(string + str(eq) + " does not match hint " + hint)
ValueError: ODE f(x)**2 + 2*Derivative(f(x), x) - f(x)/x + 3/x**2 does not match hint Riccati_special_minus2
```

oscarbenjamin:
I can't reproduce the failures in the ODE tests locally with this branch.

ThePauliPrinciple:
> I can't reproduce the failures in the ODE tests locally with this branch.

I can reproduce them locally. I set up my environment like described in https://github.com/sympy/sympy/wiki/Development-workflow.

oscarbenjamin:
But how exactly are you running the tests:
```console
$ bin/test sympy/solvers/ode/tests/test_single.py -k test_nth_order_linear_euler_eq_nonhomogeneous_undetermined_coefficients 
============================== test process starts ===============================
executable:         /home/oscar/current/sympy/sympy.git/38venv/bin/python  (3.8.9-final-0) [CPython]
architecture:       64-bit
cache:              yes
ground types:       gmpy 2.0.8
numpy:              1.21.3
random seed:        36333991
hash randomization: on (PYTHONHASHSEED=3697924441)

sympy/solvers/ode/tests/test_single.py[1] .                                   [OK]

_________________________________ slowest tests __________________________________
sympy/solvers/ode/tests/test_single.py::test_nth_order_linear_euler_eq_nonhomogeneous_undetermined_coefficients - Took 21.285 seconds
=================== tests finished: 1 passed, in 27.30 seconds ===================
$ pytest sympy/solvers/ode/tests/test_single.py::test_nth_order_linear_euler_eq_nonhomogeneous_undetermined_coefficients 
============================== test session starts ===============================
platform linux -- Python 3.8.9, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
architecture: 64-bit
cache:        yes
ground types: gmpy 2.0.8

rootdir: /home/oscar/current/sympy/sympy.git, configfile: pytest.ini
plugins: forked-1.3.0, instafail-0.4.2, xdist-2.4.0
collected 1 item                                                                 

sympy/solvers/ode/tests/test_single.py .                                   [100%]

=============================== 1 passed in 28.84s ===============================
```

ThePauliPrinciple:
`bin/test sympy/solvers/ode/` gives me the errors. I don't get errors when running the single example tests.

oscarbenjamin:
> I don't get errors when running the one specific test.

Maybe the problem is something to do with the cache (which uses `__eq__`).

ThePauliPrinciple:
I'll try removing the `__hash__` definition in `Expr`, that might be the problem.

oscarbenjamin:
> `bin/test sympy/solvers/ode/` gives me the errors

I was able to reproduce those errors before the last commit. Testing with the new commit now... looks like the same errors still.

ThePauliPrinciple:
It's not clear to me whether we're looking at a bug in the caching, or that we're now missing a cache hit because the `__eq__` is different and this should be the result when `Expr.__eq__ = Basic.__eq__`

oscarbenjamin:
Maybe if a check is added to `Expr.__eq__` for cases when it would return a different value than `Basic.__eq__` then it would be possible to run the ODE tests and get a debugger that can show which `__eq__` is actually behaving differently. This would probably be easy to fix if knew exactly what `a` and `b` are such that `a == b` no longer gives the same value.

ThePauliPrinciple:
It looks to me that the problem might be that `Basic.__eq__` is reporting `False` when comparing `Function('f')(x)` with itself:
```
___________________________________________________________________________________________________________________________________________________________________________________
___________________________________________________ sympy/solvers/ode/tests/test_systems.py:test_sysode_linear_neq_order1_type2 ___________________________________________________
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/solvers/ode/tests/test_systems.py", line 1053, in test_sysode_linear_neq_order1_type2
    assert dsolve(eqs3) == sol3
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/solvers/ode/ode.py", line 559, in dsolve
    sol = dsolve_system(eq, funcs=func, ics=ics, doit=True)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/solvers/ode/systems.py", line 2106, in dsolve_system
    canon_eqs = canonical_odes(eqs, funcs, t)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/solvers/ode/systems.py", line 1077, in canonical_odes
    canon_eqs = solve(eqs, *[func.diff(t, order[func]) for func in funcs], dict=True)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/solvers/solvers.py", line 1098, in solve
    solution = _solve_system(f, symbols, **flags)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/solvers/solvers.py", line 1758, in _solve_system
    subsol = _solve_system(subexpr, subsyms, **flags)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/solvers/solvers.py", line 1789, in _solve_system
    poly = g.as_poly(*symbols, extension=True)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/expr.py", line 1133, in as_poly
    poly = Poly(self, *gens, **args)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/polys/polytools.py", line 169, in __new__
    return cls._from_expr(rep, opt)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/polys/polytools.py", line 299, in _from_expr
    return cls._from_dict(rep, opt)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/polys/polytools.py", line 243, in _from_dict
    domain, rep = construct_domain(rep, opt=opt)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/polys/constructor.py", line 378, in construct_domain
    result = _construct_composite(coeffs, opt)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/polys/constructor.py", line 238, in _construct_composite
    domain = ground.poly_ring(*gens)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/polys/domains/domain.py", line 873, in poly_ring
    return PolynomialRing(self, symbols, order)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/polys/domains/polynomialring.py", line 25, in __init__
    ring = PolyRing(symbols, domain_or_ring, order)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/polys/rings.py", line 206, in __new__
    obj = _ring_cache.get(_hash_tuple)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/expr.py", line 166, in __eq__
    raise TypeError(result, Basic.__eq__(self, other), type(self), type(other), self.__repr__(), other.__repr__())
TypeError: (True, False, f, f, 'f(x)', 'f(x)')
```

ThePauliPrinciple:
Seems like `type(self) is type(other)` is `False`. Not sure what causes this though, assumptions seem the same on these, `hash` is also the same for `self` and `other`.

ThePauliPrinciple:
Alright looks like using `!=` instead of `not is` solved the issue. I wonder if the current fails are actually the behaviour you would expect. If `S(2) == S(2.0)`, then why shouldn't `Symbol('x')**S(2) == Symbol('x')**S(2.0)`

oscarbenjamin:
> Alright looks like using `!=` instead of `not is` solved the issue.

Can you elaborate on this? I don't understand what was not working and what was fixed and where the fix was etc.

oscarbenjamin:
> I wonder if the current fails are actually the behaviour you would expect. If `S(2) == S(2.0)`, then why shouldn't `Symbol('x')**S(2) == Symbol('x')**S(2.0)`

See #20033

ThePauliPrinciple:
> > Alright looks like using `!=` instead of `not is` solved the issue.
> 
> Can you elaborate on this? I don't understand what was not working and what was fixed and where the fix was etc.

In `Basic.__eq__` I changed `type(self) not is type(other)` to `type(self) != type(other)`. For some reason there were undefined functions `Function('f')` that were equal `==` but not `is` that caused problems in the caching in the ode solver.

asmeurer:
This seems very similar to https://github.com/sympy/sympy/pull/22184. I think for many cases we can avoid sympify entirely, for instance, Integer(1) == 1 does need need to sympify the rhs to perform the comparison. 

In general, my experience is that doing targeted comparison in subclass `__eq__` methods is more performant than having a general purpose `__eq__` in a superclass, which can matter since equality checking is done a lot.

ThePauliPrinciple:
Yes, avoiding `_sympify` as much as possible is the goal here. There just aren't a lot of classes that compare equally to another class. Exceptions already all have their own `__eq__` logic. But why should `Expr` and `Basic` be different when comparing?

Currently in master:
```python
>>> from sympy import *
>>> Expr(S(2)) == Expr(S(2.0))
False
>>> Basic(S(2)) == Basic(S(2.0))
True
```
This came up because recently there was an update to `Expr.__eq__` to suppress converting `dict` to `Dict` when comparing with a completely unrelated type. This then caused the creation of invalid `Dict`s, e.g. with `list` or `None` in them. However, this then was only solved for `Expr`s, not all `Basic`s. So I believe that there should be a single superclass `__eq__` and then targeted equality checks when appropriate in subclasses that revert to the superclass `__eq__` when there is no known equality check for the `target` type.

oscarbenjamin:
I hadn't realised that the Float special casing only applied for Expr:
```python
In [12]: x**2 == x**2.0
Out[12]: False

In [13]: FiniteSet(2) == FiniteSet(2.0)
Out[13]: True
```
That's a bug I think and it means that the code that handles this should apply to `Basic.__eq__` as well as `Expr.__eq__`. In any case I don't think that there should be a distinction between these two `__eq__` methods so `Expr.__eq__` shouldn't be defined.

Ideally there wouldn't need to be any special-casing for Float but it is needed because we don't yet have #20033 so there is a distinction between how expressions involving Float are handled as compared to equality comparisons for individual Floats.

ThePauliPrinciple:
> ```python
> In [13]: FiniteSet(2) == FiniteSet(2.0)
> Out[13]: True
> ```


Apparently this is specifically tested for (so not a bug, but a feature):
```
 _______________ sympy/sets/tests/test_sets.py:test_finite_basic ________________
  File "/home/runner/work/sympy/sympy/sympy/sets/tests/test_sets.py", line 1009, in test_finite_basic
    assert FiniteSet(1.0) == FiniteSet(1)
AssertionError
```

oscarbenjamin:
> Apparently this is specifically tested for (so not a bug, but a feature):

The test could be changed if it makes the behaviour of `__eq__` consistent in general.

asmeurer:
I guess the difference with https://github.com/sympy/sympy/pull/22184 is that there I explicitly check for `int` in the superclass and return False, based on the assumption that anything that will equal an int will define its own `__eq__`. This prevents calling `_sympify` on int objects when the equality check would be False, like `x == 1`.

This feels a bit fragile, but I'm not sure it would break anything. In library code, `Integer` is the only class that can be equal to `int`, which does define `__eq__`. If a user defined their own class that can equal `int`, I don't think they could get away without defining `__eq__` anyway (it would need to equal `Integer` too). 

Even so, the real question is if the performance gain is substantial enough. It feels like it should be, but I wasn't able to really confirm if it actually was in https://github.com/sympy/sympy/pull/22184.

ThePauliPrinciple:
> If a user defined their own class that can equal `int`, I don't think they could get away without defining `__eq__` anyway (it would need to equal `Integer` too).

They could define a `_sympy_` method to convert it to a SymPy object (or possibly an `int`).

oscarbenjamin:
> Even so, the real question is if the performance gain is substantial enough.

No the real question is avoiding bugs. The performance might not be that significant because where things are using `__eq__` in an inner loop we shouldn't mix Basic and non-Basic anyway. There might however be some cases where `sympify` is extremely expensive although using `_sympify` might help to avoid those e.g. this case doesn't happen with `_sympify`:
```python
In [10]: %time ok = sympify(np.ones((1000, 1000)))
CPU times: user 58.7 s, sys: 188 ms, total: 58.9 s
Wall time: 58.9 s
```

The reason for the change proposed here is that many objects should not be passed to `_sympify/sympify` which can do strange things with them. The reason that @ThePauliPrinciple started looking at this is because of something like this:
```python
In [2]: Basic() == ('asd', lambda: x)
/home/oscar/current/sympy/sympy.git/sympy/core/sympify.py:453: SymPyDeprecationWarning: 

String fallback in sympify has been deprecated since SymPy 1.6. Use
sympify(str(obj)) or sympy.core.sympify.converter or obj._sympy_
instead. See https://github.com/sympy/sympy/issues/18066 for more
info.

  SymPyDeprecationWarning(
Out[2]: False
```
Here `Basic.__eq__` attempted to sympify the tuple which sympifies to Tuple which then tries to sympify the elements of the tuple. This particular tuple has non-Basic elements because it isn't supposed to be anything that would sympify to Basic. The `__eq__` method should not attempt to sympify arbitrary objects. It is bad-form for `__eq__` to either raise an error or give a warning.

asmeurer:
> No the real question is avoiding bugs. The performance might not be that significant because where things are using __eq__ in an inner loop we shouldn't mix Basic and non-Basic anyway. There might however be some cases where sympify is extremely expensive although using _sympify might help to avoid those e.g. this case doesn't happen with _sympify:

It's very hard to avoid it though. `__eq__` gets called implicitly all over the place. For example, a check like `if 1 in container` (where container is a list/tuple/dict/set) will perform a `__eq__` test on every element of the container with a builtin object (`1`). You can avoid this by using `if S(1) in container`, but this starts to become unnecessarily verbose. And while it prevents calling `sympify(1)` `len(container)` times, it does still call it once, whereas with this approach it gets called zero times. If you profile some high level operation, you'll find that `__eq__` methods are called thousands of times. 

> The reason for the change proposed here is that many objects should not be passed to _sympify/sympify which can do strange things with them. 

I think we are talking about different things. 

The issue you showed is due to using `sympify` instead of `_sympify`. `_sympify` (with the underscore) should be strict enough to only accept objects it knows about. All library code ought to be using `_sympify` (this in particular means deprecating string inputs to most functions, which has been discussed elsewhere). If this PR is moving more code from `sympify` to `_sympify` that's a good thing.

The change in my PR is about is doing a more forceful bypassing of _sympify entirely in `Basic.__eq__`. This has no impact on behavior because it would only avoid calling `_sympify` on types for which it would definitely be safe, like `int` or `float`. 

By the way, the string fallback thing was deprecated primarily because of performance (https://github.com/sympy/sympy/issues/15416).

ThePauliPrinciple:
This PR does a lot more forceful bypassing of `_sympify` in `Basic.__eq__` than ignoring `int`. It now only does it if `other` has `_sympy_`. Then if a subclass of `Basic` wants to compare to `True` with a non-`Basic` it is that class' responsibility to convert the type of `other`. There are very few classes that actually need to do this. Precisely those that already had custom `__eq__` methods.

oscarbenjamin:
> The issue you showed is due to using `sympify` instead of `_sympify`. `_sympify` (with the underscore) should be strict enough to only accept objects it knows about

The same problem is seen with `_sympify`:
```python
In [1]: Basic() == ('asd', object())
/home/oscar/current/sympy/sympy.git/sympy/core/sympify.py:453: SymPyDeprecationWarning: 

String fallback in sympify has been deprecated since SymPy 1.6. Use
sympify(str(obj)) or sympy.core.sympify.converter or obj._sympy_
instead. See https://github.com/sympy/sympy/issues/18066 for more
info.

  SymPyDeprecationWarning(
Out[1]: False
```
The call to `_sympify` is here in `Basic.__eq__`:
https://github.com/sympy/sympy/blob/a3e0f9c810cb78d8ca68b128757551783564f218/sympy/core/basic.py#L348

Of course `_sympify` could be improved although I think more time is needed to remove that particular deprecated functionality.

In any case though for me the fundamental issue is that we really should not call `sympify` or `_sympify` on fully arbitrary objects. If someone does `sympy.cos(obj)` then I think it's reasonable to call `_sympify`: the operation couldn't possibly make sense if the object was not sympifiable. However with `a == b` you have to expect fully arbitrary objects and so it should not be presumed that it is reasonable to even try to sympify the other operand.

The number of situations where a sympifiable can compare equal to Basic is limited and most importantly it is connected to the types of both operands in `a == b` so the scope of any comparison is best kept to those relevant classes.

Also `sympify` performs the conversion in the wrong direction. The `Integer.__eq__` method should not convert an `int` to an `Integer` so that they can then compare their internal `int`s: `Integer.__eq__` should compare with an int by comparing it's `.p` (`int`) attribute with the other operand.

asmeurer:
Another thing to remember here is that custom classes can always implement their own `__eq__`. As long as `Basic.__eq__` and other `__eq__` methods properly return NotImplemented instead of False on unknown objects, custom objects can inject themselves into `==` however they like.

> This PR does a lot more forceful bypassing of _sympify in Basic.__eq__ than ignoring int. It now only does it if other has _sympy_. Then if a subclass of Basic wants to compare to True with a non-Basic it is that class' responsibility to convert the type of other. There are very few classes that actually need to do this. Precisely those that already had custom __eq__ methods.

I see. I just checked the diff, and we also need to check if the type is in the sympify converter dict. Otherwise users lose the ability to register a custom type as being interoperable with SymPy.  In other words, this should continue to work

```py
>>> class CustomZero:
...     pass
>>> from sympy.core.sympify import converter
>>> converter[CustomZero] = lambda x: Integer(0)
>>> S(0) == CustomZero()
True
```

The converter method is needed because sometimes you don't have control over the object to add an `__eq__` method to it. 

Aside from that this looks like the right approach. 

> The same problem is seen with _sympify:

Right, it's in the Tuple constructor. The same thing happened in https://github.com/sympy/sympy/issues/15416#issuecomment-453690358. A rogue sympify call from somewhere completely unexpected led to a 45x performance slowdown. To avoid this, we have to replace sympify with _sympify everywhere. For public APIs this cleans things up by disallowing string inputs, but even for internal APIs it can improve the performance.

My idea is that (after a deprecation) we ban non-strict sympify in library code, except in a few select places where accepting string inputs actually makes sense. This is the issue for it https://github.com/sympy/sympy/issues/11003.

oscarbenjamin:
> > The same problem is seen with _sympify:
> 
> Right, it's in the Tuple constructor.

Okay so it is. That should be fixed:
https://github.com/sympy/sympy/blob/a3e0f9c810cb78d8ca68b128757551783564f218/sympy/core/containers.py#L52

oscarbenjamin:
> My idea is that (after a deprecation) we ban non-strict sympify in library code, except in a few select places where accepting string inputs actually makes sense. This is the issue for it #11003.

I see so much string input being used in the wild that this makes me nervous. I don't know how to define where it "makes sense".

oscarbenjamin:
> we also need to check if the type is in the sympify converter dict. Otherwise users lose the ability to register a custom type as being interoperable with SymPy

This is a good point. There is probably a lot of downstream code that depends on this.

ThePauliPrinciple:
I tried changing to `_sympify` in `Tuple` before, but there were just too many cases where it was explictly expected to work that way.

If you check for being in `converter` then you'll also hit `dict` and `tuple`, which is exactually something you want to avoid converting to most of the times, unless you're comparing with `Dict` or `Tuple`. So I'm not sure how to deal with that. Perhaps there should be a different dict for user types and library types?

oscarbenjamin:
> I tried changing to `_sympify` in `Tuple` before, but there were just too many cases where it was explictly expected to work that way.

There are a lot of badly designed tests in SymPy. The work you are doing will naturally throw these up. It is possible to change those tests if it is what is needed for bringing SymPy into a consistent state.

ThePauliPrinciple:
I feel like a lot of issues with changing `sympify` to `_sympify` in `Tuple` would already be solved if `sympify(tuple())` has `Tuple` use `sympify` and `_sympify(tuple())` has it use `_sympify`. So some kind of memory of "strictness". Not sure if that would be easy to accomplish though.

ThePauliPrinciple:
Perhaps it would also be helpful if the underlying problem is solved. Which is the inconvenience of converting input arguments. Say I want to use `_sympify` but also convert any `str` to `Symbol`, I can't do that without writing some custom code. Currently this is solved in a lot of different ways e.g. with explicit logic or some recursive function, or `sympify` is used and hope for the best.

I think the most sought after features:
- apply `_sympify` element wise on arbitrarily structured iterables (without converting those iterables to SymPy objects)
- convert `str` to `Symbol`
- have some way to define additional rules only relevant for "my" case ( I suppose the above is like this too, it just appears a lot)

`_sympify( iterate = False, local_converter={})`
where `iterate=True` would make it iterate over python iterables and leave them intact and `local_converter` gets appended to `converter` (overriding `converter`) so that you can have the "rules" as you want them (e.g. `local_converter = {str: Symbol}`)

oscarbenjamin:
I still think that we shouldn't be depending on sympify in `__eq__`. As I said above it's performing the conversion in the wrong direction i.e. `Tuple.__eq__` should compare with a tuple by comparing its own internal tuple:
```python
def __eq__(self, other):
    if isinstance(other, tuple):
        return self.args == other
```
Converting other to Tuple so that they can then compare their internal args tuples is backwards.

Maybe we do need to continue to check `converters` for the benefit of downstream code that might be depending on that but within the SymPy codebase we should not use `sympify` or `_sympify` as part of `__eq__`.

ThePauliPrinciple:
For `Tuple` this is trivial, but I couldn't figure out how to avoid `_sympify` with `Number` and subclasses. 

Can you show how this can be accomplished in `numbers.py`?

If that's possible, then we can just change `converters` to  `_converters` and copy it to `converters` and the end of the SymPy import (in case people somehow extract information from `converters`). Then we can choose to only use user-defined sympifies based on comparing those two dicts.

oscarbenjamin:
> Can you show how this can be accomplished in `numbers.py`?

It takes some work but it's just a case of inverting and combining the logic used by `sympify`, `Number.__new__` and `Number.__eq__`. Look at how `converters` is used in `numbers.py`.

ThePauliPrinciple:
> Look at how `converters` is used in `numbers.py`.

How do you mean? `numbers.py` adds a bunch of conversions to the dictionary and that is all it does with it.

oscarbenjamin:
> > Look at how `converters` is used in `numbers.py`.
> 
> How do you mean? `numbers.py` adds a bunch of conversions to the dictionary and that is all it does with it.

Each of those conversions is something that needs to have logic in the corresponding `__eq__` methods e.g.:
https://github.com/sympy/sympy/blob/a3e0f9c810cb78d8ca68b128757551783564f218/sympy/core/numbers.py#L4429-L4432
```python
class Rational:
    def __eq__(self, other):
        if isinstance(other, Number):
            # logic for sympy numbers
        elif isinstance(other, Fraction):
            return self.p == other.numerator and self.q == other.denominator
        ...
```

ThePauliPrinciple:
The problem to me is that it doesn't only need logic in the corresponding `__eq__` method. Any `__eq__` method that currently uses `isinstance(other, Rational)`/`.is_Rational` will also have to define a statement for `Fraction`, everytime with exactly the same conversion logic in it, which is also probably the same as the `__new__` method of `Rational` in the first place. Having the same code in multiple places is something you want to avoid.

oscarbenjamin:
> Any `__eq__` method that currently uses `isinstance(other, Rational)`/`.is_Rational` will also have to define a statement for `Fraction`, everytime with exactly the same conversion logic in it

Not necessarily. There are several possibilities:
1. There can be a general `Number.__eq__` method that handles all the cases.
2. The fall-back case for `Rational.__eq__` could be converting to `Fraction/MPQ` [note 1].
3. The conversions will not necessarily be exactly the same in all cases.
4. If the conversions are the same then we can just put them in a function like `as_integer_ratio` or something.

There are only a few classes with nontrivial logic here.

Note 1: Ideally Rational would internally use Fraction or gmpy.mpq in which case it could just delegate the logic to its internal attribute just like `Integer` can (https://github.com/sympy/sympy/issues/20598#issuecomment-902842183).

ThePauliPrinciple:
Attempt at breaking this down into smaller parts:

1. [ ] Remove old support for python2: https://github.com/sympy/sympy/pull/22634
2. [ ] Add the same iterable skipping in `Basic` as in `Expr` (not trivial, `Dict` and `FiniteSet` will need an `__eq__` dealing with ordering). 
3. [ ] Merge `Expr.__eq__` into `Basic.__eq__` (needs changed tests)
4. [ ] Remove `_sympify` from `Basic.__eq__`
5. [ ] Rewriting of the `numbers.py` `__eq__` system
6. [ ] It would be good to have a look at `Array`, `Matrix` etc (comparison with `np.array` and nested `list`s etc)

4 still requires some details (how to stay compatible with downstream) but perhaps it becomes clearer when 1-3 are introduced. Personally not interested in doing 5 and 6, but it would be great to have.

ThePauliPrinciple:
I got stuck on point 3. 

For point 4, I'm wondering if something like this can work:

```python
    def _do_eq_sympify(self, other):
        tother = type(other)
        if tother in converter or hasattr(other,'_sympy_'):
            #if tother is known by sympy, but does not have a user defined conversion,
            #it should have already been handled explicitly in any class that could
            #compare to it (e.g. dicts, tuples, floats)
            if tother in _converter and converter[tother] == _converter[tother]):
                return NotImplemented
            try:
                return self == _sympify(other)
            except:
                return NotImplemented
        return NotImplemented

    def __eq__(self, other):
        """Return a boolean indicating whether a == b on the basis of
        their symbolic trees.

        This is the same as a.compare(b) == 0 but faster.

        Notes
        =====

        If a class that overrides __eq__() needs to retain the
        implementation of __hash__() from a parent class, the
        interpreter must be told this explicitly by setting
        __hash__ : Callable[[object], int] = <ParentClass>.__hash__.
        Otherwise the inheritance of __hash__() will be blocked,
        just as if __hash__ had been explicitly set to None.

        References
        ==========

        from http://docs.python.org/dev/reference/datamodel.html#object.__hash__
        """
        if self is other:
            return True

        if not isinstance(other, Basic):
            return self._do_eq_sympify(other)

        if type(self) != type(other):
            return NotImplemented

        return self._hashable_content() == other._hashable_content()
```
Where `_converter` would contain any conversion defined in SymPy and `converter` would contain those in `_converter` and in addition any user defined conversions (possibly overriding something in `_converter`.

I decided to use `self == _sympify(other)` because the result of `sympify` might be an object whose `__eq__` we need to resolve the equality, but returning `NotImplemented` would not trigger it since it isn't part of the original `self == other`.

It's essentially a generalisation of what @asmeurer suggested in  #22184: rather than only excluding `int` (or `iterables` as is currently implemented in SymPy), it excludes any type that SymPy should be aware of.

I'm not really sure how to implement the `_converter` `converter` distinction, so that probably indicates there is a better way to go about this, although I do believe it does exactly what we are looking for: avoid `_sympify` as much as possible in `__eq__` but keep compatibility with downstream projects.

ThePauliPrinciple:
Looking for some suggestions on point 4. In particular, the requirement to `simpify` any user made additions to `converter` is troublesome, it always leads to some circular imports when I try to find a solution for it.

ThePauliPrinciple:
In particular, here I have to add code to `sympy/__init__.py` in order to avoid such circular imports

oscarbenjamin:
> 4 still requires some details (how to stay compatible with downstream)

Do you know of any downstream code that alters the sympify converters?

ThePauliPrinciple:
Not in any library, but this comment shows an example of what is expected to work:
https://github.com/sympy/sympy/pull/22607#issuecomment-989333047

oscarbenjamin:
For full equivalence all of mro needs to be checked e.g.:
```python
In [3]: from sympy.core.sympify import converter

In [4]: class A: pass

In [5]: converter[A] = lambda a: S.Zero

In [6]: class B(A): pass

In [7]: S.Zero == B()
Out[7]: True
```
The constraint of needing to satisfy this seems quite awkward to me. It would be better if the user-customisable part of sympify was fully separate from the inbuilt part.

ThePauliPrinciple:
That is a good point. I had considered simply storing the SymPy values in something like `_converter` and setup `converter` as an empty list. People might be relying on what is defined in `converter` somehow though. It might be possible to create a custom object that acts similar to a dict but allows storing and accessing the two different use cases without giving up compatibility.

ThePauliPrinciple:
Is there any documentation available in SymPy about which non-SymPy types should equal to SymPy types in the context of the `numbers` module? e.g. `Rational(1,2) =?= 0.5`

oscarbenjamin:
> Is there any documentation available in SymPy about which non-SymPy types should equal to SymPy types in the context of the `numbers` module? e.g. `Rational(1,2) =?= 0.5`

No there isn't. Relevant things to test/consider are fractions.Fraction, decimal.Decimal, gmpy.mpz, gmpy2.mpq, gmpy2.mpfr, gmpy2.mpc, mpmath.mpf, mpmath.mpc, numpy.float64 etc.

ThePauliPrinciple:
I am not familiar enough with `numbers.py` to continue working on this. As far as I am concerned, this "proof of concept" had a successful result. Thanks everyone for helping!

PR #22615, opened 2021-12-06
Added refine-support for Min and Max

oscargus:

Related to https://github.com/sympy/sympy/discussions/22515

Just realized that there is a Discussions-tab... Found https://github.com/sympy/sympy/discussions/22515 and decided to make a stab at it.

Not really happy with the code, but can currently not figure out a better way of doing it... 

#### Other comments

I would have expected something else than an error here:
`refine(Max(a, b, c), Q.positive(a) & Q.negative(c))`

```

  File "C:\Users\Oscar\AppData\Local\Temp/ipykernel_22316/742831045.py", line 1, in <module>
    refine(Max(a, b, c), Q.positive(a) & Q.negative(c))

  File "C:\Users\Oscar\sympy\sympy\assumptions\refine.py", line 55, in refine
    ref_expr = expr._eval_refine(assumptions)

  File "C:\Users\Oscar\sympy\sympy\functions\elementary\miscellaneous.py", line 813, in _eval_refine
    if ask(i > j, assumptions):

  File "C:\Users\Oscar\sympy\sympy\assumptions\ask.py", line 480, in ask
    raise ValueError("inconsistent assumptions %s" % assumptions)

ValueError: inconsistent assumptions Q.negative(c) & Q.positive(a)
```

Not really clear why they are deemed inconsistent....

#### Release Notes

* assumptions
   * Basic support for `refine` of `Min` and `Max`.

### DIFF
--- a/sympy/core/relational.py
+++ b/sympy/core/relational.py
@@ -7,7 +7,7 @@ from .function import AppliedUndef
 from .singleton import S
 from .sympify import _sympify, SympifyError
 from .parameters import global_parameters
-from .logic import fuzzy_bool, fuzzy_xor, fuzzy_and, fuzzy_not
+from .logic import fuzzy_bool, fuzzy_xor, fuzzy_and, fuzzy_not, fuzzy_or
 from sympy.logic.boolalg import Boolean, BooleanAtom
 from sympy.utilities.exceptions import SymPyDeprecationWarning
 from sympy.utilities.iterables import sift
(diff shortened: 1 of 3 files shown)


oscargus:
Nah, seems to require more work to make sense, probably with more insights into how the new assumptions system works. Will keep it here to not forget about it.

(It works if > is used instead of >= in the `ask`, so somehow that should be fixed as well... That is, `ask(a >= c, Q.gt(a, c))` should return `True`, `ask(a > c, Q.gt(a, c))` does return `True`)

oscarbenjamin:
> I would have expected something else than an error here:
> refine(Max(a, b, c), Q.positive(a) & Q.negative(c))

I don't see an error for that:
```python
In [1]: a, b, c = symbols('a, b, c')

In [2]: refine(Max(a, b, c), Q.positive(a) & Q.negative(c))
Out[2]: Max(a, b, c)

In [3]: a, b, c = symbols('a, b, c', real=True)

In [4]: refine(Max(a, b, c), Q.positive(a) & Q.negative(c))
Out[4]: Max(a, b, c)
```

PR #22630, opened 2021-12-09
Helper function for solving differential equations by Laplace transform

### LINKED ISSUE #17098: Printing of Laplace transform of undefined function
See also #7219 which is about applying Laplace transform rules for derivatives and integrals of unknown functions.

Currently the Laplace transform of an unknown function prints like:
```julia
In [2]: t, s = symbols('t, s')

In [3]: y = Function('y')

In [4]: laplace_transform(y(t), t, s)
Out[4]: LaplaceTransform(y(t), t, s)
```
I wonder if the LaplaceTransform of an undefined function should print like `Y(s)` as in
```python
>>> LaplaceTransform(y(t), t, s)
Y(s)
```
There are other notations like `L{y}(s)` that could also be used.

Then after adding the properties in #7219 you could have something like
```python
>>> laplace_transform(Integral(y(t), t), t, s) 
Y(s)/s
```

hanspi42:

Fixes  #17098

New option in `lalpace_transforrm` that adds support for solving differential equations by Laplace transform

#### Release Notes

* integrals
  * Added support for solving differential equations by Laplace transform


oscarbenjamin:
How would we use this with a system of ODEs like:
```python
In [1]: s = Symbol('s')

In [2]: laplace_transform(f(t).diff(t) + g(t).diff(t), t, s)
Out[2]: s⋅LaplaceTransform(f(t), t, s) + s⋅LaplaceTransform(g(t), t, s) - f(0) - g(0)
```

sympy/integrals/tests/test_transforms.py
```diff
@@ -705,6 +707,14 @@ def test_laplace_transform():
         f(w), w, t, plane=0) == InverseLaplaceTransform(f(w), w, t, 0)
     assert LT(f(t)*g(t), t, s) == LaplaceTransform(f(t)*g(t), t, s)
 
+    # The following test confirms that issue #17098 is solved
+    F, _, _ = laplace_transform(diff(y(t), t, 2)
+                                + y(t)
+                                - 4*(Heaviside(t - 1)
+                                - Heaviside(t - 2)), t, s)
+    G = laplace_transform_replace_ic(F, y, Y, t, s, [1, 0])
```
oscarbenjamin, line 715: I wonder if it would be better to have a function `laplace_transform_ode` that combines these steps so that you only need to call one function. Maybe it should just be possible to provide these other arguments directly to `laplace_transform`. There might be situations where you want to calculate a `laplace_transform` in terms of an arbitrary function but it's not even about ODEs like:
```python
In [6]: laplace_transform(exp(-t)*f(t), t, s)
Out[6]: LaplaceTransform(f(t), t, s + 1)
```

hanspi42: Considering that there might be different undefined functions in an expression, could it be an idea to give the laplace_transform an argument `replacements=[]` that could be called with a list of tuples containing function pairs and initial conditions?  Example:

```
r = [(f, F, [1, 0]), (h, T, [0,0])]
laplace_transform(expr, t, s, replacements=r)
```

hanspi42:
> How would we use this with a system of ODEs like:
> 
> ```python
> In [1]: s = Symbol('s')
> 
> In [2]: laplace_transform(f(t).diff(t) + g(t).diff(t), t, s)
> Out[2]: s⋅LaplaceTransform(f(t), t, s) + s⋅LaplaceTransform(g(t), t, s) - f(0) - g(0)
> ```

Apply the replacement once for each function, this would work together with my suggestion above, https://github.com/sympy/sympy/pull/22630#discussion_r765880172

hanspi42:
> I wonder if it would be better to have a function `laplace_transform_ode` that combines these steps ...

Idea: `laplace_transform_ode` could take equation systems too.

`laplace_transform_ode(eq, t, s, funcs)` would take a list of equations for `eq`, variables for `t` and `s`, and funcs would be the list of tuples I explained above, e.g., `funcs = [(f, F, [1, 0]), (h, T, [0, 0])]`

I would keep `laplace_transform_replace_ic`, which would be called by `laplace_transform_ode`. What do you think?

hanspi42:
It took me some time to try and figure out a workable input strategy for the new ``laplace_transform_ode`` function.

In the present version,

```
from sympy.integrals.transforms import (laplace_transform_ode)
from sympy.core import diff
from sympy.core.function import Function
from sympy.core.symbol import symbols
x = Function('x')
X = Function('X')
y = Function('y')
Y = Function('Y')
s, t = symbols('s, t')
a, b, c, d, x0, y0 = symbols('a, b, c, d, x0, y0')
print(laplace_transform_ode([diff(x(t), t)-a*x(t)-b*y(t),
                             diff(y(t), t)-c*x(t)-d*y(t)],
                             t, s, 
                             ((x, X, [x0, 0, 0]),
                              (y, Y, [y0, 0, 0])) ))
```

results in 

```
{X(s): (b*y0 - d*x0 + s*x0)/(a*d - a*s - b*c - d*s + s**2), Y(s): (-a*y0 + c*x0 + s*y0)/(a*d - a*s - b*c - d*s + s**2)}
```

@oscarbenjamin: Is this the kind of function you were looking for?

sympy/integrals/transforms.py
```diff
@@ -1162,6 +1162,98 @@ def sbs(expr):
...
+    k=0
+    for c in ic:
+        if k == 0:
+            res = res.subs(f(0), c)
+        else:
+            if k == 1:
+                res = res.subs(Subs(Derivative(f(t), t), t, 0), c)
+            else:
+                res = res.subs(Subs(Derivative(f(t), (t, k)), t, 0), c)
+        k+=1
```
oscarbenjamin, line 1253: Can this not just be:
```
for k, c in enumerate(ic):
    res = res.subs(f(t).diff(t, k).subs(t, 0), c)
```

oscarbenjamin:
> In the present version,

Maybe I've misunderstood but why are the ICs for `x` given as `[x0, 0, 0]` when the ODE for `x` is only first order?

hanspi42:
> > In the present version,
> 
> Maybe I've misunderstood but why are the ICs for `x` given as `[x0, 0, 0]` when the ODE for `x` is only first order?

It does not matter how many ICs you give, it will just take those that are available, and if not enough are available, unreplaced initial conditions will be left standing.

sympy/integrals/transforms.py
hanspi42: > Can this not just be:
> 
> ```
> for k, c in enumerate(ic):
>     res = res.subs(f(t).diff(t, k).subs(t, 0), c)
> ```

A speechless Hanspi bows before the master ... thanks ;)

hanspi42:
Does a function like ``laplace_transform_ode`` require any checks whether the arguments are OK, or is it just a matter of writing good documentation?

sympy/integrals/transforms.py
```diff
@@ -1162,6 +1162,90 @@ def sbs(expr):
...
+    eqs_laplace = []
+    for eq in eqs:
+        xl = laplace_transform(eq, t, s)
+        for _f, _F, _ic in funcs:
+            xl = laplace_transform_replace_ic(xl, _f, _F, t, s, _ic)
+        eqs_laplace.append(xl)
+    Funcs = []
+    for _f, _F, _ic in funcs:
+        Funcs.append(_F(s))
+    sols_laplace = solve(eqs_laplace, Funcs)
```
oscarbenjamin, line 1180: I think it would be better to return the transform of the expressions unsolved. Then the user can call solve if they want to.

It would also be good to handle equations by transforming each side.

hanspi42: Done. It now also has a flag solve_it that defaults to `False`.

If you're happy, I'll write a doc and test cases.

hanspi42:
@oscarbenjamin I'm waiting for your "I'm happy" indicator ;)  Does it now look OK for you?

oscarbenjamin:
I guess I'm still unsure about the basic API design here.

It seems to me that the basic function of transforming `x(t)` to `X(s)` is probably something that users want even if they aren't doing anything with ODEs so it seems strange to tie it to functions called `_ode` or `_ics`. Also it probably makes more sense to apply these replacements while computing the transform in the first place.

I expect that just having these options in `laplace_transform` is probably what most users would want e.g. (more or less what you suggested above):
```python
In [12]: laplace_transform(x(t).diff(t, 2), t, s, funcs={x(t):X(s)}, ics={x(t):[2, 3]})
Out[12]:
 2
s ⋅X(s) - 2⋅s - 3
```
If we had that then would either of the functions added here be needed?

With those arguments to `laplace_transform` solving ODEs is just a case of calling `laplace_transform` then `solve` and then `inverse_laplace_transform` (which should presumably take the same `funcs={x(t):X(s)}` argument as well). That actually mirrors the steps needed conceptually when doing it by hand as well.

hanspi42:
> If we had that then would either of the functions added here be needed?

Well, I think not as public functions, but I would keep them as independent functions anyway to make the code easier to read.

> With those arguments to `laplace_transform` solving ODEs is just a case of calling `laplace_transform` then `solve` and then `inverse_laplace_transform` (which should presumably take the same `funcs={x(t):X(s)}` argument as well). That actually mirrors the steps needed conceptually when doing it by hand as well.

The `inverse_laplace_transform` should be a second step; most users in engineering would not even bother to do it as they would be more interested in the frequency content of the solution.

I'll try to do the API like this.

hanspi42:
Back to an old question: I want to implement a new named keyword in `laplace_transform`. There are many ways to do it, even inside the `sympy` code base, so could you please give me a reference to where something like this has ben done in a way that is considered best practice for sympy development?

hanspi42:
I have implemented it in the most straightforward way. `laplace_transform` can now take a tuple of replacements as well as a list of functions to transform. You see an example here:

https://github.com/sympy/sympy/blob/d17fe4ca6b14b76c6b6e5c9f4f266a7a2866954b/sympy/integrals/tests/test_transforms.py#L721-L726

I removed the `laplace_transform_ode` function.  I think it is quite OK to not provide a built-in solve as `laplace_transform` can now return an equation system. Depending on what is in that equation system, a different solver might be needed anyway.

What do you think?

sympy/integrals/transforms.py
```diff
@@ -1931,7 +1997,14 @@ def laplace_transform(f, t, s, legacy_matrix=True, **hints):
...
                 return type(f)(*f.shape, elements_trans)
 
-    return LaplaceTransform(f, t, s).doit(**hints)
+    if isinstance(f, list):
+        return [laplace_transform(x, t, s, diff_subs=diff_subs, **hints)
+                for x in f]
+
+    LT = LaplaceTransform(f, t, s).doit(**hints)
+    for _f, _r in diff_subs.items():
+        LT = laplace_transform_replace_ic(LT, _f, _r[0], t, s, _r[1])
```
oscarbenjamin, line 2006: Would it not be better to handle this directly in `_laplace_rule_diff`?

sympy/integrals/tests/test_transforms.py
```diff
@@ -705,6 +710,20 @@ def test_laplace_transform():
...
+    assert (G == s**2*Y(s) - s + Y(s) - 4*exp(-s)/s + 4*exp(-2*s)/s)
+
+    # laplace_transform covers lists and replacements
+    assert laplace_transform([diff(z(t), t, 2)-a*z(t)-b*y(t),
+                              diff(y(t), t)-c*z(t)-d*y(t)], t, s,
+                             {z: (Z, [z0, z1, 0]), y: (Y, [y0, 0, 0])}
```
oscarbenjamin, line 724: I think it's better to have the user provide `z(t)` and `Z(s)` rather than `z` and `Z`. Note that it's quite a common use pattern to define the python variable `z` as the expression `z(t)` i.e.:
```python
In [8]: t, s = symbols('t, s')

In [9]: z = Function('z')(t)

In [10]: Z = Function('Z')(s)

In [11]: z
Out[11]: z(t)

In [12]: Z
Out[12]: Z(s)
```

hanspi42:
I see your point.  I don't like it much, because then users need to provide redundant information to the API, but I have done it because it is close to common practice.

I have also added the option to give `0` instead of a list as initial condition, the all initial conditions are set to zero for that function.

Examples here:
https://github.com/hanspi42/sympy/blob/39b0394fbb07ece892f6571b2587979299aa2fc2/sympy/integrals/tests/test_transforms.py#L721-L729

sympy/integrals/transforms.py
hanspi42: I thought about this one, and the answer is no. The rules themselves should just do whatever is correct in an abstrtact manner, filling in additional information like actual names of transforms and initial conditions has got nothing to do with the rule itself.

sympy/integrals/tests/test_transforms.py
hanspi42: Done.

hanspi42:
I have updated the documentation; shall we remove ```laplace_transform_replace_ic```?

sympy/integrals/transforms.py
```diff
@@ -1885,14 +1966,25 @@ def laplace_transform(f, t, s, legacy_matrix=True, **hints):
...
     (24/s**5, 0, True)
     >>> laplace_transform(t**a, t, s)
     (gamma(a + 1)/(s*s**a), 0, re(a) > -1)
     >>> laplace_transform(DiracDelta(t)-a*exp(-a*t),t,s)
     (s/(a + s), Max(0, -a), True)
+    >>> laplace_transform(diff(z(t), t, 2), t, s)
+    s**2*LaplaceTransform(z(t), t, s) - s*z(0) - Subs(Derivative(z(t), t), t, 0)
+    >>> laplace_transform([diff(z(t), t, 2)-a*z(t)-b*y(t), diff(y(t), t)-c*z(t)-d*y(t)], t, s, {z(t): (Z(s), [z0, z1, 0]), y(t): (Y(s), [y0, 0, 0])})
+    [-a*Z(s) - b*Y(s) + s**2*Z(s) - s*z0 - z1, -c*Z(s) - d*Y(s) + s*Y(s) - y0]
+    >>> laplace_transform([diff(z(t), t, 2)-a*z(t)-b*y(t), diff(y(t), t)-c*z(t)-d*y(t)], t, s, {z(t): (Z(s), 0), y(t): (Y(s), 0)})
```
oscarbenjamin, line 1985: These lines are too long. It would be better to split this up e.g. by first defining the odes:
```
odes = [diff(z(t), t, 2)-a*z(t)-b*y(t), diff(y(t), t)-c*z(t)-d*y(t)]
```

Also I think that it would be better to visually separate these from the simpler examples of using `laplace_transform` shown above by breaking it up like:
```
>>> test1
out1

It is also to compute Laplace transforms of functions that involve undefined functions and their derivatives which can be used for working with ordinary differential equations.

>>> more test
more out
```

sympy/integrals/transforms.py
```diff
@@ -1870,6 +1938,19 @@ def laplace_transform(f, t, s, legacy_matrix=True, **hints):
     If the integral cannot be fully computed in closed form, this function
     returns an unevaluated :class:`LaplaceTransform` object.
 
+    This function can also transform differentials; in that case the result
+    will contain unevaluated expressions like ``LaplaceTransform(x(t), t, s)``
+    and initial conditions like ``x(0)`` or ``Subs(Derivative(x(t), t), t, 0)``.
```
oscarbenjamin, line 1943: It isn't really "unevaluated expressions". If the function `x(t)` is unknown then what more can we do but refer to `L{x(t)}`. That's also how it displays in LaTeX form:
```python
In [112]: print(latex(laplace_transform(f(t), t, z)))
\mathcal{L}_{t}\left[f{\left(t \right)}\right]\left(z\right)
```
Id you use this in a jupyter notebook then that's the output you'll see (the pretty printer could be changed to shbow the same in a terminal.

Also rather than `Subs(Derivative(x(t), t), t, 0)` I'd just say `x'(0)`. In pretty or latex printing it shows up better e.g.:
```
In [114]: laplace_transform(f(t).diff(t, 2), t, z)
Out[114]: 
 2                                         ⎛d       ⎞│   
z ⋅LaplaceTransform(f(t), t, z) - z⋅f(0) - ⎜──(f(t))⎟│   
                                           ⎝dt      ⎠│t=0
```
(It's really just `LaplaceTransform` that doesn't display very well here)

sympy/integrals/transforms.py
hanspi42: Good idea, thanks!

sympy/integrals/transforms.py
hanspi42: Next version is clearer ;)

oscarbenjamin:
If test sphinx fails then there is an error building the docs. You can test that by building the docs yourself. Most likely there is a problem with the rst used in a docstring.

hanspi42:
> If test sphinx fails then there is an error building the docs. You can test that by building the docs yourself. Most likely there is a problem with the rst used in a docstring.

Caught it, thanks. It was a typo in the "see also" section of ```laplace_tarnsfom```'s docstring.

hanspi42:
@oscarbenjamin: This is ready, in my opinion. What do you think?

sympy/integrals/transforms.py
```diff
@@ -1885,15 +1899,34 @@ def laplace_transform(f, t, s, legacy_matrix=True, **hints):
...
     ========
 
-    >>> from sympy import DiracDelta, exp, laplace_transform
-    >>> from sympy.abc import t, s, a
+    >>> from sympy import DiracDelta, exp, laplace_transform, diff, Function, symbols
+    >>> y = Function('y')
+    >>> Y = Function('Y')
+    >>> z = Function("z")
+    >>> Z = Function("Z")
+    >>> s, t, a, b, c, d, z0, z1, y0 = symbols('s, t, a, b, c, d, z0, z1, y0')
```
oscarbenjamin, line 1907: Most of these aren't needed until lower down. I suggest to define them where they are needed so that we can keep the first examples simple.

Also multiple functions can be created like:
```python
y, Y, z, Z = symbols('y, Y, z, Z', cls=Function)
```

sympy/integrals/transforms.py
```diff
@@ -1931,7 +1964,28 @@ def laplace_transform(f, t, s, legacy_matrix=True, **hints):
...
+            n = Wild('n')
+            x = x.replace(Subs(Derivative(_f, (t, n)), t, 0), 0)
+            x = x.subs(_f.subs(t, 0), 0)
+        else:
+            for k, c in enumerate(_r[1]):
+                x = x.subs(_f.diff(t, k).subs(t, 0), c)
+    if type(LT) is tuple:
+        return x, a, c
+    else:
+        return x
```
oscarbenjamin, line 1988: I think it would be better to factor the for loop out into a separate function (which can be defined locally inside this function) so that the logic is just like:
```python
if isinstance(LT, tuple):
    x, a, c = LT
    x = func(x)
    return x, a, c
else:
    return func(x)

```

sympy/integrals/transforms.py
```diff
@@ -1836,7 +1836,7 @@ def _try_directly(self, **hints):
         return fn, LT
 
 
-def laplace_transform(f, t, s, legacy_matrix=True, **hints):
+def laplace_transform(f, t, s, diff_subs=dict(), legacy_matrix=True, **hints):
```
oscarbenjamin, line 1839: I'm not sure `diff_subs` is such a good name for this parameter. Maybe `function_map` or something is better? Differentiation need not be involved:
```python
In [1]: t, s = symbols('t, s')

In [2]: x, X = symbols('x, X', cls=Function)

In [4]: laplace_transform(x(t), t, s, {x(t):(X(s),[])})
Out[4]: X(s)
```
Also I think this parameter should be keyword-only rather than positional.

oscarbenjamin:
Testing this out I find some examples that don't work:
```python
In [7]: t, s = symbols('t, s')

In [8]: x, X = symbols('x, X', cls=Function)

In [9]: laplace_transform(x(t)*exp(-t), t, s, {x(t):(X(s),[])})
Out[9]: LaplaceTransform(x(t), t, s + 1)
```
Should that not be `X(s + 1)`?

Is it possible to give better outputs for the examples below?
```python
In [10]: laplace_transform(x(t)*Heaviside(t-2), t, s, {x(t):(X(s),[])})
Out[10]: LaplaceTransform(x(t)⋅θ(t - 2), t, s)

In [11]: laplace_transform(x(t-1)*Heaviside(t), t, s, {x(t):(X(s),[])})
Out[11]: LaplaceTransform(x(t - 1)⋅θ(t), t, s)

In [14]: laplace_transform(t*x(t), t, s, {x(t):(X(s),[])})
Out[14]: LaplaceTransform(t⋅x(t), t, s)
```

sympy/integrals/transforms.py
hanspi42: Shouldm't legacy_matrix then also be a keyword-only argument?

hanspi42:
> In [9]: laplace_transform(x(t)*exp(-t), t, s, {x(t):(X(s),[])})
> Out[9]: LaplaceTransform(x(t), t, s + 1)
> ```
> 
> Should that not be `X(s + 1)`?

Yes, it should.  This is an easy fix.

> Is it possible to give better outputs for the examples below?

Not in general, a multiplication in the time domain would give a convolution integral in the s domain.

sympy/integrals/transforms.py
hanspi42: Thanks! I did not know this yet.

sympy/integrals/transforms.py
hanspi42: Will be done!

sympy/integrals/transforms.py
oscarbenjamin: Yes, but we should really just get rid of legacy_matrix.

oscarbenjamin:
The table here:
https://tutorial.math.lamar.edu/classes/de/laplace_table.aspx

Lists the example that `L(t*f(t)) = -F'(s)`. Should that not be computed by `laplace_transform`?

oscarbenjamin:
How do you envisage this working with `inverse_laplace_transform`?

Presumably if we have the way to say `x(t) -> X(s)` for the forward transform then we will want to be able to do the same for the inverse transform as well. How would it look to do the transform, make some rearrangements and then do the inverse transform?

oscarbenjamin:
Just testing some examples:
```python
In [18]: s, t = symbols('s, t')

In [19]: x, y = symbols('x, y', cls=Function)

In [20]: X, Y = symbols('X, Y', cls=Function)

In [21]: lt = lambda e: laplace_transform(e, t, s, func_map={x(t): (X(s),[]), y(t):(Y(s),[])})

In [22]: lt(Heaviside(t-1)*x(t-1)) # seems to work
Out[22]: 
      -s
X(s)⋅ℯ  

In [23]: lt(Heaviside(t)*x(t)) # doesn't work...
Out[23]: LaplaceTransform(x(t)⋅θ(t), t, s)

In [24]: lt(Heaviside(t-z)*x(t-z)) # fails for a symbol
TypeError 
```

> how good it is to base Open-Source software on a not openly available table.

It is better to have open sources if possible. The rule about `t**n*f(t)` seems to appear in most of the tables I've seen along with `f(t)/t` so it seems like a reasonable candidate.

It doesn't seem immediately possible to find an electronic copy of Oberhettinger's book although parts of it can be seen in google books:
https://books.google.co.uk/books?id=3OHrCAAAQBAJ&printsec=frontcover&source=gbs_ViewAPI&redir_esc=y#v=onepage&q&f=false

I notice that rule 1.4 doesn't seem to be implemented:
```python
In [35]: lt(x(t+1))
Out[35]: LaplaceTransform(x(t + 1), t, s)

In [36]: lt(x(t-1))
Out[36]: LaplaceTransform(x(t - 1), t, s)
```
I think that these general rules are probably the most important because can we not also use them with explicit functions as well e.g. here the combination of two rules can be applied to reduce this to finding the LT of `sin`:
```python
In [42]: lt(sin(t-1)/(t-1))
Out[42]: 
                ⎛sin(t - 1)      ⎞
LaplaceTransform⎜──────────, t, s⎟
                ⎝  t - 1         ⎠
```

hanspi42:
> How do you envisage this working with `inverse_laplace_transform`?
> 
> Presumably if we have the way to say `x(t) -> X(s)` for the forward transform then we will want to be able to do the same for the inverse transform as well. How would it look to do the transform, make some rearrangements and then do the inverse transform?

Here the main issue is that most people would get rational expressions in `s`, so I'd definitely first convert the `inverse_laplace_transform` to rule-based. There I am a bit stuck with the concept because even the formal structure of the inverse LT of  `1/(s**2+a*s+1)` depends on the value of `a` (harmonic function or sum of exponentials). 

`inverse_laplace_transform` can then also have a func_map argument.

oscarbenjamin:
> There I am a bit stuck with the concept because even the formal structure of the inverse LT of `1/(s**2+a*s+1)` depends on the value of `a` (harmonic function or sum of exponentials).

Generally speaking sympy will treat the "generic case" in this kind of situation i.e. for "almost all" possible values of `a` there will not be a repeated root. Hence:
```python
In [17]: eq = x(t).diff(t, 2) + a*x(t).diff(t) + x(t)

In [18]: eq
Out[18]: 
                      2      
  d                  d       
a⋅──(x(t)) + x(t) + ───(x(t))
  dt                  2      
                    dt       

In [19]: dsolve(eq)
Out[19]: 
             ⎛        ________⎞          ⎛       ________⎞ 
             ⎜       ╱  2     ⎟          ⎜      ╱  2     ⎟ 
           t⋅⎝-a + ╲╱  a  - 4 ⎠       -t⋅⎝a + ╲╱  a  - 4 ⎠ 
           ────────────────────       ─────────────────────
                    2                           2          
x(t) = C₁⋅ℯ                     + C₂⋅ℯ    
```
Also (what if `a=0`?):
```python
In [23]: roots(a*y**2 + b*y + c, y)
Out[23]: 
⎧           _____________                _____________   ⎫
⎪          ╱           2                ╱           2    ⎪
⎨   b    ╲╱  -4⋅a⋅c + b          b    ╲╱  -4⋅a⋅c + b     ⎬
⎪- ─── - ────────────────: 1, - ─── + ────────────────: 1⎪
⎩  2⋅a         2⋅a              2⋅a         2⋅a          ⎭

In [26]: solve(a*y - b, y)
Out[26]: 
⎡b⎤
⎢─⎥
⎣a⎦
```
Ideally sympy would return the symbolic conditions for the validity of things like this but the APIs haven't been designed that way. It is often possible to return Piecewise but then it isn't always practical to compute all possible cases.

oscarbenjamin:
> `inverse_laplace_transform` can then also have a func_map argument.

So how would this look? I'm mainly interested because I'm not sure about the form of arguments given to `laplace_transform` in this PR currently. I would naturally expect something like this:
```python
In [1]: x, X = symbols('x, X', cls=Function)

In [2]: t, s = symbols('t, s')

In [3]: ics = {x(t):[0,0]}

In [4]: func_map = {x(t): X(s)}

In [5]: laplace_transform(x(t).diff(t, 2), t, s, func_map=func_map, ics=ics)
Out[5]: 
 2     
s ⋅X(s)

In [7]: inverse_laplace_transform(_, s, t, func_map=func_map)
Out[7]: 
  2      
 d       
───(x(t))
  2      
dt   
```
The point here is that the same `func_map` dict can be used for both the forward and backward transform. For the backward transform the initial conditions don't make sense. What would be the analogous data to pass there?

To me it would seem more natural to separate the initial conditions from the function map.

hanspi42:
This PR is completely out of date. A similar project may be re-started after #24561 has progressed far enough.

Issue #22616, opened 2021-12-06
From mathml to python code


rspadim:
Hello Folks, could be possible convert mathml to python code?

for example a sum be converted to a for loop a summer var, and call to sum expression? And others more complex expressions and systems?

oscarbenjamin:
As far as I can tell this isn't currently possible

ThePauliPrinciple:
You could try converting it to LaTeX first: https://tex.stackexchange.com/questions/85639/i-want-to-convert-mathml-to-latex
and then use the SymPy LaTeX parser https://docs.sympy.org/latest/modules/parsing.html#experimental-mathrm-latex-parsing and then using the python codeprinter https://docs.sympy.org/latest/modules/printing.html#sympy.printing.pycode.pycode. But I would be surprised if it is plug and play

PR #22637, opened 2021-12-10
Fix for #22616 : added code that can be used for converting mathml code to python code.

gazal2708:
…python

added code that can be used for converting mathml code to python code.

Took help from the repository [(https://github.com/oerpub/mathconverter/tree/master/xsl_yarosh)] for converting first mathml to latex code, then used sympy libraries to convert latex code to python code.


oscargus:
Thanks for this. I believe that some things are not really clear.
1. What is the copyright of the added files? Probably the readme should stay in the directory? In case it is OK to use them like this to start with.
2. Now, you convert to Python, although the function etc is called tolatex. Probably a SymPy representation is the preferred output.
3. requirements.txt is for distribution purposes, not to add dependencies to SymPy itself (which is really restricted, only one currently)
4. Is this for content or presentation MathML?
5. See https://github.com/sympy/sympy/tree/master/sympy/utilities/mathml how a content to presentation converter is packages. This can serve as a guideline how to deal with some of the points about, like how to check if lxml is installed etc. Also, it may possibly be used for converting from content to presentation (if you support presentation).

gazal2708:
Apologies for the confusion. @oscargus 
1. I will add the readMe in the folder. For adding those .xsl files i took reference from github repository : [https://github.com/oerpub/mathconverter/tree/master/xsl_yarosh].
2. Yes the function name I need to change to mathMLtoPython. Thanks for pointing it out.
3. The code i added uses a dependency "lxml", which i thought to add in requirements.txt. Without that dependency the code cannot parse mathML code or .xsl files i added.
4. I was not aware of the difference. I just wrote the code which can transform mathematical equations written in mathML into Python code.
5. [https://github.com/sympy/sympy/blob/master/sympy/utilities/mathml/data/mmltex.xsl] , thanks for sharing the link. The .xsl files I have added are similar to the code inside this file.

gazal2708:
> Thanks for this. I believe that some things are not really clear.
> 
> 1. What is the copyright of the added files? Probably the readme should stay in the directory? In case it is OK to use them like this to start with.
> 2. Now, you convert to Python, although the function etc is called tolatex. Probably a SymPy representation is the preferred output.
> 3. requirements.txt is for distribution purposes, not to add dependencies to SymPy itself (which is really restricted, only one currently)
> 4. Is this for content or presentation MathML?
> 5. See https://github.com/sympy/sympy/tree/master/sympy/utilities/mathml how a content to presentation converter is packages. This can serve as a guideline how to deal with some of the points about, like how to check if lxml is installed etc. Also, it may possibly be used for converting from content to presentation (if you support presentation).

After using this link you provided: https://github.com/sympy/sympy/tree/master/sympy/utilities/mathml , I have removed the extra files I added. the "mmltex.xsl" file present inside "data" folder is same as those files which i added.

gazal2708:
> Thanks for this. I believe that some things are not really clear.
> 
> 1. What is the copyright of the added files? Probably the readme should stay in the directory? In case it is OK to use them like this to start with.
> 2. Now, you convert to Python, although the function etc is called tolatex. Probably a SymPy representation is the preferred output.
> 3. requirements.txt is for distribution purposes, not to add dependencies to SymPy itself (which is really restricted, only one currently)
> 4. Is this for content or presentation MathML?
> 5. See https://github.com/sympy/sympy/tree/master/sympy/utilities/mathml how a content to presentation converter is packages. This can serve as a guideline how to deal with some of the points about, like how to check if lxml is installed etc. Also, it may possibly be used for converting from content to presentation (if you support presentation).

@oscargus  Regarding 4th point, I wanted to know, the input type to mathml_to_python function will be content MathML or presentation mathML? Currently, the function supports presentation mathML inputs but based on requirements, I can do further modifications

sylee957:
We had updated the code with `lxml` in #25007 because of the security issues in the parsers.
If the project here continues, this should also be taken account.

PR #22638, opened 2021-12-10
core/basic: __eq__ ignores iterables like Expr.__eq__

ThePauliPrinciple:

follow-up from #22593
See also https://github.com/sympy/sympy/pull/22607

Recently `Expr.__eq__` was updated to not convert iterables with
`_sympify`. This is now also added to `Basic.__eq__`.

`FiniteSet` and `Dict` now have explicit checks.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -13,7 +13,7 @@ from ._print_helpers import Printable
 
 from sympy.utilities.decorator import deprecated
 from sympy.utilities.exceptions import SymPyDeprecationWarning
-from sympy.utilities.iterables import iterable, numbered_symbols
+from sympy.utilities.iterables import iterable, numbered_symbols, NotIterable
 from sympy.utilities.misc import filldedent, func_name
 
 from inspect import getmro
@@ -342,9 +342,15 @@ class Basic(Printable, metaclass=ManagedProperties):
             return True
 
         if not isinstance(other, Basic):
+            if iterable(other, exclude=(str, NotIterable)
+                    ) and not hasattr(other, '_sympy_'):
+                # XXX iterable self should have it's own __eq__
+                # method if the path gives a false negative
+                # comparison
+                return False
             try:
                 other = _sympify(other)
-            except SympifyError:
+            except (SympifyError, SyntaxError):
                 return NotImplemented
 
         if type(self) != type(other):
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -328,10 +328,10 @@ class Basic(Printable, metaclass=ManagedProperties):
 
         If a class that overrides __eq__() needs to retain the
         implementation of __hash__() from a parent class, the
-        interpreter must be told this explicitly by setting __hash__ =
-        <ParentClass>.__hash__. Otherwise the inheritance of __hash__()
-        will be blocked, just as if __hash__ had been explicitly set to
-        None.
+        interpreter must be told this explicitly by setting 
+        __hash__ : Callable[[object], int] = <ParentClass>.__hash__. 
+        Otherwise the inheritance of __hash__() will be blocked, 
+        just as if __hash__ had been explicitly set to None.
 
         References
         ==========
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -328,9 +328,9 @@ class Basic(Printable, metaclass=ManagedProperties):
 
         If a class that overrides __eq__() needs to retain the
         implementation of __hash__() from a parent class, the
-        interpreter must be told this explicitly by setting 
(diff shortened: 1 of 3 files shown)


sympy/core/basic.py
```diff
@@ -342,9 +342,15 @@ def __eq__(self, other):
...
+                    ) and not hasattr(other, '_sympy_'):
+                # XXX iterable self should have it's own __eq__
+                # method if the path gives a false negative
+                # comparison
+                return False
             try:
                 other = _sympify(other)
-            except SympifyError:
+            except (SympifyError, SyntaxError):
                 return NotImplemented
```
oscarbenjamin, line 354: Why is this catching SyntaxError? There can't be a good reason to catch that here.

sympy/core/containers.py
```diff
@@ -304,6 +304,13 @@ def __lt__(self, other):
     def _sorted_args(self):
         return tuple(sorted(self.args, key=default_sort_key))
 
+    def __eq__(self, other):
+        if isinstance(other, dict):
+            return self == Dict(other)
+        return super().__eq__(other)
```
oscarbenjamin, line 310: Would it not be better to do `return self._dict == other`?

sympy/core/basic.py
ThePauliPrinciple: I copied this over from `Expr.__eq__`, should it be removed in both places?

sympy/core/containers.py
ThePauliPrinciple: That would currently fail, since `Dict` uses `sympify` (in `Tuple`), it would try to match e.g. `Symbol('x')` with `'x'`:
`Dict({'x':1}) == {'x':1}` would return `False`.

This requires a larger `Dict` overhaul that I can hopefully solve in https://github.com/sympy/sympy/pull/22580

sympy/core/containers.py
```diff
@@ -304,6 +304,13 @@ def __lt__(self, other):
...
+
+    def __hash__(self):
+        return Basic.__hash__(self)
```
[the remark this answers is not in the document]
ThePauliPrinciple: mypy complains about this, I tried a bunch of things but couldn't get it to work. (Don't really have any experience with mypy)

sympy/core/containers.py
oscarbenjamin: Tuple should be fixed to use `_sympify`.

sympy/core/containers.py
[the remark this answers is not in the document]
oscarbenjamin: You can do something like
```python
from typing import Callable, Any

    __hash__: Callable[[Basic], Any]  = Basic.__hash__
```

sympy/core/containers.py
ThePauliPrinciple: I agree, but it requires a large cleanup of parts in the code that rely on `sympify` being used in `Tuple`

sympy/core/basic.py
ThePauliPrinciple: This is also in `numbers.py`. Possibly it is to catch errors with `str` parsed expressions in classes that (unfortunately) use `sympify` over `_sympify`

ThePauliPrinciple:
```python
from sympy import *
Tuple("2x")
```
Gives a `SyntaxError`

oscarbenjamin: How exactly does calling `_sympify` lead to the `SyntaxError`?
```python
In [65]: from sympy.core.sympify import _sympify

In [66]: _sympify(('2x', 'x'))
SympifyError 
```

oscarbenjamin: Sorry for the noise, I had already changed `Tuple.__new__` locally to use `_sympify`. With actual master I see the `SyntaxError`:
```python
In [1]: from sympy.core.sympify import _sympify

In [2]: _sympify(('2x', 'x'))
ValueError                                Traceback (most recent call last)
ValueError: Error from parse_expr with transformed code: "Integer (2 )Symbol ('x' )"

The above exception was the direct cause of the following exception:

SyntaxError  
```

ThePauliPrinciple: In case you're looking into solving all errors related to changing `sympify` to `_sympify` in `Tuple`, I do have a WIP branch to solve the issues related to `Qubit` (it's using `Tuple` to convert "0" and "1" strings to and I replaced it with an explicit conversion)

ThePauliPrinciple: Maybe this PR is helpful regarding this topic: https://github.com/sympy/sympy/pull/22643

oscarbenjamin:
Looks good

PR #22641, opened 2021-12-10
calculus tutorial updated

AdvaitPote:
In the calculus tutorial documentation, specifically under 'Series Expansion', it is written 'x=\infty'. Mathematically, however, that is incorrect as no variable or function is ever equal to infinity but tends to. Hence, the '=' has been replaced by a 'tends to' sign.

#### Release Notes

NO ENTRY


anutosh491:
I guess you should also do a basic check for such similar cases and report if any !

AdvaitPote:
@anutosh491 After checking, the only mathematical inconsistency I could find in the calculus tutorial documentation is the one I mentioned above. So I believe that is the only change that should be made.

smichr:
Thank you.

PR #22642, opened 2021-12-10
lower order hint turns off higher order flags that are unspecified

smichr:

`solve(..., cubics=False, quartics=False, quintics=False)` now is the same as `solve(..., cubics=False)`
#### Release Notes

* solvers
    * disabling lower-order solvers automatically disables higher order solvers: cubics=False automatically sets quartic and quintic to False, too


oscarbenjamin:
Looks good

PR #22646, opened 2021-12-10
polys/numberfields: Support conversions `ANP` --> `AlgebraicNumber`

skieffer:

* Give the `AlgebraicField` class a `to_alg_num(a)` method, which converts an `ANP` to an `AlgebraicNumber`.

* This is supported by a new method, `AlgebraicNumber.field_element()`, which is useful on its own.

* Make `AlgebraicField.integral_basis()` accept a `fmt` arg, so user can define the
  desired format for the returned basis elements.

#### Release Notes

* polys
  * Add support for conversions `ANP` --> `AlgebraicNumber`.

### DIFF
--- a/sympy/polys/numberfields/tests/test_basis.py
+++ b/sympy/polys/numberfields/tests/test_basis.py
@@ -1,6 +1,9 @@
-from sympy import QQ
 from sympy.abc import theta, x
+from sympy.core import S
+from sympy.core.numbers import AlgebraicNumber
+from sympy.functions.elementary.miscellaneous import sqrt
 from sympy.polys import Poly, cyclotomic_poly
+from sympy.polys.domains import QQ
 from sympy.polys.matrices import DomainMatrix, DM
 from sympy.polys.numberfields.basis import round_two
 from sympy.testing.pytest import raises
@@ -65,3 +68,15 @@ def test_round_two():
         # The computed basis need not equal the expected one, but their quotient
         # must be unimodular:
         assert (B.inv()*B_exp).det()**2 == 1
+
+
+def test_AlgebraicField_integral_basis():
+    alpha = AlgebraicNumber(sqrt(5), alias='alpha')
+    k = QQ.algebraic_field(alpha)
+    B0 = k.integral_basis()
+    B1 = k.integral_basis(fmt='sympy')
+    B2 = k.integral_basis(fmt='alg')
+    assert B0 == [k([1]), k([S.Half, S.Half])]
+    assert B1 == [1, S.Half + alpha/2]
+    assert B2 == [alpha.field_element([1]),
+                  alpha.field_element([S.Half, S.Half])]
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -2835,6 +2835,50 @@ class AlgebraicNumber(Expr):
                     return AlgebraicNumber(r)
         return self
 
+    def field_element(self, coeffs):
+        r"""
+        Form another element of the same number field.
+
+        Explanation
+        ===========
+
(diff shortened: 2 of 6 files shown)


sympy/polys/numberfields/tests/test_basis.py
```diff
@@ -65,3 +68,15 @@ def test_round_two():
...
         assert (B.inv()*B_exp).det()**2 == 1
+
+
+def test_AlgebraicField_integral_basis():
+    alpha = AlgebraicNumber(sqrt(5), alias='alpha')
+    k = QQ.algebraic_field(alpha)
+    B0 = k.integral_basis()
+    B1 = k.integral_basis(fmt='sympy')
+    B2 = k.integral_basis(fmt='alg')
+    assert B0 == [k([1]), k([S.Half, S.Half])]
```
oscarbenjamin, line 79: I didn't realise that a list could be passed like this. I've just noticed a problem with this that the types of the elements in the list are not checked or converted:
```python
In [16]: K = QQ.algebraic_field(sqrt(2))

In [17]: K(1)
Out[17]: ANP([mpq(1,1)], [mpq(1,1), mpq(0,1), mpq(-2,1)], QQ)

In [18]: K([1])
Out[18]: ANP([1], [mpq(1,1), mpq(0,1), mpq(-2,1)], QQ)

In [19]: K([sqrt(2)])
Out[19]: ANP([sqrt(2)], [mpq(1,1), mpq(0,1), mpq(-2,1)], QQ)
```
Perhaps `of_type` should be checked and an error raised:
```python
In [22]: K.domain.of_type(QQ(1))
Out[22]: True

In [23]: K.domain.of_type(1)
Out[23]: False
```

oscarbenjamin:
Looks good. Thanks!

sympy/polys/numberfields/tests/test_basis.py
skieffer: Oh, that slipped right by me! Well, what we currently have in the `ANP` constructor is kind of odd:

https://github.com/sympy/sympy/blob/81cd2630631e315586f796646a7f9268e71bf484/sympy/polys/polyclasses.py#L1515-L1518

When the incoming `rep` is _not_ a list, it converts for you (which is why the `K(1)` example above works out okay).

So when `rep` _is_ a list, what do you want to do? Follow suit and convert, or do type checking and raise an exception?

oscarbenjamin: I guess it can just convert.

I don't like implicit conversions in general because they slow things down. In polys there are often things that are slow because of this. With implicit conversions you don't notice unless you profile things. With a TypeError you see straight away where you need to do a convert.

Also with explicit conversions you can specify what to convert from e.g. `K1.convert(x, K2)` is often significantly faster than `K1.convert(x)` even if both work. You can only use the explicit `K2` if you are making an explicit conversion though.

skieffer: Well, I would tend to be in favor of converting, from a convenience standpoint; on the other hand, I do see your point about implicit conversions.

Digging some more, I discovered that `ANP` already has a `from_list()` method:

https://github.com/sympy/sympy/blob/81cd2630631e315586f796646a7f9268e71bf484/sympy/polys/polyclasses.py#L1600-L1604

which, arguably, is the "right" way to build an `ANP` from a list of arbitrary types.

I could see giving `AlgebraicField` its own `from_list()` method which uses this, so that the example from the unit test above becomes

```python
k.from_list([S.Half, S.Half])]
```

oscarbenjamin: Yes, I think that a `from_list` method is better. Probably `k([1])` should convert just like `k(1)` does but `k([sqrt(2)])` should be an error in any case.

PR #18164, opened 2019-12-29
Solves a Bug in Derivative

Saanidhyavats:
* core
  * Replaced Derivative with cls which was related to issue #16350

PR #22647, opened 2021-12-11
Solves issue of ignoring evaluate=False condition

### LINKED ISSUE #16350: Derivative ignores `evaluate=False` for nested derivative
Hello

There is a small issue in `Derivative` that is it is still evaluated n matter `evaluate` for nested derivative dues to these lines (1340-1343 in `cord/functions.py`:

```
        # denest
        if isinstance(expr, Derivative):
            variable_count = list(expr.variable_count) + variable_count
            expr = expr.expr
            return Derivative(expr, *variable_count, **kwargs)
```

This part is highly problematic for inheritance for two reasons:

expression is evaluated AND even if it were decided that this is necessary, the last line should be
` return self.func(expr, *variable_count, **kwargs)` so that inheritance doesn't break.

cheers

faze-geek:

Closes #16350 

Being new to the open source community ,I was looking into the easy to fix issues of sympy when I found this issue .Earlier @Saanidhyavats had fixed issue #18164 by changing the if statement and return statement .I have read and built on this code ,made changes and also added comment explaining what is being fixed by this issue so that someone else can get clarity while reading the return statement.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -1350,7 +1350,7 @@ class Derivative(Expr):
         if len(variable_count) == 0:
             return expr
 
-        evaluate = kwargs.get('evaluate', False)
+        evaluate = kwargs.get('evaluate', None)
 
         if evaluate:
             if isinstance(expr, Derivative):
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
Looks like the printing tests here need to be updated:
https://github.com/sympy/sympy/blob/2285e0bfb65d85b4b62c79526295b4014ac37c2f/sympy/printing/tests/test_latex.py#L743-L745

Probably there should be printing tests for both collapsed and uncollapsed derivatives. A collapsed Derivative that still does not evaluate the differentiation can be made with:
```python
In [3]: Derivative(x**2, (x, 2))
Out[3]: 
  2    
 d ⎛ 2⎞
───⎝x ⎠
  2    
dx 
```

PR #22649, opened 2021-12-11
Fixed 'simplify' changing state of internal random generator

### LINKED ISSUE #22592: `simplify` changes state of internal random generator
I'm afraid [`sympy.core.ramdom`](https://github.com/sympy/sympy/blob/4489b4c7c51a65966b1a95f2af6f7e0c8c461794/sympy/core/random.py#L1-L10) brings back an old issue with `random` in `sympy`. 

See
```
>>> from sympy import I, simplify
>>> from sympy.core.random import rng
>>> rng.getstate()[1][-1]
624
>>> simplify(-I)
-I
>>> rng.getstate()[1][-1]
497
```
actually this comes from an lazy import inside of `simplify`

```
>>> from sympy.core.random import rng
>>> rng.getstate()[1][-1]
624
>>> from sympy.physics.units.quantities import Quantity
>>> rng.getstate()[1][-1]
489
```
The actual problem seems to be deep in the units system when it comes to multiply when it comes to multiply two units like `mass * time`

https://github.com/sympy/sympy/blob/52f606a503cea5e9588de14150ccb9f7f9ed4752/sympy/physics/units/systems/length_weight_time.py#L102

mohajain:
Fixes #22592

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/assumptions.py
+++ b/sympy/core/assumptions.py
@@ -214,7 +214,7 @@ from .facts import FactRules, FactKB, InconsistentAssumptions
 from .core import BasicMeta
 from .sympify import sympify
 
-from sympy.core.random import shuffle
+from sympy.core.random import rng_assumptions
 
 
 _assume_rules = FactRules([
@@ -548,7 +548,7 @@ def _ask(fact, obj):
 
     # Try assumption's prerequisites
     prereq = list(_assume_rules.prereq[fact])
-    shuffle(prereq)
+    rng_assumptions.shuffle(prereq)
     for pk in prereq:
         if pk in assumptions:
             continue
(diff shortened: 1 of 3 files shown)


mohajain:
I applied the diff you suggested. Please have a look @oscarbenjamin

oscarbenjamin:
There should be a test for this.

Need to test that something like this outputs a consistent result:
```python
from sympy import *
from sympy.core.random import seed, randint

seed(4)
prod(symbols('x:10')) # invoke the assumptions system
print(randint(0, 5))
```
That will need to be run in a subprocess though to avoid interactions with the cache.

mohajain:
Okay. How can we test something like this? What would you like me to do?

mohajain:
@oscarbenjamin could you please help?

oscarbenjamin:
We can just have a test along these lines using subprocess:
https://docs.python.org/3/library/subprocess.html

We should make sure that the test is something that currently fails but that doesn't fail with this PR.

sylee957:
This is already fixed in #22592

PR #22651, opened 2021-12-11
sympy/core: Basic and Expr __eq__ are merged

ThePauliPrinciple:

Helps with https://github.com/sympy/sympy/pull/22607

Currently, `Expr` ensures that `Expr(S(1)) != Expr(S(1.0))`. However, this does not work for
`Basic` : `Basic(S(1)) == Basic(S(1.0))`. Since this is the only difference currently between
`Basic` and `Expr`'s `__eq__` method, this PR attempts to merge these two so that
`__eq__` only has to be defined in `Basic`.

This causes backwards compatibility issues, but I believe `Basic(S(1)) == Basic(S(1.0))`
should be considered a bug.

#### Release Notes

* core
  * `Basic` now behaves the same as `Expr` when comparing structural equality with `==`.

### DIFF
--- a/sympy/calculus/tests/test_singularities.py
+++ b/sympy/calculus/tests/test_singularities.py
@@ -67,6 +67,7 @@ def test_is_decreasing():
     b = Symbol('b', positive=True)
 
     assert is_decreasing(1/(x**2 - 3*x), Interval.open(Rational(3,2), 3))
+    assert is_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3))
     assert is_decreasing(1/(x**2 - 3*x), Interval.Lopen(3, oo))
     assert not is_decreasing(1/(x**2 - 3*x), Interval.Ropen(-oo, Rational(3, 2)))
     assert not is_decreasing(-x**2, Interval(-oo, 0))
--- a/sympy/calculus/tests/test_singularities.py
+++ b/sympy/calculus/tests/test_singularities.py
@@ -81,13 +81,12 @@ def test_is_strictly_decreasing():
     assert not is_strictly_decreasing(-x**2, Interval(-oo, 0))
     assert not is_strictly_decreasing(1)
     assert is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(Rational(3,2), 3))
-    print(is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3)))
     assert is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3))
-
 
 def test_is_monotonic():
     """Test whether is_monotonic returns correct value."""
     assert is_monotonic(1/(x**2 - 3*x), Interval.open(Rational(3,2), 3))
+    assert is_monotonic(1/(x**2 - 3*x), Interval.open(1.5, 3))
     assert is_monotonic(1/(x**2 - 3*x), Interval.Lopen(3, oo))
     assert is_monotonic(x**3 - 3*x**2 + 4*x, S.Reals)
     assert not is_monotonic(-x**2, S.Reals)
--- a/sympy/calculus/tests/test_singularities.py
+++ b/sympy/calculus/tests/test_singularities.py
@@ -66,7 +66,7 @@ def test_is_decreasing():
     """Test whether is_decreasing returns correct value."""
     b = Symbol('b', positive=True)
 
-    assert is_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3))
+    assert is_decreasing(1/(x**2 - 3*x), Interval.open(Rational(3,2), 3))
     assert is_decreasing(1/(x**2 - 3*x), Interval.Lopen(3, oo))
     assert not is_decreasing(1/(x**2 - 3*x), Interval.Ropen(-oo, Rational(3, 2)))
     assert not is_decreasing(-x**2, Interval(-oo, 0))
@@ -80,12 +80,14 @@ def test_is_strictly_decreasing():
         1/(x**2 - 3*x), Interval.Ropen(-oo, Rational(3, 2)))
     assert not is_strictly_decreasing(-x**2, Interval(-oo, 0))
     assert not is_strictly_decreasing(1)
+    assert is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(Rational(3,2), 3))
+    print(is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3)))
     assert is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3))
 
 
 def test_is_monotonic():
     """Test whether is_monotonic returns correct value."""
-    assert is_monotonic(1/(x**2 - 3*x), Interval.open(1.5, 3))
+    assert is_monotonic(1/(x**2 - 3*x), Interval.open(Rational(3,2), 3))
     assert is_monotonic(1/(x**2 - 3*x), Interval.Lopen(3, oo))
     assert is_monotonic(x**3 - 3*x**2 + 4*x, S.Reals)
     assert not is_monotonic(-x**2, S.Reals)
--- a/sympy/calculus/singularities.py
+++ b/sympy/calculus/singularities.py
@@ -274,9 +274,13 @@ def is_decreasing(expression, interval=S.Reals, symbol=None):
     >>> from sympy import S, Interval, oo
     >>> is_decreasing(1/(x**2 - 3*x), Interval.open(S(3)/2, 3))
     True
+    >>> is_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3))
+    True
     >>> is_decreasing(1/(x**2 - 3*x), Interval.Lopen(3, oo))
     True
     >>> is_decreasing(1/(x**2 - 3*x), Interval.Ropen(-oo, S(3)/2))
+    False
+    >>> is_decreasing(1/(x**2 - 3*x), Interval.Ropen(-oo, 1.5))
     False
     >>> is_decreasing(-x**2, Interval(-oo, 0))
     False
@@ -318,6 +322,8 @@ def is_strictly_decreasing(expression, interval=S.Reals, symbol=None):
     >>> is_strictly_decreasing(1/(x**2 - 3*x), Interval.Lopen(3, oo))
     True
     >>> is_strictly_decreasing(1/(x**2 - 3*x), Interval.Ropen(-oo, S(3)/2))
+    False
+    >>> is_strictly_decreasing(1/(x**2 - 3*x), Interval.Ropen(-oo, 1.5))
     False
     >>> is_strictly_decreasing(-x**2, Interval(-oo, 0))
     False
@@ -362,6 +368,8 @@ def is_monotonic(expression, interval=S.Reals, symbol=None):
     >>> from sympy import is_monotonic
     >>> from sympy.abc import x, y
     >>> from sympy import S, Interval, oo
+    >>> is_monotonic(1/(x**2 - 3*x), Interval.open(S(3)/2, 3))
+    True
     >>> is_monotonic(1/(x**2 - 3*x), Interval.open(1.5, 3))
     True
     >>> is_monotonic(1/(x**2 - 3*x), Interval.Lopen(3, oo))
--- a/sympy/calculus/singularities.py
+++ b/sympy/calculus/singularities.py
@@ -272,7 +272,7 @@ def is_decreasing(expression, interval=S.Reals, symbol=None):
     >>> from sympy import is_decreasing
     >>> from sympy.abc import x, y
     >>> from sympy import S, Interval, oo
-    >>> is_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3))
+    >>> is_decreasing(1/(x**2 - 3*x), Interval.open(S(3)/2, 3))
     True
     >>> is_decreasing(1/(x**2 - 3*x), Interval.Lopen(3, oo))
     True
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -2,7 +2,7 @@
 from collections import defaultdict
 from collections.abc import Mapping
 from itertools import chain, zip_longest
-from typing import Set, Tuple
+from typing import Set, Tuple, Any
 
 from .assumptions import ManagedProperties
 from .cache import cacheit
@@ -82,6 +82,7 @@ class Basic(Printable, metaclass=ManagedProperties):
                 )
 
     _args: 'Tuple[Basic, ...]'
+    _mhash: 'Any'
 
     # To be overridden with True in the appropriate subclasses
     is_number = False
@@ -140,7 +141,7 @@ class Basic(Printable, metaclass=ManagedProperties):
             raise NotImplementedError(msg)
         return super().__reduce_ex__(protocol)
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         # hash cannot be cached using cache_it because infinite recurrence
         # occurs as hash is needed for setting cache dictionary keys
         h = self._mhash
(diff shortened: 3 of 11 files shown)


ThePauliPrinciple:
Not really sure why this makes `is_decreasing` not return anything anymore.

oscarbenjamin:
> Not really sure why this makes `is_decreasing` not return anything anymore.

Maybe add something that detects when `Basic.__eq__` is about to return something different from before so you can get into it with a debugger.

ThePauliPrinciple:
Alright, looks like this comparison
```
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/calculus/tests/test_singularities.py", line 83, in test_is_strictly_decreasing
    assert is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3))
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/calculus/singularities.py", line 328, in is_strictly_decreasing
    return monotonicity_helper(expression, lambda x: x < 0, interval, symbol)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/calculus/singularities.py", line 158, in monotonicity_helper
    return interval.is_subset(predicate_interval)
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/sets/sets.py", line 399, in is_subset
    if self.intersect(other) == self:
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/core/basic.py", line 376, in __eq__
    raise TypeError
```
in `is_subset` is the problem, since the two intervals here are not the same anymore:
` <class 'sympy.sets.sets.Interval'> <class 'sympy.sets.sets.Interval'> (3/2, 3, True, True) (1.50000000000000, 3, True, True)`
as `3/2` and `1.5000000000000000` are not structurally equivalent.

oscarbenjamin:
I thought I had removed that line but apparently I just left a comment:
https://github.com/sympy/sympy/blob/81cd2630631e315586f796646a7f9268e71bf484/sympy/sets/sets.py#L395-L399
For the problem at hand though I guess trace it back a bit. Why is this comparing a set with float boundaries against one with rational boundaries?

ThePauliPrinciple:
Because it obtains a rational interval from `solveset` which it compares to a float interval from the input.

oscarbenjamin:
> Because it obtains a rational interval from `solveset` which it compares to a float interval from the input.

Maybe it shouldn't be expected to work then. Where is the test?

ThePauliPrinciple:
> I thought I had removed that line but apparently I just left a comment:

This method should probably be made to either return a default value or raise an error if no if branch returns anything.

ThePauliPrinciple:
> > Because it obtains a rational interval from `solveset` which it compares to a float interval from the input.
> 
> Maybe it shouldn't be expected to work then. Where is the test?

Yes, this shouldn't be expected to work. It is similar to the other failures, I got confused due to the `None` being returned.

oscarbenjamin:
If there are tests that only fail on PyPy then that implies that we have a difference of behaviour somehow between SymPy running on PyPy and SymPy running on CPython. That's a bug in one of SymPy, PyPy or CPython that should be reported or fixed.

How exactly can the difference in behaviour be reproduced?

sympy/calculus/tests/test_singularities.py
```diff
@@ -80,12 +80,12 @@ def test_is_strictly_decreasing():
         1/(x**2 - 3*x), Interval.Ropen(-oo, Rational(3, 2)))
     assert not is_strictly_decreasing(-x**2, Interval(-oo, 0))
     assert not is_strictly_decreasing(1)
-    assert is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3))
```
ThePauliPrinciple, line 83: This test was failing on PyPy, but not in any of the other tests. It makes sense that it should fail.

sympy/calculus/tests/test_singularities.py
```diff
@@ -80,12 +80,12 @@ def test_is_strictly_decreasing():
...
+    assert is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(Rational(3,2), 3))
 
 
 def test_is_monotonic():
     """Test whether is_monotonic returns correct value."""
-    assert is_monotonic(1/(x**2 - 3*x), Interval.open(1.5, 3))
```
ThePauliPrinciple, line 88: I changed this too, since it was similar

sympy/calculus/tests/test_singularities.py
```diff
@@ -66,7 +66,7 @@ def test_is_decreasing():
     """Test whether is_decreasing returns correct value."""
     b = Symbol('b', positive=True)
 
-    assert is_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3))
```
ThePauliPrinciple, line 69: This also fails in non pypy tests

ThePauliPrinciple:
I can't get any useful output on this unfortunately.

ThePauliPrinciple:
Looks like the problem in optional-dependencies is not due to this PR, which explains why I couldn't figure out why these changes cause that.

ThePauliPrinciple:
There really seems to be something buggy about this. Now it fails in 3.10, but doesn't fail in in the other tests (for as far as they run). Locally it fails for me:
```python
assert is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3))
```
It returns `None`.
And I would expect this to fail, not sure why it passes on any of the tests.

ThePauliPrinciple:
Surely the reported speedups aren't caused by this PR?

ThePauliPrinciple:
Looks like the result is actually non-deterministic:
```
============================================================================================= tests finished: 5 passed, 1 failed, in 1.60 seconds =============================================================================================
DO *NOT* COMMIT!
(sympy-dev) thepauli@DESKTOP-QN36B9B:/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy$ vi sympy/calculus/tests/test_singularities.py
(sympy-dev) thepauli@DESKTOP-QN36B9B:/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy$ bin/test sympy/calculus/tests/test_singularities.py
============================================================================================================= test process starts =============================================================================================================
executable:         /home/thepauli/anaconda3/envs/sympy-dev/bin/python  (3.9.7-final-0) [CPython]
architecture:       64-bit
cache:              yes
ground types:       python
numpy:              None
random seed:        63826900
hash randomization: on (PYTHONHASHSEED=2410301442)

sympy/calculus/tests/test_singularities.py[6] ....True
F.                                                                                                                                                                                     [FAIL]

_______________________________________________________________________________________________________________________________________________________________________________________________________________________________________________
___________________________________________________________________________________ sympy/calculus/tests/test_singularities.py:test_is_strictly_decreasing ____________________________________________________________________________________
  File "/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy/sympy/calculus/tests/test_singularities.py", line 85, in test_is_strictly_decreasing
    assert is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3)) is None
AssertionError

============================================================================================= tests finished: 5 passed, 1 failed, in 1.68 seconds =============================================================================================
DO *NOT* COMMIT!
(sympy-dev) thepauli@DESKTOP-QN36B9B:/mnt/c/Users/paul/Documents/PythonScripts/Libs/SymPy/sympy$ bin/test sympy/calculus/tests/test_singularities.py
============================================================================================================= test process starts =============================================================================================================
executable:         /home/thepauli/anaconda3/envs/sympy-dev/bin/python  (3.9.7-final-0) [CPython]
architecture:       64-bit
cache:              yes
ground types:       python
numpy:              None
random seed:        13438830
hash randomization: on (PYTHONHASHSEED=2112566269)

sympy/calculus/tests/test_singularities.py[6] ....None
```

Note that during one run it is printing `True` while in another it is printing `False`.

ThePauliPrinciple:
The bug seems to boil down to
```python
list(ordered(set([Interval.open(Rational(3,2), 3), Interval.open(1.50000000000000, 3)])))
```
being underterministic. Then apparently one order leads to a different result compared to the other order when those two are not equal (but not when they are equal).

This happens in the `Intersection` class, which is used in the `intersect` method which is used in 
https://github.com/sympy/sympy/blob/81cd2630631e315586f796646a7f9268e71bf484/sympy/sets/sets.py#L395-L400
`is_subset`.

Not yet sure how to fix this or what is actually the reason why this ordering is important.

oscarbenjamin:
I've seen this before. The `ordered` function does not work properly with float vs rational:
```python
In [1]: list(ordered([S.Half, 0.5]))
Out[1]: [1/2, 0.5]

In [2]: list(ordered([0.5, S.Half]))
Out[2]: [0.5, 1/2]
```
The ultimate fix for that is #20033 but it could be possible to add a workaround for in the mean time.

sympy/calculus/tests/test_singularities.py
```diff
@@ -80,12 +80,13 @@ def test_is_strictly_decreasing():
...
-    assert is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3))
-
+    assert is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(Rational(3,2), 3))
+    assert is_strictly_decreasing(1/(x**2 - 3*x), Interval.open(1.5, 3)) is None
 
 def test_is_monotonic():
     """Test whether is_monotonic returns correct value."""
-    assert is_monotonic(1/(x**2 - 3*x), Interval.open(1.5, 3))
+    assert is_monotonic(1/(x**2 - 3*x), Interval.open(Rational(3,2), 3))
+    assert is_monotonic(1/(x**2 - 3*x), Interval.open(1.5, 3)) is None
```
asmeurer, line 89: Changes like this make me hesitant to accept that making rationals and floats unequal is actually a good idea (see [this comment](https://github.com/sympy/sympy/pull/20033#issuecomment-710417294)). 1.5 is exactly equal to 3/2 as a floating point number. The answer here is fortunately not wrong, just not as helpful anymore, but one could imagine scenarios where this breaks things in a more serious way.

This is obviously more of a question for https://github.com/sympy/sympy/pull/20033 than here. This PR just makes things consistent, which is good, and the changes are pretty minimal that it wouldn't be a big deal to revert them if we decide to go the other way.

ThePauliPrinciple: Perhaps we can get the best of both worlds in this case. The mathematical question here is whether the expr is "monotonic" on a certain interval and it most certainly is monotonic along the given `Float` interval. So this function should deal with comparing based on "value equivalence" rather than "structural equivalence". So I think 
https://github.com/sympy/sympy/blob/7c0fa071cb3e2f267a6cb9767a093a8ebbf10375/sympy/sets/sets.py#L399-L400
might be a wrong use of `==` and should use a comparison based on value.

ThePauliPrinciple:
If this is the prefered solution, I do believe that the tests should reflect this with checks for both rational and float intervals.

ThePauliPrinciple:
Alright looks like this idea doesn't actually work and I was passing tests by coincidence.

Considering that currently this PR does not yield consistent results and I don't know how to make it do so, I think I want to close this, unless there are any suggestions to make this update work with `is_subset`.

ThePauliPrinciple:
This surely is a problem:
```python
>>> fi=Interval(3/2,oo)
>>> ri=Interval(Rational(3,2),oo)
>>> for i in (fi, ri):
...     for u in ((fi, ri), (ri, fi)):
...         print(i, u)
...         i.is_subset(Union(*u))
...
Interval(1.50000000000000, oo) (Interval(1.50000000000000, oo), Interval(3/2, oo))
True
Interval(1.50000000000000, oo) (Interval(3/2, oo), Interval(1.50000000000000, oo))
True
Interval(3/2, oo) (Interval(1.50000000000000, oo), Interval(3/2, oo))
Interval(3/2, oo) (Interval(3/2, oo), Interval(1.50000000000000, oo))
```
how can a be a subset of the union of a and b, but b not be a subset of the union of a and b?

oscarbenjamin:
The issue with floats is basically that we don't know the origin of the floats so e.g.:
```python
In [25]: a = Rational(3, 2) + Rational(1, 10**20)

In [26]: a
Out[26]: 
150000000000000000001
─────────────────────
100000000000000000000

In [27]: i = Interval(a, 2)

In [28]: ie = i.evalf()

In [29]: i
Out[29]: 
⎡150000000000000000001   ⎤
⎢─────────────────────, 2⎥
⎣100000000000000000000   ⎦

In [30]: ie
Out[30]: [1.5, 2.0]

In [31]: Rational(3, 2) in i
Out[31]: False

In [32]: Rational(3, 2) in ie
Out[32]: True
```
The way I think about this is that where I have a float that is equal to some exact rational number I imagine that the float represents a range of values around that number (an open set). If the statement isn't true for an arbitrarily small open set then SymPy shouldn't give True. With that interpretation I am not sure if the test results that you are currently worrying about should be expected to pass.

ThePauliPrinciple:
I managed to break down the issue more fundamentally, with `Float` and `Rational` not being equal, this no longer holds:
```python
>>> Interval(1.5,oo).intersect(Interval(Rational(3,2),oo)) == Interval(Rational(3,2),oo).intersect(Interval(3/2,oo))
False
```
I do believe it should not matter for intersect what the order is, but it does now (it chooses `Rational(3,2)` vs `3/2` based on order).

ThePauliPrinciple:
Should those intersects return `None` since they don't have a well defined meaning?

ThePauliPrinciple:
Still doesn't solve it. Since I can't get a deterministic result I'm closing this.

oscarbenjamin:
> Still doesn't solve it. Since I can't get a deterministic result I'm closing this.

I'll give this some more thought...

oscarbenjamin:
Something like this seems to work:
```diff
diff --git a/sympy/core/sorting.py b/sympy/core/sorting.py
index f255bbf..6ba314e 100644
--- a/sympy/core/sorting.py
+++ b/sympy/core/sorting.py
@@ -172,6 +172,8 @@ def _node_count(e):
     # some object has a non-Basic arg, it needs to be
     # fixed since it is intended that all Basic args
     # are of Basic type (though this is not easy to enforce).
+    if e.is_Float:
+        return 0.5
     return 1 + sum(map(_node_count, e.args))
 
 
diff --git a/sympy/sets/sets.py b/sympy/sets/sets.py
index 4e88ebd..d17658f 100644
--- a/sympy/sets/sets.py
+++ b/sympy/sets/sets.py
@@ -2463,7 +2463,7 @@ def simplify_intersection(args):
     args = set(args)
     new_args = True
     while new_args:
-        for s in args:
+        for s in ordered(args):
             new_args = False
             for t in args - {s}:
                 new_set = intersection_sets(s, t)
```

ThePauliPrinciple:
Do I understand correctly that this would change `ordered` such that it always sorts `Float` and `Rational` that have the same value in the same order? That would indeed be helpful. 
I think I would prefer putting it here:
```diff
diff --git a/sympy/sets/handlers/intersection.py b/sympy/sets/handlers/intersection.py
index 1980251c5d..fc31545494 100644
--- a/sympy/sets/handlers/intersection.py
+++ b/sympy/sets/handlers/intersection.py
@@ -426,7 +426,7 @@ def intersection_sets(a, b): # noqa:F811
             start = a.start
             left_open = a.left_open
         else:
-            start = a.start
+            start = ordered([a,b])[0].start
             left_open = a.left_open or b.left_open

         if a.end < b.end:
@@ -436,7 +436,7 @@ def intersection_sets(a, b): # noqa:F811
             end = b.end
             right_open = b.right_open
         else:
-            end = a.end
+            end = ordered([a,b])[0].end
             right_open = a.right_open or b.right_open

         if end - start == 0 and (left_open or right_open):
```

 so that `intersection_sets(a,b)=intersection_sets(b,a)` always holds. Or is there a reason not to do this?

oscarbenjamin:
> Or is there a reason not to do this?

We could do both. I was just looking at the source of the nondeterminism in this particular example.

ThePauliPrinciple:
This still needs cleanup

ThePauliPrinciple:
This is ready for review again.

oscarbenjamin:
Looks good.

oscarbenjamin:
Actually the release note needs clarification. Can you edit it here?
https://github.com/sympy/sympy/wiki/Release-Notes-for-1.10

If we say that Basic now behaves the same as Expr then what does that mean? How were they different and what is changed to make them the same?

ThePauliPrinciple:
> Actually the release note needs clarification. Can you edit it here? https://github.com/sympy/sympy/wiki/Release-Notes-for-1.10
> 
> If we say that Basic now behaves the same as Expr then what does that mean? How were they different and what is changed to make them the same?

Is it acceptable now?

oscarbenjamin:
Yes, looks good.

Issue #22654, opened 2021-12-12
Printer support for "all" classes


oscargus:
To get some impression of which SymPy classes that are (well) printable, I added the following lines to `_test_args`, which is executed on "all" classes:
```
    print("\nsrepr: ", srepr(obj))
    print("str:   ", str(obj))
    print("latex: ", latex(obj))
    print("pretty:")
    print(pretty(obj))
```

The (rather long) result is here:
[printing.txt](https://github.com/sympy/sympy/files/7698800/printing.txt)

Now, when I redirected to a file, it seems like the ASCII-printer was used, so some additional errors because of that.

I'll try to add specific issues identified in this thread, but it can be worth checking if one is up for improving the printer functionality.

oscargus:
One step would to have the class names printed using `\text` for the LaTeX-printer. For example, now it looks like:
![image](https://user-images.githubusercontent.com/8114497/145714671-4234b013-36b3-4d88-bd3a-4f0824386e18.png)
while 
![image](https://user-images.githubusercontent.com/8114497/145714678-d9811ce1-78b8-4bc2-82f5-15c42961a202.png)

is better (Firefox messes up pasting PNGs...).

oscargus:
This looks like a plain bug:
```
______ sympy\core\tests\test_args.py:test_sympy__codegen__fnodes__Extent _______
Traceback (most recent call last):
  File "c:\users\oscar\sympy\sympy\core\tests\test_args.py", line 435, in test_sympy__codegen__fnodes__Extent
    assert _test_args(Extent(x, y))
  File "c:\users\oscar\sympy\sympy\core\tests\test_args.py", line 104, in _test_args
    print("str:   ", str(obj))
  File "c:\users\oscar\sympy\sympy\core\_print_helpers.py", line 29, in __str__
    return sstr(self, order=None)
  File "c:\users\oscar\sympy\sympy\printing\printer.py", line 373, in __call__
    return self.__wrapped__(*args, **kwargs)
  File "c:\users\oscar\sympy\sympy\printing\str.py", line 997, in sstr
    s = p.doprint(expr)
  File "c:\users\oscar\sympy\sympy\printing\printer.py", line 293, in doprint
    return self._str(self._print(expr))
  File "c:\users\oscar\sympy\sympy\printing\printer.py", line 310, in _print
    return getattr(expr, self.printmethod)(self, **kwargs)
[...]

PR #22657, opened 2021-12-13
polys/domains: Improve `ANP.__init__()`.

skieffer:
#### Brief description of what is fixed or changed

In `ANP.__init__()`, ensure that the elements of the incoming `rep` are converted to the underlying domain whether
or not `rep` is a list.

#### Release Notes

NO ENTRY

PR #22661, opened 2021-12-13
LaTeX and pretty printing for symbolic statistics

oscargus:

Related to #22654 

`Covariance`, `Expectation`, `Probability` and `Variance` now have dedicated LaTex and pretty printing.
image
image

#### Other comments
Also changed so that a random domain prints using the `\in` symbol rather than the text `in`.

Added some rudimentary ASCII pretty-printing option for `Differential` (`d`), `WedgeProduct` (`/\`), and `TensorProduct` (`.*`). The last is a bit doubtful, but probably better than an error...

#### Release Notes

* printing
   * Dedicated LaTeX and pretty printing for `Covariance`, `Expectation`, `Probability` and `Variance`.

### DIFF
--- a/sympy/printing/latex.py
+++ b/sympy/printing/latex.py
@@ -2471,16 +2471,16 @@ class LatexPrinter(Printer):
         return result
 
     def _print_Expectation(self, expr):
-        return r"E\left[{}\right]".format(self._print(expr.args[0]))
+        return r"\operatorname{{E}}\left[{}\right]".format(self._print(expr.args[0]))
 
     def _print_Variance(self, expr):
-        return r"\operatorname{{var}}\left({}\right)".format(self._print(expr.args[0]))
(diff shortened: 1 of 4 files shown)


oscarbenjamin:
See here for the test failure: https://github.com/sympy/sympy/pull/22657#issuecomment-992342721

oscarbenjamin:
Looks good to me

bjodah:
I think "E" should be written either using `\operatorname` or simply `{\rm E}` since it's not a variable here, would you agree?

oscargus:
> I think "E" should be written either using `\operatorname` or simply `{\rm E}` since it's not a variable here, would you agree?

Yes, you are correct. I also changed P. Used `\operatorname` (`\rm` is TeX and should not be used, ever, as far as I understand it).

oscargus:
Merging this and restarting the ones that builds on it.

PR #22663, opened 2021-12-13
Better `srepr` for vectors and some LaTeX improvements

oscargus:
#### References to other Issues or PRs

Builds on #22661 so that should be merged first.
Related to #22654 

#### Brief description of what is fixed or changed
`srepr` for `BaseVector` and `BaseDyadic` did not return correct code (in the same way as other `srepr` does).

`DFT` and `IDFT` now have better LaTeX-printing.

`S.Catalan` is now printed as `G` (as per the Wiki-link provided as reference and many pther sources).

#### Other comments

Tests added for LaTeX-printing of all(?) NumberSymbols in `S`.

#### Release Notes

* printing
   * `S.Catalan` is now printed as `G` in LaTeX and pretty printers. 
   * Improved LaTeX-printing of `DFT` and `IDFT`. 
* vector
   * Improved `srepr` for `BaseVector` and `BaseDyadic`.

PR #22665, opened 2021-12-13
Fixed printing of all(?) quantum states and more

oscargus:

Builds on #22661 and #22663 so should be merged in order
Related to #22654 

All states in both `physics.secondquant` and `physics.quantum` can now be printed without errors.

#### Other comments
`String` and `Type` from `sympy.codegen.ast` now has a reasonable LaTeX-printing. Not that it is really required, but at least nice if something proper shows up in case one happens to try it...

`Extent` from  `sympy.codegen.ast` can be printed using symbolic arguments, which seems to be supported based on `test_args`.

Improved the printing of `BaseDyadic`

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/codegen/ast.py
+++ b/sympy/codegen/ast.py
@@ -914,6 +914,9 @@ class String(Atom, Token):
     def func(self):
         return lambda: self
 
+    def _latex(self, printer):
+        return r'\texttt{{"{}"}}'.format(self.text)
+
 class QuotedString(String):
     """ Represents a string which should be printed with quotes. """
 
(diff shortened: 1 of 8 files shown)


ThePauliPrinciple:
Sometimes `_latex` is used in the class being printed and sometimes a `_print_<classname>` method is added to the latex printer. Is there a reasoning for when which option is used?

oscargus:
> Sometimes `_latex` is used in the class being printed and sometimes a `_print_<classname>` method is added to the latex printer. Is there a reasoning for when which option is used?

Not really. I have primarily used the `_print_Class`-approach earlier, but in many of the current classes, they have relied on the `_latex`-approach so I continued along those lines. For the codegen-classes `_latex` was primarily selected as it is of very limited use and I didn't want to mess-up the LaTeX-printer too much...

oscargus:
I merge this as it isn't that critical (more that there should be fewer errors).

PR #22671, opened 2021-12-14
Renamed test_args and Updated docs

### LINKED ISSUE #22666: Test that classes are printable and srepr:able
It seems like there should be a test in test_args that all classes are printable (in the sense that str, srepr, latex and pretty does not give an error) and that `srepr` of the object can be used to recreate the same object. The printable, I know how to implement (although it is currently not the state of the code), but for the second it is not that simple.

Using `sympify` will not work for all classes it seems. Especially vectors that are recreated as a member of its parent system, e.g. `sympify(srepr(C.i))` will complain that the member `i` is not available. Using `eval` will require all imports to happen at the top level and not inside the test-functions (as is sometimes the case now). This may be the most feasible approach, but there are some duplicate class names I think.

Any ideas or opinions?

ghost:
Renamed `test_args` to `test_basic_subclasses` and Updated docs

May help #22666 

`test_args` is the only place that forces all subclasses to appear so renamed to `test_basic_subclasses`  and updated docs as per the changes.

#### Release Notes

NO ENTRY


oscargus:
I do not think the idea was this, rather to have some way to run different tests on "all" classes.

anutosh491:
@oscargus True , I don't think this would be the intended fix for the issue reported above !

PR #22677, opened 2021-12-15
MatrixElement symbol property added; validate shape compatible with indices

### LINKED ISSUE #22676: MatrixElement allows bad subs
```python
>>> m22 = MatrixSymbol("A",2,2)
>>> m11 = MatrixSymbol("B",1,1)
>>> m22.subs(m22,m11) == m11  # this is ok, just changing a symbol without context
True
>>> m22[1,1].subs(m22, m11)  # this is not ok -- it indicates a non-existent element in B
B[1, 1]
```
If you enter the element manually you will get the error:
>>> B[1, 1]
KeyError: (1, 1)
```
Such a substitution should either fail to change the object or else raise a KeyError.

### PR DESCRIPTION
**Pull request by smichr**

fixes #22676

##### symbol attribute has been added
MatrixElement behaves more like Indexed in terms of argument access:

```python
>>> IndexedBase('x')[0].base
x
>>> MatrixSymbol('x',2,2)[0,0].symbol
x
```

##### invalid substitutions no longer succeed

In master,
```python
X = MatrixSymbol("X", 2, 2)
Y = MatrixSymbol("Y", 1, 2)
assert X[1, 1].subs(X, Y) == Y[1, 1]
>>> Y[1,1]
...
IndexError: indices out of bounds
```
in this PR that substitution no longer succeeds so you don't end up with an invalid expression

#### Other comments

There is some divergence in how MatrixSymbol/IndexedBase and MatrixElement/Indexed behave in terms of free symbols and attributes. This should have a look by someone that works with matrix expressions to see if these changes make sense.

- [x] QUESTION: is there are reason to store Indexed.label as a Symbol but MatrixSymbol as a Str?
ANSWER: no (see #22391 and #22272)

#### Release Notes

* matrices
    * replacement of MatrixSymbol in MatrixElement requires shape to be compatible with indices to succeed
    * MatrixElement now has `symbol` attributes (analogous to `base` attribute of `IndexedBase` )
    * `eval_hyper_sum` returns as unchanged arguments for which `is_hypergeometric` is False and those which produce singularities at `a` or `b + 1` where `and` and `b` are the lower and upper limits of evaluation, respectively.
* stats
    * removed deprecation warning about using `evaluate=False`

### DIFF
--- a/sympy/concrete/summations.py
+++ b/sympy/concrete/summations.py
@@ -1296,7 +1296,7 @@ def _eval_sum_hyper(f, i, a):
 def eval_sum_hyper(f, i_a_b):
     i, a, b = i_a_b
 
-    if not f.is_hypergeometric(i):
+    if f.is_hypergeometric(i) is False:
         return
 
     if (b - a).is_Integer:
@@ -1316,7 +1316,8 @@ def eval_sum_hyper(f, i_a_b):
             # that neither limit causes evaluation issues,
             # but now we are testing 1 past the upper limit
             # so check here if this should be avoided:
-            if f.subs(i, b + 1).has(*illegal):
+            n_illegal = lambda x: sum(x.count(_) for _ in illegal)
+            if n_illegal(f) < n_illegal(f.subs(i, b + 1)):
                 return
             res2 = _eval_sum_hyper(f, i, b + 1)
             if res1 is None or res2 is None:
--- a/sympy/concrete/summations.py
+++ b/sympy/concrete/summations.py
@@ -1308,10 +1308,10 @@ def eval_sum_hyper(f, i_a_b):
                 return Piecewise(res, (old_sum, True))
         else:
             res1 = _eval_sum_hyper(f, i, a)
-            try:
-                res2 = _eval_sum_hyper(f, i, b + 1)
-            except IndexError:  # for f = Trace (or other matexpr?)
-                res2 = None
+            # this will fail if an indexed object gets sent
+            # here -- so don't send such objects here (i.e.
+            # fix at the call, not here
+            res2 = _eval_sum_hyper(f, i, b + 1)
             if res1 is None or res2 is None:
                 return None
             (res1, cond1), (res2, cond2) = res1, res2
--- a/sympy/concrete/summations.py
+++ b/sympy/concrete/summations.py
@@ -28,6 +28,7 @@ from sympy.logic.boolalg import And
 from sympy.polys.partfrac import apart
 from sympy.polys.polyerrors import PolynomialError, PolificationFailed
 from sympy.polys.polytools import parallel_poly_from_expr, Poly, factor
+from sympy.polys.polyutils import illegal
 from sympy.polys.rationaltools import together
 from sympy.series.limitseq import limit_seq
 from sympy.series.order import O
@@ -1295,6 +1296,9 @@ def _eval_sum_hyper(f, i, a):
 def eval_sum_hyper(f, i_a_b):
     i, a, b = i_a_b
 
+    if not f.is_hypergeometric(i):
+        return
+
     if (b - a).is_Integer:
         # We are never going to do better than doing the sum in the obvious way
         return None
@@ -1308,9 +1312,12 @@ def eval_sum_hyper(f, i_a_b):
                 return Piecewise(res, (old_sum, True))
         else:
             res1 = _eval_sum_hyper(f, i, a)
-            # this will fail if an indexed object gets sent
-            # here -- so don't send such objects here (i.e.
-            # fix at the call, not here
+            # a calling routine may have reasonably checked
+            # that neither limit causes evaluation issues,
+            # but now we are testing 1 past the upper limit
+            # so check here if this should be avoided:
+            if f.subs(i, b + 1).has(*illegal):
+                return
             res2 = _eval_sum_hyper(f, i, b + 1)
             if res1 is None or res2 is None:
                 return None
--- a/sympy/concrete/summations.py
+++ b/sympy/concrete/summations.py
@@ -1308,7 +1308,10 @@ def eval_sum_hyper(f, i_a_b):
                 return Piecewise(res, (old_sum, True))
         else:
             res1 = _eval_sum_hyper(f, i, a)
-            res2 = _eval_sum_hyper(f, i, b + 1)
+            try:
+                res2 = _eval_sum_hyper(f, i, b + 1)
+            except IndexError:  # for f = Trace (or other matexpr?)
+                res2 = None
             if res1 is None or res2 is None:
                 return None
             (res1, cond1), (res2, cond2) = res1, res2
--- a/sympy/concrete/summations.py
+++ b/sympy/concrete/summations.py
@@ -1311,17 +1311,15 @@ def eval_sum_hyper(f, i_a_b):
             if res is not None:
                 return Piecewise(res, (old_sum, True))
         else:
-            res1 = _eval_sum_hyper(f, i, a)
-            # a calling routine may have reasonably checked
-            # that neither limit causes evaluation issues,
-            # but now we are testing 1 past the upper limit
-            # so check here if this should be avoided:
             n_illegal = lambda x: sum(x.count(_) for _ in illegal)
-            if n_illegal(f) < n_illegal(f.subs(i, b + 1)):
+            had = n_illegal(f)
+            # check that no extra illegals are introduced
+            res1 = _eval_sum_hyper(f, i, a)
+            if res1 is None or n_illegal(res1) > had:
                 return
             res2 = _eval_sum_hyper(f, i, b + 1)
-            if res1 is None or res2 is None:
-                return None
+            if res2 is None or n_illegal(res2) > had:
+                return
             (res1, cond1), (res2, cond2) = res1, res2
             cond = And(cond1, cond2)
             if cond == False:
--- a/sympy/matrices/expressions/matexpr.py
+++ b/sympy/matrices/expressions/matexpr.py
@@ -6,6 +6,7 @@ from sympy.core.assumptions import check_assumptions
 from sympy.core.decorators import call_highest_priority
 from sympy.core.expr import Expr, ExprBuilder
 from sympy.core.logic import FuzzyBool
+from sympy.core.relational import is_le
 from sympy.core.symbol import Str, Dummy, symbols, Symbol
 from sympy.core.sympify import SympifyError, _sympify
 from sympy.external.gmpy import SYMPY_INTS
@@ -589,10 +590,14 @@ class MatrixElement(Expr):
     def __new__(cls, name, n, m):
         n, m = map(_sympify, (n, m))
         from sympy.matrices.matrices import MatrixBase
-        if isinstance(name, (MatrixBase,)):
+        if isinstance(name, MatrixBase):
             if n.is_Integer and m.is_Integer:
                 return name[n, m]
-        if isinstance(name, str):
+        elif isinstance(name, MatrixSymbol):
+            r, c = name.shape
+            if is_le(r, n) or is_le(c, m):
+                raise IndexError('indices out of bounds')
+        elif isinstance(name, str):
             name = Symbol(name)
         else:
             name = _sympify(name)
--- a/sympy/matrices/expressions/matexpr.py
+++ b/sympy/matrices/expressions/matexpr.py
@@ -6,7 +6,6 @@ from sympy.core.assumptions import check_assumptions
 from sympy.core.decorators import call_highest_priority
 from sympy.core.expr import Expr, ExprBuilder
 from sympy.core.logic import FuzzyBool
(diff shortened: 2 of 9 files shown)


sympy/matrices/expressions/matexpr.py
```diff
@@ -613,6 +621,17 @@ def doit(self, **kwargs):
...
         return self.args[1:]
 
+    def _eval_subs(self, old, new):
+        # when replacing the symbol, the indices cannot
+        # access out of bounds
+        if old == self.symbol:
+            if type(old) == type(new):
+                for i, j in zip(self.indices, new.shape):
+                    if i.is_Integer and j.is_Integer and i >= j:
+                        raise IndexError('indices access out of range')
```
oscarbenjamin, line 631: I think it would be better to check this in `MatrixElement.__new__` rather than `_eval_subs`.

smichr: I tried and failed but I think I got thrown off by thinking the MatrixBase clause was for MatrixSymbol...I will try again to make the change there.

smichr: OK, this is part of `__new__` now.

smichr:
@oscarbenjamin , any suggestions for working around mypy errors? Do I just put "  # type: ignore" on the class def?

oscarbenjamin:
> @oscarbenjamin , any suggestions for working around mypy errors? Do I just put " # type: ignore" on the class def?

I think that these errors are really pointing to bugs in the code somewhere but it's not obvious where those are. It's also possible that they are due to incorrect type-hints being applied somewhere. It should always be possible to silence mypy by adding `type: ignore` on the exact line number that is referenced in the mypy output.

smichr:
@Upabjojr @oscarbenjamin , does this look ok?

sympy/concrete/summations.py
```diff
@@ -1308,7 +1308,10 @@ def eval_sum_hyper(f, i_a_b):
                 return Piecewise(res, (old_sum, True))
         else:
             res1 = _eval_sum_hyper(f, i, a)
-            res2 = _eval_sum_hyper(f, i, b + 1)
+            try:
+                res2 = _eval_sum_hyper(f, i, b + 1)
+            except IndexError:  # for f = Trace (or other matexpr?)
+                res2 = None
```
oscarbenjamin, line 1314: Where is the IndexError coming from? It doesn't seem like a reasonable exception to catch here.

sympy/matrices/expressions/matexpr.py
```diff
@@ -589,10 +590,15 @@ class MatrixElement(Expr):
...
-        if isinstance(name, (MatrixBase,)):
+        if isinstance(name, MatrixBase):
             if n.is_Integer and m.is_Integer:
                 return name[n, m]
-        if isinstance(name, str):
+            name = _sympify(name)  # change mutable into immutable
+        elif isinstance(name, MatrixSymbol):
+            r, c = name.shape
+            if is_le(r, n) or is_le(c, m):
+                raise IndexError('indices out of bounds')
```
oscarbenjamin, line 600: What about negative indices?

sympy/tensor/array/expressions/arrayexpr_derivatives.py
```diff
@@ -23,7 +23,7 @@ def array_derive(expr, x):
 
 @array_derive.register(Expr) # type: ignore
 def _(expr: Expr, x: Expr):
-    return ZeroArray(*x.shape)
+    return ZeroArray(*x.shape)  # type: ignore
```
oscarbenjamin, line 26: I would rather fix these things than add `ignore`. This is clearly invalid because an arbitrary Expr does not have a `.shape` attribute.

smichr: This must be a change to mypy. It couldn't be a PR that was added or else that PR wouldn't have passed tests. And this happening on all PRs. So we either patch it until it gets fixed or wait until someone can fix it and stop all other PRs. What is your preference?

oscarbenjamin: The problem was initiated by a change in mypy. I thought that was now addressed (in the ignore sense) but the diff here shows this as a new change in SymPy.

sympy/concrete/summations.py
smichr: `Trace(A).rewrite(Sum)` -- the `hyper_sum` method is attempted and that method requires sampling at `upper limit + 1` which you cannot do with a physical object like a matrix.

sympy/tensor/array/expressions/arrayexpr_derivatives.py
smichr: #22692 is passing and has the following difference to master (and did not add those differences, to it was a change to master that did this):
```diff
diff --git a/.github/workflows/runtests.yml b/.github/workflows/runtests.yml
index 132075145e..f9c6215388 100644
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           python-version: '3.9'
       - run: python -m pip install --upgrade pip
-      - run: pip install mpmath flake8
+      - run: pip install mpmath flake8 mypy
 
       - name: Basic code quality tests
         run: bin/test quality
@@ -33,6 +33,9 @@ jobs:
       - name: Run flake8 on the sympy package
         run: flake8 sympy
 
+      - name: Run mypy on the sympy package
+        run: mypy sympy
+
       - name: Detect invalid escapes like '\e'
         run: python -We:invalid -We::SyntaxWarning -m compileall -f -q sympy/
 
@@ -49,28 +52,6 @@ jobs:
       #- name: Make sure all commits have an associated author
       #  run: bin/authors_update.py
 
-
-  # ----------------------------- mypy ----------------------------- #
-
-  mypy:
-    needs: code-quality
-
-    runs-on: ubuntu-20.04
-    steps:
-      - uses: actions/checkout@v2
-        with:
-          # Clone full git history (needed for AUTHORS/mailmap)
-          fetch-depth: '0'
-
-      - uses: actions/setup-python@v2
-        with:
-          python-version: '3.9'
-      - run: python -m pip install --upgrade pip
-      - run: pip install mpmath mypy
-
-      - name: Run mypy on the sympy package
-        run: mypy sympy
-
```

smichr: @asmeurer , see previous. I think the additions of the red and removal of green diff lines is what made master start failing mypy.

sympy/concrete/summations.py
oscarbenjamin: I don't quite follow how `hyper_sum` ends up being called. Presumably this is a case where it shouldn't be called.

smichr: caught at source, now

sympy/matrices/expressions/trace.py
```diff
@@ -144,8 +145,11 @@ def get_arg_key(x):
 
     def _eval_rewrite_as_Sum(self, expr, **kwargs):
         from sympy.concrete.summations import Sum
-        i = Dummy('i')
-        return Sum(self.arg[i, i], (i, 0, self.arg.rows-1)).doit()
+        i = uniquely_named_symbol('i', expr)
+        s = Sum(self.arg[i, i], (i, 0, self.arg.rows - 1))
+        if not isinstance(expr, MatrixExpr):
+            s = s.doit()
```
oscarbenjamin, line 151: If this call to `doit` would otherwise raise `IndexError` then that suggests that something should still be fixed somewhere because `doit` shouldn't raise `IndexError`.

smichr: Have a look at the `eval_sum_hyper` routine. It needs to access the `upper bound + 1`: `res2 = _eval_sum_hyper(f, i, b + 1)`. This will fail for any indexed object which cannot be computed at that bound. I left a note on the routine regarding indexed objects and I added a safeguard in the trace routine that is writing trace of a matrix as a sum. In the case of an actual matrix the dimensions are concrete and `eval_sum_hyper` is not called.

oscarbenjamin: Can you show example code needed to set this up so I can see the error? Would I see the error on master or only with the changes in this PR?

smichr: Saying that it shouldn't happen is equivalent to saying either "do routine writing sums with indexed objects should call the `eval_sum_hyper` routine. But calling that routine is part of the `Sum.doit` machinery. So any routine writing Sum with indexed objects should not call doit. And that is the change that I made in `trace`.

oscarbenjamin: What I was looking for was how to reproduce this. I see now that with this PR it's:
```python
In [1]: n = Symbol('n', integer=True)

In [2]: M = MatrixSymbol('M', n, n)

In [3]: trace(M).rewrite(Sum).doit()
IndexError 
```
Strangely the `integer=True` is needed here.

As you say it goes wrong from here:
https://github.com/sympy/sympy/blob/14cf005b8fea132207654603cc6495900040106a/sympy/concrete/summations.py#L1310-L1311
There could be lots of situations where just substituting `b+1` in might break something so either the exception needs to be caught here (i.e. your previous code) or there should be some code to determine whether this is an appropriate function to use in this routine at all. There isn't much documentation but I presume that `eval_sum_hyper` is only supposed to work for hypergeometric functions and `M[i, i]` is clearly not hypergeometric in `i`.

oscarbenjamin: Catching the exception is not exactly satisfactory because this kind of potentially invalid operation could lead to all sorts of different exceptions.

oscarbenjamin: These lines seem to be responsible for this output for example:
```python
In [1]: b = symbols('b', integer=True, positive=True)

In [2]: summation(1/(x-b)**2, (x, 0, b-1))
Out[2]: 
 ⎛         ⅈ⋅π⎞      
Φ⎝1, 2, b⋅ℯ   ⎠ + zoo
```

oscarbenjamin: This is giving zoo in cases where actually the result might be well defined:
```python
In [7]: Sum(1/(x-b)**2, (x, 0, b-1)).subs(b, 2).doit()
Out[7]: 5/4

In [8]: b = symbols('b', integer=True, positive=True)

In [9]: s = Sum(1/(x-b)**2, (x, 0, b-1))

In [10]: s.subs(b, 2).doit()
Out[10]: 5/4

In [11]: s.doit().subs(b, 2)
Out[11]:
ZeroDivisionError  
```

smichr: I added a `is_hypergeometric` tests to the routine and then a check that illegals are not introduced by the `b + 1` evaluation. From the wolframalpha result, it appears that a solution is possible. That's a separate issue; at least we can check that there is no `zoo`.

oscarbenjamin: I'm not sure how well `is_hypergeometric` works. Note that it just calls `hypersimp` which is already being called here I think.

smichr: bad evaluations are avoid now and tested as:
```python
def test_pr_22677():
    b = Symbol('b', integer=True, positive=True)
    assert Sum(1/x**2,(x, 0, b)).doit() == Sum(x**(-2), (x, 0, b))
    assert Sum(1/(x - b)**2,(x, 0, b-1)).doit() == Sum(
        (-b + x)**(-2), (x, 0, b - 1))
```
The first one gives the following in master:
```python
>>> Sum(1/x**2,(x,0,b)).doit()
zoo - (b**2 + 2*b + 1)*lerchphi(1, 2, b + 1)/(b + 1)**2
```

sympy/matrices/expressions/matexpr.py
```diff
@@ -589,10 +589,14 @@ class MatrixElement(Expr):
...
         n, m = map(_sympify, (n, m))
         from sympy.matrices.matrices import MatrixBase
-        if isinstance(name, (MatrixBase,)):
+        if isinstance(name, MatrixBase):
             if n.is_Integer and m.is_Integer:
                 return name[n, m]
-        if isinstance(name, str):
+            name = _sympify(name)  # change mutable into immutable
+        elif isinstance(name, MatrixSymbol):
+            if not name.valid_index(n, m):
```
oscarbenjamin, line 597: I think that `valid_index` should always be checked, not just for MatrixSymbol e.g.:
```python
In [18]: n = Symbol('n', integer=True, positive=True)

In [19]: M = MatrixSymbol('M', n, n)

In [20]: M[3, m]
Out[20]: (M)[3, m]

In [21]: M[3, m].subs(M, Matrix([[1, 2], [3, 4]]))
Out[21]: 
⎛⎡1  2⎤⎞      
⎜⎢    ⎥⎟[3, m]
⎝⎣3  4⎦⎠ 
```

smichr: good point! To me this seems like a strange constructor in that you are creating an unevaluated indexed object rather than simply indexing an object. It works, but having to put these checks here seems odd to me.

PR #22688, opened 2021-12-16
Added Essential Information In .evaluate() Doc-string Regarding Use Of Declaring Real Symbols

### LINKED ISSUE #22682: `solve()` is not able to solve an equation generated by `Line`
It can be observed that
```
Eq(-6*x-5*y-2)
```
and 
```
Eq(Line(Point(3,-4),Point(-2,2)).equation())
```
both output
```
-6⋅x - 5⋅y - 2 = 0
```
however, if I were to type
```
solve(Eq(Line(Point(3,-4), Point(-2,2)).equation()),y)
```
or
```
solve(Eq(Line(Point(3,-4), Point(-2,2)).equation()),x)
```
I get the output of 
```
[]
```
when if I were to instead type
```
solve(-6*x-5*y-2,y)
```
or
```
solve(Eq(-6*x-5*y-2),y)
```
I will get the (somewhat ugly, but correct) output of 
```
⎡  6⋅x   2⎤
⎢- ─── - ─⎥
⎣   5    5⎦
```

faze-geek:

Closes #22682

 In SymPy two different symbols with the same name but different assumptions are not considered to be equivalent. Updating the Doc-string regarding use of real symbols is important to avoid multiple wrong results.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/geometry/line.py
+++ b/sympy/geometry/line.py
@@ -2101,6 +2101,9 @@ class Line2D(LinearEntity2D, Line):
         y : str, optional
             The name to use for the y-axis, default value is 'y'.
 
+        Note : Symbols being used need to be declared as real to avoid
+               unexpected results. The best approach is to provide your
+               own symbols as .equation(x, y) .
         Returns
         =======
 


faze-geek:
Ping @oscarbenjamin @smichr .

sympy/geometry/line.py
```diff
@@ -2101,6 +2101,9 @@ def equation(self, x='x', y='y'):
         y : str, optional
             The name to use for the y-axis, default value is 'y'.
 
+        Note : Symbols being used need to be declared as real to avoid
```
ThePauliPrinciple, line 2104: Should `Symbol`s always be declared as real? Or should they be declared with "appropriate" assumptions? If they always need to be real here, perhaps it can be done by the library rather than the user?

faze-geek: I get what you are suggesting but I still have my doubts . It would make sense in this case if not declaring an assumption could be treated as declaring real ,it would be easier for the user if he isn't aware .

faze-geek:
Ping @oscarbenjamin . I believe this does not need more work .Can you review and merge if everything seems fine !

sympy/geometry/line.py
```diff
@@ -2101,6 +2101,9 @@ def equation(self, x='x', y='y'):
...
+               unexpected results. The best approach is to provide your
+               own symbols as .equation(x, y) .
```
oscarbenjamin, line 2106: I don't expect that anyone reading this would understand what point you are trying to make. A doctest example showing how to pass your own symbols and make use of them would be better.

Also probably the function should be changed so that it doesn't declare the symbols as real.

faze-geek: Thanks .I'll try to improve on it !

PR #22691, opened 2021-12-16
Remove the autowrap package as an optional dependency

asmeurer:
It was never actually a dependency. It was added by
161964db4e37e5cee54c97a7cb7a00a7b84efc75. Presumably the author of that commit
confused the autowrap *package* with the autowrap *SymPy* function/submodule.
But the two are unrelated. A grep of "autowrap" in the codebase reveals that
all references to it refer to the SymPy function, not the external package.

Thank you to Samuel Wein for pointing this out to me via private
correspondence.

#### Release Notes

NO ENTRY


oscarbenjamin:
MAybe a new release of mypy is causing CI problems

asmeurer:
Maybe we should pin mypy, that way we can update it whenever someone is able to fix the issues with a new version, but it won't break the CI. The only issue is if we will actually remember to update it.

oscarbenjamin:
Ideally we should just fix problems like this. I'm fairly sure that mypy is complaining about either a bug or an incorrect annotation somewhere.

asmeurer:
That's fair, but fixes like https://github.com/sympy/sympy/pull/22699 just workaround the problem instead of actually fixing it. We can expect that when the mypy update causes every PR build to fail. 

As an aside, I don't think mypy should be in the code quality build. It doesn't make sense to me that even a legitimate mypy error should prevent the other tests from running. Typically the code quality checks mean something very basic is wrong that should be fixed before anything else is looked at, but typing errors are often far from basic.

asmeurer:
Actually there's a simpler solution. If we split mypy into a separate build, we can temporarily make it not required when this happens. That way the failure will still be there, but it won't block merging (I think only admins are able to do this at https://github.com/sympy/sympy/settings/branch_protection_rules/18177, but we can give more people admin or maintainer access if that's a problem). I've implemented this at https://github.com/sympy/sympy/pull/22700

oscarbenjamin:
> That's fair, but fixes like #22699 just workaround the problem instead of actually fixing it.

I agree. I described the actual fix here: https://github.com/sympy/sympy/issues/22697#issuecomment-996907315

If we want to actually use mypy then it does mean that contributors need to learn how to use it and try to actually fix the problems.

oscargus:
This is good to go, right? (Maybe even worth back-porting to 1.10, just to get away with it?)

asmeurer:
Yes. I don't think there's any point in backporting it as it doesn't change anything in library code, only the release scripts. Having it there doesn't really affect anything as far as the release goes, it's just confusing.

moorepants:
LGTM!

PR #22692, opened 2021-12-16
doctest: changes several doctests that created Basics with non-Basic arguments

ThePauliPrinciple:

Helps with https://github.com/sympy/sympy/pull/22445

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/series/sequences.py
+++ b/sympy/series/sequences.py
@@ -424,7 +424,8 @@ class SeqExpr(SeqBase):
 
     >>> from sympy.series.sequences import SeqExpr
     >>> from sympy.abc import x
-    >>> s = SeqExpr((1, 2, 3), (x, 0, 10))
+    >>> from sympy.core.containers import Tuple
+    >>> s = SeqExpr(Tuple(1, 2, 3), Tuple(x, 0, 10))
     >>> s.gen
     (1, 2, 3)
     >>> s.interval
--- a/sympy/tensor/functions.py
+++ b/sympy/tensor/functions.py
@@ -89,14 +89,14 @@ def shape(expr):
 
     This function returns the shape of any object representing matrix or array.
 
-    >>> from sympy import shape, Array, Matrix, Integral
+    >>> from sympy import shape, Array, ImmutableDenseMatrix, Integral
     >>> from sympy.abc import x
     >>> A = Array([1, 2])
     >>> shape(A)
     (2,)
     >>> shape(Integral(A, x))
     (2,)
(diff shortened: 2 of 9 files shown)


sympy/series/sequences.py
```diff
@@ -424,7 +424,8 @@ class SeqExpr(SeqBase):
 
     >>> from sympy.series.sequences import SeqExpr
     >>> from sympy.abc import x
-    >>> s = SeqExpr((1, 2, 3), (x, 0, 10))
+    >>> from sympy.core.containers import Tuple
+    >>> s = SeqExpr(Tuple(1, 2, 3), Tuple(x, 0, 10))
```
oscarbenjamin, line 428: Maybe this class should sympify its args

sympy/tensor/functions.py
```diff
@@ -89,14 +89,14 @@ def shape(expr):
...
-    >>> from sympy import shape, Array, Matrix, Integral
+    >>> from sympy import shape, Array, ImmutableDenseMatrix, Integral
     >>> from sympy.abc import x
     >>> A = Array([1, 2])
     >>> shape(A)
     (2,)
     >>> shape(Integral(A, x))
     (2,)
-    >>> M = Matrix([1, 2])
+    >>> M = ImmutableDenseMatrix([1, 2])
```
oscarbenjamin, line 99: Shouldn't this get sympified somewhere?

sympy/series/sequences.py
ThePauliPrinciple: It actually doesn't have a `__new__` method at all and the docstring says `Various sequences should inherit from this class.` So I interpreted it as a class that should normally not be used by users (similar to `Basic`).

oscarbenjamin: Do the subclasses sympify their arguments?

sympy/tensor/functions.py
ThePauliPrinciple: It does in `Integral`, but not in `NewExpr` (user defined subclass). Here I thought the doc would be easier to read if `NewExpr` didn't have a `__new__` method defined

oscarbenjamin: Ok, fair enough

sympy/series/sequences.py
ThePauliPrinciple: Yes, there were also no tests that had to be changed for the subclasses as far as I can remember. They do use `sympify` over `_sympify` though (and are fairly complicated).

oscarbenjamin:
Looks good

PR #22693, opened 2021-12-17
Add guide listing all of SymPy's optional dependencies

### LINKED ISSUE #5259: List of dependencies
We should include in the README, and probably in the docs somewhere too, a list of all the "dependencies" of SymPy.  Now, obviously, SymPy's only dependency is Python, but there are several optional dependencies like IPython and gmpy that are not required but extent the capabilities of SymPy if they are installed. 

I am putting this here instead of just doing it and making a pull request because I don't know exactly how things work with things like numpy and scipy, and I also don't know if I am forgetting anything.  So far, I have this:

"""
Dependencies
------------

One of the fundamental design decisions behind SymPy is that it should not have any external dependencies besides Python to install.  Therefore, the only real "dependency" of SymPy is an installation of Python 2.4, 2.5, 2.6, or 2.7 (note that Python 2.4 support is deprecated and will no longer work in the next release).

However, there are several packages that are not required to use SymPy, but that will enhance SymPy if they are installed.  These packages are:

- IPython ( http://ipython.scipy.org/moin/ ).  IPython is a third party Python interactive shell that has many more features over the built-in Python interactive shell, such as tab-completetion and [what other important features should I mention here?].  If IPython is installed, isympy will automatically use it.  Otherwise, it will fall back to the regular Python interactive shell.  You can override this behavior by setting the -c option to isympy, like `isympy -c python`.

- gmpy ( http://code.google.com/p/gmpy/ ). gmpy is a Python wrapper around the GNU Multiple Precision Arithmetic Library (GMP). If gmpy is installed, it may make certain operations in SymPy faster, because it will use gmpy as the ground type instead of the built-in Python ground types.  If gmpy is not installed, it will fall back to the default Python ground types.  You can override this behavior by setting the SYMPY_GROUND_TYPES environment variable, such as `SYMPY_GROUND_TYPES=python isympy`. [Note: this should be an option to isympy!]

- Cython. [What exactly is the situation with Cython?]

- Numpy. [Ditto]

- Scipy. [Ditto]

- [Code generation dependencies?]

- [Other dependencies?]

Note that mpmath and pyglet, our floating point and plotting libraries respectively, are included with SymPy, so it is unnecessary to install them.  Indeed, SymPy will always use the version of mpmath or pyglet that comes with SymPy, even if a newer version is installed in the system.  This is done for compatibility reasons.

"""

OK, the text in [] are comments.  I need help from others writing the rest of this.  Note that I do not want to include any development dependencies here (like Sphinx), because the user does not care about that (maybe we could have that list somewhere else in our development docs).   

But basically for the others, explain what happens when they are installed and what happens when they are not (and how to override, if relevant).  

I just realized that the best way to do this would be through the wiki.  So if anyone has anything to add to this, please do it at https://github.com/sympy/sympy/wiki/Dependencies .  And then when we have it finished we can add it to the README and regular docs.
```

Original issue for #5259: http://code.google.com/p/sympy/issues/detail?id=2160
Referenced issues: #4887
Original owner: https://code.google.com/u/asmeurer@gmail.com/
```

asmeurer:
This is not finished yet (I plan to write something in every section).

I'm also not sure yet that I have everything. If you see something that's missing, please let me know.

I've put this in the contributor guide, since a lot of the stuff here is only useful to contributors (like dependencies that are only needed to run certain tests). However, certain bits of this will also be useful to users as well. I'm not [entirely sure](https://github.com/sympy/sympy/issues/5259#issuecomment-995614730) if this is the best way to organize this yet. I want to get the content written, then we can think if there is a better way to organize it (comments welcome).

Fixes https://github.com/sympy/sympy/issues/5259

#### Release Notes

- other
  - Add a documentation guide listing SymPy's optional dependencies.

### DIFF
--- a/doc/src/guides/contributing/dependencies.md
+++ b/doc/src/guides/contributing/dependencies.md
@@ -1,0 +1,222 @@
+# Dependencies
+
+There are several packages that, when installed, can enable certain additional
+SymPy functionality. Most users and contributors will not need to install any
+of the packages mentioned below (except for the hard dependencies), unless
+they intend to use the part(s) of SymPy that can use those packages.
+
+Every dependency listed below can be installed with conda via
+[conda-forge](https://conda-forge.org/), and most can also be installed with
+`pip`.
+
+This page does not list packages which themselves depend on SymPy, only those
+packages that SymPy depends on. An incomplete list of packages that depend on
+SymPy can be found on the [main SymPy
+webpage](https://www.sympy.org/en/index.html), and a more complete list can be
+found on
+[GitHub](https://github.com/sympy/sympy/network/dependents?dependent_type=PACKAGE)
+or [libraries.io](https://libraries.io/pypi/sympy/dependents).
+
+## Hard Dependencies
+
+SymPy only has one hard dependency, which is required for it to work: mpmath.
+
+### mpmath
+
+[mpmath](https://mpmath.org/) is a pure Python package for arbitrary precision
+arithmetic. It is used under the hood whenever SymPy calculates the
+floating-point value of a function, e.g., when using
+[evalf](sympy.core.evalf.EvalfMixin.evalf).
+
+SymPy cannot function with mpmath and will fail to import if it is not
+installed. If you get an error like
+
+```
+ImportError: SymPy now depends on mpmath as an external library. See
+https://docs.sympy.org/latest/install.html#mpmath for more information.
+```
+
+this means that you did not install mpmath correctly. [This
+page](mpmath-install) explains how to install it.
+
+Most methods of installing SymPy, such as the ones outlined in the
+[installation](installation) guide, will install mpmath automatically. You
+typically only need to install mpmath manually if you did not actually install
+SymPy, e.g., if you are developing directly on SymPy in the git repository.
+
+## Optional Dependencies
+
+These dependencies are not required to use SymPy. The vast majority of SymPy
+functions do not require them, however, a few functions such as plotting and
+automatic wrapping of code generated functions require additional dependencies
+to function.
+
+Additionally, as a contributor, when running the SymPy tests, some tests will
+be skipped if a dependency they require is not installed. The [GitHub Actions
+CI](https://github.com/sympy/sympy/actions) which is run on every SymPy pull
+request will automatically install these dependencies in the
+"optional-dependencies" build, but you may wish to install them locally if you
+are working on a part of SymPy that uses them.
+
+### Recommended Optional Dependencies
+
+These dependencies are not required for SymPy to function, but it is
+recommended that all users install them if they can, as they will improve the
+general performance of SymPy.
+
+#### gmpy2
+
+[gmpy2](https://gmpy2.readthedocs.io/en/latest/intro.html) is a Python wrapper
+for the GMP multiple-precision library. It provides large integers that are
+faster than the built-in Python `int`. When gmpy2 is installed, it is used
+automatically by certain core functions that operate on integers, such as the
+[polys](polys-docs). See {ref}`polys-domainsref` for more details.
+
+### Interactive Use
+
+#### IPython
+
+#### Jupyter Notebook
+
+### LaTeX Parsing
+
+#### antlr-python-runtime
+
+#### python-clang
+
+### Logic
+
+#### pycosat
+
+#### pysat
+
+### Plotting
+
+#### matplotlib
+
+#### pyglet
+
+### lambdify
+
+#### numpy
+
+#### scipy
+
+#### cupy
+
+#### tensorflow
+
+#### numexpr
+
+### Code Generation
+
+#### NumPy
+
+#### Cython
+
+#### Aesara
+
+#### pymc3
+
+#### llvmlite
+
+#### Wurlitzer
+
+[Wurlitzer](https://github.com/minrk/wurlitzer) is a Python package that
+allows capturing output from C extensions. It is used by some of the tests in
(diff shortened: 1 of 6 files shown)


doc/src/guides/contributing/dependencies.md
```diff
@@ -0,0 +1,222 @@
...
+
+### mpmath
+
+[mpmath](https://mpmath.org/) is a pure Python package for arbitrary precision
+arithmetic. It is used under the hood whenever SymPy calculates the
+floating-point value of a function, e.g., when using
+[evalf](sympy.core.evalf.EvalfMixin.evalf).
+
+SymPy cannot function with mpmath and will fail to import if it is not
+installed. If you get an error like
```
oscarbenjamin, line 32: cannot function *without* mpmath

asmeurer:
Need to think of a better way to organize this. The headings and subheadings are too hard to distinguish on the rendered page.

bertiewooster:
> The headings and subheadings are too hard to distinguish on the rendered page.

Concur; that's an issue site-wide: I have to carefully visually compare headings to tell an `<h2>` from an `<h3>` and even an `<h4>`. Should we consider changing the styling on those various heading levels? Perhaps give different levels different indenting, more dissimilar font size,...?

asmeurer:
There were discussions about switching to a better theme for the docs (see https://github.com/sympy/sympy/issues/22007). Furo was suggested, but I'm open to any theme that improves things. We should look at how Furo handles this.

doc/src/guides/contributing/dependencies.md
```diff
@@ -0,0 +1,222 @@
...
+faster than the built-in Python `int`. When gmpy2 is installed, it is used
+automatically by certain core functions that operate on integers, such as the
+[polys](polys-docs). See {ref}`polys-domainsref` for more details.
+
+### Interactive Use
+
+#### IPython
+
+#### Jupyter Notebook
+
```
oscargus, line 80: Mention Spyder?

doc/src/guides/contributing/dependencies.md
```diff
@@ -0,0 +1,222 @@
...
+Some tests in `sympy.utilities.tests.test_pickling.py` depend on cloudpickle
+to run. It is not otherwise required for any SymPy function.
+
+#### Sage
+
+
+## Development Dependencies
+
+Typical development on SymPy does not require any additional dependencies
+beyond Python and mpmath.
```
oscargus, line 184: pytest? (Not sure though, I haven't really understood the relation fully...)

doc/src/guides/contributing/dependencies.md
asmeurer: There's no spyder specific code in SymPy itself (that I'm aware of).

oscargus: Is there any Jupyter Notebook-specific?

asmeurer: Yes, we have hooks for MathJax printing in the Jupyter notebook (enabled by default, and customizable with `init_printing()`).

doc/src/guides/contributing/dependencies.md
asmeurer: Added something on pytest. Let me know if that clears it up or if I should add more.

oscarbenjamin: We should add something about pytest somewhere with more detail. Personally I only use pytest for running the sympy tests.

asmeurer:
I built this using Furo. Furo does make it easier to distinguish header levels by using distinct styling for each level.

Screen Shot 2022-01-05 at 4 48 43 PM

However, I feel like this is still too cluttered and hard to read. I'm going to play with other ways of formatting the page.

asmeurer:
I modified the subheadings into bullet points. The document is much easier to read now. I do still have a few level 4 headers in the code generation section, which will be easier to parse once we clean up the theming. 

This PR is now ready to review. Please let me know if I missed any optional dependencies, or any important places where an optional dependency is used in library or test code.

doc/src/guides/contributing/dependencies.md
```diff
@@ -0,0 +1,398 @@
...
+
+These dependencies are not required for SymPy to function, but it is
+recommended that all users install them if they can, as they will improve the
+general performance of SymPy.
+
+- **gmpy2**: [gmpy2](https://gmpy2.readthedocs.io/en/latest/intro.html) is a
+  Python wrapper for the GMP multiple-precision library. It provides large
+  integers that are faster than the built-in Python `int`. When gmpy2 is
+  installed, it is used automatically by certain core functions that operate
+  on integers, such as the [polys](polys-docs). See {ref}`polys-domainsref`
```
oscarbenjamin, line 74: Maybe it's worth clarifying that it is used implicitly by all kinds of things like `collect`, `cancel` etc which are themselves widely used by higher-level things like symbolic integration. Using `gmpy2` can make things a lot faster. Also often the more noticeable speedup is from mpq rather than mpz. Also `gmpy2` is an optional dependency for `mpmath` and can speed up things like `evalf`.

Although only certain parts of the codebase use gmpy2 those parts are very widely used throughout the rest of the codebase and I think it is really not obvious even to experienced users what kinds of operation will be made faster by installing gmpy2.

asmeurer: I added more here. Maybe once we have flint as another such dependency we can have another document showing some of the performance improvements you can achieve by installing these (or if not a docs page, a blog post somewhere).

oscarbenjamin:
Looks good to me

PR #22696, opened 2021-12-17
physics/optics: Updated Medium to inherit from Basic

### LINKED ISSUE #22664: `Medium` not hashable
This shows up as an issue in the LaTeX-printing, but is probably an issue elsewhere as well.

```
from sympy.physics.optics import Medium
Medium('m')
```
gives
```

  File "C:\Users\Oscar\miniconda3\lib\site-packages\IPython\core\formatters.py", line 345, in __call__
    return method()

  File "C:\Users\Oscar\sympy\sympy\core\_print_helpers.py", line 64, in _repr_latex_
    s = latex(self, mode='plain')

  File "C:\Users\Oscar\sympy\sympy\printing\printer.py", line 373, in __call__
    return self.__wrapped__(*args, **kwargs)

  File "C:\Users\Oscar\sympy\sympy\printing\latex.py", line 3011, in latex
    return LatexPrinter(settings).doprint(expr)

  File "C:\Users\Oscar\sympy\sympy\printing\latex.py", line 254, in doprint
    tex = Printer.doprint(self, expr)

  File "C:\Users\Oscar\sympy\sympy\printing\printer.py", line 293, in doprint
    return self._str(self._print(expr))

  File "C:\Users\Oscar\sympy\sympy\printing\printer.py", line 332, in _print
    return printmethod(expr, **kwargs)

  File "C:\Users\Oscar\sympy\sympy\printing\latex.py", line 1569, in _print_Symbol
    if expr in self._settings['symbol_names']:

TypeError: unhashable type: 'Medium'
```

NikhilSDate:

Fixes #22664

Changed `Medium` to inherit from `Basic` instead of `Symbol` and to store its arguments in `args` after checking if any arguments are `None`. This makes `Medium` hashable and also makes `sprepr` work correctly when called on a `Medium` object.

Also changed a condition in the `__new__` method of `Medium` that checks whether its arguments are consistent to not raise a `TypeError` due to `Relational` not having a well-defined truth value.

#### Release Notes

* physics.optics
  * changed Medium to inherit from Basic instead of Symbol


sympy/physics/optics/medium.py
```diff
@@ -69,23 +70,32 @@ class Medium(Symbol):
...
         elif permittivity is not None and permeability is not None:
-            obj._n = c*sqrt(permittivity*permeability)
+            n = c*sqrt(permittivity*permeability)
         elif permittivity is None and permeability is None:
-            obj._permittivity = _e0mksa
-            obj._permeability = _u0mksa
+            permittivity = _e0mksa
+            permeability = _u0mksa
+            n = c*sqrt(permittivity*permeability)
+        args = list(map(sympify, (permittivity, permeability, n)))
```
ThePauliPrinciple, line 93: `_sympify` is prefered over `sympify` (it is stricter and shouldn't return non-Basic types, which `sympify` might)

NikhilSDate: Alright, I'll make the change.

sympy/physics/optics/medium.py
```diff
@@ -69,23 +70,31 @@ class Medium(Symbol):
...
             if permeability is not None and permittivity is None:
-                obj._permittivity = n**2/(c**2*obj._permeability)
+                permittivity = n**2/(c**2*permeability)
             if permittivity is not None and permittivity is not None:
-                if abs(n - c*sqrt(obj._permittivity*obj._permeability)) > 1e-6:
-                   raise ValueError("Values are not consistent.")
+                expr = abs(n - c*sqrt(permittivity*permeability))
+                expr = expr.subs({meter: 1, second: 1})
+                if len(expr.free_symbols) == 0 and expr > 1e-6:
+                    raise ValueError("Values are not consistent.")
```
ThePauliPrinciple, line 85: What if `n is not None`, `permeability is None` and `permittivity is None`? I don't think it is possible to compute the values for `permeability` and `permittivity` in that case (you have two unknowns but one equation), but this is not handled explicitly here. I think it will end up with a `None` in both `permeability` and `permittivity` when it passes them to `_sympify` which will cause an error.

ThePauliPrinciple: Line 77-85 should also be an if-elif block since the options are mutually exclusive.

NikhilSDate: Regarding your first suggestion, in the case you are talking about, how do you suggest this should be handled? Should it immediately raise an rrror? 
Also, the version of lines 77-85 in the original code does not have `if-elif` blocks. I can still change that though.

ThePauliPrinciple: Yes, it should raise an error to tell the user it is using a combination of inputs that can't be used, rather than fail somewhere where the error might not be easy to understand. 

I think the `elif` would be more efficient.

NikhilSDate: Ok. Are there any other changes you suggest I should make in this section of the code? I will make all of them in a single commit to avoid too many commits with small changes in each commit.

ThePauliPrinciple: If `n is None` and only one of `permittivity` and `permeability` `is None` is also not handled.

NikhilSDate: Should that also produce an error since it's impossible to calculate unique values for both unknowns? Or should it use the default value for whichever of `permittivity` and `permeability` is `None` and then calculate `n` using that value?

ThePauliPrinciple: Other than that it looks good.

ThePauliPrinciple: I think an error is better, since it currently would also give an error at some point, just not at a place where a user can easily understand what was done wrong.

NikhilSDate: Does it look ok now? If so, I will resolve this conversation

NikhilSDate:
@ThePauliPrinciple There seems to be a doctest in `Medium` that involves creating a `Medium` object with `n` specified, but `permittivity` and `permeability` not specified. Since I changed the code to raise an error in this case, that test is currently failing. Should this test be modified to provide either the `permittivity` or `permeability` when creating the object?

ThePauliPrinciple:
People might be using it in that way then. I've come across this a couple of times, where it would be convenient if there was something like a "None" that we could use in `args` just so that we can "allign" the values there (to make sure we can maintain the invariance `obj.func(*obj.args)==obj` when some arg is missing before the last one. It would have been easy if `n` were the first argument (since then we can just have either 1 or 3 `args`). I'm not sure how to make this class work with SymPy and be backwards compatible.

oscarbenjamin:
I doubt that something like `Medium` is used much by downstream libraries so I think it's probably fine just to change it.

I'm not sure I understand exactly what the change is though. Why is it not possible to just fix this class without breaking backwards compatibility?

ThePauliPrinciple:
`Medium('m', None, None, 1)` is currently accepted, but we can't store this in `args`. An argument could be made for using default values here, however, how do we choose the default values if `n=2`, then `permittivity*permeability=2`, we one equation and two unknowns.

Maybe a `Dummy` variable could be created for `permittivity` and then `permeability` is an expression which solves the above equation (or the other way around).

Originally, only `name` was stored in `args` : `obj = super().__new__(cls, name)`

oscarbenjamin:
> An argument could be made for using default values here, however, how do we choose the default values if `n=2`, then `permittivity*permeability=2`, we have one equation and two unknowns.

Is it actually useful to define a `Medium` with unknown permittivity and permeability?

ThePauliPrinciple:
I don't think so personally (hence my original suggestion to raise an error in this case), but the doctest tests for this explicitly:
```python
        >>> from sympy.physics.optics import Medium
        >>> m = Medium('m')
        >>> m.speed
        299792458*meter/second
        >>> m2 = Medium('m2', n=1)
        >>> m.speed == m2.speed
```

oscarbenjamin:
> but the doctest tests for this explicitly:

All kinds of bogus things are added as tests or doctests by contributors who think that something is nice even if it has no actual use. I don't see why that can't just be changed.

NikhilSDate:
Are there any specific disadvantages to using `Dummy` variables for `permittivity` and `permeability` when `n` is not `None` and both `permittivity` and `permeability` are `None`? I think that's actually a nice solution since it will fix the failing tests and maintain the `obj.func(*obj.args)==obj` invariance without having to break backwards compatibility. 

@oscarbenjamin I guess a potential use for a `Medium` whose permittivity and permeability are unknown might be when we are doing simple ray optics, where we are only interested in the refractive index. The following test in `test_utils.py` is an example of such a situation (calculating the critical angle between two media)
https://github.com/sympy/sympy/blob/6981dee608382a0bf426dd8d05bf90a809fe8718/sympy/physics/optics/tests/test_utils.py#L126-L129

ThePauliPrinciple:
That is a good point. @NikhilSDate I think you can just remove that part of the doctest, unless you want to make an attempt at still supporting this option somehow.

NikhilSDate:
> If you use a Dummy variable, then two different media with undefined permittivity and permeability and the same n are always not equal. You could also use some kind of Symbol, but then they would always be equal.

They would currently still be equal since the `__eq__` of `Medium` only checks if refractive indices are equal. Since there are some tests that depend on this behavior, I haven't changed `__eq__`.

> You could try something like permittivity = sqrt(n)*default_permittivity and permeability = sqrt(n)*default_permeability.

Yes, this seems like a good suggestion. I'll change it to this.

oscarbenjamin:
Looking at it the Medium class is just a badly designed mess.

I'm not sure that the changes here really help much though. The class would be hashable if it's `__eq__` method was removed which I think is the proper fix.

ThePauliPrinciple:
I was considering to suggest to remove the `__eq__` method too. Do note that without the other changes, it would only store a `str` (not a `Str`) in `args` and nothing else, so those changes are needed too.

Any tests/code that rely on `__eq__` comparing `n` only, should probably just compare that property (e.g. `m1.refractive_index == m2.refractive_index` directly rather than using `m1 == m2` (which should be a comparison for structural equality, not mathematical equality).

NikhilSDate:
Removing __eq__ would probably make some tests fail which require __eq__ to only compare refractive indices.

I made these changes since it was suggested in #22664 that Medium should inherit from Expr or Basic instead of Symbol. Since Medium objects can't be added or multiplied, I went with Basic. This also allows Medium to store its arguments in arts, which makes print_latex and srepr work correctly with Medium. Maybe the class design issues could be fixed in another PR since that might also break backwards compatibility? The main point of this PR is to just make Medium inherit from Basic.

oscarbenjamin:
The proper way for an Atom to influence the behaviour of `__eq__` is by defining `_hashable_content` which is then used by both `Basic.__eq__` and `Basic.__hash__`.

NikhilSDate:
> The proper way for an Atom to influence the behaviour of `__eq__` is by defining `_hashable_content` which is then used by both `Basic.__eq__` and `Basic.__hash__`.

I can go ahead and change all the tests that rely on `__eq__` comparing `n` to directly compare the refractive index and then just remove `__eq__`. In this case `_hashable_content` is just `_args`, so it `__eq__` would then directly compare arguments. Backward compatibility is the only issue with this change. Should I go ahead then?

ThePauliPrinciple:
> The proper way for an Atom to influence the behaviour of `__eq__` is by defining `_hashable_content` which is then used by both `Basic.__eq__` and `Basic.__hash__`.

So then your suggestion would be to keep the class as-is, but remove `__eq__` and add
```python
def _hashable_content(self):
    return self._n
```

NikhilSDate:
I still think it is better to not have `Medium` inherit from `Symbol`. From an OOP point of view, I guess it makes more sense for Medium to inherit directly from `Basic`, since `Medium` doesn't really pass the is-a test with `Symbol`. Also, only changing hashable content would mean that `srepr` and `print_latex` would still not work correctly, since only the name would be stored as an argument. Any thoughts?

oscarbenjamin:
At the moment Medium is not an actual atom because it gives one arg to `Basic.__new__`. It could be made into an atom though. Probably it shouldn't be a subclass of Symbol but if it is to be a proper non-atom Basic subclass then it should have well-defined positional args.

I think the best way to get the original intent would be to have different classes with `Medium` as a common base class holding the bulk of the methods. One class would just take `n` as an argument. Another would take permittivity and permeability as its args. The `Medium.__new__` constructor could dispatch to the different subclasses.

NikhilSDate:
Should Medium even be an Atom, since it contains subexpressions like permittivity and permeability?

oscarbenjamin:
To be clear my suggestion about having different classes was that they would not be atoms.

ThePauliPrinciple:
> Also, what would be the advantage of doing it with subclasses over the current approach in this PR?

So that you would not have to choose default values based on some arbitrary choice. (like sqrt(n) that I suggested before)

NikhilSDate:
Hmm I'm not exactly sure then what you mean. So are you suggesting that `Medium` should inherit from `Basic`, and the subclasses should then inherit from `Medium`. And are also suggesting that there should be separate subclasses for cases where `n` is not defined but `permittivity` and `permeability` are defined, and where `permittivity` and `permeability` are defined but n is not defined? I don't think it would be a good idea to have a separate subclass in the second case, since `n` can be calculated from the `permittivity` and `permeability`

oscarbenjamin:
It would look something like this:
```python
class Medium(Basic):
    def __new__(cls, name, permittivity=None, permeability=None, n=None):
        if n is not None and permittivity is None and permeability is None:
            return MediumN(name, _sympify(n))
        elif permittivity is not None and permeability is not None:
            return MediumPP(name, _sympify(permittivity), _sympify(permeability))
        else:
            raise Error

class MediumN(Medium): 
    @property
    def n(self):
        return self.args[1]
    
class MediumPP(Medium):
    @property
    def permittivity(self):
        return self.args[1]
    
    @property
    def permeability(self):
        return self.args[2]
    
    @property
    def n(self):
        return c*sqrt(self.permittivity*self.permeability)   
```
Most methods then go in the Medium class but they just use the `.n` etc properties. The methods that only make sense for MediumPP could just be defined in MediumPP or could raise when called on MediumN. The two classes can have different implementations for `speed`.

> lot of subclasses might be necessary corresponsing to the different underspecified cases.

How many underspecified cases are there? Is anything more than the two cases I just spelled out actually needed for anything?

ThePauliPrinciple:
`MediumPP` should be able to easily handle cases where `n` and one of the others is given too, since it is possible to compute the last value. The odd one is when only `n` is specified. If only `permittivity` or `permeability` is given it should give an error

NikhilSDate:
@oscarbenjamin Thanks for the code. I see what you are suggesting now. 
I've got another idea for how the class hierarchy could work. We could have `MediumN` directly inheriting from `Basic`, with its `__new__` only accepting `name` and `n`.  Then, we could have `Medium` inheriting from `MediumN`, and defining `permittivity` and `permeability`. All the methods that only require `n` would go in `MediumN`, and everything else would go in `Medium`. This would result in one less class. I also think this is a good idea since, in this case, `Medium` extends the behavior of `MediumN` by providing the additional `permittivity` and `permeability` attributes. Also, all the tests that only specify `n` could then be changed to use `MediumN`.
@oscarbenjamin @ThePauliPrinciple What do you think of doing it this way?

ThePauliPrinciple:
> Also, all the tests that only specify `n` could then be changed to use `MediumN`.

This isn't needed with the solution proposed by @oscarbenjamin so I think that is better.

NikhilSDate:
Ok I see. For me, the only issue with @oscarbenjamin's solution is that the hierarchy kind of runs opposite to how I would intuitively understand this situation. I would look at `Medium` extending the functionality provided by `MediumN` in that `Medium` can provide all the functionality provided by `MediumN`, but `MediumN` cannot do everything `Medium` does. Do you suggest any way to get around this?

ThePauliPrinciple:
I don't understand the problem with the hierarchy, `Medium` provides all functionality needed for `MediumN` and `MediumPP`, while those classes provide their respective specific functionality. The user still can use `Medium` since it dispatches to `MediumN` or `MediumPP` depending on the input.

NikhilSDate:
Alright, I'll try implementing @oscarbenjamin's idea and see how it goes.

NikhilSDate:
@oscarbenjamin I have changed the class structure as you suggested. Is this in line with what you had in mind?

sympy/core/tests/test_args.py
```diff
@@ -4919,6 +4919,14 @@ def test_sympy__physics__optics__medium__Medium():
     from sympy.physics.optics import Medium
     assert _test_args(Medium('m'))
 
+@SKIP('should not be instantiated directly (only indirectly through instantiating Medium')
+def test_sympy__physics__optics__medium__MediumN():
+    pass
```
oscarbenjamin, line 4924: You can still add a test that instantiates a MediumN indirectly.

sympy/physics/optics/medium.py
```diff
@@ -69,50 +71,34 @@ class Medium(Symbol):
...
-                   raise ValueError("Values are not consistent.")
+                permeability = n**2/(c**2*permittivity)
+                return MediumPP(name, _sympify(permittivity), _sympify(permeability))
+            elif permeability is not None and permittivity is None:
+                permittivity = n**2/(c**2*permeability)
+                return MediumPP(name, _sympify(permittivity), _sympify(permeability))
+            elif permittivity is not None and permittivity is not None:
+                expr = abs(n - c * sqrt(permittivity * permeability))
+                expr = expr.subs({meter: 1, second: 1})
+                if len(expr.free_symbols) == 0 and expr > 1e-6:
```
oscarbenjamin, line 87: I think that passing all 3 arguments should just always be an error. This code for checking consistency is unreliable and shouldn't be attempted

sympy/physics/optics/medium.py
```diff
@@ -69,50 +71,34 @@ class Medium(Symbol):
...
-        >>> from sympy.physics.optics import Medium
-        >>> m = Medium('m')
-        >>> m.intrinsic_impedance
-        149896229*pi*kilogram*meter**2/(1250000*ampere**2*second**3)
-
-        """
-        return sqrt(self._permeability/self._permittivity)
+            permittivity = _e0mksa
+            permeability = _u0mksa
+            return MediumPP(name, _sympify(permittivity), _sympify(permeability))
```
oscarbenjamin, line 101: These don't need to be sympified.

sympy/physics/optics/medium.py
```diff
@@ -152,6 +135,73 @@ def refractive_index(self):
...
+        """
+        Compares based on refractive index of the medium.
+        """
+        return self.refractive_index < other.refractive_index
+
+    def __gt__(self, other):
+        return not self < other
+
+    def __eq__(self, other):
+        return self.refractive_index == other.refractive_index
```
oscarbenjamin, line 148: The `__eq__` method should be removed. Proper `Basic` subclasses shouldn't define `__eq__`.

sympy/physics/optics/medium.py
```diff
@@ -152,6 +135,73 @@ def refractive_index(self):
...
+
+    def __ne__(self, other):
+        return not self == other
+
+
+class MediumN(Medium):
+
+    def __new__(cls, name, n):
+        obj = super(Medium, cls).__new__(cls, name, n)
+        obj.name = name
```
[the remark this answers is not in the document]
oscarbenjamin: Also there's no reason to store `_n` because there is a property method for `n`

sympy/physics/optics/medium.py
```diff
@@ -152,6 +135,73 @@ def refractive_index(self):
...
+        obj._n = n
+        return obj
+
+    @property
+    def n(self):
+        return self._n
+
+    def __str__(self):
+        return type(self).__name__ + ': ' + sstr([self.n])
```
oscarbenjamin, line 167: This `__str__` method should be removed. The printers control the printed display of Basic

sympy/physics/optics/medium.py
```diff
@@ -152,6 +135,73 @@ def refractive_index(self):
...
+
+
+class MediumPP(Medium):
+
+    def __new__(cls, name, permittivity, permeability):
+        obj = super(Medium, cls).__new__(cls, name, permittivity, permeability)
+        obj.name = name
+        obj._permittivity = permittivity
+        obj._permeability = permeability
```
oscarbenjamin, line 176: These should all be accessed by property-methods

sympy/physics/optics/medium.py
```diff
@@ -69,50 +71,34 @@ class Medium(Symbol):
...
+                return MediumPP(name, _sympify(permittivity), _sympify(permeability))
+            else:
+                return MediumN(name, _sympify(n))
         elif permittivity is not None and permeability is not None:
-            obj._n = c*sqrt(permittivity*permeability)
+            return MediumPP(name, _sympify(permittivity), _sympify(permeability))
+        elif permittivity is not None and permeability is None:
+            raise ValueError("At least one of n and permeability must not be None if permittivity is not None")
+        elif permittivity is None and permeability is not None:
+            raise ValueError("At least one of n and permittivity must not be None if permeability is not None")
```
oscarbenjamin, line 97: I'm pretty sure that the logic of this function can be written in a simpler way.

sympy/core/tests/test_args.py
NikhilSDate: The only issue with directly instantiating `MediumN` is that it does not sympify its arguments (since `Medium`'s `__new__` sympifies arguments before passing them when creating the `MediumN` object. So, the arguments will have to be sympified first.

sympy/physics/optics/medium.py
NikhilSDate: That will cause tests to fail. Should `_hashable_content` be set to just `refractive_index` then?

sympy/core/tests/test_args.py
oscarbenjamin: I said *indirectly*

sympy/physics/optics/medium.py
oscarbenjamin: No the tests should be changed.

sympy/physics/optics/medium.py
```diff
@@ -152,6 +131,62 @@ def refractive_index(self):
...
+        """
+        Compares based on refractive index of the medium.
+        """
+        return self.refractive_index < other.refractive_index
+
+    def __gt__(self, other):
+        return not self < other
+
+    def __ne__(self, other):
+        return not self == other
```
oscarbenjamin, line 144: I would just get rid of these. The `__ne__` method shouldn't be needed because Basic already defines it. The `__lt__` and `__gt__` methods are misguided and incorrect e.g. `__gt__` should be the negation of `self <= other` but that needs `__le__` which isn't defined. If someone wants to compare two `Medium` by their refractive index then they should just do that explicitly.

sympy/physics/optics/medium.py
```diff
@@ -152,6 +131,62 @@ def refractive_index(self):
...
+
+
+class MediumN(Medium):
+
+    def __new__(cls, name, n):
```
oscarbenjamin, line 149: At least a short class docstring could explain what this class is and show two to create it (by calling `Medium`)

oscarbenjamin: Same goes for `MediumPP`.

sympy/physics/optics/medium.py
NikhilSDate: Yes, that makes sense. I'll remove those methods.

oscarbenjamin:
Looks good to me.

@ThePauliPrinciple does this look good to you?

sympy/physics/optics/medium.py
```diff
@@ -69,50 +70,31 @@ class Medium(Symbol):
...
+
         if n is not None:
             if permittivity is not None and permeability is None:
-                obj._permeability = n**2/(c**2*obj._permittivity)
-            if permeability is not None and permittivity is None:
-                obj._permittivity = n**2/(c**2*obj._permeability)
-            if permittivity is not None and permittivity is not None:
-                if abs(n - c*sqrt(obj._permittivity*obj._permeability)) > 1e-6:
-                   raise ValueError("Values are not consistent.")
+                permeability = n**2/(c**2*permittivity)
```
ThePauliPrinciple, line 78: I think it is better to `_sympify` before any calculation, e.g. if a custom object has `_sympy_` defined, but not a `__pow__`/`__mul__`, it wouldn't work with this class.

ThePauliPrinciple:
> Looks good to me.
> 
> @ThePauliPrinciple does this look good to you?

Other than the minor comment on `_sympify`, this looks good now. If it is a problem somehow to change that I'm also fine with accepting it as is.

sympy/physics/optics/medium.py
NikhilSDate: Is something like this done elsewhere in the codebase? I'm struggling to think of a situation where permittivity, permeability, or n would have `_sympy_` defined by not `__pow__/__mul__`, since permittivity, permeability, and n should be either scalars or expressions that can be evaluated to scalars. I don't have any problem making the change, but I'm just not sure if it's necessary.

ThePauliPrinciple: I would expect it to not have a sensible `__mul__` in ALL cases it has `_sympy_` defined, since `_sympy_` is a a method a user made class that is not a subclass of `Basic` can have to indicate how SymPy should work with it (or more precisely, how to convert that non-SymPy object to an equivalent SymPy object). It should also be possible to do this for user-classes by modifying `converter` in `core.sympify`. (which is also handled if that object is given to `_sympify`).

NikhilSDate: OK I see. I'll make the change then. I assume there is not a significant performance overhead to sympifying everything before the calculations, since sympify will return any sympy object as it is. Am I right here?

NikhilSDate:
I have made the change suggested by @ThePauliPrinciple. I think this PR is ready to merge now.

NikhilSDate:
ping @oscarbenjamin for merging.

sympy/physics/optics/medium.py
oscarbenjamin: It's probably best just to move all `_sympify` calls to the top of the function like:
```
n = _sympify(n) if n is not None else n
```

oscarbenjamin:
Looks good. Merging. Thanks.

NikhilSDate:
Thanks @oscarbenjamin @ThePauliPrinciple for the reviews.

oscargus:
Thanks @NikhilSDate !

PR #17822, opened 2019-10-30
geometry: Adding a module for Hyperbola

KaLiMaLi555:
#### References to other Issues or PRs

Fixes #17725 
https://github.com/sympy/sympy/pull/7709

#### Brief description of what is fixed or changed

The module for hyperbola contains
`center`, `hradius`, `vradius`, `eccentricity`, `focus distance` and many other attributes.

```
from sympy import Hyperbola, Point, Rational
h1 = Hyperbola(Point(0, 0), 5, 1)
h1.hradius, h1.vradius
(5, 1)
h2 = Hyperbola(Point(3, 1), hradius=3, eccentricity=Rational(5, 4))
h2
Hyperbola(Point2D(3, 1), 3, 9/4)

```
#### Other comments
TODO
`Rectangular Hyperbola`, `encloses_point`, `apoapsis`, `periapsis`, `is_tangent`, `normal_lines`, `tangent_lines`, `reflect`, `rotate`

#### Release Notes

- geometry
  -  Adding a module for hyperbola

PR #7709, opened 2014-07-08
Hyperbola class

akshayah3:

PR #22698, opened 2021-12-17
Implemented basic structure and functionality for hyperbola.py

anutosh491:
When I found out couple months back that the geometry module covered all conics except the hyperbola , I realized it's long time we get this in ! Hence I have implemented a basic structure and all important functionality which deals with the Hyperbola conic . 

What all have I covered ??

-  Equation of conic

-  minor/major axis, centre , vradius, hradius and eccentricity

- foci and focus distance

- equation of auxillary and director circle

-  latus rectum, arbitary point, evolute, a method to check whether hyperbola encloses a point 

What all could be implemented according to me??

- [x] Asymtotes (method to return equation)
- [x] equation of particular branch(left , right) sent as parameters in equation method()
- [ ] Equation of tangent/normal 
- [x] Rotate, scale and related transformations ?
- [ ] Rectangular Hyperbola Class to be used whenever `hradius` is equal to `vradius` ( basic implementation of the standard rectangular Hyperbola equation of the form `x*y = c**2`)
- [x] test_hyperbola.py (I'll be pushing a basic structure for this some time soon following ` test_parabola.py` and` test_ellipse.py`)
- [ ] For two Hyperbola having same centre, vradius , hradius we can't compulsorily say that they are the same conics . They can be conjugate to each other i.e (`x**2/a - y**2/b = 1` and `y**2/b - x**2/a = 1`) .The second hyperbola is called the conjugate hyperbola . Hence we would need a method /class for properties of the conjugate one , producing a conjugate hyperbola is basically changing the axis and orientation of the original one 
- [ ] Other suggestions

Probably could open an issue with the task list above as not everything above is too tough to be implemented by new contributors and also would help us keep track of what is done !

#### Other comments
I have my doubts regarding my `encloses_point` method . What I've implemented (shown at the end) should work fine even without the `if len(self.foci) == 2` block and the other part which is 
```
        else:
            test = self.radius - self.center.distance(p)

        return fuzzy_bool(test.is_negative)
```
```
            x, y = symbols('x y')
            expr = self.equation() + 1
            syms = list(self.equation().free_symbols)
            expr = expr.xreplace({syms[0]: x, syms[1]: y})
            test = expr.subs([(x, p.x), (y, p.y)])
            if test >= S.One:
                return False
            else:
                return True
```
#### Release Notes

* geometry
  * Implemented basic structure and functionality for hyperbola.py

### DIFF
--- a/sympy/geometry/hyperbola.py
+++ b/sympy/geometry/hyperbola.py
@@ -18,7 +18,7 @@ from .entity import GeometryEntity, GeometrySet
 from .exceptions import GeometryError
 from .ellipse import Circle, Ellipse
 from .line import Line, Segment, Ray2D, Segment2D, Line2D, LinearEntity3D
-from .point import Point
+from .point import Point, Point2D, Point3D
 from .polygon import Polygon
 from sympy.solvers import solve
 from sympy.utilities.misc import filldedent, func_name
@@ -771,7 +771,7 @@ class Hyperbola(GeometrySet):
         sympy.geometry.entity.GeometryEntity
         Examples
         ========
-        >>> from sympy import Hyperbola, Point, Line, Point2D
+        >>> from sympy import Hyperbola, Point, Line, Point2D, Ellipse
         >>> h = Hyperbola(Point(0, 0), 5, 7)
         >>> h.intersection(Point(0, 0))
         []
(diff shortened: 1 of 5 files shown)


sympy/geometry/hyperbola.py
```diff
@@ -0,0 +1,703 @@
...
+
+    """
+
+    def __contains__(self, o):
+        if isinstance(o, Point):
+            x = Dummy('x', real=True)
+            y = Dummy('y', real=True)
+
+            res = self.equation(x, y).subs({x: o.x, y: o.y})
+            return trigsimp(simplify(res)) is S.Zero
```
smichr, line 92: This should probably be `trigsimp(simplify(res)).is_zero` unless `contains` cannot return None. Otherwise if it must be T/F then an Undecidable (sp?) error should be raised if it doesn't resolve.

anutosh491:
I've added a branches parameter in my latest commit which would work for all hyperbola which are constructed without slope though I currently am not sure what would be the best method to get branches of a hyperbola constructed with slope 
```
>>> h = Hyperbola((0, 0), 3, 2)
>>> h.equation()
x**2/9 - y**2/4 - 1
>>> h.equation(branch = 'left')
x + 3*sqrt(y**2 + 4)/2
>>> h.equation(branch = 'right')
x - 3*sqrt(y**2 + 4)/2
```

anutosh491:
Ping @oscarbenjamin @oscargus @asmeurer @sylee957 for review. Hello All.  Implementing a new py file , I would be glad to know your views and suggestions for the same. I've mentioned what has been done or what all I have been trying to approach in the OP !

jksuom:
There is some previous work on hyperbola: https://github.com/sympy/sympy/pull/7709, https://github.com/sympy/sympy/pull/17822. Some parts of those could probably be included.

anutosh491:
> There is some previous work on hyperbola: #7709, #17822. Some parts of those could probably be included.

Thanks for mentioning this , I was completely unaware of any previous work. Will go through it ! I plan to submit code for all basic functionalities and a `basic test_hyperbola.py` file soon. All suggestions are welcome

sympy/geometry/hyperbola.py
```diff
@@ -0,0 +1,986 @@
...
+
+        return GeometryEntity.__new__(cls, center, hradius, vradius, **kwargs)
+
+    def __contains__(self, o):
+        if isinstance(o, Point):
+            x = Dummy('x', real=True)
+            y = Dummy('y', real=True)
+
+            res = self.equation(x, y).subs({x: o.x, y: o.y})
+            return trigsimp(simplify(res)).is_zero
```
oscargus, line 135: It is possible to use something less computationally demanding here? Like cancel? 

Also, is trigsimp really needed? Is the idea that the point is described using trigonometric functions?

sympy/geometry/hyperbola.py
```diff
@@ -0,0 +1,986 @@
...
+        a
+
+        >>> m = Symbol('m')
+        >>> M = m + 1
+        >>> Hyperbola(p1, m, M).minor
+        m
+
+        """
+        ab = self.args[1:3]
+        if len(ab) == 1:
```
oscargus, line 553: I do not think this case is tested. (And I think the line above will give an error for this case.)

sympy/geometry/hyperbola.py
```diff
@@ -0,0 +1,986 @@
...
+        b
+
+        >>> m = Symbol('m')
+        >>> M = m + 1
+        >>> Hyperbola(p1, m, M).major
+        m + 1
+
+        """
+        ab = self.args[1:3]
+        if len(ab) == 1:
```
oscargus, line 506: Same here.

sympy/geometry/hyperbola.py
anutosh491: Yeah currently that won't be required , so we could get rid of that but we would be using it whenever the rectangular hyperbola class is implemented !

sympy/geometry/hyperbola.py
anutosh491: Yeah true , I used `ellipse.py` as an inspiration for this part but giving a thought , I realize that trigsimp isn't really required !

anutosh491:
ping @NikhilSDate @praneethratna. The code being a lot , I would be glad if you'll could give it a review if you'll are familiar with the basic mathematics related to the `Hyperbola` conic

NikhilSDate:
I'm not extremely familiar with the math involved, but if I'm not wrong, currently only a parabola opening to the left and the right is implemented. Have you considered implementing a parabola that opens up and down?

faze-geek:
> I'm not extremely familiar with the math involved, but if I'm not wrong, currently only a parabola opening to the left and the right is implemented. Have you considered implementing a parabola that opens up and down?

Yeah I'm familiar with conic modules .Are you talking about parabola or hyperbola .Check out #22792 ,now arbitrary parabolas can be designed.

NikhilSDate:
Oops! I meant hyperbola. That was a typo.

PR #22700, opened 2021-12-17
Run mypy on a separate build on GitHub Actions

asmeurer:
This makes it so that the other tests still run when mypy fails. Although mypy
is a "code quality" check, the code-quality checks on CI cause other builds to
not run, so they should only be things that are so basic that other tests need
not run until they are fixed. Typing errors are often complicated to fix so
they don't generally fall into this category.

Splitting it like this also makes it easy to make it not required to pass if a
mypy update introduces new errors to every pull request, until those are
fixed.

#### Release Notes

NO ENTRY

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -25,16 +25,13 @@ jobs:
         with:
           python-version: '3.9'
       - run: python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


asmeurer:
We will also need to go in to the branch protection settings and make the mypy build required once this is merged (assuming that it is now passing).

PR #22701, opened 2021-12-18
DOC: replace ** -> ^ in LaTeX equation for atan2()

mgeier:

`**` is used in Python for exponentiation, but LaTeX equations need to use `^` instead.

#### Release Notes

NO ENTRY


smichr:
Thanks

PR #22707, opened 2021-12-18
quality: fix type hints and remove type: ignore

### LINKED ISSUE #22697: mypy code quality test failing
The code quality test seems to be currently failing for all pull requests. This does not seem to have anything to do with the pull requests themselves. In particular, the `mypy` run is failing with the following message:
```
sympy/tensor/array/expressions/conv_array_to_matrix.py:288: error: No overload variant of "__getitem__" of "tuple" matches argument type "List[Any]"
sympy/tensor/array/expressions/conv_array_to_matrix.py:288: note: Possible overload variants:
sympy/tensor/array/expressions/conv_array_to_matrix.py:288: note:     def __getitem__(self, int) -> Basic
sympy/tensor/array/expressions/conv_array_to_matrix.py:288: note:     def __getitem__(self, slice) -> Tuple[Basic, ...]
sympy/tensor/array/expressions/conv_array_to_matrix.py:290: error: No overload variant of "__getitem__" of "tuple" matches argument type "List[Any]"
sympy/tensor/array/expressions/conv_array_to_matrix.py:290: note: Possible overload variants:
sympy/tensor/array/expressions/conv_array_to_matrix.py:290: note:     def __getitem__(self, int) -> Basic
sympy/tensor/array/expressions/conv_array_to_matrix.py:290: note:     def __getitem__(self, slice) -> Tuple[Basic, ...]
sympy/tensor/array/expressions/conv_array_to_matrix.py:494: error: Incompatible types in assignment (expression has type "List[Tuple[Any, ...]]", variable has type "Dict[Any, Tuple[Any, ...]]")
sympy/tensor/array/expressions/conv_array_to_matrix.py:500: error: Incompatible types in assignment (expression has type "List[Any]", variable has type "Dict[Any, Tuple[Any, ...]]")
sympy/tensor/array/expressions/arrayexpr_derivatives.py:26: error: "Expr" has no attribute "shape"
sympy/tensor/array/expressions/arrayexpr_derivatives.py:66: error: "Expr" has no attribute "shape"
sympy/tensor/array/expressions/arrayexpr_derivatives.py:77: error: "Expr" has no attribute "shape"
sympy/tensor/array/expressions/arrayexpr_derivatives.py:82: error: "Expr" has no attribute "shape"
Found 8 errors in 2 files (checked 1455 source files)
```
This should probably be fixed soon as it is also preventing other checks from running.

oscarbenjamin:

Fixes #22697

Remove a lot of `# type: ignore` that aren't needed with latest mypy. Fix some inaccurate type annotations.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/tensor/array/expressions/array_expressions.py
+++ b/sympy/tensor/array/expressions/array_expressions.py
@@ -33,7 +33,7 @@ from sympy.core.sympify import _sympify
 
 
 class _ArrayExpr(Expr):
-    pass
+    shape : tTuple[Expr, ...]
 
 
 class ArraySymbol(_ArrayExpr):
@@ -1571,7 +1571,7 @@ class _EditArrayContraction:
                 args_with_ind[arg_pos].indices[rel_pos] = i
         self.args_with_ind: List[_ArgE] = args_with_ind
         self.number_of_contraction_indices: int = len(contraction_indices)
-        self._track_permutation: Optional[List[List[int]]] = None
+        self._track_permutation: List[List[int]] = None # type: ignore
 
         mapping = _get_mapping_from_subranks(base_array.subranks)
 
--- a/sympy/tensor/array/expressions/array_expressions.py
+++ b/sympy/tensor/array/expressions/array_expressions.py
@@ -1571,7 +1571,7 @@ class _EditArrayContraction:
                 args_with_ind[arg_pos].indices[rel_pos] = i
         self.args_with_ind: List[_ArgE] = args_with_ind
         self.number_of_contraction_indices: int = len(contraction_indices)
-        self._track_permutation: List[List[int]] = None # type: ignore
+        self._track_permutation: Optional[List[List[int]]] = None
 
         mapping = _get_mapping_from_subranks(base_array.subranks)
 
--- a/sympy/tensor/array/expressions/conv_array_to_matrix.py
+++ b/sympy/tensor/array/expressions/conv_array_to_matrix.py
@@ -283,11 +283,10 @@ def _(expr: PermuteDims):
             p2 = permuted[2*i+1]
             if p1 // 2 != p2 // 2:
                 return _permute_dims(mat_mul_lines, permutation)
-            posi = p1 // 2
             if p1 > p2:
-                args_array[i] = _a2m_transpose(mat_mul_lines.args[posi])
+                args_array[i] = _a2m_transpose(mat_mul_lines.args[p1 // 2])
             else:
-                args_array[i] = mat_mul_lines.args[posi]
+                args_array[i] = mat_mul_lines.args[p1 // 2]
         return _a2m_tensor_product(*args_array)
     else:
         return expr
@@ -956,7 +955,7 @@ def _array_contraction_to_diagonal_multiple_identity(expr: ArrayContraction):
                 flag = True
                 break
             free_pos = list(range(*editor.get_absolute_free_range(id1)))[0]
-            editor._track_permutation[-1].append(free_pos)
+            editor._track_permutation[-1].append(free_pos) # type: ignore
             id1.element = None
             flag = False
             break
@@ -970,7 +969,7 @@ def _array_contraction_to_diagonal_multiple_identity(expr: ArrayContraction):
     for j, e in enumerate(editor.args_with_ind):
         if e.element is None:
             editor._track_permutation[j] = None # type: ignore
-    editor._track_permutation = [i for i in editor._track_permutation if i is not None]
+    editor._track_permutation = [i for i in editor._track_permutation if i is not None] # type: ignore
     # Renumber permutation array form in order to deal with deleted positions:
     remap = {e: i for i, e in enumerate(sorted({k for j in editor._track_permutation for k in j}))}
     editor._track_permutation = [[remap[j] for j in i] for i in editor._track_permutation]
--- a/sympy/tensor/array/expressions/conv_array_to_matrix.py
+++ b/sympy/tensor/array/expressions/conv_array_to_matrix.py
@@ -172,7 +172,7 @@ def _array2matrix(expr):
     return expr
 
 
-@_array2matrix.register(ZeroArray) # type: ignore
+@_array2matrix.register(ZeroArray)
 def _(expr: ZeroArray):
     if get_rank(expr) == 2:
         return ZeroMatrix(*expr.shape)
@@ -180,12 +180,12 @@ def _(expr: ZeroArray):
         return expr
 
 
-@_array2matrix.register(ArrayTensorProduct) # type: ignore
+@_array2matrix.register(ArrayTensorProduct)
 def _(expr: ArrayTensorProduct):
     return _a2m_tensor_product(*[_array2matrix(arg) for arg in expr.args])
 
 
-@_array2matrix.register(ArrayContraction) # type: ignore
+@_array2matrix.register(ArrayContraction)
 def _(expr: ArrayContraction):
     expr = expr.flatten_contraction_of_diagonal()
     expr = identify_removable_identity_matrices(expr)
@@ -226,7 +226,7 @@ def _(expr: ArrayContraction):
             return _array_contraction(ret, *expr.contraction_indices)
 
 
-@_array2matrix.register(ArrayDiagonal) # type: ignore
+@_array2matrix.register(ArrayDiagonal)
 def _(expr: ArrayDiagonal):
     pexpr = _array_diagonal(_array2matrix(expr.expr), *expr.diagonal_indices)
     pexpr = identify_hadamard_products(pexpr)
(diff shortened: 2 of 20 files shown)


oscarbenjamin:
So on master we have:
```console
$ git grep 'type: ignore' | wc -l
560
```
This PR gives:
```console
$ git grep 'type: ignore' | wc -l
141
```
So this removes the bulk of the `# type: ignore` comments.

oscarbenjamin:
This is ready for review

sympy/tensor/array/expressions/array_expressions.py
```diff
@@ -1571,7 +1571,7 @@ def __init__(self, base_array: typing.Union[ArrayContraction, ArrayDiagonal, Arr
                 args_with_ind[arg_pos].indices[rel_pos] = i
         self.args_with_ind: List[_ArgE] = args_with_ind
         self.number_of_contraction_indices: int = len(contraction_indices)
-        self._track_permutation: Optional[List[List[int]]] = None
+        self._track_permutation: List[List[int]] = None # type: ignore
```
ThePauliPrinciple, line 1574: Would `List[List[int]] | NoneType` allow the use of `None` here without ignore?

sympy/tensor/array/expressions/conv_array_to_matrix.py
```diff
@@ -283,23 +283,23 @@ def _(expr: PermuteDims):
             p2 = permuted[2*i+1]
             if p1 // 2 != p2 // 2:
                 return _permute_dims(mat_mul_lines, permutation)
-            pos = p1 // 2
+            posi = p1 // 2
```
ThePauliPrinciple, line 286: Why store this in a variable at all, it gets used only once due to the if-else (unless it gets moved up)

sympy/tensor/array/expressions/array_expressions.py
oscarbenjamin: Maybe. I'll check...

oscarbenjamin: That gives
```console
$ mypy sympy
sympy/tensor/array/expressions/array_expressions.py:1574: error: X | Y syntax for unions requires Python 3.10
sympy/tensor/array/expressions/array_expressions.py:1574: error: Name "NoneType" is not defined
```

sympy/tensor/array/expressions/conv_array_to_matrix.py
oscarbenjamin: I'm not sure I understand, It gets used in either of two branches below.

sympy/tensor/array/expressions/array_expressions.py
ThePauliPrinciple: Reading the docs, it looks like it is exactly what `Optional` does in the first place: https://docs.python.org/3/library/typing.html#typing.Optional 

Why was that changed? It feels like a step backwards.

sympy/tensor/array/expressions/conv_array_to_matrix.py
ThePauliPrinciple: Right, it gets used in either branch, so only once. It's not exactly a complicated expression, so why not just put `p1 // 2`` directly where `posi` is used?

sympy/tensor/array/expressions/array_expressions.py
oscarbenjamin: I'm not sure exactly how to type this given that before use it will always become non-None. If you have Optional here then mypy will complain every time this is used:
```
$ mypy sympy
sympy/tensor/array/expressions/conv_array_to_matrix.py:959: error: Value of type "Optional[List[List[int]]]" is not indexable
sympy/tensor/array/expressions/conv_array_to_matrix.py:973: error: Item "None" of "Optional[List[List[int]]]" has no attribute "__iter__" (not iterable)
Found 2 errors in 1 file (checked 1455 source files)
```

ThePauliPrinciple: I see, it seems that mypy simply doesn't allow you to do this unless you handle "None" in the code explicitly everywhere even if you know it will never be `None`.

oscarbenjamin: I'll just add back the type: ignore. This class is too complicated for me to figure out how to clean it up. (I don't think it needs to be complicated but the code is really messy)

sympy/tensor/array/expressions/conv_array_to_matrix.py
oscarbenjamin: There can be good reasons for naming something even if it is only used once but since this isn't a particularly good name anyway I'll do that.

sympy/tensor/array/expressions/array_expressions.py
ThePauliPrinciple:
```diff
diff --git a/sympy/tensor/array/expressions/array_expressions.py b/sympy/tensor/array/expressions/array_expressions.py
index 75573cbb32..c42c53babb 100644
--- a/sympy/tensor/array/expressions/array_expressions.py
+++ b/sympy/tensor/array/expressions/array_expressions.py
@@ -1571,7 +1571,7 @@ def __init__(self, base_array: typing.Union[ArrayContraction, ArrayDiagonal, Arr
                 args_with_ind[arg_pos].indices[rel_pos] = i
         self.args_with_ind: List[_ArgE] = args_with_ind
         self.number_of_contraction_indices: int = len(contraction_indices)
-        self._track_permutation: Optional[List[List[int]]] = None
+        self._track_permutation: List[List[int]] = [[]]

         mapping = _get_mapping_from_subranks(base_array.subranks)

@@ -1673,7 +1673,7 @@ def to_array_contraction(self):
         contraction_indices = self.get_contraction_indices()
         expr = _array_contraction(_array_tensor_product(*args), *contraction_indices)
         expr2 = _array_diagonal(expr, *diag_indices_filtered)
-        if self._track_permutation is not None:
+        if any(self._track_permutation):
             permutation2 = _af_invert([j for i in self._track_permutation for j in i])
             expr2 = _permute_dims(expr2, permutation2)

@@ -1755,8 +1755,8 @@ def track_permutation_start(self):
     def track_permutation_merge(self, destination: _ArgE, from_element: _ArgE):
         index_destination = self.args_with_ind.index(destination)
         index_element = self.args_with_ind.index(from_element)
-        self._track_permutation[index_destination].extend(self._track_permutation[index_element]) # type: ignore
-        self._track_permutation.pop(index_element) # type: ignore
+        self._track_permutation[index_destination].extend(self._track_permutation[index_element])
+        self._track_permutation.pop(index_element)

     def get_absolute_free_range(self, arg: _ArgE) -> typing.Tuple[int, int]:
         """
```
It seems to pass `mypy`, `bin/test sympy/tensor`, `bin/doctest/ sympy/tensor`, but haven't done the full tests

oscarbenjamin: There is this which might be affected:
https://github.com/sympy/sympy/blob/c43550c31ef6355c49c4461e90a23e9629291b93/sympy/tensor/array/expressions/array_expressions.py#L1676

ThePauliPrinciple: The reason why it has that is because it uses `self._track_permutation` in a list comprehension, so the `any(self._track_permutation)` check should identify the uninitialized list.

Alternative is to use what mypy seems to imply:
```diff
diff --git a/sympy/tensor/array/expressions/array_expressions.py b/sympy/tensor/array/expressions/array_expressions.py
index 75573cbb32..079a1febe6 100644
--- a/sympy/tensor/array/expressions/array_expressions.py
+++ b/sympy/tensor/array/expressions/array_expressions.py
@@ -1755,8 +1755,9 @@ def track_permutation_start(self):
     def track_permutation_merge(self, destination: _ArgE, from_element: _ArgE):
         index_destination = self.args_with_ind.index(destination)
         index_element = self.args_with_ind.index(from_element)
-        self._track_permutation[index_destination].extend(self._track_permutation[index_element]) # type: ignore
-        self._track_permutation.pop(index_element) # type: ignore
+        if self._track_permutation is not None:
+            self._track_permutation[index_destination].extend(self._track_permutation[index_element])
+            self._track_permutation.pop(index_element)

     def get_absolute_free_range(self, arg: _ArgE) -> typing.Tuple[int, int]:
         """
```

oscarbenjamin: This code is too messy for me to try and make changes to it. I don't really understand what it does and I don't know how well tested it is so I'd rather just leave the `type: ignore` there and let someone who wants to improve this code fix it. At least I don't want to smuggle a bug in with a PR that has a large number of trivial changes.

ThePauliPrinciple: Ok, fair enough.

ThePauliPrinciple:
I don't have any further suggestions. Looks good.

PR #22712, opened 2021-12-19
Fixed limit(x**n, x, -oo) Earlier Giving -oo As Answer Irrespective Of Odd/Even n 

### LINKED ISSUE #8634: limit(x**n, x, -oo) is sometimes wrong
For example,

``` python
>>> n = Symbol('n', integer=True, positive=True)  # could be 3
>>> limit(x**n, x, -oo)
oo
```

while

``` python
>>> limit(x**3, x, -oo)
-oo
```

It looks like it assumes `n` is even, because stating `odd=True` for `n` seems to solve this issue.

faze-geek:

Fixes #8634

A fix which returns output dependent on 'n'( hence varies when n is odd/even) instead of -oo which is the wrong output for `limit(x**n, x, oo) `

#### Release Notes

NO ENTRY


oscarbenjamin:
@faze-geek how should your name be recorded in the AUTHORS file?
https://github.com/sympy/sympy/blob/e0188c79d8eb348333b2cc6968b31e7ab3baad2d/AUTHORS#L1129

faze-geek:
> @faze-geek how should your name be recorded in the AUTHORS file?
> https://github.com/sympy/sympy/blob/e0188c79d8eb348333b2cc6968b31e7ab3baad2d/AUTHORS#L1129

Hello @oscarbenjamin ,thanks for reaching out .This is Anurag Bhat , a student of IIT Jodhpur, India.

faze-geek:
Ping @smichr for review and to add `limits` label.

faze-geek:
Ping @jksuom 
I believe this is fine low level fix and doesn't require any changes .Can you check and merge if there are no potential changes to be made !

faze-geek:
Ping @jksuom @anutosh491 
Is there appetite to merge this ? If not kindly share reviews / expected changes !

anutosh491:
How are we tackling the case below ? Your Pr would give something like the following 
```
>>> m = Symbol('n', positive=True)
>>> limit(x**m, x, -oo)
oo

# I would expect something like the following
>>> limit(x**(2.5), x, -oo)
oo*I
```
I acknowledge @jksuom 's comment https://github.com/sympy/sympy/issues/8634#issuecomment-180765304 that I wouldn't expect `gruntz` to be doing this but there are few things we would have to consider .Also I'm not convinced with the check and `isinstance(ex,Mul) is False` . Not so sure why that is needed !
I'll try to suggest some changes as and when I understand the best approach here .

sympy/series/tests/test_limits.py
```diff
@@ -615,6 +615,12 @@ def test_issue_8481():
     assert limit(lamda**k * exp(-lamda) / factorial(k), k, oo) == 0
 
 
+def test_issue_8634():
+    n = Symbol('n', integer=True, positive=True)
+    x=Symbol('x')
```
anutosh491, line 620: Spaces around `=`

faze-geek: > How are we tackling the case below ? Your Pr would give something like the following
> 
> ```
> >>> m = Symbol('n', positive=True)
> >>> limit(x**m, x, -oo)
> oo
> 

Yeah this was an initial pull request where I stuck to the issue in particular .The issue was opened with the assumption that n = integer ,hence the changes are not broadened to solve something like this .We could work on this in this pr itself !?

anutosh491:
I feel this could do maybe , I had implemented some part of this , so I see maybe this doing the job here , but i'll confirm it out and see if we can do better . Till then you could confirm and see for if anything fails !
```
--- a/sympy/series/limits.py
+++ b/sympy/series/limits.py
@@ -323,20 +323,12 @@ def set_signs(expr):
                 elif ex == 0:
                     return coeff
                 elif ex.is_negative:
-                    if ex.is_integer:
-                        if cdir == 1 or ex.is_even:
-                            return S.Infinity*sign(coeff)
-                        elif cdir == -1:
-                            return S.NegativeInfinity*sign(coeff)
-                        else:
-                            return S.ComplexInfinity
+                    if cdir == 1:
+                        return S.Infinity*sign(coeff)
+                    elif cdir == -1:
+                        return S.Infinity*sign(coeff)*S.NegativeOne**ex
                     else:
-                        if cdir == 1:
-                            return S.Infinity*sign(coeff)
-                        elif cdir == -1:
-                            return S.Infinity*sign(coeff)*S.NegativeOne**ex
-                        else:
-                            return S.ComplexInfinity
+                        return S.ComplexInfinity
```

anutosh491:
we should also focus on improving test coverage here , there are only 2 tests supporting this check .
```
                        elif cdir == -1:
                           return S.NegativeInfinity*sign(coeff)
```
And two supporting the other one which is this 
```
                        elif cdir == -1:
                            return S.Infinity*sign(coeff)*S.NegativeOne**ex
```
And all other checks in the blocks are exactly same , hence currently I don't see two seperate blocks being necessary here , some good tests would maybe ensure that both blocks are necessary !

faze-geek:
> I feel this could do maybe , I had implemented some part of this , so I see maybe this doing the job here , but i'll confirm it out and see if we can do better . Till then you could confirm and see for if anything fails !

Thanks @anutosh491 . 

Ping @jksuom .Would be glad if you could comment on the latest diff .Thanks !

faze-geek:
> And all other checks in the blocks are exactly same , hence currently I don't see two separate blocks being necessary here , some good tests would maybe ensure that both blocks are necessary !

Yes ,I agree with your point here .Adding a bunch of unit tests can be done in a separate PR in the future .

sympy/series/tests/test_limits.py
```diff
@@ -615,6 +615,12 @@ def test_issue_8481():
     assert limit(lamda**k * exp(-lamda) / factorial(k), k, oo) == 0
 
 
+def test_issue_8634():
+    n = Symbol('n', integer=True, positive=True)
+    x = Symbol('x')
+    assert limit(x**n, x, -oo) == oo/(-1)**n
```
jksuom, line 621: The result looks odd. Maybe there should be `(-1)**-ex` on line 326 above.

anutosh491: if I remember correctly it used to be like that originally but Christopher  had told me to change the form , for easy interpretation of a test on one of my prs (#22491)
So it was me who had changed this , surely could be changed back but couple tests would have to be changed I reckon !

faze-geek: I had tried a decent amount of examples on terminal before making the commit .The result was consistent with forms of `oo/(-1)**n` .I'm not too sure how to change the result ( I believed the above one was expected ) ,so I would request some assistance for that !

anutosh491: I reckon we could try changing line 326 to `S.NegativeInfinity*sign(coeff)*S.NegativeOne**(S.One + ex)` and change tests based on the current implementation of line 326 to what we obtain from the new implementation , @faze-geek you could try that !

faze-geek: Okay I'll make the change in a commit soon ( assuming this is inline with what Kalevi suggested )

certik:
I edited the title, let me know if it is accurate.

anutosh491:
> 
> ```
> --- a/sympy/series/limits.py
> +++ b/sympy/series/limits.py
> @@ -323,20 +323,12 @@ def set_signs(expr):
>                  elif ex == 0:
>                      return coeff
>                  elif ex.is_negative:
> -                    if ex.is_integer:
> -                        if cdir == 1 or ex.is_even:
> -                            return S.Infinity*sign(coeff)
> -                        elif cdir == -1:
> -                            return S.NegativeInfinity*sign(coeff)
> -                        else:
> -                            return S.ComplexInfinity
> +                    if cdir == 1:
> +                        return S.Infinity*sign(coeff)
> +                    elif cdir == -1:
> +                        return S.Infinity*sign(coeff)*S.NegativeOne**ex
>                      else:
> -                        if cdir == 1:
> -                            return S.Infinity*sign(coeff)
> -                        elif cdir == -1:
> -                            return S.Infinity*sign(coeff)*S.NegativeOne**ex
> -                        else:
> -                            return S.ComplexInfinity
> +                        return S.ComplexInfinity
> ```

I am sure this change works here , if this diff is applied the issue is solved . The current implementation is faulty , if `ex` is `odd `by any chance , there is a possibility for oscillating outputs as is the case here ! I ran all checks and some test examples too ,and I feel this could be merged .

A subtle change here just to make the output look better would be the following , 
```
elif cdir == -1:
       S.NegativeInfinity*sign(coeff)*S.NegativeOne**(S.One + ex)
```
Otherwise everything looks fine here !

anutosh491:
> I edited the title, let me know if it is accurate.

it is accurate !

faze-geek:
@anutosh491 @oscarbenjamin . Can you help me figure out the failing authors test here. This PR looks ready to be merged once that passes.

anutosh491:
> This PR looks ready to be merged once that passes.

The changes look OK to me. We could get this in once the mailmap error is fixed (and the unnecessary commits are squashed I guess) . Will have to think how the mailmap error needs to be fixed . Maybe Oscar can guide us better here as I myself struggle with mailmap sometimes but will give it a try !

.mailmap
```diff
@@ -1095,6 +1095,8 @@ Robert Dougherty-Bliss <robert.w.bliss@gmail.com>
 Robert Johansson <jrjohansson@gmail.com>
 Robert Kern <robert.kern@gmail.com>
 Robert Pollak <robert.pollak@posteo.net> Robert Pollak <robert.pollak@jku.at>
+Robin Richard <raisin@ecomail.fr> robinechuca <raisin@ecomail.fr>
+Robin Richard <raisin@ecomail.fr> robinechuca <61911191+robinechuca@users.noreply.github.com>
```
oscarbenjamin, line 1099: These lines should not be added to the .mailmap file.

faze-geek: A bit strange, because I didn't manually add these lines. Maybe I forgot to delete them when I was resolving the merge conflict.

faze-geek: Where should I run `python bin/mailmap_check.py ` , I can't see such a command in the `bin` directory.

faze-geek: Personally I would like to get my work in before the release, so I'll make a separate pull request if I can't workaround the failing authors test soon.

oscarbenjamin: This PR was created before the `bin/mailmap_check.py` script was added. If you delete these two lines and rebase to current master and squash the commits then it should be fine. You might find it easier just to start a fresh PR from current master though.

Issue #22718, opened 2021-12-20
solve(particular=True) with integer variables fails with IndexError


user202729:
Example code:

```
import sympy
from sympy import Matrix, symbols, solve

A=Matrix([[1, 1, 1, 0, 1, 1, 0, 1, 0, 0, 1, 1, 1, 0], [1, 1, 0, 1, 1, 0, 1, 0, 1, 0, -1, -1, 0, 0], [-1, -1, 0, 0, -1, 0, 0, 0, 0, 0, 1, 1, 0, 1], [1, 0, 1, 1, 0, 1, 1, 0, 0, 1, -1, 0, -1, 0], [-1, 0, -1, 0, 0, -1, 0, 0, 0, 0, 1, 0, 1, 1], [-1, 0, 0, -1, 0, 0, -1, 0, 0, 0, -1, 0, 0, -1], [0, 1, 1, 1, 0, 0, 0, 1, 1, 1, 0, -1, -1, 0], [0, -1, -1, 0, 0, 0, 0, -1, 0, 0, 0, 1, 1, 1], [0, -1, 0, -1, 0, 0, 0, 0, -1, 0, 0, -1, 0, -1], [0, 0, -1, -1, 0, 0, 0, 0, 0, -1, 0, 0, -1, -1], [0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0], [0, 0, 0, 0, -1, -1, 0, -1, 0, 0, 0, 0, 0, 0]])
v=Matrix(symbols("v:14", integer=True))
B=Matrix([[2], [-2], [0], [0], [0], [0], [0], [0], [0], [0], [0], [0]])


solve(A@v-B, particular=True)
```

Error message:

```
Traceback (most recent call last):
  File "FILE", line 11, in <module>
    solve(A@v-B, particular=True)
  File "PATH/python3.9/site-packages/sympy/solvers/solvers.py", line 1097, in solve
    solution = _solve_system(f, symbols, **flags)
  File "PATH/python3.9/site-packages/sympy/solvers/solvers.py", line 1812, in _solve_system
    result = minsolve_linear_system(matrix, *symbols, **flags)
  File "PATH/python3.9/site-packages/sympy/solvers/solvers.py", line 2232, in minsolve_linear_system
    bestsol = minsolve_linear_system(system, *symbols, quick=True)
  File "PATH/python3.9/site-packages/sympy/solvers/solvers.py", line 2214, in minsolve_linear_system
[...]

Issue #22717, opened 2021-12-20
solve has trouble with getting full solution


smichr:
From this [SO post](https://stackoverflow.com/questions/70413564/sympy-seems-to-fail-solving-a-system-of-non-linear-equations-while-being-able-to/70416346#70416346):
```python
>>> eqs = Tuple(-y**2 + log(y**2/x) + 2, -2*x*y + 2*x/y)
>>> solve(eqs)
[{y: -1}, {y: 1}]
```
but that is not complete since the first equation is not 0 for all values of x:
```python
>>> eqs.subs(y,1)
(log(1/x) + 1, 0)
```
The manual option for solve resolves the solutions:
```python
>>> solve((-y**2 + log(y**2/x) + 2, -2*x*y + 2*x/y), manual=True)
[{y: -1, x: E}, {y: 1, x: E}]
```

PR #22719, opened 2021-12-20
`solved_syms` and `particular` handling issues

smichr:

#22717, #22718, #22723 (fixes first two)

Systems of equations that have failed equations need to test variables that have already been solved because not all the solved variables will necessarily satisfy the equations that haven't been tested.

#### Release Notes

* solvers
    * a bug fix in checking solutions now allows more complete solutions to some nonlinear systems using `solve`
    * use of the flag `quick` wthout setting `particular=True` will raise an error
    * an error is raised if `particular` is True but a single equation (not in a list) was passed to `solve`
    * when using `particular=True` an error is no longer raised when a solution compatible with assumptions is not found
    * a bug where less than complete solutions were being returned for systems of equations from `solve` was addressed


sympy/solvers/solvers.py
```diff
@@ -849,13 +849,21 @@ def solve(f, *symbols, **flags):
...
     bare_f = not iterable(f)
+
+    # check flag usage
+    if flags.get('particular', False):
+        if bare_f:
+            raise ValueError(
+                "pass f as a list when using 'particular=True'")
+    elif flags.get('quick', None) is not None:
+        raise ValueError(
+            "set 'particular=True' when the 'quick' flag.")
```
smichr, line 860:
```suggestion
            "set 'particular=True' when the 'quick' flag is not None.")
```

sympy/solvers/tests/test_solvers.py
```diff
@@ -1613,6 +1616,26 @@ def count(dic):
...
+        [ 0, -1, -1,  0,  0,  0,  0, -1,  0,  0,  0,  1,  1,  1],
+        [ 0, -1,  0, -1,  0,  0,  0,  0, -1,  0,  0, -1,  0, -1],
+        [ 0,  0, -1, -1,  0,  0,  0,  0,  0, -1,  0,  0, -1, -1],
+        [ 0,  0,  0,  0,  1,  1,  1,  1,  1,  1,  0,  0,  0,  0],
+        [ 0,  0,  0,  0, -1, -1,  0, -1,  0,  0,  0,  0,  0,  0]])
+    v = Matrix(symbols("v:14", integer=True))
+    B = Matrix([[2], [-2], [0], [0], [0], [0], [0], [0], [0],
+        [0], [0], [0]])
+    eqs = A@v-B
+    assert solve(eqs) == []
```
oscarbenjamin, line 1637: Why does `solve(eqs)` return no solutions here?

This is the output from `linsolve`:
```python
In [14]: linsolve((A, B))
Out[14]: 
⎧⎛          3⋅τ₂   1                                          τ₂   2          τ₂   4               
⎨⎜τ₀ + τ₁ + ──── + ─, 4/3 - τ₀, -τ₁ - 2/3, -τ₂ - 2/3, -2⋅τ₁ - ── - ─, -2⋅τ₀ - ── + ─, -2/3, 2⋅τ₀ + 
⎩⎝           4     3                                          2    3          2    3               

                                       3⋅τ₂                ⎞⎫
2⋅τ₁ + τ₂ - 2/3, -2/3, 4/3, -τ₀ - τ₁ - ──── + 1, τ₀, τ₁, τ₂⎟⎬
                                        4                  ⎠⎭

```

smichr: `solve` doesn't handle infinite solutions: the `tau` parameters are showing a parametric solution for the equations.

oscarbenjamin: Usually `solve` does handle infinite solutions for linear and polynomial systems:
```python
In [8]: solve([x + y], [x, y])
Out[8]: {x: -y}

In [9]: solve([x + 2 * y + 3 * z, 4 * x + 5 * y + 6 * z, 7 * x + 8 * y + 9 * z], [x, y, z])
Out[9]: {x: z, y: -2⋅z}

In [10]: solve([x ** 2 + y ** 2 + z, z + z], [x, y, z])
Out[10]: [(-ⅈ⋅y, y, 0), (ⅈ⋅y, y, 0)]
```

oscarbenjamin: There must be some problem with checking somewhere:
```python
In [22]: sol = solve(eqs, check=False)

In [23]: sol
Out[23]: 
⎧      v₃   v₇   1        v₃   v₅   1       v₃   v₇   5       v₃   v₅   5       v₃   v₅   v₇   1   
⎨v₀: - ── + ── + ─, v₁: - ── + ── + ─, v₁₀: ── - ── + ─, v₁₁: ── - ── + ─, v₁₂: ── + ── + ── - ─, v
⎩      4    2    2        4    2    2       4    2    6       4    2    6       4    2    2    6   

                     v₃   v₅   v₇   1                                           ⎫
₁₃: -v₃ - 2/3, v₂: - ── - ── - ── - ─, v₄: -v₅ - v₇, v₆: -2/3, v₈: -2/3, v₉: 4/3⎬
                     4    2    2    2                                           ⎭

In [24]: eqs.subs(sol).is_zero_matrix
Out[24]: True

In [25]: checksol(eqs, sol)
Out[25]: True
```

smichr: The `integer` assumption was used but `v1 = 4/3 - v11` so the assumption is violated.

sympy/solvers/solvers.py
smichr, line 860:
```suggestion
        particular = True  # `quick` is only used in this context
```

smichr, line 860:
```suggestion
    if flags.get('quick', None) is not None:
        particular = True
    elif flags.get('particular', False) and bare_f:
        raise ValueError(filldedent("""
            The 'particular' flag is usually used with systems of
            equations. Either pass your equation as a list or
            consider using a solver like `diophantine` if you are
            looking for a solution in integers."""))
```

sympy/solvers/tests/test_solvers.py
```diff
@@ -2433,3 +2456,10 @@ def test_solver_flags():
     root = solve(x**5 + x**2 - x - 1, cubics=False)
     rad = solve(x**5 + x**2 - x - 1, cubics=True)
     assert root != rad
+
+
+def test_issue_22717():
+    assert solve((-y**2 + log(y**2/x) + 2, -2*x*y + 2*x/y)) == [
+        {y: -1, x: E}, {y: 1, x: E}]
+    raises(ValueError, lambda: solve(x + 1, particular=True))
+    raises(ValueError, lambda: solve([x + 1], quick=True))
```
smichr, line 2465:
```suggestion
    raises(ValueError, lambda: solve(x + 1, particular=True))
    raises(ValueError, lambda: solve(x + 1, quick=True))
    assert solve([x + y - 5, 2*x - y - 1], quick=True) == {x: 2, y: 3}
```

sympy/solvers/tests/test_solvers.py
smichr: When this set of equations is used in master, an error is [raised](https://github.com/sympy/sympy/issues/22718#issue-1084286569), so the test is just checking that the error is no longer there. Until the `igcdLLL` or similar is added, there is not support of Diophantine solutions for a set of linear equations.

sympy/solvers/solvers.py
```diff
@@ -849,13 +849,23 @@ def solve(f, *symbols, **flags):
     def _sympified_list(w):
         return list(map(sympify, w if iterable(w) else [w]))
     bare_f = not iterable(f)
+
+    # check flag usage
+    if flags.get('quick', None) is not None:
+        particular = True
```
smichr, line 855:
```suggestion
```

sympy/solvers/tests/test_solvers.py
oscarbenjamin: Oh, I missed the integer assumptions!

sympy/solvers/solvers.py
```diff
@@ -1908,15 +1919,16 @@ def key(sym):
                 rv = sorted(rv, key=key)
             return rv
 
-        solved_syms = set(solved_syms)  # set of symbols we have solved for
         legal = set(symbols)  # what we are interested in
         # sort so equation with the fewest potential symbols is first
         u = Dummy()  # used in solution checking
+        solved_syms = set()  # ignore for sake of ordering
```
oscarbenjamin, line 1925: Why is this set before the loop if each iteration resets it anyway?

sympy/solvers/solvers.py
```diff
@@ -849,13 +849,24 @@ def solve(f, *symbols, **flags):
...
+    if flags.get('quick', None) is not None:
+        if flags.get('particular') == False:
+            raise ValueError('conflicting flags: particular=False and quick=True')
+        flags['particular'] = True
+    if flags.get('particular', False) and bare_f:
+        raise ValueError(filldedent("""
+            The 'particular/quick' flag is usually used with systems of
+            equations. Either pass your equation as a list or
+            consider using a solver like `diophantine` if you are
+            looking for a solution in integers."""))
```
oscarbenjamin, line 863: The flow control in this function is very complicated. I worry that this makes it worse.

smichr: `quick` and `particular` are for the same solution path.
The options are particular/quick one of [True/True, True/False].
So if `quick` is set (i.e., not None) it sets `particular`.
Nothing ambiguous here.

If the user passes a single equation with `particular=True`, they are told that this is an error and redirected to what they probably are intending or to resend equation in a list.
Less ambiguity here.

sympy/solvers/solvers.py
smichr: Because it is needed to get the loop going with the call to `_ok_syms`. The appearance inside the loop near the top is (to me) preferred over hiding it at the bottom after all the if blocks are done.

smichr:
I have no further additions. This is ready for final review.

smichr:
will commit tomorrow if there are no objections

smichr:
* The flag handling is being done at the top level.
* the indented code checks the solution of undetermined coeffs -- it is not included in the try block and could be in an `else;` post-try block, but then it would need to be even more indented.
* when many symbols were requested for a single equation, I feel badly solving it for symbols looking for linear then nonlinear solutions that are in terms of variables not already solved for. Why? The only way to solve for more than one symbol is if you can separate variables (i.e. the expression is a Mul). But we are not at that stage in the `_solve` routine -- and my feeling is that any Mul should have been separated into factors back in `solve`, creating a system of smaller expressions. But that would still permit an expanded product to arrive in `_solve` so I guess the situation still has to be resolved somehow: find the implicit factors containing the symbols of interest and solve for them.
* Regarding the mesy nature of the routine. I'm sure there is some metric to quantify this, but for counter point, I think you made the modification of this are of code in #21883. I found the concept difficult to understand at first but the scope of that solution process is small -- fitting about on 1 screen -- so it wasn't hard to work my way through it and I feel comfortable with the change. It's not making things worse.
* the elimination of the `solved_syms` was relatively easy to make (without increasing complexity) and was much easier for me to correct in `solve` than in `solveset` (which was much harder for me to find the solution). I still am not able to make the parallel fix there.

sympy/solvers/tests/test_solvers.py
```diff
@@ -1605,14 +1605,47 @@ def test_high_order_roots():
 
 
 def test_minsolve_linear_system():
+    assert solve([x + y - 5, 2*x - y - 1], quick=True) == {x: 2, y: 3}
+    assert solve([x + y - 5, 2*x - y - 1], quick=False) == {x: 2, y: 3}
     def count(dic):
         return len([x for x in dic.values() if x == 0])
-    assert count(solve([x + y + z, y + z + a + t], particular=True, quick=True)) \
+    assert count(solve([x + y + z, y + z + a + t], quick=True)) \
```
oscarbenjamin, line 1612: The quick=True flag is already quite misleading I think but making it imply particular=True doesn't help.

sympy/solvers/solvers.py
```diff
@@ -2222,7 +2250,10 @@ def update(determined, solution):
             if len(k.free_symbols) != 1:
                 determined[x] = S.Zero
             else:
-                val = solve(k)[0]
+                val = solve(k)
+                if not val:
+                    raise NotImplementedError('assumptions may have prohibited finding a solution')
```
oscarbenjamin, line 2255: Why is `solve` being used here if it potentially leads to this problem?

sympy/solvers/solvers.py
```diff
@@ -1320,29 +1334,41 @@ def _solve(f, *symbols, **flags):
...
+                            for d in soln:
+                                for k in d:
+                                    d[k] = simplify(d[k])
+                        elif isinstance(soln[0], tuple):
+                            soln = [tuple(simplify(i) for i in j) for j in soln]
+                        else:
+                            raise TypeError('unrecognized args in list')
+                    elif isinstance(soln, tuple):
+                        sym, sols = soln
+                        soln = sym, {tuple(simplify(i) for i in j) for j in sols}
```
oscarbenjamin, line 1353: It seems absurd to me that so many different cases need to be handled here. The code should really be refactored so that at least internally consistent types are used.

sympy/solvers/solvers.py
```diff
@@ -849,13 +853,23 @@ def solve(f, *symbols, **flags):
...
+    # check flag usage for particular/quick which should only be used
+    # with systems of equations
+    if flags.get('quick', None) is not None:
+        flags['particular'] = True  # quick trumps particular
+    if flags.get('particular', False) and bare_f:
+        raise ValueError(filldedent("""
+            The 'particular/quick' flag is usually used with systems of
+            equations. Either pass your equation as a list or
+            consider using a solver like `diophantine` if you are
+            looking for a solution in integers."""))
```
oscarbenjamin, line 866: This seems like an unnecessary restriction. If we can have a system then why can't we have a system of size 1?

sympy/solvers/tests/test_solvers.py
smichr: Setting `quick=True` has no meaning outside of the `particular=True` context. Do you think it would be better to raise an error indicating that `particular` must be True when setting `quick` to True or False?

sympy/solvers/solvers.py
smichr: It depends how you define the problem. This can also be viewed as a linear system whose solution is prohibited by assumptions. It were better for the user to have sent it to `diophantine`, but getting a `diophantine` solution from a `solve` call doesn't seem like a less complicated direction to go.

sympy/solvers/solvers.py
smichr: I agree. The different return types for different types of equations is something that got baked in at the beginning and all development just retained those types. Anyone not liking this can use `dict=True`.  There's no getting around that *feature* until we agree that it's a good time to break backwards compatibilty. The docstring of `solve` enumerates all the different input and return types.

smichr: I appreciate the significance of "internally" in your comment. It's just that at some point you have to return the expected type of the solution, regardless of what you do internally. Maybe it would be better to break out the simplification and checking process so the `_solve` routine doesn't have to display this level of complexity?

sympy/solvers/solvers.py
oscarbenjamin: I wouldn't have expected this particular `minsolve_linear_system` to be affected by assumptions at all. Also I wouldn't have expected it to use a general purpose function like `solve` at all. Is this not just solving an univariate linear equation?

sympy/solvers/solvers.py
oscarbenjamin: The different types of input and output should ideally be handled in one function only before/after any other processing so that all the logic is in one place and internally a consistent form (presumably list of dicts) should be used everywhere.

sympy/solvers/solvers.py
smichr: I didn't see the context. I agree that this should not be calling `solve`.

sympy/solvers/tests/test_solvers.py
```diff
@@ -1605,14 +1605,47 @@ def test_high_order_roots():
...
+        [ 0, -1,  0, -1,  0,  0,  0,  0, -1,  0,  0, -1,  0, -1],
+        [ 0,  0, -1, -1,  0,  0,  0,  0,  0, -1,  0,  0, -1, -1],
+        [ 0,  0,  0,  0,  1,  1,  1,  1,  1,  1,  0,  0,  0,  0],
+        [ 0,  0,  0,  0, -1, -1,  0, -1,  0,  0,  0,  0,  0,  0]])
+    v = Matrix(symbols("v:14", integer=True))
+    B = Matrix([[2], [-2], [0], [0], [0], [0], [0], [0], [0],
+        [0], [0], [0]])
+    eqs = A@v-B
+    assert solve(eqs) == []
+    raises(NotImplementedError, lambda: solve(eqs, particular=True))
```
smichr, line 1637:
```suggestion
    assert solve(eqs, particular=True) == []  # assumption violated
```

sympy/solvers/tests/test_solvers.py
oscarbenjamin: I think perhaps it would be good to give a warning (not a deprecation warning but a general warning) if `quick` is used without `particular`. I often see people try various combinations of parameters in an attempt to make `solve` faster and I can imagine seeing a parameter like `quick` and trying that without expecting this behaviour.

sympy/solvers/tests/test_solvers.py
```diff
@@ -1605,14 +1605,47 @@ def test_high_order_roots():
...
+    # check that quick flag trumps particular flag in case anyone
+    # tries to make default flags match the docstring description
+    assert all(v for v in solve([x + y + z, y + z + a]).values())
+    for _q in (True, False):
+        assert not all(v for v in solve(
+            [x + y + z, y + z + a], quick=_q,
+            particular=False).values())
+        # and give a good error message if someone tries to use
+        # particular with a single equation
+        raises(ValueError, lambda: solve(x + 1, quick=_q))
```
smichr, line 1647:
```suggestion
    assert all(v for v in solve([x + y + z, y + z + a]).values())
    for _q in (True, False):
        assert not all(v for v in solve(
            [x + y + z, y + z + a], quick=_q,
            particular=True).values())
        # raise error if quick used w/o particular=True
        raises(ValueError, lambda: solve([x + 1], quick=_q))
        raises(ValueError, lambda: solve([x + 1], quick=_q, particular=False))
    # and give a good error message if someone tries to use
    # particular with a single equation
    raises(ValueError, lambda: solve(x + 1, particular=True))
```

sympy/solvers/solvers.py
oscarbenjamin: Ideally there would be no difference between passing `[eq]` and `eq`. The fact that those aren't equivalent here is part of what makes solve such a mess and also is one of the things that caused difficulty when I tried to clean things up. A single equation given as `eq` should always just be treated as being equivalent to a system of 1 equation like `[eq]`.

sympy/solvers/tests/test_solvers.py
```diff
@@ -1605,14 +1605,48 @@ def test_high_order_roots():
 
 
 def test_minsolve_linear_system():
+    pqt = dict(quikc=True, particular=True)
+    pqf = dict(quikc=False, particular=True)
```
smichr, line 1609:
```suggestion
    pqt = dict(quick=True, particular=True)
    pqf = dict(quick=False, particular=True)
```

sympy/solvers/solvers.py
oscarbenjamin: Maybe I'm misunderstanding how this works but I thought that the first argument is always the system of equations: it can be a single equation or a list of equations. The second argument is always either a list of unknowns or the first of the unknowns.

oscarbenjamin:
> Can we get this into 1.10, @oscarbenjamin ?

No, I don't think so. Problems are already surfacing with the release and I don't want to complicate it with more changes.

sympy/solvers/solvers.py
```diff
@@ -1349,21 +1373,22 @@ def _solve(f, *symbols, **flags):
...
+                rhs_s |= vfree
                 got_s.add(xi)
                 result.append({xi: v})
             elif xi:  # there might be a non-linear solution if xi is not 0
-                failed.append(s)
-        if not failed:
-            return result
-        for s in failed:
+                nonlin_s.append(s)
+        for s in nonlin_s:
```
smichr, line 1386:
```suggestion
        if not nonlin_s:
            return result
        for s in nonlin_s:
```

sympy/solvers/solvers.py
```diff
@@ -847,13 +848,24 @@ def solve(f, *symbols, **flags):
     def _sympified_list(w):
         return list(map(sympify, w if iterable(w) else [w]))
     bare_f = not iterable(f)
+
+    # check flag usage for particular/quick which should only be used
+    # with systems of equations
+    if flags.get('quick', None) is not None:
+        if not flags.get('particular', None):
+            raise ValueError('when using `quick`, `particular` should be True')
```
oscarbenjamin, line 856: A warning would be better than an exception here. If someone is using this mistakenly then they can fix their code after seeing the warning but their code will continue to work in the mean time.

sympy/solvers/solvers.py
```diff
@@ -847,13 +848,24 @@ def solve(f, *symbols, **flags):
...
+    if flags.get('particular', False) and bare_f:
+        raise ValueError(filldedent("""
+            The 'particular/quick' flag is usually used with systems of
+            equations. Either pass your equation in a list or
+            consider using a solver like `diophantine` if you are
+            looking for a solution in integers."""))
```
oscarbenjamin, line 862: I don't think this is a good idea. It is already problematic that various codepaths depend on the `bare_f` variable and this makes it worse.

PR #22721, opened 2021-12-20
Fuzzy bool explained for vague subets

### LINKED ISSUE #9855: bad subset evaluation
> > > x, y, z = symbols('x, y, z', real=True)
> > > s1 = Interval(1,x) & Interval(y,2)
> > > s2 = Interval(1,2)
> > > s1.is_subset(s2)
> > > False
> > > simplify(s1.contains(z) >> s2.contains(z))
> > > True

ejcohe22:

Fixes #9855

Added a useful printout for the fuzzy bool returned in the "is_subset" method. This is caused by intervals defined with variables. In the case (1, x) where x is a symbol in the real number domain, if x < 1 the set is empty, leading to unclear truth tables.

#### Other comments
Please accept part of a class assignment due 12/20/21 11:59:00 PM EST

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/sets/sets.py
+++ b/sympy/sets/sets.py
@@ -399,6 +399,9 @@ class Set(Basic, EvalfMixin):
         if self.intersect(other) == self:
             return True
 
+        # Function defaults to None!
+        print("Boolean undeterminable. Set is defined vaguely")
+
     def _eval_is_subset(self, other):
         '''Returns a fuzzy bool for whether self is a subset of other.'''
         return None


sympy/sets/sets.py
```diff
@@ -399,6 +399,9 @@ def is_subset(self, other):
         if self.intersect(other) == self:
             return True
 
+        # Function defaults to None!
+        print("Boolean undeterminable. Set is defined vaguely")
```
oscarbenjamin, line 403: We don't add print statements in library code in general.

ejcohe22:
I am going to create a new pull request but with this warning print out as a note in the docstring. Thank you for the quick response!

PR #22731, opened 2021-12-20
Updated Docstring of "is_subset()"

### LINKED ISSUE #9855: bad subset evaluation

ejcohe22:

Fixes #9855

Added a useful docstring to explain the fuzzy bool returned in the "is_subset()" method. This is caused by intervals defined with variables. In the case (1, x) where x is a symbol in the real number domain, if x < 1 the set is empty, leading to unclear truth tables. This leads to confusion because the current docstring only describes True/False bools being returned.

#### Other comments
Please accept part of a class assignment due 12/20/21 11:59:00 PM EST

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/sets/sets.py
+++ b/sympy/sets/sets.py
@@ -365,7 +365,7 @@ class Set(Basic, EvalfMixin):
         False
         >>> from sympy import symbols
         >>> x, y = symbols('x, y', real = True)
-        >>> (Interval(1,x) & Interval(y,2)).is_subset(Interval(1, 2)) 
+        >>> (Interval(1,x) & Interval(y,2)).is_subset(Interval(1, 2))
         # Returns nothing
 
         """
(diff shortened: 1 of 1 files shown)


sympy/sets/sets.py
```diff
@@ -353,6 +353,7 @@ def _contains(self, other):
     def is_subset(self, other):
         """
         Returns True if ``self`` is a subset of ``other``.
+        Returns None if ``self`` or ``other`` is a vaguely defined set.
```
oscarbenjamin, line 356: This is not exactly a clear description.

oscarbenjamin:
Thanks for your contribution and sorry this was not reviewed before.

I don't think that the change here really communicates very clearly what this does and at this point it is probably better just to close the PR at this point.

PR #22732, opened 2021-12-20
DO NOT MERGE: Testing Python 3.11

oscargus:

Just to see what happens...

I have removed everything else from the tests, so do NOT merge...

#### Release Notes

NO ENTRY


oscargus:
Seems that there are no new deprecation warnings and that SymPy should work on 3.11.

Main problem right now is that NumPy is not supporting 3.11. I may change the order of the optional dependencies if I get bored at some stage.

Will keep this open in case there is some feedback.

.github/workflows/runtests.yml
```diff
@@ -90,44 +88,41 @@ jobs:
...
   tests1-latest:
-    needs: code-quality
 
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.9'
+          python-version: '3.11.0-alpha.2'
```
oscarbenjamin, line 97: The setup-python README suggests that you can use a range here like
```
python-version: '3.11.0-alpha - 3.11.0'
```
I presume that means that it will use the newest version until 3.11.0. Maybe it's possible to change it to 3.11 so it will continue using newest versions of 3.11 after 3.11.0 is released...

https://github.com/actions/setup-python

oscargus: OK! I'll try it! (I noticed that alpha-4 was available so I was about to change anyway.)

.github/workflows/runtests.yml
```diff
@@ -23,7 +23,7 @@ jobs:
 
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.9'
+          python-version: '3.11.0-alpha.2'
```
oscargus, line 26:
```suggestion
          python-version: '3.11.0-alpha - 3.11'
```

.github/workflows/runtests.yml
```diff
@@ -64,7 +63,7 @@ jobs:
...
```
oscargus, line 66:
```suggestion
          python-version: ''3.11.0-alpha - 3.11'
```

.github/workflows/runtests.yml
```diff
@@ -74,14 +73,13 @@ jobs:
...
   doctests-latest:
-    needs: code-quality
 
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.9'
+          python-version: '3.11.0-alpha.2'
```
oscargus, line 82:
```suggestion
          python-version: ''3.11.0-alpha - 3.11.0''
```

.github/workflows/runtests.yml
oscargus, line 97:
```suggestion
          python-version: ''3.11.0-alpha - 3.11.0''
```

.github/workflows/runtests.yml
```diff
@@ -90,44 +88,41 @@ jobs:
...
   tests2-latest:
-    needs: code-quality
 
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.9'
+          python-version: '3.11.0-alpha.2'
```
oscargus, line 111:
```suggestion
          python-version: '3.11.0-alpha - 3.11'
```

.github/workflows/runtests.yml
```diff
@@ -90,44 +88,41 @@ jobs:
...
   optional-dependencies:
-    needs: code-quality
 
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.9'
+          python-version: '3.11.0-alpha.2'
```
oscargus, line 125:
```suggestion
          python-version: '3.11.0-alpha - 3.11'
```

.github/workflows/runtests.yml
```diff
@@ -140,231 +135,31 @@ jobs:
...
   slow1:
-    needs: [doctests-latest, tests1-latest, tests2-latest]
 
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.9'
+          python-version: '3.11.0-alpha.2'
```
oscargus, line 148:
```suggestion
          python-version: '3.11.0-alpha - 3.11'
```

.github/workflows/runtests.yml
```diff
@@ -140,231 +135,31 @@ jobs:
...
   slow2:
-    needs: [doctests-latest, tests1-latest, tests2-latest]
 
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
         with:
-          python-version: '3.9'
+          python-version: '3.11.0-alpha.2'
```
oscargus, line 162:
```suggestion
          python-version: '3.11.0-alpha - 3.11'
```

.github/workflows/runtests.yml
oscargus: Seems to work!

(Matplotlib seems to have a harder time with alpha-4 compared to alpha-2 though... Or maybe the actual install order is randomized in some way...)

oscargus: Or rather: there is a cached numpy wheel for alpha-2. But should just be a question of time before there is one for alpha-4 as well and we are back to scipy and kiwisolver causing issues (which seems to be the ones having problems on the latest alpha-2 build).

Issue #15680, opened 2018-12-20
Replace pyglet as the default viewer in the preview() function


asmeurer:
Right now preview() by default uses pyglet to show the image. I think it would be better to have a more portable default. Some possibilities are the `PIL` `show` function, which opens the image in the default image viewer for the platform, or using `matplotlib.imshow`. Probably both should be supported and the default would try each until it finds the best one that works. We can leave the pyglet viewer there, but not as the default.

asmeurer:
I guess the change of the default should be discussed, but we should add PIL and matplotlib as viewers regardless.

avishrivastava11:
@asmeurer  +1 for the default image viewer. Although I'm not sure about using the `matplotlib.imshow` because of two reasons
1. Its slower than the default image viewer, and also slower to navigate (Or could only be happening in my system; If its so, please let me know)
2. It might have compatibility issues on certain systems where they don't have a certain software installed (I might be wrong here; If so, please correct me)
But I'm unaware of the method to find the default viewer and use that. Could you help me here (as I would like to take up this issue)?

oscarbenjamin:
I have a proof of concept for using `matplotlib.figshow` instead of `pyglet` here: https://github.com/sympy/sympy/issues/8446#issuecomment-557798066

asmeurer:
We should use the system viewer as the default. There's already code there for it (https://github.com/sympy/sympy/pull/19867), so we just need to change the default.

PR #13621, opened 2017-11-18
Various fixes to preview()

asmeurer:
- Document the `preamble` and `dvioptions` flags better, to make it clearer what the defaults are. 
- Allow passing the `mode` flag through to `latex`. 

#### Release notes

* printing
  - Document the `preamble` and `dvioptions` flags to `preview()` better, to make it clearer what the defaults are. 
  - Allow passing the `mode` flag through to `latex` in `preview()`.

PR #22734, opened 2021-12-21
Refactored LaTeX generation in preview

oscargus:

Refactored the interactive LaTeX generation a bit so that the LaTeX preamble etc are defined in a single place.

#### Other comments

As an added bonus one can specify fontsize as an int, but I do not think that warrants a release note. Although it is mentioned in the documentation.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/printing/preview.py
+++ b/sympy/printing/preview.py
@@ -9,6 +9,7 @@ except ImportError:
     pass
 
 from sympy.utilities.decorator import doctest_depends_on
+from sympy.utilities.misc import debug
 from .latex import latex
 
 __doctest_requires__ = {('preview',): ['pyglet']}
@@ -40,7 +41,7 @@ def _run_pyglet(fname, fmt):
     win = window.Window(
         width=img.width + 2*offset,
         height=img.height + 2*offset,
-        caption="sympy",
+        caption="SymPy",
         resizable=False,
         config=config
     )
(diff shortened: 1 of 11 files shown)


oscarbenjamin:
I think this looks good although I'm not really sure how to test it. I don't think I've seen `preview` work on OSX.

oscargus:
All the LaTeX generation (like in IPython and Spyder) goes through `preview`  and it is only the LaTeX-file generation that has been refactored (and slightly changed).

(I have problems testing the other aspects of preview as well.)

asmeurer:
`preview()` works just fine on Mac if you configure it to save to a file. The only issue is with the default Pyglet viewer. IMO pyglet shouldn't be the default (for preview or plotting).

Ideally `viewer='open'` (meaning use the `open` command, which would open it in Preview.app) ought to work and be the default on Mac, but it doesn't for me (it opens Preview but no image is opened). 

See also https://github.com/sympy/sympy/pull/13621 and https://github.com/sympy/sympy/issues/15680

oscargus:
I was thinking about adding a `'system'` viewer, but that requires platform-specific code and when I didn't get it to work on Windows (because the default viewers returns after starting causing the temporary file to be deleted before it is loaded in the viewer...), I didn't spend more time on it...

asmeurer:
Doing NotImplementedError on Windows for now would be fine. I will say my experience on Windows in general is that you shouldn't delete files that are open in programs like you would in Unix, so you'd have to find a way to only delete the file after the program closes. Perhaps Windows has some API for this?

oscargus:
Except from that the feature request from @asmeurer about an ability to open a file in the system viewer is not implemented, this PR does what it originally was supposed to do: generate the LaTeX-file in a single place and use the same LaTeX-file for different cases. I think it may just as well go into 1.11.

oscarbenjamin:
Not sure what the circleci error is. I guess that's because this PR predates using circleci.

@asmeurer does the circleci build use the PR branch or does it use the branch resulting from merging the PR into master (like most other things in CI do)?

oscarbenjamin:
@oscargus I haven't looked through this in detail but it looks fine. Feel free to merge if you're happy with it.

sympy/printing/preview.py
```diff
@@ -78,12 +79,51 @@ def on_expose():
 
     win.close()
 
+def _get_latex_main(expr, *, preamble=None, packages=(), preamble_lines=(),
```
oscargus, line 82:
```suggestion
def _get_latex_main(expr, *, preamble=None, packages=(), preamble_lines=None,
```

oscargus:
I changed the name of `preamble_lines` to `extra_preamble` to reduce the confusion if it is actually lines (as in a tuple or list) or just a string.

oscargus:
I rebased and then the CircleCI stuff was OK (I guess this was pretty old).

oscarbenjamin:
I think something funny happened with the rebase because it's showing extra commits that were already on master.

oscargus:
Ahh! Sorry about that! And automerge on top of that...

PR #22735, opened 2021-12-21
Updated default LaTeX printing of Basic

oscargus:

Earlier on the LaTeX printing of Basic (and therefore all classes not having dedicated printing) resulted in a quite ugly result (as no formatting was applied). Now, these are printed as `\operatorname` if they have arguments and `\text` if not. This is a better guess in most cases.
Earlier:

image

image

Now:
image
image

Edit: oh, and predicates are printed like:
image

#### Other comments

One thing to consider is how to deal with classes with _ in their names. Right now, the _ is used to indicate subscripts as in the rest of SymPy. This is of course a bit bad for a few classes that happens to include _, but I still think it is better than the opposite.

Also improved the recently introduced ast-printing slightly.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/codegen/ast.py
+++ b/sympy/codegen/ast.py
@@ -915,7 +915,8 @@ class String(Atom, Token):
         return lambda: self
 
     def _latex(self, printer):
-        return r'\texttt{{"{}"}}'.format(self.text)
+        from sympy.printing.latex import latex_escape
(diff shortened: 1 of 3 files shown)


oscarbenjamin:
Looks good to me

Issue #15743, opened 2019-01-05
evalf hangs with expression involving CRootOf


oscarbenjamin:
The following expression arises from `test_nth_linear_constant_coeff_homogeneous_rootof_sol` and seems to cause `evalf()` to hang (I haven't waited more than an hour):

```python
sol = -2*exp(CRootOf(x**5 + 11*x - 2, 0)) + exp(CRootOf(x**5 + 11*x - 2, 0))*CRootOf(x**5 + 11*x - 2, 0)**5 + 11*exp(CRootOf(x**5 + 11*x - 2, 0))*CRootOf(x**5 + 11*x - 2, 0) + 11*exp(CRootOf(x**5 + 11*x - 2, 3))*CRootOf(x**5 + 11*x - 2, 3) + exp(CRootOf(x**5 + 11*x - 2, 4))*CRootOf(x**5 + 11*x - 2, 4)**5 - 2*exp(CRootOf(x**5 + 11*x - 2, 4)) + exp(CRootOf(x**5 + 11*x - 2, 1))*CRootOf(x**5 + 11*x - 2, 1)**5 + 11*exp(CRootOf(x**5 + 11*x - 2, 2))*CRootOf(x**5 + 11*x - 2, 2) - 2*exp(CRootOf(x**5 + 11*x - 2, 2)) - 2*exp(CRootOf(x**5 + 11*x - 2, 1)) + 11*exp(CRootOf(x**5 + 11*x - 2, 1))*CRootOf(x**5 + 11*x - 2, 1) + exp(CRootOf(x**5 + 11*x - 2, 2))*CRootOf(x**5 + 11*x - 2, 2)**5 - 2*exp(CRootOf(x**5 + 11*x - 2, 3)) + exp(CRootOf(x**5 + 11*x - 2, 3))*CRootOf(x**5 + 11*x - 2, 3)**5 + 11*exp(CRootOf(x**5 + 11*x - 2, 4))*CRootOf(x**5 + 11*x - 2, 4)

sol.evalf()
```

The expression pretty-printed is:
```julia
In [17]: pprint(sol)                                                                                                                                          
            ⎛ 5              ⎞           ⎛ 5              ⎞                          5              ⎛ 5              ⎞                                      
[...]

PR #19474, opened 2020-06-02
[experiment] Move old assumptions from Basic to Expr

oscarbenjamin:
#### Brief description of what is fixed or changed

This is still an experiment at this stage. The old assumptions have no place being on Basic rather than Expr and this PR is to see what happens if they are moved to Expr.

#### Release Notes

NO ENTRY

Issue #19164, opened 2020-04-21
POC: Global numerical equation solver


oscarbenjamin:
This idea was previously suggested in #4218.

Since mpmath provides multiprecision interval arithmetic we can use it to implement the interval newton method for numerical solutions of algebraic equations:
https://en.wikipedia.org/wiki/Interval_arithmetic
https://en.wikipedia.org/wiki/Interval_arithmetic#Interval_Newton_method
https://en.wikipedia.org/wiki/Newton%27s_method#Interval_Newton's_method

This is different from ordinary Newton iteration as it *proves* the uniqueness or absence of roots to an equation in an interval. Combined with subdivision we can use this to find all roots within an interval. It can also be used to find all stationary points and hence global maxima/minima.

I have an implementation below that works for solving algebraic equations in a single real unknown. With that we can do things like:
```julia
In [3]: nsolve_interval(1 + x**2, '-inf', 'inf')  # prove no solution over the reals
Out[3]: []

In [20]: nsolve_interval(tan(x), -10, 10)  # provably find all roots in an interval
Out[20]: [-9.42478, -6.28319, -3.14159, 0.0, 3.14159, 6.28319, 9.42478]
```
There are some equations that it struggles with particularly where many derivatives are zero at the root and the derivative is a sum of terms of alternate sign:
```julia
In [28]: nsolve_interval(sin(x)**2 - x**2, -10, 10)                                                                               
^C---------------------------------------------------------------------------
KeyboardInterrupt
```
Other examples are those with infinite roots in the interval:
```julia
In [22]: nsolve_interval(sin(1/x), -10, 10)  # infinite roots
^C---------------------------------------------------------------------------
KeyboardInterrupt

In [27]: nsolve_interval(sin(1/x), 0.01, 1)  # finite roots are fine
Out[27]: 
[0.01027, 0.01061, 0.01097, 0.01137, 0.01179, 0.01224, 0.01273, 0.01326, 0.01384, 0.01447, 0.01516, 0.01591, 0.01675, 0.01768, 0
.01872, 0.01989, 0.02122, 0.02274, 0.02449, 0.02653, 0.02894, 0.03183, 0.03537, 0.03979, 0.04547, 0.05305, 0.06366, 0.07958, 0.1
061, 0.15915, 0.31831]
```

[...]

Issue #19153, opened 2020-04-20
Should RealElement classes be dynamically created?


jksuom:
Currently, there is one static class `RealElement` in the `mpelements` module. Each time a new domain of class `RealField` is created, a new `MPContext` instance is created and the `RealElement` class is assigned to the `mpf` attribute of the context. In addition, the `context` class attribute of `RealField` is patched to point to the created context. This means that the context of the single `RealElement` class will always be the last context created.
```
>>> from sympy import *
>>> a = RR(2)
>>> a.context.prec
53
>>> b = E.n(2)
>>> x = Symbol('x')
>>> p = Poly(x + b, x)
>>> a.context.prec
10
>>> a.parent() == p.domain
True
```
`MPContext` and `RealElement` are subclassed from `mpmath`'s `PythonMPContext` and `_mpf` classes. `PythonMPContext.__init__()` creates dynamically a new `_mpf` class with name `mpf` and assigns it to the `mpf` attribute. This way, instances of different `mpf` classes will preserve their context also when working precision is temporarily changed.

 `MPContext.__init__()` is an an extended copy of `PythonMPContext.__init__()` except that no new `RealElement` class is created. Should that be done in SymPy as well?

oscarbenjamin:
The way this is all handled in mpmath seems overly complicated and clearly makes things awkward e.g. 

Maybe the `RealElement` and `ComplexElement` classes should be removed and `_mpf` and `_mpc` can be used directly. In fact maybe `MPContext` should be removed as well. I don't see what purpose it serves over just using mpmath's MPContext directly.

PR #22737, opened 2021-12-21
Various improvements to `field_isomorphism()`.

### LINKED ISSUE #22736: polys/numberfields: Improvements in `field_isomorphism()` implementation
I think there are a few places where the implementation of the `field_isomorphism()` function could be sped up, simplified, and made safer.

I will open a draft PR to show the changes I want to make. I realize this is a lot to read, but even if it just gets a cursory review, I want to document the reasons for the changes I'm making.

# Background

Given two algebraic numbers `a` and `b`, the `field_isomorphism()` function,

* [in the docs](https://docs.sympy.org/dev/modules/polys/numberfields.html#sympy.polys.numberfields.subfield.field_isomorphism)
* [in the code](https://github.com/sympy/sympy/blob/5d4477c8d2fcb3dc8c46fff3e604f7790ce22640/sympy/polys/numberfields/subfield.py#L160)

tries to write `a` as a polynomial in `b`. In other words, it is checking whether `a` belongs to the number field `Q(b)`.

Internally, it relies on two separate methods for answering this question:

* [`field_isomorphiam_pslq()`](https://github.com/sympy/sympy/blob/5d4477c8d2fcb3dc8c46fff3e604f7790ce22640/sympy/polys/numberfields/subfield.py#L82) which is fast, but might be inconclusive, and
* [`field_isomorphism_factor()`](https://github.com/sympy/sympy/blob/5d4477c8d2fcb3dc8c46fff3e604f7790ce22640/sympy/polys/numberfields/subfield.py#L136) which is slower, but definitely gives an answer.

I want to make improvements in both of the internal methods.

# Proposed Improvements

# Safety

As [noted earlier](https://github.com/sympy/sympy/pull/22464#discussion_r748897805) by @oscarbenjamin, we might want a `try/finally` block here

https://github.com/sympy/sympy/blob/5d4477c8d2fcb3dc8c46fff3e604f7790ce22640/sympy/polys/numberfields/subfield.py#L98-L100

when setting `mp.dps`.

# Speed

As noted in Issue #15743, `evalf` can be very slow when evaluating a sum whose actual value is zero. We have evaluation of zero sums in the internal method  `field_isomorphism_factor()`:

https://github.com/sympy/sympy/blob/5d4477c8d2fcb3dc8c46fff3e604f7790ce22640/sympy/polys/numberfields/subfield.py#L150-L154

and this can be quite slow.

**EXAMPLE:** Under the existing implementation, this call:

```python
a = CRootOf(x**4 + x**3 + x**2 + x + 1, -1)
a._reset()
b = exp(2*I*pi/5)
assert field_isomorphism(a, b) == [1, 0]
```

takes 9.44 seconds, but with the proposed changes, it will take only 0.74 seconds!

**QUESTION:** Are there any gotchas with swapping the order of `evalf` and `+/-` that I'm overlooking? (See my PR.) Or is this fine?

# Simplifications

Both internal methods `field_isomorphism_factor()` and `field_isomorphism_pslq()` seem to be taking unmotivated steps, which may even waste time. I want to cut these steps and simplify the code, but I want to ask if there may be something I'm missing here.

### How this function works

The basic idea is simple. Let `f(x)` be the minimal polynomial for the algebraic number `a` over `Q`, and try factoring `f(x)` over the field `Q(b)`. Then the number `a` belongs to the field `Q(b)` if and only if we find the linear factor `x - a` among the factors of `f(x)`.

### How the current implementation seems needlessly complex

As we search through the factors of `f(x)`, we consider each linear factor `x - c`, as we should. But we seem to be checking not just whether `c == a`, but also whether `-c == a`:

https://github.com/sympy/sympy/blob/5d4477c8d2fcb3dc8c46fff3e604f7790ce22640/sympy/polys/numberfields/subfield.py#L140-L154

The only reason I can think of for doing this would be to try to speed up the case where `a` was a square root. For example, suppose `a == sqrt(2)` and `f(x) == x**2 - 2`. If `sqrt(2) in Q(b)`, then `f(x)` will split over `Q(b)` into the linear factors `x - sqrt(2)` and `x + sqrt(2)`. The existing code seems to be saying, "If we happened to get the linear factor `x + sqrt(2)` first, let's look ahead and try negating the constant term in case that gives the number we're looking for."

But I don't see this as an important optimization, because we were going to get the linear factor we needed on the very next iteration anyway!

More generally, this could anticipate any case where `f(x)` was an _even function_, but I don't think there's any strong motivation to do this.

On the flip side, I think this extra step is just going to _waste_ time in any case where `-a` is _not_ a conjugate of `a`, i.e. where `f(-a) != 0`. And you don't have to look far for such a case, e.g. `f(x) == x**3 - 2`.

**QUESTION:** Am I missing some reason why this extra step is valuable?

### How this function works

First, what is the PSLQ algorithm? I am no expert, but my understanding is that you pass it a list `[r1, r2, ..., rn]` of real numbers, and it tries to find an integer-linear combination of these numbers that is nearly equal to `0`.

So, how do we use PSLQ to try to write `a` as a polynomial in `b`?

0. The algebraic numbers `a` and `b` must be real. If not, give up immediately.
1. Let `f` and `g` be the minimal polynomials of the two algebraic numbers `a` and `b`, resp.
2. Let `n` and `m` be the degrees of `f` and `g` resp.
2. Let `A` and `B` be floating point evaluations of `a` and `b` resp.
3. Form the list of real numbers, `basis = [1, B, B**2, ..., B**(m-1), -A]`.
4. Ask the PSLQ algorithm for an integer-linear combination of `basis` that is (nearly) equal to zero.
5. This amounts to writing `A` _approximately_ as a polynomial in `B` with integer coeffs.
   Form this polynomial `h(x)` and now check whether in fact `h(b) == a` _exactly_, by asking whether `f(h) % g` is the zero polynomial.

### How the current implementation seems needlessly complex

Again the code plays around with +/- cases: we check not just whether `h(b) == a` but also whether `-h(b) == a`.

https://github.com/sympy/sympy/blob/5d4477c8d2fcb3dc8c46fff3e604f7790ce22640/sympy/polys/numberfields/subfield.py#L118-L129

I'm guessing the idea here may have been to try the same "optimization" (I don't think it is one!) as we just considered in the `field_isomorphism_factor()` function, but in this case I don't think it actually applies at all.

PSLQ has given us an integer-linear combination of `[1, B, B**2, ..., B**(m-1)]` which is approximately equal to `A`. The only reason to check the "negated case" would be if you believed this _same_ linear combination might actually be closer to `-A`. I think this would represent a serious flaw in the PSLQ algorithm. I don't know much about PSLQ, but I'd be surprised if it would ever return such a result.

I note that in fact line 127 quoted above has _no unit test coverage_, and I think it is probably unreachable, due to PSLQ _not_ suffering from the flaw it would take to get there.

So again I want to simplify the code by throwing out the +/- cases.

**QUESTION:** Is there any reason to keep the +/- cases here?

### LINKED ISSUE #22849: `evalf_table` missing entry for `AlgebraicNumber`
`sympy.core.evalf._create_evalf_table()` is missing an entry for `AlgebraicNumber`, causing

```python
>>> from sympy.core.evalf import evalf
>>> from sympy import AlgebraicNumber, sqrt
>>> a = -8 + 3 * sqrt(3)
>>> x = AlgebraicNumber(a)
>>> evalf(a, 1, {})
((1, mpz(1), 1, 1), None, 9, None)
>>> evalf(x, 1, {})
(None, None, None, None)
```

This is because `evalf(x, 1, {})` is falling back on "ordinary evalf", where cancelation occurs.

skieffer:

Fixes #22736 
Fixes #22849 

Improvements to `field_isomorphism()`:
* Speed: try to minimize precision for `evalf`. (See Issue #15743.)
* Code safety: use a non-global mpmath context.
* Simplifications: cut steps that consider +/- cases.
* Add explanatory comments.

#### Release Notes

* polys
  * Speed up `field_isomorphism()`.

### DIFF
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -1788,8 +1788,8 @@ def _evalf_with_bounded_error(x: 'Expr', eps: 'Expr' = None, m: int = 0,
     evalf
 
     """
+    eps = sympify(eps)
     if eps is not None:
-        eps = sympify(eps)
         if not (eps.is_Rational or eps.is_Float) or not eps > 0:
             raise ValueError("eps must be positive")
         r, _, _, _ = evalf(1/eps, 1, {})
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -1086,10 +1086,8 @@ def as_mpmath(x: Any, prec: int, options: OPT_DICT) -> tUnion[mpc, mpf]:
     if isinstance(x, NegativeInfinity):
         return mpf('-inf')
     # XXX
-    re, im, _, _ = evalf(x, prec, options)
-    if im:
-        return mpc(re or fzero, im)
-    return mpf(re)
+    result = evalf(x, prec, options)
+    return quad_to_mpmath(result)
 
 
 def do_integral(expr: 'Integral', prec: int, options: OPT_DICT) -> TMP_RES:
@@ -1557,6 +1555,21 @@ def evalf(x: 'Expr', prec: int, options: OPT_DICT) -> TMP_RES:
     return r
 
 
+def quad_to_mpmath(q):
+    """Turn the quad returned by ``evalf`` into an ``mpf`` or ``mpc``. """
+    if q is S.ComplexInfinity:
+        raise NotImplementedError
+    re, im, _, _ = q
+    if im:
+        if not re:
+            re = fzero
+        return make_mpc((re, im))
+    elif re:
+        return make_mpf(re)
+    else:
+        return make_mpf(fzero)
+
+
 class EvalfMixin:
     """Mixin class adding evalf capability."""
 
@@ -1706,17 +1719,7 @@ class EvalfMixin:
             return make_mpf(self._as_mpf_val(prec))
         try:
             result = evalf(self, prec, {})
-            if result is S.ComplexInfinity:
-                raise NotImplementedError
-            re, im, _, _ = result
-            if im:
-                if not re:
-                    re = fzero
-                return make_mpc((re, im))
-            elif re:
-                return make_mpf(re)
-            else:
-                return make_mpf(fzero)
+            return quad_to_mpmath(result)
         except NotImplementedError:
             v = self._eval_evalf(prec)
             if v is None:
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -44,7 +44,7 @@ if TYPE_CHECKING:
     from sympy.functions.elementary.integers import ceiling, floor
     from sympy.functions.elementary.trigonometric import atan
     from sympy.functions.combinatorial.numbers import bernoulli
-    from .numbers import Float, Rational, Integer
+    from .numbers import Float, Rational, Integer, AlgebraicNumber
 
 LG10 = math.log(10, 2)
 rnd = round_nearest
@@ -1055,6 +1055,10 @@ def evalf_bernoulli(expr: 'bernoulli', prec: int, options: OPT_DICT) -> TMP_RES:
         return None, None, None, None
     return b, None, prec, None
 
+
+def evalf_alg_num(a: 'AlgebraicNumber', prec: int, options: OPT_DICT) -> TMP_RES:
+    return evalf(a.to_root(), prec, options)
+
 #----------------------------------------------------------------------------#
 #                                                                            #
 #                            High-level operations                           #
@@ -1401,7 +1405,7 @@ def _create_evalf_table():
     from .add import Add
     from .mul import Mul
     from .numbers import Exp1, Float, Half, ImaginaryUnit, Integer, NaN, NegativeOne, One, Pi, Rational, \
-        Zero, ComplexInfinity
+        Zero, ComplexInfinity, AlgebraicNumber
     from .power import Pow
     from .symbol import Dummy, Symbol
     from sympy.functions.elementary.complexes import Abs, im, re
@@ -1450,6 +1454,7 @@ def _create_evalf_table():
         Piecewise: evalf_piecewise,
 
         bernoulli: evalf_bernoulli,
+        AlgebraicNumber: evalf_alg_num,
     }
 
 
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -1796,12 +1796,10 @@ def evalf_with_bounded_error(x: 'Expr', eps: 'Expr' = None, m: int = 0,
     # bitcount 1. Therefore 2**fastlog(c) = 2**(nc+1) = 2|c| is an upper bound
     # on |a|. Likewise for b and d.
     nr, ni = fastlog(c), fastlog(d)
-    if nr == MINUS_INF == ni:
-        n = 1
-    else:
-        n = max(nr, ni) + 1
-        # n-1 bits get us past the integer parts of a and b, and +1 accounts for
-        # the factor of <= sqrt(2) that is |x|/max(|a|, |b|).
+    n = max(nr, ni) + 1
+    # If x is 0, then n is MINUS_INF, and p will be 1. Otherwise,
+    # n - 1 bits get us past the integer parts of a and b, and +1 accounts for
+    # the factor of <= sqrt(2) that is |x|/max(|a|, |b|).
     p = max(1, m + n + 1)
 
     options = options or {}
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -1760,8 +1760,8 @@ def N(x, n=15, **options):
     return sympify(x, rational=True).evalf(n, **options)
 
 
-def evalf_with_bounded_error(x: 'Expr', eps: 'Expr' = None, m: int = 0,
-                             options: OPT_DICT = None) -> TMP_RES:
+def _evalf_with_bounded_error(x: 'Expr', eps: 'Expr' = None, m: int = 0,
+                              options: OPT_DICT = None) -> TMP_RES:
     """
     Evaluate *x* to within a bounded absolute error.
 
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -1784,7 +1784,7 @@ def evalf_with_bounded_error(x: 'Expr', eps: 'Expr' = None, m: int = 0,
 
     """
     if eps is not None:
-        if not eps.is_real or not eps > 0:
+        if not (eps.is_Rational or eps.is_Float) or not eps > 0:
             raise ValueError("eps must be positive")
         r, _, _, _ = evalf(1/eps, 1, {})
         m = fastlog(r)
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -149,6 +149,20 @@ def fastlog(x: Optional[MPF_TUP]) -> tUnion[int, Any]:
     return x[2] + x[3]
 
 
+def fastlog10_for_expr(v: 'Expr'):
+    """
+    Quickly compute a positive integer m such that 10^m > |v|, for any
+    :py:class:`~.Expr` *v*.
+    """
+    # Do `.evalf(1)` before taking `abs`, or else `abs` can be very slow.
+    # Taking `.evalf(1)` can make the value smaller in magnitude, but we make
+    # up for it by adding 1 at the end.
+    a = abs(v.evalf(1))
+    b = int(math.ceil(a)).bit_length()
+    c = math.ceil(b / LG10)
+    return c + 1
+
+
 def pure_complex(v: 'Expr', or_real=False) -> Optional[tTuple['Expr', 'Expr']]:
     """Return a and b if v matches a + I*b where b is not zero and
     a and b are Numbers, else None. If `or_real` is True then 0 will
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -149,20 +149,6 @@ def fastlog(x: Optional[MPF_TUP]) -> tUnion[int, Any]:
     return x[2] + x[3]
 
 
-def fastlog10_for_expr(v: 'Expr'):
-    """
-    Quickly compute a positive integer m such that 10^m > |v|, for any
-    :py:class:`~.Expr` *v*.
-    """
-    # Do `.evalf(1)` before taking `abs`, or else `abs` can be very slow.
-    # Taking `.evalf(1)` can make the value smaller in magnitude, but we make
-    # up for it by adding 1 at the end.
-    a = abs(v.evalf(1))
-    b = int(math.ceil(a)).bit_length()
-    c = math.ceil(b / LG10)
-    return c + 1
-
-
 def pure_complex(v: 'Expr', or_real=False) -> Optional[tTuple['Expr', 'Expr']]:
     """Return a and b if v matches a + I*b where b is not zero and
     a and b are Numbers, else None. If `or_real` is True then 0 will
@@ -1767,3 +1753,56 @@ def N(x, n=15, **options):
     # by using rational=True, any evaluation of a string
     # will be done using exact values for the Floats
     return sympify(x, rational=True).evalf(n, **options)
+
+
+def evalf_with_bounded_error(x: 'Expr', eps: 'Expr' = None, m: int = 0,
+                             options: OPT_DICT = None) -> TMP_RES:
+    """
+    Evaluate *x* to within a bounded absolute error.
+
+    Parameters
+    ==========
+
+    x : Expr
+        The quantity to be evaluated.
+    eps : Expr, None, optional (default=None)
+        Positive real upper bound on the acceptable error.
+    m : int, optional (default=0)
+        If *eps* is None, then use 2**(-m) as the upper bound on the error.
+    options: OPT_DICT
+        As in the ``evalf`` function.
+
+    Returns
+    =======
+
+    A tuple ``(re, im, re_acc, im_acc)``, as returned by ``evalf``.
+
+    See Also
+    ========
+
+    evalf
+
+    """
+    if eps is not None:
+        if not eps.is_real or not eps > 0:
+            raise ValueError("eps must be positive")
+        r, _, _, _ = evalf(1/eps, 1, {})
+        m = fastlog(r)
+
+    c, d, _, _ = evalf(x, 1, {})
+    # Note: If x = a + b*I, then |a| <= 2|c| and |b| <= 2|d|, with equality
+    # only in the zero case.
+    # If a is non-zero, then |c| = 2**nc for some integer nc, and c has
+    # bitcount 1. Therefore 2**fastlog(c) = 2**(nc+1) = 2|c| is an upper bound
+    # on |a|. Likewise for b and d.
+    nr, ni = fastlog(c), fastlog(d)
+    if nr == MINUS_INF == ni:
+        n = 1
+    else:
+        n = max(nr, ni) + 1
+        # n-1 bits get us past the integer parts of a and b, and +1 accounts for
+        # the factor of <= sqrt(2) that is |x|/max(|a|, |b|).
+    p = max(1, m + n + 1)
+
+    options = options or {}
+    return evalf(x, p, options)
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -1789,6 +1789,7 @@ def evalf_with_bounded_error(x: 'Expr', eps: 'Expr' = None, m: int = 0,
 
     """
     if eps is not None:
+        eps = sympify(eps)
         if not (eps.is_Rational or eps.is_Float) or not eps > 0:
             raise ValueError("eps must be positive")
         r, _, _, _ = evalf(1/eps, 1, {})
--- a/sympy/polys/densearith.py
+++ b/sympy/polys/densearith.py
@@ -1786,6 +1786,48 @@ def dmp_l1_norm(f, u, K):
     return sum([ dmp_l1_norm(c, v, K) for c in f ])
 
 
+def dup_l2_norm_squared(f, K):
+    """
+    Returns squared l2 norm of a polynomial in ``K[x]``.
+
+    Examples
+    ========
+
+    >>> from sympy.polys import ring, ZZ
+    >>> R, x = ring("x", ZZ)
+
+    >>> R.dup_l2_norm_squared(2*x**3 - 3*x**2 + 1)
+    14
+
+    """
+    if not f:
+        return K.zero
+    else:
+        return sum([ coeff**2 for coeff in f ])
+
+
+def dmp_l2_norm_squared(f, u, K):
+    """
+    Returns squared l2 norm of a polynomial in ``K[X]``.
+
+    Examples
+    ========
+
+    >>> from sympy.polys import ring, ZZ
+    >>> R, x,y = ring("x,y", ZZ)
+
+    >>> R.dmp_l2_norm_squared(2*x*y - x - 3)
+    14
+
+    """
+    if not u:
+        return dup_l2_norm_squared(f, K)
+
+    v = u - 1
+
+    return sum([ dmp_l2_norm_squared(c, v, K) for c in f ])
+
+
 def dup_expand(polys, K):
     """
     Multiply together several polynomials in ``K[x]``.
--- a/sympy/polys/densearith.py
+++ b/sympy/polys/densearith.py
@@ -1800,10 +1800,7 @@ def dup_l2_norm_squared(f, K):
     14
 
     """
-    if not f:
-        return K.zero
-    else:
-        return sum([ coeff**2 for coeff in f ])
+    return sum([coeff**2 for coeff in f], K.zero)
 
 
 def dmp_l2_norm_squared(f, u, K):
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -138,7 +138,6 @@ def field_isomorphism_pslq(a, b):
 def field_isomorphism_factor(a, b):
     """Construct field isomorphism via factorization. """
     _, factors = factor_list(a.minpoly, extension=b)
-    root_eq = None
     for f, _ in factors:
         if f.degree() == 1:
             # Any linear factor f(x) represents some conjugate of a in QQ(b).
@@ -152,9 +151,7 @@ def field_isomorphism_factor(a, b):
                 terms.append(coeff*b.root**(d - i))
             r = Add(*terms)
             # Check whether we got the number a
-            if root_eq is None:
-                root_eq = a.minpoly.root_equality_test()
-            if root_eq(r, a):
+            if a.minpoly.root_equality_test(r, a):
                 return coeffs
 
     # If none of the linear factors represented a in QQ(b), then in fact a is
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -97,7 +97,7 @@ def field_isomorphism_pslq(a, b):
         basis = [1, B] + [ B**i for i in range(2, m) ] + [-A]
 
         ctx.dps = n
-        coeffs = ctx.pslq(basis, maxcoeff=int(1e10), maxsteps=1000)
+        coeffs = ctx.pslq(basis, maxcoeff=10**10, maxsteps=1000)
 
         if coeffs is None:
             # PSLQ can't find an integer linear combination. Give up.
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -138,7 +138,7 @@ def field_isomorphism_pslq(a, b):
 def field_isomorphism_factor(a, b):
     """Construct field isomorphism via factorization. """
     _, factors = factor_list(a.minpoly, extension=b)
-    A = None
+    A, eps, evalf_ = None, None, None
     for f, _ in factors:
         if f.degree() == 1:
             # Any linear factor f(x) represents _some conjugate of_ a in QQ(b).
@@ -153,8 +153,9 @@ def field_isomorphism_factor(a, b):
             root = Add(*terms)
             # Check whether we got the number a
             if A is None:
-                A = a.root.evalf(chop=True)
-            if root.evalf(chop=True) == A:
+                eps, evalf_ = a.minpoly.root_comparison_tools()
+                A = evalf_(a)
+            if abs(evalf_(root) - A) < eps:
                 return coeffs
 
     # If none of the linear factors represented a in QQ(b), then in fact a is
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -138,10 +138,10 @@ def field_isomorphism_pslq(a, b):
 def field_isomorphism_factor(a, b):
     """Construct field isomorphism via factorization. """
     _, factors = factor_list(a.minpoly, extension=b)
-    A, eps, evalf_ = None, None, None
+    root_eq = None
     for f, _ in factors:
         if f.degree() == 1:
-            # Any linear factor f(x) represents _some conjugate of_ a in QQ(b).
+            # Any linear factor f(x) represents some conjugate of a in QQ(b).
             # We want to know whether this linear factor represents a itself.
             # Let f = x - c
             c = -f.rep.TC()
@@ -150,12 +150,11 @@ def field_isomorphism_factor(a, b):
             d, terms = len(coeffs) - 1, []
             for i, coeff in enumerate(coeffs):
                 terms.append(coeff*b.root**(d - i))
-            root = Add(*terms)
+            r = Add(*terms)
             # Check whether we got the number a
-            if A is None:
-                eps, evalf_ = a.minpoly.root_comparison_tools()
-                A = evalf_(a)
-            if abs(evalf_(root) - A) < eps:
+            if root_eq is None:
+                root_eq = a.minpoly.root_equality_test()
+            if root_eq(r, a):
                 return coeffs
 
     # If none of the linear factors represented a in QQ(b), then in fact a is
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -45,7 +45,7 @@ from sympy.polys.polyerrors import IsomorphismFailed
 from sympy.polys.polytools import Poly, PurePoly, factor_list
 from sympy.utilities import public
 
-from mpmath import pslq, mp
+from mpmath import MPContext
 
 
 def is_isomorphism_possible(a, b):
@@ -88,6 +88,7 @@ def field_isomorphism_pslq(a, b):
     g = b.minpoly.replace(f.gen)
 
     n, m, prev = 100, b.minpoly.degree(), None
+    ctx = MPContext()
 
     for i in range(1, 5):
         A = a.root.evalf(n)
@@ -95,12 +96,8 @@ def field_isomorphism_pslq(a, b):
 
         basis = [1, B] + [ B**i for i in range(2, m) ] + [-A]
 
-        dps = mp.dps
-        try:
-            mp.dps = n
-            coeffs = pslq(basis, maxcoeff=int(1e10), maxsteps=1000)
-        finally:
-            mp.dps = dps
+        ctx.dps = n
+        coeffs = ctx.pslq(basis, maxcoeff=int(1e10), maxsteps=1000)
 
         if coeffs is None:
             # PSLQ can't find an integer linear combination. Give up.
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -151,7 +151,7 @@ def field_isomorphism_factor(a, b):
                 terms.append(coeff*b.root**(d - i))
             r = Add(*terms)
             # Check whether we got the number a
-            if a.minpoly.root_equality_test(r, a):
+            if a.minpoly.same_root(r, a):
                 return coeffs
 
     # If none of the linear factors represented a in QQ(b), then in fact a is
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -49,7 +49,7 @@ from mpmath import pslq, mp
 
 
 def is_isomorphism_possible(a, b):
-    """Returns `True` if there is a chance for isomorphism. """
+    """Necessary but not sufficient test for isomorphism. """
     n = a.minpoly.degree()
     m = b.minpoly.degree()
 
@@ -93,40 +93,45 @@ def field_isomorphism_pslq(a, b):
         A = a.root.evalf(n)
         B = b.root.evalf(n)
 
-        basis = [1, B] + [ B**i for i in range(2, m) ] + [A]
+        basis = [1, B] + [ B**i for i in range(2, m) ] + [-A]
 
-        dps, mp.dps = mp.dps, n
-        coeffs = pslq(basis, maxcoeff=int(1e10), maxsteps=1000)
-        mp.dps = dps
+        dps = mp.dps
+        try:
+            mp.dps = n
+            coeffs = pslq(basis, maxcoeff=int(1e10), maxsteps=1000)
+        finally:
+            mp.dps = dps
 
         if coeffs is None:
+            # PSLQ can't find an integer linear combination. Give up.
             break
 
         if coeffs != prev:
             prev = coeffs
         else:
+            # Increasing precision didn't produce anything new. Give up.
             break
 
+        # We have
+        #   c0 + c1*B + c2*B^2 + ... + cm-1*B^(m-1) - cm*A ~ 0.
+        # So bring cm*A to the other side, and divide through by cm,
+        # for an approximate representation of A as a polynomial in B.
+        # (We know cm != 0 since `b.minpoly` is irreducible.)
         coeffs = [S(c)/coeffs[-1] for c in coeffs[:-1]]
 
+        # Throw away leading zeros.
         while not coeffs[-1]:
             coeffs.pop()
 
         coeffs = list(reversed(coeffs))
         h = Poly(coeffs, f.gen, domain='QQ')
 
+        # We only have A ~ h(B). We must check whether the relation is exact.
         if f.compose(h).rem(g).is_zero:
-            d, approx = len(coeffs) - 1, 0
-
-            for i, coeff in enumerate(coeffs):
-                approx += coeff*B**(d - i)
-
-            if A*approx < 0:
-                return [ -c for c in coeffs ]
-            else:
-                return coeffs
-        elif f.compose(-h).rem(g).is_zero:
-            return [ -c for c in coeffs ]
+            # Now we know that h(b) is in fact equal to _some conjugate of_ a.
+            # But from the very precise approximation A ~ h(B) we can assume
+            # the conjugate is a itself.
+            return coeffs
         else:
             n *= 2
 
@@ -136,23 +141,27 @@ def field_isomorphism_pslq(a, b):
 def field_isomorphism_factor(a, b):
     """Construct field isomorphism via factorization. """
     _, factors = factor_list(a.minpoly, extension=b)
-
+    A = None
     for f, _ in factors:
         if f.degree() == 1:
-            coeffs = f.rep.TC().to_sympy_list()
+            # Any linear factor f(x) represents _some conjugate of_ a in QQ(b).
+            # We want to know whether this linear factor represents a itself.
+            # Let f = x - c
+            c = -f.rep.TC()
+            # Write c as polynomial in b
+            coeffs = c.to_sympy_list()
             d, terms = len(coeffs) - 1, []
-
             for i, coeff in enumerate(coeffs):
                 terms.append(coeff*b.root**(d - i))
-
             root = Add(*terms)
-
-            if (a.root - root).evalf(chop=True) == 0:
+            # Check whether we got the number a
+            if A is None:
+                A = a.root.evalf(chop=True)
+            if root.evalf(chop=True) == A:
                 return coeffs
 
-            if (a.root + root).evalf(chop=True) == 0:
-                return [-c for c in coeffs]
-
+    # If none of the linear factors represented a in QQ(b), then in fact a is
+    # not an element of QQ(b).
     return None
 
 
--- a/sympy/polys/polytools.py
+++ b/sympy/polys/polytools.py
@@ -15,7 +15,7 @@ from sympy.core.evalf import (
     pure_complex, evalf, fastlog, evalf_with_bounded_error, quad_to_mpmath)
 from sympy.core.function import Derivative
 from sympy.core.mul import Mul, _keep_coeff
-from sympy.core.numbers import ilcm, I, Integer, Rational
+from sympy.core.numbers import ilcm, I, Integer
 from sympy.core.relational import Relational, Equality
 from sympy.core.sorting import ordered
 from sympy.core.symbol import Dummy, Symbol
--- a/sympy/polys/polytools.py
+++ b/sympy/polys/polytools.py
@@ -12,7 +12,7 @@ from sympy.core.basic import Basic
 from sympy.core.decorators import _sympifyit
 from sympy.core.exprtools import Factors, factor_nc, factor_terms
 from sympy.core.evalf import (
-    pure_complex, evalf, fastlog, evalf_with_bounded_error, quad_to_mpmath)
+    pure_complex, evalf, fastlog, _evalf_with_bounded_error, quad_to_mpmath)
 from sympy.core.function import Derivative
 from sympy.core.mul import Mul, _keep_coeff
 from sympy.core.numbers import ilcm, I, Integer
@@ -3817,7 +3817,7 @@ class Poly(Basic):
         # Then 2^n > 1/eps**2.
         m = math.ceil(n/2)
         # Then 2^(-m) < eps.
-        ev = lambda x: quad_to_mpmath(evalf_with_bounded_error(x, m=m))
+        ev = lambda x: quad_to_mpmath(_evalf_with_bounded_error(x, m=m))
 
         # Then for any complex numbers a, b we will have
         # |a - ev(a)| < eps and |b - ev(b)| < eps.
--- a/sympy/polys/polytools.py
+++ b/sympy/polys/polytools.py
@@ -3773,6 +3773,67 @@ class Poly(Basic):
 
         return r.replace(t, x)
 
+    def root_bounds(f):
+        """
+        Compute lower and upper bounds on the magnitude of the nonzero roots of
+        this polynomial.
+
+        Examples
+        ========
+
+        >>> from sympy import Poly
+        >>> from sympy.abc import x
+        >>> Poly(x**2 - 2).root_bounds()
+        (2/3, 3)
+        >>> (x * Poly(x**2 - 2)).root_bounds()
+        (2/3, 3)
+
+        Returns
+        =======
+
+        Pair ``(lb, ub)`` of :py:class:`~.Expr`
+            Gives the lower bound and upper bound. These bounds will be
+            rational if the domain of the polynomial is :ref:`ZZ` or :ref:`QQ`.
+
+        Raises
+        ======
+
+        DomainError
+            If the domain of the polynomial is not a subring of the complex
+            numbers.
+        MultivariatePolynomialError
+            If the polynomial is not univariate.
+        PolynomialError
+            If the polynomial is a monomial, since in such a case it has no
+            nonzero roots.
+
+        Notes
+        =====
+
+        The domain of the polynomial must be a subring of the complex numbers.
+        We use the `Cauchy bound`_.
+
+        .. _Cauchy bound: https://en.wikipedia.org/wiki/Geometrical_properties_of_polynomial_roots#Lagrange's_and_Cauchy's_bounds
+
+        """
+        dom = f.get_domain()
+        if dom.is_FiniteField or not dom.is_Numerical:
+            raise DomainError("Cannot compute root bounds over %s" % dom)
+        if f.is_multivariate:
+            raise MultivariatePolynomialError(
+                "Must be a univariate polynomial")
+        if f.is_monomial:
+            raise PolynomialError("Monomial has no nonzero roots")
+
+        coeffs = f.all_coeffs()
+        while coeffs[-1] == 0:
+            coeffs.pop()
+        lc, tc = coeffs[0], coeffs[-1]
+        ub = 1 + max(abs(n / lc) for n in coeffs[1:])
+        lb_inv = 1 + max(abs(n / tc) for n in coeffs[:-1])
+
+        return 1/lb_inv, ub
+
     def cancel(f, g, include=False):
         """
         Cancel common factors in a rational function ``f/g``.
--- a/sympy/polys/polytools.py
+++ b/sympy/polys/polytools.py
@@ -10,7 +10,7 @@ from sympy.core import (
 from sympy.core.basic import Basic
 from sympy.core.decorators import _sympifyit
 from sympy.core.exprtools import Factors, factor_nc, factor_terms
-from sympy.core.evalf import pure_complex
+from sympy.core.evalf import pure_complex, fastlog10_for_expr
 from sympy.core.function import Derivative
 from sympy.core.mul import Mul, _keep_coeff
 from sympy.core.numbers import ilcm, I, Integer, Rational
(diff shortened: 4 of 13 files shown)


sympy/polys/numberfields/subfield.py
```diff
@@ -93,40 +93,45 @@ def field_isomorphism_pslq(a, b):
...
 
-        dps, mp.dps = mp.dps, n
-        coeffs = pslq(basis, maxcoeff=int(1e10), maxsteps=1000)
-        mp.dps = dps
+        dps = mp.dps
+        try:
+            mp.dps = n
+            coeffs = pslq(basis, maxcoeff=int(1e10), maxsteps=1000)
+        finally:
+            mp.dps = dps
```
oscarbenjamin, line 103: It would be better to use a local context:
```python
In [30]: s2 = sqrt(2).evalf(200)

In [31]: from mpmath import mp

In [32]: ctx = mp.clone()

In [33]: ctx.dps = 100

In [34]: ctx.pslq([1, s2._mpf_, (s2**2)._mpf_])
Out[34]: [2, 0, -1]
```

sympy/polys/numberfields/subfield.py
```diff
@@ -136,23 +141,27 @@ def field_isomorphism_pslq(a, b):
...
             for i, coeff in enumerate(coeffs):
                 terms.append(coeff*b.root**(d - i))
-
             root = Add(*terms)
-
-            if (a.root - root).evalf(chop=True) == 0:
+            # Check whether we got the number a
+            if A is None:
+                A = a.root.evalf(chop=True)
+            if root.evalf(chop=True) == A:
```
oscarbenjamin, line 160: Checking `a.evalf() == b.evalf()` is not equivalent to checking `(a - b).evalf()` if the numbers differ by an amount smaller then the default precision of `evalf`:
```python
In [40]: a = 1 + Rational(1, 10**20)

In [41]: b = S(1)

In [42]: a.evalf() == b.evalf()
Out[42]: True

In [43]: (a - b).evalf()
Out[43]: 1.00000000000000e-20
```

oscarbenjamin: Maybe that's not a problem here though. We know that `a` and `b` are nonzero. Are they guaranteed to differ in sign or leading digit?

skieffer: Not necessarily sign or leading digit, but I am thinking that an isolating interval approach could be the way to go.

sympy/polys/numberfields/subfield.py
skieffer: Could we use `workdps`? As in,

```python
from mpmath import pslq, workdps
with workdps(n):
    coeffs = pslq(basis)
```

sympy/polys/numberfields/subfield.py
oscarbenjamin: Yes, that makes sense. What exactly does "conjugate" mean here? Does it mean that they have the same minimal polynomial?

skieffer: Yup, exactly that.

So, yeah, for each linear factor, we've found _some_ root of the minimal polynomial of `a`, and we just want to know: is this `a` itself? It does sound like a job for isolating intervals.

And I guess computing adequate isolating intervals can be faster than working with evalf.

oscarbenjamin: The root isolation code can definitely be improved so that refinement runs faster. The problem with evalf is that you need to isolate the root very closely but actually with a bounding interval if you know that `a` is a root of `p` then a bigger bounding interval is better than a smaller one:
```python
In [1]: r = RootOf(x**4 + x**3 + x**2 + x + 1, 0)

In [2]: r._get_interval()
Out[2]: (-1, -1/2) x (-1, -1/2)

In [3]: r.evalf()  # This is the slow part
Out[3]: -0.809016994374947 - 0.587785252292473⋅ⅈ

In [4]: r._get_interval()
Out[4]: (-932733090407273441/1152921504606846976, -1865466180814546881/2305843009213693952) x (-677670257458753271/1152921504606846976, -1355340514917506541/2305843009213693952)
```

sympy/polys/numberfields/subfield.py
```diff
@@ -93,40 +93,45 @@ def field_isomorphism_pslq(a, b):
         A = a.root.evalf(n)
         B = b.root.evalf(n)
 
-        basis = [1, B] + [ B**i for i in range(2, m) ] + [A]
```
jksuom, line 96: It is true, of course, that if `-a` is in `Q(b)` then so is `a` but I suspect that this line was the reason for the unexplained tests with `+/-a`. Clearly, `-a` should be used here to avoid confusion.

skieffer: True, I could see this having been overlooked, and then the cases further down were the result of trying to make it work.

sympy/polys/numberfields/subfield.py
skieffer: Question about `evalf`: Is there a good way to get the desired number of decimal places strictly to the right of the decimal point, without computing a log?

Example:

Consider `a = 1000000 + sqrt(2)`, and suppose I want 10 places to the right of the decimal point.
Obviously this doesn't work:
```python
>>> a.evalf(10)
1000001.414
```
I can do it this way:
```python
>>> a.evalf(10 + 1 + int(log(a, 10)))
1000001.4142135624
```
but I'm wondering if there might be some ready-made solution for this.

oscarbenjamin: I don't think that there is a ready-made solution although if you want this only for RootOf then there is probably a way to specify how small you want the bounding intervals to be.

Why exactly is this needed? There might be a different way to do this.

sympy/polys/numberfields/subfield.py
jksuom: It seems to me that `workdps` will do the same `try - finally` operation under the hood. So it could be used. Maybe it would even be preferable.

sympy/polys/numberfields/subfield.py
skieffer: Well, here's what I'm trying to do: Say I have a `Poly` `f` with root isolating intervals `I1`, `I2`, ..., `In`.
I have two `Expr`s `r` and `s`, and I know `f(r) == 0 == f(s)`. I want to determine whether `r == s`.

But I don't want to compute `(r - s).evalf()`, because it's very slow. (#15743)

My idea is to let `D` be the largest denominator occurring in any endpoint of any of the intervals `Ik`.
Then `abs(r - s) < 1/D` would suffice to conclude that `r == s`.

But to check the inequality requires use of `evalf`. To what precision? My idea is to detemine `n` large
enough to guarantee that `abs(r - r.evalf(n)) < eps = 1/(3*D)`. (And likewise for `s`.)

Then `abs(r.evalf(n) - s.evalf(n)) < eps` will suffice to conclude `r == s`.

So I was thinking of giving `Poly` a method like this:

```python
    def root_comparison_tools(f):
        roots = f.all_roots(radicals=False)
        ivals = [r._get_interval() for r in roots]
        # Let D0 be max denom among all interval endpoints.
        D0 = int(max(i.max_denom for i in ivals))
        # We need 3 times that.
        D1 = 3*D0
        # Pick n large enough to guarantee 10^n > D1.
        n = math.ceil(D1.bit_length() / 3.3) + 2
        # Make a custom evalf function that guarantees rounding error < 10^(-n).
        evalf_ = lambda x: x.evalf(n + 1 + math.ceil(int(x).bit_length() / 3.3))
        # Define a threshold that guarantees two roots of f are the same.
        eps = S(1) / D1
        return eps, evalf_
```

and then the problem could be solved like this:

```python
eps, evalf_ = f.root_compartison_tools()
if abs(evalf_(r) - evalf_(s)) < eps:
    # roots r and s are the same...
else:
    # they're different roots...

```

sympy/polys/numberfields/subfield.py
oscarbenjamin: Does `workdps` modify the `dps` for the global `mp` context? Is it thread-safe if another thread is using the `mp` context as well?

jksuom: `workdps` is defined in `mpmath/__init__.py` as `mp.workdps` where `mp = MPContext()`. I guess we could use ` with MPContext().workdps(n):` to get a temporary local context.

oscarbenjamin: Yes, that's better than using a clone of `mp` because a user might have altered the properties of the global `mp` context. I'm not sure we can use with and then use the global `mpmath.pslq` function though because I think that still gets its properties from the global context somehow. So I think it's:
```python
ctx = MPContext()
ctx.dps = n
coeffs = ctx.pslq(basis)
```
We shouldn't need to use `with` since this only modifies the properties of the throwaway `ctx`.

jksuom: > We shouldn't need to use with

Agreed. I also came to that conclusion.

sympy/polys/numberfields/subfield.py
oscarbenjamin: So `(r - s).evalf()` is slow if `r` and `s` are equal because it ends up computing `r` and `s` to high precision in an attempt to find the significant digits of `r - s`. What's faster is computing the significant digits of something that is *nonzero*.

For each interval `Ik` (including the correct one) we expect that `abs(r - Ik.mid) - Di` does not equal zero and we can compute that to find whether it is positive or not e.g.:
```python
In [51]: p = x**4 + x + 2

In [52]: rs = [RootOf(p, n) for n in range(degree(p))]

In [53]: r1, r2, r3, r4 = roots(p)

In [54]: intervals = [r._get_interval() for r in rs]

In [55]: intervals[0]
Out[55]: (-1, -1/2) x (-1, -1/2)

In [56]: intervals[0].center
Out[56]: (mpq(-3,4), mpq(-3,4))

In [57]: intervals[0].dx
Out[57]: mpq(1,2)

In [58]: intervals[0].dy
Out[58]: mpq(1,2)

In [59]: r1.evalf()
Out[59]: -0.849848427793055 - 0.6542723031575⋅ⅈ

In [60]: def centre(i):
    ...:     a, b = i.center
    ...:     return a + b*I
    ...: 

In [61]: def radius(i):
    ...:     return i.dx + i.dy
    ...: 

In [62]: centre(intervals[0])
Out[62]: 
  3   3⋅ⅈ
- ─ - ───
  4    4 

In [63]: radius(intervals[0])
Out[63]: mpq(1,1)

In [64]: (abs(r1 - centre(intervals[0])) - radius(intervals[0])).evalf(2)
Out[64]: -0.86

In [65]: (abs(r1 - centre(intervals[1])) - radius(intervals[1])).evalf(2)
Out[65]: 0.41

In [66]: (abs(r1 - centre(intervals[2])) - radius(intervals[2])).evalf(2)
Out[66]: 0.71

In [67]: (abs(r1 - centre(intervals[3])) - radius(intervals[3])).evalf(2)
Out[67]: 1.5
```
Here `Out[64]` through to `Out[67]` proves that `r1` is the root bounded by `intervals[0]`. If we end up having multiple negatives here then it means we need to shrink the intervals and try again. This is analogous to the `choose_factor` method in minpoly: we prove a factor is the minpoly for `a` by proving that every other factor is *not* the minpoly. This is done by numerically showing that something is *not* zero which can be done reliably (unlike proving that something is zero).

Note that we don't actually need to do the `evalf(2)` explicitly here as it is precisely what `a > b` does for Expr:
```python
In [69]: abs(r1 - centre(intervals[3])) > radius(intervals[3])
Out[69]: True
```

Combining this for `r` and `s` an algorithm could be:
1. For each interval compute `(r - c)**2 < d**2`
2. Also compute `(s - c)**2 < d**2`
3. If either of `r` or `s` is definitely not in this interval discard the interval.
4. If we have one remaining interval then `r = s`.
5. If we have no remaining intervals then `r != s`.
6. If there are multiple intervals remaining, shrink them and go back to step 1.

That is:
```python
while True:
    intervals_keep = []
    for interval in intervals:
         c = centre(interval)
         d = radius(interval)
         # Should be a way to compute abs(z)**2 without intermediate square roots
         if abs(r - c)**2 < d**2 and abs(s - c)**2 < d**2:
             intervals_keep.append(interval)
    if not intervals_keep:
        return False
    elif len(intervals_keep) == 1:
        return True
    else:
        intervals = [shrink_interval(interval) for interval in intervals_keep]
```

It might seem inefficient to recalculate the `evalf` for each root but at least in the case of RootOf the work done by one `evalf` is carried over to the next. It would also be possible to wrap other kinds of expression with something that caches the results of `_eval_evalf`.

We know that `abs(r - Ik.mid) - Di` will not be zero for any real root because that would imply that `r` was rational. It might potentially be zero though for a Gaussian rational like `2 + 3*I` but it's unlikely.

skieffer: Well, this is pretty cool. : )  I like the algorithm, and I like the principle that it's easy to check when something is _nonzero_, so work by process of elimination.

The note at the end about the outside possibility of an `evalf(0)` still happening is an interesting point. What if we used a transcendental radius, like `pi*(dx + dy)`. Then it's definitely different from `abs(r - c)` which is algebraic.

oscarbenjamin: Yes, that makes sense. It could be `pi/3*(dx + dy)` to make the bound reasonably tight.

oscarbenjamin: Or in fact we want `(dx + dy)/2` so `pi/6` should work.

skieffer: Sounds good. I am also still interested to see whether the algorithm I described could work as well. I added a note correcting the way the denominator `D` is chosen.

Maybe it's worth coding them both up. Even if only one is used, I think they both contain valuable ideas that could be worth revisiting at some point.

Meanwhile, maybe the comprehensive move over to ball-radius arithmetic you mentioned [before](https://github.com/sympy/sympy/issues/15743#issuecomment-985706352) will eventually make it all irrelevant?

It does seem that a fundamental problem with `evalf()` is that it wants to give you so many digits of precision, when all you may really want is so much accuracy, i.e. to bound the rounding error by `10^(-n)`.

oscarbenjamin: The algorithm I described above is quite similar to what `minpoly` does. Is it a bad idea to just compute `minpoly(r - s)`?

oscarbenjamin: Oh, I suppose that had to `(r - s).evalf()` as well as all the factoring.

oscarbenjamin:
The idea I had for improving the performance of RootOf refinement is basically the same as #19164 and is based on mpmath's interval arithmetic:
https://mpmath.org/doc/1.0.0/contexts.html#arbitrary-precision-interval-arithmetic-iv

Here's how it looks for refining bounds on an isolated real root:
```python
from sympy import *
from mpmath import MPIntervalContext


def qq_to_iv(num, iv):
    """Convert rational QQ to real MPF interval"""
    return iv.mpf(num.numerator) / iv.mpf(num.denominator)


def poly_to_iv(poly, iv):
    """Convert poly with QQ coefficients to MPF intervals"""
    return [qq_to_iv(coeff, iv) for coeff in poly]


def intersect_iv(X, Y, iv):
    a = max(X.a, Y.a)
    b = min(X.b, Y.b)
    if a <= b:
        return iv.mpf((a, b))
    else:
        return None


def is_proper_subset(X, Y):
    return X.a > Y.a and X.b < Y.b


def split_iv(X, iv):
    return iv.mpf((X.a, X.mid)), iv.mpf((X.mid, X.b))


def split_iv_zero(X, iv):
    assert 0 in X
    return iv.mpf((X.a, 0)), iv.mpf((0, X.b))


def horner(poly, x0):
    """Evaluate a poly

    >>> horner([1, 0, -1], 3)
    8
    """
    val = poly[0]
    for c in poly[1:]:
        val = c + val*x0
    return val


def interval_newton(poly, polydiff, X, prec=53):
    iv = MPIntervalContext()
    iv.dps = prec + 20

    poly_iv = poly_to_iv(poly, iv)
    polydiff_iv = poly_to_iv(polydiff, iv)
    X_iv = iv.mpf(X)

    queue = [X_iv]
    while queue:
        if len(queue) == 1:
            [Xj] = queue
            a, b = Xj.a, Xj.b
            if abs(a - b) * 2**(prec+3) < (abs(a) + abs(b)):
                return Float(Xj.mid, precision=prec)

        nextqueue = []

        for X1 in queue:
            x1 = X1.mid
            f = horner(poly_iv, x1)
            Fp = horner(polydiff_iv, X1)

            if 0 not in Fp:
                Z2 = x1 - f/Fp
                X2 = intersect_iv(Z2, X1, iv)

                if X2 is None:
                    pass # no roots
                elif is_proper_subset(Z2, X1):
                    # this interval is the one -> discard all others
                    nextqueue = [X2]
                    break
                elif 4*X2.delta < 3*X1.delta:
                    # converging -> continue
                    nextqueue.append(X2)
                else:
                    # insufficient progress -> split
                    nextqueue.extend(split_iv(X2, iv))

            else:
                # Split at the zero of the denominator
                Fp1, Fp2 = split_iv_zero(Fp, iv)
                X21 = intersect_iv(x1 - f/Fp1, X1, iv)
                X22 = intersect_iv(x1 - f/Fp2, X1, iv)
                X2s = [Xi for Xi in [X21, X22] if Xi is not None]
                nextqueue.extend(X2s)

        queue = nextqueue


def refine_root(p, a, b, dps=15):
    poly = Poly(p).rep.rep
    polydiff = Poly(p.diff()).rep.rep
    prec = dps_to_prec(dps)
    return interval_newton(poly, polydiff, (a, b), prec)


x = Symbol('x')
p = x**2 - 2
root = refine_root(p, 0, 1.5)
print(root)
```
With this you can refine a real root much faster than with the current `RootOf.evalf` method:
```python
In [3]: p = x**10 + x**4 - x**3 - 3

In [4]: %time ok = refine_root(p, 0, 2, 10000)
CPU times: user 304 ms, sys: 4 ms, total: 308 ms
Wall time: 307 ms

In [6]: %time ok2 = RootOf(p, 1).evalf(10000)
CPU times: user 1min 29s, sys: 16 ms, total: 1min 29s
Wall time: 1min 29s

In [9]: p = x**10 + x**8 + x**3 - 100

In [10]: %time ok = refine_root(p, 0, 2, 20)
CPU times: user 24 ms, sys: 0 ns, total: 24 ms
Wall time: 24.6 ms

In [11]: %time ok2 = RootOf(p, 1).evalf(20)
CPU times: user 48 ms, sys: 0 ns, total: 48 ms
Wall time: 47.4 ms

In [12]: ok == ok2
Out[12]: True
```
The code above only works for real roots but it can be extended to real systems of equations and therefore can also find the complex roots of a polynomial.

sympy/polys/numberfields/subfield.py
jksuom: The minimum distance between roots of a polynomial `p` can be estimated as follows. The roots of the resultant of `p(x)` and `p(x + y)` are the differences `x - x'` for all pairs of roots `(x, x')` of `p`.
```
In [1]: p = x**4 + x + 2                                                        

In [2]: q = p.subs(x, x + y)                                                    

In [3]: r = resultant(p, q, x)                                                  

In [4]: r                                                                       
Out[4]: 
 16       12       10        8        6         4
y   + 16⋅y   + 26⋅y   - 448⋅y  + 432⋅y  + 2021⋅y
```
After removing the zero roots, a lower bound for `|1/y|` can be found by applying e.g. [Cauchy's bound](https://en.wikipedia.org/wiki/Geometrical_properties_of_polynomial_roots#Lagrange's_and_Cauchy's_bounds) to the reversed polynomial. In the above case we get `|1/y| <= 1 + 448/2021`. The result can be sharpened by noting that the resultant is an even polynomial (for each `x - x'`, also `x' - x` is a root) hence a polynomial in `y**2` so that Cauchy's bound is also valid for `|1/y**2|`.

skieffer: Nice! So maybe the `Poly` class can have a method roughly like this:

```python
    def root_separation_lower_bound(f):
        """
        Compute a lower bound on the minimum distance between any two
        roots of this polynomial.
        """
        if f.is_multivariate:
            raise MultivariatePolynomialError(
                "must be a univariate polynomial")

        from sympy.functions.elementary.miscellaneous import sqrt

        x = f.gen
        y = Dummy('y')
        g = f.subs(x, x + y)
        h = Poly(resultant(f, g, x))
        # If f has roots ri, the roots of h are all the ri - rj.

        coeffs = h.rep.rep
        while coeffs[-1] == 0:
            coeffs.pop()
        # Use the Cauchy bound
        d = coeffs.pop()
        M = 1 + max(abs(n / d) for n in coeffs)
        # Since h is an even func., can improve bound with sqrt.
        return 1/sqrt(M)
```

oscarbenjamin: Yes, and maybe there could also be a method just called `roots_bound` that just gives root bounds in general. Using `sqrt` here is tricky. We ideally want the bounds to be rational numbers. It might be better to have something that returns the square of the bound(s).

Also `rep.rep` should not be manipulated directly as it is a private internal attribute of the immutable `Poly`. There is `all_coeffs` to get the coefficients as `Expr`.

Also any methods like this should check the domain and whether the `Poly` is univariate.

jksuom: Maybe `isqrt` could be used. Something like `Rational(isqrt(M.p*M.q) + 1, M.q)`.

skieffer: Finally was able to get back to this...

On making the bounds rational: I've written it so that the bounds will be rational if the polynomial's coefficients are. Do you think this is good enough?

skieffer: @jksuom Is it all right if I set you as first author of 9c91d52a6f? These are your ideas.

skieffer:
Some good news: Test `sympy.polys.numberfields.tests.test_subfield.test_issue_22736()` was taking over 9 seconds before these changes, and now takes only about 1/4 of a second!

sympy/core/evalf.py
```diff
@@ -149,6 +149,20 @@ def fastlog(x: Optional[MPF_TUP]) -> tUnion[int, Any]:
...
+    """
+    Quickly compute a positive integer m such that 10^m > |v|, for any
+    :py:class:`~.Expr` *v*.
+    """
+    # Do `.evalf(1)` before taking `abs`, or else `abs` can be very slow.
+    # Taking `.evalf(1)` can make the value smaller in magnitude, but we make
+    # up for it by adding 1 at the end.
+    a = abs(v.evalf(1))
+    b = int(math.ceil(a)).bit_length()
+    c = math.ceil(b / LG10)
```
oscarbenjamin, line 162: It looks like `math.ceil(abs(v))` will call `evalf` implicitly:
```python
In [9]: x = 10**100*pi

In [10]: math.ceil(x).bit_length()
Out[10]: 334
```
Why do we need `fastlog10` instead of `fastlog` above which uses base 2?

sympy/polys/numberfields/subfield.py
```diff
@@ -88,45 +88,47 @@ def field_isomorphism_pslq(a, b):
...
+        basis = [1, B] + [ B**i for i in range(2, m) ] + [-A]
 
-        dps, mp.dps = mp.dps, n
-        coeffs = pslq(basis, maxcoeff=int(1e10), maxsteps=1000)
-        mp.dps = dps
+        ctx.dps = n
+        coeffs = ctx.pslq(basis, maxcoeff=int(1e10), maxsteps=1000)
```
oscarbenjamin, line 100: I would write `int(1e10)` as `10**10`

sympy/polys/polytools.py
```diff
@@ -3773,6 +3774,178 @@ def nth_power_roots_poly(f, n):
...
+        # If f has roots ri, the roots of h are all the ri - rj.
+
+        if h.is_monomial:
+            # In this case all roots of f are the same.
+            return Integer(0)
+
+        lb, _ = h.root_bounds()
+
+        # Since h is an even function, can improve bound with square root.
+        if lb.is_rational:
```
oscarbenjamin, line 3893: This should probably be `is_Rational` rather than `is_rational`. Note the difference:
```python
In [4]: a = S(3)

In [5]: b = Symbol('b', rational=True)

In [6]: a
Out[6]: 3

In [7]: b
Out[7]: b

In [8]: a.is_rational
Out[8]: True

In [9]: a.is_Rational
Out[9]: True

In [10]: b.is_rational
Out[10]: True

In [11]: b.is_Rational
Out[11]: False
```
The test `is_Rational` is structural: is this an explicit rational number. The test `is_rational` is a *semantic* query through the assumptions system:

The polys module is mostly concerned with structural considerations rather than semantic in this sense.

sympy/polys/polytools.py
```diff
@@ -3773,6 +3774,178 @@ def nth_power_roots_poly(f, n):
...
+            # Preserve rational bound
+            return Rational(isqrt(lb.p*lb.q) + 1, lb.q)
+        else:
+            from sympy.functions.elementary.miscellaneous import sqrt
+            return sqrt(lb)
```
oscarbenjamin, line 3898: If the function just returns the square of the bound then the shenanigans here are not needed and the return value will always be within the field generated by the coefficients.

sympy/polys/polytools.py
```diff
@@ -3773,6 +3774,178 @@ def nth_power_roots_poly(f, n):
...
+                "Must be a univariate polynomial")
+        if f.is_monomial:
+            raise PolynomialError("Monomial has no nonzero roots")
+
+        coeffs = f.all_coeffs()
+        while coeffs[-1] == 0:
+            coeffs.pop()
+        lc, tc = coeffs[0], coeffs[-1]
+        ub = 1 + max(abs(n / lc) for n in coeffs[1:])
+        lb_inv = 1 + max(abs(n / tc) for n in coeffs[:-1])
```
oscarbenjamin, line 3834: It does seem somewhat backwards to use Expr here. Can we not convert to a field and compute this with domain elements e.g.:
```python
In [24]: p = Poly(x**2 + 2)

In [25]: p = p.to_field()

In [26]: coeffs = p.rep.rep

In [27]: result = 1 + max(abs(n/coeffs[0]) for n in coeffs[1:])

In [28]: result
Out[28]: mpq(3,1)

In [29]: K.to_sympy(result)
Out[29]: 3
```

sympy/polys/polytools.py
```diff
@@ -3773,6 +3774,178 @@ def nth_power_roots_poly(f, n):
...
+        eps = delta / 3
+
+        # Pick n large enough to guarantee 10^n > 1/eps.
+        L = fastlog10_for_expr
+        n = L(1/eps)
+
+        # Make a custom evalf function that guarantees rounding error < 10^(-n).
+        # This means we want more than n places to the right of the decimal
+        # point. `L(x)` gets us past any places to the left of the point.
+        evalf_ = lambda x: x.evalf(L(x) + n + 1)
```
oscarbenjamin, line 3943: Is this the reason for needing fastlog10?

Maybe `dps_to_prec` could be used here to work in binary precision.

sympy/core/evalf.py
skieffer: The `.evalf(1)` makes it faster, and is good enough for the approximation we want here.

```python
f = cyclotomic_poly(5)
v = CRootOf(f, 0)
timeit.timeit('v._reset(); math.ceil(abs(v))', number=10, globals=globals())
Out[7]: 10.512659683999999
timeit.timeit('v._reset(); math.ceil(abs(v.evalf(1)))', number=10, globals=globals())
Out[8]: 0.9664351640000035
```

skieffer: On `fastlog10` versus `fastlog`: I discuss below, where it comes up again.

sympy/polys/polytools.py
skieffer: Yes, this was part of the reason. Also I wanted a function that would accept any `Expr`, instead of the `MPF_TUP` that's required by `fastlog`.

Maybe, if we try to work with domain elements instead of `Expr`, as you discuss above, this could be reconsidered.

sympy/polys/polytools.py
skieffer: Ah, thanks for the explanation. i was wondering about this.

sympy/polys/polytools.py
oscarbenjamin: The `evalf` function in the `sympy.core.evalf` module takes a binary precision but otherwise does more or less the same as the `Expr.evalf` method.

The mpf tuple is easy to obtain:
```python
In [1]: pi.evalf(1)._mpf_
Out[1]: (0, mpz(25), -3, 5)
```

sympy/polys/polytools.py
skieffer: In other words, you're saying that sure the square root sharpens the bound, but who cares. Maybe it's not worth the extra calculation. I suppose that may be true in many cases.

Maybe the method should take a keyword arg, `take_sqrt=False` so those who care can activate it.

oscarbenjamin: The sharp bound is good. It's just that I can have a function that returns the bound or a function that returns the square of the bound and each is equally usable with the same sharpness:
```python
if abs(value) > bound():
    # do stuff

if value ** 2 > bound_squared():
    # do the same stuff
```
The difference between these two cases is that in the latter I can do all calculations within the same field e.g. if `bound_squared()` is guaranteed to return a rational number and value is a rational number then the check `value ** 2 > bound_squared()` is pure rational arithmetic and does not require computing in an extension field or using floating point.

In the complex case even `abs(value)` involves a square root but `abs(value)**2` is `re(value)**2 + im(value)**2` so if both parts are rational then the result is rational.

skieffer: Okay, those are some excellent points. I suppose as long as the method is called something like `root_separation_lower_bound_squared()`, users will know what they're getting. Also the examples in the method's docstring can show the usage patterns you just described.

sympy/polys/numberfields/subfield.py
jksuom: It is all right, though I wasn't expecting that.

sympy/polys/numberfields/subfield.py
```diff
@@ -88,45 +88,47 @@ def field_isomorphism_pslq(a, b):
     g = b.minpoly.replace(f.gen)
 
     n, m, prev = 100, b.minpoly.degree(), None
+    ctx = MPContext()
```
jksuom, line 91: Local context is a good idea but it would probably be better to use `PythonMPContext`. There are unresolved issues in the implementation of `MPContext` (see #19153). There is a single static `RealElement` (and `ComplexElement`) class whose `context` attribute is patched to point to the new `MPContext` class each time one is created. So all computations in existing `RR` and `CC` classes will always have the precision set in the last `MPContext`.

sympy/polys/polytools.py
```diff
@@ -3773,6 +3774,178 @@ def nth_power_roots_poly(f, n):
...
+        else:
+            from sympy.functions.elementary.miscellaneous import sqrt
+            return sqrt(lb)
+
+    def root_comparison_tools(f):
+        """
+        Produce a threshold ``eps`` and custom evalf function ``evalf_`` such
+        that two roots ``a``, ``b`` of this polynomial must be equal if
+        ``abs(evalf_(a) - evalf_(b)) < eps``.
+
```
skieffer, line 3905: I ran some tests, and

```python
A = evalf_(a)
for b in roots:
    abs(A - evalf_(b)) < eps
```

did not appear to be any faster than

```python
for b in roots:
    abs(evalf_(a) - evalf_(b)) < eps
```

(Due to caching, I suppose?)

So maybe this method can be simplified. It could be called `root_equality_test()`, and could simply return a binary function like

```python
lambda a, b: abs(evalf_(a) - evalf_(b)) < eps
```

(or one that incorporates the squared bound idea).

oscarbenjamin: A RootOf caches its internal interval so repeated calls to evalf are immediate:
```python
In [3]: r = RootOf(x**3 + x + 1, 0)

In [4]: r
Out[4]: 
       ⎛ 3           ⎞
CRootOf⎝x  + x + 1, 0⎠

In [5]: %time r.evalf()
CPU times: user 20 ms, sys: 0 ns, total: 20 ms
Wall time: 19 ms
Out[5]: -0.682327803828019

In [6]: %time r.evalf()
CPU times: user 0 ns, sys: 0 ns, total: 0 ns
Wall time: 878 µs
Out[6]: -0.682327803828019
```
Internally `evalf` will try increasing the precision on subterms until it considers that the overall result has the accuracy it wants. So when you have `(a - b).evalf()` if `a = b` then `evalf` will increase the precision again and again until it hits the limit. Each time it does that RootOf needs to shrink its interval so we end up needing a much smaller interval then we would for `a.evalf()` or even `(a.evalf() - b).evalf()`. This is the part that is slowest.

These caches will affect your timings if you do everything in the same process. Instead make a complete Python script and time the whole operation:
```python
from sympy import *

x = Symbol('x')
p = x**4 + x**3 + x**2 + x + 1
rs = roots(p)
r3 = RootOf(p, 3)

def f():
    return [abs(r3 - r).evalf() for r in rs]

def g():
    r3e = r3.evalf()
    return [abs(r3e - r).evalf() for r in rs]

if True:
    f()
else:
    g()
```
That script will call `f` and the timing we get is:
```console
$ time python r.py

real	0m36.530s
user	0m36.496s
sys	0m0.024s
```
Now instead let's call `g()` (change the True to False):
```console
$ time python r.py

real	0m1.505s
user	0m1.464s
sys	0m0.040s
```
So `f` is a lot slower than `g` and this slowness is absolutely on a scale that is noticeable to a user.

If you look at the outputs of `f` and `g` you can also see that they don't get the same result:
```python
In [3]: f()
Out[3]: 
[1.90211303259031 - 0.e-25⋅ⅈ, 1.17557050458495 - 0.e-25⋅ⅈ, 1.90211303259031 + 0.e-22⋅ⅈ, 2.952700857
5875e-64 - 1.25078415499626e-63⋅ⅈ]

In [4]: g()
Out[4]: [1.90211303259031, 1.17557050458495, 1.90211303259031, 7.81510642993629e-9]
```
The absolute difference computed by `f` is smaller by 50 orders of magnitude.

Personally I think that both `f` and `g` are too slow. What needs to be improved for performance here is the algorithm that shrinks the intervals.

skieffer: Thanks. I didn't realize separate processes should be called in order to get a good comparison.

In the two forms I want to consider, the `evalf` always happens before the subtraction.
So, maybe the fact about caching already settles the question.
But just to be sure I add functions `h` and `k` to the script:

```python
import sys
from sympy import *

x = Symbol('x')
p = x**4 + x**3 + x**2 + x + 1
rs = roots(p)
r3 = RootOf(p, 3)

def f():
    return [abs(r3 - r).evalf() for r in rs]

def g():
    r3e = r3.evalf()
    return [abs(r3e - r).evalf() for r in rs]

def h():
    return [abs(r3.evalf() - r.evalf()) for r in rs]

def k():
    r3e = r3.evalf()
    return [abs(r3e - r.evalf()) for r in rs]

funcs = [f, g, h, k]

F = funcs[int(sys.argv[1])]
F()
```

and we get:

```bash
$ time python r.py 0

real    0m18.880s
user    0m18.381s
sys 0m0.153s

$ time python r.py 1

real    0m1.000s
user    0m0.937s
sys 0m0.056s

$ time python r.py 2

real    0m0.960s
user    0m0.891s
sys 0m0.052s

$ time python r.py 3

real    0m0.974s
user    0m0.895s
sys 0m0.057s
```

So I do think the `root_comparison_tools()` method I wrote could be simplified. I made it return the pair `(eps, evalf_)` because I imagined it might offer users an important performance improvement to let them use form `k()`. But, if in fact form `h()` is just as good, then it can be a simpler method that returns a single test function

```python
lambda a, b: abs(evalf_(a) - evalf_(b)) < eps
```

> Personally I think that both f and g are too slow. What needs to be improved for performance here is the algorithm that shrinks the intervals.

To me it depends on the application. For certain web apps, clicking and waiting one second to see a result might be just fine, whereas waiting 10s starts to feel unresponsive.

sympy/polys/polytools.py
skieffer: Actually running into an interesting issue here, depending on the domain of the `Poly`, since `abs()` is not supported for domain elements of types `ANP` or `GuassianElement`.

What do you suppose is the right way to handle this? Should these domain types define an `__abs__()` method?

For example, I tried giving `GaussianRational` this method:

```python
    def __abs__(self):
        from sympy.polys.domains.complexfield import CC
        z = CC.from_GaussianRationalField(self, self.base)
        return abs(z)
```

which does solve the immediate problem. But unfortunately importing `CC` at the top of the module causes a cyclic import, and importing it here seems a bit ugly. Is there a better way?

oscarbenjamin: I still think that we should try to avoid introducing square roots if it's possible.

Maybe it would be better to give a domain method like `K.abs2` which can compute `abs(z)**2` since I think it should always be possible to compute that within the same ring:
```python
def abs2(K, z):
    return z.x**2 + z.y**2
```
For ANP it could be done if there was a way to compute complex conjugates...

skieffer: Yeah, if it's important to stay in the Poly's domain, then I think there will be trouble with `ANP` (e.g. the fields generated by the cube roots of 2 -- the universal counterexample!)....

Also if `Poly.root_bounds()` returns squared bounds, then the `Poly.root_separation_...` method will return the 4th power of the bound, and I feel like it's getting out of hand...

I'm thinking I'd really like to retreat a bit here. How about if `Poly.root_bounds()` only supports polys over `QQ` and `ZZ`. That's all that was originally called for. I was trying to support all domains contained in `CC` because it seemed easy, but it no longer does. If someone needs that at some point, let them open an issue at that time.

sympy/polys/numberfields/subfield.py
skieffer: Is this a problem with the `MPContext` imported directly from `mpmath`, or just with the one defined [in SymPy](https://github.com/sympy/sympy/blob/d822fcba181155b85ff2b29fe525adbafb22b448/sympy/polys/domains/mpelements.py#L45)?

I tried using `PythonMPContext`, but it has no `.pslq` attribute.

sympy/core/evalf.py
```diff
@@ -1750,3 +1753,56 @@ def N(x, n=15, **options):
...
+    A tuple ``(re, im, re_acc, im_acc)``, as returned by ``evalf``.
+
+    See Also
+    ========
+
+    evalf
+
+    """
+    if eps is not None:
+        if not eps.is_real or not eps > 0:
```
skieffer, line 1787: Was wondering if there's an `is_Real` versus `is_real` consideration here. But `is_Real` doesn't seem to be a thing?

skieffer:
## Summary of recent changes

* Dumped the `fastlog10_for_expr()` function I had written in `core.evalf`. Instead, we now have `evalf_with_bounded_error()`. I couldn't find any existing function in either `core.evalf` or `mpmath` that would let you evalf up until the _absolute error_ was less than a given bound. So that's what this function does.

* The new `Poly` methods:

    - `root_bounds()` -- and as a consequence, the other two methods as well -- now only accepts polys over `QQ` or `ZZ`.
    - `root_separation_lower_bound_squared()` does just that: it returns the _square_ of the Cauchy lower bound on the
       separation between distinct roots of this polynomial.
    - `root_equality_test()` returns a binary function `eq(a, b)` such that, if `a` and `b` are complex numbers known to be roots of this polynomial, then `eq(a, b)` is `True` if and only if they are the _same_ root.

* `field_isomorphism_factor()` now uses `root_equality_test()` and as a consequence is a lot faster!

sympy/core/evalf.py
skieffer: Now I see that a `mypy` test is failing due to an `Expr` not having an `is_real` attribute. So, what's the right way to check whether an arbitrary `Expr` represents a real number?

sympy/polys/numberfields/subfield.py
oscarbenjamin: In mpmath it's added through the IdentificationMethods mixin class:
https://github.com/fredrik-johansson/mpmath/blob/c11db84b3237bd8fc6721f5a0c5d7c0c98a24dc1/mpmath/identification.py#L837
That class is then a superclass for StandardBaseContext. Finally MPContext subclasses both BaseMPContext and StandardBaseContext:
https://github.com/fredrik-johansson/mpmath/blob/c11db84b3237bd8fc6721f5a0c5d7c0c98a24dc1/mpmath/ctx_mp.py#L58
No idea why it's all done in such a complicated way.

sympy/core/evalf.py
oscarbenjamin: Every Expr has an `is_real` attribute so maybe mypy just needs to be told that. Something like this:
https://github.com/sympy/sympy/blob/d822fcba181155b85ff2b29fe525adbafb22b448/sympy/core/basic.py#L84-L85

In fact every Basic has `is_real` but I think that really it should only be defined for Expr (#19474).

Note though that `is_real` is a *semantic* check (e.g. `Symbol('x', real=True).is_real`) which is not what is wanted here. Here I think that what is wanted is for `eps` to be an explicit elementary number like `Float/Rational`.

skieffer: Okay, thanks. I've updated the type checking.

sympy/polys/numberfields/subfield.py
skieffer: I think we might be okay. Calling `field_isomorphism_pslq()` does not appear to be altering the context of a `RealElement`:

```python
In [2]: from sympy import *
   ...: from sympy.polys.numberfields.subfield import field_isomorphism_pslq
   ...: a = RR(2)
   ...: print(a.context.prec)
   ...: b = AlgebraicNumber(sqrt(2))
   ...: c = AlgebraicNumber(sqrt(2) + sqrt(3))
   ...: print(field_isomorphism_pslq(b, c))
   ...: print(a.context.prec)
53
[1/2, 0, -9/2, 0]
53
```

skieffer:
The simplification I just introduced in `evalf_with_bounded_error()` happened to expose a really surprising issue. It should be solved by #22850.

skieffer:
This PR now includes a solution to #22849 as well as #22736, and I think it's ready from my point of view.

For reviewers: There are three different things happening in the file `sympy/core/evalf.py`:

* Introduces `quad_to_mpmath()` to do some refactoring of pre-existing code.
* Introduces `evalf_with_bounded_error()`.
* Adds the `AlgebraicNumber` class to the evalf table.

I think the changes in all the other files are more or less straightforward.

sympy/core/evalf.py
```diff
@@ -1750,3 +1758,54 @@ def N(x, n=15, **options):
...
+    A tuple ``(re, im, re_acc, im_acc)``, as returned by ``evalf``.
+
+    See Also
+    ========
+
+    evalf
+
+    """
+    if eps is not None:
+        if not (eps.is_Rational or eps.is_Float) or not eps > 0:
```
oscarbenjamin, line 1792: Maybe `eps` should be sympified.

sympy/polys/polytools.py
```diff
@@ -3773,6 +3775,174 @@ def nth_power_roots_poly(f, n):
 
         return r.replace(t, x)
 
+    def root_bounds(f):
+        """
+        Compute lower and upper bounds on the magnitude of the nonzero roots of
+        this polynomial.
```
oscarbenjamin, line 3781: I think it would be good to clarify in this docstring somewhere that we are thinking about potentially non-real roots.

sympy/polys/polytools.py
```diff
@@ -3773,6 +3775,174 @@ def nth_power_roots_poly(f, n):
...
+
+        # Then for any complex numbers a, b we will have
+        # |a - ev(a)| < eps and |b - ev(b)| < eps.
+        # So if |ev(a) - ev(b)|**2 < eps**2, then
+        # |ev(a) - ev(b)| < eps, hence |a - b| < 3*eps = delta.
+        def eq_test(a, b):
+            A, B = ev(a), ev(b)
+            return (A.real - B.real)**2 + (A.imag - B.imag)**2 < eps_sq
+
+        return eq_test
```
oscarbenjamin, line 3944: Returning the function like this seems like an odd interface

skieffer: I guess since the `Poly` is considered immutable it could just store the test function as an attribute, and return only the True/False answer to the user.

skieffer: But now that I take a closer look, I see that the `Poly` class is actually really minimal in its instance attributes. It only has `rep` and `gens`, and it declares these as its slots `__slots__ = ('rep', 'gens')`. I'm not sure I want to disturb this....

oscarbenjamin: Is there much value in storing anything?

oscarbenjamin: If there is then it could be possible to have a cached function that returns the `eq_test` but still expose an interface to the user that looks more natural like `p.same_roots(r1, r2)`.

skieffer: > Is there much value in storing anything?

I was running some timing tests in order to try to answer this, when I realized it's not scaling as well as I had hoped. I should have done these scaling tests much earlier!

I want to try `dup_zz_mignotte_bound` and see if it will do better. ([Mignotte bound](https://en.wikipedia.org/wiki/Geometrical_properties_of_polynomial_roots#Root_separation))

skieffer: Couple of findings: (a) `dup_zz_mignotte_bound` is actually a different Mignotte bound, but anyway (b) the Mignotte bound I meant isn't doing much better than the bound we already have.

Might have to think some more...

jksuom: Mignotte bound is currently misnamed.

oscarbenjamin: What exactly is slow?

skieffer: > What exactly is slow?

I'm looking at cyclotomic polynomials `phi_n`. I didn't anticipate how fast the coefficients of `Res(phi_n(x), phi_n(x + y))` would grow with `n`.

Of course for cyclotomic polys the geometry is easy, so better separation bounds could be used.

Anyway, the comparison with Mignotte is interesting. Our bound starts out several orders of magnitude better, but then gets worse. As for the run time, we're slowing down dramatically but Mignotte barely changes at all.


```python
f = lambda n: Poly(cyclotomic_poly(prime(n)))

def mignotte(f):
    n = f.degree()
    D = f.discriminant()
    eucl_norm = sqrt( sum( [cf**2 for cf in f.all_coeffs()] ) )
    denom = sqrt(n)**(n+1) * eucl_norm**(n-1)
    num = sqrt(3*abs(D))
    return num/denom

for n in range(2, 11):
    t = time.time()
    b1 = sqrt(f(n).root_separation_lower_bound_squared()).evalf()
    t1 = time.time() - t

    t = time.time()
    b2 = mignotte(f(n)).evalf()
    t2 = time.time() - t
    print(f'{prime(n)}:\n{b1}, {b2}')
    print(f'{t1:.5f}, {t2:.5f}\n')

3:
0.866025403784439, 0.612372435695794
0.00404, 0.00108

5:
0.845154254728517, 0.0541265877365274
0.00793, 0.00131

7:
0.509175077217316, 0.00327364250549328
0.01933, 0.00160

11:
0.00671827668354308, 0.00000547722557505166
0.11216, 0.00210

13:
0.000155324265551629, 1.67448988340192E-7
0.28852, 0.00230

17:
2.09849711164693E-9, 1.00818626091867E-10
1.15808, 0.00292

19:
1.82097216957742E-12, 2.05812538301822E-12
2.17408, 0.00345

23:
2.41307054558417E-20, 6.31974853510620E-16
6.41855, 0.00444

29:
4.75406277160334E-36, 1.79791507099308E-21
24.15388, 0.00471

```

skieffer:
"Root identification" is now looking to me like a whole subject of its own, and we should wrap up this PR without trying to settle that subject too. This PR contributes a little to that subject, and does several other things as well.


## Wrapping up

I think the last question we were considering was whether it was important for `Poly.root_equality_test()` to cache the test function. With the time costs involved in other parts of the process, I don't think this matters.


## Thoughts for future work on Root Identification

* References
    - [Mignotte 1982](http://people.dm.unipi.it/gianni/AC-EAG/Mignotte.pdf)
    - [Yap 2000](http://160592857366.free.fr/joe/ebooks/ShareData/Fundamental%20Problems%20in%20Algorithmic%20Algebra%201993%20Yap.pdf), Lecture VI, Sec. 7

* I am convinced of the necessity to improve the performance of `RootOf` refinement, which @oscarbenjamin has pointed to several times.

* There are several ways to compute lower bounds on root separations.

    - the Mignotte bound
    - the resultant method, which outperforms Mignotte up to a point, but then not
    - special cases, e.g. separation of `2*sin(pi/n)` for `cyclotomic_poly(n)`

* I still think the root intervals themselves can produce a better separation bound, albeit with a time tradeoff. E.g. on `f = cyclotomic_poly(29)`:
    - Mignotte is done in under 5ms, but the lower bound is `1.8E-21`, and refining roots of `f` to 21 places takes 27.4s.
    - Calling `f.intervals()` takes 1.6s, but the lower bound is `0.125`, and the root refinement takes 8.12s. Still not good, but better overall.

oscarbenjamin:
The parts of this PR that I'm not sure about is where there are changes to public interfaces. Otherwise I would be happy to just merge as is with future improvements deferred for now.

One examples is the `evalf_with_bounded_error` function. It isn't clear if this is a public or a private function: are we committing to supporting that function as used by users of SymPy in future?

From an API design perspective I would have thought that it would make more sense to make bounded error an option to `evalf` so that the interface is more like:
```
expr.evalf(20, absolute_bound=50)
```
Which would mean something like "compute the expression with a relative error of 1e-20 and an absolute error of 1e-50". Perhaps instead the bound could be given as an explicit number like `1e-50`.

In any case it's clear if you look at the `.evalf` code that implementing absolute error bounding would require deep changes throughout the stack so it's not something I propose to do. I just think that before changing public API it's important to consider what the ideal API is rather than introduce something temporary.

Also the methods that are added to Poly include `root_equality_test` which has an interface that seems unintuitive to me. Also `root_bounds` really just computes the Cauchy bounds so I wonder if it would make sense to call it `cauchy_bounds`.

Another question from a codebase design perspective is if it really makes sense to have a method like `cauchy_bounds` as something like `dup_cauchy_bounds` so that it can be used at the lower-levels without needing to wrap things up in a Poly.

skieffer:
(1)

> From an API design perspective I would have thought that it would make more sense to make bounded error an option to `evalf`

I agree, that is probably the right design. For now do you think it would be enough to rename `evalf_with_bounded_error` with a leading underscore to make it clear it's not public?

(2)

> `root_equality_test` which has an interface that seems unintuitive to me.

Yes, I no longer think that caching the test function is important, so this can be more like,

```python

    def root_equality_test(f, a, b):
        ...
        return boolean saying whether a and b are equal
```

What about the name? `root_equality_test()`? `test_root_equality()`? `roots_are_equal()`?

(3)

> have a method like `cauchy_bounds` as something like `dup_cauchy_bounds` so that it can be used at the lower-levels without needing to wrap things up in a Poly

Do you have a sense for the module where `dup_cauchy_bounds()` belongs? I'd propose `polys.rootisolation`, which is filled with many `dup_...` methods on the topic of root geometry. In fact, it already has `dup_root_upper_bound()` (which appears to apply only to real roots).

(4)

> `root_bounds` really just computes the Cauchy bounds so I wonder if it would make sense to call it `cauchy_bounds`

In a similar sense, I wonder if we should be cautious about what we're offering with `Poly.root_separation_lower_bound_squared()`. Maybe for now it would be better just to offer the Mignotte bound, since it's a known thing, and save this idea for future work.

If so, then maybe it should be `dup_mignotte_sep_bound()`, again in the `polys.rootisolation` module?

oscarbenjamin:
> For now do you think it would be enough to rename `evalf_with_bounded_error` with a leading underscore to make it clear it's not public?

Yes.

> What about the name? `root_equality_test()`? `test_root_equality()`? `roots_are_equal()`?

What about `same_root` or `is_same_root/are_same_root`?

> I'd propose `polys.rootisolation`

I agree.

> If so, then maybe it should be `dup_mignotte_sep_bound()`, again in the `polys.rootisolation` module?

Yes maybe. Not sure...

I'm less concerned about adding "public" `dup_` functions than public Poly methods. There are hundreds of `dup_` functions and it's not hard to just leave one there once it's added.

skieffer:
Okay, I have addressed the last four things discussed.

For a root separation bound, I implemented Mignotte, as `dup_mignotte_sep_bound_squared()` in `polys.rootisolation`.

sympy/polys/polytools.py
```diff
@@ -3773,6 +3775,57 @@ def nth_power_roots_poly(f, n):
...
+        delta_sq = f.domain.get_field().to_sympy(dom_delta_sq)
+        # We have delta_sq = delta**2, where delta is a lower bound on the
+        # minimum separation between any two roots of this polynomial.
+        # Let eps = delta/3, and define eps_sq = eps**2 = delta**2/9.
+        eps_sq = delta_sq / 9
+
+        r, _, _, _ = evalf(1/eps_sq, 1, {})
+        n = fastlog(r)
+        # Then 2^n > 1/eps**2.
+        m = math.ceil(n/2)
```
oscarbenjamin, line 3818: Maybe I'm overthinking this but at first glance this seems an unnecessary use of machine precision floating point...

oscarbenjamin: If `n` is guaranteed to be an an integer then I guess this would work:
```
m = (n // 2) + (n % 2)
```
(Add zero if n is even or 1 if n is odd)

sympy/polys/tests/test_densearith.py
```diff
@@ -970,6 +971,19 @@ def test_dmp_l1_norm():
...
+    assert dup_l2_norm_squared([], ZZ) == 0
+    assert dup_l2_norm_squared([1], ZZ) == 1
+    assert dup_l2_norm_squared([1, 4, 2, 3], ZZ) == 30
+
+
+def test_dmp_l2_norm_squared():
+    assert dmp_l2_norm_squared([[[]]], 2, ZZ) == 0
+    assert dmp_l2_norm_squared([[[1]]], 2, ZZ) == 1
+
+    assert dmp_l2_norm_squared(f_0, 2, ZZ) == 111
```
oscarbenjamin, line 984: Why the blank line here?

oscarbenjamin:
I had a couple of minor comments but otherwise I think this looks good.

sympy/polys/densearith.py
```diff
@@ -1786,6 +1786,48 @@ def dmp_l1_norm(f, u, K):
...
+    >>> R, x = ring("x", ZZ)
+
+    >>> R.dup_l2_norm_squared(2*x**3 - 3*x**2 + 1)
+    14
+
+    """
+    if not f:
+        return K.zero
+    else:
+        return sum([ coeff**2 for coeff in f ])
```
jksuom, line 1806: Maybe something like `return sum([coeff**2 for coeff in f], K.zero)` would suffice.

sympy/polys/polytools.py
skieffer: Interesting. Would not have thought about the efficiency here. Will change it.

sympy/polys/densearith.py
skieffer: I think so. I wrote it this way only to follow suit with the similar methods, like `dup_l1_norm`.

sympy/polys/polytools.py
oscarbenjamin: It's not so much efficiency as correctness:
```python
In [46]: e = 2 ** 53 + 1

In [47]: math.ceil(e / 2)
Out[47]: 4503599627370496

In [48]: (e // 2) + (e % 2)
Out[48]: 4503599627370497
```
That's probably way outside the range of what is possible here but in general we should avoid trusting floating point calculations unless necessary.

skieffer:
Thanks @oscarbenjamin and @jksuom for reviewing! I've addressed your most recent comments. If there's nothing else, I'll merge soon.

oscarbenjamin:
Looks good to me!

jksuom:
Looks good to me too!

PR #22749, opened 2021-12-23
use more targeted Plane.equals

smichr:

Use specific simplification routines that are expected to be able to show that two expressions are (or are not) equal except for a multiplicative constant.

#### Other comments

`simplify` is more expensive than necessary:
```python
define some planes

>>> from random import randint
>>> p=[Plane(*reshape([randint(1,100) for i in range(9)],(3,))) for i in range(50)]

define ratios of equations

>>> r=[p[0].equation(x,y,z)/p[i].equation(x,y,z) for i in range(1,len(p))]

compare times

>>> t=time();ok=list(map(simplify,r));time()-t
8.28780221939087
>>> t=time();ok=list(map(cancel,r));time()-t
0.6698627471923828

get new ratios

>>> r=[p[1].equation(x,y,z)/p[i].equation(x,y,z) for i in range(len(p)) if i-1]

compare in reverse order (cancel first, then simplify)

>>> t=time();ok=list(map(cancel,r));time()-t
0.7477538585662842
>>> t=time();ok=list(map(simplify,r));time()-t
7.3203558921813965
```

So using `simplify` is about an order of magnitude slower.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/geometry/plane.py
+++ b/sympy/geometry/plane.py
@@ -15,7 +15,6 @@ from .line import (Line, Ray, Segment, Line3D, LinearEntity, LinearEntity3D,
 from .point import Point, Point3D
 from sympy.matrices import Matrix
 from sympy.polys.polytools import cancel
-from sympy.simplify.simplify import simplify
 from sympy.solvers import solve, linsolve
 from sympy.utilities.iterables import uniq, is_sequence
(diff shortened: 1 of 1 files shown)


oscargus:
Seems like simplify is quite generously used in geometry.

PR #22752, opened 2021-12-25
functions - added code for some TODO parts of spherical_harmonics.py

praneethratna:
Added checks to make sure `n` is in `N` and `abs(m) <= n` in _eval_rewrite_as_polynomial and _eval_rewrite_as_cos , also added checks to make sure `theta` and `phi` are in `R` in _eval_conjugate.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/functions/special/spherical_harmonics.py
+++ b/sympy/functions/special/spherical_harmonics.py
@@ -184,7 +184,7 @@ class Ynm(Function):
                 return self.expand(func=True)
             elif Abs(m) > n:
                 return S.Zero
-        return self.expand(func=True)        
+        return self.expand(func=True)
 
     def _eval_rewrite_as_sin(self, n, m, theta, phi, **kwargs):
         return self.rewrite(cos)
@@ -198,7 +198,7 @@ class Ynm(Function):
             elif Abs(m) < n:
                 term = S.Zero
         else:
-            term = simplify(self.expand(func=True))              
+            term = simplify(self.expand(func=True))
         # We can do this because of the range of theta
         term = term.xreplace({Abs(sin(theta)):sin(theta)})
         return simplify(trigsimp(term))
@@ -208,7 +208,7 @@ class Ynm(Function):
         if all(map(lambda x: x.is_integer, (n, m, theta, phi))):
             if theta in S.Reals and phi in S.Reals:
                 return S.NegativeOne**m * self.func(n, -m, theta, phi)
-        return S.NegativeOne**m * self.func(n, -m, theta, phi)        
+        return S.NegativeOne**m * self.func(n, -m, theta, phi)
 
     def as_real_imag(self, deep=True, **hints):
         # TODO: Handle deep and hints
--- a/sympy/functions/special/spherical_harmonics.py
+++ b/sympy/functions/special/spherical_harmonics.py
@@ -179,12 +179,11 @@ class Ynm(Function):
             raise ArgumentIndexError(self, argindex)
 
     def _eval_rewrite_as_polynomial(self, n, m, theta, phi, **kwargs):
-        if all(map(lambda x: x.is_integer, (n, m, theta, phi))):
-            if n in S.Naturals and Abs(m) <= n:
+        if n.is_integer and n in S.Naturals:
+            if Abs(m) <= n:
                 return self.expand(func=True)
-            elif Abs(m) > n:
+            else:
                 return S.Zero
-        return self.expand(func=True)
 
     def _eval_rewrite_as_sin(self, n, m, theta, phi, **kwargs):
         return self.rewrite(cos)
@@ -192,22 +191,22 @@ class Ynm(Function):
     def _eval_rewrite_as_cos(self, n, m, theta, phi, **kwargs):
         # This method can be expensive due to extensive use of simplification!
         from sympy.simplify import simplify, trigsimp
-        if all(map(lambda x: x.is_integer, (n, m, theta, phi))):
-            if n in S.Naturals and Abs(m) <= n:
+        if n.is_integer and n in S.Naturals:
+            if Abs(m) <= n:
                 term = simplify(self.expand(func=True))
-            elif Abs(m) < n:
+            else:
                 term = S.Zero
-        else:
-            term = simplify(self.expand(func=True))
         # We can do this because of the range of theta
         term = term.xreplace({Abs(sin(theta)):sin(theta)})
         return simplify(trigsimp(term))
 
     def _eval_conjugate(self):
         n, m, theta, phi = self.args
-        if all(map(lambda x: x.is_integer, (n, m, theta, phi))):
-            if theta in S.Reals and phi in S.Reals:
+        if all(map(lambda x: x.is_Number, (theta, phi))):
+            if theta and phi in S.Reals:
                 return S.NegativeOne**m * self.func(n, -m, theta, phi)
+            else:
+                return None
         return S.NegativeOne**m * self.func(n, -m, theta, phi)
 
     def as_real_imag(self, deep=True, **hints):
--- a/sympy/functions/special/spherical_harmonics.py
+++ b/sympy/functions/special/spherical_harmonics.py
@@ -89,10 +89,7 @@ class Ynm(Function):
     We can differentiate the functions with respect
     to both angles:
 
-    >>> from sympy import Ynm, Symbol, diff
-    >>> from sympy.abc import n,m
-    >>> theta = Symbol("theta", real=True)
-    >>> phi = Symbol("phi", real=True)
+    >>> from sympy import diff
 
     >>> diff(Ynm(n, m, theta, phi), theta)
     m*cot(theta)*Ynm(n, m, theta, phi) + sqrt((-m + n)*(m + n + 1))*exp(-I*phi)*Ynm(n, m + 1, theta, phi)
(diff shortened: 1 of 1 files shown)


sympy/functions/special/spherical_harmonics.py
```diff
@@ -189,16 +192,22 @@ def _eval_rewrite_as_sin(self, n, m, theta, phi, **kwargs):
...
         # We can do this because of the range of theta
         term = term.xreplace({Abs(sin(theta)):sin(theta)})
         return simplify(trigsimp(term))
 
     def _eval_conjugate(self):
-        # TODO: Make sure theta \in R and phi \in R
         n, m, theta, phi = self.args
+        if all(map(lambda x: x.is_integer, (n, m, theta, phi))):
+            if theta in S.Reals and phi in S.Reals:
+                return S.NegativeOne**m * self.func(n, -m, theta, phi)
```
oscargus, line 210: This seems like it doesn't add anything as you do the same thing independent of the if-cases.

oscargus: Also, I do not think theta and phi should be integer?

praneethratna: Sorry, by mistakenly added the wrong code.  I'll commit the correct code again.

sympy/functions/special/spherical_harmonics.py
```diff
@@ -189,16 +192,22 @@ def _eval_rewrite_as_sin(self, n, m, theta, phi, **kwargs):
     def _eval_rewrite_as_cos(self, n, m, theta, phi, **kwargs):
         # This method can be expensive due to extensive use of simplification!
         from sympy.simplify import simplify, trigsimp
-        # TODO: Make sure n \in N
-        # TODO: Assert |m| <= n ortherwise we should return 0
```
oscargus, line 193: I interpret these as different requirements.
1) n should be in S.Naturals, if so, if `Abs(m) < n`. So while the comparison will break for complex n, I do not think one should return 0 if n is not in S.Naturals (but one may have to change something other than the comment to see...).

praneethratna: But n should return 0 when `Abs(m) > n` ? Not sure of what should be returned when n is not in S.Naturals.

sympy/functions/special/spherical_harmonics.py
praneethratna: > Also, I do not think theta and phi should be integer?

I think the check should be is_Number but I'm not sure.

sympy/functions/special/spherical_harmonics.py
oscargus: If I understand it correctly, in general, the rewrite can only happen for these cases. So unless it can be rewritten one should return self (or possibly None, not sure how it works in detail).

So, the previous code just applies the rewrite, although it should only be applied if the requirements hold. (The problem is that when you rewrite like this you can possibly evaluate cases that really are not valid if they were evaluated using the original function.) But this is how I interpret it.

oscargus: None.

Compare to this from factorial. Only if the argument is integer and non-negative, the rewrite is valid:

praneethratna: I've made a commit and rewrote the code but I'm not sure of the  `_eval_conjugate` one as of now. For that I'm checking if `phi` and `theta` are numbers and if satisfied then I'm checking the whether `phi` and `theta` are in `S.Reals` and returning `None` if they aren't in `S.Reals`.

praneethratna:
```
>>> theta = Symbol("theta")
>>> phi = Symbol("phi")
```
I've added real = True  for the above lines in the docstring but for some reason theta and phi aren't being evaluated as Reals.

sympy/functions/special/spherical_harmonics.py
```diff
@@ -179,25 +179,28 @@ def fdiff(self, argindex=4):
             raise ArgumentIndexError(self, argindex)
 
     def _eval_rewrite_as_polynomial(self, n, m, theta, phi, **kwargs):
-        # TODO: Make sure n \in N
-        # TODO: Assert |m| <= n ortherwise we should return 0
-        return self.expand(func=True)
+        if n.is_integer and n in S.Naturals:
```
oscargus, line 182:
```suggestion
        if n in S.Naturals:
```
I think this should be enough? Maybe add some tests to see what happens for different scenarios?

sympy/functions/special/spherical_harmonics.py
```diff
@@ -179,25 +179,28 @@ def fdiff(self, argindex=4):
...
+            if Abs(m) <= n:
```
oscargus, line 183: I think that there may be an issue here that this cannot always evaluate and therefore gives an error.
Maybe something like 
```
c = is_le(Abs(m), n)
if c == True:
    return self.expand(func=True)
elif c == False:
    return S.Zero
```
In this way, if the comparison cannot be evaluated it will not break anything.

sympy/functions/special/spherical_harmonics.py
```diff
@@ -179,25 +179,28 @@ def fdiff(self, argindex=4):
...
     def _eval_rewrite_as_sin(self, n, m, theta, phi, **kwargs):
         return self.rewrite(cos)
 
     def _eval_rewrite_as_cos(self, n, m, theta, phi, **kwargs):
         # This method can be expensive due to extensive use of simplification!
         from sympy.simplify import simplify, trigsimp
-        # TODO: Make sure n \in N
-        # TODO: Assert |m| <= n ortherwise we should return 0
-        term = simplify(self.expand(func=True))
+        if n.is_integer and n in S.Naturals:
```
oscargus, line 194: Similar here and below.

sympy/functions/special/spherical_harmonics.py
```diff
@@ -179,25 +179,28 @@ def fdiff(self, argindex=4):
...
+            if Abs(m) <= n:
+                term = simplify(self.expand(func=True))
+            else:
+                term = S.Zero
         # We can do this because of the range of theta
         term = term.xreplace({Abs(sin(theta)):sin(theta)})
         return simplify(trigsimp(term))
 
     def _eval_conjugate(self):
-        # TODO: Make sure theta \in R and phi \in R
```
oscargus, line 200: Do not remove this unless it is tested for.

praneethratna:
ping @oscargus I've made the changes as you've mentioned but `TODO` part of _eval_conjugate is left out.

sympy/functions/special/spherical_harmonics.py
```diff
@@ -179,19 +180,25 @@ def fdiff(self, argindex=4):
...
-        # TODO: Make sure n \in N
-        # TODO: Assert |m| <= n ortherwise we should return 0
-        term = simplify(self.expand(func=True))
+        if n in S.Naturals:
+            c = is_le(Abs(m), n)
+            if c == True:
+                term = simplify(self.expand(func=True))
+            elif c == False:
+                term = S.Zero
         # We can do this because of the range of theta
```
oscargus, line 202: I think this part should only be done in the if-clause. 

Some tests are needed which would detect e.g. this strange behavior.

sympy/functions/special/spherical_harmonics.py
```diff
@@ -201,6 +208,7 @@ def _eval_conjugate(self):
         n, m, theta, phi = self.args
         return S.NegativeOne**m * self.func(n, -m, theta, phi)
 
+
```
oscargus, line 211: Please get rid of this extra empty line (only a single empty line between methods).

sympy/functions/special/spherical_harmonics.py
```diff
@@ -36,8 +37,8 @@ class Ynm(Function):
 
     >>> from sympy import Ynm, Symbol, simplify
     >>> from sympy.abc import n,m
-    >>> theta = Symbol("theta")
-    >>> phi = Symbol("phi")
+    >>> theta = Symbol("theta", real=True)
```
oscargus, line 40: Is this needed? It doesn't look like any of the added code makes use of it.

sympy/functions/special/spherical_harmonics.py
```diff
@@ -90,8 +91,8 @@ class Ynm(Function):
 
     >>> from sympy import Ynm, Symbol, diff
     >>> from sympy.abc import n,m
```
oscargus, line 93: This part can be removed afaik. I think it is only required to import diff here. (Or one can simply write `Ynm(n, m, theta, phi).diff(theta)`)

sympy/functions/special/spherical_harmonics.py
```diff
@@ -103,8 +104,8 @@ class Ynm(Function):
 
     >>> from sympy import Ynm, Symbol, conjugate
     >>> from sympy.abc import n,m
-    >>> theta = Symbol("theta")
-    >>> phi = Symbol("phi")
+    >>> theta = Symbol("theta", real=True)
+    >>> phi = Symbol("phi", real=True)
```
oscargus, line 108: Similar here. `Ynm(n, m, theta, phi).conjugate()` should work. (And no additional imports or recomputations.)

I may be wrong though, but I am under the impression that intermediate text does not require new imports and redefinitions...

sympy/functions/special/spherical_harmonics.py
praneethratna: They're to ensure `theta` and `phi` are real but the check in `_eval_conjugate` isn't working properly.

sympy/functions/special/spherical_harmonics.py
praneethratna: I'm a bit confused with what kind of tests to add. Could you help me with it?

praneethratna:
ping @oscargus

PR #15929, opened 2019-02-05
[WIP] Added new simplifications for Piecewise and fixed issue with previous version

oscargus:
#### Brief description of what is fixed or changed
Now Piecewise of the type
```
⎧y + 2  for x = 2 ∧ y = 0
⎨                        
⎩  x        otherwise    

```
can be simplified (to `x`) as well as of the type
```
⎧ 2            
⎪x    for x ≠ 2
⎪              
⎨2⋅x  for x ≠ 4
⎪              
⎪x⋅y  otherwise
⎩    
```          
(gives ~~`2⋅x`~~ `x**2`).

Hence, for Or-type of Equality or Or of And of Equality `Or(Eq..., And(Eq.., Eq...)))` it is checked if any of the terms in the Or can be moved to the next segment. It also checks if the total measure of the segment expressions is reduced. Here, one may think of cases where one term can be move to several of the following segments and one tries to figure out what the best combination is, probably through some exhaustive search. One may also note that the fact that segments are removed are not explicitly encouraged. This will follow from the simplification of the segment expressions, moving an expression should not increase the complexity, if anything reduce it, and with a sensible ratio this should be OK. But if we try to position an Or-term it will require that one also takes this into account and that will require additional computational complexity. If there are many cases that does not simplify well here it should be possible to implement though.

For Neq expressions, it is simply checked if the next expression happens to evaluate to the same expression as the Neq expression, for the same value as the Neq expression is not allowed to take. In that case, the expression of the latter expression is used for the first and eventually merged.

I also realized that there were some cases in the previous PR #15898 that did not work correctly. Those should work now.

#### Release Notes

NO ENTRY

Issue #21360, opened 2021-04-20
Issues with simplifying Piecewise


oscarbenjamin:
The examples below show a few problems with `Piecewise` and `simplify`:

1. Simplify assumes that all conditions are Real and so `Ne(a, 0)` and `Ne(a, -2)` becomes `(a < -2) | ((-2 < a) & (a < 0)) | (0 < a)`. This does not work for complex values of `a`. I suspect that the problem is from using `as_set/as_relational` somewhere.
2. Simplify should be able to infer that if the first case is `Ne(a, 0) & Ne(a, -2)` then in subsequent cases we must have `Eq(a, 0) | Eq(a, -2)` so if the second condition is `Ne(a, 0)` then that implies `Eq(a, -2)` and we can substitute `-2` for `a` in the expression.
3. By the third case we know that `a = 0` since the condition for the second case was `Ne(a, 0)`. We must always have `Eq(a, 0)` in all cases after a case with the condition `Ne(a, 0)`. We can substitute `a=0` into the expression and conditions for all cases below the second case.
4. After 3 we know that the condition for the third case will always be `True` so the 4th case isn't needed.
5. In the output for `integrate(m.expand(), x).diff(x).simplify()` substituting `a = -2` into the expressions for the first and third cases gives the same expression so those cases can be combined. Likewise for `a=0` and the first and third cases.

With all of these taken together it should be possible for simplify to recover the original formula after integrating and differentiating without any Piecewise in the result. Note that all of these things should be done in `simplify` and *not* in `Piecewise.__new__`.

```python
In [10]: a = Symbol('a', real = True)

In [11]: m = (a*y + x**2)*(x**(a-1))

In [12]: integrate(m.expand(), x)
Out[12]: 
    ⎛⎧   a             ⎞   ⎛⎧ 2  a             ⎞
    ⎜⎪  x              ⎟   ⎜⎪x ⋅x              ⎟
    ⎜⎪  ──    for a ≠ 0⎟   ⎜⎪─────   for a ≠ -2⎟
a⋅y⋅⎜⎨  a              ⎟ + ⎜⎨a + 2             ⎟
    ⎜⎪                 ⎟   ⎜⎪                  ⎟
    ⎜⎪log(x)  otherwise⎟   ⎜⎪log(x)  otherwise ⎟
    ⎝⎩                 ⎠   ⎝⎩                  ⎠

In [13]: integrate(m.expand(), x).simplify()
Out[13]: 
⎧    a              a + 2                                          
[...]

Issue #8458, opened 2014-11-16
Simplifying Piecewise when two cases have same value


cbm755:
```
In [92]: Piecewise((0, Eq(x, 0)), (sin(x), True))
Out[92]: Piecewise((0, x == 0), (sin(x), True))

In [93]: pprint(_)
⎧  0     for x = 0
⎨                 
⎩sin(x)  otherwise

In [94]: simplify(_)
Out[94]: Piecewise((0, x == 0), (sin(x), True))
```

Perhaps this should not happen automatically, but it would be nice if `simplify()` did it.  Apologies if this is a dupe, didn't see it.  Figure its related to #1009 somehow.

---

Very simple cases work:

```
In [97]: Piecewise((10, Eq(x,0)), (10, True))
Out[97]: 10
```

And some more complicated things work too:

```
In [101]: p3 = Piecewise((10, Eq(x, 0)), (y, True))

In [101]: p3
Out[101]: Piecewise((10, x == 0), (y, True))

In [102]: p3.subs(y,10)
Out[102]: 10
```

smichr:
cf #8453

PR #22753, opened 2021-12-26
Refactored Piecewise simplification and added simplification methods

### LINKED ISSUE #17283: Eq in condition of Piecewise might be redundant
```python
>>> eq=Piecewise((0, x <= 1), (1, Eq(x, -oo)), (2, True))
>>> simplify(eq)
Piecewise((0, x <= 1), (1, Eq(x, -oo)), (2, True))
```
The result should be `Piecewise((0, x <= 1), (2, True))`: if a condition is an equality sets a previous condition to True then it is redundant. This easy case should probably be removed at instantiation.

(identified [here](https://github.com/sympy/sympy/pull/17146#discussion_r307968690))

### LINKED ISSUE #21481: simplified Piecewise is missing conditions
I have been trying to write a model for the size of symbolic powers.
I define the Piecewise expression as follows for power `b**e` with
the expressions of the Piecewise giving a value that is representative
of the region in which the size of the power is found (e.g. 1/2 if the
value of the power is between 0 and 1):

```python
var('b e')
A=Piecewise((1, Eq(b, 1) | Eq(e, 0) | (Eq(b, -1) & Eq(Mod(e, 2), 0))),
(0, Eq(b, 0) & (e > 0)), (-1, Eq(b, -1) & Eq(Mod(e, 2), 1)),
(Piecewise((2, ((b > 1) & (e > 0)) | ((b > 0) & (b < 1) & (e < 0)) |
((e >= 2) & (b < -1) & Eq(Mod(e, 2), 0)) | ((e <= -2) & (b > -1) & (b
< 0) & Eq(Mod(e, 2), 0))), (S.Half, ((b > 1) & (e < 0)) | ((b > 0) & (e >
0) & (b < 1)) | ((e <= -2) & (b < -1) & Eq(Mod(e, 2), 0)) | ((e >= 2)
& (b > -1) & (b < 0) & Eq(Mod(e, 2), 0))), (-S.Half, Eq(Mod(e, 2), 1) &
(((e <= -1) & (b < -1)) | ((e >= 1) & (b > -1) & (b < 0)))), (-2, ((e
>= 1) & (b < -1) & Eq(Mod(e, 2), 1)) | ((e <= -1) & (b > -1) & (b < 0)
& Eq(Mod(e, 2), 1)))), Eq(im(b), 0) & Eq(im(e), 0)))
```
As a test of folding and simplification I create the following alternative forms for A
```python
B=piecewise_fold(A)
sa=A.simplify()
sb=B.simplify()
```
I then test the forms over a range of `b` and `e` values:
```python
v = Tuple(-2, -1, -0.5, 0, 0.5, 1, 2)
for i in v:
 for j in v:
  r = {b:i,e:j}
  ok=[k.xreplace(r) for k in (A,B,sa,sb)]
  if len(set(ok))!=1:print('ab %s %s'%(r,ok))
```
Although `A` and `B` agree, the simplified forms of each do not and I get these results
```python
s {b: -2, e: -1} [-1/2, -1/2, -1/2, nan]
s {b: -2, e: 1} [-2, -2, -2, nan]
s {b: -1/2, e: 1} [-1/2, -1/2, -1/2, nan]
```
The `nan` indicates that the conditions for the values of `b` and `e` that were passed are not found in the Piecewise expression so a default `nan` is returned. I confirmed that the value returned by `sa` agrees with the value from `A`. I have not yet tracked down the source of the discrepancy from `sb` in which are missing the cases for -2 and -1/2:
```python
Piecewise(
    (1, Eq(b, 1) | Eq(e, 0) | (Eq(b, -1) & Eq(Mod(e, 2), 0))),
    (0, Eq(b, 0) & (e > 0)),
    (-1, Eq(b, -1) & Eq(Mod(e, 2), 1)),
    (2, Eq(im(b), 0) & Eq(im(e), 0) & ((b > 0) | (b > 1)) & ((b > 0) | (e > 0)) & ((b > 1) | (b < 1)) & ((b > 1) | (e < 0)) & ((e > 0) | (b < 1)) & ((e > 0) | (e < 0))),
    (1/2, Eq(im(b), 0) & Eq(im(e), 0) & ((b > 0) | (b > 1)) & ((b > 0) | (e < 0)) & ((b > 1) | (e > 0)) & ((b > 1) | (b < 1)) & ((e > 0) | (e < 0)) & ((b < 1) | (e <0))))
```

### LINKED ISSUE #22066: conflictive expressions in cases statement as integral solution
The following expression
```python
import sympy as sy

x   = sy.symbols("x")
L   = sy.symbols("L", nonzero = True)
k, l = sy.symbols("k, l",   integer=True, nonnegative = True )

integral = sy.Integral( sy.cos( (x*k*sy.pi) / L) * sy.cos( (x*l*sy.pi) / L), (x, -L, L) )
eq = sy.Eq( integral , sy.simplify(integral.doit()) )
eq
```
results in

image

Based on the assumptions given for the symbols the conditions stated in the cases statement are not as expected.
The solution I expected is 
 
image

There are two things that are bothering me

1. cases statement 0 and 1 are conflictive
2. k = -l is always a false statement as k and l where defined as non negative integers

Is there something I'm missing or is this a bug?

oscargus:

Fixes  #8453
Closes #15929
Closes #17443
Fixes #17283
Closes #21481 (fixed elsewhere)
Fixes #22066

Refactored the simplification of Piecewise and added more possible simplification steps.

Builds on #15929 and #17443.

One may have to fine tune the steps, but this seems like a reasonable order.

#### Release Notes

* functions
     * Better simplification of `Piecewise`

### DIFF
--- a/sympy/logic/boolalg.py
+++ b/sympy/logic/boolalg.py
@@ -48,14 +48,14 @@ def as_Boolean(e):
     """
     from sympy.core.symbol import Symbol
     if e == True:
-        return S.true
+        return true
     if e == False:
-        return S.false
+        return false
     if isinstance(e, Symbol):
         z = e.is_zero
         if z is None:
             return e
-        return S.false if z else S.true
+        return false if z else true
     if isinstance(e, Boolean):
         return e
     raise TypeError('expecting bool or Boolean, not `%s`.' % e)
@@ -357,7 +357,7 @@ class BooleanTrue(BooleanAtom, metaclass=Singleton):
 
     @property
     def negated(self):
-        return S.false
+        return false
 
     def as_set(self):
         """
@@ -432,7 +432,7 @@ class BooleanFalse(BooleanAtom, metaclass=Singleton):
 
     @property
     def negated(self):
-        return S.true
+        return true
 
     def as_set(self):
         """
@@ -500,9 +500,9 @@ class BooleanFunction(Application, Boolean):
                 if x in bin_syms and x in r.free_symbols:
                     if isinstance(r, (Eq, Ne)):
                         if not (
-                                S.true in r.args or
-                                S.false in r.args):
-                            reps[r] = S.false
+                                true in r.args or
+                                false in r.args):
+                            reps[r] = false
                     else:
                         raise TypeError(filldedent('''
                             Incompatible use of binary symbol `%s` as a
@@ -613,7 +613,7 @@ class And(LatticeOp, BooleanFunction):
                 if c in rel:
                     continue
                 elif c.negated.canonical in rel:
-                    return [S.false]
+                    return [false]
                 else:
                     rel.add(c)
             newargs.append(x)
@@ -631,7 +631,7 @@ class And(LatticeOp, BooleanFunction):
                     bad = i
                 continue
             if i == False:
-                return S.false
+                return false
             elif i != True:
                 args.append(i)
         if bad is not None:
@@ -712,7 +712,7 @@ class And(LatticeOp, BooleanFunction):
         patterns = _simplify_patterns_and()
         threeterm_patterns = _simplify_patterns_and3()
         return _apply_patternbased_simplification(rv, patterns,
-                                                  measure, S.false,
+                                                  measure, false,
                                                   threeterm_patterns=threeterm_patterns)
 
     def _eval_as_set(self):
@@ -771,7 +771,7 @@ class Or(LatticeOp, BooleanFunction):
                     continue
                 nc = c.negated.canonical
                 if any(r == nc for r in rel):
-                    return [S.true]
+                    return [true]
                 rel.append(c)
             newargs.append(x)
         return LatticeOp._new_args_filter(newargs, Or)
@@ -788,7 +788,7 @@ class Or(LatticeOp, BooleanFunction):
                     bad = i
                 continue
             if i == True:
-                return S.true
+                return true
             elif i != False:
                 args.append(i)
         if bad is not None:
@@ -824,7 +824,7 @@ class Or(LatticeOp, BooleanFunction):
             return rv
         patterns = _simplify_patterns_or()
         return _apply_patternbased_simplification(rv, patterns,
-                                                  kwargs['measure'], S.true)
+                                                  kwargs['measure'], true)
 
     def to_anf(self, deep=True):
         args = range(1, len(self.args) + 1)
@@ -1243,10 +1243,10 @@ class Implies(BooleanFunction):
         if A in (True, False) or B in (True, False):
             return Or(Not(A), B)
         elif A == B:
-            return S.true
+            return true
         elif A.is_Relational and B.is_Relational:
             if A.canonical == B.canonical:
-                return S.true
+                return true
             if A.negated.canonical == B.canonical:
                 return B
         else:
@@ -1396,17 +1396,17 @@ class ITE(BooleanFunction):
             if len(set(a.args) - bin_syms) == 1:
                 # one arg is a binary_symbols
                 _a = a
-                if a.lhs is S.true:
+                if a.lhs is true:
                     a = a.rhs
-                elif a.rhs is S.true:
+                elif a.rhs is true:
                     a = a.lhs
-                elif a.lhs is S.false:
+                elif a.lhs is false:
                     a = Not(a.rhs)
-                elif a.rhs is S.false:
+                elif a.rhs is false:
                     a = Not(a.lhs)
                 else:
                     # binary can only equal True or False
-                    a = S.false
+                    a = false
                 if isinstance(_a, Ne):
                     a = Not(a)
         else:
@@ -1426,26 +1426,26 @@ class ITE(BooleanFunction):
         a, b, c = args
         if isinstance(a, (Ne, Eq)):
             _a = a
-            if S.true in a.args:
-                a = a.lhs if a.rhs is S.true else a.rhs
-            elif S.false in a.args:
-                a = Not(a.lhs) if a.rhs is S.false else Not(a.rhs)
+            if true in a.args:
+                a = a.lhs if a.rhs is true else a.rhs
+            elif false in a.args:
+                a = Not(a.lhs) if a.rhs is false else Not(a.rhs)
             else:
                 _a = None
             if _a is not None and isinstance(_a, Ne):
                 a = Not(a)
-        if a is S.true:
+        if a is true:
             return b
-        if a is S.false:
+        if a is false:
             return c
         if b == c:
             return b
         else:
             # or maybe the results allow the answer to be expressed
             # in terms of the condition
-            if b is S.true and c is S.false:
+            if b is true and c is false:
                 return a
-            if b is S.false and c is S.true:
+            if b is false and c is true:
                 return Not(a)
         if [a, b, c] != args:
             return cls(a, b, c, evaluate=False)
@@ -3259,11 +3259,11 @@ def _simplify_patterns_and():
     # Relationals patterns should be in alphabetical order
     # (pattern1, pattern2, simplified)
     # Do not use Ge, Gt
-    _matchers_and = ((Tuple(Eq(a, b), Lt(a, b)), S.false),
-                     #(Tuple(Eq(a, b), Lt(b, a)), S.false),
-                     #(Tuple(Le(b, a), Lt(a, b)), S.false),
-                     #(Tuple(Lt(b, a), Le(a, b)), S.false),
-                     (Tuple(Lt(b, a), Lt(a, b)), S.false),
+    _matchers_and = ((Tuple(Eq(a, b), Lt(a, b)), false),
+                     #(Tuple(Eq(a, b), Lt(b, a)), false),
+                     #(Tuple(Le(b, a), Lt(a, b)), false),
+                     #(Tuple(Lt(b, a), Le(a, b)), false),
+                     (Tuple(Lt(b, a), Lt(a, b)), false),
                      (Tuple(Eq(a, b), Le(b, a)), Eq(a, b)),
                      #(Tuple(Eq(a, b), Le(a, b)), Eq(a, b)),
                      #(Tuple(Le(b, a), Lt(b, a)), Gt(a, b)),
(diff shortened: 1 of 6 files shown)


robinechuca:
Hello, this looks like the pull request 21933 that I proposed a while ago. I'm glad to hear that the idea has not been abandoned.

oscargus:
@robinechuca This is independent of that (unfortunately) and some old stuff I did a few years back. There is a bit of an issue here which should be merged first as the other will have to be modified quite a bit... But I believe that your additions definitely are worthwhile and should be able to fit in this chain of simplifications somewhere.

NotWearingPants:
Probably also fixes #8453 ?

NotWearingPants:
A few more issues to check (if they're not already handled by this PR then tell me and I'll open issues):

---

```py
Piecewise((0, x>3))
```
simplifies to `0` but it could be `NaN` when `x>=3`, so it should stay as a `Piecewise`.

---

```py
Piecewise(
    ( Piecewise( (1,Eq(x,3)), (2,True) ),    Ne(x,3) ),
    (3, True),
)
```
should realize getting 1 is impossible and simplify to
```py
Piecewise(
    (2, Ne(x,3)),
    (3, True),
)
```

---

```py
Piecewise((x, Ne(x, 1)), (1, True))
```
should simplify to `x`, similar to #8458.

---

btw is the PR still viable or is it too old to fix?

oscargus:
> ```python
> Piecewise((0, x>3))
> ```

This is indeed a bug. I do not think that this PR fixes it (unless coincidental).
> ```python
> Piecewise(
>     ( Piecewise( (1,Eq(x,3)), (2,True) ),    Ne(x,3) ),
>     (3, True),
> )
> ```

`piecewise_fold` will get the expected result. One may for sure consider doing `piecewise_fold` as part of the simplification (I almost thought that it was done).

> ```python
> Piecewise((x, Ne(x, 1)), (1, True))
> ```

There is a at least code in this PR that should simplify it.

> btw is the PR still viable or is it too old to fix?

There is probably a bit of work to sort it out, but I think it would be worthwhile. The problem is if it ends up like this and not getting reviews and then just hangs around. But as 1.11 is about to be released, I guess that one should be able to rebase and merge after that so that there is time to find any issues before the next release.

oscargus:
I rebased it and fixed the conflicts (not pushed yet). It seems like the last case is indeed fixed here. I'll add a test for that as well.

And you are right about #8453. Can't understand how I missed that...

oscargus:
@oscarbenjamin @asmeurer I've rebased this (again) and would like to have it seriously considered for merging. 

It is, naturally, heuristic, and it may be that one must run it more than once etc, but I believe that it is a significant step in terms of simplification capabilities.

oscarbenjamin:
I haven't checked through this in detail but if I understand correctly this is adding a lot of potentially expensive automatic simplification code to `Piecewise.eval`. In general I would rather remove automatic simplification rather than add more of it. I would naturally expect a lot of this to belong in `_eval_simplify` rather than `eval`. The previous `eval` method seems more appropriately restrained in its effort to apply some basic simplification but without working too hard for it.

oscargus:
> lot of potentially expensive automatic simplification code to `Piecewise.eval`

No. It just refactors that part out of `eval` as it is reused in `piecewise_simplify`.

oscarbenjamin:
> > lot of potentially expensive automatic simplification code to `Piecewise.eval`
> 
> No. It just refactors that part out of `eval` as it is reused in `piecewise_simplify`.

Oh sorry, I see what you mean. So `eval` only calls `_piecewise_collapse_arguments` which is equivalent to the existing `eval` code and the other simplification functions are only used in `piecewise_simplify/_eval_simplify`.

sympy/logic/boolalg.py
```diff
@@ -2769,6 +2769,13 @@ def simplify_logic(expr, form=None, deep=True, force=False):
         made. By setting ``force`` to ``True``, this limit is removed. Be
         aware that this can lead to very long simplification times.
 
+    dontcare : Boolean expression
+        Optimize expression under the assumption that this expression is false.
+        This is useful in e.g. Piecewise conditions, where later confitions
+        are subject to the previous conditions being false. Hence, if a previous
+        condition is And(A, B), the simplification of expr can be made with
+        don't cares for And(A, B).
```
oscarbenjamin, line 2777: typo -> confitions.

Would it make more sense to negate this and call the argument something like `assumptions`? I think maybe it's easier to understand assuming something to be True rather than assuming something to be False.

oscarbenjamin: Or perhaps a better name for the argument would be `assume_false`.

oscargus: I changed the doc-string to better explain what is actually going on.

oscargus:
OK, there are still some bugs here...

```
simplify(Piecewise((x, And(Ne(x, 1), Ne(x, 2))), (1, Eq(x, 1)), (2, Eq(x, 2))))
```
gives 
``` 
⎧x  for x = 1 ∨ x ≠ 2
⎨                    
⎩2      for x = 2    
```

I think that maybe the Ne only can handle the single case at the moment (spent some time thinking about how to deal with this, but cannot really come up with anything efficient, especially considering `And` and `Or`).

sympy/logic/boolalg.py
oscarbenjamin: Okay, that makes more sense.

oscarbenjamin:
This PR helps the example from #21360 by eliminating one case:

Master:
```python
In [1]: a = Symbol('a', real = True)

In [2]: m = (a*y + x**2)*(x**(a-1))

In [3]:  integrate(m.expand(), x).diff(x).simplify()
Out[3]: 
⎧ a - 1 ⎛       2⎞                    
⎪x     ⋅⎝a⋅y + x ⎠  for a ≠ -2 ∧ a ≠ 0
⎪                                     
⎪      a                              
⎪   a⋅x ⋅y + 1                        
⎪   ──────────          for a ≠ 0     
⎪       x                             
⎪                                     
⎨         a + 2                       
⎪  a⋅y + x                            
⎪  ────────────         for a ≠ -2    
⎪       x                             
⎪                                     
⎪     a⋅y + 1                         
⎪     ───────           otherwise     
⎪        x                            
⎩           
```
This PR:
```python
In [3]:  integrate(m.expand(), x).diff(x).simplify()
Out[3]: 
⎧ a - 1 ⎛       2⎞                    
⎪x     ⋅⎝a⋅y + x ⎠  for a ≠ -2 ∧ a ≠ 0
⎪                                     
⎪      a                              
⎨   a⋅x ⋅y + 1                        
⎪   ──────────          for a ≠ 0     
⎪       x                             
⎪                                     
⎩        x              otherwise   
```
It's hard to tell what has caused that change (mixing a refactor with nontrivial changes in the same diff makes the changes hard to follow). It looks as if the third case has been simplified with `a = 0` implying that the condition is True and also that the expression can simplify. What has not simplified is:

- The condition in the second case which could be `a = -2`.
- Then `a = -2` could be used to substitute into the second expression.
- Substituting `a = 0` or `a = -2` into the first expression gives something equivalent to the second and third expressions:
```python
In [4]:  integrate(m.expand(), x).diff(x).simplify().args[0][0].subs(a, -2)
Out[4]: 
 2      
x  - 2⋅y
────────
    3   
   x    

In [5]:  integrate(m.expand(), x).diff(x).simplify().args[1][0].subs(a, -2)
Out[5]: 
    2⋅y
1 - ───
      2
     x 
───────
   x   

In [6]:  integrate(m.expand(), x).diff(x).simplify().args[1][0].subs(a, -2).together()
Out[6]: 
 2      
x  - 2⋅y
────────
    3   
   x    

In [7]:  integrate(m.expand(), x).diff(x).simplify().args[0][0].subs(a, 0)
Out[7]: x
```
If those were handled then the Piecewise could collapse completely.

oscarbenjamin:
This URL gives the (current) diff for this PR minus the initial commit which is the refactor:
https://github.com/sympy/sympy/pull/22753/files/64cedc463177a9a9e218bc0e321208ee6581bc1a..5b67e1298dfad74f30ae5289f05c1bc5ffd2dd23

oscargus:
Well, all the actual simplification takes parts in the different functions, so factoring out one method shouldn't really be a big deal.

oscargus:
> * The condition in the second case which could be `a = -2`.

The thing is that the complexity is not higher for `a != 0`, so `simplify` chooses to not change this. It can of course be hardcoded that a single `Eq` is selected over a single `Ne` if possible.

oscargus:
With a small modification I get:

image

after the first simplification and

image

after the second and finally
image

after the third.

This all comes down to which order one should run the simplifications in.

There will probably always be cases to be found, but I'd go in the direction of improvements. (Heck, if I knew how to make a perfect Piecewise simplification I would work with it. This is probably better than Mathematica for some cases.)

oscarbenjamin:
> This all comes down to which order one should run the simplifications in.

I wonder if it would be simpler to start by applying `piecewise_exclusive` and then manage the different cases more or less independently rather than having each simplification step need to consider the ordering of the cases:
```python
In [8]: piecewise_exclusive(integrate(m.expand(), x).diff(x).simplify())
Out[8]: 
⎧ a - 1 ⎛       2⎞                    
⎪x     ⋅⎝a⋅y + x ⎠  for a ≠ -2 ∧ a ≠ 0
⎪                                     
⎪      a                              
⎪   a⋅x ⋅y + 1                        
⎪   ──────────          for a = -2    
⎪       x                             
⎪                                     
⎨         a + 2                       
⎪  a⋅y + x                            
⎪  ────────────         for a = 0     
⎪       x                             
⎪                                     
⎪     a⋅y + 1                         
⎪     ───────       for a = -2 ∧ a = 0
⎪        x                            
⎩   
```
From here is should be straight-forward to eliminate the final case and substitute values of `a` in the others. In this form we more or less know that Eq and Ne come in pairs so that an Eq in one case must clash with an Ne somewhere else.

oscargus:
Yes, that probably works well for that type of simpler conditions with Ne and Eq. For other cases it is probably a waste of time. But the problem is that it is really hard to know which the correct things to do are for the general case. But one can for sure add some optimizations for that case.

oscarbenjamin:
After the last commit it seems that `Eq` is preferred over `Ne` but the final case is not filtered out any more:
```python
In [9]: simplify(integrate(m.expand(), x).diff(x))
Out[9]: 
⎧ a - 1 ⎛       2⎞                    
⎪x     ⋅⎝a⋅y + x ⎠  for a ≠ -2 ∧ a ≠ 0
⎪                                     
⎪         2⋅y                         
⎪     1 - ───                         
⎪           2                         
⎪          x                          
⎨     ───────           for a = -2    
⎪        x                            
⎪                                     
⎪        x              for a = 0     
⎪                                     
⎪     a⋅y + 1                         
⎪     ───────           otherwise     
⎩        x  
```

oscarbenjamin:
I tried to find a paper about simplifying Piecewise. There don't seem to be many but I did find these:
https://arxiv.org/abs/cs/0702010
https://www.sciencedirect.com/science/article/pii/S0377042714005123

It doesn't look like either of them really handles the case where more than one variable appears in the conditions.

oscargus:
Yeah, you need to run it multiple times at the moment.

Maybe the best thing is to split this in multiple PRs. At the moment I am just adding features rather than making it converge...

So start with refactoring and then add new simplification functions as we go along.

oscarbenjamin:
> Maybe the best thing is to split this in multiple PRs.

It would certainly be easier to review that way.

> At the moment I am just adding features rather than making it converge...

Maybe some of these operations just need to be done in a while loop. I find it hard to follow precisely what all of them do so I'm not sure which. Since each of them runs a pass over the cases maybe it's better to frame it as a single outer pass that builds a stack and tries all methods to merge each case with the previous case until complete. Something like this maybe:

```python
def _piecewise_simplify_args(args):
    first, remaining = args[0], args[1:]
    argstack = [first]
    for case in remaining:
        newcases = [case]
        # Keep popping off the stack while we can merge cases
        while len(newcases) == 1:
            newcases = _merge_cases([argstack.pop(), newcases[0]])
        argstack.extend(newcases)
    return argstack

def _merge_cases(cases):
    while True:
        oldcases = cases
        for func in _merge_funcs:
            cases = func(cases)
            if cases != old_cases:
                break
        else:
            return cases
        
_merge_funcs = [
    _merge_cases_and_or,
    _merge_cases_ne,
    ...
]
```
Then I guess `_piecewise_simplify_args` could do any book-keeping to keep track of acccumulated conditions etc and pass those through to `_merge_cases`.

PR #22703, opened 2021-12-18
maint: add all names from commits to .mailmap

oscarbenjamin:
#### References to other Issues or PRs

Follws on from #21987 and #22650

#### Brief description of what is fixed or changed

Adds all people who are listed in any git commits to the .mailmap file. In a subsequent PR I will sort the .mailmap file into alphabetical ordering. This will make it possible for new contributors to add their identifying information to the .mailmap file without running into continuous merge conflicts. This PR just adds the missing people to .mailmap so that from now on we can require all commits to be accounted for by an entry in .mailmap (although this PR does not enforce that yet) and the AUTHORS file can be updated automatically at release time.

#### Release Notes

NO ENTRY

PR #22650, opened 2021-12-11
maint: update .mailmap and AUTHORS

oscarbenjamin:
#### Brief description of what is fixed or changed

Update .mailmap and AUTHORS files.

#### Release Notes

NO ENTRY

PR #22754, opened 2021-12-26
Check author information and the .mailmap file in CI

oscarbenjamin:

See #22725, #22711, #22703, #22650

Requires that contributors update the .mailmap file before a PR can be merged. From now on the metadata for every commit needs to be accounted for in the .mailmap file which itself needs to be maintained in sorted order.

This PR adds a script that can help with checking and ordering the .mailmap file as well as updating the AUTHORS file at release time. That script is also used to verify that the information is correct in CI.

#### Other comments

This is a second attempt at this. The previous attempt failed because of the ordering of names in the AUTHORS file. All new contributors had merge conflicts. By sorting names in the .mailmap file in lexicographic order we can avoid the merge conflicts and still preserve the ordering as listed in the AUTHORS file which corresponds more or less to the order of first contribution as determined by the ordering of commits on master.

The output of the script (which contributors will see in CI) points to here:
https://github.com/sympy/sympy/wiki/Development-workflow
That page will need to be updated to explain this after this is merged.

#### Release Notes

NO ENTRY

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -64,14 +64,8 @@
 *Marc-Etienne M.Leveille <protonyc@gmail.com>
 *Ulrich Hecht <ulrich.hecht@gmail.com>
 2torus <boris.ettinger@gmail.com>
-<abrombo@verizon.net> <brombo@GA.(none)>
-<aterrel@uchicago.edu> <andy.terrel@gmail.com>
-<aterrel@uchicago.edu> <aterrel@enthought.com>
 <felix.kaiser@fxkr.net> <kaiser.fx@gmail.com>
-<felix.kaiser@fxkr.net> <whatfxkr@gmail.com>
 <fredrik.johansson@gmail.com> <fredrik@airy.(none)>
-<jnebos@gmail.com> Jurjen N.E. Bos <devnull@localhost>
-<jo@durchholz.org> <toolforger@durchholz.org>
 Aadit Kamat <aadit.k12@gmail.com>
 Aaditya Nair <aadityanair6494@gmail.com>
 Aaron Meurer <asmeurer@gmail.com>
@@ -125,6 +119,8 @@ Akshay Siramdas <akshaysiramdas@gmail.com>
 Akshay Srinivasan <akshaysrinivasan@gmail.com>
 Akshit Agarwal <akshit.jiit@gmail.com>
 AkuBrain <76952313+Franck2111@users.noreply.github.com>
+Alan Bromborsky <abrombo@verizon.net> <abrombo@verizon.net>
+Alan Bromborsky <abrombo@verizon.net> <brombo@GA.(none)>
 Alec Kalinin <alec.kalinin@gmail.com>
 Aleksandar Makelov <amakelov@college.harvard.edu>
 Alex Argunov <sajkoooo@gmail.com>
@@ -174,6 +170,10 @@ Andrew Straw <strawman@astraw.com>
 Andrew Taber <andrew.e.taber@gmail.com>
 Andrey Grozin <A.G.Grozin@inp.nsk.su>
 Andrey Lekar <andrey_lekar@adoriasoft.com> blackyblack <andrey_lekar@adoriasoft.com>
+Andy R. Terrel <aterrel@uchicago.edu> <andy.terrel@gmail.com>
+Andy R. Terrel <aterrel@uchicago.edu> <aterrel@enthought.com>
+Andy R. Terrel <aterrel@uchicago.edu> <aterrel@uchicago.edu>
+Andy R. Terrel <aterrel@uchicago.edu> <aterrel@uchicago.edu>
 Angad Sandhu <55819847+angadsinghsandhu@users.noreply.github.com>
 Angadh Nanjangud <angadh.n@gmail.com>
 Angus Griffith <16sn6uv@gmail.com>
@@ -398,6 +398,8 @@ Fabian Pedregosa <fabian@fseoane.net> <fabian@debian.(none)>
 Faisal Anees <faisal.iiit@gmail.com>
 Faisal Riyaz <faisalriyaz011@gmail.com>
 Fawaz Alazemi <Mba7eth@gmail.com>
+Felix Kaiser <felix.kaiser@fxkr.net> <felix.kaiser@fxkr.net>
+Felix Kaiser <felix.kaiser@fxkr.net> <whatfxkr@gmail.com>
 Felix Yan <felixonmars@archlinux.org>
 Fermi Paradox <FermiParadox@users.noreply.github.com>
 Fernando Perez <Fernando.Perez@berkeley.edu>
@@ -540,6 +542,7 @@ Jens Jørgen Mortensen <jj@smoerhul.dk>
 Jeremey Gluck <jeremygluck@yahoo.com>
 Jeremias Yehdegho <j.yehdegho@gmail.com> <aevirex@aleph.(none)>
 Jeremy <twobitlogic@gmail.com>
+Jeremy Monat <jemonat@calalum.org>
 Jerry James <loganjerry@gmail.com>
 Jerry Li <jerry@jerryli.ca>
 Jezreel Ng <jezreel@gmail.com>
@@ -551,6 +554,8 @@ Jiri Kuncar <jiri.kuncar@gmail.com>
 Jisoo Song <jeesoo9595@snu.ac.kr> JSS95 <52185322+JSS95@users.noreply.github.com>
 Jisoo Song <jeesoo9595@snu.ac.kr> mcpl-sympy <59268464+mcpl-sympy@users.noreply.github.com>
 Jithin D. George <jithindgeorge93@gmail.com>
+Joachim Durchholz <jo@durchholz.org> <jo@durchholz.org>
+Joachim Durchholz <jo@durchholz.org> <toolforger@durchholz.org>
 Joan Creus <joan.creus.c@gmail.com>
 Joannah Nanjekye <joannah.nanjekye@ibm.com> Joannah Nanjekye <jnanjekye@python.org>
 Joannah Nanjekye <joannah.nanjekye@ibm.com> nanjekyejoannah <joannah.nanjekye@ibm.com>
@@ -586,6 +591,8 @@ Julien Palard <julien@palard.fr>
 Julien Rioux <julien.rioux@gmail.com>
 Julio Idichekop Filho <idichekop@yahoo.com.br>
 Jun Lin <junlin0604@gmail.com> wate123 <junlin0604@gmail.com>
+Jurjen N.E. Bos <jnebos@gmail.com> <jnebos@gmail.com>
+Jurjen N.E. Bos <jnebos@gmail.com> Jurjen N.E. Bos <devnull@localhost>
 Justin Blythe <jblythe29@gmail.com>
 Ka Yi <chua.kayi@yahoo.com.sg>
 Kaifeng Zhu <cafeeee@gmail.com>
@@ -790,6 +797,7 @@ Nikhil Gopalam <gopalamn@umich.edu> gopalamn <gopalamn@umich.edu>
 Nikhil Maan <nikhilmaan22@gmail.com> Nikhil Maan <nikhilman22@gmail.com>
 Nikhil Pappu <nkhlpappu@gmail.com>
 Nikhil Sarda <diff.operator@gmail.com>
+NikhilSDate <47920034+NikhilSDate@users.noreply.github.com>
 Nikita <nikita.student.cse19@itbhu.ac.in>
 Niklas Thörne <notrupertthorne@gmail.com>
 Nikolay Lazarov <qwerqwerqwer@abv.bg>
--- a/.mailmap
+++ b/.mailmap
@@ -1220,6 +1220,7 @@ Zeel Shah <kshah215@gmail.com>
 Zhi-Qiang Zhou <zzq_890709@hotmail.com> zhouzq-thu <zzq_890709@hotmail.com>
 Zhongshi <zj495@nyu.edu>
 Zlatan Vasović <zlatanvasovic@gmail.com>
+Zouhair <zouhair.mahboubi@gmail.com>
 Zulfikar <zulfikar97@gmail.com>
 abhinav28071999 <41710346+abhinav28071999@users.noreply.github.com>
 adhoc-king <46354827+adhoc-king@users.noreply.github.com>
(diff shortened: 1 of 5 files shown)


.mailmap
```diff
@@ -790,6 +797,7 @@ Nikhil Gopalam <gopalamn@umich.edu> gopalamn <gopalamn@umich.edu>
 Nikhil Maan <nikhilmaan22@gmail.com> Nikhil Maan <nikhilman22@gmail.com>
 Nikhil Pappu <nkhlpappu@gmail.com>
 Nikhil Sarda <diff.operator@gmail.com>
+NikhilSDate <47920034+NikhilSDate@users.noreply.github.com>
```
oscarbenjamin, line 800: @NikhilSDate what name and email do you want to be recorded in the AUTHORS file?

NikhilSDate: I would want the name to be Nikhil Date and email to be nikhil.s.date@gmail.com.

oscarbenjamin: Thanks. I've corrected the .mailmap to show this as an alias for Nikhil Date (which was already listed in the AUTHORS file).

NikhilSDate: Thanks for correcting the .mailmap entry.

oscarbenjamin:

It's important to understand that this makes a change to the workflow for new contributors so that they need to add themselves to the .mailmap file before the first PR can be merged or any time they use alternate git metadata for their commits.

This is the main blocker to automating the release process so I want this resolved before the next release.

ThePauliPrinciple:
> The output of the script (which contributors will see in CI) points to here:
> https://github.com/sympy/sympy/wiki/Development-workflow
> That page will need to be updated to explain this after this is merged.

I'd argue that this update should be written out as part of this PR in order to judge whether this is a good approach. Writing your first PR is hard as it is, there are potentially many steps that you are performing for the first time.

oscarbenjamin:
> this update should be written out as part of this PR

The update can't be directly part of the PR as the development workflow page is on the wiki rather than in the main codebase. I don't want to edit the workflow before the PR is merged but these are basically the changes I would make:

Add a section below "Create a patch file or pull request for GitHub":

## Add your name and email address to the .mailmap file.

Every author's name and email address is stored in the AUTHORS file but this file should not be edited directly. The AUTHORS file is updated automatically when a new version of SymPy is released based on the name and email addresses that are recorded in the commits. Every commit made with git stores the name and email address that git is configured with (see "Configure git settings" above). The .mailmap file is used to associate the name/email recorded in the commits with an author name and email address that will be listed in the AUTHORS file.

The first time you make a pull request you will need to add your name and email address to the .mailmap file by adding a line like
```
Joe Bloggs <joe@bloggs.com>
```
This name and email should exactly match the name and email that you have configured with git before making the commits (see "Configure git settings" above). The `bin/mailmap_check.py` script can check that this has been done correctly. If you have made a commit but not yet added yourself to the .mailmap file then you will see this:
```console
$ python bin/mailmap_check.py 
Joe Bloggs <joe@bloggs.com>

        The .mailmap file needs to be updated because there are commits with
        unrecognised author/email metadata.
        

        https://github.com/sympy/sympy/wiki/Development-workflow

                The following authors will be added to the AUTHORS file at the
                time of the next SymPy release.
```
This means that you should add your name and email address to the .mailmap file. If you add this at the end of the file then `git diff` will show:
```console
$ git diff
diff --git a/.mailmap b/.mailmap
index 3af6dc1..7fa63b1 100644
--- a/.mailmap
+++ b/.mailmap
@@ -1307,3 +1307,4 @@ zsc347 <zsc347@gmail.com>
 Øyvind Jensen <jensen.oyvind@gmail.com>
 Łukasz Pankowski <lukpank@o2.pl>
 彭于斌 <1931127624@qq.com>
+Joe Bloggs <joe@bloggs.com>
```
Now you can rerun the `bin/mailmap_check.py` script and you should see:
```console
$ python bin/mailmap_check.py 
The mailmap file was reordered

        https://github.com/sympy/sympy/wiki/Development-workflow

                The following authors will be added to the AUTHORS file at the
                time of the next SymPy release.

	Joe Bloggs <joe@bloggs.com>
```
The first line their says that the .mailmap file was "reordered". This is because the file should be in alphabetical order. The script will have moved your name into the correct position so now you can see the change as:
```console
$ git diff
diff --git a/.mailmap b/.mailmap
index 3af6dc1..7598d94 100644
--- a/.mailmap
+++ b/.mailmap
@@ -562,6 +562,7 @@ Joannah Nanjekye <joannah.nanjekye@ibm.com> Joannah Nanjekye <jnanjekye@python.o
 Joannah Nanjekye <joannah.nanjekye@ibm.com> nanjekyejoannah <joannah.nanjekye@ibm.com>
 Joaquim Monserrat <qmonserrat@mailoo.org>
 Jochen Voss <voss@seehuhn.de>
+Joe Bloggs <joe@bloggs.com>
 Jogi Miglani <jmig5776@gmail.com> jmig5776 <jmig5776@gmail.com>
 Johan Blåbäck <johan_bluecreek@riseup.net> <johan.blaback@cea.fr>
 Johan Guzman <jguzm022@ucr.edu>
```
Now if you rerun the script you will see:
```console
$ python bin/mailmap_check.py 
No changes needed in .mailmap

                The following authors will be added to the AUTHORS file at the
                time of the next SymPy release.

	Joe Bloggs <joe@bloggs.com>
```
The key information here is "No changes needed in .mailmap" which means that you have correctly updated the .mailmap file. You should now add and commit these changes as well:
```console
$ git add .mailmap
$ git commit -m 'author: add Joe Bloggs to .mailmap'
```
Sometimes a commit will be made with an incorrect name or email address or an author will make multiple commits with different names and email addresses. In this case a line should be added to the .mailmap file where the first name and email address is what should be recorded in the AUTHORS file and the others are the name and email address that was incorrectly used in the other commits. For example if the commit was recorded with the name `joeb` and the email address `wrong@email.com` but the AUTHORS file should show `Joe Bloggs` as above then there should be a line in the .mailmap file like:
```
Joe Bloggs <joe@bloggs.com> joeb <wrong@email.com>
```
A common reason that this can happen is if making commits with the GitHub web UI which always recorded the name as github username and the email as something like `1785690389+joeb@users.noreply.github.com`. In this case a line will need to be added to .mailmap like:
```
Joe Bloggs <joe@bloggs.com> joeb <1785690389+joeb@users.noreply.github.com>
```
Multiple lines like this can be added to the .mailmap file. They should record all of the different name and email address combinations that have been used by an author and map all of them to a single author name that will show in the AUTHORS file.

If your pull request is merged and you have not previously been added to the AUTHORS file then your name will be added at the time of the next release of SymPy.

asmeurer:
I think we're going to want to add even more automation to this to reduce the friction and make it easy for new contributors to understand what they need to do. It's pretty easy to use GitHub Actions to do things like comment on the PR and add a commit to the PR (or master).

bin/mailmap_check.py
```diff
@@ -0,0 +1,297 @@
...
+    write_lines(mailmap_path(), lines_mailmap_sorted)
+
+    if lines_mailmap_sorted != lines_mailmap:
+        problems = True
+        print(red("The mailmap file was reordered"))
+
+    if problems:
+        print(red("""
+        For instructions on updating the .mailmap file see:
+        https://github.com/sympy/sympy/wiki/Development-workflow"""))
```
asmeurer, line 126: Even if we don't add it now, we can add the anchor for the subheading on this page that will contain the instructions, like https://github.com/sympy/sympy/wiki/Development-workflow#update-mailmap

asmeurer:
We now have 3 scripts relating to AUTHORS and .mailmap. I think it would be best to merge them into one, with command line flags to do the different operations. We can also think about adding an interactive mode, which takes care of the most common things by asking some basic questions. 

I know this PR doesn't touch them, but I noticed in this branch (after making a test commit with a new dummy author), that 

- mailmap_update.py makes a lot of changes to .mailmap. I don't know if you ran this script before commit the changes here to .mailmap, but if you did, I think that means the sorting in it isn't consistent across runs somehow. 

- authors_update.py pops up a `less` output of `git diff`. Is that intentional? I think if we want to show the changes we should do it in a way that doesn't run through a pager (and also only shows the changes in `AUTHORS`). 

Do I understand the workflow correctly that people should not attempt to update AUTHORS, and that is now only done once at release time?

bin/mailmap_check.py
oscarbenjamin: > It's pretty easy to use GitHub Actions to do things like comment on the PR and add a commit to the PR (or master).

No, this can't be automated. The contributor has to say what name they want in the AUTHORS file. They also should be expected top understand the implication that their name is going into the AUTHORS file and why the name/email metadata in their commits is important. This is not something that anyone else can do on behalf of the contributor.

oscarbenjamin:
> We now have 3 scripts relating to AUTHORS and .mailmap. I think it would be best to merge them into one, with command line flags to do the different operations.

The other two can be deleted after this PR. The script here performs both jobs and has a command line option `--update-authors` to update the AUTHORS file.

> * mailmap_update.py makes a lot of changes to .mailmap.

I'm not sure how `mailmap_update.py` determines the order of lines in .mailmap but the script in this PR just calls `sort`. I don't see why anything else is needed.

oscarbenjamin:
> I think we're going to want to add even more automation to this to reduce the friction and make it easy for new contributors to understand what they need to do. It's pretty easy to use GitHub Actions to do things like comment on the PR and add a commit to the PR (or master).

No, this can't be automated. The contributor has to say what name they want in the AUTHORS file. They also should be expected top understand the implication that their name is going into the AUTHORS file and why the name/email metadata in their commits is important. This is not something that anyone else can do on behalf of the contributor.

asmeurer:
If those two scripts are no longer needed we should just delete them.

oscarbenjamin:
> If those two scripts are no longer needed we should just delete them.

Yes, there are a lot of things around the release scripts that need to be cleared up. Once this PR is in I can edit the workflow and then make the next release. I intend to tidy up the release files then.

That's all work that I can just do though. This PR is here to get agreement that we enable this in CI or to discuss changes that should be made to how we enable this in CI or how the script works for contributors. If that part is agreed then this can be merged and I'll get on with the rest.

asmeurer:
Is it possible to get an idea of what this looks like from the point of view of a new contributor? Maybe make a dummy PR based off this one with a commit from a dummy author (`git commit --author`).

oscarbenjamin:
> Is it possible to get an idea of what this looks like from the point of view of a new contributor?

In CI they will see the same message that I showed above:
```
Joe Bloggs <joe@bloggs.com>

        The .mailmap file needs to be updated because there are commits with
        unrecognised author/email metadata.
        

        https://github.com/sympy/sympy/wiki/Development-workflow

                The following authors will be added to the AUTHORS file at the
                time of the next SymPy release.
```

bin/mailmap_check.py
```diff
@@ -0,0 +1,297 @@
...
+    problems = False
+    missing = False
+    ambiguous = False
+    dups = defaultdict(list)
+
+    for person in git_people:
+        email = key(person)
+        dups[email].append(person)
+        if email not in who:
+            print(red("This author is not included in the .mailmap file:"""))
```
asmeurer, line 82: The extra quotes at the end here are really confusing my text editor

```suggestion
            print(red("This author is not included in the .mailmap file:"))
```

oscarbenjamin:
> Is it possible to get an idea of what this looks like from the point of view of a new contributor?

See #22794. Note also that it shows that already this PR is out of date (a new author has had a PR merged).

asmeurer:
The indentation of this line seems off

>                 The following authors will be added to the AUTHORS file at the
>                 time of the next SymPy release.

Although I'm not sure that having that there really adds much. I would remove it as it isn't relevant to the PR author.

I think this may be fine, but we should make sure the instructions are as clear as possible on what people need to do. Since .mailmap will always be up-to-date, we can basically always assume that this script failing means one of two things:

- The PR is from a new contributor
- A pre-existing contributor has committed with a new name/email

We can sometimes guess the latter case, e.g., if the same email with a different spelling already exists (I think the mailmap script already distinguishes these cases when it can). We could also use GitHub metadata to infer this as well.

We should have some sort of instructions like 

"This appears to be your first contribution to SymPy. Please add your name and email to the `.mailmap` file. If you have contributed before, please add an entry to .mailmap to indicate which name/email you prefer (see *link to dev workflow* for more information)."

or

"You appear to have contributed before with a different name/email in your git metadata. Please add an entry in .mailmap to indicate how you prefer to be credited in the SymPy AUTHORS file."

oscarbenjamin:
> Although I'm not sure that having that there really adds much. I would remove it as it isn't relevant to the PR author.

I put that there for two reasons:

1. It makes it clear to a new contributor who runs the script directly that the names will be added to the AUTHORS file later.
2. In the case that a new contributor has already had a PR merged if they make new commits before the next release with different metadata then the message will make it clear that they are already expected to be added to the AUTHORS file under the previous metadata. Hopefully someone looking at this will be able to see when there are obvious duplicates.

bin/mailmap_check.py
oscarbenjamin: I've added that.

oscarbenjamin:
I think I've addressed everything above.

This is holding up the release so I want to get it in now. Also this PR goes out of date every few days (any time a new author has a PR merged) so I don't want to leave it sitting here. The whole point of this PR is so that we don't accumulate a backlog of people whose metadata is unknown to be accurate.

Improvements can be made later as we see how this works out in practice. It doesn't sound like anyone objects to the general idea so I'll merge this shortly.

.mailmap
```diff
@@ -1211,6 +1220,7 @@ Zeel Shah <kshah215@gmail.com>
 Zhi-Qiang Zhou <zzq_890709@hotmail.com> zhouzq-thu <zzq_890709@hotmail.com>
 Zhongshi <zj495@nyu.edu>
 Zlatan Vasović <zlatanvasovic@gmail.com>
+Zouhair <zouhair.mahboubi@gmail.com>
```
oscarbenjamin, line 1223: @zouhairm how do you want your name to be recorded in the AUTHORS file?

asmeurer:
It should be clear once this is merged how well it works and what we would need to improve.

.mailmap
zouhairm: yes fine with that.

.mailmap
```diff
@@ -1295,6 +1305,7 @@ wuyudi <wuyudi119@163.com>
 ylemkimon <ylemkimon@naver.com> <mail@ylem.kim>
 znxftw <vishnu2101@gmail.com>
 zsc347 <zsc347@gmail.com>
+zzj <29055749+zjzh@users.noreply.github.com>
```
oscarbenjamin, line 1308: @zjzh how do you want your name to be recorded in the AUTHORS file?

oscarbenjamin:
Okay merging. I'll update the workflow and send a message to the mailing list.

Issue #22716, opened 2021-12-20
Improve documentation formatting by changing Sphinx theme


bertiewooster:
The headings and subheadings are too hard to distinguish on the rendered page.

_Originally posted by @asmeurer in https://github.com/sympy/sympy/issues/22693#issuecomment-997103512_

Concur; that's an issue site-wide: I have to carefully visually compare headings to tell an `<h2>` from an `<h3>` and even an `<h4>`. Should we consider changing the styling on those various heading levels? Perhaps give different levels different indenting, more dissimilar font size,...?

**Follow-up:** Comparing to [Bootstrap](https://getbootstrap.com/docs/5.1/components/buttons/) as a model of web design, a major difference is vertical spacing above headings. For example, as rendered on my laptop, Bootstrap top vertical spacing (all margin) is:

- `<h2>` headings = 48 px (pixels), or about 1.5x the font size of 31.1 px (pixels). 
- `<h3>` headings = 32 px, or about 1.17x the font size of 27.4 px.

On docs.sympy.org, headings `<h2>`, `<h3>`, and `<h4>` all have top vertical spacing (margin plus padding) of 23 px, regardless of font size (25.6, 22.4, and 19.2 px, respectively).

So differentiating SymPy's top vertical spacing by heading level could be helpful for distinguishing visual hierarchy.

_Originally posted by @bertiewooster in https://github.com/sympy/sympy/issues/22693#issuecomment-997146310_

There were discussions about switching to a better theme for the docs (see https://github.com/sympy/sympy/issues/22007). Furo was suggested, but I'm open to any theme that improves things. We should look at how Furo handles this.

_Originally posted by @asmeurer in https://github.com/sympy/sympy/issues/22693#issuecomment-997472230_

We also discussed the potential of changing the documentation theme from classic. I proposed the Furo theme used by CPython, NumPy, and Pandas, etc., but this is still up for decision and suggestions. In my view, this would give us a better layout for seamless navigation of the documentation.
_Originally posted by @nanjekyejoannah in https://github.com/sympy/sympy/issues/22007_

bertiewooster:
Because we're open to changing the theme, this issue is to discuss options. In addition to visual formatting, we should also consider accessibility: 

> Accessibility is the practice of making your websites usable by as many people as possible. We traditionally think of this as being about people with disabilities, but the practice of making sites accessible also benefits other groups such as those using mobile devices, or those with slow network connections.

[...]

Issue #15700, opened 2018-12-27
Set CSS in docs to prefer DejaVu Sans Mono


asmeurer:
The CSS in the docs should be set to prefer DejaVu Sans Mono for monospace. That way font substitution issues as described [here](https://github.com/diofant/diofant/issues/744) won't occur if that font is installed. To quote myself:

> It's a font substitution issue. The monospace font that is used for most of the characters doesn't have the ⎟character, so it is replaced from another font. The problem is that it is likely a font that isn't itself monospace, so it doesn't have the right width. Actually I'm not even sure if two different monospace fonts will have the same width as each other. I don't know if it can be fixed with CSS. The best fix I know is to install a font that has all those characters, like DejaVu Sans Mono and set your monospace font in your browser to that. Actually a simple partial fix would be to set the CSS in the docs to prefer that font if it is installed.

Here is what it looks like:

<img width="461" alt="screen shot 2018-12-27 at 2 02 08 pm" src="https://user-images.githubusercontent.com/71486/50494361-030cb700-09e0-11e9-96b9-51aad08936df.png">

Compare this to the same thing with DejaVu Sans Mono (I have it set as my monospace font in my browser settings):

<img width="514" alt="screen shot 2018-12-27 at 2 02 55 pm" src="https://user-images.githubusercontent.com/71486/50494375-1e77c200-09e0-11e9-8362-2439beecf2ba.png">

jmig5776:
I would like to work on this issue, @asmeurer can you tell me how i should change the css preference in docs because i don't know how exactly the docs work?

asmeurer:
Look up how to do CSS modifications for Sphinx docs. I think you add a custom.css to the _static directory or something like that.

jmig5776:
@asmeurer I had run the docs in my dev environment. But eventually i found that this font ```DejaVu Sans Mono``` is not available on google api fonts. Should i import this font from another site or should i  try for another font which have monospace?

asmeurer:
My suggestion is just to set the font for people who have it already installed on their computer. Making it a web font is an interesting idea. We could probably do that as well, though we would probably have to serve it ourselves. The DejaVu Sans Mono [license](https://dejavu-fonts.github.io/License.html) should allow it.

jmig5776:
Okk okk it will be seen to those who have already installed.  Yes I think having web font DejaVu Sans Mono would good as it has good looking output .

asmeurer:
[...]

PR #22755, opened 2021-12-27
DO NOT MERGE: Changing Sphinx theme to Press

bertiewooster:

Attempting to change Sphinx theme to [Press](https://pypi.org/project/sphinx-press-theme/) per issue #22716

#### Release Notes

NO ENTRY


asmeurer:
The sidebar in press doesn't expand to subsections like it does in Furo and RTD. It doesn't seem to have an option for expandable sidebars, as far as I can tell https://schettino72.github.io/sphinx_press_site/configuration.html. Unless there is a remedy for this, I would suggest not using press, as the expandable sidebar is the primary motivation for switching themes. 

I do think the general look of Press is pretty good otherwise.

asmeurer:
Another issue I just noticed with Press is that it doesn't have clickable anchors next to autodoc function headings.

asmeurer:
Here's a comparison of https://docs.sympy.org/dev/modules/matrices/matrices.html#sympy.matrices.matrices.MatrixSubspaces.orthogonalize in various themes. I screenshotted the entire page so that you can see how wide the content is in different themes, but for this analysis I'm ignoring everything on the sidebars. I included two screenshots for the themes that couldn't fit the whole thing on one page.

<details>
<summary><b>Classic (the current theme)</b></summary>
Screen Shot 2022-01-06 at 5 02 29 PM
</details>

<details>
<summary><b>Furo</b></summary>
Screen Shot 2022-01-06 at 5 02 37 PM
Screen Shot 2022-01-06 at 5 02 46 PM
</details>

<details>
<summary><b>Readthedocs</b></summary>
Screen Shot 2022-01-06 at 5 02 55 PM
Screen Shot 2022-01-06 at 5 03 01 PM
</details>

<details>
<summary><b>Press</b></summary>
Screen Shot 2022-01-06 at 5 03 05 PM
Screen Shot 2022-01-06 at 5 03 11 PM
</details>

<details>
<summary><b>Alabaster</b></summary>
Screen Shot 2022-01-06 at 5 03 15 PM
</details>

I should first note that most if not all of this styling can be edited by modifying the CSS. We definitely will want to do this for the basic background colors, but all other things being equal, it would be nice to select a theme that has decent enough styling that we don't have to mess with anything else too much. And even if we do mess with these things, maybe we can take some of the good ideas from each theme when doing so.

My thoughts:

- Only Classic and Alabaster actually fit the entire docstring on a single page in my browser, despite the fact that Alabaster is quite narrow in width. This is because the other themes add what I would consider to be excessive vertical whitespace to the page. 

- Alabaster feels quite narrow, but maybe that's just because it doesn't have a floating sidebar like the rest.

- I don't like the vertical lines to the left of the parameter descriptions in Furo and Press (I think this is how these themes render "quote" text).

- I really don't like the use of italics text in the parameter descriptions in Furo.

- I also really don't like the use of grayed out text in the parameter descriptions Press. 

Gray text and italics should both be used sparingly in my opinion, as they make text unnecessarily hard to read. They shouldn't be used for entire paragraphs.

- I've never been a fan of the red for code used by readthedocs.

- I like the "See Also" sections in Furo and Classic (or at least Classic with our theming) the best because they make it clear that they are clickable links. I somewhat dislike the smaller text size in the Furo "See Also" box, however.

- Classic and Alabaster do the best job of rendering the parameter lists, in my opinion. The rest are harder to read.

- I honestly don't really care about the choice of fonts. Maybe fonts just aren't my thing, but all the font choices here seem fine to me. The main thing that matters is if the theme properly distinguishes between different header levels (that will take a separate analysis). I will note that we do eventually want to switch to using DejaVu Sans Mono for all code in the docs because of pretty printing (https://github.com/sympy/sympy/issues/15700).

Here's a second comparison of a page with narrative documentation, from https://github.com/sympy/sympy/pull/22693.

<details>
<summary><b>Classic (the current theme)</b></summary>
Screen Shot 2022-01-06 at 5 30 36 PM
</details>

<details>
<summary><b>Furo</b></summary>
Screen Shot 2022-01-06 at 5 30 45 PM
</details>

<details>
<summary><b>Readthedocs</b></summary>
Screen Shot 2022-01-06 at 5 30 43 PM

</details>

<details>
<summary><b>Press</b></summary>
Screen Shot 2022-01-06 at 5 30 48 PM
</details>

<details>
<summary><b>Alabaster</b></summary>
Screen Shot 2022-01-06 at 5 30 40 PM
</details>

Some additional notes here:

- Classic (with our styling) puts the most content on the page, followed by Furo, with Press, RTD, and Alabaster showing the least amount of content. Part of this because each theme fixes the content to a fixed width rather than letting it overflow (n.b., this is a custom thing we did in our styling of the Classic theme; by default, it would overflow), but it's also because of excessive vertical whitespace in some of the themes (with the empty top bar in Press being the worst offender).

- Classic and Furo again are the only themes that make it clear that the internal cross-reference links are links. In the others, they just look like monospace text, and you can only tell it's a link by hovering over it (technically RTD and Alabaster style link and non-link code differently, but the distinction is not obvious).

I should note that the styling of the internal links shown in the screenshots with fixed-width font is because the link names are code (the fact that it does this even when there is custom link text as with "LaTeX parser" is a quirk of Sphinx, not the themes).

- Press has extra whitespace around its inline code styling which makes it look bad when intermixed with normal text.

- Classic and RTD are the only themes that render visited links in different colors (I'm not sure how important this is, and either way I'm pretty sure it's trivial to add or remove this in the CSS).

---

In all, I may just be biased to what I'm used to, but as far as the pure aesthetics of the styling go, I think Classic looks the best. My ranking based purely on the styling would be (from best to worst):

Classic
Alabaster
Readthedocs
Furo
Press

I really see no upside to Press, and given the [other issues](https://github.com/sympy/sympy/pull/22755#issuecomment-1007008305) I found with it, and the fact that it's not as popular as the others, I would suggest removing it from consideration, unless anyone else wants to vouch for it. 

The others all have advantages and disadvantages from my point of view. 

And like I said, most of these issues can be remedied with custom CSS. What's harder to do is the sidebar. So I suspect we will end up picking a theme that has a decent sidebar and cleaning up its styling in the CSS. That is unless there is some way to add a better sidebar to a theme that doesn't have a particularly good one like "Classic" or "Alabaster". That probably pits it between Furo or RTD (I like the sidebar aspects of Furo the best so far), but maybe we can take the above into consideration when building a custom CSS, which we should do anyway for color styling.

oscargus:
I restart this as now `doc/requirements.txt` is used for building the docs.

Issue #22757, opened 2021-12-27
manualintegrate outputs wrong result


mizusato:
The expression `manualintegrate(sin(x),y)` produces a wrong result.
```
>>> from sympy import *
>>> from sympy.integrals.manualintegrate import *
>>> x,y = symbols('x y')
>>> manualintegrate(sin(x),y)  // wrong
-cos(x)
>>> manualintegrate(log(x),y)  // correct
y*log(x)
```
version: SymPy 1.9

MiguelAMLucero:
Is this issue available to work on?

smichr:
It is open and there is no referencing PR...it's available.

dianet97:
@smichr  @oscarbenjamin manualintegrate doesn't seem to work for any of the standard trigonometric functions, i.e,
`manualintegrate(trig(x),y)` 
returns the wrong answer. Weirdly enough it does work for inverse trigonometric functions (atan,asin,acos,acot,acsc,asec). As an example:
`>>>>manualintegrate(asin(x),y) `
`y*asin(x)`
This begs the question if Log and these inverse trig functions are treated similarly in the code (for this instance anyways)? Turning on the debugging and stepping through we find that   `def integral_is_subclass(*klasses):` in manualintegrate.py  hits a *kclasses argument of `(<class 'sympy.core.mul.mul>, log, atan, asin, acos, acot, acsc, asec)` at some point

MiguelAMLucero:
@smichr @oscarbenjamin @ddtcode when you run
 `>>> integral_steps(sin(x), y)` 
the output is `TrigRule(func='sin', arg=x, context=sin(x), symbol=y)`, 
but when you run `>>> integral_steps(log(x),y)` 
the rule comes out as `ConstantRule(constant=log(x), context=log(x), symbol=y)`. 
The same ConstantRule comes out with inverse trig functions. It should detect trig functions in this case as constants.
Also, I tried to run the same integral in symPy gamma, and this happened
<img width="635" alt="Screen Shot 2021-12-28 at 20 40 19" src="https://user-images.githubusercontent.com/92400491/147628468-6400acc9-9c47-4974-8ad8-e0d6b3d2c432.png">

ddtcode:
@MiguelAMLucero Sorry for not "@-ing" you in my previous message!

ddtcode:
[...]

PR #22765, opened 2021-12-29
change of if/else order in def key code 

dianetc:
References to other Issues or PRs
#22757 

Essentially sin(x)  along with other trigonometric functions gets registered as a trigonometric function instead of a constant when  taking the integrand with respect to another variable. Consequently, the `TrigRule` is applied leading to an incorrect result. Other non-trigonometric functions (including inverse trig functions) would not enter the initial `if` leading to the correct ruling.

A possible fix is switching the order of the if/else clause in ` def key()`

#### Release Notes

* functions
  * Fixed a bug with the indefinite integral of trigonometric functions with respect to a different variable in `manualintegrate.py'

### DIFF
--- a/sympy/integrals/tests/test_manual.py
+++ b/sympy/integrals/tests/test_manual.py
@@ -555,16 +555,3 @@ def test_quadratic_denom():
 
 def test_issue_22757():
     assert manualintegrate(sin(x), y) == y * sin(x)
-    assert manualintegrate(cos(x), y) == y * cos(x)
-    assert manualintegrate(tan(x), y) == y * tan(x)
-    assert manualintegrate(cot(x), y) == y * cot(x)
-    assert manualintegrate(sec(x), y) == y * sec(x)
-    assert manualintegrate(csc(x), y) == y * csc(x)
-
-    assert manualintegrate(sin(x) * cos(x), y) == y * sin(x) * cos(x)
-    assert manualintegrate(-sec(x) * tan(x), y) == y * -sec(x) * tan(x)
-    assert manualintegrate(csc(x) * cot(x), y) == y * csc(x) * cot(x)
-    assert manualintegrate(sec(x)**2, y) == y * sec(x)**2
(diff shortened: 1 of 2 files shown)


ThePauliPrinciple:
Some formal points:

Can you add a test for this? 

Can you rewrite the description? There are some typos in there. Additionally I think it would be helpful to mention also that you are talking about integrating over `y`, because in other contexts `cos(x)` would not normally be thought of as a constant.

I don't think the release note for `solvers` is helpful since it doesn't explain anything that has changed for the user. (and is vague)

ddtcode:
@ThePauliPrinciple For the test, should I create a whole separate function in ``test_manual.py` or is it fine to add some asserts in `test_manualintegrate_trigonometry()`

ThePauliPrinciple:
Considering this PR is the result of an issue, it should be in a function with the issue number in it.

e.g.
```python
def test_issue_22757():
    assert ...
```

ddtcode:
@ThePauliPrinciple I had changed it to be that way. The error is
`AssertionError: File ends with more than 1 newline`

sympy/integrals/tests/test_manual.py
```diff
@@ -553,3 +552,19 @@ def test_quadratic_denom():
     assert manualintegrate(f, x) == 5*log(3*x**2 - 2*x + 8)/6 + 11*sqrt(23)*atan(3*sqrt(23)*(x - Rational(1, 3))/23)/69
     g = 3/(2*x**2 + 3*x + 1)
     assert manualintegrate(g, x) == 3*log(4*x + 2) - 3*log(4*x + 4)
+
+def test_issue_22757():
```
smichr, line 556: The tests are more than are needed. It feels underwhelming to add a single test, but a single one would suffice to show that the issue has been addressed (and it is clear to the reviewer that this is not fixed for only one case). So maybe just the first test is sufficient.

smichr:
Thanks, this is in.

PR #22767, opened 2021-12-30
Fix arsinh, arcosh, etc. full names

TomFryers:
The full names of the inverse hyperbolic trigonometric functions (arsinh, arcosh, artanh, arsech, arcsch, arcoth) don't have a ‘c’, so this fixes them in the LaTeX output when `inv_trig_style="full"`.

NO ENTRY

### DIFF
--- a/sympy/printing/latex.py
+++ b/sympy/printing/latex.py
@@ -915,7 +915,7 @@ class LatexPrinter(Printer):
                 if inv_trig_style == "abbreviated":
                     pass
                 elif inv_trig_style == "full":
-                    func = "arc" + func[1:]
+                    func = ("ar" if func[-1] == "h" else "arc") + func[1:]
                 elif inv_trig_style == "power":
                     func = func[1:]
                     inv_trig_power_case = True
--- a/sympy/printing/tests/test_latex.py
+++ b/sympy/printing/tests/test_latex.py
@@ -469,7 +469,7 @@ def test_latex_functions():
     assert latex(acsc(x), inv_trig_style="full") == \
         r"\operatorname{arccsc}{\left(x \right)}"
     assert latex(asinh(x), inv_trig_style="full") == \
-        r"\operatorname{arcsinh}{\left(x \right)}"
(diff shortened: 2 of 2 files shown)


oscargus:
I was quite convinced that there should be a c. Not the best reference, but e.g. https://en.wikipedia.org/wiki/Inverse_trigonometric_functions claims that.

However, it should be possible to add a `inv_trig_style="ar"` to support the ar-approach.

TomFryers:
I think that Wikipedia page is just about the inverse circular trigonometric functions (arcsin, arccos, etc.), which definitely do have a 'c'. I mean the [inverse hyperbolic trigonometric functions](https://en.wikipedia.org/wiki/Inverse_hyperbolic_functions).

ThePauliPrinciple:
> I mean the [inverse hyperbolic trigonometric functions](https://en.wikipedia.org/wiki/Inverse_hyperbolic_functions).

In particular from that wiki page:
"The ISO 80000-2 standard abbreviations consist of ar- followed by the abbreviation of the corresponding hyperbolic function (e.g., arsinh, arcosh). The prefix arc- followed by the corresponding hyperbolic function (e.g., arcsinh, arccosh) is also commonly seen, by analogy with the nomenclature for inverse trigonometric functions. These are misnomers, since the prefix arc is the abbreviation for arcus, while the prefix ar stands for area; the hyperbolic functions are not directly related to arcs.[9][10][11]"

oscargus:
Sorry, I didn't understand that it was only the hyperbolic functions that were affected. I had the impression that it was arcos, arsin etc... I totally agree then (and learnt something new).

oscargus:
The thing that needs to be fixed is this test:
https://github.com/sympy/sympy/blob/b0b66d5511575eda9f0e3c7cc048b7bf63a85706/sympy/printing/tests/test_latex.py#L471-L472

oscargus:
Thanks!

The failing test is unrelated.

oscargus:
Restarting tests and enabling auto-merge.

PR #22775, opened 2022-01-01
Allow complex transfer functions in pole-zero plot

oscargus:

The pole zero plot could only work with real-valued transfer functions (complex conjugate pole-zero pairs). No harm in allowing support for complex-valued expressions. (Realized this while I was playing around with a non-working filter design...)

#### Other comments

Also added a comment regarding a sub-optimal check. Expressions of the type `exp(s)` should not be allowed, but `s*exp(1)` can be handled. Now both are disallowed. Better to check if the numerator and denominator are polynomials in `s`.

#### Release Notes

* physics.control
   * `pole_zero_plot` now supports complex-valued transfer functions.

### DIFF
--- a/sympy/physics/control/control_plots.py
+++ b/sympy/physics/control/control_plots.py
@@ -42,6 +42,8 @@ def _check_system(system):
             " that there are no free symbols in the dynamical system other"
             " than the variable of Laplace transform.")
     if sys.has(exp):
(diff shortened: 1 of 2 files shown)


oscargus:
This should be quite non-controversial, as in there is not point in restricting it to real-only transfer functions, so I self-merge.

PR #22778, opened 2022-01-01
Add Jacobi theta and elliptic functions

oscargus:

I needed the Jacobi elliptic integrals for another thing I am working on, so I added those and the Jacobi theta function. Still there are things to do, but I wanted to get some feedback on the design so far.

My main doubt about these is that they have a first argument that completely determines the function. In e.g. Mathematica they have specific functions for each variant (in addition to the general that takes a type argument). This can be useful when writing things, but also when rewriting as it is possible to explicitly say which form is required. 

An option will be to be able to do something like:
```
sn = JacobiEllipticIntegral('sn')
v = sn(x, y)
```
But I have not managed to get this to work (in combination with `JacobiEllipticIntegral('sn', x, y)` still working).

Things to do:
- [x] Decide on the single/multiple class issue above
- [x] Decide on naming
- [ ] Add tests
- [ ] Add documentation
- [ ] Add printer support
   - [x] LaTeX
   - [x] Pretty
   - [x] Mathematica
   - [x] Maple
   - [ ] NumPy/SciPy
   - [ ] Matlab/Octave
   - [ ] MathML

#### Other comments

The naming is not obvious.
* `JacobiTheta`: `jacobi_theta`? `elliptic_theta`?
* `JacobiEllipticFunction`: `jacobi_elliptic`? `elliptic_jacobi`? `elliptic_j`?

One may consider passing a nome instead of q and m. See https://mpmath.org/doc/current/functions/elliptic.html However, the other elliptic integrals do not use that.

#### Release Notes

* functions
   * Jacobi theta and elliptic functions added: `theta1`, etc and `jacobi_cd` etc.

### DIFF
--- a/sympy/__init__.py
+++ b/sympy/__init__.py
@@ -130,7 +130,9 @@ from .functions import (factorial, factorial2, rf, ff, binomial,
         chebyshevu, chebyshevu_root, chebyshevt_root, laguerre,
         assoc_laguerre, gegenbauer, jacobi, jacobi_normalized, Ynm, Ynm_c,
         Znm, elliptic_k, elliptic_f, elliptic_e, elliptic_pi, theta1, theta2,
-        theta3, theta4, elliptic_nome_q, jacobi_cd, beta, mathieus, mathieuc,
+        theta3, theta4, elliptic_nome_q, jacobi_cd, jacobi_cn, jacobi_cs,
+        jacobi_dc, jacobi_dn, jacobi_ds, jacobi_nc, jacobi_nd, jacobi_ns,
+        jacobi_sc, jacobi_sd, jacobi_sn, beta, mathieus, mathieuc,
         mathieusprime, mathieucprime, riemann_xi, betainc, betainc_regularized)
 
 from .ntheory import (nextprime, prevprime, prime, primepi, primerange,
@@ -355,7 +357,9 @@ __all__ = [
     'elliptic_k', 'elliptic_f', 'elliptic_e', 'elliptic_pi', 'beta',
     'mathieus', 'mathieuc', 'mathieusprime', 'mathieucprime', 'riemann_xi','betainc',
     'betainc_regularized', 'theta1', 'theta2', 'theta3', 'theta4',
-    'elliptic_nome_q', 'jacobi_cd',
+    'jacobi_cd', 'jacobi_cn', 'jacobi_cs', 'jacobi_dc', 'jacobi_dn',
+    'jacobi_ds', 'jacobi_nc', 'jacobi_nd', 'jacobi_ns', 'jacobi_sc',
+    'jacobi_sd', 'jacobi_sn', 'elliptic_nome_q',
 
     # sympy.ntheory
     'nextprime', 'prevprime', 'prime', 'primepi', 'primerange', 'randprime',
--- a/sympy/__init__.py
+++ b/sympy/__init__.py
@@ -129,8 +129,9 @@ from .functions import (factorial, factorial2, rf, ff, binomial,
         meijerg, appellf1, legendre, assoc_legendre, hermite, chebyshevt,
         chebyshevu, chebyshevu_root, chebyshevt_root, laguerre,
         assoc_laguerre, gegenbauer, jacobi, jacobi_normalized, Ynm, Ynm_c,
-        Znm, elliptic_k, elliptic_f, elliptic_e, elliptic_pi, beta, mathieus,
-        mathieuc, mathieusprime, mathieucprime, riemann_xi, betainc, betainc_regularized)
+        Znm, elliptic_k, elliptic_f, elliptic_e, elliptic_pi, JacobiTheta,
+        JacobiEllipticFunction, beta, mathieus, mathieuc, mathieusprime,
+        mathieucprime, riemann_xi, betainc, betainc_regularized)
 
 from .ntheory import (nextprime, prevprime, prime, primepi, primerange,
         randprime, Sieve, sieve, primorial, cycle_length, composite,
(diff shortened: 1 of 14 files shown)


jksuom:
Bessel functions `besselj`, etc. are defined as subclasses of `BesselBase` that contains code common to all Bessel functions. Maybe we could define `jacobi_cn`, etc. as subclasses of e.g. `JacobiEllipticBase`.
`

oscargus:
There are two separate cases I think.

For the theta functions they are basically the same function but with different constants and offsets (so they can at least be seemingly written differently). For the elliptic functions, there are three different functions (plus a one) that are combined using different in the numerator/denominator. (You are probably aware of this, but others may not.)

So I am not so sure how much code there will be in each subclass. I had that approach first and then it was basically only the type string for the latter case. (One can have different `_eval` methods as well, but the current one is quite compact...)

Maybe the best thing is to make an alternative PR with this to show what it may look like?

oscarbenjamin:
> So I am not so sure how much code there will be in each subclass.

It's fine to have subclasses with very little code. I think that's the point: subclasses are cheap so we can use a different class for each. Instead of having a single class with a string arg to distinguish the cases.

oscargus:
OK! So subclasses there is. What about naming? `jacobi_sn` etc seems to make sense. `jtheta1`?

jksuom:
I think that it would not be necessary to add `j` or `jacobi` to `theta`. There is practically no risk of confusion. If Riemann't theta is ever implemented, then we can add his name. And Weierstrass' theta function is called `sigma`.

As to the indices, I personally like Mumford's presentation in 'Tata lectures on theta I`. He writes the functions as `theta_xy` where `xy` is 00, 01, 10 or 11.

oscargus:
> As to the indices, I personally like Mumford's presentation in 'Tata lectures on theta I'. He writes the functions as `theta_xy` where `xy` is 00, 01, 10 or 11.

Oh, yes, that is another aspect. Am I correct that these are basically the binary representation (+1) of 1 to 4? I didn't fully get it when trying to check it up.

oscargus:
Or it is maybe not that easy... https://en.wikipedia.org/wiki/Jacobi_theta_functions_(notational_variations)

oscargus:
For what it is worth, `mpmath` has defined 1 through 4. But as long as we agree on some notation and I understand how to map it to `mpmath` that should be fine.

(I think that there are systematic, "well looking", relations between the elliptic functions and theta{1-4}, but it should of course be possible to modify.)

oscargus:
I've come back to this, but before spending too much time on it, can we agree on the naming?`

I've gone with `theta1` etc as this follows mpmath and I couldn't really understand the mapping to/from `theta00` etc. I guess that those can be added later on.

For the Jacobi elliptic functions, I went with `jacobi_cd` etc. I think that `cd` is basically too short to make sense...

There are still eleven `jacobi_xx`-functions to define and probably one would like to use https://functions.wolfram.com/EllipticFunctions/JacobiAmplitude/ as this shows up in some derivatives, although I cannot really figure out how to evaluate it numerically in mpmath.

oscarbenjamin:
> I've come back to this, but before spending too much time on it, can we agree on the naming?

I don't know a lot about these functions but here's a quick survey of other names used in related software:

1. mpmath calls these `jtheta` and uses an integer first argument like `jtheta(1, z, q)`.
2. It looks like SciPy doesn't have these.
3. Maple uses `JacobiTheta1(z, q)`.
4. Mathematica uses `EllipticTheta[1, z, q]`.
5. Matlab doesn't seem to have these.

The only consistent thing seems to be including `theta` in the name but not *only* calling it theta. I think in that case `theta1` etc seems fine although I don't have any strong preference between `theta1` and `jtheta1`.

sylee957:
I'd just prefer the capitalized name for classes. 
I think that we should just supply trivial constructors `def jacobi_sn(...)` for stuff that should be function

HeskethGD:
Hi, I was wondering what the status of this was and if there was a reason why it was not merged or continued? Weierstrass elliptic functions were recently added to mpmath and I was contemplating getting involved to add them to sympy. Would that be welcome and if so, how might it relate to this PR, should this be fixed and merged first or would it be better to start a new PR with them all in but take this work as the starting reference? Thanks.

oscarbenjamin:
I think if you want to take this forward then probably take the commits from here onto a new branch (preserving authorship), fix up the merge conflicts and open a new PR.

HeskethGD:
> I think if you want to take this forward then probably take the commits from here onto a new branch (preserving authorship), fix up the merge conflicts and open a new PR.

There are 10 places in the [`elliptic_integral.py`](https://github.com/sympy/sympy/pull/22778/changes#diff-c4d8212328f216a87b60a94c88db8bbd8911a27549087554ae3f857da72ed10d) file where it says "(wrong)" as a comment in the doc string i.e. things like:
```
`Jacobi elliptic function: (wrong)`
```
So thats why I was hoping to get some clarification as to whether this is a good starting point or not. I was also thinking of introducing the Jacobi theta functions first as a single PR, then add Jacobi elliptic, then Weierstrass to keep it a bit smaller.

oscarbenjamin:
Okay maybe it is better to start fresh with something a bit smaller than this PR then. Not adding all the functions at once makes sense. We should make sure we know what the names (and arguments) of all of the functions would be before adding any of them though.

HeskethGD:
> Okay maybe it is better to start fresh with something a bit smaller than this PR then. Not adding all the functions at once makes sense. We should make sure we know what the names (and arguments) of all of the functions would be before adding any of them though.

I have moved the discussion re names etc for three future PRs to this discussion https://github.com/sympy/sympy/discussions/30211. There I have outlined a provisional plan. I also took the liberty of drafting the first PR re Jacobi theta only which is currently ready for review https://github.com/sympy/sympy/pull/30205

PR #22780, opened 2022-01-01
Replaced pymc3 with pymc

oscargus:

Trying to get the optional dependency tests to pass.

#### Release Notes

NO ENTRY


oscarbenjamin:
What's the difference between pymc and pymc3 here?

oscargus:
pymc doesn't use theano and there seems to be some issues for all optional dependency tests caused by theano, e.g.
```
_______ sympy/stats/tests/test_matrix_distributions.py:test_sample_seed ________
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/theano/configparser.py", line 238, in fetch_val_for_key
    return self._theano_cfg.get(section, option)
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/configparser.py", line 781, in get
    d = self._unify_values(section, vars)
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/configparser.py", line 1152, in _unify_values
    raise NoSectionError(section) from None
configparser.NoSectionError: No section: 'blas'
```

oscargus:
The not so obvious thing is that with pymc version 2.3.8 is downloaded... Right now they are about to release version 4, but it clearly states "PyMC (formerly PyMC3)"

https://github.com/pymc-devs/pymc

So probably I am using an older version, but the most recent pymc3 stopped working. When 4.0 is released this change will use the most recent, as opposed to pymc3. But maybe I am not getting everything. 

Not clear why something stopped working yesterday(?)...

oscarbenjamin:
> Not clear why something stopped working yesterday(?)...

No, it's not clear. Maybe @brandonwillard knows...

oscargus:
I added a PR to aesara that fixes it. Seems like aesara (and theano) access a numpy config directly in the file and now that config is removed.

https://github.com/aesara-devs/aesara/pull/716

brandonwillard:
> I added a PR to aesara that fixes it. Seems like aesara (and theano) access a numpy config directly in the file and now that config is removed.
> 
> [aesara-devs/aesara#716](https://github.com/aesara-devs/aesara/pull/716)

Yeah, that's exactly it; some of the old Theano code is doing something unnecessary and no longer supported.

oscarbenjamin:
> some of the old Theano code is doing something unnecessary and no longer supported.

Is this an issue for asarea as well?

brandonwillard:
> > some of the old Theano code is doing something unnecessary and no longer supported.
> 
> Is this an issue for aesara as well?

Yes, it has inherited that same code, but @oscargus has a fix that we can merge and release today: https://github.com/aesara-devs/aesara/pull/716.

oscarbenjamin:
I guess after that fix SymPy still needs to drop Theano from testing then.

brandonwillard:
> I guess after that fix SymPy still needs to drop Theano from testing then.

Yeah, that would probably be best.

oscargus:
The problem is that pymc3 depends on theano, SymPy does not import theano directly. But it seems like pymc has 4.0 in later stages, so I guess that this PR will become handy in the future. Right now, it installs 2.3.8 which is (at least) seven years old, but seems to not use the problematic code.

So if a new aesara version comes out quickly, it will probably make sense to skip pymc3 for a while. Also, the new version seems quite disruptive, even though I have no idea how much affects SymPy. https://github.com/pymc-devs/pymc/blob/main/RELEASE-NOTES.md

oscargus:
> import theano directly

As in: the optional dependency tests do not install theano directly.

oscarbenjamin:
I'm not sure I'm keeping track of all this correctly but doesn't this PR mean that we can drop the implicit dependency on Theaono by using PyMC instead of PyMC3?

oscarbenjamin:
@oscargus to be clear I haven't looked at this in detail but it seems like you have. I'm +1 to any approach that fixes CI right now if you have a PR that can fix that.

oscargus:
> using PyMC instead of PyMC3?

Yes, because it works (if the tests are restarted, the latest version of aesara will be used and the errors gone).

No, because currently using pymc gives such an old version that theano is not used, so it may not be the best idea. 

However, in the "near future", pymc 4 is released and will, if I understand it correctly, go back to the pymc naming. If the major changes in that doesn't affect SymPy, it should be possible to use this PR then.

For now, I believe disabling the pymc3 install as in #22783 is the best way forward. Easy to revert and nothing says that this PR will actually work when pymc 4 is available.

PR #22770, opened 2021-12-31
Corrected prior form leading to dead code in sympy.polys.polyroots.roots() function

faze-geek:
#### References to other Issues or PRs

Fixes #22768

#### Brief description of what is fixed or changed
Made appropriate changes which handle dead code in` sympy.polys.polyroots.roots() ` function .
#### Release Notes

NO ENTRY

``

PR #22781, opened 2022-01-01
activate dead code to factor constant term when solving binomial equation

### LINKED ISSUE #22768: Dead code in sympy.polys.polyroots.roots() function
While messing around with `roots()` method I saw things...

https://github.com/sympy/sympy/blob/27613158382a668605a21616706a8d64a3573997/sympy/polys/polyroots.py#L891

Here `f` is a `Poly` instance, and `Poly.length()` is a function.
So this expression is always `False` and followed block of code is dead.
Not sure if this can affect `roots()` correct behavior.

smichr:
Code was dead because `f.length` is a method that can never equal 2. Code has now been corrected and updated to handle the re-factoring of the constant term that is expanded when the Poly is made.

fixes #22768
closes #22770 

#### Release Notes

* solvers
    * solve will give more factored results automatically when solving binomial expressions with similar powers in the constant term -- like when solving `x**3 - (1 - y)**6` for `x`.

### DIFF
--- a/sympy/solvers/tests/test_solvers.py
+++ b/sympy/solvers/tests/test_solvers.py
@@ -2478,7 +2478,8 @@ def test_solver_flags():
 
 
 def test_issue_22768():
-    assert solve(2*x**3 - 16*(y - 1)**6*z**3, x, simplify=False
+    eq = 2*x**3 - 16*(y - 1)**6*z**3
+    assert solve(eq.expand(), x, simplify=False
         ) == [2*z*(y - 1)**2, z*(-1 + sqrt(3)*I)*(y - 1)**2,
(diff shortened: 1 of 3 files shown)


sympy/solvers/tests/test_solvers.py
```diff
@@ -2477,6 +2477,12 @@ def test_solver_flags():
     assert root != rad
 
 
+def test_issue_22768():
+    assert solve(2*x**3 - 16*(y - 1)**6*z**3, x, simplify=False
```
smichr, line 2481:
```suggestion
    eq = 2*x**3 - 16*(y - 1)**6*z**3
    assert solve(eq.expand(), x, simplify=False
```

PR #22783, opened 2022-01-01
Temporarily disable pymc3

oscargus:

Disable ~~aesara and~~ pymc3 as there is a bug when combined with newest NumPy.

~~Bug: https://github.com/aesara-devs/aesara/issues/715~~
~~Solution(?): https://github.com/aesara-devs/aesara/pull/716~~

#### Release Notes

NO ENTRY

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -130,7 +130,7 @@ jobs:
           python-version: '3.9'
       - run: sudo apt install antlr4 libgfortran5 gfortran libmpfr-dev libmpc-dev
       - run: python -m pip install --upgrade pip
-      - run: pip install mpmath gmpy2 matplotlib numpy scipy pymc3 aesara ipython \
+      - run: pip install mpmath gmpy2 matplotlib numpy scipy aesara ipython \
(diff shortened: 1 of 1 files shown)


oscargus:
The option is to limit NumPy to version 1.21.5.

oscarbenjamin:
> The option is to limit NumPy to version 1.21.5.

That would make sense if we expect that numpy might fix this. Is there a numpy issue for this?

oscargus:
I'd say it is an aesara issue. Basically they try to read directly from a file rather than querying:

https://github.com/aesara-devs/aesara/blob/ff4b91dddede43d63d348bdec8fe1079932146b0/aesara/link/c/cmodule.py#L2625-L2646

oscargus:
Pushing again with only pymc3 removed (should pass if the providers are updated with the new aesara)-

oscarbenjamin:
Okay looks good to me based on comments elsewhere (https://github.com/sympy/sympy/pull/22780#issuecomment-1003623031)

I'll merge this now

We need to make sure to remember to enable this again once there is a new version of pymc that we can use (CC @brandonwillard).

PR #22784, opened 2022-01-01
Sequences: New basic combinatorial sequences returning an initial segment of values.

PeterLuschny:
Sequences: New basic combinatorial sequences returning an initial segment of values.

This module implements some special sequences that commonly appear in combinatorial contexts. The functions return the initial segment of given size for these sequences.

The need for such a module arises from the fact that sequences often can be computed much more efficiently (for example via recurrences) than by list comprehension.

A convincing example is the computation of the Euler numbers. To compute the first 500 nonzero Euler numbers with 'euler_sequence' is more than 1000 times faster than with list comprehension of 'euler' (bench-marked with SymPy 1.10 and Python 3.10.)

Moreover the typical use case often requires the initial segment of sequences. For example Bernoulli numbers rarely come alone: their typical use is in the Taylor-MacLaurin expansion of functions which needs B0, B2, B4,..., B2n up to some n.

Contrary to the 'bernoulli' and 'euler' functions the sequences in this module do not depended on the mpmath module.

Currently the following sequences are implemented:
* andre_sequence
* bell_sequence
* bernoulli_sequence
* euler_sequence
* tangent_sequence

Example:

To compute the first 250 nonzero Bernoulli numbers with 'bernoulli_sequence' is about 70 times faster than with list comprehension (and needs no mpmath).

[bernoulli(n) for n in range(0, 500, 2)]
1743.085 ms

bernoulli_sequence(250)
25.568 ms

More examples, benchmarks and details can be found at:
http://luschny.de/math/sympy/Proposals.html

* functions
  * New functions returning initial segments of sequences of the Bernoulli, Euler, Bell, Andre, and Tangent numbers.

### DIFF
--- a/sympy/core/tests/test_args.py
+++ b/sympy/core/tests/test_args.py
@@ -2019,6 +2019,31 @@ def test_sympy__functions__combinatorial__numbers__partition():
     assert _test_args(partition(Symbol('a', integer=True)))
 
 
+def test_sympy__functions__combinatorial__sequences__bernoulli_sequence():
+    from sympy.functions.combinatorial.sequences import bernoulli_sequence
+    assert _test_args(bernoulli_sequence(x))
+
+
+def test_sympy__functions__combinatorial__sequences__andre_sequence():
+    from sympy.functions.combinatorial.sequences import andre_sequence
+    assert _test_args(andre_sequence(x))
+
+
+def test_sympy__functions__combinatorial__sequences__euler_sequence():
+    from sympy.functions.combinatorial.sequences import euler_sequence
+    assert _test_args(euler_sequence(x))
+
+
+def test_sympy__functions__combinatorial__sequences__tangent_sequence():
+    from sympy.functions.combinatorial.sequences import tangent_sequence
+    assert _test_args(tangent_sequence(x))
+
+
+def test_sympy__functions__combinatorial__sequences__bell_sequence():
+    from sympy.functions.combinatorial.sequences import bell_sequence
+    assert _test_args(bell_sequence(x))
+
+
 def test_sympy__functions__elementary__complexes__Abs():
     from sympy.functions.elementary.complexes import Abs
     assert _test_args(Abs(x))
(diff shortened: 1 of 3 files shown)


oscargus:
Thanks!

The optional dependencies test failure is caused by other issues. However, there is a requirement that all classes are tested in core/tests/test_args.py and that the test is named in a specific way (should be straightforward). Please do these tests using symbolic arguments as part of the test is that it can recreate itself from its arguments and if it evaluates to something else (like a list in this case) the purpose of the test will be lost. (We will test that this doesn't happen later on, but right now some of the older classes does not support this...)

oscarbenjamin:
It doesn't look to me like these should be symbolic functions. They return Python lists as far as I can tell. These should just be ordinary Python functions somewhere.

PeterLuschny:
> It doesn't look to me like these should be symbolic functions. They return Python lists as far as I can tell. These should just be ordinary Python functions somewhere.

What do I have to change for this?

oscarbenjamin:
> What do I have to change for this?

You currently have symbolic functions defined like this:
```python
class andre_sequence(Function):
    @classmethod
    def eval(cls, size):
        s = sympify(size)
        if not s.is_Integer:
             return
        return ...
```
You can just write this as
```python
def andre_sequence(size):
    return ...
```

PeterLuschny:
Well, if there is no fundamental reason for the class notation I will change that.

ThePauliPrinciple:
In SymPy `Function` refers to a function in the mathematical sense, not the computer science one. It might make sense to define sequence "symbols" if there is any case where something like `bernoulli_sequence(x)/bernoulli_sequence(x+1)=<a different sequence>(x)` exists.

anutosh491:
> It doesn't look to me like these should be symbolic functions. They return Python lists as far as I can tell. These should just be ordinary Python functions somewhere.

Yeah True , this was my first thought when I saw the methods introduced by this pr . Not sure where should the code be added for the same !

PR #22785, opened 2022-01-02
Basic: __eq__ now only uses sympify for user defined objects

ThePauliPrinciple:

Helps with #22607

`Basic.__eq__` now no longer always tries to convert non-Basic objects using `sympify`, instead, it only does so if the object has `_sympy_` defined, or if its type has been added `converter` by the user. Any classes in SymPy will have to explicitly add logic to compare to non-Basic objects, this is to encourage using properties directly, which should be more efficient than creating a SymPy object. 

In order for this to work, `BooleanFalse` and `BooleanTrue` now explicitly check if `other` is `True`/`False`, rather than attempt to convert `other` to `BooleanFalse`/`BooleanTrue`.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -7,14 +7,14 @@ from typing import Set, Tuple, Any
 from .assumptions import ManagedProperties
 from .cache import cacheit
 from .core import BasicMeta
-from .sympify import _sympify, sympify, SympifyError
+from .sympify import _sympify, sympify, SympifyError, _external_converter
 from .sorting import ordered
 from .kind import Kind, UndefinedKind
 from ._print_helpers import Printable
 
 from sympy.utilities.decorator import deprecated
 from sympy.utilities.exceptions import SymPyDeprecationWarning
-from sympy.utilities.iterables import iterable, numbered_symbols, NotIterable
+from sympy.utilities.iterables import iterable, numbered_symbols
 from sympy.utilities.misc import filldedent, func_name
 
 from inspect import getmro
@@ -319,6 +319,25 @@ class Basic(Printable, metaclass=ManagedProperties):
         args = len(args), tuple([inner_key(arg) for arg in args])
         return self.class_key(), args, S.One.sort_key(), S.One
 
+    def _do_eq_sympify(self, other):
+        """Returns a boolean indicating whether a == b when either a
+        or b is not a Basic. This is only done for types that were either
+        added to `converter` by a 3rd party or when the object has `_sympy_`
+        defined. This essentially reuses the code in `_sympify` that is
+        specific for this use case. Non-user defined types that are meant
+        to work with SymPy should be handled directly in the __eq__ methods
+        of the `Basic` classes it could equate to and not be converted. Note
+        that after conversion, `==`  is used again since it is not
+        neccesarily clear whether `self` or `other`'s __eq__ method needs
+        to be used."""
+        for superclass in type(other).__mro__:
+            conv = _external_converter.get(superclass)
+            if conv is not None:
+                return self == conv(other)
+        if hasattr(other, '_sympy_'):
+            return self == other._sympy_()
+        return NotImplemented
+
     def __eq__(self, other):
         """Return a boolean indicating whether a == b on the basis of
         their symbolic trees.
@@ -344,16 +363,7 @@ class Basic(Printable, metaclass=ManagedProperties):
             return True
 
         if not isinstance(other, Basic):
-            if iterable(other, exclude=(str, NotIterable)
-                    ) and not hasattr(other, '_sympy_'):
-                # XXX iterable self should have it's own __eq__
-                # method if the path gives a false negative
-                # comparison
-                return False
-            try:
-                other = _sympify(other)
-            except (SympifyError, SyntaxError):
-                return NotImplemented
+            return self._do_eq_sympify(other)
 
         # check for pure number expr
         if  not (self.is_Number and other.is_Number) and (
--- a/sympy/core/sympify.py
+++ b/sympy/core/sympify.py
@@ -28,9 +28,13 @@ class SympifyError(ValueError):
             str(self.base_exc)))
 
 
-# See sympify docstring.
 converter = {}  # type: tDict[Type[Any], Callable[[Any], Basic]]
 
+#holds the conversions defined in SymPy itself, i.e. non-user defined conversions
+_sympy_converter = {} # type: tDict[Type[Any], Callable[[Any], Basic]]
+
+#alias for clearer use in the library
+_external_converter = converter
 
 class CantSympify:
     """
@@ -77,7 +81,7 @@ def _convert_numpy_types(a, **sympify_args):
     import numpy as np
     if not isinstance(a, np.floating):
         if np.iscomplex(a):
-            return converter[complex](a.item())
+            return _sympy_converter[complex](a.item())
         else:
             return sympify(a.item(), **sympify_args)
     else:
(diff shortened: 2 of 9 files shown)


oscarbenjamin:
> or if its type has been added `converter` by the user.

I'm not sure I understand the need for this distinction. Is it so that downstream Basic subclasses can continue to depend on `__eq__` using sympify?

I wonder how much downstream code actually alters the behaviour of sympify at all. Would there ever be a need to do that?

ThePauliPrinciple:
I'm a bit surpised that this would cause a slowdown

oscarbenjamin:
> I'm a bit surpised that this would cause a slowdown

I think that the timings for the benchmarks are variable in CI so sometimes a particular benchmark will show up as being slower by chance.

asmeurer:
Keeping `converter` as a user-definable mechanism is important, but I don't know if it's necessary to continue to keep certain built-in classes in it. This is an implementation detail, IMO. So I'd say it's fine to just make a separate `sympy_converter` dict and replace instances of `converter` in library code with it. That removes the need for the `_setup_sympy_converter` function. 

Also I would make it `_sympy_converter` with the underscore to make it private.

ThePauliPrinciple:
Is there a way to avoid duplicating the code in `sympify` which checks the `converter` `dict`? Ideally the two `dict`s can be joined somewhere.

asmeurer:
You can merge two dictionaries with `{**d1, **d2}`. This will prefer the keys in `d2` if there are duplicates.

ThePauliPrinciple:
Sorry, I meant to have some way to do this only once, rather than every time `sympify` is used.

oscarbenjamin:
> Sorry, I meant to have some way to do this only once, rather than every time `sympify` is used.

There is chainmap but I would probably just duplicate the loop here:
https://github.com/sympy/sympy/blob/1409fcf748d8d73014cca509f2816acb82af7fc0/sympy/core/sympify.py#L368-L372
I don't think there is anywhere else that needs this...

ThePauliPrinciple:
I was wondering about this, before checking the mro, it also tries to see if the class exists directly in the dict, but I think the class is the first entry in the mro, so I should probably remove that code entirely. There is also an if statement related to "old style" classes, which I think can be removed too

ThePauliPrinciple:
I am happy with this now. In particular, instead of calling `_sympify`, I now directly perform a suitable conversion rather than having `_sympify` check for all posibilities again.

sympy/core/sympify.py
```diff
@@ -359,15 +363,18 @@ def sympify(a, locals=None, convert_xor=True, strict=False, rational=False,
...
-    if cls is None:
-        cls = type(a)  # Probably an old-style class
-    conv = converter.get(cls, None)
-    if conv is not None:
-        return conv(a)
 
+    #check for user defined converters
+    for superclass in getmro(cls):
+        try:
+            return _external_converter[superclass](a)
```
oscarbenjamin, line 370: It's much more likely that we will find what we are looking for in `_sympy_converter` so I would swap the order of these two loops.

ThePauliPrinciple: I was thinking that if the user wants to override SymPy's conversion that should be possible (it is currently).

sympy/core/basic.py
```diff
@@ -319,6 +319,16 @@ def inner_key(arg):
         args = len(args), tuple([inner_key(arg) for arg in args])
         return self.class_key(), args, S.One.sort_key(), S.One
 
+    def _do_eq_sympify(self, other):
+        for superclass in type(other).__mro__:
+            try:
+                return self == _external_converter[superclass](other)
+            except KeyError:
+                continue
```
oscarbenjamin, line 327: This `try/except` pattern (which already exists in the codebase) should not be used as it can also catch unintended `KeyError` e.g. from calling the converter function or from the equality comparison `==` itself. It should be structured more like:
```python
for superclass in type(other).__mro__:
    try:
        return conv = _external_converter[superclass]
    except KeyError:
        continue
    else:
        return self == conv(other)
```
Personally I would just avoid the `try/except` altogether and do:
```python
for superclass in type(other).__mro__:
    conv = _external_converter.get(superclass)
    if conv is not None:
        return self == conv(other)
```

sympy/core/sympify.py
ThePauliPrinciple: I'm considering to put the `_sympy_converter` and `_external_converter` inside the same loop (but still `_external_converter` first)

oscarbenjamin: > I'm considering to put the `_sympy_converter` and `_external_converter` inside the same loop (but still `_external_converter` first)

Currently whichever entry from the mro first matches will be chosen so I guess checking all of the external ones before any of the internal ones could lead to a different result e.g. if `_sympy_converter` matches lower in the mro than `_external_converter`. Maybe that's impossible though...

oscarbenjamin: Yes, I think so.

oscarbenjamin: We should probably have some tests that none of this has potentially altered the behaviour.

oscarbenjamin:
I think it would be good to add some comments somewhere indicating why the code is the way that it is now.

sympy/core/sympify.py
ThePauliPrinciple: I'll open a different PR for that, since we need to make sure that those tests pass in the current master too.

ThePauliPrinciple:
> I think it would be good to add some comments somewhere indicating why the code is the way that it is now.

I added some comments and a docstring. Is it clearer now?

sympy/core/sympify.py
```diff
@@ -28,9 +28,13 @@ def __str__(self):
...
 
 
-# See sympify docstring.
 converter = {}  # type: tDict[Type[Any], Callable[[Any], Basic]]
 
+#holds the conversions defined in SymPy itself, i.e. non-user defined conversions
+_sympy_converter = {} # type: tDict[Type[Any], Callable[[Any], Basic]]
+
+#alias for clearer use in the library
+_external_converter = converter
```
oscarbenjamin, line 37: It would be good to add some explanation here for why there are two converters and which should be used within the codebase.

oscarbenjamin:
I think this looks good. Can you squash the commits and add tests?

ThePauliPrinciple:
Not sure when I can do this, I'm having trouble with a recent update from github no longer allowing me to push/pull with my current workflow.

oscarbenjamin:
> I'm having trouble with a recent update from github no longer allowing me to push/pull with my current workflow.

What exactly has changed? I can still push and pull? I use ssh keys for access.

ThePauliPrinciple:
Looks like it was only temporarily disabled yesterday (but will be a permanent problem in march).

ThePauliPrinciple:
> Can you squash the commits

I'm having trouble squashing due to the merges with master, are there any tools available that will ignore commits from merges with master when trying to squash commits?

oscarbenjamin:
I just tested these steps with this branch as follows:
1. `git rebase -i master`
2. In the editor (that opens automatically after 1.) change all lines below the top line to read "squash".
3. Close the editor and git rebases
4. Finally an editor reopens and you can edit the commit message for the squashed result.

oscarbenjamin:
I suggest checking out a temporary branch to try this first:
```
git checkout -b tmp
git rebase -i master
```

ThePauliPrinciple:
> I suggest checking out a temporary branch to try this first:
> 
> ```
> git checkout -b tmp
> git rebase -i master
> ```

Thanks!

oscarbenjamin:
Looks good. The tests were added in #22844

PR #22789, opened 2022-01-03
Problem with fraction

### LINKED ISSUE #22788: RecursionError for unevluated expression in latex
`latex(Mul(Pow(Integer(2), Integer(-1), evaluate=False), Add(Integer(1), i, evaluate=False), Add(Integer(1), Pow(i, Integer(-8), evaluate=False), evaluate=False), evaluate=False))`

ghost:
Fixed the problem with fraction

Fixes #22788(WIP)

Previously when we call `convert` with `expr` we get right back here again due to which `RecursionError` occurred now it's fixed.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/simplify/radsimp.py
+++ b/sympy/simplify/radsimp.py
@@ -1089,11 +1089,11 @@ def fraction(expr, exact=False):
                     denom.append(b)
                 elif exact:
                     if ex.is_constant():
-                        denom.append(Pow(b, -ex, evaluate=not exact))
+                        denom.append(Pow(b, -ex, evaluate=False))
                     else:
                         numer.append(term)
                 else:
-                    denom.append(Pow(b, -ex, evaluate=not exact))
+                    denom.append(Pow(b, -ex, evaluate=True))
             elif ex.is_positive:
                 numer.append(term)
             elif not exact and ex.is_Mul:
--- a/sympy/simplify/radsimp.py
+++ b/sympy/simplify/radsimp.py
@@ -1089,11 +1089,11 @@ def fraction(expr, exact=False):
                     denom.append(b)
                 elif exact:
                     if ex.is_constant():
-                        denom.append(Pow(b, -ex))
+                        denom.append(Pow(b, -ex, evaluate=not exact))
                     else:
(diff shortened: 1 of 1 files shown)


sympy/simplify/radsimp.py
```diff
@@ -1089,11 +1089,11 @@ def fraction(expr, exact=False):
                     denom.append(b)
                 elif exact:
                     if ex.is_constant():
-                        denom.append(Pow(b, -ex))
+                        denom.append(Pow(b, -ex, evaluate=not exact))
```
oscargus, line 1092: Here exact is True so evaluate can be False.

sympy/simplify/radsimp.py
```diff
@@ -1089,11 +1089,11 @@ def fraction(expr, exact=False):
...
                     else:
                         numer.append(term)
                 else:
-                    denom.append(Pow(b, -ex))
+                    denom.append(Pow(b, -ex, evaluate=not exact))
```
oscargus, line 1096: And here evaluate is True.

oscargus:
Please add a test as well.

sympy/simplify/radsimp.py
KuldeepBorkar: @oscargus thanks

KuldeepBorkar:
> Please add a test as well.

There are still some changes need to be made in sympy\printing\latex.py therefore I am waiting.

anutosh491:
Please change OP to either "Fixes" or "Closes" followed by the pr number . what do you mean by "Helps" , is it solving the issue partially ?

KuldeepBorkar:
@anutosh491 Yup issue is not fixed yet therefore it's still in progress and not updated.

anutosh491:
> @anutosh491 Yup issue is not fixed yet therefore it's still in progress and not updated.

Then maybe you could write WIP - "work in progress" in braces in your PR title as is usually done and change Helps to Fixes above otherwise your PR won't be linked to the issue ! And thinking that the issue hasn't been addressed someone else might also create one for the same and there would be overlap which should be avoided.

PR #22776, opened 2022-01-01
New method to calculate nearest distance between two lines  

AdvaitPote:
#### References to other Issues or PRs

Fixes #22774 

#### Brief description of what is fixed or changed
A new method has been added which calculates the nearest distance between two lines in 3D using the diff suggested by @smichr. Minor changes like adding a missing '>' have been made. 

#### Release Notes

* geometry
  * added a method to calculate the nearest distance between two lines

PR #22790, opened 2022-01-04
Added a class method 'nearest_points()' to calculate distance between two lines in 3D.

### LINKED ISSUE #22774: have a method to calculate the distance between two lines
It seems like the geometry module should have a method for computing the distance bewteen two lines. This is discussed [here](https://math.stackexchange.com/questions/1993953/closest-points-between-two-lines). The following is a high-level (slow?) method of doing so:
```python
def nearest_points(l1, l2):
    """return the closest distance of separation between l1 and l2
    and the points connecting the lines with that distance of
    separation.

    Examples
    ========

    >> from sympy.geometry.line import nearest_points
    >>> from sympy import Line, Point
    >>> f = nearest_points
    >>> l3 = Line(Point(0, 0, 0), Point(-1, 1, 0))

    There is no separation between collinear lines:

    >>> l3b = Line(Point(0, 0, 0), Point(1, -1, 0))
    >>> f(l3, l3b)
    (0, Point3D(-t, t, 0))

    There is a uniform distance between non-intersecting
    parallel lines:

    >>> lp = Line(Point(1, 0, 0), Point(0, 1, 0))
    >>> f(l3, lp)
    (sqrt(2)/2, [Point3D(-t, t, 0), Point3D(1/2 - t, t + 1/2, 0)])
    >>> l4 = Line(Point(1, 0, 1), Point(1, 1, 1))
    >>> f(l3, l4)
    (1, [Point3D(1, -1, 0), Point3D(1, -1, 1)])
    """
    from sympy.core.symbol import uniquely_named_symbol
    from sympy.solvers.solvers import solve
    u = uniquely_named_symbol('t', (l1, l2))
    same = l1.equals(l2)
    if same:
        return S.Zero, l1.arbitrary_point(u)
    assert same == False  # if we aren't sure, don't continue
    p1 = l1.arbitrary_point(u)
    if l1.is_parallel(l2):
        d = l2.distance(p1)
        p2 = l1.perpendicular_line(p1).intersection(l2)[0]
        return d, [p1, p2]
    p2 = l2.arbitrary_point(u)
    pts = []
    for i in range(2):
        d = l2.distance(p1)
        v = solve(d.diff(u), u)[0]
        pts.append(p1.subs(u, v))
        l2, l1 = l1, l2
        p1, p2 = p2, p1
    return d.subs(u, v), pts
```
Special cases for segments and rays would be needed. This routine does not consider those entities.

faze-geek:

Fixes #22774
Closes #22776 

Added a new method which calculates the smallest between 2 lines in 3-D space as will be useful in the geometry module .
Modified `sympy/geometry/line.py`
Added Tests `sympy/geometry/tests/test_line.py`

#### Release Notes

NO ENTRY


faze-geek:
Ping @smichr for review .
All the changes as suggested by you along with appropriate tests have been added in this pull request .All the checks have passed except for `test / sphinx` which I guess is due to a missed line in doc-string which I will look into .

smichr:
Should the separation be excluded from the return values? Should there be a separate function for the separation? Does an algebraic solution perform better than the high-level solution of this PR?
```python
def separation(l1, l2):
    """return separation of two skew lines

    Examples
    ========

    >>> from sympy import Line
    >>> a, b, c, d = (0, 0, 0), (1, 2, 3), (1, 1, 1), (2, 3, 5)
    >>> separation(Line(a, b), Line(c, d))
    sqrt(5)/5
    >>> a, b, c, d = (-1, 1, 4), (1, 1, -1), (5, -3, -3), (-2, 0, 1)
    >>> separation(Line(a, b), Line(c, d))
    sqrt(110)/55
    """
    from sympy import Matrix
    n = Matrix(l1.direction.unit).cross(Matrix(l2.direction.unit))
    ps = l2.p1 - l1.p1
    return abs(n.dot(ps))/Point(0,0,0).distance(n)


def close_points(l1, l2):
    """return s and t, the parametric location of the closest
    points on l1 and l2, the actual points, and the separation between
    those two points.

    NOTE: these were checked on wolframalpha.com with query, e.g.
    "closest points on Line((0,0,0),(1,1,1)) and Line((0,1,0),(1,2,3))"

    Examples
    ========

    >>> from sympy import Line
    >>> close_points(Line((0, 0, 0), (1, 1, 1)), Line((0, 1, 0), (1, 2, 3)))
    (3/4, 1/4, Point3D(3/4, 3/4, 3/4), Point3D(1/4, 5/4, 3/4), sqrt(2)/2)
    >>> close_points(Line((0, 0, 0), (1, 2, 3)), Line((1, 1, 1), (2, 3, 5)))
    (7/5, 4/5, Point3D(7/5, 14/5, 21/5), Point3D(9/5, 13/5, 21/5), sqrt(5)/5)
    """
    # when s and t are in [0,1] the solution is also valid for segments, otherwise
    # one must determine which end is closer to ps and/or pt and use that instead of ps or pt
    # respectively
    # L1 = lambda s: l1.p1+s*l1.direction
    # L2 = lambda t: l2.p1+t*l2.direction
    # eq = lambda l1: Matrix(L1(var('s')) - L2(var('t'))).dot(l1.direction)
    # stdict = solve((
    #     eq(Line(var('x1 y1 z1'),var('x2 y2 z2'))),
    #     eq(Line(var('x3 y3 z3'),var('x4 y4 z4')))),
    #     var('s t'), dict=True)
    x1,y1,z1=p1=l1.p1
    x2,y2,z2=p2=l1.p2
    x3,y3,z3=p3=l2.p1
    x4,y4,z4=p4=l2.p2
    x21, y21, z21 = l1.direction
    x43, y43, z43 = l2.direction
    x31, y31, z31 = l2.p1 - l1.p1
    R1 = x21**2 + y21**2 + z21**2
    R2 = x43**2 + y43**2 + z43**2
    d4321 = x21*x43 + y21*y43 + z21*z43
    d3121 = x21*x31 + y21*y31 + z21*z31
    d4331 = x31*x43 + y31*y43 + z31*z43
    den = d4321**2 - R1*R2
    # R1*s - d4321*t - d3121 = 0
    # d4321*s - R2*t - d4331 = 0
    s = (d4321*d4331 - R2*d3121)/den
    t = (R1*d4331 - d4321*d3121)/den
    ps = l1.p1 + s*l1.direction
    pt = l2.p1 + t*l2.direction
    return s, t, ps, pt, ps.distance(pt)
```

faze-geek:
> Should the separation be excluded from the return values? Should there be a separate function for the separation? 

No in my opinion seperation should not be excluded from the return values . Firslty because the documentation string  itself makes it clear enough and I don't think it can lead to any kind of misinterpretations . Secondly even when we manually evaluate shortest distance between two skew lines using parametric points on the lines is the approach (most common I might add). The return values are currently the seperation distance and the points on the shortest vector between the two lines , which I believe is more interactive with the user .

anutosh491:
Yeah true , I have been looking into issues related to the geometry module lately and this Pr seems fine rather than a core algebraic solution ! what advantages do you see over the implemented pr ? @smichr

faze-geek:
Ping @oscarbenjamin I believe I have done the required changes .Could you add labels ,review it and merge if there are no potential changes to be made !?

smichr:
> this Pr seems fine rather than a core algebraic solution

```python
>>> l1,l2=Line((0, 0, 0), (1, 2, 3)), Line((1, 1, 1), (2, 3, 5))
>>> t=time();ok=close_points(l1,l2);time()-t
0.005997657775878906
>>> t=time();ok=l2.nearest_points(l1);time()-t
2.938823938369751
```
A low-level solution is about 500X faster.

faze-geek:
> A low-level solution is about 500X faster.

Thanks @smichr .I think we should then go ahead with faster algebraic solution .Also I still think that separation must be returned with the nearest point values .I'll replace the function in a commit soon .

sympy/geometry/line.py
```diff
@@ -1280,6 +1280,57 @@ def distance(self, other):
...
+        >>> f(l3, l4)
+        (1, [Point3D(1, -1, 0), Point3D(1, -1, 1)])
+        """
+        from sympy.core.symbol import uniquely_named_symbol
+        from sympy.solvers.solvers import solve
+        u = uniquely_named_symbol('t', (l1, l2))
+        same = l1.equals(l2)
+        if same:
+            return S.Zero, l1.arbitrary_point(u)
+        assert same == False  # if we aren't sure, don't continue
```
czgdp1807, line 1318:
```suggestion
        assert same is False  # if we aren't sure, don't continue
```

sympy/geometry/line.py
```diff
@@ -1280,6 +1280,57 @@ def distance(self, other):
...
+        >>> f(l3, lp)
+        (sqrt(2)/2, [Point3D(-t, t, 0), Point3D(1/2 - t, t + 1/2, 0)])
+        >>> l4 = Line(Point(1, 0, 1), Point(1, 1, 1))
+        >>> f(l3, l4)
+        (1, [Point3D(1, -1, 0), Point3D(1, -1, 1)])
+        """
+        from sympy.core.symbol import uniquely_named_symbol
+        from sympy.solvers.solvers import solve
+        u = uniquely_named_symbol('t', (l1, l2))
+        same = l1.equals(l2)
```
czgdp1807, line 1315:
```suggestion
        same = l1.equals(l2)

```

sympy/geometry/line.py
```diff
@@ -1280,6 +1280,57 @@ def distance(self, other):
...
+        if same:
+            return S.Zero, l1.arbitrary_point(u)
```
czgdp1807, line 1317:
```suggestion
            return S.Zero, l1.arbitrary_point(u)

```

sympy/geometry/line.py
czgdp1807: May be a `ValueError` with a nice message would be good for the end user.

sympy/geometry/line.py
```diff
@@ -1280,6 +1280,57 @@ def distance(self, other):
...
+        assert same == False  # if we aren't sure, don't continue
+        p1 = l1.arbitrary_point(u)
```
czgdp1807, line 1319:
```suggestion
        p1 = l1.arbitrary_point(u)

```

sympy/geometry/line.py
```diff
@@ -1280,6 +1280,57 @@ def distance(self, other):
...
+        if l1.is_parallel(l2):
+            d = l2.distance(p1)
+            p2 = l1.perpendicular_line(p1).intersection(l2)[0]
+            return d, [p1, p2]
```
czgdp1807, line 1323:
```suggestion
            return d, [p1, p2]

```

sympy/geometry/tests/test_line.py
```diff
@@ -814,3 +814,20 @@ def test_issue_8615():
     a = Line3D(Point3D(6, 5, 0), Point3D(6, -6, 0))
     b = Line3D(Point3D(6, -1, 19/10), Point3D(6, -1, 0))
     assert a.intersection(b) == [Point3D(6, -1, 0)]
+
+
+def test_nearest_points():
+    t = Symbol('t')
+    f = Line.nearest_points
```
czgdp1807, line 821:
```suggestion
    f = Line.nearest_points

```

sympy/geometry/tests/test_line.py
```diff
@@ -814,3 +814,20 @@ def test_issue_8615():
...
+    #Collinear case
+    l1 = Line(Point(0, 0, 0), Point(-1, 1, -1))
+    l1c = Line(Point(0, 0, 0), Point(1, -1, 1))
+    assert f(l1, l1c) == (0, Point3D(-t, t, -t))
```
czgdp1807, line 825:
```suggestion
    assert f(l1, l1c) == (0, Point3D(-t, t, -t))

```

sympy/geometry/tests/test_line.py
```diff
@@ -814,3 +814,20 @@ def test_issue_8615():
...
+    #Parallel 3-D case
+    l2 = Line(Point(0, 0, -1), Point(-1, 1, -1))
+    l2p = Line(Point(1, 0, -1), Point(0, 1, -1))
+    assert f(l2, l2p) == (sqrt(2)/2, [Point3D(-t, t, -1), Point3D(1/2 - t, t + 1/2, -1)])
```
czgdp1807, line 829:
```suggestion
    assert f(l2, l2p) == (sqrt(2)/2, [Point3D(-t, t, -1), Point3D(1/2 - t, t + 1/2, -1)])

```

sympy/geometry/line.py
```diff
@@ -1280,6 +1280,57 @@ def distance(self, other):
...
+        p2 = l2.arbitrary_point(u)
+        pts = []
+        for i in range(2):
+            d = l2.distance(p1)
+            v = solve(d.diff(u), u)[0]
+            pts.append(p1.subs(u, v))
+            l2, l1 = l1, l2
+            p1, p2 = p2, p1
+        return d.subs(u, v), pts
```
czgdp1807, line 1332: Is it similar to https://brilliant.org/wiki/3d-coordinate-geometry-distance/#distance-between-2-skew-lines or https://math.stackexchange.com/a/2217845? Distance will only matter if the lines are non-parallel, the formula in the above links should work in that case. Not sure though.

sympy/geometry/line.py
```diff
@@ -1280,6 +1280,57 @@ def distance(self, other):
             return S.Zero
         return self.perpendicular_segment(other).length
 
+    def nearest_points(l1, l2):
```
czgdp1807, line 1283: `shortest_distance` or that matter, just `distance`?

smichr: It's returning both the shortest distance and the points at the ends of that short segment. Perhaps we could return the segment and let the user take what they want, distance or args/pts. Then the function could be `shortest_segment_between`

sympy/geometry/line.py
faze-geek: Yes @smichr `close_points ` answers are not consistent throughout .I've been a bit dicey after some of it's answers wouldn't match with `nearest_points` earlier .
This is one I remember right now but I have encountered more ( this is a parallel case ,I believe it will be also incorrect in collinear cases)
```
>>> l1=Line(Point(0,0,0), Point(-1,1,0))
>>> l2= Line(Point(1, 0, 0), Point(0, 1, 0))
>>> separation(l1,l2)
nan

>>> close_points(l1,l2)
  File "C:\Users\kunni\sympy\sympy\sympy\geometry\point.py", line 218, in __add__
    s, o = Point._normalize_dimension(self, Point(other, evaluate=False))
  File "C:\Users\kunni\sympy\sympy\sympy\geometry\point.py", line 125, in __new__
    raise TypeError(filldedent('''
TypeError:
Expecting sequence of coordinates, not `NaN`
```

PR #22796, opened 2022-01-05
Matplotlib has improved

oscargus:

It can render both `\overline` and `\operatorname` now.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/interactive/printing.py
+++ b/sympy/interactive/printing.py
@@ -117,10 +117,6 @@ def _init_ipython_printing(ip, stringify_func, use_latex, euler, forecolor,
         return exprbuffer.getvalue().decode('utf-8')
 
     def _matplotlib_wrapper(o):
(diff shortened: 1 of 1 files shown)


ThePauliPrinciple:
Can a test be added? In case matplotlib's support for this is not stable. Considering this is something users can now use, it might be helpful if there is a release note.

oscargus:
I do not see a way to add a test. Earlier, these two where internally replaced before sending to IPython (which in turn calls Matplotlib), so from a user perspective the only difference is that it looks more as it should.

I'm quite sure that the support is stable. 
```
2011-10-25

    added support for operatorname to mathtext, including the ability to insert spaces, such as $operatorname{arg,max}$ - PI
```

`\overline` has been supported at least since 1.3.1 which is 2013 or so. (Couldn't find an exact entry.)

PR #22797, opened 2022-01-05
Improve bernoulli DE matching by changing general equation and solution

### LINKED ISSUE #22462: coefficients in Bernoulli DE
Hello, everyone! 

I've ran into problems using Bernoulli equation with certain coefficients. I wanted to solve this equation with `dsolve(de, f(x))`
```
de = Eq(f(x).diff(x), 20*f(x)**2 - 500*f(x)/7200)
```
but got stuck on a long executing time. I checked the reduced coefficients
```
de = Eq(f(x).diff(x), 2*f(x)**2 - 5*f(x)/7)
```
and got this answer:
```
Eq(x - 7*log(f(x) - 5/14)/5 + 7*log(f(x))/5, C1)
```
By using `classify_ode` I found out that equations aren't completely classified. They aren't classified as 'Bernoulli' so I can't use a hint to point the type manually. I checked all available coefficients from -2 to 2 and it turns out that equations based on some of them aren't classified as Bernoulli equations. 
```
from sympy import *
f = Function("f")
x = symbols("x")
for i in range(-2,3,1):
    for j in range(-2,3,1):
        if i != 0 and j != 0:
            de = Eq(f(x).diff(x), i*f(x) + j*f(x)**2)
            if classify_ode(de).count('Bernoulli') == 0:
                print(de) 
# gives
#Eq(Derivative(f(x), x), -2*f(x)**2 - 2*f(x))
#Eq(Derivative(f(x), x), -2*f(x)**2 - f(x))
#Eq(Derivative(f(x), x), -2*f(x)**2 + f(x))
#Eq(Derivative(f(x), x), -2*f(x)**2 + 2*f(x))
```
Using `dsolve` leads to the correct answers for this four equations. However changing `-2*f(x)**2` to `-0.5*f(x)**2` leads to incomplete answers.

I am running SymPy 1.9. I checked live version with SymPy 1.5.1 and have similar results, except that the type is classified as Bernoulli and I can manually specify it with `hint='Bernoulli'` and get the right answers.

Any suggestions to solve the initial equation with SymPy 1.9?

 Thanks in advance

mostlyaman:
Fixes #22462 

The general equation of a Bernoulli DE is stored as `f(x).diff(x) + P*f(x) - Q*f(x)**n`. This format fails to match some equations which can be solved as a Bernoulli DE. Some of these cases are described in this comment. https://github.com/sympy/sympy/issues/22462#issuecomment-966572294

More specifically, all equations where `Q >= 2` fail to match with the general Bernoulli DE as explained in this comment.
https://github.com/sympy/sympy/issues/22462#issuecomment-986268603

The actual problem occurs in this bit of code.
https://github.com/sympy/sympy/blob/1409fcf748d8d73014cca509f2816acb82af7fc0/sympy/core/operations.py#L255-L258

Here, 'expr ' is our part of our original equation in the form `- Q*f(x)**n`. To match the wild Q, there is a negation added to matched part of Q to counteract the negative sign in front of Q in the general equation. This increases the no of operations and the code returns None and fails to match the code.

A more detailed explanation is in this comment. https://github.com/sympy/sympy/issues/22462#issuecomment-987622991

The solution proposed is to change the general Bernoulli DE from `f(x).diff(x) + P*f(x) - Q*f(x)**n` to `f(x).diff(x) + P*f(x) + Q*f(x)**n`. This eliminates the need to negate the matched part of Q and hence it is now able to match all the equations which were earlier posing a problem.

Since the inbuilt `_equation` method of Bernoulli class is for internal use, I doubt it would break compatibility for users.

#### Other comments
A previous unit test has been modified as it was earlier not being recognised as a Bernoulli DE and now is.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/solvers/ode/single.py
+++ b/sympy/solvers/ode/single.py
@@ -830,7 +830,7 @@ class Bernoulli(SinglePatternODESolver):
 
     def _equation(self, fx, x, order):
         P, Q, n = self.wilds()
-        return fx.diff(x) + P*fx - Q*fx**n
+        return fx.diff(x) + P*fx + Q*fx**n
 
     def _get_general_solution(self, *, simplify_flag: bool = True):
         P, Q, n = self.wilds_match()
@@ -839,11 +839,11 @@ class Bernoulli(SinglePatternODESolver):
         (C1,) = self.ode_problem.get_numbered_constants(num=1)
         if n==1:
             gensol = Eq(log(fx), (
-            C1 + Integral((-P + Q), x)
+            C1 + Integral((-P - Q), x)
         ))
         else:
             gensol = Eq(fx**(1-n), (
-                (C1 - (n - 1) * Integral(Q*exp(-n*Integral(P, x))
+                (C1 - (n - 1) * Integral(-Q*exp(-n*Integral(P, x))
                             * exp(Integral(P, x)), x)
                 ) * exp(-(1 - n)*Integral(P, x)))
             )
--- a/sympy/solvers/ode/tests/test_ode.py
+++ b/sympy/solvers/ode/tests/test_ode.py
@@ -1059,3 +1059,15 @@ def test_issue_22604():
     assert x1sol == Eq(x1(t), sqrt(3 - sqrt(5))*(sqrt(10) + 5*sqrt(2))*cos(sqrt(2)*t*sqrt(3 - sqrt(5))/2)/20 + \
                        (-5*sqrt(2) + sqrt(10))*sqrt(sqrt(5) + 3)*cos(sqrt(2)*t*sqrt(sqrt(5) + 3)/2)/20)
     assert x2sol == Eq(x2(t), (sqrt(5) + 5)*cos(sqrt(2)*t*sqrt(3 - sqrt(5))/2)/10 + (5 - sqrt(5))*cos(sqrt(2)*t*sqrt(sqrt(5) + 3)/2)/10)
+
+def test_issue_22462():
+    de = Eq(f(x).diff(x), -20*f(x)**2 - 500*f(x)/7200)
+    assert classify_ode(de, f(x)) == ('factorable', 'separable', '1st_exact',
+        'Bernoulli', '1st_rational_riccati', '1st_power_series', 'lie_group',
+        'separable_Integral', '1st_exact_Integral', 'Bernoulli_Integral')
+    assert dsolve(de, hint = "Bernoulli") == Eq(f(x), 1/(symbols('C1')*exp(5*x/72) - 288))
+
+    de2 = Eq(f(x).diff(x), -2*f(x)**2 - 5*f(x)/7)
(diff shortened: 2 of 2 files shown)


sympy/solvers/ode/single.py
```diff
@@ -839,11 +839,11 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
         (C1,) = self.ode_problem.get_numbered_constants(num=1)
         if n==1:
             gensol = Eq(log(fx), (
-            C1 + Integral((-P + Q), x)
+            C1 + Integral((-P - Q), x)
```
ThePauliPrinciple, line 842: Can this be simplified to `C1 - Integral(P+Q, x)`, or is that not possible here/ going against convention for this particular problem?

sympy/solvers/ode/single.py
```diff
@@ -839,11 +839,11 @@ def _get_general_solution(self, *, simplify_flag: bool = True):
...
         ))
         else:
             gensol = Eq(fx**(1-n), (
-                (C1 - (n - 1) * Integral(Q*exp(-n*Integral(P, x))
+                (C1 - (n - 1) * Integral(-Q*exp(-n*Integral(P, x))
```
ThePauliPrinciple, line 846: likewise here `(C1 + (n - 1) * Integral(Q*exp(-n*Integral(P, x))`

sympy/solvers/ode/single.py
mostlyaman: Yeah I think it is a good suggestion, should work. I'll test it.

mostlyaman: https://github.com/sympy/sympy/blob/1409fcf748d8d73014cca509f2816acb82af7fc0/sympy/solvers/ode/tests/test_ode.py#L63
Apart from this failing, all other tests seem fine. 

I am not sure if it will not cause a problem though. Because we have just changed the sign of Q from general equation, it just counteracts that change, so we have the same expression as we have had from the all the previous sympy versions.

oscarbenjamin:
Maybe I don't understand but it looks to me like this is just working around the bug in `match()`. Is it not better to fix that?

mostlyaman:
I'm not sure if this is a bug in match. The code which cause the failure to match the expression is basically this.
https://github.com/sympy/sympy/blob/7a0fbd8168fd091abf14b65badf1c7c7a7a810f9/sympy/core/operations.py#L255-L258

We could just remove this if condition but there must be a reason why they were added in the first place. Though I'm not able to see why it is checking for the no of operations, I thought it is better to see why it was failing for Bernoulli matching specifically, and hence found out about the added negation caused by -ve sign infront of wild Q and that satisfies this if condition.

So, if you consider there is a bug in match, it is basically this if condition.
If removing this if condition is viable, then we could go that route.

smichr:
An alternative is given on the issue page.

PR #16272, opened 2019-03-16
Testing : Add 'coding : UTF-8' test in code quality test

sylee957:
#### References to other Issues or PRs

Fixes #16257

#### Brief description of what is fixed or changed

Current test suite does not properly detect import error for python 2, as the reason described in #16257

I have found the regular expression in
https://www.python.org/dev/peps/pep-0263/#defining-the-encoding
to match the encoding,

And used the trick from
https://www.quora.com/How-do-you-check-to-see-if-a-string-in-Python-is-in-ASCII
to detect unicode character in the string.

#### Other comments

This may be deleted after deprecating python 2.7

#### Release Notes

NO ENTRY

Issue #10974, opened 2016-04-07
solvers.py contains BOM character


asmeurer:
See https://stackoverflow.com/questions/7297888/ufeff-character-showing-up-in-files-how-to-remove-them for more information. The first character of solvers.py is U+FEFF. As far as I can tell it's impossible to tell this from GitHub, but if you `git blame sympy/solvers/solvers.py`, or use a hex editor (or even `less`), you can see it. The BOM was added by 56713720. 

We should find whatever other files have these and remove them. If a file has non-ASCII characters, it should start with `# -*- coding: utf-8 -*-`.

Nitin216:
@asmeurer I think it is only in ```solvers.py```. So should i add it manually.

PR #22798, opened 2022-01-05
Removed even more Python 2-support

oscargus:

There was still a bit of Python 2-related code left. Now, it is primarily the antlr-parsers for autolev and latex that should be updated.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/codegen/tests/test_pyutils.py
+++ b/sympy/codegen/tests/test_pyutils.py
@@ -5,5 +5,3 @@ def test_standard():
     ast = Print('x y'.split(), "coordinate: %12.5g %12.5g")
     assert render_as_module(ast, standard='python3') == \
         '\n\nprint("coordinate: %12.5g %12.5g" % (x, y))'
-    assert render_as_module(ast, standard='python2') == \
-        '\n\nprint "coordinate: %12.5g %12.5g" % (x, y)'
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -116,7 +116,7 @@ class BadArgumentsError(TypeError):
     pass
 
 
-# Python 2/3 version that does not raise a Deprecation warning
+# Python 3 version that does not raise a Deprecation warning
 def arity(cls):
     """Return the arity of the function if it is known, else None.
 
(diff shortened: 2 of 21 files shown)


oscarbenjamin:
> * BREAKING: `quality_unicode.py` removed as it was Python 2 only.

This isn't really a breaking change since this is just part of sympy's tests. I'm not sure why it should be removed though. The intention is to prevent contributors from adding unnecessary or accidental non-ascii text in code files.

oscargus:
> The intention is to prevent contributors from adding unnecessary or accidental non-ascii text in code files.

But that is only a Python 2 issue, right? Anyway, that was the reason that I removed it. If you check the comments, they all say "remove when Python 2 is not supported". 

Well, since it was deprecated when moved, it made change to note it here (otherwise I would not have done it).

oscarbenjamin:
> But that is only a Python 2 issue, right?

Currently unicode in the codebase will fail CI and that was done deliberately. There needs to be an agreement on whether unicode can be included in the sympy source. One place where some have wanted to use it is in docstrings. It would help the pretty printing in docstrings but it's not clear how many other things it would stop working e.g. because of editors/terminals/browsers that don't support enough of unicode etc.

oscarbenjamin:
The only difference with Python 2 is that the default encoding is not unicode if there is no encoding cookie at the top of the file.

asmeurer:
The Unicode code quality thing came from https://github.com/sympy/sympy/pull/16272. The original motivation was https://github.com/sympy/sympy/issues/10974. 

I think it is a least partially valuable. We should not allow non-printing Unicode characters, or characters that look exactly the same as ASCII equivalents. We should also not use non-ASCII variable names, at least without discussion. Incorrect encoding lines should be disallowed too, although I believe those will already raise an error on import. 

The coding header bits are indeed outdated. That was required for Unicode characters to work in Python 2 but isn't for Python 3 (UTF-8 is now the default encoding).

oscargus:
OK! So the message is to restore `quality_unicode`, but get rid of the header checking and the Python 2-comments?

oscarbenjamin:
The encoding cookies were just there for Python 2 compatibility so that part can definitely go. I think it's probably best to keep the quality check that prevents non-ASCII being added to non-whitelisted files.

oscargus:
This is updated.

oscarbenjamin:
Looks good!

PR #22799, opened 2022-01-05
22462: ignore sign while matching

### LINKED ISSUE #22462: coefficients in Bernoulli DE

smichr:

fixes #22462 
closes #22797 as alternate

There is some asymmetry in pattern matching when there is a leading sign on the expression as identified in #22462.
```python
>>> Q=Wild("Q")

These all work

>>> eq=5*x
>>> (Q*x).matches(eq)
{Q_: 5}
>>> (Q*x).matches(-eq)
{Q_: -5}
>>> (-Q*x).matches(-eq)
{Q_: 5}
>>> (-Q*x).matches(eq)
{Q_: -5}

But when matching two wilds, the sign is not handled as well:

>>> e=Wild('e')
>>> eq=5*x**2
>>> (Q*x**e).matches(eq)
{e_: 2, Q_: 5}
>>> (Q*x**e).matches(-eq)
{e_: 2, Q_: -5}
>>> (-Q*x**e).matches(-eq)
{e_: 2, Q_: 5}
>>> (-Q*x**e).matches(eq)  # now gives {e_: 2, Q_: -5} in stead of None
>>>
```

#### Other comments

This now allows Bernoulli ODEs to better be identified.

#### Release Notes

* core
    - a matching issue related to matching of a leading sign on products was fixed
* solvers
    - Bernoulli differential equations are more often classified as such in `classify_ode` because of a fix to matching

### DIFF
--- a/sympy/solvers/ode/tests/test_ode.py
+++ b/sympy/solvers/ode/tests/test_ode.py
@@ -740,14 +740,16 @@ def test_undetermined_coefficients_match():
     assert _undetermined_coefficients_match(cos(x**2), x) == {'test': False}
     assert _undetermined_coefficients_match(2**(x**2), x) == {'test': False}
 
-def test_issue_4785():
+
+def test_issue_4785_22462():
     from sympy.abc import A
     eq = x + A*(x + diff(f(x), x) + f(x)) + diff(f(x), x) + f(x) + 2
     assert classify_ode(eq, f(x)) == ('factorable', '1st_exact', '1st_linear',
-        'almost_linear', '1st_power_series', 'lie_group',
+        'Bernoulli', 'almost_linear', '1st_power_series', 'lie_group',
         'nth_linear_constant_coeff_undetermined_coefficients',
         'nth_linear_constant_coeff_variation_of_parameters',
-        '1st_exact_Integral', '1st_linear_Integral', 'almost_linear_Integral',
+        '1st_exact_Integral', '1st_linear_Integral', 'Bernoulli_Integral',
+        'almost_linear_Integral',
         'nth_linear_constant_coeff_variation_of_parameters_Integral')
     # issue 4864
     eq = (x**2 + f(x)**2)*f(x).diff(x) - 2*x*f(x)
@@ -759,6 +761,7 @@ def test_issue_4785():
         'lie_group', '1st_exact_Integral',
         '1st_homogeneous_coeff_subs_indep_div_dep_Integral',
         '1st_homogeneous_coeff_subs_dep_div_indep_Integral')
+
 
 def test_issue_4825():
     raises(ValueError, lambda: dsolve(f(x, y).diff(x) - y*f(x, y), f(x)))
@@ -1059,3 +1062,10 @@ def test_issue_22604():
     assert x1sol == Eq(x1(t), sqrt(3 - sqrt(5))*(sqrt(10) + 5*sqrt(2))*cos(sqrt(2)*t*sqrt(3 - sqrt(5))/2)/20 + \
                        (-5*sqrt(2) + sqrt(10))*sqrt(sqrt(5) + 3)*cos(sqrt(2)*t*sqrt(sqrt(5) + 3)/2)/20)
     assert x2sol == Eq(x2(t), (sqrt(5) + 5)*cos(sqrt(2)*t*sqrt(3 - sqrt(5))/2)/10 + (5 - sqrt(5))*cos(sqrt(2)*t*sqrt(sqrt(5) + 3)/2)/10)
+
+
+def test_issue_22462():
+    for de in [
+            Eq(f(x).diff(x), -20*f(x)**2 - 500*f(x)/7200),
+            Eq(f(x).diff(x), -2*f(x)**2 - 5*f(x)/7)]:
+        assert 'Bernoulli' in classify_ode(de, f(x))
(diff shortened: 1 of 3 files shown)


smichr:
@mostlyaman , please feel free to make PR requests against this if you want to see additional changes. It was different enough from what you proposed that I thought it would be easier to just open a new PR.

sympy/solvers/ode/tests/test_ode.py
```diff
@@ -740,14 +740,16 @@ def test_undetermined_coefficients_match():
     assert _undetermined_coefficients_match(cos(x**2), x) == {'test': False}
     assert _undetermined_coefficients_match(2**(x**2), x) == {'test': False}
 
-def test_issue_4785():
+
+def test_issue_4785_22462():
     from sympy.abc import A
     eq = x + A*(x + diff(f(x), x) + f(x)) + diff(f(x), x) + f(x) + 2
```
smichr, line 746: @mostlyaman , do you think it is a good idea to add the solution of this equation with the Bernoulli hint in case someone drops the `count_ops` approach in matching? -- didn't you get a wrong result in that case?

mostlyaman: Yes, the solution given by dsolve is incorrect but it not the fault of Bernoulli Solver. The solver produces the correct Unevaluated Integral. It is the integration procedure which adds a wrong result. I opened a issue to fix this case.

If this is not the answer to your question, then can you please explain a bit more. I dont get what `in case someone drops the count_ops approach in matching?` means?

smichr: I thought it gave the wrong result when you removed the `count_ops` test. I'll check the issue you opened.

mostlyaman: No, the wrong result is in no way related to the count_ops() test. I explained it here in this comment. https://github.com/sympy/sympy/issues/22462#issuecomment-991948813

```python
In [13]: eq                                                                                              
Out[13]: 
  ⎛           d       ⎞              d           
A⋅⎜x + f(x) + ──(f(x))⎟ + x + f(x) + ──(f(x)) + 2
  ⎝           dx      ⎠              dx          

In [14]: dsolve(eq,hint = "Bernoulli_Integral")                                                          
Out[14]: 
       ⎛     ⌠                         ⎞         
       ⎜     ⎮                ⌠        ⎟         
       ⎜     ⎮                ⎮ 1 dx   ⎟   ⌠     
       ⎜     ⎮                ⌡        ⎟  -⎮ 1 dx
       ⎜     ⎮ (A⋅x + x + 2)⋅ℯ         ⎟   ⌡     
f(x) = ⎜C₁ + ⎮ ───────────────────── dx⎟⋅ℯ       
       ⎜     ⎮         A + 1           ⎟         
       ⎝     ⌡                         ⎠         

In [15]: dsolve(eq,hint = "Bernoulli_Integral").doit()                                                   
Out[15]: 
       ⎛     ⎛⎧                   x            ⎞⎞    
       ⎜     ⎜⎪(A⋅x - A + x + 1)⋅ℯ             ⎟⎟    
       ⎜     ⎜⎪────────────────────  for A ≠ -1⎟⎟    
       ⎜     ⎜⎪       A + 1                    ⎟⎟  -x
f(x) = ⎜C₁ + ⎜⎨                                ⎟⎟⋅ℯ  
       ⎜     ⎜⎪      2                         ⎟⎟    
       ⎜     ⎜⎪     x     2⋅x                  ⎟⎟    
       ⎜     ⎜⎪     ── + ─────       otherwise ⎟⎟    
       ⎝     ⎝⎩     2    A + 1                 ⎠⎠
```
From the above code snippet, you can see that the unevaluated integral produced is correct. Upon calling doit, the wrong Piecewise is added. I dont believe the wrong result has something to do with count_ops!

PR #22801, opened 2022-01-05
docs: Support `make livehtml` in Docker

### LINKED ISSUE #22800: docs: Support `make livehtml` in the Docker image
> By the way, `make livehtml` will also open a live server, and in addition, it live rebuilds the docs whenever an update is made.

_Originally posted by @asmeurer in https://github.com/sympy/sympy/issues/22655#issuecomment-997471982_

skieffer:

Fixes #22800 

Provide a way to use the `livehtml` server, when using Docker to build the docs.

#### Release Notes

NO ENTRY


doc/src/guides/contributing/build-docs.rst
```diff
@@ -136,6 +136,67 @@ down menu and select “open file”, navigate into the ``sympy/doc/_build/html`
...
+
+   cd doc
+   make livehtml
+
+to start the server. In your terminal you should see a line that looks like::
+
+   [sphinx-autobuild] Serving on http://127.0.0.1:#####
+
+with some numbers giving the port at the end. Copy this address (including the
+port number) and navigate to this address in your web browser.
```
asmeurer, line 194: Actually I added `--open-browser` to the sphinx-autobuild command in the Makefile, because doing this manually is annoying, so it should open a browser automatically.

doc/Makefile
```diff
@@ -10,6 +10,8 @@ SPHINXBUILD  = sphinx-build
 PAPER        =
 BUILDDIR     = _build
 SOURCEDIR    = src
+LIVEHOST     = 127.0.0.1
```
asmeurer, line 13: I guess this has no effect for the local build, right? It still works for me.

doc/src/guides/contributing/build-docs.rst
skieffer: You're right, it does open automatically. I'll adapt the text.

doc/Makefile
skieffer: It works locally for me as well. But on second thought, probably better to write this as `localhost` instead of the numerical version, since that's probably more what people expect to see in their browser. I'll change it to that.

skieffer:
I guess there are no big questions on this one, and I'll merge it soon.

PR #22802, opened 2022-01-06
DO NOT MERGE: Changing Sphinx theme to Furo

bertiewooster:

Attempting to change Sphinx theme to [Furo](https://github.com/pradyunsg/furo) per issue #22716

#### Release Notes

NO ENTRY


asmeurer:
Is the SymPy Live thing just an accidental commit? I don't think we should add a submodule just for the docs build.

bertiewooster:
Yes, it's accidental, has to do with my .venv directory.

asmeurer:
In general I like this theme. One thing I like is that it makes different formatting very clear. For instance, I didn't notice that the text in reference/index.rst is currently formatted as block quotes because of the way it is indented, because the current theme does renders it in a way that makes it hard to see that. But with Furo, you can see right away that something is off with the spacing and there is a vertical line to the left of the text. Also as I mentioned on https://github.com/sympy/sympy/pull/22693, the different header levels are much easier to distinguish. One thing I'm not a fan of is the red text used in the autodoc function headers, but we will need to fix anyway. It also feels to me like Furo wastes a lot of horizontal space with unnecessary padding to the left and right. I'm not just talking about keeping the text within a certain width, which we also do, but specifically the padding to the left of the left sidebar. This could probably be fixed with the CSS, though. 

We are going to need to update the styling so that it looks unique. I would go for something similar to the existing docs in terms of colors, although i'm not opposed to modifying them a bit, especially if there are contrast issues (for context, I created the current docs "color theme" mostly based on the colors from the website and logo, and didn't pay particularly close attention to accessibility at the time). I guess we also will need to make both "light" and "dark" themes?

There's also the question of the Live extension, which is almost completely broken. I don't want to waste too much time on the Live shell, because it's something that I want to replace it (see https://github.com/sympy/sympy-live/issues/83), but we need to have something that isn't broken. I think this may end up being the biggest blocker of this, unfortunately.

bertiewooster:
> One thing I like is that it makes different formatting very clear. For instance, I didn't notice that the text in reference/index.rst is currently formatted as block quotes because of the way it is indented, because the current theme does renders it in a way that makes it hard to see that. But with Furo, you can see right away that something is off with the spacing and there is a vertical line to the left of the text.

Great point. For easy comparison, here's a screenshot of that page in each theme: current theme on top, Furo on bottom.

reference_index_classic_over_furo

Also worth checking is how a theme responds to being in a narrow-width window. For example, current theme (left) is unresponsive so sidebar takes up much of the window; Furo is responsive so by default it shows only the body and hides the sidebar (which can be shown via a hamburger menu).

reference_index_classic_over_furo

bertiewooster:
>I guess we also will need to make both "light" and "dark" themes?

Maybe? Furo handles dark mode itself fairly well, and (at least as currently configured) we don't have too many different colors so visually we're not far from default Furo. Here's the same page in light and dark mode. The green box around the diagram uses (I believe) a custom SymPy color, and it's rendered reasonably in dark mode.

furo_light_over_dark

bertiewooster:
> It also feels to me like Furo wastes a lot of horizontal space with unnecessary padding to the left and right. I'm not just talking about keeping the text within a certain width, which we also do, but specifically the padding to the left of the left sidebar. This could probably be fixed with the CSS, though.

I played with it and it appears Furo's philosophy is to keep the article (main) content width constant regardless of the window width (down to a certain window width, obviously). In the web page inspector, I added borders around the sidebar and article. The article width stays constant, and the padding to the left of the sidebar increases with the window width. If you shrink the window even more, the sidebar responsively disappears (really, gets shifted off-screen), and the article width stays constant.

I'd guess all this is to avoid the article text constantly re-flowing when you adjust the window width, which I do find distracting.

sidebar-article

oscargus:
This may be unrelated, but I sometimes read stuff on my (Android) tablet and one important thing for me is that one should be able to zoom, as in actually zooming in so that the left- and right-hand side of the "page" is outside of the screen, i.e., not reformatting with the same page width and a larger font size. Maybe this is what is mentioned above?

Anyway, if any of these themes differ in that respect, I would just like to encourage you to give that feature rather high importance...

(Never understood why some sites does one thing and others not though...)

doc/requirements.txt
```diff
@@ -1,3 +1,4 @@
+furo
```
oscargus, line 1: Maybe one should actually change the script to use `doc/requirements.txt`? But then, I believe that the `release/requirements.txt` may be used for the release, so probably both files should be in sync (i.e., only one file...).

moorepants:
I'd like to see a bit more "SymPy green" incorporated into the coloring. Does the theme allow for color pallette changes? If so, selecting a color pallette based on the SymPy green would be nice. Here is a website that generates color palletes showing some for the sympy color: https://mycolor.space/?hex=%233B5526&sub=1

asmeurer:
Right. It's actually a fairly minor point. My point is only about the space between the left side of the window and the left sidebar when the window is wide enough. It gives the impression of unused space. Compare that to, for instance, the way the [GitHub docs](https://docs.github.com/en/rest/overview/resources-in-the-rest-api), which use a very similar layout, handle responsiveness. Maybe it's something @pradyunsg would consider improving upstream. Or maybe my impression of how it would look without this isn't as good as it actually would look. 

> Maybe? Furo handles dark mode itself fairly well, and (at least as currently configured) we don't have too many different colors so visually we're not far from default Furo. Here's the same page in light and dark mode. The green box around the diagram uses (I believe) a custom SymPy color, and it's rendered reasonably in dark mode.

Presumably there are two sets of CSS stylings, one for light and one for dark. It's possible I am misunderstanding how the dark mode stuff is implemented (I haven't looked at the Furo CSS yet). So if we make the background some light shade of green, for instance, we will need some other, darker shade of green for dark mode. 

I'm not sure, but the code block styling, which is coming from the Live extension, might just be working that way because the extension changes the color for light mode only. If the extension is in-fact messing with the styling that's something we might want to change. 

By the way, another thing to check is long math formulae, like the ones on this page https://docs.sympy.org/dev/modules/integrals/g-functions.html. I think lines in the custom CSS are relevant (I believe they prevent the math from creating scrollbars): https://github.com/sympy/sympy/pull/22802/files#diff-89b7609a41acd7d2b8f63fff6a36fe070b7a940ab155a04ded46c3099259c4b4L348-L353

bertiewooster:
> I'd like to see a bit more "SymPy green" incorporated into the coloring. Does the theme allow for color pallette changes? If so, selecting a color pallette based on the SymPy green would be nice. Here is a website that generates color palletes showing some for the sympy color: https://mycolor.space/?hex=%233B5526&sub=1

Thanks for the color palettes! Concur it makes sense to customize the coloring. Yes, it does; there are currently a few touches of the existing color scheme, for example in the code blocks:

<details>
<summary>Furo mobile</summary>
<br>

furo_link_to_anchor

</details>

pradyunsg:
> Presumably there are two sets of CSS stylings, one for light and one for dark. It's possible I am misunderstanding how the dark mode stuff is implemented (I haven't looked at the Furo CSS yet). So if we make the background some light shade of green, for instance, we will need some other, darker shade of green for dark mode.

https://pradyunsg.me/furo/customisation/colors/#how-light-and-dark-mode-work

> [here's a screenshot of that page](https://github.com/sympy/sympy/pull/22802#issuecomment-1006238371)

I'm curious why there's missing spacing around the left and right of the content.

Screenshot 2022-01-07 at 11 45 54

This is especially pronounced on the mobile view for you.

> My point is only about the space between the left side of the window and the left sidebar when the window is wide enough.

Do you have a screen resolution for this?

If this is on a wide-ish monitor, it's because Furo keeps the content centered on the page -- which is a much better experience than having it hugging the left edge on even-wider screens. It's similar to how GitHub behaves when viewing a repository, for example.

If this is on a less-wide window (eg: less than iPads but more than mobile), part of why Furo does what it does is to prevent different widths in the content. It might make sense to change that stance and allow additional content widths -- but so far, I've not seen any common popular device that hits the weird gap between iPad and mobile resolutions.

pradyunsg:
> Does the theme allow for color pallette changes?

Yup! Pretty much every color on the screen is customizable. :)

See https://pradyunsg.me/furo/customisation/colors/#defining-overrides-for-defaults

bertiewooster:
@pradyunsg thanks for answering our questions.

> I'm curious why there's missing spacing around the left and right of the content.
> This is especially pronounced on the mobile view for you.

Turns out we apply a CSS file which sets the padding of .content to 0.1em (so essentially none). Disabling that rule restores the padding:

furo_live-core_css_disabled

Side note: The inspector indicates that CSS file is taken from another web site, https://live.sympy.org/static/live-core.css. Perhaps that was done to avoid duplication of CSS code across the two sites? To speed up loading, I believe the best practice is to have web pages minimize the number of files they load from other sites. So recommend we consider moving any needed CSS rules from that other site onto our main documentation site, https://docs.sympy.org. (The place I find `live-core` in the SymPy codebase is in `doc/src/sympylive.py`.)

bertiewooster:
It occurred to me that the CSS and JavaScript pulled from https://live.sympy.org are to support the SymPy Live add-in on the documentation site. So if we need to keep importing them, we will need to find a way to ensure the CSS from SymPy Live doesn't override the CSS from the documentation site.

bertiewooster:
> Is the SymPy Live thing just an accidental commit? I don't think we should add a submodule just for the docs build.

I just pushed a commit that removes that accidental submodule add.

asmeurer:
Yes, we need to make it so that the docs site can load when the live site is slow/down. Thanks for reminding me about that. I always assumed it was because of the extension. I didn't realize it was because we are loading the CSS from there. If we want do de-duplicate CSS, we should do it at the build stage, not when the docs load.

asmeurer:
Let's completely disable the Live extension for now. Once we get everything else how we like it, we can figure out what to do with it.

bertiewooster:
Automated `github-actions test / sphinx (pull_request)` fails:

```
Theme error:
no theme named 'furo' found (missing theme.conf?)
make: *** [Makefile:60: html] Error 2
Error: Process completed with exit code 2.
```

I believe this is because I added `furo` to the `requirements.txt` only in `doc/` (for the Docker build). Am I correct that I need to add `furo` to `requirements.txt` in `release/` as well for the automated test (and build) to succeed?

P.S. Same type of [failure is happening with Read The Docs (rtd)](https://github.com/sympy/sympy/pull/22803), which also required a Python package.

oscargus:
> Am I correct that I need to add `furo` to `requirements.txt` in `release/` as well for the automated test (and build) to succeed?

Yes, at the moment the tests use `release/requirements.txt` and although it would be better that the doctests use `doc/requirements.txt`, the quick way to get this demo to work is to add it to `release/requirements.txt`.

oscargus:
(And it seems like the Sphinx build passes with using `doc/requirements.txt` instead. See #22830 )

oscargus:
Restarting.

pradyunsg:
I'm unsubscribing from this, since I don't think there's much for me to add here -- feel free to @-mention me if there's questions for me. :)

PR #22658, opened 2021-12-13
docs: Supply Dockerfile for html docs build.

skieffer:
#### References to other Issues or PRs

Fixes #22655 

#### Brief description of what is fixed or changed

Provides one option to help developers build the HTML version of the docs.

#### Release Notes

NO ENTRY

Issue #20623, opened 2020-12-18
Pinned requirements in the optional dependencies tests


asmeurer:
The optional dependencies tests are installing from a pinned requirements file in the releases/ directory (see https://github.com/sympy/sympy/blob/master/release/aptinstall.sh and https://github.com/sympy/sympy/blob/master/release/requirements.txt). 

Is there a reason for this? I don't think it's a good idea to pin the dependencies, unless we have to. Otherwise, we won't catch bugs from new versions of the dependencies.

CC @oscarbenjamin 

I'm also surprised that we are using pip instead of conda. If pip works I guess that's fine, but I would expect conda to be necessary for at least some dependencies. 

Side note: is Sage being tested on GitHub Actions right now? I don't see it.

oscarbenjamin:
I made those files as a way to ensure it was possible to test all dependencies and build the docs as part of the release script where I wanted to pin the versions for reproducibility. Since I already had those ready to setup an Ubuntu 20.04 environment it was quick and easy to reuse them when migrating to actions in a hurry.

I think probably those files should be split up into several parts e.g. some parts are needed for tests and some for building the docs, some for apt and some for pip etc. There's also `update_requirements.sh` that has the bare pip command to install the latest versions of the same pieces.

The reasons for using conda over pip are not the same as they used to be since packaging with wheels etc has improved a lot over the years.

I don't know about sage. Maybe that got missed out...

asmeurer:
Well conda packaging has also improved because of conda-forge. Although I've heard that tensorflow is better to install with pip. 

For sage, we basically have to use the conda install. It also has to be a separate conda environment, because sage pins a lot of packages itself, and it also breaks mpmath when it is installed. I'm not sure why the Travis install pins a specific conda-forge channel. We might look into if that is still necessary.

oscarbenjamin:
> Although I've heard that tensorflow is better to install with pip.

I run the release script in a virtualbox VM. Initially I gave it 1GB of memory but it kept crashing while installing Tensorflow so I had to rebuild the VM with 2GB of memory. It turns out that `pip install tensorflow` uses something like 1.5GB of memory (briefly). I don't know what they're putting in that package or how `pip` is extracting the zip that that happens.

asmeurer:
[...]

PR #22803, opened 2022-01-06
DO NOT MERGE: Changing Sphinx theme to rtd

bertiewooster:

Attempting to change Sphinx theme to [rtd](https://pypi.org/project/sphinx-rtd-theme/) per issue #22716

#### Release Notes

NO ENTRY


doc/requirements.txt
```diff
@@ -6,6 +6,7 @@ myst-parser
 Sphinx==4.1.2
 sphinx-math-dollar==1.2
 sphinx-reredirects==0.0.1
+sphinx-rtd-theme
```
oscargus, line 9: The script seems to use `release/requirements.txt` (for unknown reasons).

oscarbenjamin: It's because I had made the `release/requirements.txt` file just before Travis got effectively disabled and we urgently needed to switch to Actions.

oscargus: I later on realized that the `doc/requirements.txt` was added quite recently. So the cause of the confusion is probably that rather than `release`.

asmeurer: We need to migrate the CI away from the release requirements file, and also make the requirements less pinned and only list top-level packages https://github.com/sympy/sympy/issues/20623.

asmeurer: In fact I don't know when the doc/requirements.txt was added, it also should not be using pins either, unless they are required (this is all of course completely unrelated to this PR).

asmeurer:
This line needs to be disabled for this to work, same as with the Furo PR https://github.com/sympy/sympy/pull/22803/files#diff-507f393ca988b45de131c01a7347f0aaeaaaf165cdce346bd722a8f352c0a3a8R134

asmeurer:
This theme has [quite a few options](https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html) that we should play with. For instance 

```py
html_theme_options = dict(
    collapse_navigation=False,
    navigation_depth=-1,
)
```

makes the so you can expand the sidebar items without changing pages. 

There do seem to be some bugs with the sidebar. For some reason, with the above setting, the "reference" section in the sidebar does not have expandable sections. This probably has something to do with how the RST in the pages is organized rather than the theme. 

Something to note about this theme is that it mixes both the current page and other pages in the same sidebar, similar to the existing theme. This can be considered an advantage or disadvantage. It can cause problems for pages with quite a lot of headers (for instance, the sidebar on this page https://docs.sympy.org/dev/modules/matrices/matrices.html, with the above settings, is a bit unwieldy). Part of the issue there will be solved if we can figure out autosummary (https://github.com/sympy/sympy/pull/22589), but that has some issues that may prevent it from happening soon. I do like that the sidebar scrolls independently, compared to the existing theme. The search bar is not pinned, as you have noted elsewhere. 

This theme seems to have similar problems as the existing theme for distinguishing between heading levels. There are also some bugs I noticed in the sidebar when using level 4 headers, where they don't nest properly. 

So far, my impression is that I prefer Furo to RTD. I do slightly prefer the way RTD is laid out in general to Furo, but that's purely aesthetic, and not a very strong sticking point compared to the other things.

bertiewooster:
> This line needs to be disabled for this to work, same as with the Furo PR

Thanks! Yes after disabling `html_style` in `doc/src/conf.py`, rtd builds correctly now:
rtd_logic

doc/requirements.txt
skieffer: `doc/requirements.txt` was added in #22658. I'll remove the pinned version numbers in #22801, which is currently open.

oscargus:
Restarting

PR #22806, opened 2022-01-06
Core/function : Fixes differentiation and series for Multivariate cases (more than 2 variable)

### LINKED ISSUE #22493: Series expansion introduces new variables
Hi all, I'm trying to automatically compute 2nd order accurate stencil values for a homework exercise (to set the context, doesn't matter). So just taylor expanding a function `u(x,y)` at various neighbouring points at spacing `h`.

If I run this:

```python
from sympy import *
x, y, h = symbols('x y h')
u = Function('u')
u(x-h, y).series(h, x0=0, n=3).simplify()
```

I get a perfectly reasonable output:
image

However, if I expand one of the corner values of the stencil
```python
from sympy import *
x, y, h = symbols('x y h')
u = Function('u')
u(x-h, y-h).series(h, x0=0, n=3).simplify()
```
I get this:

image

which has gained some strange `xi` and `_x` variables, resulting from the original `Subs` object in the expression. This is more clear if we print as ascii:

```
u(x, y) - h*(Derivative(u(x, y), x) + Derivative(u(x, y), y)) + h**2*(Derivative(u(x, y), (x, 2)) + Derivative(u(x, y), (y, 2)) + 2*Derivative(u(_x - xi, y), _x - xi, y))/2 + O(h**3)
```

Of course this is not what is expected. For example if I query WolframAlpha for the same question `"taylor expand u(x-h, y-h) up to 2rd order"` I get

image

which is the right answer.

Is this a bug or am I doing something wrong?

Kind regards,
Lennart

anutosh491:

Closes #22493 

While I was going through the issue , I realized that there's no error actually from sympy's side . We get the correct answer but then the function call for subs doesn't stop as intended .
For example while calculating the following we get the result as shown below 
```
>>> from sympy import *
>>> x, y, h = symbols('x y h')
>>> u = Function('u')
>>> u(x-h, y-h).series(h, x0=0, n=3).simplify()

                                             ⎛  2              2
                                           2 ⎜ ∂              ∂
                                          h ⋅⎜───(u(x, y)) + ───(u(x, y)) + 2⋅
                                             ⎜  2              2
            ⎛∂             ∂          ⎞      ⎝∂x             ∂y
u(x, y) - h⋅⎜──(u(x, y)) + ──(u(x, y))⎟ + ────────────────────────────────────
            ⎝∂x            ∂y         ⎠                                  2

      2                   ⎞
     ∂                    ⎟
───────────(u(_x - xi, y))⎟
∂y ∂_x - xi               ⎟
                          ⎠    ⎛ 3⎞
─────────────────────────── + O⎝h ⎠

```

Now the third term here is not what we expect to have , so I printed out the `self` term during the function call for `Subs` to see how sympy works for the same and after going through the code I realized there is no mistake from sympy's side as everything works as expected . We also get the correct answer but move on from that case 
```
# we encounter the correct forms here where the calls should maybe return but it goes on to give the wrong answer
    2
   ∂
───────( u(ξ₁, ξ₂))
∂ξ₂ ∂ξ₁
    2
   ∂
───────( u(ξ₁, ξ₂))
∂ξ₂ ∂ξ₁
    2
   ∂
───────( u(ξ₁, ξ₂))
∂ξ₂ ∂ξ₁
    2
   ∂
───────( u(ξ₁, ξ₂))
∂ξ₂ ∂ξ₁
 ⎛    2              ⎞│
 ⎜   ∂               ⎟│
-⎜───────( u(ξ₁, ξ₂))⎟│           # correct answer we expect but call goes on instead of stopping
 ⎝∂ξ₂ ∂ξ₁            ⎠│ξ₁=-h + x
      2
     ∂
────────────( u(ξ₁, _-h + y)) # the wrong form involving ∂_-h is encountered after the correct forms
∂_-h + y ∂ξ₁
```

The answer we expect obviously needs denesting and thus I realized that we do denest expressions of type `Sub` but we weren't doing so to expressions of type `-1*Sub` (basically type `Mul`) , hence leading to the wrong answer . Adding denesting for the above type of expressions fixes all cases as sympy was anyways working good enough for everything else

**Some wrong answers obtained on master**
```
>>> (u(x -h, y -h).diff(h, 2))

⎛  2                ⎞│            ⎛  2                ⎞│              ⎛      2
⎜ ∂                 ⎟│            ⎜ ∂                 ⎟│              ⎜     ∂
⎜────(u(ξ₁, -h + y))⎟│          + ⎜────(u(-h + x, ξ₂))⎟│          + 2⋅⎜───────
⎜   2               ⎟│            ⎜   2               ⎟│              ⎝∂_-h +
⎝∂ξ₁                ⎠│ξ₁=-h + x   ⎝∂ξ₂                ⎠│ξ₂=-h + y

                     ⎞│
                     ⎟│
─────(u(_-h + x, ξ₂))⎟│
x ∂ξ₂                ⎠│ξ₂=-h + y

>>> ((u(x -h, y -h).diff(h, 3)

 ⎛⎛  3                ⎞│            ⎛  3                ⎞│            ⎛
 ⎜⎜ ∂                 ⎟│            ⎜ ∂                 ⎟│            ⎜      ∂
-⎜⎜────(u(ξ₁, -h + y))⎟│          + ⎜────(u(-h + x, ξ₂))⎟│          + ⎜───────
 ⎜⎜   3               ⎟│            ⎜   3               ⎟│            ⎜
 ⎝⎝∂ξ₁                ⎠│ξ₁=-h + x   ⎝∂ξ₂                ⎠│ξ₂=-h + y   ⎝∂_-h +

3                     ⎞│            ⎛       3                     ⎞│
                      ⎟│            ⎜      ∂                      ⎟│
──────(u(ξ₁, _-h + y))⎟│          + ⎜─────────────(u(_-h + x, ξ₂))⎟│
 2                    ⎟│            ⎜        2                    ⎟│
y  ∂ξ₁                ⎠│ξ₁=-h + x   ⎝∂_-h + x  ∂ξ₂                ⎠│ξ₂=-h + y

    ⎛       3                     ⎞│         ⎞
    ⎜      ∂                      ⎟│         ⎟
+ 2⋅⎜─────────────(u(_-h + x, ξ₂))⎟│         ⎟
    ⎜            2                ⎟│         ⎟
    ⎝∂_-h + x ∂ξ₂                 ⎠│ξ₂=-h + y⎠

>>> u(x-h, y-h).series(h, x0=0, n=4).simplify()

                                             ⎛  2              2
                                           2 ⎜ ∂              ∂
                                          h ⋅⎜───(u(x, y)) + ───(u(x, y)) + 2⋅
                                             ⎜  2              2
            ⎛∂             ∂          ⎞      ⎝∂x             ∂y
u(x, y) - h⋅⎜──(u(x, y)) + ──(u(x, y))⎟ + ────────────────────────────────────
            ⎝∂x            ∂y         ⎠                                  2

      2                   ⎞      ⎛  3              3                    3
     ∂                    ⎟    3 ⎜ ∂              ∂                    ∂
───────────(u(_x - xi, y))⎟   h ⋅⎜───(u(x, y)) + ───(u(x, y)) + 2⋅────────────
∂y ∂_x - xi               ⎟      ⎜  3              3                2
                          ⎠      ⎝∂x             ∂y               ∂y  ∂_x - xi
─────────────────────────── - ────────────────────────────────────────────────


                        3                              3                     ⎞
                       ∂                              ∂                      ⎟
(u(_x - xi, y)) + ────────────(u(_x - xi, y)) + ─────────────(u(x, _-xi + y))⎟
                             2                              2                ⎟
                  ∂y ∂_x - xi                   ∂x ∂_-xi + y                 ⎠
──────────────────────────────────────────────────────────────────────────────
              6





    ⎛ 4⎞
 + O⎝h ⎠
```

**Correct answers obtained on committed branch**
```
>>> (u(x -h, y -h).diff(h, 2))

⎛  2                ⎞│            ⎛  2                ⎞│              ⎛    2
⎜ ∂                 ⎟│            ⎜ ∂                 ⎟│              ⎜   ∂
⎜────(u(ξ₁, -h + y))⎟│          + ⎜────(u(-h + x, ξ₂))⎟│          + 2⋅⎜───────
⎜   2               ⎟│            ⎜   2               ⎟│              ⎝∂ξ₂ ∂ξ₁
⎝∂ξ₁                ⎠│ξ₁=-h + x   ⎝∂ξ₂                ⎠│ξ₂=-h + y

           ⎞│
           ⎟│
(u(ξ₁, ξ₂))⎟│
           ⎠│ξ₁=-h + x, ξ₂=-h + y

>>> (u(x -h, y -h).diff(h, 3))


 ⎛⎛  3                ⎞│            ⎛  3                ⎞│              ⎛    3
 ⎜⎜ ∂                 ⎟│            ⎜ ∂                 ⎟│              ⎜   ∂
-⎜⎜────(u(ξ₁, -h + y))⎟│          + ⎜────(u(-h + x, ξ₂))⎟│          + 3⋅⎜─────
 ⎜⎜   3               ⎟│            ⎜   3               ⎟│              ⎜   2
 ⎝⎝∂ξ₁                ⎠│ξ₁=-h + x   ⎝∂ξ₂                ⎠│ξ₂=-h + y     ⎝∂ξ₂

              ⎞│                         ⎛    3              ⎞│
              ⎟│                         ⎜   ∂               ⎟│
───(u(ξ₁, ξ₂))⎟│                     + 3⋅⎜────────(u(ξ₁, ξ₂))⎟│
              ⎟│                         ⎜       2           ⎟│
∂ξ₁           ⎠│ξ₂=-h + y, ξ₁=-h + x     ⎝∂ξ₂ ∂ξ₁            ⎠│ξ₂=-h + y, ξ₁=-

     ⎞
     ⎟
     ⎟
     ⎟
h + x⎠
>>> (u(x -h, y -h).diff(h, 4))


⎛  4                ⎞│            ⎛  4                ⎞│              ⎛    4
⎜ ∂                 ⎟│            ⎜ ∂                 ⎟│              ⎜   ∂
⎜────(u(ξ₁, -h + y))⎟│          + ⎜────(u(-h + x, ξ₂))⎟│          + 4⋅⎜───────
⎜   4               ⎟│            ⎜   4               ⎟│              ⎜   3
⎝∂ξ₁                ⎠│ξ₁=-h + x   ⎝∂ξ₂                ⎠│ξ₂=-h + y     ⎝∂ξ₂  ∂ξ

            ⎞│                         ⎛     4              ⎞│
            ⎟│                         ⎜    ∂               ⎟│
─(u(ξ₁, ξ₂))⎟│                     + 6⋅⎜─────────(u(ξ₁, ξ₂))⎟│
            ⎟│                         ⎜   2    2           ⎟│
₁           ⎠│ξ₂=-h + y, ξ₁=-h + x     ⎝∂ξ₂  ∂ξ₁            ⎠│ξ₁=-h + x, ξ₂=-h

         ⎛    4              ⎞│
         ⎜   ∂               ⎟│
     + 4⋅⎜────────(u(ξ₁, ξ₂))⎟│
         ⎜       3           ⎟│
 + y     ⎝∂ξ₂ ∂ξ₁            ⎠│ξ₁=-h + x, ξ₂=-h + y

# All were returning wrong results before . Also checking for the issue whether that works fine .... does according to me!

>>> (u(x-h, y-h).series(h, x0=0, n=3).simplify())

                                             ⎛  2              2
                                           2 ⎜ ∂              ∂
                                          h ⋅⎜───(u(x, y)) + ───(u(x, y)) + 2⋅
                                             ⎜  2              2
            ⎛∂             ∂          ⎞      ⎝∂x             ∂y
u(x, y) - h⋅⎜──(u(x, y)) + ──(u(x, y))⎟ + ────────────────────────────────────
            ⎝∂x            ∂y         ⎠                            2

   2          ⎞
  ∂           ⎟
─────(u(x, y))⎟
∂y ∂x         ⎟
              ⎠    ⎛ 3⎞
─────────────── + O⎝h ⎠

>>> (u(x-h, y-h).series(h, x0=0, n=4).simplify())

                                             ⎛  2              2
                                           2 ⎜ ∂              ∂
                                          h ⋅⎜───(u(x, y)) + ───(u(x, y)) + 2⋅
                                             ⎜  2              2
            ⎛∂             ∂          ⎞      ⎝∂x             ∂y
u(x, y) - h⋅⎜──(u(x, y)) + ──(u(x, y))⎟ + ────────────────────────────────────
            ⎝∂x            ∂y         ⎠                            2

   2          ⎞      ⎛  3              3                 3                   3
  ∂           ⎟    3 ⎜ ∂              ∂                 ∂                   ∂
─────(u(x, y))⎟   h ⋅⎜───(u(x, y)) + ───(u(x, y)) + 3⋅──────(u(x, y)) + 3⋅────
∂y ∂x         ⎟      ⎜  3              3                2
              ⎠      ⎝∂x             ∂y               ∂y  ∂x              ∂y ∂
─────────────── - ────────────────────────────────────────────────────────────
                                                     6

           ⎞
           ⎟
──(u(x, y))⎟
 2         ⎟
x          ⎠    ⎛ 4⎞
──────────── + O⎝h ⎠

```

#### Other comments
ping @smichr @oscargus @0sidharth for review !

#### Release Notes

* series
  * Fixes differentiation and series for Multivariate cases (more than 2 variable) by adding denesting for a specific case.

### DIFF
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -2187,18 +2187,24 @@ class Subs(Expr):
 
         def denest(expr, variables, point):
             # Helper function to denest expressions involving Subs in some form.
-            # if expression isn't related to Subs, the expression passed is returned.
-            # Currently it can deal with expressions of type Mul(constant, Subs(...)).
+            # if expression isn't related to Subs, the sympified expression is returned.
+            # Currently it can deal with expressions of type Mul(expr, Subs(...)).
+            # where expr could be of type Constant, Mul, Number, Pow, Symbol.
+            # Check out functions: test_denesting_for_Subs1 and test_denesting_for_Subs2
+            # in core/tests/test_function.py for more information.
+            # Refer https://github.com/sympy/sympy/pull/22806#issue-1095031567
+            # for more information on implementation of this function.
             if isinstance(expr, Subs):
                 variables = expr.variables + variables
                 point = expr.point + point
                 expr = expr.expr
             elif isinstance(expr, Mul):
-                constant = expr.args[0]
-                _expr = expr.args[1]
-                if isinstance(_expr, Subs):
-                    expr, variables, point = denest(_expr, variables, point)
-                    expr = constant*expr
+                subs_arg = [arg for arg in expr.args if isinstance(arg, Subs)]
+                if subs_arg and len(subs_arg) == S.One:
+                    subs_arg = subs_arg[0]
+                    mul_arg = expr.extract_multiplicatively(subs_arg)
+                    expr, variables, point = denest(subs_arg, variables, point)
+                    expr = expr*mul_arg
             else:
                 expr = sympify(expr)
             return expr, variables, point
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -2188,15 +2188,17 @@ class Subs(Expr):
         def denest(expr, variables, point):
             # Helper function to denest expressions involving Subs in some form.
             # if expression isn't related to Subs, the expression passed is returned.
-            # TODO: Code to deal with expressions of type Mul(constant, Subs(...))
-            # which currently return wrong results.
+            # Currently it can deal with expressions of type Mul(constant, Subs(...)).
             if isinstance(expr, Subs):
                 variables = expr.variables + variables
                 point = expr.point + point
                 expr = expr.expr
-            elif isinstance(-expr, Subs):
-                expr, variables, point = denest(-expr, variables, point)
-                expr = -expr
+            elif isinstance(expr, Mul):
+                constant = expr.args[0]
+                _expr = expr.args[1]
+                if isinstance(_expr, Subs):
+                    expr, variables, point = denest(_expr, variables, point)
+                    expr = constant*expr
             else:
                 expr = sympify(expr)
             return expr, variables, point
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -2185,19 +2185,23 @@ class Subs(Expr):
         if not point:
             return sympify(expr)
 
-        # denest
-        if isinstance(expr, Subs):
-            variables = expr.variables + variables
-            point = expr.point + point
-            expr = expr.expr
-        elif isinstance(-expr, Subs):
-            expr = -expr
-            variables = expr.variables + variables
-            point = expr.point + point
-            expr = expr.expr
-            expr = -expr
-        else:
-            expr = sympify(expr)
+        def denest(expr, variables, point):
+            # Helper function to denest expressions involving Subs in some form.
+            # if expression isn't related to Subs, the expression passed is returned.
+            # TODO: Code to deal with expressions of type Mul(constant, Subs(...))
+            # which currently return wrong results.
+            if isinstance(expr, Subs):
+                variables = expr.variables + variables
+                point = expr.point + point
+                expr = expr.expr
+            elif isinstance(-expr, Subs):
+                expr, variables, point = denest(-expr, variables, point)
+                expr = -expr
+            else:
+                expr = sympify(expr)
+            return expr, variables, point
+
+        expr, variables, point = denest(expr, variables, point)
 
         # use symbols with names equal to the point value (with prepended _)
         # to give a variable-independent expression
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -2190,6 +2190,12 @@ class Subs(Expr):
             variables = expr.variables + variables
             point = expr.point + point
             expr = expr.expr
+        elif isinstance(-expr, Subs):
+            expr = -expr
+            variables = expr.variables + variables
+            point = expr.point + point
+            expr = expr.expr
+            expr = -expr
         else:
             expr = sympify(expr)
 
--- a/sympy/core/tests/test_function.py
+++ b/sympy/core/tests/test_function.py
@@ -1352,6 +1352,8 @@ def test_issue_22439():
 
 
 def test_denesting_for_Subs():
+    # Refer https://github.com/sympy/sympy/pull/22806#issuecomment-1037339531 to see the results
+    # being displayed in their pretty printed format.
     x, y, h = symbols('x y h')
     u = Function('u')
     _xi_1, _xi_2 = [Dummy() for i in range(2)]
--- a/sympy/core/tests/test_function.py
+++ b/sympy/core/tests/test_function.py
@@ -1351,6 +1351,36 @@ def test_issue_22439():
             3*Derivative(u(x, y), (x, 2), y))/6 + O(h**4)
 
 
+def test_denesting_for_Subs():
+    x, y, h = symbols('x y h')
+    u = Function('u')
+    _xi_1, _xi_2 = [Dummy() for i in range(2)]
+
+    assert u(x - 2*h, y - 2*h).diff(h, 2).simplify() == 4*(Subs(Derivative(u(_xi_1, -2*h + y), (_xi_1, 2)), _xi_1, -2*h + x) + \
+            Subs(Derivative(u(-2*h + x, _xi_2), (_xi_2, 2)), _xi_2, -2*h + y) + \
+            2*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, _xi_2), (_xi_1, _xi_2), (-2*h + x, -2*h + y)))
+    assert u(x + 3*h, y + 3*h).diff(h, 3).simplify() == 27*(Subs(Derivative(u(_xi_1, 3*h + y), (_xi_1, 3)), _xi_1, 3*h + x) + \
+            Subs(Derivative(u(3*h + x, _xi_2), (_xi_2, 3)), _xi_2, 3*h + y) + \
+            3*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, (_xi_2, 2)), (_xi_2, _xi_1), (3*h + y, 3*h + x)) + \
+            3*Subs(Derivative(u(_xi_1, _xi_2), (_xi_1, 2), _xi_2), (_xi_1, _xi_2), (3*h + x, 3*h + y)))
+    assert u(x + 2*h, y - 3*h).diff(h, 4).simplify() == 16*Subs(Derivative(u(_xi_1, -3*h + y), (_xi_1, 4)), _xi_1, 2*h + x) + \
+            81*Subs(Derivative(u(2*h + x, _xi_2), (_xi_2, 4)), _xi_2, -3*h + y) - \
+            216*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, (_xi_2, 3)), (_xi_2, _xi_1), (-3*h + y, 2*h + x)) + \
+            216*Subs(Derivative(u(_xi_1, _xi_2), (_xi_1, 2), (_xi_2, 2)), (_xi_2, _xi_1), (-3*h + y, 2*h + x)) - \
+            96*Subs(Derivative(u(_xi_1, _xi_2), (_xi_1, 3), _xi_2), (_xi_2, _xi_1), (-3*h + y, 2*h + x))
+    assert u(x - 2*h, y - 2*h).series(h, x0=0, n=4).simplify() == u(x, y) - 2*h*(Derivative(u(x, y), x) + Derivative(u(x, y), y)) + \
+            2*h**2*(Derivative(u(x, y), (x, 2)) + Derivative(u(x, y), (y, 2)) + 2*Derivative(u(x, y), x, y)) - \
+            4*h**3*(Derivative(u(x, y), (x, 3)) + Derivative(u(x, y), (y, 3)) + 3*Derivative(u(x, y), x, (y, 2)) + \
+            3*Derivative(u(x, y), (x, 2), y))/3 + O(h**4)
+    assert u(x + 3*h, y - 2*h).series(h, x0=0, n=6).simplify() == u(x, y) + h*(3*Derivative(u(x, y), x) - 2*Derivative(u(x, y), y)) + \
+            h**2*(9*Derivative(u(x, y), (x, 2)) + 4*Derivative(u(x, y), (y, 2)) - 12*Derivative(u(x, y), x, y))/2 + \
+            h**3*(27*Derivative(u(x, y), (x, 3)) - 8*Derivative(u(x, y), (y, 3)) + 36*Derivative(u(x, y), x, (y, 2)) - \
+            54*Derivative(u(x, y), (x, 2), y))/6 + h**4*(81*Derivative(u(x, y), (x, 4)) + 16*Derivative(u(x, y), (y, 4)) - \
+            96*Derivative(u(x, y), x, (y, 3)) + 216*Derivative(u(x, y), (x, 2), (y, 2)) - 216*Derivative(u(x, y), (x, 3), y))/24 + \
+            h**5*(243*Derivative(u(x, y), (x, 5)) - 32*Derivative(u(x, y), (y, 5)) + 240*Derivative(u(x, y), x, (y, 4)) - \
+            720*Derivative(u(x, y), (x, 2), (y, 3)) + 1080*Derivative(u(x, y), (x, 3), (y, 2)) - 810*Derivative(u(x, y), (x, 4), y))/120 + O(h**6)
+
+
 def test_negative_counts():
     # issue 13873
     raises(ValueError, lambda: sin(x).diff(x, -1))
--- a/sympy/core/tests/test_function.py
+++ b/sympy/core/tests/test_function.py
@@ -1325,6 +1325,8 @@ def test_issue_15084_13166():
 
 
 def test_issue_22439():
+    # Refer https://github.com/sympy/sympy/pull/22806#issuecomment-1007272066 to see the results
+    # being displayed in their pretty printed format.
     x, y, h = symbols('x y h')
     u = Function('u')
     _xi_1, _xi_2 = [Dummy() for i in range(2)]
--- a/sympy/core/tests/test_function.py
+++ b/sympy/core/tests/test_function.py
@@ -1324,6 +1324,30 @@ def test_issue_15084_13166():
     assert eq.diff(x, g(x)) == eq.diff(g(x), x)
 
 
+def test_issue_22439():
+    x, y, h = symbols('x y h')
+    u = Function('u')
+    assert u(x-h, y-h).diff(h, 2) == Subs(Derivative(u(_xi_1, -h + y), (_xi_1, 2)), _xi_1, -h + x) + \
+            Subs(Derivative(u(-h + x, _xi_2), (_xi_2, 2)), _xi_2, -h + y) + \
+            2*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, _xi_2), (_xi_1, _xi_2), (-h + x, -h + y))
+    assert u(x-h, y-h).diff(h, 3).simplify() == -Subs(Derivative(u(_xi_1, -h + y), (_xi_1, 3)), _xi_1, -h + x) - \
+            Subs(Derivative(u(-h + x, _xi_2), (_xi_2, 3)), _xi_2, -h + y) - \
+            3*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, (_xi_2, 2)), (_xi_2, _xi_1), (-h + y, -h + x)) - \
+            3*Subs(Derivative(u(_xi_1, _xi_2), (_xi_1, 2), _xi_2), (_xi_2, _xi_1), (-h + y, -h + x))
+    assert u(x-h, y-h).diff(h, 4) == Subs(Derivative(u(_xi_1, -h + y), (_xi_1, 4)), _xi_1, -h + x) + \
+            Subs(Derivative(u(-h + x, _xi_2), (_xi_2, 4)), _xi_2, -h + y) + \
+            4*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, (_xi_2, 3)), (_xi_2, _xi_1), (-h + y, -h + x)) + \
+            6*Subs(Derivative(u(_xi_1, _xi_2), (_xi_1, 2), (_xi_2, 2)), (_xi_1, _xi_2), (-h + x, -h + y)) + \
+            4*Subs(Derivative(u(_xi_1, _xi_2), (_xi_1, 3), _xi_2), (_xi_1, _xi_2), (-h + x, -h + y))
+    assert u(x-h, y-h).series(h, x0=0, n=3).simplify() == u(x, y) - h*(Derivative(u(x, y), x) + Derivative(u(x, y), y)) + \
+            h**2*(Derivative(u(x, y), (x, 2)) + Derivative(u(x, y), (y, 2)) + \
+            2*Derivative(u(x, y), x, y))/2 + O(h**3)
+    assert u(x-h, y-h).series(h, x0=0, n=4).simplify() == u(x, y) - h*(Derivative(u(x, y), x) + Derivative(u(x, y), y)) + \
+            h**2*(Derivative(u(x, y), (x, 2)) + Derivative(u(x, y), (y, 2)) + 2*Derivative(u(x, y), x, y))/2 - \
+            h**3*(Derivative(u(x, y), (x, 3)) + Derivative(u(x, y), (y, 3)) + 3*Derivative(u(x, y), x, (y, 2)) + \
+            3*Derivative(u(x, y), (x, 2), y))/6 + O(h**4)
+
+
 def test_negative_counts():
     # issue 13873
     raises(ValueError, lambda: sin(x).diff(x, -1))
--- a/sympy/core/tests/test_function.py
+++ b/sympy/core/tests/test_function.py
@@ -1354,6 +1354,8 @@ def test_issue_22439():
 def test_denesting_for_Subs():
     # Refer https://github.com/sympy/sympy/pull/22806#issuecomment-1037339531 to see the results
     # being displayed in their pretty printed format.
+    # This test addresses denesting for subs when we encounter expressions of type Mul(expr, Subs(...))
+    # where expr could be of type Symbol, Number/Constant.
     x, y, h = symbols('x y h')
     u = Function('u')
     _xi_1, _xi_2 = [Dummy() for i in range(2)]
@@ -1381,6 +1383,30 @@ def test_denesting_for_Subs():
             96*Derivative(u(x, y), x, (y, 3)) + 216*Derivative(u(x, y), (x, 2), (y, 2)) - 216*Derivative(u(x, y), (x, 3), y))/24 + \
             h**5*(243*Derivative(u(x, y), (x, 5)) - 32*Derivative(u(x, y), (y, 5)) + 240*Derivative(u(x, y), x, (y, 4)) - \
             720*Derivative(u(x, y), (x, 2), (y, 3)) + 1080*Derivative(u(x, y), (x, 3), (y, 2)) - 810*Derivative(u(x, y), (x, 4), y))/120 + O(h**6)
+
+
+def test_denesting_for_Subs2():
+    # Refer https://github.com/sympy/sympy/pull/22806/files#r805224808 to see the results
+    # being displayed in their pretty printed format.
+    # This test addresses denesting for subs when we encounter expressions of type Mul(expr, Subs(...))
+    # where expr could be of type Mul, Pow.
+    x, y, z, a, h = symbols('x y z a h')
+    u = Function('u')
+    _xi_1, _xi_2 = [Dummy() for i in range(2)]
+
+    assert u(x - 2*z*h, y - 2*z*h).diff(h, 3) == -8*z**3*(Subs(Derivative(u(_xi_1, -2*h*z + y), (_xi_1, 3)), _xi_1, -2*h*z + x) + \
+            Subs(Derivative(u(-2*h*z + x, _xi_2), (_xi_2, 3)), _xi_2, -2*h*z + y) + \
+            3*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, (_xi_2, 2)), (_xi_1, _xi_2), (-2*h*z + x, -2*h*z + y)) + \
+            3*Subs(Derivative(u(_xi_1, _xi_2), (_xi_1, 2), _xi_2), (_xi_2, _xi_1), (-2*h*z + y, -2*h*z + x)))
+    assert u(x - a*z**2*h, y - a*z**2*h).diff(h, 2) == a**2*z**4*(Subs(Derivative(u(_xi_1, -a*h*z**2 + y), (_xi_1, 2)), _xi_1, -a*h*z**2 + x) + \
+            Subs(Derivative(u(-a*h*z**2 + x, _xi_2), (_xi_2, 2)), _xi_2, -a*h*z**2 + y) + \
+            2*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, _xi_2), (_xi_2, _xi_1), (-a*h*z**2 + y, -a*h*z**2 + x)))
+    assert u(x - 3*z*h, y + 3*z*h).series(h, 0, 4).simplify() == u(x, y) - 3*h*z*(Derivative(u(x, y), x) - Derivative(u(x, y), y)) + \
+            9*h**2*z**2*(Derivative(u(x, y), (x, 2)) + Derivative(u(x, y), (y, 2)) - \
+            2*Derivative(u(x, y), x, y))/2 + 9*h**3*z**3*(-Derivative(u(x, y), (x, 3)) + Derivative(u(x, y), (y, 3)) - \
+            3*Derivative(u(x, y), x, (y, 2)) + 3*Derivative(u(x, y), (x, 2), y))/2 + O(h**4)
+    assert u(x + a*z**2*h, y + a*z**2*h).series(h, 0, 3).simplify() == u(x, y) + a*h*z**2*(Derivative(u(x, y), x) + Derivative(u(x, y), y)) + \
+            a**2*h**2*z**4*(Derivative(u(x, y), (x, 2)) + Derivative(u(x, y), (y, 2)) + 2*Derivative(u(x, y), x, y))/2 + O(h**3)
 
 
 def test_negative_counts():
--- a/sympy/core/tests/test_function.py
+++ b/sympy/core/tests/test_function.py
@@ -1327,22 +1327,23 @@ def test_issue_15084_13166():
 def test_issue_22439():
     x, y, h = symbols('x y h')
     u = Function('u')
-    assert u(x-h, y-h).diff(h, 2) == Subs(Derivative(u(_xi_1, -h + y), (_xi_1, 2)), _xi_1, -h + x) + \
+    _xi_1, _xi_2 = [Dummy() for i in range(2)]
+    assert u(x - h, y - h).diff(h, 2) == Subs(Derivative(u(_xi_1, -h + y), (_xi_1, 2)), _xi_1, -h + x) + \
             Subs(Derivative(u(-h + x, _xi_2), (_xi_2, 2)), _xi_2, -h + y) + \
             2*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, _xi_2), (_xi_1, _xi_2), (-h + x, -h + y))
-    assert u(x-h, y-h).diff(h, 3).simplify() == -Subs(Derivative(u(_xi_1, -h + y), (_xi_1, 3)), _xi_1, -h + x) - \
-            Subs(Derivative(u(-h + x, _xi_2), (_xi_2, 3)), _xi_2, -h + y) - \
-            3*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, (_xi_2, 2)), (_xi_2, _xi_1), (-h + y, -h + x)) - \
-            3*Subs(Derivative(u(_xi_1, _xi_2), (_xi_1, 2), _xi_2), (_xi_2, _xi_1), (-h + y, -h + x))
+    assert u(x + h, y - h).diff(h, 3) == Subs(Derivative(u(_xi_1, -h + y), (_xi_1, 3)), _xi_1, h + x) - \
+            Subs(Derivative(u(h + x, _xi_2), (_xi_2, 3)), _xi_2, -h + y) + \
+            3*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, (_xi_2, 2)), (_xi_1, _xi_2), (h + x, -h + y)) - \
+            3*Subs(Derivative(u(_xi_1, _xi_2), (_xi_1, 2), _xi_2), (_xi_1, _xi_2), (h + x, -h + y))
     assert u(x-h, y-h).diff(h, 4) == Subs(Derivative(u(_xi_1, -h + y), (_xi_1, 4)), _xi_1, -h + x) + \
             Subs(Derivative(u(-h + x, _xi_2), (_xi_2, 4)), _xi_2, -h + y) + \
             4*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, (_xi_2, 3)), (_xi_2, _xi_1), (-h + y, -h + x)) + \
             6*Subs(Derivative(u(_xi_1, _xi_2), (_xi_1, 2), (_xi_2, 2)), (_xi_1, _xi_2), (-h + x, -h + y)) + \
             4*Subs(Derivative(u(_xi_1, _xi_2), (_xi_1, 3), _xi_2), (_xi_1, _xi_2), (-h + x, -h + y))
-    assert u(x-h, y-h).series(h, x0=0, n=3).simplify() == u(x, y) - h*(Derivative(u(x, y), x) + Derivative(u(x, y), y)) + \
+    assert u(x - h, y - h).series(h, x0 = 0, n = 3).simplify() == u(x, y) - h*(Derivative(u(x, y), x) + Derivative(u(x, y), y)) + \
             h**2*(Derivative(u(x, y), (x, 2)) + Derivative(u(x, y), (y, 2)) + \
             2*Derivative(u(x, y), x, y))/2 + O(h**3)
-    assert u(x-h, y-h).series(h, x0=0, n=4).simplify() == u(x, y) - h*(Derivative(u(x, y), x) + Derivative(u(x, y), y)) + \
+    assert u(x - h, y - h).series(h, x0 = 0, n = 4).simplify() == u(x, y) - h*(Derivative(u(x, y), x) + Derivative(u(x, y), y)) + \
             h**2*(Derivative(u(x, y), (x, 2)) + Derivative(u(x, y), (y, 2)) + 2*Derivative(u(x, y), x, y))/2 - \
             h**3*(Derivative(u(x, y), (x, 3)) + Derivative(u(x, y), (y, 3)) + 3*Derivative(u(x, y), x, (y, 2)) + \
             3*Derivative(u(x, y), (x, 2), y))/6 + O(h**4)


anutosh491:
Pretty Printing all tests for reference . I am satisfied with the changes as I tried a lot of different cases ,(all possible combinations of x, y and h as parameters in our function eg `u(-x -h, y+h)/ u(-x+h, y-h)` etc etc and all seem to work fine . Quite a few anyways worked fine on master . If all tests pass in the latest commit maybe this could be merged !
```
>>> from sympy import *
>>> x, y, h = symbols('x y h')
>>> u = Function('u')
>>> pprint(u(x - h, y - h).diff(h, 2))

⎛  2                ⎞│            ⎛  2                ⎞│              ⎛    2
⎜ ∂                 ⎟│            ⎜ ∂                 ⎟│              ⎜   ∂
⎜────(u(ξ₁, -h + y))⎟│          + ⎜────(u(-h + x, ξ₂))⎟│          + 2⋅⎜───────
⎜   2               ⎟│            ⎜   2               ⎟│              ⎝∂ξ₂ ∂ξ₁
⎝∂ξ₁                ⎠│ξ₁=-h + x   ⎝∂ξ₂                ⎠│ξ₂=-h + y

           ⎞│
           ⎟│
(u(ξ₁, ξ₂))⎟│
           ⎠│ξ₁=-h + x, ξ₂=-h + y

>>> pprint(u(x + h, y - h).diff(h, 3))

⎛  3                ⎞│           ⎛  3               ⎞│              ⎛    3
⎜ ∂                 ⎟│           ⎜ ∂                ⎟│              ⎜   ∂
⎜────(u(ξ₁, -h + y))⎟│         - ⎜────(u(h + x, ξ₂))⎟│          + 3⋅⎜────────(
⎜   3               ⎟│           ⎜   3              ⎟│              ⎜   2
⎝∂ξ₁                ⎠│ξ₁=h + x   ⎝∂ξ₂               ⎠│ξ₂=-h + y     ⎝∂ξ₂  ∂ξ₁

          ⎞│                        ⎛    3              ⎞│
          ⎟│                        ⎜   ∂               ⎟│
u(ξ₁, ξ₂))⎟│                    - 3⋅⎜────────(u(ξ₁, ξ₂))⎟│
          ⎟│                        ⎜       2           ⎟│
          ⎠│ξ₁=h + x, ξ₂=-h + y     ⎝∂ξ₂ ∂ξ₁            ⎠│ξ₁=h + x, ξ₂=-h + y
>>> pprint(u(x - h, y - h).diff(h, 4))

⎛  4                ⎞│            ⎛  4                ⎞│              ⎛    4
⎜ ∂                 ⎟│            ⎜ ∂                 ⎟│              ⎜   ∂
⎜────(u(ξ₁, -h + y))⎟│          + ⎜────(u(-h + x, ξ₂))⎟│          + 4⋅⎜───────
⎜   4               ⎟│            ⎜   4               ⎟│              ⎜   3
⎝∂ξ₁                ⎠│ξ₁=-h + x   ⎝∂ξ₂                ⎠│ξ₂=-h + y     ⎝∂ξ₂  ∂ξ

            ⎞│                         ⎛     4              ⎞│
            ⎟│                         ⎜    ∂               ⎟│
─(u(ξ₁, ξ₂))⎟│                     + 6⋅⎜─────────(u(ξ₁, ξ₂))⎟│
            ⎟│                         ⎜   2    2           ⎟│
₁           ⎠│ξ₂=-h + y, ξ₁=-h + x     ⎝∂ξ₂  ∂ξ₁            ⎠│ξ₁=-h + x, ξ₂=-h

         ⎛    4              ⎞│
         ⎜   ∂               ⎟│
     + 4⋅⎜────────(u(ξ₁, ξ₂))⎟│
         ⎜       3           ⎟│
 + y     ⎝∂ξ₂ ∂ξ₁            ⎠│ξ₁=-h + x, ξ₂=-h + y

>>> pprint(u(x - h, y - h).series(h, x0 = 0, n = 3).simplify())

                                             ⎛  2              2
                                           2 ⎜ ∂              ∂
                                          h ⋅⎜───(u(x, y)) + ───(u(x, y)) + 2⋅
                                             ⎜  2              2
            ⎛∂             ∂          ⎞      ⎝∂x             ∂y
u(x, y) - h⋅⎜──(u(x, y)) + ──(u(x, y))⎟ + ────────────────────────────────────
            ⎝∂x            ∂y         ⎠                            2

   2          ⎞
  ∂           ⎟
─────(u(x, y))⎟
∂y ∂x         ⎟
              ⎠    ⎛ 3⎞
─────────────── + O⎝h ⎠

>>> pprint(u(x - h, y - h).series(h, x0 = 0, n = 4).simplify())

                                             ⎛  2              2
                                           2 ⎜ ∂              ∂
                                          h ⋅⎜───(u(x, y)) + ───(u(x, y)) + 2⋅
                                             ⎜  2              2
            ⎛∂             ∂          ⎞      ⎝∂x             ∂y
u(x, y) - h⋅⎜──(u(x, y)) + ──(u(x, y))⎟ + ────────────────────────────────────
            ⎝∂x            ∂y         ⎠                            2

   2          ⎞      ⎛  3              3                 3                   3
  ∂           ⎟    3 ⎜ ∂              ∂                 ∂                   ∂
─────(u(x, y))⎟   h ⋅⎜───(u(x, y)) + ───(u(x, y)) + 3⋅──────(u(x, y)) + 3⋅────
∂y ∂x         ⎟      ⎜  3              3                2
              ⎠      ⎝∂x             ∂y               ∂y  ∂x              ∂y ∂
─────────────── - ────────────────────────────────────────────────────────────
                                                     6

           ⎞
           ⎟
──(u(x, y))⎟
 2         ⎟
x          ⎠    ⎛ 4⎞
──────────── + O⎝h ⎠
```

anutosh491:
ping @jksuom for review

anutosh491:
Ping @smichr for review!

anutosh491:
@smichr one of your Pr's had added denesting for Subs , I had some doubts regarding few cases hence I wanted to ask whether we could do without the denesting block ? Like does it have a lot of use case or could we do without it !
Also does 1 and 2 here represent the same thing . doesn't seem too big a difference !
```
1) 
  ⎛⎛    2             ⎞│         ⎞│
  ⎜⎜   ∂              ⎟│         ⎟│
2⋅⎜⎜───────(u(ξ₁, ξ₂))⎟│         ⎟│
  ⎝⎝∂ξ₂ ∂ξ₁           ⎠│ξ₁=-h + x⎠│ξ₂=-h + y 

2)
  ⎛    2             ⎞│
  ⎜   ∂              ⎟│
2⋅⎜───────(u(ξ₁, ξ₂))⎟│
  ⎝∂ξ₂ ∂ξ₁           ⎠│ξ₁=-h + x, ξ₂=-h + y
```

anutosh491:
ping @0sidharth for review

anutosh491:
ping @jksuom for review

anutosh491:
ping @smichr

anutosh491:
ping @smichr for review !

anutosh491:
CC: @smichr haven't heard anything on this pr , would be glad to have your views on the same !

sympy/core/function.py
```diff
@@ -2190,6 +2190,12 @@ def __new__(cls, expr, variables, point, **assumptions):
             variables = expr.variables + variables
             point = expr.point + point
             expr = expr.expr
+        elif isinstance(-expr, Subs):
```
ThePauliPrinciple, line 2193: This now duplicates the code of the `isinstance(expr, Subs)` branch, could it be rewritten such that if a change needs to be made that it only needs to be made once rather than twice (or simply forgotten because that author is not aware of the specifics of the `-expr` branch?)

sympy/core/tests/test_function.py
```diff
@@ -1324,6 +1324,31 @@ def test_issue_15084_13166():
     assert eq.diff(x, g(x)) == eq.diff(g(x), x)
 
 
+def test_issue_22439():
+    x, y, h = symbols('x y h')
+    u = Function('u')
+    _xi_1, _xi_2 = [Dummy() for i in range(2)]
+    assert u(x - h, y - h).diff(h, 2) == Subs(Derivative(u(_xi_1, -h + y), (_xi_1, 2)), _xi_1, -h + x) + \
```
ThePauliPrinciple, line 1331: Reviewing these kind of tests is rather involved since it requires to check the actual result (as in, should the derivative really be equal to what is stated here). Could you create tests based on some prior result from a reliable source (e.g. definitions in wiki pages or wolfram). With a link, this would also make it easier in the future to review if some problem comes up. If it doesn't exist then so be it.

anutosh491: Cool , yeah maybe I could try this too. I was quick to realize that the rather difficult part of this PR was to add tests , cause the expression/problem mentioned in the issue reported in Op was not a trivial day to day result which a general sympy user might end up trying ! 
Hence the best I could do was to create pretty printed results of all these tests and add them into a comment for reference which is here https://github.com/sympy/sympy/pull/22806#issuecomment-1007272066

sympy/core/function.py
anutosh491: This diff would do the job for us , what say @ThePauliPrinciple .A functional implementation looks good to me otherwise it's getting too shabby and the original committed code is looking better , not sure if this could be improved but looks fine !
```
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -2185,6 +2185,21 @@ def __new__(cls, expr, variables, point, **assumptions):
         if not point:
             return sympify(expr)

+        def denest(expr, variables, point):
+            # denest
+            if isinstance(expr, Subs):
+                variables = expr.variables + variables
+                point = expr.point + point
+                expr = expr.expr
+            elif isinstance(-expr, Subs):
+                expr, variables, point = denest(-expr, variables, point)
+                expr = -expr
+            else:
+                expr = sympify(expr)
+            return expr, variables, point
+
+        expr, variables, point = denest(expr, variables, point)
         # denest
         if isinstance(expr, Subs):
             variables = expr.variables + variables
```

ThePauliPrinciple: It would be helpful if a comment is added why this is done, so a future author can easily understand it. In the context of the issue it is clear, but without that context it is not clear what is happening here.

I'm also wondering of `-Subs` is just a specific case of `Mul(constant, Subs(...))`, e.g. is `u(x - 3*h, y - 3*h)` currently supported? If not, could it be supported if we look for any `Mul` pattern with `Subs` rather than only `Mul(-1,Subs)`?

anutosh491: Will try this one out !

>it would be helpful if a comment is added why this is done, so a future author can easily understand it. In the context of the issue it is clear, but without that context it is not clear what is happening here.
Yupp, i was planning to do the same !

anutosh491: > I'm also wondering of `-Subs` is just a specific case of `Mul(constant, Subs(...))`, e.g. is `u(x - 3*h, y - 3*h)` currently supported? If not, could it be supported if we look for any `Mul` pattern with `Subs` rather than only `Mul(-1,Subs)`?

Nope , there isn't code for that currently ! cause `isinstance(-1*Mul, Subs)` is `True` if the `Mul` variable is `-Subs`, but the same isn't true for `(-constant*Mul, Subs)` . Hence master and the committed branch both give the same error !

ThePauliPrinciple: I think you can just generalise what you have here (but I'm not entirely certain, you'd have to confirm the result by hands):

Instead of checking for `isinstance(-expr, Subs)`, check for `isinstance(expr, Mul)`, then in that branch check if any (or more precise, exactly one) of the args is an instance of `Subs`, if it is, apply denest on the arg that is a `Subs` and multiply that result with all the other args (that should do the same if `expr == Mul(-1, Subs(..))` as the current code) but I think it handles any other constants too, including symbolic ones). If no `Subs` is found we `simplify` by default.

It might be possible to apply similar logic to e.g. `Pow` or `Function` too!

anutosh491: Yupp, trying it out currently . Will try out all ways and get back to you whether we can get it done somehow , or like is it somewhat approachable with the already written code !

anutosh491: > Instead of checking for `isinstance(-expr, Subs)`, check for `isinstance(expr, Mul)`, then in that branch check if any (or more precise, exactly one) of the args is an instance of `Subs`, if it is, apply denest on the arg that is a `Subs` and multiply that result with all the other args (that should do the same if `expr == Mul(-1, Subs(..))` as the current code) but I think it handles any other constants too, including symbolic ones). If no `Subs` is found we `simplify` by default.

Thanks for the suggestions , work perfectly fine ! @ThePauliPrinciple , I'll make the commit soon .

sympy/core/function.py
```diff
@@ -2185,13 +2185,25 @@ def __new__(cls, expr, variables, point, **assumptions):
...
+        def denest(expr, variables, point):
+            # Helper function to denest expressions involving Subs in some form.
+            # if expression isn't related to Subs, the expression passed is returned.
+            # Currently it can deal with expressions of type Mul(constant, Subs(...)).
+            if isinstance(expr, Subs):
+                variables = expr.variables + variables
+                point = expr.point + point
+                expr = expr.expr
+            elif isinstance(expr, Mul):
+                constant = expr.args[0]
```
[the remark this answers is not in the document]
anutosh491: Well if we are tackling `Mul(const, Subs(..))` then I realized it would end up like this ! But yeah what you say is correct . Though even if the multiplicative part isn't a constant it'll end up similarly 
If we have the expression of the form 
```
>>> x, y, h, z = symbols('x y h z')
>>> u = Function('u')
>>> u(x + z*h, y + z*h).diff(h, 2)
```
we would end up getting `z*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, _xi_2), _xi_2, h*z + y)` which is being denested fine and the answer obtained is also correct 
```

   ⎛⎛  2                 ⎞│             ⎛  2                 ⎞│
 2 ⎜⎜ ∂                  ⎟│             ⎜ ∂                  ⎟│
z ⋅⎜⎜────(u(ξ₁, h⋅z + y))⎟│           + ⎜────(u(h⋅z + x, ξ₂))⎟│           + 2⋅
   ⎜⎜   2                ⎟│             ⎜   2                ⎟│
   ⎝⎝∂ξ₁                 ⎠│ξ₁=h⋅z + x   ⎝∂ξ₂                 ⎠│ξ₂=h⋅z + y

⎛    2             ⎞│                      ⎞
⎜   ∂              ⎟│                      ⎟
⎜───────(u(ξ₁, ξ₂))⎟│                      ⎟
⎝∂ξ₂ ∂ξ₁           ⎠│ξ₂=h⋅z + y, ξ₁=h⋅z + x⎟
                                           ⎠
```
But yeah what is the best way out to handle this ?? Maybe looping through the args , taking out the subs part and keeping other things together ? Also I'm not sure how would we approach more than 2 args , like how could we deal with something like `2*z*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, _xi_2), _xi_2, 2*h*z + y)` or `a***2*z*Subs(Derivative(u(_xi_1, _xi_2), _xi_1, _xi_2), _xi_2, 2*h*z + y)` and other random cases . I'll give this a try and check if I find something substantial !

anutosh491:
The changes in the latest commit have been explained in https://github.com/sympy/sympy/pull/22806#discussion_r805224808 (I made two edits in the comment which sort of shows how to got to the point of committing the change). Anyways thanks for the great reviews on this Pr , makes me much more confident of this Pr now !

anutosh491:
CC: @ThePauliPrinciple ,does this look good to you ? I feel some good cases have been addressed in the previous 2 commits

anutosh491:
CC: @ThePauliPrinciple

anutosh491:
> Please don't repeatedly ping like this. You need to allow time for people to respond.

So sorry for that . I know I shouldn't do it , I felt this was close to go in hence thought of asking for reviews but I assure you I am not doing this again !

ThePauliPrinciple:
> It doesn't benefit most of the time

Since I already posted in this PR, I'm getting emails when the PR is updated, so pinging doesn't really help. It might be nice if there is some place where people can indicate about what kind of topics they would like to get pinged, so that at least we can be clear about that.

The suggested changes seem like the right approach to me. The tests are extensive and cover the new functionality, however, I can't confirm myself that the proposed results for the tests are mathematically correct. Is there any external reference which shows what the result of differentiation should be for these type of expressions?

anutosh491:
> The suggested changes seem like the right approach to me. The tests are extensive and cover the new functionality, however, I can't confirm myself that the proposed results for the tests are mathematically correct. Is there any external reference which shows what the result of differentiation should be for these type of expressions?

Hey sorry for bugging you ,I just got excited as I'm much more confident of this pr now after your reviews and see this going in anytime soon ! But to answer your question ,how could we approach your concern ? 
1) I've pasted pretty prints of all results in the comments above so maybe that's a reference points 
2) Secondly I can paste wolfram links for all the tests on this pr here cause I am quite confident of the results mathematically . I only committed the tests only after making sure that a number of random cases were working for the same !
 The user who reported the issue mentioned above in the OP has also pasted the answer which he would like to have or rather the mathematically correct answer for the question, and the test added for that issue perfectly with what is expected.

anutosh491:
I'll list down some results I could find from my search history . I wish I could find better ways to convey my point ! Anyways good place for reference to check mathematical correctness .

**test_issue_22439()**
[1st assertion test](https://www.wolframalpha.com/input?i2d=true&i=+D%5Bu%5C%2840%29f+-+x%5C%2844%29+h++-+x%5C%2841%29%2C%7Bx%2C2%7D%5D)
[2nd assertion test ](https://www.wolframalpha.com/input?i2d=true&i=differentiate+u%5C%2840%29f+%2B+x%5C%2844%29+g-+x%5C%2841%29+thrice)
[3rd/4rth/5th assertion test ](https://www.wolframalpha.com/input?i2d=true&i=differentiate+u%5C%2840%29f+-+x%5C%2844%29+g-+x%5C%2841%29+four+times)

**test_denesting_for_Subs()**
[1st/4rth assertion test ](https://www.wolframalpha.com/input?i2d=true&i=differentiate+u%5C%2840%29f+-+2x%5C%2844%29+g-+2x%5C%2841%29+twice)
[2nd assertion test](https://www.wolframalpha.com/input?i2d=true&i=differentiate+u%5C%2840%29f+%2B+3x%5C%2844%29+g+%2B++3x%5C%2841%29+thrice)
[3rd assertion test](https://www.wolframalpha.com/input?i2d=true&i=differentiate+u%5C%2840%29f+%2B+2x%5C%2844%29+g+-++3x%5C%2841%29+four+times)

**test_denesting_for_Subs2()**
[1st assertion test](https://www.wolframalpha.com/input?i2d=true&i=differentiate+u%5C%2840%29f+-+2ax%5C%2844%29+g-+2ax%5C%2841%29+twice)
[2nd assertion test ](https://www.wolframalpha.com/input?i2d=true&i=differentiate+u%5C%2840%29f+-+Power%5Bz%2C2%5Dax%5C%2844%29+g-+Power%5Bz%2C2%5Dax%5C%2841%29+twice)
[3rd assertion test ](https://www.wolframalpha.com/input?i2d=true&i=series+u%5C%2840%29f+-+3zx%5C%2844%29+g+%2B+3zx%5C%2841%29)

anutosh491:
I ended up trying some more sophisticated random tests today to confirm whether this is completely ready to go in. I'll paste the more basic ones here . These are some greats tests and would have surely liked to add these too but would get a bit too much as already we have good tests for the same .
```
>>> pprint(u(x -h**2, y -h**2).diff(h, 2).expand())

     ⎛  2                  ⎞│                   ⎛  2                  ⎞│
   2 ⎜ ∂  ⎛ ⎛       2    ⎞⎞⎟│                 2 ⎜ ∂  ⎛ ⎛   2        ⎞⎞⎟│
4⋅h ⋅⎜────⎝u⎝ξ₁, - h  + y⎠⎠⎟│            + 4⋅h ⋅⎜────⎝u⎝- h  + x, ξ₂⎠⎠⎟│
     ⎜   2                 ⎟│      2            ⎜   2                 ⎟│
     ⎝∂ξ₁                  ⎠│ξ₁=- h  + x        ⎝∂ξ₂                  ⎠│ξ₂=- h

             ⎛    2             ⎞│
           2 ⎜   ∂              ⎟│                             ⎛ ∂ ⎛ ⎛       2
      + 8⋅h ⋅⎜───────(u(ξ₁, ξ₂))⎟│      2            2     - 2⋅⎜───⎝u⎝ξ₁, - h
2            ⎝∂ξ₂ ∂ξ₁           ⎠│ξ₂=- h  + y, ξ₁=- h  + x     ⎝∂ξ₁
  + y


    ⎞⎞⎞│                ⎛ ∂ ⎛ ⎛   2        ⎞⎞⎞│
 + y⎠⎠⎟│      2     - 2⋅⎜───⎝u⎝- h  + x, ξ₂⎠⎠⎟│      2
      ⎠│ξ₁=- h  + x     ⎝∂ξ₂                 ⎠│ξ₂=- h  + y
>>> 
>>> 
>>> pprint(u(x -h**2, y - 1/h**2).diff(h, 2).expand())

     ⎛  2                ⎞│
   2 ⎜ ∂  ⎛ ⎛        1 ⎞⎞⎟│                ⎛ ∂ ⎛ ⎛        1 ⎞⎞⎞│
4⋅h ⋅⎜────⎜u⎜ξ₁, y - ──⎟⎟⎟│            - 2⋅⎜───⎜u⎜ξ₁, y - ──⎟⎟⎟│            -
     ⎜   2⎜ ⎜         2⎟⎟⎟│      2         ⎜∂ξ₁⎜ ⎜         2⎟⎟⎟│      2
     ⎝∂ξ₁ ⎝ ⎝        h ⎠⎠⎠│ξ₁=- h  + x     ⎝   ⎝ ⎝        h ⎠⎠⎠│ξ₁=- h  + x


  ⎛    2             ⎞│
  ⎜   ∂              ⎟│                           ⎛ ∂ ⎛ ⎛   2        ⎞⎞⎞│
8⋅⎜───────(u(ξ₁, ξ₂))⎟│      2             1    6⋅⎜───⎝u⎝- h  + x, ξ₂⎠⎠⎟│
  ⎝∂ξ₂ ∂ξ₁           ⎠│ξ₁=- h  + x, ξ₂=y - ──     ⎝∂ξ₂                 ⎠│ξ₂=y
                                            2
                                           h
───────────────────────────────────────────── - ──────────────────────────────
                       2                                         4
                      h                                         h

         ⎛  2                  ⎞│
         ⎜ ∂  ⎛ ⎛   2        ⎞⎞⎟│
       4⋅⎜────⎝u⎝- h  + x, ξ₂⎠⎠⎟│
  1      ⎜   2                 ⎟│       1
- ──     ⎝∂ξ₂                  ⎠│ξ₂=y - ──
   2                                     2
  h                                     h
──── + ───────────────────────────────────
                         6
                        h
>>>
>>> pprint(u(x + h**3, y - h**2).diff(h, 2).expand())

     ⎛  2                  ⎞│                  ⎛    2             ⎞│
   4 ⎜ ∂  ⎛ ⎛       2    ⎞⎞⎟│                3 ⎜   ∂              ⎟│
9⋅h ⋅⎜────⎝u⎝ξ₁, - h  + y⎠⎠⎟│          - 12⋅h ⋅⎜───────(u(ξ₁, ξ₂))⎟│    3
     ⎜   2                 ⎟│    3             ⎝∂ξ₂ ∂ξ₁           ⎠│ξ₁=h  + x,
     ⎝∂ξ₁                  ⎠│ξ₁=h  + x

                    ⎛  2                ⎞│
                  2 ⎜ ∂  ⎛ ⎛ 3        ⎞⎞⎟│                  ⎛ ∂ ⎛ ⎛       2
       2     + 4⋅h ⋅⎜────⎝u⎝h  + x, ξ₂⎠⎠⎟│            + 6⋅h⋅⎜───⎝u⎝ξ₁, - h  +
 ξ₂=- h  + y        ⎜   2               ⎟│      2           ⎝∂ξ₁
                    ⎝∂ξ₂                ⎠│ξ₂=- h  + y


 ⎞⎞⎞│              ⎛ ∂ ⎛ ⎛ 3        ⎞⎞⎞│
y⎠⎠⎟│    3     - 2⋅⎜───⎝u⎝h  + x, ξ₂⎠⎠⎟│      2
   ⎠│ξ₁=h  + x     ⎝∂ξ₂               ⎠│ξ₂=- h  + y
```
Wolfram links for confirming answers
[first](https://www.wolframalpha.com/input?i2d=true&i=differentiate+u%5C%2840%29f+-+Power%5Bx%2C2%5D%5C%2844%29+g-+Power%5Bx%2C2%5D%5C%2841%29+twice)
[Second](https://www.wolframalpha.com/input?i2d=true&i=differentiate+u%5C%2840%29f+-+Power%5Bx%2C2%5D%5C%2844%29+g-+Divide%5B1%2CPower%5Bx%2C2%5D%5D%5C%2841%29+twice)
[Third](https://www.wolframalpha.com/input?i2d=true&i=differentiate+u%5C%2840%29f+%2BPower%5Bx%2C3%5D%5C%2844%29+g-+Power%5Bx%2C2%5D%5C%2841%29+twice)

Now I can surely say that is good to go . Thanks for the great suggestions on this !

anutosh491:
@ThePauliPrinciple do you think something can be improved if any here or is it good to go ? I actually plan to introduce Denesting for `Add` through another pr once this is goes in . A seperate pr is better because It answers different problems i've spotted than this one (will open an issue addressing those soon) and the code requires more debugging and heavier tests than this one ! Hence I am looking forward to this going in .

anutosh491:
CC: @ThePauliPrinciple Does this look good now that I've attached references for all tests ? LGTM

anutosh491:
CC: @ThePauliPrinciple , does this look good to you ? Haven't heard your views on this since the last commit ! I am quite confident of the changes here and the tests added. I guess you had some doubts regarding the tests (as the issue requires extensive testing, the tests get a bit ginormous and hard to analyze here ). 
I had added links from wolfram for all tests in a link above for reference  and the convinience of all reviewers https://github.com/sympy/sympy/pull/22806#issuecomment-1039457376 (you can randomly cross check coeffs of some terms with their powers for your satisfaction) .
If you still have your doubts for this maybe @jksuom or @smichr could help us out here .

anutosh491:
Actually I just realized that `diofant` implements a quite simple version for these kind of tests(without denesting and couple other things), so practically the best way to maybe cross check these limits might be pretty printing from diofant's master (obviously the `Subs` part will miss denesting so not the best results but otherwise coeffs and powers can be compared)
Couple results go as follows
```
>>> # 2nd assertion test from test_denesting_for_subs
>>> pprint(u(x + 3*h, y + 3*h).diff(h, 3).simplify())
   /   3                    \|                  /   3                    \|                  //      3                    \|            \|                  //      3
   |  d                     ||                  |  d                     ||                  ||     d                     ||            ||                  ||     d
27*|------(u(xi_1, 3*h + y))||             + 27*|------(u(3*h + x, xi_2))||             + 81*||------------(u(xi_1, xi_2))||            ||             + 81*||--------
   |     3                  ||                  |     3                  ||                  ||           2               ||            ||                  ||     2
   \dxi_1                   /|xi_1=3*h + x      \dxi_2                   /|xi_2=3*h + y      \\dxi_2 dxi_1                /|xi_1=3*h + x/|xi_2=3*h + y      \\dxi_2  d

                   \|            \|
                   ||            ||
----(u(xi_1, xi_2))||            ||
                   ||            ||
xi_1               /|xi_2=3*h + y/|xi_1=3*h + x
>>>
>>>
>>> # 5th assertion test from test_denesting_for_subs
>>> pprint(u(x + 3*h, y - 2*h).series(h, x0=0, n=6).simplify())
                                                                               /  /   2              \|           /   2              \|            //      2
                                                                             2 |  |  d               ||           |  d               ||            ||     d
                                                                            h *|4*|------(u(x, xi_2))||       + 9*|------(u(xi_1, y))||       - 12*||-----------(u(xi_
                                                                               |  |     2            ||           |     2            ||            \\dxi_2 dxi_1
            /  /  d              \|           /  d              \|      \      \  \dxi_2             /|xi_2=y     \dxi_1             /|xi_1=x
u(x, y) - h*|2*|-----(u(x, xi_2))||       - 3*|-----(u(xi_1, y))||      | + ------------------------------------------------------------------------------------------
            \  \dxi_2            /|xi_2=y     \dxi_1            /|xi_1=x/                                                            2

         \|      \|      \      /    /   3              \|            /   3              \|            //      3                    \|      \|            //      3
         ||      ||      |    3 |    |  d               ||            |  d               ||            ||     d                     ||      ||            ||     d
1, xi_2))||      ||      |   h *|- 8*|------(u(x, xi_2))||       + 27*|------(u(xi_1, y))||       - 54*||------------(u(xi_1, xi_2))||      ||       + 36*||----------
         /|xi_1=x/|xi_2=y|      |    |     3            ||            |     3            ||            ||           2               ||      ||            ||     2
                         /      \    \dxi_2             /|xi_2=y      \dxi_1             /|xi_1=x      \\dxi_2 dxi_1                /|xi_1=x/|xi_2=y      \\dxi_2  dxi
-------------------------- + -----------------------------------------------------------------------------------------------------------------------------------------
                                                                                                                  6

                 \|      \|      \      /   /   4              \|            /   4              \|             //      4                    \|      \|             //
                 ||      ||      |    4 |   |  d               ||            |  d               ||             ||     d                     ||      ||             ||
--(u(xi_1, xi_2))||      ||      |   h *|16*|------(u(x, xi_2))||       + 81*|------(u(xi_1, y))||       - 216*||------------(u(xi_1, xi_2))||      ||       + 216*||-
                 ||      ||      |      |   |     4            ||            |     4            ||             ||           3               ||      ||             ||
_1               /|xi_2=y/|xi_1=x/      \   \dxi_2             /|xi_2=y      \dxi_1             /|xi_1=x       \\dxi_2 dxi_1                /|xi_1=x/|xi_2=y       \\d
---------------------------------- + ---------------------------------------------------------------------------------------------------------------------------------
                                                                                                                                                    24

      4                    \|      \|            //      4                    \|      \|      \      /     /   5              \|             /   5              \|
     d                     ||      ||            ||     d                     ||      ||      |    5 |     |  d               ||             |  d               ||
------------(u(xi_1, xi_2))||      ||       - 96*||------------(u(xi_1, xi_2))||      ||      |   h *|- 32*|------(u(x, xi_2))||       + 243*|------(u(xi_1, y))||
    2      2               ||      ||            ||     3                     ||      ||      |      |     |     5            ||             |     5            ||
xi_2  dxi_1                /|xi_2=y/|xi_1=x      \\dxi_2  dxi_1               /|xi_2=y/|xi_1=x/      \     \dxi_2             /|xi_2=y       \dxi_1             /|xi_1
----------------------------------------------------------------------------------------------- + --------------------------------------------------------------------


         //      5                    \|      \|              //       5                    \|      \|             //       5                    \|      \|
         ||     d                     ||      ||              ||      d                     ||      ||             ||      d                     ||      ||
   - 810*||------------(u(xi_1, xi_2))||      ||       + 1080*||-------------(u(xi_1, xi_2))||      ||       - 720*||-------------(u(xi_1, xi_2))||      ||       + 24
         ||           4               ||      ||              ||     2      3               ||      ||             ||     3      2               ||      ||
=x       \\dxi_2 dxi_1                /|xi_1=x/|xi_2=y        \\dxi_2  dxi_1                /|xi_2=y/|xi_1=x       \\dxi_2  dxi_1                /|xi_1=x/|xi_2=y
----------------------------------------------------------------------------------------------------------------------------------------------------------------------
                                                                       120

  //      5                    \|      \|      \
  ||     d                     ||      ||      |
0*||------------(u(xi_1, xi_2))||      ||      |
  ||     4                     ||      ||      |
  \\dxi_2  dxi_1               /|xi_2=y/|xi_1=x/    / 6\
------------------------------------------------ + O\h /
>>>
>>>
>>>
>>> x, y, z, a, h = symbols('x y z a h')
>>> # 1st assertion test from test_denesting_for_subs2
>>> pprint(u(x - 2*z*h, y - 2*z*h).diff(h, 3))
      //   3                       \|                  /   3                       \|                    //      3                    \|               \|
    3 ||  d                        ||                  |  d                        ||                    ||     d                     ||               ||
-8*z *||------(u(xi_1, -2*h*z + y))||                + |------(u(-2*h*z + x, xi_2))||                + 3*||------------(u(xi_1, xi_2))||               ||
      ||     3                     ||                  |     3                     ||                    ||           2               ||               ||
      \\dxi_1                      /|xi_1=-2*h*z + x   \dxi_2                      /|xi_2=-2*h*z + y     \\dxi_2 dxi_1                /|xi_1=-2*h*z + x/|xi_2=-2*h*z +

       //      3                    \|               \|               \
       ||     d                     ||               ||               |
   + 3*||------------(u(xi_1, xi_2))||               ||               |
       ||     2                     ||               ||               |
 y     \\dxi_2  dxi_1               /|xi_2=-2*h*z + y/|xi_1=-2*h*z + x/
>>>
>>> # 3rd assertion test from test_denesting_for_subs2
>>> pprint(u(x - 3*z*h, y + 3*z*h).series(h, 0, 4).simplify())
                                                                                    //   2              \|         /   2              \|           //      2
                                                                               2  2 ||  d               ||         |  d               ||           ||     d
                                                                            9*h *z *||------(u(x, xi_2))||       + |------(u(xi_1, y))||       - 2*||-----------(u(xi_
                                                                                    ||     2            ||         |     2            ||           \\dxi_2 dxi_1
                //  d              \|         /  d              \|      \           \\dxi_2             /|xi_2=y   \dxi_1             /|xi_1=x
u(x, y) + 3*h*z*||-----(u(x, xi_2))||       - |-----(u(xi_1, y))||      | + ------------------------------------------------------------------------------------------
                \\dxi_2            /|xi_2=y   \dxi_1            /|xi_1=x/                                                            2

         \|      \|      \           //   3              \|         /   3              \|           //      3                    \|      \|           //      3
         ||      ||      |      3  3 ||  d               ||         |  d               ||           ||     d                     ||      ||           ||     d
1, xi_2))||      ||      |   9*h *z *||------(u(x, xi_2))||       - |------(u(xi_1, y))||       + 3*||------------(u(xi_1, xi_2))||      ||       - 3*||------------(u
         /|xi_1=x/|xi_2=y|           ||     3            ||         |     3            ||           ||           2               ||      ||           ||     2
                         /           \\dxi_2             /|xi_2=y   \dxi_1             /|xi_1=x     \\dxi_2 dxi_1                /|xi_1=x/|xi_2=y     \\dxi_2  dxi_1
-------------------------- + -----------------------------------------------------------------------------------------------------------------------------------------
                                                                                                                2

             \|      \|      \
             ||      ||      |
(xi_1, xi_2))||      ||      |
             ||      ||      |
             /|xi_2=y/|xi_1=x/    / 4\
------------------------------ + O\h /
```
All results are perfectly in sync with the tests added (I have compared those which haven't been posted above too) . This is ready to go if no complaints can be spotted

anutosh491:
CC @oscarbenjamin , This is ready to go in , the only doubt here was that the test are quite extensive (as the pr solves a huge variety of wrong results) so I don't think @ThePauliPrinciple had the time to compare the results with their actual answers as he had written some days back 

>>The suggested changes seem like the right approach to me. The tests are extensive and cover the new functionality, however, I can't confirm myself that the proposed results for the tests are mathematically correct. Is there any external reference which shows what the result of differentiation should be for these type of expressions?

I had attached wolfram links to all tests some days before 
1) https://github.com/sympy/sympy/pull/22806#issuecomment-1039457376
2) Some more not so trivial and harder tests https://github.com/sympy/sympy/pull/22806#issuecomment-1040340521
3) The implementation somewhat works in diofant (not perfectly though but power and coeffs can be compared ) , hence I had attached pretty printed results from diofant too https://github.com/sympy/sympy/pull/22806#issuecomment-1050868621

Would be glad if you could maybe have a look over them and if you have doubt compare them with the wolfram links attached ! Though i am quite satisfied with the changes here .Let me know if there is something you would like to know from my side ( clarity on a test etc etc )

anutosh491:
ping @jksuom I'm quite confident with the changes here , There have been good reviews and extensive testing of the issues we are solving just because of the good number of functions we are catering too through the fix, would like to have your views here . I feel this is more than ready to go in !

anutosh491:
ping @eagleoflqj for review! I am completely satisfied with the changes here and have provided wolfram links for all added tests . I feel this is a really good change for the series module and should go in as it covered a good variety of taylor series expansions which would be fixed that aren't working currently

anutosh491:
ping @eagleoflqj for review! I am completely satisfied with the changes here and have provided wolfram links for all added tests . I feel this is a really good change for the series module and should go in as it covered a good variety of taylor series expansions which would be fixed that aren't working currently !
Could you please have a look and merge this .Thank you .

anutosh491:
> What is a simple example that demonstrates where this goes wrong on current master?

The tests I had added initially in the first commit were quite simple actually and like I wasn't aware the code had potential to address much more complex multivariate series expansions . So yeah, the tests were simple but the review thread by **ThePauliPrinciple** improved the range of cases being solved here by a vast amount , hence had to add tests for that. 
The size of tests grew because in the expansions , like not all but 1 or 2 or maybe 3 terms are faulty , the rest are correct .

The simplest thing being fixed here is 
```
>>> # On master
>>> (u(x -h, y -h).diff(h, 2))

⎛  2                ⎞│            ⎛  2                ⎞│              ⎛      2
⎜ ∂                 ⎟│            ⎜ ∂                 ⎟│              ⎜     ∂
⎜────(u(ξ₁, -h + y))⎟│          + ⎜────(u(-h + x, ξ₂))⎟│          + 2⋅⎜───────
⎜   2               ⎟│            ⎜   2               ⎟│              ⎝∂_-h +
⎝∂ξ₁                ⎠│ξ₁=-h + x   ⎝∂ξ₂                ⎠│ξ₂=-h + y

                     ⎞│
                     ⎟│
─────(u(_-h + x, ξ₂))⎟│
x ∂ξ₂                ⎠│ξ₂=-h + y
>>>
>>>
>>> # On Branch
>>> u(x - h, y - h).diff(h, 2)

⎛  2                ⎞│            ⎛  2                ⎞│              ⎛    2
⎜ ∂                 ⎟│            ⎜ ∂                 ⎟│              ⎜   ∂
⎜────(u(ξ₁, -h + y))⎟│          + ⎜────(u(-h + x, ξ₂))⎟│          + 2⋅⎜───────
⎜   2               ⎟│            ⎜   2               ⎟│              ⎝∂ξ₂ ∂ξ₁
⎝∂ξ₁                ⎠│ξ₁=-h + x   ⎝∂ξ₂                ⎠│ξ₂=-h + y

           ⎞│
           ⎟│
(u(ξ₁, ξ₂))⎟│
           ⎠│ξ₁=-h + x, ξ₂=-h + y
>>>
```

oscarbenjamin:
Oh, I see so the problem is that the first example here doesn't denest:
```python
In [3]: Subs(-Subs(f(x, y), x, z), y, t)
Out[3]: 
⎛-(f(x, y))│   ⎞│   
⎝          │x=z⎠│y=t

In [4]: Subs(Subs(f(x, y), x, z), y, t)
Out[4]: 
(f(x, y))│        
         │x=z, y=t
```
That's because the denesting code just checks to see if it has a Subs so it misses a Mul. Probably something more robust than the changes here are needed because a Subs could be deeply nested in an expression. On the other hand maybe it's okay not to denest all Subs all of the time.

In any case that in itself doesn't lead to a wrong result. Somewhere else there must be a bug that turns the nested Subs into an incorrect expression, probably using xreplace. For example this line creates an invalid expression although it doesn't seem to be used anywhere:
https://github.com/sympy/sympy/blob/e83dcff9c0519f81ef2dfe7b5630cc7a5918de78/sympy/core/function.py#L2244

It seems to me that this PR avoids the problem but doesn't quite hit where the actual bug is.

anutosh491:
> Oh, I see so the problem is that the first example here doesn't denest:
> 
> ```python
> In [3]: Subs(-Subs(f(x, y), x, z), y, t)
> Out[3]: 
> ⎛-(f(x, y))│   ⎞│   
> ⎝          │x=z⎠│y=t
> 



> That's because the denesting code just checks to see if it has a Subs so it misses a Mul. Probably something more robust than the changes here are needed because a Subs could be deeply nested in an expression. On the other hand maybe it's okay not to denest all Subs all of the time.

Yeah that's mostly it .. as you've pointed out here , we could actually do without the denest block as a whole as it would rather not give any errors but if we have a denest block we would need to have checks for all instances of Mul, Pow and Add eventually , just like I've added the `isinstance(expr, Mul)` check . 

I've only added one of them here , hence this can easily get all 2 variable cases now and get answers for something like these `u(x + 3*h, y - 2*h).series(h, x0=0, n=6)` that's because 2 variable cases won't need anything more than a `Mul` check but this code and the code on master would fail miserably to handle 3 variables as that would involve type `Add` which is more tougher to handle through such an implementation. 

Hence its not that I feel this is the only method to address the issue overall , but only keeping denesting for type `Subs` reduces the scope for multivariable expansion by a huge extent and won't be useful at all !

oscarbenjamin:
I still want to know where the actual bug is. Denesting or not isn't necessarily a bug but something is giving a wrong result somewhere.

PR #22808, opened 2022-01-06
Fixed stuff in rubi

oscargus:

I had a go at getting rubi up and running. While I didn't succeed, I thought I could at least contribute the corrections that were made.

#### Other comments

There is a function `InverseFunction` which I cannot seem to find in older versions. But that seems to be the only thing that is strictly missing to make it work (there may be bugs etc of course, but that is the only syntax error).

Also, some comparisons do not evaluate in the tests.

#### Release Notes

NO ENTRY


oscargus:
Since this is a no-brainer (and very few people care about `rubi`) I self-merge.

PR #22812, opened 2022-01-07
Set language of web pages to improve accessbility

### LINKED ISSUE #22811: Docs: Improve accessibility of current (Sphinx classic) theme by adding lang
Set the language to English to ensure visitors using screen readers have the text announced correctly. This is a simple improvement that can be made while we work on using a new theme in #22716.

From Lighthouse:

> If a page doesn't specify a lang attribute, a screen reader assumes that the page is in the default language that the user chose when setting up the screen reader. If the page isn't actually in the default language, then the screen reader might not announce the page's text correctly. Learn more.

bertiewooster:

Fixes #22811

Set `language = 'en'` in `doc/src/conf.py`

#### Release Notes

* other
  * Visitors to the documentation site, docs.sympy.org, using screen readers should have the text announced in the correct language.

### DIFF
--- a/doc/src/conf.py
+++ b/doc/src/conf.py
@@ -173,6 +173,7 @@ html_domain_indices = ['py-modindex']
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'SymPydoc'
 
+language = 'en'
 
 # Options for LaTeX output
 # ------------------------


asmeurer:
According to https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-language, this enables translation features. I'm +1 to this as long as it doesn't break any other parts of the docs build.

asmeurer:
By the way, in the future, it isn't necessary to open an issue if you are also going to open a pull request fixing the issue right away (it's fine to have a pull request that doesn't correspond to an issue).

bertiewooster:
> According to https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-language, this enables translation features.

Great--so this can help with multiple kinds of accessibility (vision impairment, and different languages).

>I'm +1 to this as long as it doesn't break any other parts of the docs build.

Build worked fine; [classic with lang = en result is online](http://whitegloveapps.com/sympy/classic/index.html). `diff` reveals that the only difference is adding `lang="en"` to `<html>` tag (other than the build date):

index_effect_of_lang


> By the way, in the future, it isn't necessary to open an issue if you are also going to open a pull request fixing the issue right away (it's fine to have a pull request that doesn't correspond to an issue).

Oh, OK. I thought we were supposed to have an issue number so we could start our branch with that number. Guess folks don't always follow that convention anyway. :-)

asmeurer:
Yeah the branch number thing is not required either. For me it makes it harder to checkout the branch locally, so I personally prefer to not have a number in there.

PR #22813, opened 2022-01-07
DO NOT MERGE: Changing Sphinx theme to Alabaster

bertiewooster:

Sphinx theme set to Alabaster per issue #22716

#### Release Notes

NO ENTRY

### DIFF
--- a/doc/src/conf.py
+++ b/doc/src/conf.py
@@ -142,7 +142,8 @@ html_static_path = ['_static']
 html_last_updated_fmt = '%b %d, %Y'
 
 # was classic
-html_theme = "classic"
+# html_theme = "classic"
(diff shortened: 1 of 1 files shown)


asmeurer:
For what it's worth, I discovered that Alabaster has a `fixed_sidebar` setting that lets the sidebar float with the content. That, along with some basic [custom CSS](https://stackoverflow.com/questions/57031848/sphinx-alabaster-theme-scroll-inside-of-fixed-sidebar/57040610) to make it scrollable, makes it a lot better than it seems. You still can't expand and collapse sections without visiting the pages. It does make me wonder just how hard it actually is to make the sidebar act like RTD or Furo in Alabaster or Classic.

PR #22817, opened 2022-01-07
geometry - projection method returns EmptySet when there is no projection

### LINKED ISSUE #22807: let some sort of EmptySet represent null object in geometry
Projections and intersections that are non-existant could be represented by EmptySet. Currently, projections raise an error and intersections return `[]`:

```python
>>> s
Segment2D(Point2D(0, 0), Point2D(0, 1))
>>> t
Segment2D(Point2D(1, 0), Point2D(2, 1/2))
>>> s.projection(t)
Segment2D(Point2D(0, 0), Point2D(0, 1/2))
>>> t.projection(s)
...
AttributeError: 'EmptySet' object has no attribute 'direction'

>>> s.intersection(t)
[]
```

Alternatively a geometric "Null" object -- not sure what it should be called, maybe GeometricEmptySet? -- could be created and used instead, so all geometric methods with it could return a null result (rather than giving EmptySet geometry related methods. e.g.
`GeometricEmptySet.projection(Line(Eq(x,0))) -> GeometricEmptySet`.

praneethratna:
Partially Fixes #22807 and now `projection` method returns EmptySet when there is no projection of one entity onto another.

#### Release Notes

* geometry
   * `projection` method of `line.py` now returns EmptySet when there is no projection of one entity onto another.

### DIFF
--- a/sympy/geometry/line.py
+++ b/sympy/geometry/line.py
@@ -28,7 +28,6 @@ from sympy.core.symbol import _symbol, Dummy, uniquely_named_symbol
 from sympy.core.sympify import sympify
 from sympy.functions.elementary.piecewise import Piecewise
 from sympy.functions.elementary.trigonometric import (_pi_coeff as pi_coeff, acos, tan, atan2)
-from sympy.sets import EmptySet
 from .entity import GeometryEntity, GeometrySet
 from .exceptions import GeometryError
 from .point import Point, Point3D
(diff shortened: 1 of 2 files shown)


praneethratna:
CC @smichr

smichr:
Looks good.

asmeurer:
Please write your release notes so that they make sense to an end-user reading the release notes file. "Made changes in `line.py`" is meaningless to someone who just wants to know what changed in the release. It's not even particularly helpful for people reviewing this PR.

praneethratna:
> Please write your release notes so that they make sense to an end-user reading the release notes file. "Made changes in `line.py`" is meaningless to someone who just wants to know what changed in the release. It's not even particularly helpful for people reviewing this PR.

Sorry for that vague release note. I've updated the Release notes now which looks much better to the end user.

PR #22818, opened 2022-01-08
adding option to skip explicitly solving for kinematic equations in KanesMethod

### LINKED ISSUE #22626: KanesMethod solves kinematics explicitly even if not asked
While dynamic equations are left in explicit form `M [u].diff(t) = f(q, u)`, the kinematic equations are solved explicitly [here](https://github.com/sympy/sympy/blob/0e4bab73562d6f90fcdf5fa079731f26a455347c/sympy/physics/mechanics/kane.py#L259):
```
            f_k = k_kqdot.LUsolve(f_k)
            k_ku = k_kqdot.LUsolve(k_ku)
            k_kqdot = eye(len(qdot))
```

The end result is that in `mass_matrix_full` the columns associated with kinematics are just the identity. This works fine most of the time, but can be unnecessarily slow if an explicit form of the equations is not desired, not to mention that it requires simplification of the explicit right hand side.

Would it make sense to avoid this step unless explicitly requested by the initialization function?

Happy to submit a PR if others here suggest how to best control that behavior.

zouhairm:
Fixes #22626 

This adds an option to KanesMethod to request that kinematics matrices are not explicitly inverted.

Default behavior should remain unchanged, so existing tests should be fine.
However, I think I have to write some tests to check that with the new behavior there aren't functions out there that are relying on the matrices being inverted (I have a suspicion that the linearization routines might :\ )

Before doing that I wanted to check that this is in the right direction

* physics.mechanics
  * Added option to leave kinematics in implicit form in KanesMethod


moorepants:
You need some unit tests that, at the minimum, exercises the new option and shows that the new results are expected.

I'd also like to see a problem that shows why this is needed. It could be the case that we need this behavior as default, because solving any of these systems symbolically has some price.

sympy/physics/mechanics/kane.py
```diff
@@ -118,7 +118,9 @@ class KanesMethod(_Methods):
     def __init__(self, frame, q_ind, u_ind, kd_eqs=None, q_dependent=None,
             configuration_constraints=None, u_dependent=None,
             velocity_constraints=None, acceleration_constraints=None,
-            u_auxiliary=None, bodies=None, forcelist=None):
+            u_auxiliary=None, bodies=None, forcelist=None,
+            explicit_kinematics=True
```
moorepants, line 122: You'll need to add an explanation of this and the new attributes to the docstring of KanesMethod.

sympy/physics/mechanics/kane.py
```diff
@@ -256,9 +258,10 @@ def _initialize_kindiffeq_matrices(self, kdeqs):
             k_ku = (msubs(kdeqs, qdot_zero) - f_k).jacobian(u)
             k_kqdot = (msubs(kdeqs, u_zero) - f_k).jacobian(qdot)
 
-            f_k = k_kqdot.LUsolve(f_k)
-            k_ku = k_kqdot.LUsolve(k_ku)
-            k_kqdot = eye(len(qdot))
+            if explicit_kinematics:
+                f_k = k_kqdot.LUsolve(f_k)
+                k_ku = k_kqdot.LUsolve(k_ku)
+                k_kqdot = eye(len(qdot))
```
moorepants, line 264: I'm looking at this more closely and, at the moment, I do not see what the purpose of these original lines are.

moorepants: I guess it is simply to force the block in the full mass matrix to be identity.

moorepants:
The rolling disc examples in the existing tests have kinematical differential equations that are non-trivial, so you could use that to show the difference in the option.

sympy/physics/mechanics/kane.py
zouhairm: will do - but hopefully it can be made moot if we agree that the =False behavior should be the default behavior?

moorepants:
I've merged the PR with some code clean up: https://github.com/sympy/sympy/pull/22823. Can you update this to build on that?

moorepants:
@zouhairm This is still on my radar. I'm just swamped teaching my new multibody dyanmics course. I'm actually about to teach about this, so maybe I can manage a proper review when I do so. I do want to fix this.

zouhairm:
Hi @moorepants - just pinging to see if you've had a chance to check the PR.
Would love to be able to close this out in the next week or so :)

moorepants:
Thanks for checking in. It is on my todo, just haven't gotten to it. Let me see if I can have a proper look this weekend. I may have a bit of time.

sympy/physics/mechanics/kane.py
```diff
@@ -40,13 +42,17 @@ class KanesMethod(_Methods):
...
-        The system's forcing vector
+        The system's forcing vector for the u's
     mass_matrix_full : Matrix
         The "mass matrix" for the u's and q's
     forcing_full : Matrix
         The "forcing vector" for the u's and q's
+    mass_matrix_kin : Matrix
+        The "mass matrix" for the q's
+    forcing_kin : Matrix
+        The "forcing vector" for the q's
```
moorepants, line 55: It'd be nice to come up with better names for this. They are kinematical differential coefficient and remainder matrices. I'm not sure what a nice concise, informative name would be though.

sympy/physics/mechanics/kane.py
```diff
@@ -244,7 +255,8 @@ def _initialize_kindiffeq_matrices(self, kdeqs):
             Kinematic differential equations in the form of f(u,q',q,t) where
             f() = 0. The equations have to be linear in the generalized
             coordinates and generalized speeds.
-
+        explicit_kinematics: bool
+            If True, force the kinematic equations to be in explicit form
```
moorepants, line 259: The qdots are still solved for, so it isn't clear how this forces things to an explicit form.

sympy/physics/mechanics/kane.py
```diff
@@ -303,6 +323,26 @@ def _initialize_kindiffeq_matrices(self, kdeqs):
...
+        Return the partial velocity w.r.t. generalized speeds
+        Throws an error if qdot terms are left
+        '''
+        self._assert_no_qdot(vel_list)
+        return partial_velocity(vel_list, self.u, self._inertial)
+
+    def _assert_no_qdot(self, vel_or_acc_list):
+        q_dot_set = set(self._qdot).union([_qd.diff() for _qd in self._qdot])
+        for vel in vel_or_acc_list:
+            dyn_symbols = find_dynamicsymbols(vel, reference_frame=self._inertial)
```
moorepants, line 337: This could be a very slow operation for very large systems. I'm not a fan of adding this. There must be a faster way to check if qdots are present. I bet differentiating wrt to the qdots is faster than this.

sympy/physics/mechanics/kane.py
```diff
@@ -581,7 +622,10 @@ def kanes_equations(self, bodies=None, loads=None):
                 km = KanesMethod(self._inertial, self.q, self._uaux,
                         u_auxiliary=self._uaux, u_dependent=self._udep,
                         velocity_constraints=(self._k_nh * self.u +
-                        self._f_nh))
+                        self._f_nh),
+                        acceleration_constraints=(self._k_dnh * self._udot +
```
moorepants, line 626: Not clear why this is added.

sympy/physics/mechanics/kane.py
moorepants: Do we really need this kwarg? You can provide these matrices without adding a kwarg here. Right?

sympy/physics/mechanics/kane.py
```diff
@@ -286,13 +298,21 @@ def _initialize_kindiffeq_matrices(self, kdeqs):
...
+                # implicit form.
+                f_k = k_kqdot.LUsolve(f_k)
+                k_ku = k_kqdot.LUsolve(k_ku)
+                k_kqdot = eye(len(qdot))
+                self._qdot_u_map = dict(zip(qdot, -(k_ku*u + f_k)))
+            else:
+                # If kinematics are not forced to be explicit, do not provide a qdot_u_map
+                # However, still compute a kindiffdict that is necessary to provide a RHS
+                # (i.e. fully explicit) version of the equations
+                self._qdot_u_map = dict()
```
moorepants, line 313: If you don't calculate this here, then want the equations of motion be incorrectly formed. The qdots have to be replaced by expressions of the u's before partial velocities are computed, no?

sympy/physics/mechanics/kane.py
zouhairm: it's fair that there is really no "mass" associated with these equations, but I kept the name to  be consistent with the `mass_matrix_full` (which has as subset of it these equations)

Happy to use something else, but I'll defer to you for a better name

sympy/physics/mechanics/kane.py
zouhairm: gotcha - I assume it's the `find_dynamicssymbols` that you expect to be slow? I'll see what I can do with using diff() instead

sympy/physics/mechanics/kane.py
zouhairm: if this is not done, then the 2nd call to KanesMethod ends up having to recompute the acceleration constraint. While in principle that's fine, it's an issue if acceleration_constraints were passed to the initial call: the initial call uses line 228 to compute `_f_dnh` and `_k_dnh` from `acc`, while the 2nd call uses line 220 to compute them from `vel` 
By passing `acceleration_constraints = self._k_dnh * self._udot +self._f_dnh`, it forces the 2nd call to use 228 which will just recover `_f_dnh` and `_k_dnh` which have been computed in the initial call regardless of whether acceleration constraints was passed or not.

Why is this relevant to this explicit kinematics PR? acceleration_constraints might need to be passed to ensure that no qdots appear in them, and not rely on KanesMethod computing them from derivatives of velocity constraints which might introduce qdots.

sympy/physics/mechanics/kane.py
zouhairm: I think we do need it: the behavior is different if one chooses to not solve for kinematics explicitly, and the method is limited to forces/velocities/accelerations only containing generalized speeds, not rates of generalized coordinates.

sympy/physics/mechanics/kane.py
zouhairm: I tried this with using diff and jacobian, and found actually that it was slower than using find_dynamicssymbols (10ms vs 1ms tested on the disc linearization test).
I left it in the code for now just to show what it looks like. Let me know if you have suggestions there for something faster, otherwise I think find_dynamicssymbols is sufficiently fast?

zouhairm: It currently fires for both cases.


I don't have a good handle for the complexity of `find_dynamicssymbols`: in the project I've used it with 30+ states I haven't encountered issues. My approach would be to keep the logic as is and only address it if someone raises an issue, but if your intuition says this will be slow for uses cases you know people might be interested in, then it's easy enough to shortcut it by returning immediately if `self.explicit_kinematics == True`.

I'm fine with either options, let me know what you prefer and I'll push one or the other.

zouhairm:
@moorepants : do you know how I fix the authors test (do I have to since it's not "required" for merging?

When I run the test locally as instructed by the [wiki](https://github.com/sympy/sympy/wiki/Development-workflow#add-your-name-and-email-address-to-the-mailmap-file) there is no issue

```
sympy 22626_kanesmethod_implici $ bin/mailmap_check.py
No changes needed in .mailmap

the next SymPy release.

	Andrii Oriekhov <andriyorehov@gmail.com>
	Arie Bovenberg <a.c.bovenberg@gmail.com>
	Carson McManus <carson.mcmanus1@gmail.com>
	Gaurav Jain <gjain369@gmail.com>
	Luis Talavera <luisfertalavera15@gmail.com>
	Shivam Sagar <technoshivam12@gmail.com>
	Sumit Kumar <mr.sumitkrr@gmail.com>
	Tirthankar Mazumder <63574588+wermos@users.noreply.github.com>
```

sympy/physics/mechanics/kane.py
moorepants: The place were I found find_dynamicsymbols to be slow was using it on this problem https://pydy.readthedocs.io/en/latest/examples/carvallo-whipple.html. For example it takes 1.6 seconds to run it on fr + frstar on my computer. Running it on the n_link_pendulum_on_cart model in the mechanics package here with n=2 is like a 1/10 to 1/20th of a second, so that isn't bad. I'm just wary of introducing find_dynamicsymbols into the core code because we typically know what symbols we need a priori.

sympy/physics/mechanics/tests/test_kane3.py
```diff
@@ -1,4 +1,4 @@
-from sympy.core.evalf import evalf
+import sympy.core.evalf as evalf
```
moorepants, line 1: This seems to import the module instead of the function.

zouhairm: yes the module is needed because of like 189.
```
    HTA                             =  evalf.N(pi / 2 - pi / 10)
```
 Before it imported just the function so that didnt work. I think it doesn't fail on the server because this test is skipped on travis

moorepants:
I had some time to look at this more closely yesterday finally. I read through your example here: https://github.com/sympy/sympy/issues/22626#issuecomment-1012617282. That example is for a single rigid body. In one sense, there is little reason to use Kane's method for a single unconstrained rigid body, because you can write the equations of motion directly. But it would be nice if `KanesMethod` produced the cleaner implicit equations you show for this specific case. And it is worth noting that this is a very specific case: single rigid body with ignorable orientation and position coordinates.

For the single rigid body and six degrees of freedom in your example the velocities can be expressed cleanly in the body fixed reference frame with generalized speeds being the measure numbers of those velocities. When the time derivatives of the momenta are then calculated you can avoid introducing any new time derivatives of the generalized coordinates because this system has ignorable orientation coordinates. But for multibody systems with only a bit more complexity (in particular non-ignorable coordinates), time derivatives of the generalized coordinates will arise when differentiating velocity vectors even if those velocity vectors are written only in terms of generalized speeds. 

There are two aspects that I think are separable:

1. Do we need to solve the kinematical differential equations for the qdots for the purposes of substituting them into velocities before finding the relevant partial velocities and substituting in accelerations to put the equations pf motion in terms of udots, u's, and q's?

and

2. Do we provide the kinematical differential equations to the user in implicit form in some way?

For 1. if there are qdots in the essential velocities, then substituting them in and then finding the partial velocities is the basic approach. These will also have to be substituted into the essential accelerations so that these are in terms of udots and u's. If the user has already defined the essential velocities and accelerations in terms of udots, u's, and q's, then no substitution needs to occur. Right now, I think we do the substitutions for the user inside `KanesMethod` if they didn't manually do this.

For 2. we should be able to provide this form regardless of whether we've substituted qdots for the user automatically or not. I think it is bit of a bad design in `KanesMethod` to formulate the kinematical differential equations as the explicit form `q' = f(u, q, t)` and present them when `mass_matrix_full` and `forcing_full` is required because it does make the equations more complex than they need to be. I think we can simply have a flag that gives `Y_q*q' = Y_u*u + f(q, t)` instead (like you've done).

Right now this PR doing both 1 and 2 and I think for your examples you've shown, that you personally only need a flag that deals with 2. There is no reason to disable the internal solving for the qdots and substitution such that the equations won't be solved (error thrown), because they can be solved.

I also think this flag may belong more in `.kanes_equation(explicit_kinematical=True)` which would cause `mass_matrix_full` and `forcing_full` to take the different form. Even better would be that `mass_matrix_full()` is a method (we can't do that due to backwards incompat, but we could make a new method `mass_matrix_full_implicit()` or something).

About the tests modifications. I'm not that enthusiastic about changing all the tests to these for loop of two items. It makes the tests messier and harder to understand. I'd rather see a couple of completely new tests that test this flag or simply calling `.kanes_equations()` twice at the end of the problem, once with the flag and once without.

Do you agree that points 1 and 2 above are completely separable? If so, my suggestion would be then to simply:

- Make `Y_q` and `Y_u*u + f(q, t)` accessible to the user as methods or properties of `KanesMethod`.
- Add a way to return `mass_matrix_full_implicit` and `forcing_full_implicit` (new methods/properties is probably best)
- Don't make the unit tests more complex than they need to be. Simply showing that `mass_matrix_full_implicit` works for a single multibody system is probably sufficient. Testing your single rigid body case example is also very appropriate because you want that form for the single body to work as you show.

Lastly, there are some things we could do for point 1. that could make things simpler, faster, and smarter inside `KanesMethod` but I think that should be tackled in another PR and would take some significant work.

zouhairm:
Thanks for taking the time to dig deeper into this. Greatly appreciated!

I understand the argument of splitting the "2 tasks" and only focusing on providing implicit kinematics equations (point 2) and leaving the handling of qdots (point 1) to a separate PR.

I agree that I can accomplish what I'm after by ensuring that the velocities are all functions of generalized speeds (that's what I'm already doing) and I can ensure no substitution of qdots will occur by doing my `assert_no_qdots` outside of KanesMethod (thus not burdening the class with more complexity). Then as long as I have the ability to return an `mass_matrix_full_implicit`, I should be fine.

Regarding tests: things would have been cleaner if I could use [pytest parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html) but looks like the repo doesnt use the full pytest module. 
Nonetheless, if I narrow the focus of the PR to point-2 then it doesnt make sense to test things the way I have so far. I can just indeed add tests which focus on this.

It'll take me a few days to refactor things, but hopefully it'll be an overall simpler PR ...


P.S. While the example I shared indeed is a single rigid body, it was meant as "minimum" illustration of the issue. In my use case, that body is the "root" body of a 13 bodies problem. And although the other 12 bodies have more "trivial" generalized coordinates/speeds, forces (for example gravity) do end up being a function of the orientation of the root body: i.e. the latter is not an "ignorable coordinate" (and it gets coupled with position due to elliptical Earth where local vertical is a function of position)

moorepants:
> Regarding tests: things would have been cleaner if I could use [pytest parametrize](https://docs.pytest.org/en/6.2.x/parametrize.html) but looks like the repo doesnt use the full pytest module.

SymPy uses an old fork of pytest. There is some effort to change that.

> It'll take me a few days to refactor things, but hopefully it'll be an overall simpler PR ...

Sorry for requesting a large change. The effort here is apparent and it is very nice work. Thanks for you willingness to refactor.

> P.S. While the example I shared indeed is a single rigid body, it was meant as "minimum" illustration of the issue. In my use case, that body is the "root" body of a 13 bodies problem.

Got it. Sounds like a cool problem!

> @moorepants : do you know how I fix the authors test (do I have to since it's not "required" for merging?

I'm not real sure about this as it is a new requirement.

moorepants:
This is the email zouhair <430175+zouhairm@users.noreply.github.com> that shows up in the CI error. Maybe you have commits with different emails associated with them. If so, the mailmap file somehow associates multiple emails with one person.

oscarbenjamin:
> This is the email zouhair [430175+zouhairm@users.noreply.github.com](mailto:430175+zouhairm@users.noreply.github.com) that shows up in the CI error.

I think what happens is that when running CI GitHub makes a merge commit and based on your GitHub profile settings it chooses to use this no-reply email address (I think because your account is set to not share your email address).

You can either change your GitHub settings or add a line to .mailmap to associate this email address with you. You are already in the .mailmap file as
```
Zouhair <zouhair.mahboubi@gmail.com>
```
which is what is shown in your commits. If you change that line to
```
Zouhair <zouhair.mahboubi@gmail.com> zouhair <430175+zouhairm@users.noreply.github.com>
```
then any commits made by GitHub will also be associated with the same name and email.

zouhairm:
Closing in favor of https://github.com/sympy/sympy/pull/23580 as per discussion

PR #22829, opened 2022-01-10
do not use None as key in local_dict

### LINKED ISSUE #22822: `parse_expr` evaluates `is None` expression to False since 1.9
Since version 1.9 `parse_expr`  evaluates a `is None` expression to False instead of True (as I would expect) when logic_data passed contains a  a None value.

```python
from sympy.parsing.sympy_parser import parse_expr

logic = "some_parameter is None"
data = {'some_parameter': None}
print(parse_expr(logic, data))  # evaluates to True in 1.8 and False in 1.9
data = {'some_parameter': 'got it'}
print(parse_expr(logic, data))  # evaluates to False
```

Expected result is True, False and not False, False

smichr:

fixes #22822
#### Release Notes

* parsing
    * corrected parsing error that happened when when local_dict value was None

### DIFF
--- a/sympy/parsing/sympy_parser.py
+++ b/sympy/parsing/sympy_parser.py
@@ -20,6 +20,8 @@ from sympy.utilities.misc import filldedent, func_name
 from sympy.functions.elementary.miscellaneous import Max, Min
 
 
+null = ''
+
 def _token_splittable(token):
     """
     Predicate for whether a token name can be split into multiple tokens.
@@ -559,11 +561,11 @@ def auto_symbol(tokens, local_dict, global_dict):
                     or (prevTok[0] == OP and prevTok[1] in ('(', ',')
                         and nextTokNum == OP and nextTokVal == '=')
                     # the name has already been defined
-                    or name in local_dict and local_dict[name] is not None):
+                    or name in local_dict and local_dict[name] is not null):
                 result.append((NAME, name))
                 continue
             elif name in local_dict:
-                local_dict.setdefault(None, set()).add(name)
+                local_dict.setdefault(null, set()).add(name)
                 if nextTokVal == '(':
                     local_dict[name] = Function(name)
                 else:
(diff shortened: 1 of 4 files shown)


oscargus:
Closing and reopening as the required tests have changed.

sympy/parsing/sympy_parser.py
```diff
@@ -20,6 +20,8 @@
 from sympy.functions.elementary.miscellaneous import Max, Min
 
 
+null = ''
```
ThePauliPrinciple, line 23: It feels to me that using an empty string is going to make this vulnerable to (hard to debug) bugs in the future. (when for some reason `''` ends up as a key for another reason)

```python
class ParserNull:
    pass
null = ParserNull()
```
Might be a bit safer, but feel free to ignore if you don't think it is needed here.

smichr: I understand the sentiment...but when would we ever parse a null string which, by definition, has nothing to parse? There is also a test that `null` does not appear in `local_dict` so if it ever does we could revisit it then and make the change in one place (with what you suggested). I'll make the change, however.

smichr, line 23:
```suggestion
class ParserNull:
    pass
null = ParserNull()
```

sympy/parsing/sympy_parser.py
```diff
@@ -1046,6 +1050,8 @@ def parse_expr(s, local_dict=None, transformations=standard_transformations,
         local_dict = {}
     elif not isinstance(local_dict, dict):
         raise TypeError('expecting local_dict to be a dict')
+    elif null in local_dict:
+        raise ValueError('cannot use "" in local_dict')
```
smichr, line 1054:
```suggestion
        raise ValueError('cannot use ParserNull as local_dict key')
```

sympy/parsing/sympy_parser.py
smichr: On second thought, let's leave it as is -- by definition nobody should be changing nothing to something.

PR #22830, opened 2022-01-10
Use doc requirements.txt for sphinx test

oscargus:

Use `doc/requirements.txt` instead of `release/requirements.txt` for the sphinx tests. 

#### Release Notes

NO ENTRY

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -272,7 +272,7 @@ jobs:
         with:
           python-version: '3.9'
       - run: release/aptinstall.sh
-      - run: pip install -r release/requirements.txt
+      - run: pip install -r doc/requirements.txt
       - run: bin/test_sphinx.sh
 
 


asmeurer:
This looks good. Maybe we can also remove the pins in the file while we are at it (this was mentioned on some other issues). 

Once we do split out the "release" requirements files we should also split out the apt packages to a separate doc/apt-requirements.txt that gets used here as well. Right now they are hard-coded in the aptinstall.sh file.

oscargus:
There is another PR that has done that. I didn't want to risk merge conflicts for that. (I think it looks good otherwise, but includes parts I do not really know much about.)

skieffer:
> There is another PR that has done that.

Do you mean #22801? I think this one's really just waiting for an "OK" from someone. It's a follow-on to #22658, now allowing use of the `make livehtml` live server.

It had passed all tests before, but I guess something has changed in the testing system since then? Should I close and re-open it, to make the tests run again?

asmeurer:
Oh I didn't know there was already a PR for that. We can merge this separately then.

oscargus:
> Do you mean #22801?

Yes, that's the one. As I said, I do not really follow the other parts happening there, but the requirements-part seems fine to me.

PR #22831, opened 2022-01-10
combinatorial - added code for Schröder Number

praneethratna:
Added code for Schröder number in combinatorial numbers.

#### Other comments
There are other combinatorial numbers which can be added and might be useful to sympy
- [ ] [Narayana Number](https://en.wikipedia.org/wiki/Narayana_number)
- [ ] [Lobb Number](https://en.wikipedia.org/wiki/Lobb_number)
- [ ] [Delannoy Number](https://en.wikipedia.org/wiki/Delannoy_number)

I'll add these along with this PR or in a separate PR after this is merged.

#### Release Notes

* functions
  * Added Schröder number in combinatorial numbers.

### DIFF
--- a/sympy/core/tests/test_args.py
+++ b/sympy/core/tests/test_args.py
@@ -1995,7 +1995,7 @@ def test_sympy__functions__combinatorial__numbers__motzkin():
 
 def test_sympy__functions__combinatorial__numbers__schroeder():
     from sympy.functions.combinatorial.numbers import schroeder
-    assert _test_args(schroeder(5))
+    assert _test_args(schroeder(x))
 
 
 def test_sympy__functions__combinatorial__numbers__fibonacci():
--- a/sympy/core/tests/test_args.py
+++ b/sympy/core/tests/test_args.py
@@ -1993,6 +1993,11 @@ def test_sympy__functions__combinatorial__numbers__motzkin():
     assert _test_args(motzkin(5))
 
 
+def test_sympy_functions_combinatorial_numbers_schroder():
+    from sympy.functions.combinatorial.numbers import schroder
+    assert _test_args(schroder(6))
+
+
 def test_sympy__functions__combinatorial__numbers__fibonacci():
     from sympy.functions.combinatorial.numbers import fibonacci
     assert _test_args(fibonacci(x))
--- a/sympy/core/tests/test_args.py
+++ b/sympy/core/tests/test_args.py
@@ -1993,9 +1993,9 @@ def test_sympy__functions__combinatorial__numbers__motzkin():
     assert _test_args(motzkin(5))
 
 
-def test_sympy_functions_combinatorial_numbers_schroder():
-    from sympy.functions.combinatorial.numbers import schroder
-    assert _test_args(schroder(6))
+def test_sympy__functions__combinatorial__numbers__schroeder():
+    from sympy.functions.combinatorial.numbers import schroeder
+    assert _test_args(schroeder(5))
 
 
 def test_sympy__functions__combinatorial__numbers__fibonacci():
--- a/sympy/functions/combinatorial/numbers.py
+++ b/sympy/functions/combinatorial/numbers.py
@@ -17,7 +17,7 @@ from sympy.core.function import Function, expand_mul
 from sympy.core.logic import fuzzy_not
 from sympy.core.mul import Mul, prod
 from sympy.core.numbers import E, pi, oo, Rational, Integer
-from sympy.core.relational import Eq, is_le, is_gt
+from sympy.core.relational import is_le, is_gt
 from sympy.external.gmpy import SYMPY_INTS
 from sympy.functions.combinatorial.factorials import (binomial,
     factorial, subfactorial)
@@ -2184,17 +2184,19 @@ class motzkin(Function):
 
 class schroeder(Function):
     r"""
-    In mathematics, the Schroeder number, also called a
-    large Schroeder number or big Schroeder number, describes
-    the number of lattice paths from the southwest corner
-    $(0, 0)$ of an $n\times n$ grid to the northeast corner $(n, n)$ using only
+    In mathematics, the Schroeder number,
+    also called a large Schroeder number or
+    big Schroeder number, describes the number of
+    lattice paths from the southwest corner $(0, 0)$
+    of an $n\times n$ grid to the northeast corner $(n, n)$ using only
     single steps north, $(0, 1)$ northeast, $(1, 1)$ or east, $(1, 0)$
-    that do not rise above the SW-NE diagonal. The nth Schroeder number
-    is the number of Schroeder paths of length n. The Schroeder numbers were
-    named after the German mathematician Ernst Schroeder.
+    that do not rise above the SW-NE diagonal. The nth
+    Schroeder number is the number of Schroeder paths
+    of length n. The Schroeder numbers  were named
+    after the German mathematician Ernst Schroeder.
 
     Schroeder numbers are the integer sequences defined by the
-    initial terms `S_0 = 1`, `S_1 = 2` and which satisfies the two-term recurrence relation
+    initial terms `S_0 = 1`, `S_1 = 2` and the two-term recurrence relation
     `S_n = \frac{6*n - 3}{n + 1} * S_{n-1} - \frac{n - 2}{n + 1} * S_{n-2}`.
 
 
@@ -2203,12 +2205,16 @@ class schroeder(Function):
 
     >>> from sympy import schroeder
 
+    >>> schroeder(2)
+    2
+    >>> schroeder(5)
+    90
     >>> schroeder.is_schroeder(2)
     True
     >>> schroeder.is_schroeder(7)
     False
     >>> schroeder.find_first_n_schroeder_numbers(1)
-    [1, 2]
+    [1]
     >>> schroeder.find_first_n_schroeder_numbers(10)
     [1, 2, 6, 22, 90, 394, 1806, 8558, 41586, 206098]
     >>> schroeder.find_schroeder_numbers_in_range(1, 10)
@@ -2231,7 +2237,7 @@ class schroeder(Function):
             n = as_int(n)
         except ValueError:
             False
-        if n>0:
+        if n > 0:
             if n in (1, 2):
                 return True
 
@@ -2239,15 +2245,12 @@ class schroeder(Function):
             sn = 2
             i = 2
             while sn < n:
-                b = ((6*i - 3)*sn - (i - 2)*sn1)/(i + 1)
+                b = ((6*i - 3)*sn - (i - 2)*sn1) // (i + 1)
                 sn1 = sn
                 sn = b
                 i = i+1
 
-            if Eq(sn, n, evaluate=True):
-                return True
-            else:
-                return False
+            return sn == n
 
         else:
             return False
@@ -2260,15 +2263,15 @@ class schroeder(Function):
             raise ValueError('The provided number must be a positive integer')
         if n < 0:
             raise ValueError('The provided number must be a positive integer')
-        schroeder = [1]
-        if n >= 1:
-            schroeder.append(2)
+        if n == 1:
+            return [1]
+        schroeder = [1, 2]
         sn1 = 2
         sn = 6
         i = 3
         while i <= n:
             schroeder.append(sn)
-            b = ((6*i - 3)*sn - (i - 2)*sn1)/(i + 1)
+            b = ((6*i - 3)*sn - (i - 2)*sn1) // (i + 1)
             sn1 = sn
             sn = int(b)
             i = i+1
@@ -2277,9 +2280,17 @@ class schroeder(Function):
 
     @staticmethod
     def find_schroeder_numbers_in_range(x, y):
-        if 0 <= x <= y:
-            schroeder = list()
-            if x <=1 <= y:
+        """
+        Returns all the schroeder numbers in the range (x, y),
+        x and y inclusive.
+
+        """
+
+        if x < 0 or y < x:
+            raise ValueError('The provided range is not valid. This condition should satisfy x <= y')
+        else:
+            schroeder = []
+            if x <= 1 <= y:
                 schroeder.append(1)
             sn1 = 1
             sn = 2
@@ -2287,15 +2298,12 @@ class schroeder(Function):
             while sn <= y:
                 if sn >= x:
                     schroeder.append(sn)
-                b = ((6*i - 3)*sn - (i - 2)*sn1)/(i + 1)
+                b = ((6*i - 3)*sn - (i - 2)*sn1) // (i + 1)
                 sn1 = sn
                 sn = int(b)
                 i = i+1
 
             return schroeder
-
-        else:
-            raise ValueError('The provided range is not valid. This condition should satisfy x <= y')
 
     @staticmethod
     @recurrence_memo([S.One, 2*S.One])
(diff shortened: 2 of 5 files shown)


ThePauliPrinciple:
Why is this a subclass of `Function`?

In German, it is allowed to replace "ö" with "oe" when it is not possible to write "ö" (even in legal documents), so I think it is more appropriate to name the function "schroeder" (one of the links in the docstring makes use of this), but not sure if SymPy has its own convention for this situation.

praneethratna:
> Why is this a subclass of `Function`?

Since it also used a function to return the nth schroeder number.
```
>>> schroeder(5)
90
```

> 
> In German, it is allowed to replace "ö" with "oe" when it is not possible to write "ö" (even in legal documents), so I think it is more appropriate to name the function "schroeder" (one of the links in the docstring makes use of this), but not sure if SymPy has its own convention for this situation.

I'm also not sure of the convention so I've changed the function name to `schroeder`.

praneethratna:
> Gröbner is called `groebner` so replacing ö with oe is probably a good idea (and feel free to use unicode in the docstring, although it may require whitelisitng the file somewhere.

I have changed the name to `schroeder'.

anutosh491:
@praneethratna Heyy nice addition. Combinatorics has been always been one of favorites in math but I was not quite sure of how could I help sympy's combinatorics module , hence I would maybe like to try adding something from the task's you mentioned ! @oscargus do you have any suggestions on addition something which would be a good addition to sympy

praneethratna:
ping @oscarbenjamin for review

praneethratna:
ping @oscarbenjamin for review.

sympy/core/tests/test_args.py
```diff
@@ -1993,6 +1993,11 @@ def test_sympy__functions__combinatorial__numbers__motzkin():
     assert _test_args(motzkin(5))
 
 
+def test_sympy__functions__combinatorial__numbers__schroeder():
+    from sympy.functions.combinatorial.numbers import schroeder
+    assert _test_args(schroeder(5))
```
oscarbenjamin, line 1998: Does `schroeder(5)` return an integer? If so then this should be changed to something like `schroeder(x)`.

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,144 @@ def eval(cls, n):
...
+    @recurrence_memo([S.One, 2*S.One])
+    def _schroeder(n, prev):
+        return ((6*n - 3)*prev[-1] - (n - 2)*prev[-2]) // (n + 1)
+
+    @classmethod
+    def eval(cls, n):
+        try:
+            n = as_int(n)
+        except ValueError:
+            raise ValueError('The provided number must be a positive integer')
```
oscarbenjamin, line 2310: There's no point having a symbolic class (subclass of Function) if it only accepts concrete integers. It should be possible to do `schroeder(x)` for a symbol `x`.

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,144 @@ def eval(cls, n):
...
+    `S_n = \frac{6*n - 3}{n + 1} * S_{n-1} - \frac{n - 2}{n + 1} * S_{n-2}`.
+
+
+    Examples
+    ========
+
+    >>> from sympy import schroeder
+
+    >>> schroeder.is_schroeder(2)
+    True
```
oscarbenjamin, line 2207: There should be an example showing that you can call e.g. `schroeder(5)`

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,144 @@ def eval(cls, n):
...
+                    schroeder.append(sn)
+                b = ((6*i - 3)*sn - (i - 2)*sn1)/(i + 1)
+                sn1 = sn
+                sn = int(b)
+                i = i+1
+
+            return schroeder
+
+        else:
+            raise ValueError('The provided range is not valid. This condition should satisfy x <= y')
```
oscarbenjamin, line 2298: Better to move this raise up to the top of the function where the condition is tested

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,144 @@ def eval(cls, n):
...
+            sn = int(b)
+            i = i+1
+
+        return schroeder
+
+    @staticmethod
+    def find_schroeder_numbers_in_range(x, y):
+        if 0 <= x <= y:
+            schroeder = list()
+            if x <=1 <= y:
```
oscarbenjamin, line 2282: Spacing is inconsistent

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,144 @@ def eval(cls, n):
...
+    def find_first_n_schroeder_numbers(n):
+        try:
+            n = as_int(n)
+        except ValueError:
+            raise ValueError('The provided number must be a positive integer')
+        if n < 0:
+            raise ValueError('The provided number must be a positive integer')
+        schroeder = [1]
+        if n >= 1:
+            schroeder.append(2)
```
oscarbenjamin, line 2265: Maybe just:
```
if n == 1:
    return [1]

schroeder = [1, 2]
```

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,144 @@ def eval(cls, n):
...
+            b = ((6*i - 3)*sn - (i - 2)*sn1)/(i + 1)
+            sn1 = sn
+            sn = int(b)
+            i = i+1
+
+        return schroeder
+
+    @staticmethod
+    def find_schroeder_numbers_in_range(x, y):
+        if 0 <= x <= y:
```
oscarbenjamin, line 2280: This needs a docstring that defines what it returns. Is the range inclusive?

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,144 @@ def eval(cls, n):
...
+        if n >= 1:
+            schroeder.append(2)
+        sn1 = 2
+        sn = 6
+        i = 3
+        while i <= n:
+            schroeder.append(sn)
+            b = ((6*i - 3)*sn - (i - 2)*sn1)/(i + 1)
+            sn1 = sn
+            sn = int(b)
```
oscarbenjamin, line 2273: Better to use integer division `//`

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,144 @@ def eval(cls, n):
...
+            while sn < n:
+                b = ((6*i - 3)*sn - (i - 2)*sn1)/(i + 1)
+                sn1 = sn
+                sn = b
+                i = i+1
+
+            if Eq(sn, n, evaluate=True):
+                return True
+            else:
+                return False
```
oscarbenjamin, line 2250: Just `return sn == n`

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,144 @@ def eval(cls, n):
...
+
+    """
+
+    @staticmethod
+    def is_schroeder(n):
+        try:
+            n = as_int(n)
+        except ValueError:
+            False
+        if n>0:
```
oscarbenjamin, line 2234: Put spaces around `>`

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,144 @@ def eval(cls, n):
...
+    the number of lattice paths from the southwest corner
+    $(0, 0)$ of an $n\times n$ grid to the northeast corner $(n, n)$ using only
+    single steps north, $(0, 1)$ northeast, $(1, 1)$ or east, $(1, 0)$
+    that do not rise above the SW-NE diagonal. The nth Schroeder number
+    is the number of Schroeder paths of length n. The Schroeder numbers were
+    named after the German mathematician Ernst Schroeder.
+
+    Schroeder numbers are the integer sequences defined by the
+    initial terms `S_0 = 1`, `S_1 = 2` and which satisfies the two-term recurrence relation
+    `S_n = \frac{6*n - 3}{n + 1} * S_{n-1} - \frac{n - 2}{n + 1} * S_{n-2}`.
```
oscarbenjamin, line 2198: Use consistent line wrapping for text in the docstring

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,144 @@ def eval(cls, n):
...
+            sn1 = sn
+            sn = int(b)
+            i = i+1
+
+        return schroeder
+
+    @staticmethod
+    def find_schroeder_numbers_in_range(x, y):
+        if 0 <= x <= y:
+            schroeder = list()
```
oscarbenjamin, line 2281: `schroeder  = []`

sympy/functions/combinatorial/tests/test_comb_numbers.py
```diff
@@ -451,6 +451,23 @@ def test_genocchi():
     raises(ValueError, lambda: genocchi(-2))
 
 
+def test_schroeder():
+    assert schroeder.is_schroeder(2) == True
```
oscarbenjamin, line 455: There are no tests for e.g. `schroeder(5)`. Add a test like:
```
nums = [...]
assert [schroeder(n) for n in range(10)] == nums
```

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,144 @@ def eval(cls, n):
...
+                schroeder.append(1)
+            sn1 = 1
+            sn = 2
+            i = 2
+            while sn <= y:
+                if sn >= x:
+                    schroeder.append(sn)
+                b = ((6*i - 3)*sn - (i - 2)*sn1)/(i + 1)
+                sn1 = sn
+                sn = int(b)
```
oscarbenjamin, line 2292: Integer division

sympy/functions/combinatorial/numbers.py
praneethratna: Yes, both x and y are inclusive, I have mentioned it in the docstring which i included now.

sympy/functions/combinatorial/numbers.py
praneethratna: I have made changes to the docstring but I'm not sure whether it is right way of formatting.

praneethratna:
ping @oscarbenjamin

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,149 @@ def eval(cls, n):
...
+                schroeder.append(1)
+            sn1 = 1
+            sn = 2
+            i = 2
+            while sn <= y:
+                if sn >= x:
+                    schroeder.append(sn)
+                b = ((6*i - 3)*sn - (i - 2)*sn1) // (i + 1)
+                sn1 = sn
+                sn = int(b)
```
oscarbenjamin, line 2303: If you're using integer division above then b should already be an int

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,149 @@ def eval(cls, n):
...
+    In mathematics, the Schroeder number,
+    also called a large Schroeder number or
+    big Schroeder number, describes the number of
+    lattice paths from the southwest corner $(0, 0)$
+    of an $n\times n$ grid to the northeast corner $(n, n)$ using only
+    single steps north, $(0, 1)$ northeast, $(1, 1)$ or east, $(1, 0)$
+    that do not rise above the SW-NE diagonal. The nth
+    Schroeder number is the number of Schroeder paths
+    of length n. The Schroeder numbers  were named
+    after the German mathematician Ernst Schroeder.
```
oscarbenjamin, line 2196: Any decent code editor will be able to fix this for you by wrapping lines to a fixed length.

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,149 @@ def eval(cls, n):
...
+            raise ValueError('The provided number must be a positive integer')
+        if n == 1:
+            return [1]
+        schroeder = [1, 2]
+        sn1 = 2
+        sn = 6
+        i = 3
+        while i <= n:
+            schroeder.append(sn)
+            b = ((6*i - 3)*sn - (i - 2)*sn1) // (i + 1)
```
oscarbenjamin, line 2274: This code for iterating over the Schroeder numbers is repeated in two different methods and this formula is also repeated in the `_schroeder` function. It should be possible to factor out common code. An iterator that yields the schroeder numbers could be used to implement at least these two methods without duplicating the code.

praneethratna: I have added a iterator function to iterate over schroeder numbers but I'm not sure whether it is the correct way.

sympy/functions/combinatorial/numbers.py
praneethratna: I have formatted the docstring such that each line has length of only 72 characters. Is it the correct way?

praneethratna:
ping @oscarbenjamin

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,141 @@ def eval(cls, n):
...
+    = \frac{6*n - 3}{n + 1} * S_{n-1} - \frac{n - 2}{n + 1} * S_{n-2}`.
+
+
+    Examples
+    ========
+
+    >>> from sympy import schroeder
+
+    >>> schroeder(2)
+    2
```
oscarbenjamin, line 2207: Some text to explain each group of examples would be useful e.g. "Compute the nth Schroeder number:"

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,141 @@ def eval(cls, n):
...
+            for i in schroeder.__iterator(y):
+                if i >= x and i <= y:
+                    schroder.append(i)
+
+            return schroder
+
+    @staticmethod
+    def __iterator(n):
+        sn, i = 2, 2
+        while i <= n:
```
oscarbenjamin, line 2295: I would probably be better just to make an infinite iterator that yields all the numbers. Also double underscore private names are not generally used in the sympy codebase.

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,155 @@ def eval(cls, n):
...
+            raise ValueError('The provided range is not valid. This condition should satisfy x <= y')
+        else:
+            schroder = []
+            if x <= 1 <= y:
+                schroder.append(1)
+            itr = schroeder._iterator()
+            for _ in range(1, y+1):
+                t = next(itr)
+                if t >= x and t <= y:
+                    schroder.append(t)
```
oscarbenjamin, line 2302: Rather than calling next the natural thing to do here would be to loop over the iterator and use break to exit the loop.

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,155 @@ def eval(cls, n):
...
+                return True
+
+            sn1 = 1
+            sn = 2
+            i = 2
+            while sn < n:
+                b = ((6*i - 3)*sn - (i - 2)*sn1) // (i + 1)
+                sn1 = sn
+                sn = b
+                i = i+1
```
oscarbenjamin, line 2260: Can this not use the iterator as well?

praneethratna: Yeah, now the code is very much simplified.

praneethratna:
ping @oscarbenjamin

praneethratna:
The `motzkin` class also needs similar refactoring of code and also there are combinatorial numbers like - [Narayana Number](https://en.wikipedia.org/wiki/Narayana_number), [Lobb Number](https://en.wikipedia.org/wiki/Lobb_number), [Delannoy Number](https://en.wikipedia.org/wiki/Delannoy_number) which can be added. Should i make these changes along with this PR or in a separate PR after this is merged?

praneethratna:
ping @oscarbenjamin

praneethratna:
@oscarbenjamin is this good to be merged?

praneethratna:
@oscarbenjamin are there any other changes required?

sympy/functions/combinatorial/numbers.py
```diff
@@ -2175,6 +2175,151 @@ def eval(cls, n):
...
+    .. [2] https://mathworld.wolfram.com/SchroederNumber.html
+
+    """
+
+    @staticmethod
+    def is_schroeder(n):
+        try:
+            n = as_int(n)
+        except ValueError:
+            False
```
faze-geek, line 2248: This probably should be return False

praneethratna:
ping @czgdp1807 for review

praneethratna:
ping @smichr for review.

PR #22841, opened 2022-01-11
watch for Add->Mul in as_numer_denom

### LINKED ISSUE #22837: Solve simplest algebraic equations with dummy parameter
Very strange behaviour of `solve` command in SymPy, when I try to solve quadratic equation with parameter `a` (which is dummy here, because solution doesn't depends on it):
```
from sympy import *
x, a = symbols("x a")
eq = Eq(0, (4 - 4*x + x**2)/(4*a**2))
print(solve(eq, x))
print(solve(simplify(eq), x))
```
Output:
```
[2 - sqrt(a**2 - 1)/a, 2 + sqrt(a**2 - 1)/a]
[2]
```
Just `solve` gives two (!) solutions, which depend on `a`. After symplifying it gives only solution `x=2`, which is correct. What happens? Command `solveset` works correctly, but I am interested in using `solve` command. My SymPy version is 1.8.

smichr:

fixes #22837 

```python
formerly

>>> Add(0,(x+y)/z/-2,evaluate=0).as_numer_denom()
(z*(x + y - 1) + 1, 2*z)

now
>>> Add(0,(x+y)/z/-2,evaluate=0).as_numer_denom()
(-x - y, 2*z)
```
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -613,7 +613,7 @@ class Add(Expr, AssocOp):
         """
         # clear rational denominator
         content, expr = self.primitive()
-        if not isinstance(expr, self.func):
+        if not isinstance(expr, Add):
             return Mul(content, expr, evaluate=False).as_numer_denom()
         ncon, dcon = content.as_numer_denom()
 
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -613,6 +613,8 @@ class Add(Expr, AssocOp):
         """
         # clear rational denominator
         content, expr = self.primitive()
+        if not isinstance(expr, self.func):
+            return Mul(content, expr, evaluate=False).as_numer_denom()
         ncon, dcon = content.as_numer_denom()
 
         # collect numerators and denominators of the terms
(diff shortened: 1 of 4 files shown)


sympy/core/add.py
```diff
@@ -613,6 +613,8 @@ def as_numer_denom(self):
         """
         # clear rational denominator
         content, expr = self.primitive()
+        if not isinstance(expr, self.func):
+            return Mul(content, expr, evaluate=False).as_numer_denom()
```
oscarbenjamin, line 617: Why not just `isinstance(expr, Add)`?

sympy/core/add.py
```diff
@@ -613,6 +613,8 @@ def as_numer_denom(self):
         """
         # clear rational denominator
         content, expr = self.primitive()
+        if not isinstance(expr, self.func):
```
smichr, line 616:
```suggestion
        if not isinstance(expr, Add):
```

sympy/core/add.py
smichr: I was being literal: if the function type changed, don't handle it here.

PR #22844, opened 2022-01-12
core/sympify: tests for resolution order of user defined sympify targets

ThePauliPrinciple:

Relates to https://github.com/sympy/sympy/pull/22785

This adds additional tests that ensure that the order in which user
defined sympify targets are resolved does not change without any
tests failing. Currently, such a change might pass during reviewing
without being intended.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/tests/test_sympify.py
+++ b/sympy/core/tests/test_sympify.py
@@ -662,6 +662,7 @@ def test_sympify_mro():
     assert sympify(b()) == Integer(2)
     assert sympify(c()) == Integer(1)
 
+
 def test_sympify_converter():
     """Tests the resolution order for classes in converter"""
     class a:
@@ -685,14 +686,16 @@ def test_sympify_converter():
         int_converter = converter[int]
     else:
         int_converter = None
-    converter[int] = MyInteger
 
-    assert sympify(1) == MyInteger(1)
+    try:
+        converter[int] = MyInteger
+        assert sympify(1) == MyInteger(1)
+    finally:
+        if int_converter is None:
+            del converter[int]
+        else:
+            converter[int] = int_converter
 
-    if int_converter is None:
-        del converter[int]
-    else:
-        converter[int] = int_converter
 
 def test_issue_13924():
     if not numpy:
--- a/sympy/core/tests/test_sympify.py
+++ b/sympy/core/tests/test_sympify.py
@@ -19,7 +19,7 @@ from sympy.sets.fancysets import Range
 from sympy.sets.sets import Interval
 from sympy.abc import x, y
 from sympy.core.sympify import (sympify, _sympify, SympifyError, kernS,
-    CantSympify)
+    CantSympify, converter)
 from sympy.core.decorators import _sympifyit
 from sympy.external import import_module
 from sympy.testing.pytest import raises, XFAIL, skip, warns_deprecated_sympy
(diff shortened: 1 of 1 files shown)


sympy/core/tests/test_sympify.py
```diff
@@ -647,6 +647,53 @@ def test_sympify_rational_numbers_set():
...
+    else:
+        int_converter = None
+    converter[int] = MyInteger
+
+    assert sympify(1) == MyInteger(1)
+
+    if int_converter is None:
+        del converter[int]
+    else:
+        converter[int] = int_converter
```
oscarbenjamin, line 695: This should use `try/finally`. Also it would be better to set a converter for a locally defined temporary class rather than messing with `int`.

sympy/core/tests/test_sympify.py
```diff
@@ -647,6 +647,53 @@ def test_sympify_rational_numbers_set():
...
+        def _sympy_(self):
+            return Integer(2)
+    class c(a):
+        pass
+
+    assert sympify(a()) == Integer(1)
+    assert sympify(b()) == Integer(2)
+    assert sympify(c()) == Integer(1)
+
+def test_sympify_converter():
```
oscarbenjamin, line 665: The convention is for two blank spaces between top-level functions/classes.

sympy/core/tests/test_sympify.py
ThePauliPrinciple: There are tests for temp defined classes (`a b c`), this is to see if overriding a sympy defined converter works.

I'll change it to a `try` block. When writing this, I thought it might be convenient if SymPy provides a context manager for this, so that users can temporarily change the behaviour of `sympify`/`converter`.

oscarbenjamin: I don't think that we should encourage users to do this. Changing the converter for something like `int` would break everything:
```python
In [1]: x + 1
Out[1]: x + 1

In [2]: from sympy.core.sympify import converter

In [3]: converter[int] = lambda e: E

In [4]: x + 1
Out[4]: x + ℯ
```

ThePauliPrinciple: While I agree that it would be a problem if a user would replace the converter with something that doesn't work, I do think this should be tested for (or explicitly forbidden), considering the changes proposed in https://github.com/sympy/sympy/pull/22785 where a future change might introduce a situation where `converter[int] =` does not actually change the behaviour of `sympify` at all (e.g. if it first checks `_sympy_converter`) eventhough the user explicitly indicated to do something different from the SymPy setting.

oscarbenjamin: Adding a test is fine but we shouldn't add utilities like a context manager to encourage anyone to do this.

ThePauliPrinciple: Is this resolved now?

oscarbenjamin: Yes, looks fine. It seems like overkill to check whether int converter is set or not: it should always be set but not a big deal. Main thing is that even if the test fails the converter will be reset.

ThePauliPrinciple: In https://github.com/sympy/sympy/pull/22785 it would fail, since `converter` would no longer contain SymPy defined converters

oscarbenjamin: Oh, I see what you mean! Ok, this is fine.

Issue #22400, opened 2021-11-01
Minpoly doesn't terminate


SofiaSL:
This code

`from sympy import *
minpoly(root(2,3)+root(3,3)+(-1+I*sqrt(3))/2*root(5,3))`

doesn't terminate.

oscarbenjamin:
There should be some limit on what `minpoly` attempts to do but in this case the problem is factoring this polynomial:
```python
> /home/oscar/current/sympy/sympy.git/sympy/polys/numberfields/minpoly.py(288)_minpoly_op_algebraic_element()
    286 
    287     r = Poly(r, x, domain=dom)
--> 288     _, factors = r.factor_list()
    289     res = _choose_factor(factors, x, op(ex1, ex2), dom)
    290     return res.as_expr()

ipdb> p r
[...]

Issue #20682, opened 2020-12-28
Better API for diophantine and handling systems


oscarbenjamin:
The `diophantine` function is useful but is awkward to use e.g.:
https://stackoverflow.com/a/65438529/9450991

It would be much more useful if `diophantine` could:

1. Have a consistent interface with predictable and usable return values
2. Solve systems of diophantine equations
3. Support combining inequalities and equations to produce finite solution sets

I have some example code that builds something more usable based on diophantine below. First let's see what's awkward about diophantine.

In basic interactive usage diophantine seems okay. It works only for a single equation but in possibly several unknowns:
```python
In [2]: diophantine(x**2 + y**2 - 1)
Out[2]: {(-1, 0), (0, -1), (0, 1), (1, 0)}

In [3]: diophantine(x**2 + y**2 + 1)
Out[3]: set()

In [4]: diophantine(x + y**2)
Out[4]: 
⎧⎛  2    ⎞⎫
⎨⎝-t , -t⎠⎬
⎩         ⎭
```
When there are infinitely many solutions diophantine returns solutions that are parametrised in terms of newly created symbols. It would be much more useful if the interface could tell you what those symbols are. They can be found using `free_symbols` (although this wouldn't work if `diophantine` supported systems with parameter symbols in):
```python
In [5]: diophantine(x + y + z)
Out[5]: {(t₀, t₁, -t₀ - t₁)}

In [6]: sol, = diophantine(x + y + z)

In [7]: set().union(*(s.free_symbols for s in sol))
Out[7]: {t₀, t₁}
```
Having access to the symbol objects makes it possible to do things like substitute values for them or solve for their values in inequalities.

The current signature of `diophantine` is
```python
solset = diophantine(equation, param=t, symbols=None)
```
where the idea is that param is used to specify the symbol that will be used when there are infinitely solutions. That doesn't work though because often more than one symbol is needed to parametrise the solution and so diophantine ends up using different symbols like `t0, t1` instead of the provided symbol `t`.

The `symbols` parameter is analogous to calling solve and specifying the unknowns except that symbols not appearing in the equation are essentially ignored:
```python
In [25]: diophantine(x, syms=[x, y, z])
Out[25]: {(0,)}

In [26]: diophantine(x+y, syms=[x, y, z])
Out[26]: {(t₀, -t₀)}

In [27]: diophantine(x+y+z, syms=[x, y, z])
Out[27]: {(t₀, t₁, -t₀ - t₁)}
```
[...]

PR #22845, opened 2022-01-12
add igcdLLL to numbers

smichr:
#### Other comments
- algorithm ref [here](https://projecteuclid.org/euclid.em/1048515660/)**
- discussion of use of the algorithm and implementations in PHIX, WREN and FREEBASIC [here](https://rosettacode.org/wiki/Diophantine_linear_system_solving)
- LLL descriptions [here](https://www.math.auckland.ac.nz/~sgal018/crypto-book/ch17.pdf) and [here](https://algo.epfl.ch/_media/en/projects/bachelor_semester/rapportetiennehelfer.pdf) and a purportedly inefficient version in #22400
- [A polynomial time algorithm for computing the HNF of a
module over the integers of a number field](http://biasse.myweb.usf.edu/papers/HNF_pol.pdf)
- [uses of the LLL routine](https://www.math.ru.nl/~bosma/onderwijs/voorjaar07/compalg8.pdf)
- 
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -439,7 +439,7 @@ def ilcm(*args):
     return a
 
 
-def igcdLLL(s, m1=2, n1=3, optimize=False):
+def igcdLLL(s, m1=2, n1=3, optimize=None):
     """return g, B where g is the gcd of the integers in s and the
     LLL-reduced basis and small multipliers such that those
     multiples of the corresponding values of s is g.
@@ -450,29 +450,70 @@ def igcdLLL(s, m1=2, n1=3, optimize=False):
     This is an iterative procedure whose quality in returning
     small multipliers is related to alpha = m1/n1 which must
     be in the range [3/8, 1]. When 1, the algorithm will iterate
-    longer but is not guaranteed to return the smallest multipliers
-    except for the case of 3 values in s.
+    longer (but is not guaranteed to return the smallest multipliers
+    except for the case of 3 values in `s` unless `optimize` is False).
 
     Examples
     ========
 
     >>> from sympy.core.numbers import igcdLLL
-    >>> from sympy import Add, Matrix, symbols
-    >>> x = symbols('x:%s' % len(s))
+    >>> from sympy import symbols
+    >>> from sympy.abc import x, y, c
+    >>> mul = lambda x, y: [i*j for i, j in zip(x, y)]
+    >>> dot = lambda x, y: sum(mul(x, y))
+    >>> dot((2, 3), (x, y))  # a helper needed below
+    2*x + 3*y
 
-    >>> s = 2, 4, 6
-    >>> eq = Add(*[si*xi for si, xi in zip(s, x)]); eq
-    2*x0 + 4*x1 + 6*x2
+    The greatest common divisor for elements of `s` and a list of multipliers are
+    computed by `igcdLLL`:
 
+    >>> s = (2, 4, 6)
     >>> g, B = igcdLLL(s)
-    >>> assert Add(*[si*mi for si, mi in zip(s, B[-1])]) == g
+    >>> g
+    2
+    >>> B
+    [[1, 1, -1], [2, -1, 0], [1, 0, 0]]
 
-    >>> p = symbols('p:%s' % (len(B) - 1))
-    >>> psol = (Matrix(B[:-1]).T*Matrix(p)).T.tolist()[0]
-    >>> dict(zip(x, psol))
-    {x0: p0 + 2*p1, x1: p0 - p1, x2: -p0}
-    >>> eq.subs(_)
+    The last row of `B` gives a list of small multipliers for the value of `s`
+    that will produce a sum equal to `g` while the others will produce a zero sum:
+
+    >>> [dot(s, b) for b in B]
+    [0, 0, 2]
+
+    One way these results can be used is to give a parametric solution
+    for `2*x + 4*y + 6*z = c`. Let `p0` and `p1` be the multiples for the
+    first two rows:
+
+    >>> p0, p1 = symbols('p:2')
+
+    Any multiple of the first rows in B will still sum to zero as will their sum:
+
+    >>> mul([p0]*3, B[0])
+    [p0, p0, -p0]
+    >>> mul([p1]*3, B[1])
+    [2*p1, -p1, 0]
+
+    Their combination will also create multipliers that will set the sum to zero:
+
+    >>> comb = [p0 + 2*p1, p0 - p1, -p0]
+    >>> dot(comb, s)
     0
+
+    In order to get the sum to equal `c`, the last row of multipliers will be
+    multiplied by `c/2`:
+
+    >>> last = mul([c/2]*3, B[-1]); last
+    [c/2, 0, 0]
+    >>> dot(_, s)
+    c
+
+    The general solution, then, is a combination of this and the homogeneous solutions from
+    the basis:
+
+    >>> gensol = [i + j for i, j in zip(comb, last)]; gensol
+    [c/2 + p0 + 2*p1, p0 - p1, -p0]
+    >>> dot(_, s)
+    c
     """
     if not iterable(s):
         raise ValueError('s should be iterable')
@@ -485,22 +526,38 @@ def igcdLLL(s, m1=2, n1=3, optimize=False):
         raise ValueError('require 3/8 <= alpha <= 1 but alpha = %s/%s' % (m1, n1))
     if optimize and len(s) > 3:
         raise ValueError('optimization for more than 3 values')
+    elif len(s) == 3 and optimize is None:
+        optimize = True
+
+    def iround(num, den):
+        # round away from 0 on tie
+        if den < 0:
+            num = -num
+            den = -den
+        neg = (num < 0)
+        if neg:
+            num = -num
+        w, r = divmod(num, den)
+        w = int(w)  # la can be float right now
+        if 2*r >= den:
+            w += 1
+        return -w if neg else w
 
     def reduce1(k, i):
         if a[i] != 0:
-            q = int(round(a[k] / a[i]))
+            q = iround(a[k], a[i])
         else:
-            if 2 * abs(la[k][i]) > D[i]:
-                q = int(round(la[k][i] / D[i]))
+            if 2*abs(la[k][i]) > D[i]:
+                q = iround(la[k][i], D[i])
             else:
                 q = 0
         if q != 0:
-            a[k] -= q * a[i]
+            a[k] -= q*a[i]
             for c in range(m):
-                B[k][c] -= q * B[i][c]
-            la[k][i] -= q * D[i]
+                B[k][c] -= q*B[i][c]
+            la[k][i] -= q*D[i]
             for j in range(i):
-                la[k][j] -= q * la[i][j]
+                la[k][j] -= q*la[i][j]
 
     def swap(k):
 
@@ -512,25 +569,24 @@ def igcdLLL(s, m1=2, n1=3, optimize=False):
         for i in range(j):
             (la[k][i], la[j][i]) = (la[j][i], la[k][i])
         for i in range(k + 1, m):
-            t = la[i][j] * D[k] - la[i][k] * la[k][j]
-            la[i][j] = (la[i][j] * la[k][j] + la[i][k] * D[j - 1]) \
-                / D[j]
-            la[i][k] = t / D[j]
-        D[j] = (D[k - 2] * D[k] + la[k][j] ** 2) / D[j]
+            t = la[i][j]*D[k] - la[i][k]*la[k][j]
+            la[i][j] = (la[i][j]*la[k][j] + la[i][k]*D[j - 1])/D[j]
+            la[i][k] = t/D[j]
+        D[j] = (D[k - 2]*D[k] + la[k][j]**2)/D[j]
 
     m = len(s)
-    B = [[0] * m for i in range(m)]
+    B = [[0]*m for i in range(m)]
     for i in range(m):
         B[i][i] = 1
-    la = [[0] * i for i in range(m)]
-    D = [1] * (m + 1)  # m+1 referenced as D[-1]
+    la = [[0]*i for i in range(m)]
+    D = [1]*(m + 1)  # m+1 referenced as D[-1]
     a = list(s)
     k = 1
     while k < m:
         j = k - 1
         reduce1(k, j)
-        if a[j] != 0 or a[j] == 0 and a[k] == 0 and n1 * (D[j - 1]
-                * D[k] + la[k][j] ** 2) < m1 * D[j] ** 2:
+        if a[j] != 0 or (a[j] == 0 and a[k] == 0 and
+                n1*(D[j - 1]*D[k] + la[k][j] ** 2) < m1*D[j]**2):
             swap(k)
             if k > 1:
                 k = j
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -439,18 +439,59 @@ def ilcm(*args):
     return a
 
 
-def igcdLLL(s, m1=2, n1=3, optimize=None):
-    """return g, B where g is the gcd of the integers in s and
-    B contains the LLL-reduced basis and Bezout coefficients.
+def iround(num, den):
+    # round away from 0 on tie
+    if den < 0:
+        num = -num
+        den = -den
+    neg = (num < 0)
+    w, r = divmod(-num if neg else num, den)
+    w = int(w)
+    if 2*r >= den:
+        w += 1
+    return -w if neg else w
+
+
+def idiv(a, b):
+    # safe a/b asserting that b divides a evenly
+    assert a%b == 0
+    return a//b
+
+
+def _swap(k, cols, A, B, la, D):
+    # swap rows k and k - 1 and calculate new la and D
+    j = k - 1
+    for M in (A, B):
+        M[k], M[j] = M[j], M[k]
+    for i in range(j):
+        la[k][i], la[j][i] = la[j][i], la[k][i]
+    for i in range(k + 1, cols):
+        t = la[i][j]*D[k] - la[i][k]*la[k][j]
+        la[i][j] = idiv(
+            la[i][j]*la[k][j] + la[i][k]*D[j - 1],
+            D[j])
+        la[i][k] = idiv(t, D[j])
+    D[j] = idiv(D[j - 1]*D[k] + la[k][j]**2, D[j])
+
+
+def igcdLLL(s, _n=2, _d=3, optimize=None):
+    """return g, B where g is the gcd of the integers in s and the
+    LLL-reduced basis and small multipliers such that those
+    multiples of the corresponding values of s is g.
 
     If there are three values in s then the smallest multipliers
     will be returned unless optimize is False.
 
     This is an iterative procedure whose quality in returning
-    small multipliers is related to alpha = m1/n1 which must
-    be in the range [3/8, 1]. When 1, the algorithm may iterate
-    longer (but is not guaranteed to return the smallest multipliers
-    except for the case of 3 values in `s` unless `optimize` is False).
+    small multipliers is related to ``alpha = _n/_d`` which should
+    be in the range of (0, 1]; the value of 3/8 is useful for selecting
+    the optimum result for 3 numbers. When alpha is 1, the algorithm
+    will iterate longer (but is not guaranteed to return the smallest
+    multipliers except for the case of 3 values in `s` unless
+    `optimize` is False).
+
+    The method used is that of Algorithm 3 in the paper by
+    Havas, Majewski and Matthews [1].  XXX how to cite properly?
 
     Examples
     ========
@@ -516,6 +557,16 @@ def igcdLLL(s, m1=2, n1=3, optimize=None):
     [c/2 + p0 + 2*p1, p0 - p1, -p0]
     >>> dot(_, s)
     c
+
+    See Also
+    ========
+    hnf - putting integer matrix into Hermitian Normal form
+
+    References
+    ==========
+
+    .. [1] https://projecteuclid.org/euclid.em/1048515660/
+    .. [2] https://rosettacode.org/wiki/Diophantine_linear_system_solving
     """
     if not iterable(s):
         raise ValueError('s should be iterable')
@@ -523,58 +574,29 @@ def igcdLLL(s, m1=2, n1=3, optimize=None):
         raise ValueError('expecting 2 or more values')
     if not all(type(i) is int for i in s):
         raise ValueError('expecting literal ints for s')
-    ok = (3*n1 <= 8*m1) and (n1 > 0) and (m1 > 0) and (m1 <= n1)
+    ok = (_d > 0) and (_n > 0) and (_n <= _d)
     if not ok:
-        raise ValueError('require 3/8 <= alpha <= 1 but alpha = %s/%s' % (m1, n1))
+        raise ValueError('alpha must be in (0, 1], but is %s/%s' % (_n, _d))
     if optimize and len(s) > 3:
         raise ValueError('optimization for more than 3 values')
     elif len(s) == 3 and optimize is None:
         optimize = True
 
-    def iround(num, den):
-        # round away from 0 on tie
-        if den < 0:
-            num = -num
-            den = -den
-        neg = (num < 0)
-        if neg:
-            num = -num
-        w, r = divmod(num, den)
-        w = int(w)  # la can be float right now
-        if 2*r >= den:
-            w += 1
-        return -w if neg else w
-
     def reduce1(k, i):
-        if a[i] != 0:
-            q = iround(a[k], a[i])
+        if A[i] != 0:
+            q = iround(A[k], A[i])
         else:
             if 2*abs(la[k][i]) > D[i]:
                 q = iround(la[k][i], D[i])
             else:
                 q = 0
         if q != 0:
-            a[k] -= q*a[i]
+            A[k] -= q*A[i]
             for c in range(m):
                 B[k][c] -= q*B[i][c]
             la[k][i] -= q*D[i]
             for j in range(i):
                 la[k][j] -= q*la[i][j]
-
-    def swap(k):
-
-        # swap rows k and k - 1 and calculate new la and D
-
-        j = k - 1
-        (a[k], a[j]) = (a[j], a[k])
-        (B[k], B[j]) = (B[j], B[k])
-        for i in range(j):
-            (la[k][i], la[j][i]) = (la[j][i], la[k][i])
-        for i in range(k + 1, m):
-            t = la[i][j]*D[k] - la[i][k]*la[k][j]
-            la[i][j] = (la[i][j]*la[k][j] + la[i][k]*D[j - 1])/D[j]
-            la[i][k] = t/D[j]
-        D[j] = (D[k - 2]*D[k] + la[k][j]**2)/D[j]
 
     m = len(s)
     B = [[0]*m for i in range(m)]
@@ -582,24 +604,28 @@ def igcdLLL(s, m1=2, n1=3, optimize=None):
         B[i][i] = 1
     la = [[0]*i for i in range(m)]
     D = [1]*(m + 1)  # m+1 referenced as D[-1]
-    a = list(s)
+    A = list(s)
     k = 1
     while k < m:
         j = k - 1
         reduce1(k, j)
-        if a[j] != 0 or (a[j] == 0 and a[k] == 0 and
-                n1*(D[j - 1]*D[k] + la[k][j] ** 2) < m1*D[j]**2):
-            swap(k)
+        if A[j] != 0 or (A[j] == 0 and A[k] == 0 and
+                _d*(D[j - 1]*D[k] + la[k][j] ** 2) < _n*D[j]**2):
+            _swap(k, m, A, B, la, D)
             if k > 1:
                 k = j
         else:
             for i in range(j - 1, -1, -1):
                 reduce1(k, i)
             k += 1
+
     # make gcd positive and update multipliers to be consistent
-    if a[-1] < 0:
-        a[-1] = -a[-1]
+    gcd = A[-1]
+    # del A, la, D - we are done with these but flake8 complains
+    if gcd < 0:
+        gcd = -gcd
         B[-1] = [-i for i in B[-1]]
+
     if optimize and len(s) == 3:
         mag = lambda x: sum([i**2 for i in x])
         add = lambda x, y: [i + j for i, j in zip(x, y)]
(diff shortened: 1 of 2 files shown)


sympy/core/numbers.py
```diff
@@ -438,6 +439,142 @@ def ilcm(*args):
...
+    except for the case of 3 values in s.
+
+    Examples
+    ========
+
+    >>> from sympy.core.numbers import igcdLLL
+    >>> from sympy import Add, Matrix, symbols
+    >>> x = symbols('x:%s' % len(s))
+
+    >>> s = 2, 4, 6
```
smichr, line 463:
```suggestion

    >>> s = 2, 4, 6
    >>> x = symbols('x:%s' % len(s))
```

sympy/core/numbers.py
```diff
@@ -438,6 +439,142 @@ def ilcm(*args):
...
+
+        j = k - 1
+        (a[k], a[j]) = (a[j], a[k])
+        (B[k], B[j]) = (B[j], B[k])
+        for i in range(j):
+            (la[k][i], la[j][i]) = (la[j][i], la[k][i])
+        for i in range(k + 1, m):
+            t = la[i][j] * D[k] - la[i][k] * la[k][j]
+            la[i][j] = (la[i][j] * la[k][j] + la[i][k] * D[j - 1]) \
+                / D[j]
```
oscarbenjamin, line 517: It's better to have a line that is slightly longer than to wrap like this.

sympy/core/numbers.py
```diff
@@ -438,6 +439,142 @@ def ilcm(*args):
...
+        raise ValueError('require 3/8 <= alpha <= 1 but alpha = %s/%s' % (m1, n1))
+    if optimize and len(s) > 3:
+        raise ValueError('optimization for more than 3 values')
+
+    def reduce1(k, i):
+        if a[i] != 0:
+            q = int(round(a[k] / a[i]))
+        else:
+            if 2 * abs(la[k][i]) > D[i]:
+                q = int(round(la[k][i] / D[i]))
```
oscarbenjamin, line 494: There are a few places here that do `int / int -> float` division. Is it possible to write this without floating point division? This should presumably fail in some cases for very large integers (`> 2^53`).

sympy/core/numbers.py
```diff
@@ -438,6 +439,142 @@ def ilcm(*args):
...
+    >>> from sympy.core.numbers import igcdLLL
+    >>> from sympy import Add, Matrix, symbols
+
+    >>> s = 2, 4, 6
+    >>> x = symbols('x:%s' % len(s))
+    >>> eq = Add(*[si*xi for si, xi in zip(s, x)]); eq
+    2*x0 + 4*x1 + 6*x2
+
+    >>> g, B = igcdLLL(s)
+    >>> assert Add(*[si*mi for si, mi in zip(s, B[-1])]) == g
```
oscarbenjamin, line 468: These examples seem almost designed to obfuscate what this function is doing.  At least a simple example that just directly shows the input and output would be a lot clearer:
```python
In [4]: gcd([2, 4, 6])
Out[4]: 2

In [5]: igcdLLL([2, 4, 6])
Out[5]: (2, [[1, 1, -1], [2, -1, 0], [1, 0, 0]])

In [6]: 1*2 + 1*4 + (-1)*6
Out[6]: 0
```
Using `Add` and other constructs complicates things.

oscarbenjamin: What exactly is this a basis for? Is `B[-1]` guaranteed to give the `Bezout` coefficients?

oscarbenjamin: There should be a reference to the paper for the algorithm

oscarbenjamin: It looks like `B` is the matrix sending the input numbers to `[0, ..., g].T`:
```python
In [5]: from sympy.core.numbers import igcdLLL

In [6]: nums = [3, 6, 12]

In [7]: g, B = igcdLLL(nums)

In [8]: g
Out[8]: 3

In [9]: B
Out[9]: [[-2, 1, 0], [0, -2, 1], [1, 0, 0]]

In [10]: Matrix(B) * Matrix(nums)
Out[10]: 
⎡0⎤
⎢ ⎥
⎢0⎥
⎢ ⎥
⎣3⎦
```
More generally I guess that if `nums` generates a Z-module then `B` is a map from that Z-module to the Z-module generated by `[0, ..., g]` where `g` is the gcd of nums.

smichr: What about
```python
The following function will compute weighted sums of values:

>>> dot = lambda x, y: sum([i*j for i, j in zip(x, y)])
>>> dot((x, y), (2, 3))
2*x + 3*y

The greatest common divisor for elements of `s` and a list of multipliers are
computed by `igcdLLL`:

>>> g, B = igcdLLL(s)
>>> g
2
>>> B
[[1, 1, -1], [2, -1, 0], [1, 0, 0]]

The last row of `B` gives a list of small multipliers for the value of `s` that will produce a
sum equal to `g` while the others will produce a zero sum:

>>> [dot(s, b) for b in B]
[0, 0, 2]
```
Should it just stop there or shall I add
```python
One way these results can be used is to give a parametric solution for `2*x + 4*y + 6*z = c`.
Let `p0` and `p1` be the multiples for the first two rows; any mutliple of them will still sum to zero.
In order to get the sum to equal `c`, the last row of multipliers will be multiplied by `c/2`:

>>> gensol = (Matrix(B).T*Matrix([p0, p1, c/2])).T.tolist()[0]; gensol
[c/2 + p0 + 2*p1, p0 - p1, -p0]
>>> dot(s, _)
c
```

sympy/core/numbers.py
smichr: I was thinking the same and had each of those lines tested in the version I tried to translate (without success) from Rosetta Code.

Another issue is the `round`. Python rounds away from zero. I'm not sure that this will make a difference. The HMM paper uses `\ceiling x \floor` and the translations on the (Rosetta Code page]() uses the round-away-even-and-odd-the-same sort of `round`, too. So we might need a custom `round` to handle unlimited integer size:
```python
def iround(num, den):
  # round away from 0 on tie
  if num < 0:
    return -iround(-num, den)
  w, r =divmod(num, den)
  r2 = 2*r
  if r2 < den:
    return w
  return w + 1
```

sympy/core/numbers.py
smichr: `black` agrees

sympy/core/numbers.py
smichr: Changes to that effect were made.

Here is where there is division which is not guaranteed to be integer:
```
        for i in range(k + 1, m):
            t = la[i][j]*D[k] - la[i][k]*la[k][j]
            #la[i][j] = (la[i][j]*la[k][j] + la[i][k]*D[j - 1])/D[j]
            la[i][j] = round_out((la[i][j]*la[k][j] + la[i][k]*D[j - 1]),D[j])
            #assert la[i][j] == int(la[i][j]), la[i][j] # [-9203, -9333, 3922] -1927141.000024883
            #la[i][k] = t/D[j]
            la[i][k] = round_out(t,D[j])
            #assert la[i][k] == int(la[i][k]), la[i][k] # [9221, -7615, -7571],-70482653.00000001
        #D[j] = (D[k - 2]*D[k] + la[k][j]**2)/D[j]
        D[j] = round_out((D[k - 2]*D[k] + la[k][j]**2),D[j])
        #assert D[j] == int(D[j]), D[j] # [9997, 9079, -5001], 182368249.99999997
```
I used the round function on them but I am not sure of the implications of this in terms of the algorithm. All I have confirmed so far is that after running for a while with 3-tuples of values in the range of [-10**4, 10**4] it doesn't hang.

sympy/core/numbers.py
```diff
@@ -438,6 +439,196 @@ def ilcm(*args):
...
+    >>> from sympy.core.numbers import igcdLLL
+    >>> from sympy import symbols
+    >>> dot = lambda x, y: sum([i*j for i, j in zip(x, y)])
+    >>> dot((2, 3), (x, y))  # a helper needed below
+    2*x + 3*y
+
+    The greatest common divisor for elements of `s` and a list of multipliers are
+    computed by `igcdLLL`:
+
+    >>> g, B = igcdLLL(s)
```
smichr, line 468:
```suggestion
    >>> from sympy import symbols
    >>> from sympy import x, y, c
    >>> dot = lambda x, y: sum([i*j for i, j in zip(x, y)])
    >>> dot((2, 3), (x, y))  # a helper needed below
    2*x + 3*y

    The greatest common divisor for elements of `s` and a list of multipliers are
    computed by `igcdLLL`:

    >>> s = (2, 4, 6)
    >>> g, B = igcdLLL(s)
```

sympy/core/numbers.py
skieffer: This looks interesting. You mentioned a paper. Do you have the reference?

I'm wondering about the division with rounding. I'm just surprised that it's not ordinary exact division, i.e. `q = a[k] // a[i]`.

sympy/core/numbers.py
```diff
@@ -438,6 +439,198 @@ def ilcm(*args):
...
+            den = -den
+        if num < 0:
+            neg = True
+            num = -num
+        else:
+            neg = False
+        w, r = divmod(num, den)
+        if 2*r < den:
+            return -w if neg else w
+        return -w + 1 if neg else w + 1
```
smichr, line 544:
```suggestion
        neg = (num < 0)
        w, r = divmod(num, den)
        w = int(w)  # la can be float right now
        if neg:
            w = -w
        if 2*r < den:
            return w
        return w + 1
```

sympy/core/numbers.py
smichr: @skieffer , refs are in the OP.  They will be added to the commit later. I still need to link this up to the `diophantine` module. There are many more applications for this as discussed on the Rosetta Code link (given in the OP).

sympy/core/numbers.py
oscarbenjamin: Okay this is a bit clearer now. The first two rows of B are a basis for the integer solutions of `2*x + 4*y + 6*z = 0`. The last row of B gives coefficients for `gcd(2, 4, 6)` as a linear combination of 2, 4 and 6.

I'm not clear how exactly this is different from the Bezout coefficients:
https://en.wikipedia.org/wiki/B%C3%A9zout%27s_identity#For_three_or_more_integers

sympy/core/numbers.py
```diff
@@ -438,6 +439,198 @@ def ilcm(*args):
...
+    >>> last = mul([c/2]*3, B[-1]); last
+    [c/2, 0, 0]
+    >>> dot(_, s)
+    c
+
+    The general solution, then, is a combination of this and the homogeneous solutions from
+    the basis:
+
+    >>> gensol = [i + j for i, j in zip(comb, last)]; gensol
+    [c/2 + p1 + 2*p1, p0 - p1, -p0]
```
smichr, line 514:
```suggestion
    >>> mul([p1]*3, B[1])
    [2*p1, -p1, 0]

    Their combination will also create multipliers that will set the sum to zero:

    >>> comb = [p0 + 2*p1, p0 - p1, -p0]
    >>> dot(comb, s)
    0

    In order to get the sum to equal `c`, the last row of multipliers will be
    multiplied by `c/2`:

    >>> last = mul([c/2]*3, B[-1]); last
    [c/2, 0, 0]
    >>> dot(_, s)
    c

    The general solution, then, is a combination of this and the homogeneous solutions from
    the basis:

    >>> gensol = [i + j for i, j in zip(comb, last)]; gensol
    [c/2 + p0 + 2*p1, p0 - p1, -p0]
```

sympy/core/numbers.py
smichr: >  how exactly this is different from the Bezout coefficients:

I don't know a lot about this, but I guess you could say the same thing about the extra parameters from the extended gcd calculation. For me they just appear to be roses called by other names.

sympy/core/numbers.py
smichr: @skieffer , the expectation is that division is exact or is explicitly floor-related. I will update the code. The reason I was getting floats is because ints-as-floats lost precision at some point and then the assertion that `int(afloat) == afloat` began to fail.

oscarbenjamin:
Is it possible to decouple the LLL part of the algorithm here from the igcd part so that we also have an indpendently usable LLL algorithm?

It would be useful to have LLL both for internal use and also as a public API.

smichr:
Although this routine affirms that `P*G = H` I get a different `P` for the data from example 7.4:
```
HMM paper referenced in routine

-10  -8 -5  1  2  3  5  3  0 -4
 -2  -1  0  1 -1  0  1  0  1 -1
-15 -11 -4  0  4  5  4  3  1 -5
  1  -1 -1  0  2 -1  0  0  0  0
  0   1 -1 -1  1 -1  2 -1  0  0
  1   0 -1 -1 -1  2  0  1 -1  0
  1   0 -2  1 -1  1 -1  1  1 -1
 -1   0  1  0  1  1 -1 -2  0  1
  1  -1  0 -1  1  0  0 -1  2 -1
  1  -2  1  1 -2  0  2 -1  0  0

This routine gives
alpha = 1/10
-12 -4 -6 -1  3  3  5  3  0 -4
  0 -4  1 -1  1  1  0  0  1 -1
-15 -9 -7  1  3  5  5  4  0 -5
 -1  2 -2  1  0  1  0 -1 -1  1
  0 -2  3 -1  1  0 -1 -1  1  0
  1 -2  0  1  1  0 -2  1  0  0
 -1  2 -2  2  0 -2  1  0  0  0
 -1  1  1  0 -2  1  0  0  0  0
 -2  4 -1 -2  1  0  0  0  0  0
  3 -8  7 -2  0  0  0  0  0  0
alpha = 5/10
-10  -8 -5  1  2  3  5  3  0 -4
 -2  -1  0  1 -1  0  1  0  1 -1
-15 -11 -4  0  4  5  4  3  1 -5
  0   0 -1  1  1 -1  0  1 -2  1
  0  -1  2  0 -1 -1  0  2 -1  0
 -1   0  1  1  1 -2  0 -1  1  0
 -1   1 -1  3 -1 -1 -1  1  0  0
 -1   2 -1 -1  2  0 -2  1  0  0
  0  -1  1  1 -1  1 -2  1  0  0
 -1   1  1  0 -2  1  0  0  0  0
alpha = 6/10
-10  -8 -5  1  2  3  5  3  0 -4
 -2  -1  0  1 -1  0  1  0  1 -1
-15 -11 -4  0  4  5  4  3  1 -5
  1  -2  2 -1  0 -1  0  1  1 -1
  0   0 -1  1  1 -1  0  1 -2  1
 -1   0  1  1  1 -2  0 -1  1  0
 -1   1 -1  3 -1 -1 -1  1  0  0
 -1   2 -1 -1  2  0 -2  1  0  0
  0  -1  1  1 -1  1 -2  1  0  0
 -1   1  1  0 -2  1  0  0  0  0
alpha = 7/10
-11  -7 -4  1  0  4  5  3  0 -4
 -2  -1  0  1 -1  0  1  0  1 -1
-15 -11 -4  0  4  5  4  3  1 -5
 -1   1  2 -2  0  0 -2  2  1 -1
  0  -1  2  0 -1 -1  0  2 -1  0
  1   0 -2  1 -1  1 -1  1  1 -1
  0   0 -1  1  1 -1  0  1 -2  1
 -1   0  1  1  1 -2  0 -1  1  0
  0  -1  1  1 -1  1 -2  1  0  0
 -1   1  1  0 -2  1  0  0  0  0
alpha = 8/10
-11  -8 -4  1  3  4  4  1  0 -3
 -2  -1  0  1 -1  0  1  0  1 -1
-15 -11 -4  0  4  5  4  3  1 -5
 -1   2 -1 -1  2  0 -2  1  0  0
 -1   1  0  1 -1  0  0  1 -2  1
 -1   0  2 -1  1 -1  1 -1 -1  1
  0  -1  2  0 -1 -1  0  2 -1  0
 -1   0  1  1  1 -2  0 -1  1  0
  0  -1  1  1 -1  1 -2  1  0  0
 -1   1  1  0 -2  1  0  0  0  0
```

sympy/core/numbers.py
```diff
@@ -438,6 +439,342 @@ def ilcm(*args):
...
+        for j in range(len(B) - 1):
+            B[j] = [-i for i in B[j]]
+
+    return (gcd, B)
+
+
+def hnf(G, _n=3, _d=4):
+    """return ``n``, ``H``, ``P`` where ``n`` is the number of
+    iterations needed to find ``H`` and ``P`` such that ``H*G = P``
+    with ``H`` being in Hermite normal form and ``P`` being unimodular.
```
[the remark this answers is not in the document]
smichr: Size of the multipliers(?). I don't understand enough to make a direct comparison, but consider the following example from the paper:
```python
>>> from sympy.matrices.normalforms import hermite_normal_form
>>> G = Matrix(10,10,lambda i,j:(i+1)**3*(j+1)**2+i+j+2)
>>> from sympy.core.numbers import hnf
>>> hermite_normal_form(G)
Matrix([
[2880, 9, 1432],
[2352, 8, 1169],
[1848, 7,  918],
[1380, 6,  685],
[ 960, 5,  476],
[ 600, 4,  297],
[ 312, 3,  154],
[ 108, 2,   53],
[   0, 1,    0],
[   0, 0,    1]])
>>> Matrix(hnf(G.tolist())[-2][:3]).T
Matrix([
[  1,  0,   0],
[  0,  1,   0],
[  7,  4,  12],
[ 22,  9,  36],
[ 45, 16,  72],
[ 76, 25, 120],
[115, 36, 180],
[162, 49, 252],
[217, 64, 336],
[280, 81, 432]])
```
Also, hnf gives the unimodular matrix in addition to the Hermitian normal form.

I tried another small random matrix and note this:
```python
>>> m
Matrix([
[98, 23, 61],
[71, 45, 78],
[20, 60, 16]])
>>> hermite_normal_form(m)
Matrix([
[21671, 4905, 19657],
[    0,    2,     1],
[    0,    0,     4]])
>>> Matrix(hnf(m.tolist())[-2])
Matrix([
[1, 0,   3871],
[0, 1,  36273],
[0, 0, 173368]])
```
Really, any more on this should be done by someone with more insight into these things. The Rosetta Code link has an extension of this code there. I started there but ended up being partially successful by going back to the original source.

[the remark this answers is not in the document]
smichr: Is the `hermite_normal_form` matrix upside down?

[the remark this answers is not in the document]
oscarbenjamin: Maybe this could be made as a change to the existing `hermite_normal_form` function rather than as a new `hnf` function.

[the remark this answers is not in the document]
oscarbenjamin: It would be good to improve HNF in general because HNF is the standard method to solve systems of linear diophantine equations which it would be good if SymPy supported (#20682).

sympy/core/numbers.py
```diff
@@ -438,6 +439,342 @@ def ilcm(*args):
...
+    ==========
+
+    .. [1] https://projecteuclid.org/euclid.em/1048515660/
+    .. [2] https://rosettacode.org/wiki/Diophantine_linear_system_solving
+    """
+    if not iterable(s):
+        raise ValueError('s should be iterable')
+    if len(s) < 2:
+        raise ValueError('expecting 2 or more values')
+    if not all(type(i) is int for i in s):
```
smichr, line 575: must assert that `i !=0`

sympy/core/numbers.py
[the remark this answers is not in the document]
skieffer: The differences we're seeing between `hnf` and `hermite_normal_form` are all due to conflicting definitions. The HMM paper calls it "row Hermite normal form". Cohen prefers to work by columns.

To make `hermite_normal_form` give the same results as `hnf` on a given matrix `M`, do this:
1. Rotate `M` 90 degrees counterclockwise
2. Pass through `hermite_normal_form`
3. Rotate 90 degrees clockwise and flip top to bottom.

I'm seeing the makings of a little math paper for the American Math Monthly. : )
"Hermite Normal Form under Matrix symmetries of the Dihedral Group D4" LOL

In particular it's kind of fascinating to compare the middle columns of the two outputs in your first example: you got squares, and I got first powers! What's going on there?!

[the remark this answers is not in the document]
skieffer: On a more practical note, I suppose what this means for SymPy is that it should have:

1. At the core, the best/fastest HNF algorithm, which is probably this new one.
2. The ability to perform all necessary flips and rotations on matrices, to accommodate users working within any set of conventions as to what "HNF" means.
3. Maybe some convenience methods that put (1) and (2) together for known cases.

I don't know how many different sets of conventions are out there, but maybe for now it's enough to have functions called `row_hnf`, which could be a synonym for this new function, while `column_hnf` would recreate the existing `hermite_normal_form` by composing the right flips and rotations with `row_hnf`.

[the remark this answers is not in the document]
skieffer: > Maybe this could be made as a change to the existing hermite_normal_form function rather than as a new hnf function.

Yes, I would also tend to think this still belongs under `polys.matrices` instead of `core.numbers`.

skieffer:
I'm getting a crash on the following test case:

```python
In [1]: from sympy.core.numbers import hnf

In [2]: G = [[5, 3, 2], [13, 11, 7], [23, 19, 17], [37, 31, 29], [47, 43, 41]]

In [3]: hnf(G)
AssertionError
```

smichr:
> I'm getting a crash on the following test case:

ok, so it *does* involve non-exact division; I removed the check

skieffer:
Here's a complete set of comparison tools, and some tests:

```python
from sympy import ZZ, Matrix, pretty_print
from sympy.polys.matrices.domainmatrix import DM, DomainMatrix
from sympy.polys.matrices.normalforms import hermite_normal_form as Cohen
from sympy.core.numbers import hnf as HMM_base

def ll_trunc(ll):
    """Chop off rows of zeros at bottom of list of lists"""
    c = ll[:]
    while c[-1] == [0] * len(c[-1]):
        c.pop()
    return c

HMM = lambda ll: ll_trunc(HMM_base(ll)[1])

# convert list of lists <--> domain matrix
ll2dm = lambda ll: DM(ll, ZZ)
dm2ll = lambda m: [[int(n) for n in row] for row in m.to_list()]

# flips and rotations
H = lambda m: DomainMatrix(list(reversed(m.to_list())), m.shape, m.domain)
T = lambda m: m.transpose()
R90cc = lambda m: H(T(m))
R90cw = lambda m: T(H(m))
T_ = lambda m: R90cc(H(m))

HMM_from_Cohen = lambda ll: dm2ll(T_(Cohen(R90cc(ll2dm(ll)))))
Cohen_from_HMM = lambda m: T_(ll2dm(HMM(dm2ll(R90cw(m)))))

G1 = Matrix(10, 10, lambda i, j: (i + 1) ** 3 * (j + 1) ** 2 + i + j + 2)
G2 = Matrix([[98, 23, 61], [71, 45, 78], [20, 60, 16]])
G3 = Matrix([[2, 7, 17, 29, 41], [3, 11, 19, 31, 43], [5, 13, 23, 37, 47]])
G4 = Matrix([[5, 3, 2], [13, 11, 7], [23, 19, 17], [37, 31, 29], [47, 43, 41]])

for G in [G1, G2, G3, G4]:
    L = G.tolist()
    D = DomainMatrix.from_Matrix(G, 'dense')
    assert HMM(L) == HMM_from_Cohen(L)
    assert Cohen(D) == Cohen_from_HMM(D)
```

oscarbenjamin:
Also here's a comparison of timings for the two implementations with larger matrices:
```python
In [29]: for n in [1, 2, 5, 10, 20, 50, 100, 200, 500, 1000]:
    ...:     print(f'{n} x {n}')
    ...:     G1 = Matrix(n, n, lambda i, j: (i + 1) ** 3 * (j + 1) ** 2 + i + j + 2)
    ...:     t = time.time(); Cohen(G1._rep); t1 = time.time() - t
    ...:     t = time.time(); HMM_base(G1._rep.to_list()); t2 = time.time() - t
    ...:     print('%.2f %.2f\n' % (t1*1000, t2*1000))
    ...: 
1 x 1
0.17 0.08

2 x 2
0.19 0.23

5 x 5
0.51 1.02

10 x 10
0.71 4.43

20 x 20
1.37 13.44

50 x 50
7.13 118.96

100 x 100
29.11 675.47

200 x 200
133.79 4288.29

500 x 500
823.70 52296.60

...
```

oscarbenjamin:
I don't know whether the extra expense comes from computing the unimodular matrix but it is definitely good to compute that.

skieffer:
The current `hermite_normal_form` is Cohen's Alg 2.4.5. It looks like his Alg 2.4.10 is almost exactly the same except that it also computes the unimodular matrix. Might be worth trying this.

skieffer:
Looking more closely, Cohen makes a remark after Alg 2.4.10 that the coefficients in the unimodular matrix `U`  tend to be large, and that some LLL-type reductions are advisable. So I suspect the new algorithm is the right one when `U` is needed.

skieffer:
Yup, the coeffs in 2.4.10 get huge:

```python
    for n in [1, 2, 5, 10, 20, 50, 100, 200, 500, 1000][:8]:
        print(f'{n} x {n}')
        G1 = Matrix(n, n, lambda i, j: (i + 1) ** 3 * (j + 1) ** 2 + i + j + 2)

        t = time.time(); H, U = _hermite_normal_form(G1._rep); t1 = time.time() - t
        t = time.time(); c, H, P = hnf(G1._rep.to_list()); t2 = time.time() - t

        L1 = log(max([abs(m) for m in U.to_list_flat()]))
        L2 = log(max(max([abs(m) for m in row]) for row in P))

        print('            time: %10.2f %10.2f' % (t1 * 1000, t2 * 1000))
        print('log of max coeff: %10.2f %10.2f\n' % (L1, L2))

1 x 1
            time:       0.08       0.02
log of max coeff:       0.00       0.00

2 x 2
            time:       0.80       0.22
log of max coeff:       3.58       2.40

5 x 5
            time:       0.28       0.42
log of max coeff:       8.46       3.53

10 x 10
            time:       0.69       2.05
log of max coeff:      11.56       2.71

20 x 20
            time:       2.09      11.77
log of max coeff:      14.48       2.08

50 x 50
            time:      12.62      88.99
log of max coeff:      18.23       1.10

100 x 100
            time:      54.45     519.66
log of max coeff:      21.03       0.69

200 x 200
            time:     205.49    3084.59
log of max coeff:      23.82       0.69
```

smichr:
This water is too deep for me. Someone (@skieffer ?) please take over.

skieffer:
Where did we leave off with this? I guess there are two questions:

* correctness of the algorithm, and
* integration with SymPy.


## Correctness

It looks like it was basically working, but still producing a slightly different transformation matrix than the one shown in the paper?

I haven't read the paper, but it looks like you are "tuning" it somehow by setting a value for `alpha`?

For the sake of confirmation, I guess we do want to see the exact same result as in the paper (for some value of `alpha`).

On the other hand, if `P*G = H` and the coefficients of `P` are very small -- which they are, in your examples -- to me that basically sounds like success.


## Integration

I guess it would be something like this:

* Put the algorithm in `polys.matrices.normalforms`, and make it accept a `DomainMatrix`.

* When the user DOES want the unimodular transformation matrix, then we use this algorithm.

* When they DON'T want it, we use the existing algorithm, since it is much faster.

oscarbenjamin:
This seems to be a faster way of getting the unimodular matrix:
```python
from sympy.matrices.normalforms import hermite_normal_form as hnf_old

def hnf_transform(A):
    A = Matrix(A)
    m, n = A.shape

    A2 = Matrix.vstack(eye(n), A)
    H2 = hnf_old(A2)

    U, H = H2[:n,:], H2[n:,:]
    assert A*U == H
    return U, H
```
This gives the inverse of the unimodular matrix compared to what is in this PR which would satisfy `U*A == H` rather than `A*U == H`. Here are timings of the simple version above, the version in this PR and flint's much faster hnf function for comparison:
```python
In [106]: for n in [2, 5, 10, 15, 20, 25, 30]:
     ...:     print('\nn =', n)
     ...:     A = randMatrix(n)
     ...:     print('hnf_transform:')
     ...:     %time U, H = hnf_transform(A)
     ...:     assert A*U == H
     ...:     print('hnf (PR):')
     ...:     %time U, H = hnf(A)
     ...:     assert U*A == H
     ...:     fA = to_flint(A)
     ...:     print('flint.hnf:')
     ...:     %time H, U = fA.hnf(transform=True)
     ...:     assert U*fA == H
     ...: 

n = 2
hnf_transform:
CPU times: user 863 µs, sys: 1e+03 ns, total: 864 µs
Wall time: 872 µs
hnf (PR):
CPU times: user 3.78 ms, sys: 876 µs, total: 4.66 ms
Wall time: 3.99 ms
flint.hnf:
CPU times: user 40 µs, sys: 1e+03 ns, total: 41 µs
Wall time: 46 µs

n = 5
hnf_transform:
CPU times: user 2.05 ms, sys: 29 µs, total: 2.08 ms
Wall time: 2.15 ms
hnf (PR):
CPU times: user 31.8 ms, sys: 398 µs, total: 32.2 ms
Wall time: 32.8 ms
flint.hnf:
CPU times: user 67 µs, sys: 1 µs, total: 68 µs
Wall time: 73.2 µs

n = 10
hnf_transform:
CPU times: user 7.87 ms, sys: 274 µs, total: 8.14 ms
Wall time: 8.45 ms
hnf (PR):
CPU times: user 225 ms, sys: 2.76 ms, total: 228 ms
Wall time: 230 ms
flint.hnf:
CPU times: user 227 µs, sys: 1e+03 ns, total: 228 µs
Wall time: 235 µs

n = 15
hnf_transform:
CPU times: user 20.3 ms, sys: 189 µs, total: 20.5 ms
Wall time: 20.7 ms
hnf (PR):
CPU times: user 809 ms, sys: 5.22 ms, total: 814 ms
Wall time: 819 ms
flint.hnf:
CPU times: user 706 µs, sys: 1 µs, total: 707 µs
Wall time: 713 µs

n = 20
hnf_transform:
CPU times: user 46.4 ms, sys: 336 µs, total: 46.8 ms
Wall time: 47.1 ms
hnf (PR):
CPU times: user 1.82 s, sys: 7.28 ms, total: 1.83 s
Wall time: 1.84 s
flint.hnf:
CPU times: user 1.39 ms, sys: 13 µs, total: 1.41 ms
Wall time: 1.41 ms

n = 25
hnf_transform:
CPU times: user 93.4 ms, sys: 520 µs, total: 93.9 ms
Wall time: 94.2 ms
hnf (PR):
CPU times: user 5.03 s, sys: 60.8 ms, total: 5.09 s
Wall time: 5.48 s
flint.hnf:
CPU times: user 1.81 ms, sys: 2 µs, total: 1.81 ms
Wall time: 1.82 ms

n = 30
hnf_transform:
CPU times: user 854 ms, sys: 14.9 ms, total: 869 ms
Wall time: 906 ms
hnf (PR):
CPU times: user 11.4 s, sys: 101 ms, total: 11.5 s
Wall time: 11.8 s
flint.hnf:
CPU times: user 4.45 ms, sys: 22 µs, total: 4.47 ms
Wall time: 4.48 ms
```

oscarbenjamin:
Apparently flint uses the ("naive") algorithm I showed above when the matrix does not have full column rank:
https://github.com/wbhart/flint2/blob/268e499cae98dcd3b1e518cf13d92929e50f6cab/fmpz_mat/hnf_transform.c#L51

If the number of rows in the RREF is equal to the number of columns (as most likely the case in the timings above) then a method based on minors is used:
https://github.com/wbhart/flint2/blob/268e499cae98dcd3b1e518cf13d92929e50f6cab/fmpz_mat/hnf_minors_transform.c#L15-L19

That's apparently based on the paper "Polynomial Algorithms for Computing the Smith and Hermite Normal Forms of an Integer Matrix, Kannan and Bachem": https://epubs.siam.org/doi/10.1137/0208040

skieffer:
> a faster way of getting the unimodular matrix

I like this simplicity of this approach, but I think it's basically the same as the Alg 2.4.10 that I tried earlier, and the drawback of this method is that the coeffs in the unimodular matrix `U` can be enormous.

Maybe the user should have a choice:
* If they want the matrix `U`, and want it quickly, and don't care about the size of the coeffs, this method should be used.
* If they want small coeffs and are willing to wait, they need the LLL-based algorithm of this PR.

skieffer:
I wonder to what extent Flint's speed is a function of its choice of algorithms, or a matter of C++ vs. Python, or both.

oscarbenjamin:
> I wonder to what extent Flint's speed is a function of its choice of algorithms, or a matter of C++ vs. Python, or both.

It's all of these things. Flint uses C rather than Python so for something low-level like multiplying two integer matrices you can expect that an optimised Python implementation might be something like 100x slower than an optimised C implementation. The actual ratio depends how much time is spent in large integer operations though because if the time is all spent in expensive calls to GMP then it's basically the same for both.

However it is absolutely the case that flint uses better algorithms for most of these things than sympy does.

> and the drawback of this method is that the coeffs in the unimodular matrix `U` can be enormous.

And yet according to my timings it's always worked out faster than what we have here in this PR. Did I not time the right cases?

> I like the simplicity of this approach

The advantage is that the implementation becomes faster if we improve the ordinary hnf function and there's fewer places to fix bugs. This PR adds a lot of code for something that can actually be achieved quite easily.

skieffer:
> according to my timings it's always worked out faster

Absolutely! It's nice and quick, which is why again I think we should give the user the option. For some users, speed might be paramount, and they don't care what the coeffs look like. In that case this algorithm is great. But, for anyone who wants to visually inspect or display the matrix, forget it. The coeffs have 100 digits.

smichr:
Looks like a different LLL was committed in #24572.

PR #22850, opened 2022-01-12
Make `evalf_table` entry for `AlgebraicNumber`

### LINKED ISSUE #22849: `evalf_table` missing entry for `AlgebraicNumber`

skieffer:

Fixes #22849 

Improves `evalf()` on instances of `AlgebraicNumber`.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -2681,7 +2681,7 @@ class AlgebraicNumber(Expr):
         ...                  [-1, -1, 0, 0], alias=zeta)
         >>> a5.as_poly().as_expr()
         -zeta**3 - zeta**2
-        >>> a5.evalf()
+        >>> a5.evalf(chop=True)
         1.61803398874989
 
         (The index ``-1`` to ``CRootOf`` selects the complex root with the
--- a/sympy/core/evalf.py
+++ b/sympy/core/evalf.py
@@ -1403,7 +1403,7 @@ def _create_evalf_table():
     from .add import Add
     from .mul import Mul
     from .numbers import Exp1, Float, Half, ImaginaryUnit, Integer, NaN, NegativeOne, One, Pi, Rational, \
-        Zero, ComplexInfinity
+        Zero, ComplexInfinity, AlgebraicNumber
     from .power import Pow
     from .symbol import Dummy, Symbol
     from sympy.functions.elementary.complexes import Abs, im, re
@@ -1427,6 +1427,8 @@ def _create_evalf_table():
         NegativeOne: lambda x, prec, options: (fnone, None, prec, None),
         ComplexInfinity: lambda x, prec, options: S.ComplexInfinity,
         NaN: lambda x, prec, options: (fnan, None, prec, None),
+        AlgebraicNumber: lambda x, prec, options: evalf(x.as_expr(),
+                                                        prec, options),
 
         exp: evalf_exp,
 
--- a/sympy/core/tests/test_evalf.py
+++ b/sympy/core/tests/test_evalf.py
@@ -7,7 +7,7 @@ from sympy.core.evalf import N
 from sympy.core.function import (Function, nfloat)
 from sympy.core.mul import Mul
 from sympy.core import (GoldenRatio)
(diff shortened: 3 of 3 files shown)


jksuom:
Would it suffice to define `AlgebraicNumber._eval_evalf`?

sympy/core/numbers.py
```diff
@@ -2681,7 +2681,7 @@ def __new__(cls, expr, coeffs=None, alias=None, **args):
         ...                  [-1, -1, 0, 0], alias=zeta)
         >>> a5.as_poly().as_expr()
         -zeta**3 - zeta**2
-        >>> a5.evalf()
+        >>> a5.evalf(chop=True)
         1.61803398874989
```
oscarbenjamin, line 2685: Why is this change needed?

skieffer:
> Would it suffice to define `AlgebraicNumber._eval_evalf`?

Good question. There actually is an `AlgebraicNumber._eval_evalf()`, and I was going to modify it, but I thought that extending the table would be the more straightforward approach.

I'm new to the whole evalf system -- so, apologies if my attempt to write about it comes out as spaghetti -- but it seems like there are two sides to it: there are the functions defined in `core.evalf`, which return quadruples `(re, im, re_acc, im_acc)`, and then there are the `.eval_evalf()` methods, which return `Float`. That second system is referred to here

https://github.com/sympy/sympy/blob/c9075243704ee133a2ab5f1ecda6fae513e21e97/sympy/core/evalf.py#L1490-L1500

as "ordinary evalf". Here we first try to use a "quadruple function" from the lookup table, and fall back on "ordinary evalf" if there is none.

For an `AlgebraicNumber`, say

```python
a = -8 + 3 * sqrt(3)
x = AlgebraicNumber(a)
```

we're always going to start with `y = x.to_expr()`, and then we're going to evalf `y`. The only question is which of the two systems -- "quads" or "Floats" -- are we going to use to evalf `y`.

Currently, `evalf(x)` does lead to `AlgebraicNumber._eval_evalf()`:

https://github.com/sympy/sympy/blob/c9075243704ee133a2ab5f1ecda6fae513e21e97/sympy/core/numbers.py#L2778-L2779

where we use "ordinary evalf" on `y`. When `y` is an `Add` (which it usually will be), and the precision is low, we can wind up with total cancelation, and return `0`.

To prevent this, we need to use the `evalf_add()` function instead (which lives on the "quad" side of the system). And I think giving `AlgebraicNumber` an entry in the lookup table is the most straightforward way to do this.

sympy/core/numbers.py
skieffer: Another good question. It started printing as

```
1.61803398874989 + 0.e-20*I
```

and I don't really understand why. Maybe the `0.e-20*I` used to be eliminated by the very same cancelation I'm trying to get rid of.

oscarbenjamin:
The evalf stack is a bit of a mess. The `_eval_evalf` methods don't have any way of passing round any of the other options besides the precision (e.g. `maxprec`).

I would have thought that it was better to evalf an AlgebraicNumber from its minpoly e.g. to guarantee getting a real number if it is real.

skieffer:
> evalf an AlgebraicNumber from its minpoly e.g. to guarantee getting a real number if it is real.

I like this idea! This will solve the issue above, with `a5.evalf()` printing as

```
1.61803398874989 + 0.e-20*I
```

But to do it I need `AlgebraicNumber.to_root()` to be in good shape, and on this branch it's a mess. So I'd like to actually close this PR, and instead solve #22849 in #22737.

PR #22861, opened 2022-01-14
Document build artifacts

### LINKED ISSUE #21829: Upload built docs for on PR CI builds
If you look at the matplotlib CI, they have a special build that builds their docs and uploads them somewhere. For example, go to some matplotlib PR https://github.com/matplotlib/matplotlib/pulls and click the "Check the rendered docs here!" build in the checks section at the bottom. 

I think they are using a combination of GitHub Actions and Circle CI. CC Matplotlib devs @tacaswell @story645 @Qulogic. How hard would it be to just copy what matplotlib is using to SymPy?

oscargus:

Closes #21829

The idea is to upload build artifacts (html and pdf) for each PR. We'll how that works out...

#### Release Notes

NO ENTRY

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -277,7 +277,7 @@ jobs:
       - uses: actions/upload-artifact@v2
         with:
           name: html
-          path: doc/html/
+          path: doc/_build/html/
       - uses: actions/upload-artifact@v2
         with:
           name: pdf
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -274,6 +274,15 @@ jobs:
       - run: release/aptinstall.sh
       - run: pip install -r doc/requirements.txt
       - run: bin/test_sphinx.sh
+      - uses: actions/upload-artifact@v2
+        with:
+          name: html
+          path: html/
+      - uses: actions/upload-artifact@v2
+        with:
+          name: pdf
+          path: _build/latex/sympy*.pdf
+
 
 
   # -------------------- Check the error message under py2 --------- #
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -277,11 +277,11 @@ jobs:
       - uses: actions/upload-artifact@v2
         with:
           name: html
-          path: html/
+          path: doc/html/
       - uses: actions/upload-artifact@v2
         with:
           name: pdf
-          path: _build/latex/sympy*.pdf
+          path: doc/_build/latex/sympy*.pdf
 
 
 


oscarbenjamin:
We should consider what this means in terms of the storage allocation that GA gives for free.

oscargus:
Indeed, probably set a quite short life span for it. Just wanted to see if
and how it worked.

oscarbenjamin:
We could add something to the comment on pr workflow that pushes the built docs to a special sympy/pullrequestdocs repo so that you could go to say sympy.github.com/io/pullrequestsdoc/12345 to see the build docs.

.github/workflows/runtests.yml
```diff
@@ -274,6 +274,15 @@ jobs:
       - run: release/aptinstall.sh
       - run: pip install -r doc/requirements.txt
       - run: bin/test_sphinx.sh
+      - uses: actions/upload-artifact@v2
+        with:
+          name: html
+          path: html/
```
oscargus, line 280:
```suggestion
          path: doc/html/
```

.github/workflows/runtests.yml
```diff
@@ -274,6 +274,15 @@ jobs:
...
+      - uses: actions/upload-artifact@v2
+        with:
+          name: pdf
+          path: _build/latex/sympy*.pdf
```
oscargus, line 284:
```suggestion
          path: doc/_build/latex/sympy*.pdf
```

.github/workflows/runtests.yml
```diff
@@ -274,6 +274,15 @@ jobs:
       - run: release/aptinstall.sh
       - run: pip install -r doc/requirements.txt
       - run: bin/test_sphinx.sh
+      - uses: actions/upload-artifact@v2
+        with:
+          name: html
+          path: doc/html/
```
oscargus, line 280:
```suggestion
          path: doc/_build/html/
```

oscargus:
OK! So the artifacts are moved and available as zip-files which is not ideal. Also, the size is quite massive...

One should probably look more in detail into how e.g. matplotlib solves this. There is a direct link in the list of tests that one can use to access the built docs.

oscarbenjamin:
> Also, the size is quite massive...

This is why I didn't do this already. If the artifacts are created once per push to any PR and each generates a 50MB html file then the 1GB storage allocation will run out within a day.

> One should probably look more in detail into how e.g. matplotlib solves this.

I don't know how they do it but my suggestion above should work and sounds straight-forward to me. Basically in the comment on pr workflow add a job that does something like:
```
git clone github.com/sympy/pullrequestdocs
cd pullrequestdocs
rm -r $PRNUMBER
unzip $ZIPFILE $PRNUMBER
git add $PRNUMBER
git push
```
Then the idea is that the pullrequestdocs repo has subfolders corresponding to each PR number and so you can access the docs for the PR 12345 from GH pages with
```
sympy.github.io/pullrequestdocs/12345
```
The only bit I'm not sure about is the permissions needed to push from one repo to another in GA. The comment on pr workflow is a separate workflow file because it needed to run with different permissions to be able to comment on the PR. Probably the same would apply here.

oscargus:
Closing as #23465 seems better.

PR #22866, opened 2022-01-15
fixed checksol causing error when passing int or float as equation

NikhilSDate:
`checksol` does not cause an `AttributeError` anymore if an `int` or `float` is passed as one of the equations. 
For example, currently on master, `checksol(0, {})` produces `AttributeError: 'int' object has no attribute 'is_Relational'`. 
After this change, `checksol(0, {})` returns `True`.

A few tests are also added to `test_checksol` to test this behavior.

#### Release Notes

* solvers
  * checksol no longer produces an error if ints or floats are passed as equations


sympy/solvers/solvers.py
```diff
@@ -272,6 +272,10 @@ def checksol(f, symbol, sol=None, **flags):
 
     if isinstance(f, BooleanAtom):
         return bool(f)
+    elif isinstance(f, int):
+        return f == 0
+    elif isinstance(f, float):
+        return abs(f) < 1e-9
```
oscarbenjamin, line 278: It would be better just to `_sympify` the input.

NikhilSDate: I tried that, but it is causing some weird errors that will have to be dealt with. For example, `checksol(0.0001, x, 0)` returns `None` instead of `False`.

NikhilSDate: I'll see if I can fix this though.

NikhilSDate: It seems that the following code is causing the problem I mentioned.
https://github.com/sympy/sympy/blob/9fcdf264f3310e82a7f1d159d69812cf2422dbd0/sympy/solvers/solvers.py#L278-L280
Fixing the issue will probably require an explicit type check for `Float`. Would that be better than checking if f is an `int` or a `float`?

oscarbenjamin: The same thing happens with `Integer`: it's not just float but also `pi`, `pi+1` etc. Think about the logic here: can you list all of the possible objects that you need to check for and call isinstance for all of them?

Maybe None is a reasonable return value here. Thinking about it the empty equation does not imply that`x=3` or anything else so maybe `checksol(1, x, 0)` should return `None` to indicate that `x=3` is not something that can be inferred from the equation. I'm not sure what the precise expected semantics of `checksol` should be but we shouldn't change anything without clarifying that first.

NikhilSDate: `Integer` incidentally gets caught by a previous condition and returns `True` if the equation is zero. So something like `checksol(Integer(0), x, 1)` returns `True`. Currently, the problem is that even something like `checksol(0.0, {})` will return `None` if `Float` is not checked for. So that needs to be fixed at least.
I would argue that, regardless of what symbols are passed in the solution, `checksol` should return `True` if the given solution satisfies all the equations in the sense that all equations can be simplified to 0 if the solution passed is substituted in and should return `False` if at least one of the equations is definitely non-zero after the solution is substituted (for example something like `checksol(pi+1, x, 0)`. If that's the case, that entire `if` condition can even be removed entirely. I've tried removing that `if` block and all tests in the solvers module seem to pass so it does not look like that behavior is being tested anywhere. 
Maybe someone else with a better idea of how `checksol` should work can help here? I'm not sure who that would be.
In any case, we could maybe push in the change where `f` is sympified so that passing an `int` or a `float` at least stops producing an `AttributeError` and I could open another PR for potentially changing the behavior for equations that have no free symbols in common with the solution keys.

NikhilSDate: > maybe checksol(1, x, 0) should return None to indicate that x=3 is not something that can be inferred from the equation.

Also, I think this logic could cause some inconsistency issues with stuff like `checksol(x - y, {x: 0, y: 0})` (which currently returns `True` even though it is impossible to infer that `x=0` and `y=0` from `x-y=0`) or even `checksol(x, {x: 0, y: 0)` (which also returns `True` even though `y=0` cannot be inferred from the given equation. I think making `checksol` return output based on what can be definitely inferred from the equation would likely require major changes and would also change a lot of existing behavior. Any thoughts on this?

smichr:
> even checksol(x, {x: 0, y: 0) (which also returns True even though y=0 cannot be inferred from the given equation

I advise against creating cases for which `checksol` was not expected to adjudicate. Given an equation and a test solution, `checksol` should indicate if the solution makes the equation 0. The intent is not to tell whether the solution can be inferred from the equation.

sympy/solvers/solvers.py
smichr:
```python
if iterable(f):
    ...
f = _sympify(f)
if not f.free_symbols:
    return f.is_zero
...
```

> Maybe someone else with a better idea of how checksol should work can help here? I'm not sure who that would be.

Look at the git blame and see who has worked on this routine.

NikhilSDate:
> The intent is not to tell whether the solution can be inferred from the equation.

Thanks for the clarification. This would mean that if the equation is just `0` (for example `checksol(0, x, 2)`, then `checksol` should return `True` no matter what the solution passed is, right?

smichr:
> no matter what the solution passed is, right?

yes.

NikhilSDate:
I see. The current changes in this PR provide this behavior then. Also, if we have something like `checksol(pi, x, 2)`, this should then return `False` by the same logic right? Currently on master, this returns `None`.

smichr:
> same logic right?

yes

NikhilSDate:
@smichr @oscarbenjamin Do you think cases such as `checksol(y, x, 2)` returning `None` (the current behavior on master) makes sense? I think this causes an inconsistency with something like `checksol(x, x, y) `(which returns `False` on master), because both result in the same expression after the solution is substituted in but still have different return values.

If this is indeed an inconsistency, the following lines can be entirely removed to fix the inconsistent behavior. This change along with sympifying the input will also fix the original issue that this PR addresses. 
https://github.com/sympy/sympy/blob/9fcdf264f3310e82a7f1d159d69812cf2422dbd0/sympy/solvers/solvers.py#L278-L280

NikhilSDate:
@oscarbenjamin @smichr Could you give my comment above a look? This PR is stuck unless I get some feedback on the question I raised in that comment.

oscarbenjamin:
> the following lines can be entirely removed

That seems reasonable I think.

sympy/solvers/solvers.py
```diff
@@ -360,12 +357,12 @@ def checksol(f, symbol, sol=None, **flags):
                 return not nz
             break
 
+        if numerical and val.is_number:
```
NikhilSDate, line 360: These `if` conditions have been reordered to prevent the `if val==was` condition preventing the numerical check from executing in cases where the equation is a number. Before this change, stuff like `checksol(1e-10, ...)` would return `False` instead of `True` because the numerical check would never execute.

sympy/solvers/solvers.py
```diff
@@ -360,12 +357,12 @@ def checksol(f, symbol, sol=None, **flags):
...
+            return (abs(val.n(18).n(12, chop=True)) < 1e-9) is S.true
         if val == was:
             continue
         elif val.is_Rational:
             return val == 0
-        if numerical and val.is_number:
-            return (abs(val.n(18).n(12, chop=True)) < 1e-9) is S.true
```
NikhilSDate, line 368: These `if` conditions have been reordered to prevent the `val==was` condition preventing the numerical check from executing in cases where the equation is a number. Before this change, stuff like `checksol(1e-10, ...)` would return `False` instead of `True` because the numerical check would never execute.

oscarbenjamin:
> Is there anything I need to do here?

You are already in the .mailmap and AUTHORS file as:
```
Nikhil Date <nikhil.s.date@gmail.com>
```
You git commits in this PR are recorded as:
```
NikhilSDate <nikhil.s.date@gmail.com>
```
This comes from your git config:
https://github.com/sympy/sympy/wiki/Development-workflow#configure-git-settings

You need to change that line in .mailmap so it says:
```
Nikhil Date <nikhil.s.date@gmail.com> NikhilSDate <nikhil.s.date@gmail.com>
```
That just means that all commits from the second entry are from the same person as the first and that the first entry is the one that should be recorded in the AUTHORS file (as it already is).

oscarbenjamin:
Actually it looks like there should be two entries in .mailmap:
```
Nikhil Date <nikhil.s.date@gmail.com> NikhilSDate <47920034+NikhilSDate@users.noreply.github.com>
Nikhil Date <nikhil.s.date@gmail.com> NikhilSDate <nikhil.s.date@gmail.com>
```

NikhilSDate:
@oscarbenjamin Thanks for the help with the `.mailmap` file. Is this ready to be merged now?

NikhilSDate:
@oscarbenjamin is there something holding this PR up? Are more reviews necessary for this to go in? I have another PR I'm working on that needs this to be merged first so I'm a bit stuck till there is any more progress on this.

oscarbenjamin:
It looks okay to me but I don't know this code so well so I'm hoping that someone else (e.g. @smichr might review it.

NikhilSDate:
@smichr Could you give this a look. This is stuck unless it gets some additional reviews from someone familiar with code.

NikhilSDate:
ping @smichr for review

NikhilSDate:
@oscarbenjamin since this PR hasn't yet gotten any more reviews, I'm not sure how long it usually takes for stuff like this to get reviewed. How much longer should I expect to wait?

sympy/solvers/tests/test_solvers.py
```diff
@@ -1310,6 +1310,10 @@ def test_checksol():
     assert checksol([x - 1, x**2 - 1], x, 1) is True
     assert checksol([x - 1, x**2 - 2], x, 1) is False
     assert checksol(Poly(x**2 - 1), x, 1) is True
+    assert checksol(0, {}) is True
+    assert checksol([1e-10, x - 2], x, 2) is True
```
oscarbenjamin, line 1314: Ideally this should fail.

NikhilSDate: Which assertion are you referring to exactly?

oscarbenjamin: The one immediately above with `1e-10` which is not zero.

NikhilSDate: The reason I added that test is because `checksol` seems to treat equations with numeric absolute values less than `1e-9` (after the solution has been substituted in) as zero when checking solutions numerically. On master, something like `checksol(x+1e-10, x, 0)` returns `True`. I added this to test the same behavior for when the equation is just a `float`.

oscarbenjamin: I can see the potential merit of doing this when there is a need to substitute something in and numerically evaluate but I think if the input is a plain number then it should just be `return num == 0`.

sympy/solvers/solvers.py
```diff
@@ -274,10 +275,6 @@ def checksol(f, symbol, sol=None, **flags):
     elif not f.is_Relational and not f:
         return True
 
-    if sol and not f.free_symbols & set(sol.keys()):
```
smichr, line 277: why was this removed?

sympy/solvers/tests/test_solvers.py
smichr: Agreed. The function is checking a solution. Such non-equations would/should have been eliminated before sending here. If there isn't a test like `assert checksol(x+1e-10, x, 0, numerical=True) == True` and `assert checksol(x+1e-10, x, 0, numerical=False) == False` then you could add those.

sympy/solvers/solvers.py
NikhilSDate: It was causing inconsistency issues. To refer to my previous comment above, 
> @smichr @oscarbenjamin Do you think cases such as checksol(y, x, 2) returning None (the current behavior on master)  makes sense? I think this causes an inconsistency with something like checksol(x, x, y) (which returns False on master), > because both result in the same expression after the solution is substituted in but still have different return values.
If this is indeed an inconsistency, the following lines can be entirely removed to fix the inconsistent behavior. This  change along with sympifying the input will also fix the original issue that this PR addresses.
https://github.com/sympy/sympy/blob/9fcdf264f3310e82a7f1d159d69812cf2422dbd0/sympy/solvers/solvers.py#L278-L280

sympy/solvers/tests/test_solvers.py
NikhilSDate: OK. I'll make the change in `checksol` and add the tests you suggest if they don't already exist.

sympy/solvers/solvers.py
smichr: Perhaps a better thing to do is to test the input against expectation: `not f.free_symbols - set(sol.keys())`. It is supposed to be answering False if we know that a solution does not reduce the expression to 0, None if it can't be determined, and True if it does reduce to zero. The case for args `(y, x, 2)` is an error while the case of `(x, x, y)` is False.

NikhilSDate: I'm not sure I understand you correctly. Are you suggesting that `not f.free_symbols & set(sol.keys())` in the if condition be replaced with `f.free_symbols - set(sol.keys())`? This would not work correctly if the args are something like `(x*y, x, 0)`.
> The case for args (y, x, 2) is an error.

Do you mean that the `None` result for `(y, x, 2)` is incorrect and should be `False` instead?

smichr: I am getting the feeling that we are turning something with a specific use case in mind (though not clearly stated) into something more general that we shouldn't (won't) need. Let's see if the following clarified possibilities make sense:

1. if a solution sets an equation to 0, return True
```
(x*y, x, 0) -> True
```
2. if a solution sets an equation to an expression that is not zero, return False
```
(cos(x)**2+sin(x)**2, x, 1) -> False
(y, x, 0) -> False (** but see below)
(x*y, x, 2) -> False
```
4.  if a solution sets an equation to a  number that can't be distinguished from zero, return None
```
(cos(x)**2+sin(x)**2 - 1, x, 1) -> None
```
(**) For the second example of 2, we might consider this a None case since there is no symbol to replace for the test to be done. But it could be the case that such an expression is just an unsimplified symbolic 0 (i.e. replace `y` with `f(y)` such that `f(y) != 0 and f(y).simplify() == 0`) so we shouldn't return False/None until the required check has been done.

NikhilSDate: I agree with you about points 1 and 2. but I'm not sure about the example for point 4. Currently, `checksol(cos(x)**2+sin(x)**2-1, x, 1)` returns `True` because it checks the solution numerically and finds it to be sufficiently close to zero. Even if `numerical` is set to `False`, `simplify` is able to simplify `sin(1)**2+cos(1)**2-1` to 0, so `checksol` still returns `True`, which is mathematically correct. Am I understanding you correctly on point 4?
Also, entirely removing the `if sol and not f.free_symbols & set(sol.keys()):` check makes `checksol` work as you describe in points 1 and 2, so I think it's a good idea to remove it. Do you have any other suggestions in mind here?

sympy/solvers/tests/test_solvers.py
NikhilSDate: I've gone with your suggestion about returning `f.is_zero` if `f.free_symbols` is empty, and I've added the tests you suggested, so I'm resolving this.

sympy/solvers/solvers.py
```diff
@@ -256,6 +256,11 @@ def checksol(f, symbol, sol=None, **flags):
             rv = None  # don't return, wait to see if there's a False
         return rv
 
+    f = _sympify(f)
+
+    if not f.free_symbols:
+        return f.is_zero
```
smichr, line 262: Since the routine will try different strategies, shouldn't this be allowed to continue? It would only be safe to do this `if f.is_Number:`

sympy/solvers/tests/test_solvers.py
```diff
@@ -1310,6 +1310,12 @@ def test_checksol():
     assert checksol([x - 1, x**2 - 1], x, 1) is True
     assert checksol([x - 1, x**2 - 2], x, 1) is False
     assert checksol(Poly(x**2 - 1), x, 1) is True
+    assert checksol(0, {}) is True
+    assert checksol([1e-10, x - 2], x, 2) is False
```
smichr, line 1314: Since `numerical=True` is the default, why is this giving False?

NikhilSDate: If you're talking about the last assertion (`assert checksol([1e-10, x - 2], x, 2) is False`), I think we agreed (https://github.com/sympy/sympy/pull/22866#discussion_r802574214 and https://github.com/sympy/sympy/pull/22866#discussion_r802618221) that for equations that are just plain numbers, `checksol` should only return `True` if the equation is exactly zero. In this case the `1e-10` "equation" is not zero, so `checksol` returns `False`. Did I misunderstand something?

smichr: OK, that makes sense: no tolerance for a number that is already not zero. I don't think the case of no free symbols needs special handling -- that's one of those corner cases that I really don't envision this dealing with. So returning None instead of False or True seems ok.

sympy/solvers/solvers.py
smichr, line 262:
```suggestion
if not isintance(f, Basic):
    return f == 0
if f.is_number:
    return f.is_zero
```

sympy/solvers/solvers.py
```diff
@@ -256,6 +256,11 @@ def checksol(f, symbol, sol=None, **flags):
             rv = None  # don't return, wait to see if there's a False
         return rv
 
+    f = _sympify(f)
+
+    if not f.free_symbols:
```
smichr, line 261:
```suggestion
    if f.is_number:
```

NikhilSDate:
Thanks @smichr for merging. I agree with the last change you made.

PR #22869, opened 2022-01-17
combinatorial: Added code for Padovan number

faze-geek:
Added code for padovan numbers along with some other relevant methods in combinatorial numbers.
https://en.wikipedia.org/wiki/Padovan_sequence

#### Other comments
The motivation for this PR was #22831 , there are some generalizations of the fibonacci series which can be useful for SymPy users. I'll add others once this PR gets merged .

#### Release Notes

* functions
  * Added Padovan Number to combinatorial numbers.


sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2292,126 @@ def ok(x):
...
+    .. [2] https://mathworld.wolfram.com/PadovanSequence.html
+
+    """
+
+    @staticmethod
+    def is_padovan(n):
+        try:
+            n = as_int(n)
+        except ValueError:
+            False
```
NikhilSDate, line 2338: This should probably be `return False`

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2292,126 @@ def ok(x):
...
+
+    @classmethod
+    def eval(cls, n):
+        if n is S.Infinity:
+            return S.Infinity
+
+        if n in (0,1,2):
+            return 1
+        if n.is_Integer and n>0:
+            return padovan(n - 2) + padovan(n - 3)
```
NikhilSDate, line 2410: You might want to use memoization here like `tribonacci` uses instead of directly calling `padovan(n-2)` and `padovan(n-3)`. I'm not too familiar with this module so maybe someone else should confirm that this would be useful.

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2292,126 @@ def ok(x):
...
+
+        if n == -1:
+            return 0
+        if n == -2:
+            return 1
+        if n.is_Integer and n<0:
+            return padovan(n + 3) - padovan(n + 1)
```
NikhilSDate, line 2417: You might be able to use some form of memoization here as well.

sympy/functions/combinatorial/numbers.py
faze-geek: Oops, I'll change this as soon as possible  .

sympy/functions/combinatorial/numbers.py
faze-geek: > You might want to use memoization here like `tribonacci` uses instead of directly calling `padovan(n-2)` and `padovan(n-3)`. 

Yeah even I was split ,between using `padovan(n-2) + padovan(n-3) ` or creating a  recurrence memo . I'm not sure what works better but looking at the module it seems the typical approach is using memoization .

sympy/functions/combinatorial/numbers.py
faze-geek: Noted from the upper comment !

sympy/functions/combinatorial/numbers.py
NikhilSDate: > it seems the typical approach is using memoization .
Yes, that's what I thought as well. Memoization could probably end up being more efficient here. You could perhaps try timing both methods and see which one is actually faster.

faze-geek: @NikhilSDate
 I have added a commit with the recurrence memo .However I had trouble in writing a common recurrence memo for both, when n is positive and when n is negative (because the recurrence relation in the two is different ).Currently I have added code for when n is positive .Can you review it and also improvise one where n (when negative) case is added with it !!
Recurrence relations -
n is positive` f(n) = f(n-2) + f(n-3)`
n is negative `f(n) = f(n+3) - f(n+1)`

NikhilSDate: I think you will need a separate recurrence for the negative case.

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2292,149 @@ def ok(x):
...
+            while sn2 < n :
+                b = sn0 + sn1
+                sn0 = sn1
+                sn1 = sn2
+                sn2 = b
+
+            if Eq(sn2, n, evaluate=True):
+                return True
+            else:
+                return False
```
oscarbenjamin, line 2357: Both `sn2` and `n` are of type `int` if I understand this correctly. In that case you just do `return sn2 == n`.

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2292,149 @@ def ok(x):
...
+    def _padovan(n, prev):
+        if n>0:
+            return (prev[-3] + prev[-2])
+
+    @classmethod
+    def eval(cls, n):
+        try:
+            n = as_int(n)
+        except ValueError:
+            raise ValueError('The provided number must be a integer')
```
oscarbenjamin, line 2439: What if it's a symbol?

sympy/functions/combinatorial/numbers.py
faze-geek: But then wouldn't I have to define another name !? I believe it is not user friendly if I have to use different func names something like `padovan_pos(x)` for positive and `padovan_neg(x)` for negative.

faze-geek: > I think you will need a separate memoization for the negative case.

I will look into this !

sympy/functions/combinatorial/numbers.py
faze-geek: > Both `sn2` and `n` are of type `int` if I understand this correctly. In that case you just do `return sn2 == n`.

I will remove unnecessary code in a future commit !

sympy/functions/combinatorial/numbers.py
faze-geek: > What if it's a symbol?

Yeah it is a subclass of functions ,it should not only accept integers but symbols too!
How would you suggest to deal with this ,I'll make appropriate changes in the next commit !!

sympy/functions/combinatorial/numbers.py
NikhilSDate: You can do it similar to how it's handled in `tribonacci` where there is a private function that is memoized which is called by `eval`.  Here you could have two private functions called by `eval` for the two separate cases.

NikhilSDate:
@faze-geek I just noticed that the release notes entry for this PR is blank. Have you left it blank for any specific reason? I think this is a change that requires a release notes entry.

faze-geek:
> @faze-geek I just noticed that the release notes entry for this PR is blank. Have you left it blank for any specific reason? I think this is a change that requires a release notes entry.

Thanks for reminding !

sympy/functions/combinatorial/numbers.py
oscarbenjamin: It's enough to return `None` from here I think rather raising `ValueError`.

faze-geek: Thanks @oscarbenjamin .I have made the change you suggested .

sympy/functions/combinatorial/numbers.py
faze-geek: @oscarbenjamin  
Could you look into this .I am not sure whether two different recurrence memo's can be created expected to be working with the same defined name `padovan()` .I have attempted to handle it but that code only gives correct results for n>0 (positive n) .Rest all looks good, once this is solved I think this would be good to merge !

faze-geek: @praneethratna I would like some hint/assistance on this .I'm not sure if 2 separate recurrence memos can be used under the same name .You could refer to my comment above as reference for what I am trying to improve upon.

praneethratna: I'm also not sure of this but i think we cannot use two recurrence memos under the same name.

faze-geek: Padovan series for negative numbers is essential ,I would not like to neglect it .For now should I go with the earlier code only ?
 ```
if n in (0,1,2):
            return 1
 if n.is_Integer and n>0:
            return padovan(n - 2) + padovan(n - 3)
```

File: sympy/functions/combinatorial/numbers.py [hunk 6]
```diff
@@ -2292,3 +2292,156 @@ def ok(x):
...
+    .. [2] https://mathworld.wolfram.com/PadovanSequence.html
+
+    """
+
+    @staticmethod
+    def is_padovan(n):
+        try:
+            n = as_int(n)
+        except ValueError:
+            return False
```
czgdp1807, line 2340: Can we somehow use `if-else` here to check for an int?

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2292,156 @@ def ok(x):
...
+        if n>0:
```
czgdp1807, line 2341:
```suggestion
        if n > 0:
```

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2292,156 @@ def ok(x):
...
+            if n == 1 :
+                return True
+
+            sn0 = 1
+            sn1 = 1
+            sn2 = 1
```
czgdp1807, line 2347:
```suggestion
            sn0, sn1, sn2 = 1, 1, 1
```

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2292,156 @@ def ok(x):
...
+                sn0 = sn1
+                sn1 = sn2
+                sn2 = b
+
+            if sn2 == n:
+                return True
+            else:
+                return False
+
+        if n<=0:
```
czgdp1807, line 2359:
```suggestion
        if n <= 0:
```

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2292,156 @@ def ok(x):
...
+            if n in(0,-1):
```
czgdp1807, line 2360:
```suggestion
            if n in (0,-1):
```

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2292,156 @@ def ok(x):
...
+                return True
+
+            sn0 = 1
+            sn1 = 0
+            sn2 = 1
```
czgdp1807, line 2365:
```suggestion
            sn0, sn1, sn2 = 1, 0, 1
```

czgdp1807: Apple this pattern to other places in the code below.

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2292,156 @@ def ok(x):
...
+                b = sn0 + sn1
+                sn0 = sn1
+                sn1 = sn2
+                sn2 = b
+                i = i + 1
+
+            return padovan
+
+        else:
+            raise ValueError('The provided range is not valid. This condition should satisfy x <= y')
```
czgdp1807, line 2426:
```suggestion
        if not (0 <= x <= y):
            raise ValueError('The provided range is not valid. This condition should satisfy x <= y')

        padovan = list()
        if x <=1 <= y:
            padovan.extend([1,1])
        sn0 = 1
        sn1 = 1
        sn2 = 1
        i = 2
        while sn2 <= y:
            if sn2 >= x:
                padovan.append(sn2)
            b = sn0 + sn1
            sn0 = sn1
            sn1 = sn2
            sn2 = b
            i = i + 1

        return padovan
```

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2292,156 @@ def ok(x):
...
+        if n in (0,1,2):
+            return 1
+        if n == -1:
+            return 0
+        if n == -2:
+            return -1
+        if n>0:
+            return padovan(n - 2) + padovan(n - 3)
+        if n<0:
+            return padovan(n + 3) - padovan(n + 1)
```
czgdp1807, line 2447:
```suggestion
        if n in (0, 1, 2):
            return 1

        if n == -1:
            return 0

        if n == -2:
            return -1

        if n > 0:
            return padovan(n - 2) + padovan(n - 3)

        if n < 0:
            return padovan(n + 3) - padovan(n + 1)
```

sympy/functions/combinatorial/tests/test_comb_numbers.py
```diff
@@ -742,3 +742,12 @@ def test_nD_derangements():
     raises(TypeError, lambda: nD({1: x}))
     raises(TypeError, lambda: nD(m={1: x}))
     raises(TypeError, lambda: nD(m={x: 1}))
+
+
+def test_padovan():
+    assert padovan.is_padovan(3) == True
+    assert padovan.is_padovan(6) == False
+    assert padovan.is_padovan(200) == True
```
czgdp1807, line 750:
```suggestion
    assert padovan.is_padovan(3) is True
    assert padovan.is_padovan(6) is False
    assert padovan.is_padovan(200) is True
```

faze-geek:
Thanks @czgdp1807 for the review .I have made the required changes .

faze-geek:
Looks good from my side .Ping @oscarbenjamin

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2293,152 @@ def ok(x):
...
+                while sn2 > n :
+                    b = sn0 - sn2
+                    sn0 = sn1
+                    sn1 = sn2
+                    sn2 = b
+
+                if Eq(sn2, n, evaluate=True):
+                    return True
+                else:
+                    return False
```
smichr, line 2370:
```suggestion
               return sn2 == n
```

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2293,152 @@ def ok(x):
...
+                while sn2 < n :
+                    b = sn0 + sn1
+                    sn0 = sn1
+                    sn1 = sn2
+                    sn2 = b
+
+                if sn2 == n:
+                    return True
+                else:
+                    return False
```
smichr, line 2354:
```suggestion
                return sn2 == n
```

sympy/functions/combinatorial/numbers.py
```diff
@@ -2292,3 +2293,152 @@ def ok(x):
...
+            b = sn0 + sn1
+            sn0 = sn1
+            sn1 = sn2
+            sn2 = b
+            i = i + 1
+
+        return padovan
+
+    @staticmethod
+    def find_padovan_numbers_in_range(x, y):
```
smichr, line 2401: perhaps following `primerange` example, call it `padovanrange` and use same semantics of not including the end point?

smichr: (At least leave off 'find_', IMO.)

faze-geek: Thanks @smichr , I had actually followed naming schemes from the motzkin number sequence which was the latest number added to combinatorial .But your suggestion makes sense . I have made the required changes !

faze-geek: @oscarbenjamin 
I had made some commits a while ago but tests haven't started running yet, which is strange .Is this some unknown mistake on my part ?

faze-geek:
Ping @oscarbenjamin @smichr @czgdp1807 . Made all the required corrections !

czgdp1807:
@faze-geek Right now you are using a loop of complexity O(n) for finding P(n)? Have you thought of using the matrix `Q` (in https://mathworld.wolfram.com/PadovanSequence.html). Raising a 3 x 3 matrix to the power n should take O(log(n)) steps and asymptotically it should work faster. Can you try comparing the loop and the matrix approach for different values of n (small, medium and large)?

PR #22870, opened 2022-01-17
Series: Fixes series for expressions of type Pow having unspecified exponents  

### LINKED ISSUE #14384: An unspecified power of x is reported to be `O(log(x)**6)`
```
>>> var('x a')
>>> series(x**a, x, 0)
O(log(x)**6)
```
This should probably return the expression `x**a` itself, as `x**a` is not `O(log(x)**6)` when a is negative.

anutosh491:

Fixes #14384 

Changes made are as follows
On **Master**
```
>>> x, a = symbols('x a')
>>> series(x**a, x)
O(log(x)**6)
>>> series(x**(-2*a), x)
O(log(x)**6)
>>> series(exp(a*log(x)), x)
O(log(x)**6)
>>> series(x**I, x)
nan
>>> series(x**(I + 1), x)
O(log(x)**6)
>>> series(exp(I*log(x)), x)
nan
```

On **Committed Branch**
```
>>> x,a = symbols('x a')
>>> series(x**a, x)
x**a
>>> series(x**(-2*a), x)
x**(-2*a)
>>> series(exp(a*log(x)), x)
exp(a*log(x))
>>> series(x**I, x)
x**I
>>> series(x**(I + 1), x)
x**(1 + I)
>>> series(exp(I*log(x)), x)
exp(I*log(x))
```

All these end up in the `_eval_nseries()` method of `exponential.py` where we can check if the series is decisive enough which is not as these depend on sign of `sign(a)` or `sign(I + 1)`. and Hence if it isn't decisive enough we return self rather than `O(log(x)**6`) which was being returned as default for all cases !
#### Release Notes

* series
  * Fixes series for expressions of type Pow having unspecified exponents

### DIFF
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -3032,8 +3032,11 @@ class Expr(Basic, EvalfMixin):
             else:
                 o = Order(x**n, x)
                 s1done = s1.doit()
-                if (s1done + o).removeO() == s1done:
-                    o = S.Zero
+                try:
+                    if (s1done + o).removeO() == s1done:
+                        o = S.Zero
+                except NotImplementedError:
+                    return s1
 
             try:
                 from sympy.simplify.radsimp import collect
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -3345,12 +3345,12 @@ class Expr(Basic, EvalfMixin):
         >>> e.nseries(x, 0, 6, logx=logx)
         sin(logx)
 
-        In the following example, the expansion works but gives only an Order term
+        In the following example, the expansion works but only returns self
         unless the ``logx`` parameter is used:
 
         >>> e = x**y
         >>> e.nseries(x, 0, 2)
-        O(log(x)**2)
+        x**y
         >>> e.nseries(x, 0, 2, logx=logx)
         exp(logx*y)
 
--- a/sympy/core/power.py
+++ b/sympy/core/power.py
@@ -1670,7 +1670,10 @@ class Pow(Expr):
             e = logcombine(e).cancel()
 
         if not (m.is_zero or e.is_number and e.is_real):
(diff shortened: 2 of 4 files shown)


anutosh491:
ping @oscarbenjamin  @smichr for review

anutosh491:
ping @jksuom @0sidharth for review !

anutosh491:
CC: @oscarbenjamin , I have made changes for all group of functions as you had pointed out in the issue reported above . Actually as pointed above , all of them would end up in the same place (not accidently though, it works out mathematically ) where I've added an if-else block to fix these .
Would be glad if you could have a look !

oscarbenjamin:
I don't know the series code so well but hopefully someone else will take a look.

anutosh491:
> I don't know the series code so well but hopefully someone else will take a look.

Cool No worries, I had some idea that you weren't much familiar with the series code (from the series module) though I thought you had some idea regarding the core module .Nonetheless, thanks for your time !

anutosh491:
ping @jksuom @0sidharth  for review ! I am satisfied with the changes here , I would be glad if you'll could have a look ! You could refer to the OP and also a comment on the mentioned issue https://github.com/sympy/sympy/issues/14384#issuecomment-1013659520 which is acts as a guidance for the change .

jksuom:
Looks good. Returning the unevaluated expression instead of a spurious order term is definitely an improvement when a series expansion does not exist. There could be other choices like nan or NotImplementeError but let us be satisfied with this for now. Thanks!

anutosh491:
> Looks good. Returning the unevaluated expression instead of a spurious order term is definitely an improvement when a series expansion does not exist. There could be other choices like nan or NotImplementeError but let us be satisfied with this for now. Thanks! 

@jksuom 
Thanks for the merging ! It's been hard to get reviews for the series/concrete module lately (which are the domains I work on) specially this vacation period of December/January . Glad to here from you today !

PR #22873, opened 2022-01-17
Printing error fixed.

### LINKED ISSUE #22764: Printing modifies internal expression
It looks like printing expressions with `math_ml` (presentational or content) and `latex` unnecessarily modifies the sympy expression. Here's an example with a scientific number notation:

```
def test_printing():
    expr = sympify("3.5*10**(-5)", evaluate=False)
    print("str:", str(expr))
    print("srepr:", srepr(expr))
    print("MATHML:")
    print_mathml(expr)
    print("Presentation MATHML:")
    print_mathml(expr, printer='presentation')
    print("Latex:", latex(expr))
```

This prints:
```
str: 3.5/10**5
srepr: Mul(Float('3.5', precision=53), Pow(Integer(10), Integer(-5)))
MATHML:
<apply>
        <divide/>
        <cn>3.50000000000000</cn>
        <cn>100000</cn>
</apply>

Presentation MATHML:
<mrow>
        <mfrac>
                <mn>3.5</mn>
                <mn>100000</mn>
        </mfrac>
</mrow>

Latex: \frac{3.5}{100000}
```

As we can see:

- `str` and `srepr` nicely reflect the original input, since the `evaluate=False` flag is passed
- `print_mathml` and `latex` strongly modify the input

I think this modification is unnecessary and harmful. Can this be changed? The srepr shows that actually all information would be there to print it correctly?

mohajain:

Fixes #22764

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/printing/mathml.py
+++ b/sympy/printing/mathml.py
@@ -206,7 +206,7 @@ class MathMLContentPrinter(MathMLPrinterBase):
             return x
 
         from sympy.simplify import fraction
-        numer, denom = fraction(expr)
+        numer, denom = fraction(expr, exact=True)
 
         if denom is not S.One:
             x = self.dom.createElement('apply')
@@ -672,7 +672,7 @@ class MathMLPresentationPrinter(MathMLPrinterBase):
 
         def multiply(expr, mrow):
             from sympy.simplify import fraction
-            numer, denom = fraction(expr)
+            numer, denom = fraction(expr, exact=True)
             if denom is not S.One:
                 frac = self.dom.createElement('mfrac')
                 if self._settings["fold_short_frac"] and len(str(expr)) < 7:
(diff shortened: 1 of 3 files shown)


mohajain:
@oscarbenjamin could you please have a look into this?

oscarbenjamin:
The diff I showed in https://github.com/sympy/sympy/issues/22764#issuecomment-1005038031 makes more changes. This PR only seems to make a subset of those changes and I'm not sure why some should be applied and others not.

mohajain:
> The diff I showed in [#22764 (comment)](https://github.com/sympy/sympy/issues/22764#issuecomment-1005038031) makes more changes. This PR only seems to make a subset of those changes and I'm not sure why some should be applied and others not.


The diff you showed made two changes in `mathml.py` and one change in `radsimp.py`. I have made those changes. Is there something I'm missing?

oscarbenjamin:
Yes, you're right. All the changes were made.

However not all the changes have tests added for the things that don't currently work on master.

mohajain:
Should I add the examples mentioned in your comment as tests?

oscarbenjamin:
Yes, we should make sure that there is a test corresponding to each of the lines of code that is changed so that without that change the test would fail.

PR #22877, opened 2022-01-18
Fixing bug in array2matrix( ) involving permutations

### LINKED ISSUE #22814: _array2matrix( ) conversion is wrong
The function `_array2matrix( )` does not return the correct axis permutation for at least one case:

```python
In [1]: from sympy.abc import n

In [3]: from sympy import MatrixSymbol, Identity

In [4]: from sympy.combinatorics import Permutation

In [5]: from sympy.tensor.array.expressions.array_expressions import *

In [7]: from sympy.abc import n

In [8]: x = MatrixSymbol("x", n, 1)

In [9]: I = Identity(n)

In [10]: I1 = Identity(1)

In [11]: expr = PermuteDims(ArrayContraction(ArrayTensorProduct(x, I, I1, x), (0, 3), (1, 7)), Permutation(2, 3))

In [12]: expr.shape
Out[12]: (n, 1, n, 1)
```

So the shape is `(n, 1, n, 1)` for the array expression, but when converting to a matrix expression:

```python
In [13]: from sympy.tensor.array.expressions.conv_array_to_matrix import _array2matrix

In [14]: _array2matrix(expr)
Out[14]: PermuteDims(ArrayTensorProduct(x*x.T, I), (1 2 3))

In [15]: _array2matrix(expr).shape
Out[15]: (n, 1, 1, n)
```

The resulting shape is different. This is a bug, `_array2matrix( )` should return exactly the same shape as its input expression.

Upabjojr:
Fixes https://github.com/sympy/sympy/issues/22814

NO ENTRY

### DIFF
--- a/sympy/tensor/array/expressions/tests/test_convert_array_to_matrix.py
+++ b/sympy/tensor/array/expressions/tests/test_convert_array_to_matrix.py
@@ -668,6 +668,7 @@ def test_convert_array_elementwise_function_to_matrix():
 
 
 def test_array2matrix():
+    # See issue https://github.com/sympy/sympy/pull/22877
     expr = PermuteDims(ArrayContraction(ArrayTensorProduct(x, I, I1, x), (0, 3), (1, 7)), Permutation(2, 3))
     expected = PermuteDims(ArrayTensorProduct(x*x.T, I1), Permutation(3)(1, 2))
     assert _array2matrix(expr) == expected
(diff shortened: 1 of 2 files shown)


sympy/tensor/array/expressions/tests/test_convert_array_to_matrix.py
```diff
@@ -665,3 +665,9 @@ def test_convert_array_elementwise_function_to_matrix():
 
     expr = ArrayElementwiseApplyFunc(Lambda(d, 1 / (2 * sqrt(d))), x)
     assert convert_array_to_matrix(expr) == S.Half * HadamardPower(x, -S.Half)
+
+
+def test_array2matrix():
```
ThePauliPrinciple, line 670: Maybe a comment with a reference to the issue number would be helpful here.

PR #22879, opened 2022-01-19
Get rid of error message in image conversion

oscargus:
See: https://simpleit.rocks/linux/ubuntu/fixing-common-google-chrome-gpu-process-error-message-in-linux/

#### Release Notes

NO ENTRY

### DIFF
--- a/doc/convert-svg-to-pdf.sh
+++ b/doc/convert-svg-to-pdf.sh
@@ -64,4 +64,4 @@ else
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
Looks good

oscargus:
Seems like it was replaced with this instead:
```
[0119/133212.969115:ERROR:gpu_init.cc(457)] Passthrough is not supported, GL is disabled, ANGLE is 
```
Not sure which is worst...

Issue #22878, opened 2022-01-19
RecursionError in trigsimp


proy87:
`trigsimp(8*sqrt(2)*(-exp(2*I*t)*sin(2*t) + (-1 + 2*I)*exp(2*I*t)*cos(2*t) + 1 + I)*exp(-t*(1 + 2*I)), method='fu')`

oscarbenjamin:
More directly the problem is:
```python
In [16]: (-t - 2*I*t + 2951*t*(1 + 2*I)).extract_additively(-t)
---------------------------------------------------------------------------
RecursionError
```

smichr:
MWE: `(-t -t*(1+I)).extract_additively(-t)`

PR #22881, opened 2022-01-19
fixes recursion error during extract_additively

smichr:

non-destructive reconstruction of an expression from which terms have been collected is now implemented by using a helper function to get the coefficient of interest and the terms which did not have the indicated factor. The old method left an unsimplified reconstruction that looked like the original (but was actually zero).

#### Other comments

I'm surprised we don't have a function to do this automatically. `match` is non-collecting:
```python
>>> a,b=symbols('a b ',cls=Wild)
>>> eq=-t-t*I
>>> eq.match(a*(-t)+b)
{b_: -I*t, a_: 1}
```

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -20,6 +20,19 @@ from mpmath.libmp.libintmath import giant_steps
 from collections import defaultdict
 
 
+def _corem(eq, c):  # helper for extract_additively
+    # return co, diff from co*c + diff
+    co = []
+    non = []
+    for i in Add.make_args(eq):
+        ci = i.coeff(c)
+        if not ci:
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
I added fixes #22878 which I presume this is for. It looks good to me.

smichr:
> I added fixes

having it in the commit title is also sufficient. I'll double check by removing the op addition and committing.

oscarbenjamin:
> having it in the commit title is also sufficient.

It might be sufficient to mean that the issue gets closed automatically but I still want to be able to look at the OP and find links to all related issues.

PR #22882, opened 2022-01-19
tensor/indexed: Indexed now sympifies base argument

ThePauliPrinciple:

Helps with https://github.com/sympy/sympy/pull/22445

`Indexed` can currently be used to symbolically indicate elements
of a `MutableArray`. However, when doing so that MutableArray is
stored in `Indexed`'s `args`, which should not be allowed since
a `MutableArray` is not a subclass of `Basic`. `Indexed` now uses
`_sympify` on its `base` argument to convert any `MutableArray`
to an `ImmutableArray`.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/tensor/indexed.py
+++ b/sympy/tensor/indexed.py
@@ -165,6 +165,8 @@ class Indexed(Expr):
             else:
                 return base[args]
 
+        base = _sympify(base)
+
         obj = Expr.__new__(cls, base, *args, **kw_args)
 
         try:


oscarbenjamin:
Is this case not covered by `test_args` somehow?

ThePauliPrinciple:
Doesn't seem like it, see the first two errors in the tests https://github.com/sympy/sympy/runs/4857274723?check_suite_focus=true from https://github.com/sympy/sympy/pull/22445

PR #22887, opened 2022-01-19
physics/quantum: wrapper class for python functions used in OracleGate

ThePauliPrinciple:
Currently python functions are stored directly in `OracleGate`'s
args. This is now updated to use a wrapper `Atom` class.

#### Release Notes

* physics.quantum
  * The wrapper class `OracleGateFunction` is now provided for python functions used in `OracleGate`.

### DIFF
--- a/sympy/physics/quantum/grover.py
+++ b/sympy/physics/quantum/grover.py
@@ -10,6 +10,7 @@ Todo:
 
 from sympy.core.numbers import pi
 from sympy.core.sympify import sympify
+from sympy.core.basic import Atom
 from sympy.functions.elementary.integers import floor
 from sympy.functions.elementary.miscellaneous import sqrt
 from sympy.matrices.dense import eye
@@ -57,6 +58,22 @@ def superposition_basis(nqubits):
 
     amp = 1/sqrt(2**nqubits)
     return sum([amp*IntQubit(n, nqubits=nqubits) for n in range(2**nqubits)])
+
+class OracleGateFunction(Atom):
+    """Wrapper for python functions used in `OracleGate`s"""
+
+    def __new__(cls, function):
+        if not callable(function):
+            raise TypeError("{func} is not a callable.".format(func = function))
+        obj = Atom.__new__(cls)
+        obj.function = function
+        return obj
+
+    def _hashable_content(self):
+        return type(self), self.function
+
+    def __call__(self, *args):
+        return self.function(*args)
 
 
 class OracleGate(Gate):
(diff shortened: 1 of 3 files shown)


sympy/physics/quantum/tests/test_grover.py
```diff
@@ -4,16 +4,16 @@
...
 
 def return_one_on_two(qubits):
     return qubits == IntQubit(2, qubits.nqubits)
-
+return_one_on_two=OracleGateFunction(return_one_on_two)
 
 def return_one_on_one(qubits):
     return qubits == IntQubit(1, nqubits=qubits.nqubits)
-
+return_one_on_one=OracleGateFunction(return_one_on_one)
```
oscarbenjamin, line 16: I guess you could use this as a decorator like:
```
@OracleGateFunction()
def return_one_on_one(qubits):
    ...
```

oscarbenjamin:
Looks good

sympy/physics/quantum/tests/test_grover.py
ThePauliPrinciple: Good idea, I'll add such an example to the docstring too, since it is potentially a lot more convincing for a user to use it in that way.

ThePauliPrinciple:
@asmeurer I approached this from a computer science point of view, so I am not sure if this is also a convenient approach from quantum physics point of view

sympy/physics/quantum/grover.py
```diff
@@ -305,7 +331,8 @@ def apply_grover(oracle, nqubits, iterations=None):
         >>> from sympy.physics.quantum.qapply import qapply
         >>> from sympy.physics.quantum.qubit import IntQubit
         >>> from sympy.physics.quantum.grover import apply_grover
-        >>> f = lambda qubits: qubits == IntQubit(2)
+        >>> from sympy.physics.quantum.grover import OracleGateFunction
+        >>> f = OracleGateFunction(lambda qubits: qubits == IntQubit(2))
         >>> qapply(apply_grover(f, 2))
```
oscarbenjamin, line 336: Maybe `apply_grover` or `OracleGate` could do this conversion automatically.

ThePauliPrinciple: I thought about that, but I couldn't find the right `__new__` function where this would make sense but I am open to suggestions. I also considered creating a generic `PythonFunctionWrapper` and add it to `_sympy_converter`, but felt it was a bit invasive/encouraging to write sloppy code

oscarbenjamin: It's probably not worth overthinking this. If the quantum module becomes maintained then hopefully someone can come up with sensible ideas.

ThePauliPrinciple: I think I've found the right place to do this automatically.

oscarbenjamin:
Looks good to me

PR #22889, opened 2022-01-20
functions : Improved floor and ceiling results 

### LINKED ISSUE #20865: Floor not simplifying/evaluating final answer
```py
In [1]: from sympy import __version__

In [2]: __version__
Out[2]: '1.8.dev'

In [3]: t = sqrt(3) + 1

In [4]: r = (1 / t).simplify()

In [5]: s = t - 1 / r

In [6]: s
Out[6]: 
     1             
- ──────── + 1 + √3
    1   √3         
  - ─ + ──         
    2   2          

In [7]: s.simplify()
Out[7]: 0

In [8]: floor(s)
Out[8]: 
     1             
- ──────── + 1 + √3
    1   √3         
  - ─ + ──         
    2   2          

In [9]: ceiling(s)
Out[9]: 
     1             
- ──────── + 1 + √3
    1   √3         
  - ─ + ──         
    2   2          

In [10]: frac(s)
Out[10]: 0
```
It would be nice to have `floor(s)` and `ceiling(s)` output 0.

faze-geek:

Fixes #20865

Modified :  `sympy/functions/elementary/integers.py` 
Added a condition that simplifies argument when needed and hence improves prior results .
Earlier On Master:
```
>>> from sympy import *
>>> t = sqrt(3) + 1
>>> r = (1 / t).simplify()
>>> s = t - 1 / r
>>> pprint(s)
     1
- ──────── + 1 + √3
    1   √3
  - ─ + ──
    2   2
>>> floor(s)
-1/(-1/2 + sqrt(3)/2) + 1 + sqrt(3)
```

After Commit : 
```
>>> from sympy import *
>>> t,r,s = symbols('t r s')
>>> t = sqrt(3) + 1
>>> r = (1 / t).simplify()
>>> s = t - 1 / r
>>> pprint(s)
     1
- ──────── + 1 + √3
    1   √3
  - ─ + ──
    2   2
>>> floor(s)
0
>>> ceiling(s)
0
```

#### Other comments
Appropriate tests have been added.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/functions/elementary/integers.py
+++ b/sympy/functions/elementary/integers.py
@@ -37,6 +37,12 @@ class RoundFunction(Function):
                 return arg.cancel()
             else :
                 return arg
+        else :
+            if arg.together().expand().cancel().is_integer :
+                for i in arg.args :
+                    if i.is_Symbol is True :
+                        return arg.cancel()
+
         if arg.is_imaginary or (S.ImaginaryUnit*arg).is_real:
             i = im(arg)
             if not i.has(S.ImaginaryUnit):
--- a/sympy/functions/elementary/integers.py
+++ b/sympy/functions/elementary/integers.py
@@ -31,7 +31,12 @@ class RoundFunction(Function):
             return v
 
         if arg.is_integer or arg.is_finite is False:
-            return arg
+            if arg.is_real is False :
+                return arg
+            elif arg.is_number is True :
+                return arg.simplify()
+            else :
+                return arg
         if arg.is_imaginary or (S.ImaginaryUnit*arg).is_real:
             i = im(arg)
             if not i.has(S.ImaginaryUnit):
--- a/sympy/functions/elementary/integers.py
+++ b/sympy/functions/elementary/integers.py
@@ -33,8 +33,8 @@ class RoundFunction(Function):
         if arg.is_integer or arg.is_finite is False:
             if arg.is_real is False :
                 return arg
-            elif arg.is_number is True :
-                return arg.simplify()
+            elif arg.is_Number is True :
+                return arg.cancel()
             else :
                 return arg
         if arg.is_imaginary or (S.ImaginaryUnit*arg).is_real:
--- a/sympy/functions/elementary/tests/test_integers.py
+++ b/sympy/functions/elementary/tests/test_integers.py
@@ -582,3 +582,9 @@ def test_issue_18689():
 def test_issue_18421():
     assert floor(float(0)) is S.Zero
     assert ceiling(float(0)) is S.Zero
+
+
+def test_issue_20865():
+    s = -1/(-1/2 + sqrt(3)/2) + 1 + sqrt(3)
+    assert floor(s) == 0
+    assert ceiling(s) == 0


faze-geek:
Ping @oscargus .

sympy/functions/elementary/integers.py
```diff
@@ -31,7 +31,12 @@ def eval(cls, arg):
             return v
 
         if arg.is_integer or arg.is_finite is False:
-            return arg
+            if arg.is_real is False :
+                return arg
+            elif arg.is_number is True :
+                return arg.simplify()
```
smichr, line 37: `return eq.cancel()` is more targeted

oscargus:
> incorrect earlier.

Strictly not incorrect since that expression is indeed zero.

> If required, I'll add appropriate tests related to this Pull Request later.

Yes, required.

I also think that you should add a case where the `is_real` condition is tested.

sympy/functions/elementary/integers.py
```diff
@@ -31,7 +31,12 @@ def eval(cls, arg):
             return v
 
         if arg.is_integer or arg.is_finite is False:
-            return arg
+            if arg.is_real is False :
```
oscargus, line 34: Can this case actually be trigged?

`(2*I).is_integer` returns  `False`, so what remains are infinite numbers. Probably something like:
```
if arg.is_integer:
    if arg.is_Number:
         return arg
    else:
        return arg.cancel()
if arg.is_finite is False:
   return arg
```
is more efficient. (I think `is_Number` is faster than `is_number`.)

Or do you have any examples which hit all three branches?

oscargus:
Here is an expression that I do not think works with your current code:
```
M = Symbol('M', integer=True)
e = -M/(sqrt(3)/2 - S(1)/2) + M + sqrt(3)*M
floor(e)
```

oscargus:
Regarding performance, using my test case above:
```
%timeit e.cancel()
2.98 ms ± 70.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%timeit e.simplify()
11.5 ms ± 527 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

oscargus:
Here is a bit of more information that can be useful to determine when and when not you should "simplify" (using the original expression s):

```
s.is_integer
Out[31]: True

s.is_Number
Out[32]: False

s.is_number
Out[33]: True
```

faze-geek:
> Strictly not incorrect since that expression is indeed zero.

I  have verified that statement in my description.
> Yes, required.
> I also think that you should add a case where the `is_real` condition is tested.

Yeah, will do that once the code snippet is certain !

sympy/functions/elementary/integers.py
faze-geek: Yeah makes sense ,thanks @oscargus for showing how it improves the code performance time-wise.

sympy/functions/elementary/integers.py
faze-geek: I had involved three conditions since there were a couple of fails related to imaginary numbers and one simplify fail in 'test_diophantine.py` which returned the simplified answer which was unexpected .
```
assert parametrize_ternary_quadratic(eq) == \
        (8*p**2 - 3*p*q, -8*p*q + 8*q**2, 5*p*q)
```
I have understood the changes you suggest and also ran tests .I will look into the suggested expression which fails according to you currently and then make a commit along with tests .

oscargus: If you get a *better* results after a change, feel free to update the test (it will be reviewed as well).

oscargus:
> Yeah, will do that once the code snippet is certain !

The thing is that to some extent it also depends on the tests as those also gives information about what actually happens. So always a good thing to add those (here it is sort of obvious that you will simply use the original issue code, so not very much work, although there may of course be more required as one goes along, like the one I proposed).

sympy/functions/elementary/integers.py
oscargus: I thought a bit more about it and I assume that the reason for returning arg here is that we knows that we have the final result, so an early exit.

I would change the original condition to 
```
if arg.is_Atom and arg.is_integer or arg.is_finite is False:
```

(which should be the same as `if arg.is_Atom and arg.is_integer or arg.is_infinite:`)

So we exit here it the value is infinite or if it is a single number/symbol and integer.
This may to some extent solve the problem (as we do not exit because it `is_integer`, but not a single value).

Here one can consider performing a cancel and then do another similar test, in case something evaluates to the something where we can do an early exit.

faze-geek: Yeah I will look into this .Thanks for the suggestions @oscargus 
I will make the changes soon !

sympy/functions/elementary/integers.py
oscarbenjamin: Neither `cancel` nor `simplify` should be used here. Both are arbitrarily expensive functions even if `cancel` is better than `simplify`.

What is the purpose of this call?

oscargus: "Simplify" expressions that do not cancel otherwise, see the examples discussed.

oscarbenjamin: Why does this check `is_number`? The purpose of `is_number` is to check whether `.evalf()` works...

The OP example is handled by something like `s.together().expand()` but I'm not sure even that should be done. We can also use `s.is_zero` but for this case that ends up calling `minpoly` which can be expensive (the call to `minpoly` should be removed).

oscargus: I do not think it should call `is_number`. It is probably better to call `is_integer and is_Atom or is_infinite` (or maybe `(is_integer or is_infinite) and is_Atom`) instead of `is_integer or not is_finite==False` in the top call (to get an early exit if nothing more can be done) and then try to "simplify" if there is something left, try `is_integer or is_finite` again (without `is_Atom`) for possible new early exit and then continue.

And, yes, there is a trade-off when things should be "simplified". Typically it is not a problem calling simplify in ones code at appropriate places, but then one would at the same time avoid expression blowup inside the library code (which is not the case here, but can be useful to avoid it elsewhere). For the problem in #20865 I do not really see the problem calling `simplify` on the obtained expression (as that will give the correct result), although if part of a larger problem it can be convenient to not have to.

oscarbenjamin: Exactly: the user can call simplify if they want. We shouldn't do it in library code because then it happens for all inputs and inevitably some things will become extremely slow for no good reason.

oscargus: With the trade-off that some things become extremely slow if you do not simplify in the library code. :-)

oscarbenjamin: There are some specific places where more specific functions (not `simplify`) can be used to simplify certain kinds of expressions. There needs to be some context for knowing when that situation occurs though. If random expression types can call simplify whenever they want then it's impossible to make anything involving those types that doesn't risk extreme slowness in some situations.

smichr:
I think the rabbit trail was started by `is_integer` being too smart. The evaluation routine must be simplifying the expression `1 + sqrt(2) - 1/(-1 + sqrt(2))` to realize that it is an integer so then it seems like the same thing should be done here in the `floor/ceiling` routine. I would rather see the reason for this resolving that expression as integer (0) being fixed rather than to see the floor/ceiling routine changed.

oscarbenjamin:
> I would rather see the reason for this resolving that expression as integer (0) being fixed rather than to see the floor/ceiling routine changed.

It happens because of this:
```python
In [1]: e = 1 + sqrt(2) - 1 / (-1 + sqrt(2))

In [2]: e._eval_is_real()
Out[2]: True

In [3]: e._eval_is_positive()
Out[3]: False

In [4]: e._eval_is_negative()
Out[4]: False
```
The resolver in the core assumptions is smart enough to figure out that `e` is therefore zero.

Now the question is how do `_eval_is_positive/negative` figure this out? The answer is here:
https://github.com/sympy/sympy/blob/a2342a1f56ae8aa67c21636d50e9cb8dba203a9c/sympy/core/expr.py#L888-L894
That's guarded by `if self.is_number` and comes after ordinary numerical evaluation seems to return zero.

There are a few things wrong with this:

1. Clearly the check should be in `_eval_is_zero`.
2. It is also possible to conclude in the opposite direction: if the minimal polynomial is not just a symbol then the number is definitely not zero but that isn't being used.
3. The `minimal_polynomial` function can be extremely expensive so it should not be used as part of the assumptions system like this (although it is probably fine to use it in `.equals()` which already calls things like `simplify` and can be expected to be slow.

I would remove the call to `minpoly` from the assumptions system. From memory there was an obscure test that depended on this somewhere but not one that couldn't just be changed.

sympy/functions/elementary/integers.py
faze-geek: From comments on #20865
> Is this causing a problem inside of SymPy (this is the reason that some other expression blows up)? If not, as you showed, calling `simplify` on the result will give you 0.
>There is indeed a trade-off, so this issue may be closed if its deemed that this utility isn't worth the slowdown.

@oscargus @0sidharth 
Why are we looking deeply into the `simplify` tradeoff when we have established that there other methods like `arg.cancel()` which surely will make the code much more efficient time wise.

Using a check like `is_zero()`(after removing the call to `minpoly`) is what I think we should go for .

faze-geek: Also according to @smichr 
>I think the rabbit trail was started by is_integer being too smart. The evaluation routine must be simplifying the expression 1 + sqrt(2) - 1/(-1 + sqrt(2)) to realize that it is an integer so then it seems like the same thing should be done here in the floor/ceiling routine. I would rather see the reason for this resolving that expression as integer (0) being fixed rather than to see the floor/ceiling routine changed.

I feel according to what @smichr and @oscarbenjamin stated I maybe could raise a stand alone issue where this can be dealt with by SymPy contributors rather than going through the comments to find the crux of the matter.

oscargus: > Using a check like is_zero()

Why? Then this will come back for another expression which doesn't evaluate to zero. Add 1 to `s` and that will not resolve.

What I cannot really understand is that you have gotten quite a bit of suggestions which you seemingly ignore. It may not be the way to ping in people (at all) and then just ignore the response.

faze-geek: My apologies , I didn't expect so many suggestions and got a bit confused regarding the best approach here .

faze-geek:
> The thing is that to some extent it also depends on the tests as those also gives information about what actually happens. So always a good thing to add those (here it is sort of obvious that you will simply use the original issue code..)

@oscargus Yeah added tests and corrected obvious performance errors in latest commit (to be worked further on)

faze-geek:
> Here is an expression that I do not think works with your current code:
> ```
> M = Symbol('M', integer=True)
> e = -M/(sqrt(3)/2 - S(1)/2) + M + sqrt(3)*M
> floor(e)
> ```

@oscargus The latest commit addresses this . I have added a condition to the evaluation routine which first simplifies the equation .The exact change that can resolve this expression to integer(0) is unknown so an additional check here could make sense .

Corrections by this commit -
```
>>> M = Symbol('M', integer=True)
>>> e = -M/(sqrt(3)/2 - S(1)/2) + M + sqrt(3)*M
>>> k = Symbol('k', integer=True)
>>> floor(e)
0
>>> t = 1
>>> floor(e+t)
1
>>> t = k + e
>>> floor(t)
k
```

faze-geek:
When this PR was opened there were multiple reviews (some similar some contradicting each other ) and I even commented that I was overwhelmed with the different approaches suggested .I would like if someone could react to [above comment](https://github.com/sympy/sympy/pull/22889#issuecomment-1020116377) showing the current improvements by this PR so that I can start working on it again .Thanks !

PR #22890, opened 2022-01-20
codegen: IndexedAssignment for better support of Indexed printing

ThePauliPrinciple:
Currently the support for printing `Assignments` with `Indexed` is
limited. This PR adds additional convenient options.
`IndexedAssignments` are always normalized to a form with explicit
`Sum`s to avoid limiting the user to write specific expressions.
Common implicit sum uses are provided through bool flags. Any
indices of `Indexed` objects that are not `Idx` are ignored for the
purposes of implicit sums.

#### Other comments
This is an early WIP, but I hope to get feedback early in the process since features along these lines have been requested several times and it should be flexible, convenient and intuitive to use.

This [colab](https://colab.research.google.com/drive/1moawTKjzM5HifVfbzzrUWBVnQ1J6Jt9f?usp=sharing) document provides a workspace to test the current version of `IndexedAssignment`
#### Release Notes

NO ENTRY


ThePauliPrinciple:
I'm especially looking for suggestions for supported input/output, both for what is currently already working (for writing eventual tests) and what still has to be implemented (for providing more flexibility)

sympy/codegen/ast.py
```diff
@@ -496,6 +498,134 @@ class Assignment(AssignmentBase):
...
+    ========
+
+    XXX: To do
+    """
+    def __new__(cls, lhs, rhs, tensor_contraction = False, explicit = False):
+        from sympy.tenser.indexed import IndexedBase
+        lhs = _sympify(lhs)
+        rhs = _sympify(rhs)
+        if tensor_contraction:
+            rhs = _tensor_contraction(rhs)
```
asmeurer, line 591: This seems like it would be better as a separate simplification function.

asmeurer:
What's the point of this vs. an Assignment with an Indexed?

ThePauliPrinciple:
There are two points

- Codewriting for assignments is a mess, there are several special cases which need to be handled (e.g. piecewise, Indexed, matrices/arrays). These special cases are not handled in the same way in every programming language, so if the code for all of that is in the same method in the printers, the entire method needs to be replaced, so splitting it up into different nodes to indicate different behaviour makes it easier to separate that.
- We want to store the dummy variables in the codegen node for the case that we do not only want to print the code, but also print variable declarations (e.g. for c and fortran), so if we would rely on the printer to do that every time, we would potentially get a different dummy everytime.

That said, I'm not set on any kind of interface, as long as it becomes possible to write more complicated expressions.

Note that the case below is not yet supported correctly (it's not initializing correctly and sometimes has an incorrect looping structure), but this is the kind of expression I want to be able to print:

```python
>>> A = IndexedBase('A')
>>> B = IndexedBase('B')
>>> C = IndexedBase('C')
>>> D = IndexedBase('D')
>>> i,j,k,l = Idx('i',m), Idx('j',m), Idx('k',m), Idx('l',m)
>>> expr = IndexedAssignment(D[i], A[i,k]*(Sum(B[l,k]-C[j],l)), explicit = True)
>>> print(fcode(expr))
      do i = 1, m
         do l = 1, m
            _Dummy_72(j, k) = B(l, k) - C(j) + _Dummy_72(j, k)
         end do
         do k = 1, m
            do j = 1, m
               _Dummy_71(i) = A(i, k)*_Dummy_72(j, k) + _Dummy_71(i)
            end do
         end do
         D(i) = _Dummy_71(i)
      end do
>>> print(ccode(expr))
for (int i=0; i<m; i++){
   for (int l=0; l<m; l++){
      _Dummy_72[m*j + k] = B[m*l + k] - C[j] + _Dummy_72[m*j + k]
   }
   for (int k=0; k<m; k++){
      for (int j=0; j<m; j++){
         _Dummy_71[i] = A[m*i + k]*_Dummy_72[m*j + k] + _Dummy_71[i]
      }
   }
   D[i] = _Dummy_71[i]
}
```

The behaviour for e.g. the `NumPyPrinter` would then be to simply convert it to an array expression and print that.

ThePauliPrinciple:
I was under the impression that a draft would supress checks, but it seems like it isn't. Is there any way to do so?

oscarbenjamin:
> Is there any way to do so?

I don't think so.

PR #22891, opened 2022-01-20
refactoring code with chained comparison which is more pythonic, concise, readable and efficient

idiomaticrefactoring:
refactoring code with chained comparison which is more pythonic, concise, readable and efficient; how do think this change which has practical value？

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/parsing/latex/_parse_latex_antlr.py
+++ b/sympy/parsing/latex/_parse_latex_antlr.py
@@ -189,7 +189,7 @@ def convert_postfix_list(arr, i=0):
                         sympy.Symbol)
                     # if the left and right sides contain no variables and the
                     # symbol in between is 'x', treat as multiplication.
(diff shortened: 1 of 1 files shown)


sympy/parsing/latex/_parse_latex_antlr.py
```diff
@@ -189,8 +189,7 @@ def convert_postfix_list(arr, i=0):
                         sympy.Symbol)
                     # if the left and right sides contain no variables and the
                     # symbol in between is 'x', treat as multiplication.
-                    if len(left_syms) == 0 and len(right_syms) == 0 and str(
-                            res) == "x":
+                    if len(left_syms) == 0 == len(right_syms) and str(res) == 'x':
```
smichr, line 192:
```suggestion
                    if not (left_syms or right_syms) and str(res) == 'x':
```

PR #22894, opened 2022-01-20
core/numbers: __eq__ no longer uses _sympify

ThePauliPrinciple:

Helps with https://github.com/sympy/sympy/pull/22607

Currently `__eq__` in `numbers` always tries to convert `other` to
a SymPy type. `__eq__` now uses direct comparisons, which should be
more efficient.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -1382,9 +1382,10 @@ class Float(Number):
     def __eq__(self, other):
         from sympy.logic.boolalg import Boolean
         if isinstance(other, float):
+            #converting here to catch inf float cases
             return self == Float(other)
         if isinstance(other, int):
-            return self == Integer(other)
+            return self.num == other
         if not isinstance(other, Basic):
             return Basic.__eq__(self, other)
         if isinstance(other, Boolean):
--- a/sympy/core/numbers.py
+++ b/sympy/core/numbers.py
@@ -1387,7 +1387,7 @@ class Float(Number):
         if isinstance(other, int):
             return self.num == other
         if not isinstance(other, Basic):
-            return Basic.__eq__(self, other)
+            return self._do_eq_sympify(other)
         if isinstance(other, Boolean):
             return False
         if other.is_NumberSymbol:
@@ -1888,7 +1888,7 @@ class Rational(Number):
         if isinstance(other, float):
             return self == Float(other)
         if not isinstance(other, Basic):
-            return Basic.__eq__(self, other)
+            return self._do_eq_sympify(other)
         if not isinstance(other, Number):
             # S(0) == S.false is False
             # S(0) == False is True
@@ -2258,6 +2258,8 @@ class Integer(Rational):
     def __eq__(self, other):
         if isinstance(other, (int, float)):
             return (self.p == other)
+        if not isinstance(self, Basic):
+            return self._do_eq_sympify(other)
         elif isinstance(other, Integer):
             return (self.p == other.p)
         return Rational.__eq__(self, other)
@@ -3464,6 +3466,8 @@ class Infinity(Number, metaclass=Singleton):
         return super().__hash__()
 
     def __eq__(self, other):
+        if not isinstance(self, Basic):
+            return self._do_eq_sympify(other)
         return other is S.Infinity or other == float('inf')
 
     def __ne__(self, other):
@@ -3630,6 +3634,8 @@ class NegativeInfinity(Number, metaclass=Singleton):
         return super().__hash__()
 
     def __eq__(self, other):
+        if not isinstance(self, Basic):
+            return self._do_eq_sympify(other)
         return other is S.NegativeInfinity or other == float('-inf')
 
     def __ne__(self, other):
@@ -3761,6 +3767,8 @@ class NaN(Number, metaclass=Singleton):
         return super().__hash__()
 
     def __eq__(self, other):
+        if not isinstance(self, Basic):
+             return self._do_eq_sympify(other)
         # NaN is structurally equal to another NaN
         return other is S.NaN
 
@@ -3886,7 +3894,7 @@ class NumberSymbol(AtomicExpr):
 
     def __eq__(self, other):
         if not isinstance(other, Basic):
-            return Basic.__eq__(self, other)
+            return self._do_eq_sympify(other)
         if self is other:
             return True
         if other.is_Number and self.is_irrational:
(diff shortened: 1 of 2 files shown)


sympy/core/numbers.py
```diff
@@ -1879,9 +1880,11 @@ def __ceil__(self):
         return self.ceiling()
 
     def __eq__(self, other):
-        try:
-            other = _sympify(other)
-        except SympifyError:
+        if isinstance(other, MPQ):
+            return self.p == other.numerator and self.q == other.denominator
+        if isinstance(other, (int, float)):
```
oscargus, line 1885: Won't this lead to that?

Rational(10000000000000001, 10000000000000000) == 1.0

ThePauliPrinciple: Yes, I believe it would, which would be incorrect. I'll also add a test for this, so it won't be possible to accidentally introduce this behaviour in the future.

oscargus: Probably better to compare `self.p == other*self.q`. For ints that should be clearly better, but for floats it may provide some other weird consequences.

sympy/core/numbers.py
```diff
@@ -3879,9 +3882,7 @@ def _eval_evalf(self, prec):
         return Float._new(self._as_mpf_val(prec), prec)
 
     def __eq__(self, other):
-        try:
-            other = _sympify(other)
-        except SympifyError:
+        if not isinstance(other, Basic):
             return NotImplemented
```
oscargus, line 3886: Won't this mean that one gets a NotImplemented doing e.g `zoo == 1`? Or am I missing something?

ThePauliPrinciple: if a.__eq__(b) returns `NotImplemented`, then it will check b.__eq__(a), if both return `NotImplemented` it will fall back to identity comparison (I believe this is `a is b`, which will probably be `False`). Basically it allows an arbitrary other class to define a sensible `a == b` and `b == a` without having to modify both classes.

sympy/core/numbers.py
ThePauliPrinciple: I tried that first (since it should be more efficient), but it caused issues with `== float('inf')`

sympy/core/numbers.py
ThePauliPrinciple: I now changed it to use `Basic.__eq__` because someone might have user defined conversions (e.g. in `converter` or because `other` has `_sympy_` defined. Although that method also returns `NotImplemented` when it doesn't know the type of `other`.

sympy/core/numbers.py
oscargus: OK! But for int it should be better?

sympy/core/numbers.py
oscargus: Ahh, all right! It is not `NotImplemtedError.

ThePauliPrinciple:
I can't seem to reproduce the optional-dependencies error locally

sympy/core/numbers.py
```diff
@@ -38,6 +38,10 @@
 from sympy.external.pythonmpq import PythonMPQ as MPQ
```
oscargus, line 38: I _think_ that if you import `MPQ` from gmpy `instead` it will basically do the lines added in this commit.

https://github.com/sympy/sympy/blob/dcdf59bbc6b13ddbc329431adf72fcee294b6389/sympy/external/gmpy.py#L83-L104

oscargus:
I'm wondering if one should use `SYMPY_INT` instead of `int` for type checking? I haven't really looked into the `GMPY`-stuff, no not sure what will happen, but that seems to also cover the "GMPY-integers", although I have no idea where they may turn up nor how it will work in the code...

(There are probably persons that have better ideas about this than me guessing though...)

ThePauliPrinciple:
I don't think there are any converters defined for "GMPY-integers", so it shouldn't currently work with SymPy either, unless they are a subclass of `int` already (in which case the current instance check should work too). 

But I too am not entirely confident on this.

ThePauliPrinciple:
I am now only converting `float`s to `Float`s, mostly because `inf` float values aren't a `Float` in SymPy but their own type, which `Float.__new__` handles.

oscargus:
Here is a snippet: 
https://github.com/sympy/sympy/blob/07022af0da7a829ae40e019fcf2a7030344d45da/sympy/core/numbers.py#L602-L603
so it seems to be possible to convert to `Integer` and afaik `int` in Python is arbitrary precision. That may not mean that it is possible to convert and/or compare directly though.

ThePauliPrinciple:
It might be possible to convert to `Integer`, but unless it is in `_sympy_converter` or `_external_converter` it won't when you are comparing with it directly. So if we would be checking for anything other than `int`, it would be an extension of the current functionality.

What types are you finding in `SYMPY_INTS`?

ThePauliPrinciple:
Commit 4587971 is to ensure that
```python
>>> from sympy import S
>>> from sympy.core.sympify import _sympify
>>> class MyNaN:
...    def _sympy_(self):
...        return S.NaN
>>> S.NaN == MyNaN()
False
>>> S.NaN == _sympify(MyNaN())
True
```
changes to
```python
>>> S.NaN == MyNaN()
True
>>> S.NaN == _sympify(MyNaN())
True
```

ThePauliPrinciple:
Actually, I don't want to commit to this, I'm basically relying on the tests and not really understanding what is happening. If someone else wants to pick this up, feel free to do so.

oscargus:
> What types are you finding in `SYMPY_INTS`?

`SYMPY_INTS` is either just `int` (if gmpy is not installed) or `int` and `mpz` (if gmpy is installed). So I think it often makes sense to use that.

`_sympify` seems quite overused indeed, so being able to skip it, if so only moving some parts out, possibly relying on `_sympify` as a final step, can indeed be worthwhile. Although it also becomes a guessing game which order they should be in. (Which it already is, not just as clearly as the code is "somewhere else".)

ThePauliPrinciple:
I think a change like this would be an improvement, but I am not very knowledgable about how the sympy number types store their information and when non sympy types should equal them.

The general idea would be that SymPy classes explicitly define comparisons with classes that SymPy claims to be aware of and conversion only happens for user-defined competability (e.g `converter[my_int] = my_int_converter` or using the `_sympy_` interface).
Those conversions can now be performed more efficiently with `self._do_eq_sympify(other)`, which is essentially the part of `_sympify` which only checks for externally defined conversions.

oscargus:
I also think that there is still on ongoing discussion if `Integer(1) == Float(1.0)` or not. (Probably because there are good arguments for both alternatives...).

To me, it seems to be somewhat easier with integers and rational numbers since those are treated exact in Python. But floats are indeed challenging. For the gmpy-types I am not really sure how they work.

I'll try to keep this in mind and extract the parts that I am pretty certain about and create a PR.

Issue #16504, opened 2019-03-30
Allow matrix with shape (1, 1) to multiply any matrix shape?


Upabjojr:
Matrices with shape `(1, 1)` may be seen as scalars. They might be allowed to multiply any other matrix without raising `ShapeError`.

Similarly matrices of shape `(k, 1)` may also be allowed to multiply any other matrix?

supreet11agrawal:
I think that this can be implemented. Numpy also supports broadcasting.
For reference: https://docs.scipy.org/doc/numpy/user/basics.broadcasting.html

cbm755:
#9190 is related.

oscarbenjamin:
This kind of special case can often lead to awkwardness down the line. Why is it advantageous? The user can just use a scalar instead of a 1x1 matrix.

This is what happens in Matlab and it makes sense there. But that's because in Malab there is no way to distinguish between a scalar and a 1x1 matrix. Why would it be useful in SymPy?

supreet11agrawal:
In making machine learning and deep learning models, broadcasting is especially beneficial. The cases where it is used mostly are somewhat like `w*x + b` where `w` is `nxn` weight matrix and `a` and `b` are `1*n` vectors.

Upabjojr:
> This kind of special case can often lead to awkwardness down the line. Why is it advantageous? The user can just use a scalar instead of a 1x1 matrix.
> 
> This is what happens in Matlab and it makes sense there. But that's because in Malab there is no way to distinguish between a scalar and a 1x1 matrix. Why would it be useful in SymPy?

Well, actually not. The 1x1 matrix may be an expression which cannot be simplified leading to a 1x1 shape.

I was actually coming along this problem while pondering about the support for matrix derivatives.

Upabjojr:
> In making machine learning and deep learning models, broadcasting is especially beneficial. The cases where it is used mostly are somewhat like `w*x + b` where `w` is `nxn` weight matrix and `a` and `b` are `1*n` vectors.

We are **never** going to support broadcasting. The (1, 1) matrix shape is a special case as it is equivalent to a scalar.

Upabjojr:
`KroneckerProduct` may be used to achieve this behaviour with (1, 1)-shaped matrices.

oscarbenjamin:
> Well, actually not. The 1x1 matrix may be an expression which cannot be simplified leading to a 1x1 shape.

Can you give an example?

asmeurer:
[...]

PR #22897, opened 2022-01-21
Conversion of array to matrix: use KroneckerProduct as a trick to broadcast matrix products

Upabjojr:
Conversion of array to matrix: use KroneckerProduct as a trick to broadcast matrix products.

NO ENTRY


Upabjojr:
Cfr https://github.com/sympy/sympy/issues/16504#issuecomment-1018339957

Issue #22892, opened 2022-01-20
SymPy Deprecation backlog


ThePauliPrinciple:
There is a bunch of code that has been deprecated for a long time. Maybe it is a good idea to set a "hard cut-off" for deprecated code e.g. "everything deprecated in version 1.5 or earlier gets removed"?

Or is it prefered to keep it in the codebase until it is actively interfering with development?

Or is the reason it is not removed because it is difficult to actually remove these pieces of code?

```grep
git grep "deprecated_since_version="
sympy/assumptions/ask.py:        deprecated_since_version="1.8"
sympy/assumptions/ask.py:        deprecated_since_version="1.8"
sympy/assumptions/assume.py:            deprecated_since_version="1.8"
sympy/assumptions/assume.py:            deprecated_since_version="1.8"
sympy/assumptions/assume.py:            deprecated_since_version="1.8"
sympy/assumptions/handlers/common.py:            deprecated_since_version="1.8"
sympy/calculus/finite_diff.py:    deprecated_since_version="1.1", issue=11410)(_as_finite_diff)
sympy/calculus/finite_diff.py:                                deprecated_since_version="1.5").warn()
sympy/codegen/array_utils.py:    deprecated_since_version="1.8"
sympy/concrete/expr_with_limits.py:                deprecated_since_version=1.6,
sympy/core/basic.py:                                deprecated_since_version="1.9").warn()
sympy/core/basic.py:    deprecated_since_version="1.10", issue=22288)(_preorder_traversal)
sympy/core/compatibility.py:    deprecated_since_version="1.10", issue=22352)(_default_sort_key)
sympy/core/compatibility.py:    deprecated_since_version="1.10", issue=22352)(_ordered)
sympy/core/compatibility.py:    deprecated_since_version="1.10", issue=22352)(__nodes)
sympy/core/compatibility.py:    deprecated_since_version="1.10", issue=22352)(_as_int)
sympy/core/compatibility.py:    deprecated_since_version="1.10", issue=22352)(_is_sequence)
sympy/core/compatibility.py:    deprecated_since_version="1.10", issue=22352)(_iterable)
sympy/core/compatibility.py:    deprecated_since_version="1.10", issue=22352)(_notiterable)
sympy/core/expr.py:                                deprecated_since_version="1.9").warn()
sympy/core/expr.py:                                deprecated_since_version="1.9").warn()
sympy/core/function.py:                deprecated_since_version="1.5").warn()
[...]

Issue #21329, opened 2021-04-15
Deprecation - Line2DBaseSeries.get_segments


Davide-sd:
The `get_segments` method implemented in the `Line2DBaseSeries` is used to convert two list of coordinates, `x` and `y`, into a list of segments used by Matplotlib's `LineCollection` to plot a line. Looking at the entire code base, `get_segments` is only called by `MatplotlibBackend` class. 

Since the list of segments is only required by Matplotlib (for example, Bokeh, Plotly, Mayavi, K3D only require lists of coordinates), it make sense to move that function inside `MatplotlibBackend` class. This will be achieved with PR #21256 .

mgeier:
I would like to add a few notes because it took me quite a while to understand how to replace my previous usage of `get_segments()`.

Previously, the method `get_points()` always returned uniformly sampled points, which meant that some functions were not plotted correctly when using `get_points()` to plot with Matplotlib.

To avoid this problem, the method `get_segments()` could be used, which used adaptive sampling and which could be used with Matplotlib's `LineCollection`.

A few months ago, this has been changed in 29a97f34602b976b6f1955bdaa6fbfe152182968, which is somewhat hidden in #21256. This change hasn't yet been released.

Since that change, `get_points()` can also use adaptive sampling (but it cannot return multiple discontinuous pieces, as `get_segments()` could previously IIRC).

Note that there are currently a few open issues with plotting discontinuous functions, see #17738, #18455, #21538.

oscarbenjamin:
I don't understand why this is deprecated. Is it not possible to have the `get_segments` function continue to work after the other changes?

Also the release note does not mention anything being deprecated (and is not very informative).

@Davide-sd is it possible to restore `get_segments` before the next release (which is very soon)?

Davide-sd:
`Line2DBaseSeries.get_segment` has been deprecated. That function has been moved into "MatplotlibBackend.get_segments`, as written in the [deprecation warning message](https://github.com/sympy/sympy/blob/master/sympy/plotting/plot.py#L609).

Up to Sympy 1.8, `Line2DBaseSeries.get_segments` **always** returns data formatted in such a way to be used by Matplotlib's `LineCollection`, irrespective of the adaptive keyword argument:

```
[[array(x0, y0), array(x1, y1)],
 [array(x1, y1), array(x2, y2)],
 [array(x2, y2), array(x3, y3)],
 ...
]
```

The idea behind #21256 is well documented. In summary:
[...]

PR #22900, opened 2022-01-22
Move deprecation policy to the documentation, and update it

### LINKED ISSUE #20015: Better way of handling deprecation removal issues
I've noticed that virtually all SymPy developers misunderstand the purpose of the "deprecation-removal" tag on GitHub. 

The idea, as outlined at https://github.com/sympy/sympy/wiki/Deprecating-policy, is twofold. One, whenever there is a deprecation, there should be an issue to track the eventual removal of it (hence the name "deprecation **removal**"). And two, the deprecation warnings are supposed to point people to that issue. So the issue text should contain information relevant for end-users on why the feature is deprecated and how to fix their code. 

That was the idea behind this, but I've not seen anyone (other than myself) do it that way. So I'm wondering if we should just change it to something that doesn't confuse the developers so much. 

Instead, I see people using the label for issues relating to deprecating a feature, rather than *removing* an existing deprecation. And people never properly fill out useful information in the issue, despite the fact that the warnings link to it. I've also seen people link to the PR that removed the warning, rather than an issue. Of people won't create new issues for removing the deprecation but just reuse the one about making the deprecation.

Should we try to track the removal of deprecations? Maybe it isn't that important, since people seem to remove deprecations eventually anyway. I've also noticed the removal issues can prompt people looking for easy to fix issues to try to remove deprecations prematurely. As an aside, we don't currently have a policy on how long deprecations last, but maybe we should adopt one to remove the ambiguity here. 

For explaining the deprecation to users, maybe instead of doing it in issues, we should do it in documentation instead. I think the original motivation for issues was so that it could have a discussion forum for each deprecation in case people have questions. But I'm not sure if this has really been necessary. This also ties in nicely with the idea of documenting what things are deprecated https://github.com/sympy/sympy/issues/19806.

asmeurer:

Fixes https://github.com/sympy/sympy/issues/20015
Discussed at https://github.com/sympy/sympy/issues/22892

I have moved the deprecation policy from [the wiki](https://github.com/sympy/sympy/wiki/Deprecating-policy) into the documentation. I have also made several updates to it. See `deprecations.md`.

- I added some more text going over what should and should not be deprecated. Please read through the document and let me know if you disagree with anything. It's also somewhat lengthy now, so if you have any suggestions on something that should be trimmed or how it should be better organized, let me know. 
  
  I did want to include some examples of what does and doesn't constitute public API. I know we don't have a policy on this yet, so I've hopefully made it clear that this is not strictly codified (but I don't want that to prevent us from saying anything about it). 

  I opened https://github.com/sympy/sympy/issues/23037 to discuss this.

- I have made the deprecation period officially **at least 1 year**. Hopefully I have made it clear that this is just a minimum period. Deprecations can last longer than this. The deprecation policy is time-based and not version-based. I have added some justifications for this. Please let me know if you agree with this policy.

- I have completely nuked the "deprecation removal issue" stuff. See https://github.com/sympy/sympy/issues/22892. Basically no one really understood this, and it was a huge mess.

  Instead, we are going to put all information about deprecations in the documentation itself. This will be done in 3 places: 

  1. the warning itself, which should contain more informative and longer messages than they currently do.
  2. a `.. deprecated::` directive in each docstring.
  3. a new "active deprecations" page in the documentation, which will list all active deprecations, and provide a longer description of each, including context for why the deprecation was made and links to relevant discussions on GitHub.

  See the document for a description of what should go in each. 

#### Other comments

Additionally, I have done the following.

- I've updated the constructor for creating deprecation warnings to be much simpler. It used to be 

  ```
  SymPyDeprecationWarning(value=None, feature=None, last_supported_version=None,
                          useinstead=None, issue=None, deprecated_since_version=None)
  ```

  now it is

  ```py
  sympy_deprecation_warning(message, *, active_deprecations_target, deprecated_since_version)
  ```

  The `message` should contain a description of the deprecated functionality and what it should be replaced with. Multiline strings with code blocks are be encouraged. The `active_deprecations_target` should be a cross-reference target in the "active deprecations" page which will be used to automatically generate a link to the document at the end of the warning. `deprecated_since_version` is the same as before. It should be the version of the next release (the version in release.py minus `.dev`). Library code should use the `sympy_deprecation_warning` function to issue a warning rather than the `SymPyDeprecationWarning` constructor directly (previously you had to call `SymPyDeprecationWarning(...).warn()`, but it was easy to forget the `warn()` part).

  (yes this is itself a backwards incompatible change. But `SymPyDeprecationWarning` should not be used by non-library code).

  You can also use `@deprecated` but it now has the exact same signature as `sympy_deprecation_warning`, and there is no effective difference between using it and just calling `sympy_deprecation_warning()` at the top of the function.

  See the diff for tons of examples of this. 

- Update all existing deprecations to use the above format. That means replacing the existing "deprecation removal" issues with text in the active deprecations document, rewriting all the warning text to be more descriptive and to use the new `SymPyDeprecationWarning` signature, and adding `.. deprecation:: <version>` directives to all relevant docstrings.

- I have not removed any currently active deprecations (except for one minor thing that was missed in https://github.com/sympy/sympy/pull/22952). Some deprecations here are quite old, but also a lot of them were documented extremely poorly, to the point that I don't know how anyone could ever tell what they were actually supposed to do to fix their code. So I would suggest keeping the existing deprecations for now, especially if we can get this into the next release.

- I have updated the `warns` and `warns_deprecated_sympy` test context managers:
  - Both now check that the `stacklevel` argument is set correctly. `stacklevel` should be set so that the lien of code for a warning is the user line of code. The default value for `sympy_deprecation_warning` is set so that you usually don't have to worry about this, but you will need to increase it if you issue the warning from some helper function. If there's no way do this consistently, you can disable it with `warns(test_stacklevel=False)`.
  - `warns()` no longer ignores all other warnings. This was hiding a use of deprecated code in the tests (plot_implicit() relies on the string fallback in `sympify()`. I have left this as XFAIL because I don't know how to fix it. It seems to depend heavily on the poor design of `exerimental_lambdify` and the interval math).

- Once this is merged, I will delete the old policy from the wiki and update any references to it, and close all "deprecation removal" issues, replacing them with links to the new document.

#### Release Notes

- other
  - Update the deprecation policy, and include it in the documentation. SymPy's official deprecation policy is now that all deprecations will last at least 1 year.
  - All active deprecations are now included in a new "active deprecations" page in the documentation.
  - All deprecation warning messages have been improved to better indicate what is deprecated and how to update code that uses it.


doc/src/guides/contributing/deprecations.md
```diff
@@ -0,0 +1,388 @@
+(deprecation-policy)=
+# Deprecation Policy
+
+This page outlines SymPy's police on doing deprecations, and describes the
+steps developers should take to properly deprecate code.
```
oscarbenjamin, line 5: policy

doc/src/guides/contributing/deprecations.md
```diff
@@ -0,0 +1,388 @@
...
+
+## When to deprecate and when not to
+
+Sometimes the API of SymPy must change in an incompatible way. We try to avoid
+this, because changing API breaks people's code, but it is often deemed worthy
+to do so. Some reasons APIs are changed can include:
+
+- The existing API is confusing.
+- There is unnecessary redundancy in the API.
+- The existing API limits what is possible.
```
oscarbenjamin, line 28: Looking through the list of currently open issues tagged with deprecation removal it seems that the most common reason is just because a change is being made to something that seems private but it isn't clear whether anyone downstream might be using. The problem is that the distinction between public and private is not clearly delineated anywhere so we end up deprecating things that probably no one cares about.

oscarbenjamin: Another reason would be because something is fundamentally broken and cannot be made to work sensibly without incompatible changes.

doc/src/guides/contributing/deprecations.md
```diff
@@ -0,0 +1,388 @@
...
+- Keyword argument default values.
+- The mathematical conventions used to define a function.
+- Submodule names for **public** functions that are not included in the
+  top-level `sympy/__init__.py`, since users must use the fully qualified
+  module name to import them (note: many such functions are not included
+  because they aren't actually public API at all. For these, no deprecations
+  are required to make breaking changes).
+
+And broadly, a non-exhaustive list of some things that don't consistent public
+API, and therefore don't require deprecations to change, include
```
oscarbenjamin, line 67: constitute

oscarbenjamin:
> * Update the `SymPyDeprecationWarning` constructor to be much simpler. Right now it is
>   ```
>   SymPyDeprecationWarning(value=None, feature=None, last_supported_version=None,
>                           useinstead=None, issue=None, deprecated_since_version=None)
>   ```
>     
>   and it automatically constructs a message from those. Instead, I plan to use
>   ```python
>   SymPyDeprecationWarning(msg, *, active_deprecations_target, deprecated_since_version)
>   ```

It would be better if this was just a function like:
```
warn_deprecated(...)
```
I think that a constructor that emits a warning is confusing.

oscarbenjamin:
I think the general principles here are all good.

oscarbenjamin:
> * @oscarbenjamin do we have an issue regarding the better codification of public vs. private APIs? I thought I remembered there being one, but I couldn't find it.

I don't think there is one. You might be thinking of #22105

doc/src/guides/contributing/deprecations.md
```diff
@@ -0,0 +1,388 @@
...
+
+- The precise form of an expression. In general, functions may be changed to
+  return a different but mathematically equivalent form of the same
+  expression. This includes a function returning a value which it was not able
+  to compute previously.
+- The name of a submodule for functions is included in the top-level
+  `sympy/__init__.py`. Such functions should be imported directly, like `from
+  sympy import <name>`.
```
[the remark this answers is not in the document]
asmeurer: I'll make it clearer. It means something like the name of the submodule `sympy.functions.elementary.exponential` isn't public API if we wanted to move the files around, because all the public names in that should be imported from the top-level.

asmeurer:
> I think that a constructor that emits a warning is confusing.

That's actually a typo on my part. The constructor doesn't emit a warning. There should be a `warn()` call at the end. But a helper function sounds good too, to avoid the same mistake.

doc/src/guides/contributing/deprecations.md
```diff
@@ -0,0 +1,388 @@
...
+indicating what the user should be using instead.
+
+This allows users a chance to update their code without it completely
+breaking. It also gives SymPy an opportunity to give users an informative
+message on how to update their code, rather than making their code simply
+error or start giving wrong answers.
+
+## When to deprecate and when not to
+
+Sometimes the API of SymPy must change in an incompatible way. We try to avoid
```
moorepants, line 22: Suggested change:

"We should, in general, avoid this,"

Below it is emphasized in multiple places why changing API is not good, so I think opening with a bit stronger statement is best.

doc/src/guides/contributing/deprecations.md
```diff
@@ -0,0 +1,388 @@
...
+this, because changing API breaks people's code, but it is often deemed worthy
```
moorepants, line 23: but it can be deemed worthy to do so in some circumstances.

moorepants: I think our default position should be to discourage API changes unless the benefit largely outweighs the cost. Of course, whether something largely outweighs the cost is subjective because we aren't measuring any information to make it objective. But I think the language in this document should give the overall impression that you should make a very strong effort in avoiding API changes at all and make a very strong effort to design APIs that have less likelihood of needing to be changed. Opening the document with language that sets a strong tone early is good. If our language is weak at the opening about this, then the actual actions of developers will be even weaker (maybe too much API changes tolerated).

doc/src/guides/contributing/deprecations.md
```diff
@@ -0,0 +1,388 @@
...
+
+- Function names.
+- Keyword argument names.
+- Keyword argument default values.
+- The mathematical conventions used to define a function.
+- Submodule names for **public** functions that are not included in the
+  top-level `sympy/__init__.py`, since users must use the fully qualified
+  module name to import them (note: many such functions are not included
+  because they aren't actually public API at all. For these, no deprecations
+  are required to make breaking changes).
```
moorepants, line 64: I don't really agree with this one. `sympy.physics.vector.frame.ReferenceFrame` isn't importable like `from sympy import ReferenceFrame`, but it is certainly public API. Does this bullet say that only names imported by `from sympy import *` are public?

doc/src/guides/contributing/deprecations.md
```diff
@@ -0,0 +1,388 @@
...
+API, and therefore don't require deprecations to change, include
+
+- The precise form of an expression. In general, functions may be changed to
+  return a different but mathematically equivalent form of the same
+  expression. This includes a function returning a value which it was not able
+  to compute previously.
+- The name of a submodule for functions is included in the top-level
+  `sympy/__init__.py`. Such functions should be imported directly, like `from
+  sympy import <name>`.
+- Positional argument names.
```
moorepants, line 76: Technically Python permits:

```python
def f(a, b, c=None):
    return a, b, c

r = f(a=12, b=13, c=14)
```

So users can rely on positional names. I don't think it is widely used, but code would break if you changed `a` and `b` to something else in the function definition. It isn't too hard to deprecate this either.

doc/src/guides/contributing/deprecations.md
oscarbenjamin: There needs to be a more nuanced definition here. For most of the codebase it is true that public API is exposed at top-level like `from sympy import X` but there are major exceptions like physics, stats, combinatorics...

doc/src/guides/contributing/deprecations.md
asmeurer: We could try to be more strict about this by using positional-only arguments and keyword-only arguments everywhere. I personally like the idea of using more keyword-only arguments, which helps avoid backwards compatibility breaks (https://github.com/sympy/sympy/pull/21803). I'm less a fan of forcing positional-only, but technically speaking, it is a way to enforce that the argument name is *not* part of the API (it also requires Python 3.9 to do syntactically, but it won't be too long before we drop <=3.8).

But even so, if you take something like `solve`, which calls its first argument `f`, I don't think users should be relying on that argument being called `f`. We should be able to rename it to something more readable like `expr` or `eq`. There's a bunch of other places where the argument is just called `x` or `expr` or something like that. I don't think users should really be using keyword arguments when they call functions. 

It's also my experience that the reverse is more common anyway. Have you seen real examples of users using keyword arguments where they aren't required? Personally I've seen users call things positionally that probably should be keyword-only, but not the other way around.

doc/src/guides/contributing/deprecations.md
asmeurer: I think you aren't understanding it. What I am trying to write here agrees with you. 

The basic point is this: if users *have* to write the module name to access a function (because it can't be imported from the top-level `from sympy import X`), then the module name itself is part of the public API. This is because if we renamed the module, it would break user code. 

OTOH if a function is in the top-level, then users can write `from sympy import X` instead of `from sympy.submodule.submodule import X`, so if we rename `submodule` it won't break user code. 

Or at least that's the ideal situation. It doesn't consider users who may be importing top-level names directly from submodules anyway. I've actually seen quite a bit of user code that does things like `from sympy.functions import exp` instead of `from sympy import exp`. The fact that we intentionally do this in library code doesn't help here. So in practice, if we ever did want to rename modules or move functions around, we might want to do a deprecation anyway. Maintaining backwards compatibility here is quite trivial (see for instance the deprecation for `sympy.printing.ccode`, which was done for technical reasons). 

I think a more codified public/private API policy should probably include things like underscores for names that aren't supposed to be public. So e.g., `sympy.functions` would become `sympy._functions` and all user code should import names from there from the top-level. But this requires a *lot* of work to actually do, so it's quite out of scope for this PR. For now, I just want to give some general guidelines. 

I'll reword these two bullet points to make them clearer.

asmeurer: I guess I'll just remove the non-public half of this. The problem is that sometimes names are in a file that aren't included in `__init__.py` but are still public API. The most common example of this is base classes, which we want people to be able to subclass. To really distinguish this requires a lot more work. 

The cleanest thing I can say for now is that submodule renames should be deprecated (which as I noted, is not difficult). Even if a function is just moved around, it's very easy to keep an import of it at the top of the file so that direct imports from the old module still work.

asmeurer: See what you think of it now.

Also it really needs to be made clear here that these lists here are just guidelines. We don't have a real public/private API policy yet. There are always exceptions to the rules.

asmeurer:
@oscarbenjamin @moorepants (and others) It's clear you have the most comments about the "what should and shouldn't be deprecated" parts of this. My intention here is just to have some guidelines, so that we can some documentation on it. I know we can't really truly say what is and isn't public API without a lot more work.

I'm more curious if you have any concerns about the other two parts, namely, 

1. the "at least 1 year" part, which would be a new official policy, 
2. the parts about how deprecations should be documented and the new proposed signature for `SymPyDeprecationWarning`. 

Especially the second one. If no one has any issues with it, I'll go ahead and start updating the existing deprecations. One possible concern is that it means writing a lot more corresponding documentation whenever a deprecation is done than we currently do. If it helps, I can change one or two of the existing deprecations to the proposed style to better show what it will look like.

doc/src/guides/contributing/deprecations.md
asmeurer: I did have something written about this but it was lower down. I've moved it to the top of the file and given it its own heading.

asmeurer:
An aside, I noticed some more flaws in our current Sphinx theme when writing this:

- Bold text is almost indistinguishable from normal text (consider how the "at least 1 year" looks in the rendered document).
- Admonitions are not rendered as boxes, only notes and warnings (I've put the list of public and non-public guidelines in `admonition` directives). 

CC @bertiewooster

doc/src/explanation/active-deprecations.md
```diff
@@ -0,0 +1,32 @@
+(active-deprecations)=
+# List of active deprecations
+
+This pages lists all active deprecations in the SymPy codebase. See the
+{ref}`deprecation-policy` page for a description of SymPy's deprecation
+policy, as well as instructions on how for contributors to deprecate things.
+
+Of particular, the deprecation policy for SymPy is for deprecations to last at
```
oscarbenjamin, line 8: In particular

doc/src/guides/contributing/deprecations.md
```diff
@@ -0,0 +1,387 @@
...
+deprecated, as there will be no way for users to fix their code to continue to
+work.
+
+Deprecations should always
+
+1. Allow users to continue to use the existing APIs unchanged during the
+   deprecation period (with a warning).
+2. Allow users to always fix their code so that it stops giving the warning.
+3. After users fix their code, it should continue to work after the deprecated
+   code is removed.
```
oscarbenjamin, line 147: Another point to consider here is how a library that attempts to support multiple versions of sympy should straddle the versions without seeing any warning. If there is a way to do it like:
```python
try:
    from sympy.x import y
except ImportError:
    from sympy.z import y
```
then that should be documented.

oscarbenjamin:
I think it would be good to have something softer than the current approach to deprecation warnings. In particular there are two softer approaches that some other libraries use:

Document something as deprecated without emitting any warnings. This means that we agree that something should be deprecated and change all usage within the sympy codebase. We also make sure that the docs do not suggest doing something deprecated and perhaps add a note somewhere that it was previously suggested to do x but that is now discouraged and users should do y instead. We can also add a release note for this.

Another approach is to emit a PendingDeprecationWarning. This means that the warning is only seen by test suites so is visible to developers but not users. Under pytest and other test libraries a PendingDeprecationWarning will be visible. This gives time for libraries to release an update before users see the warning. The warning could also be made to error in the SymPy test suite so that the deprecated approach is not used within the codebase.

doc/src/guides/contributing/deprecations.md
asmeurer: That's a good point. I think generally users should do version checks. After all, each deprecation lists the first version it is deprecated in. Otherwise they have to do some sort of warning filter, or trying the new way first and falling back to the old way if it doesn't work, which may not be simple in general.

asmeurer:
The problem is that most people aren't going to read the documentation or release notes. If someone has existing code that uses something that gets deprecated, they aren't going to revisit the docstring for that function. In fact the whole point of SymPyDeprecationWarning is that it does get omitted by default, since the built-in DeprecationWarning does not.

I've seen libraries use softer warnings and PendingDeprecationWarning type things like you describe in instances where the deprecation lasts for a very long time. In some cases, the functionality is "deprecated" in the sense that there is a much better alternative, but there are no plans at all to actually remove it, as doing so would break too much user code. I think in SymPy that would only ever happen if we were to replace some key core functionality with something much better, say, replacing old assumptions syntax entirely with new assumptions syntax (although I don't actually think we should do that, but just as an example).

asmeurer:
Also we've had the relatively loud SymPyDeprecationWarning (i.e., with filterwarnings('always')) for over 10 years (ee839cc2fc0601e4d2164731830af24a371d08eb), and to my memory, no one has ever complained about it. 

Even so, I think we should ask for feedback on this on the mailing list.

doc/src/guides/contributing/deprecations.md
oscarbenjamin: The point is that whatever downstream libraries need to do should be clearly documented. Filtering warnings doesn't seem acceptable to me but providing a clear sample of code that can be copied into somewhere does.

oscarbenjamin:
> The problem is that most people aren't going to read the documentation or release notes.

That's true but it doesn't mean that we shouldn't even try to make it possible for an author of a downstream library who wants to be diligent to do so. Also I think that usually it's good to let something bed in and make sure that it is used and is usable before insisting that users switch over to it.

You mentioned the new assumptions which is a good example. Although the old assumptions never started giving deprecation warnings the announcement that the new assumptions should be used was extremely premature. There should have been a period of time to establish that the new assumptions were usable before suggesting that users should definitely use them. Likewise the docs page that says that you should not use solve because we now have solveset was extremely premature when it so obviously doesn't work as a drop-in replacement (and never would!).

For a significant change there should be a period of time in which a new way of doing things is suggested or encouraged as an option while the old way continues to work without any warnings. The warning should only appear when the new way of doing things has been around long enough that the new way clearly works and library authors don't need to worry about continuing to use the old way to support old versions of sympy any more.

The current approach is basically that in sympy version 1.X you have to use `old_way` and then in version 1.X+1 you have to use `new_way`. That doesn't leave any room to upgrade downstream code while straddling multiple sympy versions.

asmeurer:
The goal here is that the warnings should now always be more explicit on what you should replace things with. In some instances the warning message can even be dynamic based on the user input so that it shows exactly what to replace it with. I have a simple example of this here, but maybe I should replace it with something bigger just to emphasize that the warning messages can be larger. I also plan to show by example when I update the existing messages. 

The idea is to have everything documented in multiple places, the warning message, the `deprecated` note in the docstring, the "all activate deprecations" document, and the release notes. This leads to a lot of duplication, but I think it's worth it because people may look in each of the different places to get information about deprecations. It's also fine to duplicate content between the different places. It also allows writing longer form text, for instance, in the "active deprecations" document, which would be too verbose for the docstring or error message. 

Generally, though, the idea is that we should document the old way and what to replace it with. That's separate from the actual ways a user can could support different SymPy versions. Most users won't need to do this, but some libraries do. There are different methods for doing this, I don't think it really makes a huge difference which one they choose. All of these seem fine to me.

```py
try:
    <new_way>
except <relevant exception>:
    <old way>
```

This way is the simplest, but only in situations where `<new way>` raises a reliable exception in older SymPy versions (like an `ImportError`).

```py
import sympy
if int(sympy.__version__.split('.')[1]) >= 10:
    <new way>
else:
    <old way>
```

or

```py
from sympy.utilities.exceptions import SymPyDeprecationWarning
import warnings
warnings.filterwarnings("error" category=SymPyDeprecationWarning)
try:
    <old way>
except SymPyDeprecationWarning:
    <new way>
```

This one is usually less preferable, because it will stop working when `<old way>` is finally removed. But there might be a situation where it is easier to use. Certainly silencing the warning instead of fixing code as a quick fix is a legitimate thing that libraries will want to do in some cases.

I'm not sure if we really need to document any of this though. None of it is specific to SymPy. The same applies for any feature in any library or even the standard library that is different across versions. It also applies in non-deprecation situations where there is a new feature added that you'd like to use, but can only use if the version of SymPy you are using is new enough. Maybe if a specific deprecated feature has something about it that makes this harder than usual we can add something, but otherwise, this all seems like general stuff that any seasoned Python developer will already have picked up. Resources like https://stackoverflow.com/questions/1532488/supporting-multiple-python-versions-in-your-code or https://stackoverflow.com/questions/57634170/supporting-multiple-versions-of-a-python-dependency seem to cover the basics.

> Also I think that usually it's good to let something bed in and make sure that it is used and is usable before insisting that users switch over to it.

> For a significant change there should be a period of time in which a new way of doing things is suggested or encouraged as an option while the old way continues to work without any warnings. The warning should only appear when the new way of doing things has been around long enough that the new way clearly works and library authors don't need to worry about continuing to use the old way to support old versions of sympy any more.

That's a good point. I'll add something about this. 

It does somewhat contradict what I wrote [here](https://github.com/sympy/sympy/pull/22900/files#diff-ee2d6877b6ca2d058d5a08fa49d2440f18d330b8c386dd13e6bf49607e7262b7R199-R201), however. In general, people may want to deprecate as early as possible to accelerate the removal of the old code. In some instances this is warranted because the mere existence of the old code prevents something else from being possible. In others, it's just people being a little too anxious.

asmeurer:
> It does somewhat contradict what I wrote here, however. In general, people may want to deprecate as early as possible to accelerate the removal of the old code. In some instances this is warranted because the mere existence of the old code prevents something else from being possible. In others, it's just people being a little too anxious.


Thinking about it, I'm not really sure I agree that we should always release a version with the new functionality before deprecating the old functionality. If we know that we are planning to remove the old functionality, then deprecating it sooner is better, as that gives users more time to upgrade.

This doesn't mean of course, that old features always have to be deprecated and removed at all. They quite often can stay intact alongside the newer functionality. But when they are, I don't see a benefit to waiting.

oscarbenjamin:
> Thinking about it, I'm not really sure I agree that we should always release a version with the new functionality before deprecating the old functionality. If we know that we are planning to remove the old functionality, then deprecating it sooner is better, as that gives users more time to upgrade.

Note that I said "for a *significant* change". There needs to be a more nuanced terminology here. There is a difference between deprecating something by saying "don't use that" in some place that developers might see compared to forcing warnings that are visible to users. None of us wants users to see warnings for things that should just get fixed by developers.

Our policy should aim to make it possible that everything just gets fixed without users knowing about it. If we immediately issue warnings with no prior chance for downstream change then that goal is impossible.

asmeurer:
This is a good example of what I was talking about with the top-level names thing https://github.com/sympy/sympy/issues/22288. The functions `preorder_traversal`, `postorder_traversal`, `bottom_up`, `walk`, and `use` were all moved from various places around the codebase into a new file in sympy/core/traversal.py. However, all these names except for `walk` were already available in the top-level `sympy/__init__.py` (`walk` is still not there). So is it really necessary to keep and deprecate `sympy.utilities.postorder_traversal` and so on? Any code that uses `from sympy import postorder_travsersal` will not break. 

Note that this particular deprecation was made since the last release, so there is still time to reconsider it.

Certainly, if we keep the deprecation, I think we should change the warning to recommend `sympy.postorder_traversal` instead of `sympy.core.traversal.postorder_traversal`. Importing names from specific submodules when they are in the top-level is the whole reason this sort of deprecation had to happen in the first place (that is assuming it is agreed to keep this deprecation at all, which I personally think we shouldn't). Considering not just "`sympy.postorder_traversal`" but also "`sympy.utilities.iterables.postorder_traversal`" to be "public API" means that simple file reorganizations like this require deprecation cycles. 

CC @oscargus

oscarbenjamin:
> So is it really necessary to keep and deprecate `sympy.utilities.postorder_traversal` and so on?

This is really a good example of deprecating things that should probably have been considered private in the first place. Some of those functions were clearly publicly importable but anyone who imported them from the submodule was breaking the API contract. Of course you're right that the new advice should suggest the proper public way to import them.

oscargus:
One thing we should also consider is the how we should write the documentation examples. Maybe it is clear in the instructions, but from my experience sometimes functions/classes that are available on the top level are imported with full (or partial) path. I made some improvements, but I doubt it is fully done. To me that is a step to be able to avoid deprecation when moving things around, not just that they can be imported at the top level, but also that the code examples indicate it.

It should also be clear from "somewhere" that one should import at the highest level possible.

Also, we should probably be more strict when introducing helper functions, considering if they are private or not (I believe this is pretty author dependent, but probably a reason why it is sometimes hard to restructure the code).

I've played around with matplotlib recently and they are currently making a large restructuring, making things private and deprecating public access. They have the policy of two major releases, so it takes some time, but over time it will for sure be much easier to change things without breaking other things.

We should probably also encourage keyword only arguments for new things.

asmeurer:
It might be possible to build a linter that disallows overly specific imports in doctests. 

Sphinx cross references are also annoying here as you often have to use the fully qualified name to remove ambiguity. We do document the `~.` trick in the docs style guide, but I think it's still not well known.

oscargus:
Yes, a linter checking that would be nice.

Slightly related: should we, in general, try to make more functions private? For example, in https://github.com/sympy/sympy/blob/master/sympy/utilities/source.py there are two "public" functions that are only used once in SymPy. They will probably be required for a longer time, but it also means that we cannot easily modify the behavior of them or move them.

Here one should probably deprecated them first, so it is a long term plan, but is the opinion that we should go in that direction?

oscarbenjamin:
> Slightly related: should we, in general, try to make more functions private?

Yes, absolutely.

asmeurer:
> Slightly related: should we, in general, try to make more functions private?

Yes. Anything that isn't private can't be changed without breaking things. The more internal details that are public, the harder it is to improve the implementations of various things. Right now we don't have a clear policy on what is public and what isn't, so the best way to do this is to either prefix function names with an underscore, or otherwise state in the documentation somewhere that something is private. 

Ideally, we would have the reverse situation, where everything is assumed to be private unless we explicitly mark it as public. But without that explicit messaging, we can't really expect users to know that they shouldn't use internal functions.

asmeurer:
@oscarbenjamin are you planning to make a 1.10 release soon? If so, it would be good to get this finished before then if possible.

oscarbenjamin:
> it would be good to get this finished before then if possible.

How long will that take? This release is already delayed as far as I'm concerned.

The longer it gets delayed the more people will suggest that things should get included in it when potentially it is just better to make the release now and include those other things in the next release.

asmeurer:
If the release is imminent then this doesn't need to block it. It would be nice to get some of the nicer messages out for the new deprecations, but it's not critical. I was more checking for myself to see what the plan was. I do hope get this completed this week either way.

asmeurer:
I just pushed a change that makes it so that the `stacklevel` parameter is tested by default in `warns` and `warns_deprecated_sympy`. This should be set so that the line of code shown by the warning is the user code, if possible. In a lot of cases, this is ambiguous, including most existing UserWarning warnings in SymPy, because there might be different call stacks leading to the function that issues the warning. But for deprecation warnings it should usually be unambiguous.

asmeurer:
Here's an example of what the new deprecation warnings will look like 

```py
>>> import sympy.printing.ccode
<stdin>:1: SymPyDeprecationWarning:

The sympy.printing.ccode submodule is deprecated. It has been renamed to
sympy.printing.c.

See https://docs.sympy.org/latest/explanation/active-deprecations.html#deprecated-printing-code-submodules
for details.

This has been deprecated since SymPy version 1.7. It
will be removed in a future version of SymPy.

>>>
```

This is generated from this code https://github.com/sympy/sympy/pull/22900/files#diff-77d7928eb9d1df7157f7a9d4b844360fd0fd36991b2eb3152876e451e56b124dR11-R18:

```py
sympy_deprecation_warning(
    """
    The sympy.printing.ccode submodule is deprecated. It has been renamed to
    sympy.printing.c.
    """,
    deprecated_since_version="1.7",
    active_deprecations_target="deprecated-printing-code-submodules",
)
```

The URL won't work until this is merged and released, but it will eventually point to [this](https://github.com/sympy/sympy/pull/22900/files#diff-0e1766c0a3437bc6388516043af5b7d09b8598602b16d9dbd5c467612ac0c6b6R130-R142).

Any suggestions on how to improve the automatic formatting of the message are welcome.

asmeurer:
@czgdp1807 @0sidharth can you review [`6ec6cae`](https://github.com/sympy/sympy/pull/22900/commits/6ec6caeeab133f28b81f035a1931daf7fe579b07)? I have tried my best to understand the change to StochasticProcess.distribution from https://github.com/sympy/sympy/issues/20078 and its motivation, but I'm not super familiar with sympy.stats, so please check that what I wrote covers the situation accurately.

czgdp1807:
Looks good to me as of now. @0sidharth Thoughts?

asmeurer:
Something I just realized here. If we make dynamic error messages for warnings, they will show each time the message changes. For example:

```py
>>> from sympy.stats import *
>>> X = Die('X', 6)
>>> sample(X, numsamples=3)
<stdin>:1: SymPyDeprecationWarning:

The numsamples parameter to sympy.stats.sample() is deprecated.
Either use a list comprehension, like

[sample(...) for i in range(3)]

or add a dimension to size, like

sample(..., size=(3,))

See https://docs.sympy.org/latest/explanation/active-deprecations.html#deprecated-sympy-stats-numsamples
for details.

This has been deprecated since SymPy version 1.9. It
will be removed in a future version of SymPy.

[1, 2, 3]
>>> sample(X, numsamples=4)
<stdin>:1: SymPyDeprecationWarning:

The numsamples parameter to sympy.stats.sample() is deprecated.
Either use a list comprehension, like

[sample(...) for i in range(4)]

or add a dimension to size, like

sample(..., size=(4,))

See https://docs.sympy.org/latest/explanation/active-deprecations.html#deprecated-sympy-stats-numsamples
for details.

This has been deprecated since SymPy version 1.9. It
will be removed in a future version of SymPy.

[3, 4, 5, 5]
>>> sample(X, numsamples=4)
[3, 2, 2, 4]
```

This is because the `warnings` module by default treats two warnings as the same iff they have the same message. 

This behavior is hard-coded into the warnings module and would be quite difficult to workaround. Is this something that we should consider "worth it" in order to give more helpful error messages?

(aside: I've realized like at least a dozen different limitations in the warnings module while working on this. I'm almost tempted to create a "betterwarnings" package that fixes all these things)

0sidharth:
The `sympy.stats` deprecation message looks good to me too.

moorepants:
This is maybe relevant. Just noticed SciPy 1.8 added underscores to all private API that "were private but happened to miss underscores in
their names". There is a larger section in the release notes about this.

https://github.com/scipy/scipy/releases/tag/v1.8.0

I'd like to see deprecation warnings if we do something like this, so that users and downstream libs get warned if these names change.

asmeurer:
@moorepants the question of how we distinguish what is public and private is a separate question. I opened https://github.com/sympy/sympy/issues/23037 to discuss it. I expect it will not be an easy problem to figure out.

asmeurer:
@Davide-sd @mgeier I've tried my best to summarize the deprecation of `get_segments()` (https://github.com/sympy/sympy/issues/21329) in [`b89518f` (#22900)](https://github.com/sympy/sympy/pull/22900/commits/b89518fd8dd815cb3ef9588beee55b583650220f). Please take a look and make sure everything I wrote is correct and I didn't miss anything. 

I'm a little confused what the difference between `get_points()` and get_data()` is. From what I could tell, they both return the same thing, which can be used to recreate what `get_segments()` used to return.  The issue mentions using `get_points()` instead but the PR https://github.com/sympy/sympy/pull/21256/files replaced all the uses of `get_segments()` in the tests with `get_data()`.

asmeurer:
@Upabjojr @schymans you're going to have to review my update of the deprecations in sympy.physics.units [`905eb42` (#22900)](https://github.com/sympy/sympy/pull/22900/commits/905eb426131ca9542a6b258462d9ae984e5b2563). I've done my best to understand the deprecations and document them, but the documentation in the issues and existing warnings was pretty sparse and confusing, so it's possible I misinterpreted something and wrote something that is wrong.

asmeurer:
@Upabjojr @drybalka please review the various tensor deprecations [`cba899d` (#22900)](https://github.com/sympy/sympy/pull/22900/commits/cba899d4137b0b65f6850120ee42cd4fcd4f9dbf). These were by far the most complicated deprecations I had to update here, so I'm not 100% sure I did everything correctly in the code, and I'm also not sure I documented everything correctly and completely. If you see anything wrong or that could be added to the docs that are added in that commit, please let me know.

doc/src/explanation/active-deprecations.md
```diff
@@ -0,0 +1,1231 @@
...
+```
+
+The remaining were only intended for internal SymPy use and should not be used
+by user code.
+
+Additionally, these two functions, `ordered` `default_sort_key`, used to also
+be in `sympy.utilities.iterables` but have been moved from there as well.
+
+
+## Version 1.9
```
asmeurer, line 140: Question for @oscarbenjamin before I do it. Are you opposed to adding the date for each release here, like

```
## Version 1.9 (2021-09-30)
```

This would be useful considering we are going to use a date-based policy, but it would also be one more thing to do for the release.

Upabjojr:
All deprecations in `tensor.tensor` and `physics.units` can be removed in my opinion.

asmeurer:
@Upabjojr I'm not removing deprecations in this PR. Considering how poorly the deprecations in tensor in particular were documented prior to this PR, I think it's worth doing at least one release with this better documentation. If you could review said documentation to make sure it is accurate, that would be really helpful.

doc/src/explanation/active-deprecations.md
```diff
@@ -0,0 +1,1231 @@
...
+
+(deprecated-expr-free-symbols)=
+### `expr_free_symbols`
+
+The `expr_free_symbols` attribute of various SymPy objects is deprecated.
+
+`expr_free_symbols` was meant to represent indexed objects such as
+`MatrixElement` and {class}`~.Indexed` as free symbols. This was
+intended to make derivatives of free symbols work. However, this now workw
+without making use of the method:
```
oscarbenjamin, line 162: workw -> works

doc/src/explanation/active-deprecations.md
```diff
@@ -0,0 +1,1231 @@
...
+it potentially impacts on some downstream use cases that might be similar to
+the uses of Matrix with non-Expr elements that were in the SymPy codebase. A
+potential replacement for code that used Matrix with non-Expr elements is
+DomainMatrix if the elements are something like domain elements and a domain
+object can be provided for them. Alternatively if the goal is just printing
+support then perhaps TableForm can be used.
+
+It isn't clear what to advise as a replacement here without knowing more about
+the usecase. Feel free to comment below indicating what you want to do with
+non-Expr Matrix elements and we can discuss how to achieve that.
```
oscarbenjamin: What does "feel free to comment below" mean if this is now in the docs rather than a GH issue?

doc/src/explanation/active-deprecations.md
```diff
@@ -0,0 +1,1231 @@
...
+
+  ```py
+  CoordSystem(name, patch, ['x', 'y']) # DEPRECATED
+  ```
+
+  use
+
+  ```py
+  CoordSystem(name, patch, symbols('x y', real=True)
+  ```
```
oscarbenjamin, line 571: Needs `)`

doc/src/explanation/active-deprecations.md
```diff
@@ -0,0 +1,1231 @@
...
+### Creating an indefinite `Integral` with an `Eq` argument
+
+Passing an [`Eq()`](sympy.core.relational.Equality) object to
+{func}`~.integrate` is deprecated in the case where the integral is
+indefinite. This is because if $a = b$, $\int a\,dx = \int b\,dx$ is not true
+in general, due to the arbitrary constants (which `integrate` does not
+include).
+
+If you want to make an equality of indefinite integrals, use `Eq(integrate(a, x),
+integrate(b, x))` explicitly.
```
oscarbenjamin: Maybe also worth mentioning that if you have an equality already then you can do
```
Eq(integrate(eq.lhs, x), integrate(eq.rhs, x))
```

sympy/testing/pytest.py
```diff
@@ -28,7 +31,6 @@
 
 if USE_PYTEST:
     raises = pytest.raises
-    warns = pytest.warns
```
oscarbenjamin, line 31: Why is this changed?

doc/src/guides/contributing/deprecations.md
```diff
@@ -0,0 +1,549 @@
...
+All deprecation warnings should be something that users can remove by updating
+their code. Deprecation warnings that fire unconditionally, even when using
+the "correct" newer APIs, should be avoided.
+
+This also means all deprecated code must have a completely functioning
+replacement. If there is no way for users to update their code, then this
+means API in question is not yet read to be deprecated. The deprecation
+warning should inform users of a way to change their code so that it works in
+the same version of SymPy, as well as all future versions (see
+[below](deprecation-documentation)).
```
oscarbenjamin, line 147: Really it should be possible for users to change their code so that it works in both past and future versions of sympy. In some cases it is possible because the "correct" way of using a particular SymPy feature has not changed and only an "incorrect" way is being deprecated (example: importing something from a submodule rather than using `from sympy import X`). In some cases it isn't possible because the "correct" way doesn't exist in previous sympy versions. Either way there should be at least an attempt to explain how cross-version compatibility can be achieved for any downstream library authors.

doc/src/guides/contributing/deprecations.md
```diff
@@ -0,0 +1,549 @@
...
+  ```
+
+  with
+
+  ```py
+  is_this_zero(expr1 - expr2)
+  ```
+
+  `````
+
```
oscarbenjamin, line 530: This page has a lot of discussion about when something should be deprecated. I think that it would be good if the examples used actually match that rationale i.e. neither of these is a good example. In the case of `simplify_this` there is no reason why the function couldn't just be left as a stub that calls `simplify` with a docstring saying that it's recommended to use `simplify` instead. Also in the case of `is_this_zero` there isn't any clear harm in leaving the two argument form there and just advising to use something else in the docs.

oscarbenjamin:
I've been through this and left a few comments but broadly I think this is good. Regardless of the comments above I'd say it's fine to merge apart from two points:

1. It seems that the `pytest.warns` function not being used under `pytest` (to test the stack-level)?
2. The tests have failed under pypy because the stack-level checking is seeing different results there.

I'd rather have a new e.g. `warns_stack_check` function that calls whichever `warns` function is used. I'm not sure what the implications would be of not using pytest's `warns` function when running the tests under `pytest`.

sympy/testing/pytest.py
asmeurer: See the updated PR description. I've updated `warns` to do things that the pytest version does not do, so I didn't think it would be a good idea to continue to allow the pytest one to be used when the tests are run in pytest. Specifically:

- `warns` checks the stacklevel to see if it is set to a level so that the warning shows the user code (can be disabled with `warns(test_stacklevel=False)`). A correctly set stacklevel helps users actually see what code of theirs is triggering the warning.
- For `SymPyDeprecationWarning`, it checks that the new `active_deprecations_target` flag actually corresponds to a real cross-reference target in the active-deprecations.md document (otherwise the URL printed in the warning message won't work).
- `warns` no longer ignores unrelated warnings. This is something that was lamented in several of the `test_pytest.py` tests, which I have updated. There was one instance of deprecated code being used in SymPy that was being silenced by this (I will open an issue about it later, but it's `plot_implicit`, and I had to XFAIL it for now).

doc/src/guides/contributing/deprecations.md
asmeurer: That's fair, although both of these are very similar to real warnings that already exist. There are several of examples of existing deprecations that are just renames for clarity (e.g., `tensorhead` -> `tensor_heads`). Maybe we could say these weren't really worth it to begin with, though. 

The `is_this_zero` example is basically an inverted version of a current deprecation of `Eq` (`Eq(x)` with a single argument is deprecated). I honestly think this one is fine. It's a good example of something that is both confusing and completely unnecessary IMO.

I can definitely try to come up with better examples (suggestions welcome). Or reuse an existing real deprecation as an example, if it is simple enough (although I rather like how these are completely self-contained).

asmeurer:
The pypy issue is because pypy has a different number of stack frames, which is due to the fact that it implements functools.lru_cache in pure Python. This is definitely the sort of annoyance you get with the stacklevel argument. If it becomes even more annoying in the future, we might want to remove it (or alternately, try to come up with some more intelligent logic to automatically set the stacklevel). For now, I'm going to just skip the stacklevel tests for these functions that are cached. The worst that happens when the stacklevel is wrong is you see an unrelated line of code in the warning message, which is also what happens by default when you call `warnings.warn()` without setting the stacklevel (the default stacklevel is 1). And actually the cache messes with warnings anyway (something like `Pow(Basic(), S.One)` can only ever warn once, even if you set the warnings filter to always).

doc/src/guides/contributing/deprecations.md
oscarbenjamin: I've lost count of the number of times I've copied snippets of code from somewhere and seen the deprecation warning about single argument Eq so it was definitely widely used. Having read the rationale for when to deprecate here (which I agree with) I don't think it was a reasonable thing to deprecate.

There's nothing inherently wrong with `Eq(x)` as a shorthand for `Eq(x, 0)` especially since in many contexts where an equation should logically be used sympy allows an expression with the implicit meaning that the expression is equal to zero. In that sense you can view `Eq(expr)` as simply a conversion between two different ways that sympy represents equations.

More importantly though nothing was gained by deprecating the single argument Eq. It didn't make it possible to make other useful changes or prevent the creation of nonsensical objects or anything else. It was just a deprecation for the sake of it. I agree that it would have been a cleaner API design not to allow single argument Eq in the first place but there was no real benefit from trying to change that retrospectively.

As for a better example I would go with something like `integrate(Eq(x, y), t)`. There's a clear rationale there for why it isn't possible to make that work sensibly. There is clear benefit in showing a message to users explaining why that code is probably incorrect. It isn't possible to just "fix the bug" because that requires introducing an integration constant and anyone currently depending on it won't expect that. If we add the symbol then that's incompatible with the rest of what integrate does and also we have to choose what symbol to use which should really be handled by the user. The code is broken and gives mathematically incorrect results in a way that cannot be fixed in general on SymPy's side without the user updating their code in some way. Ideally this should raise an error but since some code might depend on this we start with a deprecation warning.

asmeurer:
Yeah we are only enabling DeprecationWarning in the tests https://github.com/sympy/sympy/blob/2c9c12dfbb708fa27d95165415b05a23e5d4d0e8/sympy/testing/runtests.py#L182 We ought to be enabling (and turning into an error) all warnings. See also https://github.com/sympy/sympy/issues/23061. It's not something I plan to do in this PR, though. 

Actually pending deprecation warnings are something we didn't really discuss here. They are completely hidden by default (even more hidden than `DeprecationWarning`, which at least trigger in interactive sessions). I'm not sure what benefit they provide with those defaults, and I didn't even realize we were using them anywhere.

doc/src/guides/contributing/deprecations.md
asmeurer: I think the boat's already sailed on single-argument Eq. Personally I think it was fine because Eq(x) is just as likely to be a typo, and explicitly writing Eq(x, 0) isn't hard. It's not unlike how a long time ago we used to make `__call__` do substitution but we removed it because people were more likely to mistype multiplication. 

I'll think if there are better examples that can be used here. Like I said I do rather like that these examples are small and self-contained, even if they are imaginary (people can look at the real deprecations if they want a more concrete example).

oscarbenjamin:
> Actually pending deprecation warnings are something we didn't really discuss here

They are discussed above. I suggested using them and you said that you didn't think they were useful.

> I'm not sure what benefit they provide

The purpose of PendingDeprecationWarning is to show up in CI for downstream projects. The idea is that downstream developers should see the warnings but not users and developers should fix them so that users don't see them when they become proper warnings or errors later.

Since the whole point of PendingDeprecationWarning is to show up in CI pytest makes sure to make them visible. At least `bin/test` should do the same.

oscarbenjamin:
Since CI has now passed I'll say that I'm +1 to just merging this and making any further improvements later. I have reviewed the whole diff but I only had minor suggestions (mentioned above).

asmeurer:
I have a quick fix I need to push for the pytest tests. Apparently you can't import `sympy.testing.pytest` in library code or it will break the pytest integration (it gets run too early for the module to detect it is being run inside of pytest). So I moved `ignore_warnings` to sympy.utilities.exceptions, since it has some legitimate uses inside of library code (to prevent duplicate warnings from recursive calls to deprecated functions).

Other than that, I am also OK with merging. If anyone else has any comments on this, please make a new issue or PR.

asmeurer:
The pages are live 

https://docs.sympy.org/dev/guides/contributing/deprecations.html
https://docs.sympy.org/dev/explanation/active-deprecations.html#active-deprecations

The links in the warnings won't work until they are in a release because they use 'latest' instead of 'dev'.

I will go through all the old deprecation removal issues next week and update them.

asmeurer:
@oscarbenjamin if I redo this PR against the 1.10 branch will it be possible to get it in?

oscarbenjamin:
Yes

asmeurer:
https://github.com/sympy/sympy/pull/23080

PR #22904, opened 2022-01-23
multiset_derangements update

smichr:

The case of "covers all but 1" was not tracking the placement of the last element properly so non-derangments were being returned, e.g. the 3rd and 4th elements below have 2 `k`s and no `b` and the 7th and 8th have two `b`s and no `k`:

```python
>>> [''.join(i) for i in multiset_derangements('books')]
['kbsoo', 'ksboo', 'kksoo', 'kskoo', 'oksbo', 'oskbo', 'obsbo', 'osbbo', 'oskob', 'oksob', 'obkob', 'okbob']
```
instead of
```python
>>> [''.join(i) for i in multiset_derangements('books')]
['kbsoo', 'ksboo', 'sbkoo', 'skboo', 'oksbo', 'oskbo', 'okbso', 'obkso', 'oskob', 'oksob', 'osbok', 'obsok']
```

Also, the code to compute the placement of the last two elements of the multiset was not getting called.
#### Other comments

Thanks to @rathmann for communicating these issues!

#### Release Notes

* utilities
    * `multiset_derangements` has an optimized finishing routine
    * corrected a special-case derangement of a multiset where a single repeated element multiplicity is 1 less than the non-repeated elements (like 'mmabc' or 'mmmabcd')
    * `generate_derangements` formerly always emitted derangements from filtered permutations; now, when a multiset is involved, the much faster `multiset_derangements` is called and the ordering of derangments will be different than before.

### DIFF
--- a/sympy/utilities/iterables.py
+++ b/sympy/utilities/iterables.py
@@ -2205,7 +2205,8 @@ def multiset_derangements(s):
     n = len(s)
     # special cases
 
-    # 0) impossible case
+    # 0) one element has more than half the total cardinality of s: no
+    # derangements are possible.
     if mx*2 > n:
         return
 
@@ -2232,8 +2233,11 @@ def multiset_derangements(s):
             yield rv
         return
 
-    # 3) single repeat covers all but 1 of the non-repeats
-    if n - 2*mx == 1 and len(ms.values()) - 1 == n - mx:
+    # 3) single repeat covers all but 1 of the non-repeats:
+    # if there is one repeat then the multiset of the values
+    # of ms would be {mx: 1, 1: n - mx}, i.e. there would
+    # be n - mx + 1 values with the condition that n - 2*mx = 1
+    if n - 2*mx == 1 and len(ms.values()) == n - mx + 1:
         for i in range(len(inonM)):
             i1 = inonM[i]
             ifill = inonM[:i] + inonM[i+1:]
--- a/sympy/utilities/iterables.py
+++ b/sympy/utilities/iterables.py
@@ -2224,14 +2224,15 @@ def multiset_derangements(s):
             ifill = inonM[:i] + inonM[i+1:]
             for j in ifill:
                 rv[j] = M
-            rv[i1] = s[i1]
             for p in permutations([s[j] for j in ifill]):
+                rv[i1] = s[i1]
                 for j, pi in zip(iM, p):
                     rv[j] = pi
+                k = i1
                 for j in iM:
-                    rv[j], rv[i1] = rv[i1], rv[j]
+                    rv[j], rv[k] = rv[k], rv[j]
                     yield rv
-                    i1 = j
+                    k = j
         return
 
     def finish_derangements():
--- a/sympy/utilities/iterables.py
+++ b/sympy/utilities/iterables.py
@@ -1945,7 +1945,10 @@ def has_dups(seq):
     if isinstance(seq, (dict, set, Dict, Set)):
         return False
     unique = set()
-    return any(True for s in seq if s in unique or unique.add(s))
+    try:
+        return any(True for s in seq if s in unique or unique.add(s))
+    except TypeError:
+        return len(seq) != len(list(uniq(seq)))
 
 
 def has_variety(seq):
--- a/sympy/utilities/iterables.py
+++ b/sympy/utilities/iterables.py
@@ -2401,7 +2401,7 @@ def random_derangement(t, choice=None, strict=True):
 
 def generate_derangements(perm):
     """
-    Routine to generate unique derangements or sets or multisets.
+    Routine to generate unique derangements of sets or multisets.
 
     Examples
     ========
--- a/sympy/utilities/iterables.py
+++ b/sympy/utilities/iterables.py
@@ -9,14 +9,12 @@ from itertools import product as cartes # noqa: F401
 from operator import gt
 
 
-# this is the logical location of these functions
-# from sympy.core.compatibility import ordered
-# from sympy.core.compatibility import default_sort_key  # noqa: F401
-from sympy.utilities.misc import as_int
 
+# this is the logical location of these functions
 from sympy.utilities.enumerative import (
     multiset_partitions_taocp, list_visitor, MultisetPartitionTraverser)
 
+from sympy.utilities.misc import as_int
 from sympy.utilities.decorator import deprecated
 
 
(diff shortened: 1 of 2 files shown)


rathmann:
Thanks for the very quick fix.

Still pondering the #3 case.  It seems to work, but I want to study more closely.  

In the meantime here are a few very simple comments.  (I don't think I can do line comments, since they are not changed in this pr.)

docstring for `generate_derangements()`:

Routine to generate unique derangements **or** sets or multisets.

or-> of

In the body of generate derangements, `yield(list(p))` for the output of
itertools.permutations().

Is the list wrapper necessary?

Doesn't `itertools.permutations()` already copy the data?  Is it because you wanted to return lists rather than tuples?

smichr:
In case #3 we move the multiples into subsets of the non-repeated elements. Then permutations of the non-repeats must go into the place where the repeats were, but we must also move the extra non-repeated element somewhere. So after placing a permutation, the extra is moved through the positions containing the nonrepeated elements.
```
abcdEEE -> aEEEbcd
bEEEacd
bEEEcad
bEEEcda
abcdEEE -> aEEEbdc
bEEEadc
bEEEdac
bEEEdca
...
```

The use of `list` in generate derangments is only to keep the output the same. In the case of a multiset, the in-place list is returned from `multiset_derangement` but `generate_derangment` is not advertised to work in-place so a copy is made. I don't think it matters if it's a list or a tuple, however. But before the last change the output was list (coming from `multiset_permutations`).

rathmann:
Thanks.

Agree that it makes sense to keep the list() to make the output more predictable.

I did verify that, as expected,the ordering of the results yielded by generate_derangements() is different.  Probably worth a note in the release notes.  (As well as an indication that it is now a lot faster.)

Thanks for the explanation on the `#3` case.  I *think* I understand it now, but would like to do a few more checks tomorrow.  Very clever, BTW.

smichr:
> Probably worth a note in the release notes. (As well as an indication that it is now a lot faster.)

I added a note in this PR's release notes.

ThePauliPrinciple:
> * correct result for the case when the repeated elements cover all but one of the nonrepeated elements

I'm not sure if these bullet points are guaranteed to keep their order in the final release notes or whether it ends up sorted somehow. In the letter case, I'm not sure it is clear what this sentence is about.

smichr:
> not sure it is clear what this sentence is about.

I changed the tense and description -- I'm trying to say that a bug for the special "case 3" was fixed. Does it make more sense now?

rathmann:
Thanks for the explanation.

Here is my (somewhat long winded) description of what case `#3` is
doing.  This case is not a lot of code, but it is pretty complicated
when you try to write it down.

Case 3

Preconditions (i.e., the tests to enter this case):

1) One element, M, has multiplicity, mx, greater than 1.  All the other
   elements are singletons.
2) The number of singletons is exactly one greater than the multiplicity of M.

In this situation, the mx instances of M can be placed into the mx+1
open slots in mx+1 different ways (differing in which slot is NOT
used).

For each such placement of M, there are mx*mx! ways to place the mx+1
singletons.  The singleton *not* covered by M can be placed in any
slot other than the one it originally occupied (mx choices).  The
remaining singletons can then be placed in mx! ways into the remaining
open slots.

The code to generate these singleton placements are generated as follows:

For each permutation p of the non-constrained singletons:

We intialize the output vector, rv, placing the singletons of p, in
order, into into the positions originally occupied by M in s.  The
constrained singleton is initialized by placing it (in violation of
the derangement constraint) into the index it had occupied in s.

Then, in a loop, the constrained singleton is swapped with successive
elements of p, with a derangement yielded after each such swap.
(Somewhat analogous to the movement pattern in bubble sort.)

This code organization has the performance advantage that the
innermost loop is just a swap.

rathmann:
The point of the above explanation is that I can now see why case `#3` does correctly yield the correct set of derangements for its special case.  (Assuming I am understanding correctly.)

Suggestions/comments, not in any particular order:

1. ` len(ms.values())` in (first line of `#3`).
This is an important concept, the cardinality of the support of s,
and deserves its own variable.  Perhaps:

  `num_distinct = len(ms)`

  (I don't know if `len(ms)` or `len(ms.values())` is to be preferred.)

2. The use of multiset() and has_dups() means that the elements of the argument
   permutation multiset need to be hashable.
   
   For example, x3 = [[1], [2], [3]] fails as an argument to either
   generate_derangements() or multiset_derangements(). 

   This is perfectly reasonable, but it is a new limitation -- x3  does
   work as an argument in the released version of sympy.

   Worth mentioning in the release notes?  Maybe even in the docstrings
   for the functions.

3. I am mildly bothered that generate_derangements() and
   multiset_derangements() are mutually recursive.
   Perhaps factor out the set-specific code as its own function, something like:
   

```python
    def set_derangements(s):
        """Yield the derangements of iterable s.

        All the elements of s are assumed to be distinct.  This is not checked."""
        for p in permutations(s):
            if not any(i == j for i, j in zip(p, s)):
                yield list(p)
```

    I realize that this is a matter of taste.  I would do such a
    refactoring anyhow if/when I put in the ECO implementations.

4. Possible intro comment for case 0.

   0) One element has more than half the total cardinality of s.  No
   derangements are possible.

   (Impossible case sounds (to me) like a "can't happen" bug check.
   Argument 'aab' is a non-buggy case, but returns an empty result.

smichr:
@rathmann , all good comments, thanks. I have made changes in response to them.

BTW, "bubble sort" was the same thought I had when looking at how case #3 worked.

rathmann:
Thanks for your quick and helpful response to my comments of
yesterday.  Here are some (hopefully final) comments on this PR.

1) multiset_derangements: `# 1) singletons` - Shouldn't this now be
   calling `_set_derangements()`?

2) I assume the iterative version of `do()` is off the table, at least
   for now.  If so, I think we can remove the second sentence of the
   docstring to finish_derangements() --  "In non-recursive version,...".

3) Should there be a blank line after the last line of the body of
   do() (before `take = ...`)?

4) Just a question - why subsets() rather than itertools.combinations()?
   Don't they do the same thing?

5) Finally, given the complexity of the code, I think that
   multiset_derangements() would benefit from a roadmap.

   Perhaps something like the following:

   Roadmap: This function first tests for and handles some easier special
   cases (labeled 0 through 3).  It defines 3 inner (nested) functions
   which to handle the general case.

   The primary function is do(), which places all the instances of an
   element at one once.  It uses subsets() to iterate through the possible
   placements.  For each placement it adjusts the output list, rv, to reflect that
   placement, and recursively calls itself (yield from) to place the
   instances of the next element.

   iopen() is a simple helper to decide which positions of rv are
   currently open for the argument element, while
   finish_derangements() speeds up the placement of the last two
   elements.

smichr:
It's updated again with some additions. The clean-up of `rv` is not stricly necessary, but it is probably a good idea to make sure the user is using the output of the routine properly.

rathmann:
```python
# ... either there are two elements with multiplicity
# of mx or else there is only one with multiplicity mx
```

Not sure I understand/believe this.  In particular, 'aaabbb' triggers
case 3, but so does 'aaabbc'.

**Other than that, this PR looks good to me.  Thanks.**

I notice that subsets() is now mostly a passthrough to itertools, so
not a lot of motivation to switch.  There is a little extra overhead
caused by an extra layer of yield in subsets, but it would be better
(TODO, eventually) to just optimize subsets.

smichr:
> In particular, 'aaabbb' triggers
> case 3, but so does 'aaabbc'.

Yes. In both cases, aaa will swap places with bbb or bbc which will then get placed in aaa's positions as multiset permutations. bbb has only one such permutation while bbc can get placed as bbc, bcb, or cbb.

smichr:
I will commit this so we can at least have the right-working version in place. Thanks for the review iterations :-)

rathmann:
> smichr merged commit fd40404 into sympy:master 12 hours ago

Great!  Congratulations.  Thanks for all the quick updates.

Issue #10390, opened 2016-01-14
O(1, (n, oo)) is contained in O(sin(n), (n, oo))


gxyd:
```
>>> n = Symbol('n', integer=True, positive=True)
>>> e1 = O(1, (n, oo))
>>> e2 = O(sin(n), (n, oo))
>>> e2 in e1
True         # this is correct
>>> e1 in e2
True         # this should be false
```

tokencolour:
Hello @gxyd,
I read [this](http://docs.sympy.org/dev/modules/series.html#order-terms).  I couldn't quite understand why O(x + x**2) is O(x). Is there a condition imposed on x?
```
>>> O(x+x**2)
O(x)
```
and what the second arg (the tuple (x, 0) or (x, oo)) means
```
>>> O(x + x**2, (x, 0))
O(x)

>>> O(x + x**2, (x, oo))
O(x**2, (x, oo))
```
Can you please recommend to me some reading material so that I can comprehend this?

gxyd:
You can read this  https://en.wikipedia.org/wiki/Big_O_notation . I don't know the fix right now. 
In particular in mathematics, `BigO` is "not just" about taking limit behaviour at `oo`(which is often done in computer science) that is done specifying a tuple `(x, oo)`.  We can find the limiting behaviour at any general point as `x tends to a` using tuple `(x, a)`

tokencolour:
Thank you!

gxyd:
This will probably be solved after the implementation of `#TODO` left for  `AccumBounds.union` method. Since `limit(1/sin(n), (n, oo))` returns `<-oo, oo>` instead of `<-oo, -1> U <1, oo>`.

tokencolour:
I debugged a little to find out that the reason why `limit(1/sin(n), (n, oo))` returns `<-oo, oo>` instead of `<-oo, -1> U <1, oo>` . It seemed to me that the problem is [here](https://github.com/sympy/sympy/blob/master/sympy/calculus/util.py#L561). I couldn't understand how this is related to `AccumBounds.union` being incompletely implemented. Can you please help me, @gxyd?

Also, I am interested to complete the implementation of `union` in `AccumulationBounds` . One test case I found, where the `union` fails is:
```
AccumBounds(-oo, -1).union(AccumBounds(0, oo))
```

gxyd:
>  I couldn't understand how this is related to AccumBounds.union being incompletely implemented.

[...]

Issue #20357, opened 2020-10-29
Could AccumBounds use maximum/minimum?


asmeurer:
We have

```py
>>> limit(sin(x) + cos(x), x, oo)
AccumBounds(-2, 2)
```

But it is actually limited by +/-sqrt(2). The `maximum` and `minimum` functions can compute this:

```py
>>> maximum(sin(x) + cos(x), x)
sqrt(2)
>>> minimum(sin(x) + cos(x), x)
-sqrt(2)
```

Can we use this in the general case?

Also, `sin(oo) + cos(oo)` has the same problem, but fixing that would be harder, because it's basically `AccumBounds(1, 1) + AccumBounds(1, 1)`.

oscarbenjamin:
Perhaps related is a recent SO question asking to be able to merge linear combinations of sin and cos having the same frequency but different amplitude and phase. There doesn't seem to be a sympy function for that:
https://stackoverflow.com/questions/64481861/python-sympy-superposition-of-several-waves

asmeurer:
minimum and maximum just take the derivative and set it equal to 0. But I guess it fails for more complicated examples, like the one in the SO question. So maybe such a function could be used inside of solve to help it there, if it isn't good enough for such cases already. Or maybe there's a more general way to improve solve.

PR #22905, opened 2022-01-23
Series: Improved consistency between basic limits

### LINKED ISSUE #14525: limit: SymPy returns unevaluated limit
The following limit comes out unevaluated:
```
In [1]: func = sin(x)**2 - cos(x) + tan(x)*csc(x)
In [2]: limit(func, x, oo)
Out[2]:
     /   2                            \
 lim \sin (x) - cos(x) + tan(x)*csc(x)/
x->oo
```

anutosh491:

Fixes #14525 
While I was trying to approach another issue , I realized that some consistency between some basic limits would be better overall .

On Master
```
>>> limit(tan(x), x, oo)
AccumBounds(-oo, oo)
>>> limit(cot(x), x, oo)
cot(oo)
>>> limit(sec(x), x, oo)
nan
>>> limit(csc(x), x, oo)
Limit(csc(x), x, oo, dir='-')
```

whereas the `cot` case could maybe return the output in `AccumBounds` as `cot(oo)` is anyways a bit vague and also the `sec` and `csc` could maybe have similar outputs .

Now different answers are returned for the latter because
```
>>> sec(1/x).as_leading_term(x)
nan
>>> csc(1/x).as_leading_term(x)
AccumBounds(-oo, oo)
```

Currently `nan` would be a better output for the above cases as `AccumBounds(-oo, oo)` isn't the correct representation for these (maybe after we have a Union method for `AccumBounds` as being discussed in #10390  we could think of something like `<-oo, -1> U <1, oo>` as the output but till then `nan` could be returned .(Or if we could comprise on the answer I could maybe return `AccumBounds(-oo, oo)` for both !

Hence I implemented the `_eval_as_leading_term()` for the `csc` function and we have the following on the committed branch

```
>>> limit(tan(x), x, oo)
AccumBounds(-oo, oo)
>>> limit(cot(x), x, oo)
AccumBounds(-oo, oo)
>>> limit(sec(x), x, oo)
nan
>>> limit(csc(x), x, oo)
nan
```

#### Release Notes

* series
  * Improved Consistency between basic limits and added _eval_as_leading_term for csc function.


anutosh491:
ping @0sidharth

anutosh491:
I felt returning unevaluated `self` would do better than `nan` as an answer for the limits reported above because 
The test file shows that returning unevaluated self for limits which can't be computed.... seems as a better option currently !(I either see a raise or a `Limit` object being returned in such cases), hence I am making the second commit

faze-geek:
I was also exploring `AccumBounds()` previously once ,I would like to report some more issues related to the one's your commit is solving .You could possibly work on these in some future commits .Even if they cannot be evaluated correctly SymPy could consistently return self .

```
>>> limit(sin(x)**2 + cos(x)**2,x,oo)
AccumBounds(0, 2)
>>> limit(sec(x)**2 ,x,oo)
nan
>>> limit(sec(x)**2 - 1,x,oo)
0
>>> limit(tan(x)**2,x,oo)
Limit(tan(x)**2, x, oo, dir='-')
>>> limit(csc(x)**2 - 1,x,oo)
Limit(csc(x)**2 - 1, x, oo, dir='-')
>>> limit(sin(x)**2 + csc(x)**2,x,oo)
Limit(sin(x)**2 + csc(x)**2, x, oo, dir='-')
>>> limit(cos(x)**2 + sec(x)**2,x,oo)
0
```
Other than these there are multiple` NotImplementedError` errors which I tackled while trying various identities/ relations .

anutosh491:
> I was also exploring `AccumBounds()` previously once ,I would like to report some more issues related to the one's your commit is solving .You could possibly work on these in some future commits .Even if they cannot be evaluated correctly SymPy could consistently return self .
> Other than these there are multiple` NotImplementedError` errors which I tackled while trying various identities/ relations .

Thanks for pointing these out ! If we want consistency overall it's better to stick to `AccumBounds(-oo , oo)` as the output for something like `sec(1/x).as_leading_term(x)` and `csc(1/x).as_leading_term(x)` rather than `nan` . Btw a reason why couple of these don't work were pointed out in #20357 and that might take a while too be implemented hence I might not tackle those initially in this Pr untill need arises !! Thank you again .

anutosh491:
ping @jksuom

anutosh491:
@jksuom , @0sidharth . The Pr isn't a heavy one and could go in after a set of reviews ! I would like to know what could we expect as an output for `sec(1/x).as_leading_term(x) and csc(1/x).as_leading_term(x)`  at this current state in sympy ??

anutosh491:
ping @jksuom for review !

jksuom:
I am not sure what the leading terms of sec(1/x) and csc(1/x) should be but it looks like AccumBounds(-oo, oo) might work. It can be regarded as "real nan".

anutosh491:
> I am not sure what the leading terms of sec(1/x) and csc(1/x) should be but it looks like AccumBounds(-oo, oo) might work. It can be regarded as "real nan".

okay , though I felt for a while that `nan` would work cause otherwise we would want to change  its trig relatives  like `tan(1/x)` too . Currently we have as follows
```
>>> tan(1/x).as_leading_term(x)
tan(1/x)
>>> limit(tan(x), x, oo)
AccumBounds(-oo, oo)
```
If we change this this `tan(1/x).as_leading_term(x)` to` AccumBounds(-oo, oo)` , the current code would give back `self ` due to line 316 - 319
```
        else:
            if coeff.has(S.Infinity, S.NegativeInfinity, S.ComplexInfinity, ):
                return self
```
I would prefer the limit giving out `Accumbounds` rather than the leading_term methods . What are your thoughts for the same !

anutosh491:
I can see lot of inconsistencies here and will try to address them
```
>>> x = Symbol('x')
>>> tan(1/x).as_leading_term(x)
tan(1/x)
>>> sec(1/x).as_leading_term(x)
nan
>>> csc(1/x).as_leading_term(x)
AccumBounds(-oo, oo)
>>>
>>>
>>> limit(tan(x), x, oo)
AccumBounds(-oo, oo)
>>> limit(sec(x), x, oo)
nan
>>> limit(csc(x), x, oo)
Limit(csc(x), x, oo, dir='-')
```

I would like to see the results as following , consedering (implementing union function for `AccumBounds` is still under discussion so we can't have something like `AccumBounds(-oo, -1).union(AccumBounds(1, oo))` )
```
>>> limit(tan(x), x, oo)
AccumBounds(-oo, oo)
>>>
>>> limit(sec(x), x, oo)
Limit(csc(x), x, oo, dir='-')
>>>
>>> limit(csc(x), x, oo)
Limit(csc(x), x, oo, dir='-')
```

anutosh491:
I've made changes as denoted in the comment above , I am happy with the changes and can't find a lot to improve on ! Any suggestions are welcome!

anutosh491:
> I was also exploring `AccumBounds()` previously once ,I would like to report some more issues related to the one's your commit is solving .You could possibly work on these in some future commits .Even if they cannot be evaluated correctly SymPy could consistently return self .
> 
> ```
> >>> limit(sec(x)**2 ,x,oo)
> nan
> >>> limit(sec(x)**2 - 1,x,oo)
> 0
> >>> limit(cos(x)**2 + sec(x)**2,x,oo)
> 0
> ```

The pr fixes addresses the wrong answers which have been pointed out here too 
```
>>> limit(sec(x)**2 ,x,oo)
Limit(sec(x)**2, x, oo, dir='-')

>>> limit(sec(x)**2 - 1,x,oo)
Limit(sec(x)**2 - 1, x, oo, dir='-')

>>> limit(cos(x)**2 + sec(x)**2,x,oo)
Limit(cos(x)**2 + sec(x)**2, x, oo, dir='-')
```

sympy/functions/elementary/trigonometric.py
```diff
@@ -1826,13 +1828,16 @@ def taylor_term(n, x, *previous_terms):
...
+        from sympy.calculus.accumulationbounds import AccumBounds
         arg = self.args[0]
         x0 = arg.subs(x, 0).cancel()
         n = (x0 + S.Pi/2)/S.Pi
         if n.is_integer:
             lt = (arg - n*S.Pi + S.Pi/2).as_leading_term(x)
             return (S.NegativeOne**n)/lt
-        return self.func(x0)
+        if x0 is S.ComplexInfinity:
+            return AccumBounds(S.NegativeInfinity, S.Infinity)
```
jksuom, line 1839: AccumBounds is essentially extended real. It should not be used for expansions and limits at zoo.

anutosh491: True, this slipped out of my mind . I'll fix that !

anutosh491:
benchmarks are failing as reported in #23023 , other than that I think would work fine !

anutosh491:
ping @jksuom

sympy/functions/elementary/trigonometric.py
```diff
@@ -1826,13 +1828,19 @@ def taylor_term(n, x, *previous_terms):
...
         arg = self.args[0]
         x0 = arg.subs(x, 0).cancel()
         n = (x0 + S.Pi/2)/S.Pi
         if n.is_integer:
             lt = (arg - n*S.Pi + S.Pi/2).as_leading_term(x)
             return (S.NegativeOne**n)/lt
-        return self.func(x0)
+        if x0 is S.ComplexInfinity:
+            x0 = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
+        if x0 in [S.Infinity, S.NegativeInfinity]:
```
jksuom, line 1841: It would be more efficient to use a tuple.

sympy/series/tests/test_limits.py
```diff
@@ -65,6 +65,10 @@ def test_basic1():
     assert limit(1/tan(x**3), x, (2*pi)**Rational(1, 3), dir="-") is -oo
     assert limit(1/cot(x)**3, x, (pi*Rational(3, 2)), dir="+") is -oo
     assert limit(1/cot(x)**3, x, (pi*Rational(3, 2)), dir="-") is oo
+    assert limit(tan(x), x, oo) == AccumBounds(S.NegativeInfinity, S.Infinity)
+    assert limit(cot(x), x, oo) == AccumBounds(S.NegativeInfinity, S.Infinity)
+    assert limit(sec(x), x, oo) == Limit(sec(x), x, oo, dir='-')
```
jksuom, line 70: Could this be AccumBounds as well?

anutosh491: This has been pointed out before , I'll try to locate where( was a discussion on `AccumBounds` by Aaron ) . As per what I can recollect `limit(sec(x), x, oo)` would be `AccumBounds(-oo, -1).union(AccumBounds(-oo, -1))` but as union method for `Accumbounds` is under discussion(#10390) in we should restrict from returning the wrong answer (which in this case would be `AccumBounds(-oo, oo)` untill then !

sympy/functions/elementary/trigonometric.py
anutosh491: Cool , I'll do that otherwise looks good to me !

sympy/series/tests/test_limits.py
anutosh491: I've made requested changes in the latest commit ! LGTM and would be glad if you could confirm .

anutosh491:
Optional_dependencies are failing on all prs today ! The same is the case here , I don't think it has anything to do with this Pr in particular!

anutosh491:
ping @jksuom for review !

anutosh491:
ping @jksuom . LGTM !

anutosh491:
ping @jksuom . LGTM !

sympy/series/tests/test_limits.py
jksuom: Why does AccumBounds not appear here?

anutosh491: okay I'll try to convince you again maybe and if you think that AccumBounds could be better I'll do the needful
On master we have the following 
```
>>> limit(csc(x), x, oo)
Limit(csc(x), x, oo, dir='-')
>>> limit(sec(x), x, oo)
nan
```
Now the correct answer for this is `AccumBounds(-oo, -1).Union(AccumBounds(1, oo))` and currently sympy doesn't have any method to represent this union of AccumBound Intervals and the best approach for this is under discussion in the issue reported above, hence the best we could do is return `AccumBounds(-oo, oo)` . As `AccumBounds(-oo, oo)` doesn't completely represent the correct answer and also this being a basic limit where I feel sympy shouldn't leave any doubt of error ......I thought returning self would be a better choice in this place !

But if you think otherwise , I'll do the needful as soon as possible .

jksuom: AccumBounds is not expected to give an exact description of the set of accumulation points. There may be gaps, even big gaps. The union of two AccumBounds is more accurate than a single one but that can probably seldom be made use of. Hence the simple result seems better to me.

anutosh491: I have made the requested change in the latest commit ,would be glad to know your thoughts on the same !

anutosh491:
Now that we have Introduced leading terms for `sin/cos/sec/csc` for the expression `1/x`  , it tells us that we could maybe include these two too !
```
>>> tan(1/x).as_leading_term(x)
AccumBounds(-oo, oo)
>>> cot(1/x).as_leading_term(x)
AccumBounds(-oo, oo)
```
I'll make the commit , obviously this won't be disturbing anything . if there is some reason why this shouldn't be done or doesn't seem reasonable , i'll revert back .

sympy/series/limits.py
```diff
@@ -314,6 +314,8 @@ def set_signs(expr):
                     return r
         else:
             if coeff.has(S.Infinity, S.NegativeInfinity, S.ComplexInfinity):
+                if isinstance(coeff, AccumBounds) and ex == S.Zero:
```
jksuom, line 317: AccumBounds should probably be tested before infinities. It looks like also finite ones could be returned as is.

anutosh491: Yeah , correct I realized this should be consedered too ! Thanks , I have made the change.

anutosh491:
@jksuom please could you review this one too as and when you find time?

sympy/functions/elementary/trigonometric.py
```diff
@@ -1826,13 +1840,19 @@ def taylor_term(n, x, *previous_terms):
...
         n = (x0 + S.Pi/2)/S.Pi
         if n.is_integer:
             lt = (arg - n*S.Pi + S.Pi/2).as_leading_term(x)
             return (S.NegativeOne**n)/lt
-        return self.func(x0)
+        if x0 is S.ComplexInfinity:
+            x0 = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
+        if x0 in (S.Infinity, S.NegativeInfinity):
+            return AccumBounds(S.NegativeInfinity, S.Infinity)
+        return self.func(x0) if x0.is_finite else self
```
0sidharth, line 1855: Why is the else condition needed here (and below)?

sympy/functions/elementary/tests/test_trigonometric.py
```diff
@@ -1285,10 +1285,13 @@ def test_leading_terms():
...
     assert cos(S.Half).as_leading_term(x) == cos(S.Half)
+    assert sec(1/x).as_leading_term(x) == AccumBounds(S.NegativeInfinity, S.Infinity)
+    assert csc(1/x).as_leading_term(x) == AccumBounds(S.NegativeInfinity, S.Infinity)
+    assert tan(1/x).as_leading_term(x) == AccumBounds(S.NegativeInfinity, S.Infinity)
+    assert cot(1/x).as_leading_term(x) == AccumBounds(S.NegativeInfinity, S.Infinity)
 
     for func in [tan, cot]:
-        for a in (1/x, S.Half):
-            eq = func(a)
+            eq = func(S.Half)
```
0sidharth, line 1294: This should be re-indented to 4 spaces. Also, can `sin` and `cos` be included in the loop as well?

sympy/functions/elementary/trigonometric.py
anutosh491: True I had the given a thought about this while making the latest commit here as I couldn't really find a case involving the else check but i took inspiration for ` sin, cos, tan` functions where it has been used, though I couldn't really find any test where the following is encountered

anutosh491: I'll try to think of some case where that might be encountered but as of now I can't think of any !

anutosh491:
The `AccumBounds` suggestions work extremely well as I tried out many functions randomly....... some of which I've added as tests .It doesn't only help us with cases whose answer we expect to have as `AccumBounds(-oo, oo)` but also for all generic cases. I've tried to add tests that cater to these general cases too .
The tests added are a bit random to see if it performs badly for any case which doesn't really happen
```
    assert limit(sin(x)**2 - cos(x) + tan(x)*csc(x), x, oo) == AccumBounds(S.NegativeInfinity, S.Infinity)
    assert limit(sin(x)**2 - cos(x) + sin(x)*cot(x), x, oo) == AccumBounds(S.NegativeInfinity, S.Infinity)
    assert limit(cot(x) - tan(x)**2, x, oo) == AccumBounds(S.NegativeInfinity, S.Infinity)
    assert limit(cos(x) - tan(x)**2, x, oo) == AccumBounds(S.NegativeInfinity, S.One)
    assert limit(sin(x) - tan(x)**2, x, oo) == AccumBounds(S.NegativeInfinity, S.One)
    assert limit(cos(x)**2 - tan(x)**2, x, oo) == AccumBounds(S.NegativeInfinity, S.One)
    assert limit(tan(x)**2 + sin(x)**2 - cos(x), x, oo) == AccumBounds(-S.One, S.Infinity)
```
Some graphs for reference 
image
image

anutosh491:
Obviously we can find cases here and there where we would encounter approximations rather than bulls eye answers like one of the tests I've added `limit(sin(x) - tan(x)**2, x, oo) == AccumBounds(S.NegativeInfinity, S.One)`
Graph for reference
image
Graphically we can see that the highest point is approximately `0.381` but as `AccumBounds` isn't expected to be too strict I'm happy with the answer  `AccumBounds(S.NegativeInfinity, S.One)` . All answers above will only improve once we've addressed 
Union method for `AccumBounds` which is under discussion as of now .

anutosh491:
@jksuom does this look good to you ?

anutosh491:
CC: @0sidharth for reviews . This looks good to me !

sympy/functions/elementary/tests/test_trigonometric.py
```diff
@@ -1285,11 +1286,14 @@ def test_leading_terms():
...
-            eq = func(a)
-            assert eq.as_leading_term(x) == eq
+    assert sec(1/x).as_leading_term(x) == AccumBounds(S.NegativeInfinity, S.Infinity)
+    assert csc(1/x).as_leading_term(x) == AccumBounds(S.NegativeInfinity, S.Infinity)
+    assert tan(1/x).as_leading_term(x) == AccumBounds(S.NegativeInfinity, S.Infinity)
+    assert cot(1/x).as_leading_term(x) == AccumBounds(S.NegativeInfinity, S.Infinity)
+
+    for func in (sin, cos, tan, cot):
+        eq = func(S.Half)
+        assert eq.as_leading_term(x) == eq
```
jksuom, line 1296: What is the purpose of this test?

anutosh491: Ohh it was already there , I wasn't too sure if we would like to keep this test or get rid of it and even improve it somehow, and Sidharth ended up suggesting me to add `sin` and `cos` function too ! Originally we were just dealing with `tan` and `cot`

anutosh491: Maybe we could keep it too ! What are your thoughts on the same ?

anutosh491: CC @jksuom

jksuom: eq is a constant so it is obviously also the leading term. That is what as_leading_term returns. It is not necessary to test it with different kinds of constants.

anutosh491:
CC: @jksuom , does this look good to you now ?

anutosh491:
CC: @jksuom @0sidharth . Do you'll see any scope of improvement in this Pr ? this looks good to me !

jksuom:
This is good to go, thanks. `sin(n*pi)` is 0.

anutosh491:
> This is good to go, thanks. `sin(n*pi)` is 0.
YAYAY ! Thanks for the merge and thanks for the reviews on this @jksuom and @0sidharth too !

Yeah I agree but even `cos(n*pi/2)` is either of `-1, 0 or 1` . That's what I was trying to get your attention at . That just because it doesn't auto simplify/evaluate on it's own like `sin(n*pi)` does to `0` we return `AccumBounds` for the `cos` case !!

PR #22906, opened 2022-01-23
printing/latex: latex printing support for Idx class

ThePauliPrinciple:
Improved LaTeX printing support for Idx class. Currently the
printing for this class is too verbose for the context it is used.

#### Release Notes

* printing
  * Less verbose LaTeX printing for `Idx` class


oscargus:
Basic idea is good! Some minor things:
* Should one print a single value differently, like `Idx('i', 0)`?
* I'm a bit doubtful about the `..`, but cannot really make up my mind if e.g `\ldots` is better or not, `..` is for sure more compact

ThePauliPrinciple:
> * Should one print a single value differently, like `Idx('i', 0)`?

I considered this, but I think this would be confusing, since it will look like you are taking the first index of `i`.

> * I'm a bit doubtful about the `..`, but cannot really make up my mind if e.g `\ldots` is better or not, `..` is for sure more compact

I'd be happy with any symbol as long as it is clear and compact, I'll try a few and see what works.

ThePauliPrinciple:
https://en.wikipedia.org/wiki/Interval_(mathematics)

Wiki quotes `..` as an acceptable notation for integer intervals

oscargus:
LGTM!

ThePauliPrinciple:
Thanks for reviewing!

asmeurer:
Does Idx specify whether the range is half-open or closed (i.e., whether the upper endpoint is included)? I would expect `..` in math notation to mean a closed range, whereas a Python slice is half-open, and SymPy often follows Python conventions.

ThePauliPrinciple:
Currently this is how it is printed
afbeelding

PR #22910, opened 2022-01-24
Added Reshape operator for array expressions

Upabjojr:
Added Reshape operator for array expressions

NO ENTRY

### DIFF
--- a/sympy/tensor/array/expressions/array_expressions.py
+++ b/sympy/tensor/array/expressions/array_expressions.py
@@ -16,6 +16,7 @@ from sympy.core.mul import Mul
 from sympy.core.singleton import S
 from sympy.core.sorting import default_sort_key
 from sympy.core.symbol import (Dummy, Symbol)
+from sympy.matrices.common import MatrixCommon
 from sympy.matrices.expressions.diagonal import diagonalize_vector
 from sympy.matrices.expressions.matexpr import MatrixExpr
 from sympy.matrices.expressions.special import ZeroMatrix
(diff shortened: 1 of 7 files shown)


sympy/tensor/array/expressions/array_expressions.py
```diff
@@ -1458,6 +1459,35 @@ def as_explicit(self):
...
+    def expr(self):
+        return self._expr
+
+    def as_explicit(self):
+        ee = self.expr.as_explicit()
+        if isinstance(ee, MatrixCommon):
+            from sympy import Array
+            ee = Array(ee)
+        elif isinstance(ee, MatrixExpr):
+            return self
```
oscarbenjamin, line 1487: Why would `ee` still be `MatrixExpr` here (after the `as_explicit` call above)?

sympy/tensor/array/expressions/array_expressions.py
```diff
@@ -1458,6 +1459,35 @@ def as_explicit(self):
         return tensorcontraction(self.expr.as_explicit(), *self.contraction_indices)
 
 
+class Reshape(_CodegenArrayAbstract):
+
+    def __new__(cls, expr, shape):
+        expr = _sympify(expr)
+        if not isinstance(shape, Tuple):
+            shape = Tuple(*shape)
```
ThePauliPrinciple, line 1467: Do I understand correctly that the object is created even if the new shape does not match the original shape?

```python
Eq(Mul(*expr.shape), Mul(*shape)) != False
```
Should allow any unknown symbolic sizes, but would not allow e.g. 
```python
from sympy import Eq, Symbol, Mul
x= Symbol('x', positive = True)
shape1=(x,)
shape2=(x,2)
Eq(Mul(*shape1),Mul(*shape2)).doit() != False
```

Upabjojr: > Do I understand correctly that the object is created even if the new shape does not match the original shape?

Yes, I will add a check.

sympy/tensor/array/expressions/array_expressions.py
Upabjojr: The shape of the matrix can be symbolic. In such case, `.as_explicit()` will not evaluate.

PR #22920, opened 2022-01-25
make limited precision intersections more robust

### LINKED ISSUE #12598: make ray intersection more robust
The intersection of two rays is being rejected:

```
def bisectors(p, prec=None):
    """Return angle bisectors of a polygon. If prec is given
    then approximate the point defining the ray to that precision.

    The distance between the points defining the bisector ray is 1.

    Examples
    ========

    >>> from sympy import Polygon, Point
    >>> p = Polygon(Point(0, 0), Point(2, 0), Point(1, 1), Point(0, 3)
    >>> for k, v in bisectors(p,2).items():
    ...     print (k, v)
    ...
    (Point2D(2, 0), Ray2D(Point2D(2, 0), Point2D(1.1, 0.38)))
    (Point2D(0, 3), Ray2D(Point2D(0, 3), Point2D(0.23, 2.0)))
    (Point2D(0, 0), Ray2D(Point2D(0, 0), Point2D(0.71, 0.71)))
    (Point2D(1, 1), Ray2D(Point2D(1, 1), Point2D(0.19, 0.42)))
    """
    b = {}
    pts = list(p.args)
    pts.append(pts[0])  # close it
    for v, a in p.angles.items():
        i = pts.index(v)
        ray = Ray(pts[i], pts[i + 1]).rotate(a/2, v)
        # normalize so distance from p1 to p2 is 1
        dir = ray.direction
        ray = Ray(ray.p1, ray.p1 + dir/dir.distance((0,0)))
        if prec is not None:
            ray = Ray(ray.p1, ray.p2.n(prec))
        b[v] = ray
    return b

p=Polygon((0,0),(3,0),(4,10),(0,1))
b = bisectors(p, 2)
for s in p.sides:
 r1, r2 = [b[i] for i in s.args]
 s.args, r1.intersection(r2), Line(r1).intersection(Line(r2))
```
Output is:
```
((Point2D(0, 0), Point2D(3, 0)), [Point2D(1.6, 1.6)], [Point2D(1.6, 1.6)])
((Point2D(3, 0), Point2D(4, 10)), [], [Point2D(1.7, 1.4)])
((Point2D(0, 1), Point2D(4, 10)), [], [Point2D(1.5, 0.68)])
((Point2D(0, 0), Point2D(0, 1)), [], [Point2D(0.82, 0.82)])
```

(The last 3 should have given ray intersection identical to the Line-wrapped ray intersections.)

smichr:

fixes #12598

Containment checks can fail when limited precision points are used with geometry. If there are floats present a sanity check on the parametric values at the intersection are now checked.

```python
>>> l1=Line((0,0),(1,1));l2=Segment((-1,1),(0,-1)).n(2)
>>> l1.intersection(l2)
[Point2D(-0.33, -0.33)]
>>> pt = _[0]
>>> l1.contains(pt)
True
>>> l2.contains(pt)
False
```
Previously the routine concluded that since the pt was not contained in both entities, it must not be an intersection point. But if the non-Float version of the line and pt are used we see that the intersection is valid:
```python
>>> Segment((-1,1),(0,-1)).contains((Rational(-1,3),Rational(-1,3)))
True
```
So now, if the lines contain Float, a check is done on the parameters to see that they both indicate containment: rays must have a non-negative parameter and segments must have a parameter between 0 and 1.

#### Release Notes

* geometry
    * linear intersections of enitities defined in terms of Floats are now more robust

### DIFF
--- a/sympy/geometry/line.py
+++ b/sympy/geometry/line.py
@@ -20,7 +20,7 @@ Segment3D
 from sympy.core.containers import Tuple
 from sympy.core.evalf import N
 from sympy.core.expr import Expr
-from sympy.core.numbers import Rational, oo
+from sympy.core.numbers import Rational, oo, Float
 from sympy.core.relational import Eq
 from sympy.core.singleton import S
 from sympy.core.sorting import ordered
@@ -36,6 +36,7 @@ from sympy.logic.boolalg import And
 from sympy.matrices import Matrix
 from sympy.sets.sets import Intersection
 from sympy.simplify.simplify import simplify
+from sympy.solvers.solvers import solve
 from sympy.solvers.solveset import linear_coeffs
 from sympy.utilities.decorator import deprecated
 from sympy.utilities.exceptions import SymPyDeprecationWarning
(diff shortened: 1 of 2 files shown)


oscargus:
Feel free to self-merge if you are done and happy.

PR #22924, opened 2022-01-25
Fix `__slots__` issues (overlaps, broken inheritance)

### LINKED ISSUE #22922: Some overlapping and ineffective `__slots__`
```shell
$ slotscheck -v -m sympy
ERROR: 'sympy.codegen.ast:FloatType' defines overlapping slots.
       - 'name' is also defined in 'sympy.codegen.ast:Type'
ERROR: 'sympy.codegen.ast:FloatType' has slots but superclass does not.
       - 'sympy.codegen.ast:FloatBaseType' superclass has no slots.
ERROR: 'sympy.codegen.ast:FunctionDefinition' defines overlapping slots.
       - 'attrs' is also defined in 'sympy.codegen.ast:FunctionPrototype'
       - 'attrs' is also defined in 'sympy.codegen.ast:Node'
       - 'name' is also defined in 'sympy.codegen.ast:FunctionPrototype'
       - 'parameters' is also defined in 'sympy.codegen.ast:FunctionPrototype'
       - 'return_type' is also defined in 'sympy.codegen.ast:FunctionPrototype'
ERROR: 'sympy.codegen.ast:FunctionPrototype' defines overlapping slots.
       - 'attrs' is also defined in 'sympy.codegen.ast:Node'
ERROR: 'sympy.codegen.ast:IntBaseType' defines overlapping slots.
       - 'name' is also defined in 'sympy.codegen.ast:Type'
ERROR: 'sympy.codegen.ast:Variable' defines overlapping slots.
       - 'attrs' is also defined in 'sympy.codegen.ast:Node'
ERROR: 'sympy.codegen.ast:_SizedIntType' defines overlapping slots.
       - 'name' is also defined in 'sympy.codegen.ast:IntBaseType'
       - 'name' is also defined in 'sympy.codegen.ast:Type'
ERROR: 'sympy.codegen.fnodes:Subroutine' defines overlapping slots.
       - 'attrs' is also defined in 'sympy.codegen.ast:Node'
ERROR: 'sympy.codegen.matrix_nodes:MatrixSolve' has slots but superclass does not.
       - 'sympy.matrices.expressions.matexpr:MatrixExpr' superclass has no slots.
ERROR: 'sympy.concrete.products:Product' defines overlapping slots.
       - 'is_commutative' is also defined in 'sympy.concrete.expr_with_limits:ExprWithLimits'
ERROR: 'sympy.concrete.products:Product' has slots but superclass does not.
       - 'sympy.concrete.expr_with_intlimits:ExprWithIntLimits' superclass has no slots.
ERROR: 'sympy.concrete.summations:Sum' defines overlapping slots.
       - 'is_commutative' is also defined in 'sympy.concrete.expr_with_limits:ExprWithLimits'
ERROR: 'sympy.concrete.summations:Sum' has slots but superclass does not.
       - 'sympy.concrete.expr_with_limits:AddWithLimits' superclass has no slots.
       - 'sympy.concrete.expr_with_intlimits:ExprWithIntLimits' superclass has no slots.
ERROR: 'sympy.core.numbers:Integer' defines overlapping slots.
       - 'p' is also defined in 'sympy.core.numbers:Rational'
ERROR: 'sympy.geometry.polygon:RegularPolygon' has slots but superclass does not.
       - 'sympy.geometry.polygon:Polygon' superclass has no slots.
       - 'sympy.geometry.entity:GeometrySet' superclass has no slots.
       - 'sympy.geometry.entity:GeometryEntity' superclass has no slots.
       - 'sympy.sets.sets:Set' superclass has no slots.
ERROR: 'sympy.integrals.integrals:Integral' defines overlapping slots.
       - 'is_commutative' is also defined in 'sympy.concrete.expr_with_limits:ExprWithLimits'
ERROR: 'sympy.integrals.integrals:Integral' has slots but superclass does not.
       - 'sympy.concrete.expr_with_limits:AddWithLimits' superclass has no slots.
ERROR: 'sympy.polys.agca.extensions:ExtensionElement' has slots but superclass does not.
       - 'sympy.polys.domains.domainelement:DomainElement' superclass has no slots.
ERROR: 'sympy.polys.domains.gaussiandomains:GaussianElement' has slots but superclass does not.
       - 'sympy.polys.domains.domainelement:DomainElement' superclass has no slots.
ERROR: 'sympy.polys.domains.modularinteger:ModularInteger' has slots but superclass does not.
       - 'sympy.polys.domains.domainelement:DomainElement' superclass has no slots.
ERROR: 'sympy.polys.domains.mpelements:ComplexElement' has slots but superclass does not.
       - 'sympy.polys.domains.domainelement:DomainElement' superclass has no slots.
ERROR: 'sympy.polys.domains.mpelements:RealElement' has slots but superclass does not.
       - 'sympy.polys.domains.domainelement:DomainElement' superclass has no slots.
ERROR: 'sympy.polys.polyclasses:ANP' has slots but superclass does not.
       - 'sympy.core.sympify:CantSympify' superclass has no slots.
ERROR: 'sympy.polys.polyclasses:DMF' has slots but superclass does not.
       - 'sympy.core.sympify:CantSympify' superclass has no slots.
ERROR: 'sympy.polys.polyclasses:DMP' has slots but superclass does not.
       - 'sympy.core.sympify:CantSympify' superclass has no slots.
```

There are two common issues:
- a slot being redefined in a subclass. This wastes some space as the slot descriptor will be overridden.
- a slotted class which inherits from a non-slot class. This leaves `__dict__` on the class accessible and negates much of the memory savings.

I will submit a PR soon with the fixes

ariebovenberg:

fixes #22922 

- `__slots__` added to some base classes
- specific slot names removed that were already declared on base classes

#### Other comments

I discovered the slots issues with [slotscheck](https://github.com/ariebovenberg/slotscheck), a tool I maintain. Of course, If you like, I can add it to tox/CI as I've done for [instagram/LibCST](https://github.com/Instagram/LibCST/pull/615), [dry-python/returns](https://github.com/dry-python/returns/pull/1233), and [aio-libs/aiohttp](https://github.com/aio-libs/aiohttp/pull/6547).

#### Release Notes

* other
  * Fixed several issues with `__slots__` (overlaps and omission in base classes)

### DIFF
--- a/sympy/codegen/ast.py
+++ b/sympy/codegen/ast.py
@@ -126,7 +126,7 @@ There is a function constructing a loop (or a complete function) like this in
 
 """
 
-from typing import Any, Dict as tDict, List
+from typing import Any, Dict as tDict, List, Tuple as tTuple
 
 from collections import defaultdict
 
@@ -161,7 +161,7 @@ def _mk_Tuple(args):
 
 
 class CodegenAST(Basic):
-    pass
+    __slots__ = ()
 
 
 class Token(CodegenAST):
@@ -170,10 +170,10 @@ class Token(CodegenAST):
     Explanation
     ===========
 
-    Defining fields are set in ``__slots__``. Attributes (defined in __slots__)
+    Defining fields are set in ``_fields``. Attributes (defined in _fields)
     are only allowed to contain instances of Basic (unless atomic, see
     ``String``). The arguments to ``__new__()`` correspond to the attributes in
-    the order defined in ``__slots__`. The ``defaults`` class attribute is a
+    the order defined in ``_fields`. The ``defaults`` class attribute is a
     dictionary mapping attribute names to their default values.
 
     Subclasses should not need to override the ``__new__()`` method. They may
@@ -182,14 +182,14 @@ class Token(CodegenAST):
     in the class attribute ``not_in_args`` are not passed to :class:`~.Basic`.
     """
 
-    __slots__ = ()
+    __slots__ = _fields = ()  # type: tTuple[str, ...]
     defaults = {}  # type: tDict[str, Any]
     not_in_args = []  # type: List[str]
     indented_args = ['body']
 
     @property
     def is_Atom(self):
-        return len(self.__slots__) == 0
+        return len(self._fields) == 0
 
     @classmethod
     def _get_constructor(cls, attr):
@@ -213,20 +213,20 @@ class Token(CodegenAST):
         if len(args) == 1 and not kwargs and isinstance(args[0], cls):
             return args[0]
 
-        if len(args) > len(cls.__slots__):
-            raise ValueError("Too many arguments (%d), expected at most %d" % (len(args), len(cls.__slots__)))
+        if len(args) > len(cls._fields):
+            raise ValueError("Too many arguments (%d), expected at most %d" % (len(args), len(cls._fields)))
 
         attrvals = []
 
         # Process positional arguments
-        for attrname, argval in zip(cls.__slots__, args):
+        for attrname, argval in zip(cls._fields, args):
             if attrname in kwargs:
                 raise TypeError('Got multiple values for attribute %r' % attrname)
 
             attrvals.append(cls._construct(attrname, argval))
 
         # Process keyword arguments
-        for attrname in cls.__slots__[len(args):]:
+        for attrname in cls._fields[len(args):]:
             if attrname in kwargs:
                 argval = kwargs.pop(attrname)
 
@@ -243,13 +243,13 @@ class Token(CodegenAST):
 
         # Parent constructor
         basic_args = [
-            val for attr, val in zip(cls.__slots__, attrvals)
+            val for attr, val in zip(cls._fields, attrvals)
             if attr not in cls.not_in_args
         ]
         obj = CodegenAST.__new__(cls, *basic_args)
 
         # Set attributes
-        for attr, arg in zip(cls.__slots__, attrvals):
+        for attr, arg in zip(cls._fields, attrvals):
             setattr(obj, attr, arg)
 
         return obj
@@ -257,13 +257,13 @@ class Token(CodegenAST):
     def __eq__(self, other):
         if not isinstance(other, self.__class__):
             return False
-        for attr in self.__slots__:
+        for attr in self._fields:
             if getattr(self, attr) != getattr(other, attr):
                 return False
         return True
 
     def _hashable_content(self):
-        return tuple([getattr(self, attr) for attr in self.__slots__])
+        return tuple([getattr(self, attr) for attr in self._fields])
 
     def __hash__(self):
         return super().__hash__()
@@ -291,12 +291,12 @@ class Token(CodegenAST):
     def _sympyrepr(self, printer, *args, joiner=', ', **kwargs):
         from sympy.printing.printer import printer_context
         exclude = kwargs.get('exclude', ())
-        values = [getattr(self, k) for k in self.__slots__]
+        values = [getattr(self, k) for k in self._fields]
         indent_level = printer._context.get('indent_level', 0)
 
         arg_reprs = []
 
-        for i, (attr, value) in enumerate(zip(self.__slots__, values)):
+        for i, (attr, value) in enumerate(zip(self._fields, values)):
             if attr in exclude:
                 continue
 
@@ -329,7 +329,7 @@ class Token(CodegenAST):
         apply : callable, optional
             Function to apply to all values.
         """
-        kwargs = {k: getattr(self, k) for k in self.__slots__ if k not in exclude}
+        kwargs = {k: getattr(self, k) for k in self._fields if k not in exclude}
         if apply is not None:
             return {k: apply(v) for k, v in kwargs.items()}
         else:
@@ -849,7 +849,7 @@ class For(Token):
         ))
     ))
     """
-    __slots__ = ('target', 'iterable', 'body')
+    __slots__ = _fields = ('target', 'iterable', 'body')
     _construct_target = staticmethod(_sympify)
 
     @classmethod
@@ -893,7 +893,7 @@ class String(Atom, Token):
     String('foo')
 
     """
-    __slots__ = ('text',)
+    __slots__ = _fields = ('text',)
     not_in_args = ['text']
     is_Atom = True
 
@@ -945,7 +945,7 @@ class Node(Token):
 
     """
 
-    __slots__ = ('attrs',)
+    __slots__ = _fields = ('attrs',)  # type: tTuple[str, ...]
 
     defaults = {'attrs': Tuple()}  # type: tDict[str, Any]
 
@@ -1013,7 +1013,7 @@ class Type(Token):
     .. [1] https://docs.scipy.org/doc/numpy/user/basics.types.html
 
     """
-    __slots__ = ('name',)
+    __slots__ = _fields = ('name',)  # type: tTuple[str, ...]
 
     _construct_name = String
 
@@ -1142,12 +1142,13 @@ class Type(Token):
 
 class IntBaseType(Type):
     """ Integer base type, contains no size information. """
-    __slots__ = ('name',)
+    __slots__ = ()
     cast_nocheck = lambda self, i: Integer(int(i))
 
 
 class _SizedIntType(IntBaseType):
-    __slots__ = ('name', 'nbits',)
+    __slots__ = ('nbits',)
+    _fields = Type._fields + __slots__
 
     _construct_nbits = Integer
 
@@ -1160,6 +1161,7 @@ class _SizedIntType(IntBaseType):
 
 class SignedIntType(_SizedIntType):
     """ Represents a signed integer type. """
+    __slots__ = ()
     @property
     def min(self):
         return -2**(self.nbits-1)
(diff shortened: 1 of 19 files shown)


oscarbenjamin:
Looks like the `__slots__` are being used for something in codegen. Not sure if that's easy to fix.

ariebovenberg:
@oscarbenjamin ooo that'd be easy for me to overlook 👀 . I'll see what breaks then and let you know 😅

ThePauliPrinciple:
> Looks like the `__slots__` are being used for something in codegen. Not sure if that's easy to fix.

The role of `__slots__` in codegen/ast.py is documented in the docstring, maybe that is helpful to fix the issue.

sympy/geometry/entity.py
```diff
@@ -69,6 +69,8 @@ class GeometryEntity(Basic, EvalfMixin):
 
     """
 
+    __slots__ = ()
```
oscarbenjamin, line 72: The mypy tests are failing because mypy can't figure out the "type" of slots here. I think it needs something like this to make mypy happy:
```diff
diff --git a/sympy/geometry/entity.py b/sympy/geometry/entity.py
index 1bd1722..5d541c1 100644
--- a/sympy/geometry/entity.py
+++ b/sympy/geometry/entity.py
@@ -19,6 +19,7 @@
 R3 are currently the only ambient spaces implemented.
 
 """
+from typing import Tuple as tTuple
 
 from sympy.core.basic import Basic
 from sympy.core.containers import Tuple
@@ -69,7 +70,7 @@ class GeometryEntity(Basic, EvalfMixin):
 
     """
 
-    __slots__ = ()
+    __slots__: tTuple[()] = ()
 
     def __cmp__(self, other):
         """Comparison of two GeometryEntities."""
```

ariebovenberg:
@oscarbenjamin @ThePauliPrinciple thanks for the feedback. The mypy/authors issues are easy to fix, so only `codegen` remains.

I'd like to suggest using a separate classvar (like `__codegen_fields__` or something) instead of `__slots__` to indicate the fields for codegen. Using `__slots__` for codegen will cause overlaps. The Python docs have this to say about overlaps (emphasis mine):

> If a class defines a slot also defined in a base class, the instance variable defined by the base class slot is inaccessible (except by retrieving its descriptor directly from the base class). *This renders the meaning of the program undefined*. In the future, *a check may be added to prevent this*.

This tells us:
1. It may cause undefined behavior
2. It may be disallowed in future Python versions

Agreed to move "codegen" slots to another attribute, so `__slots__` overlaps can be removed?

sympy/geometry/entity.py
ariebovenberg: That'd make `mypy` happy, although I'd probably go for `tTuple[str, ...]` to ensure slots may be added later.

ThePauliPrinciple:
@bjodah do you think there will be unintended side effects with this change?

I don't think we should use a "custom" double underscore attribute for this, something like `_attributes` or `_fields` makes more sense to me.

We could also remove overlapping values from `__slots__` and then check the MRO for any `CodegenAST` superclass and collect all `__slots__` from those.

ariebovenberg:
@ThePauliPrinciple agree that a single `_` name would be best. First, I'll investigate your suggestion on generating codegen slots from MRO.

ariebovenberg:
I've adjusted the commits according to the development guide. What I realize now is that the branch name should reference the issue number. I can't make this change without opening a new pull request. What do you prefer?

sympy/codegen/ast.py
```diff
@@ -161,7 +161,17 @@ def _mk_Tuple(args):
 
 
 class CodegenAST(Basic):
-    pass
+    __slots__ = ()
+
+
+def _fields(cls):
+    if hasattr(cls, "_fields"):
+        yield from cls._fields
```
ThePauliPrinciple, line 169: Why do we need this?

oscarbenjamin:
> the branch name should reference the issue number

It's fine to use a different branch name. Using the issue number can be helpful so you remember what the branch is but ultimately it's just a branch in your fork rather than in the sympy repo so you can call it what you want.

sympy/codegen/ast.py
ariebovenberg: Hmmm, I did indeed forget to explain.

The reason this is needed (and we cannot only use `__slots__`) is that there is no way to 'automatically' order inherited slots in the correct way.

One sensible way seemed from in reverse MRO:

```python
class A:
    __slots__ = ('x', 'y')

class B:
    __slots__ = ('z', )

_fields(B)  # x, y, z
```

This seems to be the rule followed by most `codegen.ast` members

However, some classes (like `FunctionPrototype`) add inherited slots in a different way:

```python
class Node:
    __slots__ = ("attrs", )

class FunctionPrototype(Node):
    __slots__ = ('return_type', 'name', 'parameters') + Node.__slots__

_fields(FunctionPrototype)  # attrs, return_type, name, parameters
# SHOULD BE return_type, name, parameters, attrs
```

I briefly considered just adopting the new order, but that would of course be a breaking change.

ariebovenberg: If you prefer, I could also make `_fields` the only way to define fields (i.e. never fallback to `__slots__`). That'd simplify introspection but it'd require setting `_fields` on a lot of classes.

ThePauliPrinciple: If going by the mro is causing ordering issues then I think it is indeed better to simply define `_fields` fully. It would also make it clearer what to do for a future author who wants to create another codegen ast class.

asmeurer:
> > the branch name should reference the issue number
> 
> It's fine to use a different branch name. Using the issue number can be helpful so you remember what the branch is but ultimately it's just a branch in your fork rather than in the sympy repo so you can call it what you want.

Does it still say that in the contributing guide? Can we remove this? Typing out branch numbers is super annoying whenever I checkout someone's branch. I'd much prefer short to the point branch names.

ariebovenberg:
> > > the branch name should reference the issue number
> > 
> > 
> > It's fine to use a different branch name. Using the issue number can be helpful so you remember what the branch is but ultimately it's just a branch in your fork rather than in the sympy repo so you can call it what you want.
> 
> Does it still say that in the contributing guide? Can we remove this? Typing out branch numbers is super annoying whenever I checkout someone's branch. I'd much prefer short to the point branch names.

It's documented here, if you want to remove it: https://github.com/sympy/sympy/wiki/Development-workflow#branch-names

bjodah:
> @bjodah do you think there will be unintended side effects with this change?

No, I think it's fine as long as tests pass. I was not aware of the finer points of overriding `__slots__` in subclasses, nice to see there's a tool to teach us to do better :+1:.

ThePauliPrinciple:
I'd be in favour of merging this PR and have a seperate issue/PR for discussing the CI.

oscarbenjamin:
It's not really clear to me how useful `__slots__` actually is. I know it used to be an attempt to save memory but other ways of saving memory in CPython have been developed since then. As I understand it using `__slots__` is not good for performance in pypy.

Is there a strong advantage in using `__slots__` at all?

asmeurer:
I'm OK with adding this to the CI. 

> It's documented here, if you want to remove it: sympy/sympy/wiki/Development-workflow#branch-names

I removed it and replaced the example branch name with something better.

asmeurer:
> Is there a strong advantage in using __slots__ at all?

I think `__slots__` can also positively impact performance. It might be worth benchmarking, both the speed and memory impacts of having vs. not having `__slots__`.

As far as downsides, I'm not aware of anyone ever complaining about it. I've seen it cause weird issues with metaclasses, but I think that usually means you are doing something wrong anyway. It also makes SymPy objects unable to have weak references (without `__weakref__` in `__slots__`). And obviously it makes it impossible to assign random attributes to SymPy objects, but that's usually not something users should do anyway.

ariebovenberg:
> I think __slots__ can also positively impact performance. It might be worth benchmarking, both the speed and memory impacts of having vs. not having __slots__.

Indeed, the two advantages are:
1. better performance of attribute access
2. reduced memory footprint

There are some benchmarks floating around that should be easy to find. Of course, it's best to measure your specific case.

I'm not sure of the impact on other Python implementations like `pypy` though 🤔

oscargus:
I like the idea from @bjodah , merge this and discuss `slotcheck` in another issue.

But for what it is worth: independent of if slots are a good idea or not, I do not think that we would like to misuse them, so if that can be prevented (and it is not overly complicated or time-consuming to run it), then it is a yes from me.

ariebovenberg:
@oscarbenjamin @ThePauliPrinciple @oscargus @asmeurer I've created a ticket to discuss `slotscheck` here: #23030.

oscargus:
Any objections against merging this? Seems like an improvement to me!

ariebovenberg:
@oscargus @oscarbenjamin @ThePauliPrinciple can this be merged?

oscarbenjamin:
This seems to be the cause of CI failures on the master branch.

The last commit is 30th Jan. After that length of time the PR should be closed and reopened to rerun CI before merging.

asmeurer:
Sorry about that. I didn't realize the most recent commit was from so long ago. We can do a quick revert of this unless @ariebovenberg can provide a quick fix.

oscarbenjamin:
Presumably some other PR has changed something affected by this. My guess is https://github.com/sympy/sympy/pull/23196

ariebovenberg:
@oscarbenjamin if that's the only failure, the fix is easy. I'll submit a new PR.

ariebovenberg:
@asmeurer @oscarbenjamin the issue has been fixed in #23252

oscarbenjamin:
Great, thanks.

PR #22936, opened 2022-01-27
Add public documentation for array expressions

Upabjojr:
It's time to publicly announce the array expressions module. Added link to the documentation, will become available to everyone after the next version release.

* tensor
  * Array expression submodule has now reached a stable state and has been added to the documentation.

### DIFF
--- a/sympy/tensor/array/expressions/__init__.py
+++ b/sympy/tensor/array/expressions/__init__.py
@@ -1,0 +1,108 @@
+r"""
+Array expressions are expressions representing N-dimensional arrays, without evaluating them.
+They are *lazy* as they don't evaluate the operations on the arrays until required and
+can define operations on unknown array (in a certain way reminiscent of graphs of
+operations in TensorFlow).
+
+Every N-dimensional array operator has a corresponding lazy object.
+
+Table of correspondences:
+
+=============================== =============================
+         Array operator           Array expression operator
+=============================== =============================
+        tensorproduct                 ArrayTensorProduct
+        tensorcontraction             ArrayContraction
+        tensordiagonal                ArrayDiagonal
+        permutedims                   PermuteDims
+        derive_by_array               array_derive
+=============================== =============================
+
+Examples
+========
+
+>>> from sympy import MatrixSymbol
+>>> from sympy.tensor.array.expressions import ArrayTensorProduct, ArrayContraction
+>>> M = MatrixSymbol("M", 3, 3)
+>>> N = MatrixSymbol("N", 3, 3)
+
+Express the matrix product in the array expression form:
+
+>>> from sympy.tensor.array.expressions import convert_matrix_to_array
+>>> expr = convert_matrix_to_array(M*N)
+>>> expr
+ArrayContraction(ArrayTensorProduct(M, N), (1, 2))
+
+The expression can be converted back to matrix form:
+
+>>> from sympy.tensor.array.expressions import convert_array_to_matrix
+>>> convert_array_to_matrix(expr)
+M*N
+
+Add a second contraction on the remaining axes in order to get the trace of `M \cdot N`:
+
+>>> expr_tr = ArrayContraction(expr, (0, 1))
+>>> expr_tr
+ArrayContraction(ArrayContraction(ArrayTensorProduct(M, N), (1, 2)), (0, 1))
+
+Flatten the expression by calling ``.doit()`` and remove the nested array contraction operations:
+
+>>> expr_tr.doit()
+ArrayContraction(ArrayTensorProduct(M, N), (0, 3), (1, 2))
+
+Get the explicit form of the array expression:
+
+>>> expr.as_explicit()
+[[M[0, 0]*N[0, 0] + M[0, 1]*N[1, 0] + M[0, 2]*N[2, 0], M[0, 0]*N[0, 1] + M[0, 1]*N[1, 1] + M[0, 2]*N[2, 1], M[0, 0]*N[0, 2] + M[0, 1]*N[1, 2] + M[0, 2]*N[2, 2]], [M[1, 0]*N[0, 0] + M[1, 1]*N[1, 0] + M[1, 2]*N[2, 0], M[1, 0]*N[0, 1] + M[1, 1]*N[1, 1] + M[1, 2]*N[2, 1], M[1, 0]*N[0, 2] + M[1, 1]*N[1, 2] + M[1, 2]*N[2, 2]], [M[2, 0]*N[0, 0] + M[2, 1]*N[1, 0] + M[2, 2]*N[2, 0], M[2, 0]*N[0, 1] + M[2, 1]*N[1, 1] + M[2, 2]*N[2, 1], M[2, 0]*N[0, 2] + M[2, 1]*N[1, 2] + M[2, 2]*N[2, 2]]]
+
+Express the trace of a matrix:
+
+>>> from sympy import Trace
+>>> convert_matrix_to_array(Trace(M))
+ArrayContraction(M, (0, 1))
+>>> convert_matrix_to_array(Trace(M*N))
+ArrayContraction(ArrayTensorProduct(M, N), (0, 3), (1, 2))
+
+Express the transposition of a matrix (will be expressed as a permutation of the axes:
+
+>>> convert_matrix_to_array(M.T)
+PermuteDims(M, (0 1))
+
+Compute the derivative array expressions:
+
+>>> from sympy.tensor.array.expressions import array_derive
+>>> d = array_derive(M, M)
+>>> d
+PermuteDims(ArrayTensorProduct(I, I), (3)(1 2))
+
+Verify that the derivative corresponds to the form computed with explicit matrices:
+
+>>> d.as_explicit()
+[[[[1, 0, 0], [0, 0, 0], [0, 0, 0]], [[0, 1, 0], [0, 0, 0], [0, 0, 0]], [[0, 0, 1], [0, 0, 0], [0, 0, 0]]], [[[0, 0, 0], [1, 0, 0], [0, 0, 0]], [[0, 0, 0], [0, 1, 0], [0, 0, 0]], [[0, 0, 0], [0, 0, 1], [0, 0, 0]]], [[[0, 0, 0], [0, 0, 0], [1, 0, 0]], [[0, 0, 0], [0, 0, 0], [0, 1, 0]], [[0, 0, 0], [0, 0, 0], [0, 0, 1]]]]
+>>> Me = M.as_explicit()
+>>> Me.diff(Me)
+[[[[1, 0, 0], [0, 0, 0], [0, 0, 0]], [[0, 1, 0], [0, 0, 0], [0, 0, 0]], [[0, 0, 1], [0, 0, 0], [0, 0, 0]]], [[[0, 0, 0], [1, 0, 0], [0, 0, 0]], [[0, 0, 0], [0, 1, 0], [0, 0, 0]], [[0, 0, 0], [0, 0, 1], [0, 0, 0]]], [[[0, 0, 0], [0, 0, 0], [1, 0, 0]], [[0, 0, 0], [0, 0, 0], [0, 1, 0]], [[0, 0, 0], [0, 0, 0], [0, 0, 1]]]]
+
+"""
+
+__all__ = [
+    "ArraySymbol", "ArrayElement", "ZeroArray", "OneArray",
+    "ArrayTensorProduct",
+    "ArrayContraction",
+    "ArrayDiagonal",
+    "PermuteDims",
+    "ArrayAdd",
+    "ArrayElementwiseApplyFunc",
+    "Reshape",
+    "convert_array_to_matrix",
+    "convert_matrix_to_array",
+    "convert_indexed_to_array",
+    "array_derive",
+]
+
+from sympy.tensor.array.expressions.array_expressions import ArrayTensorProduct, ArrayAdd, PermuteDims, ArrayDiagonal, \
+    ArrayContraction, Reshape, ArraySymbol, ArrayElement, ZeroArray, OneArray, ArrayElementwiseApplyFunc
+from sympy.tensor.array.expressions.arrayexpr_derivatives import array_derive
+from sympy.tensor.array.expressions.conv_array_to_matrix import convert_array_to_matrix
+from sympy.tensor.array.expressions.conv_indexed_to_array import convert_indexed_to_array
+from sympy.tensor.array.expressions.conv_matrix_to_array import convert_matrix_to_array
--- a/sympy/tensor/array/expressions/__init__.py
+++ b/sympy/tensor/array/expressions/__init__.py
@@ -1,10 +1,9 @@
 r"""
-Array expressions are expressions representing N-dimensional arrays, without evaluating them.
-They are *lazy* as they don't evaluate the operations on the arrays until required and
-can define operations on unknown array (in a certain way reminiscent of graphs of
-operations in TensorFlow).
+Array expressions are expressions representing N-dimensional arrays, without
+evaluating them. These expressions represent in a certain way abstract syntax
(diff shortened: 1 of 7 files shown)


sympy/tensor/array/expressions/__init__.py
```diff
@@ -0,0 +1,108 @@
+r"""
+Array expressions are expressions representing N-dimensional arrays, without evaluating them.
+They are *lazy* as they don't evaluate the operations on the arrays until required and
+can define operations on unknown array (in a certain way reminiscent of graphs of
+operations in TensorFlow).
```
ThePauliPrinciple, line 5: While internally TensorFlow still has these graphs, since the TF2 release, the API exposes it as "eager" tensors, meaning they appear to not be evaluated lazily (eventhough they really are). So this comparison might be confusing to a TF user who is not familiar with TF1 or not very knowledgable about how TF2 works internally (really, TF1 and TF2 should be considered two different libraries in my opinion).

sympy/tensor/array/expressions/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+        permutedims                   PermuteDims
+        derive_by_array               array_derive
+=============================== =============================
+
+Examples
+========
+
+>>> from sympy import MatrixSymbol
+>>> from sympy.tensor.array.expressions import ArrayTensorProduct, ArrayContraction
+>>> M = MatrixSymbol("M", 3, 3)
```
ThePauliPrinciple, line 26: I would prefer examples for `Array` and `ArraySymbol`, since this is the documentation about arrays. Then competability with matrices can be introduced after it is clear how `Array`s work.

sympy/tensor/array/expressions/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+Table of correspondences:
+
+=============================== =============================
+         Array operator           Array expression operator
+=============================== =============================
+        tensorproduct                 ArrayTensorProduct
+        tensorcontraction             ArrayContraction
+        tensordiagonal                ArrayDiagonal
+        permutedims                   PermuteDims
+        derive_by_array               array_derive
```
ThePauliPrinciple, line 18: It is confusing to me to call `array_derive` a "corresponding lazy object" here and as far as I can see, it isn't lazy. e.g. when dispatched an `ArraySymbol` it will yield zero's or an identity with some shape depending on whether we derive by the same object or not. Lazy would imply to me that we could still substitute either of the objects later on. (To indicate that they are equal after all) e.g.
```python
from sympy import Derivative
from sympy.abc import x,y
Derivative(x,y)
Derivative(x,y).subs(y,x)
```

ThePauliPrinciple:
First of all, congratulations on getting the array module to this point!

Some (minor) points.

I feel this needs a release note so that users are aware that arrays now provide a stable API.

From what I understand, currently there is no "broadcasting" support. I suspect people expect this feature since they are likely to be familiar with NumPy arrays. Is this something we wan't in SymPy. I feel it would be convenient to express certain mathematical concepts, e.g. when we have an array containing the Cartesian coordinates `X(N,3)` of objects, we can express the "matrix" of all object to object vectors by writing `D = X[None,:,:] - X[:,None,:]` (element wise squaring, contracting over the last index and taking the sqrt elementwise would give a matrix of distances). While I'm not requesting this particular interface, it would be useful if it is possible to express such an operation and I'm not aware of a way to do so currently.

ThePauliPrinciple:
Currently `Reshape` doesn't appear to have a docstring.

sympy/tensor/array/expressions/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+
+Flatten the expression by calling ``.doit()`` and remove the nested array contraction operations:
+
+>>> expr_tr.doit()
+ArrayContraction(ArrayTensorProduct(M, N), (0, 3), (1, 2))
+
+Get the explicit form of the array expression:
+
+>>> expr.as_explicit()
+[[M[0, 0]*N[0, 0] + M[0, 1]*N[1, 0] + M[0, 2]*N[2, 0], M[0, 0]*N[0, 1] + M[0, 1]*N[1, 1] + M[0, 2]*N[2, 1], M[0, 0]*N[0, 2] + M[0, 1]*N[1, 2] + M[0, 2]*N[2, 2]], [M[1, 0]*N[0, 0] + M[1, 1]*N[1, 0] + M[1, 2]*N[2, 0], M[1, 0]*N[0, 1] + M[1, 1]*N[1, 1] + M[1, 2]*N[2, 1], M[1, 0]*N[0, 2] + M[1, 1]*N[1, 2] + M[1, 2]*N[2, 2]], [M[2, 0]*N[0, 0] + M[2, 1]*N[1, 0] + M[2, 2]*N[2, 0], M[2, 0]*N[0, 1] + M[2, 1]*N[1, 1] + M[2, 2]*N[2, 1], M[2, 0]*N[0, 2] + M[2, 1]*N[1, 2] + M[2, 2]*N[2, 2]]]
```
asmeurer, line 56: You should be able to line wrap these. The doctester ignores newlines in the output. 

Although is the type of this object an Array or a list? It seems to me it ought to be an Array, in which case

- The str form should include "Array(...)" so that that is clear.
- The printer should do some basic line wrapping, similar to what is done with Matrix

asmeurer:
> From what I understand, currently there is no "broadcasting" support. I suspect people expect this feature since they are likely to be familiar with NumPy arrays. Is this something we wan't in SymPy. I feel it would be convenient to express certain mathematical concepts, e.g. when we have an array containing the Cartesian coordinates X(N,3) of objects, we can express the "matrix" of all object to object vectors by writing D = X[None,:,:] - X[:,None,:] (element wise squaring, contracting over the last index and taking the sqrt elementwise would give a matrix of distances). While I'm not requesting this particular interface, it would be useful if it is possible to express such an operation and I'm not aware of a way to do so currently.

If we do, I would suggest adding it as an explicit operation, like `Broadcasted(array, shape)` (or some generalization of broadcasting for tensors), at least at the low level. The high level API would most likely benefit from being able to write NumPy-like operations that broadcast automatically. But at the low level, we will want to be able to represent array operations without implicit broadcasting so that we can more easily target languages like C that don't have broadcasting in code generation.

sympy/tensor/array/expressions/__init__.py
Upabjojr: Well, I have to remove the word *lazy* from the documentation, it's the wrong adjective. Better call it *array expression operator* or *component-implicit* (whereas *component-explicit* are traditional arrays with elements).

ThePauliPrinciple:
> least at the low level. The high level API would most likely benefit from being able to write NumPy-like operations that broadcast automatically. But at the low level, we will want to be able to represent array operations without implicit broadcasting so that we can more easily target languages like C that don't have broadcasting in code generation.

I agree.

For codegeneration, I think the c-like languages should convert array expressions to indexed and for numpy/tf like languages, it should convert indexed expressions to array expressions.

Something like `D[i,j,n] = X[i,n] - X[j,n]` can easily be written in c and SymPy Indexed objects, but requires broadcasting for languages that use arrays.

<details> <summary>What I had in mind:</summary>

```python
class ArrayBroadcast(_CodegenArrayAbstract):
    """Symbolic broadcasting for Arrays."""
    def __new__(cls, expr, broadcast_to):
        expr = _sympify(expr)
        broadcast_to = _sympify(broadcast_to)

        if hasattr(broadcast_to, 'shape'):
            broadcast_to = Tuple(*broadcast_to.shape)
        broadcast_shape = cls._broadcast(expr.shape, broadcast_to)

        obj = Expr.__new__(cls, expr, broadcast_shape)
        obj._shape = tuple(broadcast_shape)
        obj._expr = expr

        return obj

    @classmethod
    def _broadcast(cls, a, b):
        la, lb = len(a), len(b)
        l = min(la,lb)
        left = a[:-l] + b[:-l]

        def _index(ia, ib):
            if ia == ib:
                return ia
            elif ia == 1
                return ib
            elif ib == 1:
                return ia
            else:
                raise ValueError("Can not broadcast %s and %s" % ia, ib)

        right = tuple(map(_index,a[-l:], b[-l:]))
        return Tuple(*left, *right)

    def as_explicit(self):
        ee = self.expr.as_explicit()
        if not all(i.is_Integer for i in self.shape + self.expr.shape):
            raise ValueError("cannot express explicit array with symbolic shape")

        def _index_map(index):
            return [0 if b == 1 else a for a, b in zip(index, self.expr.shape)]

        data = [ee[_index_map(i)] for i in itertools.product(*[range(j) for j in self.shape])]
        return ImmutableDenseNDimArray(data).reshape(*self.shape)

    @property
    def shape(self):
        return self._shape

    @property
    def expr(self):
        return self._expr
```
```python
>>> from sympy.tensor.array.expressions.array_expressions import ArraySymbol, ArrayAdd, Broadcast, Reshape; A= Reshape(ArraySymbol('X',(5,3)),(1,5,3));  B = Reshape(ArraySymbol('X',(5,3)),(5,1,3)); ArrayAdd(Broadcast(A,B),Broadcast(B,A)).as_explicit().shape
(5, 5, 3)
>>> ArrayAdd(Broadcast(A,B),Broadcast(B,A))
ArrayAdd(Broadcast(Reshape(X, (1, 5, 3)), (5, 5, 3)), Broadcast(Reshape(X, (5, 1, 3)), (5, 5, 3)))
>>> ArrayAdd(Broadcast(A,B),Broadcast(B,A)).as_explicit()
[[[2*X[0, 0], 2*X[0, 1], 2*X[0, 2]], [X[0, 0] + X[1, 0], X[0, 1] + X[1, 1], X[0, 2] + X[1, 2]], [X[0, 0] + X[2, 0], X[0, 1] + X[2, 1], X[0, 2] + X[2, 2]], [X[0, 0] + X[3, 0], X[0, 1] + X[3, 1], X[0, 2] + X[3, 2]], [X[0, 0] + X[4, 0], X[0, 1] + X[4, 1], X[0, 2] + X[4, 2]]], [[X[0, 0] + X[1, 0], X[0, 1] + X[1, 1], X[0, 2] + X[1, 2]], [2*X[1, 0], 2*X[1, 1], 2*X[1, 2]], [X[1, 0] + X[2, 0], X[1, 1] + X[2, 1], X[1, 2] + X[2, 2]], [X[1, 0] + X[3, 0], X[1, 1] + X[3, 1], X[1, 2] + X[3, 2]], [X[1, 0] + X[4, 0], X[1, 1] + X[4, 1], X[1, 2] + X[4, 2]]], [[X[0, 0] + X[2, 0], X[0, 1] + X[2, 1], X[0, 2] + X[2, 2]], [X[1, 0] + X[2, 0], X[1, 1] + X[2, 1], X[1, 2] + X[2, 2]], [2*X[2, 0], 2*X[2, 1], 2*X[2, 2]], [X[2, 0] + X[3, 0], X[2, 1] + X[3, 1], X[2, 2] + X[3, 2]], [X[2, 0] + X[4, 0], X[2, 1] + X[4, 1], X[2, 2] + X[4, 2]]], [[X[0, 0] + X[3, 0], X[0, 1] + X[3, 1], X[0, 2] + X[3, 2]], [X[1, 0] + X[3, 0], X[1, 1] + X[3, 1], X[1, 2] + X[3, 2]], [X[2, 0] + X[3, 0], X[2, 1] + X[3, 1], X[2, 2] + X[3, 2]], [2*X[3, 0], 2*X[3, 1], 2*X[3, 2]], [X[3, 0] + X[4, 0], X[3, 1] + X[4, 1], X[3, 2] + X[4, 2]]], [[X[0, 0] + X[4, 0], X[0, 1] + X[4, 1], X[0, 2] + X[4, 2]], [X[1, 0] + X[4, 0], X[1, 1] + X[4, 1], X[1, 2] + X[4, 2]], [X[2, 0] + X[4, 0], X[2, 1] + X[4, 1], X[2, 2] + X[4, 2]], [X[3, 0] + X[4, 0], X[3, 1] + X[4, 1], X[3, 2] + X[4, 2]], [2*X[4, 0], 2*X[4, 1], 2*X[4, 2]]]]
```
</details>

Whether we then want `ArrayAdd` to attempt to create such an `ArrayBroadcast` when it's arguments don't match shape is a different discussion, but at least it would be nice to be able to represent such an operation.

Upabjojr:
> From what I understand, currently there is no "broadcasting" support. I suspect people expect this feature since they are likely to be familiar with NumPy arrays.

Well... there is no real support for arithmetic operations in the array expressions module. There this `ArrayAdd` for elementwise additions, but no equivalent for multiplications.

In theory multiplications can be represented by `ArrayDiagonal(ArrayTensorProduct(X, Y), ... proper axes diagonalized ... )`, that's why there is no implementation of elementwise multiplication.

I have created the array expression module in order to represent all possible results of matrix derivatives. I didn't have any intention of imitating NumPy.

I don't think we should support *broadcasting* the way NumPy does it (i.e. by doing it implicitly). A SymPy user should check and match the dimensions properly. 99% of SymPy users probably also have NumPy installed. There is a `Reshape` operator now, maybe we can also add an equivalent of tile/repeat, or something like [numpy.broadcast_to](https://numpy.org/doc/stable/reference/generated/numpy.broadcast_to.html#numpy.broadcast_to).

Upabjojr:
> Something like `D[i,j,n] = X[i,n] - X[j,n]` can easily be written in c and SymPy Indexed objects, but requires broadcasting for languages that use arrays.

@ThePauliPrinciple No, we can already represent `X[i,n] - X[j,n]` as:
```python
X = ArraySymbol("X", (3, 3))
O = OneArray(3)

ArrayAdd(
    PermuteDims(ArrayTensorProduct(O, X), [1, 0, 2]),
    ArrayTensorProduct(-1, O, X)
)
```

The last line is equivalent to `D[i, j, n]`, you can verify it by calling `.as_explicit()` and computing the other expression like:
```python
D = MutableDenseNDimArray.zeros(3, 3, 3)
for i, j, n in itertools.product(*[range(3) for ji in range(3)]):
    D[i, j, n] = X[i, n] - X[j, n]
```

Maybe we don't need *broadcasting* after all...

sympy/tensor/array/expressions/__init__.py
Upabjojr: https://github.com/sympy/sympy/issues/22939

ThePauliPrinciple:
> Maybe we don't need _broadcasting_ after all...

I'm absolutely open to other alternatives that does not add more code to SymPy, but how do I make it work if I have 5 "particles", i.e. `X = ArraySymbol("X", (5, 3))`

Upabjojr:
> I'm absolutely open to other alternatives that does not add more code to SymPy, but how do I make it work if I have 5 "particles", i.e. `X = ArraySymbol("X", (5, 3))`

Do you mean `X[i, n] - X[j, n]` with shape `(5, 5, 3)` ?

```python
ArrayAdd(ArrayTensorProduct(-1, OneArray(5), X), PermuteDims(ArrayTensorProduct(X, OneArray(5)), (1 2)))
```

ThePauliPrinciple:
So, in general, we can say that `ArrayTensorProduct(OneArray(N), X)` is a way to create a broadcasted index on the left (right by swapping `O` and `X`) and we can then just put it in the right index with `PermuteDims`, that makes sense to me. It's a bit inconvenient to parse it when you want to use this structure for printing to NumPy/TensorFlow, but not exactly impossible either.

ThePauliPrinciple:
In particular, you want to avoid evaluating tensorproducts if you can:
```python
>>> import numpy as np
>>> import timeit
>>> X = np.ones((1000,3))
>>> O = np.ones(1000)
>>> def test():
>>>    T=np.tensordot(O,X,axes=0)
>>>    T-T.swapaxes(0,1)
>>> print(timeit.timeit(test, number = 1000))
>>> def test():
>>>    X[None,:,:]-X[:,None,:]
>>> print(timeit.timeit(test , number = 1000))
89.66080202800003
32.12669453699891
```
Note that `np.tensordot(..., axes=0)` is the same as a tensorproduct (no contractions).

PR #22937, opened 2022-01-27
Removed the recommendation to not use solve.

moorepants:

The message pointing people away from solve is confusing to users, because solve() is perfectly fine to use. It confused my students today, so I thought I'd change the message to something that avoids recommending the newer functions (which I find the outputs of much more difficult to deal with than the myriad of options for solve).

#### Release Notes

NO ENTRY

### DIFF
--- a/doc/src/modules/solvers/solveset.rst
+++ b/doc/src/modules/solvers/solveset.rst
@@ -12,8 +12,8 @@ equations.
 What's wrong with solve():
 --------------------------
 
-SymPy already has a pretty powerful ``solve`` function. But it has a lot of major
-issues
+SymPy already has a pretty powerful ``solve`` function. But it has some
+deficiencies. For example:
 
 1. It doesn't have a consistent output for various types of solutions
    It needs to return a lot of types of solutions consistently:
@@ -29,24 +29,21 @@ issues
    * Relational: `x > 0`
    * And the most important case: "We don't Know"
 
-2. The input API is also a mess, there are a lot of parameters. Many of them
-   are not needed and they make it hard for the user and the developers to
-   work on solvers.
+2. The input API has a lot of parameters and it can be difficult to use.
 
 3. There are cases like finding the maxima and minima of function using
    critical points where it is important to know if it has returned all the
    solutions. ``solve`` does not guarantee this.
 
-
 Why Solveset?
 -------------
 
-* ``solveset`` has a cleaner input and output interface: ``solveset`` returns
-  a set object and a set object takes care of all types of output. For
-  cases where it does not "know" all the solutions a ``ConditionSet`` with a partial
-  solution is returned. For input it only takes the equation, the variables
-  to solve for and the optional argument ``domain`` over which the equation is to
-  be solved.
+* ``solveset`` has an alternative consistent input and output interface:
+  ``solveset`` returns a set object and a set object takes care of all types of
+  output. For cases where it does not "know" all the solutions a
+  ``ConditionSet`` with a partial solution is returned. For input it only takes
+  the equation, the variables to solve for and the optional argument ``domain``
+  over which the equation is to be solved.
 
 * ``solveset`` can return infinitely many solutions. For example solving for
   `\sin{(x)} = 0` returns `\{2 n \pi | n \in \mathbb{Z}\} \cup \{2 n \pi + \pi | n \in \mathbb{Z}\}`,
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
+1 to this. This page should also be changed:
https://docs.sympy.org/latest/modules/solvers/solveset.html

There are some issues with solve that should be documented but we certainly should not discourage its use unless there is an alternative that is unambiguously superior.

The basic point to get across is that `solve` gives a finite list of explicit expressions that represent solutions to a system of equations. The docs should make it clear that using the output with `subs` is really *the* reason for wanting a function like `solve` and that e.g. `solveset` cannot satisfy that use-case as an alternative.

Giving a finite list of substitutable solutions is an important use-case but it also defines a problem that is not fully solvable in general e.g. where there are infinitely many solutions (`sin(x)=0`) or where the solutions only exist conditionally (`a*x = b` where `a` might possibly be zero). Necessarily that means that there are some caveats about the generality of the solutions returned by `solve`.

The docs should also note other limitations of solve that are things that could be improved such as:

1. Inconsistent output types: using `dict=True` fixes this and ensures that the output is always in a form suitable for direct use with `subs`.
2. Using `solve` with inequalities does something weird that it probably not what is usually wanted except in very simple cases (basically don't give inequalities to `solve`).
3. The output of `solve` does not guarantee to distinguish between a situation in which no solutions exist and a situation in which `solve` simply couldn't find the solutions (although usually an error is raised in the latter case).
4. When the form of the equations is known using `solve` can be slower than using more specific routines such as `linsolve` or the various functions for finding roots of polynomials.
5. The `solve` function does not give a simple way to specify the domain to work over (e.g. reals, integers, complex). There is limited support for distinguishing this based on the assumptions of the unknown symbols.

It's also worth noting that the `nonlinsolve` function is to some extent a dropin replacement for `solve` that might work in some cases where `solve` fails but it does not handle underdetermined systems well.

Lastly for integer solutions `diophantine` should be used.

asmeurer:
> +1 to this. This page should also be changed:
docs.sympy.org/latest/modules/solvers/solveset.html

It's that the page that is updated here?

asmeurer:
I don't know how far you wanted to go with this Jason, but I'm fine with merging this as-is if this is all you planned to do. 

Improving the solvers documentation is near the top of my list of documentation things to improve. It was the top requested thing in the survey. My main issue is that I'm not as familiar with the nuances with the solvers, as it's a part of the codebase I tend to ignore relative to others. So I will need to do some research first. 

The main thing that strikes me is that most of the complexities you mention could be fixed by improving the way the various functions work. For example, I don't understand why nonlinsolve even exists. Why does there need to be a function specifically for solving *non*linear systems? What happens if you pass a linear system to it? Why can't solveset just solve a system if you pass it a list of functions, the same way that solve() does? Why are there a half dozen different functions that can solve linear systems?

I also think that it should be possible for there to be only one underlying set of solvers internally. Then both solve() and solveset() could call them. I'm not sure to what degree is this already the case. My understanding was that they were still separate, but this may have changed. My idea is that solve() could itself just call solveset() (or the same internal functions as solveset()). It could more or less just be a wrapper to solveset() that converts a set into a list or dictionary of solutions. 

Of course, this is aspirational. For the time being, we should not be afraid to document things as they are, even if they are a mess.

oscarbenjamin:
> Isn't that the page that is updated here?

No, this update is to `solvers.rst` rather than `solveset.rst`.

> I don't know how far you wanted to go with this Jason, but I'm fine with merging this as-is if this is all you planned to do.

I'm also fine with that to be clear.

moorepants:
> I don't know how far you wanted to go with this Jason, but I'm fine with merging this as-is if this is all you planned to do.

I'll fix the solveset page too, for now, and we can merge. Other improvements can be done later.

doc/src/modules/solvers/solveset.rst
```diff
@@ -220,38 +215,28 @@ Why not use dicts as output?
...
 
- Solveset is designed to be independent of the assumptions on the
- variable being solved for and instead, uses the ``domain`` argument to
- decide the solver to dispatch the equation to, namely ``solveset_real``
- or ``solveset_complex``. It's unlike the old ``solve`` which considers the
- assumption on the variable.
+Solveset is designed to be independent of the assumptions on the variable being
+solved for and instead, uses the ``domain`` argument to decide the solver to
+dispatch the equation to, namely ``solveset_real`` or ``solveset_complex``.
+It's unlike the old ``solve`` which considers the assumption on the variable.
```
moorepants, line 239: This wasn't edited, just text wrapping and indentation.

moorepants:
Ok, I've made some modifications to both pages that hopefully is a step in the right direction.

smichr:
Thanks, @moorepants

PR #22942, opened 2022-01-28
Broadcasting index-explicit into index-implicit arrays

Upabjojr:
Extend conversion function of indexed expression to arrays to support broadcasting and addition of different indices.

NO ENTRY

### DIFF
--- a/sympy/tensor/array/expressions/tests/test_convert_index_to_array.py
+++ b/sympy/tensor/array/expressions/tests/test_convert_index_to_array.py
@@ -7,7 +7,7 @@ from sympy.tensor.indexed import IndexedBase
 from sympy.combinatorics import Permutation
 from sympy.tensor.array.expressions.array_expressions import ArrayContraction, ArrayTensorProduct, \
     ArrayDiagonal, ArrayAdd, PermuteDims, ArrayElement, _array_tensor_product, _array_contraction, _array_diagonal, \
-    _array_add, _permute_dims, ArraySymbol
+    _array_add, _permute_dims, ArraySymbol, OneArray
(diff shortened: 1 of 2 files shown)


sympy/tensor/array/expressions/tests/test_convert_index_to_array.py
```diff
@@ -167,3 +168,32 @@ def test_arrayexpr_convert_indexed_to_array_out_of_bounds():
...
+    C = ArraySymbol("C", (d0, d1))
+    D = ArraySymbol("D", (d3, d1))
+
+    expr = C[i, j] + D[k, j]
+    assert convert_indexed_to_array(expr, [i, j, k]) == ArrayAdd(ArrayTensorProduct(C, OneArray(d3)), PermuteDims(ArrayTensorProduct(OneArray(d0), D), [0, 2, 1]))
+
+    X = ArraySymbol("X", (5, 3))
+
+    expr = X[i, n] - X[j, n]
+    assert convert_indexed_to_array(expr, [i, j, n]) == ArrayAdd(ArrayTensorProduct(-1, OneArray(5), X), PermuteDims(ArrayTensorProduct(X, OneArray(5)), [0, 2, 1]))
```
Upabjojr, line 197: @ThePauliPrinciple here is your example :)

ThePauliPrinciple: Thanks, this is really useful! Is there a way to perform the opposite conversion too?

Upabjojr: The opposite should be handled by `ArrayElement`, I suppose. Probably not working at this time but it should be much easier to implement.

PR #22945, opened 2022-01-28
printing: Indexed assignment support for NumPy and TensorFlow

ThePauliPrinciple:
#### Other comments
I feel that the type checks are a bit of a mess. Perhaps it is possible to create a
two variables, one for indexed-style arrays (Indexed/ArrayElement/MatrixElement
etc) and one for numpy-like arrays (IndexedBase, Array, ArraySymbol, MatrixSymbol
etc). Not sure what would be the right place for these to exist though.

#### Release Notes

* printing
  * Assignments for indexed style objects is now supported in NumPy and TensorFlow printers.

### DIFF
--- a/sympy/codegen/ast.py
+++ b/sympy/codegen/ast.py
@@ -138,7 +138,8 @@ from sympy.core.numbers import Float, Integer, oo
 from sympy.core.sympify import _sympify, sympify, SympifyError
 from sympy.utilities.iterables import (iterable, topological_sort,
                                        numbered_symbols, filter_symbols)
-
+from sympy.tensor.array.expressions.array_expressions import ArrayElement,\
+        ArraySymbol
 
 def _mk_Tuple(args):
     """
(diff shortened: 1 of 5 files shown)


ThePauliPrinciple:
```python
from sympy.codegen.ast import Element
from sympy.matrix import Matrix
from sympy.tensor.array import Array
from sympy.tensor.array.expression.array_expressions import ArrayElement, ArraySymbol
from sympy.tensor.indexed import Indexed, IndexedBase

#tuple of classes that have explicit indices
ExplicitIndexed = (Indexed, MatrixElement, ArrayElement, Element)
#tuple of classes that have shapes (implicit indices)
ImplicitIndexed = (IndexedBase, Matrix, Array, MatrixSymbol, ArraySymbol)
```
I think something like this would be helpful, as these all work in a very similar way and it should be easy to write codegen for these groups generically.

ThePauliPrinciple:
This will be easier to implement if https://github.com/sympy/sympy/pull/22950 is accepted.

ThePauliPrinciple:
Instead of handling this in the printer, I think it would be better to implement this as some preprocessing stage before handing it to the printer.

PR #22952, opened 2022-01-29
Deprecation removals

### LINKED ISSUE #11410: Deprecated as_finite_diff function
Users found `sympy.calculus.as_finite_diff`to be confusing.
It used only to operate on `Derivative` instances, which arguably makes a method better suited to host such functionality. Furthermore people expected the function to differentiate expressions.

Hence the PR #11337 deprecated the original function and introduced a new method `Derivative.as_finite_difference` with the same (but extended) capabilities, and a new differentiating function was introduced `sympy.calculus.differentiate_finite`

Related PRs:
- Partial derivative support for as_finite_diff #11242
- Enabled support for partial derivatives in as_finite_diff #11177

Related issue:
- as_finite_diff does not support partial derivatives #11007

Note that only `as_finite_diff` needs to be removed after the deprecation cycle. The private function `_as_finite_diff` still contains the logic needed by the new method and the new function.

### LINKED ISSUE #12723: Deprecate has_Field/Ring for domains and replace with is_Field/Ring
The pr modifies `sympy/utilities/tests/test_pickling.py`. This is temporary and should be removed when `has_Field` and `has_Ring` are fully replaced.

The modification is adding `check_attr=False` to all the calls of `check()` in `test_pickling_polys_domains()`.

### LINKED ISSUE #12820: Float(prec=...) is deprecated
Previously, the argument to the `Float` constructor to set the decimal precision was called `prec`. However, everywhere else in SymPy, `prec` means binary precision, whereas `dps` means decimal precision.

To change this behavior, the Float constructor is now `Float(num, dps=None, prec=None, precision=None)`. The decimal precision can still be passed as the second argument, like `Float(1.0, 100)`. What is deprecated is explicitly calling the argument `prec`. To set the decimal precision, use `dps`. To set the binary precision, use `precision`. 

In a future version of SymPy, `prec` will be removed. Or, I don't know if this was decided, we may instead remove `precision` and make `prec` act as the binary precision, which would match the rest of SymPy. 

The PR that implemented this was https://github.com/sympy/sympy/pull/12227.

### LINKED ISSUE #12859: simplify method of DiracDelta is deprecated
The simplify method of DiracDelta is deprecated. Use `expand(diracdelta=True, wrt=x)` instead.

This was done because the `simplify` method on other objects has a different API (no `wrt` flag), which was causing issues (e.g., https://github.com/sympy/sympy/issues/7716).

### LINKED ISSUE #12860: sympy.geometry.Line.equal was renamed to sympy.geometry.Line.equals
This was done because `equals` matches the rest of the SymPy API. See https://github.com/sympy/sympy/pull/9526/

### LINKED ISSUE #12865: CoordsysCartesian is deprecated
The `CoordSysCartesian` class is deprecated. Use `CoordSys3D` instead.

This was done because currently `vector` module supports different kind of curvilinear coordinate system, so name `CoordSysCartesian` is misleading .

### LINKED ISSUE #12866: CoordSysCartesian.delop() is deprecated
`CoordSysCartesian.delop()` is deprecated. Use instance `Del()` instead. 

It was done to reflect mathematical concept, that differential operator, `Del()` doesn't depend on coordinate system. 
Currently, `delop` should be created as instance of `Del()` class. 

One behavior is different during this change. Previously, we could get information about coordinate system from `delop` because it was created inside coordinate system so for object outside the coordinate system (like `int` or `Vector.zero`) we could express them in base scalars or base vectors. Currently `delop` doesn't have this information so dealing with object also withoult this information, it is not possible to leave them unevaluated. For example, for such expression:
```python 
delop ^ Vector.zero 
```
previously we got:
``` python
Derivative(0, C.y) - Derivative(0, C.z))*C.i +
Derivative(0, C.x) + Derivative(0, C.z))*C.j +
Derivative(0, C.x) - Derivative(0, C.y))*C.k
```
but now, program returns just `0`

### LINKED ISSUE #12884: coord_sys parameter in gradient, curl and divergence is deprecated
`coord_sys` parameter in `gradient`, `curl` and `divergence` is deprecated. Please do not use any argument instead.

The reason for that is we can obtain information about coordinate system strictly from argument i.e `vect` or `scalar_field` so it is not necessary to forward them by separate argument.

### LINKED ISSUE #13336: Deprecations in sympy.physics.units
Dimensional dependencies are now part of `DimensionSystem`.

All related methods inside `Dimension` have been deprecated in favor of the methods in `DimensionSystem`.

oscargus:

Closes #11410
Closes #12723
Closes #12820
Closes #12859
Closes #12860
Closes #12865
Closes #12866
Closes #12884
Parts of #13336

Removed deprecations from SymPy 1.0, 1.1, and 1.2

#### Release Notes

* calculus
    * BREAKING: removed deprecated `as_finite_diff`.
* core
    * BREAKING: removed deprecated `prec` keyword argument to `Float`.
* functions
    * BREAKING: removed deprecated `simplify` method of `DiracDelta`.
* geometry
    * BREAKING: removed deprecated `equal` method of `Line`.
* physics.units
    * BREAKING: removed deprecated `print_unit_base` method of `UnitSystem`.
    * BREAKING: removed deprecated `desc` and `name` keyword arguments of `DimensionSystem`.
    * BREAKING: removed deprecated `description` and `name` keyword arguments of `DimensionSystem.extend`.
    * BREAKING: removed deprecated `sort_dims` method of `DimensionSystem`.
    * BREAKING: removed deprecated ability to call a `DimensionSystem`.
    * BREAKING: removed deprecated ability to get an item from a `DimensionSystem`.
* polys
    * BREAKING: removed deprecated `has_Field` and `has_Ring` methods of `Domain`.
* vector
    * BREAKING: removed deprecated `CoordSysCartesian` class.
    * BREAKING: removed deprecated `delop` method of `CoordSys3D`.
    * BREAKING: removed deprecated `system` keyword argument of `Del`.
    * BREAKING: removed deprecated `coord_sys` keyword argument of `curl`, `divergence`, and `gradient`.

### DIFF
--- a/sympy/calculus/finite_diff.py
+++ b/sympy/calculus/finite_diff.py
@@ -313,25 +313,23 @@ def _as_finite_diff(derivative, points=1, x0=None, wrt=None):
     Examples
     ========
 
-    >>> from sympy import symbols, Function, exp, sqrt, Symbol, as_finite_diff
-    >>> from sympy.utilities.exceptions import SymPyDeprecationWarning
-    >>> import warnings
-    >>> warnings.simplefilter("ignore", SymPyDeprecationWarning)
+    >>> from sympy import symbols, Function, exp, sqrt, Symbol
+    >>> from sympy.calculus.finite_diff import _as_finite_diff
     >>> x, h = symbols('x h')
     >>> f = Function('f')
-    >>> as_finite_diff(f(x).diff(x))
+    >>> _as_finite_diff(f(x).diff(x))
     -f(x - 1/2) + f(x + 1/2)
 
     The default step size and number of points are 1 and ``order + 1``
     respectively. We can change the step size by passing a symbol
     as a parameter:
 
-    >>> as_finite_diff(f(x).diff(x), h)
+    >>> _as_finite_diff(f(x).diff(x), h)
     -f(-h/2 + x)/h + f(h/2 + x)/h
 
     We can also specify the discretized values to be used in a sequence:
 
-    >>> as_finite_diff(f(x).diff(x), [x, x+h, x+2*h])
+    >>> _as_finite_diff(f(x).diff(x), [x, x+h, x+2*h])
     -3*f(x)/(2*h) + 2*f(h + x)/h - f(2*h + x)/(2*h)
 
     The algorithm is not restricted to use equidistant spacing, nor
@@ -340,7 +338,7 @@ def _as_finite_diff(derivative, points=1, x0=None, wrt=None):
 
     >>> e, sq2 = exp(1), sqrt(2)
     >>> xl = [x-h, x+h, x+e*h]
-    >>> as_finite_diff(f(x).diff(x, 1), xl, x+h*sq2)
+    >>> _as_finite_diff(f(x).diff(x, 1), xl, x+h*sq2)
     2*h*((h + sqrt(2)*h)/(2*h) - (-sqrt(2)*h + h)/(2*h))*f(E*h + x)/((-h + E*h)*(h + E*h)) +
     (-(-sqrt(2)*h + h)/(2*h) - (-sqrt(2)*h + E*h)/(2*h))*f(-h + x)/(h + E*h) +
     (-(h + sqrt(2)*h)/(2*h) + (-sqrt(2)*h + E*h)/(2*h))*f(h + x)/(-h + E*h)
@@ -349,7 +347,7 @@ def _as_finite_diff(derivative, points=1, x0=None, wrt=None):
 
     >>> y = Symbol('y')
     >>> d2fdxdy=f(x,y).diff(x,y)
-    >>> as_finite_diff(d2fdxdy, wrt=x)
+    >>> _as_finite_diff(d2fdxdy, wrt=x)
     -Derivative(f(x - 1/2, y), y) + Derivative(f(x + 1/2, y), y)
 
     See also
@@ -405,15 +403,6 @@ def _as_finite_diff(derivative, points=1, x0=None, wrt=None):
     return apply_finite_diff(order, points, [
         Derivative(derivative.expr.subs({wrt: x}), *others) for
         x in points], x0)
-
-
-as_finite_diff = deprecated(
-    useinstead="Derivative.as_finite_difference",
-    deprecated_since_version="1.1", issue=11410)(_as_finite_diff)
-
-as_finite_diff.__doc__ = """
-    Deprecated function. Use Diff.as_finite_difference instead.
-    """
 
 
 def differentiate_finite(expr, *symbols,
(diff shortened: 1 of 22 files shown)


oscarbenjamin:
Looks good to me

sympy/calculus/finite_diff.py
```diff
@@ -313,25 +312,23 @@ def _as_finite_diff(derivative, points=1, x0=None, wrt=None):
     Examples
     ========
 
-    >>> from sympy import symbols, Function, exp, sqrt, Symbol, as_finite_diff
-    >>> from sympy.utilities.exceptions import SymPyDeprecationWarning
-    >>> import warnings
-    >>> warnings.simplefilter("ignore", SymPyDeprecationWarning)
+    >>> from sympy import symbols, Function, exp, sqrt, Symbol
+    >>> from sympy.calculus.finite_diff import _as_finite_diff
```
smichr, line 316: can we just use `finite_diff`? why the underscore?

oscargus: Because it earlier said `as_finite_diff` and that called `_as_finite_diff`?

I didn't look more into it, but can do.

oscargus: Not sure if I fully understand the comment. The name `as_finite_diff` was deprecated and `Derivative.as_finite_difference` should be used instead. Hence, reintroducing `finite_diff` as a public function didn't seem to be the intention then.

Maybe one should not have those code examples in the documentation for `_as_finite_diff`? Not sure if they will show up in the documentation (maybe private functions/methods shouldn't?) though, but for development documentation I think it can make sense.

asmeurer: Both public and private functions are included in the docs site. 

I'm not entirely clear what this function is, but the doctests should use the same name of the function, so if the function is called _as_finite_diff that's what the doctest should use. 

This actually should have been updated to this as soon as the function was deprecated. We should always document non-deprecated behavior. Using `warnings.simplefilter("ignore", SymPyDeprecationWarning)` in a doctest is a bad sign.

asmeurer:
I'm going to merge this so that I can include it in https://github.com/sympy/sympy/pull/22900 and avoid wasting time documenting these deprecations that are being removed anyway.

asmeurer:
Regarding the remaining part of #13336, the remaining method, is_dimensionless, doesn't have much explanation, either in the issue or the warning, but I think the point is that dimensions are not dimensionless. The current implementation raises a warning unless it is `Dimension(1)`, which means that effectively the entire method should be removed to remove the deprecation. I am going to do this in #22900 as I see no point in removing only part of that deprecation.

oscargus:
I agree. Was just too lazy to really think about it (as I had covid, I just wanted to do something mechanical...).

asmeurer:
No problem. Just wanted to note here what was being done. Hope you are feeling better.

PR #22959, opened 2022-01-30
nonlinsolve now checks if solutions are correct

### LINKED ISSUE #22628: nonlinsolve is incorrect
`nonlinsolve([h - 1, k - 1, f - 2, f - 4, -2*k], h, k, f)` returns `{(1, 1, f)}`, but this system doesn't have a solution.

### LINKED ISSUE #22874: nonlinsove returns nan
`nonlinsolve([-2*(x - exp(y)), (2*x - 3*exp(2*y))*exp(y)],x,y)` returns `{(nan, nan), (2*e^(2*_n*i*pi)/3, ImageSet(Lambda(_n, 2*_n*i*pi + ln(2/3)), Integers))}`

NikhilSDate:

Fixes #22628 
Fixes #22874 

`nonlinsolve` now checks if the solutions it produces are valid and only returns the valid solutions. 

Currently, this is only done for solutions without any `ConditionSets`, `ImageSets` etc. or any extra symbols in the solution.
An example of the second case is `nonlinsolve([sin(x)*sin(y), cos(x)*cos(y) - 1], [x, y])`, which returns `FiniteSet((x, 2*n*pi), (2*n*pi, y), (x, 2*n*pi + pi), (2*n*pi + pi, y))` (on master). This is required because `checksol` can fail with such solutions due to `is_constant` not working correctly.

Incorrect solutions in some tests have been removed, and new tests have been added for #22628 and #22874 
 

#### Other comments
#22866 needs to be merged first for this PR to pass all tests. `test_nonlinsolve_basic` in `test_solveset` currently causes an exception due to the issue fixed in #22866. I have checked that all other tests in the `solvers` module pass with this PR.
The failing authors check is also fixed in #22866

#### Release Notes

* solvers
  * nonlinsolve returns fewer incorrect solutions

### DIFF
--- a/sympy/solvers/solveset.py
+++ b/sympy/solvers/solveset.py
@@ -16,6 +16,7 @@ from sympy.core import (S, Pow, Dummy, pi, Expr, Wild, Mul, Equality,
 from sympy.core.containers import Tuple
 from sympy.core.function import (Lambda, expand_complex, AppliedUndef,
                                 expand_log, _mexpand, expand_trig)
+from sympy.core.kind import NumberKind
 from sympy.core.mod import Mod
 from sympy.core.numbers import igcd, I, Number, Rational, oo, ilcm
 from sympy.core.power import integer_log
@@ -3695,16 +3696,11 @@ def nonlinsolve(system, *symbols):
         result = substitution(
             polys_expr + nonpolys, symbols, exclude=denominators)
     valid_solns = []
-    idx = 0
     if isinstance(result, FiniteSet):
         for soln in result.args:
-            values = dict(zip(symbols, soln))
-            try:
-                if checksol(system, values) != False:
-                    valid_solns.append(soln)
-            except:
+            num_values = {sym: val for (sym, val) in zip(symbols, soln) if not isinstance(val, Set)}
+            if checksol(system, num_values) != False:
                 valid_solns.append(soln)
-            idx += 1
         if len(valid_solns) > 0:
             return FiniteSet(*valid_solns)
         else:
--- a/sympy/solvers/solveset.py
+++ b/sympy/solvers/solveset.py
@@ -3694,8 +3694,8 @@ def nonlinsolve(system, *symbols):
         # Use `substitution` method for the system
         result = substitution(
             polys_expr + nonpolys, symbols, exclude=denominators)
-    valid_solns = []
     if isinstance(result, FiniteSet):
+        valid_solns = []
         for soln in result.args:
             # the second of the following conditions is a hack
             # it is required because checksol can fail with cases like sin(n*pi)
--- a/sympy/solvers/solveset.py
+++ b/sympy/solvers/solveset.py
@@ -3694,4 +3694,19 @@ def nonlinsolve(system, *symbols):
         # Use `substitution` method for the system
         result = substitution(
             polys_expr + nonpolys, symbols, exclude=denominators)
-        return result
+    valid_solns = []
+    idx = 0
+    if isinstance(result, FiniteSet):
+        for soln in result.args:
+            values = dict(zip(symbols, soln))
+            try:
+                if checksol(system, values) != False:
+                    valid_solns.append(soln)
+            except:
+                valid_solns.append(soln)
+            idx += 1
+        if len(valid_solns) > 0:
+            return FiniteSet(*valid_solns)
+        else:
+            return S.EmptySet
+    return result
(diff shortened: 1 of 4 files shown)


sympy/solvers/solveset.py
```diff
@@ -3694,4 +3694,17 @@ def nonlinsolve(system, *symbols):
         # Use `substitution` method for the system
         result = substitution(
             polys_expr + nonpolys, symbols, exclude=denominators)
-        return result
+    valid_solns = []
+    if isinstance(result, FiniteSet):
```
oscargus, line 3698:
```suggestion
    if isinstance(result, FiniteSet):
    valid_solns = []
```

oscargus:
```
 _________ sympy/solvers/tests/test_solveset.py:test_nonlinsolve_basic __________
  File "/home/runner/work/sympy/sympy/sympy/solvers/tests/test_solveset.py", line 1562, in test_nonlinsolve_basic
    assert nonlinsolve([x - y, 0], x, y) == soln
  File "/home/runner/work/sympy/sympy/sympy/solvers/solveset.py", line 3704, in nonlinsolve
    or checksol(system, dict(zip(symbols, soln))) != False:
  File "/home/runner/work/sympy/sympy/sympy/solvers/solvers.py", line 251, in checksol
    check = checksol(fi, sol, **flags)
  File "/home/runner/work/sympy/sympy/sympy/solvers/solvers.py", line 274, in checksol
    elif not f.is_Relational and not f:
AttributeError: 'int' object has no attribute 'is_Relational'
```

It seems like one should probably `_sympify` the `system` argument

NikhilSDate:
> It seems like one should probably _sympify the system argument

@oscargus I have handled this in #22866. This error is actually the reason I created that PR.

sympy/solvers/solveset.py
NikhilSDate: Seems like a good suggestion. I've committed the change.

oscarbenjamin:
In many situations `solve` is very slow because of checking solutions retrospectively like this. It would be much better to avoid needing to check anything. The example in #22628 is linear with pure integer coefficients and it should be completely possible to discover that this does not have a solution at the time of solving the linear system rather than checking afterwards.

NikhilSDate:
@oscarbenjamin How about an optional `check` flag that is `False` by default? This wouldn't change any current behavior, but would give the option to check solutions if a user wants it. This would also not change the output format in any way, but would only remove invalid solutions. I also noticed that `_handle_zero_dimensional` uses `checksol`, maybe that could also be changed so that it only checks solutions if the `check` flag is `True`.

oscarbenjamin:
> How about an optional `check` flag that is `False` by default?

I think that's a bad idea. Checking should only be done when it is needed. Otherwise the solver should avoid generating solutions that need to be checked.

oscarbenjamin:
> > Checking should only be done when it is needed.
> 
> What could be the situations in which checking is needed?

Checking can be needed e.g. if an equation involves radicals and something like unrad is used because it can introduce spurious solutions. If nothing introduces spurious solutions then no checking is needed.

NikhilSDate:
@oscarbenjamin Ok. I see. If the substitution method is not known to return spurious solutions, I guess this can be closed. I will open a separate PR with the fix for the broken if condition discussed in #22628, if that sounds good. The other changes you suggested in #22628 might take a while to implement, so I think it might be better to fix the easily solvable problem first (that will also end up fixing #22628). What do you think?

oscarbenjamin:
The `if` condition should be changed but I don't really consider that a full fix for the issue because it won't be hard to make other examples that have the same problem e.g.
```python
In [1]: nonlinsolve([exp(y) - 1, exp(y) - 2], [y])
Out[1]: {(log(2),), ({2⋅n⋅ⅈ⋅π + log(2) │ n ∊ ℤ},)}
```

NikhilSDate:
I think fixing the if condition should at least fix all such issues with purely polynomial systems. For systems like your example, this seems to be an issue with `substitution`. 
```
>>> substitution([exp(y) - 1, exp(y) - 2], [y])
{(log(2),), (ImageSet(Lambda(_n, 2*_n*I*pi + log(2)), Integers),)}

oscarbenjamin:
> For systems like your example, this seems to be an issue with `substitution`.

The issue is with substitution in all cases. Changing the `if` means we avoid using substitution but it doesn't fix the bug in substitution.

NikhilSDate:
Yes, that's what I meant. I guess it could have been clearer. Do you think it is worth it to fix the `if` separate from fixing the underlying bug with `substitution`?

oscarbenjamin:
> Do you think it is worth it to fix the `if` separate from fixing the underlying bug with `substitution`?

Yes.

NikhilSDate:
Ok. I'll close this for now and open a separate PR for that then. Thanks for the feedback on my questions!

NikhilSDate:
I have created #23011 to fix the `if` condition.

PR #22971, opened 2022-01-31
Documentation cleanup 5

oscargus:

Primarily LaTeX:ifying stuff, but some general fixes as well.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/stats/joint_rv_types.py
+++ b/sympy/stats/joint_rv_types.py
@@ -97,14 +97,14 @@ class JointDistributionHandmade(JointDistribution):
 
 def JointRV(symbol, pdf, _set=None):
     """
-    Create a Joint Random Variable where each of its component is conitinuous,
+    Create a Joint Random Variable where each of its component is continuous,
     given the following:
 
     -- a symbol
     -- a PDF in terms of indexed symbols of the symbol given
     as the first argument
 
-    NOTE: As of now, the set for each component for a `JointRV` is
+    NOTE: As of now, the set for each component for a ``JointRV`` is
     equal to the set of all integers, which cannot be changed.
 
     Examples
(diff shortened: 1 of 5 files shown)


sympy/stats/joint_rv_types.py
```diff
@@ -713,9 +713,9 @@ def GeneralizedMultivariateLogGamma(syms, delta, v, lamda, mu):
     ====
 
     If the GeneralizedMultivariateLogGamma is too long to type use,
-    `from sympy.stats.joint_rv_types import GeneralizedMultivariateLogGamma as GMVLG`
+    ``from sympy.stats.joint_rv_types import GeneralizedMultivariateLogGamma as GMVLG``
```
ThePauliPrinciple, line 716: Is this instruction really needed? Users can import it as any name as far as I am concerned.

ThePauliPrinciple:
Changes make sense, looks good to me. Seems like the `.mailmap` needs to be updated.

sympy/stats/joint_rv_types.py
oscargus: I wasn't convinced either. But at least it should be properly formatted...

oscargus:
> Seems like the `.mailmap` needs to be updated.

Have to figure out why that email shows up here... Should probably fix one of my computers to use my private email...

oscarbenjamin:
> Have to figure out why that email shows up here...

There's only one commit here and that's the email address it shows:
https://github.com/sympy/sympy/commit/c9f6ba02dfa7ccc1b8b408d5b07d1ab96f11eb0d.patch
I'd say whichever computer you used to make this PR.

oscargus:
I have force pushed from different computers and apparently I had things not correctly setup on the first and the address is not changed "automatically" when force pushing from the one with correct email. Now I have changed it manually though (I hope).

PR #22973, opened 2022-01-31
Add convenience methods for constructing `AlgebraicField`

### LINKED ISSUE #22954: polys/numberfields: Convenience constructors for `AlgebraicField`
I'd like to propose a pair of new constructor methods, to support a couple of common use cases.

## Use Case 1

* You want a number field `Q(a)`,
* AND you have the minimal polynomial `f(x)` for `a`,
* AND you don't care _which_ root of `f` `a` is. (E.g. it's a Galois field, so it doesn't matter.),
* AND, OPTIONALLY you want an `alias` for `a` so that field elements print nicely.

## Use Case 2

* You want the `n`th cyclotomic field.
* You probably want the primitive element to be aliased either `"zeta"` or `f"zeta{n}"`.

## The Current Situation

For Use Case 1, you have to do this:
```python
alpha = AlgebraicNumber(CRootOf(f, -1), alias="alpha")
K = QQ.algebraic_field(alpha)

```
which has several problems:

* It's too long.
* You shouldn't have to import `CRootOf`.
* You shouldn't have to choose an index for `CRootOf`.

Use Case 2 is similar, and you also have to import `cyclotomic_poly`:

```python
zeta = AlgebraicNumber(CRootOf(cyclotomic_poly(19), -1), alias="zeta")
K = QQ.algebraic_field(zeta)
```

## Proposal

Alongside the `QQ.algebraic_field()` method (which takes one arg, being an `Expr`) I would propose two additional methods:

```python
    def alg_field_from_poly(self, poly, alias=None, root_index=-1):
        alpha = AlgebraicNumber(CRootOf(poly, root_index), alias=alias)
        return self.algebraic_field(alpha)

    def cyclotomic_field(self, n, alias="zeta", ss=False, gen=None, root_index=-1):
        if ss:
            alias += str(n)
        return self.alg_field_from_poly(cyclotomic_poly(n, gen), alias=alias, root_index=root_index)

```

Then the above examples become

```python
K = QQ.alg_field_from_poly(f, "alpha")
```

and

```python
K = QQ.cyclotomic_field(19)
```

## Rationale

For a "canonical" default root, `-1` is the best index to pass to `CRootOf`. In common cases like quadratic or cyclotomic fields, this will select the root you tend to think of: the positive square root, or the square root with positive imaginary part, or `exp(2*I*pi/n)` for the `n`th cyclotomic field.

skieffer:

Fixes #22954 

Adds a couple of convenience methods for constructing `AlgebraicField`.

Serves users who don't care about the complex embedding, by choosing an embedding for them.

In particular, makes constructing a cyclotomic field much easier.

#### Release Notes

* polys
  * Add convenience constructors for `AlgebraicField`

### DIFF
--- a/sympy/polys/domains/domain.py
+++ b/sympy/polys/domains/domain.py
@@ -922,7 +922,7 @@ class Domain:
         >>> K.ext.minpoly == f
         True
         >>> g = Poly(8*x**3 - 6*x - 1)
-        >>> L = QQ.alg_field_from_poly(g, alias="alpha")
+        >>> L = QQ.alg_field_from_poly(g, "alpha")
         >>> L.ext.minpoly == g
         True
         >>> L.to_sympy(L([1, 1, 1]))
@@ -962,7 +962,7 @@ class Domain:
         >>> K = QQ.cyclotomic_field(5)
         >>> K.to_sympy(K([-1, 1]))
         1 - zeta
-        >>> L = QQ.cyclotomic_field(7, ss=True)
+        >>> L = QQ.cyclotomic_field(7, True)
         >>> a = L.to_sympy(L([-1, 1]))
         >>> print(a)
         1 - zeta7
--- a/sympy/polys/domains/domain.py
+++ b/sympy/polys/domains/domain.py
@@ -3,6 +3,7 @@
 
 from typing import Any, Optional, Type
 
+from sympy.core.numbers import AlgebraicNumber
 from sympy.core import Basic, sympify
 from sympy.core.sorting import default_sort_key, ordered
 from sympy.external.gmpy import HAS_GMPY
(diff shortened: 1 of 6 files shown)


sympy/polys/domains/domain.py
```diff
@@ -891,6 +892,90 @@ def algebraic_field(self, *extension):
...
+        >>> L.ext.minpoly == g
+        True
+        >>> L.to_sympy(L([1, 1, 1]))
+        alpha**2 + alpha + 1
+
+        """
+        from sympy.polys.rootoftools import CRootOf
+        root = CRootOf(poly, root_index)
+        alpha = AlgebraicNumber(root, alias=alias)
+        return self.algebraic_field(alpha)
```
oscarbenjamin, line 935: Maybe it would be useful for this method to return the generator as well.

skieffer: I don't think it's necessary, since it can be easily accessed from the field:
```python
In [4]: K = QQ.alg_field_from_poly(Poly(8*x**3 - 6*x - 1))

In [5]: K.ext
Out[5]: CRootOf(8*x**3 - 6*x - 1, 2)
```

oscarbenjamin: Oh, of course.

oscarbenjamin:
Looks good to me

sympy/polys/domains/domain.py
```diff
@@ -891,6 +892,90 @@ def algebraic_field(self, *extension):
...
+
+    def cyclotomic_field(self, n, alias="zeta", ss=False, gen=None, root_index=-1):
```
skieffer, line 937: Checking this over, I realized the order of the args `ss` and `alias` should be swapped. It will be more common to want to activate the subscript than to want to use a symbol other than `zeta`.

I'll make this change and (if still no objections) will merge soon.

Issue #22974, opened 2022-01-31
polys/numberfields: `PrimeIdeal` printing methods need work.


skieffer:
Two methods in the `PrimeIdeal` class, `pretty()` (for printing) and `reduce_poly()`, involve generating polynomial expressions. Ideally, the generator symbol in these expressions should be able to be automatically the same as the `alias` symbol for the generator of the surrounding `AlgebraicField`.

To do this, instances of `PrimeIdeal` will need a reference to that field, which they currently lack.

skieffer:
So, in #22977 these two methods were made private, so that they could still be changed after v1.10.
Here are some thoughts on improving them for v1.11:

## `PrimeIdeal.pretty`

First, this is misnamed, because it's not "pretty printing" in the sense of 2D displays. It's really just a `repr`-type string.

Apart from this, we should have better `__str__` and LaTeX printing for the `PrimeIdeal` class, along the lines discussed in the OP (i.e. where the alias of the primitive element of the associated `AlgebraicField` (if any) can be automatically used).

Before that can happen though, it would be best if #23338 could be resolved.


## `PrimeIdeal.reduce_poly`

I'd like to improve this, and keep it around because it is useful, but I'd like to keep it private (leading underscore) in case we want to change the argument or return types as the need becomes more clear.

The basic operation is an important one: we're mapping elements of a ring of integers `R` into the finite field `R/P`, where `P` is a prime ideal. So it's like the [`nfmodpr`](http://pari.math.u-bordeaux.fr/dochtml/html-stable/General_number_fields.html#nfmodpr) function of PARI.

The existing implementation is funny though, in that it accepts a `Poly`. It makes more sense to accept a class like `ANP` or `AlgebraicNumber` (or both), which represents an algebraic number more directly. (And the method should be called something simpler like `_reduce()`).

As for the return value, we're using a `Poly` mod `p`, which does make sense as a representation of an element of a finite field of order `p^n`, but again it's not clear if this should change in the future. We might just want a `dup` instead of a `Poly`, or at some point SymPy might provide `GF(p^n)` as a domain type (for `n > 1`), and then we might want it to be one of those (in particular if `GF(p^n)` lets you choose the irreducible poly by which it's defined).

skieffer:
I've opened a separate issue (#23403) for the `PrimeIdeal.reduce_poly()` method, so that it can
be handled separately.

The present issue should be considered resolved if we fix up the printing problems for `PrimeIdeal`.

PR #22975, opened 2022-01-31
Mark two `PrimeIdeal` methods non-public.

skieffer:

Temporary way of addressing #22974 

Makes two `PrimeIdeal` methods non-public.

#### Other comments

A proper solution to #22974 may take a little time. It would be hard to squeeze that in before the release of v1.10. For now, I think it's best to keep these methods out of the public API, so that we don't have to commit to backwards compatibility on them.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/polys/numberfields/primes.py
+++ b/sympy/polys/numberfields/primes.py
@@ -70,7 +70,7 @@ class PrimeIdeal(IntegerPowerable):
         self._test_factor = None
         self.e = e if e is not None else self.valuation(p * ZK)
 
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
This looks good to me. To make into the release the PR needs to be also opened against the 1.10 branch. You can use cherry-pick for this e.g.:
```
git fetch upstream
git checkout 1.10
git checkout -b pr_id_nonpublic_methods_110
git cherry-pick pr_id_nonpublic_methods
git push # etc
```
Then when opening the PR change the base branch to 1.10.

PR #22985, opened 2022-02-01
stats - added Beta negative binomial distribution in drv_types.

praneethratna:
In probability theory, a [beta negative binomial distribution](https://en.wikipedia.org/wiki/Beta_negative_binomial_distribution) is the probability distribution of a discrete random variable X equal to the number of failures needed to get r successes in a sequence of independent Bernoulli trials where the probability p of success on each trial, while constant within any given experiment, is itself a random variable following a beta distribution, varying between different experiments.

#### Release Notes

* stats
   * added beta negative binomial distribution in drv_types.py.

### DIFF
--- a/sympy/core/tests/test_args.py
+++ b/sympy/core/tests/test_args.py
@@ -1662,6 +1662,10 @@ def test_sympy__stats__drv_types__ZetaDistribution():
     assert _test_args(ZetaDistribution(1.5))
 
 
+def test_sympy__stats__drv_types__BetaNegativeBinomialDistribution():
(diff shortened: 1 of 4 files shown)


praneethratna:
@oscarbenjamin

praneethratna:
Ping @czgdp1807 for review

czgdp1807:
LGTM. Though I would suggest other reviewers to make sure this is not a special case of an already present distribution. If I find time I will do that.

praneethratna:
@czgdp1807 are there any changes required ?

PR #22989, opened 2022-02-01
MatchPy connector: add support for pickle

Upabjojr:
Just realized that the objects in `sympy.utilities.matchpy_connector` cannot be pickled. This is a problem when using RUBI.

NO ENTRY

### DIFF
--- a/sympy/utilities/matchpy_connector.py
+++ b/sympy/utilities/matchpy_connector.py
@@ -117,8 +117,8 @@ else:
 
 @doctest_depends_on(modules=('matchpy',))
 class _WildAbstract(Wildcard, Symbol):
-    min_length: int # abstract field required in subclasses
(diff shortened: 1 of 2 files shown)


oscargus:
I was just going to give moving rubi to its own repository a go. Are you expecting to do more rubi-related work in the near future? (Just want to avoid unnecessary work for both of us.)

Upabjojr:
> I was just going to give moving rubi to its own repository a go. Are you expecting to do more rubi-related work in the near future? (Just want to avoid unnecessary work for both of us.)

This PR is unrelated to RUBI. The RUBI module has its own matchpy connector as of now (which is not optimal, this one should be used).

This is a generic matchpy connector for all of SymPy. It will work if matchpy is installed. This should remain in SymPy.

Issue #22277, opened 2021-10-14
Hardcoded "return" in the lambdify printer prevents multiline code in lambdify


ThePauliPrinciple:
Currently, the lambdify printer uses a string with "return" in order to return the result of the expr it is trying to evaluate. This precludes e.g. multiline AST `CodeBlocks`. Instead, it should be using the AST `Return`

ThePauliPrinciple:
After going through the `lambdify` code, I feel that adding full AST support will give friction with the parsing that is already performed. So instead an ast_lambdify in codegen/ast.py would be appropriate. In this case, it would accept an AST `FunctionDefinition` only. It will then only print this  `FunctionDefinition`  as is and apply the appropriate namespace magic. Any arbitraty AST node structure can be stored in the AST `FunctionDefinition`'s body attribute.

ThePauliPrinciple:
Is there any particular reason why lambdify is originally written as a function rather than a class? Would it be benificial to rewrite it as a class?

ThePauliPrinciple:
Giving it some further thought and making use of what is available in the python code printer I arrived at this solution for now:
```python
from sympy.codegen.ast import FunctionDefinition
from sympy.printing.pycode import AbstractPythonCodePrinter

class ASTLambdify():
  
  _lambdify_generated_counter=0

  def __new__(cls,*args,**kwargs):
    cls._lambdify_generated_counter+=1
    return super().__new__(cls)

  def __init__(self,ast,printer,shortened_names=False):    
    if not isinstance(ast,FunctionDefinition):
      TypeError('{0} is a {1} and not a (derivative of) {2}'.format(ast,type(ast),FunctionDefinition))
    if not isinstance(ast,AbstractPythonCodePrinter):
      TypeError('{0} is a {1} and not a (derivative of) {2}'.format(printer,type(printer),AbstractPythonCodePrinter))
    self.ast=ast
    self.printer=printer
    self.code=printer.doprint(ast)
    self._namespace(shortened_names)
    self._exec_function()

  def _exec_function(self):
    funclocals={}
    c = compile(self.code, self.filename, 'exec')
    exec(c, self.namespace, funclocals)
    self.function=funclocals[self.function_name]

  @property 
  def filename(self):
    #filenames generated by counter as class variable + class name for easier debugging with multiple lambdify classes?
    return self.__class__.__name__ + '-' + str(self._lambdify_generated_counter)

  @property
  def function_name(self):
    return self.ast.name.text

  @property
[...]

Issue #21751, opened 2021-07-15
lambdify produces incorrect results


Davide-sd:
I think there is an error somewhere within `lambdify`. Here is an example code, comparing the output of `lambdify` with the one produced by the `experimental_lambdify` module (from sympy.plotting). I also printed the content of the lambda functions.

According to [wolfram alpha](https://www.wolframalpha.com/input/?i=imag%28sqrt%28-x%29%29), the correct result is given by `experimental_lambdify`.

```
import matplotlib.pyplot as plt
import numpy as np
from sympy import im, sqrt, symbols, lambdify
from sympy.plotting.experimental_lambdify import lambdify as lam

x = symbols("x")
expr = im(sqrt(-x))
f1 = lambdify([x], expr)
f2 = lam([x], expr)

xx = np.linspace(-10, 10)
fig = plt.figure()
ax = fig.add_subplot(1, 1, 1)
ax.plot(xx, f1(xx), label="lambdify")
ax.plot(xx, [f2(t) for t in xx], label="lam")
plt.legend()
plt.show()

import inspect
print(inspect.getsource(f1))
print(f2.lambda_func.eval_str)

# def _lambdifygenerated(x):
#    return ((real(x)**2 + imag(x)**2)**(1/4)*sin((1/2)*arctan2(-imag(x), -real(x))))

# lambda x0 : ( ((re(x0)).evalf()**2 + (im(x0)).evalf()**2)**(1/4)*cmath.sin((atan2(-(im(x0)).evalf(), -(re(x0)).evalf())).evalf()/2) )
```

![Figure 3(1)](https://user-images.githubusercontent.com/9921777/125817594-ac7863f2-d638-472d-af84-6d7bc415a117.png)

Davide-sd:
Maybe it's something deep within `numpy`. If I use `mpmath` as the module for `f1`, it produces the correct result...

oscarbenjamin:
This might be a clue:
```python
In [2]: f1(10)
Out[2]: 3.1622776601683795

In [3]: f1(10.0)
Out[3]: -3.1622776601683795
```

oscarbenjamin:
The issue is to do with signed zero in floating point:
```python
In [39]: arctan2(-0, -10)
Out[39]: 3.141592653589793

In [40]: arctan2(-0.0, -10)
Out[40]: -3.141592653589793
```

Davide-sd:
Link to NumPy issue: https://github.com/numpy/numpy/issues/19484

[...]

PR #22993, opened 2022-02-02
Enable automatic rewrite for cot, sec, csc, and related inverse and hyperbolic functions

### LINKED ISSUE #22992: Lambdify Error: 'NameError: name 'cot' is not defined'
When attempting to lambify an expression with cot:

image

The following error is displayed:

image

oscargus:

Closes #22992 

`cot` is automatically rewritten to `tan` to support lambdifying.

#### Release Notes

* printing
   * `cot`, `sec`, `csc` and related inverse and hyperbolic functions are automatically rewritten if they cannot be printed. Enables `lambdify` for these expressions. 
* functions
    * hyperbolic functions can be rewritten to more other hyperbolic functions.

### DIFF
--- a/sympy/functions/elementary/hyperbolic.py
+++ b/sympy/functions/elementary/hyperbolic.py
@@ -1,19 +1,17 @@
-from sympy.core.logic import FuzzyBool
-
 from sympy.core import S, sympify, cacheit, pi, I, Rational
 from sympy.core.add import Add
 from sympy.core.function import Function, ArgumentIndexError
-from sympy.core.logic import fuzzy_or, fuzzy_and
+from sympy.core.logic import fuzzy_or, fuzzy_and, FuzzyBool
 from sympy.functions.combinatorial.factorials import (binomial, factorial,
                                                       RisingFactorial)
 from sympy.functions.combinatorial.numbers import bernoulli, euler, nC
 from sympy.functions.elementary.complexes import Abs
 from sympy.functions.elementary.exponential import exp, log, match_real_imag
-from sympy.functions.elementary.miscellaneous import sqrt
 from sympy.functions.elementary.integers import floor
-from sympy.functions.elementary.trigonometric import (acot, asin, atan, cos,
-                                              cot, sin, tan,
-                                              _imaginary_unit_as_coefficient)
+from sympy.functions.elementary.miscellaneous import sqrt
+from sympy.functions.elementary.trigonometric import (
+    acos, acot, asin, atan, cos, cot, csc, sec, sin, tan,
+    _imaginary_unit_as_coefficient)
 from sympy.polys.specialpolys import symmetric_poly
 
 
@@ -221,6 +219,12 @@ class sinh(HyperbolicFunction):
     def _eval_rewrite_as_exp(self, arg, **kwargs):
         return (exp(arg) - exp(-arg)) / 2
 
+    def _eval_rewrite_as_sin(self, arg, **kwargs):
+        return -I * sin(I * arg)
+
+    def _eval_rewrite_as_csc(self, arg, **kwargs):
+        return -I / csc(I * arg)
+
     def _eval_rewrite_as_cosh(self, arg, **kwargs):
         return -S.ImaginaryUnit*cosh(arg + S.Pi*S.ImaginaryUnit/2)
 
@@ -231,6 +235,9 @@ class sinh(HyperbolicFunction):
     def _eval_rewrite_as_coth(self, arg, **kwargs):
         coth_half = coth(S.Half*arg)
         return 2*coth_half/(coth_half**2 - 1)
+
+    def _eval_rewrite_as_csch(self, arg, **kwargs):
+        return 1 / csch(arg)
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
         arg = self.args[0].as_leading_term(x, logx=logx, cdir=cdir)
@@ -409,6 +416,12 @@ class cosh(HyperbolicFunction):
     def _eval_rewrite_as_exp(self, arg, **kwargs):
         return (exp(arg) + exp(-arg)) / 2
 
+    def _eval_rewrite_as_cos(self, arg, **kwargs):
+        return cos(I * arg)
+
+    def _eval_rewrite_as_sec(self, arg, **kwargs):
+        return 1 / sec(I * arg)
+
     def _eval_rewrite_as_sinh(self, arg, **kwargs):
         return -S.ImaginaryUnit*sinh(arg + S.Pi*S.ImaginaryUnit/2)
 
@@ -419,6 +432,9 @@ class cosh(HyperbolicFunction):
     def _eval_rewrite_as_coth(self, arg, **kwargs):
         coth_half = coth(S.Half*arg)**2
         return (coth_half + 1)/(coth_half - 1)
+
+    def _eval_rewrite_as_sech(self, arg, **kwargs):
+        return 1 / sech(arg)
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
         arg = self.args[0].as_leading_term(x, logx=logx, cdir=cdir)
@@ -657,6 +673,12 @@ class tanh(HyperbolicFunction):
     def _eval_rewrite_as_exp(self, arg, **kwargs):
         neg_exp, pos_exp = exp(-arg), exp(arg)
         return (pos_exp - neg_exp)/(pos_exp + neg_exp)
+
+    def _eval_rewrite_as_tan(self, arg, **kwargs):
+        return -I * tan(I * arg)
+
+    def _eval_rewrite_as_cot(self, arg, **kwargs):
+        return -I / cot(I * arg)
 
     def _eval_rewrite_as_sinh(self, arg, **kwargs):
         return S.ImaginaryUnit*sinh(arg)/sinh(S.Pi*S.ImaginaryUnit/2 - arg)
@@ -1017,8 +1039,17 @@ class csch(ReciprocalHyperbolicFunction):
 
             return 2 * (1 - 2**n) * B/F * x**n
 
+    def _eval_rewrite_as_sin(self, arg, **kwargs):
+        return I / sin(I * arg)
+
+    def _eval_rewrite_as_csc(self, arg, **kwargs):
+        return I * csc(I * arg)
+
     def _eval_rewrite_as_cosh(self, arg, **kwargs):
         return S.ImaginaryUnit / cosh(arg + S.ImaginaryUnit * S.Pi / 2)
+
+    def _eval_rewrite_as_sinh(self, arg, **kwargs):
+        return 1 / sinh(arg)
 
     def _eval_is_positive(self):
         if self.args[0].is_extended_real:
@@ -1067,8 +1098,17 @@ class sech(ReciprocalHyperbolicFunction):
             x = sympify(x)
             return euler(n) / factorial(n) * x**(n)
 
+    def _eval_rewrite_as_cos(self, arg, **kwargs):
+        return 1 / cos(I * arg)
+
+    def _eval_rewrite_as_sec(self, arg, **kwargs):
+        return sec(I * arg)
+
     def _eval_rewrite_as_sinh(self, arg, **kwargs):
         return S.ImaginaryUnit / sinh(arg + S.ImaginaryUnit * S.Pi /2)
+
+    def _eval_rewrite_as_cosh(self, arg, **kwargs):
+        return 1 / cosh(arg)
 
     def _eval_is_positive(self):
         if self.args[0].is_extended_real:
@@ -1186,6 +1226,19 @@ class asinh(InverseHyperbolicFunction):
 
     def _eval_rewrite_as_log(self, x, **kwargs):
         return log(x + sqrt(x**2 + 1))
+
+    def _eval_rewrite_as_atanh(self, x, **kwargs):
+        return atanh(x/sqrt(1 + x**2))
+
+    def _eval_rewrite_as_acosh(self, x, **kwargs):
+        ix = I*x
+        return I*(sqrt(1 - ix)/sqrt(ix - 1) * acosh(ix) - pi/2)
+
+    def _eval_rewrite_as_asin(self, x, **kwargs):
+        return -I * asin(I * x)
+
+    def _eval_rewrite_as_acos(self, x, **kwargs):
+        return I * acos(I * x) - I*pi/2
 
     def inverse(self, argindex=1):
         """
@@ -1337,6 +1390,22 @@ class acosh(InverseHyperbolicFunction):
     def _eval_rewrite_as_log(self, x, **kwargs):
         return log(x + sqrt(x + 1) * sqrt(x - 1))
 
+    def _eval_rewrite_as_acos(self, x, **kwargs):
+        return sqrt(x - 1)/sqrt(1 - x) * acos(x)
+
+    def _eval_rewrite_as_asin(self, x, **kwargs):
+        return sqrt(x - 1)/sqrt(1 - x) * (pi/2 - asin(x))
+
+    def _eval_rewrite_as_asinh(self, x, **kwargs):
+        return sqrt(x - 1)/sqrt(1 - x) * (pi/2 + I*asinh(I*x))
+
+    def _eval_rewrite_as_atanh(self, x, **kwargs):
+        sxm1 = sqrt(x - 1)
+        s1mx = sqrt(1 - x)
+        sx2m1 = sqrt(x**2 - 1)
+        return (pi/2*sxm1/s1mx*(1 - x * sqrt(1/x**2)) +
+                sxm1*sqrt(x + 1)/sx2m1 * atanh(sx2m1/x))
+
     def inverse(self, argindex=1):
         """
         Returns the inverse of this function.
@@ -1442,6 +1511,11 @@ class atanh(InverseHyperbolicFunction):
     def _eval_rewrite_as_log(self, x, **kwargs):
         return (log(1 + x) - log(1 - x)) / 2
 
+    def _eval_rewrite_as_asinh(self, x, **kwargs):
+        f = sqrt(1/(x**2 - 1))
+        return (pi*x/(2*sqrt(-x**2)) -
+                sqrt(-x)*sqrt(1 - x**2)/sqrt(x)*f*asinh(f))
+
     def _eval_is_zero(self):
         if self.args[0].is_zero:
             return True
@@ -1537,6 +1611,13 @@ class acoth(InverseHyperbolicFunction):
     def _eval_rewrite_as_log(self, x, **kwargs):
         return (log(1 + 1/x) - log(1 - 1/x)) / 2
 
+    def _eval_rewrite_as_atanh(self, x, **kwargs):
+        return atanh(1/x)
+
+    def _eval_rewrite_as_asinh(self, x, **kwargs):
+        return (pi*I/2*(sqrt((x - 1)/x)*sqrt(x/(x - 1)) - sqrt(1 + 1/x)*sqrt(x/(x + 1))) +
+                x*sqrt(1/x**2)*asinh(sqrt(1/(x**2 - 1))))
+
     def inverse(self, argindex=1):
         """
         Returns the inverse of this function.
@@ -1617,6 +1698,8 @@ class asech(InverseHyperbolicFunction):
                 (-1 - sqrt(5)): 3*S.Pi / 5,
                 (sqrt(6) + sqrt(2)): 5*S.Pi / 12,
                 (-sqrt(6) - sqrt(2)): 7*S.Pi / 12,
+                S.ImaginaryUnit*S.Infinity: -S.Pi*S.ImaginaryUnit / 2,
+                S.ImaginaryUnit*S.NegativeInfinity: S.Pi*S.ImaginaryUnit / 2,
             }
 
     @classmethod
@@ -1676,6 +1759,20 @@ class asech(InverseHyperbolicFunction):
 
     def _eval_rewrite_as_log(self, arg, **kwargs):
         return log(1/arg + sqrt(1/arg - 1) * sqrt(1/arg + 1))
+
+    def _eval_rewrite_as_acosh(self, arg, **kwargs):
+        return acosh(1/arg)
+
+    def _eval_rewrite_as_asinh(self, arg, **kwargs):
+        return sqrt(1/arg - 1)/sqrt(1 - 1/arg)*(S.ImaginaryUnit*asinh(S.ImaginaryUnit/arg)
+                                                + S.Pi*S.Half)
+
+    def _eval_rewrite_as_atanh(self, x, **kwargs):
+        return (I*pi*(1 - sqrt(x)*sqrt(1/x) - I/2*sqrt(-x)/sqrt(x) - I/2*sqrt(x**2)/sqrt(-x**2))
+                + sqrt(1/(x + 1))*sqrt(x + 1)*atanh(sqrt(1 - x**2)))
+
+    def _eval_rewrite_as_acsch(self, x, **kwargs):
+        return sqrt(1/x - 1)/sqrt(1 - 1/x)*(pi/2 - I*acsch(I*x))
 
 
 class acsch(InverseHyperbolicFunction):
@@ -1767,6 +1864,9 @@ class acsch(InverseHyperbolicFunction):
         if arg is S.ComplexInfinity:
             return S.Zero
 
+        if arg.is_infinite:
+            return S.Zero
+
         if arg.is_zero:
             return S.ComplexInfinity
 
@@ -1781,6 +1881,19 @@ class acsch(InverseHyperbolicFunction):
 
     def _eval_rewrite_as_log(self, arg, **kwargs):
         return log(1/arg + sqrt(1/arg**2 + 1))
+
+    def _eval_rewrite_as_asinh(self, arg, **kwargs):
+        return asinh(1/arg)
+
+    def _eval_rewrite_as_acosh(self, arg, **kwargs):
+        return S.ImaginaryUnit*(sqrt(1 - S.ImaginaryUnit/arg)/sqrt(S.ImaginaryUnit/arg - 1)*
+                                acosh(S.ImaginaryUnit/arg) - S.Pi*S.Half)
+
+    def _eval_rewrite_as_atanh(self, arg, **kwargs):
+        arg2 = arg**2
+        arg2p1 = arg2 + 1
+        return sqrt(-arg2)/arg*(S.Pi*S.Half -
+                                sqrt(-arg2p1**2)/arg2p1*atanh(sqrt(arg2p1)))
 
     def _eval_is_zero(self):
         return self.args[0].is_infinite
--- a/sympy/functions/elementary/tests/test_hyperbolic.py
+++ b/sympy/functions/elementary/tests/test_hyperbolic.py
@@ -558,6 +558,10 @@ def test_asinh():
 def test_asinh_rewrite():
     x = Symbol('x')
     assert asinh(x).rewrite(log) == log(x + sqrt(x**2 + 1))
+    assert asinh(x).rewrite(atanh) == atanh(x/sqrt(1 + x**2))
+    assert asinh(x).rewrite(asin) == asinh(x)
+    assert asinh(x*(1 + I)).rewrite(asin) == -I*asin(I*x*(1+I))
+    assert asinh(x).rewrite(acos) == I*(-I*asinh(x) + pi/2) - I*pi/2
 
 
 def test_asinh_series():
@@ -637,6 +641,14 @@ def test_acosh():
 def test_acosh_rewrite():
     x = Symbol('x')
     assert acosh(x).rewrite(log) == log(x + sqrt(x - 1)*sqrt(x + 1))
+    assert acosh(x).rewrite(asin) == sqrt(x - 1)*(-asin(x) + pi/2)/sqrt(1 - x)
+    assert acosh(x).rewrite(asinh) == sqrt(x - 1)*(-asin(x) + pi/2)/sqrt(1 - x)
+    assert acosh(x).rewrite(atanh) == \
+        (sqrt(x - 1)*sqrt(x + 1)*atanh(sqrt(x**2 - 1)/x)/sqrt(x**2 - 1) +
+         pi*sqrt(x - 1)*(-x*sqrt(x**(-2)) + 1)/(2*sqrt(1 - x)))
+    x = Symbol('x', positive=True)
+    assert acosh(x).rewrite(atanh) == \
+        sqrt(x - 1)*sqrt(x + 1)*atanh(sqrt(x**2 - 1)/x)/sqrt(x**2 - 1)
 
 
 def test_acosh_series():
@@ -721,6 +733,10 @@ def test_asech_series():
 def test_asech_rewrite():
     x = Symbol('x')
     assert asech(x).rewrite(log) == log(1/x + sqrt(1/x - 1) * sqrt(1/x + 1))
+    assert asech(x).rewrite(acosh) == acosh(1/x)
+    assert asech(x).rewrite(asinh) == sqrt(-1 + 1/x)*(-asin(1/x) + pi/2)/sqrt(1 - 1/x)
+    assert asech(x).rewrite(atanh) == \
+        sqrt(x + 1)*sqrt(1/(x + 1))*atanh(sqrt(1 - x**2)) + I*pi*(-sqrt(x)*sqrt(1/x) + 1 - I*sqrt(x**2)/(2*sqrt(-x**2)) - I*sqrt(-x)/(2*sqrt(x)))
 
 
 def test_asech_fdiff():
@@ -796,6 +812,10 @@ def test_acsch_infinities():
 def test_acsch_rewrite():
     x = Symbol('x')
     assert acsch(x).rewrite(log) == log(1/x + sqrt(1/x**2 + 1))
+    assert acsch(x).rewrite(asinh) == asinh(1/x)
+    assert acsch(x).rewrite(atanh) == (sqrt(-x**2)*(-sqrt(-(x**2 + 1)**2)
+                                                    *atanh(sqrt(x**2 + 1))/(x**2 + 1)
+                                                    + pi/2)/x)
 
 
 def test_acsch_fdiff():
@@ -864,6 +884,8 @@ def test_atanh():
 def test_atanh_rewrite():
     x = Symbol('x')
     assert atanh(x).rewrite(log) == (log(1 + x) - log(1 - x)) / 2
+    assert atanh(x).rewrite(asinh) == \
+        pi*x/(2*sqrt(-x**2)) - sqrt(-x)*sqrt(1 - x**2)*sqrt(1/(x**2 - 1))*asinh(sqrt(1/(x**2 - 1)))/sqrt(x)
 
 
 def test_atanh_series():
@@ -920,6 +942,9 @@ def test_acoth():
 def test_acoth_rewrite():
     x = Symbol('x')
     assert acoth(x).rewrite(log) == (log(1 + 1/x) - log(1 - 1/x)) / 2
+    assert acoth(x).rewrite(atanh) == atanh(1/x)
+    assert acoth(x).rewrite(asinh) == \
+        x*sqrt(x**(-2))*asinh(sqrt(1/(x**2 - 1))) + I*pi*(sqrt((x - 1)/x)*sqrt(x/(x - 1)) - sqrt(x/(x + 1))*sqrt(1 + 1/x))/2
 
 
 def test_acoth_series():
@@ -956,8 +981,8 @@ def test_leading_term():
     for func in [sinh, tanh, asinh, atanh]:
         assert func(x).as_leading_term(x) == x
     for func in [sinh, cosh, tanh, coth, asinh, acosh, atanh, acoth]:
-        for arg in (1/x, S.Half):
-            eq = func(arg)
+        for ar in (1/x, S.Half):
+            eq = func(ar)
             assert eq.as_leading_term(x) == eq
     for func in [csch, sech]:
         eq = func(S.Half)
--- a/sympy/printing/codeprinter.py
+++ b/sympy/printing/codeprinter.py
@@ -69,6 +69,18 @@ class CodePrinter(StrPrinter):
     # may be supported
     # function_to_rewrite : (function_to_rewrite_to, iterable_with_other_functions_required)
     _rewriteable_functions = {
+            'cot': ('tan', []),
+            'csc': ('sin', []),
+            'sec': ('cos', []),
+            'acot': ('atan', []),
+            'acsc': ('asin', []),
+            'asec': ('acos', []),
+            'coth': ('exp', []),
+            'csch': ('exp', []),
+            'sech': ('exp', []),
+            'acoth': ('log', []),
+            'acsch': ('log', []),
+            'asech': ('log', []),
             'catalan': ('gamma', []),
             'fibonacci': ('sqrt', []),
             'lucas': ('sqrt', []),
--- a/sympy/utilities/tests/test_lambdify.py
+++ b/sympy/utilities/tests/test_lambdify.py
@@ -12,12 +12,13 @@ from sympy.core.singleton import S
 from sympy.core.symbol import (Dummy, symbols)
 from sympy.functions.combinatorial.factorials import (RisingFactorial, factorial)
 from sympy.functions.elementary.complexes import Abs
-from sympy.functions.elementary.exponential import exp
+from sympy.functions.elementary.exponential import exp, log
 from sympy.functions.elementary.hyperbolic import acosh
 from sympy.functions.elementary.integers import floor
 from sympy.functions.elementary.miscellaneous import (Max, Min, sqrt)
 from sympy.functions.elementary.piecewise import Piecewise
-from sympy.functions.elementary.trigonometric import (acos, cos, sin, sinc, tan)
+from sympy.functions.elementary.trigonometric import (acos, cos, cot, sin,
+                                                      sinc, tan)
 from sympy.functions.special.bessel import (besseli, besselj, besselk, bessely)
 from sympy.functions.special.beta_functions import (beta, betainc, betainc_regularized)
 from sympy.functions.special.delta_functions import (Heaviside)
@@ -1348,6 +1349,24 @@ def test_issue_22739():
     assert abs(f.subs(point) - F(*point.values())) <= 1e-10
 
 
+def test_issue_22992():
+    if not numpy:
+        skip("numpy not installed")
+
+    a, t = symbols('a t')
+    expr = a*(log(cot(t/2)) - cos(t))
+    F = lambdify([a, t], expr, 'numpy')
+
+    point = {a: 10, t: 2}
+
+    assert abs(expr.subs(point) - F(*point.values())) <= 1e-10
+
+    # Standard math
+    F = lambdify([a, t], expr)
+
+    assert abs(expr.subs(point) - F(*point.values())) <= 1e-10
+
+
 def test_issue_19764():
     if not numpy:
         skip("numpy not installed")
@@ -1394,6 +1413,7 @@ def test_beta_math():
 
     assert abs(beta(1.3, 2.3) - F(1.3, 2.3)) <= 1e-10
 
+
 def test_betainc_scipy():
     if not scipy:
         skip("scipy not installed")
@@ -1402,6 +1422,7 @@ def test_betainc_scipy():
     F = lambdify((w, x, y, z), f, modules='scipy')
 
     assert abs(betainc(1.4, 3.1, 0.1, 0.5) - F(1.4, 3.1, 0.1, 0.5)) <= 1e-10
+
 
 def test_betainc_regularized_scipy():
     if not scipy:
--- a/doc/src/tutorial/simplification.rst
+++ b/doc/src/tutorial/simplification.rst
@@ -623,11 +623,12 @@ another.  This works for any function in SymPy, not just special functions.
 To rewrite an expression in terms of a function, use
 ``expr.rewrite(function)``.  For example,
 
-    >>> tan(x).rewrite(sin)
-         2
-    2⋅sin (x)
-    ─────────
-     sin(2⋅x)
+    >>> tan(x).rewrite(cos)
+       ⎛    π⎞
+    cos⎜x - ─⎟
+       ⎝    2⎠
+    ──────────
+      cos(x)
     >>> factorial(x).rewrite(gamma)
     Γ(x + 1)
 
--- a/sympy/functions/elementary/tests/test_trigonometric.py
+++ b/sympy/functions/elementary/tests/test_trigonometric.py
@@ -1,6 +1,7 @@
 from sympy.calculus.accumulationbounds import AccumBounds
 from sympy.core.add import Add
 from sympy.core.function import (Lambda, diff)
+from sympy.core.mod import Mod
 from sympy.core.mul import Mul
 from sympy.core.numbers import (E, Float, I, Rational, nan, oo, pi, zoo)
 from sympy.core.power import Pow
@@ -192,7 +193,9 @@ def test_sin_series():
 def test_sin_rewrite():
     assert sin(x).rewrite(exp) == -I*(exp(I*x) - exp(-I*x))/2
     assert sin(x).rewrite(tan) == 2*tan(x/2)/(1 + tan(x/2)**2)
-    assert sin(x).rewrite(cot) == 2*cot(x/2)/(1 + cot(x/2)**2)
+    assert sin(x).rewrite(cot) == \
+        Piecewise((0, Eq(im(x), 0) & Eq(Mod(x, pi), 0)),
+                  (2*cot(x/2)/(cot(x/2)**2 + 1), True))
     assert sin(sinh(x)).rewrite(
         exp).subs(x, 3).n() == sin(x).rewrite(exp).subs(x, sinh(3)).n()
     assert sin(cosh(x)).rewrite(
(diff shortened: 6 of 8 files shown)


sympy/printing/codeprinter.py
```diff
@@ -69,6 +69,7 @@ class CodePrinter(StrPrinter):
     # may be supported
     # function_to_rewrite : (function_to_rewrite_to, iterable_with_other_functions_required)
     _rewriteable_functions = {
+            'cot': ('tan', []),
```
oscarbenjamin, line 72: Presumably this should also be done for sec, csc, coth, sech, csch as well.

oscarbenjamin: And all the inverses asec, acsc etc

sympy/utilities/tests/test_lambdify.py
```diff
@@ -1340,6 +1341,24 @@ def test_issue_22739():
...
+    F = lambdify([a, t], expr, 'numpy')
+
+    point = {a: 10, t: 2}
+
+    assert abs(expr.subs(point) - F(*point.values())) <= 1e-10
+
+    # Standard math
+    F = lambdify([a, t], expr)
+
+    assert abs(expr.subs(point) - F(*point.values())) <= 1e-10
```
ThePauliPrinciple, line 1359: Are there any choices of `point` that would deviate substantially between `cot` and `tan**-1` due to numerical implementation? I'd argue that this test would imply to users that this accuracy holds for any point, but that might not actually be the case.

oscarbenjamin: I wouldn't have thought that there would be a substantial deviation. Rounding error from division is generally minimal.

For inverse functions there could be issues with branch cuts and signed zeros (https://github.com/sympy/sympy/issues/21751)

oscargus: I guess one can probably find cases where it is better to evaluate the original function as opposed to a rewrite, but to some extent it is choice where to get the errors... And in most cases, I assume that the rewrites are OK enough and not worse that just evaluating a large symbolic argument to `cot` like `sqrt(sqrt(2) + sin(pi/7))` or so.

sympy/printing/codeprinter.py
oscargus: Good observation! I'll add those as well (but maybe not tests for all combinations etc...).

oscarbenjamin: I think it is worth testing all combinations. Some of them won't work and some of the rewrite rules seem to be wrong or broken e.g.:
```python
In [2]: sech(x).rewrite(cosh)
Out[2]: sech(x)

In [3]: sech(x).rewrite(sinh)
Out[3]: 
   1   
───────
cosh(x)

In [4]: csch(x).rewrite(cosh)
Out[4]: 
   1   
───────
sinh(x)

In [5]: csch(x).rewrite(sinh)
Out[5]: csch(x)
```
Also if any rewrite rule is missing then `lambdify` goes into infinite recursion so that would happen for `asech` here:
```python
In [9]: asec(x).rewrite(acos)
Out[9]: 
    ⎛1⎞
acos⎜─⎟
    ⎝x⎠

In [10]: asech(x).rewrite(acosh)
Out[10]: asech(x)
```

oscargus: OK! I'll update the current one then. I found that the `a*h` could not be rewritten, but not the others.

sympy/printing/codeprinter.py
```diff
@@ -69,6 +69,18 @@ class CodePrinter(StrPrinter):
...
+            'csc': ('sin', []),
+            'sec': ('cos', []),
+            'acot': ('atan', []),
+            'acsc': ('asin', []),
+            'asec': ('acos', []),
+            'coth': ('tanh', []),
+            'csch': ('sinh', []),
+            'sech': ('cosh', []),
```
oscarbenjamin, line 80: This one doesn't work
```
In [1]: lambdify(x, sech(x))(1)
RecursionError
```
It would work if changed to `sinh` but actually I think the proper rewrite rule should be added to sech instead.
```
In [2]: sech(x).rewrite(cosh)
Out[2]: sech(x)

In [3]: sech(x).rewrite(sinh)
Out[3]: 
   1   
───────
cosh(x)
```

sympy/printing/codeprinter.py
oscargus: Some are simply not supported, while others are getting rewritten as part of automatic evaluation. For example:
```
In [3]: sech(x).rewrite(sinh)
Out[3]: 
   1   
───────
cosh(x)
```

oscargus:
This currently is enough (and a bit more) to support code generation for all the (a)cot/sec/csc(h) variants, but for each, it is in general possible to rewrite it to almost any other form (including the sin/cos/tan ones), so typically 12 variants (plus log and meijerg etc) for each (> 144 in total). Hence, this is quite a bit of work. 

I am currently happy for now. While it is of course possible to add  all, I am not sure if it is worth it. Some of them are automatically evaluated back and others have quite massive expressions for the general case. I cannot even imagine adding tests for all the cases. Maybe the way forward is to add an issue for adding the remaining one. It is pretty straightforward, although tedious, and it is easy to show by an example what code and test to add (although, evaluation can make it a bit challenging to figure out a good test case).

sympy/printing/codeprinter.py
oscarbenjamin: There should be a `sech._eval_rewrite_as_cosh` method.

oscarbenjamin: Oh, I see you've added it now

sympy/printing/codeprinter.py
oscargus: Done!

sympy/functions/elementary/tests/test_hyperbolic.py
```diff
@@ -558,6 +558,10 @@ def test_asinh():
 def test_asinh_rewrite():
     x = Symbol('x')
     assert asinh(x).rewrite(log) == log(x + sqrt(x**2 + 1))
+    assert asinh(x).rewrite(atanh) == atanh(x/sqrt(1 + x**2))
+    assert asinh(x).rewrite(asin) == asinh(x)
+    assert asinh(x*(1 + I)).rewrite(asin) == -I*asin(I*x*(1+I))
+    assert asinh(x).rewrite(acos) == I*(-I*asinh(x) + pi/2) - I*pi/2
```
smichr, line 564: Isn't this odd: no `acos`, just a rewriting in terms of original function?

smichr: And when expanded, it is just the original.

sympy/functions/elementary/tests/test_hyperbolic.py
```diff
@@ -637,6 +641,8 @@ def test_acosh():
 def test_acosh_rewrite():
     x = Symbol('x')
     assert acosh(x).rewrite(log) == log(x + sqrt(x - 1)*sqrt(x + 1))
+    assert acosh(x).rewrite(asinh) == Abs(asinh(sqrt(x**2 - 1)))
```
smichr, line 644: At `x=0` these are not the same.

sympy/functions/elementary/tests/test_hyperbolic.py
```diff
@@ -637,6 +641,8 @@ def test_acosh():
...
+    assert acosh(x).rewrite(atanh) == Abs(atanh(sqrt(x**2 - 1)/x))
```
smichr, line 645: These are not the same at 1.

sympy/functions/elementary/tests/test_hyperbolic.py
```diff
@@ -721,6 +727,9 @@ def test_asech_series():
 def test_asech_rewrite():
     x = Symbol('x')
     assert asech(x).rewrite(log) == log(1/x + sqrt(1/x - 1) * sqrt(1/x + 1))
+    assert asech(x).rewrite(acosh) == acosh(1/x)
+    assert asech(x).rewrite(asinh) == Abs(asinh(sqrt(1/x**2 - 1)))
+    assert asech(x).rewrite(atanh) == Abs(atanh(x*sqrt(1/x**2 - 1)))
```
smichr, line 732: Issues with branch, again, when `x > 1` -- do these need to be Piecewise?

sympy/functions/elementary/tests/test_hyperbolic.py
oscarbenjamin: The problem is automatic evaluation:
```python
In [2]: acos(I * x)
Out[2]: 
              π
-ⅈ⋅asinh(x) + ─
              2
```

sympy/functions/elementary/tests/test_hyperbolic.py
oscarbenjamin: Maybe some of these rewrites that aren't needed for lambdify should be removed so that the main issue can be fixed.

oscargus:
I've updated the rewrites. In some cases I used a Piecewise, for others, I went for the most general expressions available. 

I'm a bit confused about
``` 
    def _eval_rewrite_as_asinh(self, x, **kwargs):
        return sqrt(x - 1)/sqrt(1 - x)*(pi/2 + S.ImaginaryUnit*asinh(S.ImaginaryUnit*x))
```
as Sympy will evaluate `asinh(S.ImaginaryUnit*x)` as `S.ImaginaryUnit*asinh(x)`. Primarily, I wonder if the evaluation is really correct? (If so, why did Wolfram not include that in the general rewrite, would be much easier to use ` sqrt(x - 1)/sqrt(1 - x)*(pi/2 - asinh(x))`, but that is not given...)

oscarbenjamin:
I tested all combinations like this:
```python
from sympy import *

x = symbols('x')

funcs = [cot , csc , sec , acot , acsc , asec,
         coth, csch, sech, acoth, acsch, asech]

points = [
    0, 1, -1, I, -I, 1+I, 1-I, -1+I, -1-I,
]

for f1 in funcs:
    for f2 in funcs:
        expr1 = f1(x)
        expr2 = expr1.rewrite(f2)
        for p in points:
            if expr1.subs(x, p).n() != expr2.subs(x, p).n():
                print()
                print(f1, f2, p)
                print(expr1, ' --> ', expr2)
                print(expr1.subs(x, p), expr2.subs(x, p))
                print(expr1.subs(x, p).n(5), expr2.subs(x, p).n(5))
```
There are a bunch of cases that don't work out e.g.:
```python
In [32]: sec(0)
Out[32]: 1

In [33]: sec(x).rewrite(cot).subs(x, 0)
Out[33]: nan
```
The full output is
```python
csc cot 0
csc(x)  -->  (cot(x/2)**2 + 1)/(2*cot(x/2))
zoo nan
zoo nan

sec cot 0
sec(x)  -->  (cot(x/2)**2 + 1)/(cot(x/2)**2 - 1)
1 nan
1.0000 nan

acot acsc 0
acot(x)  -->  x*(-acsc(sqrt((x**2 + 1)/x**2)) + pi/2)*sqrt(x**(-2))
pi/2 nan
1.5708 nan

acot acsc I
acot(x)  -->  x*(-acsc(sqrt((x**2 + 1)/x**2)) + pi/2)*sqrt(x**(-2))
-oo*I zoo
-oo*I zoo

acot acsc -I
acot(x)  -->  x*(-acsc(sqrt((x**2 + 1)/x**2)) + pi/2)*sqrt(x**(-2))
oo*I zoo
oo*I zoo

acot acsc -1 + I
acot(x)  -->  x*(-acsc(sqrt((x**2 + 1)/x**2)) + pi/2)*sqrt(x**(-2))
-acot(1 - I) (-1 - I)*(-1 + I)*(pi/2 - acsc(sqrt((1 + (-1 + I)**2)/(-1 + I)**2)))/2
-0.55357 - 0.40236*I 0.55357 + 0.40236*I

acot acsc -1 - I
acot(x)  -->  x*(-acsc(sqrt((x**2 + 1)/x**2)) + pi/2)*sqrt(x**(-2))
-acot(1 + I) (-1 - I)*(-1 + I)*(pi/2 - acsc(sqrt((1 + (-1 - I)**2)/(-1 - I)**2)))/2
-0.55357 + 0.40236*I 0.55357 - 0.40236*I

acot asec 0
acot(x)  -->  x*sqrt(x**(-2))*asec(sqrt((x**2 + 1)/x**2))
pi/2 nan
1.5708 nan

acot asec I
acot(x)  -->  x*sqrt(x**(-2))*asec(sqrt((x**2 + 1)/x**2))
-oo*I zoo
-oo*I zoo

acot asec -I
acot(x)  -->  x*sqrt(x**(-2))*asec(sqrt((x**2 + 1)/x**2))
oo*I zoo
oo*I zoo

acot asec -1 + I
acot(x)  -->  x*sqrt(x**(-2))*asec(sqrt((x**2 + 1)/x**2))
-acot(1 - I) (-1 - I)*(-1 + I)*asec(sqrt((1 + (-1 + I)**2)/(-1 + I)**2))/2
-0.55357 - 0.40236*I 0.55357 + 0.40236*I

acot asec -1 - I
acot(x)  -->  x*sqrt(x**(-2))*asec(sqrt((x**2 + 1)/x**2))
-acot(1 + I) (-1 - I)*(-1 + I)*asec(sqrt((1 + (-1 - I)**2)/(-1 - I)**2))/2
-0.55357 + 0.40236*I 0.55357 - 0.40236*I

acsc acot 0
acsc(x)  -->  (-acot(1/sqrt(x**2 - 1)) + pi/2)*sqrt(x**2)/x
zoo nan
zoo nan

acsc acot I
acsc(x)  -->  (-acot(1/sqrt(x**2 - 1)) + pi/2)*sqrt(x**2)/x
acsc(I) pi/2 - I*acoth(sqrt(2)/2)
-0.88137*I 0.e-10 - 0.88137*I

acsc acot -I
acsc(x)  -->  (-acot(1/sqrt(x**2 - 1)) + pi/2)*sqrt(x**2)/x
-acsc(I) -pi/2 + I*acoth(sqrt(2)/2)
0.88137*I -0.e-10 + 0.88137*I

acsc acot -1 + I
acsc(x)  -->  (-acot(1/sqrt(x**2 - 1)) + pi/2)*sqrt(x**2)/x
-acsc(1 - I) (-1 - I)*(-1 + I)*(pi/2 - acot(1/sqrt(-1 + (-1 + I)**2)))/2
-0.45228 - 0.53064*I 0.45228 + 0.53064*I

acsc acot -1 - I
acsc(x)  -->  (-acot(1/sqrt(x**2 - 1)) + pi/2)*sqrt(x**2)/x
-acsc(1 + I) (-1 - I)*(-1 + I)*(pi/2 - acot(1/sqrt(-1 + (-1 - I)**2)))/2
-0.45228 + 0.53064*I 0.45228 - 0.53064*I

acsc asec I
acsc(x)  -->  -asec(x) + pi/2
acsc(I) pi/2 - asec(I)
-0.88137*I 0.e-10 - 0.88137*I

acsc asec -I
acsc(x)  -->  -asec(x) + pi/2
-acsc(I) pi/2 - asec(-I)
0.88137*I 0.e-10 + 0.88137*I

asec acot 0
asec(x)  -->  (2*acot(x - sqrt(x**2 - 1)) - pi/2)*sqrt(x**2)/x
zoo nan
zoo nan

asec acot -1 + I
asec(x)  -->  (2*acot(x - sqrt(x**2 - 1)) - pi/2)*sqrt(x**2)/x
asec(-1 + I) (-1 - I)*(-1 + I)*(-pi/2 - 2*acot(1 + sqrt(-1 + (-1 + I)**2) - I))/2
2.0231 + 0.53064*I -2.0231 - 0.53064*I

asec acot -1 - I
asec(x)  -->  (2*acot(x - sqrt(x**2 - 1)) - pi/2)*sqrt(x**2)/x
asec(-1 - I) (-1 - I)*(-1 + I)*(-pi/2 - 2*acot(1 + I + sqrt(-1 + (-1 - I)**2)))/2
2.0231 - 0.53064*I -2.0231 + 0.53064*I

csch coth 0
csch(x)  -->  (coth(x/2)**2 - 1)/(2*coth(x/2))
zoo nan
zoo nan

sech coth 0
sech(x)  -->  (coth(x/2)**2 - 1)/(coth(x/2)**2 + 1)
1 nan
1.0000 nan

asech acsch 0
asech(x)  -->  sqrt(-1 + 1/x)*(-I*acsch(I*x) + pi/2)/sqrt(1 - 1/x)
oo nan
oo nan

asech acsch 1
asech(x)  -->  sqrt(-1 + 1/x)*(-I*acsch(I*x) + pi/2)/sqrt(1 - 1/x)
0 nan
0 nan
```
I don't think any of these are rewrites that are needed for lambdify though. Many of the examples fail on master as well.

oscargus:
Well, I have only added rewrite equations for the hyperbolic functions, so cannot really say anything about the trig-rewrites. (Some of those seems to be fundamentally incorrect.)

For the hyperbolic functions, I assume that the equations (given at Wolframs function site) do assume certain things, such as 1/0 = oo (rather than NaN), so the limit will be correct but as Sympy has made the choice to do that, the results will be a bit messed up. But it is of course possible to return a Piecewise for those cases.

oscargus:
There are so many rewrites that are incorrect in the existing code that it is not really possible to fix it in this PR. I mean, not even this holds: `tan(x).rewrite('sincos')`

I'll push the latest version that fixes the introduced issues pointed out in your test and also introduces some support for +/-I*oo.

oscargus:
Here is an extended test:

```
from sympy import *

x = symbols('x')

funcs = [cos, sin, tan, cot , csc , sec , acos, asin, atan, acot , acsc , asec,
         cosh, sinh, tanh, coth, csch, sech, acosh, asinh, atanh, acoth, acsch, asech]

points = [
    0, 1, -1, I, -I, 1+I, 1-I, -1+I, -1-I, oo, -oo, I*oo, -I*oo, zoo
]

for f1 in funcs:
    for f2 in funcs:
        expr1 = f1(x)
        expr2 = expr1.rewrite(f2)
        for p in points:
            if expr1.subs(x, p).n() != expr2.subs(x, p).n():
                print()
                print(f1, f2, p)
                print(expr1, ' --> ', expr2)
                print(expr1.subs(x, p), expr2.subs(x, p))
                print(expr1.subs(x, p).n(5), expr2.subs(x, p).n(5))
```

oscarbenjamin:
> There are so many rewrites that are incorrect in the existing code that it is not really possible to fix it in this PR

I wasn't suggesting that they all needed to be fixed here.

oscarbenjamin:
> Here is an extended test:

Okay, there's a lot of fails (on master)...

oscargus:
I think all failing tests of your original test are not caused by my addition, even the ones for hyperbolic functions. Some of the basic hyperbolic rewrites are just using the reciprocal of the argument or function, which then of course fails for 0 inputs.

asmeurer:
> cot, sec, csc and related inverse and hyperbolic functions are automatically rewritten if they cannot be printed.

I hope this really just means the code printers. Things like pprint() and latex() should not be doing expression rewriting. I'm a little hesitant about having it in the code printers too. It makes sense for lambdify to do it, but I wonder if that should instead be done as a separate step from the printing.

sympy/printing/codeprinter.py
```diff
@@ -69,6 +69,18 @@ class CodePrinter(StrPrinter):
     # may be supported
     # function_to_rewrite : (function_to_rewrite_to, iterable_with_other_functions_required)
     _rewriteable_functions = {
```
asmeurer, line 71: I guess I didn't realize this is already a thing. So given that this is already here, we can keep it.

oscargus: Not sure about the mpmath printer. I do not think so as all(?) functions in SymPy already have mpmath-support(?).

The point was to support code generation for languages that didn't have support for some functions, but could be rewritten to have support. And to unify that so that one do not have to copy the same code to every printer to expand it. I *think* that it is probably only now that it has become possibly  problematic. Most (all?) earlier rewrites shouldn't have this problem.

These are the current automatic rewrites btw.
https://github.com/sympy/sympy/blob/0bc2587aef09ab150bd408b7cd9e08baf7b5fd9c/sympy/printing/codeprinter.py#L71-L94

And somehow, the problem is not really in the printing, but in the rewrites themselves. So if it is done automatically or a user have to do it to be able to print is not fundamentally different (from some perspective).

Finally, for rewriting to happen someone should add the rule to the list above. So the while it can happen that a new function is added and rather than adding support for latex and pprint, the rewrite support is added, it seems quite unlikely. Especially since this is not really a well known functionality.

asmeurer: Let's just leave this as it is for now. I didn't realize this was already happening. I think this should only happen if the function can't be printed at all. I brought up the broader question of how to do rewrites in the printers here https://github.com/sympy/sympy/issues/22277#issuecomment-1058481071, although maybe it should be a separate issue.

oscarbenjamin: It looks like `mpmath` has a bunch of those functions e.g. factorial, erfc, li, ff, sinc, ...

Maybe the rewrites should be put into groups in the CodePrinter class so that subclasses can choose which groups to use e.g.:
```python
class CodePrinter:
    _basic_rewrites = {
        'sinc': 'sin',
        ...
    }
    _reciprocal_trig_rewrites = {
        'sec': 'cos',
        ...
    }

class NumPyPrinter(CodePrinter):
    _rewrites = {
        **CodePrinter._basic_rewrites,
        **CodePrinter._reciprocal_trig_rewrites,
    }
```

oscargus: > I think this should only happen if the function can't be printed at all.

This is exactly how it is. These are used as a final resort:
https://github.com/sympy/sympy/blob/cd738be3bdde5f8a0e53bf71b4420146acff392a/sympy/printing/codeprinter.py#L433-L437

asmeurer: Honestly these class level variables that are so commonly used in the printer classes are not great for subclassing. You either get all or nothing. And furthermore, since they are mutable containers, you can get a subclass modifying them and changing all other subclasses:

```py
>>> class A:
...     x = {
...     'a': 1,
...     }
>>> class B(A):
...     def __init__(self):
...         self.x['b'] = 2
>>> A().x
{'a': 1}
>>> B()
<__main__.B object at 0x10b3fff70>
>>> A().x
{'a': 1, 'b': 2}
```

For this specific case, though, as long as the rewrites are ignored when the printer actually knows how to handle a given function, then it shouldn't be a problem to just have one big list.

oscargus: One can consider making it immutable (`collections.namedtuple()` may work from a quick googling). AFAIK there is no code (added by me) that modifies it.

oscargus:
Is there anything here that still needs to be done (except sorting out the conflicts?). I think this would be nice to have in 1.11.

To summarize: these automatic rewrites happens when a particular code printer/language does not support a given function. Hence, we can either rewrite those to something that is supported (subject to potential numerical errors) or just say that we cannot print it. This PR encourages the first approach.

oscargus:
Restarting CI.

oscarbenjamin:
I think this looks good. Let's fix the conflicts and get it in.

We should open a new issue to consider all of the rewrites and evaluation problems highlighted above (https://github.com/sympy/sympy/pull/22993#issuecomment-1053624700). In fact I'm more concerned about what happens when we don't just test special values:
```python
import numpy as np
from sympy import *

x = symbols('x')

funcs = [cos, sin, tan, cot , csc , sec , acos, asin, atan, acot , acsc , asec,
         cosh, sinh, tanh, coth, csch, sech, acosh, asinh, atanh, acoth, acsch, asech]

numtests = 10
std = 10
normal = np.random.randn
points_real = std*normal(numtests)
points_complex = std*(normal(numtests) + normal(numtests)*1j)
points = list(points_real) + list(points_complex)

def isclose(n1, n2):
    return abs(n1 - n2)/(abs(n1) + abs(n2)) < 1e-5

bad = 0
total = 0
for f1 in funcs:
    for f2 in funcs:
        total += 1
        expr1 = f1(x)
        expr2 = expr1.rewrite(f2)
        fine = True
        for p in points:
            if not isclose(expr1.subs(x, p).n(), expr2.subs(x, p).n()):
                print()
                print(f1, f2, p)
                print(expr1, ' --> ', expr2)
                print(expr1.subs(x, p), expr2.subs(x, p))
                print(expr1.subs(x, p).n(5), expr2.subs(x, p).n(5))
                fine = False
        if not fine:
            bad += 1

print(bad, 'bad rewrites out of', total)
```
This gives a lot of output which represents potential bugs in evalf, evaluation, rewrites etc. Running that on master it gives a lot of output and ends with
```
21 bad rewrites out of 576
```
Testing with this PR I get
```
16 bad rewrites out of 576
```
That's definitely something that needs investigation and shouldn't be forgotten about.

sympy/functions/elementary/hyperbolic.py
```diff
@@ -1702,6 +1750,25 @@ def inverse(self, argindex=1):
...
+
+    def _eval_rewrite_as_asinh(self, arg, **kwargs):
+        return sqrt(1/arg - 1)/sqrt(1 - 1/arg)*(S.ImaginaryUnit*asinh(S.ImaginaryUnit/arg)
+                                                + S.Pi*S.Half)
+
+    def _eval_rewrite_as_atanh(self, arg, **kwargs):
+        return (S.Pi*S.ImaginaryUnit*(1 - sqrt(arg)*sqrt(1/arg) -
+                (S.ImaginaryUnit*sqrt(-arg))/(2*sqrt(arg)) -
+                (S.ImaginaryUnit*sqrt(arg**2))/(2*sqrt(-arg**2)))
+                + sqrt(arg + 1)*sqrt(1/(arg + 1))*atanh(sqrt(1 - arg**2)))
```
oscarbenjamin, line 1764: This doesn't always work:
```python
In [2]: f = asech(x)

In [3]: f.subs(x, -1-I).n()
Out[3]: 0.530637530952518 + 2.02307477394609⋅ⅈ

In [4]: f.rewrite(atanh).subs(x, -1-I).n()
Out[4]: 0.530637530952518 - 1.11851787964371⋅ⅈ
```

oscarbenjamin: On master this works because there is no rewrite:
```python
In [1]: f = asech(x)

In [2]: f.subs(x, -1-I).n()
Out[2]: 0.530637530952518 + 2.02307477394609⋅ⅈ

In [3]: f.rewrite(atanh).subs(x, -1-I).n()
Out[3]: 0.530637530952518 + 2.02307477394609⋅ⅈ
```

sympy/functions/elementary/hyperbolic.py
```diff
@@ -1702,6 +1750,25 @@ def inverse(self, argindex=1):
     def _eval_rewrite_as_log(self, arg, **kwargs):
         return log(1/arg + sqrt(1/arg - 1) * sqrt(1/arg + 1))
 
+    def _eval_rewrite_as_acosh(self, arg, **kwargs):
+        return sqrt(1/arg - 1)/sqrt(1 - 1/arg)*acosh(1/arg)
```
oscarbenjamin, line 1754: This doesn't always work:
```python
In [1]: f = asech(x)

In [2]: f.subs(x, -1+I).n()
Out[2]: 0.530637530952518 - 2.02307477394609⋅ⅈ

In [3]: f.rewrite(acosh).subs(x, -1+I).n()
Out[3]: -2.02307477394609 - 0.530637530952518⋅ⅈ
```
(a regression compared to master)

sympy/functions/elementary/hyperbolic.py
```diff
@@ -1359,6 +1380,14 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
     def _eval_rewrite_as_log(self, x, **kwargs):
         return log(x + sqrt(x + 1) * sqrt(x - 1))
 
+    def _eval_rewrite_as_asinh(self, x, **kwargs):
+        return sqrt(x - 1)/sqrt(1 - x)*(pi/2 +
+                                        S.ImaginaryUnit*asinh(S.ImaginaryUnit*x))
+
+    def _eval_rewrite_as_atanh(self, x, **kwargs):
+        return (pi/2*sqrt(x - 1)/sqrt(1-x)*(1 - x * sqrt(1/x**2)) +
+                sqrt(x - 1)*sqrt(x + 1)/sqrt(x**2 - 1)*atanh(sqrt(x**2 - 1)/x))
```
oscarbenjamin, line 1389: This doesn't always work:
```python
In [7]: f = acosh(x)

In [8]: f.subs(x, -1+I).n()
Out[8]: 1.06127506190504 + 2.23703575928741⋅ⅈ

In [9]: f.rewrite(atanh).subs(x, -1+I).n()
Out[9]: 1.06127506190504 - 0.904556894302381⋅ⅈ
```
(a regression compared to master)

oscarbenjamin:
This one is a bug on master:
```python
In [1]: f = asec(x)

In [2]: f.subs(x, -1+I).n()
Out[2]: 2.02307477394609 + 0.530637530952518⋅ⅈ

In [3]: f.rewrite(acot).subs(x, -1+I).n()
Out[3]: -2.02307477394609 - 0.530637530952518⋅ⅈ
```

oscarbenjamin:
Also a bug on master:
```python
In [6]: f = asec(x)

In [7]: f.subs(x, -1+I).n()
Out[7]: 2.02307477394609 + 0.530637530952518⋅ⅈ

In [8]: f.rewrite(atan).subs(x, -1+I).n()
Out[8]: -2.02307477394609 - 0.530637530952518⋅ⅈ
```

oscarbenjamin:
Also a bug on master:
```python
In [9]: f = acsc(x)

In [10]: f.rewrite(acot).subs(x, -1+I).n()
Out[10]: 0.452278447151191 + 0.530637530952518⋅ⅈ

In [11]: f.subs(x, -1+I).n()
Out[11]: -0.452278447151191 - 0.530637530952518⋅ⅈ
```

oscarbenjamin:
Also a bug on master:
```python
In [1]: f = acsc(x)

In [2]: f.subs(x, -1+I).n()
Out[2]: -0.452278447151191 - 0.530637530952518⋅ⅈ

In [3]: f.rewrite(atan).subs(x, -1+I).n()
Out[3]: 0.452278447151191 + 0.530637530952518⋅ⅈ
```

oscarbenjamin:
Also a bug on master:
```python
In [1]: f = acot(x)

In [2]: f.subs(x, -1+I).n()
Out[2]: -0.553574358897045 - 0.402359478108525⋅ⅈ

In [3]: f.rewrite(acsc).subs(x, -1+I).n()
Out[3]: 0.553574358897045 + 0.402359478108525⋅ⅈ
```

oscarbenjamin:
A bug on master:
```python
In [1]: f = acot(x)

In [2]: f.subs(x, -1+I).n()
Out[2]: -0.553574358897045 - 0.402359478108525⋅ⅈ

In [3]: f.rewrite(asin).subs(x, -1+I).n()
Out[3]: 0.553574358897045 + 0.402359478108525⋅ⅈ

```

oscarbenjamin:
Bug on master:
```python
In [5]: f = acot(x)

In [6]: f.subs(x, -1+I).n()
Out[6]: -0.553574358897045 - 0.402359478108525⋅ⅈ

In [7]: f.rewrite(acos).subs(x, -1+I).n()
Out[7]: 0.553574358897045 + 0.402359478108525⋅ⅈ
```

oscarbenjamin:
A bug on master:
```python
In [8]: f = atan(x)

In [9]: f.subs(x, -1+I).n()
Out[9]: -1.01722196789785 + 0.402359478108525⋅ⅈ

In [10]: f.rewrite(asec).subs(x, -1+I).n()
Out[10]: 1.01722196789785 - 0.402359478108525⋅ⅈ
```

oscarbenjamin:
A bug on master:
```python
In [11]: f = atan(x)

In [12]: f.subs(x, -1+I).n()
Out[12]: -1.01722196789785 + 0.402359478108525⋅ⅈ

In [13]: f.rewrite(acsc).subs(x, -1+I).n()
Out[13]: 1.01722196789785 - 0.402359478108525⋅ⅈ
```

oscarbenjamin:
A bug on master:
```python
In [14]: f = atan(x)

In [15]: f.subs(x, -1+I).n()
Out[15]: -1.01722196789785 + 0.402359478108525⋅ⅈ

In [16]: f.rewrite(asin).subs(x, -1+I).n()
Out[16]: 1.01722196789785 - 0.402359478108525⋅ⅈ
```

oscarbenjamin:
A bug on master:
```python
In [17]: f = atan(x)

In [18]: f.subs(x, -1+I).n()
Out[18]: -1.01722196789785 + 0.402359478108525⋅ⅈ

In [19]: f.rewrite(acos).subs(x, -1+I).n()
Out[19]: 1.01722196789785 - 0.402359478108525⋅ⅈ
```

oscarbenjamin:
A bug on master:
```python
In [20]: f = acos(x)

In [21]: f.subs(x, -1+I).n()
Out[21]: 2.23703575928741 - 1.06127506190504⋅ⅈ

In [22]: f.rewrite(atan).subs(x, -1+I).n()
Out[22]: -0.904556894302381 - 1.06127506190504⋅ⅈ
```

oscarbenjamin:
There are 3 cases (see above) where the rewrite introduced in this PR is incorrect or at least inconsistent with mpmath. (There are other cases where the same applies on master)

oscarbenjamin:
> There are 3 cases (see above) where the rewrite introduced in this PR is incorrect

Only the `asech -> acosh` rewrite is actually used by lambdify in this PR.

sympy/functions/elementary/hyperbolic.py
oscarbenjamin: I think that `return acosh(1/arg)` is good here.

sympy/functions/elementary/hyperbolic.py
oscargus: OK! I have if nothing else double-checked that the equation is the same as https://functions.wolfram.com/ElementaryFunctions/ArcSech/27/02/10/01/01/0005/

sympy/functions/elementary/hyperbolic.py
oscargus: Yes, I'll change to that. I cannot even find the earlier rewrite anymore...

sympy/functions/elementary/hyperbolic.py
oscargus: Same as this https://functions.wolfram.com/ElementaryFunctions/ArcCosh/27/02/09/01/01/0006/

oscargus:
That particular point seems a bit problematic:
```python
from sympy import *
p = -1 + I
x = Symbol('x')
e = sqrt(x**2)/x
print(e.subs(x, p).expand())

p2 = p**2
p2 = p2.expand()
print((sqrt(p2)/p).expand())
```

So the results are different depending on the order.

oscarbenjamin:
> That particular point (and probably more different points)

Note that I identified these problematic rewrites just by choosing random (normally distributed) real and complex numbers.

oscargus:
Yes, I realize that. But I guess that if it matters which order an expression like `sqrt(x**2)/x` evaluates and most rewrites depends in this, it is "a bit tricky" to get it correct...

sympy/functions/elementary/hyperbolic.py
oscarbenjamin: Both DLMF and mpmath define `asech(x)` as `acosh(1/x)`:
https://dlmf.nist.gov/4.37
https://mpmath.org/doc/current/functions/hyperbolic.html#asech

Their definitions of `atanh` don't seem to be equivalent though. mpmath says:
```
atanh(x) = 1/2 (log(1+x) - log(1-x)
```
DLMF says that away from the branch cuts
```
atanh(x) = 1/2 log((1+x)/(1-x))
```
These are not equivalent for most complex numbers:
```python
In [26]: atanh_mpmath = (log(1 + x) - log(1 - x))/2

In [27]: atanh_dlmf = log((1 + x)/log(1 - x))/2

In [28]: atanh_mpmath.subs(x, -1 + I)
Out[28]: 
  log(2 - ⅈ)   ⅈ⋅π
- ────────── + ───
      2         4 

In [29]: atanh_dlmf.subs(x, -1 + I)
Out[29]: 
   ⎛    ⅈ     ⎞
log⎜──────────⎟
   ⎝log(2 - ⅈ)⎠
───────────────
       2       

In [30]: atanh_mpmath.subs(x, -1 + I).n()
Out[30]: -0.402359478108525 + 1.01722196789785⋅ⅈ

In [31]: atanh_dlmf.subs(x, -1 + I).n()
Out[31]: 0.036967944745527 + 1.04675131315153⋅ⅈ
```
I'm also not sure that the definitions of `acosh` are equivalent either.

oscarbenjamin:
> it is "a bit tricky" to get it correct...

I agree. My quick fix suggestion is to comment out the two new problematic rewrites:
```
acosh -> atanh
asech -> atanh
```

oscargus:
The problem became really clear here:

``` python
        return atan(sqrt(1 - x**2)/x) + Piecewise((S.Zero, And(arg_f(x) < pi/2, -pi/2 <= arg_f(x))),
                                                  (pi, True))
        return atan(sqrt(1 - x**2)/x) + (pi/2)*(1 - x*sqrt(1/x**2))
```
as it is pretty clear that the second rewrite is trying to either add or subtract 1 depending on the argument.

So I would almost claim that it is the sqrt evaluation that is really the problem here.

Regarding automatic rewrites, by rewriting unsupported hyperbolics to exp or log, I am quite sure that it will work as the branch handling is much better there.

oscarbenjamin:
> So I would almost claim that it is the sqrt evaluation that is really the problem here.

What exactly do you mean by evaluation here? Are you talking about automatic symbolic evaluation (simplification) or numerical evaluation through mpmath or scipy etc.?

oscargus:
> What exactly do you mean by evaluation here?

Good question! Did you check out the small code example above where `sqrt(x**2)/x` evaluates to either 1 or -1 depending on how it is computed? The fact that this is computed differently (in the SymPy evaluation/expand/subs) is really the reason why these expressions do evaluate to something different.

oscarbenjamin:
Okay I see what you mean. Basically it's:
```python
In [39]: p = -1 + I

In [40]: sqrt(p**2)
Out[40]: -1 + ⅈ

In [41]: sqrt(expand(p**2))
Out[41]: 1 - ⅈ
```
The correct result here should be `1 - I` because that's the principal root.

oscarbenjamin:
> The correct result here should be `1 - I` because that's the principal root.

This bug has been around for a long time...

oscargus:
OK! I didn't recall that. Anyway, that is probably the source of the incorrect rewrites.

I have pushed a change where those two problematic ones are rewritten as `Piecewise`. Now, there are 10 bad rewrites left. Although more tests are required (there may be some missing singular values here as well)

sympy/functions/elementary/hyperbolic.py
oscarbenjamin: Okay, so the problem is `subs` rather than `rewrite`:
```python
In [1]: f = asech(x)

In [2]: f.evalf(subs={x:-1-I})
Out[2]: 0.530637530952518 + 2.02307477394609⋅ⅈ

In [3]: f.rewrite(atanh).evalf(subs={x:-1-I})
Out[3]: 0.530637530952518 + 2.02307477394609⋅ⅈ
```

sympy/functions/elementary/hyperbolic.py
oscarbenjamin: Also here the problem is `subs`:
```python
In [8]: f = acosh(x)

In [9]: f.evalf(subs={x:-1-I})
Out[9]: 1.06127506190504 - 2.23703575928741⋅ⅈ

In [10]: f.rewrite(atanh).evalf(subs={x:-1-I})
Out[10]: 1.06127506190504 - 2.23703575928741⋅ⅈ
```

oscarbenjamin:
Okay, I've updated the test script to use `evalf` with `subs=` rather than using `subs`:
```python
import numpy as np
from sympy import *

x = symbols('x')

funcs = [cos, sin, tan, cot , csc , sec , acos, asin, atan, acot , acsc , asec,
         cosh, sinh, tanh, coth, csch, sech, acosh, asinh, atanh, acoth, acsch, asech]

numtests = 10
std = 10
normal = np.random.randn
points_real = std*normal(numtests)
points_complex = std*(normal(numtests) + normal(numtests)*1j)
points = list(points_real) + list(points_complex)

def isclose(n1, n2):
    return abs(n1 - n2)/(abs(n1) + abs(n2)) < 1e-5

bad = 0
total = 0
for f1 in funcs:
    for f2 in funcs:
        total += 1
        expr1 = f1(x)
        expr2 = expr1.rewrite(f2)
        fine = True
        for p in points:
            val1 = expr1.evalf(subs={x:p})
            val2 = expr2.evalf(subs={x:p})
            if not isclose(val1, val2):
                print()
                print(f1, f2, p)
                print(expr1, ' --> ', expr2)
                print(expr1.subs(x, p), expr2.subs(x, p))
                print(val1.n(5), val2.n(5))
                fine = False
        if not fine:
            bad += 1

print(bad, 'bad rewrites out of', total)
```
Now on master I get:
```python
3 bad rewrites out of 576
```
The bad rewrites are:
```
asec -> atan
acsc -> atan
acos -> atan
```
With the current state of this PR we get:
```
1 bad rewrites out of 576
```
The only bad rewrite is:
```
acsc -> atan
```

oscarbenjamin:
> The only bad rewrite is:
> 
> ```
> acsc -> atan
> ```

After more extensive testing (I increased numtests to 30) I found another case:
```python
asec acot 0.6639099696882306
asec(x)  -->  pi*Piecewise((1, Ne(im(x), 0) | (Eq(im(x), 0) & (Abs(re(x)) > 1))), (3, Eq(im(x), 0) & (re(x) > -1) & (re(x) < 0)), (0, True))/2 + Piecewise((0, Eq(im(x), 0) & Eq(Abs(x), 1)), (zoo, Eq(x, 0)), (1, (arg(x) > pi/2) | (arg(x) < -pi/2) | (Eq(im(x), 0) & (re(x) < 0))), (-1, True))*acot(sqrt(x**2 - 1))
0.967973739051589*I I*(0.967973739051589 - 1.5707963267949*I)
0.96797*I 1.5708 + 0.96797*I
2 bad rewrites out of 576
```
Probably some cases of the Piecewise are incorrect.

oscarbenjamin:
> I have pushed a change where those two problematic ones are rewritten as `Piecewise`

I think that maybe those rewrites were correct so maybe they should be put back.

oscargus:
> maybe those rewrites were correct

Which?

oscarbenjamin:
> > maybe those rewrites were correct
> 
> Which?

The "two problematic ones" that you said you rewrote using Piecewise.

> The only bad rewrite is:
> 
> ```
> acsc -> atan
> ```

I don't think this is a bad rewrite. Rather `evalf` gives the wrong result for the rewritten expression. Most likely `evalf` uses `subs` somewhere and the incorrect symbolic simplification of `sqrt` gives an incorrect result. The expression in the rewrite evaluates correctly using `cmath`:
```python
In [47]: acsc(x).rewrite(atan)
Out[47]: 
⎛      ⎛   ________⎞    ⎞    ____
⎜      ⎜  ╱  2     ⎟   π⎟   ╱  2 
⎜- atan⎝╲╱  x  - 1 ⎠ + ─⎟⋅╲╱  x  
⎝                      2⎠        
─────────────────────────────────
                x                

In [48]: acsc(x).evalf(subs={x:-1+I})
Out[48]: -0.452278447151191 - 0.530637530952518⋅ⅈ

In [49]: acsc(x).rewrite(atan).evalf(subs={x:-1+I})
Out[49]: 0.452278447151191 + 0.530637530952518⋅ⅈ

In [50]: xv = -1 + 1j

In [51]: (-cmath.atan(cmath.sqrt(xv**2 - 1)) + (1/2)*cmath.pi)*cmath.sqrt(xv**2)/xv
Out[51]: (-0.4522784471511907-0.5306375309525179j)
```

oscargus:
> Probably some cases of the Piecewise are incorrect.

I found it.

oscargus:
I rewrote it as a Piecewise and now is the latest script not giving any errors.

The earlier one still have 8 bad cases (so down by 2).

oscarbenjamin:
Are there any cases where this uses a Piecewise but ideally a Piecewise wouldn't be used if the bug in `sqrt` was fixed?

oscargus:
Yes, most of the hyperbolic I would say. Also, it would reduce the risk of actually not typing the correct Piecewise, I think (at least) one of them is not fully correct yet.

I'm slowly trying to get rid of the last bugs and converting the previous rewrites to correct ones (evaluating with 0 or pi kills quite a few of the basic rewrites, like `sin(x).rewrite(tan).subs(x, pi)`). These should probably stay for correctness.

oscarbenjamin:
Ideally we should avoid using Piecewise wherever possible (it is very awkward both symbolically and numerically e.g. in lambdify). So if we are using a Piecewise in place of a potential non-Piecewise expression just to work around some bug then it would be good to have a comment clearly showing what the non-Piecewise version of the formula would be and why it isn't being used.

I want to have a go at fixing the sqrt but but I'm not sure how easy it is to do or how disruptive it might potentially be (maybe this is too late in the release cycle). If we had more time then it would make sense to fix that bug first and then get all the formulae here perfected with minimial use of Piecewise. If we want to get this PR in for 1.11 though then it's probably best not to wait for that. We should make sure though that we can easily fix all the formulae later if/when the bug does get fixed.

oscarbenjamin:
> I want to have a go at fixing the sqrt but but I'm not sure how easy it is to do

Actually it seems to be quite easy to do (#23653)

oscarbenjamin:
Now that #23653 is merged it shouldn't be necessary to work around the sqrt bug using unnecessary Piecewise.

oscarbenjamin:
It would be good to get this in for 1.11

sympy/functions/elementary/hyperbolic.py
```diff
@@ -1354,6 +1376,22 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
     def _eval_rewrite_as_log(self, x, **kwargs):
         return log(x + sqrt(x + 1) * sqrt(x - 1))
 
+    def _eval_rewrite_as_asinh(self, x, **kwargs):
+        return Piecewise((S.Zero, Eq(x, 1)),
+                         (sqrt(x - 1)/sqrt(1 - x)*(pi/2 +
+                          S.ImaginaryUnit*asinh(S.ImaginaryUnit*x)), True))
```
oscarbenjamin, line 1382: This is `acosh(x).rewrite(asinh)`. I think it can be done without Piecewise:
```python
In [25]: 2*asinh(sqrt((x - 1)/2))
Out[25]: 
       ⎛    _______⎞
       ⎜   ╱ x   1 ⎟
2⋅asinh⎜  ╱  ─ - ─ ⎟
       ⎝╲╱   2   2 ⎠
```

sympy/functions/elementary/hyperbolic.py
```diff
@@ -1558,6 +1603,16 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
...
 
+    def _eval_rewrite_as_atanh(self, x, **kwargs):
+        return atanh(1/x)
+
+    def _eval_rewrite_as_asinh(self, x, **kwargs):
+        from sympy.functions.elementary.complexes import arg, Abs
+        return Piecewise((S.Half*asinh(2*x/(1 - x**2)) - pi*x/2*sqrt(-1/x**2),
+                          Abs(x) < 1),
+                         (-S.Half*asinh(2*x/(1 - x**2)), Abs(x) > 1),
+                         (S.Infinity*arg(x), True))
```
oscarbenjamin, line 1614: This is `acoth(x).rewrite(asinh)`. The final case in the Piecewise seems incorrect for `x=I`:
```python
In [1]: acoth(x).subs(x, I)
Out[1]: 
-ⅈ⋅π 
─────
  4  

In [2]: acoth(x).rewrite(asinh).subs(x, I)
Out[2]: ∞
```

oscarbenjamin:
This is the failing test:
https://github.com/sympy/sympy/blob/b64cfcdb640975706c71f305d99a8453ea5e46d8/sympy/functions/elementary/tests/test_trigonometric.py#L575
It has been changed to 
```
In [39]: tan(x).rewrite(sin)
Out[39]: 
⎧    0      for im(x) = 0 ∧ x mod π = 0
⎪                                      
⎪     2                                
⎨2⋅sin (x)                             
⎪─────────           otherwise         
⎪ sin(2⋅x)                             
⎩           
```
which is correct for `x = pi`. The second case still gives `zoo` for `pi/2` etc.

oscargus:
I'll try to get some time to retype the general expressions.

oscargus:
This is now updated. (But will be updated again in a few minutes as I realized that some Piecewises that should not be there was left.)

The possible controversial thing is that e.g `sin(x).rewrite(cot)` is now a Piecewise. The problem is that the earlier expression, 2*cot(x/2)/(cot(x/2)**2 + 1), did not evaluate correctly for e.g. 0.

smichr:
> Now that #23653 is merged

sorry I missed this; glad to see it was fixed

oscarbenjamin:
Okay, let's get this in.

There are some followups to be done from the discussion above but we can defer those to a future issue.

maxkway:
There is some mistake due lambdifing expressions with `cot(x)**2` (and another functions with power auto-rewrite):
`lambdify(x, cot(x)**2)` has 
`Source code:
def _lambdifygenerated(x):
      return tan(x)**(-1.0)**2`
and represents actually `tan(x)` instead of `(tan(x)**(-1.0))**2` (=`tan(x)**(-2.0)`)

oscargus:
I am not sure it is caused by this. Entering `tan(x)**(-1.0)**2` gives `tan(x)**(1.0)`

oscargus:
Ahh, I see the problem now. However, it is not clear how to get around this (at least from the changes made in this PR). Maybe one can return an unevaluated result in this case though.

oscarbenjamin:
It looks correct to me:
```
In [2]: f = lambdify(x, cot(x)**2)

In [3]: import inspect

In [4]: inspect.getsource(f)
Out[4]: 'def _lambdifygenerated(x):\n    return (tan(x)**(-1.0))**2\n'

In [5]: f(2)
Out[5]: 0.20945043706303793

In [6]: cot(2.0)**2
Out[6]: 0.209450437063038
```

oscarbenjamin:
I think that the issue was already fixed in 83f8c53cbc7239e46c2c189bfbd4d28131f80b9e from gh-25514

PR #22995, opened 2022-02-02
Series/limits : Fixes limits for basic inverse and hyperbolic functions through restructuring code in limits.py

anutosh491:
Discovered while debugging #22521 .One of my limits while debugging the reported Pr wasn't working fine though there was nothing that could go wrong in that code , hence I discovered couple of limits which aren't returning expected results leading to the error and these are pretty basic ones involving `trig/hyperbolic` function which I would expect sympy to answer !
Those are
```
>>> limit(acoth(x), x, 0, '-') # this is correct but the following results aren't correct 
I*pi/2
>>> limit(acoth(x), x, 0, '+')
I*pi/2
>>> limit(acoth(x), x, 0, '+-')
I*pi/2
>>> acoth(x).subs(x, 0.000000000000000000000000000000000000000000000000000000000000000000001)
-1.5707963267949*I
```
[Wolfram link for reference for acoth limit ](https://www.wolframalpha.com/input?i=limit+%28acoth%28x%29%29)

After correcting this through the `_eval_as_leading_term` method of  the `acoth` function , I still realized that the bidirectional limit wasn't working as expected and still gave the answer above !

This made me realize that the structure of `limits.py` isn't able to answer couple of bidirectional limits properly as we have the following and as `cdir` is `0` for bi-directional limits it goes through the default case.

```
            try:
                coeff, ex = newe.leadterm(z, cdir=cdir)
            except ValueError:
```
Another example which isn't solved on master
```
>>> limit(acot(x), x, 0, '-') # this is correct
-pi/2
>>> limit(acot(x), x, 0, '+') # this is correct
pi/2
>>> limit(acot(x), x, 0, '+-') # but this is not
pi/2

# This is because newe.leadterm(z, cdir=cdir) checks for the default case and we don't reach the bi-directional check 

>>> acot(x).as_leading_term(x, cdir = -1)
-pi/2
>>> acot(x).as_leading_term(x, cdir = 1)
pi/2
```
The same was happening for `acoth` too, hence whenever we approach `cdir = 0` we should only pass if `expr.leadterm` is same from both side else stop the evaluation !

I guess there will be a couple more like this , i'll do a basic check through all such functions and report if something more is also getting fixed .

Results on Committed branch
```
>>> limit(acoth(x), x, 0, '+')
-I*pi/2
>>> limit(acoth(x), x, 0, '-')
I*pi/2
>>> limit(acoth(x), x, 0, '+-')
ValueError: The limit does not exist since left hand limit = I*pi/2 and right hand limit = -I*pi/2

>>> limit(acot(x), x, 0, '-')
-pi/2
>>> limit(acot(x), x, 0, '+')
pi/2
>>> limit(acot(x), x, 0, '+-')
ValueError: The limit does not exist since left hand limit = -pi/2 and right hand limit = pi/2
```
#### Release Notes

* series
  * Fixes limits for basic inverse and hyperbolic functions through restructuring code in limits.py


anutosh491:
ping @oscargus @0sidharth @smichr

anutosh491:
ping @jksuom

faze-geek:
I went through the code and the diff you made .For any expression `expr` what is the significance of `cdir =0` and `cdir =1` in `expr._eval_as_leading_term()`. Are these both different , I am confused because from `limits.py` whenever we call something like `limit(expr, x, 0)` , the `dir` is assigned to `'+'` by default while makes `cdir = 1` and when this eventually calls for `new.leadterm(x, cdir)`, the `cdir` being passed is `1` which tells me that 
`expr._eval_as_leading_term(x, cdir = 1)` means taking the leading term from the right! hence cdir =1 signifies the leading term from the right !

Now suppose we pass `limit(expr, x, 0, '+-')`, we pass `cdir = 0 ` while calling `new.leadterm(x, cdir)` which tells me that 
`expr._eval_as_leading_term(x, cdir = 0)` should represent a bi-directional leading term for of the expr . 
So why should this give `pi/2`

```
>>> acot(x).as_leading_term(x) # the _eval_as_leading_term method take cdir = 0 by default and this leads to the error here.
pi/2
```

anutosh491:
Yeah true I had this doubt multiple times before and thought of viable solutions (maybe also assigning a cdir =2 or something for the bi-directional case) !
I'll basically explain what's happening here ! In ` limits.py` we always take right to be default(`dir = '+' `or `cdir = 1`) . But in case of the leading_term methods we assume `cdir = 0` to be the default case from right which shouldn't be the case as this creates differences !

So what happens in the following error is that ,`limit(acot(x), x, 0, '+-')` calls `newe.leadterm(x, cdir = 0)` which calls `_eval_as_leading_term(x, cdir = 0)` and then takes the deafault route ( not differenciating between the deafault calculation from the right and the bi-directional case) and returns `pi/2`

This shouldn't be the case right ? I propose changing the following 
1) The default `cdir` parameter of all `_eval_as_leading_term` methods to 1 , so something like `expr.as_leading_term(x)` means taking leading term from the right !!!
2) Secondly whenever we call `limit(expr, x, 0, '+-')` and eventually call `expr.as_leading_term(x, cdir = 0)` if the leading term from both directions is not equal we return a `ValueError ` and otherwise return the value if similar .Hence the above issue will be taken care of !

This will give consistent result as I can clearly see two different defaults being considered in different files ! @0sidharth @oscargus @jksuom 

A basic example would be of the diff would be

```
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -3411,7 +3411,7 @@ def compute_leading_term(self, x, logx=None):
         return s.as_leading_term(x)

     @cacheit
-    def as_leading_term(self, *symbols, logx=None, cdir=0):
+    def as_leading_term(self, *symbols, logx=None, cdir=1):
         """
         Returns the leading (nonzero) term of the series expansion of self.

diff --git a/sympy/functions/elementary/trigonometric.py b/sympy/functions/elementary/trigonometric.py
index 88cb604348..74fb13e35f 100644
--- a/sympy/functions/elementary/trigonometric.py
+++ b/sympy/functions/elementary/trigonometric.py
@@ -2884,7 +2884,7 @@ def taylor_term(n, x, *previous_terms):
             x = sympify(x)
             return S.NegativeOne**((n + 1)//2)*x**n/n

-    def _eval_as_leading_term(self, x, logx=None, cdir=0):
+    def _eval_as_leading_term(self, x, logx=None, cdir=1):
         from sympy.functions.elementary.complexes import (im, re)
         arg = self.args[0]
         x0 = arg.subs(x, 0).cancel()

@@ -2895,6 +2895,9 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
         if x0.is_zero:
             if re(cdir) < 0:
                 return self.func(x0) - S.Pi
+            if re(cdir) == 0:
+                if self._eval_as_leading_term(x, logx=None, cdir=1) != self._eval_as_leading_term(x, logx=None, cdir=-1):
+                    raise ValueError("Leading term from either sides is not equal")
             return self.func(x0)
         if re(cdir) > 0 and re(x0).is_zero and im(x0) > S.Zero and im(x0) < S.One:
             return self.func(x0) + S.Pi
```

And the results obtained would be 

```
>>> from sympy import *
>>> x = Symbol('x')

>>> limit(acot(x), x, 0, '+-')
ValueError: The limit does not exist since left hand limit = -pi/2 and right hand limit = pi/2
>>> limit(acot(x), x, 0, '+')
pi/2
>>> limit(acot(x), x, 0, '-')
-pi/2
>>> acot(x).as_leading_term(x) # default case considering cdir = 1 or direction from right
pi/2
>>> acot(x).as_leading_term(x, cdir = 0)
ValueError: Leading term from either sides is not equal

>>> acot(x).as_leading_term(x, cdir = 1) # default case 
pi/2
>>> acot(x).as_leading_term(x, cdir = -1)
-pi/2
```

anutosh491:
We can revert back to the original code in `limits.py` if the above proposed change is agreed on !! and limits.py isn't the place where the default assumption creates any issue but it's the other leading_term methods which oppose it !

sympy/series/tests/test_limits.py
```diff
@@ -96,6 +96,11 @@ def test_basic1():
     assert limit(log(1/x) + 1/sin(x), x, 0, dir = '+') == oo
     assert limit(log(1/x) + 1/x, x, 0, dir = '+') == oo
 
+    # test acoth function at 0
+    assert limit(acoth(x), x, 0, '+') == -I*pi/2
+    assert limit(acoth(x), x, 0, '-') == I*pi/2
```
jksuom, line 101: There should be more tests for cases like`acoth(-x)` and `acoth(x - I*x)` as `x -> 0`.

anutosh491: True , these are significant tests , will look into them . You could also give your thoughts regarding the comments above !
https://github.com/sympy/sympy/pull/22995#issuecomment-1028648751 this one specifically!

anutosh491: I have updated the code to answer all possible cases and added test for the same, things which work now
```
>>> limit(acosh(x), x, 0)
I*pi/2
>>> limit(acoth(x), x, 0)
-I*pi/2
>>> limit(acoth(x), x, 0, '-')
I*pi/2
>>> limit(acoth(-x), x, 0)
I*pi/2
>>> limit(acoth(-x), x, 0, '-')
-I*pi/2
>>> limit(acoth(x**2), x, 0)
-I*pi/2
>>> limit(acoth(x**2), x, 0, '-')
-I*pi/2
>>> limit(acoth(I*x), x, 0)
-I*pi/2
>>> limit(acoth(I*x), x, 0, '-')
I*pi/2
>>> limit(acoth(I*x**2), x, 0)
-I*pi/2
>>> limit(acoth(I*x**2), x, 0, '-')
-I*pi/2
>>>
>>>
>>> limit(acoth(x - I*x), x, 0)
I*pi/2
>>> limit(acoth(x - I*x), x, 0, '-')
-I*pi/2
>>> limit(acoth(x**2 - I*x**2), x, 0)
I*pi/2
>>> limit(acoth(x**2 - I*x**2), x, 0, '-')
I*pi/2
```

anutosh491:
As limits were failing for `acoth`, I was pretty convinced `acosh` would also not be working for many cases ! So I have fixed that too in the latest commit ! Just 2 of these tests work on master as of now , hence a good amount of basic stuff has been adressed !
```
    assert limit(acosh(I*x), x, 0, '+') == I*pi/2
    assert limit(acosh(I*x), x, 0, '-') == -I*pi/2
    assert limit(acosh(-I*x), x, 0, '+') == -I*pi/2
    assert limit(acosh(-I*x), x, 0, '-') == I*pi/2
    raises(ValueError, lambda: limit(acosh(I*x), x, 0, '+-'))
    assert limit(acosh(x - I*x), x, 0, '+') == -I*pi/2
    assert limit(acosh(x - I*x), x, 0, '-') == I*pi/2
    raises(ValueError, lambda: limit(acosh(x - I*x), x, 0, '+-'))
    assert limit(acosh(I*x**2), x, 0, '+-') == I*pi/2
    assert limit(acosh(x**2 - I*x**2), x, 0, '+-') == -I*pi/2
```

anutosh491:
> Yeah true I had this doubt multiple times before and thought of viable solutions (maybe also assigning a cdir =2 or something for the bi-directional case) ! I'll basically explain what's happening here ! In ` limits.py` we always take right to be default(`dir = '+' `or `cdir = 1`) . But in case of the leading_term methods we assume `cdir = 0` to be the default case from right which shouldn't be the case as this creates differences !
> 
> So what happens in the following error is that ,`limit(acot(x), x, 0, '+-')` calls `newe.leadterm(x, cdir = 0)` which calls `_eval_as_leading_term(x, cdir = 0)` and then takes the deafault route ( not differenciating between the deafault calculation from the right and the bi-directional case) and returns `pi/2`
> 
> This shouldn't be the case right ? I propose changing the following
> 
>     1. The default `cdir` parameter of all `_eval_as_leading_term` methods to 1 , so something like `expr.as_leading_term(x)` means taking leading term from the right !!!
> 
>     2. Secondly whenever we call `limit(expr, x, 0, '+-')` and eventually call `expr.as_leading_term(x, cdir = 0)` if the leading term from both directions is not equal we return a `ValueError ` and otherwise return the value if similar .Hence the above issue will be taken care of !
> 
> 
> This will give consistent result as I can clearly see two different defaults being considered in different files @0sidharth @oscargus @jksuom @smichr ! This won't be disturbing any of the tests and would rather help us find more mistakes if any
> 
> A basic example would be of the diff would be
> 
> ```
> --- a/sympy/core/expr.py
> +++ b/sympy/core/expr.py
> @@ -3411,7 +3411,7 @@ def compute_leading_term(self, x, logx=None):
>          return s.as_leading_term(x)
> 
>      @cacheit
> -    def as_leading_term(self, *symbols, logx=None, cdir=0):
> +    def as_leading_term(self, *symbols, logx=None, cdir=1):
>          """
>          Returns the leading (nonzero) term of the series expansion of self.
> 
> diff --git a/sympy/functions/elementary/trigonometric.py b/sympy/functions/elementary/trigonometric.py
> index 88cb604348..74fb13e35f 100644
> --- a/sympy/functions/elementary/trigonometric.py
> +++ b/sympy/functions/elementary/trigonometric.py
> @@ -2884,7 +2884,7 @@ def taylor_term(n, x, *previous_terms):
>              x = sympify(x)
>              return S.NegativeOne**((n + 1)//2)*x**n/n
> 
> -    def _eval_as_leading_term(self, x, logx=None, cdir=0):
> +    def _eval_as_leading_term(self, x, logx=None, cdir=1):
>          from sympy.functions.elementary.complexes import (im, re)
>          arg = self.args[0]
>          x0 = arg.subs(x, 0).cancel()
> 
> @@ -2895,6 +2895,9 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
>          if x0.is_zero:
>              if re(cdir) < 0:
>                  return self.func(x0) - S.Pi
> +            if re(cdir) == 0:
> +                if self._eval_as_leading_term(x, logx=None, cdir=1) != self._eval_as_leading_term(x, logx=None, cdir=-1):
> +                    raise ValueError("Leading term from either sides is not equal")
>              return self.func(x0)
>          if re(cdir) > 0 and re(x0).is_zero and im(x0) > S.Zero and im(x0) < S.One:
>              return self.func(x0) + S.Pi
> ```
> 
> And the results obtained would be
> 
> ```
> >>> from sympy import *
> >>> x = Symbol('x')
> 
> >>> limit(acot(x), x, 0, '+-')
> ValueError: The limit does not exist since left hand limit = -pi/2 and right hand limit = pi/2
> >>> limit(acot(x), x, 0, '+')
> pi/2
> >>> limit(acot(x), x, 0, '-')
> -pi/2
> >>> acot(x).as_leading_term(x) # default case considering cdir = 1 or direction from right
> pi/2
> >>> acot(x).as_leading_term(x, cdir = 0)
> ValueError: Leading term from either sides is not equal
> 
> >>> acot(x).as_leading_term(x, cdir = 1) # default case 
> pi/2
> >>> acot(x).as_leading_term(x, cdir = -1)
> -pi/2
> ```

What do you'll think , about this suggestions . I am somewhat convinced that we should we going for it ! @jksuom @0sidharth . This also helps couple of my other prs where I am caught up with the same confusion .

anutosh491:
@jksuom @0sidharth what do your thoughts on the comment above https://github.com/sympy/sympy/pull/22995#issuecomment-1028648751

anutosh491:
Converting to draft as the Pr needs some more thought and careful examination from my side . The Code would work for the tests added but I can't say it would work for other related cases .

PR #23000, opened 2022-02-02
Refactored import ordering in functions

oscargus:

Change the import precedence in functions using a "natural" ordering (trig before hyperbolic, elementary before special, etc). Except that combinational comes before elementary since there are more dependencies in that direction.

In some files there are improvements, in other it gets worse. However, one can use the total number of removed line as a measure.

(#22993 can be rewritten to reduce the number of imports subject to this one)

#### Other comments

Also changed some incorrect import locations in solvers.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/functions/special/polynomials.py
+++ b/sympy/functions/special/polynomials.py
@@ -18,16 +18,9 @@ from sympy.functions.elementary.miscellaneous import sqrt
 from sympy.functions.elementary.trigonometric import cos, sec
 from sympy.functions.special.gamma_functions import gamma
 from sympy.functions.special.hyper import hyper
-
-from sympy.polys.orthopolys import (
-    jacobi_poly,
-    gegenbauer_poly,
-    chebyshevt_poly,
-    chebyshevu_poly,
-    laguerre_poly,
-    hermite_poly,
-    legendre_poly
-)
+from sympy.polys.orthopolys import (chebyshevt_poly, chebyshevu_poly,
+                                    gegenbauer_poly, hermite_poly, jacobi_poly,
+                                    laguerre_poly, legendre_poly)
 
 _x = Dummy('x')
 
--- a/sympy/assumptions/cnf.py
+++ b/sympy/assumptions/cnf.py
@@ -4,12 +4,12 @@ only and should not be used anywhere else as these do not possess the
 signatures common to SymPy objects. For general use of logic constructs
 please refer to sympy.logic classes And, Or, Not, etc.
 """
-from itertools import combinations, product
-from sympy.core.singleton import S
-from sympy.logic.boolalg import (Equivalent, ITE, Implies, Nand, Nor, Xor)
+from itertools import combinations, product, zip_longest
+from sympy.assumptions.assume import AppliedPredicate, Predicate
 from sympy.core.relational import Eq, Ne, Gt, Lt, Ge, Le
+from sympy.core.singleton import S
 from sympy.logic.boolalg import Or, And, Not, Xnor
-from itertools import zip_longest
+from sympy.logic.boolalg import (Equivalent, ITE, Implies, Nand, Nor, Xor)
 
 
 class Literal:
@@ -171,7 +171,6 @@ def to_NNF(expr, composite_map=None):
     (Literal(Q.negative(x), False) | Literal(Q.zero(x), False))
     """
     from sympy.assumptions.ask import Q
-    from sympy.assumptions.assume import AppliedPredicate, Predicate
 
     if composite_map is None:
         composite_map = dict()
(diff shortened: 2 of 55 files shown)


oscarbenjamin:
It would be good to document the ordering somewhere in the code. Is it possible to move the remaining delayed imports to the bottom of the files?

oscargus:
Yes, it would be good. Ideally there can be some tool that generates it (as I have not made explicit decisions between all submodules).

I have limited success with imports towards the end of the file, so I have not even tried it.

oscarbenjamin:
> Ideally there can be some tool that generates it

I did try a couple once but they crashed with RecursionError when I pointed them at the SymPy code. The import path is just so messy that these libraries couldn't cope!

> I have limited success with imports towards the end of the file, so I have not even tried it.

I think once you've chosen an ordering between a collection of submodules it should just be a case of imports going at the top or the bottom based on that ordering except when some code actually needs to be executed at import time (e.g. creating instances of a class or calling some function). The difficulty at the moment is just that no sensible ordering has been chosen explicitly and also there are various places that instantiate and use Basic instances during the import process.

oscargus:
I have also tried some and although it worked, it provided too much detail really.

Anyway, I have started mapping imports by hand with the aim of creating results on both top-level module dependencies and inter-module dependencies. So far, concrete, one of the smaller modules, are done and this is what I got:

image

Solid lines are top-level imports, dashed lines are function-level imports. One can argue which is the highest level of summations and products, but apart from that I think it makes sense. expr_with_limit -> expr_with_intlimits -> summations/products is sort of natural.

oscargus:
And for assumptions:
image

oscargus:
I think it makes sense to do this in stages.

I've checked up to and including concrete now in the systematic approach. No ordering there, just moving imports to top level.

For functions it is combinatorial -> elementary -> special.

The next step is to either analyze the internals of combinatorics or core. But the risk is that there will be hundreds of files affected, impossible to review and conflicts. So my suggestion is to merge this (it is clearly better than before if nothing else).

It may also be that further analysis shows that there are even better orderings, but as long as it improves...

oscargus:
It probably makes sense to start separate issue for the internal ordering of submodules. I'll do that.

sympy/functions/special/polynomials.py
```diff
@@ -1025,8 +1025,6 @@ class hermite(OrthogonalPolynomial):
 
     """
 
-    _ortho_poly = staticmethod(hermite_poly)
-
```
oscarbenjamin, line 1029: Not sure I understand this

sympy/functions/special/polynomials.py
```diff
@@ -1047,6 +1045,12 @@ def eval(cls, n, x):
             else:
                 return cls._eval_at_order(n, x)
 
+    @classmethod
+    def _eval_at_order(cls, n, x):
+        if n.is_integer and n >= 0:
+            from sympy.polys.orthopolys import hermite_poly
+            return hermite_poly(int(n), _x).subs(_x, x)
```
oscarbenjamin, line 1052: I also don't understand where this has come from

oscarbenjamin:
There were a few changes I didn't understand (see the two comments above) but it seems to work so +1 to merge if you're done with this.

sympy/functions/special/polynomials.py
oscargus: A consequence of the changes was that functions cannot import at top level from polys (but the other way around). So hermite_poly cannot be available at creation time.

Maybe it could be solved by a property method?

sympy/functions/special/polynomials.py
oscargus: Since _ortho_poly cannot be used for import reasons, I had to copy the method.

Maybe it can be done in a better way though?

sympy/functions/special/polynomials.py
oscarbenjamin: On the higher level I *think* would want the polys module to precede anything in functions. Maybe that way around makes less sense though.

oscarbenjamin: > On the higher level I _think_ I would want the polys module to precede anything in functions. Maybe that way around makes less sense though.

smichr:
@oscargus , there are merge conflicts now

sympy/functions/special/polynomials.py
oscargus: Yes, I sort of expected that poly would be imported from functions and not the other way around. It can of course be arbitrarily chosen, one just have to see the consequences.

oscargus:
This is rebased now. I changed so that polys can be imported in functions on top-level.

oscarbenjamin:
Looks good to me.

PR #23002, opened 2022-02-02
fixes #22988; leave Float alone for domain=RR

### LINKED ISSUE #22988: Evalf and round odd behaviour
`round_fun = lambda n: n.round(15)`

Suppose I have the following simple polynomial:
```
poly=S.One * 88341*x**2/100000
evaled = poly.evalf(n=25)
(evaled.xreplace({f: round_fun(f) for f in evaled.atoms(Float)})).as_poly()
```
The output is `Poly(0.88341*x**2, x, domain='RR')`.

But for the following poly:
```
poly=S.One * 88341*x**2/100000 - S.One*688*x*y/125 + S.One*2439*x/312500 + S.One*643*y**2/50 - S.One*29013*y/3125 + S.One*9766278/1953125
evaled = poly.evalf(n=25)
(evaled.xreplace({f: round_fun(f) for f in evaled.atoms(Float)})).as_poly()
```
the output is `Poly(0.88341000000000003*x**2 - 5.504*x*y + 0.0078048000000000006*x + 12.86*y**2 - 9.28416*y + 5.000334336, x, y, domain='RR')`.

Notice 0.88341000000000003 instead of 0.88341.

Why is that and how to avoid it?

smichr:

closes #22988

This is an attempt to avoid spurious digits from appearing in Poly per issue #22988. Basically, it just tries to leave Float alone during conversions and when converting to mpf uses the string option to create the mpf (dtype).
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/ntheory/primetest.py
+++ b/sympy/ntheory/primetest.py
@@ -101,14 +101,25 @@ def is_square(n, prep=True):
     # >>> print(hex(magic(85)))
     # 0x121065188e001c46298213
     if not 0x2020212020202130202021202030213 & (1 << (n & 127)):
-        return False
+        return False  # e.g. 2, 3
     m = n % (99 * 91 * 85)
     if not 0x209060049048220348a410213 & (1 << (m % 99)):
-        return False
+        return False  # e.g. 17, 68
     if not 0x102e403012a0c9862c14213 & (1 << (m % 91)):
-        return False
+        return False  # e.g. 97, 388
     if not 0x121065188e001c46298213 & (1 << (m % 85)):
-        return False
+        return False  # e.g. 793, 1408
+    # n is either:
+    #   a) odd = 4*even + 1 (and square if even = k*(k + 1))
+    #   b) even with
+    #     odd multiplicity of 2 --> not square, e.g. 39040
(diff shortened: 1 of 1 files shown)


sympy/polys/domains/realfield.py
```diff
@@ -61,10 +62,19 @@ def __hash__(self):
 
     def to_sympy(self, element):
         """Convert ``element`` to SymPy number. """
-        return Float(element, self.dps)
+        _mpf = element._mpf_
+        dps = prec_to_dps(_mpf[-1])
+        return Float(_mpf, dps)
```
oscarbenjamin, line 67: I'm not sure it's a good idea to try and preserve this information in elements of RR. The RealField implementation needs to be improved but it's supposed to represent a fixed precision domain in which all elements have the same precision.

smichr:
Given @oscarbenjamin comment, I am not sure how this issue can be fixed.

sympy/polys/domains/realfield.py
smichr: marking as resolved because I don't see a requirement that all be the same precision

PR #23003, opened 2022-02-02
compat: add back deprecated stub files in iterables

### LINKED ISSUE #22997: Missing deprecated stubs for ordered and default_sort_key
There was a rush to get #22357 in at the time because of merge conflicts but it looks like the TODOs where not all done. In particular the deprecated stubs are missing for e.g.:
```
from sympy.utilities.iterables import ordered
```
This needs to be fixed before 1.10 release.

CC @oscargus

oscarbenjamin:
The ordered and default_sort_key functions where removed from
sympy.utilities.iterables in GH-22357 but no deprecated stub functions
were left behind. This commit adds and tests the stubs to ensure that
anyone depending on importing these functions like
```
  from sympy.utilities.iterables import default_sort_key
  from sympy.utilities.iterables import ordered
```
will see a deprecation warning rather than an error.

The proper way to import these functions both before and after these
changes is:
```
  from sympy import default_sort_key
  from sympy import ordered
```

Fixes #22997 (at least on the 1.10 branch)

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/utilities/iterables.py
+++ b/sympy/utilities/iterables.py
@@ -2993,15 +2993,29 @@ def is_sequence(i, include=None):
             isinstance(i, include))
 
 
-@deprecated(useinstead="sympy.core.traversal.postorder_traversal",
+@deprecated(useinstead="from sympy import postorder_traversal",
     deprecated_since_version="1.10", issue=22288)
 def postorder_traversal(node, keys=None):
     from sympy.core.traversal import postorder_traversal as _postorder_traversal
     return _postorder_traversal(node, keys=keys)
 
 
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
@asmeurer this maybe overlaps with #22900. Perhaps if you want #22900 to be in the release you could make sure that the changes here are included and that could be merged instead.

asmeurer:
You can just merge this as-is. I'll merge it into #22900 and update it once it is in master. I'm changing the signature of SymPyDeprecationWarning, so it will be impossible to "miss" any warnings in the update process. But it also means that things will be in an unmergable state until I manage to go through and update all of them.

PR #23006, opened 2022-02-03
Reordered imports 2

oscargus:

Related to #23000

Major change here:

`solvers` is moved down the import chain (so one can import almost anything on top level in solv`solvers`ers, but not import `solvers`on top level anywhere else)

`simplify` is similarly moved, but still possible to import simplify from top level in `solvers`.

This should break some of the non-obvious import cycles. For example make it easier to import from concrete at top level.

Motivation: both `simplify` and `solvers` should be restrictively used in library code. So it makes more sense to be able to import "everything else" in those two at top level.

#### Other comments

Some minor other changes, primarily moving imports to top-level.

#### Release Notes

NO ENTRY


oscarbenjamin:
Looks good to me

PR #23008, opened 2022-02-03
return precision of Floats passed to Poly in RR

smichr:

Another attempt at closing #22988

That the tests pass suggests that this is a less rarely needed option, triggered by a case like that given in the issue where numbers *rounded* to the same decimal place end up with different precisions. In this case, it seems reasonable to use the minimum precision else extra garbage digits end up showing through in the representation of the object.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/polys/tests/test_polytools.py
+++ b/sympy/polys/tests/test_polytools.py
@@ -53,6 +53,7 @@ from sympy.polys.domains import FF, ZZ, QQ, ZZ_I, QQ_I, RR, EX
 from sympy.polys.domains.realfield import RealField
 from sympy.polys.domains.complexfield import ComplexField
 from sympy.polys.orderings import lex, grlex, grevlex
+from sympy.polys.rootoftools import rootof
 
 from sympy.core.add import Add
 from sympy.core.basic import _aresame
@@ -73,9 +74,10 @@ from sympy.functions.elementary.piecewise import Piecewise
 from sympy.functions.elementary.trigonometric import sin
 from sympy.matrices.dense import Matrix
 from sympy.matrices.expressions.matexpr import MatrixSymbol
-from sympy.polys.rootoftools import rootof
+from sympy.printing.repr import srepr
 from sympy.simplify.simplify import signsimp
 from sympy.utilities.iterables import iterable
+from sympy.utilities.misc import filldedent
 
 from sympy.testing.pytest import raises, warns_deprecated_sympy
 
@@ -3567,3 +3569,15 @@ def test_issue_20985():
     w, R = symbols('w R')
     poly = Poly(1.0 + I*w/R, w, 1/R)
     assert poly.degree() == S(1)
+
+
+def test_issue_22988():
+    e = S('(3/100 + 1234/10*x)')
+    p = Poly(e.subs({k: k.round(15) for k in e.atoms(Rational)}))
+    assert str(p) == "Poly(123.4*x + 0.03, x, domain='RR')"
+    assert str(p.terms()) == (
+        '[((1,), 123.400000000000000), ((0,), 0.030000000000000000)]')
+    assert filldedent(srepr(p)) == filldedent("""
+        Poly(Add(Mul(Float('123.399999999999999994', precision=63),
+        Symbol('x')), Float('0.030000000000000000001', precision=60)),
+        Symbol('x'))""")
--- a/sympy/simplify/simplify.py
+++ b/sympy/simplify/simplify.py
@@ -74,6 +74,9 @@ def separatevars(expr, symbols=[], dict=False, force=False):
     Examples
     ========
 
+    >>> from sympy import simplify, sin rad
+    >>> abs((simplify(30.8**2 - 82.5**2 * sin(rad(11.6))**2)).evalf()
+    >>> assert _-673.447451402970 < 1e-12
     >>> from sympy.abc import x, y, z, alpha
     >>> from sympy import separatevars, sin
     >>> separatevars((x*y)**y)
--- a/sympy/simplify/simplify.py
+++ b/sympy/simplify/simplify.py
@@ -74,7 +74,7 @@ def separatevars(expr, symbols=[], dict=False, force=False):
     Examples
     ========
 
-    >>> from sympy import simplify, sin rad
+    >>> from sympy import simplify, sin, rad
     >>> abs((simplify(30.8**2 - 82.5**2 * sin(rad(11.6))**2)).evalf())
     >>> assert _-673.447451402970 < 1e-12
     >>> from sympy.abc import x, y, z, alpha
--- a/sympy/simplify/simplify.py
+++ b/sympy/simplify/simplify.py
@@ -75,7 +75,7 @@ def separatevars(expr, symbols=[], dict=False, force=False):
     ========
 
     >>> from sympy import simplify, sin rad
-    >>> abs((simplify(30.8**2 - 82.5**2 * sin(rad(11.6))**2)).evalf()
+    >>> abs((simplify(30.8**2 - 82.5**2 * sin(rad(11.6))**2)).evalf())
     >>> assert _-673.447451402970 < 1e-12
     >>> from sympy.abc import x, y, z, alpha
     >>> from sympy import separatevars, sin
--- a/sympy/polys/domains/realfield.py
+++ b/sympy/polys/domains/realfield.py
@@ -2,10 +2,11 @@
 
 
 from sympy.core.numbers import Float
+from mpmath.libmp.libmpf import prec_to_dps
 from sympy.polys.domains.field import Field
 from sympy.polys.domains.simpledomain import SimpleDomain
 from sympy.polys.domains.characteristiczero import CharacteristicZero
-from sympy.polys.domains.mpelements import MPContext
+from sympy.polys.domains.mpelements import MPContext, RealElement
 from sympy.polys.polyerrors import CoercionFailed
 from sympy.utilities import public
 
@@ -61,12 +62,19 @@ class RealField(Field, CharacteristicZero, SimpleDomain):
 
     def to_sympy(self, element):
         """Convert ``element`` to SymPy number. """
-        return Float(element, self.dps)
+        _mpf = element._mpf_
+        dps = prec_to_dps(_mpf[-1])
+        return Float(_mpf, dps)
 
     def from_sympy(self, expr):
         """Convert SymPy's number to ``dtype``. """
+        if isinstance(expr, Float):
+            # I don't know how else to get the converted
+            # expr to look like the original. I tried
+            # passing (expr._mpf_, prec=expr._prec) but
+            # that did not pass the added tests
+            return RealElement(str(expr))
         number = expr.evalf(n=self.dps)
-
         if number.is_Number:
             return self.dtype(number)
         else:


sympy/polys/constructor.py
```diff
@@ -57,9 +57,9 @@ def _construct_simple(coeffs, opt):
                 # this is a composite domain, e.g. ZZ[X], EX
                 return None
 
-    # Use the maximum precision of all coefficients for the RR or CC
+    # Use the minimum precision of all coefficients for the RR or CC
     # precision
-    max_prec = max(c._prec for c in float_numbers) if float_numbers else 53
+    max_prec = min(c._prec for c in float_numbers) if float_numbers else 53
```
oscargus, line 62: Rename it to `min_prec`?

Otherwise, this change seems to very much make sense!

smichr: myopia! I made a change here and in the composite section

sympy/polys/tests/test_constructor.py
```diff
@@ -206,3 +207,11 @@ def test_issue_11538():
         assert construct_domain(x + n)[0] == ZZ[x, n]
     assert construct_domain(GoldenRatio)[0] == EX
     assert construct_domain(x + GoldenRatio)[0] == EX
+
+
+def test_issue_22988():
+    a = pi.round(5)
+    b = (pi/1000).round(5)
+    p = construct_domain(a + b*x)[0]
+    assert str(p.from_sympy(pi)) == '3.14159'
```
smichr, line 216:
```suggestion
    assert str(p.from_sympy(pi)) == '3.14'
```

smichr:
ping @oscarbenjamin

oscarbenjamin:
The variable precision option for RR and CC is buggy. It currently uses a global precision for all instances:
https://docs.sympy.org/latest/modules/polys/domainsintro.html#real-and-complex-fields

I have an incomplete attempt to fix this up:
```diff
diff --git a/sympy/polys/domains/complexfield.py b/sympy/polys/domains/complexfield.py
index b2d40f7..91faf74 100644
--- a/sympy/polys/domains/complexfield.py
+++ b/sympy/polys/domains/complexfield.py
@@ -4,11 +4,16 @@
 from sympy.core.numbers import Float, I
 from sympy.polys.domains.characteristiczero import CharacteristicZero
 from sympy.polys.domains.field import Field
-from sympy.polys.domains.mpelements import MPContext
+from sympy.polys.domains.mpelements import MPContext, MPZ_ONE, mpf_mul, int_types, repr_dps
 from sympy.polys.domains.simpledomain import SimpleDomain
 from sympy.polys.polyerrors import DomainError, CoercionFailed
 from sympy.utilities import public
 
+
+def _convert_fallback(x, strings):
+        raise TypeError("cannot create mpf from " + repr(x))
+
+
 @public
 class ComplexField(Field, CharacteristicZero, SimpleDomain):
     """Complex numbers up to the given precision. """
@@ -25,6 +30,63 @@ class ComplexField(Field, CharacteristicZero, SimpleDomain):
 
     _default_precision = 53
 
+    def __init__(self, prec=_default_precision, dps=None, tol=None):
+
+        context = MPContext()
+        if dps is None:
+            context.prec = prec
+        else:
+            context.dps = dps
+
+        context.types = [context.mpf, context.mpc, context.constant]
+        context._str_digits = context.dps
+        context._repr_digits = repr_dps(context.prec)
+        context._convert_fallback = _convert_fallback
+        context.pretty = False
+
+        if tol is None:
+            tol = self._make_tol(context.prec)
+        elif tol is False:
+            tol = fzero
+        else:
+            tol = self._convert_tol(tol)
+
+        tolerance = context.make_mpf(tol)
+
+        self._context = context
+        self.tolerance = tolerance
+        self.max_denom = self._make_max_denom(self.tolerance)
+
+        self.dtype = context.mpc
+        self.zero = self.dtype(0)
+        self.one = self.dtype(1)
+
+    def _make_tol(self, precision):
+        hundred = (0, 25, 2, 5)
+        eps = (0, MPZ_ONE, 1-precision, 1)
+        return mpf_mul(hundred, eps)
+
+    def make_tol(self):
+        return self.context.make_mpf(self._make_tol(self.precision))
+
+    def _make_max_denom(self, tolerance):
+        if not tolerance:
+            return 1000000
+        else:
+            return int(1/tolerance)
+
+    def _convert_tol(ctx, tol):
+        if isinstance(tol, int_types):
+            return from_int(tol)
+        if isinstance(tol, float):
+            return from_float(tol)
+        if hasattr(tol, "_mpf_"):
+            return tol._mpf_
+        prec, rounding = ctx._prec_rounding
+        if isinstance(tol, str):
+            return from_str(tol, prec, rounding)
+        raise ValueError("expected a real number, got %s" % tol)
+
     @property
     def has_default_precision(self):
         return self.precision == self._default_precision
@@ -37,26 +99,13 @@ def precision(self):
     def dps(self):
         return self._context.dps
 
-    @property
-    def tolerance(self):
-        return self._context.tolerance
-
-    def __init__(self, prec=_default_precision, dps=None, tol=None):
-        context = MPContext(prec, dps, tol, False)
-        context._parent = self
-        self._context = context
-
-        self.dtype = context.mpc
-        self.zero = self.dtype(0)
-        self.one = self.dtype(1)
-
     def __eq__(self, other):
         return (isinstance(other, ComplexField)
            and self.precision == other.precision
            and self.tolerance == other.tolerance)
 
     def __hash__(self):
-        return hash((self.__class__.__name__, self.dtype, self.precision, self.tolerance))
+        return hash((self.__class__.__name__, self.precision, self.tolerance))
 
     def to_sympy(self, element):
         """Convert ``element`` to SymPy number. """
@@ -106,10 +155,7 @@ def from_RealField(self, element, base):
         return self.dtype(element)
 
     def from_ComplexField(self, element, base):
-        if self == base:
-            return element
-        else:
-            return self.dtype(element)
+        return self.dtype(element)
 
     def get_ring(self):
         """Returns a ring associated with ``self``. """
diff --git a/sympy/polys/domains/mpelements.py b/sympy/polys/domains/mpelements.py
index 51208a7..12f4050 100644
--- a/sympy/polys/domains/mpelements.py
+++ b/sympy/polys/domains/mpelements.py
@@ -170,3 +170,35 @@ def almosteq(ctx, s, t, rel_eps=None, abs_eps=None):
         else:
             err = diff/abss
         return err <= rel_eps
+
+def to_rational_2(s, max_denom, limit=True):
+    p, q = to_rational(s._mpf_)
+
+    if not limit or q <= max_denom:
+        return p, q
+
+    p0, q0, p1, q1 = 0, 1, 1, 0
+    n, d = p, q
+
+    while True:
+        a = n//d
+        q2 = q0 + a*q1
+        if q2 > max_denom:
+            break
+        p0, q0, p1, q1 = p1, q1, p0 + a*p1, q2
+        n, d = d, n - a*d
+
+    k = (max_denom - q0)//q1
+
+    number = mpq(p, q)
+    bound1 = mpq(p0 + k*p1, q0 + k*q1)
+    bound2 = mpq(p1, q1)
+
+    if not bound2 or not bound1:
+        return p, q
+    elif abs(bound2 - number) <= abs(bound1 - number):
+        return bound2._mpq_
+    else:
+        return bound1._mpq_
+
+MPContext = PythonMPContext
diff --git a/sympy/polys/domains/realfield.py b/sympy/polys/domains/realfield.py
index 4ae4235..59ba177 100644
--- a/sympy/polys/domains/realfield.py
+++ b/sympy/polys/domains/realfield.py
@@ -5,10 +5,15 @@
 from sympy.polys.domains.field import Field
 from sympy.polys.domains.simpledomain import SimpleDomain
 from sympy.polys.domains.characteristiczero import CharacteristicZero
-from sympy.polys.domains.mpelements import MPContext
+from sympy.polys.domains.mpelements import MPContext, MPZ_ONE, mpf_mul, to_rational_2, fzero, repr_dps
 from sympy.polys.polyerrors import CoercionFailed
 from sympy.utilities import public
 
+
+def _convert_fallback(x, strings):
+    raise TypeError("cannot create mpf from " + repr(x))
+
+
 @public
 class RealField(Field, CharacteristicZero, SimpleDomain):
     """Real numbers up to the given precision. """
@@ -26,6 +31,63 @@ class RealField(Field, CharacteristicZero, SimpleDomain):
 
     _default_precision = 53
 
+    def __init__(self, prec=_default_precision, dps=None, tol=None):
+
+        context = MPContext()
+        if dps is None:
+            context.prec = prec
+        else:
+            context.dps = dps
+
+        context.types = [context.mpf, context.mpc, context.constant]
+        context._str_digits = context.dps
+        context._repr_digits = repr_dps(context.prec)
+        context._convert_fallback = _convert_fallback
+        context.pretty = False
+
+        if tol is None:
+            tol = self._make_tol(context.prec)
+        elif tol is False:
+            tol = fzero
+        else:
+            tol = self._convert_tol(tol)
+
+        tolerance = context.make_mpf(tol)
+
+        self._context = context
+        self.tolerance = tolerance
+        self.max_denom = self._make_max_denom(self.tolerance)
+
+        self.dtype = context.mpf
+        self.zero = self.dtype(0)
+        self.one = self.dtype(1)
+
+    def _make_tol(self, precision):
+        hundred = (0, 25, 2, 5)
+        eps = (0, MPZ_ONE, 1-precision, 1)
+        return mpf_mul(hundred, eps)
+
+    def make_tol(self):
+        return self.context.make_mpf(self._make_tol(self.precision))
+
+    def _make_max_denom(self, tolerance):
+        if not tolerance:
+            return 1000000
+        else:
+            return int(1/tolerance)
+
+    def _convert_tol(ctx, tol):
+        if isinstance(tol, int_types):
+            return from_int(tol)
+        if isinstance(tol, float):
+            return from_float(tol)
+        if hasattr(tol, "_mpf_"):
+            return tol._mpf_
+        prec, rounding = ctx._prec_rounding
+        if isinstance(tol, str):
+            return from_str(tol, prec, rounding)
+        raise ValueError("expected a real number, got %s" % tol)
+
     @property
     def has_default_precision(self):
         return self.precision == self._default_precision
@@ -38,19 +100,6 @@ def precision(self):
     def dps(self):
         return self._context.dps
 
-    @property
-    def tolerance(self):
-        return self._context.tolerance
-
-    def __init__(self, prec=_default_precision, dps=None, tol=None):
-        context = MPContext(prec, dps, tol, True)
-        context._parent = self
-        self._context = context
-
-        self.dtype = context.mpf
-        self.zero = self.dtype(0)
-        self.one = self.dtype(1)
-
     def __eq__(self, other):
         return (isinstance(other, RealField)
            and self.precision == other.precision
@@ -105,7 +154,7 @@ def from_ComplexField(self, element, base):
 
     def to_rational(self, element, limit=True):
         """Convert a real number to rational number. """
-        return self._context.to_rational(element, limit)
+        return to_rational_2(element, self.max_denom, limit)
 
     def get_ring(self):
         """Returns a ring associated with ``self``. """
diff --git a/sympy/polys/domains/tests/test_domains.py b/sympy/polys/domains/tests/test_domains.py
index 40b1cb0..2a66cce 100644
--- a/sympy/polys/domains/tests/test_domains.py
+++ b/sympy/polys/domains/tests/test_domains.py
@@ -1155,11 +1155,6 @@ def test_canonical_unit():
     assert (K.one + i)/(i - K.one) == -i
 
 
-def test_issue_18278():
-    assert str(RR(2).parent()) == 'RR'
-    assert str(CC(2).parent()) == 'CC'
-
-
 def test_Domain_is_negative():
     I = S.ImaginaryUnit
     a, b = [CC.convert(x) for x in (2 + I, 5)]
diff --git a/sympy/polys/ring_series.py b/sympy/polys/ring_series.py
index 538db6d..47a02ad 100644
--- a/sympy/polys/ring_series.py
+++ b/sympy/polys/ring_series.py
@@ -985,7 +985,7 @@ def rs_nth_root(p, n, x, prec):
                     "this domain.")
         else:
             try:                              # RealElement doesn't support
-                const = R(c**Rational(1, n))  # exponentiation with mpq object
+                const = R(R.domain.to_sympy(c)**Rational(1, n))  # exponentiation with mpq object
             except ValueError:                # as exponent
                 raise DomainError("The given series cannot be expanded in "
                     "this domain.")
```

smichr:
@oscarbenjamin , any objection to committing this? A work-around for the issue would probably be appreciated until a more full resolution is made (and I am not aware enough of the issue that is being addressed by your diff, so it would be some time before I could provide useful feedback).

oscarbenjamin:
I'm not entirely sure what this PR is doing and whether it makes conceptual sense for a consistent definition of how RR is supposed to work which is that it should be a domain of fixed precision (53-bit) floating point numbers.

smichr:
> it should be a domain of fixed precision (53-bit) floating point numbers.

This is not how it is set up in master -- it is a domain of fixed precision number and only defaults to 53 if there are no floats in the expression. Formerly, the uniform precision was taken to be the precision of the most precise value. This causes garbage digits to show on numbers that had less precision. Now the precision of the *least* precise value is used to set the uniform precision. And that makes some sense: your precision in computing a value should not exceed the least precise value.

oscarbenjamin:
I haven't thought through all of the implications but I'm really not sure about this change. It means that a low precision float reduces the precision of everything else:
```python
In [1]: e = N(sin(1) ** 2 + cos(1) ** 2 - 1)

In [2]: Poly(x + e)
Out[2]: Poly(1.0*x - 4.0e-124, x, domain='RR')

In [3]: Poly(x + e) / 3
Out[3]: 0.3⋅x - 1.0e-124
```

smichr:
> reduces the precision of everything else

I'm not sure that is a good example -- the value of `e` is without precision and I would not expect to make any quantitative predictions from an expression containing such a value. But I can see an objection: if x is 1000 then we should get an answer of 3333.3 +/- 0.1 (or 0.05), not 3000 +/- 0.1. sigh. I guess the string-hack is the only workaround until variable precisions are supported -- but that is what the last PR I made was, but you didn't like that (so I am a little confused).

oscarbenjamin:
> so I am a little confused

I'm also confused. The first point is I don't really see a major problem to solve here. The solutions you have proposed impact on the accuracy of calculations but the problem itself seems to be mostly just aesthetic. To me there's a hierarchy here where accuracy basically always trumps aesthetics (especially if I don't fully understand all of the implications of a change).

Note that in the example in the issue the number is not one that can be represented exactly in binary:
```
In [8]: import decimal

In [9]: decimal.Decimal(0.88341)
Out[9]: Decimal('0.883410000000000028563817977556027472019195556640625')

In [10]: Rational("0.88341")
Out[10]: 
88341 
──────
100000
```
The expected 5 digit decimal number cannot be represented exactly in binary. Conversely the exact decimal representation of the binary number that we end up with is too long to fully display and so some truncation of output will happen. In this particular case the truncation was not as short as hoped for. If that's an issue (I'm not personally sure that it is) then I don't see why it shouldn't be considered a printing issue.

smichr:
> don't see why it shouldn't be considered a printing issue.

1. The numbers were rounded. 
2. Rounded numbers print properly. 
3. RR removes information about precision (upcasting precision of lower precision number) so it cannot be printed as intended. 

My original workaround was to leave all precisions in a RR expression alone and only convert to Float those that were not already Float. It worked but you said the intent of RR was to have all numbers at the same precision. (But based on the discussion here, it seems like you don't see this as a requirement.)

oscarbenjamin:
> But based on the discussion here, it seems like you don't see this as a requirement.

I do think it's a requirement, or at least that that is the current design and I don't see that it should be changed. I can't say that I fully understand what the effects of the current design are or what difference these changes would make. When thinking about this though I am *much* more concerned about trying to understand the potential effect on the accuracy of calculations then I am on the aesthetic appearance of the float in a printed display.

The issue does not in any way seem serious enough to risk making a change like this without a clear and detailed understanding of its effects. The change in this PR is really quite far-reaching. The SymPy test suite does not have good coverage for operations with floats even though it is clear that there are many problems that occur in relation to them so the fact that tests pass does not give me confidence in these changes.

smichr:
I think that when users see garbage it makes them nervous about the integrity of the product. Yes, prioritize making it do the right thing, but don't neglect showing the right thing.

I went back to the changes I had before that allow the precision of passed values to be retained.

sympy/polys/tests/test_polytools.py
```diff
@@ -3567,3 +3569,15 @@ def test_issue_20985():
...
+def test_issue_22988():
+    e = S('(3/100 + 1234/10*x)')
+    p = Poly(e.subs({k: k.round(15) for k in e.atoms(Rational)}))
+    assert str(p) == "Poly(123.4*x + 0.03, x, domain='RR')"
+    assert str(p.terms()) == (
+        '[((1,), 123.400000000000000), ((0,), 0.030000000000000000)]')
+    assert filldedent(srepr(p)) == filldedent("""
+        Poly(Add(Mul(Float('123.399999999999999994', precision=63), 
+        Symbol('x')), Float('0.030000000000000000001', precision=60)), 
+        Symbol('x'))""")
```
smichr, line 3583: In master this gives
```
>>> str(p)
"Poly(123.4*x + 0.0299999999999999989, x, domain='RR')"
>>> str(p.terms())
'[((1,), 123.400000000000000), ((0,), 0.0299999999999999989)]'
>>> print(filldedent(srepr(p)))

Poly(Add(Mul(Float('123.399999999999999994', precision=63),
Symbol('x')), Float('0.0299999999999999988898', precision=63)),
Symbol('x'))
```

Note that the precision is no longer '63' for the lower precision 0.03, it is 60.

smichr: > low precision float reduces the precision of everything else

Here is a comparison of master and this PR with the `e + x` example you gave:
```python
master

>>> e = N(sin(1) ** 2 + cos(1) ** 2 - 1)
>>> srepr(Poly(e + x))
"Poly(Add(Mul(Float('1.0', precision=7), Symbol('x')), Float('-3.782e-124', precision=7)), Symbol('x'))"
>>> ^Z

this pR

>>> e = N(sin(1) ** 2 + cos(1) ** 2 - 1)
>>> srepr(Poly(e + x))
"Poly(Mul(Float('1.0', precision=7), Symbol('x')), Symbol('x'))"
```

So this PR isn't breaking anything worse than it is.

smichr: My hunch is that the conversion of non-floats (like the Integer coefficient on x) should use max(float prec, 53).

sympy/simplify/simplify.py
```diff
@@ -74,6 +74,9 @@ def separatevars(expr, symbols=[], dict=False, force=False):
     Examples
     ========
 
+    >>> from sympy import simplify, sin rad
+    >>> abs((simplify(30.8**2 - 82.5**2 * sin(rad(11.6))**2)).evalf()
```
smichr, line 78:
```suggestion
    >>> abs((simplify(30.8**2 - 82.5**2 * sin(rad(11.6))**2)).evalf())
```

sympy/simplify/simplify.py
```diff
@@ -74,6 +74,9 @@ def separatevars(expr, symbols=[], dict=False, force=False):
     Examples
     ========
 
+    >>> from sympy import simplify, sin rad
```
smichr, line 77:
```suggestion
    >>> from sympy import simplify, sin, rad
```

sympy/simplify/simplify.py
```diff
@@ -74,6 +74,9 @@ def separatevars(expr, symbols=[], dict=False, force=False):
     Examples
     ========
 
+    >>> from sympy import simplify, sin, rad
+    >>> abs((simplify(30.8**2 - 82.5**2 * sin(rad(11.6))**2)).evalf())
+    >>> assert _-673.447451402970 < 1e-12
```
smichr, line 79: The actual value is 673.295830164072 so it looks like simplify is using the `to/from_sympy` which is odd.

PR #23009, opened 2022-02-03
Moved definition of `illegal`

oscargus:

`illegal = (S.NaN, S.Infinity, S.NegativeInfinity, S.ComplexInfinity)` was defined in `polys.polytools`, but used in `core`. It makes more sense to define this in `core.numbers`.

#### Other comments

It is still possible to import from `polys.polytools`.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/polys/polyutils.py
+++ b/sympy/polys/polyutils.py
@@ -4,6 +4,7 @@
 from sympy.core import (S, Add, Mul, Pow, Eq, Expr,
     expand_mul, expand_multinomial)
 from sympy.core.exprtools import decompose_power, decompose_power_rat
+from sympy.core.numbers import _illegal
 from sympy.polys.polyerrors import PolynomialError, GeneratorsError
 from sympy.polys.polyoptions import build_options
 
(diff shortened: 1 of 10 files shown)


smichr:
I keep thinking this whenever I need to use it. Thanks.

sympy/polys/polyutils.py
```diff
@@ -165,7 +166,6 @@ def order_no_multiple_key(f):
     else:
         return sorted(factors, key=order_no_multiple_key)
 
-illegal = [S.NaN, S.Infinity, S.NegativeInfinity, S.ComplexInfinity]
 illegal_types = [type(obj) for obj in illegal]
 finf = [float(i) for i in illegal[1:3]]
```
smichr, line 170:
```suggestion
illegal_types = [type(obj) for obj in _illegal]
finf = [float(i) for i in _illegal[1:3]]
```

oscargus:
(I force-pushed with that change in it without knowing that you had committed here... Had some weird local test failures in power that actually was not related to this...)

PR #23012, opened 2022-02-04
SymPyPrinter: And prints as & instead of and

### LINKED ISSUE #23010: lambdify - differences between `and` and `&`
I'm trying to remove the dependency of the plotting module from `experimental_lambdify` in favor of `lambdify`. Unfortunately, it doesn't appear to be that simple when interval arithmetic is involved (used by `plot_implicit` when `adaptive=True`). Let's look at the following examples.

## Correct result when using `experimental_lambdify`

```python
from sympy import *
from sympy.plotting.experimental_lambdify import experimental_lambdify
from sympy.plotting.intervalmath import interval
var("x, y")

expr = Eq(y, sin(x)) & (y > 0)
f1 = experimental_lambdify((x, y), expr, use_interval=True)
xi, yi = interval(-4.999996, -4.687508), interval(-0.312501, 0.000005)
f1(xi, yi)

# Output: intervalMembership(False, True)
```

## Wrong result when using `lambdify`

```python
import sympy.plotting.intervalmath.lib_interval as li
# use the interval math functions with lambdify
keys = [t for t in dir(li) if ("__" not in t) and (t not in ["import_module", "interval"])]
vals = [getattr(li, k) for k in keys]
d = {k: v for k, v in zip(keys, vals)}

f2 = lambdify((x, y), expr, modules=d)
f2(xi, yi)

# Output: intervalMembership(None, True)
```

## Debugging

Let's inspect the two functions:

```python
print(f1.eval_str)

# Output: lambda x0, x1 : ( (x1 > 0) & (lambda x, y: x == y)(x1, imath.sin(x0)) )
```

Note the use of the operator `&`.

```python
import inspect
print(inspect.getsource(f2))

# Output:
# def _lambdifygenerated(x, y):
#    return (y == sin(x)) and (y > 0)
```

Note the use of the operator `and`.

Obviously, by modifying the second function to use `&` we will get the correct result:

```python
def func(x, y):
    return (y == li.sin(x)) & (y > 0)
func(xi, yi)

# Output: intervalMembership(False, True)
```

## Question

Is it possible to ask `lambdify` to use `&` instead of `and`?

ThePauliPrinciple:

Fixes #23010

The SymPy printer now uses the `&` operator to represent a "logica
and" rather than `and` since `&` creates an `And` object and `and`
does not.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/printing/pycode.py
+++ b/sympy/printing/pycode.py
@@ -733,6 +733,8 @@ class SymPyPrinter(AbstractPythonCodePrinter):
 
     language = "Python with SymPy"
 
+    _operators = {**AbstractPythonCodePrinter._operators, 'and' : '&'}
+
     def _print_Function(self, expr):
         mod = expr.func.__module__ or ''
         return '%s(%s)' % (self._module_format(mod + ('.' if mod else '') + expr.func.__name__),
--- a/sympy/printing/tests/test_pycode.py
+++ b/sympy/printing/tests/test_pycode.py
@@ -421,3 +421,9 @@ def test_array_printer():
     assert prntr.doprint(ArrayDiagonal(A, [0,1], [2,3])) == 'tensorflow.linalg.einsum("aabbc->cab", A)'
     assert prntr.doprint(ArrayContraction(A, [2], [3])) == 'tensorflow.linalg.einsum("abcde->abe", A)'
     assert prntr.doprint(Assignment(I[i,j,k], I[i,j,k])) == 'I = I'
+
+def test_issue_23010():
+    from sympy.abc import x,y
+    from sympy.printing.pycode import SymPyPrinter
+
+    assert SymPyPrinter().doprint( x & y ) == 'x & y'
--- a/sympy/utilities/tests/test_lambdify.py
+++ b/sympy/utilities/tests/test_lambdify.py
@@ -1558,3 +1558,25 @@ def test_lambdify_cse():
             f = case.lambdify(cse=cse)
             result = f(*case.num_args)
             case.assertAllClose(result)
+
+
+def test_issue_23010():
+    from sympy import Eq, sin, lambdify
+    from sympy.abc import x,y
+    from sympy.plotting.intervalmath import interval
+    from sympy.plotting.intervalmath.interval_membership import intervalMembership
+    import sympy.plotting.intervalmath.lib_interval as li
+
+    expr = Eq(y, sin(x)) & (y > 0)
+    xi, yi = interval(-4.999996, -4.687508), interval(-0.312501, 0.000005)
+
+    # use the interval math functions with lambdify
+    keys = [t for t in dir(li) if ("__" not in t) and (t not in ["import_module", "interval"])]
+    vals = [getattr(li, k) for k in keys]
+    d = {k: v for k, v in zip(keys, vals)}
+
+    f2 = lambdify((x, y), expr, modules=[d, 'sympy'])
+
+    expected_code = ['def _lambdifygenerated(x, y):', '    return (y == sin(x)) & (y > 0)']
+    assert f2.__doc__.splitlines()[10:12] == expected_code
+    assert f2(xi, yi) == intervalMembership(False, True)
--- a/sympy/utilities/tests/test_lambdify.py
+++ b/sympy/utilities/tests/test_lambdify.py
@@ -1561,6 +1561,8 @@ def test_lambdify_cse():
 
 
 def test_issue_23010():
+    if not numpy:
+        skip("numpy not installed.")
     from sympy import Eq, sin, lambdify
     from sympy.abc import x,y
     from sympy.plotting.intervalmath import interval


ThePauliPrinciple:
Looks like the `lambdify` test needs numpy to actually evaluate, does anyone know how to enable the test only when `np` is defined?

sympy/printing/pycode.py
```diff
@@ -733,6 +733,8 @@ class SymPyPrinter(AbstractPythonCodePrinter):
 
     language = "Python with SymPy"
 
+    _operators = {**AbstractPythonCodePrinter._operators, 'and' : '&'}
```
asmeurer, line 736: Shouldn't this also include the other logical operations?

asmeurer:
I'm a little confused just what exactly is the SymPyPrinter? It seems the only place it is used is in lambdify with modules='sympy'. But why is that not the same thing as the str printer, which should always print Python code that recreates the SymPy expression, or the `srepr` printer, which does this but even more rigorously?

asmeurer:
To be completely honest actually, lambdify(modules='sympy') really makes little sense to me. I think it's mostly just there for historical reasons. But I don't understand why `lambdify(x, expr, modules='sympy')` it should be anything more than just a wrapper that returns `lambda _x: expr.subs({x: _x})` (without any "printing"). 

More to the point, why is @Davide-sp using modules='sympy' in https://github.com/sympy/sympy/issues/23010?

ThePauliPrinciple:
I think the SymPyPrinter could indeed just be a rename of the str/repr printer.

I suggested to try `modules='sympy'` because there was the desire to replace a plot module specific `lambdify` with the generically available `lambdify`. But considering a complete replacement of sympy plotting is currently discussed, perhaps this is no longer needed?

asmeurer:
I don't want to derail this too much with talk of refactoring SymPyPrinter. Given it's mostly historical, it probably should just stay as it is. I agree that using `&` instead of `and` looks more correct for it.

I do think we should consider whether it should be used in the plotting module, though. I really think anyone using it should reconsider whether they should instead be using something else, and that especially applies to library code.

ThePauliPrinciple:
Currently, plotting uses `experimental_lambdify`, which is it's entirely own, working in a completely different way, version of `lambidfy`, if that can be replaced with this modification, I think it is an improvement.

https://github.com/sympy/sympy/blob/master/sympy/plotting/experimental_lambdify.py

Davide-sd:
Ohh I see there are new comments here. Sorry for not replying sooner, but @asmeurer tagged a wrong person :)

> To be completely honest actually, lambdify(modules='sympy') really makes little sense to me. I think it's mostly just there for historical reasons. But I don't understand why lambdify(x, expr, modules='sympy') it should be anything more than just a wrapper that returns lambda _x: expr.subs({x: _x}) (without any "printing").

Consider this example: `lambdify(x, re(sqrt(x)), modules=[d, "math", "sympy"])`, where `d` is a dictionary of functions to be used in the evaluation. `"sympy"` is added at the end because if a particular function is not defined in `d` or `"math"` (such as `re`), then the evaluation will be performed with sympy.

The reason why I need to provide `d` is because I'm trying to remove `experimental_lambdify` and use `lambdify` instead. `d` contains all functions defined in the `sympy.plotting.intervalmath` module, which are necessary to create implicit plots.

oscarbenjamin:
I would have thought that intervalmath should be handled with a different modules argument like `modules='intervalmath'` or something.

Davide-sd:
> I would have thought that intervalmath should be handled with a different modules argument like `modules='intervalmath'` or something.

Not sure if that's a good thing to do, because then people might relies on it. Quoting from `sympy/plotting/intervalmath/interval_arithmetic.py`:

```
"""
Interval Arithmetic for plotting.
This module does not implement interval arithmetic accurately and
hence cannot be used for purposes other than plotting. If you want
to use interval arithmetic, use mpmath's interval arithmetic.
The module implements interval arithmetic using numpy and
python floating points. The rounding up and down is not handled
and hence this is not an accurate implementation of interval
arithmetic.
The module uses numpy for speed which cannot be achieved with mpmath.
"""
```

oscarbenjamin:
There can be a way to make the intervalmath implementation private. The point is that probably it should be a separate printer/module. If it's based on numpy it might make sense to use the numpy printer but have the intervalmath module provide all of implementations of the corresponding numpy functions.

> `sympy.plotting.intervalmath` module, which are necessary to create implicit plots.

Is it actually necessary to use these? I would have thought that matplotlib's contour/contourf was the simplest way to do this e.g.:
```python
x = y = np.linspace(-5, 5, 100)
X, Y = np.meshgrid(x, y)
plt.contour(x, y, X * np.exp(X) - Y, [0])
```
That comes out with a nice antialiased line unlike `plot_implicit(x*exp(x)-y)`.

Davide-sd:
> Is it actually necessary to use these? I would have thought that matplotlib's contour/contourf was the simplest way to do this e.g.:

Yes, it is necessary. Plot implicit has 2 algorithms:

1. uniform sampling to create contour plots. It is NOT able to plot Boolean expressions.
2. adaptive algorithm to create fills. This is capable of plotting Boolean expressions because it is using `sympy.plotting.intervalmath`.

oscarbenjamin:
> 2\. adaptive algorithm to create fills. This is capable of plotting Boolean expressions because it is using `sympy.plotting.intervalmath`.

Is it not possible to do this using numpy and matplotlib? I'm imagining something like
```
x = y = np.linspace(-5, 5, 1000)
X, Y = np.meshgrid(x, y)
plt.contourf(Y > X ** 2, [0.5, 1.5])
```

Davide-sd:
> Is it not possible to do this using numpy and matplotlib? I'm imagining something like

The following example plot a Boolean expression with adaptive algorithm and `sympy.plotting.intervalmath`.

```python
from sympy import *
var("x, y")
plot_implicit((y > sin(x)) & (y > 0), adaptive=True)
```

I don't think that expression can be plotted with meshgrid.

oscarbenjamin:
> I don't think that expression can be plotted with meshgrid.

Isn't it just this:
```
x = y = np.linspace(-5, 5, 1000)
X, Y = np.meshgrid(x, y)
plt.contourf((Y > np.sin(X)) & (Y > 0), [0.5, 1.5])
```

Davide-sd:
> Isn't it just this:

Ouch, I made your life easy :) I copied the wrong command... Here is the one that has bugged me:

```python
from sympy import *
var("x, y")
plot_implicit(Eq(y, sin(x)) & (y > 0), adaptive=True)
```

oscarbenjamin:
> Except actually contour has nan handling so:

Might be worth checking if that is an explicitly supported feature...

Davide-sd:
> That one produces curves with end-points so maybe cases like that would be hard to produce with contour since it necessarily produces unbounded or closed curves...

Most importantly, it wouldn't be possible to achieve that result by lambdifying the expression:

```python
plt.figure()
var("x, y")
f = lambdify([x, y], Eq(y, sin(x)) & (y > 0))
X = Y = np.linspace(-5, 5, 1000)
X, Y = np.meshgrid(X, Y)
plt.contour(X, Y, f(X, Y), [0])
```

Because `f(X, Y)` returns boolean values (in this case, False everywhere): the result is the evaluation was done point by point.

That's the power of the interval math and adaptive algorithm: it starts with a rough discretization of intervals and goes to a finer one until a condition is met, and it will find the values where the expression is satisfied.

oscarbenjamin:
It's definitely possible to make this work with contour/contourf. I won't dispute that intervalmath is a useful approach though. Personally I'd like to see much more use of interval arithmetic in sympy but currently it's just an obscure fringe feature in the plotting module.

I also don't think that it would be a problem to expose intervalmath publicly if the caveats are clearly documented.

oscarbenjamin:
This has a merge conflict now. If everyone defaults to adding tests at the end of test files then it creates a high chance of merge conflicts.

ThePauliPrinciple:
> This has a merge conflict now. If everyone defaults to adding tests at the end of test files then it creates a high chance of merge conflicts.

I usually try to find a consistent place within the tests, but a new "issue test" usually ends up at the end of a file for me. Do you have a better suggestion to place those?

ThePauliPrinciple:
I don't think so, this class isn't very helpful in the first place I think. E.g. it is not using `_module_format` that you would expect from a python printer. It's probably best to leave it as is.

PR #23032, opened 2022-02-07
Updated Docstrings

### LINKED ISSUE #22981: Docstring for Symbol is very thin
I linked to the docstring of `Symbol`, which is one of the most widely used objects in SymPy and was surprised at how little information was present in the docstring. This is all it is:

```
class Symbol(AtomicExpr, Boolean):
    """
    Assumptions:
       commutative = True
    You can override the default assumptions in the constructor.
    Examples
    ========
    >>> from sympy import symbols
    >>> A,B = symbols('A,B', commutative = False)
    >>> bool(A*B != B*A)
    True
    >>> bool(A*B*2 == 2*A*B) == True # multiplication by scalars is commutative
    True
    """
```

I would expect this to explain quite a bit more: all assumptions, passing in greek letters, or subscripts, and probably more.

ghost:
The Docstring for ``Symbol`` is now increased

Fixes #22981 

Previously the docstring for ``Symbol`` was very thin and therefore is now updated to be more informative so that readers will be able to see what symbols in ``SymPy`` can really do, like uses in **sympy expressions, preety printing of greek letters etc.**
And therefore is now updated.

#### Other comments
- [x] Opened on a seperate branch.

#### Release Notes

NO ENTRY


sympy/core/symbol.py
```diff
@@ -201,15 +201,43 @@ def numbered_string_incr(s, start=0):
...
     ========
 
-    >>> from sympy import symbols
+    >>> from sympy import Symbol, symbols
+    >>> a = Symbol('a')
+    >>> x, y, z = symbols('x y z')
+
+    SymPy will automatically pretty print symbols with greek letters.
+
+    >>> alpha, beta, gamma = symbols('alpha beta gamma')
```
anutosh491, line 229: I guess this should be `alpha, beta, gamma = symbols('α  β γ')` for the doctest to pass

KuldeepBorkar: No

anutosh491: Sorry my bad , I wrongly understood what you were trying to implement!

sympy/core/symbol.py
```diff
@@ -201,15 +201,43 @@ def numbered_string_incr(s, start=0):
...
+
+    Sympy will automatically pretty print symbols with subscripts.
+
+    >>> alpha1, alpha2 = symbols('alpha1 alpha2')
+    >>> 2*alpha1 + 3*alpha2 #doctest: +SKIP
+    2α₁ + 3α₂
+
+    You can override the default assumption in the constructor.
+
     >>> A,B = symbols('A,B', commutative = False)
```
oscarbenjamin, line 241: Commutative is one of the least useful properties to set.

It would be good to show some quick examples of more useful things like `positive=True` here and then link to this page for more information:

oscarbenjamin: Note that `even=False` is a bad example as it's almost always not going to do anything useful:
https://docs.sympy.org/latest/guides/assumptions.html#interpretation-of-the-predicates

faze-geek: @oscarbenjamin My bad ,I opened the doc and that was the first thing I got my eyes on ! It might not be an important assumption to document/add here !

KuldeepBorkar: Thank you for the review and response Sir @oscarbenjamin  and @faze-geek I just implemented the changes, can you please check once are they okay or should I add something more

KuldeepBorkar: flake8 check is failing I think due to the use of `sqrt`  is there something I can do for that to pass that check?

faze-geek: You have missed that .Otherwise I still think the documentation is partial but @oscarbenjamin could comment better on that !

sympy/core/symbol.py
```diff
@@ -201,21 +201,77 @@ def numbered_string_incr(s, start=0):
 
 class Symbol(AtomicExpr, Boolean):
     """
+    Symbols are variables, and like all other variables, they need to
+    be assigned before you can use them.
+    SymPy :func:`Symbol` function returns just a single symbol taken from argument ``name`` whereas :func:`symbols` function can be used
+    to return a sequence of symbols.
+    Symbols can be used to build SymPy expressions which are immutable.
```
oscarbenjamin, line 208: Use your code editor to wrap these lines to a consistent length.

KuldeepBorkar: > Use your code editor to wrap these lines to a consistent length.

@oscarbenjamin I guess you mean I should wrap these lines in such a way that number of lines can be reduced like now I wrapped it to line number 206.

oscarbenjamin: No that's definitely not what I meant. Read PEP 8:
https://www.python.org/dev/peps/pep-0008/#maximum-line-length

Long lines are hard to read in many situations so it is better to wrap them so that the code has a consistent line length like say 80 characters per line. This is not followed strictly in sympy but there is no reason not to do it for lines like this that are in a docstring.

Wrapping the above would give something like
```
    Symbols are variables, and like all other variables, they need to be
    assigned before you can use them. SymPy :func:`Symbol` function returns
    just a single symbol taken from argument ``name`` whereas :func:`symbols`
    function can be used to return a sequence of symbols.  Symbols can be used
    to build SymPy expressions which are immutable.
```
This isn't something that needs to be done manually because most code editors have a way to do this automatically like you select the lines and then ask the editor to wrap them.

KuldeepBorkar: > This isn't something that needs to be done manually because most code editors have a way to do this automatically like you select the lines and then ask the editor to wrap them.

@oscarbenjamin I made the changes now.
Thank you so much for explaining in detail,
I am learning a lot from you : )

KuldeepBorkar:
Should I merge the commits and force push the changes or it should be okay?

KuldeepBorkar:
@oscarbenjamin Could you please tell me why it is showing here that 28 checks are successful and 16 expected checks (Some checks are not completed yet it's showing --- **_Waiting for status to be reported_** in front of them I don't know why).

oscarbenjamin:
The sphinx job failed.

KuldeepBorkar:
> The sphinx job failed.
Please continue this with #24197

PR #23038, opened 2022-02-07
add some type hints

eagleoflqj:
Working on a relatively big change. Pre-PR typing part for easier review.
~~These functions should not return SymPy object and python object at the same time.~~
These functions should have clear return type.
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -1,4 +1,4 @@
-from typing import Tuple as tTuple
+from typing import Tuple as tTuple, TYPE_CHECKING
 from collections.abc import Iterable
 from functools import reduce
 
@@ -16,6 +16,8 @@ from sympy.utilities.iterables import has_variety, sift
 from mpmath.libmp import mpf_log, prec_to_dps
 from mpmath.libmp.libintmath import giant_steps
 
+if TYPE_CHECKING:
+    from .numbers import Number
 
 from collections import defaultdict
 
@@ -1989,7 +1991,7 @@ class Expr(Basic, EvalfMixin):
         d.update({m: c})
         return d
 
-    def as_base_exp(self):
+    def as_base_exp(self) -> tTuple['Expr', 'Expr']:
         # a -> b ** e
         return self, S.One
 
@@ -3492,7 +3494,7 @@ class Expr(Basic, EvalfMixin):
         c = c.subs(d, log(x))
         return c, e
 
-    def as_coeff_Mul(self, rational=False):
+    def as_coeff_Mul(self, rational: bool = False) -> tTuple['Number', 'Expr']:
         """Efficiently extract the coefficient of a product. """
         return S.One, self
 
--- a/sympy/core/exprtools.py
+++ b/sympy/core/exprtools.py
@@ -18,6 +18,7 @@ from sympy.utilities.iterables import (common_prefix, common_suffix,
         variations, iterable, is_sequence)
 
 from collections import defaultdict
+from typing import Tuple as tTuple
 
 
 _eps = Dummy(positive=True)
@@ -212,7 +213,7 @@ def _monotonic_sign(self):
             return rv.subs(_eps, 0)
 
 
-def decompose_power(expr):
+def decompose_power(expr: Expr) -> tTuple[Expr, int]:
     """
     Decompose power into symbolic base and integer exponent.
 
@@ -245,26 +246,27 @@ def decompose_power(expr):
     if exp.is_Number:
         if exp.is_Rational:
             if not exp.is_Integer:
-                base = Pow(base, Rational(1, exp.q))
-
-            exp = exp.p
+                base = Pow(base, Rational(1, exp.q))  # type: ignore
+            e = exp.p  # type: ignore
         else:
-            base, exp = expr, 1
+            base, e = expr, 1
     else:
         exp, tail = exp.as_coeff_Mul(rational=True)
 
         if exp is S.NegativeOne:
-            base, exp = Pow(base, tail), -1
+            base, e = Pow(base, tail), -1
         elif exp is not S.One:
-            tail = _keep_coeff(Rational(1, exp.q), tail)
-            base, exp = Pow(base, tail), exp.p
+            # todo: after dropping python 3.7 support, use overload and Literal
+            #  in as_coeff_Mul to make exp Rational, and remove these 2 ignores
+            tail = _keep_coeff(Rational(1, exp.q), tail)  # type: ignore
+            base, e = Pow(base, tail), exp.p  # type: ignore
         else:
-            base, exp = expr, 1
+            base, e = expr, 1
 
-    return base, exp
+    return base, e
 
 
-def decompose_power_rat(expr):
+def decompose_power_rat(expr: Expr) -> tTuple[Expr, Rational]:
     """
     Decompose power into symbolic base and rational exponent.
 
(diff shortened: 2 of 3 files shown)


eagleoflqj:
It seems that when `_parallel_dict_from_expr_no_gens` returns a representation of polynomial (opt.series is False), the exps must be python int. When it returns a representation of series (opt.series is True), the exps can be SymPy Rational.

oscarbenjamin:
> the exps must be python int

Is this difference intentional? Or is it just what currently happens because use of types isn't always consistent but it could be improved?

eagleoflqj:
> > the exps must be python int
> 
> Is this difference intentional? Or is it just what currently happens because use of types isn't always consistent but it could be improved?

Seems intentional. When constructing a Poly, PolyElement and DMP objects are created. Later PythonMPQ object is constructed from them. PythonMPQ object explicitly accept `int, PythonMPQ, Decimal, float, str and Fraction as parameter. It rejects SymPy Integer.

oscarbenjamin:
> PythonMPQ object explicitly accept `int, PythonMPQ, Decimal, float, str and Fraction` as parameter.

I wrote the PythonMPQ class so I can say that that was deliberate. I'm unsure though how that relates to `_parallel_dict_from_expr_no_gens` (a function that I haven't studied).

eagleoflqj:
> > PythonMPQ object explicitly accept `int, PythonMPQ, Decimal, float, str and Fraction` as parameter.
> 
> I wrote the PythonMPQ class so I can say that that was deliberate. I'm unsure though how that relates to `_parallel_dict_from_expr_no_gens` (a function that I haven't studied).

`_parallel_dict_from_expr_no_gens` is used to construct Poly and series. It calls `decompose_power`.
At first, I want to make `decompose_power` return `tTuple[Expr, Integer]`. But CI failed.
That's because the value that `_parallel_dict_from_expr_no_gens` returns may be used to construct `PythonMPQ`. So it must be int instead of `Integer`.

oscarbenjamin:
Maybe `_parallel_dict_from_expr_no_gens` should always return `int`? (I haven't yet looked at the function)

eagleoflqj:
> Maybe `_parallel_dict_from_expr_no_gens` should always return `int`? (I haven't yet looked at the function)

For polynomials, yes. But `_parallel_dict_from_expr_no_gens` is intended to return `Rational` when `opt.series` is True:
https://github.com/sympy/sympy/blob/d62e689e4f2944939181c89d4b9e99c6b8e2ffa9/sympy/polys/polyutils.py#L261-L267

oscarbenjamin:
The `PythonMPQ` class is a fallback for when gmpy2 is not installed and is used in place of gmpy2.mpq. The `mpq` type cannot accept `Integer`:
```python
In [1]: import gmpy2

In [2]: import sympy

In [3]: gmpy2.mpq(sympy.Integer(3))
Input In [3], in <module>
----> 1 gmpy2.mpq(sympy.Integer(3))

TypeError: mpq() requires numeric or string argument
```
Therefore `PythonMPQ` should not accept `Integer` either.

I'm not sure what's the best way to type `decompose_power` or how to handle the type hints so that mypy knows that PythonMPQ is not being called with `Integer`.

sympy/core/exprtools.py
```diff
@@ -245,47 +246,50 @@ def decompose_power(expr):
...
+    return base, e
 
 
-def decompose_power_rat(expr):
+def decompose_power_rat(expr: Expr) -> tTuple[Expr, Rational]:
     """
     Decompose power into symbolic base and rational exponent.
 
     """
     base, exp = expr.as_base_exp()
```
smichr, line 274: maybe change this exp to e, too?

eagleoflqj: Not applicable. `exp` is `Expr`, but e the return value is `Rational`. mypy complains:
error: Incompatible return value type (got "Tuple[Expr, Expr]", expected "Tuple[Expr, Rational]")

eagleoflqj:
Thank you for showing me that! I originally wanted to convert Integer to int inside `PythonMPQ` and now I know it's totally not feasible.

> I'm not sure what's the best way to type `decompose_power`

So I decided to keep it returning int. Just add type hints to it and use a new variable `e` to make mypy happy.

> or how to handle the type hints so that mypy knows that PythonMPQ is not being called with `Integer`.

That's another big thing, since from `decompose_power` to `PythonMPQ` there are too many layers, which are all untyped.

eagleoflqj:
ping @oscarbenjamin 
I think we can merge this?

sympy/core/exprtools.py
```diff
@@ -245,47 +246,50 @@ def decompose_power(expr):
     if exp.is_Number:
         if exp.is_Rational:
             if not exp.is_Integer:
-                base = Pow(base, Rational(1, exp.q))
-
-            exp = exp.p
+                base = Pow(base, Rational(1, cast(Rational, exp).q))
+            e = cast(Rational, exp).p
```
oscarbenjamin, line 250: I wonder how we would have to go with adding `cast` like this in different parts of the codebase. Is there a way somehow to get mypy to understand that for `Basic` instances `exp.is_Rational` is equivalent to `isinstance(exp, Rational)`?

sympy/core/expr.py
```diff
@@ -3492,7 +3494,7 @@ def leadterm(self, x, logx=None, cdir=0):
         c = c.subs(d, log(x))
         return c, e
 
-    def as_coeff_Mul(self, rational=False):
+    def as_coeff_Mul(self, rational: bool = False) -> tTuple['Number', 'Expr']:
```
oscargus, line 3497: Not your fault, but the consequences on keyword arguments are, well, not my favorite part of type checking...

sympy/core/exprtools.py
eagleoflqj:
```py
import functools

class A:
    is_B = False
    a = 0

    @functools.cached_property
    def method(self):
        return isinstance(self, B)

class B(A):
    is_B = True
    a = 1
    b = 2

def f(a: A):
    print(a.a)
    if isinstance(a, B):
        print(a.b)
    if a.is_B:
        print(a.b)
    if a.method:
        print(a.b)

f(B())
```
```
hint.py:21: error: "A" has no attribute "b"
hint.py:23: error: "A" has no attribute "b"
```
This minimal example shows even using property mypy cannot tell a is B. I expect @overload could be used to decorate a field like is_Rational in future python version. Now `cast` has to be used to avoid `# type: ignore`.

oscarbenjamin: I think I'd rather have a `# type: ignore` than the cast. Adding hints in function signatures or variable declarations to "annotate" the code is one thing but complicating the actual code with extra bits that don't do anything seems like too much to me.

Besides making improvements to mypy and typing to handle this case I can see several options here:

1. Change `exp.is_Rational` to `isinstance(exp, Rational)`. In principle I would rather write the code like this anyway but it is actually slower at runtime and also exacerbates circular import problems.
2. Use `type: ignore`. This is considered undesirable if you actually want the checker to check things for you because it effectively disables checking. Since it's only changing the code by adding a comment though this doesn't mess up the main code.
3. Use `cast`. This has a runtime cost *and* messes up the code *and* also causes problems with circular imports. Effectively the use of `cast` also disables checking at least the point is to override the checker so what is stated by the case is not actually checked.

I'm not sure why `cast` would be considered the preferred option here. Really I think it would be good to go ask the mypy people how this can be done or suggest that mypy be improved to handle this case rather than compromising on other things just for the sake of the checker.

moorepants:
Is there a discussion on adding type hints into the sympy codebase somewhere?

oscarbenjamin:
> Is there a discussion on adding type hints into the sympy codebase somewhere?

Probably #17945 is the closest issue to that.

sympy/core/exprtools.py
eagleoflqj: https://docs.python.org/3/library/typing.html#typing.cast cast shouldn't have runtime cost. I use that because `# type: ignore` ignores the whole line which may not be what we want. But I'm OK to update this PR to use `# type: ignore` instead.

oscarbenjamin: Calling any function in Python has a runtime overhead even if the function doesn't do anything:
```python
In [6]: from typing import cast

In [7]: x = Symbol('x')

In [8]: timeit x.is_Rational
81.5 ns ± 4.45 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [9]: timeit isinstance(x, Rational)
330 ns ± 3.1 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [10]: timeit cast(x, Rational)
225 ns ± 6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```
More importantly though the cast obfuscates what should be a very simple line of code making it harder to read.

oscarbenjamin:
Okay, looks good.

PR #23039, opened 2022-02-08
fix 'KeyError: 'ComplexInfinity'' with PythonCodePrinter 

amitport:

When using scypi or mpmath module, `_print_NaN` is being overridden by `_print_known_const`, which relies on the input `expr` to hold the correct expression.

`_print_ComplexInfinity` delegates to `_print_NaN` but sends the original `ComplexInfinity` expr which causes an exception since it is not in `known_constants`.

#### Release Notes

NO ENTRY

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -157,6 +157,7 @@ Amit Kumar <dtu.amit@gmail.com>
 Amit Kumar <dtu.amit@gmail.com> <aktech@users.noreply.github.com>
 Amit Kumar <dtu.amit@gmail.com> <dtu.amit+gh@gmail.com>
 Amit Manchanda <amitdelhi1995@gmail.com>
+Amit Portnoy <amit.portnoy@gmail.com> Amit Portnoy <1131991+amitport@users.noreply.github.com>
 Amit Saha <amitsaha.in@gmail.com> <asaha@redhat.com>
 Ananya H <ananyaha93@gmail.com>
 Anatolii Koval <weralwolf@gmail.com>
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,7 +4,7 @@ those who explicitly didn't want to be mentioned. People with a * next
 to their names are not found in the metadata of the git history. This
 file is generated automatically by running `./bin/authors_update.py`.
 
-There are a total of 1127 authors.
+There are a total of 1129 authors.
 
 Ondřej Čertík <ondrej@certik.cz>
 Fabian Pedregosa <fabian@fseoane.net>
@@ -1133,3 +1133,5 @@ Tom Fryers <61272761+TomFryers@users.noreply.github.com>
 Zouhair <zouhair.mahboubi@gmail.com>
 zzj <29055749+zjzh@users.noreply.github.com>
 shubhayu09 <guptashubhayu601@gmail.com>
+Siddhant Jain <siddhantashoknagar@gmail.com>
+Amit Portnoy <amit.portnoy@gmail.com>
--- a/sympy/printing/numpy.py
+++ b/sympy/printing/numpy.py
@@ -24,7 +24,8 @@ _known_constants_numpy = {
     'EulerGamma': 'euler_gamma',
     'NaN': 'nan',
     'Infinity': 'PINF',
-    'NegativeInfinity': 'NINF'
+    'NegativeInfinity': 'NINF',
+    'ComplexInfinity': 'nan',
 }
 
 _numpy_known_functions = {k: 'numpy.' + v for k, v in _known_functions_numpy.items()}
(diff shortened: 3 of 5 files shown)


bjodah:
This needs a test.

amitport:
@bjodah this is my first contribution, just happen to stumble on a solution, not sure I can add a test here and will appreciate it if someone else could look into it.

bjodah:
Great! Glad to have you submitting this PR, much appreciated!

I suspect this is the wrong fix though. But if you add a (prevously failing) test i test_lambdify.py which excercises this code path (via `lambdify`), it'll be easier to come up with suggestions.

amitport:
Hi, there may be more than one issue, in any case, added a few tests a different fix

amitport:
I traced the difference in behavior to https://github.com/sympy/sympy/blob/master/sympy/printing/pycode.py#L62 @oscargus

There was always a problem. before 'ComplexInfinity' was added to math constants, my first fix was needed, after the second fix is relevant. I guess another solution would be to revert the line above and add my first fix (sending `S.NaN` in `_print_ComplexInfinity`).

bjodah:
I'm not convinced we want to add `ComplexInfinity` to the printer if there is no obvious representation in the target printer. For example: I'd much rather have `lambdify` raise an exception, telling me that I'm trying to create a function out of an expression containing `ComplexInfinity` rather than having `lambdify` silently accepting that expression, only for me to get "nan" during "runtime" when the function is used.

Therefore, it would be instructive if you could come up with a so called "minimal working example" which shows where this print method is needed.

amitport:
I also don't see the use case it was already there and was just trying to push a fix for the buggy parts.

bjodah:
Sorry, I don't quite follow. This PR only adds to the printers right?

So using SymPy's master branch, what would a small python program look like to trigger the error you are describing in the PR description?

amitport:
`lambdify([], S.ComplexInfinity, 'mpmath')` (or `'scipy` or `numpy`)

ThePauliPrinciple:
It can appear when dividing by zero

amitport:
@ThePauliPrinciple I have something less obvious (simplify over the density function of piecewise distribution... IDK it just appeared). In any case, the printer should not crash, or at least shouldn't crash with this exception.

sylee957:
I'm not sure if the printer itself crashes when encountering unknown function, I thought that it just attempts to print as unknown function.
It could be possible to crash when lambdify with that, but I think that the user should have the controllability of using `.subs(S.NaN, ...)` to convert to object that are representable.

amitport:
In any case, without this PR, the printer also crashes directly `mapmathprinter().doprint(S.ComplexInfinity)` with Inner keyerror which is not ideal.

oscargus:
I do not recall why I added that there. Probably realized that `ComplexInfinity` would not print properly and that there were no suitable representation so used `'nan'`. Maybe that was a bad idea.

bjodah:
@amitport thank you, so a minimal case would be:
```python
from sympy.printing.pycode import MpmathPrinter
from sympy import S
print(MpmathPrinter().doprint(S.ComplexInfinity))
```
I agree that it looks like the culprit is the addition of ComplexInfinity, let's remove it for now. I would also remove `_print_ComplexInfinity` completely when there is no representation. If a user wants to represent it by NaN, I think we can instruct them to do `.replace(ComplexInfinity, NaN)` prior to printing. Having NaN being produced is most often not useful (there are of course exceptions depending on how NaNs are treated by other interacting code). At least it will make the user aware that `NaN`s may be produced by their generated code. What do you think?

amitport:
I think having a nice error is reasonable. We can suggest replacing it with NaN or `.replace(ComplexInfinity, Infinity)` also sometimes makes sense.

.mailmap
```diff
@@ -157,6 +157,7 @@ Amit Kumar <dtu.amit@gmail.com>
 Amit Kumar <dtu.amit@gmail.com> <aktech@users.noreply.github.com>
 Amit Kumar <dtu.amit@gmail.com> <dtu.amit+gh@gmail.com>
 Amit Manchanda <amitdelhi1995@gmail.com>
+Amit Portnoy <amit.portnoy@gmail.com>
```
oscarbenjamin, line 160: The AUTHORS check fails with
```
Amit Portnoy <1131991+amitport@users.noreply.github.com>
```
You should change the line above in `.mailmap` to
```
Amit Portnoy <amit.portnoy@gmail.com> Amit Portnoy <1131991+amitport@users.noreply.github.com>
```
That tells the `bin/mailmap_check.py` script that the second name/email is an alias for the first.

amitport:
@oscarbenjamin  update authors list.

Note that this is the solution that adds ComplexInfinity to mpmath and numpy, just removing it from math does not work and seems to require a larger refactor of _print_ComplexInfinity:

If we remove it from math we still need to change the following:
```
    def _print_ComplexInfinity(self, expr):
        return self._print_NaN(S.NaN) # <-- change from return self._print_NaN(expr)
```
added the alternative solution as https://github.com/amitport/sympy/tree/23039-opt2

PR #23044, opened 2022-02-08
Series/limits :Added _eval_nseries and _eval_as_leading_term method for frac function

anutosh491:
I wanted to use `frac` function for calculating something else , when I realized the `frac` function performs quite poorly while calculating limits
Some wrong results returned on master
```
>>> x =Symbol('x')
>>> limit(frac(x), x, 0, '+')
frac(_w)
>>> limit(frac(x), x, 0, '-')
0
>>> limit(frac(x + 1), x, 0, '-')
0
>>> limit(frac(-x), x, 0, '+')
0
>>> limit(frac(-x), x, 0, '-')
frac(-_w)
>>> limit(frac(-2*x + 1), x, 0, '+')
0
>>> limit(frac(x), x, 0, '+-')
    raise ValueError("The limit does not exist since "
ValueError: The limit does not exist since left hand limit = 0 and right hand limit = frac(_w)
```
All these have been been fixed , also I have implemented the `_eval_nseries` method for `frac` 
On master we had , which isn't wrong but not what one would expect either
```
>>> frac(x).series(x, 0, 6, cdir=1)
x*Subs(Derivative(frac(x), x), x, 0) + x**2*Subs(Derivative(frac(x), (x, 2)), x, 0)/2 + x**3*Subs(Derivative(frac(x), (x, 3)), x, 0)/6 + x**4*Subs(Derivative(frac(x), (x, 4)), x, 0)/24 + x**5*Subs(Derivative(frac(x), (x, 5)), x, 0)/120 + O(x**6)
```
The results for `series/nseries` of `frac` function heavily depend on the direction/`cdir` in which the `series/nseries` is being calculated , hence I have added tests accordingly.
On committed branch , I have taken care of all these cases, I've added the order term too but we can get rid of that if we decide to, some examples
```
>>> frac(x).series(x)
x + O(x**6)
>>> frac(x)._eval_nseries(x, 6, logx = None, cdir = 1)
x + O(x**6)
>>> frac(x)._eval_nseries(x, 6, logx = None, cdir = 0)
x + O(x**6)
>>> frac(x)._eval_nseries(x, 6, logx = None, cdir = -1)
1 + x + O(x**6)
>>> frac(2*x + 1)._eval_nseries(x, 6, logx = None, cdir = 1)
2*x + O(x**6)
>>> frac(2*x + 2)._eval_nseries(x, 6, logx = None, cdir = 1)
2*x + O(x**6)
>>> frac(x**2)._eval_nseries(x, 6, logx = None, cdir = 1)
x**2 + O(x**6)
>>> frac(x**2)._eval_nseries(x, 6, logx = None, cdir = -1)
x**2 + O(x**6)
>>> frac(x**3 - 1)._eval_nseries(x, 6, logx = None, cdir = -1)
1 + x**3 + O(x**6)
>>> frac(x**3 - 1)._eval_nseries(x, 6, logx = None, cdir = 1)
x**3 + O(x**6)
>>> 
>>> frac(1/x).series(x, 0, 6)
AccumBounds(0, 1) + O(x**6)
>>> frac(1/x).series(x, 0, 0)
O(1)
```

#### Release Notes

* series
  * Added _eval_nseries and _eval_as_leading_term method for frac function.


anutosh491:
After implementing `nseries`,  `calcultaing_series` of `gruntz` started working cause it depends on `lseries` which in turn depends on `nseries`  , hence `gruntz(frac(x), x, oo)` started working !
Hence earlier on we had 
```
>>> gruntz(frac(x), x, oo)
Asymptotic expansion of frac around [oo] is not implemented.
```
Now we have
```
>>> gruntz(frac(x), x, oo)
AccumBounds(0, 1)
```
Now we already had a test i.e `limit(frac(x)**x, x, oo)` which fails because of the above Asymptotic expansion not implemented case and then gets solved by heurestics after failing gruntz .
Now gruntz isn't still capable to solve this but now that gruntz(frac(x), x, oo) works , the path for the above limit changes a bit , hence it finally depends on `sign(AccumBounds)` and there wasn't any case addressing this, hence I've addressed the case too !

anutosh491:
CC: @jksuom , I'm confident of this change , just wanted your views on the order term i've added . A random expression is better for an example ! Consider following examples 
```
>>> frac(x**3 - x).series(x,3, cdir = -1)
-77 + 9*(x - 3)**2 + (x - 3)**3 + 26*x + O((x - 3)**6, (x, 3))
>>> frac(x**3 - x).series(x,3, cdir = 1)
-78 + 9*(x - 3)**2 + (x - 3)**3 + 26*x + O((x - 3)**6, (x, 3))
>>>
>>> # floor rewrite for frac 
>>> (x**3 - x - floor(x**3 - x)).series(x,3, cdir = -1)
26*x + (x - 3)**3 + 9*(x - 3)**2 - 77
>>> (x**3 - x - floor(x**3 - x)).series(x,3, cdir = 1)
26*x + (x - 3)**3 + 9*(x - 3)**2 - 78
>>>
>>>
>>> frac(x**2 + 1).series(x,5, cdir = 1)
-50 + (x - 5)**2 + 10*x + O((x - 5)**6, (x, 5))
>>> frac(x**2 + 1).series(x,5, cdir = -1)
-49 + (x - 5)**2 + 10*x + O((x - 5)**6, (x, 5))
>>>
>>>
>>> ((x**2 + 1) - floor(x**2 + 1)).series(x,5, cdir = 1)
10*x + (x - 5)**2 - 50
>>> ((x**2 + 1) - floor(x**2 + 1)).series(x,5, cdir = -1)
10*x + (x - 5)**2 - 49
```
Hence I have no doubts for the leading term or the nseries implementation except the requirement of the order term !

sympy/functions/elementary/complexes.py
```diff
@@ -378,6 +379,13 @@ def eval(cls, arg):
                 return S.ImaginaryUnit
             if arg2.is_extended_negative:
                 return -S.ImaginaryUnit
+        if isinstance(arg, AccumBounds):
+            if arg.max < S.Zero:
+                return S.NegativeOne
+            elif arg.min > S.Zero:
+                return S.One
+            elif arg.min.is_zero or arg.max.is_zero or (arg.min < S.Zero and arg.max > S.Zero):
```
0sidharth, line 387: This could just be `else`

sympy/functions/elementary/integers.py
```diff
@@ -517,6 +517,53 @@ def _eval_is_zero(self):
...
+                                            "does not exist" % self)
+                    else:
+                        return r
+                else:
+                    ndir = arg.dir(x, cdir=cdir)
+                    return r + 1 if ndir.is_negative else r
+            else:
+                return r
+        elif arg0 is S.ComplexInfinity:
+            return AccumBounds(0, 1)
```
0sidharth, line 542: Maybe this should be returned in real infinities case too.

sympy/functions/elementary/integers.py
```diff
@@ -517,6 +517,53 @@ def _eval_is_zero(self):
...
         return False
 
+    def _eval_as_leading_term(self, x, logx=None, cdir=0):
+        from sympy.calculus.accumulationbounds import AccumBounds
+        arg = self.args[0]
+        arg0 = arg.subs(x, 0)
+        r = self.subs(x, 0)
+
+        if arg0.is_finite:
+            if arg0.is_integer or r.is_zero:
```
0sidharth, line 527: Is there some case where `arg0.is_integer` does not return `True` but `r.is_zero` returns `True`?

anutosh491:
Thanks @0sidharth I'll look into the reviews soon and try to address those !

sympy/series/tests/test_limits.py
```diff
@@ -220,6 +220,27 @@ def test_ceiling_requires_robust_assumptions():
     assert limit(ceiling(5 + cos(x)), x, 0, "-") == 6
 
 
+def test_frac():
+    assert limit(frac(x), x, oo) == AccumBounds(0, 1)
+    assert limit(frac(x)**x, x, oo) == AccumBounds(0, oo)  # wolfram gives (0, 1)
```
0sidharth, line 225: Does gruntz work for this limit? I don't think it will but curious

anutosh491: Without the block for `AccumBounds` in `sign` class , we would end up with something like the following
```
>>> gruntz(frac(x)**x, x, oo)
NotImplementedError: Result depends on the sign of sign(AccumBounds(-1, 0))
```
Hence it won't reach the heurestics method as only `PoleError/ValueError `exceptions are handled after result from `gruntz` . Hence it only makes sense to add a block to tackle `AccumBounds` in `sign` as `sign(AccumBounds(-1, 0))` does make sense if a single value can be returned ,Now we get the following

```
>>> gruntz(frac(x)**x, x, oo)
ValueError: More than one value is possible for passed argument sign(AccumBounds(-1, 0))
```
This also resolves the following
```
>>> sign(AccumBounds(-1, -0.5))
-1
>>> sign(AccumBounds(2, 3))
1
>>> sign(AccumBounds(-2, 3))
ValueError: More than one value is possible for passed argument sign(AccumBounds(-2, 3))
```

sympy/functions/elementary/integers.py
anutosh491: True , I hadn't thought about this but makes complete sense .

anutosh491:
CC : @jksuom

anutosh491:
@0sidharth I've addressed requested changes would be glad if you could have a look . Other than my doubt regarding the order term this looks good to me https://github.com/sympy/sympy/pull/23044#issuecomment-1034557503

anutosh491:
CC: @jksuom , Could you look into the pr whenever you find time ! I am somewhat satisfied with the changes and I've tried to cover all cases possible here . I feel this could be close to go in after a set of reviews from your sides. Thanks .

sympy/series/tests/test_nseries.py
```diff
@@ -444,6 +445,21 @@ def test_floor():
...
+    assert frac(2*x).series(x, 0, 6, cdir=-1) == 1 + 2*x + O(x**6)
+    assert frac(2*x + 1).series(x, 0, 6, cdir=1) == 2*x + O(x**6)
+    assert frac(2*x + 2).series(x, 0, 6, cdir=1) == 2*x + O(x**6)
+    assert frac(x**2).series(x, 0, 6, cdir=1) == x**2 + O(x**6)
+    assert frac(x**2).series(x, 0, 6, cdir=1) == x**2 + O(x**6)
+    assert frac(x**3 - 1).series(x, 0, 6, cdir=-1) == 1 + x**3 + O(x**6)
+    assert frac(x**3 - 1).series(x, 0, 6, cdir=1) == x**3 + O(x**6)
+    assert frac(x**8).series(x, 0, 6, cdir=1) == O(x**6)
+    assert frac(1/x).series(x, 0, 6) == AccumBounds(0, 1) + O(x**6)
+    assert frac(1/x).series(x, 0, 0).expr == S.One
```
0sidharth, line 460: What is the point of this test?

anutosh491: Not sure if we want that , we would end up with O(1).expr for which I've returned 1,but its not too necessary a test !

anutosh491:
CC : @jksuom for review ! Does this look good to you ?

anutosh491:
Ping @jksuom for review .Would be glad to hear from you on this pr .I am somewhat happy with the changes here !

anutosh491:
I will be adding another corner case as a test in my next commit .It's this one
```
>>> limit(frac(x)**(1/x), x, oo)
1 
```
But this is one of the corner cases addressed by one of my other pr's which is now merged (#23127) so the current commits give 1 as the answer but the correct answer is `AccumBounds(0, 1)` as the function steeply falls to `0` periodically . I had to merge master into this branch so that I could add this test .
image . This completes the set of exclusive limit test for `frac` from my side .

anutosh491:
Well I realized that `test/test1-latest` is still running because of `limit(frac(x)**x, x, oo)` which runs infinitely.
 `log(AccumBounds(0, x))` was added by me to give `AccumBounds(-oo, log(x))` in my latest merged Pr , which is not wrong by any means  . I am sure the expression is caught here infinitely as `gruntz(frac(x)**x, x, oo) `involves taking series/limit/leading_term at `log(frac(1/x))` etc and `frac(1/x).as_leading_term(x)` is nothing but `AccumBounds(0, 1)`.

I am not able to catch the error though! @0sidharth could you help me out here ? The latest master + the commit here would lead to the `limit(frac(x)**x, x, oo)` hanging infinitely

anutosh491:
Cool caught the error, The error due to which` test/test1-latest` ran more than 3 hours is as follows. I'll explain the changes made through the latest Pr here too ! 

**1)** **THE CHANGE IN GRUNTZ.PY**
So it couldn't process `limit(frac(x)**x, x, oo)` which indeed requires `gruntz(frac(x)**x, x, oo)`.Now this is because 
the `mrv` wasn't being calculated for it properly . The `mrv` we need to calculate here is  `mrv(exp(x*log(frac(x)), x)` 
Now as `gruntz algo` says 
image

So checking out the `e**g` case we know if `g` tends to infinite values as `x` tends to `oo` . We return  `max(exp(x*log(frac(x)), mrv(x*log(frac(x))), x)` . 

So what happens in line `295 -296` of gruntz.py is that its trying to formulate the above set of rules (i.e if `g` goes to infinity take the `max` otherwise just `mrv(g)`)
```
        li = limitinf(e.exp, x)
        if any(_.is_infinite for _ in Mul.make_args(li)):
            s1 = SubsSet()
            e1 = s1[e]
            s2, e2 = mrv(e.exp, x)
            su = s1.union(s2)[0]
            su.rewrites[e1] = exp(e2)
            return mrv_max3(s1, e1, s2, exp(e2), su, e1, x)
```
So what happens is one of my latest merged Prs (**23127**) improves the result for `li`
Earlier we had
```
>>> gruntz(x*log(frac(x)), x, oo)
oo*sign(log(AccumBounds(0, 1)))
>>> # Now after 23127 this gives
>>> gruntz(x*log(frac(x)), x, oo)
AccumBounds(-oo, 0)
```
Now we all would take the second result anyday , so I can confirm there's nothing wrong with that 
For reference 
image

So now we have `li` as `AccumBounds(-oo, 0)` and but obviously `if any(_.is_infinite for _ in Mul.make_args(li))` can't catch that li would be tending to `infinite` values as `x` tends to `oo` and as a result if won't compute the `max(......)` case and return `mrv(g, x)` which is not what we want I mean we want the `max` case here .
```
>>> mrv(frac(x)**x, x) # which says the mrv is just x
({x: _Dummy_82}, {}, exp(_Dummy_82*log(frac(_Dummy_82))))

>>> mrv(frac(x)**x, x) # which says the mrv is exponential in nature exp(x*log(frac(x))) 
({exp(x*log(frac(x))): _Dummy_50, x: _Dummy_51}, {_Dummy_50: exp(_Dummy_51*log(frac(_Dummy_51)))}, _Dummy_50)
```
Obviously the second case is the preferred one which also used to happen when ` 23127` wasn't in , cause the `li` back then was `oo*sign(log(AccumBounds(0, 1)))` and this catches that `g` will tend to `oo ` and hence takes the `max(....)` based `mrv`. So the gruntz change addresses based on the following idea . Hope the explanation clears this

anutosh491:
2) TESTS ADDED 
a) The corner case which was addressed in ` 23127` as also pointed above . Earlier we had 
```
>>> limit(frac(x)**(1/x), x, oo)
1
```
Now we have the following which is clearly the correct answer so here's an improvement I thought of expressing
```
>>> limit(frac(x)**(1/x), x, oo)
AccumBounds(0, 1)
```
image
Clearly falls to `0` steeply with infinite slope periodically . @jksuom would approve this as he helped me implement the corner case .

b) Other tests added in test_gruntz.py
I know gruntz isn't expected to answer limits for periodic cases but these work brilliantly due to the _eval_nseries implementation , hence I like these too so I've added them 
Earlier we had

```
>>> gruntz(log(frac(x)), x, oo)
log(AccumBounds(0, 1))
>>> gruntz(x*log(frac(x)), x, oo)
AccumBounds(-oo, 0)
oo*sign(log(AccumBounds(0, 1)))
```
Now we have 
```
>>> gruntz(frac(x), x, oo)
AccumBounds(0, 1)
>>> gruntz(log(frac(x)), x, oo)
AccumBounds(-oo, 0)
>>> gruntz(x*log(frac(x)), x, oo)
AccumBounds(-oo, 0)
```
Which are anyday better results !

anutosh491:
I am quite happy with the latest commit , would be glad if this could be reviewed @jksuom @0sidharth

anutosh491:
CC @jksuom for review ! This LGTM but would like to have your views on the same ! Fixes almost all limits and series involving `frac(x)` (OP shows how poorly frac used to perform) and works as good as the `floor / ceil` rewrite of `frac` eg (`x - floor(x)`) and better at some places as I could see while implementing this Pr.

anutosh491:
The latest commit has been explained here https://github.com/sympy/sympy/pull/23127#issuecomment-1049555078 and the comments which follow it !

anutosh491:
ping @jksuom for review !

anutosh491:
ping @jksuom @0sidharth for review . I am somewhat happy would with the changes here . Would be glad to hear from you'll . thanks !

anutosh491:
I really like the `if isinstance(arg, AccumBounds):` check here . For example on master we have

```
>>> limit(sign(sin(x) + 2), x, oo)
sign(AccumBounds(1, 3))
>>> limit(sign(sin(x)), x, oo)
sign(AccumBounds(-1, 1))
>>> limit(sign(sin(x) - 1), x, oo)
sign(AccumBounds(-3, -1))
```
But we should we returning `sign(AccumBounds(-1, 1))` for `limit(sign(sin(x)), x, oo)` , even wolfram gives indeterminant for this !
Cause the graph goes like this 
image
So on this branch we raise `ValueError` for this case 
```
>>> limit(sign(sin(x)), x, oo)
  File "C:\Users\anuto\sympy\sympy\sympy\series\limits.py", line 361, in doit
    raise PoleError()
sympy.core.function.PoleError

During handling of the above exception, another exception occurred:

  File "C:\Users\anuto\sympy\sympy\sympy\series\limits.py", line 361, in doit
    raise PoleError()
sympy.core.function.PoleError

During handling of the above exception, another exception occurred:

  File "C:\Users\anuto\sympy\sympy\sympy\functions\elementary\complexes.py", line 385, in eval
    raise ValueError('More than one value is possible for passed argument sign(%s)' % (arg))
ValueError: More than one value is possible for passed argument sign(AccumBounds(-1, 1))

```
And `gruntz(sign(sin(x) ), x, oo)` returns `S.NaN` resulting into the `PoleError` too !

The other cases also look better
image
```
>>> limit(sign(sin(x) + 2), x, oo)
1
>>> limit(sign(sin(x) - 2), x, oo)
-1

```

anutosh491:
ping @jksuom for review . I am somewhat happy would with the changes here . Haven't heard from you on this pr yet so would be glad to hear from your side  . thanks !

anutosh491:
Fixed in #24020 , hence closing .

PR #21318, opened 2021-04-14
Adding integration test for numba.jit and lambdify

abbyjng:
#### References to other Issues or PRs

Addresses #20070 

#### Brief description of what is fixed or changed
- Creates file `sympy/external/tests/test_numba.py` to test that `numba.jit` works properly with `lambdify()`.
- Adds numba to the list of optional dependencies in `.travis.yaml`.
- Adds numba to the list of travis tests in `bin/test_travis.sh`.

#### Release Notes

NO ENTRY

PR #20088, opened 2020-09-15
Added test for using numba.jit with lambdify

xavierdurawa:
Added a test to check that numba.jit is working
properly with lambdified functions.

#### References to other Issues or PRs

Addresses #20070

#### Brief description of what is fixed or changed
New test was added to ensure that numba.jit works
with Sympy's lambdified functions

#### Release Notes

NO ENTRY

PR #23050, opened 2022-02-09
Add a test for lambdify used with numba 

### LINKED ISSUE #20070: Add a test that lambdify works with numba
We should have a test in our test suite that tests that lambdify() works properly if you jit it with `numba.jit`. If we have that, then numba can add it to their integration tests to avoid potentially breaking it in the future (and we can also avoid breaking it too). 

See https://github.com/numba/numba/issues/2328

sumit-158:
Fixes #20070
Adding test in [test_lumbidify.py](https://github.com/sympy/sympy/blob/master/sympy/utilities/tests/test_timeutils.py) that work for numba 

#### Release Notes

NO ENTRY

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -148,7 +148,8 @@ jobs:
       - run: python -m pip install --upgrade pip
       - run: pip install mpmath gmpy2 matplotlib numpy scipy aesara ipython \
                          symengine cython llvmlite wurlitzer \
-                         autowrap numexpr 'antlr4-python3-runtime==4.7.*'
+                         autowrap numexpr 'antlr4-python3-runtime==4.7.*' \ 
+                         numba
       # Test external imports
       - run: bin/test_external_imports.py
       - run: bin/test_submodule_imports.py
--- a/.mailmap
+++ b/.mailmap
@@ -1102,6 +1102,7 @@ Subhash Saurabh <subhashsaurabh419@gmail.com>
 Sudarshan Kamath <sudarshan.kamath97@gmail.com>
 Sudeep Sidhu <sudeepmanilsidhu@gmail.com> sidhu1012 <sudeepmanilsidhu@gmail.com>
 Sudhanshu Mishra <mrsud94@gmail.com>
+Sumit Kumar <mr.sumitkrr@gmail.com> Sumit Kumar <96618001+sumit-158@users.noreply.github.com>
 Sumith Kulal <sumith1896@gmail.com>
 Sunny Aggarwal <sunnyaggarwal1994@gmail.com>
 Supreet Agrawal <supreet11agrawal@gmail.com> stm <supreet11agrawal@gmail.com>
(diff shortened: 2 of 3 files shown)


anutosh491:
> #### Other comments
> #### Release Notes
> NO ENTRY

You need to add release notes correctly for the checks to pass , if you're not sure how to do so refer to [guide on how to write release notes](https://github.com/sympy/sympy/wiki/Writing-Release-Notes)

anutosh491:
Also line `1356` has some trailing whitespaces which need to be removed .python bin/strip_whitespace followed by the file name 
`python bin/strip_whitespace sympy/utilities/tests/test_lambdify.py` is the command which would hep you here !

sumit-158:
> Thank you for the Help!!

anutosh491:
You would also need to update your name in the `./mailmap` file , otherwise the tests would fail . I'll attach a string of comments by one of the maintainers here , it seems to be a good guide for this

sumit-158:
@oscarbenjamin @oscargus Please review this PR.

oscargus:
This is the outcome:

```
test_issue_20070 numba not installed s
```
so numba should be installed as well. That is done by adding numba here:
https://github.com/sympy/sympy/blob/7cb6a0d7c67585441c93eeb89103390e156f0319/.github/workflows/runtests.yml#L149-L151

sumit-158:
Hey, @oscargus I have done the task please review it?

oscargus:
Seems to work, but it would be better to have it on the same line. Now numpy is first installed, removed and reinstalled.

On the other hand, this limits the numpy-version, so not sure how it should be prioritized.

.github/workflows/runtests.yml
```diff
@@ -149,6 +149,7 @@ jobs:
       - run: pip install mpmath gmpy2 matplotlib numpy scipy aesara ipython \
                          symengine cython llvmlite wurlitzer \
                          autowrap numexpr 'antlr4-python3-runtime==4.7.*'
+      - run: pip install numba
```
oscarbenjamin, line 152: This should just be added to the previous line

sumit-158: can you please review now!!

.github/workflows/runtests.yml
```diff
@@ -148,8 +148,8 @@ jobs:
       - run: python -m pip install --upgrade pip
       - run: pip install mpmath gmpy2 matplotlib numpy scipy aesara ipython \
                          symengine cython llvmlite wurlitzer \
-                         autowrap numexpr 'antlr4-python3-runtime==4.7.*'
-      - run: pip install numba
+                         autowrap numexpr 'antlr4-python3-runtime==4.7.*' \ 
+                         pip install numba
```
oscargus, line 152:
```suggestion
                         numba
```

oscargus: These are package names. We do not want to install `pip` (as it is already installed) and not install `install` (as it is not needed).

.github/workflows/runtests.yml
oscargus: Please refrain from using exclamation marks, especially multiple. If it took you ten days to fix it, why the hurry now?

oscargus:
Can you please rebase and squash the commits? Right now there are nine commits for a rather marginal change. Thanks!

oscargus:
Looks good! Thanks!

(We should probably also merge the more thorough tests from #21318 and #20088 in, with correct attribution.)

PR #16970, opened 2019-06-05
Added PERT distribution

ritesh99rakesh:
#### References to other Issues or PRs

Added PERT distribution to `sympy/stats/crv_types.py`

#### Release Notes

- stats
    - Added PERT distribution to `sympy/stats/crv_types.py`

Issue #16983, opened 2019-06-06
Expectation of PERT RV is taking forever


ritesh99rakesh:
I have the following implementation of PERT distribution (see PR (https://github.com/sympy/sympy/pull/16970) and https://github.com/sympy/sympy/pull/16970#discussion_r290782910)
```python3
class PERTDistribution(SingleContinuousDistribution):
    _argnames = ('a', 'b', 'c')

    @property
    def set(self):
        return Interval(self.a, self.c)

    @staticmethod
    def check(a, b, c):
        _value_check((b > a, b < c),
            "Parameter b must be in range (%s, %s). b = %s"%(a, c, b))
        _value_check((a < c),
            "Parameter a must be less than Parameter c. a = %s, c = %s"%(a, c))

    def pdf(self, x):
        a, b, c = self.a, self.b, self.c
        alpha = (4*b + c - 5*a)/(c - a)
        beta = (5*c - a - 4*b)/(c - a)
        num = (x - a)**(alpha - 1)*(c - x)**(beta - 1)
        den = beta_fn(alpha, beta)*(c - a)**5
        return num/den

    # def expectation(self, expr, var, **kwargs):
    #     a, b, c = self.args
    #     return (a + 4*b + c)/6
```
But the following takes forever
```python3
>>> from sympy.stats import *
>>> froms sympy import symbols
>>> a, b, c = symbols('a b c')
>>> X = PERT('x', a, b, c)
>>> E(X) # takes forever without explicit definition of method `expectation`
```
**Reason:** Might be an issue with integrals (see https://github.com/sympy/sympy/pull/16970#discussion_r291090305)

jksuom:
It returns immediately on my system:
```
Python 2.7.6 (default, Nov 13 2018, 12:45:42) 
[GCC 4.8.4] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from sympy.stats import *
>>> from sympy import symbols
>>> a, b, c = symbols('a, b, c')
>>> X = PERT('x', a, b, c)
>>> E(X)
a/6 + 2*b/3 + c/6
```

ritesh99rakesh:
> It returns immediately on my system:
> 
> ```
> Python 2.7.6 (default, Nov 13 2018, 12:45:42) 
> [GCC 4.8.4] on linux2
> Type "help", "copyright", "credits" or "license" for more information.
> >>> from sympy.stats import *
> >>> from sympy import symbols
[...]

PR #23051, opened 2022-02-09
stats - added PERT Distribution in crv_types.py

praneethratna:
Closes #16970
Added [PERT](https://en.wikipedia.org/wiki/PERT_distribution) distribution to crv_types.py.

#### Release Notes

* stats
   * added PERT distribution in crv_types.py


czgdp1807:
LGTM. Would like to have someone who can verify the formulae mentioned here. @Smit-create can you please do that?

sympy/stats/crv_types.py
```diff
@@ -3444,6 +3447,111 @@ def Pareto(name, xm, alpha):
...
+    def check(a, b, c):
+        _value_check(b > a, "Parameter b must be greater than a.")
+        _value_check(c > b, "Parameter c must be greater than b.")
+        _value_check(c > a, "Parameter c must be greater than a.")
+
+    def pdf(self, x):
+        a, b, c = self.a, self.b, self.c
+        alpha = 1 + 4*(b - a/(c - a))
+        beta = 1 + 4*(c - b/(c - a))
+        num = (x - a)**(alpha - 1)*(c - x)**(beta - 1)
```
Smit-create, line 3472:
```suggestion
        alpha = 1 + 4*(b - a)/(c - a)
        beta = 1 + 4*(c - b)/(c - a)
        num = (x - a)**(alpha - 1)*(c - x)**(beta - 1)
```

sympy/stats/crv_types.py
```diff
@@ -3444,6 +3447,111 @@ def Pareto(name, xm, alpha):
...
+    def _cdf(self, x):
+        a, b, c = self.a, self.b, self.c
+        alpha = 1 + 4*(b - a/(c - a))
+        beta = 1 + 4*(c - b/(c - a))
+        z = (x - a)/(c - a)
+        return betainc_regularized(alpha, beta, 0, z)
+
+    def expectation(self, expr, var, **kwargs):
+        a, b, c = self.a, self.b, self.c
+        return (a + 4*b + c)/6
```
Smit-create, line 3485: We should generally avoid overriding `expectation` and let sympy calculate on its own. See the issue due to which the PR in OP was stalled: https://github.com/sympy/sympy/issues/16983

sympy/stats/crv_types.py
praneethratna: I'm not sure why making this change is causing the `density(X)(z)` to fail.

czgdp1807: I think you need to update the tests (the ones you have added) as well.

sympy/stats/crv_types.py
Smit-create: You can actually check that by removing these 3 lines and see if it is able to find the expectation.

praneethratna: No, sympy is unable to find expectation automatically. Should i close this PR then?

czgdp1807: Well, if SymPy is unable to calculate `expectation` automatically, then that's a bit non-ideal. Adding a method to directly obtain `expectation` in the PERT's distribution class is fine by me. 
@Smit-create What do you think?

PR #23054, opened 2022-02-09
integrals - fixed issue#19575 and now Meijerint does not fail with float input

### LINKED ISSUE #19575: Meijerint fails with float input
From stackoverflow:
https://stackoverflow.com/questions/62427258/sympy-typeerror-cannot-interpret-float-object-as-integer/62427652#62427652

```julia
In [63]: Integral(0.0625*x**3.0*exp(-x/2), (x, 0, Rn))                                                                                                        
Out[63]: 
Rn                    
⌠                     
⎮               -x    
⎮               ───   
⎮          3.0   2    
⎮  0.0625⋅x   ⋅ℯ    dx
⌡                     
0                     

In [64]: Integral(0.0625*x**3.0*exp(-x/2), (x, 0, Rn)).doit()                                                                                                 
<ipython-input-64-af398b95e17a> in <module>
----> 1 Integral(0.0625*x**3.0*exp(-x/2), (x, 0, Rn)).doit()

~/current/sympy/sympy/sympy/integrals/integrals.py in doit(self, **hints)
    591                     antideriv = None
    592                 else:
--> 593                     antideriv = self._eval_integral(
    594                         function, xab[0], **eval_kwargs)
    595                     if antideriv is None and meijerg is True:

~/current/sympy/sympy/sympy/integrals/integrals.py in _eval_integral(self, f, x, meijerg, risch, manual, heurisch, conds)
   1067                 # rewrite using G functions
   1068                 try:
-> 1069                     h = meijerint_indefinite(g, x)
   1070                 except NotImplementedError:
   1071                     from sympy.integrals.meijerint import _debug

~/current/sympy/sympy/sympy/integrals/meijerint.py in meijerint_indefinite(f, x)
   1619     results = []
   1620     for a in sorted(_find_splitting_points(f, x) | {S.Zero}, key=default_sort_key):
-> 1621         res = _meijerint_indefinite_1(f.subs(x, x + a), x)
   1622         if not res:
   1623             continue

~/current/sympy/sympy/sympy/integrals/meijerint.py in _meijerint_indefinite_1(f, x)
   1685         else:
   1686             place = None
-> 1687         r = hyperexpand(r.subs(t, a*x**b), place=place)
   1688 
   1689         # now substitute back

~/current/sympy/sympy/sympy/simplify/hyperexpand.py in hyperexpand(f, allow_hyper, rewrite, place)
   2489         if not r.has(nan, zoo, oo, -oo):
   2490             return r
-> 2491     return f.replace(hyper, do_replace).replace(meijerg, do_meijer)

~/current/sympy/sympy/sympy/core/basic.py in replace(self, query, value, map, simultaneous, exact)
   1519             return expr
   1520 
-> 1521         rv = walk(self, rec_replace)
   1522         return (rv, mapping) if map else rv
   1523 

~/current/sympy/sympy/sympy/core/basic.py in walk(rv, F)
   1503                                 if rv == e and e != newargs[i]:
   1504                                     return rv
-> 1505                 rv = F(rv)
   1506             return rv
   1507 

~/current/sympy/sympy/sympy/core/basic.py in rec_replace(expr)
   1512             result = _query(expr)
   1513             if result or result == {}:
-> 1514                 v = _value(expr, result)
   1515                 if v is not None and v != expr:
   1516                     if map:

~/current/sympy/sympy/sympy/core/basic.py in <lambda>(expr, result)
   1441                 _value = lambda expr, result: value(*expr.args)
   1442             elif callable(value):
-> 1443                 _value = lambda expr, result: value(*expr.args)
   1444             else:
   1445                 raise TypeError(

~/current/sympy/sympy/sympy/simplify/hyperexpand.py in do_meijer(ap, bq, z)
   2485 
   2486     def do_meijer(ap, bq, z):
-> 2487         r = _meijergexpand(G_Function(ap[0], ap[1], bq[0], bq[1]), z,
   2488                    allow_hyper, rewrite=rewrite, place=place)
   2489         if not r.has(nan, zoo, oo, -oo):

~/current/sympy/sympy/sympy/simplify/hyperexpand.py in _meijergexpand(func, z0, allow_hyper, rewrite, place)
   2363 
   2364     t = Dummy('t')
-> 2365     slater1, cond1 = do_slater(func.an, func.bm, func.ap, func.bq, z, z0)
   2366 
   2367     def tr(l):

~/current/sympy/sympy/sympy/simplify/hyperexpand.py in do_slater(an, bm, ap, bq, z, zfinal)
   2296                 #      t*k ... we are using polar numbers for consistency!
   2297                 premult = (t/k)**bh
-> 2298                 hyp = _hyperexpand(Hyper_Function(nap, nbq), harg, ops,
   2299                                    t, premult, bh, rewrite=None)
   2300                 res += fac * hyp

~/current/sympy/sympy/sympy/simplify/hyperexpand.py in _hyperexpand(func, z, ops0, z0, premult, prem, rewrite)
   2017     # Try to recognise a shifted sum.
   2018     p = S.Zero
-> 2019     res = try_shifted_sum(func, z0)
   2020     if res is not None:
   2021         func, nops, p = res

~/current/sympy/sympy/sympy/simplify/hyperexpand.py in try_shifted_sum(func, z)
   1671 
   1672     ops = []
-> 1673     for n in range(r - 1):
   1674         ops.append(ShiftA(n + 1))
   1675     ops.reverse()

~/current/sympy/sympy/sympy/simplify/hyperexpand.py in try_shifted_sum(func, z)
   1671 
   1672     ops = []
-> 1673     for n in range(r - 1):
   1674         ops.append(ShiftA(n + 1))
   1675     ops.reverse()

TypeError: 'Float' object cannot be interpreted as an integer
```

praneethratna:
Fixes #19575 and now `Integral(0.0625*x**3.0*exp(-x/2), (x, 0, Rn)).doit()` returns 
`-0.125*Rn**3*exp(-Rn/2) - 0.75*Rn**2*exp(-Rn/2) - 3.0*Rn*exp(-Rn/2) + 6.0 - 6.0*exp(-Rn/2)` instead of 
`TypeError: 'Float' object cannot be interpreted as an integer`.

#### Release Notes

* simplify
    * fixed a bug in `hyperexpand.py` and now float cannot be interpreted as integer is not raised for some integrals.

### DIFF
--- a/sympy/simplify/hyperexpand.py
+++ b/sympy/simplify/hyperexpand.py
@@ -79,6 +79,8 @@ from sympy.polys import apart, poly, Poly
 from sympy.series import residue
 from sympy.simplify.powsimp import powdenest
 from sympy.utilities.iterables import sift
+from sympy.utilities.misc import as_int
+
 
 # function to define "buckets"
 def _mod1(x):
@@ -1661,12 +1663,12 @@ def try_shifted_sum(func, z):
     abuckets, bbuckets = sift(func.ap, _mod1), sift(func.bq, _mod1)
     if len(abuckets[S.Zero]) != 1:
         return None
-    r = int(round(abuckets[S.Zero][0]))
+    r = as_int(abuckets[S.Zero][0])  # XXX if this fails, is round necessary?
     if r <= 0:
         return None
     if S.Zero not in bbuckets:
         return None
-    k = int(round(min(bbuckets[S.Zero])))
+    k = as_int(min(bbuckets[S.Zero]))  # XXX if this fails, is round necessary?
     if k <= 0:
         return None
 
@@ -2330,8 +2332,10 @@ def _meijergexpand(func, z0, allow_hyper=False, rewrite='default',
                 s = Dummy('s')
                 integrand = z**s
                 for b in bm:
-                    if not Mod(b, 1) and b.is_Number:
-                        b = int(round(b))
+                    if b.is_Number:
+                        ib = int(b)
+                        if ib == b:
+                            b = ib
                     integrand *= gamma(b - s)
                 for a in an:
                     integrand *= gamma(1 - a + s)
--- a/sympy/simplify/hyperexpand.py
+++ b/sympy/simplify/hyperexpand.py
@@ -1661,14 +1661,14 @@ def try_shifted_sum(func, z):
     abuckets, bbuckets = sift(func.ap, _mod1), sift(func.bq, _mod1)
     if len(abuckets[S.Zero]) != 1:
         return None
-    r = abuckets[S.Zero][0]
+    r = int(round(abuckets[S.Zero][0]))
     if r <= 0:
         return None
     if S.Zero not in bbuckets:
         return None
     l = list(bbuckets[S.Zero])
     l.sort()
-    k = l[0]
+    k = int(round(l[0]))
     if k <= 0:
         return None
 
--- a/sympy/simplify/hyperexpand.py
+++ b/sympy/simplify/hyperexpand.py
@@ -1666,9 +1666,7 @@ def try_shifted_sum(func, z):
         return None
     if S.Zero not in bbuckets:
         return None
-    l = list(bbuckets[S.Zero])
-    l.sort()
-    k = int(round(l[0]))
+    k = int(round(min(bbuckets[S.Zero])))
     if k <= 0:
         return None
 
--- a/sympy/simplify/hyperexpand.py
+++ b/sympy/simplify/hyperexpand.py
@@ -1757,10 +1757,10 @@ def try_polynomial(func, z):
     if not al0:
         return None
 
-    a = int(al0[-1]) # make int in case `a` is Float equal to int
+    a = al0[-1] # make int in case `a` is Float equal to int
     fac = 1
     res = S.One
-    for n in Tuple(*list(range(-a))):
+    for n in Tuple(*list(range(-int(a)))):
         fac *= z
         fac /= n + 1
         for a in func.ap:
--- a/sympy/simplify/hyperexpand.py
+++ b/sympy/simplify/hyperexpand.py
@@ -83,6 +83,40 @@ from sympy.utilities.iterables import sift
 
 # function to define "buckets"
 def _mod1(x):
+    """returns additive constant, mod 1
+
+    Examples
+    ========
+
+    >>> from sympy import S, Symbol
+    >>> from sympy.simplify.hyperexpand import _mod1 as m1
+    >>> from sympy.abc import x
+
+    When the input is equal to an integer, the return value
+    will be S.Zero:
+
+    >>> m1(S(3)) is m1(S(3.0)) is S.Zero
+    True
+    >>> m1(3*S.Half)
+    1/2
+    >>> m1(S(1.5))
+    0.500000000000000
+
+    Currently, integer expressions are not reduced to 0.
+
+    >>> i = Symbol('i', integer=True)
+    >>> m1(2*i)
+    2*i
+
+    For expressions with free symbols, only the additive
+    constant is modified:
+
+    >>> m1(2*x) == m1(2*x + 3) == 2*x
+    True
+    >>> m1(2*x + S(3)/2)
+    2*x + 1/2
+    """
+
     # TODO see if this can work as Mod(x, 1); this will require
     # different handling of the "buckets" since these need to
     # be sorted and that fails when there is a mixture of
@@ -544,12 +578,9 @@ class Hyper_Function(Expr):
         abuckets, bbuckets = sift(self.ap, _mod1), sift(self.bq, _mod1)
 
         def tr(bucket):
-            bucket = list(bucket.items())
-            if not any(isinstance(x[0], Mod) for x in bucket):
-                bucket.sort(key=lambda x: default_sort_key(x[0]))
-            bucket = tuple([(mod, len(values)) for mod, values in bucket if
+            bucket = sorted(list(bucket.items()), key=lambda x: default_sort_key(x[0]))
+            return tuple([(mod, len(values)) for mod, values in bucket if
                     values])
-            return bucket
 
         return (self.gamma, tr(abuckets), tr(bbuckets))
 
@@ -1662,12 +1693,12 @@ def try_shifted_sum(func, z):
     abuckets, bbuckets = sift(func.ap, _mod1), sift(func.bq, _mod1)
     if len(abuckets[S.Zero]) != 1:
         return None
-    r = int(abuckets[S.Zero][0])
+    r = abuckets[S.Zero][0]
     if r <= 0:
         return None
     if S.Zero not in bbuckets:
         return None
-    k = int(min(bbuckets[S.Zero]))
+    k = min(bbuckets[S.Zero])
     if k <= 0:
         return None
 
@@ -1675,6 +1706,12 @@ def try_shifted_sum(func, z):
     nap.remove(r)
     nbq = list(func.bq)
     nbq.remove(k)
+
+    # r and k might be floats equal to integers
+    # so convert them explicitly to ints now
+    r = int(r)
+    k = int(k)
+
     k -= 1
     nap = [x - k for x in nap]
     nbq = [x - k for x in nbq]
@@ -1720,7 +1757,7 @@ def try_polynomial(func, z):
     if not al0:
         return None
 
-    a = al0[-1]
+    a = int(al0[-1]) # make int in case `a` is Float equal to int
     fac = 1
     res = S.One
     for n in Tuple(*list(range(-a))):
--- a/sympy/simplify/hyperexpand.py
+++ b/sympy/simplify/hyperexpand.py
@@ -79,7 +79,6 @@ from sympy.polys import apart, poly, Poly
 from sympy.series import residue
 from sympy.simplify.powsimp import powdenest
 from sympy.utilities.iterables import sift
-from sympy.utilities.misc import as_int
 
 
 # function to define "buckets"
@@ -1663,12 +1662,12 @@ def try_shifted_sum(func, z):
     abuckets, bbuckets = sift(func.ap, _mod1), sift(func.bq, _mod1)
     if len(abuckets[S.Zero]) != 1:
         return None
-    r = as_int(abuckets[S.Zero][0], strict=False)  # XXX if this fails, is round necessary?
+    r = int(abuckets[S.Zero][0])
     if r <= 0:
         return None
     if S.Zero not in bbuckets:
         return None
-    k = as_int(min(bbuckets[S.Zero]), strict=False)  # XXX if this fails, is round necessary?
+    k = int(min(bbuckets[S.Zero]))
     if k <= 0:
         return None
 
@@ -2332,7 +2331,7 @@ def _meijergexpand(func, z0, allow_hyper=False, rewrite='default',
                 s = Dummy('s')
                 integrand = z**s
                 for b in bm:
-                    if b.is_Number:
+                    if b.is_Number:  # XXX is it necessary to convert to int?
                         ib = int(b)
                         if ib == b:
                             b = ib
--- a/sympy/simplify/hyperexpand.py
+++ b/sympy/simplify/hyperexpand.py
@@ -62,7 +62,7 @@ from functools import reduce
 
 from sympy import SYMPY_DEBUG
 from sympy.core import (S, Dummy, symbols, sympify, Tuple, expand, I, pi, Mul,
-    EulerGamma, oo, zoo, expand_func, Add, nan, Expr, Rational)
+    EulerGamma, oo, zoo, expand_func, Add, nan, Expr, Rational, Integer)
 from sympy.core.mod import Mod
 from sympy.core.sorting import default_sort_key
 from sympy.functions import (exp, sqrt, root, log, lowergamma, cos,
@@ -1707,17 +1707,12 @@ def try_shifted_sum(func, z):
     nbq = list(func.bq)
     nbq.remove(k)
 
-    # r and k might be floats equal to integers
-    # so convert them explicitly to ints now
-    r = int(r)
-    k = int(k)
-
     k -= 1
     nap = [x - k for x in nap]
     nbq = [x - k for x in nbq]
 
     ops = []
-    for n in range(r - 1):
+    for n in range(int(r) - 1):
         ops.append(ShiftA(n + 1))
     ops.reverse()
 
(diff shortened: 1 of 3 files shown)


sympy/simplify/hyperexpand.py
```diff
@@ -1661,14 +1661,14 @@ def try_shifted_sum(func, z):
...
-    r = abuckets[S.Zero][0]
+    r = int(round(abuckets[S.Zero][0]))
     if r <= 0:
         return None
     if S.Zero not in bbuckets:
         return None
     l = list(bbuckets[S.Zero])
     l.sort()
-    k = l[0]
+    k = int(round(l[0]))
```
smichr, line 1671: `nbp.remove(k)` appears below -- isn't it possible that `k` will not be there? or are you just getting ints as floats (in which case `round` is not necessary?

smichr: It also does not appear that sort is necessary -- can we just use `k = int(round(min(bbuckets[S.Zero])))` instead?

praneethratna: > or are you just getting ints as floats (in which case `round` is not necessary?
Yeah, it should `Int` but I'm getting a `Float`.

praneethratna: > It also does not appear that sort is necessary -- can we just use `k = int(round(min(bbuckets[S.Zero])))` instead?

Yeah, seems like we can.

praneethratna: > `nbp.remove(k)` appears below -- isn't it possible that `k` will not be there? 

I don't think it is possible that `k` will not be there.

> or are you just getting ints as floats (in which case `round` is not necessary?

Yeah, it should `Int` but I'm getting a `Float`.

smichr: If you don't use `round` then an error will be raised if k is not present during the `remove` attempt. You could put an inline note at that line saying that if this raises then we have to look at whether round should be used to calculate k and the raw k kept for removal purposes.

oscarbenjamin: What is the value of k before this line?

smichr:
@praneethratna , see what you think of the last commit that I added.

sympy/simplify/hyperexpand.py
```diff
@@ -1661,14 +1663,12 @@ def try_shifted_sum(func, z):
     abuckets, bbuckets = sift(func.ap, _mod1), sift(func.bq, _mod1)
     if len(abuckets[S.Zero]) != 1:
         return None
-    r = abuckets[S.Zero][0]
+    r = as_int(abuckets[S.Zero][0])  # XXX if this fails, is round necessary?
```
smichr, line 1666:
```suggestion
    r = as_int(abuckets[S.Zero][0], strict=False)  # XXX if this fails, is round necessary?
```

sympy/simplify/hyperexpand.py
```diff
@@ -1661,14 +1663,12 @@ def try_shifted_sum(func, z):
...
     if r <= 0:
         return None
     if S.Zero not in bbuckets:
         return None
-    l = list(bbuckets[S.Zero])
-    l.sort()
-    k = l[0]
+    k = as_int(min(bbuckets[S.Zero]))  # XXX if this fails, is round necessary?
```
smichr, line 1671:
```suggestion
    k = as_int(min(bbuckets[S.Zero]), strict=False)  # XXX if this fails, is round necessary?
```

smichr:
OK, look at what is being retrieved from abuckets and bbuckets: key S.Zero which means that Mod(x, 1) == 0. This will only contain integers (perhaps as floats) since the symbolic values which might give 0 are separated so if there were something like `3 + Symbol('i', integer=True)` it would be returned as `Symbol('i', integer=True)` -- only the coefficient would be removed. So `round` is not necessary.

sympy/simplify/hyperexpand.py
```diff
@@ -1661,14 +1663,12 @@ def try_shifted_sum(func, z):
     abuckets, bbuckets = sift(func.ap, _mod1), sift(func.bq, _mod1)
     if len(abuckets[S.Zero]) != 1:
         return None
-    r = abuckets[S.Zero][0]
+    r = as_int(abuckets[S.Zero][0], strict=False)  # XXX if this fails, is round necessary?
```
smichr, line 1666:
```suggestion
    r = int(abuckets[S.Zero][0])
```

sympy/simplify/hyperexpand.py
```diff
@@ -79,6 +79,8 @@
 from sympy.series import residue
 from sympy.simplify.powsimp import powdenest
 from sympy.utilities.iterables import sift
+from sympy.utilities.misc import as_int
```
smichr, line 82:
```suggestion
```

sympy/simplify/hyperexpand.py
```diff
@@ -1661,14 +1663,12 @@ def try_shifted_sum(func, z):
...
-    r = abuckets[S.Zero][0]
+    r = as_int(abuckets[S.Zero][0], strict=False)  # XXX if this fails, is round necessary?
     if r <= 0:
         return None
     if S.Zero not in bbuckets:
         return None
-    l = list(bbuckets[S.Zero])
-    l.sort()
-    k = l[0]
+    k = as_int(min(bbuckets[S.Zero]), strict=False)  # XXX if this fails, is round necessary?
```
smichr, line 1671:
```suggestion
    k = int(min(bbuckets[S.Zero]))
```

sympy/simplify/hyperexpand.py
```diff
@@ -2332,8 +2332,10 @@ def do_slater(an, bm, ap, bq, z, zfinal):
                 s = Dummy('s')
                 integrand = z**s
                 for b in bm:
-                    if not Mod(b, 1) and b.is_Number:
-                        b = int(round(b))
+                    if b.is_Number:
```
smichr, line 2335:
```suggestion
                    if b.is_Number:  # XXX is it necessary to convert to int?
```

smichr:
The only warning I would raise is that `_mod1` would be better to return `S.Zero` in the case that `Mod(x, 1) == 0` else we are using `S.Zero` to retrieve key `Float(0)` in code that does such retrieval from a sift using `_mod1`.

smichr:
I made the `_mod1` output for numbers be S.Zero in case `x == int(x)`.

smichr:
Will commit after 12h unless there is any correction needed (or sooner if anyone reviews).

smichr:
What do you think about this:
```diff
diff --git a/sympy/simplify/hyperexpand.py b/sympy/simplify/hyperexpand.py
index ccf0d670c3..3f92ff557e 100644
--- a/sympy/simplify/hyperexpand.py
+++ b/sympy/simplify/hyperexpand.py
@@ -83,6 +83,39 @@
 
 # function to define "buckets"
 def _mod1(x):
+    """returns additive constant, mod 1
+
+    Examples
+    ========
+
+    >>> from sympy import S, Symbol
+    >>> from sympy.simplify.hyperexpand import _mod1 as m1
+    >>> from sympy.abc import x
+
+    When the input is equal to an integer, the return value
+    will be S.Zero:
+
+    >>> m1(S(3)) is m1(S(3.0)) is S.Zero
+    True
+    >>> m1(3*S.Half)
+    1/2
+    >>> m1(S(1.5))
+    0.500000000000000
+
+    Currently, integer expressions are not reduced to 0.
+
+    >>> i = Symbol('i', integer=True)
+    >>> m1(2*i)
+    2*i
+
+    For expressions with free symbols, only the additive
+    constant is modified:
+
+    >>> m1(2*x) == m1(2*x + 3) == 2*x
+    True
+    >>> m1(2*x + S(3)/2)
+    2*x + 1/2
+    """
     # TODO see if this can work as Mod(x, 1); this will require
     # different handling of the "buckets" since these need to
     # be sorted and that fails when there is a mixture of
@@ -544,12 +577,9 @@ def build_invariants(self):
         abuckets, bbuckets = sift(self.ap, _mod1), sift(self.bq, _mod1)
 
         def tr(bucket):
-            bucket = list(bucket.items())
-            if not any(isinstance(x[0], Mod) for x in bucket):
-                bucket.sort(key=lambda x: default_sort_key(x[0]))
-            bucket = tuple([(mod, len(values)) for mod, values in bucket if
-                    values])
-            return bucket
+            items = sorted(bucket.items(), key=lambda x: x[0])
+            return tuple([(mod, len(values)) for mod, values in
+                items if values])
 
         return (self.gamma, tr(abuckets), tr(bbuckets))
 
@@ -1662,19 +1692,27 @@ def try_shifted_sum(func, z):
     abuckets, bbuckets = sift(func.ap, _mod1), sift(func.bq, _mod1)
     if len(abuckets[S.Zero]) != 1:
         return None
-    r = int(abuckets[S.Zero][0])
-    if r <= 0:
-        return None
+    else:
+        [r] = abuckets[S.Zero]
+        if r <= 0:
+            return None
     if S.Zero not in bbuckets:
         return None
-    k = int(min(bbuckets[S.Zero]))
-    if k <= 0:
-        return None
+    else:
+        k = min(bbuckets[S.Zero])
+        if k <= 0:
+            return None
 
     nap = list(func.ap)
     nap.remove(r)
     nbq = list(func.bq)
     nbq.remove(k)
+
+    # r and k might be floats equal to integers
+    # so convert them explicitly to ints now
+    r = int(r)
+    k = int(k)
+
     k -= 1
     nap = [x - k for x in nap]
     nbq = [x - k for x in nbq]
@@ -1720,7 +1758,7 @@ def try_polynomial(func, z):
     if not al0:
         return None
 
-    a = al0[-1]
+    a = int(al0[-1])  # make int in case `a` is Float equal to int
     fac = 1
     res = S.One
     for n in Tuple(*list(range(-a))):
```

praneethratna:
> What do you think about this:

This change is working , except i made few changes so that all tests pass.

sympy/simplify/hyperexpand.py
```diff
@@ -1720,7 +1757,7 @@ def try_polynomial(func, z):
     if not al0:
         return None
 
-    a = al0[-1]
+    a = int(al0[-1]) # make int in case `a` is Float equal to int
```
smichr, line 1760: Could you find a test that passes through here and change one or more of its numerical values to Floats to see that this works?

praneethratna: I'm unable to find such a case, but is that change required? The issue is fixed even without with change.

smichr: If you can't find a test that passes through there then we don't even know it is tested. And if you do find one that passes through, I'm pretty sure I can cause it to fail by replacing ints with floats.

smichr: Yes...and for `r` and `k`.

smichr: I set up a trap for a test passing through the polynomial case.

praneethratna: > Yes...and for `r` and `k`.

The test cases are not passing, also the OP issue now returns,     
```
  File "/home/praneeth/sympy/sympy/polys/polyutils.py", line 211, in _parallel_dict_from_expr_if_gens
    monom[indices[base]] = exp
KeyError: _t**4.0
```

smichr:
Do you know if you have a test for the case when `k` is not an int?

praneethratna:
This works for OP issue,
```diff
diff --git a/sympy/simplify/hyperexpand.py b/sympy/simplify/hyperexpand.py
index 0f10514b5b..c6f438d3ee 100644
--- a/sympy/simplify/hyperexpand.py
+++ b/sympy/simplify/hyperexpand.py
@@ -1715,7 +1715,7 @@ def try_shifted_sum(func, z):
     for n in range(int(r) - 1):
         ops.append(ShiftA(n + 1))
     ops.reverse()
-
+    k = int(k)
     fac = factorial(k)/z**k
     for a in nap:
         fac /= rf(a, k)
@@ -1725,7 +1725,7 @@ def try_shifted_sum(func, z):
     ops += [MultOperator(fac)]

     p = 0
-    for n in range(int(k)):
+    for n in range(k):
         m = z**n/factorial(n)
         for a in nap:
             m *= rf(a, n)
```
but this is giving an error,
```diff
diff --git a/sympy/simplify/hyperexpand.py b/sympy/simplify/hyperexpand.py
index 0f10514b5b..df9ce9eee7 100644
--- a/sympy/simplify/hyperexpand.py
+++ b/sympy/simplify/hyperexpand.py
@@ -1715,8 +1715,8 @@ def try_shifted_sum(func, z):
     for n in range(int(r) - 1):
         ops.append(ShiftA(n + 1))
     ops.reverse()
-
     fac = factorial(k)/z**k
+    k = int(k)
     for a in nap:
         fac /= rf(a, k)
     for b in nbq:
@@ -1725,7 +1725,7 @@ def try_shifted_sum(func, z):
     ops += [MultOperator(fac)]

     p = 0
-    for n in range(int(k)):
+    for n in range(k):
         m = z**n/factorial(n)
         for a in nap:
             m *= rf(a, n)
```

smichr:
> This works for OP issue,

OK, this is where we have to be careful. I don't think we should just convert to int because that can changed the output (giving an int answer instead of a float). This needs to be investigated to see what that monom is if r and k are ints  -- is it `_t` or _t**4`?

praneethratna:
> is it `_t` or _t**4`?

It is` _t**4` when `r` and `k` are ints.

smichr:
```python
>>> Poly(x**4,x)
Poly(x**4, x, domain='ZZ')
>>> Poly(x**4.,x)
```
Poly doesn't like float exp. If the returned poly would be `1.0*_t**4` then that would be ok to change the exponent to int.

oscarbenjamin:
It's unclear to me what any of these changes do. This needs more explanation including examples that clearly show what the values of these variables are and how the changes affect them.

praneethratna:
> If the returned poly would be `1.0*_t**4` then that would be ok to change the exponent to int.

So, how could we fix that here?

smichr:
> So, how could we fix that here?

`integrate` could raise an error if any exponent was a float. Although we can integrate `3.14*x**2 -> 3.14/3*x**3` we can't `symbolically` integrate `3.14*x**2.0` but we can numerically do so. So `Integral(3.14*x**2.0, x)` is ok (since it could be evaluated with `.evalf`). @jksuom or someone else would have to check me here.

I don't know enough about the structure of the hyperexpansion function to know what corresponds to the exponents to know where to raise errors in those routines (and at what level).

sympy/simplify/hyperexpand.py
```diff
@@ -2332,8 +2364,10 @@ def do_slater(an, bm, ap, bq, z, zfinal):
                 s = Dummy('s')
                 integrand = z**s
                 for b in bm:
-                    if not Mod(b, 1) and b.is_Number:
-                        b = int(round(b))
+                    if b.is_Number:  # XXX is it necessary to convert to int?
```
jksuom, line 2367: gamma function does not like floating point numbers that are integers in the sense that Mod(x, 1) is zero. So those should be converted to (first int's, and then at some point) Integers. Other numbers should not be converted so the Mod(b, 1) is needed.

PR #23065, opened 2022-02-12
refactoring with var unpack

idiomaticrefactoring:
refactoring code with var unpack which is more pythonic, concise, readable and efficient; how do think this change which has practical value？

#### Release Notes

NO ENTRY


oscargus:
I agree with @oscarbenjamin 

The problem right now is that any changes there are not tested.

But in general, I am positive to this type of changes, just not in RUBI.

oscarbenjamin:
I'm going to close this. Thank you @zjzh for the contribution but it is best to leave this particular part of the code base alone for now.

PR #23068, opened 2022-02-12
physics : Added solve_vector function in physics vector

### LINKED ISSUE #18692: Eq not working with vectors
I want to set up a vector relation and be able to solve for a certain variable.

    from sympy.physics.vector import *
    from sympy import * 

    t,l = symbols("t l")
    R = ReferenceFrame('R')
    eq1 = Eq(5*R.x + t*R.y, l*R.x + 4*R.y
    solve(eq1,t)

Then I get this error. 

      File "stackoverflow_test.py", line 6, in <module>
        Eq(5*R.x + t*R.y, l*R.x + 4*R.y)
      File "/usr/lib/python3.8/site-packages/sympy/core/relational.py", line 474, in __new__
        lhs = _sympify(lhs)
      File "/usr/lib/python3.8/site-packages/sympy/core/sympify.py", line 418, in _sympify
      File "/usr/lib/python3.8/site-packages/sympy/core/sympify.py", line 340, in sympify
        raise SympifyError(a)
    sympy.core.sympify.SympifyError: SympifyError: 5*R.x + t*R.y

I see no reason why this should not work.

faze-geek:

Fixes #18692 

Implemented `solve_vector ` that can solve for unknown variables while two equal vectors get compared.
Implemented helper method `components()` that returns components of a vector in physics.

#### Other comments
Appropriate Tests have been added .
User must go through the `notes` to be aware of `solve_vector ` functionality.

#### Release Notes

* vector
  * physics : Added solve_vector function in physics vector

### DIFF
--- a/sympy/physics/vector/functions.py
+++ b/sympy/physics/vector/functions.py
@@ -700,8 +700,8 @@ def solve_vector(f1, f2, *symbols):
     Comp_list = []
     f1 = _check_vector(f1)
     f2 = _check_vector(f2)
-    c1 = f1.components()
-    c2 = f2.components()
+    c1 = f1._components()
+    c2 = f2._components()
 
     for i in c1:
         if c1[i].is_Number and c2[i].is_Number:
--- a/sympy/physics/vector/functions.py
+++ b/sympy/physics/vector/functions.py
@@ -648,14 +648,26 @@ def solve_vector(f1, f2, *symbols):
     Parameters
     ==========
 
-    f1 : Vector1
+    f1 : Vector
+        Input for vector 1.
+    f2 : Vector
+        Input for vector 2.
+    symbols: str, optional
+        Symbols the user wants to solve for. Single or multiple symbols
+        separated by commas are accepted. If no symbol is given as input,
+        values of all symbols are returned.
 
-    f2 : Vector2
+    Raises
+    ======
 
-    symbols: (object(s) to solve for) specified as
-        - none given(other non-numeric objects will be used)
-        - single symbol
-        - multiple symbols separated by commas
+    TypeError
+        When a scalar input is given.
+
+    ValueError
+        When same component of the 2 vectors have different numeric values.
+
+    AttributeError
+        When vector not declared along '.x', '.y', '.z' basis vectors.
 
     Examples
     ========
@@ -682,19 +694,7 @@ def solve_vector(f1, f2, *symbols):
     Notes
     =====
 
-    - ``f1`` and ``f2`` must be vectors.
-    - User must input vectors along '.x', '.y', '.z' basis vectors only.
-    - If two vectors have different numeric components a ValueError is raised.
-    - User must declare symbols as real to avoid dealing with complex solutions.
-
-    >>> solve_vector(a*R.x + 4*R.y, 0)
-    Traceback (most recent call last):
-    ...
-    TypeError: A Vector must be supplied
-    >>> solve_vector(a*R.x + 4*R.y, 6*R.x + 5*R.y, a)
-    Traceback (most recent call last):
-    ...
-    ValueError: Two Vectors with different numeric components cannot be equal
+    User must declare symbols as real to avoid dealing with complex solutions.
 
     """
     Comp_list = []
--- a/sympy/physics/vector/functions.py
+++ b/sympy/physics/vector/functions.py
@@ -1,9 +1,10 @@
 from functools import reduce
-
+from sympy.core.relational import Eq
 from sympy.core.backend import (sympify, diff, sin, cos, Matrix, symbols,
                                 Function, S, Symbol)
 from sympy.integrals.integrals import integrate
 from sympy.simplify.trigsimp import trigsimp
+from sympy.solvers import solve
 from .vector import Vector, _check_vector
 from .frame import CoordinateSym, _check_frame
 from .dyadic import Dyadic
@@ -638,3 +639,81 @@ def dynamicsymbols(names, level=0,**assumptions):
 
 dynamicsymbols._t = Symbol('t')  # type: ignore
 dynamicsymbols._str = '\''  # type: ignore
+
+
+def solve_vector(f1, f2, *symbols):
+    """
+    Solves for unknown variables while equating two Vector expressions.
+
+    Parameters
+    ==========
+
+    f1 : Vector1
+
+    f2 : Vector2
+
+    symbols: (object(s) to solve for) specified as
+        - none given(other non-numeric objects will be used)
+        - single symbol
+        - multiple symbols separated by commas
+
+    Examples
+    ========
+
+    The output varies according to the input and can be seen by example:
+
+        >>> from sympy import symbols
+        >>> from sympy.physics.vector import ReferenceFrame
+        >>> from sympy.physics.vector.functions import solve_vector
+        >>> R = ReferenceFrame('R')
+        >>> a, b, c = symbols("a b c", real = True)
+
+        >>> solve_vector(a*R.x + 4*R.y, 6*R.x + b*R.y)
+        {a: 6, b: 4}
+        >>> solve_vector(a*R.x + 4*R.y, 6*R.x + b*R.y,a)
+        {a: 6}
+        >>> solve_vector((a +b +c)*R.x + 2*b*R.y + (a+2*b)*R.z, 6*R.x + 4*R.y +6*R.z)
+        {a: 2, b: 2, c: 2}
+        >>> solve_vector((a**2 + b)*R.x + b*R.y, 6*R.x + 5*R.y)
+        [{a: -1, b: 5}, {a: 1, b: 5}]
+        >>> solve_vector((a**2 + b)*R.x + b*R.y, 5*R.x + 6*R.y)
+        []
+
+    Notes
+    =====
+
+    - ``f1`` and ``f2`` must be vectors.
+    - User must input vectors along '.x', '.y', '.z' basis vectors only.
+    - If two vectors have different numeric components a ValueError is raised.
+    - User must declare symbols as real to avoid dealing with complex solutions.
+
+    >>> solve_vector(a*R.x + 4*R.y, 0)
+    Traceback (most recent call last):
+    ...
+    TypeError: A Vector must be supplied
+    >>> solve_vector(a*R.x + 4*R.y, 6*R.x + 5*R.y, a)
+    Traceback (most recent call last):
+    ...
+    ValueError: Two Vectors with different numeric components cannot be equal
+
+    """
+    Comp_list = []
+    f1 = _check_vector(f1)
+    f2 = _check_vector(f2)
+    c1 = f1.components()
+    c2 = f2.components()
+
+    for i in c1:
+        if c1[i].is_Number and c2[i].is_Number:
+            if c1[i] != c2[i]:
+                raise ValueError('Two Vectors with different numeric components cannot be equal')
+        Comp_list.append(c1[i] - c2[i])
+
+    eq1 = Eq(Comp_list[0], 0)
+    eq2 = Eq(Comp_list[1], 0)
+    eq3 = Eq(Comp_list[2], 0)
+    f = (eq1, eq2 ,eq3)
+    if not symbols:
+        symbols = set().union(*[fi.free_symbols for fi in f])
+
+    return solve(f, symbols)
--- a/sympy/physics/vector/tests/test_functions.py
+++ b/sympy/physics/vector/tests/test_functions.py
@@ -506,3 +506,19 @@ def test_dynamicsymbols():
     assert f3.is_positive
     assert f4*f5 != f5*f4
     assert f6.is_integer
+
+
+def test_solve_vector():
+    from sympy.physics.vector import ReferenceFrame
+    from sympy.physics.vector.functions import solve_vector
+    R = ReferenceFrame('R')
+    a, b, c = symbols("a b c", real = True)
+    assert solve_vector(a*R.x + 4*R.y, 6*R.x + b*R.y) == {a: 6, b: 4}
+    assert solve_vector(a*R.x + 4*R.y, 6*R.x + b*R.y,a) == {a: 6}
+    assert solve_vector((a +b +c)*R.x + 2*b*R.y + (a+2*b)*R.z, 6*R.x + 4*R.y +6*R.z) == {a: 2, b: 2, c: 2}
+    assert solve_vector((a**2 + b)*R.x + b*R.y, 6*R.x + 5*R.y) == [{a: -1, b: 5}, {a: 1, b: 5}]
+    assert solve_vector((a**2 + b)*R.x + b*R.y, 5*R.x + 6*R.y) == []
+
+    raises(TypeError, lambda: solve_vector(a*R.x + 4*R.y, 0))
+    raises(ValueError, lambda: solve_vector(a*R.x + 4*R.y, 6*R.x + 5*R.y, a))
+    raises(AttributeError, lambda : solve_vector(a*R.i + 4*R.j, 6*R.i + b*R.j))
--- a/sympy/physics/vector/vector.py
+++ b/sympy/physics/vector/vector.py
@@ -453,21 +453,12 @@ class Vector(Printable, EvalfMixin):
     __rand__ = __and__
     __rmul__ = __mul__
 
-    def components(self):
+    def _components(self):
         """
         Returns the components of this vector in the form of a
         Python dictionary mapping BaseVector instances to the
-        corresponding measure numbers.
-        Examples
-        ========
-        >>> from sympy.physics.vector import ReferenceFrame
-        >>> R = ReferenceFrame('R')
-        >>> v = 3*R.x + 4*R.y + 5*R.z
-        >>> v.components()
-        {R.x: 3, R.y: 4, R.z: 5}
-        >>> v = 3*R.x + 4*R.y
-        >>> v.components()
-        {R.x: 3, R.y: 4, R.z: 0}
+        corresponding measure numbers.Used as helper function
+        for solve_vector().
         """
         components = {}
         components[(self.args[0][1]).x] = self.args[0][0][0]
--- a/sympy/physics/vector/vector.py
+++ b/sympy/physics/vector/vector.py
@@ -453,6 +453,29 @@ class Vector(Printable, EvalfMixin):
     __rand__ = __and__
     __rmul__ = __mul__
 
+    def components(self):
+        """
+        Returns the components of this vector in the form of a
+        Python dictionary mapping BaseVector instances to the
+        corresponding measure numbers.
+        Examples
+        ========
+        >>> from sympy.physics.vector import ReferenceFrame
+        >>> R = ReferenceFrame('R')
+        >>> v = 3*R.x + 4*R.y + 5*R.z
+        >>> v.components()
+        {R.x: 3, R.y: 4, R.z: 5}
+        >>> v = 3*R.x + 4*R.y
+        >>> v.components()
+        {R.x: 3, R.y: 4, R.z: 0}
+        """
+        components = {}
+        components[(self.args[0][1]).x] = self.args[0][0][0]
+        components[(self.args[0][1]).y] = self.args[0][0][1]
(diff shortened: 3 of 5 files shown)


faze-geek:
Ping @moorepants for review .

sympy/physics/vector/functions.py
```diff
@@ -638,3 +639,81 @@ def dynamicsymbols(names, level=0,**assumptions):
 
 dynamicsymbols._t = Symbol('t')  # type: ignore
 dynamicsymbols._str = '\''  # type: ignore
+
+
+def solve_vector(f1, f2, *symbols):
```
moorepants, line 644: Shouldn't this be `solve_vector(sequence_of_vector_expressions, *symbols)` to match `solve`?

sympy/physics/vector/functions.py
```diff
@@ -638,3 +639,81 @@ def dynamicsymbols(names, level=0,**assumptions):
...
+        >>> solve_vector(a*R.x + 4*R.y, 6*R.x + b*R.y)
+        {a: 6, b: 4}
+        >>> solve_vector(a*R.x + 4*R.y, 6*R.x + b*R.y,a)
+        {a: 6}
+        >>> solve_vector((a +b +c)*R.x + 2*b*R.y + (a+2*b)*R.z, 6*R.x + 4*R.y +6*R.z)
+        {a: 2, b: 2, c: 2}
+        >>> solve_vector((a**2 + b)*R.x + b*R.y, 6*R.x + 5*R.y)
+        [{a: -1, b: 5}, {a: 1, b: 5}]
+        >>> solve_vector((a**2 + b)*R.x + b*R.y, 5*R.x + 6*R.y)
+        []
```
moorepants, line 680: dedent

sympy/physics/vector/functions.py
```diff
@@ -638,3 +639,81 @@ def dynamicsymbols(names, level=0,**assumptions):
...
+
+        >>> solve_vector(a*R.x + 4*R.y, 6*R.x + b*R.y)
+        {a: 6, b: 4}
+        >>> solve_vector(a*R.x + 4*R.y, 6*R.x + b*R.y,a)
+        {a: 6}
+        >>> solve_vector((a +b +c)*R.x + 2*b*R.y + (a+2*b)*R.z, 6*R.x + 4*R.y +6*R.z)
+        {a: 2, b: 2, c: 2}
+        >>> solve_vector((a**2 + b)*R.x + b*R.y, 6*R.x + 5*R.y)
+        [{a: -1, b: 5}, {a: 1, b: 5}]
+        >>> solve_vector((a**2 + b)*R.x + b*R.y, 5*R.x + 6*R.y)
```
moorepants, line 679: Have you tried vector expressions made up of unit vectors from multiple reference frames?

sympy/physics/vector/functions.py
```diff
@@ -638,3 +639,81 @@ def dynamicsymbols(names, level=0,**assumptions):
...
+        []
+
+    Notes
+    =====
+
+    - ``f1`` and ``f2`` must be vectors.
+    - User must input vectors along '.x', '.y', '.z' basis vectors only.
+    - If two vectors have different numeric components a ValueError is raised.
```
moorepants, line 687: Should be in "raises" section. See the sympy docstring guide.

sympy/physics/vector/functions.py
```diff
@@ -638,3 +639,81 @@ def dynamicsymbols(names, level=0,**assumptions):
...
+    - User must declare symbols as real to avoid dealing with complex solutions.
+
+    >>> solve_vector(a*R.x + 4*R.y, 0)
+    Traceback (most recent call last):
+    ...
+    TypeError: A Vector must be supplied
+    >>> solve_vector(a*R.x + 4*R.y, 6*R.x + 5*R.y, a)
+    Traceback (most recent call last):
+    ...
+    ValueError: Two Vectors with different numeric components cannot be equal
```
moorepants, line 697: Raises section

sympy/physics/vector/tests/test_functions.py
```diff
@@ -506,3 +506,19 @@ def test_dynamicsymbols():
...
+    a, b, c = symbols("a b c", real = True)
+    assert solve_vector(a*R.x + 4*R.y, 6*R.x + b*R.y) == {a: 6, b: 4}
+    assert solve_vector(a*R.x + 4*R.y, 6*R.x + b*R.y,a) == {a: 6}
+    assert solve_vector((a +b +c)*R.x + 2*b*R.y + (a+2*b)*R.z, 6*R.x + 4*R.y +6*R.z) == {a: 2, b: 2, c: 2}
+    assert solve_vector((a**2 + b)*R.x + b*R.y, 6*R.x + 5*R.y) == [{a: -1, b: 5}, {a: 1, b: 5}]
+    assert solve_vector((a**2 + b)*R.x + b*R.y, 5*R.x + 6*R.y) == []
+
+    raises(TypeError, lambda: solve_vector(a*R.x + 4*R.y, 0))
+    raises(ValueError, lambda: solve_vector(a*R.x + 4*R.y, 6*R.x + 5*R.y, a))
+    raises(AttributeError, lambda : solve_vector(a*R.i + 4*R.j, 6*R.i + b*R.j))
```
moorepants, line 524: Needs at least one nonlinear vector equation test.

Needs multiple reference frame tests.

sympy/physics/vector/vector.py
```diff
@@ -453,6 +453,29 @@ def _det(mat):
...
+        >>> v = 3*R.x + 4*R.y
+        >>> v.components()
+        {R.x: 3, R.y: 4, R.z: 0}
+        """
+        components = {}
+        components[(self.args[0][1]).x] = self.args[0][0][0]
+        components[(self.args[0][1]).y] = self.args[0][0][1]
+        components[(self.args[0][1]).z] = self.args[0][0][2]
+
+        return components
```
moorepants, line 477: There is already something like this, `.seperate()` maybe?

Also, probably best to make this private for now.

moorepants:
Good start. I left some comments that need addressing.

sympy/physics/vector/functions.py
```diff
@@ -638,3 +639,81 @@ def dynamicsymbols(names, level=0,**assumptions):
...
+    ==========
+
+    f1 : Vector1
+
+    f2 : Vector2
+
+    symbols: (object(s) to solve for) specified as
+        - none given(other non-numeric objects will be used)
+        - single symbol
+        - multiple symbols separated by commas
```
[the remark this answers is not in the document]
faze-geek: Thanks , I have made all suitable docstring style changes you have pointed out .

sympy/physics/vector/vector.py
faze-geek: > There is already something like this, `.seperate()` maybe?

I went through `.separate()` method , it is used when 2 vectors of different reference frames are added and gives
constituents of the vector in different reference frames.

moorepants: Ok, probably just keep it private with a leading underscore we it is only used internally. We don't need to expose these things to the users (same goes for `seperate()` but that made it in as public is seems).

sympy/physics/vector/tests/test_functions.py
```diff
@@ -506,3 +506,19 @@ def test_dynamicsymbols():
...
+
+def test_solve_vector():
+    from sympy.physics.vector import ReferenceFrame
+    from sympy.physics.vector.functions import solve_vector
+    R = ReferenceFrame('R')
+    a, b, c = symbols("a b c", real = True)
+    assert solve_vector(a*R.x + 4*R.y, 6*R.x + b*R.y) == {a: 6, b: 4}
+    assert solve_vector(a*R.x + 4*R.y, 6*R.x + b*R.y,a) == {a: 6}
+    assert solve_vector((a +b +c)*R.x + 2*b*R.y + (a+2*b)*R.z, 6*R.x + 4*R.y +6*R.z) == {a: 2, b: 2, c: 2}
+    assert solve_vector((a**2 + b)*R.x + b*R.y, 6*R.x + 5*R.y) == [{a: -1, b: 5}, {a: 1, b: 5}]
```
oscarbenjamin, line 519: The inconsistent return types of `solve` are a serious design flaw that should not be replicated in any new API.

moorepants:
@oscarbenjamin, one issue is that physics vector's aren't subclassed from sympy core types and are mutable. So, we've mostly had to make these kinds of special functions strictly for physics vectors. Probably better not to pollute `solve()` with special code for these types. Not sure though.

sympy/physics/vector/tests/test_functions.py
moorepants: Yes, it would be best to always return the dict form.

oscarbenjamin: It would be better to always return a list of dicts i.e. the output of `solve` with `dict=True`

oscarbenjamin:
> Probably better not to pollute `solve()` with special code for these types.

The point is that `solve` should be extensible. If `solve` knows that some objects have a `to_scalar_equations` method then it can call that without needing to special case things like matrices, vectors etc. Handling matrices is currently hard-coded:
https://github.com/sympy/sympy/blob/69662b4fb4a9c585742da50ff33c5ba896108f23/sympy/solvers/solvers.py#L928-L932
If that was just calling a `to_scalar_equations()` method on Matrix then the same method added to vector would make it work with solve automatically.

moorepants:
Oscar, I agree, but physics vectors will always need the user to pass in a reference frame object too, so it then `solve()` would need a `ref_frame+=` kwarg for it to work. This is the reason that we have to create `free_symbols(ref_frame)`, `to_matrix(ref_frame)` etc in physics vector. So it can't just be an argument-less `to_scalar_equations()`.

oscarbenjamin:
I must be missing something basic here. Suppose that I have something like this:
```python
In [19]: from sympy.physics.vector import *
    ...: from sympy import *
    ...: 
    ...: t, l = symbols("t l")
    ...: R = ReferenceFrame("R")
    ...: eq1 = 5 * R.x + t * R.y - l * R.x + 4 * R.y

In [20]: eq1
Out[20]: (5 - l) r_x + (t + 4) r_y
```
Why is a reference frame needed to extract the coefficients?

When I call `args` it looks like the equations I want are just sitting there in an internal matrix:
```python
In [23]: eq1.args
Out[23]: 
[(Matrix([
  [5 - l],
  [t + 4],
  [    0]]),
  R)]
```

moorepants:
> When I call args it looks like the equations I want are just sitting there in an internal matrix:

Yes the are if you are only working in a single reference frame.

See:

```
In [1]: import sympy as sm

In [2]: import sympy.physics.mechanics as me

In [3]: a, b, c = sm.symbols('a, b, c')

In [4]: A = me.ReferenceFrame('A')

In [5]: B = me.ReferenceFrame('B')

In [6]: B.orient_body_fixed(A, (a, b, c), 'ZXZ')

In [7]: v = a*A.x + b*A.y + c*A.z + a*B.x + b*B.y + c*B.z

In [8]: v.args
Out[8]: 
[(Matrix([
  [a],
  [b],
  [c]]),
  A),
 (Matrix([
  [a],
  [b],
  [c]]),
  B)]

In [9]: v.express(B)
Out[9]: (a*(-sin(a)*sin(c)*cos(b) + cos(a)*cos(c)) + a + b*(sin(a)*cos(c) + sin(c)*cos(a)*cos(b)) + c*sin(b)*sin(c))*B.x + (a*(-sin(a)*cos(b)*cos(c) - sin(c)*cos(a)) + b*(-sin(a)*sin(c) + cos(a)*cos(b)*cos(c)) + b + c*sin(b)*cos(c))*B.y + (a*sin(a)*sin(b) - b*sin(b)*cos(a) + c*cos(b) + c)*B.z

In [10]: v.express(B).args
Out[10]: 
[(Matrix([
  [ a*(-sin(a)*sin(c)*cos(b) + cos(a)*cos(c)) + a + b*(sin(a)*cos(c) + sin(c)*cos(a)*cos(b)) + c*sin(b)*sin(c)],
  [a*(-sin(a)*cos(b)*cos(c) - sin(c)*cos(a)) + b*(-sin(a)*sin(c) + cos(a)*cos(b)*cos(c)) + b + c*sin(b)*cos(c)],
  [                                                           a*sin(a)*sin(b) - b*sin(b)*cos(a) + c*cos(b) + c]]),
  B)]
```

moorepants:
This gives the scalar equations:
```python
In [11]: v.to_matrix(B)
Out[11]: 
Matrix([
[ a*(-sin(a)*sin(c)*cos(b) + cos(a)*cos(c)) + a + b*(sin(a)*cos(c) + sin(c)*cos(a)*cos(b)) + c*sin(b)*sin(c)],
[a*(-sin(a)*cos(b)*cos(c) - sin(c)*cos(a)) + b*(-sin(a)*sin(c) + cos(a)*cos(b)*cos(c)) + b + c*sin(b)*cos(c)],
[                                                           a*sin(a)*sin(b) - b*sin(b)*cos(a) + c*cos(b) + c]])

```

So I guess it is simple enough to do:

```python
solve([v1.to_matrix(B), v2.to_matrix(B)], x, y, z)
```

Maybe that works?

moorepants:
@faze-geek can you see if something like this works:

```python
solve([v1.to_matrix(B), v2.to_matrix(B)], x, y, z)
```

If so, we could just add an example to the documentation.

sympy/physics/vector/functions.py
sidhu1012: I agree with @moorepants, the vector expressions used in examples and tests are trivial. You should check for complex vector expressions also.

sympy/physics/vector/vector.py
```diff
@@ -453,6 +453,29 @@ def _det(mat):
     __rand__ = __and__
     __rmul__ = __mul__
 
+    def components(self):
```
sidhu1012, line 456: Instead of adding a separate private function, can't we add this as an attribute to vector class? @moorepants

moorepants: My suggestion is that this components thing is not something the user needs, so we should hide it from the user as an internal private method.

sidhu1012: This function isn't doing anything unique, just using vector args. So in my opinion we should just simply add an attribute to vector class which returns the components as a dictionary.

```
>>> v = v = 3*R.x + 4*R.y + 5*R.z
>>> v.comp
{R.x: 3, R.y: 4, R.z: 5}

```

moorepants: Sure, but private `_comp`.

moorepants:
@sidhu1012, this needs to be addressed first:

> @faze-geek can you see if something like this works:
> 
> ```python
> solve([v1.to_matrix(B), v2.to_matrix(B)], x, y, z)
> ```
> 
> If so, we could just add an example to the documentation.

sidhu1012:
Yes, even I addressed it https://github.com/sympy/sympy/pull/23068#discussion_r808755520

sympy/physics/vector/functions.py
faze-geek: @sidhu1012 Could you give an example for complex vector expressions ? 
I'm working /correcting another pr of mine and will address comments on this one soon .Thanks for your suggestions !

moorepants:
> Yes, even I addressed it https://github.com/sympy/sympy/pull/23068#discussion_r808755520

That's not the point I'm making. If all you have to do is convert to matrix and solve, we don't need a method for this. Simply show how to do it in the documentation.

sympy/physics/vector/vector.py
faze-geek: Yeah I have made it private now .In the vector module `components()` is a public function ( which was my motivation to make it similar ) . I think it is fine as not accessible to the user .But if it should be added as an attribute to vector class, I'll make changes in upcoming commits .Thanks !

faze-geek:
> @sidhu1012, this needs to be addressed first:
> 
> > @faze-geek can you see if something like this works:
> > ```python
> > solve([v1.to_matrix(B), v2.to_matrix(B)], x, y, z)
> > ``` 
> > If so, we could just add an example to the documentation.

@moorepants I guess you wanted to see something like this 
```
>>> from sympy import *
>>> from sympy.physics.vector import ReferenceFrame
>>> R = ReferenceFrame('R')
>>> a, b, c = symbols("a b c")
>>> v1 = a*R.x + 4*R.y
>>> v2 = 6*R.x + b*R.y
>>> x= v1.to_matrix(R)
>>> y =v2.to_matrix(R)
>>> t =Eq(x,y)
>>> solve(t,a,b)
{a: 6, b: 4}

```
This works fine ,actually I had thought of something like this ( common approach in `MATLAB` )  but missed `.to_matrix()` while going through the documentation . This is a trivial case so I'll try some more examples and report here .Thanks !

moorepants:
> This works fine ,actually I had thought of something like this ( common approach in MATLAB ) but missed .to_matrix() while going through the documentation . This is a trivial case so I'll try some more examples and report here .Thanks !

Great. Then adding an example or two of how to do this in the documentation is a sufficient solution.

faze-geek:
> Great. Then adding an example or two of how to do this in the documentation is a sufficient solution.

What do you suggest ?
Maybe this PR can just be about adding a public function of components (which is even public in vector module and a basic property I think physics.vector must also have anyways ) . Then I can open up a new pr regarding documentation examples .
Or I can revert the commits of this pr and add documentation in this pr itself !?
Thanks .

oscarbenjamin:
> Maybe this PR can just be about adding a public function of components

If the public function isn't needed then it is better not to add it.

moorepants:
I would suggest closing this and making a new PR with the examples.

faze-geek:
> I would suggest closing this and making a new PR with the examples.

Thanks I will work on this .I would like if you would mentioned where the examples should be added to be of most help to users ( I was a bit dicey about it ,so would like to know from your side ).

moorepants:
It depends on what you plan to write a bit. We have actual examples for vector/mechanics and there is an advanced page.

sujantkumarkv:
Hey! I request update on the state of the PR and the issue it's solving. 
What's missing and what specific results are expected before to get it merged?
@faze-geek are you still working on it? Let me know the details.

cc : @moorepants @sidhu1012 @oscarbenjamin

faze-geek:
> What's missing and what specific results are expected before to get it merged?
> @faze-geek are you still working on it? Let me know the details.

Hi ,the current status of the pr can be understood by reading the following comment -
https://github.com/sympy/sympy/pull/23068#issuecomment-1048467286

If you want to add those examples, you can go ahead. Thanks !

moorepants:
Let's close this in favor of making one or more examples in the documentation. Thanks @faze-geek.

sujantkumarkv:
Hey! 
I'll be attempting to summarise to see if I understood correctly (please correct me if wrong) and what's needed to be done. I guess it also helps someone in future to understand.

- So, it starts with issue #18692 where a user wasn't able to solve vector equations, [here](https://github.com/sympy/sympy/issues/18692#issue-568132463). `sympy.vector` was created later to `sympy.physics.vector` to eventually replace it but it hasn't yet. 

- Then, pull #23068 was created to make a new function `solve_vector()` in `sympy.physics.vector`,  a point to note: `sympy.vector` was [also not solving](https://github.com/sympy/sympy/issues/18692#issuecomment-589075318) the equation.

- The `solve()` from `sympy.physics.vector`  was able to solve it with `solve(equation.to_matrix(*ReferenceFrame)`.
Screenshot 2023-03-12 at 12 07 23 AM

-  Note: I checked that `solve()` from `sympy.vector` can't solve it.
Screenshot 2023-03-12 at 12 00 32 AM

So, what's needed to be done is: Add examples of `solve(equation.to_matrix(*ReferenceFrame)` with different equations in a new PR, no functions or extra code?

I had two questions:
- what about the `components`?  I agree that we can simply have it as an attribute.
- If changes to `solve()` in `sympy.vector` should be done since `equation.to_matrix()` doesn't exist as the plan is to replace `sympy.physics.vector`, so shouldn't it be able to do what physics vector can? 

cc: @faze-geek @moorepants @oscarbenjamin @sidhu1012

sujantkumarkv:
Pinging for response:
@faze-geek @moorepants @oscarbenjamin @sidhu1012

PR #23045, opened 2022-02-08
physics/control: label for Frequency axis in bode_plot modified

AdvaitPote:
#### References to other Issues or PRs

Fixes the direct bug in #23034 
#### Brief description of what is fixed or changed
The x-axis for bode_plot is labeled as 'Hz' while the correct unit to be used should be 'rad/s'. This has been fixed by modifying the label for the x-axis. 

#### Other comments
Appropriate tests to be added. 

#### Release Notes

NO ENTRY

PR #23074, opened 2022-02-14
physics : Improved Bode's phase and magnitude plots

### LINKED ISSUE #23034: Wrong axis label in bode_plot
The code 

```
from sympy.abc import s
from sympy.physics.control.lti import TransferFunction
from sympy.physics.control.control_plots import bode_plot
tf = TransferFunction(1, 1+s, s)
bode_plot(tf)
```

shows that the frequency axis that is labelled with the unit Hz actually should have the unit rad/s.

This should be corrected; at the same time it would be nice to have options for:

- Frequency in Hz or rad/sec
- Amplitude in dB or just as a logarithmic axis
- Phase in rad or deg

Issue created from discussion https://groups.google.com/g/sympy/c/xGz7KB1UKAU/m/twv6AGI1BAAJ?utm_medium=email&utm_source=footer

faze-geek:

Fixes #23034 
Fixes [#23045(comment)](https://github.com/sympy/sympy/pull/23045#issuecomment-1038591114)

Modified `bode_plot()` . Now user has options to choose the axis units -
1. Frequency in Hz or rad/sec
2. Phase in rad or deg
 
#### Other comments
The attribute names and error messages are in line with `MATLAB` documentation. 
Current plots are compared with `MATLAB`.

#### Release Notes

* physics.control
  * Added unit parameter to Bode's plot which returns plot accordingly.
  * Fixed labels of magnitude and phase plots.

### DIFF
--- a/sympy/physics/control/control_plots.py
+++ b/sympy/physics/control/control_plots.py
@@ -669,7 +669,7 @@ def ramp_response_plot(system, slope=1, color='b', prec=8, lower_limit=0,
     return plt
 
 
-def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs):
+def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, freq_unit='rad/sec', **kwargs):
     """
     Returns the numerical data of the Bode magnitude plot of the system.
     It is internally used by ``bode_magnitude_plot`` to get the data
@@ -686,6 +686,8 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs)
         The initial exponent of 10 of the semilog plot. Defaults to -5.
     final_exp : Number, optional
         The final exponent of 10 of the semilog plot. Defaults to 5.
+    freq_unit : string, optional
+        User can choose between radians/second and Hertz as frequency units.
 
     Returns
     =======
@@ -731,12 +733,17 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs)
 
     mag = 20*log(Abs(w_expr), 10)
 
-    return LineOver1DRangeSeries(mag,
+    x, y = LineOver1DRangeSeries(mag,
         (_w, 10**initial_exp, 10**final_exp), xscale='log', **kwargs).get_points()
+    if freq_unit == 'Hz':
+        a = pi.n()
+        x = [i/(2*a) for i in x]
+
+    return x, y
 
 
 def bode_magnitude_plot(system, initial_exp=-5, final_exp=5,
-    color='b', show_axes=False, grid=True, show=True, freq_unit='rad/sec', **kwargs):
+    color='b', show_axes=False, grid=True, show=True, hide_unit=False, freq_unit='rad/sec', **kwargs):
     r"""
     Returns the Bode magnitude plot of a continuous-time system.
 
@@ -747,14 +754,12 @@ def bode_magnitude_plot(system, initial_exp=-5, final_exp=5,
         raise ValueError('Only rad/sec and Hz are accepted frequency units.')
 
     x, y = bode_magnitude_numerical_data(system, initial_exp=initial_exp,
-        final_exp=final_exp)
-    if freq_unit == 'Hz':
-        a = pi.n()
-        x = [i/(2*a) for i in x]
+        final_exp=final_exp, freq_unit=freq_unit)
     plt.plot(x, y, color=color, **kwargs)
     plt.xscale('log')
 
-    plt.xlabel('Frequency (%s) [Log Scale]' % freq_unit)
+    if hide_unit == False:
+        plt.xlabel('Frequency (%s) [Log Scale]' % freq_unit)
     plt.ylabel('Magnitude (dB)')
     plt.title(f'Bode Plot (Magnitude) of ${latex(system)}$', pad=20)
 
@@ -770,7 +775,7 @@ def bode_magnitude_plot(system, initial_exp=-5, final_exp=5,
     return plt
 
 
-def bode_phase_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs):
+def bode_phase_numerical_data(system, initial_exp=-5, final_exp=5, freq_unit='rad/sec', phase_unit='rad', **kwargs):
     """
     Returns the numerical data of the Bode phase plot of the system.
     It is internally used by ``bode_phase_plot`` to get the data
@@ -787,6 +792,10 @@ def bode_phase_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs):
         The initial exponent of 10 of the semilog plot. Defaults to -5.
     final_exp : Number, optional
         The final exponent of 10 of the semilog plot. Defaults to 5.
+    freq_unit : string, optional
+        User can choose between radians/second and Hertz as frequency units.
+    phase_unit : string, optional
+        User can choose between radians and degree as phase units.
 
     Returns
     =======
@@ -832,8 +841,16 @@ def bode_phase_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs):
 
     phase = arg(w_expr)
 
-    return LineOver1DRangeSeries(phase,
+    x, y = LineOver1DRangeSeries(phase,
         (_w, 10**initial_exp, 10**final_exp), xscale='log', **kwargs).get_points()
+    if freq_unit == 'Hz':
+        a = pi.n()
+        x = [i/(2*a) for i in x]
+    if phase_unit == 'deg':
+        a = pi.n()
+        y = [i*(180/a) for i in y]
+
+    return x, y
 
 
 def bode_phase_plot(system, initial_exp=-5, final_exp=5,
@@ -851,13 +868,7 @@ def bode_phase_plot(system, initial_exp=-5, final_exp=5,
         raise ValueError('Only rad and deg are accepted phase units.')
 
     x, y = bode_phase_numerical_data(system, initial_exp=initial_exp,
-        final_exp=final_exp)
-    if freq_unit == 'Hz':
-        a = pi.n()
-        x = [i/(2*a) for i in x]
-    if phase_unit == 'deg':
-        a = pi.n()
-        y = [i*(180/a) for i in y]
+        final_exp=final_exp, freq_unit=freq_unit, phase_unit=phase_unit)
     plt.plot(x, y, color=color, **kwargs)
     plt.xscale('log')
 
@@ -878,7 +889,7 @@ def bode_phase_plot(system, initial_exp=-5, final_exp=5,
 
 
 def bode_plot(system, initial_exp=-5, final_exp=5,
-    grid=True, show_axes=False, show=True, freq_unit='rad/sec', phase_unit='rad', **kwargs):
+    grid=True, show_axes=False, show=True, hide_unit=True, freq_unit='rad/sec', phase_unit='rad', **kwargs):
     r"""
     Returns the Bode phase and magnitude plots of a continuous-time system.
 
@@ -902,6 +913,8 @@ def bode_plot(system, initial_exp=-5, final_exp=5,
         If ``True``, the plot will have a grid. Defaults to True.
     show_axes : boolean, optional
         If ``True``, the coordinate axes will be shown. Defaults to False.
+    hide_axis : boolean, optional
+        If ``True``, the x axis unit will not be shown . Defaults to True.
     freq_unit : string, optional
         User can choose between radians/second and Hertz as frequency units.
     phase_unit : string, optional
@@ -930,7 +943,7 @@ def bode_plot(system, initial_exp=-5, final_exp=5,
     plt.subplot(211)
     bode_magnitude_plot(system, initial_exp=initial_exp, final_exp=final_exp,
         show=False, grid=grid, show_axes=show_axes,
-        freq_unit=freq_unit, **kwargs).title(f'Bode Plot of ${latex(system)}$', pad=20)
+        hide_unit=hide_unit, freq_unit=freq_unit, **kwargs).title(f'Bode Plot of ${latex(system)}$', pad=20)
     plt.subplot(212)
     bode_phase_plot(system, initial_exp=initial_exp, final_exp=final_exp,
         show=False, grid=grid, show_axes=show_axes, freq_unit=freq_unit, phase_unit=phase_unit, **kwargs).title(None)
--- a/sympy/physics/control/control_plots.py
+++ b/sympy/physics/control/control_plots.py
@@ -9,6 +9,7 @@ from sympy.physics.control.lti import SISOLinearTimeInvariant
 from sympy.plotting.plot import LineOver1DRangeSeries
 from sympy.polys.polytools import Poly
 from sympy.printing.latex import latex
+from sympy import S
 
 __all__ = ['pole_zero_numerical_data', 'pole_zero_plot',
     'step_response_numerical_data', 'step_response_plot',
@@ -736,18 +737,25 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs)
 
 
 def bode_magnitude_plot(system, initial_exp=-5, final_exp=5,
-    color='b', show_axes=False, grid=True, show=True, **kwargs):
+    color='b', show_axes=False, grid=True, show=True, Freq_unit='rad/sec', **kwargs):
     r"""
     Returns the Bode magnitude plot of a continuous-time system.
 
     See ``bode_plot`` for all the parameters.
     """
+    Freq_units = ('rad/sec', 'Hz')
+    if Freq_unit not in Freq_units:
+        raise ValueError('Only rad/sec and Hz are accepted frequency units.')
+
     x, y = bode_magnitude_numerical_data(system, initial_exp=initial_exp,
         final_exp=final_exp)
+    if Freq_unit == 'Hz':
+        a = S.Pi.n()
+        x = [i/(2*(a)) for i in x]
     plt.plot(x, y, color=color, **kwargs)
     plt.xscale('log')
 
-    plt.xlabel('Frequency (rad/s) [Log Scale]')
+    plt.xlabel('Frequency (%s) [Log Scale]' % Freq_unit)
     plt.ylabel('Magnitude (dB)')
     plt.title(f'Bode Plot (Magnitude) of ${latex(system)}$', pad=20)
 
@@ -830,19 +838,32 @@ def bode_phase_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs):
 
 
 def bode_phase_plot(system, initial_exp=-5, final_exp=5,
-    color='b', show_axes=False, grid=True, show=True, **kwargs):
+    color='b', show_axes=False, grid=True, show=True, Phase_unit='rad', Freq_unit='rad/sec', **kwargs):
     r"""
     Returns the Bode phase plot of a continuous-time system.
 
     See ``bode_plot`` for all the parameters.
     """
+    Freq_units = ('rad/sec', 'Hz')
+    Phase_units = ('rad', 'deg')
+    if Freq_unit not in Freq_units:
+        raise ValueError('Only rad/sec and Hz are accepted frequency units.')
+    if Phase_unit not in Phase_units:
+        raise ValueError('Only rad and deg are accepted phase units.')
+
     x, y = bode_phase_numerical_data(system, initial_exp=initial_exp,
         final_exp=final_exp)
+    if Freq_unit == 'Hz':
+        a = S.Pi.n()
+        x = [i/(2*(a)) for i in x]
+    if Phase_unit == 'deg':
+        a = S.Pi.n()
+        y = [i*(180/a) for i in y]
     plt.plot(x, y, color=color, **kwargs)
     plt.xscale('log')
 
-    plt.xlabel('Frequency (Hz) [Log Scale]')
-    plt.ylabel('Phase (rad)')
+    plt.xlabel('Frequency (%s) [Log Scale]' % Freq_unit)
+    plt.ylabel('Phase (%s)' % Phase_unit)
     plt.title(f'Bode Plot (Phase) of ${latex(system)}$', pad=20)
 
     if grid:
@@ -858,7 +879,7 @@ def bode_phase_plot(system, initial_exp=-5, final_exp=5,
 
 
 def bode_plot(system, initial_exp=-5, final_exp=5,
-    grid=True, show_axes=False, show=True, **kwargs):
+    grid=True, show_axes=False, show=True, Phase_unit='rad', Freq_unit='rad/sec', **kwargs):
     r"""
     Returns the Bode phase and magnitude plots of a continuous-time system.
 
@@ -882,6 +903,10 @@ def bode_plot(system, initial_exp=-5, final_exp=5,
         If ``True``, the plot will have a grid. Defaults to True.
     show_axes : boolean, optional
         If ``True``, the coordinate axes will be shown. Defaults to False.
+    Phase_unit : string, optional
+        User can choose between radians and degree as phase units.
+    Freq_unit : string, optional
+        User can choose between radians/second and Hertz as frequency units.
 
     Examples
     ========
@@ -906,10 +931,10 @@ def bode_plot(system, initial_exp=-5, final_exp=5,
     plt.subplot(211)
     bode_magnitude_plot(system, initial_exp=initial_exp, final_exp=final_exp,
         show=False, grid=grid, show_axes=show_axes,
-        **kwargs).title(f'Bode Plot of ${latex(system)}$', pad=20)
+        Freq_unit=Freq_unit, **kwargs).title(f'Bode Plot of ${latex(system)}$', pad=20)
     plt.subplot(212)
     bode_phase_plot(system, initial_exp=initial_exp, final_exp=final_exp,
-        show=False, grid=grid, show_axes=show_axes, **kwargs).title(None)
+        show=False, grid=grid, show_axes=show_axes, Phase_unit=Phase_unit, Freq_unit=Freq_unit, **kwargs).title(None)
 
     if show:
         plt.show()
--- a/sympy/physics/control/control_plots.py
+++ b/sympy/physics/control/control_plots.py
@@ -687,7 +687,7 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, freq_unit
     final_exp : Number, optional
         The final exponent of 10 of the semilog plot. Defaults to 5.
     freq_unit : string, optional
-        User can choose between 'rad/sec' (radians/second) and 'Hz' (Hertz) as frequency units.
+        User can choose between ``'rad/sec'`` (radians/second) and ``'Hz'`` (Hertz) as frequency units.
 
     Returns
     =======
@@ -732,7 +732,7 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, freq_unit
     expr = system.to_expr()
     freq_units = ('rad/sec', 'Hz')
     if freq_unit not in freq_units:
-        raise ValueError('Only rad/sec and Hz are accepted frequency units.')
+        raise ValueError('Only "rad/sec" and "Hz" are accepted frequency units.')
 
     _w = Dummy("w", real=True)
     if freq_unit == 'Hz':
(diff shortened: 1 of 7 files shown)


faze-geek:
Ping @oscargus @akshanshbhatt 

I could not figure out how to add tests since we are dealing with graphs . 
I will show the improvements here -
```

>>> from sympy.abc import s
>>> from sympy.physics.control.lti import TransferFunction
>>> from sympy.physics.control.control_plots import bode_plot
>>> tf = TransferFunction(1,
>>> bode_plot(tf)
```
anurag1

`>>> bode_plot(tf, Freq_unit='Hz')`

anurag2

`>>> bode_plot(tf, Phase_unit='deg')`
anurag3

`>>> bode_plot(tf, Phase_unit='deg',Freq_unit='Hz')`



anurag4

```
>>> bode_plot(tf, Phase_unit='deg',Freq_unit='hz')
  File "C:\Users\kunni\sympy\sympy\sympy\physics\control\control_plots.py", line 932, in bode_plot
    bode_magnitude_plot(system, initial_exp=initial_exp, final_exp=final_exp,
  File "C:\Users\kunni\sympy\sympy\sympy\physics\control\control_plots.py", line 748, in bode_magnitude_plot
    raise ValueError('Only rad/sec and Hz are accepted frequency units.')
```

sympy/physics/control/control_plots.py
```diff
@@ -736,18 +737,25 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs)
...
 
     See ``bode_plot`` for all the parameters.
     """
+    Freq_units = ('rad/sec', 'Hz')
+    if Freq_unit not in Freq_units:
+        raise ValueError('Only rad/sec and Hz are accepted frequency units.')
+
     x, y = bode_magnitude_numerical_data(system, initial_exp=initial_exp,
         final_exp=final_exp)
+    if Freq_unit == 'Hz':
```
oscargus, line 752: Ideally, this should be passed to `bode_magnitude_numerical_data`. In that way you should be able to divide by 2*pi in the SymPy expression rather than in the vector (although I have not checked how it looks there).

sympy/physics/control/control_plots.py
```diff
@@ -736,18 +737,25 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs)
...
+        a = S.Pi.n()
+        x = [i/(2*(a)) for i in x]
```
oscargus, line 754:
```suggestion
        x = [i/(2*a) for i in x]
```
(Although better to move it.)

sympy/physics/control/control_plots.py
```diff
@@ -736,18 +737,25 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs)
 
 
 def bode_magnitude_plot(system, initial_exp=-5, final_exp=5,
-    color='b', show_axes=False, grid=True, show=True, **kwargs):
+    color='b', show_axes=False, grid=True, show=True, Freq_unit='rad/sec', **kwargs):
```
oscargus, line 740: Use lower case `freq_unit` (same for keyword arguments of the other function below).

sympy/physics/control/control_plots.py
oscargus: Should be straightforward to obtain it here:
https://github.com/sympy/sympy/blob/69662b4fb4a9c585742da50ff33c5ba896108f23/sympy/physics/control/control_plots.py#L730

ThePauliPrinciple:
The plots seem to be overlapping slightly, is there anything that can be done about this?

faze-geek:
> The plots seem to be overlapping slightly, is there anything that can be done about this?

Yeah ,I guess it is a separate issue .On my monitor when I plot it on a small window size the Frequency unit is completely hidden and as I increase the window size it gets more visible .On maximum, the unit and below plot are distinct .I could maybe print a few empty lines ? (but if it's an issue with subplots ,it would be better to handle it in a separate PR )

akshanshbhatt:
I agree with @oscargus. Division by `2*Pi` should be done in `bode_magnitude_numerical_data` and `bode_phase_numerical_data` functions.

akshanshbhatt:
Also, mention the bug that you have fixed in the release notes.

akshanshbhatt:
> I could maybe print a few empty lines ?

That might not be a good idea. I would prefer to remove the redundant x-axis label of the magnitude plot.
It is not visible in the documentation anyways.

https://docs.sympy.org/dev/modules/physics/control/control_plots.html?highlight=bode#sympy.physics.control.control_plots.bode_plot

> One more thing that can be done is what `MATLAB` does by default that since the x axis and it's values are same for both the plot unless user manually sets the option only one x axis unit ( the below one) is shown .We can add that option `show_unit = False` ?

Since the x-axis of both the plots is the same, I don't think the `show_unit` option will be of any use.

oscargus:
> I would prefer to remove the redundant x-axis label of the magnitude plot.

I agree. It is a problem with subplots and specific backends that probably should be fixed elsewhere, but for this particular case it seems to be the best idea to just skip the x-label in case both are plotted (this is probably why the first fix didn't sort out the magnitude x-label, btw, it was not visible).

faze-geek:
> That might not be a good idea. I would prefer to remove the redundant x-axis label of the magnitude plot. It is not visible in the documentation anyways.
> Since the x-axis of both the plots is the same, I don't think the `show_unit` option will be of any use.

But won't removing it directly be problematic when the user only calls `bode_magnitude_plot()` on his transfer function and then there is no unit shown on the x-axis ?

faze-geek:
> I agree with @oscargus. Division by `2*Pi` should be done in `bode_magnitude_numerical_data` and `bode_phase_numerical_data` functions.

Yeah even I agree ,I will handle this in a commit soon !

oscargus:
> But won't removing it directly be problematic when the user only calls `bode_magnitude_plot()` on his transfer function and then there is no unit shown on the x-axis ?

Sure, so that should not happen. Use a keyword argument to control it.

akshanshbhatt:
> But won't removing it directly be problematic when the user only calls `bode_magnitude_plot()` on his transfer function and then there is no unit shown on the x-axis ?

What I mean is that remove the x-axis when you are calling the `bode_magnitude_plot()` function inside the `bode_plot()` function. You can pass an argument here to hide the x-axis in the magnitude subplot -
https://github.com/sympy/sympy/blob/c1e1f50cb68caa4976b63714fbe7cb84c3b4a342/sympy/physics/control/control_plots.py#L907-L909

sympy/physics/control/control_plots.py
faze-geek: Sorry I misjudged the hint thinking I could handle the constant in the SymPy expression itself .I'll make this change .

faze-geek: @akshanshbhatt @oscargus 
I have made the changes .These are the improvements -

```
>>> from sympy.abc import s
>>> from sympy.physics.control.lti import TransferFunction
>>> tf = TransferFunction(1, 1+s, s)
>>> from sympy.physics.control.control_plots import bode_magnitude_plot
>>> bode_magnitude_plot(tf)
```
Ip1

```
>>> from sympy.physics.control.control_plots import bode_plot
>>> bode_plot(tf)
```

Ip2

`>>> bode_plot(tf,freq_unit='Hz')`

Ip3

`>>> bode_plot(tf,phase_unit='deg')`

Ip4

sympy/physics/control/control_plots.py
```diff
@@ -731,23 +733,33 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs)
 
     mag = 20*log(Abs(w_expr), 10)
 
-    return LineOver1DRangeSeries(mag,
+    x, y = LineOver1DRangeSeries(mag,
         (_w, 10**initial_exp, 10**final_exp), xscale='log', **kwargs).get_points()
+    if freq_unit == 'Hz':
```
oscargus, line 738: Move this to `w_expr` on line 732.

```
if freq_unit == 'Hz':
    repl = I*_w/(2*pi)
else:
    repl = I*_w
w_expr = expr.subs({system.var: repl})
```

and the same for phase (for `phase_unit` you can multiply `phase = arg(w_expr)` with `180/pi` in case of degrees *before*  evaluating it).

Probably it also makes sense moving the check of valid units to the numerical_data-functions.

sympy/physics/control/control_plots.py
```diff
@@ -731,23 +733,33 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs)
...
+        raise ValueError('Only rad/sec and Hz are accepted frequency units.')
+
     x, y = bode_magnitude_numerical_data(system, initial_exp=initial_exp,
-        final_exp=final_exp)
+        final_exp=final_exp, freq_unit=freq_unit)
     plt.plot(x, y, color=color, **kwargs)
     plt.xscale('log')
 
-    plt.xlabel('Frequency (rad/s) [Log Scale]')
+    if hide_unit == False:
```
oscargus, line 761:
```suggestion
    if not hide_unit:
```

oscargus: Although `show_xlabel` may be a better name.

sympy/physics/control/control_plots.py
faze-geek: Thanks for this .Even time wise this solution is better .For `phase = arg(w_expr)` multiplying works fine and gives the expected graphs .While for `repl = I*_w/(2*pi) ` there is an observation on applying the diff you suggested .

By `repl = I*_w/(2*pi)` x axis values are multiplied by 2*pi (instead of divided by it which is expected)


OG2

Having seen this it was obvious to try the inverse operation.
By `repl = I*_w*(2*pi)` x axis values are divided by 2*pi ( which gives the expected plots )

OG1

This needs to be investigated why` I*_w*(2*pi)` gives the correct results and should we go forward with that ?

faze-geek: @akshanshbhatt Any thoughts on this ?

akshanshbhatt:
`show_xlabel` is not an elegant solution. It will make the API confusing for the user -

- Why is there only an option for showing/hiding the `xlabel`? Why is there no option for showing/hiding the `ylabel`?
- Why is this option available for the Bode-Magnitude plot and not the Bode-Phase plot (or any other plot for that reason)?
- What is the practicality of having this option for the user? Will it ever be useful?
- There is already an option to show/hide the axis for all the control plots.

Here is how I would have done it.

```diff
--- a/sympy/physics/control/control_plots.py
+++ b/sympy/physics/control/control_plots.py
@@ -943,9 +943,11 @@ def bode_plot(system, initial_exp=-5, final_exp=5,
 
     """
     plt.subplot(211)
-    bode_magnitude_plot(system, initial_exp=initial_exp, final_exp=final_exp,
+    mag = bode_magnitude_plot(system, initial_exp=initial_exp, final_exp=final_exp,
         show=False, grid=grid, show_axes=show_axes,
-        show_xlabel=show_xlabel, freq_unit=freq_unit, **kwargs).title(f'Bode Plot of ${latex(system)}$', pad=20)
+        show_xlabel=show_xlabel, freq_unit=freq_unit, **kwargs)
+    mag.title(f'Bode Plot of ${latex(system)}$', pad=20)
+    mag.xlabel(None)
     plt.subplot(212)
     bode_phase_plot(system, initial_exp=initial_exp, final_exp=final_exp,
         show=False, grid=grid, show_axes=show_axes, freq_unit=freq_unit, phase_unit=phase_unit, **kwargs).title(None
)
```

Output -

bode_plot 2

faze-geek:
> * Why is there only an option for showing/hiding the `xlabel`? Why is there no option for showing/hiding the `ylabel`?
> * Why is this option available for the Bode-Magnitude plot and not the Bode-Phase plot (or any other plot for that reason)?
> * What is the practicality of having this option for the user? Will it ever be useful?
> * There is already an option to show/hide the axes for all the control plots.

Yes these questions have also cross my mind ,I even mentioned in the [#23074(comment)](https://github.com/sympy/sympy/pull/23074#issuecomment-1039820139) that this is an issue with subplot and the concerned backend so we would have to look for a particular fix for `bode_plot`. This purely tackles the overlapping of the unit and lower plot so no `ylabel `. I beleive `show_axes` function that you mentioned has different functionality .

But anyways I agree with the diff you suggested .Thanks

sympy/physics/control/control_plots.py
faze-geek: CC @oscargus @akshanshbhatt I have made all the other changes ,would like if you addressed this comment .Both the current and this suggested code run in similar time ( the suggestion a bit faster ) . Other than that the plots are identical and as expected !

akshanshbhatt: Go ahead with the suggested one. It looks better and is more pythonic.

oscargus: ` I*_w*(2*pi)` is the correct version, my bad.

oscargus: The point is to use the symbolic functionality rather than floating-point computations.

oscargus:
> Here is how I would have done it.

Clearly a better solution. (Answer to other questions: yes, everything can be added, given time and effort. That was the one that was required, given the previous solution.)

sympy/physics/control/control_plots.py
faze-geek: Thanks @oscargus @akshanshbhatt .I have made all the above changes in the latest commit .It seems that the `sphinx.sh`
check will fail (I guess because of the change `mag.xlabel(None)` ) and I'm not too familiar with it .I will look into that today !

sympy/physics/control/control_plots.py
```diff
@@ -686,6 +686,8 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs)
         The initial exponent of 10 of the semilog plot. Defaults to -5.
     final_exp : Number, optional
         The final exponent of 10 of the semilog plot. Defaults to 5.
+    freq_unit : string, optional
+        User can choose between ``rad/sec``(radians/second) and ``Hz``(Hertz) as frequency units.
```
oscargus, line 690:
```suggestion
        User can choose between ``'rad/sec'`` (radians/second) and ``'Hz'`` (Hertz) as frequency units.
```

sympy/physics/control/control_plots.py
```diff
@@ -707,6 +709,8 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs)
         The only variable in the transfer function should be
         the variable of the Laplace transform.
 
+    When incorrect frequency units are given as input.
```
oscargus, line 712: I think this may cause the sphinx error.  Remove the empty line above and indent this as the text above.

sympy/physics/control/control_plots.py
```diff
@@ -801,6 +820,8 @@ def bode_phase_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs):
         The only variable in the transfer function should be
         the variable of the Laplace transform.
 
+        When incorrect frequency or phase units are given as input.
```
oscargus, line 823: Maybe remove the empty line above? (The other case is more likely to cause the error, but maybe this one as well.)

sympy/physics/control/control_plots.py
```diff
@@ -882,6 +919,10 @@ def bode_plot(system, initial_exp=-5, final_exp=5,
         If ``True``, the plot will have a grid. Defaults to True.
     show_axes : boolean, optional
         If ``True``, the coordinate axes will be shown. Defaults to False.
+    freq_unit : string, optional
+        User can choose between ``rad/sec``(radians/second) and ``Hz``(Hertz) as frequency units.
```
oscargus, line 923:
```suggestion
        User can choose between ``'rad/sec'`` (radians/second) and ``'Hz'`` (Hertz) as frequency units.
```

sympy/physics/control/control_plots.py
```diff
@@ -882,6 +919,10 @@ def bode_plot(system, initial_exp=-5, final_exp=5,
...
+    phase_unit : string, optional
+        User can choose between ``rad``(radians) and ``deg``(degree) as phase units.
```
oscargus, line 925:
```suggestion
        User can choose between ``'rad'`` (radians) and ``'deg'`` (degree) as phase units.
```

oscargus:
Looks good now. Just some formatting and to sort out the failing test (quite sure it is related to the indentation and empty lines).

oscargus:
Here are the error messages:

```
/home/runner/work/sympy/sympy/sympy/physics/control/control_plots.py:docstring of sympy.physics.control.control_plots.bode_plot:41: WARNING: Inline literal start-string without end-string.
/home/runner/work/sympy/sympy/sympy/physics/control/control_plots.py:docstring of sympy.physics.control.control_plots.bode_plot:41: WARNING: Inline literal start-string without end-string.
/home/runner/work/sympy/sympy/sympy/physics/control/control_plots.py:docstring of sympy.physics.control.control_plots.bode_plot:45: WARNING: Inline literal start-string without end-string.
/home/runner/work/sympy/sympy/sympy/physics/control/control_plots.py:docstring of sympy.physics.control.control_plots.bode_plot:45: WARNING: Inline literal start-string without end-string.
/home/runner/work/sympy/sympy/sympy/physics/control/control_plots.py:docstring of sympy.physics.control.control_plots.bode_magnitude_numerical_data:26: WARNING: Inline literal start-string without end-string.
/home/runner/work/sympy/sympy/sympy/physics/control/control_plots.py:docstring of sympy.physics.control.control_plots.bode_magnitude_numerical_data:26: WARNING: Inline literal start-string without end-string.
/home/runner/work/sympy/sympy/sympy/physics/control/control_plots.py:docstring of sympy.physics.control.control_plots.bode_phase_numerical_data:26: WARNING: Inline literal start-string without end-string.
/home/runner/work/sympy/sympy/sympy/physics/control/control_plots.py:docstring of sympy.physics.control.control_plots.bode_phase_numerical_data:26: WARNING: Inline literal start-string without end-string.
/home/runner/work/sympy/sympy/sympy/physics/control/control_plots.py:docstring of sympy.physics.control.control_plots.bode_phase_numerical_data:30: WARNING: Inline literal start-string without end-string.
/home/runner/work/sympy/sympy/sympy/physics/control/control_plots.py:docstring of sympy.physics.control.control_plots.bode_phase_numerical_data:30: WARNING: Inline literal start-string without end-string.
```

faze-geek:
> Looks good now. Just some formatting and to sort out the failing test (quite sure it is related to the indentation and empty lines).

Thanks @oscargus . After going through the `control_plots` html file for quite some time, seeing the actual mistake was an indentation error was definitely facepalm :)
Now I've made the changes !

sympy/physics/control/control_plots.py
```diff
@@ -686,6 +686,8 @@ def bode_magnitude_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs)
         The initial exponent of 10 of the semilog plot. Defaults to -5.
     final_exp : Number, optional
         The final exponent of 10 of the semilog plot. Defaults to 5.
+    freq_unit : string, optional
+        User can choose between 'rad/sec' (radians/second) and 'Hz' (Hertz) as frequency units.
```
oscargus, line 690:
```suggestion
        User can choose between ``'rad/sec'`` (radians/second) and ``'Hz'`` (Hertz) as frequency units.
```
(Just at I suggested last time.)

sympy/physics/control/control_plots.py
```diff
@@ -820,29 +841,45 @@ def bode_phase_numerical_data(system, initial_exp=-5, final_exp=5, **kwargs):
...
     _check_system(system)
     expr = system.to_expr()
-    _w = Dummy("w", real=True)
-    w_expr = expr.subs({system.var: I*_w})
-
-    phase = arg(w_expr)
+    freq_units = ('rad/sec', 'Hz')
+    phase_units = ('rad', 'deg')
+    if freq_unit not in freq_units:
+        raise ValueError('Only rad/sec and Hz are accepted frequency units.')
```
oscargus, line 847:
```suggestion
        raise ValueError('Only "rad/sec" and "Hz" are accepted frequency units.')
```

sympy/physics/control/control_plots.py
```diff
@@ -882,6 +919,10 @@ def bode_plot(system, initial_exp=-5, final_exp=5,
         If ``True``, the plot will have a grid. Defaults to True.
     show_axes : boolean, optional
         If ``True``, the coordinate axes will be shown. Defaults to False.
+    freq_unit : string, optional
+        User can choose between 'rad/sec' (radians/second) and 'Hz' (Hertz) as frequency units.
+    phase_unit : string, optional
+        User can choose between 'rad' (radians) and 'deg' (degree) as phase units.
```
oscargus, line 925: And here

faze-geek: Thanks @oscargus .I have made all suggested changes ,you can have a look now !

akshanshbhatt:
This PR looks almost complete now. Maybe you can add a couple of unit tests for the new units you have added before merging.

faze-geek:
> Maybe you can add a couple of unit tests for the new units you have added before merging.

I have add tests of possible errors in `test_errors()` .Rest could you give an example of what kind of test you expect above ,I'm not sure if I should make something new like def test_23074 or just modify some existing ones .

faze-geek:
@akshanshbhatt

oscarbenjamin:
Do the tests for the control plots actually pass when you run them on your own computer? They don't pass on my computer if I install matplotlib:
```console
$ bin/test sympy/physics/control/tests/test_control_plots.py 
============================================================= test process starts =============================================================
executable:         /Users/enojb/current/sympy/tmp/38venv/bin/python  (3.8.5-final-0) [CPython]
architecture:       64-bit
cache:              yes
ground types:       python 
numpy:              1.22.2
random seed:        67575540
hash randomization: on (PYTHONHASHSEED=3970762046)

sympy/physics/control/tests/test_control_plots.py[6] .F.FFF                                                                              [FAIL]

_______________________________________________________________________________________________________________________________________________
______________________________________ sympy/physics/control/tests/test_control_plots.py:test_pole_zero _______________________________________
  File "/Users/enojb/current/sympy/sympy/sympy/physics/control/tests/test_control_plots.py", line 99, in test_pole_zero
    assert _to_tuple(*pole_zero_numerical_data(tf1)) == \
AssertionError
_______________________________________________________________________________________________________________________________________________
___________________________________ sympy/physics/control/tests/test_control_plots.py:test_impulse_response ___________________________________
  File "/Users/enojb/current/sympy/sympy/sympy/physics/control/tests/test_control_plots.py", line 200, in test_impulse_response
    assert impulse_res_tester(tf3, exp3)
AssertionError
_______________________________________________________________________________________________________________________________________________
____________________________________ sympy/physics/control/tests/test_control_plots.py:test_step_response _____________________________________
  File "/Users/enojb/current/sympy/sympy/sympy/physics/control/tests/test_control_plots.py", line 249, in test_step_response
    assert step_res_tester(tf3, exp3)
AssertionError
_______________________________________________________________________________________________________________________________________________
____________________________________ sympy/physics/control/tests/test_control_plots.py:test_ramp_response _____________________________________
  File "/Users/enojb/current/sympy/sympy/sympy/physics/control/tests/test_control_plots.py", line 291, in test_ramp_response
    assert ramp_res_tester(tf3, 10, exp3, 1.5)
AssertionError

============================================= tests finished: 2 passed, 4 failed, in 6.11 seconds =============================================
DO *NOT* COMMIT!
```
The problem I think is that these are not being tested with matplotlib installed in CI because they aren't listed in the optional dependency job:
https://github.com/sympy/sympy/blob/93c1f7d0e40eaa477fb27758a9bd48df66ebb0b5/bin/test_optional_dependencies.py#L20-L67
What I don't understand is that people are contributing to this module without apparently noticing this:

1. Have you tried running the tests on your own computer?
2. Do the tests actually pass when you run them?

faze-geek:
@oscarbenjamin 
I have run tests after installing matplotlib in Cl and a few of those you pointed failed consistently .These failed even before any work on this pr .But I'm was not sure if they actually are incorrect  ,these failed tests were also reported in issue #23061 .
Since then I have tested my code modifications by plotting the graphs and  went ahead when they were as expected .
Do you suggest I open up an issue which @akshanshbhatt (I did a git blame to find the author) could fix or maybe skip them currently in this pr itself.

oscarbenjamin:
I'm going to open a PR that just disables the tests because we shouldn't have tests that fail as part of the test suite. I don't know whether the tests failing actually means that the code doesn't work or that the tests themselves are incorrect but I'm not going to spend time studying this module to figure that out while putting together the 1.10 release.

If it's easy to fix the tests or the code then I suggest doing that in a separate PR that can be applied to 1.10 as well as master along with enabling the tests in CI. If someone can put that together quickly then we won't need to merge my PR that just disables the tests but I'm not going to wait for that before releasing 1.10.

faze-geek:
> If it's easy to fix the tests or the code then I suggest doing that in a separate PR that can be applied to 1.10 as well as master along with enabling the tests in CI. If someone can put that together quickly then we won't need to merge my PR that just disables the tests but I'm not going to wait for that before releasing 1.10.
@oscarbenjamin 
Yes ,some of them are basic fixes like changing 0.2499999999999999 to 0.25 .I'll make them soon and open up a PR .

akshanshbhatt:
@oscarbenjamin, I am working on the tests. I have traced back the root cause of the errors and will be opening a PR soon.

akshanshbhatt:
> Yes ,some of them are basic fixes like changing 0.2499999999999999 to 0.25 .I'll make them soon and open up a PR .

It's more than that. Since we are dealing with NumPy arrays here, you can't simply fix them by asserting equivalence.

faze-geek:
> > Yes ,some of them are basic fixes like changing 0.2499999999999999 to 0.25 .I'll make them soon and open up a PR .
> 
> It's more than that. Since we are dealing with NumPy arrays here, you can't simply fix them by asserting equivalence.

Yes I realized that after making the comment .You can go forward with the fix .Thanks !

faze-geek:
@akshanshbhatt Here is the git diff if you need to change the tests manually -

```
     assert _to_tuple(*pole_zero_numerical_data(tf1)) == \
-        ((), ((-0.24999999999999994+1.3919410907075054j), (-0.24999999999999994-1.3919410907075054j)))
+        ((), ((-0.25+1.3919410907075056j), (-0.2499999999999999-1.3919410907075054j)))
     assert _to_tuple(*pole_zero_numerical_data(tf2)) == \
-        ((0.0,), ((-0.25+0.3227486121839514j), (-0.25-0.3227486121839514j)))
+        ((0.0,), ((-0.2500000000000001+0.32274861218395146j), (-0.2500000000000001-0.32274861218395146j)))
     assert _to_tuple(*pole_zero_numerical_data(tf3)) == \
-        ((0.0,), ((-0.5000000000000004+0.8660254037844395j),
-        (-0.5000000000000004-0.8660254037844395j), (0.9999999999999998+0j)))
+        ((0.0,), ((-0.5000000000000001+0.866025403784439j),
+        (-0.5000000000000001-0.866025403784439j), (1+0j)))
     assert _to_tuple(*pole_zero_numerical_data(tf7)) == \
-        (((-0.6722222222222222+0.8776898690157247j), (-0.6722222222222222-0.8776898690157247j)),
-        ((2.220446049250313e-16+1.2797182176061541j), (2.220446049250313e-16-1.2797182176061541j),
-        (-0.7657146670186428+0.5744385024099056j), (-0.7657146670186428-0.5744385024099056j),
-        (0.7657146670186427+0.5744385024099052j), (0.7657146670186427-0.5744385024099052j)))
+        (((-0.6722222222222218+0.877689869015724j), (-0.6722222222222222-0.8776898690157243j)),
+        ((-1.1102230246251565e-16+1.279718217606154j), (0.7657146670186428+0.5744385024099055j),
+        (-6.738674333953937e-17-1.2797182176061532j), (0.7657146670186425-0.574438502409905j),
+        (-0.7657146670186421+0.574438502409905j), (-0.7657146670186423-0.5744385024099056j)))
     assert _to_tuple(*pole_zero_numerical_data(ser1)) == \
         ((), (5.0, 0.0, 0.0, 0.0))
     assert _to_tuple(*pole_zero_numerical_data(par1)) == \
-        ((-5.645751311064592, -0.5000000000000008, -0.3542486889354093),
-        ((-0.24999999999999986+1.3919410907075052j),
-        (-0.24999999999999986-1.3919410907075052j), (-0.2499999999999998+0.32274861218395134j),
-        (-0.2499999999999998-0.32274861218395134j)))
+        (((-5.645751311064593-9.951207762139537e-18j), (-0.4999999999999997-1.7224902640350256e-16j),
+        (-0.35424868893540945+1.4425660087123082e-16j)), ((-0.25000000000000044+1.3919410907075045j),
+        (-0.25000000000000044-1.3919410907075058j), (-0.25000000000000006+0.3227486121839515j),
+        (-0.25-0.32274861218395146j)))
     assert _to_tuple(*pole_zero_numerical_data(tf8)) == \
         ((),
          ((-1.1641600331447917-3.545808351896439j),
```

oscargus:
These should be tested to match within an accuracy. Different platforms will provide slightly different results (32/64-bits, but possibly also different number of bits in extended floating-point formats etc).

Not related to this PR though. I think this looks OK now!

oscargus (approved):

faze-geek:
> Not related to this PR though. I think this looks OK now!

Thanks @oscargus and @akshanshbhatt , work on this pr helped me to understand quality of code expected from contributors .Is it good to be merged now?

akshanshbhatt:
> Is it good to be merged now?

Yes, I'll merge it now. Thanks for your contribution!

faze-geek:
@oscarbenjamin Just came across release notes for 1.10, https://github.com/sympy/sympy/wiki/Release-Notes-for-1.10 .I believe this pr solves a bunch of issues and it's release notes should be added under the `physics.control` notes domain . Thanks !

oscargus:
This is merged in 1.11.

PR #23077, opened 2022-02-15
Polys - RootOf now can handle polynomials with irrational algebraic coefficients

### LINKED ISSUE #22943: RootOf for polynomials with irrational algebraic coefficients
The set of algebraic numbers is the set of all numbers that are roots of polynomials having rational coefficients. Since the set of algebraic numbers is algebraically closed it should be possible to represent the roots of any polynomial with algebraic number coefficients as the root of a polynomial with rational coefficients.

Currently rootof (and so also `real_roots`, `all_roots` etc) support: only polynomials with rational coefficients:
```python
In [15]: p = sqrt(2) * x ** 3 + sqrt(3) * x + cbrt(3)

In [16]: rootof(p, 0)
NotImplementedError
```

There is a method that can extend this which is described in
```
Computer Algebra: Symbolic and Algebraic Computation
Second Edition
Edited by
B. Buchberger, G. E. Collins, and R. Loos
in cooperation with R. Albrecht
```
The algorithm is described in the chapter "Loos, R.: Computing in Algebraic Extensions" as "Algorithm 3 (NORMAL)" and is something like:

- Input: a polynomial `B*(x)` with coefficients `a_0`, `a_1, ..., a_n` that are algebraic numbers.
- Output: a polynomial `C(x)` with rational coefficients such that every root `B*(x)` is also a root of `C(x)`

1. Find a primitive element `alpha` with minimal polynomial `A` for an extension field that contains all the coefficients `a_i` and also polynomials `B_i` with rational coefficients such that for each `i` we have `a_i = B_i(alpha)`.
2. Compute `D(y) = gcd(A(y), B_n(y))` and if it has positive degree compute `A* = A/D`, now reduce all `B_i` modulo `A*` and set `A = A*`
3. Define `B(y, x) = B_0(y) + B_1(y)*x + ... + B_n(y)*x**n.
4. Compute `C(x) = resultant(A(y), B(y, x), y)` and return `C(x)`

Here's a demonstration (skipping step 2):
```python
In [38]: A, rA, Bs = primitive_element(Poly(p, x).all_coeffs(), y, ex=True)

In [39]: B = sum(Poly(Bi, y).as_expr() * x ** i for i, Bi in enumerate(Bs[::-1]))

In [40]: r = resultant(A, B, y)

In [41]: nroots(p, 3)
Out[41]: [-1.01, -0.107 - 0.999⋅ⅈ, -0.107 + 0.999⋅ⅈ]

In [42]: nroots(r, 3)
Out[42]: 
[-1.01, -1.0, 0.493, 0.638, 1.59, 2.06, -1.04 - 0.545⋅ⅈ, -1.04 + 0.545⋅ⅈ, -1.01 - 0.92⋅ⅈ, -1.01 + 0
.92⋅ⅈ, -0.936 - 0.243⋅ⅈ, -0.936 + 0.243⋅ⅈ, -0.859 - 1.15⋅ⅈ, -0.859 + 1.15⋅ⅈ, -0.735 - 0.441⋅ⅈ, -0.7
35 + 0.441⋅ⅈ, -0.629 - 1.26⋅ⅈ, -0.629 + 1.26⋅ⅈ, -0.417 - 0.566⋅ⅈ, -0.417 + 0.566⋅ⅈ, -0.368 - 1.21⋅ⅈ
, -0.368 + 1.21⋅ⅈ, -0.107 - 0.999⋅ⅈ, -0.107 + 0.999⋅ⅈ, -0.035 - 0.594⋅ⅈ, -0.035 + 0.594⋅ⅈ, 0.182 - 
0.664⋅ⅈ, 0.182 + 0.664⋅ⅈ, 0.355 - 0.458⋅ⅈ, 0.355 + 0.458⋅ⅈ, 0.413 - 0.335⋅ⅈ, 0.413 + 0.335⋅ⅈ, 1.81 
- 0.216⋅ⅈ, 1.81 + 0.216⋅ⅈ, 1.99 - 0.152⋅ⅈ, 1.99 + 0.152⋅ⅈ]

In [43]: p
Out[43]: 
    3       2          3 ___
√2⋅x  + √3⋅x  + √3⋅x + ╲╱ 3 

In [44]: r
Out[44]: 
    36        34         33         32         31         30          29          28          27   
64⋅x   - 576⋅x   - 1152⋅x   + 1584⋅x   + 8640⋅x   + 8640⋅x   - 17280⋅x   - 57780⋅x   - 47520⋅x   + 

       26           25           24          23           22           21           20           19
68364⋅x   + 217080⋅x   + 205389⋅x   - 69012⋅x   - 428166⋅x   - 535572⋅x   - 246645⋅x   + 227448⋅x  

           18           17           16           15          14          13          12          1
 + 542232⋅x   + 548208⋅x   + 354699⋅x   + 153900⋅x   + 40014⋅x   - 10692⋅x   - 28917⋅x   - 22356⋅x 

1          10         9         8         7        6     
  - 12150⋅x   - 9720⋅x  - 7290⋅x  - 2916⋅x  - 486⋅x  + 81
```
I wonder if this should be used automatically by `rootof` or in other situations such as `real_roots` etc: `rootof(p)` could return a `CRootOf` for a rational polynomial. Alternatively it could still represent itself as a polynomial with algebraic coefficients but there could be a way of converting it to a polynomial with rational coefficients. Computing a `primitive_element` can be very slow so I'm not sure if this should be enabled by default or not.

praneethratna:
Fixes #22943 and now `rootof(sqrt(2) * x ** 3 + sqrt(3) * x ** 2 + sqrt(3) * x + cbrt(3), 0)` returns 
```
CRootOf(64*x**36 - 576*x**34 - 1152*x**33 + 1584*x**32 + 8640*x**31 + 8640*x**30 - 17280*x**29 - 57780*x**28 - 47520*x**27 + 68364*x**26 + 217080*x**25 + 205389*x**24 - 69012*x**23 - 428166*x**22 - 535572*x**21 - 246645*x**20 + 227448*x**19 + 542232*x**18 + 548208*x**17 + 354699*x**16 + 153900*x**15 + 40014*x**14 - 10692*x**13 - 28917*x**12 - 22356*x**11 - 12150*x**10 - 9720*x**9 - 7290*x**8 - 2916*x**7 - 486*x**6 + 81, 0)
``` 
instead of `NotImplementedError` .
#### Release Notes

* polys
   * `rootof` now can handle polynomials with irrational algebraic coefficients.

### DIFF
--- a/sympy/polys/rootoftools.py
+++ b/sympy/polys/rootoftools.py
@@ -351,7 +351,7 @@ class ComplexRootOf(RootOf):
         if roots is not None:
             return roots[index]
 
-        if all((i.is_algebraic and (not i.is_symbol)) for i in poly.all_coeffs()):
+        if all((i.is_algebraic and i.is_number) for i in poly.all_coeffs()):
             poly = cls ._roots_Algebraic(poly, dom)
         else:
             raise NotImplementedError("CRootOf is not supported over %s" % poly)
--- a/sympy/polys/rootoftools.py
+++ b/sympy/polys/rootoftools.py
@@ -353,6 +353,8 @@ class ComplexRootOf(RootOf):
 
         if all((i.is_algebraic and (not i.is_symbol)) for i in poly.all_coeffs()):
             poly = cls ._roots_Algebraic(poly, dom)
+        else:
+            raise NotImplementedError("CRootOf is not supported over %s" % poly)
 
         coeff, poly = preprocess_roots(poly)
         dom = poly.get_domain()
--- a/sympy/polys/rootoftools.py
+++ b/sympy/polys/rootoftools.py
@@ -351,13 +351,11 @@ class ComplexRootOf(RootOf):
         if roots is not None:
             return roots[index]
 
+        if all((i.is_algebraic and (not i.is_symbol)) for i in poly.all_coeffs()):
+            poly = cls ._roots_Algebraic(poly, dom)
+
         coeff, poly = preprocess_roots(poly)
         dom = poly.get_domain()
-
-        if not dom.is_ZZ:
-            poly = cls._roots_Algebraic(poly)
-            coeff, poly = preprocess_roots(poly)
-            dom = poly.get_domain()
 
         root = cls._indexed_root(poly, index)
         return coeff * cls._postprocess_root(root, radicals)
@@ -763,29 +761,23 @@ class ComplexRootOf(RootOf):
         return roots
 
     @classmethod
-    def _roots_Algebraic(cls, poly):
+    def _roots_Algebraic(cls, poly, dom):
         r"""Converts a Polynomial 'B(x)' with algebriac coefficients
         to a Polynomial 'C(x)' with rational coefficients such that
         every root of 'B(x)' is a root of 'C(x)'.
         """
         from sympy.polys.numberfields import primitive_element
-        from sympy.polys.polytools import gcd, resultant, reduced, div
-        x, y = symbols('x y')
-        coeffs = Poly(poly, x).all_coeffs()
+        from sympy.polys.polytools import resultant
+        gen = poly.gen
+        coeffs = Poly(poly, gen).all_coeffs()
+        m = symbols('m')
         try:
-            A, r_A, B_s = primitive_element(coeffs, y, ex=True)
+            A, r_A, B_s = primitive_element(coeffs, m, ex=True)
         except NotInvertible:
             raise NotImplementedError("Cannot convert %s to an algebraic extension" % poly)
-        B_n = Poly(B_s[::-1][len(B_s)-1], y).as_expr()
-        D = gcd(A, B_n)
 
-        if not D.is_Number and D.degree() > 0:
-            A_hat = div(A, D)
-            B_s, A_hat = reduced(A, [Poly(Bi, y).as_expr() for Bi in B_s[::-1]])
-            A = A_hat
-
-        B = sum(Poly(Bi, y).as_expr() * x ** i for i, Bi in enumerate(B_s[::-1]))
-        poly = PurePoly(resultant(A, B, y), x, domain='EX')
+        B = sum(Poly(Bi, m).as_expr() * gen ** i for i, Bi in enumerate(B_s[::-1]))
+        poly = PurePoly(resultant(A, B, m), gen, domain=dom)
         return poly
 
     @classmethod
--- a/sympy/polys/rootoftools.py
+++ b/sympy/polys/rootoftools.py
@@ -12,7 +12,8 @@ from sympy.polys.polyerrors import (
     MultivariatePolynomialError,
     GeneratorsNeeded,
     PolynomialError,
-    DomainError)
+    DomainError,
+    NotInvertible)
 from sympy.polys.polyfuncs import symmetrize, viete
 from sympy.polys.polyroots import (
     roots_linear, roots_quadratic, roots_binomial,
@@ -354,7 +355,9 @@ class ComplexRootOf(RootOf):
         dom = poly.get_domain()
 
         if not dom.is_ZZ:
-            raise NotImplementedError("CRootOf is not supported over %s" % dom)
+            poly = cls._roots_Algebraic(poly)
+            coeff, poly = preprocess_roots(poly)
+            dom = poly.get_domain()
 
         root = cls._indexed_root(poly, index)
         return coeff * cls._postprocess_root(root, radicals)
@@ -758,6 +761,32 @@ class ComplexRootOf(RootOf):
         for root in getattr(cls, method)(poly):
             roots.append(coeff*cls._postprocess_root(root, radicals))
         return roots
+
+    @classmethod
+    def _roots_Algebraic(cls, poly):
+        r"""Converts a Polynomial 'B(x)' with algebriac coefficients
+        to a Polynomial 'C(x)' with rational coefficients such that
+        every root of 'B(x)' is a root of 'C(x)'.
+        """
+        from sympy.polys.numberfields import primitive_element
+        from sympy.polys.polytools import gcd, resultant, reduced, div
+        x, y = symbols('x y')
+        coeffs = Poly(poly, x).all_coeffs()
+        try:
+            A, r_A, B_s = primitive_element(coeffs, y, ex=True)
+        except NotInvertible:
+            raise NotImplementedError("Cannot convert %s to an algebraic extension" % poly)
+        B_n = Poly(B_s[::-1][len(B_s)-1], y).as_expr()
+        D = gcd(A, B_n)
+
+        if not D.is_Number and D.degree() > 0:
+            A_hat = div(A, D)
+            B_s, A_hat = reduced(A, [Poly(Bi, y).as_expr() for Bi in B_s[::-1]])
+            A = A_hat
+
+        B = sum(Poly(Bi, y).as_expr() * x ** i for i, Bi in enumerate(B_s[::-1]))
+        poly = PurePoly(resultant(A, B, y), x, domain='EX')
+        return poly
 
     @classmethod
     def clear_cache(cls):
--- a/sympy/polys/tests/test_rootoftools.py
+++ b/sympy/polys/tests/test_rootoftools.py
@@ -15,7 +15,7 @@ from sympy.core.numbers import (Float, I, Rational)
 from sympy.core.relational import Eq
 from sympy.core.singleton import S
 from sympy.functions.elementary.exponential import (exp, log)
-from sympy.functions.elementary.miscellaneous import sqrt
+from sympy.functions.elementary.miscellaneous import sqrt, cbrt
 from sympy.functions.elementary.trigonometric import tan
 from sympy.integrals.integrals import Integral
 from sympy.polys.orthopolys import legendre_poly
@@ -173,6 +173,17 @@ def test_CRootOf_is_real():
 
 def test_CRootOf_is_complex():
     assert rootof(x**3 + x + 3, 0).is_complex is True
+
+
+def test_CRootOf_Algebraic():
+    assert rootof(sqrt(2) * x ** 3 + sqrt(3) * x ** 2 + sqrt(3) * x + cbrt(3), 0) == \
+        CRootOf(64*x**36 - 576*x**34 - 1152*x**33 + 1584*x**32 + 8640*x**31 + 8640*x**30 \
+        - 17280*x**29 - 57780*x**28 - 47520*x**27 + 68364*x**26 + 217080*x**25 \
+        + 205389*x**24 - 69012*x**23 - 428166*x**22 - 535572*x**21 - 246645*x**20 \
+        + 227448*x**19 + 542232*x**18 + 548208*x**17 + 354699*x**16 + 153900*x**15 \
+        + 40014*x**14 - 10692*x**13 - 28917*x**12 - 22356*x**11 - 12150*x**10 \
+        - 9720*x**9 - 7290*x**8 - 2916*x**7 - 486*x**6 + 81, 0)
+    raises(NotImplementedError, lambda: rootof(sqrt(2) * x ** 3 + sqrt(3) * x + cbrt(3), 0))
 
 
 def test_CRootOf_subs():
--- a/sympy/polys/tests/test_rootoftools.py
+++ b/sympy/polys/tests/test_rootoftools.py
@@ -5,7 +5,6 @@ from sympy.polys.rootoftools import (rootof, RootOf, CRootOf, RootSum,
     _pure_key_dict as D)
 
 from sympy.polys.polyerrors import (
-    DomainError,
     MultivariatePolynomialError,
     GeneratorsNeeded,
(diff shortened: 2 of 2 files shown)


sympy/polys/rootoftools.py
```diff
@@ -354,7 +355,9 @@ def __new__(cls, f, x, index=None, radicals=False, expand=True):
         dom = poly.get_domain()
 
         if not dom.is_ZZ:
-            raise NotImplementedError("CRootOf is not supported over %s" % dom)
+            poly = cls._roots_Algebraic(poly)
```
oscarbenjamin, line 358: The check should be for an AlgebraicField rather than just `not dom.is_ZZ`.

sympy/polys/rootoftools.py
```diff
@@ -759,6 +762,32 @@ def _get_roots(cls, method, poly, radicals):
...
+    @classmethod
+    def _roots_Algebraic(cls, poly):
+        r"""Converts a Polynomial 'B(x)' with algebriac coefficients
+        to a Polynomial 'C(x)' with rational coefficients such that
+        every root of 'B(x)' is a root of 'C(x)'.
+        """
+        from sympy.polys.numberfields import primitive_element
+        from sympy.polys.polytools import gcd, resultant, reduced, div
+        x, y = symbols('x y')
+        coeffs = Poly(poly, x).all_coeffs()
```
oscarbenjamin, line 774: What exactly does `Poly(poly, x).all_coeffs()` do?

sympy/polys/tests/test_rootoftools.py
```diff
@@ -175,6 +175,17 @@ def test_CRootOf_is_complex():
...
 
+def test_CRootOf_Algebraic():
+    assert rootof(sqrt(2) * x ** 3 + sqrt(3) * x ** 2 + sqrt(3) * x + cbrt(3), 0) == \
+        CRootOf(64*x**36 - 576*x**34 - 1152*x**33 + 1584*x**32 + 8640*x**31 + 8640*x**30 \
+        - 17280*x**29 - 57780*x**28 - 47520*x**27 + 68364*x**26 + 217080*x**25 \
+        + 205389*x**24 - 69012*x**23 - 428166*x**22 - 535572*x**21 - 246645*x**20 \
+        + 227448*x**19 + 542232*x**18 + 548208*x**17 + 354699*x**16 + 153900*x**15 \
+        + 40014*x**14 - 10692*x**13 - 28917*x**12 - 22356*x**11 - 12150*x**10 \
+        - 9720*x**9 - 7290*x**8 - 2916*x**7 - 486*x**6 + 81, 0)
+    raises(NotImplementedError, lambda: rootof(sqrt(2) * x ** 3 + sqrt(3) * x + cbrt(3), 0))
```
oscarbenjamin, line 186: This shouldn't raise `NotImplementedError`. It looks like there's a bug in `primitive_element` that causes it to raise when there's a zero in the inputs.

sympy/polys/rootoftools.py
praneethratna: How CAN i check for AlgebraicField instead of checking for domain?

sympy/polys/rootoftools.py
```diff
@@ -759,6 +762,32 @@ def _get_roots(cls, method, poly, radicals):
...
+            A, r_A, B_s = primitive_element(coeffs, y, ex=True)
+        except NotInvertible:
+            raise NotImplementedError("Cannot convert %s to an algebraic extension" % poly)
+        B_n = Poly(B_s[::-1][len(B_s)-1], y).as_expr()
+        D = gcd(A, B_n)
+
+        if not D.is_Number and D.degree() > 0:
+            A_hat = div(A, D)
+            B_s, A_hat = reduced(A, [Poly(Bi, y).as_expr() for Bi in B_s[::-1]])
+            A = A_hat
```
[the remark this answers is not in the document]
praneethratna: Sorry, missed that part. Will remove this code.

sympy/polys/rootoftools.py
```diff
@@ -759,6 +762,32 @@ def _get_roots(cls, method, poly, radicals):
...
+        """
+        from sympy.polys.numberfields import primitive_element
+        from sympy.polys.polytools import gcd, resultant, reduced, div
+        x, y = symbols('x y')
+        coeffs = Poly(poly, x).all_coeffs()
+        try:
+            A, r_A, B_s = primitive_element(coeffs, y, ex=True)
+        except NotInvertible:
+            raise NotImplementedError("Cannot convert %s to an algebraic extension" % poly)
+        B_n = Poly(B_s[::-1][len(B_s)-1], y).as_expr()
```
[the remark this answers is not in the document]
praneethratna: Yeah, But since the 2nd step of the algorithm is not needed here, this line can be removed.

sympy/polys/tests/test_rootoftools.py
praneethratna: Yes, But I was unable to find a fix for that bug so raised a `NotImplementedError` for the polynomial.

sympy/polys/rootoftools.py
oscarbenjamin: Oh, I thought that input here was already a list of coefficients.

sympy/polys/rootoftools.py
praneethratna: I'm not sure of how to check for AlgebraicField here , should i check for `dom.is_QQ` ?

oscarbenjamin: The check should be that all coefficients are algebraic which could be done with Expr:
```python
In [1]: sqrt(2).is_algebraic
Out[1]: True
```
You have to be careful about `Symbol('x', algebraic=True)` though.

praneethratna: So will this `all(i.is_algebraic for i in poly.all_coeffs())` would work?

oscarbenjamin: It might also be necessary to check that there are no symbols.

sympy/polys/rootoftools.py
```diff
@@ -759,6 +762,32 @@ def _get_roots(cls, method, poly, radicals):
...
+        r"""Converts a Polynomial 'B(x)' with algebriac coefficients
+        to a Polynomial 'C(x)' with rational coefficients such that
+        every root of 'B(x)' is a root of 'C(x)'.
+        """
+        from sympy.polys.numberfields import primitive_element
+        from sympy.polys.polytools import gcd, resultant, reduced, div
+        x, y = symbols('x y')
+        coeffs = Poly(poly, x).all_coeffs()
+        try:
+            A, r_A, B_s = primitive_element(coeffs, y, ex=True)
```
jksuom, line 776: Better take only those coeffs that are not rational to speed up primitive_element.

sympy/polys/tests/test_rootoftools.py
skieffer: It looks like the trouble occurs in this loop

https://github.com/sympy/sympy/blob/c1e1f50cb68caa4976b63714fbe7cb84c3b4a342/sympy/polys/numberfields/subfield.py#L368-L380

when `p` is linear, so `L == QQ`. In particular, the call to `_switch_domain()` seems to be messed up, since there isn't really anything to `inject()`.

Maybe a fix could be as simple as starting the loop with something like this?

```python
    for ext in extension[1:]:
        p = minimal_polynomial(ext, x, polys=True)
        if p.degree() == 1:
            coeffs.append(0)
            continue
        L = QQ.algebraic_field((p, ext))
```

I haven't tried it against all unit tests, but it seems to allow the case at hand to pass...

praneethratna: Yeah, seems like it and also it is passing all the unit tests as well.
But with this change along with changes with this PR, many tests of `rootoftools.py` are raising `TypeError: cannot unpack non-iterable NoneType object` .

sympy/polys/rootoftools.py
praneethratna: Yeah, maybe that can be done.

sympy/polys/tests/test_rootoftools.py
jksuom: I would rather try something like this:

    for ext in extension[1:]:
        if ext.is_Rational:
            coeffs.append(0)    # rational ext is not included in the expression of a primitive element
            reps.append(K.convert(ext))    # but it is included in reps
            continue
        p = minimal_polynomial(ext, x, polys=True)
        L = QQ.algebraic_field((p, ext))

praneethratna: This change is working and all the unit tests are passing, so should i include these changes along with this PR or in a separate PR along with relevant tests?

oscarbenjamin: Let's fix the bug in a separate PR.

praneethratna: > I would rather try something like this:
> 
> ```
> for ext in extension[1:]:
>     if ext.is_Rational:
>         coeffs.append(0)    # rational ext is not included in the expression of a primitive element
>         reps.append(K.convert(ext))    # but it is included in reps
>         continue
>     p = minimal_polynomial(ext, x, polys=True)
>     L = QQ.algebraic_field((p, ext))
> ```

`ext` should be sympified i guess because I'm getting an error `AttributeError: 'int' object has no attribute 'is_Rational'`.

praneethratna:
I'm not sure why rootof(y*x**3 + y*x + 2*y, x, 0) is raising `NotImplementedError` . Seems like `NotImplementedError` should be returned when all the coefficients are not symbols but I'm not sure how that can be done.

sympy/polys/rootoftools.py
```diff
@@ -350,12 +351,14 @@ def __new__(cls, f, x, index=None, radicals=False, expand=True):
         if roots is not None:
             return roots[index]
 
+        if all((i.is_algebraic and (not i.is_symbol)) for i in poly.all_coeffs()):
```
oscarbenjamin, line 354: I'm not sure exactly what `is_symbol` means but `is_number` would be a better check here. If `is_number` is true then it means that the expression has no free symbols and can be evaluated to a complex number with `evalf`. Presumably this should also check `is_rational` as well although I haven't followed all the logic.

oscarbenjamin:
RootOf is supposed to order the roots so that the real roots come first in increasing order and then the non-real roots, ordered by real parts first and imaginary parts second. I don't see any code here that tries to figure out how to map the nth roots of the polynomial with irrational coefficients to the mth root of the higher degree polynomial with rational coefficients.

Also I'm not sure that this transformation should be done automatically. Possibly it would be good to represent the RootOf with irrational coefficients just as it is without automatically converting it into a RootOf with rational coefficients.

This will need a lot more testing including tests for things like `real_roots`, `Poly(...).all_roots` etc.

praneethratna:
> RootOf is supposed to order the roots so that the real roots come first in increasing order and then the non-real roots, ordered by real parts first and imaginary parts second. I don't see any code here that tries to figure out how to map the nth roots of the polynomial with irrational coefficients to the mth root of the higher degree polynomial with rational coefficients.

Right now `rootof` returns `CRoot(equivalent rational polynomial)`  for polynomials with irrational coeffefficients , so we have to find the roots for the equation with rational coefficients and then map them with roots of the irrational polynomial?

> Also I'm not sure that this transformation should be done automatically. Possibly it would be good to represent the RootOf with irrational coefficients just as it is without automatically converting it into a RootOf with rational coefficients.

I'm not sure what to do here, could you explain it?

oscarbenjamin:
> so we have to find the roots for the equation with rational coefficients and then map them with roots of the irrational polynomial?

Yes. I think the way to do this is to substitute the roots of the rational polynomial into the irrational one and prove numerically that it gives something nonzero e.g.:
```python
In [11]: p_rational = rootof(p_irrational, x, 0).poly.as_expr()

In [12]: p_irrational = sqrt(2) * x ** 3 - x - 1

In [13]: p_rational = rootof(p_irrational, x, 0).poly.as_expr()

In [14]: p_irrational
Out[14]: 
    3        
√2⋅x  - x - 1

In [15]: p_rational
Out[15]: 
   6    2          
2⋅x  - x  - 2⋅x - 1

In [16]: [p_irrational.subs(x, r).n(2) for r in Poly(p_rational).all_roots()]
Out[16]: [-0.73, 0.e-109, -0.e-117 + 0.e-117⋅ⅈ, -0.e-117 - 0.e-117⋅ⅈ, -2.6 + 2.0⋅ⅈ, -2.6 - 2.0⋅ⅈ]
```
Unfortunately that's quite slow though so maybe there is a better way. Or maybe we just need to improve the speed of the root refinement code.

It is necessary to do this though because otherwise we will get the wrong roots of the rational polynomial:
```python
In [17]: rootof(p_irrational, 0)
Out[17]: 
       ⎛   6    2             ⎞
CRootOf⎝2⋅x  - x  - 2⋅x - 1, 0⎠

In [18]: _.n(2)
Out[18]: -0.64

In [19]: nroots(p_irrational, 2)
Out[19]: [1.2, -0.58 - 0.53⋅ⅈ, -0.58 + 0.53⋅ⅈ]
```

PR #23083, opened 2022-02-16
Removed #doctest: +SKIP

ghost:
Removed #doctest: +SKIP *(wherever possible)*

Helps #6959

Removed #doctest: +SKIP from the following files:
*modules/evalf.rst
tutorial/gotchas.rst
combinatorics/perm_groups.py
holonomic.py (Not removing from here now, since some checks are failing when removed from this file)
stats/rv.py*
#### Other comments
 - [x] Opened on a separate branch.
 - [ ] Has release note.

#### Release Notes

NO ENTRY


KuldeepBorkar:
When run locally,  **bin/doctest --force-colors** shows no fails for holonomic.py

doc/src/modules/evalf.rst
```diff
@@ -343,7 +343,7 @@ digits in a fraction of a second with a simple command:
     >>> f = factorial
     >>> n = Symbol('n', integer=True)
     >>> R = 9801/sqrt(8)/Sum(f(4*n)*(1103+26390*n)/f(n)**4/396**(4*n),
-    ...                         (n, 0, oo)) #doctest: +SKIP
+    ...                         (n, 0, oo))
```
smichr, line 346: Was this skipped because it is slow?

sympy/stats/rv.py
```diff
@@ -1775,7 +1775,7 @@ def sample_stochastic_process(process):
     >>> from sympy import Matrix
     >>> T = Matrix([[0.5, 0.2, 0.3],[0.2, 0.5, 0.3],[0.2, 0.3, 0.5]])
     >>> Y = DiscreteMarkovChain("Y", [0, 1, 2], T)
-    >>> next(sample_stochastic_process(Y)) in Y.state_space # doctest: +SKIP
```
smichr, line 1778: Why was this one skipped? Was it previously nondeterministic?

sympy/stats/rv.py
```diff
@@ -1775,7 +1775,7 @@ def sample_stochastic_process(process):
...
+    >>> next(sample_stochastic_process(Y)) in Y.state_space
     True
     >>> next(sample_stochastic_process(Y))  # doctest: +SKIP
```
smichr, line 1780: Can this one be unskipped, too?

doc/src/modules/evalf.rst
KuldeepBorkar: @smichr I don't know the real reason why it was skipped, I just checked where the #doctest: +SKIP is used and does it make any difference when they are removed or are they necessary for a specific reason for knowing this I just run tests(checks were OK) and there were no fails so I just thought they can be removed.
*Is there a way to check if this was slow? I am rookie in these things ; (*

smichr: It takes about 2 seconds to calculate and print on `replit.com`. I suppose part of the question is whether we really need to keep computing this in a doctest since it (should) already be tested in the tests. We can show that is is possible but don't actually need to recompute it every time. @asmeurer  ?

doc/src/tutorial/gotchas.rst
```diff
@@ -266,7 +266,7 @@ division, depending on whether you are in Python 2 or Python 3, and depending
 on whether or not you have run ``from __future__ import division``:
```
oscargus, line 266: This should be rewritten as ``from __future__ import division`` is Python 2 which is no longer supported.

KuldeepBorkar: @oscargus How about I mention the `SymPy` version too from where Python 2 is not supported like:
``from __future__ import division`` for Python 2 which is no longer supported for versions above `SymPy 1.5.1`

.mailmap
```diff
@@ -629,6 +629,7 @@ Kristian Brünn <hello@kristianbrunn.com> Kristianmitk <hello@kristianbrunn.com>
 Krit Karan <kritkaran.b13@iiits.in> <kritkaran94@users.noreply.github.com>
 Kshitij <kshitijparwani.mat18@itbhu.ac.in> Kshitij Parwani <44468674+P-Kshitij@users.noreply.github.com>
 Kshitij Saraogi <KshitijSaraogi@gmail.com>
+Kuldeep <kuldeepborkar.jr@gmail.com>
 Kuldeep Borkar Jr <kuldeepborkarjr765@gmail.com> Kuldeep Borkar Jr <74557588+KuldeepBorkar@users.noreply.github.com>
```
oscarbenjamin, line 633: This will generate two separate entries in the AUTHORS file, Are `Kuldeep <kuldeepborkar.jr@gmail.com>` and `Kuldeep Borkar Jr <kuldeepborkarjr765@gmail.com>` the same person or different people?

KuldeepBorkar: > That makes the second name/email combination an alias of the first.

@oscarbenjamin I see, 
```
name <email> alias
```
Means if I want to add more alias I can just do like this;
```
Kuldeep Borkar Jr <kuldeepborkarjr765@gmail.com> alias
Kuldeep Borkar Jr <kuldeepborkarjr765@gmail.com> alias
```
In the authors file I checked this name and email has been added so I think I should use this as first name/email and make alias in front of that.

Thank you so much for the clarification : )

oscargus:
Can you please rebase and squash the commits? 20 commits for 15 lines is a bit much...

oscargus:
It is hard to say without seeing the whole context. But the idea seems correct, keep the first and fixup the rest. This may have something to do with the merge commits though, so maybe you should keep those for now.

I think one problem here is that you have merged you documentation branch and then reverted those commits (while it would have been better to just not merge it).

If you can paste the list you are trying to apply it may be easier to help out.

(You can always use `git rebase --abort` if you get stuck in these situations to just start all over again.)

KuldeepBorkar:
> It is hard to say without seeing the whole context. But the idea seems correct, keep the first and fixup the rest. This may have something to do with the merge commits though, so maybe you should keep those for now.
> I think one problem here is that you have merged you documentation branch and then reverted those commits (while it would have been better to just not merge it).
> If you can paste the list you are trying to apply it may be easier to help out.

@oscargus I consider by the list you mean the list of commits which I can see after the `rebase` command.

commits after rebase

*I think some of the commits I made from my master branch then I created this branch from the master so, it showed that commits too( Later I realized this was a mistake to make commits from master)*

eagleoflqj:
Let's try to make it easy. You want one commit that contains the changes shown in the "File changed" page on GitHub, right?
Follow these command:
`git reset --hard master`, so that you discard changes and switch to your local master branch, which should be behind sympy's master several commits.
`git pull https://github.com/sympy/sympy master`, so that you are up-to-date to sympy's master
`git diff bb7275c103..f721225540 > ~/patch`, this calculates the changes we need and stores it in your home dir. bb7275c103 is the last common commit of your branch and sympy's master, and f721225540 is your latest commit.
`git branch -D neww && git checkout -b neww`, this deletes your neww branch and creates a new one based on master.
`git apply ~/patch`, this needs no explanation, just use `git status` to see the result.
`git add -A` and `git commit -m 'YOUR COMMIT MSG'`, also needs no explanation.
`git push -f origin neww`, you are done!
Remember to remove the temporary patch file `rm ~/patch`

eagleoflqj:
I suggest you learn git more and DIY. But in case you don't want now, you may temporarily give me write access to your repo at https://github.com/KuldeepBorkar/sympy/settings/access so that I could clean it for you.

KuldeepBorkar:
> I suggest you learn git more and DIY. But in case you don't want now, you may temporarily give me write access to your repo at https://github.com/KuldeepBorkar/sympy/settings/access so that I could clean it for you.

@eagleoflqj It's okay for me to give access but I think I should try this on my own so that I can learn more : )

eagleoflqj:
@oscargus Is it good to go?

sympy/stats/rv.py
eagleoflqj: No. It produces 0,1,2 randomly.

eagleoflqj:
This should be good to go @sympy/developers-with-push-access-to-everything (I don't know if it's the proper name to ping

sympy/combinatorics/perm_groups.py
```diff
@@ -3694,7 +3694,7 @@ def schreier_sims_random(self, base=None, gens=None, consec_succ=10,
         >>> from sympy.combinatorics.named_groups import SymmetricGroup
         >>> S = SymmetricGroup(5)
         >>> base, strong_gens = S.schreier_sims_random(consec_succ=5)
-        >>> _verify_bsgs(S, base, strong_gens) #doctest: +SKIP
```
eagleoflqj, line 3697: This is causing trouble randomly.

KuldeepBorkar: @eagleoflqj The doctest should be skipped here?
Also, I don't know how things work after a PR gets merged like, this PR is merged now so if I want to make any further changes here then how should I be doing that?
The regular way we do changes to PR that is not merged yet,
OR
The PR needs to be unmerged first,
OR
A new PR should be created to make further changes?

sympy/stats/rv.py
KuldeepBorkar: @eagleoflqj means there should be no changes here(the doctest should be skipped)?

sympy/combinatorics/perm_groups.py
eagleoflqj: According to docstring of `schreier_sims_random`, this indeed can give wrong answer in a low probability. The easiest way is to add `# doctest: +SKIP` back in a new PR, though there might also be other way like use a fixed random seed.

PR #23087, opened 2022-02-16
Series/Limits : Fixes leading term for exponential functions involving AccumBounds based arguments

### LINKED ISSUE #18473: Bug in limit
`limit(cos (x)**(1/x),x,oo)` should give `AccumBounds` instead of `1`.

anutosh491:

Fixes #18473 
On Master
```
>>> limit(sin(x)**(1/x), x, oo)
1
>>> limit(cos(x)**(1/x), x, oo)
1
>>> limit(tan(x)**(1/x), x, oo)
1
>>> exp(x*log(cos(1/x))).as_leading_term(x)
1
>>> exp(x*log(sin(1/x))).as_leading_term(x)
1
>>> exp(x*log(tan(1/x))).as_leading_term(x)
1
>>> log(cos(1/x)).as_leading_term(x)
log(AccumBounds(-1, 1))
>>> log(sin(1/x)).as_leading_term(x)
log(AccumBounds(-1, 1))
>>> log(tan(1/x)).as_leading_term(x)
log(AccumBounds(-oo, oo))
```
The limits shown above cannot be correctly claimed to be `1` . For eg the following
image

Firstly the function isn't continuous in the real domain as `x` tends to `oo` and even if it does the answer is not going to be `1` .
This is because something like `log(cos(1/x)).as_leading_term(x)` returns `log(AccumBounds(-1, 1))` which wouldn't be supportive in the `real` domain . Unfortunately a `subs` operation for `x` by `0` is taking place changing the `x*log(AccumBounds(-1, 1))` to `0` and returning `exp(0)` which is `1` . Hence the Pr fixes the above not really correct result.

On Committed branch . `S.NaN` seems to be the best option we have to tackle non real values of `log` inside `AccumBounds` !
```
>>> x = Symbol('x')
>>> limit(sin(x)**(1/x), x, oo)
Limit(sin(x)**(1/x), x, oo, dir='-')
>>> limit(cos(x)**(1/x), x, oo)
Limit(cos(x)**(1/x), x, oo, dir='-')
>>> limit(tan(x)**(1/x), x, oo)
Limit(tan(x)**(1/x), x, oo, dir='-')
>>> exp(x*log(cos(1/x))).as_leading_term(x)
nan
>>> exp(x*log(sin(1/x))).as_leading_term(x)
nan
>>> exp(x*log(tan(1/x))).as_leading_term(x)
nan
>>> log(cos(1/x)).as_leading_term(x)
nan
>>> log(tan(1/x)).as_leading_term(x)
nan
```
#### Release Notes

* series
  * Fixes leading term for exponential functions involving AccumBounds based arguments.


anutosh491:
CC: @jksuom I have returned `self` as of now for these expressions but we can also return indeterminant(in our case `S.NaN`) for something like `limit(cos(x)**(1/x), x, oo)` . My thinking behind returning `self` as of now is that all over `test_limits.py` ,i've seen `self` being used/preferred over other outcomes and currently the use of `S.NaN` is pretty minimal . Hence to maintain this consistency, I thought of using self otherwise `S.NaN` is an equally good output i feel

sympy/functions/elementary/exponential.py
```diff
@@ -1055,6 +1058,8 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
             cdir = arg0.dir(x, cdir)
         if x0.is_real and x0.is_negative and im(cdir).is_negative:
             return self.func(x0) - 2*I*S.Pi
+        if isinstance(arg, AccumBounds) and arg.min < S.Zero:
+            return S.NaN
```
jksuom, line 1062: A better place for this might be `_eval`. If arg.min is 0, then `AccumBounds(-oo, arg.max)`.

anutosh491: Cool , would give this a thought !

anutosh491: I am a bit confused about the `arg` your using . Our expression is `log(cos(1/x))` . So the `arg` which `eval` uses is `self.args[0] `i.e `cos(1/x)` . But the `arg` we use in `_eval_as_leading_term` is `cos(1/x).as_leading_term(x)` which is `AccumBounds(-1, 1)`. So as per what you've written if we deal with this in eval we must be using something like `arg.as_leading_term(x).min == 0`

anutosh491: Maybe we could end up with a block like this in `eval`
```
        leading_term = arg.as_leading_term(*list(arg.free_symbols))
        if isinstance(leading_term, AccumBounds):
            from sympy import oo
            if leading_term.min < S.Zero:
                return AccumBounds(-oo, leading_term.max)
```
If we include this we end up with something as follows
```
>>> log(cos(1/x))
AccumBounds(-oo, 1)
```

jksuom: Returning min = -oo only applies when the argument AccumBounds has min == 0. Is it is < 0, then nan should be returned. Currently, there is only the min > 0 case in `eval`. The two new cases should be added.

anutosh491: Thanks for this suggestions , works better now . I also wanted your assistance on another thing which is the limit we are fixing but for the `tan/cot/sec/csc` cases . I realized that wolfram returns indefinite for 
`limit(sin(1/x)**x, x, 0)` and `limit(cos(1/x)**x, x, 0)` but for similar cases of `tan/cot/sec/csc` , it returns 1 . I was a bit confused but when after applying your suggestion it seems that's how it is supposed to be .
The diff I used 
```
diff --git a/sympy/functions/elementary/exponential.py b/sympy/functions/elementary/exponential.py
index 2030458b83..9095d9b5f5 100644
--- a/sympy/functions/elementary/exponential.py
+++ b/sympy/functions/elementary/exponential.py
@@ -703,6 +703,10 @@ def eval(cls, arg, base=None):
         elif isinstance(arg, AccumBounds):
             if arg.min.is_positive:
                 return AccumBounds(log(arg.min), log(arg.max))
+            elif arg.min.is_zero:
+                return AccumBounds(S.NegativeInfinity, log(arg.max))
+            elif arg.min.is_negative:
+                return S.NaN
             else:
                 return
         elif isinstance(arg, SetExpr):
@@ -1058,8 +1062,8 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
             cdir = arg0.dir(x, cdir)
         if x0.is_real and x0.is_negative and im(cdir).is_negative:
             return self.func(x0) - 2*I*S.Pi
-        if isinstance(arg, AccumBounds) and arg.min < S.Zero:
-            return S.NaN
+        if isinstance(arg, AccumBounds):
+            return log(arg)
         return self.func(arg)
```
Now I get 
```
>>> limit(sin(1/x)**x, x, 0) # returning self because exp(x*log(sin(1/x))).as_leading_term is nan
Limit(sin(1/x)**x, x, 0)
>>> limit(cos(1/x)**x, x, 0)
Limit(cos(1/x)**x, x, 0)
>>> limit(tan(1/x)**x, x, 0) # returning 1 because exp(x*log(tan(1/x))).as_leading_term(x) is 1
1
>>> limit(cot(1/x)**x, x, 0)
1
>>> limit(csc(1/x)**x, x, 0)
1
>>> limit(sec(1/x)**x, x, 0)
1
```
This is because 
```
>>> log(AccumBounds(-1, 1))
nan
>>> log(AccumBounds(-1, oo))
nan
>>> log(AccumBounds(-oo, oo))
log(AccumBounds(-oo, oo))
```

Wolfram pics for reference
image
image
image

anutosh491: Hence I wanted to know that now that our changes reported above correctly work for `tan/cot/sec/csc` and return `1 ` , I hope there isn't anything suspicious or doubtful here ?? I hope the results shown above are the ones we need to return .

anutosh491: If this is what we expect, I'll make the commit soon .

anutosh491:
I quite happy with the latest commit , works much better than the previous code . 
Now we have 
```
>>> limit(cos(x)**(1/x), x, oo)
Limit(cos(x)**(1/x), x, oo, dir='-')
>>> limit((cos(x) + 2)**(1/x), x, oo)
1
>>> limit((cos(x) - 2)**(1/x), x, oo)
Limit((cos(x) - 2)**(1/x), x, oo, dir='-')
```
image
image
image

anutosh491:
My only doubt/slight confusion is `limit((cos(x) + 1)**x, x, oo)` . Currently I return `1` for it !
```
>>> limit((cos(x) + 1)**(1/x), x, oo)
1
```
image

It;s not discontinuous like `limit((cos(x))**x, x, oo)` and steeply goes to `0` periodically . If `1` isn't the correct answer here(would like to have your thoughts on what you think is correct) I'll try to make the changes and then I don't think there's anything left to be addressed in this Pr as all cases have been covered . This can be merged then .

anutosh491:
CC : @jksuom does this look better to you ? I've tried to address the corner case too (limit((cos(x) + 1)**x, x, oo). It's a corner case because It;s not discontinuous like limit((cos(x))**x, x, oo) and steeply goes to 0 periodically . I have shown the graph above , but will attach a zoomed in pic here .

```
>>> log(cos(1/x) + 1).as_leading_term(x)
AccumBounds(-oo, log(2))
>>> exp(x*log(cos(1/x) + 1)).as_leading_term(x)
AccumBounds(0, 1)
>>> limit((cos(1/x) + 1)**(x), x, 0)
AccumBounds(0, 1)
>>> limit((cos(x) + 1)**(1/x), x, oo)
AccumBounds(0, 1)
```
Periodically steeply goes to 0 at (2n +1)*pi. I've shown `11, 13, 15 *pi` for reference
image

All other cases go like this 
`limit((cos(x) + n)**x, x, oo) ` is indeterminant for `n<1` and it is `1` for `n>1` . The graphs have been shown here !
https://github.com/sympy/sympy/pull/23087#issuecomment-1044043851

sympy/functions/elementary/exponential.py
```diff
@@ -529,8 +529,13 @@ def _taylor(self, x, n):
         return Add(*l)
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
+        from sympy.calculus.util import AccumBounds
         arg = self.args[0].cancel().as_leading_term(x, logx=logx)
         arg0 = arg.subs(x, 0)
+        if arg is S.NaN:
+            return S.NaN
+        if isinstance(arg0, AccumBounds) and arg0.min == S.NegativeInfinity:
```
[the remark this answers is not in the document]
anutosh491: Okay I'll try to convey my understanding of this to you .

**Overview:** Consider the case `exp(x*log(cos(1/x) + n)).as_leading_term(x)` . We always calculate leading term of `self.args[0]` first(stored in `arg`). Then we substitute `x` as `0` in arg(stored in `arg0`)

**Case1:** **exp(x*log(cos(1/x))).as_leading_term(x)**

- arg = x * log(cos(1/x)).as_leading_term(x) = nan 
- arg0 = arg.subs(x, 0) = nan
- we return nan from the first if statement (if arg == S.NaN: return S.NaN)

**Case2:** **exp(x*log(cos(1/x) + 2)).as_leading_term(x)**

- arg = x * log(cos(1/x) + 2).as_leading_term(x) = x*AccumBounds(0, log(3)) 
- arg0 = arg.subs(x, 0) = 0*AccumBounds(0, log(3)) = 0
- we return exp(0) from the fourth if statement (if arg0.is_infinite is False: return exp(arg0)

**Case2:** **exp(x*log(cos(1/x) + 1)).as_leading_term(x)** (Corner Case)

- arg = x * log(cos(1/x) + 1).as_leading_term(x) = x*AccumBounds(-oo, log(2))
This is because we return AccumBounds(-oo, log(x)) for log(AccumBounds(0, x)).

- arg0 = arg.subs(x, 0) = 0*AccumBounds(-oo, log(2)) = AccumBounds(-oo, 0)
- we return the answer from the new if statement implemented base on AccumBounds !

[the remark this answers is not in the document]
anutosh491: Hence your doubt is completely valid . We will be encountering `0* AccumBounds` for `arg0` and this will only end up not being `0` when `AccumBounds.min == -oo` . Otherwise be it AccumBounds or anything multiplied by `0` , we would end up `0` unless it would be something like `zoo, oo , nan` . So maybe we just need to check whether `arg0` is AccumBounds , cause for all other cases like the 2nd case we will end up with `arg0` being `0` .
Ps: I tried my best to explain haha ! Hope it helps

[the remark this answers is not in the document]
jksuom: > I don't think there would be any case where arg0.min is not -oo 

So it seems that it would not be necessary to test `arg0.min == -oo`, `isinstance(arg0, AccumBounds)` alone should suffice.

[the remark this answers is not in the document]
anutosh491: True agreed , had just put it for safety !but I realize that being a corner case it's not required .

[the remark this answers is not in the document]
anutosh491: Both the new tests added are also corner cases (will periodically fall to 0 steeply in pi or 2*pi period)
I'll explain one of those here 
```
>>> limit((tan(x)**2)**(2/x) , x, oo)
AccumBounds(0, oo)
```
**Step 1)** At the final step we would end up calculating `log(tan(1/x)**2).as_leading_term(x)` where we would calculate `(tan(1/x)**2).as_leading_term(x)` and as the result i.e the `arg` we get is of type `AccumBounds` , we return ` log(arg)`
```
>>> (tan(1/x)**2).as_leading_term(x)
AccumBounds(0, oo)
>>> log(tan(1/x)**2).as_leading_term(x)
AccumBounds(-oo, oo)
```
**Step 2)** At the second last step this would involve calculating the leading term of `exp(2*x*log(tan(1/x)**2))` where the `arg` returned from the above step is `2*x*AccumBounds(-oo, oo)` . Now `arg0 ` would be `0*AccumBounds(-oo, oo)` which is ` AccumBounds(-oo, oo)` and this also is solved through the corner case implementation where `arg0` is on type `AccumBounds`

**Note :The corner case would only involve 2 cases one where we have `AccumBounds(-oo, 0)` and other where we have `AccumBounds(-oo, oo)`, hence we can always return `AccumBounds(0, exp(arg0.max))` .** 
And hence in this case we get `AccumBounds(0, oo)` as our answer for limit .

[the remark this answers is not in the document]
anutosh491: And as expected we have our graph for `(tan(x)**2)**(2/x)` as follows . This rise steeply with infinite slope at every `n*pi/2`  and goes to zero steeply at every integer `pi`
image

The other new test case added (also a corner case)
```
>>> limit((sin(x)**2)**(1/x) , x, oo)
AccumBounds(0, 1)
```
image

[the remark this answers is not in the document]
anutosh491: I am quite happy with the latest commit but would like to hear from your side too . The corner case not being very intuitive I have added a set of comments which represent the changes made and the reason behind them, so that it's easy for Us to address this case in future if we make changes in `AccumBounds` etc
I am happy with the latest commit But would like to hear from your side ! @jksuom

sympy/functions/elementary/exponential.py
```diff
@@ -534,8 +534,13 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
...
             return S.NaN
-        if isinstance(arg0, AccumBounds) and arg0.min == S.NegativeInfinity:
-            return AccumBounds(exp(arg0.min), exp(arg0.max))
+        if isinstance(arg0, AccumBounds):
+            # This check addresses a corner case involving AccumBounds.
+            # if isinstance(arg, AccumBounds) is True, then arg0 can either be 0,
+            # AccumBounds(-oo, 0) or AccumBounds(-oo, oo).
+            # Check out function: test_issue_18473() in test_exponential.py and
+            # test_limits.py for more information.
+            return AccumBounds(0, exp(arg0.max))
```
jksuom, line 543: Could this be `return exp(arg0)`?

anutosh491: Yeah sure why not , slipped out of my mind , should have caught that !

anutosh491:
I have made the requested change ! Does this seem better now ?

anutosh491:
Cc: @jksuom .

jksuom:
There are 20 commits, maybe some of them could be squashed. Otherwise this looks good.

anutosh491:
> There are 20 commits, maybe some of them could be squashed. Otherwise this looks good. 

Cool will squash those which weren't necessary.  I am not sure why my git messes up whenever I am opening a new pr these days. This has been happening since my last 3 prs and the code from the previous pr automatically gets added here . Though it never shows up in the git diff , I'll have to look into this .Fixing this led to some unnecessary commits.

oscarbenjamin:
> I am not sure why my git messes up whenever I am opening a new pr these days.

Most likely you have commits in your master branch that are not from the main repo master branch. I recommend starting a new PR branch from the upstream master branch like this:
```
git fetch upstream
git checkout upstream/master
git checkout -b my_pr_branch
```

anutosh491:
> > I am not sure why my git messes up whenever I am opening a new pr these days.
> 
> Most likely you have commits in your master branch that are not from the main repo master branch. I recommend starting a new PR branch from the upstream master branch like this:
> 
> ```
> git fetch upstream
> git checkout upstream/master
> git checkout -b my_pr_branch
> ```

Thanks Oscar , will surely keep this in mind . Gets quite messy cause all 10 + commits of the previous branch were carried on in this Pr along with the changes here . Gets quite messy and then I get confused how to clear up that mess !

jksuom:
There are still commits that do not belong here, those dealing with Fourier series, for example.

anutosh491:
> There are still commits that do not belong here, those dealing with Fourier series, for example.

Yeah I am trying my best to get fix this . I haven't faced this anytime earlier , just started happening recently where commits from a seperate branch also were included in my new prs .Will try to get this sorted soon

anutosh491:
@oscarbenjamin I am seeing the above commit fail due to `test/authors` . I see this message here and locally too , I don't think I am at fault though . It this something where we would expect the mailmap.py check to fail ?
```
Ambiguous names in .mailmap
This email address appears for multiple entries:
Person: Akshānsh <akshansh@tuta.io>
Mailmap entries:
Akshansh Bhatt <akshansh@tuta.io> Akshansh <53227127+akshanshbhatt@users.noreply.github.com>
Akshansh Bhatt <akshansh@tuta.io> akshanshbhatt <53227127+akshanshbhatt@users.noreply.github.com>

Lines should be added to .mailmap to indicate the correct name and
email aliases for all commits.

https://github.com/sympy/sympy/wiki/Development-workflow#update-
mailmap
```

the next SymPy release.

	Akshānsh <akshansh@tuta.io>
	Tirthankar Mazumder <635745[8](https://github.com/sympy/sympy/runs/5259760678?check_suite_focus=true#step:6:8)8+wermos@users.noreply.github.com>
Error: Process completed with exit code 1

anutosh491:
@oscargus could you help me out here ! I'm a bit stuck and struggling to understand what's the reason behind this ( quote from a comment above)
>>> I am not sure why my git messes up whenever I am opening a new pr these days. This has been happening since my last 3 prs and the code from the previous pr automatically gets added here . Though it never shows up in the git diff , I'll have to look into this .Fixing this led to some unnecessary commits.

I tried rebase/revert etc but still nothing is really helping here ! Mostly shows error .

anutosh491:
okay even other prs fail due to the authors checks error #23125 ! So I don't think it's an error from my side

oscargus:
I think/hope that #23126 will fix the test issue (note that the authors test is not required, so it is still possible to merge, although probably less likely that someone checks out PRs that have not passed...).

oscargus:
In general, it is better to rebase and force-push than to merge.

anutosh491:
Closed in favour of #23127

PR #23088, opened 2022-02-16
Documenting matrix derivatives

Upabjojr:
It's time to announce the matrix expression derivative algorithm in the documentation, so that word will spread about this algorithm (probably currently the only CAS apart from the website [matrixcalculus.org](http://www.matrixcalculus.org/) to support such a feature).

* matrices
  * Matrix expression derivative algorithm add support for matrix calculus. The derivative of a matrix expression by another matrix expression can now be computed and the result is either an array expression (4-dimensional) or, if two of the resulting dimensions can be suppressed, a matrix expression (2-dimensional).

### DIFF
--- a/doc/src/modules/matrices/expressions.rst
+++ b/doc/src/modules/matrices/expressions.rst
@@ -22,6 +22,22 @@ The Matrix expression module allows users to write down statements like
 
 where ``X`` and ``Y`` are :class:`MatrixSymbol`'s rather than scalar symbols.
 
+Matrix expression derivatives are supported. The derivative of a matrix by another matrix
+is generally a 4-dimensional array, but if some dimensions are trivial or diagonal,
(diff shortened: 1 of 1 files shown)


Upabjojr:
@oscarbenjamin can we put this into the 1.10 branch, so that we can announce this algorithm publicly?

oscarbenjamin:
> @oscarbenjamin can we put this into the 1.10 branch, so that we can announce this algorithm publicly?

Yes, I think so. I just cherry-picked this to #23090 and moved the release note there.

It needs a better release note though i.e. the note should probably announce the feature itself rather than saying that examples were added to the documentation. Can you edit that?

oscarbenjamin:
@Upabjojr can you manually edit the release note here:
https://github.com/sympy/sympy/wiki/Release-Notes-for-1.10

PR #23091, opened 2022-02-17
Use LaTeX for labels in matplotlib backend

oscargus:

Now, x-, y-, and z-labels are rendered by default using "LaTeX" (or Mathtext) when plotting using `plot` and `plot3d` with the Matplotlib backend. In this way, code like:
```
from sympy import *
x = Symbol('x_Delta')
plot(sin(x), (x, -1, 1))
```
renders as
image

#### Other comments
As one can set the labels separately, the logic is that if there is a `$` in the label, it is assumed to already be in LaTeX-format. There are some odd cases where that is not the case, but still quite unlikely that one creates a variable with a $ in the name and expect LaTeX-rendering.

I realize that setting the labels using `latex` actually only should be applied if a backend supporting this is used. Is there are way figure out which backend will be used at this stage?

#### Release Notes

* plotting
   * Labels etc can be set to expressions, including the default values, and are latexified using the Matplotlib backend.

### DIFF
--- a/sympy/plotting/plot.py
+++ b/sympy/plotting/plot.py
@@ -25,15 +25,18 @@ every time you call ``show()`` and the old one is left to the garbage collector.
 
 from collections.abc import Callable
 
+
+from sympy.core.basic import Basic
 from sympy.core.containers import Tuple
 from sympy.core.expr import Expr
+from sympy.core.function import arity, Function
 from sympy.core.symbol import (Dummy, Symbol)
 from sympy.core.sympify import sympify
 from sympy.external import import_module
-from sympy.core.function import arity
+from sympy.printing.latex import latex
+from sympy.utilities.exceptions import sympy_deprecation_warning
 from sympy.utilities.iterables import is_sequence
 from .experimental_lambdify import (vectorized_lambdify, lambdify)
-from sympy.utilities.exceptions import sympy_deprecation_warning
 
 # N.B.
 # When changing the minimum module version for matplotlib, please change
@@ -53,6 +56,11 @@ def unset_show():
     """
     global _show
     _show = False
+
+def _str_or_latex(label):
+    if isinstance(label, Basic):
+        return latex(label, mode='inline')
+    return str(label)
 
 ##############################################################################
 # The public interface
@@ -110,9 +118,9 @@ class Plot:
     The global options for a figure are:
 
     - title : str
-    - xlabel : str
-    - ylabel : str
-    - zlabel : str
+    - xlabel : str or Symbol
+    - ylabel : str or Symbol
+    - zlabel : str or Symbol
     - legend : bool
     - xscale : {'linear', 'log'}
     - yscale : {'linear', 'log'}
@@ -664,7 +672,7 @@ class LineOver1DRangeSeries(Line2DBaseSeries):
     def __init__(self, expr, var_start_end, **kwargs):
         super().__init__()
         self.expr = sympify(expr)
-        self.label = kwargs.get('label', None) or str(self.expr)
+        self.label = kwargs.get('label', None) or self.expr
         self.var = sympify(var_start_end[0])
         self.start = float(var_start_end[1])
         self.end = float(var_start_end[2])
@@ -792,6 +800,7 @@ class LineOver1DRangeSeries(Line2DBaseSeries):
         list_y = f(list_x)
         return (list_x, list_y)
 
+
 class Parametric2DLineSeries(Line2DBaseSeries):
     """Representation for a line consisting of two parametric SymPy expressions
     over a range."""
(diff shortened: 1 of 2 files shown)


oscarbenjamin:
> I realize that setting the labels using `latex` actually only should be applied if a backend supporting this is used. Is there are way figure out which backend will be used at this stage?

Not sure. @Davide-sd do you know how this should be handled?

Davide-sd:
When managing labels, my module is currently coded as SymPy: it uses the string representation of a symbol. If a user wants fancy rendering, `xlabel` and `ylabel` must be manually set. That's because:

1. some backend doesn't support Latex (for example Bokeh).
2. other backends might support Latex, but not on all browsers (for example Plotly).

sympy/plotting/plot.py
```diff
@@ -1843,8 +1853,8 @@ def plot(*args, show=True, **kwargs):
                     'The same variable should be used in all '
                     'univariate expressions being plotted.')
     x = free.pop() if free else Symbol('x')
-    kwargs.setdefault('xlabel', x.name)
-    kwargs.setdefault('ylabel', 'f(%s)' % x.name)
+    kwargs.setdefault('xlabel', latex(x))
+    kwargs.setdefault('ylabel', 'f(%s)' % latex(x))
```
asmeurer, line 1857: Should we set mode='inline'? I think there are some places in the printer that print different things for inline vs. equation

asmeurer:
This is causing a failure in the Sphinx build.

sympy/plotting/plot.py
oscargus: That is probably a good idea, but maybe it is not a good idea to try to do this, at least not in this trivial way.

oscargus:
Thanks @Davide-sd for the clarification. I was worried it was like that. It is a bit tempting to be able to pass a `Symbol` as a label and then in the backend decide what to do with it. (If a string, one simply uses that.)

I'll mark this as draft and think a bit (and fix the issue in physics).

oscargus:
After giving this some thinking. Would it make sense to be able to set the labels to expressions/symbols and then leave it to the backends to detect if it is an expression or string and deal with it accordingly? One can even provide (private) convenience functions (although those will probably be simple wrappers to `str` and `latex`), possibly as a mixin class.

@Davide-sd regarding
> 2. other backends might support Latex, but not on all browsers (for example Plotly).

Does the backend know if the browser supports Latex or not?

Then one should decide if `plot` etc should set ylabel to f(symbol) or the actual expression. It is of course nice with the actual expression, but only if it is "small enough"... (The Bode plots in `physics.control` as the transfer function which is not always a good idea, but nice when it is.)

Here I am primarily focusing on "simple graphs", getting something sensible quickly, rather than getting "publication quality" graphs where one probably would like to set things explicitly.

Davide-sd:
> After giving this some thinking. Would it make sense to be able to set the labels to expressions/symbols and then leave it to the backends to detect if it is an expression or string and deal with it accordingly?

I was thinking of doing exactly this on my module.

> Does the backend know if the browser supports Latex or not?

@oscargus unfortunately not. But my module comes with default user-configurable options. So, I can create an option `use_latex=True`. Then, if a user detects that a particular backend/browser combination doesn't work well (for example, Plotly/Firefox), the user set `use_latex=False` only once and from that moment all plots created with a particular backend will fall back to string representation.

Davide-sd:
By the way, does Matplotlib 3D support latex labels?

oscargus:
I have updated this to allow both strings and symbols/expressions, including legend labels.

Not sure if it may break other backends though... (Some plots in physics seems to already set labels to symbols/expressions.)

oscargus:
(Btw, the documentation could use some time to standardize argument formatting etc. But if we are aiming for a new plot module, it is probably not highly prioritized.)

Davide-sd:
> Btw, the documentation could use some time to standardize argument formatting etc. But if we are aiming for a new plot module, it is probably not highly prioritized.

New plotting module has that covered!

sympy/plotting/plot.py
oscargus: Fixed in updated version,

oscarbenjamin:
Given the discussion above I'm not sure where that leaves this. Does the "new plotting module" have any bearing on whether this PR should be merged?

oscargus:
This introduces support for using expressions as labels (which some code already did, relying on str-conversion). The new plotting module does support it as well, so for consistency it should make sense to merge this. (Then I do not know if the old plotting will be removed and when the new may be merged, so it may be that this won't make it into any release, but that would be the only "drawback" as I see it. And that people get LaTeX-labels instead of strings by default, but hardly a drawback.)

oscarbenjamin:
@Davide-sd does this PR look good to you?

Davide-sd:
Look's good!

oscarbenjamin:
Okay, let's merge this then.

PR #23093, opened 2022-02-17
Fix partial derivative index order

### LINKED ISSUE #23073: tensor PartialDerivative behaves in unexpected (wrong?) way
I'm trying to do some symbolic GR work in sympy, but I get stuck on behaviour of tensor PartialDerivative that I don't understand and can't seem to deal with.

Bringing it down to simplest case, I have a tensor (g) with 1 upper index (a) in 2 dimensions and I take the partial derivative with respect to both dimensions (index -b). However, when I look at the result after `replace_with_arrays`, it seems that it has swapped the a and -b indices. In more complicated cases (taking the partial derivative of  rank 2 tensor, it seems the indices are reversed.

Am I missing something simple (please tell me what it is, then), or is this a bug? I can't seem to find a way to compensate for this behaviour. I'm trying to implement an (well-known) expression like `PartialDerivative(g(-a,-b), x(m)) + PartialDerivative(g(-a,-m), x(b)) - PartialDerivative(g(-b,-m),x(a))` where the changing order of indices a, b, m is crucial, but the eventual added result is wrong, due (I think) to the fact that PartialDerivative does not give properly ordered output after `replace_with_arrays`.

Here is the basic code for the artificial rank 1 example:
```
import sympy
from sympy.tensor.tensor import TensorIndexType, tensor_indices, tensor_heads
from sympy.tensor.toperators import PartialDerivative
Lorentz = TensorIndexType('Lorentz', dummy_name='L', dim=2)
a, b = tensor_indices('a,b', Lorentz)
r, theta = sympy.symbols("r, theta")
x, g = tensor_heads('x g', [Lorentz])
repl = {x(a): [r, theta], g(a): [0, r]}
def Show(T):
    print(T.get_indices())
    print(T.replace_with_arrays(repl))
pd = PartialDerivative(g(a),x(b))
# we expect to see one 1 at a==1, -b==0, because the partial derivative vanishes except for dg(1)/dx(0)
# but somehow the indices are reversed, so the 1 is at a==0, -b==1
Show( pd ) # --> gives [a, -b], [[0, 1], [0, 0]] instead of [a, -b], [[0, 0], [1, 0]]
```

Upabjojr:
Fixes https://github.com/sympy/sympy/issues/23073

NO ENTRY

### DIFF
--- a/sympy/tensor/tests/test_tensor_operators.py
+++ b/sympy/tensor/tests/test_tensor_operators.py
@@ -29,6 +29,13 @@ def test_invalid_partial_derivative_valence():
 
 def test_tensor_partial_deriv():
     # Test flatten:
+    expr = PartialDerivative(PartialDerivative(A(i), A(j)), A(k))
+    assert expr == PartialDerivative(A(i), A(j), A(k))
+    assert expr.get_indices() == [-k, -j, i]
+    assert expr.get_free_indices() == [-k, -j, i]
+    assert expr.expr == A(i)
+    assert expr.variables == (A(j), A(k))
+
     expr = PartialDerivative(PartialDerivative(A(i), A(j)), A(i))
     assert expr.expr == A(L_0)
     assert expr.variables == (A(j), A(L_0))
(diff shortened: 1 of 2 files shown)


ThePauliPrinciple:
This change is not backwards compatable and might break people's current code. Can't the `Array` that is returned by `.replace_with_arrays` just have its indices swapped so that everything is consistent?

I like the update in the docs.

Upabjojr:
> Can't the `Array` that is returned by `.replace_with_arrays` just have its indices swapped so that everything is consistent?

In principle yes. Though I have used the convention of putting the deriving dimensions before the base dimensions everywhere else in SymPy.

Maybe we should leave an option to keep the old behaviour with a deprecation.

I was wondering... Can we find literature about this convention?

Upabjojr:
Not going to merge this as the issue has been solved in https://github.com/sympy/sympy/pull/23100

PR #23095, opened 2022-02-18
Removed ^ and & from the mechanics documentation.

moorepants:

The `__xor__` and `__and__` methods are overridden for physics vectors to be
the cross product and dot product, but I want to leave these as
undocumented features and even possibly remove in the distant future. I
don't think it is good practice to override builtin methods with
functionality that does not resemble the intended functionality, i.e.
don't make `__add__` do multiplication. Using the .cross/.dot and
cross()/dot() methods and functions is clearer to read and more
pythonic. We also don't need 3 ways to do one thing (unpythonic). This
PR just stops advertising the overridden methods.

#### Release Notes

NO ENTRY

/cc @sidhu1012


oscarbenjamin:
Looks good.

PR #23099, opened 2022-02-18
polys - fixed bug in primitive_element which now does not raise zero divisor error when zero is present in input

praneethratna:
Helps in PR #23077 and issue #22943 and now `primitive_element` does not raise zero divisor error when zero is present in the input.
In current master,
```
>>> primitive_element([sqrt(2), 0], x, ex=True)
sympy.polys.polyerrors.NotInvertible: zero divisor
```
After this change,
```
>>> primitive_element([sqrt(2), 0], x, ex=True)
(x**2 - 2, [1, 0], [[MPQ(1,1), MPQ(0,1)], []])
```

#### Release Notes

* polys
   * fixed bug in `primitive_element` which now does not raise zero divisor error when zero is present in the input.

### DIFF
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -387,7 +387,12 @@ def primitive_element(extension, x=None, *, ex=False, polys=False):
         ogen = K.unit - s*erep  # old gen as element of K
         reps = [dup_eval(_.rep, ogen, K) for _ in reps] + [erep]
 
-    H = [_.rep for _ in reps]
+    if K.ext.root.is_Rational:  # all extensions are rational
+        H = [K.convert(_).rep for _ in extension]
+        coeffs = [0]*len(extension)
+        f = cls(x, domain=QQ)
+    else:
+        H = [_.rep for _ in reps]
     if not polys:
         return f.as_expr(), coeffs, H
     else:
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -36,7 +36,7 @@ from sympy.core.add import Add
 from sympy.core.numbers import AlgebraicNumber
 from sympy.core.singleton import S
 from sympy.core.symbol import Dummy
-from sympy.core.sympify import sympify
+from sympy.core.sympify import sympify, _sympify
 from sympy.ntheory import sieve
 from sympy.polys.densetools import dup_eval
 from sympy.polys.domains import QQ
@@ -340,7 +340,7 @@ def primitive_element(extension, x=None, *, ex=False, polys=False):
     """
     if not extension:
         raise ValueError("Cannot compute primitive element for empty extension")
-    extension = [sympify(ext) for ext in extension]
+    extension = [_sympify(ext) for ext in extension]
 
     if x is not None:
         x, cls = sympify(x), Poly
@@ -351,6 +351,9 @@ def primitive_element(extension, x=None, *, ex=False, polys=False):
         gen, coeffs = extension[0], [1]
         g = minimal_polynomial(gen, x, polys=True)
         for ext in extension[1:]:
+            if ext.is_Rational:
+                coeffs.append(0)
+                continue
             _, factors = factor_list(g, extension=ext)
             g = _choose_factor(factors, x, gen)
             s, _, g = g.sqf_norm()
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -366,6 +366,11 @@ def primitive_element(extension, x=None, *, ex=False, polys=False):
     K = QQ.algebraic_field((f, gen))  # incrementally constructed field
     reps = [K.unit]  # representations of extension elements in K
     for ext in extension[1:]:
+        ext = sympify(ext)
+        if ext.is_Rational:
+            coeffs.append(0)    # rational ext is not included in the expression of a primitive element
+            reps.append(K.convert(ext))    # but it is included in reps
+            continue
         p = minimal_polynomial(ext, x, polys=True)
         L = QQ.algebraic_field((p, ext))
         _, factors = factor_list(f, domain=L)
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -340,6 +340,7 @@ def primitive_element(extension, x=None, *, ex=False, polys=False):
     """
     if not extension:
         raise ValueError("Cannot compute primitive element for empty extension")
+    extension = [sympify(ext) for ext in extension]
 
     if x is not None:
         x, cls = sympify(x), Poly
@@ -366,7 +367,6 @@ def primitive_element(extension, x=None, *, ex=False, polys=False):
     K = QQ.algebraic_field((f, gen))  # incrementally constructed field
     reps = [K.unit]  # representations of extension elements in K
     for ext in extension[1:]:
-        ext = sympify(ext)
         if ext.is_Rational:
             coeffs.append(0)    # rational ext is not included in the expression of a primitive element
             reps.append(K.convert(ext))    # but it is included in reps
--- a/sympy/polys/numberfields/tests/test_subfield.py
+++ b/sympy/polys/numberfields/tests/test_subfield.py
@@ -273,7 +273,7 @@ def test_primitive_element():
     assert primitive_element([a, b, I], x) == (x**4 + 6*x**2 + 1, [1, 0, 0])
 
     assert primitive_element([sqrt(2), 0], x, ex=True) == (x**2 - 2, [1, 0], [[MPQ(1,1), MPQ(0,1)], []])
-    assert primitive_element([0], x) == (x, [1])
+
 
 def test_to_number_field():
     assert to_number_field(sqrt(2)) == AlgebraicNumber(sqrt(2))
--- a/sympy/polys/numberfields/tests/test_subfield.py
+++ b/sympy/polys/numberfields/tests/test_subfield.py
@@ -4,7 +4,7 @@ from sympy.core.numbers import (AlgebraicNumber, I, pi, Rational)
 from sympy.core.singleton import S
 from sympy.functions.elementary.exponential import exp
 from sympy.functions.elementary.miscellaneous import sqrt
-from sympy.external.pythonmpq import PythonMPQ as MPQ
+from sympy.external.gmpy import MPQ
 from sympy.polys.numberfields.subfield import (
     is_isomorphism_possible,
     field_isomorphism_pslq,
--- a/sympy/polys/numberfields/tests/test_subfield.py
+++ b/sympy/polys/numberfields/tests/test_subfield.py
@@ -4,6 +4,7 @@ from sympy.core.numbers import (AlgebraicNumber, I, pi, Rational)
 from sympy.core.singleton import S
 from sympy.functions.elementary.exponential import exp
 from sympy.functions.elementary.miscellaneous import sqrt
+from sympy.external.pythonmpq import PythonMPQ as MPQ
 from sympy.polys.numberfields.subfield import (
     is_isomorphism_possible,
     field_isomorphism_pslq,
@@ -271,6 +272,8 @@ def test_primitive_element():
     a, b = I*sqrt(2*sqrt(2) + 3), I*sqrt(-2*sqrt(2) + 3)
     assert primitive_element([a, b, I], x) == (x**4 + 6*x**2 + 1, [1, 0, 0])
 
+    assert primitive_element([sqrt(2), 0], x, ex=True) == (x**2 - 2, [1, 0], [[MPQ(1,1), MPQ(0,1)], []])
+    assert primitive_element([0], x) == (x, [1])
 
 def test_to_number_field():
     assert to_number_field(sqrt(2)) == AlgebraicNumber(sqrt(2))
--- a/sympy/polys/numberfields/tests/test_subfield.py
+++ b/sympy/polys/numberfields/tests/test_subfield.py
@@ -272,7 +272,10 @@ def test_primitive_element():
     a, b = I*sqrt(2*sqrt(2) + 3), I*sqrt(-2*sqrt(2) + 3)
     assert primitive_element([a, b, I], x) == (x**4 + 6*x**2 + 1, [1, 0, 0])
 
+    assert primitive_element([sqrt(2), 0], x) == (x**2 - 2, [1, 0])
+    assert primitive_element([0, sqrt(2)], x) == (x**2 - 2, [1, 1])
     assert primitive_element([sqrt(2), 0], x, ex=True) == (x**2 - 2, [1, 0], [[MPQ(1,1), MPQ(0,1)], []])
+    assert primitive_element([0, sqrt(2)], x, ex=True) == (x**2 - 2, [1, 1], [[], [MPQ(1,1), MPQ(0,1)]])
 
 
 def test_to_number_field():
--- a/sympy/polys/densetools.py
+++ b/sympy/polys/densetools.py
@@ -273,9 +273,6 @@ def dup_eval(f, a, K):
     11
 
     """
-    if not a:
-        return dup_TC(f, K)
-
     result = K.zero
 
     for c in f:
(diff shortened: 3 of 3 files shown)


sympy/polys/numberfields/tests/test_subfield.py
```diff
@@ -271,6 +272,8 @@ def test_primitive_element():
     a, b = I*sqrt(2*sqrt(2) + 3), I*sqrt(-2*sqrt(2) + 3)
     assert primitive_element([a, b, I], x) == (x**4 + 6*x**2 + 1, [1, 0, 0])
 
+    assert primitive_element([sqrt(2), 0], x, ex=True) == (x**2 - 2, [1, 0], [[MPQ(1,1), MPQ(0,1)], []])
```
oscarbenjamin, line 275: Without `ex=True` this still gives an error:
```python
In [3]: primitive_element([sqrt(2), 0])
OptionError
```

praneethratna: Should'nt that give an error? or should it's output be same as that of with `ex=True`.

skieffer: Might want to also have a test case where `ex=True` but the _first_ extension is rational, such as

```python
primitive_element([0, sqrt(2)], x, ex=True)
```

since I'm wondering if in such a case these lines

https://github.com/sympy/sympy/blob/403067f7bd22d63693a152c659d8681a3125fb9b/sympy/polys/numberfields/subfield.py#L364-L367

might cause trouble. Maybe here if `gen.isRational` then `K = QQ` and adjust `coeffs` and `reps` accordingly (and don't compute `f`).

praneethratna: > ```python
> primitive_element([0, sqrt(2)], x, ex=True)
> ```

Yeah this test is failing with this error,
```
  File "/home/praneeth/sympy/sympy/polys/numberfields/subfield.py", line 387, in primitive_element
    H = [_.rep for _ in reps]
  File "/home/praneeth/sympy/sympy/polys/numberfields/subfield.py", line 387, in <listcomp>
    H = [_.rep for _ in reps]
AttributeError: 'PythonMPQ' object has no attribute 'rep'
```

praneethratna: > @praneethratna I don't think it should be an error. It's just that when `ex=False` then it's a completely separate block of code,

Oh, then seems like this is not the correct fix then. I'm not sure of how this can be fixed.

oscarbenjamin:
> Do you mean something like this?

No what I mean is

1. Uninstall gmpy2 (e.g. `pip uninstall gmpy2`)
2. Run the tests
3. Install gmpy2 (e.g. `pip install gmpy2`)
4. Run the tests again

You need to write your code so that the tests pass in both cases.

sympy/polys/numberfields/tests/test_subfield.py
skieffer: Maybe the whole function needs a rewrite.

The following is passing all tests under `sympy/polys`. I have not tested it outside of there, though.
(Note: need to import `_sympify` at top of file.)

```python
    if not extension:
        raise ValueError("Cannot compute primitive element for empty extension")
    extension = [_sympify(ext) for ext in extension]

    if x is not None:
        x, cls = sympify(x), Poly
    else:
        x, cls = Dummy('x'), PurePoly

    coeffs = []
    reps = []
    # Seek first irrational generator
    for i, gen in enumerate(extension):
        if gen.is_Rational:
            coeffs.append(0)
            reps.append([gen] if gen else [])
        else:
            coeffs.append(1)
            f = minimal_polynomial(gen, x, polys=True)
            if ex:
                K = QQ.algebraic_field((f, gen))
                reps.append(K.unit.rep)
            break
    else:
        # There were no irrational generators
        f = Poly(x)

    if not ex:
        for ext in extension[i+1:]:
            if ext.is_Rational:
                coeffs.append(0)
                continue
            _, factors = factor_list(f, extension=ext)
            f = _choose_factor(factors, x, gen)
            s, _, f = f.sqf_norm()
            gen += s*ext
            coeffs.append(s)

        if not polys:
            return f.as_expr(), coeffs
        else:
            return cls(f), coeffs

    for ext in extension[i+1:]:
        if ext.is_Rational:
            coeffs.append(0)    # rational ext is not included in the expression of a primitive element
            reps.append(K.convert(ext).rep)    # but it is included in reps
            continue
        p = minimal_polynomial(ext, x, polys=True)
        L = QQ.algebraic_field((p, ext))
        _, factors = factor_list(f, domain=L)
        f = _choose_factor(factors, x, gen)
        s, g, f = f.sqf_norm()
        gen += s*ext
        coeffs.append(s)
        K = QQ.algebraic_field((f, gen))
        h = _switch_domain(g, K)
        erep = _linsolve(h.gcd(p))  # ext as element of K
        ogen = K.unit - s*erep  # old gen as element of K
        reps = [dup_eval(_, ogen, K).rep for _ in reps] + [erep.rep]

    if not polys:
        return f.as_expr(), coeffs, reps
    else:
        return f, coeffs, reps
```

sympy/polys/numberfields/subfield.py
```diff
@@ -340,32 +340,52 @@ def primitive_element(extension, x=None, *, ex=False, polys=False):
...
         raise ValueError("Cannot compute primitive element for empty extension")
+    extension = [_sympify(ext) for ext in extension]
 
     if x is not None:
         x, cls = sympify(x), Poly
     else:
         x, cls = Dummy('x'), PurePoly
 
+    coeffs = []
+    reps = []
```
jksuom, line 351: reps is needed only in the ex=True case.

sympy/polys/numberfields/tests/test_subfield.py
jksuom: I would keep the two ex cases separate and only add these lines to the False case:

    If ext.is_Rational:
        coeffs.append(0)
        continue

sympy/polys/numberfields/subfield.py
praneethratna: Yeah, reps is being returned only during `ex=True` case i think.

sympy/polys/numberfields/tests/test_subfield.py
praneethratna: > I would keep the two ex cases separate and only add these lines to the False case:

There are few failing test cases, like `primitive_element([sqrt(2), sqrt(3)], x)` gives `(x**2 - 2, [1])` which is incorrect i guess.

jksuom: This is what the current implementation gives:
```
In [11]: primitive_element([sqrt(2), sqrt(3)], x)                               
Out[11]: 
⎛ 4       2            ⎞
⎝x  - 10⋅x  + 1, [1, 1]⎠
```
Do you mean that adding the three lines in https://github.com/sympy/sympy/pull/23099#discussion_r810463900 would change the result?

praneethratna: No, what is meant is that this implementation for `ex=False` case, the code is
```python
if not ex:
        for ext in extension[i+1:]:
            if ext.is_Rational:
                coeffs.append(0)
                continue
            _, factors = factor_list(f, extension=ext)
            f = _choose_factor(factors, x, gen)
            s, _, f = f.sqf_norm()
            gen += s*ext
            coeffs.append(s)
```
But If I'm correct what you have suggested is
```python
if not ex:
        for ext in extension[i+1:]:
            if ext.is_Rational:
                coeffs.append(0)
                continue
```
then the output is changing.

jksuom: The current implementation looks like this:
```
    if not ex:
        gen, coeffs = extension[0], [1]
        g = minimal_polynomial(gen, x, polys=True)
        for ext in extension[1:]:
            _, factors = factor_list(g, extension=ext)
            g = _choose_factor(factors, x, gen)
            s, _, g = g.sqf_norm()
            gen += s*ext
            coeffs.append(s)
```
After this change:
```
    if not ex:
        gen, coeffs = extension[0], [1]
        g = minimal_polynomial(gen, x, polys=True)
        for ext in extension[1:]:
            if ext.is_Rational:
                coeffs.append(0)
                continue
            _, factors = factor_list(g, extension=ext)
            g = _choose_factor(factors, x, gen)
            s, _, g = g.sqf_norm()
            gen += s*ext
            coeffs.append(s)
```
I get
```
In [1]: primitive_element([sqrt(2), sqrt(3)], x)                                
Out[1]: 
⎛ 4       2            ⎞
⎝x  - 10⋅x  + 1, [1, 1]⎠
```

praneethratna: Sorry, i thought the implementation that you meant was https://github.com/sympy/sympy/pull/23099#discussion_r810397651, Yeah the change that you have mentioned is working except for one case `primitive_element([0, sqrt(2)], x, ex=True)` which is still giving an error.

jksuom: It seems that `dup_eval` returns an mpq object. I don't think that should happen.

praneethratna: I'm not sure but this change is fixing the error, But some test cases are failing.
```diff
diff --git a/sympy/polys/numberfields/subfield.py b/sympy/polys/numberfields/subfield.py
index d9e3aec008..343b795347 100644
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -382,13 +382,12 @@ def primitive_element(extension, x=None, *, ex=False, polys=False):
         h = _switch_domain(g, K)
         erep = _linsolve(h.gcd(p))  # ext as element of K
         ogen = K.unit - s*erep  # old gen as element of K
-        reps = [dup_eval(_.rep, ogen, K) for _ in reps] + [erep]
+        reps = [dup_eval(_, ogen, K).rep for _ in reps] + [erep.rep]

-    H = [_.rep for _ in reps]
     if not polys:
-        return f.as_expr(), coeffs, H
+        return f.as_expr(), coeffs, reps
     else:
-        return f, coeffs, H
+        return f, coeffs, reps

```

jksuom: This will make `dup_eval` return an ANP object with `rep`.
```
--- a/sympy/polys/densetools.py
+++ b/sympy/polys/densetools.py
@@ -273,9 +273,6 @@ def dup_eval(f, a, K):
     11
 
     """
-    if not a:
-        return dup_TC(f, K)
-
     result = K.zero
 
     for c in f:
```
It will give
```
In [1]: primitive_element([0, sqrt(2)], x, ex=True)                             
Out[1]: (x**2 - 2, [1, 1], [[], [mpq(1,1), mpq(0,1)]])
```

praneethratna: Yeah, every test case is passing now, I'll push these changes.

jksuom: This seems to work as well:
```
--- a/sympy/polys/densetools.py
+++ b/sympy/polys/densetools.py
@@ -274,7 +274,7 @@ def dup_eval(f, a, K):
 
     """
     if not a:
-        return dup_TC(f, K)
+        return K.convert(dup_TC(f, K))
 
     result = K.zero
```
That might be slightly more efficient though the case `a == 0` is probably not common.

praneethratna: So should i commit this change or the previous change is fine?

jksuom: This change goes straight to the essential point, so maybe it would be preferable.

jksuom:
This looks good to me.

oscarbenjamin:
I'm not sure about these:
```python
In [10]: primitive_element([0, 1], ex=True)
Out[10]: (_x, [1, 0], [[mpq(1,1), mpq(0,1)], [mpq(1,1)]])

In [11]: primitive_element([1, 0], ex=True)
Out[11]: (_x - 1, [1, 0], [[mpq(1,1), mpq(0,1)], []])
```
What do we want the output to be if all exts are actually rational? I think it would be fine to raise a dedicated error in this case.

At least though I would expect that all the representing polynomials have degree 0 (i.e. length 1 or 0). The obvious invariant that should be maintained is that the primitive element alpha is a root of the polynomial and the coefficients for ex=True represent the input elements as polynomials in alpha, but this does not uniquely describe the output when the primitive element itself is rational. In `Out[10]` we have that `alpha=0` and then `1` is represented as `[1,0]` meaning `1*alpha+0` but it could just be `[0]` or `[]` which both indicate zero more directly.

praneethratna:
> I think it would be fine to raise a dedicated error in this case.

Maybe we can raise something like `ValueError("Cannot compute unique output when all the inputs coefficients are rational")` ?

jksuom:
A primitive element cannot be expected to be unique. It is the representation of a rational number as polynomial of another rational number that is not unique. I am not sure that would be problem though.

oscarbenjamin:
Although the choice of a primitive element for an extension field is not unique the representation of elements of the extension field as polynomials in the primitive element is unique when the primitive element is a proper irrational algebraic generator for the field extension. In the field `Q(sqrt(2))` every element of the field can be uniquely represented as `a*sqrt(2)+b` where `a` and `b` are uniquely specified elements of `Q`. In this case we don't have an irrational generator though:
```python
In [15]: primitive_element([1, 0], ex=True)
Out[15]: (_x - 1, [1, 0], [[mpq(1,1), mpq(0,1)], []])
```
We're now constructing the field `Q(1)` (really just `Q` again) and the representation of any element of the field as `a*1+b` is not unique e.g. above we have decided to represent `1` as `[1,0]` meaning `1*1+0` but we could have instead chosen to represent it as simply `[1]` which is short for `[0,1]` meaning `0*1+1`.

In the absence of a unique specification for the coefficients I think it would be better to say that any element of `q` of `Q` should be represented as `[q]` except zero which would be `[]`. Then the representation of rationals like `q` in `Q(1)` is the same as their representation in `Q(sqrt(2))`.

jksuom:
I agree that the output should be canonical also when all "extensions" are rational. `reps` can be constructed directly from the extension list and all `coeffs` could be zero. (That seems to be the most natural choice.) Maybe something like this (after the loop):

    if K.ext.root.is_Rational:  # all extensions are rational
        H = [K.convert(_).rep for _ in extension]
        coeffs = [0]*len(extension)
    else:
        H = [_.rep for _ in reps]

praneethratna:
> I agree that the output should be canonical also when all "extensions" are rational. `reps` can be constructed directly from the extension list and all `coeffs` could be zero. (That seems to be the most natural choice.)

This is the output after implementing this change, but is it correct? can we also raise a warning that there won't be a unique representation of elements of the extension field as polynomials in the primitive element when all the input coefficients are rational?
```python
>>> from sympy import *
>>> from sympy.abc import *
>>> primitive_element([1, 0], ex=True)
(_x - 1, [0, 0], [[mpq(1,1)], []])
```

jksuom:
The canonical minimal polynomial should be simply `x` (I did forget to change it). Actually, any polynomial will do as the reps are constant.

praneethratna:
> The canonical minimal polynomial should be simply `x` (I did forget to change it). Actually, any polynomial will do as the reps are constant.

So a change like this would work?
```diff
diff --git a/sympy/polys/numberfields/subfield.py b/sympy/polys/numberfields/subfield.py
index c314084253..3aa12d1401 100644
--- a/sympy/polys/numberfields/subfield.py
+++ b/sympy/polys/numberfields/subfield.py
@@ -387,7 +387,12 @@ def primitive_element(extension, x=None, *, ex=False, polys=False):
         ogen = K.unit - s*erep  # old gen as element of K
         reps = [dup_eval(_.rep, ogen, K) for _ in reps] + [erep]

-    H = [_.rep for _ in reps]
+    if K.ext.root.is_Rational:  # all extensions are rational
+        H = [K.convert(_).rep for _ in extension]
+        coeffs = [0]*len(extension)
+        f = x
+    else:
+        H = [_.rep for _ in reps]
     if not polys:
         return f.as_expr(), coeffs, H
     else:
```

jksuom:
`f` is expected to be a polynomial, so probably `cls(x, domain=QQ)`.

jksuom:
I think that this ready to go, thanks!

PR #23103, opened 2022-02-18
Added simplification of matrix multiplication with inverse matrices containing MatMul objects

Upabjojr:
Just trying to simplify expressions like `Inverse(X*X.T)*X*X.T`, which can be useful to deal with equations of matrix expressions.

NO ENTRY

### DIFF
--- a/sympy/matrices/expressions/matmul.py
+++ b/sympy/matrices/expressions/matmul.py
@@ -320,8 +320,27 @@ def combine_powers(mul):
     factor, args = mul.as_coeff_matrices()
     new_args = [args[0]]
 
-    for B in args[1:]:
+    for i in range(1, len(args)):
         A = new_args[-1]
+        B = args[i]
+
+        if isinstance(B, Inverse) and isinstance(B.arg, MatMul):
+            Bargs = B.arg.args
+            l = len(Bargs)
+            if list(Bargs) == new_args[-l:]:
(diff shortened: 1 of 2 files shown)


oscargus:
Isn't there a risk that we run into the x/x discussion here again? What if the matrices are not invertible?

Upabjojr:
> Isn't there a risk that we run into the x/x discussion here again? What if the matrices are not invertible?

The code is already simplifying expressions like `Inverse(X)*X` or even `Inverse(X*Y)*X*Y` if `X` and `Y` are both square matrices. It's already happening without checking for singularity.

This PR just adds the possibility of extending it to inverses of matrix multiplications: `Inverse(X*Y)*X*Y` when `X` and `Y` are not square matrices but their product is.

If you worry about the x/x problem, it's already there.

oscargus:
OK! I know that x/x is already there, but also that it is debated. 

Anyway, this doesn't introduce any new "issue", so fair enough.

In the long term one should possibly add some assumption for matrices and matrixsymbols to indicate if they are non-singular (maybe there already is, if so, it should probably be used).

Upabjojr:
> In the long term one should possibly add some assumption for matrices and matrixsymbols to indicate if they are non-singular (maybe there already is, if so, it should probably be used).

This is a bit complicated for this PR as it deals with non-square matrices whose product is square and invertible.

PR #23104, opened 2022-02-18
Deprecate redundant static methods in carmichael

### LINKED ISSUE #23057: Deprecate redundant methods in `carmichael`
The following static methods are redundant and should be deprecated:

https://github.com/sympy/sympy/blob/dc4f807f4116c449b82433fe4ae583dfab06bf5c/sympy/functions/combinatorial/numbers.py#L109-L119

Just look at where `is_square` and `isprime` is imported from and point to that as `useinstead`. For `divides` it is enough to type the corresponding Python code as `useinstead`.

**Note that there is a change in the deprecation procedure**, see #22900 , so it makes sense to wait with introducing these deprecations until that is merged and done (and to follow the new procedure).

oscargus:

Closes #23057

Deprecate static wrapper methods.

#### Other comments

Formatted the documentation a bit better.

#### Release Notes

* functions
   * DEPRECATION: `is_prime`, `is_perfect_square`, and `divides` methods of `carmichael` deprecated.

### DIFF
--- a/sympy/functions/combinatorial/numbers.py
+++ b/sympy/functions/combinatorial/numbers.py
@@ -24,6 +24,7 @@ from sympy.functions.combinatorial.factorials import (binomial,
     factorial, subfactorial)
 from sympy.ntheory.primetest import isprime, is_square
 from sympy.utilities.enumerative import MultisetPartitionTraverser
+from sympy.utilities.exceptions import sympy_deprecation_warning
 from sympy.utilities.iterables import multiset, multiset_derangements, iterable
 from sympy.utilities.memoization import recurrence_memo
 from sympy.utilities.misc import as_int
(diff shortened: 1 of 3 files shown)


sympy/functions/combinatorial/numbers.py
```diff
@@ -59,38 +60,36 @@ class carmichael(Function):
...
-    with every number smaller than themselves.
+    Unfortunately, certain composite numbers (non-primes) still pass the Fermat
+    test with every number smaller than themselves.
     These numbers are called Carmichael numbers.
 
-    A Carmichael number will pass a Fermat primality test to every base b relatively prime to the number,
-    even though it is not actually prime. This makes tests based on Fermat's Little Theorem less effective than
-    strong probable prime tests such as the Baillie-PSW primality test and the Miller-Rabin primality test.
-    mr functions given in sympy/sympy/ntheory/primetest.py will produce wrong results for each and every
-    carmichael number.
```
oscargus, line 82: This sentence didn't make sense to me so I removed it. In what sense "wrong results"? (It may have been fixed since added though.)

oscarbenjamin:
I think this looks good. I'm not sure this meets the criteria outlined in the deprecation docs but I expect that these methods are not widely used and I support removing redundancy in the codebase in general.

oscargus:
Not sure either, but I guess these should never have been added in the first place. (With that said, I am not sure about the history, it may be that those methods were more complicated initially and not just simple wrappers, the documentation may indicate that.)

oscargus:
This would also be good to get in 1.11.

oscargus:
Restarting CI.

oscarbenjamin:
Looks good.

I edited the release note to say DEPRECATION rather than BREAKING CHANGE.

PR #23111, opened 2022-02-19
Add .free_dynamicsymbols to physics vectors.

moorepants:

Adds a new method to physics vectors that extracts the functions of time. This complements `.free_symbols()`.

#### Release Notes

* physics.vector
  * Adds a new method to physics vectors that extracts the functions of time: `.free_dynamicsymbols()`.


sympy/physics/vector/vector.py
```diff
@@ -751,20 +751,57 @@ def angle_between(self, vec):
...
+        Returns
+        =======
+        set of Symbol
+            set of symbols present in the measure numbers of
+            ``reference_frame``.
+
+        See Also
+        ========
+
+        - :meth:`~sympy.core.basic.Basic.free_symbols`
```
moorepants, line 772: I'm struggling to get these cross references working. I get this error:
```
/home/moorepants/src/sympy/sympy/physics/vector/vector.py:docstring of sympy.physics.vector.vector.Vector.free_dynamicsymbols:24: WARNING: py:obj reference target not found: -
/home/moorepants/src/sympy/sympy/physics/vector/vector.py:docstring of sympy.physics.vector.vector.Vector.free_symbols:24: WARNING: py:obj reference target not found: -
```
I've also tried `:obj:`~.free_symbols`` and `:obj:`sympy.physics.mechanics.functions.find_dynamicsymbols` as suggested in the docstring style guide and get the same error.

oscarbenjamin: For a method I think it should be `:meth:~.Basic.free_symbols` but for a property method it's probably better to use `:attr:` rather than `:meth:`.

moorepants: I tried:

```rst
        - :attr:`~.Basic.free_symbols`
        - :func:`~.find_dynamicsymbols`
```

And I get the same errors.

moorepants:
This failure is unrelated:

```
Ambiguous names in .mailmap
This email address appears for multiple entries:
Person: Akshānsh <akshansh@tuta.io>
Mailmap entries:
Akshansh Bhatt <akshansh@tuta.io> Akshansh <53227127+akshanshbhatt@users.noreply.github.com>
Akshansh Bhatt <akshansh@tuta.io> akshanshbhatt <53227127+akshanshbhatt@users.noreply.github.com>

Lines should be added to .mailmap to indicate the correct name and
email aliases for all commits.

https://github.com/sympy/sympy/wiki/Development-workflow#update-
mailmap

the next SymPy release.

	Akshānsh <akshansh@tuta.io>
	Tirthankar Mazumder <63574588+wermos@users.noreply.github.com>
```

sympy/physics/vector/vector.py
oscarbenjamin: Maybe it's not possible. I missed the fact that you're adding this in the "see also" section which is a numpydoc thing that seems to have a completely different set of referencing rules. I can't find how to refer to a method/attribute in the see also section in the docs for numpy doc.

ThePauliPrinciple:
Authors are required to have been correctly added to mailmap, but a member who merges a PR needs to be registered correctly too, otherwise the master branch suddenly fails the author check (eventhough the PR checks pass). This has happened a couple of times already. In the most recent master, this has already been solved, so simply pulling the master into this branch should resolve the error.

sympy/physics/vector/vector.py
moorepants: I wouldn't think that would matter. The fully qualified name should work in all cases, I would have thought. I just deleted them.

moorepants:
Ok, this passes now, I've removed the cross references since I can't figure them out.

sympy/physics/vector/vector.py
oscarbenjamin: I think that the "see also" uses a completely different "namespace" from ordinary sphinx referencing so it does matter.

moorepants: Interesting.

sidhu1012:
LGTM!!

sympy/physics/vector/vector.py
asmeurer: I see you figured this out, but yeah, the numpydoc extension takes the lines in "See Also" and rewrites them into cross references. It really ought to detect when the line looks like a cross reference itself and just leave it alone (or at least give a better error message).

moorepants: Ah! I actually didn't figure it out. I just deleted it.

PR #23112, opened 2022-02-19
Adjust statement about namespace implementation detail.

moorepants:

I came across this statement in the docstring guide and do not consider
module layout an "implementation detail". Modules without a prepended
underscore in their filename should be considered public API because
people will import from them, regardless if a sentence in the
documentation warns them not to or if the fully qualified name is not
displayed in the cross reference. The `sympy.module.<tab>` in IPython
will show all variables, including module names, that do not start with
an underscore, so people will import from anywhere. I have yet to come
across a library that shifts module layout without deprecation and
warning to users.

#### Release Notes

NO ENTRY

### DIFF
--- a/doc/src/guides/contributing/docstring.rst
+++ b/doc/src/guides/contributing/docstring.rst
@@ -673,16 +673,16 @@ of objects that have conventions here:
 1. Objects that are included in ``from sympy import *``, for example,
 ``sympy.acos``.
 
-For these, use ``:obj:`~.acos()```. The ``~`` makes it so that
-the text in the rendered HTML only shows ``acos``. Without it, it would use the
-fully qualified name ``sympy.functions.elementary.trigonometric.acos``. However,
-for names that are part of the global ``sympy`` namespace, we do not want to
-encourage accessing them from their specific submodule, as this is an
-implementation detail that could change. The ``.`` makes it so that the function
-name is found automatically. Sometimes, Sphinx will give a warning that there
-are multiple names found. If that happens, replace the ``.`` with the full name.
-For example, ``:obj:`~sympy.solvers.solvers.solve()```. For functions, methods,
-and classes, it is a convention to add () after the name to indicate such.
+For these, use ``:obj:`~.acos()```. The ``~`` makes it so that the text in the
+rendered HTML only shows ``acos``. Without it, it would use the fully qualified
+name ``sympy.functions.elementary.trigonometric.acos``. For names that are part
+of the global ``sympy`` namespace, we want to encourage accessing them from
+``sympy.*``, not the specific submodule. The ``.`` makes it so that the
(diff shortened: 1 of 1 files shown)


moorepants:
If I search for python files with `from sympy.simplify` I get 10k+ results:

https://github.com/search?l=&q=%22from+sympy.simplify%22+language%3APython&type=code

I can't figure out how to negate in a github search, but if I search for code in the path `sympy/simplify` that would likely identify people that have the sympy source in their repo:

https://github.com/search?q=%22from+sympy.simplify%22+path%3Asympy%2Fsimplify+language%3APython&type=Code&ref=advsearch&l=Python&l=

That's like 500 hits.

This isn't a fullproof careful test, but I think it shows that people likely do import from fully qualified path names from sympy.

moorepants:
My search isn't so great. I'd really like to do `NOT path:"*/sympy/simplify"` but I can't get globs to work.

moorepants:
I guess wildcards/globs don't work: https://github.com/isaacs/github/issues/402

oscargus:
I think the change to say how we want it to be imported, rather than how it not should be imported makes sense.

One can consider if we want to include a motivation for why or not. I think a motivation makes sense as it stresses the importance of following it, although the wording possibly should change.

doc/src/guides/contributing/docstring.rst
```diff
@@ -673,16 +673,16 @@ of objects that have conventions here:
...
+For these, use ``:obj:`~.acos()```. The ``~`` makes it so that the text in the
+rendered HTML only shows ``acos``. Without it, it would use the fully qualified
+name ``sympy.functions.elementary.trigonometric.acos``. For names that are part
+of the global ``sympy`` namespace, we want to encourage accessing them from
+``sympy.*``, not the specific submodule. The ``.`` makes it so that the
+function name is found automatically. Sometimes, Sphinx will give a warning
+that there are multiple names found. If that happens, replace the ``.`` with
+the full name.  For example, ``:obj:`~sympy.solvers.solvers.solve()```. For
+functions, methods, and classes, it is a convention to add () after the name to
+indicate such.
```
smichr, line 685:
```suggestion
For these, use ``:obj:`~.acos()```. The ``~`` makes it so that the text in the
rendered HTML only shows ``acos`` instead of the fully qualified name
``sympy.functions.elementary.trigonometric.acos``. (This will encourage importing
names from the global ``sympy`` namespace instead of a specific submodule.)
The ``.`` makes it so that the function name is found automatically. (If Sphinx gives
a warning that there are multiple names found, replace the ``.`` with
the full name.  For example, ``:obj:`~sympy.solvers.solvers.solve()```.) Adding a trailing
pair of parentheses is a convention for indicating the name is a function, method, or
class.
```

PR #23096, opened 2022-02-18
Changed the name of the node visiting function from `visit_Num` to `visit_Constant`.

wermos:
#### References to other Issues or PRs

Fixes #23092 

#### Brief description of what is fixed or changed
I changed the name of the node visiting function `visit_Num` to `visit_Constant` because the former is deprecated since version 3.8.

#### Release Notes

* parsing
  * Changed the name of the `visit_Num` node visitor function to `visit_Constant`, as the former is deprecated.

Issue #23092, opened 2022-02-17
ast_parser: PendingDeprecationWarning: visit_Num is deprecated; add visit_Constant


jwilk:
`sympy.parsing.ast_parser` triggers `PendingDeprecationWarning` in Python ≥ 3.8:
```console
$ python3 -Wd
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from sympy.parsing.ast_parser import parse_expr
>>> parse_expr('6 * 7', {})
/usr/lib/python3.8/ast.py:371: PendingDeprecationWarning: visit_Num is deprecated; add visit_Constant
  return visitor(node)
42
```

asmeurer:
This should be easy for someone to fix. I think we still support 3.7 so it will need to have a branch.

oscarbenjamin:
We should really run the tests in CI so that a PendingDeprecationWarning shows up as an error.

oscarbenjamin:
Reopening because the fix in #23096 was incomplete: https://github.com/sympy/sympy/pull/23096#discussion_r810234437

wermos:
There are two ways I can think of fixing this:
We can use the Python version info (which we can get using `sys.version_info`) to conditionally define the relevant function, C preprocessor-style.

Alternatively we can define both functions with identical code, with plans to remove `visit_Num` when we drop support for 3.7. If I understood the AST node visiting code correctly (which I think I did), the Interpreter tries to call `visit_Constant` before falling back to `visit_Num` in versions >= 3.8, so the existence of both functions won't result in a warning.

Since I'm new to the project, I am not sure which method we prefer, so I would appreciate feedback on this.

jwilk:
> we can define both functions with identical code

They can't be literally identical, because `ast.Constant.n` is deprecated too.
(Although at the moment you don't get any warning about this: <https://bugs.python.org/issue46797>.)

oscarbenjamin:
I would rather avoid an explicit Python version check if it isn't needed. I don't really know anything about ast parsing so I can't say what's the best approach though. Either way there should be comments making clear if something should be removed/changed in a future version of sympy.

wermos:
I think we can close this issue now.

@jwilk, I would appreciate if you could look over the new additions and see if we overlooked something again or not, before closing (since you were the one who found the issue with #23096) 😅

jwilk:
[...]

PR #23122, opened 2022-02-19
Put the `visit_Num` function back into `ast_parser.py`

wermos:
This was required to keep SymPy compatible with Python 3.7.

This is a follow-up to #23096, where it was discovered that the `visit_Num` must be kept for compatilibity with Python 3.7.
Fixes the new issues raised in #23092.

I added the `visit_Num` function back in. Now, `ast_parser.py` has both `visit_Constant` (for compatibility with Python >= 3.8), and `visit_Num` (for compatibility with Python 3.7).

I also changed the way the `value` attribute of the `Constant` node was being accessed, because `Constant.n` is deprecated in Python 3.11.

#### Release Notes

* parsing
  * Added the `visit_Num` function back into `ast_parser.py`.

### DIFF
--- a/sympy/parsing/ast_parser.py
+++ b/sympy/parsing/ast_parser.py
@@ -33,6 +33,17 @@ class Transform(NodeTransformer):
         self.global_dict = global_dict
 
     def visit_Constant(self, node):
+        if isinstance(node.value, int):
+            return fix_missing_locations(Call(func=Name('Integer', Load()),
+                    args=[node], keywords=[]))
+        elif isinstance(node.value, float):
+            return fix_missing_locations(Call(func=Name('Float', Load()),
+                    args=[node], keywords=[]))
+        return node
+
+    def visit_Num(self, node):
+        """This function exists for backwards compatibility with Python 3.7.
+           It should be removed when SymPy removes support for Python 3.7."""
         if isinstance(node.n, int):
             return fix_missing_locations(Call(func=Name('Integer', Load()),
                     args=[node], keywords=[]))
--- a/sympy/parsing/tests/test_ast_parser.py
+++ b/sympy/parsing/tests/test_ast_parser.py
@@ -1,3 +1,4 @@
+from sympy.core.numbers import Rational
 from sympy.core.singleton import S
 from sympy.core.symbol import symbols
 from sympy.parsing.ast_parser import parse_expr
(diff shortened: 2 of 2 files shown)


oscarbenjamin:
It would be good to have tests that show that this is working.

wermos:
I will add a couple of tests, as soon as I read the Sympy contribution guidelines regarding tests 😅

wermos:
@oscarbenjamin Is it possible to write tests based on Python version in `pytest`? Because we need to check for the `PendingDeprecationWarning` in Python >= 3.8, and it shouldn't raise any warnings for 3.7.

The best solution I can think of is using `sys.version_info`, which you said we should avoid if possible.

oscarbenjamin:
Why would we need to check the Python version? We want the same behaviour under all versions i.e. that the parsing works without emitting any warnings.

I would have thought that the idea would be to set up a warning filter that enabled PendingDeprecationWarning and then just have a test that does whatever parsing is needed and asserts that *no* warning is emitted. Then we just want to know that that test passes under all Python versions which will be checked in CI.

wermos:
Thank you for the explanation. There was a misunderstanding on my part, so I thought that checking for version info would be required.

wermos:
@oscarbenjamin Do you think these tests are sufficient?

oscarbenjamin:
The idea is that the test should fail without the changes here so that it demonstrates the need for the change. I just tried running the code from the test under sympy master on Python 3.7 and also 3.10 and I didn't get any error in either case.

Under what conditions would this test fail?

wermos:
@oscarbenjamin Thank you for the speedy reply.

The tests are _supposed_ to fail whenever a) the Python version is greater than or equal to 3.8 and `visit_Constant` does not exist, and b) the Python version is 3.7 and `visit_Num` does not exist.

I just tried it on my machine, and here are the results with Python 3.8:
screenshot of the parsing tests failing
This error showed up after I manually deleted the `visit_Constant` code which I have written. Note that this function is [already in master](https://github.com/sympy/sympy/blob/9c7767cf21c7b664b185c1e183c52f9be1ad4afc/sympy/parsing/ast_parser.py#L35) because #23096 was accepted into the mainline of work before @jwilk pointed out a flaw we had overlooked.

However, while investigating, I realized the test for Python 3.7 is incorrectly written because I forgot that `==` in Python checks for semantic equality, and not exact equality.

I have fixed this by checking the values using `==` and then checking the type using `isinstance`.

oscarbenjamin:
Okay, I can confirm that the test works and that both functions are needed to get the correct behaviour in Python 3.8+ and 3.7-.

Setting to automerge if tests pass.

Thanks!

PR #23125, opened 2022-02-19
Remove contractions in documentation

oscargus:

Primarily in documentation, but sometimes some exceptions are also changed.

#### Release Notes

NO ENTRY


smichr:
-1/10...This doesn't bother me. Not sure if it is the same with others.

oscargus:
Technical/scientific writing should not include it.

oscarbenjamin:
> Technical/scientific writing should not include it.

I agree but like @smichr my feelings are not that strong either way.

oscargus:
It is one thing to do the changes, another to merge it. :-)

PR #23126, opened 2022-02-19
Fix .mailmap

oscargus:
#### Release Notes

NO ENTRY

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -105,8 +105,8 @@ Akash Trehan <akash.trehan123@gmail.com>
 Akash Vaish <akash.9712@gmail.com>
 Akhil Rajput <akh1lrjput@gmail.com> Akhil Rajput <43316490+akhil-dh1@users.noreply.github.com>
 Akira Kyle <ak@akirakyle.com>
-Akshansh Bhatt <akshansh@tuta.io> Akshansh <53227127+akshanshbhatt@users.noreply.github.com>
-Akshansh Bhatt <akshansh@tuta.io> akshanshbhatt <53227127+akshanshbhatt@users.noreply.github.com>
+Akshansh Bhatt <akshansh@tuta.io>
+Akshansh Bhatt <akshansh@tuta.io> <53227127+akshanshbhatt@users.noreply.github.com>
 Akshat Jain <akshat.jain@students.iiit.ac.in>
 Akshat Maheshwari <akshat14714@gmail.com>
 Akshat Sood <68052998+akshatsood2249@users.noreply.github.com>


smichr:
Does `mailmap_check` flag this, too?

oscargus:
> Does `mailmap_check` flag this, too?

Yes.

.mailmap
```diff
@@ -105,8 +105,8 @@ Akash Trehan <akash.trehan123@gmail.com>
 Akash Vaish <akash.9712@gmail.com>
 Akhil Rajput <akh1lrjput@gmail.com> Akhil Rajput <43316490+akhil-dh1@users.noreply.github.com>
 Akira Kyle <ak@akirakyle.com>
-Akshansh Bhatt <akshansh@tuta.io> Akshansh <53227127+akshanshbhatt@users.noreply.github.com>
-Akshansh Bhatt <akshansh@tuta.io> akshanshbhatt <53227127+akshanshbhatt@users.noreply.github.com>
+Akshansh Bhatt <akshansh@tuta.io>
+Akshansh Bhatt <akshansh@tuta.io> <53227127+akshanshbhatt@users.noreply.github.com>
```
oscarbenjamin, line 109: This change should not have been necessary. There must be a bug somehow in the logic in the `mailmap_check.py` script.

oscargus: It seems like if you have the same email twice on the right hand side, possibly with different names this happens. (At least that is my best guess.)

oscarbenjamin: This error only emerged after a new commit from @akshanshbhatt was merged today:
https://github.com/sympy/sympy/commit/8c9cd9b4d394e19c410813d7566172f55531ebc0

I don't think that's a coincidence. Somehow the commits that exist on master affected the logic that checks for duplicate email addresses which it shouldn't.

oscarbenjamin: @ThePauliPrinciple clarified this here: https://github.com/sympy/sympy/pull/23111#issuecomment-1046128710

The change is because @akshanshbhatt is the author of the merge commit after merging #23074.

I still don't understand why the diff here in the .mailmap file is needed though. The original two lines should have been fine as they both just declare some other name/email to be  an alternative for `Akshansh Bhatt <akshansh@tuta.io>`.

Looking in git log the merge commit says:
```
commit 8c9cd9b4d394e19c410813d7566172f55531ebc0
Merge: 555ad6c 2f1fd71
Author: Akshānsh <akshansh@tuta.io>
Date:   Sat Feb 19 21:31:51 2022 +0530

    Merge pull request #23074 from faze-geek/bode_plot
    
    physics : Improved Bode's phase and magnitude plots
```
The name `Akshānsh` has a non-ASCII character that I don't see anywhere in the .mailmap. I'm not sure why that doesn't show as a separate author.

PR #22872, opened 2022-01-17
Created SetKind for Set class #22860 

me-t1me:
#### References to other Issues or PRs

Issue #22860 

#### Brief description of what is fixed or changed
Creates Class `SetKind` for all Subclasses of `Set` such as `Interval`, `FiniteSet`, `Union` etc

```python
>>> Interval(1,2).kind
SetKind(NumberKind)

>>> FiniteSet(Matrix([1, 2])).kind
FiniteSet(Matrix([1, 2])).kind
>>> FiniteSet(0, Matrix([1, 2])).kind
SetKind(UndefinedKind)
```

#### Release Notes

NO ENTRY

PR #23127, opened 2022-02-19
Series/Limits : Fixes leading term for exponential functions involving AccumBounds based arguments

### LINKED ISSUE #18473: Bug in limit

anutosh491:

Fixes #18473

On Master
```
>>> limit(sin(x)**(1/x), x, oo)
1
>>> limit(cos(x)**(1/x), x, oo)
1
>>> limit(tan(x)**(1/x), x, oo)
1
>>> exp(x*log(cos(1/x))).as_leading_term(x)
1
>>> exp(x*log(sin(1/x))).as_leading_term(x)
1
>>> exp(x*log(tan(1/x))).as_leading_term(x)
1
>>> log(cos(1/x)).as_leading_term(x)
log(AccumBounds(-1, 1))
>>> log(sin(1/x)).as_leading_term(x)
log(AccumBounds(-1, 1))
>>> log(tan(1/x)).as_leading_term(x)
log(AccumBounds(-oo, oo))
```
The limits shown above cannot be correctly claimed to be `1` . For eg the following
image

Firstly the function isn't continuous in the real domain as `x` tends to `oo` and even if it does the answer is not going to be `1` .
This is because something like `log(cos(1/x)).as_leading_term(x)` returns `log(AccumBounds(-1, 1))` which wouldn't be supportive in the `real` domain . Unfortunately a `subs` operation for `x` by `0` is taking place changing the `x*log(AccumBounds(-1, 1))` to `0` and returning `exp(0)` which is `1` . Hence the Pr fixes the above not really correct result.

On Committed branch . `S.NaN` seems to be the best option we have to tackle non real values of `log` inside `AccumBounds` !
```
>>> x = Symbol('x')
>>> limit(sin(x)**(1/x), x, oo)
Limit(sin(x)**(1/x), x, oo, dir='-')
>>> limit(cos(x)**(1/x), x, oo)
Limit(cos(x)**(1/x), x, oo, dir='-')
>>> limit(tan(x)**(1/x), x, oo)
Limit(tan(x)**(1/x), x, oo, dir='-')
>>> exp(x*log(cos(1/x))).as_leading_term(x)
nan
>>> exp(x*log(sin(1/x))).as_leading_term(x)
nan
>>> exp(x*log(tan(1/x))).as_leading_term(x)
nan
>>> log(cos(1/x)).as_leading_term(x)
nan
>>> log(tan(1/x)).as_leading_term(x)
nan
```

#### Other comments
All Discussion for this Pr had taken place in #23087 but due to couple wrong commands in the past my PR  was based on a master branch which wasn't a straight-copy of the main sympy master branch. Hence since past few Prs after switching to a new branch from master , I would get unwanted code pushed everytime  . Hence I was having problems in squashing, rebasing,reverting commits which was showing fatal error message. I followed @oscarbenjamin 's comment on another Pr helping another fellow contributor to fix a similar issue. This involved setting the main sympy repo as upstream again . This will help me in the future too as now I am sure this problem isn't going to occur again!
This is what I followed to sort the error (https://github.com/sympy/sympy/pull/22872#issuecomment-1031365942)

#### Release Notes

* series
  * Fixes leading term for exponential functions involving AccumBounds based arguments.

### DIFF
--- a/sympy/functions/elementary/exponential.py
+++ b/sympy/functions/elementary/exponential.py
@@ -529,8 +529,18 @@ class exp(ExpBase, metaclass=ExpMeta):
         return Add(*l)
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
+        from sympy.calculus.util import AccumBounds
         arg = self.args[0].cancel().as_leading_term(x, logx=logx)
         arg0 = arg.subs(x, 0)
+        if arg is S.NaN:
+            return S.NaN
+        if isinstance(arg0, AccumBounds):
+            # This check addresses a corner case involving AccumBounds.
+            # if isinstance(arg, AccumBounds) is True, then arg0 can either be 0,
+            # AccumBounds(-oo, 0) or AccumBounds(-oo, oo).
+            # Check out function: test_issue_18473() in test_exponential.py and
+            # test_limits.py for more information.
+            return exp(arg0)
         if arg0 is S.NaN:
             arg0 = arg.limit(x, 0)
         if arg0.is_infinite is False:
@@ -701,8 +711,10 @@ class log(Function):
         elif isinstance(arg, AccumBounds):
             if arg.min.is_positive:
                 return AccumBounds(log(arg.min), log(arg.max))
+            elif arg.min.is_zero:
+                return AccumBounds(S.NegativeInfinity, log(arg.max))
             else:
-                return
+                return S.NaN
         elif isinstance(arg, SetExpr):
             return arg._eval_func(cls)
 
@@ -1041,6 +1053,7 @@ class log(Function):
         return res + Order(x**n, x)
 
     def _eval_as_leading_term(self, x, logx=None, cdir=0):
+        from sympy.calculus.util import AccumBounds
         arg0 = self.args[0].together()
 
         arg = arg0.as_leading_term(x, cdir=cdir)
@@ -1055,6 +1068,8 @@ class log(Function):
             cdir = arg0.dir(x, cdir)
         if x0.is_real and x0.is_negative and im(cdir).is_negative:
             return self.func(x0) - 2*I*S.Pi
+        if isinstance(arg, AccumBounds):
+            return log(arg)
         return self.func(arg)
 
 
--- a/sympy/functions/elementary/tests/test_exponential.py
+++ b/sympy/functions/elementary/tests/test_exponential.py
@@ -11,7 +11,7 @@ from sympy.functions.elementary.complexes import (adjoint, conjugate, re, sign, 
 from sympy.functions.elementary.exponential import (LambertW, exp, exp_polar, log)
 from sympy.functions.elementary.hyperbolic import (cosh, sinh, tanh)
 from sympy.functions.elementary.miscellaneous import sqrt
-from sympy.functions.elementary.trigonometric import (cos, sin)
+from sympy.functions.elementary.trigonometric import (cos, sin, tan)
 from sympy.matrices.expressions.matexpr import MatrixSymbol
 from sympy.polys.polytools import gcd
 from sympy.series.order import O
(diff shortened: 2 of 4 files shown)


anutosh491:
CC: @jksuom . Sorry for the disorganization and confusion on the other Pr but due to an unwanted mistake with my master branch(as pointed in other comments in the OP) whenever I would try a rebase, squash or revert command , it used to show a fatal error message . But I realized Oscar had helped another fellow contributor to restore the master branch correctly , I realized after trying out all possible options that the error lies there and referring his comment I have fixed the error . This would also help me for future Pr's !
This looks good to go in . The authors file failure isn't a fault of this Pr and is being fixed through another Pr (#23126).
Thanks for your patience and reviews !

anutosh491:
ping @jksuom

jksuom:
This looks good now, thanks!

anutosh491:
> This looks good now, thanks! 

Thanks for the great reviews and suggestions!

anutosh491:
CC: @jksuom . I just realized something which we could improve on here . The inspiration for this was one of my latest prs implementing series and leading term for the frac functions #23044 , which is yet to be reviewed by you !
As `frac(1/x).as_leading_term(x)` is `AccumBounds(0, 1)`, I thought of adding a test which is a corner case like the above ones for eg ( `limit((cos(x) + 1)**(1/x), x, oo)` ) , which should be addressed by this part 
```
        if isinstance(arg0, AccumBounds):
            # This check addresses a corner case involving AccumBounds.
            # if isinstance(arg, AccumBounds) is True, then arg0 can either be 0,
            # AccumBounds(-oo, 0) or AccumBounds(-oo, oo).
            # Check out function: test_issue_18473() in test_exponential.py and
            # test_limits.py for more information.
            return exp(arg0)
```
So we have the following
```
>>> limit(frac(x)**(1/x), x, oo)
AccumBounds(0, 1)
```
image

But then I realized the implementing was missing out on cases involving `-oo` . 
```
# Currently on master
>>> limit((cos(x) + 1)**(1/x), x, oo)
AccumBounds(0, 1)
>>> limit((cos(x) + 1)**(1/x), x, -oo)
AccumBounds(0, 1)
```
But as the graph shows  our implementation is correct for `oo` but won't work for `-oo` implementation 
image
image

So if you think carefully the negative case is just the inversion of the positive case so this is solved by the following . Will be linking all expected results and graphs for the same in a comment below !
```
        if isinstance(arg0, AccumBounds):
            # This check addresses a corner case involving AccumBounds.
            # if isinstance(arg, AccumBounds) is True, then arg0 can either be 0,
            # AccumBounds(-oo, 0) or AccumBounds(-oo, oo).
            # Check out function: test_issue_18473() in test_exponential.py and
            # test_limits.py for more information.
            if cdir == 1:
                return exp(arg0)
            return 1/exp(arg0)
```

anutosh491:
After applying above suggested diff , we can see as while travelling towards positive infinity , the function steeply falls to `0` periodically , similarly while travelling in the opposite direction the function rises steeply to `oo` 

**1)** `(sin(x)**2)**(1/x)`
```
>>> limit((sin(x)**2)**(1/x), x, oo)
AccumBounds(0, 1)
>>> limit((sin(x)**2)**(1/x), x, -oo)
AccumBounds(1, oo)
```
image


**2)** `limit((tan(x)**2)**(2/x) , x, oo)`
```
>>> limit((tan(x)**2)**(2/x) , x, oo)
AccumBounds(0, oo)
>>> limit((tan(x)**2)**(2/x) , x, -oo)
AccumBounds(0, oo)
```
image

**3)** Also a test from the latest pr as reported above #23044 , which is where I spotted this improvement which is `frac(x)**(1/x)`
```
>>> limit(frac(x)**(1/x), x, oo)
AccumBounds(0, 1)
>>> limit(frac(x)**(1/x), x, -oo)
AccumBounds(1, oo)
```
image

anutosh491:
Other cases work perfectly fine at negative infinity too because , either the limit does not exist (where we return self) or the function approaches a constant value in both direction , so we don't need to touch them . For eg this function goes to `1` from both sides 
```
>>> limit((cos(x) + 2)**(1/x), x, oo)
1
>>> limit((cos(x) + 2)**(1/x), x, -oo)
1
```
image

I will

PR #23128, opened 2022-02-19
fix(physics.control): failing tests in test_control_plots.py

### LINKED ISSUE #22686: test_pole_zero in test_control_plots.py failing on master
In the following test, 
https://github.com/sympy/sympy/blob/6ebec9263f50082a65cce272c490eaf40aa04610/sympy/physics/control/tests/test_control_plots.py#L94-L116
the second, third, fourth, and sixth assertions are failing on master. However, the differences between the required and actual values are very small and may have been caused due to floating point error. For example, for the third assertion, the value returned by `_to_tuple(*pole_zero_numerical_data(tf2))` is `((0.0,), ((-0.49999999999999983+0.8660254037844384j), (-0.49999999999999983-0.8660254037844384j), (1+0j)))`, while the required value is `((0.0,), ((-0.5000000000000004+0.8660254037844395j), (-0.5000000000000004-0.8660254037844395j), (0.9999999999999998+0j))) ` 

This should probably be changed to check if the difference between the expected and returned values is smaller than something small like `1e-6` for every tuple element.

akshanshbhatt:
Fixed failing tests in test_control_plots.py

https://github.com/sympy/sympy/pull/23074#issuecomment-1045992888
https://github.com/sympy/sympy/pull/21763#issuecomment-939192905

Fixes #22686 

- Numerical computation results depend heavily on the hardware on which it is computed, so comparing for equality can lead to assertion errors as it did in the case of `test_pole_zero()`. I fixed it by allowing a tolerance range. Because the order of roots computed by NumPy can sometimes vary (especially when there are multiple imaginary roots), it is **not** assured that the test will pass even after allowing the values for some variation. For this reason, I have added the `XFAIL` marker on the test.
- `23964739753087.42` and `23964739753087.418` are very close numbers, but their difference is much more significant than `10e-8` (allowed tolerance). Instances of entries like this made the other three tests fail. A simple solution was to add a condition to deal with these big outlier entries.
- Referenced control module in test_optional_dependencies.py.

#### Release Notes

* physics.control
  * Fixed the failing tests in control_plots module.


sympy/physics/control/tests/test_control_plots.py
```diff
@@ -92,33 +92,35 @@ def test_errors():
     raises(ValueError, lambda: ramp_response_plot(tf1, slope=-0.1))
 
 
+@XFAIL
```
oscargus, line 95: Why is this XFAIL:ed? Seems like if anything an XFAIL-should be removed.

Also, why is this skipped based on matplotlib? Seems like this is only numerical methods?

Is numpy requited to generate the numerical data? If not, one should maybe create a local version of `allclose`? There is an `isclose` in math that one can iterate over.

sympy/physics/control/tests/test_control_plots.py
```diff
@@ -149,7 +151,7 @@ def test_bode_data(sys):
 
 
 def check_point_accuracy(a, b):
-    return all(Abs(a_i - b_i) < 1e-12 for \
+    return all(((Abs(a_i - b_i) < 1e-12) or (Abs((a_i / b_i) - 1) < 1e-8)) for \
```
oscargus, line 154: I think it should be possible to replace this with `math.isclose`. If one only makes it possible to traverse arbitrary structure of numbers (tuples of tuples etc), is should be possible to use that for the above tests. Something like:

sympy/physics/control/tests/test_control_plots.py
akshanshbhatt: > Why is this XFAIL:ed? Seems like if anything an XFAIL-should be removed.

**The test runs perfectly fine on my computer.** However, the issue is that the order of roots computed by NumPy (`np.roots()` is used in the pole-zero plot) seems to vary, either due to updated code in newer releases of NumPy or due to different CPU architecture (I'm not sure). The order of complex roots returned by the `pole_zero_numerical_data()` earlier was different from now. It can lead to exceptions when running the test suite since we make a one-to-one comparison for each array element.

Is there any way to ensure that the roots are returned in a fixed order?

I added this XFAIL marker because I thought this test might fail for the users for which the order of computed roots is different. I can remove it if someone can confirm that the order of roots returned is same for every user.

> Is numpy requited to generate the numerical data? If not, one should maybe create a local version of `allclose`? There is an `isclose` in math that one can iterate over.

This seems to be an excellent option. Only the `pole_zero_numerical_data()` function requires NumPy for calculation. All other functions in the module utilize numeric computation native to sympy.

oscarbenjamin:
Thanks!

PR #23135, opened 2022-02-20
PermuteDims: added easier API

Upabjojr:
PermuteDims and permutedims now have an easier API based on index orders. The corresponding permutation is then derived internally.

Now you don't need to know the permutation anymore, just the old index order and the new index order.

NO ENTRY

### DIFF
--- a/sympy/tensor/array/expressions/array_expressions.py
+++ b/sympy/tensor/array/expressions/array_expressions.py
@@ -714,6 +714,10 @@ class PermuteDims(_CodegenArrayAbstract):
                 raise ValueError("Permutation not defined")
             return PermuteDims._get_permutation_from_index_orders(index_order_old, index_order_new, dim)
         else:
+            if index_order_new is not None:
+                raise ValueError("index_order_new cannot be defined with permutation")
+            if index_order_old is not None:
+                raise ValueError("index_order_old cannot be defined with permutation")
             return permutation
 
     @classmethod
--- a/sympy/tensor/array/expressions/array_expressions.py
+++ b/sympy/tensor/array/expressions/array_expressions.py
@@ -414,7 +414,7 @@ class PermuteDims(_CodegenArrayAbstract):
 
     >>> from sympy.tensor.array.expressions import ArraySymbol, PermuteDims
     >>> M = ArraySymbol("M", (1, 2, 3, 4, 5))
-    >>> expr = PermuteDims(M, index_order_old=["i", "j", "k", "l", "m"], index_order_new=["k", "i", "j", "m", "l"])
+    >>> expr = PermuteDims(M, index_order_old="ijklm", index_order_new="kijml")
     >>> expr
     PermuteDims(M, (0 2 1)(3 4))
     >>> expr.shape
--- a/sympy/tensor/array/expressions/array_expressions.py
+++ b/sympy/tensor/array/expressions/array_expressions.py
@@ -410,6 +410,16 @@ class PermuteDims(_CodegenArrayAbstract):
     >>> cg.shape
     (2, 3)
 
+    There are optional parameters that can be used as alternative to the permutation:
+
+    >>> from sympy.tensor.array.expressions import ArraySymbol, PermuteDims
+    >>> M = ArraySymbol("M", (1, 2, 3, 4, 5))
+    >>> expr = PermuteDims(M, index_order_old=["i", "j", "k", "l", "m"], index_order_new=["k", "i", "j", "m", "l"])
+    >>> expr
+    PermuteDims(M, (0 2 1)(3 4))
+    >>> expr.shape
+    (3, 1, 2, 5, 4)
+
     Permutations of tensor products are simplified in order to achieve a
     standard form:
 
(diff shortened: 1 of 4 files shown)


ThePauliPrinciple:
Instead of having the named arguments, perhaps some simple logic can be used to detect whether `perm` is an iterable or an iterable of length 2 of iterables? Or just a check whether there are 2 or 3 arguments.

I'm not against adding this, but why is this considered an easier API? Can you give an example where there is a problem identifying the new order of the indices but where it is easy to define the permutation with the new API? Is there any example of another library/program which defines permutations along the lines of the proposed API?

Upabjojr:
> Instead of having the named arguments, perhaps some simple logic can be used to detect whether `perm` is an iterable or an iterable of length 2 of iterables? Or just a check whether there are 2 or 3 arguments.

I was actually thinking about not overloading the main method and instead create a different method. I ended up deciding to keep the old method in this PR. Requiring the argument names to be explicit is a better distinction from the old method.

> I'm not against adding this, but why is this considered an easier API?

Well, have you tried using the current API? Guessing the correct permutation isn't that easy after all.

> Can you give an example where there is a problem identifying the new order of the indices but where it is easy to define the permutation with the new API?

Usually people reason with indices, not with permutations. This API is more user-friendly.

> Is there any example of another library/program which defines permutations along the lines of the proposed API?

No idea.

ThePauliPrinciple:
Yes, I have used the current API. Since an `Array` doesn't have named indices, isn't the current API just a shorthand for `index_order_old = [0,1,2,..n], index_order_new = perms`?

Upabjojr:
> Yes, I have used the current API. Since an `Array` doesn't have named indices, isn't the current API just a shorthand for `index_order_old = [0,1,2,..n], index_order_new = perms`?

Yes. But sometimes you have the inverse permutation, sometimes you just reason with indices.

I am just giving the users the possibility of avoiding reasoning with permutations. There might be users who are not familiar with them.

Upabjojr:
I am going to merge this PR as it is useful for writing a tutorial for people without experience in permutations.

Upabjojr:
> Personally I find the string format of numpy's einsum really convenient for these purposes, e.g. `"cba->abc"`. (it could easily be integrated in this PR, by allowing `perms` to be a string and then using `.split()`s to obtain `index_order_old` and `index_order_new`)

Well, `"cba->abc"` does not really make it clear which one are the old and new indices. You can interpret it either way.

Upabjojr:
I am trying to minimize the possibility of confusion in the permutation direction: the permutation shall remap the new indices to the old ones, that's our convention.

But it is not the only possible interpretation, some people could reason starting from the _old_ indices by finding the permutation to the new ones. This reasoning would use the inverse permutation.

By adding _old_ and _new_ it's pretty clear which ones are the previous and following axes orders.

ThePauliPrinciple:
Ah I see, the docs were using lists of single letter strings, so I was under the impression that this was enforced, but it seems that if single letters are enough to identify your indices you can already just use strings like:
`permutedims(b, index_order_old="cba", index_order_new="abc")`

sympy/tensor/array/expressions/array_expressions.py
```diff
@@ -410,6 +410,16 @@ class PermuteDims(_CodegenArrayAbstract):
     >>> cg.shape
     (2, 3)
 
+    There are optional parameters that can be used as alternative to the permutation:
+
+    >>> from sympy.tensor.array.expressions import ArraySymbol, PermuteDims
+    >>> M = ArraySymbol("M", (1, 2, 3, 4, 5))
+    >>> expr = PermuteDims(M, index_order_old=["i", "j", "k", "l", "m"], index_order_new=["k", "i", "j", "m", "l"])
```
smichr, line 417:
```suggestion
    >>> expr = PermuteDims(M, index_order_old="ijklm", index_order_new="kijml")
```

smichr:
I, too, prefer a mutable interpretation of indices so one can just do `permutedims(A, old, new)` or `permutedims(A, perm)` instead of having to use named args or a None place holder as `permutedims(A, None, old, new)`
```python
def permutedims(A, p, old=None):
   if new is None:
        p = Permutation(p)  # this will check it for you and raise an error if there are issues; p(iterable) -> reordered iterable
    else:
        assert len(p) == len(old) and all(type(i) == type(j) for i, j in zip(p, old))
```
Putting it this way retains the idea that the 2nd arg is the permutation, defined, perhaps, in terms of the order as it was before. If that seems backward then `old` can be changed to `new` with `p` being the old form. My preference is for the former.

Upabjojr:
> I, too, prefer a mutable interpretation of indices so one can just do `permutedims(A, old, new)` or `permutedims(A, perm)` instead of having to use named args or a None place holder as `permutedims(A, None, old, new)`

I don't like the fact that the first argument is overloaded (that is, both a permutation and a list of indices). I would rather create another function, say `permutedims_by_indices` and a static method `PermuteDims.by_indices`.

But that's not the point. By making the user write the whole parameter name, i.e. `index_order_old`, it is clear whether the indices are the old or new ones. It just improves the readability of the code.

Advanced users with no issue of readability should just use the permutation object. The new arguments are just meant to both improve readability and decrease ambiguity in the code.

Upabjojr:
Also notice that Wolfram Mathematica uses the inverse permutation as convention: 
https://reference.wolfram.com/language/ref/TensorTranspose.html

That is, in Mathematica you have to specify the permutation from the _old_ indices to the _new_ ones, which is the inverse of SymPy.

SymPy follows the same convention as [NumPy's transpose](https://numpy.org/doc/stable/reference/generated/numpy.transpose.html).

I hope I've been clear on why I would like users to explicitly write the words _old_ and _new_ regarding to the indices in this PR. It minimizes any possible ambiguity.

smichr:
Since there are two conventions in the computing environment that SymPy shares, it makes sense to be explicit. Any reservations, @ThePauliPrinciple ?

smichr:
OK, let's go with what you have. `list` is not necessary in the tests, either, but that's a nit.

Issue #23144, opened 2022-02-21
UndefinedFunction.diff broken in sympy 1.9, 1.10rc2


mkoeppe:
```
>>> import sympy
>>> sympy.__version__
'1.10rc2'
>>> from sympy.abc import x, y
>>> from sympy import Function
>>> f = Function('f')
>>> f.diff(x)
1
>>> f.diff(x,x)
1
>>> f.diff(x, 2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mkoeppe/s/sage/sage-rebasing/local/var/lib/sage/venv-python3.9/lib/python3.9/site-packages/sympy/core/expr.py", line 3526, in diff
    return _derivative_dispatch(self, *symbols, **assumptions)
  File "/Users/mkoeppe/s/sage/sage-rebasing/local/var/lib/sage/venv-python3.9/lib/python3.9/site-packages/sympy/core/function.py", line 1919, in _derivative_dispatch
    return Derivative(expr, *variables, **kwargs)
  File "/Users/mkoeppe/s/sage/sage-rebasing/local/var/lib/sage/venv-python3.9/lib/python3.9/site-packages/sympy/core/function.py", line 1287, in __new__
    raise ValueError("First variable cannot be a number: %i" % v)
ValueError: First variable cannot be a number: 2
>>> f.diff(x,2,y,2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mkoeppe/s/sage/sage-rebasing/local/var/lib/sage/venv-python3.9/lib/python3.9/site-packages/sympy/core/expr.py", line 3526, in diff
    return _derivative_dispatch(self, *symbols, **assumptions)
  File "/Users/mkoeppe/s/sage/sage-rebasing/local/var/lib/sage/venv-python3.9/lib/python3.9/site-packages/sympy/core/function.py", line 1919, in _derivative_dispatch
    return Derivative(expr, *variables, **kwargs)
  File "/Users/mkoeppe/s/sage/sage-rebasing/local/var/lib/sage/venv-python3.9/lib/python3.9/site-packages/sympy/core/function.py", line 1287, in __new__
    raise ValueError("First variable cannot be a number: %i" % v)
ValueError: First variable cannot be a number: 2
```

(seen in https://github.com/sympy/sympy/runs/4929630109?check_suite_focus=true, failure for `src/sage/symbolic/expression_conversions.py`)

[...]

PR #23145, opened 2022-02-21
Derivative, ArrayDerivative, MatrixExpr: Accept numbers.Integral as order of differentiation, index

mkoeppe:

First reported in #23144.

`Derivative` accepts integers to indicate order of differentiation with respect to a variable. 
We generalize the test for integers from a hardcoded list of types to the numbers ABC `numbers.Integral`. (Such objects can be sympified to `Integer` because the existence of an `__int__` method is [guaranteed by the ABC](https://docs.python.org/3/library/numbers.html#numbers.Integral).) This fixes the following regression when passing Sage integers:

```
sage: var('y')
y
sage: f_sage = function('f_sage')(x, y)
sage: f_sage
f_sage(x, y)
sage: f_sympy = f_sage._sympy_()
sage: f_sympy
f_sage(x, y)
sage: f_sympy.diff(x, 2)
ValueError: 
Can't calculate derivative wrt 2.
```

We make similar changes to `ArrayDerivative` to fix the following problem:
```
sage:     from sympy.matrices.expressions.matexpr import MatrixSymbol
sage:     from sympy.tensor.array.array_derivatives import ArrayDerivative
....:     A = MatrixSymbol("A", 2, 2)
sage: arrder = ArrayDerivative(A, A, 2, evaluate=False)
ValueError: 
Can't calculate derivative wrt 2.
```

We make similar changes to `MatrixSymbol`, in particular the `__getitem__` method.
```
sage: from sympy import MatrixSymbol
....: A = MatrixSymbol('A', 3, 3)
sage: A[1]
IndexError: Invalid index, wanted A[i,j]
```

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -32,6 +32,7 @@ There are three types of functions implemented in SymPy:
 
 from typing import Any, Dict as tDict, Optional, Set as tSet, Tuple as tTuple, Union as tUnion
 from collections.abc import Iterable
+import numbers
 
 from .add import Add
 from .assumptions import ManagedProperties
@@ -1276,12 +1277,11 @@ class Derivative(Expr):
         # derivative.
         variable_count = []
         array_likes = (tuple, list, Tuple)
-        integer_likes = (int, Integer)
 
         from sympy.tensor.array import Array, NDimArray
 
         for i, v in enumerate(variables):
-            if isinstance(v, integer_likes):
+            if isinstance(v, (int, Integer, numbers.Integral)):
                 if i == 0:
                     raise ValueError("First variable cannot be a number: %i" % v)
                 count = v
--- a/sympy/matrices/expressions/matexpr.py
+++ b/sympy/matrices/expressions/matexpr.py
@@ -1,5 +1,6 @@
 from typing import Tuple as tTuple
 from functools import wraps
+import numbers
 
 from sympy.core import S, Integer, Basic, Mul, Add
 from sympy.core.assumptions import check_assumptions
@@ -271,7 +272,7 @@ class MatrixExpr(Expr):
 
     def valid_index(self, i, j):
         def is_valid(idx):
-            return isinstance(idx, (int, Integer, Symbol, Expr))
+            return isinstance(idx, (int, Integer, numbers.Integral, Symbol, Expr))
         return (is_valid(i) and is_valid(j) and
                 (self.rows is None or
                 (0 <= i) != False and (i < self.rows) != False) and
@@ -291,7 +292,7 @@ class MatrixExpr(Expr):
                 return self._entry(i, j)
             else:
                 raise IndexError("Invalid indices (%s, %s)" % (i, j))
-        elif isinstance(key, (SYMPY_INTS, Integer)):
+        elif isinstance(key, (SYMPY_INTS, Integer, numbers.Integral)):
             # row-wise decomposition of matrix
             rows, cols = self.shape
             # allow single indexing if number of columns is known
@@ -313,8 +314,8 @@ class MatrixExpr(Expr):
         raise IndexError("Invalid index, wanted %s[i,j]" % self)
 
     def _is_shape_symbolic(self) -> bool:
-        return (not isinstance(self.rows, (SYMPY_INTS, Integer))
-            or not isinstance(self.cols, (SYMPY_INTS, Integer)))
+        return (not isinstance(self.rows, (SYMPY_INTS, Integer, numbers.Integral))
+            or not isinstance(self.cols, (SYMPY_INTS, Integer, numbers.Integral)))
 
     def as_explicit(self):
         """
--- a/sympy/tensor/array/array_derivatives.py
+++ b/sympy/tensor/array/array_derivatives.py
@@ -1,4 +1,5 @@
 from typing import Optional
+import numbers
 
 from sympy.core.expr import Expr
 from sympy.core.function import Derivative
@@ -85,7 +86,7 @@ class ArrayDerivative(Derivative):
         # object, the default in `Basic` will call a loop over
         # `_eval_derivative`:
 
-        if not isinstance(count, (int, Integer)) or ((count <= 0) == True):
+        if not isinstance(count, (int, Integer, numbers.Integral)) or ((count <= 0) == True):
             return None
 
         # TODO: this could be done with multiple-dispatching:


sympy/matrices/expressions/matexpr.py
```diff
@@ -313,8 +314,8 @@ def __getitem__(self, key):
         raise IndexError("Invalid index, wanted %s[i,j]" % self)
 
     def _is_shape_symbolic(self) -> bool:
-        return (not isinstance(self.rows, (SYMPY_INTS, Integer))
-            or not isinstance(self.cols, (SYMPY_INTS, Integer)))
+        return (not isinstance(self.rows, (SYMPY_INTS, Integer, numbers.Integral))
```
ThePauliPrinciple, line 317: Is this needed? I don't think `self.rows` and `self.cols` should be anything but a SymPy object.

ThePauliPrinciple:
Since
```python
>>> import gmpy2 as gmpy
>>> isinstance(gmpy.mpz(0), Integral)
True
>>> isinstance(0, Integral)
True
```
Shouldn't it just replace `SYMPY_INTS` (and `int`)  rather than be there in addition?

It's not clear to me if there could be any unwanted side effects.

sympy/core/function.py
```diff
@@ -1276,12 +1277,11 @@ def __new__(cls, expr, *variables, **kwargs):
         # derivative.
         variable_count = []
         array_likes = (tuple, list, Tuple)
-        integer_likes = (int, Integer)
 
         from sympy.tensor.array import Array, NDimArray
 
         for i, v in enumerate(variables):
-            if isinstance(v, integer_likes):
+            if isinstance(v, (int, Integer, numbers.Integral)):
```
ThePauliPrinciple, line 1284: `numbers.Integral` should be correct here, since it ensures you can perform addition and convert to `int`, which should be all we really need here. The question is, do we really want to start supporting this, if we do it somewhere, it would imply that it should be possible in other places too.

SymPy should probably be clearer on what types it allows in its API and when it becomes the responsibility of the user/downstream library to make sure the types are correct.

oscarbenjamin:
> ```python
> >>> isinstance(gmpy.mpz(0), numbers.Integral)
> ```

I think this is a new feature only in the most recent gmpy2 versions. With `gmpy2==2.0.8` you get:
```python
In [1]: import gmpy2, numbers

In [2]: isinstance(gmpy2.mpz(0), numbers.Integral)
Out[2]: False
```

sympy/core/function.py
oscarbenjamin: It looks like sage has been using sympy like this for some time. Really what sage depends on is sympify turning its types into sympy types. Usually that's fine because we can just sympify things. The problem here is that sympify isn't being used at the outset because there was a problem with doing that hence the changes in #22551. Maybe it would still be possible to use sympify though but with more careful logic rather than just passing everything in.

ThePauliPrinciple: I think if we check whether `v` is an `array_like` first, then we can safely use `_sympify` on the remaining `v`. Or just try to sympify and if it fails check for `array_like`s

sympy/matrices/expressions/matexpr.py
```diff
@@ -271,7 +272,7 @@ def I(self):
 
     def valid_index(self, i, j):
         def is_valid(idx):
-            return isinstance(idx, (int, Integer, Symbol, Expr))
+            return isinstance(idx, (int, Integer, numbers.Integral, Symbol, Expr))
```
ThePauliPrinciple, line 275: How about
```python
def is_valid(idx):
    try:
        idx = _sympify(idx)
        return isinstance(idx, (Integer, Symbol, Expr))
    except SympifyError:
        return False
```

sympy/tensor/array/array_derivatives.py
```diff
@@ -85,7 +86,7 @@ def _dispatch_eval_derivative_n_times(cls, expr, v, count):
         # object, the default in `Basic` will call a loop over
         # `_eval_derivative`:
 
-        if not isinstance(count, (int, Integer)) or ((count <= 0) == True):
+        if not isinstance(count, (int, Integer, numbers.Integral)) or ((count <= 0) == True):
```
ThePauliPrinciple, line 89: No idea if I understand correctly (it's really hard to trace down how you arrive here), but I think this shouldn't be needed if the count in `function.py` is converted.

sympy/matrices/expressions/matexpr.py
ThePauliPrinciple: I have never used sage, but it might be possible to trace down how something other than `int`/`Integer` ends up here by adding these lines to this method
```python
        if not isinstance(self.rows, (SYMPY_INTS, Integer)):
            print(type(self.rows), self)
        if not isinstance(self.cols, (SYMPY_INTS, Integer)):
            print(type(self.cols), self)
```
and running tests.

sympy/matrices/expressions/matexpr.py
oscarbenjamin: Both `Integer` and `Symbol` are subclasses of `Expr` so just `isinstance(idx, Expr)` should do it.

sympy/tensor/array/array_derivatives.py
oscarbenjamin: Yes, that's what I thought. I also haven't traced how to get here because the differentiation code is so convoluted now.

sympy/matrices/expressions/matexpr.py
ThePauliPrinciple: I don't have a good way to check for `numbers` that are compatible with SymPy without converting though, it was the same thing I got stuck on when trying to improve `__eq__` for `Number` classes. Perhaps ABCs from `numbers` is the right approach, but that might not be stable enough (e.g. gmpy2 only being compatible in newer versions). It is also not clear to me how that would work precisely.

oscarbenjamin: Yes, I mean after sympifying.

ThePauliPrinciple: I got that, but ideally we would not need to convert `int` to `Integer` just to check if it is a valid index. Now putting in an extra instance check for `int` would be easy, but then why not also for `mpz` when it is available? (i.e. `SYMPY_INTS`). But if we go that route, we might indeed allow `numbers.Integral`. But that would possibly introduce a bug. I don't know if it will always work in that case.

Maybe the right approach is to check if it is an `Expr` and then try to convert it with `int(idx)`.

sympy/core/function.py
mkoeppe: > The question is, do we really want to start supporting this

It was supported; it's a regression introduced in the SymPy 1.10 series

oscarbenjamin: It was supported via sympify rather than in terms of the numbers ABCs. It previously worked because the sage Integer type defines `_sympy_` but would fail with some other class that happened to implement the `numbers.Integral` ABC.

ThePauliPrinciple: The support was that these arguments were converted with `sympify`, not that anything which passes an instance check with `numbers.Integral` is allowed here. I think that means that this does not only revert the regression, but also extends what types are allowed.

ThePauliPrinciple: Something like this should handle any class/object that is correctly set up to work with SymPy but nothing else.

```diff
diff --git a/sympy/core/function.py b/sympy/core/function.py
index 171231ce24..bc70aaba92 100644
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -47,7 +47,7 @@
 from .parameters import global_parameters
 from .rules import Transform
 from .singleton import S
-from .sympify import sympify
+from .sympify import sympify, SympifyError, _sympify

 from .sorting import default_sort_key, ordered
 from sympy.utilities.exceptions import (sympy_deprecation_warning,
@@ -1276,12 +1276,16 @@ def __new__(cls, expr, *variables, **kwargs):
         # derivative.
         variable_count = []
         array_likes = (tuple, list, Tuple)
-        integer_likes = (int, Integer)

         from sympy.tensor.array import Array, NDimArray

         for i, v in enumerate(variables):
-            if isinstance(v, integer_likes):
+            try:
+                sv = _sympify(v)
+            except SympifyError:
+                sv = v
+            if isinstance(sv, Integer):
+                v = sv
                 if i == 0:
                     raise ValueError("First variable cannot be a number: %i" % v)
                 count = v
```

oscarbenjamin: Presumably the reason you didn't do it like that before was to avoid calling `sympify` on `v` if it was an array-like. I think it probably makes sense just to invert the logic in the loop so that `array_likes` is checked first and then `_sympify` is called after.

ThePauliPrinciple: Actually, `Basic._do_eq_sympify` could also be used here, although it seems that I have chosen the name a bit unfortunately. maybe `_do_external_sympify` would have been better.

Inverting the logic didn't seem trivial to me btw, there is a reason why it isn't just one large if-elif block.

mkoeppe: > The support was that these arguments were converted with `sympify`, not that anything which passes an instance check with `numbers.Integral` is allowed here. I think that means that this does not only revert the regression, but also extends what types are allowed.

It does not actually extend it: `sympify` is able to handle all instances of `numbers.Integral` because the method `__int__` is defined.
```
>>> class Eight():
...     def __int__(self): return 8
... 
>>> sympy.sympify(Eight())
8
>>> type(_)
<class 'sympy.core.numbers.Integer'>
```

mkoeppe:
Is `MatrixExpr.valid_index` intended to be public API?

sympy/core/function.py
oscarbenjamin: Perhaps to put it another way: it would be better if all handling of foreign types was done in one place. Ideally this function should depend on sympify doing the right thing and then it shouldn't need to worry about things like the Integral ABC. If sympify can handle the numbers ABCs then the rest of sympy can benefit from that.

sympy/matrices/expressions/matexpr.py
mkoeppe: Is `MatrixExpr.valid_index` intended to be public API? I agree that its only library use, from `MatrixExpr.__getitem__`, would not need the change to handle `numbers.Integral` because it is run on sympified data... but then neither would it need to test for `int`!
As it seems to be public, the change to allow `numbers.Integral` is an improvement.

mkoeppe:
As #23162/#23166 fix this, we can close this PR.

PR #23146, opened 2022-02-22
more autosimplification with infinite terms

### LINKED ISSUE #23120: interactions between finite and infinite quantities
```python
>>> 1/(var('f', finite=True) + oo)
1/(f + oo)
>>> 1/oo
0
```
If symbolic expressions are finite, it seems like they should disappear when added to `oo`.

smichr:

fixes #23120 

`real*oo + nonzero + 1/nonzero == real*oo`

#### Release Notes

* core
    * multiplying a nonzero (finite) number by +/-oo returns +/-oo

### DIFF
--- a/sympy/core/add.py
+++ b/sympy/core/add.py
@@ -204,6 +204,7 @@ class Add(Expr, AssocOp):
         from sympy.tensor.tensor import TensExpr
         rv = None
         if len(seq) == 2:
+            # this is only to handle the case of adding a Rational
             a, b = seq
             if b.is_Rational:
                 a, b = b, a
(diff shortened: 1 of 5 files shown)


sympy/core/add.py
```diff
@@ -315,10 +316,17 @@ def flatten(cls, seq):
         # now let's construct new args:
         # [2*x**2, x**3, 7*x**4, pi, ...]
         newseq = []
+        imfinite = []
```
oscargus, line 319: cmplxfinite? (Not that it matters as such, but maybe a better name?)

And similar two lines down.

smichr: Hmm...I'll change this back to a draft and think more about this.

Issue #10805, opened 2016-03-12
Safety


The-Penultimate-Defenestrator:
I think `sympy` shoudl be made safer for untrusted input, as discussed in the comments on [this stackoverflow issue](http://stackoverflow.com/q/35929640/4414003). Since `sympy` uses `eval` under the hood, it's vulnerable to attacks. There's really no reason to allow attribute access, etc, which leads to vulnerabilities like [these](http://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html). I'd love a safer sympy that I could really use with foreign input.

asmeurer:
My idea is to disable attribute access. That doesn't fix everything (see https://stackoverflow.com/questions/35804961/python-eval-is-it-still-dangerous-if-i-disable-builtins-and-attribute-access), but does fix most things, and is completely unnecessary for what sympify needs.

asmeurer:
Unless I'm missing something this should be fixable at the tokenizer level.

asmeurer:
The other option is to do an AST transformation from Python AST -> SymPy AST. This would almost certainly be slower, however (it's basically re-inventing `eval`, but in Python).

The-Penultimate-Defenestrator:
Itnlooks like "f-strings" could be blacklisted pretty easily as well. Are there any plans to implement better safety?

The-Penultimate-Defenestrator:
So, if you have this idea and want to implement it, do you have any kind of timeline for when this might go in? I've looked at `stringify_expr` but it doesn't seem easy to get simpler code from this without `eval`.

asmeurer:
No one is working on it as far as I know, so I can't give a timeline. I'm not familiar enough with the internals of the parser to know how easy this would be. Is it as simple as blacklisting tokens from the tokenizer or does it need to be done at the AST level?

The-Penultimate-Defenestrator:
Cool. I think parsing with AST and whitelisting is a good approach, I'm not sure if `tokenize` can recognize attribute access. I'll look into it and see if the scope of this would be small enough that I could make a pull request. Are pull requests accepted here?

asmeurer:
Yes, absolutely.

The-Penultimate-Defenestrator:
I just spent about 10 minutes on this and came up with a pretty simple solution. Can you see any holes in the following code?
```python
def is_safe(inp_string):
    """ Blacklist attribute access, simply by checking for any period that is
[...]

PR #12524, opened 2017-04-10
Add safe flag to sympify()

asmeurer:
True (the default) causes it to raise UnsafeSympifyError on strings with AST elements that aren't in a whitelist (in particular, things like attribute access, comprehensions, and f-strings are not allowed). It also does a filter on the namespace to disallow names that might be used to bypass it, including `sympify` itself. See the added notes in the `sympify` docstring for more information. 

This requires several str printers that used `.` to be changed, so that they can still be round-tripped with `sympify`. The one that I haven't changed yet is new assumptions predicates, like `Q.positive`, as I'm not sure how I feel about it. Maybe we should only require the `srepr` printer to be fully compatible with safe sympify? 

**Example**

```py
>>> sympify('a.b')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "./sympy/core/sympify.py", line 434, in sympify
    transformations=transformations, evaluate=evaluate, safe=safe)
  File "./sympy/parsing/sympy_parser.py", line 954, in parse_expr
    code = stringify_expr(s, local_dict, global_dict, transformations, safe=safe)
  File "./sympy/parsing/sympy_parser.py", line 856, in stringify_expr
    check_string_for_safety(res)
  File "./sympy/parsing/safe_parser.py", line 134, in check_string_for_safety
    raise UnsafeSympifyError(s, reason='Non-whitelisted AST node %s found' % node)
sympy.core.sympify.UnsafeSympifyError: "Symbol ('a' ).b " cannot be parsed safely. Reason: Non-whitelisted AST node <_ast.Attribute object at 0x11fa2a198> found.
>>> sympify('sympify("a.b", safe=False)')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "./sympy/core/sympify.py", line 434, in sympify
    transformations=transformations, evaluate=evaluate, safe=safe)
  File "./sympy/parsing/sympy_parser.py", line 954, in parse_expr
    code = stringify_expr(s, local_dict, global_dict, transformations, safe=safe)
  File "./sympy/parsing/sympy_parser.py", line 856, in stringify_expr
    check_string_for_safety(res)
  File "./sympy/parsing/safe_parser.py", line 139, in check_string_for_safety
    raise UnsafeSympifyError(s, reason='disallowed name %r found' % node.id)
sympy.core.sympify.UnsafeSympifyError: 'sympify ("a.b",safe =False )' cannot be parsed safely. Reason: disallowed name 'sympify' found.
```

Fixes #10805.

PR #23147, opened 2022-02-22
Offer a "controlled" eval option for `sympify`.

skieffer:

This is an attempt to address #10805 

Provides a `controlled` keyword arg to `sympify`, which means that instead of using built-in `eval`, we use a custom AST node visitor to evaluate the expression.

The node visitor only allows whitelisted callables, and tries to limit arithmetic on excessively large operands.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/sympify.py
+++ b/sympy/core/sympify.py
@@ -99,7 +99,7 @@ def _convert_numpy_types(a, **sympify_args):
 
 
 def sympify(a, locals=None, convert_xor=True, strict=False, rational=False,
-        evaluate=None):
+        evaluate=None, controlled=True):
     """
     Converts an arbitrary expression to a type that can be used inside SymPy.
 
@@ -492,7 +492,7 @@ sympify() performed the string fallback resulting in the following string:
 
     try:
         a = a.replace('\n', '')
-        expr = parse_expr(a, local_dict=locals, transformations=transformations, evaluate=evaluate)
+        expr = parse_expr(a, local_dict=locals, transformations=transformations, evaluate=evaluate, controlled=controlled)
     except (TokenError, SyntaxError) as exc:
         raise SympifyError('could not parse %r' % a, exc)
 
--- a/sympy/parsing/sympy_parser.py
+++ b/sympy/parsing/sympy_parser.py
@@ -14,7 +14,10 @@ import types
 from sympy.assumptions.ask import AssumptionKeys
 from sympy.core.basic import Basic
 from sympy.core import Symbol
-from sympy.core.function import arity, Function
+from sympy.core.function import arity, Function, UndefinedFunction
+from sympy.core.numbers import Integer
+from sympy.core.relational import Relational
+from sympy.core.sympify import sympify, _sympify
 from sympy.utilities.iterables import iterable
 from sympy.utilities.misc import filldedent, func_name
 from sympy.functions.elementary.miscellaneous import Max, Min
@@ -918,7 +921,7 @@ def eval_expr(code, local_dict, global_dict):
 
 
 def parse_expr(s, local_dict=None, transformations=standard_transformations,
-               global_dict=None, evaluate=True):
+               global_dict=None, evaluate=True, controlled=True):
     """Converts the string ``s`` to a SymPy expression, in ``local_dict``
 
     Parameters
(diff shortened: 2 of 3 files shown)


sympy/core/sympify.py
```diff
@@ -99,7 +99,7 @@ def _convert_numpy_types(a, **sympify_args):
 
 
 def sympify(a, locals=None, convert_xor=True, strict=False, rational=False,
-        evaluate=None):
+        evaluate=None, controlled=True):
```
ThePauliPrinciple, line 102: Is `True` by default the right choice? Does this come with some performance hit to the library while it is not needed for private use? Perhaps a global parameter to switch the default behaviour might be useful.

skieffer: Same answer, this is just here temporarily for experimentation, to force all unit tests to go through the new evaluator.

asmeurer: Why "controlled" instead of "safe" like at https://github.com/sympy/sympy/pull/12524?

sympy/parsing/sympy_parser.py
```diff
@@ -1256,3 +1273,225 @@ def __getitem__(self, t):
         return tuple([_transformation[_] for _ in i])
 
 T = _T()
+
+
+class ControlledEvaluationException(Exception):
```
asmeurer, line 1278: This should subclass SympifyError.

asmeurer: Although maybe not. Is SympifyError used inside of ParseExpr?

asmeurer:
We might want to do the Mul and Pow limiting stuff separately. I expect that will be more complicated. It's also basically a duplicate of https://github.com/sympy/sympy/issues/6835. The idea there is to make `Mul` and `Pow` themselves automatically keep big numbers unevaluated. If that were implemented, then `sympify('10**10**10')` would just create `Pow(10, Pow(10, 10))` which would remain unevaluated as `Pow(10, 10000000000)`.

sympy/parsing/sympy_parser.py
```diff
@@ -1086,13 +1089,36 @@ def parse_expr(s, local_dict=None, transformations=standard_transformations,
...
+            # Here evaluation is achieved by applying another node visitor
+            # to the AST.
+            ce = ControlledEvaluator(local_dict, global_dict)
+            node = ast.Expr(expr_value)
+            rv = ce.visit(node)
+        else:
+            # FIXME: the assertion was just here to ensure we were testing the
+            #  new method in all unit tests. What's the right way to test?
+            #  Duplicate all tests in `test_sympify.py` and `test_sympy_parser.py`?
+            assert False
```
asmeurer, line 1113: You can also trigger this with `sympify("parse_expr('__import__(\"os\").system(\"ls\")', controlled=False)")`.

sympy/core/sympify.py
skieffer: I just didn't want to use "safe" until I was 100% sure. I think I'd feel better about using that word with an implementation such as discussed [here](https://github.com/sympy/sympy/issues/10805#issuecomment-1047956743), where the whitelist of callables would be explicit, and would list only functions each of which had been thoroughly audited.

sympy/parsing/sympy_parser.py
skieffer: Ah, so `parse_expr` would have to be removed from the whitelist as well (like `sympify`).

skieffer:
> We might want to do the Mul and Pow limiting stuff separately.

Sounds good. I'm in favor of focusing on one problem at a time, and the ideas from #6835 sound promising.

sympy/core/sympify.py
asmeurer: It's true that "safe" may imply a stronger guarantee than we can actually provide. But I also feel like it's a much more self descriptive name for the argument. It's also not easy to change keyword argument names in the future so we should try to pick a good one.

sympy/parsing/sympy_parser.py
skieffer: It's not used in there yet, but I don't see why it couldn't be.

asmeurer:
Some of the test cases from https://github.com/sympy/sympy/pull/12524 can be ported over here.

asmeurer:
This approach is nice in that it lets us keep attribute access. However, I does also seem to widen the possible attack surface. If some rogue SymPy object does something in its `__getattr__` or `__getitem__` it could be exploited. It wouldn't even have to be a public class that does it. You can pretty easily dig into the namespace of any sympy module (for instance, `simplify.__globals__` gives every variable defined in sympy/simplify/simplify.py).

PR #23150, opened 2022-02-22
BUG: respect global_dict in parse_expr (master branch)

### LINKED ISSUE #23148: BUG: sympy 1.10rc2 breaks unyt
Minimal Reproducer:
```
python3.8 -m venv .env
source .env/bin/activate/
pip install -U pip
pip install --pre sympy
pip install unyt

python -c "import unyt"
```
output
```
  File "<string>", line 1, in <module>
  File "/private/tmp/.env/lib/python3.8/site-packages/unyt/__init__.py", line 75, in <module>
    from unyt import unit_symbols
  File "/private/tmp/.env/lib/python3.8/site-packages/unyt/unit_symbols.py", line 33, in <module>
    _namespace[_alt_name] = _Unit(_canonical_name, registry=_registry)
  File "/private/tmp/.env/lib/python3.8/site-packages/unyt/unit_object.py", line 227, in __new__
    raise UnitParseError(
unyt.exceptions.UnitParseError: Unit representation must be a string or sympy Expr. 'Min' has type '<class 'sympy.core.function.FunctionClass'>'.
```

oscarbenjamin:
If parse_expr is called with a global_dict then builtins will not be
added to the global_dict so that builtin names do not conflict with
possible symbol names for the parsed expression e.g.:
```
>>> parse_expr("sum", global_dict={"Symbol":Symbol}) == Symbol("sum")
True
```

Fixes https://github.com/sympy/sympy/issues/23148 on the 1.10 branch
Same as #23149 but for the master branch

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/parsing/sympy_parser.py
+++ b/sympy/parsing/sympy_parser.py
@@ -1054,6 +1054,14 @@ def parse_expr(s, local_dict=None, transformations=standard_transformations,
     if global_dict is None:
         global_dict = {}
         exec('from sympy import *', global_dict)
+
+        builtins_dict = vars(builtins)
+        for name, obj in builtins_dict.items():
+            if isinstance(obj, types.BuiltinFunctionType):
+                global_dict[name] = obj
+        global_dict['max'] = Max
+        global_dict['min'] = Min
+
     elif not isinstance(global_dict, dict):
         raise TypeError('expecting global_dict to be a dict')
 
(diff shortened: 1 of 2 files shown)


asmeurer:
Instead of disabling this completely when globals is passed, couldn't it just not overwrite names that are already in the dict?

asmeurer:
Oh actually thinking about it this is correct. If you pass a globals dict (not a locals dict) that should overwrite the behavior of builtins.

PR #23159, opened 2022-02-24
Furo docs theme with styling

asmeurer:

This is based off of https://github.com/sympy/sympy/pull/22864, but I've updated the styling of the Furo theme. This is not at all finished yet, and we also haven't formally decided to switch to Furo yet either. 

Demo site here: https://www.asmeurer.com/sympy-furo-demo/

I've done a lot of styling, but there are still things that can be improved. Please build the docs and play around with it. If you find something that is low contrast or the wrong color, it probably means it is something I didn't notice.

Also please try both the light and dark theme (you can click the icon at the top of the page).

I haven't changed any fonts or font sizes, except for the code examples. It's a possibility to do this, but I'd like feedback. 

CC @bertiewooster 

#### Other comments

**TODOs**

- [x] Check if fonts should be adjusted (different fonts, sizes, etc.)
- [x] Add a version selector to the sidebar
- [x] Adjust the logo size on mobile (or alternately, make the whole sidebar scrollable)
- [x] Fix all remaining contrast issues
- [x] Remove all text referencing SymPy Live

#### Release Notes

- other
  - The documentation now uses the [Furo](https://pradyunsg.me/furo/) Sphinx theme.


asmeurer:
Note: we should probably rebase away the submodule addition before merging any of this.

asmeurer:
I've uploaded a demo site so you don't have to build this yourself https://github.com/asmeurer/sympy-furo-demo

bertiewooster:
Thanks @asmeurer. To save folks a click, the built site is at https://www.asmeurer.com/sympy-furo-demo/

bertiewooster:
Great start! Comments:

1. The Search box appears to connect to the body of the page, because they're about the same color. The Furo example site has the search box the same color as the rest of the left navbar. Is that possible and advisable? I can imagine that having a dark green background would make it hard to read black text being input; is it possible to make the input text a light color?
2. [Lighthouse reports](https://googlechrome.github.io/lighthouse/viewer/?psiurl=https%3A%2F%2Fwww.asmeurer.com%2Fsympy-furo-demo%2Fmodules%2Fcore.html%23module-sympy.core.random&strategy=mobile&category=performance&category=accessibility&category=best-practices&category=seo&category=pwa&utm_source=lh-chrome-ext) that the accessibility is very good, at 98/100. The one problem it found is `Background and foreground colors do not have a sufficient contrast ratio.`
    1. One situation where this occurs is the light green text in a function template light gray box, for example [`sympy.core.sympify.`](https://www.asmeurer.com/sympy-furo-demo/modules/core.html#id1). It also occurs for links (for example, "[next section](https://www.asmeurer.com/sympy-furo-demo/tutorial/gotchas.html#tutorial-gotchas-symbols)"), which are light green on white background. Perhaps make those types of text dark green instead of light? Personally, I prefer lighter colors for large page elements, so maybe make the left navbar a light green, and the text in the function template a dark green?
    2. Another is the [text in a code block](https://www.asmeurer.com/sympy-furo-demo/tutorial/intro.html#what-is-symbolic-computation). Perhaps make the code block background color lighter?

asmeurer:
> The Search box appears to connect to the body of the page, because they're about the same color. The [Furo example site](https://sphinx-themes.org/sample-sites/furo/) has the search box the same color as the rest of the left navbar. Is that possible and advisable? I can imagine that having a dark green background would make it hard to read black text being input; is it possible to make the input text a light color?

Sure that's doable. I'm not sure how good it would look. Having it blend in does make it harder to see. I guess the flip side of that is that it's less distracting. The current docs site has the search bar white https://docs.sympy.org/dev/index.html.

> [Lighthouse reports](https://googlechrome.github.io/lighthouse/viewer/?psiurl=https%3A%2F%2Fwww.asmeurer.com%2Fsympy-furo-demo%2Fmodules%2Fcore.html%23module-sympy.core.random&strategy=mobile&category=performance&category=accessibility&category=best-practices&category=seo&category=pwa&utm_source=lh-chrome-ext) that the accessibility is very good, at 98/100. The one problem it found is Background and foreground colors do not have a sufficient contrast ratio.

Yeah, I noticed some of these things as well (the Chrome inspect accessibility thing flagged them). 

I'm a little unsure how to fix some of them. A problem I noticed with these color contrast checkers is that they only consider the contrast of the text against its background. They don't consider the contrast of the text with the surrounding text. This is particularly important for the link color. We can make the (currently green) links have a better contrast against the background by making them darker, but then they become harder to see as links against the surrounding black text (you can see what I mean if you right click "inspect" a link in Chrome and then click the color in the inspector and choose the suggested higher contrast color).

I can try flipping the dark and light greens to see how it looks. 

I was also planning on adjusting the code block background (which should be easier than trying to restyle all the different syntax highlighting colors). The problem is the Chrome dev tools only support suggesting foreground colors, not background colors. But I did find https://accessible-colors.com/ which allows both. AAA is hard to achieve while still keeping distinguishable colors, but we should be able to get AA I think. I also tried just messing with the color picker, but I had a hard time finding something I liked aesthetically. I do personally like how the code blocks stand out as I think the examples are the most important part of documentation.

doc/src/conf.py
```diff
@@ -142,7 +142,52 @@
...
+
+    "admonition-font-size": "var(--font-size--normal)",
+    "admonition-title-font-size": "var(--font-size--normal)",
+    "code-font-size": "var(--font-size--small)",
+    "font-stack--monospace": 'DejaVu Sans Mono,"SFMono-Regular",Menlo,Consolas,Monaco,Liberation Mono,Lucida Console,monospace;'
+}
+
+html_theme_options = {
+    "light_css_variables": css_settings_common,
+    "dark_css_variables": {**css_settings_common,
```
pradyunsg, line 183: You shouldn't need this -- the light mode variables get inherited for dark mode.

pradyunsg: See https://pradyunsg.me/furo/customisation/colors/#how-light-and-dark-mode-work for the relevant details.

pradyunsg:
> Add a version selector to the sidebar

Relevant here: https://pradyunsg.me/furo/customisation/sidebar/. See also https://github.com/pradyunsg/furo/issues/372.

moorepants:
This is looking pretty nice. I like the green!
I noticed that the API rendering isn't so great:
image

The name of the function/class/method doesn't stand out. The bold "parameters" "returns" etc overwhelms the name of the function. The name of the functions should be the most distinct thing on the page. Also the notes are way too strong (in color I guess):
image

This also detracts from the function name which should be the most prominent thing.

moorepants:
If you scroll through the api listing, the function names should be the only thing you can detect while scrolling (or the primary thing).

namannimmo10:
Thanks for this.  The new docs site looks really nice!

doc/src/conf.py
asmeurer: I didn't realize this. It would be nice if this were mentioned at https://pradyunsg.me/furo/customisation/#light-css-variables-dark-css-variables

asmeurer: This was working, but I updated to the latest Furo and it stopped working and I had to separate them again.

asmeurer:
Addressed some comments and additional things I noticed:
- Removed the fancy scrolling when jumping to an anchor (it was too slow on long pages)
- Made the API headings easier to see. They are also now the only part of the page that is colored. 
- "See also" and "Notes" now use a gray background instead of green. "Warning" is still red. 
- Removed the extra vertical space in the parameters section.
- Made the "Parameters", "Returns", "Examples", and so on more prominent. 
- Added a GitHub icon to the footer.
- Made the code block background color slightly dimmer to improve the color contrast. 
- Various small fixes to improve color contrast, especially in dark mode.

I had one idea also, which is to make the function names easier to "spot" on the page by inverting the colors to make the background green (it is currently grey), and the text black (or some other color that's easy to read on green). I'm a little unsure about it, though. Let me know if you think it would be better that way or the way it is now.

asmeurer:
Another note: we are doing enough modifications in custom.css that it might be a good idea to pin the Furo version. We can update it every once in a while, but whenever we do, we have to check if any CSS modifications are required. Even without custom.css, after updating my local version Furo, there were some additional "supported" CSS variables added that required updating.

asmeurer:
> I had one idea also, which is to make the function names easier to "spot" on the page by inverting the colors to make the background green (it is currently grey), and the text black (or some other color that's easy to read on green). I'm a little unsure about it, though. Let me know if you think it would be better that way or the way it is now.

Here's a quick mockup of what I mean. Right now, it looks like this:

Screen Shot 2022-03-03 at 5 18 45 PM

I'm wondering if we should instead change it to something like this:

Screen Shot 2022-03-03 at 5 18 34 PM

Or would that be too noisy on the page? It would only apply to function and class definitions in the API docs, not other headers.

bertiewooster:
Thanks for doing all this detailed work, @asmeurer. I do find the second version visually noisy, as I prefer blocks of color to be subtle; that also makes it easier to achieve sufficient color contrast. How about largely keeping the first version, and making the module prefix names text the darker green? The actual function name could be bold to distinguish it from the module prefixes.

asmeurer:
Pushed a few more fixes:
- The logo is now high resolution.
- Various contrast fixes for dark mode, especially in the syntax highlighting. 

I'm honestly still not sure how to handle the lighthouse contrast issues in light mode. Unlike dark mode, most of the suggestions for contrast improvements in light mode seem to make the text too close to black to be distinguishable. 

Is there an easy way to run lighthouse on every page, with both light and dark mode? I've been running it one page at a time, but different pages have different elements. There are way too  many pages to run each page manually.

moorepants:
I like both, definitely fixes the issues I reaised. Thanks! For the second, the green bars are strong. Maybe a lighter shade of green would be better. The first one is good too though, and sufficient.

bertiewooster:
> Is there an easy way to run lighthouse on every page, with both light and dark mode?

[Lighthouse can be run via npm](https://www.npmjs.com/package/lighthouse); I haven't tried that.

I'm not sure if there's a way to automatically do that in both light and dark mode, as the URL doesn't change when I click dark mode. The mode selection does persist when you go from one page to another, so perhaps you could run the Lighthouse scans in light mode, then change to dark mode, then re-running the Lighthouse scans might use the dark mode? I'm not sure if that mode will be picked up by the npm version of Lighthouse.

asmeurer:
It looks like you can force dark mode with the `--chrome-flags`. But I don't see an easy way to use this to scan every page. It would be nice if there were a tool that just scanned the local HTML files.

asmeurer:
Another thing we can do with Furo is to add a header to each page in the development docs like this:
Screen Shot 2022-03-04 at 6 30 36 PM

bertiewooster:
> Another thing we can do with Furo is to add a header to each page in the development docs like this:

I was thinking it'd be nice to _temporarily_ add headers to each page (in both released and dev docs) when we want to publicize a survey or the like to the community--that would reach anyone who uses the docs, not just people who follow SymPy on Github, Twitter, Gitter, etc. Is there an easy way to do that?

bertiewooster:
> I don't see an easy way to use this to scan every page. It would be nice if there were a tool that just scanned the local HTML files.

I think Lighthouse only works if the page is posted on a server. I wonder if there's a way to give it a sitemap to have it visit all the mapped pages?

asmeurer:
The announcement feature makes it quite easy to add it to the dev docs. To add it to the "latest" docs we would need to manually rebuild the docs from that git tag with the given header (which isn't too hard).

asmeurer:
Some updates

- The current page is now highlighted in the sidebar. For example: Screen Shot 2022-03-21 at 6 28 25 PM

- I've made it so that the whole sidebar scrolls. Previously the logo would be pinned to the top of the page and only the table of contents would scroll. This should make the sidebar more usable on mobile, where the logo takes up a large amount of vertical space.

- I've added a version selection menu to the sidebar. You can see what it looks like at the bottom of the sidebar at https://www.asmeurer.com/sympy-furo-demo/dev/index.html.
   Screen Shot 2022-03-21 at 6 25 21 PM

   This uses a file called `versions.json` which will be placed at the root of the sympy_doc repo, which looks like

   ```json
   {
       "latest": "1.10",
       "dev": "1.11.dev"
   }
   ```

   If you build the docs locally the file won't be present and the docs will look like this 
   Screen Shot 2022-03-21 at 6 23 17 PM

   This is much simpler than the [current solution](https://github.com/sympy/sympy_doc/blob/gh-pages/generate_indexes.py) which injects code into the page after it is built. It also has the advantage that the version selector now appears on every page, not just the index page. Some new code will need to be added to the release process to update the versions.json file. 

- I have removed all references to the SymPy Live extension (mostly from the tutorial). If we ever add a replacement that uses pyiodide, we can reintroduce this text. For now, the extension has been removed because it is not compatible with the Furo theme, and has other issues as well. 

This is now ready for review. I have made all the style changes I wanted to make, so please let me know if you see anything that you feel should be changed in any way. Also if anyone is knowledgeable about CSS or Javascript I would appreciate a review of my CSS in `custom.css` and Javascript in `versions.html`, as I'm not great at either language.

oscarbenjamin:
> do you want me to add this to the release scripts or do you want to do it?

It's best if you do this. The relevant place is here:
https://github.com/sympy/sympy/blob/ff5755cecd165cfde49adbe6b4b976dc486fccf3/release/update_docs.py#L60

> This would need to be run whenever either the `latest` or `dev` version changes

Updates to the dev docs are still handled by Travis:
https://github.com/sympy/sympy/blob/ff5755cecd165cfde49adbe6b4b976dc486fccf3/.travis.yml#L121

If we can migrate that to GitHub Actions then we can ditch the Travis scripts as well.

asmeurer:
> It's best if you do this. The relevant place is here:

Is the update_docs.py script just run directly by you on the command line or is it called from somewhere else? If I am reading it correctly, we will need to add a new flag to pass in the dev version in addition to the release version (unless there's another way to get that).

oscarbenjamin:
> Is the update_docs.py script just run directly by you on the command line or is it called from somewhere else?

I run it directly. The sequence is:

1. Push tag and wait for CI
2. Download release files from CI
3. Run `release/github_release.py`
4. Run `twine upload`
5. Run `release/update_docs.py`
6. Edit release notes
7. Send an email to the mailing list.

It's fine to pass in the dev version manually to the update_docs.py script.

asmeurer:
I updated the release script. It's kind of hard to test, so you'll have to let me know if everything works properly when we do the next release. For now, I have added `versions.json` to the doc repo manually. I will delete the `generate_indexes.py` script there once this is merged.

oscarbenjamin:
> It's kind of hard to test, so you'll have to let me know if everything works properly when we do the next release.

We can check when that happens. To be clear it's not that hard to test though:

1. Clone the docs repo as ../sympy_doc
2. Checkout your PR branch
3. Change the version in sympy/release.py to 1.11
4. Run `release/build_docs.py 1.11 .`
5. Run `release/update_docs.py ../sympy_doc ./sympy-docs-xxx.zip 1.11`

If you **don't** pass `--push` to update_docs.py it will do everything except for the final push so you can open `index.html` and check how it looks.

asmeurer:
OK, I was able to get it working (and fixed a bug with it). Note that the last line will now be `release/update_docs.py ../sympy_doc ./sympy-docs-xxx.zip 1.11 1.12.dev`.

Also note that if you just open `index.html` in your browser, it will say "other versions not found". This is because browsers' cross-origin policies will not allow Javascript to load the versions.json file when using a `file://` URL. If you instead use a web server, like `python -m http.server` it will load the versions correctly (and it will obviously work correctly on the real docs.sympy.org site).

bertiewooster:
Coming together nicely!

## Color issues

- Contrast is a bit low for links of regular-sized text. (This is the only accessibility issue I've found using Lighthouse, giving pages a score of 98/100, very good.)
- Bright green text color (h2's, links, etc.) strikes me as a little garish. 

Possible solution to both: Could the green in the body text (h2 and links) be the dark green, and the green background of the navbar be a lighter green? (This may have already been tried and found unsatisfactory?)

## Content
The following are sort of tangential, could be a separate pull request: On homepage, suggest renaming top-level categories, and making them the same in the body and in the left navbar:

- SymPy Tutorials -> Tutorials
- SymPy Guides -> How-to Guides (only left navbar name to be changed)
- SymPy API Reference / Reference Documentation -> API Reference

Similarly, moving pages from Miscellaneous into one of the existing Diataxis categories, probably Guides.

asmeurer:
> Nitpick: seems like there are two different green link colors, for example on [style guide](https://www.asmeurer.com/sympy-furo-demo/dev/guides/contributing/documentation-style-guide.html), for example link in sphinx.ext.linkcode seems lighter green, and link in numpydoc seems darker green.

The darker green means the link has been visited. Maybe this is too subtle? I've noticed that many websites these days don't even distinguish these any more (in the old days, unvisited links were blue and visited links were purple). Note that our current docs theme does this as well. 

> Contrast is a bit low for links of regular-sized text. (This is the only accessibility issue I've found using Lighthouse, giving pages a score of 98/100, very good.)

My issue with this has been making sure the contrast of the link color is itself high enough relative to the surrounding text. If you let the Chrome dev tools suggest a better contrast color, it will be easier to read against the background, but will look almost indistinguishable from the surrounding black text. That's bad because you want to be able to tell that a link is a link. 

Here's the current style with no links visited

Screen Shot 2022-03-24 at 4 44 34 PM

Here's the current style with the first link visited

Screen Shot 2022-03-24 at 4 45 13 PM

And here's the current style with all links visited

Screen Shot 2022-03-24 at 4 44 11 PM

Perhaps we can just use latter for all links, and have no distinction between visited and unvisited. 

Personally, I prefer the current style, but I also don't have issues seeing color (both are readable to me), so I'm not sure just how bad the contrast issue of the current link color is. 

Just to be clear, the current link color is #29A329, which has a ratio of [3.3](https://whocanuse.com/?b=ffffff&c=29a329&f=16&s=) on a white background and the current visited link color is #307748, which has a ratio of [5.44](https://whocanuse.com/?b=ffffff&c=307748&f=16&s=) on a white background. AA is 4 and AAA is 7. I've found that AAA is basically untenable because it makes the links indistinguishable from the surrounding black text.

In dark mode, #307748 has lower contrast in dark mode ([3.86](https://whocanuse.com/?b=000000&c=307748&f=16&s=)), whereas #29A329 has good contrast ([6.37](https://whocanuse.com/?b=000000&c=29a329&f=16&s=)). Here, Chrome's suggested 7 contrast ratio color [#2FB92F](https://whocanuse.com/?b=000000&c=2fb92f&f=16&s=) is actually distinghable from the surrounding white text, at least to my eyes. 

I just found https://webaim.org/resources/linkcontrastchecker/, which actually lets you check link text against the surrounding text and the background text. According to it, using [#29A329 for dark mode](https://webaim.org/resources/linkcontrastchecker/?fcolor=FFFFFF&bcolor=000000&lcolor=29A329) and [#307748 for light mode](https://webaim.org/resources/linkcontrastchecker/?fcolor=000000&bcolor=FFFFFF&lcolor=307748) are both good contrast against both the background and the surrounding text (AA). It also correctly identifies that Chrome's "contrast ratio 7" suggestions have too poor contrast for the link text against the surrounding text. For instance, Chrome suggests [#2FB92F](https://webaim.org/resources/linkcontrastchecker/?fcolor=FFFFFF&bcolor=000000&lcolor=2FB92F) for dark mode, which is too low contrast against white text, and [#1A661A](https://webaim.org/resources/linkcontrastchecker/?fcolor=000000&bcolor=FFFFFF&lcolor=1A661A) for light mode, which is too low contrast against black text.

Dark mode is going to have better contrast in general, since I also adjusted the syntax highlighting colors to be better in dark mode.

**So tl;dr: my suggestion is to ditch the current visited/unvisited distinction for link colors and use the current darker "visited" color in light mode and the current lighter "unvisited" color in dark mode.** If we want to keep the visited/unvisited distinction we will need to think harder about this. 

> Bright green text color (h2's, links, etc.) strikes me as a little garish.

Regarding the green h2, this is actually only the case on the main page. It's because on that page, [the headers are links](https://github.com/sympy/sympy/blob/master/doc/src/index.rst). As I recall, we did this so that the sections are very evident as clickable things on the page, while still being part of the top-level table of contents. If you have any suggestions on a better way to do this, I'm open to it. I don't think it's a perfect solution because this also makes them headers with anchors, which is confusing. I don't think we should change the styling of links everywhere else just because this one thing happens to be rendering as a link right now. 

We actually have some pretty inconsistent use of headers throughout the documentation, which Furo is going to make more obvious. One example is that sometimes we have a header per reference function, like at https://www.asmeurer.com/sympy-furo-demo/dev/modules/core.html, which produces a list of all the functions in the sidebar. In other cases we only have top-level headers, like at https://www.asmeurer.com/sympy-furo-demo/dev/modules/ntheory.html. We should probably prefer the former style since it produces the nice list in the sidebar. But it's not trivial: the latter style generally comes from using `automodule` rather than explicitly listing each function with `autofunction` in the `rst`, so if we switch to always explicitly listing each function, we need to make better use of things like the `bin/coverage_doctest.py` script to make sure we don't omit functions from Sphinx. Alternately, we might be able to modify autodoc so that `automodule` automatically produces headers (this might be possible by messing with autodoc templates, I'm not sure). 

There are other problems too. For example, if you click on "number theory" on the sidebar, it takes you to this page, which is kind of useless: https://www.asmeurer.com/sympy-furo-demo/dev/reference/public/numbertheory/index.html. You have to click on the headers under it to go to the actual page on number theory. But those headers should really be on the right sidebar, not the left one. This is an issue with the way the rst pages and tables of contents are laid out for the number theory page, not with Furo.

> Possible solution to both: Could the green in the body text (h2 and links) be the dark green, and the green background of the navbar be a lighter green? (This may have already been tried and found unsatisfactory?)

Lighter green in the sidebar definitely does not look good. There are actually two lighter green colors we are using, and both are too bright for a background color.

Screen Shot 2022-03-24 at 4 35 45 PM
Screen Shot 2022-03-24 at 4 35 34 PM

And it's obviously even worse with dark mode. It's actually kind of nice with the current background color that it looks good in both light and dark mode. The current sidebar green is dark enough to be unobtrusive but also gives the SymPy docs its flair.

> The following are sort of tangential, could be a separate pull request: On homepage, suggest renaming top-level categories, and making them the same in the body and in the left navbar:

The content is independent of the theme. Let's make those changes in a separate PR. They can be made before or after this is merged.

bertiewooster:
> I just found https://webaim.org/resources/linkcontrastchecker/, which actually lets you check link text against the surrounding text and the background text.

Great resource. Thanks!



> So tl;dr: my suggestion is to ditch the current visited/unvisited distinction for link colors and use the current darker "visited" color in light mode and the current lighter "unvisited" color in dark mode.

I concur.

I'll work on a pull request for changing top-level category names, and moving Miscellaneous into Guides.

asmeurer:
I've made the suggested changes to link colors.

bertiewooster:
@asmeurer thanks for the changing the link colors so they're all dark, regardless of whether they've been visited or not.

I ran [Lighthouse](https://googlechrome.github.io/lighthouse/viewer/?psiurl=https%3A%2F%2Fwww.asmeurer.com%2Fsympy-furo-demo%2Fdev%2Fguides%2Fcontributing%2Fdocumentation-style-guide.html&strategy=desktop&category=performance&category=accessibility&category=best-practices&category=seo&category=pwa&utm_source=lh-chrome-ext) again and found a couple issues:

- The only remaining accessibility contrast issue I found is the bluish text in a code block, for example the class name and docstring (below, "loggamma" and "r""" The loggamma function...". My naive suggestion would be to decrease the saturation (intensity of color) of the code-block background, that is ["tint"](https://en.wikipedia.org/wiki/Tint%2C_shade_and_tone) it. (Personally, I find blocks of intense color distracting anyway.)
Screen Shot 2022-03-25 at 10 08 44 PM
- In terms of search engine optimization (SEO), one thing the pages are missing is a meta description: "Meta descriptions may be included in search results to concisely summarize page content." I learned that [Sphinx can support meta descriptions](https://docs.readthedocs.io/en/stable/guides/technical-docs-seo-guide.html) using reStructuredText like

```
.. meta::
    :description lang=en:
```

so I'm guessing the issue is those descriptions aren't auto-generated? I suppose we could manually generate descriptions for the non-API pages...

asmeurer:
> The only remaining accessibility contrast issue I found is the bluish text in a code block, for example the class name and docstring (below, "loggamma" and "r""" The loggamma function...". My naive suggestion would be to decrease the saturation (intensity of color) of the code-block background, that is ["tint"](https://en.wikipedia.org/wiki/Tint%2C_shade_and_tone) it. (Personally, I find blocks of intense color distracting, so this is my favored solution. :-) )

I already have decreased intensity of the code background color (in light mode). It was originally the same color as it is in the current site. 

I had also adjusted some of the code highlighting colors in dark mode to be better contrast. But I'm wondering if instead we ought to adjust the pygments theme, which is separate from the docs theme. Right now it's set to "sphinx" which is designed around the original docs theme, and probably wasn't designed with contrast in mind as much as a more modern one would be. It's actually the same theme used by the official Python docs (see e.g., https://docs.python.org/3/tutorial/introduction.html). It just happens to work well with the green background, so we never changed it. 

I'll see what the other options are. In either case, I'm realizing now that adjusting the syntax highlighting via CSS is the wrong approach. It would be better to adjust the pygments theme itself. 

I will say that I do like having the code blocks stand out somewhat. The doctest examples are the most important part of the docstring, so having them be prominent is a good thing. 

> In terms of search engine optimization (SEO), one thing the pages are missing is a meta description: "Meta descriptions may be included in search results to concisely summarize page content." I learned that [Sphinx can support meta descriptions](https://docs.readthedocs.io/en/stable/guides/technical-docs-seo-guide.html) using reStructuredText like

Does this need to be added to each page manually, or is the meta description something that can be generated automatically? 

I was considering enabling https://github.com/wpilibsuite/sphinxext-opengraph. I'm not sure if this is related to this or if it's something else. 

Another thing I noticed that Furo is missing compared to the current site is the `rel="canonical"` tag (see https://github.com/sympy/sympy/blob/master/doc/src/_templates/layout.html), although to be honest, it wasn't clear to me that that ever actually did much to the Google results.

bertiewooster:
> The [Open Graph protocol](https://ogp.me/) enables any web page to become a rich object in a social graph. For instance, this is used on Facebook to allow any web page to have the same functionality as any other object on Facebook.
>[Basic Metadata](https://ogp.me/#metadata)
`<meta property="og:title" content="The Rock" />`

My guess is that these meta properties do not apply to search engines; this is the first I've learned about it, so I could be wrong.

> > In terms of search engine optimization (SEO), one thing the pages are missing is a meta description: "Meta descriptions may be included in search results to concisely summarize page content." I learned that [Sphinx can support meta descriptions](https://docs.readthedocs.io/en/stable/guides/technical-docs-seo-guide.html) using reStructuredText like
> 
> Does this need to be added to each page manually, or is the meta description something that can be generated automatically?

From [Read The Docs](https://docs.readthedocs.io/en/stable/guides/technical-docs-seo-guide.html#use-meta-tags):
>Typically search engines will use the first few sentences of a page if no meta description is provided.

So if we don't supply anything, search engines will extract some text from the page. As these Google search results for "sympy solve" demonstrate, that may not be the most helpful text for a searcher to decide if the page is relevant:
Screen Shot 2022-03-28 at 5 46 59 PM

Here's the guidance for a meta description:
> - Your meta description should have the most relevant content of the page. A searcher should know whether they’ve found the right page from the description.
> - The meta description should be between 150-300 characters and it may be truncated down to around 150 characters in some situations.
> - Meta descriptions are used for display but not for ranking.

So to improve on what's automatically extracted, I think the descriptions would need to be generated manually. Having said that, arguably the meta description should also be the first (or nearly) text that the visitor encounters, at the top of the page. For example, on our [Matrices page](https://docs.sympy.org/dev/modules/matrices/matrices.html), the first text the visitor encounters is:

> ## Matrices (linear algebra)
> ### Creating Matrices
> The linear algebra module is designed to be as simple as possible. First, we import and declare our first Matrix object:

It would definitely benefit from a better overview of the topic, which could be the same as the meta description.

So perhaps we should update the documentation style guide to include a 1-2 sentence overview for each page, which would also be used as the meta description. I haven't investigated whether there's some way to automate tagging that overview as the meta description.

asmeurer:
The search engine thing is specifically the rel="canonical". See https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls. It theoretically would make it so that Google would prefer "latest" URLs over "dev" URLs. It also should make it so that Google doesn't try to serve versions of the docs that people have built and published themselves (for instance, my demo site). In practice, it's not perfect in my experience. We used to serve many old versions of the docs, and Google would generally serve up a random old URL rather than the latest one, even with the rel="canonical" (see https://github.com/sympy/sympy_doc/pull/33). Even so, if it does help even a little bit it's worth having there. 

Apparently rel="canonical" is supported directly in Sphinx https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_baseurl. So I've added that, although I think it doesn't actually work currently because of a bug in Furo https://github.com/pradyunsg/furo/discussions/406.

For the opengraph and meta description stuff, let's do it in a separate pull request. It's basically unrelated to the theme, and doesn't exist on the current site so it's not a regression. 

For the code highlighting, it looks like the "sphinx" style we have been using is basically the [friendly](https://pygments.org/styles/) pygments style with a [yellow background](https://github.com/sphinx-doc/sphinx/blob/b12b39db05d605cfb946f84afd27e0f6cf3c9e13/sphinx/pygments_styles.py). 

For dark mode, Furo is defaulting to monokai. The [Pygments styles page](https://pygments.org/styles/) says that both friendly and monokai are lower contrast. 

I'm not really sure what I want to do for the syntax highlighting. I'm not really keen on building a custom syntax highlighting theme just for the SymPy docs. So I would suggest trying to find an existing theme that works well. Or just using the Furo defaults. I'm not sure if pygments/sphinx would accept upstream improvements to improve the contrast of their themes, or if they would consider changing the theme colors to be a backwards incompatibility break.

asmeurer:
For the syntax highlighting, I found this https://github.com/mpchadwick/pygments-high-contrast-stylesheets, which has a tool that allows updating pygments styles to be WCAG AA compliant. I used this to create versions of the two themes that have better contrast. I outlined the process I used in [this gist](https://gist.github.com/asmeurer/d78409ea033aeeb764a11bb4acaeccc2). It's a bit hacky, and ideally the whole thing would just be one Python script, but since I don't plan on doing this again, I am OK with the hacky process.

This also reverts back to the original darker yellow background for the light mode, since the lighter yellow didn't even work for better contrast (it actually made things like the blue docstrings worse), and I like the darker yellow anyway (it highlights the code blocks more, and it fits well with the green theme). 

We might want to try upstreaming these changes, although I'm unsure if pygments or sphinx would accept color changes to their themes or if they would consider that to be backwards incompatible. @pradyunsg you may also be interested in this if you want Furo to have better WCAG compatibility. In particular, the default style you use for dark mode, "native", has some color contrast issues that are fixed here. I didn't check if the Furo default style for light mode has any such issues. If you're interested in upstreaming any of the other custom.css changes I have here or the versions menu, I would be open to that as well.

I think this should be ready to go now. The rel="canonical" thing will need to be fixed upstream in Furo, and the meta descriptions and opengraph stuff can be done separately, as can the modifications to the text and the headings layout.

bertiewooster:
Thanks @asmeurer. Is the version posted at https://www.asmeurer.com/sympy-furo-demo/dev/ the latest version?

asmeurer:
Yes it is.

pradyunsg:
You likely want to override `--color-sidebar-background-border` as well.

pradyunsg:
> And I would need to figure out how to make it cross-link with Javascript since it doesn't seem to be Jinja templated.

You can use `{% block announcement %}` in page.html, to provide the internal content of the announcement, if that's what you want -- as long as you set `announcement` in `html_theme_options` to a truthy value.

I should probaby document how that works.

doc/src/_static/custom.css
```diff
@@ -0,0 +1,116 @@
+/* Completely disable blockquotes. The Furo defaults make block quotes very*/
+/* apparent. However, the parameter blocks from numpydoc render as block
+ * quotes, which looks bad. Furthermore, it is very easy for things to
+ * accidentally have block quotes due to the fact that RST makes indented
+ * blocks render as a block quote (unlike Markdown which makes indented blocks
+ * render as code). There aren't any places in the documentation that use a
```
pradyunsg, line 6: Well, this is your documentation, so do whateverr you want -- I'll note that I made blockquotes apparent *because* it was too easy to inject them by accident with reST, making them unusable when you *actually* want them. :)

doc/src/_static/custom.css
```diff
@@ -0,0 +1,116 @@
...
+   color as the icon which is illegible on the green background. */
+.highlight-link a {
+    color: white !important;
+}
+
+/* Make the "class", "property" and so on before function names in the API*/
+/* docs bold. */
+em.property:first-child {
+    font-weight: bold;
+}
```
pradyunsg, line 68: The entire signature is bold by default, so you shouldn't need this.

doc/src/_templates/sidebar/versions.html
```diff
@@ -0,0 +1,56 @@
+<!-- Versions Menu. In order to work, there should be a file called versions.json one level up from the documentation listing the "dev" and "latest" version numbers, like
+{
+    "latest": "1.10",
+    "dev": "1.11.dev"
+}
+-->
+<div class="sidebar-tree sidebar-versions">
+  <h5 class="versions-header">Documentation Version</h4>
```
pradyunsg, line 8: I would suggest adding a padding-left to this, to align it with the rest of the sidebar content. You might want to consider if you want this to look like sidebar captions instead of a heading.

doc/src/conf.py
```diff
@@ -142,12 +146,96 @@
...
+
+    "code-font-size": "var(--font-size--small)",
+    "font-stack--monospace": 'DejaVu Sans Mono,"SFMono-Regular",Menlo,Consolas,Monaco,Liberation Mono,Lucida Console,monospace;'
+    }
+
+html_theme_options = {
+    "light_css_variables": common_theme_variables,
+    # The dark variables automatically inherit values from the light variables
+    "dark_css_variables": {
+        **common_theme_variables,
```
pradyunsg, line 206: You likely do not need this.

doc/src/conf.py
```diff
@@ -220,6 +308,7 @@
 # SymPy logo on title page
 html_logo = '_static/sympylogo.png'
 latex_logo = '_static/sympylogo_big.png'
+html_favicon = '../_build/logo/sympy-notailtext-favicon.ico'
```
pradyunsg, line 311: This should likely be logo/... instead? Otherwise, you won't get the logo on the first build (unless I'm misunderstanding your build process).

pradyunsg:
> We might want to try upstreaming these changes, although I'm unsure if pygments or sphinx would accept color changes to their themes or if they would consider that to be backwards incompatible. @pradyunsg you may also be interested in this if you want Furo to have better WCAG compatibility.

If pygments gets new better-contrast themes or improves their themes, I'd be *very* happy to see that. I had considered writing custom pygments theme that had a better contrast story, but I decided against that to minimise how much stuff I was doing in the theme; to instead try contributing to those project.

I haven't actually had the time/mental space to come around to doing that though, so I welcome this! :)

doc/src/_static/custom.css
asmeurer: I could make an effort to remove all the blockquotes, and maybe I should. This is a much easier thing for now. A big problem is that numpydoc is using blockquote for the parameters, which makes them look pretty bad. So we'd need to fix that (we're actually currently using a forked version of numpydoc, which could be part of the problem, but that just highlights the complexity). SymPy has a lot of documentation (over 270 RST files, most of which are pulling in documentation from docstrings). So any kind of content clean up like this would take a bit of effort, which is more than what I want to do for this PR.

doc/src/_static/custom.css
asmeurer: Looks like you are right. This must have been left over from a previous iteration.

doc/src/_templates/sidebar/versions.html
asmeurer: I'll have to play with it some more. I'm OK with the way it currently looks but I'll see if it looks better this way. Note that I'm not a designer and my HTML/Javascript/CSS skills are terrible, so if you see something that looks off, it's most likely unintentional so please let me know.

doc/src/conf.py
asmeurer: If I comment this line out, the light specific modifications do not carry over to dark mode. I think it was working before when I had an older version of Furo but it stopped working when I updated it.

doc/src/conf.py
asmeurer: Yeah, we compile the logo from svg to png and ico as part of the build process (the Makefile dependencies ensure it's compiled first).

asmeurer:
Thanks for the review @pradyunsg.

> You likely want to override --color-sidebar-background-border as well.

Wow that's a very subtle thing, especially with my other overrides. I guess the best thing to set this to when the sidebar has a different background color as in our case is to `var(--color-background-primary)`.

> You can use {% block announcement %} in page.html, to provide the internal content of the announcement, if that's what you want -- as long as you set announcement in html_theme_options to a truthy value.

I don't think we have any announcements we want to add at this point, but that's good to know for the future. We could consider adding one for the dev pages, but I still unsure if it's a good idea.

> If pygments gets new better-contrast themes or improves their themes, I'd be very happy to see that. I had considered writing custom pygments theme that had a better contrast story, but I decided against that to minimise how much stuff I was doing in the theme; to instead try contributing to those projects instead.

I more or less feel the same way. I didn't want to spend a whole lot of time on this, as I know it could end up being a huge rabbit hole that would ultimately produce a lot of good work, but not end up being related to the SymPy documentation. That's why I came up with a pretty hacky solution that just reuses the Ruby script I found at https://github.com/mpchadwick/pygments-high-contrast-stylesheets/blob/gh-pages/tools/make-stylesheet. It gets the job done for us, and makse the SymPy docs WCAG compatible.

Ideally the whole thing would be done in Python (I'm not sure what the best library for that would be), and and be more convenient to use. For instance, one could have a function that just takes a pygments theme and automatically makes it WCAG AA compatible without requiring any manual. Or alternately, just change the themes upstream to be AA compatible. But given that pygments [specifically notates](https://pygments.org/styles/) which styles have bad contrast gives me the impression that simply updating them is not a simple thing for them to do. The "sphinx" style is actually part of Sphinx, so updating that would be a separate discussion (but again, likely not a simple proposition, as it would affect a lot of projects, including docs.python.org itself).

So for now I'm going to leave this as an open project for someone else to implement, but if you or anyone else ever does, please let me know.

doc/src/conf.py
pradyunsg: Hmm... Fascinating. I'll dig into what's happening here later. I'm pretty sure this won't do the wrong thing though, so this is definitely fine to do anyway. :)

bertiewooster:
The [Lighthouse report](https://googlechrome.github.io/lighthouse/viewer/?psiurl=https%3A%2F%2Fwww.asmeurer.com%2Fsympy-furo-demo%2Fdev%2Fmodules%2Fntheory.html&strategy=desktop&category=performance&category=accessibility&category=best-practices&category=seo&category=pwa&utm_source=lh-chrome-ext) reveals an accessibility contrast issue with class etc. green names against a gray background, whether the green text is [regular](https://www.asmeurer.com/sympy-furo-demo/dev/modules/ntheory.html#sympy.ntheory.generate.Sieve) or [bold](https://www.asmeurer.com/sympy-furo-demo/dev/modules/ntheory.html#sympy.ntheory.generate.Sieve.extend):

sympy-docs-contrast-issue-gray-background

That situation seems to be the only contrast issue; the code blocks are fine.

asmeurer:
I fixed that and cleaned up some of the colors a bit (we now just have the two greens in color-brand-content and color-brand-primary). Also fixed some contrast issues that only showed when selecting a header (lighthouse does not see these unless you give a URL with an anchor). I also changed the hover color in the left sidebar because it had bad contrast (lighthouse does not look at hover colors).

doc/src/_templates/sidebar/versions.html
asmeurer: Actually this was supposed to have the exact same `padding` as the rest of the sidebar, but my editor inserted an incorrect line break in the CSS and messed it up. Should look better now.

bertiewooster:
Great, thanks for going beyond even the automated Lighthouse checks to ensure that as many people as possible can use SymPy documentation.

Comparing the current version (above) to the restyled version (below) demonstrates how much cleaner, more navigable, and more functional the restyled site will be:

sympy-current-example
sympy-furo-example
.

asmeurer:
If there are no further comments, I'd like to merge this. If we find something later that I've missed, we can make additional PRs to fix. It will be good to have this in the dev docs soon so that we can find any such issues before it gets released to the main "latest" documentation. 

Let's open new issues for the items we discussed above that were not addressed in this PR.

oscarbenjamin:
Looks good to me. Let's get this in.

asmeurer:
Looks like I forgot to update the dependencies for Travis. Fix forthcoming.

asmeurer:
I opened several issues for things that were discussed here but not addressed (see the cross references above this comment). If I missed anything, please open an issue for it.

PR #23160, opened 2022-02-24
codegen: Make List hashable

ThePauliPrinciple:
`List` had `__eq__` defined but not `__hash__` and was for that
reason not hashable. Tests are added.

Note that the codegen `List` is actually a `Tuple`, but allows for printing objects as `list` with the codegen AST, this is why it can be made hashable (which it should be since it is a subclass of `Basic`).

#### Release Notes

NO ENTRY


oscargus:
LGTM!

PR #23162, opened 2022-02-24
core/function: Fix regression in derivative

ThePauliPrinciple:
The arguments are ran through `_sympify` so that types registered
in `converter` are also allowed. This fixes a regression from https://github.com/sympy/sympy/commit/626064e1a582d6dc37fc3bcf54e3508d38fe6877 reported in #23144 and #23145.

https://github.com/ThePauliPrinciple/sympy/runs/5318939986?check_suite_focus=true
This PR passes `src/sage/symbolic/expression_conversions.py`.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/tests/test_function.py
+++ b/sympy/core/tests/test_function.py
@@ -777,6 +777,14 @@ def test_diff_wrt_not_allowed():
     assert diff(exp(x*y), x*y, 0) == exp(x*y)
 
 
+def test_diff_wrt_intlike():
+    class Two:
+        def __int__(self):
+            return 2
+
+    assert cos(x).diff(x, Two()) == -cos(x)
+
+
 def test_klein_gordon_lagrangian():
     m = Symbol('m')
     phi = f(x, t)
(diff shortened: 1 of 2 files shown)


sympy/core/tests/test_function.py
```diff
@@ -777,6 +777,14 @@ def test_diff_wrt_not_allowed():
     assert diff(exp(x*y), x*y, 0) == exp(x*y)
 
 
+def test_diff_wrt_intlike():
+    class Two:
+        def __int__(self):
```
ThePauliPrinciple, line 782: Do we want to test for a class with `_sympy_` too? How about a class added to `converter` ?

oscarbenjamin: Yes, maybe. At least this particular test would have picked up on this particular case. We also have the SAGE integration tests and that's the only known application for this but I thought it would be good to add something in the main test suite.

ThePauliPrinciple: Fair enough, I guess we can't put in tests for every possible situation.

Issue #23140, opened 2022-02-21
solve() returns overly complex results for simple linear equations


moorepants:
I can develop a work around for this, but I think the behavior should be what we see prior to SymPy 1.7. We should try to return the simpler, clearer result for a linear solve.

master:

```python
In [1]: import sympy as sm

In [2]: import sympy.physics.mechanics as me

In [3]: q1, q2, q3, u1, u2, u3 = me.dynamicsymbols('q1, q2, q3, u1, u2, u3')

In [4]: eqs = me.kinematic_equations([u1, u2, u3], [q1, q2, q3], 'body', 'ZXZ')

In [5]: eqs
Out[5]: 
[-(u1(t)*sin(q3(t)) + u2(t)*cos(q3(t)))/sin(q2(t)) + Derivative(q1(t), t),
 -u1(t)*cos(q3(t)) + u2(t)*sin(q3(t)) + Derivative(q2(t), t),
 (u1(t)*sin(q3(t)) + u2(t)*cos(q3(t)))*cos(q2(t))/sin(q2(t)) - u3(t) + Derivative(q3(t), t)]

In [6]: sm.solve(eqs, [u1, u2, u3])
Out[6]: 
{u1(t): sin(q2(t))*sin(q3(t))*Derivative(q1(t), t)/(sin(q3(t))**2 + cos(q3(t))**2) + cos(q3(t))*Derivative(q2(t), t)/(sin(q3(t))**2 + cos(q3(t))**2),
 u2(t): sin(q2(t))*cos(q3(t))*Derivative(q1(t), t)/(sin(q3(t))**2 + cos(q3(t))**2) - sin(q3(t))*Derivative(q2(t), t)/(sin(q3(t))**2 + cos(q3(t))**2),
 u3(t): cos(q2(t))*Derivative(q1(t), t) + Derivative(q3(t), t)}
```

sympy 1.6:

```python
In [1]: import sympy as sm

In [2]: import sympy.physics.mechanics as me

In [3]: q1, q2, q3, u1, u2, u3 = me.dynamicsymbols('q1, q2, q3, u1, u2, u3')

In [4]: eqs = me.kinematic_equations([u1, u2, u3], [q1, q2, q3], 'body', 'ZXZ')

In [5]: sm.solve(eqs, [u1, u2, u3])
Out[5]: 
{u1(t): sin(q2(t))*sin(q3(t))*Derivative(q1(t), t) + cos(q3(t))*Derivative(q2(t), t),
[...]

PR #23171, opened 2022-02-26
Restores pre-1.7 simplify behavior for orient_body_fixed()

### LINKED ISSUE #23130: ang_vel_in() could give a simpler result for body fixed angular velocities
If you set up a body fixed rotation and then call `.ang_vel_in()` I would expect the simplest form of the angular velocity to be returned, because we know apriori what the angular velocity is for any body fixed rotation.
```python
import sympy as sm
import sympy.physics.mechanics as me

psi, theta, phi = me.dynamicsymbols('psi, theta, varphi')

A = me.ReferenceFrame('A')
B = me.ReferenceFrame('B')
B.orient_body_fixed(A, (psi, theta, phi), 'ZXZ')
B.ang_vel_in(A)
```
gives:
```
⎛                         d                              d            ⎞       ⎛                         d                              d
⎜sin(θ(t))⋅sin(varphi(t))⋅──(ψ(t))        cos(varphi(t))⋅──(θ(t))     ⎟       ⎜sin(θ(t))⋅cos(varphi(t))⋅──(ψ(t))        sin(varphi(t))⋅──(
⎜                         dt                             dt           ⎟       ⎜                         dt                             dt
⎜───────────────────────────────── + ─────────────────────────────────⎟ b_x + ⎜───────────────────────────────── - ───────────────────────
⎜   2                 2                 2                 2           ⎟       ⎜   2                 2                 2                 2
⎝sin (varphi(t)) + cos (varphi(t))   sin (varphi(t)) + cos (varphi(t))⎠       ⎝sin (varphi(t)) + cos (varphi(t))   sin (varphi(t)) + cos (

          ⎞
θ(t))     ⎟
          ⎟       ⎛          d          d            ⎞
──────────⎟ b_y + ⎜cos(θ(t))⋅──(ψ(t)) + ──(varphi(t))⎟ b_z
          ⎟       ⎝          dt         dt           ⎠
varphi(t))⎠


```
We know the answer to this and I though that the code was supposed to return the simpler form for a lookup table. The simple form is:
```python
>>> B.ang_vel_in(A).simplify()
⎛                         d                         d       ⎞       ⎛                         d                         d       ⎞       ⎛
⎜sin(θ(t))⋅sin(varphi(t))⋅──(ψ(t)) + cos(varphi(t))⋅──(θ(t))⎟ b_x + ⎜sin(θ(t))⋅cos(varphi(t))⋅──(ψ(t)) - sin(varphi(t))⋅──(θ(t))⎟ b_y + ⎜c
⎝                         dt                        dt      ⎠       ⎝                         dt                        dt      ⎠       ⎝

         d          d            ⎞
os(θ(t))⋅──(ψ(t)) + ──(varphi(t))⎟ b_z
         dt         dt           ⎠


```
The manual calculation is:
```python
mnx = me.dot(B.y.express(A).dt(A), B.z)
mny = me.dot(B.z.express(A).dt(A), B.x)
mnz = me.dot(B.x.express(A).dt(A), B.y)
A_w_B = mnx*B.x + mny*B.y + mnz*B.z
A_w_B.simplify()
```
We have this function:
https://github.com/sympy/sympy/blob/master/sympy/physics/vector/functions.py#L227
which has this big lookup table but I don't think it is used.

moorepants:

Fixes #23130

`solve()` used `simplify()` internally when solving linear equations prior to SymPy 1.7 (see https://github.com/sympy/sympy/issues/23140). `ReferenceFrame.orient()` (and now `ReferenceFrame.orient_body_fixed()`) relied on this behavior to return the simplest statements for the angular velocities. This restores that behavior.

#### Release Notes

* physics.vector
  * Restored pre-1.7 simplest form for body fixed angular velocities.

### DIFF
--- a/sympy/physics/vector/frame.py
+++ b/sympy/physics/vector/frame.py
@@ -892,6 +892,13 @@ class ReferenceFrame:
             u2 = expand(td[u2])
             u3 = expand(td[u3])
             wvec = u1 * self.x + u2 * self.y + u3 * self.z
+            # NOTE : SymPy 1.7 removed the call to simplify() that occured
+            # inside the solve() function, so this restores the pre-1.7
+            # behavior. See:
+            # https://github.com/sympy/sympy/issues/23140
+            # and
+            # https://github.com/sympy/sympy/issues/23130
+            wvec = wvec.simplify()
         except (CoercionFailed, AssertionError):
             wvec = self._w_diff_dcm(parent)
         self._ang_vel_dict.update({parent: wvec})
--- a/sympy/physics/vector/tests/test_frame.py
+++ b/sympy/physics/vector/tests/test_frame.py
@@ -477,21 +477,22 @@ def test_orient_body():
 def test_orient_body_simple_ang_vel():
     """orient_body_fixed() uses kinematic_equations() internally and solves
     those equations for the measure numbers of the angular velocity. This test
-    ensures that the simplest form of that linear system solution is
-    returned."""
+    ensures that the simplest form of that linear system solution is returned,
+    thus the == for the expression comparison."""
 
     psi, theta, phi = dynamicsymbols('psi, theta, varphi')
+    t = dynamicsymbols._t
     A = ReferenceFrame('A')
     B = ReferenceFrame('B')
     B.orient_body_fixed(A, (psi, theta, phi), 'ZXZ')
     A_w_B = B.ang_vel_in(A)
-    mx = A_w_B.args[0][0][0]
-    my = A_w_B.args[0][0][1]
-    mz = A_w_B.args[0][0][2]
     assert A_w_B.args[0][1] == B
-    assert mx.count_ops() <= 15
-    assert my.count_ops() <= 15
-    assert mz.count_ops() <= 9
+    assert A_w_B.args[0][0][0] == (sin(theta)*sin(phi)*psi.diff(t) +
+                                   cos(phi)*theta.diff(t))
+    assert A_w_B.args[0][0][1] == (sin(theta)*cos(phi)*psi.diff(t) -
+                                   sin(phi)*theta.diff(t))
+    assert A_w_B.args[0][0][2] == cos(theta)*psi.diff(t) + phi.diff(t)
+
 
 def test_orient_space():
     A = ReferenceFrame('A')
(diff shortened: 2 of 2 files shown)


oscarbenjamin:
This would have been picked up before if there had been tests covering this case so maybe it's good to add some.

oscarbenjamin:
These two statements seem contradictory:

> I thought we aren't really supposed to test exact forms
> ...
> But the form does matter here

We should test everything that matters. Sometimes tests will need to be changed because of an unimportant change in form. I think that's fine and it's good to see that change in the diff for a PR so you can see that the form has changed and consider whether it represents something wider than the particular tested examples. In some cases the tested expressions would be too complicated though so that testing an exact form is a bad idea because the precise form has no value and it would be hard to investigate the meaning of any change in form if the test began to fail.

smichr:
> If I add the test, should I test the `srepr()` for an exact string match?

I would recommend using `count_ops` and testing that it is not too much greater than current acceptable op count. You can apply `count_ops` to an iterable (if needed).

oscarbenjamin:
> I would recommend using `count_ops`

I don't think that `count_ops` is needed here. The test case for the original issue was simple enough to be explicitly in the tests.

moorepants:
I went with the `count_ops()` suggestion as it felt wrong to test an `srepr()` exact string match.

oscarbenjamin:
The expression involved here is not that complicated i.e.:
```python
psip = psi.diff(t)
phip = phi.diff(t)
thetap = theta.diff(t)

expected = (
    (sin(theta)*sin(phi)*psip + cos(phi)*thetap) * B.x
  + (sin(theta)*cos(phi)*psip - sin(phi)*thetap) * B.y
  + (cos(theta)*psip + phip) * B.z
)
```
If that's the expected output then I'm not sure why you wouldn't test it directly.

moorepants:
As far as I understand, the only way to check is to do something like `assert simplify(expected - result)) == 0` so that would not check for an exact symbolic match.

moorepants:
How do I check for an exact expression match without using `srepr()`?

ThePauliPrinciple:
`==` checks for structural equality, is that not the right comparison here?

moorepants:
I actually don't know. I've never used `==` for assertions for expression in sympy, always avoid it.

ThePauliPrinciple:
It will compare `_hashable_content` which should the the `type` of the object together with its `args` (which will then happen recursively because `args` are also subclasses of `Basic`). Leave nodes (`Atom`s) will contain python objects in their `_hashable_content`.

moorepants:
Ok, this should be good to merge.

PR #23175, opened 2022-02-27
Improve `gf_edf_zassenhaus()`.

### LINKED ISSUE #23174: Problem with `gf_edf_zassenhaus()`
This:

```python
from sympy import Poly, cyclotomic_poly
    f = Poly(cyclotomic_poly(17), modulus=2)
    print(f.factor_list())
```

never terminates.

It gets stuck in `gf_edf_zassenhaus()`.

`gf_edf_zassenhaus()` is a probabilistic algorithm, but each iteration has at least a 1/2 chance of success. (And on my test case exactly 1/2 chance.)

On several test runs, I managed to get through 10 iterations without success. Each time, that's a 1 in 1024 chance. This can't be right.

-------------------

The existing code does seem to be a correct implementation of Alg 8.9 from [Geddes92], but something is clearly wrong.
My best guess is that maybe using a random polynomial `r` ignores the assumption `v(x) in W` of Thm 8.13? (Again in Geddes.)
I'm not sure.

Anyway, handling the case `p = 2` with Alg 3.4.8 from [Cohen93] instead seems to work. And it seems to promise to be much faster in general, since the test polynomial `h` is now computed with a loop of `n - 1` iterations, instead of `2**(n*N - 1) - 1`.

For example, in the case of `cyclotomic_poly(17)`, this is 7 iterations instead of 32,767.

skieffer:

Fixes #23174 

For the case p == 2, we use Algorithm 3.4.8 of [Cohen93], instead of the current procedure.

The current algorithm was failing to terminate on at least one known case (factoring cyclotomic_poly(17) mod 2).

#### Release Notes

* polys
  * Improve equal degree factorization mod powers of 2.

### DIFF
--- a/sympy/polys/galoistools.py
+++ b/sympy/polys/galoistools.py
@@ -1877,6 +1877,7 @@ def gf_edf_zassenhaus(f, n, p, K):
 
     .. [1] [Gathen99]_
     .. [2] [Geddes92]_
+    .. [3] [Cohen93]_
 
     """
     factors = [f]
@@ -1888,18 +1889,19 @@ def gf_edf_zassenhaus(f, n, p, K):
     if p != 2:
         b = gf_frobenius_monomial_base(f, p, K)
 
+    t = [K.one, K.zero]
     while len(factors) < N:
-        r = gf_random(2*n - 1, p, K)
-
         if p == 2:
-            h = r
+            h = r = t
 
-            for i in range(0, 2**(n*N - 1)):
+            for i in range(n - 1):
                 r = gf_pow_mod(r, 2, f, p, K)
                 h = gf_add(h, r, p, K)
 
             g = gf_gcd(f, h, p, K)
+            t += [K.zero, K.zero]
         else:
+            r = gf_random(2 * n - 1, p, K)
             h = _gf_pow_pnm1d2(r, n, f, b, p, K)
             g = gf_gcd(f, gf_sub_ground(h, K.one, p, K), p, K)
 
--- a/sympy/polys/galoistools.py
+++ b/sympy/polys/galoistools.py
@@ -1872,6 +1872,12 @@ def gf_edf_zassenhaus(f, n, p, K):
        >>> gf_edf_zassenhaus([1,1,1,1], 1, 5, ZZ)
        [[1, 1], [1, 2], [1, 3]]
 
+    Notes
+    =====
+
+    The case p == 2 is handled by Cohen's Algorithm 3.4.8. The case p odd is
+    as in Geddes Algorithm 8.9 (or Cohen's Algorithm 3.4.6).
+
     References
     ==========
 
--- a/sympy/polys/galoistools.py
+++ b/sympy/polys/galoistools.py
@@ -1876,8 +1876,8 @@ def gf_edf_zassenhaus(f, n, p, K):
     ==========
 
     .. [1] [Gathen99]_
-    .. [2] [Geddes92]_
-    .. [3] [Cohen93]_
+    .. [2] [Geddes92]_ Algorithm 8.9
+    .. [3] [Cohen93]_ Algorithm 3.4.8
 
     """
     factors = [f]
--- a/sympy/polys/tests/test_galoistools.py
+++ b/sympy/polys/tests/test_galoistools.py
@@ -620,6 +620,13 @@ def test_gf_edf():
     assert gf_edf_shoup(f, 2, 3, ZZ) == g
 
 
+def test_issue_23174():
+    f = ZZ.map([1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1])
+    g = ZZ.map([[1, 0, 0, 1, 1, 1, 0, 0, 1], [1, 1, 1, 0, 1, 0, 1, 1, 1]])
+
+    assert gf_edf_zassenhaus(f, 8, 2, ZZ) == g
+
+
 def test_gf_factor():
     assert gf_factor([], 11, ZZ) == (0, [])
     assert gf_factor([1], 11, ZZ) == (1, [])


sympy/polys/galoistools.py
```diff
@@ -1877,6 +1877,7 @@ def gf_edf_zassenhaus(f, n, p, K):
 
     .. [1] [Gathen99]_
     .. [2] [Geddes92]_
+    .. [3] [Cohen93]_
```
oscarbenjamin, line 1880: So which algorithm is implemented now? It would be good to add a note clarifying which of these references is the one that describes the implemented algorithm.

skieffer:
Wait a minute. I think in the Geddes algorithm `h` was only meant to have a linear number of terms as well. Maybe that's all that was wrong.

sympy/polys/galoistools.py
skieffer: I see your point. A more direct reference could be made.

skieffer:
Okay, only one line of the existing implementation actually needed to change:

https://github.com/sympy/sympy/blob/bb7275c103895b51ff3736b2c1e099a810a5c955/sympy/polys/galoistools.py#L1897

becomes

```python
for i in range(n - 1):
```

and then the tests pass. And `range(n - 1)` is confirmed by Cohen's Proposition 3.4.7.

But I think there were two errors. A correct transcription of Alg 8.9 as given in Geddes would have been

```python
for i in range(n*N - 1):
```

but even Alg 8.9 appears to be wrong. With the above line, the new `test_issue_23174()` still hangs.

skieffer:
I still think moving to Cohen's version is a good idea though. Cohen states that while you _can_ use a random polynomial when `p == 2`, it is better to use the polynomials `x, x**3, x**5, x**7, ...`.

jksuom:
Yes, the exponents in the sum should be powers of 2: 1, 2, 4,...

skieffer:
I think this one's good to go if there's agreement on switching to Cohen's Alg 3.4.8 in the case of `p == 2`.

To be clear, all this means is that instead of choosing a random polynomial on each iteration, we start with `x`, then move to `x^3`, `x^5`, `x^7`, and so on, until it works, which it is guaranteed to do by `x^(2*n-1)` at latest.

A more conservative approach would be to make this a simple one-line bug fix. But if the new algorithm is an improvement, maybe it's worth taking the occasion to switch to it?

Cohen doesn't offer much explanation as to exactly _how_ it is better. He writes (p. 129) that in fact this approach can be extended to general primes `p`, but is "useful in practice only if `p` is small`. (Can't get much smaller than 2.)

He then points to Exercise 14, where you're meant to try an experiment and, "Compare in practice the expected speed of the resulting algorithm" with the more traditional one (i.e. what we already have).

So it sounds like something that's hard to prove, but experience shows it's better.

oscarbenjamin:
I'm not sure exactly what makes good test cases for the performance of this function. I used `random_poly` to find a polynomial of degree 200 that hits this code path (many examples skip out early in `gf_edf_zassenhaus`):
```python
from sympy import *

x = symbols('x')

p = (x**196 + x**195 + x**194 + x**193 + x**192 + x**191 + x**189 + x**188 +
        x**183 + x**178 + x**175 + x**168 + x**167 + x**166 + x**165 + x**164 +
        x**161 + x**160 + x**159 + x**157 + x**156 + x**155 + x**153 + x**151 +
        x**149 + x**148 + x**146 + x**145 + x**144 + x**143 + x**141 + x**140 +
        x**139 + x**138 + x**137 + x**136 + x**133 + x**130 + x**126 + x**124 +
        x**122 + x**120 + x**116 + x**115 + x**113 + x**107 + x**105 + x**104 +
        x**101 + x**100 + x**99 + x**98 + x**96 + x**91 + x**87 + x**86 + x**85
        + x**83 + x**79 + x**78 + x**77 + x**76 + x**75 + x**69 + x**68 + x**67
        + x**63 + x**62 + x**61 + x**60 + x**59 + x**57 + x**54 + x**53 + x**51
        + x**50 + x**48 + x**45 + x**43 + x**42 + x**39 + x**37 + x**33 + x**31
        + x**29 + x**28 + x**27 + x**26 + x**24 + x**20 + x**18 + x**17 + x**16
        + x**15 + x**14 + x**13 + x**12 + x**11 + x**8 + x**7 + x**6 + x**4 +
        x**2 + 1)
```
Then using isympy with master I get:
```python
In [2]: %time factor(p, modulus=2)
CPU times: user 2.35 s, sys: 0 ns, total: 2.35 s
Wall time: 2.35 s

In [3]: %timeit factor(p, modulus=2)
2.39 s ± 331 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
The PR gives:
```python
In [2]: %time factor(p, modulus=2)
CPU times: user 1.84 s, sys: 4 ms, total: 1.85 s
Wall time: 1.85 s

In [3]: %timeit factor(p, modulus=2)
1.78 s ± 3.92 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
So it seems to be a bit faster for this case.

jksuom:
> this one's good to go if there's agreement on switching to Cohen's Alg 3.4.8 in the case of p == 2.

Agreed. I would also prefer Cohen's deterministic algorithm in the case of p == 2 as it does not seem to be slower.

skieffer:
Okay, sounds like there's agreement.

I did try generating more random polys and running timing tests, and found the same results as @oscarbenjamin: the Cohen method seems to be slightly faster. Not a really significant improvement, but I would agree with @jksuom that deterministic + "not slower" is enough of a recommendation.

Thanks for reviewing!

oscarbenjamin:
I think it's hard to tell how slow `gf_edf_zassenhaus` is because it potentially only takes up a small amount of the time in `factor`. The previous version can be quite slow even for small inputs e.g.:
```python
In [12]: from gmpy2 import mpz
    ...: 
    ...: f = [mpz(1), mpz(1), mpz(1), mpz(1), mpz(0), mpz(1), mpz(1), mpz(0), mpz(1), mpz(1), mpz(1)]
    ...: 
    ...: n = 5
    ...: p = 2
    ...: K = ZZ

In [13]: for _ in range(10):
    ...:     %time gf_edf_zassenhaus(f, n, p, K)
    ...: 
CPU times: user 304 ms, sys: 0 ns, total: 304 ms
Wall time: 303 ms
CPU times: user 1.24 s, sys: 0 ns, total: 1.24 s
Wall time: 1.24 s
CPU times: user 1.88 s, sys: 0 ns, total: 1.88 s
Wall time: 1.89 s
CPU times: user 136 ms, sys: 0 ns, total: 136 ms
Wall time: 135 ms
CPU times: user 1.25 s, sys: 0 ns, total: 1.25 s
Wall time: 1.25 s
CPU times: user 420 ms, sys: 0 ns, total: 420 ms
Wall time: 420 ms
CPU times: user 388 ms, sys: 0 ns, total: 388 ms
Wall time: 387 ms
CPU times: user 220 ms, sys: 0 ns, total: 220 ms
Wall time: 220 ms
CPU times: user 1.12 s, sys: 0 ns, total: 1.12 s
Wall time: 1.12 s
CPU times: user 1.18 s, sys: 0 ns, total: 1.18 s
Wall time: 1.17 s
```
After merging this we get:
```python
In [1]: from gmpy2 import mpz
   ...: 
   ...: f = [mpz(1), mpz(1), mpz(1), mpz(1), mpz(0), mpz(1), mpz(1), mpz(0), mpz(1), mpz(1), mpz(1)]
   ...: n = 5
   ...: p = 2
   ...: K = ZZ

In [2]: from sympy.polys.galoistools import gf_edf_zassenhaus

In [3]: for _ in range(10):
   ...:     %time gf_edf_zassenhaus(f, n, p, K)
   ...: 
CPU times: user 0 ns, sys: 0 ns, total: 0 ns
Wall time: 849 µs
CPU times: user 4 ms, sys: 0 ns, total: 4 ms
Wall time: 744 µs
CPU times: user 0 ns, sys: 0 ns, total: 0 ns
Wall time: 769 µs
CPU times: user 0 ns, sys: 0 ns, total: 0 ns
Wall time: 781 µs
CPU times: user 0 ns, sys: 0 ns, total: 0 ns
Wall time: 859 µs
CPU times: user 0 ns, sys: 0 ns, total: 0 ns
Wall time: 1.23 ms
CPU times: user 4 ms, sys: 0 ns, total: 4 ms
Wall time: 905 µs
CPU times: user 0 ns, sys: 0 ns, total: 0 ns
Wall time: 705 µs
CPU times: user 0 ns, sys: 0 ns, total: 0 ns
Wall time: 613 µs
CPU times: user 0 ns, sys: 0 ns, total: 0 ns
Wall time: 610 µs
```
So I think this implementation is actually a lot faster but it still only takes up a small portion of the total time in `factor`:
```python
In [4]: x = symbols('x')
   ...: 
   ...: p = (x**196 + x**195 + x**194 + x**193 + x**192 + x**191 + x**189 + x**188 +
   ...:         x**183 + x**178 + x**175 + x**168 + x**167 + x**166 + x**165 + x**164 +
   ...:         x**161 + x**160 + x**159 + x**157 + x**156 + x**155 + x**153 + x**151 +
   ...:         x**149 + x**148 + x**146 + x**145 + x**144 + x**143 + x**141 + x**140 +
   ...:         x**139 + x**138 + x**137 + x**136 + x**133 + x**130 + x**126 + x**124 +
   ...:         x**122 + x**120 + x**116 + x**115 + x**113 + x**107 + x**105 + x**104 +
   ...:         x**101 + x**100 + x**99 + x**98 + x**96 + x**91 + x**87 + x**86 + x**85
   ...:         + x**83 + x**79 + x**78 + x**77 + x**76 + x**75 + x**69 + x**68 + x**67
   ...:         + x**63 + x**62 + x**61 + x**60 + x**59 + x**57 + x**54 + x**53 + x**51
   ...:         + x**50 + x**48 + x**45 + x**43 + x**42 + x**39 + x**37 + x**33 + x**31
   ...:         + x**29 + x**28 + x**27 + x**26 + x**24 + x**20 + x**18 + x**17 + x**16
   ...:         + x**15 + x**14 + x**13 + x**12 + x**11 + x**8 + x**7 + x**6 + x**4 +
   ...:         x**2 + 1)

In [5]: %time ok = factor(p, modulus=2)
CPU times: user 1.92 s, sys: 0 ns, total: 1.92 s
Wall time: 1.92 s
```
That difference in `gf_edf_zassenhaus` speed does make me wonder how many other things in there could be massively improved.

PR #23178, opened 2022-02-27
Some pep8 cleanup of sympy.physics.vector.

moorepants:

Simple pep8 code cleanup, fixing linting warnings.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/physics/vector/frame.py
+++ b/sympy/physics/vector/frame.py
@@ -66,8 +66,8 @@ class CoordinateSym(Symbol):
         return self._id[0]
 
     def __eq__(self, other):
-        #Check if the other object is a CoordinateSym of the same frame
-        #and same index
+        # Check if the other object is a CoordinateSym of the same frame and
+        # same index
         if isinstance(other, CoordinateSym):
             if other._id == self._id:
                 return True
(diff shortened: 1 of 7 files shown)


sympy/physics/vector/frame.py
```diff
@@ -130,7 +130,8 @@ def __init__(self, name, indices=None, latexs=None, variables=None):
         >>> vlatex(P.x)
         'A1'
 
-        symbols() can be used to create multiple Reference Frames in one step, for example:
+        ``symbols()`` can be used to create multiple Reference Frames in one
```
oscargus, line 133: Even better is probably `:func:~.symbols` (with correct backticks around `~.symbols`, but not sure how to get those here...)

sympy/physics/vector/point.py
```diff
@@ -32,7 +32,8 @@ class Point:
     >>> O.acc(N)
     u1'*N.x + u2'*N.y + u3'*N.z
 
-    symbols() can be used to create multiple Points in a single step, for example:
+    ``symbols()`` can be used to create multiple Points in a single step, for
```
oscargus, line 35: Same here. Possible also for Point, but maybe that looks a bit weird as the full path must be included (I think?).

oscargus:
Looks good! Some minor comment, but nothing really required.

sympy/physics/vector/frame.py
moorepants: I'll work on those in other PRs.

moorepants:
Thanks. I'll merge and work on the cross referencing later.

Issue #23164, opened 2022-02-25
Refine and Abs don't work for compact domains


aplund:
I want the variable a to satisfy 0 < a < 1 and then refine absolute value expressions. 

```
>>> from sympy import *
>>> a = Symbol('a', positive=True)
>>> sqrt((1-a)**2).refine(Q.lt(a,1))
Abs(a - 1)
>>> Abs(a-1).refine(Q.lt(a,1))
Abs(a - 1)
```

My expectation is that both of these would return 1-a.

```
>>> sympy.__version__
'1.9'
```

```
$ python -V
Python 3.10.2
```

oscargus:
Currently Abs does not have any refine functionality. Even after adding a primitive one (in #23180), it is clear that the refine system is not really that powerful (yet). So this example does not work.

oscargus:
This actually works:

```
sqrt((1-a)**2).refine(Q.negative(a-1))
Abs(a-1).refine(Q.negative(a-1))
```

(And I was wrong that there were no refine support for Abs.)

TiloRC:
Fixed by #25544

PR #23180, opened 2022-02-27
Add refine for Abs

oscargus:

Related to #23164

Add basic refine functionality for Abs.

#### Other comments

Things like `ask(x >= 0, x >= 1)` do not work, but once that is operational, this will probably sort out #23164. Also `Abs(x).refine(Q.ge(x, 1))` does not simplify.

#### Release Notes

* assumptions
   * Basic refine functionality for `Abs`.

### DIFF
--- a/sympy/functions/elementary/complexes.py
+++ b/sympy/functions/elementary/complexes.py
@@ -694,6 +694,19 @@ class Abs(Function):
     def _eval_rewrite_as_conjugate(self, arg, **kwargs):
         return sqrt(arg*conjugate(arg))
 
+    def _eval_refine(self, assumptions, **kwargs):
+        from sympy.assumptions.ask import ask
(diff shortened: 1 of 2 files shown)


sylee957:
Doesn't `sympy.refine.refine_abs` work or can be reused?

oscargus:
> Doesn't `sympy.refine.refine_abs` work or can be reused?

I've missed that. Just noted that the very basic refine in the added test didn't seem to work and when searching I couldn't find anything doing refine. But I clearly missed that.

oscargus:
Well, I guess one should really sort out the assumptions system support for various approaches. 

`Abs(x).refine(x > 0)` does not work with the current approach, but `Abs(x).refine(Q.positive(x))` does.

While the latter is OK for simple queries, if somethings like #23164 should work in the future, the first is probably required as well. 

I'll try to merge the approaches...

PR #22180, opened 2021-09-29
WIP: Add return type hints for some functions

MaddyGuthridge:
#### References to other Issues or PRs

Should begin to address #17945

#### Brief description of what is fixed or changed
Type hints have been added to a few functions in SymPy. This helps speed up peformance of language servers when SymPy is used within another project, as the server can determine return types from the hints rather than by analysing code. In VS Code's PyLance, I noted that load times for inline documentation for the changed SymPy functions decreased from several minutes to less than a second, and RAM usage by the language server decreased significantly too.

So far, I've only made changes for the `diff()`, `integrate()` and `sympify()` functions as a proof-of-concept, but if you guys think this is a good idea, I'll do some more work on it.

#### Other comments
This is my first PR to SymPy (and to any project that isn't my own), so it's probably far from perfect. I'm putting it up early to get feedback sooner rather than later. Let me know if I'm overlooking anything, or if I'm making any mistakes. I'm keen to learn!

#### Release Notes

NO ENTRY

PR #23181, opened 2022-02-28
Fix type inconsistency inside symbols

eagleoflqj:

Related to #22180

Add type hints to `symbols`.
It's **impossible** to add type hints to it **accurately**, since return type depends on whether there are `,` and ` ` in `names`.
However, in most use cases, we call it using string literal. (Imagine calling `res = symbols(s)` when `s` is a string var, then check whether `res` is a tuple or a `Symbol`.) So we know how many `Symbol`s we want.
If we want only one symbol, we can just write `x = Symbol('x')` or `x, = symbols('x,')` or even `x, = symbols('x', seq=True)`. There is really no need for writing `x = symbols('x')` in sympy code base. For downstream users, we can give them choice: either receive type hints from `symbols`, or still using `x = symbols('x')` without being broken and without correct type hints.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/symbol.py
+++ b/sympy/core/symbol.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from .assumptions import StdFactKB, _assume_defined
 from .basic import Basic, Atom
 from .cache import cacheit
@@ -17,6 +19,8 @@ import string
 import re as _re
 import random
 from itertools import product
+from typing import Any
+
 
 class Str(Atom):
     """
@@ -577,7 +581,8 @@ class Wild(Symbol):
 
 _range = _re.compile('([0-9]*:[0-9]+|[a-zA-Z]?:[a-zA-Z])')
 
-def symbols(names, *, cls=Symbol, **args):
+
+def symbols(names, *, cls=Symbol, **args) -> Any:
     r"""
     Transform strings into instances of :class:`Symbol` class.
 
@@ -697,16 +702,16 @@ def symbols(names, *, cls=Symbol, **args):
 
     if isinstance(names, str):
         marker = 0
-        literals = [r'\,', r'\:', r'\ ']
-        for i in range(len(literals)):
-            lit = literals.pop(0)
-            if lit in names:
+        splitters = r'\,', r'\:', r'\ '
+        literals: list[tuple[str, str]] = []
+        for splitter in splitters:
+            if splitter in names:
                 while chr(marker) in names:
                     marker += 1
                 lit_char = chr(marker)
                 marker += 1
-                names = names.replace(lit, lit_char)
-                literals.append((lit_char, lit[1:]))
+                names = names.replace(splitter, lit_char)
+                literals.append((lit_char, splitter[1:]))
         def literal(s):
             if literals:
                 for c, l in literals:
@@ -739,7 +744,8 @@ def symbols(names, *, cls=Symbol, **args):
                 result.append(symbol)
                 continue
 
-            split = _range.split(name)
+            split: list[str] = _range.split(name)
+            split_list: list[list[str]] = []
             # remove 1 layer of bounding parentheses around ranges
             for i in range(len(split) - 1):
                 if i and ':' in split[i] and split[i] != ':' and \
@@ -747,30 +753,30 @@ def symbols(names, *, cls=Symbol, **args):
                         split[i + 1].startswith(')'):
                     split[i - 1] = split[i - 1][:-1]
                     split[i + 1] = split[i + 1][1:]
-            for i, s in enumerate(split):
+            for s in split:
                 if ':' in s:
-                    if s[-1].endswith(':'):
+                    if s.endswith(':'):
                         raise ValueError('missing end range')
                     a, b = s.split(':')
                     if b[-1] in string.digits:
-                        a = 0 if not a else int(a)
-                        b = int(b)
-                        split[i] = [str(c) for c in range(a, b)]
+                        a_i = 0 if not a else int(a)
+                        b_i = int(b)
+                        split_list.append([str(c) for c in range(a_i, b_i)])
                     else:
                         a = a or 'a'
-                        split[i] = [string.ascii_letters[c] for c in range(
+                        split_list.append([string.ascii_letters[c] for c in range(
                             string.ascii_letters.index(a),
-                            string.ascii_letters.index(b) + 1)]  # inclusive
-                    if not split[i]:
+                            string.ascii_letters.index(b) + 1)])  # inclusive
+                    if not split_list[-1]:
                         break
                 else:
-                    split[i] = [s]
+                    split_list.append([s])
             else:
                 seq = True
-                if len(split) == 1:
-                    names = split[0]
+                if len(split_list) == 1:
+                    names = split_list[0]
                 else:
-                    names = [''.join(s) for s in product(*split)]
+                    names = [''.join(s) for s in product(*split_list)]
                 if literals:
                     result.extend([cls(literal(s), **args) for s in names])
                 else:
--- a/sympy/matrices/expressions/tests/test_derivatives.py
+++ b/sympy/matrices/expressions/tests/test_derivatives.py
@@ -27,8 +27,7 @@ from sympy.tensor.array.array_derivatives import ArrayDerivative
 from sympy.matrices.expressions import hadamard_power
 from sympy.tensor.array.expressions.array_expressions import ArrayAdd, ArrayTensorProduct, PermuteDims
 
-k = symbols("k")
-i, j = symbols("i j")
+i, j, k = symbols("i j k")
 m, n = symbols("m n")
 
 X = MatrixSymbol("X", k, k)
--- a/sympy/solvers/diophantine/diophantine.py
+++ b/sympy/solvers/diophantine/diophantine.py
@@ -550,7 +550,7 @@ class BinaryQuadratic(DiophantineEquationType):
                 sqc = isqrt(c)
                 _c = e*sqc*D - sqa*E
                 if not _c:
-                    z = symbols("z", real=True)
+                    z = Symbol("z", real=True)
                     eq = sqa*g*z**2 + D*z + sqa*F
                     roots = solveset_real(eq, z).intersect(S.Integers)
                     for root in roots:


sympy/matrices/expressions/tests/test_derivatives.py
```diff
@@ -27,7 +27,7 @@
 from sympy.matrices.expressions import hadamard_power
 from sympy.tensor.array.expressions.array_expressions import ArrayAdd, ArrayTensorProduct, PermuteDims
 
-k = symbols("k")
+k, = symbols("k,")
```
moorepants, line 30: Does this imply the former statement no longer works? if so, I don't think this test should be changed.

sympy/solvers/diophantine/diophantine.py
```diff
@@ -550,7 +550,7 @@ def solve(self, parameters=None, limit=None) -> DiophantineSolutionSet:
                 sqc = isqrt(c)
                 _c = e*sqc*D - sqa*E
                 if not _c:
-                    z = symbols("z", real=True)
+                    z = Symbol("z", real=True)
```
moorepants, line 553: Same here, why would this change be needed?

moorepants:
Hi, on the face of it, I'm a strong -1 to adding these typing statements to sympy. It clutters the code with incomprehensible typing declarations. I think we need a sympep for something like this before these PRs are merged.

sympy/matrices/expressions/tests/test_derivatives.py
eagleoflqj: No. The former statement still works at runtime. It just mean mypy complains about it.

eagleoflqj:
> It clutters the code with incomprehensible typing declarations.

That’s because this function’s functionality is too flexible. If that’s the only concern, I can remove overloads and let the return type be tuple only.

As #22180 said, type hints are not only for type checking but also for IDE auto completion which is very important for downstream users.

moorepants:
I moved my concerns over to the main issue. I don't have an issue with this specific PR or the implementation, but just want to get clarity on whether sympy is, in fact, moving to a typed code base. It's a big change for a 16 year old code base.

sympy/matrices/expressions/tests/test_derivatives.py
oscarbenjamin: I'm not sure about this. We need to be clear about whether using symbols like this is discouraged because otherwise mypy is going to complain about user code.

moorepants: I don't think we should have to write unusual forms of code like this to satisfy a type checker. Seems like the type checker is wrong in this case.

eagleoflqj: > I'm not sure about this. We need to be clear about whether using symbols like this is discouraged because otherwise mypy is going to complain about user code.

If a user uses mypy, it means the user definitely wants stronger type support and inevitably fix code according to type checker. Original `symbols` provides nothing.

sympy/solvers/diophantine/diophantine.py
eagleoflqj: Even if finally we decide not to add type hints to `symbols`, I still go for this change. Because
1. otherwise the following code involving `z` can not be type-checked.
2. a plural form function returns multiple results in common sense, but we only need one here
3. `symbols` has more overhead than `Symbol` as it calls `Symbol("z", real=True)` in the end

sympy/matrices/expressions/tests/test_derivatives.py
oscarbenjamin: I tried editing a file in vscode with pylance (which is the "recommended" plugin). If I write `x = symbols('x')` and then hover the mouse over `x` it says that `x` has type `Tuple[Symbol, ...]` which is incorrect.

eagleoflqj: > I tried editing a file in vscode with pylance (which is the "recommended" plugin). If I write `x = symbols('x')` and then hover the mouse over `x` it says that `x` has type `Tuple[Symbol, ...]` which is incorrect.

True. That's why we'd better not suggest writing `symbols('x')`. Is there a scene where it cannot be replaced by `Symbol('x')`?

sympy/solvers/diophantine/diophantine.py
oscargus: Look at the first example in the tutorial: https://docs.sympy.org/latest/tutorial/preliminaries.html#exercises

(Then, I personally always use `Symbol` for a single symbol, but it seems like the "official" SymPy way is, or has at some stage been, to use `symbols`. So it makes more sense, as @moorepants said somewhere, to update the documentation first.)

sympy/matrices/expressions/tests/test_derivatives.py
oscargus: Why can't we make symbols return a single element as well?

sympy/core/symbol.py
```diff
@@ -577,6 +578,15 @@ def matches(self, expr, repl_dict=None, old=False):
 
 _range = _re.compile('([0-9]*:[0-9]+|[a-zA-Z]?:[a-zA-Z])')
 
+@overload
```
oscargus, line 581: I think it is time to start using those pyi-files or whatever they were called. This looks just horrendous... (And does NOT simplify for a developer to read the code etc.)

sympy/matrices/expressions/tests/test_derivatives.py
oscarbenjamin: I think the problem is that a correct type hint here would look something like
```python
def symbols(...) -> Symbol | Tuple[Symbol]
```
but then when you do
```python
x = symbols('x')
```
the type checker would not be able to infer that x has type `Symbol`. Then something like `x.is_positive` would appear to be a type failure (tuple has no attribute `is_positive`). The problem is that `symbols('x')` and `symbols('x, y')` return different types but the type is determined by the *value* of the argument rather than the types of the arguments.

sympy/core/symbol.py
oscarbenjamin: I agree that this looks horrendous but maybe common sympy functions should not have such complicated type semantics in the first place. Given that the semantics actually are this complicated maybe it's good to make that clear with annotations.

Almost all use of `symbols` is just calling it with a single string and getting back either a single symbol or a tuple of symbols. I very rarely see anyone using any of the other overloads so I think it was a mistake to add them.

oscargus: And then it would "just" be:
`def symbols(names: str, *, cls=Symbol, **args) -> Union[Symbol, List[Symbol]]`
?

oscargus: Except that we should probably consistently rename `List` to `tList` as we have our own `List` class (and it will be a mess to remember which should be renamed and not, maybe have a renaming file in `utilities` so that one can import `tUnion` etc directly?).

oscarbenjamin: With PEP 585 you don't need `typing.Tuple` etc so as of Python 3.9 it's just:
```
Symbol | tuple[Symbol]
```
https://www.python.org/dev/peps/pep-0585/

oscarbenjamin: Actually the `|` syntax is Python 3.10:
https://www.python.org/dev/peps/pep-0604/

eagleoflqj:
I know this PR is very controversial. I was too ambitious, and there seem no way to add type hints to the signature of `symbols` elegantly. After trying many ways I decide to give up and only fix the internal type inconsistency inside `symbols` in this PR.

sympy/core/symbol.py
```diff
@@ -17,6 +17,8 @@
 import re as _re
 import random
 from itertools import product
+from typing import List, Tuple as tTuple, Any
```
oscargus, line 20: I think it is better to start using an alias for List as well since there is a List-class in SymPy. Even better to create a `utilities.typing` with these aliases and import from there.

sympy/core/symbol.py
```diff
@@ -697,16 +700,16 @@ def symbols(names, *, cls=Symbol, **args):
 
     if isinstance(names, str):
         marker = 0
-        literals = [r'\,', r'\:', r'\ ']
-        for i in range(len(literals)):
-            lit = literals.pop(0)
-            if lit in names:
+        splitters = r'\,', r'\:', r'\ '
+        literals: List[tTuple[str, str]] = []
```
oscargus, line 704: I will never be a big supporter of this. While I may see the point of functions having (well-)defined inputs and outputs, bloating the internals like this... (An end-of-line comment would be better from a readability point-of-view, if it is now a problem to understand how it works. Though I understand that there are possible advantages for static code checkers to use this.)

So let us just say that I do not like it, but will not block this.

oscargus: I like the remaining though!

sympy/solvers/diophantine/diophantine.py
oscargus: This should be reverted.

sympy/core/symbol.py
```diff
@@ -739,38 +742,39 @@ def literal(s):
...
+                    split_list.append([s])
             else:
                 seq = True
-                if len(split) == 1:
-                    names = split[0]
+                if len(split_list) == 1:
+                    names = split_list[0]
                 else:
-                    names = [''.join(s) for s in product(*split)]
+                    names = [''.join(s) for s in product(*split_list)]
```
oscargus, line 777: I do not follow the logic of the rewrites in this part (nor of the original code), but if it passes the tests that should be fine.

One may question if append is better or worse than updating a pre-allocated list, but I guess this is not a performance bottleneck anyway...

sympy/solvers/diophantine/diophantine.py
eagleoflqj: > This line should be reverted.

I gave the reason for this change. What’s the reason for reverting?

eagleoflqj:
```py
Python 3.9.5 (default, Jun  4 2021, 12:28:51) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.0.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from sympy import *

In [2]: %timeit Symbol('x', real=True)
863 ns ± 6.21 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [3]: %timeit symbols('x', real=True)
2.3 µs ± 6.05 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
Only for efficiency `symbols('x')` shouldn't be endorsed inside sympy project, while users of sympy may choose between higher efficiency and more flexible syntax.

eagleoflqj:
> Regarding the question about using `Symbol` vs. `symbols`, yes, `symbols` should be preferred in the documentation. This is because it is a single function that works for all use cases: a single symbol, multiple symbols, and functions. See https://github.com/sympy/sympy/wiki/Idioms-and-Antipatterns#creating-symbols.

Even if users are encouraged to use `symbols`, it doesn't mean `symbols` should be used inside sympy when you need only one `Symbol`. There is an obvious efficiency difference https://github.com/sympy/sympy/pull/23181#discussion_r817894732. Also sympy actually uses `Symbol` more often than `symbols` in one-symbol scene
```sh
$ git grep 'symbols("[^,: "]*"' **/*.py | grep -v test | grep -v '>>>' | wc -l
13
$ git grep 'Symbol("[^,: "]*"' **/*.py | grep -v test | grep -v '>>>' | wc -l
64
```
Tests and docstrings are filtered out.

sympy/core/symbol.py
eagleoflqj: I took a look at https://www.python.org/dev/peps/pep-0585/#id5
It seems better to directly use `list` with `from __future__ import annotation`, as `from typing import List` is deprecated.

asmeurer: Yes, let's do this. I don't like these `tTuple` things as they are confusing. This should be fixed everywhere in the codebase, though, so it can be done in a separate PR. I think we can also use the `|` syntax. It just isn't supported if you write it directly in Python  < 3.10 (i.e., without using `from __future__ import annotations), but mypy itself will support it in any Python version. `from __future__ import annotations` also means we can convert string and comment annotations into real annotations.

asmeurer:
> Even if users are encouraged to use symbols, it doesn't mean symbols should be used inside sympy when you need only one Symbol. There is an obvious efficiency difference https://github.com/sympy/sympy/pull/23181#discussion_r817894732. Also sympy actually uses Symbol more often than symbols in one-symbol scene

Yes, inside of library code, using Symbol is fine. The point is about documentation. 

Actually I think we decided at some point to use sympy.abc in all docstrings. I'll have to look up that conversation to see why that decision was made.

eagleoflqj:
@oscarbenjamin is it good to go?

sympy/solvers/diophantine/diophantine.py
asmeurer: I think it's fine to use Symbol in library code.

sympy/core/symbol.py
```diff
@@ -577,7 +581,8 @@ def matches(self, expr, repl_dict=None, old=False):
 
 _range = _re.compile('([0-9]*:[0-9]+|[a-zA-Z]?:[a-zA-Z])')
 
-def symbols(names, *, cls=Symbol, **args):
+
+def symbols(names, *, cls=Symbol, **args) -> Any:
```
asmeurer, line 585: Does this do anything? And couldn't it be `Union[Symbol,Tuple[Symbol]]`?

oscarbenjamin: It's probably not possible to type this but really the type is the return value of the `cls` parameter (or a tuple of the same).

eagleoflqj: `Any` forces mypy to check function body. Apparently mypy doesn’t complain this function on master. But if you add `Any`, it will.

eagleoflqj: Having Union as return type of not-overloaded function always makes trouble, as at call site `isinstanceof` has to be inserted.
I personally hope the `OneOf` type proposed (here)[https://github.com/python/typing/issues/1096#issuecomment-1055673065] could be added to standard and we can have union return type here.

oscarbenjamin:
To me this looks fine.

sympy/core/symbol.py
asmeurer: > Any forces mypy to check function body. Apparently mypy doesn’t complain this function on master. But if you add Any, it will.

I would have expected the opposite, that without `Any` it tries to do type inference on the body, but with `Any` it takes that as a signal to not try type inference.

asmeurer:
Looks good to me too.

Issue #23156, opened 2022-02-24
sympy.Sum() bug


yz3558:
x=sympy.Symbol("x")
a=sympy.Symbol("a")
sympy.Sum(1/sympy.gamma(x),(x,0,0)).doit()
sympy.Sum(1/sympy.gamma(x),(x,0,a)).doit().subs({a:0})

These two lines give different results. The first one gives 0, while the second one gives e.
The first result is correct.

anutosh491:
Well I can say that the result for a random integer `a` is correct 
```
>>> Sum(1/gamma(x), (x, 0, a)).doit()
-E*a*lowergamma(a, 1)/gamma(a + 1) + E
```

anutosh491:
I can confirm that all cases involving positive `a` look fine but doesn't work otherwise . So does it mean the general sum should only be returned if `a` is positive .
```
>>> Sum(1/gamma(x), (x, 0, a)).doit().subs(a, -2)
E
```
Yeah I would expect `nan` as pointed above but otherwise can't comment just by surface level analysis , cause the general sum till a random positive integer `a` looks fine to me !
Obviously the code snippet answering `Sum(1/gamma(x), (x, 0, 0)).doit()` is as follows so it returns `1/zoo` 
```
def eval_sum(f, limits):
    (i, a, b) = limits
    if a == b:
        return f.subs(i, a)
```

yz3558:
Can we define `0*lowergamma(0,1)=1`? This will make it consistent between 0 and positive `a` values.

anutosh491:
> Can we define `0*lowergamma(0,1)=1`? This will make it consistent between `a=0` and positive `a` values.

I don't think we can do this cause `lowergamma(0, 1)` evaluates `oo` and as per sympy convention `0*oo = nan`

yz3558:
If we identify a lowergamma(0,1) in the expression, we can always see if we have a 0 multiplies this term. If so, we get a 1. I think it is better nan.

anutosh491:
> If we identify a`lowergamma(0,1)` in the expression, instead of converting it to `oo` first, can we first see if we have a `0` multiplies this term? If so, we get a `1`.

I don't think this would be the best way to address this . We shouldn't be writing code to address a particular cases explicitly and try to buid an approach which generally works fine overall !

PR #23190, opened 2022-03-01
Functions/special/gamma : Added special case (when a is 0) for evaluating lowergamma function

anutosh491:
Addresses https://github.com/sympy/sympy/issues/23156#issuecomment-1049992233

Some examples from wolfram for reference

imageimageimage
---------------------------------------------------------------------------------------------------------------------------------------
image

#### Release Notes

* series
  * Added _eval_rewrite_as_tractable method for the lowergamma function.
* functions
  * Added special case for evaluation of lowergamma function.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1238,8 +1238,8 @@ alijosephine <alijosephine@gmail.com>
 amsuhane <ayushsuhane99@iitkgp.ac.in> amsuhane <“ayushsuhane99@iitkgp.ac.in”>
 anca-mc <anca-mc@users.noreply.github.com>
 andreo <andrey.torba@gmail.com>
-anutosh491 <andersonbhat491@gmail.com> Anutosh Bhat <87052487+anutosh491@users.noreply.github.com>
 anutosh491 <87052487+anutosh491@users.noreply.github.com>
+anutosh491 <andersonbhat491@gmail.com> Anutosh Bhat <87052487+anutosh491@users.noreply.github.com>
 arooshiverma <av22@iitbbs.ac.in>
 bluebrook <perl4logic@gmail.com>
 carstimon <carstimon@gmail.com>
--- a/.mailmap
+++ b/.mailmap
@@ -105,8 +105,8 @@ Akash Trehan <akash.trehan123@gmail.com>
 Akash Vaish <akash.9712@gmail.com>
 Akhil Rajput <akh1lrjput@gmail.com> Akhil Rajput <43316490+akhil-dh1@users.noreply.github.com>
 Akira Kyle <ak@akirakyle.com>
-Akshansh Bhatt <akshansh@tuta.io> Akshansh <53227127+akshanshbhatt@users.noreply.github.com>
-Akshansh Bhatt <akshansh@tuta.io> akshanshbhatt <53227127+akshanshbhatt@users.noreply.github.com>
+Akshansh Bhatt <akshansh@tuta.io>
+Akshansh Bhatt <akshansh@tuta.io> <53227127+akshanshbhatt@users.noreply.github.com>
 Akshat Jain <akshat.jain@students.iiit.ac.in>
 Akshat Maheshwari <akshat14714@gmail.com>
 Akshat Sood <68052998+akshatsood2249@users.noreply.github.com>
@@ -1238,8 +1238,8 @@ alijosephine <alijosephine@gmail.com>
 amsuhane <ayushsuhane99@iitkgp.ac.in> amsuhane <“ayushsuhane99@iitkgp.ac.in”>
 anca-mc <anca-mc@users.noreply.github.com>
 andreo <andrey.torba@gmail.com>
-anutosh491 <87052487+anutosh491@users.noreply.github.com>
 anutosh491 <andersonbhat491@gmail.com> Anutosh Bhat <87052487+anutosh491@users.noreply.github.com>
+anutosh491 <andersonbhat491@gmail.com> anutosh491 <87052487+anutosh491@users.noreply.github.com>
 arooshiverma <av22@iitbbs.ac.in>
 bluebrook <perl4logic@gmail.com>
 carstimon <carstimon@gmail.com>
--- a/.mailmap
+++ b/.mailmap
@@ -1239,6 +1239,7 @@ amsuhane <ayushsuhane99@iitkgp.ac.in> amsuhane <“ayushsuhane99@iitkgp.ac.in”
 anca-mc <anca-mc@users.noreply.github.com>
 andreo <andrey.torba@gmail.com>
 anutosh491 <andersonbhat491@gmail.com> Anutosh Bhat <87052487+anutosh491@users.noreply.github.com>
+anutosh491 <87052487+anutosh491@users.noreply.github.com>
 arooshiverma <av22@iitbbs.ac.in>
 bluebrook <perl4logic@gmail.com>
 carstimon <carstimon@gmail.com>
--- a/sympy/functions/special/gamma_functions.py
+++ b/sympy/functions/special/gamma_functions.py
@@ -319,7 +319,7 @@ class lowergamma(Function):
 
         # Special values.
         if a.is_Number:
-            if a is S.Zero and x.is_number:
+            if a is S.Zero and x.is_zero is False:
                 return S.Infinity
             if a is S.One:
                 return S.One - exp(-x)
--- a/sympy/functions/special/gamma_functions.py
+++ b/sympy/functions/special/gamma_functions.py
@@ -319,6 +319,8 @@ class lowergamma(Function):
 
         # Special values.
         if a.is_Number:
+            if a is S.Zero and not x is S.Zero:
+                return oo
             if a is S.One:
                 return S.One - exp(-x)
             elif a is S.Half:
--- a/sympy/functions/special/gamma_functions.py
+++ b/sympy/functions/special/gamma_functions.py
@@ -320,8 +320,7 @@ class lowergamma(Function):
         # Special values.
         if a.is_Number:
             if a is S.Zero and x.is_number:
-                if x.is_zero is False:
-                    return S.Infinity
+                return S.Infinity
             if a is S.One:
                 return S.One - exp(-x)
             elif a is S.Half:
--- a/sympy/functions/special/gamma_functions.py
+++ b/sympy/functions/special/gamma_functions.py
@@ -319,8 +319,9 @@ class lowergamma(Function):
 
         # Special values.
         if a.is_Number:
-            if a is S.Zero and not x is S.Zero:
-                return oo
+            if a is S.Zero and x.is_number:
+                if x.is_zero is False:
+                    return S.Infinity
             if a is S.One:
                 return S.One - exp(-x)
             elif a is S.Half:
--- a/sympy/functions/special/gamma_functions.py
+++ b/sympy/functions/special/gamma_functions.py
@@ -2,7 +2,7 @@ from sympy.core import Add, S, sympify, Dummy, expand_func
 from sympy.core.expr import Expr
 from sympy.core.function import Function, ArgumentIndexError, PoleError
 from sympy.core.logic import fuzzy_and, fuzzy_not
-from sympy.core.numbers import Rational, pi, oo, I
+from sympy.core.numbers import Rational, pi, oo, I, zoo
 from sympy.core.power import Pow
 from sympy.functions.special.zeta_functions import zeta
 from sympy.functions.special.error_functions import erf, erfc, Ei
@@ -320,7 +320,7 @@ class lowergamma(Function):
         # Special values.
         if a.is_Number:
             if a is S.Zero and x.is_zero is False:
-                return S.Infinity
+                return zoo
             if a is S.One:
                 return S.One - exp(-x)
             elif a is S.Half:
--- a/sympy/functions/special/gamma_functions.py
+++ b/sympy/functions/special/gamma_functions.py
@@ -385,6 +385,9 @@ class lowergamma(Function):
     def _eval_rewrite_as_uppergamma(self, s, x, **kwargs):
         return gamma(s) - uppergamma(s, x)
 
+    def _eval_rewrite_as_tractable(self, s, x, **kwargs):
+        return self.rewrite(uppergamma)
+
     def _eval_rewrite_as_expint(self, s, x, **kwargs):
         from sympy.functions.special.error_functions import expint
         if s.is_integer and s.is_nonpositive:
--- a/sympy/series/limits.py
+++ b/sympy/series/limits.py
@@ -280,7 +280,7 @@ class Limit(Expr):
                 newe = e.subs(z, z + z0)
             try:
                 coeff, ex = newe.leadterm(z, cdir=cdir)
-            except ValueError:
+            except (ValueError, NotImplementedError):
                 pass
             else:
                 if ex > 0:
--- a/sympy/functions/special/tests/test_gamma_functions.py
+++ b/sympy/functions/special/tests/test_gamma_functions.py
@@ -120,8 +120,10 @@ def tn_branch(s, func):
 def test_lowergamma():
     from sympy.functions.special.error_functions import expint
     from sympy.functions.special.hyper import meijerg
+    n = Symbol('n', nonzero = True)
     assert lowergamma(0, 1) == oo
     assert lowergamma(0, oo) == oo
+    assert lowergamma(0, n) == oo
     assert lowergamma(x, 0) == 0
     assert lowergamma(x, y).diff(y) == y**(x - 1)*exp(-y)
     assert td(lowergamma(randcplx(), y), y)
--- a/sympy/functions/special/tests/test_gamma_functions.py
+++ b/sympy/functions/special/tests/test_gamma_functions.py
@@ -121,9 +121,9 @@ def test_lowergamma():
     from sympy.functions.special.error_functions import expint
     from sympy.functions.special.hyper import meijerg
     n = Symbol('n', nonzero = True)
-    assert lowergamma(0, 1) == oo
-    assert lowergamma(0, oo) == oo
-    assert lowergamma(0, n) == oo
+    assert lowergamma(0, 1) == zoo
+    assert lowergamma(0, oo) == zoo
+    assert lowergamma(0, n) == zoo
     assert lowergamma(x, 0) == 0
     assert lowergamma(x, y).diff(y) == y**(x - 1)*exp(-y)
     assert td(lowergamma(randcplx(), y), y)
--- a/sympy/functions/special/tests/test_gamma_functions.py
+++ b/sympy/functions/special/tests/test_gamma_functions.py
@@ -120,6 +120,8 @@ def tn_branch(s, func):
 def test_lowergamma():
     from sympy.functions.special.error_functions import expint
     from sympy.functions.special.hyper import meijerg
+    assert lowergamma(0, 1) == oo
+    assert lowergamma(0, oo) == oo
     assert lowergamma(x, 0) == 0
     assert lowergamma(x, y).diff(y) == y**(x - 1)*exp(-y)
     assert td(lowergamma(randcplx(), y), y)
--- a/sympy/series/tests/test_limits.py
+++ b/sympy/series/tests/test_limits.py
@@ -327,6 +327,11 @@ def test_lowergamma_at_origin():
     assert limit(lowergamma(a, x), a, 0, '-') == -oo
     assert limit(lowergamma(I*a, x), a, 0) == -oo*I
     assert limit(lowergamma(I*a, x), a, 0, '-') == oo*I
+    assert limit(lowergamma(a, 1), a, 0) == oo
+    assert limit(x*lowergamma(x, 1)/gamma(x + 1), x, 0) == 1
+    # issue 23156
+    expr = Sum(1/gamma(x), (x, 0, a)).doit()
+    assert limit(expr, a, 0) == 0
 
 
 @XFAIL
--- a/sympy/series/tests/test_limits.py
+++ b/sympy/series/tests/test_limits.py
@@ -16,7 +16,7 @@ from sympy.functions.elementary.trigonometric import (acos, acot, acsc, asec, as
                                                       atan, cos, cot, csc, sec, sin, tan)
 from sympy.functions.special.bessel import (besselj, besselk)
 from sympy.functions.special.error_functions import (Ei, erf, erfc, erfi, fresnelc, fresnels)
-from sympy.functions.special.gamma_functions import (digamma, gamma, uppergamma)
+from sympy.functions.special.gamma_functions import (digamma, gamma, uppergamma, lowergamma)
 from sympy.integrals.integrals import (Integral, integrate)
 from sympy.series.limits import (Limit, limit)
 from sympy.simplify.simplify import simplify
@@ -319,6 +319,14 @@ def test_series_AccumBounds():
     # https://github.com/sympy/sympy/issues/12312
     e = 2**(-x)*(sin(x) + 1)**x
     assert limit(e, x, oo) == AccumBounds(0, oo)
+
+
+def test_lowergamma_at_origin():
+    a, x = symbols('a x')
+    assert limit(lowergamma(a, x), a, 0) == oo
+    assert limit(lowergamma(a, x), a, 0, '-') == -oo
+    assert limit(lowergamma(I*a, x), a, 0) == -oo*I
+    assert limit(lowergamma(I*a, x), a, 0, '-') == oo*I
 
 
 @XFAIL


jksuom:
It seems that the limit of `lowergamma(a, x)` as `a -> 0` depends on the direction. It is `oo` when `a -> 0+` and `-oo` when `a -> 0-`.

sympy/functions/special/gamma_functions.py
```diff
@@ -319,6 +319,8 @@ def eval(cls, a, x):
 
         # Special values.
         if a.is_Number:
+            if a is S.Zero and not x is S.Zero:
+                return oo
```
oscarbenjamin, line 323: The test `not x is S.Zero` is insufficient because `x` could be an expression that is possibly equal to zero but not literally `S.Zero`.

anutosh491:
> It seems that the limit of `lowergamma(a, x)` as `a -> 0` depends on the direction. It is `oo` when `a -> 0+` and `-oo` when `a -> 0-`.

Yeah , I was considering this to be a quick fix hence , this consider too many details but your suggestions are completely valid @jksuom @oscarbenjamin

anutosh491:
> It seems that the limit of `lowergamma(a, x)` as `a -> 0` depends on the direction. It is `oo` when `a -> 0+` and `-oo` when `a -> 0-`.

This is correct . So like .... were you trying to point out  some possible improvements in limits ?

```
>>> limit(limit(lowergamma(a, x), a, 0, '-'), x, 0) # This should be -oo I guess
oo

```
Also other mistakes I can see are 
```
>>> limit(lowergamma(a, x), a, 0) # This should be lowergamma(0, x)
oo
>>> limit(limit(lowergamma(a, x), a, 0), x, 1) # This should be lowergamma(0, 1)
oo
```
I wasn't consedering this to be a pr which needed much thought, I just thought a speical case evaluation has been missed out but now I see how the pr could be improved !

sympy/functions/special/gamma_functions.py
anutosh491: How do I tackle this ? Like do you mean to say I need to use something like `simplify/solve` (to simplify or solve the equation) here ?

oscarbenjamin: Definitely do not use `simplify` or `solve` as those are far too expensive. You could potentially use something like `x.is_zero is False` but even that can be expensive. It isn't necessary in all cases for the function to evaluate so you should consider whether it is even necessary to try and resolve any particular case.

anutosh491: Actually I realize what we are trying to handle has already been taken care of , we can get rid of this check ! Cause the cases I was trying to tackle as being handled by lines (307-308) which go as follows
```
        if x is S.Zero:
            return S.Zero
```

anutosh491:
>It seems that the limit of lowergamma(a, x) as a -> 0 depends on the direction. It is oo when a -> 0+ and -oo when a -> 0-.

@jksuom did I answer your doubt or am I still missing something , not sure! would be glad if you could maybe point out an example where you could see the commits going wrong

anutosh491:
```
>>> limit(lowergamma(a, x), a, 0)
lowergamma(0, x)
>>> limit(lowergamma(-a, x), a, 0)
lowergamma(0, x)
>>> limit(lowergamma(a*I, x), a, 0)
lowergamma(0, x)
```
Cool this goes through `gruntz ` as leading term methods don't exist for `lowergamma`/`uppergamma` yet . I'll try to approach this through gruntz soon if can be fixed through that ! Will also try implementing the leading term methods in near future maybe on another pr .......Never thought how  could those methods be implemented for functions depending on two variables (in this case `a` and `x`) . Will be interesting though !

anutosh491:
Btw I realized we can get this through `gruntz` easily , cause `uppergamma  `has an `_eval_rewrite_as_tractable` function so we could just include a similar rewrite function and get all correct results . 
```
--- a/sympy/functions/special/gamma_functions.py
+++ b/sympy/functions/special/gamma_functions.py
@@ -396,6 +396,9 @@ def _eval_is_zero(self):
         if x.is_zero:
             return True

+    def _eval_rewrite_as_tractable(self, s, x, **kwargs):
+        return self.rewrite(uppergamma)
+
```
This gives us 
```
>>> a, x = symbols('a x')
>>> limit(lowergamma(a, x), a, 0)
oo
>>> limit(lowergamma(a, x), a, 0, '-')
-oo
>>> limit(lowergamma(I*a, x), a, 0, '-')
oo*I
>>> limit(lowergamma(I*a, x), a, 0)
-oo*I
```

Didn't look at it as a pr fixing limits originally when I made the pr , but now maybe labels for `functions` /`limits` can be added here . I'll make the the above now !

anutosh491:
I thought of going with a number based name for the function but it only made sense to me later that something like `test_lowergamma_at_origin` would be better !! So I have added it where I felt we could have that (not seeing any proper dictionary based order for non -number based tests) . We could change that if we find somewhere better to place it . 

Other than that if @oscarbenjamin gives the green flag with this https://github.com/sympy/sympy/pull/23190#discussion_r817678908 or wants me to address anything in particular  , this could be ready to go !

anutosh491:
Okay so `test/test2-latest` fails as shows this message which tells me that `Run pip install mpmath` didn't run correctly ! But what's weird in this case is `test/test1-latest` needs to command too and there it worked perfectly fine , also I ran all tests locally so I didn't expect anything to fail 

```
Run pip install mpmath
Collecting mpmath
  WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x[7](https://github.com/sympy/sympy/runs/5401528515?check_suite_focus=true#step:5:7)fb91da9d700>: Failed to establish a new connection: [Errno 101] Network is unreachable')': /packages/d4/cf/3965bddbb4f1a61c49aacae0e7[8](https://github.com/sympy/sympy/runs/5401528515?check_suite_focus=true#step:5:8)fd1fe36b5dc36c7[9](https://github.com/sympy/sympy/runs/5401528515?check_suite_focus=true#step:5:9)7b31f30cf07dcbbb7/mpmath-1.2.1-py3-none-any.whl
  WARNING: Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x7fb91da9d9d0>: Failed to establish a new connection: [Errno [10](https://github.com/sympy/sympy/runs/5401528515?check_suite_focus=true#step:5:10)1] Network is unreachable')': /packages/d4/cf/3965bddbb4f1a61c49aacae0e78fd1fe36b5dc36c797b31f30cf07dcbbb7/mpmath-1.2.1-py3-none-any.whl
  WARNING: Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x7fb91da9db80>: Failed to establish a new connection: [Errno 101] Network is unreachable')': /packages/d4/cf/3965bddbb4f1a61c49aacae0e78fd1fe36b5dc36c797b31f30cf07dcbbb7/mpmath-1.2.1-py3-none-any.whl
  WARNING: Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x7fb91da9dd30>: Failed to establish a new connection: [Errno 101] Network is unreachable')': /packages/d4/cf/3965bddbb4f1a61c49aacae0e78fd1fe36b5dc36c797b31f30cf07dcbbb7/mpmath-1.2.1-py3-none-any.whl
  WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x7fb91da9dee0>: Failed to establish a new connection: [Errno 101] Network is unreachable')': /packages/d4/cf/3965bddbb4f1a61c49aacae0e78fd1fe36b5dc36c797b31f30cf07dcbbb7/mpmath-1.2.1-py3-none-any.whl
ERROR: Could not install packages due to an OSError: HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Max retries exceeded with url: /packages/d4/cf/3965bddbb4f1a61c49aacae0e78fd1fe36b5dc36c797b31f30cf07dcbbb7/mpmath-1.2.1-py3-none-any.whl (Caused by NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x7fb91da9dfd0>: Failed to establish a new connection: [Errno 101] Network is unreachable'))

```

anutosh491:
okay so 

> Okay so `test/test2-latest` fails and shows this message which tells me that `Run pip install mpmath` didn't run correctly ! But what's weird in this case is `test/test1-latest` needs the same command too and there it worked perfectly fine , also I ran all tests locally so I didn't expect any sympy test to fail
> 
> ```
> Run pip install mpmath
> Collecting mpmath
>   WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x[7](https://github.com/sympy/sympy/runs/5401528515?check_suite_focus=true#step:5:7)fb91da9d700>: Failed to establish a new connection: [Errno 101] Network is unreachable')': /packages/d4/cf/3965bddbb4f1a61c49aacae0e7[8](https://github.com/sympy/sympy/runs/5401528515?check_suite_focus=true#step:5:8)fd1fe36b5dc36c7[9](https://github.com/sympy/sympy/runs/5401528515?check_suite_focus=true#step:5:9)7b31f30cf07dcbbb7/mpmath-1.2.1-py3-none-any.whl
> 
> Error: Process completed with exit code 1.
> ```

Okay so closing and re-opening fixed the mpmath run issue (still not sure why it had failed , never happened before atleast on my prs) . I was sure that this could be fixed by closing and re-opening would fix that but I was waiting if I someone could educate me regarding why the test failed . Anyways all tests should pass now as they do locally !

anutosh491:
@jksuom does this look better now ?

sympy/functions/special/gamma_functions.py
anutosh491: I'll try getting rid of that check cause the cases which I want to handle (when `x.is_number` and `not x.is_zero`) will always be handled correctly from the above reported line ! If you feel something should be reverted back , i'll do the needful @oscarbenjamin

oscarbenjamin: I see now that this is in an `a.is_Number` block. In that case we do know whether `a` is equal to zero so it's fine to check `a is S.Zero` (although I would write it as `a is not S.Zero`).

anutosh491: Why do you prefer it as `a is not S.Zero` . I mean the special case goes as follows right ! 
1) if `a==0 `and `x == 0` : we should return `0`
2) else if `a== 0` and x is any other number : we should return `oo `

Like the special case is focused on the cases where `a== 0` , hence I don't see the negation case here , am I missing something . The first case is being answered on line 307 -308 and so I only need to address the second case which i've done in the latest commit !

anutosh491:
I've also made a basic check , good that the `0` case was pointed out (was a corner case which was being ignored).Other limits when `a` tends to all others values works quite well I must say , exact results as wolfram !

anutosh491:
Also @jksuom I had a doubt hear , now that the above limits should work , we would expect something like following shown below to work but we get an error
```
>>> limit(lowergamma(x, 1), x, 0)
NotImplementedError: lowergamma has no _eval_as_leading_term routine
```


Now this is because `lowergamma(x, 1).is_meromorphic(x, 0)` is `True`  and hence it enters the following block and encounters the except block where we just have been catching any sort of  `ValueError` .


```
--- a/sympy/series/limits.py
+++ b/sympy/series/limits.py
@@ -272,6 +272,7 @@ def set_signs(expr):
         if e.is_meromorphic(z, z0):
             if abs(z0) is S.Infinity:
                 newe = e.subs(z, 1/z)
@@ -280,7 +281,7 @@ def set_signs(expr):
                 newe = e.subs(z, z + z0)
             try:
                 coeff, ex = newe.leadterm(z, cdir=cdir)
            except ValueError:
```

Obviously this is solved by the following .
```
-            except ValueError:
+            except (ValueError, NotImplementedError):
```

Is this not been added for a reason like the block following that (line 309) uses this . I'll maybe run a test adding the above change to see if anything fails !
```
            coeff, ex = newe.leadterm(z, cdir=cdir)
        except (ValueError, NotImplementedError, PoleError):
            # The NotImplementedError catching is for custom functions
```

anutosh491:
Maybe we can add it , can solve cool random issues , check this out 
image

On Master
```
>>> e = x*lowergamma(x, 1)/gamma(x + 1)
>>> e.limit(x, 0)
  File "C:\Users\anuto\sympy\sympy\sympy\core\expr.py", line 3390, in limit
    return limit(self, x, xlim, dir)
  File "C:\Users\anuto\sympy\sympy\sympy\series\limits.py", line 64, in limit
    return Limit(e, z, z0, dir).doit(deep=False)
  File "C:\Users\anuto\sympy\sympy\sympy\series\limits.py", line 282, in doit
    coeff, ex = newe.leadterm(z, cdir=cdir)
  File "C:\Users\anuto\sympy\sympy\sympy\core\expr.py", line 3485, in leadterm
    l = self.as_leading_term(x, logx=logx, cdir=cdir)
  File "C:\Users\anuto\sympy\sympy\sympy\core\cache.py", line 70, in wrapper
  File "C:\Users\anuto\sympy\sympy\sympy\core\expr.py", line 3448, in as_leading_term
    obj = self._eval_as_leading_term(x, logx=logx, cdir=cdir)
  File "C:\Users\anuto\sympy\sympy\sympy\core\mul.py", line 2022, in _eval_as_leading_term
    return self.func(*[t.as_leading_term(x, logx=logx, cdir=cdir) for t in self.args])
  File "C:\Users\anuto\sympy\sympy\sympy\core\mul.py", line 2022, in <listcomp>
    return self.func(*[t.as_leading_term(x, logx=logx, cdir=cdir) for t in self.args])
  File "C:\Users\anuto\sympy\sympy\sympy\core\cache.py", line 70, in wrapper
  File "C:\Users\anuto\sympy\sympy\sympy\core\expr.py", line 3448, in as_leading_term
    obj = self._eval_as_leading_term(x, logx=logx, cdir=cdir)
  File "C:\Users\anuto\sympy\sympy\sympy\core\function.py", line 826, in _eval_as_leading_term
    raise NotImplementedError(
NotImplementedError: lowergamma has no _eval_as_leading_term routine
```

If we add the change !
```
>>> e = x*lowergamma(x, 1)/gamma(x + 1)
>>> e.limit(x, 0)
1
```

anutosh491:
This somewhat solves the issue denoted above too , wasn't expecting this but it does . Without the change raise `NotImplementedError` after big traceback 
```
>>> e = summation(1/gamma(n), (n, 0, x))
>>> limit(e, x, 0)
0
```

jksuom:
> NotImplementedError: lowergamma has no _eval_as_leading_term routine

The natural fix is to implement the missing method instead of masking the error.

anutosh491:
> > NotImplementedError: lowergamma has no _eval_as_leading_term routine
> 
> The natural fix is to implement the missing method instead of masking the error.

True agreed , but I feel we could go with this . I'll tell you why I feel this way .. so  `limits` involving `lowergamma` give out `NotImplementedError` in 2 ways only .

```
>>> limit(lowergamma(n + 1, n), n, -1)
NotImplementedError: lowergamma has no _eval_as_leading_term routine
>>> 
>>> limit(lowergamma(n + 1, n), n, oo)
NotImplementedError: MRV set computation for functions in several variables not implemented.
>>> 
>>> limit(lowergamma(n + 1, n)/factorial(n), n, oo)  # example from #14072
NotImplementedError: MRV set computation for functions in several variables not implemented.
```
Hence now that we implement this change (maybe untill we have an implementation for leading term methods) , the ones which used to give this error `NotImplementedError: lowergamma has no _eval_as_leading_term routine` would still give the `NotImplementedError ` but through gruntz .

```
>>> limit(lowergamma(n + 1, n), n, -1)
NotImplementedError: MRV set computation for functions in several variables not implemented.
```
Hence we aren't loosing any info here . So those changes which can't be truly addressed through the commits here still represent similar info but at the other hand we will be able to solve good limits at origin !

I obviously would like to plan on implementing the leading term method in near future(mostly not through this pr though as the significance of this pr is somewhat different) but I'm quite impressed with what `gruntz` can do here and even the limits you pointed out work quite well .Couple of limits like this one even took time on wolfram and worked quite fast here . 
```
>>> e = x*lowergamma(x, 1)/gamma(x + 1)
>>> e.limit(x, 0)
1
```
If hope this maybe convinces you but if not then I guess we could keep the limit changes for a seperate pr . Not sure why but I like the previous commits :)

anutosh491:
I guess we could have a good solid `as_leading_term` method for the `expint(nu, z)` function . Then use it for the `uppergamma(a, x) `function which inturn would be used for the `lowergamma(a, x)` case . I'll try to have write a prototype of the code sometime soon for later use !

anutosh491:
This could be a basic implementation of the leading term method considering some basic cases and cuts . I still would like to have the `_eval_rewrite_as_tractable`  implementation but a leading term method implementation for `expint` function would be great as we would get all results for `lowergamma/ uppergamma`  and other logarithmic integrals too ! 

What all would you like me to add /check here @jksuom . Would like your guidance here as this is the first time i'm implementing a leading term method for a function depending on 2 parameter but as of now this much is sufficient to solve **all limits** you pointed out above and the ones i've added which are `limit( x*lowergamma(x, 1)/gamma(x + 1), x, 0) ` etc . 

```
diff --git a/sympy/functions/special/error_functions.py b/sympy/functions/special/error_functions.py
index 1b0f58826f..f8c796e081 100644
--- a/sympy/functions/special/error_functions.py
+++ b/sympy/functions/special/error_functions.py
@@ -1400,6 +1400,44 @@ def _eval_rewrite_as_Si(self, nu, z, **kwargs):
     _eval_rewrite_as_Chi = _eval_rewrite_as_Si
     _eval_rewrite_as_Shi = _eval_rewrite_as_Si

+            nu = self.args[0]
+            arg = self.args[1].as_leading_term(x, logx=logx, cdir=cdir)
+            arg0 = arg.subs(x, 0)
+
+            if nu.is_Number:
+                if nu == 1:
+                     f = self._eval_rewrite_as_Si(*self.args) # could be put to use but class Si 
+                    c, e = arg.as_coeff_exponent(x)         # lacks leading term implementation yet 
+                    return -EulerGamma -log(c) -e*log(x)
+                elif nu.is_Integer and nu > 1:
+                    f = self._eval_rewrite_as_Ei(*self.args)
+                    return f._eval_as_leading_term(x, logx, cdir)
+            else:
+                return self.func(nu, arg0)
+
+


diff --git a/sympy/functions/special/gamma_functions.py b/sympy/functions/special/gamma_functions.py
index 0089b42a0e..dbd99a1995 100644
--- a/sympy/functions/special/gamma_functions.py
+++ b/sympy/functions/special/gamma_functions.py
@@ -383,9 +383,11 @@ def _eval_aseries(self, n, args0, x, logx):

+        from sympy.functions.special.error_functions import expint
+        return self.rewrite(expint).as_leading_term(x, logx = logx, cdir = cdir)
+
```

anutosh491:
Okay so I find this is implementation good enough to express the limits mentioned here . 
Also solves this one 
```
>>> limit(lowergamma(n, n - 1), n, 0, '-')
-oo
>>> limit(lowergamma(n, n - 1), n, 0, '-')
oo
```

.mailmap
```diff
@@ -1238,6 +1238,7 @@ alijosephine <alijosephine@gmail.com>
 amsuhane <ayushsuhane99@iitkgp.ac.in> amsuhane <“ayushsuhane99@iitkgp.ac.in”>
 anca-mc <anca-mc@users.noreply.github.com>
 andreo <andrey.torba@gmail.com>
+anutosh491 <87052487+anutosh491@users.noreply.github.com>
 anutosh491 <andersonbhat491@gmail.com> Anutosh Bhat <87052487+anutosh491@users.noreply.github.com>
```
oscarbenjamin, line 1242: This is listing you as two separate authors.

sympy/functions/special/gamma_functions.py
```diff
@@ -319,6 +319,8 @@ def eval(cls, a, x):
 
         # Special values.
         if a.is_Number:
+            if a is S.Zero and x.is_number:
+                return S.Infinity
```
oscarbenjamin, line 323: This doesn't seem right. What is the `x.is_number` check for?

.mailmap
anutosh491: Sure, will fix this

sympy/functions/special/gamma_functions.py
anutosh491: I tried to somewhat justify this in a comment above https://github.com/sympy/sympy/pull/23190#discussion_r819197236.

The special case addition is good and helping a lot of limits as the discussions here show but we don't want this right (which obviously is a result of getting rig of the `is_number` check )?
```
>>> lowergamma(0, x)
oo
```
The only case which I plan to affect is when `a = 0` and x is either `0` or non zero 
`lowergamma(0, 0) = 0` and `lowergamma(0, non-zero) = oo` as shown in the Op (wolfram results)  . What Kalevi pointed out is that most limits which we encounter would be dealing with either `a` or `x` being `0` or both , hence though this wasn't a pr addressing limits originally but it sort of fell into that path , hence his doubts were quite valid !

I hope I conveyed the point across !

oscarbenjamin: Read here:

You need to bear in mind that `x.is_zero` can also give None. There are 3 cases to handle: True, False and None. What should happen if `x` is an expression that is not known to be zero or nonzero? Also what if `x` is `Symbol('x', nonzero=True)`?

The check could maybe be `if a is S.Zero and x.is_zero is False`.

anutosh491: Oops my bad , cases like this just skipped my mind `Symbol('x', nonzero=True)` . Thanks for pointing this , will keep in mind for future work 

And yeah this would do it !
>if a is S.Zero and x.is_zero is False

oscarbenjamin: What should happen if `x` is infinite?
```python
In [57]: lowergamma(0, oo).evalf()
...
ValueError: gamma function pole
```
Again what if `x` is a symbol that is possibly infinite?

Numerically evaluating close to `0, zoo` gives different answers for different phases:
```python
In [61]: for u1 in [1, I, -1, -I]:
    ...:     for u2 in [1, I, -1, -I]:
    ...:         print(lowergamma(u1*1e-10, u2*1e10).evalf(5))
    ...: 
1.0000e+10
1.0e+10 - 0.21257*I
1.0778e+4342944809 + 3.3859e+4342944799*I
1.0e+10 + 0.21257*I
-0.57722 - 1.0e+10*I
-0.63206 - 1.0e+10*I
1.0778e+4342944809 + 2.4816e+4342944800*I
-0.6024 - 1.0e+10*I
-1.0000e+10
-1.0e+10 - 0.18292*I
1.0778e+4342944809 - 3.3859e+4342944799*I
-1.0e+10 + 0.18292*I
-0.57722 + 1.0e+10*I
-0.6024 + 1.0e+10*I
1.0778e+4342944809 - 2.4816e+4342944800*I
-0.63206 + 1.0e+10*I
```
Only one of these sixteen combinations resembles `oo`. We only hit this code path if `a.is_Number` so `a` would have to be real but still this doesn't look like it is approaching `oo` as `a` goes to zero:
```python
In [64]: lowergamma(-1e-10, 1e10)
Out[64]: -10000000000.5772
```

anutosh491: Well , as `lowergamma(a, x)` is better defined as `gamma(a) - uppergamma(a, x)`  as mentioned clearly in sympy and wolfram docs too . Hence we have 
```
>>> lowergamma(0, x).rewrite(uppergamma)
-expint(1, x) + zoo
>>> expint(1, oo).n()
0
>>> expint(1, 1).n()
0.219383934395520
>>> expint(1, -1).n()
-1.89511781635594 - 3.14159265358979*I
>>> expint(1, -oo).n()
-oo - 3.14159265358979*I
```
Now any of these values subtracted from `zoo` obviously gives `zoo ` . I went with `oo` just because of what's shown below and because of what wolfram says !
```
>>> lowergamma(0, 1).n() # these give oo and not zoo 
oo 
```
The best part here is wolfram returns `oo` for something like `lowergamma(0, 1)/lowergamma(0, oo) / lowergamma(0, -oo) `and `zoo` for `lowergamma(-1, 1) / lowergamma(-1, oo) / lowergamma(-1, -oo)`

The former ones have been shown in the Op , the later one is here  
image

I also checked out what `diofant` return for more clarity on this and it returns `zoo` for something like `lowergamma(0, 1)` .
I'm not too sure about values where `a` is around/tending to origin but when it is at origin , I'm quite sure comparing with other CAS and wiki that it deserves `oo` or `zoo `

anutosh491: Also not forgetting to answer your question ,we have the following currently which looks OK to me !
>>> x = Symbol('x', infinite=True)
>>> lowergamma(0, x)
oo

oscarbenjamin: That's definitely incorrect because `x` might represent `-oo`. I think probably `zoo` is the right result here.

anutosh491: Yeah but we have the following so having x as -oo shouldn't be an issue but I still see `zoo` being a better result here 
image .
I hope you also went through the comment above https://github.com/sympy/sympy/pull/23190#discussion_r820238810, which sort of explains why I ended up using `oo` instead of `zoo`

sympy/functions/special/gamma_functions.py
```diff
@@ -319,6 +319,8 @@ def eval(cls, a, x):
 
         # Special values.
         if a.is_Number:
+            if a is S.Zero and x.is_zero is False:
```
jksuom, line 322: a = 0 is not the only pole of lowergamma. They should preferably all be handled in the same place.

anutosh491: Yeah correct, I was reading wiki just now to make sure I'm not missing something when I realized this and as you pointed the check needs improvement , I am also doubtful about few other cases . Will makes the changes as I get more clarity here !

sympy/series/limits.py
```diff
@@ -280,7 +280,7 @@ def set_signs(expr):
                 newe = e.subs(z, z + z0)
             try:
                 coeff, ex = newe.leadterm(z, cdir=cdir)
-            except ValueError:
+            except (ValueError, NotImplementedError):
```
jksuom, line 283: This is not a good fix. There are many functions that do not have leading term and series expansion methods implemented, yet. Not raising `NotImplementedError` will make it harder to debug issues caused by the missing cases.

anutosh491: Yeah correct , I completely agree.  I also agree more to what you said which is 
> The natural fix is to implement the missing method instead of masking the error.

This is because as you were doubting my implementation behind the `_eval_rewrite_as_tractable` method for `lowergamma` function , I was thinking that the idea is mostly related to `_eval_rewrite_as_tractable` for the `uppergamma `function too  . This was implemented couple years ago 
```
    def _eval_rewrite_as_tractable(self, s, x, **kwargs):
        return exp(loggamma(s)) - lowergamma(s, x)
```
 Hence for some clarity I computed quite a good number of basic limits for `uppergamma ` but I realize this method might end up to be quite buggy . Even Basic limits for `upergamma` have good amount of bugs , the most basic one being 
```
>>> limit(uppergamma(a, x), a, 0)
oo
>>> limit(uppergamma(a, x), a, 0, '-')
 NotImplementedError: loggamma has no _eval_as_leading_term routine
```
The answer for this should be `expint(1, x) ` , I mean obviously if x tends to 0 here then we can expect oo but otherwise we shouldn't 
```
>>> limit(expint(1, x), x, 0)
oo
>>> limit(expint(1, x), x, 0, '-')
oo
```
Obviously if we comment this out ,we get the correct answer 
```
>>> limit(uppergamma(a, x), a, 0)
expint(1, x)
>>> limit(uppergamma(a, x), a, 0, '-')
expint(1, x)
```
image

I didn't even open this pr considering fixing limits but now I this has pr showed me so much scope of improvements in limits

anutosh491: Though fundamentally like concept wise I don't really see any obvious mistake here ..... I mean both functions used here the `exponential` one and the `lowergamma` one would end up having `aseries` , hence this doesn't seem wrong
```
    def _eval_rewrite_as_tractable(self, s, x, **kwargs):
        return exp(loggamma(s)) - lowergamma(s, x)
```

PR #23193, opened 2022-03-01
Fix the URL printed by mailmap_check in the CI

asmeurer:
#### Release Notes

- utilities
  - Pass keyword arguments of `filldedent()` through to `textwrap.fill()`.

### DIFF
--- a/bin/mailmap_check.py
+++ b/bin/mailmap_check.py
@@ -136,7 +136,8 @@ def main(*args):
     if problems:
         print(red(filldedent("""
         For instructions on updating the .mailmap file see:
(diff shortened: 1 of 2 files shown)


bin/mailmap_check.py
```diff
@@ -136,7 +136,8 @@ def key(line):
     if problems:
         print(red(filldedent("""
         For instructions on updating the .mailmap file see:
-        https://github.com/sympy/sympy/wiki/Development-workflow#update-mailmap""")))
+        https://github.com/sympy/sympy/wiki/Development-workflow#update-mailmap""",
```
oscarbenjamin, line 139: It actually needs to be:

I added that URL before the section existed so I didn't know what the correct URL would be.

PR #23194, opened 2022-03-02
Made all the parentheses in the LaTeX printer use \left and \right.

wermos:

Fixes the point raised [here](https://github.com/sympy/sympy/issues/23058#issuecomment-1037000230).

I made the LaTeX printer for `sympy.vector` objects use `\left(` and `\right)` in all cases.

#### Other comments
I think having extra `\left`s and `\right`s is better than having too few, because if resizing is not needed, then LaTeX automatically uses the regular parentheses. The only downside is that the expressions will be a little harder for humans to now read.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/printing/latex.py
+++ b/sympy/printing/latex.py
@@ -750,7 +750,7 @@ class LatexPrinter(Printer):
                 elif v == -1:
                     o1.append(' - ' + k._latex_form)
                 else:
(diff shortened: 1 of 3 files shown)


moorepants:
This makes sense to me. I've noticed some ugly parentheses output in the physics vectors recently.

Issue #22935, opened 2022-01-27
Periodicity of trigonometric expression incorrect .


faze-geek:
The period of powers of sin(x) and cos(x) should vary with the power n and should not always return 2*pi .

```
>>> x=Symbol('x')
>>> n=Symbol('n',integer = true , positive = true)
>>> periodicity(sin(x),x)
2*pi
>>> periodicity(sin(x)**n,x)
2*pi
>>> periodicity((cos(x)**2),x)
2*pi
```
When power n is even the period should be pi itself .

Is this a legit issue ? If I am not missing something here, I would like to work on this .

oscargus:
That seems to be an indeed valid issue.

Also, this is a bit "interesting":
```
In [6]: periodicity(Abs(sin(x)),x)
Out[6]: π

In [7]: periodicity(Abs(sin(x)**2),x)
Out[7]: 2⋅π
 
In [8]: periodicity(Abs(sin(x))**2,x)
Out[8]: 2⋅π
```
Clearly, there are things to improve here.

faze-geek:
I will work on this soon ,till then suggestions of other contributors are welcome !

AhmedTareek:
I would like to work on this, also I am new to contribution any advices on how to approach the problem ?

oscarbenjamin:
The docstring does not really specify what the `periodicity` function should be expected to return e.g. is it "some period" or the "minimal period". It is technically true that `cos(x)**2` is periodic with period `2*pi`. It is also true that it is periodic with period `pi`. Sometimes that distinction doesn't matter though. Maybe it's also useful to have a function that cheaply computes some period without trying too hard to find the minimal period. In any case if the expectation is that `periodicity` computes (or even just tries to compute) the minimum period then that should be made explicit.

> any advices on how to approach the problem

Look at the code for the `periodicity` function and maybe follow through in a debugger to see how it computes this.

AhmedTareek:
I divided the given period by equal spaces and substituted in the function to see if I get the same results again like this
![image](https://user-images.githubusercontent.com/92989834/160110624-2007509b-0b7a-4e67-be8d-92241b628c1f.png)

`from sympy import *

x = symbols('x')
f = cos(x)**2
a = periodicity(f, x)
space = a / 20
lists = []
y = 0
for i in range(20):
    lists.append(y)
    y += space
check = []
for i in lists:
    print(f.subs(x,i))
    print("----------------------------")
    check.append((f.subs(x,i).evalf()))
[...]

PR #23199, opened 2022-03-02
calculus : Added condition to improve is-convex() results 

faze-geek:
While working on #22935 ,I came across a wrong result which would need a minor fix.
The results are also consistent with online documentation on convexity/concavity .
Some links for reference -
[Wikipedia](https://en.wikipedia.org/wiki/Convex_function) ( See examples section )
[Symbolab concavity calculator](https://www.symbolab.com/solver/function-concavity-calculator) (Notice that it returns suitable domain of concavity but never includes singularities)

Currently on master - 
```
>>> x = Symbol('x')
>>> is_convex(1/x**2,x)
True
```

On branch -
```
>>> x = Symbol('x')
>>> is_convex(1/x**2,x)
False
>>> is_convex(1/x**2,x, domain=Interval.Lopen(0,oo))
True
```
#### Other comments
A couple of tests which show how both convex and concave part of a function is handled are added .
A test has been modified to satisfy the improvement.

#### Release Notes

* calculus
  * Added singularity check to is_convex()

### DIFF
--- a/sympy/calculus/util.py
+++ b/sympy/calculus/util.py
@@ -666,6 +666,8 @@ def is_convex(f, *syms, domain=S.Reals):
     True
     >>> is_convex(x**3, x, domain = Interval(-1, oo))
     False
+    >>> is_convex(1/x**2, x, domain=Interval.open(0, oo))
+    True
 
     References
     ==========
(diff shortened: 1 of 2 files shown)


faze-geek:
Ping @oscargus for review

sympy/calculus/util.py
```diff
@@ -686,6 +688,11 @@ def is_convex(f, *syms, domain=S.Reals):
 
     f = _sympify(f)
     var = syms[0]
+    s = singularities(f, var)
+    for ele in s:
+        if ele in domain:
+            return False
```
smichr, line 694:
```suggestion
    if any(s in domain for s in singularities(f, var)):
        return False
```

PR #23200, opened 2022-03-02
Different symbols for Laplacian

wermos:
SymPy uses `\triangle` to denote the Laplacian, which is nonstandard. However, using `\Delta` for that is very widespread:
* The [Comprehensive LaTeX Symbol List](https://mirror.kku.ac.th/CTAN/info/symbols/comprehensive/symbols-a4.pdf) lists \Delta as the symbol for the Laplacian on page 318. This is also mentioned in [this answer on tex.stackexchange.com](https://tex.stackexchange.com/a/76563/203821).
image
* As [this answer to a question about what to use for the Laplace Operator](https://tex.stackexchange.com/a/76564/203821) states, one of the aliases of U+2206 (increment) is indeed [Laplace Operator](https://unicode-table.com/en/2206/).
* [Wikipedia](https://en.wikipedia.org/wiki/Laplace_operator) exclusively uses \Delta, or U+0394 to denote the Laplace operator.

Because using `\Delta` is much more common than using `\triangle` for the Laplace operator, I propose changing the symbol used in the LaTeX printer.
#### Release Notes

* printing
  * The LaTeX printer now uses `\Delta` instead of `\triangle` for the Laplace operator.

### DIFF
--- a/sympy/printing/latex.py
+++ b/sympy/printing/latex.py
@@ -502,7 +502,7 @@ class LatexPrinter(Printer):
 
     def _print_Laplacian(self, expr):
         func = expr._expr
-        return r"\triangle %s" % self.parenthesize(func, PRECEDENCE['Mul'])
+        return r"\Delta %s" % self.parenthesize(func, PRECEDENCE['Mul'])
 
     def _print_Mul(self, expr):
         from sympy.physics.units import Quantity
--- a/sympy/printing/tests/test_latex.py
+++ b/sympy/printing/tests/test_latex.py
@@ -370,11 +370,11 @@ def test_latex_vector_expressions():
     assert latex(x*Gradient(A.x)) == r"x \left(\nabla \mathbf{{x}_{A}}\right)"
     assert latex(Gradient(x*A.x)) == r"\nabla \left(\mathbf{{x}_{A}} x\right)"
 
-    assert latex(Laplacian(A.x)) == r"\triangle \mathbf{{x}_{A}}"
+    assert latex(Laplacian(A.x)) == r"\Delta \mathbf{{x}_{A}}"
     assert latex(Laplacian(A.x + 3*A.y)) == \
-        r"\triangle \left(\mathbf{{x}_{A}} + 3 \mathbf{{y}_{A}}\right)"
-    assert latex(x*Laplacian(A.x)) == r"x \left(\triangle \mathbf{{x}_{A}}\right)"
-    assert latex(Laplacian(x*A.x)) == r"\triangle \left(\mathbf{{x}_{A}} x\right)"
+        r"\Delta \left(\mathbf{{x}_{A}} + 3 \mathbf{{y}_{A}}\right)"
+    assert latex(x*Laplacian(A.x)) == r"x \left(\Delta \mathbf{{x}_{A}}\right)"
+    assert latex(Laplacian(x*A.x)) == r"\Delta \left(\mathbf{{x}_{A}} x\right)"
 
 def test_latex_symbols():
     Gamma, lmbda, rho = symbols('Gamma, lambda, rho')


ThePauliPrinciple:
> because most people use `\Delta`, as described [here](https://tex.stackexchange.com/q/76553/203821).

The accepted answer uses `\newcommand*\Laplace{\mathop{}\!\mathbin\bigtriangleup}`. Not sure where your conclusion that "most people use `\Delta`" comes from.

asmeurer:
What is the difference between the two?

sylee957:
I think that `\triangle` prints out more thinner font, while `\Delta` prints out the correct reflection of the `\nabla`. But this can depend on the fonts.
(Check out in https://katex.org/)

wermos:
@ThePauliPrinciple It was my bad for being unclear, but I linked to the question itself, and not the accepted answer, because I expected people to read the comments. The comments state that the Comprehensive LaTeX Symbol List lists `\Delta` as the symbol for the Laplacian:
image
this is on page 318 on [here](https://mirror.kku.ac.th/CTAN/info/symbols/comprehensive/symbols-a4.pdf). (I was unable to find it on page 50 or 150, as suggested by the comments in that question.) This is also mentioned in [this answer](https://tex.stackexchange.com/a/76563/203821).

Moreover, as [this answer to that question](https://tex.stackexchange.com/a/76564/203821) states, one of the aliases of U+2206 (increment) is indeed [Laplace Operator](https://unicode-table.com/en/2206/).

Finally, we can see that, as suggested in the comments to that question, we see that [Wikipedia](https://en.wikipedia.org/wiki/Laplace_operator) exlcusively uses `\Delta`, or U+0394 to denote the Laplace operator.

(Technically, U+0394 is incorrect, because U+2206 is the one which has "Laplace operator" as an alias, but I digress.)

oscargus:
I guess it depends on the context. If the d'Alembert operator is also used (which I do not think it is in SymPy), then \bigtriangle may be better. However, considering where it is defined etc, I think that matching with \nabla is more important, so I cast my vote on \Delta.

(An option is to use \nabla^2...)

oscargus:
If it turns out to be a problem, it is always possible to add a keyword option to the printer that determines how it is printed. But since it is incorrect at the moment, any solution is better than the current.

oscargus:
Sorry, I came in to this thinking that it printed as \nabla...

I'd then suggest to add a keyword argument to let the user select. And I think that \triangle should be the default, for historical reasons (it is not obviously wrong).

Note that it is better to use "sensible keywords" than to simply let the user select whatever LaTeX-code they want to put there. Maybe 't', 'b', and 'd' for triangle, bigtriangle, and delta. In that way we can check and throw and error (with a message including valid options).

wermos:
That's a good idea. I will implement that, if the others (@ThePauliPrinciple, @asmeurer, @sylee957) are onboard with this idea.

ThePauliPrinciple:
I agree that there should be the option to choose. If one would take the time to read all points of view in the linked stackexchange article it is clear that there is no consensus on what symbol should be used. The OP should be updated to reflect this as it might cause confusion in the future when this PR is used as a reference.

wermos:
@ThePauliPrinciple Do you have any suggestions as to what the OP should be changed to?

ThePauliPrinciple:
It should indicate that there is discussion about what is the right symbol and that SymPy provides an option for that reason.

wermos:
@ThePauliPrinciple @asmeurer @sylee957 @oscargus Is the new description and proposed change better? I will have to update this PR to reflect the changes.

asmeurer:
IMO we should just use \Delta. It seems that is the best option. We could add a configuration option for it, but I don't really see a need to do that unless someone requests it. Note that it's always possible for users to fine-tune the printer manually by subclassing it and overwriting the methods.

anutosh491:
> IMO we should just use \Delta. It seems that is the best option. We could add a configuration option for it, but I don't really see a need to do that unless someone requests it. Note that it's always possible for users to fine-tune the printer manually by subclassing it and overwriting the methods.

+1, I feel this is the correct way to go here .

wermos:
@asmeurer What should I do to this PR then? Should I implement a keyword argument, or can it be merged as is?

sylee957:
I think that I'm +1 to merge this, unless we receive an argument in a couple of days, for the reasons that

- We don't consider printing out the LaTeX in exact string literal a 'feature', than printing out in correct LaTeX.
- The implementation without options is more simple and easy to extend than the implementation with options.

But I think that you should first change the release notes before the merge.

oscargus:
Yeah, I am also not objecting to not implementing configuration at this stage. But please revert the release note. (Sorry about that.)

wermos:
@oscargus The release notes and the description have both been changed, as requested.

oscargus:
Thanks!

PR #23201, opened 2022-03-02
Use sympify less

oscargus:

The idea was to not use sympify in private functions, but to make sure that the arguments was already sympified. This turned out to be quite costly in terms of lines in polytools, but is otherwise a good idea.

It also turned out that there were quite a few unnecessary sympify-calls all over the code base.

And some small rewrites where I saw potential.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/algebras/quaternion.py
+++ b/sympy/algebras/quaternion.py
@@ -51,7 +51,7 @@ class Quaternion(Expr):
     is_commutative = False
 
     def __new__(cls, a=0, b=0, c=0, d=0, real_field=True):
-        a, b, c, d = sympify(a), sympify(b), sympify(c), sympify(d)
+        a, b, c, d = map(sympify, (a, b, c, d))
 
         if any(i.is_commutative is False for i in [a, b, c, d]):
             raise ValueError("arguments have to be commutative")
--- a/sympy/algebras/quaternion.py
+++ b/sympy/algebras/quaternion.py
@@ -7,7 +7,7 @@ from sympy.functions.elementary.trigonometric import (acos, cos, sin, atan2)
 from sympy.simplify.trigsimp import trigsimp
 from sympy.integrals.integrals import integrate
 from sympy.matrices.dense import MutableDenseMatrix as Matrix
-from sympy.core.sympify import sympify
+from sympy.core.sympify import sympify, _sympify
 from sympy.core.expr import Expr
 from sympy.core.logic import fuzzy_not, fuzzy_or
 
@@ -51,10 +51,7 @@ class Quaternion(Expr):
     is_commutative = False
 
     def __new__(cls, a=0, b=0, c=0, d=0, real_field=True):
-        a = sympify(a)
-        b = sympify(b)
-        c = sympify(c)
-        d = sympify(d)
+        a, b, c, d = sympify(a), sympify(b), sympify(c), sympify(d)
 
         if any(i.is_commutative is False for i in [a, b, c, d]):
             raise ValueError("arguments have to be commutative")
@@ -184,10 +181,10 @@ class Quaternion(Expr):
         return self.add(other*-1)
 
     def __mul__(self, other):
-        return self._generic_mul(self, other)
+        return self._generic_mul(self, _sympify(other))
 
     def __rmul__(self, other):
-        return self._generic_mul(other, self)
+        return self._generic_mul(_sympify(other), self)
 
     def __pow__(self, p):
         return self.pow(p)
@@ -301,7 +298,7 @@ class Quaternion(Expr):
         (2 + 3*I)*(3 + 4*I) + (2 + 3*I)*(2 + 5*I)*i + 0*j + (2 + 3*I)*(7 + 8*I)*k
 
         """
-        return self._generic_mul(self, other)
+        return self._generic_mul(self, _sympify(other))
 
     @staticmethod
     def _generic_mul(q1, q2):
@@ -313,7 +310,7 @@ class Quaternion(Expr):
         q1 : Quaternion or symbol
         q2 : Quaternion or symbol
 
-        It's important to note that if neither q1 nor q2 is a Quaternion,
+        It is important to note that if neither q1 nor q2 is a Quaternion,
         this function simply returns q1 * q2.
 
         Returns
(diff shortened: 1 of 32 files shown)


sympy/polys/polytools.py
```diff
@@ -1678,6 +1681,7 @@ def pexquo(f, g):
         ExactQuotientFailed: 2*x - 4 does not divide x**2 + 1
 
         """
+        g = sympify(g)
         _, per, F, G = f._unify(g)
```
oscarbenjamin, line 1685: Aren't these cases handled by `_unify_`?

oscargus: They were, but sometimes that leads to double sympification. And it seems like it makes sense that "private" functions do not sympify.

oscarbenjamin: I think that `_unify_` is different even if it is an internal call because it's purpose is essentially the same as sympify. Potentially some of these shouldn't use sympify though. They expect Poly rather than general Expr anyway e.g.:
```python
In [7]: Poly(x**2).pexquo(x)
CoercionFailed

In [16]: Poly(x**2).pexquo(Poly(x))
Out[16]: Poly(x, x, domain='ZZ')
```
I'm not sure exactly what the set of acceptable inputs here is if it is more than just `int/Poly`. It seems like it will maybe  accept any Expr that coerces to the ground domain e.g.:
```python
In [13]: Poly(x**2*y, x).pexquo(y)
Out[13]: Poly(y**3*x**2, x, domain='ZZ[y]')
```
There are separate top-level functions that take arbitrary Expr and then call these methods internally:
```python
In [14]: pexquo(x**2, x)
Out[14]: x
```
It's not really clear what the expected behaviour is.

oscargus: OK! I just went with the idea of sympify in fewer locations and `_unify` was really the outlier when it came to consequences. However, there will be fewer sympify calls in total now, although on more lines.

But it is of course possible to revert the change in that file if that is preferred.

oscargus:
It is also interesting that most `eval` methods do not sympify, but a few did (until now). Probably because there were doc-tests for that behavior. However, I do not really see why one should be able to call a subset of functions as `SingularityFunction(x, a, n).eval(3, 5, 1)`.

sympy/polys/polytools.py
oscarbenjamin: What causes the unnecessary sympify calls? How are you measuring this?

oscargus: In current master, add calls sympify first and then in `_unity`. 
https://github.com/sympy/sympy/blob/0bc2587aef09ab150bd408b7cd9e08baf7b5fd9c/sympy/polys/polytools.py#L1446-L1451

After this change, all calls are moved from `_unity` to the method calling `_unity`, so no redundant calls.

(Then I noted the _polifyit decorator here, so maybe there are still redundant sympifications, but beyond my knowledge if it should be possible to use e.g. the add method separately or if the overloading of + is enough.)

oscarbenjamin: I don't completely understand the design of `__add__` vs `add`. I would have made it so that only `__add__` sympifies but then when you know you have two Polys you can call `add`.

oscargus: Agreed. So the question is really if someone uses add instead? Maybe one should deprecated `add` and introduce `_add` instead? (Which is basically only called from `__add__`, or from library code where one know that it is sympified.)

Not in this PR though... (There are quite a lot of "public" functions and methods that clearly only are internal helpers. As you most likely know, but I was struck by it searching the code...)

oscarbenjamin: Maybe there should be a `_unify_sympify` function or conversely a `_unify_no_sympify` function. The main idea of the current design of `_unify` is to reduce repetition in these methods and the change here works against that.

oscargus: I reverted this file.

oscargus: But, yes, there is a discrepancy between redundant code and redundant calls.

sympy/algebras/quaternion.py
```diff
@@ -51,10 +51,7 @@ class Quaternion(Expr):
     is_commutative = False
 
     def __new__(cls, a=0, b=0, c=0, d=0, real_field=True):
-        a = sympify(a)
-        b = sympify(b)
-        c = sympify(c)
-        d = sympify(d)
+        a, b, c, d = sympify(a), sympify(b), sympify(c), sympify(d)
```
smichr, line 54:
```suggestion
        a, b, c, d = map(sympify, (a, b, c, d))
```

sympy/functions/elementary/miscellaneous.py
```diff
@@ -391,14 +391,12 @@ def __new__(cls, *args, **assumptions):
...
-            args = frozenset(args)
-
-        if evaluate:
             # remove redundant args that are easily identified
             args = cls._collapse_arguments(args, **assumptions)
             # find local zeros
-            args = cls._find_localzeros(args, **assumptions)
+            args = frozenset(cls._find_localzeros(args, **assumptions))
+        else:
+            args = frozenset(args)
```
smichr, line 399:
```suggestion
            args = cls._find_localzeros(args, **assumptions)
        args = frozenset(args)
```

sympy/algebras/quaternion.py
oscargus: Yes, one of multiple ways of doing it... 
```
a, b, c, d = [sympify(e) for e in (a, b, c, d)]
```

sympy/functions/elementary/miscellaneous.py
oscargus: Yeah, became clear for when one saw the diffs.

sympy/physics/continuum_mechanics/beam.py
```diff
@@ -136,7 +136,7 @@ def __init__(self, length, elastic_modulus, second_moment, area=Symbol('A'), var
         self._base_char = base_char
         self._boundary_conditions = {'deflection': [], 'slope': []}
         self._load = 0
-        self._area = area
+        self.area = area
```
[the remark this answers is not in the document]
oscargus: https://github.com/sympy/sympy/blob/0bc2587aef09ab150bd408b7cd9e08baf7b5fd9c/sympy/physics/continuum_mechanics/beam.py#L187-L194

So, it is not more public now than earlier. The difference is that area is always sympified.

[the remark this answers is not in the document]
oscargus: Ahh, but actually both lines should probably be there to initialize the member in `__init__`.

oscarbenjamin:
Looks good.

PR #23218, opened 2022-03-08
feat(physics): `quantity_simplify` can now optionally simplify units across dimensions

### LINKED ISSUE #21234: Can't get joule/m^3 to simplify into pascal
Sample:
```python
from sympy.physics.units import *
print(joule/meter**3)
print(simplify(joule/meter**3))
```
I expected one of these to give me pascal as the unit, but instead I get `joule/meter**3` in both cases.

If I try `(joule/meter**3).convert_to(pascal)` instead, I get `AttributeError: 'Mul' object has no attribute 'convert_to'`. However, `convert_to(joule/meter**3, pascal)` succeeds.

I also tried `sympy.physics.units.util.quantity_simplify(joule/meter**3)`, but that also gives me `joule/meter**3`

dyc3:

Fixes #21234

Added an optional flag, `across_dimensions` to the function `quantity_simplify`, which allows units to be simplified across dimensions, eg. joule/coulomb -> volt.

#### Release Notes

* physics.units
  * `Quantity`: it is now possible to determine if a Quantity is prefixed (eg. `nanometers`) with the property `is_prefixed`.
  * `Quantity`: it is now possible to determine if a Quantity represents a physics constant (eg. `speed_of_light`) with the property `is_physical_constant` or if is an instance of the new subclass `PhysicalConstant`.
  * `quantity_simplify()` can now simplify units across dimensions using the optional flag `across_dimensions=True`.


dyc3:
I'm confused, am I supposed to update the authors/mailmap?

sympy/physics/units/util.py
```diff
@@ -154,6 +158,46 @@ def quantity_simplify(expr):
         ref = v[0]/v[0].scale_factor
         expr = expr.xreplace({vi: ref*vi.scale_factor for vi in v[1:]})
 
+    if across_dimensions:
+        # combine quantities of different dimensions into a single
+        # quantity that is equivalent to the original expression
+
+        unit_system = UnitSystem.get_unit_system("SI")
```
Upabjojr, line 165: What about if you are not using SI unit systems?

Upabjojr:
Simplifying quantities across dimensions may be dependent on the unit system that is being used (i.e. there could be different results depending on the unit system being used).

I believe the unit system needs to be specified, otherwise the results may not correspond to the desired ones.

sympy/physics/units/util.py
dyc3: Ah, yeah I forgot to fix that

dyc3: I've added an extra argument for the unit system. However, I am uncertain on whether or not this sufficiently solves the problem because of the hard coded dict `preferred_dimension_to_unit`. 

Realistically though, I don't think we should consider it a blocker for merging because in 99.99% of cases, people are going to be using SI.

sympy/physics/units/util.py
```diff
@@ -120,21 +123,24 @@ def get_total_scale_factor(expr):
     return expr_scale_factor * Mul.fromiter((1/get_total_scale_factor(u) * u) ** p for u, p in zip(target_units, depmat))
 
 
-def quantity_simplify(expr):
+def quantity_simplify(expr, across_dimensions: bool=False, unit_system="SI"):
```
Upabjojr, line 126: I would suggest to go as far as not assuming a default unit system. That is, requiring the users to specify the unit system they want to use. The purpose of the _units_ module is to be useful for generic unit systems and I would attempt at distancing from _SI_ as much as possible.

Maybe `unit_system: Optional[Union[UnitSystem, str]] = None` and then raise an exception if `across_dimensions=True` and `unit_system` is `None`?

sympy/physics/units/util.py
```diff
@@ -154,6 +160,46 @@ def quantity_simplify(expr):
...
+        target_dimension = None
+        for result_dim, result_deps in dimension_system.dimensional_dependencies.items():
+            if result_deps == dim_deps:
+                target_dimension = result_dim
+                break
+
+        from . import volt, ohm, ampere, pascal, farad, second, watt, henry, meter, newton, weber, siemens, coulomb, tesla, candela, steradian
+        from . import voltage, impedance, current, pressure, capacitance, time, power, inductance, volume, force, luminous_intensity, magnetic_flux, conductance, charge, magnetic_density
+
+        preferred_dimension_to_unit = {
```
Upabjojr, line 181: This dictionary is assuming the base units of _SI_. I think these values can be extracted from the `unit_system` variable. This would make the code independent of the unit system.

oscargus:
> I'm confused, am I supposed to update the authors/mailmap?

Yes, the mailmap-file, see https://github.com/sympy/sympy/wiki/Development-workflow#add-your-name-and-email-address-to-the-mailmap-file  (Authors will be updated automatically on release.)

sympy/physics/units/util.py
dyc3: Other functions that take `unit_system` assume a default value of `"SI"`, like `convert_to()`. I think it would be better to at least remain consistent.

sympy/physics/units/util.py
dyc3: I agree. I'm unsure if this is possible without adding additional metadata to `Quantity`. For example, `bar` and `pascal` are both units of pressure. How do we know which one to pick?

I'll try it out and see what happens.

sympy/physics/units/util.py
Upabjojr: What about uniting the parameters `across_dimensions` and `unit_system`?

Upabjojr: > Other functions that take `unit_system` assume a default value of `"SI"`, like `convert_to()`

I regret that decision.

sympy/physics/units/util.py
Upabjojr: The unit system object should have a dictionary of scale factors, IIRC. Maybe that could be used to determine the base unit?

sympy/physics/units/util.py
dyc3: Uniting the parameters sounds confusing from an end user's perspective. Should this feature be moved out into a separate function? What should it be called? Something like `quantity_unify()`?

sympy/physics/units/util.py
dyc3: I tried filtering units by scale factors, and I was getting some weird results. I'll push a new commit to show my current solution momentarily.

sympy/physics/units/util.py
Upabjojr: > Uniting the parameters sounds confusing from an end user's perspective.

Probably, it was just an idea to avoid having too many parameters.

oscargus:
These are the current errors in code-quality (in case you missed them):

```
 sympy/physics/units/unitsystem.py:13:1: F401 'sympy.physics.units.prefixes.BIN_PREFIXES' imported but unused
sympy/physics/units/util.py:6:1: F401 'typing' imported but unused
```

sympy/physics/units/util.py
```diff
@@ -154,6 +158,45 @@ def quantity_simplify(expr):
...
+
+        from . import volt, ohm, farad, second, henry, meter, weber, siemens, tesla
+        from . import voltage, impedance, capacitance, inductance, magnetic_flux, conductance, magnetic_density, velocity
+
+        non_prefixed = unit_system.get_units_non_prefixed()
+        preferred_dimension_to_unit = { unit.dimension.name: unit for unit in non_prefixed if unit.dimension.name.is_symbol }
+
+        # HACK: add missing/wrong units to the map.
+        if unit_system.name == "SI":
+            preferred_dimension_to_unit.update({
```
Upabjojr, line 184: what about this:

```python
from sympy.physics.units import definitions

preferred_dimension_to_unit = {}

for name in dir(definitions):
    quantity = getattr(definitions, name)
    if not isinstance(quantity, Quantity):
        continue
    dim = unit_system.get_quantity_dimension(quantity)
    cc = convert_to(quantity, unit_system._base_units)
    if cc.as_coeff_Mul()[0] == 1:
        preferred_dimension_to_unit[dim.name] = quantity
```

dyc3: Unfortunately, this method appears to be more correct, but it absolutely kills performance. The unit test takes noticeably longer. Also, the velocity test case fails using this method. I think we can make something like this work, if we compute it beforehand. 

It's really annoying that there are all these prefixed units like `picojoule` and `teranewton`. IMO, they should not exist, we already have the prefixes defined using `Prefix`, that should be sufficient.

dyc3: Ultimately, I think that this is going to need some metadata changes, and _maybe_ breaking changes.

Upabjojr: What about using this method as a code generator? That dictionary needs to be computed only once in the end... Maybe when you install SymPy or when you use it the first time?

dyc3: I think that doing code gen is going to be harder than it sounds. An LRU cache seems like a better idea.

Upabjojr: > If an application using sympy were to allow for unescaped user input, it could be possible to end up saving and executing malicious code.

Well, the code generator should not be called by end users, but by the developers before releasing a new version of SymPy.

> Regardless, I've already implemented 2 new properties for Quantity: `is_prefixed` and `is_physics_constant`. With these, I'm pretty sure the call to `convert_to()` will not be necessary, and none of this will be required. I'll commit a new implementation shortly.

That's great to hear :)

sympy/physics/units/quantities.py
```diff
@@ -63,13 +64,19 @@ def __new__(cls, name, abbrev=None, dimension=None, scale_factor=None,
...
+
         obj = AtomicExpr.__new__(cls, name, abbrev)
         obj._name = name
         obj._abbrev = abbrev
         obj._latex_repr = latex_repr
         obj._unicode_repr = pretty_unicode_repr
         obj._ascii_repr = pretty_ascii_repr
         obj._mathml_repr = mathml_presentation_repr
+        obj._is_prefixed = is_prefixed
+        obj._is_physics_constant = is_physics_constant
```
Upabjojr, line 79: Be careful that this information is lost in case you call `Quantity(*quantity_variable.args)` or `Quantity.func(*quantity_variable.args)`, because `is_prefixed` and `is_physics_constant` are not passed to the superclass as arguments.

This could be an issue if you call some SymPy functionality like the term rewriter or the substitution.

sympy/physics/units/tests/test_quantities.py
```diff
@@ -510,3 +512,32 @@ def test_issue_22164():
 
     # deprecation warning is not expected here
     SI._collect_factor_and_dimension(expr)
+
+def test_prefixed_flag():
```
Upabjojr, line 516: Maybe `test_prefixed_property`, I wouldn't call it a `flag` as it's supposed to be a constant.

sympy/physics/units/quantities.py
Upabjojr: What about subclassing `Quantity` into a new class `Constant` just defining `is_constant` as a static field?
e.g.
```python
class Quantity:
    is_physics_constant = False
    ...

class Constant(Quantity):
    is_physics_constant = True
```

dyc3: > Be careful that this information is lost

Could you give me a test case for this? I'm not familiar enough with this SymPy behavior.

> What about subclassing `Quantity`

That seems like a better idea.

sympy/physics/units/quantities.py
```diff
@@ -232,3 +241,17 @@ def convert_to(self, other, unit_system="SI"):
...
+
+    @property
+    def is_prefixed(self):
+        """Whether or not the quantity is prefixed. Eg. `kilogram` is prefixed, but `gram` is not."""
+        return self._is_prefixed
+
+class PhysicalConstant(Quantity):
+    """Represents a physical constant, eg. `speed_of_light` or `avogadro_constant`."""
+
+    is_commutative = True
```
Upabjojr, line 253: aren't these already inherited from the superclass?

dyc3: Ah yeah, they are. Fixed.

sympy/physics/units/systems/si.py
```diff
@@ -56,6 +56,7 @@
 for u in units:
     all_units.extend(prefix_unit(u, PREFIXES))
 
+all_units.extend(units)
```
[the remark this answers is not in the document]
dyc3: No, it's not because the output of `prefix_unit()` does not contain `u` without a prefix. It's very weird.

sympy/physics/units/util.py
```diff
@@ -120,21 +123,43 @@ def get_total_scale_factor(expr):
...
 
-def quantity_simplify(expr):
+def _build_preferred_dimension_to_unit(unit_system="SI") -> tDict[Symbol, Quantity]:
+    from sympy.physics.units import  velocity, length, time
+
+    unit_system = UnitSystem.get_unit_system(unit_system)
+
+    preferred_dimension_to_unit = {}
+
+    for quantity in unit_system.get_units_non_prefixed():
```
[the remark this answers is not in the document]
dyc3: You bring up a good point. The purpose of this is to build the lookup table using base units, because otherwise it'll end up picking units like `picometer` or `picojoule`. As a result, `gram` will be picked over `kilogram`, because `kilogram` is prefixed.

I think it would be pretty feasible to grab the degree of the resulting number, and select the closest prefix. However, I feel like it could lead to unexpected behavior. It's extra complexity to debug later, and it shouldn't go in this function. In the interest of trying to keep this PR small, I was going to follow up this PR with another to add a new function or optional arg to this function that does this.

sympy/physics/units/tests/test_quantities.py
```diff
@@ -510,3 +512,33 @@ def test_issue_22164():
...
+def test_physics_constant():
+    from sympy.physics.units import definitions
+
+    for name in dir(definitions):
+        quantity = getattr(definitions, name)
+        if not isinstance(quantity, Quantity):
+            continue
+        if name.endswith('_constant'):
+            assert isinstance(quantity, PhysicalConstant), f"{quantity} must be PhysicalConstant, but is {type(quantity)}"
+            assert quantity.is_physics_constant, f"{name} is not marked as physics constant when it should be"
```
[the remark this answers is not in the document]
dyc3: Yes. It helps enforce that any new constants that get added will get added correctly, using `PhysicalConstant`. If you are worried about unit test performance, there are many more egregious tests than this one.

sympy/physics/units/util.py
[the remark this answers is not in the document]
Upabjojr: Maybe the preferred units for a given unit system should just be hard-coded into a dictionary in the end?

Upabjojr:
By the way, why are you modifying the authors fail? Usually no one does it in a PR.

dyc3:
I'm not? I only updated the mailmap file, as requested above. Idk why test/authors is still failing.

sympy/physics/units/util.py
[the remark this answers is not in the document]
dyc3: But then it won't be unit system agnostic.

.mailmap
```diff
@@ -122,8 +122,8 @@ Alan Bromborsky <abrombo@verizon.net> <brombo@GA.(none)>
 Alec Kalinin <alec.kalinin@gmail.com>
 Aleksandar Makelov <amakelov@college.harvard.edu>
 Alex Argunov <sajkoooo@gmail.com>
-Alex Lindsay <adlinds3@ncsu.edu> <lindsayad@users.noreply.github.com>
 Alex Lindsay <adlinds3@ncsu.edu> Alex Lindsay <alexlindsay239@gmail.com>
+Alex Lindsay <adlinds3@ncsu.edu> <lindsayad@users.noreply.github.com>
```
oscarbenjamin, line 126: I'm not sure why these kinds of changes sometimes happen. I guess you've used the `mailmap_update.py` script rather than the `mailmap_check.py` script but I can't be sure. Do you know which script you ran?

The authors check is failing and says this:
```
Carson McManus <carson.mcmanus1@gmail.com>
Carson McManus <dyc3@users.noreply.github.com>

The .mailmap file needs to be updated because there are commits with
unrecognised author/email metadata.
The mailmap file was reordered

```
The lines that you should add to the .mailmap file would be something like:
```
Carson McManus <carson.mcmanus1@gmail.com> Carson McManus <dyc3@users.noreply.github.com>
```
That line means that `Carson McManus` with the gmail address is a listed author and that any commits with the `dyc3` email address should be associated with the same author.

After adding that line you should run the `mailmap_check.py` script twice: first time reorders the file and the second time should confirm that the file is now up to date.

Upabjojr:
> I'm not? I only updated the mailmap file, as requested above. Idk why test/authors is still failing.

I think that script is usually run only 2 or 3 times per year, before the release of a new SymPy versions.

.mailmap
dyc3: Ah yeah I did use the `mailmap_update.py` script. Seems like this process needs some better instructions haha

Upabjojr:
In reality the issue of the difference between physical constants and units is a lot more subtle. You can define a unit system with physical constants as units, so maybe it does not make much sense to have different classes for `Quantity` and `PhysicalConstant`. In the end, everything is just a `Quantity`... it's just a convention to use some as units and some others as physical constants. Maybe the class `PhysicalConstant` does not make that much sense.

I'm not sure how the issue of simplification across dimensions could be resolved, maybe with a dictionary of preferred units associated to the unit system? Maybe using a complexity score for the expression as used by SymPy's `simplify( )`?

dyc3:
I'm a little confused. I'm not entirely sure what is blocking this, (besides the obvious merge conflicts and mailmap stuff). I think that this implementation is pretty good, and I really don't want this PR to sit and collect dust.

Upabjojr:
> I'm a little confused. I'm not entirely sure what is blocking this, (besides the obvious merge conflicts and mailmap stuff). I think that this implementation is pretty good, and I really don't want this PR to sit and collect dust.

I think that units and physical constants should be defined inside the `UnitSystem` class, not globally. That is the problem I see with the current implementation.

dyc3:
Rebased and added preferred units as hard coded dict, controlled by the `UnitSystem` class.

I also took the liberty of renaming some imports to be more explicit and clear in some of the unit system definitions.

sympy/physics/units/systems/mks.py
```diff
@@ -15,19 +15,30 @@
...
+all_units.extend([gram, joule, newton, watt, pascal, hertz])
 
 for u in units:
     all_units.extend(prefix_unit(u, PREFIXES))
-all_units.extend([G, c])
+all_units.extend([gravitational_constant, speed_of_light])
 
 # unit system
-MKS = UnitSystem(base_units=(m, kg, s), units=all_units, name="MKS", dimension_system=dimsys_length_weight_time)
+MKS = UnitSystem(base_units=(meter, kilogram, second), units=all_units, name="MKS", dimension_system=dimsys_length_weight_time, preferred_units={
```
Upabjojr, line 30: what about calling the dict _derived units_ instead of _preferred units_? Indeed, they are the [officially derived units](https://en.wikipedia.org/wiki/SI_derived_unit).

Upabjojr:
One doubt: are physical constants really physical constants in all unit systems or do they depend on the chosen unit system?

For example, the speed of light is equal to 1 and dimensionless in natural units. Can it still be considered a constant?

What about a unit system defined using Planck units as base units? A lot of physical constants would turn out to be units instead of constants.

dyc3:
According to the [Physical Constant](https://en.wikipedia.org/wiki/Physical_constant) article on wikipedia:

> A physical constant, sometimes fundamental physical constant or universal constant, is a [physical quantity](https://en.wikipedia.org/wiki/Physical_quantity) that is generally believed to be both universal in nature and have [constant](https://en.wikipedia.org/wiki/Constant_(mathematics)) value in time. It is contrasted with a [mathematical constant](https://en.wikipedia.org/wiki/Mathematical_constant), which has a fixed numerical value, but does not directly involve any physical measurement.

Under this definition, IMO, the speed of light is correctly defined as a fundamental physical constant, and should be denoted as such. 

Looking at our current definition for the natural unit system, I'm not completely sure if it's correct. I've filed #23261 as it is out of the scope of this PR.

sympy/physics/units/systems/mks.py
dyc3: Sounds good, will do.

dyc3:
Looking for one last review, and it should be ready to merge in.

sympy/physics/units/unitsystem.py
```diff
@@ -192,3 +199,9 @@ def _collect_factor_and_dimension(self, expr):
             return S.One, expr
         else:
             return expr, Dimension(1)
+
+    def get_units_non_prefixed(self) -> tSet[Quantity]:
```
Upabjojr, line 203: Maybe this could be private, i.e. start with an underscore `_` character. Do end-users usually need to access this method?

sympy/physics/units/unitsystem.py
```diff
@@ -26,7 +28,7 @@ class UnitSystem(_QuantityMapper):
 
     _unit_systems = {}  # type: tDict[str, UnitSystem]
 
-    def __init__(self, base_units, units=(), name="", descr="", dimension_system=None):
+    def __init__(self, base_units, units=(), name="", descr="", dimension_system=None, derived_units: tDict[Symbol, Quantity]={}):
```
Upabjojr, line 31: What about mapping `tDict[Dimension, Quantity]` instead of the name of the dimension?

Upabjojr:
OK, just one doubt about the implicit unit system. `quantity_simplify` now implicitly assumes SI units, this is dangerous. I would make the `unit_system` parameter mandatory in case `across_dimensions=True`. I would suggest to specify `unit_system=None` and then check if it is specified when `across_dimensions=True`. The purpose of the units module is to be independent of the chosen unit system, although some historical code is still implicitly assuming SI as the preferred unit system.

The problem is that once a new SymPy version will be released, we are not supposed to change the API anymore (unless you go to some lengthy deprecation cycle). It's therefore convenient to take some time and reason about a good API before it is too late to change it.

sympy/physics/units/unitsystem.py
dyc3: That might work, I'm not sure if `Dimension` can be used as a dictionary key.

sympy/physics/units/unitsystem.py
Upabjojr: OK.

Upabjojr:
I have merged this PR as it is now good enough to be included. 

Thanks for your contribution @dyc3 !

Issue #20502, opened 2020-11-28
We shouldn't use .. [1], .. [2] for references 


sylee957:
https://docs.sympy.org/latest/documentation-style-guide.html#references-section

I think that this causes problems in migration to napoleon builder because it throws lots of duplicate reference errors.

Although writing references like `.. [1], .. [2]` looks fine solely in the docstring viewpoint, I see it causes problems for duplicate references maybe because it's used throughout all the sympy codebase or the documentation could be concatenated to single rst page technically.

The reason that our numpydoc builder runs fine is that it have some automatic renumbering of references
https://github.com/sympy/sympy/blob/8ec76cc615f73c3a525b9bd4b1f1a8c925dbdc06/doc/ext/numpydoc.py#L66-L86
But napoleon lacks that capability and I think that napoleon may expect to use `[#]` or `[*]` instead of `[1], [2]`
https://docutils.sourceforge.io/docs/user/rst/quickref.html#footnotes

Although we can use quick search-and-replace changes for changing this throughout all sympy codebase, I saw some sympy documentation is using those as actual symbolic references such as, so it's not always a good idea to search and replace whole  `[1]` to `[#]`
```
    Find the Euler-Lagrange equations [1]_ for a given Lagrangian.

    Parameters
    ==========
```
So it's better to carefully identify what kind of things should better be renamed to `[foo]_` and what kind of things should be replaced with `[#]`.

Anantha-Rao12:
Hey! I'll be glad to give this issue a try and help out. Any suggestions on how I should proceed and improve the references section?
Thanks

moorepants:
I'd start by adding this to the documentation style guide. Once the recommendation is solidified there, you can start changing everything to be consistent.

Anantha-Rao12:
Working on it! Thanks, @moorepants !

sylee957:
It’s just one of my speculation, but before we write any documentation tutorial, we should check which one ([#] or [*]) works better for dummy references to use throughout sympy by testing out a single file and running sphinx.

PR #23232, opened 2022-03-13
Improved some documentation related to the stats module

### LINKED ISSUE #23136: Stats Module Documentation
I think there is a mistake in stats module documentation on this part  `BetaNoncentral` CRV, since it's not showing code at that place properly.
Sympy stats doc

ghost:
Improved documentation in the stats module

Fixes #23136 

1.  In the parameters, spaces were missing due to which the whole sentence and the parameter name was appearing in bold letters therefore a space is added now at most of the places.
2.  In the docstrings for some symbols like mu it was mentioned with `'mu'` due to which it appeared the same way on the documentation therefore that was updated so that the `Symbol` can be seen.
3. Some characters were missing like`\lambda \ge 0` due to which it was not appearing as it should be, here it should be `\lambda \geq 0` therefore added now.
4. Parameters and Notes were not mentioned properly in joint_rv_types.py so added now.
5. Sentence to use python `as` keyword as a Note in GeneralizedMultivariateLogGamma, Omega is now converted into a codeblock

#### Other comments

- [x] Opened on a seperate branch.
- [ ] Has unit tests.
- [ ] Has Release Notes.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/stats/crv_types.py
+++ b/sympy/stats/crv_types.py
@@ -516,7 +516,7 @@ def BetaNoncentral(name, alpha, beta, lamda):
 
     alpha : Real number, `\alpha > 0`, a shape
     beta : Real number, `\beta > 0`, a shape
-    lamda: Real number, `\lambda \ge 0`, noncentrality parameter
+    lamda : Real number, `\lambda \geq 0`, noncentrality parameter
 
     Returns
     =======
@@ -671,7 +671,7 @@ def BoundedPareto(name, alpha, left, right):
     Parameters
     ==========
 
-    alpha : Real Number, `alpha > 0`
+    alpha : Real Number, `\alpha > 0`
         Shape parameter
     left : Real Number, `left > 0`
         Location parameter
@@ -1244,9 +1244,9 @@ def ExGaussian(name, mean, std, rate):
     ==========
 
     mu : A Real number, the mean of Gaussian component
-    std: A positive Real number,
+    std : A positive Real number,
         :math: `\sigma^2 > 0` the variance of Gaussian component
-    lambda: A positive Real number,
+    lambda : A positive Real number,
         :math: `\lambda > 0` the rate of Exponential component
 
     Returns
@@ -2006,8 +2006,8 @@ def Gumbel(name, beta, mu, minimum=False):
     Parameters
     ==========
 
-    mu : Real number, '\mu', a location
-    beta : Real number, '\beta > 0', a scale
+    mu : Real number, `\mu`, a location
+    beta : Real number, `\beta > 0`, a scale
     minimum : Boolean, by default ``False``, set to ``True`` for enabling minimum distribution
 
     Returns
@@ -2077,13 +2077,13 @@ def Gompertz(name, b, eta):
     .. math::
         f(x) := b \eta e^{b x} e^{\eta} \exp \left(-\eta e^{bx} \right)
 
-    with :math: 'x \in [0, \inf)'.
+    with :math:`x \in [0, \infty)`.
 
     Parameters
     ==========
 
-    b: Real number, 'b > 0', a scale
-    eta: Real number, '\eta > 0', a shape
+    b : Real number, `b > 0`, a scale
+    eta : Real number, `\eta > 0`, a shape
 
     Returns
     =======
@@ -3485,7 +3485,7 @@ def PowerFunction(name, alpha, a, b):
     Parameters
     ==========
 
-    alpha: Positive number, `0 < alpha`, the shape paramater
+    alpha : Positive number, `0 < \alpha`, the shape paramater
     a : Real number, :math:`-\infty < a`, the left boundary
     b : Real number, :math:`a < b < \infty`, the right boundary
 
@@ -3883,13 +3883,13 @@ def ShiftedGompertz(name, b, eta):
     .. math::
         f(x) := b e^{-b x} e^{-\eta \exp(-b x)} \left[1 + \eta(1 - e^(-bx)) \right]
 
-    with :math: 'x \in [0, \inf)'.
+    with :math:`x \in [0, \infty)`.
 
     Parameters
     ==========
 
-    b: Real number, 'b > 0', a scale
-    eta: Real number, '\eta > 0', a shape
+    b : Real number, `b > 0`, a scale
+    eta : Real number, `\eta > 0`, a shape
 
     Returns
     =======
@@ -4625,7 +4625,7 @@ def WignerSemicircle(name, R):
     Returns
     =======
 
-    A `RandomSymbol`.
+    A RandomSymbol.
 
     Examples
     ========
--- a/sympy/stats/crv_types.py
+++ b/sympy/stats/crv_types.py
@@ -551,7 +551,8 @@ def BetaNoncentral(name, alpha, beta, lamda):
     /____,
     k = 0
 
-    Compute cdf with specific 'x', 'alpha', 'beta' and 'lamda' values as follows :
+    Compute cdf with specific 'x', 'alpha', 'beta' and 'lamda' values as follows:
+
     >>> cdf(BetaNoncentral("x", 1, 1, 1), evaluate=False)(2).doit()
     2*exp(1/2)
 
@@ -1243,10 +1244,11 @@ def ExGaussian(name, mean, std, rate):
     Parameters
     ==========
 
-    mu : A Real number, the mean of Gaussian component
+    name : A string giving a name for this distribution
+    mean : A Real number, the mean of Gaussian component
     std : A positive Real number,
         :math: `\sigma^2 > 0` the variance of Gaussian component
-    lambda : A positive Real number,
+    rate : A positive Real number,
         :math: `\lambda > 0` the rate of Exponential component
 
     Returns
--- a/sympy/stats/joint_rv_types.py
+++ b/sympy/stats/joint_rv_types.py
@@ -54,8 +54,8 @@ def marginal_distribution(rv, *indices):
     Parameters
     ==========
 
-    rv: A random variable with a joint probability distribution.
-    indices: component indices or the indexed random symbol
+    rv : A random variable with a joint probability distribution.
+    indices : Component indices or the indexed random symbol
         for which the joint distribution is to be calculated
 
     Returns
@@ -100,11 +100,18 @@ def JointRV(symbol, pdf, _set=None):
     Create a Joint Random Variable where each of its component is continuous,
     given the following:
 
-    -- a symbol
-    -- a PDF in terms of indexed symbols of the symbol given
-    as the first argument
+    Parameters
+    ==========
 
-    NOTE: As of now, the set for each component for a ``JointRV`` is
+    symbol : Symbol
+        Represents name of the random variable.
+    pdf : a PDF in terms of indexed symbols of the symbol given
+        as the first argument
+
+    NOTE
+    ====
+
+    As of now, the set for each component for a ``JointRV`` is
     equal to the set of all integers, which cannot be changed.
 
     Examples
@@ -198,7 +205,7 @@ def MultivariateNormal(name, mu, sigma):
 
     mu : List representing the mean or the mean vector
     sigma : Positive semidefinite square matrix
-        Represents covariance Matrix
+        Represents covariance Matrix.
         If `\sigma` is noninvertible then only sampling is supported currently
 
     Returns
@@ -358,11 +365,11 @@ def MultivariateT(syms, mu, sigma, v):
     Parameters
     ==========
 
-    syms: A symbol/str
+    syms : A symbol/str
         For identifying the random variable.
-    mu: A list/matrix
+    mu : A list/matrix
         Representing the location vector
-    sigma: The shape matrix for the distribution
+    sigma : The shape matrix for the distribution
 
     Examples
     ========
@@ -435,15 +442,15 @@ def NormalGamma(sym, mu, lamda, alpha, beta):
     Parameters
     ==========
 
-    sym: A symbol/str
+    sym : A symbol/str
         For identifying the random variable.
-    mu: A real number
+    mu : A real number
         The mean of the normal distribution
-    lamda: A positive integer
+    lamda : A positive integer
         Parameter of joint distribution
-    alpha: A positive integer
+    alpha : A positive integer
         Parameter of joint distribution
-    beta: A positive integer
+    beta : A positive integer
         Parameter of joint distribution
 
     Returns
@@ -506,7 +513,7 @@ def MultivariateBeta(syms, *alpha):
     Parameters
     ==========
 
-    alpha: Positive real numbers
+    alpha : Positive real numbers
         Signifies concentration numbers.
 
     Returns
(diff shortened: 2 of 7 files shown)


KuldeepBorkar:
- Due to flake8 check I guess I should not convert the below line to a code block or should I skip the test for that line sympy.stats.joint_rv_types.GeneralizedMultivariateLogGamma as GMVLG
I thought it would be good if the line is shown in a code block.
- I saw on the documentation in most of the places the references are given as "can be found at [1]"
but when we check the reference section it is mentioned as some reference ID like [R902] rather than [1] I suppose it should be changed too but I just don't know the proper way to do that.
I was thinking of opening an issue for that but asking here first would be good I guess.
- At some places it is `lamda` is mentioned I think it should be `lambda`, can someone tell if this was done for a reason or probably just a minor mistake. Symbol was not shown on the documentation for lambda therefore, I was trying to change that to `lambda` but in the code the word `lamda` is used instead therefore before changing the whole thing I wanted to ask here first.
- Some more work in the same way is needed to be done then it will be all good, is it okay to push more changed in this `PR`, 
I thought it would be time taking and annoying for reviewers to review so much changes at the same place therefore I should ask first.
- In `rv.py` even after adding a space after `..` in `..math::` I think the line used here is not proper I think its written in latex which I don't know how exactly it should be written there it would be good if someone helps with that.
 .. math::
        Q(p) = inf{x \in (-\infty, \infty) such that p <= F(x)}

Any thoughts and reviews regarding the above points and changes would be helpful : )

sympy/stats/crv_types.py
```diff
@@ -1244,9 +1244,9 @@ def ExGaussian(name, mean, std, rate):
     ==========
 
     mu : A Real number, the mean of Gaussian component
-    std: A positive Real number,
+    std : A positive Real number,
```
oscargus, line 1247: There seems to be some inconsistency between the function and the parameters described. While not caused by you, feel free to correct it. Function uses `name, mean, std, rate` so `std` is the only one actually matching the documentation.

sympy/stats/crv_types.py
```diff
@@ -516,7 +516,7 @@ def BetaNoncentral(name, alpha, beta, lamda):
 
     alpha : Real number, `\alpha > 0`, a shape
     beta : Real number, `\beta > 0`, a shape
-    lamda: Real number, `\lambda \ge 0`, noncentrality parameter
+    lamda : Real number, `\lambda \geq 0`, noncentrality parameter
```
oscargus, line 519: \ge and \geq are the same. So any goes, but no need to modify and more of these.

sympy/stats/crv_types.py
```diff
@@ -3883,13 +3883,13 @@ def ShiftedGompertz(name, b, eta):
     .. math::
         f(x) := b e^{-b x} e^{-\eta \exp(-b x)} \left[1 + \eta(1 - e^(-bx)) \right]
 
-    with :math: 'x \in [0, \inf)'.
+    with :math:`x \in [0, \infty)`.
```
oscargus, line 3886: For info: in this situation there is no need to write `:math:` in front as SymPy will default single backtick to equation. (It is also possible to use $ instead of single backtick.)

(No need to update this.)

sympy/stats/crv_types.py
```diff
@@ -4625,7 +4625,7 @@ def WignerSemicircle(name, R):
     Returns
     =======
 
-    A `RandomSymbol`.
+    A RandomSymbol.
```
oscargus, line 4628: An option is to change this to 
```
A :class:`~.RandomSymbol`.
```
which should provide a clickable link to the RandomSymbol class (subject to that I do not remember the exact syntax...).

But this change is clearly more correct than it was.

sympy/stats/crv_types.py
KuldeepBorkar: What could be the possible reason that on the documentation in `BetaNonCentral` parameter it shows "lamda: Real number, `lambda >= 0`, noncentrality parameter"
And not the `symbol` for lambda?

After updating this with `geq` I was able to see `symbol` for lambda.

sympy/stats/joint_rv_types.py
```diff
@@ -712,8 +719,10 @@ def GeneralizedMultivariateLogGamma(syms, delta, v, lamda, mu):
     ====
 
     If the GeneralizedMultivariateLogGamma is too long to type use,
-    ``from sympy.stats.joint_rv_types import GeneralizedMultivariateLogGamma as GMVLG``
-    If you want to pass the matrix omega instead of the constant delta, then use,
+
+    >>> from sympy.stats.joint_rv_types import GeneralizedMultivariateLogGamma as GMVLG
+
+    If you want to pass the matrix omega instead of the constant delta, then use
```
oscargus, line 725: This was discussed when I updated parts of the documentation and there were opinions that it maybe was not really relevant. However, now that you have made it a doctest, one should possibly go ahead and also use GMVLG to define a variable.

sympy/stats/joint_rv_types.py
```diff
@@ -100,11 +100,18 @@ def JointRV(symbol, pdf, _set=None):
...
 
-    -- a symbol
-    -- a PDF in terms of indexed symbols of the symbol given
-    as the first argument
+    Parameters
+    ==========
+
+    symbol : Symbol
+        Represents name of the random variable.
+    pdf : a PDF in terms of indexed symbols of the symbol given
```
oscargus, line 108:
```suggestion
    pdf : A PDF in terms of indexed symbols of the symbol given
```

sympy/stats/joint_rv_types.py
```diff
@@ -823,9 +833,9 @@ def Multinomial(syms, n, *p):
     Parameters
     ==========
 
-    n: Positive integer
+    n : Positive integer
         Represents number of trials
-    p: List of event probabilites
+    p : List of event probabilites
         Must be in the range of [0, 1]
```
oscargus, line 839:
```suggestion
        Must be in the range of $[0, 1]$.
```

sympy/stats/stochastic_process_types.py
```diff
@@ -1746,10 +1746,10 @@ def probability(self, condition, given_condition=None, evaluate=True, **kwargs):
         Parameters
         ==========
 
-        condition: Relational
+        condition : Relational
                 Condition for which probability has to be computed. Must
                 contain a RandomIndexedSymbol of the process.
-        given_condition: Relational/And
+        given_condition : Relational/And
```
oscargus, line 1752:
```suggestion
        given_condition : Relational, Logic
```
As on line 1730.

sympy/stats/stochastic_process_types.py
```diff
@@ -1627,12 +1627,12 @@ class BernoulliProcess(DiscreteTimeStochasticProcess):
     Parameters
     ==========
 
-    sym: Symbol/str
-    success: Integer/str
+    sym : Symbol/str
+    success : Integer/str
             The event which is considered to be success, by default is 1.
```
oscargus, line 1632:
```suggestion
            The event which is considered to be success. Default: 1.
```

sympy/stats/stochastic_process_types.py
```diff
@@ -1627,12 +1627,12 @@ class BernoulliProcess(DiscreteTimeStochasticProcess):
...
     failure: Integer/str
             The event which is considered to be failure, by default is 0.
```
oscargus, line 1634:
```suggestion
            The event which is considered to be failure. Default: 0.
```

sympy/stats/stochastic_process_types.py
oscargus: Or maybe it should be `Relational, And`? Not sure if e.g. `Or` works?

sympy/stats/rv.py
```diff
@@ -1192,7 +1192,7 @@ def quantile(expr, evaluate=True, **kwargs):
     Quantile is defined as the value at which the probability of the random
     variable is less than or equal to the given probability.
 
-    ..math::
+    .. math::
         Q(p) = inf{x \in (-\infty, \infty) such that p <= F(x)}
```
oscargus, line 1196:
```suggestion
        Q(p) = \max{x \in (-\infty, \infty) : p \le F(x)}
```
I think is slightly better, although maybe not completely correct.

oscargus:
Some comments, but overall I think this is a good

> * Due to flake8 check I guess I should not convert the below line to a code block or should I skip the test for that line sympy.stats.joint_rv_types.GeneralizedMultivariateLogGamma as GMVLG
>   I thought it would be good if the line is shown in a code block.

One can probably remove the whole thing, but if you want to keep it, I think that adding the new abbreviation may be enough (and helpful).

> * I saw on the documentation in most of the places the references are given as "can be found at [1]"
>   but when we check the reference section it is mentioned as some reference ID like [R902] rather than [1] I suppose it should be changed too but I just don't know the proper way to do that.
>   I was thinking of opening an issue for that but asking here first would be good I guess.

What you can do is to change [1] to [1]_ where it is references. Then it will update to the given number. There is a bit of an issue using numbered references like that, it can be better to name them something, but most of the time we just use numbers and then leave it to sphinx to sort it out. If it works, it is OK (but can be improved, although that would require changing all of the SymPy documentation...).

> * At some places it is `lamda` is mentioned I think it should be `lambda`, can someone tell if this was done for a reason or probably just a minor mistake. Symbol was not shown on the documentation for lambda therefore, I was trying to change that to `lambda` but in the code the word `lamda` is used instead therefore before changing the whole thing I wanted to ask here first.

Since `lambda` is a Python keyword, the incorrect spelling `lamda` is used to avoid potential errors. So if anything, any class that use `lambda` should be changed to `lamda` (although still `\lambda` is used for equations etc.).

> * Some more work in the same way is needed to be done then it will be all good, is it okay to push more changed in this `PR`,
>   I thought it would be time taking and annoying for reviewers to review so much changes at the same place therefore I should ask first.

There are two aspects here: 
1. It is easier for a reviewer to see the changes if you add another commit.
2. We want to avoid getting too many commits and right now we do not have the option to squeeze at merge time.

So ideally, you add a new commit with changes from the review comments (or anything else that is not reviewed) and at some stage, like when fixing the last final things, you do a rebase, squash them and force push. (Although this is currently a subject of debate among the developers.) But as long as there are only a few commits it is not a big deal, more like if there are tens of commits, some very minor.

> * In `rv.py` even after adding a space after `..` in `..math::` I think the line used here is not proper I think its written in latex which I don't know how exactly it should be written there it would be good if someone helps with that.
>   .. math::
>   Q(p) = inf{x \in (-\infty, \infty) such that p <= F(x)}

I gave a suggestion, not sure how "such that" really should be written, but the rest is OK (I think). At least better than the current one...

sympy/stats/crv_types.py
oscargus: OK! I wouldn't expect that to be the reason (more that the space enables equations in the description, without a space everything will be the "parameter name" where one maybe cannot have equations).

If the official docmentation shows that, it can be that I had a similar PR (although clearly still a bit sloppy) that changed it to \ge some time ago, that probably didn't make it into 1.10. If you look at the dev-documentation it says `lambda ge 0` https://docs.sympy.org/dev/modules/stats.html?highlight=betanoncentral#sympy.stats.BetaNoncentral (which is caused by the lack of space before :).

oscargus: If you can add an empty line before line 555 that would also be nice. (I cannot mark there directly, but that line does not show up as code.)

sympy/stats/rv.py
KuldeepBorkar: @czgdp1807 Sir, Could you please confirm this once?

sympy/stats/crv_types.py
KuldeepBorkar: I don't know much about the `dev-documentation` is it a bit different like, are there some changes which are different from here?

KuldeepBorkar:
> Since `lambda` is a Python keyword, the incorrect spelling `lamda` is used to avoid potential errors. So if anything, any class that use `lambda` should be changed to `lamda` (although still `\lambda` is used for equations etc.).

There are some places where `Lambda` is used to avoid use of keyword then I think its okay to use lambda `symbol` in the docstring currently I saw at one place its `lamda > 0` so, I will just convert that.

> What you can do is to change [1] to [1]_ where it is references. Then it will update to the given number. There is a bit of an issue using numbered references like that, it can be better to name them something, but most of the time we just use numbers and then leave it to sphinx to sort it out. If it works, it is OK (but can be improved, although that would require changing all of the SymPy documentation...).

Is that okay if I just open a seperate issue for this? It's quite problematic for me I tried at first but then at most of the places [1] as a reference is used again and again that makes it quite difficult.
One solution to this can be that we can use number the references from starting in order like [1] [2] [3]........... in sequence from starting.
But that would be some large work to do for the whole stats documentation here.
Instead if there is time someone could possibly do this and work on it as a seperate issue later on.

KuldeepBorkar:
@oscargus I was not able to find why the code-quality check is failing I run the tests locally and were OK but here it says "detected invalid escape like \e" can you please point out?

oscargus:
> Is that okay if I just open a seperate issue for this?

There is already one: https://github.com/sympy/sympy/issues/20502

But you can for sure skip it here.

KuldeepBorkar:
> You will either further esdape the , so \gamma or add an r in front of the string (which does not support escaped characters like \n), so r""" on line 2252.

It says in file stochastic_process_types.py
there, on line 2252 I have not made any changes so how?
or its 2311?

oscargus:
You are right. 2311 it is. I looked a bit too fast and took the wrong doc-string...

sympy/stats/rv.py
oscargus: I checked https://en.wikipedia.org/wiki/Quantile_function and it looks correct, except that it should be \min (or \operatorname{inf}), not \max.

KuldeepBorkar: `Q(p) = \min\{x \in (-\infty, \infty) : p \le F(x)\}`
or 
`Q(p) = \inf\{x \in (-\infty, \infty) : p \le F(x)\}`
So, the above line right?
I was not knowing the operator inf or min refers to the same thanks : )

sympy/stats/crv_types.py
KuldeepBorkar: Here I changed that `inf` to `infty` was that correct I am not sure.

sympy/stats/rv.py
oscargus: Not sure that \inf works, but if it does then it should be OK.

sympy/stats/crv_types.py
oscargus: Yes. \inf is the infimum function and \infty is the infinity symbol. (I checked it, and, yes, \inf does exist. Regarding my comment below.)

sympy/stats/crv_types.py
oscargus: The dev documentation is for the head of the master branch. So when modifications are made it will only be shown in dev until a release is made. (Right now 1.10 is the "official" documentation, your changes here will show up in 1.11, but until then one can see them in the dev documentation.)

oscargus:
Looks good! Thanks! Check out the dev documentation tomorrow or so and you will see your changes in action.

sympy/stats/crv_types.py
KuldeepBorkar: Thank you : )
Now, I know the difference clearly.

PR #23234, opened 2022-03-13
Improve documentation

oscargus:

Improve the documentation. Especially, the intersphinx extension is used so that it is possible to link to other projects. So far, some links to matplotlib are used.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/plotting/plot.py
+++ b/sympy/plotting/plot.py
@@ -73,10 +73,10 @@ class Plot:
     Explanation
     ===========
 
-    For interactive work the function ``plot`` is better suited.
+    For interactive work the function :func:`plot()` is better suited.
 
     This class permits the plotting of SymPy expressions using numerous
-    backends (matplotlib, textplot, the old pyglet module for sympy, Google
+    backends (:mod:`matplotlib`, textplot, the old pyglet module for SymPy, Google
     charts api, etc).
 
     The figure can contain an arbitrary number of plots of SymPy expressions,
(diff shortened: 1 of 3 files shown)


sympy/plotting/plot.py
```diff
@@ -1213,32 +1217,32 @@ class BaseBackend:
...
-        processed and a list of instances of the `BaseSeries` class is created,
-        containing the necessary information to plot the expressions (eg the
-        expression, ranges, series name, ...). Eventually, these objects will
-        generate the numerical data to be plotted.
-    2. A Plot object is instantiated, which stores the list of series and the
+        processed and a list of instances of the :class:`BaseSeries` class is
+        created, containing the necessary information to plot the expressions
+        (e.g. the expression, ranges, series name, ...). Eventually, these
+        objects will generate the numerical data to be plotted.
+    2. A :class:`~.Plot` object is instantiated, which stores the list of series and the
```
ThePauliPrinciple, line 1224: Is this line now too long or does `:class:~` not count for the length?

sympy/plotting/plot.py
```diff
@@ -2216,41 +2223,44 @@ def plot3d(*args, show=True, **kwargs):
     Arguments
     =========
 
-    ``expr`` : Expression representing the function along x.
+    expr : Expression representing the function along x.
 
-    ``range_x``: (x, 0, 5), A 3-tuple denoting the range of the x
-    variable.
+    range_x : (x, 0, 5), A 3-tuple denoting the range of the x
+    variable,
```
ThePauliPrinciple, line 2229: on 2101 (in the new file) the tuple was placed at the end. I think it would be better to make this consistent. Why the change of `.` to `,` here?

oscargus: I moved it back and forgot to change back the , to .

Will think about how to write this. I am not really happy with any of the current approaches. (But want to see if the matplotlib is working.)

sympy/plotting/plot.py
oscargus: I do not think we enforce line length in SymPy and since this is text it should wrap anyway. 

But I probably missed re-wrapping the source anyway.

ThePauliPrinciple: I don't think it is forced either, but since I saw you had rewrapped it in other places, I assumed it was intended to be so.

sympy/plotting/plot.py
oscargus: I changed these to 

```
    range_x : (:class:`Symbol`, float, float)
        A 3-tuple denoting the range of the x variable, e.g. (x, 0, 5).
```

PR #23238, opened 2022-03-13
liealgebras/cartan_type: CartanType_generator class is no longer a Basic

ThePauliPrinciple:

Helps with #22445

This class does not make use of any of the properties of `Basic`,
so it should also not subclass it. It also seems to fail
`_test_args`. (the only class in `core/tests/test_args.py` that currently does so)

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/tests/test_args.py
+++ b/sympy/core/tests/test_args.py
@@ -18,7 +18,7 @@ from sympy.functions.elementary.exponential import (exp, log)
 from sympy.functions.elementary.miscellaneous import sqrt
 from sympy.functions.elementary.trigonometric import sin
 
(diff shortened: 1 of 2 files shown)


oscargus:
Should this even be a class? Would it be any difference it is was a function?

ThePauliPrinciple:
> Should this even be a class? Would it be any difference it is was a function?

Yes, some user who instantiates the class rather than uses the intermediate variable will get errors if we change it to a function (`CartanType_generator()(<string>)` vs `CartanType_generator(<string>)`). Otherwise, no I don't believe so.

oscargus:
Looks good to me!

PR #23241, opened 2022-03-15
Series/Limits: Fixed leading term method inconsistency for expressions involving variables in powers 

### LINKED ISSUE #23231: Sympy giving the wrong solution
Hello, 

While doing an example with 
```
f(x) =( 2^(1/x) -2^(-1/x))/( 2^(1/x) +2^(-1/x))
```
Where you want to see the left end-behavior and the right end-behavior for f(1/x). Now left end-behavior is -1 and right end-behavior is 1 but according to SymPy its results is equal to 1. 

```
julia> using PythonCall
 
julia> sympy = pyimport("sympy")
Python module: <module 'sympy' from '/Users/verzani/.julia/environments/v1.7/.CondaPkg/env/lib/python3.10/site-packages/sympy/__init__.py'>
 
julia> sympy.limit((2^x - 2^(-x))/(2^x + 2^(-x)), x, -sympy.oo)^C
 
julia> x = sympy.symbols("x")
Python Symbol: x
 
julia> sympy.limit((2^x - 2^(-x))/(2^x + 2^(-x)), x, -sympy.oo)
Python One: 1
 
julia> sympy.__version__
Python str: '1.10'
```

Any reason behind this error?

anutosh491:

Fixes #23231 
 
The block we are getting rid of which is  as follows .
```
--- a/sympy/core/power.py
+++ b/sympy/core/power.py
@@ -1792,11 +1792,7 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
             raise PoleError("Cannot expand %s around 0" % (self))
         elif e.has(x):
             lt = exp(e * log(b))
-            try:
-                lt = lt.as_leading_term(x, logx=logx, cdir=cdir)
-            except PoleError:
-                pass
-            return lt
+            return lt.as_leading_term(x, logx=logx, cdir=cdir)
```

This block was responsible for handling the following cases (cases where expressions have args involving variables in exponents/powers for eg `x**x` , `(2**x + x**(1/x))` ). Hence getting rid of this block will spoil almost all cases of limits and order of such type as something like  `(x**(1/x)).as_leading_term(x)` would give a `NotImplementedError/PoleError` .
```
# On master
>>> (x**(1/x)).as_leading_term(x)
exp(log(x)/x)
>>> (2**(1/x)).as_leading_term(x)
exp(log(2)/x)
>>> ((1/x)**(1/x)).as_leading_term(x)
exp(log(1/x)/x)
```

What would be potentially spoiled here .
```
# results after making the change
>>> O(x**x, (x, oo))      # on master O(exp(x*log(x)), (x, oo))
O((1/x)**x, (x, oo))
>>> O(x**x + 2**x, (x, oo))   # on master O(exp(x*log(x)), (x, oo))
O(1, (x, oo))
>>> O(x**x + 1/x**2, (x, oo))   # on master O(exp(x*log(x)), (x, oo))
O(x**(-2), (x, oo))
>>>
>>> O(x**x + x**2, (x, oo))   # on master O(exp(x*log(x)), (x, oo))
O(x**2, (x, oo))
```

Hence now this case needs to expressed in the leading term method of `exponential.py` . Order for such cases (variables present in exponents) was under tested(just one trivial case i guess) or rather not tested at all , hence I have addressed a test for this too .

#### Release Notes

* series
  * Fixed leading term method inconsistency for expressions involving variables in powers.

### DIFF
--- a/sympy/functions/elementary/exponential.py
+++ b/sympy/functions/elementary/exponential.py
@@ -541,8 +541,6 @@ class exp(ExpBase, metaclass=ExpMeta):
             # Check out function: test_issue_18473() in test_exponential.py and
             # test_limits.py for more information.
             return exp(arg0)
-        if isinstance(arg.as_numer_denom()[0], log):
-            return self
         if arg0 is S.NaN:
             arg0 = arg.limit(x, 0)
         if arg0.is_infinite is False:
--- a/sympy/functions/elementary/exponential.py
+++ b/sympy/functions/elementary/exponential.py
@@ -541,6 +541,8 @@ class exp(ExpBase, metaclass=ExpMeta):
             # Check out function: test_issue_18473() in test_exponential.py and
             # test_limits.py for more information.
             return exp(arg0)
+        if isinstance(arg.as_numer_denom()[0], log):
+            return self
         if arg0 is S.NaN:
             arg0 = arg.limit(x, 0)
         if arg0.is_infinite is False:
--- a/sympy/core/power.py
+++ b/sympy/core/power.py
@@ -1793,11 +1793,7 @@ class Pow(Expr):
             raise PoleError("Cannot expand %s around 0" % (self))
         elif e.has(x):
             lt = exp(e * log(b))
-            try:
-                lt = lt.as_leading_term(x, logx=logx, cdir=cdir)
-            except PoleError:
-                pass
-            return lt
+            return lt.as_leading_term(x, logx=logx, cdir=cdir)
         else:
             from sympy.functions.elementary.complexes import im
             f = b.as_leading_term(x, logx=logx, cdir=cdir)
--- a/sympy/functions/elementary/tests/test_exponential.py
+++ b/sympy/functions/elementary/tests/test_exponential.py
@@ -210,12 +210,6 @@ def test_exp_leading_term():
     # raises(NotImplementedError, lambda: exp(x + 1/x).as_leading_term(x))
 
 
-def test_issue_23231():
-    assert (x**(1/x)).as_leading_term(x) == exp(log(x)/x)
-    assert (2**(1/x)).as_leading_term(x) == exp(log(2)/x)
-    assert ((1/x)**(1/x)).as_leading_term(x) == exp(log(1/x)/x)
-
-
 @_both_exp_pow
 def test_exp_taylor_term():
     x = symbols('x')
--- a/sympy/functions/elementary/tests/test_exponential.py
+++ b/sympy/functions/elementary/tests/test_exponential.py
@@ -210,6 +210,12 @@ def test_exp_leading_term():
     # raises(NotImplementedError, lambda: exp(x + 1/x).as_leading_term(x))
 
 
+def test_issue_23231():
+    assert (x**(1/x)).as_leading_term(x) == exp(log(x)/x)
+    assert (2**(1/x)).as_leading_term(x) == exp(log(2)/x)
+    assert ((1/x)**(1/x)).as_leading_term(x) == exp(log(1/x)/x)
+
+
 @_both_exp_pow
 def test_exp_taylor_term():
     x = symbols('x')
--- a/sympy/series/order.py
+++ b/sympy/series/order.py
@@ -254,8 +254,6 @@ class Order(Expr):
                                 expr = new_expr.expr
                             elif expr.is_Mul:
                                 expr = Mul(*[a.expr for a in orders])
-                            elif expr.is_Pow:
-                                expr = orders[0].expr**orders[1].expr
                     expr = expr.as_independent(*args, as_Add=False)[1]
 
                     expr = expand_power_base(expr)
--- a/sympy/series/order.py
+++ b/sympy/series/order.py
@@ -3,6 +3,7 @@ from sympy.core.cache import cacheit
 from sympy.core.containers import Tuple
 from sympy.core.function import Function, PoleError, expand_power_base, expand_log
 from sympy.core.sorting import default_sort_key
+from sympy.functions.elementary.exponential import exp, log
 from sympy.sets.sets import Complement
 from sympy.utilities.iterables import uniq, is_sequence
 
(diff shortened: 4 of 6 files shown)


anutosh491:
cc @jksuom

sympy/functions/elementary/exponential.py
```diff
@@ -541,6 +541,8 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
             # Check out function: test_issue_18473() in test_exponential.py and
             # test_limits.py for more information.
             return exp(arg0)
+        if isinstance(arg.as_numer_denom()[0], log):
```
jksuom, line 544: It seems that this will only work in very special cases not including this:
```
>>> O(x**(2*x), (x, oo))
O(exp(x*log(1/x)), (x, oo))
```

anutosh491: Yupp I am fixing this , i realized this will become a bit specialized . Hence I turned this into draft . Working on this

jksuom:
It seems that `Order.__new__` will give wrong results when the expression is a power that has no leading term: https://github.com/sympy/sympy/blob/ebda8eb85f7c07c7c01d6bf7678aca03243a2ef3/sympy/series/order.py#L257-L258
This is clearly wrong: The order of `b**e` cannot be derived from the orders of `b` and `e`. For example, the order of `x**2` is not `x**1`. (This will be handled by a different code path as there will be no `PoleError` but the expression should still be valid also in this case.)

anutosh491:
Also I have a doubt here , does it make sense to have something like this on master . This was what got me a bit confused and therefore I ended up writing a very specific case above , which after commiting I realized would clearly not answer evrything

```
>>> (2**(1/x)).as_leading_term(x)
exp(log(2)/x)
>>> exp(log(2)/x).as_leading_term(x)
  File "C:\Users\anuto\sympy\sympy\sympy\core\cache.py", line 70, in wrapper
  File "C:\Users\anuto\sympy\sympy\sympy\core\expr.py", line 3447, in as_leading_term
    obj = self._eval_as_leading_term(x, logx=logx, cdir=cdir)
  File "C:\Users\anuto\sympy\sympy\sympy\functions\elementary\exponential.py", line 548, in _eval_as_leading_term
    raise PoleError("Cannot expand %s around 0" % (self))
sympy.core.function.PoleError: Cannot expand exp(log(2)/x) around 0
```
At the end both are same things right ?? Are these results completely correct according to you ?

anutosh491:
> This is clearly wrong: The order of `b**e` cannot be derived from the orders of `b` and `e`. For example, the order of `x**2` is not `x**1`. (This will be handled by a different code path as there will be no `PoleError` but the expression should still be valid also in this case.)

Completely agreed , thanks for pointing this out . Will have a look into this !

anutosh491:
> This is clearly wrong: The order of `b**e` cannot be derived from the orders of `b` and `e`. For example, the order of `x**2` is not `x**1`. (This will be handled by a different code path as there will be no `PoleError` but the expression should still be valid also in this case.)

Actually I realize , we don't even need to deal with this as of now . Neither would this be affecting any test now would this be giving us any mathematically wrong results . I am not sure why the author of this code used it like the way it is because aren't any tests dependent on that block either . 
If we just get rid of that block , we end up getting better results 
```
>>> O(2**x + x**x , (x, oo))
O((1/x)**(-x), (x, oo))
>>> O(x**(2*x) + x**x , (x, oo))
O((1/x)**(-2*x), (x, oo))
>>> O(x**(x**(2)) + x**(2*x) , (x, oo))
O((1/x)**(-x**2), (x, oo))
```

Should we let it be how it is , I'll make the commit cause this looks somewhat decent to me , you can guide me from there and obviously would work better than the current committed code !

anutosh491:
The reason behind naming this as follows it that now the inconsistencies in this commit have been addressed (https://github.com/sympy/sympy/pull/23241#issuecomment-1068040386)
We have 
```
>>> (2**(1/x)).as_leading_term(x)
..............................
sympy.core.function.PoleError: Cannot expand exp(log(2)/x) around 0
>>> exp(log(2)/x).as_leading_term(x)
.................................................
sympy.core.function.PoleError: Cannot expand exp(log(2)/x) around 0
```

anutosh491:
The fail here isn't too big a concern 
```
>>> x = Symbol('x')
>>> lowergamma(x, 2).series(x, oo, 3)
2**x*(1 + 2/(x + 1))*exp(-2)/x + O(exp(x*log(2))/x**3, (x, oo))
>>> # On committed branch
>>> lowergamma(x, 2).series(x, oo, 3)
2**x*(1 + 2/(x + 1))*exp(-2)/x + O(2**x/x**3, (x, oo))
```

jksuom:
There are several ways to handle tho order of a power raising `PoleError`:

1. Remove the `elif expr.is_Pow:` block.
    This means essentially that `expr = orders[0].expr**orders[1].expr` is replaced by `pass` leaving `expr` intact.
2. Set `expr = exp(e*log(b)` copying what is done in `Pow._eval_as_leading_term`.
3. `raise NotImplementedError`
    This is a plausible possibility as long as there is no clear understanding what the order of an expression with no power series expansion should mean.

I would experiment with 2, in addition to 1. If the results are not satisfactory, 3 could be considered.

anutosh491:
> There are several ways to handle the order of a power raising `PoleError`:
> I would experiment with 2, in addition to 1. If the results are not satisfactory, 3 could be considered.

Thanks for the suggestions
I feel we wouldn't need 3 , first two would be good enough . I'll try these suggestions out , I had thought about 2 but 1 wasn't bad either(which is what I tried in the latest commit) ! I'll let you know which one performs best in some time

anutosh491:
This looks much better to me than previous commits and I'm quite sure would do the job correctly for us . Looks good to go , I would admit  . What  are you thoughts here ?

anutosh491:
Also happy with how order would be dealt with for expressions of type pow , it goes as follows
```
>>> O(x**x + 2**x, (x, oo))
O(exp(x*log(x)), (x, oo))
>>> O(x**x + x**2, (x, oo))
O(exp(x*log(x)), (x, oo))
>>> O(x**x + 1/x**2, (x, oo))
O(exp(x*log(x)), (x, oo))
>>> O(2**x + 3**x , (x, oo))
O(exp(x*log(3)), (x, oo))
```

anutosh491:
cc @jksuom

jksuom:
Looks good, thanks.

PR #23242, opened 2022-03-15
Fixed exception message

islem-esi:
Changed exception messages in two files (one is wrong an the other is vague)

NO ENTRY

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -626,6 +626,7 @@ Ilya Pchelintsev <ilya.pchelintsev@outlook.com> ILLIA PCHALINTSAU <ilya.pchelint
 Imran Ahmed Manzoor <imran.manzoor31@gmail.com>
 Ishan Joshi <ishanaj98@gmail.com>
 Islam Mansour <is3mansour@gmail.com>
+Islem BOUZENIA <fi_bouzenia@esi.dz> islem-esi <fi_bouzenia@esi.dz>
 Isuru Fernando <isuruf@gmail.com> <isuru.11@cse.mrt.ac.lk>
 Itay4 <31018228+Itay4@users.noreply.github.com>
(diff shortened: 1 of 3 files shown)


dyc3 (approved):

oscargus:
Looks good!

See https://github.com/sympy/sympy/wiki/Development-workflow#add-your-name-and-email-address-to-the-mailmap-file
to get rid of the failing test (add your name and email to the mailmap file).

.mailmap
```diff
@@ -1325,3 +1325,4 @@ zzj <29055749+zjzh@users.noreply.github.com>
 Øyvind Jensen <jensen.oyvind@gmail.com>
 Łukasz Pankowski <lukpank@o2.pl>
 彭于斌 <1931127624@qq.com>
+Islem BOUZENIA <fi_bouzenia@esi.dz>
```
oscargus, line 1328:
```suggestion
Islem BOUZENIA <fi_bouzenia@esi.dz> Islem BOUZENIA <61057090+islem-esi@users.noreply.github.com>
```

oscargus: Seems like you had a previous entry if I interpreted the error message correctly. I do not really follow this, but you should have a line similar to the above, with both your email addresses.

oscarbenjamin: The message is confusing but the problem is that neither of the names here actually matches what is in the git commits (`islem-esi <fi_bouzenia@esi.dz>`). The line to be added should be:
```
Islem BOUZENIA <fi_bouzenia@esi.dz> islem-esi <fi_bouzenia@esi.dz>
```
There the name "islem-esi" is what is actually in the commits.

After adding that line the `bin/mailmap_check.py` script should be run to sort the file.

PR #23249, opened 2022-03-16
Removed type annotations from lambdify()

moorepants:
The type annotations add extra noise when trying to read a doc string
and offer no useful addition over the docstring itself. The annotations
on lambdify's first two args are removed.

No decision has been made on whether we are introducing these type hints or not, see #17945 

I noticed that the docstring had changed when writing code and thought that function signature was no longer easily readable. Here is a screenshot from my editor:
image

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/utilities/lambdify.py
+++ b/sympy/utilities/lambdify.py
@@ -840,7 +840,7 @@ or tuple for the function arguments.
                 )
 
     # Get the names of the args, for creating a docstring
-    iterable_args: Iterable = (args,) if isinstance(args, Expr) else args
+    iterable_args = (args,) if isinstance(args, Expr) else args
     names = []
 
     # Grab the callers frame, for getting the names by inspection (if needed)
--- a/sympy/utilities/lambdify.py
+++ b/sympy/utilities/lambdify.py
@@ -27,14 +27,14 @@ __doctest_requires__ = {('lambdify',): ['numpy', 'tensorflow']}
 
 # Default namespaces, letting us define translations that can't be defined
 # by simple variable maps, like I => 1j
-MATH_DEFAULT = {}  # type: tDict[str, Any]
-MPMATH_DEFAULT = {}  # type: tDict[str, Any]
-NUMPY_DEFAULT = {"I": 1j}  # type: tDict[str, Any]
-SCIPY_DEFAULT = {"I": 1j}  # type: tDict[str, Any]
-CUPY_DEFAULT = {"I": 1j}  # type: tDict[str, Any]
-TENSORFLOW_DEFAULT = {}  # type: tDict[str, Any]
-SYMPY_DEFAULT = {}  # type: tDict[str, Any]
-NUMEXPR_DEFAULT = {}  # type: tDict[str, Any]
+MATH_DEFAULT = {}
+MPMATH_DEFAULT = {}
+NUMPY_DEFAULT = {"I": 1j}
+SCIPY_DEFAULT = {"I": 1j}
+CUPY_DEFAULT = {"I": 1j}
+TENSORFLOW_DEFAULT = {}
+SYMPY_DEFAULT = {}
+NUMEXPR_DEFAULT = {}
 
 # These are the namespaces the lambda functions will use.
 # These are separate from the names above because they are modified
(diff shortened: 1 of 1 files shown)


sympy/utilities/lambdify.py
```diff
@@ -179,7 +179,7 @@ def _import(module, reload=False):
 
 
 @doctest_depends_on(modules=('numpy', 'scipy', 'tensorflow',), python_version=(3,))
-def lambdify(args: tUnion[Iterable, 'sympy.core.expr.Expr'], expr: 'sympy.core.expr.Expr', modules=None, printer=None, use_imps=True,
```
asmeurer, line 182: If we did have a type hint here it would be better to write it more concisely as `Iterable | Expr`. 

However, I think this is an example of something I touched on in https://github.com/sympy/sympy/issues/17945. The hint here is not actually correct. The `args` can be an arbitrarily nested iterable, like `((x, y), z)` (it also really should be Sequence, not Iterable, since unordered iterables like `set` are not allowed). To make nested sequences work, you need something like this https://github.com/numpy/numpy/pull/20129.

sympy/utilities/lambdify.py
```diff
@@ -3,7 +3,7 @@
 lambda functions which can be used to calculate numerical values very fast.
 """
 
-from typing import Any, Dict as tDict, Iterable, Union as tUnion, TYPE_CHECKING
```
asmeurer, line 6: You removed Dict and Any from the import but they don't seem to be removed in the file. Were they unused or is this removing too much?

moorepants: My linter said they were unused. I'll have to check, maybe they weren't.

moorepants: There are a ton of comments like this:

```python
# type: tDict[str, str]
```

I'm not sure if they actually have functionality.

moorepants: I removed all those comments.

sympy/utilities/lambdify.py
moorepants: Yes, I noticed it wasn't correct either. It is just confusing, is what it is.

moorepants:
The mypy ci run reports:

```
Run mypy sympy
sympy/utilities/lambdify.py:24: error: Need type annotation for "MATH_DEFAULT" (hint: "MATH_DEFAULT: Dict[<type>, <type>] = ...")
sympy/utilities/lambdify.py:25: error: Need type annotation for "MPMATH_DEFAULT" (hint: "MPMATH_DEFAULT: Dict[<type>, <type>] = ...")
sympy/utilities/lambdify.py:29: error: Need type annotation for "TENSORFLOW_DEFAULT" (hint: "TENSORFLOW_DEFAULT: Dict[<type>, <type>] = ...")
sympy/utilities/lambdify.py:30: error: Need type annotation for "SYMPY_DEFAULT" (hint: "SYMPY_DEFAULT: Dict[<type>, <type>] = ...")
sympy/utilities/lambdify.py:31: error: Need type annotation for "NUMEXPR_DEFAULT" (hint: "NUMEXPR_DEFAULT: Dict[<type>, <type>] = ...")
sympy/utilities/lambdify.py:89: error: Need type annotation for "SCIPY_TRANSLATIONS" (hint: "SCIPY_TRANSLATIONS: Dict[<type>, <type>] = ...")
sympy/utilities/lambdify.py:90: error: Need type annotation for "CUPY_TRANSLATIONS" (hint: "CUPY_TRANSLATIONS: Dict[<type>, <type>] = ...")
sympy/utilities/lambdify.py:92: error: Need type annotation for "TENSORFLOW_TRANSLATIONS" (hint: "TENSORFLOW_TRANSLATIONS: Dict[<type>, <type>] = ...")
sympy/utilities/lambdify.py:94: error: Need type annotation for "NUMEXPR_TRANSLATIONS" (hint: "NUMEXPR_TRANSLATIONS: Dict[<type>, <type>] = ...")
Found 9 errors in 1 file (checked 1453 source files)
```

moorepants:
Closing in favor of https://github.com/sympy/sympy/pull/23251

asmeurer:
I guess let's leave those in. It makes sense that if you have `a = {}` then mypy needs an explicit annotation to know the types of the keys and values. I think we can instead write it as

```py
MATH_DEFAULT: dict[str, Any] = {}
```

which is more readable. Personally I don't think variable annotations are as bad as function signature annotations because they only affect the code itself, not the documentation, and they only need to be used in cases like this where it's otherwise impossible to infer the type automatically.

moorepants:
> Personally I don't think variable annotations are as bad as function signature annotations because they only affect the code itself, not the documentation, and they only need to be used in cases like this where it's otherwise impossible to infer the type automatically.

I agree they aren't as bad, but it still is unnecessary clutter.

PR #23255, opened 2022-03-17
solvers - fixes issue#23253 and improved LambertW solution detection

### LINKED ISSUE #23253: improve LambertW solution detection
The [SO question](https://stackoverflow.com/a/71513677/1089161) has an answer and work-around. It looks like `a**2/log(b/x + c)**2-x**2` needs a change of variables to `a**2/log(y)**2-(b/(y-c))**2` in order to solve (and then conversion of answers for `y` to `x`).

praneethratna:
Fixes #23253 by applying the changes mentioned in https://github.com/sympy/sympy/issues/23253#issuecomment-1071370065 and improved solution detection of LambertW.

#### Release Notes

* solvers
    * improved detection of Lambert solutions when a power of x appear in `log`

### DIFF
--- a/sympy/solvers/solvers.py
+++ b/sympy/solvers/solvers.py
@@ -2719,9 +2719,9 @@ def _tsolve(eq, sym, **flags):
         # 2**(3*x) + x**3*log(2)**3 + 3*x**2*log(2)**2 + 3*x*log(2) + 1
         # to 2**(3*x) + (x*log(2) + 1)**3
         logs = eq.atoms(log)
-        spow = {i.exp.as_coeff_Mul()[0] for j in logs for i in j.atoms(Pow) if i.base == sym}
+        spow = {i.exp for j in logs for i in j.atoms(Pow) if i.base == sym}
         spow = min(spow) if spow else 1
-        if spow > 1 and eq.subs(sym**spow, Dummy()).has_free(sym):
+        if eq.subs(sym**spow, Dummy()).has_free(sym):
             spow = 1  # there was a free-symbol or non-pow generator with sym
         if spow != 1:
             u = Dummy('bivariate-cov')
--- a/sympy/solvers/solvers.py
+++ b/sympy/solvers/solvers.py
@@ -2718,21 +2718,23 @@ def _tsolve(eq, sym, **flags):
         # lambert forms may need some help being recognized, e.g. changing
         # 2**(3*x) + x**3*log(2)**3 + 3*x**2*log(2)**2 + 3*x*log(2) + 1
         # to 2**(3*x) + (x*log(2) + 1)**3
+
+        # make generator in log have exponent of 1
         logs = eq.atoms(log)
-        spow = {i.exp for j in logs for i in j.atoms(Pow) if i.base == sym}
-        spow = min(spow) if spow else 1
-        if eq.subs(sym**spow, Dummy()).has_free(sym):
-            spow = 1  # there was a free-symbol or non-pow generator with sym
+        spow = min(
+            {i.exp for j in logs for i in j.atoms(Pow)
+             if i.base == sym} or {1})
         if spow != 1:
-            u = Dummy('bivariate-cov')
             p = sym**spow
+            u = Dummy('bivariate-cov')
             ueq = eq.subs(p, u)
-            sol = solve(ueq, u, **flags)
-            inv = solve(p - u, sym)
-            rv = []
-            for i in inv:
-                rv.extend([i.subs(u, s) for s in sol])
-            return rv
+            if not ueq.has_free(sym):
+                sol = solve(ueq, u, **flags)
+                inv = solve(p - u, sym)
+                rv = []
+                for i in inv:
+                    rv.extend([i.subs(u, s) for s in sol])
+                return rv
 
         g = _filtered_gens(eq.as_poly(), sym)
         up_or_log = set()
--- a/sympy/solvers/solvers.py
+++ b/sympy/solvers/solvers.py
@@ -2719,7 +2719,7 @@ def _tsolve(eq, sym, **flags):
         # 2**(3*x) + x**3*log(2)**3 + 3*x**2*log(2)**2 + 3*x*log(2) + 1
         # to 2**(3*x) + (x*log(2) + 1)**3
         logs = eq.atoms(log)
-        spow = {i.exp for j in logs for i in j.atoms(Pow) if i.base == sym}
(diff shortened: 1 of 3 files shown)


sympy/solvers/solvers.py
```diff
@@ -2718,6 +2718,22 @@ def equal(expr1, expr2):
         # lambert forms may need some help being recognized, e.g. changing
         # 2**(3*x) + x**3*log(2)**3 + 3*x**2*log(2)**2 + 3*x*log(2) + 1
         # to 2**(3*x) + (x*log(2) + 1)**3
+        logs = eq.atoms(log)
+        spow = {i.exp for j in logs for i in j.atoms(Pow) if i.base == sym}
```
smichr, line 2722:
```suggestion
        spow = {i.exp.as_coeff_Mul()[0] for j in logs for i in j.atoms(Pow) if i.base == sym}
```

smichr:
I'm not sure if it is best to put this here or to locate it within the code that identifies the "log(x) + x" code -- I suspect the latter. I spent a lot of time originally trying to work out the varieties of ways the forms could appear. I missed this one (at least), however. I wonder if it applies to the other cases in some way, too. These would be worth looking into if you have the time and inclination.

sympy/solvers/tests/test_solvers.py
```diff
@@ -1896,6 +1896,17 @@ def test_lambert_bivariate():
     assert solve((x**3)**(x/2) + pi/2, x) == [
         exp(LambertW(-2*log(2)/3 + 2*log(pi)/3 + I*pi*Rational(2, 3)))]
 
+    # issue 23253
+    assert solve((1/log(sqrt(x) + 2)**2 - 1/x)) == [
```
smichr, line 1900: Does `solve((1/log(1/sqrt(x) + 2)**2 - x))` also work and give `[(LambertW(-exp(-2), -1) + 2)**-2`]?

sympy/solvers/tests/test_solvers.py
```diff
@@ -1896,6 +1896,17 @@ def test_lambert_bivariate():
...
+        (LambertW(-exp(-2), -1) + 2)**2]
+    assert solve((1/log(x**2 + 2)**2 - x**-4)) == [
```
smichr, line 1902:
```suggestion
    assert solve((1/log(1/sqrt(x) + 2)**2 - x)) == [
        (LambertW(-exp(-2), -1) + 2)**-2]
    assert solve((1/log(x**2 + 2)**2 - x**-4)) == [
```

praneethratna:
> These would be worth looking into if you have the time and inclination.

Sorry for the delay, I have tried a diff which seems to be working fine for all cases, not sure if that is the right way.

sympy/solvers/solvers.py
```diff
@@ -2719,9 +2719,9 @@ def equal(expr1, expr2):
         # 2**(3*x) + x**3*log(2)**3 + 3*x**2*log(2)**2 + 3*x*log(2) + 1
         # to 2**(3*x) + (x*log(2) + 1)**3
         logs = eq.atoms(log)
-        spow = {i.exp.as_coeff_Mul()[0] for j in logs for i in j.atoms(Pow) if i.base == sym}
+        spow = {i.exp for j in logs for i in j.atoms(Pow) if i.base == sym}
         spow = min(spow) if spow else 1
```
smichr, line 2723: this will fail with symbolic exponents...what was the failing expression that got here from dsolve?

praneethratna: This is the failing expression - `(x + sqrt(-x*f(x) + f(x)**2))*diff(f(x), x) - f(x)`

praneethratna: I think this diff should work for symbolic exponents as well, correct me if I'm wrong.
```diff
diff --git a/sympy/solvers/solvers.py b/sympy/solvers/solvers.py
index 07ac28db2d..d5c677ad6f 100644
--- a/sympy/solvers/solvers.py
+++ b/sympy/solvers/solvers.py
@@ -2719,7 +2719,7 @@ def equal(expr1, expr2):
         # 2**(3*x) + x**3*log(2)**3 + 3*x**2*log(2)**2 + 3*x*log(2) + 1
         # to 2**(3*x) + (x*log(2) + 1)**3
         logs = eq.atoms(log)
-        spow = {i.exp for j in logs for i in j.atoms(Pow) if i.base == sym}
+        spow = {i.exp.as_coeff_Mul()[1] for j in logs for i in j.atoms(Pow) if i.base == sym}
         spow = min(spow) if spow else 1
         if eq.subs(sym**spow, Dummy()).has_free(sym):
             spow = 1  # there was a free-symbol or non-pow generator with sym
```

praneethratna:
ping @smichr

sympy/solvers/solvers.py
smichr:
```python
>>> sym=x
>>> logs = log(x**2), log(x**y)
>>> spow = {i.exp.as_coeff_Mul()[1] for j in logs for i in j.atoms(Pow) if i.base == sym}
>>> spow
{1,y}
>>> min(spow)
...
TypeError: cannot determine truth value of Relational
```

smichr: > This is the failing expression

Since there are no log terms, I don't understand why this is failing....

sympy/solvers/solvers.py
```diff
@@ -2718,6 +2718,22 @@ def equal(expr1, expr2):
...
         # to 2**(3*x) + (x*log(2) + 1)**3
+        logs = eq.atoms(log)
+        spow = {i.exp for j in logs for i in j.atoms(Pow) if i.base == sym}
+        spow = min(spow) if spow else 1
+        if eq.subs(sym**spow, Dummy()).has_free(sym):
+            spow = 1  # there was a free-symbol or non-pow generator with sym
+        if spow != 1:
+            u = Dummy('bivariate-cov')
+            p = sym**spow
+            ueq = eq.subs(p, u)
```
smichr, line 2729:
```suggestion
        p = sym**spow
        u = Dummy('bivariate-cov')
        ueq = eq.subs(p, u)
        if not ueq.has(sym):
```

smichr: But extracting the `min`, even though `spow > 1` is not used, will fail if any power has a symbolic exponent. I can give a better test later if you don't come up with one.

smichr: How about `solve((1/log(x**(2*y) + 2)**2 - x**-(4*y)))`?

smichr: What is the value of the `spow` **set** when solving that equation?

PR #23256, opened 2022-03-18
Minor type hints cleaning

eagleoflqj:
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -1,8 +1,9 @@
 """Base class for all the objects in SymPy"""
+from __future__ import annotations
+
 from collections import defaultdict
 from collections.abc import Mapping
 from itertools import chain, zip_longest
-from typing import Set, Tuple, Any
 
 from .assumptions import ManagedProperties
 from .cache import cacheit
@@ -81,8 +82,8 @@ class Basic(Printable, metaclass=ManagedProperties):
                  '_assumptions'
                 )
 
-    _args: 'Tuple[Basic, ...]'
-    _mhash: 'Any'
+    _args: tuple[Basic, ...]
+    _mhash: int | None
 
     # To be overridden with True in the appropriate subclasses
     is_number = False
@@ -521,8 +522,10 @@ class Basic(Printable, metaclass=ManagedProperties):
         return result
 
     @property
-    def free_symbols(self) -> 'Set[Basic]':
+    def free_symbols(self) -> set[Basic]:
         """Return from the atoms of self those which are free symbols.
+
+        Not all free symbols are ``Symbol``. Eg: IndexedBase('I')[0].free_symbols
 
         For most expressions, all symbols are free symbols. For some classes
         this is not true. e.g. Integrals use Symbols for the dummy variables
@@ -533,7 +536,7 @@ class Basic(Printable, metaclass=ManagedProperties):
 
         Any other method that uses bound variables should implement a
         free_symbols method."""
-        empty: 'Set[Basic]' = set()
+        empty: set[Basic] = set()
         return empty.union(*(a.free_symbols for a in self.args))
 
     @property
@@ -741,7 +744,7 @@ class Basic(Printable, metaclass=ManagedProperties):
         return self.__class__
 
     @property
-    def args(self) -> 'Tuple[Basic, ...]':
+    def args(self) -> tuple[Basic, ...]:
         """Returns a tuple of arguments of 'self'.
 
         Examples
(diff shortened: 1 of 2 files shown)


sympy/core/basic.py
```diff
@@ -517,7 +522,7 @@ def atoms(self, *types):
         return result
 
     @property
-    def free_symbols(self) -> 'Set[Basic]':
+    def free_symbols(self) -> 'set[Symbol]':
```
oscarbenjamin, line 525: II don't think this is accurate given that
```python
In [17]: IndexedBase('I')[0].free_symbols
Out[17]: {I, I[0]}
```

eagleoflqj: Thanks for pointing that out!

eagleoflqj: Added to docstring as I may not be the only one that has this misunderstanding.

eagleoflqj:
Could anyone give a counterexample that free symbols are not even an `Expr`?

eagleoflqj:
I did a quick search by reading the implementations of these and didn't find a counterexample:
```py
$ git grep -n 'def free_symbols(self)'
sympy/codegen/ast.py:658:    def free_symbols(self):
sympy/concrete/expr_with_limits.py:306:    def free_symbols(self):
sympy/core/basic.py:522:    def free_symbols(self) -> set[Basic]:
sympy/core/function.py:207:    def free_symbols(self):
sympy/core/function.py:1680:    def free_symbols(self):
sympy/core/function.py:2036:    def free_symbols(self):
sympy/core/function.py:2322:    def free_symbols(self):
sympy/core/symbol.py:374:    def free_symbols(self):
sympy/diffgeom/diffgeom.py:892:    def free_symbols(self):
sympy/geometry/curve.py:157:    def free_symbols(self):
sympy/integrals/integrals.py:103:    def free_symbols(self):
sympy/integrals/transforms.py:109:    def free_symbols(self):
sympy/matrices/common.py:1412:    def free_symbols(self):
sympy/matrices/expressions/matexpr.py:699:    def free_symbols(self):
sympy/physics/quantum/operator.py:608:    def free_symbols(self):
sympy/physics/quantum/qexpr.py:101:    def free_symbols(self):
sympy/physics/quantum/state.py:825:    def free_symbols(self):
sympy/physics/units/quantities.py:232:    def free_symbols(self):
sympy/polys/polytools.py:328:    def free_symbols(self):
sympy/polys/polytools.py:4282:    def free_symbols(self):
sympy/polys/rootoftools.py:390:    def free_symbols(self):
sympy/polys/rootoftools.py:1189:    def free_symbols(self):
sympy/series/limits.py:173:    def free_symbols(self):
sympy/series/order.py:333:    def free_symbols(self):
sympy/series/sequences.py:84:    def free_symbols(self):
sympy/series/series_class.py:40:    def free_symbols(self):
sympy/sets/conditionset.py:182:    def free_symbols(self):
sympy/simplify/cse_main.py:359:    def free_symbols(self):
sympy/stats/rv.py:321:    def free_symbols(self):
sympy/stats/rv.py:345:    def free_symbols(self):
sympy/tensor/array/array_comprehension.py:83:    def free_symbols(self):
sympy/tensor/array/dense_ndim_array.py:208:    def free_symbols(self):
sympy/tensor/array/sparse_ndim_array.py:195:    def free_symbols(self):
sympy/tensor/indexed.py:343:    def free_symbols(self):
sympy/tensor/indexed.py:757:    def free_symbols(self):
sympy/vector/scalar.py:46:    def free_symbols(self):
sympy/vector/vector.py:402:    def free_symbols(self):
```
`SetSymbol` seems a union type of `Symbol`, `Indexed`, `BaseVector` and others, which (in my practice) adds complexity when being a return type. `Expr` is probably OK in common cases, like taking an element `x` from `set[Expr]` then performing `x+1`.

oscarbenjamin:
> didn't find a counterexample

There might not be a counterexample now but I think it should be possible to make a symbol corresponding to any kind of expression.

eagleoflqj:
> I'm not sure of an example but there is no reason in principle why there couldn't be something like a SetSymbol.

Now I see what you mean. Then it may be better to keep `free_symbols` `set[Basic]` now.

sympy/core/basic.py
```diff
@@ -81,8 +83,8 @@ class Basic(Printable, metaclass=ManagedProperties):
                  '_assumptions'
                 )
 
-    _args: 'Tuple[Basic, ...]'
-    _mhash: 'Any'
+    _args: tuple[Basic, ...]
+    _mhash: Any
```
oscarbenjamin, line 87: Is it possible for `_mhash` to be more specific? It's supposed to be an `int` but it doesn't get initalised until it is needed and until then it is None. Should this be `Optional[int]` rather than `Any`?

PR #23258, opened 2022-03-18
Series/limits : Fixes limit computations where arg0 would evaluate to S.NaN for the ceiling and floor functions

### LINKED ISSUE #14355: Bidirectional limit of floor(sin(x)/x) at 0, is not correct
```python
>>> pprint(Limit(floor(sin(x)/x),x,0,'+-'))

    ⎢sin(x)⎥
lim ⎢──────⎥
x─→0⎣  x   ⎦

>>> limit(floor(sin(x)/x),x,0,'+')
1

>>> limit(floor(sin(x)/x),x,0,'-')
1
```
The correct limit should evaluate to 0.
Also confirmed from [WA](http://www.wolframalpha.com/input/?i=limit(floor(sin(x)%2Fx),x,0)).

anutosh491:

Fixes #14355 

Also addresses https://github.com/sympy/sympy/issues/14355#issuecomment-372121314

This Pr fixes cases where `arg0` would evaluate to `S.NaN` and this case currently lacks support. Few results which have been addressed have been noted below !! 

```
>>> limit(floor((x**3 + x)/(x**2 - x)), x, 0, '+')  # -1 on master
-2
>>> limit(floor((x**3 + x)/(x**2 - x)), x, 0, '-')   # -1 on master but not through intended code
-1
>>> limit(floor((cos(x) - 1)/x), x, 0, '-')   # 0 on master but not through intended code
0
>>> limit(floor((cos(x) - 1)/x), x, 0, '+')  # 0 on master
-1
>>> limit(floor(sin(x)/x), x, 0, '+')   # 1 on master
0
>>> limit(floor(sin(x)/x), x, 0, '-')   # 1 on master
0
>>> limit(floor(-tan(x)/x), x, 0, '-')   # -1 on master
-2
>>> limit(floor(-tan(x)/x), x, 0, '+')   # -1 on master
-2
>>> # ceiling cases
>>> limit(ceiling((x**3 + x)/(x**2 - x)), x, 0, '-')  # -1 on master
0
>>> limit(ceiling((x**3 + x)/(x**2 - x)), x, 0, '+')   # -1 on master but not through intended code
-1
>>> limit(ceiling((cos(x) - 1)/x), x, 0, '+')  # 0 on master but not through intended code
0
>>> limit(ceiling((cos(x) - 1)/x), x, 0, '-')  # 0 on master
1
```

The issue deserved calculation of the **leading term** and then dealing with it similarly as all other cases for **leading terms** are dealt , hence this has good resonance with already implemented code !

#### Release Notes

* series
  * Fixes limit computations where arg0 would evaluate to S.NaN for the ceiling and floor functions.

### DIFF
--- a/sympy/functions/elementary/integers.py
+++ b/sympy/functions/elementary/integers.py
@@ -11,7 +11,7 @@ from sympy.core.numbers import Integer
 from sympy.core.relational import Gt, Lt, Ge, Le, Relational, is_eq
 from sympy.core.symbol import Symbol
 from sympy.core.sympify import _sympify
-from sympy.functions.elementary.complexes import im
+from sympy.functions.elementary.complexes import im, re
 from sympy.multipledispatch import dispatch
 
 ###############################################################################
@@ -144,10 +144,12 @@ class floor(RoundFunction):
         arg = self.args[0]
         arg0 = arg.subs(x, 0)
         r = self.subs(x, 0)
-        if arg0.is_finite or arg0 is S.NaN:
+        if arg0 is S.NaN:
+            limit = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
+            ndir = arg.dir(x, cdir=cdir)
+            return limit - 1 if ndir.is_negative else limit
+        if arg0.is_finite:
             if arg0 == r:
-                leading_term = arg.as_leading_term(x, logx=logx, cdir=cdir)
-                leading_term = leading_term.subs(x, 0)
                 if cdir == 0:
                     ndirl = arg.dir(x, cdir=-1)
                     ndir = arg.dir(x, cdir=1)
@@ -156,12 +158,8 @@ class floor(RoundFunction):
                                     "does not exist" % self)
                 else:
                     ndir = arg.dir(x, cdir=cdir)
-                if arg0 is S.NaN:
-                    return leading_term - 1 if ndir.is_negative else leading_term
                 return r - 1 if ndir.is_negative else r
             else:
-                if arg0 is S.NaN:
-                    return leading_term
                 return r
         return arg.as_leading_term(x, logx=logx, cdir=cdir)
 
@@ -308,10 +306,12 @@ class ceiling(RoundFunction):
         arg = self.args[0]
         arg0 = arg.subs(x, 0)
         r = self.subs(x, 0)
-        if arg0.is_finite or arg0 is S.NaN:
+        if arg0 is S.NaN:
+            limit = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
+            ndir = arg.dir(x, cdir=cdir)
+            return limit if ndir.is_negative else limit + 1
+        if arg0.is_finite:
             if arg0 == r:
-                leading_term = arg.as_leading_term(x, logx=logx, cdir=cdir)
-                leading_term = leading_term.subs(x, 0)
                 if cdir == 0:
                     ndirl = arg.dir(x, cdir=-1)
                     ndir = arg.dir(x, cdir=1)
@@ -320,12 +320,8 @@ class ceiling(RoundFunction):
                                     "does not exist" % self)
                 else:
                     ndir = arg.dir(x, cdir=cdir)
-                    if arg0 is S.NaN:
-                        return leading_term if ndir.is_negative else leading_term + 1
                 return r if ndir.is_negative else r + 1
             else:
-                if arg0 is S.NaN:
-                    return leading_term
                 return r
         return arg.as_leading_term(x, logx=logx, cdir=cdir)
 
--- a/sympy/functions/elementary/integers.py
+++ b/sympy/functions/elementary/integers.py
@@ -145,11 +145,8 @@ class floor(RoundFunction):
         arg0 = arg.subs(x, 0)
         r = self.subs(x, 0)
         if arg0 is S.NaN:
-            limit = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
-            ndir = arg.dir(x, cdir=cdir)
-            if not limit.is_integer:
-                return floor(limit)
-            return limit - 1 if ndir.is_negative else limit
+            arg0 = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
+            r = floor(arg0)
         if arg0.is_finite:
             if arg0 == r:
                 if cdir == 0:
@@ -168,19 +165,16 @@ class floor(RoundFunction):
     def _eval_nseries(self, x, n, logx, cdir=0):
         arg = self.args[0]
         arg0 = arg.subs(x, 0)
+        r = self.subs(x, 0)
         if arg0 is S.NaN:
-            ndir = arg.dir(x, cdir=cdir if cdir != 0 else 1)
-            series = arg._eval_nseries(x, n, logx, cdir).subs(x, 0)
-            if not series.is_integer:
-                return floor(series)
-            return series - 1 if ndir.is_negative else series
+            arg0 = arg._eval_nseries(x, n, logx, cdir).subs(x, 0)
+            r = floor(arg0)
         if arg0.is_infinite:
             from sympy.calculus.accumulationbounds import AccumBounds
             from sympy.series.order import Order
             s = arg._eval_nseries(x, n, logx, cdir)
             o = Order(1, (x, 0)) if n <= 0 else AccumBounds(-1, 0)
             return s + o
-        r = self.subs(x, 0)
         if arg0 == r:
             ndir = arg.dir(x, cdir=cdir if cdir != 0 else 1)
             return r - 1 if ndir.is_negative else r
@@ -315,11 +309,8 @@ class ceiling(RoundFunction):
         arg0 = arg.subs(x, 0)
         r = self.subs(x, 0)
         if arg0 is S.NaN:
-            limit = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
-            ndir = arg.dir(x, cdir=cdir)
-            if not limit.is_integer:
-                return ceiling(limit)
-            return limit if ndir.is_negative else limit + 1
+            arg0 = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
+            r = ceiling(arg0)
         if arg0.is_finite:
             if arg0 == r:
                 if cdir == 0:
@@ -338,19 +329,16 @@ class ceiling(RoundFunction):
     def _eval_nseries(self, x, n, logx, cdir=0):
         arg = self.args[0]
         arg0 = arg.subs(x, 0)
+        r = self.subs(x, 0)
         if arg0 is S.NaN:
-            ndir = arg.dir(x, cdir=cdir if cdir != 0 else 1)
-            series = arg._eval_nseries(x, n, logx, cdir).subs(x, 0)
-            if not series.is_integer:
-                return ceiling(series)
-            return series if ndir.is_negative else series + 1
+            arg0 = arg._eval_nseries(x, n, logx, cdir).subs(x, 0)
+            r = ceiling(arg0)
         if arg0.is_infinite:
             from sympy.calculus.accumulationbounds import AccumBounds
             from sympy.series.order import Order
             s = arg._eval_nseries(x, n, logx, cdir)
             o = Order(1, (x, 0)) if n <= 0 else AccumBounds(0, 1)
             return s + o
-        r = self.subs(x, 0)
         if arg0 == r:
             ndir = arg.dir(x, cdir=cdir if cdir != 0 else 1)
             return r if ndir.is_negative else r + 1
--- a/sympy/functions/elementary/integers.py
+++ b/sympy/functions/elementary/integers.py
@@ -144,23 +144,23 @@ class floor(RoundFunction):
         arg = self.args[0]
         arg0 = arg.subs(x, 0)
         r = self.subs(x, 0)
-        if arg0 is S.NaN:
-            limit = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
-            ndir = arg.dir(x, cdir=cdir)
-            return limit - 1 if ndir.is_negative else limit
-        if arg0.is_finite:
-            if arg0 == r:
-                if cdir == 0:
-                    ndirl = arg.dir(x, cdir=-1)
-                    ndir = arg.dir(x, cdir=1)
-                    if ndir != ndirl:
-                        raise ValueError("Two sided limit of %s around 0"
-                                    "does not exist" % self)
-                else:
-                    ndir = arg.dir(x, cdir=cdir)
-                return r - 1 if ndir.is_negative else r
+        if arg0.is_finite or arg0 is S.NaN:
+            if cdir == 0:
+                ndirl = arg.dir(x, cdir=-1)
+                ndir = arg.dir(x, cdir=1)
+                if ndir != ndirl:
+                    raise ValueError("Two sided limit of %s around 0"
+                                     "does not exist" % self)
             else:
-                return r
+                ndir = arg.dir(x, cdir=cdir)
+            if arg0.is_finite:
+                if arg0 == r:
+                    return r - 1 if ndir.is_negative else r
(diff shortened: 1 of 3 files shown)


anutosh491:
cc @jksuom

sympy/functions/elementary/integers.py
```diff
@@ -144,8 +144,10 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
         arg = self.args[0]
         arg0 = arg.subs(x, 0)
         r = self.subs(x, 0)
-        if arg0.is_finite:
+        if arg0.is_finite or arg0 is S.NaN:
```
jksuom, line 147: These two cases should probably be treated in separation. If `arg0` is nan (i.e. subs fails), then limit of arg should be attempted.

anutosh491: Cool , completely agreed . I had thought of going for that , but this seemed to be a viable option . I'll switch to this in the next commit !

anutosh491: I've made a commit fixing this , this looks better than the previous commit to me , what do you think here ?

anutosh491: Actually I realize , we won't be able to handle bi-directional limits through the latest commit
```
        if arg0 is S.NaN:
            limit = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
            ndir = arg.dir(x, cdir=cdir)
            return limit if ndir.is_negative else limit + 1
```
Consider calculating `limit(ceiling(sin(x)/x), x, 0, '+-')` , now through this we clearly get limits 1 from both sides but we won't be able to work this for the bi-directional case
```
>>> x = Symbol('x')
>>> limit(ceiling(sin(x)/x), x, 0)
1
>>> limit(ceiling(sin(x)/x), x, 0, '-')
1
>>> limit(ceiling(sin(x)/x), x, 0, '+-')
2
```
This is because through 
```
# line 1 limit = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+') we get limit as 1
# line 2 ndir = arg.dir(x, cdir=cdir) we get ndir as 0
# line 3 return limit if ndir.is_negative else limit + 1 we get answer as 2
```
To fix this we need a similar logic to that as the one written for cases where `arg0 is finite`
```
        if arg0.is_finite:
            if arg0 == r:
                if cdir == 0:
                    ndirl = arg.dir(x, cdir=-1)
                    ndir = arg.dir(x, cdir=1)
                    if ndir != ndirl:
                        raise ValueError("Two sided limit of %s around 0"
                                    "does not exist" % self)
                else:
                    ndir = arg.dir(x, cdir=cdir)
```
And clearly there is no point writing this twice !!! which could be done here . Hence I'll make the change for this , you could let me know if that could be improved .
Otherwise we won't be correctly solving the issue which states (**Bidirectional limit of floor(sin(x)/x) at 0, is not correct**) as of now . I'll also add relevant tests for this !

anutosh491: I am quite happy with the latest commit here , answers all cases correctly . I guess this might be good to go once you review this !

jksuom: arg0 should not be NaN in this block. limit should be taken before this line.

anutosh491: > arg0 should not be NaN in this block. limit should be taken before this line.
Ohh so you mean the second commit was correct ?
```
-        if arg0.is_finite or arg0 is S.NaN:
+       if arg0 is S.NaN:
+           limit = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
+           ndir = arg.dir(x, cdir=cdir)
+            return limit - 1 if ndir.is_negative else limit
        if arg0.is_finite:
            if arg0 == r:
                if cdir == 0:
                    ndirl = arg.dir(x, cdir=-1)
                    ndir = arg.dir(x, cdir=1)
```

anutosh491: okay now that you've pointed that there should be a separate block, I would like to know your thoughts on this . I feel this would be the best way to go here . Because if we don't go for the same block there will be duplication of the following check shown below , which wouldn't be the correct way to go about here I presume.
```
        if cdir == 0:
            ndirl = arg.dir(x, cdir=-1)
            ndir = arg.dir(x, cdir=1)
            if ndir != ndirl:
                raise ValueError("Two sided limit of %s around 0"
                                 "does not exist" % self)
```
Hence we could go with this and use the check for the bi-directional cases of both (when **arg0 is finite or is S.NaN**)
```
        arg = self.args[0]
        arg0 = arg.subs(x, 0)
        r = self.subs(x, 0)
        if cdir == 0:
            ndirl = arg.dir(x, cdir=-1)
            ndir = arg.dir(x, cdir=1)
            if ndir != ndirl:
                raise ValueError("Two sided limit of %s around 0"
                                 "does not exist" % self)
        else:
            ndir = arg.dir(x, cdir=cdir)
        if arg0 is S.NaN:
            limit = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
            return limit if ndir.is_negative else limit + 1
        if arg0.is_finite:
            if arg0 == r:
                return r if ndir.is_negative else r + 1
            else:
                return r
        return arg.as_leading_term(x, logx=logx, cdir=cdir)
```

anutosh491: @jksuom , I've reverted the last commit , as the second commit would be according to what  you would prefer and maybe like to have !!

jksuom:
It would be better to have direct tests for leading terms like `(floor(sin(x)/x)).as_leading_term(x)`. If those are correct, then the limits would also succeed and not so many of those tests would be needed.

anutosh491:
> It would be better to have direct tests for leading terms like `(floor(sin(x)/x)).as_leading_term(x)`. If those are correct, then the limits would also succeed and not so many of those tests would be needed.

Sure will make the changes soon , though I'll still keep the sin and tan case tested , just to have some reference regarding the issue

anutosh491:
> It would be better to have direct tests for leading terms like `(floor(sin(x)/x)).as_leading_term(x)`. If those are correct, then the limits would also succeed and not so many of those tests would be needed.

I've made the required changes , should be good to go I guess !

anutosh491:
I would also like to address series for such functions , sometime soon(once my current midterm evaluations at university are over and I get more time) through a different pr as not really related to this issue . For eg
```
>>> floor(sin(x)/x).series(x)
nan
>>> # hence does it make sense to return series expansion of expr inside floor ??
>>> floor(sin(x)/x).series(x)
floor(-x**2/6 + x**4/120 + O(x**6)) + 1
```
The nan handling is missing here too I assume and I'm not sure if this is what should be returned ? is the second computation showed above even correct ? @jksuom

jksuom:
> is the second computation showed above even correct ?

I think that the series expansions of floor and ceiling should contain only one constant term, no powers of x.

anutosh491:
> > is the second computation showed above even correct ?
> 
> I think that the series expansions of floor and ceiling should contain only one constant term, no powers of x.

sure, will keep this in mind . Also what are you're thoughts on the pr otherwise , are you happy with the changes here ?

anutosh491:
> > is the second computation showed above even correct ?
> 
> I think that the series expansions of floor and ceiling should contain only one constant term, no powers of x.

I went on implementing the series support case too , now that you made it clear and simple for me regarding how we could go about this !

anutosh491:
CC @jksuom do you like the changes now, I'm quite satisfied with the changes overall , would like to have your thoughts here !

anutosh491:
cc @jksuom

anutosh491:
> f = floor(sin(x)/x/2)

Wow , no clue how you ever able that ! But i'll work on that now.

anutosh491:
> ```
> >>> f = floor(sin(x)/x/2)
> >>> f.as_leading_term(x)
> 1/2
> ```
> 

I guess I have addressed this case ,you could let me know if you know any other such cases which I might have overlooked :)
Though I am quite certain there shouldn't be other cases as I ended up trying a decent number of random limits on the branch based on this error !

anutosh491:
cc @jksuom

anutosh491:
Dear Kalevi ,

I would also be glad if you could take some time out to go through my draft Gsoc 22' proposal on the [`Current Application wiki page`](https://github.com/sympy/sympy/wiki/GSoC-2022-Current-Applications) which is the only one there currently.

Through my knowledge and experience of contributing since past 4-5 months, I knew of 8-9 **domains/topics** which could be improved in **series** module of sympy .
Those are 

1. Improve `leading term` & `series` method for functions involving `branch cuts/ branch points`
2. Adding `nseries`, `aseries` and `leading term` support to special functions
3. Extending Functionality of `Multivariate series expansions` 
4. `Ring series` Improvements and Implementations
5. Extending coverage for `limit_seq.py` and `formal.py`
6. Fixing symbolic limit errors through `gruntz` and other gruntz errors
7. Refactor `Order` code and order.py to handle cases better
8. Improving `AccumBounds` for computing limits
9. Miscellaneous

Though I had posted about my project on the google group a week back or so and had addressed my doubts there ....... I am currently struggling with somethings where I would like some guidance and assistance from your side !
These are the follows

- I am not sure which of these topics/domains would sympy like to have first and give priority too . As I was not sure regarding the priority , I eventually went on explaining the issues and proposed solutions/ideas /pseudocode to tackle them so that I could just get rid of any topic if sympy doesn't need it in the near future( or rather the Gsoc coding period). Hence I would like some assistance from your side here . **Even before deep diving in a topic, If you could just give a quick read across the topic and the issues I plan to address , I would be glad as you could get back to me and let me know how to arrange these ideas.**

- Also as I tried my best to cover what I would be addressing for each topic, the proposed ideas section now range from page **12** to page **85** , which is quite a bit and I am not sure how much of this could be addressed in a Gsoc coding period and hence a priority analysis is quite important here ,so that I could know what ideas can be kept for the summer and what could be kept for future Gsoc projects and contributions in general.

- I obviously can't form a timeline /deadlines to be met section unless we finalize which ideas I would be addressing . I would also think out the time and weeks I would devote to each topic after this .

- Last but not least , I would also like to have clarity on the project length . Though I would prefer a **175 hour** project variant , I am quite comfortable with the other one too (**350 hours**) and obviously we end up addressing more content there. 


**I know this might not be the best place to contact you for the above reason but I wasn't quite sure regarding the platform where I could reach out to you and as your active on github I thought of writing to you here, hence I'll be marking the conversation as of topic as soon as I hear from your side. Apologies in advance !**

jksuom:
> which of these topics/domains would sympy like to have first and give priority too

I think that the question is not about priorities but about "which issues will you be able to solve". It is often hard to tell in advance how much work will be needed. Fixing one issue may reveal other issues that also need to be solved. Therefore you should have at least an experimental implementation ready in order to see all the implications. That implementation should also be described in the proposal.

I personally think that the series and leading term methods should be improved (1) and implemented (2). Multivariate series (3)  and ring series (4) could be handled later. As to gruntz (6), I'm not sure how you would be planning to solve the issues with symbolic coefficients as the algorithm was designed to work with explicit real constants. It also looks like Order (7) has problems but I am not clear how you would be planning to fix those.

anutosh491:
cc @jksuom do you see any other cases of concern here otherwise this could go in I guess ! Also I've continued our discussion about the proposal on the gitter channel , would like to hear from you on the issue I mentioned there . Till then i'll move onto some other issue , i'll be addressing!

anutosh491:
This is ready from my side maybe @0sidharth could have a look through the changes and give us the green flag here !

sympy/functions/elementary/integers.py
```diff
@@ -144,6 +144,12 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
         arg = self.args[0]
         arg0 = arg.subs(x, 0)
         r = self.subs(x, 0)
+        if arg0 is S.NaN:
+            limit = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
```
jksuom, line 148: A global name like `limit` is not a good choice for a local variable. If the limit exists, it means that `arg` can be extended continuously to `x = 0` and the limit is its value at 0. So this could be `arg0` redefined.

sympy/functions/elementary/integers.py
```diff
@@ -144,6 +144,12 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
...
+            ndir = arg.dir(x, cdir=cdir)
+            if not limit.is_integer:
```
jksuom, line 150: `is.integer` does not work with floating point numbers. There is no guarantee that the limit would never be a `Float`. It seems that the test used in the next block, `arg0 == r`, would be more reliable. For that purpose also `r` should be redefined, probably as `floor(arg0)` (where `arg0` is the limit). There is no need to repeat the code here. We could just let the control pass to the next block after `arg0` and `r` have been redefined.

anutosh491: Good catch this , had completely overlooked this and didn't cross my mind even once.
> There is no guarantee that the limit would never be a Float

I also get what you're trying to convey . The redefining approach will also make it easier to debug or catch errors in the future .Will make the changes. Thanks !

anutosh491:
CC @jksuom

sympy/functions/elementary/integers.py
anutosh491: Resolving as addressed !

sympy/functions/elementary/integers.py
```diff
@@ -162,13 +165,16 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
     def _eval_nseries(self, x, n, logx, cdir=0):
         arg = self.args[0]
         arg0 = arg.subs(x, 0)
+        r = self.subs(x, 0)
+        if arg0 is S.NaN:
+            arg0 = arg._eval_nseries(x, n, logx, cdir).subs(x, 0)
```
jksuom, line 170: What is the purpose of starting with a series expansion? `subs(x, 0)` will wipe out all positive powers of `x` and only the constant term will remain (unless there are negative powers).

anutosh491: I mean why not , I somewhat like the idea of having a series here . This is because of two things 

1) Firstly we address this , I don't think a series through floor should give out `nan`
```
>>> floor(sin(x)/x).series(x, 0, 100, cdir = 1)  
nan
```

2) Secondly as leading term didn't have this case already implemented all wrong results were through gruntz , so having a series would help if anytime the leading term implementation would be caught up due to something
```
>>>gruntz(floor(sin(x)/x), x, 0)
nan                  # on master
>>> gruntz(floor(sin(x)/x), x, 0)
0
```
And the initial idea behind using `subs(x, 0)` was through a reply of yours 

The question I had asked was 
```
>>> # does it make sense to return series expansion of expr inside floor ??
>>> floor(sin(x)/x).series(x)
floor(-x**2/6 + x**4/120 + O(x**6)) + 1
```
To which you had replied 
> I think that the series expansions of floor and ceiling should contain only one constant term, no powers of x.

anutosh491: If I remember correctly, even wolfram does plot 0 to be the series approximation or expansion for something like `floor(sin(x)/x)` when expanded till order 20 . So I'm not fully sure how we could change what I've implemented and possibly return an alternative to this . **Do you see any other approach which we could adapt here ??**

I'll possibly have a look regarding what other CAS would return in this case !

jksuom: I would consider using `limit'.

anutosh491: Ohhh ok do you mean something like this rather than using `_eval_nseries` and then doing the `subs(x, 0)` step. 

```
        if arg0 is S.NaN:
            arg0 = arg.limit(x, 0, dir='-' if re(cdir).is_negative else '+')
            r = ceiling(arg0)
```
I guess this is what you mean , I'll make the change now !

anutosh491: I've made the change, maybe this is ready to go if all checks pass here , let me know if you think we can improve on the changes made here or have some doubts regarding the current implementation .

anutosh491: CC @jksuom

jksuom:
This looks good now, thanks.

anutosh491:
Thanks for the great reviews on this one !

PR #23270, opened 2022-03-22
Use Dimension objects instead of symbols/expressions in the dimensional dependecy dictionaries

Upabjojr:
Use `Dimension` objects instead of symbols/expressions in the dimensional dependecy dictionaries.

There was once a bug which prevented `Dimension` objects from having a proper hash, thus they could not be used as keys in a dictionary. This bug has since been fixed, so no need to hide the `Dimension` objects anymore.

NO ENTRY

### DIFF
--- a/doc/src/modules/physics/units/examples.rst
+++ b/doc/src/modules/physics/units/examples.rst
@@ -38,9 +38,9 @@ should be :math:`L^3 M^{-1} T^{-2}`.
     >>> F
     Dimension(acceleration*mass)
     >>> dimsys_SI.get_dimensional_dependencies(F)
-    {'length': 1, 'mass': 1, 'time': -2}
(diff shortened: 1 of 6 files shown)


doc/src/modules/physics/units/examples.rst
```diff
@@ -38,9 +38,9 @@ should be :math:`L^3 M^{-1} T^{-2}`.
     >>> F
     Dimension(acceleration*mass)
     >>> dimsys_SI.get_dimensional_dependencies(F)
-    {'length': 1, 'mass': 1, 'time': -2}
+    {Dimension(length): 1, Dimension(mass, M): 1, Dimension(time): -2}
```
smichr, line 41: What is the purpose of "M" (and similar) in some of the instances?

Upabjojr: The abbreviation of the name.

PR #23271, opened 2022-03-22
Make HNF require a `mode` argument.

### LINKED ISSUE #23260: Hermite normal form works in a strange way
It works flipped and transposed, and this part isn't documented. It took me really really long to figure it out. 

Here's the expected result of what's commonly understood as the [Hermite normal form](https://en.wikipedia.org/wiki/Hermite_normal_form). This is the row-style HNF. Notably, [MATLAB](https://www.mathworks.com/help/symbolic/hermiteform.html) and [Sagemath](https://doc.sagemath.org/html/en/reference/matrices/sage/matrix/matrix_integer_dense.html#sage.matrix.matrix_integer_dense.Matrix_integer_dense.hermite_form) work this way. 
Input: 
`[[12, 19, 28, 34], [19, 30, 44, 53]]`
Output: 
`[[1, 0, -4, -13], [0, 1, 4, 10]]`

There's also the column-style HNF, which works like this. 
Input: 
`[[12, 19], [19, 30], [28, 44], [34, 53]]`
Output: 
`[[1, 0], [0, 1], [-4, 4], [-13, 10]]`

With Sympy, the only way to find the expected result is like this. 
Input: 
`[[34, 53], [28, 44], [19, 30], [12, 19]]`
Output: 
`[[10, -13], [4, -4], [1, 0], [0, 1]]`

That's to say I feel like using
`hermite_normal_form (Matrix (A))`
but I have to type
`np.flip (hermite_normal_form (Matrix (np.flip (A)).T).T))`

Suggestions: 
Please add the option to make it work the good way (preferably making it default) and the current behavior should be documented clearly.

skieffer:

Resolves #23260 

The `hermite_normal_form()` functions (one under `matrices` and one under `polys`) now
require a `mode` argument.

This allows us to work within any of the (several) popular definitions of Hermite Normal Form.

#### Other comments

This changes behavior that was new in SymPy 1.10.
It will be considered a breaking change by users for whom `HNFMode.RIGHT` was the expected behavior.
It will be considered a bugfix by users who expected different behavior.

#### Release Notes

* matrices
  * HNF requires a mode argument
* polys
  * HNF requires a mode argument

### DIFF
--- a/sympy/matrices/normalforms.py
+++ b/sympy/matrices/normalforms.py
@@ -7,8 +7,12 @@ from sympy.polys.matrices.normalforms import (
         smith_normal_form as _snf,
         invariant_factors as _invf,
         hermite_normal_form as _hnf,
+        HNFMode as _HNFMode,
     )
 
+# Redefine here so that users intending only to work with Matrices do not
+# have to import from under polys.
+HNFMode = _HNFMode
 
 def _to_domain(m, domain=None):
     """Convert Matrix to DomainMatrix"""
@@ -67,7 +71,7 @@ def invariant_factors(m, domain=None):
     return factors
 
 
-def hermite_normal_form(A, *, D=None, check_rank=False):
+def hermite_normal_form(A, mode, *, D=None, check_rank=False, truncate=True):
     r"""
     Compute the Hermite Normal Form of a Matrix *A* of integers.
 
@@ -75,15 +79,18 @@ def hermite_normal_form(A, *, D=None, check_rank=False):
     ========
 
     >>> from sympy import Matrix
-    >>> from sympy.matrices.normalforms import hermite_normal_form
+    >>> from sympy.matrices.normalforms import hermite_normal_form, HNFMode
     >>> m = Matrix([[12, 6, 4], [3, 9, 6], [2, 16, 14]])
-    >>> print(hermite_normal_form(m))
+    >>> print(hermite_normal_form(m, HNFMode.RIGHT))
     Matrix([[10, 0, 2], [0, 15, 3], [0, 0, 2]])
 
     Parameters
     ==========
 
     A : $m \times n$ ``Matrix`` of integers.
+
+    mode : :py:class:`~HNFMode`
+        Say on which side of the pivots the reduced entries should go.
 
     D : int, optional
         Let $W$ be the HNF of *A*. If known in advance, a positive integer *D*
@@ -97,6 +104,10 @@ def hermite_normal_form(A, *, D=None, check_rank=False):
         checking it for you. If you do want this to be checked (and the
         ordinary, non-modulo *D* algorithm to be used if the check fails), then
         set *check_rank* to ``True``.
+
+    truncate : boolean, optional (default=True)
+        If ``True``, chop off rows or columns of zeros, according to the
+        *mode*. Setting ``False`` is not available when working mod *D*.
 
     Returns
     =======
@@ -124,4 +135,5 @@ def hermite_normal_form(A, *, D=None, check_rank=False):
     # Accept any of Python int, SymPy Integer, and ZZ itself:
     if D is not None and not ZZ.of_type(D):
         D = ZZ(int(D))
-    return _hnf(A._rep, D=D, check_rank=check_rank).to_Matrix()
+    return _hnf(A._rep, mode,
+                D=D, check_rank=check_rank, truncate=truncate).to_Matrix()
--- a/sympy/polys/matrices/domainmatrix.py
+++ b/sympy/polys/matrices/domainmatrix.py
@@ -679,6 +679,22 @@ class DomainMatrix:
         """Matrix transpose of ``self``"""
         return self.from_rep(self.rep.transpose())
 
+    def flip_top_to_bottom(self):
+        return DomainMatrix(list(reversed(self.to_list())), self.shape, self.domain)
+
+    def rotate_clockwise(self):
+        return self.flip_top_to_bottom().transpose()
+
+    def rotate_counterclockwise(self):
+        return self.transpose().flip_top_to_bottom()
+
+    def rotate_180(self):
+        return self.rotate_clockwise().rotate_clockwise()
+
+    def anti_transpose(self):
+        """Flip across the *other* diagonal. """
+        return self.flip_top_to_bottom().rotate_counterclockwise()
+
     def flat(self):
         rows, cols = self.shape
         return [self[i,j].element for i in range(rows) for j in range(cols)]
--- a/sympy/polys/matrices/normalforms.py
+++ b/sympy/polys/matrices/normalforms.py
@@ -1,6 +1,7 @@
 '''Functions returning normal forms of matrices'''
 
 from collections import defaultdict
+from enum import Enum
 
 from .domainmatrix import DomainMatrix
 from .exceptions import DMDomainError, DMShapeError
@@ -174,14 +175,17 @@ def _gcdex(a, b):
     return x, y, g
 
 
-def _hermite_normal_form(A):
+def _hermite_normal_form(A, truncate=True):
     r"""
-    Compute the Hermite Normal Form of DomainMatrix *A* over :ref:`ZZ`.
+    Compute the Hermite Normal Form of DomainMatrix *A* over :ref:`ZZ`,
+    in :py:class:`~HNFMode` ``RIGHT``.
 
     Parameters
     ==========
 
     A : :py:class:`~.DomainMatrix` over domain :ref:`ZZ`.
+    truncate : boolean, optional (default=True)
+        If true, chop off columns of zeros on the left.
 
     Returns
     =======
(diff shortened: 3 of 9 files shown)


oscarbenjamin:
This all seems a bit like overkill to me. Can we not just document the different conventions and how to convert between them?

I don't think we need methods like `M.flip_top_to_bottom()` when they're all things that can be done with slicing and transpose like `M[::-1, :]` etc (which already works efficiently for sparse and dense matrices).

Neither DomainMatrix nor Matrix has a method for computing the  HNF yet so it's not too late to add those methods and use a different convention by default in those. My preference would be to aim for consistency with sage and python_flint which are both in the Python ecosystem. Also the convention suggested in the linked issue is consistent with Matlab, Maple and Mathematica.

You've mentioned Cohen and PariGP as exceptions. I'm not too familiar with HNF in general and I don't know what other implementations there are to compare with but from my survey it seems like there is a fairly common convention even if it is arbitrary.

smichr:
> I don't think we need methods like `M.flip_top_to_bottom()` when they're all things that can be done with slicing and transpose like `M[::-1, :]` etc (which already works efficiently for sparse and dense matrices).

I don't have a strong feeling against this. While row reversal is easy to think about. It takes me more than a second to figure out things like rotations. And things that take more than a second I would write down on a reference page but would not need to if the method-in-words existed (like `rotate_clockwise`).

Rather than making this a breaking change, can the mode default to its current value? Anyone who reads the help/docstring will see that there is a mode option which they should understand and use for their own purposes.

skieffer:
> Can we not just document the different conventions and how to convert between them?

That wouldn't be unreasonable, but I also don't think it hurts to provide a ready-made implementation.

I haven't been able to do a very large literature survey, but, as a start:

* Hermite himself (1851): reduced entries go above the pivot
* Minkowski (1893): reduced entries go on the right
* Chatelet (1911): below
* the Kannen-Bachem paper (1979): left
* Donaldson (1979): above
* the algorithm in Pohst-Zassenhaus (1989): left
* the algorithms in both of Cohen's books (1993 and 2000): right
* Wikipedia: left or above (no preference)

Considering how chaotic the situation seems, I think providing an option to do the flipping and rotating for the user is warranted.

I wouldn't, for example, suggest providing the same for Gaussian elimination. I don't know of any conflicting definitions of RREF. (Probably putting the word "row" in the name was a good idea!)

I think people's expectations will be a function of how they got here. If they came to SymPy from another CAS, then, given all the examples you have mentioned, they may well expect the reduced entries to go above the pivot. But if they came here directly from the literature, they might expect something else.

skieffer:
> things that can be done with slicing and transpose like M[::-1, :] etc (which already works efficiently for sparse and dense matrices).

The methods I added to DomainMatrix are very inefficient. If there's agreement on providing this at all, I would redo them.

skieffer:
> Rather than making this a breaking change, can the mode default to its current value? Anyone who reads the help/docstring will see that there is a mode option which they should understand and use for their own purposes.

Yes, making it a keyword arg with a default value, along with the docstring, would be enough I suppose.

oscarbenjamin:
If there are to be convenience methods for doing these transformations on matrices then the names should be consistent with other languages/libraries e.g. both matlab and numpy use the same names `flipud`, `rot90` etc.:
https://www.mathworks.com/help/matlab/ref/rot90.html
https://numpy.org/doc/stable/reference/generated/numpy.rot90.html

It looks like Mathematica just has `Reverse` and `Transpose` and same for Maple except reverse is called `FlipDimension`.

As for the default convention used to define HNF I think it's more important to be consistent with other usable computing environments than textbooks. It's much more likely that someone would try to port some code from say Matlab to SymPy (or vice-versa) than that someone would want to compare the output with a textbook.

skieffer:
I guess I'm contemplating the educational use case. You have the student who's learning the theory, and then turns to a CAS to try things out and see the theory in action. It would be nice if they could easily make the CAS conform to the conventions their particular textbook is using.

For this, the default setting is not as important as having the ability to change it.

Using the same names as numpy sounds like the right plan.

Question: When it comes to composing transformations, how do you feel about the tradeoff of speed versus sticking to a small set of methods? For example: a `rot180` operation implemented directly would be faster than calling `rot90` twice. I guess numpy didn't see the need, and maybe it's best to follow suit.

smichr:
> speed versus sticking to a small set of methods

I would consider this in the context of what you would be doing with the matrices. Re-arrangement will be a small part of the total work for a given meaningful matrix operation, won't it? In that case, the time taken for that would be small relative to the other operations, wouldn't it? If so, then I would stick with using `rot90` twice.

Out of curiosity, how much faster and how much more complicated is the routine to do a 180 directly?

oscarbenjamin:
> When it comes to composing transformations, how do you feel about the tradeoff of speed versus sticking to a small set of methods?

Methods like this are a convenience that I don't see the need for. Certainly if this is to be added then there's no need to add any nontrivial code. Something like `rot180` is just `M[::-1,::-1]`: I don't see why any implementation should be more complicated than that.

In particular I want to minimise the number of methods that are implemented separately by SDM and DDM because those have to be duplicated. I also generally want to minimise the number of methods that DomainMatrix has as well. DomainMatrix is not intended to be a high-level class (that's what Matrix is for).

At the DomainMatrix level I would prefer to just have a single `hnf` method that uses the convention most commonly found elsewhere and clearly documents that. If a `hnf` method is added to Matrix then we can think about whether it makes sense to add these kinds of convenience routines.

smichr:
> I don't see why any implementation should be more complicated than that.

That sounds good, too. Any optimization (if needed) would/should be handled by the routines that provide those slices, then. So all the sugar manipulations would be written in terms of transpose and reversal-slicing.

skieffer:
Generally speaking I'm in favor of the syntactic sugar methods, both for the reason @smichr mentions (they save you having to think it through each time) and because they make for self-documenting code. When a person reading your code encounters `m[::-1,::-1]` they have to decipher what you're doing. When they encounter `m.rot180()` they know immediately what you're doing.

For the issue at hand I don't care too much. Internally, the `hermite_normal_form()` function could just do the necessary combinations of transpose and slice.

sympy/polys/matrices/normalforms.py
```diff
@@ -336,7 +339,97 @@ def add_columns_mod_R(m, R, i, j, a, b, c, d):
     return DomainMatrix(W, (m, m), ZZ).to_dense()
 
 
-def hermite_normal_form(A, *, D=None, check_rank=False):
+class HNFMode(Enum):
+    r"""
```
skieffer, line 343: What about the `Enum`? It's possible that with this I was thinking more like a software engineer, and less like a user.

My impression is that SymPy doesn't use a lot of `Enum`s. Typically a function or method that needs a mode switch seems to take a string. E.g. `limit` accepts `dir="+"` or `dir="-"`. `DomainMatrix` accepts `fmt="sparse"` or `fmt="dense"`.

From the user's perspective, it would certainly be handier to be able to just write `mode="a"` than to have to (a) import an enum class and then (b) write the much longer `mode=HNFMode.ABOVE`. In retrospect I'm thinking I might like the string approach better. Any opinions?

smichr: I don't know anything about Enum but I am supportive of the text mode. Possibilities include "NEWS" to indicate a map direction (N=north/above pivot); or glyph-like "^v<>" to point where the reduced entries should go.

asmeurer: Can we set the Enum values to strings? That could be a middle ground. I think typing a string is more convenient as a user, but the Enum can also be useful in that it makes it easy to see what the options are from tab completion. 

Allowing different equivalent variants of the same string is useful too (like `"ABOVE"` or `"above"`, or @smichr's suggestions). I don't know if Enum supports that, but it could accept it in the function.

sympy/matrices/normalforms.py
```diff
@@ -67,24 +71,27 @@ def invariant_factors(m, domain=None):
     return factors
 
 
-def hermite_normal_form(A, *, D=None, check_rank=False):
+def hermite_normal_form(A, mode, *, D=None, check_rank=False, truncate=True):
```
asmeurer, line 74: Does this need to be added as a required argument without a default value? This is a backwards incompatible change.

skieffer: I think a kwarg with default value would be fine. Originally I felt it was important to force the user to be aware of the issue by making it a required arg, but perhaps that's overdoing it. (And not worth the compatibility break.)

sympy/polys/matrices/normalforms.py
skieffer: I like the compass and glyph-like alternative strings. A user could easily forget whether it was supposed to be "above" or "up", but "^" is clear.

skieffer: > Can we set the Enum values to strings?

I'm new to `Enum`s myself, and don't know if there's a handy way to check them simultaneously as a string and as an actual `Enum` member. You can set strings as values, but the enum members still do not test equal to strings:

```python
>>> from enum import Enum
>>> class Foo(Enum):
...     BAR = "BAR"
...     SPAM = "SPAM"
>>> "BAR" == Foo.BAR
False
```

If you want a single test that accepts both strings and enum members, maybe an `or` is the best you can do?

```python
>>> test = lambda x: x is Foo.BAR or x == Foo.BAR.value
>>> test("BAR")
True
>>> test(Foo.BAR)
True
```

sympy/polys/matrices/domainmatrix.py
```diff
@@ -679,6 +679,22 @@ def transpose(self):
         """Matrix transpose of ``self``"""
         return self.from_rep(self.rep.transpose())
 
+    def flip_top_to_bottom(self):
+        return DomainMatrix(list(reversed(self.to_list())), self.shape, self.domain)
+
+    def rotate_clockwise(self):
```
smichr, line 685: FYI, `Matrix` aleady has the `rot90` method.

PR #23278, opened 2022-03-24
Make all Greek capitals upright in LaTeX output

TomFryers:
Greek capital letters are typically typeset upright. Those that have TeX commands do this automatically, but for those that don’t, we need `\mathrm`.

NO ENTRY


asmeurer:
There was a discussion at https://github.com/sympy/sympy/issues/15934#issuecomment-921222731 about the subtleties of \mathrm that should be considered here

asmeurer:
How does LaTeX (amsmath?) itself render things like \Pi?

TomFryers:
LaTeX

TomFryers:
LaTeX renders Greek capitals upright by default. For example,
```latex
\documentclass{article}
\begin{document}
\[A B \Gamma \Delta E Z H \Theta I K \Lambda M N \Xi O \Pi P \Sigma T \Upsilon \Phi X \Psi \Omega\]
\[\mathrm{A} \mathrm{B} \Gamma \Delta \mathrm{E} \mathrm{Z} \mathrm{H} \Theta \mathrm{I} \mathrm{K} \Lambda \mathrm{M} \mathrm{N} \Xi \mathrm{O} \Pi \mathrm{P} \Sigma \mathrm{T} \Upsilon \Phi \mathrm{X} \Psi \Omega\]
\end{document}
```
produces this:
Letters
`amsmath` seems to make no difference. XeTeX behaves the same, except it opens up to possibility of the `unicode-math` package, which also provides `\Alpha`, `\Beta`, etc. macros. These are upright by default.

Re `\mathrm` vs. e.g. `\text`, I think `\mathrm` is probably still the best option, but it’s certainly not ideal.

asmeurer:
Sorry for being unclear. I mean what does LaTeX do under the hood to make the greek letters upright. Is `\Pi` a shorthand for `\mathrm{Π}`? Or is this simply an artifact of the way Π is represented in the Computer Modern font? Or something else?

TomFryers:
I don’t fully understand the LaTeX font mechanisms, but it seems that `\Pi` is defined [here](https://github.com/latex3/latex2e/blob/ba246a87310372784ed0a2fb9d01ff36e6735176/base/fontdef.dtx#L917) as
```latex
\DeclareMathSymbol{\Pi}{\mathalpha}{operators}{"05}
```
where `operators` [is defined as roman](https://github.com/latex3/latex2e/blob/ba246a87310372784ed0a2fb9d01ff36e6735176/base/fontdef.dtx#L644):
```latex
\DeclareSymbolFont{operators}   {OT1}{cmr} {m}{n}
```
In contrast, the lowercase Greek letters use `letters`, which is
```latex
\DeclareSymbolFont{letters}     {OML}{cmm} {m}{it}
```
Then `\mathrm` is defined as switching to `operators`:
```latex
\DeclareSymbolFontAlphabet{\mathrm}    {operators}
```

sympy/printing/latex.py
```diff
@@ -39,35 +39,35 @@
...
-    'Kappa': 'K',
+    'Iota': r'\mathrm{I}',
+    'Kappa': r'\mathrm{K}',
     'Lambda': r'\Lambda',
-    'Mu': 'M',
-    'Nu': 'N',
+    'Mu': r'\mathrm{M}',
+    'Nu': r'\mathrm{N}',
     'Xi': r'\Xi',
     'omicron': 'o',
```
oscargus, line 56:
```suggestion
    'omicron': r'\mathrm{o}',
```
for consistency?

(I note that TeXStudio uses the default for the their Greek list, so no `\mathrm`, but I understand why that may be incorrect.)

oscargus:
I think that this makes sense, all Greek characters are upright (that is why we have e.g. `\varTheta` etc.

Regarding `\text` vs `\mathrm` I have actually changed my mind after years of encouraging `\text`. Turns out that one should not interpret the `rm` part of `\mathrm` that strict. However, the drawback is that Beamer does that, so using `\mathrm` with the default Beamer style will mess things up.

I think that there should be a setting to determine which of `\mathrm`, `\text`, and `\operatorname` is preferred for "upright". (Drawback with `\mathrm` is that dashes will be interpreted as minus, so it is not as easy as just having two options...)

sympy/printing/latex.py
TomFryers: I think it is just the capitals that are supposed to be upright. See e.g. [this answer](https://tex.stackexchange.com/a/119262/200616).

asmeurer: Curiously, the `O` in big O notation is supposedly an omicron, but I'm not sure I've ever seen it presented as upright.

oscargus:
Here it is also worth noting that not all "Greek" capital letters actually have a LaTeX command. For example, `\Alpha` doesn't really work in LaTeX. So anything we do can be based on our definition and if someone doesn't like it they can just use `A` instead.

So if no one objects, I will merge this in a few days.

oscargus:
Closing, reopening and enabling auto-merge.

PR #23291, opened 2022-03-25
Calculus : Added function to calculate minimum periodicity.

### LINKED ISSUE #22935: Periodicity of trigonometric expression incorrect .

AhmedTareek:

Fixes #22935 

The Periodicity function gives a correct period but not always the smallest period so by using this function we can get the minimum period always

#### Release Notes

* calculus
  * Added a new function to get the minimum period.

### DIFF
--- a/sympy/calculus/util.py
+++ b/sympy/calculus/util.py
@@ -841,3 +841,34 @@ def minimum(f, symbol, domain=S.Reals):
         return function_range(f, symbol, domain).inf
     else:
         raise ValueError("%s is not a valid symbol." % symbol)
+
+        
+def min_period(f):
+    old_period = periodicity(f, x)
+    if old_period is None:
+        return None
+    equal_spaces = old_period / 20
+    equal_spaces_list = []
+    y = 0
+    for i in range(20):
+        equal_spaces_list.append(y)
+        y += equal_spaces
+    period_values_list = []
+    for i in equal_spaces_list:
+        period_values_list.append((f.subs(x, i).evalf()))
+    num_of_periods = 1
+    startidx = 0
+    p1 = 1
+    p2 = startidx
+    while p1 < 20:
+        if period_values_list[p1] == period_values_list[p2]:
+            final = p1
+            while period_values_list[p1] == period_values_list[p2] and (p2 < final) and (p1 < 20) and (p2 < 20):
+                p1 += 1
+                p2 += 1
+                if final - p2 == 1:
+                    num_of_periods += 1
+                    break
+        p1 += 1
+    return old_period / num_of_periods
+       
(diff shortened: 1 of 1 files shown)


smichr:
Can the period ever be less than half of the computed period?

faze-geek:
> Can the period ever be less than half of the computed period?

Not sure but when I had raised the issue this pr intends to solve ,I had tried a bunch of trigonometric functions and their manipulations , but never got a period which was less than half of the actual period.

asmeurer:
Can you explain what this algorithm is doing? What is the point of the number 20 that is used throughout?

AhmedTareek:
> Can you explain what this algorithm is doing? What is the point of the number 20 that is used throughout?
its mainly built on the periodicity function and try and error 
I explained it here https://github.com/sympy/sympy/issues/22935#issuecomment-1078921296
20 is just a number I chose it could be any number
another point for improving the algorithm which I did not implement yet is to chose a non zero number and track it for f(x+p) = f(x) where x is the chosen number and p is the period I deduced

anutosh491:
> > Can you explain what this algorithm is doing? What is the point of the number 20 that is used throughout?
>
> its mainly built on the periodicity function and try and error I explained it here [#22935 (comment)](https://github.com/sympy/sympy/issues/22935#issuecomment-1078921296) 20 is just a number I chose it could be any number another point for improving the algorithm which I did not implement yet is to chose a non zero number and track it for f(x+p) = f(x) where x is the chosen number and p is the period I deduced

I'm not sure of this logic and have the same doubt as reported above , building an algorithm based on a randomly chosen number doesn't look like the best approach to me ! The thought behind the algorithm looks some what okay though .

smichr:
Here is an algebraic approach -- some investigation would have to be done to see how well it fairs for more complicate expressions. We want to find the periodicity of $cos(x)^2$ so we seek a value $y$ which gives us the same value and derivative of the function when added to $x$:
```python
>>> from sympy.abc import x, y, a, b
>>> eq = cos(x)
>>> eqs = [eq - eq.subs(x,x+y)]
>>> eqs.append(eqs[0].diff(x))
>>> e = Tuple(*[i.rewrite(exp).expand().subs(exp(I*x),a).subs(exp(I*y),b) for i in eqs])
>>> gb = groebner([i.as_numer_denom()[0] for i in e],a,b); sol = solve(gb[-1], dict=True)
>>> [i for i in sol if e[0].subs(i).equals(e[1].subs(i))]
[{b: -1}, {b: 1}]
>>> [solve(exp(I*y)-i[b])[0] for i in _]
[pi, 0]
```
And we are only interested in the non-zero value.

PR #23292, opened 2022-03-25
WIP: feat(physics.units): `quantity_simplify()` can optionally derive the most appropriate unit prefix with `add_prefix=True`

dyc3:
Adds a new optional flag to `quantity_simplify()`: `add_prefix` that allows it
to derive the most appropriate unit prefix.

#### Other comments
It's unclear whether or not we should grab the prefixed unit from the unit system as a single `Quantity` (eg. `microsecond` or `kilometer`),
or if it's better to have prefixes as `Prefix*Quantity` (eg. `micro*second` or `kilo*meter`).

#### Release Notes

* physics.units
  * `quantity_simplify()` can add prefixes to the resulting units with the optional flag `add_prefix=True` to make the resulting value more human readable.

### DIFF
--- a/sympy/physics/units/util.py
+++ b/sympy/physics/units/util.py
@@ -4,6 +4,9 @@ Several methods to simplify expressions involving unit objects.
 from functools import reduce
 from collections.abc import Iterable
 from typing import Optional
+from matplotlib.pyplot import sca
+
+from matplotlib.style import available
 
 from sympy import default_sort_key
 from sympy.core.add import Add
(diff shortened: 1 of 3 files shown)


sympy/physics/units/util.py
```diff
@@ -188,6 +196,40 @@ def quantity_simplify(expr, across_dimensions: bool=False, unit_system=None):
...
+
+        # find the prefixed unit that matches
+        matching_units = list(filter(lambda u: u.is_prefixed and u.scale_factor == closest_prefix.scale_factor and u.name.name.endswith(unit.name.name), unit_system._units))
+        assert len(matching_units) == 1
+        prefixed_unit: Quantity = matching_units[0]
+
+        scale /= closest_prefix.scale_factor
+        print("before:", expr)
+        expr = scale * prefixed_unit
+        print("after:", expr)
```
asmeurer, line 231: These print statements shouldn't be here.

PR #22522, opened 2021-11-18
core/basic: argument invariance for all Atom classes - class

ThePauliPrinciple:
#### Brief description of what is fixed or changed
The following identities now hold for `Atom` classes:
```python
assert obj.func == AtomSubClass
assert AtomSubClass == obj.func
assert hash(AtomSubClass) == hash(obj.func)
assert {AtomSubClass: True}[obj.func]
assert {obj.func: True}[AtomSubClass]
assert obj.func() == obj
assert obj.func(*obj.args) == obj
assert str(obj.func) == str(AtomSubClass)
```
These seem to all be needed since several parts of SymPy
assumes that `obj.func` returns `obj.__class__`, although it
is not needed for every class.

Several classes still behave as previously, these are:

`Quantity`
`BaseVector`
`BaseDyadic`
`BaseScalar`

It seems that these classes need a redesign rather than
a fix to their argument invariance, hence the previous
behaviour was maintained.

#### Other comments
It might be the case that some "solutions" provided to issues I encountered are actually not needed as the issues were only introduced due to changes I made (and reverted).

```python
@property
def func(self):
    return lambda: self
```
Has several intertwined issues and I expanded from that. In particular:
* printing of `obj.func`
* `Singleton`s and their "create once" behaviour
* comparisons of `obj.func == obj.class`
* use of `obj.__class__` hash for comparison with `obj.func` hash (e.g. in `dict` lookups)

were use cases in SymPy that were assumed to work as if `obj.func` always returns
`obj.__class__`.

#### Release Notes

* core
  * Atom classes now correctly implement the invariance `obj.func(*obj.args)`

PR #23293, opened 2022-03-25
Glossary

asmeurer:

This adds a glossary of SymPy terms to the documentation. This was suggested in a few places. 

I have still not written all the definitions here. I am looking for feedback on terms that should be included here. If you have any suggestions, please let me know. 

#### Release Notes

- other
  - Add a glossary of common SymPy terminology to the documentation.


bertiewooster:
Suggested terms to include:

- Atom
- Canonical
- GreaterThan, StrictGreaterThan; LessThan, StrictLessThan
- [negated (relationship)](https://docs.sympy.org/latest/modules/core.html?highlight=core%20relational#sympy.core.relational.Relational.negated), reversed, strict, weak
- Unequality

asmeurer:
It's not clear to me that the last three bullet points appear anywhere other than in the documentation of themselves. Some of them, like the relational class names, should really never be seen by users or mentioned in docs anywhere. Note that the point of the glossary isn't to documente every class or function. The docstrings themselves do that just fine. The point is to document terms that might appear in many places. Most things in the glossary are not expected to be functions at all, except for a few core examples like `Expr` that would appear everywhere (I also included `oo` and `zoo` because they are confusingly named).

Atom and canonical could be helpful (although "atom" is currently not very well defined in my opinion).

oscarbenjamin:
> Atom and canonical could be helpful (although "atom" is currently not very well defined in my opinion).

All the more reason to say something about it. If the term is used but not well defined then that should be made clear.

asmeurer:
>> Atom and canonical could be helpful (although "atom" is currently not very well defined in my opinion).

> All the more reason to say something about it. If the term is used but not well defined then that should be made clear.

That's fair, although it would be nice to have a decision about https://github.com/sympy/sympy/pull/22522 to really pin down a good definition. Right now "atoms" really are distinct things because they are treated differently. But with the change suggested in that PR, the difference is not that important.

bertiewooster:
> It's not clear to me that the last three bullet points appear anywhere other than in the documentation of themselves. Some of them, like the relational class names, should really never be seen by users or mentioned in docs anywhere. Note that the point of the glossary isn't to documente every class or function. The docstrings themselves do that just fine. The point is to document terms that might appear in many places.

Understood. Regarding GreaterThan etc., my thinking was that that term is not necessarily taught in school as `>=`, rather it may be taught as `>`, so the term's usage in SymPy might confuse some people.

asmeurer:
I would rather just change the class names to be less confusing.

asmeurer:
This is now ready for review.

asmeurer:
You can see a preview of the build using the new docs preview CI (click "Click here to see a preview of the documentation" link in the CI builds). Here is the build for this PR https://output.circle-artifacts.com/output/job/a9fa0426-c6b1-4be1-a41d-dc2c01e523a0/artifacts/0/doc/_build/html/explanation/glossary.html

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,743 @@
...
+
+`args`
+
+    The *`args`* property of a SymPy {term}`expression` `expr` is a tuple of
+    the top-level {term}`subexpressions <subexpression>` used to create
+    `expr`. They are the arguments to the class used to create the expression.
+    The args of any expression can be obtained by the `.args` attribute. For
+    example, `(1 + x*y).args` is `(1, x*y)`, because it equals `Add(1, x*y)`.
+    The `args` together with {term}`func` completely define an expression. It
+    is always possible to walk the {term}`expression tree` and extract any
```
bertiewooster, line 58: Recommend changing "_walk_ the expression tree" with "_traverse_" to

- make the analogy with traversing the DOM (document object model), which is a common phrase in web development
- be more inclusive (less ableist)

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,743 @@
...
+    sometimes called the "new assumptions", allows assumptions to be assumed
+    using separate predicate objects like `Q.positive` and does queries using
+    the {func}`~.ask` function. The newer assumptions system is able to
+    support more complex queries, but is also not as well developed as the
+    older one. Most users of SymPy should prefer the older assumptions system
+    at this time.
+
+Atom
+
+    An *atom* is an expression whose {term}`args` is the empty tuple `()`.
```
bertiewooster, line 102: Suggest giving an example of an atom in context. Also, this might be a good place to give the non-intuitive case where, for example, an atom is `x >= 0.3` from `(x >= 0.3) & (x <= 1.0)`.

bertiewooster (changes requested):
I reviewed A-C so far :-) Nice work! I think this page will be a really good starting point for understanding SymPy, not only terminology, but also how different parts relate (for example, how `Expr` is a subclass of `Basic`). Just a thought: maybe we point people to it at the end of the tutorial?

doc/src/explanation/glossary.md
asmeurer: Would you be happy with language that uses both, like "walk/traverse the expression tree"? Grepping, I see both uses in various places, including a function called `walk` that does just this.

The DOM is also a tree, so similar language will be used when talking about it. We can also add a glossary entry for walk/traverse. Another tree structure people may be familiar with is a file system. I think it depends on people's backgrounds which analogies they may be familiar with.

I linked to the Wikipedia article https://en.wikipedia.org/wiki/Tree_(data_structure) in the "expression tree" definition which lists the DOM as an example, but I'm happy to link to other, better references as well.

doc/src/explanation/glossary.md
asmeurer: But that isn't an atom:

```
>>> (x >= 0.3).args
(x, 0.300000000000000)
```

If there's some other documentation suggesting this is an atom, that should be fixed, because the definition I wrote here is what we should be using to mean "atom". I rather suspect there may be some instances where documentation elsewhere hasn't been written very rigorously and will need to be updated to match the glossary. We can consider it to be both descriptive and prescriptive.

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,743 @@
...
+    either `True`, `False`, or `None`.
+
+    Assumptions are either *assumed* or *queried*. For example, a symbol `x`
+    might be *assumed* to be positive by defining it as `x = symbols('x',
+    positive=True)`. Then an assumption might be *queried* on the expression
+    containing this symbol, like `(x + 1).is_real`, which in this case would
+    return `True`.
+
+    If no assumptions are assumed on a symbol, then by default symbols are
+    assumed to be general complex numbers. Setting assumptions is important
```
oscarbenjamin, line 82: This statement is tricky because symbols are not automatically assumed to be finite.

oscarbenjamin:
I would recommend keeping glossary entries as short as possible and using links to other documents that explain concepts in more detail. The most important thing missing from most docs are links to other parts of the docs. The glossary page should very briefly define each term but at the same time link through in a useful way for more detail.

For example the entry for args can just clarify that `x + y` is represented as `Add(x, y)` and that the args are `x` and `y`. Then there can be a see also for `func` and a reference to a page that describes the tree structure in detail including how to use it.

Similarly there is already a page that explains assumptions in detail so the assumptions entry here should link to that rather than try to explain everything. Or perhaps a new page is needed but either way the detail should be minimal in a glossary.

doc/src/explanation/glossary.md
asmeurer: I do want the text here to be correct, but I also don't want to bog down the definitions with technical details that wouldn't matter to most people. 

The point here is just to be that nothing is assumed, not really about the "default assumptions". Maybe I can tweak the wording a bit to not imply that finite is assumed by default.

asmeurer:
I've tried to keep the entries brief and no more than a paragraph, although some of the entries ended up being a little longer because they are more complex to explain. I'm sure a lot can be trimmed here, however. Writing a short definition is certainly harder than writing a long one. Can you be more specific about what you would remove from "args" and "assumptions", and anything else?

oscarbenjamin:
> Can you be more specific about what you would remove from "args" and "assumptions", and anything else?

Firstly rather than removing anything there should be links to anywhere that has more information. Those links are the most valuable thing that we can include in a short glossary entry. Likewise we should make sure that the glossary entries themselves can be linked to from other places e.g. like `See also: antiderivative`.

Secondly though I would limit the glossary entries to just 2-3 sentences. The value of a glossary page is that many things are listed in one place but that becomes unwieldy if each individual entry is not very short. If we want to provide more explanations for some of these things then we should make other pages that expand on these concepts and link to them from the glossary.

As an example:
```
Antiderivative
    An *antiderivative* of a function $f(x)$ with respect to $x$ is an
    function $F(x)$ such that $\frac{d}{dx}F(x) = f(x).$ It is also sometimes
    called an "indefinite integral" of $f(x)$, and written as $\int
    f(x)\,dx.$ Antiderivatives in SymPy can be computed with
    {func}`~.integrate`. Any given function has an infinite family of
    antiderivatives, typically obtained by adding an arbitrary constant, but
    {func}`~.integrate` only returns one
    such antiderivative, which is sometimes informally referred to as "the
    antiderivative" of $f(x)$. Note some sources call this the "primitive" of
    $f(x)$, but this terminology is not used in SymPy because it is not as
    universally used as "antiderivative", and because "primitive" has other
    meanings in mathematics.
```
The longest I would write here is
```
Antiderivative
    An *antiderivative* of a function $f(x)$ with respect to $x$ is an
    function $F(x)$ such that $\frac{d}{dx}F(x) = f(x).$ It is also sometimes
    called an "indefinite integral" of $f(x)$, and written as $\int
    f(x)\,dx.$ Antiderivatives in SymPy can be computed with
    {func}`~.integrate`.
```
I'm assuming that the reference to `integrate` here is a clickable link: that is where the explanation of what integrate does should be. There could also be a glossary entry for "indefinite integral" which should just say "see antiderivative" (with a clickable link).

I don't think there's any need to refer to a "primitive" or other rarely used terminology. Most people know what an antiderivative is and if they want more information they can go read wikipedia. A link from here to wikipedia would be reasonable.

asmeurer:
I guess I'm using the glossary in two ways here. One is to define terms for users. The other is to give a set of terms which should be used in the documentation and other SymPy resources, and terms which should be avoided for some reason or another. Maybe the latter use is better to go in the style guide or somewhere else. I think it is important that people use "antiderivative" instead of "primitive" in SymPy documentation (and there are several other examples of this in other definitions, like "expression" vs. "equation" and "symbolic" vs. "closed-form"). 

The other sentence is a long winded way of excusing the use of "*the* antiderivative" even if that's not a mathematically consistent thing to say strictly speaking. I agree it should either be shortened or removed.

asmeurer:
> Firstly rather than removing anything there should be links to anywhere that has more information. Those links are the most valuable thing that we can include in a short glossary entry. Likewise we should make sure that the glossary entries themselves can be linked to from other places e.g. like See also: antiderivative.

I've tried to link as much as possible. Some things just aren't documented very well elsewhere yet, or not documented at all. 

We definitely need to cross-reference to the glossary. I've left this as a future work. It's quite easy to do, just use ``` :term:`args` ``` (or ``` {term}`args` ``` in Markdown). See the comment at the top of the glossary.

oscarbenjamin:
> I've tried to link as much as possible. Some things just aren't documented very well elsewhere yet, or not documented at all.

That's true but we should be careful not to allow the glossary to become a new place to put documentation that doesn't really belong there just because there isn't a better alternative place. It would be better to make a page in guides or explanation that discusses integrals, antiderivatives, `integrate`, `Integral`, `dsolve`, `Integral.evalf` etc. That page can have more detail and discussion of nuances about the meaning of things like antiderivatives. The glossary should just have a clear definition and links for more detail.

It's important to bear in mind that after you've written this glossary it will only ever get longer over time as others add to it. More entries will be added and the existing entries will be added to. The standard should be set now that for more information the glossary links elsewhere. It's better to create very short pages that explain these things (and can be expanded later) than to add overly long glossary entries now.

asmeurer:
> I guess I'm using the glossary in two ways here. One is to define terms for users. The other is to give a set of terms which should be used in the documentation and other SymPy resources, and terms which should be avoided for some reason or another. Maybe the latter use is better to go in the style guide or somewhere else. My thinking was that it can be useful for users to see those distinctions as well. I think it is important that people use "antiderivative" instead of "primitive" in SymPy documentation (and there are several other examples of this in other definitions, like "expression" vs. "equation" and "symbolic" vs. "closed-form"), so we should write these down somewhere.

Also we should remember that one of the original motivations for the glossary was to do this sort of disambiguation for people. For instance, the entry for "solve" explains that it is not used to denote things like computing an integral, which people might often colloquially use "solve" for. 

Right now, if you Google ["sympy primitive"](https://www.google.com/search?q=sympy+primitive) the first result is the `primitive()` function and the second result is the calculus section of the tutorial, because it happens to say "antiderivative, or primitive". If you use the built-in search of the docs, all the results are `primitive` functions based on the polys definition. Given that there are people who use this term instead of "antiderivative", it may be hard for them to find what they are looking for unless they know that SymPy prefers the latter terminology. Perhaps it may actually be beneficial to add "primitive" itself to the glossary with the polys definition and mentioning that "antiderivative" is preferred for indefinite integrals.

> That's true but we should be careful not to allow the glossary to become a new place to put documentation that doesn't really belong there just because there isn't a better alternative place. It would be better to make a page in guides or explanation that discusses integrals, antiderivatives, integrate, Integral, dsolve, Integral.evalf etc. That page can have more detail and discussion of nuances about the meaning of things like antiderivatives. The glossary should just have a clear definition and links for more detail.
>
> It's important to bear in mind that after you've written this glossary it will only ever get longer over time as others add to it. More entries will be added and the existing entries will be added to. The standard should be set now that for more information the glossary links elsewhere. It's better to create very short pages that explain these things (and can be expanded later) than to add overly long glossary entries now.

Yes, this is fair. I can try trimming things down. I've already tried to avoid adding any doctests to the glossary. Some definitions are very hard to describe without a small example, but even there, I've kept the description inline instead of showing code. 

On the other hand, if someone does want to add just a short definition of something, this does give a place for it to go, whereas now it would either have to be expanded to a full fledged guide or else shoehorned into some docstring somewhere.

doc/src/explanation/glossary.md
bertiewooster: While I think someone from the disability community would be better to answer that question, I appreciate that there are deeply embedded terms such as `walk` and that changing them all would be difficult for maintainers, and confusing for users. So yes, I think using both terms is a good idea at present.

That Wikipedia link seems good and sufficient.

doc/src/explanation/glossary.md
bertiewooster: Understand that isn't an atom. However, using the atoms(Relational) method on a solved equation returns that as an "atom":

```
>>> from sympy import *
>>> from sympy.core.relational import Relational
>>> x = symbols('x')
>>> eq = solve([x >= 0.5, x <= 0.7])
>>> [atom for atom in eq.atoms(Relational)]
[0.5 <= x, x <= 0.7]
```

Apparently this is a quirk of using Relational as an argument of atoms() (and perhaps other arguments), because omitting an argument does atoms that meet the definition given:

```
>>> [atom for atom in eq.atoms()]
[x, 0.700000000000000, 0.500000000000000]
```

In the interest of keeping the glossary short, it may well be best to discuss such subtleties elsewhere.

asmeurer: Oh, I see. The `atoms` method is a bit confusing relating to the definition I wrote here, because if you pass it a class it just returns all the instances of that class inside of an expression, even if those instances aren't "atomic" expressions. I guess I can add this to the reasons I don't really like the term "atom". I'll need to think about how to clarify this. Maybe a note should just be added to the `atoms` docstring about this.

asmeurer:
I've gone through this and tried to trim down the entries as best as I can.

I'd like to get this merged by next week so that I can reference the built page in the CZI report, which is due August 1.

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    top-level {term}`subexpressions <subexpression>` used to create it. They
+    are the arguments to the class used to create the expression. The args of
+    any expression can be obtained by the `.args` attribute. For example,
+    `(1 + x*y).args` is `(1, x*y)`, because it equals `Add(1, x*y)`. The `args`
+    together with {term}`func` completely define an expression. It is always
+    possible to walk the {term}`expression tree` and extract any subexpression
+    of a SymPy expression by repeated use of `.args`. Every SymPy expression
+    can be rebuilt exactly with `func` and `args`, that is,
+    `expr.func(*expr.args) == expr` will always be true of any SymPy
+    expression `expr`. The args of an expression may be the empty tuple `()`,
```
oscarbenjamin, line 58: I think "every" and "will always be true" is a bit strong. This doesn't hold for expressions created with evaluate=False.

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    meaning the expression is an {term}`atom`.
+
+Assumptions
+
+    *Assumptions* are a set of predicates on a {term}`symbol` or
```
oscarbenjamin, line 63: There is a long guide that could be linked to here:

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    Various methods on {term}`Basic` and {term}`Expr` can be defined on
+    subclasses using special *`_eval_*`* methods. For example, an object can
+    define how it will be processed by the {func}`~.diff` function by defining
+    a `_eval_derivative` method. `_eval_*` methods used are instead of
+    overriding the method itself so that the method defined on the base class
+    can do pre-processing before dispatching to the `_eval_*` method.
+
+`evalf`
+
+    [*`evalf`*](sympy.core.evalf.EvalfMixin.evalf) is the method present on
```
oscarbenjamin, line 193: I think it's worth mentioning `.n()` and `N()` here. It's not obvious that these are all equivalent.

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    expressions do not make mathematical sense in classes like {class}`~.Add`
+    or {class}`~.Mul`.
+
+Expression
+
+    Any SymPy object, that is, any instance of {term}`Basic`, may be called an
+    *expression*. Sometimes, the term "expression" is reserved for
+    {term}`Expr` objects, which are algebraic expressions. Expressions are not
+    to be confused with {term}`equations <equation>`, which are a specific
+    type of expression with that represents a mathematical equality.
```
oscarbenjamin, line 230: Change "with that". Also the pluralisation is confused.

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    strictly by {term}`structural equality`, one may also think of an
+    expression tree as being a
+    [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph), where
+    identical subexpressions are only represented in the graph once.
+
+`func`
+
+    The *`func`* property is the function of an {term}`expression`, which can
+    be obtained by `expr.func`. This is usually the same as `type(expr)`, but
+    may differ in some cases, so it should be preferred to `type(expr)` when
```
oscarbenjamin, line 255: to use

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    rebuilding expressions with {term}`args`. Every SymPy expression can be
+    rebuilt exactly with `func` and `args`, that is, `expr.func(*expr.args) ==
+    expr` will always be true of any SymPy expression `expr`.
```
oscarbenjamin, line 258: Again "every" seems too strong

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    Python functions which always return a value. For example, most
+    simplification functions like {term}`simplify() <simplification>` cannot
+    be represented symbolically.
+
+
+Immutable
+
+    In Python, objects are *immutable* if they can not be modified in-place.
+    In order to change an immutable object, a new object must be created. In
+    SymPy, all objects are immutable. This means that all functions that
```
oscarbenjamin, line 308: Matrix is not immutable

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    when using SymPy interactively, but is generally frowned upon for
+    programmatic usage, where importing names explicitly just using `import
+    sympy` is preferred.
+
+`is_*`
+
+    Attributes in SymPy that start with *`is_`* and use a *lowercase* name
+    query the given {term}`assumption <assumptions>` on that object (note:
+    there are a few properties that are an exception to this because they do
+    not use the assumptions system). For example, `x.is_integer` will query
```
oscarbenjamin, line 341: More explanation that could be linked to here:
https://docs.sympy.org/latest/guides/assumptions.html#other-is-properties

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    *Number* can refer to two things in SymPy:
+
+    - The class {class}`~.Number`, which is the base class for explicit
+      numbers ({class}`~.Integer`, {class}`~.Rational`, and {class}`~.Float`).
+      Symbolic numeric constants like {class}`pi <sympy.core.numbers.Pi>` are
+      not instances of `Number`.
+
+    - Lowercase "*number*", as in the `is_number` property, refers to any
+      {term}`expression` that can be {term}`evalfed <evalf>` into an explicit
+      `Number`. This includes symbolic constants like {class}`pi
```
oscarbenjamin, line 441: It's probably worth mentioning symbols here. Most of the time the distinction is about not having free symbols although it is also about distinguishing a number from something like a matrix.

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    wrapper around the polynomial factorization algorithms that are
+    implemented in the polys. The classes in the polys are implemented using
+    efficient data structures, and are not subclasses of {term}`Basic` like
+    the other classes in SymPy.
+
+Printing
+
+    *Printing* refers to the act of taking an {term}`expression` and
+    converting it into a form that can be viewed on screen. SymPy has several
+    printers which represent expressions using different formats. Some of the
```
oscarbenjamin, line 474: Printing is also often used to refer to code generation.

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    solving in this sense. It is not used in the sense of "solving a problem".
+    For instance, one would generally prefer to say "compute an integral" or
+    "evaluate an integral" rather than "solve an integral" to refer to
+    symbolic integration using the function {func}`~.integrate`.
+
+Structural Equality
+
+    Two SymPy objects are *structurally equal* if they are equal as
+    {term}`expressions <expression>`, that is, they have the same as
+    {term}`expression trees <expression tree>`. Two structurally equal
```
oscarbenjamin, line 548: Remove as

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    because the {class}`~.Add` constructor automatically sorts its arguments,
+    making them both the same.
+
+Subexpression
+
+    A *subexpression* is an {term}`expression` that is contained within a
+    larger expression. A subexpression appears somewhere in the
+    {term}`expression tree`. In some contexts, a subtree of the expression
+    tree is also considered a subexpression. For instance, `x + y` may
+    sometimes be considered a subexpression of `x + y + z`, even though the
```
oscarbenjamin, line 576: I've never seen the word "subtree" used in this way.

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+    <simplification>` to equivalent symbolic expressions using those
+    properties. The goal of SymPy is to represent and manipulate symbolic
+    expressions representing various mathematical objects.
+
+    Some sources use the phrases "analytic solution" or "closed-form" to refer
+    to the concept of "symbolic", but this terminology is not used in SymPy.
+    If used in SymPy, "analytic" would refer to the property of being [an
+    analytic function](https://en.wikipedia.org/wiki/Analytic_function), and
+    in SymPy {term}`solve` refers only to a certain type of symbolic
+    operation. "Closed-form" in SymPy would typically refer the mathematical
```
oscarbenjamin, line 614: to the

doc/src/explanation/glossary.md
```diff
@@ -0,0 +1,712 @@
...
+
+    *{class}`~.Symbol`* is the class for symbol objects. A symbol represents a
+    single mathematical variable in an expression. The {class}`~.Symbol` class
+    is a subclass of {term}`Expr` and is {term}`atomic <atom>`. A `Symbol`
+    contains a name, which is any string, and {term}`assumptions`. Symbols are
+    typically defined with the `Symbol` constructor or the {func}`~.symbols`
+    function. Two Symbols with the same name and assumptions are considered
+    {term}`equal <structural equality>`. Symbols are implicitly assumed to be
+    independent or constant with respect to one another. SymPy does not
+    distinguish between symbols that are constants, variables, or parameters.
```
oscarbenjamin, line 630: In the core no such distinction is made but when calling a function like solve these are distinguished.

oscarbenjamin:
> I'd like to get this merged

If you want to get this in then let's just get it in and improve later. The release branch is created now so there is plenty of time to improve things after they're merged.

I left a bunch of comments above. Some are simple typos but others are possibly harder. I think it's fine to merge though if you want to address those later.

doc/src/explanation/glossary.md
asmeurer: Maybe I should word this better. They're not intrinsically distinguished. A symbol being constant or not depends on how it is used (e.g., if it is passed as an argument to solve or not). It's a common misconception I see from new users that they need to specify that a symbol is constant or not when defining it.

doc/src/explanation/glossary.md
asmeurer: You mean the word "subtree" to refer to a subset of a (mathematical) tree? I would assume that's common, but perhaps I can find a reference (or more commonly used term).

Or do you mean you haven't seen `x + y` referred to as a subexpression of `x + y + z`? I see that a lot, especially in functions like `subs` that operate at a level higher than pure expression trees.

doc/src/explanation/glossary.md
asmeurer: Thanks I didn't notice we already had documentation for this.

asmeurer:
I'm not sure how much I want to bog down definitions with "except in this corner case because the design of SymPy isn't as clean as we'd like it to be" caveats. And one should consider that once the glossary is merged, it will be treated both as descriptive and prescriptive by people looking to add new things to the library. Maybe there are less extreme words I can use than "every" and "all" that don't imply strict compliance but at least indicate that this is how things ought to be.

doc/src/explanation/glossary.md
oscarbenjamin: At the bottom of this wikipedia page there is a footnote:

This is different from the formal definition of subtree used in graph theory, which is a subgraph that forms a tree – it need not include all descendants. For example, the root node by itself is a subtree in the graph theory sense, but not in the data structure sense (unless there are no descendants).
https://en.wikipedia.org/wiki/Tree_(data_structure)

I'm not sure that the word subtree as used here corresponds to either of those definitions though:

- The data structure subtree would not allow `x + y` to be a subtree of `x + y + z`.
- The graph theory subtree would allow `x + y*z` to be a subtree of `x + y*z*t`.

oscarbenjamin:
We can use the glossary to define terminology that makes it easier to express exceptional cases accurately. For example:

Evaluation is **idempotent**. For every **evaluated expression** `expr` it is always true that `expr.func(*expr.args) == expr`.

An **unevaluated expression** is created with `evaluate=False`. For every **expression** (evaluated or unevaluated) `expr.func(*expr.args, evaluate=False) == expr`.

Of course there will still be exceptions to those in the codebase but those can truly be considered bugs.

doc/src/explanation/glossary.md
asmeurer: Really this only applies to the associative operators Add and Mul, where you can take subsets of them and consider that to be a "subexpression". You wouldn't consider `f(x, y)` to be a subexpression of `f(x, y, z)` for any other `f`. So maybe instead of trying to describe this in terms of graph theory, it can just mention that sometimes commutative or associative laws are taken into account for deciding what is a subexpression, which can expand past the literal "subtree" definition.

We really should have more formal ways of thinking about this sort of thing in SymPy, along the lines of MatchPy's commutative/associative pattern matching (and I think e.g., Mathematica has similar notions). But presently it's all treated in an ad hoc manner.

oscarbenjamin: The examples I showed use only associative, commutative operations. I think what you are describing is like the data structure subtree except that only a subset of children needs to be taken at the second level of the tree (the level below the outer Add).

doc/src/explanation/glossary.md
asmeurer: I'm really not clear what you think I should add here. Symbols are not numbers, according to this definition (they don't represent a specific numerical value).

asmeurer:
We should probably discuss the evaluate=False stuff separately. I disagree that we should document the invariant as `expr.func(*expr.args, evaluate=False) == expr`. To me, `expr.func(*expr.args)` makes sense when `expr` is well-formed, which happens when it's been properly canonicalized with its constructor. I see `evaluate=False` as a bit of a hack, because it disables that canonicalization. Some things might still happen to work, but some things won't, and rebuilding is one of the things that won't. 

If write "For every expression (evaluated or unevaluated) `expr.func(*expr.args, evaluate=False) == expr`." in the glossary, that's implying that functions that rebuild expressions should use `evaluate=False`. But that's not actually how we want to be handling unevaluated expressions. If every single function has to think about both evaluated and unevaluated expressions it gets too messy. It's already a mess for the handful of functions that do (the most common being the printers, and even there are there are still tons of bugs with it).

In fact, I would say it's misleading to even say that there's a distinction between "evaluated" and "unevaluated" expressions. There's well-formed expressions, which have gone through the expected canonicalizations in their constructors, and expressions that have skipped those canonicalizations, and which may or may not be canonicalized. For example, there's no material difference between `Add(x, y)` and `Add(x, y, evaluate=False)` (but there *is* between `Add(y, x) and `Add(y, x, evaluate=False)`).

I'd rather treat `evaluate=False` as what it is, a hack around the fact that some constructors do too much automatic simplification. I do mention `evaluate=False` and automatic simplification in several places in the glossary, but for most definitions, I think they should just apply to well-formed expressions, i.e., ones that actually go through the proper canonicalization. Because the reality is that most functions actually do expect well-formed expressions, and might fail or give wrong answers if given one that isn't.

`expr.func(*args) == expr` is really saying that expressions should be assumed to be canonicalized by their constructors. It's much like saying all `args` should be `Basic`. In a case where it's not actually true, it's really a bug, because it leads to real issues, since potentially any function will assume that. And functions *should* be able to assume that.

asmeurer:
Anyway, I need to merge this now, because I need a URL for the report as I noted, but feel free to make any more comments here and I will address them in another PR.

oscarbenjamin:
> I disagree that we should document the invariant as `expr.func(*expr.args, evaluate=False) == expr`

Perhaps it wasn't clear but what I meant was that we could write that in the section that defines what an **unevaluated expression** is. My suggestion for the func-args variant section would be to explicitly refer to **evaluated expressions** but otherwise just document the func-args invariant as you have done since it *should* hold for **evaluated expressions** which are the common case. Currently the glossary adds an entry for **Unevaluated** but no entry for **Evaluated**. If there was an entry for **evaluated expression** then it could be used to define the func-args invariant accurately.

> I'd rather treat `evaluate=False` as what it is, a hack around the fact that some constructors do too much automatic simplification

This is not what most users use `evaluate=False` for. It is a really common case to want to be able to just represent an expression as-is with no evaluation or canonicalisation. SymPy doesn't support that case very well at all but that's a deficiency of SymPy. Note that the glossary entry for **Unevaluated** here explicitly states that "unevaluated expressions are supported".

PR #23298, opened 2022-03-28
add type hints to quadratic_residues

eagleoflqj:
Originally pyright deduces the return type as `list[Unknown]`.
Converting `set` to `list` before sorting is not necessary.
Remove a duplicated test case.

#### Release Notes

NO ENTRY


oscarbenjamin:
Looks good

PR #23311, opened 2022-04-02
Proof of concept: Make Function classes behave as Basic instances

ThePauliPrinciple:

https://github.com/sympy/sympy/pull/22445

This is a proof of concept to demonstrate that it is possible to
make `Function` classes behave as if they are `Basic` instances,
without interfering with instances of those `Function` classes.
I'm currently looking for counter examples where this fails.

Some ideas to think about if something along these lines were to
be added to SymPy:

Should `isinstance(sin, Basic) == True`? It's possible to do this
technically, but it would potentially alter many code paths.

Should the information of "specific" functions (e.g. `sin`) be
stored in the `FunctionSymbol`s? This would allow a move towards
something along the lines of `Function(FunctionSymbol(), x, y, ...)`
by seperating changes in the library and the actual deprecation
process. Perhaps even allowing the current function system as well
as a new one to work at the same time.

#### Release Notes

NO ENTRY


oscarbenjamin:
Can you explain how the mechanism here works?

This seems to me like it is working around the problem rather than fixing it but perhaps it makes behaviour that would be more compatible with the ideal future behaviour.

Ultimately I think that the problem is the design that expression heads are Python classes and expressions are instances of those classes. I have come to think now that a head should just be an expression and that it should just be the first expression in the args list of any expression that it generates so e.g. `sin(x)` should just be represented as `Expression((sin, x))` where both `sin` and `x` are themselves expressions.

This is how it works in e.g. Mathematica:
```mathematica
In[10]:= e = Sin[x^2]
Out[10]:= Sin[x ^ 2]

In[11]:= e
Out[11]:= Sin[x ^ 2]

In[12]:= Head[e]
Out[12]:= Sin

In[13]:= e[[0]]
Out[13]:= Sin

In[14]:= e[[1]]
Out[14]:= x ^ 2
```
In fact the head does not even need to be an atom which makes it possible to represent derivatives of functions without using any symbols:
```mathematica
In[15]:= e2 = D[f[x], x]
Out[15]:= f'[x]

In[16]:= Head[e2]
Out[16]:= f'

In[17]:= Head[e2] // FullForm
Out[17]:= Derivative[1][f]

In[18]:= e3 = f'[0]
Out[18]:= f'[0]

In[19]:= e3 // FullForm
Out[19]:= Derivative[1][f][0]
```
In Mathematica `D[e,x]` means derivative wrt `x` but `Derivative[1][f]` just means the derivative of the function `f` regardless of any symbols.

I think that Maple is similar:
https://www.maplesoft.com/support/help/maple/view.aspx?path=op

ThePauliPrinciple:
Yes, this is definately a work around and I absolutely agree with your analysis on the ideal solution.

There are two major obstacles when we want to change SymPy to make functions work like that: rewriting large parts of the library and some very inconvenient backwards incompetability.

I hope that this kind of python "trickery" can allow us to rewrite the library a small piece at a time without introducing backwards incompetability until the last moment.

As an example, if we want `isinstance(sin, Basic) == True`, a lot of code paths would currently break. It is possible to make this currently evaluate to `True` by modifying `__instancecheck__` in `BasicMeta`  by checking e.g. for the attribute `function_symbol` (or perhaps even `.__sympy__ is True`). This means that we can change the code paths, without changing the interface to `Function.`

Ideally, all code can be replaced before we make the change to a new system and we can have a deprecation period where the old way of using `Function` works together with a/the new way.

As for a less distant future, this can help with the current problem that `Function` classes are being used inside SymPy as if they are `Basic` at least in one non-trivial case.

As for the mechanism: when you access an attribute of a class, the `__getattribute__` method of its meta class is used to obtain it. Here I override `__getattribute__` in `FunctionClass` (which is `Function`'s meta class) to look up attributes in a `FunctionSymbol` instance when that attribute name is in a whitelist. So `sin.args` actually returns `FunctionSymbol(sin).args` (which is an empty `tuple` here). A `FunctionSymbol` object is added to the `Function` class objects when they are created by the meta class.

oscarbenjamin:
> As an example, if we want `isinstance(sin, Basic) == True`, a lot of code paths would currently break.

The bigger issue with changing to something like I suggested is changing `isinstance(sin(x), sin)`.

> As for the mechanism:

Can you explain this more clearly? What is the type of `f = Function('f')`? What is the type of `f(x)`? How do these types relate?

We can mess around with `__instance_check__` but how does that affect something like `type(f(x)) is f`?

ThePauliPrinciple:
> The bigger issue with changing to something like I suggested is changing `isinstance(sin(x), sin)`.

You can just do this by defining `__instancecheck__` in the class, rather than the metaclass. (Python makes nearly no exceptions, classes are just objects, meta classes are just classes are just objects):

```python
class A:
    pass

class B:
    def __instancecheck__(self, inst):
        return type(inst) is A

isinstance(A(), B())
```

> Can you explain this more clearly? What is the type of `f = Function('f')`? What is the type of `f(x)`? How do these types relate?

Nothing changed here w.r.t. to how the library currently works. Just the attribute lookup when accessing them from the class object directly changed.
```python
>>> from sympy import Function; from sympy.abc import x
>>> type(Function('f')(x)).__mro__
(f, AppliedUndef, Function, Application, <class 'sympy.core.expr.Expr'>, <class 'sympy.core.basic.Basic'>, <class 'sympy.printing.defaults.Printable'>, <class 'sympy.core.evalf.EvalfMixin'>, <class 'object'>)
>>> type(Function('f')).__mro__
(<class 'sympy.core.function.UndefinedFunction'>, <class 'sympy.core.function.FunctionClass'>, <class 'sympy.core.assumptions.ManagedProperties'>, <class 'sympy.core.core.BasicMeta'>, <class 'type'>, <class 'object'>)
```

> We can mess around with `__instance_check__` but how does that affect something like `type(f(x)) is f`?

I don't think we can in any way influence what `type` and `is` do in python, one of the few exceptions I suppose. (type doesn't even look up what is in `__class__`, which is something that we can change). With the changes in this PR, the result of this is not yet influenced (since the type of any function is still the same), but this is something that needs to be changed inside SymPy if we want to change how functions work along the lines of your suggestions. Hopefully it is easier to perform this replacement if `f` can already behave like a `basic`. Is there any case where `is` is absolutely required? `==` behaviour for the class can be modified in the metaclass.

ThePauliPrinciple:
> The bigger issue with changing to something like I suggested is changing `isinstance(sin(x), sin)`.

Also, the problem is not so much to make this equal to true, it is that `if isinstance(x, Basic): ` branches would start to trigger for e.g. `cos`, which are currently usually trig

ThePauliPrinciple:
> Also, the problem is not so much to make this equal to true, it is that `if isinstance(x, Basic): ` branches would start to trigger for e.g. `cos`, which are currently usually triggering a different branch e.g. `isinstance(x, Function)`.
> 
> EDIT: clicked wrong button and closed it

Actually, this problem doesn't seem too bad either, I just had to extend the whitelist it seems. It does now need a seperate exclusion for pickling (since pickling of classes and instances works differently)

ThePauliPrinciple:
Currently there are two points that I'm a bit stuck on:

I would like for `FunctionSymbol` to be an actual `Symbol`, however, `Symbol` inherits from `Boolean`, which is outside of core and imports from core/function.py. Since the `FunctionSymbol` is created when the class is defined, I can't delay the import.

I'm not clear what would be the prefered interface for defining specific functions (e.g. `sin`, `cos`). Currently they are classes, but that's not what we want. Then what should it look like if a user (or the library) wants to define a function with specific properties?

ThePauliPrinciple:
Actually, the problem with `Boolean` using `Function`s has exposed another issue to me. Operators like `Or` and `And` are actually functions in SymPy. Do we then also want `Or(A, B)` to be represented as `Call(Or, A, B)`? If not, then what kind of functions do we want and what kind of functions don't we want to behave in the new way?

oscarbenjamin:
I don't see why `Or` needs to subclass `Function` (or rather `Application`).

ThePauliPrinciple:
It currently does. And it isn't exactly "wrong" to consider it a "function" (or "application"). `Application` seems to be what allows for substituting the `class` rather than the `args` (see its `_eval_subs`).

So by what "rule" do we decide which classes get changed in their behaviour and more importently why?

oscarbenjamin:
I would expect And and Or to behave the same as Mul and Add. 

I'm not really sure how to specify in general what should happen. The more I think about it I think that using classes for different expression nodes is a bad idea in general.

ThePauliPrinciple:
Unless it can be properly defined what is and what is not a function that should work according to a new "style", I don't think it is a good idea to work towards such a change. 

On the other hand, I do think I have demonstrated that it is feasible to have a class that behaves similar to a `Basic` using `__getattribute__`.

Alternatively, this can also be achieved with a "getter":

```python
>>> class property_getter_a():
...     def __get__(self, instance, owner):
...         if instance is None:
...             print('do stuff for class')
...         else:
...             print('do stuff for instance')
...         return 0
>>> class A():
...     a = property_getter_a()
>>> A.a
do stuff for class
0
>>> A().a
do stuff for instance
0
```

The challenge is still to find all properties/methods that need to be handled in this way though.

ThePauliPrinciple:
Tried to take a couple of steps back. There is now a `FunctionSymbol(Symbol)` class. On `_sympify`, `Function` classes are replaced with this new class. It stores the `Function` as an instance (I think this is needed for pickling?). Currently it still fails some matches (mostly because I don't really understand how it works).

PR #23316, opened 2022-04-04
Fix `QRdecomposition` for symbolic cases

### LINKED ISSUE #23313: QR factorization fails if matrix has Symbols
Make a matrix with a Symbol in it.  Simplest case:
```
>>> var('x')
>>> A = Matrix([x])
```
Takes its QR decomposition:
```
>>> Q, R = A.QRdecomposition()
>>> print(Q)
Matrix(1, 0, [])
>>> print(R)
Matrix(0, 1, [])
```
Similarly for bigger matrix:
```
>>> B = Matrix([[x, 0],[0,x]])
>>> Q, R = B.QRdecomposition()
>>> print(Q)
Matrix(2, 0, [])
>>> print(R)
Matrix(0, 2, [])
```
Tested on 1.8, 1.9. 1.10.1 and master.

In SymPy <= 1.7.1 this worked:
```
>>> var('x')
>>> A = Matrix([x])
>>> Q, R = A.QRdecomposition()

>>> pprint(Q)
⎡ x ⎤
⎢───⎥
⎣│x│⎦
>>> pprint(R)
[│x│]

>>> B = Matrix([[x, 0],[0,x]])
Q, R = B.QRdecomposition()
>>> pprint(Q)
⎡ x      ⎤
⎢───   0 ⎥
⎢│x│     ⎥
⎢        ⎥
⎢      x ⎥
⎢ 0   ───⎥
⎣     │x│⎦
>>> pprint(R)
⎡│x│   0 ⎤
⎢        ⎥
⎣ 0   │x│⎦
```
I have not yet bisected this.

sylee957:

Fixes https://github.com/sympy/sympy/issues/23313

I'm attempting to fix the issue by changing the fuzzy logic to infer `is_zero == None` as nonzero.

#### Release Notes

- matrices
  - Fixed `Matrix.QRdecomposition` giving wrong results for matrices containing symbols (for most cases where zero test can automatically infer the correct results)

### DIFF
--- a/sympy/matrices/decompositions.py
+++ b/sympy/matrices/decompositions.py
@@ -1363,7 +1363,7 @@ def _QRdecomposition_optional(M, normalize=True):
             Q[:, j] -= Q[:, i] * R[i, j]
 
         Q[:, j] = dps(Q[:, j])
-        if Q[:, j].is_zero_matrix is False:
+        if Q[:, j].is_zero_matrix is not True:
             ranked.append(j)
             R[j, j] = M.one
 
--- a/sympy/matrices/tests/test_decompositions.py
+++ b/sympy/matrices/tests/test_decompositions.py
@@ -1,7 +1,9 @@
 from sympy.core.function import expand_mul
-from sympy.core.numbers import (I, Rational)
+from sympy.core.numbers import I, Rational
 from sympy.core.singleton import S
+from sympy.core.symbol import Symbol
 from sympy.functions.elementary.miscellaneous import sqrt
+from sympy.functions.elementary.complexes import Abs
 from sympy.simplify.simplify import simplify
 from sympy.matrices.matrices import NonSquareMatrixError
 from sympy.matrices import Matrix, zeros, eye, SparseMatrix
@@ -152,6 +154,18 @@ def test_QR():
     assert Q.T * Q == eye(Q.cols)
     assert R.is_upper
     assert A == Q*R
+
+    x = Symbol('x')
+    A = Matrix([x])
+    Q, R = A.QRdecomposition()
+    assert Q == Matrix([x / Abs(x)])
+    assert R == Matrix([Abs(x)])
+
+    A = Matrix([[x, 0], [0, x]])
+    Q, R = A.QRdecomposition()
+    assert Q == x / Abs(x) * Matrix([[1, 0], [0, 1]])
+    assert R == Abs(x) * Matrix([[1, 0], [0, 1]])
+
 
 def test_QR_non_square():
     # Narrow (cols < rows) matrices


sympy/matrices/tests/test_decompositions.py
```diff
@@ -153,6 +155,18 @@ def test_QR():
     assert R.is_upper
     assert A == Q*R
 
+    x = Symbol('x')
```
cbm755, line 158: probably should be `x = Symbol('x', nonzero=True)`?  Otherwise, the `Q` matrix is not well-defined.

sympy/matrices/tests/test_decompositions.py
```diff
@@ -153,6 +155,18 @@ def test_QR():
...
+    A = Matrix([x])
+    Q, R = A.QRdecomposition()
+    assert Q == Matrix([x / Abs(x)])
+    assert R == Matrix([Abs(x)])
+
+    A = Matrix([[x, 0], [0, x]])
+    Q, R = A.QRdecomposition()
+    assert Q == x / Abs(x) * Matrix([[1, 0], [0, 1]])
+    assert R == Abs(x) * Matrix([[1, 0], [0, 1]])
```
cbm755, line 167: I'd probably use a new `def test_QR_with_symbols():` but don't feel strongly.

sympy/matrices/tests/test_decompositions.py
cbm755: ...and would the original code have worked in that case?  That suggests that a better fix may be possible.

cbm755: It does seem suspicious to be able to take a path through this GS/MGS and get empty (degenerate) matrices.  

Here's another case to worry about:
```
Matrix([[0]]).QRdecomposition()
Out[42]: (Matrix(1, 0, []), Matrix(0, 1, []))
```
(That should give `Q=[[1]]` and `R=[[0]]`.)

cbm755: Oh I see, this is a "thin" rank-revealing QR factorization.

```
>>> q, r = Matrix([[0]]).QRdecomposition()
>>>: q
Matrix(2, 0, [])
>>> r
Matrix(0, 2, [])

>>> q*r
Matrix([[0]])
```
I also tested with 2x2 zero matrix and it also works!  That is pretty neat!  At least some of this seems to be unit-tested too.

sympy/matrices/decompositions.py
```diff
@@ -1363,7 +1363,7 @@ def dot(u, v):
             Q[:, j] -= Q[:, i] * R[i, j]
 
         Q[:, j] = dps(Q[:, j])
-        if Q[:, j].is_zero_matrix is False:
+        if Q[:, j].is_zero_matrix is not True:
```
cbm755, line 1366: Ok, having experimented with the below, I agree with this change.

But perhaps it could use a comment:
```
  # if we are sure the column is zero, we do not need to include it
  if Q[:, j].is_zero_matrix in (False, None):
```
Or how about this without a comment?
```
  if Q[:, j].is_zero_matrix:
       continue
  ranked.append(j)
  R[j, j] = M.one
```

sympy/matrices/tests/test_decompositions.py
sylee957: I've attempted to use the preceding boilerplates for tests, but sympy is incapable of proving 'Q * Q.H == I' because 'x * conjugate(x) / Abs(x)**2' is not simplified even with 'simplify'

oscarbenjamin:
Looks good to me.

PR #23320, opened 2022-04-05
add type hints to continued_fraction[_periodic]

eagleoflqj:
`continued_fraction` and `continued_fraction_periodic` are mutually recursive, so pyright fails to deduce the return type of both.
After adding return type, mypy complains the type of `terms`. After that, the `terms[:i] + [terms[i:]]` must be ignored due to https://github.com/python/typeshed/issues/2383

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/ntheory/continued_fraction.py
+++ b/sympy/ntheory/continued_fraction.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 from sympy.core.exprtools import factor_terms
 from sympy.core.numbers import Integer, Rational
 from sympy.core.singleton import S
@@ -6,7 +7,7 @@ from sympy.core.sympify import _sympify
 from sympy.utilities.misc import as_int
 
 
-def continued_fraction(a):
+def continued_fraction(a) -> list:
     """Return the continued fraction representation of a Rational or
     quadratic irrational.
 
@@ -71,7 +72,7 @@ def continued_fraction(a):
         'expecting a rational or quadratic irrational, not %s' % e)
 
 
-def continued_fraction_periodic(p, q, d=0, s=1):
+def continued_fraction_periodic(p, q, d=0, s=1) -> list:
     r"""
     Find the periodic continued fraction expansion of a quadratic irrational.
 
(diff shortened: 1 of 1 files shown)


sympy/ntheory/continued_fraction.py
```diff
@@ -177,7 +178,7 @@ def continued_fraction_periodic(p, q, d=0, s=1):
         q = (d - p**2)//q
 
     i = pq[(p, q)]
-    return terms[:i] + [terms[i:]]
+    return terms[:i] + [terms[i:]]  # type: ignore
```
oscarbenjamin, line 181: I'm not sure I understand the benefit of adding type hints if it always ends up needing to have type ignore somewhere. What is the actual type of the return value here?

oscarbenjamin: My question is really what value the hints give if they can't actually express the types involved. Here the return type is just `list`. Does that really help anything?

I don't understand why out of all of the possible places that a hint could be added you have picked this one obscure function that can't actually be typed properly anyway.

eagleoflqj: Because I'm adding more functionalities to https://github.com/eagleoflqj/sympy_beta. Wherever I can't get type hints from pyright extension for vscode, I have to check the sympy source to make sure I handle return value correctly, then I try to add type hints. Last time was `quadratic_residue` and it's online https://sympy-beta.vercel.app/input/12. I'm just doing continued fraction these days.
Of course there are places that are more important than here to be typed. I tried several but they either require python 3.8+ for Literal, or can't make mypy happy without hacking, or are too complex to satisfy other sympy members. They are all lying silently on my local repo.

eagleoflqj: > Here the return type is just `list`. Does that really help anything?

Previously even `list` can't be deduced. I just think it better than nothing. SymPy (or any open source project) is never a "perfect or nothing" thing. I believe even small improvement is worthy.

oscarbenjamin: I think it's important to try and feed these things back up to the type checkers that they should improve things like this. Here it should be straight-forward to infer that the return type is a list.

eagleoflqj: Yes. It's a duplicate of https://github.com/microsoft/pyright/issues/3166 that won't be fixed by pyright.

PR #23324, opened 2022-04-06
Inclusion of identification of dimensionless quantities using Buckingham pi theorem

ShrohanMohapatra:

Added the functionality to be able to identify a set of dimensionless numbers from a given set of characteristic physical quantities by the use of the prescriptions from the Buckingham pi theorem. The function by the name 'getDimensionlessNumbers' added to the class DimensionSystem as a public property invocable on its own with the definition of some globally defined base dimensions.

#### Other comments

First ever PR for the author

#### Release Notes

NO ENTRY


sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,74 @@ def is_consistent(self):
...
+            'amount_of_substance': 5,
+            'luminous_intensity': 6
+            }
+        """
+            We extract the exponent matrix from the dimensional dependencies of the
+            quantities
+        """
+        for quantity in list_of_derived_Quantities:
+            ElementForQuantity = [0 for k in range(7)]
+            dimensions = dimsys_SI.get_dimensional_dependencies(
```
Upabjojr, line 607: You are using a reference to `dimsys_SI`, which is an instance of `Dimensions`, inside the class `Dimension`. It makes the algorithm dependent on the dimension system of SI units only. The units module should be independent of the chosen unit/dimension system.

What about `self.get_dimensional_dependencies` instead?

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,74 @@ def is_consistent(self):
...
+    	essentially dimensionful) and returns one set of all possible dimensionless
+	    quantities, as guided by the Buckingham's pi theorem (a generalization of 
+	    Rayleigh's method of dimensional analysis .....)
+	    
+        Please refer and run to the unit testing file '/units/test/test_dimensionLess.py' for
+            the format of the input .... As a heads up, we first need to define
+            a global set of base dimensions and with those we can define the 
+            derived quantities .....
+        """
+        ExponentMatrix = []
```
Upabjojr, line 591: In SymPy, we are trying to use *CamelCase* style for class names, we prefer *snake_case* for variable names.

It's not mandatory, but having a unified style makes the code more readable.

sympy/physics/units/dimensions.py
ShrohanMohapatra: @Upabjojr thanks much Francesco. So may I include it as an edit here or would it be a separate commit somewhere?

sympy/physics/units/dimensions.py
ShrohanMohapatra: @Upabjojr correct, makes sense! I'll correct the variables in the future edits.

ShrohanMohapatra (commented):
@Upabjojr Thanks for the suggestions ....

sympy/physics/units/dimensions.py
Upabjojr: What do you mean by _edit_? An edit will create a new commit.

sympy/physics/units/tests/test_dimensionLess.py
```diff
@@ -0,0 +1,117 @@
+# Testing the function that gets a set of non-dimensional numbers
```
smichr, line 1: please make this like the other test files in the repo

ShrohanMohapatra: Sure, right away! :)

Upabjojr:
SymPy does requires the code to have these conditions:

1. no line can end with whitespaces (tests are failing because of this),
2. exactly one newline at the end of the file (after the last expression),
3. no unused imports.

Condition 1. is not satisfied right now.

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,64 @@ def is_consistent(self):
...
+        This function is designed to take in a set of physical variables (all 
+    	essentially dimensionful) and returns one set of all possible dimensionless 
+	    quantities, as guided by the Buckingham's pi theorem (a generalization of 
+	    Rayleigh's method of dimensional analysis .....)
+        Please refer and run to the unit testing file '/units/test/test_dimensionLess.py' for
+        the format of the input .... As a heads up, we first need to define a global set of 
+        base dimensions and with those we can define the derived quantities .....
+        """
+        exponent_matrix = []
+        index_dict = {
```
Upabjojr, line 590: This dictionary is not necessary, we already have `self.base_dims`. It's a list, but you could use `self.base_dims.index(dimension)` to make it work the same way as `index_dict`.

ShrohanMohapatra: Thanks for this, but does self.base_dims have a property/attribute by the name index? I looked up the complete file but could not find any such attribute so thought to ask you before pushing in a new commit.

ShrohanMohapatra: Hi @Upabjojr Francesco, I tried removing the redundant whitespaces from the end of each line, and replaced the index_dict with the self.base_dims, and also edited the release notes. But still it shows the two PR tests not working so .....

Upabjojr: By looking at the logs:

> AssertionError: File contains trailing whitespace: /home/runner/work/sympy/sympy/sympy/physics/units/dimensions.py, line 580.

So, line 580 of file `dimensions.py`.

You can run the code quality tests on your computer by calling `sympy.test("sympy/testing")`

ShrohanMohapatra: Thanks @Upabjojr Francesco, everything else seems to work, the only two errors that remain are the release notes error and the following on the authors' inclusion:

 * Run bin/mailmap_check.py
  bin/mailmap_check.py
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.9.1[2](https://github.com/sympy/sympy/runs/5983251488?check_suite_focus=true#step:6:2)/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/[3](https://github.com/sympy/sympy/runs/5983251488?check_suite_focus=true#step:6:3).9.12/x6[4](https://github.com/sympy/sympy/runs/5983251488?check_suite_focus=true#step:6:4)/lib
Shrohan Mohapatra <4[6](https://github.com/sympy/sympy/runs/5983251488?check_suite_focus=true#step:6:6)8182[7](https://github.com/sympy/sympy/runs/5983251488?check_suite_focus=true#step:6:7)9+ShrohanMohapatra@users.noreply.github.com>

It'd be great if you could suggest something .... Thanks a lot once again :)

Upabjojr: I wouldn't be too concerned with the mailmap file, once a new SymPy version is released, the mailmap file will be automatically adjusted by taking the author and email list from the git commits. So just make sure your email is correctly inserted into your git commits.

ShrohanMohapatra: Thanks @Upabjojr. I think the write access stays with the mentors, so if everything else works (I think yes my email address is there and correct in the commits), may I be able to request you to merge the 22 commits in the main sympy master branch? I would like to continue to contribute to sympy this summer so ....

Upabjojr: If the tests don't pass, there is no way to merge this. Merging is only possible after all tests have passed.

There are still some issues with the code that I see right now. I will write later.

ShrohanMohapatra: Okay sure! Awaiting your response ....

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,44 @@ def is_consistent(self):
         # dimensions
         # in vector language: the set of vectors do not form a basis
         return self.inv_can_transf_matrix.is_square
+    def get_dimensionless_numbers(self, list_of_derived_quantities):
```
Upabjojr, line 575: one newline before `def` statements makes the code more readable.

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,44 @@ def is_consistent(self):
...
+        those we can define the derived quantities ..... """
+        exponent_matrix = []
+        """ We extract the exponent matrix from the dimensional dependencies of the quantities """
+        for quantity in list_of_derived_quantities:
+            element_for_quantity = [0 for k in range(7)]
+            dimensions = self.get_dimensional_dependencies(list_of_derived_quantities[quantity])
+            for base_dims in dimensions:
+                element_for_quantity[self.base_dims.index(base_dims)] = dimensions[base_dims]
+            exponent_matrix.append(element_for_quantity)
+        """ Next we obtain the null space of the exponent matrix """
```
Upabjojr, line 590: I would suggest to use comments `# ` for comments inside the code. Docstrings `"""  ...  """` should just appear once after the `def ...` statement.

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,44 @@ def is_consistent(self):
...
+        """ list_of_derived_quantities is the input list of derived quantities in the form of a python dictionary of tuples (key, item) where key is the name of
+        the physical quantity and item is an instance of Dimension class which possesses the dimensions of the quantity. This function is designed to take in
+        a set of physical variables (all essentially dimensionful) and returns one set of all possible dimensionless quantities, as guided by the
+        Buckingham's pi theorem (a generalization of Rayleigh's method of dimensional analysis .....) Please refer and run to the unit testing file
+        /units/test/test_dimensionless.py for the format of the input .... As a heads up, we first need to define a global set of base dimensions and with
+        those we can define the derived quantities ..... """
+        exponent_matrix = []
+        """ We extract the exponent matrix from the dimensional dependencies of the quantities """
+        for quantity in list_of_derived_quantities:
+            element_for_quantity = [0 for k in range(7)]
```
Upabjojr, line 585: why is there `range(7)`? The number of dimensions depends on the dimension system that is being used.

sympy/physics/units/tests/test_dimensionless.py
```diff
@@ -0,0 +1,60 @@
...
+    p_c = M/L/T**2 # Characteristic pressure scale
+    F_bc = M*L/T**2 # Characteristic body forces scale
+    list_of_quantities = {'Lc': Lc, 'Uc' : Uc, 'rho' : rho_c, 'mu' : mu, 'pc' : p_c, 'Fbc' : F_bc}
+    set_of_dimless_quant, set_of_dimless_nums = dummy_system.get_dimensionless_numbers(list_of_quantities)
+    flag = True
+    for dimension in set_of_dimless_nums:
+        flag = flag and dimension == Dimension(1)
+    assert flag
+    """ You should be able to see the Reynolds number, Euler's number and Froude number in the output """
+def test_mhd_case():
```
Upabjojr, line 28: what about having tests with examples of electromagnetism with both SI and cgs or gaussian systems? _cgs_ and _gaussian_ dimension systems are good tests because they have one less base dimension than _SI_ (**current** is a base dimension in _SI_ but it is a derived dimension in _cgs_ and _gauss_).

ShrohanMohapatra: Yeah thanks for the ideas @Upabjojr Francesco! Much appreciated! Still would like to have your reviews, actually I wanted to join you guys in sympy this summer so a quicker feedback would help me work faster on the problem before the application deadline so .... And again I think the only problem lies in the .mailmap issue so if you could help ....

Upabjojr: No problem for GSoC. What project are you applying to?

Try not to edit the mailmap file. I don't know what the problem is.

Can you think of an application of Buckingham's Pi theorem in cgs or gaussian dimensions where the result is different than SI? This would test the stability of the code and make sure it works in dimensiom systems other than SI.

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,49 @@ def is_consistent(self):
         # dimensions
         # in vector language: the set of vectors do not form a basis
         return self.inv_can_transf_matrix.is_square
+    def get_dimensionless_numbers(self, list_of_derived_quantities):
+        """ list_of_derived_quantities is the input list of derived quantities in the form of a python dictionary of tuples (key, item) where key is the name of
+        the physical quantity and item is an instance of Dimension class which possesses the dimensions of the quantity. This function is designed to take in
+        a set of physical variables (all essentially dimensionful) and returns one set of all possible dimensionless quantities, as guided by the
+        Buckingham's pi theorem (a generalization of Rayleigh's method of dimensional analysis .....) Please refer and run to the unit testing file
+        /units/test/test_dimensionless.py for the format of the input .... As a heads up, we first need to define a global set of base dimensions and with
+        those we can define the derived quantities ..... """
```
Upabjojr, line 581: These docstring lines are too long.

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,49 @@ def is_consistent(self):
...
+        # We need to make sure that we obtain integer nullspaces because
+        # Dimension() is not able to simplify non-diophantine exponents
+        null_space_dims = len(exponents)
+        m = 1
+        while True:
+            flag = True
+            for k in range(null_space_dims):
+                for l in range(len(exponents[k])):
+                    exponents[k][l] = exponents[k][l] * m
+                    flag = flag and (exponents[k][l]-int(exponents[k][l])) == 0.0
```
Upabjojr, line 602: We don't use floating points in SymPy unless really necessary.

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,49 @@ def is_consistent(self):
...
+            if flag:
+                # Just to make sure that the exponents are all integers
+                for k in range(null_space_dims):
+                    for l in range(len(exponents[k])):
+                        exponents[k][l] = int(exponents[k][l])
```
Upabjojr, line 607: This flag block looks pretty useless, why not just store integers in the matrix at its creation?

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,49 @@ def is_consistent(self):
...
+                break
+            m = m + 1
+        set_of_dimless_nums = [Dimension(1) for k in range(null_space_dims)]
+        set_of_dimless_quant = ['' for k in range(null_space_dims)]
+        # Here as a final step we take the dot product of the nullspace with the derived quantities
+        for k in range(null_space_dims):
+            m = 0
+            for quantity in list_of_derived_quantities:
+                set_of_dimless_nums[k] = set_of_dimless_nums[k]*(list_of_derived_quantities[quantity]**exponents[k][m])
+                set_of_dimless_quant[k] = set_of_dimless_quant[k] + '[' + quantity + ']' + '**(' + str(exponents[k][m]) + ')*'
```
Upabjojr, line 617: Why is this string necessary? Strings should go into the printing module (usually).

sympy/physics/units/tests/test_dimensionless.py
ShrohanMohapatra: Thanks a lot for the suggestions, would be going over them in a short while. I have a proposal by the name "Automated reduction of differential Groebner bases to simplify systems of PDEs
" which aims to implement the Kolchin-Ritt algorithm (the analogue of Buchberger's algorithm) to simplify any given system of PDEs, a topic which I have been reading extensively on for a while as a pursuit of my own research interests.

I think I remember obtaining different dimensionless quantities in Gaussian especially for the MHD case and circuit QED case, because of the vaccuum permittivity being unitless in the CGS system. But by definition, they were referring to the same quantity but with different dimensions in SI. I will still do a quick recheck just in case, not to worry! Probably if I send the print commands to the printing module, it might help. Thanks once again :)

sympy/physics/units/dimensions.py
ShrohanMohapatra: Hi Francesco @Upabjojr .... thanks for your recommendations! this time I have separated the modules, one that takes care of printing the module and the other creates objects of class Dimension that represent those quantities (probably merely for testing purposes). I have also reduced the triple loop to an argument to the nullspace() function which does the job. Similar corresponding changes also reflected in the test program. Would like to know your reviews/suggestions :)

Upabjojr: We usually associate the printers to classes. In this case, the printer is acting on a list. It's quite unusual for SymPy. Do we really need this printer?

sympy/physics/units/tests/test_dimensionless.py
```diff
@@ -0,0 +1,105 @@
+from sympy.physics.units import Dimension
+from sympy.physics.units import DimensionSystem
+from sympy import Rational
+"""
+We need to define the base quantities globally to work with and
+define the derived quantities from there on
+"""
+
+M = Dimension('mass')
```
Upabjojr, line 9: We already have a definition of these dimensions. Could these just be imports?

sympy/physics/units/tests/test_dimensionless.py
```diff
@@ -0,0 +1,105 @@
...
+    Uc = L/T # Velocity dimensions
+    list_of_quantities = {'hbar': hbar, 'Uc' : Uc, 'omega' : omega, 'epsilon' : epsilon, 'Vm' : Vm, 'Qe' : Qe}
+    set_of_dimless_nums = dummy_si_system.get_dimensionless_numbers(list_of_quantities)
+    flag = True
+    for dimension in set_of_dimless_nums:
+        flag = flag and dummy_si_system.get_dimensional_dependencies(dimension) == {}
+    assert flag
+    # One of the values in the output shows the fine-structure interaction
+
+M = Dimension('mass')
```
Upabjojr, line 68: Redefinition

sympy/physics/units/tests/test_dimensionless.py
```diff
@@ -0,0 +1,105 @@
...
+We need to define the base quantities globally to work with and
+define the derived quantities from there on
+"""
+
+M = Dimension('mass')
+L = Dimension('length')
+T = Dimension('time')
+I = Dimension('current')
+
+dummy_si_system = DimensionSystem((L, M, T, I))
```
Upabjojr, line 14: why not just import the proper SI system?

sympy/physics/units/tests/test_dimensionless.py
```diff
@@ -0,0 +1,105 @@
...
+    epsilon = I**2*T**4*M**(-1)*L**(-3) # Vaccuum permittivity
+    Vm = L**3 # Volume of the resonator
+    Qe = I*T # Dimensions of the electronic charge
+    Uc = L/T # Velocity dimensions
+    list_of_quantities = {'hbar': hbar, 'Uc' : Uc, 'omega' : omega, 'epsilon' : epsilon, 'Vm' : Vm, 'Qe' : Qe}
+    set_of_dimless_nums = dummy_si_system.get_dimensionless_numbers(list_of_quantities)
+    flag = True
+    for dimension in set_of_dimless_nums:
+        flag = flag and dummy_si_system.get_dimensional_dependencies(dimension) == {}
+    assert flag
```
Upabjojr, line 65: why not assert on the output of `.get_dimensionless_numbers` ?

sympy/physics/units/tests/test_dimensionless.py
```diff
@@ -0,0 +1,105 @@
...
+def test_navier_stokes_case_si():
+    """ In this case my function will be able to generate one set of
+    dimensionless quantities in Navier-Stokes equations ......"""
+    Lc = L # Characteristic length scale
+    Uc = L/T # Characteristic velocity scale
+    rho_c = M/L**3 # Characteristic density scale
+    mu = M/L/T # Dynamic viscosity
+    p_c = M/L/T**2 # Characteristic pressure scale
+    F_bc = M*L/T**2 # Characteristic body forces scale
+    list_of_quantities = {'Lc': Lc, 'Uc': Uc, 'rho': rho_c, 'mu': mu, 'pc': p_c, 'Fbc': F_bc}
```
Upabjojr, line 25: why create a dictionary? What about just passing a list instead?

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,51 @@ def is_consistent(self):
         # dimensions
         # in vector language: the set of vectors do not form a basis
         return self.inv_can_transf_matrix.is_square
+    def buckingham_pi_theorem(self, list_of_derived_quantities):
+        """
+        This function is designed to take in a set of physical variables (all essentially dimensionful) and returns one set of all possible exponents,
+        as guided by the Buckingham's pi theorem"""
```
Upabjojr, line 578: Can you add some usage examples? The ones with `>>> ` notation where everyone can see the output. It makes the code easier to read and the documentation gets filled.

sympy/physics/units/tests/test_dimensionless.py
ShrohanMohapatra: Hi Francesco, @Upabjojr Actually I tried a lot on this, but I found several hurdles along this way: The resulting Dimension class either comes out to be 1 (since the input quantities are derived quantities and these are sort of stored in the base forms only) or an equivalent radical which eventually is equivalent to one. Thus I could not find in any equivalent list based representation that worked. The second one being that there are some derived quantities such as momentum (for example, there is Planck's momentum in SI dimension system in the package but not many may ask for it directly), viscosity, elasticity, gauge field units etc. which may be of explicit importance to the users. Thus I think it's appropriate to use the dictionary where the user is allowed to define the name of the quantity of their own choice, and tuple it with the corresponding dimensions.

sympy/physics/units/dimensions.py
ShrohanMohapatra: Thanks for the suggestion. Was a great new experience with docstrings :)

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,127 @@ def is_consistent(self):
...
+            >>> mu = M/L/T # Dynamic viscosity
+            >>> j0 = M**(S.One/2)/L**(S.One/2)/T**2 # Characteristic range value of current density
+            >>> sigma_0 = T**(-1) # Characteristic conductivity scale
+            >>> B0 = L**(-S.One/2)*M**(S.One/2)/T # Magnetic field characteristic scale
+            >>> dict_of_quantities = {'Lc': Lc, 'Uc' : Uc, 'rho': rho_c, 'mu': mu, 'j0': j0, 'sigma_0': sigma_0, 'B0': B0}
+            >>> dimsys_cgs.get_dimensionless_numbers(dict_of_quantities)
+            [Dimension(1),
+             Dimension(sqrt(mass)/(length**(3/2)*sqrt(mass/length**3))),
+             Dimension(1),
+             Dimension(sqrt(mass)/(length**(3/2)*sqrt(mass/length**3)))]
```
Upabjojr, line 639: The input is a dictionary while the output is a list. That does not seem very consistent.

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,126 @@ def is_consistent(self):
...
+            >>> p_c = M/L/T**2 # Characteristic pressure scale
+            >>> F_bc = M*L/T**2 # Characteristic body forces scale
+            >>> list_of_quantities = [('Lc',Lc), ('Uc',Uc), ('rho',rho_c), ('mu',mu), ('p_c',p_c), ('Fbc',F_bc)]
+            >>> dimsys_SI.verify_dimensionless_numbers(list_of_quantities)
+            [Dimension(1), Dimension(1), Dimension(1)]
+            >>> set_of_dimless_nums = dimsys_SI.verify_dimensionless_numbers(list_of_quantities)
+            >>> flag = True
+            >>> for dimension in set_of_dimless_nums:
+            ...     flag = flag and dimsys_SI.get_dimensional_dependencies(dimension) == {}
+            >>> flag
```
Upabjojr, line 608: This code is definitely hard to understand... This should not go into the documentation. Can you rather think of some easy examples of applications that a SymPy user may need that can be explained with some theory in the comments?

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,126 @@ def is_consistent(self):
...
+            True
+            >>> dimsys_SI.get_dimensionless_numbers(list_of_quantities)
+            [Dimension(mu/(Lc*Uc*rho)),
+             Dimension(p_c/(Uc**2*rho)),
+             Dimension(Fbc/(Lc**2*Uc**2*rho))]
+            >>> # The first line of the output is the Reynolds number Re = mu/(Uc*Lc*rho)
+            >>> # The second line of the output is the Euler number Eu = pc/(rho*Uc**2)
+            >>> # The third line of the output is the Froude number Fr, such that Fr**2 = Fbc/(rho*Uc**2*Lc**2)
```
Upabjojr, line 616: `>>> #` is not necessary in a docstring.

By the way, you can use LaTeX in a docstring using the backtick `\`` symbol. Maybe some equations in the explanation would make the code easier to understand?

ShrohanMohapatra: Thanks Francesco (@Upabjojr ) for this one, made it much easier for me to try typesetting in the docstring. Also I have tried writing a much simpler example with an RLC circuit where we get more than one relevant dimensionless quantities popular in use, also provided a wiki-reference there, I hope it works well with the documentation. Awaiting you reviews :)

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,87 @@ def is_consistent(self):
...
+        list_of_derived_quantities is the input of derived quantities in the form of a list of tuples (quantity_name, dims) where quantity_name
+        is the name of the physical quantity and dims is an instance of Dimension class which possesses the dimensions of the quantity. This function
+        is designed to take in a set of physical variables (all essentially dimensionful) and returns one set of all possible exponents,
+        as guided by the Buckingham's pi theorem (raise the list of quantities to the list of exponents to get the dimensionless numbers).
+
+        One of the next functions get_dimensionless_numbers() is used to produce the requisite result in the form of a list of instances of Dimension
+        class signifying the actual dimensionless quantities.
+
+        For example, consider an RLC circuit with resistance ```$R$```, inductance ```$L$``` and capacitance ```$C$``` connected to a constant
+        voltage source ```$V_{amp}$```, where current ```$I(t)$``` is given by the equation (https://en.wikipedia.org/wiki/RLC_circuit).
```
Upabjojr, line 587: Sorry for my previous message, I think it works with just one backtick (I erroneously wrote three earlier). Have you tried running `make html` in `sympy/doc`? It should create the HTML documentation so you can see whether the LaTeX expressions are rendered correctly.

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,87 @@ def is_consistent(self):
...
+        >>> from sympy.physics.units.systems.si import dimsys_SI
+        >>> resistance = mass*length**2*time**(-3)*current**(-2)
+        >>> inductance = mass*length**2*time**(-2)*current**(-2)
+        >>> capacitance = mass**(-1)*length**(-2)*time**4*current**2
+        >>> voltage_amp = mass*length**2*time**(-3)*current**(-1)
+        >>> current_amp = current
+        >>> omega = time**(-1)
+        >>> list_of_quantities = [('R', resistance),('L',inductance),('C',capacitance),('V_amp',voltage_amp),('I_amp',current_amp),('omega',omega)]
+        >>> dimsys_SI.get_dimensionless_numbers(list_of_quantities)
+        [Dimension(C*R**2/L), Dimension(I_amp*R/V_amp), Dimension(L*omega/R)]
```
Upabjojr, line 607: OK, this documentation starts looking a lot better now!

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,87 @@ def is_consistent(self):
         # dimensions
         # in vector language: the set of vectors do not form a basis
         return self.inv_can_transf_matrix.is_square
+
+    def buckingham_pi_theorem(self, list_of_derived_quantities):
+        """
```
Upabjojr, line 577: You can put an `r` before the string start in order to require only one backslash `\` in the string, i.e. start the string with `r""" ... `

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,87 @@ def is_consistent(self):
...
+
+        >>> from sympy.physics.units import length, mass, time, current
+        >>> from sympy.physics.units.systems.si import dimsys_SI
+        >>> resistance = mass*length**2*time**(-3)*current**(-2)
+        >>> inductance = mass*length**2*time**(-2)*current**(-2)
+        >>> capacitance = mass**(-1)*length**(-2)*time**4*current**2
+        >>> voltage_amp = mass*length**2*time**(-3)*current**(-1)
+        >>> current_amp = current
+        >>> omega = time**(-1)
+        >>> list_of_quantities = [('R', resistance),('L',inductance),('C',capacitance),('V_amp',voltage_amp),('I_amp',current_amp),('omega',omega)]
```
Upabjojr, line 605: Would it be easier to just pass the dimensions? For example, `list_of_quantities = [resistance, inductance, capacitance, ... ]` instead of tuples with their string abbreviations? Does the algorithm really need to know that `'R'` is the abbreviation for `resistance`? I think this info is not necessary.

When creating a new API, it's important to make it as simple as you can, so that new SymPy users will find it faster to learn how to use it. If the API is too complicated, people will probably not use this code, which is a pity!

(By the way, SymPy is trying not to change the API of code that has already been released, so it's important to create a good API at the first Pull Request, I know this makes things longer, but I hope you understand the problem)

ShrohanMohapatra: Hi @Upabjojr Francesco, sorry I had exams on the way, so it took me a while to come up with solutions. Finally I have managed to reduce the parameters to the buckingham_pi_theorem() function to a simple list, and you can check out the new edition in the recent commits. Actually I tried a lot on the make_html but that did not work in my system somehow. And also, I am currently seeing a new test by the name CircleCI, which's not going through. Could you please help me?

Thanks in advance
Shrohan

sympy/physics/units/dimensions.py
```diff
@@ -572,3 +572,99 @@ def is_consistent(self):
...
+
+    def get_dimensionless_numbers(self, *list_of_derived_quantities):
+        """
+        Give the resultant list the quantities (as an instance of class Dimension), which are actually the dimensionless numbers
+        in the original unit system.
+        """
+        exponents = self.buckingham_pi_theorem(*list_of_derived_quantities)
+        import traceback, re
+        args_storer = []
+        for elems in traceback.extract_stack(): args_storer.append(elems.line)
```
asmeurer, line 656: What is this code doing? Surely there is a way to do this that doesn't involve parsing traceback strings.

ShrohanMohapatra: So my intention with the Buckingham Pi API is that the user may be able to input his set of dimensions as named variables (or even simply a list of dimensionful quantities as simple instances of class Dimension) and the possible forms of output are two in number:

1. Either named variables are passed (by that I mean something like force = Dimension('mass*length/time**2'), density = Dimension('mass/length**3'), diffusivity = Dimension('length**2/time')  and so on) and then in the output the user gets to see the actual 'dimensionless quantities' as power law combinations that comes from the buckingham pi procedure (for example, [Dimension('force/density/diffusivity**2'), ..., ..., ...]). To recall those names, I wanted the addresses and the variable names passed as a parameter to the function live in scope so that's why I used strings from the traceback.stack(). I also tried using packages like inspect, symtable and ast etc. but not much to avail. But in the hindsight I feel this approach might ain't work in case of a user merely passing a list of Dimension objects without declaring a new variable. But still, I am welcome to suggestions in this option.

2. Just return the null-space of the exponent matrix and explain in the documentation to the user as to how to interpret the output with examples. This sounds more traditional and feasible at the time being. Here I wanted to ask you a question: do IDEs like Anaconda, Jupyter have Sphinx by default? I had Anaconda quite a while ago but since I have a limited memory Mac and I received an FEM project so I had to compromise my Anaconda. The question arises because my terminal does not process make html command from sympy/docs and is unable to render the LaTeX part of the documentation and hence unable to decipher problems with CircleCI. Can you please help me with this? Thanks a lot in advance :)

asmeurer: I'm still not completely following what you are trying to do (and you need to fix the formatting in your comment). However, in general, you should avoid functions that take in expressions as strings. Expressions should be constructed directly from the objects. 

We have a guide for building the documentation here https://docs.sympy.org/dev/guides/contributing/build-docs.html.

ShrohanMohapatra: HI Aaron ... Thanks a lot for the suggestions and the help with the docs. I have removed the part with the strings completely and retained the original form of the definitions with appropriate description in the documentation part of the code. I think Sphinx works for me now, and helped me quite a lot with the results of the LaTeX typesetting. Would like to know your reviews on the same :)

sympy/physics/units/dimensions.py
ShrohanMohapatra: Hi @Upabjojr Francesco, would like to have some of your final reviews on this patch, if at all it could be merged into the next release (or merge of this pull request). Currently I feel there is only some issue with the CircleCI which I could not figure out at all. And there were some other relevant pieces of advice which have also been incorporated so ....

PR #23325, opened 2022-04-06
22531: as_coefficients_dict accepts symbols

### LINKED ISSUE #22531: allow for specification of symbols for `Expr.as_coefficients_dict`
On SO the question often arises about how to get the coefficients of terms in sum.
* [example1](https://stackoverflow.com/questions/22955888/how-to-extract-all-coefficients-in-sympy)
* [example2](https://stackoverflow.com/questions/69958709/collecting-sub-expressions-of-a-multivariate-polynomial-function-in-sympy/69959128?noredirect=1#comment123675865_69959128)
* [example3](https://stackoverflow.com/questions/70037738/how-to-get-the-coefficients-of-the-equation-in-sympy/70040012#70040012)

In one of the answers I suggested the following whose code:
```python
def codict(expr, *x):
  collected = Poly(expr, *x).as_expr()
  i, d = collected.as_independent(*x, as_Add=True)
  rv = dict(i.as_independent(*x, as_Add=False)[::-1] for i in Add.make_args(d))
  if i:
      assert 1 not in rv
      rv.update({S.One: i})
  return rv
```
Examples:
```python
>>> var('a x z y')
(a, x, z, y)
>>> expr = 3 + x + x**2 + a*x*2
>>> codict(expr, x)
{x**2: 1, x: 2*a + 1, 1: 3}
>>> codict(expr+y+z, x)
{x**2: 1, x: 2*a + 1, 1: y + z + 3}
>>> codict(expr+y+z, x,y)
{y: 1, x**2: 1, x: 2*a + 1, 1: z + 3}
>>> codict(expr+y+z, x,y,z)
{y: 1, z: 1, x**2: 1, x: 2*a + 1, 1: 3}
```
It might make sense to let this be renamed `as_coefficients_dict` so one may indicate the symbols of interest for the collection of coefficients.

In [example3](https://stackoverflow.com/questions/70037738/how-to-get-the-coefficients-of-the-equation-in-sympy/70040012#70040012) I showed how Polys can return the raw information about terms and their coefficients. It should be easy to create an "all_coeffs" response for a multivariate expression by returning 0 for any missing monomials and generating all monomials as partitions of the maximum order represented, n, in k variables, e.g. for `z**4 + x**2*y + 2*y**2` the maximum order is 4 and there are are three generators so all coefficients would correspond to monomials of:
```python
>>> Poly(Add(*[Mul(*i) for i in subsets([x]*4+[y]*4+[z]*4, 4)])).monoms()
[(4, 0, 0), (3, 1, 0), (3, 0, 1), (2, 2, 0), (2, 1, 1), (2, 0, 2), (1, 3, 0), (1, 2, 1), (1, 1, 2), (1, 0, 3), (0, 4, 0), (0, 3, 1), (0, 2, 2), (0, 1, 3), (0, 0, 4)]
```
<there's probably a better way of generating those or that Polynomial>

smichr:

closes #22531

#### Release Notes

* core
    * `Expr.as_coefficient_dict` now accepts one or more symbols so symbolic coefficients can be extracted

### DIFF
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -1982,15 +1982,35 @@ class Expr(Basic, EvalfMixin):
         {x: 3*a}
 
         """
-        d = defaultdict(int)
-        c, m = self.as_coeff_mul(*syms)
-        if not c.is_Rational:
-            c = S.One
-            m = self
+        if not syms:
+            d = defaultdict(list)
+            for ai in Add.make_args(self):
+                c, m = ai.as_coeff_Mul()
+                d[m].append(c)
+            for k, v in d.items():
+                if len(v) == 1:
+                    d[k] = v[0]
+                else:
+                    d[k] = Add(*v)
+            di = defaultdict(int)
+            di.update(d)
+            return di
         else:
-            m = Mul(*m, evaluate=False)
-        d.update({m: c})
-        return dict(d) if syms else d
+            d = defaultdict(list)
+            ind, dep = self.as_independent(*syms, as_Add=True)
+            for i in Add.make_args(dep):
+                if i.is_Mul:
+                    c, x = i.as_coeff_mul(*syms)
+                    if c is S.One:
+                        d[i].append(c)
+                    else:
+                        d[i._new_rawargs(*x)].append(c)
+                else:
+                    d[i].append(S.One)
+            d = {k: Add(*d[k]) for k in d}
+            if ind is not S.Zero:
+                d.update({S.One: ind})
+            return d
 
     def as_base_exp(self) -> tuple[Expr, Expr]:
         # a -> b ** e
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -1960,8 +1960,7 @@ class Expr(Basic, EvalfMixin):
     def as_coefficients_dict(self, *syms):
         """Return a dictionary mapping terms to their Rational coefficient.
         Since the dictionary is a defaultdict, inquiries about terms which
-        were not present will return a coefficient of 0. If an expression is
-        not an Add it is considered to have a single term.
+        were not present will return a coefficient of 0.
 
         If symbols ``syms`` are provided, any multiplicative terms
         independent of them will be considered a coefficient and a
@@ -1971,7 +1970,7 @@ class Expr(Basic, EvalfMixin):
         Examples
         ========
 
-        >>> from sympy.abc import a, x
+        >>> from sympy.abc import a, x, y
         >>> (3*x + a*x + 4).as_coefficients_dict()
         {1: 4, x: 3, a*x: 1}
         >>> _[a]
@@ -1980,10 +1979,12 @@ class Expr(Basic, EvalfMixin):
         {a*x: 3}
         >>> (3*a*x).as_coefficients_dict(x)
         {x: 3*a}
+        >>> (3*a*x).as_coefficients_dict(y)
+        {1: 3*a*x}
 
         """
+        d = defaultdict(list)
         if not syms:
-            d = defaultdict(list)
             for ai in Add.make_args(self):
                 c, m = ai.as_coeff_Mul()
                 d[m].append(c)
@@ -1992,11 +1993,7 @@ class Expr(Basic, EvalfMixin):
                     d[k] = v[0]
                 else:
                     d[k] = Add(*v)
-            di = defaultdict(int)
-            di.update(d)
-            return di
         else:
-            d = defaultdict(list)
             ind, dep = self.as_independent(*syms, as_Add=True)
             for i in Add.make_args(dep):
                 if i.is_Mul:
@@ -2005,12 +2002,14 @@ class Expr(Basic, EvalfMixin):
                         d[i].append(c)
                     else:
                         d[i._new_rawargs(*x)].append(c)
-                else:
+                elif i:
                     d[i].append(S.One)
             d = {k: Add(*d[k]) for k in d}
             if ind is not S.Zero:
                 d.update({S.One: ind})
-            return d
+        di = defaultdict(int)
+        di.update(d)
+        return di
 
     def as_base_exp(self) -> tuple[Expr, Expr]:
         # a -> b ** e
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -1957,11 +1957,16 @@ class Expr(Basic, EvalfMixin):
         d.update(dict([self.as_base_exp()]))
         return d
 
-    def as_coefficients_dict(self):
+    def as_coefficients_dict(self, *syms):
         """Return a dictionary mapping terms to their Rational coefficient.
         Since the dictionary is a defaultdict, inquiries about terms which
         were not present will return a coefficient of 0. If an expression is
         not an Add it is considered to have a single term.
+
+        If symbols ``syms`` are provided, any multiplicative terms
+        independent of them will be considered a coefficient and a
+        regular dictionary of syms-dependent generators as keys and
+        their corresponding coefficients as values will be returned.
 
         Examples
         ========
@@ -1973,15 +1978,19 @@ class Expr(Basic, EvalfMixin):
         0
         >>> (3*a*x).as_coefficients_dict()
         {a*x: 3}
+        >>> (3*a*x).as_coefficients_dict(x)
+        {x: 3*a}
 
         """
-        c, m = self.as_coeff_Mul()
+        d = defaultdict(int)
+        c, m = self.as_coeff_mul(*syms)
         if not c.is_Rational:
             c = S.One
             m = self
-        d = defaultdict(int)
+        else:
+            m = Mul(*m, evaluate=False)
         d.update({m: c})
-        return d
+        return dict(d) if syms else d
 
     def as_base_exp(self) -> tuple[Expr, Expr]:
         # a -> b ** e
--- a/sympy/core/tests/test_expr.py
+++ b/sympy/core/tests/test_expr.py
@@ -1554,6 +1554,8 @@ def test_as_coefficients_dict():
     assert eq.as_coefficients_dict(x) == {x: b, 1/x: c,
         x*(x + 1): a}
     assert eq.expand().as_coefficients_dict(x) == {x**2: a, x: a + b, 1/x: c}
+    assert x.as_coefficients_dict() == {S.One: x}
+
 
 def test_args_cnc():
     A = symbols('A', commutative=False)
--- a/sympy/core/tests/test_expr.py
+++ b/sympy/core/tests/test_expr.py
@@ -1552,9 +1552,8 @@ def test_as_coefficients_dict():
     assert (3.0*x*y).as_coefficients_dict()[3.0*x*y] == 0
     eq = x*(x + 1)*a + x*b + c/x
     assert eq.as_coefficients_dict(x) == {x: b, 1/x: c,
-        x*(x + 1): a, 1: 0}
-    assert eq.expand().as_coefficients_dict(x) == {x: a + b, x**2: a,
-        1/x: c, 1: 0}
+        x*(x + 1): a}
+    assert eq.expand().as_coefficients_dict(x) == {x**2: a, x: a + b, 1/x: c}
 
 def test_args_cnc():
     A = symbols('A', commutative=False)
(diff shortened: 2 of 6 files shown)


oscarbenjamin:
It seems odd that `as_coefficients_dict` has separate implementations for Expr, Add and Mul but these don't call each other at all. Should it not just be a single implementation for Expr that uses `Add.make_args` so it always treats the expression as an Add?

sympy/core/expr.py
```diff
@@ -1980,15 +1985,32 @@ def as_coefficients_dict(self):
...
+                    d[k] = Add(*v)
+            di = defaultdict(int)
+            di.update(d)
+            return di
+        else:
+            d = defaultdict(list)
+            ind, dep = self.as_independent(*syms, as_Add=True)
+            for i in Add.make_args(dep):
+                c, x = i.as_independent(*syms, as_Add=False)
+                d[x].append(c)
```
smichr, line 2010:
```suggestion
                if i.is_Mul:
                    c, x = i.as_coeff_mul(*syms)
                    if c is S.One:
                        d[i].append(c)
                    else:
                        d[i._new_rawargs(*x)].append(c)
                else:
                    d[i].append(S.One)
```

smichr:
One thing to note is that the `Mul` version was cached. I modified the `Expr` routine to use `as_coeff_mul` which is also cached (and will be faster at creating the symbol-dependent portion of the expression from which the coefficient was extracted).

sympy/core/expr.py
```diff
@@ -1980,15 +1985,38 @@ def as_coefficients_dict(self):
...
+                if i.is_Mul:
+                    c, x = i.as_coeff_mul(*syms)
+                    if c is S.One:
+                        d[i].append(c)
+                    else:
+                        d[i._new_rawargs(*x)].append(c)
+                else:
+                    d[i].append(S.One)
+            d = {k: Add(*d[k]) for k in d}
+            d.update({S.One: ind})
```
smichr, line 2018:
```suggestion
            if ind is not S.Zero:
                d.update({S.One: ind})
```

sympy/core/tests/test_expr.py
```diff
@@ -1550,7 +1550,11 @@ def test_as_coefficients_dict():
     assert [(3.0*x*y).as_coefficients_dict()[i] for i in check] == \
         [0, 0, 0, 3.0, 0]
     assert (3.0*x*y).as_coefficients_dict()[3.0*x*y] == 0
-
+    eq = x*(x + 1)*a + x*b + c/x
+    assert eq.as_coefficients_dict(x) == {x: b, 1/x: c,
+        x*(x + 1): a, 1: 0}
```
smichr, line 1555:
```suggestion
        x*(x + 1): a}
```

sympy/core/expr.py
```diff
@@ -1980,15 +1985,39 @@ def as_coefficients_dict(self):
         0
         >>> (3*a*x).as_coefficients_dict()
         {a*x: 3}
+        >>> (3*a*x).as_coefficients_dict(x)
+        {x: 3*a}
```
smichr, line 1989:
```suggestion
        {x: 3*a}
        >>> from sympy.abc import b, c
        >>> (x*(x + 1)*a + x*b + c/x).as_coefficients_dict(x)
```

sympy/core/tests/test_expr.py
```diff
@@ -1550,7 +1550,11 @@ def test_as_coefficients_dict():
     assert [(3.0*x*y).as_coefficients_dict()[i] for i in check] == \
         [0, 0, 0, 3.0, 0]
     assert (3.0*x*y).as_coefficients_dict()[3.0*x*y] == 0
-
+    eq = x*(x + 1)*a + x*b + c/x
+    assert eq.as_coefficients_dict(x) == {x: b, 1/x: c,
+        x*(x + 1): a}
+    assert eq.expand().as_coefficients_dict(x) == {x: a + b, x**2: a,
+        1/x: c, 1: 0}
```
smichr, line 1557:
```suggestion
    assert eq.expand().as_coefficients_dict(x) == {x: b, 1/x: c, x*(x + 1): a}
```

sympy/core/tests/test_expr.py
```diff
@@ -1550,7 +1550,10 @@ def test_as_coefficients_dict():
     assert [(3.0*x*y).as_coefficients_dict()[i] for i in check] == \
         [0, 0, 0, 3.0, 0]
     assert (3.0*x*y).as_coefficients_dict()[3.0*x*y] == 0
-
+    eq = x*(x + 1)*a + x*b + c/x
+    assert eq.as_coefficients_dict(x) == {x: b, 1/x: c,
+        x*(x + 1): a}
+    assert eq.expand().as_coefficients_dict(x) == {x: b, 1/x: c, x*(x + 1): a}
```
smichr, line 1556:
```suggestion
    assert eq.expand().as_coefficients_dict(x) == {x**2: a, x: a + b, 1/x: c}
```

smichr:
@oscarbenjamin does this seem like something we should add:
```diff
diff --git a/sympy/simplify/radsimp.py b/sympy/simplify/radsimp.py
index 91078be3af..2b8407809e 100644
--- a/sympy/simplify/radsimp.py
+++ b/sympy/simplify/radsimp.py
@@ -121,6 +121,12 @@ def collect(expr, syms, func=None, evaluate=None, exact=False, distribute_order_
         >>> collect(a*x**7 + b*x**7, x**7, exact=True)
         x**7*(a + b)
 
+    If you want to collect on any object containing symbols, set
+    ``exact`` to None:
+
+        >>> collect(x*exp(x) + sin(x)*y + sin(x)*2 + 3*x, x, exact=None)
+        x*exp(x) + (y + 2)*sin(x) + 3*x
+
     You can also apply this function to differential equations, where
     derivatives of arbitrary order can be collected. Note that if you
     collect with respect to a function or a derivative of a function, all
@@ -165,6 +171,7 @@ def collect(expr, syms, func=None, evaluate=None, exact=False, distribute_order_
     """
     expr = sympify(expr)
     syms = [sympify(i) for i in (syms if iterable(syms) else [syms])]
+
     # replace syms[i] if it is not x, -x or has Wild symbols
     cond = lambda x: x.is_Symbol or (-x).is_Symbol or bool(
         x.atoms(Wild))
@@ -182,6 +189,25 @@ def collect(expr, syms, func=None, evaluate=None, exact=False, distribute_order_
             return {urep.get(k, k).xreplace(urep): v.xreplace(urep)
                     for k, v in rv.items()}
 
+    # see if other expressions should be considered
+    if exact is None:
+        _syms = set()
+        for i in Add.make_args(expr):
+            if not i.has_free(*syms):
+                continue
+            if not i.is_Mul and i not in syms:
+                _syms.add(i)
+            else:
+                g = i._new_rawargs(*i.as_coeff_mul(*syms)[1])
+                if g not in syms:
+                    _syms.add(g)
+        simple = all(i.is_Pow and i.base in syms for i in _syms)
+        syms = syms + list(ordered(_syms))
+        if not simple:
+            return collect(expr, syms,
+            func=func, evaluate=evaluate, exact=False,
+            distribute_order_term=distribute_order_term)
+
     if evaluate is None:
         evaluate = global_parameters.evaluate
 
```

sympy/core/expr.py
```diff
@@ -1980,15 +1985,39 @@ def as_coefficients_dict(self):
...
+                c, m = ai.as_coeff_Mul()
+                d[m].append(c)
+            for k, v in d.items():
+                if len(v) == 1:
+                    d[k] = v[0]
+                else:
+                    d[k] = Add(*v)
+            di = defaultdict(int)
+            di.update(d)
+            return di
```
oscarbenjamin, line 2004: Why does this return a defaultdict? It's usually better not to because they can behave in unexpected ways.

smichr: Because that is the advertised behavior of the default case (no symbols) per docstring:
```
        """Return a dictionary mapping terms to their Rational coefficient.
        Since the dictionary is a defaultdict, inquiries about terms which
        were not present will return a coefficient of 0. If an expression is
        not an Add it is considered to have a single term.
```
I'm not sure it is worth a deprecation.

sympy/core/expr.py
```diff
@@ -1957,12 +1957,17 @@ def as_powers_dict(self):
         d.update(dict([self.as_base_exp()]))
         return d
 
-    def as_coefficients_dict(self):
+    def as_coefficients_dict(self, *syms):
         """Return a dictionary mapping terms to their Rational coefficient.
         Since the dictionary is a defaultdict, inquiries about terms which
         were not present will return a coefficient of 0. If an expression is
         not an Add it is considered to have a single term.
```
smichr, line 1964:
```suggestion
```

sympy/simplify/radsimp.py
```diff
@@ -121,6 +121,12 @@ def collect(expr, syms, func=None, evaluate=None, exact=False, distribute_order_
         >>> collect(a*x**7 + b*x**7, x**7, exact=True)
         x**7*(a + b)
 
+    If you want to collect on any object containing symbols, set
+    ``exact`` to None:
+
+        >>> collect(x*exp(x) + sin(x)*y + sin(x)*2 + 3*x, x, exact=None)
+        x*exp(x) + 3*x + (y + 2)*sin(x)
```
smichr, line 128: I'm not sure why the `x*exp(x) + 3*x` doesn't become `x*(exp(x) + 3)`.

sympy/core/expr.py
```diff
@@ -1957,11 +1957,13 @@ def as_powers_dict(self):
...
+    def as_coefficients_dict(self, *syms):
         """Return a dictionary mapping terms to their Rational coefficient.
-        Since the dictionary is a defaultdict, inquiries about terms which
-        were not present will return a coefficient of 0. If an expression is
-        not an Add it is considered to have a single term.
+
+        If symbols ``syms`` are provided, any multiplicative terms
+        independent of them will be considered a coefficient and a
+        regular dictionary of syms-dependent generators as keys and
+        their corresponding coefficients as values will be returned.
```
smichr, line 1966:
```suggestion
        regular dictionary of syms-dependent generators as keys (and
        their corresponding coefficients as values) will be returned.
```

sympy/core/expr.py
```diff
@@ -1957,11 +1957,13 @@ def as_powers_dict(self):
         d.update(dict([self.as_base_exp()]))
         return d
 
-    def as_coefficients_dict(self):
+    def as_coefficients_dict(self, *syms):
         """Return a dictionary mapping terms to their Rational coefficient.
```
smichr, line 1961:
```suggestion
        """Return a dictionary mapping terms to their Rational coefficient.
            Since the dictionary is a defaultdict, inquiries about terms which
            were not present will return a coefficient of 0.
```

sympy/core/expr.py
```diff
@@ -1973,15 +1977,39 @@ def as_coefficients_dict(self):
...
+                    else:
+                        d[i._new_rawargs(*x)].append(c)
+                else:
+                    d[i].append(S.One)
+            d = {k: Add(*d[k]) for k in d}
+            if ind is not S.Zero:
+                d.update({S.One: ind})
+            dd = defaultdict(int)
+            dd.update(d)
+            return dd
```
smichr, line 2012:
```suggestion
            dd = defaultdict(int)
            dd.update(d)
            return dd
        else:
            d = defaultdict(list)
            ind, dep = self.as_independent(*syms, as_Add=True)
            for i in Add.make_args(dep):
                if i.is_Mul:
                    c, x = i.as_coeff_mul(*syms)
                    if c is S.One:
                        d[i].append(c)
                    else:
                        d[i._new_rawargs(*x)].append(c)
                else:
                    d[i].append(S.One)
            d = {k: Add(*d[k]) for k in d}
            if ind is not S.Zero:
                d.update({S.One: ind})
            return d
```

sympy/core/expr.py
```diff
@@ -1957,11 +1957,15 @@ def as_powers_dict(self):
...
         return d
 
-    def as_coefficients_dict(self):
+    def as_coefficients_dict(self, *syms):
         """Return a dictionary mapping terms to their Rational coefficient.
-        Since the dictionary is a defaultdict, inquiries about terms which
-        were not present will return a coefficient of 0. If an expression is
-        not an Add it is considered to have a single term.
+            Since the dictionary is a defaultdict, inquiries about terms which
+            were not present will return a coefficient of 0.
```
smichr, line 1963:
```suggestion
        Since the dictionary is a defaultdict, inquiries about terms which
        were not present will return a coefficient of 0.
```

smichr:
It is an unecessary break of expectations for function to return dict instead of defaultdict, @oscarbenjamin. `as_powers_dict` does the same thing. What is the unexpected behavior that you are concerned about?

oscarbenjamin:
If it already returns a defaultdict then I guess it doesn't need to be changed.

PR #23329, opened 2022-04-07
Added methods for nilpotent, abelian and cyclic numbers.

OrestisVaggelis:
Implemented methods to check if a number is nilpotent, abelian or cyclic. Details about the algorithm can be found in
[this](http://www2.math.ou.edu/~shankar/papers/nil2.pdf) paper.

#### Release Notes

* combinatorics
  * Added methods for nilpotent, abelian and cyclic numbers.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -122,8 +122,8 @@ Alan Bromborsky <abrombo@verizon.net> <brombo@GA.(none)>
 Alec Kalinin <alec.kalinin@gmail.com>
 Aleksandar Makelov <amakelov@college.harvard.edu>
 Alex Argunov <sajkoooo@gmail.com>
-Alex Lindsay <adlinds3@ncsu.edu> <lindsayad@users.noreply.github.com>
 Alex Lindsay <adlinds3@ncsu.edu> Alex Lindsay <alexlindsay239@gmail.com>
+Alex Lindsay <adlinds3@ncsu.edu> <lindsayad@users.noreply.github.com>
 Alex Lubbock <code@alexlubbock.com>
 Alex Malins <github@alexmalins.com>
 Alex Meiburg <timeroot.alex@gmail.com>
@@ -330,8 +330,8 @@ David Joyner <wdjoyner@gmail.com>
 David Joyner <wdjoyner@gmail.com> <wdj@tinah.(none)>
 David Ju <Sgtmook314@gmail.com> David Ju <david@David-PC.(none)>
 David Lawrence <dmlawrence@gmail.com>
-David Li <l33tnerd.li@gmail.com> <li.davidm96@gmail.com>
 David Li <l33tnerd.li@gmail.com> unknown <David@David-PC.(none)>
+David Li <l33tnerd.li@gmail.com> <li.davidm96@gmail.com>
 David Marek <h4wk.cz@gmail.com> <davidm@atrey.karlin.mff.cuni.cz>
 David Menéndez Hurtado <david.menendez.hurtado@scilifelab.se>
 David P. Sanders <dpsanders@gmail.com>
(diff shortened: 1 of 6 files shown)


AUTHORS
```diff
@@ -1133,3 +1133,13 @@ Tom Fryers <61272761+TomFryers@users.noreply.github.com>
...
+Siddhant Jain <siddhantashoknagar@gmail.com>
+Tirthankar Mazumder <63574588+wermos@users.noreply.github.com>
+Sumit Kumar <mr.sumitkrr@gmail.com>
+Shivam Sagar <technoshivam12@gmail.com>
+Gaurav Jain <gjain369@gmail.com>
+Andrii Oriekhov <andriyorehov@gmail.com>
+Luis Talavera <luisfertalavera15@gmail.com>
+Arie Bovenberg <a.c.bovenberg@gmail.com>
+Carson McManus <carson.mcmanus1@gmail.com>
+Jack Schmidt <1107865+jackschmidt@users.noreply.github.com>
```
oscarbenjamin, line 1145: The changes to this file should be reverted. The AUTHORS file is updated at the time of the next release. It's only necessary to add your details in the .mailmap file.

.mailmap
```diff
@@ -831,7 +831,7 @@ Ondřej Čertík <ondrej@certik.cz> <ondrej.certik@gmail.com>
 Ondřej Čertík <ondrej@certik.cz> <ondrej@certik.us>
 Ondřej Čertík <ondrej@certik.cz> ondrej.certik <devnull@localhost>
 Or Dvory <gidesa@gmail.com>
-OrestisVaggelis <orestis22000@gmail.com>
+OrestisVaggelis <orestis22000@gmail.com> Orestis Vaggelis <orestisvaggelis@mail.com>
```
oscarbenjamin, line 834: Which name and email do you want to appear in the AUTHORS file?

OrestisVaggelis: I'd prefer Orestis Vaggelis <orestisvaggelis@mail.com> to appear, but without the first one, I was getting an error during the tests.

sympy/combinatorics/group_numbers.py
```diff
@@ -0,0 +1,119 @@
...
+    >>> is_nilpotent_number(21)
+    False
+    >>> is_nilpotent_number(randprime(1, 30)**12)
+    True
+
+    References
+    ==========
+
+    .. [1] Pakianathan, J., Shankar, K., Nilpotent Numbers,
+           The American Mathematical Monthly
```
jksuom, line 25: There should be more details like year, issue, pages..

.mailmap
oscarbenjamin: Then the above should be reversed like:
```
Orestis Vaggelis <orestisvaggelis@mail.com>  OrestisVaggelis <orestis22000@gmail.com>
```
Also you should remove the line `OrestisVaggelis <orestis22000@gmail.com>` from the AUTHORS file. The new name will then be added to AUTHORS automatically at the time of the next release.

AUTHORS
```diff
@@ -1133,3 +1132,9 @@ Tom Fryers <61272761+TomFryers@users.noreply.github.com>
 Zouhair <zouhair.mahboubi@gmail.com>
 zzj <29055749+zjzh@users.noreply.github.com>
 shubhayu09 <guptashubhayu601@gmail.com>
+Siddhant Jain <siddhantashoknagar@gmail.com>
+Tirthankar Mazumder <63574588+wermos@users.noreply.github.com>
+Sumit Kumar <mr.sumitkrr@gmail.com>
+Shivam Sagar <technoshivam12@gmail.com>
+Gaurav Jain <gjain369@gmail.com>
+Andrii Oriekhov <andriyorehov@gmail.com>
```
oscarbenjamin, line 1140: These lines should also be removed.

OrestisVaggelis:
Thank you for your feedback. I've now reverted the changes made in the AUTHORS file, updated the .mailmap file and added more information about the reference.

AUTHORS
```diff
@@ -4,7 +4,7 @@ those who explicitly didn't want to be mentioned. People with a * next
 to their names are not found in the metadata of the git history. This
 file is generated automatically by running `./bin/authors_update.py`.
 
-There are a total of 1127 authors.
+There are a total of 1137 authors.
```
oscarbenjamin, line 7: This change also needs to be reverted.

When looking at the PR you can click "files changed" at the top to see what is actually being changed.

OrestisVaggelis: So there should be 1126 authors now, since I removed myself, right?

oscarbenjamin: The number will be updated at the time of the next release. For now it is best just to leave it unchanged.

The reason for not changing things like this is that if any two PRs changes the same line of the same file then there will be merge conflicts. If every PR tries to change this line then every PR will have merge conflicts.

oscargus:
Should this be Abelian group in the docs? Wolfram seems to use Abelian, while Wikipedia and many other sources use abelian. Since Abel is a person it is not obvious to me as non-native.

OrestisVaggelis:
> Should this be Abelian group in the docs? Wolfram seems to use Abelian, while Wikipedia and many other sources use abelian. Since Abel is a person it is not obvious to me as non-native.

Other docs in sympy seem to use abelian as well (see permutation groups for example). Also the paper referenced uses a lower case a too.

PR #23339, opened 2022-04-08
Support `alias` for prim. elt. of `AlgebraicField`

### LINKED ISSUE #23338: LaTeX printing for `k.ext` is not good.
By `k.ext` I mean the primitive element of an `AlgebraicField` `k`.
`k.ext` is just an `AlgebraicNumber`, but, due to the way an `AlgebraicField` is constructed, things tend to go awry.
Specifically:
* printing is extremely slow
* the terms are out of order

## Example

The output of this script:

```python
import time
from sympy import QQ, latex
k = QQ.cyclotomic_field(17)
alpha = k.ext.field_element([1, 1, 1, 1, 1])

t0 = time.time()
L = latex(alpha)
t1 = time.time()

print(L)
print(f'Took {t1 - t0:.2f} seconds')
```

is:

```
1 + \zeta^{4} + \zeta + \zeta^{3} + \zeta^{2}
Took 8.63 seconds
```

which has two big problems:

* The powers of zeta should be in order (either descending or ascending), but they're all mixed up.
* Since it's intended to be a purely symbolic expression, there should be no expensive computation involved, and taking 8 seconds is unacceptable.

What's happening is that, instead of being treated like a symbol, the algebraic number is being treated like a number, and this means that `sympy.core.expr.Expr.as_terms()` is converting it with `complex()` (hence the 8 second delay) and `sympy.core.expr.Expr._parse_order()` is sorting the terms according to real part and imaginary part (hence the mixed up order).

## Expected behavior

The situation is complicated by the fact that sometimes an `AlgebraicNumber` has an alias, and sometimes it doesn't. When it has one, you want it to be treated like a symbol; when it doesn't, you want it to be treated like a number.

For example, setting no alias,

```python
>>> alpha = AlgebraicNumber(2**(S(1)/5) + 3**(S(1)/7))
```

we expect `alpha` to print as some expression in radicals, like this:

```python
>>> latex(alpha)
'\\sqrt[5]{2} + \\sqrt[7]{3}'
```

When we set an alias,

```python
>>> alpha = AlgebraicNumber(2**(S(1)/5) + 3**(S(1)/7), alias="alpha")
```

we expect `alpha` to print as just this symbol:

```python
>>> latex(alpha)
'\\alpha'
```

## The cause of the problem

When you construct an `AlgebraicField` `k`, the primitive element `k.ext` is never equal to any `AlgebraicNumber` you passed as an extension, and therefore any alias you might have set gets buried. (It's still there, but not where it needs to be.)

The reason it's done this way is so that you can pass mutiple extensions. When you say you want the field `QQ(alpha, beta, gamma)`, a primitive element `theta` is first constructed so that `QQ(theta) == QQ(alpha, beta, gamma)`, and this `theta` is what becomes `k.ext`.

But it works the same way even when you pass just a single extension element. In other words, if you define

```python
alpha = AlgebraicNumber(2**(S(1)/5) + 3**(S(1)/7), alias="alpha")
k = QQ.algebraic_field(alpha)
```

then `k.ext != alpha`. Instead, a new `AlgebraicNumber` `theta` is constructed, which has `alpha` as its `.root`:

```python
>>> theta = k.ext
>>> theta.root is alpha
True
```

Your number `alpha` still has its alias, but the new number `theta` has none:

```python
>>> theta.root.alias
alpha
>>> theta.alias is None
True
```

## Quick fix?

A quick fix would be to change the latex printing method,

https://github.com/sympy/sympy/blob/19ed9ef7c0eb8a6e135ef66289c0cb896d800dee/sympy/printing/latex.py#L626-L632

so that instead of merely checking whether `expr` had an alias, it would also check to see if `expr.root` had an alias.

But I don't think this is the right solution, because it ignores the deeper problem.

## A better solution

The root of the problem is that there's currently no way to make `k.ext` have an alias. In a hacky way you could manually set it after the fact, but SymPy classes are usually supposed to be immutable, so this isn't the intended design.

When you want a single extension `Q(alpha)`, you can "bury" an alias in `k.ext.root.alias`, as we've seen above. While this doesn't make printing work right, it's the reason why "zeta" appeared at all in the output from the example script I started with.

When you want a multiple extension `Q(alpha, beta, gamma)`, there's no way to have an alias.

So the solution I'd propose is:

1. Provide a way to set an alias for the primitive element of an `AlgebraicField`, with the end
    result that this alias is _always_ found at `k.ext.alias`, whether it was a single or multiple extension; and then

2. the LaTeX printing problem will solve itself, without even needing to change the existing method.

This has been a convoluted discussion, but I'll open a draft to make the proposed changes clear.

skieffer:

Fixes #23338 

We make it so that, when you construct an `AlgebraicField` `k`, you can set an alias for its primitive element `k.ext`.

Furthermore, if the field is constructed on a single extension, which is itself an `AlgebraicNumber` having an alias, then this alias is adopted by the `AlgebraicField` as well. As a consequence, LaTeX printing of elements of `AlgebraicField`s is repaired.

#### Release Notes

* polys
  * Support setting of an alias for the primitive element of an `AlgebraicField`
* printing
  * LaTeX printing of elements of `AlgebraicField`s is repaired (in terms of both speed and appearance).


oscarbenjamin:
This is marked as draft but it looks good to me if you're done with it.

skieffer:
> looks good to me if you're done with it.

Okay, great. Yes, I'm happy with this solution.

PR #23011, opened 2022-02-04
nonlinsolve now handles polynomials using polynomial methods

NikhilSDate:
#### References to other Issues or PRs

Fixes #22628 

#### Brief description of what is fixed or changed
A broken if condition in `nonlinsolve` that was preventing polynomial systems from being solved using polynomial methods has been fixed. This also results in less spurious solutions when solving polynomial systems.

#### Release Notes

* solvers
  * nonlinsolve now handles all polynomial equations using polynomial methods.
  * solve_poly_system accepts a keyword argument strict and raises NotImplementedError if it is not able to find a complete solution and strict is True
  * solve_generic accepts a keyword argument strict and raises NotImplementedError if it is not able to find a complete solution and strict is True

PR #23341, opened 2022-04-09
Improved functionality of invert_real() in case of Pow expressions equal to something

### LINKED ISSUE #23340: Solveset gives wrong result
The code
```
solveset(sqrt((x-3)/x), x, domain=Interval(3, +oo))
```
gives the answer `EmptySet`, which is not correct, since `x=3` is a solution.

Instead
```
solveset(sqrt(x-3), x, domain=Interval(3, +oo))
```
gives `FiniteSet(3)`, which is correct.

The issue seems in the `invert_real()` function, since `invert_real(x-3, 0, x)` gives `(x, {3})`, while `invert_real((x-3)/x, 0, x)` gives `((x - 3)/x, {0})`.

maurogaravello:

Fixes #23340 

`solveset(sqrt((x-3)/x), x, domain=Interval(3, +oo))` returns `EmptySet` in place of `FiniteSet(3)`, since `invert_real((x-3)/x, 0, x)` gives `((x - 3)/x, {0})`.

The change in the commit is in the behavior of `invert_real()` in case of `Pow`.

#### Other comments

Some tests added.

#### Release Notes

* solvers
  * A bug leading to incorrect solutions of some radical equations from `solveset` was fixed.

### DIFF
--- a/sympy/solvers/solveset.py
+++ b/sympy/solvers/solveset.py
@@ -251,12 +251,22 @@ def _invert_real(f, g_ys, symbol):
                 num, den = expo.as_numer_denom()
 
                 if den % 2 == 0 and num % 2 == 1 and den.is_zero is False:
+                    # Here we have f(x)**(num/den) = y
+                    # where den is nonzero and even and y is an element
+                    # of the set g_ys.
+                    # den is even, so we are only interested in the cases
+                    # where both f(x) and y are positive.
+                    # Restricting y to be positive (using the set g_ys_pos)
+                    # means that y**(den/num) is always positive.
+                    # Therefore it isn't necessary to also constrain f(x)
+                    # to be positive because we are only going to
+                    # find solutions of f(x) = y**(d/n)
+                    # where the rhs is already required to be positive.
                     root = Lambda(n, real_root(n, expo))
                     g_ys_pos = g_ys & Interval(0, oo)
                     res = imageset(root, g_ys_pos)
-                    base_positive = solveset(base >= 0, symbol, S.Reals)
                     _inv, _set = _invert_real(base, res, symbol)
-                    return (_inv, _set.intersect(base_positive))
+                    return (_inv, _set)
 
                 if den % 2 == 1:
                     root = Lambda(n, real_root(n, expo))
--- a/sympy/solvers/tests/test_solveset.py
+++ b/sympy/solvers/tests/test_solveset.py
@@ -403,7 +403,9 @@ def test_solveset_sqrt_1():
     assert solveset_real(sqrt(x) - 2 - 5, x) == FiniteSet(49)
     assert solveset_real(sqrt(x**3), x) == FiniteSet(0)
     assert solveset_real(sqrt(x - 1), x) == FiniteSet(1)
-
+    assert solveset_real(sqrt((x-3)/x), x) == FiniteSet(3)
+    assert solveset_real(sqrt((x-3)/x)-Rational(1, 2), x) == \
+        FiniteSet(4)
 
 def test_solveset_sqrt_2():
     x = Symbol('x', real=True)


sympy/solvers/solveset.py
```diff
@@ -239,6 +239,22 @@ def _invert_real(f, g_ys, symbol):
         if g is not S.One:
             return _invert_real(h, imageset(Lambda(n, n/g), g_ys), symbol)
 
+        if g_ys == FiniteSet(S.Zero):
+            res = S.EmptySet
+            test = True
```
eagleoflqj, line 244: You may use [for-else](https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops) to get rid of `test` variable.

sympy/solvers/solveset.py
```diff
@@ -239,6 +239,22 @@ def _invert_real(f, g_ys, symbol):
...
+            for f1 in f.args:
+                (a, b) = _invert_real(f1, g_ys, symbol)
+                if a == symbol and isinstance(b, FiniteSet):
+                    if all(checksol(f, symbol, sol) for sol in b):
+                        res = res.union(b)
```
oscarbenjamin, line 249: It would be better not to use `checksol` here but also I'm not sure how the logic can be formally correct. Why do we check with `all`? What would happen if some of the solutions are good and some are not?

I would have thought that it would be best to have something here to split the numerator and denominator. If you want to invert `f(x)/g(x) = a` then (provided `g(x)!=0`) that's the same as inverting `f(x) - a*g(x) = 0`.

Also does the same issue not arise with the rhs being nonzero like this:
```python
In [16]: solveset(sqrt((x-3)/x)-Rational(1, 2), x, Reals)
Out[16]: ∅
```
I'm not sure where the logic falls through here but at least it should be a ConditionSet rather than simply an empty set.

sympy/solvers/solveset.py
maurogaravello: Sure! Thanks! I will change the code accordingly.

sympy/solvers/solveset.py
maurogaravello: I used `checksol` to eliminate non correct solutions like `x=0` in `log(1+x)/x`, which produced a failure in tests. Is there any obstruction to the use of `checksol`?

You right. The check with `all` is not good. I will change it including the possibility that some solutions are good and some not.

I imagine that splitting numerator and denominator works in the case. But `invert_real((x-3)*(x+5), 0, x)` returns `((x - 3)*(x + 5), {0})` and in this case `invert_real` should not be affected. Also `solveset` in this case gives the wrong result.

I do not understand your last comment. `x=4` is a solution. The output seems wrong.

oscarbenjamin: Since `checksol` can be very expensive it should be used sparingly. There should be a simpler way of eliminating `x=0` from `log(1+x)/x` i.e. we want roots of the numerator that are not roots of the denominator.

> But `invert_real((x-3)*(x+5), 0, x)` returns `((x - 3)*(x + 5), {0})`

Well `solveset` can solve this so where in `solveset` is this handled? Maybe the case considered here should be handled there as well.

> I do not understand your last comment. `x=4` is a solution. The output seems wrong.

Yes, the output is wrong and the changes here don't fix it even though it is the same function to be inverted. That's because the solution here doesn't really invert the function: it's a special case for when the rhs is zero.

maurogaravello: What do you think about the code?
```
        if g_ys == FiniteSet(S.Zero):
            res = _solve_as_rational(f, symbol, S.Reals)
            return (symbol, res)
```
The function `_solve_as_rational()` works in the spirit of your suggestion.

> Yes, the output is wrong and the changes here don't fix it even though it is the same function to be inverted. That's because the solution here doesn't really invert the function: it's a special case for when the rhs is zero.

I take a look of the case for understanding where the logic does not work.

oscarbenjamin: I'm not sure that this really belongs in `invert` rather than somewhere else. It doesn't look like `invert` is intended to handle anything nontrivial that can't just be inverted one function at a time e.g.:
```python
In [20]: invert_real(x**2 + x, 0, x)
Out[20]: 
⎛ 2         ⎞
⎝x  + x, {0}⎠
```

eagleoflqj: > I'm not sure that this really belongs in `invert` rather than somewhere else. It doesn't look like `invert` is intended to handle anything nontrivial that can't just be inverted one function at a time e.g.:

Solving `f(g(x))=0` can be simplified when `invert_real` becomes more powerful, as we proceed to solving `g(x)=f_inv(0)` instead. In your example `g(x)=x` which indeed doesn't show benefit. But what if `g(x)` is much more complicated?

maurogaravello: After some debug, I am convinced that the issue lies in a return of `_invert_real()`, in which a set were intersected with a domain for base positive.
I think that this is fundamental only if the inverse function really works.

In particular this change seems to fix also `solveset(sqrt((x-3)/x)-Rational(1, 2), x, Reals)`.

sympy/solvers/solveset.py
```diff
@@ -255,7 +256,9 @@ def _invert_real(f, g_ys, symbol):
                     res = imageset(root, g_ys_pos)
                     base_positive = solveset(base >= 0, symbol, S.Reals)
                     _inv, _set = _invert_real(base, res, symbol)
-                    return (_inv, _set.intersect(base_positive))
+                    if _inv == symbol:
+                        return (_inv, _set.intersect(base_positive))
+                    return (_inv, _set)
```
oscarbenjamin, line 261: So if `_invert_real` does not fully invert to symbol then we shouldn't automatically intersect the two sets. That makes sense but does there not need to be something else to handle the kind of case that `base_positive` is intended to handle?

maurogaravello: Probably, in order to be completely correct, we need to do the intersection with the set `base_positive` transformed with the same transformation the function `_invert_real` maps `base` to `_inv`.
I do not have any idea how to obtain the transformation used inside `_invert_real`.

oscarbenjamin: I'm not sure how to implement this either.

The change here is an improvement though so I think we could use this. We should add a comment though clearly indicating that something is not being handled properly.

What would be a test case where not handling the sign of the base properly leads to an incorrect result?

maurogaravello: Here is the code
```
                if den % 2 == 0 and num % 2 == 1 and den.is_zero is False:
                    root = Lambda(n, real_root(n, expo))
                    g_ys_pos = g_ys & Interval(0, oo)
                    res = imageset(root, g_ys_pos)
                    base_positive = solveset(base >= 0, symbol, S.Reals)
                    _inv, _set = _invert_real(base, res, symbol)
                    if _inv == symbol:
                        return (_inv, _set.intersect(base_positive))
                    return (_inv, _set)
```
We have to invert `f = g_y` and `f` is `Pow` where the exponent has not symbol (i.e. it is a constant).
In particular the exponent is rational with even denominator. In this case `res` is a set of positive numbers (I do not see what is the advantage of using `g_y_pos` instead of `g_y`).
Therefore we want to solve `base = res` using `_invert_real`. So what is the use of `base_positive`, since `res` is already positive?

By the way, I tried to comment out the code containing `base_positive`. The test `test_issue_21022` in `test_solveset.py` is the only one failing. 
The test is quite complicated and I do not know why it fails.

oscarbenjamin: That particular test was recently changed in #23011. Did you try with the old or the new version of the test?

maurogaravello: I tried, but it fails.
The problem is that `len(res)` now is `8` and not `4`. So it seems that `base_positive` needs at some point.

maurogaravello: On master it works!

maurogaravello: Yes, that makes sense.

maurogaravello:
> Okay, I think this looks good but given that it's completely different from the original changes can you squash the commits?
> 

I think I did some messy with squash. Probably it is better to open a new pull request


> Also it would be good to add some comments to this block of code that explain how it works and what it's doing.

eagleoflqj:
> I think I did some messy with squash. Probably it is better to open a new pull request

No you don't need to open a new PR.
Run
```sh
git diff f11bc5e42ce3f84a14dbd669d51429171af827b9..a8f143bb6a62c2b07614a8dddc413d624f0d215f > ~/patch
git checkout master
git branch -D invert_issue
git checkout -b invert_issue
git apply ~/patch
```
Then do `git add` and `git commit`, finally `git push -f origin invert_issue`

maurogaravello:
> 
> No you don't need to open a new PR. Run
> 
 
Thank you for your help!

sympy/solvers/solveset.py
```diff
@@ -240,6 +240,7 @@ def _invert_real(f, g_ys, symbol):
         if g is not S.One:
             return _invert_real(h, imageset(Lambda(n, n/g), g_ys), symbol)
 
+
     if f.is_Pow:
```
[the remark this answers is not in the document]
maurogaravello: Done!

oscarbenjamin:
I've edited the release note to mention `solveset`. Set to auto-merge.

Thanks!

PR #23344, opened 2022-04-10
Explicitly export __version__

eagleoflqj:
After https://github.com/sympy/sympy/pull/22337 is merged, pyright starts to complain sympy usages in user code.

test_import.py:
```py
import numpy
import sympy

print(numpy.__version__)
print(sympy.__version__)
```

For sympy 1.10.1,
```sh
$ pyright test_import.py 
No configuration file found.
No pyproject.toml file found.
stubPath .../typings is not a valid directory.
Assuming Python platform Linux
Searching for source files
Found 1 source file
.../test_import.py
.../test_import.py:5:13 - error: "__version__" is not exported from module "sympy" (reportPrivateImportUsage)
1 error, 0 warnings, 0 informations 
Completed in 0.616sec
```

This can also be reproduced on current master by installing (`python setup.py install`) and running `pyright test_import.py`.
Numpy doesn't has this problem as it explicitly exports `__version__`:
https://github.com/numpy/numpy/blob/65a701fd40d5ab3b772131daf45679d6ecf3d721/numpy/__init__.py#L227
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/__init__.py
+++ b/sympy/__init__.py
@@ -254,6 +254,8 @@ from .interactive import init_session, init_printing, interactive_traversal
 evalf._create_evalf_table()
 
 __all__ = [
+    '__version__',
+
     # sympy.core
     'sympify', 'SympifyError', 'cacheit', 'Basic', 'Atom',
     'preorder_traversal', 'S', 'Expr', 'AtomicExpr', 'UnevaluatedExpr',


eagleoflqj:
Maybe we could have a pyright label instead of uncategorised?

PR #23345, opened 2022-04-10
add type hints to base and exp of Pow

eagleoflqj:
test_exp.py:
```py
from sympy import Symbol, Pow
x = Symbol('x')
p = Pow(x, 2)
print(p.base + p.exp)
```
```sh
$ pyright test_exp.py 
No configuration file found.
No pyproject.toml file found.
stubPath .../typings is not a valid directory.
Assuming Python platform Linux
Searching for source files
Found 1 source file
.../test_exp.py
  .../test_exp.py:4:7 - error: Operator "+" not supported for types "Basic" and "Basic"
    Operator "+" not supported for types "Basic" and "Basic" when expected type is "object" (reportGeneralTypeIssues)
1 error, 0 warnings, 0 informations 
Completed in 0.991sec
```
Solely adding `-> Expr` makes pyright extension for vscode complain, as `_args` is declared `tuple[Basic, ...]`.
According to https://github.com/sympy/sympy/blob/9c7ebddbb7be78267f0162c75e2e66ff5b2ff5a3/sympy/core/basic.py#L765-L773
replacing `_args` makes sense.
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/power.py
+++ b/sympy/core/power.py
@@ -1,4 +1,5 @@
-from typing import Callable, Tuple as tTuple
+from __future__ import annotations
+from typing import Callable
 from math import log as _log, sqrt as _sqrt
 from itertools import product
 
@@ -275,7 +276,8 @@ class Pow(Expr):
 
     __slots__ = ('is_commutative',)
 
-    args: tTuple[Expr, Expr]
+    args: tuple[Expr, Expr]
+    _args: tuple[Expr, Expr]
 
     @cacheit
     def __new__(cls, b, e, evaluate=None):
@@ -383,11 +385,11 @@ class Pow(Expr):
         return None
 
     @property
-    def base(self):
+    def base(self) -> Expr:
         return self._args[0]
 
     @property
-    def exp(self):
+    def exp(self) -> Expr:
         return self._args[1]
 
     @property


sympy/core/power.py
```diff
@@ -383,12 +383,12 @@ def inverse(self, argindex=1):
         return None
 
     @property
-    def base(self):
-        return self._args[0]
+    def base(self) -> Expr:
+        return self.args[0]
```
oscarbenjamin, line 387: Maybe it would be better to declare that the args of `Pow` are `Expr`.

eagleoflqj: Yes, it is already there
https://github.com/sympy/sympy/blob/641a8b2839f0756808fb4cf941b82955f527a231/sympy/core/power.py#L278

oscarbenjamin: So why does pyright think that it's Basic?

eagleoflqj: If I only add return type hint and keep function body as `return _args[0]` then pyright will complain, as it doesn’t know `_args` and `args` is the same thing.
Now it doesn’t complain.

oscarbenjamin: Maybe there should also be a type-hint for `Pow._args` then.

eagleoflqj: _args isn’t used elsewhere in this file, and its usage (for both developer and user) is not encouraged either. It’s more like an internal representation of args for Basic and not its subclasses.

ThePauliPrinciple: What if instead we define the type in `def __new__(cls, b, e, evaluate):`? Shouldn't it be able to figure that the values in `_arg` are actually `Expr` instead of `Basic`?

eagleoflqj: > What if instead we define the type in `def __new__(cls, b, e, evaluate):`? Shouldn't it be able to figure that the values in `_arg` are actually `Expr` instead of `Basic`?

This `__new__` function doesn't always return a `Pow` object, like `Pow(1,1)` returns `Integer(1)`. And it delegates to `Expr.__new__` for creating `Pow` object (and after that we still don't know if it's a `Pow`).
https://github.com/sympy/sympy/blob/641a8b2839f0756808fb4cf941b82955f527a231/sympy/core/power.py#L372-L377
Given the complexity, I don't think pyright can track all the way down to know `_arg` is `Expr`.

oscarbenjamin: It looks to me like `_args` is being used rather than `args` because it is maybe slightly faster. I'm not sure we should change that.

eagleoflqj: The usage of `_args` here dates back to the first git commit of sympy in 2007. Later in 2008, @certik made the note in https://github.com/sympy/sympy/commit/b89a24efab8e4e25aed2ecb567905a1720b4bb62 that `args` should always be used. Could @certik or @asmeurer please give some advices?

oscarbenjamin: I just don't think it's necessary to change this. As far as I can tell all that is needed is a type-hint on `_args` to clarify that `Pow._args` is `tuple[Expr, Expr]`.

eagleoflqj:
@oscarbenjamin this should be good to go.

oscarbenjamin:
Looks good

PR #23346, opened 2022-04-10
Added More Distribution in Sampling with NumPy

ghost:
More distributions in Sampling(in stats module)

More distributions to Sampling Module can be included and so I tried to include as much as possible in Sampling with Numpy.

The following has been registered:
* *FDistribution Distribution*
* *Gumbel Distribution*
* *Laplace Distribution*
* *Logistic Distribution*
* *Rayleigh Distribution*
* *Triangular Distribution*
* *HyperGeometric Distribution*

#### Other comments
- [x] Opened on a seperate branch
- [x] Has unit tests
- [x] Follows Documentation Guidelines
- [x] Has Release Notes

#### Release Notes

* stats
  * Registered more distributions that can be sampled using numpy.

### DIFF
--- a/sympy/stats/sampling/sample_numpy.py
+++ b/sympy/stats/sampling/sample_numpy.py
@@ -34,35 +34,35 @@ def _(dist: ExponentialDistribution, size, rand_state):
 
 @do_sample_numpy.register(FDistributionDistribution)
 def _(dist: FDistributionDistribution, size, rand_state):
-    return rand_state.f(dfn=float(dist.dfn), dfd=float(dist.dfd), size=size)
+    return rand_state.f(dfnum = float(dist.d1), dfden = float(dist.d2), size=size)
 
 @do_sample_numpy.register(GammaDistribution)
 def _(dist: GammaDistribution, size, rand_state):
-    return rand_state.gamma(float(dist.k), float(dist.theta), size=size)
+    return rand_state.gamma(shape = float(dist.k), scale = float(dist.theta), size=size)
 
 @do_sample_numpy.register(GumbelDistribution)
 def _(dist: GumbelDistribution, size, rand_state):
-    return rand_state.gumbel(float(dist.mu), float(dist.beta), size=size)
+    return rand_state.gumbel(loc = float(dist.mu), scale = float(dist.beta), size=size)
 
 @do_sample_numpy.register(LaplaceDistribution)
 def _(dist: LaplaceDistribution, size, rand_state):
-    return rand_state.laplace(float(dist.mu), float(dist.beta), size=size)
+    return rand_state.laplace(loc = float(dist.mu), scale = float(dist.b), size=size)
 
 @do_sample_numpy.register(LogisticDistribution)
 def _(dist: LogisticDistribution, size, rand_state):
-    return rand_state.logistic(float(dist.mu), float(dist.sigma), size=size)
+    return rand_state.logistic(loc = float(dist.mu), scale = float(dist.s), size=size)
 
 @do_sample_numpy.register(LogNormalDistribution)
 def _(dist: LogNormalDistribution, size, rand_state):
-    return rand_state.lognormal(float(dist.mean), float(dist.std), size=size)
+    return rand_state.lognormal(mean = float(dist.mean), sigma = float(dist.std), size=size)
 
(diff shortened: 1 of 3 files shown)


KuldeepBorkar:
Could someone help me with the following please:

1) Like consider the file in which I made changes here i.e.  *`stats\sampling\sample_numpy.py`*,
 are these 2 lines the same?
*`return rand_state.function_name`* and *`return numpy.random.function_name`*

Also I was not able to understood what *`@singledispatch`* and *`rand_state`* are actually doing there (If someone could tell that would be really helpful for me.)

2) We have to register here those distributions only which are already implemented in `SymPy` right?

3) Should I include Release Notes in this way or `NO ENTRY` should be okay?
     * stats
           Registered some more distributions in sampling with numpy
4) Is it okay to register some more distributions in Sampling with PyMC3 and SciPy if I found any or should I open another `PR` for them?

KuldeepBorkar:
For anyone who don't know about @singledispatch maybe the below link might help.
(https://peps.python.org/pep-0443/)

I tried to learn more about the @singledispatch decorator in python.
It is used to decorate a generic function and the dispatch happens on the type of first argument.

Let's consider an example:

```
from functools import singledispatch

@singledispatch
def learning(arg):
  print(arg)

@learning.register(int)
def _(arg):
  print("Dispatched for int as first argument")

@learning.register(list)
def _(arg):
  print("Dispatched for list as first argument")

a = [1,2,3]
b = 10
c =  "For string nothing was registered"

print(learning(a))
print(learning(b))
print(learning(c))
```

And we will get Outputs:

```
Dispatched for list as first argument
Dispatched for int as first argument
For string nothing was registered
```

czgdp1807:
This is looking good to me. I think this is an extension of already existing sampling framework by adding new distributions to it. @Smit-create @Upabjojr What are your thoughts?

czgdp1807 (approved):

Upabjojr:
It looks good to me as well. Did you do some numerical checks to verify that the sampled distributions are exactly the same as expected?

KuldeepBorkar:
**Example:**
> 
> ```
> from sympy.stats import Gumbel, sample_iter
> X = Gumbel('X', 1, 2)
> expr = X*X + 3
> iterator = sample_iter(expr, library='numpy' ,numsamples=3) 
> list(iterator) 
> ```
> 
> _The above code will result in error NotImplemented but once this PR gets merged it would be possible to do the above thing._

@Upabjojr I don't think so, I just used the above example to try sampling and it worked perfectly and then I just tried sampling using `NumPy` and then in `SymPy` and the results were almost same as expected(So, I thought it would be okay).

I was planning to make a same `PR` to register more distributions using `SciPy` and `PyMC3`, what kind of numerical checks should I do to confirm the sampled distributions are exactly as expected?

I tried to confirm things in this way:

```
from sympy.stats import sample
from sympy.stats import Gumbel
X = Gumbel('X', 1, 2)
sample(X) #or a iterator as used in above example and check say some 10 or 100 samples or plot them.
```
and then with NumPy
```
import numpy as np
mu, beta = 1, 2
s = np.random.gumbel(mu, beta, 1)
s
```

Although I was not able to see any negative sample while sampling in `SymPy` but was able to see in `NumPy`
Do you think that's a problem because I think it is?

Smit-create (approved):
LGTM, thanks!!

Upabjojr:
> what kind of numerical checks should I do to confirm the sampled distributions are exactly as expected?

Sometimes different libraries use slightly different conventions for the parameters of the distributions. It's usually not the case, but it's better to check.

For example, using NumPy it's possible to generate lots of data, and then check that some quantiles (or other statistics) of the sampled data are the same as expected for a given set of parameters.

Maybe we should add some function to the tests that performs these checks. But this can be done in the future, this PR is OK right now.

Upabjojr:
Anyway, LGTM

KuldeepBorkar:
> Maybe we should add some function to the tests that performs these checks. But this can be done in the future, this PR is OK right now.

I will open an Issue for this so that this function would be added sooner or later.

KuldeepBorkar:
@Upabjojr , @czgdp1807 , @Smit-create  I created a similar Pull Request #23422 to register more distributions to be sampled using SciPy but I am confused a bit there. It would really be helpful if you could please take a look at that PR too.

Issue #21884, opened 2021-08-16
Adopt NEP 29 - Python support


oscarbenjamin:
Should SymPY adopt NEP 29?
https://numpy.org/neps/nep-0029-deprecation_policy.html

It prescribes a policy for how long to support different Python versions (and numpy versions) for the scientific Python ecosystem. The relevant part for SymPy is just how long to support a given version of Python.

Adopting NEP 29 would mean dropping support for Python 3.6 in the next SymPy release (it could be removed from CI right now).

The main advantage in adopting NEP 29 as I see it is just that it means there is a clearly written support policy. If that policy is problematic then there can be a broader discussion than just SymPy but otherwise I think it makes sense for SymPy to synchronise with related projects.

asmeurer:
We do have a policy https://github.com/sympy/sympy/wiki/Python-version-support-policy

Our current policy is based on when Python itself drops support, whereas NEP 29 drops support sooner, meaning it supports fewer versions. I think it's a good idea to do this, because the increased CPython release cadence that was implemented in the past few years means that there are too many versions of Python to support. Personally I don't like supporting more than 3 versions of CPython at once. I think with NEP 29 we would actually drop Python 3.7 very soon too (by the end of the year). Note that 3.10 will be released in October.

oscarbenjamin:
Right, there is a policy. The difference with NEP 29 is to drop support for older Python versions before they reach EOL but in a predictable way. I'm not sure it really matters that much to anyone downstream but the noticeable differences to me are just that we don't need to test so many versions in CI and we can also use relatively recent language features.

moorepants:
I think this should be based on time, not Python versions. We should support the past 1 or 2 years of Python, something like that.

oscarbenjamin:
NEP 29 is based primarily on a time window:
```
When a project releases a new major or minor version, we recommend
that they support at least all minor versions of Python introduced and
released in the prior 42 months from the anticipated release date with
a minimum of 2 minor versions of Python
```

moorepants:
That looks good to me.

oscarbenjamin:
I've opened #21886 to drop support for Python 3.6 and PyPy 3.6 immediately.

If NEP 29 is adopted then the support policy will need to be updated:
https://github.com/sympy/sympy/wiki/Python-version-support-policy

asmeurer:
We should probably discuss this on the mailing list. A change in this policy would potentially affects users of SymPy.

[...]

PR #23347, opened 2022-04-11
Update sage ci to use python 3.8

eagleoflqj:

Related to #21884

Re-pick the first commit of #21941 to run on ubuntu groovy instead of debian buster.

#### Release Notes

NO ENTRY

### DIFF
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -91,9 +91,9 @@ jobs:
       fail-fast: false
       max-parallel: 32
       matrix:
-        # debian-buster uses system python 3.7
+        # ubuntu-focal uses system python 3.8
         # archlinux-latest is at the cutting edge
-        tox_system_factor: [debian-buster, archlinux-latest]
+        tox_system_factor: [ubuntu-focal, archlinux-latest]
         # "standard" installs lots of system packages, reducing the full build and test of the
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
I'm not really sure why this is needed.

eagleoflqj:
> I'm not really sure why this is needed.

Originally at https://github.com/sympy/sympy/pull/21941/commits/1e835fb708f1d8fa7fde04311f53ce0c07d8ef0c all supported python versions of sympy were tested, which was then considered unnecessary, so finally only the oldest and newest python version are tested. That's why we have debian-buster and arch-latest.
As we decide to drop python 3.7 support, 3.8 becomes the oldest version we support, this should be reflected on sage ci.

eagleoflqj:
@mkoeppe What do you think?

mkoeppe:
`ubuntu-groovy` is past EOL (https://fridge.ubuntu.com/2021/07/25/ubuntu-20-10-groovy-gorilla-end-of-life-reached-on-july-22-2021/). I'd suggest to use `ubuntu-focal`, which is an LTS and also has 3.8 as the default python3.

eagleoflqj:
I think @mkoeppe approves this change. Could we move on dropping python 3.7? @oscarbenjamin

oscarbenjamin:
Looks good. Thanks both.

PR #23352, opened 2022-04-11
Add hyperbola class

oleander-twig:
Implemented hyperbola class

Attributes: hradius, vradius, center, focus_distance, foci, eccentricity
The Hyperbola class can be used as an instrument to calculate asymptotes (their equations), to draw a function, rotate and scale it. All attributes are recalculated automatically after changes (rotation, scaling). The class is also writes 2 types of hyperbola equation with any symbols you choose and moves function to a new center point. 

* class
 * functions
  * move() - recalculates the hyperbola equation with a center point you choose 
  * equation() - prints 2 types of hyperbola equation with any symbols you choose; recalculate equations after rotation, scaling, moving 
  * asymptote() - prints the equations of asymptotes of hyperbola function 
  * border_point() - returns bool statement if point is in the hyperbola area 
  * random_point - returns a random point from hyperbola area 
  * scale() - scaling the function 
  * draw() - draws plot with asymptotes

### DIFF
--- a/sympy/geometry/hyperbola.py
+++ b/sympy/geometry/hyperbola.py
@@ -1,0 +1,124 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# In[ ]:
+
+
+import sympy
+import random
+import math
+import decimal
+import numpy as np
+import matplotlib.pyplot as plt
+from sympy import Symbol, symbols, plot_implicit
+from sympy import Rational
+from sympy.geometry import Point
+from sympy.core.symbol import _symbol
+
+
+class Hyperbola:
+    def __init__(self, hradius, vradius, angle=None, centre=None):
+        self.hradius = hradius
+        self.vradius = vradius
+        self.angle = 0
+        self._centre = Point(0,0)
+        if angle:
+            self.angle = angle
+        if centre:
+            self._centre = centre
+    
+    @property
+    def focus_distance(self):
+        return (sympy.sqrt(self.hradius ** 2 + self.vradius ** 2))
+    
+    @property
+    def foci(self):
+        return Point(self.focus_distance, 0), Point(- self.focus_distance, 0)
+        
+    @property
+    def eccentricity(self):
+        return self.focus_distance / self.hradius
+    
+    @property
+    def centre(self):
+        return self._centre
+    
+    def move(self, x1, y1):
+        self._centre = Point(x1, y1)
+    
+    def equation(self, x=None, y=None, expression=False):
+        if x:
+            x = Symbol(x)
+        else:
+            x = symbols('x')
+        if y:
+            y = Symbol(y)
+        else:
+            y = symbols('y')
+        x = x - self._centre[0]
+        y = y - self._centre[1]
+        x1 = x * sympy.cos(self.angle) + y * sympy.sin(self.angle)
+        y = - x * sympy.sin(self.angle) + y * sympy.cos(self.angle)
+        x = x1
+        if not expression:
+            t1 = (x * self.vradius) ** 2
+            t2 = (y * self.hradius) ** 2
+            return (t1 - t2 - (self.hradius * self.vradius) ** 2).simplify()
+        t1 = (x / self.hradius) ** 2
+        t2 = (y / self.vradius) ** 2
+        return sympy.Eq((t1 - t2).simplify(), 1)
+    
+    def asymptote(self, x=None, y=None, expression=False):
+        if x:
+            x = Symbol(x)
+        else:
+            x = symbols('x')
+        if y:
+            y = Symbol(y)
+        else:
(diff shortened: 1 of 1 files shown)


faze-geek:
Hi @oleander-twig 
I believe you are new to SymPy ,so maybe you need to get familiar with how a pull request OP and release notes are framed .

Thanks for the contribution but I think it is always better to check if the things you are implementing have been attempted earlier . Currently we already have 2 pr's #17822 and #22698 ( which is a recent modified version ) with about 2000 lines of code diff .I guess they have also gone through an initial set of reviews and have tests to support the functionalities they have added .

sympy/geometry/hyperbola.py
```diff
@@ -0,0 +1,124 @@
...
+
+# In[ ]:
+
+
+import sympy
+import random
+import math
+import decimal
+import numpy as np
+import matplotlib.pyplot as plt
```
oscargus, line 12: Neither numpy not matplotlib are dependencies of SymPy, so this must be handled with differently.

sympy/geometry/hyperbola.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# In[ ]:
+
+
+import sympy
+import random
```
oscargus, line 8: Use the SymPy random module instead.

sympy/geometry/hyperbola.py
oscargus: Are all these imports even used?

sympy/geometry/hyperbola.py
```diff
@@ -0,0 +1,124 @@
...
+import math
+import decimal
+import numpy as np
+import matplotlib.pyplot as plt
+from sympy import Symbol, symbols, plot_implicit
+from sympy import Rational
```
oscargus, line 14: Please import using fully qualified names as this will help us avoid cyclic imports in the long run.

sympy/geometry/hyperbola.py
```diff
@@ -0,0 +1,124 @@
...
+from sympy.geometry import Point
+from sympy.core.symbol import _symbol
+
+
+class Hyperbola:
```
oscargus, line 19: For example, Parabola inherits from GeomerySet. This probably should do that as well to be able to operated correctly.

sympy/geometry/hyperbola.py
```diff
@@ -0,0 +1,124 @@
...
+            x = Symbol(x)
+        else:
+            x = symbols('x')
+        if y:
+            y = Symbol(y)
+        else:
+            y = symbols('y')
+        x = x - self._centre[0]
+        y = y - self._centre[1]
+        x1 = x * sympy.cos(self.angle) + y * sympy.sin(self.angle)
```
oscargus, line 60: Import cos and sin from sympy instead.

sympy/geometry/hyperbola.py
```diff
@@ -0,0 +1,124 @@
...
+    
+    @property
+    def centre(self):
+        return self._centre
+    
+    def move(self, x1, y1):
+        self._centre = Point(x1, y1)
+    
+    def equation(self, x=None, y=None, expression=False):
+        if x:
```
oscargus, line 50: You probably want to check if x is a string here as it could make sense to allow it to be a Symbol.

sympy/geometry/hyperbola.py
```diff
@@ -0,0 +1,124 @@
...
+            x = Symbol(x)
+        else:
+            x = symbols('x')
```
oscargus, line 53: symbols('x') and Symbol('x') will do exactly the same in this case.

sympy/geometry/hyperbola.py
```diff
@@ -0,0 +1,124 @@
...
+            y = symbols('y')
+        x = x - self._centre[0]
+        y = y - self._centre[1]
+        x1 = x * sympy.cos(self.angle) + y * sympy.sin(self.angle)
+        y = - x * sympy.sin(self.angle) + y * sympy.cos(self.angle)
+        x = x1
+        if not expression:
+            t1 = (x * self.vradius) ** 2
+            t2 = (y * self.hradius) ** 2
+            return (t1 - t2 - (self.hradius * self.vradius) ** 2).simplify()
```
oscargus, line 66: Avoid calling `simplify` in library code.

eagleoflqj:
Thanks for your contribution but I think #22698 is more close to finish. Closing.

anutosh491:
> Thanks for your contribution but I think #22698 is more close to finish. Closing.

I haven't worked on this one since some and plan to finish it as soon as my exams at university are over . Thanks for pointing this out though, even I feel it is close to finish . Let me know if you see improvements in the original pr by commenting there.

PR #12322, opened 2017-03-16
v2pt_theory: Added zero velocity constraint requirement.

angadhn:
I'm unsure if what I am adding here, which is a trivial detail, was
intentionally left out; here is the description of what I have done:
P and O are two points of a rigid body B that is moving in frame N.
The velocity of P in N, N^v^P, can be computed using v2pt_theory,
which it does admirably. However, the function should also set
B^v^P = 0; that is what I have added. In fact, the function should
likely also constrain the B^v^O to be 0; that is what makes P and O
2 points of the same body.
Further, this constraint of B^v^O should also likely be added to v1pt
theory. If you agree, I can commit changes to that next time, as well.

<!-- If this pull request fixes an issue please indicate which issue by typing "Fixes #NNNN" below.>

PR #23359, opened 2022-04-13
Add zero velocity constraint in v1pt and v2pt

tjstienstra:

Fixes the problem presented in pull request [12322](https://github.com/sympy/sympy/pull/12322).

Add a check if the velocity of the `otherpoint` in the velocity theorems is indeed set to `0`

#### Other comments
Possibly also add a check for the acceleration theorems (i.e. `a1pt_theory` and `a2pt_theory`).

#### Release Notes

* physics.vector
  * Add a zero velocity requirement to `v1pt_theory` and `v2pt_theory`

### DIFF
--- a/sympy/physics/vector/point.py
+++ b/sympy/physics/vector/point.py
@@ -68,6 +68,23 @@ class Point:
         if not isinstance(other, Point):
             raise TypeError('A Point must be supplied')
 
+    def _check_not_moving(self, frame):
+        """Checks whether the point is indeed not moving in a frame."""
+        point_valid = True
+        try:
+            if not self.vel(frame) == 0:
+                point_valid = False
+        except ValueError:
+            pass
+        try:
+            if not self.acc(frame) == 0:
+                point_valid = False
+        except (ValueError, KeyError):
+            pass
+        if not point_valid:
+            raise ValueError('Point ' + self.name + ' is not fixed in ' +
+                             'ReferenceFrame ' + frame.name)
+
     def _pdict_list(self, other, num):
         """Returns a list of points that gives the shortest path with respect
         to position, velocity, or acceleration from this point to the provided
@@ -151,7 +168,7 @@ class Point:
         >>> B = ReferenceFrame('B')
         >>> B.set_ang_vel(N, 5 * B.y)
         >>> O = Point('O')
-        >>> P = O.locatenew('P', q * B.x)
+        >>> P = O.locatenew('P', q * B.x + q2 * B.y)
         >>> P.set_vel(B, qd * B.x + q2d * B.y)
         >>> O.set_vel(N, 0)
         >>> P.a1pt_theory(O, N, B)
@@ -162,6 +179,7 @@ class Point:
         _check_frame(outframe)
         _check_frame(interframe)
         self._check_point(otherpoint)
+        otherpoint._check_not_moving(interframe)
         dist = self.pos_from(otherpoint)
         v = self.vel(interframe)
         a1 = otherpoint.acc(outframe)
@@ -213,6 +231,7 @@ class Point:
         _check_frame(outframe)
         _check_frame(fixedframe)
         self._check_point(otherpoint)
+        otherpoint._check_not_moving(fixedframe)
         dist = self.pos_from(otherpoint)
         a = otherpoint.acc(outframe)
         omega = fixedframe.ang_vel_in(outframe)
@@ -433,7 +452,7 @@ class Point:
         >>> B = ReferenceFrame('B')
         >>> B.set_ang_vel(N, 5 * B.y)
         >>> O = Point('O')
-        >>> P = O.locatenew('P', q * B.x)
+        >>> P = O.locatenew('P', q * B.x + q2 * B.y)
         >>> P.set_vel(B, qd * B.x + q2d * B.y)
         >>> O.set_vel(N, 0)
         >>> P.v1pt_theory(O, N, B)
@@ -444,6 +463,7 @@ class Point:
         _check_frame(outframe)
         _check_frame(interframe)
         self._check_point(otherpoint)
+        otherpoint._check_not_moving(interframe)
         dist = self.pos_from(otherpoint)
         v1 = self.vel(interframe)
         v2 = otherpoint.vel(outframe)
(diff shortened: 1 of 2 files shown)


sympy/physics/mechanics/models.py
```diff
@@ -197,6 +197,7 @@ def n_link_pendulum_on_cart(n=1, cart_force=True, joint_torques=False):
         frames.append(Bi)
 
         Pi = points[-1].locatenew('P{}'.format(i + 1), l[i] * Bi.y)
+        points[-1].set_vel(Bi, 0)
```
moorepants, line 200: Seeing that we have to add this now, I'm not sure this is the solution we want. I thought that we only want to through an error if the velocity of the fixed point is such that it has a value and that value indicates it isn't fixed. So we should allow the velocity to 1) be undefined or 2) fixed in the shared frame. Having to set this velocity to zero is just going to be annoying for users.

sympy/physics/vector/point.py
```diff
@@ -444,6 +444,9 @@ def v1pt_theory(self, otherpoint, outframe, interframe):
         _check_frame(outframe)
         _check_frame(interframe)
         self._check_point(otherpoint)
+        if not otherpoint.vel(interframe) == 0:
```
moorepants, line 447: Given my comment above, this should be something like "if otherpoint velocity is defined and it isn't zero in the shared frame, then ...."

TJStienstra: Yes, I agree. We could at most display a warning. The best option I see is to use `try` `except` to catch the `ValueError`, when the velocity is not defined. This will result in something like:
```py
other_point_is_valid = True
try:
    if not otherpoint.vel(interframe)==0:
        otherpoint_is_valid = False
except ValueError:
    pass
if not other_point_is_valid:
    raise ValueError(...)
```
Advantage is that this is a extensive test, which really checks whether the velocity is defined. Disadvantages: it uses `try` `except` and will possibly add `{interframe: 0}` to the `otherpoint._vel_dict' by calling `otherpoint.vel` (not sure if this is a disadvantage).

Shall I implement this solution? Or is there a better one? Also should I revert my previous commit, so the test files and `model.py` will not be changed?

moorepants: It is fine to do explicit checks over the try/except clause. Maybe preferred in this case.

We try to avoid warnings in SymPy, but we already have some for this particular piece of code. I think it is fine to throw an error if the velocity is defined and it is not constant. If it is not defined, the behavior should be as it is, don't do anything.

TJStienstra: Should I also undo my previous changes to the other files?

moorepants: Yes, unfortunately. I don't think we want to add this requirement that you have to set the velocity of the base point in the shared frame. It is really unnecessary. The alternative is, as @angadhn suggested, to set that velocity automatically for the user. It's probably ok to do that. So, if the velocity is undefined, set it to 0 and if it is defined and not zero, then throw an error. The only thing I'm hung up on that is that we don't set more than one velocity with any other methods. But I may be warmed up to the idea by now. If it throws an error if it is not constant and sets it to zero otherwise, I can't see a major issue, but then we wouldn't return that second velocity.  But that method really shouldn't return a velocity, it should just set them. If you set the velocity to zero and make that very clear in the documentation, then that's probably fine.

TJStienstra: I will just revert my changes and implement the [suggestion](https://github.com/sympy/sympy/pull/23359#discussion_r850122669). For this one the velocity in that frame will only be set to `0` if it is already known that it is `0`.

TJStienstra:
Just cleaned up a bit and changed the implementation as well. However I found out that there is one big problem, namely that it will fail on cases where the velocity term contains a kde. This happens in the last example [here](https://docs.sympy.org/latest/modules/physics/vector/kinematics.html?highlight=particle%20moving%20ring), which can be replicated with the following code:
```py
from sympy.physics.vector import dynamicsymbols, Point, ReferenceFrame
from sympy import Symbol, cos, sin

N = ReferenceFrame('N')
u1, u2 = dynamicsymbols('u1 u2')
q1, q2 = dynamicsymbols('q1 q2')
l = Symbol('l')
R = Symbol('R')
C = N.orientnew('C', 'Axis', [q1, N.x])
C.set_ang_vel(N, u1 * N.x)
O = Point('O')
O.set_vel(N, 0)
Q = O.locatenew('Q', -l * C.z)
P = Q.locatenew('P', R * (cos(q2) * C.x + sin(q2) * C.y))
P.set_vel(C, R * u2 * (-sin(q2) * C.x + cos(q2) * C.y))
print(O.vel(C))
Q.v2pt_theory(O, N, C)
```

TJStienstra:
> An option would be to check that there are no derivative terms in the velocity

Another problem could be given if someone manages to define something double with different variable.
Another options would be replacing it with a warning.

TJStienstra:
@moorepants or @sidhu1012 what shall we do with this PR:
- Just let it be, not adding a zero velocity constraint, only edit the small bugs in the tests that were found
- Check that there are no derivative terms in the checks: `self.vel(frame) == 0` and `self.acc(frame) == 0`
- Replace `raise ValueError` with a warning

moorepants:
Hi @tjstienstra is there a reason to close this? It seems to have been forgotten, but looks like there are some aspects that may be desired. Or maybe its been addressed in other work?

sympy/physics/vector/tests/test_point.py
```diff
@@ -76,8 +96,8 @@ def test_point_funcs():
     B = ReferenceFrame('B')
     B.set_ang_vel(N, 5 * B.y)
     O = Point('O')
-    P = O.locatenew('P', q * B.x)
-    assert P.pos_from(O) == q * B.x
+    P = O.locatenew('P', q * B.x + q2 * B.y)
+    assert P.pos_from(O) == q * B.x + q2 * B.y
```
moorepants, line 100: Was this an error in the example?

tjstienstra: Yes, the next line sets the velocity as `P.set_vel(B, qd * B.x + q2d * B.y)`

PR #23365, opened 2022-04-14
Fix latex

georgesk:
#### Changes
Mardown and RST files were modified to generate the code $\textrm{\LaTeX}$, which is the right way to generate text in math dollar-environment with latex and the package amsmath.

#### Other comments
Those changes are part of a patch which I am using to maintain the Debian package, see https://tracker.debian.org/pkg/sympy 
 
NO ENTRY

### DIFF
--- a/doc/src/guides/contributing/dependencies.md
+++ b/doc/src/guides/contributing/dependencies.md
@@ -109,7 +109,7 @@ images rendered with LaTeX. `preview()` can either save the image to a file or
 show it with a viewer.
 
 (dependencies-latex)=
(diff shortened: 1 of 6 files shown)


asmeurer:
This has made it so that `\LaTeX` shows up unrendered in the HTML docs using MathJAX. https://docs.sympy.org/dev/tutorial/printing.html#textrm-latex

asmeurer:
It looks like this change was just wrong. I checked and the previous `\mathrm{\LaTeX}` renders correctly in both the HTML and PDF. @georgesk can you clarify what the issue you were having here was?

Issue #22994, opened 2022-02-02
Lobb and Delannoy Combinatorial Numbers


praneethratna:
I would like to add the Lobb Number and Delannoy Number to the combinatorial module. More on them here -
[Lobb Number](https://en.wikipedia.org/wiki/Lobb_number)
[Delannoy Number](https://en.wikipedia.org/wiki/Delannoy_number)

EveningShower:
Are you working on this right now? If not I wanna work on it so I can submit my application to GSoC22.

asmeurer:
Both sounds like they are legitimate thing that can be added.

HESleagacy:
It seems that Delannoy Number is still pending to be added, if i may add it ?

PR #23370, opened 2022-04-15
Added Lobb numbers to sympy/sympy/functions/combinatorial/numbers.py

EveningShower:

Implemented the first part from #22994

I added Lobb numbers to the combinatorics numbers, Lobb numbers counts the number of ways that `n + m` open parentheses 
    and `n - m` close parentheses can be arranged to form the start of a valid sequence 
    of balanced parentheses, they are closely related to Catalan numbers, in fact, `lobb(0, n) = catalan(n)`.

#### Other comments
I'm doing this so I can submit my application to GSoC2022, so I hope it can be merged as soon as possible, any feedback would be much appreciated.
Also I can't fix the issue with authors, I followed every manual there is but I can't fix the problem, because it involves another user, it shows me the following error:
`This author is not included in the .mailmap file:`
`Georges Khaznadar <georgesk@debian.org>`
#### Release Notes

* functions
  * Added Lobb numbers to combinatorial numbers.

### DIFF
--- a/sympy/functions/combinatorial/numbers.py
+++ b/sympy/functions/combinatorial/numbers.py
@@ -1408,6 +1408,52 @@ class partition(Function):
         if n.is_nonnegative and n.is_integer:
             return True
 
+class lobb(Function):
+    r"""
+    The Lobb number `L_{m,n}` counts the number of ways that `n + m` open parentheses
+    and `n - m` close parentheses can be arranged to form the start of a valid sequence
+    of balanced parentheses.
+
+    Lobb numbers form a natural generalization of the Catalan numbers,
+    which count the number of complete  strings of balanced parentheses of a given length.
+    Thus, the `n^{th}` Catalan number equals the Lobb number `L_{0,n}`.
+
+    The `(m, n)^{th}` Lobb number `L_{m,n}` is given by the formula:
+
+    .. math:: L_{m,n} = \frac{2m+1}{m+n+1}\binom{2n}{m+n}
+
+    Examples
+    ========
+    >>> from sympy.functions.combinatorial.numbers import lobb
+    >>> for i in range(0,6):
+    ...     if i == 0: print("1")
+    ...     else:
+    ...         for j in range(i):
+    ...             print(lobb(j, i), end=" ")
+    ...         print("1")
+    1
+    1 1
+    2 3 1
+    5 9 5 1
+    14 28 20 7 1
+    42 90 75 35 9 1
+
+    See Also
+    ========
+    bell, bernoulli, catalan, euler, fibonacci, harmonic, lucas, partition, tribonacci,
+    genocchi
+    References
+    ==========
+    .. [1] https://en.wikipedia.org/wiki/Lobb_number
+    """
+
+
+    @classmethod
+    def eval(cls, m, n):
+
+        if n.is_Integer and n.is_nonnegative and m.is_Integer and m.is_nonnegative:
+            return binomial(2*n, m + n)*(2*m + 1)/(m + n + 1)
+
 
 #######################################################################
 ###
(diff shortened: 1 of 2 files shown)


anutosh491:
I am not sure if just the eval method of Lobb Number class should be merged , it is incomplete and doesn't make complete sense regarding why it is a good addition  . maybe something like #22831 would look better !

EveningShower:
> I am not sure if just the eval method of Lobb Number class should be merged , it is incomplete and doesn't make complete sense regarding why it is a good addition . maybe something like #22831 would look better !

There is not much about Lobb numbers even in the wikipedia article, that's why the eval method is the only one I could add.
Can you please explain to me what do you mean by referencing #22831, would it be better if I implemented all the numbers mentionned there? Because I was willing to do that but I was in a hurry to finish my GSoC proposal, if this can be merged, I will get to that as soon as I finish up my proposal.

oscarbenjamin:
> There is not much about Lobb numbers even in the wikipedia article

This suggests that they might not be worth including in sympy.

> I hope it can be merged as soon as possible

If you want to get something merged quickly then I suggest addressing an already open issue rather than trying to add something new.

anutosh491:
> There is not much about Lobb numbers even in the wikipedia article, that's why the eval method is the only one I could add. Can you please explain to me what do you mean by referencing #22831, would it be better if I implemented all the numbers mentionned there? Because I was willing to do that but I was in a hurry to finish my GSoC proposal, if this can be merged, I will get to that as soon as I finish up my proposal.

I don't see this getting merged quickly , because reviewers even those who review code for combinatorics might not know about the Lobb's number , hence they'll have to study the math and everything behind that first ! I suggest taking up something more straightforward if you are short of time as it seems!

EveningShower:
> > I hope it can be merged as soon as possible
> 
> If you want to get something merged quickly then I suggest addressing an already open issue rather than trying to add something new.

Can you please suggest me something of that sort? I went through the open issues multiple times but nothing really seems doable or that it could get merged quickly

EveningShower:
> > There is not much about Lobb numbers even in the wikipedia article
> 
> This suggests that they might not be worth including in sympy.
But it's the same for other combinatorial numbers that are already included, like Genocchi numbers.

sympy/functions/combinatorial/numbers.py
```diff
@@ -1408,6 +1408,52 @@ def _eval_is_positive(self):
...
+    and `n - m` close parentheses can be arranged to form the start of a valid sequence
+    of balanced parentheses.
+
+    Lobb numbers form a natural generalization of the Catalan numbers,
+    which count the number of complete  strings of balanced parentheses of a given length.
+    Thus, the `n^{th}` Catalan number equals the Lobb number `L_{0,n}`.
+
+    The `(m, n)^{th}` Lobb number `L_{m,n}` is given by the formula:
+
+    .. math:: L_{m,n} = \frac{2m+1}{m+n+1}\binom{2n}{m+n}
```
asmeurer, line 1423: Do not copy text directly from [Wikipedia](https://en.wikipedia.org/wiki/Lobb_number). This is effectively plagiarism, since it is done without attribution (and even if there were attribution, the license of the text on Wikipedia is not compatible with the SymPy license).

faze-geek:
> I am not sure if just the eval method of Lobb Number class should be merged , it is incomplete and doesn't make complete sense regarding why it is a good addition . maybe something like #22831 would look better !

I agree, you could probably see if writing a recurrence memo( guess by observation since a series of numbers are present ) is feasible for Lobb number ?

faze-geek:
Also to address the failing check
You will need to add yourself to the .mailmap file as described here to get rid of failing authors test:

faze-geek:
> Can you please suggest me something of that sort? I went through the open issues multiple times but nothing really seems doable or that it could get merged quickly

Try to land on some issues with  `easy to fix` and `good first issue` labels
`

PR #23373, opened 2022-04-16
Use @staticmethod in Symbol

eagleoflqj:
`staticmethod()` is [not a pythonic way](https://www.programiz.com/python-programming/methods/built-in/staticmethod) to define static method.
Previously `__new_stage2__` seems a normal method at first look, which confuses pyright.
```sh
$ pyright sympy/core/symbol.py
...
.../sympy/core/symbol.py:302:28 - error: Argument of type "Self@Symbol" cannot be assigned to parameter "cls" of type "Type[Expr]" in function "__new__"
...
```
Using staticmethod decorator is a better way.
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/symbol.py
+++ b/sympy/core/symbol.py
@@ -295,7 +295,8 @@ class Symbol(AtomicExpr, Boolean):
         cls._sanitize(assumptions, cls)
         return Symbol.__xnew_cached_(cls, name, **assumptions)
 
-    def __new_stage2__(cls, name, **assumptions):
+    @staticmethod
+    def __xnew__(cls, name, **assumptions):  # never cached (e.g. dummy)
         if not isinstance(name, str):
             raise TypeError("name should be a string, not %s" % repr(type(name)))
 
@@ -319,10 +320,10 @@ class Symbol(AtomicExpr, Boolean):
         obj._assumptions._generator = tmp_asm_copy  # Issue #8873
         return obj
 
-    __xnew__ = staticmethod(
-        __new_stage2__)            # never cached (e.g. dummy)
-    __xnew_cached_ = staticmethod(
-        cacheit(__new_stage2__))   # symbols are always cached
+    @staticmethod
+    @cacheit
+    def __xnew_cached_(cls, name, **assumptions):  # symbols are always cached
+        return Symbol.__xnew__(cls, name, **assumptions)
 
     def __getnewargs_ex__(self):
         return ((self.name,), self.assumptions0)


asmeurer:
I'm not sure if we should consider these weird methods to be public API or not, but if we do this is breaking.

eagleoflqj:
> I'm not sure if we should consider these weird methods to be public API or not, but if we do this is breaking.

The behavior of `__xnew__` and `__xnew_cached_` are not changed, only `__new_stage2__` is removed.
`__new_stage2__` is not documented, so I don't think it belongs to public API.
Moreover, it's a normal method, but shouldn't be used as normal method. Whoever uses it should just use `__xnew__`. And I guess they (if exist) already do it that way, because undocumented function can only be understood through reading source code, and they will notice `__xnew__` and `__xnew_cached_` are the only places that use `__new_stage2__`, and static method is actually the proper way to use it.
In a word, user of `__new_stage2__` should just change their code.

oscarbenjamin:
Should they not be class methods rather than static methods?

eagleoflqj:
> Should they not be class methods rather than static methods?

Indeed, they are more like class methods. Pyright reports a warning regarding the parameter name `cls`:
`Static methods should not take a "self" or "cls" parameter`
However, all usages of  `__xnew__` are treating it static method:
https://github.com/sympy/sympy/blob/da665dadec268a9ecdd0136bb5bda24f21dcf904/sympy/core/symbol.py#L437
If we change it to class method, this should be `cls.__xnew__(name, **assumptions)`.
If you don't mind also breaking `__xnew__` and `__xnew_cached_` (they are not documented either), I'd give it a try.

oscarbenjamin:
Searching here shows a couple of uses of `__xnew__`:
https://grep.app/search?q=__xnew__

eagleoflqj:
Great website, thanks for showing that. The good news is `__new_stage2__` doesn't seem used.
Converting to classmethod does break them.
It's possible to convert `__xnew__` to class method and provide a `sympy_deprecation_warning` like:
```py
    @classmethod
    def __xnew__(*args, **assumptions):  # never cached (e.g. dummy)
        if len(args) == 3:
            args = args[1:]
            sympy_deprecation_warning(...)
        cls, name = args
```
However, when filling `active_deprecations_target` parameter, I took a look at `active-deprecations.md`, where no such undocumented function seems ever been mentioned. I doubt it not worthy to do so.
Anyway I'm open to follow your suggestions.

oscarbenjamin:
It's best not to change anything unless there is a good reason. It would be good to clean up the symbol creation code and ideally get rid of these methods or at least stop using them. Until then it is probably best not to change anything unnecessarily though.

The change to `__xnew_cached_` looks harmless. Does changing `__xnew__` to a `staticmethod` potentially break anything?

eagleoflqj:
> The change to `__xnew_cached_` looks harmless. Does changing `__xnew__` to a `staticmethod` potentially break anything?

`__xnew__` is originally a static method. This PR doesn’t change that. Only `__new_stage2__` is removed.

oscarbenjamin:
Oh, I see. Yes, and clearly `__new_stage2__` is not something that should every be called because it is just used as input to `staticmethod`. Okay, then this looks good.

PR #23376, opened 2022-04-17
New section on function signatures to the docstring style guide.

moorepants:

I opened this primarily as a way to have some rules about adding type hints to SymPy based on my concerns and the conversations here https://github.com/sympy/sympy/issues/17945. This seemed like the appropriate place for the guidelines and necessitated a section on function signatures in the style guide. I took the liberty of adding two other recommendations to the new section, as those seem to reflect the current expectations in SymPy.

Function signatures are effectively part of the docstring, so we should
have some guidelines for them. This introduces a section in the guide
with three recommendations:

- use meaningful argument names
- avoid using *args and **kwargs
- add type hints only to stub (.pyi) files

#### Release Notes

NO ENTRY

### DIFF
--- a/doc/src/guides/contributing/docstring.rst
+++ b/doc/src/guides/contributing/docstring.rst
@@ -133,12 +133,13 @@ Sections
 In SymPy’s docstrings, it is preferred that function, class, and method
 docstrings consist of the following sections in this order:
 
-1. Single-Sentence Summary
-2. Explanation
-3. Examples
-4. Parameters
-5. See Also
-6. References
+1. Function Signature
+2. Single-Sentence Summary
+3. Explanation
+4. Examples
+5. Parameters
+6. See Also
+7. References
 
 The Single-Sentence Summary and Examples sections are **required** for every
 docstring. Docstrings will not pass review if these sections are not included.
@@ -156,7 +157,27 @@ is unnecessary, do not use it. Unnecessary sections and cluttered docstrings
 can make a function harder to understand. Aim for the minimal amount of
 information required to understand the function.
 
-1. Single-Sentence Summary
+1. Function Signature
+^^^^^^^^^^^^^^^^^^^^^
+
+The function signature is part of the Python source code but is almost always
+displayed along with the doc string. Function signatures should have
+informative argument names that follow Python variable name best practices,
+e.g. use ``coef_matrix`` instead of ``m``. It is also best to avoid the use of
+the generic ``*args`` and ``**kwargs`` in function signatures. Instead, make
+explcitly named arguments and keyword arguments as this is more informative for
+the user.
+
+Python supports `type hints`_ but these may decrease the human readability of
+the function signatures. Type hints should not be added to the function
+signatures in Python modules (``.py`` files). Instead, they should only be
+added in `stub files`_. These stub files have the extension ``.pyi`` and mirror
+the modules which are being typed.
+
+.. _type hints: https://docs.python.org/3/library/typing.html
+.. _stub files: https://mypy.readthedocs.io/en/stable/stubs.html
+
+2. Single-Sentence Summary
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 This section is **required** for every docstring. A docstring will not pass
@@ -176,7 +197,7 @@ in the Sphinx directive::
 
 See :ref:`deprecation-documentation` for more details.
 
-2. Explanation Section
+3. Explanation Section
 ^^^^^^^^^^^^^^^^^^^^^^^
 
 This section is encouraged. If you choose to include an Explanation section in
@@ -193,7 +214,7 @@ class, or method does when the concise Single-Sentence Summary will not
 suffice. This section should be used to clarify functionality in several
 sentences or paragraphs.
 
-3. Examples Section
+4. Examples Section
 ^^^^^^^^^^^^^^^^^^^^
 
 This section is **required** for every docstring. A docstring will not pass
@@ -275,7 +296,7 @@ assumptions), in which case use ``symbols`` like ``x, y = symbols('x y')``.
 In general, you should run ``./bin/doctest`` to make sure your examples run
 correctly, and fix them if they do not.
 
-4. Parameters Section
+5. Parameters Section
 ^^^^^^^^^^^^^^^^^^^^^^
 
 This section is encouraged. If you choose to include a Parameters section in
@@ -333,7 +354,7 @@ Here is an example of a correctly formatted Parameters section::
 
 .. _style_guide_see_also:
 
-5. See Also Section
+6. See Also Section
 ^^^^^^^^^^^^^^^^^^^^^^
 
 This section is encouraged. If you choose to include a See Also section in your
@@ -390,7 +411,7 @@ Here is a correctly formatted See Also section with just a list of names::
 
         """
 
-6. References Section
+7. References Section
 ^^^^^^^^^^^^^^^^^^^^^^
 
 This section is encouraged. If you choose to include a References section in
--- a/doc/src/guides/contributing/docstring.rst
+++ b/doc/src/guides/contributing/docstring.rst
@@ -165,7 +165,7 @@ displayed along with the doc string. Function signatures should have
 informative argument names that follow Python variable name best practices,
 e.g. use ``coef_matrix`` instead of ``m``. It is also best to avoid the use of
 the generic ``*args`` and ``**kwargs`` in function signatures. Instead, make
-explcitly named arguments and keyword arguments as this is more informative for
+explicitly named arguments and keyword arguments as this is more informative for
 the user.
 
 Python supports `type hints`_ but these may decrease the human readability of


eagleoflqj:
-0.5 for not adding type hints directly in .py for even readable ones
+1 for adding type hints in .pyi for overloaded function

doc/src/guides/contributing/docstring.rst
```diff
@@ -156,7 +157,27 @@ is unnecessary, do not use it. Unnecessary sections and cluttered docstrings
...
-1. Single-Sentence Summary
+1. Function Signature
+^^^^^^^^^^^^^^^^^^^^^
+
+The function signature is part of the Python source code but is almost always
+displayed along with the doc string. Function signatures should have
+informative argument names that follow Python variable name best practices,
+e.g. use ``coef_matrix`` instead of ``m``. It is also best to avoid the use of
+the generic ``*args`` and ``**kwargs`` in function signatures. Instead, make
+explcitly named arguments and keyword arguments as this is more informative for
```
jksuom, line 168: explicitly

moorepants:
> -0.5 for not adding type hints directly in .py for even readable ones

If we allow some type hints in .py files then please propose wording to explain what is and isn't allowed in the .py file. Otherwise every type hint PR will cause bikeshedding.

doc/src/guides/contributing/docstring.rst
moorepants, line 168:
```suggestion
explicitly named arguments and keyword arguments as this is more informative for
```

moorepants, line 168:
```suggestion
explicitly named arguments and keyword arguments as this is more informative for
```

moorepants:
> -0.5 for not adding type hints directly in .py for even readable ones

Also, what is the negative consequence of not allowing them in the .py files? Or what does adding them into the py files solve that the pyi files don't? As far as I understand, adding the pyi files would solve all of the needs people have expressed.

moorepants:
> This doesn't strictly belong in the docstring style guide. If we had a code style guide it would be better to put it there.

There isn't one, so I put it here. When there is a code style guide we could move it.

eagleoflqj:
> Also, what is the negative consequence of not allowing them in the .py files? Or what does adding them into the py files solve that the pyi files don't? As far as I understand, adding the pyi files would solve all of the needs people have expressed.

I don't think this should be a rule to enforce, as readability is subjective. It's fine to be a kind suggestion though.
For most non-public functions, they are not overloaded and have simple parameter/return type. It is not necessary to maintain both .py and .pyi for them.
Moreover, local variables can only be typed in .py files. They work together with function type. Separating function type and variable type is just exchanging a kind of readability for another kind of readability.
Finally, there are already lots of type hints for functions. You need extra effort for moving them to .pyi.
Well, I'm not a sympy member. You don't need my approval to proceed.

oscarbenjamin:
I don't think that the use of type-hints in function signatures should be prohibited. Instead it should say something to the effect that overly complicated type-hints or overloads that harm readability should be listed in separate .pyi files.

Again this belongs somewhere else rather than the docstring guide but I think it's worth mentioning that ideally the types accepted and (especially) returned by a function should be sufficiently well defined that the type hints for its signature would not be complicated to express.

Since this is about the docs style guide does it make any difference to the docs whether type hints are inline or in separate .pyi files?

moorepants:
> Instead it should say something to the effect that overly complicated type-hints or overloads that harm readability should be listed in separate .pyi files.

Seems like this would just open for bike shedding in every PR with out any explicitness to the statement and type hints wouldn't actually end up in the stub files.

> Since this is about the docs style guide does it make any difference to the docs whether type hints are inline or in separate .pyi files?

Currently, if type hints are added, these hints would be displayed in any viewing mode of the docstrings: ipython ?, help(), sympy sympy docs website, spyder help, jupyter, source code in a text editor, etc. The function signature is displayed as documentation anywhere that the docstring is viewed from, so it very much seems like part of the documentation as much as the operational code. Secondly, type hints have no runtime functionality, so that makes them only documentation in my book.

I don't think we should adding typing at all to SymPy (at least not at this point in time). Adding type hints only to stub files is a compromise because some users and developers want them. NumPy, for example, made https://github.com/numpy/numpy-stubs which was like a 4 year effort (initial discussions then development of that repo) before the stubs were moved into the NumPy source tree. And even when they were moved in, no types were added to the .py files. That seems like a conservative and safe approach, at least from my perspective. If we allow type hints in stub files now, even into the main sympy repo, that at least lets people that want them to add them and gives an incremental step to whatever the future will look like of Python + typing. It solves the two primary issues:

1. Type hints can be used by source code editors that know what to do with them.
2. Some bugs may be caught by adding type hints and running mypy over the source code.

Maybe when runtime checks occur on type hints and all the docstring viewers know how to handle them, we then merge in the stub files into the py files and remove all the `instance()` checks and utilization of duck typing.

It doesn't seem unreasonable to allow type hints to be added to SymPy right now if we start with an only stub file approach. This doesn't mean it won't change in the future, but it is a non-invasive way to allow the work on typing to move forward without having to change and function signatures. Once all the editors and various ways you can view a function signature have ways to parse and display a human readable form of the function signature with types, then we can relax this rule. I don't see why it has to be all or nothing on type hints. I'm guessing the creators of typing in Python made the stub files, at least partially, for this very purpose. I was very happy to see that they exist and that other big projects (even Python itself) made use of them to ease their way into this very invasive change to how Python code is written.

moorepants:
Once type hints are built out in the stub files, we will all be able to see how complicated they actually are for SymPy. If they aren't complicated then we can move them over to the py files once we know.

fakuivan:
+1 to using stubs. Adding typing directly into the .py files would require using `if typing.TYPE_CHECKING` and forward references or PEP 563 support.

ThePauliPrinciple:
Would it make sense to generate these `pyi` automatically based on the existing doc strings? It would automatically "flush" out any inconsistencies between the written documentations and the actual types found by e.g. mypy, as well as force parts of SymPy that are currently not documented but do have codehints to become documented.

moorepants:
There are tools that generate the pyi files automatically from the function signatures, but the types would need to be added manually. I'm not sure if any of the tools can scrape the docstrings also, maybe!

sylee957:
> Currently, if type hints are added to the py files, these hints would be displayed in any viewing mode of the docstrings: ipython ?, help(), sympy docs website, spyder help, jupyter, source code in a text editor, etc. The function signature is displayed as documentation anywhere that the docstring is viewed from, so it very much seems like part of the documentation as much as the operational code. Secondly, type hints have no runtime functionality, so that makes them only documentation in my book.

This only applies for the function signatures or classes where docstrings are applied, but for local variables, the type signature shouldn't be exposed anyway for them.
And furthermore, there are standard libraries that uses type signatures in runtime like [singledispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch) or [dataclass](https://docs.python.org/3/library/dataclasses.html) in runtime. 
So I think that banning the type signature can't be spoken as strict for them.

doc/src/guides/contributing/docstring.rst
```diff
@@ -156,7 +157,27 @@ is unnecessary, do not use it. Unnecessary sections and cluttered docstrings
...
+1. Function Signature
+^^^^^^^^^^^^^^^^^^^^^
+
+The function signature is part of the Python source code but is almost always
+displayed along with the doc string. Function signatures should have
+informative argument names that follow Python variable name best practices,
+e.g. use ``coef_matrix`` instead of ``m``. It is also best to avoid the use of
+the generic ``*args`` and ``**kwargs`` in function signatures. Instead, make
+explicitly named arguments and keyword arguments as this is more informative for
+the user.
```
asmeurer, line 169: I think what you really mean to say here is that functions should use explicit keyword arguments in the signature itself whenever they are known.

`*args` and `**kwargs` are the only way to do the things that they do. Any function that accepts an arbitrary number of arguments has to use `*args`, and there are a lot of those in SymPy (e.g., `Add`, `Mul`, `Piecewise`, etc.). Of course, if a function doesn't actually accept an arbitrary number of arguments, then `*args` is probably not correct, but I don't think this is a common enough thing that it needs to be mentioned here. 

`**kwargs` also also legitimately used whenever a function needs to pass arbitrary keyword arguments through. For example, `doit(**hints)` passes arbitrary hints through, and any hint can be defined by any Function subclass. 

I would also like to move toward using keyword-only arguments in more places, but the exact policy for that is harder to state (#21803).

doc/src/guides/contributing/docstring.rst
```diff
@@ -156,7 +157,27 @@ is unnecessary, do not use it. Unnecessary sections and cluttered docstrings
...
 information required to understand the function.
 
-1. Single-Sentence Summary
+1. Function Signature
+^^^^^^^^^^^^^^^^^^^^^
+
+The function signature is part of the Python source code but is almost always
+displayed along with the doc string. Function signatures should have
+informative argument names that follow Python variable name best practices,
+e.g. use ``coef_matrix`` instead of ``m``. It is also best to avoid the use of
```
asmeurer, line 166: We might also consider starting to use positional-only arguments in cases where the argument name shouldn't be used as a keyword, which at least prevents the argument name from becoming part of the API.

asmeurer:
Even in pyi files, we should aim to make type hints as readable as possible (e.g., use `|` instead of `Union`, `| None` instead of `Optional`, use readable names for aliases, etc.).

One question I have is which tools actually pull in type hints for human consumption when they are in pyi files? For example, will things like Sphinx, `help()`, and/or various editor environments like VS Code show users type hints if they are only in a stub? Are are stubs *only* read by automated type checkers like mypy?

sylee957:
Unfortunately, I don't think that the stub files are technically feasible solution to satisfy the both world
https://github.com/microsoft/pylance-release/issues/1238
It has serious limitation that is only intended to type check against only with the imported, but not its own implementation

sylee957:
I think that the statement that type annotation should be banned is too broad, the argument applies even for non-public functions and classes, where the supplier-consumer model doesn't work like how .pyi is designed, 
and it's rather an new introduction of the the coding style because we are rejecting syntactially correct code anyway

sylee957:
`pyi` files are only for consumers. It won't help library developers for using type hints to make the code precise or anyway. 
It is overly conservative.

I would rather think in other direction, it should be free for library developers to type as precisely as they should, 
but we could provide `pyi` for some part as alternative for consumers to type as readably as they can.

doc/src/guides/contributing/docstring.rst
sylee957: I rather think that this is a good example where type annotation can be better and make less verbose variables `coef: Matrix`, 
because we wouldn't need to make the names of the variable longer like [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) if its semantics directly link to the type.

asmeurer:
> but the types would need to be added manually. I'm not sure if any of the tools can scrape the docstrings also, maybe!

I really don't think we can do it automatically. If that worked, this could just become the job of the type checking tool.

The problem is that type hints are very often much more complicated than what you'd want to put in a docstring in order to be completely technically correct for the type checker. In SymPy an example I can think of is that any function that accepts `Expr` should also accept `int` or `float`, which would be converted to `Expr` by `sympify` at the top of the function (in fact, I'd hope that the ability to check whether or not `sympify` has been applied to the function arguments would be one positive outcome of using type hints). So in order for this to work correctly, the type hint would need to be an alias like `Sympifiable` or something like that instead of `Expr`. A fully correct definition of `Sympifiable` would be quite complicated due to the large number of user customizable content that sympify can accept (and I doubt it would even be possible to make it work with user entries in the `converter` dictionary). In the docstrings, it would be better to just list such types as `Expr`, even if they technically include anything that can be sympified to `Expr` since this gets at the crux of the type for user purposes.

There are other scenarios too. The matrices have a complicated type heirarchy due to the way subclassing is used, and it might be important to use specific classes in the type hints. But user documentation should generally just say Matrix or at best MutableMatrix or ImmutableMatrix since the other classes aren't really intended for user consumption.

ThePauliPrinciple:
> The problem is that type hints are very often much more complicated than what you'd want to put in a docstring in order to be completely technically correct for the type checker. In SymPy an example I can think of is that any function that accepts `Expr` should also accept `int` or `float`, which would be converted to `Expr` by `sympify` at the top of the function (in fact, I'd hope that the ability to check whether or not `sympify` has been applied to the function arguments would be one positive outcome of using type hints). So in order for this to work correctly, the type hint would need to be an alias like `Sympifiable` or something like that instead of `Expr`. A fully correct definition of `Sympifiable` would be quite complicated due to the large number of user customizable content that sympify can accept (and I doubt it would even be possible to make it work with user entries in the `converter` dictionary). In the docstrings, it would be better to just list such types as `Expr`, even if they technically include anything that can be sympified to `Expr` since this gets at the crux of the type for user purposes.

Isn't this exactly what abstract base classes are meant to do? Passing `assert isinstance(1, Sympifiable)` can be done by simply looking up if `type(1)` is in the converter dicts.

oscarbenjamin:
Type checkers use static analysis so they won't have access to the converters dict. I expect there's some other way to make it work but I think it will need explicit type declarations somehow for the type checker to understand what exactly is sympifiable.

doc/src/guides/contributing/docstring.rst
moorepants: Happy to improve the wording.

asmeurer: How about:

> Use explicit keyword arguments whenever they are known instead of `**kwargs`. For example, if a function takes a flag `simplify=True`, include `f(expr, simplify=True)` in the signature instead of using `f(expr, **kwargs)` with `kwargs.get('simplify', True)` in the body. Doing this will show the actual flag name and the default value in `help()` and the online documentation. If the default is something that can't be included in the function signature itself, use `None` instead and document what the default is in the docstring. `**kwargs` should only be used for functions that accept arbitrary keyword arguments, for example, `**assumptions` for functions that accept assumptions or `**hints` for functions like `doit()` that recursively pass hints through.

Again, I don't think we need to mention `*args` (unless you've seen people misusing it?). Ideally there would also be something there about making keyword arguments keyword-only, but we probably need to think a little more about what that policy should look like first.

asmeurer: OK, I actually did just see this pattern in SymPy:

```py
def func(*args):
    x, y = args
```

So I guess people are misusing it :) `*args` is completely wrong there. If the number of args is a fixed number that is known ahead of time, the arguments should just be in the signature itself, like

```py
def func(x, y):
```

But in general, `*args` is actually needed in a lot of cases, so this needs to be worded in a way that only disallows `*args` for cases like the above. Or sometimes the number of arguments is a fixed number but it can be something like 2 args or 3 args. In those cases, it's sometimes simpler to just use `*args` and explain the signature in the docstring, rather than trying to emulate it without using `*args`. An example is `Range`, which accepts 1, 2, or 3 arguments, but the 1st argument is different for the 1-argument version, so trying to write it without `*args` would require something obtuse like `Range(start_or_stop, stop=None, step=None)`. Instead it just uses `Range(*args)` the the three versions are described at the top of the docstring.

oscarbenjamin: Range is an odd one that is designed to match `range` which also has a strange signature.

One thing I think that should be discouraged is mixing `*args` with accepting an iterable arg e.g.:
```python
In [14]: solve([x+y, x-y-1], [x, y])
Out[14]: {x(t): 1/2, y(t): -1/2}

In [15]: solve([x+y, x-y-1], x, y)
Out[15]: {x(t): 1/2, y(t): -1/2}
```
There isn't really a robust way to tell which form the user used. It is better just to be clear and ask for a single iterable argument. I think that `*args` should generally be discouraged unless there is some good reason for it like `Add(*args)`.

asmeurer: > One thing I think that should be discouraged is mixing *args with accepting an iterable arg e.g.:

I agree with that (although which is better depends on the function). That's actually one benefit of type hints. When the types start getting complicated (like `Sequence[Expr] | Expr` instead of just `Expr`) it's a sign that there you might want to reconsider the API design. Although we do embrace polymorphic typing a lot in SymPy for the sake of usability (consider for instance, what the type signature of `subs` would be).

Issue #11254, opened 2016-06-18
Integration of sech and csch functions


nchisholm:
Sympy does not evaluate the integrals involving the hyperbolic secant and hyperbolic cosecant functions when I try to do so.
~~~
In [1]: import sympy

In [2]: sympy.init_session()
IPython console for SymPy 1.0 (Python 3.5.1-64-bit) (ground types: gmpy)
...

In [3]: integrate(sech(x), (x, 0, 1))
Out [3]:
1
⌠          
⎮ sech(x) dx
⌡           
0

In [4]: integrate(sech(x)**2, x)
Out[4]: 
⌠            
⎮     2      
⎮ sech (x) dx
⌡        
~~~
The result is similar for `csch`.  However, `sinh`, `cosh`, `tanh`, and `coth` all evaluate as expected, so I can do
```
In [5]: integrate(1/cosh(x)**2, x)
Out[5]:
       ⎛x⎞  
 2⋅tanh⎜─⎟  
       ⎝2⎠  
────────────
    2⎛x⎞    
tanh ⎜─⎟ + 1
     ⎝2⎠    

In [6]: _.simplify()
Out [6]: tanh(x)
```
which is the result I expected for integrating `sech(x)**2`.  Is there a reason for this behavior?

kritkaran94:
@nchisholm It's a bug I guess. I am working on this.

kritkaran94:
@jksuom @debugger22 Any idea about this issue. Looks a bug to me.

jksuom:
Integration of hyperbolic functions is not fully implemented. It could be done in the Risch integration module, but that is not complete yet. Currently, some hyperbolic integrands are handled by the parallel (or poor man's) Risch integration module `heurisch`. However, that only works with some integrands, and there is no easy way to improve its performance.
It seems to me that the best way to deal with integrands involving hyperbolic functions would be within `manualintegrate`. That should first rewrite such functions in terms of the exponential function, and then integrate them by substituting the exponential as a new integration variable. Finally, one could attempt to transform the result back to a form containing hyperbolic functions, if possible.

kritkaran94:
@jksuom Thanks a lot for that. Can I work on this issue ?

jksuom:
You are welcome to work on that.

nchisholm:
Thanks for the response! It sounds like @jksuom is roughly suggesting the following procedure:
```python
# isympy -I
[...]

PR #23379, opened 2022-04-17
manualintegrate hyperbolic

### LINKED ISSUE #6947: Step by step for hyperbolic functions
```
Sympy Gamma knows how to do the integral of sin(x) step by step: http://www.sympygamma.com/input/?i=integrate%28sin%28x%29%2C+x%29 but not the similar sinh(x): http://www.sympygamma.com/input/?i=integrate%28sinh%28x%29%2C+x%29
```

Original issue for #6947: http://code.google.com/p/sympy/issues/detail?id=3848
Original author: https://code.google.com/u/107490137238222069432/
Original owner: https://code.google.com/u/112898427768461421869/

eagleoflqj:

Fixes #6947

Add step support for integrating all 6 kinds of hyperbolic functions
#### Release Notes

* integrals
  * Support integral steps for hyperbolic functions

### DIFF
--- a/sympy/integrals/manualintegrate.py
+++ b/sympy/integrals/manualintegrate.py
@@ -869,15 +869,27 @@ def hyperbolic_rule(integral: tuple[Expr, Symbol]):
             return HyperbolicRule('sinh', symbol, integrand, symbol)
         if integrand.func == cosh:
             return HyperbolicRule('cosh', symbol, integrand, symbol)
+        u = Dummy('u')
         if integrand.func == tanh:
             rewritten = sinh(symbol)/cosh(symbol)
-        elif integrand.func == coth:
+            return RewriteRule(rewritten,
+                   URule(u, cosh(symbol), None,
+                   ReciprocalRule(u, 1/u, u), rewritten, symbol), integrand, symbol)
+        if integrand.func == coth:
             rewritten = cosh(symbol)/sinh(symbol)
-        elif integrand.func in (sech, csch):
-            rewritten = integrand.rewrite(tanh)
+            return RewriteRule(rewritten,
+                   URule(u, sinh(symbol), None,
+                   ReciprocalRule(u, 1/u, u), rewritten, symbol), integrand, symbol)
         else:
-            assert False
-        return RewriteRule(rewritten, integral_steps(rewritten, symbol), integrand, symbol)
+            rewritten = integrand.rewrite(tanh)
+            if integrand.func == sech:
+                return RewriteRule(rewritten,
+                       URule(u, tanh(symbol/2), None,
+                       ArctanRule(S(2), S.One, S.One, 2/(u**2 + 1), u), rewritten, symbol), integrand, symbol)
+            if integrand.func == csch:
+                return RewriteRule(rewritten,
+                       URule(u, tanh(symbol/2), None,
+                       ReciprocalRule(u, 1/u, u), rewritten, symbol), integrand, symbol)
 
 @cacheit
 def make_wilds(symbol):
--- a/sympy/integrals/manualintegrate.py
+++ b/sympy/integrals/manualintegrate.py
@@ -17,6 +17,7 @@ To enable simple substitutions, add the match to find_substitutions.
 
 """
 
+from __future__ import annotations
 from typing import Dict as tDict, Optional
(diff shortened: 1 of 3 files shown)


sympy/integrals/manualintegrate.py
```diff
@@ -859,6 +861,24 @@ def root_mul_rule(integral):
...
+
+def hyperbolic_rule(integral: tuple[Expr, Symbol]):
+    integrand, symbol = integral
+    if isinstance(integrand, HyperbolicFunction) and integrand.args[0] == symbol:
+        if integrand.func == sinh:
+            return HyperbolicRule('sinh', symbol, integrand, symbol)
+        if integrand.func == cosh:
+            return HyperbolicRule('cosh', symbol, integrand, symbol)
+        if integrand.func == tanh:
+            rewritten = sinh(symbol)/cosh(symbol)
```
oscarbenjamin, line 873: Why not just return the integral of tanh here?

oscarbenjamin, line 873: Why not just return the integral of tanh here?

eagleoflqj: It may not be a trivial thing for all people I guess.
IMHO `integrate(tanh(x))=log(cosh(x))` is not "atomic" enough. Atomic (or basic) integration rules come directly from basic derivation rules. We memorize `diff(sin(x))=cos(x)`, so we automatically get `integrate(cos(x))=sin(x)`. I don't think people need to memorize `diff(log(cosh(x))=tanh(x)`.

oscarbenjamin: Is there a more direct way to indicate what the next step should be then?

I don't know manualintegrate so well but it looks like `RewriteRule` goes back to the beginning and tries to figure out what rule to use again. Here at this stage we already know exactly what rules should be applied to compute the integral fully so that seems wasteful.

eagleoflqj: That's a good point. Hard-coding further rules make code more complicated, but I do think the test cases I add run a bit slow. I will update it and do benchmark to see if it's worth.

oscarbenjamin: The `manualintegrate` routine is very slow and can often be the slowest part of a call to `integrate` so it makes general integration slow as well.

asmeurer: I do agree that manualintegrate should aim to give user understandable steps and not just to produce an answer, since the steps can be used for things like "integration steps" in SymPy Gamma and similar tools. 

If the API were to allow specifying which rule to apply that would also mean it would be possible for the user to specify the rules directly instead of just having manualintegrate guess.

oscarbenjamin: I think it's possible here to specify the next step directly rather than asking `integral_steps` to choose the next step.

eagleoflqj:
1231275ccb7e83b4f3b77629bbf4e0c3e4f34bd8 spends 3s on `test_hyperbolic`, while 273f6002b3ab3385ae3257eefa70134a491521fd spends 0.75s.

sympy/integrals/tests/test_manual.py
```diff
@@ -561,3 +568,16 @@ def test_issue_23348():
...
     assert constant_times_step.context == constant_times_step.constant * constant_times_step.other
+
+
+def test_hyperbolic():
+    assert_is_integral_of(sinh(2*x), cosh(2*x)/2)
+    assert_is_integral_of(x*cosh(x**2), sinh(x**2)/2)
+    assert_is_integral_of(tanh(x), log(cosh(x)))
+    assert_is_integral_of(coth(x), log(sinh(x)))
+    f, F = sech(x), 2*atan(tanh(x/2))
+    assert manualintegrate(f, x) == F
```
oscarbenjamin, line 579: Some of these integrals cannot be computed by the main `integrate` routine on master. Tests should be added for `integrate` to see that it can compute these. Also I would put these tests along with the main trig tests.

oscarbenjamin:
Looks good

oscargus:
Looks like this also solved https://github.com/sympy/sympy/issues/11254

There are some tests in failing_integrals.py that can be moved to actual tests (with the correct result, now the test just checks if it evaluates or not)-

PR #23388, opened 2022-04-19
Code printing: Fix C-style comments: /* ... */

bjodah:
Use correct comment format in C printer (C++ style comments `//` are not universally supported).

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/printing/c.py
+++ b/sympy/printing/c.py
@@ -255,7 +255,7 @@ class C89CodePrinter(CodePrinter):
         return codestring if codestring.endswith(';') else codestring + ';'
 
     def _get_comment(self, text):
-        return "// {}".format(text)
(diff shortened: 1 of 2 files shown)


oscargus:
LGTM!

asmeurer:
What about C++? Another advantage of /* */ comments is that they support multiple lines.

bjodah:
> And they can be inside of expressions.

Yes, I think that's another reason to prefer them. They are fully supported by C++.

asmeurer:
I mean does the C++ printer also use them?

bjodah:
I see, yes it does:
```python
In [6]: cxxcode(Comment("foobar"))
Out[6]: '/* foobar */'
```

PR #23392, opened 2022-04-19
Fix PinJoint velocity calculation

### LINKED ISSUE #22956: Buggy kinematics for serial chain manipulators
I am trying out the serial chain manipulator [example](https://docs.sympy.org/latest/modules/physics/mechanics/examples/multi_degree_freedom_holonomic_system.html) in the sympy mechanics docs but am finding out that the kinematics sympy computes in the backend are buggy. I am attaching the file I use to test it. In the [example](https://docs.sympy.org/latest/modules/physics/mechanics/examples/multi_degree_freedom_holonomic_system.html), the velocity of the center of mass of the simple pendulum should be : 

> u1(t)*N_frame.x + l*u2(t)*B_frame.x + l*(u2(t) + u3(t))*C_frame.x

However, sympy returns : 

> u1(t)*N_frame.x + l*u3(t)/3*B_frame.x + l*u3(t)*C_frame.x

I tried to investigate this by looking at the definition of a PinJoint in sympy. It turns out every joint, when initialized calls the `_set_linear_velocity` function. In case of the PinJoint the `_set_linear_velocity` function calls `child.masscenter.v2pt_theory(parent.masscenter, parent.frame, child.frame)` to impose velocity constraints between the child and parent link mass centers. This velocity constraint is valid under the assumption that child and parent mass centers are fixed in the child frame (which is the frame attached to the center of mass of the child). This would hold true if the parent link has mass concentrated at one end (or at the joint i.e. COM and joint location coincide) but not if the joint location is anywhere other than the COM of the parent link.

Is there a way to define a serial chain manipulator in sympy without making such assumptions about the mass distribution of the links? I am attaching the code that I am using to test

```python
from sympy import zeros, symbols
from sympy.physics.mechanics import Body, PinJoint, PrismaticJoint, JointsMethod, inertia
from sympy.physics.mechanics import dynamicsymbols

q1, q2, q3, u1, u2, u3 = dynamicsymbols('q1, q2, q3, u1, u2, u3')
l, k, c, g, kT = symbols('l, k, c, g, kT')
ma, mb, mc, IBzz= symbols('ma, mb, mc, IBzz')

wall = Body('N')
block = Body('A', mass=ma)
IB = inertia(block.frame, 0, 0, IBzz)
compound_pend = Body('B', mass=mb, central_inertia=IB)
simple_pend = Body('C', mass=mc)

bodies = (wall, block, compound_pend, simple_pend)

slider = PrismaticJoint('J1', wall, block, coordinates=q1, speeds=u1)
rev1 = PinJoint('J2', block, compound_pend, coordinates=q2, speeds=u2,
                child_axis=compound_pend.z, child_joint_pos=l*2/3*compound_pend.y,
                parent_axis=block.z)
rev2 = PinJoint('J3', compound_pend, simple_pend, coordinates=q3, speeds=u3,
                child_axis=simple_pend.z, parent_joint_pos=-l/3*compound_pend.y,
                parent_axis=compound_pend.z, child_joint_pos=l*simple_pend.y)

joints = (slider, rev1, rev2)

print(simple_pend.masscenter.vel(wall.frame))
```

tjstienstra:

Fixes #22956 

Fix incorrect velocity definition of the center of mass of the child body, when the center of mass is not at the joint position.

#### Release Notes

* physics.mechanics
   * Fixed incorrect center of mass velocity bug.

### DIFF
--- a/sympy/physics/mechanics/joint.py
+++ b/sympy/physics/mechanics/joint.py
@@ -526,9 +526,9 @@ class PinJoint(Joint):
 
     def _set_linear_velocity(self):
         self.parent_point.set_vel(self.parent.frame, 0)
-        self.child_point.set_vel(self.parent.frame, 0)
+        self.child_point.set_vel(self.child.frame, 0)
         self.child_point.set_pos(self.parent_point, 0)
-        self.child.masscenter.v2pt_theory(self.parent.masscenter,
+        self.child.masscenter.v2pt_theory(self.parent_point,
                                           self.parent.frame, self.child.frame)
 
 
--- a/sympy/physics/mechanics/tests/test_joint.py
+++ b/sympy/physics/mechanics/tests/test_joint.py
@@ -181,8 +181,7 @@ def test_pinjoint_arbitrary_axis():
     assert A.ang_vel_in(N) == omega*N.x
     assert A.ang_vel_in(N).express(A) == omega * A.y
     assert A.ang_vel_in(N).magnitude() == sqrt(omega**2)
-    angle = A.ang_vel_in(N).angle_between(A.y)
-    assert angle.xreplace({omega: 1}) == 0
+    assert A.ang_vel_in(N).cross(A.y) == 0
     assert C.masscenter.vel(N) == omega*A.z
     assert C.masscenter.pos_from(P.masscenter) == -A.x
     assert (C.masscenter.pos_from(P.masscenter).express(N).simplify() ==
--- a/sympy/physics/mechanics/tests/test_joint.py
+++ b/sympy/physics/mechanics/tests/test_joint.py
@@ -203,7 +203,7 @@ def test_pinjoint_arbitrary_axis():
     assert A.ang_vel_in(N).magnitude() == sqrt(omega**2)
     angle = A.ang_vel_in(N).angle_between(A.x)
     assert angle.xreplace({omega: 1}) == 0
-    assert C.masscenter.vel(N).simplify() == - omega*N.z
+    assert C.masscenter.vel(N) == 0
     assert C.masscenter.pos_from(P.masscenter) == N.x
 
     # Both joint pos id defined but different axes
@@ -282,16 +282,15 @@ def test_pinjoint_arbitrary_axis():
     angle = A.ang_vel_in(N).angle_between(A.x+A.y-A.z)
     assert angle.xreplace({omega: 1}) == 0
     assert (C.masscenter.vel(N).simplify() ==
-            (m*omega*N.y + m*omega*N.z + n*omega*A.y + n*omega*A.z)/sqrt(3))
+            sqrt(3)*n*omega/3*A.y + sqrt(3)*n*omega/3*A.z)
     assert C.masscenter.pos_from(P.masscenter) == m*N.x - n*A.x
     assert (C.masscenter.pos_from(P.masscenter).express(N).simplify() ==
             (m + n*(2*cos(theta) - 1)/3)*N.x + n*(2*sin(theta + pi/6) +
             1)/3*N.y - n*(2*cos(theta + pi/3) + 1)/3*N.z)
     assert (C.masscenter.vel(N).express(N).simplify() ==
-            -2*n*omega*sin(theta)/3*N.x + (sqrt(3)*m +
-                                           2*n*cos(theta + pi/6))*omega/3*N.y
-            + (sqrt(3)*m + 2*n*sin(theta + pi/3))*omega/3*N.z)
-    assert expand_mul(C.masscenter.vel(N).angle_between(m*N.x - n*A.x)) == pi/2
+            - 2*n*omega*sin(theta)/3*N.x + 2*n*omega*cos(theta + pi/6)/3*N.y +
+            2*n*omega*sin(theta + pi/3)/3*N.z)
+    assert C.masscenter.vel(N).dot(N.x - N.y + N.z).simplify() == 0
 
 
 def test_pinjoint_pi():


TJStienstra:
The checks are already added by correcting the tests. Because the necessary test is checking the velocity of `child.masscenter`, when it is not positioned on the joint axis. Also note that I changed some other incorrect assertions.

sympy/physics/mechanics/joint.py
```diff
@@ -526,9 +526,9 @@ def _set_angular_velocity(self):
 
     def _set_linear_velocity(self):
         self.parent_point.set_vel(self.parent.frame, 0)
-        self.child_point.set_vel(self.parent.frame, 0)
+        self.child_point.set_vel(self.child.frame, 0)
```
moorepants, line 529: In the pin joint, wouldn't the velocity of both the parent and child points be zero in both frames?

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -203,7 +203,7 @@ def test_pinjoint_arbitrary_axis():
     assert A.ang_vel_in(N).magnitude() == sqrt(omega**2)
     angle = A.ang_vel_in(N).angle_between(A.x)
     assert angle.xreplace({omega: 1}) == 0
```
moorepants, line 205: The angle between the angular velocity vector and the rotation axis should be zero. Odd that you have to subsitute omega: 1 here. Not quite sure what that is about. The angle shouldn't be a function of omega.

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -203,7 +203,7 @@ def test_pinjoint_arbitrary_axis():
...
-    assert C.masscenter.vel(N).simplify() == - omega*N.z
+    assert C.masscenter.vel(N) == 0
```
moorepants, line 206: N is the reference frame of body P. The mass center of body C is at the joint because it wasn't defined. So the velocity should be zero. Makes sense.

sympy/physics/mechanics/tests/test_joint.py
```diff
@@ -282,16 +282,15 @@ def test_pinjoint_arbitrary_axis():
     angle = A.ang_vel_in(N).angle_between(A.x+A.y-A.z)
     assert angle.xreplace({omega: 1}) == 0
     assert (C.masscenter.vel(N).simplify() ==
-            (m*omega*N.y + m*omega*N.z + n*omega*A.y + n*omega*A.z)/sqrt(3))
+            sqrt(3)*n*omega/3*A.y + sqrt(3)*n*omega/3*A.z)
```
moorepants, line 285: Did you work this out by hand to verify it?

TJStienstra: I have used a few methods, since the answer was a bit different than I expected. I used the following code which reuses the defined ReferenceFrames:
```py
Pm = Point("Pm")
Pm.set_vel(N, 0)
Pj = Pm.locatenew('Pj', m * N.x)
Pj.set_vel(N, 0)
Pj.set_vel(A, 0)
Cm = Pj.locatenew('Cj', - n * A.x)
Cm.set_vel(A, 0)
print(Cm.v2pt_theory(Pj, N, A).simplify())
```
Besides this I also plotted the situation with which you can see, that the velocity `n*omega` is split in three parts, of which only the `y` and `z` component appear in the velocity since the `x` is parallel to the arm. I do also expect that the last check is a logical change, since it now checks if indeed the velocity is perpendicular to the axis of the PinJoint.

sympy/physics/mechanics/tests/test_joint.py
TJStienstra: `angle=acos(omega_J(t)/sqrt(omega_J(t)**2))`, so it kind of is `0`, I can replace this one with a dot-product. That is a bit more general and does the same.

sympy/physics/mechanics/joint.py
TJStienstra: Yes, but this way I thought that the `vel` method gets a bit easier job searching and defining both will probably give cycles.

TJStienstra: Setting this velocity to `0` may actually not be correct at all, see #23393. As example you may actually use a `PinJoint` in combination with a `PrismaticJoint`.

moorepants: What's the precise example that demonstrates this?

Also, I'm wondering if we can only really set these velocities once the whole system of body connections is defined. Maybe it is ill designed thinking we can set velocities correctly as we go along.

sympy/physics/mechanics/tests/test_joint.py
moorepants: This is due to a different issue. The bottom should be `Abs(omega_J)`. See https://github.com/sympy/sympy/issues/23173

moorepants: This is fine, but it would probably reduce without subbing in 1 if it was Abs(omega_J).

sympy/physics/mechanics/joint.py
TJStienstra:
```py
from sympy.physics.mechanics import Body, PinJoint, PrismaticJoint, dynamicsymbols

q1, q2, u1, u2 = dynamicsymbols('q1, q2, u1, u2')

wall = Body('N')
block = Body('A')
pend = Body('B')

slider = PrismaticJoint('J1', wall, block, coordinates=q1, speeds=u1)
hinge = PinJoint('J2', block, pend, coordinates=q2, speeds=u2)

for point in (block.masscenter, slider.parent_point, slider.child_point):
    print(point)
    print(point.vel(wall.frame))
    print(point.vel(block.frame), end='\n\n')
```
These should all be the same.

TJStienstra: The only change is that the velocity should not be set to `0`, but use the `masscenter` as reference.

TJStienstra: I feel like we are missing something. Could it be the case that the different `ReferenceFrames` can have a different zero-velocity reference? Because than the entire part about the `PrismaticJoint` isn't an issue.

sympy/physics/mechanics/tests/test_joint.py
TJStienstra: I by the way meant `cross` product. That would always just return `0` as long as they are parallel (at least in my tests).

moorepants:
LGTM! Thanks :)

PR #23400, opened 2022-04-20
Update single.py

### LINKED ISSUE #23380: Wrong Answer in the sympy.dsolve for ode
There is an issue in the sympy.dsolve for solving an ode. Here is an example, which provides the wrong answer.
When I was trying to solve the ode: B(t)^2 - B'(t) +1 =0. The proper result should be tanh(C_1 - t), while the given result is the 1/tanh(C_1 - t).

IMG_7136

Here is the [discussion](https://groups.google.com/g/sympy/c/JFEfn_ECzqs) in the [mailing list](https://groups.google.com/g/sympy/c/F0DKwUYaAw0).

KangOxford:
Issue 23380 Fixed : Wrong Answer in the sympy.dsolve

Fixes #23380

One of the ode hint 1st_exact is not right. I have fixed it, with reasons explained and tests posted in the #23380

#### Release Notes

* solvers
  * fixed the bug in the ode First_Exact situation

### DIFF
--- a/sympy/solvers/ode/single.py
+++ b/sympy/solvers/ode/single.py
@@ -563,7 +563,7 @@ class FirstExact(SinglePatternODESolver):
             factor = exp(Integral(factor, integration_variable))
             m *= factor
             n *= factor
-            self._wilds_match[P] = m.subs(y, fx)
+            self._wilds_match[P] = m.subs(y, 0)
             self._wilds_match[Q] = n.subs(y, fx)
             return True
 
@@ -577,8 +577,8 @@ class FirstExact(SinglePatternODESolver):
         m = m.subs(fx, y)
         n = n.subs(fx, y)
 
-        gen_sol = Eq(Subs(Integral(m.subs(y,0), x)
-                + Integral(n - Integral(m.subs(y,0), x).diff(y), y), y, fx), C1)
+        gen_sol = Eq(Subs(Integral(m, x)
+                + Integral(n - Integral(m, x).diff(y), y), y, fx), C1)
         return [gen_sol]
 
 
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
This has caused a test failure for one of the example ODES included in the tests. If you run the tests yourself then you can identify which ODE it is that fails and see what solution is now returned by dsolve.

KangOxford:
> [](/apps/github-actions) ** test / tests2-latest (pull_request) **

Thank you for you reply!  This is the test : `test / tests2-latest (pull_request) `
The error comes from the constant term
```python
from sympy import *
x = Symbol('x')
f = Function("f")
eq = cos(f(x)) - (x*sin(f(x)) - f(x)**2)*f(x).diff(x)
Eq(x*cos(f(x)) + f(x)**3/3, Rational(1, 3))
```
image

KangOxford:
I failed the `slow2` test, is there anyone who knows what this test is, and how can I fix it?
image
image

sympy/solvers/ode/single.py
```diff
@@ -563,7 +563,7 @@ def _verify(self, fx) -> bool:
             factor = exp(Integral(factor, integration_variable))
             m *= factor
             n *= factor
-            self._wilds_match[P] = m.subs(y, fx)
+            self._wilds_match[P] = m.subs(y, 0)
```
jksuom, line 566: 0 is not a good choice if factor has a singularity at y=0.

asmeurer: This is why the ODE solvers generally avoid typical definite integral type solutions you might find in textbooks. They require that the integration bounds be chosen in some "nice" region of the function's domain, but we can't do that for arbitrary expressions. Instead we use definite integrals and "substituted" integrals (like `integrate(f(x), (x, a))`) and let this complexity live in the implicit restrictions on the arbitrary constants. The general solution in the [documentation](https://docs.sympy.org/latest/modules/solvers/ode.html#st-exact) is based on `x0` and `y0`, but those have to be chosen based on the specific ODE.

PR #23404, opened 2022-04-21
Improve printing for `PrimeIdeal`

### LINKED ISSUE #22974: polys/numberfields: `PrimeIdeal` printing methods need work.

skieffer:

Fixes #22974 

* It is now possible for a `Submodule` representing the maximal ideal of an `AlgebraicField` to have a reference back to that field.
* When the maximal ideal is constructed via `AlgebraicField.maximal_order()` or `AlgebraicField.integral_basis()` it _does_ have such a reference.
* As a consequence, we're able to improve the printing methods for `PrimeIdeal`, so that they use the alias of the associated field's primitive element, if any. In particular, we:
  - Support latex printing
  - Rename `_pretty()` --> `repr()` since this is not 2D printing.
  - Provide a `__str__()` method, which prints less info than the `__repr__()` method.

#### Release Notes

* polys
  * Improve printing for `PrimeIdeal`

### DIFF
--- a/sympy/polys/numberfields/modules.py
+++ b/sympy/polys/numberfields/modules.py
@@ -712,11 +712,9 @@ class PowerBasis(Module):
         K = None
         if isinstance(T, AlgebraicField):
             K, T = T, T.ext.minpoly_of_element()
-        if T.domain == QQ:
-            try:
-                T = Poly(T, domain=ZZ)
-            except CoercionFailed:
-                raise ValueError('A polynomial over ZZ is required')
+        # Sometimes incoming Polys are formally over QQ, although all their
+        # coeffs are integral. We want them to be formally over ZZ.
+        T = T.set_domain(ZZ)
         self.K = K
         self.T = T
         self._n = T.degree()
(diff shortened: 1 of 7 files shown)


sympy/polys/numberfields/modules.py
```diff
@@ -678,15 +702,30 @@ def __init__(self, T):
...
+            or (2) an :py:class:`~.AlgebraicField` whose primitive element
+            is the generator of the power basis.
 
         """
+        K = None
+        if isinstance(T, AlgebraicField):
+            K, T = T, T.ext.minpoly_of_element()
+        if T.domain == QQ:
+            try:
+                T = Poly(T, domain=ZZ)
```
oscarbenjamin, line 717: The direct way to do this is `T.set_domain(ZZ)`.

oscarbenjamin: I guess it isn't necessary to check if the domain is `QQ` since it should also raise for e.g. `ZZ[x]` if it can't be coerced to `ZZ`.

Another possibility would be to use `clear_denoms` I guess.

oscarbenjamin:
Looks good to me.

sympy/polys/numberfields/modules.py
skieffer: `set_domain()` looks good for this. Thanks, was not aware of this method!

PR #23409, opened 2022-04-23
Improve `PrimeIdeal` reduction methods.

### LINKED ISSUE #23403: `PrimeIdeal._reduce_poly` needs work
This method was kept private in v1.10. It should be repaired before v1.11.

There are different ways to "reduce" an algebraic number mod a prime ideal.
I would say the existing method lands halfway between the functionality offered by
the `nfmodpr` and `nfeltreduce` methods of PARI. While it could be useful in some
circumstances, it's not really a canonical or well-recognized form of reduction.
In many cases it's essentially the same as `nfmodpr`, but it would fail there in
cases where the rational prime `p` divided the index of the number field.

It's also a funny operation in that it accepts a poly, instead of one of our
classes that more directly represents an algebraic number.

I think an attempt at a function like `nfmodpr` is a bit premature right now.

What we do need right now is a function like PARI's `nfeltreduce`, in order to best resolve #23402.
And for this there is a simple algorithm in the literature, namely Alg 1.4.12 of Cohen's second
book (Advanced Topics in Computational Number Theory).

skieffer:

Resolves #23403 

Provides reduction mod ideals, according to Alg 1.4.12 of [Cohen00].

#### Release Notes

* polys
  * Provide method to reduce an algebraic number to a small representative mod an ideal.

### DIFF
--- a/sympy/polys/numberfields/modules.py
+++ b/sympy/polys/numberfields/modules.py
@@ -1166,7 +1166,9 @@ class Submodule(Module, IntegerPowerable):
         If this submodule $B$ has defining matrix $W$ in square, maximal-rank
         Hermite normal form, then, given an element $x$ of the parent module
         $A$, we produce an element $y \in A$ such that $x - y \in B$, and the
-        $i$th coordinate of $y$ satisfies $0 \leq y_i < w_{i,i}$.
+        $i$th coordinate of $y$ satisfies $0 \leq y_i < w_{i,i}$. This
+        representative $y$ is unique, in the sense that every element of
+        the coset $x + B$ reduces to it under this procedure.
 
         Explanation
         ===========
--- a/sympy/polys/numberfields/modules.py
+++ b/sympy/polys/numberfields/modules.py
@@ -180,6 +180,7 @@ to support this.
 
 from sympy.core.numbers import igcd, ilcm
 from sympy.core.symbol import Dummy
+from sympy.polys.polyclasses import ANP
 from sympy.polys.polytools import Poly
 from sympy.polys.densetools import dup_clear_denoms
 from sympy.polys.domains.algebraicfield import AlgebraicField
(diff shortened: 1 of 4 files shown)


sympy/polys/numberfields/modules.py
```diff
@@ -1127,6 +1161,76 @@ def __mul__(self, other):
...
+        $A$, we produce an element $y \in A$ such that $x - y \in B$, and the
+        $i$th coordinate of $y$ satisfies $0 \leq y_i < w_{i,i}$.
+
+        Explanation
+        ===========
+
+        In the special case where $A$ is a power basis for a number field $K$,
+        and $B$ is a submodule representing an ideal $I$, this operation
+        represents one of a few important ways of reducing an element of $K$
+        modulo $I$ to obtain a "small" representative. See [Cohen00] Section
```
oscargus, line 1177:
```suggestion
        modulo $I$ to obtain a "small" representative. See [Cohen00]_ Section
```
Should create a clickable link.

skieffer: Thanks!

sympy/polys/numberfields/modules.py
```diff
@@ -1127,6 +1161,76 @@ def __mul__(self, other):
     def _first_power(self):
         return self
 
+    def reduce_element(self, elt):
+        r"""
+        If this submodule $B$ has defining matrix $W$ in square, maximal-rank
+        Hermite normal form, then, given an element $x$ of the parent module
+        $A$, we produce an element $y \in A$ such that $x - y \in B$, and the
+        $i$th coordinate of $y$ satisfies $0 \leq y_i < w_{i,i}$.
```
oscarbenjamin, line 1169: Does this define a unique element `y` for all of `x + I`?

skieffer: I think it does, yes. Maybe the docstring should mention that.

Check:
Suppose `x + I` = `u + I`.
Suppose `x` reduces to `y` under this procedure, and `u` reduces to `v`.
We want to show that `y = v`.

Let `W` be the matrix representing the submodule `I`.
Let `d = y - v`.

The reduction conditions say that `-wkk < dk < wkk` for `k = 1, ..., n`.

But we have `y + I = x + I = u + I = v + I`, so `d in I`,
which means `Wc = d` for some vector `c` of integers.

Then since `W` is upper triangular, the bounds `-wkk < dk < wkk` force `c` to be the zero vector,
hence `d` is also zero, which we wanted to show.

skieffer, line 1169:
```suggestion
        $i$th coordinate of $y$ satisfies $0 \leq y_i < w_{i,i}$. This
        representative $y$ is unique in the sense that every element of
        the coset $x + B$ reduces to it under this procedure.
```

oscarbenjamin:
Looks good to me

PR #23419, opened 2022-04-24
Provide a `lazy` option in `CRootOf` constructor.

### LINKED ISSUE #23418: `CRootOf` should have a `lazy` option
When you construct a `CRootOf`, it makes an initial approximation of the roots
right then, at construction time. It would be good if `CRootOf` offered a `lazy` option, so
that these approximations wouldn't be computed until they were needed.

I imagine this might have many applications, but at the moment I just have one application in mind:
When you form an `AlgebraicNumber` whose root is given by a `CRootOf`, it can be extremely slow, due
to this approximation of roots.

This is not good, because there are many things you can do with an `AlgebraicNumber` that do not
require these approximations at all. For many applications, it is just wasted time.

For example, just forming a field:

```python
>>> import time
>>> from sympy import QQ
>>> t0 = time.time(); k = QQ.cyclotomic_field(47); t1=time.time();
>>> print(t1-t0)
39.494821071624756
```

can take 40 seconds. This should be instantaneous.

skieffer:

Resolves #23418 

* `CRootOf` constructor accepts a kwarg, `lazy=False`. If `True`, approximating the roots is postponed until needed.
* `Domain.alg_field_from_poly()` uses `lazy=True`, so that constructing fields is instantaneous.

#### Release Notes

* polys
  * Provide a `lazy` option in `CRootOf` constructor.
  * Speed up construction of number fields.

### DIFF
--- a/sympy/polys/domains/domain.py
+++ b/sympy/polys/domains/domain.py
@@ -919,7 +919,7 @@ class Domain:
 
         """
         from sympy.polys.rootoftools import CRootOf
-        root = CRootOf(poly, root_index)
+        root = CRootOf(poly, root_index, lazy=True)
         alpha = AlgebraicNumber(root, alias=alias)
         return self.algebraic_field(alpha, alias=alias)
 
--- a/sympy/polys/rootoftools.py
+++ b/sympy/polys/rootoftools.py
@@ -126,7 +126,7 @@ def _imag_count_of_factor(f):
 
 
 @public
-def rootof(f, x, index=None, radicals=True, expand=True):
+def rootof(f, x, index=None, radicals=True, expand=True, lazy=False):
     """An indexed root of a univariate polynomial.
 
     Returns either a :obj:`ComplexRootOf` object or an explicit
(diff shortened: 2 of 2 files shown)


skieffer:
As an experiment I've just tried setting `lazy=True` as the default, and running all unit tests again, and I'm getting some errors.

So maybe this PR still has some flaws....

skieffer:
This is actually trickier than I thought, because of what happens when you pass a _composite_ poly to `CRootOf`.

For irreducibles, this PR might have been okay, and so I'd like to suggest #23426 instead.

Explanation:

If `f` is a composite poly, and you set
```python
r = CRootOf(f, n)
```
then `r.poly` is not going to be equal to `f`. It's going to be some irreducible factor `g` of `f`.
Accordingly, `r.index` is not in general equal to `n`. It's adjusted so that it points to that root of `g` which is the `n`th root of `f`.

And we can't compute the adjusted index before we've done all the work of initializing and refining (so they can be sorted) the isolating intervals.

So I think the most I would try to do would be to make `CRootOf` lazy on irreducibles. See #23426.

PR #23420, opened 2022-04-24
Fix manualintegrate x**x

eagleoflqj:

https://sympy-beta.vercel.app/input/integrate(diff(x**x))
Left is on sympy 1.10.1, right is on this branch.
Screenshot from 2022-04-24 18-57-02

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/integrals/manualintegrate.py
+++ b/sympy/integrals/manualintegrate.py
@@ -211,9 +211,6 @@ def find_substitutions(integrand, symbol, u_var):
         if u_diff == 0:
             return False
         substituted = integrand / u_diff
-        if symbol not in substituted.free_symbols:
-            # replaced everything already
-            return False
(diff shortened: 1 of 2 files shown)


sympy/integrals/manualintegrate.py
```diff
@@ -211,9 +211,6 @@ def test_subterm(u, u_diff):
         if u_diff == 0:
             return False
         substituted = integrand / u_diff
-        if symbol not in substituted.free_symbols:
-            # replaced everything already
-            return False
```
asmeurer, line 216: So the idea is this previously wouldn't allow u-substitutions where the du was just the entire integrand, as in the example? Such as substitution seems unhelpful from a steps point of view, but it's also useful if it does mean it can compute more integrals.

eagleoflqj: Exactly.

PR #23426, opened 2022-04-25
Make `CRootOf` lazy on irreducibles.

### LINKED ISSUE #23418: `CRootOf` should have a `lazy` option

skieffer:

Resolves #23418
Replaces #23419 

If `f` is irreducible, and you construct `CRootOf(f, n)`, then the index `n` need not be adjusted.
This means we are free to postpone calculation of the isolating intervals until they are needed,
and this PR makes this so.

If `f` is composite, then the index `n` may need to be adjusted, which requires calculation of the
isolating intervals at construction time, so we cannot postpone it.

#### Release Notes

* polys
  * Lazy interval calculation for `CRootOf` on irreducible polys

### DIFF
--- a/sympy/polys/rootoftools.py
+++ b/sympy/polys/rootoftools.py
@@ -356,7 +356,7 @@ class ComplexRootOf(RootOf):
         if not dom.is_ZZ:
             raise NotImplementedError("CRootOf is not supported over %s" % dom)
 
-        root = cls._indexed_root(poly, index)
+        root = cls._indexed_root(poly, index, lazy=True)
         return coeff * cls._postprocess_root(root, radicals)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


oscargus:
Looks good!

Do you know if the current tests cover the newly added code or should one add some specific test for this?

skieffer:
Yeah, good question. I ran pytest with a coverage report, and all the new lines do have coverage, just from existing tests.

When I thought about adding new tests I could only think of ones that would involve timings, and I figured I shouldn't add those, due to differences in machines.

skieffer:
On second thought, I could add a test showing the intervals cache is being used in the expected way.

oscarbenjamin:
Looks good

asmeurer:
I expect the code here could be simplified by making the caches special objects that automatically call the init methods when they are accessed.

skieffer:
> making the caches special objects that automatically call the init methods when they are accessed

I like the idea of this design, but it would be hard to give time to it right now. Something to keep in mind though.

Issue #23421, opened 2022-04-25
Regression(?) when using sympy objects wrapped in numpy scalar array


bjodah:
I needed to rerun some analysis code I wrote a while back and noticed a regression, here's a minimal example (this used to work in sympy <1.6):
```python
import numpy as np
import sympy as sy
one = np.array(sy.S.One, dtype=object)
print(sy.exp(one)*sy.exp(one))
```
<details>
<summary>Click to expand full traceback.<pre>TypeError: unsupported operand type(s) for ** or pow(): 'Exp1' and 'ImmutableDenseNDimArray'</pre></summary>
<pre>
/home/bjorn/vc/sympy/sympy/functions/elementary/exponential.py:86: SymPyDeprecationWarning:                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                           
Using non-Expr arguments in Mul is deprecated (in this case, one of                                                                                                                                                                                                                        
the arguments has type 'ImmutableDenseNDimArray').                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                           
If you really did intend to use a multiplication or addition operation with                                                                                                                                                                                                                
this object, use the * or + operator instead.                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                           
See https://docs.sympy.org/latest/explanation/active-deprecations.html#non-expr-args-deprecated                                                                                                                                                                                            
for details.                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                           
This has been deprecated since SymPy version 1.7. It                                                                                                                                                                                                                                       
will be removed in a future version of SymPy.                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                           
  return self.func(1), Mul(*self.args)                                                                                                                                                                                                                                                     
Traceback (most recent call last):                                                                                                                                                                                                                                                         
  File "sympy-regression-scalar-numpy-array.py", line 4, in <module>                                                                                                                                                                                                                       
    print(sy.exp(one)*sy.exp(one))                                                                                                                                                                                                                                                         
  File "/home/bjorn/vc/sympy/sympy/core/decorators.py", line 236, in _func                                                                                                                                                                                                                 
    return func(self, other)                                                                                                                                                                                                                                                               
  File "/home/bjorn/vc/sympy/sympy/core/decorators.py", line 106, in binary_op_wrapper                                                                                                                                                                                                     
    return func(self, other)                                                                                                                                                                                                                                                               
[...]

PR #23428, opened 2022-04-26
sympify: unpack numpy 0-d arrays

bjodah:
This partially addresses the issues in discuessed in gh-23421

Prior to sympy-1.6, np.array(1) used to sympify into `S.One`, since sympy-1.6, sympify needed the strict flag set to `True` for this conversion to keep happen. With default `strict=False` it was instead wrapped in a `ImmutableDenseNDimArray`.

#### Other comments
Having different output from `sympify` based on `strict` flag is undesirable, the strict flag should only cause exceptions to be thrown more readily when set.

#### Release Notes

* core
  * Sympify of 0-dimensional numpy arrays are no longer wrapped in an ImmutableDenseNDimArray

### DIFF
--- a/sympy/core/sympify.py
+++ b/sympy/core/sympify.py
@@ -416,7 +416,7 @@ def sympify(a, locals=None, convert_xor=True, strict=False, rational=False,
         flat = getattr(a, "flat", None)
         if flat is not None:
             shape = getattr(a, "shape", None)
-            if shape is not None:
+            if shape is not None and shape != ():
                 from sympy.tensor.array import Array
                 return Array(a.flat, a.shape)  # works with e.g. NumPy arrays
 
--- a/sympy/core/tests/test_sympify.py
+++ b/sympy/core/tests/test_sympify.py
@@ -822,8 +822,10 @@ def test_issue_14706():
     assert sympify(numpy.array([[[1]]])) == ImmutableDenseNDimArray([1], (1, 1, 1))
     assert sympify(z1) == ImmutableDenseNDimArray([0], (1, 1))
     assert sympify(z2) == ImmutableDenseNDimArray([0, 0, 0, 0], (2, 2))
-    assert sympify(z3) == ImmutableDenseNDimArray([0], ())
-    assert sympify(z3, strict=True) == 0.0
+    for strict in [False, True]:  # see gh-23421
+        z3_s = sympify(z3, strict=strict)
+        assert type(z3_s) == Float
+        assert z3_s == 0.0
 
     raises(SympifyError, lambda: sympify(numpy.array([1]), strict=True))
     raises(SympifyError, lambda: sympify(z1, strict=True))


oscargus:
Looks good to me!

ThePauliPrinciple:
Considering that this regression has been in SymPy for a while and considering that changing it back is also potentially a regression for people having written code since this change, it might be worthwhile to consider this a "bug" or "inconsistency" instead and have a debate about what would be the preferable usage? SymPy arrays/matrices are also not converted to scalars when they get reduced to 0d as far as I understand.

It would also make sense to me that `strict` would work the other way round, if you state `strict = True`, it should be similar to the input, while `strict = False` less so. My suspicious is that the current choice is a legacy of not having an array-like subclass of `Basic` at the time this was introduced, which would preclude having it as a result from `strict = True`, but this is no longer the case.

Otherwise, if the stated outcome is desired, the changes look good to me.

asmeurer:
I agree that `strict` should only be about disabling certain types of inputs. In particular, we should expect functions in library code to call `sympify(strict=True)` (or equivalently `_sympify()`) on their inputs.

sympy/core/sympify.py
```diff
@@ -416,7 +416,7 @@ def sympify(a, locals=None, convert_xor=True, strict=False, rational=False,
         flat = getattr(a, "flat", None)
```
asmeurer, line 416: I don't like that this works by duck typing. The scalar check above doesn't. If it's going to do that, it should at least work using methods that are commonly available in other array libraries, which `flat` is not.

asmeurer:
We do treat array scalars differently from arrays already it looks like. Given that, I personally think it makes sense for a 0-D array to return an array in SymPy.

sympy/core/sympify.py
asmeurer: Ideally we would make use of the [array API](https://data-apis.org/array-api/latest/), but that's not perfect because it isn't implemented everywhere yet (even NumPy itself only implements it as a sub-module).

bjodah:
I can be swayed both ways regarding *what* should be returned for 0-d arrays. So, I'll change the PR so that `ImmutableDenseNDimArray` is returned also for `shape==()` (for both `strict=True` & `strict=False`)?

asmeurer:
I have a hard time having a very strong opinion on this because I haven't really used the NDimArray class in SymPy very much. Mathematically a 0-D array is more or less a scalar (though it's not completely true if you want to be 100% pedantic). But I doubt that 0-D ImmutableDenseNDimArray instances work the same as Expr. One will have the expected array methods (like `shape`) and one will have the expected Expr methods. It doesn't even look like ImmutableDenseNDimArray subclasses from Expr, only Basic. 

There's even more complexity here, because the whole situation with NumPy itself is a bit of a mess. NumPy has 0-D arrays and "scalar instances". Some operations give one and some give the other. They are mostly interchangeable from one another, but they also have a lot of subtle differences. In many ways, Expr vs. a 0-D ImmutableDenseNDimArray instance is the same problem in SymPy. The difference is that in NumPy, the array is the first class thing and the scalar instances are the weird offshoot that the mostly wish they could get rid of, whereas in SymPy, Expr is the primary class and ImmutalbeDenseNDimArray is just one, relatively obscure, submodule. 

So my best suggestion would be to not "take sides" and just convert scalars to scalars (which we already do), and convert arrays to arrays, even if they are 0-D.

bjodah:
Yes, I had misunderstood the distinction of "scalar" vs. 0-D array in numpy. I'll close this PR since it would be a breaking change since a few releases (even though techincally it reverts another unannounced breaking change).

PR #23445, opened 2022-04-29
jacobi_symbol: remove dead code paths

### LINKED ISSUE #23443: jacobi_symbol: dead code path
If `m` is negative, this will turn it into positive since `n` is always a positive integer.

https://github.com/sympy/sympy/blob/09c03dc8a17ebd2c65c90f4a12af78a91c715d90/sympy/ntheory/residue_ntheory.py#L998-L999

But a little bit later there's another test for negative values that will never be reached.

https://github.com/sympy/sympy/blob/09c03dc8a17ebd2c65c90f4a12af78a91c715d90/sympy/ntheory/residue_ntheory.py#L1008-L1011

oittaa:
Fixes #23443

The first part of the code was never reached since `m %= n` forces `m` to be non-negative.

The later `n != 1` condition never matched either since the gcd calculation prevented that from ever happening.
```python
    if igcd(m, n) != 1:
        return 0
```

https://en.wikipedia.org/wiki/Jacobi_symbol#Implementation_in_Lua

#### Release Notes

* ntheory
  * Remove dead code paths from `jacobi_symbol`

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1419,6 +1419,7 @@ mohit <39158356+mohitacecode@users.noreply.github.com> mohit <42018918+mohitshah
 naelsondouglas <naelson17@gmail.com>
 noam simcha finkelstein <noam.finkelstein@protonmail.com>
 numbermaniac <5206120+numbermaniac@users.noreply.github.com>
+oittaa <8972248+oittaa@users.noreply.github.com>
 pekochun <hamburg_hamburger2000@yahoo.co.jp>
 prshnt19 <prashant.rawat216@gmail.com>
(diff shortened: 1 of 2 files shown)


airpods69:
You will have to add your name to the .mailmap
this might help

skieffer:
This looks good, and it looks like we now have 100% test coverage for this function.

skieffer:
For the release notes, why `functions` instead of `ntheory`?

eagleoflqj:
@oittaa Thanks for your contribution! I edited the release note. FYI the change you made is under `sympy/ntheory/` so it should be `ntheory`.

PR #23446, opened 2022-04-30
parse_latex \int\limits

eagleoflqj:
Previously:
```py
>>> from sympy.parsing.latex import parse_latex
>>> parse_latex(R'\int\limits_0^{1} x dx')
Integral(limits_{0}**1*x, x)
```
#### Release Notes

NO ENTRY


Upabjojr:
I'm going a bit out of topic (i.e. this comment is not related to this PR), but I think we need a probabilistic parser for LaTeX expressions.

eagleoflqj:
> I think we need a probabilistic parser for LaTeX expressions.

I think everything that can be represented in both LaTex and SymPy object should be parse-able and printable, and there should be some input/output consistency, eg `Integral` outputs `\limits` by default (`itex=False`) but the LaTex parser rejects it before this PR.

asmeurer:
What is a "probabilistic parser"?

eagleoflqj:
I thought that was a parser that can parse `P(X=3|Y=4)` things, but very likely I misunderstood.

PR #16968, opened 2019-06-04
fixes #16925

Alex031544:
Fix LaTeX printing with sub- and superscripts

#### References to other Issues or PRs

Fixes #16925

#### Release Notes

* printing
    * The function names in the LaTeX printer, take the include sub- and superscripts into account then determining the name length. (e.g.: now `a_1` is interpreted as a one letter name with subscript)

PR #17926, opened 2019-11-19
Allow the LatexPrinter to print objects implementing _repr_latex_

eric-wieser:
This way, when sympy enables latex-printing of lists, it will allow those lists to contain any types that ipython already knows how to print.

#### References to other Issues or PRs

* #19389 - types that have already explicitly opted-in to sympy latex printing will be rendered as LaTeX in jupyter
* This PR - types that have _not_ opted-in to _sympy_ printing, but _have_ opted into IPython printing, will be embedded in math mode in order to render from within sympy collections.

#### Brief description of what is fixed or changed

With latex printing enabled in sympy, the following now displays as latex:
```python
class Foo:
    def _repr_latex_(self):
        return r'$\theta$'
    
(Foo(),)
```

#### Release Notes

* interactive
  * respect `_repr_latex_` on custom types

~~cc @abrombo, this might make printing behave better for pygae/galgebra~~

PR #23448, opened 2022-04-30
Fix LaTeX printing with sub- and superscripts

### LINKED ISSUE #16925: fails to print Derivative for latex
Hello,

I create a symbolic expression and want do print it's derivation with latex.
I expect `\frac{d}{d t} q_{1}`, but I receive `\frac{d}{d t} \operatorname{q_{1}}{\left(t \right)}`

Look at this Python3 example:
```py
from sympy import *
from sympy import init_printing
init_printing(use_latex=True)

from sympy.core.function import Function

t = Symbol('t')
q1 = Function('q1')(t)
q1d = diff(q1,t)

print("q1: %s"%q1)
print("q1d: %s"%q1d)
print("latex(q1d): %s"%latex(q1d))
```
results in
```
q1: q1(t)
q1d: Derivative(q1(t), t)
latex(q1d): \frac{d}{d t} \operatorname{q_{1}}{\left(t \right)}
```

How can this be solved?

eagleoflqj:

Fixes https://github.com/sympy/sympy/issues/16925
Revive #16968
#### Release Notes

* printing
    * The function names in the LaTeX printer, take the include sub- and superscripts into account then determining the name length. (e.g.: now `a_1` is interpreted as a one letter name with subscript)


oscargus:
Thanks for correcting your attempt!

I reiterate that what is really needed here, whuch for some reason was not clear enough in the previous PR, despite stating it multole times, is handling of superscript.

At the minimum a test using superscript should be added. After that I suspect that there will be changes required as I do not think that superscripts produce correct LaTeX with the current code.

Also, as Sympy traditionally support __ (double _) to mark superscript, it would be good if that is still supported here. Hence one should use the already existing function to handle the sub- and superscript rather than reinventing a non-Sympy-compliant version.

eagleoflqj:
I hope these are correct tests to add.

.mailmap
```diff
@@ -251,6 +251,8 @@ Alex Lubbock <code@alexlubbock.com>
 Alex Malins <github@alexmalins.com>
 Alex Meiburg <timeroot.alex@gmail.com>
 AlexCQY <alex_chua@u.nus.edu>
+Alexander Behrens <Alex031544@users.noreply.github.com>
+Alexander Behrens <alex.git@gmx.net> Alex031544 <alex.git@gmx.net>
```
oscarbenjamin, line 255: This should not be two separate entries:
https://github.com/sympy/sympy/blob/4165a5016bd8708ee83aea11a22e6217fb38800b/.mailmap#L139-L145

eagleoflqj: Thank you. Fixed.

oscargus:
I'd like to see the following tests as well

```
latex(Function('a^12')(x)**ab)
latex(Function('a__12')(x)**ab)
latex(Function('a__12')(x))
latex(Function('a_1__1_2')(x))
```

These names all work for Symbol. And they currently work for Function and the pretty printer.

oscargus:
I'm still not convinced that this is the best/correct way to fix it, but it is clearly a step in the right direction and the results are syntactically correct.

One may want to discuss if the sub- and superscripts of a function name actually should be put inside `\operatorname` (which I think they should) or not, but it is maybe more a matter of taste. Although if the idea is that it should be upright it will be better to put it inside:
image
vs
image

(Firefox cannot properly paste png:s, so ignore the black border and skewed image.)

But as I said: a step in the right direction.

eagleoflqj:
Thanks for reviewing. I personally don't have strong opinions about this. For better control, maybe #17926 could be made use of.
I'll wait for objections then merge it.

oscargus:
For control one can simply use a LaTeX-string as function name. Or rather, if the function name has a `{`, all translation is skipped (I think, it is like that for `Symbol` anyway). The drawback of that is that the name works badly using `pprint`.

So I think the purpose really is to sort out the conversion from the "SymPy markup".

PR #23454, opened 2022-05-02
isprime: test with `pow(2, n, n)` up to 31417

### LINKED ISSUE #23451: isprime: Is there a reason for the constant to be `23001`?
Is there a particular reason why the value is set to 23001, which makes it optimal, that I don't understand?

https://github.com/sympy/sympy/blob/14198353847f9c5a15cb3def322aedb43e790493/sympy/ntheory/primetest.py#L589-L590

A quick testing reveals if it was changed to 31417 (non-inclusive), only one additional special case would be needed while increasing the range by almost 40%. After that it gets more complicated since there are a lot of special cases right after 31417.

```python
    if n < 31417:
        return pow(2, n, n) == 2 and n not in [7957, 8321, 13747, 18721, 19951, 23377]
```

oittaa:

Fixes #23451

Only one additional special case is needed to test up to 31417. And at least on my computer the difference in used time in `pow` calculation is quite small. Example with primes near the limits:
```bash
$ python3 -m timeit 'pow(2, 31397, 31397)'
500000 loops, best of 5: 675 nsec per loop
$ python3 -m timeit 'pow(2, 22993, 22993)'
500000 loops, best of 5: 645 nsec per loop
```

#### Release Notes

* ntheory
  * `isprime`: test with `pow(2, n, n)` up to 31417

### DIFF
--- a/sympy/ntheory/primetest.py
+++ b/sympy/ntheory/primetest.py
@@ -586,8 +586,8 @@ def isprime(n):
         return False
     if n < 2809:
         return True
-    if n <= 23001:
-        return pow(2, n, n) == 2 and n not in [7957, 8321, 13747, 18721, 19951]
+    if n < 31417:
+        return pow(2, n, n) == 2 and n not in [7957, 8321, 13747, 18721, 19951, 23377]
 
     # bisection search on the sieve if the sieve is large enough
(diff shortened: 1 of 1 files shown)


eagleoflqj (approved):
LGTM

PR #23455, opened 2022-05-02
isprime deterministic to 3317044064679887385961981

oittaa:

This extends deterministic MR solutions to 3317044064679887385961981 (over 2^81)

#### Other comments
https://arxiv.org/pdf/1509.00864.pdf
https://oeis.org/A014233

#### Release Notes

* ntheory
  * `isprime` extend deterministic solutions to over 2^81

### DIFF
--- a/sympy/ntheory/primetest.py
+++ b/sympy/ntheory/primetest.py
@@ -607,6 +607,9 @@ def isprime(n):
     #    https://miller-rabin.appspot.com/
     # for lists.  We have made sure the M-R routine will successfully handle
     # bases larger than n, so we can use the minimal set.
+    # In September 2015 deterministic numbers were extended to over 2^81.
+    #    https://arxiv.org/pdf/1509.00864.pdf
+    #    https://oeis.org/A014233
     if n < 341531:
         return mr(n, [9345883071009581737])
     if n < 885594169:
@@ -621,6 +624,10 @@ def isprime(n):
         return mr(n, [2, 123635709730000, 9233062284813009, 43835965440333360, 761179012939631437, 1263739024124850375])
     if n < 18446744073709551616:
         return mr(n, [2, 325, 9375, 28178, 450775, 9780504, 1795265022])
+    if n < 318665857834031151167461:
+        return mr(n, [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37])
+    if n < 3317044064679887385961981:
+        return mr(n, [2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41])
 
     # We could do this instead at any point:
     #if n < 18446744073709551616:


oittaa:
This might need more benchmarking, but at least on my computer several rounds of Miller-Rabin seems faster than Lucas-test on average. And it might be desirable to have deterministic solutions even though there are no known counterexamples to one round of Miller-Rabin test + Lucas test.

sympy/ntheory/primetest.py
```diff
@@ -621,6 +624,10 @@ def isprime(n):
         return mr(n, [2, 123635709730000, 9233062284813009, 43835965440333360, 761179012939631437, 1263739024124850375])
     if n < 18446744073709551616:
         return mr(n, [2, 325, 9375, 28178, 450775, 9780504, 1795265022])
+    if n < 318665857834031151167461:
```
smichr, line 627: These look suspiciously small relative to the numbers in the other ranges.

asmeurer:
Can you clarify what this does? `isprime(3317044064679887385961981)` returns False already in master.

sympy/ntheory/primetest.py
oittaa: https://arxiv.org/pdf/1509.00864.pdf

oittaa:
> Can you clarify what this does? `isprime(3317044064679887385961981)` returns False already in master.

Yes, there aren't any known pseudoprimes that would pass Baillie–PSW primality test, but to my knowledge it has been proven to be correct "only" up to 2^64. https://arxiv.org/pdf/1509.00864.pdf asserts that Miller-Rabin test with the provided bases is deterministic up to 3317044064679887385961981 (over 2^81). 

The question becomes, is it worth to add complexity to have proven deterministic solutions to around 2^81 instead of 2^64?

eagleoflqj:
> The question becomes, is it worth to add complexity to have proven deterministic solutions to around 2^81 instead of 2^64?

I haven't read the thesis but I think it worth.

eagleoflqj:
I'm not able to judge whether the thesis tells the truth. I just found the table above 4.2 has an obvious typo. `1247050339261*2494100678521` should be `3110269097300703345712981`, not the repeated `3317044064679887385961981`.

jksuom:
> These look suspiciously small relative to the numbers in the other ranges.

It seems that small bases do not make the test less reliable, only less efficient as the range is shorter or the list of bases longer. That can be seen by comparing the tests of [wikipedia](https://en.wikipedia.org/wiki/Miller%E2%80%93Rabin_primality_test#Testing_against_small_sets_of_bases):

if n < 2,047, it is enough to test a = 2;
if n < 1,373,653, it is enough to test a = 2 and 3;
if n < 9,080,191, it is enough to test a = 31 and 73;
if n < 25,326,001, it is enough to test a = 2, 3, and 5;

with more efficient bases like these: https://github.com/sympy/sympy/blob/68c37df362e8585d72fc7ef490013bb8eff16e3e/sympy/ntheory/primetest.py#L610-L615
So it seems that there only remains this question:
> is it worth to add complexity to have proven deterministic solutions to around 2^81 instead of 2^64?

smichr:
OK, it is a small added complexity to make this deterministic to a significantly larger number.

PR #23456, opened 2022-05-02
series - Fixes series expansion around float failing with NotImplementedError

### LINKED ISSUE #23432: Series expansion around float fails with NotImplementedError
This works fine with rational `S.Half` but fails with float `0.5`:
```python
In [5]: series(1/sqrt(1-x**2), x, S.Half)
Out[5]: 
                                      2                   3                   4                    
2⋅√3   4⋅√3⋅(x - 1/2)   8⋅√3⋅(x - 1/2)    112⋅√3⋅(x - 1/2)    608⋅√3⋅(x - 1/2)    1088⋅√3⋅(x - 1/2)
──── + ────────────── + ─────────────── + ───────────────── + ───────────────── + ─────────────────
 3           9                 9                  81                 243                 243       

5                         
     ⎛         6         ⎞
─ + O⎝(x - 1/2) ; x → 1/2⎠
                          

In [6]: series(1/sqrt(1-x**2), x, 0.5)
NotImplementedError                       Traceback (most recent call last)
<ipython-input-6-ab154a1ba9fb> in <module>
----> 1 series(1/sqrt(1-x**2), x, 0.5)

~/current/sympy/sympy.git/sympy/series/series.py in series(expr, x, x0, n, dir)
     61     """
     62     expr = sympify(expr)
---> 63     return expr.series(x, x0, n, dir)

~/current/sympy/sympy.git/sympy/core/expr.py in series(self, x, x0, n, dir, logx, cdir)
   2956                 rep2 = x
   2957                 rep2b = -x0
-> 2958             s = self.subs(x, rep).series(x, x0=0, n=n, dir='+', logx=logx, cdir=cdir)
   2959             if n is None:  # lseries...
   2960                 return (si.subs(x, rep2 + rep2b) for si in s)

~/current/sympy/sympy.git/sympy/core/expr.py in series(self, x, x0, n, dir, logx, cdir)
   2966             # replace x with an x that has a positive assumption
   2967             xpos = Dummy('x', positive=True)
-> 2968             rv = self.subs(x, xpos).series(xpos, x0, n, dir, logx=logx, cdir=cdir)
   2969             if n is None:
   2970                 return (s.subs(xpos, x) for s in rv)

~/current/sympy/sympy.git/sympy/core/expr.py in series(self, x, x0, n, dir, logx, cdir)
   2974         from sympy.series.order import Order
   2975         if n is not None:  # nseries handling
-> 2976             s1 = self._eval_nseries(x, n=n, logx=logx, cdir=cdir)
   2977             o = s1.getO() or S.Zero
   2978             if o:

~/current/sympy/sympy.git/sympy/core/power.py in _eval_nseries(self, x, n, logx, cdir)
   1735             _, d = g.leadterm(x)
   1736             if not d.is_positive:
-> 1737                 raise NotImplementedError()
   1738 
   1739         from sympy.functions.elementary.integers import ceiling

NotImplementedError: 
```

praneethratna:
Fixes #23432 and now `series(1/sqrt(1-x**2), x, 0.5)` returns 
```
0.769800358919501 + 1.539600717839*(x - 0.5)**2 + 2.39493444997178*(x - 0.5)**3 + 4.33369090947275*(x - 0.5)**4 + 7.75502583800386*(x - 0.5)**5 + 0.769800358919501*x + O((x - 1/2)**6, (x, 1/2))
```
instead of `NotImplementedError`.

#### Release Notes

* series
  * fixes bug in `series` and now series expansion with float does not give `NotImplementedError`.

### DIFF
--- a/sympy/core/power.py
+++ b/sympy/core/power.py
@@ -1735,9 +1735,14 @@ class Pow(Expr):
                 raise NotImplementedError()
         if not d.is_positive:
             g = g.simplify()
+            if g.is_zero:
+                return f**e
             _, d = g.leadterm(x)
             if not d.is_positive:
-                raise NotImplementedError()
+                g = ((b - f)/f).expand()
+                _, d = g.leadterm(x)
+                if not d.is_positive:
+                    raise NotImplementedError()
 
         from sympy.functions.elementary.integers import ceiling
         gpoly = g._eval_nseries(x, n=ceiling(maxpow), logx=logx, cdir=cdir).removeO()
--- a/sympy/series/tests/test_series.py
+++ b/sympy/series/tests/test_series.py
@@ -362,3 +362,9 @@ def test_issue_21245():
 def test_issue_21938():
     expr = sin(1/x + exp(-x)) - sin(1/x)
     assert expr.series(x, oo) == (1/(24*x**4) - 1/(2*x**2) + 1 + O(x**(-6), (x, oo)))*exp(-x)
+
+
+def test_issue_23432():
+    expr = 1/sqrt(1 - x**2)
+    result = expr.series(x, 0.5)
+    assert result.is_Add and len(result.args) == 7


praneethratna:
CC @jksuom

sympy/core/power.py
```diff
@@ -1736,6 +1736,11 @@ def mul(d1, d2):
         if not d.is_positive:
             g = g.simplify()
             _, d = g.leadterm(x)
+            if not d.is_positive:
+                if g.is_zero:
+                    return f**e
```
jksuom, line 1741: This block could come  immediately after simplification on line 1737. If g is zero, there is no need to compute the leadterm.

sympy/core/power.py
```diff
@@ -1736,6 +1736,11 @@ def mul(d1, d2):
...
+                g = ((b - f)/f).expand()
+                _, d = g.leadterm(x)
             if not d.is_positive:
                 raise NotImplementedError()
```
jksuom, line 1745: This block could be indented in the same way as lines 1742-3. It is not necessary to repeat the test `not d.is_positive` when d is positive after line 1738.

praneethratna:
When I'm trying to add a test case, with the following change - Even though the answer is correct, I'm getting an `AssertionError`. @jksuom could you help me out here?
```diff
diff --git a/sympy/series/tests/test_series.py b/sympy/series/tests/test_series.py
index 34b598a415..f526b3bf0d 100644
--- a/sympy/series/tests/test_series.py
+++ b/sympy/series/tests/test_series.py
@@ -362,3 +362,11 @@ def test_issue_21245():
 def test_issue_21938():
     expr = sin(1/x + exp(-x)) - sin(1/x)
     assert expr.series(x, oo) == (1/(24*x**4) - 1/(2*x**2) + 1 + O(x**(-6), (x, oo)))*exp(-x)
+
+
+def test_issue_23432():
+    expr = 1/sqrt(1 - x**2)
+    assert expr.series(x, 0.5) == \
+        0.769800358919501 + 1.539600717839*(x - 0.5)**2 +\
+        2.39493444997178*(x - 0.5)**3 + 4.33369090947275*(x - 0.5)**4 +\
+        7.75502583800386*(x - 0.5)**5 + 0.769800358919501*x + O((x - 1/2)**6, (x, 1/2))
```

sympy/core/power.py
```diff
@@ -1735,12 +1735,10 @@ def mul(d1, d2):
...
         if not d.is_positive:
             g = g.simplify()
-            _, d = g.leadterm(x)
-            if not d.is_positive:
-                if g.is_zero:
+            if g.is_zero:
                     return f**e
-                g = ((b - f)/f).expand()
-                _, d = g.leadterm(x)
```
jksuom: The indentation of these lines (after `if not d.is_positive:`) should be preserved.

praneethratna: Do you mean something like this?
```diff
diff --git a/sympy/core/power.py b/sympy/core/power.py
index cefba7fe5c..f179650b97 100644
--- a/sympy/core/power.py
+++ b/sympy/core/power.py
@@ -1736,7 +1736,7 @@ def mul(d1, d2):
         if not d.is_positive:
             g = g.simplify()
             if g.is_zero:
-                    return f**e
+                return f**e
             g = ((b - f)/f).expand()
             _, d = g.leadterm(x)
             if not d.is_positive:
```

jksuom:
https://github.com/sympy/sympy/blob/a189657135dc9f6f33704c03e263f79ca575576b/sympy/core/power.py#L1735-L1739
Two changes: First, two lines between lines 1736 and 1737:

                  if g.is_zero:
                      return f**e
    
Second, 4 lines in place of 1739:

                      g = ((b - f)/f).expand()
                      _, d = g.leadterm(x)
                      if not d.is_positive:
                          raise NotImplementedError()

sympy/core/power.py
```diff
@@ -1735,9 +1735,13 @@ def mul(d1, d2):
                 raise NotImplementedError()
         if not d.is_positive:
             g = g.simplify()
-            _, d = g.leadterm(x)
```
jksuom, line 1738: `d` should be recomputed when it is not zero as it may be positive for the simplified `g`. This line should not be removed. It should come before the test below (line 1739 1740).

sympy/core/power.py
```diff
@@ -1735,11 +1735,11 @@ def mul(d1, d2):
                 raise NotImplementedError()
         if not d.is_positive:
             g = g.simplify()
+            _, d = g.leadterm(x)
```
jksuom, line 1738: There is no need to compute `d` if `g` is zero. Hence this this line is better placed after the next two lines immediately before the test.

sympy/core/power.py
```diff
@@ -1735,11 +1735,11 @@ def mul(d1, d2):
...
             if g.is_zero:
                 return f**e
             if not d.is_positive:
                 g = ((b - f)/f).expand()
-                _, d = g.leadterm(x)
```
jksuom: This line is needed when `g` has be recomputed.

jksuom:
> Even though the answer is correct, I'm getting an AssertionError.

Comparisons of expressions containing Floats are notoriously hard to implement. It seems that when a Float is printed, it is truncated, and when the printed expression is input and sympified, it will again be slightly modified so it will often not be exactly the same as the original Float. Comparison of printed strings is sometimes used as a work-around, but that should probably be avoided.

It is possible to use `almosteq` but that would make the test fairly complicated as all the Floats should be compared pairwise. In this case, I think that it would suffice to devise a test that would show that no error will be raised; for example, that `expr.series(x, 0.5)` is an Add.

praneethratna:
@jksuom I have added the test case - `assert expr.series(x, 0.5) != NotImplementedError` , but I'm not sure whether it is the right way or not.

jksuom:
The purpose of the tests is to give an indication of possible regressions when new code is added. `assert expr.series(x, 0.5) != NotImplementedError` would allow almost any changes to the series expansion; for example, it would pass even if an error of a different type were raised.
I think that a test of the following type would suffice in most cases:

    res = expr.series(x, 0.5)
    assert res.is_Add and len(res.args) == 7

jksuom:
Can you squash the commits? A single commit with a message would suffice.

jksuom:
Thanks, this is ready to go.

PR #23457, opened 2022-05-02
make `quantity_simplify()` able to simplify more complex expressions

### LINKED ISSUE #23452: Complex expressions with mixed units don't auto evaluate/cancel out.
When working with units, its fairly common to get a resulting expression that looks like this.

```
13140019104.1279*(-7209794853.0*ampere*ohm + 9125000000.0*volt)/(ampere*meter**3*ohm)
```

But it is equivalent to

```
2.51658322199041e+19/meter**3
```

 Basically, sympy can't cancel out equivalent units.

dyc3:
- add area, volume to si derived units
- fix(physics.units): quantity_symplify is now able to symplify more complex expressions

fixes #23452

`quantity_simplify()` and `convert_to()` struggle when the expression is not trivial. This is because the unit's dimensionality is not fully taken into consideration when simplifing expressions.
Additionally, `quantity_simplify()` struggles when the target dimension is not in the derived units map.

For example:
Both sides of this equality are equivalent dimensionally, but sympy doesn't think so because it sees these as 2 different expressions.
```
ohm*ampere == volt
```

This PR attempts to remedy this problem at least a little bit by providing a mechanism to convert the expression into a compound unit.

#### Other comments
I am not super confident about this solution. I would feel more comfortable if we had more test cases that were as complex as the one in #23452.

This PR is a work in progress. I'm looking for feedback on what direction to go in before I work on this any more.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/physics/units/systems/si.py
+++ b/sympy/physics/units/systems/si.py
@@ -18,7 +18,7 @@ from sympy.physics.units.definitions.dimension_definitions import (
     acceleration, action, current, impedance, length, mass, time, velocity,
     amount_of_substance, temperature, information, frequency, force, pressure,
     energy, power, charge, voltage, capacitance, conductance, magnetic_flux,
-    magnetic_density, inductance, luminous_intensity
+    magnetic_density, inductance, luminous_intensity, volume, area,
 )
 from sympy.physics.units.definitions import (
     kilogram, newton, second, meter, gram, cd, K, joule, watt, pascal, hertz,
@@ -81,6 +81,8 @@ SI = MKSA.extend(base=(mol, cd, K), units=all_units, name='SI', dimension_system
     current: ampere,
     voltage: volt,
     length: meter,
+    area: meter**2,
+    volume: meter**3,
     frequency: hertz,
     inductance: henry,
     temperature: kelvin,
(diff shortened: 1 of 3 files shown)


oscargus:
Looks promising! Regarding the mypy error, I do not really know how to get around it, probably some casting.

(I guess that apart from removing the prints, maybe shorten some lines, and this should be good to go? Although we do not have a strict limit there seems to be some possibilities to narrow most long lines down, especially comments.)

oscarbenjamin:
> Regarding the mypy error, I do not really know how to get around it, probably some casting.

There are two errors:
```
mypy sympy
sympy/physics/units/util.py:199: error: "Basic" has no attribute "name"
sympy/physics/units/util.py:221: error: Missing return statement
Found 2 errors in 1 file (checked 1455 source files)
```
The first error points out that `target_dimension.free_symbols` gives `set[Basic]` but it isn't guaranteed that a `Basic` instance has a `.name` attribute. Why is `free_symbols` being used there?

The second error points out that the `__find_compound_dimension` function is typed as returning `Dimension` but it doesn't look like it's guaranteed to return anything. Is it intended that the function would sometimes return None?

Also I think a single leading underscore is sufficient to note that this is an internal private function. Double underscore private functions/methods are not generally used in the SymPy codebase.

Also why are there print statements in the code? Those should be removed.

dyc3 (commented):
As I indicated above, this PR is a work in progress. It is very much not final, and not ready to be merged. I simply committed the code that I currently have. The reason for this is because I don't think I quite have the right solution, and would like feedback on how we could improve the simplification algorithm.

How does simplification for normal, regular old numeric expressions work? It's pretty robust, and I think we should make quantity_simplify work more like that.

Additionally, I would like to produce some more complex test cases.

PR #14664, opened 2018-04-26
Allow numpy to be import under pypy

extraymond:
Pypy has evolved greatly, it might be a good idea to remove the import ban.

#### References to other Issues or PRs

Fixes #14658

#### Brief description of what is fixed or changed

Remove import ban under pypy

PR #23461, opened 2022-05-04
Allow numpy to be import under pypy

### LINKED ISSUE #14658: Lambdify with numpy failed on pypy
Hi! I've try to use sympy lambdify to convert my code while running in pypy.
However the below error is raised:

```
can't import 'None' with 'import_module('numpy')' command
```

I do have numpy installed, just that using numpy as lambdify modules is not working.
BTW, if using cython as lambdify module it worked.

eagleoflqj:

Fixes #14658
Revive #14664

Please review commit by commit (e.g. checkout the first commit, reproduce the result I get, then checkout the next commit). I'll amend and force-push when needed.
## The 1st commit removes the import restriction
My local environment is Ubuntu 22.04 x64, pypy3 from apt.
I can't install scipy (build fail), so I test with numpy only.
```sh
$ pypy3 -m pytest sympy/utilities/tests/test_lambdify.py
...
FAILED sympy/utilities/tests/test_lambdify.py::test_integral - NameError: name 'scipy' is not defined
FAILED sympy/utilities/tests/test_lambdify.py::test_double_integral - NameError: name 'scipy' is not defined
```
It is strange since I have no scipy locally, why it tries to import scipy? If I run the `test_integral` only, it gives another error:
```sh
$ pypy3 -m pytest sympy/utilities/tests/test_lambdify.py::test_integral
...
FAILED sympy/utilities/tests/test_lambdify.py::test_integral - NameError: name 'Integral' is not defined
```
So the previous error is dependent with tests ran before.
## The 2nd commit fixes the scipy problem
A test runs before `test_integral` tries to import scipy. At line 149 it executes `import numpy` first, which succeed and updates `namespace`, which is the global dict `SCIPY`.
After that, `test_integral` tries to import scipy. As the `namespace` is modified, at line 132 the condition becomes `True` and the function successfully returns, giving a false conclusion that scipy is available.
So this commit re-arranges the order of imports. If scipy is not available, `SCIPY` isn't modified anymore, and tests run consistently.
https://github.com/sympy/sympy/blob/ed4e06eff860e001fd7c4da7cf17fd8c1cac8c6a/sympy/utilities/lambdify.py#L132-L150
Now testing `test_lambdify.py` will also lead to the undefined `Integral` exception.
## The 4th commit test optional dependencies for pypy
gmpy2, symengine, llvmlite and numba cannot be built. Skip them.
## The 5th commit fix the test error in the previous [PR](https://github.com/sympy/sympy/pull/14664#issuecomment-591618494)
Explanation can be found at https://stackoverflow.com/a/7396043.

#### Release Notes

* utilities
  * It is now possible to use `lambdify` with NumPy while running under PyPy.

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -133,17 +133,20 @@ jobs:
     needs: code-quality
 
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: ['3.9', 'pypy-3.8']
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v3
         with:
-          python-version: '3.9'
+          python-version: ${{ matrix.python-version }}
       - run: sudo apt install antlr4 libgfortran5 gfortran libmpfr-dev libmpc-dev
       - run: python -m pip install --upgrade pip
-      - run: pip install mpmath gmpy2 matplotlib numpy scipy aesara ipython \
-                         symengine cython llvmlite wurlitzer \
-                         autowrap numexpr 'antlr4-python3-runtime==4.7.*' \ 
-                         numba
+      - run: pip install mpmath matplotlib numpy scipy aesara ipython cython \
+                         wurlitzer autowrap numexpr 'antlr4-python3-runtime==4.7.*'
+      - if: ${{ matrix.python-version == '3.9' }}
+        run: pip install gmpy2 symengine llvmlite numba  # not available for pypy
       # Test external imports
       - run: bin/test_external_imports.py
       - run: bin/test_submodule_imports.py
--- a/.mailmap
+++ b/.mailmap
@@ -1396,6 +1396,7 @@ dranknight09 <cbhaavan@gmail.com>
 dustyrockpyle <dustyrockpyle@gmail.com>
 elvis-sik <e.sikora@grad.ufsc.br>
 erdOne <36414270+erdOne@users.noreply.github.com>
+extraymond <extraymond@gmail.com>
 faizan2700 <syedfaizan824@gmail.com>
 foice <foice.news@gmail.com>
 goddus <39923708+goddus@users.noreply.github.com>
--- a/sympy/utilities/lambdify.py
+++ b/sympy/utilities/lambdify.py
@@ -100,7 +100,7 @@ MODULES = {
     "math": (MATH, MATH_DEFAULT, MATH_TRANSLATIONS, ("from math import *",)),
     "mpmath": (MPMATH, MPMATH_DEFAULT, MPMATH_TRANSLATIONS, ("from mpmath import *",)),
     "numpy": (NUMPY, NUMPY_DEFAULT, NUMPY_TRANSLATIONS, ("import numpy; from numpy import *; from numpy.linalg import *",)),
-    "scipy": (SCIPY, SCIPY_DEFAULT, SCIPY_TRANSLATIONS, ("import numpy; import scipy; from scipy import *; from scipy.special import *",)),
+    "scipy": (SCIPY, SCIPY_DEFAULT, SCIPY_TRANSLATIONS, ("import scipy; import numpy; from scipy import *; from scipy.special import *",)),
     "cupy": (CUPY, CUPY_DEFAULT, CUPY_TRANSLATIONS, ("import cupy",)),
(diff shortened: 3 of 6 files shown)


eagleoflqj:
It took more than 20 min to build packages, and gmpy2 is incompatible with pypy. Skip that.

eagleoflqj:
Symengine should also be skipped.

.mailmap
```diff
@@ -1396,6 +1396,7 @@ dranknight09 <cbhaavan@gmail.com>
 dustyrockpyle <dustyrockpyle@gmail.com>
 elvis-sik <e.sikora@grad.ufsc.br>
 erdOne <36414270+erdOne@users.noreply.github.com>
+extraymond <extraymond@gmail.com>
```
oscarbenjamin, line 1399: @extraymond, how do you want your name to be recorded in the AUTHORS file?

asmeurer:
The changes here all make sense to me. For the CI, I don't know if it's necessary to have a separate optional dependencies build just for pypy. We can install the optional dependencies that work with pypy in the normal pypy build. The point of the optional dependencies build is to make sure we test everything without any optional dependencies installed in the normal tests, but I don't think that's important to do for pypy specifically.

.mailmap
extraymond: Thanks for mentioning it, it's my pleasure to be included.

eagleoflqj: > Thanks for mentioning it, it's my pleasure to be included.

Oscar means, whether you want your real name shown, or just github user name. And if real name, what is it.

extraymond: Oh! Handle should be fine.

eagleoflqj:
I make a separate CI job for this because
* Pypy tests are much slower than their cpython counterpart. We split test1 and test2 even for cpython because running as a whole is too long.
* We can compare optional-dependencies-pypy with optional-dependencies, so that we know what dependencies are disabled for pypy, and try adding a package back in the future if it's supported by pypy.

Let test cases covered by test1+test2 be set A, test cases covered by test_optional_dependencies.py be set B. I don't know the relationship between set A and B. If B is a subset of A, then we may run A with optional dependencies. If not, we need this new CI job.

oscarbenjamin:
I think it's reasonable to add this as an extra job.

eagleoflqj:
I'm going to merge it in order to fix the issue. Feel free to remove the job if it becomes a burden.

sympy/utilities/lambdify.py
```diff
@@ -100,7 +100,7 @@
     "math": (MATH, MATH_DEFAULT, MATH_TRANSLATIONS, ("from math import *",)),
     "mpmath": (MPMATH, MPMATH_DEFAULT, MPMATH_TRANSLATIONS, ("from mpmath import *",)),
     "numpy": (NUMPY, NUMPY_DEFAULT, NUMPY_TRANSLATIONS, ("import numpy; from numpy import *; from numpy.linalg import *",)),
-    "scipy": (SCIPY, SCIPY_DEFAULT, SCIPY_TRANSLATIONS, ("import numpy; import scipy; from scipy import *; from scipy.special import *",)),
+    "scipy": (SCIPY, SCIPY_DEFAULT, SCIPY_TRANSLATIONS, ("import scipy; import numpy; from scipy import *; from scipy.special import *",)),
```
oscarbenjamin, line 103: Why is this change needed?

.github/workflows/runtests.yml
```diff
@@ -151,6 +151,29 @@ jobs:
       # Test modules with specific dependencies
       - run: bin/test_optional_dependencies.py
 
+  # -------------------- Optional dependency tests for pypy ----------------- #
+
+  optional-dependencies-pypy:
+    needs: code-quality
```
oscarbenjamin, line 157: Can this be added using `matrix:` rather than as a separate listing?

The only difference with the job above is that pip doesn't install quite as many things but we could guard that with a conditional.

sympy/utilities/lambdify.py
eagleoflqj: See my description for the 2nd commit. If the environment has numpy but doesn't have scipy, `test_lambdify.py` produces strange behavior. It's the case for both cpython and pypy.

.github/workflows/runtests.yml
eagleoflqj: I'll read the GitHub Actions doc to see how to achieve that.

eagleoflqj:
@oscarbenjamin Required tests should be changed.

PR #23462, opened 2022-05-04
Solving main page

bertiewooster:
# Purpose
This is the main (landing) solving page for the [Season of Docs 2022 project](https://github.com/sympy/sympy/wiki/Season-of-Docs-2022-Organization-Application). The goal is to

- direct traffic to this page on docs.sympy.org from web searches and visitors to the [guides page](https://docs.sympy.org/dev/guides/index.html#guides)
- help visitors find the appropriate sub-page (pages to be made later) that will "solve" their type of problem.

"Solving" is defined broadly, in accord with how searchers use the term, for example "solve integral". By contrast, in SymPy, "solving" has the specific meaning “find the roots of an equation.”

# Page status and feedback sought
This draft pull request has all the intended text of the final page, including link texts. However, the links currently lead nowhere; I will fill their targets in after addressing any feedback on the content of this page.

 At this point, I request feedback on

- Page title
- Page text blocks
- Links to solving sub-pages: link text, organization, etc.

# Methodology
The solving topics and ordering were [derived from Google Search Console data as shown in this analysis](http://www.sympy.org/sympy-search-analytics/search-terms.html). Briefly, all search terms containing "solv" in a year's worth of data were categorized into 15 types designed to represent searchers' conception the type of solving requested, for example ordinary differential equation(s), inequality, integral, etc. The categories were then ordered by the number of impressions (times SymPy documentation page results were presented to Google searchers) to represent how often those terms lead Google to suggest that searchers visit SymPy's documentation site.

# Search engine optimization
To encourage search engines to rank this page highly (present it near the top of the search results), the page contains verbose, relevant terms. For example, the title is "Solve Using **Python** SymPy" because searchers may know that they need a way to solve an equation in Python, and be unaware of SymPy, as evidenced by search terms such as "**python** solve equation".

Similarly, each link is a verbose description of the type of solving, for example "Solve an equation algebraically", because

- search engines use the link text to help determine the topic of the target page
- page visitors perceive link texts as avenues to go down.

(We might consider even more verbose links like "Solve an equation algebraically **using Python SymPy**".)

Added Solving main page.

#### Release Notes

* other
  * A new, user-friendly page on solving of various types in SymPy was added to the Guides page.


asmeurer:
Comments from our meeting today:

- The title should be human readable. The current form "Solve Using Python SymPy" is ungrammatical. Something like "the SymPy Python library" may be better.
- The page titles already include "SymPy 1.10.1 documentation" automatically from Sphinx. So including "SymPy" again may be unnecessary.
- The links currently go nowhere. We will link them to the best existing pages for each.
- The order of the links is based on the search console data. I suggested looking into whether adding a short example for each page would make it easier to navigate.

doc/src/guides/solving.md
```diff
@@ -0,0 +1,25 @@
+# Solve Using Python SymPy
+
+Python package SymPy can symbolically solve equations, differential equations,
```
asmeurer, line 3: "*The* Python package..."

doc/src/guides/solving.md
```diff
@@ -0,0 +1,25 @@
...
+- [Solve an equation numerically]()
+- [Solve a partial differential equation algebraically]()
+- [Solve ordinary differential equations algebraically]()
+- [Solve a system of linear equations algebraically]()
+- [Solve a system of nonlinear equation algebraically]()
+- [Solve a matrix problem algebraically]()
+- [Solve an inequality]() or [solve a system of inequalities]() algebraically
+- [Solve a polynomial algebraically]()
+- [Solve a diophantine equation algebraically]()
+- [Solve an integral symbolically]()
```
asmeurer, line 19: In the glossary I wrote

>   SymPy generally uses the word "solve" to mean equation solving in this
    sense. It is not used in the sense of "solving a problem". For instance,
    one would generally prefer to use a word like "compute" or "evaluate" to
    refer to the computation of an integral using a function like
    integrate.

https://github.com/sympy/sympy/pull/23293/files#diff-f3a7e8594f4f56ce385b5d650df921ee5cb516cc7b94db0b0c42679a70bdd6a8R558

So I think we should stick with that and avoid using "solve" to mean integration here. We can also add some similar text to this page to help people understand this distinction.

bertiewooster: That makes sense from a SymPy nomenclature standpoint.

To check how that might affect search engine optimization (SEO) and presumably visitor [information architecture](https://www.usability.gov/what-and-why/information-architecture.html) (how content is organized, structured, and labeled so visitors can find what they need), I investigated further the [year's worth of Google Search Console data](https://docs.google.com/spreadsheets/d/1zuQDU1Mj8Ua8zAAxRu5hr7b23wCRlIJ2cOK6FPMnJq0/edit?usp=sharing) and found that "sympy evaluate integral" has somewhat more impressions than "sympy solve integral", 146 vs. 103, so "evaluate" is also preferable from an SEO standpoint. (There are no search terms in the top 1000 including "compute" and "integral".)

So I'll change "Solve an integral" to "Compute an integral".

I reviewed the list of solving types and didn't notice any others where "solve" conflicts with the definition in the glossary. If anyone believes there are others, please let me know.

gutow: I am confused. Wouldn't it help with searching and be more consistent with common usage, while also being consistent with Sympy vocabulary to do something like this:

"Evaluate (solve) an integral". 

Why use a term (compute) that people use rarely?

bertiewooster: Oops! Thanks. I had "compute" on the brain. I changed it to "evaluate" in the pull request, and edited my comment above so someone reviewing my comment later won't notice the same mistake.

bertiewooster:
As @asmeurer suggested, added example mathematical formulas for each solving type. Thought they looked best in a table because width of each description and example may be different.

Request feedback on the layout of the table with description and example columns. Here's a preview where the table width is about the body section width on site as viewed on a desktop, so the table row widths should be about accurate--each solving type seems to fit in one row, no wrapping required. (I'm getting an error when building docs so this is not in SymPy documentation theme yet.)

SymPy-solving-page-with-table-descr-example

Issues to address later:

- Couldn't get a matrix to format correctly in Markdown
- The math of the examples may be nonsense; I focused on the formatting
- The formulas are currently surrounded simply with dollar signs e.g. `$\int f\,dx$`; I believe I should be using `{math}` but couldn't get that to work yet

asmeurer:
Dollar signs are the correct way to do latex math in Markdown, or two dollar signs for equation math.

bertiewooster:
Do folks agree that making a table (columns for a) textual description and b) mathematical formula) is a good way to present the solving types?

I'm encountering some technical difficulties in rendering a table in Markdown that includes links. I think there are ways to solve the difficulties (for example, by adding Sphinx extension [sphinx-markdown-tables](https://pypi.org/project/sphinx-markdown-tables/)); if people think a table is a good format, I'll pursue them.

asmeurer:
Does sphinx-markdown-tables work with myst? The docs mention recommonmark, which works differently from myst. I would ask the myst community for help with myst. Or if it ends up being easier in RST, we can just use RST for the index page.

asmeurer:
Or if a table is too hard, we can present it as the links as headers with a short example under each header.

bertiewooster:
Turns out the problem was simply that Sphinx insists there be a target for the link: `[link text]()` raises an error whereas `[link text](http://docs.sympy.org)` is fine. So the table format seems feasible.

All the content should be there now with these caveats:

- Some of the math may be nonsense
- The link targets have not been set individually, they all just point to http://docs.sympy.org for now

moorepants:
This looks great. What is the different in "solve system of equations algebraically" and "solve a system of linear equations algebraically"? It seems they both have the same linear style example equations.

bertiewooster:
@moorepants thanks and good point. I made the example for "Solve system of equations algebraically" have one nonlinear and one linear equation. We're basing the categories on the type of terms that people search for, so there may be some overlap between categories.

solving-main-page-3

doc/src/guides/solving.md
```diff
@@ -0,0 +1,30 @@
...
+
+Learn how to use SymPy computer algebra system to:
+
+| Description                                                  | Example                                                                                                                     |
+|--------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------|
+| [ Solve an equation algebraically ](http://docs.sympy.org)                        | [  $x^2 - 1 = 0$  ](http://docs.sympy.org)                                                                                                       |
+| [ Solve a system of equations algebraically ](http://docs.sympy.org)              | [  $x^2 + y = 1, x - y = 0$  ](http://docs.sympy.org)                                                                                              |
+| [ Solve an equation numerically ](http://docs.sympy.org)                          | [ $x^2 + x = 0.274$ ](http://docs.sympy.org)                                                                                                     |
```
asmeurer, line 13: I would suggest using an equation that doesn't have a closed-form solution (the example I like to use is cos(x) = x)

asmeurer:
I think we decided that the links here should link to the best documentation that currently exists (in most cases, the docstring of `solve` or another function).

asmeurer:
If there's room, maybe we can show an example equation and the solutions, like `x^2 = 2 -> [-sqrt(2), sqrt(2)]`.

doc/src/guides/solving.md
```diff
@@ -0,0 +1,30 @@
...
+| [ Solve a system of inequalities algebraically ](http://docs.sympy.org)           | [ $ x^2 + x > 0, x < 1$ ](http://docs.sympy.org)                                                                                                 |
+| [ Solve a polynomial algebraically ](http://docs.sympy.org)                       | [ $ 4x^3 - 3x^2 + x = 0$ ](http://docs.sympy.org)                                                                                                |
+| [ Solve a diophantine equation algebraically ](http://docs.sympy.org)             | [ $ax^2 + bxy + cy^2 + dx + ey + f = 0$ ](http://docs.sympy.org)                                                                                 |
+| [ Evaluate an integral symbolically ](http://docs.sympy.org)                      | [ $\int f(x)\,dx$ ](http://docs.sympy.org)                                                                                                          |                                                                |
```
asmeurer, line 23: Let's replace `f(x)` with an explicit function. Otherwise, this might appear to be a page for solving integral equations, which is a different thing (which SymPy doesn't really have functions to do yet).

doc/src/guides/solving.md
bertiewooster: Great, I just used that example in my latest commit. (Are we sure HackerRank hasn't copyrighted that example? ;-)

bertiewooster:
> I think we decided that the links here should link to the best documentation that currently exists (in most cases, the docstring of `solve` or another function).

Yep, now that it seems like there's consensus on the format and content of the "solving" types, my next task is to seek the best existing documentation for each.

doc/src/guides/solving.md
bertiewooster: Thanks, good idea. Done.

bertiewooster:
> If there's room, maybe we can show an example equation and the solutions, like `x^2 = 2 -> [-sqrt(2), sqrt(2)]`.

I'll start trying this out. Hopefully it won't take up significantly more vertical space. If it does, I'll ask for feedback on whether it's worth including the solutions.

bertiewooster:
## Include solutions in table?
I added solutions for each type of "solving." That does increase the height of the table noticeably, from 766 to 1246 pixels. As I review the overall effect, I think it makes the table harder to scan for the type of problem a visitor is searching for. (I suppose we could mitigate that by adding another column for solutions, but that would add more vertical space.) So I'm favoring leaving the solutions off of this page, and considering putting them near the top of the sub-pages to give the visitor confidence that the page will lead to the type of solution they expect. Thoughts?

<details>
  <summary>Without solutions</summary>
  
solving-no-solns
</details>

<details>
  <summary>With solutions</summary>
  
solving-with-solns
</details>

## Type of matrix problem to give
The one type I am unclear what kind of example we should use is matrix, currently described as "Solve a matrix problem algebraically". The top Google search query for "solv*" and "matrix" is "sympy solve matrix equation", and the other two are "sympy matrix solve" and "sympy solve matrix." Do folks think that means [solving a system of linear equations in matrix form](https://www.varsitytutors.com/hotmath/hotmath_help/topics/solving-systems-of-linear-equations-using-matrices.html)?

bertiewooster:
## Include solutions in table?
Just temporarily, I'm including both versions of the table, one with solutions and one without, so folks can compare. If no one gives feedback soon, I'll go with the no-solutions version, and plan to put each solution on its problem type's sub-page.

By the way, I'd forgotten to include one equation type in the no-solutions table screenshot, so that table's height is slightly taller than the measurement above.

## Type of matrix problem to give
I checked [StackOverflow for "matrix solve" tagged SymPy with high scores](https://stackoverflow.com/search?tab=votes&q=%5bsympy%5d%20matrix%20solve) and high view. Unfortunately, there's no consistent type of problem people ask about: questions include

- system of linear equations
- matrix = 0
- M * x + N * y = 0 (M,N = matrices, x,y = vectors)
- finding the eigenvalues and eigenvectors of a matrix

so I think it makes sense to go with system of linear equations on the main page. Perhaps on the matrix solving sub-page (to be created later) we could include multiple problem types.

doc/src/guides/solving.md
```diff
@@ -0,0 +1,45 @@
...
+| [ Solve a matrix problem algebraically ](http://docs.sympy.org)                   | [ $ \left[\begin{array}{cc} 1 & 1\\1 & -1\end{array}\right] \left[\begin{array}{cc} x\\y\end{array}\right] = \left[\begin{array}{cc} 2\\0\end{array}\right] $ ](http://docs.sympy.org) |
+| [ Solve an inequality algebraically ](http://docs.sympy.org)                      | [ $ x^2 < 4 $ ](http://docs.sympy.org)                                                                                                        |
+| [ Solve a system of inequalities algebraically ](http://docs.sympy.org)           | [ $ x^2 < 4, x > 0 $ ](http://docs.sympy.org)                                                                                                 |
+| [ Solve a polynomial algebraically ](http://docs.sympy.org)                       | [ $ x^2 - x = 0 $ ](http://docs.sympy.org)                                                                                                |
+| [ Solve a diophantine equation algebraically ](http://docs.sympy.org)             | [ $x^2 - 4xy + 8y^2 - 3x + 7y - 5 = 0 $ ](http://docs.sympy.org)                                                                                 |
```
oscarbenjamin, line 38: It might be more intuitive to describe this as "find integer solutions to an equation".

doc/src/guides/solving.md
```diff
@@ -0,0 +1,45 @@
...
+| [ Evaluate an integral symbolically ](http://docs.sympy.org)                      | [ $\int 2x\,dx 	$ ](http://docs.sympy.org)                                                                                                          |                                                                |
```
oscarbenjamin, line 39: I don't really think this belongs here.

doc/src/guides/solving.md
```diff
@@ -0,0 +1,45 @@
...
+| [ Solve a system of nonlinear equations algebraically ](http://docs.sympy.org)    | [  $x^2 + y^3 = 1, x^3 - y^2 = 0 $  ](http://docs.sympy.org)                                                                                      |
+| [ Solve a matrix problem algebraically ](http://docs.sympy.org)                   | [ $ \left[\begin{array}{cc} 1 & 1\\1 & -1\end{array}\right] \left[\begin{array}{cc} x\\y\end{array}\right] = \left[\begin{array}{cc} 2\\0\end{array}\right] $ ](http://docs.sympy.org) |
+| [ Solve an inequality algebraically ](http://docs.sympy.org)                      | [ $ x^2 < 4 $ ](http://docs.sympy.org)                                                                                                        |
+| [ Solve a system of inequalities algebraically ](http://docs.sympy.org)           | [ $ x^2 < 4, x > 0 $ ](http://docs.sympy.org)                                                                                                 |
+| [ Solve a polynomial algebraically ](http://docs.sympy.org)                       | [ $ x^2 - x = 0 $ ](http://docs.sympy.org)                                                                                                |
```
oscarbenjamin, line 37: I would use the word roots here: "Find roots of a polynomial".

oscarbenjamin: I think it would also be good to have a page that discusses solving systems of polynomials with Groebner bases.

oscarbenjamin:
The type of matrix problem should be solving a linear system of equations but as compared to the linear solvers page you could maybe discuss more about determinants, inverse, nullspace, LU decomposition etc since these all relate to solving linear systems of equations.

doc/src/guides/solving.md
asmeurer: Agree. People might not know what "Diophantine" means, although it is a good idea to tell them since that is the term used in SymPy.

bertiewooster: Here are the [search queries in the top 1000](https://docs.google.com/spreadsheets/d/1zuQDU1Mj8Ua8zAAxRu5hr7b23wCRlIJ2cOK6FPMnJq0/edit?usp=sharing) involving _diophantine_ or _integer_:

Top queries | Clicks | Impressions | Click-through rate | Position
-- | -- | -- | -- | --
_diophantine_ equation python | 235 | 662 | 35.50% | 1.06
sympy _integer_ | 222 | 387 | 57.36% | 1.1
sympy _diophantine_ | 193 | 292 | 66.10% | 1
python _diophantine_ equation | 63 | 165 | 38.18% | 1.07
_diophantine_ equation solver python | 53 | 131 | 40.46% | 1.01

In total, there are 1250+ searches a year (3.4+ per day) involving _diophantine_, so I think it's well worth including that term here. Here's a revised description to tell visitors what Diophantine means:

`Solve a Diophantine equation (find integer solutions to a polynomial equation) algebraically`

doc/src/guides/solving.md
bertiewooster: The only search query in the top 1000 containing _groebner_ is

Top queries | Clicks | Impressions | CTR | Position
-- | -- | -- | -- | --
sympy _groebner_ basis | 60 | 76 | 78.95% | 1

So while Groebner bases are conceptually important, I'm not sure there's enough Google searches (which is a proxy for overall interest because we don't collect site analytics) for them to justify including them on this page, or at least in the tranche of types planned for the Season of Docs project. A page discussing the technique may well be useful, though again is probably beyond the scope of this project.

bertiewooster: Here are the [search queries in the top 1000](https://docs.google.com/spreadsheets/d/1zuQDU1Mj8Ua8zAAxRu5hr7b23wCRlIJ2cOK6FPMnJq0/edit?usp=sharing) involving _polynomial_, with queries involving **solve** or **roots** bolded:

Top queries | Clicks | Impressions | CTR | Position
-- | -- | -- | -- | --
polynomial manipulation | 240 | 1763 | 13.61% | 2.88
sympy polynomial | 1079 | 1574 | 68.55% | 1
sympy polynomial coefficients | 155 | 406 | 38.18% | 1.35
sympy polynomial division | 254 | 365 | 69.59% | 1
**sympy roots of polynomial** | 238 | **365** | 65.21% | 1.04
polynomial manipulation in data structure | 74 | 344 | 21.51% | 4.34
sympy characteristic polynomial | 213 | 336 | 63.39% | 1
sympy factor polynomial | 157 | 330 | 47.58% | 1.03
**sympy solve polynomial** | 218 | **300** | 72.67% | 1
sympy polynomials | 171 | 242 | 70.66% | 1
sympy factorize polynomial | 162 | 219 | 73.97% | 1
sympy expand polynomial | 126 | 211 | 59.72% | 1
sympy coefficients of polynomial | 93 | 175 | 53.14% | 1.18
**sympy polynomial roots** | 90 | **144** | 62.50% | 1.28
sympy polynomial from coefficients | 52 | 138 | 37.68% | 1.26

So there are more searches involving **roots** than **solve**, 509 vs. 365. Given those two terms are used with fairly comparable frequency, I think a description including both is worthwhile from the perspective of both search engine optimization and user recognition:

`Solve (find the roots of) a polynomial algebraically`

asmeurer: "Groebner" is somewhat analogous to the Diophantine/integer equations situation. It's the technical name for the algorithm used to solve systems of polynomial equations. It might be helpful to include "Groebner" in the title text, but the main thing people will look for is systems of equations.

Also, in my experience, people won't necessarily recognize that they have a "system of polynomial equations" vs. just a "system of equations". It partially depends on how mathematically sophisticated they are. But also SymPy itself makes the distinction because the methods used to solve them are different, but people who aren't familiar with those methods might not make the same distinctions. For example, Groebner bases make it straightforward to solve a system of *polynomial* equations, but if the system includes a non-polynomial equation it is much harder because the algorithm no longer applies. But people may not appreciate that and might not even realize that their system is in fact non-polynomial. These are the sorts of caveats we will have to include in the solvers guides themselves.

oscarbenjamin: I often see polynomial equations being described as "nonlinear". It might be good to clarify when talking about nonlinear equations that polynomials are an important special case.

doc/src/guides/solving.md
```diff
@@ -0,0 +1,45 @@
+# Solve Using the SymPy Python Library
```
asmeurer, line 1: I know we chose this name for SEO purposes, but this rather verbose title also shows up in the sidebar. I don't know if there's a way to make the one different from the other. I agree it's useful to have this longer name for the title of the page in the Google search results, but once you are already in the SymPy documentation, it actually makes it harder to navigate because of the verbosity.

doc/src/guides/solving.md
```diff
@@ -0,0 +1,45 @@
...
+| [ Solve a system of equations algebraically ](http://docs.sympy.org)              | [  $x^2 + y = 2, x - y = 4 \to [(x = -3, y = -7), (x = 2, y = 2)]$  ](http://docs.sympy.org)                                                                                              |
+| [ Solve an equation numerically ](http://docs.sympy.org)                          | [ $cos(x) = x \to x \approx 0.739085133215161$ ](http://docs.sympy.org)                                                                                                     |
+| [ Solve a partial differential equation algebraically ](http://docs.sympy.org)    | [${\frac {\partial ^{2}(x^3 y^3)}{\partial x\partial y}}=9 \to x \ne 0, y = \pm\frac{1}{x} $](http://docs.sympy.org)                                                                     |
+| [ Solve an ordinary differential equation algebraically  ](http://docs.sympy.org) | [ $f\prime\prime(x) + 9f(x)=0 \to f(x)=C_{1} sin(3x)+ C_{2} cos(3x)$ ](http://docs.sympy.org)                                                    |
```
asmeurer, line 15: This should just be `'` instead of `\prime`. See https://tex.stackexchange.com/questions/87134/what-is-the-advantage-of-using-f-prime-instead-of-f. If you do use `\prime`, it should be `^\prime`. Right now the prime symbols are too low and look odd.

asmeurer: This should be `\sin` and `\cos`

doc/src/guides/solving.md
```diff
@@ -0,0 +1,45 @@
...
+
+| Description                                                  | Example                                                                                                                     |
+|--------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------|
+| [ Solve an equation algebraically ](http://docs.sympy.org)                        | [  $x^2 = 4 \to x =[-2, 2] $  ](http://docs.sympy.org)                                                                                                       |
+| [ Solve a system of equations algebraically ](http://docs.sympy.org)              | [  $x^2 + y = 2, x - y = 4 \to [(x = -3, y = -7), (x = 2, y = 2)]$  ](http://docs.sympy.org)                                                                                              |
+| [ Solve an equation numerically ](http://docs.sympy.org)                          | [ $cos(x) = x \to x \approx 0.739085133215161$ ](http://docs.sympy.org)                                                                                                     |
```
asmeurer, line 13: `\cos`

doc/src/guides/solving.md
```diff
@@ -0,0 +1,45 @@
...
+| [ Solve a partial differential equation algebraically ](http://docs.sympy.org)    | [${\frac {\partial ^{2}(x^3 y^3)}{\partial x\partial y}}=9 \to x \ne 0, y = \pm\frac{1}{x} $](http://docs.sympy.org)                                                                     |
```
asmeurer, line 14: This isn't really a PDE. A PDE would have an unknown function of multiple variables, similar to an ODE. SymPy's PDE support isn't very good, so for an example you may need to copy one from the docs just to be sure that SymPy can actually solve it (`pdsolve`). 

Alternatively, we can just omit PDEs from here for now, since SymPy's symbolic PDE solving support is so weak, especially for the sorts of PDEs people might actually want to solve analytically.

asmeurer:
Some things we discussed in the call today:

- We talked a bit about how things like "solve integrals" should appear on the page. Some of my thoughts were that we can move it from the table but include a short paragraph explaining how "solve" is used in the SymPy and pointing to other parts of the docs for things like integration. I also discussed that we can iterate on what exactly is here once it is live and we get updated search data. @bertiewooster stressed the importance of getting people to the right place from their search terms (I'm probably not doing your arguments justice here, so feel free to comment further).
- We should aim to get this merged by the end of the month to stay on schedule for the season of docs project.
- We need to replace all the existing links with links to the best existing documentation.
- We discussed whether the table should include solutions. One concern is that the table in its current form makes the leftmost column too narrow. A possible fix for this is to put the solutions in separate math environments, or even on a separate line or separate column (we will need to try all of these to see which renders the best). If the table still squishes the columns too much, we can report it upstream to Furo, since most likely the fix would involve improving the Furo CSS.
- I requested that the math not be links, just the text. It's not obvious that the math is a link, and it makes everything too green and harder to read.

It wasn't brought up at the meeting, but another comment I have is that we should verify that each solution in the table is correct and that SymPy can actually solve it.

doc/src/guides/solving.md
```diff
@@ -0,0 +1,45 @@
...
+
+The Python package SymPy can symbolically solve equations, differential equations, 
+linear equations, nonlinear equations, matrix problems, inequalities, 
+Diophantine equations, and evaluate integrals. SymPy can also solve numerically.
+
+Learn how to use SymPy computer algebra system to:
+
+| Description                                                  | Example                                                                                                                     |
+|--------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------|
+| [ Solve an equation algebraically ](http://docs.sympy.org)                        | [  $x^2 = 4 \to x =[-2, 2] $  ](http://docs.sympy.org)                                                                                                       |
```
asmeurer, line 11: Since we're using math notation, let's use set notation `\{-2, 2\}` so that this doesn't look like an interval.

doc/src/guides/solving.md
asmeurer: Also let's use `y'` instead of `f'(x)` so that it matches what people are more likely to encounter in a textbook.

doc/src/guides/solving.md
bertiewooster: Thanks--this exceeds my level of math education :-) This is the example I found on the [`pdsolve` page](https://docs.sympy.org/dev/modules/solvers/pde.html?highlight=pdsolve#sympy.solvers.pde.pdsolve):

```
>>> from sympy.solvers.pde import pdsolve
>>> from sympy import Function, Eq
>>> f = Function('f')
>>> u = f(x, y)
>>> ux = u.diff(x)
>>> uy = u.diff(y)
>>> eq = Eq(1 + (2*(ux/u)) + (3*(uy/u)), 0)
pdsolve(eq)
```

Is there a way to express this in one line of math, or at least fewer than four?

doc/src/guides/solving.md
bertiewooster: I shortened to _Solve Using SymPy_. The word _Python_ is at the top of the page, so hopefully search engines will take that into account.

Or if people are OK with the grammar, _Solve Using Python's SymPy_ would be a compromise.

doc/src/guides/solving.md
asmeurer: Let's just nix the PDE section for now. If we want to include it we should take a look at stackoverflow and other sources to get an idea of common problems that people are trying to solve, and whether SymPy can actually solve them or not.

bertiewooster: OK, I'll remove for now. PDE is the [fourth most common category of solving type](http://www.sympy.org/sympy-search-analytics/search-terms.html) (16215 Google search impressions per year = 44 per day), so we should try to add it back in later. The [common types seem to be those used for physics applications (heat, wave, etc.)](https://en.wikipedia.org/wiki/Partial_differential_equation#See_also).

A specific example is [heat flow in a uniform rod](https://en.wikipedia.org/wiki/Heat_equation#Heat_flow_in_a_uniform_rod). If someone with more math expertise could formulate and solve that, or another common and simple example, in SymPy, I'd appreciate it and would add it to this landing page and a PDE sub-page.

bertiewooster:
> we should verify that each solution in the table is correct and that SymPy can actually solve it
Yes, I've used SymPy to solve all the examples in the table, except PDE (which I'll remove for other reasons), in a Jupyter notebook. I can't think of a way to tie that to the page like a doctest; if there is, please let me know. I'm happy to commit the Jupyter notebook to the SymPy repo if that's appropriate, or pull its code out into a .py file and commit that.

> @bertiewooster stressed the importance of getting people to the right place from their search terms
Yes, the best way to ensure that search engines present the SymPy docs link first (as opposed to other sites) is to use the words that searchers use, for example "sympy solve integral". Having noted that, integral is the only "solving" type in the categories we identified that we've decided shouldn't be in the table, so I'm fine with calling _integral_ out in a paragraph outside the table and linking to the best existing integral page.

doc/src/guides/solving.md
bertiewooster: Here are the search queries in the top 1000 that involve integral:

Top queries | Clicks | Impressions | CTR | Position
-- | -- | -- | -- | --
integral python | 82 | 5931 | 1.38% | 9.28
python integral | 53 | 4522 | 1.17% | 9.8
sympy integral | 2222 | 3163 | 70.25% | 1.03
sympy definite integral | 568 | 1151 | 49.35% | 1.08
integral sympy | 727 | 1035 | 70.24% | 1
indefinite integral python | 210 | 721 | 29.13% | 1.77
sympy double integral | 220 | 704 | 31.25% | 1.79
python indefinite integral | 194 | 665 | 29.17% | 1.73
integral python sympy | 305 | 453 | 67.33% | 1.01
definite integral sympy | 202 | 399 | 50.63% | 1.03
sympy integrals | 246 | 382 | 64.40% | 1.01
sympy indefinite integral | 119 | 193 | 61.66% | 1.01
sympy evaluate integral | 87 | 146 | 59.59% | 1.01
sympy solve integral | 69 | 103 | 66.99% | 1

So to the extent that searchers use a verb, it is _solve_. However, most searchers don't use a verb.

bertiewooster:
I tried moving the solutions to a new column and believe it's an improvement on both

- desktop, because the example columns is easier to scan visually
- mobile, because the example column fits on the screen (at least of my iPhone 13 mini), and visitors can scroll horizontally to get to the solution column

The only downside of the three-column format is that, on mobile, the solution column is off the screen. I don't think that's a big problem because the solutions are nice to have rather than essential for visitors finding their desired solving type.

So I suggest using the three-column table.

<details>
  <summary>Three columns</summary>
 
Desktop
three-col-table-desktop

Mobile
three-col-table-mobile

</details>

<details>
  <summary>Two columns</summary>

Desktop
two-col-table-desktop

Mobile
two-col-table-mobile

</details>

bertiewooster:
I'm starting to replace the placeholder links with links to the best available content. I'm getting an error when I try to include an anchor in the target; for example, this works:

`[ Solve a system of equations algebraically ](../modules/solvers/solvers.md)`

whereas this gives error `make: *** [Makefile:62: html] Error 1`
`[ Solve a system of equations algebraically ](../modules/solvers/solvers.md#algebraic-equations)`

when I execute `docker run --rm -v /Users/jemonat/Projects/sympy:/sympy sympy_htmldoc`.

The only tip I found was in Myst's documentation [Markdown link resolution improvements](https://myst-parser.readthedocs.io/en/latest/develop/_changelog.html?highlight=anchor#markdown-link-resolution-improvements)

> If header-anchors are enabled, anchor links are also supported, e.g. `[text](file.md#anchor)`

Our `doc/src/conf.py` already included `myst_heading_anchors = 2`; I tried increasing it to 3 and 5 just to make sure the depth of anchors wasn't an issue. I also added to `myst_enable_extensions` this: `"heading_anchors"` to `doc/src/conf.py`. Neither solved the problem.

Any ideas?

asmeurer:
If you want to link to a heading in RST you need to add a xref before the heading, like

```
.. _heading-name:

Heading
=======
```

(in Markdown this would be `(heading-name)=`). Then you can just link to `heading-name`. Note the extra underscore in the xref. The reference names should be globally unique so should be named something like `solvers-heading-name`.

bertiewooster:
@asmeurer thanks; I'm not quite following for Markdown though. For example, I want to link to https://docs.sympy.org/dev/modules/solvers/solvers.html#algebraic-equations. I tried the following and all gave errors:

- [ Solve a system of equations algebraically ](#algebraic-equations)
- [ Solve a system of equations algebraically ](#algebraic-equations=)
- [ Solve a system of equations algebraically ](algebraic-equations=)
- [ Solve a system of equations algebraically ](algebraic-equations)=)
- [ Solve a system of equations algebraically ](heading-name)=algebraic-equations)

Could you give what the markdown should be for this situation?

asmeurer:
"heading-name" was just an example. You would replace that with the actual shorthand text for the given heading. If you look in the RST file you can see other examples.

For example, you might have in `solvers.rst`

```
.. _solvers-algebraic-equations:

Algebraic Equations
===================
```

And then the link to it would be `[Solve a system of equations algebraically](solvers-algebraic-equations)`. 

If `solvers.rst` were in Markdown, you would use

```
(solvers-algebraic-equations)=
# Algebraic Equations
```

instead, and link to it would be the same.

bertiewooster:
Ah, thanks. I was confused about whether the format you meant (RST or Markdown) was for the target file (`solvers.rst`) or the origin file (`solving.md`).

Just to be clear, if the target file does not already have a shorthand text (such as `.. _solvers-algebraic-equations:`) for a heading that I need to link to, I need to add that shorthand text for the heading in the target file, correct? I had to do that for `Algebraic Equations` in `solving.rst`.

Next question: how do I point to a method that comes from API documentation? For example, for nsolve, I need to target https://docs.sympy.org/dev/modules/solvers/solvers.html?highlight=nsolve#sympy.solvers.solvers.nsolve, which is generated by the [nsolve docstring](https://github.com/sympy/sympy/blob/6ea48db2bd108a0c5b3c53aa2f1ef9e5d617bf3c/sympy/solvers/solvers.py#L2811-L3008).

bertiewooster:
Also, a math question: the ordinary differential equation currently has
$y''(x) + 9f(x)=0 $ 
with solution
$ f(x)=C_{1} \sin(3x)+ C_{2} \cos(3x)$ 
Should each of those $f(x)$ be $y(x)$?

asmeurer:
Yes, all the f(x) should be y(x).

asmeurer:
> Next question: how do I point to a method that comes from API documentation? For example, for nsolve, I need to target [docs.sympy.org/dev/modules/solvers/solvers.html?highlight=nsolve#sympy.solvers.solvers.nsolve](https://docs.sympy.org/dev/modules/solvers/solvers.html?highlight=nsolve#sympy.solvers.solvers.nsolve), which is generated by the [nsolve docstring](https://github.com/sympy/sympy/blob/6ea48db2bd108a0c5b3c53aa2f1ef9e5d617bf3c/sympy/solvers/solvers.py#L2811-L3008).

For this it's a little more complicated because the usual link syntax doesn't work. You have to do ``` {func}`~.nsolve` ``` to create a link to `nsolve()` and ``` {func}`link text <sympy.solvers.solvers.nsolve` ``` will create a link with `link text` as the text. Note that

- `func` is replaced with `class` if the object is a class, or `meth` if it's a method.
- The `~.` makes it hide the `sympy.solvers.solvers` part. But you should only include it when you don't provide the link text.
- When you provide the link text, you have to use the fully qualified name of the function (i.e., include the full module name). 

Also, we ought to update the documentation style guide for some of this Markdown stuff.

bertiewooster:
Thanks. `` {func}`~.nsolve` `` works, creating a link to `nsolve()` on the appropriate page and making the link text `nsolve()`.

`` {func}`link text <sympy.solvers.solvers.nsolve` `` does not--it gives an error when I try to build the docs locally in Sphinx. I got it to work by adding a closing angle bracket: `` {func}`link text <sympy.solvers.solvers.nsolve>` ``. That syntax does result in a gray box around the link text as for `Solve an equation numerically`:

solve-table-third-link-boxed

That is the standard formatting for link text enclosed in backticks (such as a function name), so I understand why Sphinx renders it that way. I tried removing the backticks, but then Sphinx didn't render it as a link. I tried double and triple backticks; that renders the same as single backticks. Is there perhaps a slightly different syntax that would present as a regular (non-backtick) link?

I went ahead and pushed my current version because it's pretty close, just that link formatting issue would be nice to address.

> Also, we ought to update the documentation style guide for some of this Markdown stuff.

Good idea. Is there documentation of this Markdown link-target stuff? I haven't been able to find much beyond simple linking like `[link text](target)`.

asmeurer:
The rendering of links to functions as code is a Sphinx issue. I remember seeing it before but I don't remember if there's an upstream issue about it. It also occurs with RST. Anyway, that won't be a problem any more once we replace the links with links to actual documents, so I wouldn't worry about it too much. 

> Good idea. Is there documentation of this advanced Markdown link-target stuff? I haven't been able to find much beyond simple linking like [link text](target).

The MyST docs are going to be your best bet, although they don't really cover the details with autodoc very well. More or less what I we discussed above is sufficient for the SymPy guide.

asmeurer:
Is this ready to be merged? I like the two column format. It's much more readable.

doc/src/guides/solving.md
```diff
@@ -0,0 +1,28 @@
...
+
+The Python package SymPy can symbolically solve equations, differential equations, 
+linear equations, nonlinear equations, matrix problems, inequalities, 
+Diophantine equations, and evaluate integrals. SymPy can also solve numerically.
+
+Learn how to use SymPy computer algebra system to:
+
+| Description                                                  | Example                                                                                                                     | Solution |
+|--------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------|--------------|
+| [ Solve an equation algebraically ](../modules/solvers/solvers.rst)                        | $x^2 = 4$ | $x =\{-2,2\}$                                                                                                |
```
asmeurer, line 11:
```suggestion
| [ Solve an equation algebraically ](../modules/solvers/solvers.rst)                        | $x^2 = 4$ | $x \in \{-2,2\}$                                                                                                |
```

doc/src/guides/solving.md
```diff
@@ -0,0 +1,28 @@
+# Solve Using SymPy
```
asmeurer, line 1: Would "Solve Equations" be better here? 

Still not sure about including "using SymPy" in the title. It does feel redundant once you are on the docs page. From Google, I'm not sure if it's required since "SymPy" is already in the page title of every page in the docs.

bertiewooster: I changed it to "Solve Equations"--good point that "SymPy" is in the page title.

bertiewooster:
> I like the two column format. It's much more readable.

Compared to the three-column format? I find the three-column format more readable per the [screenshots and reasons above](https://github.com/sympy/sympy/pull/23462#issuecomment-1133789357).

> Is this ready to be merged?

Other than resolving the number-of-table-columns question, yes.

asmeurer:
Sorry I mean the three-column format, as it is now.

bertiewooster:
Great, then I propose it's ready to merge.

bertiewooster:
This Solve Equations page is now live at https://docs.sympy.org/dev/guides/solving.html

PR #23464, opened 2022-05-04
Repair a couple of epigraphs in the docs.

skieffer:
#### Other comments

Before, these were using plain indentation.
Now they use the actual `.. epigraph::` RST directive.

#### Release Notes

NO ENTRY


asmeurer:
Great. I like requiring explicit `epigraph` to create a block quote.

PR #23265, opened 2022-03-21
2 fixes for `Circle` from equation

gschintgen:
#### References to other Issues or PRs

Fixes #23264

#### Brief description of what is fixed or changed
With this PR both issues described in #23264 are fixed. Example session:
```
In [1]: from sympy.geometry import Circle

In [2]: eq = (x - 1)**2 + y**2 - 9

In [3]: expand(2*eq)
Out[3]:
   2            2
2⋅x  - 4⋅x + 2⋅y  - 16

In [4]: Circle(_3)
Out[4]: Circle(Point2D(1, 0), 3)

In [5]: Circle(eq)
Out[5]: Circle(Point2D(1, 0), 3)
```
Previously, the first `Circle` call would return an incorrect radius and the second one would throw an exception.

#### Release Notes

* geometry
  * Fixed `Circle` from equation (wrong radius in some cases; allow non-expanded equation).

PR #23465, opened 2022-05-04
Add a Circle CI job to show a preview of the docs on PRs

### LINKED ISSUE #23460: Set up a docs build preview on CI
We should set up a docs build preview on CI. 

I'm not sure how to do it on GitHub pages, but you can see how NumPy does it with Circle CI https://github.com/numpy/numpy/blob/main/.circleci/config.yml. The idea is there should just be a button down with the checks where if you click on "details" it takes you to the docs built for that PR. 

If it is possible to do this on GitHub Actions that would be preferable than setting up a separate CI service, but there might be a reason it can't work.

asmeurer:
Fixes #23460.

#### Release Notes

NO ENTRY

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,7 @@ jobs:
           no_output_timeout: 25m
           command: |
             sudo apt update
-            sudo apt install imagemagick graphviz
+            sudo apt install imagemagick graphviz librsvg2-bin
             cd doc
             pip install -r requirements.txt
             make html
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,32 @@
+version: 2
+
+# Aliases to reuse
+_defaults: &defaults
+  docker:
+    # CircleCI maintains a library of pre-built images
+    # documented at https://circleci.com/docs/2.0/circleci-images/
+    - image: cimg/python:3.10.2
+  working_directory: ~/repo
+
+jobs:
+  build_page:
+    <<: *defaults
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/
+      - run:
+          name: build docs
+          no_output_timeout: 25m
+          command: |
+            cd doc
+            pip install -r requirements.txt
+            make html
+      - store_artifacts:
+          path: _build/html
+
+workflows:
+  version: 2
+  default:
+    jobs:
+      - build_page
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,6 +24,7 @@ jobs:
           name: build docs
           no_output_timeout: 25m
           command: |
+            sudo apt update
             sudo apt install imagemagick graphviz
             cd doc
             pip install -r requirements.txt
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,7 @@ jobs:
             pip install -r requirements.txt
             make html
       - store_artifacts:
-          path: _build/html
+          path: doc/_build/html
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,8 @@
+# This builds a preview of the docs which can be seen on pull requests. It
+# also uses the .github/workflows/docs-preview.yml GitHub Actions workflow.
+
+# This is separate from the GitHub Actions build that builds the docs, which
+# also builds the docs PDF.
 version: 2
 
 # Aliases to reuse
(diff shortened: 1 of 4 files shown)


asmeurer:
Circle worked but I can't get the GitHub Actions build to appear (maybe I just need to wait?). 

Also the authors build is failing. Is it because it's on a "push" build instead of a PR build?

asmeurer:
The [page for the GitHub Actions](https://github.com/larsoner/circleci-artifacts-redirector-action) says this might not work for PRs from forks. But it seems to work on every NumPy PR I checked, so I'm not sure if that's still true. If it doesn't work we can switch to Jason's idea.

oscarbenjamin:
> The [page for the GitHub Actions](https://github.com/larsoner/circleci-artifacts-redirector-action) says this might not work for PRs from forks.

This is the reason that the benchmarks comment-on-pr workflow is separate from the main runtests workflow. Basically the runtests workflow doesn't have permissions to do things like comment on a PR. A separate workflow that runs with the master branch version of the workflow files can have these permissions.

oscarbenjamin:
I see that the authors test is failing:
```
Run bin/mailmap_check.py --skip-last-commit
fatal: ambiguous argument 'HEAD^2': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
  File "/home/runner/work/sympy/sympy/bin/mailmap_check.py", line 327, in <module>
    sys.exit(main(*sys.argv[[1](https://github.com/sympy/sympy/runs/6298124542?check_suite_focus=true#step:6:1):]))
  File "/home/runner/work/sympy/sympy/bin/mailmap_check.py", line 56, in main
    git_people = get_authors_from_git(skip_last=args.skip_last_commit)
  File "/home/runner/work/sympy/sympy/bin/mailmap_check.py", line 252, in get_authors_from_git
    move(git_people, 2, 0, 'Ondřej Čertík')
  File "/home/runner/work/sympy/sympy/bin/mailmap_check.py", line 246, in move
    x = l.pop(i1)
IndexError: pop index out of range
```
That ("unknown revision") suggests that the commit use in CI (in the runtests workflow) doesn't have two ancestors for the merge commit. Not sure why that might be...

asmeurer:
> That ("unknown revision") suggests that the commit use in CI (in the runtests workflow) doesn't have two ancestors for the merge commit. Not sure why that might be...

Because it's a "push" build. I pushed the branch directly up to the repo in addition to my fork so that I could test Circle without merging. You'll probably get the same error whenever you make a release branch.

asmeurer:
> A separate workflow that runs with the master branch version of the workflow files can have these permissions.

I wonder if that's why the GitHub Actions preview "build" isn't showing up. Where are those permissions set?

asmeurer:
I removed the merge logic from the authors script, and it seems to be passing. Was it adding some special GitHub user for the merge commit before? It doesn't seem to be doing that https://github.com/sympy/sympy/runs/6314266852?check_suite_focus=true#step:6:12

asmeurer:
I can't figure out how to make the actual GitHub Actions step of this work. It isn't reporting the status with the link to the built page. Maybe it would only work once it is in master itself, although it seems to me it ought to at least work on the push build.

oscarbenjamin:
> I removed the merge logic from the authors script, and it seems to be passing.

The merge logic is needed for someone who has correctly set up their metadata in their commits (with `git config`) but whose github settings mean that github uses different metadata in any commits that it creates for them: https://github.com/sympy/sympy/pull/23265#issuecomment-1074484573.

asmeurer:
OK, I can't figure out how to get the GitHub Actions link to actually show up. I even tried making a PR within the same repo so there are no forks, and it didn't work there either https://github.com/sympy/sympy/pull/23469. @larsoner do you have any suggestions? I've more or less directly copied this from another repo where I know it works. 

The only thing I can think of is that it has to be merged into master first before it will work. I do already have this as a branch pushed to the main repo, so it seems to me that ought to be enough, but maybe it isn't. So we can try merging this. @oscarbenjamin are you OK with the mailmap_update change?

If we can't get this working, I guess we can try using @moorepants's approach instead.

asmeurer:
Oh I didn't see your comments. I'll just revert the mailmap script change. You might need to do something about it for the release branch later. I'm also surprised it doesn't cause the master builds to fail. 

It sounds pretty likely that this just isn't working because it isn't in master. Since I can't think of anything else, I'm just going to merge this (after reverting the mailmap script change). If it doesn't work then, I'll keep hammering on it, or try Jason's approach if I can't figure it out.

larsoner:
> The only thing I can think of is that it has to be merged into master first before it will work.

Yes because it operates on the `[status]` event that only gets processed on `main` AFAIK, you'll probably have to merge it and then see if it works

PR #23470, opened 2022-05-06
Use `json.load` instead of `json.loads`

Ombucha:

The code is shortened by using `json.load` instead of `json.loads`.

NO ENTRY


.mailmap
```diff
@@ -1423,6 +1423,7 @@ naelsondouglas <naelson17@gmail.com>
 noam simcha finkelstein <noam.finkelstein@protonmail.com>
 numbermaniac <5206120+numbermaniac@users.noreply.github.com>
 oittaa <8972248+oittaa@users.noreply.github.com>
+Omkaar <79257339+Pysics@users.noreply.github.com>
```
asmeurer, line 1426: This needs to be placed above with the capital Os.

PR #23472, opened 2022-05-06
Fix docs preview link

asmeurer:
#### Release Notes

NO ENTRY

### DIFF
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -9,7 +9,7 @@ jobs:
         uses: larsoner/circleci-artifacts-redirector-action@master
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-          artifact-path: 0/_build/html/index.html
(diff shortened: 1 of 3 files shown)


asmeurer:
This should work. For instance https://output.circle-artifacts.com/output/job/a9fa0426-c6b1-4be1-a41d-dc2c01e523a0/artifacts/0/doc/_build/html/index.html for #23293

asmeurer:
The preview link doesn't work here, but I think as before it uses what's in master, not the PR, so we can't tell if this will work until after we merge this.

oscarbenjamin:
You could also use the `comment-on-pr` workflow to post the URL to the PR.

asmeurer:
I made use of the Furo announcement feature to add a header the preview pages that points back to the PR and commit, and should make it so that people don't accidentally end up on a PR preview page instead of the real docs. https://output.circle-artifacts.com/output/job/151814d4-bd70-4343-8306-286c2c5e2363/artifacts/0/doc/_build/html/index.html

PR #23473, opened 2022-05-07
Run pyupgrade on printing

oscargus:

Stumbled upon pyupgrade that can automatically rewrite code to "newer" variants. Gave it a go on the printing module.

#### Other comments

It doesn't modify the alignment in case the arguments are split over multiple lines. 

Not sure it is a good idea, but wanted to test it. Mainly changed the printing from `%`-printing to `.format`-printing. This is faster, so no harm. But may possibly introduce conflicts in old PRs.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/printing/codeprinter.py
+++ b/sympy/printing/codeprinter.py
@@ -200,7 +200,7 @@ class CodePrinter(StrPrinter):
         if text != lhs_printed:
             lines.extend(openloop)
             if assign_to is not None:
-                text = self._get_statement("%s = %s" % (lhs_printed, text))
+                text = self._get_statement("{} = {}".format(lhs_printed, text))
             lines.append(text)
             lines.extend(closeloop)
 
@@ -241,7 +241,7 @@ class CodePrinter(StrPrinter):
 
                         lines.extend(openloop)
                         lines.extend(openloop_d)
-                        text = "%s = %s" % (lhs_printed, StrPrinter.doprint(
+                        text = "{} = {}".format(lhs_printed, StrPrinter.doprint(
                             self, assign_to + term))
                         lines.append(self._get_statement(text))
                         lines.extend(closeloop_d)
@@ -376,7 +376,7 @@ class CodePrinter(StrPrinter):
         else:
             lhs_code = self._print(lhs)
             rhs_code = self._print(rhs)
-            return self._get_statement("%s = %s" % (lhs_code, rhs_code))
+            return self._get_statement("{} = {}".format(lhs_code, rhs_code))
 
     def _print_AugmentedAssignment(self, expr):
         lhs_code = self._print(expr.lhs)
@@ -386,7 +386,7 @@ class CodePrinter(StrPrinter):
                  [lhs_code, expr.op, rhs_code])))
 
     def _print_FunctionCall(self, expr):
-        return '%s(%s)' % (
+        return '{}({})'.format(
             expr.name,
             ', '.join(map(lambda arg: self._print(arg),
                           expr.function_args)))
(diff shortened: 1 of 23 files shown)


sympy/printing/codeprinter.py
```diff
@@ -200,7 +200,7 @@ def _doprint_loops(self, expr, assign_to=None):
         if text != lhs_printed:
             lines.extend(openloop)
             if assign_to is not None:
-                text = self._get_statement("%s = %s" % (lhs_printed, text))
+                text = self._get_statement("{} = {}".format(lhs_printed, text))
```
oscarbenjamin, line 203: When `pyupgrade` was previously used it was decided not to change `%s` to format (there is an option to disable that aspect).

I don't necessarily mind changing it now but would it make sense to go to f-strings rather than format?
```python
                # Which of these?
                text = self._get_statement("%s = %s" % (lhs_printed, text))
                text = self._get_statement("{} = {}".format(lhs_printed, text))
                text = self._get_statement(f"{lhs_printed} = {text}")
```

sympy/printing/latex.py
```diff
@@ -664,20 +664,20 @@ def _print_Pow(self, expr):
...
+                return r"{}^{{{}}}".format(expr.base, expr.exp)
             # special case for (1/x)^(-y) and (-1/-x)^(-y), issue 20252
             if expr.base.is_Rational and \
                     expr.base.p*expr.base.q == abs(expr.base.q):
                 if expr.exp == -1:
-                    return r"\frac{1}{\frac{%s}{%s}}" % (expr.base.p, expr.base.q)
+                    return r"\frac{{1}}{{\frac{{{}}}{{{}}}}}".format(expr.base.p, expr.base.q)
                 else:
-                    return r"\frac{1}{(\frac{%s}{%s})^{%s}}" % (expr.base.p, expr.base.q, abs(expr.exp))
+                    return r"\frac{{1}}{{(\frac{{{}}}{{{}}})^{{{}}}}}".format(expr.base.p, expr.base.q, abs(expr.exp))
```
eagleoflqj, line 680: This change makes code a bit unreadable

sympy/printing/codeprinter.py
asmeurer: I agree f-strings would be even better in many cases, although `format` or `%` may be preferred if the formatting arguments are large and would make the f-string hard to read. 

I don't have a preference between `format` vs. `%`. I think pyflakes can catch errors with both, so we don't really need to worry about that.

eagleoflqj: I generally prefer f-string, unless the string contains `{}`, eg latex. In that case format function is even worse than f-string. `%s` seems the only readable choice.

asmeurer: I agree let's continue to use `%` for LaTeX and anywhere else `{}` is already part of the string (maybe in some of the set printers?).

bjodah:
Regarding f-strings and format, I would personally not change to either of those wherever doing so would introduce triple braces (`{{{`) due to escaping. I always liked that Python offers more than one way to template strings, sometimes braces have a meaning in the targeted language, and then `%`-formatting is more natural, and sometimes percent sign has a special meaning, in which case `{}`-formatting is easier.

oscargus:
I agree that f-strings are actually better and that the brace-escaping produces quite unreadable results. Since this took less than 15 minutes to accomplish and I learnt a bit in the process, I am happy to close it without further action.

I may try it again without the string-changes at a later stage (although there are probably quite limited changes remaining then).

(I believe that % is the slowest of the printing methods though, but not sure if that is actually a problem.)

asmeurer:
I just tried a very basic benchmark of the three string formatting methods and the times were about the same. And all three are so fast that I would expect the actual time to be dominated by the call to `str` on the arguments. So we should prefer whichever option gives the best readability.

oscargus:
format is actually slower than % though, so no benefit of this PR from that aspect.

https://www.scivision.dev/python-f-string-speed/

oscargus:
It actually converts format-strings to f-string, but one need to run it twice to get f-strings from %.

Still, some of the resulting code looks quite bad... And not all format are converted.

PR #23475, opened 2022-05-07
parse_latex x^\frac{1}{2}

eagleoflqj:
Previously:
```py
>>> from sympy.parsing.latex import parse_latex
>>> parse_latex(R'x^\frac{1}{2}')
x
```
#### Release Notes

* parsing
  * Latex parser treats \frac and \binom as atomic (needn't wrap with braces on exponent), and supports \dfrac, \tfrac


oscargus:
Looks good! 

Checking the diff (without knowing anything about it really): shouldn't e.g. `binom` be dealt with similarly? `x^\binom{N}{k}` etc. Probably most of those that are now `comp` should be `atom` to handle `^` properly?

eagleoflqj:
I tested with MathJax and KaTex and the result is
token|is atomic|example
-|-|-
abs_group|False|x^\|x\|
func|False|x^\sin{x}
binom|True|x^\binom{3}{2}
floor,cell|False|x^\lfloor x \rfloor

binom should be added but others shouldn't.

oscargus:
If you can add support for `\dfrac` as well, that would be great (and rather straightforward).

oscargus:
Great!

oscargus:
Seems to be a merge conflict now. I was about to merge otherwise. Feel free to self-merge when the conflict is resolved.

oscargus:
Now it merged...

Issue #20479, opened 2020-11-24
No proper way to recurse in evalf


oscarbenjamin:
The problem is that in the FiniteSet case evalf recurses to the args like this:
https://github.com/sympy/sympy/blob/b3cae120dad022dd83e338df69c45f60dd95efd9/sympy/sets/sets.py#L1886-L1887

Since that calls `_evalf` the call will go through to `_eval_evalf` which in the case of an `Add` means `AssocOp._eval_evalf`:
https://github.com/sympy/sympy/blob/b3cae120dad022dd83e338df69c45f60dd95efd9/sympy/core/operations.py#L354

However calling `x.evalf()` leads to `evalf(x, ...)` where the `evalf` function is the internal function here:
https://github.com/sympy/sympy/blob/b3cae120dad022dd83e338df69c45f60dd95efd9/sympy/core/evalf.py#L1321

The dispatches through the table to call `evalf_add`:
https://github.com/sympy/sympy/blob/b3cae120dad022dd83e338df69c45f60dd95efd9/sympy/core/evalf.py#L519

The `evalf_add` function has a special handler to ensure accuracy in the face of cancellation.

It seems that when recursing down the args in an evalf call the function to use should be `evalf(x, ...)` (as is done in the internal routines in the evalf module). The alternatives are all problematic:
1) The `x.evalf()` method takes the wrong kind of precision and is the public method.
2) The `x._evalf()` method just called `x._eval_evalf` and handles a `None` return. Perhaps this method isn't needed at all.
3) The `x._eval_evalf` method bypasses the table so it won't e.g. lead to `evalf_add`.

However the `evalf(x, ...)` function also isn't suitable for recursion from `x._eval_evalf` because the `_eval_evalf` method doesn't receive the arguments that should be passed to `evalf(x, ...)`. The `_eval_evalf` method is clearly the way for non-core classes to implement `evalf` but there does not seem to be a proper way for such a method to recurse to its args.

This all seems like a bit of a mess to me and indeed usage around the codebase of `.evalf`, `._evalf` and `._eval_evalf` is inconsistent.

_Originally posted by @oscarbenjamin in https://github.com/sympy/sympy/issues/20379#issuecomment-730563821_

oscarbenjamin:
We need a new method to replace `_eval_evalf` etc that can actually propagate options for evalf.

sidhu1012:
[...]

PR #20321, opened 2020-10-22
Fixed `FiniteSet.evalf(...)` with a subs argument

Maelstrom6:
#### References to other Issues or PRs

Fixes #20291 

#### Brief description of what is fixed or changed

`FiniteSet(a).evalf(subs={a: 1})` would return `FiniteSet(a)` instead of `FiniteSet(1)`. This PR fixes this and similar issues.

#### Release Notes

* sets
    * `FiniteSet.evalf` with a subs argument now does the substitution. Previously, the argument was ignored.

Issue #17280, opened 2019-07-27
Cleaning up Basic and subclasses for evaluate/doit


oscarbenjamin:
TL;DR I think that SymPy's model for Basic is messy in places and I think that there could be significant benefits in cleaning it up. Below are some ideas.

SymPy's Basic evaluation model broadly works as follows:

Basic subclasses define `__new__` which
1. Usually sympifies the arguments
2. Possibly raises if the arguments are of the wrong type/value
3. Possibly evaluates by simplifying the arguments (e.g. `re(x+I) -> re(x)`)
4. Possibly evaluates to return a different object (`Mul(2, 3) -> Integer(6)`)
5. Otherwise sends the arguments to `Basic.__new__` to create the object, set
args etc.

Evaluation is typically controlled in a few ways:
1. Some `__new__` functions don't evaluate when given `evaluate=False` giving en "unevaluated object".
2. Others don't evaluate by default and always return unevaluated.
3. Unevaluated objects can be evaluated by calling doit.
4. doit has kwargs like `deep` which can control whether doit is called recursively on args.

That all seems like the nice model but it isn't implemented consistently across the SymPy code base and presumably by downstream authors/users writing their own Basic subclasses.

## 1. Inconsistent support for evaluate kwarg

There are 351 `__new__` methods in SymPy:
```shell
$ git grep 'def __new__' | wc -l
     351
```
I (crudely) count 3 `__new__` methods that have an evaluate parameter and 124 that have `**kwargs`. That means that the majority of these classes canot accept an evaluate kwarg. Of those `__new__` methods that take `**kwargs` many do not respect any `evaluate` kwarg and will evaluate regardless.

Classes like Derivative/Integral/Subs that don't evaluate by default are inconsistent in accepting an evaluate kwarg. For example Derivative respects evaluate=True but Integral raises. There is also no programmatic way to tell whether a particular class accepts an evaluate kwarg or whether it would or would not evaluate by default.

I propose that all Basic types should accept an evaluate=True kwarg. As a kwarg it should default to None but True, False and None should be accepted.  In the case of None a class attribute (`cls.autodoit`?) should be checked. This allows Integral etc to disable evaluation by default at the class level. Subclasses should then not define `__new__` but something else like `_eval_new` that would be called by `Basic.__new__` (if evaluate=True - perhaps this should be `_eval_doit` - see below).

Simpification should be handled in `Basic.__new__` perhaps controlled by a
[...]

PR #23488, opened 2022-05-11
Add a guide on writing custom functions

asmeurer:
This is still a work in progress, but early reviews are always welcome. 

Reminder to reviewers that you can use the PR docs preview by clicking "Click here to see a preview of the documentation." in the checks list below.

#### Release Notes

- other
  - Add a guide to writing custom functions.


doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,414 @@
...
+>>> f(1)
+2
+>>> f(x)
+x + 1
+```
+
+If you find yourself defining an `eval` method on a `Function` subclass where
+you always return a value and never return `None`, you should consider just
+using a normal Python function instead, as there is no benefit to using a
+symbolic `Function` subclass in that case.
```
oscarbenjamin, line 111: There is potential benefit if you can use `evaluate=False` and also have printing support. Another point to consider is that maybe it should be possible to call the function with symbols as arguments so the eval method should maybe handle that.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,414 @@
...
+
+Note that in many cases, functions like these can be represented directly
+using SymPy classes. For example, the above function can be represented
+symbolically using `Piecewise`. The function can then be evaluated using
+{meth}`subs() <sympy.core.basic.Basic.subs>`.
```
oscarbenjamin, line 116: Also perhaps worth mentioning `Lambda` and maybe even `lambdify` here:
```python
In [19]: f = Lambda(x, Piecewise((0, Eq(x, 0)), (x+1, True)))

In [20]: f
Out[20]: 
    ⎧  0    for x = 0
x ↦ ⎨                
    ⎩x + 1  otherwise

In [21]: f(0)
Out[21]: 0
```

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,414 @@
...
+  ...    def eval(cls, x):
+  ...        return 1 - cos(x)
+  ```
+
+  However, this would make it so that `versin(x)` would *always* return `1 -
+  cos(x)`, regardless of what `x` is. If all you want is a quick shorthand to
+  `1 - cos(x)`, that is fine. But would be much simpler and more explicit to
+  just use a Python function, as [described
+  above](custom-functions-fully-evaluated). If we defined `versin` like this,
+  it would never actually return as `versin(x)`, and none of the other
```
oscarbenjamin, line 257: Unless `evaluate=False` is given. Perhaps worth mentioning here and also to clarify that `__new__` should not be used for `Function` subclasses because `Function.__new__` already takes care of things like this.

doc/src/guides/custom-functions.md
asmeurer: Mentioning lambdify is a good idea. I thought about Lambda, but I'm a little unsure because I'm not sure just how well supported it is throughout SymPy. Very few things make use of it.

doc/src/guides/custom-functions.md
asmeurer: But do we really want to recommend that? evaluate=False is very fragile. It's a much better design to just avoid automatic evaluation and move it to some other method such as `doit`. Even for something like printing, it can be more complicated than it might seem, because that doesn't really tell the whole story. What if you want to manipulate the expression somewhat to get it in some form to print. What if you later decide to do some stuff that isn't printing. The more code you have build up around a function, the harder it is to remove automatic evaluation from it. That's my experience with SymPy functions and I believe it is also true of user defined functions.

And even if you are just doing printing and nothing else wouldn't it still be simpler to just have a function that doesn't evaluate (in fact, you could just use a `Function('f')` like I described earlier in the guide) than passing evaluate=False everywhere?

> Another point to consider is that maybe it should be possible to call the function with symbols as arguments so the eval method should maybe handle that.

I'm not following this. Both a `def` function and an `eval` method of a `Function` subclass should handle symbols as arguments.

doc/src/guides/custom-functions.md
asmeurer: Good point, I'll mention to avoid `__init__` and `__new__`. 

Actually it would be nice if the `eval` logic were on `Basic` rather than `Function`. It might have to be generalized a bit, but it would be great it 99% of SymPy objects did not need to define `__new__`.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,414 @@
...
+functions <special-functions>` like {class}`~.gamma()` or {class}`~.Si()`, and
+{ref}`combinatorial functions <combinatorial-functions>` and {ref}`number
+theory functions <ntheory-module>` like {class}`~.factorial()` or
+{func}`~.primepi()`. Thus custom user functions are able to do the same things
+as functions defined by SymPy.
+
+This guide describes how to define functions that map a subset of
+$\mathbb{C}^n$ to $\mathbb{C}$. Functions that accept or return other kinds of
+objects should subclass another class, such as {class}`~.Boolean`,
+{class}`~.MatrixExpr`, {class}`~.Expr`, or {class}`~.Basic`. Much of what is
```
asmeurer, line 16: Hopefully this will change in the future if we can integrate the kind system into Function (https://github.com/sympy/sympy/issues/4787, which might change other aspects of this guide as well).

doc/src/guides/custom-functions.md
asmeurer: The only benefit I can see is maybe if we implement https://github.com/sympy/sympy/issues/4787, you could have a symbolic function object `f` which always evaluates. But even then, I don't know if having functions as objects that can be inside of other expressions is something that too many people really need. https://github.com/sympy/sympy/issues/4787 is important, but that's the most important part of it. Not many people request that feature.

The reason I am stressing the "don't just return an expression in `eval`" and also the "use a `def` function if you just want a simple shortcut" is because it's a common thing I see people do on StackOverflow and the mailing list. I think part of it is that people don't really sit down and think about what they actually want, but also part of it is that they don't completely understand how `eval` works and maybe don't even realize that just returning an expression in `eval` has this implication. 

Finally, I would hope that everything in this guide should apply equally well to functions in library code as to user defined functions, and indeed I expect it to be useful for contributors. So I'm trying to apply the highest standard to the recommendations here. Users are of course free to do whatever they want, but we should not recommend something that we wouldn't do ourselves. The goal is to replace this wiki page, which isn't written very well https://github.com/sympy/sympy/wiki/Development-Tips-:-About-implementing-special-functions.

oscarbenjamin: I agree that automatic evaluation should be discouraged for symbolic functions. My suggestion though is that if someone is in the position of writing a Function subclass with an eval method that always returns maybe they should consider changing it so that the eval method does not always return so that then their function can also work with symbolic arguments.

doc/src/guides/custom-functions.md
oscarbenjamin: The potential advantage of `Lambda` is that you get a symbolic object that can be used like a function with Python's call syntax (assuming that's what is wanted):
```python
In [1]: eq_y = Lambda(x, Piecewise((0, Eq(x, y)), (1, True)))

In [2]: eq_y
Out[2]: 
    ⎧0  for x = y
x ↦ ⎨            
    ⎩1  otherwise

In [3]: eq_y(y+1)
Out[3]: 1

In [4]: eq_y(y)
Out[4]: 0
```
Unlike other options the `Lambda` function can be used with `subs` to substitute something that is not the "argument" of the function:
```python
In [5]: eq_4 = eq_y.subs(y, 4)

In [6]: eq_4
Out[6]: 
    ⎧0  for x = 4
x ↦ ⎨            
    ⎩1  otherwise
```

asmeurer: The thing I don't understand about `Lambda` is that it's an `Expr`. But does it actually make sense to include a `Lambda` inside of other expressions? You can't call it unless the top-level expression is a `Lambda`. Does something like `Lambda(...) + 1` even make sense?

doc/src/guides/custom-functions.md
asmeurer: I think which one you really want requires introspection as the whether you're looking for a symbolic function or just a shorthand. I can try to make it clearer here that there are two options.

doc/src/guides/custom-functions.md
oscarbenjamin: I would prefer it if `Lambda` wasn't a subclass of Expr. There is a backwards way of calling it:
```python
In [1]: f = Lambda(x, x**2)

In [2]: f
Out[2]: 
     2
x ↦ x 

In [3]: exp(f)
Out[3]: 
      2
 x ↦ x 
ℯ      

In [4]: exp(f).rcall(3)
Out[4]: 
 9
ℯ
```
I'm not sure I've ever seen that used in the wild though.

asmeurer:
GitHub has added $\mathrm{\LaTeX}$ math support to GitHub comments: $$\int_0^1\sin(x)\, dx.$$ Just use dollar signs, like `$$\int_0^1\sin(x)\, dx.$$`. Looks like they have a bug with `\,` and `\:`. Hopefully that will be fixed soon.

oscarbenjamin:
> GitHub has added 
> LATEX math

Nice!

(Although apparently it doesn't work with quoting...)

agryman:
Very cool.

agryman (commented):
The new LaTeX math support if great!

bertiewooster:
Suggest splitting this guide into several pages, to decrease page size and length.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,879 @@
...
+This guide will describe how to create custom function classes in SymPy.
+Custom functions use the same mechanisms as the {ref}`functions <functions>`
+that are included with SymPy such as the common {ref}`elementary functions
+<elementary-functions>` like {class}`~.exp()` or {class}`~.sin()`,
+{ref}`special functions <special-functions>` like {class}`~.gamma()` or
+{class}`~.Si()`, and {ref}`combinatorial functions <combinatorial-functions>`
+and {ref}`number theory functions <ntheory-module>` like
+{class}`~.factorial()` or {func}`~.primepi()`.
+
+This guide describes how to define functions that map a subset of
```
bertiewooster, line 16: Suggest describing the goal of this function in words, and/or by giving an example, in addition to in mathematical symbols: someone who wants to write a mathematically simple custom function could get confused by the math already.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,879 @@
...
+
+Ultimately, the correct tool for the job depends on what you are doing and
+what exact behavior you want.
+
+(custom-functions-function-subclass)=
+## Creating a custom function
+
+The basic way to create a custom function is to subclass {class}`~.Function`.
+The name of the subclass will be the name of the function. Different methods
+should then e defined on this subclass, depending on what functionality you
```
bertiewooster, line 165: typo: _e_ -> _be_

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,879 @@
...
+```
+
+Finally, note that automatic evaluation on floating-point inputs happens
+automatically once `evalf` is defined. See the section on [defining numerical
+evaluation with `evalf`](custom-functions-evalf) below.
+
+(custom-functions-eval-best-practices)=
+#### Best Practices for `eval`
+
+- **Don't just return an expression.**
```
bertiewooster, line 293: Per [SymPy tone preferences](https://docs.sympy.org/dev/guides/contributing/documentation-style-guide.html#tone-preferences), suggest phrasing this positively. For example, "Return _description_ instead of an expression".

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,879 @@
...
+  [differentiation](custom-functions-differentiation), and so on.
+
+  If you find yourself doing this, you should think about what you actually
+  want to achieve. If you just want a shorthand function for an expression, it
+  will be simpler to just [define a Python
+  function](custom-functions-fully-evaluated). If you really do want a
+  symbolic function, think about when you want it to evaluate to something
+  else and when you want it to stay unevaluated.
+
+- **Avoid too much automatic evaluation.**
```
bertiewooster, line 331: Per [SymPy tone preferences](https://docs.sympy.org/dev/guides/contributing/documentation-style-guide.html#tone-preferences), suggest phrasing this positively. For example, "Use an appropriate amount of  automatic evaluation".

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,879 @@
...
+#### `as_real_imag`
+
+#### Miscellaneous `_eval_*` methods
+
+
+(custom-functions-complete-examples)=
+## Complete Examples
+
+(custom-functions-example-fma)=
+### FMA
```
bertiewooster, line 838: Suggest spelling out FMA as fused multiply-add here because the H3 appears in the page table of contents (right navbar), so a visitor scanning the table of contents might not have read the section above where FMA is defined.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,879 @@
...
+...         elif argindex == 3:
+...             return 1
+```
+
+## Additional Tips
+
+- SymPy includes dozens of functions. These can serve as useful examples for
+  how to write a custom function, especially if the function is similar to one
+  that is already implemented. Remember that everything in this guide applies
+  equally well to functions that are included with SymPy and user defined
```
bertiewooster, line 868: user defined functions -> user-defined functions (add a hyphen)

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,879 @@
...
+  functions.
+
+- If you have many custom functions that share common logic, you can use a
+  common base class to contain this shared logic. For an example of this, see
+  the source code for the trigonometric functions in SymPy, which use
+  `TrigonometricFunction`, `InverseTrigonometricFunction`, and
```
bertiewooster, line 874: Suggest linking to the source code for those functions if possible.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,879 @@
...
+  `ReciprocalTrigonometricFunction` base classes with some shared logic.
+
+- As with any code, it is a good idea to write extensive tests for your
+  function. The SymPy test suite is a good resource for examples of how to
```
bertiewooster, line 878: Suggest linking to the SymPy test suite if possible.

asmeurer:
Thanks for the review. How would you suggest splitting this? I agree the guide is a bit lengthy, though hopefully that's less of a concern in general now that we have the Furo theme with the right sidebar.

bertiewooster:
Maybe split into pages like this:

- Main page contains the intro and then links to the sub-pages
- Sub-page for Easy Cases
- Sub-page for Creating a Custom Function. Or maybe multiple sub-pages because that section alone is a dozen screens tall (on a good-sized monitor), for example
    - eval and evalf (does it make sense to put those in adjacent sections, and move Assumptions after evalf?)
    - Assumptions and Rewriting and Simplification
    - Differentiation and Series Expansions
    - Printing and Other Methods
- Sub-page for Complete Examples
- Sub-page for Additional Tips, or could be combined with Complete Examples if each section remains short

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,879 @@
...
+accurately represent symbolic values. For example, the above Python `def`
+definition of `f`, `f(x)` implicitly assumes that `x` is nonzero. The
+`Piecewise` version handles this case correctly and won't evaluate to the $x
+\neq 0$ case unless `x` is known to not be zero.
+
+Ultimately, the correct tool for the job depends on what you are doing and
+what exact behavior you want.
+
+(custom-functions-function-subclass)=
+## Creating a custom function
```
bertiewooster, line 161: Use Title Case: Creating a **C**ustom **F**unction

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,879 @@
...
+### Numerical Evaluation with `evalf`
+
+<!-- TODO: this goes over the basics, but it might be useful to have a
+     separate guide dedicated to numerical evaluation. -->
+
+Here we show how to define how to define how a function should  numerically
+evaluate to a floating point {class}`~.Float` value, for instance, via `evalf`.
+
+If your function has the same name as a function in
+[mpmath](https://mpmath.org/doc/), as is the case for most functions in SymPy,
```
bertiewooster, line 591: This link is invalid; https://mpmath.org/doc/current is valid

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1613 @@
...
+```py
+>>> divides(m, n).is_integer
+True
+>>> divides(m, n).is_nonnegative
+True
+```
+
+In general, however, the assumptions of a function depend on the assumptions
+of its inputs. In this case, you should define an <code class="docutils literal notranslate"><span
+class="pre">\_eval\_*assumption*</span></code> method.
```
asmeurer, line 574: I was going to add an admonition here about not defining `is_assumption` as a @property method and using `_eval_is_assumption` instead, but I'm not actually sure if anything bad would happen if you do that, other than just being bad style. @oscarbenjamin do you know if anything particularly bad can happen if you do this? I tested it and it seems to actually work.

oscarbenjamin: It won't work with the assumptions cache and won't compute implications.
```
In [110]: exp(2)._assumptions
Out[110]: {}

In [111]: exp(2).is_zero
Out[111]: False

In [112]: exp(2)._assumptions
Out[112]: {'zero': False}

In [113]: exp(2).is_negative
Out[113]: False

In [114]: exp(2)._assumptions
Out[114]: 
{'zero': False,
 'finite': True,
 'infinite': False,
 'extended_negative': False,
 'negative': False}

In [115]: class A(Expr):
     ...:     @property
     ...:     def is_positive(self):
     ...:         return True
     ...: 

In [116]: a = A()

In [117]: a.is_positive
Out[117]: True

In [118]: a._assumptions
Out[118]: {}
```

asmeurer: Ah you are right. For some reason I didn't notice this when I was playing with it before:

```
>>> class Test(Function):
...     @property
...     def is_zero(self):
...         return self.args[0].is_zero
>>> t = Test(0)
>>> t.is_zero
True
>>> t._assumptions
{}
>>> t.is_negative # This is not computed
>>> class Test2(Function):
...     def _eval_is_zero(self):
...         return self.args[0].is_zero
>>> t2 = Test2(0)
>>> t2.is_zero
True
>>> t2._assumptions
{'zero': True, 'negative': False, 'nonzero': False, 'extended_negative': False, 'integer': True, 'infinite': False, 'complex': True, 'odd': False, 'extended_positive': False, 'extended_nonpositive': True, 'algebraic': True, 'irrational': False, 'extended_real': True, 'nonnegative': True, 'imaginary': False, 'prime': False, 'composite': False, 'finite': True, 'transcendental': False, 'positive': False, 'extended_nonzero': False, 'even': True, 'real': True, 'noninteger': False, 'nonpositive': True, 'hermitian': True, 'extended_nonnegative': True, 'rational': True, 'commutative': True}
>>> t2.is_negative
False
```

asmeurer: I wonder if we ought to try to detect and warn about this in the metaclass (as long as it wouldn't hurt performance).

asmeurer: I have a proof of concept of that at https://github.com/sympy/sympy/pull/23640. However, there are a few issues with it that we'd need to figure out first. So for now, I've just included a warning in this doc about this. If that PR gets merged we can remove the warning, or at least tone it down.

asmeurer:
This is now ready for final review. A few notes:

- I've opened several issues relating to various things that I noticed while writing this guide (see the cross references). In some cases, I've omitted documentation for something because I either felt that the existing APIs were incomplete or poorly designed, or, in the case of series expansion, because it's not something I'm yet comfortable enough with documenting (and it's big enought to warrent its own guide anyway). However, with the exception of series expansion, I feel this guide touches all the important things. 

- I have tried to emphasize `doit` as a better place to put advanced evaluation rather than `eval`. I see two primary use-cases for doit:
  - A place to put more advanced simplifications, particularly symbolic ones that make use of assumptions.
  - A way to define an "unevaluated placeholder" function that a user may want to sometimes evaluate to what it represents (see the FMA example).

- I'm still undecided on whether this ought to be split into multiple pages or not. Personally I think it's fine, especially given the right sidebar navigation, but I'd like others' feedback on this.

- I'm a little unsure how to handle the second person ("you") vs. the third person ("we") in the guide. Right now there's a little mix of both. Should we try to always use one or the other?

doc/src/guides/custom-functions.md
oscarbenjamin: Are you objecting to the use of the word "avoid"?

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1693 @@
...
+  It is recommended to minimize what is evaluated automatically by `eval()`.
+  It is typically better to put more advanced simplifications in [other
+  methods](custom-functions-rewriting-and-simplification), like
+  [`doit()`](custom-functions-doit). Remember that whatever you define for
+  automatic evaluation will *always* evaluate.[^evaluate-footnote] As in the
+  previous point, if you evaluate every value, there is little point to even
+  having a symbolic function in the first place. For example, we might be
+  tempted to evaluate some trig identities on `versin` in `eval()`, but then
+  these identities would always evaluate, and it wouldn't be possible to
+  represent one half of the identity.
```
oscarbenjamin, line 371: It might be good to have an example here showing what the implications would be and why we might not want to use automatic evaluation.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1693 @@
...
+  assumptions system is not able to deduce this:
+
+  ```py
+  >>> print(((m**2 + m)/2).is_integer)
+  None
+  ```
+
+  SymPy's assumptions system is always improving, but there will always be
+  cases like this that it cannot deduce, due to the fundamental computational
+  complexity of the problem.
```
oscarbenjamin, line 480: The problem is not just computationally complex but is in fact mathematically impossible in general.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1693 @@
...
+
+  ```
+
+(custom-functions-assumptions)=
+### Assumptions
+
+The next thing we might want to define are the assumptions on our function.
+The assumptions system allows defining what mathematical properties your
+function has given its inputs, for example, "$f(x)$ is *positive* when $x$ is
+*real*."
```
oscarbenjamin, line 538: It's a bit strange to switch from "we" to "your" here.

doc/src/guides/custom-functions.md
asmeurer: I added an example. See what you think.

doc/src/guides/custom-functions.md
asmeurer: I guess it's both. Sometimes a problem is decidable but intractable. And I'm hoping the assumptions system can evolve into a state where it only returns simple deductions by default and only tries harder when explicitly told to.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1773 @@
...
+method `_eval_evalf(self, prec)`, where `prec` binary precision of the input.
+The method should return the expression evaluated to the given precision, or
+`None` if this is not possible.
+
+```{note}
+
+The `prec` argument to `_eval_evalf()` is the *binary* precision, that is, the
+number of bits in the floating-point representation. This differs from the
+first argument to the `evalf()` method, which is the *decimal* precision, or
+`dps`. For example, the default binary precision of `Float` is 53,
```
oscarbenjamin, line 769: Perhaps worth mentioning the conversion functions `prec_to_dps` et al

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1773 @@
...
+method recursively calls evalf on another expression, it should call
+`expr._eval_evalf(prec)` rather than `expr.evalf(prec)`, as the latter will
+incorrectly use `prec` as the decimal precision.
+
+```
+
+We can define numerical evaluation for [our example $\operatorname{versin}(x)$
+function](custom-functions-versine-definition) by recursively evaluating
+$2\sin^2\left(\frac{x}{2}\right)$, which is a more numerically stable way of writing $1 -
+\cos(x)$.
```
oscarbenjamin, line 780: Maybe not worth mentioning here but perhaps check out the discussion in #20479 and #20493

oscarbenjamin: See also #20321

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1773 @@
...
+
+-->
+
+(custom-functions-doit)=
+#### `doit()`
+
+The {meth}`doit() <sympy.core.basic.Basic.doit>` method is used to evaluate
+"unevaluated" functions. To define `doit()` implement `doit(self, deep=True,
+**hints)`. If `deep=True`, `doit()` should recursively call `doit()` on the
+arguments. `**hints` will be any other keyword arguments passed to the user,
```
oscarbenjamin, line 893: It would be much better to have `_eval_doit` so that subclasses don't need to implement recursion (#17280)

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1773 @@
...
+...     # inputs are not nonintegral.
+...     @classmethod
+...     def eval(cls, m, n):
+...         # Evaluate for explicit integer m and n.
+...         if isinstance(m, Integer) and isinstance(n, Integer):
+...             return int(n % m == 0)
+...
+...         # For symbolic arguments, require m and n to be integer.
+...         if m.is_integer is False or n.is_integer is False:
+...             raise TypeError("m and n should be integers")
```
oscarbenjamin, line 929: Ideally we wouldn't use `eval` for type checking so that we can still check types when `evaluate=False`.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1773 @@
...
+x
+>>> FMA(x, y, z).diff(z)
+1
+>>> FMA(x**2, x + 1, y).diff(x)
+x**2 + 2*x*(x + 1)
+```
+
+To leave a derivative unevaluated, raise
+`sympy.core.function.ArgumentIndexError(self, argindex)`. This is the default
+behavior if `fdiff()` is not defined. Here is an example function $f(x, y)$ that
```
oscarbenjamin, line 1145: Does returning `None` not work?

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1773 @@
...
+>>> versin(x).expand(complex=True)
+I*sin(re(x))*sinh(im(x)) - cos(re(x))*cosh(im(x)) + 1
+```
+
+#### Miscellaneous `_eval_*` methods
+
+There are many other functions in SymPy whose behavior can be defined on
+custom functions via a custom `_eval_*` method, analogous to the ones
+described above. See the documentation of the specific function for details on
+how define each method.
```
oscarbenjamin, line 1320: how to define

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1773 @@
...
+...     # Define basic assumptions.
+...     def _eval_is_nonnegative(self):
+...         x = self.args[0]
+...         # versin(x) is nonnegative if x is real
+...         return x.is_real
+...
+...     def _eval_is_positive(self):
+...         x = self.args[0]
+...         # versin(x) is positive if x is real and not an even multiple of pi
+...         return fuzzy_and([x.is_real, fuzzy_not((x/pi).is_even)])
```
oscarbenjamin, line 1385: Creating a new expression like `x/pi` in an assumptions query is problematic. This kind of thing is bad for performance and also often leads to infinite recursion.

doc/src/guides/custom-functions.md
asmeurer: I'm not a fan of that function, because it loses precision (so to speak). Multiple binary prec's correspond to the same dps. Ideally evalf would just take `prec` as an optional argument as an alternative to dps (I can't remember if this has been discussed before).

doc/src/guides/custom-functions.md
asmeurer: Yes definitely. I noticed this problem for several methods. `expand` does it correctly, but most methods don't.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1773 @@
...
+...         if deep:
+...            m, n = m.doit(deep=deep, **hints), n.doit(deep=deep, **hints)
+...
+...         # divides(m, n) is 1 iff n/m is an integer. Note that m and n are
+...         # already assumed to be integers because of the logic in eval().
+...         isint = (n/m).is_integer
+...         if isint is True:
+...             return 1
+...         elif isint is False:
+...             return 0
```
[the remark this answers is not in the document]
asmeurer: Oh I guess so. That's another problem with defining raw methods rather than `_eval_*` methods.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1773 @@
...
+...         else:
+...             return divides(m, n)
```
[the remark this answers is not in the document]
asmeurer: I don't think there are any cases where this can make a difference. I'm also unclear even if there were if that would be what you'd want.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1773 @@
...
+
+```
+
+For our [$\operatorname{versin}$ example
+function](custom-functions-versine-definition), the derivative is $\sin(x)$.
+
+```py
+>>> class versin(Function):
+...     def fdiff(self, argindex=1):
+...         return sin(self.args[0])
```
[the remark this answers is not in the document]
asmeurer: It's mentioned in the text above but I'll also note it here with a comment. I decided to ignore invalid `argindex`, since that should never happen from `diff` and users shouldn't really ever call `fdiff` (see https://github.com/sympy/sympy/issues/23490).

doc/src/guides/custom-functions.md
asmeurer: https://github.com/sympy/sympy/issues/23490

doc/src/guides/custom-functions.md
asmeurer: Hmm, OK. I thought this was a pretty common pattern, but I can see how it might cause problems. 

So is a better pattern to parse the expression (like with the _peeloff_pi helper in the SymPy trig functions)? That seems more complex, especially for end users. What's the simplest way to write this?

oscarbenjamin: This kind of thing is more common in the codebase than it should be. There might not really be a better way but there could be some discussion about this in this guide.

The problem is that creating an expression leads to evaluation which leads to assumptions queries. So if assumptions queries create expressions then we have recursion. Whenever you have recursion you need some proof that your recursion terminates. If you know that you are always creating a smaller expression than the one that you started with then that is some proof. Likewise if your assumptions queries only ever query assumptions directly on their args then you know that you are always recursing to a query on a smaller expression tree.

In this case it is probable fine to do this but for a different reason: you are recursing from a query on a `versin` expression to a query on an expression from the set of expressions that do not involve `versin`. If you know that the set of expressions that do not involve `versin` will never create a `versin` then that is safe.

Getting this right is non-trivial and there should be clear warnings about it. The general rule is:

1. Ideally an assumptions query on x is resolved just by doing some direct assumptions queries on `x.args`.
2. Assumption queries should aboid creating any new expressions.
3. If it is necessary to create any new expression then you need to think very carefully about whether this recursion is guaranteed to terminate.

In this particular case you could use `as_independent` to do a structural check that you do actually have a `Mul` involving `pi`:
```
coeff, pi_ = x.as_independent(pi, as_Add=False)
if pi_ is pi:
    return fuzzy_and([x.is_real, fuzzy_not(coeff.is_even)])
elif x.is_real is False:
    return False
```
Here we always know that `as_independent` only creates smaller expressions. At worst it returns the original expression as the coeff and 1 for `pi_`.

asmeurer: I see. I do remember seeing issues in the past with that sort of thing. It's not something that we can really show with this example, because `versin` isn't used anywhere else in the codebase, so it's impossible for it to sneak in. But it can happen if someone is editing assumptions for a more core function in SymPy. 

Actually I don't even think it's possible to "think very carefully" about whether the recursion will terminate. It really depends on what assumptions queries are being done elsewhere, and what expressions might create other expressions, which you cannot possibly know unless you've read the whole codebase. It's tough though because there's a lot of advanced logic in other class' assumptions handlers and you really want to be able to reuse that whenever possible.

This really is a major design flaw in the old assumptions. It makes the handlers implement both the specification of what the predicates are and the actual method by which they are resolved. That leads to problems like this where a handler would need to have global information to do that properly, but it can't. It also makes certain optimizations in the inference impossible, even basic ones like not evaluating the same inference twice. Contrast that with the new assumptions where the predicates are specified symbolically and a global solver determines how to resolve everything and which sub-predicates need to be evaluated. Global solvers (like SAT solvers) can handle cyclical inference chains just fine without infinite recursion:

```py
>>> list(satisfiable(And(x >> y, y >> z, z >> x), all_models=True))
[{y: False, z: False, x: False}, {y: True, z: True, x: True}]
```

oscarbenjamin: > Actually I don't even think it's possible to "think very carefully" about whether the recursion will terminate

In general no it isn't possible if arbitrary code can be in any of the handlers. However it is possible if we limit what can go into the handlers. Two rules are enough but they need to be applied universally:

1. Assumptions queries should not create new expressions.
2. A query on expr should be resolved only by querying assumptions on the args of expr (and not other queries on expr itself).

With these two rules we always know that any assumptions query only ever recurses to queries on smaller expressions and therefore the recursion must at some point terminate.

> This really is a major design flaw in the old assumptions

The flaw is expecting the old assumptions to be something that they are not. They are relatively well defined for their purpose which is to be a limited inference system for use in evaluation. Remember that much of what they do is not really about "assumptions" in the sense of assuming things about symbols but actually just about computing basic properties often of numeric expressions.

asmeurer:
Review comments have been addressed. I used your example for the `_eval_is_positive` method. Ad wrote an admonition against creating expressions in assumptions handlers. 

It's a little unfortunate that we have to use `if`/`elif` instead of just a single `fuzzy_*` method with your version. Maybe it would help to add `fuzzy_if` (or `fuzzy_implies`). It wouldn't evaluate lazily (which is a problem with all the `fuzzy_` methods), but it would be more robust to correct `None` handling.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1827 @@
...
+handler**. Always pull apart the args of the function using structural methods
+like `as_independent`.
+
+```py
+>>> from sympy.core.logic import fuzzy_and, fuzzy_not
+>>> class versin(Function):
+...     def _eval_is_nonnegative(self):
+...         # versin(x) is nonnegative iff x is real
+...         x = self.args[0]
+...         return x.is_real
```
oscarbenjamin, line 690: We need to be more careful here:
```python
In [12]: (1 - cos(pi+I*pi))
Out[12]: 1 + cosh(π)
```
If `x` is not real then that does not imply that `versin(x)` is not nonnegative.

asmeurer: Ah good catch. Of course, the real range is just [0, 2], so there are complex values that map it to (2, oo).

Really I would like to just write `(1 - cos(x)).is_real` here. If you look in the guide, a lot of the methods do just delegate to `1 - cos(x)`. Partly that is just me being lazy/keeping the guide simple. But I also do think there are legitimate reasons why you'd want to do that in general. If something can be reduced to something already implemented somewhere else, you'd like to be able to use it. It's generally going to be more robust to reuse something already in SymPy as a user since it's going to be way better tested, and even for library code, it reduces duplication which also minimizes bugs. So it's really unfortunate to have a "no recursive expressions" rule, which I agree is better for performance and avoiding infinite recursion, but makes it harder to avoid rewriting hard mathematical logic.

But also, one of the use-cases for custom functions that this guide talks about is functions that are shorthands for another function (the see FMA example in the guide). In that case, you have a function that represents another expression and evaluates with doit. 

Maybe we just need to have a specific way to create something like this, since the basic pattern is

```py
def Unevaluated(Function):
    @classmethod
    def eval(cls, *args):
        # Evaluate to <expression> on explicit numerical inputs

    def doit(self, deep=True):
        return <expression>

    def _eval_whatever(self):
        return <expression>._eval_whatever()
```

where the last part is basically the same for every possible method that you would want to define. This whole thing can be wrapped up into a single `Function` subclass which you could subclass to specify the expression you want to represent, and only override methods if you don't want the default behavior (like defining a custom code printer).

asmeurer: Made https://github.com/sympy/sympy/issues/23701 about the wrapper function superclass idea

asmeurer:
Any more comments on this?

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1850 @@
...
+>>> from sympy.core.logic import fuzzy_and, fuzzy_not
+>>> class versin(Function):
+...     def _eval_is_nonnegative(self):
+...         # versin(x) is nonnegative if x is real
+...         x = self.args[0]
+...         if x.is_real is True:
+...             return True
+...
+...     def _eval_is_positive(self):
+...         # versin(x) is positive iff x is real and not an even multiple of pi
```
oscarbenjamin, line 712: The example above shows that this is not an if and only if.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1850 @@
...
+We can define numerical evaluation for [our example $\operatorname{versin}(x)$
+function](custom-functions-versine-definition) by recursively evaluating
+$2\sin^2\left(\frac{x}{2}\right)$, which is a more numerically stable way of writing $1 -
+\cos(x)$.
+
+```py
+>>> from sympy import sin
+>>> class versin(Function):
+...     def _eval_evalf(self, prec):
+...         return (2*sin(self.args[0]/2)**2)._eval_evalf(prec)
```
oscarbenjamin, line 834: Maybe this is okay but in general I would prefer not to invoke symbolic evaluation as part of numeric evaluation.

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1850 @@
...
+Contrast this with the
+[versine](custom-functions-versine-full-example) example, which treats
+`versin` as a first-class function in its own regard. Even though `versin(x)`
+can be expressed in terms of other functions (`1 - cos(x)`) it does not
+evaluate on general symbolic inputs in `versin.eval()`, and `versin.doit()` is
+not defined at all.
+
+`FMA` also represents an example of a continuous function defined on multiple
+vriables, which demonstrates how `argindex` works in the
+[`fdiff`](custom-functions-differentiation) example.
```
oscarbenjamin, line 1700: variables

doc/src/guides/custom-functions.md
```diff
@@ -0,0 +1,1850 @@
...
+(custom-functions-versin-eval-example)=
+```py
+>>> from sympy import pi, Integer
+>>> class versin(Function):
+...    @classmethod
+...    def eval(cls, x):
+...        # If x is an integer multiple of pi, x/pi will cancel and be an Integer
+...        n = x/pi
+...        if isinstance(n, Integer):
+...            return 1 - (-1)**n
```
oscarbenjamin, line 257: Maybe it's worth discussing the use of sympify somewhere. I think you've changed it so that both the inputs and outputs of this function are automatically sympified. That's very different from other Basic subclasses so it should probably be clarified somewhere.

oscarbenjamin:
I had a couple of comments but I think this looks good in any case.

oscarbenjamin:
Actually I meant to leave that for you to merge but hit the wrong button...

As I said I think it looks good!

doc/src/guides/custom-functions.md
asmeurer: I'm not following. What is your concern here?

doc/src/guides/custom-functions.md
asmeurer: With things being sympified automatically it's not something you really need to worry about. I'd love for all `Basic` subclasses and methods to work like that, and also for all classes to more generally use an `eval` pattern rather than overriding `__new__`. There are too many bugs caused by people forgetting to sympify inputs/outputs, plus people tend to use sympify instead of _sympify when they do it themselves.

asmeurer:
Followup at https://github.com/sympy/sympy/pull/23732

PR #23489, opened 2022-05-12
manualintegrate 1/sqrt(a+b*x+c*x**2)

eagleoflqj:
Previously these all can't be calculated.
```py
from sympy import *
x=Symbol('x')
integrate(1/sqrt(3*x**2+4*x+5))
integrate(1/sqrt(-3*x**2+4*x+5))
integrate(1/sqrt(3*x**2+4*x-5))
```
Moreover, `manualintegrate(1/sqrt(3*x**2+4), x)` gives unnecessary step. Left is current master, right is this branch:

Screenshot from 2022-05-11 23-28-59

#### Release Notes

* integrals
  * Support integrate(1/sqrt(a+bx+cx**2))

### DIFF
--- a/sympy/integrals/manualintegrate.py
+++ b/sympy/integrals/manualintegrate.py
@@ -457,13 +457,11 @@ def inverse_trig_rule(integral):
     base, exp = integrand.as_base_exp()
     a = Wild('a', exclude=[symbol])
     b = Wild('b', exclude=[symbol])
-    match = base.match(a + b*symbol**2)
+    c = Wild('c', exclude=[symbol, 0])
+    match = base.match(a + b*symbol + c*symbol**2)
 
     if not match:
         return
-
-    def negative(x):
-        return x.is_negative or x.could_extract_minus_sign()
 
     def ArcsinhRule(integrand, symbol):
         return InverseHyperbolicRule(asinh, integrand, symbol)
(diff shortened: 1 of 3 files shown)


sympy/integrals/manualintegrate.py
```diff
@@ -457,58 +457,47 @@ def inverse_trig_rule(integral):
...
     possibilities = []
 
     if simplify(2*exp + 1) == 0:
-        possibilities.append((ArcsinRule, exp, a, 1, -b, -1, And(a > 0, b < 0)))
-        possibilities.append((ArcsinhRule, exp, a, 1, b, 1, And(a > 0, b > 0)))
-        possibilities.append((ArccoshRule, exp, -a, -1, b, 1, And(a < 0, b > 0)))
+        h, k = -b/(2*c), a - b**2/(4*c)  # rewrite base to k + c*(symbol-h)**2
+        possibilities.append((ArcsinRule, k, 1, -c, -1, h, And(k > 0, c < 0)))  # 1-x**2
+        possibilities.append((ArcsinhRule, k, 1, c, 1, h, And(k > 0, c > 0)))  # 1+x**2
+        possibilities.append((ArccoshRule, -k, -1, c, 1, h, And(k < 0, c > 0)))  # -1+x**2
```
oscarbenjamin, line 497: What happens if k or c is non-real?

eagleoflqj:
```py
from sympy import *
from sympy.integrals.manualintegrate import *
x=Symbol('x')
manualintegrate(1/sqrt(x**2+I), x)
```
An exception will be raised, both current master and this branch.
I never learned how to deal with that case.

oscarbenjamin: It seems that it can be done with atanh:
```python
In [70]: a = atanh(x/sqrt(x**2 + y))

In [71]: ratsimp(diff(a, x))
Out[71]: 
     1     
───────────
   ________
  ╱  2     
╲╱  x  + y 
```
Is that valid for any nonzero complex number `y`?

eagleoflqj: I tried some specific complex `y` and it seems valid. But I don't know how it works for complex quadratic coefficient.
I just noticed [Wolfram|Alpha](https://www.wolframalpha.com/input?i=integrate+1%2Fsqrt%28a%2Bbx%2Bcx%5E2%29) provides a general solution: `ln(2*sqrt(c)*sqrt(a+b*x+c*x**2)+b+2*c*x)/sqrt(c)`. By manually diff, it should be correct as all simplifications are universal (e.g. no `sqrt(a)*sqrt(b)==sqrt(a*b)` used).
The problem is, should it belong to manual integration? How one is supposed to get this result manually?

eagleoflqj: Just pushed a commit to solve that.
After completing square, only if we know both k and c are real, we apply inverse trig/hyperbolic rules. Otherwise, fall back to the general solution.

oscarbenjamin:
With this we have
```
In [11]: integral_steps(1/sqrt(a + b*x + c*x**2), x)
Out[11]: ReciprocalSqrtQuadraticRule(a=a, b=b, c=c, context=1/sqrt(a + b*x + c*x**2), symbol=x)
```
Maybe it would make more sense to have a "complete the square" rule. Then to solve these cases the steps would be:
1. Complete the square
2. Make the substitution to bring it to a canonical form
3. Apply a specific rule like inverse hyperbolic

eagleoflqj:
> ```
> In [11]: integral_steps(1/sqrt(a + b*x + c*x**2), x)
> Out[11]: ReciprocalSqrtQuadraticRule(a=a, b=b, c=c, context=1/sqrt(a + b*x + c*x**2), symbol=x)
> ```

That's expected. Completing square are only applied when both k and c are real. The general integration formula provided by Wolfram Alpha doesn't seem achieved by completing square. It seems achieved by algebraic transformation.

> have a "complete the square" rule

OK. This is a special RewrittenRule that only differs in wording. If SymPy Beta uses this branch and handles the new CompleteSquareRule, the result is
Screenshot from 2022-05-14 11-33-57

oscarbenjamin:
Are there other cases where it would make sense to apply trig/hyperbolic rules? With this we have:
```python
In [1]: from sympy import *
   ...: from sympy.integrals.manualintegrate import *
   ...: x=Symbol('x')
   ...: manualintegrate(1/sqrt(x**2+I), x)
Out[1]: 
   ⎛           ________⎞
   ⎜          ╱  2     ⎟
log⎝2⋅x + 2⋅╲╱  x  + ⅈ ⎠
```
Potentially it's nicer to use `atanh` in that case:
```
In [7]: print(ratsimp(atanh(x/sqrt(x**2+I)).diff(x)))
1/sqrt(x**2 + I)
```
Maybe that's not nicer though given that we end up using complex numbers either way...

eagleoflqj:
I prefer the `log` version, as `atanh` is just a `log`. In complex world it's not simpler than `log` anymore (it may be subjective because my Complex Function course book talks about `log`, but no hyperbolic functions).

oscarbenjamin:
I think this looks good.

PR #23498, opened 2022-05-15
functions - fixed inconsistency with binomial factorial function

### LINKED ISSUE #23497: binomial(-1, -1) returns 0, should return 1
I am in the process of converting some nuclear physics code from Maple to SymPy and I ran into a bug which I traced to the behaviour of the binomial coefficient function. Specifically:

Maple: binomial(-1, -1) = 1
SymPy: binomial(-1, -1) = 0

The SymPy docs for binomial(n, k) state: For the sake of convenience for negative integer k this function will return zero no matter what valued is the other argument.
See https://docs.sympy.org/latest/modules/functions/combinatorial.html#binomial

Therefore, the code is working as described. However, it is not mathematically correct. The Maple result is correct. Here's the reasoning.

For all n, k, we must have:
binomial(n, k) = binomial(n, n - k)

Therefore:
binomial(-1, -1) = binomial(-1, -1 - (-1)) = binomial(-1, 0)

The generalized definition for binomial(n, k) given in in the SymPy docs uses the falling factorial function, ff(n, k):
binomial(n, k) = ff(n, k)/k!

By definition of ff(n, k) we have ff(n, 0) = 1 for all n.
Therefore binomial(-1, 0) = ff(-1, 0)/0! = 1/1 = 1.

I think it is appropriate to correct this behavior for integer arguments. I understand that this changes the behaviour and could potentially break existing code, but that is probably doubtful. However, if you really want to preserve this behaviour, perhaps you could create a new function that handles the case of negative k "correctly"?

fyi, the Maple docs explain how binomial(n, k) should be extended to complex arguments. I'd be happy to share that if there is interest.

praneethratna:
Fixes #23497 and now `binomial(-1, -1)` returns `1` instead of `0` .

#### Release Notes

* functions
  * fixed inconsistency with `binomial(-1, -1)`

### DIFF
--- a/sympy/functions/combinatorial/factorials.py
+++ b/sympy/functions/combinatorial/factorials.py
@@ -955,6 +955,8 @@ class binomial(CombinatorialFunction):
         if (k - 1).is_zero or ((n_nonneg or n_isint is False)
                 and (d - 1).is_zero):
             return n
+        if k == -1 and n == -1:
+            return S.One
         if k.is_integer:
             if k.is_negative or (n_nonneg and n_isint and d.is_negative):
                 return S.Zero
--- a/sympy/functions/combinatorial/factorials.py
+++ b/sympy/functions/combinatorial/factorials.py
@@ -955,9 +955,9 @@ class binomial(CombinatorialFunction):
         if (k - 1).is_zero or ((n_nonneg or n_isint is False)
                 and (d - 1).is_zero):
             return n
-        if k == -1 and n == -1:
-            return S.One
         if k.is_integer:
+            if k.is_negative and (n.is_negative and n_isint):
+                return S.One
             if k.is_negative or (n_nonneg and n_isint and d.is_negative):
                 return S.Zero
             elif n.is_number:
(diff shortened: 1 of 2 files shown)


sympy/functions/combinatorial/factorials.py
```diff
@@ -955,6 +955,8 @@ def eval(cls, n, k):
         if (k - 1).is_zero or ((n_nonneg or n_isint is False)
                 and (d - 1).is_zero):
             return n
+        if k == -1 and n == -1:
+            return S.One
```
oscarbenjamin, line 959: If we were going to make this change then it would be to evaluate for general negative integer values rather than just special casing these particular two values.

praneethratna: I think there was a discussion regarding this, and `binomial(-1, -1)` and in general `binomial(kn, kn)` where `kn` is a negative value returning `0` change was brought by PR #14575. I've made a lot of tests where both `k` and `n` are negative evaluate to 1.

praneethratna: I think `summation(binomial(2*k, k)/5**k, (k, -oo, oo))` is not evaluated since after this PR binomial gives 1 for negative numbers but `summation(binomial(2*k, k)/5**k, (k, 0, oo))` gives the correct answer equal to `sqrt(5)`. Should this test be changed too?

oscarbenjamin: I think it's better to discuss on the issue. There needs to be a decision on changing the behaviour of binomial before making a PR.

praneethratna: Since the OP issue is closed, should i close this PR as well?

PR #22213, opened 2021-10-03
Add `nsimplify` call in case limit argument has a `Float`

0sidharth:
#### References to other Issues or PRs

Fixes #12312
Fixes #19453 

#### Brief description of what is fixed or changed
Convert floats like `0.5` to exact SymPy numbers like `S.Half`, to
prevent rounding errors which can lead to unexpected execution
of conditional blocks that work on comparisons

#### Other comments
See #19453 

#### Release Notes

* series
  * Limit now converts `Float`'s in the expression to `Rational`s or exact SymPy numbers when possible

PR #23499, opened 2022-05-15
test: fix use of Float/Rational in tests

oscarbenjamin:
This commit changes tests that assert comparisons between Floats and
Rationals as well as Floats with different precision. Where a tested
function returns a Float (or an expression involving Floats) then the
test should compare the returned value with a Float. Likewise if exact
Rationals are returned then the test should assert that.

There were also many tests that use Floats with specific precision but
compare against Floats with different precision. This commit ensures
that a correct precision Float is always used in these comparisons.

This extracts some of the harmless changes from #20033.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/codegen/ast.py
+++ b/sympy/codegen/ast.py
@@ -1535,7 +1535,7 @@ class Variable(Node):
         >>> decl1.variable.value == NoneToken()  # OK
         True
         >>> decl2 = x.as_Declaration(value=42.0)
-        >>> decl2.variable.value == 42
+        >>> decl2.variable.value == 42.0
         True
 
         """
--- a/sympy/core/tests/test_numbers.py
+++ b/sympy/core/tests/test_numbers.py
@@ -1597,25 +1597,25 @@ def test_Rational_gcd_lcm_cofactors():
     assert Integer(4).cofactors(Integer(2)) == \
         (Integer(2), Integer(2), Integer(1))
 
-    assert Integer(4).gcd(Float(2.0)) == S.One
+    assert Integer(4).gcd(Float(2.0)) == Float(1.0)
     assert Integer(4).lcm(Float(2.0)) == Float(8.0)
-    assert Integer(4).cofactors(Float(2.0)) == (S.One, Integer(4), Float(2.0))
+    assert Integer(4).cofactors(Float(2.0)) == (Float(1.0), Float(4.0), Float(2.0))
 
-    assert S.Half.gcd(Float(2.0)) == S.One
+    assert S.Half.gcd(Float(2.0)) == Float(1.0)
     assert S.Half.lcm(Float(2.0)) == Float(1.0)
     assert S.Half.cofactors(Float(2.0)) == \
-        (S.One, S.Half, Float(2.0))
+        (Float(1.0), Float(0.5), Float(2.0))
 
 
 def test_Float_gcd_lcm_cofactors():
-    assert Float(2.0).gcd(Integer(4)) == S.One
+    assert Float(2.0).gcd(Integer(4)) == Float(1.0)
     assert Float(2.0).lcm(Integer(4)) == Float(8.0)
-    assert Float(2.0).cofactors(Integer(4)) == (S.One, Float(2.0), Integer(4))
+    assert Float(2.0).cofactors(Integer(4)) == (Float(1.0), Float(2.0), Float(4.0))
 
-    assert Float(2.0).gcd(S.Half) == S.One
+    assert Float(2.0).gcd(S.Half) == Float(1.0)
     assert Float(2.0).lcm(S.Half) == Float(1.0)
     assert Float(2.0).cofactors(S.Half) == \
-        (S.One, Float(2.0), S.Half)
+        (Float(1.0), Float(2.0), Float(0.5))
 
 
 def test_issue_4611():
@@ -2013,7 +2013,7 @@ def test_issue_10020():
 def test_invert_numbers():
     assert S(2).invert(5) == 3
     assert S(2).invert(Rational(5, 2)) == S.Half
-    assert S(2).invert(5.) == 0.5
+    assert S(2).invert(5.) == S.Half
     assert S(2).invert(S(5)) == 3
     assert S(2.).invert(5) == 0.5
     assert S(sqrt(2)).invert(5) == 1/sqrt(2)
@@ -2222,7 +2222,7 @@ def test_floordiv():
 
 def test_negation():
     assert -S.Zero is S.Zero
-    assert -Float(0) is not S.Zero and -Float(0) == 0
+    assert -Float(0) is not S.Zero and -Float(0) == 0.0
 
 
 def test_exponentiation_of_0():
--- a/sympy/core/tests/test_sympify.py
+++ b/sympy/core/tests/test_sympify.py
@@ -51,10 +51,9 @@ def test_sympify1():
     assert sympify("   x") == Symbol("x")
     assert sympify("   x   ") == Symbol("x")
     # issue 4877
-    n1 = S.Half
-    assert sympify('--.5') == n1
-    assert sympify('-1/2') == -n1
-    assert sympify('-+--.5') == -n1
+    assert sympify('--.5') == 0.5
+    assert sympify('-1/2') == -S.Half
+    assert sympify('-+--.5') == -0.5
     assert sympify('-.[3]') == Rational(-1, 3)
     assert sympify('.[3]') == Rational(1, 3)
     assert sympify('+.[3]') == Rational(1, 3)
@@ -132,6 +131,7 @@ def test_sympify_mpmath():
     assert sympify(
         mpmath.pi).epsilon_eq(Float("3.14159"), Float("1e-6")) == False
 
+    mpmath.mp.dps = 15
     assert sympify(mpmath.mpc(1.0 + 2.0j)) == Float(1.0) + Float(2.0)*I
 
     assert sympify(mpq(1, 2)) == S.Half
@@ -407,7 +407,7 @@ def test_int_float():
     f1_1c = F1_1c()
     assert sympify(i5) == 5
     assert isinstance(sympify(i5), Integer)
-    assert sympify(i5b) == 5
+    assert sympify(i5b) == 5.0
     assert isinstance(sympify(i5b), Float)
     assert sympify(i5c) == 5
     assert isinstance(sympify(i5c), Integer)
@@ -417,7 +417,7 @@ def test_int_float():
 
     assert _sympify(i5) == 5
     assert isinstance(_sympify(i5), Integer)
-    assert _sympify(i5b) == 5
+    assert _sympify(i5b) == 5.0
     assert isinstance(_sympify(i5b), Float)
     assert _sympify(i5c) == 5
     assert isinstance(_sympify(i5c), Integer)
@@ -625,20 +625,31 @@ def test_sympify_numpy():
     assert equal(sympify(np.float32(1.123456)), Float(1.123456, precision=24))
     assert equal(sympify(np.float64(1.1234567891234)),
                 Float(1.1234567891234, precision=53))
+
+    # The exact precision of np.longdouble, npfloat128 and other extended
+    # precision dtypes is platform dependent.
+    ldprec = np.finfo(np.longdouble(1)).nmant + 1
     assert equal(sympify(np.longdouble(1.123456789)),
-                 Float(1.123456789, precision=80))
+                 Float(1.123456789, precision=ldprec))
+
     assert equal(sympify(np.complex64(1 + 2j)), S(1.0 + 2.0*I))
     assert equal(sympify(np.complex128(1 + 2j)), S(1.0 + 2.0*I))
-    assert equal(sympify(np.longcomplex(1 + 2j)), S(1.0 + 2.0*I))
+
+    lcprec = np.finfo(np.longcomplex(1)).nmant + 1
+    assert equal(sympify(np.longcomplex(1 + 2j)),
+                Float(1.0, precision=lcprec) + Float(2.0, precision=lcprec)*I)
 
     #float96 does not exist on all platforms
     if hasattr(np, 'float96'):
+        f96prec = np.finfo(np.float96(1)).nmant + 1
         assert equal(sympify(np.float96(1.123456789)),
-                    Float(1.123456789, precision=80))
+                    Float(1.123456789, precision=f96prec))
+
     #float128 does not exist on all platforms
     if hasattr(np, 'float128'):
+        f128prec = np.finfo(np.float128(1)).nmant + 1
         assert equal(sympify(np.float128(1.123456789123)),
-                    Float(1.123456789123, precision=80))
+                    Float(1.123456789123, precision=f128prec))
 
 
 @XFAIL
@@ -820,9 +831,9 @@ def test_issue_14706():
 
     assert sympify(numpy.array([1])) == ImmutableDenseNDimArray([1], 1)
     assert sympify(numpy.array([[[1]]])) == ImmutableDenseNDimArray([1], (1, 1, 1))
-    assert sympify(z1) == ImmutableDenseNDimArray([0], (1, 1))
-    assert sympify(z2) == ImmutableDenseNDimArray([0, 0, 0, 0], (2, 2))
-    assert sympify(z3) == ImmutableDenseNDimArray([0], ())
+    assert sympify(z1) == ImmutableDenseNDimArray([0.0], (1, 1))
+    assert sympify(z2) == ImmutableDenseNDimArray([0.0, 0.0, 0.0, 0.0], (2, 2))
+    assert sympify(z3) == ImmutableDenseNDimArray([0.0], ())
     assert sympify(z3, strict=True) == 0.0
 
     raises(SympifyError, lambda: sympify(numpy.array([1]), strict=True))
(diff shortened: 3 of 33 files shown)


oscarbenjamin:
This is ready for review.

sympy/core/tests/test_numbers.py
```diff
@@ -2002,7 +2002,7 @@ def test_issue_10020():
 def test_invert_numbers():
     assert S(2).invert(5) == 3
     assert S(2).invert(Rational(5, 2)) == S.Half
-    assert S(2).invert(5.) == 0.5
```
ThePauliPrinciple, line 2005: Is `5.1` not a float here? Maybe because of sympify/python? It appears to me as if this test was meant to return a float, but perhaps it stopped doing so during some commit?

sympy/core/tests/test_sympify.py
```diff
@@ -626,10 +626,11 @@ def equal(x, y):
     assert equal(sympify(np.float64(1.1234567891234)),
                 Float(1.1234567891234, precision=53))
     assert equal(sympify(np.longdouble(1.123456789)),
-                 Float(1.123456789, precision=80))
+                 Float(1.123456789, precision=64))
     assert equal(sympify(np.complex64(1 + 2j)), S(1.0 + 2.0*I))
     assert equal(sympify(np.complex128(1 + 2j)), S(1.0 + 2.0*I))
-    assert equal(sympify(np.longcomplex(1 + 2j)), S(1.0 + 2.0*I))
+    assert equal(sympify(np.longcomplex(1 + 2j)),
```
ThePauliPrinciple, line 632: Why does this become `precision=64`? Shouldn't it be 128 since a longcomplex is 256 bit https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.clongdouble

sympy/physics/units/tests/test_quantities.py
```diff
@@ -253,7 +253,7 @@ def test_mul_div():
 
 
 def test_units():
-    assert convert_to((5*m/s * day) / km, 1) == 432
+    assert convert_to((5*m/s * day) / km, 1) == 432.0
```
ThePauliPrinciple, line 256: Is this not exactly 432? Why should this be a float?

sympy/series/tests/test_limits.py
```diff
@@ -372,10 +372,10 @@ def test_issue_4547():
 
 def test_issue_5164():
     assert limit(x**0.5, x, oo) == oo**0.5 is oo
-    assert limit(x**0.5, x, 16) == S(16)**0.5
+    assert limit(x**0.5, x, 16) == 4 # Should this be a float?
```
ThePauliPrinciple, line 375: I believe so, why would it lose its "floatness"?

ThePauliPrinciple:
Structural comparisons (`==`) between floats of different precision should clearly result in unequal, but it is (still) not so clear to me what should happen for `Eq`. 

In general I feel that float inputs into functions should be remembered. If people feel that in their case floats can safely be converted to integers they can round it explicitly, it should not be assumed that it can be.

sympy/core/tests/test_numbers.py
oscarbenjamin: I'm not sure if this is something that was changed at some point or if the original author simply didn't check.

sympy/core/tests/test_sympify.py
oscarbenjamin: The precision refers to the number of bits in the mantissa so e.g. np.float64 corresponds to a precision of 53 bits. I think np.float128 should have 113 bits of precision. That's not what sympify does though:
```python
In [29]: sympify(np.longcomplex(1 + 2j)).args[0]._prec
Out[29]: 64
```

sympy/physics/units/tests/test_quantities.py
oscarbenjamin: I'm not sure where it comes from.

sympy/series/tests/test_limits.py
oscarbenjamin: Probably because the float is converted into (or interpreted as) a Rational somewhere.

oscarbenjamin:
To be clear the purpose of this PR is not to change any behaviour of SymPy. The idea is just to make sure that the tests are explicit about what they are testing. I also want to reduce the diff for #20033 so that it's clearer what the actual changes are and so that I can also use it to extract other more useful changes.

sympy/core/tests/test_sympify.py
ThePauliPrinciple: So then one could state that the result of `sympify(np.longcomplex(1 + 2j))` is incorrect? Since it creates a `Float` with the wrong precision?

asmeurer: The code in `sympify` is using `np.finfo(dtype).nmant`, which does return `63` for `np.longcomplex` https://github.com/sympy/sympy/blob/42248dd7be300bc3893642fd92793c0ff4c19a89/sympy/core/sympify.py#L91

ThePauliPrinciple:
> To be clear the purpose of this PR is not to change any behaviour of SymPy. The idea is just to make sure that the tests are explicit about what they are testing. I also want to reduce the diff for #20033 so that it's clearer what the actual changes are and so that I can also use it to extract other more useful changes.

This is how I interpreted this PR. My concern is mostly that some tests have now been changed to, what appears to me, as an incorrect result, when float comparisons should be considered more strictly. e.g. `assert limit(x**0.5, x, 16) == 4` to me should be `4.0 == 4`, which should be false. 

So while this paves the way to a better comparison of floats (and this is absolutely the right direction), it might also be hiding incorrect results of other tests because they are based on the current outcome of the tests and not on what the outcome should be. These tests are currently (as in, before this PR) only passing because float comparisons are implemented in an inconsistent way in the first place.

sympy/core/tests/test_sympify.py
oscarbenjamin: It comes from here:
https://github.com/sympy/sympy/blob/42248dd7be300bc3893642fd92793c0ff4c19a89/sympy/core/sympify.py#L90-L96
We can try that manually:
```python
In [7]: np.finfo(np.float128(1)).nmant + 1
Out[7]: 64

In [8]: np.finfo(np.float64(1)).nmant + 1
Out[8]: 53
```
The numpy docs say 
```
nmant : int
The number of bits in the mantissa.
```
https://numpy.org/doc/stable/reference/generated/numpy.finfo.html

However if you look here then you can see that the actual size of these datatypes is platform dependent:
https://numpy.org/doc/stable/user/basics.types.html#extended-precision

So actually np.float128 is most likely not the same as IEEE 128 bit floating point. I guess that on my computer (and also in CI) what we actually get is an 80-bit floating point number (as was used in x87 FPUs in the 32 bit days):
https://en.wikipedia.org/wiki/Extended_precision#Extended_precision_implementations

ThePauliPrinciple: I wonder if `finfo` then is not accurate. Why would `sympy.sympify(np.complex64(1 + 2j)).args[0]._prec == sympy.sympify(np.complex128(1 + 2j)).args[0]._prec == 53`?

ThePauliPrinciple: I see, so then the test should probably be rewritten in some platform independent way?

oscarbenjamin:
> My concern is mostly that some tests have now been changed to, what appears to me, as an incorrect result, when float comparisons should be considered more strictly.

I have added comments in places where I thought that was the case. It is always necessary to apply some interpretation when reading the results of the test suite.

> it might also be hiding incorrect results of other tests because they are based on the current outcome of the tests and not on what the outcome should be.

I don't know what you mean by this. This PR just makes it explicit in the test code itself what is being returned. If we want to enforce that in future then we can use something like #23499 except only have it run in CI rather than actually making the change for end users. In the mean time #23499 can be used to see whether there are any tests that mix floats and rationals in comparisons.

ThePauliPrinciple:
> I don't know what you mean by this

Again, the example `assert limit(x**0.5, x, 16) == 4`. The argument here is apparently to make the change from ` ... == 4.0` to `... == 4` because that is currently the result of `limit(x**0.5, x, 16)` is `4`. And you want to be able to later on change the structural comparison of `Float` and `Integer` to fail.

While I understand this reasoning, the result is now that we are testing for `limit(x**0.4, x, 16) == 4`, which I think is not the right test. I would argue that this should return `False`, eventhough it currently does not.

sympy/core/tests/test_sympify.py
asmeurer: See https://stackoverflow.com/questions/9062562/what-is-the-internal-precision-of-numpy-float128/17023995#17023995 and https://github.com/numpy/numpy/issues/10288.

oscarbenjamin:
> While I understand this reasoning, the result is now that we are testing for `limit(x**0.4, x, 16) == 4`, which I think is not the right test.

That's why I added a comment on that line "should this be a float?". The point is to make it clear to someone looking at it that the test itself can be questioned and should not be considered definitive. The comment will also show up in the output of the test runner:
```
  File "/home/oscar/current/sympy/sympy.git/sympy/series/tests/test_limits.py", line 375, in test_issue_5164
    assert limit(x**0.5, x, 16) == 4 # Should this be a float?
AssertionError
```

ThePauliPrinciple:
The original test was correct, so I don't think it should be replaced with the incorrect result and a comment.

sympy/core/tests/test_sympify.py
oscarbenjamin: I've changed the test to be based on np.finfo

oscarbenjamin:
> The original test was correct, so I don't think it should be replaced with the incorrect result and a comment.

I will revert the change but I think that if the goal is not to hide potentially incorrect behaviour then leaving the test as it is in master is worse.

ThePauliPrinciple:
> if the goal is not to hide potentially incorrect behaviour then leaving the test as it is in master is worse.

This is what I don't understand. How I see it, when the change is made for `Float` to compare inequal to `Integer` this test will start to fail. It can then become an "xfail" test to indicated it is known to fail but should eventually pass. Or alternatively, in the mean time `limits` is changed to give `4.0` as a result.

oscarbenjamin:
> when the change is made for `Float` to compare inequal to `Integer`

I don't think it's a foregone conclusion that that change would ever be made (there isn't really any agreement if you look at the PR). The fact that I am working on this now is not because I think that the change is necessarily going to happen but rather because I want to extract other useful things from the PR.

> It can then become an "xfail" test to indicated it is known to fail but should eventually pass.

In practice xfail tests are pretty useless in sympy. Most contributors don't notice them because they don't cause CI to fail. Most people who write xfail tests don't know how to write a test that actually works so that it will pass one day.

To me the most basic point in the test code is that I want to be able to judge from reading the code in the test what the current behaviour of sympy is. If there is some behaviour that I care about then I want the test suite to show me when that behaviour changes (for better or for worse).

Currently in master those tests are misleading: they appear to assert a result that is not what actually happens (and then still pass). That means that the test code itself gives a misleading impression of current sympy behaviour. Whether that behaviour is correct or should be changed is a separate point and this is where I think that there should be more comments in the test code.

ThePauliPrinciple:
In previous versions this limit actually did return a float:
```python
>>> import sympy
>>> x = sympy.abc.x
>>> print(sympy.limit(x**0.5, x, 16))
>>> print(sympy.__version__)
4.00000000000000
1.7.1
```
So we're really looking at a regression here that wasn't picked up because of how `Float` currently compares.

oscarbenjamin:
I've bisected the change to limit to cec5c2b7a79a37d1fcf181eeca1a4772d7d8ca3c from #22213. The change seems deliberate to me

CC @0sidharth

ThePauliPrinciple:
Alright, if these changes are intentional than that is what it is. It would be nice if we could have some consistent definition of what a `Float` is in SymPy and when it is allowed to be automatically converted to some other type without the user knowing about it.

0sidharth:
These changes are indeed intended. I don't think a better alternative for dealing with the original issue of #22213 exists, so maybe just mentioning to the user when this conversion is done on purpose would be the way to go.

oscarbenjamin:
Maybe the conversion to rational could be temporary and then the result could be converted back to float later. You can something similar here:
https://github.com/sympy/sympy/blob/832c24fec1046eaa544a4cab4c69e3af3e651759/sympy/solvers/solveset.py#L3520-L3525
Then later:
https://github.com/sympy/sympy/blob/832c24fec1046eaa544a4cab4c69e3af3e651759/sympy/solvers/solveset.py#L3550-L3553

0sidharth:
Thanks for the suggestion, this should be easily implementable. Although I don't like having to add such a conditional check, the consistency is probably valuable, so I'll try to see if I can get this in within the next day.

sympy/physics/units/tests/test_quantities.py
smichr:
```diff
diff --git a/sympy/physics/units/util.py b/sympy/physics/units/util.py
index 08f611858e..ee6e4c6ee5 100644
--- a/sympy/physics/units/util.py
+++ b/sympy/physics/units/util.py
@@ -99,16 +99,20 @@ def convert_to(expr, target_units, unit_system="SI"):
         target_units = [target_units]
 
     if isinstance(expr, Add):
-        return Add.fromiter(convert_to(i, target_units, unit_system) for i in expr.args)
+        return Add.fromiter(convert_to(i, target_units, unit_system)
+            for i in expr.args)
 
     expr = sympify(expr)
+    target_units = sympify(target_units)
 
     if not isinstance(expr, Quantity) and expr.has(Quantity):
-        expr = expr.replace(lambda x: isinstance(x, Quantity), lambda x: x.convert_to(target_units, unit_system))
+        expr = expr.replace(lambda x: isinstance(x, Quantity),
+            lambda x: x.convert_to(target_units, unit_system))
 
     def get_total_scale_factor(expr):
         if isinstance(expr, Mul):
-            return reduce(lambda x, y: x * y, [get_total_scale_factor(i) for i in expr.args])
+            return reduce(lambda x, y: x * y,
+                [get_total_scale_factor(i) for i in expr.args])
         elif isinstance(expr, Pow):
             return get_total_scale_factor(expr.base) ** expr.exp
         elif isinstance(expr, Quantity):
@@ -120,7 +124,9 @@ def get_total_scale_factor(expr):
         return expr
 
     expr_scale_factor = get_total_scale_factor(expr)
-    return expr_scale_factor * Mul.fromiter((1/get_total_scale_factor(u) * u) ** p for u, p in zip(target_units, depmat))
+    return expr_scale_factor * Mul.fromiter(
+        (1/get_total_scale_factor(u)*u)**p for u, p in
+        zip(target_units, depmat))
 
 
 def quantity_simplify(expr, across_dimensions: bool=False, unit_system=None):
```

The `target_units` needs to be sympified.

oscarbenjamin:
I've reverted the changes to units tests after #23537

smichr:
Should this wait for #23539?

sympy/core/tests/test_numbers.py
oscarbenjamin: The code that handles this is here:
https://github.com/sympy/sympy/blob/857a97e66c6bb162eb922179aea20226c7a0ed3b/sympy/core/numbers.py#L546-L550
I have no idea what it is trying to do but it ignores the type of `m` if `m > 1`. Hence:
```python
In [1]: S(2).invert(5.)
Out[1]: 1/2

In [2]: S(2.).invert(5.)
Out[2]: 0.500000000000000

In [3]: S(2.).invert(5)
Out[3]: 0.500000000000000
```
If both are integers:
```python
In [4]: S(2).invert(5)
Out[4]: 3
```

sympy/series/tests/test_limits.py
oscarbenjamin: I reverted this change

oscarbenjamin:
> Should this wait for #23539?

No, I've reverted the change to that test.

There shouldn't be anything controversial here. Reviews welcome.

sympy/codegen/ast.py
```diff
@@ -1535,7 +1535,7 @@ def as_Declaration(self, **kwargs):
         >>> decl1.variable.value == NoneToken()  # OK
         True
         >>> decl2 = x.as_Declaration(value=42.0)
-        >>> decl2.variable.value == 42
+        >>> decl2.variable.value == 42.0
```
asmeurer, line 1538: I guess you don't have to fix this here, but I don't like doctests that are written like this. I would rather see

```py
>>> decl2.variable.value
42.0
```

We should only write doctests with `==` this when it's impossible to doctest the output for some reason (and even then, maybe just skipping the test can be better). 

Even when the output is `None` (which doctest hides), I'd rather see

```py
>>> print(x.is_integer)
None
```

than

```py
>>> x.is_integer == None
True
```

If these doctests were written this way in the first place you wouldn't need to fix them here.

asmeurer:
Just scrolling through this looks good, and does make #20033 convincing, since quite a few of these tests were basically wrong before. 

I'm wondering in #20033, how do Floats with the non-default precision work when comparing to `float`? Will `Float('1.0', 5) == 1.0` give True or False?

oscarbenjamin:
> Will `Float('1.0', 5) == 1.0` give True or False?

In #20033 a normal Python float is considered to be whatever sympify would make of it:
```python
In [3]: srepr(S(1.0))
Out[3]: "Float('1.0', precision=53)"

In [4]: Float('1.0', 5) == 1.0
Out[4]: False

In [5]: S(1.0) == 1.0
Out[5]: True
```
When I can get this merged I'll update #20033 to see what can be done with it.

oscarbenjamin:
Okay I'll merge this now.

PR #23500, opened 2022-05-15
add type hints to printing/latex.py

eagleoflqj:
#### Other comments
There are many `_print_*` methods just return `str`. I only pick some of them whose return type can't be deduced automatically by pyright.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/printing/latex.py
+++ b/sympy/printing/latex.py
@@ -1,20 +1,21 @@
 """
 A Printer which converts an expression into its LaTeX equivalent.
 """
-
-from typing import Any, Dict as tDict
+from __future__ import annotations
+from typing import Any, Callable, TYPE_CHECKING
 
 import itertools
 
-from sympy.core import Add, Float, Mod, Mul, Number, S, Symbol
+from sympy.core import Add, Float, Mod, Mul, Number, S, Symbol, Expr
 from sympy.core.alphabets import greeks
 from sympy.core.containers import Tuple
-from sympy.core.function import AppliedUndef, Derivative
+from sympy.core.function import Function, AppliedUndef, Derivative
 from sympy.core.operations import AssocOp
 from sympy.core.power import Pow
 from sympy.core.sorting import default_sort_key
 from sympy.core.sympify import SympifyError
-from sympy.logic.boolalg import true
+from sympy.logic.boolalg import true, BooleanTrue, BooleanFalse
+from sympy.tensor.array import NDimArray
 
 # sympy.printing imports
 from sympy.printing.precedence import precedence_traditional
@@ -27,6 +28,9 @@ from mpmath.libmp.libmpf import prec_to_dps, to_str as mlib_to_str
 from sympy.utilities.iterables import has_variety
 
 import re
+
+if TYPE_CHECKING:
+    from sympy.vector.basisdependent import BasisDependent
 
 # Hand-picked functions which can be used directly in both LaTeX and MathJax
 # Complete list at
@@ -81,7 +85,7 @@ other_symbols = {'aleph', 'beth', 'daleth', 'gimel', 'ell', 'eth', 'hbar',
                      'hslash', 'mho', 'wp'}
 
 # Variable name modifiers
-modifier_dict = {
+modifier_dict: dict[str, Callable[[str], str]] = {
     # Accents
     'mathring': lambda s: r'\mathring{'+s+r'}',
     'ddddot': lambda s: r'\ddddot{'+s+r'}',
@@ -119,7 +123,7 @@ _between_two_numbers_p = (
 )
 
 
-def latex_escape(s):
+def latex_escape(s: str) -> str:
     """
     Escape a string such that latex interprets it as plaintext.
 
@@ -137,7 +141,7 @@ def latex_escape(s):
 class LatexPrinter(Printer):
     printmethod = "_latex"
 
-    _default_settings = {
+    _default_settings: dict[str, Any] = {
         "full_prec": False,
         "fold_frac_powers": False,
         "fold_func_brackets": False,
@@ -162,7 +166,7 @@ class LatexPrinter(Printer):
         "min": None,
         "max": None,
         "diff_operator": "d",
-    }  # type: tDict[str, Any]
+    }
 
     def __init__(self, settings=None):
         Printer.__init__(self, settings)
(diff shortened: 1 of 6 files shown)


sympy/printing/latex.py
```diff
@@ -1615,15 +1615,16 @@ def _print_Order(self, expr):
                 s += self._print(expr.point[0])
         return r"O\left(%s\right)" % s
 
-    def _print_Symbol(self, expr, style='plain'):
-        if expr in self._settings['symbol_names']:
-            return self._settings['symbol_names'][expr]
+    def _print_Symbol(self, expr: Symbol, style='plain') -> str:
+        name: str = self._settings['symbol_names'].get(expr)
```
eagleoflqj, line 1619: minor optimization

sympy/printing/latex.py
```diff
@@ -1879,15 +1880,15 @@ def _print_NDimArray(self, expr):
...
+            left_delim: str = self._settings['mat_delim']
             right_delim = self._delim_dict[left_delim]
             block_str = r'\left' + left_delim + block_str + \
                         r'\right' + right_delim
 
         if expr.rank() == 0:
             return block_str % ""
 
-        level_str = [[]] + [[] for i in range(expr.rank())]
+        level_str: list[list[str]] = [[] for i in range(expr.rank() + 1)]
```
eagleoflqj, line 1891: minor optimization.
Even if I write `level_str: list[list[str]] = [[]] + [[] for i in range(expr.rank())]`, pyright still thinks `level_str` is `list[list[Unknown]]`.

sympy/printing/latex.py
```diff
@@ -1914,7 +1915,7 @@ def _print_NDimArray(self, expr):
 
         return out_str
 
-    def _printer_tensor_indices(self, name, indices, index_map={}):
+    def _printer_tensor_indices(self, name, indices, index_map={}) -> str:
```
ThePauliPrinciple, line 1918: Why doesn't Pyright get this right? Is there any value for the result of `latex_escape(s)` that can make it not return a `str`? If there isn't, this should be reported as a bug there instead of requiring a modification in SymPy

ThePauliPrinciple: *same for all other similar cases

eagleoflqj: `latex_escape` doesn't belong to those `_print_*` functions whose return type should be deduced automatically. Without hinting its parameter pyright can't deduce its return type. With hinting its parameter, pyright can, and I still add explicit type hint to it just for consistency.
`_printer_tensor_indices` is indeed weird. `index_map` has a mutable default value which is a bad practice. If modify it to `index_map: dict` then pyright can get it right.

ThePauliPrinciple: But regardless of the return type of `latex_escape`, the return type will be a string. I don't think `str % anything` returns anything other than a `str`?

eagleoflqj: Which function are you referring to? There is no calls to `latex_escape` or `%` format for `_printer_tensor_indices`.

eagleoflqj: I’m reluctant to open more issues in pyright. As you can see, most likely I’ll get an “as designed” response. It seems their culture that users should always modify their legal python code to fit their tool, otherwise just don’t use it.

hmc-cs-mdrissi:
```python
class Foo(str):
    def __rmod__(self, other: object) -> int:
        return 0


print("x" % Foo())  # Gives 0
```

So yes it's possible for `str % anything` to be not a string.

ThePauliPrinciple: > I’m reluctant to open more issues in pyright. As you can see, most likely I’ll get an “as designed” response. It seems their culture that users should always modify their legal python code to fit their tool, otherwise just don’t use it.

That is a good point. Then maybe the problem is that `%` is used over `.format()`?

ThePauliPrinciple: Does pyright understand these:
```python
class Foo(str):
    def __rmod__(self, other: object) -> int:
        return 0


print("x" % Foo())  # Gives 0
print(type("{0}".format(Foo())))
print(type(f"{Foo()}"))
```
I think that fstrings are very elegant in this way.

eagleoflqj: Thank you. This is the counter-example I expected from the pyright issue.

eagleoflqj: > Why doesn't Pyright get this right?

Some are because the return value of a `_print_*` is a call to a helper function that is defined afterward, and I added type hints from the beginning to the end, so at the time I check the deduced return type of `_print_*` pyright cannot give correct answer.
Some are because pyright needs parameter type, although when I hover on any return value in vscodium, pyright tells me it's `str`.
Now there's only one function `_print_Function` that has to be explicitly typed. And this is obviously the reason:
https://github.com/sympy/sympy/blob/918eb81f4bb3ae062cef98a9dcb9ecda058e33d5/sympy/printing/latex.py#L940

sympy/tensor/array/ndim_array.py
```diff
@@ -143,6 +143,9 @@ def __new__(cls, iterable, shape=None, **kwargs):
         from sympy.tensor.array import ImmutableDenseNDimArray
         return ImmutableDenseNDimArray(iterable, shape, **kwargs)
 
+    def __getitem__(self, index):
+        ...
```
oscarbenjamin, line 147: What does this do?

ThePauliPrinciple: Wouldn't this mean that the code (either before or after this PR) would fail if an `NDimArray` (and not one of its subclasses) was printed for some reason and has `.rank == 0`?

eagleoflqj: The `__new__` function returns an `ImmutableDenseNDimArray`, so if you call `NDimArray([])` it won't return a base class instance.

ThePauliPrinciple: I was under the impression that baseclasses in sympy usually raise NotImplementedError if their methods are not well defined in the base class (see e.g. `Number` in core.numbers)

eagleoflqj: I don't think `NotImplementedError` is a proper way for this. It should be used for real not-implemented branch of an algorithm. If we do want to prevent a base class instantiation, we should use `abc`:
```py
>>> import abc
>>> class C(abc.ABC):
...     @abc.abstractmethod
...     def foo(self):
...             return 0
... 
>>> C()
TypeError: Can't instantiate abstract class C with abstract method foo
```

eagleoflqj:
Any suggestions? Otherwise I'll merge it.

sympy/tensor/array/ndim_array.py
ThePauliPrinciple: This still hasn't been adressed

eagleoflqj: Thanks. Converted to abstractmethod.

sympy/core/core.py
```diff
@@ -64,7 +66,7 @@ def __delattr__(self, name):
 all_classes = set()
 
 
-class BasicMeta(type):
+class BasicMeta(ABCMeta):
```
eagleoflqj, line 69: To make abstract class this change is needed.

ThePauliPrinciple: I don't think we should make this kind of change to sympy just to get type checking right for this very particular case. 

If it is a good idea to include this, it should be a seperate PR, that also makes this consistent across the entire library (e.g. then the use of `NotImplementedError` should be replaced consistently).

eagleoflqj: That’s also not what I aimed for at the beginning. That’s why I left __getitem__’s body empty previously.

eagleoflqj: OK. I changed my mind. `NotImplementedError` may fit here, indicating that a concrete method is not implemented (instead of this abstract method).

PR #23501, opened 2022-05-15
maint(ci): bump python to 3.10 and add 3.11b1

oscarbenjamin:

Closes #22732

#### Release Notes

NO ENTRY

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -320,7 +320,7 @@ jobs:
   # -------------------- Build the html/latex docs ----------------- #
 
   sphinx:
-    #needs: [doctests-latest, tests1-latest, tests2-latest]
+    needs: [doctests-latest, tests1-latest, tests2-latest]
 
     runs-on: ubuntu-20.04
     steps:
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -150,9 +150,15 @@ jobs:
     needs: code-quality
 
     runs-on: ubuntu-20.04
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
+      fail-fast: false
       matrix:
         python-version: ['3.10', 'pypy-3.8']
+        experimental: [false]
+        include:
+          - python-version: '3.11.0-alpha - 3.11'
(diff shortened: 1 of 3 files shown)


oscarbenjamin:
Everything works with python 3.10 except the optional dependency job. Looks like the optional dependencies fail on Python 3.10 in the llvmjit code. There's no useful output from the test runner (it seems to just crash). I think I've seen that error before from Travis before most travis jobs were disabled. Probably the temporary fix there is to disable whatever tests are crashing the test runner so we can at least see that the other tests pass.

For 3.11b1 the optional dependency test fails because there are no wheels for numpy and building numpy 1.22.3 from source fails with a compiler error.

The 3.11b1 doctest and test2 jobs fail because of some problem with pickling. The change comes from https://github.com/python/cpython/issues/70766 and gives errors like
```python
___________________ sympy.polys.polyutils.PicklableWithSlots ___________________
File "/home/runner/work/sympy/sympy/sympy/polys/polyutils.py", line 455, in sympy.polys.polyutils.PicklableWithSlots
Failed example:
    some2 = loads(dumps(some))
Exception raised:
      File "/opt/hostedtoolcache/Python/3.11.0-beta.1/x64/lib/python3.11/doctest.py", line 1346, in __run
        exec(compile(example.source, filename, "single",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "<doctest sympy.polys.polyutils.PicklableWithSlots[9]>", line 1, in <module>
        some2 = loads(dumps(some))
                      ^^^^^^^^^^^
      File "/home/runner/work/sympy/sympy/sympy/polys/polyutils.py", line 474, in __getstate__
        d.update(c.__getstate__(self, c))
                 ^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/runner/work/sympy/sympy/sympy/polys/polyutils.py", line 474, in __getstate__
        d.update(c.__getstate__(self, c))
                 ^^^^^^^^^^^^^^^^^^^^^^^
    TypeError: object.__getstate__() takes no arguments (1 given)
```
I've just pushed a quick fix for that here.

PR #23507, opened 2022-05-16
Replaced i with a dummy symbol with no assumptions for get_residue_factor

### LINKED ISSUE #23491: Incorrect symbolic residue summation
This summation should clearly be a positive number:
```python
In [2]: s = Sum(1/(n**2+1), (n, 1, oo))

In [3]: s
Out[3]: 
  ∞         
 ____       
 ╲          
  ╲     1   
   ╲  ──────
   ╱   2    
  ╱   n  + 1
 ╱          
 ‾‾‾‾       
n = 1       

In [4]: s.doit() # <--- incorrect
Out[4]: -1/2
```
The correct result seems to be:
```python
In [5]: (pi*coth(pi)-1)/2
Out[5]: 
  1   π⋅coth(π)
- ─ + ─────────
  2       2    

In [6]: N(_)
Out[6]: 1.07667404746858

In [7]: N(s)
Out[7]: 1.07667404746858
```
The bug is in `eval_sum_residue` somewhere.

ghost:
A dummy symbol with no assumptions set for get_residue_factor

Fixes #23491 

In _sympy/concrete/summations.py_ **get_residue_factor(**),  'i' was used with assumption set **integer = True** but the problem is that **cot(S.Pi * i) gives zoo(complex infinity)** therefore we need to use a dummy symbol with no assumption set to get_residue_factor to calculate the residue around the pole rather than evaluate the expression at the pole.

#### Other comments
- [x] Opened on a seperate branch
- [x] Has unit tests
- [x] Follows Documentation Guidelines
- [x] Has Release Notes

#### Release Notes

* concrete
   * A bug leading to incorrect results when evaluating some infinite summations was fixed. Previously an incorrect result might be found in some cases if the summation symbol was defined with the `integer=True` assumption set.

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -20,7 +20,7 @@ jobs:
 
       - uses: actions/setup-python@v3
         with:
-          python-version: '3.9'
+          python-version: '3.10'
       - run: python -m pip install --upgrade pip
       - run: pip install mpmath flake8
 
@@ -57,7 +57,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v3
         with:
-          python-version: '3.9'
+          python-version: '3.10'
       - run: python -m pip install --upgrade pip
       - run: pip install mpmath mypy
 
@@ -74,7 +74,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v3
         with:
-          python-version: '3.9'
+          python-version: '3.10'
       - run: python -m pip install --upgrade pip
       - run: pip install sphinx-lint
 
@@ -93,7 +93,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-python@v3
         with:
-          python-version: '3.9'
+          python-version: '3.10'
       - run: python -m pip install --upgrade pip
       - run: pip install mpmath
       - run: bin/mailmap_check.py --skip-last-commit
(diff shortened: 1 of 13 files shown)


KuldeepBorkar:
I will add a test case once all the checks are successful,
I am actually not able to decide with every PR that I should add Release Notes or not could someone please help me with that?
Whenever I make some changes related to the documentation I use NO ENTRY but with changes other than documentation I am confused ; )

oscarbenjamin:
The release note could be
```
* concrete
  * A bug leading to incorrect results when evaluating some infinite summations
    was fixed. Previously an incorrect result might be found in some cases if
    the summation symbol was defined with the `integer=True` assumption set.
```

KuldeepBorkar:
Thank you @oscarbenjamin : )

KuldeepBorkar:
It says **Python version 3.8 or above is required for SymPy** when I run bin/test_optional_dependencies.py
locally but SymPy supports Python 3.7 I think, (optional-dependencies check is failing for me the first time)

oscarbenjamin:
The next release of sympy will not support 3.7 so "support" is already dropped on master. Probably you can just edit that line in the script though so that it still runs though. It isn't actually necessary to run that script though: if you have the dependencies installed then the ordinary test runner will run the same tests.

The failures in the optional dependencies are only seen under Python 3.10 and 3.11 though so to reproduce them you need to install Python 3.10 or Python 3.11b1 (and also many of the optional dependencies).

More importantly those failures are unrelated to this PR. Those jobs fail on the master branch which is why they are not required to pass for the PR to be merged.

KuldeepBorkar:
Thank you @oscarbenjamin for your help (I always thought release notes should be something like one liner by looking at the examples mentioned while writing comment for a PR, also now I think writing Release Notes is a good practice now instead of simply writing NO ENTRY if change is something other than documentation)

@eagleoflqj  I have added the test case now : )

PR #13463, opened 2017-10-16
Bidirectional limits

gxyd:
Commits takes from PR #11694. @asmeurer approved that PR, but it has been sometime since those changes were made, and the PR didn't receive many comments.

I have taken partial commits from the PR since the rest of commits in PR #11694 rather only revert these changes. I have closed the other PR.

[Edit]: Related issue where discussion was there #11667.

Issue #11667, opened 2016-09-27
limit(1/x, x, 0) == oo ??


latot:
Hi, well, what definition of limit are you using? or this is a bug.

```
>>> limit(1/x, x, 0)
oo
>>> limit(1/x, x, 0, '+')
oo
>>> limit(1/x, x, 0, '-')
-oo
```

I search if this issue exist and i can't found it, if i search bad close it please.

Thx. Cya.

viditjain1:
I want to work on this. Can someone give me a hint on where to start..?

latot:
I don't know to much about Sympy, but the logic say to me, when you don't send a direction you should calc the limit in both sides, the compare it, if are equals return the value, if not return nan.

asmeurer:
There are some older issues about this, if you search. It would be nice to make limit bidirectional by default.

viditjain1:
bidirectional by default meaning the same logic that is described in the previous comment by latot?

asmeurer:
I think it would make more sense to raise an exception when the different sided limits differ.

latot:
Hi, sadly i'm not agree, conceptually the limit of an discontinuous functions is simply not defined, i think that is nan, or the expression you like, but an exception i think its too heavy, you can check the limit basically for tree reasons, get the right limit, get the left limit and get the limit of a point, in this last case its the same to check if the limit exist, actually Sympy works with discontinuous functions, at least to me don't have to much sense send the exception with this in mind.

Thx.Cya.

asmeurer:
The nice thing about an exception is that it could contain a message indicating to the user why the limit failed. `nan` is, in some sense, a non-failure state. It is designed to propagate to further calculations without error.

latot:
mmm, if i'm right.. ```zoo``` represent an indeterminate point?, the main reason of why i don't like the exception is because if you use that the must start checking exceptions...

so, can be ```zoo``` an option? or some expression?

only in a practical case, i start working with discontinuos function, think in piecewise or ```1/x``` format,  if i try calc limits or diff, i will start getting a lot of exceptions...

asmeurer:
`zoo` is the value of the complex limit of `1/x`, but it doesn't make sense as a generic value. You were right that `nan` is the most correct value for "doesn't exist". 

I don't see why you wouldn't need to check for `nan` too. If you are taking a limit that doesn't exist, you'd want to know that fact. Maybe I'm not seeing your use case properly where it would be more useful to have a nan propagate.

[...]

PR #23516, opened 2022-05-19
limit calculates bi-directional limit by default

eagleoflqj:

This is talked about several times, e.g. https://github.com/sympy/sympy/issues/11667#issuecomment-636179241 and https://github.com/sympy/sympy/pull/13463#issuecomment-338243685. The current behavior is very [confusing](https://www.reddit.com/r/math/comments/44j0nm/sympy_gamma_good_wolfram_alpha_alternative/czqz4bc/?utm_source=reddit&utm_medium=web2x&context=3) and contradicts with https://docs.sympy.org/latest/tutorial/calculus.html#limits
Change public APIs to calculate bi-directional limit by default.
Some internal usages without an explicit direction parameter are also changed.

#### Release Notes

* series
  * BREAKING: limit calculates bi-directional limit by default

### DIFF
--- a/doc/src/tutorial/calculus.rst
+++ b/doc/src/tutorial/calculus.rst
@@ -255,9 +255,9 @@ counterpart, ``Limit``.  To evaluate it, use ``doit``.
 
     >>> expr = Limit((cos(x) - 1)/x, x, 0)
     >>> expr
-         ⎛cos(x) - 1⎞
-     lim ⎜──────────⎟
-    x─→0⁺⎝    x     ⎠
+        ⎛cos(x) - 1⎞
+    lim ⎜──────────⎟
+    x─→0⎝    x     ⎠
     >>> expr.doit()
     0
 
--- a/sympy/core/expr.py
+++ b/sympy/core/expr.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import Literal, TYPE_CHECKING
 from collections.abc import Iterable
 from functools import reduce
 import re
@@ -22,6 +22,7 @@ from mpmath.libmp.libintmath import giant_steps
 
 if TYPE_CHECKING:
     from .numbers import Number
+    from .symbol import Symbol
 
 from collections import defaultdict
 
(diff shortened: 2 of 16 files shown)


anutosh491:
This might take some effort and time as currently almost all leading term and series methods for all functions assume that the limit or the series is being calculated from the right by default .

anutosh491:
@eagleoflqj  If this is only an API change and doesn't solve any bugs then don't you think this should involve a deprecation somehow ??  users should have a chance to adapt their code to changes in the public API and any downstream code should continue to work in the same way until the deprecation period is expired.

jksuom:
I am not sure this would be a good idea. Bidirectional limits are not used in the library. I think that the default should be the current one-sided limit to avoid having to add `dir='+'` everywhere.

anutosh491:
> I am not sure this would be a good idea. Bidirectional limits are not used in the library. I think that the default should be the current one-sided limit to avoid having to add `dir='+'` everywhere.

+ 1  I too don't really like idea of having Bidirectional limits by default but the comments/references pointed out in the OP seemed somewhat valid hence didn't really speak out before !

eagleoflqj:
> Bidirectional limits are not used in the library.

SymPy’s code is not only used internally. Public APIs shouldn’t provide confusing behaviors for users just because of its internal implementation. Most users won’t expect a limit at a point is evaluated single-side by default. Why? Because math formulas don’t do it this way. We have to put a + or - explicitly to show we want single-side limit. And that’s why currently there is inconsistency between doc and implementation.

> the default should be the current one-sided limit to avoid having to add dir='+' everywhere.
I don’t think that a bigger problem than providing confusing behavior.

eagleoflqj:
> @eagleoflqj If this is only an API change and doesn't solve any bugs then don't you think this should involve a deprecation somehow ??

Thanks for pointing that out. Yes, there should be a deprecation warning if users don’t provide a dir arg.

asmeurer:
Not sure I like deprecating not providing the dir arg since we don't intend to make it required, although I can't think of any other way to change this without a warning.

oscarbenjamin:
Also I wonder if something better can be done with bidirectional limits that don't exist. This case for example might exist depending on the value of `y`:
```python
In [7]: limit(y*x/abs(x), x, 0, dir='+-')
...
ValueError: The limit does not exist since left hand limit = -y and right hand limit = y
```

eagleoflqj:
I fixed it at SymPy Beta https://github.com/eagleoflqj/sympy_beta/commit/8660f9042b92b28073e4f868b26d4865dd15b4d1 to address the complaint at reddit.
Now https://sympy-beta.vercel.app/input/limit(tan(x),%20x,%20pi%2F2) calculates bi-direction limit.

PR #23521, opened 2022-05-19
Fixing some bugs in str and latex printers for MatAdd

Upabjojr:
NO ENTRY

### DIFF
--- a/sympy/printing/latex.py
+++ b/sympy/printing/latex.py
@@ -1761,17 +1761,14 @@ class LatexPrinter(Printer):
                 return r"%s^{\dagger}" % s
 
     def _print_MatMul(self, expr):
+        from sympy import MatMul
 
-        parens = lambda x: self.parenthesize(x, precedence_traditional(expr),
-                                             False)
+        # Parenthesize nested MatMul but not other types of Mul objects:
+        parens = lambda x: self._print(x) if isinstance(x, Mul) and not isinstance(x, MatMul) else \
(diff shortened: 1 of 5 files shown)


sympy/printing/latex.py
```diff
@@ -1772,7 +1772,7 @@ def _print_MatMul(self, expr):
         else:
             args = list(args)
 
-        if isinstance(expr, MatMul) and expr.could_extract_minus_sign():
+        if len(args) > 0 and args[0].is_real and (args[0] < 0) == True:
```
Upabjojr, line 1775: @smichr I am reverting this as `.could_extract_minus_sign()` introduced a double minus sign bug to the LaTeX printer. Plus, negating the expression may some time trigger some non-trivial transformations that make the printed LaTeX expression structurally different from its internal representation. This may cause difficulties in manipulating expressions in Jupyter notebooks.

asmeurer: What does the logic for this look like in the normal Mul printer? This ought to emulate that.

In general, I would say printers should avoid doing things that could be expensive. Even the use of assumptions here is something I'm a little uneasy about. It could mean odd behavior in the printer depending on whether SymPy is able to compute `is_real` or not (and btw, why not just `args[0].is_negative`?).

Upabjojr: Turns out there was a bug in `.could_extract_minus_sign()` in `MatMul`.

PR #23522, opened 2022-05-19
Fix MatAdd.expand( ) not expanding completely

Upabjojr:
NO ENTRY

### DIFF
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -1963,6 +1963,10 @@ class Basic(Printable, metaclass=ManagedProperties):
     def could_extract_minus_sign(self):
         return False  # see Expr.could_extract_minus_sign
 
+    @classmethod
+    def _evaluate(cls, expr):
+        return expr
+
 
 class Atom(Basic):
     """
--- a/sympy/core/basic.py
+++ b/sympy/core/basic.py
@@ -1963,10 +1963,6 @@ class Basic(Printable, metaclass=ManagedProperties):
     def could_extract_minus_sign(self):
         return False  # see Expr.could_extract_minus_sign
 
-    @classmethod
-    def _evaluate(cls, expr):
-        return expr
-
 
 class Atom(Basic):
     """
--- a/sympy/matrices/expressions/matadd.py
+++ b/sympy/matrices/expressions/matadd.py
@@ -52,11 +52,16 @@ class MatAdd(MatrixExpr, Add):
             validate(*args)
 
         if evaluate:
-            if not any(isinstance(i, MatrixExpr) for i in args):
-                return Add(*args, evaluate=True)
-            obj = canonicalize(obj)
+            obj = cls._evaluate(obj)
 
         return obj
+
+    @classmethod
+    def _evaluate(cls, expr):
+        args = expr.args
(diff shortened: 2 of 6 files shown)


sympy/core/basic.py
```diff
@@ -1963,6 +1963,10 @@ def _sage_(self):
     def could_extract_minus_sign(self):
         return False  # see Expr.could_extract_minus_sign
 
+    @classmethod
+    def _evaluate(cls, expr):
+        return expr
```
oscarbenjamin, line 1968: I don't think that this method should be added to Basic. There are already too many methods on Basic that are added haphazardly and then used by only a tiny handful of subclasses. New methods should not be added without defining and documenting what the overall design is supposed to be.

sympy/matrices/expressions/matadd.py
```diff
@@ -52,12 +52,17 @@ def __new__(cls, *args, evaluate=False, check=False, _sympify=True):
...
+            obj = cls._evaluate(obj)
 
         return obj
 
+    @classmethod
+    def _evaluate(cls, expr):
+        args = expr.args
+        if not any(isinstance(i, MatrixExpr) for i in args):
+            return Add(*args, evaluate=True)
+        return canonicalize(expr)
```
oscarbenjamin, line 64: Rather than testing for MatrixExpr it would be better to use `kind` here.

This looks like masking something like a bug though. Why would a MatAdd be constructed with non-matrix args?

This would make sense for a MatMul since a MatMul can always have some non-matrix args. There's never any reason to have any non-matrix args in a MatAdd though.

sympy/core/basic.py
Upabjojr: This is related to the well-known problem of SymPy that it is not well defined when expressions evaluate and when they do not. `.expand( )` on matrix expressions does not completely expand, as the constructors in the matrix expressions module do not evaluate by default. This behaviour makes sense (I would rather change the behaviour of the rest of SymPy, the matrix expressions module has a robust design).

Maybe singledispatch could avoid the problem of too many overloaded methods in `Basic` ?

sympy/matrices/expressions/matadd.py
Upabjojr: I have just copied the previously existing code. I'm not sure about changing the behaviour of `MatAdd` in this PR.

sympy/core/basic.py
oscarbenjamin: I don't think singledispatch vs methods really makes any difference. The point is really that every time core logic gets complicated by needing to call additional functions/methods it becomes hard to improve that core logic in future e.g. to make it faster or simpler or more consistent.

Here if the goal is to change the behaviour of `MatAdd.expand` then why isn't the logic in the `MatAdd.expand` function? At least that would localise it to where it is needed.

Upabjojr: > Here if the goal is to change the behaviour of `MatAdd.expand` then why isn't the logic in the `MatAdd.expand` function? At least that would localise it to where it is needed.

That could be an option. Though, are we giving up completely the idea of not evaluating all `__new__` constructors by default?

PR #23526, opened 2022-05-19
Some fixes to eval() in Function subclasses

asmeurer:
- Made the return value of `eval()` automatically sympified (so things like `return 1` will result in `Integer(1)`).
- Includes some fixes to 3-argument `pow` in the code that were necessary after this.
- Raise an exception if `eval` is not defined as a `@classmethod` (common mistake).

#### Release Notes

- core
  - The return values of `eval` on `Function` subclasses are now automatically sympified.
  - `Function` subclasses now raise an exception if the `eval` method is not defined as a `@classmethod`.

### DIFF
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -47,7 +47,7 @@ from .operations import LatticeOp
 from .parameters import global_parameters
 from .rules import Transform
 from .singleton import S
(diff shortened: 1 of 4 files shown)


oscarbenjamin:
Looks good

Issue #19528, opened 2020-06-11
Remove antlr generated files from the repository


sylee957:
I think that we should remove the antlr generated files because it is not making the parser really standalone, and we always need to get huge changes to the script files whenever the grammar files are generated.

The other problem of antlr is that they contain the version information
https://github.com/sympy/sympy/blob/de62108d34a1265cc1b2b9c5155d6070b842b672/sympy/parsing/latex/_antlr/latexlexer.py#L455
Although for user side, I see it only raises some warnings for different versions and runs without much critical problems, but it is worse for the development side because using different version of `antlr` gives differently structured script files.

https://github.com/sympy/sympy/blob/de62108d34a1265cc1b2b9c5155d6070b842b672/sympy/parsing/latex/_build_latex_antlr.py#L78

And also, I see we are already blindly removing the line 1~2 of the generated antlr script files because it contains personal information. 
So I have doubts whether antlr's automatic code generation was ever designed to be compatible with team development.

So I think that the only good option is to generate the antlr script files in runtime if the user have antlr installed in environment or is importable, and it can also be a good option to have sympy installable with antlr if some options are specified.
But I'm not sure how to make it possible with setup scripting.

asmeurer:
IMHO we should move from antlr to a pure Python library. The only advantage of antlr is performance, and that isn't that important for this use-case.

asmeurer:
CC @bollwyvl. I know we discussed these issues a lot when the parser was added.

asmeurer:
> and that isn't that important for this use-case.

At least for the LaTeX parser. It may be important for the other parsers.

asmeurer:
https://github.com/sympy/sympy/issues/14004

bollwyvl:
I chimed in over on #14004, but for a little more depth here: I'm still really a grammar (and LaTeX (and sympy)) dilettante, but we have a reference implementation (really the whole goal of that initial work) to test against for both correctness and performance, and left a way to provide additional implementations. I might be able to finagle some time to work on prototyping something, and if I did, I would probably take a crack at it with [lark-parser](https://github.com/lark-parser/lark). The things that seem useful:

- has active, receptive maintainers
- pretty good docs
- has no runtime dependencies beyond python's stdlib (parsimonious still needs `six`)
- shows strong performance
- handles ambiguities that PEG parsers cannot
- uses a dedicated, self-described format, 
  - that cannot include implementation details (e.g. inline python) 
  - which can be stored inline in a py file
[...]

PR #23535, opened 2022-05-22
parse_latex support f'(x)

### LINKED ISSUE #23515: parse_latex f'(x)
parse_latex should handle f'(x), f''(x) and so on. But currently it throws an exception.
There are 2 possible solutions.
1. parse as `Function("f'")(x)`, and let users substitute it to `Derivative(f(x), x)` on demand.
2. parse as `Derivative(f(u), u).subs(u, x)`, where `u` is a `Dummy`. The substitution is needed for parsing `f'(0)`, but that case you will get a `Subs` object, and if you convert it back to latex it will be very different from `f'(0)`.

I propose to implement solution 1. What do you think?

eagleoflqj:

Closes #23515

Single quotes with subscripts are legal latex, and their order doesn't matter.
#### Release Notes

* parsing
  * Latex parser supports f'(x)


ThePauliPrinciple:
It looks good to me, but I do have a question about the large amount of changes for this seemingly trivial feature. Why is it needed? It seems to me that a lot of lines with `state = #` have changed. I don't understand why that is needed.

eagleoflqj:
The `latexlexer.py` and `latexparser.py` are generated by ANTLR from `LaTeX.g4`, by running `python setup.py antlr` (Everytime I run this the head comment of the 2 generated files are corrupted, so I have to delete these 2 file before re-generating them).
Ideally generated files shouldn't be included in SCM. Discussion happens at #19528.
There may be a better way to provide out-of-the-box usage of latex parser. That is, for developers using SymPy git version, let them generate themselves. But when releasing, generate these files and include them in the SymPy wheel.

Issue #19453, opened 2020-05-28
Limit changes from simplification of original expression


B-Lorentz:
I have an ugly symbolic expression:
[...]

PR #23539, opened 2022-05-24
Changed Float handling in limits

0sidharth:

#23499 

If a float is present in the original limit expression, the returned
result should also have floats

#### Release Notes

* series
  * If the argument to limit contains floats, the final answer will also contain floats.

### DIFF
--- a/sympy/series/limits.py
+++ b/sympy/series/limits.py
@@ -1,6 +1,7 @@
 from sympy.calculus.accumulationbounds import AccumBounds
 from sympy.core import S, Symbol, Add, sympify, Expr, PoleError, Mul
 from sympy.core.exprtools import factor_terms
+from sympy.core.function import nfloat
 from sympy.core.numbers import Float, _illegal
 from sympy.functions.combinatorial.factorials import factorial
 from sympy.functions.elementary.complexes import (Abs, sign)
@@ -261,13 +262,6 @@ class Limit(Expr):
                         return expr.args[0] if abs_flag else S.One
             return expr
 
-        if e.has(Float):
-            # Convert floats like 0.5 to exact SymPy numbers like S.Half, to
-            # prevent rounding errors which can lead to unexpected execution
-            # of conditional blocks that work on comparisons
-            # Also see comments in https://github.com/sympy/sympy/issues/19453
-            from sympy.simplify.simplify import nsimplify
-            e = nsimplify(e)
         e = set_signs(e)
 
 
@@ -347,6 +341,14 @@ class Limit(Expr):
 
         l = None
 
+        inexact = e.has(Float)
+        if inexact:
+            # Convert floats like 0.5 to exact SymPy numbers like S.Half, to
+            # prevent rounding errors which can lead to unexpected execution
+            # of conditional blocks that work on comparisons
+            # Also see comments in https://github.com/sympy/sympy/issues/19453
+            from sympy.simplify.simplify import nsimplify
+            e = nsimplify(e)
         try:
             if str(dir) == '+-':
                 r = gruntz(e, z, z0, '+')
@@ -366,4 +368,4 @@ class Limit(Expr):
             if r is None:
                 return self
 
-        return r
+        return nfloat(r) if inexact else r
--- a/sympy/series/tests/test_limits.py
+++ b/sympy/series/tests/test_limits.py
@@ -2,7 +2,7 @@ from itertools import product
 
 from sympy.concrete.summations import Sum
 from sympy.core.function import (Function, diff)
-from sympy.core import EulerGamma
+from sympy.core import EulerGamma, N
 from sympy.core.numbers import (E, I, Rational, oo, pi, zoo)
 from sympy.core.singleton import S
 from sympy.core.symbol import (Symbol, symbols)
@@ -137,7 +137,7 @@ def test_log():
     tau, x = symbols('tau x', positive=True)
     # The value of manualintegrate in the issue
     expr = tau**2*((tau - 1)*(tau + 1)*log(x + 1)/(tau**2 + 1)**2 + 1/((tau**2\
-            + 1)*(x + 1)) - (-2*tau*atan(x/tau) + (tau**2/2 - 1/2)*log(tau**2\
+            + 1)*(x + 1)) - (-2*tau*atan(x/tau) + (tau**2/2 - S.Half)*log(tau**2\
             + x**2))/(tau**2 + 1)**2)
     assert limit(expr, x, oo) == pi*tau**3/(tau**2 + 1)**2
 
@@ -192,7 +192,7 @@ def test_floor():
 
     # https://github.com/sympy/sympy/issues/14478
     assert limit(x*floor(3/x)/2, x, 0, '+') == Rational(3, 2)
-    assert limit(floor(x + 1/2) - floor(x), x, oo) == AccumBounds(-S.Half, S(3)/2)
+    assert limit(floor(x + S.Half) - floor(x), x, oo) == AccumBounds(-S.Half, S(3)/2)
 
 
 def test_floor_requires_robust_assumptions():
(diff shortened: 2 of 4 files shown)


0sidharth:
@oscarbenjamin Not sure if this is the expected change and the type of handling you were looking for. Please do take a look and suggest changes if any.

sympy/series/tests/test_limits.py
```diff
@@ -139,7 +139,7 @@ def test_log():
     expr = tau**2*((tau - 1)*(tau + 1)*log(x + 1)/(tau**2 + 1)**2 + 1/((tau**2\
             + 1)*(x + 1)) - (-2*tau*atan(x/tau) + (tau**2/2 - 1/2)*log(tau**2\
             + x**2))/(tau**2 + 1)**2)
-    assert limit(expr, x, oo) == pi*tau**3/(tau**2 + 1)**2
+    assert limit(expr, x, oo) == N(pi)*tau**3/(tau**2 + 1)**2
```
jksuom, line 142: It seems that this is inexact because of 1/2 in expr. Maybe we could make that exact without essentially changing the test.

sympy/series/limits.py
```diff
@@ -261,7 +262,8 @@ def set_signs(expr):
                         return expr.args[0] if abs_flag else S.One
             return expr
 
-        if e.has(Float):
+        inexact = e.has(Float)
+        if inexact:
             # Convert floats like 0.5 to exact SymPy numbers like S.Half, to
             # prevent rounding errors which can lead to unexpected execution
             # of conditional blocks that work on comparisons
```
oscarbenjamin, line 269: Maybe the proper fix is to identify where these comparisons are and fix them to work with floats.

sympy/series/tests/test_limits.py
0sidharth: I had thought so too but wasn't sure if the test should be changed, or the inexact 1/2 should intentionally have been preserved... I will change it to S.Half

sympy/series/limits.py
0sidharth: From #19453
> The cause of this is a rounding error. In this block of code
> 
> https://github.com/sympy/sympy/blob/b4777fdcef467b7132c055f8ac2c9a5059e6a145/sympy/core/mul.py#L1910-L1915
> 
> `sum(powers)` returns the sum of `(-0.333333333333333, 0.333333333333333)` as `-5.55111512312578e-17`.
> An alternative to get the correct answer is to use exact SymPy numbers, i.e. `S.Half` instead of `0.5`.
> 
> ```python
> In [1]: beta = Symbol("beta", real=True, positive=True)
>    ...: h = Symbol("h", real=True, positive=True)
>    ...: m = Symbol("m", real=True, positive=True)
>    ...: w = Symbol("omega", real=True, positive=True)
>    ...: g = Symbol("g", real=True, positive=True)
>    ...: 
>    ...: e = exp(1)
>    ...: q = 3*h**2*beta*g*e**(S.Half*h*beta*w)
>    ...: p = m**2*w**2
>    ...: s = e**(h*beta*w) - 1
>    ...: Z = -q/(4*p*s) - q/(2*p*s**2) - q*(e**(h*beta*w)+1)/(2*p*s**3) + e**(S.Half*h*beta*w)/s
>    ...: 
>    ...: E = -diff(log(Z), beta)
>    ...: limit(E - S.Half*h*w, beta, oo), limit(E.simplify() - S.Half*h*w, beta, oo)
> Out[1]: (0, 0)
> ```

I can't think of methods to change this comparison and make it work with floats without hardcoding a tolerance value. It also doesn't appear that much can be changed before the comparison (I doubt there is an alternative to `sum` that would return 0, as that is not expected from Python).

Do you have any suggestions?

jksuom: > we shouldn't apply nsimplify to the whole expression but rather target the relevant parts (the exponents).

This would fix the exponents making test_issue_19453 pass:
```
--- a/sympy/core/mul.py
+++ b/sympy/core/mul.py
@@ -1964,6 +1964,8 @@ def coeff_exp(term, x):
             ords3 = [coeff_exp(term, x) for term in fac]
             coeffs, powers = zip(*ords3)
             power = sum(powers)
+            if abs(power) < sum(abs(_) for _ in powers)/10**15:
+                power = S.Zero
             if (power - n).is_negative:
                 res += Mul(*coeffs)*(x**power)
```
However, the test will take 54.8 s vs. 3.1 s on the current current master. The reason for the slowdown seems to be the same as in this comment: https://github.com/sympy/sympy/blob/6ea48db2bd108a0c5b3c53aa2f1ef9e5d617bf3c/sympy/series/gruntz.py#L667-L671 
It seems that the series expansions are currently working but slow.

It also looks like applying nsimplify to exponents only may not suffice as gruntz will create new exponents here:  https://github.com/sympy/sympy/blob/6ea48db2bd108a0c5b3c53aa2f1ef9e5d617bf3c/sympy/series/gruntz.py#L637-L646
(c will usually be a Float if there are other Floats.)

So it seems to me that applying nsimplify to the whole expression would be the natural choice at least in gruntz.

0sidharth:
The tests for this example (Line 450 in beam_problems.rst) - 
```
>>> from sympy.physics.continuum_mechanics.beam import Beam
>>> from sympy import symbols
>>> E,I,M,V = symbols('E I M V')
>>> b = Beam(8, E, I)
>>> E,I,R1,R2 = symbols('E I R1 R2')
>>> b.apply_load(R1, 1, -1)
>>> b.apply_load(R2, 7, -1)
>>> b.apply_load(10, 0, 0, end=2)
>>> b.apply_load(20, 5, -1)
>>> b.apply_load(8, 7.5, -1)
>>> b.solve_for_reaction_loads(R1, R2)
>>> b.reaction_loads
{R₁: -26, R₂: -22}
>>> b.load
      0             -1             0             -1             -1              -1
10⋅<x>  - 26⋅<x - 1>   - 10⋅<x - 2>  + 20⋅<x - 5>   - 22⋅<x - 7>   + 8⋅<x - 7.5>

>>> b.shear_force()
        1             0             1             0             0              0
- 10⋅<x>  + 26⋅<x - 1>  + 10⋅<x - 2>  - 20⋅<x - 5>  + 22⋅<x - 7>  - 8⋅<x - 7.5>
```
reveal that the coefficients are inconsistent between floats and integers, as `solve_for_reaction_loads` calls `limit`, which will henceforth return a float, for float inputs. This float is mixed with integers (calculated without using `limit`). I wonder if we can add `nsimplify` call to https://github.com/sympy/sympy/blob/109ea4017bfdd2819916c881eadb0cfbc072efa9/sympy/physics/continuum_mechanics/beam.py#L830-L831 to tackle the change in limit behavior. cc @moorepants

0sidharth:
This is probably not the best way to solve the issue with `beam_problems`, but I think it is fine for now instead of waiting. Can improve in a later patch.

oscarbenjamin:
I'm adding the 1.11 milestone.

> This is probably not the best way to solve the issue with `beam_problems`, but I think it is fine for now instead of waiting

I agree but can you add comments explaining why `nsimplify` is being used there? If there is a problem somewhere in the code and we add something to work around it then there needs to be clear comments explaining that so that someone reading it can understand why it is there without just running all the tests.

@jksuom @ThePauliPrinciple does this look good to you?

jksuom:
Would it suffice to apply nsimplify only in gruntz?

oscarbenjamin:
Another question is whether it should be only exponents that get nsimplified. I don't know the limits code so well so I don't know if floats in other parts of the expression would end up leading to floats in the exponents. Otherwise if the problem is just to do with exponents it might make sense to only nsimplify those.

jksuom:
> if floats in other parts of the expression would end up leading to floats in the exponents.

That can happen in `grunz.rewrite` on the last line here: https://github.com/sympy/sympy/blob/596257ecb35453b8450470e80b450f2e0b118924/sympy/series/gruntz.py#L637-L646
The exponent `c` of `wsym` is computed from the limit of two partial expressions on the second line. (`f.exp` and `g.exp` are arguments of `exp` functions created by the `moveup2` function out of some parts of the original expression.) That is why I think that it might suffice to apply `nsimplify` in `gruntz` alone instead of `Limit.doit`,

0sidharth:
I still have to address Kalevi's latest comment, but haven't been able to find the time to do it

jksuom:
This looks good but I would like to see if it would suffice to apply nsimplify in gruntz alone.

oscarbenjamin:
If there isn't time to address the suggestion about applying this only in Gruntz is it better to defer this or to merge as it is with the potential for future improvements?

oscarbenjamin:
This is the main outstanding issue for the 1.11 release. Is it likely to get resolve any time soon? If not then I will move the milestone to 1.12.

jksuom:
>  I will move the milestone to 1.12.

Maybe that should be done. This is not a critical fix.

PR #23541, opened 2022-05-25
Improve performance of inverse trigonometric functions

eendebakpt:

This PR improves performance of the inverse trigonometric functions by:

* Caching the lookup tables in `InverseTrigonometricFunction` (so the tables are not constructed for each invocation of the function)
* Consistently using the variable `pi` instead of `S.Pi` (which saves an attribute lookup)

The main speed improvement is for numbers. Two benchmarks:

```
%timeit atan(np.random.rand()) 
#main: 670 µs ± 19.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
#PR: 488 µs ± 1.47 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

```
%timeit asin(np.random.rand()) # 670 us on main, 512 us with class init of _atsin_table
#main: 865 µs ± 10.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
#PR: 513 µs ± 705 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
#### Other comments

The same improvements can be applied to other functions in the sympy codebase. The changes in this PR are limited to a single file to keep the PR small and first discuss whether the approach is good.

For catching the sympy `cacheit` method is used. Since the lookup table methods have no arguments, this is a bit overkill. We could replace it by `functools.lru_cache(size=1)`

Another approach would be to replace the `_atan_table` with a class attribute. This failed because of circular references.

#### Release Notes

* functions
  * Improve performance of trigonometric functions

### DIFF
--- a/sympy/functions/elementary/trigonometric.py
+++ b/sympy/functions/elementary/trigonometric.py
@@ -2,7 +2,7 @@ from typing import Tuple as tTuple
 
 from sympy.core.add import Add
 from sympy.core.basic import sympify, cacheit
-from sympy.core.expr import Expr
+from sympy.core.expr import Expr, _imaginary_unit_as_coefficient
 from sympy.core.function import Function, ArgumentIndexError, PoleError, expand_mul
 from sympy.core.logic import fuzzy_not, fuzzy_or, FuzzyBool, fuzzy_and
 from sympy.core.numbers import igcdex, Rational, pi, Integer
@@ -20,12 +20,6 @@ from sympy.ntheory import factorint
 from sympy.polys.specialpolys import symmetric_poly
 from sympy.utilities.iterables import numbered_symbols
 
-def _imaginary_unit_as_coefficient(arg):
-    """ Extracts symbolic coefficient for imaginary unit """
-    if getattr(arg, 'is_real', True):
-        return None
-    else:
-        return arg.as_coefficient(S.ImaginaryUnit)
 
 ###############################################################################
 ########################## TRIGONOMETRIC FUNCTIONS ############################
--- a/sympy/functions/elementary/trigonometric.py
+++ b/sympy/functions/elementary/trigonometric.py
@@ -19,7 +19,6 @@ from sympy.functions.elementary.piecewise import Piecewise
 from sympy.ntheory import factorint
 from sympy.polys.specialpolys import symmetric_poly
 from sympy.utilities.iterables import numbered_symbols
-from sympy.core.cache import cacheit
 
 ###############################################################################
 ########################## TRIGONOMETRIC FUNCTIONS ############################
--- a/sympy/functions/elementary/trigonometric.py
+++ b/sympy/functions/elementary/trigonometric.py
@@ -20,6 +20,13 @@ from sympy.ntheory import factorint
 from sympy.polys.specialpolys import symmetric_poly
 from sympy.utilities.iterables import numbered_symbols
 
+def _imaginary_unit_as_coefficient(arg):
+    """ Extracts symbolic coefficient for imaginary unit """
+    if getattr(arg, 'is_real', True):
+        return None
+    else:
+        return arg.as_coefficient(S.ImaginaryUnit)
+
 ###############################################################################
 ########################## TRIGONOMETRIC FUNCTIONS ############################
 ###############################################################################
@@ -305,7 +312,7 @@ class sin(TrigonometricFunction):
         if arg.could_extract_minus_sign():
             return -cls(-arg)
 
-        i_coeff = arg.as_coefficient(S.ImaginaryUnit)
+        i_coeff = _imaginary_unit_as_coefficient(arg)
         if i_coeff is not None:
             from sympy.functions.elementary.hyperbolic import sinh
             return S.ImaginaryUnit*sinh(i_coeff)
@@ -598,7 +605,7 @@ class cos(TrigonometricFunction):
         if arg.could_extract_minus_sign():
             return cls(-arg)
 
-        i_coeff = arg.as_coefficient(S.ImaginaryUnit)
+        i_coeff = _imaginary_unit_as_coefficient(arg)
         if i_coeff is not None:
             from sympy.functions.elementary.hyperbolic import cosh
             return cosh(i_coeff)
@@ -1059,7 +1066,7 @@ class tan(TrigonometricFunction):
         if arg.could_extract_minus_sign():
             return -cls(-arg)
 
-        i_coeff = arg.as_coefficient(S.ImaginaryUnit)
+        i_coeff = _imaginary_unit_as_coefficient(arg)
         if i_coeff is not None:
             from sympy.functions.elementary.hyperbolic import tanh
             return S.ImaginaryUnit*tanh(i_coeff)
@@ -1392,7 +1399,7 @@ class cot(TrigonometricFunction):
         if arg.could_extract_minus_sign():
             return -cls(-arg)
 
-        i_coeff = arg.as_coefficient(S.ImaginaryUnit)
+        i_coeff = _imaginary_unit_as_coefficient(arg)
         if i_coeff is not None:
             from sympy.functions.elementary.hyperbolic import coth
             return -S.ImaginaryUnit*coth(i_coeff)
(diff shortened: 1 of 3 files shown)


oscargus:
> is for numbers

I guess that there would be an even bigger improvement if the argument was checked for being a `Float` first and in that case skip all the remaining tests? This probably holds for most evaluations as they focus on symbolic values.

eendebakpt:
@oscargus I tested skipping the tests in `atan.eval` with a check `isinstance(arg, Float)`. Most tests don't have a significant impact on the running time. The only exception is `i_coeff = arg.as_coefficient(S.ImaginaryUnit)`, which is quite expensive.

```
%timeit atan(np.random.rand()) 
#main: 670 µs ± 19.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
#PR: 488 µs ± 1.47 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
#PR + fix as_coefficient: 211 µs ± 882 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
I consider two options to fix the `rg.as_coefficient(S.ImaginaryUnit)`:

1) Add a check with `isinstance(arg, Float)` in the `eval` method. Works, but has to be done at multiple locations (a quick grep on `as_coefficient(S.ImaginaryUnit)` gives 36 hits
2) Improve `arg.as_coefficient`. Something like
```
    def as_coefficient(self, expr):
        if isinstance(self, Float) and expr == S.ImaginaryUnit:
            return None
        
        r = self.extract_multiplicatively(expr)
        if r and not r.has(expr):
            return r
```
works, but it not really generic.

I would also be fine with creating a new issue for this to keep this PR focused.

eendebakpt:
@oscargus If all expressions of interest have `.is_real`, we could replace all occurrences of `as_coefficient(S.ImaginaryUnit)` with a specialized method:
```
def as_coefficient_imaginary_unit(arg):
   if arg.is_real:
      return None
   else:
      return arg.as_coefficient(S.ImaginaryUnit)
```

sympy/functions/elementary/trigonometric.py
```diff
@@ -20,6 +20,13 @@
 from sympy.polys.specialpolys import symmetric_poly
 from sympy.utilities.iterables import numbered_symbols
 
+def _imaginary_unit_as_coefficient(arg):
```
oscarbenjamin, line 23: This function seems out of place at this line in the file.

sympy/functions/elementary/trigonometric.py
```diff
@@ -20,6 +20,13 @@
...
+    """ Extracts symbolic coefficient for imaginary unit """
+    if getattr(arg, 'is_real', True):
+        return None
```
oscarbenjamin, line 26: Why would the arg not have an `is_real` attribute?

sympy/functions/elementary/trigonometric.py
eendebakpt: @oscarbenjamin I moved it to `sympy.core.expr`.

oscarbenjamin:
Using `is_real` is a semantic check whereas I think what is really needed here is a structural one (also for performance reasons because `is_real` can be slow).

Consider something like:
```python
In [10]: x = symbols('x', imaginary=True)

In [11]: e = x*I

In [12]: e
Out[12]: ⅈ⋅x

In [13]: e.is_real
Out[13]: True

In [14]: sin(e)
Out[14]: ⅈ⋅sinh(x)
```
This PR changes that behaviour:
```python
In [3]: sin(e)
Out[3]: sin(ⅈ⋅x)
```

eendebakpt:
> Using `is_real` is a semantic check whereas I think what is really needed here is a structural one (also for performance reasons because `is_real` can be slow).
> 
> Consider something like:
> 
> ```python
> In [10]: x = symbols('x', imaginary=True)
> 
> In [11]: e = x*I
> 
> In [12]: e
> Out[12]: ⅈ⋅x
> 
> In [13]: e.is_real
> Out[13]: True
> 
> In [14]: sin(e)
> Out[14]: ⅈ⋅sinh(x)
> ```
> 
> This PR changes that behaviour:
> 
> ```python
> In [3]: sin(e)
> Out[3]: sin(ⅈ⋅x)
> ```

@oscarbenjamin I updated the method to use `isinstance(arg, Float)`. In a separate PR I created a regression test.

oscarbenjamin:
> That is interesting, but probably for another PR.

I disagree. If it's better to do it there then there's no need to go around making changes to each of the subclasses. Also I suspect that skipping eval altogether would give a bigger speedup than making minor changes in eval.

eendebakpt:
> > That is interesting, but probably for another PR.
> 
> I disagree. If it's better to do it there then there's no need to go around making changes to each of the subclasses. Also I suspect that skipping eval altogether would give a bigger speedup than making minor changes in eval.

@oscarbenjamin This PR contains 3 optimizations:

* Cache the trigoniometric tables
* Replace `S.Pi` with `pi`
* Fast path for `Float` arguments via `_imaginary_unit_as_coefficient`

The first two seem worthwhile because they are also effective for non `Float` arguments. The last one is indeed irrelevant if we can implement your suggestion to modify the parent class.

I created PR #23565 to test the changes to the parent class. I will put this PR on hold until that PR is either merged or rejected.

asmeurer:
> For catching the sympy cacheit method is used. Since the lookup table methods have no arguments, this is a bit overkill. We could replace it by functools.lru_cache(size=1)

If cacheit is slow for this case (which we should check if it actually is), then we should optimize this at the cacheit level, so that it just saves the result for functions without arguments rather than using a lru_cache. However, I suspect this optimization may already exist in lru_cache, so it's possible this doesn't even matter.

asmeurer:
> For catching the sympy cacheit method is used. Since the lookup table methods have no arguments, this is a bit overkill. We could replace it by functools.lru_cache(size=1)

If cacheit is slow for this case (which we should check if it actually is), then we should optimize this at the cacheit level, so that it just saves the result for functions without arguments rather than using a lru_cache. However, I suspect this optimization may already exist in lru_cache, so it's possible this doesn't even matter.

eendebakpt:
> > For catching the sympy cacheit method is used. Since the lookup table methods have no arguments, this is a bit overkill. We could replace it by functools.lru_cache(size=1)
> 
> If cacheit is slow for this case (which we should check if it actually is), then we should optimize this at the cacheit level, so that it just saves the result for functions without arguments rather than using a lru_cache. However, I suspect this optimization may already exist in lru_cache, so it's possible this doesn't even matter.

The overhead of cacheit is not a lot, so I think this is fine. (the overhead is in checking for unhashable types, which would be hard to change). A special check for functions without arguments is possible, but not worth the effort here.

eendebakpt:
@oscarbenjamin As #23582 is not a good alternative for the `_imaginary_unit_as_coefficient`, can you review again?

oscarbenjamin:
Timing here I don't see much of a difference between the PR and master.

This is master:
```python
In [1]: %timeit atan(np.random.rand())
954 µs ± 4.98 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [2]: %timeit asin(np.random.rand())
1.43 ms ± 15.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
This is the PR:
```python
In [1]: %timeit atan(np.random.rand())
920 µs ± 4.35 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [2]: %timeit asin(np.random.rand())
1.59 ms ± 120 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
What happened to the speedups reported in the OP?

oscarbenjamin:
I'm not sure what happened there. I just tried (possibly on a different computer because I don't remember which I used before) and I can see the difference.

This is master:
```python
In [1]: %timeit atan(np.random.rand())
911 µs ± 58.7 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [2]: %timeit asin(np.random.rand())
1.39 ms ± 4.96 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
This is the PR:
```python
In [1]: %timeit atan(np.random.rand())
326 µs ± 7.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [2]: %timeit asin(np.random.rand())
523 µs ± 1.19 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

oscarbenjamin:
Okay, I think this looks good.

oscarbenjamin:
Thanks!

PR #23546, opened 2022-05-27
simplify function is added in util.py

### LINKED ISSUE #23375: Units get lost when using convert_to()
In some circumstances, `convert_to()` seems to result in a loss of units.

Example code:
```python
import sympy
from sympy import *
from sympy.physics.units import *

expr1 = Mul(Float('2.9357798165137616e-13', precision=53), Pow(centimeter, Integer(-2)), Pow(meter, Integer(-2)), kilogram, Add(Mul(Float('415734822253.0', precision=53), Pow(centimeter, Integer(2))), Mul(Float('29293750.0', precision=53), Pow(meter, Integer(2)))))
print(expr1.simplify())
print(convert_to(expr1, kg/m**2).simplify())
```

Output:
```
2.08050590019229e-5*kilogram/centimeter**2
0.208050590019229*kilogram
```

Observed on conda python 3.9 with the following package:
```
> conda list sympy
# Name                    Version                   Build  Channel
sympy                     1.10.1           py39haa95532_0
```

OMH-G:
convert_to function in util.py is modified used simplify function from sympy to tackle the unit not showing problem

#### Code which does not show the unit before modifying util.py
```
import sympy
from sympy import *
from sympy.physics.units import *

expr1 = Mul(Float('2.9357798165137616e-13', precision=53), Pow(centimeter, Integer(-2)), Pow(meter, Integer(-2)), kilogram, Add(Mul(Float('415734822253.0', precision=53), Pow(centimeter, Integer(2))), Mul(Float('29293750.0', precision=53), Pow(meter, Integer(2)))))
print(expr1.simplify())
print(convert_to(expr1, kg/m**2).simplify())
```
#### Output
```
2.08050590019229e-5*kilogram/centimeter**2
0.208050590019229*kilogram
```

#### After modifying the util.py convert_to function 

Screenshot (186)

* core
  * simplify function is passed to convert expression to simplified expression 
* printing
  * The output is displayed with the units

### DIFF
--- a/sympy/physics/units/util.py
+++ b/sympy/physics/units/util.py
@@ -11,7 +11,6 @@ from sympy.core.containers import Tuple
 from sympy.core.mul import Mul
 from sympy.core.power import Pow
 from sympy.core.sorting import ordered
-from sympy.simplify.simplify import simplify
 from sympy.core.sympify import sympify
 from sympy.matrices.common import NonInvertibleMatrixError
 from sympy.physics.units.dimensions import Dimension, DimensionSystem
--- a/sympy/physics/units/util.py
+++ b/sympy/physics/units/util.py
@@ -11,6 +11,7 @@ from sympy.core.containers import Tuple
 from sympy.core.mul import Mul
 from sympy.core.power import Pow
 from sympy.core.sorting import ordered
+from sympy.simplify.simplify import simplify
 from sympy.core.sympify import sympify
 from sympy.matrices.common import NonInvertibleMatrixError
 from sympy.physics.units.dimensions import Dimension, DimensionSystem
@@ -49,6 +50,7 @@ def _get_conversion_matrix_for_expr(expr, target_units, unit_system):
 
 
 def convert_to(expr, target_units, unit_system="SI"):
+    expr=expr.simplify()
     """
     Convert ``expr`` to the same expression with all of its units and quantities
     represented as factors of ``target_units``, whenever the dimension is compatible.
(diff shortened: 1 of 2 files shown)


sympy/physics/units/util.py
```diff
@@ -49,6 +49,7 @@ def _get_conversion_matrix_for_expr(expr, target_units, unit_system):
 
 
 def convert_to(expr, target_units, unit_system="SI"):
+    expr=expr.simplify()
     """
     Convert ``expr`` to the same expression with all of its units and quantities
```
oscarbenjamin, line 54: We generally don't want to use `simplify` in library code. Also there shouldn't be code before a docstring.

OMH-G: Thankyou for acknowledging , but checking if the `dir(expr)` has simplify in it then the expr is having simplified version of itself . so `simplify `function was helping to get out of the unit that it wanted and in future PR it adding code before docstring will not be happen .

OMH-G: @oscarbenjamin I have tried to check whether simplify is present in expr if it is present then it simplifies the expression when I run through sample of test cases as well as the test case where unit was not shown up , it started working well
Screenshot (187)
so is there any alternate way to add simplify?, as when adding simplify function to util makes the test case to pass and unit was also showing up 

This patch was added to get accepted behavior in util.py 
```
if 'simplify' in dir(expr):
        expr=expr.simplify()
```

oscarbenjamin: We don't want to use `simplify` here.

Issue #22863, opened 2022-01-14
Hangs: integrate((3*x**3-x**2+2*x-4)/sqrt(x**2-3*x+2), (x, 0, 1))


certik:
The `Integral(...).n()` works, but the same `integrate(...)` hangs:
```
In [1]: Integral((3*x**3-x**2+2*x-4)/sqrt(x**2-3*x+2), (x, 0, 1))
Out[1]: 
1                       
⌠                       
⎮    3    2             
⎮ 3⋅x  - x  + 2⋅x - 4   
⎮ ─────────────────── dx
⎮     ______________    
⎮    ╱  2               
⎮  ╲╱  x  - 3⋅x + 2     
⌡                       
0                       

In [2]: i = Integral((3*x**3-x**2+2*x-4)/sqrt(x**2-3*x+2), (x, 0, 1))

In [3]: i.n()
Out[3]: -2.98126694400554

In [4]: integrate((3*x**3-x**2+2*x-4)/sqrt(x**2-3*x+2), (x, 0, 1))
[HANGS]
```

oscarbenjamin:
It hangs in manulintegrate somewhere.

geekboi777:
Can I be assigned this issue ?

geekboi777:
Well, I believe that this may be due to the fact that SymPy initially takes up a lot of time finding for elementary anti-derivatives using the reisch algorithm or other available algorithms before it tries for the most suitable ones for definite integrals.

Because, using the Meijer G methods actually prevented the hanging and gave the correct output.

<img width="766" alt="Screenshot 2022-01-19 at 2 54 30 AM" src="https://user-images.githubusercontent.com/85869106/150022428-abf25b98-5e5c-4e65-af54-60d43e3c3106.png">

Well its one speculation from my side. There may be other reasons for the extra time consumption.

asmeurer:
It's a known issue that heurisch can hang, but manualintegrate shouldn't hang. It should be investigated. manualintegrate is done before meijerg because it's generally faster. It's likely it is stuck in an infinite loop of sort.

asmeurer:
Actually for me in master, the integral returns unevaluated in about 20 seconds. @certik were you using master or a release version of SymPy? 

The fact that it can't do the integral isn't too surprising. SymPy isn't very good at algebraic integrals.

certik:
[...]

Issue #17383, opened 2019-08-11
Sympy.integrate returns incorrect result. (Problem with DiracDelta functions.)


Volker-Weissmann:
Long story short:
`integrate(exp(x)* DiracDelta(x,1) , (x,-1,z))`
returns
`-Heaviside(z)`
. This is false! The correct answer would be
`DiracDelta(z, 0) - Heaviside(z)`
as can be proven using partial integration.
Interestingly, Sympy is able to correctly tell that
`integrate((1+x+b*x**2)* DiracDelta(x,1) , (x,-1,z))`
is
`DiracDelta(z, 0) - Heaviside(z)`
and that
`integrate((sinh(x)+cosh(x))* DiracDelta(x,1) , (x,-1,z))`
is
`DiracDelta(z, 0) - Heaviside(z)`
. Note that sinh(x)+cosh(x) = exp(x)

jksuom:
> integrate(exp(x)* DiracDelta(x,1) , (x,-1,z))

Integrals involving generalized functions are well defined only when there are no singularities at the endpoints of integration interval. So, strictly speaking, this is defined only for `z != 0`, and its value is `-Heaviside(z)`. However, if we want to interpret the integral expression as an antiderivative, then it should be `DiracDelta(z) - Heaviside(z)` (which is a generalized function with no proper value for `z == 0`).

It seems that `deltaintegrate` is designed to compute only the definite integral. The other terms of an antiderivative (as a distribution) are missing in this code
https://github.com/sympy/sympy/blob/6c03115c5c055f135100d4866b90f5f65167ddca/sympy/integrals/deltafunctions.py#L185-L194
There should be several terms (containing delta and its derivatives) instead of only one (that will give the definite integral).

jksuom:
Here are some thoughts about antiderivatives of `f(x)*DiracDelta(x, n)` where `f(x) = rest_mult` in the code (and `point = 0` to simplify notation). This is a [distribution with point support](https://en.wikipedia.org/wiki/Distribution_(mathematics)#Distributions_with_point_support), hence it can be represented as a finite linear combination with *constant coefficients* of Dirac delta and its derivatives. Given such a representation, it is easy to find an antiderivative (`a*DiracDelta(x, n-1)` for `a*DiracDelta(x, n)` if `n > 0`, and `a*Heaviside(x)` for `a*DiracDelta(x, 0)`).

[...]

PR #23559, opened 2022-05-30
manualintegrate x**n*(trig or Derivative)

### LINKED ISSUE #16161: Integral of x*sec(x)**2 doesn't evaluate
This comes from #11131.

This integral doesn't evaluate but Wolfram gives an answer that differentiates back to a form equivalent to the integrand:
```julia
In [14]: f = x*sec(x)**2                                                                                                                                      

In [15]: F = x*tan(x) + log(cos(x))                                                                                                                           

In [16]: f                                                                                                                                                    
Out[16]: 
     2   
x⋅sec (x)

In [17]: F                                                                                                                                                    
Out[17]: x⋅tan(x) + log(cos(x))

In [18]: integrate(f, x)                                                                                                                                      
Out[18]: 
⌠             
⎮      2      
⎮ x⋅sec (x) dx
⌡             

In [19]: F.diff(x)                                                                                                                                            
Out[19]: 
  ⎛   2       ⎞   sin(x)         
x⋅⎝tan (x) + 1⎠ - ────── + tan(x)
                  cos(x)         

In [20]: simplify(F.diff(x)-f)                                                                                                                                
Out[20]: 0
```

### LINKED ISSUE #23495: Integral fails to integrate
Attempting to integrate:
`integrate(z*sec(z)*tan(z))`

Expected behavior is that it either successfully computes and shows the antiderivative or it returns an error or warning. 

This is really those two issues. By hand, the result is x*sec(x)-ln(|sec(x)+tan(x)|)

eagleoflqj:

Fixes #23495 
Fixes #16161
$\int x^n f^{(n+1)}(x)\mathrm{d}x=\int x^n\mathrm{d}f^{(n)}(x)$
$=x^n f^{(n)}(x)-n\int x^{n-1}f^{(n)}(x)\mathrm{d}x$
$=x^n f^{(n)}(x)-nx^{n-1}f^{(n-1)}(x)+n(n-1)x^{n-2}f^{(n-2)}(x)-\cdots+(-1)^{n}n!f(x)$
$=\Sigma_{i=0}^n (-1)^{i}\frac{n!}{(n-i)!}x^{n-i} f^{(n-i)}(x)$

Edit:
This rule is too aggressive that may slow down many cases.
Enlarge the range of `dv` when `u` is algebraic for parts rule instead.
#### Release Notes

* integrals
  * Support integrate(x**n*(trig or Derivative))

### DIFF
--- a/sympy/integrals/manualintegrate.py
+++ b/sympy/integrals/manualintegrate.py
@@ -580,7 +580,7 @@ def _parts_rule(integrand, symbol):
     def pull_out_algebraic(integrand):
         integrand = integrand.cancel().together()
         # iterating over Piecewise args would not work here
-        algebraic = ([] if isinstance(integrand, Piecewise)
+        algebraic = ([] if isinstance(integrand, Piecewise) or not integrand.is_Mul
             else [arg for arg in integrand.args if arg.is_algebraic_expr(symbol)])
         if algebraic:
             u = Mul(*algebraic)
@@ -633,7 +633,9 @@ def _parts_rule(integrand, symbol):
                         return
 
             # Can integrate a polynomial times OrthogonalPolynomial
-            if rule == pull_out_algebraic and isinstance(dv, OrthogonalPolynomial):
+            if rule == pull_out_algebraic:
+                if dv.is_Derivative or dv.has(TrigonometricFunction) or \
+                        isinstance(dv, OrthogonalPolynomial):
                     v_step = integral_steps(dv, symbol)
                     if contains_dont_know(v_step):
                         return
@@ -723,12 +725,7 @@ def parts_rule(integral):
             return PartsRule(u, dv, v_step,
                              make_second_step(steps[1:], v * du),
                              integrand, symbol)
-        else:
-            steps = integral_steps(integrand, symbol)
-            if steps:
-                return steps
-            else:
-                return DontKnowRule(integrand, symbol)
+        return integral_steps(integrand, symbol)
 
     if steps:
         u, dv, v, du, v_step = steps[0]
--- a/sympy/integrals/tests/test_failing_integrals.py
+++ b/sympy/integrals/tests/test_failing_integrals.py
@@ -8,7 +8,7 @@ from sympy.functions.elementary.exponential import (exp, log)
 from sympy.functions.elementary.hyperbolic import (sech, sinh)
 from sympy.functions.elementary.miscellaneous import sqrt
 from sympy.functions.elementary.piecewise import Piecewise
-from sympy.functions.elementary.trigonometric import (acos, atan, cos, sec, sin, tan)
+from sympy.functions.elementary.trigonometric import (acos, atan, cos, sin, tan)
 from sympy.functions.special.delta_functions import DiracDelta
 from sympy.functions.special.gamma_functions import gamma
 from sympy.integrals.integrals import (Integral, integrate)
@@ -118,13 +118,6 @@ def test_issue_16396a():
 def test_issue_16396b():
     i = integrate(x*sin(x)/(1+cos(x)**2), (x, 0, pi))
     assert not i.has(Integral)
-
-
-@XFAIL
-def test_issue_16161():
-    i = integrate(x*sec(x)**2, x)
-    assert not i.has(Integral)
-    # assert i == x*tan(x) + log(cos(x))
 
 
 @XFAIL
@@ -266,3 +259,15 @@ def test_issue_4311_slow():
 def test_issue_20370():
     a = symbols('a', positive=True)
     assert integrate((1 + a * cos(x))**-1, (x, 0, 2 * pi)) == (2 * pi / sqrt(1 - a**2))
+
+
+@XFAIL
+def test_polylog():
+    # log(1/x)*log(x+1)-polylog(2, -x)
+    assert not integrate(log(1/x)/(x + 1), x).has(Integral)
+
+
+@XFAIL
+def test_polylog_manual():
+    # Make sure _parts_rule does not go into an infinite loop here
+    assert not integrate(log(1/x)/(x + 1), x, manual=True).has(Integral)
--- a/sympy/integrals/tests/test_integrals.py
+++ b/sympy/integrals/tests/test_integrals.py
@@ -15,7 +15,7 @@ from sympy.functions.elementary.exponential import (LambertW, exp, exp_polar, lo
 from sympy.functions.elementary.hyperbolic import (acosh, asinh, cosh, coth, csch, sinh, tanh, sech)
 from sympy.functions.elementary.miscellaneous import (Max, Min, sqrt)
 from sympy.functions.elementary.piecewise import Piecewise
-from sympy.functions.elementary.trigonometric import (acos, asin, atan, cos, sin, sinc, tan)
+from sympy.functions.elementary.trigonometric import (acos, asin, atan, cos, sin, sinc, tan, sec)
 from sympy.functions.special.delta_functions import DiracDelta
 from sympy.functions.special.error_functions import (Ci, Ei, Si, erf, erfc, erfi, fresnelc, li)
 from sympy.functions.special.gamma_functions import (gamma, polygamma)
@@ -2005,3 +2005,10 @@ def test_sqrt_quadratic():
 
     assert integrate(x*sqrt(x**2+2*x+4)) == \
         (x**2/3 + x/6 + S(5)/6)*sqrt(x**2 + 2*x + 4) - 3*asinh(sqrt(3)*(x + 1)/3)/2
+
+
+def test_mul_pow_derivative():
+    assert integrate(x*sec(x)*tan(x)) == x*sec(x) - log(tan(x) + sec(x))
+    assert integrate(x*sec(x)**2, x) == x*tan(x) + log(cos(x))
+    assert integrate(x**3*Derivative(f(x), (x, 4))) == \
+           x**3*Derivative(f(x), (x, 3)) - 3*x**2*Derivative(f(x), (x, 2)) + 6*x*Derivative(f(x), x) - 6*f(x)
--- a/sympy/integrals/tests/test_manual.py
+++ b/sympy/integrals/tests/test_manual.py
@@ -74,8 +74,6 @@ def test_manualintegrate_parts():
     assert manualintegrate((3*x**2 + 5) * exp(x), x) == \
         3*x**2*exp(x) - 6*x*exp(x) + 11*exp(x)
     assert manualintegrate(atan(x), x) == x*atan(x) - log(x**2 + 1)/2
-    # Make sure _parts_rule does not go into an infinite loop here
-    assert manualintegrate(log(1/x)/(x + 1), x).has(Integral)
 
     # Make sure _parts_rule doesn't pick u = constant but can pick dv =
     # constant if necessary, e.g. for integrate(atan(x))
@@ -658,3 +656,11 @@ def test_manualintegrate_sqrt_quadratic():
 
     assert_is_integral_of(x*sqrt(x**2+2*x+4),
                           (x**2/3 + x/6 + S(5)/6)*sqrt(x**2 + 2*x + 4) - 3*asinh(sqrt(3)*(x + 1)/3)/2)
+
+
+def test_mul_pow_derivative():
+    assert_is_integral_of(x*sec(x)*tan(x), x*sec(x) - log(tan(x) + sec(x)))
+    assert_is_integral_of(x*sec(x)**2, x*tan(x) + log(cos(x)))
+    assert_is_integral_of(x**3*Derivative(f(x), (x, 4)),
+                          x**3*Derivative(f(x), (x, 3)) - 3*x**2*Derivative(f(x), (x, 2)) +
+                          6*x*Derivative(f(x), x) - 6*f(x))


sympy/integrals/manualintegrate.py
```diff
@@ -1601,6 +1628,18 @@ def eval_derivativerule(integrand, symbol):
...
+
+@evaluates(MulPowDerivativeRule)
+def eval_mul_pow_derivative(n: Integer, steps: list, results: list[Expr], integrand: Expr, x: Symbol):
+    result = 0
+    coeff = 1
+    for i in range(n+1):
+        e = n - i
+        result += coeff * x**e * results[i]
+        coeff *= -e
+    return result
```
oscarbenjamin, line 1640: Does this guarantee to return a sympified result?

It's generally inefficient to build an Add term by term with `+=` because each time a new `Add` is created there is `O(n)` work done to canonicalise. It is better then to do `return Add(*terms)`.

eagleoflqj: > Does this guarantee to return a sympified result?

I think so, as `x` is `Symbol` and `n` is positive. I can make `result = S.Zero` for safety.

> It's generally inefficient to build an Add term by term with `+=` because each time a new `Add` is created there is `O(n)` work done to canonicalise. It is better then to do `return Add(*terms)`.

I'll fix that.

sympy/integrals/manualintegrate.py
```diff
@@ -898,6 +899,31 @@ def root_mul_rule(integral):
         return URule(u, u_func, None, next_step, integrand, symbol)
 
 
+def mul_pow_derivative_rule(integral: tuple[Expr, Symbol]):
+    """
+    Apply "Integral(x**n*Derivative(f(x), (x, n + 1)), x) equals to x**n*Derivative(f(x), (x, n))
+    - n*Integral(x**(n - 1)*Derivative(f(x), (x, n)), x)" n times
```
oscarbenjamin, line 905: Is it possible to use Cauchy's formula rather than repeated integration here?

https://en.wikipedia.org/wiki/Cauchy_formula_for_repeated_integration

eagleoflqj: Here we need all intermediate $f^{(n-i)}(x)$, not only for constructing final result $\Sigma_{i=0}^n (-1)^{i}\frac{n!}{(n-i)!}x^{n-i} f^{(n-i)}(x)$, but also for showing manual steps, so the formula doesn't seem applicable here.

oscarbenjamin: I wondered if it might be more efficient to compute the nth integral once and then differentiate it to get the intermediate results. Maybe the integral implied by the formula would be harder than the original integral though...

oscarbenjamin: How well does this perform if `n` is large?

eagleoflqj: > Maybe the integral implied by the formula would be harder than the original integral though...

That's the case. This PR assumes `x^n * f(x)` is hard to integrate, and if `f(x)` can be integrated `n+1` times, then `x^n * f(x)` can be computed. The Cauchy formula assumes a definite integration `(x-t)^n * f(t)` is easy to compute, so it can be used to integrate `f(x)` `n+1` times. That's just the inverse problem.

> How well does this perform if `n` is large?

If `n` is large, $f^{(n+1)}(x)$ tends to be complicated, so it's likely not be integrated by existing algorithms.
```py
>>> from sympy import *
>>> x=Symbol('x')
>>> [integrate(tan(x).diff(x, n), x, manual=True) for n in range(4)]
[-log(cos(x)), tan(x), tan(x)**2, -6*x + 8*tan(x) + 6*Integral(tan(x)**4, x)]
```

oscarbenjamin: I guess it's not worth using the formula.

As an aside I just tried this and it's really slow (on master, due to heurisch):
```python
In [6]: %time integrate(x**20*sin(x))
CPU times: user 1min 16s, sys: 116 ms, total: 1min 16s
Wall time: 1min 16s
Out[6]: 
   20              19               18                17                  16                   15  
- x  ⋅cos(x) + 20⋅x  ⋅sin(x) + 380⋅x  ⋅cos(x) - 6840⋅x  ⋅sin(x) - 116280⋅x  ⋅cos(x) + 1860480⋅x  ⋅s

                  14                     13                      12                       11       
in(x) + 27907200⋅x  ⋅cos(x) - 390700800⋅x  ⋅sin(x) - 5079110400⋅x  ⋅cos(x) + 60949324800⋅x  ⋅sin(x)

                 10                         9                          8                           
 + 670442572800⋅x  ⋅cos(x) - 6704425728000⋅x ⋅sin(x) - 60339831552000⋅x ⋅cos(x) + 482718652416000⋅x

7                            6                             5                              4        
 ⋅sin(x) + 3379030566912000⋅x ⋅cos(x) - 20274183401472000⋅x ⋅sin(x) - 101370917007360000⋅x ⋅cos(x) 

                      3                               2                                            
+ 405483668029440000⋅x ⋅sin(x) + 1216451004088320000⋅x ⋅cos(x) - 2432902008176640000⋅x⋅sin(x) - 243

                       
2902008176640000⋅cos(x)

```

eagleoflqj:
I put the new rule at the end of the list to try, to avoid giving a more complicated answer to
https://github.com/sympy/sympy/blob/0ebfb1c777e1ad874c7a865fb2848624f072e6a9/sympy/integrals/tests/test_manual.py#L239-L242
Now only
https://github.com/sympy/sympy/blob/0ebfb1c777e1ad874c7a865fb2848624f072e6a9/sympy/integrals/tests/test_integrals.py#L1680-L1683
fails. I investigated a bit and it seems the new rule tries to integrate `_u**2/sqrt(-_u**2 + _u*h)`. That should have been rsolved before trying this rule. I'll leave this PR draft for a while and open a new PR to address that.

eagleoflqj:
> Was this fixed somewhere else?

Yes, this pattern is covered by #23608

But this PR brings a 4x slow for `test_manualintegrate_parts`. I'd perf it before merging.

eagleoflqj:
The problem is at
https://github.com/sympy/sympy/blob/18afc6b4d724ed286381dcb6f0f1994578669e32/sympy/integrals/tests/test_manual.py#L76
It tries many new possibilities brought by `mul_pow_derivative_rule` before failing.
However the result from Wolfram Alpha is correct:
```py
>>> from sympy import *
>>> x=Symbol('x')
>>> F=log(1/x)*log(x+1)-polylog(2, -x)
>>> expand_func(F.diff(x))
log(1/x)/(x + 1)
```

eagleoflqj:
I don't have quick solution. Since it actually has closed form result, I'd move it to failing tests.

jksuom:
Does this method also handle derivatives of Dirac delta? (cf. https://github.com/sympy/sympy/issues/17383#issuecomment-526914713)

eagleoflqj:
> Does this method also handle derivatives of Dirac delta? (cf. [#17383 (comment)](https://github.com/sympy/sympy/issues/17383#issuecomment-526914713))

```py
>>> integrate(x*DiracDelta(x,1), manual=True)
Integral(x*DiracDelta(x, 1), x)
```
No, I think it's because manual integration hasn't handled `DiracDelta` yet.
```py
>>> integrate(DiracDelta(x,1), manual=True)
Integral(DiracDelta(x, 1), x)
```

eagleoflqj:
With #23670 and this PR we have
```py
>>> integrate(x*DiracDelta(x,1), manual=True)
x*DiracDelta(x) - Heaviside(x)
```
However manual is not preferred
```py
>>> integrate(x*DiracDelta(x,1))
-Heaviside(x)
```
So this PR can't fix that issue. Even for manual, only `x**n * DiracDelta` is supported. For arbitrary `f(x) * DiracDelta` it seems generic parts rule should be improved.

asmeurer:
Would it be possible (or even a good idea) to subsume deltaintegrate into manualintegrate?

eagleoflqj:
> Would it be possible (or even a good idea) to subsume deltaintegrate into manualintegrate?

I think so, as ideally we should provide steps for any kinds of integrand.

eagleoflqj:
@oscarbenjamin Is it good to go?

oscarbenjamin:
The difficulty I have with all of these additions to manual integrate is that I don't have a clear understanding of the control flow in manual integrate myself. In principle each of the many PRs that have recently been merged just adds some extra rule(s) that are correct for integration and increase manualintegrate's coverage of different kinds of integrals.

However I also know that manualintegrate can be very slow in some cases and that it can also be the cause of slowness for general integration since manualintegrate is tried as the last method there. The slowness when I have investigated it seems to come from manualintegrate trying large numbers of possibilities so each time a new rule is added there is the risk that it just increases the number of possibilities. We don't have good test or benchmark coverage of different integrals for manualintegrate so I worry that while many of these rules might help in simple cases they might also lead to big slowdowns in more complicated cases including the common case where the integral cannot ultimately be computed.

The problem seems to be that there isn't really any consideration in the manualintegrate code of how to limit the number of possibilities. In principle a pattern matching integrater like `manualintegrate` can be something that is very fast but you can see how slow it is just for something simple:
```python
In [1]: from sympy.integrals.manualintegrate import manualintegrate

In [2]: %time integrate(x, manual=True)
CPU times: user 428 ms, sys: 4 ms, total: 432 ms
Wall time: 432 ms
Out[2]: 
 2
x 
──
2 
```
How can that possibly be so slow? Surely it should just match the obvious power of x rule and return immediately?

The problem that you refer to above is this integral. On master it takes 8 seconds and the result is still just an unevaluated integral:
```python
In [1]: %time integrate(log(1/x)/(x + 1), x)
CPU times: user 8.46 s, sys: 32 ms, total: 8.49 s
Wall time: 8.48 s
Out[1]: 
⌠          
⎮    ⎛1⎞   
⎮ log⎜─⎟   
⎮    ⎝x⎠   
⎮ ────── dx
⎮ x + 1    
⌡  
```
With the PR we have the same except that it takes 42 seconds:
```python
In [1]: %time integrate(log(1/x)/(x + 1), x)
CPU times: user 42.2 s, sys: 40 ms, total: 42.2 s
Wall time: 42.2 s
Out[1]: 
⌠          
⎮    ⎛1⎞   
⎮ log⎜─⎟   
⎮    ⎝x⎠   
⎮ ────── dx
⎮ x + 1    
⌡   
```
On master most of the time is already spent in manualintegrate.

How can we improve the general structure of manualintegrate to be smarter about all of the possibilities that it generates?

We had a similar problem previous with `dsolve` where every time a new solver is added it slows everything down even if the existing solvers are already capable of computing a solution. Some refactoring has helped there so that simple cases are a lot faster now but it can still be improved a lot. With manualintegrate it's more difficult though because it looks like the rules can potentially combine to generate combinatorially many cases and I haven't seen anything that tries to limit the cost of computing all of those cases.

eagleoflqj:
> How can that possibly be so slow? Surely it should just match the obvious power of x rule and return immediately?

I think that's just because it's the first time that `manualintegrate` is called. You can see subsequent calls are faster, even for a more complicated expression:
```py
In [1]: %time integrate(x, manual=True)
CPU times: user 190 ms, sys: 0 ns, total: 190 ms
Wall time: 189 ms
Out[1]: 
 2
x 
──
2 

In [2]: %time integrate(x**3/sqrt(1+x+x**2), manual=True)
CPU times: user 65.9 ms, sys: 697 µs, total: 66.6 ms
Wall time: 66.4 ms
Out[2]: 
                                         ⎛2⋅√3⋅(x + 1/2)⎞
⎛ 2           ⎞    ____________   7⋅asinh⎜──────────────⎟
⎜x    5⋅x   1 ⎟   ╱  2                   ⎝      3       ⎠
⎜── - ─── - ──⎟⋅╲╱  x  + x + 1  + ───────────────────────
⎝3     12   24⎠                              16          

In [3]: %time integrate(x, manual=True)
CPU times: user 37.4 ms, sys: 0 ns, total: 37.4 ms
Wall time: 37.1 ms
Out[3]: 
 2
x 
──
2 
```

oscarbenjamin:
I think that 40 milliseconds is still very slow to integrate `x`.

eagleoflqj:
> I think that 40 milliseconds is still very slow to integrate `x`.

Yes. That's a different meta problem for manual integration that worth investigation.

eagleoflqj:
> each time a new rule is added there is the risk that it just increases the number of possibilities.

That's true. We have 3 categories:
1. integrands that can already be integrated
2. integrands that has closed-form integral, but we can't handle yet
3. integrands that doesn't have closed-form, and of course we can never handle it

Adding new rules aim to move integrands from category 2 to 1. What we should be careful is to not slow down category 1 much.
For integrands in category 3, there isn't much that we can do currently to prevent its slowdown, because
* this category is too huge
* manual integration mimics integrating manually, and we human won't tell a complicated integrand has closed-form or not very fast, and we always make wrong guess that an integrand doesn't have closed-form before we try enough approaches. Searching thoroughly is the purpose that we use program to integrate.

`log(1/x)/(x+1)` is a different story. It's in category 2. We always meet that situation, like #22863. In diofant the solution was to add a test that shows this integrand fail fast, which IMHO is not much constructive. The correct solution is to **support** it, so it can be computed fast.

In conclusion, I think performance is important, but making integrands that have closed-form integral computed, is more important than making integrands that don't have closed-form integral fail early. An extreme example is that a program which computes nothing is fast but useless.

oscarbenjamin:
> An extreme example is that a program which computes nothing is fast but useless.

This is not really the situation though because usually we can do other things if the integral cannot be computed. For example in the case of `dsolve` we can try a different solver altogether and typically the different solvers lead to different integrals. If `integrate` hangs then we can't try another integral or another approach.

Another way to look at this is that the vast majority of integrals cannot be computed so actually the common case is really case 3. If you want to use something like `integrate` as part of an application then you clearly need to have some fallback for integrals that can't be computed and you want integrate to just handle some things quickly but otherwise fail fast. The fallback might be something like computing the integral numerically.

Also I have the impression that manualintegrate can be done in a cleverer way. I think that right now many possibilities that are tried are redundant. When debugging slow integrals that go through manualintegrate I find it very hard to keep track of what is happening but what I can definitely see is that the design does not seem to make any effort to limit the potential explosion of possible cases that can be attempted. The approach used by `possible_subterms` is a good example of this where it really just tries to collect everything that could conceivably be tried without e.g. giving any consideration to whether there is an obvious best substitution to use.

Another problem here I think is that manualintegrate is designed primarily for `integral_steps` but we also need it as part of `integrate`. I think maybe that's where the tension is here: for `integral_steps` there is not much benefit in failing fast but for `integrate` there very often is.

eagleoflqj:
@oscarbenjamin I try to let parts_rule accept these kinds of input. No slow down for `integrate(log(1/x)/(x + 1)` now.

oscarbenjamin:
> No slow down for `integrate(log(1/x)/(x + 1))` now.

I'll take a closer look later but I think this can probably be merged.

To be clear my point though is not about how we can add rules to improve particular examples but how we can limit the worst case runtime in general. I know that heurisch can be slow but I also know what I can do to improve that. The heurisch algorithm is well defined but it uses subroutines that are not efficiently implemented (polynomial gcd, linear algebra etc). The same goes for risch, meijerg etc.

In the case of manualintegrate there doesn't really seem to be any particular known algorithm that is being implemented and it can be extremely slow. There needs to be some work to improve this somehow but I don't personally have much understanding of how it works so I'm not sure what should be done. It seems like it tries some very slow things but I don't know how to judge what the value is of the different things it tries.

The example from #23704 is a good one to consider:
```python
In [1]: c, x, a, b = symbols('c x a b')

In [2]: %time integrate(log(x)/x**2/(c*x**2+b*x+a),x, manual=True)
# I don't know how long this takes...
```
Why is it so slow? Is that a bug? When I try to debug manualintegrate I just get lost...

eagleoflqj:
> how we can limit the worst case runtime in general.

Generally recursive calls to `integral_steps` without knowing new integrand is simpler should be avoided. Substitution rule and parts rule are 2 dangerous rules. The complexity can be defined naively by operation numbers, or more smart (but controversial), like `log` is simpler than `exp`, so don't substitute `u=log(x)` for `log(x)/x**2/(c*x**2+b*x+a)` to get a complicated expression that contains `exp`.

> Why is it so slow? Is that a bug? When I try to debug manualintegrate I just get lost...

I use open source PyCharm CE's debugger, pause from time to time, and the stack frame should contain several `integral_steps` calls. The `integrand` variable tells what it's trying to integrate, and other frames between `integral_steps` tells what specific rule it's trying. Every integration PRs of mine should give credits to the debugger.

Screenshot from 2022-06-29 23-19-26

asmeurer:
Shouldn't manualintegrate in principle be very easy to debug? Do we have a function that prints the integration steps?

eagleoflqj:
> Shouldn't manualintegrate in principle be very easy to debug?

Yes for wrong result, or hang due to a single slow step. No for hang due to too many possible branches.

> Do we have a function that prints the integration steps?

Only `print(integral_steps(integrand, x))` I think. Human readable steps were designed to be in another repo.

oscarbenjamin:
I think this looks good.

PR #23565, opened 2022-05-31
[WIP] Optimize evaluation for float arguments

eendebakpt:

This PR explores how we can improve the evaluation of floating point types (e.g. `float`, `sympy.Float`, complex numbers). In #23541 it was found that the performance of evaluations of floats is unneccary slow because in the `eval` method of a `Function` much work is peformed for general arguments that can be avoided with a fast path for floats.

The general workflow for an evaluation of a `Function` like `atan` on a float, e.g. `atan(2.3)` is:

1) Sympify the argument 2.3. Result: `Float(2.3)` 
2) In the `__new__` of the `Function` the `eval` method of `atan` is called.
3) The `eval` handles special cases (like `sympy.core.singleton.Infinity`)
4) If evaluation is requested (default: True), then a call to `evalf` is made.

**Options**

1) In #23541 a fast path for float arguments was added to the `eval` of `atan` and several other methods. This approach is specific to these functions are could lead to many small exceptions.

2) In the approach of this PR a more general approach was tried. In the `__new__` of the method it is checked whether the arguments are of float type or not (more details below). This works for many methods (e.g. `atan`), but some of the sympy tests have errors. One of the themes is that some functions rely (partly) on `eval` for the evaluation of floats. (e.g. `DiracDelta`,  does not implement the `evalf` itself)
Fixing the tests might be possible, but does not guarantee that the behavior is unchanged. (not everything is covered by tests, users can have subclassed sympy objects)

3) A third approach is to add a new attribute to the `Function` class `skip_eval_for_float`. The default value could be set to `False`, but for functions where skipping the eval is safe (checked manually) we can set it to `True`. An advantage is that it allows users to disable to new optimization for subclasses.
A variation is to set the default `skip_eval_for_float=True`, and set it to `False` for all function that fail tests.
This option is tested in #23582.

**Float type**

Which objects do we consider for a fast path? In any case the `Float` objects. But also complex numbers could be interesting. The check should be fast, otherwise performance of the evaluation of non-float numbers will suffer. Here are some options:

i) Only check on `Float`:

```
%timeit [isinstance(q, Float) for q in args]
189 ns ± 1.27 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

ii) Check with `_should_evalf`. This was already used in the `__new__` of `Function` (on the `result.args`, we need it on the `args`)
```
%timeit [atan._should_evalf(a) for a in args]
299 ns ± 5.22 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
iii) Test for complex numbers and floats. (could probably be improved)
```
def is_float_like(q):
    if isinstance(q, Float):
        return True
    if isinstance(q, Add):
        terms = q.args
        if len(terms)==2:
            if isinstance(terms[0], Float):
                if isinstance(terms[1], Mul):
                    terms = terms[1].args
                    print(terms)
                    if isinstance(terms[0], Float) and terms[1]==sympy.core.numbers.ImaginaryUnit():
                        return True
    return False
%timeit [is_float_like(q) for q in args]
241 ns ± 0.384 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

#### Release Notes

WIP

### DIFF
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -480,13 +480,17 @@ class Function(Application, Expr):
                 'plural': 's'*(min(cls.nargs) != 1),
                 'given': n})
 
-        evaluate = options.get('evaluate', global_parameters.evaluate)
-        result = super().__new__(cls, *args, **options)
-        if evaluate and isinstance(result, cls) and result.args:
-            pr2 = min(cls._should_evalf(a) for a in result.args)
-            if pr2 > 0:
-                pr = max(cls._should_evalf(a) for a in result.args)
-                result = result.evalf(prec_to_dps(pr))
+        evaluate = options.pop('evaluate', global_parameters.evaluate)
+ 
+        if evaluate and len(args)>0:
+             should = [cls._should_evalf(_sympify(a)) for a in args]
+             if not any(p <= 0 for p in should):
+                 pr = max(should)
+                 result = super().__new__(cls, *args, evaluate=False, **options)
+                 return result.evalf(prec_to_dps(pr))
+ 
+        result = super().__new__(cls, *args, evaluate=evaluate, **options)
+
 
         return _sympify(result)
 
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -481,7 +481,7 @@ class Function(Application, Expr):
                 'given': n})
 
         evaluate = options.pop('evaluate', global_parameters.evaluate)
- 
+
         if evaluate and len(args)>0:
              should = [cls._should_evalf(_sympify(a)) for a in args]
              if not any(p <= 0 for p in should):
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -483,7 +483,8 @@ class Function(Application, Expr):
         evaluate = options.pop('evaluate', global_parameters.evaluate)
 
         if evaluate and len(args)>0:
-             should = [cls._should_evalf(_sympify(a)) for a in args]
+             args =[sympify(a)for a in args]
+             should = [cls._should_evalf(a) for a in args]
              if not any(p <= 0 for p in should):
                  pr = max(should)
                  result = super().__new__(cls, *args, evaluate=False, **options)
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -488,9 +488,8 @@ class Function(Application, Expr):
                  pr = max(should)
                  result = super().__new__(cls, *args, evaluate=False, **options)
                  return result.evalf(prec_to_dps(pr))
- 
-        result = super().__new__(cls, *args, evaluate=evaluate, **options)
 
+        result = super().__new__(cls, *args, evaluate=evaluate, **options)
 
         return _sympify(result)
 


eendebakpt:
@oscarbenjamin This PR still has a number of failing tests, some help would be appreciated. One example I picked out: 
```
from sympy import DiracDelta
r=DiracDelta(5.1)
print(r)
```
returns `DiracDelta(5.1)` (should be `0`). 

Originally the value 5.1 was handled in `eval` (https://github.com/sympy/sympy/blob/master/sympy/functions/special/delta_functions.py#L218 I think). In this PR `eval` is skipped, but there is no key for `DiracDelta` in `sympy.core.evalf`.

eendebakpt:
I checked a few more failing tests. One of them is
```
from sympy import sin, cos, re, im
n = sin(1)**2 + cos(1)**2 - 1
w=n.n(2)
v=re(w)
print(v) # should be -0.e-110, but in this PR it is -1.e-110
```

The  problem is in the `re` function. The `eval` is not executed any more, hence `arg.is_extended_real` is never triggered.
```
    @classmethod
    def eval(cls, arg):
        if arg is S.NaN:
            return S.NaN
        elif arg is S.ComplexInfinity:
            return S.NaN
        elif arg.is_extended_real: # <------
            return arg
```

eendebakpt:
@oscarbenjamin It seems that skipping the `eval` part, even though the arguments are of Float type (the actual check is with `_should_evalf`) is not valid. Unless you have other suggestions I will close this PR.

oscarbenjamin:
Seems like a few problems that could be improved. The DiracDelta and SchurNumber cases suggest that `eval` should be tried if the numerical evaluation fails.

The problem with `re` is I guess a bug in numerical evaluation or Float handling of `_prec`.

eendebakpt:
> Seems like a few problems that could be improved. The DiracDelta and SchurNumber cases suggest that `eval` should be tried if the numerical evaluation fails.
> 

How would we check for failure? Do you mean check the output of `result.evalf(prec_to_dps(pr))`, e.g.
```
...
        if evaluate and len(args)>0:
             args =[sympify(a)for a in args]
             should = [cls._should_evalf(a) for a in args]
             if not any(p <= 0 for p in should):
                 pr = max(should)
                 result = super().__new__(cls, *args, evaluate=False, **options)
                 evaluated_result = result.evalf(prec_to_dps(pr))
                 if isinstance(evaluated_result, (Float,Add)): # or an Add of two floats
                     return evaluated_result
                 else:
                     # could not evaluate, go to full evaluation
                     result = super().__new__(cls, *args, evaluate=evaluate, **options)
                     evaluated_result = result.evalf(prec_to_dps(pr))
                     return evaluated_result
....
```

oscarbenjamin:
Yes, it could be something like that. Another possibility is to check if the result returned from evalf still has the same type (rather than evaluating to something else).

eendebakpt:
> Seems like a few problems that could be improved. The DiracDelta and SchurNumber cases suggest that `eval` should be tried if the numerical evaluation fails.
> 
> The problem with `re` is I guess a bug in numerical evaluation or Float handling of `_prec`.

Indeed fixes the `DiracDelta` issue. But it is not very elegant as we then have to retry and do the float evaluation again.

eendebakpt:
@oscarbenjamin  On the first message in this PR I added an overview of the experiments performed. My preference at this point is to continue with the third option. I created PR https://github.com/sympy/sympy/pull/23582 as an implementation. The PR still needs some improvements (documentation, tests), but I would like to have some feedback on the following items:

- The PR passes all tests (except for one unrelated test). But the behaviour of the `Function` constructor has changed, so we cannot be sure the behaviour of sympy in all cases is unchanged. Is this acceptable?
- Performance of evaluation of float arguments improves significantly for many functions (see first message of the PR). The evaluation of exact or symbolic expressions a bit slower. I tested with
```
from sympy import cos

w=S.Pi/4
def go():
    cos.__new__.cache_clear()
    return cos(w)
%timeit go()
# master: 18 µs ± 258 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
# PR: 19.7 µs ± 490 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
The performance is a bit lower because in this PR for `cos(pi/4)` we evaluate `should = [cls._should_evalf(a) for a in args]`, which is not required in the master branch.  The performance seems ok, since in practice `cos(pi/4)` will be cached, and in #23587, #23586 performance is improved so combined the performance is also better for exact and symbolic arguments.
In the end, this is a trade-off between the performance on symbolic expressions and float-like expressions.
     
- Should we make the new attribute `_skip_eval_float_arguments` public? (and perhaps give it a better name)

oscarbenjamin:
> My preference at this point is to continue with the third option.

That's probably best in terms of backwards compatibility. Does it also improve the problem with symbolic evaluation?

asmeurer:
I don't see the need for a new `skip_eval_for_float` attribute. `_should_evalf` should be sufficient.

> The performance is a bit lower because in this PR for cos(pi/4) we evaluate should = [cls._should_evalf(a) for a in args], which is not required in the master branch.

There's basically two ways that we can do this. Either we first call `eval` and only check if `_should_evalf` if it fails (what we currently do), or we call `_should_evalf` first only call `eval` if it returns False. One optimizes for the floating point case and one optimizes for the non-floating point case. In practice, I would say the most common case is not to have floating point inputs, so it may make more sense to keep it as it is. That also has the advantage of not breaking existing code that doesn't implement `_should_evalf` or that implements it in a way that doesn't match what `eval` would do. Although I expect very little user code overrides `_should_evalf` since isn't not even documented, and we don't even override it anywhere in SymPy.

Issue #23566, opened 2022-06-01
integrate(1/sqrt(x**2-1)) shouldn't be acosh(x)


eagleoflqj:
### originally
```py
>>> from sympy import *
>>> x=Symbol('x')
>>> integrate(1/sqrt(x**2-1))
acosh(x)
>>> integrate(1/sqrt(x**2-1), (x, 1, 2))
acosh(2)
>>> _.evalf()
1.31695789692482
>>> integrate(1/sqrt(x**2-1), (x, -2, -1))
-acosh(-2) + I*pi
>>> _.evalf()
-1.31695789692482 - 0.e-21*I
```
### now those all give the correct result, but requesting meijerg gives
```python
>>> integrate(1/sqrt(x**2 - 1), x, meijerg=True)
Piecewise((acosh(x), Abs(x**2) > 1), (-I*asin(x), True))
```

eagleoflqj:
The problem for heurisch is at
https://github.com/sympy/sympy/blob/ff00553cf2a03ce048bf46c25c8a3cbe2a636c2c/sympy/integrals/heurisch.py#L461-L465
It's easy to fix by using a `log` instead of `acosh`, although the `_eval_derivative` of `log` has to be implemented to deal with this special case.
The problem for meijerg is at
https://github.com/sympy/sympy/blob/ff00553cf2a03ce048bf46c25c8a3cbe2a636c2c/sympy/integrals/meijerint.py#L1747
where `sqrt(x**2)` becomes `x` instead of `Abs(x)`. I don't know if it's intended behavior for either `powdenest` itself or this place. It seems fixable by adding a real assumption to `x`. I'm not familiar with meijerg so I'll leave it for others to fix.
The problem for manual is easy to fix for me.

eagleoflqj:
Now this issue is only valid for meijerg.

oscarbenjamin:
> Now this issue is only valid for meijerg.

I guess this is down to this code although I'm not entirely sure how to interpret it:
https://github.com/sympy/sympy/blob/d0134b305ef088a3189c09e496934714ad303df9/sympy/simplify/hyperexpand.py#L134-L137

PR #23568, opened 2022-06-02
acosh(x).diff(x) == 1/(sqrt(x-1)*sqrt(x+1))

eagleoflqj:
#### References to other Issues or PRs

This justifies #23566
Wolfram Alpha result: https://www.wolframalpha.com/input?i=arccosh%28x%29+derivative
#### Release Notes

* functions
  * Fix derivative of acosh

PR #23573, opened 2022-06-03
Fix integrate 1/sqrt(x**2-1) for heurisch

eagleoflqj:

#23566 heurisch part. It shouldn't be closed.

Use `log` instead of `acosh`. ~~Implement a special case for evaluating derivative of log.~~ Patch heurisch to simplify a special case of derivative. 
#### Release Notes

* integrals
  * Fix integrate(1/sqrt(x**2-1))

### DIFF
--- a/sympy/integrals/heurisch.py
+++ b/sympy/integrals/heurisch.py
@@ -15,7 +15,7 @@ from sympy.core.sorting import ordered
 from sympy.core.traversal import iterfreeargs
 
 from sympy.functions import exp, sin, cos, tan, cot, asin, atan
-from sympy.functions import log, sinh, cosh, tanh, coth, asinh, acosh
+from sympy.functions import log, sinh, cosh, tanh, coth, asinh
 from sympy.functions import sqrt, erf, erfi, li, Ei
 from sympy.functions import besselj, bessely, besseli, besselk
 from sympy.functions import hankel1, hankel2, jn, yn
@@ -403,6 +403,7 @@ def heurisch(f, x, rewrite=False, hints=None, mappings=None, retries=3,
             rewrite = True
 
     terms = components(f, x)
+    dcache = DiffCache(x)
 
     if hints is not None:
         if not hints:
@@ -462,15 +463,16 @@ def heurisch(f, x, rewrite=False, hints=None, mappings=None, retries=3,
 
                         if M is not None and M[b].is_positive:
                             if M[a].is_positive:
-                                terms.add(acosh(sqrt(M[a]/M[b])*x))
+                                dF = 1/sqrt(M[a]*x**2 - M[b])
+                                F = log(2*sqrt(M[a])*sqrt(M[a]*x**2 - M[b]) + 2*M[a]*x)/sqrt(M[a])
+                                dcache.cache[F] = dF  # hack: F.diff(x) doesn't automatically simplify to f
+                                terms.add(F)
                             elif M[a].is_negative:
                                 terms.add(-M[b]/2*sqrt(-M[a])*
                                            atan(sqrt(-M[a])*x/sqrt(M[a]*x**2 - M[b])))
 
         else:
             terms |= set(hints)
-
-    dcache = DiffCache(x)
 
     for g in set(terms):  # using copy of terms
         terms |= components(dcache.get_diff(g), x)
--- a/sympy/integrals/tests/test_integrals.py
+++ b/sympy/integrals/tests/test_integrals.py
@@ -1,3 +1,4 @@
+import math
 from sympy.concrete.summations import (Sum, summation)
 from sympy.core.add import Add
 from sympy.core.containers import Tuple
@@ -1040,7 +1041,7 @@ def test_issue_4403():
     assert integrate(sqrt(x**2 + z**2), x) == \
         z**2*asinh(x/z)/2 + x*sqrt(x**2 + z**2)/2
     assert integrate(sqrt(x**2 - z**2), x) == \
-        -z**2*acosh(x/z)/2 + x*sqrt(x**2 - z**2)/2
+        x*sqrt(x**2 - z**2)/2 - z**2*log(x + sqrt(x**2 - z**2))/2
 
     x = Symbol('x', real=True)
     y = Symbol('y', positive=True)
@@ -1918,6 +1919,12 @@ def test_issue_18527():
     expr = (cos(x)/(4+(sin(x))**2))
     res_real = integrate(expr.subs(x, xr), xr, manual=True).subs(xr, x)
     assert integrate(expr, x, manual=True) == res_real == Integral(expr, x)
+
+
+def test_issue_23566():
+    i = integrate(1/sqrt(x**2-1), (x, -2, -1))
+    assert i == -log(2 - sqrt(3))
+    assert math.isclose(i.n(), 1.31695789692482)
 
 
 def test_pr_23583():
--- a/sympy/integrals/tests/test_heurisch.py
+++ b/sympy/integrals/tests/test_heurisch.py
@@ -201,6 +201,9 @@ def test_heurisch_hacking():
     assert heurisch(1/sqrt(9 + 4*x**2), x, hints=[]) == \
         asinh(x*Rational(2, 3))/2
 
+    assert heurisch(1/sqrt(3*x**2-4), x, hints=[]) == \
+           sqrt(3)*log(3*x + sqrt(3)*sqrt(3*x**2 - 4))/3
+
 
 def test_heurisch_function():
     assert heurisch(f(x), x) is None
--- a/sympy/solvers/ode/tests/test_single.py
+++ b/sympy/solvers/ode/tests/test_single.py
@@ -40,7 +40,7 @@ from sympy.core.singleton import S
 from sympy.core.symbol import (Dummy, symbols)
 from sympy.functions.elementary.complexes import (im, re)
 from sympy.functions.elementary.exponential import (LambertW, exp, log)
-from sympy.functions.elementary.hyperbolic import (acosh, asinh, cosh, sinh, tanh)
+from sympy.functions.elementary.hyperbolic import (asinh, cosh, sinh, tanh)
 from sympy.functions.elementary.miscellaneous import (cbrt, sqrt)
 from sympy.functions.elementary.piecewise import Piecewise
 from sympy.functions.elementary.trigonometric import (acos, asin, atan, cos, sec, sin, tan)
@@ -2859,7 +2859,8 @@ def _get_examples_ode_sol_1st_homogeneous_coeff_best():
 
     '1st_homogeneous_coeff_best_08': {
         'eq': f(x)**2 + (x*sqrt(f(x)**2 - x**2) - x*f(x))*f(x).diff(x),
-        'sol': [Eq(log(x), C1 - log(f(x)/x) + acosh(f(x)/x))],
+        'sol': [Eq(f(x), -sqrt(-x*exp(2*C1)/(x - 2*exp(C1)))), Eq(f(x), sqrt(-x*exp(2*C1)/(x - 2*exp(C1))))],
+        'checkodesol_XFAIL': True  # solutions are valid in a range
     },
     }
     }


sympy/functions/elementary/exponential.py
```diff
@@ -645,6 +645,23 @@ def fdiff(self, argindex=1):
...
 
+    def _eval_derivative(self, s):
+        if self.args[0].has(Pow):
+            a = Wild('a', exclude=[s])
+            b = Wild('b', exclude=[s, 0])
+            c = Wild('c', exclude=[s, 0])
+            d = Wild('d', exclude=[s])
+            e = Wild('e', exclude=[s])
+            f = Wild('f', exclude=[s, 0])
+            M = self.args[0].match(a+b*s+c*sqrt(d+e*s+f*s**2))
```
oscarbenjamin, line 656: This seems pretty expensive for a derivative calculation.

oscarbenjamin: It also seems a surprisingly specific.

Why is this change to differentiating logs needed?

eagleoflqj: > Why is this change to differentiating logs needed?

When I changed the derivative of `acosh` in #23568, this test failed
https://github.com/sympy/sympy/blob/ff00553cf2a03ce048bf46c25c8a3cbe2a636c2c/sympy/integrals/tests/test_integrals.py#L1042-L1043
which means heurisch uses `acosh`'s derivative and compare it to the original expression.
Here I replaced `acosh` to `log`. For the same reason, the derivative of `log` will be compared to original expression. Without this change, heurisch can't deal with this test.

This change is tested in `test_log_diff`. I can't find a way to simplify the derivative of `log(b + 2*sqrt(c)*sqrt(a + b*x + c*x**2) + 2*c*x)` to `sqrt(c)/sqrt(a + b*x + c*x**2)`, and even if there exist a way, it's usually not recommended to post-simplify a step in SymPy code base.

> This seems pretty expensive for a derivative calculation.

1.10.1:
```py
In [3]: %timeit log(x**2+1).diff()
81.8 µs ± 459 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
79dee2fd50166253be1eb8eb3577ab38f605218f:
```py
In [3]: %timeit log(x**2+1).diff()
38.7 ms ± 340 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
So, yes it's too expensive.
I'll push a change to not use that much `Wilds` and the time should be
```py
In [3]: %timeit log(x**2+1).diff()
821 µs ± 2.34 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

oscarbenjamin: Unless there is a bug in the derivative of `log` I would rather just XFAIL the test than make any changes to the derivative of `log` regardless of speed.

Is the antiderivative in the test even correct though? With #23573 I get:
```python
In [32]: x = symbols('x')

In [33]: z = symbols('z', positive=True)

In [34]: F = -z**2*acosh(x/z)/2 + x*sqrt(x**2 - z**2)/2

In [35]: f = sqrt(x**2 - z**2)

In [36]: zero = F.diff(x) - f

In [37]: zero.subs({z:1,x:-2})
Out[37]: 
√3
──
3 
```

eagleoflqj: The original antiderivative `-z**2*acosh(x/z)/2 + x*sqrt(x**2 - z**2)/2` was wrong. So I changed it to `x*sqrt(x**2 - z**2)/2 - z**2*log(x + sqrt(x**2 - z**2))/2` in this PR and the test can pass with the `log` derivative implementation.
Is XFAIL better even if it can be corrected? The zen of SymPy says
https://github.com/sympy/sympy/blob/645bbb06cd289165a9144d4c2c64bf17b664f8ac/sympy/this.py#L13-L14

oscarbenjamin: The reason I don't want to change log's derivative is because as far as I can tell it is correct. You haven't said that anything about it is incorrect so why is this change to log needed?

We can't just add random special cases everywhere. The correctness and simplicity of log is more important than evaluating any integral. If there is something that can be improved in heurisch or simplify or something then that can be considered but this is PR is trying to fix the problem in the wrong place.

oscarbenjamin: I think `simplify` can certainly be improved here although I'm not sure what's the best approach. Only `apart` seems capable of handling something like this with all of these square roots:
```python
In [67]: f = F.diff(x).subs(sqrt(a + b*x + c*x**2), z)

In [68]: f
Out[68]: 
     ⎛b      ⎞      
2⋅√c⋅⎜─ + c⋅x⎟      
     ⎝2      ⎠      
────────────── + 2⋅c
      z             
────────────────────
 b + 2⋅√c⋅z + 2⋅c⋅x 

In [69]: f.simplify()
Out[69]: 
√c⋅(b + 2⋅c⋅x) + 2⋅c⋅z
──────────────────────
z⋅(b + 2⋅√c⋅z + 2⋅c⋅x)

In [70]: ratsimp(f)
Out[70]: 
          3/2          
b⋅√c + 2⋅c   ⋅x + 2⋅c⋅z
───────────────────────
            2          
b⋅z + 2⋅√c⋅z  + 2⋅c⋅x⋅z

In [71]: radsimp(f)
Out[71]: 
 2           3/2        5/2  2      3/2  2
b ⋅√c + 4⋅b⋅c   ⋅x + 4⋅c   ⋅x  - 4⋅c   ⋅z 
──────────────────────────────────────────
     ⎛ 2                2  2        2⎞    
   z⋅⎝b  + 4⋅b⋅c⋅x + 4⋅c ⋅x  - 4⋅c⋅z ⎠    

In [72]: f.together()
Out[72]: 
√c⋅(b + 2⋅c⋅x) + 2⋅c⋅z
──────────────────────
z⋅(b + 2⋅√c⋅z + 2⋅c⋅x)

In [73]: f.expand()
Out[73]: 
                                     3/2                              
          b⋅√c                    2⋅c   ⋅x                 2⋅c        
─────────────────────── + ─────────────────────── + ──────────────────
            2                         2             b + 2⋅√c⋅z + 2⋅c⋅x
b⋅z + 2⋅√c⋅z  + 2⋅c⋅x⋅z   b⋅z + 2⋅√c⋅z  + 2⋅c⋅x⋅z                     

In [74]: f.expand(denom=True)
Out[74]: 
     ⎛b      ⎞      
2⋅√c⋅⎜─ + c⋅x⎟      
     ⎝2      ⎠      
────────────── + 2⋅c
      z             
────────────────────
 b + 2⋅√c⋅z + 2⋅c⋅x 

In [87]: factor(f)
Out[87]: 
          3/2          
b⋅√c + 2⋅c   ⋅x + 2⋅c⋅z
───────────────────────
 z⋅(b + 2⋅√c⋅z + 2⋅c⋅x)

In [75]: f.apart(z)
Out[75]: 
√c
──
z 
```
Note that if you substitute c=1 then almost all of the above will succeed:
```python
In [88]: f.subs(c, 1)
Out[88]: 
      ⎛b    ⎞
    2⋅⎜─ + x⎟
      ⎝2    ⎠
2 + ─────────
        z    
─────────────
b + 2⋅x + 2⋅z

In [89]: f.subs(c, 1).together()
Out[89]: 
1
─
z
```

eagleoflqj: > why is this change to log needed?

[Wolfram alpha](https://www.wolframalpha.com/input?i=derivative+of+log%28b+%2B+2*sqrt%28c%29*sqrt%28a+%2B+b*x+%2B+c*x**2%29+%2B+2*c*x%29) already provides a simplified answer. I don't know the black magic behind it but it doesn't need users to manually call simplify.
Also, providing a simpler result does take more time but can also reduce processing time afterwards. Given `log(b + 2*sqrt(c)*sqrt(a + b*x + c*x**2) + 2*c*x).diff(x).simplify() == (sqrt(c)*(b + 2*c*x) + 2*c*sqrt(a + b*x + c*x**2))/(sqrt(a + b*x + c*x**2)*(b + 2*sqrt(c)*sqrt(a + b*x + c*x**2) + 2*c*x))`, it's not easy for user to simplify this result afterwards, and complexity accumulates fast.

eagleoflqj:
With checking `s**2` and `S.Half`, any expressions without these 2 elements won't be affected much. With multi-stage match, most irrelevant expressions will fail and break early. The expressions that pass all matches are worth a try to simplify.

oscarbenjamin:
I'll repeat what I said earlier:

Unless there is a bug in the derivative of log I would rather just XFAIL the integrals test than make any changes to the derivative of log *regardless of speed*.

While the speed does concern me the more important point is that the code for differentiation of logs should not be complicated by adding checks for random special cases. Just think for a moment how many other possible expression patterns we could check for here: this is clearly not a scalable way to implement differentiation. Ideally we would use *much* better algorithms for differentiation but that's impossible if those algorithms are expected to produce arbitrarily defined results in random special cases.

eagleoflqj:
It’s not a *random* special case. It helps fixing a long standing bug elsewhere, meanwhile provides a nicer result for derivative itself. SymPy is a rule-based library so there will never be too many rules to add (think about the model that SymPy tries to compete: mathematica, its size is perhaps 100x of SymPy, it even hard-codes solution for some special quintic equations). There are only awkward rules that should be replaced by smarter rules.

oscarbenjamin:
> It helps fixing a long standing bug elsewhere,

That bug can be fixed without the change here. If necessary the integral test can be XFAILed. It previously succeeded only by depending on buggy behaviour in the derivative of `acosh` and the result that it tested was incorrect.

Fixing `acosh.diff` now prevents `heurisch` from giving a wrong result in this case which is good. Of course it is good to investigate the problem with heurisch and to think about other improvements that are related to the change (as you are doing). Finding some other way for `heurisch` to evaluate that integral correctly is not necessarily a precondition for fixing the bug in `acosh.diff` though.

Making this particular integral work is not a reason to change `log.diff`. The complicated derivative that you showed should be easier to simplify but that's a bug in the *simplification* code. The output from `log` is perfectly correct and fine and follows the obvious definition for differentiation of logs:
```python
In [8]: a, b, c = symbols('a, b, c')

In [9]: F = log(b + 2*sqrt(c)*sqrt(a + b*x + c*x**2) + 2*c*x)

In [10]: F.diff(x)
Out[10]: 
             ⎛b      ⎞              
        2⋅√c⋅⎜─ + c⋅x⎟              
             ⎝2      ⎠              
     ─────────────────── + 2⋅c      
        ________________            
       ╱              2             
     ╲╱  a + b⋅x + c⋅x              
────────────────────────────────────
            ________________        
           ╱              2         
b + 2⋅√c⋅╲╱  a + b⋅x + c⋅x   + 2⋅c⋅x
```
That's a nice clear application of the chain rule for the `log` and then the `sqrt` which is precisely what it should be. The problem of course is why it doesn't simplify e.g.:
```python
In [7]: F.diff(x).factor()
Out[7]: 
                                 ________________         
                  3/2           ╱              2          
        b⋅√c + 2⋅c   ⋅x + 2⋅c⋅╲╱  a + b⋅x + c⋅x           
──────────────────────────────────────────────────────────
   ________________ ⎛            ________________        ⎞
  ╱              2  ⎜           ╱              2         ⎟
╲╱  a + b⋅x + c⋅x  ⋅⎝b + 2⋅√c⋅╲╱  a + b⋅x + c⋅x   + 2⋅c⋅x⎠
```
The reason is that most simplification routines depend in some way on the polynomial code but the polynomial code doesn't understand how to treat `sqrt(c)` and `c` as being related e.g.:
```python
In [3]: Poly(sqrt(c) + c**2)
Out[3]: Poly(c**2 + (sqrt(c)), c, sqrt(c), domain='ZZ')

In [4]: Poly(sqrt(c) + c**2).gens
Out[4]: (c, √c)

In [5]: Poly(sqrt(c) + c**2) ** 2
Out[5]: Poly(c**4 + 2*c**2*(sqrt(c)) + (sqrt(c))**2, c, sqrt(c), domain='ZZ')
```
Notice that `c` and `sqrt(c)` are being treated as independent symbols (generators). That means that polys doesn't know that `sqrt(c)**2 = c` and it also doesn't understand that `c` can be factored as `sqrt(c)*sqrt(c)`. This causes all sorts of things `cancel`, `factor`, `together` etc to fail.

Something like this can make the polys code smarter about square-rooted symbols (although it needs to be applied in more places):
```diff
diff --git a/sympy/polys/polyutils.py b/sympy/polys/polyutils.py
index 82c8c83..1798171 100644
--- a/sympy/polys/polyutils.py
+++ b/sympy/polys/polyutils.py
@@ -273,6 +273,28 @@ def _is_coeff(factor):
 
         reprs.append(terms)
 
+    #
+    # See if e.g. sqrt(c) and c are both in gens
+    # If they are then rewrite c as sqrt(c)**2
+    #
+    to_replace = {}
+    for g1 in gens:
+        if g1.is_Pow and g1.base in gens and g1.exp.is_Rational and g1.exp.p == 1:
+            to_replace[g1.base] = (g1, g1.exp.q)
+
+    if to_replace:
+
+        to_replace_set = to_replace.keys()
+
+        for terms in reprs:
+            for coeff, term in terms:
+                for g2 in to_replace_set & term:
+                    exp2 = term.pop(g2)
+                    g1, exp1 = to_replace[g2]
+                    term[g1] = term.get(g1, 0) + exp1*exp2
+
+        gens -= to_replace_set
+
     gens = _sort_gens(gens, opt=opt)
     k, indices = len(gens), {}
 
```
With that you can have:
```python
In [13]: F.diff(x).factor()
Out[13]: 
         √c        
───────────────────
   ________________
  ╱              2 
╲╱  a + b⋅x + c⋅x  
```

eagleoflqj:
> the polynomial code doesn't understand how to treat `sqrt(c)` and `c` as being related e.g.:

Yes. I have a local fix for polynomial but it breaks some solver tests and I haven't fixed them.
OK, I'll try to patch heurisch code.

eagleoflqj:
How to XFAIL this test? I'm not familiar with ODE.
https://github.com/sympy/sympy/blob/78bbfa804d00899aa92d61eb4f012d6f31d928c3/sympy/solvers/ode/tests/test_single.py#L2860-L2863

oscarbenjamin:
You can do it like
https://github.com/sympy/sympy/blob/78bbfa804d00899aa92d61eb4f012d6f31d928c3/sympy/solvers/ode/tests/test_single.py#L673
The list is the list of `dsolve` hints that fails (check `classify_ode`).

oscarbenjamin:
> How to XFAIL this test?

Why does it fail?

sympy/functions/elementary/exponential.py
```diff
@@ -645,6 +645,37 @@ def fdiff(self, argindex=1):
         else:
             raise ArgumentIndexError(self, argindex)
 
+    def _eval_derivative(self, s):
```
asmeurer, line 648: Redefining this makes `fdiff` above effectively dead code.

eagleoflqj: I'll revert this and patch heurisch instead.

asmeurer:
I've mentioned it before, but it would really help to have a separate method for computing derivatives just for heurisch. See https://github.com/sympy/sympy/issues/4300 and the cross-linked issues.

eagleoflqj:
> Why does it fail?

The 2 new solutions are valid only on a given range related to `C1`:
```py
from sympy import *

x,f,C1 = symbols('x f C1')

def test(f):
    equ = f**2 + (x*sqrt(f**2 - x**2) - x*f)*f.diff(x)
    print(equ.subs({x: -1, C1: -1}).n())
    print(equ.subs({x: 1, C1: -1}).n())
    print(equ.subs({x: -1, C1: 1}).n())
    print(equ.subs({x: 1, C1: 1}).n())

test(-sqrt(-x*exp(2*C1)/(x - 2*exp(C1))))
test(sqrt(-x*exp(2*C1)/(x - 2*exp(C1))))
```
output:
```
0.e-126
-2.45041787584235
0.e-125
0.e-125
-0.122888165622335
0.e-125
-1.32633468246037
1.29008974262908
```
I XFAIL it for checksol.

eagleoflqj:
> I've mentioned it before, but it would really help to have a separate method for computing derivatives just for heurisch.

I saw PRs aimed for it. They are not finished as the problem is not as trivial as previously thought. To fix https://github.com/sympy/sympy/issues/23566 that seems over complicated.

oscarbenjamin:
> I XFAIL it for checksol.

Would the previous solution also fail if the derivative of `acosh` was fixed as in #23566?

eagleoflqj:
> Would the previous solution also fail if the derivative of `acosh` was fixed as in #23566?

No, to my surprise it still passes. 
It is because when checking solution, the solution is `posify`ed at https://github.com/sympy/sympy/blob/6a04cdb14427489fe49974f33a84001a299d3688/sympy/solvers/ode/subscheck.py#L276

So whether `acosh(x).diff(x)==1/sqrt(x**2-1)` or `acosh(x).diff(x)==1/(sqrt(x+1)*sqrt(x-1))` doesn't matter here:
```py
x,y=symbols('x y', positive=True)
simplify(sqrt(x+y)*sqrt(x-y))  # == sqrt(x**2-y**2)
```

oscarbenjamin:
> the solution is `posify`ed at

Yeah, that shouldn't be there...

sympy/integrals/heurisch.py
```diff
@@ -462,16 +463,17 @@ def heurisch(f, x, rewrite=False, hints=None, mappings=None, retries=3,
 
                         if M is not None and M[b].is_positive:
                             if M[a].is_positive:
-                                terms.add(acosh(sqrt(M[a]/M[b])*x))
+                                dF = 1/sqrt(M[a]*x**2 - M[b])
+                                F = log(2*sqrt(M[a])*sqrt(M[a]*x**2 - M[b]) + 2*M[a]*x)/sqrt(M[a])
+                                dcache.cache[F] = dF  # hack: F.diff(x) doesn't automatically simplify to f
```
oscarbenjamin, line 468: We should open a separate issue about this not simplifying.

sympy/integrals/tests/test_integrals.py
```diff
@@ -1920,6 +1920,12 @@ def test_issue_18527():
     assert integrate(expr, x, manual=True) == res_real == Integral(expr, x)
 
 
+def test_issue_23566():
+    i = integrate(1/sqrt(x**2-1), (x, -2, -1))
+    assert i == -log(2 - sqrt(3))
+    assert str(i.n()) == '1.31695789692482'
```
oscarbenjamin, line 1926: This should use `math.isclose`.

oscarbenjamin:
I think this looks good.

PR #23580, opened 2022-06-05
adding option to return kinematic equations implicitly in KanesMethod #22818

### LINKED ISSUE #22626: KanesMethod solves kinematics explicitly even if not asked

zouhairm:
Fixes #22626 

This adds an option to KanesMethod to expose kinematic matrices in implicit form.

Default behavior should remain unchanged, so existing tests should be fine. Added one test to exercise the new code.

Closing PR https://github.com/sympy/sympy/pull/22818 to take a different approach as proposed by @moorepants in https://github.com/sympy/sympy/pull/22818#issuecomment-1086811861 . Doing this in a different PR in case we decide to pursue the other method at some point (and because I have some 3rd party code relying on that branch right now which I'll migrate to this PR once it's approved)

* physics.mechanics
  * Added option to return kinematics and full mass matrix in implicit form in KanesMethod

### DIFF
--- a/sympy/physics/mechanics/kane.py
+++ b/sympy/physics/mechanics/kane.py
@@ -35,25 +35,32 @@ class KanesMethod(_Methods):
     loads : iterable
         Iterable of (Point, vector) or (ReferenceFrame, vector) tuples
         describing the forces on the system.
-    auxiliary : Matrix
+    auxiliary_eqs : Matrix
         If applicable, the set of auxiliary Kane's
         equations used to solve for non-contributing
         forces.
     mass_matrix : Matrix
-        The system's mass matrix
+        The system's dynamics mass matrix: [k_d; k_dnh]
     forcing : Matrix
-        The system's forcing vector
+        The system's dynamics forcing vector: -[f_d; f_dnh]
     mass_matrix_kin : Matrix
-        The "mass matrix" for kinematic differential equations
+        The "mass matrix" for kinematic differential equations: k_kqdot
     forcing_kin : Matrix
-        The forcing vector for kinematic differential equations
+        The forcing vector for kinematic differential equations: -(k_ku*u + f_k)
     mass_matrix_full : Matrix
-        The "mass matrix" for the u's and q's with explicit kinematics
+        The "mass matrix" for the u's and q's with dynamics and explicit kinematics
     forcing_full : Matrix
-        The "forcing vector" for the u's and q's with explicit kinematics
+        The "forcing vector" for the u's and q's with dynamics and explicit kinematics
 
-    By default, the kinematic and "full" mass matrix and forcing vector are given
-    with the kinematic equations in explicit form. The method also provides an `_implicit` version for those matrices/vectors where the kinematics are left in implicit form
+    Notes
+    =====
+
+    The mass matrices and forcing vectors which involve kinematic equations are given in explicit form.
+    In other words, mass_matrix_kin = k_kqdot = identity
+
+    The Kane object also provides an ``_implicit`` version for those matrices/vectors where the kinematics
+    are left in implicit form, e.g. with mass_matrix_kin_implicit is not necessarily an identity
+    This can provide more compact equations for non-simple kinematics (see https://github.com/sympy/sympy/issues/22626)
 
     Examples
     ========
@@ -559,7 +566,7 @@ class KanesMethod(_Methods):
         constraints) the length of the returned vectors will be o - m + s in
         length. The first o - m equations will be the constrained Kane's
         equations, then the s auxiliary Kane's equations. These auxiliary
-        equations can be accessed with the auxiliary_eqs().
+        equations can be accessed with the auxiliary_eqs property.
 
         Parameters
         ==========
@@ -672,19 +679,22 @@ class KanesMethod(_Methods):
             raise ValueError('Need to compute Fr, Fr* first.')
         return -Matrix([self._f_d, self._f_dnh])
 
+
     def _mass_matrix_kin(self, explicit=True):
         """
-        Returns the "mass matrix" Y_q for the kinematic differential equations
-        Y_q*q' = f_q = Y_u*u + f(q, t)
-        If explicit=True, Y_q = identity
+        Returns the "mass matrix" k_qqdot for the kinematic differential equations
+
+        k_qqdot * q' = f_q = -k_ku * u - f_k(q, t)
+        If explicit=True (default behavior), k_qqdot = identity
         """
         return self._k_kqdot if explicit else self._k_kqdot_implicit
 
     def _forcing_kin(self, explicit=True):
         """
         Returns the "forcing vector" for the kinematic differential equation
-        Y_q*q' = f_q = Y_u*u + f(q, t)
-        If explicit=True (default behavior), assumes Y_q = identity
+
+        k_qqdot * q' = f_q = -k_ku * u - f_k(q, t)
+        If explicit=True (default behavior), implies k_qqdot = identity
         """
         if explicit:
             return -(self._k_ku * Matrix(self.u) + self._f_k)
@@ -693,22 +703,22 @@ class KanesMethod(_Methods):
 
     @property
     def mass_matrix_kin(self):
-        """The mass matrix of the the kinematic differential equations in explicit form."""
+        """The mass matrix of the the kinematic differential equations k_qqdot (explicit form, i.e. k_qqdot=identity)."""
         return self._mass_matrix_kin(True)
 
     @property
     def mass_matrix_kin_implicit(self):
-        """The mass matrix of the the kinematic differential equations in implicit form."""
+        """The mass matrix of the the kinematic differential equations k_qqdot (implicit form)."""
         return self._mass_matrix_kin(False)
 
     @property
     def forcing_kin(self):
-        """The focing vector  the the kinematic differential equations in explicit form."""
+        """The focing vector  the the kinematic differential equations -k_ku * u - f_k (explicit form)."""
         return self._forcing_kin(True)
 
     @property
     def forcing_kin_implicit(self):
-        """The focing vector  the the kinematic differential equations in implicit form."""
+        """The focing vector  the the kinematic differential equations -k_ku * u - f_k (implicit form)."""
         return self._forcing_kin(False)
 
     def _mass_matrix_full(self, explicit=True):
--- a/sympy/physics/mechanics/kane.py
+++ b/sympy/physics/mechanics/kane.py
@@ -43,10 +43,17 @@ class KanesMethod(_Methods):
         The system's mass matrix
     forcing : Matrix
         The system's forcing vector
+    mass_matrix_kin : Matrix
+        The "mass matrix" for kinematic differential equations
+    forcing_kin : Matrix
+        The forcing vector for kinematic differential equations
     mass_matrix_full : Matrix
-        The "mass matrix" for the u's and q's
+        The "mass matrix" for the u's and q's with explicit kinematics
     forcing_full : Matrix
-        The "forcing vector" for the u's and q's
+        The "forcing vector" for the u's and q's with explicit kinematics
+
+    By default, the kinematic and "full" mass matrix and forcing vector are given
+    with the kinematic equations in explicit form. The method also provides an `_implicit` version for those matrices/vectors where the kinematics are left in implicit form
 
     Examples
     ========
@@ -281,6 +288,10 @@ class KanesMethod(_Methods):
                        'coordinates.')
                 raise ValueError(msg.format(nonlin_vars))
 
+            self._f_k_implicit = f_k.xreplace(uaux_zero)
+            self._k_ku_implicit = k_ku.xreplace(uaux_zero)
+            self._k_kqdot_implicit = k_kqdot
+
             # Solve for q'(t) such that the coefficient matrices are now in
             # this form:
             #
@@ -289,19 +300,19 @@ class KanesMethod(_Methods):
             # NOTE : Solving the kinematic differential equations here is not
             # necessary and prevents the equations from being provided in fully
             # implicit form.
-            f_k = k_kqdot.LUsolve(f_k)
-            k_ku = k_kqdot.LUsolve(k_ku)
-            k_kqdot = eye(len(qdot))
-            self._qdot_u_map = dict(zip(qdot, -(k_ku*u + f_k)))
+            f_k_explicit = k_kqdot.LUsolve(f_k)
+            k_ku_explicit = k_kqdot.LUsolve(k_ku)
+            self._qdot_u_map = dict(zip(qdot, -(k_ku_explicit*u + f_k_explicit)))
 
-            self._f_k = f_k.xreplace(uaux_zero)
-            self._k_ku = k_ku.xreplace(uaux_zero)
-            self._k_kqdot = k_kqdot
+            self._f_k = f_k_explicit.xreplace(uaux_zero)
+            self._k_ku = k_ku_explicit.xreplace(uaux_zero)
+            self._k_kqdot = eye(len(qdot))
+
         else:
             self._qdot_u_map = None
-            self._f_k = Matrix()
-            self._k_ku = Matrix()
-            self._k_kqdot = Matrix()
+            self._f_k_implicit = self._f_k = Matrix()
+            self._k_ku_implicit = self._k_ku = Matrix()
+            self._k_kqdot_implicit = self._k_kqdot = Matrix()
 
     def _form_fr(self, fl):
         """Form the generalized active force."""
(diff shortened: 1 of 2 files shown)


zouhairm:
@moorepants - plz let me know if this along the lines of what you're looking for.

moorepants:
@TJStienstra would you like to review this? It would be a good practice for your GSoC.

TJStienstra:
Yeah fine, thanks for showing. Will have a lot more time this week, so will look into it somewhere in the coming days (probably tomorrow).

zouhairm:
thanks for the comments @TJStienstra - will try to address them this week.

sympy/physics/mechanics/tests/test_kane.py
```diff
@@ -348,3 +353,113 @@ def test_input_format():
...
+    # Expecting implicit form to be less than 5% of the flops
+    n_ops_explicit = sum(
+        [x.count_ops() for x in KM.forcing_full] +
+        [x.count_ops() for x in KM.mass_matrix_full]
+    )
+    n_ops_implicit = sum(
+        [x.count_ops() for x in KM.forcing_full_implicit] +
+        [x.count_ops() for x in KM.mass_matrix_full_implicit]
+    )
+    assert n_ops_implicit / n_ops_explicit < .05
```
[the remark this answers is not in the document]
zouhairm: the other test is actually in `test_two_dof`

I added another test, although turned out this is not so easy as I couldn't use the whole explicit form as a way to check things against since the explicit equations are too slow to handle (hence the whole point of this PR :p)

I've  basically used the explicit form for the linear velocities which are 'trivial', and a 'textbook' solution for the quaternion rates which are more involved and thus their explicit form is too complex.

zouhairm:
@TJStienstra - sorry for the much longer than expected delay as some higher priorities ended up taking over.

I believe I've addressed all of your comments. I'm hoping we can get this merged as it'd be great to get our internal codebase updated to use a 'built wheel' from sympy as opposed to having to build it from scratch from my branch.

Thanks!

TJStienstra:
Nice work, like the tests and I agree it would be nice to have this in the code base as soon as possible. Overall there are some small typos, but will be addressing a more general refactoring of `KanesMethod` in #24093.

However just discussed the PR with @moorepants and @brocksam and came with the following proposal to reduce the current polutting (~100 lines of code for these properties and 3 extra external methods and 3 protected methods). By implementing an attribute `explicit`, which is defaultly set to `True` we can drastically reduce it. Here is some example code:

```py
kanes = KanesMethod(...)
kanes.kanes_equations()
kanes.forcing_kin  # Explicit kinematic forcing vector
kanes.mass_matrix_full # Explicit full mass matrix
kanes.explicit = False
kanes.forcing_kin  # Implicit kinematic forcing vector
```

@zouhairm What do you think of this proposal (it is in the end a bit of what you've started with)?

Will be giving the entire suggestion, that you can check and commit, in a few minutes.

sympy/physics/mechanics/tests/test_kane.py
```diff
@@ -77,6 +79,10 @@ def test_two_dof():
     assert simplify(KM.rhs() -
                     KM.mass_matrix_full.LUsolve(KM.forcing_full)) == zeros(4, 1)
 
+    # Check that the implicit form also gives the same result
+    assert simplify(KM.rhs() -
+                    KM.mass_matrix_full_implicit.LUsolve(KM.forcing_full_implicit)) == zeros(4, 1)
```
TJStienstra, line 84:
```sugggestion
    rhs = KM.rhs()
    assert simplify(rhs -
                    KM.mass_matrix_full.LUsolve(KM.forcing_full)) == zeros(4, 1)

    # Check that the implicit form also gives the same result
    KM.explicit = False
    assert simplify(rhs -
                    KM.mass_matrix_full.LUsolve(KM.forcing_full)) == zeros(4, 1)
```

sympy/physics/mechanics/tests/test_kane.py
```diff
@@ -348,3 +354,147 @@ def test_input_format():
...
+
+    # Expecting implicit form to be less than 5% of the flops
+    n_ops_explicit = sum(
+        [x.count_ops() for x in KM.forcing_full] +
+        [x.count_ops() for x in KM.mass_matrix_full]
+    )
+    n_ops_implicit = sum(
+        [x.count_ops() for x in KM.forcing_full_implicit] +
+        [x.count_ops() for x in KM.mass_matrix_full_implicit]
+    )
```
TJStienstra, line 465:
```suggestion
    KM.explicit = False
    n_ops_implicit = sum(
        [x.count_ops() for x in KM.forcing_full] +
        [x.count_ops() for x in KM.mass_matrix_full]
    )
```

sympy/physics/mechanics/tests/test_kane.py
```diff
@@ -348,3 +354,147 @@ def test_input_format():
...
+    assert n_ops_implicit / n_ops_explicit < .05
+
+    # Ideally we would check that implicit and explicit equations give the same result as done in test_one_dof
+    # But the whole raison-d'etre of the implicit equations is to deal with problems such
+    # as this one where the explicit form is too complicated to handle, especially the angular part
+    # (i.e. tests would be too slow)
+    # Instead, we check that the kinematic equations are correct using more fundamental tests:
+    #
+    # (1) that we recover the kinematic equations we have provided
+    assert (KM.mass_matrix_kin_implicit * KM.q.diff() - KM.forcing_kin_implicit) == Matrix(kinematic_eqs)
```
TJStienstra, line 475:
```suggestion
assert (KM.mass_matrix_kin * KM.q.diff() - KM.forcing_kin) == Matrix(kinematic_eqs)
```

sympy/physics/mechanics/tests/test_kane.py
```diff
@@ -348,3 +354,147 @@ def test_input_format():
...
+
+    # (2) that rate of quaternions matches what 'textbook' solutions give
+    # Note that we just use the explicit kinematics for the linear velocities
+    # as they are not as complicated as the angular ones
+    qdot_candidate = KM.forcing_kin
```
TJStienstra, line 480:
```suggestion
    KM.explicit = True
    qdot_candidate = KM.forcing_kin
```

sympy/physics/mechanics/tests/test_kane.py
```diff
@@ -348,3 +354,147 @@ def test_input_format():
...
+    # due to the number of operations involved
+    qdot_candidate[-1] = quat_dot_textbook[0] # lambda_0, note the [-1] as sympy's Kane puts the dependent coordinate last
+    qdot_candidate[0]  = quat_dot_textbook[1] # lambda_1
+    qdot_candidate[1]  = quat_dot_textbook[2] # lambda_2
+    qdot_candidate[2]  = quat_dot_textbook[3] # lambda_3
+
+    # sub the config constraint in the candidate solution and compare to the implicit rhs
+    lambda_0_sol = solve(config_cons[0], q_att_vec[0])[1]
+    lhs_candidate = simplify(KM.mass_matrix_kin_implicit * qdot_candidate).subs({q_att_vec[0]: lambda_0_sol})
+    assert lhs_candidate == KM.forcing_kin_implicit
```
TJStienstra, line 500:
```suggestion
    KM.explicit = False
    lhs_candidate = simplify(KM.mass_matrix_kin * qdot_candidate).subs({q_att_vec[0]: lambda_0_sol})
    assert lhs_candidate == KM.forcing_kin
```

tjstienstra (commented):
I think I've added all change suggestion. Only in the `__init__` you still have to add `self.explicit = True` as last line.

zouhairm:
The suggestion sounds reasonable to me (In fact that was the initial [implementation](https://github.com/sympy/sympy/pull/22818#discussion_r836788688) but we moved away from it)

--
Thanks for taking the time to type up the suggestions, unfortunately, github is not allowing me to commit them (due to this issue https://github.com/isaacs/github/issues/1755)

I can manually type them up, but if you happen to have this in branch already, can you either submit a PR to my fork, or perhaps post a patch file?

TJStienstra:
Almost have a patch file if you'd like

TJStienstra:
[explicit_attribute.zip](https://github.com/sympy/sympy/files/9677135/explicit_attribute.zip)

TJStienstra:
Using `explicit_kinematics` as attribute is indeed more clear.

sympy/physics/mechanics/tests/test_kane.py
```diff
@@ -74,8 +78,31 @@ def test_two_dof():
...
+            assert KM_matrices['explicit'][k] == KM_matrices['implicit'][k]
+        else:
+            # but matrix/vector involving kinematics are expected to be different
+            assert KM_matrices['explicit'][k] != KM_matrices['implicit'][k]
+
+    # Check that whether using implicit or explicit kinematics the RHS equations are consisten with the matrix form
+    for explicit_kinematics in [False, True]:
+        KM.explicit_kinematics = explicit_kinematics
+        assert simplify(KM.rhs() -
+                        KM.mass_matrix_full.LUsolve(KM.forcing_full)) == zeros(4, 1)
```
TJStienstra, line 105: I would personally not usethese kind of test loops, since they are not nice to debug. Rather write them out, since you don't even use less lines in the end. Besides that I would also suggest to directly check the implicit kinematic mass matrix and forcing vector (they are really simple), i.e. `KM.mass_matrix_kin == Matrix([[1/2, 0], [0, 2]])` (is also bit more useful and accurate)

P.S. Try to keep the lines within 80 characters

moorepants:
> The suggestion sounds reasonable to me (In fact that was the initial https://github.com/sympy/sympy/pull/22818#discussion_r836788688 but we moved away from it)

Sorry to run you in circles. That comment was still tied up with the 2 parts, so maybe not as relevant now. I also made a comment further down about the flag belonging in `.kanes_equations()`. Having a flag and no new attributes does seem cleaner at this point.

sympy/physics/mechanics/tests/test_kane.py
moorepants: Let's just try to get this in. We've nitpicked @zouhairm enough :)

zouhairm:
Not sure why `test/sphinx` is failing - seems to be unrelated to my changes as it's failing at the installation step:
```
Err:38 http://azure.archive.ubuntu.com/ubuntu focal-updates/main amd64 libgs9-common all 9.50~dfsg-5ubuntu4.5
  404  Not Found [IP: 52.252.75.106 80]
Err:42 http://azure.archive.ubuntu.com/ubuntu focal-updates/main amd64 libgs9 amd64 9.50~dfsg-5ubuntu4.5
  404  Not Found [IP: 52.252.75.106 80]
Err:43 http://azure.archive.ubuntu.com/ubuntu focal-updates/main amd64 ghostscript amd64 9.50~dfsg-5ubuntu4.5
  404  Not Found [IP: 52.252.75.106 80]
Fetched 665 MB in 15s (43.5 MB/s)
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/g/ghostscript/libgs9-common_9.50~dfsg-5ubuntu4.5_all.deb  404  Not Found [IP: 52.252.75.106 80]
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/g/ghostscript/libgs9_9.50~dfsg-5ubuntu4.5_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/g/ghostscript/ghostscript_9.50~dfsg-5ubuntu4.5_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Error: Process completed with exit code 100.
```

TJStienstra:
The problem should be fixed in the master now

sympy/physics/mechanics/kane.py
```diff
@@ -117,7 +136,7 @@ class KanesMethod(_Methods):
     def __init__(self, frame, q_ind, u_ind, kd_eqs=None, q_dependent=None,
             configuration_constraints=None, u_dependent=None,
             velocity_constraints=None, acceleration_constraints=None,
-            u_auxiliary=None, bodies=None, forcelist=None):
+            u_auxiliary=None, bodies=None, forcelist=None, explicit_kinematics=True):
```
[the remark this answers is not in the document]
zouhairm: the way we've implemented it know, explicit_kinematics can be "hot-swapped" as both the implicit and explicit forms are available.
This is because we are only doing the "2nd" part of the original PR of providing the mass/forcing matrices for kinematics in implicit form, i.e. this has no impact on how the dynamics (Kane's) equations themselves are constructed.

I've dropped it from the initializer for now, and we can later add it to the `kanes_equations()` call when the "1" part of the original PR is eventually implemented (i.e. where it can no longer be hotswapped as it would now have an impact on how the dynamics are computed)
I remember the argument for putting it in `kanes_equations()`, which makes sense regarding the "1st" part of the original PR, but since kinematic equations are available before  `kanes_equations()` is invoked, I think it still makes sense to set it in the initial constructor. Perhaps in the future we can decide that `kanes_equations()` can also take it as an argument to overwrite the original value if so desired.


For now I'm tempted to make it a 'private' member so as to signal that it should not be changed mid-run (even though that'd the only way to do it right now if we remove it from the initializer) and to remove it from the initializer and kick the can down the road to figure this out based on how the "1st" part of the original PR is ultimately implemented.

Thoughts?

sympy/physics/mechanics/tests/test_kane.py
zouhairm: - the for loop was on purpose, not to save lines of code, but to make it clear that the only thing changing between the two tests is the explicit flag. pytest should print out the variables when something fails, so hopefully shouldn't be too difficult to debug.

-  I've removed one of them anyway by implementing a more, excuse the pun, explicit matrix check as you suggested as opposed to just checking that they are different

- sorry about the 80 chars: I know it's a PEP8 thing but I didn't see it enforced by code_quality so I assumed the codebase uses a different value (we certainly do in ours as 80 is just too small for most modern monitors). If you feel strongly about it lmk and I'll clean it up.

TJStienstra: I do agree with the first point, was just a bit surprised to see this kind of code in tests. Furthermore, thanks for taking the time to update. As for the last point, the 80 character line width is in SymPy as far as I know not a strict rule, but personally I like consistency in formatting (even though I personally prefer 120 chars).

sympy/physics/mechanics/kane.py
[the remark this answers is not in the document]
TJStienstra: Making it a private member is indeed more future safe. As for where we can set the argument, I'm with either of them. `kanes_equations` is probably the most logical location from a users perspective, but `__init__` is the location where we actually need it. So I would prefere making it a private member and putting it in either the `__init__` or `kanes_equations`. Biggest disadvantage I see here is that it is an extra argument, especially in the `__init__` case.

@moorepants what is your opinion?

sympy/physics/mechanics/tests/test_kane.py
```diff
@@ -74,8 +78,20 @@ def test_two_dof():
...
 
-    assert simplify(KM.rhs() -
-                    KM.mass_matrix_full.LUsolve(KM.forcing_full)) == zeros(4, 1)
+    # Check that the explicit form is the default and kinematic mass matrix is identity
+    assert KM.explicit_kinematics
+    assert KM.mass_matrix_kin == eye(2)
+
+    # Check that for the implicit form the mass matrix is not identity
+    KM.explicit_kinematics = False
+    assert KM.mass_matrix_kin == Matrix([[1/2, 0], [0, 2]])
```
TJStienstra, line 87: You probably have to make the `1/2` a sympy object, i.e. `S(1)/2`

sympy/physics/mechanics/kane.py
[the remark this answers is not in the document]
moorepants: I don't have any issues with adding new arguments to `__init__`. The KanesMethod class does alot already, of course, but it seems in line to let it also return the explicit form of the kinematic equations.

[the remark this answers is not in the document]
zouhairm: So is the PR fine as is? If so I'll leave to one of you to please merge.
Otherwise please lmk what is left for us to do.

tjstienstra (approved):

sympy/physics/mechanics/kane.py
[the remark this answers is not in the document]
TJStienstra: @moorepants Based on your comments I assumed that you do agree with the current implementation, i.e. `explicit_kinematics` as attribute and as kwarg in the `__init__`, which is also fine with me.
Since I have no more comments on the PR (all suggestions have been taken into account), can this PR be merged?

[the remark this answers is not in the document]
zouhairm: bump - any chance you can merge please @moorepants ?
Thanks :)

moorepants:
Merging! Thanks and sorry that we took so long on this.

PR #23582, opened 2022-06-05
[WIP] Optimize evaluation of float arguments (version 2)

eendebakpt:

See the discussion at #23565.

Benchmark master:

```
from sympy import cos, atan
from random import random

%timeit atan(random())
542 µs ± 9.09 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)


%timeit cos(random())
393 µs ± 4.35 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
Benchmark PR
```
%timeit atan(random())
89.2 µs ± 1.5 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%timeit cos(random())
90.4 µs ± 1.8 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

Optimize evaluation of float arguments 

#### Release Notes

* functions
   * Optimize evaluation of float arguments. A new class attribute `_skip_eval_float_arguments` is added to the `Function` class. If set to `True`, then on numeric evaluation of the class the `eval` is skipped.

### DIFF
--- a/sympy/codegen/cfunctions.py
+++ b/sympy/codegen/cfunctions.py
@@ -192,7 +192,7 @@ class exp2(Function):
     log2
     """
     nargs = 1
-
+    _skip_eval_float_arguments = False
 
     def fdiff(self, argindex=1):
         """
--- a/sympy/combinatorics/schur_number.py
+++ b/sympy/combinatorics/schur_number.py
@@ -35,6 +35,7 @@ class SchurNumber(Function):
     536
 
     """
+    _skip_eval_float_arguments = False
 
     @classmethod
     def eval(cls, k):
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -512,8 +512,6 @@ class Function(Application, Expr):
         evaluated_result = super().__new__(cls, *args, evaluate=run_evaluation, **options)
 
         if evalf_precision >= 0:
-             if getattr(Function, 'debug_print', False):
-                        print(f'{cls}: {args} _skip_eval_float_arguments {cls._skip_eval_float_arguments} run_evaluation {run_evaluation} evalf_precision {evalf_precision}')
              if isinstance(evaluated_result, cls):
                  evaluated_result = evaluated_result.evalf(prec_to_dps(evalf_precision))
 
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -455,7 +455,7 @@ class Function(Application, Expr):
     """
 
     _skip_eval_float_arguments = True
-    
+
     @property
     def _diff_wrt(self):
         return False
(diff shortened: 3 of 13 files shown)


oscarbenjamin:
I really want to see some proper benchmarks for things like this. The benchmark suite needs to be improved. It currently doesn't handle caching properly so I don't know if it can meaningfully measure the effect of something like this.

asmeurer:
See my comment here https://github.com/sympy/sympy/pull/23565#issuecomment-1147978591. I am currently -1 on this because:

- I don't see the need for a new flag. `_should_evalf` should suffice. Also `_should_evalf` currently defaults to evalfing on Float but it could be redefined to evalf anything, so naming the flag in this way is misleading. 
- Calling `_should_evalf` (and `evalf`) before `eval` is a behavior change that could affect custom functions.
- This seems to optimize the case of floating-point arguments over the case of symbolic arguments. But symbolic arguments are far more common, so they should be the case that is more optimized. 
- The current logic doesn't require any code to redefine `_should_evalf`. With this, the flag has to be added to several functions. This suggests that the current behavior is a better default, and furthermore, that making a change like this would likely break any user code that is similar.

eendebakpt:
@asmeurer I agree with your observation that sympy should optimize for symbolic arguments and I share your concerns about changing behavior.

To give a bit of context: this PR was started because there are some methods with an `eval` that is really expensive. In #23541 I added a fast path for floats and @oscarbenjamin suggested an alternative optimization that is generic and not (as in # 23541) specific for one particular function. That suggestion was worked out to this PR.

* The changing behaviour is the main concern here. For this reason the attribute `_skip_eval_for_float` was added so we can disable the new behaviour for functions that rely on `eval` (in this PR this was required for about 10 functions to make all tests pass). Or, we can set `_skip_eval_for_float=False` by default and only enable it for functions we have checked to be safe.

* The check `_should_evalf` is really fast compared to the `eval`. So we can gain a lot on the evaluation of floats, while having more or less the same performance on symbolic expressions (in particular because the function evaluations are cached). There are already some benchmarks in #23565 and some open PRs to improve performance independent of this PR. 
I would be happy to perform more benchmarking, but the main point is the change of behaviour.

@asmeurer @oscarbenjamin Based on your comments I gather that having `_skip_eval_for_float=True` is not an option due to the change in behaviour. 
We can still add the mechanism with `_skip_eval_for_float` and allow this for specific functions. This would allow #23541 to remove the special `imaginary_unit_as_coefficient` method. It that an option? If not, I will close the PR and continue with #23541.

oscarbenjamin:
> I gather that having `_skip_eval_for_float=True` is not an option due to the change in behaviour.

We have to consider the fact that downstream code somewhere that subclasses Function probably depends on the current behaviour and we should try to preserve compatibility for that.

PR #23583, opened 2022-06-05
Fix integrate 1/sqrt(x**2-1) for manual

eagleoflqj:

#23566 manual part.
The `acosh` is replaced with a ~~`Piecewise`~~ `log`
#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/integrals/manualintegrate.py
+++ b/sympy/integrals/manualintegrate.py
@@ -38,7 +38,7 @@ from sympy.core.symbol import Dummy, Symbol, Wild
 from sympy.functions.elementary.complexes import Abs
 from sympy.functions.elementary.exponential import exp, log
 from sympy.functions.elementary.hyperbolic import (HyperbolicFunction, csch,
-    cosh, coth, sech, sinh, tanh, acosh, asinh, acoth, atanh)
+    cosh, coth, sech, sinh, tanh, asinh, acoth, atanh)
 from sympy.functions.elementary.miscellaneous import sqrt
 from sympy.functions.elementary.piecewise import Piecewise
 from sympy.functions.elementary.trigonometric import (TrigonometricFunction,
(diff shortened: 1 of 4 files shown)


oscarbenjamin:
Maybe this is undesirable:
```python
In [16]: manualintegrate(1/sqrt((x-I)**2 - 1), x)
Out[16]: 
⎧ acosh(x - ⅈ)   for x - ⅈ ≥ 1 
⎪                              
⎨ⅈ⋅asinh(x - ⅈ)  for x - ⅈ > -1
⎪                              
⎩-acosh(-x + ⅈ)    otherwise
```

eagleoflqj:
I understand that `integrate` should support $\int f(x)\text{d}x$ when `f(x)` is complex. Is it also supposed to support $\int f(x)\text{d}x$ when `x` is complex? I didn't find documentation for a clear scope of it.

sympy/integrals/tests/test_failing_integrals.py
```diff
@@ -287,3 +287,9 @@ def test_issue_4311_slow():
 def test_issue_20370():
     a = symbols('a', positive=True)
     assert integrate((1 + a * cos(x))**-1, (x, 0, 2 * pi)) == (2 * pi / sqrt(1 - a**2))
+
+
+@XFAIL
+def test_pr_23583():
+    # meijerg gives wrong Piecewise with Abs result
+    assert integrate(1/sqrt((x - I)**2-1)) == log(2*x + 2*sqrt(x**2 - 2*I*x - 2) - 2*I)
```
oscarbenjamin, line 295: I find XFAIL tests like this difficult because the correct result is not necessarily going to be in that form. We don't know that fixing the bug would make this pass.

This is the kind of situation where I would rather have a test that fails as soon as the output of integrate is changed in any way for this case. In other words I would rather have a test that asserts that the output is what it currently is but that is clearly commented to say that the output is wrong and the test should be changed as soon as integrate is fixed. Then if a change causes the test to fail someone can look into it to see if the bug is fixed.

Also if the wrong output currently comes from meijerg it would be good to have separate tests for `integrate` and for `integrate` specifically with `meijerg`.

oscarbenjamin:
Looks good. Thanks!

Issue #23487, opened 2022-05-11
Wrong nargs for Function with *args


asmeurer:
```py
>>> class f(Function):
...     @classmethod
...     def eval(cls, x, *args):
...         return None
>>> f.nargs
Naturals0
>>> class f(Function):
...     @classmethod
...     def eval(cls, x, y, z, *args):
...         return None
>>> f.nargs
Naturals0
```

The answers should be `Naturals` and `Range(3, oo)`. 

It looks like there is also some code that assumes the only possible infinite `nargs` is `Naturals0`, so that will need to be updated.

PR #23586, opened 2022-06-06
Cache FiniteSet for function calls

eendebakpt:

**Update**: since #23593 is merged, the analysis below is not valid any more.

The evaluation of a `Function` can be improved by avoiding the repeated construction of a `FiniteSet` object for simple arguments. Profiling shows this is a significant part of the evaluation

function_finiteset_overhead

(profiling results in the context of #23582)

Benchmark results: **main**
```
In [5]: %timeit re(random())
36.1 µs ± 343 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```
Benchmark results: **this PR**
```
15.3 µs ± 127 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

#### Release Notes

* functions
  * Improve performance of function evaluation by caching the construction of `nargs`

### DIFF
--- a/sympy/core/cache.py
+++ b/sympy/core/cache.py
@@ -149,12 +149,15 @@ else:
     raise RuntimeError(
         'unrecognized value for SYMPY_USE_CACHE: %s' % USE_CACHE)
 
+_cached_property_sentinel = object()
+
+
 def cached_property(func):
     '''Decorator to cache property method'''
     attrname = '__' + func.__name__
     def propfunc(self):
-        val = getattr(self, attrname, None)
-        if val is None:
+        val = getattr(self, attrname, _cached_property_sentinel)
+        if val is _cached_property_sentinel:
             val = func(self)
             setattr(self, attrname, val)
         return val
--- a/sympy/core/cache.py
+++ b/sympy/core/cache.py
@@ -151,7 +151,7 @@ else:
 
 def cached_property(func):
     '''Decorator to cache property method'''
-    attrname = '_' + func.__name__
+    attrname = '__' + func.__name__
     def propfunc(self):
         val = getattr(self, attrname, None)
         if val is None:
--- a/sympy/core/cache.py
+++ b/sympy/core/cache.py
@@ -148,3 +148,14 @@ elif USE_CACHE == 'debug':
 else:
     raise RuntimeError(
         'unrecognized value for SYMPY_USE_CACHE: %s' % USE_CACHE)
+
+def cached_property(func):
+    '''Decorator to cache property method'''
+    attrname = '_' + func.__name__
+    def propfunc(self):
+        val = getattr(self, attrname, None)
+        if val is None:
+            val = func(self)
+            setattr(self, attrname, val)
+        return val
+    return property(propfunc)
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -36,7 +36,7 @@ from collections.abc import Iterable
 from .add import Add
 from .assumptions import ManagedProperties
 from .basic import Basic, _atomic
-from .cache import cacheit
+from .cache import cacheit, cached_property
 from .containers import Tuple, Dict
 from .decorators import _sympifyit
 from .evalf import pure_complex
@@ -223,7 +223,7 @@ class FunctionClass(ManagedProperties):
         # to return the right value
         return lambda rule, **_: rule.get(self, self)
 
-    @property
+    @cached_property
     def nargs(self):
         """Return a set of the allowed number of arguments for the function.
 
@@ -2016,7 +2016,7 @@ class Lambda(Expr):
                 yield args
         return tuple(_variables(self.signature))
 
-    @property
+    @cached_property
     def nargs(self):
         from sympy.sets.sets import FiniteSet
         return FiniteSet(len(self.signature))
--- a/sympy/core/tests/test_cache.py
+++ b/sympy/core/tests/test_cache.py
@@ -1,4 +1,4 @@
-from sympy.core.cache import cacheit
+from sympy.core.cache import cacheit, cached_property
 from sympy.testing.pytest import raises
 
 def test_cacheit_doc():
@@ -54,3 +54,23 @@ def test_cachit_exception():
     # Unhashable type
     raises(TypeError, lambda: testf2([]))
     assert len(a) == 1
+
+def test_cached_property():
+    class A:
+        def __init__(self, value):
+            self.value = value
+            self.calls = 0
+
+        @cached_property
+        def prop(self):
+            self.calls = self.calls + 1
+            return self.value
+
+    a = A(2)
+    assert a.calls == 0
+    assert a.prop == 2
+    assert a.calls == 1
+    assert a.prop == 2
+    assert a.calls == 1
+    b = A(None)
+    assert b.prop == None
--- a/sympy/solvers/ode/single.py
+++ b/sympy/solvers/ode/single.py
@@ -9,6 +9,7 @@ from typing import Dict as tDict, Type, Iterator, List, Optional
 
 from .riccati import match_riccati, solve_riccati
 from sympy.core import Add, S, Pow, Rational
+from sympy.core.cache import cached_property
 from sympy.core.exprtools import factor_terms
 from sympy.core.expr import Expr
 from sympy.core.function import AppliedUndef, Derivative, diff, Function, expand, Subs, _mexpand
@@ -38,18 +39,6 @@ from .lie_group import _ode_lie_group
 class ODEMatchError(NotImplementedError):
     """Raised if a SingleODESolver is asked to solve an ODE it does not match"""
     pass
-
-
-def cached_property(func):
-    '''Decorator to cache property method'''
-    attrname = '_' + func.__name__
-    def propfunc(self):
-        val = getattr(self, attrname, None)
-        if val is None:
-            val = func(self)
-            setattr(self, attrname, val)
-        return val
-    return property(propfunc)
 
 
 class SingleODEProblem:


oscarbenjamin:
Maybe `nargs` can just be a property so that the FiniteSet is only created when accessed. Then we can skip using it altogether in the ordinary case and use a regular Python set or just integers representing the min and max number of allowed args.

asmeurer:
For the default `nargs` that's created from the `eval` signature, I don't see why we need to check it at all. We can just call eval and if the number of arguments is wrong Python will raise a TypeError. However if a subclass defines its own `nargs` we can't do this because it might be more restrictive than the `eval` signature, and rely on the pre-filtering.

asmeurer:
I'm curious just how effective this is. Looking at the function.py code, `nargs` is used in quite a few places, so I imagine it wouldn't take long for it to be evaluated.

sympy/core/function.py
```diff
@@ -320,7 +324,7 @@ def __new__(cls, *args, **options):
             #  - AppliedUndef with no nargs like Function('f')(1).nargs
             nargs = obj._nargs  # note the underscore here
         # convert to FiniteSet
-        obj.nargs = FiniteSet(*nargs) if nargs else Naturals0()
+        obj.nargs = _FiniteSetCache(*nargs) if nargs else S.Naturals0
```
asmeurer, line 327: Is there a reason this has to be set in the constructor? Otherwise we could just add @cacheit to the nargs property definition in FunctionClass.

eendebakpt: It was there, so I did not move it

eendebakpt:
@asmeurer @oscarbenjamin I do not know exactly where nargs is used, but it is used in the `Function.__new__`, which is a performance issue. This PR solves this by caching the `FiniteSet`. 

In PR #23593 the problem in `__new__` is solved by having a direct method (which is faster than this, but does not address other invocations of `nargs`)

The best solution would be to replace `nargs` with a `tuple` (suggestion by oscar in #23593), but this changes the public API. This patch:
```
diff --git a/sympy/core/function.py b/sympy/core/function.py
index 710c5cfc74..f4115a6f0d 100644
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -256,10 +256,9 @@ def nargs(self):
         >>> len(f(1).args)
         1
         """
-        from sympy.sets.sets import FiniteSet
         # XXX it would be nice to handle this in __init__ but there are import
         # problems with trying to import FiniteSet there
-        return FiniteSet(*self._nargs) if self._nargs else S.Naturals0
+        return set(self._nargs) if self._nargs else S.Naturals0
 
     def __repr__(cls):
         return cls.__name__
```
passes most of the tests. The few tests failing (I suspect) are easy to fix.

If changing the type if `nargs` is an option, then I can make a PR and investigate the failing tests.

asmeurer:
I don't think we need to change the public API of nargs to get better performance here. The default nargs is just based on the eval signature, so we can optimize for that case internally, leaving the relatively slower general case for the instances where a class defines its own `nargs` (which is rare). The internals can be optimized in such a way that it doesn't affect the public `nargs` attribute, and only applies to subclasses that don't redefine `nargs`.

Additionally, if FiniteSet specifically is slow, we should determine if it can be improved directly, rather than working around it. For things like `1 in FiniteSet(1)`, that seems to be related to https://github.com/sympy/sympy/pull/22184, which I thought was already fixed (CC @ThePauliPrinciple), but maybe there is still more that can be done there? Ideally a `FiniteSet` should have comparable performance to a Python `set`.

oscarbenjamin:
> Ideally a `FiniteSet` should have comparable performance to a Python `set`.

In the case of `n in S` where `n` is in `S` it should be possible to get a lot closer e.g.:
```python
In [1]: S = FiniteSet(1)

In [2]: %timeit 1 in S
2.81 µs ± 47.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [3]: %timeit 1 in S._args_set
264 ns ± 2.85 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: Ss = {1}

In [5]: %timeit 1 in Ss
46.3 ns ± 0.194 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

asmeurer:
Ah, so it looks like `Set.__contains__` calls `_sympify(other)`.

https://github.com/sympy/sympy/blob/8c180043a9150de62a6b6a6f12ced1c7bacf15a3/sympy/sets/sets.py#L789

Otherwise it would be completely identical to `1 in S._args_set` (plus some additional function calls). 

https://github.com/sympy/sympy/blob/8c180043a9150de62a6b6a6f12ced1c7bacf15a3/sympy/sets/sets.py#L2019-L2020

The sympify call should be unnecessary if it's just comparing since `__eq__` already does that automatically, but in a way that smartly avoids it. But I'm not sure if removing it would break something else. 

The `not in` case is another story, since it is using the same nasty auto-evaluation that happens in `Eq`. But the `in` case definitely matters more for optimization, as the `not in` case is an error anyway.

eendebakpt:
@asmeurer Timings on my system:
```
In [30]: f=FiniteSet(1)
    ...: %timeit sympify(1)
    ...: %timeit 1 in f
    ...: %timeit sympify(1) in f
    ...: %timeit 1 in f._args_set
836 ns ± 19.7 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
1.48 µs ± 13.2 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
1.78 µs ± 11.5 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
189 ns ± 8.97 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

Saving > 1us from the 1.48 us that it currently takes (e.g. `1 in f`) seems worthwhile. We have several options:

* #23593 (which avoids the FiniteSet altogether)
* Using `1 in f._args_set` (perhaps make this public, e.g. `FiniteSet.contains_no_sympify(1)`?)
* Modyfing `FiniteSet` to not automatically `sympify` the arguments (as you note: this might break things)
* Making `sympify(1)` faster.

I have a preference for one of the first two options.

oscarbenjamin:
My preference is definitely to avoid FiniteSet altogether. Something low-level like this should not depend on using any symbolic objects. I'm not sure what the best approach in the absence of FiniteSet is though.

eendebakpt:
> It actually looks like there are some bugs with nargs anyway. Aside from #23487 which I mentioned before, it looks like you can't actually define nargs to be a set. This is the sort of thing I would expect to break things if we do some of the suggested refactors here.
> 
> ```python
> >>> class f(Function):
> ...     @classmethod
> ...     def eval(self, x1, x2, y1=None, y2=None):
> ...         return
> ...     nargs = FiniteSet(2, 4) # If y1 is passed require y2 as well
>   File "/Users/aaronmeurer/Documents/Python/sympy/sympy/./sympy/utilities/misc.py", line 545, in as_int
>     return operator.index(n)
> TypeError: 'FiniteSet' object cannot be interpreted as an integer
> 
> During handling of the above exception, another exception occurred:
> 
>   File "<stdin>", line 1, in <module>
>   File "/Users/aaronmeurer/Documents/Python/sympy/sympy/./sympy/core/function.py", line 185, in __init__
>     nargs = (as_int(nargs),)
>   File "/Users/aaronmeurer/Documents/Python/sympy/sympy/./sympy/utilities/misc.py", line 547, in as_int
>     raise ValueError('%s is not an integer' % (n,))
> ValueError: {2, 4} is not an integer
> ```
> 
> I'm not really sure I see the point of using `nargs` like this anyway. The default of computing it from the `eval` signature is sufficient for 99% of cases. If you want more filtering than that, you can do it in `eval` itself. I wouldn't mind deprecating the automatic nargs filtering that happens in `FunctionClass.__call__` in favor of manual filtering in `eval`.
> 
> Other than in `__call__`, it looks like it's also used in the class key, which we could probably replace with something equivalent that is cheaper to compute, and it's used in `_eval_nseries`. That use is probably legitimate, but it's not something that gets computed at function creation time so I'm not too worried about it.
> 
> The main value of it that I see is for undefined functions, like `Function('f', nargs=2)` (which also doesn't support sets).

@asmeurer I cannot follow all the arguments here. But either this PR or #23593 solves the current performance bottleneck in the `nargs` lookup in the `Function` without changing the public API. Can we consider either (or both) PRs and move discussion about a possible replacement for the current nargs somewhere else?

eendebakpt:
@oscarbenjamin @asmeurer Apart from moving the `cached_property` code to a common location, this PR is two lines: adding the `cached_property` to `nargs`. It helps to cache `nargs`, since a `FiniteSet` is constructed which takes time.

But with #23593 merged, I am not sure `nargs` is called often. In any case not in the evaluation of a `Function`.

sympy/core/cache.py
```diff
@@ -148,3 +148,14 @@ def _getenv(key, default=None):
 else:
     raise RuntimeError(
         'unrecognized value for SYMPY_USE_CACHE: %s' % USE_CACHE)
+
+def cached_property(func):
```
oscarbenjamin, line 152: There should be two blank lines between the top-level functions.

oscarbenjamin:
If we want to make `cached_property` a proper reusable function it would be good to have some tests.

eendebakpt:
@oscarbenjamin Let me know whether you want to merge the PR or not. If not, I will remove the optimizations and just keep the unit tests for the `cached_property` decorator.

sympy/core/cache.py
```diff
@@ -148,3 +148,17 @@ def _getenv(key, default=None):
 else:
     raise RuntimeError(
         'unrecognized value for SYMPY_USE_CACHE: %s' % USE_CACHE)
+
+_cached_property_sentinel = object()
+
+
+def cached_property(func):
```
oscarbenjamin, line 155: It's probably a little tidier to put the sentinel inside the `cached_property` function.

eendebakpt:
> > Let me know whether you want to merge the PR or not
> 
> Right now it can't be merged in any case because the tests failed.
> 
> Does the `cached_property` decorator here make a noticeable difference to timings? How does the profile plot you computed above look now?

The decorator makes a large difference for the `nargs` property, but since this one is not used any more in the object constructor the property is not called often. The tests are failing, because something else has changed. I will close this PR and open a new one just for the tests.

Note: for some of the failing tests it is not clear to be that these are valid tests. For example `test_nargs_inheritance` contains the code
```
...
    class f4(f3):
        nargs = 1,2
...
```
The property `nargs` is supposed to be a `FiniteSet`, so assigning a tuple to it seems wrong.

oscarbenjamin:
> The property `nargs` is supposed to be a `FiniteSet`, so assigning a tuple to it seems wrong.

It's hard to say what's right or wrong with these things. There is code to handle a sequence here and that code has been around a long time so downstream code might be doing all sorts of things with it.

PR #23588, opened 2022-06-06
Refactor the ordering function to prevent unnecessary dict access

eendebakpt:

Refactor the `ordering` function to prevent unnecessary `dict` access

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/sorting.py
+++ b/sympy/core/sorting.py
@@ -282,10 +282,12 @@ def ordered(seq, keys=None, default=True, warn=False):
 
     d = defaultdict(list)
     if keys:
-        if not isinstance(keys, (list, tuple)):
-            keys = [keys]
-        keys = list(keys)
-        f = keys.pop(0)
+        if isinstance(keys, (list, tuple)):
+            keys = list(keys)
+            f = keys.pop(0)
+        else:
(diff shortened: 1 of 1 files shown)


sympy/core/sorting.py
```diff
@@ -283,9 +283,11 @@ def ordered(seq, keys=None, default=True, warn=False):
...
     if keys:
         if not isinstance(keys, (list, tuple)):
-            keys = [keys]
-        keys = list(keys)
-        f = keys.pop(0)
+            f = keys
+            keys = []
+        else:
+            keys = list(keys)
+            f = keys.pop(0)
```
eagleoflqj, line 290: Could you please swap the if and else, as there is no need for using `not`?

eendebakpt:
I think so as well. The only reason I kept it like this, is that the pop
can release some memory. I will update the pr with your suggestion

Op ma 6 jun. 2022 19:01 schreef Liumeo ***@***.***>:

eendebakpt:
@eagleoflqj Thanks for the review, I addressed the comments.

eagleoflqj (approved):
LGTM. Squashing and merging.

PR #23590, opened 2022-06-06
[WIP] Add small integer lookup table for Function

eendebakpt:

An expensive part of a function evaluation is the check whether the given number of arguments is valid for the function. The check is performed as:
```
...
n=len(args)
if n not in cls.nargs:
   ....
```
This check is rather slow because: i) cls.nargs is a `FiniteSet` and ii) len(args) is an `int` on which `sympify` is called during the check. This PR improves the speed of the check by creating a fast lookup for the `sympify` of integer types.

#### Other comments

The `Function.nargs` is part of the public API, so we keep that intact. An alternative would be to create a second structure (e.g. a python set `._nargs_plain` to perform the check.

#### Release Notes

NO ENTRY

### DIFF
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -66,7 +66,7 @@ from collections import Counter
 _sympy_int_cache = {x: sympify(x) for x in range(10)}
 
 def _integer_sympify(x : int):
-    """ Fast sympiofy for small integers """
+    """ Fast sympify for small integers """
     value = _sympy_int_cache.get(x, None)
     if value is None:
         value = sympify(x)
--- a/sympy/core/function.py
+++ b/sympy/core/function.py
@@ -63,6 +63,16 @@ from mpmath.libmp.libmpf import prec_to_dps
 import inspect
 from collections import Counter
 
+_sympy_int_cache = {x: sympify(x) for x in range(10)}
+
+def _integer_sympify(x : int):
+    """ Fast sympiofy for small integers """
+    value = _sympy_int_cache.get(x, None)
+    if value is None:
+        value = sympify(x)
+    return value
+
+
 def _coeff_isneg(a):
     """Return True if the leading Number is negative.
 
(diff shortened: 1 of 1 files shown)


oscarbenjamin:
I would really like to have better benchmarks before thinking about changes like this.

asmeurer:
If this really is a bottleneck wouldn't it be better to improve `FiniteSet.__contains__` so that it doesn't sympify ints. Although based on some basic timings I'm not yet convinced it is. Also, doesn't sympify already have its own cache?

eendebakpt:
> If this really is a bottleneck wouldn't it be better to improve `FiniteSet.__contains__` so that it doesn't sympify ints. Although based on some basic timings I'm not yet convinced it is. Also, doesn't sympify already have its own cache?

@asmeurer The `FiniteSet` stores its elements as sympy objects:
```
>>> f=FiniteSet(2,3)
>>> print(type(list(f)[0]))
<class 'sympy.core.numbers.Integer'>
>>> print(2 in f)
True
```
So to check whether an `int` is contained in a `FiniteSet` it neems to be sympified. (or we should do some special checks inside the `FiniteSet` specially for integers, which I think is not a good approach)

Besides the benchmark on top of the PR, this is a basic timing:
``` 
In [8]: %timeit _integer_sympify(2) in f
   ...: %timeit 2 in f
937 ns ± 5.91 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
1.4 µs ± 7.96 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
The `sympify` does not cache the integers, which explains the difference here.

I do realize now that the `FunctionClass` already has a field `_nargs` which we can use, and this is much faster: 
```
%timeit 5 in cos._nargs
48.9 ns ± 0.445 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
I will close this PR and create a new one.

Issue #12034, opened 2017-01-10
Add contour plot function


chrstphrchvz:
(Mentioned on [gitter](https://gitter.im/sympy/sympy?at=58740606cbcb2817704aa5dd), no responses yet.)

I saw 2d plots mentioned on [this wiki page](https://github.com/sympy/sympy/wiki/Plotting-capabilities#types-of-plots), but I understand that might be outdated/inaccurate.

Looking through what `plot3d()` does, I used `Plot(ContourSeries(…)).show()` as a workaround.
It seems like much of the infrastructure is in place to support a contour plotting function, would there be any difficulty doing so?

asmeurer:
My guess is it wouldn't be too difficult to add it.

zaz:
Sorry if this is user error, but why can I not find `contour_plot` alongside `plot`, `plot3d` and other such functions?

    >>> sp.__version__
    '1.10.1'
    >>> sp.plotting.contour_plot
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
    AttributeError: module 'sympy.plotting' has no attribute 'contour_plot'

chrstphrchvz:
@zaz the function is named `plot_contour`, see https://github.com/sympy/sympy/pull/14896#issuecomment-405382223

oscarbenjamin:
Maybe the `plot_contour` function should be exposed at top-level like `from sympy import plot_contour`.

zaz:
Sorry for mixing that up, but that is not there either:

    >>> sp.__version__
    '1.10.1'
    >>> sp.plotting.plot_contour
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
    AttributeError: module 'sympy.plotting' has no attribute 'plot_contour'

And the only mention of "contour" in `help(sp.plotting)` is under `plot_implicit`:

> ...
> a generating a **contour** using a mesh grid of fixed number of points. By
> ...

Even though I can see `def plot_contour` and docstring in `~/.local/lib/python3.9/site-packages/sympy/plotting/plot.py`

oscarbenjamin:
It can be imported like this:
```python
In [6]: from sympy.plotting.plot import plot_contour

In [7]: plot_contour(sin(x)*sin(y), (x, -5, 5), (y, -5, 5))
Out[7]: <sympy.plotting.plot.Plot at 0x7fbc9c11d550>
```
It should probably be easier to import though (i.e. from sympy directly).

zaz:
[...]

PR #23591, opened 2022-06-06
enable `sympy.plotting.plot_contour`

zaz:
instead of `from sympy.plotting.plot import plot_contour`

Improve: issue #12034

* plotting
  * Enable `sympy.plotting.plot_contour`

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1371,6 +1371,7 @@ Zach Carmichael <20629897+craymichael@users.noreply.github.com> Zach <20629897+c
 Zach Raines <raineszm@gmail.com>
 Zachariah Etienne <zachetie@gmail.com>
 Zamrath Nizam <zamiguy_ni@yahoo.com> <zamiguy.ni@gmail.com>
+Zaz Brown <zazbrown@zazbrown.com>
 Zeel Shah <kshah215@gmail.com>
 Zhi-Qiang Zhou <zzq_890709@hotmail.com> zhouzq-thu <zzq_890709@hotmail.com>
 Zhongshi <zj495@nyu.edu>
--- a/sympy/plotting/__init__.py
+++ b/sympy/plotting/__init__.py
@@ -4,7 +4,7 @@ from .textplot import textplot
 from .pygletplot import PygletPlot
 from .plot import PlotGrid
 from .plot import (plot, plot_parametric, plot3d, plot3d_parametric_surface,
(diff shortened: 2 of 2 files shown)


oscarbenjamin:
I guess it should probably be imported at top-level like `from sympy import plot_contour`.

@Davide-sd what do you think?

zaz:
Why at the top level and not with the other plot functions under `sympy.plotting.`?

Is there something special about `plot_contour` vs `plot3d`?

What is the rationale behind putting some plot commands in the top level while others are under `plotting`?  Is this just an inconvenience that has developed out of the need for backwards-compatibility?

It seems most sensible to have:

    sympy.plot.plot()
              .3d()
              .contour()
              .
              .
              .

The way plotting commands are currently structured is just all around odd for me.

Davide-sd:
I think it should be available both at the top level, `from sympy import plot_contour`, as well as `from sympy.plotting import plot_contour`. The last one is justified from by the fact that I have seen a couple of questions at StackOverflow about people importing plot functions from `sympy.plotting`. As a matter of fact, reading [the documentation](https://docs.sympy.org/latest/modules/plotting.html) it is implied to use `from sympy.plotting import plot_function_name`.

zaz:
It would be nice to have consistency with this: Either expose all the plotting functions at the top level, or keep them all in `plotting`, IMO.  Personally, I think that changing the name of `plotting` to `plot` and removing `plot_` gives us the best of both worlds; the user can do:

    from sympy import plot
    plot.3d(...)
    plot.parametric()

Which has the same level of succinctness, but more versatility than importing from the top-level.  Better yet, it allows the user to just `import sympy as sp` and still have a succinct way to make various kinds of plots, obviating the need to import plot functions explicitly.

The current `sympy.plot()` could be either retained permanently or deprecated using `__call__`:

```
>>> import sympy as sp
>>> class CallableModule(sp.__class__):
...     def __call__(self, **args):
...         sp.plotting.plot(**args)
... 
>>> sp.plotting.__class__ = CallableModule
>>> sp.plot = sp.plotting
>>> sp.plot()
>>> sp.plot.parametric()
```

asmeurer:
This looks good. We should discuss the other changes in a new issue as they are more controversial since they would break compatibility.

zaz:
Thank you.  I created a new issue for that: #23599

PR #23592, opened 2022-06-07
[GSoC Pr1] Refactored leading term method for log Function

### LINKED ISSUE #6682: gruntz( exp(2*Ei(-z))/z**2, z, 0)
```
The following limit containing the expüonential integral fails:

gruntz( exp(2*Ei(-z))/z**2, z, 0)

The correct result is exp(2*EulerGamma).
```

Original issue for #6682: http://code.google.com/p/sympy/issues/detail?id=3583
Original author: https://code.google.com/u/107490137238222069432/
Original owner: https://code.google.com/u/113518207426301808764/

### LINKED ISSUE #21227: Nested logarithms add unnecessary order term to series expansions
What works - 
```py
In [1]: f = log(x)

In [2]: f.nseries(x, 0, logx=y)
Out[2]: y

In [3]: f.nseries(x, 0, logx=-x)
Out[3]: -x

```
What doesn't work - 
```py
In [4]: f = log(-log(x))

In [5]: f.nseries(x, 0, logx=y)
Out[5]: 
           ⎛ 6⎞
log(-y) + O⎝x ⎠

In [6]: f.nseries(x, 0, logx=-x)
Out[6]: 
      ⎛ 6⎞
-x + O⎝x ⎠

In [7]: f = log(log(x))

In [8]: f.nseries(x, 0, logx=y)
Out[8]: 
          ⎛ 6⎞
log(y) + O⎝x ⎠

In [9]: f.nseries(x, 0, logx=-x)
Out[9]: 
           ⎛ 6⎞
ⅈ⋅π - x + O⎝x ⎠

In [10]: f.nseries(x, 0, logx=x)
Out[10]: 
     ⎛ 6⎞
x + O⎝x ⎠

```
From the documentation of `nseries` - The optional ``logx`` parameter can be used to replace any log(x) in the returned series with a symbolic value to avoid evaluating log(x) at 0. A symbol to use in place of log(x) should be provided.

I would expect no Order term to be added in any of the expansions shown above.

### LINKED ISSUE #21721: Bug in integration solver

### LINKED ISSUE #22220: Bug in the evaluation of a log limit
image
The limit is evaluated differently at the discontinuity points. In one case (atan) right and left limits are different while in the other they are evaluated the same.

### LINKED ISSUE #23596: Integral of real function has complex result
A minimal example:
```
import sympy
from sympy import symbols, expand, simplify, integrate, S
t = symbols('t', positive=True)
pdf = ((1+t)/t**3)*sympy.exp(-1/t)
print(pdf)

E= integrate(pdf*t, (t, 0, 1e9))
print(E.evalf()) # seems fine

E= integrate(pdf*t, (t, 0, S.Infinity))
print(E.evalf()) # complex number!
``` 
has output
```
(t + 1)*exp(-1/t)/t**3
21.1460501720449
0.422784335098467 + 3.14159265358979*I
```

anutosh491:

Fixes #21721
Fixes #21227
Fixes #22220
Fixes #23596 
Fixes #6682 
This PR introduces a new algorithm/logic to compute leading terms of expressions of type `log` which has been broken down into 5 step explained below.
1) In leading term methods the parameter `x` is expected to tend to `0` along the half-ray `x = cdir*t` where `t` is `real` and `positive`. Hence we define a positive dummy variable `t` and substitute `cdir*t` for `x` in the argument of `log: z = arg0.subs(x, cdir*t)`. 
2) In the next step we calculate coefficient and exponent of the leading term of `z` through `c, e = z.leadterm(t)` to take decisions based on `c` and `e` for the incoming steps.
3) In this step we handle the standard case of  `c, e == 1, 0`
4) Next, we expand `log(c*t**e)` using `log(t) = log(x) - log(cdir)` or `logx - log(cdir)`.
5) Finally we append `-2*I*pi` if `c` is `negative` and the imaginary part of `z` is `negative` for small `t > 0`

#### Release Notes

* series
  * Refactored leading term method for log Function.
  * Added logx parameter to leading term methods of type add and power.
  * Introduced arg_flag in Limit.doit.set_signs.
  * Fixed leading term method for Ei, Chi and bessely functions.
  * Implemented branch cut handling for leading term methods of inverse trigonometric functions.
  * Fixed nseries for nested logarithms where logx parameter passed is not None.
  * Fixes integral for probability distribution function of cauchy distribution


anutosh491:
Note : Limits/Series which need to be fixed . (These involve simply replacing the functions with their `log` rewrites and testing whether the new code is able to answer all of these).

1) Problems at `I*oo/ -I*oo` . Some handling is missing for these cases
```
>>> limit(atan(x), x, -I*oo)
-pi/2
>>> limit(atan(x).rewrite(log), x, -I*oo)     # expected pi/2
-pi/2
>>>
>>> limit(asinh(x), x, I*oo)
-oo
>>> limit(asinh(x).rewrite(log), x, I*oo)
log(zoo + oo*I)
>>> asinh(x).rewrite(log).series(x, I*oo)
log(zoo + oo*I)
>>>
>>> log(1 + x).series(x, x0=I*oo)    
O(1, (x, Limit(x + oo*I, x, 0)))    
>>> # expected answer for above
1/(5*x**5) - 1/(4*x**4) + 1/(3*x**3) - 1/(2*x**2) + 1/x + I*pi/2 + log(-I*x) + O(x**(-6), (x, oo*I)))
>>>
```
2) Limits/Series which are hang infinitely
```
>>> limit(acos(x).rewrite(log), x, 1)
>>> limit(acosh(x).rewrite(log), x, -oo)
>>> limit(asech(x).rewrite(log), x, -1)
>>> limit(asech(x).rewrite(log), x, 0, '-')
```

anutosh491:
Also the second set of errors which hang infinitely do so because of `log._eval_nseries()` . For eg
```
>>> expr =  log(sqrt(1 - (x + 1)**2) + I*(x + 1))
>>> expr._eval_nseries(x, 0, None)
I*pi/2 + O(x**2)                # expected ans:  I*pi/2 + sqrt(2)*sqrt(x) + O(x**2)
```

sympy/functions/elementary/exponential.py
```diff
@@ -1009,15 +1009,22 @@ def coeff_exp(term, x):
...
         if p.has(exp):
             p = logcombine(p)
         if isinstance(p, Order):
             n = p.getn()
         _, d = coeff_exp(p, x)
         if not d.is_positive:
-            return log(a) + b*logx + Order(x**n, x)
+            res = log(a) + b*logx
+            if (res.subs(logx, log(x)) == self
+                or res.subs(-logx, -log(x)) == self):
```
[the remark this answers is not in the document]
anutosh491: Actually another test from `test1_latest` which fails , is due to this sort of handling but in `mul.py` actually lands up on this 
```
>>> expr = x**2*_eis(-2*log(1/x))/2
>>> expr._eval_nseries(x, n = 6, logx = -p)
x**2*_eis(-2*p)/2 + O(x**6)
```
But the order term shouldn't be there 
Now this is because I've added a simple case like 
```
        if res != self:
            if res.subs(logx, log(x)) == self:
                return res
```
But here `res  = x**2*_eis(-2*p)/2` and `logx = -p` so we get the substituted expression as `x**2*_eis(2*log(x))/2` which is correct but not in the form of self (uses log(1/x) so yeah here the case gets more weird , we would need to have something like 
```
         if res != self:
+            if (res.subs(logx, log(x)) == self
+                or res.subs(-logx, -log(x)) == self
+                or res.subs(-logx, log(1/x)) == self):
                         return res;
```
Then we get the correct result

jksuom:
> im(z) != 0

It looks like this test does not suffice guarantee that not all terms of the lseries be real.

sympy/functions/elementary/exponential.py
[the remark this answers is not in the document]
jksuom: We cannot possibly have substitutions for many different expressions like `log(1/x)`. Instead, we should probably first try to force `self` into a more canonical form by something like `expand_log` and then substitute for `logx` (or `-logx` if `logx` is negative).
```
In [11]: res = x**2*_eis(-2*p)/2  

In [12]: logx = -p

In [13]: res.subs(-logx, -log(x))
Out[13]: 
 2               
x ⋅_eis(2⋅log(x))
─────────────────
        2        

In [14]: expr = x**2*_eis(-2*log(1/x))/2

In [17]: expand_log(expr, force=True)
Out[17]: 
 2               
x ⋅_eis(2⋅log(x))
─────────────────
        2
```

jksuom:
The loop should be fixed by something like this:

        for i, term in enumerate(z.lseries(t)):
            if not term.is_real or i == 5:
                break
        if i < 5:
            coeff, _ = term.as_coeff_exponent(t)
            res += -2*I*S.Pi*Heaviside(-im(coeff), 0)

The test `im(z) != 0` can be improved but it is probably impossible to cover all cases, so limiting the loop count looks like the best solution.

anutosh491:
> > im(z) != 0
> 
> It looks like this test does not suffice to guarantee that not all terms of the lseries be real.

Yeah correct , wouldn't just using `im(z).is_negative` work here (I guess that's the main case we're trying to address here .. where we would have to subtract the `2*S.Pi*I` from `res`).  Not saying that's the permanent fix here but addresses the relevant problems here quickly 
```
>>> expr = Ei(-2*log(x))/2 - log(log(x))/(2*x**2)
>>> limit(expr , x, 1, '-')
EulerGamma/2 + log(2)/2 - I*pi/2
>>> limit(acosh(x).rewrite(log), x, -oo)
oo
>>> limit(asech(x).rewrite(log), x, -1)
I*pi
>>> limit(asech(x).rewrite(log), x, 0, '-')
oo
```

sympy/functions/elementary/exponential.py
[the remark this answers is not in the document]
anutosh491: Thanks for this, wasn't aware of `expand_log`, I was just about to commit using the following , then saw your message
``
            expr = expr.expand(**dict(deep=True, log=True, mul=False, power_exp=False,
                  power_base=False, multinomial=False, basic=False, force=False,
                  factor=False)
``

jksuom:
> wouldn't just using im(z).is_negative work here

That does not cover the case where the sign of `im(z)` is unknown, for example `z = -1 + a*t` where `a` is a symbol.

anutosh491:
Fixed a failing test 
```
>>> integrate(log(log(x)) / x**3, [x, 1, oo])
>>> # The above test from test_integrals.py would be calling the following
>>> limit(Ei(-2*log(x))/2 - log(log(x))/(2*x**2), x, oo)
NotImplementedError: Don't know how to calculate the mrv of 'O(_w, (_w, oo))'
```
And the workflow would involved leading term methods of `Ei` and `Chi` function , both of which were implemented wrongly ! For eg
```
>>> # On master
>>> limit(Ei(x), x, 0)
EulerGamma
>>> limit(Ei(x) - log(x), x, 0)
EulerGamma
>>> limit(Chi(x), x, 0)
EulerGamma
>>>
>>> # On branch
>>> limit(Ei(x), x, 0)
-oo
>>> limit(Ei(x) - log(x), x, 0)
EulerGamma
>>> limit(Chi(x), x, 0)
-oo
```
[Reference link](https://www.wolframalpha.com/input?i=limit+calculator&assumption=%7B%22F%22%2C+%22Limit%22%2C+%22limit%22%7D+-%3E%220%22&assumption=%7B%22F%22%2C+%22Limit%22%2C+%22limitfunction%22%7D+-%3E%22Ei%28x%29%22)

This also closes closely related issues (6682, 23596) for which I've added tests !

sympy/functions/special/error_functions.py
```diff
@@ -1212,10 +1212,15 @@ def _eval_rewrite_as_tractable(self, z, limitvar=None, **kwargs):
...
+        from sympy import re
         x0 = self.args[0].limit(x, 0)
+        arg = self.args[0].as_leading_term(x, cdir=cdir)
+        cdir = arg.dir(x, cdir)
         if x0.is_zero:
-            f = self._eval_rewrite_as_Si(*self.args)
-            return f._eval_as_leading_term(x, logx=logx, cdir=cdir)
+            c, e = arg.as_coeff_exponent(x)
+            logx = log(x) if logx is None else logx
+            return log(c) + e*log(x) + S.EulerGamma - (
```
jksuom, line 1222: Should there be logx instead of log(x)?

anutosh491: Yeah correct , will make the change in the upcoming commits today !

anutosh491:
In the latest commit I've fixed leading term methods at points lying on branch cuts for all inverse trigonometric functions . I've added couple of TODO's which are remaining from our side
```
    # TODO : acos(x + 1).as_leading_term(x) hangs infinitely
    # assert acos(x + 1).as_leading_term(x) == sqrt(2)*sqrt(x)

    # TODO : atan(x - I).as_leading_term(x, cdir = -1) should be equal to
    # I*(log(x) + I*pi/2)/2 but currently gives I*(log(x) - I*pi/2)/2
    # assert atan(x - I).as_leading_term(x, cdir=-1) == I*(log(x) + I*pi/2)/2
```
Graph being 
image
So leading term from right side is fine but left side I've added as TODO 

Some basic limits which weren't responsive enough now work
```
>>> limit(atan(x), x, I)  # now I*oo
Limit(atan(x), x, I)
>>> limit(acot(x), x, I)  # now  -I*oo
Limit(acot(x), x, I)
>>> limit(asec(x), x, 0)   # now I*oo
zoo  
>>> limit(acsc(x), x, 0, '+')  # now -I*oo
zoo
```

sympy/core/add.py
```diff
@@ -996,20 +996,20 @@ def _eval_as_leading_term(self, x, logx=None, cdir=0):
...
             old = piecewise_fold(old)
 
         # This expansion is the last part of expand_log. expand_log also calls
         # expand_mul with factor=True, which would be more expensive
         if any(isinstance(a, log) for a in self.args):
-            logflags = dict(deep=True, log=True, mul=False, power_exp=False,
-                power_base=False, multinomial=False, basic=False, force=False,
-                factor=False)
-            old = old.expand(**logflags)
+            old = expand_log(old)
```
0sidharth, line 1012: Is there any reason for using `expand_log`? The above comment should be removed if such a case is there.

sympy/core/mul.py
```diff
@@ -1902,7 +1902,7 @@ def coeff_exp(term, x):
 
         try:
             for t in self.args:
-                coeff, exp = t.leadterm(x, logx=logx)
+                coeff, exp = t.leadterm(x)
```
0sidharth, line 1905: Why should the `logx` parameter not be passed? Is it not required, or giving a wrong result?

sympy/functions/elementary/exponential.py
```diff
@@ -1007,15 +1009,22 @@ def coeff_exp(term, x):
             while s.is_Order:
                 n += 1
                 s = arg.nseries(x, n=n, logx=logx)
-        a, b = s.removeO().leadterm(x)
+        if logx.has(x) and logx != log(x):
```
0sidharth, line 1012: Is the `and` clause really required? When is `logx` = `log(x)`?

sympy/functions/elementary/exponential.py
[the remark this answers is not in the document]
0sidharth: I think, we can try with the above snippet anutosh posted, with `force=True`, and see if that works. `expand_log` can be more expensive.

sympy/functions/elementary/tests/test_exponential.py
```diff
@@ -484,6 +484,25 @@ def test_log_apply_evalf():
     assert value.epsilon_eq(Float("0.58496250072115618145373"))
 
 
+def test_log_leading_term():
+    p = Symbol('p')
+
+    # Test for STEP 3
+    assert log(1 + x + x**2).as_leading_term(x, cdir = 1) == x
```
0sidharth, line 491: There should be no spaces around '=' in function argument acc to PEP8

sympy/functions/elementary/tests/test_exponential.py
```diff
@@ -484,6 +484,25 @@ def test_log_apply_evalf():
...
+    # Test for STEP 4
+    assert log(2*x).as_leading_term(x, cdir = 1) == log(x) + log(2)
+    assert log(2*x).as_leading_term(x, cdir = -1) == log(x) + log(2)
+    assert log(-2*x).as_leading_term(x, cdir = 1, logx = p) == p + log(2) + I*pi
+    assert log(-2*x).as_leading_term(x, cdir = -1, logx = p) == p + log(2) - I*pi
+    # Test for STEP 5
+    assert log(-2*x + (3 - I)*x**2).as_leading_term(x, cdir = 1) == log(x) + log(2) - I*pi
+    assert log(2*x + (3 - I)*x**2).as_leading_term(x, cdir = -1) == log(x) + log(2) - 2*I*pi
```
0sidharth, line 499: Is this supposed to be missing a '-' sign in front of 2*x?

anutosh491: No ,the real essence of STEP 5 comes into picture for `log(2*x + (3 - I)*x**2)` through any negative `cdir`. Yeah it might look weird because I've added both tests(the upper one and this) which would go through STEP 5 and neglected the counterparts which would skip the step . I guess I'll add them in the next commit to make more sense , as those tests would then act as controls and would showcase cases which won't need STEP 5

sympy/core/add.py
anutosh491: No not really , had introduced that change to see if it helps in fixing bugs in `add._eval_as_leading_term()`, had planned to change it back after I get some clarity( done in latest commit)

sympy/core/mul.py
anutosh491: Yeah it is not required and also would give many wrong result . It went unnoticed as we weren't exactly dealing with the `logx` parameter, but now that the PR introduces the real essence of `logx`, having a `logx = logx` there changes the path/execution of calls involving expressions of type `Mul` where `logx` is not `None`

anutosh491:
Changes introduces in last few commits .
1) After fixing issues related to `add._eval_as_leading_term()`, few tests started failing in `test_bessel.py` . Good that they failed as leading term for `bessely` was buggy which has been fixed now . For eg
```
>>> # On master  All results here are wrong due to minute errors in leading term method
>>> bessely(0, x).as_leading_term(x)
(2*log(x) - 2*log(2))/pi
>>> limit(bessely(1, x), x, 0)
0
>>> limit(bessely(2, x), x, 0)
oo
>>> limit(bessely(3, x), x, 0)
oo
>>> limit(bessely(-1, x), x, 0)
0
>>> limit(bessely(-2, x), x, 0)
oo
>>> limit(bessely(-3, x), x, 0)
-oo
>>>
>>> # On branch
>>> bessely(0, x).as_leading_term(x)
(2*log(x) - 2*log(2) + 2*EulerGamma)/pi
>>> limit(bessely(1, x), x, 0)
-oo
>>> limit(bessely(2, x), x, 0)
-oo
>>> limit(bessely(3, x), x, 0)
-oo
>>> limit(bessely(-1, x), x, 0)
oo
>>> limit(bessely(-2, x), x, 0)
-oo
>>> limit(bessely(-3, x), x, 0)
oo
````
2) Another thing we might have to consider is that after introducing this change in the leading term method for add
```
        if logx is None:
            _logx = Dummy('logx')
            lgx = _logx
        else:
            lgx = logx
        leading_terms = [t.as_leading_term(x, logx=lgx, cdir=cdir) for t in expr.args]
```
Expressions involving multiple log arguments might be evaluated differently . we won't get wrong results as such but consider a test in `test_order.py`
```
>>> O(log(x) + 1/log(x))
O(log(x))
>>> O(log(x) + 1/log(x))
O((log(x)**2 + 1)/log(x))
```
The second result isn't wrong either and is the only test concerning the change as of now , we could maybe look into this more carefully if required !

File: sympy/functions/elementary/exponential.py (hunk 5 above)
Comment (anutosh491) [reply sympy/functions/elementary/exponential.py:1012]: Not really , I would say no , actually that shoudn't be there , put it just by choice . From what I understand the parameter `logx` should be replacing all instances of `log(x)` in the series/leading term. For eg
```
>>> (log(x) + log(1+x)).nseries(x,n=3, logx = p)
x - x**2/2 + p + O(x**3)
```
**But quite some times we encounter `log(x)` as the `logx` parameter, which I guess means there shouldn't be any change in the expression correct ?( like replacing log(x) by log(x) )**
```
 >>> (log(x) + log(1+x)).nseries(x,n=3, logx = log(x))
log(x) + x - x**2/2 + O(x**3)
```
So yeah few tests like the following involve such a change and I had put that check by choice so that the output looks somewhat better
```
>>> log(1 + 1/x).nseries(x, n=5) # calls for log(1 + 1/x)._eval_nseries(x, n=5 , logx=log(x))
log(1/x) + x - x**2/2 + x**3/3 - x**4/4 + O(x**5)    # without the check

x - log(x) - x**2/2 + x**3/3 - x**4/4 + O(x**5)      # with the check 
```
if log(x) replaces log(x) it means there shouldn't be any change from the normal series expansion so the answer should be 
```
>>> log(1 + 1/x).series(x, 0, 5)
log(1/x) + x - x**2/2 + x**3/3 - x**4/4 + O(x**5)
```
Also another case being
```
>>> e = -log(x) + x*(-log(x) + log(sin(2*x))) + log(sin(2*x))
>>> e.series(x, n=5)
log(2*x) - log(x) + x*log(2) - 2*x**2/3 - 2*x**3/3 - 4*x**4/45 + O(x**5)
>>> # The first -log(x) doesn't really take part in the expansion but with the check we get
log(2) + log(2)*x - 2*x**2/3 - 2*x**3/3 - 4*x**4/45 + O(x**5)
```
So yeah , we could do with or without the check !

Comment (0sidharth) [reply sympy/functions/elementary/exponential.py:1012]: The check looks counterintuitive to me, so I think it can be removed. I meant to ask if there are any example cases where we don't explicitly pass `logx=log(x)`, but it gets passed during the run of the code.

Comment (anutosh491) [reply sympy/functions/elementary/exponential.py:1012]: Ahh ok got your point , will look into this . If I remember correctly , there are cases in `test_aseries.py` and `test_nseries.py`, where somehow a `log(x)` is being passed as the parameter. Not sure of its usecase and like where is it happening but I'll look into this soon

File: sympy/functions/elementary/exponential.py (hunk 1 above)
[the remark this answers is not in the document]
Comment (anutosh491) [reply sympy/functions/elementary/exponential.py:1025]: The only case which we would be missing out here would be the ones having `log(1/x)`.For eg As of now we have the following 
```
>>> log(log(1/x))._eval_nseries(x, n= 3, logx=-p)  
log(p) + O(x**3)    # order term is unwanted
 ```
And let's say we stick to something like 
```
        if not d.is_positive:
            res = log(a) + b*logx
            expr = self.expand(**dict(deep=True, log=True, mul=False,
                    power_exp=False, power_base=False, multinomial=False,
                    basic=False, force=True, factor=False))
            if (res.subs(logx, log(x)) == expr
                or res.subs(-logx, -log(x)) == expr):
                return res
```
where `res = log(p)`, `logx= -p` . We would miss out on cases having log(1/x) as shown below 
```
>>> expand_log(log(log(1/x)), force=True)
log(log(x)) + I*pi
```
This sort of implementation would have liked the answer as `log(-log(x))` rather than `log(log(x)) + I*pi`

anutosh491:
I feel we should be able to answer #21721 through the changes introduced here, we are somewhat close . Boils down to this step 

```
>>> (-I*log(-a - I + 1/x)+ I*log(-a + I + 1/x))
-I*log(-a - I + 1/x) + I*log(-a + I + 1/x)
>>> _.as_leading_term(x, cdir =-1)   # expected 0
2*pi*Heaviside(im(a) - 1, 0) - 2*pi*Heaviside(im(a) + 1, 0)
>>>
>>> expr1 = -I*log(-a - I + 1/x)
>>> expr2 = I*log(-a + I + 1/x)
>>> expr1.as_leading_term(x, cdir=-1).expand()
I*log(x) - 2*pi*Heaviside(im(a) + 1, 0) + 2*pi
>>> expr2.as_leading_term(x, cdir=-1).expand()
-I*log(x) + 2*pi*Heaviside(im(a) - 1, 0) - 2*pi
```
The expected result here is `0` , the limit/leading_term doesn't really depend on `a`. I guess the series shows that
```
>>> expr = (-I*log(-a - I + 1/x)+ I*log(-a + I + 1/x))
>>> expr.series(x, 0, cdir=-1)
-2*x + x**3*(2/3 - 2*a**2) + x**4*(-2*a**3 + 2*a) + x**5*(-2*a**4 + 4*a**2 - 2/5) - 2*a*x**2 + O(x**6)
>>> _.as_leading_term(x, cdir =-1)
-2*x
```
Though the algorithm doesn't seem to make any obvious mistakes here, need to think how we can handle such cases !

sympy/functions/elementary/exponential.py
anutosh491: > The check looks counterintuitive to me, so I think it can be removed. I meant to ask if there are any example cases where we don't explicitly pass logx=log(x), but it gets passed during the run of the code.

Actually I realize we would need some kind of check here like `logx != log(x)`. This is because , I had overlooked line `976-977` . 
```
        from sympy.series.order import Order
        from sympy.simplify.simplify import logcombine
        if not logx:
            logx = log(x)
```
So whenever we pass `logx` as `None` originally , it is changed to `log(x)` and the check I've added is for addressing a corner case as follows
```
def test_issue_21227():
    f = log(x)
    assert f.nseries(x, logx=-x) == -x

    f = log(-log(x))
    assert f.nseries(x, logx=-x) == log(x)

    f = log(log(x))
    assert f.nseries(x, logx=-x) == log(-x)
    assert f.nseries(x, logx=x) == log(x)
```
Here `logx` contains an argument based on `x` which is the variable in consideration for the expansion for which there is no handling in the code currently ! The original code is capable to address all other cases through `a, b = s.removeO().leadterm(x)` . We should have the following diff I believe
```
--- a/sympy/functions/elementary/exponential.py
+++ b/sympy/functions/elementary/exponential.py
@@ -973,6 +973,7 @@ def _eval_nseries(self, x, n, logx, cdir=0):
         #      IMPORTANT.
         from sympy.series.order import Order
         from sympy.simplify.simplify import logcombine
+        _logx = logx
         if not logx:
             logx = log(x)
         if self.args[0] == x:
@@ -1009,10 +1010,13 @@ def coeff_exp(term, x):
             while s.is_Order:
                 n += 1
                 s = arg.nseries(x, n=n, logx=logx)
-        if logx.has(x):
+        if _logx and logx.has(x):
             a, b = arg.subs(log(x), logx).as_leading_term(x), S.Zero
         else:
-            a, b = s.removeO().leadterm(x)
+            try:
+                a, b = s.removeO().leadterm(x)
+            except:
+                a, b = s.removeO().as_leading_term(x), S.Zero
```
The last try except block would be for something like this
```
>>> log(sqrt(-x**2 -2*x))._eval_nseries(x, 3, None)
log(sqrt(2)*sqrt(-x)) + x/4 - x**2/16 + O(x**3)
>>> 
>>> # Without the block 
>>> log(sqrt(-x**2 -2*x))._eval_nseries(x, 3, None)"
ValueError:
cannot compute leadterm(-sqrt(2)*x**2*sqrt(-x)/32 +
sqrt(2)*x*sqrt(-x)/4 + sqrt(2)*sqrt(-x), x). The coefficient should
have been free of x but got sqrt(2)*sqrt(-x)
```

sympy/functions/elementary/exponential.py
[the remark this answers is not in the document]
anutosh491: I've addressed this now , everything works fine now . Hence resolving (failing tests not due to the latest commit , due to some other warnings I see)

sympy/integrals/tests/test_integrals.py
```diff
@@ -1893,6 +1893,16 @@ def test_issue_21024():
     assert F_true == integrate(f, x)
 
 
+def test_issue_21721():
+    a = Symbol('a')
+    b = Symbol('b', real=True)
+    expr = -I*(-2*I*pi*Heaviside(im(a) - 1, 0) + 2*I*pi*Heaviside(im(a) + 1, 0))/(2*pi)
+    assert integrate(1/(pi*(1+(x-a)**2)),(x,-oo,oo)) == expr
```
jksuom, line 1900: if we put `integrate(1/(pi*(1+(x-a)**2)),(x,-oo,oo)).expand()` on the left hand side, them `expr` would look more readable.

sympy/integrals/tests/test_integrals.py
```diff
@@ -1893,6 +1893,16 @@ def test_issue_21024():
...
+    assert integrate(1/(pi*(1+(x-b)**2)),(x,-oo,oo)) == 1
+    for a in [-I/2, I/2]:
+        assert integrate(1/(pi*(1+(x-a)**2)),(x,-oo,oo)) == 1
```
jksuom, line 1903: These tests are not needed when the antiderivative is given above.

anutosh491: Sure , will change them . It's just that all other cases were working on master except this one which is `-1< im(a) < 1` . This case used to give `0` on master, so what we really addressed here was this case. But we anyways have the integral in the form of parameter `a`, so we can get rid of this .

sympy/core/mul.py
jksuom, line 1905: Why should logx not be included? A comment would be helpful.

sympy/series/tests/test_limits.py
```diff
@@ -1123,6 +1152,37 @@ def test_issue_22181():
...
+
+    # test https://github.com/sympy/sympy/issues/22220#issuecomment-1077618340
+    expr = expr = (-log(tan(x/2) - I) +log(tan(x/2) + I))
+    assert limit(expr, x, pi, '+') == 2*I*pi
+    assert limit(expr, x, pi, '-') == 0
+
+
+def test_issue_22982():
+    x = Symbol('x', real=True, positive=True)
+    assert limit((log(E + 1/x) - 1)**(1 - sqrt(E + 1/x)), x, oo) == 0
```
jksuom, line 1179: It looks like the limit should be `oo`. (It will suffice to define `x` with `positive=True` as that implies `real=True`.)

anutosh491: Yeah correct , not sure how I missed this . Will look into it !

sympy/functions/elementary/tests/test_trigonometric.py
```diff
@@ -923,6 +922,13 @@ def test_asin_series():
     assert t5 == 3*x**5/40
     assert asin(x).taylor_term(7, x, t5, 0) == 5*x**7/112
 
+def test_asin_leading_term():
+    assert asin(x).as_leading_term(x) == x
+    assert asin(x + 1).as_leading_term(x) == pi/2
+    assert asin(x - 1).as_leading_term(x) == -pi/2
+    assert asin(1/x).as_leading_term(x, cdir=1).expand() == I*log(x) + pi/2 - I*log(2)
```
jksuom, line 929: It is probably better to add `expand` to `asin._eval_as_leading_term` (and to other similar methods where `rewrite(log)` will introduce nontrivial factors).

sympy/functions/elementary/tests/test_trigonometric.py
```diff
@@ -923,6 +922,13 @@ def test_asin_series():
     assert t5 == 3*x**5/40
     assert asin(x).taylor_term(7, x, t5, 0) == 5*x**7/112
 
+def test_asin_leading_term():
+    assert asin(x).as_leading_term(x) == x
+    assert asin(x + 1).as_leading_term(x) == pi/2
+    assert asin(x - 1).as_leading_term(x) == -pi/2
+    assert asin(1/x).as_leading_term(x, cdir=1) == I*log(x) + pi/2 - I*log(2)
+    assert asin(1/x).as_leading_term(x, cdir=-1) == -I*log(x) - 3*pi/2 + I*log(2)
```
jksuom, line 930: It looks like there should be `-pi/2` instead of `-3*pi/2`:
```
In [49]: asin(-2).n()
Out[49]: -1.5707963267949 + 1.31695789692482⋅ⅈ
In [50]: asin(-10).n()
Out[50]: -1.5707963267949 + 2.99322284612638⋅ⅈ
In [51]: asin(-100).n()         
Out[51]: -1.5707963267949 + 5.29829236561048⋅ⅈ
```
It is probably best to comment this out now and fix in a later PR. Something like this can also be expected in case of other inverse functions.

sympy/series/tests/test_order.py
```diff
@@ -157,15 +157,15 @@ def test_add_1():
     assert Order(3*x - 2*x**2) == Order(x)
     assert Order(1 + x) == Order(1, x)
     assert Order(1 + 1/x) == Order(1/x)
-    assert Order(log(x) + 1/log(x)) == Order(log(x))
+    assert Order(log(x) + 1/log(x)) == Order((log(x)**2 + 1)/log(x))
```
jksuom, line 160: This can probably be fixed later in order.py.

sympy/functions/elementary/tests/test_trigonometric.py
anutosh491: I guess it is correct , atleast that's what the graph and wolfram shows ( I hope you didn't forget the negative cdir here ) 
image

Also we can achieve this on master through gruntz 
```
>>> x = Symbol('x')
>>> asin(1/x).rewrite(log)
-I*log(sqrt(1 - 1/x**2) + I/x)
>>> expand_log(_, force=True)
-I*log(sqrt(1 - 1/x**2) + I/x)
>>> gruntz(-I*log(sqrt(1 - 1/x**2) + I/x) + I*log(x), x, 0, '-')
I*(log(2) + 3*I*pi/2)
```

sympy/series/tests/test_order.py
anutosh491: Yeah I remember asking this on the gitter channel , if we could do with this ans . Not sure if I got any suggestion on that . Would this do ? or we should stick to `log(x)` as before . In any case I'll give it a quick try regarding how this can be fixed , otherwise make a TODO for that !

anutosh491: Ahh yeah I remember , this comes out of this line we had added
```
        _logx = Dummy('logx') if logx is None else logx
        leading_terms = [t.as_leading_term(x, logx=_logx, cdir=cdir) for t in expr.args]
```
Treating log as constants means log(x) and 1/log(x) both are rather treated as constants/symbols , hence the second log(x) term isn't eliminated !

sympy/functions/elementary/tests/test_trigonometric.py
jksuom: The difference between `-pi/2` and `-3*pi/2` is accounted for by the imaginary part of `log(x)` for negative `x`. So there is no problem after all..

jksuom:
This is ready to go, thanks!

anutosh491:
Thanks a lot @jksuom and @0sidharth for the reviews !

PR #23595, opened 2022-06-07
Series: Added Leading term and Series Methods for Arg Function

anutosh491:
Added leading term and series methods for Arg function which weren't implemented yet .

#### Release Notes

* series
  * Added leading term and nseries Methods for Arg Function.


anutosh491:
Not sure why `test / optional-dependencies (3.11.0-alpha - 3.11, true)` fails here , closing and reopening to see the reason behind this !

anutosh491:
cc @jksuom . I've implemented the leading term and series method for `arg` function as discussed. Let me know what you see any change required here !

jksuom:
The `arg` function is not complex analytic, so it has no power series (of a complex variable) in general. Only a constant value can be regarded as a power series. It is probably not feasible to use much time to examine whether the function is constant or not. I would suggest that only two values be returned: If `z` (with positive `t` substituted) is positive, then return 0, and if it is negative, then return `pi`. Otherwise raise an error, maybe `NotImplementedError`.
`_eval_as_leading_term` can be implemented this way and then `_eval_nseries` can use it to return the same constant values.

anutosh491:
>  It is probably not feasible to use much time to examine whether the function is constant or not.

Yeah correct, this makes sense . Will address the changes

anutosh491:
I had added these tests though after confirming from Wolfram . 
```
    assert arg(x + I).as_leading_term(x, cdir = 1) == pi/2
    assert arg(x + I).as_leading_term(x, cdir = -1) == pi/2
```
Now something like `limit(arg(x + I), x, 0, '+')` will give a `NotImplementedError` through gruntz and the leading term method if we introduce a change like the following
```
        arg0 = self.args[0]
        t = Dummy('t', positive=True)
        z = arg0.subs(x, cdir*t)
        if z.is_positive:
            return S.Zero
        elif z.is_negative:
            return S.Pi
        else:
            raise NotImplementedError('as_leading_term(%s, %s)' % (self, x))

        from sympy.series.order import Order
        if n <= 0:
            return Order(1)
        return self._eval_as_leading_term(x, logx=logx, cdir=cdir)
```

anutosh491:
I had added these tests though after confirming from Wolfram . 
```
    assert arg(x + I).as_leading_term(x, cdir = 1) == pi/2
    assert arg(x + I).as_leading_term(x, cdir = -1) == pi/2
```
Now something like `limit(arg(x + I), x, 0, '+')` will give a `NotImplementedError` through gruntz and the leading term method if we introduce a change like the following
```
        arg0 = self.args[0]
        t = Dummy('t', positive=True)
        z = arg0.subs(x, cdir*t)
        if z.is_positive:
            return S.Zero
        elif z.is_negative:
            return S.Pi
        else:
            raise NotImplementedError('as_leading_term(%s, %s)' % (self, x))

        from sympy.series.order import Order
        if n <= 0:
            return Order(1)
        return self._eval_as_leading_term(x, logx=logx, cdir=cdir)
```

anutosh491:
cc @jksuom ,would like to have your thoughts on the above comment .

jksuom:
Limit.doit could catch the error and try simple substitution (e.g. x -> 0).

anutosh491:
> Limit.doit could catch the error and try simple substitution (e.g. x -> 0).

Yeah heuristics could do that substitution for us but we should maybe use a `PoleError` in that case. So that the error can be caught from line 362 onwards
https://github.com/sympy/sympy/blob/ce53e7b5aeee7019cf019b9e7cbea07fe3f8cbac/sympy/series/limits.py#L362-L365

As we don't explicitly catch `NotImplementedError` after the `gruntz` call and before `heuristics`... and as there won't be any series particularly (other than when `z` is `positive` and `negative`)  , a `PoleError` won't be a bad choice here.

```
>>> arg(x + I).nseries(x)
sympy.core.function.PoleError: Cannot expand arg(x + I) around 0
```

oscarbenjamin:
@anutosh491 what's the status of this?

anutosh491:
cc @arnabnandikgp I think this was almost done from my side. Would be really good if we could revamp this. Could you please give it a thorough review and let me know if any thing needs to be addressed here.

arnabnandikgp:
I was looking into this PR a few days ago, there isn't much left to do, this looks fine to me.  As for computing things like limits containing args, the use of leading term routines barely takes place as things get sorted in the preliminary heuristics in limits.py .

anutosh491:
Closing and opening to re-run tests

anutosh491:
cc @oscarbenjamin I think this looks reasonable to me. Kalevi made me do this (I forgot the exact use-case) 2 years back during my GSoC but I didn't get his final verdict on this. Could you review and let Arnab and me know if something is left here.

arnabnandikgp:
maybe you need to rebase to current master as the authors and docs files have been changed in the recent releases.

oscarbenjamin:
Looks good to me.

anutosh491:
> Looks good to me.

Thanks for the review. Merging !

